Issue #23991, opened 2017-01-13
Opening a Vector Geopackage in QGIS alters its file even when no changes are done by user


qgib:
Author Name: **Johannes Kroeger** (Johannes Kroeger)
Original Redmine Issue: [16076](https://issues.qgis.org/issues/16076)
Affected QGIS version: 3.6.0
Redmine category:unknown

---
Opening a Vector Geopackage in QGIS alters its file. Simply loading the file is enough, you do not even need to load any actual data (visibly to the user).

```
$ wget http://www.geopackage.org/data/geonames_belgium.gpkg
2017-01-13 17:58:54 (1.28 MB/s) - ‘geonames_belgium.gpkg’ saved [11594752/11594752]

$ cp geonames_belgium.gpkg geonames_belgium.gpkg_original

$ md5sum geonames_belgium.gpkg*

4dacf68ae3a5f2c2fb55a35d062dc580  geonames_belgium.gpkg
4dacf68ae3a5f2c2fb55a35d062dc580  geonames_belgium.gpkg_original

$ ogrinfo -so geonames_belgium.gpkg
INFO: Open of `geonames_belgium.gpkg'
      using driver `GPKG' successful.
1: administrative (3D Measured Point)
2: hydrography (3D Measured Point)
3: leisure (3D Measured Point)
4: places (3D Measured Point)
5: roads (3D Measured Point)
6: spots (3D Measured Point)
7: terrain (3D Measured Point)
8: undersea (3D Measured Point)
9: vegetation (3D Measured Point)

$ md5sum geonames_belgium.gpkg*
4dacf68ae3a5f2c2fb55a35d062dc580  geonames_belgium.gpkg
4dacf68ae3a5f2c2fb55a35d062dc580  geonames_belgium.gpkg_original

hannes@hilmar:/mnt/earx/geodata$ qgis geonames_belgium.gpkg
Warning: loading of qgis translation failed [/usr/share/qgis/i18n//qgis_en_US]
Warning: loading of qt translation failed [/usr/share/qt/translations/qt_en_US]
Warning: QCss::Parser - Failed to load file  "/style.qss"

```

Here the layer chooser dialog appeared on which I clicked Cancel and quit QGIS.

```
$ md5sum geonames_belgium.gpkg*
99fad69efe058b1d308088b96d60e915  geonames_belgium.gpkg
4dacf68ae3a5f2c2fb55a35d062dc580  geonames_belgium.gpkg_original

``` 

[...]

PR #30647, opened 2019-07-09
Update 'world' easter egg to work with GeoPackage

arongergely:
## Description
The _world_ easter egg is used by some of us in non-easter eggy ways: 
To quickly get a layer when teaching qgis, to create overview maps in map layouts or to get quickly orientated within a project, etc.

The GeoPackage, besides being more modern could also serve as a container for a few other future
'demonstration'  data.

This PR switches out the `world_map.shp` and `world_map.qml` for a GeoPackage. 
The GeoPackage was prepared using the above mentioned shapefile + QML styling. 

## Checklist


nyalldawson:
This is also used in QgsCoordinateBoundsPreviewMapWidget.

Any chance we could get the states and provinces lines in here too? They would be very useful for the CRS bounds preview.

arongergely:
Thanks @nyalldawson, 

Sound like a good idea, I could do that -  would be best to include it as a separate layer in the GeoPackage, otherwise it could ruin the 'world' for some of us (sorry for the pun)

However I have not found any suitable sources so far. They are either too detailed, incomplete, or not free for commercial use. 

The closest to what I would be happy to include is [Natural Earth's global Admin level 1 data](https://www.naturalearthdata.com/downloads/10m-cultural-vectors/10m-admin-1-states-provinces/).
BUT at  10m generalization it is too detailed and their higher-generalized counterparts (at 50m, 110m) are very incomplete.

nirvn:
Adding political boundaries will likely create some undesirable tensions (disputed border etc). Might be better to stay away from that :)

arongergely:
@nirvn thanks, great point! But to be fair, in a few cases the current world_map.shp is already doing that I think.
Or do you mean not to introduce more admin boundary data as that would just introduce additional topics of dispute?

raymondnijssen:
@nyalldawson What about using the OSM online map as a background for the projection extents, as it is already being used in the About dialog? I know it is inconvenient for users without an internet connection, but is would be very nice for all the other users having a more detailed background map. And you leave the conflicted area problem to the "data people" in this case at OSM.

For the world map in QGIS I'd like to propose to run a dissolve to remove all country borders and end up with a just land/water type of map and stay out of border trouble.

nyalldawson:
@nirvn 
> Adding political boundaries will likely create some undesirable tensions (disputed border etc). Might be better to stay away from that :)

Can we add the NE disputed boundaries layer too then?

@arongergely 
> The closest to what I would be happy to include is Natural Earth's global Admin level 1 data.
BUT at 10m generalization it is too detailed and their higher-generalized counterparts (at 50m, 110m) are very incomplete.

I believe the existing layer is just a simplified version of the NE countries layer. We could do the same with this.

@raymondnijssen 
> What about using the OSM online map as a background for the projection extents, as it is already being used in the About dialog? I know it is inconvenient for users without an internet connection, but is would be very nice for all the other users having a more detailed background map. And you leave the conflicted area problem to the "data people" in this case at OSM.

I'm a -1 to this, for a couple of reasons:
- I think OSM is far too detailed for the simple preview maps, and will be too visually "noisy" and detract from the highlight rectangle
- It will also convey a false sense of accuracy to the bound previews -- e.g. by allowing zoom down to a street level, it will appear as though the projection bounds are accurate to this level (when they are not)
- I would very much like to add a second default template project out-of-the-box (we ship one already with OSM) which gives users a nicely styled vector countries and states map. (For this reason I'd be in favour of shipping with the 10m NE countries, states and disputed boundaries layers in their original detail.) I think this would be a great aid for completely new users -- on their first load of QGIS they'll then be given a choice between creating a project with an OSM background or a "world countries" map.

> For the world map in QGIS I'd like to propose to run a dissolve to remove all country borders and end up with a just land/water type of map and stay out of border trouble.

-1 to this - a dissolved version makes the projection selector useless, because you'd have no way to differentiate african/european/south american countries. It'd make the selector only useful for Australia and New Zealand :wink:

raymondnijssen:
@nyalldawson From a UX perspective I completely agree with you. Borders, including admin2 level would be very helpful while selecting reference systems. I just wanted to avoid future discussions about disputed state borders. But adding those separately and display them as dashed lines for example, could also solve the problem. I'm fine with that.

On the other hand I wonder if we should ship data at all with QGIS. I would rather have a robust framework for resources, similar to the plugin repository. It could hold symbol markers (svg+png), templates, palettes, data (layers), north arrows, styles and so on, And it should be more native than the Resources plugin. Yesterday, @arongergely and I were discussing writing a QEP for this

m-kuhn:
> On the other hand I wonder if we should ship data at all with QGIS. I would rather have a robust framework for resources, similar to the plugin repository. It could hold symbol markers (svg+png), templates, palettes, data (layers), north arrows, styles and so on, And it should be more native than the Resources plugin. Yesterday, @arongergely and I were discussing writing a QEP for this

That sounds like a great plan. But I wonder if it's really a contradiction to add some very basic data **and** adding a native resource integration.

Having some worldwide data available "in your face" in the templates section when first launching QGIS sounds very good to me. And if there's a link below "search for more templates in the online resource database" would bring us just a bit closer to perfection (with a good amount of additional work involved of course) ;)

Data-wise I think political borders would be desirable too (most of the world has them worked out clearly and it's just a fact that we live in this man-made world and not just between water and land) but including the disputed boundaries to give a lower attack surface.

Saijin-Naib:
I can curate admin boundaries for the USA from Cenus/TIGER data. From my reading of the license, this should be an appropriate usage of that data.

They also provide hydro, transport, etc, so we can give a lot more context/detail.

It doesn't hurt that I've been compiling a country-wide aggregate TIGER dataset for basically this purpose for myself...

nyalldawson:
> That sounds like a great plan. But I wonder if it's really a contradiction to add some very basic data and adding a native resource integration.

I'm with @m-kuhn here -- it's a good idea, but not for this particular use case.

If you get the data packaged, I'm happy to do the follow up work to add the world map template AND allow this to work correctly when saving/restoring projects (since we'd need to handle this as a special case, given that the QGIS install folder containing the files isn't in a fixed location).

nirvn:
@nyalldawson , yeah, adding that disputed borders dataset would likely be enough on our part to avoid starting WWIII ;)

nyalldawson (changes requested):
as above, please include state and disputed border layers

arongergely:
@Saijin-Naib 
> I can curate admin boundaries for the USA from Cenus/TIGER data. From my reading of the license, this should be an appropriate usage of that data.

Thanks for offering. I think the US States datasets are easy to come by and they are already included in the countries/provinces of [Natural Earth](https://www.naturalearthdata.com/). 
I would prefer to rather "include" data (public domain data!) in this GeoPackage than process it / have it processed by others. This way it is easy to point to the source and version of the data, and there would be no licensing problem. 

@nyalldawson 
> If you get the data packaged, I'm happy to do the follow up work to add the world map template AND allow this to work correctly when saving/restoring projects (since we'd need to handle this as a special case, given that the QGIS install folder containing the files isn't in a fixed location).

Ok thanks, agreed. 

@m-kuhn @nyalldawson @raymondnijssen 
Raymond already mentioned that we discussed writing a QEP for native resource sharing.
My original thought was having easy-to-invoke data on hand for test,teaching, etc in this GeoPackage. 
Apart from that I share Raymond's views and I am happy to help working towards 'native' resource sharing.

@nyalldawson  I will update the GeoPackage then with 

- Disputed Borders
- States / Provinces 
- Countries (correct topology)

Since the plan is to have these included in project templates,
I would use the 10m generalized data from Natural Earth and point to the source + version.

Saijin-Naib:
> 
> Thanks for offering. I think the US States datasets are easy to come by and they are already included in the countries/provinces of Natural Earth.
> I would prefer to rather "include" data (public domain data!) in this GeoPackage than process it / have it processed by others. This way it is easy to point to the source and version of the data, and there would be no licensing problem.

Sounds reasonable. I look forward to seeing what all gets included. I don't know if it is out of scope, but especially where I am from in New York State, having hydrological features is incredibly helpful in helping locate yourself if the basemap will have no labels and will be fairly large scale, so I hope the Natural Earth physical features can be integrated as well along with the Cultural dataset. Thanks for doing this all, by the way.

arongergely:
Updated the GeoPackage as per above, the current size is 37mb after VACUUM. Is that acceptable?

I added defaults stylings for new layers. The 3 layers together now look like this:
world_map

nyalldawson:
Any objections to a 37mb inclusion? I'm +0, but I'm also admittedly focused on the benefits of including this rather than any potential impact to packagers...

m-kuhn:
+0 too, it can still be shipped in a separate package.

kannes:
1) 37 MB is massive for little gain in the big picture (in my opinion). But how about compressing it and using GDAL's vsi[zip|gzip] feature? That should half the size at least. No idea about performance considerations, I guess it would ruin the benefits of spatial or other indexes.

2) Please lowercase the (ugly SHPly) attribute field names.

arongergely:
@kannes, ok. 

> 1. 37 MB is massive for little gain in the big picture (in my opinion). But how about compressing it and using GDAL's vsi[zip|gzip] feature? That should half the size at least. No idea about performance considerations, I guess it would ruin the benefits of spatial or other indexes.

I could perhaps switch out the data for a more highly-generalized one. 

If I were to use the [1:50m gen. level data from NE,](https://www.naturalearthdata.com/downloads/50m-cultural-vectors/) the gpkg size would be almost 1/10th of the current one, I think. Downside would be that the states/provinces layer would be very incomplete- has only the US and Canada states/provinces. (That is why I avoided it so far) 

@nyalldawson, @m-kuhn what do you think? 
Would we be able to live with that for now?

I am also intentionally trying to avoid to do the generalization myself, and rather include readily-available data as possible. But it remains an option..


> 2. Please lowercase the (ugly SHPly) attribute field names.


Ok, can do - once/if we resolve the size question related to your 1st point

nyalldawson:
> has only the US and Canada states/provinces. (That is why I avoided it so far)

hmm - that's certainly of limited use!

arongergely:
Another Idea: Since more than half of this geopackage size is just the states/provinces, could remove provinces of countries that are small (could set threshold based on total area?). Also throwing away some of the attributes fields could save space.

nyalldawson:
@arongergely that works for me. We could drop ALL the attributes here.

arongergely:
@nyalldawson. Ok, I would not mind doing that, perhaps keeping the country names / codes if anyone would want to do some labelling. But otherwise the attribute tables are massive, and even redundant here and there as well.

kannes:
When I first saw this, I assumed it was just geometries. But the initial post made a case for using this always available data for different needs and that sounds like a great idea. If you drop attributes, please keep appropriate identifiers (eg `iso_` and `wb_`) so other data can easily be joined to it.

Hey, this data could even be loaded by default instead of the recent project list (as _optional_ startup mode). I use "world" all the time for quick reference data.

arongergely:
Sorry for the force-push: Instead of new commit,  I amended my last one of the geopackage as I figured it would be better not to have that 37mb geopackage binary in the history.

I have amended the GeoPackage to incorporate the latest ideas on size reduction:
- Removed attribute table fields for all layers, except the names and identification codes (ISO, etc)
- states/provinces now only contain data for the top 15 countries by area

The new size is **17.9MB** after VACUUM; so about a 50% redution in size. I am not sure whether this size is forgivable.

Here is an updated quick view of the layers:
world_map

nyalldawson:
Sounds great to me! Any objections?

kannes:
I'd go either all or none for the states/provinces.

nirvn:
Is there indeed a reason why a fair amount of countries don't have 1st level boundaries? For e.g., mainland Southeast Asia has none, that's odd.

arongergely:
@nirvn, please see my commit message for the answer. 

@kannes.
Thanks, Well, this is me trying to find a middle ground and make everyone happy. We already been at the point without  provinces data or full (as complete as reasonably possible) provices data.  

It was apparent that we would have to make a compromise somewhere (size, level of detail, scope). 

Will wait and read more opinions, for now.

nyalldawson:
How's the size if we go with no attributes, all states? I'd like to see this merged before 3.10 if we can!

arongergely:
@nyalldawson keeping the full datasets but dropping all attributes (except FID): 

**32.3 mb** after VACUUM 
... vs ...
**37.0 mb** of the same but with all attributes. 

So nut much of a relief. There are just a lot of vertices. Particularly in the states / provinces layer.

I was trying to avoid editing the data but since it creeped up on me:

Should I look into simplifying the states / provinces layer, then also run a dissolve on it per country?
We could then hopefully arrive at a reduced vertex count for the countries and states / provinces layer. And the two layers should align then. But I can't make a promise on topology for now.. :(

m-kuhn:
What if you run a snap to grid algorithm on it to reduce precision and vertex density? That should preserve topology.

arongergely:
@m-kuhn thanks, I tried to snap to grid on 0.001 degree precision (working in WGS84). I fiddled around with the tolerance a bit but I did not get any reduction in size, I figure the snapped coordinate values still get stored with the same bit sizes. Perhaps doing this would save space on the spatial index only?

However apart from this, I am getting somewhere with a topologically correct simplification thanks to GRASS **v.generalize** . Stand by for my next comment soon

arongergely:
After some fiddling I managed to do a topologically correct simplificationon the 1:10m Natural Earth layers.

Encouraged by a [minimalistic demo](https://www.youtube.com/watch?v=h8CFi7WMBx0) by Faunalia, I took GRASS **v.generalize** as a basis. 


### Processing steps and QGIS model
What I ended up doing:
1) Taking 1:10m NE states/provinces layer as input rerpoject to a meter-based CRS. I chose Robinson
2) run *v.generalize* with the default simplification algorithm (set threshold to 1000m; remove polygons that are <1km2)
3) Run "Fix geometries" to sort out a few invalid polygons
4) clip the fixed geometries to a safe valid coordinate space of WGS84 (I used -179.9, 179.9, -89.9, 89.9)
This was necessary to avoid polygons that cross the entire world map due to being cut half by the 180deg meridian.
5) Dissolve to adm 1 level (becasue v.generalize broke all multipart geometries into singleparts)
6) for the countries layer, dissolve again on country level

I built a QGIS model of these steps to make this process reviewable and repeatable:

model

I can't paste the model file itself here, but I am happy to share with anyone who would like to have it.


### The simplified layers in GeoPackage 
At this point the states_provinces and countries together are **16.1mb** in a GeoPackage.
Here is 

a visual comparison of the simplified states_provinces layer (red) and the original (blue):

comparison_world
A bunch of islands mainly in the Indian and Pacific Oceans have fallen victim of the 1km2 area threshold.





Here is a close-up on the islands of Croatia:
comparison_croatia



Another one centered around Wellington, New Zealand:
comparison_new_zealand






### One step further

I figured since I am altering the data why not:

1) clean up the attribute table of the states_provinces layer since it has dozens of fields that are of limited use
2) since the countries layer is now derived from the states_provinces, it is lacking useful country codes and names,
so lets join and copy those in from the original 1:10m NE countries layer


Finally I copied in the 1:10m NE disputed borders layer. Note that it now does not line up *perfectly* with the borders in the other layers.


### Result
The current GeoPackage size is around *12.7mb*. 

In this We have 

- states_provinces and countries that fully align, topologically correct and even have some useful attributes.
- disputed borders that do not align 100% but they are there and we show the complete story this way
This is how it looks styled:  

result

Slightly misaligned disputed border in Moldova:
result_moldove_disputed_border

Let me know your opinion, also if you want the QGIS model above. Based on the reactions  I can push the updated GeoPackage.

nyalldawson:
That sounds (and looks) perfect to me. I've no issue with the simplification and lack of alignment at large scales --  it's totally fine for the intended purpose of this layer.

And 13mb is fine! Great work!

I'd like to see the model included in this PR so that we can easily update when new NE data is released.

The only other thing I'd like to see is the disputed borders restyled to avoid them standing out so much. I'd suggest they should be the same color as normal borders, but in a dotted or dashed style instead. (Either that or only show them at large scales, so they aren't so visually dominant at a global level).

arongergely:
I pushed the updated GPKG - including the styling for disputed borders. This is the new look: 
image

However I am not so sure whether taking up this layer's management within the QGIS movement is a good Idea.. I mean, starting managing data, and hte data processing pipeline, etc.

Surely I am happy to share that model I quickly put together, but If we go down this path, fast forward to adding more layers and we would have to see to maintaining more, possibly different datasources, have to deal with new wishes on data, etc.. right now it does not seem to me like a good idea. Maybe we can discuss during hackfest?

nyalldawson:
Let's do this -- there's been no objections over the lengthy review period, and there's MANY benefits here.

3nids:
I don't know if it's caused by this PR, but with master as soons as I open QGIS `resources/data/world_map.gpkg` gets modified.

saberraz:
> I don't know if it's caused by this PR, but with master as soons as I open QGIS `resources/data/world_map.gpkg` gets modified.
See this https://github.com/qgis/QGIS/issues/31044
or
https://github.com/qgis/QGIS/issues/23991

PR #30657, opened 2019-07-10
[expressions] fix comparison of intervals

3nids:
This basically fixes this expression evaluation:  `to_interval('1 minute') < to_interval('20 days')` which currently return false on master.

I looked into `testqgsexpression.cpp` to add this one...but I fail a bit to understand where to put it. Shall I create a new method for this (like `test_date_comparison`)?


roya0045:
Just my grain of salt, but I think it would be simpler to implement the test in `test_qgsexpresssion.py`

nyalldawson:
I'd add the tests in here:

    https://github.com/qgis/QGIS/blob/master/tests/src/core/testqgsexpression.cpp#L593

and make sure you've got equality, greater than, less than, greater than equal to, etc...

PR #30662, opened 2019-07-10
[oauth2] Fix network error crash; guard reply ptr; remove error phrase

dakcarto:
## Description
When processing the network error slot, access to a casted reply object's members/properties causes crash if the reply is deleted elsewhere. Guard cast reply object using QPointer and access all properties immediately.

While this works for all scenarios that I tested, including very fast zoom in/out operations, it doesn't address the deletion of the reply object elsewhere, which I suppose can still cause issues. Really, this fix only attempts to quickly access the reply object before such deletions occur.

## Checklist

### DIFF
--- a/src/auth/oauth2/qgsauthoauth2method.cpp
+++ b/src/auth/oauth2/qgsauthoauth2method.cpp
@@ -33,6 +33,7 @@
 #include <QDesktopServices>
 #include <QDir>
 #include <QEventLoop>
+#include <QPointer>
 #include <QString>
 #include <QMutexLocker>
 
@@ -435,8 +436,8 @@ void QgsAuthOAuth2Method::onNetworkError( QNetworkReply::NetworkError err )
 {
   QMutexLocker locker( &mNetworkRequestMutex );
   QString msg;
-  QNetworkReply *reply = qobject_cast<QNetworkReply *>( sender() );
-  if ( !reply )
+  QPointer<QNetworkReply> reply = qobject_cast<QNetworkReply *>( sender() );
+  if ( reply.isNull() )
   {
 #ifdef QGISDEBUG
     msg = tr( "Network error but no reply object accessible" );
@@ -444,16 +445,19 @@ void QgsAuthOAuth2Method::onNetworkError( QNetworkReply::NetworkError err )
 #endif
     return;
   }
+
+  // Grab some reply properties before object is deleted elsewhere
+  QVariant replyStatus = reply->attribute( QNetworkRequest::HttpStatusCodeAttribute );
+  QVariant replyAuthProp = reply->property( "authcfg" );
+  const QString replyErrString = reply->errorString();
+
   if ( err != QNetworkReply::NoError && err != QNetworkReply::OperationCanceledError )
   {
-    msg = tr( "Network error: %1" ).arg( reply->errorString() );
+    msg = tr( "Network error: %1" ).arg( replyErrString );
     QgsMessageLog::logMessage( msg, AUTH_METHOD_KEY, Qgis::MessageLevel::Warning );
   }
 
-  // TODO: update debug messages to output to QGIS
-
-  QVariant status = reply->attribute( QNetworkRequest::HttpStatusCodeAttribute );
-  if ( !status.isValid() )
+  if ( !replyStatus.isValid() )
   {
     if ( err != QNetworkReply::OperationCanceledError )
     {
@@ -462,23 +466,20 @@ void QgsAuthOAuth2Method::onNetworkError( QNetworkReply::NetworkError err )
     }
     return;
   }
-  QVariant phrase = reply->attribute( QNetworkRequest::HttpReasonPhraseAttribute );
-  if ( phrase.isValid() )
-  {
-    if ( err != QNetworkReply::OperationCanceledError )
-    {
-      msg = tr( "Network error, HTTP status: %1" ).arg( phrase.toString() );
-      QgsMessageLog::logMessage( msg, AUTH_METHOD_KEY, Qgis::MessageLevel::Info );
-    }
-  }
 
-
-  if ( status.toInt() == 401 )
+  if ( replyStatus.toInt() == 401 )
   {
     msg = tr( "Attempting token refresh…" );
     QgsMessageLog::logMessage( msg, AUTH_METHOD_KEY, Qgis::MessageLevel::Info );
 
-    QString authcfg = reply->property( "authcfg" ).toString();
+
+    if ( !replyAuthProp.isValid() )
+    {
+      msg = tr( "Token refresh FAILED: authcfg property invalid" );
+      QgsMessageLog::logMessage( msg, AUTH_METHOD_KEY, Qgis::MessageLevel::Warning );
+      return;
+    }
+    QString authcfg = replyAuthProp.toString();
     if ( authcfg.isEmpty() )
     {
       msg = tr( "Token refresh FAILED: authcfg empty" );


dakcarto:
@elpaso, @nyalldawson or anyone else,

How can one protect against a cast `sender()` object in a slot that might be deleted elsewhere? I don't want to explicitly use `deleteLater()`, since that is not appropriate in the authentication method, nor do I think it will 'pause' the object's deletion until the slot has completed.

I could memory copy the reply object somehow, though that seems hacky, and it would mean copying a `QObject` that doesn't have a public ctor (`QNetworkReply` objects can not be directly created).

nyalldawson:
@dakcarto I personally think we need to rethink this whole approach. If I read the code correctly, this is connecting to any request made to catch errors, and use these to detect if the token has been revoked on the server (and to trigger a new token refresh). But I can't see anyway to make this fully robust, and we're just patching short term workarounds on a fragile design.

Even copying the reply won't help (if that were possible), because it could be deleted mid-way through a copy by the other thread.

I think we ultimately need to kill the updateNetworkReply method, as I can't see how we can fix this.

But that still leaves the original issue of handling revoked tokens gracefully. Just brainstorming, but maybe an alternative approach here would be to handle all this in updateNetworkRequest. We could partially copy the request, send it off as a "ping" to the server to check the response header alone (without getting the full response content) , and if this trial request fails then trigger the token refresh. This would all happen while blocking `updateNetworkRequest`, so the original caller will be stalled while this is occurring. We could put a timer check in here so that we only do the "ping" check every XXX minutes to avoid unwanted network activity.

What do you think?

dakcarto:
@nyalldawson wrote:
> @dakcarto I personally think we need to rethink this whole approach. If I read the code correctly, this is connecting to any request made to catch errors, and use these to detect if the token has been revoked on the server (and to trigger a new token refresh). But I can't see anyway to make this fully robust, and we're just patching short term workarounds on a fragile design.

Agreed, it is a fragile design, but it is also currently causing a crash in 3.4 and 3.8 releases. I recommend we push this change, then work on a refactor.

> Even copying the reply won't help (if that were possible), because it could be deleted mid-way through a copy by the other thread.

Yeah, I could not find a way that could be assured to succeed.

> I think we ultimately need to kill the updateNetworkReply method, as I can't see how we can fix this.

You mean for this particular auth method plugin, or for all of them (i.e. the auth manager's abstracted call as well)? I agree with the former, though the latter is still a valid method for auth method plugins, since `updateNetworkReply` _should_ be used directly after getting such a reply from a network operation (not in a slot or some time later).

There are certainly valid reasons for wanting to inspect the reply object from within the auth method, e.g. so things can be cached during a session. OAuth2 just happens to be the only auth method plugin that utilizes it right now.

> But that still leaves the original issue of handling revoked tokens gracefully. Just brainstorming, but maybe an alternative approach here would be to handle all this in updateNetworkRequest. We could partially copy the request, send it off as a "ping" to the server to check the response header alone (without getting the full response content) , and if this trial request fails then trigger the token refresh. This would all happen while blocking `updateNetworkRequest`, so the original caller will be stalled while this is occurring. We could put a timer check in here so that we only do the "ping" check every XXX minutes to avoid unwanted network activity.

Hmm, that approach, besides blocking as you mentioned, looks tricky, because:

- the URL to ping is not necessarily easily determined from the data source's URL (might have to have general logic per provider)
- some services can disallow HEAD requests (GeoServer for example), so a GET or POST endpoint would need to be known

As opposed to inspecting a ping response, it would be ideal to have an endpoint that can test token validity. Unfortunately, while there appears to be a proposed [standard solution](https://tools.ietf.org/html/rfc7662) for token introspection to the [current per-auth-vendor mess](https://stackoverflow.com/questions/12296017) that currently exists, there is no guarantee it is widely implemented.

> What do you think?

Since the token refresh is really just needs triggered when there is a 401 return status, I propose having a _new_ auth manager method like `updateNetworkReply`, maybe...

```cpp
bool QgsAuthManager::processNetworkReply( QNetworkReply *reply, const QString &authcfg, const QString &dataprovider )
```

- Should only be used when the developer is processing a **finished** reply object, e.g. in a slot that is fired by the reply object's `finished()` signal or similar signal from a network access manager (which passes the reply object).
- Auth method plugins can then be assured they are processing a finished reply object and can cache what makes sense.
- The `authcfg` parameter should be _optional_, as this auth manager call may not reside within a slot of the request-calling object, i.e. the `authcfg` token may be unknown and may not be part of the request properties. The coding pattern should probably be to store the `authcfg` token in the reply object upon creation, using `updateNetworkReply` (like done in OAuth2 auth plugin).

So, I see a reasonable solution using `updateNetworkReply` and `processNetworkReply` together, excepting how to readily block similar failing, threaded requests while doing the token refresh.

Thoughts?

dakcarto:
Followup to last post. Essentially, auth method plugins should _never_ expect to work with or manipulate a reply object from within a slot **connected to one of reply's signals** (as done here in this fragile code). The updating/processing of a reply object should only be done directly, and where it makes sense, in the code flow of the class actually responding to the network response.

nyalldawson:
@dakcarto

> You mean for this particular auth method plugin, or for all of them (i.e. the auth manager's abstracted call as well)? I agree with the former, though the latter is still a valid method for auth method plugins, since updateNetworkReply should be used directly after getting such a reply from a network operation (not in a slot or some time later).

I'm arguing for complete removal of it -- it can't be made safe, so leaving it in the api is just asking for issues.

> the URL to ping is not necessarily easily determined from the data source's URL (might have to have general logic per provider)

I was initially thinking of using the same request URL which triggered the auth request -- although, i guess that's dangerous (e.g. if it was a "delete feature" request!).

> I propose having a new auth manager method like updateNetworkReply, maybe...

dakcarto:
@nyalldawson wrote in response to @dakcarto:
> > You mean for this particular auth method plugin, or for all of them (i.e. the auth manager's abstracted call as well)? I agree with the former, though the latter is still a valid method for auth method plugins, since updateNetworkReply should be used directly after getting such a reply from a network operation (not in a slot or some time later).
> 
> I'm arguing for complete removal of it -- it can't be made safe, so leaving it in the api is just asking for issues.

I see your point here. Though, I think the call is still a valid thing to do, e.g. auth methods parsing a reply. See proposal at end for one way this might be considered safe to use.

> > the URL to ping is not necessarily easily determined from the data source's URL (might have to have general logic per provider)
> 
> I was initially thinking of using the same request URL which triggered the auth request -- although, i guess that's dangerous (e.g. if it was a "delete feature" request!).

Not sure of the best way to do the _up front_ token check across varied types of requests.

> > I propose having a new auth manager method like updateNetworkReply, maybe...
> 
> So this would mean ALL users of the network access manager in QGIS would need to manually call this for their replies, just in case the network request involves authcfg?

Yeah, that would be excessive _and_ require devs to follow a non-intuitive workflow to use the auth system.

---

So, it looks like we should do something I should have done from the beginning with the auth system: **add auth support directly to `QgsNetworkAccessManager`**. This would be similar to what you have already done in `QgsBlockingNetworkRequest`.

Basically, these changes would be needed:

1. Add auth system support to internal operations in `QgsNetworkAccessManager`
   - All network request calls updated to have an `authcfg` and `dataprovider` optional parameters, or just getter/setters. This former means overriding GET, POST, etc. calls as wrapped calls to `QgsNetworkAccessManager::createRequest` with an `authcfg` parameter, but is the simplest to refactor. The `dataprovider` optional parameter allows auth method plugins to branch logic according to known providers, if necessary.
   - If an `authcfg` token is passed in with `QgsNetworkAccessManager::createRequest` then the `QNetworkRequest` object has the token and any `dataprovider` added as a _properties_. In this way any later introspection of the request, like from within a reply object or `QgsNetworkReplyContent`, could easily resolve whether the auth system was called and which configuration was used.
   - Add the proposed network reply process operation **only** to `QgsNetworkAccessManager::onReplyFinished`, [right before emitting `finished`](https://github.com/qgis/QGIS/blob/master/src/core/qgsnetworkaccessmanager.cpp#L297). Also, its signature should be **`bool QgsAuthManager::processNetworkReplyContent( const QgsNetworkReplyContent &replyContent )`**, _instead_ of accepting a pointer to a reply object that could be deleted elsewhere. Both `authcfg` token and any `dataprovider` can be resolved from the `QgsNetworkReplyContent->request()` data.

2. Remove all direct calls to `QgsNetworkAccessManager` update and process methods from all providers and wherever else they are used, and replace with simple passing of `authcfg` token parameter.

Let me know what you think. This is more of a refactoring than a simple bug fix.

---

**The immediate crash in 3.4.9 and 3.8.0 is still there. I feel strongly that users should not unnecessarily suffer it, if this PR fixes it for most situations. These changes should be pushed today so they are in 3.4.10 and 3.8.1, then we can work out a more suitable refactoring.**

nyalldawson:
I'll respond to your proposal layer, but agreed - let's get a bit of extra safety in for the next point releases. Merge away.

PR #30709, opened 2019-07-15
[callouts] Ensure simple callouts are rendered below all map labels

nyalldawson:
...instead of being rendered on top of some. For this callout style,
we don't want callouts overlapping labels (rather the opposite). But
leave API in place to allow other callout styles to render below
their associated labels only, as this may be wanted for some styles
(e.g. balloon style callouts)


nirvn:
@nyalldawson , nice, that takes care of pretty much the only visual issue I had with simple/manhattan callouts.

PR #30727, opened 2019-07-15
[Feature] Symbol Aware Legend Expression Round 2

roya0045:
## Description
This part of the  project is mostly centered around the aggregation, expression and fids storage.

Here are the major changes that have been implemented:

- Changing the fid/symbol input of QgsVectorLayer::aggregate
- Using the counter directly in qgsvectorlayer instead of the stored hashes (removed extra signals, now relying on the counter directly)

This only leaves feature stacking as the last dependency.

I have added the new parameter in the expressions for the following reasons: more easy to control by the user, no need to parse, can leverage the Fids filter.

The parameter was placed before some other existing parameters that are not constantly used, this parameter can be part of the 'basic' set and the others have been moved back, otherwise it would be too complicated to have a permanent argument after semi-permanent arguments (concatenator, order by, etc).

## Checklist


nyalldawson:
@roya0045 nice work getting pt 1 over the line! 

Can you please split this up into the component changes? E.g. the aggregate changes in a separate PR to the rest?

roya0045:
@nyalldawson This might be a hassle but I'll see what I can do. ( splitting commit and fiddling with forks is annoying but I should be able to do something.)

roya0045:
@m-kuhn @nyalldawson this PR will contain only changes regarding the vector layer, this means the aggregate method and how the fids and counts are stored. As requested by Mr. Dawson.

roya0045:
@nyalldawson I just wanted to say that the other PR will never build as it assumes that the input given to the qgsvectorlayer::aggregate method is a string and not a list of fids, whereas the string parameter is done in this PR. Are you sure that splitting them is a wise choice or should I revert the changes?

roya0045:
@nyalldawson I'm getting some weird behaviour with concatenate. I have changed the default value of orderby in aggregatestring but when it comes to concatenate and concatenate unique 4 is provided by the functions.

Now I'm sure how the changes that I have made are only affecting the concatenate functions, this might have to do with the fact that those are the only one with an 'orderby' that is not -1. But I'm not sure how this can remove the commas from most output.

nyalldawson:
@roya0045 I'm more than happy to assist - but I need this PR split up into seperate modular components in order to do so. My brain can only process so much code at the one time :laughing:

roya0045:
@nyalldawson I'm having a hard time to split things for parallel processing as the expression changes are dependent on the vectorlayer changes.

Considering the changes to the vectorlayer I'm mostly awaiting feedback from Matthias as he originally proposed to used the counter instead of storing things on the vectorlayer.

The part that I would need assistance with is the expression parsing. Since you were the one to implement the orderBy parameter, I would assume that you could provide assistance with matters only within `expressionfunction.cpp`, you can disregard the rest of the PR as 50% is documentation or test.

roya0045:
@nyalldawson after debuging, it would appear that somehow when it comes to concatenation, there is always 6 items in the value list, this means that the trigger for the 6th node will always activate. 

I'm not sure if this is intended or if I should find a more robust way to get the symbol_id node ( I find that using the node number is risky, finding something in a hash would be easier).

roya0045:
@nyalldawson somehow during the tests the concatenator node get "" and the orderBy node get "','"
This seems to me that there is an extra node in the first few positions that offsets the rest of the data. I'm not sure how this is happening.

roya0045:
@m-kuhn Looking at how nodes are behaving something weird occurs with string, it might be the concatenator but it might also be how the orderby node works. Things just seem to be shifted somehow, even before I get to do anything. This might be because I have changed parsing unknowingly.

I am thinking that moving the symbol_id node before the orderby node and keeping last might be the best option. But before I do that I wanted to make sure that you agreed about this approach of adding a new node.

The other change should be to the vectorlayer, I followed your proposition of integrating the counter rather than storing things, it seems to works ok.

roya0045:
@nyalldawson ok I have solve the node order issues. But the fact that cdash is broken is not helping me to debug the last test.

roya0045:
I compiled it and something weird is happening with the counter. When I count in the legend the base count is displayed but the symbol count stays N/A. I'm not too sure if the problem is with the counter itself or if the refresh signal is not triggered properly.

roya0045:
Somehow when used as a parameter, the symbol_id doesn't seem to be evaluated and simply passed on, as denoted by https://travis-ci.org/qgis/QGIS/jobs/562174310#L1357 I'm not sure how this can happen.

roya0045:
@m-kuhn bumblefuckery should be over, everything works as intended, I will await your feedback if you have some free time.

roya0045:
There is something strange going on, when I was preserving the counter itself, things evaluated properly but now that I went back to caching it in the layer, it seems that things don't have the time to be fully cached before the tests are executed. 

I'll try to find what might be causing this.

roya0045:
Ok the counting completes but either the signal is not handled properly, the task cannot finish or the values are not assigned somehow..

roya0045:
There is something weird going on with the connections. The featurecounter does its work properly (albeit things are stored under weird names). But somehow the signals indicating the end of the counting are not captured by the vectorlayer and this means that the hahses are not transferred to the vectorlayer.

I'm not sure if this is a timing issue, if this issue is specific to the tests and not the full App or if there is someone weird going on.

roya0045:
@nyalldawson @m-kuhn I have no clue as to why the connections or the slot don't work, this might be a Travis issue.

Should I disable the tests using the fids and simply assume that the layer will not be broken or is there a way to properly cache the count/fids without breaking the api and without using the connections. I thought about exposing the 'OnSymbolsCounted' slot and keeping a pointer to the parent in the counter to trigger that function, but this would not pass review.

roya0045:
@m-kuhn The tests are fine when I force caching directly, I assume this is related to Travis and some VM limitation that might prevent proper slot triggering.

Now that I have validate that this PR work at the expense of exposing a slot, should I make the tests partially work to compensate for travis in order to prevent exposing this slot?

This might not indicate when a change would break this feature...

roya0045:
@m-kuhn commit history has been cleaned up, awaiting further feedback concerning the forced caching. Otherwise the core of the code is stable.
@nyalldawson as you desire the PR was split, the feature caching part did not receive any feedback and neither did the expression changes that are in this PR. Your feedback would be appreciated.

roya0045:
@nyalldawson thanks for merging the id caching PR, now this branch should mostly contain changes to the expression parser and the related docs.

Changes are fairly straightforward and since you are the one most familiar with the expression parser I assume you will be the one to review this when you have the time to.

roya0045:
@nyalldawson @m-kuhn out of curiousity, do you think this has a possibility to be merged before the freeze? can this be freeze exempt (since part of the feature is already in) ? Or will have to way for the next round?

nyalldawson:
There's still a lot of changes required before this PR is in a merge-ready state, so I don't think a freeze exemption is applicable. I suggest we focus on getting https://github.com/qgis/QGIS/pull/9915 merged (can happen during freeze), and https://github.com/qgis/QGIS/pull/30297 merged before freeze deadline on Friday (i'll re-review #30297 later today/early tomorrow)

roya0045:
Very well, as for PR 30297 it can wait, I might have some way to test this branch next week or the weeks afterward to tweak things on my end ( I can't use my free time much these days). 
As for #9915 I'm ok with it getting priority, I'm not sure if @rldhont intended to review it also or simply wanted me to clean up the commits.

src/core/qgsvectorlayer.h
```diff
@@ -2229,6 +2248,9 @@ class CORE_EXPORT QgsVectorLayer : public QgsMapLayer, public QgsExpressionConte
 
     bool accept( QgsStyleEntityVisitorInterface *visitor ) const override;
 
+    //! cache the results of the feature counter
+    void onSymbolsCounted() SIP_SKIP;
```
roya0045: @nyalldawson @m-kuhn when I talked about forcing caching in my comments it was related to exposing this function as public and calling it in the tests. Otherwise travis tests failed since that slot/function was never called despite the connection. I am aware that this is bad practice and should not remain but I wanted to prove that my changes worked properly with the tests that I made. As stated if this is removed, the PR should still work but the travis tests will not be able to properly indicate when something fails as this slot never got triggered.

What do you propose on this issue?

roya0045:
Tests confirm that forcing caching in travis is the only way to get the tests to pass. Is it a know issue that signals don't trigger slot on travis?

src/core/qgsvectorlayer.h
roya0045: @m-kuhn @nyalldawson simple ping to get feedback on this aspect of the code.

roya0045:
@m-kuhn @nyalldawson Any chances to get this ready before the end of the freeze, other than the previously mentioned point, there should be nothing nothing extravagant in this PR and everything should comply with the points mentioned in the first part.

src/core/qgsvectorlayer.h
roya0045: @m-kuhn @nyalldawson once again a new ping, I know that this change will raise a red flag and I want to know if the ability to test or hiding this slot is more important.

m-kuhn:
FWIW, I would have preferred another - more rule-based than a feature id based - approach. I'm fine with this to continue, but I'm out of the review team in this case :)

roya0045:
> FWIW, I would have preferred another - more rule-based than a feature id based - approach. I'm fine with this to continue, but I'm out of the review team in this case :)

In this case could you elaborate what you mean with rule-based? I'm not quite sure which part should be change if I were to migrate from feature id to rule-based?

A simple example would be helpful to get me thinking.

m-kuhn:
Using the expression rules of the symbol to also count / filter / check features.

roya0045:
> Using the expression rules of the symbol to also count / filter / check features.

I'm possibly not understanding correctly what you have in mind. But what is currently implemented only asks for the symbol id. With that under the hood the feature Ids are fetched and the aggregate expression are evaluated on that subset.`

Isn't this the same than using the expression rules of the symbol, to filter out features, get their id and then perform the aggregation but in a more efficient way since the ids are already cached?

m-kuhn:
> Isn't this the same than using the expression rules of the symbol, to filter out features, get their id and then perform the aggregation but in a more efficient way since the ids are already cached?

... and at the expense of a higher memory footprint for a very specific use case and [an additional cache which needs to be invalidated](https://martinfowler.com/bliki/TwoHardThings.html) and an API that is specific for feature ids (it allows for stacking expressions specifically with fids and "AND", but not not with "OR" or with a rectangle or ...). But I really don't want to reiterate this at that point, the decision has been taken and the API is in there, so it's out of my responsibility :)

roya0045:
@nyalldawson what is your opinion on what Matthias proposed?

 I don't mind changing the code (the filtering PR will not be affected anyway). I always thought that caching would be better, especially when doing atlas and reports. If there is an agreement that the memory footprint outweigh the speed gain I will make the necessary changes.

roya0045:
@m-kuhn other than the conversion to sld, is there any other straightforward way to convert classified and categorized symbol to their rule-base equivalent? I don't think splitting the sld function is optimal and without that conversion step it make things harder to convert a symbol to its expression. This could be used as an expression or to implement thing with an expression rather than storing lists of long long (a smaller data format would have been nicer to take up less memory).

If I have a way to easily convert symbols to expression and if @nyalldawson agree that it would be better to not store fids I'll make the necessary changes, this should be relatively easy to do.

m-kuhn:
@roya0045 
I would recommend to ask one question at a time and discuss that first before going into implementation details. @nyalldawson has accepted your PR for fids, I assume he was in favor of it or he wouldn't have accepted. It's certainly ok to ask for re-affirmation after my statement, but not yet the time to talk about sld.

Right now he's probably busy - like we all - ironing out issues with the upcoming 3.10 release, before spending his precious volunteer time on reviewing new fancy features.

roya0045:
not stale

roya0045:
@m-kuhn I'm trying to see how to implement the expression method (mostly as a fallback method for big layers), to implement in this PR or a follow up PR.  And I'm having issue with the creation of expression. I was thinking of using `QgsRuleBasedRenderer::convertFromRenderer ` but this would break the symbology of a layer and there is the issue of the `ELSE` statements. 

This means that I probably need a function similar to `QgsRuleBasedRenderer::convertFromRenderer` and I would need to store the expressions for every symbol in the layer and to append this to the expression instead of the fids 'expression'/filter.

Another point is to have a method to get the size of the layer, if it cannot be done without counting this mean that in order to have it as a fallback method I would need to have a threshold during the count where the fids are invalidated, skipped and to have the converted run afterward.

Your feedback would be appreciated on this approach.

roya0045:
@nyalldawson should I add a parameter to the featurecounter and the QgsVectorLayer itself or the count method in QgsVectorLayer to only store the fids in some conditions?

If the user only wants the count this would prevent storing the fids and if the user wants to use the feature, counting would be carried out again  with the new params. This would reduce memory usage.

roya0045:
@nyalldawson I'm not sure why but I find that there is a ton of call to the data function of the qgslegendmodel and they are not coming from refresh functions. I was simply wondering where those call might come from and how to reduce them to prevent re-evaluating everything 12+ times.

I have checked and it doesn't come from my force refresh or from QgsLayoutItemLegend::updateFilterByMap.

roya0045:
Something seems to always refresh the legend by calling QgsLegendModel::data, I'm not sure what it is but I get 63 call to evaluate the legend expression. But my refresh function is only called once, when counting is done.

roya0045:
@nyalldawson @m-kuhn concerning the changes made to the counting, as fids caching was a distinct PR, I assume that I should also make another PR to make fids caching optional instead of having it in this PR?

roya0045:
@nyalldawson Due to the multiple request to data and hence the evaluation of the expression, I was wondering if I should find a way to cache the expression by keeping the same expressioncontext. This mean that I would need to find a way to invalidate the context somehow, especially if it uses expression that are based on things that can change such as the layout extent or an atlas attribute. What are your thoughts on this?

roya0045:
Just had an idea:
To prevent using bad values I think I should evaluate the sub expression, this way if the subexpression variabrles (any filters) are to change, the cached value should change too. This would mean that if the subexpression is actually static during eval, the result should stay the same and this should not affect the current API ad possibly improve it.

roya0045:
@nyalldawson I've been trying to find a way to retreive the context cache to prevent re-evaluating expression but somehow my pointer to my context gets removed during expression parsing and it segfaults.

roya0045:
Removed any caching attempt from this branch, too unstable and does not work as intended with sub queries.

roya0045:
@nyalldawson any hopes for getting this reviewed or was there something else that you wanted me to do?

roya0045:
@nyalldawson Another ping in the hopes to get a first review.

I would hoping that stalebot would ping this but nothing occurred after 15 days due to the merge conflict.

For information, I had to put the symbol parameter before some others due to the fact that concatenator and order by are not part of the 'base' arguments and this would have made finding the symbol parameter more annoying. By moving it to the base set this makes retrieval more reliable. But it might affect people that have code that relies on the current order without binding. I am open to discussion.

roya0045:
@nirvn  since you have made changes to expression functions in the past, would you be willing to glance at this PR? Thanks.

roya0045:
Not stale, awaiting review from @nyalldawson  and whoever else would like to review.

roya0045:
Not stale, no backport or labeled PR should be needed.

PR #30729, opened 2019-07-15
Show layer rendering/data access errors in the layer tree 

nyalldawson:
This makes these errors more visible to users, instead of hiding them away in the log panel

E.g. the attached screencast shows the error indicator displaying when trying to render a broken jpg image:

Peek 2019-07-16 09-22


nyalldawson:
Code wise, we piggy-back on the existing bad layer indicator here. If a layer has a bad source, this state is shown in the indicator. If not, the indicator falls back to showing any data source errors (with a less prominent icon)

NathanW2:
Very nice.  Much nicer than just dumping into the log view and ignoring it for the user.

nyalldawson:
@elpaso are you happy with these changes?

PR #30734, opened 2019-07-16
[FEATURE] HANA database provider

mrylov:
This pull request contains a provider implementation for SAP HANA database.
The provider uses a third-party [library](https://github.com/SAP/odbc-cpp-wrapper) to make ODBC calls for accessing a database.


NathanW2:
Sorry, we might need a bit more context on what this is and why it needs to be a core provider?

nirvn:
Greetings @mrylov . HANA database, that's new to me :)

Generally speaking, the QGIS project is not too hot to the idea of adding a new provider into its tree due to added burden of long term maintenance, especially when a provider would not be of use to a vast majority of its users.

My advice: serve your provider as a QGIS plugin, that's the easiest way forward on the basis of the code you already produced

You could also go down the path of adding support for this database into the gdal library itself. Doing so would reach a broader community of open source users, including QGIS.

mrylov:
@NathanW2 , are there any additional requirements for core providers?

@nirvn, SAP HANA is In Memory database which has a spatial extension that is compliant with version 1.1 of the OGC standard.
> Generally speaking, the QGIS project is not too hot to the idea of adding a new provider into its tree due to added burden of long term maintenance, especially when a provider would not be of use to a vast majority of its users.
How you estimate the popularity of a database?

> You could also go down the path of adding support for this database into the gdal library itself. Doing so would reach a broader community of open source users, including QGIS.
I agree. We are planning to add a HANA driver to the GDAL library as well. Could you shed some light on how a GDAL plugin can act as a proper QGIS data provider?

> My advice: serve your provider as a QGIS plugin, that's the easiest way forward on the basis of the code you already produced
Is it possible to have a standalone shared/dynamic library for a provider without additional changes in qgisapp.cpp file?

haubourg:
@mrylov thanks for the work! 
 I'd like to point out the process of QGIS Enhancement Proposal that is the process to discuss major features, before going deeply in implementation like in Pull requests . Just create an issue in https://github.com/qgis/QGIS-Enhancement-Proposals/issues

vpicavet:
Hi,

Thanks for the proposal. Support for SAP HANA would be great for QGIS to tackle GIS issues in new markets.

I share @nirvn concern : any implementation for new providers should first be done in GDAL/OGR if possible, giving access to this format to a lot of FOSS4G softwares at once, QGIS included. 

Adding direct dependencies to QGIS is not something we love to do neither.

Another concern is the licence : QGIS is GPLv2+, your PR adds a linked library as a dependency, wich is itself licensed as Apache 2. The latter is not compatible with GPLv2, which means than linking with odbc-cpp-wrapper force to distribute QGIS as GPLv3. And doing that would prevent to link and distribute with any other library compatible with GPLv2 and not with GPLv3, which is (really) annoying. I did not check if this is currently the case, but we should fix that anyway. 
I filed an issue in odbc-cpp-wrapper with a solution.

nyalldawson:
@mrylov 

This is certainly an impressive piece of work, but like others I also share concern about accepting new data providers into master. Historically when we've done this is becomes a huge maintainance burden on the project, and ones like this are larger because availability of data sources to test with is limited (e.g. due to licensing costs). The existing db2 provider is a good example of one which (in hindsight) should never have been allowed into master, and which always should have been a separate plugin.

> I agree. We are planning to add a HANA driver to the GDAL library as well. Could you shed some light on how a GDAL plugin can act as a proper QGIS data provider?

It should all be transparent -- if OGR can read a data source, it will work in QGIS

> Is it possible to have a standalone shared/dynamic library for a provider without additional changes in qgisapp.cpp file?

It should be - especially given recent refactoring done by @wonder-sk. (If not, PRs adding any missing required hooks would be very welcome!)

stefanuhrig:
I have created a QGIS Enhancement Proposal to follow-up on this: qgis/QGIS-Enhancement-Proposals#151

PR #30735, opened 2019-07-16
Some ports from Kadas

m-kuhn:
**Code cleanups**

- Hide away *default translator* implementation
- Rubberband loops (not from Kadas)
- Const correctness
- ~Add inline keyword to declarations to match definitions and avoid "redeclared without dllimport" warnings~ Removed again at least for now, see https://isocpp.org/wiki/faq/inline-functions#where-to-put-inline-keyword

**Feature**

- Add the possibility to use an SVG as rubberband marker
- Add [mil](https://en.wikipedia.org/wiki/Milliradian) units

### DIFF
--- a/src/core/qgsunittypes.cpp
+++ b/src/core/qgsunittypes.cpp
@@ -1256,6 +1256,8 @@ double QgsUnitTypes::fromUnitToUnitFactor( QgsUnitTypes::AngleUnit fromUnit, Qgs
           return 3600;
         case AngleTurn:
           return 1.0 / 360.0;
+        case AngleMil:
+          return M_PI / 180.0 * 3200 / M_PI;
         case AngleUnknownUnit:
           break;
       }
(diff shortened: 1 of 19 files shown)


src/core/qgsunittypes.cpp
```diff
@@ -1256,6 +1256,8 @@ double QgsUnitTypes::fromUnitToUnitFactor( QgsUnitTypes::AngleUnit fromUnit, Qgs
           return 3600;
         case AngleTurn:
           return 1.0 / 360.0;
+        case AngleMil:
```
nyalldawson: These ones will need to be added to existing angle unit tests

nyalldawson (commented):
Nice work!

src/core/qgsunittypes.cpp
m-kuhn: Done, some expected failures in their current form

src/gui/qgsrubberband.cpp
```diff
@@ -21,6 +21,7 @@
 #include "qgsproject.h"
 #include "qgsrectangle.h"
 #include <QPainter>
+#include <QSvgRenderer>
```
m-kuhn:
```suggestion
```

src/core/qgsunittypes.cpp
m-kuhn: Done, any more tests to do?

PR #30748, opened 2019-07-16
Fix addEntry for reference relation widget 

troopa81:
## Description

When adding an entry from the reference relation widget, the widget was using the referencing field index and not the referenced one, so it fails to update correctly the combo box after the form pop up. 

## Checklist

### DIFF
--- a/src/gui/editorwidgets/qgsrelationreferencewidget.cpp
+++ b/src/gui/editorwidgets/qgsrelationreferencewidget.cpp
@@ -208,7 +208,6 @@ void QgsRelationReferenceWidget::setRelation( const QgsRelation &relation, bool 
       mComboBox->setIdentifierField( mReferencedField );
 
     mReferencedFieldIdx = mReferencedLayer->fields().lookupField( relation.fieldPairs().at( 0 ).second );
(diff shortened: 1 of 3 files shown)


nyalldawson:
Does this need backporting?

troopa81:
Yes, I just add the labels

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
```diff
@@ -208,7 +208,6 @@ void QgsRelationReferenceWidget::setRelation( const QgsRelation &relation, bool
       mComboBox->setIdentifierField( mReferencedField );
 
     mReferencedFieldIdx = mReferencedLayer->fields().lookupField( relation.fieldPairs().at( 0 ).second );
-    mReferencingFieldIdx = mReferencingLayer->fields().lookupField( relation.fieldPairs().at( 0 ).first );
```
m-kuhn: This should still be initialized I guess?

troopa81: actually, the variable has no use anymore. I just remove it 4ee2a53

rldhont (approved):

PR #30750, opened 2019-07-16
[Tests] Creating tests about usedAttributes method

rldhont:
## Description
QgsFeatureRenderer usedAttributes method Returns a list of attributes required by this renderer. Attributes not listed in here may not have been requested from the provider at rendering time.

QgsSymbol usedAttributes method returns a list of attributes required to render this feature. This should include any attributes required by the symbology including the ones required by expressions. 

QgsSymbolLayer usedAttributes method returns the set of attributes referenced by the symbol layer. This includes attributes required by any data defined properties associated with the symbol layer (QgsPropertyCollection referencedFields).

QgsPropertyCollection referencedFields method returns the set of any fields referenced by the active properties from the collection

This methods are useful to speed up the rendering by requesting only needed attributes, but if one fail and do not return all the needed attributes, the rendering could be broken. Tests are needed.

The already existing tests are available for:
* QgsProperty
* QgsPropertyCollection
* QgsCategorizedSymbolRenderer
* QgsGraduatedSymbolRenderer

New tests has to be added to avoid regression in rendering.

## Checklist

### DIFF
--- a/tests/src/python/test_qgsrulebasedrenderer.py
+++ b/tests/src/python/test_qgsrulebasedrenderer.py
@@ -15,6 +15,9 @@
 *   (at your option) any later version.                                   *
 *                                                                         *
 ***************************************************************************
+
+From build dir, run: ctest -R PyQgsRulebasedRenderer -V
+
 """
 
 __author__ = 'Matthias Kuhn'
@@ -369,6 +372,24 @@ class TestQgsRulebasedRenderer(unittest.TestCase):
         self.assertEqual(len(rootrule.rulesForFeature(ft, ctx)), 1)
         self.assertEqual(rootrule.rulesForFeature(ft, ctx)[0], self.rx3)
         renderer.stopRender(ctx)
+
+    def testUsedAttributes(self):
+        ctx = QgsRenderContext.fromMapSettings(self.mapsettings)
+
+        # Create rulebased style
+        sym2 = QgsFillSymbol.createSimple({'color': '#71bd6c', 'outline_color': 'black'})
+        sym3 = QgsFillSymbol.createSimple({'color': '#1f78b4', 'outline_color': 'black'})
+
+        self.rx2 = QgsRuleBasedRenderer.Rule(sym2, 0, 0, '"id" = 200')
+        self.rx3 = QgsRuleBasedRenderer.Rule(sym3, 1000, 100000000, 'ELSE')
+
+        rootrule = QgsRuleBasedRenderer.Rule(None)
+        rootrule.appendChild(self.rx2)
+        rootrule.appendChild(self.rx3)
+
+        renderer = QgsRuleBasedRenderer(rootrule)
+
+        self.assertCountEqual(renderer.usedAttributes(ctx), {'id'})
 
 
 if __name__ == '__main__':
--- a/tests/src/python/test_qgsrulebasedrenderer.py
+++ b/tests/src/python/test_qgsrulebasedrenderer.py
@@ -446,14 +446,10 @@ class TestQgsRulebasedRenderer(unittest.TestCase):
 
         # for symbol layer
         self.assertCountEqual(l1.usedAttributes(ctx), {'Heading'})
-        self.assertIn('Heading', l1.usedAttributes(ctx))
         # for symbol
         self.assertCountEqual(sym1.usedAttributes(ctx), {'Heading'})
-        self.assertIn('Heading', sym1.usedAttributes(ctx))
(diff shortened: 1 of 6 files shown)


tests/src/python/test_qgssymbollayer.py
```diff
@@ -579,6 +635,12 @@ def testQgsSimpleFillSymbolLayer(self):
         mMessage = 'Expected "%s" got "%s"' % (mExpectedValue, mValue)
         assert mExpectedValue == mValue, mMessage
 
+        ctx = QgsRenderContext()
+        mExpectedValue = 0
+        mValue = len(mSymbolLayer.usedAttributes(ctx))
+        mMessage = 'Expected no attributes used'
+        assert mExpectedValue == mValue, mMessage
```
m-kuhn: is this not possible?
It normally gives better messages for debugging and is more modern.

```suggestion
        self.assertEqual(mExpectedValue, mValue)
```

m-kuhn:
:tada: more tests are always very welcome!

tests/src/python/test_qgssymbollayer.py
```diff
@@ -527,6 +566,23 @@ def testRenderMarkerLayerDataDefined(self):
...
+        ctx.setExpressionContext(exp_ctx)
+        # for symbol layer
+        self.assertEqual(len(layer.usedAttributes(ctx)), 1)
+        self.assertTrue('Class' in layer.usedAttributes(ctx))
+        # for symbol
+        self.assertEqual(len(symbol.usedAttributes(ctx)), 1)
+        self.assertTrue('Class' in symbol.usedAttributes(ctx))
+        # for symbol renderer
+        self.assertEqual(len(points_layer.renderer().usedAttributes(ctx)), 1)
+        self.assertTrue('Class' in points_layer.renderer().usedAttributes(ctx))
```
nyalldawson: On the same note as @m-kuhn, there's also self.assertIn and self.assertCountEqual (Google it, it doesn't do what you'd expect!) which would be nice here.

tests/src/python/test_qgssymbollayer.py
rldhont: This code is conformed to all the tests in this file. Do you think we have to update all the tests ?

m-kuhn: That would be nice but I didn't dare to ask that from you.

nyalldawson: Ideally, but at least we can avoid introducing more of the old style test approach

tests/src/python/test_qgssymbollayer.py
rldhont: Thanks to pointing these, I did not know them

tests/src/python/test_qgssymbollayer.py
rldhont: Ok, thanks for the review

tests/src/python/test_qgsrulebasedrenderer.py
```diff
@@ -370,6 +377,87 @@ def testWillRenderFeatureTwoElse(self):
...
+        self.rx3 = QgsRuleBasedRenderer.Rule(sym3, 1000, 100000000, 'ELSE')
+
+        rootrule = QgsRuleBasedRenderer.Rule(None)
+        rootrule.appendChild(self.rx2)
+        rootrule.appendChild(self.rx3)
+
+        renderer = QgsRuleBasedRenderer(rootrule)
+
+        self.assertCountEqual(renderer.usedAttributes(ctx), {'id'})
+        self.assertIn('id', renderer.usedAttributes(ctx))
```
nyalldawson: This test is redundant -- it's covered by the assertCountEqual check (which despite the misleading name, actually checks if the contents of a list are equal while ignoring the order!).

rldhont: I did not know it. I was thinking that asserCountEqual is only about count and that it is better to have the same list to explain what the test are waiting for.
Thanks to pointing it. I will update the tests.

nyalldawson: I thought the same initially too. It's a really badly chosen name.

rldhont: Done, thanks

rldhont: Done, thanks

nyalldawson:
Looks good to me!

nyalldawson (approved):

PR #30751, opened 2019-07-16
Followup mil angle addition

nyalldawson:
This PR:
- adds missing unit tests for mil angle conversion
- exposes mil as an option for the angle measure tool

and, most importantly, corrects the definition of mil to the SI definition (1/1000 radian), instead of PI/3200. From https://en.wikipedia.org/wiki/Milliradian :

"The true definition of a milliradian is based on a unit circle with a radius of one and an arc divided into 1000 mils per radian, hence 2000π or approximately 6283.185 milliradians in one turn, and rifle scope adjustments and reticles are calibrated to this definition.[1] There are also other definitions used for land mapping and artillery which are rounded to more easily be divided into smaller parts for use with compasses. For instance there are artillery sights and compasses with 6400 NATO mils, 6000 Warsaw Pact mils or 6300 Swedish "streck" per turn instead of 360° or 2000π, achieving higher resolution than a 360° compass while also being easier to divide into parts than if true milliradians were used."

I think we should follow SI conventions here, not region/military-specific tweaks.


nyalldawson:
ping @m-kuhn

rouault:
For me, 'mil' evokes more the NATO mil (6400 mil = 2 PI) than the milliradian unit, but I'm biased since in my "previous life" I was tainted by defense stuff (even a few years ago people in the military still reasoned in terms of NATO mil for artillery). Why not leaving mil to be NATO mil (since that was likely the original intent of https://github.com/qgis/QGIS/commit/bc4bd9199ef401fdd04d4831bbcaaaec9e20065f), and have a dedicated mrad unit to avoid the name ambiguity ?
At the very least, the comment of the enumeration value at https://github.com/qgis/QGIS/commit/bc4bd9199ef401fdd04d4831bbcaaaec9e20065f#diff-b83ac1cddc90c80c83bbda928b9fa052R106 and the text of https://github.com/qgis/QGIS/commit/bc4bd9199ef401fdd04d4831bbcaaaec9e20065f#diff-bd07ee2342745306a92d2812b068faedR1436 should be clarified

m-kuhn:
Same as @rouault here (except for the "previous life" statement), the main reason for me is that many compasses here come with these units (but we use to talk about artillery mil instead of NATO mil which I have never heard before like that).

nyalldawson:
Fair points - guess we need two seperate options here ("milliradians (SI)" vs "mil (NATO)")

nyalldawson:
Ok, done - both SI milliradians and NATO mil are options now.

m-kuhn (approved):
Perfect, good job

rouault:
I just came accross https://github.com/OSGeo/PROJ/blob/master/data/sql/unit_of_measure.sql#L75 , where EPSG has a "mil_6400" unit, which reminded me this discussion

PR #30756, opened 2019-07-16
Improve UX for labeling map tools

nyalldawson:
When label map tools are used to select a label and the user clicks on overlapping labels, prioritise either the labels in the current layer OR fallback to picking the smallest candidate label (since it will be the most difficult to select)


nirvn:
Thanks for that wait cursor, there's nothing worse than a hanged UI missing a wait-a-minute visual queue.

nirvn (approved):

PR #30770, opened 2019-07-17
only install/copy srs.db that we're using (fix 56d03f783f041)

jef-n:


nyalldawson:
@jef-n won't we still have an issue if users mix pre/post proj 6 builds on the one machine? E.g running 2.18 and 3.8?

jef-n:
> @jef-n won't we still have an issue if users mix pre/post proj 6 builds on the one machine? E.g running 2.18 and 3.8?

The srs.db should be in separate directories.  If you're running from the build directory then yes - but that should already addressed in c715bc14d9

nyalldawson:
@jef-n ok, fair point. The original bug report was found on gentoo, so I'm not sure how that package works (or if it was also a custom build). But I'm happy to write any downstream packaging issues off as not our responsibility anyway :wink:

PR #30771, opened 2019-07-17
Allow more contextual info to be set for font buttons, use correctly in layout widgets

nyalldawson:
Gives more font/symbol buttons access to the message bar and other useful contextual information.

### DIFF
--- a/src/gui/qgslabelinggui.cpp
+++ b/src/gui/qgslabelinggui.cpp
@@ -135,8 +135,7 @@ void QgsLabelingGui::updateCalloutWidget( QgsCallout *callout )
 }
 
 QgsLabelingGui::QgsLabelingGui( QgsVectorLayer *layer, QgsMapCanvas *mapCanvas, const QgsPalLayerSettings &layerSettings, QWidget *parent, QgsWkbTypes::GeometryType geomType )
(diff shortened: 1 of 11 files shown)


src/gui/qgslabelinggui.cpp
```diff
@@ -135,8 +135,7 @@ void QgsLabelingGui::updateCalloutWidget( QgsCallout *callout )
 }
 
 QgsLabelingGui::QgsLabelingGui( QgsVectorLayer *layer, QgsMapCanvas *mapCanvas, const QgsPalLayerSettings &layerSettings, QWidget *parent, QgsWkbTypes::GeometryType geomType )
-  : QgsTextFormatWidget( mapCanvas, parent, QgsTextFormatWidget::Labeling )
-  , mLayer( layer )
```
nirvn: Don't you need to keep this line?

nirvn (approved):

src/gui/qgslabelinggui.cpp
nyalldawson: Ah - thanks, I missed off removing the mLayer member from QgsLabelingGui. It's been moved to the QgsTextFormatWidget base class.

nyalldawson: Fixed

nirvn:
Looks good to merge whenever travis turns green.

PR #30772, opened 2019-07-17
Read-only style manager dialog UX improvement

nyalldawson:
When a style manager dialog is opened in read only mode, still allow symbols to be double clicked to view their properties. Just disable actually saving any edited symbols instead.


nyalldawson:
@nirvn as we've previously chatted about...

nirvn (approved):

nirvn:
@nyalldawson , looking good.

PR #30774, opened 2019-07-18
Show a warning when the move label tool cannot be used on a label

nyalldawson:
E.g. when a layer is set to store label x/y in a real field (not
an aux field), then that layer must be editable in order to move
labels in it. Otherwise the move operation fails, but users are
left no clue as to why the label can't be shifted...

(this one has confused/frustrated me for a long time, and I only just today worked out why i couldn't move some labels...!)

### DIFF
--- a/src/app/qgsmaptoolchangelabelproperties.cpp
+++ b/src/app/qgsmaptoolchangelabelproperties.cpp
@@ -21,6 +21,8 @@
 #include "qgsrubberband.h"
 #include "qgsvectorlayer.h"
 #include "qgsmapmouseevent.h"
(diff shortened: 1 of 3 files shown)


nyalldawson:
Better approach pushed - we now automatically enable editing on the layer (when we can), and notify users that we did this via the message bar

nirvn:
Excellent

nirvn:
You'll need the same adjustments for the rotate tool.

nyalldawson:
@nirvn done

nirvn (approved):

PR #30775, opened 2019-07-18
Allow data defined settings in text formats 

nyalldawson:
This PR adds the API and GUI changes to allow data defined properties to be set in QgsTextFormat, so that these are automatically respected whenever the format is rendered using QgsTextRenderer.

This is done transparently to the caller - so by porting away from the raw QPainter text apis to use QgsTextRenderer instead, then they immediately gain the ability to use data defined properties in their text formats.

Included in this PR is also a new expression variable for use when rendering scale bar text - @scale_value. By using this variable within data defined text formats for the scale bar text, users can customise the appearance of individual labels inside a scalebar. E.g.:

image

(this is an added bonus -- the important work here is the transparent handling of data defined settings EVERYWHERE text renderer is used)


src/gui/qgslabelinggui.h
```diff
@@ -52,28 +52,12 @@ class GUI_EXPORT QgsLabelingGui : public QgsTextFormatWidget
...
 
-    /**
-     * Deactivate a field from data defined properties and update the
-     * corresponding button.
-     *
-     * \param key The property key to deactivate
-     *
-     * \since QGIS 3.0
-     */
-    void deactivateField( QgsPalLayerSettings::Property key );
```
nirvn: Is this not breaking the API?

nirvn (approved):
This is quality work, nothing to say about it after review. Big boost which will benefit many corners of QGIS.

src/gui/qgslabelinggui.h
nirvn: Ignore this, I saw it was moved to the parent class. Somehow github won't let me delete this.

nyalldawson: Also - that class is marked as non stable anyway

nyalldawson:
Thanks @nirvn!

PR #30776, opened 2019-07-18
Fix lengthy delay when clicking change label tool on a label for the first time

nyalldawson:
Previously this was firing off the styleChanged for every new auxiliary
field created, which is a very expensive call to process (as it involves
rebuilding GUI widgets). Instead, delay the call and only it fire it
once, if we actually changed something in the style.


nirvn:
Brilliant

nyalldawson:
(Thanks to hotspot)

PR #30801, opened 2019-07-19
Disable CRS and transform cache permenantly when exiting Qgis

nyalldawson:
Hopefully this will prevent additional items being added to the cache after we've gracefully finalised proj operations, which results in the infamous crash-on-exit fiasco...


nirvn:
Oh yeah! Let's get make sure this doesn't miss the 3.4.10 boat.

nirvn:
@nyalldawson , I've made a build including the commits in this PR, and I can report nothing is broken AFAIK. I'm on linux and never saw the exit crash we're trying to plug here, but the code looks sound and the extra precaution cannot hurt here.

PR #30802, opened 2019-07-19
When saving a symbol to the style database, prepopulate the tags widget with existing tags

nyalldawson:


nirvn (approved):
Looks good, it's a nice gap fill until we eventually get a QgsTagEdit widget or something :)

PR #30811, opened 2019-07-19
Fix missing cache skip in proj 6 code path

nyalldawson:


src/core/qgscoordinatetransform.cpp
```diff
@@ -821,7 +821,10 @@ bool QgsCoordinateTransform::setFromCache( const QgsCoordinateReferenceSystem &s
   if ( sourceKey.isEmpty() || destKey.isEmpty() )
     return false;
 
-  sCacheLock.lockForRead();
+  QgsReadWriteLocker locker( sCacheLock, QgsReadWriteLocker::Read );
```
m-kuhn: Could this be moved to after the condition to avoid the locking/unlocking in case caching is disabled?

nyalldawson: I'm trying to be super-safe here and also protect the disable cache static boolean value.

m-kuhn: Hmm... ok.
I wouldn't expect issues (if there's a concurrent write at the same time changing the value the result will be a matter of timing with or without locking). But yes it still feels slightly cleaner.

nyalldawson: I'm still inclined to be overly safe here, especially given that the cache is only ever disabled on shutting down qgis, and we would ideally expect this situation to never arise (accessing cache during the shutdown process).

m-kuhn: All good :)

nyalldawson: Let's just hope this finally puts this issue to bed...

m-kuhn: fingers crossed!

nirvn: Going to a nearby pagoda to light up so incense! ;)

PR #30816, opened 2019-07-19
Update the project ellipsoid if CRS is taken from first layer

### LINKED ISSUE #30815: Ellipsoid not set when setting CRS from first layer
- When configuring QGIS to set the project CRS based on the CRS from the first layer, the ellipsoid is not taken into account
- When QGIS is configured to take a default CRS for new projects, the ellipsoid is set
- There is no option to configure a "default ellipsoid"

This is not consistent. QGIS should take the ellipsoid from the CRS of the first layer.
And if that's not enough for someone, we should add a configuration option for the default ellipsoid (and another option to "follow the project CRS ellipsoid), but that would be a separate ticket / feature request if it's of interest to someone.

m-kuhn:
Fix #30815

### DIFF
--- a/python/core/auto_generated/qgsproject.sip.in
+++ b/python/core/auto_generated/qgsproject.sip.in
@@ -164,9 +164,11 @@ Returns the project's native coordinate reference system.
 .. versionadded:: 3.0
 %End
 
-    void setCrs( const QgsCoordinateReferenceSystem &crs );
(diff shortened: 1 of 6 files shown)


nyalldawson (approved):

m-kuhn:
Backporting should be done, but I'm not sure if the API change is acceptable for this?

PR #30830, opened 2019-07-20
Fix change label map tool overwritting pre-existing non-field expressions

nirvn:
## Description
This PR fixes a regression-like that was introduced alongside the fantastic auxiliary field feature whereas preexisting non-field expressions would be overwritten with a newly-created auxiliary field. 

While this might be acceptable for the move label map tool, it's undesirable for the change label map tool, whereas it's not unlikely that non-field expressions are set (e.g., a multilingual project that sets a label font family via an expression looking for a variable that sets the currently displayed language, font color set on basis on a list of conditions, etc.) and should _not_ be overwritten by a new auxiliary field. 

With this PR applied, the change label tool plays nicely with preexisting non-field expression by simply disabling such properties. Any properties not field by such an expression will get an auxiliary field (or index reference to an existing field).

@nyalldawson , took me losing three long, long expressions while testing out callouts and multiline alignments to realize how important this is.

## Checklist

### DIFF
--- a/src/app/qgsmaptoolchangelabelproperties.cpp
+++ b/src/app/qgsmaptoolchangelabelproperties.cpp
@@ -73,7 +73,7 @@ void QgsMapToolChangeLabelProperties::canvasPressEvent( QgsMapMouseEvent *e )
   if ( !mCurrentLabel.layer->isEditable() )
   {
     QgsPalIndexes indexes;
-    bool newAuxiliaryLayer = createAuxiliaryFields( indexes );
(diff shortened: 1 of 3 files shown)


nyalldawson:
Oh good fix. I'd be cursed by this one to last week, but it slipped my mind. Thanks!

nyalldawson (approved):

nirvn:
@nyalldawson , I feel you brother, I feel you.

PR #30836, opened 2019-07-20
[FEATURE] Allow rendering raster layers outside the reported extent

### LINKED ISSUE #30251: QGIS clips features at the border of WMS layers
When features are at the border of a WMS layer, QGIS may show symbols or labels clipped.

WMS_Extent2

This seems to be a client issue https://github.com/qgis/QGIS/pull/30071#issuecomment-502706974
> I think this is a separate issue. The "issue" you are referring to is related to this [thread](https://lists.osgeo.org/pipermail/qgis-developer/2019-May/thread.html#57225) initiated by @andreasneumann  on the QGIS-Developer mailing list. And please see the [answer](https://lists.osgeo.org/pipermail/qgis-developer/2019-May/057237.html) I provided in the thread, which @elpaso [agreed](https://lists.osgeo.org/pipermail/qgis-developer/2019-May/057238.html) with :)

Some interesting input by @elpaso is linked there

> I agree with you that QGIS Server does the right thing here, I think that
the main question is:
> 
> 1. is the WMS GetCapabilities layer's BoundingBox meant to be the features
> BBOX or can it be larger than that?
> 2. if the latter is true, we need a way to tell QGIS Server that he needs
> to advertise a BoundingBox in GetCapabilities which is not the layer's BBOX
> stored in the QGIS project but it's a different (probably larger) one.
> 
> all the rest will follow, because the client will get a larger BBOX from
> GetCapabilities and it will request a larger image that has enough buffer
> for the symbols.
> 
> Note that I checked mapserver and it behaves by default exactly like QGIS
> Server does (I didn't check the single point but the symbols are cut-off at
> the layer's bbox in general), except that mapserver allows you to override
> the layer extent per-layer.
> 
> IMO the fix is in the client, either by allowing to override the layer
> extent advertised by the server and to store it in the project itself (this
> may require some work in the server side too in order to handle the
> override) or by setting an option in the WMS provider that will always
> request the canvas extent.

Something like the last option sounds good to me. But I wonder if this just should be done always. Or is there a good reason to request the layer bbox only?

m-kuhn:
The extent reported by raster layers may be smaller than the area which can be rendered.
Notably for WMS with symbology that takes more space than the data.

For reference / visual changelog: this is a screenshot of the final implementation

Screenshot from 2019-07-25 10-25-17

Fix #30251

### DIFF
--- a/python/core/auto_generated/raster/qgsrasterdataprovider.sip.in
+++ b/python/core/auto_generated/raster/qgsrasterdataprovider.sip.in
@@ -479,6 +479,14 @@ Resolutions are calculated in the provider's crs().
 .. versionadded:: 3.8.0
 %End
 
+    virtual bool ignoreExtents() const;
+%Docstring
+Returns true if the extents reported by the data provider are not reliable
+and it's possible that there is renderable content outside of these extents.
+
+.. versionadded:: 3.10.0
+%End
+
   signals:
 
     void statusChanged( const QString & ) const;
--- a/python/core/auto_generated/raster/qgsrasterlayer.sip.in
+++ b/python/core/auto_generated/raster/qgsrasterlayer.sip.in
@@ -382,7 +382,7 @@ Writes the symbology of the layer into the document provided in SLD 1.0.0 format
 .. versionadded:: 3.6
 %End
 
-    bool ignoreExtent() const;
+    bool ignoreExtents() const;
 %Docstring
 If the ignoreExtent flag is set, the layer will also render outside the
 bounding box reported by the data provider.
@@ -391,17 +391,6 @@ to be drawn outside the data extent.
 
 .. versionadded:: 3.10
 %End
-
-    void setIgnoreExtent( bool ignoreExtent );
-%Docstring
-If the ignoreExtent flag is set, the layer will also render outside the
-bounding box reported by the data provider.
-To be used for example for WMS layers with labels or symbology that happens
-to be drawn outside the data extent.
-
-.. versionadded:: 3.10
-%End
-
 
   public slots:
     void showStatusMessage( const QString &message );
--- a/python/core/auto_generated/raster/qgsrasterlayer.sip.in
+++ b/python/core/auto_generated/raster/qgsrasterlayer.sip.in
@@ -382,6 +382,26 @@ Writes the symbology of the layer into the document provided in SLD 1.0.0 format
 .. versionadded:: 3.6
 %End
 
+    bool ignoreExtent() const;
+%Docstring
+If the ignoreExtent flag is set, the layer will also render outside the
+bounding box reported by the data provider.
+To be used for example for WMS layers with labels or symbology that happens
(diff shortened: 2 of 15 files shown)


nyalldawson:
Would this be better suited to a provider setting, if it mostly affects wms layers?

m-kuhn:
The code was situated in qgsrasterlayer, I think it makes sense to leave it there (compared to copying it to all raster providers except WMS).
GUI wise it could be hidden for other providers, making it appear cleaner and reducing risks of unwanted instabilities on other providers (no idea what they do if activated). But this would also eliminate any potential use for other providers, but I'm not sure if there's any.
I'm 50/50 on that.

m-kuhn:
@elpaso , @elemoine what's your opinion on this?

elpaso:
I think it's fine to leave it the raster layer code, as long as we can only show the setting when relevant (is it WMS only?).

elpaso:
I think it's fine to leave it the raster layer code, as long as we can only show the setting when relevant (is it WMS only?).

elpaso:
I think it's fine to leave it the raster layer code, as long as we can only show the setting when relevant (is it WMS only?).

elpaso:
I think it's fine to leave it the raster layer code, as long as we can only show the setting when relevant (is it WMS only?).

m-kuhn:
> is it WMS only

That's the question. WMS has been reported. For others it might potentially be interesting. But we can remove it and see if someone complains.

nyalldawson:
I think it should be moved to the WMS connection properties gui. To me it sits along the workarounds sitting in that dialog, which are there to workaround issues for particular servers only:

image

E.g. add it here, and rename it as "Ignore reported layer extents"

m-kuhn:
That would mean they are global for all layers from a particular service (and saved in the users settings and not the project file)?

nyalldawson:
Yes - but isn't that what you'd want?

m-kuhn:
To have it on service level, maybe. To have it stored in the settings sounds bad to me, this means if you share a project with your colleague he'll have to setup the WMS service at his computer too. Having it inside .qgs/.qgz/.qlr would IMO be preferable.

nyalldawson:
That's still possible - it just becomes a property on the data source uri

m-kuhn:
Done!

nyalldawson:
Ok, looks good to me!

bgisb:
Could you please advise where I can find this setting? 

I use QGIS Server version 3.6.1. Noosa and QGIS Desktop version 3.8.3. Zanzibar. 

What does it mean "property on the data source uri"?
croppedwms

m-kuhn:
The property is in client when you connect to the WMS server.

bgisb:
Thank you for your reply and sorry for bothering you. 

It turns out, that I was looking for this functionality in QGIS 3.8.3. However, it will be available soon in QGIS 3.10
qgis383ignorereportedlayerextents

PR #30841, opened 2019-07-21
Rework "show pinned labels" icon

nyalldawson:
The original icon had an unintended effect of always looking enabled
on some platforms (due to the solid blue background rectangle)

Previous:

image

New:

image


nyalldawson:
@nirvn ?

nirvn:
Looks good. While we're trying to improve things here, could you add a separator between the 2nd and 3rd icon as well as the 4th and the 5th? It'll  regroup map tools vs toggles vs main styling dialogs together.

nirvn:
@nyalldawson , BTW, the show unplaced label icon could be further simplified and only have the <abc| drawing in red, it's sit quite nicely next to the blue one.

nyalldawson:
@nirvn is this what you had in mind?

image

nirvn:
@nyalldawson , exactly that, much improved.

PR #30842, opened 2019-07-22
Allow setting a custom path pre-processor for QgsPathResolver

nyalldawson:
QgsPathResolver::setPathPreprocessor allows setting a custom path pre-processor function, which allows for manipulation of paths and data sources prior to resolving them to file references or layer sources.

The processor function must accept a single string argument (representing the original file path or data source), and return a processed version of this path.

The path pre-processor function is called before any bad layer handler.

Example - replace an outdated folder path with a new one:

    def my_processor(path):
      return path.replace('c:/Users/ClintBarton/Documents/Projects', 'x:/Projects/')

    QgsPathResolver.setPathPreprocessor(my_processor)

Example - replace a stored database host with a new one:

    def my_processor(path):
      return path.replace('host=10.1.1.115', 'host=10.1.1.116')

    QgsPathResolver.setPathPreprocessor(my_processor)

Example - replace stored database credentials with new ones:

    def my_processor(path):
      path= path.replace("user='gis_team'", "user='team_awesome'")
      path = path.replace("password='cats'", "password='g7as!m*'")
      return path

    QgsPathResolver.setPathPreprocessor(my_processor)


elpaso:
:+1:  nice!

roya0045:
I'm not sure if this can be related to my changes to the bad layer handler, but while we are at it, should we rely more on qt for filepath handling. At least everything will be converted to unix style and then adapted back to the host system while resolving.

nyalldawson:
@roya0045 this shouldn't affect your work -- it's done as a step prior to the bad layer handler dialog, and only when a custom in-house plugin implements a preprocessor.

> should we rely more on qt for filepath handling

Yes, I also think this is good practice as opposed to using pythons os path manipulation code, but following that will be entirely in the hands of organisations who implement this function. (I've used a windows path example in the documentation here to hint to Windows users that they need to use / instead of \)

raymondnijssen:
Thanks, this is very useful!

arongergely:
That was fast, thanks @nyalldawson. I saw you tagged it as a new feature, does that mean it would first appear in the next regular release ( QGIS 3.10 )?

rldhont:
@dmarteau can you take a look ?

nyalldawson:
@arongergely 

>  I saw you labelled it as a new feature, does that mean it would first appear in the next regular release ( QGIS 3.10 )?

Correct, although potentially we could backport to 3.8

cxcandid:
looks like it doesn't work with GPKG data sources. See https://github.com/qgis/QGIS/issues/35650

PR #30844, opened 2019-07-22
Improve appearance of multiline tooltips

nyalldawson:
Avoids multiline tooltips being crammed up into a single block of text


nyalldawson:
Result:

Peek 2019-07-22 11-45

nirvn (approved):

PR #30864, opened 2019-07-23
Don't calculate the symbol outputUnit and mapUnitScale only for retrieval by unused, deprecated methods

nyalldawson:
Whilst technically this is an API break, it's costly to calculate these for symbols and the results are completely unused, except for in some deprecated getters which are very unlikely to be used by external plugins (and which shouldn't be used, in any case...)

Sponsored by QGIS grant program, identified by KDAB's hotspot

### DIFF
--- a/src/core/symbology/qgssymbol.cpp
+++ b/src/core/symbology/qgssymbol.cpp
@@ -424,9 +424,9 @@ void QgsSymbol::startRender( QgsRenderContext &context, const QgsFields &fields 
   Q_ASSERT_X( !mStarted, "startRender", "Rendering has already been started for this symbol instance!" );
   mStarted = true;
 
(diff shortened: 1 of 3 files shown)


nyalldawson:
Any objections to the (harmless) API break introduced here? @m-kuhn?

m-kuhn:
The accessors are deprecated already, correct?

Should be ok, but it would be nice to add docs to the already deprecated methods including suggestions what to user instead.

nyalldawson:
> The accessors are deprecated already, correct?

Yes

> Should be ok, but it would be nice to add docs to the already deprecated methods including suggestions what to user instead.

That's the thing with these -- I'm not sure what the alternative is, because there's absolutely no use cases for them in the symbol render context! I suspect they're just remnants from long long ago which have never been cleaned up.

PR #30892, opened 2019-07-23
 Add a reserve method to QgsGeometryCollection

nyalldawson:
Attempts to allocate memory for at least the specified number of geometries.

If the number of geometries is known in advance, calling this function
prior to adding geometries will prevent reallocations and memory fragmentation.

Sponsored by QGIS grant program, identified by KDAB's hotspot


m-kuhn (approved):
:+1:

PR #30905, opened 2019-07-24
Digitize the geometry from the relation reference widget add button

troopa81:
## Description

This PR add the possibility to draw the geometry in the mapCanvas when trying to create a new referenced feature from the relation reference widget

video_feature_parent_add_child

## Checklist

### DIFF
--- a/src/gui/editorwidgets/qgsrelationreferencewidget.cpp
+++ b/src/gui/editorwidgets/qgsrelationreferencewidget.cpp
@@ -38,6 +38,7 @@
 #include "qgsvectorlayer.h"
 #include "qgsattributetablemodel.h"
 #include "qgsmaptoolidentifyfeature.h"
+#include "qgsmaptooldigitizefeature.h"
 #include "qgsfeatureiterator.h"
 #include "qgsfeaturelistcombobox.h"
 #include "qgsexpressioncontextutils.h"
@@ -169,7 +170,6 @@ QgsRelationReferenceWidget::~QgsRelationReferenceWidget()
 {
   deleteHighlight();
   unsetMapTool();
-  delete mMapTool;
 }
 
 void QgsRelationReferenceWidget::updateIndex()
@@ -254,7 +254,9 @@ void QgsRelationReferenceWidget::setRelation( const QgsRelation &relation, bool 
 void QgsRelationReferenceWidget::setRelationEditable( bool editable )
 {
   if ( !editable )
+  {
     unsetMapTool();
+  }
 
   mFilterContainer->setEnabled( editable );
   mComboBox->setEnabled( editable );
@@ -452,9 +454,11 @@ void QgsRelationReferenceWidget::setEditorContext( const QgsAttributeEditorConte
   mCanvas = canvas;
   mMessageBar = messageBar;
 
-  delete mMapTool;
-  mMapTool = new QgsMapToolIdentifyFeature( mCanvas );
-  mMapTool->setButton( mMapIdentificationButton );
+  mMapToolIdentify.reset( new QgsMapToolIdentifyFeature( mCanvas ) );
+  mMapToolIdentify->setButton( mMapIdentificationButton );
+
+  mMapToolDigitize.reset( new QgsMapToolDigitizeFeature( mCanvas, context.cadDockWidget() ) );
+  mMapToolDigitize->setButton( mAddEntryButton );
 }
 
 void QgsRelationReferenceWidget::setEmbedForm( bool display )
@@ -717,17 +721,10 @@ void QgsRelationReferenceWidget::mapIdentification()
   if ( !mCanvas )
     return;
 
-  mMapTool->setLayer( mReferencedLayer );
-  mCanvas->setMapTool( mMapTool );
+  mMapToolIdentify->setLayer( mReferencedLayer );
+  setMapTool( mMapToolIdentify );
 
-  mWindowWidget = window();
-
-  mCanvas->window()->raise();
-  mCanvas->activateWindow();
-  mCanvas->setFocus();
-
-  connect( mMapTool, static_cast<void ( QgsMapToolIdentifyFeature::* )( const QgsFeature & )>( &QgsMapToolIdentifyFeature::featureIdentified ), this, &QgsRelationReferenceWidget::featureIdentified );
-  connect( mMapTool, &QgsMapTool::deactivated, this, &QgsRelationReferenceWidget::mapToolDeactivated );
+  connect( mMapToolIdentify, qgis::overload<const QgsFeature &>::of( &QgsMapToolIdentifyFeature::featureIdentified ), this, &QgsRelationReferenceWidget::featureIdentified );
 
   if ( mMessageBar )
   {
@@ -809,13 +806,44 @@ void QgsRelationReferenceWidget::featureIdentified( const QgsFeature &feature )
   unsetMapTool();
 }
 
+void QgsRelationReferenceWidget::setMapTool( QgsMapTool *mapTool )
+{
+  mCurrentMapTool = mapTool;
+  mCanvas->setMapTool( mapTool );
+
+  mWindowWidget = window();
+
+  mCanvas->window()->raise();
+  mCanvas->activateWindow();
+  mCanvas->setFocus();
+  connect( mapTool, &QgsMapTool::deactivated, this, &QgsRelationReferenceWidget::mapToolDeactivated );
+}
+
 void QgsRelationReferenceWidget::unsetMapTool()
 {
-  // deactivate map tool if activated
-  if ( mCanvas && mMapTool )
+  // deactivate map tools if activated
+  if ( mCurrentMapTool )
   {
     /* this will call mapToolDeactivated */
-    mCanvas->unsetMapTool( mMapTool );
+    mCanvas->unsetMapTool( mCurrentMapTool );
+
+    if ( mCurrentMapTool == mMapToolDigitize )
+    {
+      disconnect( mCanvas, &QgsMapCanvas::keyPressed, this, &QgsRelationReferenceWidget::onKeyPressed );
+      disconnect( mMapToolDigitize, &QgsMapToolDigitizeFeature::digitizingCompleted, this, &QgsRelationReferenceWidget::entryAdded );
+    }
+    else
+    {
+      disconnect( mMapToolIdentify, qgis::overload<const QgsFeature &>::of( &QgsMapToolIdentifyFeature::featureIdentified ), this, &QgsRelationReferenceWidget::featureIdentified );
+    }
+  }
+}
+
+void QgsRelationReferenceWidget::onKeyPressed( QKeyEvent *e )
+{
+  if ( e->key() == Qt::Key_Escape )
+  {
+    unsetMapTool();
   }
 }
 
@@ -940,7 +968,53 @@ void QgsRelationReferenceWidget::filterChanged()
 
 void QgsRelationReferenceWidget::addEntry()
 {
-  QgsFeature f( mReferencedLayer->fields() );
+  if ( !mReferencedLayer )
+    return;
+
+  const QgsVectorLayerTools *tools = mEditorContext.vectorLayerTools();
+  if ( !tools )
+    return;
+  if ( !mCanvas )
+    return;
+
+  // no geometry, skip the digitizing
+  if ( mReferencedLayer->geometryType() == QgsWkbTypes::UnknownGeometry || mReferencedLayer->geometryType() == QgsWkbTypes::NullGeometry )
+  {
+    QgsFeature f( mReferencedLayer->fields() );
+    entryAdded( f );
+    return;
+  }
+
+  mMapToolDigitize->setLayer( mReferencedLayer );
+  setMapTool( mMapToolDigitize );
+
+  connect( mMapToolDigitize, &QgsMapToolDigitizeFeature::digitizingCompleted, this, &QgsRelationReferenceWidget::entryAdded );
+  connect( mCanvas, &QgsMapCanvas::keyPressed, this, &QgsRelationReferenceWidget::onKeyPressed );
+
+  if ( mMessageBar )
+  {
+    QString title = tr( "Relation %1 for %2." ).arg( mRelation.name(), mReferencingLayer->name() );
+
+    QgsExpressionContext context( QgsExpressionContextUtils::globalProjectLayerScopes( mReferencingLayer ) );
+    if ( mCanvas )
+      context.appendScope( QgsExpressionContextUtils::mapSettingsScope( mCanvas->mapSettings() ) );
+
+    QgsExpression exp( mReferencingLayer->displayExpression() );
+    context.setFeature( mFormFeature );
+    exp.prepare( &context );
+    QString displayString = exp.evaluate( &context ).toString();
+
+    QString msg = tr( "Link feature to %1 \"%2\" : Digitize the geometry for the new feature on layer %3. Press &lt;ESC&gt; to cancel." )
+                  .arg( mReferencingLayer->name(), displayString, mReferencedLayer->name() );
+    mMessageBarItem = QgsMessageBar::createMessage( title, msg, this );
+    mMessageBar->pushItem( mMessageBarItem );
+  }
+
+}
+
+void QgsRelationReferenceWidget::entryAdded( const QgsFeature &feat )
+{
+  QgsFeature f( feat );
   QgsAttributeMap attributes;
 
   // if custom text is in the combobox and the displayExpression is simply a field, use the current text for the new feature
@@ -954,7 +1028,7 @@ void QgsRelationReferenceWidget::addEntry()
     }
   }
 
-  if ( mEditorContext.vectorLayerTools()->addFeature( mReferencedLayer, attributes, QgsGeometry(), &f ) )
+  if ( mEditorContext.vectorLayerTools()->addFeature( mReferencedLayer, attributes, f.geometry(), &f ) )
   {
     QVariantList attrs;
     for ( const QString &fieldName : qgis::as_const( mReferencedFields ) )
@@ -964,6 +1038,8 @@ void QgsRelationReferenceWidget::addEntry()
 
     mAddEntryButton->setEnabled( false );
   }
+
+  unsetMapTool();
 }
 
 void QgsRelationReferenceWidget::updateAddEntryButton()
@@ -1008,4 +1084,9 @@ void QgsRelationReferenceWidget::emitForeignKeysChanged( const QVariantList &for
   emit foreignKeyChanged( foreignKeys.at( 0 ) );
   Q_NOWARN_DEPRECATED_POP
   emit foreignKeysChanged( foreignKeys );
+}
+
+void QgsRelationReferenceWidget::setFormFeature( const QgsFeature &formFeature )
+{
+  mFormFeature = formFeature;
 }
--- a/src/gui/qgsattributeeditorcontext.h
+++ b/src/gui/qgsattributeeditorcontext.h
@@ -27,6 +27,7 @@
 #include "qgsproject.h"
 
 class QgsMapCanvas;
+class QgsAdvancedDigitizingDockWidget;
 
 /**
  * \ingroup gui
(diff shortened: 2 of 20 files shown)


src/gui/editorwidgets/qgsrelationreferencewidget.cpp
```diff
@@ -880,7 +909,34 @@ void QgsRelationReferenceWidget::filterChanged()
...
+  const QgsVectorLayerTools *tools = mEditorContext.vectorLayerTools();
+  if ( !tools )
+    return;
+  if ( !mCanvas )
+    return;
+
+  mMapToolDigitize->setLayer( mReferencedLayer );
+  setMapTool( mMapToolDigitize );
+
+  connect( mMapToolDigitize, static_cast<void ( QgsMapToolDigitizeFeature::* )( const QgsFeature & )>( &QgsMapToolDigitizeFeature::digitizingCompleted ), this, &QgsRelationReferenceWidget::entryAdded );
```
elpaso: can you check if `qgis::overload<T>` works for you here?

tests/src/gui/testqgsrelationreferencewidget.cpp
```diff
@@ -408,12 +406,21 @@ void TestQgsRelationReferenceWidget::testAddEntry()
...
   DummyVectorLayerTools tools;
   context.setVectorLayerTools( &tools );
-  w.setEditorContext( context, nullptr, nullptr );
+  context.setCadDockWidget( &cadDockWidget );
+  w.setEditorContext( context, &canvas, nullptr );
   w.addEntry();
 
+  QgsFeature feat( mLayer1->fields() );
+  feat.setAttribute( QStringLiteral( "pk" ), 13 );
+  feat.setAttribute( QStringLiteral( "material" ), "steel" );
```
elpaso:
```suggestion
  feat.setAttribute( QStringLiteral( "material" ), QStringLiteral( "steel" ) );
```

elpaso:
Nice! :+1:

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
troopa81: Actually, there is no need of `qgis::overload` or `static_cast<>` because there is only one version of the signal. I correct it. I wasn't aware of this problem, I see this way of connecting signal in the file and thought it was a coding guideline. I learn something today :) thanks!

elpaso: If you find any static_cast on connections, feel free to replace them with `qgis::overload<T>` it's much more readable and it will be easily replaced with `qOverload` when we will raise QT min version.

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
```diff
@@ -749,13 +751,40 @@ void QgsRelationReferenceWidget::featureIdentified( const QgsFeature &feature )
...
   {
     /* this will call mapToolDeactivated */
-    mCanvas->unsetMapTool( mMapTool );
+    mCanvas->unsetMapTool( mCurrentMapTool );
+
+    if ( mCurrentMapTool == mMapToolDigitize )
+    {
+      disconnect( mCanvas, &QgsMapCanvas::keyPressed, this, &QgsRelationReferenceWidget::onKeyPressed );
+    }
+    disconnect( mCurrentMapTool, 0, this, 0 );
```
nyalldawson: 0 -> nullptr.

But in general this looks quite dangerous to me. Isn't it possible to only disconnect the connected signals?

src/gui/qgsattributeeditorcontext.h
```diff
@@ -132,6 +133,20 @@ class GUI_EXPORT QgsAttributeEditorContext
      */
     inline QgsMapCanvas *mapCanvas() const { return mMapCanvas; }
 
+    /**
+     * Sets the associated CAD dock widget, \a cadDockWidget, (e.g. to be used in map tools).
+     * \see cadDockWidget()
+     * \since QGIS 3.10
+     */
+    inline void setCadDockWidget( QgsAdvancedDigitizingDockWidget *cadDockWidget ) { mCadDockWidget = cadDockWidget; }
```
nyalldawson:
```suggestion
    void setCadDockWidget( QgsAdvancedDigitizingDockWidget *cadDockWidget ) { mCadDockWidget = cadDockWidget; }
```

modern compilers decide this for themselves -- the inline keyword has no effect

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
troopa81: OK :ok_hand:

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
troopa81: Indeed, it's better to not disconnect everything. Corrected

src/gui/qgsmaptooldigitizefeature.h
```diff
@@ -17,31 +17,65 @@
...
-//! This tool digitizes geometry of new point/line/polygon features on already existing vector layers
-class APP_EXPORT QgsMapToolDigitizeFeature : public QgsMapToolCapture
+/**
+ * \ingroup gui
+ * \brief This tool digitizes geometry of new point/line/polygon features on already existing vector layers
+ * Once the map tool is enable, user can digitize the feature geometry
+ * A signal will then be emitted.
+ * \since QGIS 3.10
+ */
+class GUI_EXPORT QgsMapToolDigitizeFeature : public QgsMapToolCapture
```
lbartoletti: Good job!
Maybe this change deserves a comment on api_break?

troopa81: Is it? because it don't break the API, it adds one more class. Is there a special place to write that API has been augmented?

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
```diff
@@ -155,7 +156,9 @@ QgsRelationReferenceWidget::~QgsRelationReferenceWidget()
 {
   deleteHighlight();
   unsetMapTool();
-  delete mMapTool;
+  unsetMapTool();
```
m-kuhn: Double unset?

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
```diff
@@ -155,7 +156,9 @@ QgsRelationReferenceWidget::~QgsRelationReferenceWidget()
...
+  delete mMapToolIdentify;
+  delete mMapToolDigitize;
```
m-kuhn: Better use a `QObjectUniquePtr`?

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
```diff
@@ -880,7 +913,34 @@ void QgsRelationReferenceWidget::filterChanged()
...
+
+  mMapToolDigitize->setLayer( mReferencedLayer );
+  setMapTool( mMapToolDigitize );
+
+  connect( mMapToolDigitize, &QgsMapToolDigitizeFeature::digitizingCompleted, this, &QgsRelationReferenceWidget::entryAdded );
+  connect( mCanvas, &QgsMapCanvas::keyPressed, this, &QgsRelationReferenceWidget::onKeyPressed );
+
+  if ( mMessageBar )
+  {
+    QString title = tr( "Relation %1 for %2." ).arg( mRelationName, mReferencingLayer->name() );
```
m-kuhn: Would the field name be more interesting here? And possibly the displayString of the edited feature?
I'm imagining something like

> *Link feature to layer manhole "XC123"* : Digitize the geometry for the new feature on layer cover. Press Esc to cancel.

src/gui/qgsattributeeditorcontext.h
```diff
@@ -132,6 +133,20 @@ class GUI_EXPORT QgsAttributeEditorContext
      */
     inline QgsMapCanvas *mapCanvas() const { return mMapCanvas; }
 
+    /**
+     * Sets the associated CAD dock widget, \a cadDockWidget, (e.g. to be used in map tools).
+     * \see cadDockWidget()
+     * \since QGIS 3.10
+     */
+    void setCadDockWidget( QgsAdvancedDigitizingDockWidget *cadDockWidget ) { mCadDockWidget = cadDockWidget; }
```
m-kuhn: Can you tag this as private API?

I'd very much like to replace management of cad data in a separate class next to the widget (which would be the one to put in here, once it's there). The current approach has too much things in the dock widget.

And better move the definition to the cpp file (it's an implementation detail, not much to win by leaving it here, except for longer compile times when it needs to be adjusted).

src/gui/qgsattributeeditorcontext.h
```diff
@@ -132,6 +133,20 @@ class GUI_EXPORT QgsAttributeEditorContext
...
+
+    /**
+     * Returns the associated map canvas (e.g. to zoom to related features).
+     * \see setCadDockWidget()
+     * \since QGIS 3.10
+     */
+    inline QgsAdvancedDigitizingDockWidget *cadDockWidget() const { return mCadDockWidget; }
```
m-kuhn: What's the reason for inlining this?

src/gui/qgsmaptoolcapture.cpp
```diff
@@ -51,7 +51,8 @@ QgsMapToolCapture::QgsMapToolCapture( QgsMapCanvas *canvas, QgsAdvancedDigitizin
   connect( canvas, &QgsMapCanvas::currentLayerChanged,
            this, &QgsMapToolCapture::currentLayerChanged );
 
-  currentLayerChanged( canvas->currentLayer() );
+  if ( canvas )
+    currentLayerChanged( canvas->currentLayer() );
```
m-kuhn: Line 51 should probably also be in here if canvas can be `nullptr`

src/gui/qgsmaptooldigitizefeature.h
```diff
@@ -17,31 +17,65 @@
...
+#include "qgis_gui.h"
 
 class QgsFeature;
 
-//! This tool digitizes geometry of new point/line/polygon features on already existing vector layers
-class APP_EXPORT QgsMapToolDigitizeFeature : public QgsMapToolCapture
+/**
+ * \ingroup gui
+ * \brief This tool digitizes geometry of new point/line/polygon features on already existing vector layers
+ * Once the map tool is enable, user can digitize the feature geometry
```
m-kuhn:
```suggestion
 * Once the map tool is enabled, user can digitize the feature geometry.
```

src/gui/qgsmaptooldigitizefeature.h
```diff
@@ -17,31 +17,65 @@
...
+     */
+    QgsMapToolDigitizeFeature( QgsMapCanvas *canvas, QgsAdvancedDigitizingDockWidget *cadDockWidget, CaptureMode mode = QgsMapToolCapture::CaptureNone );
 
     void cadCanvasReleaseEvent( QgsMapMouseEvent *e ) override;
 
-    virtual void digitized( QgsFeature &f );
+    /**
+     * Change the layer edited by the map tool
+     * \param vl the layer to be edited by the map tool
+     * \since 3.10
```
m-kuhn: I'd remove the \since here, it "inherits" the \since from the class.

src/gui/qgsmaptooldigitizefeature.h
```diff
@@ -17,31 +17,65 @@
...
+     */
+    void setLayer( QgsMapLayer *vl ) { mLayer = vl; }
```
m-kuhn: I'd move this implementation to the cpp file.

I'd expect some additional safety to be added in the future (e.g. cleaning up a rubberband if geometry type changes - or always, etc.). What it does is an implementation detail and not public API.

src/gui/qgsmaptooldigitizefeature.h
```diff
@@ -17,31 +17,65 @@
...
+     */
     void digitizingFinished( );
 
   protected:
 
+    /**
+     * Finish the feature digitizing
+     * \param f the new created feature
+     */
+    virtual void digitized( QgsFeature &f );
```
m-kuhn: private?

m-kuhn: `const QgsFeature`?

src/gui/qgsmaptooldigitizefeature.h
```diff
@@ -17,31 +17,65 @@
...
+    /**
+     * Emitted whenever the digitizing has been ended without digitizing
+     * any feature
+     */
     void digitizingFinished( );
 
   protected:
 
+    /**
+     * Finish the feature digitizing
```
m-kuhn:
```suggestion
     * Called when the feature has been digitized.
```

m-kuhn (commented):
Nice feature.
This has been on my wish list for ages!
There are a couple of considerations I left in the code. Would be awesome to bring this into 3.10, but the time is getting short.

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
troopa81: Corrected

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
troopa81: Done.
Just curious, why using a QObjectUniquePtr instead of a std::unique_ptr or a QScopedPointer ?

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
troopa81: I change the message but the feature display string is way more difficult to add. Indeed, I assume you talk about the **Column preview** that change the main display of feature list.

The relation reference widget has no reference on this column preview expression. We would need to add a variable in QgsAttributeEditorContext and set it from the QgsAttributeForm or something like that. Do you think it worth the modification only for this message to be displayed?

src/gui/qgsattributeeditorcontext.h
troopa81: > Can you tag this as private API?

How do you do that? Is there a special doxygen keyword or something?

> I'd very much like to replace management of cad data in a separate class next to the widget (which would be the one to put in here, once it's there). The current approach has too much things in the dock widget.

I agree

> And better move the definition to the cpp file (it's an implementation detail, not much to win by leaving it here, except for longer compile times when it needs to be adjusted).

There is no cpp file at the moment. Do you want me to create one, with just this 2 methods? I'm not sure to understand why it matters.

src/gui/qgsattributeeditorcontext.h
troopa81: It was to be consistent with other methods declaration (except I forgot to put it in front on setCadDockWidget), but I'm pretty sure this is useless here. 

I removed it.

src/gui/qgsmaptooldigitizefeature.h
troopa81: > private?

The digitized method is inherited by the [QgsMapToolAddFeature](https://github.com/troopa81/QGIS/blob/feature_createfeat_from_relationrefwidget/src/app/qgsmaptooladdfeature.cpp#L62)

> const QgsFeature?

Could be. But we will need to copy it later in order to modify it (when we display the form for instance). I didn't modify the original code.

m-kuhn: used by or implemented by?
Private virtual can be overridden.

The question regarding const is if the implementing method is expected to change the data. If yes, the meaning and effect of such a change should be documented.

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
m-kuhn: QObjects can have a parent and if they have, they are owned by it. If the parent object is deleted first, other smart pointers are left pointing to an already deleted object.

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
m-kuhn: Not sure what the column preview is, the main display uses the displayString of the layer if I'm not wrong.

src/gui/qgsattributeeditorcontext.h
m-kuhn: Just add a note that this is unstable API, there is no official tag or similar.

Regarding the cpp, feel free to add one. It doesn't mater that much, it's just about isolation API (headers) from implementation (cpp)

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
troopa81: OK, thks

src/gui/qgsmaptooldigitizefeature.h
m-kuhn: Not really. Additions are not documented.
If someone wants a diff, it should be possible to generate that from the dox (would be a nice little project ;) )

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
troopa81: Something like this 89e76ca ? 

I add a setFormFeature method on the qgsrelationreferencewidget, and call the method from the wrapper. Is that OK to do this ?

m-kuhn: Yes, that should be ok.
Just one small thing, prefer putting the method definition into the cpp file.

src/gui/qgsattributeeditorcontext.h
troopa81: OK done

src/gui/qgsattributeeditorcontext.h
```diff
@@ -135,10 +135,12 @@ class GUI_EXPORT QgsAttributeEditorContext
 
     /**
      * Sets the associated CAD dock widget, \a cadDockWidget, (e.g. to be used in map tools).
+     * This method will eventually be replaced by another method using CAD data instead of
```
m-kuhn: Add something like the following instead.

```
\note Unstable API. This method is unstable API and may be modified or removed at any time.
```

src/gui/qgsmaptooldigitizefeature.h
troopa81: Done

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
troopa81: OK, done

src/gui/qgsattributeeditorcontext.h
```diff
@@ -132,6 +133,21 @@ class GUI_EXPORT QgsAttributeEditorContext
...
+    /**
+     * Sets the associated CAD dock widget, \a cadDockWidget, (e.g. to be used in map tools).
+     * \note Unstable API. This method is unstable API and may be modified or removed at any time.
+     * \see cadDockWidget()
+     * \since QGIS 3.10
+     */
+    void setCadDockWidget( QgsAdvancedDigitizingDockWidget *cadDockWidget );
+
+    /**
+     * Returns the associated map canvas (e.g. to zoom to related features).
```
m-kuhn: Can you fix this docstring and add a unstable api note too?

src/gui/qgsmaptoolcapture.cpp
m-kuhn: Any feedback on that?

tests/src/gui/testqgsrelationreferencewidget.cpp
```diff
@@ -408,12 +406,21 @@ void TestQgsRelationReferenceWidget::testAddEntry()
...
+  context.setCadDockWidget( &cadDockWidget );
+  w.setEditorContext( context, &canvas, nullptr );
   w.addEntry();
 
+  QgsFeature feat( mLayer1->fields() );
+  feat.setAttribute( QStringLiteral( "pk" ), 13 );
+  feat.setAttribute( QStringLiteral( "material" ), QStringLiteral( "steel" ) );
+  feat.setAttribute( QStringLiteral( "diameter" ), 140 );
+  feat.setAttribute( QStringLiteral( "raccord" ), "collar" );
+  w.mMapToolDigitize->digitized( feat );
```
m-kuhn: Not completely sure: shouldn't that be a new test instead of modifying an existing one?

m-kuhn (commented):
Just some small things left. Would be nice if you could have a look at the three comments soon to get this merged early on this week.

src/gui/qgsattributeeditorcontext.h
troopa81: Done

tests/src/gui/testqgsrelationreferencewidget.cpp
troopa81: Actually, the old behavior is not existing anymore except for referenced layer with no geometry. And, it was not working. I fixed it and add a test for layer with no geometry. c4e7cf1

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
```diff
@@ -880,7 +908,53 @@ void QgsRelationReferenceWidget::filterChanged()
...
+  const QgsVectorLayerTools *tools = mEditorContext.vectorLayerTools();
+  if ( !tools )
+    return;
+  if ( !mCanvas )
+    return;
+
+  // no geometry, skip the digitizing
+  if ( mReferencedLayer->geometryType() == QgsWkbTypes::UnknownGeometry || mReferencedLayer->geometryType() == QgsWkbTypes::NullGeometry )
+  {
+    QgsFeature f( mReferencedLayer->fields(), 0 );
```
m-kuhn: What's the 0 for?

troopa81: for nothing actually, it's just a bad copy paste. I fixed it

troopa81:
> Any feedback on that?

I just remove it, it was a remaining from a previous version. 

I commented here because for some reason, I cannot comment under the review or resolved it. I don't know why.

PR #30923, opened 2019-07-25
Copy/paste symbols in style manager

nyalldawson:
Allows symbols to be copied or pasted in the style manager dialog.

@nirvn, as previously discussed...

### DIFF
--- a/src/gui/symbology/qgsstylemanagerdialog.cpp
+++ b/src/gui/symbology/qgsstylemanagerdialog.cpp
@@ -2211,8 +2211,11 @@ void QgsStyleManagerDialog::listitemsContextMenu( QPoint point )
   // Clear all actions and create new actions for every group
   mGroupListMenu->clear();
 
+  const QModelIndexList indices = listItems->selectionModel()->selectedRows();
+
   if ( !mReadOnly )
   {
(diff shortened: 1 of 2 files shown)


src/gui/symbology/qgsstylemanagerdialog.cpp
```diff
@@ -222,6 +223,11 @@ QgsStyleManagerDialog::QgsStyleManagerDialog( QgsStyle *style, QWidget *parent,
     mCopyToDefaultButton->hide();
   }
 
+  mActionCopyItem = new QAction( tr( "Copy Item" ), this );
+  connect( mActionCopyItem, &QAction::triggered, this, &QgsStyleManagerDialog::copyItem );
+  mActionPasteItem = new QAction( tr( "Paste Item…" ), this );
```
m-kuhn: I think the paste normally has no ellipsis in other places, but prove me wrong

```suggestion
  mActionPasteItem = new QAction( tr( "Paste Item" ), this );
```

m-kuhn (approved):

src/gui/symbology/qgsstylemanagerdialog.cpp
nyalldawson: In this case it opens a dialog asking for symbol name and tags, so an ellipsis is appropriate

PR #30926, opened 2019-07-25
Default to QAbstractItemView::ExtendedSelection in pg and oracle source select

m-kuhn:
This aligns the default selection behavior of the postgres and oracle source selection with the rest of the application and other providers like geopackage.

For newcomers, the MultiSelection is rather confusing. Especially with the current behavior inside the unified data source dialog, that remembers previous selections it's very hard to clear pre-existing selections in large lists that are partially hidden.


nyalldawson:
Big +1. I'm also +1 to drop the double click option (or at least hide it from the UI), it's total overkill to have this customisable.

m-kuhn:
I am also in favour of removing the option (I just wanted to keep the controversy risk for this PR minimal), will also remove that if no one objects.

nyalldawson (approved):

nyalldawson:
I **think** I saw in the past that other providers have copied this same double click heading code too (even without actually exposing the option!!)

PR #30927, opened 2019-07-25
 Reset selection when data source select dialog is reopened

m-kuhn:
When the data source select dialog is closed and reopened it's much more likely that the user wants to add a different layer than last time he was using the dialog. So instead of offering him to add the same layer again, the selection is cleared and he is requested to make a new selection.


nirvn:
I don't have strong feelings on this one, but the argument makes sense :) code looks good.

nirvn (approved):

PR #30929, opened 2019-07-25
[QgsQuick] Custom filepaths for external resource widget

sklencar:
* Added support for defining defaultPath:
 while taking pictures with external widget's photoPanel. Default path is taken from a widget config `path -> default path`. Also options `Relative to project path` or `Relative to default path` are taken into account. In addition, default path can also contain expression, which is evaluated in global, project and layer scope. Therefore these properties have been added while creating widgets.
* Loading PhotoCapturePanel only when needed:
The panel has been loaded with a form even if it wasn't needed at that moment. In addition, it was asking permissions for accessing camera on form load which was odd. Since the loader is used, its loaded after 'capture' action is initialised. 
* Long filenames wrapping
WrapAtWordBoundaryOrAnywhere is used now to prevent overflowing of a long name of files.

### DIFF
--- a/src/quickgui/plugin/editor/qgsquickexternalresource.qml
+++ b/src/quickgui/plugin/editor/qgsquickexternalresource.qml
@@ -24,6 +24,33 @@ import QgsQuick 0.1 as QgsQuick
  * Requires various global properties set to function, see qgsquickfeatureform Loader section
  * Do not use directly from Application QML
  * The widget is interactive which allows interactions even in readOnly state (e.g showing preview), but no edit!
+ *
+ * Overview of path handling
+ * -------------------------
+ *
+ * Project home path: comes from QgsProject::homePath() - by default it points to the folder where
+ * the .qgs/.qgz project file is stored, but can be changed manually by the user.
+ *
+ * Default path: defined in the field's configuration. This is the path where newly captured images will be stored.
+ * It has to be an absolute path. It can be defined as an expression (e.g. @project_home || '/photos') or
+ * by plain path (e.g. /home/john/photos). If not defined, project home path is used.
+ *
+ * In the field's configuration, there are three ways how path to pictures is stored in field values:
+ * absolute paths, relative to default path and relative to project path. Below is an example of how
+ * the final field values of paths are calculated.
+ *
+ *   variable         |     value
+ * -------------------+--------------------------------
+ * project home path  |  /home/john
+ * default path       |  /home/john/photos
+ * image path         |  /home/john/photos/img0001.jpg
+ *
+ *
+ *    storage type          |   calculation of field value    |      final field value
+ * -------------------------+---------------------------------+--------------------------------
+ * absolute path            |  image path                     |   /home/john/photos/img0001.jpg
+ * relative to default path |  image path - default path      |   img0001.jpg
+ * relative to project path |  image path - project home path |   photos/img0001.jpg
  */
 Item {
   signal valueChanged(var value, bool isNull)
@@ -86,6 +113,10 @@ Item {
     externalResourceHandler.onFormCanceled(fieldItem)
   }
 
+  function getAbsolutePath(prefix, pathFromValue) {
+    return (prefix) ? prefixToRelativePath + "/" + pathFromValue : pathFromValue
+  }
+
   id: fieldItem
   enabled: true // its interactive widget
   height: customStyle.fields.height * 3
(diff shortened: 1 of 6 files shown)


src/quickgui/plugin/editor/qgsquickexternalresource.qml
```diff
@@ -36,6 +36,30 @@ Item {
...
+      }
+    }
+
+    if (expression) {
+      QgsQuick.Utils.evaluateExpression(featurePair, activeProject, expression)
+    } else {
+      config["DefaultRoot"] ? config["DefaultRoot"] : homePath
+    }
+  }
+  property string prefixToRelativePath: relativeStorageMode === 2 ? targetDir : homePath
```
wonder-sk: for absolute paths this should be probably empty

src/quickgui/plugin/editor/qgsquickexternalresource.qml
```diff
@@ -67,13 +91,13 @@ Item {
...
 
-  QgsQuick.PhotoCapture {
-    id: photoCapturePanel
-    visible: false
-    height: window.height
-    width: window.width
-    edge: Qt.RightEdge
-    imageButtonSize: fieldItem.iconSize
+  Loader {
+    id: photoCaptuePanelLoader
```
wonder-sk: s/Captue/Capture/

src/quickgui/plugin/editor/qgsquickexternalresource.qml
```diff
@@ -97,7 +121,7 @@ Item {
 
       MouseArea {
         anchors.fill: parent
-        onClicked: externalResourceHandler.previewImage(homePath + "/" + image.currentValue)
+        onClicked: externalResourceHandler.previewImage(prefixToRelativePath + "/" + image.currentValue)
```
wonder-sk: handling of absolute paths is missing (should be just image.currentValue). I would suggest to wrap it into a function that would return either "prefix/image.currentValue" or "image.currentValue" based on the absolute/relative path settings

src/quickgui/plugin/qgsquickfeatureform.qml
```diff
@@ -62,6 +62,17 @@ Item {
           */
         property var removeImage: function removeImage(itemWidget, imagePath) {
         }
+
+        /**
+          * Called when clicked on the OK icon after taking a photo with the Photo panel.
+          * \param itemWidget editorWidget for modified field to send valueChanged signal.
+          * \param prefixToRelativePath Absolute path to the image extracted by current value.
```
wonder-sk: This is not absolut path to the image right? just the path prefix?

src/quickgui/plugin/qgsquickphotopanel.qml
```diff
@@ -22,7 +22,12 @@ import QtGraphicalEffects 1.0
 import QgsQuick 0.1 as QgsQuick
 
 Drawer {
+  // Capture path
   property var targetDir
+  // Path to project
+  property var homePath
```
wonder-sk: Is this needed? I can't see it used anywhere in the file

src/quickgui/plugin/qgsquickphotopanel.qml
```diff
@@ -22,7 +22,12 @@ import QtGraphicalEffects 1.0
...
+  // Along with lastPhotName creates an absolute path to a photo. Its either project path or defaultRoot.
```
wonder-sk: s/Phot/Photo/

wonder-sk (commented):
Because the path handling is not completely trivial, I would suggest to add an overview like this somewhere in the external resource widget source file (it took me a bit of time to figure out how all the things are supposed to work)
```
Overview of path handling
-------------------------

Project home path: comes from QgsProject::homePath() - by default it points to the folder where
the .qgs/.qgz project file is stored, but can be changed manually by the user.

Default path: defined in the field's configuration. This is the path where newly captured images will be stored.
It has to be an absolute path. It can be defined as an expression (e.g. @project_home || '/photos') or
by plain path (e.g. /home/john/photos). If not defined, project home path is used.

In the field's configuration, there are three ways how path to pictures is stored in field values:
absolute paths, relative to default path and relative to project path. Below is an example of how
the final field values of paths are calculated.

  variable         |     value
-------------------+--------------------------------
project home path  |  /home/john
default path       |  /home/john/photos
image path         |  /home/john/photos/img0001.jpg


   storage type          |   calculation of field value    |      final field value
-------------------------+---------------------------------+--------------------------------
absolute path            |  image path                     |   /home/john/photos/img0001.jpg
relative to default path |  image path - default path      |   img0001.jpg
relative to project path |  image path - project home path |   photos/img0001.jpg
```

wonder-sk (approved):

Issue #26948, opened 2018-06-05
Crash when clicking browse button in attachment edit widget (PLEASE RE-ADD THE BUTTON)


qgib:
26.7.2019 - @m-kuhn 

> The browse button has been removed as a band aid to prevent the crash.
> Fixing the browse button will require additional work based on the analysis in https://github.com/qgis/QGIS/issues/26948#issuecomment-514972558 .

-------------------

Author Name: **Tom Chadwin** (@tomchadwin)
Original Redmine Issue: [19118](https://issues.qgis.org/issues/19118)
Affected QGIS version: 3.7(master)
Redmine category:edit_widget

---
Steps to recreate:

1. Open a vector layer with at least one text field
2. Change a text field's edit widget to Attachment
3. Open the layer's attribute table
4. Toggle editing
5. Highlight a value in the column whose edit widget you just changed
6. Click the ellipsis browse button

On Win 7 and 10, we have confirmations that this causes a full QGIS crash, with no QGIS crash handler.

---

- [2019-01-29_09-29-35.gif](https://issues.qgis.org/attachments/download/14190/2019-01-29_09-29-35.gif) (Tom Chadwin)
- [Screenshot_2019-05-22_11-37-15.png](https://issues.qgis.org/attachments/download/14971/Screenshot_2019-05-22_11-37-15.png) (Grga Franges)

---

Related issue(s): #27961 (duplicates), #29182 (duplicates), #29271 (duplicates), #29408 (duplicates), #29590 (relates)
Redmine related issue(s): [20140](https://issues.qgis.org/issues/20140), [21365](https://issues.qgis.org/issues/21365), [21454](https://issues.qgis.org/issues/21454), [21592](https://issues.qgis.org/issues/21592), [21775](https://issues.qgis.org/issues/21775)

---

gioman:
This is still true on 3.4.9.

m-kuhn:
Is there a stack trace available for this crash?

tomchadwin:
Not on Windows, no. QGIS crashes without the crash handler dialog appears, so no stack trace.

nirvn:
@m-kuhn , sending over a stack trace. While I figured out the crash has to do with the widget we pass as first argument in QFileDialog::getOpenFileName, I can't find what's wrong here. Hope you'll have more luck than me.

```
#0  0x00007ffff326ded7 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
[...]

PR #30930, opened 2019-07-25
Hide browse button on attachment widget in attribute table

m-kuhn:
Basically a workaround for a crash in the attachment/external resource widget. By hiding the browse button it's simply not possible for a user to trigger the crash.

It would be nicer to "really" fix the issue, but that will probably involve advanced black magic, to control the focus of the editor while the file browser dialog is open. Or prevent the widget from being deleted when the focus changes (but still making sure it's deleted at an appropriate time). Or maybe it's not that complicated and someone just needs to find the right Qt docs that advise how to properly solve the issue.

Be it as it may, this band aid patch avoids a crash/data corruption.

References #26948

### DIFF
--- a/src/gui/editorwidgets/qgsexternalresourcewidgetwrapper.cpp
+++ b/src/gui/editorwidgets/qgsexternalresourcewidgetwrapper.cpp
@@ -175,7 +175,8 @@ void QgsExternalResourceWidgetWrapper::initWidget( QWidget *editor )
     }
     if ( cfg.contains( QStringLiteral( "FileWidgetButton" ) ) )
     {
-      mQgsWidget->fileWidget()->setFileWidgetButtonVisible( cfg.value( QStringLiteral( "FileWidgetButton" ) ).toBool() );
+      // Prevent from showing the button in the attribute table, see https://github.com/qgis/QGIS/issues/26948
+      mQgsWidget->fileWidget()->setFileWidgetButtonVisible( cfg.value( QStringLiteral( "FileWidgetButton" ) ).toBool() && context().formMode() != QgsAttributeEditorContext::Popup );
     }
     if ( cfg.contains( QStringLiteral( "DocumentViewer" ) ) )
     {


gioman:
@m-kuhn but... without the button how a user will select a file and have the path filled in the table?

m-kuhn:
He won't.
I've spent quite a bit of time looking for a better solution, but couldn't come up with something better than this. I'll be happy to have another look if the priority is high enough for someone to invest into this.

As @elpaso said, funds from "Spare Time Inc." are very limited

gioman:
> As @elpaso said, funds from "Spare Time Inc." are very limited

I understand, but this if this is not a proper fix then is much "better" to remove completely the "attachment" widget and just say something like "removed as it is broken and no one could fix it (in its spare time)". I'm very -1 on this patch, looking at it as a user.

m-kuhn:
It will still work properly inside the form. I think only reducing functionality in the attribute table where it's broken is preferable over reducing it everywhere.

gioman:
> I think only reducing functionality in the attribute table where it's broken is preferable over reducing it everywhere.

I 'm still against it. We will forget about it and no one will ever fix what is an obvious regression compared to our old LTR. Removing the entire "attachment" widget will make more obvious that a functionality has been removed and need to be fixed and reinstated (not that I like this option too).

m-kuhn:
I changed it from "fixes" to "references" so the issue will stay open.

gioman:
Even better, leave at it is. Maybe someone that is bothered by the crash will support the work for a real fix.

m-kuhn:
Description of the current status:

1. the issue is severe (causing data loss by crashing qgis because of a simple bad click)
2. the widget has much more to offer (e.g. picture preview, showing a hyperlink) and works fine in the form
3. the bug has been around since QGIS 3.0, nobody cared enough

For me point 1 is enough to do something, point 2. is enough to not completely remove it and point 3. a reason to assume that it's unlikely to help if we do nothing.

But let's see what others say.

gioman:
> the bug has been around since QGIS 3.0, nobody cared enough

true (that someone has not stepped in to fund the fix, but that does not mean that is not affecting a lot of people), but we should never ship a new LTR with a clear regression compared to the old one, especially when it regards not an esoteric functionality but rather something that is pretty common to use.

gioman:
I still really prefer leave it as it is. Document the crash and at the next round of paid bug fixing the PSC should really make a list of tickets that must be fixed first, rather than leaving it completely to the choice of the applying developers.

nirvn:
-1 to not apply a bandaid to a crash that can lead to data loss.

gioman:
> -1 to not apply a bandaid to a crash that can lead to data loss.

is does not lead to data loss more than any other crash a user can get. @nirvn

nyalldawson:
Yeah, I'm +1 to the workaround for now. I've also investigated this one in the past and could not find a solution. 

Unless @wonder-sk steps in with an inspired brainwave, there is likely NO possible fix.

gioman:
> there is likely NO possible fix.

that is NOT a good news. But anyway we should get a better band aid rather than just removing the button. So still -1 for now.

nyalldawson:
@gioman what do you propose? No one has any other ideas for this one, despite multiple Devs trying seperate approaches. Currently we're stuck at a stale mate between a know severe crash vs preventing a crash with loss of functionality.

nirvn:
Let me rephrase: -1 to not apply an available bandaid to a known crash which causes data loss.

Of course this isn't the only way QGIS can crash but I'm against not patching the crashes we are aware of an know how to plug.

nirvn:
Also, for the record, this isn't a scenario where it crashes sometimes and work some other time. It'll kill QGIS 100 times out of 100. So there's no actual loss.

m-kuhn:
> vs preventing a crash with loss of functionality

compared to 2.18 it's a loss of functionality. compared to 3.4 it's loss of a crash with no loss of functionality.

m-kuhn:
> there is likely NO possible fix.

Btw, I still think there is a possible fix by working with event filter or possibly implementing some virtual methods on the table view / delegate, but this needs more time than I can afford right now.

So let's go ahead and kill a button whose single purpose right now is to kill QGIS and postpone everything until the next bugfixing budget from qgis.org is out or someone else extremely annoyed by the missing functionality jumps in.

gioman:
> So let's go ahead and kill a button whose single purpose right now is to kill QGIS and postpone everything until the next bugfixing budget from qgis.org is out or someone else extremely annoyed by the missing functionality jumps in.


What is the right way to NEVER FORGET about this after the button is removed? By the way: I'm annoyed by this bug. It broke workflows to a point that in a few cases I had to stick to 2.18 (that from some stats is still the most used version). Now do not ask me to support this fix right now, I support the project in several ways (including financially) and I really think that ultimately fixing regressions from one LTR to the other should be mainly on the project shoulders.

m-kuhn:
@gioman we all sit in the same boat here. I'd LOVE to be able to offer a better fix and I hate to remove functionality that once worked.
Still I think under the conditions right now we need to be pragmatic and prevent the crash with the means we have.
You are regularly asked to prepare lists of high priority bugs for the bugfixing sprints, I'd say it's best if you put that on the list for the next release.

m-kuhn:
Thanks for the feedback everyone.
I'm sorry I couldn't offer a better solution for now. Let's keep this on the todo list for 3.10 though!

gioman:
> Thanks for the feedback everyone.

@m-kuhn please revert immediately, this needs to be discussed more. I was about to start writing an email to the PSC asking to support the fix.

m-kuhn:
@gioman you are very welcome to write an email to the PSC (or anyone else with the resources to provide a proper fix). Merging this pull request is no statement that a proper fix wouldn't be preferred.

gioman:
> @gioman you are very welcome to write an email to the PSC (or anyone else with the resources to provide a proper fix). Merging this pull request is no statement that a proper fix wouldn't be preferred.

@m-kuhn you failed yesterday to tell me how you pretend that we will NEVER forget about this regression in order to FIX it asap, among the other things.  I must say that in 10+ years of following the project thos is probably the first time I'm very displeased abiut the process of how a controversial patch has been applied. I kindly ask again to revert and wait further discussion.

m-kuhn:
@gioman, I fail to understand how you pretend that allowing QGIS to crash and lead to data loss will substantially increase the chance of FIXING this situation reported over a year ago.

luipir:
@gioman what about to leave merged (because of the above tech about the difficulties to fix) and maintain the issue in a todo list ? is it enough to avoid to forget it?

gioman:
> @gioman what about to leave merged (because of the above tech about the difficulties to fix) and maintain the issue in a todo list ? is it enough to avoid to forget it?

@luipir I'm afraid it is not. The supported/paid bug fixing effort do not mandates to fix any particular bug, what will garauntee that this will be ever fixed really?

nyalldawson:
@gioman - your argument is supposing this one CAN be fixed. I'm not convinced. I spent a good 3 hours or so investigating it last year, and Matthias and Mathieu have independently clocked up some hours.

Throwing more hours at a problem doesn't gaurantee a fix. It probably needs work in qt itself, which none of us are qualified to do.

nirvn:
@gioman , you seem to suggest we should allow a known crasher leading to data loss to live on continue negatively impacting our user base in order to eventually crash one time too many on someone's machine out there who could either have the coding skills or the money to finance a fix. That doesn't sound like a good approach to me.

Let's protect our users from losing data (which @m-kuhn just did), keep the 13-month-old issue open (with a regression flag), and stress it's importance in case the functionality regression can be addressed down the road.

nyalldawson:
(now I'm wondering where for future bug fixing sprints we should approach and fund kdab to handle qt issues...?)

luipir:
> (now I'm wondering where for future bug fixing sprints we should approach and fund kdab to handle qt issues...?)

+1

elpaso:
I'm +1 on this, of course I also agree that this must be fixed with high priority, but disabling it for now looks the right thing to do.

m-kuhn:
> (now I'm wondering whether for future bug fixing sprints we should approach and fund kdab to handle qt issues...?)

Yes, having 1 week of kdab development power for a next release sounds like a very valuable thing to invest into.

nyalldawson:
Qt bugs I can think of immediately which impact us directly:
- https://bugreports.qt.io/browse/QTBUG-76821
- https://bugreports.qt.io/browse/QTBUG-69463
- https://bugreports.qt.io/browse/QTBUG-1396
.... and the infamous, longstanding pain in QGIS' side:
- https://bugreports.qt.io/browse/QTBUG-1865

gioman:
> I'm +1 on this, of course I also agree that this must be fixed with high priority, but disabling it for now looks the right thing to do.

There was no need to RUSH a merge, we are not near a new point release. This would have allowed a wider discussion, involving also the (affected) users as clearly this github discussions are biased towards the opinion of the developers.

nyalldawson:
There's also a bug which impacted reading delimited text on macos, which was due to a limitation in qt's text parsing engine (and I can't remember specifics now - @NathanW2 I remember looking into this with you once...)

nyalldawson:
Ah  - this one. https://bugreports.qt.io/browse/QTBUG-18038 , causing https://github.com/qgis/QGIS/issues/21976, https://github.com/qgis/QGIS/issues/17364, https://github.com/qgis/QGIS/issues/17190

gioman:
> That doesn't sound like a good approach to me.

I'm usually (of course) in favor of anything that avoids crashes. I'm not in favor of removing crash by removing a functionality (used by many, that are sitting on 2.18 and awaiting for a fix on 3) without any guarantee that at the next round of supported bug fixing this will solved one way or the other. If this is an upstream issue renders the actual functionality unfix-able in QGIS then an alternative solution (to re-instate the functionality) must be found, and the project would need to support all the necessary effort.

m-kuhn:
Working with KDAB on this might also help to establish communication channels for new functionality like support for CMYK.

m-kuhn:
@gioman you are right that there was no rush and a merge could have waited. Personally, I don't think that would have changed much given that there is still no alternative under discussion.

This aside, I will not merge the backports until some days before the next patch release so we can take any outcome of broader discussions that you trigger into account there!

gioman:
@m-kuhn 

>  you are right that there was no rush and a merge could have waited. Personally, I don't think that would have changed much given that there is still no alternative under discussion.

my question stand, now that the functionality has been removed, when this will be tackled to look for a proper fix (it does not matter if is upstream or not)?

do you (or any other developer) have any plan to tackle this as a priority during the next (paid) bu squashing effort?

Can you please broadcast to the psc and the user lists that a functionality has been removed (leaving a lot of people, that were awaiting for a fix, scratching their heads) and that this will need a  serious investigation/effort to make it happen again? Thanks.

m-kuhn:
Ok, I will write to the PSC.

jukos:
I am also waiting for an solution since 1 year. And I am not alone ( see for example https://gis.stackexchange.com/questions/310446/no-images-in-popups-only-url-is-displayed/310974#310974). 
I want use the plugin qgis2web from Tom Chadwin. But with this bug it is impossible to use the plugin fully. qgis2web is the easiest and the best way to produce interactive maps, very fast. 
PLEASE do all to solve the problem! Waiting is not an option. 
There are no alternative to solve the problem. 
I am very shocked about this bug - week for week for week for week.....nothing.

m-kuhn:
Interesting, I am not sure how this is connected to the qgis2web plugin, this just affects a button in the attribute table. It's not related to image preview and I'm unsure how qgis2web could rely on this.

@tomchadwin might be able to shed some light.

jukos:
I wrote an comment within the gis.stackexchange thread above. I am sure he can give a better description of the problem.

tomchadwin:
Fields with image edit widgets result in images displayed in feature popups in webmaps exported by qgis2web. If the image edit widget is not selected, the path appears in the popup instead of the image. It's therefore important functionality. I've had quite a few bug reports which I've avoided investigating because of this upstream QGIS attribute table crash.

m-kuhn:
@tomchadwin I still fail to see the connection to qgis2web. This issue only affects the browse button, how is it required to click this button to work with qgis2web?

tomchadwin:
@m-kuhn Because that widget is required, but had an associated crash, I've told people to avoid it. I'm perfectly happy for the browse button to be removed as a temporary fix. 

So is this a Qt regression from 4 to 5?

m-kuhn:
Thanks 👍 If that's the connection it can be treated as resolved with this patch.

Yes, for what I can see it's an issue that is caused by upgrading from Qt 4 to Qt 5.

jukos:
Hello,

> "Yes, for what I can see it's an issue that is caused by upgrading from Qt 4 to Qt 5."

Sure, it is good to identify the problem of the source. 
But......
I get from you the information as a user. A user of some or many, who is waiting of the repair of the function. What should  I think now? Should I wait on more year for an solution? Should I change the software? Should I work with the old QGIS2? What?
What I want is: I expect not a wonder, but I expect more informations: Is it possible to solve the problem in the next weeks? Works anybody of the programmers - in this time? Can I have hope to have this function in the NEAR future?
I want only have informations.
Is that too much?
I am a little bit annoyed.

Greetings, Jürgen

m-kuhn:
Hi Jürgen,

The QGIS project has initiated a project to work with upstream Qt developers to look into this issue and others. At the current point it is unfortunately not possible to give any guarantee of an ETA for a fix.

Here the available options you have as a user that come to my mind:

 - Stick to QGIS 2
 - Switch the software
 - Contact your local point of contact of the company you have a QGIS support contract with
 - Manage images via feature forms instead of the attribute table
 - Implement a plugin to work around this functionality until it's fixed upstream

Hope this clears the dust a bit.

jukos:
Thank you for your answer and commitment.
Greetings, Jürgen

PR #30935, opened 2019-07-25
[processing] Use lists, not maps in overlap analysis algorithm

nyalldawson:
Ensures consistent field ordering and that fields are always matched to input layer order


nyalldawson:
FYI @anitagraser

PR #30947, opened 2019-07-26
[FEATURE] [needs-docs] Add circular data dependencies

troopa81:
## Description

This PR is the first step of the proposed GRANT [Snap Cache Improvements](https://docs.google.com/document/d/1VzgLJ-hy2GtcWrAy5geivhDDdXRGJUDcikCWUtXRe9k/edit#heading=h.uau3di2m4nvn). 

If fixes the QGEP issue described [here](https://github.com/QGEP/QGEP/issues/492)

It add the possibility to define circular data dependencies (layer1 depends on layer2 that depends on layer 1), and by extension on the layer itself (layer1 depends on layer1). This last is useful for instance when you modify one line in a line layer that trigger connected lines modifications on database side. In this case, `dataChanged` signal needs to be fired in order to refresh the snapping cache (and avoid what is described in QGEP issue). 

This PR doesn't propose any optimisations on snap cache indexing. It will be done in a separate PR.

**API BREAK** : I remove the `QgsMapLayer::hasDependencyCycle` method because it has no use anymore. Even if the method was protected, it changes the API. Is it an issue? Is it better to keep it and mark it as deprecated? I remove it for now.

cc @mhugo @lbartoletti @m-kuhn @haubourg 

## Checklist

### DIFF
--- a/src/core/qgsmaplayer.h
+++ b/src/core/qgsmaplayer.h
@@ -1501,8 +1501,8 @@ class CORE_EXPORT QgsMapLayer : public QObject
 
     /**
      * Checks whether a new set of dependencies will introduce a cycle
-     * \deprecated since QGIS 3.10, this method always return false, because circular dependencies
-     * are now correctly managed.
+     * this method is now deprecated and always return false, because circular dependencies are now correctly managed.
+     * \deprecated since QGIS 3.10
      */
     Q_DECL_DEPRECATED bool hasDependencyCycle( const QSet<QgsMapLayerDependency> & ) const {return false;}
 
--- a/src/core/qgsmaplayer.h
+++ b/src/core/qgsmaplayer.h
@@ -1499,9 +1499,6 @@ class CORE_EXPORT QgsMapLayer : public QObject
     //! List of layers that may modify this layer on modification
     QSet<QgsMapLayerDependency> mDependencies;
 
-    //! Checks whether a new set of dependencies will introduce a cycle
-    bool hasDependencyCycle( const QSet<QgsMapLayerDependency> &layers ) const;
-
     bool mIsRefreshOnNofifyEnabled = false;
     QString mRefreshOnNofifyMessage;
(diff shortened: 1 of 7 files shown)


mhugo (approved):
Good for me. It seems to suppress more code than it adds, even better :)

src/core/qgsmaplayer.h
```diff
@@ -1499,9 +1499,6 @@ class CORE_EXPORT QgsMapLayer : public QObject
     //! List of layers that may modify this layer on modification
     QSet<QgsMapLayerDependency> mDependencies;
 
-    //! Checks whether a new set of dependencies will introduce a cycle
-    bool hasDependencyCycle( const QSet<QgsMapLayerDependency> &layers ) const;
```
nyalldawson: Instead of removing this and breaking API, I'd suggest leaving it, marking it as deprecated, and always returning false

troopa81: OK 8983695

troopa81:
Please don't stale!

haubourg:
just tested in production environment and it works nicely! 
fixed_snapping_index_out_of_sync_for_transactiongroup_and_dbtriggers

@nyalldawson or @m-kuhn  Can someone merge this. I think having this in LTR will be a great relief for many application using database side intelligence :)

Of course this slows the digitizing work, but #31374  should help.

haubourg:
Note for documentation team (note for myself), this feature is barely visible in the UI. 
It only adds one line in the layer depency dialog.

Chaz6:
Thanks for this improvement! I would really like to read a blog post about using this feature and its applications (i.e. what kind of editing tasks with some simple sql examples).

troopa81:
@Chaz6 Here is a [blog post](https://oslandia.com/en/2020/01/10/en-qgis-snapping-improvements/) where we talk (among other things) about data depedencies

PR #9648, opened 2019-03-28
Symbol aware legend expression

roya0045:
## Description
Supersede #8911 

ext

The gist of the PR is to allow the user to be able to set up expressions in the legend, yes such a thing is already doable since a recent PR from Mr.Dawson, but this PR seeks to allow greater usability.

Every legend node is evaluated independently. The expression has access to the layer scope, the legend scope, the map scope and the symbol scope.

When using aggregation operations, the symbol scope will be used automatically to filter out the features to use and provide the statistics for the features that are part of the current symbol.

The GUI should allow the user to set the expression of the layertreelayer. The current labels would be accessible as an expression and so the end user could insert an expression anywhere in relation to the label text of the layer/symbol.

In order to ensure that the counting is done when evaluating the expression, a refresh of the layout legend will be triggered once all counting tasks are completed. In order to do this some signals slots and a list have been implemented in both the vector layer class and the item layout legend class to keep track of all these tasks and force a refresh only when everything is counted.

I have also added more verbose sip error output for debuging ( my sipify script generated bad characters, and I think this verbosity is necessary to match the verbosity of other errors).

## Checklist

> Reviewing is a process done by project maintainers, mostly on a volunteer basis. We try to keep the overhead as small as possible and appreciate if you help us to do so by completing the following items. Feel free to ask in a comment if you have troubles with any of them.

- [x] Commit messages are descriptive and explain the rationale for changes
- [x] Commits which fix bugs include `fixes #11111` in the commit message next to the description
- [x] Commits which add new features are tagged with `[FEATURE]` in the commit message
- [x] Commits which change the UI or existing user workflows are tagged with `[needs-docs]` in the commit message and contain sufficient information in the commit message to be documented
- [x] I have read the [QGIS Coding Standards](https://docs.qgis.org/testing/en/docs/developers_guide/codingstandards.html) and this PR complies with them
- [x] This PR passes all existing unit tests (test results will be reported by travis-ci after opening this PR)
- [x] New unit tests have been added for core changes
- [x] I have run [the `scripts/prepare-commit.sh` script](https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTING.md#contributing-to-qgis) before each commit

PR #30950, opened 2019-07-26
[API?] QgsVectorLayer, rely on the counter instead of storing the results

roya0045:
## Description
This is a change that is in branch #30727 initially but that I'm branching off.

As proposed initially by Matthias, this changes aim to rely on the counter directly rather than storing its results in the qgsvectorlayer.

This changes remove two useless signals that would now be useless.

This also means that the feature counter will be usable instead of being deleted after use. Thus enabling the retrieval of things like the fids without having to store it in the layer and without having to always empty and maintain the previously stored hash.

I hope this change is not controversial. Also please squash, as usual my branches have a lot of extra crap that are already implemented.

## Checklist


roya0045:
Test failure seems unrelated.

nyalldawson:
> This also means that the feature counter will be usable instead of being deleted after use. 

That's NOT the case - task ownership is transferred to the task manager, and they will be deleted after they complete by the manager.

@m-kuhn can you elaborate on what you were proposing?

roya0045:
@nyalldawson For reference, this PR came [from this exchange](https://github.com/qgis/QGIS/pull/9648#discussion_r296443651), it is highly possible that I have misunderstood something.

roya0045:
> That's NOT the case - task ownership is transferred to the task manager, and they will be deleted after they complete by the manager.

Out of curiosity, if they were deleted by the manager, what was the goal of [those two functions?](https://github.com/qgis/QGIS/pull/30950/files#diff-106be592650b55ccae79c83eecd15cebL4474)

nyalldawson:
To avoid dangling pointers I assume.

roya0045:
But in all fairness, I have been using the counter in my other PR and everything works fine. All the tests as ok and the fids can be fetched many time without segfault, as demonstrated by the tests.

nyalldawson:
That's the nature of thread safety issues though - you won't see a crash, until you do 😆

m-kuhn:
My proposal was to create a separate new counter in the layout legend instead of relying on the cached results from the layer and adding more cached data (fids).

The idea is that a layout may have a different symbology than the "main" layer.

Main thing was not to "pollute" the layer api if not for a good reason. And to keep things modular.

The opinion is not very strong. Modifying vector layer for fid per symbol cache can be discussed too.

roya0045:
@m-kuhn when you talk about a counter in the layout, do you mean a counter to hold the result or an actual counter that would count things?

I think that doing layout specific counters will be hard to implement/access and may lead to higher memory use.

By using the current counter I rely on the existing counting procedure and this does not add any loop. It simply need to be ran once.

Also when you state that a layout may have distinct symbology that the layer. Here I'm not too sure if you are actually talking about the source layer or the QgsVectorLayer. I have always understood that the layout will use a QgsVectorLayer symbology and that it could not 'warp' or change that symbology. 

Personally the way I do things is needed is to duplicate the vector layer and change its symbology. In theory this duplicated would have a distinct counter since they are tied to the QgsVectorLayer and not the source itself. This is why I always thought that the current way ( relying on the current counter) was optimal. 

I also think it would be strange to get the feature count from one place and their id from another place, since this is pretty much the same task.

nyalldawson:
@roya0045 actually, through map themes and layer styles it is possible to have layout maps with entirely different symbology (per layer) in a layout.

roya0045:
> @roya0045 actually, through map themes and layer styles it is possible to have layout maps with entirely different symbology (per layer) in a layout.

From what I understood the themes are mostly a way to control which layer/symbol are displayed. But I must admit that I have not dealt with the layer styles, unless I'm thinking of something else.

roya0045:
@m-kuhn I have some question regarding the layout specific symbology.
1) Can the style manager aggregate or alter the reference value of a data-defined symbology?
(As long as the groups of fids are the same this should not be problematic as the id of the symbol is used to fetch the fids and not the order of the symbols)
2) assuming the style manager can completely change the symbology of a layer (entirely distinct categories/classes) between the canvas and the layer. How frequent is the use of the style manager for such operations?
3) Is it worth it to complexify things a lot to cover some fringe cases? (I have no clue where to implement the counter since the current counter need a QgsVectorLayer, this mean that a discinct QgsVectorLayer instance would exist for the canvas and one for the layout? Again I though the canvas fetched things from the layertree and I thought the qgslegendmodel(?) only copied the layertree in order to not alter the main tree.)

m-kuhn:
I would suggest you have a look at map themes and layer styles.

DelazJ:
@roya0045 for (non-dev) information on map theme, https://docs.qgis.org/3.4/en/docs/user_manual/introduction/general_tools.html#configuring-map-themes and https://docs.qgis.org/3.4/en/docs/user_manual/print_composer/composer_items/composer_map.html#layers

roya0045:
> @roya0045 for (non-dev) information on map theme, https://docs.qgis.org/3.4/en/docs/user_manual/introduction/general_tools.html#configuring-map-themes and https://docs.qgis.org/3.4/en/docs/user_manual/print_composer/composer_items/composer_map.html#layers

Thank I have used map themes extensively and I know that they do not alter the symbology of the vector layers, this is why I was confused by its mention. But its more about the capabilities of the style manager and how it can affect the categories of a layer and if it could changes the categories themselves and the features that it represents.

If no option actually changes the list of feature that is represented by a given symbol, I think that using the QgsVectorLayer is the best bet, otherwise I'll check elsewhere.

roya0045:
@m-kuhn I did some test (mostly with locking layers) and I can see what you mean. Even when it comes to the counter this part does not work.

This either means that we would need to either alter the counter extensively or redo a new counter.

From what I understand when it comes to locking or style overhaul at draw time, it is only the style of the features that is changed based on a xml /formatting information. This "temporary" change (since it does not affect the layer but only patches the data at paint-time) I'm not sure how this could be done. I assume that it might be possible to use the new formatting to pass the layer through, currently a new rendercontext and an expression context is used to that effect. I'm not sure how the style formatting could be used instead when provided.

This still leaves the issue of storage and invalidation, especially if many layout use distinct locked/overhauled symbology. 

The issue is that I have no clue how to use the xml style string to reclassify the features. @nyalldawson , are you familiar with the xml symbology string that are used in QgsMapLayerStyle

Oh and another point, doesn't this mean that those features have no symbol id either? Meaning that we would need to pass the whole list of fids in the input of the expression (as the fids is not valid). 

This seems to be too much of a hassle when the user can just duplicate the vectorlayer, make themes with the proper symbology (the proper layer) and this would not break anything as they would be distinct QgsVectorLayer with the proper fids and a matching symbology.

roya0045:
@m-kuhn @nyalldawson considering the previous two comments, would you be ok if I did as I originally intended to do, I.E.: store them in the QgsVectorLayer and maintain the current behavior in regards to the counter and its related signals?

I think that covering the option of overwritten symbology when drawing the map layout it not worth it right now but could be a good pr in the future if someone wants to fund that feature. Otherwise since this functionality is not even remotely present, I do not personally wish to tackle this ordeal considering the previously mentioned elements.

m-kuhn:
@roya0045 this is functionality which is present already for a long time, not something that needs funding in the future.

See "Layers->Follow Map Themes" and "Lock styles for layers"

image

roya0045:
> @roya0045 this is functionality which is present already for a long time, not something that needs funding in the future.
> 
> See "Layers->Follow Map Themes" and "Lock styles for layers"
> 
> image

Sorry for the confusion, I was not talking about the layer lock and themes, I am well aware of their existence and already use them,

What I am talking about is having a counter that can handle locked style.

If you have a layer with a classified symbology, freeze it in the layout, change the classes/source column of the symbology and then do the count in the layout. You will see that only the count of the QgsVectorlayer in the QgsLayerTree will be displayed and not the count of the frozen layer.

This being due to the fact that the layer is passed to the layout map as is and then the symbology is altered with the xml information stored in the QgsMapLayerStyle.

Since that symbology is only applied to override the current symbology and represent the layer in its frozen state, this representation has no attribute and it is not possible to get a proper count or list of fids for that frozen symbology since there is no instance of that symbology.

This lack of instance is why I don't want to make a counter that can handle those cases and only focus on the qgsvectorlayer.

Is that more clear?

m-kuhn:
The main reason for the request originally was that it's altering the QgsVectorLayer API and adds a cache with fids for a very particular use case.

If that's not required, I don't care that much. If it is required, I'd prefer to have other opinions on the topic.

roya0045:
> The main reason for the request originally was that it's altering the QgsVectorLayer API and adds a cache with fids for a very particular use case.
> 
> If that's not required, I don't care that much. If it is required, I'd prefer to have other opinions on the topic.

It is required as caching leads to  thread safety issues. I will make a PR with the caching in the vectorlayer as a way to get more feedback on that particular approach.

PR #30952, opened 2019-07-26
Allow changing the default ellipsoid to planimetric

m-kuhn:
The overall goal for this PR is to enable a user to have new projects defaulting to his favourite CRS while keeping measurements planimetric.
Is this ok where it is or should it go to the measurement settings?

Screenshot from 2019-07-30 11-52-34

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -5515,7 +5515,7 @@ bool QgisApp::fileNew( bool promptToSaveFlag, bool forceBlank )
   const QgsCoordinateReferenceSystem srs = QgsCoordinateReferenceSystem( settings.value( QStringLiteral( "/projections/defaultProjectCrs" ), GEO_EPSG_CRS_AUTHID, QgsSettings::App ).toString() );
   // write the projections _proj string_ to project settings
   const bool planimetric = settings.value( QStringLiteral( "/qgis/measure/planimetric" ), true ).toBool();
-  prj->setCrs( srs, !planimetric ); // If the default ellipsoid is not planimetric, set it from the defualt crs
+  prj->setCrs( srs, !planimetric ); // If the default ellipsoid is not planimetric, set it from the default crs
   if ( planimetric )
     prj->setEllipsoid( GEO_NONE );
 
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -5515,7 +5515,7 @@ bool QgisApp::fileNew( bool promptToSaveFlag, bool forceBlank )
   const QgsCoordinateReferenceSystem srs = QgsCoordinateReferenceSystem( settings.value( QStringLiteral( "/projections/defaultProjectCrs" ), GEO_EPSG_CRS_AUTHID, QgsSettings::App ).toString() );
   // write the projections _proj string_ to project settings
   const bool planimetric = settings.value( QStringLiteral( "/qgis/measure/planimetric" ), true ).toBool();
-  prj->setCrs( srs );
+  prj->setCrs( srs, !planimetric ); // If the default ellipsoid is not planimetric, set it from the defualt crs
   if ( planimetric )
     prj->setEllipsoid( GEO_NONE );
 
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -5514,7 +5514,7 @@ bool QgisApp::fileNew( bool promptToSaveFlag, bool forceBlank )
   // set project CRS
   const QgsCoordinateReferenceSystem srs = QgsCoordinateReferenceSystem( settings.value( QStringLiteral( "/projections/defaultProjectCrs" ), GEO_EPSG_CRS_AUTHID, QgsSettings::App ).toString() );
   // write the projections _proj string_ to project settings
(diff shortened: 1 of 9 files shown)


src/app/qgsoptions.cpp
```diff
@@ -83,6 +83,9 @@
 
 #include "qgsconfig.h"
 
+// TODO Move to generic place
```
m-kuhn: Note to self: if you add a todo, fix it before you hit the merge button.

nyalldawson:
Can you confirm that this is overwritten as soon as the project crs is changed?

m-kuhn:
Yes.
It's only to control the default value set for new projects.
There could also be an explicit "follow crs" option on project level that could be added but that's not on our requirements right now.

nyalldawson:
Ok.

Don't hate me, but I feel like the mega-thread which starts at https://lists.osgeo.org/pipermail/proj/2019-March/008333.html is required reading before we touch this option. Sorry :grimacing:

m-kuhn:
What's the main point in there?
Acronym not being a good unique id?

nyalldawson:
Ok - if that's the consideration, why not swap the option to "always use planimetric measurements by for new projects" (or similar?)

nyalldawson:
To expand: I'm concerned that exposing this as a choice of ellipsoid may make the situation worse. If all you are wanting is an option to disable ellipsoidal measurements by default, then I'd prefer that to be the option. It completely avoids adding extra complexity regarding the (fragile) code which syncs ellipsoid to project crs.

m-kuhn:
To make this a boolean option works for me.

But I fail to see how the current proposal makes things worse, can you elaborate?

nyalldawson:
> But I fail to see how the current proposal makes things worse, can you elaborate?

Because it conveys the message that you can set a fixed ellipsoid which will always be respected. But this isn't the case - if you change the project crs, it will be reset immediately....

m-kuhn:
... which would be fixed by having a "follow CRS (current behavior)" vs. "use a specific ellipsoid" option. Do you agree?

nyalldawson:
> vs. "use a specific ellipsoid" option. 

But this is exactly the option we need to avoid -- it would allow users to set a default ellipsoid and then it would be used regardless of the project's CRS, which is exactly in contrast to the recommendations from that proj thread...

m-kuhn:
@nyalldawson adjusted according to our discussion

src/app/qgisapp.cpp
```diff
@@ -5515,7 +5515,7 @@ bool QgisApp::fileNew( bool promptToSaveFlag, bool forceBlank )
   const QgsCoordinateReferenceSystem srs = QgsCoordinateReferenceSystem( settings.value( QStringLiteral( "/projections/defaultProjectCrs" ), GEO_EPSG_CRS_AUTHID, QgsSettings::App ).toString() );
   // write the projections _proj string_ to project settings
   const bool planimetric = settings.value( QStringLiteral( "/qgis/measure/planimetric" ), true ).toBool();
-  prj->setCrs( srs );
+  prj->setCrs( srs, !planimetric ); // If the default ellipsoid is not planimetric, set it from the defualt crs
```
nyalldawson: default

src/app/qgsoptions.cpp
```diff
@@ -556,6 +556,7 @@ QgsOptions::QgsOptions( QWidget *parent, Qt::WindowFlags fl, const QList<QgsOpti
   {
     mKeepBaseUnitCheckBox->setChecked( false );
   }
+  mPlanimetricMeasurementsComboBox->setChecked( mSettings->value( QStringLiteral( "/qgis/measure/planimetric" ), false ).toBool() );
```
nyalldawson: QgsSettings::Core ?

src/gui/layertree/qgslayertreemapcanvasbridge.cpp
```diff
@@ -115,8 +115,12 @@ void QgsLayerTreeMapCanvasBridge::setCanvasLayers()
     switch ( projectCrsBehavior )
     {
       case QgsGui::UseCrsOfFirstLayerAdded:
-        QgsProject::instance()->setCrs( mFirstCRS, true );
+      {
+        const bool planimetric = QgsSettings().value( QStringLiteral( "/qgis/measure/planimetric" ), true ).toBool();
```
nyalldawson: Note to self - we should consider a "QgsProjectDefaultSettings" class or the like, to avoid scattered code like this...

nyalldawson (commented):
Looks good to me! Can you also add this to the .ini settings template?

src/gui/layertree/qgslayertreemapcanvasbridge.cpp
m-kuhn: Wouldn't that be covered by the grant proposal for settings by @3nids? (Can't find it right now)

nyalldawson: It wasn't successful though, was it?

m-kuhn: No ...

src/app/qgsoptions.cpp
m-kuhn: Done. What does that actually do?

PR #30955, opened 2019-07-27
Allow data defined control over "label every part of multipart features" setting

nyalldawson:
This was  one of the few remaining labeling settings which could not be data defined, yet it's a useful setting to have per-feature control over.


nirvn (approved):
Nice

nirvn:
@nyalldawson , it'd be useful to have that exposed in the change label tool too.

PR #30960, opened 2019-07-28
[needs-docs] Drop setting to control label solution method from project labeling settings

nyalldawson:
This is an unnecessary setting, resulting in 1000's of extra, very complex
and untested lines of code. Furthermore, the differences are almost non-existant
and it's extremely unlikely end users would (or should) need to change this
setting.


giohappy:
Just for historical reference these algos come from the original PAL library. Some papers about POPMUSIC and the others can be found here: http://pal.heig-vd.ch/index.php?page=support

wonder-sk (approved):

src/core/qgslabelingenginesettings.h
```diff
@@ -74,10 +77,17 @@ class CORE_EXPORT QgsLabelingEngineSettings
     //! Sets number of candidate positions that will be generated for each label feature
     void setNumCandidatePositions( int candPoint, int candLine, int candPolygon ) { mCandPoint = candPoint; mCandLine = candLine; mCandPolygon = candPolygon; }
 
-    //! Sets which search method to use for removal collisions between labels
-    void setSearchMethod( Search s ) { mSearchMethod = s; }
-    //! Which search method to use for removal collisions between labels
-    Search searchMethod() const { return mSearchMethod; }
+    /**
+     * Sets which search method to use for removal collisions between labels
```
m-kuhn:
```suggestion
     * Used to set which search method to use for removal collisions between labels
```

m-kuhn:
Good move

PR #30966, opened 2019-07-29
Expose "label all parts" option via label properties tool

nyalldawson:
Makes it easier to set this option per-feature


nyalldawson:
@nirvn as requested in the earlier PR

nirvn (approved):
Thanks for addressing this.

PR #30972, opened 2019-07-29
Fix virtual fields that use functions requiring all attributes

3nids:
the requested fields to evaluate the virtual field where not fetched if only the virtual field was requested at the request creation

### DIFF
--- a/src/core/qgsvectorlayerfeatureiterator.cpp
+++ b/src/core/qgsvectorlayerfeatureiterator.cpp
@@ -286,6 +286,13 @@ QgsVectorLayerFeatureIterator::~QgsVectorLayerFeatureIterator()
   close();
 }
 
+/// @cond private
+
+/**
+ * This class guards against infinite recursion.
+ * The counter will be created per thread and hasStackOverflow will return
+ * true if more than maxDepth instances are created in parallel in a thread.
+ */
 class QgsThreadStackOverflowGuard
 {
   public:
@@ -326,6 +333,7 @@ class QgsThreadStackOverflowGuard
     int mMaxDepth;
 };
 
+/// @endcond private
 
 bool QgsVectorLayerFeatureIterator::fetchFeature( QgsFeature &f )
 {
@@ -340,8 +348,6 @@ bool QgsVectorLayerFeatureIterator::fetchFeature( QgsFeature &f )
 
   if ( guard.hasStackOverflow() )
     return false;
-
-  int depth = guard.depth();
 
   if ( mRequest.filterType() == QgsFeatureRequest::FilterFid )
   {
@@ -680,9 +686,6 @@ void QgsVectorLayerFeatureIterator::prepareExpression( int fieldIdx )
 
   if ( guard.hasStackOverflow() )
     return;
-
-
-  int depth = guard.depth();
 
   const QList<QgsExpressionFieldBuffer::ExpressionField> &exps = mSource->mExpressionFieldBuffer->expressions();
 
--- a/src/core/qgsvectorlayerfeatureiterator.cpp
+++ b/src/core/qgsvectorlayerfeatureiterator.cpp
@@ -296,40 +296,53 @@ QgsVectorLayerFeatureIterator::~QgsVectorLayerFeatureIterator()
 class QgsThreadStackOverflowGuard
 {
   public:
-    QgsThreadStackOverflowGuard( QThreadStorage<int> &storage, int maxDepth )
+
+    QgsThreadStackOverflowGuard( QThreadStorage<QStack<QString>> &storage, const QString &stackFrameInformation, int maxDepth )
       : mStorage( storage )
       , mMaxDepth( maxDepth )
     {
       if ( !storage.hasLocalData() )
       {
-        storage.setLocalData( 0 );
+        storage.setLocalData( QStack<QString>() );
       }
-      else
-      {
-        storage.setLocalData( storage.localData() + 1 );
-      }
+
+      storage.localData().push( stackFrameInformation );
     }
 
     ~QgsThreadStackOverflowGuard()
     {
-      mStorage.setLocalData( mStorage.localData() - 1 );
+      mStorage.localData().pop();
     }
 
     bool hasStackOverflow() const
     {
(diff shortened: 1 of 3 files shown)


troopa81 (approved):
Looks good to me. Thanks!

m-kuhn:
I'd love to have a review by someone here.

Short problem statement.
A virtual fields obviously cannot depend on "itself", this will lead to an infinite recursion.
So we detected so far always if it's itself in "referencedFields()", however, if it's definition contains a method that requires all attributes (like `eval` or others), it would also report itself, be detected as recursive and return NULL always.

So this PR updates the code to not check for recursive dependencies (because there is no way we can always check that on prepare time, it's simply impossible).
And instead add a thread local recursion guard which is triggered when we have more than 255 nested feature iterators open.

troopa81:
> if it's definition contains a method that requires all attributes (like eval or others), it would also report itself, be detected as recursive and return NULL always.

I cannot reproduce what you're describing here with current master. I try it with a simple layer containing two fields `a`, and `b`, and try to create a virtual field `c` with the expression `eval('a+b')` and it works. 

What is your issue use case? Is it when you have reference to other attribute layer (with get_feature for instance)?

m-kuhn:
Try `eval('a')`

m-kuhn:
It will already work, if the required attributes are requested for some reason (e.g. in the attribute table, the feature will be requested fully populated, and the expression succeeds.

Try running the following code on your sample layer:

```
request = QgsFeatureRequest()
request.setSubsetOfAttributes(['c'], layer.fields())
feature = next(layer.getFeatures(request))
feature['c']
```

troopa81:
OK, I think I understand, thanks. 

Wouldn't be possible to keep track of the "computed field call stack" in the thread storage to detect circular reference, instead of just keeping an int and stopping at 255 ?

We could push the current evaluated fieldIndex in a QStack in the thread storage. Before preparing expression, we check that it's not already in the stack. It should work with case like "test2 = sum(test2)"

m-kuhn:
This would need to be a combination of layer and field (index or name). This is probably a good idea for debugging purpose.

m-kuhn:
Actually, it's not even related to the field any more. It's just checking that we do not have more than max stack depth (255) features being fetched nested. We have no handle there as to what (or which field) caused this.

I was thinking about throwing an exception, but suddenly have "fetchFeature" throwing an exception seems risky.

troopa81:
> This would need to be a combination of layer and field (index or name)

Yes, indeed

> Actually, it's not even related to the field any more. We have no handle there as to what (or which field) caused this.

Why is that ? As we evaluate complex expression, we always go through `prepareField` method so we know the field and layer evaluated for the current stack. No?

m-kuhn:
See https://github.com/qgis/QGIS/pull/30972#issuecomment-531386574

> So this PR updates the code to not check for recursive dependencies (because there is **no way we can always check that on prepare time**, it's simply impossible).

With the latest commit, most recursions will be detected at evaluation time, in which case it will report the layers. If the recursion is already detected at prepare time, it will report the field as well. Does this look good to you?

troopa81:
I give it a try to better understand the issue. 

I think the guard in prepareExpression is never used. The localStorage is always empty, because prepareExpression always finish (and pop locatStorage data) without calling itself recursively directly or indirectly. And so, as you said, it's not possible to do any check on prepare time.

The fix doesn't work because I fail to acquire ogr connection in QgsConnectionPool while calling recursively fetchFeature.

I was thinking about [this](https://github.com/troopa81/QGIS/blob/test_fix_recursive_expression/src/core/qgsvectorlayerfeatureiterator.cpp#L436) to solve the issue and prevent the use of the number 255. It seems to work, but I tested it poorly. It raise an error message with expression like with `test2=sum(test2)` and `test2=eval('sum(test2)')`.

Hope it helps.

m-kuhn:
> I think the guard in prepareExpression is never used. The localStorage is always empty, because prepareExpression always finish (and pop locatStorage data) without calling itself recursively directly or indirectly. And so, as you said, it's not possible to do any check on prepare time.

Wouldn't it be possible that the nested prepare expressions prepares yet another iterator, recursively?

https://github.com/qgis/QGIS/pull/30972/files#diff-d2d2d088f36a3991d43cc9b57101bf5fR723

> The fix doesn't work because I fail to acquire ogr connection in QgsConnectionPool while calling recursively fetchFeature.

Can you elaborate on what's wrong here?

My idea was
 * to be more generic than just virtual fields and catch *any* infinite recursion (but right now I fail to come up with a scenario apart from virtual fields, so that might be a non issue)
 * to not disable recursion completely, and just block infinite recursion. Just like python etc do:

```
>>> def a():
>>>     return a()
>>> 
>>> a()

  File "<stdin>", line 1, in <module>
  File "<stdin>", line 2, in a
  File "<stdin>", line 2, in a
  File "<stdin>", line 2, in a
  [Previous line repeated 996 more times]
RecursionError: maximum recursion depth exceeded
```

troopa81:
> Wouldn't it be possible that the nested prepare expressions prepares yet another iterator, recursively?

I would say no. I take a quick look and it doesn't seem that any `prepare` expression node method use QgsVectorLayerIterator. And it will be odd that you call an iterator while you are preparing an expression. But, I can't be sure, maybe it's better to check.

> Can you elaborate on what's wrong here?

I made a test with a GeoPackage file and the condition `test2=sum(test2)` and it stops [here](https://github.com/troopa81/QGIS/blob/test_fix_recursive_expression/src/core/qgsconnectionpool.h#L110). 

My guess is that every time a QgsVectorLayerIterator is created (recursively) it tries to acquire an OGR connection to the file. But the connection is limited and so the acquire method waits until a connection will be released which won't happen because we are in the recursive loop. The guard limit (255) must be below the connection limit and won't break the recursive loop.

> My idea was  to be more generic

Yes, why not. It has the advantage to catch all scenarios even the one I can't imagine now. On the other hand, if it's less generic, we are able to stop the recursion sooner and display a more specific message ("There is a recursion on test2 field" for instance). 

I'm ok with both ways, just need to deal with the blocking acquire connection issue if you pick the generic ways.

m-kuhn:
Thanks everyone for the help and reviews

PR #30977, opened 2019-07-29
Fixes #28966 - disallow save project in postgres when < 9.5

### LINKED ISSUE #28966: Save project to PostGreSQL doesn't work with PostGreSQL 9.4 and older
Author Name: **Michel Stuyts** (@mstuyts)
Original Redmine Issue: [21148](https://issues.qgis.org/issues/21148)
Affected QGIS version: 3.4.4
Redmine category:project_loading/saving

---
The functionality to save a project to PostgreSQL will only work with PostgreSQL versions 9.5 and newer, because:
* the datatype JSONB was introduced in PostgreSQL 9.4
* the ON CONFLICT clause was introduced in  PostgreSQL 9.5

Problem:
* It's possible to check "Allow saving/loading QGIS projects in the database" in the Database Connection Settings for PostgreSQL databases older than version 9.5.
* When this checkbox is checked, it is possible to select Save To -> PostgreSQL in the Menu. After selecting a database, a schema and a project it will show an error that doesn't show the problem is the PostgreSQL version that is being used.

Expected behavior:
* It's not possible to check the checkbox "Allow saving/loading QGIS projects in the database" for PostgreSQL databases older than version 9.5. 
* Somewhere should be mentioned that this functionality only works with version 9.5 and newer.

Ideal behavior:
* Saving projects to PostgreSQL also works in older versions of PostgreSQL

stev-0:
## Description
Disables save project in database where Postgres database is version 9.4 or less. Relies on testing connection to determine postgres version. 

## Checklist

### DIFF
--- a/src/providers/postgres/qgspgnewconnection.cpp
+++ b/src/providers/postgres/qgspgnewconnection.cpp
@@ -208,9 +208,8 @@ void QgsPgNewConnection::testConnection()
 
   if ( conn->pgVersion() < 90500 )
   {
-    cb_projectsInDatabase->setChecked( false );
     cb_projectsInDatabase->setEnabled( false );
-    cb_projectsInDatabase->setToolTip( QStringLiteral( "Saving projects in databases not available for Postgres databases earlier than 9.5" ) );
+    cb_projectsInDatabase->setToolTip( tr( "Saving projects in databases not available for Postgres databases earlier than 9.5" ) );
   }
   else
   {
--- a/src/providers/postgres/qgspgnewconnection.cpp
+++ b/src/providers/postgres/qgspgnewconnection.cpp
@@ -115,6 +115,7 @@ void QgsPgNewConnection::accept()
   QString baseKey = QStringLiteral( "/PostgreSQL/connections/" );
   settings.setValue( baseKey + "selected", txtName->text() );
   bool hasAuthConfigID = !mAuthSettings->configId().isEmpty();
+  testConnection();
 
   if ( !hasAuthConfigID && mAuthSettings->storePasswordIsChecked( ) &&
        QMessageBox::question( this,
@@ -204,6 +205,18 @@ void QgsPgNewConnection::testConnection()
   }
 
   QgsPostgresConn *conn = QgsPostgresConn::connectDb( uri.connectionInfo( false ), true );
+
+  if ( conn->postgresVersion() < 95000 )
+  {
+    cb_projectsInDatabase->setChecked( false );
+    cb_projectsInDatabase->setEnabled( false );
+    cb_projectsInDatabase->setToolTip( QStringLiteral( "Saving projects in databases not available for Postgres databases earlier than 9.5" ) );
+  }
+  else
+  {
+    cb_projectsInDatabase->setEnabled( true );
+    cb_projectsInDatabase->setToolTip( QStringLiteral( "" ) );
+  }
 
   if ( conn )
   {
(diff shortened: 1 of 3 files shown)


src/providers/postgres/qgspgnewconnection.cpp
```diff
@@ -205,6 +206,18 @@ void QgsPgNewConnection::testConnection()
...
+  if ( conn->postgresVersion() < 95000 )
+  {
+    cb_projectsInDatabase->setChecked( false );
+    cb_projectsInDatabase->setEnabled( false );
+    cb_projectsInDatabase->setToolTip( QStringLiteral( "Saving projects in databases not available for Postgres databases earlier than 9.5" ) );
+  }
+  else
+  {
+    cb_projectsInDatabase->setEnabled( true );
+    cb_projectsInDatabase->setToolTip( QStringLiteral( "" ) );
```
nyalldawson:
```suggestion
    cb_projectsInDatabase->setToolTip( QString() );
```

jef-n:
Why does that need to be disabled?

stev-0:
See https://github.com/qgis/QGIS/issues/28966

src/providers/postgres/qgspostgresconn.cpp
```diff
@@ -871,6 +871,11 @@ bool QgsPostgresConn::hasPointcloud()
   return mPointcloudAvailable;
 }
 
+int QgsPostgresConn::postgresVersion()
```
jef-n: duplicates QgsPostgresConn::pgVersion()

src/providers/postgres/qgspgnewconnection.cpp
```diff
@@ -205,6 +206,18 @@ void QgsPgNewConnection::testConnection()
 
   QgsPostgresConn *conn = QgsPostgresConn::connectDb( uri.connectionInfo( false ), true );
 
+  if ( conn->pgVersion() < 90500 )
+  {
+    cb_projectsInDatabase->setChecked( false );
+    cb_projectsInDatabase->setEnabled( false );
+    cb_projectsInDatabase->setToolTip( QStringLiteral( "Saving projects in databases not available for Postgres databases earlier than 9.5" ) );
```
m-kuhn:
```suggestion
    cb_projectsInDatabase->setToolTip( tr( "Saving projects in databases not available for Postgres databases earlier than 9.5" ) );
```

src/providers/postgres/qgspgnewconnection.cpp
```diff
@@ -205,6 +206,18 @@ void QgsPgNewConnection::testConnection()
 
   QgsPostgresConn *conn = QgsPostgresConn::connectDb( uri.connectionInfo( false ), true );
 
+  if ( conn->pgVersion() < 90500 )
+  {
+    cb_projectsInDatabase->setChecked( false );
```
m-kuhn: What's this for?

gioman:
@stev-0 
if the goal of this patch was to disable saving projects within PostgreSQL DBs if postgresql is < 9.5 then it does not work on both LTR and 3.12. Users are still able to try save the project and get an error message in return.

stev-0:
@gioman possibly this is an old connection that was previously setup? Any new connection that is created should disallow save as project, but this doesn't work retrospectively, if I remember rightly because it would be expensive to check for this apart from at the moment the connection is first created, but there may be another way to deal with this, if this is indeed the issue.

gioman:
> possibly this is an old connection that was previously setup?

@stev-0 yes, but it happens also with newly created connections.

PR #30980, opened 2019-07-30
Port cmake fixes from kadas

nyalldawson:
Cross-platform compilation fixes ported from downstream kadas project

@elpaso I know at one stage you were actively using ming to cross compile - could you take a look over these?


manisandro:
@nyalldawson Thanks for jumping ahead and upstreaming these already. Please note that we do intend to upstream all QGIS changes by the end of the kadas 2.0 development cycle the latest, so that kadas becomes a separate application which is uses stock qgis libraries. As far as cross-compiling is concerned, this is currently only tested with Fedoras MinGW environment - but it works very well there.

elpaso:
@nyalldawson I'm still using mingw but only through MXE (https://mxe.cc/ cross-compile), I've a working recipe here https://github.com/qgis/QGIS/tree/master/ms-windows/mxe#mxe-cross-build , I use it occasionally, the only problem is that it does not support Python.
You can even run the generated QGIS on wine.

nyalldawson:
Thanks for the confirmation @manisandro  (and I've got to say -- the new approach to implementing Kadas looks great. Nice work!).

I've dropped the mingw build script from this PR for now as it needed non-trivial fixes to pass the shellcheck tests, but the rest of the cmake changes look good and would be valuable to have in master now.

PR #30981, opened 2019-07-30
Inizialize QgsGpsInformation with sane values…

nyalldawson:
…, add QgsGpsInformation::isValid and QgsGpsInformation::fixStatus

Ported from kadas

### DIFF
--- a/src/core/gps/qgsgpsconnection.cpp
+++ b/src/core/gps/qgsgpsconnection.cpp
@@ -25,6 +25,48 @@
 
 #include "qgsnmeaconnection.h"
 #include "qgslogger.h"
+#include "info.h"
+
+
+bool QgsGpsInformation::isValid() const
+{
+  bool valid = false;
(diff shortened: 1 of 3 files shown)


src/core/gps/qgsgpsconnection.cpp
```diff
@@ -25,6 +25,48 @@
...
 #include "qgslogger.h"
+#include "info.h"
+
+
+bool QgsGpsInformation::isValid() const
+{
+  bool valid = false;
+  if ( status == 'V' || fixType == NMEA_FIX_BAD || quality == 0 ) // some sources say that 'V' indicates position fix, but is below acceptable quality
+  {
+    return false;
```
m-kuhn: Can we either have a local return value variable and a single return statement in the end or directly return the value everywhere?
The current code is very confusing to read.

```suggestion
    valid = false;
```

m-kuhn:
Looks good apart from the one small comment

src/core/gps/qgsgpsconnection.cpp
nyalldawson: Fixed

PR #30995, opened 2019-07-30
Do not specify srid in dbmanager. It goes crazy

### LINKED ISSUE #30787: DB manager cannot load anymore a PostGIS layer that is not registered in the geometry_columns table
In DB Manager run a "create table" query without doing the casting for the geometry, i.e.

CREATE TABLE test as
  SELECT 
row_number() over () as gid,
    a.gid AS id,  
    b.gid AS id2,
    ST_MakeLine(a.geom, b.geom) AS geom
  FROM points1 a, points2 b

In QGIS 2.18:

as there was no casting the table/layer is shown (just) with a "?" icon, but it *can* be loaded in the QGIS project.

Screenshot_20190718_100819

In QGIS 3 (tested master):

the table/layer is shown multiple times: with the "?" icon as also with the icons for points, lines and polygons (but of course the real geometry is only one of them). Moreover trying to add the layer in the project it returns an error.

Screenshot_20190718_101409

m-kuhn:
If srid is specified in an uri, it disables type detection.

Fix #30787

### DIFF
--- a/python/plugins/db_manager/db_plugins/plugin.py
+++ b/python/plugins/db_manager/db_plugins/plugin.py
@@ -985,7 +985,6 @@ class VectorTable(Table):
 
     def uri(self):
         uri = super().uri()
-        uri.setSrid(str(self.srid))
         for f in self.fields():
             if f.primaryKey:
                 uri.setKeyColumn(f.name)


rldhont (approved):

m-kuhn:
Superseded by #31047

PR #30473, opened 2019-06-28
[Processing][xyz tiles] Multi-threading tile generation with ThreadPoolExecutor

isghj5:
Modifies the XYZ Tiles function of the processing plugin to use all available cores on a machine to render tiles, speedup making use of local CPU hardware.

speed boost

This was achieved by taking most of code in the loop, relocating to a function and using concurrent.futures::ThreadPoolExecutor to pool all the tiles into jobs that are handled by threads in the TPE.

Introduces concurrent.futures as a dependency, but since it's standard python library it shouldn't introduce additional install requirement.

One known bug: ~~If you [Close] the UI before you [Cancel] a running job, the threads will not stop working. I'm not sure where the [Close] button code is to fix.~~ Nevermind, I didn't realize you could re-open the running task on the bottom bar, guess this isn't actually a bug.

Missing piece: I think the user should be able to set the thread count, but I'm not sure how to modify the user interface to add as an option

Aside from running on my machine(cheap desktop, Intel 4760k), I have not tested further. 

[TilesXYZ.py.txt](https://github.com/qgis/QGIS/files/3340832/TilesXYZ.py.txt)

PR #31012, opened 2019-07-31
[Processing][TilesXYZ] Multithreaded Tile generation with ThreadPoolExecutor

isghj5:
Rendering tiles is changed from single thread to multithread using python standard library ThreadPoolExecutor (TPE) class, resulting in a large performance boost for multithreaded CPUs.

Most of the looped code in TilesXYZAlgorithmBase::generate() was relocated into a separate function (per metatile), then all metatiles are mapped to this function by the TPE, which handles thread management for us. To prevent the thread conflicts on the settings, settings were duplicated per thread. 

~~There is one additional addition I think would be good to add: The user interface should have an option to set how many threads are created for the rendering job, right now the code uses all cores available to it(multithreading::cpu_count()) but I don't know how to add this to the user interface.~~ done

This is a re-attempt of a previous pull-request: https://github.com/qgis/QGIS/pull/30473

## Checklist

Not sure if this counts as a feature..? It's more a change than an addition I think.

This isn't core, right? It's processing, and I don't know how to make unit tests or what new tests would need to be created

### DIFF
--- a/python/plugins/processing/algs/qgis/TilesXYZ.py
+++ b/python/plugins/processing/algs/qgis/TilesXYZ.py
@@ -45,7 +45,8 @@ from qgis.core import (QgsProcessingException,
                        QgsCoordinateTransform,
                        QgsCoordinateReferenceSystem,
                        QgsMapRendererCustomPainterJob,
-                       QgsLabelingEngineSettings)
+                       QgsLabelingEngineSettings,
+                       QgsApplication)
 from processing.algs.qgis.QgisAlgorithm import QgisAlgorithm
 import threading
 from concurrent.futures import ThreadPoolExecutor
@@ -238,6 +239,8 @@ class TilesXYZAlgorithmBase(QgisAlgorithm):
         self.tile_format = self.formats[self.parameterAsEnum(parameters, self.TILE_FORMAT, context)]
         self.quality = self.parameterAsInt(parameters, self.QUALITY, context)
         self.metatilesize = self.parameterAsInt(parameters, self.METATILESIZE, context)
+        self.maxThreads = QgsApplication.maxThreads()
+        self.maxThreads = cpu_count() if self.maxThreads == -1 else self.maxThreads # if unbound, maxThreads() returns -1
         try:
             self.tile_width = self.parameterAsInt(parameters, self.TILE_WIDTH, context)
             self.tile_height = self.parameterAsInt(parameters, self.TILE_HEIGHT, context)
@@ -251,12 +254,8 @@ class TilesXYZAlgorithmBase(QgisAlgorithm):
         project = context.project()
         self.src_to_wgs = QgsCoordinateTransform(project.crs(), wgs_crs, context.transformContext())
         self.wgs_to_dest = QgsCoordinateTransform(wgs_crs, dest_crs, context.transformContext())
-
-        self.maxThreads = cpu_count() # from multithreading
-        #self.maxThreads = 2
         # without re-writing, we need a different settings for each thread to stop async errors
         # naming doesn't always line up, but the last number does
-        #self.settingsDictionary = {'ThreadPoolExecutor-o_' + str(i): QgsMapSettings() for i in range(self.maxThreads)}
         self.settingsDictionary = {str(i): QgsMapSettings() for i in range(self.maxThreads)}
         for thread in self.settingsDictionary:
             self.settingsDictionary[thread].setOutputImageFormat(QImage.Format_ARGB32_Premultiplied)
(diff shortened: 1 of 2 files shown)


saberraz:
I agree, having number of cores will be good. But, meanwhile, maybe you can use the number of core user has set to render maps?

isghj5:
> I agree, having number of cores will be good. But, meanwhile, maybe you can use the number of core user has set to render maps?

Assuming you mean Options->Rendering->Rendering Behavior->"Max Cores to use", I'm not sure yet how to get this value. I think it's in QGSServerSettings, however

`from qgis.sever import QGSServerSettings
...
maxThreads = QGSServerSettings().maxThreads()`
Gets me a ModuleNotFoundError: No module named 'qgis.server', even though several other files can find qgis.server, TilesXYZ.py cannot...? Not sure what's going on here, might be a quirk of my machine.

roya0045:
Shouldn't it be in the general settings and not exclusive to the server? I assume that users should also benefit from this as the settings are also located in the rendering/rendering behavior.

isghj5:
Found a function on QgsApplication that returns the core restriction.

I still think there should be an option for the user to set the core count just for a heavy job like rendering tiles on all cores, in case they want to lower cores for just that one task and not have to leave the window, go back to settings, change it, then revert after the task is over, but this is a better default.

python/plugins/processing/algs/qgis/TilesXYZ.py
```diff
@@ -180,102 +184,132 @@ def prepareAlgorithm(self, parameters, context, feedback):
...
+            # Haven't found a better way to break than to make all the new threads return instantly
+
+        if "Dummy" in threading.current_thread().name: # single thread testing
+            threadSpecificSettings = list(self.settingsDictionary.values())[0]
+        else:
+            try:
+                # guess we're not the only one using TPE in QGIS, cannot assume 0_#, it's sometimes 3 or 4_#
+                threadSpecificSettings = self.settingsDictionary[threading.current_thread().name[-1]] # last number only
+            except:
+                print("Exception! our threads don't match with our settings! ")
```
m-kuhn: Use the message log or other means for debug messages. Or if the exception really shouldn't happen and you want someone to debug that if it happens, just remove the try-except to raise the exception including stack trace for debugging reasons.

python/plugins/processing/algs/qgis/TilesXYZ.py
```diff
@@ -180,102 +184,132 @@ def prepareAlgorithm(self, parameters, context, feedback):
...
-                    tile_img = image.copy(tile_width * r, tile_height * c, tile_width, tile_height)
-                    writer.write_tile(tile, tile_img)
-
-                progress += 1
-                feedback.setProgress(100 * (progress / metatiles_count))
+        self.writer = writer
+
+        self.progressThreadLock = threading.Lock()
+        if self.maxThreads > 1:
+            feedback.pushConsoleInfo('Using %s CPU Threads:' % self.maxThreads)
```
m-kuhn:
```suggestion
            feedback.pushConsoleInfo(self.tr('Using {max_threads} CPU Threads:').format(max_threads=self.maxThreads))
```

python/plugins/processing/algs/qgis/TilesXYZ.py
```diff
@@ -180,102 +184,132 @@ def prepareAlgorithm(self, parameters, context, feedback):
...
+            feedback.pushConsoleInfo('Pushing all tiles at once: %s tiles.' % len(allMetatiles))
```
m-kuhn: As above, translate and use format

python/plugins/processing/algs/qgis/TilesXYZ.py
```diff
@@ -180,102 +184,132 @@ def prepareAlgorithm(self, parameters, context, feedback):
...
+            with ThreadPoolExecutor(max_workers=self.maxThreads) as threadPool:
+                threadPool.map(self.renderSingleMetatile, allMetatiles)
+        else:
+            feedback.pushConsoleInfo('Using 1 CPU Thread:')
```
m-kuhn: translate everywhere, actually :)

isghj5:
Am I correct in interpreting that the reason I cannot find a UI/Layout for this alg is because QgsAlgorithm creates one automatically from the parameter list? 

If this is true, and I wanted to make a new UI element for CPU core to use, where would be the best place to put it in the current layout?

If a layout file was created, could the layout be changed directly? I think it's a bit weird that each small item like zoom_min and zoom_max need to be on separate lines taking up more horizontal space than is really needed...

nyalldawson:
@lsghj5

Processing interfaces are automatically constructed, based on the parameters and output for the algorithm.

I'd suggest this new setting is instead exposed as a processing configuration option, i.e. is set from the Options - Processing tab, and is then usable for other algorithms too.

isghj5:
> I'd suggest this new setting is instead exposed as a processing configuration option, i.e. is set from the Options - Processing tab, and is then usable for other algorithms too.

In Options->Processing->General? Not sure how to add a UI element to set a max core count there, myself.

I still feel for a CPU heavy task like this the user might want to lower their threads for just one task and not have to go searching in a different menu for the option, changing it for one job then changing it back, but we can use the processing setting as a default when it gets added.

nyalldawson:
Well - I'm -1 about exposing an option for this in the algorithm itself. It's cluttering the algorithm's interface, making it more complex for users, and will need to be duplicated across many other algorithms in future.

Ideally, this would be added to QgsProcessingContext as a per-algorithm option, and then we'd have a button in the algorithm dialog to open an "algorithm settings" dialog, with settings like this, a map-extent filter option, options to override the default invalid geometry handling, set output CRSes, etc. 

But that's not a trivial change. I'd love to mentor it if you want to take it on though!

Otherwise, I'd rather a temporary band-aid solution (such as a QSettings key not exposed in the UI, or a Processing configuration key) vs an explicit algorithm parameter which we'll be forced to keep... forever.

isghj5:
> Otherwise, I'd rather a temporary band-aid solution (such as a QSettings key not exposed in the UI, or a Processing configuration key) vs an explicit algorithm parameter which we'll be forced to keep... forever.

Wait wait, I meant adding a parameter to TilesXYZAlgorithmBase, not to QgsAlgorithm. I just wrote a working parameter and it's < 15 lines of code that looks really easy to remove later if so desired.

Is there some project directive I missed saying no UI elements are every allowed to be removed after they are added, as part of design policy or something? I mean, removing UI elements isn't great design, but your suggested setting would replace it, only the UI element would be removed opening up space, the functionality would be replaced not removed completely.

nyalldawson:
> Wait wait, I meant adding a parameter to TilesXYZAlgorithmBase, not to QgsAlgorithm. I just wrote a working parameter and it's < 15 lines of code that looks really easy to remove later if so desired

Right - but that parameter effectively becomes part of stable API, and can never be removed in future when we get a proper way of handling this use case.

> Is there some project directive I missed saying no UI elements are ever allowed to be removed after they are added, as part of design policy or something? 

No -- the issue is about stable API. Processing algorithm's ARE bound by the stable API contract, and adding parameters to an algorithm adds this parameter to the algorithm's API. So we can't remove these in future without breaking the stable API contract.

isghj5:
So, can't make a UI element because it's auto generated from parameters which are used for API too...

Well, how do we add that setting to the general settings then?

nyalldawson:
> Well, how do we add that setting to the general settings then?

Have a look around

https://github.com/qgis/QGIS/blob/master/python/plugins/processing/core/ProcessingConfig.py#L67

mash-graz:
unfortunately i have to report some serious issues related to this changes. -- see  #32471

Issue #29627, opened 2019-04-09
Copy & Paste of Symbol from one category to another broken


qgib:
Author Name: **Brett Carlock** (@Saijin-Naib)
Original Redmine Issue: [21812](https://issues.qgis.org/issues/21812)
Affected QGIS version: 3.6.1
Redmine category:symbology

---
STR:
Copy symbol of one category
Select another category on the same layer, in the same Layer Symbology tab, without closing/reopening Layer properties or navigating away at any point
Paste symbol to another catgegory

Expected results:
Symbol is pasted to the selected category

Actual results:
Symbol is not pasted to the selected category

Example:
Copy symbol for "C" (County Route) to "SH" (State 900 Highway)
Does not transfer symbology, must be re-built manually.

---

- [Annotation 2019-04-10 151920.png](https://issues.qgis.org/attachments/download/14776/Annotation%202019-04-10%20151920.png) (Brett Carlock)

nyalldawson:
Fixed in master

PR #31017, opened 2019-07-31
Add more copy/paste symbol actions

nyalldawson:
Adds a few more copy and paste symbol actions in the GUI, allowing for easier transfer of symbols between parts of QGIS


nirvn (approved):
Nice work.

DelazJ:
@nyalldawson does this relate to #29627?

nyalldawson:
Yes, it fixes that

RudiU:
That would be a nice and welcome feature, but as of 3.8.2 it isn't working yet as a fix to #29627

nyalldawson:
@RudiU not all fixes are applied to stable branches. We only do it where it's deemed safe to do so.

PR #31020, opened 2019-07-31
topological editing: fix topological editing for trim/extend map tool

lbartoletti:
## Description

Add support of topological editing for trim/extend map tool.

The intersection point is added on the limit layer as on the modified layer.

Before
before

After
after

## Checklist

- [X] This PR passes all existing unit tests (test results will be reported by travis-ci after opening this PR)

### DIFF
--- a/src/app/qgsmaptooltrimextendfeature.cpp
+++ b/src/app/qgsmaptooltrimextendfeature.cpp
@@ -221,6 +221,7 @@ void QgsMapToolTrimExtendFeature::canvasReleaseEvent( QgsMapMouseEvent *e )
         {
           getPoints( match, pLimit1, pLimit2 );
           mStep = StepExtend;
+          mLimitLayer = match.layer();
         }
         break;
       case StepExtend:
@@ -231,6 +232,8 @@ void QgsMapToolTrimExtendFeature::canvasReleaseEvent( QgsMapMouseEvent *e )
 
           match.layer()->beginEditCommand( tr( "Trim/Extend feature" ) );
           match.layer()->changeGeometry( match.featureId(), mGeom );
+          match.layer()->addTopologicalPoints( mIntersection );
+          mLimitLayer->addTopologicalPoints( mIntersection );
           match.layer()->endEditCommand();
           match.layer()->triggerRepaint();
 
--- a/src/app/qgsmaptooltrimextendfeature.cpp
+++ b/src/app/qgsmaptooltrimextendfeature.cpp
@@ -255,8 +255,11 @@ void QgsMapToolTrimExtendFeature::canvasReleaseEvent( QgsMapMouseEvent *e )
 
             match.layer()->beginEditCommand( tr( "Trim/Extend feature" ) );
             match.layer()->changeGeometry( match.featureId(), mGeom );
-            match.layer()->addTopologicalPoints( mIntersection );
-            mLimitLayer->addTopologicalPoints( mIntersection );
+            if ( QgsProject::instance()->topologicalEditing() )
+            {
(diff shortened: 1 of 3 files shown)


src/app/qgsmaptooltrimextendfeature.cpp
```diff
@@ -221,6 +221,7 @@ void QgsMapToolTrimExtendFeature::canvasReleaseEvent( QgsMapMouseEvent *e )
         {
           getPoints( match, pLimit1, pLimit2 );
           mStep = StepExtend;
+          mLimitLayer = match.layer();
```
troopa81: Even if I understand there is no change to happen, maybe it will be better to reset mLimitLayer to 0 on deactivate

src/app/qgsmaptooltrimextendfeature.cpp
```diff
@@ -229,12 +230,18 @@ void QgsMapToolTrimExtendFeature::canvasReleaseEvent( QgsMapMouseEvent *e )
...
-          match.layer()->beginEditCommand( tr( "Trim/Extend feature" ) );
-          match.layer()->changeGeometry( match.featureId(), mGeom );
-          match.layer()->endEditCommand();
-          match.layer()->triggerRepaint();
+          if ( match.layer() )
+          {
+
+            match.layer()->beginEditCommand( tr( "Trim/Extend feature" ) );
+            match.layer()->changeGeometry( match.featureId(), mGeom );
+            match.layer()->addTopologicalPoints( mIntersection );
```
troopa81: Are we sure that mIntersection is linked to our match? what happen if we match and update intersection, then we move again and break before updating mIntersection like [here](https://github.com/troopa81/QGIS/blob/master/src/app/qgsmaptooltrimextendfeature.cpp#L128) or [here](https://github.com/troopa81/QGIS/blob/master/src/app/qgsmaptooltrimextendfeature.cpp#L119)

troopa81 (commented):
Good job! and really well tested

src/app/qgsmaptooltrimextendfeature.cpp
lbartoletti: I think so, since mIsModified is my flag for these cases.

src/app/qgsmaptooltrimextendfeature.cpp
lbartoletti: You're right, normally there is no chance to happen, but it's good for me to reset these layers. I added other constraints too.

lbartoletti:
Thanks for your review.

src/app/qgsmaptooltrimextendfeature.cpp
troopa81: Yes, but mIsModified is only updated if mIntersection is updated, so it won't be if we break in either of the 2 `if`, no?

nirvn:
@lbartoletti , it'd be good to finish this and merge prior to the next point release on August 16, the crasher fixed here is really easy to stumble upon.

src/app/qgsmaptooltrimextendfeature.cpp
lbartoletti: Yes, but we don't want to update the topology if the geometry isn't modified.
But, maybe I'm missing something.

lbartoletti: OK, with our conversation, I forced the flag to false even if it's very difficult to produce this case IMHO. :wink:

lbartoletti:
@nirvn I added another commit with some prevention. So if travis is green (of course it will be ;) ) You can merge it.

For the UX, I'll prepare another PR.

PR #31031, opened 2019-08-01
Use real symbol buttons in categorized and graduated renderer widgets, instead of fake ones

nyalldawson:
Gives these buttons the full power of the usual QgsSymbolButton,
including copy/paste symbols, color/opacity changes, etc. Plus,
some nice additional interface consistency!


nirvn (approved):
Looks good

PR #31047, opened 2019-08-02
[db_manager] support loading mixed curved and straight type tables

m-kuhn:
Allow postgres to properly detect mixed tables with curved and straight geometries of the same geometry type.
E.g. LineString and CompoundCurve. This follows the same approach that is used for multi types too.


src/core/geometry/qgswkbtypes.h
```diff
@@ -421,11 +423,155 @@ class CORE_EXPORT QgsWkbTypes
...
+
+        case MultiPolygonZM:
+        case MultiSurfaceZM:
+          return MultiSurfaceZM;
+
+        case NoGeometry:
+          return NoGeometry;
+
+        case Point25D:
+        case MultiPoint25D:
```
m-kuhn: Hmm, should these be handled like triangles (Unknown) since there's no curve type for these 25D ones available?

PR #31053, opened 2019-08-02
Add checkbox to allow non core devs to indicate backport desire

DelazJ:
## Description
Backport labels are a nice way to ease PR creation in other branches. But since label addition is a feature only core developers can trigger, the benefit is not really shared by the whole contributors, who still have to go the manual route.
This PR tries to fill the gap, allowing any contributor to indicate the branch the automatic backport can be done to (then those with write access can add the appropriate label) while the responsibility of the code is still on the contributor shoulders.
Also set existing texts as comments.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,15 +1,22 @@
 ## Description
-Include a few sentences describing the overall goals for this PR (pull request). If applicable also add screenshots.
+<!-- Include below a few sentences describing the overall goals for this PR (pull request). If applicable also add screenshots.-->
 
 ## Checklist
 
-> Reviewing is a process done by project maintainers, mostly on a volunteer basis. We try to keep the overhead as small as possible and appreciate if you help us to do so by completing the following items. Feel free to ask in a comment if you have troubles with any of them.
+<!-- Reviewing is a process done by project maintainers, mostly on a volunteer basis. We try to keep the overhead as small as possible and appreciate if you help us to do so by completing the following items. Feel free to ask in a comment if you have troubles with any of them.
+-->
 
 - [ ] Commit messages are descriptive and explain the rationale for changes
 - [ ] Commits which fix bugs include `fixes #11111` in the commit message next to the description
 - [ ] Commits which add new features are tagged with `[FEATURE]` in the commit message
 - [ ] Commits which change the UI or existing user workflows are tagged with `[needs-docs]` in the commit message and contain sufficient information in the commit message to be documented
 - [ ] I have read the [QGIS Coding Standards](https://docs.qgis.org/testing/en/docs/developers_guide/codingstandards.html) and this PR complies with them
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,9 +1,10 @@
 ## Description
-Include a few sentences describing the overall goals for this PR (pull request). If applicable also add screenshots.
+<!--Include a few sentences describing the overall goals for this PR (pull request). If applicable also add screenshots.-->
```
m-kuhn: Actually, I'd prefer to make the below check list shorter instead of hiding this line.
It's IMO the most important one (although quite often skipped), whereas things below like "this PR passes all existing unit tests" are pretty pointless.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -13,3 +14,11 @@ Include a few sentences describing the overall goals for this PR (pull request).
 - [ ] This PR passes all existing unit tests (test results will be reported by travis-ci after opening this PR)
 - [ ] New unit tests have been added for core changes
 - [ ] I have run [the `scripts/prepare-commit.sh` script](https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTING.md#contributing-to-qgis) before each commit
+
+## Backport
+
+<!-- Bug fixes can concern more than a branch and require backporting. Appropriate labels allow to automatically open the pull request in the corresponding branch, easing the process. You are however responsible of your code and should ensure there's no regression beforehand. Indicate below the version(s) you want to backport to.
+--> 
+
+- [ ] I request the backport of the changes to the Regular Release
```
jef-n: Latest

.github/PULL_REQUEST_TEMPLATE.md
m-kuhn: Oh wait, this will only be hidden once the PR is opened but still shown when entering data? If it works like that, that's a good thing (but doesn't change the point of the superfluous checkboxes).

.github/PULL_REQUEST_TEMPLATE.md
DelazJ: Aaaah!!! That's it! I was wondering while "Regular Release" abbrveiation was LR at https://qgis.org/en/site/getinvolved/development/roadmap.html#release-schedule

.github/PULL_REQUEST_TEMPLATE.md
DelazJ: Yes, it will only hide it in the preview. I think I will use `Include **below** a few sentences describing` to avoid people replacing the text between the tags (hence display nothing).
Any checkbox other than the units tests row, you'd like to remove?

DelazJ:
Travis was unhappy because [allows does not allow everything](https://github.com/qgis/QGIS/blob/master/scripts/spell_check/spelling.dat#L395) so I reword the instructions.

m-kuhn:
Ok, I'm not yet completely sold on the new checkboxes, I think it was handled very well handled as comments, but let's see how it goes.

PR #31054, opened 2019-08-02
Fix r.grow.distance

### LINKED ISSUE #31048: r.grow.distance in the QGIS 3.8.1 with GRASS 7.6.1
Hello

There is the following bug in the r.grow.distance in the QGIS 3.8.1 with GRASS 7.6.1

If i try to use the r.grow.distance with the setting "Output distances in meters instead of map units" I get the error "ERROR: Output distance in meters for lat/lon is only possible with 'metric=geodesic'"

There are only 4 options for metrics euclidean,manhattan,squared,maximum.

**TO FIX IT**

Find the r.grow.distance.txt in the pc and add the geodesic like the following line

QgsProcessingParameterEnum|metric|Metric|euclidean;squared;maximum;manhattan;geodesic|False|0|True

dimitrisal:
Bug fix for ERROR: Output distance in meters for lat/lon is only possible with 'metric=geodesic'

Fix #31048

### DIFF
--- a/python/plugins/processing/algs/grass7/description/r.grow.distance.txt
+++ b/python/plugins/processing/algs/grass7/description/r.grow.distance.txt
@@ -2,7 +2,7 @@ r.grow.distance
 Generates a raster layer of distance to features in input layer.
 Raster (r.*)
 QgsProcessingParameterRasterLayer|input|Input input raster layer|None|False
-QgsProcessingParameterEnum|metric|Metric|euclidean;squared;maximum;manhattan|False|0|True
+QgsProcessingParameterEnum|metric|Metric|euclidean;squared;maximum;manhattan;geodesic|False|0|True
 *QgsProcessingParameterBoolean|-m|Output distances in meters instead of map units|False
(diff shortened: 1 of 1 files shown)


dimitrisal:
Bug fix for ERROR: Output distance in meters for lat/lon is only possible with 'metric=geodesic'

roya0045:
Bugfix for #31048

m-kuhn:
@alexbruy does that look good to you?

alexbruy (approved):

m-kuhn:
@alexbruy and everyone, backport to 3.8 only or 3.4 too?

alexbruy:
@m-kuhn I would vote for backport to 3.8 and 3.4 as this is bugfix. If I'm not wrong 3.4 uses same GRASS version as 3.8 and master, so this shoiuld be safe.

m-kuhn:
Perfect, I've opened both pull requests.
Thanks @dimitrisal @roya0045 and @alexbruy

PR #31055, opened 2019-08-02
Fix misalignment of CRS label in Map Item properties

DelazJ:
The current issue is like this (affects 3.6 hence, I guess, 3.8 also - backport welcome? 3.4 is good.)
image


DelazJ:
Thanks @m-kuhn

PR #31061, opened 2019-08-02
[API/feature?] Cache feature ids for each symbol in the QgsVectorLayer

roya0045:
## Description
This is a proposal to cache the feature ids in the vector layer, similar to what is currently done with the counts.

The method to the the Fids is an extra in case someone would need it eventually.

This replaces the PR #30950 where it was proposed to use the counter directly but was discarded due to thread safety issues as highlighted by @nyalldawson 

The need to have the fids comes from the symbology expression PRs where I need to provide the fids of the features to the expression and I need to store and access this list of fids from somewhere when the aggregate function is evaluated.

Any feedback is appreciated

## Checklist

### DIFF
--- a/src/core/qgsvectorlayer.cpp
+++ b/src/core/qgsvectorlayer.cpp
@@ -697,10 +697,16 @@ long QgsVectorLayer::featureCount( const QString &legendKey ) const
   if ( !mSymbolFeatureCounted )
     return -1;
 
-  return mSymbolFeatureCountMap.value( legendKey );
+  return mSymbolFeatureCountMap.value( legendKey, -1 );
 }
 
+QgsFeatureIds QgsVectorLayer::symbolFeatureIds( const QString &legendKey ) const
+{
+  if ( !mSymbolFeatureCounted )
+    return QgsFeatureIds();
 
+  return mSymbolFeatureIdMap.value( legendKey, QgsFeatureIds() );
+}
 
 QgsVectorLayerFeatureCounter *QgsVectorLayer::countSymbolFeatures()
 {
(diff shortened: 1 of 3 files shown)


roya0045:
@nyalldawson Your opinion would be appreciated on the subject.

Is this caching fine with you and does the function clutter the QgsVectorLayer APi too much or is this tolerable?

roya0045:
@m-kuhn are you ok with this or do you still want to see what others have to say?

roya0045:
@m-kuhn  @nyalldawson any opinion on this proposal?

src/core/qgsvectorlayer.cpp
```diff
@@ -4020,7 +4028,11 @@ QVariant QgsVectorLayer::aggregate( QgsAggregateCalculator::Aggregate aggregate,
   {
     return QVariant();
   }
-
+  bool hasFids = false;
```
nyalldawson: Can you remove these changes please -- they are unrelated

src/core/qgsvectorlayer.h
```diff
@@ -997,6 +997,13 @@ class CORE_EXPORT QgsVectorLayer : public QgsMapLayer, public QgsExpressionConte
      */
     long featureCount( const QString &legendKey ) const;
 
+    /**
+     * Ids of features rendered with specified legend key. Features must be first
+     * calculated by countSymbolFeatures()
+     * \returns Ids of features rendered by symbol or -1 if failed or Ids are not available
```
nyalldawson: Nearly there!

Last change - can you add

    \since QGIS 3.10

Here?

roya0045: Should be done, took far longer than I wanted.

PR #31065, opened 2019-08-03
Show auxiliary fields in field calculator dialog

nyalldawson:
BUT only if they have a non-hidden editor widget.

Rationale: it's very handy to be able to bulk-update these in certain
circumstances (e.g. select a bunch of features, set their label alignment
to "right"). But on the other hand, using the label tools creates a LOT
of auxiliary fields which clutter this dialog (which is why they were
hidden in the first place).

SO instead, default to hiding them all, but if a user HAS specifically
changed the field to show it, then allow them to edit it through
field calculator....

### DIFF
--- a/src/app/qgsfieldcalculator.cpp
+++ b/src/app/qgsfieldcalculator.cpp
@@ -29,6 +29,7 @@
 #include "qgsguiutils.h"
 #include "qgsproxyprogresstask.h"
 #include "qgsexpressioncontextutils.h"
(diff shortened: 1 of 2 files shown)


nirvn:
Sounds good to me.

Is the same logic already applied to the attribute table's top calculator bar?

nyalldawson:
> Is the same logic already applied to the attribute table's top calculator bar?

No - I looked into this too, but it's non-trivial.

PR #31071, opened 2019-08-04
Apply same logic regarding joined fields as is used in field calculator...

nyalldawson:
to QgsFieldProxyModel set to hiding read only fields

I.e. show editable joined fields, but only if they are set to a non-hidden
editor widget


nirvn:
@nyalldawson , logic sounds good, +1 to merge.

PR #31076, opened 2019-08-04
Accept authcfg with or without quotes

### LINKED ISSUE #31045: Authcfg support for PostGIS Rasters
PostGIS Rasters can be only loaded using connections with plain username & password, not the safe authentication configurations. This is because the GDAL provider doesn't use the authentication system to resolve the authcfg uri parameter, but it badly limits the use of PostGIS Raster to either not stored or unencrypted passwords (what is deprecated since quite a while in my impression).

elpaso:
Fix #31045


m-kuhn (commented):
Does the removed space in the beginning make no difference?

elpaso:
No, I am not sure why I added it in the first place.

m-kuhn (approved):

borysiasty (commented):
LGTM (and works as well). Thanks!

PR #31080, opened 2019-08-05
Add API for custom handlers for rendered features

nyalldawson:
This PR adds an interface for classes which provider custom handlers for features which are rendered as part of a map render job.

QgsRenderedFeatureHandlerInterface objects are registered in the QgsMapSettings objects used to construct map render jobs. During the rendering operation, the handleRenderedFeature() method will be called once for every rendered feature, allowing the handler to perform some custom task based on the provided information.

They can be used for custom tasks which operate on a set of rendered features, such as creating spatial indexes of the location and rendered symbology bounding box of all features rendered on a map.

The initial use case for this is for GeoPDF exports, where we need to collect a list of all rendered features, along with the approximate bounds of the symbol used to represent the feature. But the API here has been specifically designed with other demonstrated future use cases in mind, such as:

- creating an index of rendered point feature bounds, for use in QGIS server identify requests, allowing for identification to work correctly with offset or displaced points
- creating an index of rendered point features bounds for use in the QGIS desktop "rotate point features" or "offset point features" map tools, to allow existing marker symbols to be clicked on a map and moved/rotated (instead of the current behaviour, where the actual point geometry location must be clicked)
- allowing the displaced point rendered to collect and register the actual rendered location of displaced points for use when labeling these points 
...etc

### DIFF
--- a/src/core/qgsrenderedfeaturehandlerinterface.h
+++ b/src/core/qgsrenderedfeaturehandlerinterface.h
@@ -17,9 +17,13 @@
 #define QGSRENDEREDFEATUREHANDLERINTERFACE_H
 
 #include "qgis_core.h"
+#include <QSet>
+#include <QString>
 
 class QgsFeature;
 class QgsGeometry;
+class QgsRenderContext;
+class QgsVectorLayer;
 
 /**
  * \ingroup core
@@ -67,6 +71,11 @@ class CORE_EXPORT QgsRenderedFeatureHandlerInterface
      */
     virtual void handleRenderedFeature( const QgsFeature &feature, const QgsGeometry &renderedBounds, const QgsRenderedFeatureHandlerInterface::RenderedFeatureContext &context ) = 0;
 
+    /**
+     * Returns a list of attributes required by this handler, for the specified \a layer. Attributes not listed in here may
+     * not be requested from the provider at rendering time.
+     */
+    virtual QSet<QString> usedAttributes( QgsVectorLayer *layer, const QgsRenderContext &context ) const { Q_UNUSED( layer ); Q_UNUSED( context ); return QSet< QString >(); }
 };
 
 #endif // QGSRENDEREDFEATUREHANDLERINTERFACE_H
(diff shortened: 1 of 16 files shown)


nyalldawson:
@elpaso - we've previously discussed this, specifically in relation to the server API case. I'd be interested to hear if you think this API would meet the requirements for that use case.

nyalldawson:
Also - one other possible use case (which I'm ... hesitant... to raise, but bears thinking about anyway):

- we could potentially use this to resurrect the older QGIS behaviour where we built a snapping index at render time, but in a MUCH cleaner way. (ping @mhugo )

src/core/qgsrenderedfeaturehandlerinterface.h
```diff
@@ -0,0 +1,81 @@
...
+class CORE_EXPORT QgsRenderedFeatureHandlerInterface
+{
+  public:
+    virtual ~QgsRenderedFeatureHandlerInterface() = default;
+
+    struct CORE_EXPORT RenderedFeatureContext
+    {
+      ///@cond PRIVATE
+      // required to allow compilation only until real members are present
+      bool dummy;
```
m-kuhn: Can we make it disappear from sip with

```suggestion
    private:
      bool dummy;
```

m-kuhn:
Nice! That will allow for some really cool features.

While not in the scope of this PR, I wonder for the server use case if the index would be build at GetMap time (and if yes at what time it would be destroyed again, traditional stateful/stateless question) or in another rendering job at `GetFeatureInfo` time.

nyalldawson:
>  I wonder for the server use case if the index would be build at GetMap time (and if yes at what time it would be destroyed again, traditional stateful/stateless question) or in another rendering job at GetFeatureInfo time.

@elpaso and I have previously discussed this -- it would need to be done at GetFeatureInfo time. We'd probably want to add a new render context flag "dont actually DRAW anything" to speed stuff up too - this would do all the other map rendering stuff, but shortcut the actual drawing operations.

But that's a future change :)

PR #31081, opened 2019-08-05
Misc 3.4 Backports

nyalldawson:
Backporting some misc fixes

### DIFF
--- a/src/app/qgsapplayertreeviewmenuprovider.cpp
+++ b/src/app/qgsapplayertreeviewmenuprovider.cpp
@@ -691,6 +691,8 @@ void QgsAppLayerTreeViewMenuProvider::editVectorSymbol()
     singleRenderer->setSymbol( symbol.release() );
     layer->triggerRepaint();
     mView->refreshLayerSymbology( layer->id() );
+    layer->emitStyleChanged();
+    QgsProject::instance()->setDirty( true );
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/app/qgsapplayertreeviewmenuprovider.cpp
```diff
@@ -691,6 +691,8 @@ void QgsAppLayerTreeViewMenuProvider::editVectorSymbol()
     singleRenderer->setSymbol( symbol.release() );
     layer->triggerRepaint();
     mView->refreshLayerSymbology( layer->id() );
+    layer->emitStyleChanged();
+    QgsProject::instance()->setDirty( true );
```
m-kuhn: Would it make sense to connect the layer style changed signal directly to a project setDirty slot (by the project, when a layer is added)?

nyalldawson: That signal is fired for internal reasons too, whereas dirtying should only ever happen as a direct result of user interaction

m-kuhn: In this case a `void emitStyleChanged( bool makeProjectDirty = true );` might be something to consider.

Just thinking it's too easy to miss this.

nyalldawson: Well really having an emitXxx method in the first place is an antipattern, hinting that this whole part of the api needs a rework.

m-kuhn: Ok, thanks!

m-kuhn (approved):

PR #31085, opened 2019-08-05
Fix layout of QML widget config

signedav:
## Description
So the code edit field size is adapted to the form size.

Avoiding to have a small code-box like that:
qml_1

I know it's very minor - but has been annoying.


m-kuhn:
Nice! :+1:

PR #31111, opened 2019-08-06
[text renderer] Add rotation-based orientation mode

nirvn:
## Description
This PR adds an orientation mode - for label mode only - that switches between horizontal and vertical based on the label rotation. GIF:

This is needed to properly label line features in parallel placement mode:
image

@nyalldawson , I'd appreciate your review on this too. 

## Checklist


nirvn:
Oh BTW, travis will fail here as I've updated the default spacing between characters in vertical orientation mode to be slightly narrower by default.

src/core/qgspallabeling.cpp
```diff
@@ -1550,8 +1554,38 @@ void QgsPalLayerSettings::calculateLabelSize( const QFontMetricsF *fm, const QSt
...
         maxLineLength = std::max( maxLineLength, line.length() );
       }
-      h = labelHeight * maxLineLength + ( maxLineLength - 1 ) * letterSpacing;
+      h = fm->ascent() * maxLineLength + ( maxLineLength - 1 ) * letterSpacing;
       break;
+    }
+
+    case QgsTextFormat::RotationBasedOrientation:
+    {
+      double widthHorizontal = 0.0;
```
nyalldawson: If I'm reading this right - it's returning the largest size possible, right? I.e. not the actual label size, but the worst case scenario?

If so, that's far from ideal, because it means candidates in the unused areas will be stripped, labels won't be placed near the edges of the map, polygons set to place only labels which fit inside won't be labeled, etc. 

I think you're going to have to push this back into the pal library -- so PAL would need a concept of horizontal or vertical label size, and it would use the appropriate one based on the candidate's rotation.

nirvn:
@nyalldawson , I've updated the commit to have dynamic label size based on a given angle.

The dynamic sizing happens when creating candidates on/around points, but not when placing curved / parallel labels. I think we need to disable orientation altogether for curved, and for parallel, things are looking good already as is (on top of which, since the angle is dynamically calculated on the basis of a given label width to begin with, I'm not sure how that'd be done).

src/core/qgspallabeling.h
```diff
@@ -917,7 +917,7 @@ class CORE_EXPORT QgsPalLayerSettings
      * Calculates the space required to render the provided \a text in map units.
      * Results will be written to \a labelX and \a labelY.
      */
-    void calculateLabelSize( const QFontMetricsF *fm, const QString &text, double &labelX, double &labelY, const QgsFeature *f = nullptr, QgsRenderContext *context = nullptr );
+    void calculateLabelSize( const QFontMetricsF *fm, const QString &text, double &labelX, double &labelY, const QgsFeature *f = nullptr, QgsRenderContext *context = nullptr, double *rotatedLabelX = nullptr, double *rotatedLabelY = nullptr );
```
nyalldawson: SIP_OUT for these? (Also explanation in the dox)

src/core/qgstextrenderer.cpp
```diff
@@ -3537,13 +3615,13 @@ void QgsTextRenderer::drawTextInternal( TextPart drawType,
...
 
-        double labelHeight = fontMetrics->ascent() + fontMetrics->descent() + ( fontMetrics->ascent() + fontMetrics->descent() + letterSpacing ) * ( line.length() - 1 );
+        double labelHeight = fontMetrics->ascent() + ( fontMetrics->ascent() + letterSpacing ) * ( line.length() - 1 );
 
         Component subComponent;
         subComponent.text = line;
         subComponent.size = QSizeF( labelWidth, labelHeight );
         subComponent.offset = QPointF( 0.0, 0.0 );
-        subComponent.rotation = -component.rotation * 180 / M_PI;
+        subComponent.rotation = -component.rotation * 180 / M_PI;  ;
```
nyalldawson:
```suggestion
        subComponent.rotation = -component.rotation * 180 / M_PI;
```

src/core/qgstextrenderer.h
```diff
@@ -1022,6 +1022,7 @@ class CORE_EXPORT QgsTextFormat
     {
       HorizontalOrientation, //!< Vertically oriented text
       VerticalOrientation, //!< Horizontally oriented text
+      RotationBasedOrientation, //!< Horizontally or vertically oriented text based on rotation
```
nyalldawson: Better note here that it's only usable during map labelling, not in standalone text rendering

src/core/qgslabelfeature.cpp
```diff
@@ -78,6 +78,15 @@ QgsFeature QgsLabelFeature::feature() const
   return mFeature;
 }
 
+QSizeF QgsLabelFeature::size( double angle ) const
+{
+  if ( mRotatedSize.isEmpty() )
+    return mSize;
+
+  // Between 45 to 135 and 235 to 313 degrees, return the rotated size
```
DelazJ: Forgive my ignorance: why 313 and not 315?

nirvn: Good catch, it's a typo, should be 315.

DelazJ: indeed I see the right number in code later.

PR #31133, opened 2019-08-07
Add option to simplify PDF or SVG exports 

nyalldawson:
This new setting, "Simplify geometries to reduce output file size", is exposed in the SVG or PDF export settings dialogs shown when exporting a layout/atlas/report to PDF or SVG.

If checked (the new default), geometries will be simplified while exporting layouts in order to remove any redundant vertices which are not discernably different at the export DPI. (e.g. if export DPI is 300 dpi, vertices less then 1/600 inch different from each other will be removed).

This avoids exporting ALL geometry vertices during PDF/SVG export, which can result in a ridiculously complex and large export file size. Aside from the file size issues, it also causes problems when trying to load these outputs into other applications, e.g. Inkscape, which chokes on the huge number of vertices and grinds to a crawl.

Sponsored by the GeoPDF export group

### DIFF
--- a/python/core/auto_generated/layout/qgslayoutexporter.sip.in
+++ b/python/core/auto_generated/layout/qgslayoutexporter.sip.in
@@ -190,6 +190,8 @@ Constructor for PdfExportSettings
 
       QgsRenderContext::TextRenderFormat textRenderFormat;
 
+      bool simplifyGeometries;
+
     };
 
     ExportResult exportToPdf( const QString &filePath, const QgsLayoutExporter::PdfExportSettings &settings );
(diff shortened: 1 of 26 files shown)


nyalldawson:
For reference:

This (crappy) little map using an original full quality boundaries layer went from a 2.5mb to 500kb PDF, with no apparent loss of quality in the exported PDF:
image

If we zoom out more, the difference is even greater: 2.2mb to 210kb.

image

nirvn (commented):
Nice; code looks good.

nyalldawson:
Thanks for the speedy review @nirvn!

PR #31143, opened 2019-08-07
Add mutex locker when accessing QgsPostGressConn attribute member

troopa81:
## Description

I come with dead lock/seg fault issues in QgsPostGresConn trying to parallelize prepareIndex in QgsSnappingUtils while working on Snap Cache improvements (https://docs.google.com/document/d/1VzgLJ-hy2GtcWrAy5geivhDDdXRGJUDcikCWUtXRe9k/edit#heading=h.uau3di2m4nvn).

Postgres connections are shared and can be accessed from different threads during a parallel rendering process for instance (or while computing snapping cache for several postgres layer in parallel).

But some of the methods that access attribute members are not mutex protected. This PR fixes that. 

I didn't add some unit tests because it seems quite impossible to write a deterministic test while testing thread. 

## Checklist

### DIFF
--- a/src/providers/postgres/qgspostgresconn.cpp
+++ b/src/providers/postgres/qgspostgresconn.cpp
@@ -371,8 +371,15 @@ QgsPostgresConn::~QgsPostgresConn()
   mConn = nullptr;
 }
 
+void QgsPostgresConn::ref()
+{
+  QMutexLocker locker( &mLock );
+  ++mRef;
+}
+
 void QgsPostgresConn::unref()
 {
+  QMutexLocker locker( &mLock );
   if ( --mRef > 0 )
     return;
 
@@ -386,6 +393,8 @@ void QgsPostgresConn::unref()
     connections.remove( key );
   }
 
+  // to avoid destroying locked mutex
+  locker.unlock();
   delete this;
 }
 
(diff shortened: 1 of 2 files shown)


rldhont:
It seems good to me. Someone else to validate ?

elpaso:
LGTM I wonder if this also solves the deadlock issue of loading two identical postgis rasters at project load. Probably yes.

wonder-sk:
:thinking: a single connection should not be used by multiple threads at the same time - each thread should request a connection separately from the connection pool...

Is there a reason why you need to share a single connection between threads?

troopa81:
> Is there a reason why you need to share a single connection between threads?

@wonder-sk as we discuss yesterday, the postgres connection is shared when you use a transaction connection. And so it could be access from different threads.

troopa81:
> LGTM I wonder if this also solves the deadlock issue of loading two identical postgis rasters at project load.

No, I still reproduce the issue even with this PR. At first look, It seems to come from gdal...

wonder-sk:
Yeah after chatting with Julien I have realized that because of the "transaction" connections we need the locking. I have forgotten about that concept and assumed that connections are never shared. So +1 from me to merge this PR.

On a side note, I have a feeling that the whole implementation of transaction connections may be flawed, with some functionality broken due to the fact that only a single connection is shared across the whole QGIS instance.

nyalldawson:
@wonder-sk I would agree with that. When I implemented the transaction support for Oracle recently I was very surprised that the system relies on sharing a connection, and that this works at all! That said, there's no easy answers here that I could see, and we'd need to rework the transaction API substantially. Something to consider for 4.0.

In the meantime, I also think we should merge this and try to get a bit more stability in the feature.

PR #31147, opened 2019-08-07
Always clear list of previously encountered errors on each new algorithm execution

nyalldawson:
Avoids the situation where errors reported during an algorithm
which was overall successully executed are append to the errors
for later execution steps


nyalldawson:
Unrelated test failure

PR #31150, opened 2019-08-08
Allow setting the raster renderer for invalid raster layers

nyalldawson:
Fixes inaccessible (and unsettable) raster renderer information and styling properties when a raster layer is invalid.

This brings invalid raster layer handling closer in line with invalid vector layer handling, where the renderer information is still available for view in the UI for layers with broken paths.

No intention to backport -- it's too risky, and there'll likely be follow up fixes required.

Sponsored by SLYR

### DIFF
--- a/src/core/raster/qgsrasterlayer.cpp
+++ b/src/core/raster/qgsrasterlayer.cpp
@@ -809,30 +809,34 @@ void QgsRasterLayer::setDataProvider( QString const &provider, const QgsDataProv
 
 void QgsRasterLayer::setDataSource( const QString &dataSource, const QString &baseName, const QString &provider, const QgsDataProvider::ProviderOptions &options, bool loadDefaultStyleFlag )
 {
-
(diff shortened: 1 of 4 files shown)


nyalldawson:
@elpaso I know you previously investigated this, can you see if this change looks ok to you?

elpaso:
@nyalldawson LGTM.

Generally speaking, (for vectors too), I don't like very much the idea of creating renderers for invalid layers just for the purpose of (un)marshalling the styles, it looks like a patch to me and  I sense there should be a better design (but I don't know what could it be).

nyalldawson:
>  I don't like very much the idea of creating renderers for invalid layers just for the purpose of (un)marshalling the styles, it looks like a patch to me and I sense there should be a better design (but I don't know what could it be).

Actually it goes deeper then just marshalling (and maybe that code isn't required anymore now). It's more about always having access to the renderers for invalid layers, so things like legends don't break.

PR #31152, opened 2019-08-08
Color code welcome page items

nyalldawson:
For consideration: the welcome page now has 3 distinct types of items (projects, templates and news). I think we need greater visual distinction between these. 

This mockup shows a possible approach - using colors based on the QGIS logo to shade the different item types:

image

### DIFF
--- a/src/app/qgsprojectlistitemdelegate.cpp
+++ b/src/app/qgsprojectlistitemdelegate.cpp
@@ -41,33 +41,32 @@ void QgsProjectListItemDelegate::paint( QPainter *painter, const QStyleOptionVie
   QAbstractTextDocumentLayout::PaintContext ctx;
   QStyleOptionViewItem optionV4 = option;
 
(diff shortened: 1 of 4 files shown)


nyalldawson:
For reference, here's the current appearance:

image

m-kuhn:
Good move.

Some suggestions:

 - Change the red tone of the news to blue (color blindness and a bit less strong)
 - Move the news item below the template projects. The news are an offer to the user which he should go to when he wants. In his daily life his project are likely more valuable to him than the project news.

nirvn:
@nyalldawson , tried this out. One comment linked directly to this PR: the rounded rectangle outline stroke needs more transparency so it better blenders against the selected item background highlight. ATM, it doesn't look superb.

m-kuhn:
Actually, the reason to use these colors are quite compelling :+1: . So it's just their distribution among the categories which could be discussed.

m-kuhn:
@nyalldawson can we revive this?

PR #31156, opened 2019-08-08
Shorten pull request template

m-kuhn:
## Description

This shortens the pull request template.

Let's assume that most people will not be able to rewrite their git history, so for giving hints for commit messages it's too late and some things are better specified on PR level than on commit level anyway. Some commit message related suggestions are left for educational reasons.

## Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,5 +1,5 @@
 ## Description
-<!-- Include below a few sentences describing the overall goals for this PR (pull request). If applicable also add screenshots.-->
+<!-- Include below a few sentences describing the overall goals for this pull request (PR). If applicable also add screenshots.-->
 
 ## Checklist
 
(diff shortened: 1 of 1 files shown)


m-kuhn:
CC @DelazJ

DelazJ:
LGTM. Simpler, with just the necessary. Thanks.
I guess it preludes your grant proposal.

m-kuhn:
Thanks for the feedback @DelazJ 
It wasn't triggered by the grant proposal, but while working on this PR I realized that there's a connection.

nirvn:
@m-kuhn , somewhat related: while I had no issue with adding the backport checkboxes per say, seeing live has me wondering whether we should remove them.

The issue has to do with it meaning an less than 100% check count when looking at PRs through the list. Makes it more difficult to spot which PRs have required steps missing.

An alternative checkbox could be [x] I have evaluated whether it is appropriate for this PR to be backported. Rely on labels for actual requests.

m-kuhn:
@nirvn good point which expands to a more generic requirement of having the checkboxes being formulated in a way they are applicable to every single pull request.

PR #31158, opened 2019-08-08
"Allowed gaps" layer for gap check

m-kuhn:
This adds a new option for the gaps geometry check to add an "allowed gaps" layer with a buffer parameter.

This allowed gaps layer acts as a mask in the check and gaps in this area are not reported as errors. This can e.g. be interesting for a topological land use map where there are lakes which are allowed to be excluded.
The buffer parameter can be used for tolerance for slight variations along the borders of the lake.

There is a new fix available to add a reported gap as a new polygon to the allowed gaps layer. It is the users responsibility to guarantee that this can be done fully automated (e.g. no NOT NULL constraints etc.).
One thing to note is that if a gap grows and is added again, it will just be added on top of the existing gap polygon. There are alternatives to that (like only adding the missing bits, but especially with buffering enabled, that's almost impossible to do, or removing overlapping geometries, but removing geometries also seems uncontrollable)

Peek 2019-08-06 11-06

### DIFF
--- a/src/analysis/vector/geometry_checker/qgsgeometrycheckcontext.h
+++ b/src/analysis/vector/geometry_checker/qgsgeometrycheckcontext.h
@@ -27,12 +27,13 @@
  * \note This class is a technology preview and unstable API.
  * \since QGIS 3.4
  */
-struct ANALYSIS_EXPORT QgsGeometryCheckContext
+class ANALYSIS_EXPORT QgsGeometryCheckContext
 {
+  public:
     QgsGeometryCheckContext( int precision,
                              const QgsCoordinateReferenceSystem &mapCrs,
                              const QgsCoordinateTransformContext &transformContext,
-                             const QgsProject *project );
+                             const QgsProject *mProject );
 
     /**
      * The tolerance to allow for in geometry checks.
@@ -66,7 +67,10 @@ struct ANALYSIS_EXPORT QgsGeometryCheckContext
      *
      * \since QGIS 3.10
      */
-    const QgsProject *project;
+    const QgsProject *project() const;
+
+  private:
+    const QgsProject *mProject;
 
   private:
 #ifdef SIP_RUN
--- a/src/analysis/vector/geometry_checker/qgsgeometrycheckcontext.h
+++ b/src/analysis/vector/geometry_checker/qgsgeometrycheckcontext.h
@@ -30,6 +30,10 @@
 class ANALYSIS_EXPORT QgsGeometryCheckContext
 {
   public:
+
+    /**
+     * Creates a new QgsGeometryCheckContext.
+     */
     QgsGeometryCheckContext( int precision,
                              const QgsCoordinateReferenceSystem &mapCrs,
                              const QgsCoordinateTransformContext &transformContext,
--- a/src/analysis/vector/geometry_checker/qgsgeometrycheckcontext.h
+++ b/src/analysis/vector/geometry_checker/qgsgeometrycheckcontext.h
@@ -31,7 +31,8 @@ struct ANALYSIS_EXPORT QgsGeometryCheckContext
 {
     QgsGeometryCheckContext( int precision,
                              const QgsCoordinateReferenceSystem &mapCrs,
-                             const QgsCoordinateTransformContext &transformContext );
+                             const QgsCoordinateTransformContext &transformContext,
+                             const QgsProject *project );
 
     /**
      * The tolerance to allow for in geometry checks.
@@ -57,6 +58,12 @@ struct ANALYSIS_EXPORT QgsGeometryCheckContext
      * The coordinate transform context with which transformations will be done.
      */
     const QgsCoordinateTransformContext transformContext;
+
+    /**
+     * The project ... blablabla
+     * Only to be used in the main thread (prepare method)
+     */
+    const QgsProject *project;
 
   private:
 #ifdef SIP_RUN
--- a/src/analysis/vector/geometry_checker/qgsgeometrycheckcontext.h
+++ b/src/analysis/vector/geometry_checker/qgsgeometrycheckcontext.h
@@ -22,6 +22,7 @@
 #include "qgsfeaturepool.h"
 
 /**
+ * \ingroup analysis
  * Base configuration for geometry checks.
  *
  * \note This class is a technology preview and unstable API.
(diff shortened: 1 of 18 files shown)


src/analysis/vector/geometry_checker/qgsgeometrycheckcontext.h
```diff
@@ -58,6 +59,15 @@ struct ANALYSIS_EXPORT QgsGeometryCheckContext
...
     const QgsCoordinateTransformContext transformContext;
 
+    /**
+     * The project can be used to resolve additional layers.
+     *
+     * This must only be accessed from the main thread (i.e. do not access from the collectError method)
+     *
+     * \since QGIS 3.10
+     */
+    const QgsProject *project;
```
nyalldawson: Might be safest to make this a function, and Q_ASSERT that it IS being called only from the main thread

src/analysis/vector/geometry_checker/qgsgeometrygapcheck.cpp
```diff
@@ -19,27 +19,69 @@
...
+    QVector<QgsGeometry> allowedGaps;
+    QgsFeatureRequest request;
+    request.setSubsetOfAttributes( QgsAttributeList() );
+    QgsFeatureIterator iterator = mAllowedGapsSource->getFeatures( request );
+    QgsFeature feature;
 
+    while ( iterator.nextFeature( feature ) )
+    {
+      QgsGeometry geom = feature.geometry();
+      QgsGeometry gg = geom.buffer( mAllowedGapsBuffer, 20 );
```
nyalldawson: What's 20 for?

src/analysis/vector/geometry_checker/qgsgeometrygapcheck.cpp
```diff
@@ -19,27 +19,69 @@
...
+      allowedGaps.append( gg );
+    }
+
+    std::unique_ptr< QgsGeometryEngine > allowedGapsEngine = QgsGeometryCheckerUtils::createGeomEngine( nullptr, mContext->tolerance );
+    allowedGapsEngine->prepareGeometry();
```
nyalldawson: No need for this -- prepared geometries are used in GEOS combine operations (only the spatial predicates utilise prepared geometries)

src/analysis/vector/geometry_checker/qgsgeometrygapcheck.cpp
```diff
@@ -97,9 +139,10 @@ void QgsGeometryGapCheck::collectErrors( const QMap<QString, QgsFeaturePool *> &
   }
 
   // For each gap polygon which does not lie on the boundary, get neighboring polygons and add error
-  for ( int iPart = 0, nParts = diffGeom->partCount(); iPart < nParts; ++iPart )
+  QgsGeometryPartIterator parts = diffGeom->parts();
+  while ( parts.hasNext() )
   {
-    std::unique_ptr<QgsAbstractGeometry> gapGeom( QgsGeometryCheckerUtils::getGeomPart( diffGeom.get(), iPart )->clone() );
+    QgsAbstractGeometry *gapGeom = parts.next();
```
nyalldawson:
```suggestion
    const QgsAbstractGeometry *gapGeom = parts.next();
```

src/app/qgsvectorlayerproperties.cpp
```diff
@@ -462,6 +462,27 @@ QgsVectorLayerProperties::QgsVectorLayerProperties(
       cb->setChecked( activeChecks.contains( factory->id() ) );
       mGeometryCheckFactoriesGroupBoxes.insert( cb, factory->id() );
       topologyCheckLayout->addWidget( cb );
+      QString id = factory->id();
+      if ( factory->id() == QStringLiteral( "QgsGeometryGapCheck" ) )
+      {
+        const QVariantMap gapCheckConfig = mLayer->geometryOptions()->checkConfiguration( QStringLiteral( "QgsGeometryGapCheck" ) );
+
+        mGapCheckAllowExceptionsActivatedCheckBox = new QgsCollapsibleGroupBox( tr( "Allowed gaps" ) );
```
nyalldawson:
```suggestion
        mGapCheckAllowExceptionsActivatedCheckBox = new QgsCollapsibleGroupBox( tr( "Allowed Gaps" ) );
```

src/app/qgsvectorlayerproperties.cpp
```diff
@@ -462,6 +462,27 @@ QgsVectorLayerProperties::QgsVectorLayerProperties(
...
+        mGapCheckAllowExceptionsActivatedCheckBox->setCheckable( true );
+        mGapCheckAllowExceptionsActivatedCheckBox->setChecked( gapCheckConfig.value( QStringLiteral( "allowedGapsEnabled" ), false ).toBool() );
+        QFormLayout *layout = new QFormLayout();
+        mGapCheckAllowExceptionsActivatedCheckBox->setLayout( layout );
+        topologyCheckLayout->addWidget( mGapCheckAllowExceptionsActivatedCheckBox );
+        mGapCheckAllowExceptionsLayerComboBox = new QgsMapLayerComboBox();
+        mGapCheckAllowExceptionsLayerComboBox->setFilters( QgsMapLayerProxyModel::PolygonLayer );
+        mGapCheckAllowExceptionsLayerComboBox->setLayer( QgsProject::instance()->mapLayer( gapCheckConfig.value( QStringLiteral( "allowedGapsLayer" ) ).toString() ) );
```
nyalldawson: Should this combo box filter out the target layer itself too? (i.e. don't allow setting an "allowed gaps" layer for a polygon layer as itself)

src/app/qgsvectorlayerproperties.cpp
```diff
@@ -462,6 +462,27 @@ QgsVectorLayerProperties::QgsVectorLayerProperties(
...
+        layout->addRow( new QLabel( tr( "Layer" ) ), mGapCheckAllowExceptionsLayerComboBox );
+        mGapCheckAllowExceptionsBufferLineEdit = new QLineEdit();
```
nyalldawson: Why not a QgsDoubleSpinBox instead?

src/analysis/vector/geometry_checker/qgsgeometrygapcheck.cpp
m-kuhn: number of segments

src/app/qgsvectorlayerproperties.cpp
m-kuhn: Can do.
I never liked spin boxes. One advantage they have though is the suffix which can be used for the units.

PR #31167, opened 2019-08-08
Update widgets with default values on attribute form

signedav:
Updates widget values on real time while editing the referenced fields.
default_value_01

To avoid recursion it stores the currently updated fields index in `mAlreadyUpdatedFields`. The widgets contains as default value the value of the previous one plus 1:
default_value_02

To Do:
-[ ] Testfunction in `testqgsattributeform`

### DIFF
--- a/src/core/qgis.cpp
+++ b/src/core/qgis.cpp
@@ -297,5 +297,5 @@ uint qHash( const QVariant &variant )
 
 bool qgsVariantEqual( const QVariant &lhs, const QVariant &rhs )
 {
-  return lhs.isNull() == rhs.isNull() && lhs == rhs;
+  return ( lhs.isNull() == rhs.isNull() && lhs == rhs ) || ( lhs.isNull() && rhs.isNull() );
 }
--- a/src/core/qgis.cpp
+++ b/src/core/qgis.cpp
@@ -28,6 +28,7 @@
 #include "qgsconfig.h"
 #include "qgslogger.h"
 #include "qgswkbtypes.h"
+#include <QDebug>
 
 #include <gdal.h>
 #include <ogr_api.h>
@@ -297,5 +298,5 @@ uint qHash( const QVariant &variant )
 
 bool qgsVariantEqual( const QVariant &lhs, const QVariant &rhs )
 {
-  return ( lhs.isNull() == rhs.isNull() && lhs == rhs ) || ( lhs.isNull() && rhs.isNull() );
+  return ( lhs.isNull() == rhs.isNull() && lhs == rhs ) || ( lhs.isNull() && rhs.isNull() && lhs.isValid() && rhs.isValid() );
 }
--- a/src/core/qgis.cpp
+++ b/src/core/qgis.cpp
@@ -28,7 +28,6 @@
 #include "qgsconfig.h"
 #include "qgslogger.h"
 #include "qgswkbtypes.h"
-#include <QDebug>
 
 #include <gdal.h>
 #include <ogr_api.h>
--- a/src/gui/qgsattributeform.cpp
+++ b/src/gui/qgsattributeform.cpp
@@ -319,12 +319,7 @@ bool QgsAttributeForm::saveEdits()
         QVariant dstVar = dst.at( eww->fieldIdx() );
         QVariant srcVar = eww->value();
 
-        // need to check dstVar.isNull() != srcVar.isNull()
-        // otherwise if dstVar=NULL and scrVar=0, then dstVar = srcVar
-        // be careful- sometimes two null qvariants will be reported as not equal!! (e.g., different types)
-        bool changed = ( dstVar != srcVar && !dstVar.isNull() && !srcVar.isNull() )
-                       || ( dstVar.isNull() != srcVar.isNull() );
-        if ( changed && srcVar.isValid() && fieldIsEditable( eww->fieldIdx() ) )
+        if ( !qgsVariantEqual( dstVar, srcVar ) && srcVar.isValid() && fieldIsEditable( eww->fieldIdx() ) )
         {
           dst[eww->fieldIdx()] = srcVar;
 
@@ -437,12 +432,7 @@ bool QgsAttributeForm::updateDefaultValues( const int originIdx )
         QVariant dstVar = dst.at( eww->fieldIdx() );
         QVariant srcVar = eww->value();
 
-        // need to check dstVar.isNull() != srcVar.isNull()
-        // otherwise if dstVar=NULL and scrVar=0, then dstVar = srcVar
-        // be careful- sometimes two null qvariants will be reported as not equal!! (e.g., different types)
-        bool changed = ( dstVar != srcVar && !dstVar.isNull() && !srcVar.isNull() )
-                       || ( dstVar.isNull() != srcVar.isNull() );
-        if ( changed && srcVar.isValid() && fieldIsEditable( eww->fieldIdx() ) )
+        if ( !qgsVariantEqual( dstVar, srcVar ) && srcVar.isValid() && fieldIsEditable( eww->fieldIdx() ) )
         {
           dst[eww->fieldIdx()] = srcVar;
         }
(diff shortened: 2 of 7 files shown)


src/gui/qgsattributeform.cpp
```diff
@@ -418,6 +419,66 @@ bool QgsAttributeForm::saveEdits()
...
+      QgsEditorWidgetWrapper *eww = qobject_cast<QgsEditorWidgetWrapper *>( ww );
+      if ( eww )
+      {
+        QVariant dstVar = dst.at( eww->fieldIdx() );
+        QVariant srcVar = eww->value();
+
+        // need to check dstVar.isNull() != srcVar.isNull()
+        // otherwise if dstVar=NULL and scrVar=0, then dstVar = srcVar
+        // be careful- sometimes two null qvariants will be reported as not equal!! (e.g., different types)
+        bool changed = ( dstVar != srcVar && !dstVar.isNull() && !srcVar.isNull() )
```
m-kuhn: Would `qgsVariantEqual` take care of all that?

src/gui/qgsattributeform.h
```diff
@@ -444,6 +450,12 @@ class GUI_EXPORT QgsAttributeForm : public QWidget
 
     QMap<QWidget *, QSvgWidget *> mIconMap;
 
+    //! dependency map for default values
```
m-kuhn:
```suggestion
    //! Dependency map for default values. Attribute index -> widget wrapper.
    //! Attribute indexes will be added multiple times if more than one widget depends on them.
```

tests/src/gui/testqgsattributeform.cpp
```diff
@@ -906,6 +907,107 @@ void TestQgsAttributeForm::testAttributeFormInterface()
...
+  //col1 - 10
+  //col2 - 15
+  //col3 - 15
+
+  QCOMPARE( ww0->value().toInt(), 5 );
+  QCOMPARE( ww1->value().toInt(), 10 );
+  QCOMPARE( ww2->value().toInt(), 15 );
+  QCOMPARE( ww3->value().toInt(), 15 );
+
+  /*
```
m-kuhn: What's up with this part of the code?

src/gui/qgsattributeform.cpp
```diff
@@ -1417,6 +1483,32 @@ void QgsAttributeForm::init()
...
   }
 
+  //create defaultValueDependencies
+  for ( QgsWidgetWrapper *ww : qgis::as_const( mWidgets ) )
+  {
+    QgsEditorWidgetWrapper *eww = qobject_cast<QgsEditorWidgetWrapper *>( ww );
+    if ( eww )
+    {
+      QgsExpression exp( eww->field().defaultValueDefinition().expression() );
+      for ( const QString &referencedColumn : exp.referencedColumns().toList() )
```
m-kuhn:
```suggestion
      const QList<QString> referencedColumns = exp.referencedColumns().toList();
      for ( const QString &referencedColumn : referencedColumns )
```

... or the list will detach. I wonder if you could directly iterate the QSet and avoid the conversion to a list too?

src/gui/qgsattributeform.cpp
signedav: It would care for the case that when a `NULL` value changes to 0, but for what is mentioned in the comment (I took it from `saveEdit` because there the behavior is the same) `qgsVariantEqual` does not care for the case when a null value does not change, but "two null qvariants will be reported as not equal." This would mean both are `isNull` but the comparison is not true. This could be solved by changing the `qgsVectorEqual` as this:
```
  return lhs.isNull() == rhs.isNull() && lhs == rhs || ( lhs.isNull()  && rhs.isNull() ) ;
```
But I'm not sure about it.

src/gui/qgsattributeform.cpp
signedav: yes

tests/src/gui/testqgsattributeform.cpp
signedav: Preparations for my later commit to test if there is a recursion.

src/gui/qgsattributeform.cpp
m-kuhn: I think `lhs.isNull == rhs.isNull ` already takes care of exactly this.

signedav: But the `&& lhs == rhs` makes it false again. When we pass two `NULL` values. the compared `isNull`s returns true, but are connected with the comparison that could be false (regarding to the comment).

What the `lhs.isNull == rhs.isNull` does is, that it returns false in the end in case we compare `0` with `NULL` what is nice...

m-kuhn: Just because of the comment or did you experience this?
In the end I think `qgsVariantCompare` was implemented exactly for cases like this, so it would be a pity if it's not usable here.

signedav: Just because the comment. That's why I'm not sure at all. But it should be same like in `saveEdits` and so I didn't want that I risk issues with this case...

m-kuhn: Let's see, the check in saveEdits was introduced in 2fddc0079f153c6dd7cb312c8de96548d9812094 (by @nyalldawson)
The method `qgsVariantEqual` later in 90e2c45502457cfc2fa8ea84b011b78d5b4764e1 (by @elpaso)

@nyalldawson / @elpaso, could the above patch be merged into `qgsVariantEqual`?

m-kuhn: Basically the question is, should we add a testcase for `QVERIFY( QVariant( QVariant::Int ) == QVariant( QVariant::Double ) )`. I'd say yes.

src/gui/qgsattributeform.cpp
```diff
@@ -418,6 +419,66 @@ bool QgsAttributeForm::saveEdits()
   return success;
 }
 
+bool QgsAttributeForm::updateDefaultValues( const int originIdx )
+{
+  if ( !mDefaultValueDependencies.contains( originIdx ) )
+    return false;
+
+  // create updated Feature
+  QgsFeature updatedFeature = QgsFeature( mFeature );
```
nyalldawson: won't we need QgsVectorLayerUtils::createFeature here?

src/gui/qgsattributeform.cpp
nyalldawson: +1 to all the above ;)

src/gui/qgsattributeform.cpp
```diff
@@ -418,6 +419,66 @@ bool QgsAttributeForm::saveEdits()
...
+        //do not update when when mMode is not AddFeatureMode and it's not applyOnUpdate
+        if ( mMode != QgsAttributeEditorContext::AddFeatureMode && !eww->field().defaultValueDefinition().applyOnUpdate() )
+        {
+          continue;
+        }
+
+        //do not update when this widget is already updating (avoid recursions)
+        if ( mAlreadyUpdatedFields.contains( eww->fieldIdx() ) )
+        {
+          qDebug() << "we don't update field [" << eww->fieldIdx() << "] " << eww->field().name() << " because it's already 'updating'";
```
nyalldawson: remove qDebug before merge

src/gui/qgsattributeform.cpp
m-kuhn: Won't that lead to a sudden increase for postgres serials with "enable default values server side" enabled?

nyalldawson: Quite possibly!

src/gui/qgsattributeform.cpp
signedav: Is it correct that `qgsVariantEqual` should return `true` in case of two `NULL` values from different types, but `false` if one of them is invalid?
Means these test are correct:
```
  QVERIFY( qgsVariantEqual( QVariant( QVariant::Int ), QVariant( QVariant::String ) ) );

  QVERIFY( !qgsVariantEqual( QVariant( QVariant::Int ), QVariant() ) );
```

With that it won't have an issue anymore on the formerly failing test for the `ValueRelationWidget` but what is happening here, is still a little bit weird https://github.com/qgis/QGIS/blob/master/src/gui/editorwidgets/qgsvaluerelationwidgetwrapper.cpp#L236
```
      QVariant v( mComboBox->itemData( i ) );
      if ( qgsVariantEqual( v, value ) )
      {
        idx = i;
        break;
      }
```
There it won't return `true` anymore if `value` is invalid, but still it wouldn't had returned `true` before also if `v` would not be recognized as `NULL` if `mComboBox->itemData( i )` returns `0`. So shouldn't here be set a type with e.g. `toInt()`?

src/core/qgis.cpp
```diff
@@ -28,6 +28,7 @@
 #include "qgsconfig.h"
 #include "qgslogger.h"
 #include "qgswkbtypes.h"
+#include <QDebug>
```
m-kuhn: That's a leftover, I guess?

signedav: Yes.

m-kuhn:
```suggestion
```

PR #31168, opened 2019-08-08
Do not create sAuthIdToQgisSrsIdMap for proj5 builds

PeterPetrik:
I hit the https://bugs.llvm.org/show_bug.cgi?id=42515 issue for cross compilation of OSGeo4A for android on mac...

### DIFF
--- a/src/core/qgscoordinatereferencesystem_legacy.h
+++ b/src/core/qgscoordinatereferencesystem_legacy.h
@@ -19,6 +19,7 @@
 #include <QMap>
 #include <QString>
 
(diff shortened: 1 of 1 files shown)


nyalldawson:
Could you test instead if using a std hash instead of qt fixes the problem?

Otherwise this is only a very short term fix. Everyone who can be using proj 6 should be now.

m-kuhn:
For reference, upstream issue: https://bugs.llvm.org/show_bug.cgi?id=42515

nyalldawson:
Merging as a short-term workaround for @PeterPetrik

PR #31154, opened 2019-08-08
Geopdf export (API)

nyalldawson:
API additions allowing GeoPDF export of print layouts. 

Requires GDAL 3 + (and no UI for now!)

PR #31175, opened 2019-08-09
Layout API groundwork for GeoPDF work

nyalldawson:
This PR splits out the self-contained API changes from #31154 (to avoid a mega PR forming) .


roya0045:
I have a question regarding the signification of this PR. From what I can remember PDF made with Qgis can be imported in Qgis and are properly georeferenced. Whereas GeoPDF seem to be a distinct (proprietary?) format.

Could you explain what this entails and what are the advantages compared to the typical georeferenced PDF that are made?

nyalldawson:
@roya45 currently, the pdfs created in qgis can be used as raster datasets. By contrast geospatial pdf/geopdf can be used as a vector format, storing the features and attributes themselves.

It has other benefits too - such as allowing individual layers to be toggled on and off in the pdf reader interactively (which works in almost all pdf readers, eg. Open source ones as well as acrobat reader)

src/core/qgsrenderedfeaturehandlerinterface.h
```diff
@@ -50,10 +50,19 @@ class CORE_EXPORT QgsRenderedFeatureHandlerInterface
...
     struct CORE_EXPORT RenderedFeatureContext
     {
-      ///@cond PRIVATE
-      // required to allow compilation only until real members are present
-      bool dummy; SIP_SKIP
-      ///@endcond
+
+      /**
+       * Constructor for RenderedFeatureContext.
+       * \param renderContext The render context which was used while rendering feature.
```
nirvn: I assume the pointer here isn't made to have its ownership transferred?

tests/src/core/testqgslayoutexporter.cpp
```diff
@@ -0,0 +1,260 @@
...
+  layerIds.clear();
+  layerNames.clear();
+  mapLayerIds.clear();
+
+  QgsLayoutItemLabel *label = new QgsLayoutItemLabel( &l );
+  items << label;
+  res = exporter.handleLayeredExport( items, exportFunc );
+  QCOMPARE( res, QgsLayoutExporter::Success );
+  QCOMPARE( layerIds, QList< unsigned int >() << 1 << 2 );
+  QCOMPARE( layerNames, QStringList() << QStringLiteral( "Pages" ) << QStringLiteral( "<Label>" ) );
```
nirvn: Is there a reason you enclose Label within lesser than / greater than signs, and not do so for Page(s)? I'd remove the symbols here (you also don't do it when naming layers containing multiple item types, e.g. "Label, Shape").

tests/src/core/testqgslayoutexporter.cpp
```diff
@@ -0,0 +1,260 @@
...
+
+  // multiple scalebars can be placed in the same layer
+  QgsLayoutItemScaleBar *scaleBar2 = new QgsLayoutItemScaleBar( &l );
+  items << scaleBar2;
+  QgsLayoutItemScaleBar *scaleBar3 = new QgsLayoutItemScaleBar( &l );
+  items << scaleBar3;
+  res = exporter.handleLayeredExport( items, exportFunc );
+  QCOMPARE( res, QgsLayoutExporter::Success );
+  QCOMPARE( layerIds, QList< unsigned int >() << 1 << 2 << 3 << 4 << 5 );
+  QCOMPARE( layerNames, QStringList() << QStringLiteral( "Pages" ) << QStringLiteral( "Labels, Shape" ) << QStringLiteral( "<Scalebar>" ) << QStringLiteral( "<Label>" ) << QStringLiteral( "Scalebars" ) );
```
nirvn: <> vs non <>.

tests/src/core/testqgslayoutexporter.cpp
```diff
@@ -0,0 +1,260 @@
...
+  // with an item which has sublayers
+  QgsVectorLayer *linesLayer = new QgsVectorLayer( TEST_DATA_DIR + QStringLiteral( "/lines.shp" ),
+      QStringLiteral( "lines" ), QStringLiteral( "ogr" ) );
+  QVERIFY( linesLayer->isValid() );
+  QgsVectorLayer *pointsLayer = new QgsVectorLayer( TEST_DATA_DIR + QStringLiteral( "/points.shp" ),
+      QStringLiteral( "points" ), QStringLiteral( "ogr" ) );
+  QVERIFY( pointsLayer->isValid() );
+
+  p.addMapLayer( linesLayer );
+  p.addMapLayer( pointsLayer );
```
nirvn: You are not using the pointsLayer anywhere after adding it to the project, did you mean to test multiple exported layer name IDs somewhere?

src/core/layout/qgslayoutitem.h
```diff
@@ -414,16 +414,60 @@ class CORE_EXPORT QgsLayoutItem : public QgsLayoutObject, public QGraphicsRectIt
...
 
+    /**
+     * Behavior of item when exporting to layered outputs.
+     * \since QGIS 3.10
+     */
+    enum ExportLayerBehavior
+    {
+      CanGroupWithAnyOtherItem, //!< Item can be placed on a layer with any other item (default behavior)
+      CanGroupWithItemsOfSameType, //!< Item can only be placed on layers with other items of the same type, but multiple items of this type can be grouped together
+      MustPlaceInOwnLayer, //!< Item must be placed in its own individual layer
```
nirvn: Should you edit the export layer behavior of 3D map items to be set to MustPlaceInOwnLayer ?

src/core/layout/qgslayoutexporter.cpp
```diff
@@ -1466,6 +1436,169 @@ bool QgsLayoutExporter::georeferenceOutputPrivate( const QString &file, QgsLayou
   return true;
 }
 
+QString nameForLayerWithItems( const QList< QGraphicsItem * > items, unsigned int layerId )
+{
+  if ( items.count() == 1 )
+  {
+    if ( QgsLayoutItem *layoutItem = dynamic_cast<QgsLayoutItem *>( items.at( 0 ) ) )
+      return layoutItem->displayName();
```
nirvn: I commented in your added tests on the less-than-ideal mixture of layer names using lesser than / greater than symbols vs. names not surrounded by those symbols. E.g. Page, Labels, scalebar, Map 1: lines, <Label>. IMO, it'd be worth fixing this if possible.

src/core/layout/qgslayoutexporter.cpp
```diff
@@ -1466,6 +1436,169 @@ bool QgsLayoutExporter::georeferenceOutputPrivate( const QString &file, QgsLayou
   return true;
 }
 
+QString nameForLayerWithItems( const QList< QGraphicsItem * > items, unsigned int layerId )
```
nirvn: Any reason why this function is not part of the QgsLayoutExporter class?

nirvn (commented):
@nyalldawson , looked at the code, looking good (as always). A few comments added.

tests/src/core/testqgslayoutexporter.cpp
nyalldawson: It's coming straight from the layout item "displayName" (i.e. the value shown in the layout items tree widget). By default this is of the format <Scalebar>, <Legend>, etc, UNLESS the item has an ID set, in which case it's the ID. Also, some item types override this, e.g. labels, which return just the first portion of the label's text.

So while in principal I agree that the mix looks messy, I can't think of a clean way to avoid this.

tests/src/core/testqgslayoutexporter.cpp
nyalldawson: Just a copy/paste error, I'll remove. Handling of multiple map layer's is tested in depth in testqgslayoutitemmap.cpp

src/core/layout/qgslayoutitem.h
nyalldawson: I don't think so -- these become images in the exports, so are easy to manage when in the same layer as other simple item types. I want to keep MustPlaceInOwnLayer only for complex items which have lots of parts in vector exports (like legends), and would be tricky to manage if they were mixed in a layer with other items.

src/core/layout/qgslayoutexporter.cpp
nyalldawson: It's just a little implementation detail -- I don't see any need to expose it to the class.

tests/src/core/testqgslayoutexporter.cpp
nirvn: Could you clean it up in the nameForLayerWithItems() function, whereas you could strip the first and last chars of a single item layer's displayName of those are < and > ?

src/core/layout/qgslayoutitem.h
nirvn: K; I was wondering whether it was worth harmonizing 3D vs 2D map items, i.e. if 2D is a layer of its own, 3D should follow that. But there was indeed no reason other than that. No strong feeling about this, perfectly fine for it to stay as is.

src/core/qgsrenderedfeaturehandlerinterface.h
nyalldawson: I swapped it to a const reference for clarity

src/core/layout/qgslayoutexporter.cpp
```diff
@@ -1441,7 +1441,13 @@ QString nameForLayerWithItems( const QList< QGraphicsItem * > items, unsigned in
   if ( items.count() == 1 )
   {
     if ( QgsLayoutItem *layoutItem = dynamic_cast<QgsLayoutItem *>( items.at( 0 ) ) )
-      return layoutItem->displayName();
+    {
+      QString name = layoutItem->displayName();
+      // cleanup default item ID format
+      if ( name.startsWith( '<' ) && name.endsWith( '>' ) )
+        name = name.mid( 1, name.length() - 2 );
```
nirvn: @nyalldawson , :tada:

PR #31178, opened 2019-08-09
[FEATURE] allow to define multiple field pairs for relations (composite keys)

3nids:
This is the first part of a work to allow using composite keys in relations.
This focuses on the UI for defining the relations.

1) Several field pairs can be listed in the discovery dialog

image

2) Several field pairs can be added in the add relation dialog

image

I had to flip the direction of the items to allow adding lines.


src/app/qgsrelationmanagerdialog.cpp
```diff
@@ -53,6 +53,13 @@ void QgsRelationManagerDialog::addRelation( const QgsRelation &rel )
   if ( ! rel.isValid() )
     return;
 
+  QString referencingFields, referencedFields;
+  for ( int i = 0; i < rel.fieldPairs().count(); i++ )
+  {
+    referencingFields.append( QStringLiteral( "%1%2" ).arg( referencingFields.isEmpty() ? "" : ", " ).arg( rel.fieldPairs().at( i ).referencingField() ) );
```
troopa81: Wouldn't be better to use intermediary QStringlist and join?

src/gui/attributetable/qgsdualview.cpp
```diff
@@ -73,7 +73,7 @@ QgsDualView::QgsDualView( QWidget *parent )
   auto createShortcuts = [ = ]( const QString & objectName, void ( QgsFeatureListView::* slot )() )
   {
     QShortcut *sc = QgsGui::shortcutsManager()->shortcutByName( objectName );
-    Q_ASSERT( sc ); // the shortcut must have been registered in the shortcuts manager
+    //Q_ASSERT( sc ); // the shortcut must have been registered in the shortcuts manager
```
troopa81: Is it normal to comment this? What happen if shortcut hasn't been registered

src/app/qgsrelationadddlg.cpp
```diff
@@ -14,79 +14,260 @@
...
+{
+  int index = mFieldPairWidgets.count();
+  QgsFieldPairWidget *w = new QgsFieldPairWidget( index, this );
+  w->setReferencingLayer( mReferencingLayerCombobox->currentLayer() );
+  w->setReferencedLayer( mReferencedLayerCombobox->currentLayer() );
+  mFieldPairsLayout->addWidget( w );
+  mFieldPairWidgets << w;
+
+  connect( w, &QgsFieldPairWidget::configChanged, this, &QgsRelationAddDlg::checkDefinitionValid );
+  connect( w, &QgsFieldPairWidget::pairDisabled, this, [ = ]() {emit fieldPairRemoved( w );} );
```
troopa81: Wouldn't be better to use pairAdded and pairRemoved names instead of enabled/disabled?

troopa81 (commented):
Nice work!

nyalldawson:
I'm a little confused -- why can you set multiple different **parent** layers here?

nyalldawson:
Ah - ignore my question. I misinterpreted the UI.

nyalldawson:
Am I correct in concluding that the core support for this already exists, it's just not exposed to users? (before this PR, that is)

src/gui/attributetable/qgsdualview.cpp
3nids: That's a left over.
The issue is that it was crashing when running the tests locally, no idea why it works on Travis.

3nids:
@nyalldawson yes everything is in core already. The only thing is the relation reference widget which is not ready to use them, but this comes in a following PR

src/gui/attributetable/qgsdualview.cpp
m-kuhn: Maybe a mac issue?

src/app/qgsrelationmanagerdialog.cpp
3nids: I used a simpler approach now, considering that 99.5% of the time we'll have only a single pair of fields.

PR #31144, opened 2019-08-07
Thread safety in duplicates check

m-kuhn:
It's dangerous to work on a pointer owned by a temporary.
Especially if the temporary is implicitly shared with a threaded cache.
Fix #30798
Fix #29470

PR #31188, opened 2019-08-10
[FEATURE] Add option to georeference PDFs and TIFs to save map canvas as image/PDF

nirvn:
## Description
Title says it all. This comes in quite handy when users need to quickly export georeferenced PDFs or TIFs to use with apps such as Avenza Maps et cie.

@nyalldawson , review appreciated.

## Checklist

### DIFF
--- a/src/core/qgsmaprenderertask.cpp
+++ b/src/core/qgsmaprenderertask.cpp
@@ -19,10 +19,14 @@
 #include "qgsannotationmanager.h"
 #include "qgsmaprenderertask.h"
 #include "qgsmapsettingsutils.h"
-
+#include "qgsogrutils.h"
+#include "qgslogger.h"
 #include <QFile>
 #include <QTextStream>
 #include <QPrinter>
+
+#include "gdal.h"
+#include "cpl_conv.h"
 
 QgsMapRendererTask::QgsMapRendererTask( const QgsMapSettings &ms, const QString &fileName, const QString &fileFormat, const bool forceRaster )
   : QgsTask( tr( "Saving as image" ) )
@@ -86,7 +90,8 @@ bool QgsMapRendererTask::run()
     printer->setOutputFormat( QPrinter::PdfFormat );
     printer->setOrientation( QPrinter::Portrait );
     // paper size needs to be given in millimeters in order to be able to set a resolution to pass onto the map renderer
-    printer->setPaperSize( mMapSettings.outputSize()  * 25.4 / mMapSettings.outputDpi(), QPrinter::Millimeter );
+    QSizeF outputSize = mMapSettings.outputSize();
+    printer->setPaperSize( outputSize  * 25.4 / mMapSettings.outputDpi(), QPrinter::Millimeter );
     printer->setPageMargins( 0, 0, 0, 0, QPrinter::Millimeter );
     printer->setResolution( mMapSettings.outputDpi() );
 
@@ -190,6 +195,25 @@ bool QgsMapRendererTask::run()
       QRectF rect( 0, 0, img.width(), img.height() );
       pp.drawImage( rect, img, rect );
       pp.end();
+
+      if ( mSaveWorldFile )
+      {
+        CPLSetThreadLocalConfigOption( "GDAL_PDF_DPI", QString::number( mMapSettings.outputDpi() ).toLocal8Bit().constData() );
+        gdal::dataset_unique_ptr outputDS( GDALOpen( mFileName.toLocal8Bit().constData(), GA_Update ) );
+        if ( outputDS )
+        {
+          double a, b, c, d, e, f;
+          QgsMapSettingsUtils::worldFileParameters( mMapSettings, a, b, c, d, e, f );
+          c -= 0.5 * a;
+          c -= 0.5 * b;
+          f -= 0.5 * d;
+          f -= 0.5 * e;
+          double geoTransform[6] = { c, a, b, f, d, e };
+          GDALSetGeoTransform( outputDS.get(), geoTransform );
+          GDALSetProjection( outputDS.get(), mMapSettings.destinationCrs().toWkt().toLocal8Bit().constData() );
+        }
+        CPLSetThreadLocalConfigOption( "GDAL_PDF_DPI", nullptr );
+      }
 #else
       mError = ImageUnsupportedFormat;
       return false;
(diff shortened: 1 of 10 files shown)


src/core/qgsmaprenderertask.cpp
```diff
@@ -190,6 +194,21 @@ bool QgsMapRendererTask::run()
       QRectF rect( 0, 0, img.width(), img.height() );
       pp.drawImage( rect, img, rect );
       pp.end();
+
+      if ( mSaveWorldFile )
+      {
+        CPLSetConfigOption( "GDAL_PDF_DPI", QString::number( mMapSettings.outputDpi() ).toLocal8Bit().constData() );
```
nyalldawson: You need to split this and save the QString::number( mMapSettings.outputDpi() ) result to a local variable before taking the data of it (otherwise it can point to a temporary location)

src/core/qgsmaprenderertask.cpp
```diff
@@ -210,14 +229,30 @@ bool QgsMapRendererTask::run()
...
+        {
+          gdal::dataset_unique_ptr outputDS( GDALOpen( mFileName.toLocal8Bit().constData(), GA_Update ) );
+          if ( outputDS )
+          {
+            double a, b, c, d, e, f;
+            QgsMapSettingsUtils::worldFileParameters( mMapSettings, a, b, c, d, e, f );
+            double geoTransform[] = { c, a, b, f, d, e };
+            GDALSetGeoTransform( outputDS.get(), geoTransform );
+            GDALSetProjection( outputDS.get(), mMapSettings.destinationCrs().toWkt().toLocal8Bit().constData() );
+          }
```
nyalldawson: else raise an error?

src/core/qgsmaprenderertask.cpp
```diff
@@ -210,14 +229,30 @@ bool QgsMapRendererTask::run()
...
+        if ( outputSuffix == QStringLiteral( "tif" ) || outputSuffix == QStringLiteral( "tiff" ) )
+        {
+          gdal::dataset_unique_ptr outputDS( GDALOpen( mFileName.toLocal8Bit().constData(), GA_Update ) );
+          if ( outputDS )
+          {
+            double a, b, c, d, e, f;
+            QgsMapSettingsUtils::worldFileParameters( mMapSettings, a, b, c, d, e, f );
+            double geoTransform[] = { c, a, b, f, d, e };
+            GDALSetGeoTransform( outputDS.get(), geoTransform );
+            GDALSetProjection( outputDS.get(), mMapSettings.destinationCrs().toWkt().toLocal8Bit().constData() );
```
nyalldawson: Again, save to a local variable before taking the data

src/core/qgsmapsettingsutils.h
```diff
@@ -39,6 +39,19 @@ class CORE_EXPORT QgsMapSettingsUtils
...
+     * \param mapSettings map settings
+     * \param a the a parameter
+     * \param b the b parameter
+     * \param c the c parameter
+     * \param d the d parameter
+     * \param e the e parameter
+     * \param f the f parameter
+     * \since QGIS 3.10
+     */
+    static void worldFileParameters( const QgsMapSettings &mapSettings, double &a, double &b, double &c, double &d, double &e, double &f );
```
nyalldawson: needs SIP_OUT

(also a unit test?)

src/core/qgsmaprenderertask.cpp
nirvn: Hmm, not sure about the error here. Maybe simply write a word file if the georeferencing embed fails instead?

src/core/qgsmaprenderertask.cpp
rouault: There might be multithreading issues in using CPLSetConfigOption() because it affects the whole process. Better use CPLSetThreadLocalConfigOption() to restrict the effect to the current thread.This remark holds for most uses of CPLSetConfigOption() in QGIS where one sets it and unsets it a bit afterwards

src/core/qgsmaprenderertask.cpp
```diff
@@ -190,6 +194,21 @@ bool QgsMapRendererTask::run()
...
       pp.end();
+
+      if ( mSaveWorldFile )
+      {
+        CPLSetConfigOption( "GDAL_PDF_DPI", QString::number( mMapSettings.outputDpi() ).toLocal8Bit().constData() );
+        gdal::dataset_unique_ptr outputDS( GDALOpen( mFileName.toLocal8Bit().constData(), GA_Update ) );
+        if ( outputDS )
+        {
+          double a, b, c, d, e, f;
+          QgsMapSettingsUtils::worldFileParameters( mMapSettings, a, b, c, d, e, f );
```
rouault: The use of worldFileParameters() (provided that this method does return values for a ESRI .wld !) is not really appropriate to set a GDAL geotransform. You did the coefficient order change right, but there's also a half-pixel difference in the origin between worldfile and GDAL geotransform conventions. See https://github.com/OSGeo/gdal/blob/master/gdal/gcore/gdal_misc.cpp#L1841 for the relationships between both

src/core/qgsmaprenderertask.cpp
rouault: > You need to split this and save the QString::number( mMapSettings.outputDpi() ) result to a local variable before taking the data of it (otherwise it can point to a temporary location)

That's not true because here it is the argument of a function (that would be different if constData() would be affected to a variable, and then passed to CPLSetConfigOption()). The C++ standard guarantees that all temporary variables will be destroyed afterwards.
The output of the following snippet
```
#include <iostream>

struct X
{
    X() { std::cout << "X()" << std::endl; }
    ~X() { std::cout << "~X()" << std::endl; }
    int v() { return 0; }
};

void foo(int x)
{
    std::cout << "foo()" << std::endl;
}

int main()
{
    std::cout << "right" << std::endl;
    foo(X().v());

    std::cout << "wrong" << std::endl;
    int x = X().v();
    foo(x);
    return 0;
};

```
is
```
X()
foo()
~X()
wrong
X()
~X()
foo()
```

nirvn:
@rouault , merci pour les tuyaux, very useful. Always good to have your eyes fly over gdal-related code.

src/core/qgsmaprenderertask.cpp
m-kuhn: @rouault I guess this does not hold true if the returned value is a pointer, correct?
Specifically clarifying with [#31144](https://github.com/qgis/QGIS/pull/31144/files) in mind.

```
feature.geometry().constGet()->centroid();
```

src/core/qgsmaprenderertask.cpp
nyalldawson: Ok, sounds reasonable

PR #30610, opened 2019-07-08
Fix multi-selection on value relation widget using string fields

signedav:
In the Value Relation Widget there is the `allowMulti` possibility allowing to store multiple reference values. On using string fields, they are stored on the widget as a formatted string like this: `{1,2,3}`

This fixes #30557

***Why not making conversion on provider side?***
During the improvement of using JSON fields in Postgresql and Geopackages there has been the possibility to pass the values as e.g. lists from the provider to the widget and back, without making any conversation. A provider reads JSON-fields as JSON and stored it back there in the end. The widget can use the `QVariant` (in this case `QVariantMap`) as it is. But in case the field is in the source only a string, the provider does not care if this string contains a list or not. It should not care, because it does not know what this string - however formatted - is used for. That's why it should pass the string as a string to the widget. The value relation widget decodes the formatted string to a list, uses it as a list, and in the end it needs to convert it back into a string, in case that it's not coming from a JSON field.

***Format***
The `{1,2,3}` list format is coming historically - I guess - from HSTORE. In case the provider recognizes an HSTORE (in PG), it possibly could convert it already to a data format and the widgets can use it like that \*. But on a text field, the provider does not have the competence to decide if this string is used as a list or as a string or as whatever on the widget side.

\* this is not the case. Hstore are only maps not lists. I have been missleaded because of old comments calling it hstore...

***GDAL Version***
Some tests are expecting JSON fields in Spatialite Files, what requests GDAL version >= 2.4.
That's why i made the `#ifdef`s there.

***Open Problems***
- [x] The test `testWithJsonInSpatialiteTextFk` will pretty sure fail, since it's not possible to handle key values containing `,` and `{` with the formatted strings. I guess this requests an improvement. -> solved

PR #31189, opened 2019-08-10
Backport 3.8:  Fix multi-selection on value relation widget using string fields

m-kuhn:
Manual backport of #30610


signedav:
Thanks @m-kuhn

src/core/fieldformatter/qgsvaluerelationfieldformatter.cpp
```diff
@@ -173,43 +174,50 @@ QStringList QgsValueRelationFieldFormatter::valueToStringList( const QVariant &v
...
-    {
-      newVal = QVariant( newVal.toString().trimmed().mid( 1 ).mid( 0, newVal.toString().length() - 2 ).prepend( '[' ).append( ']' ) );
-    }
-    if ( newVal.toString().trimmed().startsWith( '[' ) )
+    QVariantList valuesList;
+    if ( value.type() == QVariant::String )
     {
-      try
+      // This must be an array representation
+      auto newVal { value };
```
m-kuhn:
```suggestion
      QVariant newVal { value };
```

src/core/fieldformatter/qgsvaluerelationfieldformatter.cpp
```diff
@@ -173,43 +174,50 @@ QStringList QgsValueRelationFieldFormatter::valueToStringList( const QVariant &v
...
+            }
           }
         }
-      }
-      catch ( json::parse_error &ex )
-      {
-        qDebug() << QString::fromStdString( ex.what() );
+        catch ( json::parse_error &ex )
+        {
+          qDebug() << QString::fromStdString( ex.what() );
```
m-kuhn: @signedav is this something that can happen in real life? In this case wouldn't a message log entry be better?

src/core/qgspostgresstringutils.h
```diff
@@ -0,0 +1,65 @@
...
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+#ifndef QGSPOSTGRESSTRINGUTILS_H
+#define QGSPOSTGRESSTRINGUTILS_H
+
+#include "qgis_core.h"
+#include "qgis.h"
+#include "qlist.h"
```
m-kuhn:
```suggestion
#include <QList>
```

m-kuhn (commented):
@signedav looks safe to backport to me. Can you have a look at those suggestions (which I think would make sense on master too)

src/core/fieldformatter/qgsvaluerelationfieldformatter.cpp
signedav: Afaik, okay.

src/core/fieldformatter/qgsvaluerelationfieldformatter.cpp
signedav: Rarely but still possible. In the case that someone used json in string fields to fill up the valuerelation, what would be technically possible since 3.8 - and if this json strings are not formatted correctly.

I'm okay with log as well...

m-kuhn: Can you do a PR for this on master and we can backport it on top of this one?

signedav: #31286

Issue #31177, opened 2019-08-09
QGIS Server WMS getfeatureinfo does not work for arcgis mapserver layers


elpaso:
**Describe the bug**

WMS getfeatureinfo does not work for arcgis mapserver layers

**How to Reproduce**

Create a project with the following ARCGIS mapserver layers:

http://mapservices.pjm.com/arcgis/rest/services/Interregional/LMP/MapServer/

http://services.gis.ca.gov/arcgis/rest/services/GeoscientificInformation/Potential_Landslides/MapServer

Launch the QGIS Server to serve that project.

Open QGIS desktop (or use any other client).

Add the WMS layers served by the a.m. project

Try to get info on the WMS layers served by the QGIS Server WMS.

Actual result: empty getfeatureinfo
Expected result: list of field values from the WMS at the clicked point

**QGIS and OS versions**

Master

PR #31191, opened 2019-08-10
Fix server WMS cascading getfeatureinfo

elpaso:
Fix for #31177

### DIFF
--- a/src/server/services/wms/qgswmsrenderer.cpp
+++ b/src/server/services/wms/qgswmsrenderer.cpp
@@ -1174,7 +1174,7 @@ namespace QgsWms
     {
       bool validLayer = false;
       bool queryableLayer = true;
-      for ( QgsMapLayer *layer : layers )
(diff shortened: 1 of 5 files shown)


src/server/services/wms/qgswmsrenderer.cpp
```diff
@@ -1173,7 +1174,7 @@ namespace QgsWms
     {
       bool validLayer = false;
       bool queryableLayer = true;
-      for ( QgsMapLayer *layer : qgis::as_const( layers ) )
+      for ( QgsMapLayer *layer : layers )
```
m-kuhn: Was that changed on purpose?
https://github.com/qgis/QGIS/pull/31157/files

elpaso: nope, I didn't change it at all, because I did this work locally before your commit https://github.com/qgis/QGIS/commit/6fd35f6f1769534f3c5b70a4ee7ebb46eb6477c8, I should have rebased.

PR #31200, opened 2019-08-12
Fix unnecessary layer reloads triggering after project load, which also causes a leak

nyalldawson:


m-kuhn:
Where's the leak?

nyalldawson:
> Where's the leak?

Good question. It comes from QgsFields somehow(!).

m-kuhn:
And this fixes it?
In all cases or just masked through some code paths?

nyalldawson:
Valgrind no longer complains when loading a project. That's all i know :) 

To explain a bit further -- during project load the map canvases are all frozen, to avoid any unnecessary work. But the redrawAllLayers method DOES do work even while a canvas is frozen, because it triggers the reload method. So my main goal was to stop using this method so that we're not doing anything wasteful during project load.

I'm honestly unsure where/why the leak was occurring. If I had to guess, I'd say there's something here which is triggering an event loop during the layer reload method...

m-kuhn:
The idea is good in any case, it's just puzzling where this leak is...

nyalldawson:
Sorry - I think the leak was a red herring. I can't reproduce it now, and I suspect it was actually fixed by c56797448650c1cb8271a02c75f0e81b4d80d627

m-kuhn:
Good to know. Merge anyway.

PR #31209, opened 2019-08-12
Fix multilayered layout exports mess up label placement and other cross-layer dependant rendering

nyalldawson:
This adds a new map renderer job subclass, designed for step-by-step rendering of a map to different QPainters.

It exposes an API which lets callers iterate over parts of a map (e.g. individual map layers, labels), specifying a destination QPainter for each step in the iteration.

This is designed to improve the existing print layout export to layered output map rendering, where currently the maps are rendered by multiple renderer jobs, each only rendering one layer at a time.
This current approach of rendering individual layers are separate jobs is fundamentally broken (and cannot be fixed), because it prevents any handling of interactions between map layers (e.g. shifting
colliding labels).

The existing layout support for exporting to multi-layer SVG files (and in future multi-layered PDF files) has been upgraded to utilise this.

Now it's safe to create multi-layered layout exports without the results looking vastly different to the layout preview!

Sponsored by the GeoPDF funding group.


roya0045:
> This current approach of rendering individual layers are separate jobs is fundamentally broken (and cannot be fixed), because it prevents any handling of interactions between map layers (e.g. shifting
> colliding labels).

Would it be possible to do a first pass to detect elements, shift them and then use this approach to correct the placement of this or is this the most we can do and this will always generate suboptimal map layouts?

nyalldawson:
> Would it be possible to do a first pass to detect elements, shift them and then use this approach to correct the placement of this

Can you elaborate on this? It sounds like what the labelling engine already does, but it only works correctly when given ALL the information required to layout the labels. If we render each separate layer in its own unique render job, it has no knowledge of the features which will be labelled in the other jobs, how they'll interact with labels in this layer (eg obstacle settings), and it also breaks expression based rendering which uses functions like is_layer_visibile. That's why I think it's best to just do the render job exactly how it is done in the canvas and layout designer, but allow it to be done in stages so we can give each stage it's own surface to render onto.

roya0045:
I was simply curious if conflict resolution was actually done before rendering or if conflict resolution is done during rendering.

I was thinking that maybe will all the information that we have before rendering there might be a way to find optimal position and render everything instead of rendering sequentially and resolving conflicts at the rendering stage.

nyalldawson:
It's done mid-rendering 😆 basically it goes
1. Render layer's symbology, and register the features to be labelled with the labeling engine (this is dependent on the symbol rendering, because labeling considers symbol shapes while placing labels)
2. Calculate the optimal label placements for all registered features
3. Draw the labels

The current approach does steps 1-3 multiple times, once for each layer. In this new approach we do steps 1-3 once (for all layers together), but pause before each rendering step to request a new QPainter destination for that rendering step alone.

Basically now it's identical to canvas and non -layered layout exports, we are just swapping the paper out from under the pen as the cartographer is finished drawing each layer!

nirvn:
@nyalldawson , I've reviewed the code, and the only issue I found was that, well, it left me comment-less :wink: the logic is sound and the code elegant. 

BTW, this PR should really be titled 'The promise of multi-layer SVG export in QGIS fully realized".

PR #31210, opened 2019-08-12
 [needs-docs][layout] Give user explicit option to georeference (or not) PDF exports

nirvn:
## Description
@nyalldawson , as discussed, this PR adds a switch for the user to explicitly control whether an exported PDF will be georeferenced or not. 

I've also updated the 'Export RDF metadata' to add a few examples of what is being attached to the PDF. The idea here is to avoid people attaching identity fingerprint without realizing it via demystifying  the option :)

## Checklist

### DIFF
--- a/src/core/layout/qgslayoutexporter.cpp
+++ b/src/core/layout/qgslayoutexporter.cpp
@@ -532,9 +532,10 @@ QgsLayoutExporter::ExportResult QgsLayoutExporter::exportToPdf( const QString &f
   ExportResult result = printPrivate( printer, p, false, settings.dpi, settings.rasterizeWholeImage );
   p.end();
 
+  bool shouldAppendGeoreference = settings.appendGeoreference && mLayout && mLayout->referenceMap() && mLayout->referenceMap()->page() == 0;
   if ( settings.appendGeoreference || settings.exportMetadata )
   {
-    georeferenceOutputPrivate( filePath, nullptr, QRectF(), settings.dpi, settings.appendGeoreference, settings.exportMetadata );
+    georeferenceOutputPrivate( filePath, nullptr, QRectF(), settings.dpi, shouldAppendGeoreference, settings.exportMetadata );
   }
   return result;
 }
--- a/src/core/layout/qgslayoutexporter.cpp
+++ b/src/core/layout/qgslayoutexporter.cpp
@@ -532,10 +532,9 @@ QgsLayoutExporter::ExportResult QgsLayoutExporter::exportToPdf( const QString &f
   ExportResult result = printPrivate( printer, p, false, settings.dpi, settings.rasterizeWholeImage );
   p.end();
 
-  const bool shouldGeoreference = mLayout->pageCollection()->pageCount() == 1;
-  if ( shouldGeoreference || settings.exportMetadata )
+  if ( settings.appendGeoreference || settings.exportMetadata )
   {
-    georeferenceOutputPrivate( filePath, nullptr, QRectF(), settings.dpi, shouldGeoreference, settings.exportMetadata );
+    georeferenceOutputPrivate( filePath, nullptr, QRectF(), settings.dpi, settings.appendGeoreference, settings.exportMetadata );
   }
   return result;
 }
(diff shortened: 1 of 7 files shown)


src/core/layout/qgslayoutexporter.cpp
```diff
@@ -532,10 +532,9 @@ QgsLayoutExporter::ExportResult QgsLayoutExporter::exportToPdf( const QString &f
   ExportResult result = printPrivate( printer, p, false, settings.dpi, settings.rasterizeWholeImage );
   p.end();
 
-  const bool shouldGeoreference = mLayout->pageCollection()->pageCount() == 1;
```
nyalldawson: What about the page count restriction?

NathanW2:
Great idea on showing the metadata

src/core/layout/qgslayoutexporter.cpp
nirvn: @nyalldawson , I've exported a georeferenced mutli-page PDF here and it worked just fine. Is that restriction based on an obsolete bottleneck?

nyalldawson: It's more about it potentially being misleading. (Eg multi page export with many maps)

nirvn: @nyalldawson , OK, what about georeferencing multi-page PDFs only if the mLayout->referenceMap()->page() == 0 ?

nyalldawson: No strong feelings :)

nirvn: @nyalldawson , good, I've implemented that logic :) If a PDF can't be georeferenced (no reference map present or reference map isn't on the first page), the export dialog will reflect that by graying out the option:
Screenshot from 2019-08-12 16-00-36

Users will not always know what they are exporting beyond what's visible on their PDF viewers ;)

src/ui/layout/qgssvgexportoptions.ui
```diff
@@ -36,7 +36,7 @@
       <item row="2" column="0" colspan="2">
        <widget class="QCheckBox" name="mIncludeMetadataCheckbox">
         <property name="text">
-         <string>Export RDF metadata</string>
+         <string>Export RDF metadata (title, author, etc.)</string>
```
DelazJ: How about as a tooltip (allowing to be more verbose)? Keeping the label short as is would avoid us updating screenshots in docs. Also these information could be backported without affecting the UI.

nirvn: @DelazJ , I recently faced a situation where someone's safety was put at risk because he or she did not realize he/she was exporting author metadata. A tooltip is not visible enough.

Plus, it really only adds three word to a label, this isn't abusive by any definition of over-verbose :wink:

nirvn:
@nyalldawson , test case added.

src/ui/layout/qgssvgexportoptions.ui
DelazJ: Most of the options in this dialog are not self-explanatory and are showing long tooltip (btw, they could be worth wrapping on multiple lines). And if I were about to use an option I'm not sure of, I think checking whether there's a minimal explanation in tooltip should be the minimal step. No?
> Plus, it really only adds three word to a label, this isn't abusive by any definition of over-verbose 

Sure. By verbose, I meant you could add more information on which metadata is exported (or not) in the tooltip. And deprecates the screenshot.

nirvn: The screenshot here is already deprecated due to the additional checkbox added to control georeferencing.

I like your idea of adding the actual metadata strings into the tooltip, might do that later. That said, this recent security incident convinced the label needed a bit more context outright.

DelazJ: > The screenshot here is already deprecated due to the additional checkbox added to control georeferencing.

True, for PDF (!). But I agree the option should show the same label in both dialogs.
My actual concern is that we are "understaffed" for [screenshot updates](https://github.com/qgis/QGIS-Documentation/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+label%3AScreenshots+) in docs (in docs, _tout court_, but another story...) so generally speaking, if we can avoid GUI changes when it's not necessary...
My 2cts.

nirvn: @DelazJ , agreed on avoiding unnecessary burden. In this case, it's both doc and translation teams impacted. But in this case, the increased security through in-your-face clarification warrants this extra burden I'm afraid.

Valuable 2cts, keep those discussions alive :)

PR #31221, opened 2019-08-13
Optional "Toggle labels" layer tree widget

nyalldawson:
When enabled for a layer (via the Layer's Properties - Legend setting), this layer tree widget adds a checkbox to allow labels to be quickly toggled on and off for a layer.

Peek 2019-08-13 14-57

Also available as a plugin for existing QGIS versions: https://plugins.qgis.org/plugins/toggle_labels_widget/

### DIFF
--- a/python/core/auto_generated/qgsvectorlayer.sip.in
+++ b/python/core/auto_generated/qgsvectorlayer.sip.in
@@ -698,6 +698,8 @@ Returns whether the layer contains labels which are enabled and should be drawn.
 
 .. seealso:: :py:func:`setLabelsEnabled`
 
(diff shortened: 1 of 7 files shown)


m-kuhn:
Nice.

One thing though, would it be possible to visually discern the widgets from the legend better? E.g. decreased font size or different label or background color?

nirvn:
Are you up for making this even better? :) it could be quite useful to append a field/expression combobox to the widget to change the displayed label.

The UI would look like:
```
Label<newline>
[x] [ NAME ______|V]
```

Beyond the benefit of a quick label switch, it'd improve the UI in terms of avoiding the [x] show label checkbox being misinterpreted as a child layer.

PR #31223, opened 2019-08-13
Add an indicator in the status bar while a map redraw is occurring in the background

nyalldawson:
Otherwise there's no way to tell if a redraw is humming away in
the background, or is completed...


nirvn:
@nyalldawson , nice, most welcome UX improvement.

PR #31231, opened 2019-08-13
[FEATURE][expression] Optional format parameter for to_date, to_datetime, to_time functions

nirvn:
## Description

This PR adds an optional format parameter to the to_date(), to_datetime(), and to_time() to dramatically increase the parsing capability of those functions. For e.g., it is now possible to convert date strings such as 'June 29, 2019' to date objects.

## Checklist


resources/function_help/json/to_date
```diff
@@ -1,8 +1,13 @@
 {
   "name": "to_date",
   "type": "function",
-  "description": "Convert a string into a date object.",
-  "arguments": [ {"arg":"string","description":"string representing a date value"}],
-  "examples": [ { "expression":"to_date('2012-05-04')", "returns":"2012-05-04"}
+  "description": "Convert a string into a date object. An optional format string can be provided to parse the string; see QDate::fromString for additional documentation on the format.",
```
nyalldawson: "see QDate::fromString" -- that's kinda... useless... advice for most of our users :P Could we either link to the Qt docs directly here, or (Better) link to a similar page in the QGIS docs?

src/core/expression/qgsexpressionfunction.cpp
```diff
@@ -1019,7 +1019,18 @@ static QVariant fcnToString( const QVariantList &values, const QgsExpressionCont
...
 static QVariant fcnToDateTime( const QVariantList &values, const QgsExpressionContext *, QgsExpression *parent, const QgsExpressionNodeFunction * )
 {
-  return QVariant( QgsExpressionUtils::getDateTimeValue( values.at( 0 ), parent ) );
+  QString format = QgsExpressionUtils::getStringValue( values.at( 1 ), parent );
+  if ( format.isEmpty() )
+    return QVariant( QgsExpressionUtils::getDateTimeValue( values.at( 0 ), parent ) );
+
+  QString datetimestring = QgsExpressionUtils::getStringValue( values.at( 0 ), parent );
+  QDateTime datetime = QDateTime::fromString( datetimestring, format );
+  if ( !datetime.isValid() )
```
nyalldawson: What about empty string values? They shouldn't cause an evaluation error, but instead should return a null value without error.

tests/src/core/testqgsexpression.cpp
```diff
@@ -1269,6 +1269,9 @@ class TestQgsExpression: public QObject
       QTest::newRow( "time - time" ) << "to_time('08:30:00') - to_time('05:15:00')" << false << QVariant( QgsInterval( 3 * 60 * 60 + 15 * 60 ) );
       QTest::newRow( "epoch" ) << "epoch(to_datetime('2017-01-01T00:00:01+00:00'))" << false << QVariant( 1483228801000LL );
       QTest::newRow( "epoch invalid date" ) << "epoch('invalid')" << true << QVariant();
+      QTest::newRow( "date from formatted string" ) << "to_date('June 29, 2019','MMMM d, yyyy')" << false << QVariant( QDate( 2019, 6, 29 ) );
+      QTest::newRow( "datetime from formatted string" ) << "to_datetime('June 29, 2019 @ 11:00','MMMM d, yyyy @ HH:mm')" << false << QVariant( QDateTime( QDate( 2019, 6, 29 ), QTime( 11, 0 ) ) );
+      QTest::newRow( "time from formatted string" ) << "to_time('12:34:56','HH:mm:ss')" << false << QVariant( QTime( 12, 34, 56 ) );
```
nyalldawson: add some tests with bad string and format values too, and one with an empty string value

nyalldawson (commented):
Nice - very handy addition!

src/core/expression/qgsexpressionfunction.cpp
nirvn: @nyalldawson , I think we should stick to current behavior (predating this PR) whereas all of our `to_*()` functions return an error when passing on an empty string. That's actually the reason why I added `try()` a while back.

resources/function_help/json/to_date
nirvn: @nyalldawson , done, I've also taken the opportunity to update the format_date function help hyperlink pointing to an old 4.8 doc link.

PR #31234, opened 2019-08-13
Fix multi edit button not displayed for auxiliary storage attribute

troopa81:
## Description

The multi-edit button is not displayed when attribute form deal with field coming from auxiliary storage attribute.

aux_storage_and_multiedit

This PR fixes this. 

## Checklist

### DIFF
--- a/src/gui/qgsattributeform.cpp
+++ b/src/gui/qgsattributeform.cpp
@@ -2390,28 +2390,16 @@ void QgsAttributeForm::updateJoinedFields( const QgsEditorWidgetWrapper &eww )
 
 bool QgsAttributeForm::fieldIsEditable( int fieldIndex ) const
 {
-  bool editable = false;
-
   if ( mLayer->fields().fieldOrigin( fieldIndex ) == QgsFields::OriginJoin )
   {
     int srcFieldIndex;
     const QgsVectorLayerJoinInfo *info = mLayer->joinBuffer()->joinForFieldIndex( fieldIndex, mLayer->fields(), srcFieldIndex );
 
     if ( info && !info->hasUpsertOnEdit() && mMode == QgsAttributeEditorContext::AddFeatureMode )
-      editable = false;
-    else if ( info && info->isEditable() && info->joinLayer()->isEditable() )
-      editable = fieldIsEditable( *( info->joinLayer() ), srcFieldIndex, mFeature.id() );
+      return false;
   }
-  else
-    editable = fieldIsEditable( *mLayer, fieldIndex, mFeature.id() );
 
-  return editable;
-}
-
-bool QgsAttributeForm::fieldIsEditable( const QgsVectorLayer &layer, int fieldIndex,  QgsFeatureId fid ) const
-{
-  return !layer.editFormConfig().readOnly( fieldIndex ) &&
-         ( ( layer.dataProvider() && layer.dataProvider()->capabilities() & QgsVectorDataProvider::ChangeAttributeValues ) || FID_IS_NEW( fid ) );
+  return QgsVectorLayerUtils::fieldIsEditable( mLayer, fieldIndex, mFeature.id() );
 }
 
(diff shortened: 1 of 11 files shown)


src/gui/qgsattributeformeditorwidget.cpp
```diff
@@ -214,7 +215,19 @@ void QgsAttributeFormEditorWidget::updateWidgets()
 {
   //first update the tool buttons
   bool hasMultiEditButton = ( editPage()->layout()->indexOf( mMultiEditButton ) >= 0 );
-  bool fieldReadOnly = layer()->editFormConfig().readOnly( mWidget->fieldIdx() );
+
+  bool fieldReadOnly = false;
```
nyalldawson: This looks almost identical to QgsAttributeForm::fieldIsEditable. Could we make a common function for this instead to avoid the extra code?

troopa81: Indeed, and [here](https://github.com/qgis/QGIS/blob/master/src/gui/attributetable/qgsattributetablemodel.cpp#L767) also.

I added a method in QgsVectorLayerUtils. Hope it's the right place

src/gui/qgsattributeform.cpp
```diff
@@ -2204,28 +2204,16 @@ void QgsAttributeForm::updateJoinedFields( const QgsEditorWidgetWrapper &eww )
...
 bool QgsAttributeForm::fieldIsEditable( int fieldIndex ) const
 {
-  bool editable = false;
-
   if ( mLayer->fields().fieldOrigin( fieldIndex ) == QgsFields::OriginJoin )
   {
     int srcFieldIndex;
     const QgsVectorLayerJoinInfo *info = mLayer->joinBuffer()->joinForFieldIndex( fieldIndex, mLayer->fields(), srcFieldIndex );
 
     if ( info && !info->hasUpsertOnEdit() && mMode == QgsAttributeEditorContext::AddFeatureMode )
```
nyalldawson: Is this check needed in the new method too?

troopa81: Yes, in case info is null or not editable [here](https://github.com/troopa81/QGIS/blob/fix_multiedit_button_with_auxiliary_storage/src/core/qgsvectorlayerutils.cpp#L922).

Do you prefer several return? I just keep the method the way it was written in the first place.

nyalldawson: No, I mean I think it should be in QgsVectorLayerUtils::fieldIsEditable. If upsert isn't possible then the field can only be edited if the joined feature already exists, and this will impact the multi edit mode too.

troopa81: You mean like this f3f7a77 ?

So now, the multieditbutton is visible if all selected feature are not readOnly. That means that the visibility button change upon selected feature. Is that OK?

Once we agree on the behavior, I will add some unit test.

troopa81:
I  refactor `fieldIsEditable` in this PR, and added a `layer->isEditable()` condition to the field editable condition. It seems quite normal to me that a field is editable only if its layer is editable.

As a consequence a layer must be editable to update the default value (before this PR it was working without being editable). I fixed the test accordingly c67d011 

Can someone confirm this is the right approach?

nyalldawson:
> before this PR it was working without being editable

When would this actually occur?

troopa81:
> When would this actually occur?

I'm not sure it could happen from the application. From the test, it happens when you evaluate default definition, your layer need to be edited now. 

From the application, I try several things and it doesn't change anything. I try default value with joined layer, one edited and the one not, and there is still the same behaviour.

nyalldawson:
Ok, great.

I've removed the 3.4 backport tag -- I don't think the severity of this bug offsets the risk of introducing this much new logic into the LTR release.

troopa81:
Thanks @nyalldawson

Issue #31232, opened 2019-08-13
QGIS Crashes when deleting a map in a layout


maltaesousa:
Steps to reproduce:

1. Download this [QGIS project](https://sitn.ne.ch/web/qgis/TestCrashLayout.qgz) and open it
2. Open "Atlas" Layout
3. Delete the Map 1 object
4. Save Layout > Crash

Bug occurs in 3.9, 3.8.1, 3.6

Map 1 object has uuid `b1501156-b5af-4aef-9a3b-e7096050e640`
Editing the qgs file and deleting all the `<LayoutMultiFrame>` tags that have `mapUuid="{b1501156-b5af-4aef-9a3b-e7096050e640}"` attribute solves the issue.

```
Crash ID: 729143510354b98179f61ccfc688c8156fd82ddc

Stack Trace

QgsImageLegendNode::exportSymbolToJson :
QgsImageLegendNode::exportSymbolToJson :
QgsImageLegendNode::exportSymbolToJson :
QgsImageLegendNode::exportSymbolToJson :
QgsImageLegendNode::exportSymbolToJson :
QgsImageLegendNode::exportSymbolToJson :
QgsImageLegendNode::exportSymbolToJson :
QgsImageLegendNode::exportSymbolToJson :
QgsVectorLayerProperties::onAuxiliaryLayerDeleteField :
QgsVectorLayerProperties::onAuxiliaryLayerDeleteField :
QgsVectorLayerProperties::onAuxiliaryLayerDeleteField :
QgsVectorLayerProperties::onAuxiliaryLayerDeleteField :
QMetaObject::activate :
QAction::activate :
QgsVectorLayerProperties::onAuxiliaryLayerDeleteField :
QgsVectorLayerProperties::onAuxiliaryLayerDeleteField :
QgsVectorLayerProperties::onAuxiliaryLayerDeleteField :
QgsVectorLayerProperties::onAuxiliaryLayerDeleteField :
QMetaObject::activate :
QAction::activate :
QAbstractButton::click :
QAbstractButton::mouseReleaseEvent :
QToolButton::mouseReleaseEvent :
QWidget::event :
QApplicationPrivate::notify_helper :
QApplication::notify :
QgsImageLegendNode::exportSymbolToJson :
QCoreApplication::notifyInternal2 :
QApplicationPrivate::sendMouseEvent :
QSizePolicy::QSizePolicy :
QSizePolicy::QSizePolicy :
QApplicationPrivate::notify_helper :
[...]

PR #31244, opened 2019-08-14
[layout] Fix crash an attribute table item tries to access a destroyed map

nirvn:
## Description

This PR fixes issues #31232 -- crash casued by the attribute table item trying to access a destroyed reference map item when writing its properties to XML.

## Checklist


nirvn:
@nyalldawson , while fixing this crash, I noticed deleted attribute table items would still be saved into the project file. Is there a reason for that? It led to a crasher when saving a project with a deleted attribute table referring to a non-existing map here. I'm adding another commit to skip writing xml for empty multiframe.

nirvn:
I should add that beyond the empty attribute table multiframe item crash I referred about, saving those empty multiframe items that never get deleted means that a project file will drag those forever.

nyalldawson (approved):

PR #31248, opened 2019-08-14
fix build when QT defines QT_NO_PRINTER

PeterPetrik:
fix build when QT_NO_PRINTER is defined (e.g. ios architecture)

### DIFF
--- a/src/core/qgsmultirenderchecker.cpp
+++ b/src/core/qgsmultirenderchecker.cpp
@@ -119,6 +119,9 @@ QgsLayoutChecker::QgsLayoutChecker( const QString &testName, QgsLayout *layout )
 
 bool QgsLayoutChecker::testLayout( QString &checkedReport, int page, int pixelDiff, bool createReferenceImage )
 {
+#ifdef QT_NO_PRINTER
+  return false;
+#else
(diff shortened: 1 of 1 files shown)


src/core/qgsmultirenderchecker.cpp
```diff
@@ -119,6 +119,9 @@ QgsLayoutChecker::QgsLayoutChecker( const QString &testName, QgsLayout *layout )
 
 bool QgsLayoutChecker::testLayout( QString &checkedReport, int page, int pixelDiff, bool createReferenceImage )
 {
+#ifdef QT_NO_PRINTER
+  return false;
+#else
```
nyalldawson: Where is QPrinter used here?

PeterPetrik: The class QgsLayoutExporter is not defined at all with QT_NO_PRINTER, see the qgslayoutexporter.hpp

nyalldawson: Hm, ok. That seems odd to me, when only a few methods in QgsLayoutExporter use QPrinter. All the image and svg based exports should still work without issue.

PR #31249, opened 2019-08-14
[Backport release-3_4] Fix server WMS cascading getfeatureinfo

rldhont:
Manually backport #31191.

### DIFF
--- a/src/server/services/wms/qgswmsrenderer.cpp
+++ b/src/server/services/wms/qgswmsrenderer.cpp
@@ -1746,9 +1746,6 @@ namespace QgsWms
     }
 
     QgsRasterIdentifyResult identifyResult;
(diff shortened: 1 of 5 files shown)


src/server/services/wms/qgswmsrenderer.cpp
```diff
@@ -68,6 +68,8 @@
 #include "qgsdxfexport.h"
 #include "qgssymbollayerutils.h"
 #include "qgsserverexception.h"
+#include "qgsexpressioncontextutils.h"
```
rldhont: Remove this line in cc5d449

elpaso: I don't think that import is really necessary, I might have copy pasted it by mistake.

elpaso (approved):

PR #31257, opened 2019-08-15
Final tweaks to vertically oriented text / labels

nirvn:
## Description

@nyalldawson , curved placement isn't compatible with non horizontal orientation for a few reasons. When it comes to rotation-based orientation, that mode can result is a mixture of rotated and non-rotated characters for the same label (bad!); for the vertical orientation mode, since we break strings into graphemes, it ends up rotating 90 degrees along the curves, which is also nonsensical.

## Checklist

### DIFF
--- a/src/core/qgstextrenderer.cpp
+++ b/src/core/qgstextrenderer.cpp
@@ -3550,8 +3550,28 @@ void QgsTextRenderer::drawTextInternal( TextPart drawType,
 
       bool adjustForAlignment = alignment != AlignLeft && ( mode != Label || textLines.size() > 1 );
 
-      const auto constTextLines = textLines;
-      for ( const QString &line : constTextLines )
+      // apply some character replacement to draw symbols in vertical presentation
+      QString lines = textLines.join( '\n' );
+      lines = lines.replace( ',', QChar( 65040 ) ).replace( QChar( 8229 ), QChar( 65072 ) ); // comma & two-dot leader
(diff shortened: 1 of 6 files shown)


nirvn:
The second commit in this PR (and hopefully the final one) fixes rendering of a number of characters (parenthesis, ellipses, etc.) in vertically oriented text by substituting the characters with the vertical presentation counterparts.

Screenshot from 2019-08-16 11-04-23

src/core/qgstextrenderer.cpp
```diff
@@ -3550,8 +3550,28 @@ void QgsTextRenderer::drawTextInternal( TextPart drawType,
 
       bool adjustForAlignment = alignment != AlignLeft && ( mode != Label || textLines.size() > 1 );
 
-      const auto constTextLines = textLines;
-      for ( const QString &line : constTextLines )
+      // apply some character replacement to draw symbols in vertical presentation
```
nyalldawson: Might be nice to move this to QgsStringUtils

nyalldawson:
> The second commit in this PR (and hopefully the final one) fixes rendering of a number of characters (parenthesis, ellipses, etc.) in vertically oriented text by substituting the characters with the vertical presentation counterparts.

Nice! I was wondering if that would be an issue :)

src/core/qgstextrenderer.cpp
nirvn: @nyalldawson , good idea, will do.

PR #31258, opened 2019-08-15
Fix replacement of fill/stroke opacity in parameterized svgs with opa…

nyalldawson:
…city in style attributes


nirvn (approved):
Good catch.

PR #31271, opened 2019-08-16
fix qml loading for ios

PeterPetrik:


PeterPetrik:
if no complains, I would like to merge it later today..

PR #31280, opened 2019-08-17
Allow more actions from toolbar to be listed/managed in the customization dialog

### LINKED ISSUE #21923: "Add Circular String" is not deactivatable by Settings > Customization
Author Name: **Thomas Nogatz** (@TomNog)
Original Redmine Issue: [13901](https://issues.qgis.org/issues/13901)

Redmine category:customisation_framework

---
A possibility to deactivate the "Add Circular String" Button would be fine. Cannot find it in Settings > Customization > Toolbars > mDigitizeToolBar.

DelazJ:
This PR fixes #21923 based on the remark at https://github.com/qgis/QGIS/issues/21923#issuecomment-520326904..
Also add more items.

A before (from earlier version) vs after screenshot.
image

## Checklist

### DIFF
--- a/src/app/qgscustomization.cpp
+++ b/src/app/qgscustomization.cpp
@@ -763,12 +763,32 @@ void QgsCustomization::updateMainWindow( QMenu *toolBarMenu )
         for ( QAction *action : constActions )
         {
           if ( action->objectName().isEmpty() )
-          {
             continue;
-          }
+
           visible = mSettings->value( action->objectName(), true ).toBool();
           if ( !visible )
+          {
             tb->removeAction( action );
+            continue;
+          }
+
+          if ( visible && action->metaObject()->className() == QStringLiteral( "QWidgetAction" ) )
+          {
+            mSettings->beginGroup( action->objectName() );
+            QWidgetAction *widgetAction = qobject_cast<QWidgetAction *>( action );
+            QWidget *widget = widgetAction->defaultWidget();
+            QList<QAction *> childActions = widget->actions();
+            for ( QAction *wAction : childActions )
+            {
+              if ( wAction->objectName().isEmpty() )
+                continue;
+
+              visible = mSettings->value( wAction->objectName(), true ).toBool();
+              if ( !visible )
+                widget->removeAction( wAction );
+            }
+            mSettings->endGroup();
+          }
         }
         mSettings->endGroup();
       }
--- a/src/app/qgscustomization.cpp
+++ b/src/app/qgscustomization.cpp
@@ -881,7 +881,7 @@ void QgsCustomization::customizeWidget( const QString &path, QWidget *widget, QS
 
   // Qt may insert some internal classes in the tree, e.g. QTabWidgetPrivate inserts
   // qt_tabwidget_stackedwidget, such widgets do not appear in the tree generated
-  // from ui files and do not have sense from user poin of view -> skip
+  // from ui files and do not have sense from user point of view -> skip
 
   if ( !QgsCustomization::sInternalWidgets.contains( name ) )
   {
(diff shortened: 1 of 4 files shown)


src/app/qgscustomization.cpp
```diff
@@ -609,7 +610,27 @@ void QgsCustomization::createTreeItemToolbars()
     tbItem->setFlags( Qt::ItemIsEnabled | Qt::ItemIsUserCheckable | Qt::ItemIsSelectable );
     tbItem->setCheckState( 0, Qt::Checked );
 
-    addTreeItemActions( tbItem, tb->actions() );
+    QList<QWidgetAction *> tbWidgetActions = tb->findChildren<QWidgetAction *>( QString(), Qt::FindDirectChildrenOnly );
```
m-kuhn: Prefix this line with const

src/app/qgscustomization.cpp
```diff
@@ -742,12 +763,32 @@ void QgsCustomization::updateMainWindow( QMenu *toolBarMenu )
         for ( QAction *action : constActions )
         {
           if ( action->objectName().isEmpty() )
-          {
             continue;
-          }
```
m-kuhn: Off topic. We should start to enforce a single style for this via astyle. Half the people add those curly braces for one liners, the other half removes them...

src/app/qgscustomization.cpp
```diff
@@ -742,12 +763,32 @@ void QgsCustomization::updateMainWindow( QMenu *toolBarMenu )
...
+
           visible = mSettings->value( action->objectName(), true ).toBool();
           if ( !visible )
+          {
             tb->removeAction( action );
+            continue;
+          }
+
+          if ( visible && action->metaObject()->className() == QStringLiteral( "QWidgetAction" ) )
```
m-kuhn: It will always be visible here, given the above continue...

src/app/qgscustomization.cpp
```diff
@@ -742,12 +763,32 @@ void QgsCustomization::updateMainWindow( QMenu *toolBarMenu )
...
+          {
+            mSettings->beginGroup( action->objectName() );
+            QWidgetAction *widgetAction = qobject_cast<QWidgetAction *>( action );
+            QWidget *widget = widgetAction->defaultWidget();
+            QList<QAction *> childActions = widget->actions();
```
m-kuhn: Const

src/app/qgscustomization.cpp
DelazJ: Indeed. I was surprised to not get feedback from astyle.
What would be the rule? I just remembered an advice of this no-braces style from comments on a PR a while ago and followed it despite no mention is in the Dev guidelines.

src/app/qgscustomization.cpp
```diff
@@ -609,7 +610,27 @@ void QgsCustomization::createTreeItemToolbars()
...
+      QStringList actstrs;
+      actstrs << act->objectName() << act->text();
+      QTreeWidgetItem *item = new QTreeWidgetItem( tbItem, actstrs );
+      item->setIcon( 0, act->icon() );
+      item->setFlags( Qt::ItemIsEnabled | Qt::ItemIsUserCheckable | Qt::ItemIsSelectable );
+      item->setCheckState( 0, Qt::Checked );
+
+      QWidgetAction *widgetAction = qobject_cast<QWidgetAction *>( act );
+      QWidget *widget = widgetAction->defaultWidget();
+      QList<QAction *> childActions = widget->actions();
```
DelazJ: const here also?

m-kuhn: 1. always when the list is after used "read only"
2. especially when using a range based `for` loop

DelazJ:
I think I addressed all the requests, here.

DelazJ:
Thanks @m-kuhn

PR #31281, opened 2019-08-17
[style manager] Fetch remote style XML file using QgsNetworkContentFetcherTask

nirvn:
## Description
Advantages include:
- automatic handling of URL redirects
- use the user's authentication settings

@nyalldawson , as discussed a while back.

## Checklist

### DIFF
--- a/src/gui/symbology/qgsstyleexportimportdialog.cpp
+++ b/src/gui/symbology/qgsstyleexportimportdialog.cpp
@@ -417,7 +417,7 @@ void QgsStyleExportImportDialog::downloadStyleXml( const QUrl &url )
         mTempFile->close();
         populateStyles();
       }
-      delete progressDlg;
+      progressDlg->deleteLater();
     } );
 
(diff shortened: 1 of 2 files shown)


src/gui/symbology/qgsstyleexportimportdialog.cpp
```diff
@@ -387,79 +385,45 @@ void QgsStyleExportImportDialog::importFileChanged( const QString &path )
...
+                                    tr( "HTTP Error! Download failed: %1." ).arg( reply->errorString() ) );
+      }
+      else
+      {
+        mTempFile->write( reply->readAll() );
+        mTempFile->flush();
+        mTempFile->close();
+        populateStyles();
+      }
+      delete progressDlg;
```
nyalldawson: Might be safest to deleteLater here

nirvn: Done.

PR #31282, opened 2019-08-17
[needs-docs] Add shortcut to Add Delimited Text Layer

### LINKED ISSUE #27543: Add Keyboard Shortcut for the the 'Add Delimited Text Layer'
Author Name: **Devdatta Tengshe** (@devdattat)
Original Redmine Issue: [19718](https://issues.qgis.org/issues/19718)

Redmine category:actions

---
The 'Add Delimited Text Layer' button is the only one which does not have a keyboard shortcut, and that is the one Which I use the Most. We should add a Keyboard shortcut for this as well

---

- [qgis.png](https://issues.qgis.org/attachments/download/13205/qgis.png) (Devdatta Tengshe)

DelazJ:
Fix #27543 using `Ctrl+Shift+T`

## Checklist

### DIFF
--- a/src/ui/qgisapp.ui
+++ b/src/ui/qgisapp.ui
@@ -2482,6 +2482,9 @@ Acts on the currently active editable layer only.</string>
    <property name="toolTip">
     <string>Add Delimited Text Layer</string>
    </property>
+   <property name="shortcut">
+    <string>Ctrl+Shift+T</string>
+   </property>
   </action>
   <action name="mActionAddVirtualLayer">
    <property name="icon">


roya0045:
Is the need-doc tag still used? In my PR that changed the UI the tag was never added so I assume it is not needed, and the feature is still not in the doc, but its a 'work in progress' (until everything is merged).

DelazJ:
Until @m-kuhn grant proposal lands, i think it's worth adding [needs-docs] or [feature]. Otherwise, no way for writers to be aware of this change (or someone would manually create the issue there, and I don't want to be that guy).
Not sure I understand the case you mention but if your PR is not merged to the source code, then the feature does not exist yet and can not be documented.

roya0045:
@DelazJ The change to the GUI is implemented, but the feature was decomposed in many PRs and some are still pending.

nyalldawson:
This is another one I'd question the value of -- is it used so frequently that it needs an OOTB shortcut?

DelazJ:
I recall I've read people mentioning on forum their frequent (/massive?) use of csv-like files (the issue reporter seems to be among them). Is it for joins or the geometry? I can't say with certainty.
Since I had suggested the custom shortcut in the discussion and a year later, the issue is still kept open... Anyways... I was simply trying to fix "small issues" reported. No opinion. Feel free to close both if you think it's better.

nyalldawson:
Happy to defer to your logic/findings here @Delazj

PR #31283, opened 2019-08-17
[needs-docs] Remove duplicate 'Ctrl+Shift+M' shortcuts

### LINKED ISSUE #31183: Shortcuts to add layer do not work for some data providers
The shortcuts Crtl+Shift+2 (for DB2 layers) as well as Ctrl+Shift+M (for MSSQL) do not trigger the Data Source Manager dialog with appropriate tabs. Nothing happens when used.
Tested on Windows with 3.9

DelazJ:
Fix #31183 
`Ctrl+Shift+M` shortcut failed because it's shared by three commands all accessible from the main window:
  * by "New 3D Map View" whose close neighbour "New Map View" uses `Ctrl+M` (note that `Ctrl+Alt+M` is used by the "Processing Graphical Modeler" whose neighbors are also of `Ctrl+Alt+` type)
  * by "Measure Line" whose neighbors "Measure Area" has `Ctrl+Shift+J` and "Measure Angle" has nothing
  * by "Add MSSQL Layer" whose neigbors are of `Ctrl+Shift+` type

So here's a proposal:
- Keep `Ctrl+Shift+M` for ""Add MSSQL Layer"
- The "New 3D Map View" takes `Ctrl+Alt+M`
- Then "Processing Graphical Modeler" moved to `Ctrl+Alt+G` (I checked, it's not assigned by default)
- For the Measure Line, I go with `Ctrl+Shift+K` (if we really need a shortcut and since J is used for area). And Angle?

## Checklist

### DIFF
--- a/python/plugins/processing/ProcessingPlugin.py
+++ b/python/plugins/processing/ProcessingPlugin.py
@@ -213,11 +213,11 @@ class ProcessingPlugin:
 
         self.modelerAction = QAction(
             QgsApplication.getThemeIcon("/processingModel.svg"),
-            QCoreApplication.translate('ProcessingPlugin', 'Graphical &Modeler…'), self.iface.mainWindow())
+            QCoreApplication.translate('ProcessingPlugin', '&Graphical Modeler…'), self.iface.mainWindow())
         self.modelerAction.setObjectName('modelerAction')
         self.modelerAction.triggered.connect(self.openModeler)
         self.iface.registerMainWindowAction(self.modelerAction,
-                                            QKeySequence('Ctrl+Alt+M').toString(QKeySequence.NativeText))
+                                            QKeySequence('Ctrl+Alt+G').toString(QKeySequence.NativeText))
         self.menu.addAction(self.modelerAction)
 
         self.historyAction = QAction(
--- a/src/ui/qgisapp.ui
+++ b/src/ui/qgisapp.ui
@@ -1257,7 +1257,7 @@
     <string>Measure Line</string>
    </property>
    <property name="shortcut">
-    <string>Ctrl+Shift+M</string>
+    <string>Ctrl+Shift+K</string>
    </property>
   </action>
   <action name="mActionMeasureArea">
@@ -3060,7 +3060,7 @@ Acts on the currently active editable layer only.</string>
     <string>New 3D Map View</string>
    </property>
    <property name="shortcut">
-    <string>Ctrl+Shift+M</string>
+    <string>Ctrl+Alt+M</string>
    </property>
   </action>
   <action name="mActionShowLayoutManager">
--- a/src/ui/qgisapp.ui
+++ b/src/ui/qgisapp.ui
@@ -1257,7 +1257,7 @@
     <string>Measure Line</string>
    </property>
    <property name="shortcut">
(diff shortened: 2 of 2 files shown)


nyalldawson:
> Keep Ctrl+Shift+M for ""Add MSSQL Layer"
The "New 3D Map View" takes Ctrl+Alt+M
Then "Processing Graphical Modeler" moved to Ctrl+Alt+G (I checked, it's not assigned by default)
For the Measure Line, I go with Ctrl+Shift+K (if we really need a shortcut and since J is used for area). And Angle?

My gut feeling is that measure line would be used MUCH more than Add MSSQL layer, so I think it would be good to keep the more recognisable shortcut for measuring.

Honestly, my gut feeling is that adding layers via the source specific tabs wouldn't be done frequently enough to warrant an out-of-the-box shortcut?

DelazJ:
@nyalldawson If I rephrase you:
- "Measure Line" ->Ctrl+Shift+M
- "New 3D Map View" -> Ctrl+Alt+M
- "Processing Graphical Modeler" -> Ctrl+Shift+G
- "Add MSSQL Layer" -> no shortcut
- "Measure Angle" -> no suggestion/opinion

?

nyalldawson:
@DelazJ sounds good to me!

Graphical modeler is probably another one which doesn't really need an out-of-the-box shortcut. I doubt anyone is creating new models so frequently that they'd really need a shortcut for this.

roya0045:
I was wondering something about shortcuts, could it be possible to have a dedicated shortcut to copy and paste the symbology? It can be annoying to go through 3 menus everytime and this is the feature that I used far more than adding a specific provider layer (I usually use the browser to drag and drop).

I'm not sure if it would be a good idea to use the ctrl-c-ctrl-v but something akin to that would be nice and probably more useful than most shortcut.

DelazJ:
@roya0045 Do you mean a shortcut to deploy the menus and help you chosse the categories you would like to copy/paste or a shortcut to copy and paste the whole style of the layer? If latter, these actions are available in the "keyboard shortcuts" dialog so I guess you can set your own and use them. Btw, they are also accessible from layer menu.

roya0045:
> @roya0045 Do you mean a shortcut to deploy the menus and help you chosse the categories you would like to copy/paste or a shortcut to copy and paste the whole style of the layer? If latter, these actions are available in the "keyboard shortcuts" dialog so I guess you can set your own and use them. Btw, they are also accessible from layer menu.

I meant the latter, I was simply curious since the current shortcuts might be less frequently used than some other potential shortcuts.

As far as setting my own shortcut I have always looked for symbology instead of style, hence why I thought it didn't exist (but I had only looked once).

roya0045:
> @roya0045 Do you mean a shortcut to deploy the menus and help you chosse the categories you would like to copy/paste or a shortcut to copy and paste the whole style of the layer? If latter, these actions are available in the "keyboard shortcuts" dialog so I guess you can set your own and use them. Btw, they are also accessible from layer menu.

I meant the latter, I was simply curious since the current shortcuts might be less frequently used than some other potential shortcuts.

As far as setting my own shortcut I have always looked for symbology instead of style, hence why I thought it didn't exist (but I had only looked once).

PR #31285, opened 2019-08-19
[FEATURE][oracle] Database transaction support

nyalldawson:
Implements transaction support for the Oracle provider.


gioman:
>  the whole QGIS transaction support in general seems completely broken to me. (e.g. start editing in a project with transaction groups, move a feature, rollback changes, try to move a feature -> crash).

does not configure as a regression?

nyalldawson:
> does not configure as a regression?

My gut feeling is that you'll find this crash way back since transaction support was added - seems a design issue rather than a code issue

nyalldawson:
This is ready for merge/review now

PR #30991, opened 2019-07-30
topological editing: Fix Z for add feature

lbartoletti:
## Description

Topological editing does not work for Z geometries. NaN value is returned instead of the default Z value.

This PR fixes it.

Before
before

After
after

## Checklist

> Reviewing is a process done by project maintainers, mostly on a volunteer basis. We try to keep the overhead as small as possible and appreciate if you help us to do so by completing the following items. Feel free to ask in a comment if you have troubles with any of them.

- [X] Commit messages are descriptive and explain the rationale for changes
- [ ] Commits which fix bugs include `fixes #11111` in the commit message next to the description
- [ ] Commits which add new features are tagged with `[FEATURE]` in the commit message
- [ ] Commits which change the UI or existing user workflows are tagged with `[needs-docs]` in the commit message and contain sufficient information in the commit message to be documented
- [X] I have read the [QGIS Coding Standards](https://docs.qgis.org/testing/en/docs/developers_guide/codingstandards.html) and this PR complies with them
- [ ] This PR passes all existing unit tests (test results will be reported by travis-ci after opening this PR)
- [X] New unit tests have been added for core changes
- [X] I have run [the `scripts/prepare-commit.sh` script](https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTING.md#contributing-to-qgis) before each commit

PR #31163, opened 2019-08-08
Topological editing: fix vertex tool (Z support)

lbartoletti:
## Description

Fix topological point and add z support.

The Z snapped on a segment is always the default Z. I will correct this behavior in a dedicated PR.

Before
before

After
after

Sponsored by QWAT group @ponceta

## Checklist

- [X] Commit messages are descriptive and explain the rationale for changes
- [ ] Commits which fix bugs include `fixes #11111` in the commit message next to the description
- [ ] Commits which add new features are tagged with `[FEATURE]` in the commit message
- [ ] Commits which change the UI or existing user workflows are tagged with `[needs-docs]` in the commit message and contain sufficient information in the commit message to be documented
- [X] I have read the [QGIS Coding Standards](https://docs.qgis.org/testing/en/docs/developers_guide/codingstandards.html) and this PR complies with them
- [X] New unit tests have been added for core changes
- [X] I have run [the `scripts/prepare-commit.sh` script](https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTING.md#contributing-to-qgis) before each commit

## Backport

- [x] I request the backport of the changes to the Latest Release
- [x] I request the backport of the changes to the Long Term Release

PR #31035, opened 2019-08-01
topological editing: fix topological editing for reshape map tool

lbartoletti:
## Description

I continue the work on topological editing. Today: reshape map tool.  

Before
before

After
after

## Checklist

> Reviewing is a process done by project maintainers, mostly on a volunteer basis. We try to keep the overhead as small as possible and appreciate if you help us to do so by completing the following items. Feel free to ask in a comment if you have troubles with any of them.

- [X] Commit messages are descriptive and explain the rationale for changes
- [ ] Commits which fix bugs include `fixes #11111` in the commit message next to the description
- [ ] Commits which add new features are tagged with `[FEATURE]` in the commit message
- [ ] Commits which change the UI or existing user workflows are tagged with `[needs-docs]` in the commit message and contain sufficient information in the commit message to be documented
- [X] I have read the [QGIS Coding Standards](https://docs.qgis.org/testing/en/docs/developers_guide/codingstandards.html) and this PR complies with them
- [X] This PR passes all existing unit tests (test results will be reported by travis-ci after opening this PR)
- [X] New unit tests have been added for core changes
- [X] I have run [the `scripts/prepare-commit.sh` script](https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTING.md#contributing-to-qgis) before each commit

PR #31289, opened 2019-08-19
add interpolate Z on segment snapping

lbartoletti:
## Description

Snap on a segment in Z layer returns always the Z default value (except for some tools like trim/extend which uses Z interpolation). This is not correct. 

Before
before

After
after

Sponsored by QWAT group @ponceta

## Checklist

### DIFF
--- a/src/gui/qgsmaptoolcapture.cpp
+++ b/src/gui/qgsmaptoolcapture.cpp
@@ -397,18 +397,19 @@ int QgsMapToolCapture::fetchLayerPoint( const QgsPointLocator::Match &match, Qgs
       if ( !f.geometry().vertexIdFromVertexNr( match.vertexIndex(), vId ) )
         return 2;
 
+      const QgsGeometry geom( f.geometry() );
       if ( QgsProject::instance()->topologicalEditing() && match.hasEdge() )
       {
         QgsVertexId vId2;
         if ( !f.geometry().vertexIdFromVertexNr( match.vertexIndex() + 1, vId2 ) )
           return 2;
-        QgsLineString line( f.geometry().constGet()->vertexAt( vId ), f.geometry().constGet()->vertexAt( vId2 ) );
+        QgsLineString line( geom.constGet()->vertexAt( vId ), geom.constGet()->vertexAt( vId2 ) );
 
         layerPoint = QgsGeometryUtils::closestPoint( line,  QgsPoint( match.point() ) );
       }
       else
       {
-        layerPoint = f.geometry().constGet()->vertexAt( vId );
+        layerPoint = geom.constGet()->vertexAt( vId );
       }
 
       // ZM support depends on the target layer
--- a/src/gui/qgsmaptoolcapture.cpp
+++ b/src/gui/qgsmaptoolcapture.cpp
@@ -383,7 +383,7 @@ int QgsMapToolCapture::fetchLayerPoint( const QgsPointLocator::Match &match, Qgs
 {
   QgsVectorLayer *vlayer = qobject_cast<QgsVectorLayer *>( mCanvas->currentLayer() );
   QgsVectorLayer *sourceLayer = match.layer();
-  if ( match.isValid() && match.hasVertex() && sourceLayer &&
+  if ( match.isValid() && ( match.hasVertex() || match.hasEdge() ) && sourceLayer &&
        ( sourceLayer->crs() == vlayer->crs() ) )
   {
     QgsFeature f;
@@ -396,7 +396,14 @@ int QgsMapToolCapture::fetchLayerPoint( const QgsPointLocator::Match &match, Qgs
       if ( !f.geometry().vertexIdFromVertexNr( match.vertexIndex(), vId ) )
         return 2;
 
-      layerPoint = f.geometry().constGet()->vertexAt( vId );
+      if ( match.hasEdge() )
+      {
+        layerPoint = match.interpolatedPoint();
+      }
+      else
+      {
+        layerPoint = f.geometry().constGet()->vertexAt( vId );
+      }
 
       // ZM support depends on the target layer
       if ( !QgsWkbTypes::hasZ( vlayer->wkbType() ) )
--- a/src/gui/qgsmaptoolcapture.cpp
+++ b/src/gui/qgsmaptoolcapture.cpp
@@ -398,7 +398,12 @@ int QgsMapToolCapture::fetchLayerPoint( const QgsPointLocator::Match &match, Qgs
 
       if ( match.hasEdge() )
       {
-        layerPoint = match.interpolatedPoint();
+        QgsVertexId vId2;
+        if ( !f.geometry().vertexIdFromVertexNr( match.vertexIndex() + 1, vId2 ) )
+          return 2;
+        QgsLineString line( f.geometry().constGet()->vertexAt( vId ), f.geometry().constGet()->vertexAt( vId2 ) );
+
+        layerPoint = QgsGeometryUtils::closestPoint( line,  QgsPoint( match.point() ) );
       }
       else
       {
--- a/src/gui/qgsmaptoolcapture.cpp
+++ b/src/gui/qgsmaptoolcapture.cpp
@@ -32,6 +32,7 @@
 #include "qgssettings.h"
 #include "qgsapplication.h"
 #include "qgsadvanceddigitizingdockwidget.h"
+#include "qgsproject.h"
 
 #include <QAction>
 #include <QCursor>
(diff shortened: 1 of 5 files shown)


src/gui/qgsmaptoolcapture.cpp
```diff
@@ -396,7 +396,14 @@ int QgsMapToolCapture::fetchLayerPoint( const QgsPointLocator::Match &match, Qgs
       if ( !f.geometry().vertexIdFromVertexNr( match.vertexIndex(), vId ) )
         return 2;
 
-      layerPoint = f.geometry().constGet()->vertexAt( vId );
+      if ( match.hasEdge() )
+      {
+        layerPoint = match.interpolatedPoint();
```
nyalldawson: Why re-fetch the geometry here (inside interpolatedPoint), when you've already retrieved it by this stage?

lbartoletti: Since, it's retrieved in 2D only when match hasEdge. But maybe, I'm missing something - and that would explain why it works when digitinzing into the app and why some tests pass when they should fail. Did you see a better way to catch the Z value?

m-kuhn:
That is a good thing for many cases. But I wonder if we should always assume linear interpolation along lines? This seems a bit adventurous...

m-kuhn:
Also on moving the point to a new snapping target it won't be updated any more. I have a gut feeling that we need to be able to configure the default value for Z a bit more precisely (similar to the default values for attributes).

src/gui/qgsmaptoolcapture.cpp
nyalldawson: > Since, it's retrieved in 2D only when match hasEdge.

I mean, you've got f.geometry() already. It's used in the not edge case. So it's rather inefficient to grab this again, when you could just do the interpolation on f.geometry() directly...

lbartoletti: Oh, yeah, right. I'll fix it. thanks :)

lbartoletti:
> But I wonder if we should always assume linear interpolation along lines? This seems a bit adventurous...

I asked myself the same question. Maybe we need to add an option to configure the default Z.

- Use the interpolated value if there is a snap, otherwise the default Z
- Always use the default Z

m-kuhn:
I could imagine even more options (e.g. take default Z from a elevation raster) or update Z on geometry edit.

Apart from that, the default Z really is something that needs to be rethought too. It should be a data provider option (similar to NoData for rasters) instead of a system setting.

ponceta:
> That is a good thing for many cases. But I wonder if we should always assume linear interpolation along lines? This seems a bit adventurous...

When discussing about that, we assumed that when and ONLY when Topology is activated in that case, we could assume to want the interpolated value.

lbartoletti:
> > 
> 
> When discussing about that, we assumed that when and ONLY when Topology is activated in that case, we could assume to want the interpolated value.

@m-kuhn What do you think about Arnaud's proposal?

Personnaly, I live well with linear interpolation when I snap to a segment.

> the default Z really is something that needs to be rethought too

Yes, including raster options, and other idea maybe.

m-kuhn:
Linear interpolation with topology activated works for me too.

I still think it's a weak connection with some open questions (what happens if I add a new vertex on an existing line by moving the centerpoint? since we lose the information which ones are interpolated we also cannot update when they are moved down the road because they have been incorrectly placed the first time, side effects like setting a "vertical accuracy" flag cannot be controlled). So please keep the possibility to add this as a new configuration option in the back of your head in case someone reports issues with more complex workflows.

I would:

- not backport it, that's the kind of changes that are hard to track down and a behaviour you don't want to suddenly have changed in a patch release because it can be deemed data corruption
- clearly document it (but I guess you planned to already)

src/gui/qgsmaptoolcapture.cpp
```diff
@@ -396,7 +397,19 @@ int QgsMapToolCapture::fetchLayerPoint( const QgsPointLocator::Match &match, Qgs
...
+        QgsVertexId vId2;
+        if ( !f.geometry().vertexIdFromVertexNr( match.vertexIndex() + 1, vId2 ) )
+          return 2;
+        QgsLineString line( f.geometry().constGet()->vertexAt( vId ), f.geometry().constGet()->vertexAt( vId2 ) );
+
+        layerPoint = QgsGeometryUtils::closestPoint( line,  QgsPoint( match.point() ) );
+      }
+      else
+      {
+        layerPoint = f.geometry().constGet()->vertexAt( vId );
```
m-kuhn: That's risky, it's using the pointer of a temporary of a shared object. Take a copy of geometry at line 399.

lbartoletti:
Is this PR still considered as frozen? Thanks

nyalldawson:
Hmm - I originally tagged it as a feature, but on second thoughts, I think you could make a valid case that it's actually a bug fix.

nyalldawson:
I'll leave the approval in @m-kuhn's hands given that he's already reviewed

m-kuhn:
What makes me hesitate is the fact that this could break valid use cases (working in topology mode and not wanting estimated Z values in the data), is this too much of an edge case to be considered for a merge-in-freeze?

lbartoletti:
Well, this PR is part of all my other PRs to fix the snapping on the Z and the topology. 

With your previous comment, I made sure that the interpolation was only done during the topological editing. For me it's a bug fix.

As we discussed before, we can open another PR to choose the behavior of Z during editing (default Z, interpolation on a segment, value of a raster). But this is a feature.

m-kuhn:
The thing is, that *derived* values are saved indistinguishable from *measured* values, transparently. The question I have is, if we *really* want that without any way to opt in or out.

Let's take for example the QGEP datamodel. There are the following value lists:

*reach_point_elevation_accuracy*
- more_than_6cm
- plusminus_1cm
- plusminus_3cm
- plusminus_6cm
- unkown

*reach_elevation_determination*
- accurate
- unknown
- inaccurate

What I am afraid of is, that one would need to take care of the required business logic of keeping these project specific attributes up-to-date.
From what I can see, there is no API in place to deal with that (which would so far have been possible database side or by hooking into signals inside the QGIS API). So we are effectively loosing control and possibilities.

We do not personally have a project running right now, where this would be an issue. So it's ok for me. But I think we need to be fully conscious about these implications of this. This question is about mergability without any opt-out at all (and not about mergability during freeze period). If it results in being mergable, then no worries to merge it during freeze.

lbartoletti:
> The thing is, that _derived_ values are saved indistinguishable from _measured_ values, transparently. The question I have is, if we _really_ want that without any way to opt in or out.

I can agree, and as said before, we can work on adding this option (directly or via a QEP?) -- but after this PR :wink: . For now, thise cases should be handled in the DB API as we already done with QGEP/QWAT when Z wasn't very well supported in QGIS.

> We do not personally have a project running right now, where this would be an issue

Not directly related, but we may need to get the Z from a raster on some project. 

> So it's ok for me. But I think we need to be fully conscious about these implications of this.

So we agree and we must prepare this option.
We will contact each other to define this need and propose it to the community.

Thank you Matthias.

m-kuhn (approved):
Works for me as an intermediate solution.

A follow up simple option "Linear Z interpolation on snapping to edges" (linked or not to topology) would be much appreciated to restore flexibility. Any other advanced option to control Z can be added on top of that.

lbartoletti:
@m-kuhn thank you.

lbartoletti:
> A follow up simple option "Linear Z interpolation on snapping to edges" (linked or not to topology) would be much appreciated to restore flexibility. Any other advanced option to control Z can be added on top of that.

and ok for me too ;)

lbartoletti:
@m-kuhn @nyalldawson Is this PR can be merged, since it's a bugfix? Or it will be to the next version?

nyalldawson:
It will need to wait for 3.10.1 now -- 3.10.0 is in "hard freeze"

haubourg:
@nyalldawson I'm lost. This is a feature and we can merge it to a bugfix release ?

nyalldawson:
See earlier discussion - this is tagged as a bug

haubourg:
> See earlier discussion - this is tagged as a bug

Okay

m-kuhn:
-1 for this procedure

See my earlier [comment](https://github.com/qgis/QGIS/pull/31289#issuecomment-530227701)

> that's the kind of changes that are hard to track down and a behaviour you don't want to suddenly have changed in a patch release because it can be deemed data corruption

I am in favor of merging it now or postponing for 3.12
This will allow to communicate it clearly in a changelog and allow our power users to design 3.10 based systems on a stable featureset.

ponceta:
If I may add my comment here, this is the last part of a series of fixes :

https://github.com/qgis/QGIS/pull/30991
https://github.com/qgis/QGIS/pull/31020
https://github.com/qgis/QGIS/pull/31035
https://github.com/qgis/QGIS/pull/31142
https://github.com/qgis/QGIS/pull/31163

Which have been merged towards QGIS 3.4.10 LTR to enhance topological issues due to QGIS inconsistent behaviour between edition tools.

I do agree that the timing is bad but the delivery time will go from november 2019 to november 2020, without saying that we will have to work with a QGIS LTR version with these identified issues for over a year and a half.

These changes were approved and then frozen. I'm not very familiar with this but on the user perspective, this seems a bit akward.

lbartoletti:
> I am in favor of merging it now or postponing for 3.12

As we have already discussed, and given Arnaud's message, I would like to see it merged now; it could have been last week. For me, postponing it to 3.12 is not consistent with all the other bugfixes made on the subject in recent months.

Thank you.

PR #31293, opened 2019-08-19
Fix nlohman::json array list initializer broken on GCC

elpaso:
the issue:

`json j { something_returning_json_array_of_values() };`

produces a `json::array( json::array( values ) ) )`

instead of `json::array( values )`

by using the assignment all is fine:

`json j = something_returning_json_array_of_values();`

produces a `json::array( values )` as expected

### DIFF
--- a/src/core/qgsjsonutils.cpp
+++ b/src/core/qgsjsonutils.cpp
@@ -235,10 +235,9 @@ json QgsJsonExporter::exportFeaturesToJsonObject( const QgsFeatureList &features
     { "type", "FeatureCollection" },
     { "features", json::array() }
   };
-  const auto constFeatures = features;
-  for ( const QgsFeature &feature : constFeatures )
+  for ( const QgsFeature &feature : qgis::as_const( features ) )
   {
-    data[ "features" ].push_back( exportFeatureToJsonObject( feature ) );
+    data["features"].push_back( exportFeatureToJsonObject( feature ) );
   }
   return data;
 }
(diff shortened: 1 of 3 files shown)


elpaso:
@m-kuhn I think you've already seen this this bug recently.

src/core/qgsjsonutils.cpp
```diff
@@ -238,7 +238,7 @@ json QgsJsonExporter::exportFeaturesToJsonObject( const QgsFeatureList &features
   const auto constFeatures = features;
   for ( const QgsFeature &feature : constFeatures )
   {
-    data["features"].push_back( exportFeatureToJsonObject( feature ) );
+    data[ "features" ].push_back( exportFeatureToJsonObject( feature ) );
```
m-kuhn: Since this is a cosmetic one, can you use QStringLiteral on "features" and drop 238 in favor of `qgis::as_const`? ;)

m-kuhn:
@signedav you looked into it in more detail, what do you think?

src/core/qgsjsonutils.cpp
elpaso: no literal here, it's a std::strings, ok for the as_const

signedav:
I confirm what @elpaso writes regarding the json.

Like the observation I made during debugging and wrote down here https://github.com/qgis/QGIS/pull/30758:
> Apparently `{ QgsJsonUtils::jsonFromVariant( value ) }` returns an array containing one element and that's an object while `= QgsJsonUtils::jsonFromVariant( value )` contains the object itself not nested in an array.

elpaso:
@signedav just to be sure, which compiler are you using? Because I was not able to reproduce this on clang, but only on GCC.

signedav:
GCC as well.

elpaso:
no need to backport

PR #31298, opened 2019-08-19
Remove QGIS version check

m-kuhn:
The current way QGIS checks the version is not very convincing and there are a couple of issues with it. Very often it targets the wrong audience in bigger organisations, the people who see it do not even have the possibility to upgrade. There is also no distinction between regular releases and LTR.

So I figured it might be good to just close this chapter and remove it and maybe think of a smarter way for this in the future.

Critique and better ideas explicitly encouraged!

### DIFF
--- a/src/app/main.cpp
+++ b/src/app/main.cpp
@@ -536,7 +536,6 @@ int main( int argc, char *argv[] )
 
   bool myHideSplash = false;
   bool settingsMigrationForce = false;
-  bool mySkipVersionCheck = false;
   bool hideBrowser = false;
 #if defined(ANDROID)
   QgsDebugMsg( QStringLiteral( "Android: Splash hidden" ) );
@@ -625,10 +624,6 @@ int main( int argc, char *argv[] )
         else if ( arg == QLatin1String( "--version-migration" ) )
         {
           settingsMigrationForce = true;
-        }
-        else if ( arg == QLatin1String( "--noversioncheck" ) || arg == QLatin1String( "-V" ) )
-        {
-          mySkipVersionCheck = true;
         }
         else if ( arg == QLatin1String( "--noplugins" ) || arg == QLatin1String( "-P" ) )
         {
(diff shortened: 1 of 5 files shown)


andreasneumann:
And we also now have the "News" section on the startscreen where we can announce the availability of new versions.

nyalldawson:
I'm -1 to this. It's used a lot by end users, and the news feed doesn't replace it (just augments it). 

There's already settings to disable this which can be enabled in enterprise rollouts or in downstream linux distros, why can't they be used instead?

nirvn:
I'm -0.5 to remove the check. What we'd need to do really here is fix normal vs LTR to have distinct version update check files.

m-kuhn:
@nyalldawson is there another possibility besides spcifying `--noversioncheck`?

DelazJ:
Settings --> Options --> General? (I had forgotten that one)
> What we'd need to do really here is fix normal vs LTR to have distinct version update check files.

+1

nyalldawson:
@m-kuhn 

Yes, see https://github.com/qgis/QGIS/blob/master/resources/qgis_global_settings.ini#L3 and https://github.com/qgis/QGIS/blob/master/resources/qgis_global_settings.ini#L12

m-kuhn:
Brilliant

Gustry:
It has been explained in the documentation how to disable version check when it's deployed by a system administrator: https://docs.qgis.org/3.4/en/docs/user_manual/introduction/qgis_configuration.html#deploying-qgis-within-an-organization

m-kuhn:
So how about this alternative:

 - Add a new web service for LTR
 - Add a button directly on the version check to set `checkVersion=false` without going to the settings

nyalldawson:
Both sound good!

m-kuhn:
@jef-n is there information for LTR available like on https://version.qgis.org/version.txt ?

m-kuhn:
@jef-n do you have a hint about the server side of version.txt (for User-Agent checking)?

jef-n:
> @jef-n do you have a hint about the server side of version.txt (for User-Agent checking)?

Yes, qgis/QGIS-Website@e10533371f6f adds a version-ltr.txt and the apache configuration rewrites accessed to version.txt from the LTR to it based on it's user-agent string.

PR #31299, opened 2019-08-19
Server: catch Python exceptions from plugins

elpaso:
Make sure Python exceptions are converted into
QgsServerException and catched by the main
service executor.

### DIFF
--- a/src/server/qgsserverogcapihandler.h
+++ b/src/server/qgsserverogcapihandler.h
@@ -108,8 +108,10 @@ class SERVER_EXPORT QgsServerOgcApiHandler
     /**
      * Returns the default response content type in case the client did not specifically
      * ask for any particular content type.
+     * The default implementation returns the first content type returned by
+     * contentTypes() or JSON if that list is empty.
      */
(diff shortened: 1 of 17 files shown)


elpaso:
@rldhont @pblottiere @dmarteau I know we should make some housekeeping on the exception hierarchy/handling, but for now this seems a decent solution in order to catch Python errors coming from the plugin's code.

src/server/qgsserverogcapihandler.h
```diff
@@ -111,12 +111,20 @@ class SERVER_EXPORT QgsServerOgcApiHandler
      */
     virtual QgsServerOgcApi::ContentType defaultContentType() const  { return QgsServerOgcApi::ContentType::JSON; }
 
+    /**
+     * Returns the list of content types this handler can serve, default to JSON and HTML.
+     * In case a specialized type (such as GEOJSON) is supported,
+     * the generic type (such as JSON) should not be listed.
+     * \note not available in Python bindings
```
pblottiere: `\since`

pblottiere:
> but for now this seems a decent solution in order to catch Python errors coming from the plugin's code. 

Good move :+1:. Tanks @elpaso!

PR #31303, opened 2019-08-20
Export multiple themes via layout GeoPDF export

nyalldawson:
No UI yet, but this adds the API hooks necessary to allow multiple map themes to be exported when creating a GeoPDF file from a layout.

Wait, wat?

...this screencast demonstrates the result
Peek 2019-08-20 11-18

Using this api, it's possible to create a geopdf with a structure showing multiple map themes, allowing them to be toggled between. When the UI for this lands it will consist of a list of the current map themes in the project, with check boxes to select which themes to include in the geopdf output.

### DIFF
--- a/src/core/qgsabstractgeopdfexporter.cpp
+++ b/src/core/qgsabstractgeopdfexporter.cpp
@@ -83,15 +83,14 @@ QString QgsAbstractGeoPdfExporter::geoPDFAvailabilityExplanation()
 
 bool QgsAbstractGeoPdfExporter::finalize( const QList<ComponentLayerDetail> &components, const QString &destinationFile, const ExportDetails &details )
 {
-#if GDAL_VERSION_NUM < GDAL_COMPUTE_VERSION(3,0,0)
-  Q_UNUSED( components )
-  Q_UNUSED( destinationFile )
-  Q_UNUSED( details )
-  return false;
-#else
   if ( details.includeFeatures && !saveTemporaryLayers() )
     return false;
 
+#if GDAL_VERSION_NUM < GDAL_COMPUTE_VERSION(3,0,0)
+  Q_UNUSED( components )
+  Q_UNUSED( destinationFile )
+  return false;
+#else
   const QString composition = createCompositionXml( components, details );
   QgsDebugMsg( composition );
   if ( composition.isEmpty() )
@@ -137,7 +136,7 @@ QString QgsAbstractGeoPdfExporter::generateTemporaryFilepath( const QString &fil
 }
 
 
-void QgsAbstractGeoPdfExporter::pushRenderedFeature( const QString &layerId, const QgsAbstractGeoPdfExporter::RenderedFeature &feature )
+void QgsAbstractGeoPdfExporter::pushRenderedFeature( const QString &layerId, const QgsAbstractGeoPdfExporter::RenderedFeature &feature, const QString &group )
 {
   // because map layers may be rendered in parallel, we need a mutex here
   QMutexLocker locker( &mMutex );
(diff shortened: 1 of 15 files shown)


src/core/qgsabstractgeopdfexporter.cpp
```diff
@@ -253,31 +256,78 @@ QString QgsAbstractGeoPdfExporter::createCompositionXml( const QList<ComponentLa
...
+        {
+          groupLayerMap[ component.group ].appendChild( layer );
+        }
+        else
+        {
+          QDomElement group = doc.createElement( QStringLiteral( "Layer" ) );
+          group.setAttribute( QStringLiteral( "id" ), QStringLiteral( "group_%1" ).arg( component.group ) );
+          group.setAttribute( QStringLiteral( "name" ), component.group );
+          group.setAttribute( QStringLiteral( "initiallyVisible" ), groupLayerMap.empty() ? QStringLiteral( "true" ) : QStringLiteral( "false" ) );
+          group.setAttribute( QStringLiteral( "mutuallyExclusiveGroupId" ), QStringLiteral( "__mutually_exclusive_groups__" ) );
```
nirvn: It's nice PDF spec allows for mutually exclusive groups.

tests/src/core/testqgslayoutgeopdfexport.cpp
```diff
@@ -191,27 +205,151 @@ void TestQgsLayoutGeoPdfExport::testCollectingFeatures()
...
+
+  QgsLayoutGeoPdfExporter geoPdfExporter2( &l );
+  const QString outputFile2 = geoPdfExporter2.generateTemporaryFilepath( QStringLiteral( "test_src2.pdf" ) );
+  settings = QgsLayoutExporter::PdfExportSettings();
+  settings.writeGeoPdf = true;
+  settings.exportMetadata = false;
+  exporter.exportToPdf( outputFile, settings );
+
+  // check that features were collected
+  lineFeatures = geoPdfExporter2.mCollatedFeatures.value( QString() ).value( linesLayer->id() );
```
nirvn: When exporting multiple themes, an empty group string value represents what?

nirvn (commented):
Looking good.

tests/src/core/testqgslayoutgeopdfexport.cpp
nyalldawson: No group is associated/needed (e.g. not exporting themes)

src/core/qgsabstractgeopdfexporter.cpp
nyalldawson: Yep :) There's actually a lot of other nice stuff possible here (but which can wait for future releases)

nirvn (approved):

roya0045:
@nyalldawson I was wondering if this might have the same issues as the map legend has when it comes to themes. With a layer with a classified or rule-based symbology, when there is a theme with only one class selected, the whole layer/all the classes will be displayed in the legend but not the map.

 I was curious if in such cases would the whole layer be added or only the desired features that are selected in the theme?

nyalldawson:
@roya0045 yes, it will suffer those same issues as legends and themes unfortunately. I guess it's less impactful in this particular situation because for a multi-theme export a user would want to design a legend that makes sense for all exported themes.

That makes me think though of a nice follow up we could do here, where we could allow particular layout items to be assigned to a particular logical layer in the exported geopdf too. Then you'd be able to have per-theme legends which toggle along with the map content. (But that's dependant on a fix for that underlying legend issue in the first place!) 

I might see if I can do a focused run on layout specific issues during the upcoming freeze period.

roya0045:
@nyalldawson Good to hear, this is a minor convenience thing that should only affect an handful of users but I have made a feature request or a QEP some time ago about it, just in case.

PR #31305, opened 2019-08-20
GeoPDF layout UI

nyalldawson:
Exposes the GeoPDF export options and settings during layout PDF exports

### DIFF
--- a/src/app/layout/qgslayoutpdfexportoptionsdialog.h
+++ b/src/app/layout/qgslayoutpdfexportoptionsdialog.h
@@ -24,7 +24,7 @@
 #include "qgsrendercontext.h"
 
 /**
- * A dialog for customizing the properties of an exported PDF file.
+ * A dialog for customizing the properties of an exported PDF file from a layout.
 */
 class QgsLayoutPdfExportOptionsDialog: public QDialog, private Ui::QgsPdfExportOptionsDialog
 {
--- a/src/app/layout/qgslayoutpdfexportoptionsdialog.h
+++ b/src/app/layout/qgslayoutpdfexportoptionsdialog.h
@@ -68,6 +68,10 @@ class QgsLayoutPdfExportOptionsDialog: public QDialog, private Ui::QgsPdfExportO
 
     void showHelp();
 
+  private:
+
+    bool mGeopdfAvailable = true;
+
 };
 
 #endif // QGSLAYOUTPDFEXPORTOPTIONSDIALOG_H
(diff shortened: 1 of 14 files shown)


nirvn:
@nyalldawson , you're missing a vertical spacer in the PDF export dialog:
Screenshot from 2019-08-20 15-11-01

src/app/layout/qgslayoutpdfexportoptionsdialog.h
```diff
@@ -0,0 +1,77 @@
...
+#ifndef QGSLAYOUTPDFEXPORTOPTIONSDIALOG_H
+#define QGSLAYOUTPDFEXPORTOPTIONSDIALOG_H
+
+#include <QDialog>
+#include "ui_qgspdfexportoptions.h"
+
+#include "qgsrendercontext.h"
+
+/**
+ * A dialog for customizing the properties of an exported PDF file.
```
nirvn: If one wants to be super picky, one would ask for a reference to layout export in this class description. I.e., '... properties of an exported layout PDF file' to give a bit more context to the reader (who would skip actually reading the QgsClassName :) )

nirvn (approved):
Nice to see you crossing the finish line. This looks good to merge, with or without addressing the documentation comment I dropped here :) 

Oh, but let's not forget the vertical spacer missing in the .UI file.

nyalldawson:
> you're missing a vertical spacer in the PDF export dialog:

Hm- can't see immediately what's wrong there. I'll fix in the final PR.

src/core/qgsabstractgeopdfexporter.cpp
```diff
@@ -338,18 +338,21 @@ QString QgsAbstractGeoPdfExporter::createCompositionXml( const QList<ComponentLa
   page.appendChild( dpi );
   // assumes DPI of 72, which is an assumption on GDALs/PDF side. It's only related to the PDF coordinate space and doesn't affect the actual output DPI!
   QDomElement width = doc.createElement( QStringLiteral( "Width" ) );
-  width.appendChild( doc.createTextNode( QString::number( std::ceil( details.pageSizeMm.width() / 25.4 * 72 ) ) ) );
+  const double pageWidthPdfUnits = std::ceil( details.pageSizeMm.width() / 25.4 * 72 );
```
rouault: shouldn't 72 be replaced by the above DPI ? I'm a bit confused by the above comment. The DPI setting in the XML should affect the DPI of the output PDF

nyalldawson: Hardcoding 72 was the only way I could get things to line up - it seems the coordinate space isn't affected by the actual dpi setting itself.

nyalldawson: That said -- it was early in this development that I reached that conclusion, and it's possible something else was messing with things. I'll re-test.

PR #31312, opened 2019-08-20
GeoPDF export (last part!)

nyalldawson:
Last part of the GeoPDF package of works! This one adds a UI for configuring layer-by-layer export attributes and layer groups to GeoPDF export.

It allows users to set on a layer-by-layer basis whether layers should be identifiable and include vector features in a GeoPDF export, and optionally whether to group multiple layers together into a single logical, switchable entry in the GeoPDF layer tree.

(After this one, there's just UI cleanup work remaining....)

### DIFF
--- a/src/app/layout/qgsgeopdflayertreemodel.cpp
+++ b/src/app/layout/qgsgeopdflayertreemodel.cpp
@@ -1,0 +1,170 @@
+/***************************************************************************
+  qgsgeopdflayertreemodel.cpp
+ ---------------------
+ begin                : August 2019
+ copyright            : (C) 2019 by Nyall Dawson
+ email                : nyall dot dawson at gmail dot com
+ ***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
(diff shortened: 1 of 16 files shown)


nirvn:
Screenshot?

src/app/layout/qgsgeopdflayertreemodel.cpp
```diff
@@ -0,0 +1,170 @@
...
+{
+  switch ( index.column() )
+  {
+    case LayerColumn:
+    {
+      if ( role == Qt::CheckStateRole )
+      {
+        if ( QgsVectorLayer *vl = vectorLayer( index ) )
+        {
+          vl->setCustomProperty( QStringLiteral( "geopdf/includeFeatures" ), value.toInt() == Qt::Checked );
```
nirvn: You've decided to attach those flags to the vector layer, which means it'll be a setting that will be carried across layouts. Is there a reason to think this would ever become problematic / undesirable. I.e., would those settings better sit as custom properties of a layout?

nyalldawson: I debated this too. Ultimately whichever way we go we'll frustrate one set of users! I think it's more likely that settings will apply across a whole project though, which is why I've gone with this approach.

nirvn: @nyalldawson , I can see the dilemma. I trust your gut then.

nirvn (approved):
Looks good.

PR #31313, opened 2019-08-21
[ui] Improve height of save map canvas as PDF dialog

nirvn:
## Description

The save map [canvas] as PDF dialog's height is getting out of control:
Screenshot from 2019-08-21 11-10-20

This PR implements an alternative way to inform users of layers using advanced effects, which avoids a long vertical list in the dialog itself:
peekKY6M6Z

The bonus of this alternative approach is that we can list _all_ layers using advanced effects instead of clipping the list to ten.

@nyalldawson , thoughts?

## Checklist


nirvn:
The need for this change is even more obvious when GeoPDF support is present:
image

nyalldawson (approved):

nyalldawson:
Good move!

PR #31315, opened 2019-08-21
"Center of segment" placement mode for marker and hash line  symbol layers

nyalldawson:
Adds a new placement option to place markers/hash lines at the center point of individual line segments

Refs #29785

Sponsored by SLYR


nirvn:
Looking good, nice addition.

PR #31319, opened 2019-08-21
New ctors for connections interface

elpaso:
Tries to make it clearer what are the use cases for the two
constructors (name and uri).

The new API:

```python
# loads an existing connection from the settings
md = QgsProviderRegistry.instance().providerMetadata('postgres')
conn = md.createConnection('qgis_test1')
# Save it back
md.saveConnection(conn, 'qgis_test1')
# Save with another name
md.saveConnection(conn, 'qgis_test2')
md.deleteConnection('qgis_test1')

# creates a "floating" connection
conn = md.createConnection(uri, { "dontResolveType": True})
# Save it in the settings
md.saveConnection(conn, 'qgis_test3')
```

```


nyalldawson (approved):
Looks good!

PR #31320, opened 2019-08-21
Update warp.py

### LINKED ISSUE #31276: Problems with Warp (Reproject) and user-defined output extent
**Description**

A problem occurs when using Warp (Reproject) from the Processing Toolbox or  gdal:warpreproject in Python when the CRS of a user-defined output extent is implicit or is specified in a user-defined coordinate system. In the former case the GDAL/OGR console call contains the ‘-te_srs’ string but no CRS definition.  In the latter case the -te_srs parameter is specified as an authid rather than being translated into a Proj4 string and this presumably causes the GDAL routine to crash. The error message in both cases is 'Translating source or target SRS failed'. If the Source CRS and Target CRS are explicitly specified as user-defined, these are translated to Proj4 in the console call.

**How to reproduce (implicit projection)**

Run Warp (Reproject) from the Processing Toolbox on any suitable raster and a define a georeferenced extent for the output in Advanced Parameters (e.g. the canvas extent) while leaving the CRS of the target raster extent blank.

**How to reproduce (user-defined projection)**

1.  Load a raster with a user-defined CRS into QGIS (example provided).
2.  Run Warp (Reproject) and select the user-defined CRS in the Source, Target and Extent drop-down lists. Define an extent for the output.
3.  Note the difference in the way the GDAL/OGR console call defines the -s_srs, -t_srs and -te_srs parameters.
4.  The routine crashes if -te_srs is a USER: rather than EPSG: system. 
[test_uklidar.zip](https://github.com/qgis/QGIS/files/3509481/test_uklidar.zip)

**QGIS and OS versions**

QGIS version
3.8.1-Zanzibar
QGIS code revision
dcd95cc648
Compiled against Qt
5.11.2
Running against Qt
5.11.2
Compiled against GDAL/OGR
2.4.1
Running against GDAL/OGR
2.4.1
Compiled against GEOS
3.7.2-CAPI-1.11.0
Running against GEOS
3.7.2-CAPI-1.11.0 b55d2125
PostgreSQL Client Version
9.2.4
SpatiaLite Version
4.3.0
QWT Version
6.1.3
QScintilla2 Version
2.10.8
Compiled against PROJ
5.2.0
Running against PROJ
Rel. 5.2.0, September 15th, 2018
OS Version
Windows 10 (10.0)

**Additional context**

Problem also occurs with the 'processing.run('gdal:warpreproject', parameters)' Python command  (TARGET_EXTENT_CRS parameter)

gkimbell:
## Description
Changes proposed to address bug reported in https://github.com/qgis/QGIS/issues/31276. The change to line 218 addresses the case where an output extent is defined but its CRS is left blank (implicit). The change to line 220 ensures that a USER: coordinate system is translated into a Proj4 string that is recognised by GDAL.

Apologies if this pull request doesn't comply with some items on the checklist (I'm unfamiliar with this process)

Fixes #31276

## Checklist

### DIFF
--- a/python/plugins/processing/tests/GdalAlgorithmsRasterTest.py
+++ b/python/plugins/processing/tests/GdalAlgorithmsRasterTest.py
@@ -1660,13 +1660,13 @@ class TestGdalRasterAlgorithms(unittest.TestCase, AlgorithmsTestBase.AlgorithmsT
                  outdir + '/check.jpg'])
 
             # with target using custom projection and user-defined extent
-            custom_crs = 'proj4: +proj=longlat +a=6378388 +b=6356912 +no_defs'
+            custom_crs2 = 'proj4: +proj=longlat +a=6378388 +b=6356912 +no_defs'
             self.assertEqual(
                 alg.getConsoleCommands({'INPUT': source,
-                                        'SOURCE_CRS': custom_crs,
-                                        'TARGET_CRS': custom_crs,
+                                        'SOURCE_CRS': custom_crs2,
+                                        'TARGET_CRS': custom_crs2,
                                         'TARGET_EXTENT': '18.67,18.70,45.78,45.81',
-                                        'TARGET_EXTENT_CRS': custom_crs,
+                                        'TARGET_EXTENT_CRS': custom_crs2,
                                         'OUTPUT': outdir + '/check.jpg'}, context, feedback),
                 ['gdalwarp',
                  '-s_srs "+proj=longlat +a=6378388 +b=6356912 +no_defs" -t_srs "+proj=longlat +a=6378388 +b=6356912 +no_defs" -r near -te 18.67 45.78 18.7 45.81 -te_srs "+proj=longlat +a=6378388 +b=6356912 +no_defs" -of JPEG ' +
--- a/python/plugins/processing/tests/GdalAlgorithmsRasterTest.py
+++ b/python/plugins/processing/tests/GdalAlgorithmsRasterTest.py
@@ -1669,9 +1669,9 @@ class TestGdalRasterAlgorithms(unittest.TestCase, AlgorithmsTestBase.AlgorithmsT
                                         'TARGET_EXTENT_CRS': custom_crs2,
                                         'OUTPUT': outdir + '/check.jpg'}, context, feedback),
                 ['gdalwarp',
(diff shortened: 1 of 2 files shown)


roya0045:
Have you tested this?

I assume that the best way to ensure that this work would be to make a test in there https://github.com/qgis/QGIS/blob/master/python/plugins/processing/tests/GdalAlgorithmsRasterTest.py

nyalldawson:
Fix looks good!

As @roya0045 has pointed out, we also need a unit test covering this. It would slot in right near these similar tests: https://github.com/qgis/QGIS/blob/master/python/plugins/processing/tests/GdalAlgorithmsRasterTest.py#L1638

python/plugins/processing/tests/GdalAlgorithmsRasterTest.py
```diff
@@ -1659,6 +1659,20 @@ def testWarp(self):
...
+                                        'SOURCE_CRS': custom_crs,
+                                        'TARGET_CRS': custom_crs,
+                                        'TARGET_EXTENT': '18.67,18.70,45.78,45.81',
+                                        'TARGET_EXTENT_CRS': custom_crs,
+                                        'OUTPUT': outdir + '/check.jpg'}, context, feedback),
+                ['gdalwarp',
+                 '-s_srs "+proj=utm +zone=36 +south +a=63785 +b=6357 +towgs84=-143,-90,-294,0,0,0,0 +units=m +no_defs" -t_srs "+proj=utm +zone=36 +south +a=63785 +b=6357 +towgs84=-143,-90,-294,0,0,0,0 +units=m +no_defs" -r near -te 18.67 45.78 18.7 45.81 -te_srs "+proj=utm +zone=36 +south +a=63785 +b=6357 +towgs84=-143,-90,-294,0,0,0,0 +units=m +no_defs" -of JPEG ' +
+                 source + ' ' +
+                 outdir + '/check.jpg'])
+
```
roya0045: I would suggest you to create a distinct test. Basically copy this function, and remove elements that belong to the other test and validate elements related to your changes. This way we can more easily monitor when something breaks this functionality.

gkimbell:
I've added a tentative unit test at lines 1662-1674 of https://github.com/gkimbell/QGIS/blob/master/python/plugins/processing/tests/GdalAlgorithmsRasterTest.py#L1662 

Not too confident about this as I have been unable to run the unit test properly on my PC (combination of trying to do things in Windows and extreme incompetence). Running an equivalent test manually in QGIS Desktop works OK after the edits to warp.py.

roya0045:
Since there are so many tests in this classes it should be fine, I didn't think that all the warp parameters were tested in the same function. This should be fine.

roya0045:
@gkimbell could you do a minimal commit to trigger to force travis to build and test again. I don't think your changes are what is causing the test to fail. If you force travis to rebuild well be able to know if everything is ok and the devs will be able to merge it once they get the green check mark.

nyalldawson (approved):

PR #31330, opened 2019-08-21
More explicit documentation of geometry classes, with warnings

### LINKED ISSUE #31275: Add warnings for "stupid", non-spatial GEOS-based functions like length and area
**Feature description.**
Please add notes or warnings to the docs of all the not-spatially-aware GEOS-based functions used for calculating things like lengths and areas.

For example https://qgis.org/pyqgis/master/core/QgsGeometry.html#qgis.core.QgsGeometry.area and https://qgis.org/pyqgis/master/core/QgsGeometry.html#qgis.core.QgsGeometry.length will treat all coordinates as if they were on a flat cartesian plane. This is confusing to users who assume their GIS abstracts these issues for them.

This extends to expression functions like `area()` and `length()`, though these two already have appropriate warnings in their help texts:

> Calculations are always planimetric in the Spatial Reference System (SRS) of this geometry, and the units of the returned area will match the units for the SRS. This differs from the calculations performed by the $area function, which will perform ellipsoidal calculations based on the project's ellipsoid and area unit settings.

I don't know which functions and methods all might be affected.

**Additional context**
Motivated by https://gis.stackexchange.com/questions/332179/qgsgeometry-length-gives-wrong-result/332180 . It is a very common problem.

nyalldawson:
that calculations are Cartesian only and point users to QgsDistanceArea
for ellipsoidal calculations

Fixes #31275

### DIFF
--- a/python/core/auto_generated/geometry/qgsabstractgeometry.sip.in
+++ b/python/core/auto_generated/geometry/qgsabstractgeometry.sip.in
@@ -25,7 +25,7 @@ Abstract base class for all geometries
 .. note::
 
    QgsAbstractGeometry objects are inherently Cartesian/planar geometries. They have no concept of geodesy, and none
-   of the methods or properties exposed from the QgsAbstractGeometry API (or QgsGeometry API) utilise
+   of the methods or properties exposed from the QgsAbstractGeometry API (or QgsGeometry API) utilize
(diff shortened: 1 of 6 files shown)


m-kuhn (approved):
Very nice addition!

PR #31332, opened 2019-08-22
Auto fix other matching broken paths when a layer's path is fixed

nyalldawson:
When a layer path is fixed in a project, auto scan through all other broken paths and try to auto-fix any others which were also pointing to the same original broken file path

AKA the ArcMap behavior

Peek 2019-08-22 10-28


nirvn (approved):
Nice.

nyalldawson:
@elpaso can you look this one over too?

elpaso (approved):
Nice!

Issue #31253, opened 2019-08-14
Loading too many features in attribute table


plutoscarab:
## User Feedback

Crashed while I was loading a ridiculous amount of rows in Attribute Table. It crapped out at around 900,000 features from the Geonames gazetteer dataset (http://download.geonames.org/export/dump/allCountries.zip).

## Report Details

*Crash ID*: ffc485992db59bda633a330389c3d8b52f05270c

*Stack Trace*
<pre>
VSICreateCachedFile :
</pre>

*QGIS Info*
QGIS Version: 3.8.0-Zanzibar
QGIS code revision: 11aff65f10
Compiled against Qt: 5.11.2
Running against Qt: 5.11.2
Compiled against GDAL: 2.4.1
Running against GDAL: 2.4.1

*System Info*
CPU Type: x86_64
Kernel Type: winnt
Kernel Version: 10.0.18362

gioman:
> Crashed while I was loading a ridiculous amount of rows in Attribute Table. It crapped out at around 900,000 features from the Geonames gazetteer dataset (http://download.geonames.org/export/dump/allCountries.zip).

It also brings a high end workstation to a complete freeze when opening the table after having imported the dataset into PostGIS.

900k is not a "ridiculous amount" in my opinion, nowadays is quite common to need to work with very large datasets and I think I will not offend anyone when I say that working with such amount of data in the QGIS table of attributes in QGIS has been historically very bad. Is also known that the developers say that the underlying code is not quite good. At some point I think that as project we should face this issue and define what are the necessary steps (and resources) needed to improve the situation.

stelf:
QGIS is for some while in use at Sofia/Bulgaria/ Municipality's organization for planning and under my advice they are moving hundreds of layers in QGIS, some featuring 200-300-400k elements. While most smaller sets (less than 100k) are stale bigger ones tend to bring the system to a halt, or just crash/close QGIS. And yes, in 2019 it's not big data to have some million rows in a table. It's 10TB what people consider big data.

IMHO A much wider discussion is needed reg. how to offload as-many-as-possible operations to the DB when ST-enabled DB is the source for the layer, as this will greatly reduce CPU operations and backends are really optimized for large data-sets.

elpaso:
The attribute table issue with big tables is a well know longstanding issue, a deep refactoring of the code is required.

roya0045:
[...]

PR #31352, opened 2019-08-22
[Feature]Partially load the attribute table instead of loading it all.

roya0045:
## Description
Disclaimer: I'm not too sure what I'm doing so feedback/guidance would be helpful.

This is a PR related to #31253 and the proposal that I made within it to @elpaso .

- [x] signals/proper data handling

- [x] UI for a button to load the full table

- [x] possibly a variable to set the maximum number of features to load per batch?

Here is a demo of the feature in action:
partial_loader

## Checklist


Saijin-Naib:
Question: Do you envision the ability to batch load via a UI button, or perhaps soft-stop/scroll at the boundary, sort of like how some continuous content websites will force a bounce-back then actually load more when the user continues trying to scroll?

roya0045:
> Question: Do you envision the ability to batch load via a UI button, or perhaps soft-stop/scroll at the boundary, sort of like how some continuous content websites will force a bounce-back then actually load more when the user continues trying to scroll?

The behaviour that I'm going for is the following: load the first x layers (in this case 1000), when the users scrolls down to 90% load the next 1000. The trigger range can be absolute or relative, in this case it is relative.

I am planning to add a button to load the full table (which would be the current method employed).

Suggestions are welcome.

stelf:
@roya0045 , as a side note, it also makes sense to have an option to actually free the memory used for records not currently shown. 

basically in most cases it is when someone wants to export or do anything else involving all records that these need be loaded. also valid for filtered views. so that large data-sets are never completely fed into memory, because basically most use cases users don't really need to keep them there.

also - in case data shown in the the table results in selection of features, then only the geometry + pk is needed.

this way arbitrary sized attribute table may be subject to navigation, and in case a proper geospatial DB is the source for the layer, the paging may be handled effectively by the back-end. I've seen postgre pump data into QGIS at 2mbits rate without ever increasing the CPU on the server side.

roya0045:
> @roya0045 , as a side note, it also makes sense to have an option to actually free the memory used for records not currently shown.
> 
> basically in most cases it is when someone wants to export or do anything else involving all records that these need be loaded. also valid for filtered views. so that large data-sets are never completely fed into memory, because basically most use cases users don't really need to keep them there.
> 
> also - in case data shown in the the table results in selection of features, then only the geometry + pk is needed.
> 
> this way arbitrary sized attribute table may be subject to navigation, and in case a proper geospatial DB is the source for the layer, the paging may be handled effectively by the back-end. I've seen postgre pump data into QGIS at 2mbits rate without ever increasing the CPU on the server side.

I agree but the bottleneck is the tableview, this is the part that you must keep in mind. There is no easy way to 'dump' data in the table view, otherwise it would already be implemented or I would have implemented it. But data cells must be implemented one by one and this is why its so slow.

nyalldawson:
@roya0045 - my honest advice is not to go down this track. The attribute table needs to be rebuilt from the ground up, and the work involved is MAMMOTH and requires in depth understanding of the complex interplay between the attribute form, editor widgets, the vector layer cache mechanism, and the selection models. For reference, @elpaso and I looked into this and estimated it's at least one month of full time work to do (likely more). I fear that going down this path is a certain track to madness! 😱

roya0045:
@nyalldawson I beleive you, thing seems to be set up in quite a weird way but I want to see if I can set up something to ease the pain in the meantime. I'm fine if this is not merged but I'll eventually try to test it locally and see if I can patch things somewhat.

nyalldawson:
@roya0045 yes, it's a royal spaghetti tangle stuck together with a bonfire of qt signals!

roya0045:
Strangely this approach seems to increase the number of row by one, which seems strange to me. And hence cause most of the test failures. But at least the table was loaded.

gioman:
> But at least the table was loaded.

I'm curious, so you were able to have the table load in chunks while scrolling down? if yes have you tried with a very large table?

roya0045:
> > But at least the table was loaded.
> 
> I'm curious, so you were able to have the table load in chunks while scrolling down? if yes have you tried with a very large table?

I'm saying this judging from the tests results, I cannot compile things on this system, hence why I do things blind and patch according to the travis results.

roya0045:
Not stale.

roya0045:
Quick update, got some time to build and test, I found the cause of the extra row, should be solved now. Though the trigger to load more rows does not seem to work properly, need to fix this.

roya0045:
@nyalldawson Welp, it works fine, the only issue at the moment is to trigger the loading, you must use the mouse on the slider otherwise I think that using the mouse wheel doesn't send the signal as stated in the qt doc. If you have the time to try it out and decide if its a good thing or not, there is no hurry for this PR.

@gioman @Saijin-Naib PR seems operational if you want to try it out. I did minimal tests but everything seems ok to me.

Oh and when filtering, it will filter out features that were loaded, and not all features. It keeps track of things, which might be good or bad.

roya0045:
@elpaso since you have talked about this with Nyall this is a simple ping, in case you might be interested to check the code or try it out. Or if you have suggestions.

roya0045:
Mouse wheeling is not supported.

Though I get random errors with the filter svg and I got an error where it said "Qt Error: showing notification".

I'm not sure what might be causing those or if they are a local thing or not, hence why some external testing would be appreciated.

roya0045:
Added a param in the settings/options. Works fine!

nirvn:
@roya0045 , have you tried scrolling through the dual view feature list with your commits applied? Is incremental working there too?

I think this PR's bottleneck won't be UI/UX-related, but the underlying architecture. Nyall already hinted at the poor state of affairs when it comes to attribute table, and I'm not intimidate enough with that part of the code to review.

Looking at your SVG, the icon itself is not super clear as to what it does (i.e, it looks like a button you'd hit if you want to go at the very bottom of the list). It's not an easy one to come up with. Also, don't forget to compress your SVGs (there are a few online compressors you can use).

roya0045:
@nirvn Thanks the review! As for the dual view suggestion, I forgot to rig it and I forgot about the SVG compression thing.

I did some more tests and only two things generate strange results.
1) When undoing some deletion, all layers that were removed are added to the active table.
2) Sometimes somehow when doing a mass delete, rows will remain with an error. This error seems unstable and occur randomly, mostly when deleting a lot of rows instead of deleting the features in the map.

The dualview has been patched.

I am surprised by how easy this modification is. Despite the fact that Nyall said that this was a clusterfak I think that I did the changes in the right place to cause the desired behavior without any changes to highly rigged components. By having this PR is an optimal state this lets core devs review it better and estimate if this implementation can lead to a good gain without breaking fragile parts.

To be honest I think this approach doesn't break anything and doesn't seem to depend and interact with the parts that Nyall mentioned (at least I didn't see anything horrible ).

roya0045:
Ok I found the source of the error.
When features are deleted, somehow the resetModel gets called.

But its intermittent. :/

roya0045:
@elpaso Could you change the prototype tag to a feature tag. It works fine other than the intermittent weird refresh on mass deletion.

src/gui/attributetable/qgsattributetablemodel.cpp
```diff
@@ -967,3 +1007,13 @@ const QgsFeatureRequest &QgsAttributeTableModel::request() const
 {
   return mFeatureRequest;
 }
+
+
+void QgsAttributeTableModel::sliderCheck( int position )
```
nyalldawson: Instead of depending on the movement of the slider, what we should instead be doing is using calls to QgsAttributeTableModel::data to detect which features are being requested by the view, and fetching the batches when a previously-unfetched feature is required. This removes the fragile approach used here which is mixing gui code with model code and which is the source of the issues you've identified.

I actually believe all this code is already in place, and it's how the attribute table used to work in very old QGIS versions. Features where fetched from the provider only when the model required them. I believe this was changed when sorting support was added to the table, as that required a complete load of the table in advance. I would suggest investigating whether disabling the initial full table load still falls back to the load-features-on-demand approach, and whether all that code is still in place and working correctly. Then we can investigate possible approaches to avoid the initial bulk-load when sorting is required.

roya0045: I have commented out the calls to loadLayer and the only thins that results are headers. I don't think the former code exists.

As far as relying on ::data to detect needed features, I am not sure how it would be possible. The table viewer by default doesn't request specific features. As this tool is more used to simply explorer the content of a layer and modify that content, there is no requests that are made usually. The only approach where I can see this being of use if when a user selects a layer and if we want that selected layer to appear in the table.

I'm also not sure if calls to data are being made, possibly these are only for display/rendering?

roya0045: @nyalldawson ok I did some test using QgsAttributeModel::data to see how the data is structured. Basically nothing is ordered sequentially. I have no clue why and this is especially the case with the dual view (I still need to patch it more).

Even in the base model view, number have a random number. Using my pointer and around row 20 to 31, The row value of the model index goes something like: 33,9,46,2,21,68..

Using this as a threshold would mean that the user can be randomly mousing over a cell and trigger the database to load another batch even if he is in the first 10 rows, or just go on a random hunt for the magic cells that will trigger another batch to load by mousing over everything. This is a far worse option than using the GUI.

Also I have checked the code from 2013 and the table that was loaded was based on a rectangle or selection but I couldn't identify the part that loaded the rest of the table when needed.

nyalldawson: @roya0045 you'll need to do some research in the qt model/view infrastructure. It's a mental twist at first, but a useful concept to understand fully. Unfortunately the qgis attribute table is NOT a good starting point for learning models... There's just too much happening here! But, short answer is, we **cannot** use the current approach of mixing the gui portions (the view) with the underlying model data. You've already bumped into some limitations of this approach, and we'll hit many more (probably unsolvable) ones in future if we go down this track (and the last thing we want is to make these classes more complex and confusing! 😱😜). So we need to find an approach which relies solely on changes in the model and/or attribute table cache only...

roya0045: I'll see what I can do come up with. Thought I don't quite like how often the data call is actually used in qt models ( as stated in my symbology PR), this make things more noisy. I also feel like for such an approach using ui as a trigger since it doesn't care about the content but only the position. Though I think that the issue I encountered are more related to the model. The error values are only element that are not longer present. The only wrong behaviour that I saw was due to the fact that the wrong function was called when too many features/row were deleted and that function didn't refresh the content of the table, and that call was inconsistent. I would not link this to interweaving the UI and model and the UI check seem to be resistant to the tests that I carried out.  Closing and reopening the table would close the only occasional issue I encountered. Further testing and on a bigger scale might help to show if this approach is truly bad.

As it is mostly aimed at interface element, I'm confused as to why the solution must not imply ui given that this part of the code does not appear to be problematic.

We can go over this later, I would appreciate if my more mature PR could be discussed beforehand. In the meantime I will see what I can do.

nyalldawson: Thanks @roya0045! I'd really encourage you to dig into how models are used elsewhere in qgis and use this to get a good grounding in the concept. They took me quite some time to get my head around at first! Good ones to study as examples would be QgsMapLayerModel and QgsFieldModel

roya0045:
@nyalldawson 
Here is a sample of what I get when loading the first 100 features:
image
As you can see the order is random, with lots of duplicates, I am aware that the qmodel are special thing but this does not make sense and cannot be used to control any sort of behavior in itself.

roya0045:
Not 'stale'.

jratike80:
Strong +1 for having some way for loading just part of the attribute table. I have recently met two use cases when loading the whole attribute table just does not work. The first one is OGF API Features service with 4 million address points and another is Geopackage db with 400 million measurement points. Both data sources work fine on the map view if scale is kept big enough.

I would be happy if there was an alternative to query the attributes of just those features which are visible on the screen.

stelf:
This feature is so important IMHO as optimization that eventually benefits everyone, that should be considered candidate for some grant.

roya0045:
> This feature is so important IMHO as optimization that eventually benefits everyone, that should be considered candidate for some grant.

It works pretty well for me at the moment. But I fear that to implement the changes suggested by Nyall would be quite complicated as the data model seem to store stuff quite randomly.

If you wish I would suggest you to grab master, merge this in and compile it and report anything in this thread to see if the proposed solution behaves correctly for your.

(I would rather have my old PR get merged before this since its been many months without any review on them...)

gioman:
> It works pretty well for me at the moment

so can I test your branch? I just need to compile it?

roya0045:
> > It works pretty well for me at the moment
> 
> so can I test your branch? I just need to compile it?

Absolutely, anyone can use my PRs as they see fit. I just rebased it, it should be better and still be viable for you to investigate bugs.

gioman:
> Absolutely, anyone can use my PRs as they see fit. I just rebased it, it should be better and still be viable for you to investigate bugs.

compiling fails here:

```
[ 24%] Building CXX object src/gui/CMakeFiles/qgis_gui.dir/attributetable/qgsattributetablemodel.cpp.o
/home/giovanni/Downloads/QGIS_1-gradual_table_view/src/gui/attributetable/qgsattributetablemodel.cpp: In member function ‘virtual void QgsAttributeTableModel::loadLayerPart()’:
/home/giovanni/Downloads/QGIS_1-gradual_table_view/src/gui/attributetable/qgsattributetablemodel.cpp:516:40: error: no matching function for call to ‘QgsAttributeTableModel::featureAdded(QgsFeatureId, bool)’
         featureAdded( mFeat.id(), true );
                                        ^
/home/giovanni/Downloads/QGIS_1-gradual_table_view/src/gui/attributetable/qgsattributetablemodel.cpp:218:6: note: candidate: virtual void QgsAttributeTableModel::featureAdded(QgsFeatureId)
 void QgsAttributeTableModel::featureAdded( QgsFeatureId fid )
      ^~~~~~~~~~~~~~~~~~~~~~
/home/giovanni/Downloads/QGIS_1-gradual_table_view/src/gui/attributetable/qgsattributetablemodel.cpp:218:6: note:   candidate expects 1 argument, 2 provided
At global scope:
cc1plus: warning: unrecognized command line option ‘-Wno-redundant-move’
src/gui/CMakeFiles/qgis_gui.dir/build.make:1358: recipe for target 'src/gui/CMakeFiles/qgis_gui.dir/attributetable/qgsattributetablemodel.cpp.o' failed
make[2]: *** [src/gui/CMakeFiles/qgis_gui.dir/attributetable/qgsattributetablemodel.cpp.o] Error 1
CMakeFiles/Makefile2:1555: recipe for target 'src/gui/CMakeFiles/qgis_gui.dir/all' failed
make[1]: *** [src/gui/CMakeFiles/qgis_gui.dir/all] Error 2
Makefile:162: recipe for target 'all' failed
make: *** [all] Error 2
```

roya0045:
> > Absolutely, anyone can use my PRs as they see fit. I just rebased it, it should be better and still be viable for you to investigate bugs.
> 
> compiling fails here:
> 
> ```
> [ 24%] Building CXX object src/gui/CMakeFiles/qgis_gui.dir/attributetable/qgsattributetablemodel.cpp.o
> /home/giovanni/Downloads/QGIS_1-gradual_table_view/src/gui/attributetable/qgsattributetablemodel.cpp: In member function ‘virtual void QgsAttributeTableModel::loadLayerPart()’:
> /home/giovanni/Downloads/QGIS_1-gradual_table_view/src/gui/attributetable/qgsattributetablemodel.cpp:516:40: error: no matching function for call to ‘QgsAttributeTableModel::featureAdded(QgsFeatureId, bool)’
>          featureAdded( mFeat.id(), true );
>                                         ^
> /home/giovanni/Downloads/QGIS_1-gradual_table_view/src/gui/attributetable/qgsattributetablemodel.cpp:218:6: note: candidate: virtual void QgsAttributeTableModel::featureAdded(QgsFeatureId)
>  void QgsAttributeTableModel::featureAdded( QgsFeatureId fid )
>       ^~~~~~~~~~~~~~~~~~~~~~
> /home/giovanni/Downloads/QGIS_1-gradual_table_view/src/gui/attributetable/qgsattributetablemodel.cpp:218:6: note:   candidate expects 1 argument, 2 provided
> At global scope:
> cc1plus: warning: unrecognized command line option ‘-Wno-redundant-move’
> src/gui/CMakeFiles/qgis_gui.dir/build.make:1358: recipe for target 'src/gui/CMakeFiles/qgis_gui.dir/attributetable/qgsattributetablemodel.cpp.o' failed
> make[2]: *** [src/gui/CMakeFiles/qgis_gui.dir/attributetable/qgsattributetablemodel.cpp.o] Error 1
> CMakeFiles/Makefile2:1555: recipe for target 'src/gui/CMakeFiles/qgis_gui.dir/all' failed
> make[1]: *** [src/gui/CMakeFiles/qgis_gui.dir/all] Error 2
> Makefile:162: recipe for target 'all' failed
> make: *** [all] Error 2
> ```

Thanks, should be fixed, I forgot that Nyall changed some things and forgot to adapt my code, the error is mine.

gioman:
> Thanks, should be fixed, I forgot that Nyall changed some things and forgot to adapt my code, the error is mine.

works now, thanks! but I can't figure how it works. Can you give me some tip?

roya0045:
> > Thanks, should be fixed, I forgot that Nyall changed some things and forgot to adapt my code, the error is mine.
> 
> works now, thanks! but I can't figure how it works. Can you give me some tip?

Basically in the option you should have a new setting to set up the size of rows to load by batch.

When you open the attribute table in normal or dual view and you scroll down at the bottom of the first x features, another batch of x features should load. x being the size that you indicated in the attribute table options. In the attribute table itself you should also have a button at the top to load the full table.

gioman:
> Basically in the option you should have a new setting to set up the size of rows to load by batch.
> 
> When you open the attribute table in normal or dual view and you scroll down at the bottom of the first x features, another batch of x features should load. x being the size that you indicated in the attribute table options. In the attribute table itself you should also have a button at the top to load the full table.
> 
> If I recall the default is 'all' (i e 0 ) in order to not confuse people and make them think that something is broken.

oh I see.

I tested a bit with a 300k points postgis remote layer.

I tried with 100 in "rows to load initially" and it is nice to see more rows to be loaded while scrolling down! Anyway using a bigger value and the lag while scrolling become very noticeable.

If I search ("field filter") for a feature that isn't loaded in the table then it takes a long (like it is searching in the whole table) to... show nothing.

I was hoping (probably isn't covered by this patch) in a faster time to open the table like when we use the the general option "show features visible on map", but it seems that the initial load time is similar to the one when all the records are loaded.

roya0045:
@gioman yeah something might be slowing things down at load time, I'm not sure what though.

As for field filtering I'd have to look at the code. I'm not sure if I want to adapt it to this case. You could try to apply the filter and then load the full table to see how it behaves ( I don't recall a test with a missing variable in the filter).

I wonder what might be the bottleneck since loading a part of the table and the full table seem to take similar time this tells me that the bottleneck might occur when fetching the table itself before recreating it in QGIS. Otherwise with partial loading there should be far less iterations to populate the table.

When you talk about the 'lag' yes its normal as it must fetch more features and those can only be fetched sequentially at the moment.

Hope this clear things up and thanks for testing!

gioman:
> You could try to apply the filter and then load the full table to see how it behaves ( I don't recall a test with a missing variable in the filter).

in this case it works as expected.

>  wonder what might be the bottleneck since loading a part of the table and the full table seem to take similar time this tells me that the bottleneck might occur when fetching the table itself before recreating it in QGIS. Otherwise with partial loading there should be far less iterations to populate the table.

I think this is the major issue: one of the most annoying things in QGIS is the load time of large tables. One can use the "show features visible on map" to workaround it, as load times become much faster (if the number of visible features is small enough). My impression is that fsater load time are also expected when using the "rows to load initially" parameter... otherwise is not worth load only a bunch of features and wait as long as when all features are loaded.

Anyway seems to me that this patch is heading in the right direction!

roya0045:
> I think this is the major issue: one of the most annoying things in QGIS is the load time of large tables. One can use the "show features visible on map" to workaround it, as load times become much faster (if the number of visible features is small enough). My impression is that fsater load time are also expected when using the "rows to load initially" parameter... otherwise is not worth load only a bunch of features and wait as long as when all features are loaded.

I agree and it was the expected behavior. I can see a small improvement but I'm not sure yet what is the bottleneck. I would need to investigate things further.

roya0045:
I'll close this temporarily, until the more important stuff gets merged.

TurboGraphxBeige:
@roya0045 Hey I think this PR is a nice proposition for this crippling big attribute tables issue. Is it still alive?

roya0045:
> @roya0045 Hey I think this PR is a nice proposition for this crippling big attribute tables issue. Is it still alive?

Not really, as far as I could recall, this isn't feasible without major change to how layers are handled. The current process will always want to fully read the layer regardless, so you don't gain any efficiency in read time, just less data displayed. I also don't recall if a new read operation was needed to reload each 'chunk'.

This mostly mean that it would result in something more aesthetic rather than speeding up things.

weca-theo:
agreed with @gioman here, that a major annoyance is opening an attribute table of a large dataset can take forever, when you might only want to view 100 rows to quickly see what values exist in the table. I have tables in Azure hosted PostgreSQL which are usually <1 million rows... opening the attribute table can take **minutes** and also hangs QGIS whilst it loads ('not responding' text appears). When this happens, my users will often kill the qgis process and reopen their project file (a quicker process than waiting for the attribute table to open!).

roya0045:
@weca-theo having this be possible without the big loading time would require a major rework of how data is handled. Partly reading a file and resuming the read is not something trivial, and if you  keep everything in memory you'll have a bigger footprint and more issues. This would be easier to properly implement for web data sources but not local files.

PR #31377, opened 2019-08-23
Prevent installation of ZIP plugins if they don't have the root folder

### LINKED ISSUE #30063: Install from zip erases plugin directory
If to try to `install from zip` badly zipped plugin it can erase whole plugin directory

1. Go to Plugins - Manage and Install Plugins -> Install from ZIP
2. Choose plugin_killer.zip
This plugin is zipped without dir, just files

QGIS 3.6.3, Windows 10, Windows 7

[plugin_killer.zip](https://github.com/qgis/QGIS/files/3247376/plugin_killer.zip)

seralra96:
Fixes #30063

## Description

Prevent installation of ZIP plugins if they don't have the root folder. This also prevent to erase all folders inside the plugin folder in this case.

## Checklist

### DIFF
--- a/python/pyplugin_installer/installer.py
+++ b/python/pyplugin_installer/installer.py
@@ -27,7 +27,7 @@ import os
 import json
 import zipfile
 
-from qgis.PyQt.QtCore import Qt, QObject, QDir, QUrl, QFileInfo, QFile
+from qgis.PyQt.QtCore import Qt, QObject, QDir, QUrl, QFileInfo, QFile, QSettings
 from qgis.PyQt.QtWidgets import QApplication, QDialog, QDialogButtonBox, QFrame, QMessageBox, QLabel, QVBoxLayout
 from qgis.PyQt.QtNetwork import QNetworkRequest
 
@@ -556,7 +556,20 @@ class QgsPluginInstaller(QObject):
         pluginFileName = os.path.splitext(os.path.basename(filePath))[0]
 
         if not pluginName:
-            QMessageBox.warning(iface.mainWindow(), self.tr("QGIS Python Install from ZIP Plugin Installer"), self.tr('The plugin directory was not found inside the ZIP file. You should create a plugin folder, put the files inside and create the ZIP file again.'), QMessageBox.Ok)
+            import webbrowser
+            import re
+            msg_box = QMessageBox()
+            msg_box.setIcon(QMessageBox.Warning)
+            msg_box.setWindowTitle(self.tr("QGIS Python Install from ZIP Plugin Installer"))
+            msg_box.setText(self.tr("The Zip file is not a valid QGIS python plugin. No root folder was found inside."))
+            msg_box.setStandardButtons(QMessageBox.Ok)
+            more_info_btn = msg_box.addButton(self.tr("More Information"), QMessageBox.HelpRole)
+            msg_box.exec()
+            version = 'testing' if 'master' in Qgis.QGIS_VERSION.lower() else re.findall(r'^\d.[0-9]*', Qgis.QGIS_VERSION)[0]
+            locale = QSettings().value("locale/userLocale", type=str)
+            qgis_lang = str( locale[:2] )
+            if msg_box.clickedButton() == more_info_btn:
+                webbrowser.open("https://docs.qgis.org/{version}/{lang}/docs/user_manual/plugins/plugins.html#the-install-from-zip-tab".format(version=version, lang=qgis_lang))
             return
 
         pluginsDirectory = qgis.utils.home_plugin_path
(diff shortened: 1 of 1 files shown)


python/pyplugin_installer/installer.py
```diff
@@ -555,6 +555,10 @@ def installFromZipFile(self, filePath):
 
         pluginFileName = os.path.splitext(os.path.basename(filePath))[0]
 
+        if not pluginName:
+            QMessageBox.warning(iface.mainWindow(), self.tr("QGIS Python Install from ZIP Plugin Installer"), self.tr('The plugin directory was not found inside the ZIP file. You should create a plugin folder, put the files inside and create the ZIP file again.'), QMessageBox.Ok)
```
m-kuhn: @DelazJ can you check that for the wording?

I would suggest to use something like "The ZIP file is not a valid QGIS python plugin. No root folder was found inside." (i.e.  I would not assume that the one who installs the plugin is the one who developed the plugin).

m-kuhn (commented):
Looks good from a technical perspective.
I think we need to fine tune the wording before merging.

python/pyplugin_installer/installer.py
m-kuhn: Ping @DelazJ

DelazJ: Oups, Sorry! I forgot that one. Though not sure how my wording in English can be better than yours, @m-kuhn :thinking: .
Anyways...  I agree that the person who installs is not necessarily the developer, and because of that I think we should give them some hints (what's a valid plugin structure?) to avoid them unnecessarily wait for the next point release of the plugin. And my personal use of this tool was to download a branch from github so I guess, no fix to expect from the dev and as far as i remember I had to manually rename things to get it work. Maybe we need some magic in the tool to have a proper name or, keep your rewording with a "more info" that would point to https://docs.qgis.org/3.4/en/docs/user_manual/plugins/plugins.html#the-install-from-zip-tab where we could explain what a proper structure of zip file should be.
Not sure whether i'm not out of the scope of this PR... :thinking: :

seralra96: @DelazJ I think this would be great (add more info), if apart from the message is explained to the user what is the desired structure for the zip, could be achieved in a next attempt not make the same mistake again.

I can change the wording like @m-kuhn says and i can make the docs to explain the desired structure, we can do this in this PR if you think that is ok.

python/pyplugin_installer/installer.py
```diff
@@ -556,20 +556,15 @@ def installFromZipFile(self, filePath):
...
             msg_box.setText(self.tr("The Zip file is not a valid QGIS python plugin. No root folder was found inside."))
             msg_box.setStandardButtons(QMessageBox.Ok)
             more_info_btn = msg_box.addButton(self.tr("More Information"), QMessageBox.HelpRole)
             msg_box.exec()
-            version = 'testing' if 'master' in Qgis.QGIS_VERSION.lower() else re.findall(r'^\d.[0-9]*', Qgis.QGIS_VERSION)[0]
-            locale = QSettings().value("locale/userLocale", type=str)
-            qgis_lang = str( locale[:2] )
             if msg_box.clickedButton() == more_info_btn:
-                webbrowser.open("https://docs.qgis.org/{version}/{lang}/docs/user_manual/plugins/plugins.html#the-install-from-zip-tab".format(version=version, lang=qgis_lang))
+                QgsHelp.openHelp("plugins/plugins.html#the-install-from-zip-tab")
```
seralra96: As @DelazJ suggested

python/pyplugin_installer/installer.py
```diff
@@ -556,7 +556,20 @@ def installFromZipFile(self, filePath):
...
+            msg_box.setWindowTitle(self.tr("QGIS Python Install from ZIP Plugin Installer"))
+            msg_box.setText(self.tr("The Zip file is not a valid QGIS python plugin. No root folder was found inside."))
+            msg_box.setStandardButtons(QMessageBox.Ok)
+            more_info_btn = msg_box.addButton(self.tr("More Information"), QMessageBox.HelpRole)
+            msg_box.exec()
+            version = 'testing' if 'master' in Qgis.QGIS_VERSION.lower() else re.findall(r'^\d.[0-9]*', Qgis.QGIS_VERSION)[0]
+            locale = QSettings().value("locale/userLocale", type=str)
+            qgis_lang = str( locale[:2] )
+            if msg_box.clickedButton() == more_info_btn:
+                webbrowser.open("https://docs.qgis.org/{version}/{lang}/docs/user_manual/plugins/plugins.html#the-install-from-zip-tab".format(version=version, lang=qgis_lang))
```
m-kuhn: I think that's a candidate for QgsHelp

```
QgsHelp.openHelp( "plugins/plugins.html#the-install-from-zip-tab" )
```

m-kuhn:
Thank you @seralra96

DelazJ:
Yes, Thank you @seralra96 . If ever you need guidance for the docs update, do not hesitate.

PR #31387, opened 2019-08-24
[needs-docs] Show 'template layer' constraints info in the Refactor fields algorithm's UI

### LINKED ISSUE #34458: Refactor fields algorithm - show fields with constraints
When using the refactor fields algorithm to prepare data for an import into an existing layer, it would be nice to see more information about the target layer.
Most important NOT NULL constraints (but actually any information like comments, allowed number ranges, data type, widget configuration, ... would be interesting)

This information is very valuable at configuration time as it allows to make data conformant to a target layer without repeatedly closing and reopening this (modal) dialog and the layer properties (modal) dialog.

I see that https://github.com/qgis/QGIS/pull/31387 already implemented this and would vote to reopen this one. If this is seen as too intrusive, we could connect the visual cues to a checkbox.

CC @nyalldawson @arnaud-morvan @gacarrillor

gacarrillor:
## Description
Dialog: Refactor fields algorithm's UI.
Objective: Highlight field rows when the 'template layer' field has constraints.

**TODO**
 + [x] Add an orange color (same as used in attribute form warnings) to the whole row if a template layer field has constraints.
 + [x] Adjust text in help panel to let users know why some rows are coloured. 
 + [x] Remove NOT NULL column.
 + [x] New column "Template properties."
     + Which content to show?
        + "Constraints active"?
     + What the tooltip should show?
        + Constraints, widget configuration, ...?
 + [x] Replace "Load fields from layer" by "Load fields from template layer."

**Screenshots**

image

Fix #34458

## Checklist

- [x] I have evaluated whether it is appropriate for this PR to be backported, backport requests are left as label or comment.

### DIFF
--- a/python/plugins/processing/algs/help/qgis.yaml
+++ b/python/plugins/processing/algs/help/qgis.yaml
@@ -330,6 +330,8 @@ qgis:refactorfields: >
 
   The original layer is not modified. A new layer is generated, which contains a modified attribute table, according to the provided fields mapping.
 
+  Rows in orange are not allowed to be NULL in the layer used as target. However, since the algorithm creates a new layer with no constraints, this can be taken only as a hint.
+
 qgis:regularpoints: >
   This algorithm creates a point layer with a given number of regular points, all of them within a given extent.
 
--- a/python/plugins/processing/algs/help/qgis.yaml
+++ b/python/plugins/processing/algs/help/qgis.yaml
@@ -330,7 +330,7 @@ qgis:refactorfields: >
 
   The original layer is not modified. A new layer is generated, which contains a modified attribute table, according to the provided fields mapping.
 
-  Rows in orange are not allowed to be NULL in the layer used as target. However, since the algorithm creates a new layer with no constraints, this can be taken only as a hint.
+  Rows in orange have constraints in the template layer from which these fields were loaded. Treat this information as a hint during configuration. No constraints will be added on an output layer nor will this be checked or enforced by the algorithm.
 
 qgis:regularpoints: >
   This algorithm creates a point layer with a given number of regular points, all of them within a given extent.
--- a/python/plugins/processing/algs/help/qgis.yaml
+++ b/python/plugins/processing/algs/help/qgis.yaml
@@ -330,7 +330,7 @@ qgis:refactorfields: >
 
   The original layer is not modified. A new layer is generated, which contains a modified attribute table, according to the provided fields mapping.
 
-  Rows in orange have constraints in the template layer from which these fields were loaded. Treat this information as a hint during configuration. No constraints will be added on an output layer nor will this be checked or enforced by the algorithm.
+  Rows in orange have constraints in the template layer from which these fields were loaded. Treat this information as a hint during configuration. No constraints will be added on an output layer nor will they be checked or enforced by the algorithm.
 
 qgis:regularpoints: >
   This algorithm creates a point layer with a given number of regular points, all of them within a given extent.
--- a/python/plugins/processing/algs/qgis/ui/FieldsMappingPanel.py
+++ b/python/plugins/processing/algs/qgis/ui/FieldsMappingPanel.py
@@ -217,7 +217,7 @@ class FieldsMappingModel(QAbstractTableModel):
 
         if role == Qt.ToolTipRole:
             if column_def['name'] == 'constraints' and 'constraints' in field:
-                return ", ".join([self.constraints[constraint] for constraint in field['constraints']])
+                return "<br>".join([self.constraints[constraint] for constraint in field['constraints']])
 
     def setData(self, index, value, role=Qt.EditRole):
         field = self._mapping[index.row()]
--- a/python/plugins/processing/algs/qgis/ui/FieldsMappingPanel.py
+++ b/python/plugins/processing/algs/qgis/ui/FieldsMappingPanel.py
@@ -34,6 +34,12 @@ from qgis.PyQt.QtCore import (
     pyqtSlot,
     QCoreApplication
 )
+
+from qgis.PyQt.QtGui import (
+    QBrush,
+    QColor
+)
+
 from qgis.PyQt.QtWidgets import (
     QComboBox,
     QHeaderView,
@@ -54,6 +60,7 @@ from qgis.core import (
     QgsProcessingUtils,
     QgsProject,
     QgsVectorLayer,
+    QgsFieldConstraints
 )
 from qgis.gui import QgsFieldExpressionWidget
 
@@ -109,6 +116,10 @@ class FieldsMappingModel(QAbstractTableModel):
             'name': 'precision',
             'type': QVariant.Int,
             'header': self.tr("Precision")
+        }, {
+            'name': 'not_null',
+            'type': QVariant.Bool,
+            'header': self.tr("NOT NULL")
         }]
 
     def columnIndex(self, column_name):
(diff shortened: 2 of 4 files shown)


python/plugins/processing/algs/help/qgis.yaml
```diff
@@ -466,6 +466,8 @@ qgis:refactorfields: >
 
   The original layer is not modified. A new layer is generated, which contains a modified attribute table, according to the provided fields mapping.
 
+  Rows in orange are not allowed to be NULL in the layer used as target. However, since the algorithm creates a new layer with no constraints, this can be taken only as a hint.
```
nyalldawson: I really don't see the value of this given the listed limitation. It's very confusing for users, who would expect this flag to apply to the newly created layer.

gacarrillor: @nyalldawson, if you use the Refactor Fields algorithm, you know it's a bit overwhelming, this PR aims to make it easier for users to prioritize field mappings. 

We actually use the refactor fields along with an "Append Features to Layer" algorithm (3rd party plugin that I also suggested to port to the QGIS core) to send the result into an existing layer. So, it acts as an entire ETL algorithm where knowing the target field constraints makes (a lot of) sense, especially for layers with a large number of fields.

nyalldawson: I still don't see the value sorry, and foresee the inevitable bug reports incoming about how the value isn't reflected in the output layer. Maybe for your particular use case you are best off subclassing the existing algorithm into your plugin so that it can be designed to meet your particular use case, rather then changing the core one itself.

arnaud-morvan: I agree with @nyalldawson that this is very confusing.
This only give the information that the layer used as a template for fields have not null constraints, but the purpose of this algorithm is to create a new layer.
This would make sense in an algorithm that writes features in an existing layer.

gacarrillor:
Understood, closing...

m-kuhn:
Reopened for reconsideration, as this has been seen as a wider requirement (i.e. this is not an isolated use case and therefore a generic approach as proposed in here is preferable over a subclassing approach). As stated in the below ticket, if this is considered as too intrusive, the coloring can be made optional by a switch / checkbox.

https://github.com/qgis/QGIS/issues/34458

python/plugins/processing/algs/help/qgis.yaml
m-kuhn:
```suggestion
  Rows in orange have a NOT NULL constraint in the layer from which these fields were loaded.
  Treat this information as a hint during configuration. No constraints will be added on an output layer nor will this be checked or enforced by the algorithm.
```

I think that's slightly more on topic.

nyalldawson:
Honestly, I'm still -1. If this is crucial functionality, I think it needs to be exposed in a much less confusing way. My feeling is always that if **I'm** (as a qgis developer) struggling to understand an option in qgis, a regular user will find that option incredibly confusing...

m-kuhn:
It is very helpful. If you have an existing target layer (or template thereof), showing it's constraints is very helpful. Especially since the modeler dialog is modal.

I think the most misleading part is the "NOT NULL" column. It looks very much like the other columns which result in real column properties of the created layer. Would it be sufficient to just remove that?

nyalldawson:
> It is very helpful. If you have an existing target layer (or template thereof)

Can you clarify what you mean by "target layer"? Currently there's no concept of that at all, you only have a source layer.

m-kuhn:
I see the layer specified in "load fields from layer" as a "template" of the target layer.

Situation:
Customer has an old database and a new database schema and wants to migrate data from db version 1 (source) to db version 2 (target). He loads both layers and uses refactor fields to make data match. To configure the algorithm the constraint information of the "template target layer" at hand is a big help.

Currently this means repeatedly closing/reopening this dialog and layer properties to check for constraints (because both dialogs are modal they cannot be opened side by side).

Currently this also means a final manual step to copy/paste all features into the target layer in the end (or use the append to layer plugin as mentioned by @gacarrillor). This could in the future be done automatically but even then will never be within the scope of this algorithm.

nyalldawson:
> I see the layer specified in "load fields from layer" as a "template" of the target layer.

Ok fair enough, I'll concede to that :smile:

But can we improve the UI here? I'm still concerned that the original way of exposing this is going to be very confusing for users. Just brainstorming, but what about a new column for "Source Properties" which shows a tooltip with detailed explanatory text when hovered over?

m-kuhn:
Agreed on improving the UX.
An additional column with this information would be good. I'd avoid the term "source" though to avoid confusion. It might be good to introduce a new term here. Adding "template" to the load fields from layer label and using template in the column header could help a user link the two.

nyalldawson:
> An additional column with this information would be good. I'd avoid the term "source" though to avoid confusion. It might be good to introduce a new term here. Adding "template" to the load fields from layer label and using template in the column header could help a user link the tw

Yes, I like this idea!

gacarrillor:
Happy to see how this PR has evolved.
I can continue working on it. I've written the required adjustments down in the PR's description. 
Feel free to suggest changes/adjustments to that list.

python/plugins/processing/algs/help/qgis.yaml
gacarrillor: Added a very similar text in the end, but now not only limited to NOT NULL constraints.

python/plugins/processing/algs/help/qgis.yaml
```diff
@@ -330,7 +330,7 @@ qgis:refactorfields: >
 
   The original layer is not modified. A new layer is generated, which contains a modified attribute table, according to the provided fields mapping.
 
-  Rows in orange are not allowed to be NULL in the layer used as target. However, since the algorithm creates a new layer with no constraints, this can be taken only as a hint.
+  Rows in orange have constraints in the template layer from which these fields were loaded. Treat this information as a hint during configuration. No constraints will be added on an output layer nor will this be checked or enforced by the algorithm.
```
m-kuhn:
```suggestion
  Rows in orange have constraints in the template layer from which these fields were loaded. Treat this information as a hint during configuration. No constraints will be added on an output layer nor will they be checked or enforced by the algorithm.
```

I think it makes it slightly more readable if it refers to the first part of the sentence.

python/plugins/processing/algs/qgis/ui/FieldsMappingPanel.py
```diff
@@ -199,7 +213,12 @@ def data(self, index, role=Qt.DisplayRole):
             return hAlign + Qt.AlignVCenter
 
         if role == Qt.BackgroundRole:
-            return QBrush(QColor(255, 224, 178)) if 'not_null' in field and field['not_null'] else QVariant()
+            return QBrush(QColor(255, 224, 178)) if 'constraints' in field and field['constraints'] else QVariant()
+
+        if role == Qt.ToolTipRole:
+            if column_def['name'] == 'constraints':
+                return ", ".join(
```
m-kuhn:
```suggestion
                return "<br> ".join(
```

Would this make it more readable?

python/plugins/processing/algs/help/qgis.yaml
gacarrillor: Applied!

python/plugins/processing/algs/qgis/ui/FieldsMappingPanel.py
gacarrillor: I think so. Applied on a single commit with the other suggestion. (This one shows an Outdated tag and it is not possible to commit it from GitHub)

m-kuhn:
@nyalldawson how does that look to you?

m-kuhn (approved):

PR #31396, opened 2019-08-25
Replace paths pointing to inbuilt data folder with "inbuilt:" prefix

nyalldawson:
...during project write, and redirect paths beginning with "inbuilt:" prefix to actual local install folder during project read

Allows projects using data like the built-in world map layer to work correctly across different installs


nirvn:
Good idea

PR #31399, opened 2019-08-25
Display Help button on the 3D Map Configuration dialog

DelazJ:
I'm trying to add a help button to the 3D Map Configuration dialog accessible from the 3D Map view. It does seem to work but while the "showhelp" function used to be a private function, I had to make it public since I could find no way to get to it from the canvas widget.
Should there be a 3dMapConfigDialog class? Also, the dialog does not remember its previous geometry (you have to always resize it to get it in a workable size) and I wonder if a class would have made it.
Sorry for the noob questions, if any.

## Checklist

D

### DIFF
--- a/src/app/3d/qgs3dmapconfigwidget.h
+++ b/src/app/3d/qgs3dmapconfigwidget.h
@@ -33,6 +33,8 @@ class Qgs3DMapConfigWidget : public QWidget, private Ui::Map3DConfigWidget
     explicit Qgs3DMapConfigWidget( Qgs3DMapSettings *map, QgsMapCanvas *mainCanvas, QWidget *parent = nullptr );
 
     void apply();
+    //! Displays help on the 3D Map configuration
+    void showHelp();
 
   signals:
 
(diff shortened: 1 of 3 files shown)


src/app/3d/qgs3dmapconfigwidget.h
```diff
@@ -33,6 +33,8 @@ class Qgs3DMapConfigWidget : public QWidget, private Ui::Map3DConfigWidget
     explicit Qgs3DMapConfigWidget( Qgs3DMapSettings *map, QgsMapCanvas *mainCanvas, QWidget *parent = nullptr );
 
     void apply();
+    //! Displays help on the 3D Map configuration
+    void showHelp();
```
nyalldawson: As far as I can see this isn't called from Qgs3DMapConfigWidget. Why not place it in Qgs3DMapCanvasDockWidget instead?

DelazJ: @nyalldawson Because I fail to find the proper way to write the connect code (I don't remember the build complaints, with my attempts). As indicated in my initial comment, there used to be a "qgs...Dialog" class for almost all the widgets I had to add the help button to, so here I'm a bit in an unknown area with a dialog created "on the fly" (whose resizing is btw always reset).

Another thing is that the help button is placed on the Config dialog, and should actually address the config options description (even if I pointed it to the 3D Map View section since there's no help button there (- yet?))

PR #31402, opened 2019-08-25
Save/Restore Shape digitize active tool

DelazJ:
Because the tool i often use is not necessarily the default...

## Checklist

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -2975,7 +2975,17 @@ void QgisApp::createToolBars()
   tbAddCircularString->setPopupMode( QToolButton::MenuButtonPopup );
   tbAddCircularString->addAction( mActionCircularStringCurvePoint );
   tbAddCircularString->addAction( mActionCircularStringRadius );
-  tbAddCircularString->setDefaultAction( mActionCircularStringCurvePoint );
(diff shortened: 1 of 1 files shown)


src/app/qgisapp.cpp
```diff
@@ -14293,6 +14367,42 @@ void QgisApp::toolButtonActionTriggered( QAction *action )
     settings.setEnumValue( QStringLiteral( "UI/defaultVertexTool" ), QgsVertexTool::AllLayers );
   else if ( action == mActionVertexToolActiveLayer )
     settings.setEnumValue( QStringLiteral( "UI/defaultVertexTool" ), QgsVertexTool::ActiveLayer );
+  else if ( action == mActionCircularStringCurvePoint )
+    settings.setValue( QStringLiteral( "UI/defaultCircularString" ), 0 );
```
m-kuhn: Any reason not to use `setEnumValue`?

DelazJ: Yes! I don't know what it is nor how that plays! :thinking:

m-kuhn: It saves enum names instead of magic numbers.
But sorry, that won't work here, since there are no enum involved.

PR #31405, opened 2019-08-25
Sketch line symbol rendering

wonder-sk:
This is a work in progress... not ready for inclusion in core yet. 

Original (sketch lines disabled):

image

Sketch lines enabled:

image

Sketch lines enabled - different sketch parameter values:

image

Sketch line rendering - when enabled, there are three parameters to tweak for different appearance:
- roughness
- bowing
- max offset

Code ported from https://github.com/gicentre/handy/blob/master/Handy/src/org/gicentre/handy/HandyRenderer.java

Open questions / issues:
- have this as an optional feature for "simple line" symbol or shall we have a new "sketch line" symbol?
- the output is random so the rendered output changes on every map refresh
- any volunteers to implement sketch fills (hachures) ? :-)

cc @andreasneumann @timlinux @nirvn

### DIFF
--- a/python/core/auto_generated/symbology/qgslinesymbollayer.sip.in
+++ b/python/core/auto_generated/symbology/qgslinesymbollayer.sip.in
@@ -229,6 +229,16 @@ used to render polygon rings.
 .. seealso:: :py:func:`drawInsidePolygon`
 %End
 
+    bool isSketchEnabled() const;
+    float sketchRoughness() const;
+    float sketchBowing() const;
+    float sketchMaxOffset() const;
+
+    void setSketchEnabled( bool enabled );
+    void setSketchRoughness( float roughness );
+    void setSketchBowing( float bowing );
+    void setSketchMaxOffset( float maxOffset );
+
 };
 
 
--- a/src/core/symbology/qgslinesymbollayer.cpp
+++ b/src/core/symbology/qgslinesymbollayer.cpp
@@ -177,6 +177,15 @@ QgsSymbolLayer *QgsSimpleLineSymbolLayer::create( const QgsStringMap &props )
     l->setRingFilter( static_cast< RenderRingFilter>( props[QStringLiteral( "ring_filter" )].toInt() ) );
   }
 
+  if ( props.contains( "sketch" ) )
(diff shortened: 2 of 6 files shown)


andreasneumann:
Cool! As discussed we need a way to predictably render the styles in the same way for repeated renderings: important for tiling, etc. You suggested deriving randomness from the geometry vertices, or some checksums derived from geometries?

nyalldawson:
Have you considered exposing this as an expression function instead? To me it seems a better fit within geometry generators vs dedicated settings in the symbols. Otherwise we'll be chasing our tails forever - someone will want this with marker line symbols, then shapeburst, then raster fill, etc...

nirvn:
@nyalldawson , while I don't mind having this sketch thing exposed as a geometry expression (which would output a multiline feature?), I'm -1 to kill the symbol layer offering. For one, geometry generator symbol layers are dramatically slower, but maybe more importantly, it's just not as discoverable and easy to use.

For the marker outline, the filled marker symbol layer will allow for marker borders to have sketchy style, correct?

@wonder-sk , make it a different symbol layer altogether, there are already too many parameters to IMHO have it attached to simple line. And, as we chatted on hangout, we need a fixed seed value to avoid randomness across rendering.

nyalldawson:
Yeah, ok. After I looked at the actual code I can't see how this could fit into an expression function anyway, since it's all about painter path manipulation.

> And, as we chatted on hangout, we need a fixed seed value to avoid randomness across rendering.

I think there's a need for a "random seed" member in QgsRenderContext, which could help here....

nirvn:
@nyalldawson , hm, I like this idea of a random seed member!

roya0045:
How would this random seed be managed and how much of an impact should it have?

I assume it wouldn't be shareable but should it be saved in the project or reset everytime?

nirvn:
@wonder-sk , any chance you could push this through the finish line prior to freeze?

wonder-sk:
@nirvn unfortunately I am too busy these days catching up with work after the hackfest + foss4g. But if you have some spare time please feel free to finish it... I have pushed the existing code to lower the chance that I will forget about it completely and to get some feedback on it :-)

PR #31412, opened 2019-08-26
[needs-docs] Labels toolbar actions no longer need the layer to be in…

DelazJ:
… edit mode

## Checklist

### DIFF
--- a/src/ui/qgisapp.ui
+++ b/src/ui/qgisapp.ui
@@ -1886,7 +1886,7 @@
    </property>
    <property name="toolTip">
     <string>Moves a Label or Diagram
-Acts on all editable layers.</string>
+Acts on all layers.</string>
    </property>
   </action>
   <action name="mActionRotateLabel">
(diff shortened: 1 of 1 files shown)


nyalldawson (approved):

src/ui/qgisapp.ui
```diff
@@ -1886,7 +1886,7 @@
    </property>
    <property name="toolTip">
     <string>Moves a Label or Diagram
-Acts on all editable layers.</string>
+Acts on all layers.</string>
```
m-kuhn: What is the added value of leaving this message? Isn't it just stating the obvious and it would better be removed?

nyalldawson: Some of the labeling tools operate on the active layer only -- those also have an explanation in their tooltip reflecting this.

m-kuhn: We can also do that, but then we'll have to do it for all the feature tools etc. too I think?

nyalldawson: Sounds like a good idea in general to me!

PR #31413, opened 2019-08-26
Fix single feature renderer symbol levels reset when changing style

### LINKED ISSUE #26547: Symbol levels reset when changing style settings
Author Name: **Klas Karlsson** (@klakar)
Original Redmine Issue: [18659](https://issues.qgis.org/issues/18659)
Affected QGIS version: 3.0.0
Redmine category:symbology

---
A line layer with a symbol layer order that is +not+ in the default order will apply properly, but when a style property for the line is changed, the symbol levels are reset to the default order.

Not sure if this is a regression...

Style with three layers, and no symbol levels.

Applying symbol levels so that drawing order change works.

But if I change for example the line width for anything in the styles, the symbol levels will revert to the starting drawing order for symbol levels. For complex rule based layers with multiple styles and symbol levels, this is a problem.

WORKAROUND:
Make sure your symbol layers are in the correct drawing order before applying symbol levels. 

It may be a small problem, but symbol level reordering feels broken...

---

- [levels.png](https://issues.qgis.org/attachments/download/12577/levels.png) (Klas Karlsson)
- [no_levels.png](https://issues.qgis.org/attachments/download/12578/no_levels.png) (Klas Karlsson)

nirvn:
## Description

This PR fixes #26547, whereas the symbol levels were reset to the values set when opening the style panel / layer properties window whenever the layer symbol was modified.

I've taken the opportunity to cleanup the symbol selector {widget, dialog} class a bit.

## Checklist

### DIFF
--- a/src/gui/symbology/qgssymbolselectordialog.h
+++ b/src/gui/symbology/qgssymbolselectordialog.h
@@ -128,57 +128,11 @@ class GUI_EXPORT QgsSymbolSelectorWidget: public QgsPanelWidget, private Ui::Qgs
     QgsSymbol *symbol() { return mSymbol; }
 
     /**
-     * Load the given symbol into the widget.
+     * Loads the given symbol into the widget.
      * \param symbol The symbol to load.
      * \param parent The parent symbol layer item. If the parent parameter is null, the whole symbol and model will be reset.
-     * \note not available in Python bindings
      */
     void loadSymbol( QgsSymbol *symbol, SymbolLayerItem *parent = nullptr ) SIP_SKIP;
-
-  protected:
-
-    /**
-     * Reload the current symbol in the view.
-     */
-    void loadSymbol();
-
-    /**
-     * Update the state of the UI based on the currently set symbol layer.
-     */
-    void updateUi();
-
-    /**
-     * Update the lock button states based on the current symbol layer.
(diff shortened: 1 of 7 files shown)


src/gui/symbology/qgssymbolselectordialog.h
```diff
@@ -127,21 +127,21 @@ class GUI_EXPORT QgsSymbolSelectorWidget: public QgsPanelWidget, private Ui::Qgs
      */
     QgsSymbol *symbol() { return mSymbol; }
 
+    /**
+     * Load the given symbol into the widget.
+     * \param symbol The symbol to load.
+     * \param parent The parent symbol layer item. If the parent parameter is null, the whole symbol and model will be reset.
+     * \note not available in Python bindings
+     */
+    void loadSymbol( QgsSymbol *symbol, SymbolLayerItem *parent = nullptr ) SIP_SKIP;
```
nyalldawson: Can you add a TODO QGIS 4 note here saying "QGIS 4: transfer ownership of symbol", and in the dox note that ownership is NOT transferred and that the symbol MUST exist for the lifetime of the widget

nirvn: Done.

Issue #31410, opened 2019-08-25
3D Symbol Y rotation no persisted after applying


timlinux:
In QGIS 3.9 on macOS

![3dstate](https://user-images.githubusercontent.com/178003/63656865-71d62600-c7a2-11e9-9a5c-1d6dcf238c7d.gif)

When I set the z rotation then save the layer properties the rotation does not get saved and when I open the properties dialog again, it has reset.

PR #31416, opened 2019-08-26
Fix loading of rotation in 3d point symbols in GUI

wonder-sk:
Fix for #31410

The conversion to Euler angles may give angles in range (-180, 180)
but spin boxes in GUI only expect range (0, 360)

cc @timlinux


timlinux:
didn't realise that! Thanks

PR #31423, opened 2019-08-26
Prevents decoding WMS capabilities requests

### LINKED ISSUE #31192: Add WMS server fails with QGIS Server with Postgresql project
**Context**

QGIS desktop should be able to work with QGIS Server, if the project is stored in Postgresql.

When QGIS Server is used with several different projects, the MAP parameter must be present to tell which project to use.

The following GetCapabilities requests are used against two different projects: one stored on the server filesystem and another stored in Postgresql:
1. http://qgis.demo/cgi-bin/qgis_mapserv.fcgi?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetCapabilities&MAP=/home/qgis/projects/cartageologica.qgz
2. http://qgis.demo/cgi-bin/qgis_mapserv.fcgi?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetCapabilities&MAP=postgresql%3A%3Fservice%3Dpg_geotuga%26sslmode%3Ddisable%26dbname%3D%26schema%3Dortos%26project%3Dteste

The second URL, which includes the Postgresql project, **must be properly quoted** to be properly handled on the QGIS Server side:
```python
python -c "import urllib, sys; print urllib.quote(sys.argv[1])" "postgresql:?service=pg_geotuga&sslmode=disable&dbname=&schema=ortos&project=teste"
```

**Describe the bug**

There is no problem to add a QGIS Server url to QGIS desktop, using a `MAP=<project on filesystem>`.

wms with project from filesystem

There is no way to add QGIS Server urls with `MAP=<Postgresql project>`.

error add qis server with postgresql project

I've tried with both the quoted and unquoted URL:
1. http://qgis.demo/cgi-bin/qgis_mapserv.fcgi?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetCapabilities&MAP=postgresql%3A%3Fservice%3Dpg_geotuga%26sslmode%3Ddisable%26dbname%3D%26schema%3Dortos%26project%3Dteste
1. http://qgis.demo/cgi-bin/qgis_mapserv.fcgi?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetCapabilities&MAP=postgresql:?service=pg_geotuga&sslmode=disable&dbname=&schema=ortos&project=teste

The error is the same. The string passed to QGIS server is the same.

**How to Reproduce**

1. Go to Add WMS/WMTS layer
2. Add a QGIS Server with a URl that includes the Postgresql project
2. Click connect to get the capabilities document
2. The GetCapabilities fails.

**QGIS and OS versions**

1. QGIS 3.8, 3.9-Master
1. Ubuntu 19.04
1. GDAL 3.0.1

**Additional context**

The server is able to answer only if the MAP string is quoted.

The entry point in the source code is the [`QgsWmsSettings::parseUri`](https://github.com/qgis/QGIS/blob/bf31ce1d8cf2e24e3d2ef499a15afa44f9eb5780/src/providers/wms/qgswmscapabilities.cpp#L38) function.

This calls `uri.setEncodedUri( uriString );` which does not properly quote the project part and send the request to the server like:
`QUERY_STRING: VERSION=1.3.0&MAP=postgresql:?service=pg_geotuga&sslmode=disable&dbname=&schema=ortos&project=teste&SERVICE=WMS&REQUEST=GetCapabilities`
The server breaks the parameters using the `&` character and the MAP becomes only:
`21:04 INFO Server[7789]: MAP:postgresql:?service=pg_geotuga`.

There are several possibilities...
1. Change the postgresql project string format, removing the `&` (this might break existing projects, but using `&` will be always a problem). Better now than later.
2. Make this `uri.setEncodedUri( uriString );` function aware of the MAP parameter and create a proper quoted MAP string.
3. Improve QGIS Server to recreate the proper MAP parameter. QGIS Server receives all the parameters, but separated by `&`.

jgrocha:
## Description

When a user wants to create a new WMS/WMTS connection, he must provide the WMS end point. If the end point needs any additional parameter, it is users's responsibility to provide it properly encoded.

Right now, if the user wants to provide the end point of a WMS QGIS server, adding an additional parameter MAP with a Postgresql stored project, QGIS Desktop will fail to retrieve the Capabilities document.

Fix #31192.

## Solution

To "fix" this issue, I've changed only the code related to the GetCapabilities request. I avoided more changes that might break other calls to the WMS/WMTS service.

## Checklist

### DIFF
--- a/src/providers/wms/qgswmsprovider.cpp
+++ b/src/providers/wms/qgswmsprovider.cpp
@@ -164,6 +164,24 @@ QgsWmsProvider::QgsWmsProvider( QString const &uri, const ProviderOptions &optio
   QgsDebugMsg( QStringLiteral( "exiting constructor." ) );
 }
 
+QString QgsWmsProvider::prepareUriWithoutDecode( QString uri )
+{
+  if ( uri.contains( QLatin1String( "SERVICE=WMTS" ) ) || uri.contains( QLatin1String( "/WMTSCapabilities.xml" ) ) )
+  {
+    return uri;
+  }
+
+  if ( !uri.contains( QLatin1String( "?" ) ) )
+  {
+    uri.append( '?' );
+  }
+  else if ( uri.right( 1 ) != QLatin1String( "?" ) && uri.right( 1 ) != QLatin1String( "&" ) )
+  {
+    uri.append( '&' );
+  }
+
+  return uri;
+}
 
 QString QgsWmsProvider::prepareUri( QString uri )
 {
--- a/src/providers/wms/qgswmsprovider.cpp
+++ b/src/providers/wms/qgswmsprovider.cpp
@@ -164,29 +164,13 @@ QgsWmsProvider::QgsWmsProvider( QString const &uri, const ProviderOptions &optio
   QgsDebugMsg( QStringLiteral( "exiting constructor." ) );
 }
 
-QString QgsWmsProvider::prepareUriWithoutDecode( QString uri )
-{
-  if ( uri.contains( QLatin1String( "SERVICE=WMTS" ) ) || uri.contains( QLatin1String( "/WMTSCapabilities.xml" ) ) )
-  {
-    return uri;
-  }
-
-  if ( !uri.contains( QLatin1String( "?" ) ) )
-  {
-    uri.append( '?' );
-  }
-  else if ( uri.right( 1 ) != QLatin1String( "?" ) && uri.right( 1 ) != QLatin1String( "&" ) )
-  {
-    uri.append( '&' );
-  }
-
-  return uri;
-}
-
-QString QgsWmsProvider::prepareUri( QString uri )
+QString QgsWmsProvider::prepareUri( QString uri, bool decode )
 {
   // some services provide a percent/url encoded (legend) uri string, always decode here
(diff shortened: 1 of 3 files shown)


rldhont:
Does a unit test exist ?

jgrocha:
> Does a unit test exist ?

That would be better, I agree. We can postpone this PR, if you think it is important.

It would make sense to add a test to raise an error in the current implementation, if the GetCapabilities URL is decoded before being sent to the remote WMS.

There is no interest in decoding URLs provided by users. The users either copies the URL from some metadata service (and we should keep it as pasted) or the user knows how to encoded it properly, if he wants to pass non standard parameters to the server.

rldhont:
@jgrocha can you add a test in this file tests/src/providers/testqgswmsprovider.cpp

Same issue probably in WFS and WCS providers, isn't it ?

jgrocha:
Thanks @rldhont for pushing me. I started to work on the tests and the WMS client breaks in more places. This fix would only allow the first GetCapabilities to work. More GetCapabilities are issued against the server and those fail. Probably other providers, as you said, will also fail.

I'll closed it now.

Maybe it is easier to provide an alternative syntax for Postgresql stored projects, without the exclamation mark and ampersands.

PR #31432, opened 2019-08-26
[needs-docs]Set the service area algorithms speed to 50km/h

DelazJ:
using the same value as the "Shortest Path" algorithms
I don't know the reason of 5km/h hence assume it's a typo (50km/h?) Also 50km/h represents the mean allowed speed in urban areas here so it has some meanings to me. and it's what is used in the three other network algorithms
## Checklist


nyalldawson:
Unrelated Travis failure

PR #31445, opened 2019-08-27
Expose font kerning setting to text format widget/label settings

nyalldawson:
Allows users to control whether label fonts are kerned or not


nirvn (approved):

nyalldawson:
Funnily enough - this setting just goes to show that Qt's handling of kerning is really quite poor in general :(

Issue #31443, opened 2019-08-27
QGIS crashed while exiting and saving


valsigiara:
This is the trace:

## User Feedback

QGIS crashed while closing and saving	

## Report Details

*Crash ID*: 62f1fcd2045780cb9f127836e6b579b8665f3e63

*Stack Trace*
<pre>
proj_lpz_dist :
proj_lpz_dist :
QgsCoordinateTransform::transformPolygon :
QgsCoordinateTransform::transformPolygon :
QgsCoordinateTransform::QgsCoordinateTransform :
QHashData::free_helper :
QgsCoordinateTransform::addToCache :
QgsCoordinateTransform::invalidateCache :
QgsApplication::exitQgis :
QgisApp::~QgisApp :
CPLStringList::operator[] :
main :
BaseThreadInitThunk :
RtlUserThreadStart :
</pre>

*QGIS Info*
QGIS Version: 3.8.2-Zanzibar
QGIS code revision: 4470baa1a3
Compiled against Qt: 5.11.2
Running against Qt: 5.11.2
Compiled against GDAL: 2.4.1
Running against GDAL: 2.4.1

*System Info*
CPU Type: x86_64
Kernel Type: winnt
Kernel Version: 6.1.7601

jef-n:
Duplicate of #28104

PR #31446, opened 2019-08-27
make the crash ID clickable on GitHub

Gustry:
## Description

I'm trying to make the crash ID clickable on github.
For instance this report https://github.com/qgis/QGIS/issues/31443 would have a link to 
https://github.com/qgis/QGIS/search?q=62f1fcd2045780cb9f127836e6b579b8665f3e63&type=Issues

I didn't try this PR, I don't have Windows with the crash handler. Just trying to make things a little bit easier on github.

@gioman What do you think?

I was thinking about hiding this new `htmlToMarkdown` from SIP, because it's an incomplete implementation, quite simple.

## Checklist

### DIFF
--- a/src/core/qgsstringutils.cpp
+++ b/src/core/qgsstringutils.cpp
@@ -448,6 +448,28 @@ QString QgsStringUtils::insertLinks( const QString &string, bool *foundLinks )
   return converted;
 }
 
+QString QgsStringUtils::htmlToMarkdown( const QString &html )
+{
+
+  QString converted = html;
+  converted.replace( QLatin1String( "<br>" ), QLatin1String( "\n" ) );
+  converted.replace( QLatin1String( "<b>" ), QLatin1String( "**" ) );
+  converted.replace( QLatin1String( "</b>" ), QLatin1String( "**" ) );
+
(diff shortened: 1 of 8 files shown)


NathanW2:
Super good idea.  In fact when I build the crash handler that was the main vision but searching the old ticket system was a bit hit and miss so I didn't bother. Works well now though on Github.

gioman:
> @gioman What do you think?

nice!

src/core/qgsstringutils.cpp
```diff
@@ -448,6 +448,28 @@ QString QgsStringUtils::insertLinks( const QString &string, bool *foundLinks )
   return converted;
 }
 
+QString QgsStringUtils::htmlToMarkdown( const QString &html )
+{
+
+  QString converted = html;
+  converted.replace( QLatin1String( "<br>" ), QLatin1String( "\n" ) );
+  converted.replace( QLatin1String( "<b>" ), QLatin1String( "*" ) );
```
nyalldawson: Shouldn't this be "**"?

Gustry: Indeed, this was an error before.

PR #31451, opened 2019-08-27
[FEATURE] Add z distance

ismailsunni:
## Description
The contents:
- Adding Z-Distance to the 3D Measurement dialog (suggested by @timlinux )
- Fixing not clearing the measurement table when restart using right-click
z-distance

## Checklist

### DIFF
--- a/src/app/3d/qgs3dmaptoolmeasureline.cpp
+++ b/src/app/3d/qgs3dmaptoolmeasureline.cpp
@@ -159,6 +159,7 @@ void Qgs3DMapToolMeasureLine::handleClick( Qt3DRender::QPickEvent *event, const 
   {
     // Finish measurement
     mDone = true;
+    restart();
   }
   else if ( event->button() == Qt3DRender::QPickEvent::MiddleButton )
   {
@@ -239,6 +240,7 @@ void Qgs3DMapToolMeasureLine::restart()
   mPoints.clear();
   mDone = true;
   updateMeasurementLayer();
+  mDialog->resetTable();
 }
 
 void Qgs3DMapToolMeasureLine::undo()
--- a/src/app/3d/qgs3dmeasuredialog.cpp
+++ b/src/app/3d/qgs3dmeasuredialog.cpp
@@ -82,7 +82,7 @@ void Qgs3DMeasureDialog::addPoint()
     if ( !mTool->done() )
     {
       // Add new entry in the table
-      addMeasurement( lastDistance(), lastZDistance() );
+      addMeasurement( lastDistance(), lastZDistance(), lastHorisontalDistance() );
       mTotal += lastDistance();
       editTotal->setText( formatDistance( convertLength( mTotal, mDisplayedDistanceUnit ) ) );
     }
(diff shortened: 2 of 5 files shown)


wonder-sk:
Nice stuff... I am wondering if we should have also horizontal distance displayed there. Because now I am not sure what the numbers on the right mean - is it 3d distance or is it just horizontal distance? (because vertical distance is shown separately)

ismailsunni:
Hi @wonder-sk 
Hmm, that makes sense. Should I put three distance then: Z-distance, horizontal-distance, and 3D-distance?

Or will it just make the user confuse? Adding toggle to show the Z-distance and horizontal distance probably?

roya0045:
I think that the 3 distinct elements would be helpful, I assume that the default measure tool should cover the flat distance (not 3D view though..). But we would need a way to use height and full 3D with distinct button preferably, otherwise people might forget where to look for something if the label is not perfect.

I assume that inspiring yourself from the current measure tool would let users know that they can choose distinct measurement method and I would let the full 3D one as the 'top' icon for the 3D map.

wonder-sk:
@ismailsunni Yeah I guess a checkbox to show/hide also horizontal and vertical distances for each segment would be good...

ismailsunni:
hi @roya0045 thanks for the comment

Do you mean in the right of the line edit? Or the right refers to above the line edit?

wonder-sk:
Usually @nirvn has some good advice on good UX :-)

My suggestion would be to have just a single checkbox to show/hide both horizontal and vertical lengths (with default value to have them hidden). And maybe have that checkbox below the "total 3d distance" field, so it is less prominent?

Gustry:
@wonder-sk Do you think this PR will be reopen?

ismailsunni:
Hi @Gustry 
I have updated my code to make it simpler. Hopefully, next week I can push it.

But it seems only the core dev can re-open it, isn't it?

roya0045:
@ismailsunni don't you have a button below the comment box to open the PR again?

DelazJ:
@roya0045 when the PR is closed by the stale bot, common contributors seem to have no mean to reopen it. Could be worth someone check whether there's not a workaround to allow PR author to reopen (because sometimes, the PR is closed while we are waiting for feedback/merge from the core devs)

ismailsunni:
hi @roya0045 
Yeah, as @DelazJ said, I can't re-open it :)

Thanks @nyalldawson for reopening it.

@wonder-sk I have updated with a checkbox to show and hide the extra distance

image

image

roya0045:
> hi @roya0045
> Yeah, as @DelazJ said, I can't re-open it :)
> 
> Thanks @nyalldawson for reopening it.
> 
> @wonder-sk I have updated with a checkbox to show and hide the extra distance
> 
> image
> 
> image

Looking at your images, I'm beginning to question if the button is needed. The 3 column approach is nice and displaying only one feels like a waste of space. But if a button is needed I think this is fine.

nyalldawson:
@nirvn would value your opinion here, but I'm in agreement with @roya0045 that the checkbox adds unnecessary complexity vs just always showing the columns...

nirvn:
+1 to remove the toggle and have the 3-column display always shown.

ismailsunni:
Hi @nyalldawson and @nirvn 
Thanks for the feedback, I will remove it then :)

ismailsunni:
hi @nirvn @nyalldawson @roya0045 

Here is the updated UI now.

image

roya0045:
@ismailsunni Looks good, thanks for the nice feature!

ismailsunni:
@roya0045 and thanks for the input! :)

xtc0r:
@ismailsunni: Should the 3D measurement tool work for point clouds as well? I loaded a point cloud from a UAV scan but can't click on any point to start the measurement in the 3D view using the 3D measurement tool. It seems to work sometimes but I think then it is using the 2D map below.

PR #31460, opened 2019-08-28
Partially revert 596ffd06

nyalldawson:
Just to see if this is throwing off Travis...


nyalldawson:
Hmm -- so it IS this change which has killed Travis.

3nids:
yeah, I was reading more info on QT5_ADD_RESOURCES...

nyalldawson:
Maybe we could limit this inclusion to the single picture test it's required for?

3nids:
I think we should use autorcc
https://cmake.org/cmake/help/v3.15/prop_tgt/AUTORCC.html#prop_tgt:AUTORCC
will push to this branch

nirvn:
@3nids , you should stream your struggle on twitch; man vs machine :wink:

PR #31470, opened 2019-08-28
fix missing SIP_OUT

3nids:
This removes the param doc but it's an issue from sipify.pl
issue https://github.com/qgis/pyqgis/issues/35


nyalldawson (commented):
Isn't this an in/out parameter?

3nids:
indeed

PR #31475, opened 2019-08-28
Add a context menu to the QgsMessageLogViewer's tab bar...

borysiasty:
…so all unwanted tabs can be closed at once. Because it's not easy to debug stuff if your plugins tend to reopen 10+ tabs ;)

Zrzut_20190828_215349

## Checklist

### DIFF
--- a/src/gui/qgsmessagelogviewer.cpp
+++ b/src/gui/qgsmessagelogviewer.cpp
@@ -41,6 +41,48 @@ QgsMessageLogViewer::QgsMessageLogViewer( QWidget *parent, Qt::WindowFlags fl )
            this, static_cast<void ( QgsMessageLogViewer::* )( const QString &, const QString &, Qgis::MessageLevel )>( &QgsMessageLogViewer::logMessage ) );
 
   connect( tabWidget, &QTabWidget::tabCloseRequested, this, &QgsMessageLogViewer::closeTab );
+
+  mTabBarContextMenu = new QMenu( this );
+  tabWidget->tabBar()->setContextMenuPolicy( Qt::CustomContextMenu );
+  connect( tabWidget->tabBar(), &QWidget::customContextMenuRequested, this, &QgsMessageLogViewer::showContextMenuForTabBar );
+}
+
+void QgsMessageLogViewer::showContextMenuForTabBar( QPoint point )
+{
+  if ( point.isNull() )
+  {
+    return;
+  }
+
+  mTabBarContextMenu->clear();
+
+  int tabIndex = tabWidget->tabBar()->tabAt( point );
+
+  QAction *actionCloseTab = new QAction( tr( "Close Tab" ), mTabBarContextMenu );
+  connect( actionCloseTab, &QAction::triggered, this, [this, tabIndex]
+  {
+    closeTab( tabIndex );
+  }
+         );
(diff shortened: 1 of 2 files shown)


src/gui/qgsmessagelogviewer.cpp
```diff
@@ -41,6 +42,47 @@ QgsMessageLogViewer::QgsMessageLogViewer( QWidget *parent, Qt::WindowFlags fl )
...
+void QgsMessageLogViewer::showContextMenuForTabBar( QPoint point )
+{
+  if ( point.isNull() )
+  {
+    return;
+  }
+
+  int tabIndex = tabWidget->tabBar()->tabAt( point );
+  QMenu *menu = new QMenu( this );
+  QAction *mActionCloseTab = new QAction( tr( "Close tab" ), menu );
```
nyalldawson: "Close Tab"

src/gui/qgsmessagelogviewer.cpp
```diff
@@ -41,6 +42,47 @@ QgsMessageLogViewer::QgsMessageLogViewer( QWidget *parent, Qt::WindowFlags fl )
...
+  connect( mActionCloseTab, &QAction::triggered, this, [this, tabIndex]
+  {
+    closeTab( tabIndex );
+  }
+         );
+  menu->addAction( mActionCloseTab );
+
+  if ( tabWidget->tabBar()->count() > 1 )
+  {
+    QAction *mActionCloseOtherTabs = new QAction( tr( "Close other tabs" ), menu );
```
nyalldawson: "Close Other Tabs"

src/gui/qgsmessagelogviewer.cpp
```diff
@@ -41,6 +42,47 @@ QgsMessageLogViewer::QgsMessageLogViewer( QWidget *parent, Qt::WindowFlags fl )
...
+
+void QgsMessageLogViewer::showContextMenuForTabBar( QPoint point )
+{
+  if ( point.isNull() )
+  {
+    return;
+  }
+
+  int tabIndex = tabWidget->tabBar()->tabAt( point );
+  QMenu *menu = new QMenu( this );
```
nyalldawson: Menu is leaked here. Well, not really leaked (since it's cleaned up at log destruction), but it's kept around for the lifetime of QGIS.

I'd suggest making mMenu a member here, and creating it once in the widget constructor. Then replace this line with

    mMenu->clear();

nyalldawson (commented):
Great idea!

src/gui/qgsmessagelogviewer.cpp
borysiasty: @nyalldawson thanks! So also both actions should be members, just connected and disconnected to the lambda every time, shouldn't they? Now I see I mistakenly named them like mActionCloseTab, despite they are not members (yet).

nyalldawson: Actually I'd leave them as constructed on demand like you have here, and only make the menu a member. If you clear it just before showing it will delete the existing actions, so at most only the menu+ 2 actions will exist...

borysiasty: Great, now should be ok.

nyalldawson (approved):

PR #31478, opened 2019-08-29
Allow array values as a valid result for data defined offset or size properties

### LINKED ISSUE #31444: Labels position through formulas is uneffective
Dear all,

I just tried to use the label position formulas but it seems to be effectless. I never used this feature before. I am using the 3.8.2 QGIS version, and the problem was there with the 3.8.0 one, but I don't know whether it occured in a former version.

As I use a French version, I am not sure about the correct translation of the different fields. I'll use screenshots to be more explicit.

Here is my labelling option windows of my point layer :
labelling options
You can notice with the yellow background right the X,Y shifting fields that the formula is really activated, and the set up formula is : 
`array( -10 * cos(radians(180 -  "angle" )), 10 * sin(radians("angle")))`
where angle is a layer field holding a integer value (an angle in degrees).
The preview of the formula result shown is (for a 60° value of angle) : 
`[ 4.999999999999998, 8.660254037844386 ]`

But whatever the value is, the label doesn't move. It looks like the formula is not taken into account at all, only the value input in the X,Y fields.

Any help would be greatly appreciated.

Many thanks,

Remi.

nyalldawson:
Previously only string values of the format 'x,y' would be permitted. But I've seen MANY bug reports and questions about this, so also allow arrays of doubles as a valid result. E.g. "array(3,5)". In any case, it's
just nicer.

Fixes #31444

### DIFF
--- a/python/core/auto_generated/symbology/qgssymbollayerutils.sip.in
+++ b/python/core/auto_generated/symbology/qgssymbollayerutils.sip.in
@@ -94,6 +94,23 @@ Decodes a QSizeF from a string.
 .. seealso:: :py:func:`decodeSize`
 %End
 
+    static QPointF toPoint( const QVariant &value, bool *ok /Out/ = 0 );
+%Docstring
+Converts a ``value`` to a point.
+
+:param value: value to convert
+
+:return: - converted point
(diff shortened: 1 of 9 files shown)


nirvn (approved):
Nice tweak.

PR #31480, opened 2019-08-29
Allow expression geometry creation functions to use array inputs

nyalldawson:
Adds a new expression function "collect_geometries"

Collects a set of geometries into a multi-part geometry object.
Geometry parts can either be specified as seperate arguments to the
function, or (more flexibly), as an array of geometry parts.

This allows geometries to be generated using iterator based approaches,
such as transforming an array generated using generate_series, e.g:

  collect_geometries(
    array_foreach(
      generate_series( 0, 330, 30),
      project($geometry, .2, radians(@element))
    )
  )

Gives a nice radial effect of points surrounding the central feature point when used as a MultiPoint geometry generator:

image

Also, adds a make_line expression function variant which accepts an array of points. Allows creation of lines from variable numbers of points, and from sequences from aggregates/dynamically generated sequences:

image

### DIFF
--- a/resources/function_help/json/collect_geometries
+++ b/resources/function_help/json/collect_geometries
@@ -1,0 +1,19 @@
+{
+  "name": "collect_geometries",
+  "type": "function",
+  "description": "Collects a set of geometries into a multi-part geometry object.",
+  "variants": [
+  { "variant": "List of arguments variant",
+      "variant_description": "Geometry parts are specified as seperate arguments to the function.",
+      "variableLenArguments": true,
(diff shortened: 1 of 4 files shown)


nirvn (approved):

nirvn:
Merge away!

DelazJ:
> image

@nyalldawson How did you get the symbols to get rotated. I tried to reproduce it for https://github.com/qgis/QGIS-Documentation/pull/8492 but couldn't find the correct settings. Any tips? Thanks

PR #31500, opened 2019-08-29
Refactor graduated symbol renderer

3nids:
This is the implementation of https://github.com/qgis/QGIS-Enhancement-Proposals/issues/147

Some work remains to be done in the widget and reading/writing to XML.


src/core/symbology/classification/qgsclassificationmethod.cpp
```diff
@@ -0,0 +1,246 @@
...
+    maximum = *result.second;
+  }
+  else
+  {
+    minimum = vl->minimumValue( fieldIndex ).toDouble();
+    maximum = vl->maximumValue( fieldIndex ).toDouble();
+  }
+
+  // get the breaks
+  const QList<double> breaks = calculateBreaks( minimum, maximum, values, numberOfClasses );
```
nyalldawson: Just had a quick glance over -- one thing that strikes me immediately is that there's potentially a huge regression in speed for some methods here. E.g. the existing ones like equal interval, which only require a min and max value from a layer will now be significantly slower as they require a full layer iteration.

Could this be avoided, e.g. via some flags the subclass could return to say "i only need min and max" or "i need all values"?

3nids: yep, already implemented ;)
Look at the non-virtual public method in QgsClassificationMethod.
And every implementation can define if values are required when calling the parent constructor.

3nids: but I kept only one pure virtual private method to be implemented to make things simpler.

3nids:
Missing bits were completed and this is now tested.

The only changes for the user:
* an editable combobox with values from the ranges for the symmetry point (instead of spinbox)
* ranges values can now be edited and legend will be correctly recalculated (for Std Dev for instance)

3nids:
The PR is ready for review.

Last missing thing to get green light from Travis is a warning when building the API docs probably due to use of `Q_DECL_DEPRECATED` on an enum. See warning on the build [log](https://travis-ci.org/qgis/QGIS/jobs/579128956#L740) or:
image
Will try with a newer Doxygen thanks to using Bionic on Travis (instead of Xenial), if not I will just drop the deprecation.

Regarding the review, I have mainly one point I'd like opinions. It's in the graduated symbol renderer which keeps a pointer to the method. A new method can be set using a setter.
I have used a sharer pointer for the method. See [here](https://github.com/qgis/QGIS/pull/31500/files#diff-43779a76c26f526818cd7fd779c3bd14R438). Is this a wise choice (or a dummy one)?

python/CMakeLists.txt
```diff
@@ -118,6 +118,7 @@ INCLUDE_DIRECTORIES(
   ${CMAKE_SOURCE_DIR}/src/core/raster
   ${CMAKE_SOURCE_DIR}/src/core/scalebar
   ${CMAKE_SOURCE_DIR}/src/core/symbology
+  ${CMAKE_SOURCE_DIR}/src/core/symbology/classification
```
nyalldawson: We should move this out of symbology -- there's many use cases outside this realm

src/core/CMakeLists.txt
```diff
@@ -1317,6 +1335,7 @@ INCLUDE_DIRECTORIES(
   renderer
   scalebar
   symbology
+  symbology/classification
```
nyalldawson: Don't forget to add the new folder to the dox cmake template

src/core/symbology/classification/qgsclassificationequalinterval.cpp
```diff
@@ -0,0 +1,99 @@
...
+    breaks.reserve( nclasses );
+    double value = ( distBelowSymmetricValue < distAboveSymmetricValue ) ?  minimum : maximum - nclasses * step;
+
+    for ( int i = 0; i < nclasses; i++ )
+    {
+      value += step;
+      breaks << value;
+    }
+    breaks[nclasses - 1] = maximum;
+  }
```
nyalldawson: For review purposes -- was any of the actual logic changed in these?

src/core/symbology/classification/qgsclassificationjenks.cpp
```diff
@@ -0,0 +1,171 @@
...
+}
+
+
+QList<double> QgsClassificationJenks::calculateBreaks( double minimum, double maximum,
+    const QList<double> &values, int nclasses )
+{
+  // Jenks Optimal (Natural Breaks) algorithm
+  // Based on the Jenks algorithm from the 'classInt' package available for
+  // the R statistical prgramming language, and from Python code from here:
+  // http://danieljlewis.org/2010/06/07/jenks-natural-breaks-algorithm-in-python/
```
nyalldawson: note to self - we should resync this with classInt and double check we aren't missing any upstream fixes...

src/core/symbology/classification/qgsclassificationjenks.cpp
```diff
@@ -0,0 +1,171 @@
...
+  {
+    return values;
+  }
+
+  QVector<double> sample;
+
+  // if we have lots of values, we need to take a random sample
+  if ( values.size() > mMaximumSize )
+  {
+    // for now, sample at least maximumSize values or a 10% sample, whichever
```
nyalldawson: is it worth moving the support for max sample sizes outside of this, so that the feature request used to fetch the values has a setLimit on it?

src/core/symbology/classification/qgsclassificationmethod.cpp
```diff
@@ -0,0 +1,300 @@
...
+#include "qgsvectorlayerutils.h"
+#include "qgsvectorlayer.h"
+#include "qgsgraduatedsymbolrenderer.h"
+#include "qgsapplication.h"
+#include "qgsclassificationmethodregistry.h"
+
+const int QgsClassificationMethod::MAX_PRECISION = 15;
+const int QgsClassificationMethod::MIN_PRECISION = -6;
+
+static const QRegExp RE_TRAILING_ZEROES = QRegExp( "[.,]?0*$" );
```
nyalldawson: QRegExp -> QRegularExpression

nyalldawson: Also am wondering -- is this thread safe? There's a strong requirement to ensure that the classification methods ARE, so that they can be used in processing.

src/core/symbology/classification/qgsclassificationmethod.cpp
```diff
@@ -0,0 +1,300 @@
...
+QList<QgsClassificationRange> QgsClassificationMethod::classes( const QgsVectorLayer *vl, const QString &expression, int nclasses )
+{
+  if ( expression.isEmpty() )
+    return QList<QgsClassificationRange>();
+
+  if ( nclasses < 1 )
+    nclasses = 1;
+
+  QList<double> values;
+  double minimum;
```
nyalldawson: = std::numeric _limits...

src/core/symbology/classification/qgsclassificationmethod.cpp
```diff
@@ -0,0 +1,300 @@
...
+
+QList<QgsClassificationRange> QgsClassificationMethod::classes( const QList<double> &values, int nclasses )
+{
+  auto result = std::minmax_element( values.begin(), values.end() );
+  double minimum = *result.first;
+  double maximum = *result.second;
+
+  // get the breaks
+  QList<double> breaks = calculateBreaks( minimum, maximum, values, nclasses );
+  breaks.insert( 0, minimum );
```
nyalldawson: why is this done here, and not in the subclasses?

src/core/symbology/classification/qgsclassificationmethod.h
```diff
@@ -0,0 +1,290 @@
...
+#include "qgsclassificationquantile.h"
+#include "qgsclassificationstandarddeviation.h"
+% End
+#endif
+
+
+
+/**
+ * \ingroup core
+ * QgsClassificationRange contains the information about a classification range
```
nyalldawson: \since

src/core/symbology/classification/qgsclassificationmethod.h
```diff
@@ -0,0 +1,290 @@
...
+      sipType = 0;
+    SIP_END
+#endif
+
+  public:
+
+    //! Defines the class position
+    enum ClassPosition
+    {
+      LowerBound,
```
nyalldawson: dox

src/core/symbology/classification/qgsclassificationmethod.h
```diff
@@ -0,0 +1,290 @@
...
+      Inner,
+      UpperBound
+    };
+
+    /**
+      * Creates a classification method.
+      * \param valuesRequired if TRUE, this means that the method requires a set of values to determine the classes
```
nyalldawson: instead of the separate bool arguments here, I'd rather see a flags approach. It'll be more readable and more extensible.

src/core/symbology/classification/qgsclassificationmethod.h
```diff
@@ -0,0 +1,290 @@
...
+     * Returns a clone of the method.
+     * Implementation can take advantage of copyBase method which copies the parameters of the base class
+     */
+    virtual QgsClassificationMethod *clone() const = 0 SIP_FACTORY;
+
+    //! The readable and translate name of the method
+    virtual QString name() const = 0;
+
+    //! The id of the method
+    virtual QString id() const = 0; // as saved in the project, must be unique in registry
```
nyalldawson: should the comment here become part of the dox?

src/core/symbology/classification/qgsclassificationmethod.h
```diff
@@ -0,0 +1,290 @@
...
+
+    /**
+     * Returns the label for a range
+     */
+    virtual QString labelForRange( const double &lowerValue, const double &upperValue, ClassPosition position = Inner ) const;
+
+
+    //! Writes extra information about the method
+    virtual void saveExtra( QDomElement &element, const QgsReadWriteContext &context ) const {Q_UNUSED( element ); Q_UNUSED( context )}
```
nyalldawson: probably better to stick with the quasi-standard readXml/writeXml here

src/core/symbology/classification/qgsclassificationmethod.h
```diff
@@ -0,0 +1,290 @@
...
+     */
+    QList<QgsClassificationRange> classes( const QgsVectorLayer *vl, const QString &expression, int nclasses );
+
+    /**
+     * This will calculate the breaks for a list of values.
+     * The breaks do not contain the uppper and lower bounds (minimum and maximum values)
+     * \param values The list of values
+     * \param nclasses The number of classes to be returned
+     */
+    QList<QgsClassificationRange> classes( const QList<double> &values, int nclasses );
```
nyalldawson: I think there's a missing method here -- there's no way for a caller to get the numeric breaks alone (ie as a list of doubles) from the public api.

src/core/symbology/classification/qgsclassificationmethod.h
```diff
@@ -0,0 +1,290 @@
...
+    static QList<double> listToValues( const QList<QgsClassificationRange> classes );
+
+    /**
+     * This will calculate the breaks for a given layer to define the classes.
+     * The breaks do not contain the uppper and lower bounds (minimum and maximum values).
+     * \param vl The vector layer
+     * \param fieldName The name of the field on which the classes are calculated
+     * \param nclasses The number of classes to be returned
+     */
+    QList<QgsClassificationRange> classes( const QgsVectorLayer *vl, const QString &expression, int nclasses );
```
nyalldawson: vl -> "layer"

src/core/symbology/classification/qgsclassificationmethod.h
```diff
@@ -0,0 +1,290 @@
...
+    QString formatNumber( double value ) const;
+
+  private:
+
+    /**
+     * Calculate the breaks, should be reimplemented, values might be an empty list
+     * If the symmetric mode is available, the implementation is responsible of applying the symmetry
+     * The maximum value is expected to be added at the end of the list, but not the minimum
+     */
+    virtual QList<double> calculateBreaks( double minimum, double maximum,
```
nyalldawson: Do private virtuals work ok in sip?

src/core/symbology/classification/qgsclassificationmethod.h
```diff
@@ -0,0 +1,290 @@
...
+
+    //! This is called after calculating the breaks or restoring from XML, so it can rely on private variables
+    virtual QString valueToLabel( const double &value ) const {return formatNumber( value );}
+
+
+    //! Create a list of ranges from a list of classes
+    QList<QgsClassificationRange> breaksToClasses( const QList<double> &breaks ) const;
+
+    // implementation properties (set by initialization)
+    bool mValuesRequired; // if all values are required to calculate breaks
```
nyalldawson: initialize these

src/core/symbology/classification/qgsclassificationmethodregistry.h
```diff
@@ -0,0 +1,55 @@
...
+class CORE_EXPORT QgsClassificationMethodRegistry
+{
+  public:
+    QgsClassificationMethodRegistry();
+
+    //! Adds a method to the registry
+    void addMethod( QgsClassificationMethod *method SIP_TRANSFER );
+
+    //! Return a new instance of the method for the given id
+    QgsClassificationMethod *method( const QString &id );
```
nyalldawson: SIP_FACTORY

src/core/symbology/classification/qgsclassificationmethodregistry.h
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * \since QGIS 3.10
+ */
+class CORE_EXPORT QgsClassificationMethodRegistry
+{
+  public:
+    QgsClassificationMethodRegistry();
+
+    //! Adds a method to the registry
+    void addMethod( QgsClassificationMethod *method SIP_TRANSFER );
```
nyalldawson: Make bool, return false if duplicate id already exists

src/core/symbology/qgsgraduatedsymbolrenderer.cpp
```diff
@@ -337,317 +354,16 @@ bool QgsGraduatedSymbolRenderer::accept( QgsStyleEntityVisitorInterface *visitor
...
-
-static QStringList _breaksAsStrings( const QList<double> &breaks ) // get QStringList from QList<double> without maxi break (min is not in)
-{
-  QStringList breaksAsStrings;
-  for ( int i = 0; i < breaks.count() - 1; i++ )
-  {
-    breaksAsStrings << QString::number( breaks.at( i ), 'f', 2 );
-  }
-  return breaksAsStrings;
+  QgsClassificationEqualInterval *method = new QgsClassificationEqualInterval();
```
nyalldawson: No need to create on heap

src/core/symbology/qgsgraduatedsymbolrenderer.h
```diff
@@ -400,10 +435,7 @@ class CORE_EXPORT QgsGraduatedSymbolRenderer : public QgsFeatureRenderer
     //! \note not available in Python bindings
     static const char *graduatedMethodStr( GraduatedMethod method ) SIP_SKIP;
 
-    bool mUseSymmetricMode = false;
-    double mSymmetryPoint = 0.0;
-    QStringList mListForCboPrettyBreaks = QStringList();
-    bool mAstride = false;
+    std::shared_ptr<QgsClassificationMethod> mClassificationMethod;
```
nyalldawson: Why is this a shared pointer?

src/core/symbology/qgsgraduatedsymbolrenderer.h
```diff
@@ -15,16 +15,19 @@
 #ifndef QGSGRADUATEDSYMBOLRENDERER_H
 #define QGSGRADUATEDSYMBOLRENDERER_H
 
+#include <QRegExp>
```
nyalldawson: Not needed?

src/core/symbology/classification/qgsclassificationequalinterval.cpp
3nids: no.
The only thing I find weird is that breaks always include maximum but not minimum values. I kept this behavior.

src/core/symbology/classification/qgsclassificationjenks.cpp
3nids: That makes sense but a beyond the scope of this PR. I believe this can be easily added later on, even in classification base class.

src/core/symbology/qgsgraduatedsymbolrenderer.h
3nids: As you re-run classify on the widget, a new method is created and assigned the renderer. I am worried that someone keeps on using it afterwards.

nyalldawson: I think in that case they should be taking a clone of it themselves.

src/core/symbology/classification/qgsclassificationmethod.h
3nids: there's a static method classes to breaks for this.
otherwise, I'd need to duplicate all 3 classes methods

src/core/symbology/classification/qgsclassificationmethod.h
3nids: I already have create/save. I wanted to clearly state it's for additional parameters. Is this obvious?

src/core/symbology/classification/qgsclassificationmethod.cpp
3nids: it is set a few lines after.
What would be the reasoning to set it?

src/core/symbology/classification/qgsclassificationmethod.cpp
3nids: since the current way it works.
The breaks contains the maximum but not the minimum.
It would make sense to change this, but it might break existing code.

src/core/symbology/classification/qgsclassificationmethod.cpp
nyalldawson: any uninitialised bool/int/double makes me nervous -- it just looks wrong...

nyalldawson:
@3nids a minor regression here:

image
 
(duplicate "std dev" text)

PR #31508, opened 2019-08-30
correct acres abbreviation to "ac"

kgjenkins:
## Description

This is just a trivial change to correct the abbreviation for acres that appears in the "identify results" panel.  Acres is already an area unit, so there is no need for the superscript 2.


nyalldawson:
Thanks!

Issue #31311, opened 2019-08-20
Map tips changed behaviour


northwestview:
In version 3.8.1, the map tips dialog stays open when the pointer is moved off of the point, allowing any links in the dialog to be clicked on. In 3.8.2 as soon as the pointer is moved off of the point with the map tip, the map tip dialog disappears. If there is a setting to set the length of time the map tip stays open, it should be noted somewhere in the documentation.

gioman:
>  If there is a setting to set the length of time the map tip stays open, it should be noted somewhere in the documentation

there is: configuration > options > canvas and legend > map tips

https://docs.qgis.org/testing/en/docs/user_manual/introduction/qgis_configuration.html?highlight=map%20tips#canvas-and-legend-settings

northwestview:
Thank you for the response.

That setting only seems to affect how long it takes for the map tip to
open, not how long it stays open. As I mentioned, in 3.8.2 as soon as I
move the pointer off of the point that has the map tip, the map tip
disappears. In 3.8.1 the map tip stayed open when moving the pointer off
the point with the map tip. Having the map tip disappear means that any
clickable links in the map tip cannot be clicked on (to open an image for
instance).

There is a work-around: by right-clicking on the point after the map tip
opens. This opens a dialog for the selection of which layer to show the
identify feature dialog for, then it is possible to left click in the map
tip dialog and the map tip dialog stays open so that links can be clicked.
Much different than the way the map tip functioned in 3.8.1.

gioman:
I confirm it, is no longer possible to have a maptip stay on open when moving a little the mouse, so is not possible anymore to click on (i.e.) links.

gioman:
The change appeared in 3.8.2 (works ok on 3.8.1) and affects also 3.4.11 (works ok on 3.4.10).

nirvn:
@SrNetoChan , this regression was most likely caused by your series of commits (commit 8b052fd and 6a1b572) during the 3.8.2 dev cycle, please review.

SrNetoChan:
@nirvn those two commits were later changed by @nyalldawson https://github.com/qgis/QGIS/blame/master/src/gui/qgsmaptip.cpp#L149

```
  int cursorOffset = 0;
  // attempt to shift the tip away from the cursor.
  if ( QgsApplication::instance() )
  {
    // The following calculations are taken
    // from QgsApplication::getThemeCursor, and are used to calculate the correct cursor size
[...]

PR #31515, opened 2019-09-01
Revert "Moves MapTip away from the mouse cursor (revision)"

SrNetoChan:
Reverts qgis/QGIS#30976

The improvement in the Maptip causes problems #31311 when the user tries to click hyperlinks in the Map tip. The workaround for the map tips being covered by the cursor is far easier than the needed workaround to be able to click the hyperlinks.


SrNetoChan:
@gioman @nirvn @nyalldawson Can any of you add the backport labels?

nyalldawson:
I'll leave this open until just before the next point release, just in case we can find a solution which fixes both issues in the meantime

PR #31516, opened 2019-09-01
Show message when stack trace can't be generated from  missing symbols

NathanW2:
Updates to the crash handler to not show the stack trace when the symbols can't be loaded. When symbols aren't loaded the stack trace is pretty much useless and misleading meaning people think it's caused by something it really isn't.

Other tweaks in this PR include:

- Fixing the wording to move from "Crashed" to "unexpectedly ended" as it's less aggressive 
- Fixed spacing in the message wording
- Added some more help text

image

In future, I would like to make this dialog also fetch the symbols in the background if they are not found but that is out of scope for this PR.

Should also hopefully be good for a backport into 3.4

### DIFF
--- a/src/crashhandler/qgscrashdialog.cpp
+++ b/src/crashhandler/qgscrashdialog.cpp
@@ -26,8 +26,8 @@ QgsCrashDialog::QgsCrashDialog( QWidget *parent )
   setupUi( this );
   setWindowTitle( tr( "Oh Uh!" ) );
 
-  mCrashHeaderMessage->setText( tr( ":( QGIS Crashed" ) );
-  mCrashMessage->setText( tr( "Sorry. It looks something unexpected happened that we didn't handle and QGIS crashed." ) );
+  mCrashHeaderMessage->setText( tr( "QGIS unexpectedly ended" ) );
+  mCrashMessage->setText( tr( "Sorry :( It looks something unexpected happened that we didn't handle and QGIS ended unexpectedly." ) );
   connect( mReloadQGISButton, &QPushButton::clicked, this, &QgsCrashDialog::reloadQGIS );
   connect( mCopyReportButton, &QPushButton::clicked, this, &QgsCrashDialog::createBugReport );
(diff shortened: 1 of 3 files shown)


nirvn:
Nice, nice! So happy the stack trace is not showing instead of throwing wrong calls.

Only comment: can you make sure all the sentences end with a dot, and that the dot isn't part of any hyperlink? 

Glad we have you back coding! ;)

nirvn:
Oh, and moving the ascii emoji away from the headline? Yay!

NathanW2:
> Nice, nice! So happy the stack trace is not showing instead of throwing wrong calls.
> 
> Only comment: can you make sure all the sentences end with a dot, and that the dot isn't part of any hyperlink?
> 
> Glad we have you back coding! ;)

Good to be back :)

Added the dots.

nirvn:
One missing, the sentence that starts with "Include as much info as possible..." needs a dot. Also, maybe use information instead of info there, looks a bit cleaner.

nirvn:
Looking good; I think we need to backport this to 3.4, to improve bug reporting.

NathanW2:
All good to merge @nirvn?

nirvn (approved):
Yep, go ahead.

Gustry:
Nice.

Side question @NathanW2 is-it possible to make the crash ID clickable too (opening the web browser)? And maybe we can ask people first to check if there is another ticket open before to create a new one?

NathanW2:
There is a PR already for that I am pretty sure. It should be ok to merge
from last time I looked

Gustry:
Yes, it was me ;-) And it has been merged.
It was for the crash report when it's on github. But I don't think the link will open on the QGIS application? I can't try. Maybe if we should ask our users to check first if there is a ticket in this dialog.

Sorry, I didn't try what I did as I don't have windows.

NathanW2:
Ahh sorry was on my phone so didn't even look. Yeah I can have a look and
see what I can do here

PR #31517, opened 2019-09-01
[Server] Extend the WMS FILTER whitelist with IS, NOT, NULL

rldhont:
The QGIS Server Filter parameter accept OGC Filter and an SQL where clause.

In the case of the SQL where clause, to avoid injection and garanty some safety, QGIS Server uses a whitelist of allowed characters and keywords.
The whitelist was AND,OR,IN,=,<,>=,>,>=,!=,',',(,),DMETAPHONE,SOUNDEX.

We need to be able to handle NULL values.

## Checklist

### DIFF
--- a/tests/src/python/test_qgsserver_wms_getmap.py
+++ b/tests/src/python/test_qgsserver_wms_getmap.py
@@ -903,10 +903,29 @@ class TestQgsServerWMSGetMap(QgsServerTestBase):
             "FILTER": "Country,Country_Diagrams: \"name\" IN ( 'africa , 'eurasia' );Hello: \"color\" IN ( 'magenta' , 'cerese' )"
         }.items())])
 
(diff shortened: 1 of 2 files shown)


tests/src/python/test_qgsserver_wms_getmap.py
```diff
@@ -908,6 +908,25 @@ def test_wms_getmap_filter(self):
...
+            "SERVICE": "WMS",
+            "VERSION": "1.1.1",
+            "REQUEST": "GetMap",
+            "LAYERS": "Country,Hello",
+            "STYLES": "",
+            "FORMAT": "image/png",
+            "BBOX": "-16817707,-4710778,5696513,14587125",
+            "HEIGHT": "500",
+            "WIDTH": "500",
+            "CRS": "EPSG:3857", ,
```
rldhont:
```suggestion
            "CRS": "EPSG:3857",
```

pblottiere:
If memory serves me right, I think that this PR fixes an issue raised by @elemoine some weeks ago, but I didn't succeed in retrieving the dedicated ticket...

elemoine:
> If memory serves me right, I think that this PR fixes an issue raised by @elemoine some weeks ago, but I didn't succeed in retrieving the dedicated ticket...

Indeed.

And the change looks good to me.

rldhont:
@mhugent, @elpaso any objections to merge?

Would someone like it to backport?

elpaso:
+1 for merge and backport

PR #31523, opened 2019-09-02
[Backport] Show message when stack trace can't be generated from  missing symbols

nirvn:
Backporting @NathanW2 's crash dialog stack trace fix.


nyalldawson (approved):

NathanW2:
Thanks.

PR #31524, opened 2019-09-02
New class QgsBookmarkManager

nyalldawson:
Attached to QgsProject, this provides a stable, supported method of managing project bookmarks (vs the old undocumented, not stable approach of directly manipulating project keys)

This is a minimal implementation which just handles the basics of the existing project bookmark functionality. There's no new features (unless you count stable API + unit tests as features!).

Ideally, there's a lot more we should do here:
- ~~port the existing "global" bookmarks to a QgsBookmarkManager instance attached to QgsApplication~~ DONE!
- ~~move the models out of app, ensure they correctly use model/view architecture and add tests for them~~ DONE!
- use these models to expose bookmark functionality elsewhere in qgis, e.g. in layout designer map properties
- clean up the delegate related UX, which is rather messy at the moment

But for now, this "just" gives us nicer API for scripts and plugins to modify and interact with project bookmarks.

Sponsored by SLYR - the immediate use case is translating MXD project bookmarks to QGIS project bookmarks.


nirvn:
Bookmarks getting some love? Great! I often wished we could upgrade the usefulness of bookmarks by connecting it to various bit (for e.g., setting a layout map item extent by selecting a bookmark; adding setting of extent in our extent widget via saved bookmark, etc.).

nyalldawson:
@nirvn yes, me too. I'd say this is a preliminary prerequisite for that kind of thing....

nirvn:
@nyalldawson , good.

roya0045:
Neat, could this be used to export an 'atlas' from the bookmarks? (Maybe this is already doable, I always went for an atlas layer directly).

nyalldawson:
This is ready for review

nirvn:
You've just upgraded bookmarks to become first class citizens, nice. Will review later this morning.

src/core/qgsbookmarkmanager.cpp
```diff
@@ -0,0 +1,510 @@
...
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+#include "qgsbookmarkmanager.h"
+#include "qgsproject.h"
+#include "qgssettings.h"
+#include "qgssqliteutils.h"
```
nirvn: I think this can go away.

src/core/qgsbookmarkmanager.cpp
```diff
@@ -0,0 +1,510 @@
...
+#include "qgsapplication.h"
+#include <QUuid>
+#include <sqlite3.h>
```
nirvn: Same, this likely can go away.

src/core/qgsbookmarkmanager.cpp
```diff
@@ -0,0 +1,510 @@
...
+      const double maxX = mProject->readDoubleEntry( QStringLiteral( "Bookmarks" ), QStringLiteral( "/Row-%1/MaxX" ).arg( i ) );
+      const double maxY = mProject->readDoubleEntry( QStringLiteral( "Bookmarks" ), QStringLiteral( "/Row-%1/MaxY" ).arg( i ) );
+      const long srid = mProject->readNumEntry( QStringLiteral( "Bookmarks" ), QStringLiteral( "/Row-%1/SRID" ).arg( i ) );
+      QgsBookmark b;
+      b.setId( QStringLiteral( "bookmark_%1" ).arg( i ) );
+      b.setName( mProject->readEntry( QStringLiteral( "Bookmarks" ), QStringLiteral( "/Row-%1/Name" ).arg( i ) ) );
+      b.setExtent( QgsReferencedRectangle( QgsRectangle( minX, minY, maxX, maxY ), QgsCoordinateReferenceSystem::fromSrsId( srid ) ) );
+
+      bool added = false;
+      addBookmark( b, &added );
```
nirvn: Is that going to trigger a bunch of signals when reading an XML containing say 50-100 bookmarks? Is this something we might not want to have happening? 


I.e., if I'm listening for project bookmark manager signals of added bookmarks, loading a new project with a bunch of bookmarks could be problematic. We might be better adding a different signal here, like bookmarksRead.

src/core/qgsbookmarkmanager.cpp
```diff
@@ -0,0 +1,510 @@
...
+
+  QFile f( path );
+  if ( !f.open( QFile::WriteOnly | QIODevice::Truncate ) )
+  {
+    f.close();
+    return false;
+  }
+
+  QTextStream out( &f );
+  out.setCodec( "UTF - 8" );
```
nirvn:
```suggestion
  out.setCodec( "UTF-8" );
```

src/core/qgsbookmarkmanager.cpp
```diff
@@ -0,0 +1,510 @@
...
+      f.close();
+      return;
+    }
+
+    QDomDocument doc;
+    QDomElement elem = writeXml( doc );
+    doc.appendChild( elem );
+
+    QTextStream out( &f );
+    out.setCodec( "UTF - 8" );
```
nirvn:
```suggestion
    out.setCodec( "UTF-8" );
```

src/core/qgsbookmarkmanager.h
```diff
@@ -0,0 +1,295 @@
...
+
+/**
+ * \ingroup core
+ * \class QgsBookmarkManager
+ *
+ * \brief Manages storage of a set of bookmarks.
+ *
+ * QgsBookmarkManager handles the storage, serializing and deserializing
+ * of geographic bookmarks. Usually this class is not constructed directly, but
+ * rather accessed through a QgsProject via QgsProject::bookmarkManager().
```
nirvn: access through a QgsProject... or QgsApplication via QgsApplication::bookmarkManager().

tests/src/python/test_qgsbookmarkmanager.py
```diff
@@ -0,0 +1,543 @@
...
+        self.assertEqual(p.bookmarkManager().bookmarkById('bookmark_1').extent().crs().authid(), 'EPSG:4283')
+        self.assertEqual(p.bookmarkManager().bookmarkById('bookmark_1').extent().toString(1), '149.0,-21.6 : 149.4,-21.1')
+
+        self.assertEqual(p.bookmarkManager().bookmarkById('bookmark_2').name(), 'b3')
+        self.assertEqual(p.bookmarkManager().bookmarkById('bookmark_2').extent().crs().authid(), 'EPSG:28355')
+        self.assertEqual(p.bookmarkManager().bookmarkById('bookmark_2').extent().toString(1), '807985.7,7450916.9 : 876080.0,7564407.4')
+
+    def testFileStorage(self):
+        """
+        Test QSettings bound manager
```
nirvn: Test XML file bound manager?

nirvn (commented):
Great work.

src/core/qgsbookmarkmanager.cpp
nirvn: Blah ignore, I saw it was needed when importing old sqlite-based bookmarks. I don't understand why github won't let me delete comments in draft mode.

src/core/qgsbookmarkmanager.cpp
nirvn: Diddo.

src/app/qgsbookmarks.cpp
```diff
@@ -239,32 +168,32 @@ void QgsBookmarks::zoomToBookmark()
...
-      QMessageBox::warning( this, tr( "Empty Extent" ), tr( "Reprojected extent is empty." ) );
+      QgisApp::instance()->messageBar()->pushWarning( tr( "Zoom to Bookmark" ), tr( "Bookmark extent is empty" ) );
       return;
     }
   }
 
   // set the extent to the bookmark and refresh
-  QgisApp::instance()->setExtent( rect );
+  QgisApp::instance()->mapCanvas()->setExtent( canvasExtent );
   QgisApp::instance()->mapCanvas()->refresh();
```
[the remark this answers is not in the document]
nyalldawson: seems to be (and is done elsewhere). I suspect it's this way to avoid the expense of redrawing intermediate steps when e.g. a pan/zoom is done consecutively.

src/core/qgsbookmarkmanager.cpp
nyalldawson: We could add a second signal, but we need the modular signal for the model to correctly update.

nirvn: @nyalldawson , hmm, let me think here. I guess one way around being flooded with signals would be (for e.g. the browser panel) to disconnect the bookmarkAdded signal while a project is loading?

nirvn:
Merge away! 🎉

PR #31530, opened 2019-09-02
Added batch_number variable in algorithm scope

luipir:
Added batch_number variable in algorithm scope to allow use it in the expression editor in Processing batch context.

In some condition, auto generation of filename with automatic numbering is not the expected one and a more complex filename format is necessary

Not sure if it's a bugfix or new feature

## Checklist

### DIFF
--- a/python/plugins/processing/gui/BatchPanel.py
+++ b/python/plugins/processing/gui/BatchPanel.py
@@ -218,7 +218,7 @@ class BatchPanelFillWidget(QToolButton):
             alg_scope.setVariable(k, v, True)
 
         # add batchCount in the alg scope to be used in the expressions. 0 is only an example value
-        alg_scope.setVariable('row_number', 0, True)
+        alg_scope.setVariable('row_number', 0, False)
 
         expression_context.appendScope(alg_scope)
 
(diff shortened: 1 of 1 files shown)


python/plugins/processing/gui/BatchPanel.py
```diff
@@ -217,6 +217,9 @@ def populateByExpression(self, adding=False):
         for k, v in params.items():
             alg_scope.setVariable(k, v, True)
 
+        # add batchCount in the alg scope to be used in the expressions. 0 is only an example value
+        alg_scope.setVariable('batch_number', 0, True)
+
```
nyalldawson: I'd go with row_number here to match other parts of qgis

luipir:
wait for @nyalldawson complete review to merge or modify... delay due to different time zone

python/plugins/processing/gui/BatchPanel.py
```diff
@@ -217,6 +217,9 @@ def populateByExpression(self, adding=False):
         for k, v in params.items():
             alg_scope.setVariable(k, v, True)
 
+        # add batchCount in the alg scope to be used in the expressions. 0 is only an example value
+        alg_scope.setVariable('row_number', 0, True)
```
nyalldawson: this should be

    alg_scope.setVariable('row_number', 0, False)

because it's not a static variable -- it's changing with each evaluation

luipir: right! I missed the real meaning of static... thinking more to an expression evaluation

nyalldawson: it's used when expressions are prepared -- if a variable is marked as static, it's replaced in the expression internals by a static value to optimise multiple evaluations.

PR #31533, opened 2019-09-02
[FEATURE] add logarithmic scale method for graduated renderer

3nids:
image

This implements a new classification method which can be used in graduated method renderer. It creates a logarithmic scale using the desired number of classes. This works correctly for positive values only. 

This implementation has not been copied from anywhere, it has been created on demand, but the logic looks quite straightforward to me.

This PR also brings icons to the methods, credits to @nirvn for the design

This feature has been sponsored by [gem.foundation](https://thegemfoundation.com/)


nyalldawson:
Looks ok to me, but I'd love to know where this actual logic is derived from. Was it ported from elsewhere?

nirvn:
I'm also curious about this classification, and I assume the documentation team will love some more details too :)

3nids:
@nirvn @nyalldawson I updated the main PR comment

src/core/classification/qgsclassificationlogarithmic.cpp
```diff
@@ -0,0 +1,97 @@
...
+  // create the value
+  for ( int i = 0; i < breaks.count(); i++ )
+    breaks[i] = std::pow( 10, breaks.at( i ) );
+
+  return breaks;
+}
+
+QString QgsClassificationLogarithmic::valueToLabel( double value ) const
+{
+  QString label = QString( QStringLiteral( "10^%1" ) ).arg( std::log10( value ) );
```
nyalldawson: This would look nicer as 10E03 etc, wouldn't it?

src/core/classification/qgsclassificationlogarithmic.cpp
```diff
@@ -0,0 +1,97 @@
...
+      lowerLabel = formatNumber( lowerValue ); // avoid to have float exponent for the minimum value
+      break;
+    case Inner:
+    case UpperBound:
+      lowerLabel = valueToLabel( lowerValue );
+      break;
+  }
+
+  QString label( labelFormat() );
+  label.replace( QLatin1String( "%1" ), lowerLabel ).replace( QLatin1String( "%2" ), upperLabel );
```
nyalldawson: would 

    return labelFormat().arg( lowerLabel, upperLabel );

work instead? Tooke me a few reads to work out what the existing code is doing

src/core/classification/qgsclassificationlogarithmic.cpp
3nids: not sure, this notations doesn't necessarily speak to all.
But if you're sure, I have no objection.

src/core/classification/qgsclassificationlogarithmic.cpp
3nids: yep, works

3nids:
This feature has been sponsored by [gem.foundation](https://thegemfoundation.com/)

PR #31540, opened 2019-09-03
fix standard deviation label

3nids:


nyalldawson (commented):
Might be time for some unit tests of these classes ;)

3nids:
> Might be time for some unit tests of these classes ;)

I'll try again at the next grant proposal round ;)

PR #31547, opened 2019-09-04
Allow setting layout map item extent to an existing bookmark

### LINKED ISSUE #20279: Spatial bookmarks and the print composer
Original Redmine Issue: [12083](https://issues.qgis.org/issues/12083)

---
It will be good to have an option in the print composer for the extent to access spatial bookmarks.
For example, atlas can then access all the spatial bookmarks for coverages.

nyalldawson:
This new button in the map item properties toolbar shows a menu which allows directly setting a map item to the extent of that bookmark.

Fixes #20279
bookmarks


nirvn:
@nyalldawson , code looks good, merge away.

PR #31561, opened 2019-09-05
More bookmark browser functionality

nyalldawson:
Allows:
- drag and drop of bookmarks to canvases (useful in multi-canvas projects!)
- bookmarks can be copied by drag and drop in browser
- bookmarks can be renamed or deleted in browser

Peek 2019-09-05 11-07


nyalldawson:
@elpaso this has revealed that browser based geopackage table renaming is broken following the provider connection work

nirvn (approved):
Good stuff.

PR #31563, opened 2019-09-05
"Reselect Features" action in Edit -> Select menu

nyalldawson:
Allows restoration of a layer's selection following a selection clear operation. Handy for anyone who has ever painstakingly built a custom selection only to accidently click and clear this selection...

Peek 2019-09-05 13-53

### DIFF
--- a/python/core/auto_generated/qgsvectorlayer.sip.in
+++ b/python/core/auto_generated/qgsvectorlayer.sip.in
@@ -2360,6 +2360,21 @@ Deselects features by their ID
 Clear selection
 
 .. seealso:: :py:func:`selectByIds`
+
+.. seealso:: :py:func:`reselect`
+%End
+
+    void reselect();
+%Docstring
+Reselects the previous set of selected features. This is only applicable
+after a prior call to removeSelection().
+
+Any other modifications to the selection following a call to removeSelection() clears
+memory of the previous selection and consequently calling reselect() has no impact.
+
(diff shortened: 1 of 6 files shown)


nirvn (approved):
Code looks good. I'm wondering whether it'd be nice for the deselect toolbar button would actually be a deselect if selected <-> reselect if no selection.

nyalldawson:
>  I'm wondering whether it'd be nice for the deselect toolbar button would actually be a deselect if selected <-> reselect if no selection.

I don't think that works -- deselect is a global action which applies to multiple layers, but reselect is per-layer...

nirvn:
@nyalldawson , riight. Merge away then :)

j-lippmann:
Nice! Additional: deselect all Features of the Current Layer would  be cool😀 at the Moment its only possible in the attributtable..

j-lippmann:
hi @nyalldawson,
 thanks for this little nice feature! one further idea on this "selection" theme:
 https://github.com/qgis/QGIS/issues/31598#issue-490615294

gioman:
Agree, very nice and much needed! I also have an idea on this theme: what about going back and forward in the selection history, as much as the editing history but not necessarily with the panel editing comes with (just two shortcuts or buttons).

nyalldawson:
@giomann that's a good idea! @nathanw2 has also explored the concept of stored named "selection sets" before. I'd suggest these two ideas could be combined with a nice API in core, and exposed via a python plugin...

PR #31567, opened 2019-09-05
New algorithm "Split features by character"

nyalldawson:
This algorithm splits features into multiple output features by
splitting a field's value with a specified character.

For instance, if a layer contains features with multiple comma
separated values contained in a single field, this algorithm can
be used to split these values up across multiple output features.

Geometries and other attributes remain unchanged in the output.

Optionally, the separator string can be a regular expression for
added flexibility.

Designed for use in models which need to process input files
with multiple concatenated values in a single attribute, e.g.
geocoding a table with "address1,address2,address3" format strings


DelazJ:
Nyall, does it mean that my feature with "address1,address2,address3" would become 3 features with same original geometry and other attributes, and only the address field would have the different values?

nyalldawson:
@Delazj exactly!

nirvn (approved):
All good.

nyalldawson:
Thanks @nirvn!

PR #31569, opened 2019-09-05
More browser bookmark tweaks

nyalldawson:
Improves the UX of handling bookmarks in browser, by avoiding rebuilding the bookmark tree each time bookmarks are changed. Also adds the ability to delete multiple bookmarks at once.

### DIFF
--- a/src/app/qgsappbrowserproviders.cpp
+++ b/src/app/qgsappbrowserproviders.cpp
@@ -645,7 +645,6 @@ QgsBookmarkManagerItem::QgsBookmarkManagerItem( QgsDataItem *parent, const QStri
         addChildItem( new QgsBookmarkGroupItem( this, newDetails.group(), mManager ), true );
       }
     }
-
   } );
   connect( mManager, &QgsBookmarkManager::bookmarkChanged, this, [ = ]( const QString & id )
   {
@@ -676,6 +675,11 @@ QgsBookmarkManagerItem::QgsBookmarkManagerItem( QgsDataItem *parent, const QStri
             {
               newGroup->addBookmark( newDetails );
             }
+            else
+            {
+              // need to create a new group for this (will automatically add the new bookmark)
+              addChildItem( new QgsBookmarkGroupItem( this, newDetails.group(), mManager ), true );
+            }
           }
           break;
         }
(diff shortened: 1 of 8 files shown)


src/app/qgsappbrowserproviders.cpp
```diff
@@ -627,38 +629,173 @@ QgsBookmarkManagerItem::QgsBookmarkManagerItem( QgsDataItem *parent, const QStri
...
+          // ok, found old group, now compare
+          if ( bookmarkItem->bookmark().group() == newDetails.group() )
+          {
+            // good, not the group. Just update the existing bookmark then.
+            bookmarkItem->setBookmark( newDetails );
+          }
+          else
+          {
+            // group has changed!
+            // first remove from existing group
```
nirvn: I think we need some additional code to remove the departed group if that group didn't have any other bookmarks in it.

src/app/qgsappbrowserproviders.cpp
```diff
@@ -627,38 +629,173 @@ QgsBookmarkManagerItem::QgsBookmarkManagerItem( QgsDataItem *parent, const QStri
...
+      }
+    }
+  } );
+  connect( mManager, &QgsBookmarkManager::bookmarkAboutToBeRemoved, this, [ = ]( const QString & id )
+  {
+    QgsBookmark b = mManager->bookmarkById( id );
+    if ( !b.group().isEmpty() )
+    {
+      if ( QgsBookmarkGroupItem *group = groupItem( b.group() ) )
+        group->removeBookmarkChildById( id );
```
nirvn: Similarly, I think we need to remove the group if that group is empty.

src/app/qgsappbrowserproviders.cpp
```diff
@@ -795,6 +971,43 @@ void QgsBookmarksItemGuiProvider::populateContextMenu( QgsDataItem *item, QMenu
       QgisApp::instance()->showBookmarks( true );
     } );
     menu->addAction( showBookmarksPanel );
+    menu->addSeparator();
+    QAction *importBookmarks = new QAction( tr( "Import Bookmarks…" ), menu );
```
nirvn: +Spatial (for all import, export, and delete items, won't comment on the remaining ones :) )

nirvn (approved):
Yay!

PR #31587, opened 2019-09-06
More browser bookmark tweaks...

nyalldawson:
- bookmark groups can be renamed
- bookmark groups can be exported


nirvn (approved):
Merge away, all good.

PR #31592, opened 2019-09-06
Fix QGIS goes super-slow in certain circumstances

nyalldawson:
A bunch of optimisations related to a situation encountered today where qgis would grind to a halt when certain projects were opened.

Underlying cause was the expensive QFileInfo::canonicalFilePath() call in QgsProject::homePath() when projects were stored on a network drive. This function is called many many times, especially when project expression context scopes are created. And these are created a lot, as each call to generating a symbol preview pixmap creates a new expression context...

Accordingly, i've implemented some related optimisations here:
1. cache QgsProject::homePath, to avoid the underlying expensive call
2. move responsibility of creating project expression scopes to QgsProject, and cache these. Creating the scope from scratch is much more costly than cloning an existing one
3. use a shared QgsStyleModel for the default application style, so that there's a shared icon cache which can be used across the application (avoids duplicate generation of symbol preview icons whenever a new symbol list is created)

### DIFF
--- a/src/core/qgsproject.cpp
+++ b/src/core/qgsproject.cpp
@@ -486,6 +486,8 @@ void QgsProject::setPresetHomePath( const QString &path )
     return;
 
   mHomePath = path;
+  mCachedHomePath.clear();
+
   emit homePathChanged();
 
   setDirty( true );
@@ -570,6 +572,8 @@ void QgsProject::setFileName( const QString &name )
   QString oldHomePath = homePath();
 
   mFile.setFileName( name );
+  mCachedHomePath.clear();
+
   emit fileNameChanged();
 
   QString newHomePath = homePath();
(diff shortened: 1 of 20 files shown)


src/core/qgsproject.cpp
```diff
@@ -1587,6 +1607,67 @@ QgsExpressionContext QgsProject::createExpressionContext() const
   return context;
 }
 
+QgsExpressionContextScope *QgsProject::createExpressionContextScope() const
+{
+  // MUCH cheaper to clone then build
```
nirvn: then -> than

src/core/symbology/qgssymbollayerutils.h
```diff
@@ -223,11 +223,13 @@ class CORE_EXPORT QgsSymbolLayerUtils
      * \param padding space between icon edge and symbol
      * \param customContext render context to use when rendering symbol
      * \param selected set to TRUE to render the symbol in a selected state
+     * \param expressionContext optional custom expression context
```
nirvn: Note parameter added in 3.10

nirvn (approved):

PR #31593, opened 2019-09-06
[3D] Add details to error dialog

DelazJ:
better wording is welcome...


src/app/qgisapp.cpp
```diff
@@ -11505,13 +11505,13 @@ void QgisApp::new3DMapCanvas()
   // we can't go any further - invalid extent would break everything
   if ( fullExtent.isEmpty() || !fullExtent.isFinite() )
   {
-    QMessageBox::warning( this, tr( "Error" ), tr( "Project extent is not valid." ) );
+    QMessageBox::warning( this, tr( "New 3D Map View" ), tr( "Project extent is not valid. Please switch on the visibility of a layer." ) );
```
roya0045: Do you mean something like "Please add or activate a layer to render" ?

PR #31597, opened 2019-09-07
[needs-docs] Replace "Label with" with "Value"

### LINKED ISSUE #30159: Label panel wording for combo/expression box.
We currently have "Label with" as the start label for the combo for the drop down.  The wording here feels a bit off to me (although I think I wrote it originally) 

image

We also don't use wording like this for symbols so might be best to pick a standard name and run with that for both panels

image

DelazJ:
and harmonize with symbology renderers (fixes #30159)

image

## Checklist

### DIFF
--- a/src/ui/qgscategorizedsymbolrendererwidget.ui
+++ b/src/ui/qgscategorizedsymbolrendererwidget.ui
@@ -20,7 +20,7 @@
       </sizepolicy>
      </property>
      <property name="text">
-      <string>Column</string>
+      <string>Value</string>
      </property>
     </widget>
    </item>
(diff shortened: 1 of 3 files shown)


nirvn:
-1, this label is inaccurate/misleading as the associated output can be an expression.

You're trying to fix the wrong end of the problem ;) IMHO we need to harmonize this away from 'label with' and 'column' to 'Value' or 'Column/Value'.

The reason why the renderers say use column is that it originally only supported columns. A few centuries ago ;-P

m-kuhn:
In addition to what @nirvn said, if the new term still contains "Column" it should be replaced with "Field Name" for consistency.

nyalldawson:
+1 for "Value"

m-kuhn (approved):

PR #31602, opened 2019-09-07
fix gdal translate when translating multi band rasters and changing the datatype

### LINKED ISSUE #31544: Translate (Convert Format) produces a blank image
When I trying to convert a Landsat-8 imagery from UInt16 (Unsigned Integer 16 Bit) to Byte (Unsigned Integer 8 Bit), the algorithm produces a blank image.

This video shows the issue: 
https://drive.google.com/file/d/1MnEaYDVF4b4363RTN05ZLsNb4UyHj-aT/view?usp=sharing

**Another question:** I was unable to include the parameter **-scale** to stretch data correcty as I do via GDAL command prompt

QGIS version 3.4.11 64 Bit, Operational System Windows 10 Pro.

gioman:
fixes #31544

### DIFF
--- a/python/plugins/processing/algs/gdal/translate.py
+++ b/python/plugins/processing/algs/gdal/translate.py
@@ -47,6 +47,7 @@ class translate(GdalAlgorithm):
     TARGET_CRS = 'TARGET_CRS'
     NODATA = 'NODATA'
     COPY_SUBDATASETS = 'COPY_SUBDATASETS'
+    SCALE = 'SCALE'
     OPTIONS = 'OPTIONS'
     EXTRA = 'EXTRA'
     DATA_TYPE = 'DATA_TYPE'
@@ -71,6 +72,9 @@ class translate(GdalAlgorithm):
                                                        optional=True))
         self.addParameter(QgsProcessingParameterBoolean(self.COPY_SUBDATASETS,
                                                         self.tr('Copy all subdatasets of this file to individual output files'),
+                                                        defaultValue=False))
+        self.addParameter(QgsProcessingParameterBoolean(self.SCALE,
+                                                        self.tr('Rescale the input pixels values'),
                                                         defaultValue=False))
 
         options_param = QgsProcessingParameterString(self.OPTIONS,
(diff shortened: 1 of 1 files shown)


python/plugins/processing/algs/gdal/translate.py
```diff
@@ -72,6 +73,9 @@ def initAlgorithm(self, config=None):
         self.addParameter(QgsProcessingParameterBoolean(self.COPY_SUBDATASETS,
                                                         self.tr('Copy all subdatasets of this file to individual output files'),
                                                         defaultValue=False))
+        self.addParameter(QgsProcessingParameterBoolean(self.SCALE,
```
nyalldawson: Either "input pixel values" or "input pixel's value"

nyalldawson (commented):
nice! Can you add a unit test covering this to https://github.com/qgis/QGIS/blob/master/python/plugins/processing/tests/GdalAlgorithmsRasterTest.py#L132 ?

python/plugins/processing/algs/gdal/translate.py
gioman: Right (I just copied the wording directly from the GDAL docs).

alexbruy:
Do we really need this new parameter? All GDAL algorithms now have an optional "Additional command-line parameters" parameter, which was added for cases when one need to pass a command-line argument(s) not exposed in the algorithm UI.

gioman:
> Do we really need this new parameter? All GDAL algorithms now have an optional "Additional command-line parameters" parameter, which was added for cases when one need to pass a command-line argument(s) not exposed in the algorithm UI.

@alexbruy Hi Alex, isn't that only for "layer creation options" (-lco)?

alexbruy:
No. I have added parameter to all GDAL tools exactly for passing options not exposed in the UI (like this one and in #31692). There is "Additional creation options" parameter for -lco and "Additional command-line parameters" for passing command line arguments.

gioman:
> No. I have added parameter to all GDAL tools exactly for passing options not exposed in the UI (like this one and in #31692). There is "Additional creation options" parameter for -lco and "Additional command-line parameters" for passing command line arguments.

not stale. On second thoughts I think that exposing this parameter would be a valuable addition for the users.

gioman:
Not stale. This should be merged.

nyalldawson:
@gioman it's still missing tests

PR #31610, opened 2019-09-09
On debug builds, show an obnoxious, super-visible error in the message bar...

nyalldawson:
...whenever a Qt warning is thrown

These reveal deep, underlying issues in QGIS, and should NEVER be ignored

### DIFF
--- a/src/app/main.cpp
+++ b/src/app/main.cpp
@@ -105,6 +105,7 @@ typedef SInt32 SRefCon;
 #include "qgsversionmigration.h"
 #include "qgsfirstrundialog.h"
 #include "qgsproxystyle.h"
+#include "qgsmessagebar.h"
 
 #include "qgsuserprofilemanager.h"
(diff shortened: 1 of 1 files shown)


NathanW2:
What kind of things have we seen logged here?

nyalldawson:
@NathanW2 

Here's one relating to #31611 
image

NathanW2:
Right yeah that is a good one to raise to the UI

nyalldawson:
@NathanW2 I've seen others too - there's a sporadic one thrown by the attribute table about use of an index from a different model, which I've yet to investigate. But it's probably a very serious one too!

nyalldawson:
@nathanw2 happy for me to merge? this has identified (and resulted in a fix for) at least one serious issue, and i suspect will reveal many more...

NathanW2:
I have no problem with it.  No users will see it as it's only in debug.

PR #31613, opened 2019-09-09
Address QGIS 2.x -> 3.x regression in offset features tool

nyalldawson:
In QGIS 2 it was possible to offset features from a non-active layer
which would result in an offset copy of these features being inserted
into the (editable) active layer

This behavior was lost in QGIS 3, which only permits offsetting
features within the active layer.

Resurrect this behavior, with some tweaks:
- the copy-and-offset mode is activated only when ctrl is held
while offsetting. In earlier 3.x versions this also resulted in
copy-and-offset, but only for features within the current layer.
Now, copy-and-offset respects the project's snapping configuration
and potentially copies from other layers
- We use refined logic to correctly map attributes across from
the source feature layer to the destination layer, and apply
default values and provider constraints on feature copy (also
applies to copy-and-offsets where the source and destination
layer are the same)

### DIFF
--- a/src/app/qgsmaptooloffsetcurve.cpp
+++ b/src/app/qgsmaptooloffsetcurve.cpp
@@ -31,6 +31,7 @@
 #include "qgisapp.h"
 #include "qgsmapmouseevent.h"
 #include "qgslogger.h"
+#include "qgsvectorlayerutils.h"
 
 QgsMapToolOffsetCurve::QgsMapToolOffsetCurve( QgsMapCanvas *canvas )
   : QgsMapToolEdit( canvas )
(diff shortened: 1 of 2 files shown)


nyalldawson:
Note that the offset tool is currently completely broken as a result of https://github.com/qgis/QGIS/pull/31374/files

src/app/qgsmaptooloffsetcurve.cpp
```diff
@@ -71,15 +72,25 @@ void QgsMapToolOffsetCurve::canvasReleaseEvent( QgsMapMouseEvent *e )
...
 
-    QgsPointLocator::Match match = mCanvas->snappingUtils()->snapToCurrentLayer( e->pos(),
-                                   QgsPointLocator::Types( QgsPointLocator::Edge | QgsPointLocator::Area ) );
+    QgsPointLocator::Match match;
 
-    if ( ( match.hasEdge() || match.hasArea() ) && match.layer() )
+    if ( e->modifiers() & Qt::ControlModifier )
     {
-      mLayer = match.layer();
+      match = mCanvas->snappingUtils()->snapToMap( e->pos(), nullptr );
```
troopa81: This line won't work because of PR #31374

See my comment : https://github.com/qgis/QGIS/pull/31648#issuecomment-529652349

nirvn (approved):
Looks good to merge.

nyalldawson:
Thanks! Obviously it's not working correctly yet, due to https://github.com/qgis/QGIS/pull/31648#issuecomment-529652349, but that won't affect the 3.8 backport.

Issue #31568, opened 2019-09-05
QgsSingleBandPseudoColorRenderer.setBand() crashed QGIS


jakimowb:
**Problem**

The QgsSingleBandPseudoColorRenderer.setBand() crashes QGIS.

**How to get the error**

![image](https://user-images.githubusercontent.com/1404870/64318378-3285b180-cfba-11e9-8ef6-f147d2eb2036.png)

Let `layer` be a QgsRasterLayer with 3 bands and a QgsSingleBandPseudoColorRenderer on the first band. The following python lines crashed QGIS, at least on my windows plattform (all plugins disabled, new user profile):

```
assert isinstance(layer, QgsRasterLayer)
assert layer.bandCount() > 1
renderer = layer.renderer()
assert renderer.band() == 1
assert isinstance(renderer, QgsSingleBandPseudoColorRenderer)

# clone the render
newRenderer = renderer.clone()
newRenderer.setBand(2)
layer.setRenderer(newRenderer)
```

A QGIS Project file + python script is appended
[QgsSingleBandPseudoColorRendererCrash.zip](https://github.com/qgis/QGIS/files/3577830/QgsSingleBandPseudoColorRendererCrash.zip)

**Stack Trace**

```
QgsSingleBandPseudoColorRenderer::setBand :
PyInit__core :
PyMethodDef_RawFastCallKeywords :
PyMethodDef_RawFastCallKeywords :
PyEval_EvalFrameDefault :
PyEval_EvalCodeWithName :
PyImport_Import :
PyImport_Import :
PyMethodDef_RawFastCallKeywords :
PyMethodDef_RawFastCallKeywords :
PyEval_EvalFrameDefault :
PyMethodDef_RawFastCallKeywords :
PyEval_EvalFrameDefault :
PyEval_EvalCodeWithName :
PyMethodDef_RawFastCallKeywords :
PyEval_EvalFrameDefault :
PyEval_EvalCodeWithName :
PyMethodDef_RawFastCallKeywords :
PyEval_EvalFrameDefault :
PyMethodDef_RawFastCallKeywords :
PyEval_EvalFrameDefault :
PyMethodDef_RawFastCallKeywords :
PyEval_EvalFrameDefault :
PyMethodDef_RawFastCallKeywords :
PyEval_EvalFrameDefault :
PyFunction_FastCallDict :
PyMethodDef_RawFastCallDict :
PyObject_Call :
PyInit_sip :
PyInit_Qsci :
QWidget::event :
[...]

PR #31621, opened 2019-09-09
[singleband pseudocolor renderer] Fix crash when calling setBand with no input

nirvn:
## Description

This PR fixes a crasher with the singleband pseudocolor renderer whereas the setBand function did not prevent access to a null input raster interface pointer. Reported in #31568 .

## Checklist

### DIFF
--- a/src/core/raster/qgssinglebandpseudocolorrenderer.cpp
+++ b/src/core/raster/qgssinglebandpseudocolorrenderer.cpp
@@ -38,11 +38,16 @@ QgsSingleBandPseudoColorRenderer::QgsSingleBandPseudoColorRenderer( QgsRasterInt
 
 void QgsSingleBandPseudoColorRenderer::setBand( int bandNo )
 {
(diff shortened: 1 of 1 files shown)


src/core/raster/qgssinglebandpseudocolorrenderer.cpp
```diff
@@ -38,7 +38,7 @@ QgsSingleBandPseudoColorRenderer::QgsSingleBandPseudoColorRenderer( QgsRasterInt
 
 void QgsSingleBandPseudoColorRenderer::setBand( int bandNo )
 {
-  if ( bandNo > mInput->bandCount() || bandNo <= 0 )
+  if ( !mInput || bandNo > mInput->bandCount() || bandNo <= 0 )
```
nyalldawson: actually in this case we should skip the check, and just set the band number directly. Otherwise it breaks the workflow of invalid layers

nirvn: True.

nyalldawson (approved):

PR #31622, opened 2019-09-09
Improve UX when a missing grid transform is desired/required

nyalldawson:
Instead of just showing users the URL to download the grid from (and with no clues about what to DO with the downloaded file), also present them with a user-friendly dialog allowing them to install the grid to the required location:

Peek 2019-09-09 17-05

(ignore the messed up dialog stacking - that's the screen recorder's fault)

TODO: also expose this when the full transform selection dialog is shown

### DIFF
--- a/src/app/qgsinstallgridshiftdialog.cpp
+++ b/src/app/qgsinstallgridshiftdialog.cpp
@@ -68,6 +68,6 @@ void QgsInstallGridShiftFileDialog::installFromFile()
   }
   else
   {
-    QMessageBox::critical( this, tr( "Install Grid File" ), tr( "Could not copy %1 to %2. Please check folder permissions and ret" ).arg( mGridName, destPath ) );
+    QMessageBox::critical( this, tr( "Install Grid File" ), tr( "Could not copy %1 to %2. Please check folder permissions and retry." ).arg( mGridName, destPath ) );
   }
 }
--- a/src/app/qgsinstallgridshiftdialog.cpp
+++ b/src/app/qgsinstallgridshiftdialog.cpp
@@ -1,0 +1,73 @@
+/***************************************************************************
+                         qgsinstallgridshiftdialog.cpp
+                             -------------------
+    begin                : September 2019
+    copyright            : (C) 2019 by Nyall Dawson
+    email                : nyall dot dawson at gmail dot com
+ ***************************************************************************/
+
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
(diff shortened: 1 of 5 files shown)


m-kuhn:
So, the download and unzipping still needs to be done manually, but the copy to final destination is a one-click action?
Definitely a huge step in the right direction!

nyalldawson:
@m-kuhn yep. THere's actually a hidden button here which is "download and install", but that's not hooked up yet. And we'll need the manual install option for grids which aren't available in the standard packages (or for offline installs)

src/ui/qgsinstallgridshiftdialog.ui
```diff
@@ -0,0 +1,143 @@
...
+     </property>
+     <property name="readOnly">
+      <bool>true</bool>
+     </property>
+     <property name="html">
+      <string>&lt;!DOCTYPE HTML PUBLIC &quot;-//W3C//DTD HTML 4.0//EN&quot; &quot;http://www.w3.org/TR/REC-html40/strict.dtd&quot;&gt;
+&lt;html&gt;&lt;head&gt;&lt;meta name=&quot;qrichtext&quot; content=&quot;1&quot; /&gt;&lt;style type=&quot;text/css&quot;&gt;
+p, li { white-space: pre-wrap; }
+&lt;/style&gt;&lt;/head&gt;&lt;body style=&quot; font-family:'Cantarell'; font-size:11pt; font-weight:400; font-style:normal;&quot;&gt;
+&lt;p style=&quot; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;&lt;span style=&quot; font-family:'Cantarell';&quot;&gt;The preferred transform between EPSG:26952 [NAD83 / Arkansas South] and EPSG:4326 [WGS 84] is not available for use on the system. This transformation requires the grid file “CHENyx06a.gsb”, which is not available for use on the system.&lt;/span&gt;&lt;/p&gt;
```
nirvn: Can you remove the hard-coded example in the .ui file?

src/app/qgsinstallgridshiftdialog.cpp
```diff
@@ -0,0 +1,73 @@
...
+  const QString destPath = QgsApplication::qgisSettingsDirPath() + QStringLiteral( "proj/" ) + fi.fileName();
+
+  if ( QFile::copy( gridFilePath, destPath ) )
+  {
+    QMessageBox::information( this, tr( "Install Grid File" ), tr( "The %1 grid shift file has been successfully installed. Please restart QGIS for this change to take effect." ).arg( mGridName ) );
+    accept();
+  }
+  else
+  {
+    QMessageBox::critical( this, tr( "Install Grid File" ), tr( "Could not copy %1 to %2. Please check folder permissions and ret" ).arg( mGridName, destPath ) );
```
m-kuhn: "And ret", missing something??

nyalldawson: "retry". fixed.

PR #31631, opened 2019-09-09
Use floating connections from browser items

elpaso:
Fixes an unreported issue when deleting/renaming GPKG layers
from the browser if the connection was not stored


elpaso:
@nyalldawson this should fix the issue

nyalldawson (approved):
Looks good to me!

nyalldawson:
Not sure if this PR also fixes it, but the "create a new layer or table" action for gpkg is also broken

PR #31645, opened 2019-09-09
Tweak widgets visibility in decorations at the opening

DelazJ:
In View menu --> Decorations, the dialogs have an item that is enabled while all of them should be disabled by default (until the user hits "draw/enable.. decoration" checkbox. This PR fixes some of them but we stiil have Grid line symbol, and Image colors that would need some tweaks. @nirvn


nirvn:
@DelazJ , thanks, keep those papercut fixes coming!

Issue #31427, opened 2019-08-26
Loss of labels background from older project/style


DelazJ:
Layers whose labels have a background do not display that background in master 3.9, potentially breaking old map rendering.

How to reproduce:
1. Save in 3.8 a layer with style containing a label with background (attached file has labels buffer, background, shadow )
1. Open the layer in 3.9, the background is missing

QGIS a3f411acce Windows 10

[issue_background.zip](https://github.com/qgis/QGIS/files/3541884/issue_background.zip)

PR #31647, opened 2019-09-09
Fix text formatting QML and enable/disable

### LINKED ISSUE #31428: "Text formats" dialog shows modifiable buffer properties while the "draw text buffer" is unchecked
1. Open Style Manager, text format tab
1. Click + to add a new item
1. in the new dialog, browse the "Shadow", "Buffer", "Background" tabs; their properties are modifiable despite the top "Draw text XXX" is still unchecked.
1. If you were creating a "Label setting" item instead, you'd notice that the same properties are grayed until you check the "Draw text XXX" button

Wonder if it has to do with [the non persistent labels background issue](https://github.com/qgis/QGIS/issues/31427)

QGIS a3f411a Windows 10

elpaso:
Fixes #31428 and #31427

### DIFF
--- a/src/core/qgstextrenderer.cpp
+++ b/src/core/qgstextrenderer.cpp
@@ -805,6 +805,7 @@ void QgsTextBackgroundSettings::readFromLayer( QgsVectorLayer *layer )
     setPaintEffect( nullptr );
 }
 
+
 void QgsTextBackgroundSettings::readXml( const QDomElement &elem, const QgsReadWriteContext &context )
 {
   QDomElement backgroundElem = elem.firstChildElement( QStringLiteral( "background" ) );
@@ -1924,14 +1925,6 @@ void QgsTextFormat::readXml( const QDomElement &elem, const QgsReadWriteContext 
   else
   {
     mShadowSettings.readXml( textStyleElem );
-  }
-  if ( textStyleElem.firstChildElement( QStringLiteral( "background" ) ).isNull() )
-  {
-    mBackgroundSettings.readXml( elem, context );
-  }
-  else
-  {
-    mBackgroundSettings.readXml( textStyleElem, context );
   }
 
   QDomElement ddElem = textStyleElem.firstChildElement( QStringLiteral( "dd_properties" ) );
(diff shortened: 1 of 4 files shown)


elpaso:
@nyalldawson I'm not totally sure about the fix in case of DD properties for the background, mind having a look?

nyalldawson:
#31428 is deliberate, in that we also need these properties to be editable if data defined enable buffer is used.

src/core/qgstextrenderer.cpp
```diff
@@ -1925,25 +1926,24 @@ void QgsTextFormat::readXml( const QDomElement &elem, const QgsReadWriteContext
...
-    mBackgroundSettings.readXml( elem, context );
-  }
-  else
-  {
-    mBackgroundSettings.readXml( textStyleElem, context );
-  }
 
-  if ( textStyleElem.firstChildElement( QStringLiteral( "dd_properties" ) ).isNull() )
+  QDomElement ddElem = textStyleElem.firstChildElement( QStringLiteral( "dd_properties" ) );
+  if ( ddElem.isNull() )
```
nyalldawson: Hmm - this doesn't look right. We'd still want to restore these even when data defined properties are present. What's the logic here?

elpaso: > properties are present. What's the logic here?

I said it wasn't clear to me :)

The question is what is the current (broken) logic ? 

In case of missing DD and issue #31427, the QML is read twice:
First time here:
https://github.com/qgis/QGIS/blob/master/src/core/qgstextrenderer.cpp#L1934
which gets the correct config
and second time here:
https://github.com/qgis/QGIS/blob/master/src/core/qgstextrenderer.cpp#L1939
which gets the wrong config

The DD getter also seems to be duplicated:
here is checked intially https://github.com/qgis/QGIS/blob/master/src/core/qgstextrenderer.cpp#L1937
and here is checked again: https://github.com/qgis/QGIS/blob/master/src/core/qgstextrenderer.cpp#L1946
I moved the getter before to simplify the code.

elpaso:
> #31428 is deliberate, in that we also need these properties to be editable if data defined enable buffer is used.

Ok, I've committed some more complex logic to handle frame enabled but I have some doubts about the UX/UI: is the the DD is an override of the main checkbox?
If it is, I suppose it overrides the checkbox completely, so why the checkbox is automatically checked whenever I set a DD?

Anyway I think we should make this clearer to the user.

PR #31648, opened 2019-09-09
Parallelize snap caching

troopa81:
## Description

This PR fixes the issue related to this PR #31374.
EDIT: It now also contains PR #31374 commits because it has been reverted for 3.10.0 in #32297

The snapToCurrentLayer method doesn't prepare the index and expect to be executed synchronously. So the temporary point locator must be synchronous.

It also fixes the crashes on exit.

## Checklist

### DIFF
--- a/src/core/qgspointlocator.cpp
+++ b/src/core/qgspointlocator.cpp
@@ -645,6 +645,10 @@ bool QgsPointLocator::init( int maxFeaturesToIndex, bool relaxed )
 void QgsPointLocator::waitForFinished() const
 {
   mInitTask->waitForFinished();
+
+  // force process of signal emitted from task thread
+  // so rebuildIndexFinished and taskTerminated are called
+  QCoreApplication::processEvents();
 }
 
 bool QgsPointLocator::hasIndex() const
--- a/src/core/qgspointlocator.cpp
+++ b/src/core/qgspointlocator.cpp
@@ -22,11 +22,15 @@
 #include "qgis.h"
 #include "qgslogger.h"
 #include "qgsrenderer.h"
+#include "qgsapplication.h"
+#include "qgsvectorlayerfeatureiterator.h"
 #include "qgsexpressioncontextutils.h"
 #include "qgslinestring.h"
+#include "qgspointlocatorinittask.h"
 #include <spatialindex/SpatialIndex.h>
 
 #include <QLinkedListIterator>
+#include <QtConcurrent>
 
 using namespace SpatialIndex;
 
@@ -554,6 +558,10 @@ QgsCoordinateReferenceSystem QgsPointLocator::destinationCrs() const
 
 void QgsPointLocator::setExtent( const QgsRectangle *extent )
 {
+  if ( mIsIndexing )
+    // already indexing, return!
+    return;
+
   mExtent.reset( extent ? new QgsRectangle( *extent ) : nullptr );
 
   destroyIndex();
@@ -561,6 +569,10 @@ void QgsPointLocator::setExtent( const QgsRectangle *extent )
 
 void QgsPointLocator::setRenderContext( const QgsRenderContext *context )
 {
+  if ( mIsIndexing )
+    // already indexing, return!
+    return;
+
   disconnect( mLayer, &QgsVectorLayer::styleChanged, this, &QgsPointLocator::destroyIndex );
 
   destroyIndex();
@@ -574,27 +586,103 @@ void QgsPointLocator::setRenderContext( const QgsRenderContext *context )
 
 }
 
-bool QgsPointLocator::init( int maxFeaturesToIndex )
+void QgsPointLocator::onInitTaskTerminated()
 {
-  return hasIndex() ? true : rebuildIndex( maxFeaturesToIndex );
+  mIsIndexing = false;
+  mRenderer.reset();
+  mSource.reset();
 }
 
+void QgsPointLocator::onRebuildIndexFinished( bool ok )
+{
+  onInitTaskTerminated();
+
+  // treat added and deleted feature while indexing
+  for ( QgsFeatureId fid : mAddedFeatures )
+    onFeatureAdded( fid );
+
+  for ( QgsFeatureId fid : mDeletedFeatures )
+    onFeatureDeleted( fid );
+
+  emit initFinished( ok );
+}
+
+bool QgsPointLocator::init( int maxFeaturesToIndex, bool relaxed )
+{
+  const QgsWkbTypes::GeometryType geomType = mLayer->geometryType();
+  if ( geomType == QgsWkbTypes::NullGeometry // nothing to index
+       || hasIndex()
+       || mIsIndexing ) // already indexing, return!
+    return true;
+
+  mRenderer.reset( mLayer->renderer() ? mLayer->renderer()->clone() : nullptr );
+  mSource.reset( new QgsVectorLayerFeatureSource( mLayer ) );
+
+  if ( mContext )
+  {
+    mContext->expressionContext() << QgsExpressionContextUtils::layerScope( mLayer );
+  }
+
+  mIsIndexing = true;
+
+  if ( relaxed )
+  {
+    mInitTask = new QgsPointLocatorInitTask( this );
+    connect( mInitTask, &QgsPointLocatorInitTask::rebuildIndexFinished, this, &QgsPointLocator::onRebuildIndexFinished );
+    connect( mInitTask, &QgsPointLocatorInitTask::taskTerminated, this, &QgsPointLocator::onInitTaskTerminated );
+    QgsApplication::taskManager()->addTask( mInitTask );
+    return true;
+  }
+  else
+  {
+    const bool ok = rebuildIndex( maxFeaturesToIndex );
+    mIsIndexing = false;
+    emit initFinished( ok );
+    return ok;
+  }
+}
+
+void QgsPointLocator::waitForFinished() const
+{
+  mInitTask->waitForFinished();
+}
 
 bool QgsPointLocator::hasIndex() const
 {
-  return mRTree || mIsEmptyLayer;
+  return mIsIndexing || mRTree || mIsEmptyLayer;
 }
 
+bool QgsPointLocator::prepare( bool relaxed )
+{
+  if ( mIsIndexing )
+  {
+    if ( relaxed )
+      return false;
+    else
+      waitForFinished();
+  }
+
+  if ( !mRTree )
+  {
+    init( -1, relaxed );
+    if ( ( relaxed && mIsIndexing ) || !mRTree ) // relaxed mode and currently indexing or still invalid?
+      return false;
+  }
+
+  return true;
+}
 
 bool QgsPointLocator::rebuildIndex( int maxFeaturesToIndex )
 {
+  QTime t;
+  t.start();
+
+  QgsDebugMsg( QStringLiteral( "RebuildIndex start : %1" ).arg( mSource->id() ) );
+
   destroyIndex();
 
   QLinkedList<RTree::Data *> dataList;
   QgsFeature f;
-  QgsWkbTypes::GeometryType geomType = mLayer->geometryType();
-  if ( geomType == QgsWkbTypes::NullGeometry )
-    return true; // nothing to index
 
   QgsFeatureRequest request;
   request.setNoAttributes();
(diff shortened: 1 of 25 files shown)


nyalldawson (commented):
Thanks!

troopa81:
I think there is other problems with my previous PR #31374.

snapToCurrenLayer is always synchronous while snapToMap depends on the asynchronous parameter. 

Indeed, the idea of the PR was to make snapToMap in QGIS application asynchronous, so while the user move the cursor, the snap cache is built, and the snapping information are displayed as soon as they are built.

But, it can be possible that people wants to use the snapToMap method synchronously in the application (like in PR #31613 for instance), and that won't work.

I'm considering 3 possibilities:
- Let's the way it is, snapToMap is always asynchronous is QGIS app, tool that want to call this method has to call it on mouseMove (and display snapping information) so it will be ready when clicking
- Move asynchronous boolean on snapToMap method. If synchronous, prepareIndex waits for point locator to be finished. It looks a good solution but it's not easy to wait for QRunnable thread to be finished
- make snapToMap always synchronous, replace QgsTaskManager with a QtConcurrent blocking call. It will still be quicker than before the PR, but user will have to wait for all indexes to be ready.

wonder-sk:
Patching just snapToCurrentLayer seems a bit ad-hoc and as you say, there may be problems with other methods as well (including the ones in QgsPointLocator).

I would suggest to keep the original API calls (snapToMap, snapToCurrentLayer etc) all synchronous, and introduce new methods with a new naming (e.g. snapToMapRelaxed?) that would be allowed to return invalid results if the index is being built... Then we would just modify some places to use the "relaxed" version so that this choice / optimization is always deliberate. A new variant of those functions compared to addition of a new true/false flag is IMHO a better indication that some special behavior is requested...

troopa81:
@wonder-sk Thank you for your advices. I agree with your comments and will quickly propose a rewrite of my code to provide a snapToMap and snapToMapRelaxed methods. 

I propose to not share the asynchronous and synchronous locators, so snapToMap will deal with its own synchronous locators map and snapToMapRelaxed with its asynchronous locators map.

troopa81:
I added a snapToMapRelaxed method and also a boolean to get asynchronous point locators (needed in qgsvertextools). I use this method for maptools which provides some snapping information feedback and let the other use the classic snapToMap method.

nyalldawson:
@wonder-sk?

haubourg:
@wonder-sk any opinion here?

troopa81:
Please don't stale

nyalldawson:
@troopa81 3.10 is in hard freeze now and we've still got many open bugs relating to snapping due to this work not being complete/merged in time. Can you please open a pr reverting the earlier work for 3.10.0 asap so that we can get master back in release-ready state?

troopa81:
@nyalldawson I would rather fix bugs you are mentionning than postponed this work.

I take a quick look on qgis issues, and fail to find other issues related to this work. There is, if I recall correctly, 5 map tools that are currently broken and that PR should fix these. Is there other issues I'm not aware of?

Most important thing I think is to agree on the API, because we won't change it after hard freeze. @wonder-sk did you have the time to check API changes?

Anyway, if there is too much issues or complexity related to this work, I will revert my commit but again, I would prefer not to.

mhugo:
@troopa81 About your proposed API changes, could you confirm what should be approved. I see it sums up to these two changes:
- QgsSnappingUtils constructor : the optional "asynchronous = false" parameter is removed
- QgsPointLocator *QgsSnappingUtils::locatorForLayer() : one new optional parameter "asynchronous = false" is added.

Is there something else ?

troopa81:
@mhugo there is also the new method QgsSnappingUtils::snapToMapRelaxed()

But this PR is the following of this one https://github.com/qgis/QGIS/pull/31374

Actually the API changes bring by this work (the 2 PR) is:
- new method QgsSnappingUtils::snapToMapRelaxed()
- QgsPointLocator *QgsSnappingUtils::locatorForLayer() : one new optional parameter "asynchronous = false" is added.
- QgsPointLocator constructor: one new optional parameter "asynchronous = false" is added.
- new method QgsPointLocator::isIndexing

wonder-sk:
Oops sorry everyone - my notifications to this PR got lost in my mail among 500+ other unread qgis notifications :-( Going to have a look...

wonder-sk:
Hmm I am not sure I understand why we need to have separately synchronous and asynchronous locators... Could you please explain the reasoning?

I am worried that this just adds extra memory+CPU overhead and higher code complexity without an obvious reason. I would expect to have just a single locator for each layer - if the locator is being indexed in the background and there's a call to non-relaxed version of snapToMap (or other function), it would simply block until the indexing in the background has finished.

For QgsPointLocator, rather than having the whole class either sync or async, it may be good to have init() and initAsync() and also have relaxed versions of the search functions like nearestVertex or nearestEdge.

troopa81:
@wonder-sk  Thanks for your review

> I am worried that this just adds extra memory+CPU overhead

Yes, indeed

> Could you please explain the reasoning?

Main reasons was : 
- I cannot see a way by the time to wait for QgsPointLocatorInitTask to finish, but it looks like this is exactly the purpose of QgsTask::waitForFinished, so my bad
- I didn't want to pass an asynchronous parameter all the time like you can see in this commit 3de2eff

> it may be good to have init() and initAsync() and also have relaxed versions of the search functions like nearestVertex or nearestEdge.

Don't you think that adding a relaxed version for all 8 methods (nearestVertex, nearestArea, edgesInrect...) will add complexity, meaning it would add 3 methods per each(nearestVertex, nearestVertexRelaxed and _nearestVertex private one) ? Isn't an asynchronous boolean (default to false) a better option?

Same remark apply to snapToMap

wonder-sk:
> Don't you think that adding a relaxed version for all 8 methods (nearestVertex, nearestArea, edgesInrect...) will add complexity, meaning it would add 3 methods per each(nearestVertex, nearestVertexRelaxed and _nearestVertex private one) ? Isn't an asynchronous boolean (default to false) a better option?

Yeah, it is a bit tricky decision to do. My intention was mainly to avoid the boolean trap in API, to have the code easier to read: https://ariya.io/2011/08/hall-of-api-shame-boolean-trap
But I can live also with the extra boolean in the calls. Just not sure if 'asynchronous' would be a good name for the parameter - the queries are not asynchronous, it's rather an indication if we can live with incorrect result or not :-)

troopa81:
> My intention was mainly to avoid the boolean trap in API, to have the code easier to read: https://ariya.io/2011/08/hall-of-api-shame-boolean-trap

interesting read, I understand your point. But I think I prefer the boolean way over having too much methods. 

If we stick to the boolean way, I propose to remove the snapToMapRelaxed and add a boolean to snapToMap in order to have the QgsSnappingUtils and QgsPointLocator consistent. Are you agree?

>  Just not sure if 'asynchronous' would be a good name for the parameter

I renamed it "relaxed".

wonder-sk:
> If we stick to the boolean way, I propose to remove the snapToMapRelaxed and add a boolean to snapToMap in order to have the QgsSnappingUtils and QgsPointLocator consistent. Are you agree?

Yeah sounds good to me!

troopa81:
> Yeah sounds good to me!

Done b06cee2 

I had to fix QgsTaskManager::waitForFinished 1bde15f so it waits the task to be started before waiting to be finished

tests/src/python/test_qgslayoutlegend.py
```diff
@@ -463,7 +458,7 @@ def testSymbolExpressionRender(self):
         group = legend.model().rootGroup().addGroup("Group [% 1 + 5 %] [% @layout_name %]")
         layer_tree_layer = group.addLayer(point_layer)
         counterTask = point_layer.countSymbolFeatures()
-        counterTask.waitForFinished() # does this even work?
+        counterTask.waitForFinished()
```
roya0045: Thanks for solving this, thought this might be an out of place fix for this PR.

troopa81: Yes, I was thinking about making a PR for the waitForFinished fix and maybe backporting it to LTR because it may produce some crashes.

roya0045: I would assume that core devs will suggest having that change as a standalone PR.

troopa81: Done https://github.com/qgis/QGIS/pull/32453

troopa81:
I create a PR https://github.com/qgis/QGIS/pull/32453 for waitforfinished fix. This later should be merged before this one.

src/core/qgspointlocator.cpp
```diff
@@ -574,27 +586,107 @@ void QgsPointLocator::setRenderContext( const QgsRenderContext *context )
...
+void QgsPointLocator::onRebuildIndexFinished( bool ok )
+{
+  onInitTaskTerminated();
+
+  // treat added and deleted feature while indexing
+  for ( QgsFeatureId fid : mAddedFeatures )
+    onFeatureAdded( fid );
+
+  for ( QgsFeatureId fid : mDeletedFeatures )
+    onFeatureDeleted( fid );
```
wonder-sk: Do these arrays of added/deleted features ever get cleared?

src/core/qgspointlocator.cpp
```diff
@@ -574,27 +586,107 @@ void QgsPointLocator::setRenderContext( const QgsRenderContext *context )
...
+  }
+}
+
+void QgsPointLocator::waitForFinished() const
+{
+  mInitTask->waitForFinished();
+
+  // force process of signal emitted from task thread
+  // so rebuildIndexFinished and taskTerminated are called
+  QCoreApplication::processEvents();
```
wonder-sk: Hmm... whenever I see processEvents() call, I get really scared :-) This call is very unpredictable in its side effects unfortunately... for example, it can also get into processing of the next mouse move event (while in the middle of processing the current mouse move event) and completely break the expected flow of code, often leading to crashes.

Is there any chance we can get rid of sending a signal from the task from the background thread and instead just use the services of task manager to find out that the task has finished?

src/core/qgssnappingutils.cpp
```diff
@@ -74,17 +81,18 @@ QgsPointLocator *QgsSnappingUtils::temporaryLocatorForLayer( QgsVectorLayer *vl,
...
 
-bool QgsSnappingUtils::isIndexPrepared( QgsVectorLayer *vl, const QgsRectangle &areaOfInterest )
+bool QgsSnappingUtils::isIndexPrepared( QgsPointLocator *loc, const QgsRectangle &areaOfInterest )
 {
-  if ( vl->geometryType() == QgsWkbTypes::NullGeometry || mStrategy == IndexNeverFull )
-    return false;
-
-  QgsPointLocator *loc = locatorForLayer( vl );
+  if ( loc->isIndexing() )
+    return true;
```
wonder-sk: I'm not sure if I understand the logic here... if locator is still indexing, how can we return "true" in isIndexPrepared() ?

src/core/qgspointlocator.h
```diff
@@ -251,68 +260,116 @@ class CORE_EXPORT QgsPointLocator : public QObject
...
+     *
+     * \see QgsPointLocator()
+     */
+    bool isIndexing() const { return mIsIndexing; }
+
+    /**
+     * If the point locator has been initialized relaxedly and is currently indexing,
+     * this methods waits for the indexing to be finished
+     */
+    void waitForFinished() const;
```
wonder-sk: Can we rename to something like waitForIndexingFinished() so it is clear it is related to indexing, not searching?

wonder-sk (commented):
The code is looking really good, just few notes from me...

src/core/qgspointlocator.h
troopa81: Done

src/core/qgssnappingutils.cpp
troopa81: It was to factorize this piece of code but it's strange indeed, I extract it from isIndexPrepared d0e26fb

src/core/qgspointlocator.cpp
troopa81: Yes, this way should be better d246d9f

src/core/qgspointlocator.cpp
troopa81: Done

wonder-sk:
Thanks for the updates! Can we get rid of the rebuildIndexFinished() signal as well? it should be unused now - then I think we should be good to go :)

troopa81:
> Thanks for the updates! Can we get rid of the rebuildIndexFinished() signal as well?

Yes, done! Thanks for the time spending reviewing this PR (and the one before)

wonder-sk (approved):
Cool - let's get it merged!

PR #31673, opened 2019-09-10
Server api wfs3 acl

elpaso:
QGIS Server WFS3 ACL + alias + excluded fields

Takes into account aliases, excluded attributes and ACL plugins

I classify this as a bugfix, since it fills the configuration gap between actual WFS and WFS3.


elpaso:
@pblottiere mind having a look?

PR #31663, opened 2019-09-10
Move info bar creation before welcome page (fixes crash on start)

wonder-sk:
Welcome page creates QgsTemplateProjectsModel which adds a file watcher which triggers a Qt warning which tried to use null info bar.

PR #31682, opened 2019-09-11
Revert "Move info bar creation before welcome page (fixes crash on start)"

m-kuhn:
Reverts qgis/QGIS#31663

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -825,13 +825,6 @@ QgisApp::QgisApp( QSplashScreen *splash, bool restorePlugins, bool skipVersionCh
   // what type of project to auto-open
   mProjOpen = settings.value( QStringLiteral( "qgis/projOpenAtLaunch" ), 0 ).toInt();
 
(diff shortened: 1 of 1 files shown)


nirvn:
@m-kuhn , the reason why I haven't done this yet is because I think it might be safer to figure out why the UI placement is broken, vs. reverting the commit altogether. It might be the case that the welcome page triggers more message bar content in the future, which would re-trigger the crash.

m-kuhn:
Maybe this https://doc.qt.io/qt-5/qwidget.html#stackUnder ?

nirvn:
@m-kuhn , maybe. I think it's the parameter of the centralLayout->addWdiget(...) calls which might need modification. 

Before, we were pushing the central container via addWidget( mCentralContainer, 0, 0, 2, 1 ) and then the bar via addWidget( mInfoBar, 0, 0, 1, 1). By using 0 values for the fromRow and fromColumn parameters in both function, that stacked the bar to be above the central container. I'm trying to tweak those numbers ATM to see if that fixes things. Waiting for `make -j6` to end its job :)

PR #31686, opened 2019-09-11
[point displacement renderer] Fix crasher when trying to access out of bound QVector element

nirvn:
## Description

@nyalldawson , I stumbled on a crasher with the point displacement renderer, whereas the computing of ring displacement was trigger a QVector out of bound assert. The PR avoids doing so by adding an additional check in the appropriate loop. Not sure if there's an underlying issue that needs addressing.

## Checklist


nyalldawson:
Can you still reproduce this? Like you guessed, it's just masking a deeper underlying issue, because:

- when calculateSymbolAndLabelPositions is called, nPosition MUST equal the size of diagonals
- following through the logic from there, featureIndex at line 297 must be < size of diagonals

So something is screwy somewhere, but I can't see where from the code alone.

nyalldawson:
Superseded by #31702 31702

PR #31687, opened 2019-09-11
Add iface.pluginHelpMenu() 

nyalldawson:
Returns a reference to the main window "Plugin Help" sub-menu.

Plugins are encouraged to insert help and about actions in this submenu instead of creating
a submenu under the pluginMenu() which solely contains Plugin Help or About actions.

For example, a tightly-integrated plugin which only implements a processing provider should not
create additional ui menus just to link to the plugin help pages. Rather, they should use this new submenu (available under the Help menu) to insert their help links.

### DIFF
--- a/python/gui/auto_generated/qgisinterface.sip.in
+++ b/python/gui/auto_generated/qgisinterface.sip.in
@@ -208,6 +208,16 @@ Returns a reference to the main window "Settings" menu.
 Returns a reference to the main window "Plugin" menu.
 %End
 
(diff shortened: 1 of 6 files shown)


nirvn:
Nice, that'd indeed be a nice UX improvement should authors embrace this. If we want this to be adopted anytime soon, we'd need to push it into 3.4 as well as 3.10. I offer my preemptive +1 to that.

nyalldawson:
Let's just merge for 3.10 -- it's the new LTR. I don't see any pressing need to backport.

PR #31681, opened 2019-09-11
[needs-docs] Do not display graduated method combobox

DelazJ:
when the layer is of polygon geometry type to avoid a single item combobox (fixes #31638)

## Checklist

- [ ] Commit messages are descriptive and explain the rationale for changes
- [ ] Commits which fix bugs include `Fixes #11111` at the bottom of the commit message
- [ ] I have read the [QGIS Coding Standards](https://docs.qgis.org/testing/en/docs/developers_guide/codingstandards.html) and this PR complies with them
- [ ] New unit tests have been added for core changes
- [ ] I have run [the `scripts/prepare-commit.sh` script](https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTING.md#contributing-to-qgis) before each commit
- [ ] I have evaluated whether it is appropriate for this PR to be backported, backport requests are left as label or comment

PR #31700, opened 2019-09-12
Graduated renderer fixes

nyalldawson:
Supersedes #31681


src/core/qgsunittypes.cpp
```diff
@@ -1624,6 +1624,7 @@ double QgsUnitTypes::fromUnitToUnitFactor( QgsUnitTypes::VolumeUnit fromUnit, Qg
       return 1.0;
     }
   }
+  return 1.0;
```
nirvn: Is that change intentional?

nirvn (approved):
Nice cleanup.

src/core/qgsunittypes.cpp
nyalldawson: yeah, it was just throwing up a warning during compiling

PR #31702, opened 2019-09-12
Fix crash in point displacement renderer

nyalldawson:
Supersedes #31686


nirvn:
@nyalldawson , thanks.

PR #31710, opened 2019-09-12
Fix cannot use GPS panel to draw features in multi polygon/line 

nyalldawson:
And cleanup a bunch of code, and fix other unreported issues


jonnyforestGIS:
@Nyall can I help you testing this new fixes? I have a tablet with Windows OS where I can test these fixes on the field with master?

jonnyforestGIS:
Hi, @nyalldawson.  I will start the tests with the master. I already see that if I add a Polygon in the layers tree GPS Information Panel recognizes because the button label change to Add Polygon.

Any tip on how I can do the tests to share with you? File formats, different geometries? With and without Z? Others?

Another thing this fix will be backported to 3.8.x and 3.4.x or only to 3.8.?

Regards,
João

PR #31724, opened 2019-09-12
Aggregate calculations on empty set should return NULL (not 0)

m-kuhn:
## Description

 Fix aggregate calculations for result size of 0

Previously:
```sql
sum([]) -> 0
```

Now:
```sql
sum([]) -> NULL
```

Detected in https://github.com/qgis/QGIS/pull/30972

### DIFF
--- a/python/plugins/processing/tests/testdata/expected/aggregate_field.gml
+++ b/python/plugins/processing/tests/testdata/expected/aggregate_field.gml
@@ -24,8 +24,6 @@
       <ogr:geometryProperty><gml:MultiPolygon srsName="EPSG:4326"><gml:polygonMember><gml:Polygon><gml:outerBoundaryIs><gml:LinearRing><gml:coordinates>6.24145873320538,-0.054510556621882 7.24145873320538,-1.05451055662188 5.24145873320538,-1.05451055662188 6.24145873320538,-0.054510556621882</gml:coordinates></gml:LinearRing></gml:outerBoundaryIs></gml:Polygon></gml:polygonMember></gml:MultiPolygon></ogr:geometryProperty>
       <ogr:fids>polys.1,polys.9</ogr:fids>
       <ogr:name>dd</ogr:name>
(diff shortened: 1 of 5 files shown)


m-kuhn:
A random guess: is it possible that it was mostly to fix `count([])` etc? Because they should return 0, not NULL.

nirvn:
I...do...not...remember; how odd is it to read oneself and have no idea why the comment was written then! :) 

I *suspect* that this was an issue very similar to a discussion we once had over to_date('improper') returning null (leading to field calculator unable to update large datasets when hitting null etc.). That discussion led to the addition of the try() expression function.

With the existence of try(), I wouldn't have any problem with sum([]) being null.

nirvn:
At the time, I'd have been messing around with battleship-as-QGIS-project, and I was doing some geometry overlap sums to check whether hits felt over ships or not. The null return ATM must have been problematic. It wouldn't be now when surrounded by a try().

m-kuhn:
But then we'd need to raise an error if sum is called with an empty set to trigger the try? Would this be appropriate?

nirvn:
Hmm actually you are right, this is more of a  coalesce() case. Then I'm totally at lost as to what was the problem back in 2016 :)

m-kuhn:
It also returns NULL for [Oracle](https://stackoverflow.com/a/6530371/2319028) and in the [stack exchange answer for MSSQL there is even an answer referencing the standard](https://dba.stackexchange.com/a/92128/170323).

Should we follow the SQL lead or keep it for the QGIS 3 series because the risk involved in changing this is too high?

m-kuhn:
Bump
@nyalldawson any feedback?
I don't intend to spend any energy on this if there is no support for it by other core devs.

nyalldawson:
@m-kuhn 
Alright, proceed away...

nyalldawson:
@m-kuhn time to merge this one?

PR #31726, opened 2019-09-12
Hash expressions

lbartoletti:
## Description

This PR adds hash functions to QGIS. No new dependency is required since it uses [QCryptographicHash](https://doc.qt.io/qt-5/qcryptographichash.html) available in QtCore.

Use case is the same as PgCrypto but in client side for all supported format.
 
## Checklist


nyalldawson (approved):
Just gotta fix the build ;)

lbartoletti:
> hash(value, algorithm='keccak_224')

:+1: 

I would like to keep also separated functions since I often use md5(value) / sha256(value) etc. :)
Do you agree?

Gustry:
IMHO, if it's a single function, it may fit in the already existing group `Conversions` with `to_hash` (same pattern as existing QGIS expressions) ?

m-kuhn:
I can see the use for shorthands for sha256 and md5, but I wonder if this necessarily means that we have to provide all of them. No strong opinion.
Concerning the method naming, I'd stick to `hash`, hasing is a one-way operation (in contrast to the usual to_xxx cast conversions)

NathanW2:
Yeah I think just a single hash function makes the most sense. Avoids
adding more functions then are needed

lbartoletti:
So, only one function `hash(value, string)`. Dedicated group or existing group (which one : `Conversions`? )?

Thanks

m-kuhn:
How about
* `hash(value, method)`
* `md5(value)`
* `sha256(value)`
* and `keccak256(value)`?

in `Conversions`?

lbartoletti:
@m-kuhn done. Good for you?

src/core/expression/qgsexpressionfunction.cpp
```diff
@@ -4991,8 +4991,90 @@ static QVariant fcnFileSize( const QVariantList &values, const QgsExpressionCont
...
+  return QString( QCryptographicHash::hash( str.toUtf8(), algorithm ).toHex() );
+}
 
+static QVariant fcnGenericHash( const QVariantList &values, const QgsExpressionContext *, QgsExpression *parent, const QgsExpressionNodeFunction * )
+{
+  QVariant hash;
+  QString str = QgsExpressionUtils::getStringValue( values.at( 0 ), parent );
+  QString method = QgsExpressionUtils::getStringValue( values.at( 1 ), parent ).toLower();
+
+  if ( method == QString( "md4" ) )
```
```suggestion
  if ( method == QLatin1String( "md4" ) )
```

m-kuhn:
Yep, pretty much, thanks :+1:

m-kuhn (approved):

Issue #26087, opened 2018-02-23
Choose a colour for nodata values


qgib:
Author Name: **Jon Morris** (Jon Morris)
Original Redmine Issue: [18196](https://issues.qgis.org/issues/18196)

Redmine category:rasters

---
There may already be a feature request in for this, but I couldn't find one. Please could we have an option to set the colour of nodata in raster layers?

There is a workaround (see https://gis.stackexchange.com/questions/75865/how-to-display-nodata-of-a-raster-in-qgis), but it would be much easier if there was a dropdown as there is in ArcMap.

vesnikos:
This is an on-going annoyance in my QC procedures. 
There is no way to colour the nodata with a colour of my choosing. 

I don't understand why both Arcgis and Arcgis pro now have that functionality and qgis does not :/

gioman:
> I don't understand why both Arcgis and Arcgis pro now have that functionality and qgis does not :/

why is so strange? I could make several examples of the contrary.

vesnikos:
if you are suggesting that the decision for this functionality was a conscious one, can you please redirect me to the relevant discussion. I am VERY much interested to see the arguments in it. 

As it stands right now, and correct me if I am wrong, the framework to support this functionality is already in place.

gioman:
> if you are suggesting that the decision for this functionality was a conscious one, can you please redirect me to the relevant discussion. I am VERY much interested to see the arguments in it.

I'm not sure but I just think that it was just an overlooked aspect of the raster symbology. I suggest you to raise a wider discussion by writing in the developers mailing list.

PR #31728, opened 2019-09-13
Add option to set color for rendering nodata pixels in raster layers

nyalldawson:
Allows raster nodata pixels to be colored in a specific (non-transparent) color. Designed to match the "Display background value" option from ArcMap.

Sponsored by SLYR


nirvn:
This was reported a while back as an implementation gap by ESRI users migrating to QGIS (see for e.g. issue #26087).

nirvn:
@nyalldawson , tested the branch, nodata color working with all of our raster renderers, and the UI looks nice. Good job:
Screenshot from 2019-09-13 14-47-45

nirvn (approved):

PR #31731, opened 2019-09-13
fix  broken gpkg actions

### LINKED ISSUE #31730: Broken GPKG browser actions
On master, some browser actions for GPKG files are broken:
- "compact database" (can't retrieve connection)
- "create a new layer or table" - no error, nothing happens

elpaso:
Fix #31730

@PeterPetrik can you please have a look to the itemguiprovider part?
I moved it into a lambda because it's simpler and because the original implementation did not work due to the fact that `data[QStringLiteral( "item" )].value<QPointer< QgsGeoPackageCollectionItem >>();` was invalid.

What was the reason for the QPointer btw? I cannot guess the reason why the data item could be destroyed when the action is triggered.

### DIFF
--- a/src/core/providers/ogr/qgsgeopackagedataitems.cpp
+++ b/src/core/providers/ogr/qgsgeopackagedataitems.cpp
@@ -211,11 +211,11 @@ void QgsGeoPackageCollectionItem::deleteConnection()
   mParent->refreshConnections();
 }
 
-bool QgsGeoPackageCollectionItem::vacuumGeoPackageDb( const QString &name, QString &errCause )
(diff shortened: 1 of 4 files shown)


PeterPetrik:
re `lambda`, not sure if that is related to this: https://github.com/qgis/QGIS/pull/30234/files?file-filters%5B%5D=.cpp&file-filters%5B%5D=.h#diff-10241ee6591720ea5b27935f35ea5784L262

re `QPointer`, do not remember, maybe to have something that fits into `QVariantMap`?

elpaso:
@PeterPetrik thanks for checking! 
I've added back the guarded pointer, now we should be on the safe side.

PR #31735, opened 2019-09-13
[processing] add support for WFS as input (fix #29663)

### LINKED ISSUE #29663: Processing GDAL algorithms does not handle correctly WFS input layers
Original Redmine Issue: [21848](https://issues.qgis.org/issues/21848)
Affected QGIS version: 3.6.1
Redmine category:processing/ogr
Assignee: Giovanni Manghi

---
the command is not created the right way, it must be something along this lines:

ogr2ogr -f PostgreSQL PG:"dbname='database' host=host port=5432 user='xxx' password='xxx'" WFS:"http://serviceaddress?SERVICE=WFS&VERSION=1.0.0" layername -nln schemaname.tablename

alexbruy:
## Description

Correctly generate connection string for WFS layers, making them usable with GDAL algorithms. 
Fixes #29663.

## Checklist

### DIFF
--- a/python/plugins/processing/algs/gdal/GdalAlgorithm.py
+++ b/python/plugins/processing/algs/gdal/GdalAlgorithm.py
@@ -32,7 +32,8 @@ from qgis.core import (QgsApplication,
                        QgsProcessingAlgorithm,
                        QgsProcessingContext,
                        QgsProcessingFeedback,
-                       QgsProviderRegistry)
+                       QgsProviderRegistry,
+                       QgsDataSourceUri)
 
 from processing.algs.gdal.GdalAlgorithmDialog import GdalAlgorithmDialog
 from processing.algs.gdal.GdalUtils import GdalUtils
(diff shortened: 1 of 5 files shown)


alexbruy:
Unfortunately without tests, as I can not find fast and reliable public WFS server to rely on.

Will backport to the LTR manually after merge.

python/plugins/processing/algs/gdal/GdalAlgorithm.py
```diff
@@ -115,6 +116,11 @@ def getOgrCompatibleSource(self, parameter_name, parameters, context, feedback,
                 #for the command line preview since it has no meaning outside of a QGIS session!
                 ogr_data_path = GdalUtils.ogrConnectionStringAndFormatFromLayer(input_layer)[0]
                 ogr_layer_name = GdalUtils.ogrLayerName(input_layer.dataProvider().dataSourceUri())
+        elif input_layer.dataProvider().name().lower() == 'wfs':
```
nyalldawson: Could you add a unit test covering this?

alexbruy: As I wrote in the previous comment, I will be happy to cover it with test. But need to find fast and reliable public WFS server (to create a valid layer). Already tried to create layer using fake connection string, but as resulting layer is not valid calls to its `dataProvider()` method returns `None`. Or I miss something and it can be done without layer creation?

nyalldawson: You could try QgsMapLayer.providerType(). It's badly named, but this will return the data provider name even when the layer is invalid/provider doesn't exist.

alexbruy: Unfortunately this does not work. While provider recognized correctly, layer still invalid and can not be added to project.

nyalldawson: Ok, let's merge anyway?

alexbruy: I would say yes, if this looks good.

nyalldawson:
+1 to merge (minus the world_map conflict)

PR #31743, opened 2019-09-13
Fix QT font database initializing (FCGI with IIS)

stefan-pascal:
When using FCGI with IIS, environment variables (QT_QPA_FONTDIR in this case) are lost after fcgi_accept().

In order to specify the directory where font files are located, one must define the "QT_QPA_FONTDIR" environment variable before running qgis_mapserv.fcgi.exe. When using FCGI with IIS, you can specify environment variables to be set before starting qgis_mapserv.fcgi.exe process, but these are lost immediately after the call to fcgi_accept.
We force a QFontDatabase initialization before fcgi_accept(), in order to allow QFontDatabase to initialize the list of fonts available in "QT_QPA_FONTDIR" (this is still accessible at this time).

### DIFF
--- a/src/server/qgis_map_serv.cpp
+++ b/src/server/qgis_map_serv.cpp
@@ -26,6 +26,7 @@
 #include <fcgi_stdio.h>
 #include <cstdlib>
 
+#include <QFontDatabase>
 #include <QString>
 
 int fcgi_accept()
@@ -71,7 +72,7 @@ int main( int argc, char *argv[] )
   // When using FCGI with IIS, environment variables (QT_QPA_FONTDIR in this case) are lost after fcgi_accept().
   QFontDatabase fontDB;
 #endif
-  
+
   // Starts FCGI loop
   while ( fcgi_accept() >= 0 )
   {
--- a/src/server/qgis_map_serv.cpp
+++ b/src/server/qgis_map_serv.cpp
@@ -65,6 +65,13 @@ int main( int argc, char *argv[] )
 #ifdef HAVE_SERVER_PYTHON_PLUGINS
   server.initPython();
 #endif
+
+#ifdef Q_OS_WIN
+  // Initialize font database before fcgi_accept.
+  // When using FCGI with IIS, environment variables (QT_QPA_FONTDIR in this case) are lost after fcgi_accept().
+  QFontDatabase fontDB;
+#endif
+  
   // Starts FCGI loop
   while ( fcgi_accept() >= 0 )
   {


elpaso:
> When using FCGI with IIS, environment variables (QT_QPA_FONTDIR in this case) are lost after fcgi_accept().
> 
> In order to specify the directory where font files are located, one must define the "QT_QPA_FONTDIR" environment variable before running qgis_mapserv.fcgi.exe. When using FCGI with IIS, you can specify environment variables to be set before starting qgis_mapserv.fcgi.exe process, but these are lost immediately after the call to fcgi_accept.

This is bad, do you have any idea why that happens?

Does that also mean that you cannot set (for example) `QGIS_PROJECT_FILE` in the environment and pass it to the server?


> We force a QFontDatabase initialization before fcgi_accept(), in order to allow QFontDatabase to initialize the list of fonts available in "QT_QPA_FONTDIR" (this is still accessible at this time).

Looks good to me.

elpaso:
btw, fix indentation, see: https://travis-ci.org/qgis/QGIS/jobs/584606457#L862

stefan-pascal:
> Does that also mean that you cannot set (for example) `QGIS_PROJECT_FILE` in the environment and pass it to the server?
I didn't check that; will look into and get back with the results.

stefan-pascal:
> > Does that also mean that you cannot set (for example) `QGIS_PROJECT_FILE` in the environment and pass it to the server?
> 
> I didn't check that; will look into and get back with the results.

You can. The difference is that `QGIS_PROJECT_FILE` is loaded and stored in `QgsServer::sSettings` before `fcgi_accept()`. When a new request is handled in the `while (fcgi_accept())` loop, if there is no `&map=C:/dir/file.qgs` param in the query string, it will use the value cached in `QgsServer::sSettings` (see `QgsServer::configPath`).

elpaso:
> > > Does that also mean that you cannot set (for example) `QGIS_PROJECT_FILE` in the environment and pass it to the server?
> > 
> > 
> > I didn't check that; will look into and get back with the results.
> 
> You can. The difference is that `QGIS_PROJECT_FILE` is loaded and stored in `QgsServer::sSettings` before `fcgi_accept()`. When a new request is handled in the `while (fcgi_accept())` loop, if there is no `&map=C:/dir/file.qgs` param in the query string, it will use the value cached in `QgsServer::sSettings` (see `QgsServer::configPath`).

Thanks for checking, but what I meant was to set `QGIS_PROJECT_FILE` per-request, it's the typical way to set the project from an URL like `http://myserver.com/project/my_project/?SERVICE=WMS ...` instead of using `MAP=path/to/my_project` in the query string (which is bad and evil and makes little kittens die).

stefan-pascal:
> > > > Does that also mean that you cannot set (for example) `QGIS_PROJECT_FILE` in the environment and pass it to the server?
> > > 
> > > 
> > > I didn't check that; will look into and get back with the results.
> > 
> > 
> > You can. The difference is that `QGIS_PROJECT_FILE` is loaded and stored in `QgsServer::sSettings` before `fcgi_accept()`. When a new request is handled in the `while (fcgi_accept())` loop, if there is no `&map=C:/dir/file.qgs` param in the query string, it will use the value cached in `QgsServer::sSettings` (see `QgsServer::configPath`).
> 
> Thanks for checking, but what I meant was to set `QGIS_PROJECT_FILE` per-request, it's the typical way to set the project from an URL like `http://myserver.com/project/my_project/?SERVICE=WMS ...` instead of using `MAP=path/to/my_project` in the query string (which is bad and evil and makes little kittens die).

I didn't know you can set the project file from something like `http://myserver.com/project/my_project/?SERVICE=WMS...`. I'm wondering what configuration would be necessary in IIS (or Apache) in order to achieve this.

elpaso:
For apache: mod_rewrite, for nginx: https://github.com/elpaso/qgis3-server-vagrant/blob/master/resources/nginx/qgis-server-fcgi#L39

elpaso:
@stefan-pascal I'd like to merge this one, but I cannot test it on windows and I would really like to know if we are missing something more general here.
To my understanding, the server process will try to get the project path when processing a request this way:

1. get it from query string `MAP` parameter
2. get it from project settings (initially set by the environment variable `QGIS_PROJECT_FILE`) when the server starts
3. if none of the above are set, it will try to get it from `sConfigFilePath` that is normally the first project stored the server directory
4. if none of the above are set, it will read it from the environment variable `QGIS_PROJECT_FILE` (which may have been changed by the web server rewrite rules)

--

4. is the typical and recommended way to pass `QGIS_PROJECT_FILE` in the environment through a rewrite rule.

--

@rldhont can you confirm this behavior, I'm not totally sure about the workflow here.

Now, you say that the environment variables are not passed from IIS to the QGIS server so I assume that almost all of the above mentioned behavior would not work correctly, that's why I suspect that your fix is only a partial fix and we should try to address the whole issue, not just for `QT_QPA_FONTDIR` and the font database.

stefan-pascal:
> Now, you say that the environment variables are not passed from IIS to the QGIS server so I assume that almost all of the above mentioned behavior would not work correctly, that's why I suspect that your fix is only a partial fix and we should try to address the whole issue, not just for `QT_QPA_FONTDIR` and the font database.

The environment variables **are passed** from IIS to the qgis_mapserv.fcgi worker, but there is a specific workflow (at least on IIS; i can not test it with Apache):
1. IIS starts the worker process (qgis_mapserv.fcgi.exe) when the first web / http request is received;
2. There is a initialization logic performed in `main()` (creating QgsApplication, QgsServer etc.), before entering `fcgi_accept`. All environment variables **are available** during this phase;
3. In the `fcgi_accept` loop, there are no more environment variables (**`getenv` on any variable will return NULL**).

The main difference between `QT_QPA_FONTDIR` and other variables is related to the fact that all other variables (`QGIS_PROJECT_FILE` included) are stored in `static QgsServerSettings sSettings` (initialized through `QgsServerSettings ::initSettings`). 
When a `server.handleRequest( request, response )` is performed, if there is no query string `MAP` parameter, the server will use the cached (loaded at startup) sSettings value. 
If you try to change the `QGIS_PROJECT_FILE` environemnt variable in IIS settings, IIS will stop / restart the worker process, thus passing through steps 1-2-3 described above.

In the case of `QT_QPA_FONTDIR` variable, it is **always** read directly from the environment by the `QFontDatabase`. By creating the static instance before step 3 described above, the environment is still accessible at this time, allowing QFontDatabase to load the fonts from the specified path (location).

elpaso:
ok, I would say that the only way to rewrite the address on IIS is then to manipulate the `MAP` query string argument (because restarting the process on any env change is a performance killer), this would break any OGC API workflow (OPENAPI WFS3 specs does not allow any query string parameter), but this is another issue.

> In the fcgi_accept loop, there are no more environment variables (getenv on any variable will return NULL).

This looks really wierd to me: (F)CGI is based on environment variables passing and QGIS server uses `getenv` all over the place, if they are not available inside the loop and return `NULL`, how can the server have access to `REQUEST_URI`, `SCRIPT_NAME`, `SERVER_NAME`, `SERVER_PORT` etc.?

StefanNiemeier:
This is not only a problem with IIS. There is also the same issue in QGIS/QGIS Server 3.8.3 on Apache on Windows12.
Love to get this work merged.

Regard,

Stefan Niemeier

elpaso (approved):

Issue #31644, opened 2019-09-09
UI: Fast close hotkey from SQL Filter


Komzpa:
I open the filter. I want to quickly change a number and be gone. I change it, try Enter, try Ctrl+Enter, nothing works, I go grab my mouse. 
![image](https://user-images.githubusercontent.com/810638/64537024-c3de8600-d322-11e9-8347-104990f68b95.png)

Ctrl+Enter can apply+save.

nirvn:
It's indeed the case that there's no way to use this dialog only with the keyboard as TABing away from the sql edit widget doesn't work (it instead inserts a tab).

nirvn:
Well, well, well, TIL: Alt+O (i.e. the letter O) is the dialog shortcut to apply the filter. Enjoy.

Komzpa:
Thank you!

PR #31758, opened 2019-09-14
[needs-docs][ui] Use ctrl+return shortcut to accept/use expression and filter text

nirvn:
## Description

Keyboard-based UX is broken with expression and SQL editors  as those are not focused away when hitting the Tab key. 

To fix this, this commit uses the Ctrl+Enter key sequence shortcut in the following dialogs to enable keyboard interaction:
- In the query builder (i.e. layer SQL filter), ctrl+enter will apply filter
- In the expression builder, ctrl+enter will apply the expression
- In the select by expression dialog, ctrl+enter applies the selection

This fixes the usability issue reported @ #31644. 

## Checklist


nirvn:
... Alt+O ... :scream: closing.

PR #31763, opened 2019-09-14
use PostgreSQL quote_ident() function to quote schema and table names when listing tables

alexbruy:
## Description
Found this one when working with 3rd party database with quite unusual table names like `""seksyen""`.

If table or schema name contains quotes (or even multiple quotes/other special characters) QGIS fails to get list of available tables in the PostgreSQL database. Corresponding query fails with syntax error. Only core provider affected, DB Manager works fine as it uses different queries.

Using `quote_ident()` funciton instead of manual quoting fixes this issue.

## Checklist


nyalldawson:
LGTM. @elpaso?

nyalldawson (approved):

PR #31765, opened 2019-09-14
[WFS provider] Re-inject custom query parameters in DCP HTTP URLs (fixes #31026)

### LINKED ISSUE #31026: QGIS WFS stripping of viewparam KVP from WFS Feature Requests
**Describe the bug**
For our WFS service on https://data.linz.govt.nz we use the viewparams KVP request option to provide a data changesets extension over the top of WFS 1.1 and 2.0. This viewparam option is supported in Geoserver https://docs.geoserver.org/stable/en/user/data/database/sqlview.html. This support is for all of our vector layers where we have a primary key defined and can be URLs can be generated when the user is logged in. e.g https://data.linz.govt.nz/layer/50772-nz-primary-parcels/history/

An example URL is: 

https://data.linz.govt.nz/services;key=42eb71dd76a84b3d8b2e69a7b96a4d57/wfs/layer-50772-changeset?SERVICE=WFS&REQUEST=GetCapabilities&viewparams=from:2019-07-20T06:17:24.767945Z;to:2019-07-27T00:31:26.575136Z

In QGIS 2.18 the WFS provider will load this URL ok and fetch the data. However under QGIS 3.4 and 3.8 for each GetFeature request the viewparams KVP is stripped off and the query no longer work (returns 0 features). e.g:

http://data.linz.govt.nz/services;key=42eb71dd76a84b3d8b2e69a7b96a4d57/wfs/layer-50772-changeset?SERVICE=WFS&REQUEST=GetFeature&VERSION=2.0.0&TYPENAMES=data.linz.govt.nz:layer-50772-changeset&TYPENAME=data.linz.govt.nz:layer-50772-changeset&STARTINDEX=0&COUNT=12000000&SRSNAME=urn:ogc:def:crs:EPSG::4167&BBOX=-52.47523994920035761,164.42657957612391328,-27.91622356177029474,185.88148236811673542,urn:ogc:def:crs:EPSG::4167

I know the viewparams is not part of the standard, but it certainly was great that QGIS supported it. 

Is this WFS provider change intensional or a side effect of some other WFS provider changes? In any case can this functionality please be added back?

**QGIS and OS versions**

Tested to be an issue in all OS versions with QGIS 3.4.3 and 3.8.1

**Additional context**

See old closed issue #27339 and https://gis.stackexchange.com/questions/317573/request-wfs-data-using-geoservers-viewparams-with-pyqgis

rouault:
Since f1b59876344bf0c0f066c81bf8ef73ef25365023, we honour DCP HTTP endpoints
exposed in GetCapabilities.
But some (all?) servers fail to add in those endpoints the custom query parameters
the user might have injected in the GetCapabilities URL.
So re-add them manually.

### DIFF
--- a/src/providers/wfs/qgswfsdatasourceuri.cpp
+++ b/src/providers/wfs/qgswfsdatasourceuri.cpp
@@ -19,6 +19,8 @@
 #include "qgswfsdatasourceuri.h"
 #include "qgsmessagelog.h"
 
+#include <QUrlQuery>
+
 QgsWFSDataSourceURI::QgsWFSDataSourceURI( const QString &uri )
   : mURI( uri )
 {
@@ -45,9 +47,8 @@ QgsWFSDataSourceURI::QgsWFSDataSourceURI( const QString &uri )
 
     QUrl url( uri );
     // Transform all param keys to lowercase
-    QList<queryItem> items( url.queryItems() );
-    const auto constItems = items;
-    for ( const queryItem &item : constItems )
+    const auto items( url.queryItems() );
+    for ( const queryItem &item : items )
     {
       url.removeQueryItem( item.first );
       url.addQueryItem( item.first.toLower(), item.second );
@@ -106,9 +107,8 @@ QgsWFSDataSourceURI::QgsWFSDataSourceURI( const QString &uri )
     do
     {
       somethingChanged = false;
-      QList<queryItem> items( url.queryItems() );
-      const auto constItems = items;
-      for ( const queryItem &item : constItems )
+      const auto items( url.queryItems() );
+      for ( const queryItem &item : items )
       {
         const QString lowerName( item.first.toLower() );
         if ( lowerName == QgsWFSConstants::URI_PARAM_OUTPUTFORMAT )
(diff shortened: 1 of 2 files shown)


elpaso:
@rouault the implementation LGTM, but I have some more generic doubts: I share your concerns https://github.com/qgis/QGIS/pull/31765/files#diff-01dca2db594133fa0b756eb887f408f0R201
What I'm afraid of is the possible leak of credentials and auth tokens in case a QGIS authentication plugin injects the credentials in the URL, in theory this should not happen because we would have stored the `authcfg` in the data source URI and it's the `QgsNetworkAccessManager` who transforms the authfg into the actual credentials, but worth checking that it does not really happen.

rouault:
> but worth checking that it does not really happen.

I don't have test servers which require authentication and associated credentials. I just tested a public server and used the basic authentication scheme, and I can't see anything leaking in the URLs sent to the server

elpaso:
That's because HTTP basic doesn't use any token injected in the URL (it only changes the headers), some oauth servers might do that.
Btw, I think that's safe to assume that this patch does not change anything in that regard, feel free to merge.

Issue #30937, opened 2019-07-26
Crash: right click on image ->export ->save as->"save raster layer as"


EvanJHamilton:
QGIS crashes when exporting large images (12GB compressed satellite imagery).

Crash ID: de13be2a3f01cec4b12e9b0211da1bf5ded28dd7

Stack Trace

QgsRasterFileWriter::writeImageRaster :
QgsRasterFileWriter::writeRaster :
QgsRasterFileWriterTask::run :
QgsTask::start :
QThreadPoolPrivate::reset :
QThread::start :
BaseThreadInitThunk :
RtlUserThreadStart :

QGIS Info
QGIS Version: 3.8.1-Zanzibar
QGIS code revision: dcd95cc648
Compiled against Qt: 5.11.2
Running against Qt: 5.11.2
Compiled against GDAL: 2.4.1
Running against GDAL: 2.4.1

System Info
CPU Type: x86_64
Kernel Type: winnt
Kernel Version: 6.1.7601

gioman:
Any chance of a download link for such raster layer?

gioman:
Closing for lack of feedback. Please reopen if necessary.

PR #31766, opened 2019-09-14
QgsRasterFileWriter: assorted set of fixes (might be related to refs #30937)

rouault:
- Lack of error checking on destination provider creation (in saveAsImage mode)
- Wrong computation of number of blocks if dimension is exactly a multiple of
  the block size
- Lack of error checking when writing blocks
- Slow computation of non-premultiplied r,g,b values, and inappropriate use
  of memcpy()

### DIFF
--- a/tests/src/python/test_qgsrasterfilewriter.py
+++ b/tests/src/python/test_qgsrasterfilewriter.py
@@ -18,7 +18,8 @@ import tempfile
 
 from osgeo import gdal
 from qgis.PyQt.QtCore import QTemporaryFile, QDir
-from qgis.core import (QgsRaster,
+from qgis.core import (QgsContrastEnhancement,
+                       QgsRaster,
                        QgsRasterLayer,
                        QgsRasterChecker,
                        QgsRasterPipe,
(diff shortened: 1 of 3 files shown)


rouault:
> https://doc.qt.io/qt-5/qcolor.html#qUnpremultiply might be more appropriate to use here

done

nyalldawson (approved):

tests/src/python/test_qgsrasterfilewriter.py
```diff
@@ -223,6 +226,62 @@ def testGeneratePyramidsInternal(self):
...
+
+        pipe = QgsRasterPipe()
+        self.assertTrue(pipe.set(provider.clone()))
+
+        self.assertEqual(fw.writeRaster(pipe,
+                                        provider.xSize(),
+                                        provider.ySize(),
+                                        provider.extent(),
+                                        provider.crs()), QgsRasterFileWriter.DestProviderError)
+        del fw
```
elpaso: Is there a reason for calling `del` explicitly?

rouault: hum, copy&paste from previously existing tests. I assume that the original intent was to make sure the target file was properly closed before re-reading it, but AFAICS in QgsRasterFileWriter there's no variable member with the output dataset, so probably useless

PR #31770, opened 2019-09-14
[bugfix] remove label input in layer expression

### LINKED ISSUE #31761: [layout] Unexpected result when adding expression to legend items label
The attached screenshot shows the same expression applied to a single symbol rendered "buildings" layer and a categorized "Restaurants" layer. I think something is no coherent in the result but  can't tell since I'm only trying to understand the logic behind the tool.
image

Long story:
1. Add a legend item to a map layout
1. In the 'Legend items' section, select a layer (I tested with a single symbol rendered layer named "buildings") 
1. Click the "Add an expression to the vector layer and each child symbol's label" button
The expression dialog opens with a variable `@symbol_label` already entered. 
If I OK the expression dialog with the following expression in the field 
   ```
   'test ' || @symbol_label
   ```
I'd expect the legend to display `test buildings` and not `buildingstest buildings` (as shown in the item properties) nor `buildings` (as shown on the map). 

If the symbol_label has to be at the beginning (which is a bit restrictive imho), then I think the button label should use "Append" and not "Add" AND the expression dialog should not be prefilled with the variable.

Moreover, when applying the same expression to a categorized layer, the categories are correctly rendered, BUT the layer item keeps its original label. If the layer label is not supposed to change maybe should we review the button label (it currently implies it changes layer name AND symbol labels though it'd be only symbol labels)

master 3.9 cdd3a55e9e

roya0045:
## Description
Should fix #31761 

## Checklist

### DIFF
--- a/src/core/layout/qgslayoutitemlegend.cpp
+++ b/src/core/layout/qgslayoutitemlegend.cpp
@@ -947,7 +947,7 @@ QVariant QgsLegendModel::data( const QModelIndex &index, int role ) const
         }
       }
       else if ( QgsSymbolLegendNode *symnode = qobject_cast<QgsSymbolLegendNode *>( legendnodes.first() ) )
-        name = symnode->evaluateLabel( expressionContext, name );
+        name = symnode->evaluateLabel( expressionContext );
     }
     return name;
   }


roya0045:
Can anyone merge this, the change is quite straighforward.

DelazJ:
I gave it a try and indeed, the legend items frame shows the expected result from the entered expression and is displayed in the legend. Looks good to me, then.
Thanks @roya0045

roya0045:
> I gave it a try and indeed, the legend items frame shows the expected result from the entered expression and is displayed in the legend. Looks good to me, then.
> Thanks @roya0045

No need to thank me, I was the cause of the error :p 
The error was intentional at first but I'll reserve the alternate behaviour for other cases (API and other rare calls to add a custom label or expression along with the symbol expression.

roya0045:
@nyalldawson could you please merge this. Unless you think there are something wrong with this fix?

nyalldawson:
Thanks @roya0045!

Issue #31579, opened 2019-09-05
UI: Classify on Categorized adds new colors of some colors


Komzpa:
I have classified layer:

![image](https://user-images.githubusercontent.com/810638/64355878-a2238d00-d00a-11e9-99c9-93812ac48b1d.png)

I added classes in underlying data source. I push once again. It fills them, and all new ones are grey.

![image](https://user-images.githubusercontent.com/810638/64355997-cbdcb400-d00a-11e9-8b8c-65b18611fda6.png)

They should be of different colors after that.

roya0045:
Simply change the ramp. I don't think this is a bug. The ramp is applied to existing element, when new elements are added simply reapply the ramp. Otherwise always recomputing the ramp whenever something is added can be done but in some cases you want to distinguish your added element from classified one. By default when you add a new element the default style is applied.

Komzpa:
I expect that in "classified" mode with a selected ramp the values added by pushing "classify" button are getting classified. In above scenario, they are not.

NathanW2:
@Komzpa hmm I agree it should use the ramp and generate new colors when you press classify again.  I thought it did but seems not. That is a valid UX issue as the first click is different too the second click might throw people off.

nirvn:
Click on the random color -> select 'shuffle random colors', it'll re-colorize all category. The reason why there are no random colors added when appending values to pre-existing list is to avoid losing colors of already defined categories. If you are fine with losing that, go ahead and shuffle away :)

Komzpa:
@nirvn what is the problem to add the colors randomly without losing the existing assigned ones?

nirvn:
@Komzpa , on second thoughts: https://github.com/qgis/QGIS/pull/31774 , that'll fix the scenario you raised.

NathanW2:
Thanks @nirvn

PR #31774, opened 2019-09-15
[categorized renderer] When color ramp is set to random, randomize color of appended categories to pre-existing list when hitting the classify button

nirvn:
## Description

This PR address issue #31579, whereas classifying a categorized renderer on a pre-existing list  (i.e. appending new values) would not assign random colors to the appended items when the color ramp was set to random colors.

## Checklist


src/gui/symbology/qgscategorizedsymbolrendererwidget.cpp
```diff
@@ -831,6 +831,9 @@ void QgsCategorizedSymbolRendererWidget::addCategories()
   {
     QgsCategoryList prevCats = mRenderer->categories();
     keepExistingColors = !prevCats.isEmpty();
+    QgsRandomColorRamp *randomColors = new QgsRandomColorRamp();
```
nyalldawson: leaky - you should be able to avoid the heap creation here entirely and just create on the stack instead...

PR #31784, opened 2019-09-15
[Plugin Installer] Only force http icon protocol for http-based repositories

### LINKED ISSUE #31783: Plugin Manager looks for icon in http://
Plugin installer always adds http:// scheme to icon paths for not installed plugins - even if that repository is accessed with another protocol (e.g. file:// or just https:// )
See https://lists.osgeo.org/pipermail/qgis-developer/2019-September/058562.html

borysiasty:
Fixes #31783

### DIFF
--- a/python/pyplugin_installer/installer_data.py
+++ b/python/pyplugin_installer/installer_data.py
@@ -409,7 +409,9 @@ class Repositories(QObject):
                         trusted = True
                     icon = pluginNodes.item(i).firstChildElement("icon").text().strip()
                     if icon and not icon.startswith("http"):
-                        icon = "http://{}/{}".format(QUrl(self.mRepositories[reposName]["url"]).host(), icon)
+                        url = QUrl(self.mRepositories[reposName]["url"])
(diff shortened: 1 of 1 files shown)


python/pyplugin_installer/installer_data.py
```diff
@@ -409,7 +409,10 @@ def xmlDownloaded(self):
                         trusted = True
                     icon = pluginNodes.item(i).firstChildElement("icon").text().strip()
                     if icon and not icon.startswith("http"):
-                        icon = "http://{}/{}".format(QUrl(self.mRepositories[reposName]["url"]).host(), icon)
+                        if QUrl(self.mRepositories[reposName]["url"]).scheme() == 'http':
```
nyalldawson:
```suggestion
                        url = QUrl(self.mRepositories[reposName]["url"])
                        if url.scheme() in ('http', 'https'):
                            icon = "{}://{}/{}".format(url.scheme(), url.host(), icon)
```

borysiasty: Good catch, thanks!

PR #31790, opened 2019-09-16
remove duplicated code with single method for QgsVectorLayer::min and max values

3nids:


src/core/qgsvectorlayer.cpp
```diff
@@ -3867,6 +3867,16 @@ QStringList QgsVectorLayer::uniqueStringsMatching( int index, const QString &sub
...
+{
+  return minOrMax( index, true );
+}
+
+QVariant QgsVectorLayer::maximumValue( int index ) const
+{
+  return minOrMax( index, false );
+}
+
+QVariant QgsVectorLayer::minOrMax( int index, bool minimum ) const
```
nirvn: maybe rename this function to be a bit more explicit (even though it's not part of the public API)? minimumOrMaximumValue(...) ?

m-kuhn (approved):

PR #31793, opened 2019-09-16
use a single comment checking unittest in DB Manager SQL window

Gustry:
## Description

use a single unittest in DB Manager SQL window

## Checklist

### DIFF
--- a/tests/src/python/test_db_manager_sql_window.py
+++ b/tests/src/python/test_db_manager_sql_window.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 """QGIS Unit tests for the DBManager SQL Window
 
+From build dir, run: ctest -R PyQgsDBManagerSQLWindow -V
+
 .. note:: This program is free software; you can redistribute it and/or modify
 it under the terms of the GNU General Public License as published by
 the Free Software Foundation; either version 2 of the License, or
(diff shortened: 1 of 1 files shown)


m-kuhn:
Can you elaborate on the advantages of a monolithic test?

Gustry:
Having a single line in PyCharm when you run the tests ;-)
All these tests are using a single function `check_comments_in_sql` without using the GUI, doing only some regexp.

Ref https://github.com/qgis/QGIS/commit/a39a413b92e05d04307122e9164ef263ac3b9a49

tests/src/python/test_db_manager_sql_window.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
```
m-kuhn: Was this removed on purpose?

Gustry: We are on Python3, so should we keep it?

m-kuhn: No idea, a quick internet search reveals that it might indeed be save to remove them.

In order to not mix topics: can you open a separate pull request where you remove all of them and where we can discuss about the implications?

Gustry: Ok, I have reverted

m-kuhn:
One final question, if one of these tests fails, what's the message?

For individual tests like before, the method name would point to which scenario is failing, is there still something meaningful (apart from the line number) now?

nyalldawson:
Not suitable for backport

rldhont:
Sorry @nyalldawson, I was thinking that unittest has to be backported.

m-kuhn:
It's not a new unit test. It's "just" surface polishing of an existing one.

Gustry:
This file doesn't exist in 3.4 anyway ;-)

@m-kuhn It's the normal error message with the line number indeed.
Usually a tested function is grouped in a single method `test_xxxx`.

I was working on this file when I was investigating about https://github.com/qgis/QGIS/pull/31746. I got a `git stash` on my laptop for a test.

m-kuhn:
@Gustry For me normally unit tests are written to make debugging as easy as possible if they fail. If they test a single function or not is secondary.

Let's say I do a change to the parser and I can see that test 1 and 3 fails, that's additional information that we have a problem with the `--` not being at the start of the line. And that's more information than I get from a single monolithic failing test.

We can merge this, but I'd not make it a precedent.

Gustry:
Ok, I understand @m-kuhn.
Thanks for the merge. We agree, it was small polishing ;-)

PR #31804, opened 2019-09-16
Visual cleanup for qgsProject in pyqgis doc

DelazJ:
Trying to address some issues visible at https://qgis.org/pyqgis/master/core/QgsProject.html:
1. a note that is not displayed as such
1. `/deprecated` information is weirdly replaced, ie `* \deprecated since QGIS 3.4 use QgsMapLayer::setFlags() instead` becomes `Deprecated since version since: QGIS 3.4 use QgsMapLayer.setFlags() instead`

The build will fail because no sip changes have been done yet. This PR is opened in other to get feedback whether I correctly understand the issues above:
1. Is that the proper way to address the note and are all the information part of the note?
1. What actually replaces /deprecated and where should begin the second sentence in the information (at "since" or at "use"?)

### DIFF
--- a/src/core/qgsproject.h
+++ b/src/core/qgsproject.h
@@ -77,14 +77,15 @@ class QgsBookmarkManager;
  * \ingroup core
  * Reads and writes project states.
  *
-  \note
-
-  Has two general kinds of state to make persistent.  (I.e., to read and
-  write.)  First, QGIS proprietary information.  Second plug-in information.
-
-  A singleton since there shall only be one active project at a time; and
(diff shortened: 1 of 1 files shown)


DelazJ:
It passes 🙄 ?

3nids:
This is definitely weird that it passes, but regarding the issue with since and deprecated, I believe that it doesn't like to have both since and deprecated. Maybe by putting since first (makes more sense anyway).

nyalldawson:
Superseded by #32104

PR #31805, opened 2019-09-16
[locator] fix sorting of grouped results

3nids:
backport of #31797


m-kuhn:
The test was added later?

nyalldawson:
Please rethink whether a backport of this is appropriate. I personally think at this stage 3.4 should be entirely "hands off" apart from absolutely mission critical fixes, and any "paper cut" type fixes like this should NOT be backported.

3nids:
I understand your concerns but I am currently developing a locator filter plugin for 3.4 and this is required to work.

The commit log only shows single (although 3 commits) addition to the locator being the context menu, so in the case of the model basically just the data method handling another role.
I also backported the test.

So, I'd like to get this merged as this is required and I believe is safe.

nyalldawson (approved):

PR #31750, opened 2019-09-13
Processing editing in place: handle constraints

elpaso:
Fixes #31634

PR #31810, opened 2019-09-16
Bugfix gh31634 processing in place unique constraints 3 4 backport

elpaso:
Semi-manual backport of PR #31750


nyalldawson (approved):

nyalldawson:
@elpaso While I approved this based on the changes alone, I'm actually a -1 to backporting fixes like this to 3.4. I think at this stage 3.4 should be entirely "hands off" apart from absolutely mission critical fixes.

elpaso:
@nyalldawson I had the same feeling when I saw that there was a missing API (createFeatures) but at that point the solution was evident to me and I thought it was a shame to throw away the time spent on it. 

But I'll take no offense if this is reverted.

PR #31822, opened 2019-09-17
Ensure atlas expression evaluation has access to coverage layer scope

### LINKED ISSUE #31807: is_selected() not working as filter in print composer
**Describe the bug**

Using QGIS 3.8.2 adding `is_selected()` as filter in the print composer returns an error 

**How to Reproduce**

1. select some features
2. add a print composer
3. enable atlas function
4. set the map-item-properties as to be controlled by the atlas
5. define the filter `is_selected()`

message: > no matching atlas features found

When trying to use the same expression in the attribute table, using the 'Advanced Filter' with the same expression `is_selected()`, it works fine.  Idem dito with the rule based styling of the layer only

In QGIS 2.18 you could use an expression (provided by a plugin): `isselected('layername')` which worked fine.

nyalldawson:
Fixes #31807

### DIFF
--- a/src/core/layout/qgslayoutatlas.cpp
+++ b/src/core/layout/qgslayoutatlas.cpp
@@ -439,7 +439,7 @@ QString QgsLayoutAtlas::currentFilename() const
   return mCurrentFilename;
 }
 
-QgsExpressionContext QgsLayoutAtlas::createExpressionContext()
+QgsExpressionContext QgsLayoutAtlas::createExpressionContext() const
 {
   QgsExpressionContext expressionContext;
   expressionContext << QgsExpressionContextUtils::globalScope();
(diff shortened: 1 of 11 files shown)


src/core/layout/qgslayoutatlas.cpp
```diff
@@ -439,18 +439,18 @@ QString QgsLayoutAtlas::currentFilename() const
...
+QgsExpressionContext QgsLayoutAtlas::createExpressionContext() const
 {
   QgsExpressionContext expressionContext;
   expressionContext << QgsExpressionContextUtils::globalScope();
   if ( mLayout )
     expressionContext << QgsExpressionContextUtils::projectScope( mLayout->project() )
                       << QgsExpressionContextUtils::layoutScope( mLayout );
 
-  expressionContext.appendScope( QgsExpressionContextUtils::atlasScope( this ) );
+  expressionContext.appendScope( QgsExpressionContextUtils::atlasScope( const_cast< QgsLayoutAtlas * >( this ) ) );
```
m-kuhn: Any reason not to const-change the signature `QgsExpressionContextUtils::atlasScope`?

nyalldawson: I tried originally, but there's non const getters used in atlasScope and the changes started snowballing...

nyalldawson: Actually it wasn't that bad in the end!

PR #31828, opened 2019-09-17
Layout map item grids

DelazJ:
Before
image

Because the interval x/y min/max elements are put under a stackedwidget, they are not aware of the alignment of the other features in the GUI. The PR removes the stackedwidget.
After

image

The fix works. Now is it the right/best way? And not sure also about the doc and methods/parameters naming.
Also @nyalldawson is the no margin in the widget deliberate? I can see it's also the case when setting the map item labeling properties

## Checklist

### DIFF
--- a/src/app/layout/qgslayoutmapgridwidget.cpp
+++ b/src/app/layout/qgslayoutmapgridwidget.cpp
@@ -640,11 +640,11 @@ void QgsLayoutMapGridWidget::setGridItems()
     case QgsLayoutItemMapGrid::MapUnit:
     case QgsLayoutItemMapGrid::MM:
     case QgsLayoutItemMapGrid::CM:
-      mIntervalStackedWidget->setCurrentIndex( 0 );
+      showFixedIntervalWidth( true );
       break;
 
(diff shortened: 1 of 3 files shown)


nyalldawson:
This was originally done to indicate that the x/y settings apply to the interval -- which is no longer immediately apparent.

nirvn:
@nyalldawson , isn't the 'interval [ map unit ]' line a bit misleading since I assume offsets also use the unit type set there?

I propose:

```
Unit type ______ [ Map unit ]
                           [ X 0.000   ]
Interval
                           [ Y 0.000   ]
                           [ X 0.000   ]
Offset
                           [ Y 0.000   ]
```

DelazJ:
>This was originally done to indicate that the x/y settings apply to the interval -- which is no longer immediately apparent.

Indeed, I did not remark the "indentation"; I was more focused on the spinbox size & alignment. Adding a horizontal spacer at the left of the interval subitems could do the trick, no?

DelazJ:
@nirvn the dialog provides also this:
image

DelazJ:
Edited the screenshot with the offset

DelazJ:
Gone stale... 🙄

PR #31843, opened 2019-09-17
[layer tree] move logic from QgisApp to QgsLayerTreeView to determine insertion point

3nids:
and create bindings to get the current insertion point

### DIFF
--- a/src/gui/layertree/qgslayertreeview.h
+++ b/src/gui/layertree/qgslayertreeview.h
@@ -169,6 +169,29 @@ class GUI_EXPORT QgsLayerTreeView : public QTreeView
      */
     int layerMarkWidth() const { return mLayerMarkWidth; }
 
+    /**
+     * Returns the insertion group with its insertion index
+     * \since QGIS 3.10
+     */
+    QPair<QgsLayerTreeGroup *, int> insertionPoint() const SIP_SKIP;
+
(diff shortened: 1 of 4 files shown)


3nids:
@wonder-sk aka LayerTree's grandpa, would you like to have a look?

src/gui/layertree/qgslayertreeview.h
```diff
@@ -169,6 +169,29 @@ class GUI_EXPORT QgsLayerTreeView : public QTreeView
      */
     int layerMarkWidth() const { return mLayerMarkWidth; }
 
+    /**
+     * Returns the insertion group with its insertion index
+     * \since QGIS 3.10
+     */
+    QPair<QgsLayerTreeGroup *, int> insertionPoint() const SIP_SKIP;
```
nyalldawson: Can we avoid the pair and return a struct instead?

3nids: actually I believe this relies in app in Qgis(App)Interface. It makes more sense to me since this logic can be app-dependent (with the logic for embedded layers for instance). I have another PR coming.

nyalldawson: Ok, fair enough

3nids:
superseded by #31859

PR #31848, opened 2019-09-17
Fix crash in QgsCoordinateTransformPrivate at application exit (fixes #31762)

rouault:
This fixes the following Valgrind reported error:
```
==1703== Invalid read of size 4
==1703==    at 0xE614195: internal_pj_ctx_get_errno (in /home/even/proj/install-proj-master/lib/libproj.so.15.2.0)
==1703==    by 0xE6116AE: internal_pj_free (in /home/even/proj/install-proj-master/lib/libproj.so.15.2.0)
==1703==    by 0xE624E78: internal_proj_destroy (in /home/even/proj/install-proj-master/lib/libproj.so.15.2.0)
==1703==    by 0x7D3EFE7: QgsCoordinateTransformPrivate::freeProj() (qgscoordinatetransform_p.cpp:659)
==1703==    by 0x7D3CD23: QgsCoordinateTransformPrivate::~QgsCoordinateTransformPrivate() (qgscoordinatetransform_p.cpp:120)
==1703==    by 0x7D3AC46: QExplicitlySharedDataPointer<QgsCoordinateTransformPrivate>::~QExplicitlySharedDataPointer() (qshareddata.h:165)
==1703==    by 0x7D36D63: QgsCoordinateTransform::~QgsCoordinateTransform() (qgscoordinatetransform.cpp:148)
==1703==    by 0x7D3C425: QHashNode<QPair<QString, QString>, QgsCoordinateTransform>::~QHashNode() (qhash.h:149)
==1703==    by 0x7D3C459: QHash<QPair<QString, QString>, QgsCoordinateTransform>::deleteNode2(QHashData::Node*) (qhash.h:536)
==1703==    by 0x9BEEB78: QHashData::free_helper(void (*)(QHashData::Node*)) (in /opt/qt59/lib/libQt5Core.so.5.9.1)
==1703==    by 0x7D3B81C: QHash<QPair<QString, QString>, QgsCoordinateTransform>::freeData(QHashData*) (qhash.h:576)
==1703==    by 0x7D3B2B3: QHash<QPair<QString, QString>, QgsCoordinateTransform>::~QHash() (qhash.h:254)
==1703==    by 0x7D3B7DA: QHash<QPair<QString, QString>, QgsCoordinateTransform>::operator=(QHash<QPair<QString, QString>, QgsCoordinateTransform>&&) (qhash.h:260)
==1703==    by 0x7D3B239: QHash<QPair<QString, QString>, QgsCoordinateTransform>::clear() (qhash.h:582)
[...]

Issue #31762, opened 2019-09-14
QGIS master crash upon exit @ QgsCoordinateTransformPrivate::freeProj on Windows platform


nirvn:
**Describe the bug**

While we have a fair number ( :cry: ) of "QGIS crash upon exit" for 3.4 and 3.8 (both running proj < 6), this issue covers a crash upon exit for the master builds (i.e. 3.9) on windows, where it also dies with proj >= 6.

My system is now able to duplicate the crasher every time I exit QGIS with a project loaded, and I've been able to install the debug symbol needed for a proper stack to be outputted. 

Without further due:

```
osgeo::proj::datum::Datum::conventionalRS :
osgeo::proj::datum::Datum::conventionalRS :
osgeo::proj::datum::Datum::conventionalRS :
QgsCoordinateTransformPrivate::freeProj qgscoordinatetransform_p.cpp:664
QgsCoordinateTransformPrivate::~QgsCoordinateTransformPrivate qgscoordinatetransform_p.cpp:120
QgsCoordinateTransformPrivate::`scalar deleting destructor' :
QExplicitlySharedDataPointer::~QExplicitlySharedDataPointer qshareddata.h:169
QgsCoordinateTransform::~QgsCoordinateTransform qgscoordinatetransform.cpp:148
QHashNode,QgsCoordinateTransform>::~QHashNode,QgsCoordinateTransform> :
QHashNode,QgsCoordinateTransform>::`scalar deleting destructor' :
QHash,QgsCoordinateTransform>::deleteNode2 qhash.h:547
QHashData::free_helper :
QHash,QgsCoordinateTransform>::freeData qhash.h:586
QHash,QgsCoordinateTransform>::~QHash,QgsCoordinateTransform> qhash.h:254
QHash,QgsCoordinateTransform>::operator= qhash.h:260
QHash,QgsCoordinateTransform>::clear qhash.h:591
QgsCoordinateTransform::invalidateCache qgscoordinatetransform.cpp:958
QgsApplication::exitQgis qgsapplication.cpp:1274
QgisApp::~QgisApp qgisapp.cpp:1662
QgisApp::`vector deleting destructor' :
main main.cpp:1579
WinMain mainwin.cpp:170
__scrt_common_main_seh exe_common.inl:253
BaseThreadInitThunk :
RtlUserThreadStart :
```

[...]

PR #31857, opened 2019-09-17
QgsCoordinateTransform: remove entries from cache when a thread exits (relates to #31762)

rouault:
The first commit comes from https://github.com/qgis/QGIS/pull/31848
This PR is for the second commit "QgsCoordinateTransform: remove entries from cache when a thread exits (relates to #31762)"

### DIFF
--- a/src/core/qgscoordinatetransform.cpp
+++ b/src/core/qgscoordinatetransform.cpp
@@ -958,6 +958,31 @@ void QgsCoordinateTransform::invalidateCache( bool disableCache )
   sTransforms.clear();
 }
 
+#if PROJ_VERSION_MAJOR>=6
+void QgsCoordinateTransform::removeFromCacheObjectsBelongingToCurrentThread( void *pj_context )
+{
+  // Not completely sure about object order destruction after main() has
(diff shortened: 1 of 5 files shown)


nyalldawson:
@rouault any chance of a proj < 6 version too? I think given the severity of this issue it'd be justified...

nirvn:
@rouault , great news, the crash is gone (on linux). I'll test on windows as soon as a nightly build is pushed which includes these fixes. I'm highly optimistic the crash on exit is a thing of the best. Yay!

rouault:
> any chance of a proj < 6 version too? I think given the severity of this issue it'd be justified...

- The crasher fix is the first commit, and I've done PROJ < 6 and PROJ >= 6 code paths
- The second commit which is PROJ >= 6 doesn't fix crash, but a probably not so severe memory leak.

nyalldawson:
Ok, fair enough! The sooner we make proj 6 a hard requirement the better anyway...

nirvn:
Spent one day with a build containing this commit, nothing bad to report. Let's merge.

Issue #26186, opened 2018-03-01
PostGIS Data Loads Extemely Slow


qgib:
Author Name: **Mickey Dietrich** (Mickey Dietrich)
Original Redmine Issue: [18295](https://issues.qgis.org/issues/18295)
Affected QGIS version: 3.4.5
Redmine category:data_provider/postgis

---
I have added data from PostGIS in Version 2.18 and the one data set loads in about 2 seconds. With the new 3.0 version, it takes 2 minutes to load that same data layer.  All my data sets load really slow in Version 3.0.  Not sure what the difference is from 2.18 to 3.0, but it has really caused problems with using the newer version, since we utilize the PostGIS database for accessing most all of our data.  Is there something I am missing to make it load faster like in 2.18 or is there something behind the scenes causing this problem?  Thanks, Mickey

strk:
Regis: do you mean `qgis_editor_widget_styles` contains records for many tables across 170 schemas ?
Looking at the code dealing with that table I see many optimization opportunities, like avoiding querying it for every single table, but also avoiding the count(*)...

BTW, the code doesn't schema-qualify those tables (both `qgis_editor_widget_styles` and `layer_styles`) which I'm not sure is a good idea (and surely does NOT use an index, which is only defined on the tuple schema,table.

strk:
I've just run a test creating the qgis_editor_widget_styles and populating it with 180224 records.
The `count(*)` query for such table takes 1.012 ms for me, and it is followed by a query for each attribute of the table I add to the canvas. In my case I'm adding a PostGIS topology edge_data table, which has 9 fields, this results in 9 queries to the `qgis_editor_widget_styles` table, the slower of which takes `0.193 ms`. This sums up to a worst case of under 3ms in total. QGIS logs are within the 2ms range:
```
2019-09-17 19:43:26.613 CEST [4361] strk@strk LOG:  duration: 1.012 ms  statement: SELECT COUNT(*) FROM information_schema.tables WHERE table_name=
'qgis_editor_widget_styles'
2019-09-17 19:43:26.614 CEST [4361] strk@strk LOG:  duration: 0.193 ms  statement: SELECT type, config FROM qgis_editor_widget_styles WHERE schema_name = 'linz01' and table_name = 'edge_data' and field_name = 'edge_id' LIMIT 1
[...]

PR #31858, opened 2019-09-17
Improve handling of pgsql-stored layer styles and field widgets

strk:
See #26186

### DIFF
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -1132,7 +1132,8 @@ void QgsPostgresProvider::setEditorWidgets()
   }
 
   QStringList quotedFnames;
-  for ( const QString &name : mAttributeFields.names() )
+  const QStringList fieldNames = mAttributeFields.names();
+  for ( const QString &name : fieldNames )
   {
     quotedFnames << quotedValue( name );
   }
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -66,8 +66,8 @@ inline qint32 FID2PKINT( qint64 x )
 
 static bool tableExists( QgsPostgresConn &conn, const QString &name )
 {
-  QgsPostgresResult res( conn.PQexec( "SELECT COUNT(*) FROM information_schema.tables WHERE table_name=" + QgsPostgresConn::quotedValue( name ) ) );
-  return res.PQgetvalue( 0, 0 ).toInt() > 0;
+  QgsPostgresResult res( conn.PQexec( "SELECT EXISTS ( SELECT oid FROM pg_catalog.pg_class WHERE relname=" + QgsPostgresConn::quotedValue( name ) + ")" ) );
+  return res.PQgetvalue( 0, 0 ).startsWith( 't' );
 }
 
 QgsPostgresPrimaryKeyType
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -1126,36 +1126,58 @@ bool QgsPostgresProvider::loadFields()
 
 void QgsPostgresProvider::setEditorWidgets()
 {
-  if ( tableExists( *connectionRO(), EDITOR_WIDGET_STYLES_TABLE ) )
+  if ( ! tableExists( *connectionRO(), EDITOR_WIDGET_STYLES_TABLE ) )
   {
-    for ( int i = 0; i < mAttributeFields.count(); ++i )
-    {
-      // CREATE TABLE qgis_editor_widget_styles (schema_name TEXT NOT NULL, table_name TEXT NOT NULL, field_name TEXT NOT NULL,
-      //                                         type TEXT NOT NULL, config TEXT,
-      //                                         PRIMARY KEY(schema_name, table_name, field_name));
-      QgsField &field = mAttributeFields[i];
-      const QString sql = QStringLiteral( "SELECT type, config FROM %1 WHERE schema_name = %2 and table_name = %3 and field_name = %4 LIMIT 1" ).
-                          arg( EDITOR_WIDGET_STYLES_TABLE, quotedValue( mSchemaName ), quotedValue( mTableName ), quotedValue( field.name() ) );
-      QgsPostgresResult result( connectionRO()->PQexec( sql ) );
-      for ( int i = 0; i < result.PQntuples(); ++i )
-      {
-        const QString type = result.PQgetvalue( i, 0 );
-        QVariantMap config;
-        if ( !result.PQgetisnull( i, 1 ) ) // Can be null and it's OK
-        {
-          const QString configTxt = result.PQgetvalue( i, 1 );
-          QDomDocument doc;
-          if ( doc.setContent( configTxt ) )
-          {
-            config = QgsXmlUtils::readVariant( doc.documentElement() ).toMap();
-          }
-          else
-          {
-            QgsMessageLog::logMessage( tr( "Cannot parse widget configuration for field %1.%2.%3\n" ).arg( mSchemaName, mTableName, field.name() ), tr( "PostGIS" ) );
-          }
-        }
+    return;
+  }
 
+  QStringList quotedFnames;
+  foreach ( const QString &name, mAttributeFields.names() )
+  {
+    quotedFnames << quotedValue( name );
+  }
+
+  // We expect the table to be created like this:
+  //
+  // CREATE TABLE qgis_editor_widget_styles (schema_name TEXT NOT NULL, table_name TEXT NOT NULL, field_name TEXT NOT NULL,
+  //                                         type TEXT NOT NULL, config TEXT,
+  //                                         PRIMARY KEY(schema_name, table_name, field_name));
+  const QString sql = QStringLiteral( "SELECT field_name, type, config "
+                                      "FROM %1 WHERE schema_name = %2 "
+                                      "AND table_name = %3 "
+                                      "AND field_name IN ( %4 )" ) .
+                      arg( EDITOR_WIDGET_STYLES_TABLE, quotedValue( mSchemaName ),
+                           quotedValue( mTableName ), quotedFnames.join( "," ) );
+  QgsPostgresResult result( connectionRO()->PQexec( sql ) );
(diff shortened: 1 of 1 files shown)


strk:
I left it to have a Travis backed proof of the removed code NOT being needed for what the test was testing. Just in case my test run was bogus...

So I take you agree with the removal @pvalsecc  ?

pvalsecc:
> So I take you agree with the removal @pvalsecc ?

No, not at all.
But I don't have time to fix it... and I don't have the energy to fight for it.
So suit yourself.

strk:
I did the fixing work already so it's not much energy to keep it, but it takes energy to PROVE it is needed. If the only existing test passes even w/out that code, what's the point of keeping it ?

It doesn't take a fight to keep the code, just a document describing how it works and a test proving it works.

I'm happy to give advice a to how to make the support not bring down performances

haubourg:
> It doesn't take a fight to keep the code, just a document describing how it works and a test proving it works.

Much agreed, this is the second time this hidden feature takes a lot of time to a some of us. And causes performance issues for users. 

@pvalsecc  Can you please (again) document the feature in the code and push a link to the eventual tool using it (hopefully it is a public plugin). If this feature is for a non publishable use and it is not possible to document it, I tend to think we should remove it faster than proposed in the dev's list, ie mark it deprecated and wait for 4.0 to remove it.  cc @nyalldawson @yjacolin

strk:
@haubourg can you confirm the code in this PR fixed the speed issue for you ?

strk:
Travis fail with a segfault while the relevant test on my machine works just fine, I'm confused

m-kuhn:
I still tend to think we shouldn't just drop it on such a short notice.

What about this:

 - Add a QSettings based opt-in approach to still enable this feature with 3.10.
 - Deprecate it for removal with 3.12 (the next non-LTR)

Eventually this might be used by a production system somewhere. Dropping support now for 3.10 is too quick to react for any business process and potentially leaves customers with mixed feelings about the reliability of QGIS.

@pvalsecc that would give you a year before it's dropped from LTR (and 4 months before it's dropped from the code). Is this enough time to check with your project lead / the client how we can come to a future proof solution?

strk:
I like the QSettings opt-in approach idea. It would also give us a chance to specify exact schema in which the table would be found, further improving the code reading it..

strk:
As I keep getting Travis failures of various kind (last one was an indentation error when the change reduced to ONLY removing lines) I've now force-pushed a no-change commit (all commits reverted). Will slowly re-introduce commits if there's a green, but reading actual errors from Travis is really revealing problematic.

m-kuhn:
Did you setup the pre commit hook? That should sort out all the indentation and formatting.

strk:
Yes my pre-commit hook is in place, as a symlink to ../../scripts/prepare-commit.sh (and I do see it working)

strk:
Ok, I now understand that the existing test will actually FAIL if we remove the support (Travis confirms this). What the test does is it checks that `editorWidgetSetup()` called against a QgsField of a QgsVectorLayer loaded from PostgreSQL  is not null, when a configuration is specified, for that field, in the `qgis_editor_widget_styles` table.

pvalsecc:
Happy to here I didn't mess up too much back then :-)

strk:
I get a segfault from PyQgsPostgresProvider test, upon calling `conn->PQexec( sql );`
The sql string seems fine, but that call never returns, so I suspect it's a deleted connection, not sure how it gets deleted...

strk:
It seems to be the destruction of a QgsPostgresResult:
```
#0  0x00007f1eb778f7d5 in PQclear () at /usr/lib/x86_64-linux-gnu/libpq.so.5
#1  0x00007f1e84467215 in QgsPostgresResult::~QgsPostgresResult() (this=0x7ffd4d4d49b0, __in_chrg=<optimized out>)
    at /usr/src/qgis/qgis-improve-metadata-tables-creation-and-usage/src/providers/postgres/qgspostgresconn.cpp:52
#2  0x00007f1e844245a1 in QgsPostgresProviderMetadata::saveStyle(QString const&, QString const&, QString const&, QString const&, QString const&, QString const&, bool, QString&) (this=
    0x34769e0, uri=..., qmlStyle=..., sldStyle=..., styleName=..., styleDescription=..., uiFileContent=..., useAsDefault=false, errCause=...)
    at /usr/src/qgis/qgis-improve-metadata-tables-creation-and-usage/src/providers/postgres/qgspostgresprovider.cpp:4757
```

I guess I made the scope of those variables too wide

strk:
Interesting (or not?) re-using a QgsPostgresResult seems to trigger the segfault. Using another object doesn't have problems

strk:
I think I found the culprit.
This segfaults:
> res = QgsPostgresResult( conn->PQexec( checkQuery ) );

This is fine:
> res = conn->PQexec( checkQuery );

@jef-n I suspect we'd want to add some protection in the copy costructor for QgsPostgresResult...

strk:
@m-kuhn @pvalsecc  are you ok with merging this PR as-is (no QSetting, just improved query, passing existing test) ? We can do the QSettings in 3.10

nyalldawson:
Hey @strk I'm lost with the status on this one - my understanding is that the current pr is just optimisations which doesn't change current behaviour. Is that correct?

strk:
@nyalldawson that's correct, no behavioral change, just optimization

strk:
@jgrocha I saw you changed the code for layer_styles recently, can you please check if things still work for you with this PR ?

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -1126,36 +1126,58 @@ bool QgsPostgresProvider::loadFields()
...
+    {
+      QgsMessageLog::logMessage(
+        tr( "Cannot parse widget configuration for field %1.%2.%3\n" )
+        .arg( mSchemaName, mTableName, fname ), tr( "PostGIS" )
+      );
+      continue;
+    }
 
+    // Set corresponding editor widget
+    for ( auto &field : mAttributeFields )
```
m-kuhn:
```suggestion
    for ( auto &field : qgis::as_const( mAttributeFields ) )
```

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -1126,36 +1126,58 @@ bool QgsPostgresProvider::loadFields()
...
 {
-  if ( tableExists( *connectionRO(), EDITOR_WIDGET_STYLES_TABLE ) )
+  if ( ! tableExists( *connectionRO(), EDITOR_WIDGET_STYLES_TABLE ) )
   {
-    for ( int i = 0; i < mAttributeFields.count(); ++i )
+    return;
+  }
+
+  QStringList quotedFnames;
+  foreach ( const QString &name, mAttributeFields.names() )
```
m-kuhn:
```suggestion
  const QStringList fieldNames;
  for ( const QString &name : fieldNames )
```

m-kuhn (approved):

src/providers/postgres/qgspostgresprovider.cpp
strk: this suggestion isn't clear, partial (fieldNames will be empty?)

src/providers/postgres/qgspostgresprovider.cpp
strk: field needs be non-const in order to be able to do field.setEditorWidgetSetup

src/providers/postgres/qgspostgresprovider.cpp
m-kuhn: Sorry, updated.

src/providers/postgres/qgspostgresprovider.cpp
m-kuhn: Most classes will detach and work on a copy of the container if iterating them non const.
Does this code here work as expected with QgsFields?

strk: As tests do exist and used to fail when I made a mistake I'd say it's working as-is

pvalsecc:
> @m-kuhn @pvalsecc are you ok with merging this PR as-is

Yes, I'm OK with the PR. Thanks a lot.

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -1132,7 +1132,7 @@ void QgsPostgresProvider::setEditorWidgets()
   }
 
   QStringList quotedFnames;
-  foreach ( const QString &name, mAttributeFields.names() )
+  for ( const QString &name : mAttributeFields.names() )
```
m-kuhn: This will detach

m-kuhn (changes requested):
I had a good reason to suggest the changes the way I did ;)

src/providers/postgres/qgspostgresprovider.cpp
strk: Where can I read more about this "detach" thing ?
I'm iterating it as const (you mentioned a condition being iterating as non-const?).
Also, why   const QStringList fieldNames = mAttributeFields.names() instead of holding a const reference  ? Let's take the chance to see if test will fail with current code ?

strk: btw, I'm sure there's a more elegant way to call a function for each element in a container...

m-kuhn: Tests won't fail. It's just a matter of loosing some memory and cpu cycles.
There's a thorough explanation here: https://www.kdab.com/goodbye-q_foreach/

m-kuhn: This one looks nice https://stackoverflow.com/questions/10516495/calling-a-function-on-every-element-of-a-c-vector I've never tested it though

m-kuhn (approved):
Thanks, I'll ignore the doubtful commit message

jgrocha (approved):
I'm ok with the PR. Thanks for improving the initial PR.

strk:
Would this be good to backport to 3.10 @m-kuhn @nyalldawson ?

m-kuhn:
Yes, I think so. Should we wait for a couple of days in master before merging the backport or is it low-risk?

strk:
Given how young 3.10 is I'd say it's low risk. We'll get a bunch of issues reported for 3.10 anyway, so we'll be around to fix them

m-kuhn:
The risk isn't related to the age of 3.10 ;)

strk:
Backport PR: #32573

PR #31865, opened 2019-09-18
Server fix getLegendGraphic default WIDTH/HEIGHT when BBOX

### LINKED ISSUE #31846: Server GetLegendGraphics contextual (BBOX) fails if no WIDTH and HEIGHT are supplied
For instance, using the projects and layers used in our CI tests:
```
?MAP=/dev/QGIS/tests/testdata/qgis_server/test_project_wms_grouped_nested_layers.qgs&SERVICE=WMS&VERSION=1.3&REQUEST=GetLegendGraphic&LAYER=areas%20and%20symbols&FORMAT=image/png&CRS=EPSG:4326&BBOX=52.4124,10.6727,52.4649,10.7794&SLD_VERSION=1.1
```

fails with 
```xml
<ServiceException code="InvalidParameterValue">The requested map size is too large</ServiceException>
</ServiceExceptionReport>
```

elpaso:
is set.

Fixes #31846


pblottiere:
It looks good to me, thanks @elpaso :+1:

src/server/services/wms/qgswmsgetlegendgraphics.cpp
```diff
@@ -153,10 +153,24 @@ namespace QgsWms
       throw QgsBadRequestException( QgsServiceException::QGIS_InvalidParameterValue,
                                     parameters[QgsWmsParameter::BBOX] );
     }
+    // If we have a contextual legend (BBOX is set)
+    // make sure WIDTH and HEIGHT are > 0, default to 800x600
+    if ( ! parameters.bbox().isEmpty() )
+    {
+      if ( parameters.width().isEmpty() )
+      {
+        parameters.set( QgsWmsParameter::WIDTH, 800 );
```
pblottiere: I think that to be perfectly explicit, we could use SRCWIDTH in this case.

elpaso: Yes, I thought about it, I will change and test this.

PR #31871, opened 2019-09-18
Fix double event fired by QSpinBox in slow graduated

### LINKED ISSUE #31635: UI: any click on number of classes updown triggers twice
In layer properties, whenever I click on the number of classes updown, it gets changed (slowly recalculates), and repeats the action after first recalculation gets finished. Very annoying on slow coloring modes like Quantiles.

image

elpaso:
... renderer widget

Fixes #31635

For the record of the underlying issue:

https://lists.qt-project.org/pipermail/interest/2013-July/007936.html
https://forum.qt.io/topic/82181/qt-doublespin-box-value-changed-slot-has-called-twice
https://forum.qt.io/topic/96094/qspinbox-value-changed-slot-has-called-twice-on-mouse-click
https://www.qtcentre.org/threads/53709-slot-valueChanged(double)-of-doubleSpinBox-called-twice-if-breakpoint-is-set

### DIFF
--- a/src/gui/symbology/qgsgraduatedsymbolrendererwidget.cpp
+++ b/src/gui/symbology/qgsgraduatedsymbolrendererwidget.cpp
@@ -486,6 +486,8 @@ QgsGraduatedSymbolRendererWidget::QgsGraduatedSymbolRendererWidget( QgsVectorLay
   spinPrecision->setMinimum( QgsClassificationMethod::MIN_PRECISION );
   spinPrecision->setMaximum( QgsClassificationMethod::MAX_PRECISION );
 
+  spinGraduatedClasses->setShowClearButton( false );
+
   btnColorRamp->setShowRandomColorRamp( true );
 
   // set project default color ramp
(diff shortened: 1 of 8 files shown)


src/gui/symbology/qgsgraduatedsymbolrendererwidget.cpp
```diff
@@ -906,6 +910,8 @@ void QgsGraduatedSymbolRendererWidget::symmetryPointEditingFinished( )
 
 void QgsGraduatedSymbolRendererWidget::classifyGraduated()
 {
+
+  QApplication::setOverrideCursor( Qt::WaitCursor );
```
nyalldawson: There's a scoped class in GUI which handles this nicely

nyalldawson: (QgsTemporaryCursorOverride)

nyalldawson:
Gosh, that's pretty... yuck. Oh well. 

I'm guessing QgsDoubleSpinBox should have the same workaround?

src/gui/symbology/qgsgraduatedsymbolrendererwidget.cpp
elpaso: Thanks! I was pretty sure that there was a scoped helper but I didn't search hard enough.
Do you think we could collect all this little helpers in a dedicated doxy section?

elpaso:
> Gosh, that's pretty... yuck. Oh well.
> 
> I'm guessing QgsDoubleSpinBox should have the same workaround?

I think so, it also inherits from QAbstractSpinBox. Would a mixin base class work for us in this case?

nyalldawson:
> I think so, it also inherits from QAbstractSpinBox. Would a mixin base class work for us in this case?

Not in c++ it won't :P

elpaso:
I'm still eating too much Python these days ;)

PR #31885, opened 2019-09-19
Add "concatenate_unique" and "array_agg" as aggregate parameter

### LINKED ISSUE #31877: [Feedback] Function_Help.json file for aggregate its missing a valid option "array_agg"
Hi! 

So I was reading about aggregates in QGIS with the expression builder and came across with a valid option in the aggregate function that is missing in the help panel that came form the function help JSON files.

the option is and aggregate with valid option "array_agg"

I used in the fallowing expression:  

`aggregate( 'my_layer', 'array_agg',  "COD_MGN_2021" , intersects(  $geometry , geometry(@parent) ),  ' , '  ) `

And works perfectly, so I search for the help panel in the expression builder to see the list of all valid options when using the aggregate function and didn't see the "array_agg" option, but in my expression works so it most be implemented.

I search for the Function Help Json files, and see that the option is not writed there,
just to let you know that is a missing valid option in the Function Help JSON file for the aggregate function. 

I dont know if there are more valid options that not appear in the help file. 

Thaks for all the attention and your work!! <3

DelazJ:
to aggregate and relation_aggregate functions (fixes #31877)

Btw what does represent the `stdevsample` options that is mentioned in the list of aggregate values?


edurojasr:
Thanks @DelazJ 
Yes that was missing in the help files

DelazJ:
@edurojasr Thanks for reporting

nyalldawson:
> Btw what does represent the stdevsample options that is mentioned in the list of aggregate values?

There's stdev (standard deviation of a complete population) and stdevsample (standard deviation of sampled population).

DelazJ:
Indeed. I can find both in the statistics panel but the latter is not exposed as a function _per se_ in the expression builder (while mentioned as a sub function of aggregate and relation_aggregate). On purpose?
And from the statistic panel, I see that variety is also missing in expression...

nyalldawson:
Is this one ready to merge?

I think the reason variety was skipped was that when aggregates were first added we had no list/array support in expressions. We should address this gap now though.

DelazJ:
> Is this one ready to merge?

I'd say yes; it fixes the report. The remaining point (expose variety and stdevsample as functions) is out of scope.

PR #31892, opened 2019-09-19
Allow to collapse the GeoPdf options when exporting to pdf

DelazJ:
When available these options take too much space if you do not need them. 
## Checklist


nyalldawson:
This was an intentional move -- I didn't want the warning message to be hidden by default when geopdf export is not available.

Can you ensure that the box IS expanded by default when geopdf export isn't available?

DelazJ:
Like this?
image
from the doc, the qgscollapsiblegroupboxasic does not save the last state so i guess it would remain as configured

PR #31899, opened 2019-09-20
Update general.py

paulwittle:
## Description

Correction of error return type from bool to None (as per comments). This is to resolve unexpected exception message (AttributeError: 'bool' object has no attribute 'show') when algorithm is not found.

I'm not sure I've created this pull request correctly at all as I'm still new to open source development. Do I need to create a bug report and then rename this pull request?

I'm also pretty unsure regarding line endings etc as I'm using a Windows PC but I'm hoping it has not changed them as I've only changed one word and I edited the file directly in Visual Studio which should have recognised it as a python script.

## Checklist

### DIFF
--- a/python/plugins/processing/tools/general.py
+++ b/python/plugins/processing/tools/general.py
@@ -168,7 +168,7 @@ def createAlgorithmDialog(algOrName, parameters={}):
         alg = QgsApplication.processingRegistry().createAlgorithmById(algOrName)
 
     if alg is None:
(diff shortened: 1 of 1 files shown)


m-kuhn:
Hi @paulwittle 

All good with this PR as is.

For future ones, if you change the commit message to be more verbose regarding what changed exactly that would be even better. Creating an issue report would be awesome but is not strictly required.

PR #31901, opened 2019-09-20
[processing] correctly handle case when layer source is not a plain path (fix #29336)

### LINKED ISSUE #29336: Processing "Build virtual vector" now fails in certain cases
Original Redmine Issue: [21519](https://issues.qgis.org/issues/21519)
Affected QGIS version: 3.6.0
Redmine category:processing/qgis
Assignee: Victor Olaya

---
1. In 3.x, if I load shape files as directories (i.e. drag and drop the containing directory into QGIS, rather than the shapefile itself), and then try to use those layers as inputs, I get an error message:
```
File "C:/PROGRA~1/QGIS3~1.6/apps/qgis/./python/plugins\processing\algs\qgis\Datasources2Vrt.py", line 102, in processAlgorithm
feedback=feedback)
File "C:/PROGRA~1/QGIS3~1.6/apps/qgis/./python/plugins\processing\algs\qgis\Datasources2Vrt.py", line 133, in mergeDataSources2Vrt
self.tr('Invalid datasource: {}'.format(inFile)))
_core.QgsProcessingException: Invalid datasource: //path_obfuscated/Area_2/BD34_0109|layername=BD34_0109

Execution failed after 0.10 seconds

```
It succeeds without error in 2.18.x.

2. If two of the input layers have the same name and I create a "unioned" VRT using 3.x the vrt contains one of them twice and omits the other one.  2.18.x correctly creates a vrt containing both.

3. There is another buglet that isn't a regression as 2.18.x did not support loading the output file after creating it: run the algorithm with 'create "unioned" VRT' unchecked and "open output file after running algorithm" checked, and only one layer in the output VRT will be _loaded_.  However, the VRT was actually created correctly (drag and drop the VRT into QGIS and you will get the dialog to select which layers to load).

alexbruy:
## Description
If input layer path is not a plain path (e.g. contains `|layername=` or `|layerid=`) Build virtual layer algorithm fails. Fixes #29336.

BTW, IMHO this algorithm should be in the GDAL provider, not in QGIS.

## Checklist


python/plugins/processing/algs/qgis/Datasources2Vrt.py
```diff
@@ -82,151 +81,44 @@ def defaultFileExtension(self):
...
             vrt += '<OGRVRTUnionLayer name="UnionedLayer">'
 
-        total = 100.0 / len(dataSources) if dataSources else 1
-        for current, layer in enumerate(dataSources):
+        total = 100.0 / len(input_layers) if input_layers else 0
+        for current, layer in enumerate(input_layers):
             if feedback.isCanceled():
                 break
 
+            uri = layer.source()
```
nyalldawson: Shouldn't this use QgsProviderRegistry.instance().decodeUri instead?

python/plugins/processing/algs/qgis/Datasources2Vrt.py
```diff
@@ -82,151 +81,44 @@ def defaultFileExtension(self):
...
-            return 'String'
-
-    def XmlEsc(self, text):
-        return xml.sax.saxutils.escape(text)
+        with open(vrtPath, 'w', encoding='utf-8') as f:
+            f.write(vrt)
+
+        return {self.OUTPUT: vrtPath, self.VRT_STRING: vrt}
+
+    def escape(self, text):
```
nyalldawson: Could this be replaced by QString::toHtmlEscaped()?

nyalldawson:
> BTW, IMHO this algorithm should be in the GDAL provider, not in QGIS.

Agreed. But I don't think we can do that now without breaking api. We'd need a new way to define manual algorithm renames and hook that into QgsProcessingRegistry::algorithmById -- e.g. we could add a method to QgsProcessingRegistry:

    void addAlgorithmAlias( const QString &aliasProviderId, const QString& aliasAlgorithmName, const QString& actualProviderId, const QString& actualAlgorithName );

Which would populate a map of alias -> actual inside QgsProcessingRegistry, and then QgsProcessingRegistry::algorithmById would have to check this map for overrides, much like it currently does to redirect "qgis:" algorithms to "native:"....

python/plugins/processing/algs/qgis/Datasources2Vrt.py
alexbruy: If I'm not wrong, we can not use `QString` in python now, it is mapped to `str`.

python/plugins/processing/algs/qgis/Datasources2Vrt.py
alexbruy: I think better to use GDAL connection string here, to allow also database and WFS layers. Should be fixed now.

python/plugins/processing/algs/qgis/Datasources2Vrt.py
nyalldawson: Ok, what about the python html.escape method?

python/plugins/processing/algs/qgis/Datasources2Vrt.py
nyalldawson: Good move!

alexbruy:
@nyalldawson do you plan to merge #31974 before 3.10?

nyalldawson:
@alexbruy yes, done!

nyalldawson (approved):

PR #31920, opened 2019-09-21
Fix Server OPENAPI validation and unreported QStringList to JSON conversion error

elpaso:
- Fix JSON unreported issue with QStringList to json conversion
-  Fix some OPENAPI validation issues.

Swagger is happy now, except for the MAP=/...
in the query string: no query string parameters
are allowed in the endpoint, but we can
fix this in the web server configuration.

### DIFF
--- a/resources/server/api/ogc/schema.json
+++ b/resources/server/api/ogc/schema.json
@@ -264,7 +264,7 @@
       "limit" : {
         "name" : "limit",
         "in" : "query",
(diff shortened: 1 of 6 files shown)


m-kuhn:
> Fix JSON unreported issue with QStringList to json conversion

What's this exactly? (Just for the record)

elpaso:
A QStringList was converted to a plain CSV string instead of an array

PR #31936, opened 2019-09-22
Avoid touching unchanged sip files in prepare commit script

nyalldawson:
Since any modification to these files triggers a new cmake run and rebuilding of the sip bindings (slow!), we want to avoid touching the sip files in any way if there's no actual changes to push.

Speeds up rebuilding in some circumstances by avoiding needless rebuilding of sip bindings.

### DIFF
--- a/scripts/prepare-commit.sh
+++ b/scripts/prepare-commit.sh
@@ -120,11 +120,15 @@ for f in $MODIFIED; do
     if grep -Fq "$sip_file" "${TOPLEVEL}"/python/"${module}"/"${module}"_auto.sip; then
       sip_file=$(${GP}sed -r 's@^src/(core|gui|analysis|server|3d)@\1/auto_generated@; s@\.h$@.sip.in@' <<<"$f" )
       m=python/$sip_file.$REV.prepare
-      touch python/"$sip_file"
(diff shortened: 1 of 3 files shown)


scripts/prepare-commit.sh
```diff
@@ -120,11 +120,15 @@ for f in $MODIFIED; do
...
       m=python/$sip_file.$REV.prepare
-      touch python/"$sip_file"
+      if [ ! -f python/"$sip_file" ]; then
+        touch python/"$sip_file"
+      fi
       cp python/"$sip_file" "$m"
-      "${TOPLEVEL}"/scripts/sipify.pl -s python/"$sip_file" -p python/"${module}"/auto_additions/"${pyfile}" "$f"
-      if ! diff -u "$m" python/"$sip_file" >>"$SIPIFYDIFF"; then
+      "${TOPLEVEL}"/scripts/sipify.pl -s $m -p python/"${module}"/auto_additions/"${pyfile}" "$f"
+      # only replace sip files if theygit  have changed
```
DelazJ: Theygit?

nyalldawson: Thanks, fixed

PR #31948, opened 2019-09-23
Reorganize Grid decoration GUI

DelazJ:
Put together all the properties relative to the grid drawing
Make line and maker symbols display mutually exclusive
The current 
image

The proposed
image

## Checklist


nirvn:
Nice, that dialog is in need of some love.

Comment: please but the internal x and internal y spinbox on separate lines (with the interval label spanning over two rows), same for offset.

Beyond better matching our UI generally speaking, I'm hoping I'll have time to move decorators into a unified decorator panel in the next dev cycle, and we need to keep the overall width as narrow as possible.

DelazJ:
@nirvn Here it's
image

nirvn:
Man, what a difference this UI cleanup made.

PR #31951, opened 2019-09-23
Feature update layer selection relation widgets

troopa81:
## Description

Before this PR, selected features in relation widget and feature selection dialog for link existing child was not connected to layer selection. 

This is problematic because you cannot use map selection if you want to choose child to create a relation with a given parent. This is what it looks like with the PR. 

selection2

The existing behaviour seems to be made on purposed using the `QgsGenericFeatureSelectionManager`. This class was only used for this 2 widgets. So, I would like to know if the behavior is OK. Once we agree, I will complete the PR with some unit test.

To be complete, I plan to add the feature selection widget widget_sel1 and widget_sel2 to the feature selection dialog and make it non modal. But, it would be the object of a separate PR.

cc @haubourg 

## Checklist

### DIFF
--- a/src/gui/attributetable/qgsvectorlayerselectionmanager.h
+++ b/src/gui/attributetable/qgsvectorlayerselectionmanager.h
@@ -48,6 +48,14 @@ class GUI_EXPORT QgsVectorLayerSelectionManager : public QgsIFeatureSelectionMan
     QgsVectorLayer *layer() const;
 
   private slots:
+
+    /**
+     * Called whenever layer selection was changed
+     *
+     * \param selected        Newly selected feature ids
+     * \param deselected      Ids of all features which have previously been selected but are not any more
+     * \param clearAndSelect  In case this is set to TRUE, the old selection was dismissed and the new selection corresponds to selected
+     */
     virtual void onSelectionChanged( const QgsFeatureIds &selected, const QgsFeatureIds &deselected, bool clearAndSelect );
 
   private:
--- a/src/gui/attributetable/qgsvectorlayerselectionmanager.h
+++ b/src/gui/attributetable/qgsvectorlayerselectionmanager.h
@@ -41,6 +41,10 @@ class GUI_EXPORT QgsVectorLayerSelectionManager : public QgsIFeatureSelectionMan
     void deselect( const QgsFeatureIds &ids ) override;
     void setSelectedFeatures( const QgsFeatureIds &ids ) override;
     const QgsFeatureIds &selectedFeatureIds() const override;
+    QgsVectorLayer *layer() const;
+
+  protected slots:
+    virtual void onSelectionChanged( const QgsFeatureIds &selected, const QgsFeatureIds &deselected, bool clearAndSelect );
 
   private:
     QgsVectorLayer *mLayer = nullptr;
--- a/src/gui/attributetable/qgsvectorlayerselectionmanager.h
+++ b/src/gui/attributetable/qgsvectorlayerselectionmanager.h
@@ -41,9 +41,13 @@ class GUI_EXPORT QgsVectorLayerSelectionManager : public QgsIFeatureSelectionMan
     void deselect( const QgsFeatureIds &ids ) override;
     void setSelectedFeatures( const QgsFeatureIds &ids ) override;
     const QgsFeatureIds &selectedFeatureIds() const override;
+
+    /**
+     * Returns the vector layer
+     */
     QgsVectorLayer *layer() const;
 
-  protected slots:
+  private slots:
     virtual void onSelectionChanged( const QgsFeatureIds &selected, const QgsFeatureIds &deselected, bool clearAndSelect );
 
   private:
--- a/src/gui/qgsrelationeditorwidget.cpp
+++ b/src/gui/qgsrelationeditorwidget.cpp
@@ -272,6 +272,7 @@ void QgsRelationEditorWidget::initDualView( QgsVectorLayer *layer, const QgsFeat
   mFeatureSelectionMgr = new QgsFilteredSelectionManager( layer, request, mDualView );
   mDualView->setFeatureSelectionManager( mFeatureSelectionMgr );
   connect( mFeatureSelectionMgr, &QgsIFeatureSelectionManager::selectionChanged, this, &QgsRelationEditorWidget::updateButtons );
+  updateButtons();
 }
 
 void QgsRelationEditorWidget::setRelations( const QgsRelation &relation, const QgsRelation &nmrelation )
--- a/src/gui/qgsrelationeditorwidget.cpp
+++ b/src/gui/qgsrelationeditorwidget.cpp
@@ -30,6 +30,7 @@
 #include "qgslogger.h"
 #include "qgsvectorlayerutils.h"
 #include "qgsmapcanvas.h"
+#include "qgsvectorlayerselectionmanager.h"
 
 #include <QHBoxLayout>
 #include <QLabel>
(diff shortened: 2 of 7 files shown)


troopa81:
@m-kuhn I'm interested about your opinion on this.

troopa81:
> What happens if the user selects a feature on the canvas which is not a child of the current feature and then deletes the selected child features?

Nothing. The form still displays the parent and its child features. I try also the delete selected child or parent features from the canvas or the form and it works like expected, the form and canvas are refreshed according to selection and existing feature.

m-kuhn:
Sorry, I wasn't precise enough with my question; are features deleted which are not visible on the relation widget?

troopa81:
OK, I get your point. If you select features, and some of them are not part of the current relation child features, if you click on delete from relation widget, it delete all the selected features even the one not visible in the relation.

2 possibilities: 
- We consider the delete button as a simple shortcut to the global one. It deletes all the selected feature even if there are not part of the relation (but it could lead to unexpected behavior, and so it's maybe better to remove it here)
- We delete only the feature selected AND which are part of the current relation

I would vote for the second one.

Same question apply also to zoom to selected feature.

m-kuhn:
Yep, that sounds like a good approach :+1:

src/gui/attributetable/qgsvectorlayerselectionmanager.h
```diff
@@ -41,6 +41,10 @@ class GUI_EXPORT QgsVectorLayerSelectionManager : public QgsIFeatureSelectionMan
     void deselect( const QgsFeatureIds &ids ) override;
     void setSelectedFeatures( const QgsFeatureIds &ids ) override;
     const QgsFeatureIds &selectedFeatureIds() const override;
+    QgsVectorLayer *layer() const;
+
+  protected slots:
+    virtual void onSelectionChanged( const QgsFeatureIds &selected, const QgsFeatureIds &deselected, bool clearAndSelect );
```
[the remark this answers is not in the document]
troopa81: If I do that, the QgsIFeatureSelectionManager::selectionChanged signal will be emitted twice, one with all selected features, one will the selected features that belongs to the relation.

The point of virtualization is to replace the QgsVectorLayerManager behavior so I can emit just the signal once with the filtered feature.

src/gui/qgsrelationeditorwidget.cpp
```diff
@@ -30,11 +30,70 @@
...
+      return mSelectedFeatureIds;
+    }
+
+
+    int selectedFeatureCount() override
+    {
+      return mSelectedFeatureIds.count();
+    }
+
+  public slots:
```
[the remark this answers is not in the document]
m-kuhn: Regardless of the outcome of the discussion the slot should be private.

src/gui/attributetable/qgsvectorlayerselectionmanager.h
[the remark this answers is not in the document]
m-kuhn: Sounds good :+1:

nyalldawson:
@m-kuhn does this look good to you?

src/gui/attributetable/qgsvectorlayerselectionmanager.h
```diff
@@ -42,6 +42,14 @@ class GUI_EXPORT QgsVectorLayerSelectionManager : public QgsIFeatureSelectionMan
     void setSelectedFeatures( const QgsFeatureIds &ids ) override;
     const QgsFeatureIds &selectedFeatureIds() const override;
 
+    /**
+     * Returns the vector layer
+     */
+    QgsVectorLayer *layer() const;
+
+  private slots:
+    virtual void onSelectionChanged( const QgsFeatureIds &selected, const QgsFeatureIds &deselected, bool clearAndSelect );
```
m-kuhn: Can you add a docstring to that explaining default behavior and specs for overriding?

src/gui/qgsrelationeditorwidget.cpp
```diff
@@ -30,11 +30,74 @@
...
 #include "qgsmapcanvas.h"
+#include "qgsvectorlayerselectionmanager.h"
 
 #include <QHBoxLayout>
 #include <QLabel>
 #include <QMessageBox>
 
+/// @cond PRIVATE
+
+class QgsFilteredSelectionManager : public QgsVectorLayerSelectionManager
```
m-kuhn: Can you add some docstrings explaining the rational for this class

src/gui/qgsrelationeditorwidget.cpp
```diff
@@ -198,10 +258,18 @@ void QgsRelationEditorWidget::setRelationFeature( const QgsRelation &relation, c
...
     QgsFeatureRequest myRequest = mRelation.getRelatedFeaturesRequest( mFeature );
-    mDualView->init( mRelation.referencingLayer(), nullptr, myRequest, mEditorContext );
+    initDualView( mRelation.referencingLayer(), myRequest );
   }
 }
 
+void QgsRelationEditorWidget::initDualView( QgsVectorLayer *layer, const QgsFeatureRequest &request )
+{
+  mDualView->init( layer, nullptr, request, mEditorContext );
+  mFeatureSelectionMgr = new QgsFilteredSelectionManager( layer, request, mDualView );
```
m-kuhn: This iterates over the features twice, correct?
Once in the dual view and once in the feature selection mgr. I wonder if this could be reduced to one single time.

m-kuhn:
Thanks for the update.
I wonder what happens if you add a new child (create or link) in the form and then select and delete it? Is that not filtered by the filter model which initializes it's "selectable fids" once at instantiation?

troopa81:
> I wonder what happens if you add a new child (create or link) in the form and then select and delete it? Is that not filtered by the filter model which initializes it's "selectable fids" once at instantiation?

No because we are connected to layer selectionChanged so we update the selection every time it change. And if the object is already selected and we link the feature, updateUi is called and the QgsFilteredManager is rebuilt with new request.

m-kuhn (approved):
Thanks for the clarification.
Looks good to me

src/gui/qgsrelationeditorwidget.cpp
troopa81: Actually, there was and issue on the way I built the selected feature ids list b23a972
So now, it iterates only on selected feature list to check if the feature matches the relation.

I cannot a see a correct way to do this one single time.

src/gui/qgsrelationeditorwidget.cpp
troopa81: Done

src/gui/attributetable/qgsvectorlayerselectionmanager.h
troopa81: Done

troopa81:
@nyalldawson Is there a reason why you label this PR as bug? I removed it because I consider this PR a feature

troopa81:
Thanks for the review and merge @m-kuhn

src/gui/qgsrelationeditorwidget.cpp
```diff
@@ -198,10 +259,19 @@ void QgsRelationEditorWidget::setRelationFeature( const QgsRelation &relation, c
...
+  mDualView->setFeatureSelectionManager( mFeatureSelectionMgr );
```
3nids: @troopa81 
I am getting crashes from here
image

troopa81: @3nids Doing what exactly? Did you create an issue? Sorry for late answer, I was on leave

PR #31956, opened 2019-09-23
[dbmanager] use CRS selector in the Create table dialog (fix #25535)

### LINKED ISSUE #25535: DB Manager: Create Layer dialog should be closed or cleared once the new layer is generated
Original Redmine Issue: [17638](https://issues.qgis.org/issues/17638)
Affected QGIS version: master
Redmine category:db_manager

---
Open the DB Manager
Select a gpkg database (i didn't test other formats)
In Table Menu, select "Create a table"
Fill the dialog and click Create
You'll see the layer added to the database but the "create new layer" dialog is still open.
It should either be open and cleared or closed. (unless it's by design, to help recreate new layers from the previous one? not sure what behavior is better)

alexbruy:
## Description
Create table dialog still used plain line edit for SRID, this PR replaces it with QGIS CRS selector. Also reset dialog UI to the initial state on successful table creation. Fixes #25535.

## Checklist

### DIFF
--- a/python/plugins/db_manager/dlg_create_table.py
+++ b/python/plugins/db_manager/dlg_create_table.py
@@ -28,6 +28,7 @@ from qgis.PyQt.QtCore import Qt, QModelIndex
 from qgis.PyQt.QtWidgets import QItemDelegate, QComboBox, QDialog, QPushButton, QDialogButtonBox, QMessageBox, QApplication
 from qgis.PyQt.QtCore import QItemSelectionModel, pyqtSignal
 
+from qgis.core import QgsProject
 from qgis.utils import OverrideCursor
 
 from .db_plugins.data_model import TableFieldsModel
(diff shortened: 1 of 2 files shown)


python/plugins/db_manager/dlg_create_table.py
```diff
@@ -288,7 +291,7 @@ def createTable(self):
             geomType = self.GEOM_TYPES[self.cboGeomType.currentIndex()]
             geomDim = self.spinGeomDim.value()
             try:
-                geomSrid = int(self.editGeomSrid.text())
+                geomSrid = self.widgetTargetSrid.crs().postgisSrid()
```
nyalldawson: How does postgisSrid work with a custom/user projection? I would like to see us avoid this method wherever possible -- it has no place in the wonderful world of proj >= 6 and will likely be removed in QGIS 4.0. Rather, we should lookup the correct code here from the spatial_ref_sys table by looking up the authority and identifier (and falling back to lookup by proj string for user projections).

alexbruy: As I can see it never worked with custom projections, as it was a simple line edit and entered value was casted to integer id before using (and falling back to 0 in case of failure)

nyalldawson: Right, the previous approach put the responsibility on the user to enter a valid postgis srs id. But by flipping to the user-friendly widget, QGIS is taking on that responsibility. And currently we aren't handling it correctly...

PR #31959, opened 2019-09-23
Fix browser crash when adding (broken) WMTS

### LINKED ISSUE #31927: Crash on adding WMS
3.4.12 - Windows

1) Add this WMTS to your connections:
http://geoportal.eberswalde.de:8080/geoserver/gwc/service/wmts

2) (optional) restart qgis.

3) Double click the WMTS connection in the browser so that it connects. (this is the important part)

4) Add this WMS connection:
http://apps.siba.geosphere.fr/cgi-bin/mapserv.exe?map=C:/ms4w/geosphere/client1/Carte3_IA.map&amp;

5) now double click your new WMS connection. QGIS crashes 100% of the time repeatably for me when doing this (about 6 times so far to write this).

If you don't do step 3, then step 5 works absolutely fine. It's only when they're together it crashes.

elpaso:
The crash was caused by duplicated paths in data items.

While in theory identifiers should be unique, a broken
getCapabilities might serve duplicated (or empty)
identifiers.

Fixes #31927

### DIFF
--- a/src/gui/qgsbrowsertreeview.cpp
+++ b/src/gui/qgsbrowsertreeview.cpp
@@ -143,7 +143,7 @@ bool QgsBrowserTreeView::treeExpanded( const QModelIndex &index )
 
 bool QgsBrowserTreeView::hasExpandedDescendant( const QModelIndex &index ) const
 {
-  if ( !model() )
+  if ( !model() || !index.isValid() )
     return false;
 
   for ( int i = 0; i < model()->rowCount( index ); i++ )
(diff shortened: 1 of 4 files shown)


elpaso:
Manual backport required, I'm adding the tag anyway.

src/gui/qgsbrowsertreeview.cpp
```diff
@@ -143,7 +143,7 @@ bool QgsBrowserTreeView::treeExpanded( const QModelIndex &index )
 
 bool QgsBrowserTreeView::hasExpandedDescendant( const QModelIndex &index ) const
 {
-  if ( !model() )
+  if ( !model() && !index.isValid() )
```
nyalldawson: Should this be or?

nyalldawson (approved):

src/gui/qgsbrowsertreeview.cpp
elpaso: Yes, of course, thanks!

PR #31963, opened 2019-09-23
fix false boolean not shown as so in attribute table

3nids:
If an attribute is of type boolean, QGIS does not display `false` in the attribute table while `true` is written.

image

image

before => after
image

***
also small astyle improvement for a better alignments of multilines

### DIFF
--- a/src/core/fieldformatter/qgscheckboxfieldformatter.cpp
+++ b/src/core/fieldformatter/qgscheckboxfieldformatter.cpp
@@ -18,6 +18,7 @@
 
 #include "qgscheckboxfieldformatter.h"
 #include "qgsapplication.h"
+#include "qgsvectorlayer.h"
 
 
 QString QgsCheckBoxFieldFormatter::id() const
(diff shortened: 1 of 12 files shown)


m-kuhn (approved):
LGTM

Would look even better to me with a test.

src/core/fieldformatter/qgscheckboxfieldformatter.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+QString QgsCheckBoxFieldFormatter::representValue( QgsVectorLayer *layer, int fieldIndex, const QVariantMap &config, const QVariant &cache, const QVariant &value ) const
+{
+  Q_UNUSED( layer )
+  Q_UNUSED( fieldIndex )
+  Q_UNUSED( cache )
+  Q_UNUSED( config )
+
+
+  if ( value.isNull() || !value.canConvert<bool>() )
```
nyalldawson: `!value.canConvert<bool>()`

Is this even possible, or is every qvariant convertible to a bool?

But -- more importantly -- shouldn't this formatter also consider the true/false values which can be set for a non-boolean field using the checkbox wrapper? Or should the raw values for non-boolean variant types be used instead (I think this is what I'd prefer).

nyalldawson:
Nice fix @3nids -- this has bothered me in the past too, so it's great to see it being addressed!

src/core/fieldformatter/qgscheckboxfieldformatter.cpp
3nids: > Is this even possible, or is every qvariant convertible to a bool?
good question. 

> But -- more importantly -- shouldn't this formatter also consider the true/false values which can be set for a non-boolean field using the checkbox wrapper? Or should the raw values for non-boolean variant types be used instead (I think this is what I'd prefer).

That's really correct this has to be handled as well, i.e. compare the values to the true/false configuration.

I am really not sure we should display the raw values. Let's consider that there is no boolean in SQLite for instance. I'd prefer have true/false than 0/1 written.
It would be more interesting to have configurable represent value but that's maybe over-engineering the problem and out-of-scope here.

m-kuhn: I also tend towards true/false(/NULL)
Other values should only ever be shown if there's a problem converting them to a boolean (let's say the configuration is for `t` and `f` and a `p` is in the field, then it show something like `(p)`)

nyalldawson: I like the idea of `(p)`. I'm definitely NOT a fan of silently mapping these across to NULL -- that's misleading.

m-kuhn (approved):

src/core/fieldformatter/qgscheckboxfieldformatter.cpp
3nids: ok this is done now.

```
if field type is bool:
  NULL => nullRepresentation
  true => tr("true")
  false => tr("false")
else
  if cannot convert to string (like json integer list) => (invalid)
  if == checkedstate => tr("true")
  if == uncheckedstate => tr("false")
  else (value.toString)
```

PR #8725, opened 2018-12-20
Correct mapFromSource method when reordering columns. fixes #20673

troopa81:
## Description

This PR corrects the issue https://issues.qgis.org/issues/20673
 When we reorder column (last column becomes first column for instance) with select on top activated, and we select an entire row the currentIndex is no longer the first column, and consequently it breaks edition.

## Checklist

- [x] Commit messages are descriptive and explain the rationale for changes
- [x] Commits which fix bugs include `fixes #11111` in the commit message next to the description
- [ ] Commits which add new features are tagged with `[FEATURE]` in the commit message
- [ ] Commits which change the UI or existing user workflows are tagged with `[needs-docs]` in the commit message and contain sufficient information in the commit message to be documented
- [x] I have read the [QGIS Coding Standards](https://docs.qgis.org/testing/en/docs/developers_guide/codingstandards.html) and this PR complies with them
- [ ] This PR passes all existing unit tests (test results will be reported by travis-ci after opening this PR)
- [x] New unit tests have been added for core changes
- [x] I have run [the `scripts/prepare-commit.sh` script](https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTING.md#contributing-to-qgis) before each commit

PR #31966, opened 2019-09-23
Fix move selected to top

### LINKED ISSUE #28914: Move selection to top doesn't worked when table is unsaved
Author Name: **Patrick Dunford** (@enzedrailmaps)
Original Redmine Issue: [21096](https://issues.qgis.org/issues/21096)
Affected QGIS version: 3.6.2

---
If changing data in a file based geopackage table.

Select a rown in the table, if the button to move selection to top is pressed, that record will not move to the top if there are records in the table that have not been saved.

---

- [Screenshot from 2019-01-29 16-14-10.png](https://issues.qgis.org/attachments/download/14206/Screenshot%20from%202019-01-29%2016-14-10.png) (Etienne Trimaille)

---

Related issue(s): #29035 (duplicates), #29166 (duplicates)
Redmine related issue(s): [21217](https://issues.qgis.org/issues/21217), [21348](https://issues.qgis.org/issues/21348)

---

m-kuhn:
Fix #28914

@troopa81 this is a partial revert of https://github.com/qgis/QGIS/pull/8725 .
According to the description of #8725 this was not part of the fix, but I'm not entirely sure. Do you remember why you changed these lines?

### DIFF
--- a/src/gui/attributetable/qgsattributetablefiltermodel.cpp
+++ b/src/gui/attributetable/qgsattributetablefiltermodel.cpp
@@ -250,11 +250,14 @@ void QgsAttributeTableFilterModel::setSelectedOnTop( bool selectedOnTop )
     Qt::SortOrder order = sortOrder();
 
     // set default sort values if they are not correctly set
-    if ( column < 0 || ( order != Qt::AscendingOrder && order != Qt::DescendingOrder ) )
-    {
-      sort( 0, Qt::AscendingOrder );
-      invalidate();
-    }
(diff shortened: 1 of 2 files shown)


nyalldawson:
Could you add a test for this? This feature breaks all the time...

troopa81:
@m-kuhn I don't remember exactly why I did this, it was not indeed the main part of the fix. I think it was to avoid calling sort when it doesn't need to (but I understand it needs to for some reason).

PR #31973, opened 2019-09-24
Respect user set default vector output format in vector split algorithm

nyalldawson:
The created outputs will now be created using the format specified from the Processing setting default vector format setting.


nyalldawson:
Alternative approach to #31910. 

@alexbruy what's your thoughts on this approach? As noted in #31910 I think we should eventually expose this option inside the processing algorithm dialog, along with other options to override the global settings on a per-execution run. But for now, this allows users to change the format and provides the underlying API required to respect this choice.

alexbruy (approved):
Looks good.

Probably at some point we can replace extensions with user-friendly format names. But this can be done later or even left as is.

SrNetoChan:
@nyalldawson thanks!!

PR #31974, opened 2019-09-24
[processing] Add API support for algorithm aliases

nyalldawson:
Allows us to freely move algorithms between providers without breaking existing scripts


alexbruy (approved):
Cool

Issue #25830, opened 2018-01-23
repository files modified by `make check`


qgib:
Author Name: **Sandro Santilli** (@strk)
Original Redmine Issue: [17934](https://issues.qgis.org/issues/17934)
Affected QGIS version: master
Redmine category:build/install
Assignee: Sandro Santilli

---
Here's a list of source files that are modified by running `make check`:

        modified:   tests/testdata/control_images/compositionconverter/expected_importComposerAtlas_0/expected_importComposerAtlas_0.png
        modified:   tests/testdata/control_images/compositionconverter/expected_importComposerTemplateAttributeTable_0/expected_importComposerTemplateAttributeTable_0.png
        modified:   tests/testdata/control_images/compositionconverter/expected_importComposerTemplateHtml_0/expected_importComposerTemplateHtml_0.png
        modified:   tests/testdata/control_images/compositionconverter/expected_importComposerTemplateLabel_0/expected_importComposerTemplateLabel_0.png
        modified:   tests/testdata/control_images/compositionconverter/expected_importComposerTemplateLegend_0/expected_importComposerTemplateLegend_0.png
        modified:   tests/testdata/control_images/compositionconverter/expected_importComposerTemplateMap_0/expected_importComposerTemplateMap_0.png
        modified:   tests/testdata/control_images/compositionconverter/expected_importComposerTemplateScaleBar_0/expected_importComposerTemplateScaleBar_0.png
        modified:   tests/testdata/control_images/compositionconverter/expected_importComposerTemplate_0/expected_importComposerTemplate_0.png
        modified:   tests/testdata/control_images/compositionconverter/expected_importComposerTemplate_1/expected_importComposerTemplate_1.png
        modified:   tests/testdata/polys_overlapping_with_id.dbf

Source files should be read-only, always

strk:
This is still an issue, as of commit 27d687eac594b74bf5e5832b27a330fb068f7a97

After `make check`, source code has these changes:

> M       tests/testdata/bug_17878.gpkg
> M       tests/testdata/polys_overlapping_with_id.dbf

elpaso:
@strk I can fix this: tests/testdata/bug_17878.gpkg

strk:
thanks @elpaso it would be one less :)

strk:
Still an issue in release-3_4, release-3_8 and master branches at the time of writing

strk:
Updated list of files modified by `make check`:
```
[...]

PR #31980, opened 2019-09-24
WIP: Do not override test files during testing

strk:
See #25830


strk:
As of this PR there are still modified files:
```
        modified:   ../python/plugins/processing/tests/testdata/custom/circular_strings.gpkg
        modified:   ../python/plugins/processing/tests/testdata/custom/pol.gpkg
        modified:   ../python/plugins/processing/tests/testdata/expected/merged_pol.gpkg
        modified:   ../resources/data/world_map.gpkg
        modified:   ../tests/testdata/polys_overlapping_with_id.dbf
        modified:   ../tests/testdata/provider/bug_17795.gpkg
```

strk:
I've added automated testing for having unmodified sources. The test uses `git` so only works when under a git repository. If not being in a git repository the current test still works, although emitting a possibly confusing error message...

strk:
Interesting, even the code layout changes sources !
https://travis-ci.org/qgis/QGIS/jobs/591915632#L941-L944

@jef any idea how to prevent that ?

python/testing/__init__.py
```diff
@@ -422,6 +422,14 @@ def start_app(cleanup=True):
         QGISAPP = QgsApplication(argvb, myGuiFlag)
 
         QGISAPP.initQgis()
+
+        # Make sure geopackages are not written-to, during tests
+        # See https://github.com/qgis/QGIS/issues/25830
+        # NOTE: this needs to happen _after_ QgsApplication::initQgis()
```
nyalldawson: it doesn't reset them -- it just correctly initializes the storage so that QgsSettings can be safely used. I think this comment can be dropped

nyalldawson (commented):
Looks nice to me!

Will this also catch files **added** in the src dir during tests? That frustrates me too...

python/testing/__init__.py
strk: If it is unsafe to call QgsSettings before ::initQgis maybe the constructor should throw an exception ? I think it's important for the comment to stay, maybe I'll change the wording, but it's useful information (I spent a long time to find out)

strk:
Files *added* to source dir are not handled by the current state of this PR. They also annoy me so once I get a green on first pass (no files changed!) I'll see how to deal with that part.

strk:
Note that the change is still not effective (I guess there might be some .cpp file failing to set that walForSqlite3 value...): https://travis-ci.org/qgis/QGIS/jobs/592114030#L1675-L1685

> 	modified:   ../resources/data/world_map.gpkg
> 	modified:   ../tests/testdata/curved_polys.gpkg
> 	modified:   ../tests/testdata/points_gpkg.gpkg
> 	modified:   ../tests/testdata/polys_overlapping_with_id.dbf
> 	modified:   ../tests/testdata/qgis_server/bug_gh30264_empty_layer_wrong_bbox.gpkg
> 	modified:   ../tests/testdata/qgis_server/test_project_wms_filter.gpkg
> 	modified:   ../tests/testdata/qgis_server/test_project_wms_grouped_layers.gpkg
> 	modified:   ../tests/testdata/qgis_server_accesscontrol/dimensions.gpkg
> 	modified:   ../tests/testdata/qgis_server_accesscontrol/geo.gpkg

strk:
Just a note (for the sake of beating CMake...): automake automatically defines a `distcheck` rule that errors out if source tree is modified during build...

strk:
The qgs_server data is probably modified by tests/src/python/test_qgsserver.py which does NOT subclass TestCase, thus doesn't get the walForSqlite3 setting. @elpaso does it make sense to have that test inherit from TestCase too ?

strk:
Sorry, I meant... it doesn't use `start_app` from qgis.testing. I'll try doing that

strk:
The `PyQgsServerWMSDimension` test ends up modifying `tests/testdata/qgis_server_accesscontrol/dimensions.gpkg` but it doesn't look it ever attempts to _write_ into that file. Debug lines added show that `qgis/walForSqlite3` is always false within QgsOgrProviderUtils::GDALOpenWrapper -- still that file is modified, what else could trigger that ?!

strk:
It looks like the test somehow ends up asking to open the file in read-write mode:
> 528: XXX walForSqlite is 0
528: XXX GDAL NOT setting WAL journal
528: XXX GDAL  OGR_SQLITE_JOURNAL is (null) before opening file /usr/src/qgis/qgis-no-override-testdata/tests/testdata/qgis_server_accesscontrol/di
mensions.gpkg for update

From the /src/core/qgsproject.h header I see there's no concept of opening a "project" in read-only mode, could that make sense to add ?

strk:
Forcing `walForSqlite3` results in failures like `Offline Editing Plugin" "database is locked"`.
I've understood that even w/out WAL sqlite could do read/write, by using external files ? 
Did I misunderstand, @rouault  ?

strk:
An OGR provider fails in a way that might reveal a bug in the test:
https://travis-ci.org/qgis/QGIS/jobs/593559402#L1365-L1385

Basically, this snippet of tests/src/python/test_provider_ogr_gpkg.py:

       # Commit changes
        cbk = ErrorReceiver()
        vl.dataProvider().raiseError.connect(cbk.receiveError)
        self.assertTrue(vl.commitChanges())
        self.assertIsNone(cbk.msg)


It passes the first assert (commitChanges( ) returns true) but does not pass the second (no error is printed).

Unless the OGR provider returned a false success from `.commitChanges()`, we should maybe tolerate "informational" messages ?

rouault:
> I've understood that even w/out WAL sqlite could do read/write, by using external files ?
> Did I misunderstand, @rouault ?

This is more complex than that. See point 2 at the beginning of https://www.sqlite.org/wal.html : "WAL provides more concurrency as readers do not block writers and a writer does not block readers. Reading and writing can proceed concurrently. "

strk:
I filed PR #32145 for the false success from .commitChanges()

strk:
> This is more complex than that. See point 2 at the beginning of https://www.sqlite.org/wal.html : "WAL provides more concurrency as readers do not block writers and a writer does not block readers. Reading and writing can proceed concurrently. "

During testing we don't care about "improved concurrency". I'm happy with readers blocking writers and writes blocking readers. Especially since writers should _copy_ the file elsewhere before writing to it. All I'm asking is whether use of WAL is _really_ optional, and things should just work (maybe more slowly, due to blocked readers or writers) when setting it OFF

strk:
Count of modified files got down to:
```
< 	modified:   tests/testdata/polys_overlapping_with_id.dbf
< 	modified:   tests/testdata/qgis_server/test_project_wms_grouped_layers.gpkg
< 	modified:   tests/testdata/qgis_server_accesscontrol/dimensions.gpkg
```

Lastest commit in this PR addresses the first one (polys_overlapping_with_id).
The other two will probably need setting the WAL variabel for qgis _server_.
Maybe @elpaso you have hints about that ?

elpaso:
> Count of modified files got down to:
> 
> ```
> < 	modified:   tests/testdata/polys_overlapping_with_id.dbf
> < 	modified:   tests/testdata/qgis_server/test_project_wms_grouped_layers.gpkg
> < 	modified:   tests/testdata/qgis_server_accesscontrol/dimensions.gpkg
> ```
> 
> Lastest commit in this PR addresses the first one (polys_overlapping_with_id).
> The other two will probably need setting the WAL variabel for qgis _server_.
> Maybe @elpaso you have hints about that ?

No idea what are you talking about ;) 

Just copy them into a temporary directory before running the test. 

Like it's done here: https://github.com/qgis/QGIS/blob/master/tests/src/python/test_qgsserver_api.py#L107

strk:
> Just copy them into a temporary directory before running the test.

But the test is only _reading_ those files, so it sounds a bit of an overkill to copy them over.
Rather, I'd like to instruct OGR NOT to use embedded WAL...

strk:
Travis confirms the [left-over modified files](https://travis-ci.org/qgis/QGIS/jobs/624297367#L2541-L2542) are:
> 	modified:   tests/testdata/qgis_server/test_project_wms_grouped_layers.gpkg
> 	modified:   tests/testdata/qgis_server_accesscontrol/dimensions.gpkg

I verified that runing `PyQgsServerWFS` test changes only the `test_project_wms_grouped_layers.gpkg` file.
(`ctest -R QgsServerWFS$ --output-on-failure`)

I verified that running PyQgsServerWFSDimension` test changes only the `dimensions.gpkg` file.
(`ctest -R QgsServerWMSDimension --output-on-failure`)

strk:
It turns out that even with useWal set to false (as printed by the OGR provider), the source file `tests/testdata/qgis_server/test_project_wms_grouped_layers.gpkg` is still modified when running the `QgsServerWFS` test. Maybe the server always opens files in read-write mode ?
```
579: XXXX defUseWal: 0
579: XXXX useWal (from QgsSettings): 0
```

elpaso:
@strk the solution is to copy the test file(s) to a temporary directory before each test.

strk:
@elpaso I'm keepign that solution as the last resort. For now I'm wondering if the server could avoid asking for read-write mode when opeining a layer, unless it really needs to ?

elpaso:
AFAIK the server does nothing special when loading layers.

strk:
Ok, I'll take the path to copy test data. If not else it will give us the CI guarding for source files not being modified. We can improve later, if we see a way

strk:
Now I'm getting `ERROR 1: sqlite3_exec(COMMIT) failed: database is locked` -- I suspect this is due to concurrent opening with WAL off of the same test file. I thought sqlite  would _block_ instead of erroring out.
If this is not the case, we just can NOT turn WAL off and can only ever copy the test datafiles. @rouault  does it happen to you that .commit can fail instead of block/wait ?

rouault:
> @rouault does it happen to you that .commit can fail instead of block/wait ?

yes, I don't think SQLite will ever block. It is up to the application to implement a retry logic, if wished, when it returns a locking error. That said, someone added in recent OGR version a call to sqlite3_busy_timeout(hDB, 5000) in the SQLite/GPKG driver per https://github.com/OSGeo/gdal/commit/a694e230393890e77b19369b9bb71d99bd5c7ecc, whose real effect I'm not completely sure to understand. But in your case, it seems more to be related to a SQLITE_LOCKED error than a SQLITE_BUSY, which are subtelty different according to https://www.sqlite.org/rescode.html#busy. Those aspects of concurrency haven't been explored deeply in the OGR side.

strk:
I'm seriously thinking to change strategy and do a HUGE up-front copy of the whole testsuite prior to run any test. It may be overkill but it's really frustrating to try at tackling each use everytime...

strk:
CI is reporting a segfault running the WMS server test (I think): https://travis-ci.org/qgis/QGIS/jobs/644423987#L2415 -- but the changes in this PR are only against python files, so not sure how that's possible. @elpaso can  you think of a reason why python would segfault ?

strk:
CI keeps getting segfaults. @rouault is it possible your recent null-checking work will help here ?

strk:
> @strk the solution is to copy the test file(s) to a temporary directory before each test.

@elpaso I decided to take this route. Now it's about deciding whether _each_ test should do that (the full copy of testdata) or whether it should be done once (faster, but possibly polluting cross-test). What's your opinion on this ?

elpaso:
@strk since now I think that the copy has been made inside each test case (at least for some of those) only in case they were going to modify the test file, so basically when the underlying data is GPKG/spatialite or when the test is modifying (editing) the data in some way.

My personal way to decide if a copy is needed is if I find the test data file changed (git status) after running a test.

A full copy of the data directory on each test seems overkill to me.

About cross-pollution, I don't think we have that problem now, because the test data are separate for each test, and when they are not a copy is already made.

But I may be wrong of course, I don't regularly run the full test suite locally, just the individual tests or a few of them.

strk:
`ctest -R invertedpolygon --output-on-failure` is modifying curved_polys.gpkg

strk:
`ctest -R qgis_processingtest$ --output-on-failure` changes `tests/testdata/points_gpkg.gpkg`

strk:
`output/bin/qgis_processingtest parameters`, in particular, is one of the processing tests that do modify `src/testdata/points_gpkg.gpkg` -- the C tests do not all invoke some centralized function from which we could set wal to false...

strk:
`output/bin/qgis_processingalgstest packageAlg` modifies `testdata/control_images/expected_raster_linedensity/linedensity_testcase1.tif.aux.xml` and `testdata/control_images/expected_raster_linedensity/linedensity_testcase2.tif.aux.xml`

strk:
With latest commit, which simply sets walForSqlite3 to off for QgsServer tests, this segfaults:
> ctest -R PyQgsServerApi

strk:
This gives some hints:

> valgrind python3 ${srcdir}/tests/src/python/test_qgsserver_api.py

```
==15073== Invalid read of size 8
==15073==    at 0x153025EF: QgsApplication::exitQgis() (qgsapplication.cpp:1271)
==15073==    by 0x361C1DDD: meth_QgsApplication_exitQgis (sip_corepart3.cpp:327284)
==15073==    by 0x50A8AE: UnknownInlinedFun (methodobject.c:234)
==15073==    by 0x50A8AE: UnknownInlinedFun (methodobject.c:294)
==15073==    by 0x50A8AE: call_function.lto_priv.1824 (ceval.c:4851)
==15073==    by 0x50C5B8: _PyEval_EvalFrameDefault (ceval.c:3335)
==15073==    by 0x509D47: UnknownInlinedFun (ceval.c:754)
==15073==    by 0x509D47: UnknownInlinedFun (ceval.c:4933)
==15073==    by 0x509D47: fast_function.lto_priv.2239 (ceval.c:4968)
==15073==    by 0x50AA7C: call_function.lto_priv.1824 (ceval.c:4872)
==15073==    by 0x50C5B8: _PyEval_EvalFrameDefault (ceval.c:3335)
==15073==    by 0x509D47: UnknownInlinedFun (ceval.c:754)
==15073==    by 0x509D47: UnknownInlinedFun (ceval.c:4933)
==15073==    by 0x509D47: fast_function.lto_priv.2239 (ceval.c:4968)
==15073==    by 0x50AA7C: call_function.lto_priv.1824 (ceval.c:4872)
==15073==    by 0x50C5B8: _PyEval_EvalFrameDefault (ceval.c:3335)
==15073==    by 0x508244: UnknownInlinedFun (ceval.c:754)
==15073==    by 0x508244: _PyEval_EvalCodeWithName.lto_priv.1836 (ceval.c:4166)
==15073==    by 0x509641: _PyFunction_FastCallDict (ceval.c:5084)
==15073==  Address 0x445d7ff0 is 400 bytes inside an unallocated block of size 2,272 in arena "client"
```

strk:
Avoiding the call to initQgis fixes the segfault. Next:
> ctest -R 'PyQgsServerWMSDimension$'
Modifies tests/testdata/qgis_server_accesscontrol/dimensions.gpkg

strk:
Orrible experience:

Running this does NOT modify the file:
> python3 ${srcdir}/tests/src/python/test_qgsserver_wms_dimension.py  TestQgsServerWMSDimension.test_wms_getmap_with_filter

Running this does NOT modify the file:
> python3 ${srcdir}/tests/src/python/test_qgsserver_wms_dimension.py TestQgsServerWMSDimension.test_wms_getmap_mix_values

Running this DOES modify the file:
> python3 ${srcdir}/tests/src/python/test_qgsserver_wms_dimension.py TestQgsServerWMSDimension.test_wms_getmap_mix_values TestQgsServerWMSDimension.test_wms_getmap_with_filter

@elpaso can you think of a reason why running each test by itself does NOT change `tests/testdata/qgis_server_accesscontrol/dimensions.gpkg` while running them both in the same process does ?

strk:
NOTE: order in which tests are run does not matter

strk:
worse: I merged the tests in those two functions into a single function, the function makes a total of 3 requests. Commenting out any of the 3 requests prevents changing the file.

strk:
isLocalGpkg is always 0 (false) in `src/core/providers/ogr/qgsogrprovider.cpp` so Wal would not be set to true. What other reason do we have for that gpkg to be changed by a GetMap request ? Is QgsServer using some other way to open a GPKG ? @elpaso ?

strk:
interesting enough: `chmod 600 tests/testdata/qgis_server_accesscontrol/dimensions.gpkg` prevents the file from being modified, and results in WAL not being used

strk:
Figured: the file is changed when QgsOgrProviderUtils::GDALOpenWrapper is called with bUpdate=1:

```
src/core/providers/ogr/qgsogrprovider.cpp:4028 : (GDALOpenWrapper) [2ms] XXXX GDALOpen called with bIsLocalGpkg=0 and bUpdate=1
src/core/providers/ogr/qgsogrprovider.cpp:4028 : (GDALOpenWrapper) [122ms] XXXX GDALOpen called with bIsLocalGpkg=0 and bUpdate=0
src/core/providers/ogr/qgsogrprovider.cpp:4028 : (GDALOpenWrapper) [128ms] XXXX GDALOpen called with bIsLocalGpkg=0 and bUpdate=0
```

Forcing bUpdate to false prevents modifying the file. I dunno why WAL is affected by that switch though (an OGR confusion? Ideas @rouault ?)

I'm not sure where the bUpdate value comes from, but I suspect it comes from QgsOgrProviderUtils::getLayer

nyalldawson:
@strk I think the path you're going down will eventually lead to madness... 😁 Have you considered just copying the whole test folder to a temp path instead?

strk:
> @strk I think the path you're going down will eventually lead to madness... grin Have you considered just copying the whole test folder to a temp path instead?

I'm already there. Yes, I've considered it, was thinking about it today: doing it once from CMake itself. Would also avoid writing the .shm etc. around.

Still, not understanding WHY those damn .GPKG keep being updated even if WAL is NOT requested smells like hiding dust under the carpet

strk:
qgis_distanceareatest fails on Travis but succeeds on my machine, what can the difference be ?
@nyalldawson this time I'm focused on just copying the damn test data outside of source dir, as you suggested in previous commit. Eventually I might revert all commits but the one copying data to build dir AND the one enforcing the check that source tree needs not be modified

strk:
Still important to reach as a goal. Some good commits in this PR (the testcase, possibly the global test files copy to build dir)

strk:
Still important to reach as a goal. Some good commits in this PR (the testcase, possibly the global test files copy to build dir)

strk:
Still relevant

midnight-dev:
@strk Tricky to follow over several months with plenty of commits. Were you able to figure out what was modifying the files? I'm also curious why testing failed on Travis but not on your system.

strk:
Sometimes files were changed by tests that were testing writing the
files. These case might have all been fixed in this PR already.

The rest are due to files being changed by sqlite itself when using
embedded WAL (embedded locks). This is well explained in the early
comments I think, long standing issue.

Differences between Travis and people's system is another big issue
that QGIS is having.

nyalldawson:
@strk can we leave this closed until it's ready for review/merge? The PR queue is still too long to be easily maintainable, and we need to avoid having open PRs for long-term wip items like this...

strk:
Ok, closing

PR #31986, opened 2019-09-24
[Bugfix] enabling spatialite gpkg amphibious mode

### LINKED ISSUE #27914: GeoPackage SQL issues in DB Manager
Author Name: **Antoine Fonoxo** (Antoine Fonoxo)
Original Redmine Issue: [20092](https://issues.qgis.org/issues/20092)
Affected QGIS version: 3.5(master)
Redmine category:db_manager

---
DB Manager issues with GPKG geometries on perfectly valid GeoPackage files (made by QGIS or ogr2ogr).
The following SQL queries fail in DB Manager
( but succeed in DB Browser 4 SQLite with EnableGpkgAmphibiousMode ) :

SELECT AsWKT(geom) FROM mypolygon;
--> returns NULL ; should return: MULTIPOLYGON(...

SELECT ST_Area(geom) FROM mypolygon;
--> returns NULL ; should return numeric

SELECT isvalid(geom) FROM mypolygon;
--> returns -1 (which means: UNKNOWN) ; should return 0 or 1

roya0045:
## Description
proposal to fix #27914 

@rouault @elpaso approval would be appreciated

## Checklist

### DIFF
--- a/python/utils.py
+++ b/python/utils.py
@@ -677,6 +677,13 @@ using the "mod_spatialite" extension (python3)"""
             break
     if not found:
         raise RuntimeError("Cannot find any suitable spatialite module")
+    if any(['.gpkg' in arg for arg in args]):
+        try:
+            cur.execute("SELECT EnableGpkgAmphibiousMode()")
+        except (sqlite3.Error, sqlite3.DatabaseError, sqlite3.NotSupportedError):
+            QgsMessageLog.logMessage(u"warning:{}".format("Could not enable geopackage amphibious mode"),
+                                     QCoreApplication.translate("Python", "Python warning"))
+
     cur.close()
     con.enable_load_extension(False)
     con.create_function("regexp", 2, fcnRegexp)
--- a/tests/src/python/test_db_manager_gpkg.py
+++ b/tests/src/python/test_db_manager_gpkg.py
@@ -24,6 +24,8 @@ from qgis.testing import start_app, unittest
 from plugins.db_manager.db_plugins import supportedDbTypes, createDbPlugin
 from plugins.db_manager.db_plugins.plugin import TableField
 
+from utilities import unitTestDataPath
+
 
 def GDAL_COMPUTE_VERSION(maj, min, rev):
     return ((maj) * 1000000 + (min) * 10000 + (rev) * 100)
@@ -442,6 +444,22 @@ class TestPyQgsDBManagerGpkg(unittest.TestCase):
         #     table = tables[i]
         #     info = table.info()
 
+        connection.remove()
+
+    def testAmphibiousMode(self,):
+        connectionName = 'geopack1'
+        plugin = createDbPlugin('gpkg')
+        uri = QgsDataSourceUri()
+        test_gpkg = os.path.join(os.path.join(unitTestDataPath(), 'provider'), 'test_json.gpkg')
+
+        uri.setDatabase(test_gpkg)
+        plugin.addConnection(connectionName, uri)
+        connection = createDbPlugin('gpkg', connectionName)
+        connection.connect()
+        db = connection.database()
+        res = db.connector._execute(None, "SELECT St_area({}) from foo".format(db.tables()[0].fields()[1].name))
+        results = [row for row in res]
+        self.assertEqual(results, [(215229.265625,), (247328.171875,), (261752.78125,), (547597.2109375,), (15775.7578125,), (101429.9765625,), (268597.625,), (1634833.390625,), (596610.3359375,), (5268.8125,)])
         connection.remove()
 
 


m-kuhn:
Sounds like a good idea to me.
Can you raise an exception instead of using print?

python/utils.py
```diff
@@ -677,6 +677,12 @@ def fcnRegexp(pattern, string):
             break
     if not found:
         raise RuntimeError("Cannot find any suitable spatialite module")
+    if any(['.gpkg' in arg for arg in args]):
+        try:
+            cur.execute("SELECT EnableGpkgAmphibiousMode()")
+        except Exception as e:
+            print(e)
+            raise Exception("Could not enable geopackage amphibious mode")
```
m-kuhn: - Does it make sense to catch and re-raise here? What's the rationale behind this?
- Don't `print` the exception. The python console is not a place to report errors. If that needs to be logged, use a builtin error reporting mechanism. But since this is in `utils.py` we don't really know if it's going to be used in desktop / standalone / server, so maybe it's the wrong place anyway for any kind of logging / reporting and instead that should be done on a different level and here we only raise an exception (or let the exception go through).

python/utils.py
```diff
@@ -677,6 +677,12 @@ def fcnRegexp(pattern, string):
             break
     if not found:
         raise RuntimeError("Cannot find any suitable spatialite module")
+    if any(['.gpkg' in arg for arg in args]):
+        try:
+            cur.execute("SELECT EnableGpkgAmphibiousMode()")
+        except Exception as e:
+            print(e)
```
Gustry: Use QgsMessageLog IMHO

python/utils.py
roya0045: The issue is that enabling this function is nice and can be useful, people might even depend on it for some things but on the other hand I don't want to break the connection when this cannot be enabled. Yet I want the user to know when this could not be enabled. 
Since this might be used as a standalone or in the app I'm not sure how to make a call that will be properly handled.

roya0045: I'll just settle for a warn.

python/utils.py
```diff
@@ -677,6 +677,11 @@ def fcnRegexp(pattern, string):
             break
     if not found:
         raise RuntimeError("Cannot find any suitable spatialite module")
+    if any(['.gpkg' in arg for arg in args]):
+        try:
+            cur.execute("SELECT EnableGpkgAmphibiousMode()")
+        except:
+            warnings.warn("Could not enable geopackage amphibious mode")
```
m-kuhn: Can you align it with https://github.com/qgis/QGIS/blob/3be17e5192898476a0e385fc3f3c331ac1d50512/python/utils.py#L69 please?

roya0045: Sure thing, though I changed the formatwarning to simply using warning, I doubt it was necessary to write to a file, but let me know if I should use formatwarning instead.

m-kuhn: Can you do an experiment for me on the console please?

```
test = warnings.warn('Repeat with me: I will stick to the standard way of doing things in QGIS and prefer the way existing code is doing it.')
print(test)
```

Thank you.

python/utils.py
```diff
@@ -677,6 +677,13 @@ def fcnRegexp(pattern, string):
             break
     if not found:
         raise RuntimeError("Cannot find any suitable spatialite module")
+    if any(['.gpkg' in arg for arg in args]):
+        try:
+            cur.execute("SELECT EnableGpkgAmphibiousMode()")
+        except:
```
m-kuhn: Final thing, can you catch specific exceptions here? Catch all exceptions tend to give misleading error messages.

nyalldawson:
If I'm interpreting this PR correctly -- it ONLY applies to python code, right?

m-kuhn:
That is correct

nyalldawson:
Hmmm - isn't this potentially confusing? E.g. some functions will work in db manager, but the same won't work in other areas of QGIS. (e.g. setting a layer filter).

m-kuhn:
A valid point. I'm not sure what would be involved to change everything.

elpaso:
Maybe better to add that logic into https://github.com/qgis/QGIS/blob/master/src/core/providers/ogr/qgsgeopackageproviderconnection.cpp#L273 and use that method for all operations on gpkg instead?

roya0045:
> Maybe better to add that logic into https://github.com/qgis/QGIS/blob/master/src/core/providers/ogr/qgsgeopackageproviderconnection.cpp#L273 and use that method for all operations on gpkg instead?

If this is the case is the python code still needed?

m-kuhn:
Can you answer the question please, @roya0045 ?

Pushing untested code is taking advantage of reviewer resources, users time and putting the reputation of the QGIS application at risk. I know you had troubles with your dev environment at some point. I hope you could iron out those issues and ask you to thoroughly test before pushing things for review!

roya0045:
@m-kuhn I didn't have the time to test it, I'm currently relying on Travis. I'm still trying to iron out build issues on windows. Using a windows vm is proving more problematic than anything else but I am force to use it and I always get issues with qt creator or the final exe made with Visual Studio.

m-kuhn:
Can you focus on your dev environment first?
Anything else is just a loss of time.

m-kuhn:
Can you focus on your dev environment first?
Anything else is just a loss of time for everyone, you, us, QGIS.

roya0045:
@m-kuhn the newest cmake solved all my build issues, though I am focusing on the bad layer handler feature. I will report back when I have tested this PR.

roya0045:
@m-kuhn When I load my compiled qgis I get the 'couldn't load DLL' error when qgis initiate PyQt5.QtCore. I'm not sure what might be causing this as my python path contains the same element as when I use python 3.7 directly in osgeo. 

I have tried to use py_compile but I get the 'No TARGET ' error. I'll try to get it solved but in the meantime, do you know what might fix this?

m-kuhn:
I would write to the devel list and include the complete error. No guarantee that there is an answer, as build errors are often system dependemt. But once in a while it rings a bell somewhere and you get the right hints.

roya0045:
@m-kuhn

I had to build at home to prevent the sip issue but here is the result in db manager:
  Before: (3.4.4 win 10)
bfore
After: (master + patch, ubuntu)
aftr

roya0045:
@m-kuhn @elpaso Judging from the test, the function seems to work, at least in DB manager. I'm not too sure how to test st_area on layers in the core. But the message did not pop when loading a layer and since both call are the same it seems fine.

Should be good for a final review.

roya0045:
@elpaso since db manager still uses spatialite_connect, was there a plan to now rely on the ogr connector? I didn't see any sip bindings so I assume that the current way will stay for some time.

roya0045:
Tried to use triggers, didn't work, the trigger gets set but when updating or inserting features I get an error that OGR couldn't commit.  I'm not sure of the issue come from the fact that the trigger might not be transferable from the db connection to the qgis connection or if the amphibious mode was not properly enabled in the c++ code.

roya0045:
> Maybe better to add that logic into https://github.com/qgis/QGIS/blob/master/src/core/providers/ogr/qgsgeopackageproviderconnection.cpp#L273 and use that method for all operations on gpkg instead?

I had the time to do more test and assume that this method was used to manage connections with gpkg but I was provent wrong. It seems that this is not called at all when dealing with a geaopackage as ogr seem to handle everything.

Would you recommend to add the function somewhere in the ogr provider or call the function after the layer is loaded by ogr?

This is mostly to support extended triggers.

roya0045:
@rouault Even, would you happen to know if amphibious mode could conflict with journal mode and some other existing behavior? I'm getting weird strange results since amphibious mode is activated from the c++ connection.

roya0045:
changes to symbology files slipped in when reseting, should fix when I have better internet.

roya0045:
@rouault Out of curiosity, is it possible that calling the enable amphibious mode or something related to this might create additional connections?
This might explain why there are 71 connections counted and why journal mode cannot be enabled. Though I assumed that calling the function only enabled the behavior, but I am not sure where this is stored and my current way might set up multiple connection and not leverage the amphibious mode in the c++ code.

roya0045:
Yeah cpp code need to be reworked, I need to find a way to grab the existing connection and use the function on that, otherwise I'm just spawning parallel connections that don't get used.

src/core/providers/ogr/qgsgeopackageproviderconnection.cpp
```diff
@@ -263,6 +265,21 @@ void QgsGeoPackageProviderConnection::setDefaultCapabilities()
 #endif
 }
 
+void QgsGeoPackageProviderConnection::enableAmphibiousMode( const QString &uri )
```
rouault: I doubt this method does anything useful. The SQL execute here will only be valid during the lifetime of the SQLite connectino, that is the gdal::ogr_datasource_unique_ptr hDS object, which is destroyed at the end of this method.

roya0045: Yeah what I noticed was that this made new connections to the gpkg, that connection had amphibious mode activated but this was not the main connection where things occured. I initially made this since I was not too sure how to catch an existing connection and then call the desired function. I was hoping that it would still work somehow but alas I was wrong.

I am not too sure how connections are represented, store and how a query can be passed to them as I could not find such examples in this or the ogr provider connection. I'll have to dig further and possibly change more things to do what I aimed to do.

roya0045: @rouault what confuses me is the fact that I replicated in part executeGDalSqlPrivate. I'll try with the connection release.

roya0045: Ok with release connection it removes most issues, though I am left with 1 extra connection, which is confusing. I also need to test the behaviour but this will be hard to do as python is broken (and so are plugins) in my builds...

roya0045:
@rouault After releasing the resulting layer I can drop down the connection numbers but I am still stuck with an extra connection and I don't want calls to be made to that connection erroneously when using spatialite functions (and to make the test pass). This should be unlikely as my test pass but I would like to solve it.

Do you have any idea what might cause this extra connection? Or should it be safe to alter the failing test?

src/core/providers/ogr/qgsogrprovider.cpp
```diff
@@ -4418,6 +4418,9 @@ void QgsOgrProvider::open( OpenMode mode )
 
     QgsDebugMsg( "OGR opened using Driver " + mGDALDriverName );
 
+    if ( mGDALDriverName == "GPKG" )
+      QgsGeoPackageProviderConnection::enableAmphibiousMode( mFilePath.split( '|' ).at( 0 ) );
```
rouault: I don't see the need for this. This is already done in the OGR GPKG driver per https://github.com/OSGeo/gdal/blob/5870c3fc8ccc05ead6b9c8a5dd9fad76969fd890/gdal/ogr/ogrsf_frmts/gpkg/ogrgeopackagedatasource.cpp#L6732

roya0045: Ah great! I just thought it wasn't enabled anywhere, I'll just remove the c++ code, leave the test and the python code. Though when I made a trigger with db manager it would simply break in qgis, and it thought this might solve this.

roya0045: Welp it works fine without, maybe its simply that between the dbmanager trigger and the use in QGIS the trigger changed or got removed.

roya0045:
@elpaso the c++ code was not needed, only db manager needed some help. Care to merge this?

roya0045:
@m-kuhn less code was needed, thankfully. Only DB manager was changed, should be ok for merge when you have some free time.

tests/src/python/test_db_manager_gpkg.py
```diff
@@ -444,6 +444,32 @@ def testAllGeometryTypes(self):
 
         connection.remove()
 
+    def testAmphibiousMode(self,):
+        connectionName = 'geopack1'
+        plugin = createDbPlugin('gpkg')
+        uri = QgsDataSourceUri()
+        test_gpkg = os.path.join(self.basetestpath, 'test_amphibiousGpkg.gpkg')
```
rouault: can't we just use one of the existing .gpkg files like tests/testdata/points_gpkg.gpkg ?

roya0045: Possibly but since all the other tests in this file use the same method and since this does not depend on a common source I felt that this was more appropriate. I'm open to change if needed.

rouault: I believe other tests need to create a file to test a particular aspect. Here you can use any .gpkg file and use any ST_ function from Spatialite to demonstrate it's working as expected. The less code of lines, the better

roya0045: Strangely the test fail, maybe its because the polygons are using curved lines.

roya0045: Yeah curved features could not be processed, test is fine now.

roya0045:
@rouault Thanks for the review, let me know if there is anything else.

rouault:
LGTM. Could you just amend your commit message to someting more explicit like: "spatialite_connect(): enable Spatialite 'amphibious' mode for GeoPackage to be able to use Spatialite functions"

roya0045:
Done, I tend to forget this step with all the commit merging that goes on.

roya0045:
@rouault Thank your for the assistance Even!

PR #31990, opened 2019-09-24
[processing] Extend api for retrieving a layer in a compatible format

nyalldawson:
A few releases ago a bug fix was implemented which forced conversions of multi-layer sources. This was a valid bug fix, but the consequence was that any algorithm using this api with a source file containing multiple layers (e.g. gpkg) performed a complete copy of the target layer to a new file, severely impacting performance.

This commit adds new API to retrieve a compatible layer path in the case when an algorithm CAN correctly handle specific target layer names. In this case, the forced copy of the source layer is avoided when using multi-layer inputs like geopackage.

Allows fix for performance regression in processing R provider


nyalldawson:
@alexbruy look ok to you?

alexbruy (approved):
Nice improvement.

PR #31991, opened 2019-09-24
Various conditional formatting fixes

nyalldawson:
Biggest fix here is the repair of the broken display of conditional styles in the feature list view (on the side of the feature form). The underlying code was all in place and working correctly, but the delegate used to paint the list wasn't respecting it.


nirvn (approved):
Nice set of fixes / cleanups.

nyalldawson:
Thanks @nirvn!

PR #31992, opened 2019-09-24
Bulk astyle update

nyalldawson:
(I'm personally not a huge fan of this change, but gotta get the CI passing again!)

### DIFF
--- a/scripts/astyle.options
+++ b/scripts/astyle.options
@@ -7,7 +7,7 @@
 --indent-labels
 --indent-namespaces
 --indent-switches
---max-instatement-indent=80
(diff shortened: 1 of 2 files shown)


nyalldawson:
Wow - big breakage with this one. I'm going to revert the original astyle change for now @3nids until we can get it passing on the whole codebase.

nyalldawson:
@3nids I tried in this PR originally to do the bulk update, but there was a LOT of failures in the result. That may be due to my version of astyle vs the one on travis, but it was gonna be a good hour or so of manually stuffing around to fix.

Just a warning in advance if you head into this work 😉

PR #32001, opened 2019-09-25
Add active python plugins in the about dialog

nirvn:
3.4 backport of a commit to help getting more accurate issue reports.

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -4490,6 +4490,15 @@ void QgisApp::about()
     versionString += "</tr><tr><td colspan=4>" + tr( "This copy of QGIS writes debugging output." ) + "</td>";
 #endif
 
+#ifdef WITH_BINDINGS
+    if ( mPythonUtils && mPythonUtils->isEnabled() )
+    {
+      versionString += "</tr><tr><td>" + tr( "Active python plugins" ) + "</td><td colspan=3>";
+      const QStringList activePlugins = mPythonUtils->listActivePlugins();
+      versionString += activePlugins.join( "; <br>" ) + "</td>";
+    }
+#endif
+
     versionString += QLatin1String( "</tr></table></div></body></html>" );
 
     sAbt->setVersion( versionString );
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -4428,74 +4428,68 @@ void QgisApp::about()
     sAbt = new QgsAbout( this );
     QString versionString = QStringLiteral( "<html><body><div align='center'><table width='100%'>" );
 
-    versionString += QLatin1String( "<tr>" );
-    versionString += "<td>" + tr( "QGIS version" )       + "</td><td>" + Qgis::QGIS_VERSION + "</td><td>";
-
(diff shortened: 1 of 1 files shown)


src/app/qgisapp.cpp
```diff
@@ -4490,6 +4490,15 @@ void QgisApp::about()
     versionString += "</tr><tr><td colspan=4>" + tr( "This copy of QGIS writes debugging output." ) + "</td>";
 #endif
 
+#ifdef WITH_BINDINGS
+    if ( mPythonUtils && mPythonUtils->isEnabled() )
+    {
+      versionString += "</tr><tr><td>" + tr( "Active python plugins" ) + "</td><td colspan=3>";
```
m-kuhn: QLatin1String?

nirvn: Code behaves like the other pre existing string appends in this function. Should all of those be moved to qlatin1string?

m-kuhn: Oh, I've only seen the one at line 5002 which happened to have qlatin1string already.

Yes, I think we should almost always have QLatin1String / QStringLiteral, so if you can also change them, perfect.

nirvn: Will do.

src/app/qgisapp.cpp
```diff
@@ -4490,6 +4490,15 @@ void QgisApp::about()
...
+      const QStringList activePlugins = mPythonUtils->listActivePlugins();
+      versionString += activePlugins.join( "; <br>" ) + "</td>";
```
m-kuhn: Actually, what's the `;` for? I think they should either be `;` separated or new line separated

nirvn: The ;s are there to separate plugins when pasted in github (the html to markdown parsing kills new lines). 

New lines are because it makes for an easier list to visually glance over in the about dialog.

m-kuhn (approved):

src/app/qgisapp.cpp
```diff
@@ -4428,66 +4428,69 @@ void QgisApp::about()
...
     {
-      versionString += tr( "QGIS code branch" ) + QStringLiteral( "</td><td><a href=\"https://github.com/qgis/QGIS/tree/release-%1_%2\">Release %1.%2</a></td>" )
-                       .arg( Qgis::QGIS_VERSION_INT / 10000 ).arg( Qgis::QGIS_VERSION_INT / 100 % 100 );
+      versionString += QStringLiteral( "%1</td><td><a href=\"https://github.com/qgis/QGIS/tree/release-%1_%2\">Release %1.%2</a></td>" )
+                       .arg( tr( "QGIS code branch" ) ).arg( Qgis::QGIS_VERSION_INT / 10000 ).arg( Qgis::QGIS_VERSION_INT / 100 % 100 );
     }
     else
     {
-      versionString += tr( "QGIS code revision" ) + QStringLiteral( "</td><td><a href=\"https://github.com/qgis/QGIS/commit/%1\">%1</a></td>" ).arg( Qgis::QGIS_DEV_VERSION );
+      versionString += QStringLiteral( "%1</td><td><a href=\"https://github.com/qgis/QGIS/commit/%1\">%1</a></td>" ).arg( tr( "QGIS code revision" ) ).arg( Qgis::QGIS_DEV_VERSION );
```
Gustry: Hi, I think this broke the link to GitHub in QGIS 3.4 about the commit revision number.

PR #32015, opened 2019-09-26
allow to use remote content for init code file path in attribute form

3nids:
documentation PR: https://github.com/qgis/QGIS-Documentation/pull/4261


elpaso:
I'm a little worried about the security implications here, am I getting it right that we are potentially downloading and executing untrusted python code from a remote URL?
Shouldn't it be an opt-in setting?

3nids:
can't you basically do this from within any python code ? like in the attribute form open?

elpaso:
True, the only difference is that when python code is stored directly in the project you can immediately see it and check for malicious code, the potential issue I see here is that the remote code could be changed without the user knowing. 

Can you actually see the code that is executed here? 

Is it downloaded and shown in the python code textbox? 

If yes, I see no problems.

Anyway, I'm not against this PR, I just wanted to be sure about the potential security implications, I know that some customers are very sensitive about this kind of potential vulnerabilities.

Perhaps we should add a page on the documentation about the potential security issues in QGIS (starting from the python plugins, down to the other parts of the code where arbitrary Python code execution is possible).

3nids:
> Can you actually see the code that is executed here?
> Is it downloaded and shown in the python code textbox?

No, similarly to when using a file path, the code is loaded and run.


> Perhaps we should add a page on the documentation about the potential security issues in QGIS

This sounds reasonable. Adding a guard to Python form similarly to the project macros sounds like a good addition too (and would make the latter a bit more useful).

nyalldawson:
I'm with @elpaso here -- running any python code without user acceptance makes me very nervous. I think both this and the existing form code should be blocked behind the existing "enable project macros" option and not run until that is accepted. That's the minimal effort we can do to respect our user's security...

m-kuhn:
Yup, I'm also on the same page. That has been bugging me too for a while!

Given that this is an existing issue and any python code in forms was able to download and execute remote code since the dawn of time, it's an issue unrelated to this particular PR.

I'd propose to open a new issue and collect places suffering from this (forms and actions. Maybe custom models too?)

3nids:
this is adressed in #32095

PR #32017, opened 2019-09-26
[locator] allow to trigger locator search from the API

3nids:


src/gui/qgisinterface.h
```diff
@@ -970,6 +970,14 @@ class GUI_EXPORT QgisInterface : public QObject
      */
     virtual void preloadForm( const QString &uifile ) = 0;
 
+    /**
+     * This will perforom a search in the locator bar
```
nyalldawson: "This will perform a search in the locator bar by setting the line edit text to \a searchText and automatically displaying any  results."

nyalldawson (commented):
Looks good to me

PR #32030, opened 2019-09-27
Move page background setting to a per-page option

### LINKED ISSUE #25695: [Layout] Background color from page properties should not be applied to all the composition
Original Redmine Issue: [17799](https://issues.qgis.org/issues/17799)
Affected QGIS version: master

---
Create a multi page layout with different size of pages
Right click on one page and select "Page properties..."
Set a background color: it's applied to all pages. Given that the rest of the information is about one page (obvious in case of multi size pages - see #25694), it looks weird to have that one applied to all the pages. It should be applied to the selected page imho.

Maybe we can have a global setting or a checkbox near the option (better imho) to allow replication to other pages.

nyalldawson:
The UI for this setting was sitting under the page properties panel,
which led users to believe it was a per-page setting (rather than
applying to ALL pages in the layout).

Instead, move this property to sit within individual layout item pages
so that the behavior matches what the UI suggests.

Fixes #25695

### DIFF
--- a/python/core/auto_generated/layout/qgslayoutitempage.sip.in
+++ b/python/core/auto_generated/layout/qgslayoutitempage.sip.in
@@ -38,6 +38,7 @@ Item representing the paper in a layout.
 %Docstring
 Constructor for QgsLayoutItemPage, with the specified parent ``layout``.
 %End
+    ~QgsLayoutItemPage();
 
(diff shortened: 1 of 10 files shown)


nirvn (approved):
Good correction.

PR #32031, opened 2019-09-27
rename QgsLayerTreeRegistryBridge.InsertionPoint.parent to group

3nids:
because I find it confusing to do `insertionPoint.parent.insertLayer()`
better do `insertionPoint.group.insertLayer()`


nirvn:
Is it an acceptable API break?

3nids:
Yes, it is since 3.10, I added this last week

nirvn:
@3nids , hah, sorry :)

3nids:
you will rewrite twice by hand (on paper) the whole source code for not following the API closely enough!

PR #32036, opened 2019-09-27
[locator] Fix crash when all layer features locator searching against an invalid layer

nirvn:
## Description

This PR insures that invalid layers missing a data provider are not added to the list of prepared layers used to search for features in the locator. 

This fix is sponsored by QField.

## Checklist

### DIFF
--- a/src/app/locator/qgsinbuiltlocatorfilters.cpp
+++ b/src/app/locator/qgsinbuiltlocatorfilters.cpp
@@ -353,7 +353,7 @@ void QgsAllLayersFeaturesLocatorFilter::prepare( const QString &string, const Qg
   for ( auto it = layers.constBegin(); it != layers.constEnd(); ++it )
   {
     QgsVectorLayer *layer = qobject_cast< QgsVectorLayer *>( it.value() );
-    if ( !layer || !layer->flags().testFlag( QgsMapLayer::Searchable ) )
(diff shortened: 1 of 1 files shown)


src/app/locator/qgsinbuiltlocatorfilters.cpp
```diff
@@ -353,7 +353,7 @@ void QgsAllLayersFeaturesLocatorFilter::prepare( const QString &string, const Qg
   for ( auto it = layers.constBegin(); it != layers.constEnd(); ++it )
   {
     QgsVectorLayer *layer = qobject_cast< QgsVectorLayer *>( it.value() );
-    if ( !layer || !layer->flags().testFlag( QgsMapLayer::Searchable ) )
```
3nids: shouldn't we use isValid instead?

nirvn: If isValid() does a data provider check, then we could. The problem here is that the creation of a new vector layer feature source blindly assumes that the passed on vector layer has a data provider: https://github.com/qgis/QGIS/blob/96dd42b1501d1d139ca6ef24fa2a6f43df79047b/src/core/qgsvectorlayerfeatureiterator.cpp#L34

3nids: ok

3nids (approved):

PR #9734, opened 2019-04-06
Qgis as a namespace

3nids:
following discussion in #9437 

making Qgis a namespace instead of a class.
We could define general enums in there (like map layer type, or message level which is already there) AND use forward declarations.

I could not have the value defined in the source file and using `extern`. This goes a bit beyond my understanding of compilation and I am not sure about the implications.

some reference: https://stackoverflow.com/questions/42080571/c-referencing-extern-const-within-a-namespace

PR #32039, opened 2019-09-27
use an enum for enable macros possibilities

3nids:
This creates a scoped enum as `Qgis::PythonMacroMode`

for whoever is interested, there is no implicit casting to int for scoped enums.
This is why I introduced static casting when using `QMetaEnum::valueToKey`.

Since this is fixed, we could move the enums to `Qgis` class as scoped enums. We would/could have `Qgis::LayerType::VectorLayer`. 

For reference to the topic: https://github.com/qgis/QGIS/pull/9734
Last question is, can we forward declare by doing `namespace Qgis {class LayerType;}`. But that's out of scope for this PR.


3nids:
Superseded by #32095

PR #32043, opened 2019-09-27
Fix crash when opening symbology of invalid layer

### LINKED ISSUE #32041: When opening the symbology properties of an invalid layer, QGIS crashes
on master

m-kuhn:
Fix #32041

### DIFF
--- a/src/gui/symbology/qgscategorizedsymbolrendererwidget.cpp
+++ b/src/gui/symbology/qgscategorizedsymbolrendererwidget.cpp
@@ -578,7 +578,7 @@ QgsCategorizedSymbolRendererWidget::QgsCategorizedSymbolRendererWidget( QgsVecto
   advMenu->addAction( tr( "Match to Saved Symbols" ), this, SLOT( matchToSymbolsFromLibrary() ) );
   advMenu->addAction( tr( "Match to Symbols from File…" ), this, SLOT( matchToSymbolsFromXml() ) );
   advMenu->addAction( tr( "Symbol Levels…" ), this, SLOT( showSymbolLevels() ) );
-  if ( mCategorizedSymbol->type() == QgsSymbol::Marker )
+  if ( mCategorizedSymbol && mCategorizedSymbol->type() == QgsSymbol::Marker )
   {
     QAction *actionDdsLegend = advMenu->addAction( tr( "Data-defined Size Legend…" ) );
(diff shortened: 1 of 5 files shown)


src/gui/symbology/qgscategorizedsymbolrendererwidget.cpp
```diff
@@ -535,8 +535,11 @@ QgsCategorizedSymbolRendererWidget::QgsCategorizedSymbolRendererWidget( QgsVecto
   }
 
   mCategorizedSymbol.reset( QgsSymbol::defaultSymbol( mLayer->geometryType() ) );
-  btnChangeCategorizedSymbol->setSymbolType( mCategorizedSymbol->type() );
-  btnChangeCategorizedSymbol->setSymbol( mCategorizedSymbol->clone() );
+  if ( mCategorizedSymbol )
+  {
+    btnChangeCategorizedSymbol->setSymbolType( mCategorizedSymbol->type() );
```
nyalldawson: Hmmmm, we'll still need to call this method somehow. The layer itself will usually have a wkb type available even when it's invalid. Can you use this to set the symbol type when mCategorizedSymbol doesn't exist?

m-kuhn: The symbol is created from the layer type, just one line before and the type is `UnknownGeometry` https://github.com/qgis/QGIS/pull/32043/files/5b12071026be7741e2426124624217cc136572c3#diff-bcfbc8c19bf3c403c220c39a7756db57R537

nyalldawson: Ah, missed that. So the crash must only be happening with an older project, before we started recording the wkb type alongside layers in projects....

nyalldawson: Anyway, +1 to merge. does graduated need the same fix?

m-kuhn: Yes, it's an older project.
And yea, apparently graduated too

m-kuhn: Just fixed whatever crash I could trigger.

m-kuhn:
If a crash doesn't warrant a merge during hard freeze... what does?

PR #32051, opened 2019-09-29
Improvements and big optimisations for QgsTessellator

nyalldawson:
This PR optimises QgsTesselator, by using the knowledge that the tesellator only operates on QgsPolygon geometries and accordingly all rings are QgsLineStrings. In this case we can use heavily optimised point getters and setters and avoid the generic, inefficient virtual methods.

It also adds a new constructor which takes a geometry bounds, and scales coordinates using these bounds. This avoids numerical instability with very small coordinate differences, e.g. when tesellating a geometry in geographic coordinates.

### DIFF
--- a/src/core/qgstessellator.cpp
+++ b/src/core/qgstessellator.cpp
@@ -31,6 +31,7 @@
 #include <QMatrix4x4>
 #include <QVector3D>
 #include <algorithm>
+#include <unordered_set>
 
 
 static void make_quad( float x0, float y0, float z0, float x1, float y1, float z1, float height, QVector<float> &data, bool addNormals )
@@ -215,6 +216,16 @@ static void _normalVectorToXYVectors( const QVector3D &pNormal, QVector3D &pXVec
   pYVector = QVector3D::normal( pNormal, pXVector );
 }
 
+struct float_pair_hash
+{
+  std::size_t operator()( const std::pair<float, float> pair ) const
+  {
+    std::size_t h1 = std::hash<float>()( pair.first );
+    std::size_t h2 = std::hash<float>()( pair.second );
+
+    return h1 ^ h2;
+  }
+};
 
 static void _ringToPoly2tri( const QgsLineString *ring, std::vector<p2t::Point *> &polyline, QHash<p2t::Point *, float> &zHash )
 {
@@ -225,16 +236,18 @@ static void _ringToPoly2tri( const QgsLineString *ring, std::vector<p2t::Point *
   const double *srcXData = ring->xData();
   const double *srcYData = ring->yData();
   const double *srcZData = ring->zData();
+  std::unordered_set<std::pair<float, float>, float_pair_hash> foundPoints;
+
   for ( int i = 0; i < pCount - 1; ++i )
   {
     const float x = *srcXData++;
     const float y = *srcYData++;
     const float z = *srcZData++;
 
-    const bool found = std::find_if( polyline.begin(), polyline.end(), [x, y]( p2t::Point *&p ) { return p->x == x && p->y == y; } ) != polyline.end();
-
-    if ( found )
+    auto res = foundPoints.insert( std::make_pair( x, y ) );
+    if ( !res.second )
     {
+      // already used this point, don't add a second time
       continue;
     }
 
--- a/src/core/qgstessellator.cpp
+++ b/src/core/qgstessellator.cpp
@@ -231,7 +231,7 @@ static void _ringToPoly2tri( const QgsLineString *ring, std::vector<p2t::Point *
     const float y = *srcYData++;
     const float z = *srcZData++;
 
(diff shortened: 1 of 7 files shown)


nirvn:
Looking good. Nice optimization.

src/core/qgstessellator.cpp
```diff
@@ -331,41 +387,55 @@ static bool _check_intersecting_rings( const QgsPolygon &polygon )
...
 }
 
 
 void QgsTessellator::addPolygon( const QgsPolygon &polygon, float extrusionHeight )
 {
-  const QgsCurve *exterior = polygon.exteriorRing();
+  const QgsLineString *exterior = qgsgeometry_cast< const QgsLineString * >( polygon.exteriorRing() );
 
-  const QVector3D pNormal = _calculateNormal( exterior, mOriginX, mOriginY, mInvertNormals );
+  const QVector3D pNormal = !mNoZ ? _calculateNormal( exterior, mOriginX, mOriginY, mInvertNormals ) : QVector3D();
```
wonder-sk: I am not sure I understand this change... If there's no Z value, we set normal to null vector? The normal vector should be a unit vector though, in case of no Z values pointing up...?

src/core/qgstessellator.cpp
```diff
@@ -412,9 +481,13 @@ void QgsTessellator::addPolygon( const QgsPolygon &polygon, float extrusionHeigh
     const QgsPoint ptStart( exterior->startPoint() );
     const QgsPoint pt0( QgsWkbTypes::PointZ, ptStart.x(), ptStart.y(), std::isnan( ptStart.z() ) ? 0 : ptStart.z() );
 
+    const float scaleX = !mBounds.isNull() ? 10000.0 / mBounds.width() : 1.0;
+    const float scaleY = !mBounds.isNull() ? 10000.0 / mBounds.height() : 1.0;
```
wonder-sk: Is this arbitrarily chosen value to scale the coordinates to 0...10000 ?

src/core/qgstessellator.h
```diff
@@ -42,6 +43,17 @@ class CORE_EXPORT QgsTessellator
     //! Creates tessellator with a specified origin point of the world (in map coordinates)
     QgsTessellator( double originX, double originY, bool addNormals, bool invertNormals = false, bool addBackFaces = false );
 
+    /**
+     * Creates tessellator with a specified \a bounds of input geometry coordinates.
+     * This constructor allows the tessellator to map input coordinates to a desirable range for numerically
```
wonder-sk: s/numerically/numerical/ ?

wonder-sk (approved):
Really nice improvements all over the place! +1 to merge.

Just curious, have you run the tessellation somehow also with geographical coordinates? Because currently the 3D map view should try to stop you from using unprojected map CRS - various other bits of code do not play well (yet) with different scale of coordinates.

src/core/qgstessellator.cpp
nyalldawson: The no z mode is intended for purely 2d tessellation, where the normals aren't even used at all. But I could make this a unit vector if you'd rather?

src/core/qgstessellator.cpp
nyalldawson: Yes.. it's a magic number based on.... Nothing in particular. Maybe it should even be higher?

src/core/qgstessellator.h
nyalldawson: Yep

nyalldawson:
> Just curious, have you run the tessellation somehow also with geographical coordinates? Because currently the 3D map view should try to stop you from using unprojected map CRS - various other bits of code do not play well (yet) with different scale of coordinates

Yes, but outside of qgis 3d. I.e. the raw tessellation algorithm. And my ultimate goal here is to use tessellations for efficient and uniformally distributed random point in polygon generation (see https://observablehq.com/@scarysize/finding-random-points-in-a-polygon) (ultimately leading to a random marker fill pattern symbol layer type)

src/core/qgstessellator.cpp
wonder-sk: ah ok... all good then :-)

src/core/qgstessellator.cpp
wonder-sk: I was actually thinking originally to scale it to lower numbers... e.g. 0...1 but then realized it will likely not improve the numerical stability.

wonder-sk:
Cool - I guess that's going to be more efficient random point generator than just generating points in the whole extent and then discarding points lying outside of the polygon :)

src/core/qgstessellator.cpp
nyalldawson: That was my V1. But it still encountered the precision warnings and no results. I guess I'll bump it further if we hit issues...

PR #32064, opened 2019-09-30
Do not assume PostGIS 2.0+ has raster

strk:
People installing PostGIS via scripts do not necessarely install
raster support, even after PostGIS 2.0.
Starting with PostGIS 3.0 the raster support is not enforced as
active even when installing via extension.

This change detects raster support by the presence of a "raster"
named custom type.


elpaso (approved):

strk:
Sorry for the noise, 3.4 didn't have core support for postgis raster it looks like, so no backport needed

PR #32073, opened 2019-10-01
Add stable api for setting/retrieving project custom scales

nyalldawson:
Previously this was all done through raw project key manipulation, but there's a use case for exposing this nicely for use by plugins.


nirvn (approved):
Nice cleanup.

nyalldawson:
Thanks @nirvn!

PR #32081, opened 2019-10-01
 [needs-docs] revive the GDAL 'rasterize over' tool that was available in the old ltr and for some reason not activated/added in QGIS 3

### LINKED ISSUE #27298: GDAL rasterize - add support for "burn in" to existing rasters, 'add' parameter
Author Name: **Rijk Zuurmond** (Rijk Zuurmond)
Original Redmine Issue: [19470](https://issues.qgis.org/issues/19470)

Redmine category:processing/gdal
Assignee: Giovanni Manghi

---
missing -add function in GDAL_rasterize

---

- [after-gdal-rasterize.jpg](https://issues.qgis.org/attachments/download/13012/after-gdal-rasterize.jpg) (Rijk Zuurmond)
- [polylines.jpg](https://issues.qgis.org/attachments/download/13013/polylines.jpg) (Rijk Zuurmond)
- [qgis2.18_menu_GDAL-rasterize.jpg](https://issues.qgis.org/attachments/download/13014/qgis2.18_menu_GDAL-rasterize.jpg) (Rijk Zuurmond)

gioman:
This implementation/set of options for GDAL rasterize is very valuable for a number of GIS analysis (i.e. adding a physical barrier over a digital elevation model and then modelling its consequences).

I argue that this would need to be backported to LTR, as this tool was available in the old LTR.

fixes https://github.com/qgis/QGIS/issues/27298

### DIFF
--- a/python/plugins/processing/algs/gdal/rasterize_over.py
+++ b/python/plugins/processing/algs/gdal/rasterize_over.py
@@ -46,6 +46,7 @@ class rasterize_over(GdalAlgorithm):
     INPUT_RASTER = 'INPUT_RASTER'
     ADD = 'ADD'
     EXTRA = 'EXTRA'
+    BURN = 'BURN'
 
     def __init__(self):
         super().__init__()
@@ -64,7 +65,13 @@ class rasterize_over(GdalAlgorithm):
                                                       None,
                                                       self.INPUT,
                                                       QgsProcessingParameterField.Numeric,
-                                                      optional=False))
+                                                      optional=True))
+
+        self.addParameter(QgsProcessingParameterNumber(self.BURN,
+                                                       self.tr('A fixed value to burn'),
+                                                       type=QgsProcessingParameterNumber.Double,
+                                                       defaultValue=0.0,
+                                                       optional=True))
 
         add_param = QgsProcessingParameterBoolean(self.ADD,
                                                      self.tr('Add burn in values to existing raster values'),
@@ -108,6 +115,9 @@ class rasterize_over(GdalAlgorithm):
         if fieldName:
             arguments.append('-a')
             arguments.append(fieldName)
+        else:
+            arguments.append('-burn')
+            arguments.append(self.parameterAsDouble(parameters, self.BURN, context))
 
         if self.parameterAsBool(parameters, self.ADD, context):
             arguments.append('-add')
--- a/python/plugins/processing/algs/gdal/rasterize_over.py
+++ b/python/plugins/processing/algs/gdal/rasterize_over.py
@@ -21,16 +21,27 @@ __author__ = 'Alexander Bruy'
 __date__ = 'September 2013'
 __copyright__ = '(C) 2013, Alexander Bruy'
 
-import os
+# This will get replaced with a git SHA1 when you do a git archive
 
-from qgis.core import QgsProcessingUtils
+__revision__ = '$Format:%H$'
+
+import os
 
 from qgis.PyQt.QtGui import QIcon
 
+from qgis.core import (QgsRasterFileWriter,
+                       QgsProcessingParameterDefinition,
+                       QgsProcessingParameterFeatureSource,
+                       QgsProcessingParameterField,
+                       QgsProcessingParameterRasterLayer,
+                       QgsProcessingParameterNumber,
+                       QgsProcessingParameterString,
+                       QgsProcessingParameterEnum,
+                       QgsProcessingParameterExtent,
+                       QgsProcessingParameterBoolean,
+                       QgsProcessingParameterRasterDestination)
 from processing.algs.gdal.GdalAlgorithm import GdalAlgorithm
 from processing.algs.gdal.GdalUtils import GdalUtils
-
-from processing.tools import dataobjects
 
 pluginPath = os.path.split(os.path.split(os.path.dirname(__file__))[0])[0]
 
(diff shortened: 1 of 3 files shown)


python/plugins/processing/algs/gdal/rasterize_over.py
```diff
@@ -21,38 +21,59 @@
...
-        self.addParameter(ParameterRaster(self.INPUT_RASTER,
-                                          self.tr('Existing raster layer'), False))
+        self.units = [self.tr("Pixels"),
+                      self.tr("Georeferenced units")]
+
+        self.addParameter(QgsProcessingParameterFeatureSource(self.INPUT,
+                                                              self.tr('Input vector layer')))
+        self.addParameter(QgsProcessingParameterRasterLayer(self.INPUT_RASTER, self.tr('Input raster layer')))
+        self.addParameter(QgsProcessingParameterField(self.FIELD,
+                                                      self.tr('Field to use for a burn-in value'),
```
nyalldawson: "field to use for burn in value"

python/plugins/processing/algs/gdal/rasterize_over.py
```diff
@@ -21,38 +21,59 @@
...
+                                                      None,
+                                                      self.INPUT,
+                                                      QgsProcessingParameterField.Numeric,
+                                                      optional=False))
+        add_param = QgsProcessingParameterBoolean(self.ADD,
+                                                     self.tr('Adds the new value to the existing raster'),
```
nyalldawson: "Add burn in values to existing raster values"

nyalldawson (commented):
Nice! Can you please add unit tests for this too?

nyalldawson:
(also ping @alexbruy - can you remember if there was a particular reason this one wasn't originally ported?)

alexbruy (changes requested):
May I ask to add  (if applicable) "Additional command-line arguments" parameter like in other GDAL algorithms to avoid cluttering UI with new parameters when one will require something specific and not exposed in the current implementation

python/plugins/processing/algs/gdal/rasterize_over.py
```diff
@@ -21,38 +21,59 @@
...
+                       QgsProcessingParameterDefinition,
+                       QgsProcessingParameterFeatureSource,
+                       QgsProcessingParameterField,
+                       QgsProcessingParameterRasterLayer,
+                       QgsProcessingParameterNumber,
+                       QgsProcessingParameterString,
+                       QgsProcessingParameterEnum,
+                       QgsProcessingParameterExtent,
+                       QgsProcessingParameterBoolean,
+                       QgsProcessingParameterRasterDestination)
```
alexbruy: Seems this import is not used

alexbruy:
+1 for Nyall's request for tests.

@nyalldawson I don't remeber exactly why it was not ported. Maybe lack of time or something else.

python/plugins/processing/algs/gdal/rasterize_over.py
```diff
@@ -21,38 +21,59 @@
 __date__ = 'September 2013'
 __copyright__ = '(C) 2013, Alexander Bruy'
 
-import os
+# This will get replaced with a git SHA1 when you do a git archive
+
+__revision__ = '$Format:%H$'
```
Gustry: These lines should be removed.

python/plugins/processing/algs/gdal/rasterize_over.py
gioman: changed.

python/plugins/processing/algs/gdal/rasterize_over.py
gioman: changed.

gioman:
> May I ask to add (if applicable) "Additional command-line arguments" parameter like in other GDAL algorithms to avoid cluttering UI with new parameters when one will require something specific and not exposed in the current implementation

@alexbruy added "Additional command-line arguments" . The "creation options" must not be added to this implementation of gdal rasterize because no new output is generated.

python/plugins/processing/algs/gdal/rasterize_over.py
gioman: Removed.

gioman:
> +1 for Nyall's request for tests.

@nyalldawson @alexbruy sorry for the dumb question, but as this implementation of gdal rasterize does not create any new output as it modifies the input raster layer, is this a problem when adding a test?

python/plugins/processing/algs/gdal/rasterize_over.py
gioman: I removed a few imports, but the one for boolean I think is necessary (for the "add" parameter), am I wrong?

Gustry:
> @nyalldawson @alexbruy sorry for the dumb question, but as this implementation of gdal rasterize does not create any new output as it modifies the input raster layer, is this a problem when adding a test?

GDAL/OGR tests are testing the command line generation.
https://github.com/qgis/QGIS/blob/master/python/plugins/processing/tests/GdalAlgorithmsVectorTest.py

python/plugins/processing/algs/gdal/rasterize_over.py
```diff
@@ -23,36 +23,68 @@
...
+        self.addParameter(QgsProcessingParameterRasterLayer(self.INPUT_RASTER, self.tr('Input raster layer')))
+
+        self.addParameter(QgsProcessingParameterField(self.FIELD,
+                                                      self.tr('Field to use for burn in value'),
+                                                      None,
+                                                      self.INPUT,
+                                                      QgsProcessingParameterField.Numeric,
+                                                      optional=True))
+
+        self.addParameter(QgsProcessingParameterNumber(self.BURN,
```
nyalldawson: Wouldn't it be better to split this to two algorithms, to avoid the same issue that #32097 is addressing?

DelazJ:
@gioman can you check whether these are documented and add the convenient [needs-docs] if applicable? Thanks

python/plugins/processing/algs/gdal/rasterize_over.py
gioman: @nyalldawson in my opinion the situation is similar but not exactly the same. The UI for that SAGA tool was too complicated (I think that this is not arguable) and moreover to make the tool work was necessary to enter a fake parameter (an empty reclassification table). Both things contributed to make the tool not very user friendly. If I'm not wrong this happens because for Processing tools based on description files (and using just description files) we cannot define any type of logic (i.e. If this parameter is set then ignore the other). On the other hand for GDAL based tools we can easily manage logic with parameters, in the case for gdal rasterize (and I guess for several other similar cases) the values to be transferred to the pixels can come from a column of the attribute table, or a fixed, user-defined value. The latter is used by default unless the former has been defined. This way of functioning is not ideal but is perfectly acceptable for a user, and so we are not obligated to split the tool. Anyway this is just my opinion, if we choose to split... I will split.

nyalldawson: Ok, let's get a second opinion from @alexbruy?

alexbruy: Personally I preffer to have simpler algorithms with parameters specific for single task. Of course, we can add conditional logic to handle cases like "if paramA is present use it, otherwise use paramB" but IMHO this is error prone. Also algorithm UI becomes much complex, for example look at the import into PostGIS and some other algorithms, sometimes their dialogs even does not fit into screen and have a lot of parameters with not-so-obvious behavior.

I also think that is algorithm use different parameters for the same value, better to split it into two or use dynamic parameters if possible.

Can we use a data-defined parameter here?

nyalldawson: > Can we use a data-defined parameter here?

Nope - that's a (killer) feature solely available to the native algorithms.

python/plugins/processing/algs/gdal/rasterize_over.py
```diff
@@ -90,7 +83,7 @@ def name(self):
         return 'rasterize_over'
 
     def displayName(self):
-        return self.tr('Rasterize (write over existing raster)')
+        return self.tr('Rasterize (write attribute value over existing raster)')
```
nyalldawson: A lengthy string like this should be moved to the shortDescription() method, and the name should be kept as brief as possible. Maybe "Rasterize (overwrite with attribute)"?

PR #32082, opened 2019-10-01
Makes GPKG prefered input format for OGR algs #Fixes 29097

### LINKED ISSUE #29097: Processing/OGR: do not use shapefile as format for temp input files
Author Name: **Tobias Wendorff** (Tobias Wendorff)
Original Redmine Issue: [21279](https://issues.qgis.org/issues/21279)

Redmine category:processing/core

---
The algorithms from GDAL's toolbox for vector processing still use shapefiles as intermediate format. This results in corrupted fieldnames, since those are limited in the DBF format. Also, field contents might get corrupted if they're longer than 255 chars.

@ogr2ogr -f "ESRI Shapefile" C:/Users/tobwen/AppData/Local/Temp/processing_5fe343f547494d5e9fc31fe5cc38c868/680b663b94af46ab9516f2acd4141efd/OUTPUT.shp R:/demo.geojson@

Please use GPKG for this.

SrNetoChan:
## Description

Currently, when a GDAL algorithm uses a memory layer (or runs from the temporary output in a processing model) the input layer is converted to a shapefile, with all its drawbacks. The bigger one being that the names of the columns get truncated.

This makes the default transaction format Geopackage.

Fixes #29097

## Checklist

### DIFF
--- a/python/plugins/processing/algs/gdal/GdalAlgorithm.py
+++ b/python/plugins/processing/algs/gdal/GdalAlgorithm.py
@@ -89,6 +89,7 @@ class GdalAlgorithm(QgsProcessingAlgorithm):
                 # and extract selection if required
                 ogr_data_path = self.parameterAsCompatibleSourceLayerPath(parameters, parameter_name, context,
                                                                           QgsVectorFileWriter.supportedFormatExtensions(),
+                                                                          'gpkg',
                                                                           feedback=feedback)
                 ogr_layer_name = GdalUtils.ogrLayerName(ogr_data_path)
             else:
--- a/python/plugins/processing/algs/gdal/GdalAlgorithm.py
+++ b/python/plugins/processing/algs/gdal/GdalAlgorithm.py
@@ -87,9 +87,13 @@ class GdalAlgorithm(QgsProcessingAlgorithm):
             if executing:
                 # parameter is not a vector layer - try to convert to a source compatible with OGR
                 # and extract selection if required
+                if 'gpkg' in QgsVectorFileWriter.supportedFormatExtensions():
+                    preferred_format = 'gpkg'
+                else:
+                    preferred_format = 'shp'
                 ogr_data_path = self.parameterAsCompatibleSourceLayerPath(parameters, parameter_name, context,
                                                                           QgsVectorFileWriter.supportedFormatExtensions(),
-                                                                          'gpkg',
+                                                                          preferred_format,
                                                                           feedback=feedback)
                 ogr_layer_name = GdalUtils.ogrLayerName(ogr_data_path)
             else:
--- a/python/plugins/processing/algs/gdal/GdalAlgorithm.py
+++ b/python/plugins/processing/algs/gdal/GdalAlgorithm.py
@@ -87,13 +87,9 @@ class GdalAlgorithm(QgsProcessingAlgorithm):
             if executing:
                 # parameter is not a vector layer - try to convert to a source compatible with OGR
                 # and extract selection if required
-                if 'gpkg' in QgsVectorFileWriter.supportedFormatExtensions():
-                    preferred_format = 'gpkg'
-                else:
-                    preferred_format = 'shp'
                 ogr_data_path = self.parameterAsCompatibleSourceLayerPath(parameters, parameter_name, context,
                                                                           QgsVectorFileWriter.supportedFormatExtensions(),
-                                                                          preferred_format,
+                                                                          QgsVectorFileWriter.supportedFormatExtensions()[0],
                                                                           feedback=feedback)
                 ogr_layer_name = GdalUtils.ogrLayerName(ogr_data_path)
             else:
--- a/python/plugins/processing/tests/GdalAlgorithmsGeneralTest.py
+++ b/python/plugins/processing/tests/GdalAlgorithmsGeneralTest.py
@@ -109,11 +109,11 @@ class TestGdalAlgorithms(unittest.TestCase):
         self.assertIsNotNone(alg)
         parameters = {'INPUT': 'testmem'}
         feedback = QgsProcessingFeedback()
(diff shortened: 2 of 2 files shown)


python/plugins/processing/algs/gdal/GdalAlgorithm.py
```diff
@@ -89,6 +89,7 @@ def getOgrCompatibleSource(self, parameter_name, parameters, context, feedback,
                 # and extract selection if required
                 ogr_data_path = self.parameterAsCompatibleSourceLayerPath(parameters, parameter_name, context,
                                                                           QgsVectorFileWriter.supportedFormatExtensions(),
+                                                                          'gpkg',
```
lbartoletti: Please, don't hardcode this format (even if it's better than shp :) )

`QgsVectorFileWriter.supportedFormatExtensions()[0]` returns 'gpkg' is GDAL have gpkg support (it can be disabled) instead 'shp'.

SrNetoChan: I wouldn't call it hardcoded... but who am I? :-p

If gdal has support to gpkg, why isn't the first format from that list used? It must be forced somewhere else. No?

SrNetoChan:
@lbartoletti seems that the SHP is "hardcoded" directly on the parameterAsCompatibleSourceLayerPath() method, because if no preferedFormat argument is given, "shp" is used:

https://qgis.org/api/classQgsProcessingParameters.html#a9a9de0b49b654398111a2f9190154972

So, it seems to me that this is the only way to do it, unless we want to change the default format for the the function, which I suppose will affect many more algorithms.

What I propose is that at least for GDAL algorithms, which supports gpkg (and it's no probable that someone switch it off), we start to use geopackage as preferredFormat instead.

If we agree on this, I will work on fixing the tests (that are probably expecting an shapefile as output)

nyalldawson:
I believe what @lbartoletti is asking is that you first test to see if gpkg is an available format for writing vectors, and if not, fallback to shapefile.

SrNetoChan:
@nyalldawson @lbartoletti

Ahh! So, confirm that the gpkg driver for GDAL driver before setting it as default? If it is, use gpkg, if not (how knows why the user have turned it off) use shapefile instead (which AFAIK is always on)?

Can someone point me to the class or function where I could check that? thanks!

lbartoletti:
> I believe what @lbartoletti is asking is that you first test to see if gpkg is an available format for writing vectors, and if not, fallback to shapefile.

Totally :)

>  supports gpkg (and it's no probable that someone switch it off), 

Unfortunately, I know some people who disable it (and others because they require dependencies). ¯\_(ツ)_/¯

SrNetoChan:
Nevermind, as soon as you turn the gpkg driver off, the QgsVectorFileWriter.supportedFormatExtensions() no longer returns the gpkg, but as you were expecting, the preferred format is used and caBum!

Is there a better way to check if the gpkg is active that simpling calling QgsVectorFileWriter.supportedFormatExtensions() and check it it's there?

SrNetoChan:
This seems to work, but I am not sure it's a very elegant way of going it.

SrNetoChan:
Maybe an even better way would be to "fix" the classQgsProcessingParameters (or some of the methods in which it relies upon) to make sure that the preferredFormat is present in the compatibleFormats list, before trying to use it.

nyalldawson:
This approach looks fine to me. I think the logic **does** belong here, since only the provider would know what fallbacks to use in case of a missing preferred format. 

@lbartoletti are you happy with this one?

python/plugins/processing/algs/gdal/GdalAlgorithm.py
```diff
@@ -87,8 +87,13 @@ def getOgrCompatibleSource(self, parameter_name, parameters, context, feedback,
             if executing:
                 # parameter is not a vector layer - try to convert to a source compatible with OGR
                 # and extract selection if required
+                if 'gpkg' in QgsVectorFileWriter.supportedFormatExtensions():
+                    preferred_format = 'gpkg'
+                else:
+                    preferred_format = 'shp'
                 ogr_data_path = self.parameterAsCompatibleSourceLayerPath(parameters, parameter_name, context,
                                                                           QgsVectorFileWriter.supportedFormatExtensions(),
+                                                                          preferred_format,
```
lbartoletti: `QgsVectorFileWriter.supportedFormatExtensions()[0]` will works more simply.

What I got on my system when GDAL is built:

Without external option

```
QgsVectorFileWriter.supportedFormatExtensions()
['shp', '000', 'bna', 'csv', 'dgn', 'dxf', 'geojson', 'geojsonl', 'geojsons', 'gml', 'gpx', 'gxt', 'json', 'kml', 'tab', 'txt', 'xml']
```

With gpkg/sqlite3 (and other formats):

```
QgsVectorFileWriter.supportedFormatExtensions()
['gpkg', 'shp', '000', 'bna', 'csv', 'dgn', 'dxf', 'geojson', 'geojsonl', 'geojsons', 'gml', 'gpx', 'gxt', 'ili', 'itf', 'json', 'kml', 'ods', 'sqlite', 'tab', 'txt', 'xlsx', 'xml', 'xtf']
```

By default, supportedFormatExtensions() returns the list sorted by recommanded formats, so it's not necessary to add this if/else statement.

lbartoletti:
Thanks @SrNetoChan

alexbruy:
As this changes default output format, GDAL tests should be adjusted too.

SrNetoChan:
@alexbruy I have worked on that already. Would it make sense to add a new test to test this fallback mechanism?

If so, how could I disabled and enable gdal gpkg driver using PyQGIS?

SrNetoChan:
@lbartoletti I have changed the code to use your latest suggestion. It looks much better.

@alexbruy I think the changes I made to the tests should be enough.

lbartoletti (approved):

PR #32088, opened 2019-10-02
change confusing parameter name in DB Manager

gioman:
Explanation:

https://lists.osgeo.org/pipermail/qgis-developer/2019-October/058742.html

The new option name has been chosen to be consistent with a similar option used in OGR ("promote_to_multi").


rldhont (approved):

Gustry:
Thanks, I'm also very confused by these options in the DB Manager import dialog.
Maybe a tooltip to explain the behavior?

m-kuhn:
Good move. Backporting this will break existing translations (i.e. everyone will see it in english after, even if the old version was translated), is this desired?

nyalldawson:
No backports here please!

gioman:
> No backports here please!

label removed

PR #32090, opened 2019-10-02
monkey patch sys.argv if AttributeError  in tests

Gustry:
## Description

In QGIS Master, there is already this fix:
https://github.com/qgis/QGIS/blob/master/python/testing/__init__.py#L408
which is not present in QGIS 3.4.

Can we add it in QGIS 3.4?

When I launch tests using the QGIS docker image, tests are passing on master, but not 3.4.

@elpaso Related commit in master: https://github.com/qgis/QGIS/commit/ed9709b4faa7b565ee75b7298e4b02069bcaf994

## Checklist

### DIFF
--- a/python/testing/__init__.py
+++ b/python/testing/__init__.py
@@ -394,6 +394,11 @@ def start_app(cleanup=True):
     except NameError:
         myGuiFlag = True  # All test will run qgis in gui mode
 
+        try:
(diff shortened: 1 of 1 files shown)


elpaso (approved):

Gustry:
Thanks @elpaso

Issue #32002, opened 2019-09-25
Decrease the number of round-trips to postgres servers when connecting using the Data source manager


palmerj:
**Feature description.**
We are testing if it's feasible for desktop clients to connect to large PostgreSQL database over the WAN. Our current database has over 2000 tables, of which 1100 have a PostGIS column. The average WAN response time from the QGIS desktop client to the PostgreSQL server is 35ms. Bandwidth is not a major issue in this case.

When connecting via PostgreSQL connections in the Data source manager (using the "Also list tables with no geometry", "User estimated table metadata" options) it takes about 1.5mins to fully populate the dialog. We have tested multiple configurations but have not found any of the options any faster for our use cases. This 1.5min process runs over 5000 SQL queries, and with the 35ms latency that's why the process takes so long. The top queries by volume are:

| calls | rows  | query                                                                                                                                                                                               |
|-------|-------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| 3089  | 496  | select description from pg_catalog.pg_description where objoid = $1 and objsubid = ?                                                                                                                |
| 2009  | 21597 | SELECT attname, CASE WHEN typname = ANY(ARRAY[?,?,?]) THEN ? ELSE ? END AS isSpatial FROM pg_attribute JOIN pg_type ON atttypid=pg_type.oid WHERE attrelid=regclass(?) AND attnum>? ORDER BY attnum |
| 20    | 20    | SELECT proj4text FROM spatial_ref_sys WHERE srid = ? LIMIT ?                                                                                                                                        |

In this situation we also have we also have 380 queries to try and determine the geometry type for columns that don't have a specific geometry typemod set. e.g `SELECT DISTINCT upper(geometrytype) xxx`

It looks like most of queries are coming from determining the column type information https://github.com/qgis/QGIS/blob/3be3984e8edd980a8cf36a1c59bf1f0595e0860b/src/providers/postgres/qgspostgresprovider.cpp#L744 and https://github.com/qgis/QGIS/blob/b837e65662d5328863dc1048e180c8b28752c23b/src/providers/postgres/qgspostgresconn.cpp#L402

I think the process could be quickly improved using lazy loading of the column info. The info is not needed immediately in the dialog (only schema, tablename, comment, spatial column, geometry type, srid, relation type, primary key seem to be needed). Surely the other information could be populated if required (e.g double click to set SQL, or additional potential key information if it's a view).

[...]

PR #32091, opened 2019-10-02
WIP: improve UX with high-latency PostgreSQL connections

strk:
See #32002

### DIFF
--- a/src/providers/postgres/qgscolumntypethread.cpp
+++ b/src/providers/postgres/qgscolumntypethread.cpp
@@ -67,43 +67,42 @@ void QgsGeomColumnTypeThread::run()
     return;
   }
 
-  int i = 0, n = layerProperties.size();
-  for ( QVector<QgsPostgresLayerProperty>::iterator it = layerProperties.begin(),
-        end = layerProperties.end();
-        it != end; ++it )
+  int totalLayers = layerProperties.length();
+  int addedLayers = 0;
+
+  emit progress( addedLayers, totalLayers );
+
+  QVector<QgsPostgresLayerProperty *> unrestrictedLayers;
+
+  for ( auto &layerProperty : layerProperties )
   {
-    QgsPostgresLayerProperty &layerProperty = *it;
-    if ( !mStopped )
+    if ( !layerProperty.geometryColName.isNull() &&
+         ( layerProperty.types.value( 0, QgsWkbTypes::Unknown ) == QgsWkbTypes::Unknown ||
+           layerProperty.srids.value( 0, std::numeric_limits<int>::min() ) == std::numeric_limits<int>::min() ) )
     {
-      emit progress( i++, n );
-      emit progressMessage( tr( "Scanning column %1.%2.%3…" )
-                            .arg( layerProperty.schemaName,
-                                  layerProperty.tableName,
-                                  layerProperty.geometryColName ) );
-
-      if ( !layerProperty.geometryColName.isNull() &&
-           ( layerProperty.types.value( 0, QgsWkbTypes::Unknown ) == QgsWkbTypes::Unknown ||
-             layerProperty.srids.value( 0, std::numeric_limits<int>::min() ) == std::numeric_limits<int>::min() ) )
-      {
-        if ( dontResolveType )
-        {
-          QgsDebugMsg( QStringLiteral( "skipping column %1.%2 without type constraint" ).arg( layerProperty.schemaName, layerProperty.tableName ) );
-          continue;
-        }
-
-        mConn->retrieveLayerTypes( layerProperty, mUseEstimatedMetadata );
-      }
+      unrestrictedLayers << &layerProperty;
     }
+  }
 
-    if ( mStopped )
-    {
-      layerProperty.types.clear();
-      layerProperty.srids.clear();
-      break;
-    }
+  if ( mStopped )
+  {
+    emit progress( 0, 0 );
+    emit progressMessage( tr( "Table retrieval stopped." ) );
+    QgsPostgresConnPool::instance()->releaseConnection( mConn );
+    mConn = nullptr;
+    return;
+  }
 
-    // Now tell the layer list dialog box...
+  if ( ! dontResolveType )
+  {
+    mConn->retrieveLayerTypes( unrestrictedLayers, mUseEstimatedMetadata );
+  }
+
+  for ( const auto &layerProperty : layerProperties )
+  {
+    // Tell the layer list dialog box about the completed layers
     emit setLayerType( layerProperty );
+    emit progress( ++addedLayers, totalLayers );
   }
 
   emit progress( 0, 0 );
--- a/src/providers/postgres/qgspostgresconn.cpp
+++ b/src/providers/postgres/qgspostgresconn.cpp
@@ -28,6 +28,7 @@
 #include "qgswkbtypes.h"
 #include "qgssettings.h"
 #include "qgsjsonutils.h"
+#include "qgspostgresstringutils.h"
 
 #include <QApplication>
 #include <QThread>
@@ -399,15 +400,16 @@ void QgsPostgresConn::unref()
 }
 
 /* private */
+// TODO: deprecate this function
 void QgsPostgresConn::addColumnInfo( QgsPostgresLayerProperty &layerProperty, const QString &schemaName, const QString &viewName, bool fetchPkCandidates )
 {
   // TODO: optimize this query when pk candidates aren't needed
   //       could use array_agg() and count()
   //       array output would look like this: "{One,tWo}"
-  QString sql = QStringLiteral( "SELECT attname, CASE WHEN typname = ANY(ARRAY['geometry','geography','topogeometry','raster']) THEN 1 ELSE null END AS isSpatial FROM pg_attribute JOIN pg_type ON atttypid=pg_type.oid WHERE attrelid=regclass('%1.%2') AND attnum>0 ORDER BY attnum" )
+  QString sql = QStringLiteral( "SELECT attname, CASE WHEN typname in ('geometry','geography','topogeometry','raster') THEN 1 ELSE null END AS isSpatial FROM pg_attribute JOIN pg_type ON atttypid=pg_type.oid WHERE attrelid=regclass('%1.%2') AND NOT attisdropped AND attnum>0 ORDER BY attnum" )
                 .arg( quotedIdentifier( schemaName ),
                       quotedIdentifier( viewName ) );
-  //QgsDebugMsg( sql );
+  QgsDebugMsg( "getting column info: " + sql );
   QgsPostgresResult colRes( PQexec( sql ) );
 
   layerProperty.pkCols.clear();
@@ -442,7 +444,7 @@ bool QgsPostgresConn::getTableInfo( bool searchGeometryColumnsOnly, bool searchP
   int nColumns = 0;
   int foundInTables = 0;
   QgsPostgresResult result;
-  QgsPostgresLayerProperty layerProperty;
+  QString query;
 
   //QgsDebugMsg( QStringLiteral( "Entering." ) );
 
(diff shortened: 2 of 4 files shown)


strk:
It looks like there are not tests for the PostgreSQL Datasource selector dialog (because I know Travis has been green for changes I pushed to that code which did break things...)

strk:
Further analisys of what happens, currently, with options "also show tables with no geometries" and "do NOT only look in metadata tables".

For EACH SCHEMA (in this PR) or ONCE FOR WHOLE DATABASE (master) we run 6 queries to fetch a list of layers:

  1. Look in geometry_columns
  2. Look in geography_columns
  3. Look in topology.layer
  4. Look in raster_columns
  5. Look everywhere for tables having at least one spatial column
  6. Look everywhere for any table

It seems to me that given we have to do 6, all the previous 5 queries are just useless ?

strk:
In its current status, this PR brings down the number of database queries to 439.
Target database has 1956 table (`relkind IN ('v','r','m','p')`) in 73 schemas.

439 is about 6 * 73  (=378) so I guess we could further divide this number by 6 by merging the multiple (spatial/non-spatial/metadata) queries into one.

@palmerj it would be great if you could test this before merging it to master

gioman:
awesome

strk:
After correcting the number, I shall add that we still have a one-query-per-table for unconstrained geometry tables, unless `Don't resolve type of unrestricted columns (GEOMETRY)` is checked, in which case load is close to immediate for me, once reverting the very first commit in this PR (the one running a query for each schema, so to bring down the number of queries to exactly 6 !

strk:
More stats: when resolving geometry types (unchecking `Don't resolve type of unrestricted columns (GEOMETRY)`)  results in 394 additional queries. That's the number of unconstrained spatial columns in the database (spread across 358 tables as some tables have multiple spatial columns).

strk:
One possible idea to reduce round trips for the resolution of unrestricted columns could be to run a single query to resolve all columns at once. The query could be a `UNION` returning schema.table.column.detected_type.detected_srid ...

strk:
> awesome

Giovanni, if you could test this it'd be great :)

I suspect AWS/RDS engages some form of pg proxy that
only _pretends_ to keep a connection alive but rather
re-creates it for every statement, making it impossible
for QGIS to assume a trascurable time for each query..

strk:
It looks to me that checking the `Don't resolve type of unrestricted column` prevents tables from being added to the list of layers :/

palmerj:
> It seems to me that given we have to do 6, all the previous 5 queries are just useless ?
It's a valid point. However I think there is an option checkbox to only show layers in the layer registries, which would mean queries 1-4

strk:
When both "only look in registries" and "also show tables without a geometry" are set, we'll end up doing 6 anyway (scan all tables).

palmerj:
> @palmerj it would be great if you could test this before merging it to master

I can setup my build environment tomorrow. But it might take some time!

palmerj:
> When both "only look in registries" and "also show tables without a geometry" are set, we'll end up doing 6 anyway (scan all tables).

Ok then we should optimise this.

palmerj:
> One possible idea to reduce round trips for the resolution of unrestricted columns could be to run a single query to resolve all columns at once. The query could be a UNION returning schema.table.column.detected_type.detected_srid ...

Thanks - sounds good. You could also use something like query_to_xml to run a dynamic SQL subselect for each table to return the SRID and geometry type from standard table set query

strk:
Another option is pipelinening sql queries: https://2ndquadrant.github.io/postgres/libpq-batch-mode.html

palmerj:
Does the QGIS project have a minimum version of libpq and PostgresSQL server for runtime? This feature needs PostgreSQL >= 9.6 and server >= 8.4. Might be easier to avoid pipelinening.

strk:
Ok I'm taking the UNION approach. What I'm not understanding is why a QgsPostgresLayerProperty has a QList of srids and types, are them list of possibly different types/srids found in data ? @jef-n  ?

palmerj:
Won't UNION create a horrendously long SQL string for 100s of tables?

elpaso:
> Ok I'm taking the UNION approach. What I'm not understanding is why a QgsPostgresLayerProperty has a QList of srids and types, are them list of possibly different types/srids found in data ? @jef-n ?

Yes exactly. I think there are some test cases here on the higher level connection API: https://github.com/qgis/QGIS/blob/master/tests/src/python/test_qgsproviderconnection_postgres.py#L87

strk:
Great to have tests ! What I read in code seems to be wrong (like there's an attempt to match a type with a SRID but I dubt we're really wanting that in the UI). I'll see if the existing test is good enough or improve it.

Horrendously long query ? YES, so what ? How's an horrendously big query worse than N queries of "horrendously big query"/N size ? Anyway I'm writing the code to be callable with a list of layers, so the caller might eventually use "chunks" of layers to tweak the size of the single query.

palmerj:
I mean string length. Not the nicest when statement logging is on. But I guess it's only a very minor issue.

strk:
75171 characters of a query, if we are to query all of 394 unconstrained spatial columns of your test database...

strk:
Alright, the UNION based change is now committed in this PR.
Number of queries goes down from 396 + 6 to  7 (the 6 queries for layer registries plus the new one for unconstrained layer columns type/srid determination).

It's very fast to load layers info now. So fast that what became more annoying to me is the startup time, when we fetch database info (postgis version, raster/topology/pgpatch support, endiannes...).
I'm could improve those ones as well, since I'm at it, but before going there I'll make sure to get a green Travis and verify the type/srid detection code against the existing test.

strk:
@elpaso I don't understand the failure in your test: https://cdash.orfeo-toolbox.org/testDetails.php?test=11668908&build=17300

Is the intention to trigger an error creating a "myNewSchema" schema ? I just don't see why my changes should trigger that ...

strk:
Ok an more basic question: why do we want to assert this:

Fatal: ASSERT: "types.size() == srids.size()" in file /usr/src/qgis/qgis-master/src/providers/postgres/qgspostgresconn.h, line 89

In other words, @elpaso, if `types` and `srids` are expected to be possibly SRIDs and possibly TYPES found in a table, why do we want them to be the same number of possibilities ? Can't we have only points, in a table, but all of different SRIDs ? Or can't we have differently typed geometries but of the same SRID ?

elpaso:
Well, the output is mangled (mixed stream from stdout and stderr probably) but, as far as I can see, 
you don't get geometryless table anymore:  `self.assertTrue('geometryless_table' in table_names)`

The other failure is related:
```
table_property = self._table_by_name(table_properties, 'myNewAspatialTable')            self.assertIsNotNone(table_property)
```

elpaso:
> Ok an more basic question: why do we want to assert this:
> 
> Fatal: ASSERT: "types.size() == srids.size()" in file /usr/src/qgis/qgis-master/src/providers/postgres/qgspostgresconn.h, line 89
> 
> In other words, @elpaso, if `types` and `srids` are expected to be possibly SRIDs and possibly TYPES found in a table, why do we want them to be the same number of possibilities ? Can't we have only points, in a table, but all of different SRIDs ? Or can't we have differently typed geometries but of the same SRID ?

I didn't write that code ... but, AFAIK the pg provider can only handle a single tuple of (CRS, Geometry Type) for each layer.

strk:
>  I didn't write that code ... but, AFAIK the pg provider can only handle a single tuple of (CRS, Geometry Type) for each layer.

So I should, in the scenario I represented abve, create a unique tuple for each combination ?

@jef-n we really need your help here...

> you don't get geometryless table anymore

Ok thanks, I'll look into this

strk:
Alright I made the type/srid detection code always return all combinations of found values.
Also I reverted a commit that was trying to be smart but evidently broke the spatial-less tables detection. Let's see if Travis is happy now.

src/providers/postgres/qgscolumntypethread.cpp
```diff
@@ -67,43 +67,48 @@ void QgsGeomColumnTypeThread::run()
...
+    emit progress( 0, 0 );
+    emit progressMessage( mStopped ? tr( "Table retrieval stopped." ) : tr( "Table retrieval finished." ) );
+    QgsPostgresConnPool::instance()->releaseConnection( mConn );
+    mConn = nullptr;
+  }
 
-    // Now tell the layer list dialog box...
+  mConn->retrieveLayerTypes( unrestrictedLayers, mUseEstimatedMetadata );
+
+  for ( auto layerProperty : unrestrictedLayers )
```
elpaso:
```suggestion
  for ( const auto &layerProperty : qgis::as_const(unrestrictedLayers) )
```

untested, but you get the idea, please also do the same for other iterators in this PR.

strk: Done, as part of an interactive rebase to master branch, to reorganize commits in logical steps.
I've also found the reason why tables where not shown, hopefully next travis build will be green (your test was great!)

strk:
Updated timing info for my case (phone-connected to RDS against that 1956 tables in 73 schemas, with 396 unconstrained geometry columns):

 - 15 seconds for the initial connection (fetching versions and capabilities)
 - 27 seconds to fetch all available layers (2007 entries)

I _think_ this can be further optimized:

  - the 15 seconds for initial connection can be reduced by running a single catch-all query
  - the 27 seconds to fetch all available layers can be reduced by running a single query instead of 6

strk:
I've pushed another commit further reducing the layer registry queries from 4 to 1.

strk:
I suspect the problem with the test is that it cheats QGIS into thinking a table exists while in fact it doesn't. Such error would be tolerated before my code change in that it would be a single SQL query to fail, while with my change that single SQL query is for all tables (so all-or-nothing)

jef-n:
> I didn't write that code ... but, AFAIK the pg provider can only handle a single tuple of (CRS, Geometry Type) for each layer.

upper layers can only handle single tuples - so the provider filters by (existing) CRS and geometry type combinations and returns the selected one.

strk:
> upper layers can only handle single tuples - so the provider filters by (existing) CRS and geometry type combinations and returns the selected one.

What do you mean by "upper layers" ?

is QgsPostgresLayerProperty an interface used by anything
outside of the PostgreSQL provider ?
Or the provider itself could build combinations w/out
requiring those "srids" and "types" members to be of
the same length...

jef-n:
> What do you mean by "upper layers" ? 

QgsVectorDataProvider - each has one CRC and one wkbType.  And each tuple is represented as one row in the section UI.

palmerj:
> I can setup my build environment tomorrow. But it might take some time!

@strk I've tested on Ubuntu 18.04 and the speed improvements are great.  However as already noticed by @elpaso, tables with a geometry column, but without a defined type are not showing up

strk:
I actually do see geometry columns with non-defined geometry type and srid, are you maybe mixing the problem of the LIMIT for detection @palmerj ?

strk:
My local test (connecting to `qgis_test` database) does show  the geometryless_table, when the corresponding option is enabled in GUI. So I'm not sure why Travis doesn't find it.

I see the test is using `QgsPostgresProviderConnection::tables`, passing it the `qgis_test` schema as a parameter, but the code of that function is forcing a schema of `public` for some reason I don't understand (is master branch overlooking that parameter?). Here's the line of code: https://github.com/qgis/QGIS/blob/master/src/providers/postgres/qgspostgresproviderconnection.cpp#L325

strk:
Please disreguard previous comment, I misread the code. Enabling debug lines I even see `geometryless_table` being returned by that function, so I'm puzzled

strk:
Figured! Wrong classification for the table:
428: XXX layer qgis_test.geometryless_table. is spatial (has 1 spatial cols)

Will fix that in the next 24 hours hopefully

strk:
Alright the support for "aspatial" tables (as retrived by testcase, as PostgreSQL layer selector worked already) is now fixed. Next failure is about the SRID/Types array:

        geometries_table = self._table_by_name(conn.tables('qgis_test'), 'geometries_table')
        srids = [t.crs.postgisSrid() for t in geometries_table.geometryColumnTypes()]
        self.assertEqual(srids, [0, 0, 0, 3857, 4326, 0])
        types = [t.wkbType for t in geometries_table.geometryColumnTypes()]
        self.assertEqual(types, [7, 2, 1, 1, 1, 3])

The test expects srids/types arrays to not be all combinations but list of actual pairs found in database. Sounds odd to me, but I will conform for now.

strk:
@elpaso: I've changed your test to be immune from ordering of the type/srid arrays (maintaining their correspondence) - still fine ?

elpaso:
@strk LGTM, thanks!

PR #32095, opened 2019-10-02
allow to use remote content for init code file path in attribute form

### LINKED ISSUE #32040: Python macros are executed without authorization in forms
Similarly to the project macros, form macros should be executed only when allowed.

3nids:
superseds #32015
fixes #32040 

**Content of this PR:**
* allow to use URL for Python init code in attribute forms (similarly to what is possible with custom forms)
* creates an enum (`Qgis::PythonMacroMode`) instead of using integers
* creates a method `static bool QgsGui::pythonMacroAllowed( void ( *lambda )() = nullptr, QgsMessageBar *messageBar = nullptr )`
  * determines if Python is allowed to be run depending on current setting
  * if no lambda is given, it is modal: if the mode is `Qgis::PythonMacroMode::Ask`, a modal message box will be shown (a new mode 'NotForThisSession` has been added to not ask anymore).
  * if a lambda is given, there is no message box, but a message bar item is pushed to the given message bar

**Follow-up:**
* I think it would be nice to have a new mode `TrustedProject` which would contain a list of trusted project and ask otherwise. 

**Further notes:**
* Due to an issue in former version of SIP, I could not use a scoped enum for `Qgis::PythonMacroMode`:
  * if scoped and traditional enums are mixed, there is an issue: https://www.riverbankcomputing.com/hg/sip/rev/ccc4eda868de
  * if I make all enums scoped, there is an issue with a signal with a signature containing the scoped enum: https://www.riverbankcomputing.com/pipermail/pyqt/2019-October/042190.html
* => I hate SIP

This is ready for review.

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -6297,17 +6297,8 @@ bool QgisApp::addProject( const QString &projectFile )
     {
       if ( !QgsProject::instance()->readEntry( QStringLiteral( "Macros" ), QStringLiteral( "/pythonCode" ), QString() ).isEmpty() )
       {
-        Qgis::PythonMacroMode enableMacros = settings.enumValue( QStringLiteral( "qgis/enableMacros" ), Qgis::PythonMacroMode::Ask );
-
-        if ( enableMacros == Qgis::PythonMacroMode::Always || enableMacros == Qgis::PythonMacroMode::SessionOnly )
-        {
-          enableProjectMacros();
-        }
-        else if ( enableMacros == Qgis::PythonMacroMode::Ask )
-        {
-          auto lambda = []() {QgisApp::instance()->enableProjectMacros();};
-          QgsGui::pythonMacroAllowed( lambda, mInfoBar );
-        }
+        auto lambda = []() {QgisApp::instance()->enableProjectMacros();};
+        QgsGui::pythonMacroAllowed( lambda, mInfoBar );
       }
     }
 #endif
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -1333,7 +1333,7 @@ QgisApp::QgisApp( QSplashScreen *splash, bool restorePlugins, bool skipVersionCh
     toggleMapTips( true );
   }
 
-  mTrustedMacros = false;
+  mPythonMacrosEnabled = false;
 
   // setup drag drop
   setAcceptDrops( true );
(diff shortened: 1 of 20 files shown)


nyalldawson:
Just want to confirm - this is for 3.12, right?

3nids:
> this is for 3.12, right?

I have no strong opinion for the macro security part.
But the URL for the macro is more a bugfix to me as it was a missing bit with respect to the UI.

Although I have struggled a bit with the enums again, I think this PR is very well contained. And determining if the security part is a bugfix or a feature....
But again no strong opinion.

src/app/qgisapp.cpp
```diff
@@ -6289,40 +6297,16 @@ bool QgisApp::addProject( const QString &projectFile )
     {
       if ( !QgsProject::instance()->readEntry( QStringLiteral( "Macros" ), QStringLiteral( "/pythonCode" ), QString() ).isEmpty() )
       {
-        int enableMacros = settings.value( QStringLiteral( "qgis/enableMacros" ), 1 ).toInt();
-        // 0 = never, 1 = ask, 2 = just for this session, 3 = always
+        Qgis::PythonMacroMode enableMacros = settings.enumValue( QStringLiteral( "qgis/enableMacros" ), Qgis::PythonMacroMode::Ask );
 
-        if ( enableMacros == 3 || enableMacros == 2 )
+        if ( enableMacros == Qgis::PythonMacroMode::Always || enableMacros == Qgis::PythonMacroMode::SessionOnly )
```
nyalldawson: switch would be nicer here

src/gui/qgsgui.cpp
```diff
@@ -229,6 +234,68 @@ QgsGui::QgsGui()
...
+      return true;
+    case Qgis::PythonMacroMode::Never:
+    case Qgis::PythonMacroMode::NotForThisSession:
+      return false;
+    case Qgis::PythonMacroMode::Ask:
+      if ( !lambda )
+      {
+        QMessageBox msgBox( QMessageBox::Information, "Python Macros",
+                            tr( "Python macros are currently disabled. Do you allow this macro to run?" ) );
+        QAbstractButton *stopSessionButton = msgBox.addButton( tr( "Don't ask anymore" ), QMessageBox::DestructiveRole );
```
nyalldawson:
```suggestion
        QAbstractButton *stopSessionButton = msgBox.addButton( tr( "Don't Ask Anymore" ), QMessageBox::DestructiveRole );
```

src/app/qgisapp.cpp
3nids: fixed, not needed anymore

PR #32097, opened 2019-10-02
split SAGA reclass raster values in order to improve UX, fixes #32006

### LINKED ISSUE #32006: SAGA Reclassify Values Tool broken - 'Index out of range' error in log  :bug
For QGIS 3.8

I cannot successfully run the tool using any of the 3 methods. The reclassification doesn't work and I cannot fill the nodata values either. Here is the log output:

Algorithm 'Reclassify values' starting…
Input parameters:
{ 'INPUT' : 'D:/Matt/Projects/Seine/site3ligra/Base/tourond.tif', 'MAX' : 237, 'METHOD' : 1, 'MIN' : 236, 'NEW' : 1, 'NODATA' : 0, 'NODATAOPT ' : True, 'OLD' : 0, 'OTHEROPT ' : False, 'OTHERS' : 0, 'RESULT' : 'C:/Users/info/Desktop/2.sdat', 'RETAB' : [None], 'RNEW' : 238, 'ROPERATOR' : 0, 'SOPERATOR' : 1, 'TOPERATOR' : 2 }

File "C:/OSGEO4~1/apps/qgis/./python/plugins\processing\algs\saga\SagaAlgorithm.py", line 282, in processAlgorithm
s = '{}\t{}\t{}\n'.format(values[i], values[i + 1], values[i + 2])
IndexError: list index out of range

Execution failed after 0.05 seconds

gioman:
SAGA reclass raster/grid values has 3 modes it can work: single, range and table.

The implementation we use has all the 3 modes mixed within the same UI, making this tool cumbersome to use.

This PR splits the tool in 3 to make it easier to use and along the way fixes https://github.com/qgis/QGIS/issues/32006


nyalldawson:
I think it's a good fix, but it's going to break existing models/scripts which use the all-in-one algorithm (since a matching algorithm will no longer exist). I'm happy to look into a way of deprecating saga algs so that we can leave the old one in place but hide it from the ui (we do that with native algs already), but can you please add some unit tests for these new algs in the meantime?

alexbruy:
I'm not a SAGA guy, but fix looks good to me. Only one thing, already mentioned by @nyalldawson — we can break existing workflows, better to leave old algorithm but hide it from toolbox/modeller.

gioman:
un-split tool re-added.

PR #32100, opened 2019-10-02
[Plugin installer] Fix handling zip file names with a dash as a version suffix separator

borysiasty:
Currently, the zip files can be named either package_name.zip or package_name.version_suffix.zip, i.e. the optional suffix must be delimited with a dot. However, there is a recurrent confusion when people use dash instead (e.g. my_plugin-1.2.3.zip). In that case the plugin installer unzips the package without any warning and then fails to localize it.

See e.g. https://lists.osgeo.org/pipermail/qgis-developer/2019-October/058747.html

As both characters can't be used in Python package name anyway, I'd simply accept them both as a version delimiter.

### DIFF
--- a/python/pyplugin_installer/installer_data.py
+++ b/python/pyplugin_installer/installer_data.py
@@ -397,7 +397,11 @@ class Repositories(QObject):
                     fileName = pluginNodes.item(i).firstChildElement("file_name").text().strip()
                     if not fileName:
                         fileName = QFileInfo(pluginNodes.item(i).firstChildElement("download_url").text().strip().split("?")[0]).fileName()
-                    name = fileName.partition(".")[0]
+                    match = re.match('(.*?)[.-]', fileName)
+                    if match:
+                        name = match.groups()[0]
+                    else:
+                        name = fileName
                     experimental = False
                     if pluginNodes.item(i).firstChildElement("experimental").text().strip().upper() in ["TRUE", "YES"]:
                         experimental = True


python/pyplugin_installer/installer_data.py
```diff
@@ -397,7 +397,11 @@ def xmlDownloaded(self):
                     fileName = pluginNodes.item(i).firstChildElement("file_name").text().strip()
                     if not fileName:
                         fileName = QFileInfo(pluginNodes.item(i).firstChildElement("download_url").text().strip().split("?")[0]).fileName()
-                    name = fileName.partition(".")[0]
+                    match = re.match('(.*?)[\.-]', fileName)
```
nyalldawson: It would be nice to use QFileInfo or python os.path methods here

borysiasty: @nyalldawson Probably there is a misunderstanding. I'm extracting a part of the base name here, not splitting the file name to the base name and extension. I used QFileInfo's method just a line above, where it was applicable.

nyalldawson: Ok. I think there's an issue with the regex here though - it need the "r" prefix or the \. won't work as intended

borysiasty: @nyalldawson @m-kuhn Could you please explain why and in what conditions it may not work? Honestly I can't see any connection... You mean the former dot or the latter? Of course the backslash is not necessary, I just used to put it for readability, in order to express it's a raw dot.

m-kuhn: The `\` is currently not escaped (swallowed away by python and not sent to the regexp).

As a test in python

```
print('\.')
print(r'\.')
```

borysiasty: Ok, I removed the backslash as it was more confusing than helping... Is it now satisfying for you? ;-) Btw. if you really insist for the raw string, now I can use it, as now there is no risk it would interfere with that backslash ;)

nyalldawson: Thanks. Python throws warnings because of these, and the warnings appear in the qgis log window. It's not a nice experience for users when a default install raises python warnings by itself...!

borysiasty: Ah now I see! I didn't notice those deprecation warnings - there are always tons of them (at least in Debian). Thanks!

m-kuhn: Thank you @borysiasty 
I realized, that my example actually didn't even proof the point I was trying to make :D

PR #32103, opened 2019-10-03
Don't try to load HTML content when a layout is created in a background thread

nyalldawson:
The underlying QWebPage class used to render these can ONLY be used from the main thread. So we raise an error and render nothing to avoid crashes.


m-kuhn:
Ouch.
Talking about. Has there anything been going on on the QWebEngine front in the last 2 years?

nyalldawson:
> Talking about. Has there anything been going on on the QWebEngine front in the last 2 years?

Nope, screw that. 

QtWebKit on the other hand, is being actively worked on...

see https://github.com/qtwebkit/qtwebkit

nyalldawson:
@nirvn no, that's the reverse.

nirvn:
Oh, then 🎉

PR #32104, opened 2019-10-03
Visual cleanup for qgsProject in pyqgis doc

nyalldawson:
Supersedes #31804


3nids:
will you try putting back `since. I'd say you can put it back at least whereever there is no `\since`. And if that's the issue, I'd still prefer to drop `\since` so we can stay coherent and have `since`everywhere.

nyalldawson:
@3nids done

3nids:
This looks better, cool!

PR #32108, opened 2019-10-03
Fix layer exporter force to single-part

### LINKED ISSUE #32089: DB manager import option "Create single-part geometries instead of multi-part" is broken
Despite its name, that has been always confusing (PR to change the name here https://github.com/qgis/QGIS/pull/32088 ), this option in DB Manager never transformed geometries to SINGLE from MULTI along the import process.

What the option means is that if I have a layer that is declared as MULTI but actually has no multipart geometries in it, then it can be imported as a non-MULTI table.

In past QGIS releases (i.e. 2.18 LTR)  if you had a layer with a multi-geometry and checked
that option then DB manager returned an error like

```
Error 7
Feature write errors:
Creation error for features from #0 to #0. Provider errors was:
Geometry type MultiLineString not compatible with provider type LineString.
PostGIS error while adding features: ERROR: Geometry type (MultiLineString)
does not match column type (LineString)
```

now it does **not** and imports to MULTI anyway and silently. And it does also if you don't have any multi-geometry, which of course are both bugs and regressions.

QGIS 3.4 and master.

elpaso:
Try to respect the option and fails if is there
any real MULTI geometry in the input layer.

Fixes #32089

### DIFF
--- a/src/core/qgsvectorlayerexporter.cpp
+++ b/src/core/qgsvectorlayerexporter.cpp
@@ -15,12 +15,14 @@
  *                                                                         *
  ***************************************************************************/
 
+
 #include "qgsfields.h"
 #include "qgsfeature.h"
 #include "qgsfeatureiterator.h"
 #include "qgsgeometry.h"
 #include "qgslogger.h"
 #include "qgsmessagelog.h"
+#include "qgsgeometrycollection.h"
 #include "qgscoordinatereferencesystem.h"
 #include "qgsvectorlayerexporter.h"
 #include "qgsproviderregistry.h"
@@ -28,6 +30,7 @@
 #include "qgsexception.h"
 #include "qgsvectordataprovider.h"
 #include "qgsvectorlayer.h"
+#include "qgsabstractgeometry.h"
 
 #include <QProgressDialog>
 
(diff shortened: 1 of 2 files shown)


src/core/qgsvectorlayerexporter.cpp
```diff
@@ -398,6 +411,27 @@ QgsVectorLayerExporter::exportLayer( QgsVectorLayer *layer,
...
       }
     }
+
+    // Handles conversion to single-part
+    if ( convertGeometryToSinglePart && fet.geometry().isMultipart() )
+    {
+      QgsGeometry singlePartGeometry { fet.geometry() };
+      // We want a failure if the geometry cannot be converted to single-part without data loss!
+      // check if there are more than one part
+      const QgsGeometryCollection *c = qgsgeometry_cast<QgsGeometryCollection *>( singlePartGeometry.constGet() );
```
nyalldawson:
```suggestion
      const QgsGeometryCollection *c = qgsgeometry_cast<const QgsGeometryCollection *>( singlePartGeometry.constGet() );
```

Hmm, this really shouldn't compile without the const...

nyalldawson (approved):

nyalldawson:
Nice fix, thanks!

elpaso:
@nyalldawson thank you for reviewing!

gioman:
@elpaso works like a charm (tested in db manager), thanks!

elpaso:
Thank you for testing!

gioman:
> Thank you for testing!

my duty.

I have anyway just spotted an error (on master) that I hope is not related to this fix:

In the browser I have two connection to GPKG datasources and I'm trying to import a layer from one GPKG into the other, regardless of the source layer I get messages like:


```
Failed to import some vector layers!

Feature write errors:
Failed to transform a feature with ID '1' to single part. Writing stopped.
```

gioman:
Just tested on 3.4.12 and works as expected :(

PR #32112, opened 2019-10-04
CMakeLists.txt: bump minimum cmake version to 3.7.0

rouault:
Commit 869ee9bc666fb939f3052047c54719d2a60575fa added the use of
VERSION_GREATER_EQUAL which is a 3.7 addition. So bump to it


rouault:
@nyalldawson I found in https://github.com/omnisci/omniscidb/pull/264/files an alternative trick using VERSION_GREATER which would just requires 3.3

nyalldawson:
I've done the alternative approach in #32136

PR #32114, opened 2019-10-04
Fix featureCount on postgres view when flag estimatedmetadata is set

troopa81:
We [can't estimate rows on postgres view](https://stackoverflow.com/questions/28157246/how-estimate-row-count-of-a-postgresql-view). 

## Checklist

### DIFF
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -3332,20 +3332,35 @@ long QgsPostgresProvider::featureCount() const
   // use estimated metadata even when there is a where clause,
   // although we get an incorrect feature count for the subset
   // - but make huge dataset usable.
+  long num = 0;
   if ( !mIsQuery && mUseEstimatedMetadata )
   {
-    sql = QStringLiteral( "SELECT reltuples::bigint FROM pg_catalog.pg_class WHERE oid=regclass(%1)::oid" ).arg( quotedValue( mQuery ) );
+    // parse explain output to estimate feature count
+    // we don't use pg_class reltuples because it returns 0 for view
+    sql = QStringLiteral( "EXPLAIN (FORMAT JSON) SELECT count(*) FROM %1%2" ).arg( mQuery, filterWhereClause() );
+    QgsPostgresResult result( connectionRO()->PQexec( sql ) );
+
+    const QString json = result.PQgetvalue( 0, 0 );
+    const QVariantList explain = QgsJsonUtils::parseJson( json ).toList();
+    const QVariantMap countPlan = explain.count() ? explain[0].toMap().value( "Plan" ).toMap() : QVariantMap();
+    const QVariantList queryPlan = countPlan.value( "Plans" ).toList();
+    const QVariant nbRows = queryPlan.count() ? queryPlan[0].toMap().value( "Plan Rows" ) : QVariant();
+
+    if ( nbRows.isValid() )
+      num = nbRows.toInt();
+    else
+      QgsLogger::warning( QStringLiteral( "Cannot parse JSON explain result to estimate feature count (%1) : %2" ).arg( sql, json ) );
   }
   else
   {
     sql = QStringLiteral( "SELECT count(*) FROM %1%2" ).arg( mQuery, filterWhereClause() );
+    QgsPostgresResult result( connectionRO()->PQexec( sql ) );
+
+    QgsDebugMsg( "number of features as text: " + result.PQgetvalue( 0, 0 ) );
+
+    num = result.PQgetvalue( 0, 0 ).toLong();
   }
 
-  QgsPostgresResult result( connectionRO()->PQexec( sql ) );
-
-  QgsDebugMsg( "number of features as text: " + result.PQgetvalue( 0, 0 ) );
-
-  long num = result.PQgetvalue( 0, 0 ).toLong();
   mShared->setFeaturesCounted( num );
 
   QgsDebugMsg( "number of features: " + QString::number( num ) );
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -3332,24 +3332,27 @@ long QgsPostgresProvider::featureCount() const
   // use estimated metadata even when there is a where clause,
   // although we get an incorrect feature count for the subset
   // - but make huge dataset usable.
-  long num = 0;
+  long num = -1;
   if ( !mIsQuery && mUseEstimatedMetadata )
   {
-    // parse explain output to estimate feature count
-    // we don't use pg_class reltuples because it returns 0 for view
-    sql = QStringLiteral( "EXPLAIN (FORMAT JSON) SELECT count(*) FROM %1%2" ).arg( mQuery, filterWhereClause() );
-    QgsPostgresResult result( connectionRO()->PQexec( sql ) );
+    if ( connectionRO()->pgVersion() >= 90000 )
+    {
+      // parse explain output to estimate feature count
+      // we don't use pg_class reltuples because it returns 0 for view
+      sql = QStringLiteral( "EXPLAIN (FORMAT JSON) SELECT count(*) FROM %1%2" ).arg( mQuery, filterWhereClause() );
+      QgsPostgresResult result( connectionRO()->PQexec( sql ) );
 
-    const QString json = result.PQgetvalue( 0, 0 );
-    const QVariantList explain = QgsJsonUtils::parseJson( json ).toList();
-    const QVariantMap countPlan = explain.count() ? explain[0].toMap().value( "Plan" ).toMap() : QVariantMap();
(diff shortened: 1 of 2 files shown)


troopa81:
Not sure about the backport 3.4 though. It's quite an easy fix, so I would say yes

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -3304,7 +3304,9 @@ long QgsPostgresProvider::featureCount() const
   // use estimated metadata even when there is a where clause,
   // although we get an incorrect feature count for the subset
   // - but make huge dataset usable.
-  if ( !mIsQuery && mUseEstimatedMetadata )
+  // don't use it on view because it will always return 0 (view as no rows)
```
mhugo: as => has

ponceta:
```suggestion
  // don't use it on view because it will always return 0 (view has no rows)
```

nyalldawson:
What's the consequence of the bug? Only fixes for crashes or data corruption should be backported now.

troopa81:
> What's the consequence of the bug?

If you choose the option "Use estimated table metadata" in PostGIS
connexion window and you want to add a view, the feature count will be
always 0.

It doesn't crash nor corrupt data. So, I remove the backport label.

jef-n:
Doesn't this make loading expensive views painfully slow?

strk:
It surely does. I suggest to use EXPLAIN and parse its output, for estimating number of rows contained in a query.

strk:
EXPLAIN could actually be also used for queries, dropping another performance bottleneck when user asked for "estimated" metadata...

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -3304,7 +3304,9 @@ long QgsPostgresProvider::featureCount() const
   // use estimated metadata even when there is a where clause,
   // although we get an incorrect feature count for the subset
   // - but make huge dataset usable.
-  if ( !mIsQuery && mUseEstimatedMetadata )
+  // don't use it on view because it will always return 0 (view has no rows)
+  const QgsPostgresProvider::Relkind type = relkind();
+  if ( !mIsQuery && mUseEstimatedMetadata && type != Relkind::View )
```
strk: Please try using the EXPLAIN method to estimate returned rows for queries and views.

tests/src/python/test_provider_postgres.py
```diff
@@ -1415,6 +1415,15 @@ def testCheckTidPkOnViews(self):
...
             self.assertNotEqual(f.attribute(0), NULL)
 
+    def testFeatureCountOnView(self):
+        """
+        Test feature count on view when with estimatedmetadata = true
+        """
+        self.execSQLCommand('CREATE VIEW qgis_test.somedataview AS SELECT * FROM qgis_test."someData"')
+        vl = QgsVectorLayer(self.dbconn + ' sslmode=disable key=\'pk\' estimatedmetadata=true srid=4326 type=POINT table="qgis_test"."somedataview" (geom) sql=', 'test', 'postgres')
+        self.assertTrue(vl.isValid())
+        self.assertEqual(self.source.featureCount(), vl.dataProvider().featureCount())
```
strk: The test is good to have, but since  you're asking for _estimated_ data, it's a bit picky to check the estimation to be exactly equal to the actual count.

strk (changes requested):
I think EXPLAIN should be used for the estimation

troopa81:
> It surely does. I suggest to use EXPLAIN and parse its output, for estimating number of rows contained in a query.

Isn't it dangerous to rely on EXPLAIN output to estimate the number of rows? How can we be sure that the output is not gonna change in the future, that someday `rows` will be changed in `number_of_rows` ?

strk:
> Isn't it dangerous to rely on EXPLAIN output to estimate the number of rows? How can we be sure that the output is not gonna change in the future, that someday `rows` will be changed in `number_of_rows` ?

Explain has XML output support since 9.0 (https://www.postgresql.org/docs/9.0/sql-explain.html).
Shall format change in future we'll deal with it. I don't see it that dangerous.

troopa81:
> Explain has XML output support since 9.0 (https://www.postgresql.org/docs/9.0/sql-explain.html).

Thanks for the tip, I was not aware about explain formats. I choose JSON over XML, because it seems to me easy to write 7352e3f .

tests/src/python/test_provider_postgres.py
troopa81: EXPLAIN is based on statistics so if the statistics are up to date, we must have the right count, no? I force ANALYZE in my tests to be sure.

strk: Statistics are not exact counts. When you run ANALYZE PostgreSQL makes an estimation about the number of "live" tuples in each block of data, and applies that to the count of blocks (something like that). I guess for small tables (as in our case) it may get an exact count, but I would not bet on it :)

strk: Uhm, the cure is worst than the disease ! How about keeping the test but use a tolerance for estimated rows count ?

jef-n:
should be guarded by `connectionRO()->pgVersion() >= 90200` as other occurrences of json.   And the featureCount is unreliable with "estimated metadata" anyway and widely used to avoid expensive queries.  I think we should just return -1 instead of 0, if there's no shortcut like EXPLAIN.

troopa81:
> should be guarded by connectionRO()->pgVersion() >= 90200

OK, I'll add it (with 90000, this feature was already available in [9.0](https://www.postgresql.org/docs/9.0/sql-explain.html) )

> I think we should just return -1 instead of 0, if there's no shortcut like EXPLAIN

And displaying -1 in the UI? or "Unknown"?

tests/src/python/test_provider_postgres.py
troopa81: Yes I agree but what tolerance. I was thinking about checking the returned value is just greater than 0, but can we be sure that the explain value is always greater than 0?

troopa81: I restore the test and keep only >0 has a test (if estimate fails, it returns -1)

troopa81:
I let the feature count display as it is, so if the estimate doesn't work, it will display -1

troopa81:
@strk is it OK for requested changes?

strk (approved):
Thanks, looks fine here

haubourg:
Anyone to merge this ?

haubourg:
thanks @elpaso !

Gustry:
@troopa81 I understand why you made this change for views, it was indeed correct.
But why you didn't keep the previous behavior for tables ?

The `Plan Rows` makes quite different results and is not following the `VACUUM ANALYSE` that people might run on their database.

Look this example:

```sql
EXPLAIN (FORMAT JSON) SELECT 1 FROM pgmetadata.contact
-- QUERY PLAN [{'Plan': {'Node Type': 'Seq Scan', 'Parallel Aware': False, 'Relation Name': 'contact', 'Alias': 'contact', 'Startup Cost': 0.0, 'Total Cost': 15.1, 'Plan Rows': 510, 'Plan Width': 4}}]
-- > 'Plan Rows': 510

VACUUM ANALYSE pgmetadata.contact;
SELECT reltuples as approximate_row_count  FROM pg_class WHERE oid = 'pgmetadata.contact'::regclass;
-- 0.0

SELECT COUNT(*) FROM pgmetadata.contact;
-- 0

EXPLAIN (FORMAT JSON) SELECT 1 FROM pgmetadata.contact;
-- Still 510
```

troopa81:
> @Gustry But why you didn't keep the previous behavior for tables ?

Because it would cost an extra request to check if mQuery is a view or a table (calling [relkind](https://github.com/qgis/QGIS/blob/e91abdc23f944318db5e3559d47b38e36a81ae9e/src/providers/postgres/qgspostgresprovider.cpp#L4893))

But, I reproduce your issue when you remove all your data from your table. I'll try to find some time to fix this.

Gustry:
> Because it would cost an extra request to check if mQuery is a view or a table

Hum, I though indeed we would have this info already in the context.

> But, I reproduce your issue when you remove all your data from your table.

My tables are indeed neatly all empty for now.

But I tried with a simple line, it's still wrong.

```sql
SELECT COUNT(*) FROM pgmetadata.contact;
-- 1

EXPLAIN (FORMAT JSON) SELECT 1 FROM pgmetadata.contact;
-- Still 510
```

Should I create a ticket ?

troopa81:
> Hum, I though indeed we would have this info already in the context.

It could be, I don't see any reason to retrieve this information several times

> But I tried with a simple line, it's still wrong.

My guess here is that the *Plan rows* approach is more approximate than the reltuples one. I tried with a table containing 3+06 rows and removed all rows. The reltuples did say 0 while the *Plan Rows* says 300. Both have changed and both are kind of true for an estimation. But the reltuples is more true :)

> Should I create a ticket ?

Yes, please

PR #32120, opened 2019-10-04
Fix Server getlegendgraphic duplicated symbols

elpaso:
Fix Server getlegendgraphic duplicated symbols
... and fix a few build warnings

Now uses the hit test from QGIS API.

Also: WMS contextual legend: fix unreported issue with reprojection:

When using contextual WMS legends with a different projection
in the canvas and in the layer, the extent was not transformed.


src/providers/wms/qgswmsprovider.cpp
```diff
@@ -3461,6 +3461,8 @@ QgsImageFetcher *QgsWmsProvider::getLegendGraphicFetcher( const QgsMapSettings *
   {
     scale = mapSettings->scale();
     mapExtent = mapSettings->visibleExtent();
+    QgsCoordinateTransform ct { mapSettings->destinationCrs(), crs(), mapSettings->transformContext() };
+    mapExtent = ct.transformBoundingBox( mapExtent );
```
nyalldawson: Better catch QgsCsException here

Issue #32053, opened 2019-09-29
PostGIS layers with type of GEOMETRY missing in QGIS browser and PostgreSQL Data Source Manager


palmerj:
- [x] Search through existing issue reports and gis.stackexchange.com to check whether the issue already exists
- [x] Test with a [clean new user profile](https://docs.qgis.org/testing/en/docs/user_manual/introduction/qgis_configuration.html?highlight=profile#working-with-user-profiles).
- [x] Create a light and self-contained sample dataset and project file which demonstrates the issue

**Describe the bug**
PostGIS layers with a type of GEOMETRY, in which the first 100 rows contain null geometries, are not showing in the QGIS browser. This is even when the Use Estimated Metadata option is NOT checked for the connection details.

**How to Reproduce**

1. Create a table in the PostGIS database:

```sql
DROP TABLE IF EXISTS public.foo;

CREATE TABLE public.foo(
   id INTEGER PRIMARY KEY NOT NULL,
   shape GEOMETRY(GEOMETRY, 4326)
);

INSERT INTO public.foo(id, shape)
SELECT g.id, NULL
FROM generate_series(1, 100) AS g (id);

INSERT INTO public.foo(id, shape)
SELECT g.id,'SRID=4326;MULTIPOLYGON(((170 -45, 171 -45,  171 -46, 170 -45 )))'
FROM generate_series(101, 200) AS g (id);

ANALYSE public.foo;
```

2. Establish a new PostGIS connection to the database in the QGIS Browser, by right clicking on the PostGIS node. Do not set any of the checkbox settings

3. Connect to the server, and expand the public schema. The table public.foo does not show.

**QGIS and OS versions**

QGIS version | 3.8.2-Zanzibar | QGIS code revision | 4470baa1a3
-- | -- | -- | --
Compiled against Qt | 5.12.3 | Running against Qt | 5.12.3
Compiled against GDAL/OGR | 2.4.1 | Running against GDAL/OGR | 2.4.1
Compiled against GEOS | 3.7.2-CAPI-1.11.2 | Running against GEOS | 3.7.2-CAPI-1.11.2 b55d2125
PostgreSQL Client Version | 11.3 | SpatiaLite Version | 4.3.0a
QWT Version | 6.1.4 | QScintilla2 Version | 2.11.1
Compiled against PROJ | 5.2.0 | Running against PROJ | Rel. 5.2.0, September 15th, 2018
OS Version | macOS Mojave (10.14) |   |  

**Additional context**

[...]

PR #32125, opened 2019-10-05
[db_manager] Allow selecting geometry type in context menu

### LINKED ISSUE #32119: Support for Postgis tables with multiple geometry types with Z
### Support for Postgis tables with multiple geometry types

When a Postgis table has more than one kind of geometry (Point, LineString, Polygon), DB Manager allow the user to add a layer with just Point features, or with just Polygon features, for example.

That behaviour is illustrated in the following DB Manager dialogue, where we can see three different layers that can be added for the same Potsgis table `edificio`.

Captura de ecrã de 2019-10-04 14-32-34

This works well, if the geometry are only 2d.

### Describe the bug

If the Postgis table has Z geometry (geometry with 3D), the layers are added to QGIS without the Z component. If the Postgis table has points with Z, lines with Z or polygons with Z, the layers are added, but the Z component is lost.

### How to Reproduce

To reproduce this limitation, we need to create a Postgis table with the `geometry` type with 3 dimensions.

```
CREATE TABLE public.curso_de_agua_z (
	identificador uuid NOT NULL DEFAULT uuid_generate_v1mc(),
	inicio_objeto timestamp NOT NULL,
	fim_objeto time NULL,
	delimitacao_conhecida bool NOT NULL,
	CONSTRAINT curso_de_agua_z_pkey PRIMARY KEY (identificador)
);

SELECT AddGeometryColumn ('public','curso_de_agua_z','geometria',3763,'GEOMETRY',3);
```

#### Add layer to QGIS

The previous table is shown and can be added to QGIS, using DB Manager.

layer has 3d

The layer is reported as having 3 dimensions.

#### Added layer does not have 3 dimensions

After adding the layer, it does not preserve the 3 dimensions.

This can be checked immediately on the layer tree (mouse over the layer with report it as `Polygon`, not as `PolygonZ`, for example).

Editing the layer, and showing the vertice editor, the 3 dimension does not appear.

Captura de ecrã de 2019-10-04 14-49-45

### QGIS and OS versions

Tested with 3.8.1-Zanzibar
```
POSTGIS="2.5.2 r17328" [EXTENSION] PGSQL="110" GEOS="3.7.0-CAPI-1.11.0 673b9939" PROJ="Rel. 5.2.0, September 15th, 2018" GDAL="GDAL 2.3.2, released 2018/09/21" LIBXML="2.9.4" LIBJSON="0.12.1" LIBPROTOBUF="1.3.1" RASTER
```

### Additional context

Support for 3D works, if the Postgis table has only one kind of geometry.

### LINKED ISSUE #32121: layer_styles table does not support different styles from different kinds of geometry
**Support for different geometry types**

QGIS supports Postgis tables with multiple geometry types. The same table can have, for example, points and polygons.

**Describe the bug**

For the same table, different layers are offered by DB Manager,

If the user adds a layer from such tables, the style for such layers should be saved on layer_styles table with the geometry type as key.

**How to Reproduce**

Add two layers from the same table.

Save both styles with the same name on the database.

**QGIS and OS versions**

3.8.1-Zanzibar

**Additional context**

Maybe we need to change the DDL of layer_styles.

layer_style

m-kuhn:
Peek 2019-10-05 11-12

When a layer has a generic "GEOMETRY" type, it will only be listed once in the db_manager
but with an additional context menu entry for advanced properties, where one can select
the geometry type and the CRS.

Fix #32119
References #30787

Edge case for feature vs bugfix

CC @gioman @jgrocha

### DIFF
--- a/python/plugins/db_manager/db_plugins/plugin.py
+++ b/python/plugins/db_manager/db_plugins/plugin.py
@@ -1158,7 +1158,7 @@ class VectorTable(Table):
         def selectedCrs():
             return crsSelector.crs()
 
-        addButton = QPushButton(self.tr('Load layer'))
+        addButton = QPushButton(self.tr('Load Layer'))
         addButton.clicked.connect(lambda: self.addLayer(selectedGeometryType(), selectedCrs()))
         btns = QDialogButtonBox(QDialogButtonBox.Cancel)
         btns.addButton(addButton, QDialogButtonBox.ActionRole)
--- a/python/plugins/db_manager/db_plugins/plugin.py
+++ b/python/plugins/db_manager/db_plugins/plugin.py
@@ -22,11 +22,25 @@ email                : brush.tyler@gmail.com
 from builtins import str
 from builtins import range
 
-from qgis.PyQt.QtCore import Qt, QObject, pyqtSignal
-from qgis.PyQt.QtWidgets import QApplication, QAction, QMenu, QInputDialog, QMessageBox
+from qgis.PyQt.QtCore import Qt, QObject, pyqtSignal, QByteArray
+
+from qgis.PyQt.QtWidgets import (
+    QFormLayout,
+    QComboBox,
+    QCheckBox,
+    QDialogButtonBox,
+    QPushButton,
+    QLabel,
+    QApplication,
+    QAction,
+    QMenu,
+    QInputDialog,
+    QMessageBox,
+    QDialog
+)
+
 from qgis.PyQt.QtGui import QKeySequence, QIcon
 
-from qgis.gui import QgsMessageBar
 from qgis.core import (
     Qgis,
     QgsApplication,
@@ -35,7 +49,17 @@ from qgis.core import (
     QgsWkbTypes,
     QgsProviderConnectionException,
     QgsProviderRegistry,
+    QgsVectorLayer,
+    QgsRasterLayer,
+    QgsProject,
+    QgsMessageLog
 )
+
+from qgis.gui import (
+    QgsMessageBarItem,
+    QgsProjectionSelectionWidget
+)
+
 from ..db_plugins import createDbPlugin
 
 
@@ -297,8 +321,6 @@ class Database(DbItemObject):
         return "row_number() over ()"
 
     def toSqlLayer(self, sql, geomCol, uniqueCol, layerName="QueryLayer", layerType=None, avoidSelectById=False, filter=""):
-        from qgis.core import QgsVectorLayer, QgsRasterLayer
-
         if uniqueCol is None:
             if hasattr(self, 'uniqueIdFunction'):
                 uniqueFct = self.uniqueIdFunction()
@@ -552,22 +574,6 @@ class Database(DbItemObject):
                 table = self.tablesFactory(t, self, schema)
                 ret.append(table)
 
-                # Similarly to what to browser does, if the geom type is generic geometry,
-                # we additionnly add three copies of the layer to allow importing
-                if isinstance(table, VectorTable):
-                    if table.geomType == 'GEOMETRY':
-                        point_table = self.tablesFactory(t, self, schema)
-                        point_table.geomType = 'POINT'
-                        ret.append(point_table)
-
-                        line_table = self.tablesFactory(t, self, schema)
-                        line_table.geomType = 'LINESTRING'
-                        ret.append(line_table)
-
-                        poly_table = self.tablesFactory(t, self, schema)
-                        poly_table.geomType = 'POLYGON'
-                        ret.append(poly_table)
-
         return ret
 
     def createTable(self, table, fields, schema=None):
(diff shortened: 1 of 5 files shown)


nyalldawson:
Yay!

But... I think we should consider a stance of discouraging features being added now to db manager without counterparts in browser. Browser is the future! Any chance this behavior could be ported to browser too?

python/plugins/db_manager/db_plugins/plugin.py
```diff
@@ -1100,6 +1115,71 @@ def runAction(self, action):
...
+                geomType += 'Z'
+            if mCheckBox.isChecked():
+                geomType += 'M'
+
+            return geomType
+
+        def selectedCrs():
+            return crsSelector.crs()
+
+        addButton = QPushButton(self.tr('Load layer'))
```
nyalldawson:
```suggestion
        addButton = QPushButton(self.tr('Load Layer'))
```

python/plugins/db_manager/db_plugins/plugin.py
```diff
@@ -1100,6 +1115,71 @@ def runAction(self, action):
...
+        dlg.exec_()
+
+        settings = QgsSettings()
+        settings.setValue("/DB_Manager/advancedAddDialog/geometry", dlg.saveGeometry())
+
+    def addExtraContextMenuEntries(self, menu):
+        """Called whenever a context menu is shown for this table. Can be used to add additional actions to the menu."""
+
+        if self.geomType == 'GEOMETRY':
+            menu.addAction(self.tr("Advanced add layer…"), self.showAdvancedVectorDialog)
```
nyalldawson:
```suggestion
            menu.addAction(self.tr("Add Layer (Advanced)…"), self.showAdvancedVectorDialog)
```

m-kuhn:
> But... I think we should consider a stance of discouraging features being added now to db manager without counterparts in browser. Browser is the future! Any chance this behavior could be ported to browser too?

Yes, but not before in ~2 weeks

jgrocha:
Awesome!

This fix #32119 (not #32121)

jgrocha:
@m-kuhn I didn't do a proper test (yet). I run the PR and it is a really nice addition. Thanks!

I just noticed a minor limitation (not sure): the projection is not recognized.

CRS not recognized

Table was created with a known CRS:

```sql
SELECT AddGeometryColumn ('public','curso_de_agua_z','geometria',3763,'GEOMETRY',3);
```

I'll check it more carefully tomorrow. The table does not appear in the `geometry_view`, but as we can see in the screenshot, the CRS is displayed in the table properties. It would be nice to get the proper CRS from the table.

m-kuhn:
> I just noticed a minor limitation (not sure): the projection is not recognized.

Fixed

If anyone sees this fit into the QGIS 3.10 release, feel free to approve and merge

jgrocha:
Thank you @m-kuhn. Awesome work, in such a short time (on YOUR weekend). The CRS is now detected.

Captura de ecrã de 2019-10-06 11-38-53

python/plugins/db_manager/db_plugins/plugin.py
```diff
@@ -1100,6 +1121,72 @@ def runAction(self, action):
...
+            self.mainWindow.infoBar.pushItem(QgsMessageBarItem(msgLabel, Qgis.Warning))
+
+    def showAdvancedVectorDialog(self):
+        dlg = QDialog()
+        dlg.setObjectName('dbManagerAdvancedVectorDialog')
+        settings = QgsSettings()
+        dlg.restoreGeometry(settings.value("/DB_Manager/advancedAddDialog/geometry", QByteArray(), type=QByteArray))
+        layout = QFormLayout()
+        dlg.setLayout(layout)
+        dlg.setWindowTitle(self.tr('Add layer {}').format(self.name))
```
nyalldawson: "Add Layer"

nyalldawson:
I think it's fine for 3.10

m-kuhn:
I'm out of office, I'd be glad if someone could apply this fix (using github suggestions?) and merge it for me :)

python/plugins/db_manager/db_plugins/plugin.py
elpaso:
```suggestion
        dlg.setWindowTitle(self.tr('Add Layer {}').format(self.name))
```

elpaso:
I don't have permissions to push to this repo :)

m-kuhn:
Are you sure?
IIRC I left it enabled to allow changes to my PRs from maintainers.

m-kuhn:
But thanks for the suggestion :)

elpaso:
Github said so:

immagine

gioman:
> CC @gioman @jgrocha

@m-kuhn I just compiled your branch https://github.com/m-kuhn/QGIS/tree/db_manager_select_geom_type very cool!

Just one thing, I still see "invalid projection" as in  https://github.com/qgis/QGIS/pull/32125#issuecomment-538698310 I guess I have compiled a branch with the fix for this too, but maybe not(?).

jgrocha:
> > CC @gioman @jgrocha
> 
> @m-kuhn I just compiled your branch https://github.com/m-kuhn/QGIS/tree/db_manager_select_geom_type very cool!
> 
> Just one thing, I still see "invalid projection" as in [#32125 (comment)](https://github.com/qgis/QGIS/pull/32125#issuecomment-538698310) I guess I have compiled a branch with the fix for this too, but maybe not(?).

@gioman I use `git checkout pr/32125` to get a local copy of a pull request. This creates a new local branch, which I use to compile and test the PR. 

I think you have to edit your `.git/config` to support this.
```
[remote "upstream"]
        url = https://github.com/qgis/QGIS.git
        fetch = +refs/heads/*:refs/remotes/upstream/*
        fetch = +refs/pull/*/head:refs/remotes/upstream/pr/*
```

In this specific case, I've tested the changes made by @m-kuhn and it does recognize the CRS, if defined.

gioman:
> 've tested the changes made by @m-kuhn and it does recognize the CRS, if defined.

all ok then :)

strk:
> > But... I think we should consider a stance of discouraging features being added now to db manager without counterparts in browser. Browser is the future! Any chance this behavior could be ported to browser too?
> 
> Yes, but not before in ~2 weeks

@m-kuhn did you ever start with that ? A ticket requesting the work exists in #32053, for the record

m-kuhn:
> @m-kuhn did you ever start with that ? A ticket requesting the work exists in #32053, for the record

I don't think so, no

PR #32128, opened 2019-10-06
Ensure polygon rings generated for rendering are always closed

nyalldawson:
Avoids issues with unclosed rings and use of spatial algorithms
during rendering (e.g. GEOS routines)


m-kuhn:
Issues like spikes?

m-kuhn (approved):
Code looks good.

nyalldawson:
> Issues like spikes

Not sure - I've only seen it exhibited as geos spewing warnings on the log

PR #32134, opened 2019-10-06
Get the username from current_user to fill layer_styles' owner field for Postgresql based connections

### LINKED ISSUE #32133: Owner not recorded in layer_styles when using Postgresql service based connections
**Describe the bug**

Styles can be saved in Postgresql (and using other providers, as well). When saved, the username used in the connection is stored in the `owner` field, in the table `layer_styles`.

If the Postgresql connection is based on [services](https://docs.qgis.org/testing/en/docs/user_manual/managing_data_source/opening_data.html#postgresql-service-connection-file), the username is not available from `QgsDataSourceUri`. **No information about the `owner` is stored in `layer_styles` in such case.**

**How to Reproduce**

* Create a connection based on pg_services.
* Store a style on the database
* The `owner` field is empty.

no owner using pg_services

**QGIS and OS versions**

QGIS 3.4, QGIS 3.8 and master

**Solution**

PR under preparation

jgrocha:
## Description

Fix #32133.

For connections using pg_services the connection's username can be retrieved using `CURRENT_USER` on the server side. This works for any kind of Postgresql connection, since it is managed on the Postgresql server side.

The DDL changed to:

```sql
    QgsPostgresResult res( conn->PQexec( "CREATE TABLE layer_styles("
                                         "id SERIAL PRIMARY KEY"
                                         ",f_table_catalog varchar"
                                         ",f_table_schema varchar"
                                         ",f_table_name varchar"
                                         ",f_geometry_column varchar"
                                         ",styleName text"
                                         ",styleQML xml"
                                         ",styleSLD xml"
                                         ",useAsDefault boolean"
                                         ",description text"
                                         ",owner varchar(63) DEFAULT CURRENT_USER"
                                         ",ui xml"
                                         ",update_time timestamp DEFAULT CURRENT_TIMESTAMP"
                                         ")" ) );
```

Both INSERT and UPDATE queries will set `owner` to `CURRENT_USER`. This supports previous lay_styles tables without the `owner` default value set.

## Checklist

### DIFF
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -4728,7 +4728,7 @@ bool QgsPostgresProviderMetadata::saveStyle( const QString &uri, const QString &
                                          ",styleSLD xml"
                                          ",useAsDefault boolean"
                                          ",description text"
-                                         ",owner varchar(63)"
+                                         ",owner varchar(63) DEFAULT CURRENT_USER"
                                          ",ui xml"
                                          ",update_time timestamp DEFAULT CURRENT_TIMESTAMP"
                                          ")" ) );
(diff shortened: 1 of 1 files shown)


src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -4759,9 +4759,9 @@ bool QgsPostgresProviderMetadata::saveStyle( const QString &uri, const QString &
   // two values are both replaced in the final .arg call of the string construction.
 
   QString sql = QString( "INSERT INTO layer_styles("
-                         "f_table_catalog,f_table_schema,f_table_name,f_geometry_column,styleName,styleQML,styleSLD,useAsDefault,description,owner%11"
+                         "f_table_catalog,f_table_schema,f_table_name,f_geometry_column,styleName,styleQML,styleSLD,useAsDefault,description%11"
                          ") VALUES ("
-                         "%1,%2,%3,%4,%5,XMLPARSE(DOCUMENT %16),XMLPARSE(DOCUMENT %17),%8,%9,%10%12"
```
jef-n: Why not just set `owner` to `current_user` instead of the uri param with current user?  If the table already exists the new default won't be set.

jgrocha: Thanks @jef-n for the review.

Really helpful your suggestion! It supports old `layer_styles` tables, with the default value declared. Implemented, as suggested.

jgrocha:
@jef-n Can you review this PR again? Do you think it is ok to merge?

PR #32138, opened 2019-10-07
Support for styles for different geometry types for the same Postgresql table

### LINKED ISSUE #32121: layer_styles table does not support different styles from different kinds of geometry

jgrocha:
## Description

Users can take advantage of Postgresql tables with generic geometry type, by choosing and adding each geometry type as a layer. After #32125 gets merged, there are no more limitations using this approach (layers with Z can be added as well).

But when the user save styles in the provider, the key to recover that style again does not include the geometry type. We are not able to distinguish different styles for different geometry types, on the same geometry column.

To be able to save styles for the Point based geometries separated from LineString based geometries, for example, I've added a new column `type` to `layer_styles` table (as `varchar (30)` borrowed from `geometry_columns` view).

The DDL is now:

```sql
    QgsPostgresResult res( conn->PQexec( "CREATE TABLE layer_styles("
                                         "id SERIAL PRIMARY KEY"
                                         ",f_table_catalog varchar"
                                         ",f_table_schema varchar"
                                         ",f_table_name varchar"
                                         ",f_geometry_column varchar"
                                         ",styleName text"
                                         ",styleQML xml"
                                         ",styleSLD xml"
                                         ",useAsDefault boolean"
                                         ",description text"
                                         ",owner varchar(63)"
                                         ",ui xml"
                                         ",update_time timestamp DEFAULT CURRENT_TIMESTAMP"
                                         ",type varchar(30)"
                                         ")" ) );
```

## Decisions

We just need to distinguish from `Point`, `LineString` and `Polygon` types. It is irrelevant if the layer is `Multi-` or have `Z` axis. The style will be the same for the same class.

That's why I choose to get the geometry type string with:
```cpp
QgsWkbTypes::geometryDisplayString( QgsWkbTypes::geometryType( dsUri.wkbType() ) )
```

For other layers (with just one geometry type per geometry column), this additional column `type` is not necessary, but it is nice to have it on the table.

For the record, the qml style itself also has the geometry type stored, but is quite complicated to use it for this kind of usage, to retrieve and manage styles.

Extract of qml for a polygon layer:

```xml
 <layerGeometryType>2</layerGeometryType>
```

## Screenshot

With this change, the `layer_styles` table will look like this:

style with geometry type

As we can see, even the style name can be the same for different geometry types of the same table.

This PR fixes #32121 

## Additional changes

As suggested by @Gustry, we change the previous `layer_styles` table, if this new `type` column is missing.

## Checklist


src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -4834,14 +4842,19 @@ bool QgsPostgresProviderMetadata::saveStyle( const QString &uri, const QString &
...
                                .arg( QgsPostgresConn::quotedValue( dsUri.schema() ) )
                                .arg( QgsPostgresConn::quotedValue( dsUri.table() ) )
-                               .arg( QgsPostgresConn::quotedValue( dsUri.geometryColumn() ) );
+                               .arg( QgsPostgresConn::quotedValue( dsUri.geometryColumn() ) )
+                               .arg( wkbTypeString );
+
     sql = QStringLiteral( "BEGIN; %1; %2; COMMIT;" ).arg( removeDefaultSql, sql );
   }
 
+  QgsDebugMsg( QStringLiteral( "-------------------------COSTA NOVA: %1 " ).arg( sql ) );
```
elpaso: can you remove this line?

elpaso:
Nice! 
I would be even nicer if you could add a test to `def testStyle(self):` in `test_provider_postgres.py` :)

elpaso:
You can probably re-use this test table: https://github.com/qgis/QGIS/blob/master/tests/testdata/provider/testdata_pg.sql#L629

Gustry:
Does existing `layer_styles` table get an update of the column?

src/providers/postgres/qgspostgresprovider.cpp
jgrocha: Ups! Fixed.

jgrocha:
> Nice!
> I would be even nicer if you could add a test to `def testStyle(self):` in `test_provider_postgres.py` :)

Good point. I've created a new `testStyleWithGeometryType` with layers from `geometries_table` as you suggested. Thank you.

jgrocha:
> Does existing `layer_styles` table get an update of the column?

Good point @Gustry. It should, right? I'll do my best to support it.

jgrocha:
> Does existing `layer_styles` table get an update of the column?

@Gustry I've added the `ALTER TABLE` if `layer_styles` exists, but do not have this new column yet.

Thank you for your input.

jgrocha:
I think can be included in 3.10.1.

gioman:
> I think can be included in 3.10.1.

is to be reopened? @jgrocha

jgrocha:
Yes, please. I think I can't reopen it.

jgrocha:
Without this PR, we can  not use the official cartography in Portugal. All new cartography is distributed in Postgresql (the national map agency moved from CAD to Postgresql). There is only one table for each theme, but features can have different geometry types in the same theme.

gioman:
> Without this PR, we can not use the official cartography in Portugal. All new cartography is distributed in Postgresql (the national map agency moved from CAD to Postgresql). There is only one table for each theme, but features can have different geometry types in the same theme.

@jgrocha correct. It was also be nicer if the DGT had a much more direct involvement in the QGIS project now that they mandates using PostGIS (good) and the best software to manage it is QGIS...

gioman:
@jgrocha this bug https://github.com/qgis/QGIS/issues/34982 is also probably important in the context you refers.

jgrocha:
> @jgrocha this bug #34982 is also probably important in the context you refers.

Yes, it is. Thanks for pointing out.

jgrocha:
Ping!

gioman:
Not stale.

jgrocha:
Still valid.

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -4863,15 +4886,17 @@ bool QgsPostgresProviderMetadata::saveStyle( const QString &uri, const QString &
     uiFileValue = QStringLiteral( ",XMLPARSE(DOCUMENT %1)" ).arg( QgsPostgresConn::quotedValue( uiFileContent ) );
   }
 
+  QString wkbTypeString = QgsPostgresConn::quotedValue( QgsWkbTypes::geometryDisplayString( QgsWkbTypes::geometryType( dsUri.wkbType() ) ) );
```
elpaso, line 4889: can it be const?
```suggestion
  const QString wkbTypeString = QgsPostgresConn::quotedValue( QgsWkbTypes::geometryDisplayString( QgsWkbTypes::geometryType( dsUri.wkbType() ) ) );
```

elpaso (approved):
LGTM, thank you!

src/providers/postgres/qgspostgresprovider.cpp
jgrocha: Sure! Thanks!

PR #32140, opened 2019-10-07
Improve layer unknown CRS handling

nyalldawson:
The current approach taken for handling layers with invalid/unknown CRS is horrible UX. By defaulting to a blocking dialog forcing users to make a CRS choice, we create situations resulting in effective application hangs. E.g.

- loading 100 layers from a folder into QGIS can result in 100 unknown CRS prompts. This is effectively a hang, since users either have to patiently click through 100 dialogs one by one or force kill QGIS
- running a lengthy script can be paused with a blocking dialog asking users to pick a CRS, even when the script does not require a valid CRS for a layer to proceed. Very frustrating after leaving QGIS running overnight and hoping for completed results in the morning!
- by forcing users to make CRS choices upfront, we don't allow them to do useful things like explore the layer's metadata or the coordinate ranges contained within the layer in order to make this choice. Deferring the choice allows users to do these steps at a later stage and make a better informed choice of CRS.
- By forcing users to make the choice upfront, we run the risk that users will click anything in the dialog just to get rid of it. And then the random choice is remembered and treated as the canonical value for the lifetime of that project, leaving no hints to the user that they should really investigate and fix the situation...
- Even worse, we have logic in place which falls back to a forced WGS84 CRS when a user cancels the dialog. This is very bad, since it's almost always the incorrect choice yet gives users NO visual indication of this. WGS84 is also bad(tm), and we, as knowledgeable members of the geospatial community, need to be guiding users AWAY from this CRS as quickly as possible.
- Plugins (and even core QGIS code) have been hacking around this situation by temporarily overriding user settings and forcing unknown layers to load in some random CRS choice. This is a horrible and fragile development practice (but with no choice other choice, it was the only available option for developers).

So this PR adds a new option for handling invalid layer CRS: "Leave as an unknown CRS (take no action)". It's now the default, with a new settings key created so that ALL users will be moved initially to this new default value.

When set, no validation of layer CRS is performed upfront. Instead, layers are loaded with no crs set. They gain a new indicator icon showing that the layer has no CRS (better icon needed!), and any coordinates in the layer are treated as purely raw coordinate values (just like if a project itself has no crs set). Users can click the indicator at a later stage to set the layer's CRS, OR can highlight multiple layers in the tree and set 100 layer's CRSes at once. Phew!

Additionally, new API hooks have been added to allow plugins to indicate that all CRS validation should be skipped when loading a particular layer, regardless of the user's actual settings. This avoids the need for the hacky temporary settings overrides used to date...

### DIFF
--- a/src/app/qgsoptions.h
+++ b/src/app/qgsoptions.h
@@ -44,6 +44,20 @@ class APP_EXPORT QgsOptions : public QgsOptionsDialogBase, private Ui::QgsOption
   public:
 
     /**
+     * Behavior to use when encountering a layer with an unknown CRS
+     * \since QGIS 3.10
+     */
+    enum UnknownLayerCrsBehavior
+    {
+      NoAction = 0, //!< Take no action and leave as unknown CRS
+      PromptUserForCrs = 1, //!< User is prompted for a CRS choice
+      UseProjectCrs = 2, //!< Copy the current project's CRS
+      UseDefaultCrs = 3, //!< Use the default layer CRS set via QGIS options
+    };
+    Q_ENUM( UnknownLayerCrsBehavior )
+
+
+    /**
      * Constructor
      * \param parent Parent widget (usually a QgisApp)
      * \param name name for the widget
--- a/src/core/mesh/qgsmeshlayer.h
+++ b/src/core/mesh/qgsmeshlayer.h
@@ -121,7 +121,7 @@ class CORE_EXPORT QgsMeshLayer : public QgsMapLayer
        *
        * \since QGIS 3.10
        */
-      bool allowInvalidCrs = false;
+      bool skipCrsValidation = false;
     };
 
     /**
--- a/src/core/mesh/qgsmeshlayer.h
+++ b/src/core/mesh/qgsmeshlayer.h
@@ -107,6 +107,21 @@ class CORE_EXPORT QgsMeshLayer : public QgsMapLayer
       {}
 
       QgsCoordinateTransformContext transformContext;
+
+      /**
+       * Controls whether the layer is allowed to have an invalid/unknown CRS.
+       *
+       * If TRUE, then no validation will be performed on the layer's CRS and the layer
+       * layer's crs() may be invalid() (i.e. the layer will have no georeferencing available
+       * and will be treated as having purely numerical coordinates).
+       *
+       * If FALSE (the default), the layer's CRS will be validated using QgsCoordinateReferenceSystem::validate(),
+       * which may cause a blocking, user-facing dialog asking users to manually select the correct CRS for the
+       * layer.
+       *
+       * \since QGIS 3.10
+       */
+      bool allowInvalidCrs = false;
     };
 
     /**
--- a/src/core/processing/qgsprocessingutils.cpp
+++ b/src/core/processing/qgsprocessingutils.cpp
@@ -214,7 +214,7 @@ QgsMapLayer *QgsProcessingUtils::loadMapLayerFromString( const QString &string, 
   {
     QgsVectorLayer::LayerOptions options { transformContext };
     options.loadDefaultStyle = false;
-    options.allowInvalidCrs = true;
+    options.skipCrsValidation = true;
     std::unique_ptr< QgsVectorLayer > layer = qgis::make_unique<QgsVectorLayer>( string, name, QStringLiteral( "ogr" ), options );
     if ( layer->isValid() )
     {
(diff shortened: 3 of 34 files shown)


nyalldawson:
@wonder-sk I know we've discussed this in the past - can you look over this PR? I'm aiming it for 3.10.

nirvn:
Without having reviewed the code yet, big +1 on the UI/UX front.

nirvn:
@nyalldawson , code looks good.

I believe this PR can also address the hack / TODO here:

https://github.com/qgis/QGIS/blob/master/src/gui/qgsbrowserdockwidget_p.cpp#L198

nyalldawson:
@nirvn that's already done here

nirvn:
Oh good, how did I miss that? :)

elpaso:
Do you think it would also be possible to have some logic (and a setting) that automatically picks up a sensible choice? For instance the first not deprecated transformation?

nyalldawson:
> Do you think it would also be possible to have some logic (and a setting) that automatically picks up a sensible choice? For instance the first not deprecated transformation?

What do you mean here? It's not related to transformations, but CRSes themselves. I don't think there's anyway we can (or should) try to guess these...

elpaso:
> > Do you think it would also be possible to have some logic (and a setting) that automatically picks up a sensible choice? For instance the first not deprecated transformation?
> 
> What do you mean here? It's not related to transformations, but CRSes themselves. I don't think there's anyway we can (or should) try to guess these...

My mistake, I thought it was about the other  blocking dialog that asks for the transformation when you load a layer that requires transformation.

nyalldawson:
Nope, that one already has an option for the logic you described

roya0045:
Out of curiosity, with my bad layer auto-finder it seems that the crs selection popup triggers twice, I assume that this PR would prevent both events?

nyalldawson:
@roya0045  yes, correct. Without this PR there's no way to workaround that in your code.

jgrocha:
@nyalldawson We already have the option to use a default CRS or the project CRS, to prevent the modal form. Users can already load 100 layers at once.

This PR adds another option **Leave as an unknown CRS**, right? This is will be the default, so it can be the first listed here:

handling unknown crs

nyalldawson:
@jgrocha 

Here's what it looks like now:
image

nirvn:
The default CRS for layers widget would probably be better placed below  "use default layer CRS".

nyalldawson:
@nirvn it doesn't just apply when that option is selected though -- it's also used when creating a completely new layer. That's why I want to avoid the visual association between those two.

nyalldawson:
I'd like to merge this asap for maximum user testing before release. If anyone has issues, please speak up quickly!

roya0045:
If I want to prevent the CRS double-tap with my changes to the bad layer handler, should I include some change as soon as this is merged or should I leave my PR as is and do a fix once this API has been proven?

src/app/qgsoptions.h
```diff
@@ -43,6 +43,20 @@ class APP_EXPORT QgsOptions : public QgsOptionsDialogBase, private Ui::QgsOption
     Q_OBJECT
   public:
 
+    /**
+     * Behavior to use when encountering a layer with an unknown CRS
```
nirvn: Semantically, is it an unknown CRS, or a layer with missing CRS?

src/core/mesh/qgsmeshlayer.h
```diff
@@ -107,6 +107,21 @@ class CORE_EXPORT QgsMeshLayer : public QgsMapLayer
       {}
 
       QgsCoordinateTransformContext transformContext;
+
+      /**
+       * Controls whether the layer is allowed to have an invalid/unknown CRS.
```
nirvn: invalid/unknown -> invalid/missing ?

src/core/processing/qgsprocessingutils.cpp
```diff
@@ -229,17 +207,14 @@ QgsMapLayer *QgsProcessingUtils::loadMapLayerFromString( const QString &string,
...
-  ( void )restorer; // no warnings
-
   QString name = fi.baseName();
 
   // brute force attempt to load a matching layer
   if ( typeHint == LayerHint::UnknownType || typeHint == LayerHint::Vector )
   {
     QgsVectorLayer::LayerOptions options { transformContext };
     options.loadDefaultStyle = false;
+    options.skipCrsValidation = true;
```
nirvn: This is so much cleaner, nice!

nirvn (approved):
OK, I've had another look at this, a little more closely this time. Other than a semantic question, this looks good.

I really like the skipCrsValidation option implementation here.

src/app/qgsoptions.h
nyalldawson: I guess it's either -- an un-readable CRS (e.g. a broken proj string, or WKT which can't be parsed) results in an invalid CRS, so the same situation as if the layer had no CRS to begin with.

nirvn: Makes sense.

nyalldawson: So do you think a wording change is required here?

nirvn: Na, I see your point now, it's fine.

IMHO, this is ready to merge, and as you pointed out, we'll need all the time between now and release day to stress test this most needed UI/UX behavior fix.

nyalldawson: Ok thanks... let's go!

SrNetoChan:
How about something like this for the icon?

image

SrNetoChan:
Or this:

image

SrNetoChan:
@nyalldawson @nirvn

PR #32141, opened 2019-10-07
Serve canvas background color / feature selection via proper QgsProject API

nirvn:
## Description

Code cleanup to avoid using a bunch of writeEntry / readNumEntry across the code, and serve a stable API setter/getter with test cases for 3rd parties to make use of.

## Checklist

### DIFF
--- a/src/gui/qgsmapcanvas.cpp
+++ b/src/gui/qgsmapcanvas.cpp
@@ -1823,14 +1823,6 @@ void QgsMapCanvas::setSelectionColor( const QColor &color )
     return;
 
   mSettings.setSelectionColor( color );
-  QgsDebugMsg( "XXXXXXXXXXXXXXXXX" );
-  QgsDebugMsg( "XXXXXXXXXXXXXXXXX" );
-  QgsDebugMsg( "XXXXXXXXXXXXXXXXX" );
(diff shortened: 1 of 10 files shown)


3nids:
really nice, wanted to do this for a while.
But what about going further and make a generic method in `QgsProject::const QString &scope, const QString &key, const QColor &color ) SIP_PYNAME( writeColorEntry );`

nirvn:
@3nids , identified this API gap while fixing QField's canvas background color situation :)

nyalldawson:
Much cleaner! One last request - can you test and ensure that the new signals aren't emitted when setting a color equal to the existing one?

nirvn:
@nyalldawson , done.

src/gui/qgsmapcanvas.cpp
```diff
@@ -1823,7 +1823,14 @@ void QgsMapCanvas::setSelectionColor( const QColor &color )
     return;
 
   mSettings.setSelectionColor( color );
-
+  QgsDebugMsg( "XXXXXXXXXXXXXXXXX" );
```
nyalldawson: ?

nirvn: Removed. It's my old fashioned / messy debugging style :)

PR #32145, opened 2019-10-07
Return false on commit failure from QgsOgrProvider::changeGeometryValues

### LINKED ISSUE #32144: False success from ogr provider's committChanges
Spinoff of https://github.com/qgis/QGIS/pull/31980#issuecomment-538909172

Basically, this snippet of tests/src/python/test_provider_ogr_gpkg.py:

       # Commit changes
        cbk = ErrorReceiver()
        vl.dataProvider().raiseError.connect(cbk.receiveError)
        self.assertTrue(vl.commitChanges())
        self.assertIsNone(cbk.msg)

It passes the first assert (commitChanges( ) returns true) but does not pass the second (no error is printed).

strk:
Fixes #32144

### DIFF
--- a/src/core/providers/ogr/qgsogrprovider.cpp
+++ b/src/core/providers/ogr/qgsogrprovider.cpp
@@ -2292,7 +2292,7 @@ bool QgsOgrProvider::changeAttributeValues( const QgsChangedAttributesMap &attr_
 
   if ( inTransaction )
   {
-    commitTransaction();
+    if ( ! commitTransaction() ) return false;
   }
 
   if ( mTransaction )
@@ -2483,7 +2483,7 @@ bool QgsOgrProvider::deleteFeatures( const QgsFeatureIds &id )
 
   if ( inTransaction )
   {
-    commitTransaction();
+    if ( ! commitTransaction() ) return false;
   }
 
   if ( mTransaction )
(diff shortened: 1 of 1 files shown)


rouault (approved):

strk:
Now to think about a testcase to secure this fix. What's a good reason for ::changeGeometryValues to fail committing ? Change file permission _after_ opening it ? (I stumbled upon this issue while playing with `walForSqlite3` but it'd be better for a test to be immune from that setting, which has its problems)

strk:
@rouault sorry I pushed another two changes handlng returns from commitTransaction, can you take another look ?

rouault:
> can you take another look ?

look fine.

> What's a good reason for ::changeGeometryValues to fail committing ?

Good question. I just tried to delete a sqlite file when a transaction was opened, but the commit succeeded !!! (just playing with command line sqlite3). This was in default journaling mode (not WAL)

rouault:
@strk Are you considering trying to add tests ? or should it be merged as it ?

strk:
I'm out of ideas as per generating a testcase so I'd say let's merge as-is.
Rather, do you think this should be backported @rouault ?

strk:
Well, marking for backport

strk:
Manual backport for 3.4 branch is in https://github.com/qgis/QGIS/pull/32198

PR #32154, opened 2019-10-08
Fix export to geopdf skips raster and mesh layers

### LINKED ISSUE #32148: Geopdf export does not include raster layers
**Describe the bug**

Trying the new geopdf export functionality, seems raster layers are not correctly included in the resulting file.

**How to Reproduce**

1. Setup a new map with some vectors layers and a raster backgound layer
2. Setup a new layout and open the export to pdf dialog
3. In this dialog check **_Always export as vectors_** and **_Create geopdf_**
4. Open the resulting pdf in any pdf reader, background raster is missing

**QGIS and OS versions**

QGIS 3.9

**Additional context**

Note that checking **_Always export as vectors_** option without _**Create geopdf**_ option will include the raster layer.

nyalldawson:
Fixes #32148


nirvn:
@nyalldawson , looking good. Is there no additional test case here due to travis not supporting GeoPDF export yet?

nyalldawson:
@nirvn I've just added one

src/core/qgsmaplayerrenderer.h
```diff
@@ -66,6 +67,13 @@ class CORE_EXPORT QgsMapLayerRenderer
     //! Gets access to the ID of the layer rendered by this class
     QString layerId() const { return mLayerID; }
 
+    /**
+     * Resets the painter to use for the job to a new \a painter.
```
PeterPetrik: how it is with ownership of the painter?

question: don't we kind of assume the painter cannot be null in the renderers?

thinking loud: I am wondering if the rendering context should not be part of the `QgsMapLayerRenderer`, since now we have a virtual setter in base class that does not in principle contains at all the context/painter.

nyalldawson: >  I am wondering if the rendering context should not be part of the QgsMapLayerRenderer

it probably should be, yes...

nyalldawson: @PeterPetrik does c1d156d look better to you? I've moved the render context member to the QgsMapLayerRenderer base class. There's a further cleanup which should be done in QGIS 4, which is to make the base class member a reference rather than a pointer, but I can't do that without breaking stable API.

PeterPetrik: looks right, thanks

PeterPetrik (approved):

PR #32156, opened 2019-10-08
[expression] Fix is_layer_visible function not taking scale-based visibility into account

nirvn:
## Description

Title says it all. 

## Checklist

### DIFF
--- a/src/core/expression/qgsexpressioncontextutils.cpp
+++ b/src/core/expression/qgsexpressioncontextutils.cpp
@@ -803,6 +803,10 @@ QgsExpressionContextUtils::GetLayerVisibility::GetLayerVisibility( const QList<Q
   }
 }
 
+QgsExpressionContextUtils::GetLayerVisibility::GetLayerVisibility()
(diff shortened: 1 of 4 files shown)


src/core/expression/qgsexpressioncontextutils.cpp
```diff
@@ -801,18 +802,25 @@ QVariant QgsExpressionContextUtils::GetLayerVisibility::func( const QVariantList
     return false;
   }
 
+  bool isVisible = false;
   QgsMapLayer *layer = QgsExpressionUtils::getMapLayer( values.at( 0 ), parent );
-  if ( layer )
+  if ( layer && mLayers.contains( layer ) )
```
nyalldawson: This isn't thread safe -- you'll need to record the min/max scales for each layer in the constructor and use them here

nirvn: Are you saying this function was _never_ thread safe until now?

nyalldawson (approved):

Issue #22883, opened 2016-05-31
"NULL" text in null field should disappear as soon as a field gets the focus


qgib:
Author Name: **Patrick Dunford** (@enzedrailmaps)
Original Redmine Issue: [14932](https://issues.qgis.org/issues/14932)
Affected QGIS version: 3.7(master)
Redmine category:forms

---
When I enter a value into a form or table all the fields are set to Null to begin with. When I type into the field the displayed word NULL turns into the text NULL. This means I have to select and delete that text if I don't want NULL appearing in that field.

The display of the word NULL in an empty field should not be with actual text that just makes extra work when entering data into fields.

---

- [null_values.gif](https://issues.qgis.org/attachments/download/12428/null_values.gif) (gcarrillo -)
- [null_values_2.gif](https://issues.qgis.org/attachments/download/14860/null_values_2.gif) (gcarrillo -)

signedav:
The tested behavior on current master is like this with GeoPackage source:
- text data in text edit widget: null representing value is grey/italic and on mouse click selected, but not removed
- integer in range widget: null representing value is black/nontitalic on mouse click disappears 
- integer in text edit widget: null representing value is grey/nonitalic and on text enter disappears
- real in range widget: null representing value is black/nontitalic on mouse click nothing happens (cursor in the middle of the word NULL)
- real in text edit widget: null representing value is grey/nonitalic and on text enter disappears

Do you confirm @gacarrillor ?

gacarrillor:
Hi @signedav, using the latest build I have available, I can confirm what you mentioned. 

Did you consider date/time fields? Because they also have this behaviour (see [GIF](https://issues.qgis.org/attachments/download/14860/null_values_2.gif)): Grey and italic NULL value and on click the NULL remains there.

signedav:
Okay @gacarrillor I haven't noticed that. 
Then it needs another fix.

- null representing value is grey/italic and on mouse click on null representing text it changes to current date placeholder text in gray, what can be edited -> that's fine.

- null representing value is grey/italic and on mouse click on null representing text nothing happens (cursor at the of the word NULL)  -> has to be fixed

- date is entered on clear the null representing text appears, and focus is lost (and then on click the behavior above) -> should not loose focus and display the current date placeholder text in grey

![datetimenull](https://user-images.githubusercontent.com/28384354/66113472-1b43df00-e5cd-11e9-8e1f-7578c38819a7.gif)

signedav:
[...]

Issue #31849, opened 2019-09-17
QgsDateTimeEdit doesn't behave correctly for NULL values


Gustry:
**Describe the bug**

Given this Python code:
```python
dialog = QDialog()
layout = QHBoxLayout()
dialog.setLayout(layout)

edit = QgsDateTimeEdit()
edit.clear()  # Should display NULL
layout.addWidget(edit)
dialog.exec_()
```
that we can run from the Python console, the widget is displaying the current date as a placeholder. `NULL` should be displayed instead.

Documentation of `clear`:
https://qgis.org/api/classQgsDateTimeEdit.html#ac764f28b1ceaa092de8df0e2bf24e18b

Another example:
```python
dialog = QDialog()
layout = QHBoxLayout()
dialog.setLayout(layout)
edit = QgsDateTimeEdit()
layout.addWidget(edit)
dialog.exec_()
```

Simple click on the "clear" button will indeed set `NULL`.
But a double click on the "clear "button" will set the current date as a placeholder instead of `NULL`.

**QGIS and OS versions**

QGIS version | 3.9.0-Master | QGIS code revision | 99a525c513
-- | -- | -- | --
Compiled against Qt | 5.12.2 | Running against Qt | 5.12.2
Compiled against GDAL/OGR | 2.4.0 | Running against GDAL/OGR | 2.4.0
Compiled against GEOS | 3.7.1-CAPI-1.11.1 | Running against GEOS | 3.7.1-CAPI-1.11.1 27a5e771
PostgreSQL Client Version | 11.5 (Ubuntu 11.5-1.pgdg19.04+1) | SpatiaLite Version | 4.3.0a
QWT Version | 6.1.4 | QScintilla2 Version | 2.10.4
Compiled against PROJ | 5.2.0 | Running against PROJ | Rel. 5.2.0, September 15th, 2018
OS Version | Ubuntu 19.04 |   |

signedav:
Hi @Gustry 

With the implementation done on PR #32158 this would look like this:
```
dialog = QDialog()
layout = QHBoxLayout()
dialog.setLayout(layout)

edit = QgsDateTimeEdit()
edit.clear()  # Should display NULL
layout.addWidget(edit)
edit = QgsDateTimeEdit()
layout.addWidget(edit)
dialog.exec_()
```
![python](https://user-images.githubusercontent.com/28384354/66399028-1d4bda80-e9df-11e9-82e3-ee68083e80f0.gif)

[...]

PR #32158, opened 2019-10-08
Null representation in QgsDateTimeEdit on getting focus

signedav:
There has been several issues in the behavior of getting and leaving focus on the `QgsDateTimeEditWidget` (`QgsDateTimeEdit`). 

This is definitely a bug fix and in my opinion it should be in the 3.10. But please, if you see any risks or inputs how it could be solved better, than tell me. I'm very open on that, since I have to confess that I did possibly not see through all the ideas, how it should have worked before.

### Issues (before)
datetimenull

#### 1. On getting focus
On mouse click on null representing text it changes to current date text in gray, what can be edited and that's fine. But when clicking outside the null representing text it keeps the null representing text, what has been not nice.

#### 2. On leaving focus
On leaving focus it set back the null representing text when the cursor has been on the end of the text. If the cursor has been inside the text, it did not change back to the null representing text, what has been not nice.

#### 3. On clear button
On clear button it set back the null representing text and leaves the focus. This has been not nice at all, since the user is most probably still editing the field when he presses the clear button.

The issue report for it is this: https://github.com/qgis/QGIS/issues/22883

### Solutions (now)
datetimenull-after

#### 1. On getting focus
I could not find out the logic that it changes the text on cursor inside the null representing text and otherwise not. But I decided that it's anyway more consistent, when it changes the text to the current date on getting focus (on `FocusInEvent` calling function `displayCurrentDate`).

#### 2. On leaving focus
On off Focus it calls the function `displayNull` and there it set's the cursor to the end of the text. I did not found out why the setting of the null representation text had no effect if the cursor has not been at the end. But like this it works and I see no danger in it.

#### 3. On clear button
On clear button it does not set the null representing text, but the current date instead (and the value NULL of course like before). It does not leave the focus anymore.

**Added some tests for that...**

### DIFF
--- a/python/gui/auto_generated/editorwidgets/qgsdatetimeedit.sip.in
+++ b/python/gui/auto_generated/editorwidgets/qgsdatetimeedit.sip.in
@@ -115,6 +115,8 @@ Signal emitted whenever the value changes.
 
     virtual void focusOutEvent( QFocusEvent *event );
 
+    virtual void focusInEvent( QFocusEvent *event );
+
     virtual void wheelEvent( QWheelEvent *event );
 
     virtual void showEvent( QShowEvent *event );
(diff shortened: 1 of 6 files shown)


Gustry:
Hi @signedav 

Did you see the issue https://github.com/qgis/QGIS/issues/31849 ? It might related to your PR. Can you try these Python snippets?

rldhont (approved):

3nids (approved):

signedav:
Thanks @rldhont and @3nids for reviewing 👍

PR #32161, opened 2019-10-08
Fix range widget negative integers -> NULL

### LINKED ISSUE #32149: PostGIS/geojson: Cannot input negative numbers in int fields
While working with PostGIS or geojson or WFS layers, when adding a new feature or editing the attribute table, a negative number (e.g. -10) is added in a field with datatype `int` it is automatically changed to `NULL`. 

In more detail:
When creating a new feature, in the dialog that appears when finished drawing, only values >0 are accepted. If negative values are entered they are automatically changed to `NULL`.
When editing the attribute table it's the same, only values >0 are accepted and negative values are changed to `NULL`.
But in the attribute table, while in edit mode, if the column is selected and in the expression panel a negative number is added and then the Update All or Update Selected or Update Filtered button is clicked then the values are changes correctly.

So, the issue is of course not with storing the negative numbers but with the way inserting them.

The column is `int` in postgresql which corresponds to the int32 of QGIS. When altering the column to `bigint` it corresponds to the int64 of QGIS and the problem is fixed. But it seems pointless to convert everything to `bigint`.

Note: This is not an issue when working with shp.

elpaso:
Fixes #32149

With tests

### DIFF
--- a/src/gui/editorwidgets/qgsrangewidgetwrapper.cpp
+++ b/src/gui/editorwidgets/qgsrangewidgetwrapper.cpp
@@ -142,10 +142,11 @@ void QgsRangeWidgetWrapper::initWidget( QWidget *editor )
     QgsSpinBox *qgsWidget = qobject_cast<QgsSpinBox *>( mIntSpinBox );
     if ( qgsWidget )
       qgsWidget->setShowClearButton( allowNull );
-    int minval = min.toInt();
+    int minval =  min.isValid() ? min.toInt() : std::numeric_limits<int>::lowest();
+    int maxval = max.isValid() ? max.toInt() : std::numeric_limits<int>::max();
+    uint stepval = step.isValid() ? step.toUInt() : 1;
     if ( allowNull )
     {
-      uint stepval = step.isValid() ? step.toUInt() : 1;
       // make sure there is room for a new value (i.e. signed integer does not overflow)
       int minvalOverflow = uint( minval ) - stepval;
       if ( minvalOverflow < minval )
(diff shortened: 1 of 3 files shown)


src/gui/editorwidgets/qgsrangewidgetwrapper.cpp
```diff
@@ -142,10 +142,11 @@ void QgsRangeWidgetWrapper::initWidget( QWidget *editor )
     QgsSpinBox *qgsWidget = qobject_cast<QgsSpinBox *>( mIntSpinBox );
     if ( qgsWidget )
       qgsWidget->setShowClearButton( allowNull );
-    int minval = min.toInt();
+    int minval =  min.isValid() ? min.toInt() : std::numeric_limits<int>::lowest();
```
signedav: It's okay IMO. But what is the reason doing it here and not do it on reading the config on line 90? https://github.com/qgis/QGIS/pull/32161/files#diff-76aae134610defb9f49b7af131458184R90

elpaso:
@signedav your comment disappeared, I hope I didn't delete it accidentally.


> In src/gui/editorwidgets/qgsrangewidgetwrapper.cpp:

> @@ -142,10 +142,11 @@ void QgsRangeWidgetWrapper::initWidget( QWidget *editor )
>     QgsSpinBox *qgsWidget = qobject_cast<QgsSpinBox *>( mIntSpinBox );
>    if ( qgsWidget )
>       qgsWidget->setShowClearButton( allowNull );
> -    int minval = min.toInt();
> +    int minval =  min.isValid() ? min.toInt() : std::numeric_limits<int>::lowest();

> It's okay IMO. But what is the reason doing it here and not do it on reading the config on line 90? https://github.com/qgis/QGIS/pull/32161/files#diff-76aae134610defb9f49b7af131458184R90

Because at that point we don't know if it's an int or a double. 

Btw, I didn't rewrite the whole method but I basically just copied the logic that was here: https://github.com/qgis/QGIS/pull/32161/files#diff-76aae134610defb9f49b7af131458184R98
and added a test.

signedav:
> Because at that point we don't know if it's an int or a double.

Ah I see. So if we do it at the point where reading config, we would need to decide there what type it is. Well okay, then I'm good with your solution. And it's good that it's done same like with the double values. Thanks.

signedav (approved):
So I approve this fix. @elpaso you can attach the line I suggested in the test if you want, but you don't need to. Thanks.

PR #31801, opened 2019-09-16
Application options: defer disabling of drivers after application restart (relates to #29212)

rouault:
This is a follow-up of https://github.com/qgis/QGIS/pull/31772

Driver de-registration is now defered after application restart to avoid
any risk of potential crashes if a dataset using a disabled driver was
already in use.

PR #31772, opened 2019-09-14
[GDAL provider] Be more robust to driver being disabled after provider creation (fixes #29212)

rouault:
This is still quite a dangerous practice, and this commit is more a band-aid
than a definitive fix.
In case a GDALDataset would hold and use a pointer to the GDALDriver that has
created it (which is not common in GDAL drivers though), crash would occur.
Safer but more involved fixes could be:
- to prevent disabling a driver that is in use
- to post-pone the effect of driver disabling to application restart, or when
  the last dataset using the driver has been closed

Issue #29212, opened 2019-02-26
QGIS crashes after deactivating a GDAL driver in use in the project


qgib:
Author Name: **Alexandre Neto** (@SrNetoChan)
Original Redmine Issue: [21395](https://issues.qgis.org/issues/21395)
Affected QGIS version: 3.4.5
Redmine category:data_provider/ogr

---
If you disable a GDAL driver in a QGIS session while having a file (using the driver) loaded, trying to open the layer's properties will originate a QGIS Crash.

Although it would be expected for the layer to stop working properly, it would be nicer if we could avoid the crash. I tried JP2 format and GeoTIFF, but This probably may affect all formats.

To replicate the issue:

1. Open the TIF file provided.

3. Go to Settings > Options > GDAL and disable the used driver (GeoTiff)

4. Try to open the layer properties.

Crash ID: 9061f923e086217e2f7aa8fb6923be2342099ae5

Stack Trace

QString::QString :
QtPrivate::trimmed :
QString::trimmed_helper :
QString::trimmed :
QgsProcessingParameterString::asScriptCode :
PyInit__core :
dataCapabilities :
QgsRasterLayer::htmlMetadata :
QgsRasterLayerProperties::sync :
QgsRasterLayerProperties::QgsRasterLayerProperties :
QgisApp::showLayerProperties :
QMetaObject::activate :
QAction::activate :
QMenu::actionGeometry :
QMenu::actionGeometry :
QMenu::mouseReleaseEvent :
QWidget::event :
QMenu::event :
QApplicationPrivate::notify_helper :
QApplication::notify :
QgsApplication::notify :
QCoreApplication::notifyInternal2 :
QApplicationPrivate::sendMouseEvent :
QSizePolicy::QSizePolicy :
QSizePolicy::QSizePolicy :
QApplicationPrivate::notify_helper :
QApplication::notify :
QgsApplication::notify :
QCoreApplication::notifyInternal2 :
QGuiApplicationPrivate::processMouseEvent :
QWindowSystemInterface::sendWindowSystemEvents :
QEventDispatcherWin32::processEvents :
DispatchMessageW :
DispatchMessageW :
QEventDispatcherWin32::processEvents :
qt_plugin_query_metadata :
QEventLoop::exec :
QMenu::exec :
QgsLayerTreeView::contextMenuEvent :
QWidget::event :
QFrame::event :
QAbstractItemView::viewportEvent :
QCoreApplicationPrivate::sendThroughObjectEventFilters :
[...]

PR #32162, opened 2019-10-08
Application options: defer disabling of drivers after application restart (relates to #29212)

rouault:
This is a follow-up of #31772

Driver de-registration is now defered after application restart to avoid
any risk of potential crashes if a dataset using a disabled driver was
already in use.

(revives https://github.com/qgis/QGIS/pull/31801 that was forgotten and address its review comments)

### DIFF
--- a/src/app/qgsoptions.cpp
+++ b/src/app/qgsoptions.cpp
@@ -2187,7 +2187,9 @@ void QgsOptions::optionsStackedWidget_CurrentChanged( int index )
 
 void QgsOptions::loadGdalDriverList()
 {
-  QStringList mySkippedDrivers = QgsApplication::skippedGdalDrivers();
+  QgsApplication::registerGdalDriversFromSettings();
+
+  const QStringList mySkippedDrivers = QgsApplication::skippedGdalDrivers();
   GDALDriverH myGdalDriver; // current driver
   QString myGdalDriverDescription;
   QStringList myDrivers;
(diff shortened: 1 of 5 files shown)


src/app/qgsoptions.cpp
```diff
@@ -2304,19 +2306,38 @@ void QgsOptions::loadGdalDriverList()
...
+      if ( deferredSkippedGdalDrivers.contains( driverName ) )
+      {
+        deferredSkippedGdalDrivers.removeAll( driverName );
+      }
     }
   }
-  mSettings->setValue( QStringLiteral( "gdal/skipList" ), QgsApplication::skippedGdalDrivers().join( QStringLiteral( " " ) ) );
+  if ( driverUnregisterNeeded )
+  {
+    QMessageBox::information( this, tr( "Drivers disabled" ),
```
nyalldawson: "Drivers Disabled" (capitalisation)

nyalldawson (approved):

src/app/qgsoptions.cpp
rouault: For my education, why capitalisation on Disabled ?

jef-n: Title case.

nyalldawson: We follow qt hig guidelines, which are nicely summarised at https://hig.kde.org/style/writing/capitalization.html

Basically - title case for dialog files, combo entries, menu actions, group titles. Sentence case for labels.

rouault: Thanks for the pointer. Must be an English specific thing. The French translation only capitalizes the first letter of the sentence.

PR #32165, opened 2019-10-08
do not stop package layer algo when an error occured with one layer

Gustry:
## Description

Some layers can fail to be imported into the geopackage.
Before, the algorithm would continue until the end with other layers. And only at the end an exception is raised.

@nyalldawson I know this has been discussed before, and I saw your comment: https://github.com/qgis/QGIS/blob/master/src/analysis/processing/qgsalgorithmpackage.cpp#L129

Let's not raise an exception, just report errors. Users can have the list of layers packaged in the output.
I'm using this algorithm in a python plugin so I need to programmatically know layers which haven't been imported. Ideally, we should have another output with not packaged layers ID?

Before:
```
Packaging layer 1/24: _val_raepa_etat_canal_ass
Packaging layer 2/24: _val_raepa_forme_canal_ass
Packaging layer 3/24: _val_raepa_precision_annee
Packaging layer 4/24: _val_raepa_type_intervention_ass
Packaging layer 5/24: Appareils
Packaging layer failed: Unsupported type for field _temp_data
Packaging layer 6/24: Canalisations
Packaging layer failed: Unsupported type for field _temp_data
Packaging layer 7/24: raepa_canalass_l
Packaging layer failed: Unsupported type for field _temp_data
Packaging layer 8/24: Ouvrages
Packaging layer failed: Unsupported type for field _temp_data
Packaging layer 9/24: sys_organisme_gestionnaire
Packaging layer 10/24: val_raepa_cat_canal_ae
Packaging layer 11/24: val_raepa_cat_canal_ass
Packaging layer 12/24: val_raepa_fonc_app_ae
Packaging layer 13/24: val_raepa_fonc_app_ass
Packaging layer 14/24: val_raepa_fonc_canal_ae
Packaging layer 15/24: val_raepa_fonc_canal_ass
Packaging layer 16/24: val_raepa_fonc_ouv_ae
Packaging layer 17/24: val_raepa_fonc_ouv_ass
Packaging layer 18/24: val_raepa_materiau
Packaging layer 19/24: val_raepa_mode_circulation
Packaging layer 20/24: val_raepa_qualite_anpose
Packaging layer 21/24: val_raepa_qualite_geoloc
Packaging layer 22/24: val_raepa_support_reparation
Packaging layer 23/24: val_raepa_typ_reseau_ass
Packaging layer 24/24: val_raepa_type_defaillance
Error obtained while packaging one or more layers.
Execution failed after 1.84 seconds

Loading resulting layers
Algorithm 'Package layers' finished
```

## Checklist

### DIFF
--- a/src/analysis/processing/qgsalgorithmpackage.cpp
+++ b/src/analysis/processing/qgsalgorithmpackage.cpp
@@ -169,7 +169,7 @@ QVariantMap QgsPackageAlgorithm::processAlgorithm( const QVariantMap &parameters
   }
 
   if ( errored )
-    throw QgsProcessingException( QObject::tr( "Error obtained while packaging one or more layers." ) );
+    feedback->reportError( QObject::tr( "Error obtained while packaging one or more layers." ) );
 
   QVariantMap outputs;
   outputs.insert( QStringLiteral( "OUTPUT" ), packagePath );


nyalldawson:
I disagree with this. While usually I'm in favour of forgiving algorithms, I think failure to save a layer which should be saved constitutes a valid fatal error which should never be ignored...

nirvn:
Out of curiosity, what's the field type that isn't supported?

rldhont:
The field is a postgresql hstore.

Gustry:
@nyalldawson Thanks for your inputs, I clearly understand your point. But how can I know which layer failed to import from a Python script?

I needed access to the `OUTPUT_LAYERS` list of layers which have been packaged.
Can we give this kind of info from the QgsProcessingFeedback?

nirvn:
Seems that if there's a problem to fix here, it's the handling of hstore field export. It should fallback to string, not error out.

nyalldawson:
> But how can I know which layer failed to import from a Python script?

There's no way to catch errors in an easily parseable way. You can subclass QgsProcessingFeedback and make feedback object which collects error messages, but these may be translated and aren't designed for parsing.

Seems like there's an API gap here, and we need a way of adding extra information to QgsProcessingException. Eg a dict of contextual values...

Gustry:
OK thanks for your inputs @nyalldawson !

@nirvn Yes, it makes sense! An HStore field should be exportable ;-)

PR #32166, opened 2019-10-08
Remove T attribute on project QGZ files on windows

### LINKED ISSUE #32118: Why would .qgz files need a temporary (T) attribute (Windows)
**Describe the bug**

I just discovered that project files (.qgz) on our Win Server are not backed up by our Synology backup application. The reason for this behaviour is the T attribute which is added to the file in the moment it is saved. It occurs both on the windows server and on my local windows computer. Only if I save the project file as .qgs it doesn't get that T attribute. Since project files are not truely temporary files, and I'm not aware of a configuration option I suspect it to be a bug.
(The phenomenon was described earlier on gis.stackexchange, but there was no answer whatsoever:
https://gis.stackexchange.com/questions/306458/qgz-files-and-replication-thereof)
**How to Reproduce**

**QGIS and OS versions**
OS: Win 10 Pro, Win Server 2019
QGIS: At least all versions since April 2019 including 3.8.3

**Additional context**

elpaso:
Fixes #32118

### DIFF
--- a/src/core/qgsarchive.cpp
+++ b/src/core/qgsarchive.cpp
@@ -21,6 +21,11 @@
 #include "qgsmessagelog.h"
 #include "qgsauxiliarystorage.h"
 
+
+#ifdef Q_OS_WIN
+#include <windows.h>
+#endif
+
 #include <QStandardPaths>
 #include <QUuid>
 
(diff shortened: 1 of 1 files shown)


nirvn:
Nice fix

nyalldawson (approved):

RudiU:
Thank you for that fix! However, it might well be that removing the T attribute comes too late for automatic backups. On our system I need an additional rename operation to 'inform' the NAS backup about the newly saved file.
So I'd suggest to move the attribute removal up, operating on the tmpfile before it is renamed. I suppose, rename should delete the tmpfile anyway, so there would be no need for it keeping the T attribute(?)

georegio:
I just identified the same T - Attribut as the problem with synchronizing with Synology. And windows cant find *.qgz files with the search bar as long as they have the T attribute. Do i understand the above thread correct, that the fix will be implemented in future qgis-versions or does the fix have to be implemented manualy on every computer that uses the files in the cloud?

PR #32169, opened 2019-10-09
Format legend name for remote rasters added via http/https/ftp

### LINKED ISSUE #31636: Better default legend names for remote rasters added via http/https/ftp
The functionality (to add remote rasters) is super cool, but I think the result could benefit from better layer names once added to the legend.

By default the name is something like:

`/vsicurl/https://myserver/myraster.mbtiles`

but if the remote address has authentication then things got even worst for example if I used a QGIS authentication profile:

`/vsicurl/https://myserver/myraster.mbtiles authcfg='xxxxxxx'`

and if I used basic http authentication

`/vsicurl/https://username:password@myserver/myraster.mbtiles`

If possible I think the layer name should be only the (remote) file name part, and leave the connection informations, possibly not showing any password) to be shown when hovering over the layer name, as we do for any other datasource.

jgrocha:
## Description

As described in the issue #31636, legend names can became quite huge and **expose the user credentials**. This PR tries to get the filename from the _uri_, discarding anything until the last slash and everything after the first space. It is based on the regular expression:

```cpp
QRegularExpression reRasterFile( "^/vsi(.+/)*([^ ]+)( .+)?$", QRegularExpression::CaseInsensitiveOption );
```

On the screenshot below, we can see the names before this PR and after the PR. For both names, we toggle the _formatLayerName_ advanced setting (so the legend gets capitalized).

remote raster legend name

Fix #31636

## Checklist

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -13430,12 +13430,12 @@ QgsRasterLayer *QgisApp::addRasterLayerPrivate(
   }
 
   QString shortName = name;
-  QRegularExpression reRasterFile( "^/vsi(.+/)*([^ ]+)( .+)?$", QRegularExpression::CaseInsensitiveOption );
+  QRegularExpression reRasterFile( QStringLiteral( "^/vsi(.+/)*([^ ]+)( .+)?$" ), QRegularExpression::CaseInsensitiveOption );
   QRegularExpressionMatch matchRasterFile = reRasterFile.match( name );
 
   if ( matchRasterFile.hasMatch() )
   {
-    if ( matchRasterFile.captured( 2 ).length() > 5 )
+    if ( matchRasterFile.captured( 2 ).length() > 0 )
     {
       shortName = matchRasterFile.captured( 2 );
     }
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -13429,8 +13429,20 @@ QgsRasterLayer *QgisApp::addRasterLayerPrivate(
     refreshBlocker = qgis::make_unique< QgsCanvasRefreshBlocker >();
   }
 
+  QString shortName = name;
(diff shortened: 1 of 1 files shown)


nirvn:
I'm +1 to get rid of the /vsi.../ as well as the credentials (the latter is a pretty serious security issue). 

I'll merge when later this am.

src/app/qgisapp.cpp
```diff
@@ -13429,8 +13429,20 @@ QgsRasterLayer *QgisApp::addRasterLayerPrivate(
     refreshBlocker = qgis::make_unique< QgsCanvasRefreshBlocker >();
   }
 
+  QString shortName = name;
+  QRegularExpression reRasterFile( "^/vsi(.+/)*([^ ]+)( .+)?$", QRegularExpression::CaseInsensitiveOption );
```
nirvn:
```suggestion
  QRegularExpression reRasterFile( QStringLiteral( "^/vsi(.+/)*([^ ]+)( .+)?$" ), QRegularExpression::CaseInsensitiveOption );
```

src/app/qgisapp.cpp
```diff
@@ -13429,8 +13429,20 @@ QgsRasterLayer *QgisApp::addRasterLayerPrivate(
...
+  QRegularExpressionMatch matchRasterFile = reRasterFile.match( name );
+
+  if ( matchRasterFile.hasMatch() )
+  {
+    if ( matchRasterFile.captured( 2 ).length() > 5 )
```
nirvn: Why greater than five here? E.g., m.tif is five characters. I'd merely insure that the captured group isn't empty.

jgrocha:
@nirvn Thank you for your review and valuable feedback. I've incorporated your suggestions.

nirvn:
@jgrocha , great job, thanks. If you feel like going the extra mile, a similar fix would be most welcome for remote vectors too.

PR #32171, opened 2019-10-09
Fix detection of GEOS release candidate versions

### LINKED ISSUE #32170: CMake Error at cmake/FindGEOS.cmake:162
CMake Error at cmake/FindGEOS.cmake:162
  STRING sub-command REGEX, mode MATCH needs at least 5 arguments total to
  command.

This is happening with geos 3.8.0rc3 installed

strk:
Closes #32170


strk:
I guess we could also consider "alpha" and "beta" suffixes, as they might show up at some point...

nyalldawson (approved):

PR #32175, opened 2019-10-09
Qgs quick/external widget update #2

sklencar:
Some improvements related to external resource widget and photopanel

- Added back button on PhotoPanel
- Split preview and camera output - preview's dimension depends on captured image and if the image doesn't fill a whole screen, camera output is still visible on the background (fixed by this PR)

Back button looks like this:
Screenshot_2019-10-09-14-23-48-257_uk co lutraconsulting

### DIFF
--- a/src/quickgui/images/ic_back.svg
+++ b/src/quickgui/images/ic_back.svg
@@ -1,10 +1,1 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!-- Generator: Adobe Illustrator 16.0.4, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
-<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
-<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
-	 width="30px" height="30px" viewBox="0 0 30 30" enable-background="new 0 0 30 30" xml:space="preserve">
-<path d="M18.125,3.047C18.49,2.683,18.934,2.5,19.453,2.5c0.521,0,0.965,0.183,1.328,0.547L22.5,4.844
-	c0.365,0.364,0.547,0.808,0.547,1.328s-0.182,0.938-0.547,1.25L15,15l7.5,7.5c0.365,0.364,0.547,0.808,0.547,1.328
-	s-0.182,0.964-0.547,1.328l-1.719,1.797c-0.363,0.364-0.807,0.547-1.328,0.547c-0.52,0-0.963-0.183-1.328-0.547L7.5,16.328
-	C7.137,15.964,6.953,15.521,6.953,15s0.184-0.964,0.547-1.328L18.125,3.047z"/>
-</svg>
+<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd"><svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" width="24" height="24" viewBox="0 0 24 24"><path d="M20,11V13H8L13.5,18.5L12.08,19.92L4.16,12L12.08,4.08L13.5,5.5L8,11H20Z" /></svg>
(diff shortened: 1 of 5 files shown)


src/quickgui/images/ic_back.svg
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 16.0.4, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 width="30px" height="30px" viewBox="0 0 30 30" enable-background="new 0 0 30 30" xml:space="preserve">
+<path d="M18.125,3.047C18.49,2.683,18.934,2.5,19.453,2.5c0.521,0,0.965,0.183,1.328,0.547L22.5,4.844
+	c0.365,0.364,0.547,0.808,0.547,1.328s-0.182,0.938-0.547,1.25L15,15l7.5,7.5c0.365,0.364,0.547,0.808,0.547,1.328
+	s-0.182,0.964-0.547,1.328l-1.719,1.797c-0.363,0.364-0.807,0.547-1.328,0.547c-0.52,0-0.963-0.183-1.328-0.547L7.5,16.328
+	C7.137,15.964,6.953,15.521,6.953,15s0.184-0.964,0.547-1.328L18.125,3.047z"/>
+</svg>
```
wonder-sk: Let's replace this image by official material icon

src/quickgui/plugin/qgsquickphotopanel.qml
```diff
@@ -38,12 +38,21 @@ Drawer {
   property var captureButtonIcon: QgsQuick.Utils.getThemeIcon("ic_camera_alt_border")
   property var confirmButtonIcon: QgsQuick.Utils.getThemeIcon("ic_check_black")
   property var cancelButtonIcon: QgsQuick.Utils.getThemeIcon("ic_clear_black")
+  property var backButtonSource: QgsQuick.Utils.getThemeIcon("ic_back")
```
wonder-sk: Let's move the getThemeIcon() calls to qgsquickfeatureformstyling.qml so it's easier to override them when needed...

src/quickgui/plugin/qgsquickphotopanel.qml
```diff
@@ -139,87 +150,113 @@ Drawer {
...
+          sourceSize.width: imageButtonSize
+          height: imageButtonSize
+          width: imageButtonSize
+          source: photoPanel.confirmButtonIcon
+        }
+      }
+    }
+
+    Rectangle {
+      id: backButton
```
wonder-sk: I would suggest to remove this circle - the back arrow is not properly aligned and it does not look great... better to have outline with the icon just like in the case of camera icon. And we will also get rid of the scaleRatio...

src/quickgui/plugin/editor/qgsquickexternalresource.qml
```diff
@@ -138,6 +138,13 @@ Item {
     }
   ]
 
+  onVisibleChanged: {
+    if (visible) {
+      image.currentValue = value
+      image.source = image.getSource()
+    }
+  }
+
```
wonder-sk: I am a bit worried this workaround may not work well - the change of the image source should not be dependent on the item visibility change. There is already image source being changed on 1. onCurrentValueChanged, 2. Component.onCompleted - so let's try to figure out the flow of events so we can have a more robust fix...

src/quickgui/plugin/qgsquickphotopanel.qml
```diff
@@ -38,12 +38,21 @@ Drawer {
...
   property var confirmButtonIcon: QgsQuick.Utils.getThemeIcon("ic_check_black")
   property var cancelButtonIcon: QgsQuick.Utils.getThemeIcon("ic_clear_black")
+  property var backButtonSource: QgsQuick.Utils.getThemeIcon("ic_back")
   property real imageButtonSize: 45 * QgsQuick.Utils.dp
   property real buttonSize: imageButtonSize * 1.2
   property var buttonsPosition
 
   signal confirmButtonClicked(string path, string filename)
 
+  function cancelEvent() {
```
wonder-sk: can we rename this method - e.g. discardCapturedImage() would make it more clear that we're not canceling the whole panel, only the image we already took...

src/quickgui/plugin/qgsquickphotopanel.qml
```diff
@@ -63,6 +72,7 @@ Drawer {
       camera.stop()
       photoPreview.visible = false
     }
+    photoPreview.source = ""
```
wonder-sk: I don't fully understand why we need to set photo preview source to empty string here - especially why both when making the item visible and also when hiding it?

wonder-sk (approved):

PR #32187, opened 2019-10-10
Adds better Layer panel indicator for noCRS

SrNetoChan:
## Description

Adds a new icon for the NoCRS Layer indicator in the layers panel as requested by @nyalldawson  in #32140.

It intends to be a cleaner version of the one used now. This is how it looks:

image

## Checklist


nirvn:
Can you provide a before vs after shot?

SrNetoChan:
@nirv

Before on the bottom, after on the top

image

looking at the icons in more detail.

Before

image

After

image

This icons are shown at a very small size, which makes it hard to create clear images. I wanted to keep the CRS icon reference, but I am not usre it's intuitive enough.

nirvn:
The question mark is pretty hard to decipher in-place. Could you try to make it bigger?

SrNetoChan:
I made it a bit bigger, but I am not sure you notice much diference.

SrNetoChan:
image

roya0045:
Could you make the outline of the CRS symbol more apparent?
 In your initial choices I felt the outline was too apparent and was strange with the ?, but having a third shade of grey might help make the outline of the underlying symbol clearer.

SrNetoChan:
@roya0045 I am trying to keep the same color "pallete" as the other indicators, which only have dark gray and soft gray.. with this icon size, I am thinking if I shouldn't simply fill the CRS symbol with the soft gray. Besides, the question mark kinda overlaps the the CRS outline. Here's a zoomed version:

image

roya0045:
What would it look as a small icon like if you inverted the two grays?

SrNetoChan:
@roya0045 Does not work well

image

roya0045:
Indeed, I think this is why the previous icon had more shades of grey. I feel like the issue is that one symbol is too simple and the other too complex, and using only two tones highlights one or the other.

SrNetoChan:
The previous icon was designed by me, but for another context. Those shades won't be visible in such small icon size. 

@nirvn do we really need those boxes around the icons? I understand that the original idea was to make it opaque so its no confused with the background, but the icons get soo small...

nyalldawson:
Merging as is for hard freeze - can be tweaked in follow up prs

PR #32188, opened 2019-10-10
When using estimated metadata, skip NULL geometries

strk:
This allows determining srid/type of columns having a NULL
geometries in the first 1000 records, at the cost of possibly
scanning whole tables when geometries are ALL nulls.

References #32053

### DIFF
--- a/src/providers/postgres/qgspostgresconn.cpp
+++ b/src/providers/postgres/qgspostgresconn.cpp
@@ -1656,10 +1656,11 @@ void QgsPostgresConn::retrieveLayerTypes( QVector<QgsPostgresLayerProperty *> &l
       // our estimatation ignores that a where clause might restrict the feature type or srid
       if ( useEstimatedMetadata )
       {
-        table = QStringLiteral( "(SELECT %1 FROM %2%3 LIMIT %4) AS t" )
+        table = QStringLiteral( "(SELECT %1 FROM %2%3 WHERE %4 IS NOT NULL LIMIT %5) AS t" )
                 .arg( quotedIdentifier( layerProperty.geometryColName ),
                       table,
                       layerProperty.sql.isEmpty() ? QString() : QStringLiteral( " WHERE %1" ).arg( layerProperty.sql ) )
+                .arg( layerProperty.geometryColName )
                 .arg( GEOM_TYPE_SELECT_LIMIT );
       }
       else if ( !layerProperty.sql.isEmpty() )
--- a/src/providers/postgres/qgspostgresconn.cpp
+++ b/src/providers/postgres/qgspostgresconn.cpp
@@ -2112,7 +2112,7 @@ QgsDataSourceUri QgsPostgresConn::connUri( const QString &connName )
   }
   QString database = settings.value( key + "/database" ).toString();
 
-  bool useEstimatedMetadata = settings.value( key + "/estimatedMetadata", false ).toBool();
+  bool estimatedMetadata = useEstimatedMetadata( connName );
   QgsDataSourceUri::SslMode sslmode = settings.enumValue( key + "/sslmode", QgsDataSourceUri::SslPrefer );
 
   QString username;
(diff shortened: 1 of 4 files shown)


elpaso:
@strk there is now a new API that abstracts the connections handling, would you please check if it's possible to use it here, and if that API also honors the estimated metadata option?

The new API is here: https://qgis.org/api/classQgsAbstractDatabaseProviderConnection.html

There are methods to retrieve table information as well as schemas and for filtering by table type.

strk:
Wow, interesting. Are you suggesting that that QgsPostgresProviderConnection is supposed to replace QgsPostgresConn ?

elpaso:
@strk, no, the new API is more higher level, but it should eventually replace all dataitems provider specific code. 

We should be able to create data items using the new API (which is in core) instead of the provider specific implementation (which is moved to private QgsPostgresProviderConnection).

I didn't really dig into this PR, I just wanted to make sure you was aware about the new API because the idea is to handle connections (I mean connections storage in QGIS settings) and metadata about connections in a more abstracted way.

That's why I think that the patch you are working on should also be tested in the new API or we will eventually have the same problem when we will throw away provider's dataitems specific code (not sure when this will happen though).

strk:
@elpaso if you see my change is very small, I added, after this line:

>   bool dontResolveType = QgsPostgresConn::dontResolveType( mConnectionName );

This other line:

>   bool estimatedMetadata = QgsPostgresConn::useEstimatedMetadata( mConnectionName );

So not changing things drastically.

Rewriting the whole data item class to use the new API is out of scope in this PR.

Looking at the API, btw, suggests that the `useEstimatedMetadata` is not available as a `TableFlag` (maybe it should?)

strk:
Ah @elpaso now I see what you mean. Indeed `QgsPostgresProviderConnection` is `not` honouring the `estimatedMetadata` configuration. It registers it as a configuration but then doesn't use it from `QgsPostgresProviderConnection::tables`, where it hard-codes a `true` value for it. Will push a commit fixing that

strk:
@elpaso do you know which code path uses that new API for me to test if the changes I do have an actual effect ?

palmerj:
@strk I guess the remaining issue here is an empty table with a typeless geometry column not showing up in the browser.

strk:
@palmerj I think that one would be better handled by another PR, maybe looking at https://github.com/qgis/QGIS/pull/32125

palmerj:
Maybe but that PR is scoped to the db manager.

strk:
This one was scoped to "estimated metadata" doing a better estimation (skipping NULLS). It was already overloaded with the fix to actually use the "estimated metadata" configuration ...

palmerj:
Ok I will comment back in #32053

palmerj:
Thanks @strk this good to me in my testing

palmerj:
Hang on. When I right click on a large layer in the browser and select `Layer Properties...` it runs a query like this:

```
SELECT 0, array_agg(st_srid("shape")::text || ':' || UPPER(geometrytype("shape")))  
FROM "xxx"."xxxx"
```

and QGIS is hangs and the dialog never loads. Layers with a small number of rows are ok. Maybe this is related to the changes in https://github.com/qgis/QGIS/commit/bbdbca41?

Couple of things:

1. For a table with many rows this bring back a single row with an array with a length of many characters (6 million in my case). Maybe the query should have a DISTINCT statement to only the return unique SRID, GEOMETRY type combinations? e.g

```
SELECT 0, array_agg(DISTINCT st_srid("shape")::text || ':' || UPPER(geometrytype("shape")))  
FROM "xxx"."xxxx"
```

2. I have the estimated metadata flag set for the connection. Shouldn't the layer properties dialog also apply the `limit 100` to the SQL when loading the layer metadata?

palmerj:
I have raised a follow-up issue for this https://github.com/qgis/QGIS/issues/32223

Issue #32193, opened 2019-10-10
The recent projects menu marks projects from custom storages as unavailable.


borysiasty:
Since #31386, the recent project menu grays out unavailable projects. The problem is that menu is created before plugins have chance to register custom storages, so all projects coming from custom storages are unrecognized and marked unavailable. Later, if I open any project, that menu is updated and all those projects become properly available.

Note it only affects the menu. The Welcome Page is working properly, as the project existence is examined later, in QgsRecentProjectItemsModel::flags.

For most users this problem may be irrelevant, but for corporations using custom storages extensively this regression is too confusing and annoying to let it slip into 3.10 ;)

The best solution would be probably to listen to a signal emited from QgsProjectStorageRegistry::registerProjectStorage and call QgisApp::updateRecentProjectPaths()

An alternative quick hack would be to simply call QgisApp::updateRecentProjectPaths() later, when plugins are loaded.

borysiasty:
@jgrocha @nyalldawson - It's your territory ;) I made a PR with the second dirty solution.

PR #32194, opened 2019-10-10
Fix recent project menu showing projects from custom storages as unavailable.

borysiasty:
This is a quick fix for  #32193. 
The recent projects menu is created after plugins registered possible custom storages.


borysiasty:
As we're in hard freeze now, I'm going to equip my clients (affected ones only) with a Python patch that replaces the menu with a fixed one, however, it would be nice to fix it for 3.10.1.

PR #32215, opened 2019-10-11
Use optimised point getters/setters on GEOS >= 3.8

nyalldawson:
Should squeeze some more performance out of QGIS <-> GEOS conversions.

@dbaston look good to you?

### DIFF
--- a/src/core/geometry/qgsgeos.cpp
+++ b/src/core/geometry/qgsgeos.cpp
@@ -1225,12 +1225,19 @@ std::unique_ptr<QgsLineString> QgsGeos::sequenceToLinestring( const GEOSGeometry
   double *m = mOut.data();
   for ( unsigned int i = 0; i < nPoints; ++i )
   {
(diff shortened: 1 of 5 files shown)


dbaston:
Yep. That #if bit is a mouthful though, maybe think about something like

```
#define HAVE_GEOS_38 defined(GEOS_VERSION_MAJOR) && defined(GEOS_VERSION_MINOR) && (GEOS_VERSION_MAJOR>=3) && (GEOS_VERSION_MINOR>=8 || GEOS_VERSION_MAJOR>3)

#if HAVE_GEOS_38
// new way
#else
// old way
#endif
```

Or you could conditionally define `GEOSCoordSeq_getXY` , etc. within QGIS if you have GEOS < 38. Just some thoughts to try and keep the call sites clean.

nyalldawson:
@dbaston you're right -- I've pushed a considerably simpler version

PR #8243, opened 2018-10-19
[Bugfix][Server][WFS] In GML geometry has all to be multi

rldhont:
## Description
If in the layer level, the geometry is defined as multi, in the GML all the geometry has to be converted to multi. It's needed to validate GML with schema provided by DescribeFeatureType request.

## Checklist

- [ ] Commit messages are descriptive and explain the rationale for changes
- [ ] I have read the [QGIS Coding Standards](https://docs.qgis.org/testing/en/docs/developers_guide/codingstandards.html) and this PR complies with them
- [ ] This PR passes all existing unit tests (test results will be reported by travis-ci after opening this PR)
- [ ] New unit tests have been added for core changes
- [ ] I have run [the `scripts/prepare-commit.sh` script](https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTING.md#contributing-to-qgis) before each commit

PR #32216, opened 2019-10-11
[Server] WFS Service: force Geometry to Multi in GML to be compliant with XML schema provided by DescribeFeatureType

rldhont:
## Description
It's an adapted to QGIS 3 forward porting of https://github.com/qgis/QGIS/pull/8243

If in the layer level, the geometry is defined as multi, in the GML all the geometry has to be converted to multi. It's needed to validate GML with schema provided by DescribeFeatureType request.

## Checklist

### DIFF
--- a/src/server/services/wfs/qgswfsgetfeature.cpp
+++ b/src/server/services/wfs/qgswfsgetfeature.cpp
@@ -1307,32 +1307,23 @@ namespace QgsWfs
 
         QDomElement geomElem = doc.createElement( QStringLiteral( "qgs:geometry" ) );
         QDomElement gmlElem;
+        QgsGeometry cloneGeom( geom );
         if ( params.geometryName == QLatin1String( "EXTENT" ) )
         {
-          QgsGeometry bbox = QgsGeometry::fromRect( geom.boundingBox() );
-          gmlElem = QgsOgcUtils::geometryToGML( bbox, doc, prec );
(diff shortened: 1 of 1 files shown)


rldhont:
@nyalldawson why do you consider to freeze this PR while it has to correct a regression ?

I have to add tests to avoid futur regression.

nyalldawson:
@rldhont 
>  why do you consider to freeze this PR while it has to correct a regression ?

Check the release roadmap -- 3.10.0 entered hard freeze today, which means only 100% mission critical bug fixes (e.g. crash or data corruption) will be permitted for the next two weeks.

rldhont:
Thanks to pointing it.

If I want to fix 3.10.x, I have to wait for 3.10.0 release, isn't it ?

Can I propose a PR, to fix the issue in 3.4.x ?

nyalldawson:
> If I want to fix 3.10.x, I have to wait for 3.10.0 release, isn't it ?

Yes

> Can I propose a PR, to fix the issue in 3.4.x ?

I think you need to wait till after 3.4.13 now

rldhont:
I cannot provide data to create tests.

Par exemple, les fichiers de forme ESRI ou GeoJSON peuvent contenir des géométries simples et multiples, mais OGR fournit toutes les géométries en mode multi. So I cannot build tests based on files.

The issue has been identified with an Oracle Layer. The Oracle provider authorizes to mix pulti and simple in the same column.

PR #32229, opened 2019-10-13
Only pick distinct srid/type combinations from vector layers

### LINKED ISSUE #32223: QGIS Layer property dialog is slow to load from the browser for PostGIS layers
Checklist before submitting

- [x] Search through existing issue reports and gis.stackexchange.com to check whether the issue already exists
- [x] Test with a [clean new user profile](https://docs.qgis.org/testing/en/docs/user_manual/introduction/qgis_configuration.html?highlight=profile#working-with-user-profiles).
- [x] Create a light and self-contained sample dataset and project file which demonstrates the issue

**Describe the bug**
Caused by #32002. When I right click on a large layer in the browser and select `Layer Properties...` it runs a PG query like this:

```sql
SELECT 0, array_agg(st_srid("shape")::text || ':' || UPPER(geometrytype("shape")))  
FROM "xxx"."xxxx"
```

and QGIS is hangs when built in debug mode and the dialog never loads. Layers with a small number of rows are ok.  I see in the debug log that I get repeated messages like:

src/providers/postgres/qgspostgresconn.cpp:1759 : (retrieveLayerTypes) [0ms] Analyzing layer's xxx.xxx.xxxx sridAndType 4167:POLYGON against 2 found combinations
src/providers/postgres/qgspostgresconn.cpp:1845 : (retrieveLayerTypes) [0ms] Pre-existing compatible combination 1/2 found for layer xxx.xxx.xxxx 

Couple of things:

1. For a table with many rows this bring back a single row with an array with a length of many characters (6 million in my case). I DISTINCT statement should be added to the query to only the return unique SRID, GEOMETRY type combinations? e.g

```sql
SELECT 0, array_agg(DISTINCT st_srid("shape")::text || ':' || UPPER(geometrytype("shape")))  
FROM "xxx"."xxxx"
```

2. I have the estimated metadata flag set for the connection. The layer properties dialog should also apply the `limit 100` to the SQL when loading the layer metadata

See change 
https://github.com/qgis/QGIS/commit/bbdbca41#diff-0cb4da952909b6ec6b33ef1c353df79eR1675

**QGIS and OS versions**

All versions

strk:
Closes #32223

### DIFF
--- a/src/providers/postgres/qgspostgresconn.cpp
+++ b/src/providers/postgres/qgspostgresconn.cpp
@@ -594,7 +594,7 @@ bool QgsPostgresConn::getTableInfo( bool searchGeometryColumnsOnly, bool searchP
     else if ( columnTypeInt == SctGeography ) columnType = SctGeography;
     else if ( columnTypeInt == SctTopoGeometry ) columnType = SctTopoGeometry;
     else if ( columnTypeInt == SctPcPatch ) columnType = SctPcPatch;
-    else if ( columnTypeInt == SctRaster ) columnType = SctPcPatch;
+    else if ( columnTypeInt == SctRaster ) columnType = SctRaster;
     else
     {
       QgsDebugMsg( QStringLiteral( "Unhandled columnType index %1" )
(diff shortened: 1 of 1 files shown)


strk:
I've also found raster query was bogus. I guess it would be useful to improve the existing testcase, although I'm afraid it uses the new API

PR #32232, opened 2019-10-13
Hide instead of disable useless widgets for text brackground with marker symbol

DelazJ:
avoiding people trying desperately to enable these settings (fill/stroke color/units) while they can't.
current state:
image


DelazJ:
@nyalldawson could we backport this one too?

PR #32234, opened 2019-10-13
 [WFS provider] Do not expand authentication config to username/password when adding layers from source select or browser

rouault:
Related suggested QGIS 4.0 API change : https://github.com/qgis/qgis4.0_api/issues/100


nyalldawson (approved):

nyalldawson:
@rouault can you rebase this fix please?

rouault:
Those 2 commits have already been merged into master per the OAPIF provider pull request. This PR is replaced by https://github.com/qgis/QGIS/pull/32436 which is the same but with 3.10 as the target

PR #32238, opened 2019-10-14
commenting noisy warnings

3nids:
These are annoying and prevent from correctly testing QGIS nightlies.
Will be reverted when #32176 is merged

### DIFF
--- a/src/gui/attributetable/qgsfeaturelistview.cpp
+++ b/src/gui/attributetable/qgsfeaturelistview.cpp
@@ -204,12 +204,12 @@ void QgsFeatureListView::setEditSelection( const QModelIndex &index, QItemSelect
 {
   bool ok = true;
   emit aboutToChangeEditSelection( ok );
-
+/*
 #ifdef QGISDEBUG
   if ( index.model() != mModel->masterModel() )
     qWarning() << "Index from wrong model passed in";
 #endif
-
+*/
   if ( ok )
     mCurrentEditSelectionModel->select( index, command );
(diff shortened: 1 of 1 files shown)


src/gui/attributetable/qgsfeaturelistview.cpp
```diff
@@ -204,14 +204,20 @@ void QgsFeatureListView::setEditSelection( const QModelIndex &index, QItemSelect
 {
   bool ok = true;
   emit aboutToChangeEditSelection( ok );
-
-#ifdef QGISDEBUG
-  if ( index.model() != mModel->masterModel() )
+  
+  if ( index.isValid() && index.model() != mModel->masterModel() )
```
m-kuhn: Is there a reason for the additional conditional code changes compared to just adding the `index.isValid()` condition inside warning?

For what I can see it doesn't make any difference in reality but makes the code harder to read because it looks like a deliberate choice.

nyalldawson: Also -- isn't this a bad bug, if something results in passing a index from the wrong model? Why is this happening in the first place?

m-kuhn: Yes, normally. That's why there is the warning, which therefore IMO should be the _only_ thing within this branch.  
There is one valid case though which is an invalid index, where model is `nullptr`.

m-kuhn (approved):
Perfect

PR #32241, opened 2019-10-14
[FEATURE] Random marker fill symbol layer type

nyalldawson:
This commit adds a new fill symbol layer type "Random marker fill". It
allows polygon features to be rendered using a marker symbol placed
at random locations within the polygon boundary.

Options include:
- number of marker symbols to render for the polygon
- whether markers rendered near the edges of polygons should be clipped
to the polygon boundary or not
- an optional random number seed, to give consistent placement
of markers whenever maps are refreshed (also allows random placement
to play nice with QGIS server and tile-based rendering)

Sponsored by the SLYR project

image


nyalldawson:
Also should note: this is faaaaaast. There's a method floating around for achieving this via geometry generators, but it's very slow because it relies on multiple point-in-polygon intersection tests performed through the expression engine. This PR uses the native c++ random-points-in-polygon function which relies on tesselation of polygons and doesn't require any point in polygon tests.

That said, I think there's potentially further speed boosts possible by testing whether the area of a polygon is similar to the area of its bounding box. If so, we could potentially fall back on the random-points-in-extent brute force method which may still end up faster for these special cases.

nyalldawson:
(ideas for other options welcome!)

nirvn:
@nyalldawson , is the point count meant to be point count per polygon part for multipart polygons?

nyalldawson:
@nirvn it's currently count per part -- I need to add some extra hooks to the symbol rendering API in order to get count per feature working

nirvn:
@nyalldawson , K; the only option I can think of here would be to have the possibility to declare the nb of markers as a density value vs. absolute value. It can already be done via data-defined properties, but it might be nice to have it in the UI itself (plus, it'd take into account multpart area, etc.)

That said, Nyall isn't lying, this is blazing fast, and very cool :)

andreasneumann:
Very cool - this is on my wishlist!

One thing I would like to see in this context is the ability to support different symbols from a set of given symbols.

Applications would be a mixed forest with two different symbols for coniferous trees and deciduous trees.

Or an even more complex representation of scree or block areas where there would be a set of 5-10 different symbols where both the choice of symbols within a given set, as well as size and rotation of the symbols could be randomized.

Bernhard Jenny has written a paper on such scree representation and how they could be best represented. See https://www.researchgate.net/publication/239752664_Automatic_Scree_Representation_for_Topographic_Maps

One idea he had was deriving the dot density from a relief representation (terrain model).

Looking forward to see that supported in an upcoming QGIS version.

timlinux:
Lovely stuff!

SrNetoChan:
Great improvement! Thanks SLYR :)

As for extra settings, how about the minimum distance between random points? It could help guarantee that any two points don't get too close or overlap each other.

Another interesting option would be to set a minimum distance to the boundary of the polygon, to avoid having to click symbols, when you don't want them ouside the polygons, which in some cases don't look very good.

nyalldawson:
@SrNetoChan 

> As for extra settings, how about the minimum distance between random points? It could help guarantee that any two points don't get too close or overlap each other.

I've been thinking about this, but I think in many ways this could be better handled by adding a "random offset" check box to the existing marker fill symbol layer. This would result in semi-random but evenly distributed markers, which I think is what you're after here.

> Another interesting option would be to set a minimum distance to the boundary of the polygon, 

Let me think on this one :)

nyalldawson:
@andreasneumann 
> One thing I would like to see in this context is the ability to support different symbols from a set of given symbols.
> Applications would be a mixed forest with two different symbols for coniferous trees and deciduous trees.
> Or an even more complex representation of scree or block areas where there would be a set of 5-10 different symbols where both the choice of symbols within a given set, as well as size and rotation of the symbols could be randomized.

This should already be possible, since this symbol layer doesn't behave the same as the marker fill symbol layer and doesn't force rasterize the markers into an image. This means that each marker is drawn individually, so data defined properties are evaluated per marker. You could have a svg marker with data-defined path, using a random number to pick the rendered svg file.

That said -- it's not a particular user friendly approach. Possibly what we could do here is have a new marker symbol type for "random marker" . This symbol layer type would allow multiple marker sub symbols, and would pick at render time one of these sub symbols to use. 

> One idea he had was deriving the dot density from a relief representation (terrain model).

Should be possible using this PR. You'd need to do a bit of data prep in advance, basically something like taking the dem, reclassifying it, vectorizing the result, and rendering random markers in this vector.

andreasneumann:
> This should already be possible, since this symbol layer doesn't behave the same as the marker fill symbol layer and doesn't force rasterize the markers into an image. This means that each marker is drawn individually, so data defined properties are evaluated per marker. You could have a svg marker with data-defined path, using a random number to pick the rendered svg file.
> 
> That said -- it's not a particular user friendly approach. Possibly what we could do here is have a new marker symbol type for "random marker" . This symbol layer type would allow multiple marker sub symbols, and would pick at render time one of these sub symbols to use.

I was thinking that it might already be possible - but wasn't sure.

One general question that I have about this renderer is: are the seed values stored somewhere? It would be quite disturbing if with each rendering we would get a completely different seed value and thus different positions of the symbols. This would make it unusable for tile generation and other applications where it would look quite bad if the map changes with each rendering.

I think we talked about this before and you said that we might be able to generate some seed value from the geometry itself - or I might be wrong.

andreasneumann:
> > One idea he had was deriving the dot density from a relief representation (terrain model).
> 
> Should be possible using this PR. You'd need to do a bit of data prep in advance, basically something like taking the dem, reclassifying it, vectorizing the result, and rendering random markers in this vector.

NIce! Yes - preprocessing would be fine in this case.

nyalldawson:
@andreasneumann 

> One general question that I have about this renderer is: are the seed values stored somewhere?

They are just another setting for individual symbol layers -- you can either leave it as randomised (different placement on each render), or put some seed value in to get the same placement with each render. (By default, a newly created random marker layer will use a fixed seed value (which is itself randomly generated!))

nyalldawson:
@nirvn this is ready for review/merge now

src/core/symbology/qgsfillsymbollayer.cpp
```diff
@@ -3909,3 +3912,306 @@ void QgsRasterFillSymbolLayer::applyPattern( QBrush &brush, const QString &image
...
+QgsRandomMarkerFillSymbolLayer::QgsRandomMarkerFillSymbolLayer( int pointCount, unsigned long seed )
+  : mPointCount( pointCount )
+  , mSeed( seed )
+{
+  setSubSymbol( new QgsMarkerSymbol() );
+}
+
+QgsSymbolLayer *QgsRandomMarkerFillSymbolLayer::create( const QgsStringMap &properties )
+{
+  const int pointCount = properties.value( QStringLiteral( "point_count" ), QStringLiteral( "1" ) ).toInt();
```
nirvn: Maybe we should have a better default here, like 10?

src/core/symbology/qgsfillsymbollayer.cpp
```diff
@@ -3909,3 +3912,306 @@ void QgsRasterFillSymbolLayer::applyPattern( QBrush &brush, const QString &image
...
+
+  unsigned long seed = 0;
+  if ( properties.contains( QStringLiteral( "seed" ) ) )
+    seed = properties.value( QStringLiteral( "seed" ), QStringLiteral( "0" ) ).toULong();
```
nirvn: If the computation of a random seed number (currently in the else clause) isn't too expensive, what about always coming up which such a number and use that instead of "0" as fallback value here?

src/core/symbology/qgsfillsymbollayer.cpp
```diff
@@ -3909,3 +3912,306 @@ void QgsRasterFillSymbolLayer::applyPattern( QBrush &brush, const QString &image
...
+  QgsStringMap map;
+  map.insert( QStringLiteral( "point_count" ), QString::number( mPointCount ) );
+  map.insert( QStringLiteral( "seed" ), QString::number( mSeed ) );
+  map.insert( QStringLiteral( "clip_points" ), QString::number( mClipPoints ) );
+  return map;
+}
+
+QgsRandomMarkerFillSymbolLayer *QgsRandomMarkerFillSymbolLayer::clone() const
+{
+  std::unique_ptr< QgsRandomMarkerFillSymbolLayer > x = qgis::make_unique< QgsRandomMarkerFillSymbolLayer >( mPointCount, mSeed );
```
nirvn: What's with the `x`? You're usually the one calling for more explicit variable name, even in short context spam ;-P

src/core/symbology/qgsfillsymbollayer.h
```diff
@@ -1702,6 +1722,126 @@ class CORE_EXPORT QgsPointPatternFillSymbolLayer: public QgsImageFillSymbolLayer
...
+ */
+class CORE_EXPORT QgsRandomMarkerFillSymbolLayer : public QgsFillSymbolLayer
+{
+  public:
+
+    /**
+     * Constructor for QgsRandomMarkerFillSymbolLayer, with the specified \a pointCount.
+     *
+     * Optionally a specific random number \a seed can be used when generating points. A \a seed of 0 indicates that
+     * a truly random sequence should be used.
```
nirvn: Maybe it's worth being a bit more explicit here as to what 0 does. Something like: a seed of 0 indicates that a truly random sequence will be used on every rendering?

src/core/symbology/qgsfillsymbollayer.cpp
nyalldawson: Actually this default isn't being used -- the .contains() check above negates it

nirvn: @nyalldawson , ah true, an dif the seed value isn't a number (i.e. somebody setting a wrong property value in the map), it wouldn't fallback to the default anyways. All good, sorry for the noise ;-P

nyalldawson:
@nirvn thanks, all addressed!

nirvn:
@nyalldawson , thanks. +1 to merge as soon as travis turns green.

gunerphp:
> @SrNetoChan
> 
> > As for extra settings, how about the minimum distance between random points? It could help guarantee that any two points don't get too close or overlap each other.
> 
> I've been thinking about this, but I think in many ways this could be better handled by adding a "random offset" check box to the existing marker fill symbol layer. This would result in semi-random but evenly distributed markers, which I think is what you're after here.


Hi @nyalldawson, are you still working on this feature? I really need that feature "semi-random fill with minimum offset from each marker

nyalldawson:
@gunerphp Not actively -- email me on nyall.dawson@gmail.com if you'd like to discuss sponsoring this development!

PR #32252, opened 2019-10-15
[oracle] fixes insertion of dates

mhugo:
## Description

Fixes the insertion of dates in the Oracle provider. cf https://github.com/qgis/QGIS/issues/27087

I've added a unit test for that particular regression. I've also had to disable tests for transactions that do not pass (anymore ?).
It has been tested on a 18.4.0 database (via a Docker instance provided by Oracle)

Since the bug was originally tagged as High Priority and Regression, I would say it qualifies for breaking the hard freeze, but I am not sure ...

## Checklist


nyalldawson:
>  I've also had to disable tests for transactions that do not pass (anymore ?).

Please don't -- these tests DO pass. (Try commenting out other parts of the test suite and you'll see). It's just that the provider itself hits the deadlock after a certain number of connections are made. I suspect this is due to the not-deleting-connections issue previously discussed.

tests/src/python/test_provider_oracle.py
```diff
@@ -168,6 +168,20 @@ def testAddFeatureWrongGeomType(self):
         """
         pass
 
+    def testDeleteFeaturesUpdateExtent(self):
+        # FIXME
+        """
+        This test does not pass for the Oracle provider.
```
nyalldawson: These ones pass for me -- but again, I suspect the deeper underlying issue is causing instability here.

mhugo:
@nyalldawson Indeed transaction tests seem to pass from a freshly restarted server ..., I'll enable them back.

About the extent tests (after insertion or deletion), I have other problems however. I've isolated one test and ran it on a newly started server. Results are kind of strange.

Results from ctest:
```
513: ======================================================================
513: FAIL: testDeleteFeaturesUpdateExtent (__main__.TestPyQgsOracleProvider)
513: ----------------------------------------------------------------------
513: Traceback (most recent call last):
513:   File "/home/hme/src/QGIS/tests/src/python/test_provider_oracle.py", line 180, in testDeleteFeaturesUpdateExtent
513:     self.assertEqual(l.dataProvider().extent().toString(1), '-71.1,66.3 : -65.3,78.3')
513: AssertionError: '-80.1,66.3 : -53.7,78.3' != '-71.1,66.3 : -65.3,78.3'
513: - -80.1,66.3 : -53.7,78.3
513: ?  ^^             --
513: + -71.1,66.3 : -65.3,78.3
513: ?  ^^           + +
```

Which seems to mean the extent is incorrect. Then manually testing with sqlplus:

```
SQL> select sdo_tune.extent_of('qgis.edit_data', 'geom') from dual;

SDO_TUNE.EXTENT_OF('QGIS.EDIT_DATA','GEOM')(SDO_GTYPE, SDO_SRID, SDO_POINT(X, Y,
--------------------------------------------------------------------------------
SDO_GEOMETRY(2003, 4326, NULL, SDO_ELEM_INFO_ARRAY(1, 1003, 3), SDO_ORDINATE_ARR
AY(-80.09754, 66.33, -53.747097, 78.3))


SQL> select sdo_aggr_mbr(geom) from qgis.edit_data;

SDO_AGGR_MBR(GEOM)(SDO_GTYPE, SDO_SRID, SDO_POINT(X, Y, Z), SDO_ELEM_INFO, SDO_O
--------------------------------------------------------------------------------
SDO_GEOMETRY(2003, 4326, NULL, SDO_ELEM_INFO_ARRAY(1, 1003, 3), SDO_ORDINATE_ARR
AY(-71.123, 66.33, -65.32, 78.3))
```

So the answer given by SDO_TUNE.EXTENT_OF is wrong and is different from the one given by SDO_ADDR_MBR. I'll investigate a bit further ...

Jean-Roc:
@mhugo did you test against a previous version of oracle such as 11g or 12c ?

mhugo:
The problem seems to come from the use of SDO_TUNE.EXTENT_OF on a table that has NULL geometry values. In that case, the returned MBR is wrong.

When trying to know if it is a known issue and if it has been fixed in a newer version (I am testing on 18.4.0), I was looking for a place on the Internet where issues are declared, but I am apparently too naive in assuming that place would be free and public.

mhugo:
@Jean-Roc Existing unit tests do not work with 11g. The last time I tried SDO_TUNE.EXTENT_OF on 11g resulted in a query that took several **minutes** (yes minutes) to execute on a 5-row table.

Jean-Roc:
@mhugo please see with @troopa81, this bug is part of the recent contracting we did with oslandia and these oracle versions are those we use in production (and unfortunately this query delay doesn't surprise me).

mhugo:
@Jean-Roc I've just tested with a 11g. The bogus result is still there with sdo_tune.extent_of and null geometries. Moreover, the query takes ages to execute.

mhugo:
Back to the failing test, reading the [provider code](https://github.com/qgis/QGIS/blob/master/src/providers/oracle/qgsoracleprovider.cpp#L2293), sdo_tune.extent_of is used to compute the extent, whereas there is no guarantee this function returns the actual extent. I would say sdo_aggr_mbr should be used instead when the layer's URI contains "estimatedmetadata=false".

Jean-Roc:
@mhugo, to be clear, I was referring to the date fixe not sdo_tune.extent_of :)

mhugo:
@Jean-Roc Sorry, I was lost in my bug. Yes, the fix works with 11g

mhugo:
New issue related to the use of extent_of: https://github.com/qgis/QGIS/issues/32261

I've enabled back tests on transactions, but kept extent tests as disabled, waiting for the above-mentioned issue to be solved.

mhugo:
@nyalldawson Do you think it is ok to merge ? For 3.10.0 or the next one ?

nyalldawson:
@mhugo I honestly don't know -- the fix looks good to me, but I'm really unclear on the hard freeze procedure.

mhugo:
@nyall Not clear for me either :-/

After some more thoughts, I think merging now is ok, considering that:
- it is a fix for a regression from 2.x
- the change is very limited in scope (fixing https://github.com/qgis/QGIS/issues/32261 on the contrary might have a broader impact and would be fixed after the hard freeze)
- I'd like to see it in 3.4 and it seems to me kind of strange to have to wait for 3.4.14 to be in sync with 3.10.1

mhugo:
Anyone for an approval ? Or any objection ?

rldhont (approved):

rldhont:
Thnaks @mhugo.

If you want to add it to release-3_4 you have to manually backport your changes.

mhugo:
@rldhont So, you approved the fix, but you're against a merge in 3.10.0, right ?

rldhont:
@mhugo I am not against a merge in 3.10.0, I just understood that only critical bug fixes (e.g. crash or data corruption) could be merged.

So If you don't want to wait for 3.4.14, you have to manually backport.

mhugo:
@rldhont Correct. Thanks. I've created a PR for 3.4 [here](https://github.com/qgis/QGIS/pull/32329)

nyalldawson:
@mhugo Hmmmm, Travis refuses to acknowledge that this pr exists.... Can you open a new one?

PR #32258, opened 2019-10-15
[sipify] correctly output deprecation

3nids:


m-kuhn (approved):
Does that also apply to versionadded:: etc?

PR #32262, opened 2019-10-15
[FEATURE] Add OGC API - Features (OAPIF) provider

rouault:
Implements what was mentionned in https://lists.osgeo.org/pipermail/qgis-developer/2019-September/058649.html

Contents of the pull request:
- Logic from the WFS provider to deal with background downloading of features and caching them in a temporary Spatialie DB has been separated from WFS-specific code to be re-usable.
- The OAPIF provider (non-GUI part) is added as a standalone provider (in the same .so/.dll as the WFS one), which uses the above common code (the gist of the provider is in the commit "[Feature] Add 'OGC API - Features' provider, shortnamed as OAPIF"). The provider expects the server to implement the final specification of http://docs.opengeospatial.org/is/17-069r3/17-069r3.html (some compatibility hacks exist though to handle a few servers still on recent drafts)
- For the GUI, the OAPIF provider uses the same one as the WFS one (browser and "source select"). It is mostly transparent to users. They provide an URL (for OAPIF, must be the one of the "landing page") and the GUI part figures out if its is WFS or OAPIF automatically. Users can also override this in case the auto-detection would fail, or if an endpoint would implement both protocols, in which case WFS is used by default, as it is more capable.
- The OAPIF provider handles layer metadata, server-side spatial and temporal filtering.

The commit history should be hopefully clean and proceeds steps by steps.

The OAPIF provider can be tested against following endpoints (at the time of writing...)
- https://api.planet.com/analytics/ (requires an API key)
- https://stac.boundlessgeo.io/
- https://beta-paikkatieto.maanmittauslaitos.fi/maastotiedot/wfs3/v1
- http://www.pvretano.com/cubewerx/cubeserv/default/wfs/3.0.0/framework (partially as the server doesn't implement paging yet)
- https://services.interactive-instruments.de/t15/daraa
- https://www.ldproxy.nrw.de/kataster
- https://demo.pygeoapi.io/master
- Local instance of QGIS server 3.10.0

Funded by Planet


elpaso:
Did you test it with QGIS Server too? I'd like to know if is there anything I've missed, I should probably update the implementation to the final specification.

rouault:
> Did you test it with QGIS Server too?

No, is there an easy way to do so (like a service available online) ? Never played locally with QGIS server

elpaso:
I use this script locally to run it for interactive tests https://termbin.com/hmqy 
you should probably change some env vars/paths to adapt it to your installation.

rouault:
> I use this script locally to run it for interactive tests https://termbin.com/hmqy

Thanks for the hint. I did manage to get qgis_server working. There are non-conformities of it to the final OAPIF spec ("name" instead of "id" in collections description, extent/spatial should be extent/spatial/bbox and bbox an array of bbox). But the show stopper with the above commits what in the landing page with the rel=service_desc for the /api . It should be service-desc (dash, and not underscore). With the commit e747124 on the client side, I managed to get the provider working. Another thing is that in the connection definition dialog, you must force OAPIF, otherwise it will autodetect WFS 1.1 (perhaps there should be something on the server side, or in the rewrite rules, so that an explicit /wfs3/ endpoint only has OAPIF capabilities ?)

elpaso:
> > I use this script locally to run it for interactive tests https://termbin.com/hmqy
> 
> Thanks for the hint. I did manage to get qgis_server working. There are non-conformities of it to the final OAPIF spec ("name" instead of "id" in collections description, extent/spatial should be extent/spatial/bbox and bbox an array of bbox). But the show stopper with the above commits what in the landing page with the rel=service_desc for the /api . It should be service-desc (dash, and not underscore). With the commit [e747124](https://github.com/qgis/QGIS/commit/e7471249dddfbccb2ea82760cd5971c0df804f89) on the client side, I managed to get the provider working. 

Thank you for the feed-back, I'll fix the issues.

> Another thing is that in the connection definition dialog, you must force OAPIF, otherwise it will autodetect WFS 1.1 (perhaps there should be something on the server side, or in the rewrite rules, so that an explicit /wfs3/ endpoint only has OAPIF capabilities ?)

The original idea was to maintain compatibility with the "old" services and the current logic is that we first check for `SERVICE` in the query string, and only if it is not found then we start to search for matching APIs (i.e. '/wfs3').

I think we will eventually move the legacy services to the the new API so that in the future we will have:

/wms -> WMS
/wfs -> WFS
/wcs -> WCS 
...
/wfs3 -> OAPIF

But we could also revert the logic right now and check for matches with the APIs first and fall back to the legacy `SERVICE`s code if no matches were found.

@pblottiere , @rldhont what's your feeling about this?

elpaso:
See https://github.com/qgis/QGIS/pull/32271 for the `rel_*`

elpaso:
@rouault for  "name" -> "id", do you mean the change must be in:

https://github.com/qgis/QGIS/blob/master/tests/testdata/qgis_server/api/test_wfs3_api_project.json#L182

https://github.com/qgis/QGIS/blob/master/tests/testdata/qgis_server/api/test_wfs3_collections_project.json#L37
?

elpaso:
> There are non-conformities of it to the final OAPIF spec (extent/spatial should be extent/spatial/bbox and bbox an array of bbox). 

Are you sure? From the specs:

```
While the spatial and temporal extents support multiple bounding boxes (bbox array) and time intervals (interval array) for advanced use cases, implementations SHOULD provide only a single bounding box or time interval unless the use of multiple values is important for the use of the dataset and agents using the API are known to be support multiple bounding boxes or time intervals.
```

rouault:
>  for "name" -> "id", do you mean the change must be in:

yes. See "Schema for a feature collection" in http://docs.opengeospatial.org/is/17-069r3/17-069r3.html

> Are you sure?

yes, see "Example 4. Feature collections response document". There is provision to support multiple bbox, even if a single one is strongly recommended
```
"extent": {
        "spatial": {
          "bbox": [ [ 7.01, 50.63, 7.22, 50.78 ] ]
        },
```

src/providers/wfs/qgsbackgroundcachedfeatureiterator.h
```diff
@@ -0,0 +1,352 @@
...
+
+  private:
+    QgsBackgroundCachedSharedData *mSharedBase;
+    QgsFeatureDownloader *mDownloader;
+    QWidget *mMainWindow = nullptr;
+    QMutex mMutexCreateProgressDialog;
+};
+
+// Sorry for ugliness. Due to QgsFeatureDownloaderImpl that cannot derived from QObject
+#define CONNECT_PROGRESS_DIALOG(actual_downloader_impl_class) do { \
```
elpaso: Is there any alternative to those macros? Like templates?

src/providers/wfs/qgsoapifcollection.cpp
```diff
@@ -0,0 +1,497 @@
...
+  if ( !id.is_string() )
+    return false;
+  mId = QString::fromStdString( id.get<std::string>() );
+
+  mLayerMetadata.setType( QStringLiteral( "dataset" ) );
+
+  const auto links = QgsOAPIFJson::parseLinks( j );
+  const auto selfUrl = QgsOAPIFJson::findLink( links,
+                       QStringLiteral( "self" ),
+                       QStringList() <<  QStringLiteral( "application/json" ) );
```
elpaso: nothing really important, but you could also do
```suggestion
                       { QStringLiteral( "application/json" ) } );
```

src/providers/wfs/qgsoapifcollection.cpp
```diff
@@ -0,0 +1,497 @@
...
+          }
+          metadataExtent.setTemporalExtents( temporalExtents );
+        }
+      }
+    }
+
+    mLayerMetadata.setExtent( metadataExtent );
+  }
+
+  // From STAC specification
```
elpaso: what is it?

src/providers/wfs/qgsoapifcollection.cpp
```diff
@@ -0,0 +1,497 @@
...
+        if ( licenseSet.find( license ) == licenseSet.end() )
+        {
+          licenseSet.insert( license );
+          licenses << license;
+        }
+      }
+    }
+    if ( licenses.isEmpty() && isProprietaryLicense )
+    {
+      licenses << "proprietary";
```
elpaso:
```suggestion
      licenses << QStringLiteral( "proprietary" );
```

src/providers/wfs/qgsoapifcollection.cpp
```diff
@@ -0,0 +1,497 @@
...
+              // use the one from the collection set.
+              collection.mLayerMetadata.setLicenses( licenses );
+            }
+            mCollections.emplace_back( collection );
+          }
+        }
+      }
+    }
+
+    // Paging informal extension used by api.planet.com/
```
elpaso: I thought that next/prev it were "permitted" by the specs. Is this an extension?

src/providers/wfs/qgsoapifcollection.cpp
```diff
@@ -0,0 +1,497 @@
...
+    mNextUrl = QgsOAPIFJson::findLink( links,
+                                       QStringLiteral( "next" ),
+                                       QStringList() <<  QStringLiteral( "application/json" ) );
+  }
+  catch ( const json::parse_error &ex )
+  {
+    mErrorCode = QgsBaseNetworkRequest::ApplicationLevelError;
+    mAppLevelError = ApplicationLevelError::JsonError;
+    mErrorMessage = errorMessageWithReason( tr( "Cannot decode JSon document: %1" ).arg( QString::fromStdString( ex.what() ) ) );
```
elpaso:
```suggestion
    mErrorMessage = errorMessageWithReason( tr( "Cannot decode JSON document: %1" ).arg( QString::fromStdString( ex.what() ) ) );
```

src/providers/wfs/qgsoapifcollection.cpp
```diff
@@ -0,0 +1,497 @@
...
+  try
+  {
+    const json j = json::parse( utf8Text.toStdString() );
+    mCollection.deserialize( j );
+  }
+  catch ( const json::parse_error &ex )
+  {
+    mErrorCode = QgsBaseNetworkRequest::ApplicationLevelError;
+    mAppLevelError = ApplicationLevelError::JsonError;
+    mErrorMessage = errorMessageWithReason( tr( "Cannot decode JSon document: %1" ).arg( QString::fromStdString( ex.what() ) ) );
```
elpaso:
```suggestion
    mErrorMessage = errorMessageWithReason( tr( "Cannot decode JSON document: %1" ).arg( QString::fromStdString( ex.what() ) ) );
```

src/providers/wfs/qgsoapifcollection.h
```diff
@@ -0,0 +1,143 @@
...
+  //! Description
+  QString mDescription;
+
+  //! Bounding box (in CRS84)
+  QgsRectangle mBbox;
+
+  //! Layer metadata
+  QgsLayerMetadata mLayerMetadata;
+
+  //! Fills a collection from its JSon serialization
```
elpaso:
```suggestion
  //! Fills a collection from its JSON serialization
```

src/providers/wfs/qgsoapifprovider.cpp
```diff
@@ -0,0 +1,789 @@
...
+#if 0
+  request.setLimit( 1 );
+#endif
+  return !getFeatures( request ).nextFeature( f );
+
+};
+
+bool QgsOapifProvider::setSubsetString( const QString &sql, bool updateFeatureCount )
+{
+  QgsDebugMsgLevel( QStringLiteral( "sql = '%1'" ).arg( sql ), 4 );
```
elpaso: this isn't really SQL, is it?

elpaso (approved):
Great work! Thank you.

I admit that I only had a (relatively) quick look at the code but it looks good to me.

I didn't really get if any of the URLs are constructed by appending `/collections` or `/items` to the path ony. 
Just make sure that any extra query string argument is passed through, in case a client wanted to use a specific token exposed in the query string (I know it's not best practice, but still).

rouault:
> I didn't really get if any of the URLs are constructed by appending `/collections` or `/items` to the path ony.

The collections link is retrieved from the landing page (but that could be hardcoded as it is mandated by the spec). The /items is appended to the path of the collection

> Just make sure that any extra query string argument is passed through, in case a client wanted to use a specific token exposed in the query string (I know it's not best practice, but still).

No extra query string arguments are not passed through currently. It is not clear if that would work properly with all the possible paths.

src/core/qgscoordinatereferencesystem.cpp
```diff
@@ -392,7 +392,7 @@ bool QgsCoordinateReferenceSystem::createFromOgcWmsCrs( const QString &crs )
 
   QString wmsCrs = crs;
 
-  QRegExp re_uri( "http://www\\.opengis\\.net/def/crs/([^/]+).+/(\\d+)", Qt::CaseInsensitive );
+  QRegExp re_uri( "http://www\\.opengis\\.net/def/crs/([^/]+).+/([^/]+)", Qt::CaseInsensitive );
```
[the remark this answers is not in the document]
rouault: yes, see commit bc512570a74aed69af4f3aa7d7b891b160c43b33

src/providers/wfs/qgsbackgroundcachedfeatureiterator.h
rouault: I considered that, but I'm not sure if that would work here. Templates are a pain for me :-)

src/providers/wfs/qgsoapifcollection.cpp
rouault: A close friend of OAPIF : https://stacspec.org/

src/providers/wfs/qgsoapifcollection.cpp
rouault: It is not clear to me if paging of collections themselves is somewhat really foreseen by the spec. It is in the "spirit" of it, but never mentioned, contrary to paging of features

src/providers/wfs/qgsoapifprovider.cpp
```diff
@@ -0,0 +1,789 @@
...
+
+  mShared->mCapabilityExtent = collectionRequest.collection().mBbox;
+
+  mLayerMetadata = collectionRequest.collection().mLayerMetadata;
+  // Merge contact info from /api
+  mLayerMetadata.setContacts( apiRequest.metadata().contacts() );
+
+  mShared->mItemsUrl = mShared->mCollectionUrl +  QStringLiteral( "/items" );
+
+  QgsOapifItemsRequest itemsRequest( mShared->mURI.uri(), mShared->mItemsUrl + QStringLiteral( "?limit=10" ) );
```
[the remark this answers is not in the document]
rouault: Looks to me like it should be supported by a server: http://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_parameter_limit

src/providers/wfs/qgsoapifcollection.cpp
rouault: URL added to comment

src/providers/wfs/qgsoapifcollection.cpp
rouault: taken into account

src/providers/wfs/qgsoapifcollection.cpp
rouault: taken into account

src/providers/wfs/qgsoapifcollection.h
rouault: taken into account

src/providers/wfs/qgsoapifitemsrequest.cpp
```diff
@@ -0,0 +1,175 @@
...
+      {
+        mNumberMatched = numberMatched.get<int>();
+      }
+    }
+  }
+  catch ( const json::parse_error &ex )
+  {
+    mErrorCode = QgsBaseNetworkRequest::ApplicationLevelError;
+    mAppLevelError = ApplicationLevelError::JsonError;
+    mErrorMessage = errorMessageWithReason( tr( "Cannot decode JSon document: %1" ).arg( QString::fromStdString( ex.what() ) ) );
```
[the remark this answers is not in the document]
rouault: taken into account

src/providers/wfs/qgsoapiflandingpagerequest.cpp
```diff
@@ -0,0 +1,154 @@
...
+                        QStringLiteral( "collections" ),
+                        apiTypes );
+    }
+#endif
+  }
+  catch ( const json::parse_error &ex )
+  {
+    mErrorCode = QgsBaseNetworkRequest::ApplicationLevelError;
+    mAppLevelError = ApplicationLevelError::JsonError;
+    mErrorMessage = errorMessageWithReason( tr( "Cannot decode JSon document: %1" ).arg( QString::fromStdString( ex.what() ) ) );
```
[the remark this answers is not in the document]
rouault: taken into account

src/providers/wfs/qgsoapifprovider.cpp
rouault: right. renaming variable and debug message to filter

src/providers/wfs/qgswfsprovider.cpp
```diff
@@ -118,11 +121,12 @@ QgsWFSProvider::QgsWFSProvider( const QString &uri, const ProviderOptions &optio
   //Failed to detect feature type from describeFeatureType -> get first feature from layer to detect type
   if ( mShared->mWKBType == QgsWkbTypes::Unknown )
   {
-    QgsWFSFeatureDownloader downloader( mShared.get() );
-    connect( &downloader, static_cast<void ( QgsWFSFeatureDownloader::* )( QVector<QgsWFSFeatureGmlIdPair> )>( &QgsWFSFeatureDownloader::featureReceived ),
+    auto downloader = qgis::make_unique<QgsFeatureDownloader>();
+    downloader->setImpl( qgis::make_unique<QgsWFSFeatureDownloaderImpl>( mShared.get(), downloader.get() ) );
+    connect( downloader.get(), static_cast<void ( QgsFeatureDownloader::* )( QVector<QgsFeatureUniqueIdPair> )>( &QgsFeatureDownloader::featureReceived ),
```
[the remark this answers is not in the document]
rouault: taken into account

src/providers/wfs/qgsoapifcollection.cpp
rouault: taken into account

src/providers/wfs/qgsoapifcollection.cpp
rouault: Taken into account

src/app/qgisapp.cpp
```diff
@@ -11395,7 +11395,7 @@ QgsVectorLayer *QgisApp::addVectorLayerPrivate( const QString &vectorLayerPath,
   {
     if ( guiWarning )
     {
-      QString message = layer->dataProvider()->error().message( QgsErrorMessage::Text );
+      QString message = layer->dataProvider() ? layer->dataProvider()->error().message( QgsErrorMessage::Text ) : tr( "Invalid provider" );
```
nyalldawson: This looks like a bug fix which should be applied to 3.10 also?

src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
```diff
@@ -0,0 +1,870 @@
+/***************************************************************************
+    qgsbacckgroundcachedfeatureiterator.h
```
nyalldawson:
```suggestion
    qgsbackgroundcachedfeatureiterator.h
```

src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
```diff
@@ -0,0 +1,870 @@
...
+
+void QgsFeatureDownloaderImpl::hideProgressDialog()
+{
+  mSharedBase->setHideProgressDialog( true );
+  mProgressDialog->deleteLater();
+  mProgressDialog = nullptr;
+}
+
+// Called from GUI thread
+void QgsFeatureDownloaderImpl::createProgressDialog( int numberMatched )
```
nyalldawson: I **really* don't like this gui specific logic sitting in here, it's just completely out of place. I would much prefer to see this replaced by a QgsTask.

src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
```diff
@@ -0,0 +1,870 @@
...
+      int idx = cachedFeature.fields().indexFromName( QgsBackgroundCachedFeatureIteratorConstants::FIELD_HEXWKB_GEOM );
+      Q_ASSERT( idx >= 0 );
+
+      const QVariant &v = cachedFeature.attributes().value( idx );
+      if ( !v.isNull() && v.type() == QVariant::String )
+      {
+        QByteArray wkbGeom( QByteArray::fromHex( v.toString().toLatin1() ) );
+        QgsGeometry g;
+        unsigned char *wkbClone = new unsigned char[wkbGeom.size()];
+        memcpy( wkbClone, wkbGeom.data(), wkbGeom.size() );
```
nyalldawson: I don't think this copy is required -- there's a QgsGeometry::fromWkb which takes a QByteArray value already (and the char* version doesn't take ownership in any case)

src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
```diff
@@ -0,0 +1,870 @@
...
+  {
+    // Initialize a reader stream if there's a writer stream available
+    if ( !mReaderStream )
+    {
+      {
+        QMutexLocker locker( &mMutex );
+        if ( mWriterStream )
+        {
+          // Transfer writer variables to the reader
+          delete mWriterStream;
```
nyalldawson: unique_ptr candidates?

src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
```diff
@@ -0,0 +1,870 @@
...
+      }
+      // Instantiates the reader stream from memory buffer if not empty
+      if ( !mReaderByteArray.isEmpty() )
+      {
+        mReaderStream = new QDataStream( &mReaderByteArray, QIODevice::ReadOnly );
+      }
+      // Otherwise from the on-disk file
+      else if ( !mReaderFilename.isEmpty() )
+      {
+        mReaderFile = new QFile( mReaderFilename );
```
nyalldawson: unique_ptr candidate?

src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
```diff
@@ -0,0 +1,870 @@
...
+        if ( !mReaderFile->open( QIODevice::ReadOnly ) )
+        {
+          QgsDebugMsg( QStringLiteral( "Cannot open %1" ).arg( mReaderFilename ) );
+          delete mReaderFile;
+          mReaderFile = nullptr;
+          return false;
+        }
+        mReaderStream = new QDataStream( mReaderFile );
```
nyalldawson: also potentially unique_ptr

src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
```diff
@@ -0,0 +1,870 @@
...
+
+
+void QgsBackgroundCachedFeatureIterator::copyFeature( const QgsFeature &srcFeature, QgsFeature &dstFeature, bool srcIsCache )
+{
+  //copy the geometry
+  QgsGeometry geometry = srcFeature.geometry();
+  if ( mShared->hasGeometry() && !geometry.isNull() )
+  {
+    QgsGeometry g;
+    g.fromWkb( geometry.asWkb() );
```
nyalldawson: why not just set directly? You'll get the benefits of the implicit sharing and a near-zero cost copy...

src/providers/wfs/qgsbackgroundcachedshareddata.cpp
```diff
@@ -0,0 +1,1281 @@
...
+void QgsBackgroundCachedSharedData::invalidateCache()
+{
+  // Cf explanations in registerToCache() for the locking strategy
+  QMutexLocker lockerMyself( &mMutexRegisterToCache );
+
+  QMutexLocker locker( &mMutex );
+
+// to prevent deadlock when waiting the end of the downloader thread that will try to take the mutex in serializeFeatures()
+  mMutex.unlock();
+  delete mDownloader;
```
nyalldawson: unique_ptr candidate?

src/providers/wfs/qgsbackgroundcachedshareddata.cpp
```diff
@@ -0,0 +1,1281 @@
...
+  mFeatureCountExact = false;
+  mFeatureCountRequestIssued = false;
+  mTotalFeaturesAttemptedToBeCached = 0;
+  if ( !mCacheDbname.isEmpty() && mCacheDataProvider )
+  {
+    // We need to invalidate connections pointing to the cache, so as to
+    // be able to delete the file.
+    mCacheDataProvider->invalidateConnections( mCacheDbname );
+  }
+  delete mCacheDataProvider;
```
nyalldawson: unique_ptr?

src/providers/wfs/qgsbackgroundcachedshareddata.cpp
```diff
@@ -0,0 +1,1281 @@
...
+
+  QgsFields dataProviderFields = mCacheDataProvider->fields();
+  const int uniqueIdIdx = dataProviderFields.indexFromName( QgsBackgroundCachedFeatureIteratorConstants::FIELD_UNIQUE_ID );
+
+  // To avoid excessive memory consumption in expression building, do not
+  // query more than 1000 ids at a time.
+  for ( int i = 0; i < featureList.size(); i ++ )
+  {
+    if ( !first )
+      expr += QLatin1String( "," );
```
nyalldawson:
```suggestion
      expr += ',';
```

nyalldawson: (elsewhere too)

src/providers/wfs/qgsbasenetworkrequest.cpp
```diff
@@ -0,0 +1,466 @@
...
+#include <QNetworkCacheMetaData>
+#include <QCryptographicHash> // just for testin file:// fake_qgis_http_endpoint hack
+#include <QFuture>
+#include <QtConcurrent>
+
+QgsBaseNetworkRequest::QgsBaseNetworkRequest( const QgsAuthorizationSettings &auth, const QString &translatedComponent )
+  : mAuth( auth )
+  , mTranslatedComponent( translatedComponent )
+  , mErrorCode( QgsBaseNetworkRequest::NoError )
+  , mIsAborted( false )
```
nyalldawson: Move initialization to header

src/providers/wfs/qgsbasenetworkrequest.cpp
```diff
@@ -0,0 +1,466 @@
...
+    {
+      QString args = modifiedUrlString.mid( posQuotationMark );
+      if ( modifiedUrlString.size() > 256 )
+      {
+        QgsDebugMsgLevel( QStringLiteral( "Args before MD5: %1" ).arg( args ), 4 );
+        args = QCryptographicHash::hash( args.toUtf8(), QCryptographicHash::Md5 ).toHex();
+      }
+      else
+      {
+        args.replace( QLatin1String( "?" ), QLatin1String( "_" ) );
```
nyalldawson: could all be replaced by character based methods for speed

src/providers/wfs/qgsoapifprovider.cpp
```diff
@@ -0,0 +1,789 @@
...
+
+  QString errorMsg;
+  if ( !mShared->computeServerFilter( errorMsg ) )
+  {
+    QgsMessageLog::logMessage( errorMsg, tr( "OAPIF" ) );
+    mValid = false;
+    return;
+  }
+
+  qRegisterMetaType<QgsRectangle>( "QgsRectangle" );
```
nyalldawson: Should not be required, or if it is, moved to QgsApplication alongside the other registrations

src/app/qgisapp.cpp
rouault: Maybe, but unlikely to happen in real world situations. I noticed this in an intermediate stage where I tried to add a layer for a provider that doesn't exist.

src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
rouault: yeah... This is not new code. Just the existing one of the WFS provider I moved a bit between files. I don't really feel like going through the QgsTask road in this pull request though

src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
rouault: taken into account

src/providers/wfs/qgsbackgroundcachedshareddata.cpp
rouault: taken into account

src/providers/wfs/qgsbasenetworkrequest.cpp
rouault: taken into account

src/providers/wfs/qgsbasenetworkrequest.cpp
rouault: taken into account

src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
rouault: indeed. fromWkb( const QByteArray &wkb ) was added in 3.0. This code predates it :-) . Taken into account.
(note you mention that the char* version doesn't take ownership, but the code says the contrary)

nyalldawson: Argh, you're right. Too early in the day for PR reviews :P

src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
rouault: taken into account

src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
rouault: taken into account

src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
rouault: taken into account

src/providers/wfs/qgsbackgroundcachedshareddata.cpp
rouault: taken into account

src/providers/wfs/qgsbackgroundcachedshareddata.cpp
rouault: taken into account

src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
rouault: taken into account. I was a bit worried about the thread-safety of the sharing (which could be an issue here), but looking at QgsGeometry code, it looks OK with respect to that. Taken into account

src/providers/wfs/qgsoapifprovider.cpp
rouault: was not needed indeed.

src/gui/qgsnewhttpconnection.h
```diff
@@ -103,6 +103,16 @@ class GUI_EXPORT QgsNewHttpConnection : public QDialog, private Ui::QgsNewHttpCo
...
   protected:
 
+    //! Index of wfsVersionComboBox
+    enum WfsVersionIndex
+    {
+      WFS_VERSION_MAX = 0,
+      WFS_VERSION_1_0 = 1,
+      WFS_VERSION_1_1 = 2,
+      WFS_VERSION_2_0 = 3,
+      WFS_VERSION_API_FEATURES = 4,
```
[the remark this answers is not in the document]
elpaso: This is what I mean, because http://docs.opengeospatial.org/is/17-069r3/17-069r3.html is Version 1.0
```suggestion
      WFS_VERSION_API_FEATURES_1_0 = 4,
```
if the situation with extensions becomes messy, we might need a capability flag later.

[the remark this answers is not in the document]
rouault: done

palmerj:
Is this going to be merged for 3.10?

rouault:
> Is this going to be merged for 3.10?

No, the development started even after the feature freeze.

palmerj:
Ok thanks. Pity it will miss the next LTR

PR #32269, opened 2019-10-16
Optimize wms raster information loading time and fix anchor hyperlink

### LINKED ISSUE #32213: Information tab in WMS layer properties is unresponsive if a WMS service contains many layers
**Issue**
When a WMS service contains many layers and the user opens the layer properties, QGIS becomes unresponsive for a while:

* Opening the layer properties takes quite a long time (often several seconds)
* Esp. opening the "Information" tab of the WMS layer takes very long (several seconds, up to a minute)

**How to Reproduce**

Here is a WMS service with many layers for testing:
https://geo.so.ch/api/wms

You can pick any layer, all have the same "unresponsiveness". e.g. you can pick "ch.so.arp.wanderwege"

**QGIS and OS versions**

Tested with Master, 3.4 and 3.8 on both Linux and Windows

elpaso:
This patch makes loading ~20 times faster for WMS with many layers.

Fixes #32213

Also fixes an (unreported?) issue with anchor hyperlinks ("Selected layers" and "Other layers")
Also fixes an (unreported?) issue with geotiff rasters that will try to show them in-browser (and fail)

If you are curious, the main speed gain comes from `QTextBrowser` -> `QgsWebView` widget change, I discovered that `QTextBrowser` is deadly slow when rendering large HTML: something to be aware of.

I've also changed a little the CSS style because it was really messy and unreadable with all those nested tables.


elpaso:
@andreasneumann I'd appreciate if you could give this a try.

src/app/qgsrasterlayerproperties.cpp
```diff
@@ -74,6 +75,7 @@
 #include <QMouseEvent>
 #include <QVector>
 #include <QUrl>
+#include <QWebFrame>
```
m-kuhn: I think this should be `#ifdef`'d as well

andreasneumann:
> @andreasneumann I'd appreciate if you could give this a try.

I tested it. It is really A LOT BETTER!

I also like the improved styling in the information tab and the quick link to the Selected vs Other Layers in the GetCapability view.

Thanks a lot for this improvement!

src/app/qgsrasterlayerproperties.cpp
elpaso: I probably don't need it anymore, it's a leftover from a previous attempt to adapt content size to parent (now done better with CSS).

src/app/qgsrasterlayerproperties.cpp
```diff
@@ -437,6 +437,22 @@ QgsRasterLayerProperties::QgsRasterLayerProperties( QgsMapLayer *lyr, QgsMapCanv
...
+  // Setup information tab
+  const int horizontalDpi = qApp->desktop()->screen()->logicalDpiX();
+  // Adjust zoom: text is ok, but HTML seems rather big at least on Linux/KDE
+  if ( horizontalDpi > 96 )
+  {
+    mMetadataViewer->setZoomFactor( mMetadataViewer->zoomFactor() * 0.9 );
+  }
+  mMetadataViewer->page()->setLinkDelegationPolicy( QWebPage::LinkDelegationPolicy::DelegateAllLinks );
+  connect( mMetadataViewer->page(), &QWebPage::linkClicked, this, &QgsRasterLayerProperties::urlClicked );
+  mMetadataViewer->page()->setViewportSize( QSize( 200, 200 ) );
```
nyalldawson: What's this for?

src/app/qgsrasterlayerproperties.cpp
```diff
@@ -2043,6 +2051,27 @@ bool QgsRasterLayerProperties::rasterIsMultiBandColor()
   return mRasterLayer && nullptr != dynamic_cast<QgsMultiBandColorRenderer *>( mRasterLayer->renderer() );
 }
 
+void QgsRasterLayerProperties::updateInformationContent()
+{
+  // We are using QgsWebView and the renderer is completely different from QTextBrowser
+  // add some extra style here
+  QString myStyle = QgsApplication::reportStyleSheet();
```
nyalldawson: What do you think about making adding an enum argument to reportStyleSheet to control whether QTextBrowser or QWebView stylesheet is returned?

src/app/qgsrasterlayerproperties.cpp
```diff
@@ -2043,6 +2051,27 @@ bool QgsRasterLayerProperties::rasterIsMultiBandColor()
...
+                                  "h1 { line-height: inherit; } "
+                                  "td, th { word-wrap: break-word; vertical-align: top; } "
+                                  ".list-view th:first-child, .list-view td:first-child { width: 15%; } "
+                                  ".list-view.highlight { padding-left: inherit; } "
+                                  ".tabular-view th:first-child, .tabular-view td:first-child { width: 20%; } "
+                                  ".tabular-view th.strong { background-color: #eee; }"
+                                  ".tabular-view th, .tabular-view td { "
+                                  "  border: solid 1px #eee;"
+                                  "} \n " ) );
+  const QString html { mRasterLayer->htmlMetadata().replace( QStringLiteral( "<head>" ), QStringLiteral( R"raw(<head><style type="text/css">%1</style>)raw" ) ).arg( myStyle ) };
```
nyalldawson: What does raw do here?

src/app/qgsrasterlayerproperties.cpp
elpaso: leftover, it does nothing, I'll cut it.

src/app/qgsrasterlayerproperties.cpp
elpaso: Sounds like a good idea.

src/app/qgsrasterlayerproperties.cpp
elpaso: it's a raw string, I find it much more readable when there are double quotes.
I always use `raw` as a prefix/suffix but I'm open to suggestions.

nyalldawson: Ah - I've gotta admit, I've never used these and mistakenly thought this was some new js/html something which I'd never seen before!

elpaso: they are very handy especially when you have to deal with regular expressions: you don't have do escape backslashes or double quotes.

src/app/qgsrasterlayerproperties.cpp
elpaso: See https://github.com/qgis/QGIS/pull/32269/commits/30eea3dfca534a82a7629fe63ac0dccdcb3d78ff

src/core/qgsapplication.cpp
```diff
@@ -1320,92 +1320,140 @@ QString QgsApplication::reportStyleSheet()
...
+                            ".tabular-view .odd-row {"
+                            "  background-color: #f9f9f9;"
+                            "}"
+                            ".section {"
+                            "  font-weight: bold;"
+                            "  padding-top:25px;"
+                            "}" );
+
+  // We have some subtle differences between Qt based style and QWebKit style
+  if ( styleSheetType == StyleSheetType::Qt )
```
nyalldawson: if -> switch

src/core/qgsapplication.h
```diff
@@ -463,17 +477,20 @@ class CORE_EXPORT QgsApplication : public QApplication
...
      *
-     * \returns QString containing the CSS 2.1 compliant stylesheet.
-     * \note you can use the special Qt extensions too, for example
-     * the gradient fills for backgrounds.
+     * \returns the stylesheet CSS rules.
+     * \note if styleSheetType equals StyleSheetType::Qt you can use the special Qt extensions too,
+     * for example the gradient fills for backgrounds.
      */
-    static QString reportStyleSheet();
+    static QString reportStyleSheet( const StyleSheetType &styleSheetType = StyleSheetType::Qt );
```
nyalldawson:
```suggestion
    static QString reportStyleSheet( StyleSheetType styleSheetType = StyleSheetType::Qt );
```

nyalldawson (commented):
Looks great, thank you!

nyalldawson (approved):

PR #32271, opened 2019-10-16
Server WFS3 fix rel=service_desc -> rel=service-desc

### LINKED ISSUE #32268: [SERVER] [WFS3] Wrong rel=service_desc
The landing page reports a rel=service_desc for the API endpoint. The value should rather be service-desc with a dash, not a underscore. See https://github.com/qgis/QGIS/pull/32262#issuecomment-542607819

elpaso:
Fixes #32268

### DIFF
--- a/src/server/qgsserverogcapi.cpp
+++ b/src/server/qgsserverogcapi.cpp
@@ -116,7 +116,9 @@ const QHash<QgsServerOgcApi::ContentType, QList<QgsServerOgcApi::ContentType> > 
 std::string QgsServerOgcApi::relToString( const Rel &rel )
 {
   static QMetaEnum metaEnum = QMetaEnum::fromType<QgsServerOgcApi::Rel>();
-  return metaEnum.valueToKey( rel );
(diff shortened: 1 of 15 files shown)


elpaso:
@rouault wold you mind approving this too? I think we are safe to merge during the freeze since the specs were finalized just two days ago and OAPIF is a new and well insulated feature.

rouault (approved):
I'm OK with the changes themselves, but I'm not clear of what the consensus is about what should go or not go during this hard freeze period.

elpaso:
AFAIK rules are not carved in stone, and I can't see the point in deliberately shipping a bugged WFS3 server.

My understanding is that a change might cause any regression (or introduce any issue) it should not be accepted during the freeze. 

But here we are talking about a server feature that is totally new and insulated and I don't see any possible impact on any other functionality.

rouault:
Agreed that it seems safe to merge

nyalldawson:
I also agree this is a significantly high impact issue (resulting in annoying workarounds being required by other libraries indefinitely) and the fix is very low risk.

Issue #32276, opened 2019-10-16
Apply correct SQL condition for the QGIS Browser PG Layer Properties Dialog


palmerj:
Checklist before submitting

- [x] Search through existing issue reports and gis.stackexchange.com to check whether the issue already exists
- [x] Test with a [clean new user profile](https://docs.qgis.org/testing/en/docs/user_manual/introduction/qgis_configuration.html?highlight=profile#working-with-user-profiles).
- [ ] Create a light and self-contained sample dataset and project file which demonstrates the issue

If the issue concerns a **third party plugin**, then it **cannot** be fixed by the QGIS team. Please raise your issue in the dedicated bug tracker for that specific plugin (as listed in the plugin's description). -->

**Describe the bug**

When you right click and select the layer properties for a QGIS PG table browser layer that contains multiple base geometry types (e.g one table but shows up multiple times in the browser for each polygon/line/point type) the SRID and Geometry Type query is called without the geometry type condition . The count and extents queries do correctly apply this filter. 

Likely the following SQL needs to be appended to the query

```sql

WHERE upper(geometrytype("the_geom")) IN xxxx
````

But I don't understand why the SRID and Geometry Types are required for this dialogue as they have already been determined to populated the Browser nodes. Maybe this query is not needed at all? 

**QGIS and OS versions**
QGIS version | 3.9.0-Master | QGIS code revision | 5f24a0b214
-- | -- | -- | --
Compiled against Qt | 5.12.3 | Running against Qt | 5.12.3
Compiled against GDAL/OGR | 2.4.1 | Running against GDAL/OGR | 2.4.1
Compiled against GEOS | 3.7.2-CAPI-1.11.2 | Running against GEOS | 3.7.2-CAPI-1.11.2 b55d2125
Compiled against SQLite | 3.28.0 | Running against SQLite | 3.28.0
PostgreSQL Client Version | 11.3 | SpatiaLite Version | 4.3.0a
QWT Version | 6.1.4 | QScintilla2 Version | 2.11.1
Compiled against PROJ | 5.2.0 | Running against PROJ | Rel. 5.2.0, September 15th, 2018
OS Version | macOS 10.15
Active python plugins | QuickWKT; qgisnetworklogger; Mergin; processing; db_manager; MetaSearch

strk:
It looks like getting layer properties is implemented creating a temporary layer:

[...]

PR #32292, opened 2019-10-17
Improve browser property dialog speed

strk:
Avoids double-checking from browser and data selector

See #32276

### DIFF
--- a/src/providers/postgres/qgspostgresdataitems.cpp
+++ b/src/providers/postgres/qgspostgresdataitems.cpp
@@ -343,20 +343,19 @@ QString QgsPGLayerItem::createUri()
     return QString();
   }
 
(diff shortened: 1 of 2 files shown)


strk:
@palmerj noticed there are 2 more queries performed w/out type/SRID filter:

  - Extent query (`SELECT st_extent("shape") FROM "bde"."crs_parcel"`)
  - Count query (`SELECT count(*) FROM "bde"."crs_parcel"`)

I'll see if those can also be tweaked to include a filter in this same PR

strk:
Last two commits would be good to backport to 3.4 branch as per #32378

strk:
I copied the non-debatable part of this PR into #32380 to be able to backport it

nyalldawson (approved):

PR #32297, opened 2019-10-17
Revert "Merge pull request #31374 from troopa81/parallelize_snap_caching

troopa81:
## Description

This reverts commit 87b1aa9a5d6efb279693aea68dc2dcb5b51b3e86

According to what has been discussed on the qgis-dev mailing list the parallel snap caching feature is reverted for 3.10.0 and will be merged back into master in order to be available in 3.10.1. 

Once this PR is merged in master, I will update the PR  #31648 so it could be merged once the hard freeze is over.

## Checklist


troopa81:
cc @nyalldawson @wonder-sk

troopa81:
After a second thought, I'm not sure this is the best thing to revert and report this feature to 3.10.1.

What happens with changelog? and API changes between 3.10.0 and 3.10.1?

nirvn:
IMHO, your intial conclusion is right, reverting the feature for 3.10 is the right thing to do. There's simply not enough time to test the other PR (which has yet to be finalized, approved, etc.)

nyalldawson:
> What happens with changelog? 

In the past I've just added them to the changelog for the next release, i.e. 3.12. So they actually become available earlier than the changelog states, but they still get a chance to be widely advertised...

> and API changes between 3.10.0 and 3.10.1?

Shouldn't have any impact -- regardless of whether it happens at 3.10.0 or .1, it can't break API. API additions have happened a lot in point releases in the past and there's no blocking these.

troopa81:
@nyalldawson  OK, so let's merge this one

nyalldawson (approved):

nyalldawson:
@troopa81 done -- thanks for you understanding and flexibility here!

Issue #32302, opened 2019-10-17
Crash on "zoom to layer" with global raster in EPSG:3031 (antarctic)


kannes:
**Describe the bug**
QGIS crashes when I "zoom to layer" with a global raster in EPSG:3031 (antarctic).

**How to Reproduce**

1. Load HYP_50M_SR_W.tif from https://www.naturalearthdata.com/http//www.naturalearthdata.com/download/50m/raster/HYP_50M_SR_W.zip (I guess it happens on any raster with the same extents but did not test).
2. Set the OTF-CRS to EPSG:3031
3. Use "zoom to layer" on the raster layer
4. QGIS crashes

**QGIS and OS versions**

QGIS version | 3.9.0-Master | QGIS code revision | 1a7294bc7f
-- | -- | -- | --
Compiled against Qt | 5.13.1 | Running against Qt | 5.13.1
Compiled against GDAL/OGR | 3.0.1 | Running against GDAL/OGR | 3.0.1
Compiled against GEOS | 3.7.2-CAPI-1.11.2 | Running against GEOS | 3.7.2-CAPI-1.11.2 b55d2125
Compiled against SQLite | 3.30.0 | Running against SQLite | 3.30.1
PostgreSQL Client Version | 11.5 | SpatiaLite Version | 4.3.0a
QWT Version | 6.1.4 | QScintilla2 Version | 2.11.2
Compiled against PROJ | 6.2.0 | Running against PROJ | Rel. 6.2.0, September 1st, 2019
OS Version | Arch Linux
Active python plugins | quick_map_services; SelectByRelationship; GeoCoding; qconsolidate3; web2qgis; MemoryLayerSaver; QPackage; timemanager; nominatim; QuickWKT; pointsamplingtool; QuickOSM; DotMap; GeodesicDensifier; qgis2web; Qgis2threejs; qtiles3; local_dominance; db_manager; processing; MetaSearch

**Additional context**
This is master from a while ago. Got no debugging symbols compiled in so no backtrace, sorry.

kannes:
Thanks!

PR #32307, opened 2019-10-18
[raster renderer] Fix freeze/crash by not rendering negative width/height raster viewport

nirvn:
## Description

This is part 2 of properly fixing #32302 . While the crasher is gone, it merely led to QGIS freezing altogether due to reprojected raster viewport ending up with negative width and height values. 

The PR adds a negative width / height check, and bails out from rendering if either of these values are negative.

## Checklist


nirvn:
Closing, something better coming soon.

PR #32325, opened 2019-10-20
[geonode] a generic way to consume geonode api. fixes #31376

### LINKED ISSUE #31376: GeoNode provider does not work with newer versions of GeoNode
**Describe the bug**
Some valid GeoNode connections return no layers. I suspect it is because the QGIS provider needs to be updated for newer GeoNode versions. 

**How to Reproduce**

1. Create a GeoNode connection to http://geonode.inasafe.org/ (GeoNode 2.9)
1. See how it connects and expands to show available layers
1. Create a GeoNode connection to http://geonode.ingc.gov.mz/
1. Test the connection and note that it is a successful connection
1. Note that the connection does not expand to show any layers. 

**QGIS and OS versions**

QGIS 3
Mac OSX

myarjunar:
## Description

fix #31376 
  -  Remove the geonode version checking. Because I just found out that the geonode version has nothing to do with the format of the returned data.
  -  Redo the logic to construct OWS endpoint of layers.

## Checklist

### DIFF
--- a/src/core/geocms/geonode/qgsgeonoderequest.cpp
+++ b/src/core/geocms/geonode/qgsgeonoderequest.cpp
@@ -269,111 +269,106 @@ QList<QgsGeoNodeRequest::ServiceLayerDetail> QgsGeoNodeRequest::parseLayers( con
   const QJsonObject jsonObject = jsonDocument.object();
   const QVariantMap jsonVariantMap = jsonObject.toVariantMap();
   const QVariantList layerList = jsonVariantMap.value( QStringLiteral( "objects" ) ).toList();
-  qint16 majorVersion;
-  qint16 minorVersion;
-  if ( jsonVariantMap.contains( QStringLiteral( "geonode_version" ) ) )
+
+  QString wmsURLFormat, wfsURLFormat, xyzURLFormat;
+
+  for ( const QVariant &layer : qgis::as_const( layerList ) )
   {
-    QRegularExpression re( "((\\d+)(\\.\\d+))" );
-    QRegularExpressionMatch match = re.match( jsonVariantMap.value( QStringLiteral( "geonode_version" ) ).toString() );
-    if ( match.hasMatch() )
+    QgsGeoNodeRequest::ServiceLayerDetail layerStruct;
+    const QVariantMap layerMap = layer.toMap();
+    QVariantList layerLinks = layerMap.value( QStringLiteral( "links" ) ).toList();
+
+    if ( layerMap.value( QStringLiteral( "typename" ) ).toString().isEmpty() )
     {
-      const QStringList geonodeVersionSplit = match.captured( 0 ).split( '.' );
-      majorVersion = geonodeVersionSplit.at( 0 ).toInt();
-      minorVersion = geonodeVersionSplit.at( 1 ).toInt();
+      const QStringList splitURL = layerMap.value( QStringLiteral( "detail_url" ) ).toString().split( '/' );
+      layerStruct.typeName = splitURL.at( splitURL.length() - 1 );
     }
-    else
+    layerStruct.uuid = layerMap.value( QStringLiteral( "uuid" ) ).toString();
+    layerStruct.id = layerMap.value( QStringLiteral( "id" ) ).toString();
+    layerStruct.name = layerMap.value( QStringLiteral( "name" ) ).toString();
+    layerStruct.typeName = layerMap.value( QStringLiteral( "typename" ) ).toString();
+    layerStruct.title = layerMap.value( QStringLiteral( "title" ) ).toString();
+
+    layerStruct = parseOWSUrl( layerStruct, layerLinks );
+
+    if ( layerStruct.wmsURL.isEmpty() && layerStruct.wfsURL.isEmpty() && layerStruct.xyzURL.isEmpty() )
     {
-      return layers;
-    }
-  }
-  else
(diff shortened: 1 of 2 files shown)


nirvn:
Can we try to push this into 3.10.0? As it is, we'd ship QGIS with a semi broken geonode provider.

src/core/geocms/geonode/qgsgeonoderequest.cpp
```diff
@@ -269,111 +269,106 @@ QList<QgsGeoNodeRequest::ServiceLayerDetail> QgsGeoNodeRequest::parseLayers( con
...
+          const QJsonDocument resourceUriDocument = QJsonDocument::fromJson( this->lastResponse() );
+          const QJsonObject resourceUriObject = resourceUriDocument.object();
+          const QVariantMap resourceUriMap = resourceUriObject.toVariantMap();
+          QVariantList resourceUriLinks = resourceUriMap.value( QStringLiteral( "links" ) ).toList();
+          QgsGeoNodeRequest::ServiceLayerDetail tempLayerStruct;
+          tempLayerStruct = parseOWSUrl( tempLayerStruct, resourceUriLinks );
+
+          // Avoid iterating all the layers to get the service url. Instead, generate a string format once we found one service url
+          // for every service (wms, wfs, xyz). And then use the string format for the other layers since they are identical.
+          if ( tempLayerStruct.server == "qgis-server" )
```
elpaso:
```suggestion
          if ( tempLayerStruct.server == QStringLiteral( "qgis-server" ) )
```

src/core/geocms/geonode/qgsgeonoderequest.cpp
```diff
@@ -269,111 +269,106 @@ QList<QgsGeoNodeRequest::ServiceLayerDetail> QgsGeoNodeRequest::parseLayers( con
...
+          {
+            wmsURLFormat = !tempLayerStruct.wmsURL.isEmpty() ? tempLayerStruct.wmsURL.replace( layerStruct.name, "%1" ) : "";
```
elpaso:
```suggestion
            wmsURLFormat = ! tempLayerStruct.wmsURL.isEmpty() ? tempLayerStruct.wmsURL.replace( layerStruct.name, QStringLiteral( "%1" )  ) : QString();
```

elpaso: following also.

src/core/geocms/geonode/qgsgeonoderequest.h
```diff
@@ -213,6 +217,11 @@ class CORE_EXPORT QgsGeoNodeRequest : public QObject
      */
     void setProtocol( const QString &protocol );
 
+    /**
+     * Returns the updated ServiceLayerDetail struct with WMS/WFS/XYZ url.
+     */
+    QgsGeoNodeRequest::ServiceLayerDetail parseOWSUrl( QgsGeoNodeRequest::ServiceLayerDetail &layerStruct, const QVariantList &layerLinks );
```
elpaso: I think the QGIS convention is to capitalize acronyms
```suggestion
    QgsGeoNodeRequest::ServiceLayerDetail parseOwsUrl( QgsGeoNodeRequest::ServiceLayerDetail &layerStruct, const QVariantList &layerLinks );
```

src/core/geocms/geonode/qgsgeonoderequest.h
```diff
@@ -86,6 +88,8 @@ class CORE_EXPORT QgsGeoNodeRequest : public QObject
       QString wfsURL;
       //! XYZ tileserver URL for layer
       QString xyzURL;
+      //! Backend server (geoserver or qgis-server)
+      QString server;
```
m-kuhn: Should this be an enum instead?

src/core/geocms/geonode/qgsgeonoderequest.cpp
myarjunar: ah yes of course, I must have missed some strings. thanks

src/core/geocms/geonode/qgsgeonoderequest.h
myarjunar: yeah I agree I think that is a better type. thanks!

timlinux:
Anything more to do here?

src/core/geocms/geonode/qgsgeonoderequest.cpp
```diff
@@ -269,111 +269,108 @@ QList<QgsGeoNodeRequest::ServiceLayerDetail> QgsGeoNodeRequest::parseLayers( con
...
-    else
+    QgsGeoNodeRequest::ServiceLayerDetail layerStruct;
+    const QVariantMap layerMap = layer.toMap();
+    QVariantList layerLinks = layerMap.value( QStringLiteral( "links" ) ).toList();
+
+    if ( layerMap.value( QStringLiteral( "typename" ) ).toString().isEmpty() )
     {
-      return layers;
+      const QStringList splitURL = layerMap.value( QStringLiteral( "detail_url" ) ).toString().split( '/' );
+      layerStruct.typeName = splitURL.at( splitURL.length() - 1 );
```
elpaso: Please add a check that  "splitURL.length() - 1" is a valid index or we might have a crash.

elpaso (approved):
please just add the check for at() index validity and IMHO we are good to go

PR #32334, opened 2019-10-22
Fix potential overflow in raster renderer

nyalldawson:
Relates to #32302 , Supersedes #32307


nirvn (approved):

nirvn:
Code looks good and runs fine. It'll need to be backported to 3.4.

+1 to merge now.

PR #32336, opened 2019-10-22
PG source select dlg: show multiline comments in tooltips

### LINKED ISSUE #32257: Data Source UI - PostgreSQL table multi-line comments make the grid table hard to read
Checklist before submitting

- [x] Search through existing issue reports and gis.stackexchange.com to check whether the issue already exists
- [x] Test with a [clean new user profile](https://docs.qgis.org/testing/en/docs/user_manual/introduction/qgis_configuration.html?highlight=profile#working-with-user-profiles).
- [x] Create a light and self-contained sample dataset and project file which demonstrates the issue

**Describe the bug**
When a PostgreSQL table has a multi-line table comment it causes the UI row height to expand. This causes the UI to be harder to use with grid scrolling and very large catalogs. Please consider using a fixed single row height for all tables to ensure a consistent UI presentation. 

data_manager_ui_multiline_comment

**How to Reproduce**
Create a table with a multi-line comment. e.g

```sql

CREATE TABLE public.test (id INTEGER PRIMARY KEY, shape Geometry);

COMMENT ON TABLE public.test
    IS '
This
is a
multiline comment
';
```

**QGIS and OS versions**

Tested on Mac and Ubuntu master

elpaso:
Fixes #32257

Screenshot_20191022_094850

### DIFF
--- a/src/providers/postgres/qgspgsourceselect.cpp
+++ b/src/providers/postgres/qgspgsourceselect.cpp
@@ -40,6 +40,7 @@ email                : sherman at mrcc.com
 #include <QStringList>
 #include <QStyledItemDelegate>
 
+
 //! Used to create an editor for when the user tries to change the contents of a cell
 QWidget *QgsPgSourceSelectDelegate::createEditor( QWidget *parent, const QStyleOptionViewItem &option, const QModelIndex &index ) const
 {
@@ -159,7 +160,7 @@ void QgsPgSourceSelectDelegate::setModelData( QWidget *editor, QAbstractItemMode
   {
     if ( index.column() == QgsPgTableModel::DbtmType )
     {
-      QgsWkbTypes::Type type = ( QgsWkbTypes::Type ) cb->currentData().toInt();
+      QgsWkbTypes::Type type = static_cast< QgsWkbTypes::Type >( cb->currentData().toInt() );
 
       model->setData( index, QgsPgTableModel::iconForWkbType( type ), Qt::DecorationRole );
       model->setData( index, type != QgsWkbTypes::Unknown ? QgsPostgresConn::displayStringForWkbType( type ) : tr( "Select…" ) );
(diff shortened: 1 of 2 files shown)


palmerj:
Thanks @elpaso !!

elpaso:
no problem!

strk:
@elpaso I tried this against a [LINZ bde schema](https://github.com/linz/linz-bde-schema) and the output isn't very nice, for a couple of things:

   1. The comment shown in the grid is vertically truncated to show partial glyphs (see `crs_land_district`) -- I think the ideal would be showing only the first line of the commend and only up to a given number of characters
   2. The tooltop for some table is very wide and takes the whole display -- I think the ideal would be to implement line wrapping for the toolbar

I'm attaching a screenshot (taken with a camera because the popup would disappear upon hitting a screenshot key on keyboard [this may or may not be something we can fix]).

photo_2019-10-22_11-59-58

elpaso:
@strk can you attach a comment dump to reproduce the issue?

strk:
@elpaso it's here: https://github.com/linz/linz-bde-schema/blob/master/sql/06-schema_comments.sql

You can load the whole bde schema (it's only tables, and comments) by using the `linz-bde-schema-loader` script in that repository.

strk:
As of 5ec15b8,

Fixed issues:

 - Long lines in tooltip are nicely wrapped

Left over issues:

  - The whole "comments" column in the grid has empty values (this is probably because BDE schema comments all start with a newline, so the very first line is empty - could be fixed by skipping any newline from start of string)
  - Tooltip for a layer with comment is not removed when moving mouse to a layer without a comment, so it looks like the comment belongs to a wrong table (ie: you loose track of which table you are seeing the comment for). I guess this could be fixed by either hiding the old tooltip or printing a "[none]" comment for tables without a comment...

Keep up the great work !

elpaso:
@strk I could fix the first one, the second is a won't fix for me, it's handled by Qt (or maybe the window manager).

strk (approved):

elpaso:
> > @strk I could fix the first one, the second is a won't fix for me, it's handled by Qt (or maybe the window manager).
> 
> For the second one: how about just setting an empty string as a tooltip ? Does it make sense ?

It won't solve the issue entirely: only if the user moves the mouse over another comment, or any other "tooltipped" widget.

elpaso:
oops I merged too fast, it wasn't meant for 3.10

PR #32337, opened 2019-10-22
Fix oracle provider and tests

troopa81:
## Description

This PR makes oracle provider and its tests stable and reliable. This is still WIP because I would like the tests to be launched by the CI. 

Some details about commits:
- e1b05b4 : Connections can be shared by different threads when using a transaction, so we need to protect members from concurrent access
- f6f25a6 : We have to compute exact extent when the boolean mUseEstimatedMetadata isn't set, not an estimated one
- 7a2ef62 : That's the main reason for the tests being flaky. If we don't delete the connection when closing the transaction, the QOCISpatialDriver::close is not called, the connection is not well closed, and so following tests will possibly deadlock.

All tests are now stable and OK when running with oracle-18.4.0(c) Express Edition.

FYI @mhugo @Jean-Roc 

## Checklist


nyalldawson:
Thanks -- looks good! Did you see the earlier discussion about connections not being closed when providers are deleted? I think that's also a cause of issues in these tests...

troopa81:
> Did you see the earlier discussion about connections not being closed when providers are deleted?

You mean [this](https://github.com/qgis/QGIS/issues/27087#issuecomment-541568701) ? 

I agree that the QgsOracleProvider should have a mConnRO and mConnRW member in order to not call connectDB everytime we want to acess to the current connexion (like in disconnectDB for instance). But still, it should work the way it is (because of the QgsOracleConn::sConnections map).

With these modifications all tests pass. I would rather setup the CI correctly before going any further.

troopa81:
@jef-n I squashed all continuous integration together, is that OK for you?

troopa81:
I think the commit d32b942 may have its own PR because error happened when I disable cache for docker images. What do you think?

nyalldawson:
@troopa81 what's the status with this?

troopa81:
@nyalldawson The PR is complete and need to be reviewed.

troopa81:
Looks like one microsoft CI job is stuck in queued state, while an other one is green.

roya0045:
> Looks like one microsoft CI job is stuck in queued state, while an other one is green.

That seems to be the case for many other PR.

troopa81:
@nyalldawson Thanks!

PR #32342, opened 2019-10-22
[Server] WMS: Use QGIS Style instead of SLD to restore layer style

rldhont:
## Description
To save default style, QGIS Server used the SLD style. In SLD style, labeling is not saved so when style is restored with SLD style, labeling was lost.

To be sure to restore layer style, QGIS style is used.

## Checklist

### DIFF
--- a/src/server/services/wms/qgslayerrestorer.cpp
+++ b/src/server/services/wms/qgslayerrestorer.cpp
@@ -16,6 +16,7 @@
  ***************************************************************************/
 
 #include "qgslayerrestorer.h"
+#include "qgsmessagelog.h"
 #include "qgsmaplayer.h"
 #include "qgsvectorlayer.h"
 #include "qgsrasterlayer.h"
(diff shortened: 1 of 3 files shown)


andreassteffens:
Some profiling i did in order to identify a performance problem with QGIS Server revealed that saving the original SLD for all layers in the `QgsLayerRestorer` so it can be restored after rendering actually leads to poor performance for projects with many vector layers and/or complex symbology (take a project with ALKIS layers as created by the Norbit importer/plugin as example). 

Wouldn't it make more sense to pass the entire `QgsWmsRenderContext` to the `QgsLayerRestorer` and check whether an alternative SLD styling has been supplied for a layer in the request BEFORE serializing the layer's style to SLD for restoration?

pblottiere:
Hi @andreassteffens,

> me profiling i did in order to identify a performance problem with QGIS Server revealed that saving the original SLD for all layers in the QgsLayerRestorer so it can be restored after rendering actually leads to poor performance for projects with many vector layers

Thanks for your input, it's very valuable and we definitely have to investigate :+1:. I take a look asap.

PR #32344, opened 2019-10-22
[Release-3_4]Fix featureCount on postgres view when flag estimatedmetadata is set

rldhont:
We [can't estimate rows on postgres view](https://stackoverflow.com/questions/28157246/how-estimate-row-count-of-a-postgresql-view). 

Manually backport  #32114

## Checklist

### DIFF
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -3186,7 +3186,9 @@ long QgsPostgresProvider::featureCount() const
   // use estimated metadata even when there is a where clause,
   // although we get an incorrect feature count for the subset
   // - but make huge dataset usable.
-  if ( !mIsQuery && mUseEstimatedMetadata )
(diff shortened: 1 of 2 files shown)


rldhont:
@mhugo @troopa81 I have manually backported the PR associated.

troopa81:
I'm not against 3.4 backport but as @nyalldawson mentioned "Only fixes for crashes or data corruption should be backported now" which is not the case here. Do you think this PR should really be backported?

strk:
We actually _can_ estimate rows in views, using EXPLAIN and fetching the _estimated_ number of rows returned by outermost node

rldhont:
@troopa81 the mention "Only fixes for crashes or data corruption should be backported now" is only for hard fix in master branch before version 3.10.0 will be released.

nyalldawson:
It also applies to 3.4 - that branch is now in a "super cautious" mode, and only critical fixes should be committed

PR #32352, opened 2019-10-23
Server: fix schemaLocation in WMS GetCapabilities document

### LINKED ISSUE #32085: QGIS server 3.x generates invalid schemaLocation for WMS 1.3.0 GetCapabilities request
The XML generated by the GetCapabilities request for WMS version 1.3.0 includes a `xsi:schemaLocation` attribute that is not valid, it appears to be missing the namespace for a schema extensions it includes.

`xsi:schemaLocation="http://www.opengis.net/wms http://schemas.opengis.net/wms/1.3.0/capabilities_1_3_0.xsd http://www.opengis.net/sld http://schemas.opengis.net/sld/1.1.0/sld_capabilities.xsd http://qgis34.lan/qgis?MAP=/data/qgis/map_3_4.qgs&amp;SERVICE=WMS&amp;REQUEST=GetSchemaExtension"`

You can see at the end the inclusion of the URL pointing back to the source QGIS server to retrieve the schema extensions content, but the URL is not prefixed by a namespace, which looking at the content of the URL looks like it should be `http://www.qgis.org/wms`.

So the schemaLocation should be:

`xsi:schemaLocation="http://www.opengis.net/wms http://schemas.opengis.net/wms/1.3.0/capabilities_1_3_0.xsd http://www.opengis.net/sld http://schemas.opengis.net/sld/1.1.0/sld_capabilities.xsd http://www.qgis.org/wms http://qgis34.lan/qgis?MAP=/data/qgis/map_3_4.qgs&amp;SERVICE=WMS&amp;REQUEST=GetSchemaExtension"`

The `getCapabilities` method in `qgswmsgetcapabilities.cpp` adds this schema location at line 220 with the following code:

`      schemaLocation += " " + hrefString + "SERVICE=WMS&REQUEST=GetSchemaExtension";`

which should be (something like):

`      schemaLocation += " http://www.qgis.org/wms " + hrefString + "SERVICE=WMS&REQUEST=GetSchemaExtension";`

Earlier code to add other schema location entries use `QStringLiteral` and `QLatin1String` to add the schema location entries so I'm not sure if the original code or the updated code above should also be using those when creating the schema location value.

Alternatively, this schema, if it is not supposed to be linked to a namespace, should be included in the `xsi:noNamespaceSchemaLocation` attribute rather than the `xsi:schemaLocation` attribute (but I don't think that is the case).

pblottiere:
## Description

Fixes https://github.com/qgis/QGIS/issues/32085

### DIFF
--- a/src/server/services/wms/qgswmsgetcapabilities.cpp
+++ b/src/server/services/wms/qgswmsgetcapabilities.cpp
@@ -209,6 +209,7 @@ namespace QgsWms
       schemaLocation += QLatin1String( " http://schemas.opengis.net/wms/1.3.0/capabilities_1_3_0.xsd" );
       schemaLocation += QLatin1String( " http://www.opengis.net/sld" );
       schemaLocation += QLatin1String( " http://schemas.opengis.net/sld/1.1.0/sld_capabilities.xsd" );
+
       if ( QgsServerProjectUtils::wmsInspireActivate( *project ) )
       {
         wmsCapabilitiesElement.setAttribute( QStringLiteral( "xmlns:inspire_common" ), QStringLiteral( "http://inspire.ec.europa.eu/schemas/common/1.0" ) );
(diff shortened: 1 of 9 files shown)


elpaso (approved):

sbrunner (approved):
Looks good :-)

PR #32360, opened 2019-10-23
Fix merge feature attrs: use the correct formatter for data

### LINKED ISSUE #32208: "Merge features" and "merge feature attributes" have broken numerical functions
See what is reported as count in the following screenshot (master 3.9 a3ef3899 but also occurs on 3.8 (and maybe 3.4?))image. The other functions also return wrong values

elpaso:
and use actual data for calculations and to return values

Fixes #32208

### DIFF
--- a/src/app/qgsmergeattributesdialog.cpp
+++ b/src/app/qgsmergeattributesdialog.cpp
@@ -114,7 +114,7 @@ void QgsMergeAttributesDialog::createTableWidgetContents()
   for ( int idx = 0; idx < mFields.count(); ++idx )
   {
     const QgsEditorWidgetSetup setup = QgsGui::editorWidgetRegistry()->findBest( mVectorLayer, mFields.at( idx ).name() );
-    if ( setup.type() == QLatin1String( "Hidden" ) || setup.type() == QLatin1String( "Immutable" ) )
+    if ( setup.type() == QStringLiteral( "Hidden" ) || setup.type() == QStringLiteral( "Immutable" ) )
     {
       mHiddenAttributes.insert( idx );
       continue;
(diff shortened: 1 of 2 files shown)


nyalldawson:
What's the actual cause of this issue? I can't tell directly from the diff

elpaso:
Yeah, commit message is misleading, the formatter fix is just cosmetic, the core fix is that it is not possible to take the content of a cell (the display value) for calculations, the original unformatted value must be taken instead, I'm storing the original unformatted value in Qt::UserData and using this for the calculations. The root cause is that we display value formatting according to locale.
What surprised me is that this bug was reported just now, I'm asking myself how often this feature is used with non-en (non-decimal-separator-dot) locales.

elpaso:
Yes, the real issue it's triggered by non-dot locales, but the formatter (cosmetic) issue has been there for all locales.

nyalldawson:
Ok. Lgtm then!

src/app/qgsmergeattributesdialog.cpp
```diff
@@ -319,12 +320,12 @@ void QgsMergeAttributesDialog::refreshMergedValue( int col )
...
 
-  int fieldIdx = mTableWidget->horizontalHeaderItem( col )->data( FieldIndex ).toInt();
+  const int fieldIdx = mTableWidget->horizontalHeaderItem( col )->data( FieldIndex ).toInt();
 
   //evaluate behavior (feature value or min / max / mean )
-  QString mergeBehaviorString = comboBox->currentData().toString();
+  const QString mergeBehaviorString = comboBox->currentData().toString();
   QVariant mergeResult; // result to show in the merge result field
-  if ( mergeBehaviorString == QLatin1String( "concat" ) )
+  if ( mergeBehaviorString == QStringLiteral( "concat" ) )
```
m-kuhn: I think that's better to leave as before, but don't loose too much time on these micro-optimizations :)

https://doc.qt.io/qt-5/qstring.html#operator-eq-eq

elpaso: thanks, I missed that one

PR #32370, opened 2019-10-23
Fix browser DD on GPKG/PG with multi geom

### LINKED ISSUE #32362: Importing using D&D in QGIS browser does not work anymore if input is multipart
On master

Try to D&D a MULTI layer from GPKG/PostGIS to GPKG/PostGIS, it will fail with an error like
`
Failed to import some vector layers!Feature write errors:Failed to transform a feature with ID '1' to single part. Writing stopped.`

Does not happen on 3.4

related to https://github.com/qgis/QGIS/pull/32108 ?

elpaso:
Fixes #32362

### DIFF
--- a/src/core/qgsvectorlayerexporter.cpp
+++ b/src/core/qgsvectorlayerexporter.cpp
@@ -267,6 +267,7 @@ QgsVectorLayerExporter::exportLayer( QgsVectorLayer *layer,
   }
 
   QgsFields fields = layer->fields();
+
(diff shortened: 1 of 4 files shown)


nyalldawson (approved):

nyalldawson:
@elpaso I think this qualifies for merging for 3.10.0, given it's a fix-for-a-fix. How risky is it in your opinion?

elpaso:
Acceptable risk I think.

PR #32383, opened 2019-10-24
Display Rulelabel when requested in GetLegendGraphics

signedav:
On a WMS `GetLegegendGraphics` request a legend is created like this. If there are no style categories, the layer is displayed with a title and below the symbol with the same title again.

norulesbefore

There are the parameters `LAYERTITLE` and `RULELABEL`. On having `LAYERTITLE=false` but `RULELABEL=true`, `RULELABEL` had no impact when the layer had no style categorieska, what is surely wrong:

`RULELABEL=true&LAYERTITLE=true` -> displayed layertitle and rulelabel - fine
`RULELABEL=false&LAYERTITLE=false` -> displayed no title and label at all - fine
`RULELABEL=false&LAYERTITLE=true` -> displayed only layertitle - fine
`RULELABEL=true&LAYERTITLE=false` -> displayed no title and label at all - wrong, it should display only the rulelabel

Before on `RULELABEL=true&LAYERTITLE=false`:
notitlesbutrulelabeltrue

Now on `RULELABEL=true&LAYERTITLE=false`:
image

In general I'm not complete happy with the behavior of the legend: E.g. in PrintComposer it's displayed like this. 
print_composer
First afaik we cannot display it like this over the `GetLegendGraphics` (Because there are layertitles on the layers with rules an no layertitles on the layer without rules) and second it's not that nice, that there is no visual difference between a rule and a layer without a rule. Are there any toughts already arround?


m-kuhn:
Sounds reasonable to me, but I may be missing a use case (in which case a new parameter to put the layer title into the individual items would be required I think).

@rldhont @elpaso ?

signedav:
The tests would result in a new behavior. Before on `LAYERTITLE=FALSE` there has been nothing.
image
Now - because `RULELABEL` is not set to false, the label appears. Regarding to my description above it's what it should. I would need to set it or change the images.

The behavior before, in my opinion, has been definitely wrong. But the problem is that secondary applications maybe concerned about it, and would now behave differently...

elpaso:
LGTM, thanks!

signedav:
@elpaso thanks for already commenting. Could you review it now including all changes on the tests?

signedav:
Are there any inputs? Otherwise I plan to merge that later this week.

signedav:
No backport @nyalldawson ? I think it makes sense to have it in 3.10 as well since the behavior is wrong otherwise and not what the user expects (except if he already knows the issue and is doing workarounds with font size 0 etc...) 
Thanks.

nyalldawson:
The backport bot is dead -- I just deleted the label and it's removed from all prs

m-kuhn:
Wouldn't it make sense to leave it to keep track (and discuss) what deserves manual backporting?

nyalldawson:
@m-kuhn I was concerned that it was mistakenly giving the message that backporting was still automatic. By removing it we clearly show everyone that it's back to being their own responsibility... I think there's more chance of missed backports if we leave the tag vs not.

signedav:
Maybe a differently named label like "backport needed" or something?

nyalldawson:
There's already one for that and has been for... Years?

signedav:
Ah right. Sorry, I forgot about that label since we used the backport bot meanwhile. 
Then that works for me. Thanks.

m-kuhn:
That sounds good :+1: 

Looking forward to @3nids fix of the backport bot

jef-n:
> Looking forward to @3nids fix of the backport bot

Does the (forked) backport bot support setting of labels via commit message?

m-kuhn:
I don't think that should be the job of this bot, but another app instead.
Maybe [something like this](https://stackoverflow.com/questions/8359533/github-commit-message-tagging)

PR #32395, opened 2019-10-24
[feature] Port of Densify Geometries by Count algorithm to C++

root676:
## Description
This PR contains the port of the Densify Geometries by Count algorithm to a native C++ algorithm. Overall the performance of the algorithm is improved and a dynamic parameter for the number of vertices to add is implemented. Other than that, no other functionality has been added to the algorithm. 

I am looking forward to any reviews/comments! 

## Checklist


Gustry (commented):
Nice, thanks.
It seems, you didn't remove old python files and the previous help description in the YML file.

root676:
@Gustry thanks for the hint! I just deleted the old algorithm and the helpstrings.

src/analysis/processing/qgsalgorithmdensifygeometriesbycount.cpp
```diff
@@ -0,0 +1,124 @@
...
+}
+
+QString QgsDensifyGeometriesByCountAlgorithm::groupId() const
+{
+  return QStringLiteral( "vectorgeometry" );
+}
+
+QString QgsDensifyGeometriesByCountAlgorithm::shortHelpString() const
+{
+  return QObject::tr( "This algorithm takes a polygon or line layer"
```
nyalldawson: Be careful of missing trailing spaces here -- this string is missing spaces between "layerand", "geometrieshave", etc.

Also would be good to add `\n\n` in places where the original help had a paragraph break

src/analysis/processing/qgsalgorithmdensifygeometriesbycount.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+void QgsDensifyGeometriesByCountAlgorithm::initParameters( const QVariantMap &configuration )
+{
+  Q_UNUSED( configuration )
+  std::unique_ptr<QgsProcessingParameterNumber> verticesCnt = qgis::make_unique<QgsProcessingParameterNumber>( QStringLiteral( "VERTICES" ),
+      QObject::tr( "Number of vertices to add" ),
+      QgsProcessingParameterNumber::Integer,
+      1, false, 1, 10000000 );
+  verticesCnt->setIsDynamic( true );
+  verticesCnt->setDynamicPropertyDefinition( QgsPropertyDefinition( QStringLiteral( "VerticesCount" ), QObject::tr( "Vertices count" ), QgsPropertyDefinition::IntegerPositiveGreaterZero ) );
```
nyalldawson:
```suggestion
  verticesCnt->setDynamicPropertyDefinition( QgsPropertyDefinition( QStringLiteral( "VerticesCount" ), QObject::tr( "Vertices count" ), QgsPropertyDefinition::IntegerPositive ) );
```

I think we should allow zero counts here -- e.g. if you only want to densify lines greater than a certain threshold

src/analysis/processing/qgsalgorithmdensifygeometriesbycount.cpp
```diff
@@ -0,0 +1,124 @@
...
+}
+
+QgsFeatureList QgsDensifyGeometriesByCountAlgorithm::processFeature( const QgsFeature &feature, QgsProcessingContext &context, QgsProcessingFeedback *feedback )
+{
+  Q_UNUSED( context )
+  Q_UNUSED( feedback )
+  QgsFeature densifiedFeature = feature;
+
+  int verticesCnt = mVerticesCnt;
+  if ( mDynamicVerticesCnt )
```
nyalldawson: Very minor optimisation, but this logic could be moved into the `if ( feature.hasGeometry() )` branch -- it's unused when the feature has no geometry

src/analysis/processing/qgsalgorithmdensifygeometriesbycount.cpp
```diff
@@ -0,0 +1,124 @@
...
+    verticesCnt = mVerticesCntProperty.valueAsInt( context.expressionContext(), verticesCnt );
+
+  if ( feature.hasGeometry() )
+    densifiedFeature.setGeometry( feature.geometry().densifyByCount( verticesCnt ) );
```
nyalldawson: I think we should put a safety check here of `if ( verticesCnt > 0 )`. It's possible a data defined parameter value may return a negative count... and I'm unsure if densifyByCount will handle that gracefully

nyalldawson (commented):
Looks great!!

root676:
Thank you @nyalldawson for reviewing this port and your feedback! I added the suggested changes.

PR #32425, opened 2019-10-25
Introduce GRASS_USE_REXTERNAL setting

pmav99:
GRASS GIS supports in addition to importing data in its native formats via
`r.in.gdal`/`v.in.ogr` (and likewise `r.import`/v.import`),
`r.external`/`v.external` which only link the data into the GRASS DB.

Nevertheless, both `r.external` and `v.external` occasionally have problems,
especially on windows. E.g.:

- https://trac.osgeo.org/grass/ticket/3927

GRASS 7 Processing Plugin already has a setting that controls whether
`v.external` is used (disabled by default).

With this commit the complementary setting for `r.external` gets added too.
Nevertheless, in this case `r.external` is enabled by default in order to
maintain backwards compatibility.

## Checklist


pmav99:
Not sure if a warning similar to the one here should also get added: https://github.com/qgis/QGIS/commit/676238b28fad1c453c1f51addd71065537e85b75

nyalldawson:
>in this case r.external is enabled by default in order to
maintain backwards compatibility.

Shouldn't this be disabled by default, especially given
> Nevertheless, both r.external and v.external occasionally have problems, especially on windows.

nyalldawson:
Can you copy the logic from https://github.com/qgis/QGIS/blob/dcc1cd92ef56e3306818382d365a9a4f76330b8a/python/plugins/processing/algs/grass7/Grass7Utils.py#L402 and adapt for external rasters too?

pmav99:
@nyalldawson 
> Shouldn't this be disabled by default, 

Personally, I would disable it, but breaking backward compatibility on QGIS is not my call :)

That being said, r.external is reading through the GDAL driver and as a result is faster + it does not take disk space.

> Can you copy the logic from `QGIS/python/plugins/processing/algs/grass7/Grass7Utils.py` and adapt for external rasters too?

I forced pushed.

nyalldawson:
> Personally, I would disable it, but breaking backward compatibility on QGIS is not my call :)

Why would this break backward compatibility? I'm confused... the current behavior is NOT using r external, right?

NikosAlexandris:
> > Personally, I would disable it, but breaking backward compatibility on QGIS is not my call :)
> 
> Why would this break backward compatibility? I'm confused... the current behavior is NOT using r external, right?

The current behavior is hardcoded `external = True`, see [Grass7Algorithm.py#L694-L713](https://github.com/qgis/QGIS/blob/8a3c7b14c367771d096b4a6d006aa3c4b1017dd5/python/plugins/processing/algs/grass7/Grass7Algorithm.py#L694-L713).

pmav99:
I am afraid it does use `r.external`. More specifically, both `loadRasterLayerFromParameter` and `loadRasterLayer` use `external=True` as the default argument and e.g. `processInput()` does not override it: 
https://github.com/qgis/QGIS/blob/8a3c7b14c367771d096b4a6d006aa3c4b1017dd5/python/plugins/processing/algs/grass7/Grass7Algorithm.py#L468-L469

That being said, there are cases where `external=False` is hardcoded. E.g. `r.null`
https://github.com/qgis/QGIS/blob/8a3c7b14c367771d096b4a6d006aa3c4b1017dd5/python/plugins/processing/algs/grass7/ext/r_null.py#L39-L40

and `r.category`
https://github.com/qgis/QGIS/blob/8a3c7b14c367771d096b4a6d006aa3c4b1017dd5/python/plugins/processing/algs/grass7/ext/r_category.py#L42-L49

Probably due to `r.external` misbehaving with these specific modules

NikosAlexandris:
> That being said, there are cases where `external=False` is hardcoded. E.g. `r.null`
> https://github.com/qgis/QGIS/blob/8a3c7b14c367771d096b4a6d006aa3c4b1017dd5/python/plugins/processing/algs/grass7/ext/r_null.py#L39-L40
> 
> and `r.category`
> https://github.com/qgis/QGIS/blob/8a3c7b14c367771d096b4a6d006aa3c4b1017dd5/python/plugins/processing/algs/grass7/ext/r_category.py#L42-L49
> 
> Probably due to `r.external` misbehaving with these specific modules

It's rather that the implementation of the NULL concept (in the case of `r.null` mentioned above at least) is GRASS GIS specific. GRASS GIS needs data in its native format to operate on NULL cells. More or less.

Similar, `r.category` is meant for native GRASS GIS raster maps, to manage a raster's (category) values and possibly corresponding labels (which can be more than just a 'static' description, see for example [r.category's default-and-dynamic-category-labels concept](https://grass.osgeo.org/grass79/manuals/r.category.html#default-and-dynamic-category-labels). This would not work on external raster data models.

NikosAlexandris:
> > That being said, there are cases where `external=False` is hardcoded. E.g. `r.null`
> > https://github.com/qgis/QGIS/blob/8a3c7b14c367771d096b4a6d006aa3c4b1017dd5/python/plugins/processing/algs/grass7/ext/r_null.py#L39-L40
> > 
> > and `r.category`
> > https://github.com/qgis/QGIS/blob/8a3c7b14c367771d096b4a6d006aa3c4b1017dd5/python/plugins/processing/algs/grass7/ext/r_category.py#L42-L49
> > 
> > Probably due to `r.external` misbehaving with these specific modules
> 

~~> It's rather that the implementation of the NULL concept (in the case of `r.null` mentioned above at least) is GRASS GIS specific. GRASS GIS needs data in its native format to operate on NULL cells. More or less.
> 
> Similar, `r.category` is meant for native GRASS GIS raster maps, to manage a raster's (category) values and possibly corresponding labels (which can be more than just a 'static' description, see for example [r.category's default-and-dynamic-category-labels concept](https://grass.osgeo.org/grass79/manuals/r.category.html#default-and-dynamic-category-labels). This would not work on external raster data models.~~

No intention to add noise, but for the sake of clarity, I add some more details.
In short, however, `r.in.gdal` is some times (and currently?) required, thus the need to control whether a map should be linked (via `r.external`) or imported.

In the case of categories, I have to correct myself. The concept works just fine, at least under Linux, as we deal with independent an file(s) under `GRASS_Data_Base/Location/Mapset/cats/cat`.

About NULL cells, indeed it does not work the way expected. Operating on an "external" GDAL-linked map inside a GRASS GIS data base, with the `r.null` module, will not crash nor report an error when setting, for example, NULL cells to be of some value (`r.null external_raster null=0`) but it won't do anyway. That is, the NULL cells will still remain NULL. In a native raster map, the command `r.null native_raster_map null=0` will convert NULL cells to 0.

Anyhow, problems appear when working under Windows, as discussed in https://trac.osgeo.org/grass/ticket/3857#comment:1.

nyalldawson:
> Anyhow, problems appear when working under Windows, as discussed in https://trac.osgeo.org/grass/ticket/3857#comment:1.

Given that there's known issues with defaulting to true which can be fixed my moving to false, I think we should play things safe and default to the slower, safer method. (I think this qualifies as a bug fix and should be backported)

pmav99:
I force pushed after changing the default value + editing the commit message

Q: WRT the backport, can the bot handle this, or should I do it myself?

nyalldawson (approved):

python/plugins/processing/tests/testdata/grass7_algorithms_raster_tests.yaml
```diff
@@ -1109,13 +1109,13 @@ tests:
         type: raster
     results:
       blue:
-        hash: 7fab6db4c0517dbd4a3d61181874ad0544383896ba5f4292c9418be2
+        hash: 144242915491497a375db9818189a9ed228b678afe81afdf52f88e5f
```
nyalldawson:
```suggestion
        hash: 
        - 144242915491497a375db9818189a9ed228b678afe81afdf52f88e5f
        - 6ffe41469e35fcca748f4947da527a7163cd06fb3b9a916f35083d12
```

nyalldawson:
Superseded by #33578 (updates tests)

PR #32437, opened 2019-10-26
Fix compilation with QT 5.9.0

rouault:


src/core/expression/qgsexpressionfunction.cpp
```diff
@@ -5063,6 +5064,7 @@ static QVariant fcnGenericHash( const QVariantList &values, const QgsExpressionC
   {
     hash = fcnHash( str, QCryptographicHash::Keccak_512 );
   }
+#endif
```
```suggestion
#endif
  else
  {
    parent->setEvalErrorString( QObject::tr( "Hash method %1 is not available on this system." ).arg( str ) );
  }
```

rouault: indeed. done

PR #32448, opened 2019-10-28
Replace QGIS raster resampling with GDAL calls

nyalldawson:
This PR replaces QGIS inbuilt raster resampling handling with GDAL's raster resampling code.

Why? Because

1. QGIS ones are sloooow. Horribly slow and optimized (vs which are heavily optimised)
2. QGIS ones are poorly tested and effectively unmaintained, so we benefit by using upstream maintained ones
3. GDAL exposes MANY more resampling techniques we could easily expose using this, e.g. Lanczos, Average, Mode, Max, Min, ... etc


src/core/qgsgdalutils.cpp
```diff
@@ -90,6 +97,65 @@ gdal::dataset_unique_ptr QgsGdalUtils::createSingleBandTiffDataset( QString file
...
+  blue.resize( sizeof( float ) * count );
+  alpha.resize( sizeof( float ) * count );
+
+  float *rData = reinterpret_cast<float *>( red.data() );
+  float *gData = reinterpret_cast<float *>( green.data() );
+  float *bData = reinterpret_cast<float *>( blue.data() );
+  float *aData = reinterpret_cast<float *>( alpha.data() );
+
+  gdal::dataset_unique_ptr hSrcDS( QgsGdalUtils::createMultiBandMemoryDataset( GDT_Float32, 4, extent, image.width(), image.height(), crs ) );
+  for ( qgssize i = 0; i < count; ++i )
```
nyalldawson: @rouault I'm not super-happy with this approach, but I can't see any other way to populate the raster bands in the memory dataset from a QImage. Have you got any ideas for a more optimised alternative?

src/core/qgsgdalutils.cpp
```diff
@@ -120,6 +186,132 @@ void QgsGdalUtils::resampleSingleBandRaster( GDALDatasetH hSrcDS, GDALDatasetH h
...
+    CPLFree( bData );
+    CPLFree( aData );
+    return QImage();
+  }
+
+  float *rIn = rData;
+  float *gIn = gData;
+  float *bIn = bData;
+  float *aIn = aData;
+  for ( qgssize i = 0; i < count; ++i )
```
nyalldawson: This is another iteration I'd love to optimise, but I can't think of any approach to avoid this either

nyalldawson:
(Unscientific test, but on a 10800x5400 source raster the cubic resampler is about twice as fast on a debug build now)

nirvn:
Unscientific tests are the best! :wink: Resampling improvements in QGIS are most welcome.

src/core/qgsgdalutils.cpp
```diff
@@ -90,6 +97,65 @@ gdal::dataset_unique_ptr QgsGdalUtils::createSingleBandTiffDataset( QString file
...
 
+gdal::dataset_unique_ptr QgsGdalUtils::imageToMemoryDataset( const QImage &image, QgsRectangle extent, const QgsCoordinateReferenceSystem &crs )
+{
+  if ( image.isNull() )
+    return nullptr;
+  QByteArray red, green, blue, alpha;
+
+  const QRgb *rgb = reinterpret_cast<const QRgb *>( image.constBits() );
+
+  // potentially an overflow here -- that's on Qt!
```
rouault: is this comment true ? I guess that on 32 bit builds, Qt won't let you allocate an image with more than 4 billion pixels

src/core/qgsgdalutils.cpp
rouault: The GDAL MEM driver allows to instanciate a dataset with existing pixel arrays by passing a DATAPOINTER=address. The recommended workflow is:
- GDALCreate() with 0 bands
- use GDALAddBand() as many times as you have bands with DATAPOINTER, PIXELOFFSET and LINEOFFSET options  (see https://gdal.org/drivers/raster/mem.html)

That said, it seems that you want to resampling to happen on Float32. If you follow my above method, you'll get a Byte datase, and thus resampling will occur fundamentally on Byte, even if you specify GDT_Float32 in the RasterIO calls (the promotion of Byte to Float32 will be done after resampling)

If you don't go to that road, a slight optimization to your existing code can save the need of the temporary rData, gData, etc... arrays by directly piping the rgb array to GDALRasterIO() with calls like
```
GDALRasterIO( GDALGetRasterBand( hSrcDS.get(), 1 ), GF_Write, 0, 0, image.width(), image.height(), (GByte*)rgb, image.width(), image.height(), GDT_Byte, sizeof(QRgb), 0 );

GDALRasterIO( GDALGetRasterBand( hSrcDS.get(), 2 ), GF_Write, 0, 0, image.width(), image.height(), ((GByte*)rgb)+1, image.width(), image.height(), GDT_Byte, sizeof(QRgb), 0 );
```
The changes are: the passed pointer, the data type, and the pixel stride
You need to find the proper byte offsets of the R,G,B,A components in the QRgb structure (this would be endianness dependent if you care about QGIS on big-endian hosts)

src/core/qgsgdalutils.cpp
```diff
@@ -120,6 +186,132 @@ void QgsGdalUtils::resampleSingleBandRaster( GDALDatasetH hSrcDS, GDALDatasetH h
   GDALDestroyGenImgProjTransformer( psWarpOptions->pTransformerArg );
 }
 
+gdal::dataset_unique_ptr QgsGdalUtils::resampleImage( const QImage &image, const QgsRectangle &extent, QSize outputSize, GDALResampleAlg resampleAlg )
+{
+  gdal::dataset_unique_ptr srcDS = QgsGdalUtils::imageToMemoryDataset( image, extent, QgsCoordinateReferenceSystem() );
```
rouault: This is rather inefficient to use the warp algorithm to do just resampling without reprojection (this applies to the above resampleSingleBandRaster() method). You could get even better results with using GDALRasterIOEx() with the GDALRasterIOExtraArg.eResampleAlg

src/core/qgsgdalutils.cpp
```diff
@@ -120,6 +186,132 @@ void QgsGdalUtils::resampleSingleBandRaster( GDALDatasetH hSrcDS, GDALDatasetH h
...
+
+  GDALRasterBandH rBand = GDALGetRasterBand( srcDS.get(), 1 );
+  GDALRasterBandH gBand = GDALGetRasterBand( srcDS.get(), 2 );
+  GDALRasterBandH bBand = GDALGetRasterBand( srcDS.get(), 3 );
+  GDALRasterBandH aBand = GDALGetRasterBand( srcDS.get(), 4 );
+
+  const qgssize count = static_cast< qgssize >( outputSize.width() ) * outputSize.height();
+
+  float *rData = ( float * ) CPLMalloc( count * sizeof( float ) );
+  CPLErr err = GDALRasterIO( rBand, GF_Read, 0, 0, outputSize.width(), outputSize.height(),
```
rouault: same as above (in reversed direction): you could likely GDALRasterIO() directly into QRgb* rgb

rouault:
> (Unscientific test, but on a 10800x5400 source raster the cubic resampler is about twice as fast on a debug build now)

With a debug or optimized GDAL build ? If you want to compare apples to apples, you should compare either QGIS debug+GDAL debug before and after your changes, or more relevant, QGIS optimized+GDAL optmized before and after. GDAL has SSE/SSE2 optimized paths and the difference between debug and release is abysmal.

src/core/qgsgdalutils.cpp
rouault: If your workflow is QImage (Byte data type) to QImage (Byte data type) then you definitely don't need to use a GDAL Float32 dataset. Non-nearest resampling in GDAL on Byte data type will use internally float/double and will output the rounded result. Using Float32 doesn't bring anything, except a useless initial Byte->Float32 conversion, slower Float32 resampling  (resampling paths in GDAL are especially optimized for Byte and UInt16 data types), and useless Float32 -> Byte final step

src/core/qgsgdalutils.cpp
roya0045: Is there a way to perform operations on arrays instead of iterating? I'm not sure if array librairies in c++ exist but what I have in mind is numpy. But then again I'm not sure how to leverage vectorization in c++.

nyalldawson:
@rouault thanks for your valuable comments! Are you able to see what's wrong in https://github.com/qgis/QGIS/pull/32448/files#diff-1a60a264960e5fcbc65b73c67ab775ddR236 ? valgrind is reporting a write 0 bytes after the block size allocated, but I can't see what I'm doing wrong...

src/core/qgsgdalutils.cpp
```diff
@@ -120,6 +171,103 @@ void QgsGdalUtils::resampleSingleBandRaster( GDALDatasetH hSrcDS, GDALDatasetH h
...
+
+  GDALRasterIOExtraArg extra;
+  INIT_RASTERIO_EXTRA_ARG( extra );
+  extra.eResampleAlg = resampleAlg;
+
+  QImage res( outputSize, QImage::Format_ARGB32 ); // premultiplied??
+  if ( res.isNull() )
+    return QImage();
+
+  QRgb *rgb = reinterpret_cast<QRgb *>( res.bits() );
```
rouault: you should cast to GByte*, not QRgb*, for your below +1, +2, +3 to be correct

src/core/qgsgdalutils.cpp
```diff
@@ -97,61 +97,46 @@ gdal::dataset_unique_ptr QgsGdalUtils::createSingleBandTiffDataset( QString file
   return hDstDS;
 }
 
-gdal::dataset_unique_ptr QgsGdalUtils::imageToMemoryDataset( const QImage &image, QgsRectangle extent, const QgsCoordinateReferenceSystem &crs )
+gdal::dataset_unique_ptr QgsGdalUtils::imageToMemoryDataset( const QImage &image )
```
rouault: I haven't check at the comments of the method, but there should be a mention that the exact QImage objet passed to this method (or at the very list its pixel array) should be kept alive (and not modified, if QImage is mutable) while the GDALDataset is in use, otherwise it will point to garbage

nyalldawson: Good point. I'll probably make all these private anyway...

src/core/qgsgdalutils.cpp
nyalldawson: Thank you!

nyalldawson:
Success! So, on release builds rendering with the cubic resampling is 3x faster (that's of the total rendering time, not just the resampling portion only), i.e. a very noticable difference. Bilinear/average is a similar rendering time (likely because the QImage scaler used before is also heavily optimised, just like GDAL). 

Aside from the speed boost and ability to easily expose additional resampling methods, we also fix these kind of rendering issues with the cubic resampler and nodata pixels:

Before:
image

After:
image

nyalldawson:
@rouault is this ok to merge?

nyalldawson:
Thanks for the review and assistance @rouault ! (and for making a great resampler!)

nirvn:
@nyalldawson , I've been using this PR for the last few hours while working on resampled rasters, all good.

PR #32450, opened 2019-10-28
add QGIS version in model exported as Python

Gustry:
## Description

A model created in QGIS 3.8, might not be compatible with an older version of QGIS:
* The API might be improved
* Algorithm not available, etc

I just tried a model from QGIS 3.8 in QGIS 3.6 and I got a Python exception.

This PR is adding the QGIS version which has been used to export the model as Python script, in the docstring metadata.
Ideally, we should add the QGIS version which as been used to create the model too. (It needs to be added in the `.model3` file)

## Checklist

### DIFF
--- a/src/core/processing/models/qgsprocessingmodelalgorithm.cpp
+++ b/src/core/processing/models/qgsprocessingmodelalgorithm.cpp
@@ -19,6 +19,7 @@
 #include "qgsprocessingregistry.h"
 #include "qgsprocessingfeedback.h"
 #include "qgsprocessingutils.h"
+#include "qgis.h"
 #include "qgsxmlutils.h"
 #include "qgsexception.h"
 #include "qgsvectorlayer.h"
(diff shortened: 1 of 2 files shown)


src/core/processing/models/qgsprocessingmodelalgorithm.cpp
```diff
@@ -424,7 +434,7 @@ QStringList QgsProcessingModelAlgorithm::asPythonCode( const QgsProcessing::Pyth
           importLines << importString;
       }
 
-      lines << QStringLiteral( "import processing" );
+      lines << QStringLiteral( "from qgis import processing" );
```
nyalldawson: I wasn't planning on changing this, because it causes breakage with older qgis versions which we can easily avoid

Gustry: Yes, I understand. I was doing what has been done in the script template:
https://github.com/qgis/QGIS/blob/master/python/plugins/processing/script/ScriptTemplate.py#L21

nyalldawson: Hm, we should probably revisit that...

PR #32453, opened 2019-10-28
Fix waitforfinished

troopa81:
## Description

It could lead to crashes (If you access some shared memory data believing the task has finished to run) so I will be in favor to backporting this to 3.4.

## Checklist

### DIFF
--- a/src/core/qgstaskmanager.cpp
+++ b/src/core/qgstaskmanager.cpp
@@ -57,8 +57,8 @@ qint64 QgsTask::elapsedTime() const
 
 void QgsTask::start()
 {
-  mNotStartedMutex.unlock();
   mNotFinishedMutex.lock();
+  mNotStartedMutex.unlock();
   mStartCount++;
   Q_ASSERT( mStartCount == 1 );
 
--- a/src/core/qgstaskmanager.cpp
+++ b/src/core/qgstaskmanager.cpp
@@ -29,6 +29,7 @@ QgsTask::QgsTask( const QString &name, Flags flags )
   : mFlags( flags )
   , mDescription( name )
 {
+  mNotStartedMutex.lock();
 }
 
 QgsTask::~QgsTask()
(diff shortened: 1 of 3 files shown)


tests/src/python/test_qgslayoutlegend.py
```diff
@@ -408,7 +404,6 @@ def testSymbolExpressions(self):
         label1 = legendnodes[0].evaluateLabel()
         label2 = legendnodes[1].evaluateLabel()
         label3 = legendnodes[2].evaluateLabel()
-        count.waitForFinished()
```
nyalldawson: Why these removals?

troopa81: There is already one waitForFinished before the first line modified, counterTask and count are the same object. I'm not sure why there was 2 waitForFinished calls in the first place (maybe to maximize the chance to actually wait that the task start?!), but the only one call is needed IMHO.

@roya0045 do you confirm?

roya0045: Yes, these were extra ways in an attempt to get the counter to finish before testing for the condition. If the waitforfinish actually works properly then these extra measures are not necessary anymore and I believe this should be a good implicit test for this fix.

src/core/qgstaskmanager.cpp
```diff
@@ -55,6 +57,7 @@ qint64 QgsTask::elapsedTime() const
 
 void QgsTask::start()
 {
+  mNotStartedMutex.unlock();
```
nyalldawson: I think you should flip this order -- there's a chance wait for finished might return in between these statements otherwise

troopa81: Yes, you right. Done

nyalldawson:
I've removed the 3.4 backport -- I don't think this is safe to backport

PR #32456, opened 2019-10-28
[FEATURE][symbology] Add density-based point count for the random marker fill

nirvn:
## Description

This PR adds a density-based point count method to @nyalldawson 's brand new random marker fill. 

GIF time:
Peek 2019-10-28 16-00

Edit: i should probably expand a little bit more here. The benefit of this new method is that contrary to an absolute count, the fill density will remain the same on different scale / zoom levels (as demonstrated in the GIF above). 

## Checklist


nirvn:
@nyalldawson , I wasn't so sure of the newly-added function wording, feel free to suggest better alternatives if you can think of any.

I'll add tests once the code is reviewed.

src/core/symbology/qgsfillsymbollayer.h
```diff
@@ -1734,13 +1734,19 @@ class CORE_EXPORT QgsRandomMarkerFillSymbolLayer : public QgsFillSymbolLayer
 {
   public:
 
+    enum CountMethod
+    {
+      AbsoluteCount, //!< The point count is used as an absolute count of markers
+      DensityBasedCount, //!< The point count is part of a marker density count
+    };
```
roya0045: Just a suggestion: as count might be misleading, what about Threshold for the enum name and something like MaximumDensity for the second enum variable?

nirvn: I don't really see threshold as a better wording here. But that might just be the fact that I'm not a native English speaker. Let's see what Nyall comes up with.

src/core/symbology/qgsfillsymbollayer.cpp
```diff
@@ -4047,12 +4056,26 @@ void QgsRandomMarkerFillSymbolLayer::render( QgsRenderContext &context, const QV
...
 
+
   int count = mPointCount;
   if ( mDataDefinedProperties.isActive( QgsSymbolLayer::PropertyPointCount ) )
   {
     context.expressionContext().setOriginalValueVariable( count );
     count = mDataDefinedProperties.valueAsInt( QgsSymbolLayer::PropertyPointCount, context.expressionContext(), count );
   }
+
+  if ( mCountMethod == DensityBasedCount )
```
nyalldawson: if -> switch

src/core/symbology/qgsfillsymbollayer.h
nyalldawson: I don't think threshold works either -- it suggests it's going to render some number up to the count, yet it will **always** render exactly that number of points

src/core/symbology/qgsfillsymbollayer.h
```diff
@@ -1734,13 +1734,20 @@ class CORE_EXPORT QgsRandomMarkerFillSymbolLayer : public QgsFillSymbolLayer
...
+
     /**
      * Constructor for QgsRandomMarkerFillSymbolLayer, with the specified \a pointCount.
      *
      * Optionally a specific random number \a seed can be used when generating points. A \a seed of 0 indicates that
      * a truly random sequence will be used on every rendering, causing points to appear in different locations with every map refresh.
      */
-    QgsRandomMarkerFillSymbolLayer( int pointCount = 10, unsigned long seed = 0 );
+    QgsRandomMarkerFillSymbolLayer( int pointCount = 10, unsigned long seed = 0, CountMethod method = AbsoluteCount, double densityArea = 250.0 );
```
nyalldawson: I'd suggest we move seed to the end, as it's least likely to be set

src/core/symbology/qgsfillsymbollayer.h
```diff
@@ -1815,6 +1822,64 @@ class CORE_EXPORT QgsRandomMarkerFillSymbolLayer : public QgsFillSymbolLayer
...
 
+    /**
+     * Returns the count method used to randomly fill the polygon.
+     *
+     * \see setCountMethod()
+     */
+    CountMethod countMethod() const;
+
+    /**
+     * Sets the coiunt \a method used to randomly fill the polygon.
```
nyalldawson:
```suggestion
     * Sets the count \a method used to randomly fill the polygon.
```

src/core/symbology/qgsfillsymbollayer.h
```diff
@@ -1815,6 +1822,64 @@ class CORE_EXPORT QgsRandomMarkerFillSymbolLayer : public QgsFillSymbolLayer
...
+     *
+     * \see countMethod()
+     */
+    void setCountMethod( CountMethod method );
+
+    /**
+     * Returns the density area used to count the number of points to randomly fill the polygon.
```
nyalldawson: "Only used when the count method is ... etc"
"Units are specified via ..."

src/gui/symbology/qgssymbollayerwidget.cpp
```diff
@@ -4119,11 +4119,18 @@ QgsRandomMarkerFillSymbolLayerWidget::QgsRandomMarkerFillSymbolLayerWidget( QgsV
   QgsSymbolLayerWidget( parent, vl )
 {
   setupUi( this );
+
+  mCountMethodComboBox->addItem( tr( "Absolute count" ), QgsRandomMarkerFillSymbolLayer::AbsoluteCount );
```
nyalldawson:
```suggestion
  mCountMethodComboBox->addItem( tr( "Absolute Count" ), QgsRandomMarkerFillSymbolLayer::AbsoluteCount );
```

src/gui/symbology/qgssymbollayerwidget.cpp
```diff
@@ -4119,11 +4119,18 @@ QgsRandomMarkerFillSymbolLayerWidget::QgsRandomMarkerFillSymbolLayerWidget( QgsV
...
+  mCountMethodComboBox->addItem( tr( "Density-based count" ), QgsRandomMarkerFillSymbolLayer::DensityBasedCount );
```
nyalldawson:
```suggestion
  mCountMethodComboBox->addItem( tr( "Density-based Count" ), QgsRandomMarkerFillSymbolLayer::DensityBasedCount );
```

src/core/symbology/qgsfillsymbollayer.cpp
```diff
@@ -4047,12 +4056,26 @@ void QgsRandomMarkerFillSymbolLayer::render( QgsRenderContext &context, const QV
...
+  {
+    double densityArea = mDensityArea;
+    if ( mDataDefinedProperties.isActive( QgsSymbolLayer::PropertyDensityArea ) )
+    {
+      context.expressionContext().setOriginalValueVariable( densityArea );
+      densityArea = mDataDefinedProperties.valueAsDouble( QgsSymbolLayer::PropertyPointCount, context.expressionContext(), densityArea );
+    }
+    densityArea = context.convertToPainterUnits( densityArea, mDensityAreaUnit, mDensityAreaUnitScale );
```
nyalldawson: This method converts distances, not areas. I think you'll need to adapt this accordingly.

src/core/symbology/qgssymbollayer.h
```diff
@@ -182,6 +182,7 @@ class CORE_EXPORT QgsSymbolLayer
       PropertyPointCount, //!< Point count
       PropertyRandomSeed, //!< Random number seed
       PropertyClipPoints, //!< Whether markers should be clipped to polygon boundaries
+      PropertyDensityArea, //<! Density area
```
nyalldawson: You've missed the definition for this in the .cpp file

src/core/symbology/qgsfillsymbollayer.cpp
nirvn: Hmm, so taking the root square value, then converting, then taking the square value of that?

nirvn:
@nyalldawson , your comments have been addressed, and a test added.

src/core/qgsrendercontext.cpp
```diff
@@ -438,6 +438,10 @@ double QgsRenderContext::convertFromMapUnits( double sizeInMapUnits, QgsUnitType
 
 double QgsRenderContext::convertMetersToMapUnits( double meters ) const
 {
+  QgsDebugMsg( mDistanceArea.sourceCrs().description() );
```
nyalldawson: Drop

src/core/symbology/qgssymbollayer.cpp
```diff
@@ -98,6 +98,7 @@ void QgsSymbolLayer::initPropertyDefinitions()
     { QgsSymbolLayer::PropertyPointCount, QgsPropertyDefinition( "pointCount", QObject::tr( "Point count" ), QgsPropertyDefinition::IntegerPositive, origin )},
     { QgsSymbolLayer::PropertyRandomSeed, QgsPropertyDefinition( "randomSeed", QgsPropertyDefinition::DataTypeNumeric, QObject::tr( "Random number seed" ), QObject::tr( "integer > 0, or 0 for completely random sequence" ), origin )},
     { QgsSymbolLayer::PropertyClipPoints, QgsPropertyDefinition( "clipPoints", QObject::tr( "Clip markers" ), QgsPropertyDefinition::Boolean, origin )},
+    { QgsSymbolLayer::PropertyClipPoints, QgsPropertyDefinition( "densityArea", QObject::tr( "Density area" ), QgsPropertyDefinition::IntegerPositive, origin )},
```
nyalldawson:
```suggestion
    { QgsSymbolLayer::PropertyDensityArea, QgsPropertyDefinition( "densityArea", QObject::tr( "Density area" ), QgsPropertyDefinition::DoublePositive, origin )},
```

src/gui/symbology/qgssymbollayerwidget.cpp
```diff
@@ -4147,7 +4159,21 @@ void QgsRandomMarkerFillSymbolLayerWidget::setSymbolLayer( QgsSymbolLayer *layer
   whileBlocking( mSeedSpinBox )->setValue( mLayer->seed() );
   whileBlocking( mClipPointsCheckBox )->setChecked( mLayer->clipPoints() );
 
+  const bool showDensityBasedCountWidgets = mLayer->countMethod() == QgsRandomMarkerFillSymbolLayer::DensityBasedCount;
```
nyalldawson: Ideally use a switch for this logic

nyalldawson (commented):
Nearly there!

nyalldawson (approved):

PR #32463, opened 2019-10-28
[FEATURE][needs-docs][processing] add gdal_viewshed algorithm

alexbruy:
## Description

Expose new `gdal_viewshed` tool via Processing toolbox. Note: this requires GDAL >= 3.1.

## Checklist

### DIFF
--- a/python/plugins/processing/algs/gdal/viewshed.py
+++ b/python/plugins/processing/algs/gdal/viewshed.py
@@ -1,0 +1,158 @@
+# -*- coding: utf-8 -*-
+
+"""
+***************************************************************************
+    viewshed.py
+    ---------------------
+    Date                 : October 2019
+    Copyright            : (C) 2019 by Alexander Bruy
+    Email                : alexander dot bruy at gmail dot com
+***************************************************************************
+*                                                                         *
+*   This program is free software; you can redistribute it and/or modify  *
+*   it under the terms of the GNU General Public License as published by  *
+*   the Free Software Foundation; either version 2 of the License, or     *
+*   (at your option) any later version.                                   *
+*                                                                         *
+***************************************************************************
+"""
+
+__author__ = 'Alexander Bruy'
+__date__ = 'October 2019'
+__copyright__ = '(C) 2019, Alexander Bruy'
+
+import os
+
+from qgis.PyQt.QtGui import QIcon
+
+from qgis.core import (QgsRasterFileWriter,
+                       QgsProcessingException,
+                       QgsProcessingParameterDefinition,
+                       QgsProcessingParameterRasterLayer,
+                       QgsProcessingParameterBand,
+                       QgsProcessingParameterPoint,
+                       QgsProcessingParameterNumber,
+                       QgsProcessingParameterString,
+                       QgsProcessingParameterRasterDestination)
+from processing.algs.gdal.GdalAlgorithm import GdalAlgorithm
+from processing.algs.gdal.GdalUtils import GdalUtils
+
+from processing.tools.system import isWindows
+
+pluginPath = os.path.split(os.path.split(os.path.dirname(__file__))[0])[0]
+
+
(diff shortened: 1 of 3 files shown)


python/plugins/processing/algs/gdal/viewshed.py
```diff
@@ -0,0 +1,158 @@
...
+    def initAlgorithm(self, config=None):
+        self.addParameter(QgsProcessingParameterRasterLayer(self.INPUT,
+                                                            self.tr('Input layer')))
+        self.addParameter(QgsProcessingParameterBand(self.BAND,
+                                                     self.tr('Band number'),
+                                                     1,
+                                                     parentLayerParameterName=self.INPUT))
+        self.addParameter(QgsProcessingParameterPoint(self.OBSERVER,
+                                                      self.tr('Observer location')))
+        self.addParameter(QgsProcessingParameterNumber(self.OBSERVER_HEIGHT,
```
nyalldawson: What units are these?

python/plugins/processing/algs/gdal/viewshed.py
```diff
@@ -0,0 +1,158 @@
...
+                                                       self.tr('Observer height'),
+                                                       type=QgsProcessingParameterNumber.Double,
+                                                       minValue=0.0,
+                                                       defaultValue=1.0))
+        self.addParameter(QgsProcessingParameterNumber(self.TARGET_HEIGHT,
+                                                       self.tr('Target height'),
+                                                       type=QgsProcessingParameterNumber.Double,
+                                                       minValue=0.0,
+                                                       defaultValue=1.0))
+        self.addParameter(QgsProcessingParameterNumber(self.MAX_DISTANCE,
```
nyalldawson: Same with this - what unit is it? Maybe a distance parameter is more suitable here

alexbruy: Changed to distance parameter

python/plugins/processing/algs/gdal/viewshed.py
alexbruy: If I read code correctly, it is this DEM units.

nyalldawson: I think it would be worth adding that to the parameter description?

rouault:
I would just want to mention the gdal_viewshed is super fresh and would need more work in GDAL itself on the testing and documentation sides: https://github.com/OSGeo/gdal/issues/1917 . Contributions welcome there too.

python/plugins/processing/algs/gdal/viewshed.py
alexbruy: Done. Although I'd like to see our users read docs before using any tool and not blindly press buttons.

alexbruy:
I'd like to merge this if there are no objections.

nyalldawson (approved):

imiara:
Hi,
I've just installed QGIS nightly from OSGeo4W (that was built against gdal 3.1.0dev)
I have viewshed.py but I don't see the gdal-viewshed processing algorithm in the Processing Toolbox list.
Am I missing something here?

gioman:
>  Am I missing something here?

try qgis master.

imiara:
> > Am I missing something here?
> 
> try qgis master.

Thanks for the quick reply!
I've installed the package named qgis-dev (3.11.0-28) from OSGeo4W (windows x64)
it has code revision https://github.com/qgis/QGIS/commit/4c234f21b4
which was committed 21 hours ago to master.
Did you advise I'll use something else?

pigreco:
I can't find the algorithm in the processing tools :-(

image

PedroVenancio:
This is why gdal_viewshed is not shown in Processing:

http://osgeo-org.1560.x6.nabble.com/Wrong-GDAL-VersionInfo-td5437195.html

PR #32487, opened 2019-10-29
[feature] Value relation restore missing layers from DBs

elpaso:
Checks for missing layers in value relation widget and loads them automatically (if possible), also warns the user if the layer could not be found.

The API is generic enough to be used by other widgets too, if they happen to require one or more other layers to work correctly.

The use case we address here is when a layer stored in PG or GPKG or SpatiaLite is configured to use a value-relation widget but the required layer was not already loaded in the project (or it was loaded but it was moved to another schema), in that case the weak resolution won't find the layer and the widget will not work. With this implementation, the layer is first searched in the loaded layers (matching the provider name and the layer name) and if it is not found it will be searched in the same connection as the parent layer (first in the same schema, then in the other schemas) and loaded if found.

The implementation also works when applying a style that uses a value relation widget to an existing layer.

Funded by **ARPA Piemonte**

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -2070,7 +2070,7 @@ void QgisApp::checkVectorLayerDependencies( QgsVectorLayer *vl )
         }
         else
         {
-          messageBar()->pushSuccess( tr( "Missing layer form dependency" ), tr( "Layer form dependency '%2' required by '%1' was automatically loaded." )
+          messageBar()->pushSuccess( tr( "Missing layer form dependency" ), tr( "Layer dependency '%2' required by '%1' was automatically loaded." )
                                      .arg( vl->name() )
                                      .arg( dependency.name ) );
         }
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -2041,7 +2041,6 @@ void QgisApp::checkVectorLayerDependencies( QgsVectorLayer *vl )
                 std::unique_ptr< QgsVectorLayer > newVl = qgis::make_unique< QgsVectorLayer >( layerUri, dependency.name, providerName );
                 if ( newVl->isValid() )
                 {
-                  connect( newVl.get(), &QgsVectorLayer::styleLoaded, this, &QgisApp::vectorLayerStyleLoaded );
                   QgsProject::instance()->addMapLayer( newVl.release() );
                   return true;
                 }
@@ -5189,12 +5188,6 @@ bool QgisApp::addVectorLayersPrivate( const QStringList &layerQStringList, const
   QgsProject::instance()->addMapLayers( layersToAdd );
   for ( QgsMapLayer *l : qgis::as_const( layersToAdd ) )
   {
-    // Connect style changed signal to run dependencies checks
-    QgsVectorLayer *vl { qobject_cast< QgsVectorLayer * >( l ) };
-    if ( vl )
-    {
-      connect( vl, &QgsVectorLayer::styleLoaded, this, &QgisApp::vectorLayerStyleLoaded );
-    }
     bool ok;
     l->loadDefaultStyle( ok );
     l->loadDefaultMetadata( ok );
@@ -6472,15 +6465,13 @@ bool QgisApp::addProject( const QString &projectFile )
 #endif
 
     // Check for missing layer widget dependencies
-    for ( QgsVectorLayer *vl : QgsProject::instance()->layers<QgsVectorLayer *>( ) )
+    const auto constVLayers { QgsProject::instance()->layers<QgsVectorLayer *>( ) };
+    for ( QgsVectorLayer *vl : constVLayers )
     {
       if ( vl->isValid() )
       {
         checkVectorLayerDependencies( vl );
       }
-      // Connect style changed signal to make sure the check is run again
-      // if a new style is loaded
-      connect( vl, &QgsVectorLayer::styleLoaded, this, &QgisApp::vectorLayerStyleLoaded );
     }
 
     emit projectRead(); // let plug-ins know that we've read in a new
@@ -11554,8 +11545,6 @@ QgsVectorLayer *QgisApp::addVectorLayerPrivate( const QString &vectorLayerPath, 
 
       askUserForDatumTransform( layer->crs(), QgsProject::instance()->crs(), layer );
 
-      // Connect style changed signal to run dependencies checks
-      connect( layer, &QgsVectorLayer::styleLoaded, this, &QgisApp::vectorLayerStyleLoaded );
       bool ok;
       layer->loadDefaultStyle( ok );
       layer->loadDefaultMetadata( ok );
@@ -12728,6 +12717,7 @@ void QgisApp::layersWereAdded( const QList<QgsMapLayer *> &layers )
       connect( vlayer, &QgsVectorLayer::editingStopped, this, &QgisApp::layerEditStateChanged );
       connect( vlayer, &QgsVectorLayer::readOnlyChanged, this, &QgisApp::layerEditStateChanged );
       connect( vlayer, &QgsVectorLayer::raiseError, this, &QgisApp::onLayerError );
+      connect( vlayer, &QgsVectorLayer::styleLoaded, this, &QgisApp::vectorLayerStyleLoaded );
 
       provider = vProvider;
     }
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -2031,7 +2031,7 @@ void QgisApp::checkVectorLayerDependencies( QgsVectorLayer *vl )
             }
 
             // Helper to find layers in connections
-            auto layerFinder = [ &conn, &dependency, &providerName, this ]( const QString & tableSchema, const QString & tableName ) -> bool
+            auto layerFinder = [ &conn, &dependency, &providerName ]( const QString & tableSchema, const QString & tableName ) -> bool
             {
               // First try the current schema (or no schema if it's not supported from the provider)
               try
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -1982,45 +1982,71 @@ void QgisApp::checkVectorLayerDependencies( QgsVectorLayer *vl )
       // try to aggressively resolve the broken dependencies
       bool loaded = false;
       const QString providerName { vl->dataProvider()->name() };
-      QgsProviderMetadata *md { QgsProviderRegistry::instance()->providerMetadata( providerName ) };
-      if ( md )
+      QgsProviderMetadata *providerMetadata { QgsProviderRegistry::instance()->providerMetadata( providerName ) };
+      if ( providerMetadata )
       {
-        std::unique_ptr< QgsAbstractDatabaseProviderConnection > conn { static_cast<QgsAbstractDatabaseProviderConnection *>( md->createConnection( vl->dataProvider()->uri().uri(), {} ) ) };
+        // Retrieve the DB connection (if any)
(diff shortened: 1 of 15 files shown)


src/app/qgisapp.cpp
```diff
@@ -1940,6 +1950,137 @@ QgsMessageBar *QgisApp::visibleMessageBar()
...
 
+QList<QgsVectorLayerRef> QgisApp::findBrokenWidgetDependencies( QgsVectorLayer *vl )
+{
+  QList<QgsVectorLayerRef> brokenDependencies;
+  // Check for missing layer widget dependencies
+  for ( int i = 0; i < vl->fields().count(); i++ )
+  {
+    std::unique_ptr<QgsEditorWidgetWrapper> ww;
+    ww.reset( QgsGui::editorWidgetRegistry()->create( vl, i, nullptr, nullptr ) );
+    for ( QgsVectorLayerRef &dependency : ww->layerDependencies() )
```
m-kuhn: Take a const copy before iterating a container with a range-based for loop or donald trump riding a unicorn singing elton john songs will appear every night in front of your window. And believe me it sounds terrible when he sings elton john.

elpaso: It's not that I like Elton John, but `resolveWeakly` is not const.

m-kuhn: Hmm... So what happens is you get the `layerDependencies`, take a deep copy of it to iterate, then `resolveWeakly()` on each of the items and drop the modified container again, while forwarding the individual items. This kind of works, so maybe you get the same scenario but muted. If you can live with that, I can as well :+1: I think to get rid of unicorn and president as well you'll need to take a copy only of each item in the loop without a detach in the beginning.

src/app/qgisapp.cpp
```diff
@@ -6315,6 +6461,18 @@ bool QgisApp::addProject( const QString &projectFile )
     }
 #endif
 
+    // Check for missing layer widget dependencies
+    for ( QgsVectorLayer *vl : QgsProject::instance()->layers<QgsVectorLayer *>( ) )
```
m-kuhn: No elton john excuses here ;)

src/app/qgisapp.h
```diff
@@ -1700,6 +1701,11 @@ class APP_EXPORT QgisApp : public QMainWindow, private Ui::MainWindow
 
     void onActiveLayerChanged( QgsMapLayer *layer );
 
+    /**
+     * Triggered when a vector layer style has changed, check for widget config layer dependencies
```
```suggestion
     * Emitted when a vector layer style has changed, check for widget config layer dependencies
```

m-kuhn:
Some lovely work in here :+1:

src/app/qgisapp.cpp
```diff
@@ -1940,6 +1950,137 @@ QgsMessageBar *QgisApp::visibleMessageBar()
...
+              {
+                break;
+              }
+            }
+          }
+        }
+      }
+      if ( ! loaded )
+      {
+        const QString msg { tr( "Form for layer '%1' requires layer '%2' to be loaded but '%2' could not be found, please load it manually if possible." )
```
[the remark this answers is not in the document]
elpaso: Yes, for now the broken dependency only comes from a widget configuration, that's why I added "Form", but I'm ok to drop it, I had doubts about it in the first place.

src/core/qgsvectorlayer.cpp
```diff
@@ -4687,10 +4688,12 @@ QString QgsVectorLayer::loadNamedStyle( const QString &theURI, bool &resultFlag,
       QDomDocument myDocument( QStringLiteral( "qgis" ) );
       myDocument.setContent( qml );
       resultFlag = importNamedStyle( myDocument, errorMsg );
-      return QObject::tr( "Loaded from Provider" );
+      returnMessage = QObject::tr( "Loaded from Provider" );
     }
   }
-  return QgsMapLayer::loadNamedStyle( theURI, resultFlag, categories );
+  returnMessage = QgsMapLayer::loadNamedStyle( theURI, resultFlag, categories );
+  emit styleLoaded();
```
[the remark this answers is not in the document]
elpaso: good idea!

src/app/qgisapp.cpp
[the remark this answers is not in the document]
m-kuhn: I just want to be sure we don't forget it when that ever changes, good ideas welcome!

src/core/qgsmaplayer.h
```diff
@@ -1370,6 +1370,13 @@ class CORE_EXPORT QgsMapLayer : public QObject
      */
     void dataSourceChanged();
 
+    /**
+     * Emitted when a style has been loaded
+     * \param categories style categories
+     * \since QGIS 3.12
+     */
+    void styleLoaded( const QgsMapLayer::StyleCategories &categories );
```
nyalldawson:
```suggestion
    void styleLoaded( QgsMapLayer::StyleCategories categories );
```

src/app/qgisapp.cpp
```diff
@@ -660,6 +661,18 @@ void QgisApp::onActiveLayerChanged( QgsMapLayer *layer )
   emit activeLayerChanged( layer );
 }
 
+void QgisApp::vectorLayerStyleLoaded( const QgsMapLayer::StyleCategories &categories )
```
nyalldawson:
```suggestion
void QgisApp::vectorLayerStyleLoaded( QgsMapLayer::StyleCategories categories )
```

src/app/qgisapp.cpp
```diff
@@ -660,6 +661,18 @@ void QgisApp::onActiveLayerChanged( QgsMapLayer *layer )
...
+{
+  if ( categories.testFlag( QgsMapLayer::StyleCategory::Forms ) )
+  {
+    QgsVectorLayer *vl = qobject_cast<QgsVectorLayer *>( sender() );
+    if ( vl && vl->isValid( ) )
```
nyalldawson: Question -- why does this require a valid layer? You could potentially load a style for a layer with a broken path, and still get the dependencies resolved...

src/app/qgisapp.cpp
```diff
@@ -1941,6 +1954,143 @@ QgsMessageBar *QgisApp::visibleMessageBar()
...
+        }
+      }
+    }
+  }
+  return brokenDependencies;
+}
+
+void QgisApp::checkVectorLayerDependencies( QgsVectorLayer *vl )
+{
+  if ( vl && vl->isValid() )
```
nyalldawson: Again, not sure if we need to restrict this to valid layers

src/app/qgisapp.cpp
```diff
@@ -1941,6 +1954,143 @@ QgsMessageBar *QgisApp::visibleMessageBar()
...
+  {
+    const auto constDependencies { findBrokenWidgetDependencies( vl ) };
+    for ( const QgsVectorLayerRef &dependency : constDependencies )
+    {
+      const QgsVectorLayer *depVl = nullptr;
+      // The default resolution logic does not recognize a positive match if the datasource is
```
nyalldawson: This logic should belong in _LayerRef

src/app/qgisapp.cpp
```diff
@@ -5040,6 +5189,12 @@ bool QgisApp::addVectorLayersPrivate( const QStringList &layerQStringList, const
   QgsProject::instance()->addMapLayers( layersToAdd );
   for ( QgsMapLayer *l : qgis::as_const( layersToAdd ) )
   {
+    // Connect style changed signal to run dependencies checks
+    QgsVectorLayer *vl { qobject_cast< QgsVectorLayer * >( l ) };
```
nyalldawson: Instead of implementing this here, I think it would be safer to listen to the project's layer added signals and respond to that. There's many ways to add layers to qgis which don't go through addVectorLayersPrivate

elpaso: Do you mean the connection, right?

src/app/qgisapp.cpp
elpaso: If the layer is invalid I see no point in resolving dependencies, because you will not able to use them anyways. But, if you fix the broken layer later, then the style is applied again and the resolution will be attempted.

src/app/qgisapp.cpp
elpaso: Yeah, I though about it, not sure: do you think at adding another method that `resolveAggresively` ?

But the main reason I wanted it in app and not in core was that I wanted to communicate with the user, what would you suggest? A signal connected to app ?

nyalldawson: I just mean the first part of this branch -- the part where it scans through project layers for ones which match the name and provider

elpaso: Sorry, I don't follow you: `resolveWeakly` does that (but it matches the public source exactly), the only option I see (and that's the one I proposed in my previous comment) is to have a `resolveAggressively` (or whatever you want to name it) that also searches for matching layers by provider AND name. But that won't be sufficient to solve my problem here, because _LayerRef is designed to find layers already defined in the current project while what I'm doing here is searching for layers in the provider that do not exist in the project.

src/app/qgisapp.cpp
```diff
@@ -1941,6 +1954,131 @@ QgsMessageBar *QgisApp::visibleMessageBar()
...
 }
 
+QList<QgsVectorLayerRef> QgisApp::findBrokenWidgetDependencies( QgsVectorLayer *vl )
+{
+  QList<QgsVectorLayerRef> brokenDependencies;
+  // Check for missing layer widget dependencies
+  for ( int i = 0; i < vl->fields().count(); i++ )
+  {
+    std::unique_ptr<QgsEditorWidgetWrapper> ww;
+    ww.reset( QgsGui::editorWidgetRegistry()->create( vl, i, nullptr, nullptr ) );
```
m-kuhn: @elpaso I get a crash (failing assert) since this change here when loading a project (qgep).

image

What I wonder now is if the approach to *create* a disposable widget wrapper here is the right way to go. Or if this kind of handling would be better suited in the factory directly (maybe even in the fieldformatter in core since it's not really GUI related and could be used by qfield or theoretically even server too).

elpaso: Do you mean moving `QgsEditorWidgetWrapper::layerDependencies()` logic into the field formatter?

I'm not sure that it's the right place, the configuration about widgets "dependencies" is stored in the widget configuration and it belongs to the widget.

What exactly are you suggesting?

Isn't actually better to check if the assert can be removed putting some safe guards wherever in the widget logic there is a danger of misbehavior due to vectorLayerTools being nullptr?

Or even easier: create the`QgsAttributeEditorContext` and pass it to the factory?

m-kuhn: My preferred approach would indeed be to move it to the field formatter.

The field formatter is also based on the same configuration (see e.g. https://github.com/qgis/QGIS/blob/master/src/core/fieldformatter/qgsvaluerelationfieldformatter.cpp#L119 ) and lives in core so the logic can be used everywhere. They are created once per application lifetime, so there is no need to create disposable objects.

But, since I'm not completely sure that there are no hidden dependencies involved, the second proposal less risky but also less generic would be to move it to the widget factories. They are also only created once per application lifetime.

Your proposal will be the easiest in terms of work (because it follows the structure set up in this PR), but is also mostly putting the dust under the carpet, to resolve layer dependencies there is no widget (gui element) required, so I don't think creating a throw away widget wrapper for this purpose is the best way to implement it.

elpaso: Ok, I'll look into the field formatter approach.

m-kuhn: Thanks.
Let me know if you run into questions!

PR #32494, opened 2019-10-29
fix gdal calc compression options

### LINKED ISSUE #32266: GDAL raster calculator fails when using compression
QGIS 3.8.3 on Windows 10, GDAL v2.4.1

Using any compression options in the additional creation options leads to a syntax error in the GDAL raster calculator:
`gdal_calc.py: error: no such option: -c`

The problem is in the syntax created by the tool. Instead of --co COMPRESS=..., it will use -co COMPRESS=...

The issue was reproduced running the GDAL command created by the tool in the command line. After changing -co to --co it runs successfully.

I remember being able to change the GDAL/OGR console call (at least on some tools) in prior QGIS versions. This does not seem to be possible anymore.

gioman:
Fixes https://github.com/qgis/QGIS/issues/32266

Not sure is the most elegant way, but it works.

### DIFF
--- a/python/plugins/processing/algs/gdal/gdalcalc.py
+++ b/python/plugins/processing/algs/gdal/gdalcalc.py
@@ -255,7 +255,11 @@ class gdalcalc(GdalAlgorithm):
 
         options = self.parameterAsString(parameters, self.OPTIONS, context)
         if options:
-            arguments.extend(GdalUtils.parseCreationOptions(options))
+            parts = options.split('|')
+            options = []
+            for p in parts:
+                option = '--co ' + p
+                arguments.append(option)
 
         if self.EXTRA in parameters and parameters[self.EXTRA] not in (None, ''):
             extra = self.parameterAsString(parameters, self.EXTRA, context)


luipir:
may I have to review without tests and with some test failing?
BTW if you follow the failed test you can find where to modify current tests to adapt to the new command parameter management

python/plugins/processing/algs/gdal/gdalcalc.py
```diff
@@ -255,7 +255,11 @@ def getConsoleCommands(self, parameters, context, feedback, executing=True):
 
         options = self.parameterAsString(parameters, self.OPTIONS, context)
         if options:
-            arguments.extend(GdalUtils.parseCreationOptions(options))
+            parts = options.split('|')
+            options = []
```
luipir: not used

luipir: and remove previous options get in line 256

alexbruy:
Superseded by #32598

PR #9857, opened 2019-04-24
New Processing Algorithm to generate raster XYZ tiles

marcel-dancak:
## Description
This is a new processing tool for exporting map as XYZ raster tiles. It supports directory structure and MBTiles output format. Tiles can be stored in PNG or JPG image format.

Sponsored by:

- Hansestadt Herford (www.Herford.de)
- "SWK mbH" (https://swk.herford.de/)
- d.b.g. Datenbankgesellschaft mbH (https://datenbankgesellschaft.de/)

PR #32499, opened 2019-10-30
successive ThreadPoolExecutors for all zoom levels

### LINKED ISSUE #32471: race condition affecting the TilesXYZ MBTiles export plugin
[i already reported the bug to original author (https://github.com/lutraconsulting/qgis-xyz-tiles/issues/30), but because the plugin is in the meanwhile a default component of QGIS and the issue could perhaps familiar to the developers of the main application, i try to report it here as well.]

your TilesXYZ MBTiles export plugin (`qgis/python/plugins/processing/algs/qgis/TilesXYZ.py`) included in QGIS 3.10 for linux seems to be plagued by race conditions when it becomes executed in multiple threads.

the issue doesn't appear in case of the DirectoryWriter, but the MBTilesWriter will hardly work at all and produce corrupt files with lots of missing tiles and other defects...

as a symptom of this particular issue, you should often see error messages like this:

    ERROR 5: /tmp/test7.mbtiles: Access window out of range in RasterIO().  Requested (-569088,-368384) of size 256x256 on raster of 512x512.

the strange negative numbers are a side effect of unexpected changed values of the `self._first_tile` variable resp. colliding zoom factors which are running in parallel and overwrite the expected state one another...

i did try to fix the issue by replacing `self._first_tile` and `self._zoom_ds` with dictionaries indexed by the zoom factor to reduce the observable collisions. this had some positive effect, but it doesn't completely fix the issues. :(

unfortunately i do not know, how to fix this kind of thread lock requirements in case of QGIS plugins in a more adequate manner. 

the versions of my affected debian linux installation:

```
QGIS-Version: 3.10.0-A Coruña
QGIS-Codeversion: 6c816b4204
Qt-Version: 5.11.3
GDAL-Version: 2.4.2
GEOS-Version: 3.8.0-CAPI-1.13.1 
PROJ-Version: Rel. 5.2.0, September 15th, 2018
Verarbeite Algorithmus…
Algorithmus XYZ-Kacheln erzeugen (MBTiles) startet…
Eingabeparameter:
{ 'BACKGROUND_COLOR' : QColor(0, 0, 0, 0), 'DPI' : 96, 'EXTENT' : '1716053.858612443,1721455.9642655044,5951485.136729757,5955838.399322829 [EPSG:3857]', 'METATILESIZE' : 4, 'OUTPUT_FILE' : '/tmp/test7.mbtiles', 'QUALITY' : 75, 'TILE_FORMAT' : 0, 'ZOOM_MAX' : 16, 'ZOOM_MIN' : 12 }

Using 8 CPU Threads:
Pushing all tiles at once: 11 tiles.
Ausführung nach 2.34 Sekunden abgeschlossen
Ergebnisse:
{'OUTPUT_FILE': '/tmp/test7.mbtiles'}
```

mash-graz:
## Description
instead of using only one ThreadPoolExecutor for all MetaTiles, its now handled in a loop and new ThreadPools for the next zoom level will be started when all work for the previous one is already done. i guess, this will fix the issues and interferences described in #32471 and shouldn't be much slower.

Fixes #32471

### DIFF
--- a/python/plugins/processing/algs/qgis/TilesXYZ.py
+++ b/python/plugins/processing/algs/qgis/TilesXYZ.py
@@ -23,6 +23,7 @@ __copyright__ = '(C) 2019 by Lutra Consulting Limited'
 
 import os
 import math
+import re
 from uuid import uuid4
 
 import sqlite3
@@ -177,6 +178,7 @@ class TilesXYZAlgorithmBase(QgisAlgorithm):
                                                        minValue=1,
                                                        maxValue=20,
                                                        defaultValue=4))
+        self.thread_nr_re = re.compile('[0-9]+$') # thread number regex
 
     def prepareAlgorithm(self, parameters, context, feedback):
         project = context.project()
@@ -189,10 +191,11 @@ class TilesXYZAlgorithmBase(QgisAlgorithm):
             return
             # Haven't found a better way to break than to make all the new threads return instantly
 
-        if "Dummy" in threading.current_thread().name: # single thread testing
+        if "Dummy" in threading.current_thread().name or len(self.settingsDictionary) == 1: # single thread testing
             threadSpecificSettings = list(self.settingsDictionary.values())[0]
         else:
-            threadSpecificSettings = self.settingsDictionary[threading.current_thread().name[-1]] # last number only
+            thread_nr = self.thread_nr_re.search(threading.current_thread().name)[0] # terminating number only
+            threadSpecificSettings = self.settingsDictionary[thread_nr]
 
         size = QSize(self.tile_width * metatile.rows(), self.tile_height * metatile.columns())
         extent = QgsRectangle(*metatile.extent())
--- a/python/plugins/processing/algs/qgis/TilesXYZ.py
+++ b/python/plugins/processing/algs/qgis/TilesXYZ.py
@@ -296,9 +296,10 @@ class TilesXYZAlgorithmBase(QgisAlgorithm):
         self.progressThreadLock = threading.Lock()
         if self.maxThreads > 1:
             feedback.pushConsoleInfo(self.tr('Using {max_threads} CPU Threads:').format(max_threads=self.maxThreads))
-            feedback.pushConsoleInfo(self.tr('Pushing all tiles at once: {meta_count} tiles.').format(meta_count=len(allMetatiles)))
-            with ThreadPoolExecutor(max_workers=self.maxThreads) as threadPool:
-                threadPool.map(self.renderSingleMetatile, allMetatiles)
+            for zoom in range(self.min_zoom, self.max_zoom + 1):
+                feedback.pushConsoleInfo(self.tr('Generating tiles for zoom level: {zoom}').format(zoom=zoom))
+                with ThreadPoolExecutor(max_workers=self.maxThreads) as threadPool:
+                    threadPool.map(self.renderSingleMetatile, metatiles_by_zoom[zoom])
         else:
             feedback.pushConsoleInfo(self.tr('Using 1 CPU Thread:'))
             for zoom in range(self.min_zoom, self.max_zoom + 1):


roya0045:
Since you have provided a test to reproduce the error in QGIS, it would be desirable to add this tests to the current tests (I am assuming that the tile function has tests implemented) and add your python script to confirm that your change solve this issue and that no other change will break that function.

mash-graz:
> Since you have provided a test to reproduce the error in QGIS, it would be desirable to add this tests to the current tests ...

the included script to reproduce the issue is just a little helper to make the human test process more comfortable and similar between different testers, but it doesn't represent a solution, which could be simply used as an automatic unit test.

sure, QGIS versions affected by  #32471 will produce corrupt MBTiles output files in most cases, but on any run it will look different -- that's the nature of *data races* -- and in some cases you will not even see the expected symptoms at all.

i think, you have to read and review the actual code to understand the actual issue and my proposed solution.

it somehow tries to combine the intentions of the original plugin author withe the efforts of @isghj5 to make it faster. the original developer wrote some code, that only works well in a strict sequential execution order in a single threaded environment, because it shares and reuses execution states in a rather efficient manner. but the efforts of this other developer, which introduced parallel execution, also make some sense. if we want to combine the advantages of both approaches, we have to shape the actual implementation in a form,  where the **order of tile processing stays fixed** and **tiles of different zoom levels will never be proceeded simultaneously**, otherwise it will crash against fundamental assumptions of the original code. unfortunately this happened by merging PR #31012.

i think, my proposed solution to this issue should be easily comprehensible by just watching the changed code.

[btw.: this bug fix should be backported to 3.10, because the MBTile writer plugin included in the actual release will still produce garbage otherwise.]

mash-graz:
i just added another patch, to fix issues concerning single threaded use of the MBTiles generator.

it should now:

1. handle those cases, where the max. number of threads is set to 1 in the preferences (in this case, ` threading.current_thread().name` isn't terminated by a number, but the string `MainThread')

2. work on machines with many cores and a default thread count >10, which need more digits than the last character.

mash-graz:
@roya0045 @nyalldawson

because i'm not used to contribute to the QGIS project, i don't know, which person i should contact to review and accept this really trival tiny bug fixes. i would be very happy, if they could finally find their way into the manged code base.

if you have any questions/concerns/requirements for improvement, don't hesitate to contact me.

roya0045:
@mash-graz sadly I cannot merge either (I still have PRs in limbo awaiting reviews/merge) so I cannot help you with this. You'll need to wait when a core developer has some time to review your PR.

mash-graz:
@roya0045 
thank's for this info!
it doesn't matter, if it takes some time. i was just a little bit confused, that this PR wasn't even marked with labels etc., while a lot of more recent merge requests got handled in the meanwhile...

mash-graz:
@javicasnov 
thanks for testing this PR and your very competent review of the suggested solution!

although it's a rather trivial fix, it really helpful, if others review the code and test it on different machines, because i'm not used to develop code for QGIS.

this multi threading related issues are always very hard to solve. e.g. the issue concerning the extremely limited number of digitis used for thread indexing, which i fixed in my second commit, usually only affects machines with more then 5 CPU cores, etc.

but your considerations about the performance penalties of the suggested fix are justified!
i was indeed more focused on finding a solution, which eliminates the problematic side effects, but still doesn't modify the original implementation overly. otherwise we would have to change much more and rewrite large parts with multi threading compatibility in mind from the start on and most likely also utilizing the QGIS specific parallelization infrastructure. 

the performance degradation concerning your specific usage scenario -- i.e. using relative large metatiles resp. an uncommon small number of metatile writer invocations -- is indeed an unpleasant consequence of my solution. if i would spot an utterly simple solution to solve this aspect as well, i definitely would improve the code. but given the actual code base, i'm honestly more inclined to accept the performance penalties in this [hopefully rare] practical corner cases as an unavoidable remedy resp. limitation of the original implementation instead of rewriting to much code. i hope, my suggested patches fix at least the most obvious actual bugs, although i also see the potential of further improvements.

mash-graz:
> If you think this pull request should be merged, ...

sure, it should be merged!
but unfortunately i don't see, what i could do, to get this PR finally reviewed and merged. :(

elpaso:
@marcel-dancak is there any hope you can have a look?

marcel-dancak:
I'm not related to Tiles XYZ plugin anymore, and I don't even know about changes that was made after initial version written by me. Also I have very little experience with multi-threading in Python.

Maybe @saberraz could manage review for this PR.

saberraz:
Let me ask my colleagues to look into that.

elpaso:
@saberraz thank you!

mash-graz:
> I'm not related to Tiles XYZ plugin anymore, and I don't even know about changes that was made after initial version written by me. Also I have very little experience with multi-threading in Python.

thanks @marcel-dancak for this feedback!

it's indeed not your original code, which cause this troubles. all the actual issues are in fact related to a later applied commit ("Multithreaded Tile generation with ThreadPoolExecutor" #31012).

maybe the fact, that this code is written in python, could be seen as the reason, why nobody did like to review or accept it until now, because @nyalldawson already wrote in a [very early comment](https://github.com/qgis/QGIS/pull/9857#issuecomment-486515987) about this module:

> My strong preference would be that any new algorithm, especially one as complex as this, is written in c++ and not Python. The Python algorithms have just proved too fragile in the past, and we very much rely on compile time checks to ensure that they don't break. Just something for (everyone) to keep in mind for future work...

and that's indeed a good point.

nevertheless i only concentrated my effort on just repairing the most obvious flaws in the actual code. rewriting it in C++ would be a much more ambitious task. i would perhaps do it just for fun, if [rust](https://www.rust-lang.org/), which is usually a very well choice to prevent this kind of issues, would be an option as well. but QGIS and its very C++ targeted extension API looks like one of the rather rare real world examples, where this alternative also doesn't promise a suitable solution.

elpaso:
> > I'm not related to Tiles XYZ plugin anymore, and I don't even know about changes that was made after initial version written by me. Also I have very little experience with multi-threading in Python.
> 
> thanks @marcel-dancak for this feedback!
> 
> it's indeed not your original code, which cause this troubles. all the actual issues are in fact related to a later applied commit ("Multithreaded Tile generation with ThreadPoolExecutor" #31012).
> 
> maybe the fact, that this code is written in python, could be seen as the reason, why nobody did like to review or accept it until now, because @nyalldawson already wrote in a [very early comment](https://github.com/qgis/QGIS/pull/9857#issuecomment-486515987) about this module:
> 
> > My strong preference would be that any new algorithm, especially one as complex as this, is written in c++ and not Python. The Python algorithms have just proved too fragile in the past, and we very much rely on compile time checks to ensure that they don't break. Just something for (everyone) to keep in mind for future work...
> 
> and that's indeed a good point.
> 
> nevertheless i only concentrated my effort on just repairing the most obvious flaws in the actual code. rewriting it in C++ would be a much more ambitious task. i would perhaps do it just for fun, if [rust](https://www.rust-lang.org/), which is usually a very well choice to prevent this kind of issues, would be an option as well. but QGIS and its very C++ targeted extension API looks like one of the rather rare real world examples, where this alternative also doesn't promise a suitable solution.

I can't speak for Nyall, but my personal experience is that most of the times the choice to review or not to review something is just a matter or time scarcity, priorities and familiarity with a certain area of the code. 

And I wouldn't be totally honest if I wouldn't say that when I spend my free/volunteer time on something there is also a personal bias towards certain people (I'm not referring to this particular PR!), for me personal attitude does matter.

mash-graz:
> I can't speak for Nyall, but my personal experience is that most of the times the choice to review or not to review something is just a matter or time scarcity, priorities and familiarity with a certain area of the code.
>
> And I wouldn't be totally honest if I wouldn't say that when I spend my free/volunteer time on something there is also a personal bias towards certain people (I'm not referring to this particular PR!), for me personal attitude does matter.

that's perfectly comprehensible and acceptable to me! -- thanks!

wonder-sk:
@mash-graz Apologies for taking so long to look at your PR. Your contribution is definitely a good improvement and I am happy to merge it.

The code for multi-threading support looks still quite fragile though... for example a single gdal dataset object is being accessed by multiple threads - I would expect some crashes there (but maybe thanks to Python's big interpreter lock that works fine). So let's merge your fix and if I find some time, I will port this code to C++ ... this will also ease other worries I have such as starting map rendering from a worker thread :)

wonder-sk (approved):

elpaso:
> @mash-graz Apologies for taking so long to look at your PR. Your contribution is definitely a good improvement and I am happy to merge it.
> 
> The code for multi-threading support looks still quite fragile though... for example a single gdal dataset object is being accessed by multiple threads - I would expect some crashes there (but maybe thanks to Python's big interpreter lock that works fine). So let's merge your fix and if I find some time, I will port this code to C++ ... this will also ease other worries I have such as starting map rendering from a worker thread :)

Thank you @wonder-sk for the review! Speaking about your worries, what do you think about this:  https://github.com/qgis/QGIS/pull/32999/files#diff-1016422ba06e23c6fff0c693d70ea7c1R258 ?

mash-graz:
thanks @wonder-sk for accepting the PR!

> The code for multi-threading support looks still quite fragile though... for example a single gdal dataset object is being accessed by multiple threads - I would expect some crashes there (but maybe thanks to Python's big interpreter lock that works fine). So let's merge your fix and if I find some time, I will port this code to C++ ... this will also ease other worries I have such as starting map rendering from a worker thread :)

these are indeed important aspects, which are better solved by someone, who is more familiar with all the QGIS internals. i have to confess, that i'm hardly able to spot the relevant details...

wonder-sk:
@elpaso okay I will comment there not to spread the discussion across multiple PRs :)

PR #32508, opened 2019-10-30
Add QgsProjectViewSettings API, allows for default view to be set for a project

nyalldawson:
Allows a QgsProject created using QGIS api only (i.e., not via the desktop
application) to specify the default extent which should be visible when
that project is opened.

Straight from the dox:

"Sets the default view extent, which should be used as the initial map extent
when this project is opened.

\warning When a project is opened in the QGIS desktop application and saved, individual
map canvases will store their own previous view extent as custom project properties. Reloading
this saved version of the project will trigger the canvases to restore their individual last view
extents. Accordingly, in the QGIS desktop application, this setting only forms a default, initial
view used when the project is opened for the very first time."

My initial use case is for SLYR, where I need to create a QgsProject with the initial view extent set. But I can also see there be compelling use cases for this for QField and Input too.

I'd love to extend QgsProjectViewSettings in future, adding things like a "project area of interest" extent, which sets the limits of the extents you can zoom/pan to in the canvas...


src/core/qgsprojectviewsettings.cpp
```diff
@@ -0,0 +1,148 @@
+/***************************************************************************
+    qgsprojectviewsettings.cpp
+    -----------------------------
+    begin                : November 2018
```
elpaso:
```suggestion
    begin                : October 2019
```

src/core/qgsprojectviewsettings.h
```diff
@@ -0,0 +1,145 @@
...
+     * The scales list consists of a list of scale denominator values, e.g.
+     * 1000 for a 1:1000 scale.
+     *
+     * \see setMapScales()
+     * \see mapScalesChanged()
+     */
+    QVector<double> mapScales() const;
+
+    /**
+     * Sets whether project mapScales() are \a enabled.
```
nirvn: Should we be more verbose here? If project scales are enabled, are they added to the QGIS-wide list of predefined scales or are they overwriting those? I assume it's the latter, worth specifying here.

nirvn (approved):

src/core/qgsprojectviewsettings.h
nyalldawson: resolved

nyalldawson:
Thanks @nirvn!

PR #32513, opened 2019-10-30
[dxf export] initialize PointZ with 0

m-kuhn:
DXF does not know the concept of NaN

This fixes a regression introduced by https://github.com/qgis/QGIS/commit/794a8efc815d6d9fe9101f77c0ffee1128c6ac4b#diff-63bca8c2765a9e460eccb1e6e428dd27R74 where the default values for coordinates were changed from 0.0 to nan.

### DIFF
--- a/src/core/dxf/qgsdxfexport.cpp
+++ b/src/core/dxf/qgsdxfexport.cpp
@@ -580,11 +580,11 @@ void QgsDxfExport::writeHeader( const QString &codepage )
 
   // EXTMIN
   writeGroup( 9, QStringLiteral( "$EXTMIN" ) );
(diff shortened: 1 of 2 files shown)


nyalldawson:
Are you sure that commit is the culprit? It doesn't look like it changed the constructor which accepts an explicit wkb type. Anyway, changes look good to me.

nyalldawson (approved):

m-kuhn:
No, I'm not sure. I just noticed that it works with 3.4 (and apparently 3.8) but is broken in 3.10. So I took the first commit that looked suspicious. Sorry if I blamed the wrong one.

PR #32517, opened 2019-10-30
Fix crash on exit

m-kuhn:
Global static QHash can crash on exit.

Upstream issues have been closed as unreproducible, I still have an issue here.

https://bugreports.qt.io/browse/QTBUG-40813
https://bugreports.qt.io/browse/QTBUG-27005

### DIFF
--- a/src/gui/qgsmapcanvastracer.cpp
+++ b/src/gui/qgsmapcanvastracer.cpp
@@ -26,16 +26,12 @@
 
 #include <QAction>
 
-
-QHash<QgsMapCanvas *, QgsMapCanvasTracer *> QgsMapCanvasTracer::sTracers;
-
-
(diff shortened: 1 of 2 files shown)


nyalldawson (commented):
Looks good! I wonder if it would be easier to understand in future if we just made the tracer a regular member of the canvas though...

nyalldawson:
Also, is this in 3.10?

m-kuhn:
Maybe 3.10, but as long as we don't have a crash report other than a crash in my debug environment I'm hesitating.
I think ownership with the tracer is a very brittle concept at the moment, I touched it as little as possible. What's the idea, a `QgsMapCanvas::tracer()` method that creates on first call?

m-kuhn:
So after recent experiences I am against backporting this. We'd need to move ALL non-POD statics to Q_GLOBAL_STATIC and that's far away from trivial + apparently only affects few systems.

PR #32527, opened 2019-10-30
Fix copy items from value map configuration

### LINKED ISSUE #32256: Value map Widget: add copy & paste functionality
**Allow copy'n'paste from value map widget**
Under "Layer properties" > "Attributes Form" there is the "value map" widget. Sometimes you take a lot of time to fill in a long list of value-descripton pairs in the value map widget. It would be cool if this list could be copied to Excel (or any spreadsheet programm) with copy'n'paste: select all entries in value map widget -> ctr.+c -> go to spreadsheet programm -> ctr.+v -> save the spreadsheet if you like

I use QGIS 3.4.12

m-kuhn:
QAbstractItemView internally uses the KeyPress signal, so in order to
make sure our code takes precedence we need to react to the same signal.
Apparently it's not enough to trust Qt that the KeyReleased is always
triggered after a KeyPress.

Fix #32256

### DIFF
--- a/src/gui/editorwidgets/qgsvaluemapconfigdlg.cpp
+++ b/src/gui/editorwidgets/qgsvaluemapconfigdlg.cpp
@@ -202,12 +202,13 @@ void QgsValueMapConfigDlg::populateComboBox( QComboBox *comboBox, const QVariant
 bool QgsValueMapConfigDlg::eventFilter( QObject *watched, QEvent *event )
 {
   Q_UNUSED( watched )
-  if ( event->type() == QEvent::KeyRelease )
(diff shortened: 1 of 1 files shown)


nyalldawson (commented):
Looks good, can you remove the 3.4 backport?

nyalldawson (approved):

m-kuhn:
I think this deserves backporting to 3.4.
It's a quite good risk to improvement ratio. Having unstable behaviour (even if no "hard" data courruption is involved, it's surprising doing the exactly same operation twice with different results) is nothing that we should take easily.

saemiw:
works perfectly on my machine now (with nightly build)! thanks

PR #32528, opened 2019-10-30
Create child feature with geometry from the relation editor

troopa81:
## Description

This PR allows to create a child feature by digitizing its geometry in the canvas directly from the relation editor widget. 

createchild

## Checklist

### DIFF
--- a/src/gui/qgsattributeeditorcontext.h
+++ b/src/gui/qgsattributeeditorcontext.h
@@ -28,6 +28,7 @@
 
 class QgsMapCanvas;
 class QgsAdvancedDigitizingDockWidget;
+class QgsMessageBar;
 
 /**
  * \ingroup gui
@@ -79,6 +80,7 @@ class GUI_EXPORT QgsAttributeEditorContext
       : mParentContext( &parentContext )
       , mVectorLayerTools( parentContext.mVectorLayerTools )
       , mMapCanvas( parentContext.mMapCanvas )
+      , mMainInfoBar( parentContext.mMainInfoBar )
       , mCadDockWidget( parentContext.mCadDockWidget )
       , mDistanceArea( parentContext.mDistanceArea )
       , mFormFeature( parentContext.mFormFeature )
@@ -91,6 +93,7 @@ class GUI_EXPORT QgsAttributeEditorContext
       : mParentContext( &parentContext )
       , mVectorLayerTools( parentContext.mVectorLayerTools )
       , mMapCanvas( parentContext.mMapCanvas )
+      , mMainInfoBar( parentContext.mMainInfoBar )
       , mCadDockWidget( parentContext.mCadDockWidget )
       , mDistanceArea( parentContext.mDistanceArea )
       , mRelation( relation )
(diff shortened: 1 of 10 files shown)


src/gui/qgsattributeeditorcontext.h
```diff
@@ -261,11 +264,24 @@ class GUI_EXPORT QgsAttributeEditorContext
       return metaEnum.valueToKey( static_cast<int>( mAttributeFormMode ) );
     }
 
+    /**
+     * Set current \a messageBar as the main information bar
+     * \since QGIS 3.12
+     */
+    void setMainInfoBar( QgsMessageBar *messageBar ) { mMainInfoBar = messageBar; }
```
m-kuhn: Is the term `mainInfoBar` (or just `infoBar`) used often? Or should `messageBar` be used which would be more coherent in taxonomy?

m-kuhn (commented):
Nice functionality.

- What happens if the user switches the map tool while he is in digitize child feature mode?
- How long is the message shown in the message bar on top? Would it be possible (or is it already the case?) to show the message until the user finished drawing or cancelled the map tool?

troopa81:
Thanks for review

> What happens if the user switches the map tool while he is in digitize child feature mode?

the new map tool is activated, cursor changes accordingly to this new map tool `QgsRelationEditorWidget::mapToolDeactivated` is called, the message bar disappear 

>    How long is the message shown in the message bar on top? Would it be possible (or is it already the case?) to show the message until the user finished drawing or cancelled the map tool?

It's already the case

m-kuhn:
Oh sorry, I missed that last time.
What happens if the user manually closes the message bar and then deactivates the tool? Is there no crash in this scenario? I.e. is there a risk that `mMessageBarItem` points to an already deleted item?

troopa81:
> Is there no crash in this scenario? 

No, because I never delete the item but...

> I.e. is there a risk that mMessageBarItem points to an already deleted item?

Yes. I added a QPointer cfa028f to avoid dangling pointer.

m-kuhn:
> > Is there no crash in this scenario?
> 
> No, because I never delete the item but...
> 

But you remove it from the message bar, so there is a risk for use after free ...

> > I.e. is there a risk that mMessageBarItem points to an already deleted item?
> 
> Yes. I added a QPointer [cfa028f](https://github.com/qgis/QGIS/commit/cfa028f5299075cb2df2b3f6e019790fbe0cda2f) to avoid dangling pointer.

... which should be solved by this :+1:

m-kuhn (approved):

src/gui/qgsattributeeditorcontext.h
troopa81: Is `MainMessageBar` OK?

My point was to be clear about what messageBar it is . This is the "main" one, the "application" one, the "canvas" one. This is not the one that you have in some widget like the QgsAttributeForm for instance.

m-kuhn: I see. My point was indeed rather on the `Info` than on the `main` part of the name. Maybe canvas would be good (since it's used to for reporting messages for the map canvas tool), but no strong opinion.

troopa81: I choose mainMessageBar because it's not mandatory to be in the canvas (it is actually but could be otherwise)

mhugo:
Since @troopa81 is on leave and @m-kuhn seems to be ok with the proposition, I'll merge.

PR #32542, opened 2019-10-31
Changed Python inspect.getargspec()

### LINKED ISSUE #32516: python3 inspect.getargspec() error
Hi !

From [inspect.getargspec doc](https://docs.python.org/3.8/library/inspect.html#inspect.getargspec) we can see the following :

`Deprecated since version 3.0: Use getfullargspec() for an updated API that is usually a drop-in replacement, but also correctly handles function annotations and keyword-only parameters.`

I want to add annotations to my QGIS forms, but I can't because [qgsattributeform.cpp](https://github.com/qgis/QGIS/blob/b2e712149312390b483cfb3c9be4c63e875694f0/src/gui/qgsattributeform.cpp#L1767) still uses `getargspec()` instead of `getfullargspec()`.

The error I run into:
`ValueError: Function has keyword-only parameters or annotations, use getfullargspec() API which can support them`

Djedouas:
inspect.getargspec() is deprecated since python 3.
replacement with inspect.getfullargspec().

Python documentation extract : 

```
Deprecated since version 3.0: Use getfullargspec() for an updated API that is usually a drop-in replacement, but also correctly handles function annotations and keyword-only parameters.
```

Fixes #32516

### DIFF
--- a/src/gui/qgsattributeform.cpp
+++ b/src/gui/qgsattributeform.cpp
@@ -1764,7 +1764,7 @@ void QgsAttributeForm::initPython()
     QString numArgs;
 
     // Check for eval result
-    if ( QgsPythonRunner::eval( QStringLiteral( "len(inspect.getargspec(%1)[0])" ).arg( initFunction ), numArgs ) )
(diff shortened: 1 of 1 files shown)


m-kuhn (approved):
Confirmed and successfully tested.
Good work!

PR #32546, opened 2019-10-31
[feature] Ignored credentials temporary cache

elpaso:
## Description

The problem we want to solve here is when a project contains multiple layers that require access to the same connection and that connection is not available.

What happens now is that the credentials dialog will open for each (broken) layer and there is no way to tell QGIS to ignore all future requests to the same connection.

## Proposed solution

A new "Ignore" button is added to the dialog, with a tooltip explaining what it does.

Screenshot_20191031_140513

When the "Ignore" button is clicked, the connection is added to a temporary cache of ignored connections and all future credentials requests to the same connection will be automatically rejected without the user intervention.

The connection is automatically removed from the cache after ~~5~~ 60 seconds to allow future retries to the same connection.

Funded by **ARPA Piemonte**

### DIFF
--- a/src/gui/qgscredentialdialog.cpp
+++ b/src/gui/qgscredentialdialog.cpp
@@ -24,6 +24,11 @@
 
 #include <QPushButton>
 #include <QThread>
+#include <QTimer>
+
+QSet<QString> QgsCredentialDialog::sIgnoredConnectionsCache;
+QMutex QgsCredentialDialog::sIgnoredConnectionsCacheMutex;
+
 
 static QString invalidStyle_( const QString &selector = QStringLiteral( "QLineEdit" ) )
 {
@@ -46,6 +51,27 @@ QgsCredentialDialog::QgsCredentialDialog( QWidget *parent, Qt::WindowFlags fl )
            this, &QgsCredentialDialog::requestCredentialsMasterPassword,
            Qt::BlockingQueuedConnection );
   mOkButton = buttonBox->button( QDialogButtonBox::Ok );
+  QPushButton *ignoreButton { buttonBox->button( QDialogButtonBox::StandardButton::Ignore ) };
+  ignoreButton->setToolTip( tr( "All requests for this connection will be automatically rejected for the next 5 seconds." ) );
+
+  // Keep a cache of ignored connections, and ignore them for 5 seconds.
+  connect( ignoreButton, &QPushButton::clicked, [ = ]( bool )
+  {
+    const QString realm { labelRealm->text() };
+    {
+      QMutexLocker locker( &sIgnoredConnectionsCacheMutex );
+      // Insert the realm in the cache of ignored connections
+      sIgnoredConnectionsCache.insert( realm );
+    }
+    QTimer::singleShot( 5000, [ = ]()
+    {
+      QgsDebugMsgLevel( QStringLiteral( "Removing ignored connection from cache: %1" ).arg( realm ), 4 );
+      QMutexLocker locker( &sIgnoredConnectionsCacheMutex );
+      sIgnoredConnectionsCache.remove( realm );
+    } );
+    accept( );
+  } );
+
   leMasterPass->setPlaceholderText( tr( "Required" ) );
   chkbxPasswordHelperEnable->setText( tr( "Store/update the master password in your %1" )
                                       .arg( QgsAuthManager::AUTH_PASSWORD_HELPER_DISPLAY_NAME ) );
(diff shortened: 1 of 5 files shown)


nyalldawson:
Oh gosh yes! I hit this many many times a day in my current work. 

Why limiting to 5 seconds though? I think this could potentially apply to the whole session...

elpaso:
@nyalldawson mmm, I though about it, but I think that there are use cases where the user can fix the broken connection (for example by configuring a proxy or connecting to the network or starting the server or retrieving the new credentials) and retry.

I think we can make it 60 seconds instead of 5.

m-kuhn:
Can we leave it at 5 seconds or change to 10?
Someone starting the db and then wondering why it's still not working is a very valid scenario IMO.

m-kuhn:
Wouldn't it actually be be better to reset the flag upon a signal which is emitted when the project read finished? At least for postgres that should work.

```
connect( project, &QgsProject::readProject, this, [&ignoredConnectionsCache, realm]() { ignoredConnectionsCache.remove( realm ); } );
```

elpaso:
> Wouldn't it actually be be better to reset the flag upon a signal which is emitted when the project read finished? At least for postgres that should work.
> 
> ```
> connect( project, &QgsProject::readProject, this, [&ignoredConnectionsCache, realm]() { ignoredConnectionsCache.remove( realm ); } );
> ```

That would not work in all scenarios: a broken connection can happen at any time (even for layers already loaded), also, the credentials dialog can be triggered by a queued signal if called from non-GUI thread and I'm not sure if that will happen after the project read is completed (probably not, because the project read is sequential).

There is also the use case for HTTP layers (WMS & C.) where the connection goes down, that happens in threads for sure and it has nothing to do with the project load.

src/gui/qgscredentialdialog.cpp
```diff
@@ -24,6 +24,11 @@
 
 #include <QPushButton>
 #include <QThread>
+#include <QTimer>
+
+QSet<QString> QgsCredentialDialog::sIgnoredConnectionsCache;
```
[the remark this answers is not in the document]
elpaso: I'll try with a local static, it sounds simpler to me.

m-kuhn:
Alright, I managed to reproduce a repeated dialog in other ways, let's keep it :+1:

- What do you think about relabeling the button to "Ignore for 10 seconds"?
- And adding a toolbutton (dropdown style) "Ignore for session"? (Just for @nyalldawson )

elpaso:
> Alright, I managed to reproduce a repeated dialog in other ways, let's keep it +1
> 
>     * What do you think about relabeling the button to "Ignore for 10 seconds"?
> 
>     * And adding a toolbutton (dropdown style) "Ignore for session"? (Just for @nyalldawson )

this is a nice solution :+1:

m-kuhn:
And if you want to stick to global static, I just found a safe solution
https://github.com/qgis/QGIS/pull/32517

src/gui/qgscredentialdialog.cpp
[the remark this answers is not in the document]
elpaso: > And if you want to stick to global static, I just found a safe solution
> #32517

`Q_GLOBAL_STATIC` does the trick?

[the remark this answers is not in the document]
m-kuhn: Exactly

src/gui/qgscredentialdialog.cpp
```diff
@@ -45,7 +56,58 @@ QgsCredentialDialog::QgsCredentialDialog( QWidget *parent, Qt::WindowFlags fl )
   connect( this, &QgsCredentialDialog::credentialsRequestedMasterPassword,
            this, &QgsCredentialDialog::requestCredentialsMasterPassword,
            Qt::BlockingQueuedConnection );
-  mOkButton = buttonBox->button( QDialogButtonBox::Ok );
+
+  // Setup ignore button
+  mIgnoreButton->setToolTip( tr( "All requests for this connection will be automatically rejected" ) );
+  QMenu *menu = new QMenu( mIgnoreButton );
+  QAction *ignoreTemporarily = new QAction( tr( "Ignore for 10 seconds" ), menu );
```
nyalldawson: Action text should be title case. "Ignore for 10 Seconds")

src/gui/qgscredentialdialog.cpp
```diff
@@ -45,7 +56,58 @@ QgsCredentialDialog::QgsCredentialDialog( QWidget *parent, Qt::WindowFlags fl )
...
+  ignoreTemporarily->setToolTip( tr( "All requests for this connection will be automatically rejected for 10 seconds" ) );
+  QAction *ignoreForSession = new QAction( tr( "Ignore for session" ), menu );
```
nyalldawson: Ignore for Session

src/gui/qgscredentialdialog.cpp
```diff
@@ -45,7 +56,58 @@ QgsCredentialDialog::QgsCredentialDialog( QWidget *parent, Qt::WindowFlags fl )
...
+  // Keep a cache of ignored connections, and ignore them for 10 seconds.
+  connect( mIgnoreButton, &QPushButton::clicked, [ = ]( bool )
+  {
+    const QString realm { labelRealm->text() };
+    {
+      QMutexLocker locker( &sIgnoredConnectionsCacheMutex );
+      // Insert the realm in the cache of ignored connections
+      sIgnoredConnectionsCache->insert( realm );
+    }
+    if ( mIgnoreMode == IgnoreTemporarily )
```
nyalldawson: Use a switch here instead of if

nyalldawson (commented):
Looks great!! Thanks for the new option!

nyalldawson (approved):

nyalldawson:
Fwiw, I would be happy to see this labelled as a bug fix (it fixes an irritating ui behaviour) and backported

src/gui/qgscredentialdialog.cpp
elpaso: why? I'm not handling the other case.

nyalldawson: Makes it much easier to extend this in future, because you get compiler warnings wherever a new enum value is added. It's one of the core C++ guidelines - you should always use switches for enum based logic.

elpaso: I fully agree in the general case where a switch is better than a chin of if/else if, but here it would just make the code much less readable because the enum is just a replacement for a bool.

planemad:
Thank you for this fix @elpaso ❤️ . Closes https://github.com/qgis/QGIS/issues/32899

PR #32551, opened 2019-10-31
[FEATURE] Add datetime_from_epoch (MSec from epoch) expression function

rduivenvoorde:
## Description

Adds a function to create a DateTime from  a seconds since epoch timestamp, to be used in expresssions

As there is an 'epoch' function working in MilliSeconds (and the original Qt function also does MilliSeconds, I create it to use mSecs).

The help file I've added does not pop up in the help, should I 'registre' it?

Screenshot-20191031160653-935x647

Another discussion, related to all DateTime functions is that most of them are (silently) returning a DateTime in your local Timezone, see https://doc.qt.io/qt-5/qdatetime.html#fromMSecsSinceEpoch

This makes the example in the Help of 'epoch' give another result then shown when I copy paste the example from the help.
I'm not sure what others think.
We could:
- make it more implicit that this is happening (by putting it in the help text, like I did for from_epoch)
- we could create from_epoch_utc and epoch_utc, maybe a little overkill

I will also create a test line in 
https://github.com/qgis/QGIS/blob/master/tests/src/core/testqgsexpression.cpp

### DIFF
--- a/resources/function_help/json/from_epoch
+++ b/resources/function_help/json/from_epoch
@@ -1,5 +1,5 @@
 {
-  "name": "from epoch",
+  "name": "from_epoch",
   "type": "function",
   "description": "Returns a datetime whose date and time are the number of milliseconds, msecs, that have passed since 1970-01-01T00:00:00.000, Coordinated Universal Time (Qt.UTC), and converted to Qt.LocalTime.",
   "arguments": [ {"arg":"int","description":"number (milliseconds)"} ],
--- a/resources/function_help/json/from_epoch
+++ b/resources/function_help/json/from_epoch
@@ -1,0 +1,7 @@
+{
+  "name": "from epoch",
+  "type": "function",
+  "description": "Returns a datetime whose date and time are the number of milliseconds, msecs, that have passed since 1970-01-01T00:00:00.000, Coordinated Universal Time (Qt.UTC), and converted to Qt.LocalTime.",
+  "arguments": [ {"arg":"int","description":"number (milliseconds)"} ],
+  "examples": [ { "expression":"from epoch(1483225200000)", "returns":"2017-01-01T00:00:00"} ]
+}
--- a/resources/function_help/json/from_epoch
+++ b/resources/function_help/json/from_epoch
@@ -1,7 +1,7 @@
 {
-  "name": "from_epoch",
+  "name": "datetime_from_epoch",
   "type": "function",
   "description": "Returns a datetime whose date and time are the number of milliseconds, msecs, that have passed since 1970-01-01T00:00:00.000, Coordinated Universal Time (Qt.UTC), and converted to Qt.LocalTime.",
   "arguments": [ {"arg":"int","description":"number (milliseconds)"} ],
-  "examples": [ { "expression":"from epoch(1483225200000)", "returns":"2017-01-01T00:00:00"} ]
+  "examples": [ { "expression":"datetime_from_epoch(1483225200000)", "returns":"2017-01-01T00:00:00"} ]
 }
--- a/src/core/expression/qgsexpressionfunction.cpp
+++ b/src/core/expression/qgsexpressionfunction.cpp
@@ -1969,7 +1969,7 @@ static QVariant fcnEpoch( const QVariantList &values, const QgsExpressionContext
   }
 }
 
-static QVariant fcnFromEpoch( const QVariantList &values, const QgsExpressionContext *, QgsExpression *parent, const QgsExpressionNodeFunction * )
+static QVariant fcnDateTimeFromEpoch( const QVariantList &values, const QgsExpressionContext *, QgsExpression *parent, const QgsExpressionNodeFunction * )
 {
   long long millisecs_since_epoch = QgsExpressionUtils::getIntValue( values.at( 0 ), parent );
   // no sense to check for strange values, as Qt behaviour is undefined anyway (see docs)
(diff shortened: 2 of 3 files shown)


m-kuhn:
:tada: nice

can we name it `datetime_from_epoch`?

rduivenvoorde:
@m-kuhn sure, no problem. 

What is your opinion about the TimeZone part? Keep it like this (with a clear description in the help json: "Returns a datetime whose date and time are the number of milliseconds, msecs, that have passed since 1970-01-01T00:00:00.000, Coordinated Universal Time (Qt.UTC), and converted to Qt.LocalTime.")

Or 2 separate functions:
datetime_utc_from_epoch (also setting it to UTC)
datetime_from_epoch (current version)

m-kuhn:
I have a slight preference for 2 distinct functions.

resources/function_help/json/from_epoch
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "from epoch",
```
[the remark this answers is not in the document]
Gustry: As you write, I see them WITH underscores too, that's why you should add it too ;-)

Look to my diff...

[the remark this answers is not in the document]
rduivenvoorde: Ah, duh, I see now :-) Sorry

src/core/expression/qgsexpressionfunction.cpp
```diff
@@ -1969,6 +1969,13 @@ static QVariant fcnEpoch( const QVariantList &values, const QgsExpressionContext
   }
 }
 
+static QVariant fcnFromEpoch( const QVariantList &values, const QgsExpressionContext *, QgsExpression *parent, const QgsExpressionNodeFunction * )
+{
+  long millisecs_since_epoch = QgsExpressionUtils::getIntValue( values.at( 0 ), parent );
```
nyalldawson:
```suggestion
  long long millisecs_since_epoch = QgsExpressionUtils::getIntValue( values.at( 0 ), parent );
```

src/core/expression/qgsexpressionfunction.cpp
```diff
@@ -5271,6 +5278,7 @@ const QList<QgsExpressionFunction *> &QgsExpression::Functions()
         << new QgsStaticExpressionFunction( QStringLiteral( "minute" ),  QgsExpressionFunction::ParameterList() << QgsExpressionFunction::Parameter( QStringLiteral( "datetime" ) ), fcnMinute, QStringLiteral( "Date and Time" ) )
         << new QgsStaticExpressionFunction( QStringLiteral( "second" ),  QgsExpressionFunction::ParameterList() << QgsExpressionFunction::Parameter( QStringLiteral( "datetime" ) ), fcnSeconds, QStringLiteral( "Date and Time" ) )
         << new QgsStaticExpressionFunction( QStringLiteral( "epoch" ), QgsExpressionFunction::ParameterList() << QgsExpressionFunction::Parameter( QStringLiteral( "date" ) ), fcnEpoch, QStringLiteral( "Date and Time" ) )
+        << new QgsStaticExpressionFunction( QStringLiteral( "from_epoch" ), QgsExpressionFunction::ParameterList() << QgsExpressionFunction::Parameter( QStringLiteral( "long" ) ), fcnFromEpoch, QStringLiteral( "Date and Time" ) )
```
nyalldawson:
```suggestion
        << new QgsStaticExpressionFunction( QStringLiteral( "from_epoch" ), QgsExpressionFunction::ParameterList() << QgsExpressionFunction::Parameter( QStringLiteral( "ms" ) ), fcnFromEpoch, QStringLiteral( "Date and Time" ) )
```

or milliseconds? or epoch?

tests/src/core/testqgsexpression.cpp
```diff
@@ -1284,6 +1284,7 @@ class TestQgsExpression: public QObject
       QTest::newRow( "time - time" ) << "to_time('08:30:00') - to_time('05:15:00')" << false << QVariant( QgsInterval( 3 * 60 * 60 + 15 * 60 ) );
       QTest::newRow( "epoch" ) << "epoch(to_datetime('2017-01-01T00:00:01+00:00'))" << false << QVariant( 1483228801000LL );
       QTest::newRow( "epoch invalid date" ) << "epoch('invalid')" << true << QVariant();
+      QTest::newRow( "from_epoch" ) << "from_epoch(1483228801000)" << false << QVariant( QDateTime( QDate( 2017, 1, 1 ), QTime( 0, 0, 1 ), Qt::UTC ) );
```
nyalldawson: Just need a test with invalid inputs, e.g. a string value

nyalldawson:
Regarding timezones, I think we should keep everything timezone agnostic for now. We just don't have any support for handling these in date time values, and things will get very weird if this function respects them yet getting datetime values from field values doesn't...

nyalldawson:
(great feature by the way, nice work!)

rduivenvoorde:
> Regarding timezones, I think we should keep everything timezone agnostic for now. We just don't have any support for handling these in date time values, and things will get very weird if this function respects them yet getting datetime values from field values doesn't..

I'm fine with that. 
It is just that this 'hides' something and get's bad when people do not realize that they go from UTC to localtime. 
As a side note: the example in the helptext is not the same for me:
`epoch(to_date('2017-01-01'))`  does not give me `
1483203600000` as in the help text, but `1483225200000`

Personnally I'd prefer to always work in UTC in general.

I will rename the function to `datetime_from_epoch` and add a test.

rduivenvoorde:
Added changes, @Gustry I think I made a mistake and fixed it wrong. Both name and examples should have the version with the underscores in the json help.

Let me know if we want this to be 'squashed' into 1 commit

DelazJ:
Misses [feature] tag? :smile:

rduivenvoorde:
Had some trouble fixing the tests... as creating a Datetime from an epoch will currently (that is: TimeZone-ignorant) return a (local) DateTime, zo depending on the timezone where you run this test...
So ended up by creating an epoch, created by our own expression, feeding it with an UTC datetime.. Anyway, this works.

Then I also wanted to add a negative test, by feeding the expression a STRING instead of a long:

QTest::newRow( "datetime_from_epoch_string" ) << "datetime_from_epoch('1483228801000')" << true <<  QVariant(QVariant::Invalid);

But that does NOT fail, as I would expect... Because the expression functions apparently do not check the arguments? So '1483228801000' just works, just like 1483228801000. Both in the test AND in QGIS itself.
So I decided to just remove that test.

As said: I can squash this.

rduivenvoorde:
@DelazJ nope, was already added by Nyall, though I'm not sure if this is 'feature' enough to be called a 'feature' :-)

DelazJ:
> @DelazJ nope, was already added by Nyall, though I'm not sure if this is 'feature' enough to be called a 'feature' :-)

Hmmm.. unless there are changes I'm not aware of, that would not generate an issue in the doc repo after merge. Would it? And something we did not have or could do and then it's possible within QGIS **IS** a feature imho

rduivenvoorde:
Not sure if this is a real failing test?
Or is the comment line a problem?
./scripts/prepare-commit.sh does not fix anything...

tests/src/core/testqgsexpression.cpp
```diff
@@ -1284,6 +1284,9 @@ class TestQgsExpression: public QObject
       QTest::newRow( "time - time" ) << "to_time('08:30:00') - to_time('05:15:00')" << false << QVariant( QgsInterval( 3 * 60 * 60 + 15 * 60 ) );
       QTest::newRow( "epoch" ) << "epoch(to_datetime('2017-01-01T00:00:01+00:00'))" << false << QVariant( 1483228801000LL );
       QTest::newRow( "epoch invalid date" ) << "epoch('invalid')" << true << QVariant();
+      // datetime_from_epoch will always return a local datetime, so here we create some circular magic to create a local datetime during test (so test can be ran in every timezone...)
+      QTest::newRow( "datetime_from_epoch" ) << "datetime_from_epoch(epoch(to_datetime('2017-01-01T00:00:01')))" << false << QVariant( QDateTime( QDate( 2017, 1, 1 ), QTime( 0, 0, 1 ), Qt::LocalTime ) );
+      QTest::newRow( "datetime_from_epoch_null" ) << "datetime_from_epoch(NULL)" << false <<  QVariant(QVariant::Invalid);
```
m-kuhn:
```suggestion
      QTest::newRow( "datetime_from_epoch_null" ) << "datetime_from_epoch(NULL)" << false <<  QVariant( QVariant::Invalid );
```

src/core/expression/qgsexpressionfunction.cpp
```diff
@@ -1969,6 +1969,13 @@ static QVariant fcnEpoch( const QVariantList &values, const QgsExpressionContext
   }
 }
 
+static QVariant fcnDateTimeFromEpoch( const QVariantList &values, const QgsExpressionContext *, QgsExpression *parent, const QgsExpressionNodeFunction * )
+{
+  long long millisecs_since_epoch = QgsExpressionUtils::getIntValue( values.at( 0 ), parent );
+  // no sense to check for strange values, as Qt behaviour is undefined anyway (see docs)
```
```suggestion
  // no sense to check for strange values, as Qt behavior is undefined anyway (see docs)
```

m-kuhn (commented):
One would have been fixed by prepare commit, one was a typo (or rather English localization variant) that needs manual fixing.

On the other question, expressions are like SQL, type checking is not really enforced. So it's possible to do `4 + '3'` (which is `7`) like in postgres.

rduivenvoorde:
@m-kuhn @nyalldawson should I squash this to one commit?

m-kuhn:
Thanks @rduivenvoorde !
It has just been forgotten

rduivenvoorde:
@m-kuhn thanks! On hind-sight, it would probably better if I had squased:

Screenshot-20191122122127-1007x366

I thought git was doing some magic during merging... These are too much commits for such a trivial feature... Anyway, thanks!

m-kuhn:
There would actually even be an option for that in the github merge button, but that's really just a detail. 

There's now a clean implementation (thanks) and the discussion history of the feature inside this PR here which is linked to the individual commits.

PR #32561, opened 2019-10-31
[FEATURE] Add expression functions for converting to/from wkb

nyalldawson:
Adds geom_from_wkb and geom_to_wkb, which mirror the existing
geom_from_wkt/geom_to_wkt functions but for WKB representations
of geometries.

Since QGIS 3.6 we've had good support for binary blob values in
expressions and field values, so adding these functions allows
users to work with binary blob fields containing WKB representations
of geometries (e.g. with a geometry generator showing the encoded
geometries)

### DIFF
--- a/resources/function_help/json/geom_from_wkb
+++ b/resources/function_help/json/geom_from_wkb
@@ -1,0 +1,7 @@
+{
+  "name": "geom_from_wkb",
+  "type": "function",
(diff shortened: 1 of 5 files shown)


nyalldawson:
Sponsored by the SLYR project

nirvn (approved):
Looking good.

nyalldawson:
Thanks @nirvn!

PR #32562, opened 2019-10-31
fix Processing/GRASS v_net_centrality

gioman:
Trying to fix this GRASS module after

https://lists.osgeo.org/pipermail/qgis-developer/2019-October/059084.html

I removed a couple of parameters that were completely wrong (never existed even in old GRASS releases)

The module description now looks more like as it was in 2.18. And there was a reason why it was like that. Basically the Processing implementation only works if extracting nodes from the line network, and not by using an existing node layer (which makes the tool return always an empty layer). For this reason the "Add points on nodes" parameter was hardcoded and the parameters that have to do with an existing node layer were removed.

I have a gut feeling that there are possibly many other GRASS tools with similar issues.

### DIFF
--- a/python/plugins/processing/algs/grass7/description/v.net.centrality.txt
+++ b/python/plugins/processing/algs/grass7/description/v.net.centrality.txt
@@ -1,5 +1,5 @@
 v.net.centrality
-Computes degree, centrality, betweeness, closeness and eigenvector centrality measures in the network.
+Computes degree, centrality, betweenness, closeness and eigenvector centrality measures in the network.
 Vector (v.*)
 QgsProcessingParameterFeatureSource|input|Input vector line layer (network)|1|None|False
 QgsProcessingParameterString|degree|Name of output degree centrality column|degree|False|True
--- a/python/plugins/processing/algs/grass7/description/v.net.centrality.txt
+++ b/python/plugins/processing/algs/grass7/description/v.net.centrality.txt
@@ -1,20 +1,15 @@
 v.net.centrality
-Computes degree, centrality, betweenness, closeness and eigenvector centrality measures in the network.
+Computes degree, centrality, betweeness, closeness and eigenvector centrality measures in the network.
 Vector (v.*)
 QgsProcessingParameterFeatureSource|input|Input vector line layer (network)|1|None|False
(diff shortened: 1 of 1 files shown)


gioman:
Not stale.

nyalldawson:
Is this one ready for merge?

gioman:
> Is this one ready for merge?

yes, the module is in better shape than before (completely wrong parameters).

python/plugins/processing/algs/grass7/description/v.net.centrality.txt
```diff
@@ -1,20 +1,15 @@
 v.net.centrality
-Computes degree, centrality, betweenness, closeness and eigenvector centrality measures in the network.
+Computes degree, centrality, betweeness, closeness and eigenvector centrality measures in the network.
```
nyalldawson: Actually "betweenness" is correct, it's causing the spelling failure

gioman: rats, I copied it from here https://github.com/qgis/QGIS/blob/release-2_18/python/plugins/processing/algs/grass7/description/v.net.centrality.txt#L2

nyalldawson: There's a good lesson there in not trusting 2.18 for **anything** 🙃

gioman: :)

nyalldawson:
@gioman can you update the test results so we can merge this?

gioman:
Not stale.

gioman:
Not stale.

nyalldawson:
@gioman do you plan on fixing the tests here? Otherwise could we let this pr close until you are ready to revisit? (A lengthy pr queue makes reviewing difficult and stressful!)

nirvn:
Funny, I was about to hit the comment button with a similar message :) this PR very much is the dictionary definition of stalled. We all have priorities that get in the way, there's no problem in closing a PR and revisiting later. It makes for a more relevant list of actively worked PRs.

alexbruy:
Superseded by #33613

gioman:
> Superseded by #33613

thanks @alexbruy

PR #32570, opened 2019-11-01
[feature] Add API for event tracing

wonder-sk:
This is more developer oriented feature to help with profiling, debugging and optimization - especially when dealing with multi-threaded code. I am very excited about the new possibilities it can give us :rocket: 

From doxygen:

A utility class that provides event tracing functionality. When tracing
is enabled, events from different threads can be recorded and stored
in a JSON file. This helps understanding what is going on in multi-threaded
environment of QGIS and trace some performance issues that are otherwise
not that easy to spot in profiler output.

Created traces can be viewed in Chrome/Chromium browser - simply set
the URL to about:tracing and load the JSON file. There is also a tool
called trace2html that turns the JSON file into a standalone HTML page
that can be viewed anywhere.

Trace viewer project is hosted here (created by Chrome developers):
https://github.com/catapult-project/catapult/tree/master/tracing

Event trace format specification:
https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview

The intended usage:
1. call startTracing()
2. repeatedly call addEvent()
3. call stopTracing() and writeTrace() to export the data to JSON

---

I am going to use it for improving the 3D rendering pipeline, to better understand what is going on under the hood. So far I briefly tested it with 2D map rendering on a project with XYZ:

trace-events-0

(It has quickly shown a bunch of performance issues there which were not obvious before)

The trace viewer GUI is really good, it allows zooming in/out, panning, measuring time intervals and even allows things like selecting a bunch of events to get aggregate statistics on time.

This is just a start - we could/should add some more information to the events to make the traces even more useful. Also there are more event types that the trace viewer recognizes that we could use...

I was thinking we could have the event tracer used for 2D map rendering - and with a particular shortcut we would enable tracing for the next renderer job or something like that - this could also improve our debugging options e.g. when some users a getting long rendering time we could instruct them to record the trace and send us the JSON...

Inspiration from @pablo-odorico

### DIFF
--- a/src/core/qgseventtracing.cpp
+++ b/src/core/qgseventtracing.cpp
@@ -1,0 +1,98 @@
+/***************************************************************************
+  qgseventtracing.cpp
+  --------------------------------------
+  Date                 : October 2019
+  Copyright            : (C) 2019 by Martin Dobias
+  Email                : wonder dot sk at gmail dot com
+ ***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+#include "qgseventtracing.h"
+
+#include <QFile>
+#include <QThread>
+
+/// @cond PRIVATE
+
+bool QgsEventTracing::sIsTracing = false;
+QElapsedTimer QgsEventTracing::sTracingTimer;
+QVector<QgsEventTracing::TraceItem> QgsEventTracing::sTraceEvents;
+QMutex QgsEventTracing::sTraceEventsMutex;
+
+
+bool QgsEventTracing::startTracing()
+{
+  if ( sIsTracing )
+    return false;
+
+  sIsTracing = true;
+  sTraceEventsMutex.lock();
+  sTracingTimer.start();
+  sTraceEvents.clear();
+  sTraceEvents.reserve( 1000 );
+  sTraceEventsMutex.unlock();
+  return true;
+}
+
+bool QgsEventTracing::stopTracing()
+{
+  if ( !sIsTracing )
+    return false;
+
+  sIsTracing = false;
+  sTracingTimer.invalidate();
+  return false;
+}
+
+bool QgsEventTracing::writeTrace( const QString &fileName )
+{
+  if ( sIsTracing )
+    return false;
+
+  QFile f( fileName );
+  if ( !f.open( QIODevice::WriteOnly ) )
+    return false;
+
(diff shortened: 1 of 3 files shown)


nyalldawson:
@wonder-sk have you tried kdab's hotspot before? It provides a similar hi, and I've used it extensively for identifying code to optimise in qgis. ( It doesn't require any changes to the code to work either)

wonder-sk:
Yeah I have used Hotspot before. It is a great tool, but the event tracing has strengths elsewhere:
- you only instrument what you need - the trace can be more high level and you can add custom data, moreover you can connect events like start and finish of a network request that are handled in different parts of the code
- cross-platform (not just linux perf)
- can be used on systems without dev tools installed to get data from users running it in production

src/core/qgseventtracing.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+#include "qgseventtracing.h"
+
+#include <QFile>
+#include <QThread>
+
+/// @cond PRIVATE
+
+bool QgsEventTracing::sIsTracing = false;
+QElapsedTimer QgsEventTracing::sTracingTimer;
```
nyalldawson: @m-kuhn will have a fit if you merge these :rofl:

wonder-sk: hmm... is there anything wrong with that code? The static variables are not QObject based or dependent on any other QGIS stuff, so it should be all safe...

m-kuhn: Non-POD global statics are somewhere in the gray zone of being reliable constructs and happened to crash QGIS on exit. That's why I took the time to kill many of them just the last days.

https://github.com/KDE/clazy/blob/master/docs/checks/README-non-pod-global-static.md

wonder-sk: okay thanks for the pointer - changed to use Q_GLOBAL_STATIC

Saijin-Naib:
Would this event tracing be usable for performance/rendering profiling as per my GIS.SE question?  
https://gis.stackexchange.com/questions/342020/how-to-profile-rendering-performance-using-qgis-3-10-x-or-above

wonder-sk:
@Saijin-Naib eventually yes, but the problem is that currently the map rendering code has network/storage access interleaved with rendering, so you can't easily separate time for those and the actual rendering (painting) time. Maybe at some point if we separate data access and painting, this will be easier to figure out...

Saijin-Naib:
Sounds great. I imagine this is on the horizon due to the focus on QGIS server performance and profiling in order to get it certified, correct?

wonder-sk:
Not sure - certification and performance optimizations are separate things possibly handled by different people... And performance optimization is a task that is never really done :-)

NathanW2:
This is a super neat idea.  I really support this kind of tooling inside
QGIS core itself. Would really help to debug things.

PR #32575, opened 2019-11-01
Fix loading of PostgreSQL geometry-less views

### LINKED ISSUE #32523: PostGIS Views not valid layers
**Describe the bug**

When trying to add Views from a PostGIS database I get an error message that the layer is not valid. This was not a problem before 3.10. I also tried adding views from Spatialite and they work perfectly. Have not tried others.

**How to Reproduce**

1. Go to PostGIS database in Layer Browser
2. Double-Click on any View in the database or drag view into map canvas
3. Error appears in red at the top of the canvas

**QGIS and OS versions**
Vers. 3.10 on Ubuntu

**Additional context**

strk:
Closes #32523

Bug was introduced with commit bbdbca4

Thanks @elpaso for finding the root cause of it

### DIFF
--- a/src/providers/postgres/qgspostgresconn.cpp
+++ b/src/providers/postgres/qgspostgresconn.cpp
@@ -838,11 +838,11 @@ bool QgsPostgresConn::getTableInfo( bool searchGeometryColumnsOnly, bool searchP
       QString table   = result.PQgetvalue( i, 0 ); // relname
       QString schema  = result.PQgetvalue( i, 1 ); // nspname
       QString relkind = result.PQgetvalue( i, 2 ); // relation kind
-      QString attributes = result.PQgetvalue( i, 3 ); // attributes array
+      QString comment = result.PQgetvalue( i, 3 ); // table comment
+      QString attributes = result.PQgetvalue( i, 4 ); // attributes array
       bool isView = relkind == QLatin1String( "v" ) || relkind == QLatin1String( "m" );
       bool isMaterializedView = relkind == QLatin1String( "m" );
       bool isForeignTable = relkind == QLatin1String( "f" );
-      QString comment = result.PQgetvalue( i, 3 ); // table comment
 
       //QgsDebugMsg( QStringLiteral( "%1.%2: %3" ).arg( schema ).arg( table ).arg( relkind ) );
 
--- a/tests/src/python/test_qgsproviderconnection_postgres.py
+++ b/tests/src/python/test_qgsproviderconnection_postgres.py
@@ -111,6 +111,14 @@ class TestPyQgsProviderConnectionPostgres(unittest.TestCase, TestPyQgsProviderCo
         self.assertFalse('geometries_table' in table_names)
         self.assertFalse('geometries_view' in table_names)
 
+        tables = conn.tables('qgis_test', QgsAbstractDatabaseProviderConnection.Aspatial | QgsAbstractDatabaseProviderConnection.View)
+        table_names = self._table_names(tables)
+        b32523_view = self._table_by_name(tables, 'b32523')
+        self.assertTrue(b32523_view)
+        pks = b32523_view.primaryKeyColumns()
+        self.assertTrue('pk' in pks)
+        self.assertTrue('random' in pks)
+
         geometries_table = self._table_by_name(conn.tables('qgis_test'), 'geometries_table')
         srids_and_types = [[t.crs.postgisSrid(), t.wkbType]
                            for t in geometries_table.geometryColumnTypes()]
--- a/tests/testdata/provider/testdata_pg.sql
+++ b/tests/testdata/provider/testdata_pg.sql
@@ -662,3 +662,12 @@ CREATE VIEW qgis_test.some_poly_data_shift_bbox AS
 
 CREATE TABLE qgis_test.b31799_test_table AS (SELECT (ST_DumpPoints(ST_GeneratePoints(ST_Expand('SRID=4326;POINT(0 0)'::geometry,90),10))).geom, random());
 CREATE VIEW qgis_test.b31799_test_view_ctid AS (SELECT ctid, geom, random() FROM qgis_test.b31799_test_table, pg_sleep(0.1));
+
+---------------------------------------------
+--
+-- Geometryless view
+-- See https://github.com/qgis/QGIS/issues/32523
+--
+CREATE VIEW qgis_test.b32523 AS
+  SELECT pk, random()
+  FROM qgis_test.some_poly_data;


strk:
@elpaso I'm not sure the test I added is really testing anything. My impression is that what I broke was the list of _candidate_ primary keys, so I'd need to find out an API that fetches that list of candidates, rather than specifying a key myself, right ?

elpaso:
I think so, but I didn't really look at the code, as I mentioned, the error I could see is that instead of the PK candidate there is the view comment text, it's maybe just a wrong index when retrieving data from the tuple.

strk:
Yes, it was a wrong index, and is fixed, I'm just hoping to secure the fix with a test. Do you know of tests checking for "candidate" primary keys ? Are those maybe part of the new API you wrote ?

strk:
@elpaso I think I finally got it, with latest commit, can you please double-check ? I'm not sure about the bitwise-or, as it looks like that calls to connection.tables() is also returning tables (in addition to views) which is surprising to me

strk:
Travis confirms the test _works_ (fails w/out the fix): https://travis-ci.org/qgis/QGIS/jobs/609163362#L1399

mlebihan:
I'm still encountering the problem on QGIS 3.12.0

A view 
```
CREATE OR REPLACE VIEW public.view_desendettement_communes_2018
 AS
 SELECT d.codecommune,
    d.nomcommune,
    d.capacitedesendettement,
    c.geom
   FROM desendettement_communes_2018 d
     JOIN communes_2019 c ON d.codecommune::text = c.insee::text;
```

created from tables 
```
CREATE TABLE public.communes_2019
(
    gid integer NOT NULL DEFAULT nextval('communes_2019_gid_seq'::regclass),
    insee character varying(80) COLLATE pg_catalog."default",
    nom character varying(80) COLLATE pg_catalog."default",
    wikipedia character varying(80) COLLATE pg_catalog."default",
    surf_ha numeric,
    geom geometry(MultiPolygon,4326),
    CONSTRAINT communes_2019_pkey PRIMARY KEY (gid)
        USING INDEX TABLESPACE data
)
WITH (
    OIDS = FALSE
)
TABLESPACE data;
```
by a shp2pgsql command on an OpenStreetMap shapefile file,

and another another small table I have :
```
CREATE TABLE public.desendettement_communes_2018
(
    codecommune character varying(5) COLLATE pg_catalog."default" NOT NULL,
    nomcommune character varying COLLATE pg_catalog."default" NOT NULL,
    capacitedesendettement numeric(3,0),
    CONSTRAINT desendettement_communes_2018_pkey PRIMARY KEY (codecommune)
        USING INDEX TABLESPACE data
)
WITH (
    OIDS = FALSE
)
TABLESPACE data;
```
 
The view created from them, view_desendettement_communes_2018, has a correct content :
select * from view_desendettement_communes_2018 limit 3;

codecommune, nomcommune, d.capacitedesendettement, c.geom
"59590"	"Thiennes"	14	"010600..."
"77157"	"Dhuisy"		20	"010600..."
"59054"	"Bavinchove"	13	"010600..."

But I still have the problem :

"_CRITICAL    La couche n'est pas valide : La couche dbname='comptesfrance' host=localhost port=5432 sslmode=disable key='"codeCommune"' srid=4326 type=MultiPolygon table="public"."view_desendettement_communes_2018" (geom) n'est pas une couche valide et ne peut pas être ajoutée à la carte. Raison :_ "
(with no reason following)

gioman:
> I'm still encountering the problem on QGIS 3.12.0

@mlebihan can you post a dump of a minimal dataset (a schema with the necessary tables and views) we can use to try replicate this? thanks.

mlebihan:
Here is the table to be joined with its data :
[desendettement_communes.txt](https://github.com/qgis/QGIS/files/4304536/desendettement_communes.txt)

The communes_2019 table is created from a shapefile coming from data.gouv.fr 
[Contour des communes 2019](https://www.data.gouv.fr/fr/datasets/r/61b8f19d-66ce-4ad3-a9c4-82502dc9d550)

and generated by this command :
```
export PGPASSWORD=postgres
shp2pgsql -s 4326 -I /data/comptes-france/territoire/2019/communes.shp public.communes_2019 | psql -d comptesfrance -U postgres -h localhost

```
It creates the spatial index. A single city among 34000 has an invalid geometry inside it but removing it doesn't change anything. I added this index :
```
CREATE INDEX idx_communes_2019_insee ON communes_2019(insee) TABLESPACE data;

CREATE VIEW view_desendettement_communes_2018 AS 
			SELECT d.codeCommune, d.nomCommune, d.capaciteDesendettement, c.geom FROM desendettement_communes_2018 d 
			JOIN communes_2019 c ON d.codeCommune = c.insee;

```

For other settings, here is the "_About QGIS..._" display on my computer :
qgis_env

gioman:
> Attempting to use this view as a layer then fails.

@mlebihan thanks for the data/effort, but I would really appreciate if you could post a DB/schema dump, with the two tables, the join and the view. Thanks!

mlebihan:
I can't do more than what I've done, but you see the principle of what kind of table are joined with another and maybe you can find a faster way than this to set a test environment.

mlebihan:
Currently the work around is to copy the view content into a physical table : 
```
CREATE TABLE public.temp_desendettement_communes_2018(
    codecommune character varying(5) COLLATE pg_catalog."default" NOT NULL,
    nomcommune character varying COLLATE pg_catalog."default" NOT NULL,
    capacitedesendettement numeric(3,0),
	geom geometry(MultiPolygon,4326),
    CONSTRAINT temp_desendettement_communes_2018_pkey PRIMARY KEY (codecommune)
        USING INDEX TABLESPACE data
)
WITH (
    OIDS = FALSE
)
TABLESPACE data;

INSERT INTO temp_desendettement_communes_2018(codecommune, nomcommune, capacitedesendettement, geom)
SELECT codecommune, nomcommune, capacitedesendettement, geom
	FROM view_desendettement_communes_2018;
```

And then, it works : 
works_with_physical_table

But a the cost of data duplication, of course.

gioman:
> I can't do more than what I've done. Sorry. But I should explain better what to do with this.
> 
>     1. Run desendettement_communes.txt : it creates the first table.

@mlebihan 
```

ERROR:  syntax error at or near "07114"
LINE 80: 07114 Labatie-d'Andaure 5
         ^
********** Error **********

ERROR: syntax error at or near "07114"
SQL state: 42601
Character: 2339
```

posting a dump (pg_dump in binary format) of a sample schema with all the tables in it would avoid this problems and make testers life easier...

mlebihan:
It's strange because I used the backup command of PgAdmin 4 to do that.

But enough !
"make testers life easier..."
Each developer is responsible of his bugs. If I were someone who had only knowledge in cooking, I should also enter the arcana of a database because you in QGis team are to lazy to focus on your problems to create accurate tests ?

You have a bug. You correct it. **Even without test case.** It's your duty, because it's your bug.

the principle of what cause the bug is well explained enough. A tree field table without geometry field joining another who has it in a view. Stop waiting now.

gioman:
> But enough !
> "make testers life easier..."
> Each developer is responsible of his bugs. If I were someone who had only knowledge in cooking, I should also enter the arcana of a database because you in QGis team are too lazy to focus on your problems, to create accurate tests ?
> 
> You have a bug. You correct it. **Even without test case.** It's your duty, because it's your bug.
> 
> The principle of what cause the bug is explained enough. A tree field table without geometry field joining another who has it in a view. Stop waiting now.

@mlebihan I'm sorry o bring you the news... but you got it all wrong. Anyway the same issue has been reported by others, and they are being much more collaborative. So no need to open a new ticket, just stay tuned.

gioman:
@mlebihan by the way I have been able to restore your SQL file and then followed the rest of steps: here the view loads just fine:

Screenshot_20200311_124915

mlebihan:
Thanks a lot for your efforts.
If I meet the problem with my 3.12, it means that the installation I've received isn't the same package than yours.

[My configuration](https://user-images.githubusercontent.com/30030683/76181469-7a7fce00-61c1-11ea-9c36-89e7e13b2076.png)

gioman:
> If I meet the problem with my 3.12, it means that the installation I've received isn't the same package than yours. I still have the bug packaged on my 3.12, and you : no more.

@mlebihan 

Screenshot_20200311_160841

mlebihan:
As the hashcode shows, we don't share the same QGIS code revision, even if we have the same QGIS version.
Mine is **409cf8eaa**, yours : **cd141490ec**
QGis distributions 3.12 still hold the bug in some packages.

gioman:
>  Yours : [cd14149](https://github.com/qgis/QGIS/commit/cd141490eccf4340b658eab7fab1e7fb80c06214)

@mlebihan it is the latest release available on osgeo4w, which means that is more recent than yours. Moreover also no issues on master:

Screenshot_20200311_205854

mlebihan:
The difference between QGis [409cf8eaa](https://github.com/qgis/QGIS/commit/409bcf8eaa) and [cd14149](https://github.com/qgis/QGIS/commit/cd141490eccf4340b658eab7fab1e7fb80c06214) commits should be the cause of what is happening.
I do my installations by an apt-get install on _Debian 10_, with these entries in _/etc/apt/sources.list_ :
 ```
deb https://qgis.org/debian buster main
deb-src https://qgis.org/debian buster main
```

Where is the difference between these two installations that causes the trouble ? I don't know.
But using the same tables and view than you, I still have a message :
"_CRITICAL La couche n'est pas valide : La couche dbname='comptesfrance' host=localhost port=5432 sslmode=disable key='"codeCommune"' srid=4326 type=MultiPolygon table="public"."view_desendettement_communes_2018" (geom) n'est pas une couche valide et ne peut pas être ajoutée à la carte. Raison :_ "
that your package have no more.

**Comment by gioman** [comment]
> The difference between QGis [409cf8eaa](https://github.com/qgis/QGIS/commit/409bcf8eaa) and [cd14149](https://github.com/qgis/QGIS/commit/cd141490eccf4340b658eab7fab1e7fb80c06214) commits should be the cause of what is happening.

@mlebihan On Ubuntu 18.04:

Screenshot_20200312_121844

**Comment by mlebihan** [comment]
Humm... Then I have no clue.
But the bug is here.

**Comment by gioman** [comment]
>  But the bug is here.

@mlebihan the problem is likely to be in your system (client or server side).

**Comment by mlebihan** [comment]
With a reason "" ? What is the meaning of the reason "" ? ==> The programmer has not been able to trap errors correctly.

But I have a clue : I might be using Postgresql 12 with Postgis 3.0 since a last update.

**Comment by gioman** [comment]
> But I have a clue : I might be using Postgresql 12 with Postgis 3.0 since a last update.

@mlebihan let me check that, you can be right.

**Comment by gioman** [comment]
> But I have a clue : I might be using Postgresql 12 with Postgis 3.0 since a last update.

@mlebihan so... let's try to clarify:

1) tested with pgdql 12 and postgis 3, does not make any difference here (your view loads)

2) the view loads fine using the QGIS Browser and the "add PostGIS layer" dialog

3) the view does NOT load using DB Manager:

4) it does not load in any recent QGIS version as also on 2.18

5) moreover (on top of point 5) on 3.12.0 in DB Manager the mechanism the recognizes views and materialized views is broken, but it was already fixed on master, so it will also be on 3.12.1

6) your view loads OK also with DB Manager (in any QGIS release, except 3.12.0 for now, see point 5) if you modify the query as

```
CREATE VIEW view_desendettement_communes_2018 AS 
   SELECT row_number() OVER () AS gid, d.codeCommune, d.nomCommune, d.capaciteDesendettement, c.geom FROM desendettement_communes_2018 d 
   JOIN communes_2019 c ON d.codeCommune = c.insee;
```

you should always add

`row_number() OVER () AS gid`

to your views and materialized views to avoid issues (in QGIS and other software).

**Comment by mlebihan** [comment]
Good !

Your trick works : creating a view adding what you suggest.

CREATE VIEW view_desendettement_communes_2018 AS 
   **SELECT row_number() OVER () AS gid**, d.codeCommune, d.nomCommune, d.capaciteDesendettement, c.geom FROM desendettement_communes_2018 d 
   JOIN communes_2019 c ON d.codeCommune = c.insee;

This a good work around.
However, _QGis_ should be able to work without. You can't expect that it will met views or tables containing this _gid_ column each time, and you have not always the ability to change the view you are using. (especially if you work in a company and that you are not the one who have created the view).

So currently, the issue has to remain open until a view without _gid_ is able to work too.
By the way, why a table without a _gid_ field can work well on _QGis_, but a view cannot ?

**Comment by gioman** [comment]
>  By the way, why a table without a _gid_ field can work well on _QGis_, but a view cannot ?

@mlebihan the bottom line is that since I can remember QGIS always expected (for database tables) a column (call it gid or as you want) acting as PK. It does not mean that also works without, you can add your original view from the browser or the "add postgis layer" dialog. Having such column is just a best practice, there is no overhead in adding it to your views.

**Comment by haubourg** [comment]
>  However, _QGis_ should be able to work without. You can't expect that it will met views or tables containing this _gid_ column each time, and you don't always have the ability to change the view you are using (...especially if you work in a company and you are not the one who have created the view !).

Hi, I confirm a unique key has always been necessary in QGIS. Any database system needs a Primary key. Some just hide it from your eyes. 

Yes, the error message for invalid layers should be more explicit. We look for support to implement this. That would help so much people troubleshooting QGIS. 

Just add a primary key constraint on your field `code_commune` and you will probably find you have duplicates. If you don't have a unique field, you can generate one in a query, but this is a workaround for a crappy dataset.  

About the wording earlier in this thread  :

Yes troubleshooting issues can be hard. 
No, shouting does not help.

@gioman and most reviewers are not paid by a corporation to help you. We mostly are dedicated passionate trying to help you. Being constructive will help.
 
If you are really blocked, please consider getting commercial support and have your issues fixed. 

And finally, please read the GPL licence more carefully : 

```
15. Disclaimer of Warranty.

THERE IS NO WARRANTY FOR THE PROGRAM, TO THE EXTENT PERMITTED BY APPLICABLE LAW. EXCEPT WHEN OTHERWISE STATED IN WRITING THE COPYRIGHT HOLDERS AND/OR OTHER PARTIES PROVIDE THE PROGRAM “AS IS” WITHOUT WARRANTY OF ANY KIND, EITHER EXPRESSED OR IMPLIED, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE. THE ENTIRE RISK AS TO THE QUALITY AND PERFORMANCE OF THE PROGRAM IS WITH YOU. SHOULD THE PROGRAM PROVE DEFECTIVE, YOU ASSUME THE COST OF ALL NECESSARY SERVICING, REPAIR OR CORRECTION.
```

**Comment by gioman** [comment]
> posting a dump (pg_dump in binary format) of a sample schema with all the tables in it would avoid this problems and make testers life easier...

@haubourg I think empathy broke down at this point, when I seemingly asked too much. I don''t think that ask users to help us triage/test/troubleshoot is too much. There certainly was no need to answer me as in https://github.com/qgis/QGIS/pull/32575#issuecomment-597363304 I think that I'm usually very constructive (including this case) about finding the root cause of an issue report.

**Comment by mlebihan** [comment]
I agree that you hint with _gid_ solves the trouble, but my view relies on a table _desendettement_communes_2018_ that had a primary key already. 

```
CREATE TABLE public.temp_desendettement_communes_2018(
     codecommune character varying(5) COLLATE pg_catalog."default" NOT NULL,
     nomcommune character varying COLLATE pg_catalog."default" NOT NULL,
     capacitedesendettement numeric(3,0),
 	geom geometry(MultiPolygon,4326),
     CONSTRAINT temp_desendettement_communes_2018_pkey PRIMARY KEY (codecommune)
         USING INDEX TABLESPACE data
)
 WITH (
     OIDS = FALSE
)
TABLESPACE data;
 
CREATE VIEW view_desendettement_communes_2018 AS 
    d.codeCommune, d.nomCommune, d.capaciteDesendettement, c.geom FROM desendettement_communes_2018 d 
    JOIN communes_2019 c ON d.codeCommune = c.insee;

```
So the trouble comes from elsewhere. Even if a gid helps, you can see that in :
```
CREATE VIEW view_desendettement_communes_2018 AS 
   SELECT row_number() OVER () AS gid, d.codeCommune, d.nomCommune, d.capaciteDesendettement, c.geom FROM desendettement_communes_2018 d 
   JOIN communes_2019 c ON d.codeCommune = c.insee;
```

SELECT row_number() OVER () AS gid
creates a field gid
but doesn't promote it to a primary key.

gioman:
> The part :
> " _SELECT row_number() OVER () AS gid_ "
> creates a _gid_ field, yes,
> but doesn't promote it to a primary key.

(afaik) there are no columns declared as PK in views, you just need one that has a unique identifier (to please QGIS, so to speak).

mlebihan:
But the view has the codecommune field that is an unique identifier, because it's a primary key coming from a table and is unique.

When QGis sees in a view a numeric field gid with 1, 2, 3, 4 it works
But when it sees a string field with "01001", "01002", "01003", "01004", it doesn't ?
both are unique.

gioman:
> _codecommune_

@mlebihan it is a varchar column, I think it must be a number (integer or maybe even decimal).

mlebihan:
It's what I think too...
So there many unknown things

1) Why does a view need an unique numerical value, even if it not declared a PK,
2) While a table doesn't ?
3) Why a VARCHAR column can't be eligible ?
4) Does the numerical field has to be named "_gid_" (I guess not !).

Can the message be more explanative about the error encountered, instead of a "" message ?

I think this issue would need a debug session to find how the code really works and where it fails and for what reason. It has a strange behavior.

gioman:
> Why does a view need an unique numerical value, even if it not declared a PK,

it has always been like that... but let's clarify again: this is affecting just views and **just** in DB Manager, a tool that we are aware is (unfortunately( riddled with problems and inconsistencies. You can add your view with the QGIS browser and the "add postgis layer" dialog.



> Does the numerical field has to be named "_gid_" (I guess not !).

no.

> I think this issue would need a debug session to find how the code really works and where it fails and for what reason. It has a strange behavior.

the good news is that DB manager is Python, so there is a wider possibility that someone can look at the code and find what to fix/change.

nagyrobir:
I have encountered the same issue now in QGIS 3.24, i think it has something to do with how are the views defined. I cannot drag and drop the layers through the browser in case of method 1. In case of method 2 works like a charm.

`Select geom , id from public.”PointsOfInterest”` Will create an invalid view,
`Select  id, geom from public.”PointsOfInterest”` Will create a perfectly valid view.

PR #32581, opened 2019-11-02
Make sure Python console text size can't be accidentally changed 

### LINKED ISSUE #26116: Python console text size bug
Author Name: **Stephen Knox** (Stephen Knox)
Original Redmine Issue: [18225](https://issues.qgis.org/issues/18225)
Affected QGIS version: 3.0.0
Redmine category:gui

---
When you select the cursor in the Python console and press Ctrl -> - , the text size shrinks. However, when you remain in the console and click Ctrl -> +, the text size does not get bigger, but it does zoom the map. I think the only way to reset the size is to restart QGIS.
Bit of a minor issue as it can be worked around, but annoying nonetheless

stev-0:
Fixes #26116

Previously when pressing Control -> - in the Scintilla Python console the text would get smaller and there was no way of making it bigger

### DIFF
--- a/python/console/console_sci.py
+++ b/python/console/console_sci.py
@@ -428,6 +428,10 @@ class ShellScintilla(QsciScintilla, code.InteractiveInterpreter):
             sys.stdout.fire_keyboard_interrupt = True
             return
 
+        #prevent accidental change in text size
+        if e.modifiers() & (Qt.ControlModifier) and e.key() in (Qt.Key_Plus, Qt.Key_Minus):
+            return
+
         line, index = self.getCursorPosition()
         cmd = self.text(line)
 


m-kuhn:
Thanks for looking into this.
Did you check if it would be possible to redirect Ctrl + '+' into the editor as well? Possibly also affecting the size of the output widget (and not just the input widget)?

stev-0:
@m-kuhn Not entirely sure what you mean, but if I understand correctly you are asking if the current (pre-pull request) situation is that a Ctrl + '+' has any effect on the Python console widgets, as well as Ctrl + '-' as already detailed?

The answer to that is no, the reason I noticed was because Ctrl + '-' made the console input text smaller but Ctrl + '+' did not make it bigger again.

Note this only occurs on the **input** window, not the output window, and only when the focus is on that window.

Now, with the pull request the strange thing that happens which I admit I can't really explain is:

- when you are focused on the map widget, Ctrl + '-' and Ctrl + '+' function as zoom as you would expect
- when you are focused on the input python console, Ctrl + '-' has no effect, and Ctrl + '+' zooms into the map window.

However, as this is recoverable, this is not as bad as it was before, where the worst case scenario was you got an unreadable console input font and had to restart QGIS

m-kuhn:
I can confirm the strange effect that currently ctrl+'+' affects the map widget while ctrl+'-' affects the input, while output is completely unaffected.

I could see advantage in the behavior that ctrl+'+' and ctrl+'-' would both work on input and output. E.g. in trainings it's sometimes nice to be able to quickly change the font size. So I wondered if you looked into what would be required fix ctrl+'+'.

If you won't spend more time on this, no worries and we can merge this pull request.

stev-0:
OK, I see what you mean.

I have been able to control the size in the input window only by something like the following:
```python
font = self.lexer.defaultFont(0)
  if e.modifiers() & (Qt.ControlModifier) and e.key() == Qt.Key_Minus:
  font.setPointSize(font.pointSize()-2) # or whatever value you want to decrement by
  self.lexer.setDefaultFont(font)
   self.lexer.setFont(font, -1)
   return
```

However, I can't do the same for Qt.Key_Plus. Which makes me think there is some sort of underlying bug there which was causing the odd behaviour in the first place. Not knowing a lot about QScintilla, think I will just add a similar fix to `console_editor.py` which suffers from the same issue as `console_sci.py` and leave it at that.

stev-0:
On second thoughts, might be worth some additional investigation. Control + '+' and Control + '-' keys on the numeric keypad are working fine, but control + '=/+' (on my keyboard) is being blocked somewhere else by QGIS I guess as the event isn't showing up at all. I will try and do some further investigations

m-kuhn:
https://stackoverflow.com/questions/40691389/how-to-capture-ctrlplus-in-qt ?

stev-0:
yeah I saw that, strangely this seems to be the opposite problem, i.e. the event from the numeric keypad appears, but the event  from the normal keyboard doesn't. I presume it's something in the main QGIS code which is consuming the event / not propagating it even when the focus is not on the map window.

m-kuhn:
On your keyboard, you press a modifier combination (like shift + ctrol + some key) to get the +?
Then it's exactly this.

stev-0:
I changed the line of my code above to
`if e.modifiers() and Qt.ControlModifier != 0 and e.key() == Qt.Key_Plus:`
and the keypress event still doesn't fire..

python/console/console_sci.py
```diff
@@ -428,6 +428,10 @@ def keyPressEvent(self, e):
             sys.stdout.fire_keyboard_interrupt = True
             return
 
+        #prevent accidental change in text size
+        if e.modifiers() & (Qt.ControlModifier) and e.key() in (Qt.Key_Plus, Qt.Key_Minus):
+            return
```
m-kuhn: Can you try to add a debug message in here (before the return) to see if the plus is caught as well?

stev-0:
I might close this and the bug, as I've discovered that although Shift +/- do not really work on the standard keyboard, they do work on the numeric keyboard, and the mouse, so although this is odd, not totally sure of the value

stev-0:
Closing due to above, will also request close of bug

m-kuhn:
Couldn't this PR be adjusted to call zoomIn() and zoomOut() instead of return?

Issue #32404, opened 2019-10-25
QGIS Server Font Directory


StefanNiemeier:
QGIS / QGIS Server 3.8.3  on Apache on Windows2012.

Theres a difference between displaying fonts in QGIS and displaying fonts from WMS by QGIS Server.
For example ARIAL in QGIS is not ARIAL by QGIS Server.

Even setting the environment variable QT_QPA_FONTDIR in httpd_qgis.conf doesn't work. The qgis-server-font-loader Plugin (https://github.com/avinet/qgis-server-font-loader) has also no effect.

Seems to be a bug.

Regards,
Stefan Niemeier

StefanNiemeier:
Update:
Also in the new 3.10.0 QGIS/QGIS Server on Windows10 there is no effect by setting the QT_QPA_FONTDIR in the httpd_qgis.conf. 
I had thougth that the solution by #31743 solved the problem but it's still the same.

**My httpd_qgis.conf:**
![httpd_qgis conf](https://user-images.githubusercontent.com/18546726/67631705-f247fe80-f899-11e9-90e7-8eed8ad05893.PNG)

**The qgis_server.log:**
[qgis_server.log](https://github.com/qgis/QGIS/files/3775625/qgis_server.log)

**The Screenshots:**
Text in QGIS Project: (ARIAL):
![QGIS_Beschriftung](https://user-images.githubusercontent.com/18546726/67631743-8dd96f00-f89a-11e9-8602-4dae4ecf5f6d.PNG)

Text by QGIS Server WMS:
![QGIS_Server_WMS_Beschriftung](https://user-images.githubusercontent.com/18546726/67631783-f3c5f680-f89a-11e9-91ac-950f6efe9609.PNG)

Text combined:
![zusammen](https://user-images.githubusercontent.com/18546726/67631791-15bf7900-f89b-11e9-9ed8-5d9be4e03ac0.PNG)

Livava1:
Same problem here: 3.10.2 QGIS/QGIS Server on Windows Server 2016 / Apache24 uses only default font in WMS. When project files from earlier versions have points symbolised with Font marker this is a show-stopper. The correct fonts are displayed in QGIS Desktop on the same server.

Another example: Setting _ITEMFONTFAMILY=Courier_ in GetLegendGraphic-requests has no effect.

Livava1:
[...]

Issue #30270, opened 2019-06-19
QGIS Server: Wrong Font display


freedom7979:
**Describe the bug**

In QGIS Server, fonts are poorly displayed on Windows.
This is how it looks in QGIS  - font Times New Roman
![Label in QGIS](https://user-images.githubusercontent.com/45253533/59769266-12413300-92a6-11e9-8267-c554c1298834.jpg)
This is how it looks label from QGIS Server by WMS  - font is unknown for my and wrong char "ě"
![Label from QGIS Server by WMS](https://user-images.githubusercontent.com/45253533/59769265-12413300-92a6-11e9-937b-c3760c48b719.jpg)

**How to Reproduce**
1. Create virtual directory on IIS with setting in 

> c:\Windows\System32\inetsrv\Config\applicationHost.config
```
<add name="QGISServer" autoStart="true" enable32BitAppOnWin64="false" managedRuntimeVersion="">
<processModel identityType="SpecificUser" userName="user" password="pass" />
    <environmentVariables>
        <add name="PATH" value="C:/OSGeo4W64/apps/Python37;C:/OSGeo4W64/apps/Python37/Scripts;c:/OSGeo4W64/bin/gdalplugins;c:/OSGeo4W64/apps/Qt5/bin;C:/OSGeo4W64/bin;C:/OSGeo4W64/apps/qgis/bin;C:/OSGeo4W64/apps/grass/grass-@grassversion@/bin;C:/OSGeo4W64/apps/grass/grass-@grassversion@/lib;C:/WINDOWS/system32;C:/WINDOWS;C:/WINDOWS/System32/Wbem" />
        <add name="QT_QPA_FONTDIR" value="c:/Windows/Fonts" />
        <add name="QGIS_PREFIX_PATH" value="C:/OSGeo4W64/apps/qgis" />
        <add name="QT_PLUGIN_PATH" value="C:/OSGeo4W64/apps/qgis/qtplugins;C:/OSGeo4W64/apps/qt5/plugins" />
        <add name="TEMP" value="C:/Users/Svoboda/AppData/Local/Temp" />
        <add name="PYTHONHOME" value="C:/OSGeo4W64/apps/Python37" />
        <add name="PYTHONIOENCODING" value="UTF-8" />
        <add name="GDAL_DATA" value="C:/OSGeo4W64/share/gdal" />
        <add name="GDAL_DRIVER_PATH" value="C:/OSGeo4W64/bin/gdalplugins" />
[...]

Issue #29244, opened 2019-02-28
QGIS Server Labels character problem


qgib:
Author Name: **Bela Kutfej** (@kutfejbela)
Original Redmine Issue: [21427](https://issues.qgis.org/issues/21427)
Affected QGIS version: 3.4.5
Redmine category:qgis_server

---
I updated QGIS Server from "http://qgis.org/debian-ltr stretch" and after that, the labels in WMS:GetMap result does not show the "ő" (hungarian) character.

---

- [qgis.server-wms.character.problem.png](https://issues.qgis.org/attachments/download/14481/qgis.server-wms.character.problem.png) (Bela Kutfej)

kutfejbela:
2 weeks ago the QGIS server shows this message:
"QFontDatabase: Cannot find font directory /usr/lib/x86_64-linux-gnu/fonts.
Note that Qt no longer ships fonts. Deploy some (from http://dejavu-fonts.org for example) or switch to fontconfig."
After I run the "ln -s /usr/share/fonts/truetype/*/* /usr/lib/x86_64-linux-gnu/fonts" command the characters are appears correctly.

RobiFag:
The same problem occurs with other labels style options like buffer and placement. If I set a buffer or change the label placement using an offset value, Qgis desktop shows labels correctly, but Qgis server does not show the style options applied.
We are using Qgis server 3.4.14-2 with Lizmap as web client.

Pedro-Murteira:
@kutfejbela Hello, is this still valid on more recent releases?

Issue #27788, opened 2018-09-27
Cyrillic symbols in QGIS Server


qgib:
Author Name: **Alexander Novikov** (@novikov-alexander)
Original Redmine Issue: [19966](https://issues.qgis.org/issues/19966)
Affected QGIS version: 3.2.3
Redmine category:qgis_server

---
I have a project with Cyrillic attributes and label values. It works perfect in QGIS Desktop.
But it seems not working in QGIS Server. There are empty squares instead letters in labels. Rule-based labeling also doesn't work.
I've already tried what I could. I've changed system encoding, I've changed encoding in Apache config. Nothing helps me.

---

- [bad.PNG](https://issues.qgis.org/attachments/download/13375/bad.PNG) (Alexander Novikov)

jgrocha:
Related to #29244 and #30270

PR #32588, opened 2019-11-02
WIP: QGIS Server GetServerSettings request

jgrocha:
## Description (update)

This PR implements a new request `GetServerSettings` to provide valuable feedback to tune QGIS Server deployment.

This request is not enabled by default. Only works if `QGIS_REVEAL_SERVER_SETTINGS` is defined with true or 1. This can be done, for example, in the Apache configuration with:
```
FcgidInitialEnv QGIS_REVEAL_SERVER_SETTINGS 1
```
## Output (JSON encoded)

The output is divided in four sections:
```json
{
    "about": {
    },
    "fonts": [
    ],
    "plugins": {
    },
    "settings": {
   }
}
```
### About contents
The `about` section contains more or less the same contents as the regular QGIS about dialogue: QGIS version, QT version and so on.
```json
    "about": {
        "Available memory": "1191 Mb",
        "Compiled against GDAL/OGR": "3.0.1",
        "Compiled against GEOS": "3.24.0",
        "Compiled against PROJ": "6.1.0",
        "Compiled against Qt": "5.11.1",
        "Cores": 8,
        "Hostname": "zoe",
        "Kernel version": "4.18.0-25-generic",
        "OS Version": "Ubuntu 18.10",
        "PostgreSQL Client Version": "11.4 (Ubuntu 11.4-1.pgdg18.10+1)",
        "QGIS code revision": "https://github.com/qgis/QGIS/commit/b28aa902c4",
        "QScintilla2 Version": "2.10.4",
        "QWT Version": "6.1.3",
        "Running against GDAL/OGR": "3.0.1",
        "Running against GEOS": "3.24.0",
        "Running against PROJ": "Rel. 6.1.0, May 15th, 2019",
        "Running against Qt": "5.11.1",
        "SpatiaLite Version": "4.3.0a",
        "System Memory": "15464 Mb",
        "Version": "QGIS Server - 3.11.0-Master ('Master')"
    }
```
There are a few additional entries that might help DevOps/SysOps: the number of available cores and memory. The available memory is only displayed on Linux servers (code is protected by `#ifdef Q_OS_LINUX`).
```json
    "about": {

        "Cores": 8,
        "Available memory": "1191 Mb",
        "System Memory": "15464 Mb",

    }
```
### Font contents
The prevent a large response, only the font families and the available styles are included. The sizes of each style is omitted (still in the code, but commented).

```json
"fonts": [
        {
            "family": "Libertinus Sans",
            "styles": [
                { "style": "Italic" },
                { "style": "Regular" },
                { "style": "Bold" }
            ]
        },
```
### Plugins contents
If plugins are loaded, they will be listed here.
```json
    "plugins": {
        "enabled": [
            "ServerSimpleBrowser"
        ]
    },
```
### Settings contents
This section contains all the settings defined on the server instance.
```json
"settings": {
        "ini": "/.local/share/QGIS/QGIS3/profiles/default/QGIS/QGIS3.ini",
        "summary": [
            {
                "description": "Override the default path for user configuration",
                "key": "",
                "setting": "QGIS_OPTIONS_PATH",
                "source": "DEFAULT_VALUE",
                "value": ""
            },
            {
                "description": "Activate/Deactivate parallel rendering for WMS getMap request",
                "key": "/qgis/parallel_rendering",
                "setting": "QGIS_SERVER_PARALLEL_RENDERING",
                "source": "DEFAULT_VALUE",
                "value": "false"
            },
            {
                "description": "Number of threads to use when parallel rendering is activated",
                "key": "/qgis/max_threads",
                "setting": "QGIS_SERVER_MAX_THREADS",
                "source": "DEFAULT_VALUE",
                "value": "-1"
            },
            {
                "description": "Log level",
                "key": "",
                "setting": "QGIS_SERVER_LOG_LEVEL",
                "source": "ENVIRONMENT_VARIABLE",
                "value": "0"
            },
            {
                "description": "Log file",
                "key": "",
                "setting": "QGIS_SERVER_LOG_FILE",
                "source": "ENVIRONMENT_VARIABLE",
                "value": "/var/log/qgis/qgisserver.log"
            },
            {
                "description": "Activate/Deactivate logging to stderr",
                "key": "",
                "setting": "QGIS_SERVER_LOG_STDERR",
                "source": "DEFAULT_VALUE",
                "value": "false"
            },
            {
                "description": "QGIS project file",
                "key": "",
                "setting": "QGIS_PROJECT_FILE",
                "source": "DEFAULT_VALUE",
                "value": ""
            },
            {
                "description": "Specify the maximum number of cached layers",
                "key": "",
                "setting": "MAX_CACHE_LAYERS",
                "source": "DEFAULT_VALUE",
                "value": "100"
            },
            {
                "description": "Specify the cache directory",
                "key": "/cache/directory",
                "setting": "QGIS_SERVER_CACHE_DIRECTORY",
                "source": "DEFAULT_VALUE",
                "value": "//.local/share/QGIS/QGIS3/profiles/default/cache"
            },
            {
                "description": "Specify the cache size",
                "key": "/cache/size",
                "setting": "QGIS_SERVER_CACHE_SIZE",
                "source": "DEFAULT_VALUE",
                "value": "52428800"
            },
            {
                "description": "Show group (thousands) separator",
                "key": "/locale/showGroupSeparator",
                "setting": "QGIS_SERVER_SHOW_GROUP_SEPARATOR",
                "source": "DEFAULT_VALUE",
                "value": "false"
            },
            {
                "description": "Override system locale",
                "key": "/locale/userLocale",
                "setting": "QGIS_SERVER_OVERRIDE_SYSTEM_LOCALE",
                "source": "DEFAULT_VALUE",
                "value": ""
            },
            {
                "description": "Maximum height for a WMS request. The lower one of this and the project configuration is used.",
                "key": "/qgis/max_wms_height",
                "setting": "QGIS_SERVER_WMS_MAX_HEIGHT",
                "source": "DEFAULT_VALUE",
                "value": "-1"
            },
            {
                "description": "Maximum width for a WMS request. The most conservative between this and the project one is used",
                "key": "/qgis/max_wms_width",
                "setting": "QGIS_SERVER_WMS_MAX_WIDTH",
                "source": "DEFAULT_VALUE",
                "value": "-1"
            },
            {
                "description": "Base directory where HTML templates and static assets (e.g. images, js and css files) are searched for",
                "key": "/qgis/server_api_resources_directory",
                "setting": "QGIS_SERVER_API_RESOURCES_DIRECTORY",
                "source": "DEFAULT_VALUE",
                "value": "/home/jgr/dev/cpp/build-QGIS/output/data/resources/server/api"
            },
            {
                "description": "Maximum value for \"limit\" in a features request, defaults to 10000",
                "key": "/qgis/server_api_wfs3_max_limit",
                "setting": "QGIS_SERVER_API_WFS3_MAX_LIMIT",
                "source": "DEFAULT_VALUE",
                "value": "10000"
            },
            {
                "description": "Enable/disable the GetServerSettings QGIS Server request",
                "key": "/qgis/reveal_server_settings",
                "setting": "QGIS_REVEAL_SERVER_SETTINGS",
                "source": "ENVIRONMENT_VARIABLE",
                "value": "1"
            }
        ]
```
## Discussion
I didn't changed the logic to route this request to the WMS service. This requires that a project is passed as a parameter. To issue a GetServerSettings request, a project must be present (or a default project defined).

**Valid request:**

http://localhost/cgi-bin/qgis_mapserv.fcgi?MAP=/home/qgis/projects/world.qgz&SERVICE=WMS&VERSION=1.3.0&REQUEST=GetServerSettings

**Invalid request** (without the `MAP` parameter and no default project defined on the configuration):

http://localhost/cgi-bin/qgis_mapserv.fcgi?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetServerSettings

The server will complain with:
```xml
<ServerException>Project file error</ServerException>
```
## Full example
On my development setup, the full GetServerSettings response is this 
[qgis_mapserv.fcgi.json.zip](https://github.com/qgis/QGIS/files/3827489/qgis_mapserv.fcgi.json.zip).

## Description (old)

When preparing projects for QGIS Server, we need to know which fonts are available on the server side.

There are also people complaining (like #30270, #32404, #29244, #27788) about the differences between fonts on client side and server side. With this PR, those issues are easier to debug and solve. 

This PR enables to user to know the fonts on the server side. Users will know which fonts they have to use to render properly on the server.

For sysadmins, this helps them to know which fonts they have to install on the server, if required by the users.

Geoserver, for example, also provides the available fonts on the web administration interface.

## Implementation

This PR adds the server supported fonts to the `GetProjectSettings` output. The fonts are provided by [QT QFontDatabase](https://doc.qt.io/qt-5/qfontdatabase.html) class.

GetProjectSettingsWithFonts

### Fonts for QGIS Server requests

Besides the fonts used in the project (for labels, layouts, and so on), this list is also useful to know the exact names of the fonts to use in request like `GetLegendGraphic`, in parameters like `ITEMFONTFAMILY=Century Schoolbook L`, as shown in the following screen capture.

ExactFontName

## Checklist

### DIFF
--- a/src/server/qgsserversettings.h
+++ b/src/server/qgsserversettings.h
@@ -66,7 +66,8 @@ class SERVER_EXPORT QgsServerSettingsEnv : public QObject
       QGIS_SERVER_WMS_MAX_HEIGHT, //! Maximum height for a WMS request. The most conservative between this and the project one is used (since QGIS 3.8)
       QGIS_SERVER_WMS_MAX_WIDTH, //! Maximum width for a WMS request. The most conservative between this and the project one is used (since QGIS 3.8)
       QGIS_SERVER_API_RESOURCES_DIRECTORY, //! Base directory where HTML templates and static assets (e.g. images, js and css files) are searched for (since QGIS 3.10).
-      QGIS_SERVER_API_WFS3_MAX_LIMIT //! Maximum value for "limit" in a features request, defaults to 10000 (since QGIS 3.10).
+      QGIS_SERVER_API_WFS3_MAX_LIMIT, //! Maximum value for "limit" in a features request, defaults to 10000 (since QGIS 3.10).
+      QGIS_REVEAL_SERVER_SETTINGS //! Enables QGIS Server legacy request GetServerSettings
     };
     Q_ENUM( EnvVar )
 };
@@ -114,6 +115,11 @@ class SERVER_EXPORT QgsServerSettings
     void logSummary() const;
 
     /**
+     * Log a summary of settings currently loaded in JSON format
+     */
+    QJsonObject logSummaryJson() const;
+
+    /**
      * Returns the ini file loaded by QSetting.
      * \returns the path of the ini file or an empty string if none is loaded.
      */
@@ -124,6 +130,12 @@ class SERVER_EXPORT QgsServerSettings
      * \returns TRUE if parallel rendering is activated, FALSE otherwise.
      */
     bool parallelRendering() const;
+
+    /**
+     * Returns reveal server settings setting.
+     * \returns TRUE if reveal server settings is enabled, FALSE otherwise.
+     */
+    bool revealServerSettings() const;
 
     /**
      * Returns the maximum number of threads to use.
--- a/src/server/services/wms/qgswmsgetserversettings.cpp
+++ b/src/server/services/wms/qgswmsgetserversettings.cpp
@@ -19,18 +19,18 @@
 #include "qgswmsrenderer.h"
 #include "qgswmsserviceexception.h"
 #include "qgsserverplugins.h"
-#include "ogr_core.h"
-#include "sqlite3.h"
-#include "geos_c.h"
-#include "pg_config.h"
-#include "spatialite.h"
-#include "qwt/qwt_global.h"
+#include <ogr_core.h>
+#include <sqlite3.h>
+#include <geos_c.h>
+#include <pg_config.h>
+#include <spatialite.h>
+#include <qwt/qwt_global.h>
 #if PROJ_VERSION_MAJOR>=6
 #include <proj.h>
 #else
 #include <proj_api.h>
 #endif
-#include "qgis.h"
+#include <qgis.h>
 #include <QThread>
 #include <QFontDatabase>
 #ifdef Q_OS_LINUX
--- a/src/server/services/wms/qgswmsgetserversettings.cpp
+++ b/src/server/services/wms/qgswmsgetserversettings.cpp
@@ -25,7 +25,7 @@
 #include "pg_config.h"
 #include "spatialite.h"
 #include "qwt/qwt_global.h"
-#include "proj.h"
+#include <proj.h>
 #include "qgis.h"
 #include <QThread>
 #include <QFontDatabase>
--- a/src/server/services/wms/qgswmsgetserversettings.cpp
+++ b/src/server/services/wms/qgswmsgetserversettings.cpp
@@ -25,7 +25,11 @@
 #include "pg_config.h"
 #include "spatialite.h"
 #include "qwt/qwt_global.h"
+#if PROJ_VERSION_MAJOR>=6
 #include <proj.h>
+#else
+#include <proj_api.h>
+#endif
 #include "qgis.h"
 #include <QThread>
 #include <QFontDatabase>
--- a/src/server/services/wms/qgswmsgetserversettings.cpp
+++ b/src/server/services/wms/qgswmsgetserversettings.cpp
@@ -135,6 +135,7 @@ namespace QgsWms
     QJsonObject serverSettings;
     serverSettings["about"] = about();
     serverSettings["settings"] = serverIface->serverSettings()->logSummaryJson();
+    serverSettings["fonts"] = availableFonts();
 
     QJsonObject serverPlugins;
     QJsonArray availablePlugins = {};
@@ -144,8 +145,6 @@ namespace QgsWms
     serverPlugins["enabled"] = availablePlugins;
     serverSettings["plugins"] = serverPlugins;
 
-    serverSettings["settings"] = serverIface->serverSettings()->logSummaryJson();
-    serverSettings["fonts"] = availableFonts();
     QJsonDocument document( serverSettings );
 
     response.setHeader( QStringLiteral( "Content-Type" ), QStringLiteral( "application/json" ) );
--- a/src/server/services/wms/qgswmsgetserversettings.cpp
+++ b/src/server/services/wms/qgswmsgetserversettings.cpp
@@ -30,7 +30,7 @@
 #else
 #include <proj_api.h>
 #endif
-#include <qgis.h>
+#include "qgis.h"
 #include <QThread>
 #include <QFontDatabase>
 #ifdef Q_OS_LINUX
--- a/src/server/services/wms/qgswmsgetserversettings.cpp
+++ b/src/server/services/wms/qgswmsgetserversettings.cpp
@@ -1,0 +1,154 @@
+/***************************************************************************
+                              qgswmsgetserversettings.cpp
+                              ---------------------------
+  begin                : Nov 7, 2019
+  copyright            : (C) 2019 by Jorge Gustavo Rocha
+  email                : jgr at geomaster dot pt
+ ***************************************************************************/
+
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+#include "qgswmsutils.h"
+#include "qgswmsgetprint.h"
+#include "qgswmsrenderer.h"
+#include "qgswmsserviceexception.h"
+#include "qgsserverplugins.h"
+#include "ogr_core.h"
+#include "sqlite3.h"
+#include "geos_c.h"
+#include "pg_config.h"
+#include "spatialite.h"
+#include "qwt/qwt_global.h"
+#include "proj.h"
+#include "qgis.h"
+#include <QThread>
+#include <QFontDatabase>
+#ifdef Q_OS_LINUX
+#include <unistd.h>
+#endif
+
+namespace QgsWms
+{
+
+  QJsonArray availableFonts()
+  {
+    QJsonArray availableFamilies = {};
+    QFontDatabase database;
+    const QStringList fontFamilies = database.families();
+    for ( const QString &family : fontFamilies )
+    {
+      QJsonObject thisFamily = {};
+      thisFamily["family"] = family;
+      QJsonArray familyStyles = {};
+      const QStringList fontStyles = database.styles( family );
+      for ( const QString &style : fontStyles )
+      {
+        QJsonObject thisStyle = {};
+        thisStyle["style"] = style;
+        // uncomment to provide all font sizes available
+        // QJsonArray styleSizes = {};
+        // QString sizes;
+        // const QList<int> smoothSizes = database.smoothSizes( family, style );
+        // for ( int points : smoothSizes )
+        // {
+        //   styleSizes.append( points );
+        // }
+        // thisStyle["sizes"] = styleSizes;
+        familyStyles.append( thisStyle );
+      }
+      thisFamily["styles"] = familyStyles;
+      availableFamilies.append( thisFamily );
+    }
+    return availableFamilies;
+  }
+
+  QJsonObject about()
+  {
+    QJsonObject aboutServer = {};
+    QString caption = QStringLiteral( "QGIS Server - %1 ('%2')" ).arg( Qgis::QGIS_VERSION, Qgis::QGIS_RELEASE_NAME );
+    aboutServer["Version"] = caption;
+    if ( QString( Qgis::QGIS_DEV_VERSION ) == QLatin1String( "exported" ) )
+    {
+      caption = QStringLiteral( "https://github.com/qgis/QGIS/tree/release-%1_%2" )
+                .arg( Qgis::QGIS_VERSION_INT / 10000 ).arg( Qgis::QGIS_VERSION_INT / 100 % 100 );
+      aboutServer["QGIS code branch"] = caption;
+    }
+    else
+    {
+      caption = QStringLiteral( "https://github.com/qgis/QGIS/commit/%2" ).arg( Qgis::QGIS_DEV_VERSION );
+      aboutServer["QGIS code revision"] = caption;
+    }
+    aboutServer["Compiled against Qt"] = QT_VERSION_STR;
+    aboutServer["Running against Qt"] = qVersion();
+
+    aboutServer["Compiled against GDAL/OGR"] = GDAL_RELEASE_NAME;
+    aboutServer["Running against GDAL/OGR"] = GDALVersionInfo( "RELEASE_NAME" );
+
+    aboutServer["Compiled against GEOS"] = GEOS_CAPI_VERSION;
+    aboutServer["Running against GEOS"] = GEOSversion();
+
+    aboutServer["Compiled against GEOS"] = SQLITE_VERSION;
+    aboutServer["Running against GEOS"] = sqlite3_libversion();
+
+#ifdef HAVE_POSTGRESQL
+    aboutServer["PostgreSQL Client Version"] = QStringLiteral( PG_VERSION );
+#else
+    aboutServer["PostgreSQL Client Version"] = QStringLiteral( "No support" );
+#endif
+
+    aboutServer["SpatiaLite Version"] = spatialite_version();
+    aboutServer["QWT Version"] = QWT_VERSION_STR;
+    aboutServer["QScintilla2 Version"] = QSCINTILLA_VERSION_STR;
+
+#if PROJ_VERSION_MAJOR > 4
+    PJ_INFO info = proj_info();
+    aboutServer["Compiled against PROJ"] = QStringLiteral( "%1.%2.%3" ).arg( PROJ_VERSION_MAJOR ).arg( PROJ_VERSION_MINOR ).arg( PROJ_VERSION_PATCH );
+    aboutServer["Running against PROJ"] = info.release;
+#else
+    aboutServer["PROJ.4 Version"] = QString::number( PJ_VERSION );
+#endif
+
+    aboutServer["OS Version"] = QSysInfo::prettyProductName();
+    aboutServer["Kernel version"] = QSysInfo::kernelVersion();
+    aboutServer["Hostname"] = QSysInfo::machineHostName();
+    aboutServer["CPU Cores"] = QThread::idealThreadCount();
+
+#ifdef Q_OS_LINUX
+    long pages = sysconf( _SC_PHYS_PAGES );
+    long pagesAvailable = sysconf( _SC_AVPHYS_PAGES );
+    long page_size = sysconf( _SC_PAGE_SIZE );
+    aboutServer["System Memory (Mb)"] = QStringLiteral( "%1" ).arg( pages * page_size / 1024 / 1024 );
+    aboutServer["Available memory (Mb)"] = QStringLiteral( "%1" ).arg( pagesAvailable * page_size / 1024 / 1024 );
+#endif
+
+    return aboutServer;
+  }
+
+  void writeServerSettings( QgsServerInterface *serverIface, QgsServerResponse &response )
+  {
+    QJsonObject serverSettings;
+    serverSettings["about"] = about();
+    serverSettings["settings"] = serverIface->serverSettings()->logSummaryJson();
+
+    QJsonObject serverPlugins;
+    QJsonArray availablePlugins = {};
+    QStringList plugins = QgsServerPlugins::serverPlugins();
+    for ( int i = 0; i < plugins.size(); ++i )
+      availablePlugins.append( plugins.at( i ) );
+    serverPlugins["enabled"] = availablePlugins;
+    serverSettings["plugins"] = serverPlugins;
+
+    serverSettings["settings"] = serverIface->serverSettings()->logSummaryJson();
+    serverSettings["fonts"] = availableFonts();
+    QJsonDocument document( serverSettings );
+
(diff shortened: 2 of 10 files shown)


elpaso:
Nice and useful! Thank you.

jgrocha:
I'll need to rewrite the `GetProjectSettings` test (that's why the WIP flag). Every computer running the tests will return a different set of fonts, so the test just need to check if the `<Font>` tag is there. The contents can not be compared line by line with an existent default, as we now have in `test_getprojectsettings`.

elpaso:
@jgrocha you can try to use `assertIn()` and put your font plus one of the fonts installed on Travis The question is how to know what they are, either you manage to replicate the Travis environment locally, (that is possible but not easy to do) or you print them when the test fails and you make a Travis run to see them.

elpaso:
Just one question here, the fonts installed on a system could potentially be many, did you consider/estimate the additional overhead in terms of XML size?

jgrocha:
> Just one question here, the fonts installed on a system could potentially be many, did you consider/estimate the additional overhead in terms of XML size?

@elpaso Each font takes about 250 bytes. My 270 fonts increased the response from 27 kb to 93 kb. With 500 fonts on a server, it would make the response about 125 kb. 1000 fonts would create a response of about 250 kb. It is not a request people would request often.

We can make it optional, if you think the response size can be a problem.

m-kuhn:
Nice tool!
Is `GetProjectSettings` the right place to add this to?
Or should it be a `GetServerInformation` request (which could be disabled in production)?

andreasneumann:
Yes - I would be in favour of having a separate GetServerInformation or GetServerStatus for such information. It would be really useful. But it also exposes information not everybody out there needs/should know.

F.e. I would be interested to see what projects and layers are in cache, system resources consumed by server, maybe even get statistics or access to logfile, etc.

But really useful to see what fonts are available!

jgrocha:
Thank you @m-kuhn for the valuable feedback. I'll start to work on a `GetServerInformation` request, that can be enabled by setting an environment variable in the server configuration. It's the right place to put this list of fonts and enable other features, as suggested by @andreasneumann.

jgrocha:
Since WFS 3/OGC API is no longer XML based, we will start/already started to move from the XML to JSON. May I use JSON as the output format of this new `GetServerInformation` request?

andreasneumann:
I think JSON would be fine. It is "proprietary QGIS" anyway.

The credits for the GetServerInformation request idea should go to @m-kuhn though. He suggested it first, I just agreed that it would be a good idea, and it probably will be extended with additional information soon. I am pretty sure that others will find it useful.

Thanks for working on this!

pblottiere:
Hi @jgrocha,

> This PR enables to user to know the fonts on the server side. Users will know which fonts they have to use to render properly on the server.

Nice, good idea :+1:!

> Or should it be a `GetServerInformation` request (which could be disabled in production)?

+1

> May I use JSON as the output format of this new `GetServerInformation` request?

+1

src/server/qgsserversettings.h
```diff
@@ -66,7 +66,8 @@ class SERVER_EXPORT QgsServerSettingsEnv : public QObject
       QGIS_SERVER_WMS_MAX_HEIGHT, //! Maximum height for a WMS request. The most conservative between this and the project one is used (since QGIS 3.8)
       QGIS_SERVER_WMS_MAX_WIDTH, //! Maximum width for a WMS request. The most conservative between this and the project one is used (since QGIS 3.8)
       QGIS_SERVER_API_RESOURCES_DIRECTORY, //! Base directory where HTML templates and static assets (e.g. images, js and css files) are searched for (since QGIS 3.10).
-      QGIS_SERVER_API_WFS3_MAX_LIMIT //! Maximum value for "limit" in a features request, defaults to 10000 (since QGIS 3.10).
+      QGIS_SERVER_API_WFS3_MAX_LIMIT, //! Maximum value for "limit" in a features request, defaults to 10000 (since QGIS 3.10).
+      QGIS_REVEAL_SERVER_SETTINGS //! Enables QGIS Server legacy request GetServerSettings
```
elpaso: Can you please rename it to something that starts with QGIS_SERVER_ like QGIS_SERVER_ENABLE_GETSERVERSETTINGS?

elpaso:
Actually, looking to the future of OGC services and to the fact that we may want to add more server information to this call, I was thinking that this information could be more useful if abstracted out to it's own class instead of limiting it to WMS (or WFS), so I think you could create a new `QgsServerInformation` class (just because `QgsServerSettings` is taken)  that constructs the JSON and exposes it through  the new `QgsServerApi` interface by creating a new endpoint (something like `serversettings`) that returns JSON information. This doesn't exclude that you can also return the same information from a  `getServerSettings` request in WMS.

For what concerns the code, I would recommend you using the new (and a little bit faster) json library that we are using in all the new OAPIF (WFS3) API and in the core ( `nlohmann::json` ) instead of QJson classes.

src/server/qgsserversettings.h
m-kuhn: What about `QGIS_SERVER_DEBUG_MODE` (assuming that it makes it clear that this is not for production)?

jgrocha: > What about `QGIS_SERVER_DEBUG_MODE` (assuming that it makes it clear that this is not for production)?

@m-kuhn Both option are good (thanks!), but maybe `QGIS_SERVER_ENABLE_GETSERVERSETTINGS` is less ambiguous. People might think that `QGIS_SERVER_DEBUG_MODE` has other side effects, like running in DEBUG MODE, instead of production mode. We already have `QGIS_DEBUG` and this `QGIS_SERVER_DEBUG_MODE` could be understood as the same, but for running `qgis_mapserv.fcgi`.

jgrocha:
> Actually, looking to the future of OGC services and to the fact that we may want to add more server information to this call, I was thinking that this information could be more useful if abstracted out to it's own class instead of limiting it to WMS (or WFS), so I think you could create a new `QgsServerInformation` class (just because `QgsServerSettings` is taken) that constructs the JSON and exposes it through the new `QgsServerApi` interface by creating a new endpoint (something like `serversettings`) that returns JSON information. This doesn't exclude that you can also return the same information from a `getServerSettings` request in WMS.
> 

Sounds like a good plan. To make sure I understood: I have to create an `QgsServerOgcApiHandler` instance (that can be called `QgsServerInformation`) to handle this new `serversettings` endpoint, right? The call to this request will be something like: `qgis_mapserv.fcgi/api/serversettings`, right?

If we implement it as an API call, we can drop the `GetServerSettings` request in WMS. No need to have it under different handlers.

> For what concerns the code, I would recommend you using the new (and a little bit faster) json library that we are using in all the new OAPIF (WFS3) API and in the core ( `nlohmann::json` ) instead of QJson classes.

Good! I'll check the `QgsJsonUtils` class, that already uses the `nlohmann::json`. If I can, I'll do it based on `QgsJsonUtils`.

src/server/qgsserversettings.h
m-kuhn: Good point. Let's just document this clearly with comments in all the examples that demonstrate setting up a server.

elpaso:
> > Actually, looking to the future of OGC services and to the fact that we may want to add more server information to this call, I was thinking that this information could be more useful if abstracted out to it's own class instead of limiting it to WMS (or WFS), so I think you could create a new `QgsServerInformation` class (just because `QgsServerSettings` is taken) that constructs the JSON and exposes it through the new `QgsServerApi` interface by creating a new endpoint (something like `serversettings`) that returns JSON information. This doesn't exclude that you can also return the same information from a `getServerSettings` request in WMS.
> 
> Sounds like a good plan. To make sure I understood: I have to create an `QgsServerOgcApiHandler` instance (that can be called `QgsServerInformation`) to handle this new `serversettings` endpoint, right? The call to this request will be something like: `qgis_mapserv.fcgi/api/serversettings`, right?
> 

Yes, that is one way, technically this call is not par of any OGC API, so my first thought was to recommend you to just use `QgsServerApi`, but there is a lot of functionality inside `QgsServerOgcApi / Handler` that make it probably easier for you to use that implementation.

In any event, please make sure you also provide an HTML template and use content negotiation to serve HTML or JSON according to the incoming request ("Accept" header), see the implementation in `QgsServerOgcApiHandler`, so, ideally you should just produce JSON, and use `QgsServerOgcApiHandler::write`, user using a browser will see a nice HTML page with all the information they need, other clients will get JSON.

src/server/services/wms/qgswmsgetserversettings.cpp
```diff
@@ -25,7 +25,7 @@
 #include "pg_config.h"
 #include "spatialite.h"
```
m-kuhn: You'll need to `<>` all which are not `qgs.....h`

jgrocha: Thank you @m-kuhn!

Gustry:
Very nice addition @jgrocha ! Thanks.
I had a quick look for now. I would add or switch to use the QGIS version as integer, instead of text. And also plugin as a dictionary with their version. But this is extra requests, feel free to skip, we can come back on that before the release of 3.12.

src/server/qgsserversettings.h
```diff
@@ -113,6 +114,11 @@ class SERVER_EXPORT QgsServerSettings
      */
     void logSummary() const;
 
+    /**
+     * Log a summary of settings currently loaded in JSON format
```
pblottiere:
```suggestion
     * Logs a summary of settings currently loaded in JSON format.
     * \since QGIS 3.12
```

src/server/qgsserversettings.h
```diff
@@ -125,6 +131,12 @@ class SERVER_EXPORT QgsServerSettings
      */
     bool parallelRendering() const;
 
+    /**
+     * Returns reveal server settings setting.
+     * \returns TRUE if reveal server settings is enabled, FALSE otherwise.
```
pblottiere:
```suggestion
     * \returns TRUE if reveal server settings is enabled, FALSE otherwise.
     * \since QGIS 3.12
```

src/server/services/wms/qgswmsgetserversettings.cpp
```diff
@@ -0,0 +1,157 @@
...
+  {
+    QJsonObject serverSettings;
+    serverSettings["about"] = about();
+    serverSettings["settings"] = serverIface->serverSettings()->logSummaryJson();
+    serverSettings["fonts"] = availableFonts();
+
+    QJsonObject serverPlugins;
+    QJsonArray availablePlugins = {};
+    QStringList plugins = QgsServerPlugins::serverPlugins();
+    for ( int i = 0; i < plugins.size(); ++i )
```
pblottiere: Do you think we could do something like `for ( const QString &plugin : plugins )` instead?

src/server/services/wms/qgswmsgetserversettings.h
```diff
@@ -0,0 +1,40 @@
...
+   */
+  QJsonArray availableFonts();
+
+  /**
+   * Creates Json version of the about
+   */
+  QJsonObject about();
+
+  /**
+   * Output GetPrint response
```
pblottiere: copy/paste from `qgswmsgetprint.h` :)?

src/server/services/wms/qgswmsgetserversettings.h
```diff
@@ -0,0 +1,40 @@
...
+namespace QgsWms
+{
+
+  /**
+   * Creates Json object with the available fonts
+   */
+  QJsonArray availableFonts();
+
+  /**
+   * Creates Json version of the about
```
pblottiere:
```suggestion
   * Creates Json version of the about
   * \since QGIS 3.12
```

src/server/services/wms/qgswmsgetserversettings.h
```diff
@@ -0,0 +1,40 @@
...
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+namespace QgsWms
+{
+
+  /**
+   * Creates Json object with the available fonts
```
pblottiere:
```suggestion
   * Creates Json object with the available fonts
   * \since QGIS 3.12
```

src/server/services/wms/qgswmsgetserversettings.cpp
m-kuhn: And in this case don't foget to `const QStringList plugins ...` on 146

nyalldawson:
This pr is stale and needs further work before is merge ready. @jgrocha please feel free to reopen when you have time to pick this up again.

m-kuhn:
This would be a very desirable functionality. @jgrocha if you need anything to continue, please don't hesitate to ask!

pblottiere:
Hello @jgrocha,

Considering the recent conversations on the dev mailing list, I'd like to dig a little deeper on this subject and I'm thinking about writing a QEP. 

Can I take the lead on this? Considering that you're the initial author of this PR/work, I'd like to have your approval before going any further :).

jgrocha:
> Hello @jgrocha,
> 
> Considering the recent conversations on the dev mailing list, I'd like to dig a little deeper on this subject and I'm thinking about writing a QEP.
> 
> Can I take the lead on this? Considering that you're the initial author of this PR/work, I'd like to have your approval before going any further :).

Yes, please. Thank you for taking the lead!

pblottiere:
> Yes, please. Thank you for taking the lead!

Thanks @jgrocha :+1:

andreasneumann:
It would be appreciated. It would help "mere mortals" like myself, with no debugging skills, to get a better idea on why certain things fail or not, or what the state of the server is.

Gustry:
Thanks @pblottiere !
This would be a nice addition !

elpaso:
Sorry to be the party breaker here, but I see this functionality the ideal candidate for a Python plugin implementation instead of a C++ core functionality.

Because:
1. this will never enabled by default for security reasons (too much information disclosure)
2. this is not targeted to end users but to who? Sysadmins? Integrators? Developers? "Mere mortals"? All these roles have access to the same information through a console. But I see the point for mere mortals (and similar roles, that are in the middle between all of the mentioned roles). 
3. if enabled it will probably go behind an authorization wall (that we don't currently have in the server implementation)
4. speed is not critical for this functionality

I see an OGC API based Python plugin the ideal way to implement this functionality. You could also much quickly implement this in Python, not to mention the lower barrier and the much larger code contributors base for a Python implementation.

andreasneumann:
> 2\. this is not targeted to end users but to who? Sysadmins? Integrators? Developers?

Definitely sysadmins are one class that should have access to this information. Currently, QGIS server is a bit of a blackbox to them.



> 3\. if enabled it will probably go behind an authorization wall (that we don't currently have in the server implementation)

Definitely. This is just too sensitive information and it shouldn't be revealed to the public

There should also be an easy way to enable/disable this functionality. I guess a Python plugin would be fine.

pblottiere:
Hi@elpaso,

> Sorry to be the party breaker here, but I see this functionality the ideal candidate for a Python plugin implementation instead of a C++ core functionality.

Well, it'll probably be the conclusion, but I still think it'd be interesting to identify needs. Because as pointed out by @andreasneumann , QGIS Server is indeed a blackbox for sysadmins.

However, I think that some actions cannot be done by a Python plugin, like reloading a cached project for example (at least for now).

elpaso:
> Hi@elpaso,
> 
> > Sorry to be the party breaker here, but I see this functionality the ideal candidate for a Python plugin implementation instead of a C++ core functionality.
> 
> Well, it'll probably be the conclusion, but I still think it'd be interesting to identify needs. Because as pointed out by @andreasneumann , QGIS Server is indeed a blackbox for sysadmins.
> 
> However, I think that some actions cannot be done by a Python plugin, like reloading a cached project for example (at least for now).

Was it part of this PR?

Anyway it's something we should allow in the Python API, I've also thought about it and I see the use case for DB store projects and for network filesystem stored project (and it's a quick fix too).

pblottiere:
> Was it part of this PR?

Oh no, it's not part of this PR, just a general need we have to talk about in a dedicated QEP I have to write :). But I think that this PR is the starting point to raise the main discussion (cf comments of Andreas, Matthias or Etienne). 

 > Anyway it's something we should allow in the Python API, I've also thought about it and I see the use case for DB store projects and for network filesystem stored project (and it's a quick fix too).

Totally agree. That's why I want to talk about it in a QEP in order to estimate the needs and assess the underlying work (and where/what we have to tweak).

elpaso:
Ok, let's see if we can make this QGIS Server developer online meeting, I think that we have definitely something to discuss. Perhaps we should first try to define the "mission" of the QGIS server project (what kind of product we want to realize).

jgrocha:
Thank you @pblottiere for converting this QGIS Server discussion into actions. Your QEP will be very welcome.

@elpaso Can you take the lead on this QGIS Server meeting and suggest a date/time? (Or use doodle or some equivalent so we can get the best option for the majority). It would be nice to draft a QGIS Server roadmap out of this meeting, based on the comments provided by all.

In the short term, besides Paul's QEP and probably a clear road map from the upcoming meeting, allowing things like this PR `GetServerSettings` does not hurt and can be replaced in the future with a proper QGIS Server (web) console or something else. Note: my point here is not specifically related with this PR. Is just to point out that sometimes we can go forward with small steps (since we don't have so many developers writing server's code), instead of a great/huge plan.

Looking forward for Paul's QEP and for QGIS Server meeting.

elpaso:
> @elpaso Can you take the lead on this QGIS Server meeting and suggest a date/time? (Or use doodle or some equivalent so we can get the best option for the majority). It would be nice to draft a QGIS Server roadmap out of this meeting, based on the comments provided by all.
> 

Sure, I've already started a discussion within the PSC if we are making it an official QGIS meeting or not and then we'll go public.

> In the short term, besides Paul's QEP and probably a clear road map from the upcoming meeting, allowing things like this PR `GetServerSettings` does not hurt and can be replaced in the future with a proper QGIS Server (web) console or something else. Note: my point here is not specifically related with this PR. Is just to point out that sometimes we can go forward with small steps (since we don't have so many developers writing server's code), instead of a great/huge plan.
> 

Absolutely agree: can you make this PR as a Python plugin? It should be quick.

If you need an example of an API developed with the new OGC API classes have a look to this files (the project itself is completely WIP and unfinished!!! so don't look at it): https://github.com/elpaso/qgis-server-standalone-docker/blob/getlegendgraphics/plugins/landingpage/handlers.

vpicavet:
Hi, late to the party, but shouldn't we use a REST API model now that OGC ( finally ) made the switch in new protocols ?

PR #32386, opened 2019-10-24
Use CRS in WMS GetFeatureInfo when info_format is in json Fixes #32326

pblottiere:
## Description

Fixes https://github.com/qgis/QGIS/issues/32326

PR #32596, opened 2019-11-03
Fix server OAPIF crs transformation on items

elpaso:
fixes an unreported issue when serving items not in
4326.

Cherry-picked from master 7d2000bc30.

### DIFF
--- a/src/server/services/wfs3/qgswfs3handlers.cpp
+++ b/src/server/services/wfs3/qgswfs3handlers.cpp
@@ -1077,8 +1077,7 @@ void QgsWfs3CollectionsItemsHandler::handleRequest( const QgsServerApiContext &c
     QgsJsonExporter exporter { mapLayer };
     exporter.setAttributes( featureRequest.subsetOfAttributes() );
     exporter.setAttributeDisplayName( true );
-    // Trick! See: https://github.com/qgis/QGIS/commit/7d2000bc30b2160692baf43c22d7ef4b47e52cff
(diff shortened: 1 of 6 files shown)


src/server/services/wfs3/qgswfs3handlers.cpp
```diff
@@ -1077,7 +1077,8 @@ void QgsWfs3CollectionsItemsHandler::handleRequest( const QgsServerApiContext &c
     QgsJsonExporter exporter { mapLayer };
     exporter.setAttributes( featureRequest.subsetOfAttributes() );
     exporter.setAttributeDisplayName( true );
-    exporter.setSourceCrs( mapLayer->crs() );
+    // Trick! See: https://github.com/qgis/QGIS/commit/7d2000bc30b2160692baf43c22d7ef4b47e52cff
+    exporter.setSourceCrs( QgsCoordinateReferenceSystem( 4326, QgsCoordinateReferenceSystem::EpsgCrsId ) );
```
nyalldawson: I don't understand this -- are you trying to disable transformation? if so, just setting the source crs to an invalid crs should work.

elpaso: Yes, that was the intent.

elpaso: In any event, this is a transitory solution, until this is backported: https://github.com/qgis/QGIS/pull/32386#issuecomment-549176578 and I can use an explicit call to `exporter.setTransformGeometries( false );`

pblottiere (approved):

PR #32602, opened 2019-11-03
Widespread use of AUTOMOC

m-kuhn:
Our two famous header lists (HDRS and MOC_HDRS) are not really user-friendly for new comers.
I'm sure many hours have been lost looking for obscure vtable or staticMetaObject missing problems.

Then there is also a conceptual issue with these lists. Or I'm not getting how it's supposed to work. The MOC list includes all MOC headers (QObject etc), the meaning of the other list is quite obscure. I was supposing it's a list of files that would be installed to the systems. But actually also the headers inside the MOC list are installed (mostly... except when they are not in some of our libs like analysis). There are also a lot of private headers listed which are not supposed to be installed.

Some of the modules also install MOC headers to the system. Some not.

Furthermore, we have a massive duplication of header files in there (not completely though, it seems to be kind of random which files have been duplicated and which not).

Automoc should at least solve part of this problem


m-kuhn:
Aparently automoc doesn't like our sip defines

nyalldawson:
Nice work!!

PR #32610, opened 2019-11-04
Another connect without context

nyalldawson:


src/gui/providers/ogr/qgsgeopackageitemguiprovider.cpp
```diff
@@ -94,7 +94,7 @@ void QgsGeoPackageItemGuiProvider::populateContextMenu( QgsDataItem *item, QMenu
     // Add table to existing DB
     QAction *actionAddTable = new QAction( tr( "Create a New Layer or Table…" ), collectionItem->parent() );
     QPointer<QgsGeoPackageCollectionItem>collectionItemPtr { collectionItem };
-    connect( actionAddTable, &QAction::triggered, [ collectionItemPtr ]
+    connect( actionAddTable, &QAction::triggered, actionAddTable, [ collectionItemPtr ]
```
nirvn: Out of curiosity, why not using the collectionItemPtr pointer instead of the actionAddTAble pointer here? In theory, isn't a signal sent to a null receiver harmless (in that it'll merely not send the signal)?

nyalldawson: I've pushed an improved version now!

nirvn:
@nyalldawson , improved it is indeed.

PR #32613, opened 2019-11-04
Raise a Python OwnershipException when trying to /Transfer/

nyalldawson:
an object which is already owned by some other c++ object

Try 2...

### DIFF
--- a/python/core/auto_generated/geometry/qgsmultipoint.sip.in
+++ b/python/core/auto_generated/geometry/qgsmultipoint.sip.in
@@ -38,8 +38,26 @@ Multi point geometry collection.
 
     virtual int nCoordinates() const;
 
-    virtual bool addGeometry( QgsAbstractGeometry *g /Transfer/ );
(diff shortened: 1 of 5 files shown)


nyalldawson:
@m-kuhn @wonder-sk @3nids @elpaso

I've no luck confirming this approach with upstream, can you see any potential issues with it? (Like all sip work, it's just random keyboard bashing until something compiles 🙃)

3nids:
I had a look to the code and as far as I can tell it looks good to me.
If we tend to reuse this, it would be nice to have a macro or use sipify to generate the code, since there seems to be only 2 variables (message and convert class).

nyalldawson:
> If we tend to reuse this, it would be nice to have a macro or use sipify to generate the code, since there seems to be only 2 variables (message and convert class).

I agree. Ideally, I'd partner this with a change in the c++ api to use a unique_ptr argument....

elpaso:
As far as I understand looks good, I'm looking forward to see how you'll handle unique_ptr, I thought about that in the past but I couldn't find a way to handle them with SIP.

PR #32618, opened 2019-11-04
More connect fun

nirvn:
## Description

I wanna play too! :wink: 

## Checklist


nirvn:
On basis of build failure, does this mean this QObject::connect(...) warning is a false positive?

m-kuhn:
> On basis of build failure, does this mean this QObject::connect(...) warning is a false positive?

Probably yes, we take care of disconnecting in the destructor, so the slot should not be triggered after deletion

> It's very hard to not have any false-positives. If you find any you probably can just pass the sender again, as 3rd parameter.

PR #32619, opened 2019-11-04
Followup 1f3b039: replace python bits

nirvn:
That shuts down a number of python errors I got after compiling QGIS.


m-kuhn (approved):
Sorry for that!

nyalldawson (commented):
Isn't this an API break? I use Qgis.QGIS_VERSION in plugins....

m-kuhn:
It is...
https://github.com/qgis/QGIS/commit/1f3b039fe2c9abaf4284b22ed2f94d74bf32d876#diff-d9ae6b71fecabe9e2f20481a7a951bf4R7

 should have taken care of that, is that not enough? will check ...

nirvn:
@m-kuhn , OK, after further investigation, problem was a wrong monkey patched variable name. Updated PR fixes that.

m-kuhn:
Ouch, thanks!!

PR #32628, opened 2019-11-04
[feature] Add save multiple styles action to style menu

elpaso:
Adds the possibility to save/export multiple styles
at once in QML/SLD and in the DB, in case of file system
export an incremental number is appended to the base
file name i.e. `export.qml`, `export_1.qml` etc.
in case of DBs the user can specify a new name
for the exported syles, and in that case an incremental
number is appended to the base style name i.e. name,
`name 1`, `name 2` etc., if the new name is left blank,
the style name is used and in case of conflicts the
incremental number is appended as well.

ezgif com-optimize

Funded by **ARPA Piemonte**

### DIFF
--- a/src/app/qgsvectorlayersavestyledialog.cpp
+++ b/src/app/qgsvectorlayersavestyledialog.cpp
@@ -85,6 +85,9 @@ QgsVectorLayerSaveStyleDialog::QgsVectorLayerSaveStyleDialog( QgsVectorLayer *la
   mStyleCategoriesListView->setModel( mModel );
 
   mStyleCategoriesListView->adjustSize();
+
+  setupMultipleStyles();
+
 }
 
 void QgsVectorLayerSaveStyleDialog::accept()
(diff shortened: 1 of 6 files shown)


src/app/qgsvectorlayersavestyledialog.cpp
```diff
@@ -156,6 +176,55 @@ void QgsVectorLayerSaveStyleDialog::readUiFileContent( const QString &filePath )
   }
 }
 
+void QgsVectorLayerSaveStyleDialog::setupMultipleStyles()
+{
+  // Show/hide part of the UI according to multiple style support
+  if ( ! mSaveOnlyCurrentStyle )
+  {
+    const auto mgr { mLayer->styleManager() };
+    const auto constStyles = mgr->styles();
```
nyalldawson: auto :-1:

nyalldawson (commented):
Code looks good to me!

src/app/qgsvectorlayersavestyledialog.cpp
nirvn: Could we try to come up with a coding guideline for auto? As was suggested a few times already, we should limit auto to the for loops iterating through lists / maps really.

Declaring a variable type goes right along using informative/clear variable names.

elpaso: I know about the policy: this is the first iteration and I'll clean it up before merge. I was more interested in comments regarding the functionality in this phase. 
But thank you for reminding me.

elpaso:
Merging in its current state, if is there anything I missed please let me know.

PR #32629, opened 2019-11-04
[dxf] HAlign/VAlign support for TEXT

m-kuhn:
halign / valign support for dxf

### DIFF
--- a/src/core/dxf/qgsdxfexport.cpp
+++ b/src/core/dxf/qgsdxfexport.cpp
@@ -2331,7 +2331,7 @@ void QgsDxfExport::drawLabel( const QString &layerId, QgsRenderContext &context,
 
   if ( mFlags & FlagNoMText )
   {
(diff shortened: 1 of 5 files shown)


src/core/dxf/qgsdxfexport.cpp
```diff
@@ -1391,20 +1452,30 @@ void QgsDxfExport::writeCircle( const QString &layer, const QColor &color, const
...
   writeHandle();
   writeGroup( 100, QStringLiteral( "AcDbEntity" ) );
+  // writeGroup( 6, "Continuous" ); // Line style
+  // writeGroup( 370, 18 ); // Line weight
   writeGroup( 100, QStringLiteral( "AcDbText" ) );
   writeGroup( 8, layer );
   writeGroup( color );
   writeGroup( 0, pt );
+  if ( hali != HAlign::Undefined || vali != VAlign::Undefined )
+    writeGroup( 1, pt ); // Second alignment point
```
m-kuhn: TODO: what exactly is the second alignment point

m-kuhn: Since no one knows and the results look ok, leave as is for now...

Issue #31449, opened 2019-08-27
Symbology tab widget does not horizontally expand in the layer properties dialog extent


DelazJ:
Open the layer properties dialog
Go to the symbology tab and resize the dialog: it behaves correctly
Enable another tab and enlarge the dialog again: widgets should follow
Now go back to the symbology tab: they did not follow there and you get something like 

![midfilleddialog](https://user-images.githubusercontent.com/7983394/63775864-87516a00-c8e0-11e9-82d6-790c94f5bb7e.PNG)

Observed on Windows 10, 4470baa1a3 (3.8.2) and a3f411acce (3.9)

elpaso:
I cannot reproduce on Linux/Master, I guess it's windows only

DelazJ:
No, I also get it on bionic (older than a week). But in today's windows nigthly build, the issue is still there (and for some reason, the dialog becomes even smaller than it was/should be). And as soon as I begin to resize the main dialog, the resizing is done for the symbol selector.
![midfilleddialog2](https://user-images.githubusercontent.com/7983394/64534138-2c266b00-d315-11e9-9f45-19d066f2f2f7.PNG)

elpaso:
I cannot reproduce this exact issue on Linux but I see some similar issues in the same dialog but different tabs.

elpaso:
@DelazJ waiting for your feed-back if you have a chance to test this patch, or we can just merge and you will find it in the nightly builds.

DelazJ:
Thanks @elpaso but sorry I'm not sure I can find a slot for building and testing. If it solves an issue on your side, let's merge and when available I'll check my issues on nightly

elpaso:
ok, np, I tested on windows and it also fixes your issue.

PR #32631, opened 2019-11-04
Resize source select dialogs according to container

### LINKED ISSUE #32615: Broken UI when loading layers through the Datasource Manager
# Problem

I am just trying to load a layer using the Datasource manager and I experience a weird behavior with the UI.

Broken UI
broken-ui

broken_ui

save-1

os

elpaso:
Fixes #32615

This was tricky and I don't really know what was causing the issue,
but looks like the patch doesn't break the other dialogs (mainly
vector layer properties) that were affected by #31449.

Fixes also an unreported issue with OGR source select inner dialog
not being painted when opening the source manager dialog.


nyalldawson:
Odd...

Can't comment on the fix, but I trust you to handle this one :)

nyalldawson (approved):

PR #32633, opened 2019-11-04
fixes default GRASS output geometry type

### LINKED ISSUE #32560: Processing "output type" should default to "auto"
In QGIS 3/Processing/GRASS the parameter "output type" has no selection by default and is mandatory.

This is *very* inconvenient, up to the old LTR this parameter defaulted to "auto", which is the right choice.

Moreover there are Processing/GRASS modules where the output type is hardcoded and this parameter does not make sense anyway, i.e.

https://github.com/qgis/QGIS/blob/master/python/plugins/processing/algs/grass7/ext/v_net_centrality.py

using "auto" as default would also help create less confusion with this modules.

gioman:
Fixes https://github.com/qgis/QGIS/issues/32560


alexbruy:
@gioman may I ask you to include `(fix #issue_number)` in the commit message next time? This helps when exploring history.

alexbruy (approved):

gioman:
> may I ask you to include `(fix #issue_number)` in the commit message next time? This helps when exploring history.

yes sorry. Can we backport this patch?

PR #32639, opened 2019-11-05
Fix unreported wfs3 Url without offset and limit regex

tudorbarascu:
@elpaso 

The regex was too greedy and it also stripped any layer named limit* or offset*

Before, starting from an url like
*  ``..wfs3/collections/limits/items.html?`` would have gone to 
* ``..wfs3/collections/s/items.html?&offset=10&limit=10``
Notice ``limits`` has been cut the ``limit`` part.


elpaso:
Superceeded by https://github.com/qgis/QGIS/pull/32645

elpaso:
@tudorbarascu thank you for finding the cause of this issue and the solution, but I'd prefer to get rid of the regexp completely ans use QUrl methods instead.

tudorbarascu:
@elpaso! Sure! Nice work!

PR #32642, opened 2019-11-05
add Q_ENUM macro to QgsVectorLayer::EditResult and SelectBehavior (and fix master build)

3nids:
useful for QML

### DIFF
--- a/python/core/auto_additions/qgsvectorlayer.py
+++ b/python/core/auto_additions/qgsvectorlayer.py
@@ -1,0 +1,3 @@
+# The following has been generated automatically from src/core/qgsvectorlayer.h
+QgsVectorLayer.EditResult.baseClass = QgsVectorLayer
+QgsVectorLayer.SelectBehavior.baseClass = QgsVectorLayer
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -721,6 +721,7 @@ SET(QGIS_CORE_HDRS
   qgsfeaturestore.h
   qgsfeedback.h
   qgsfield.h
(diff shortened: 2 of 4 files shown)


m-kuhn (approved):
Looks good to me
And thanks for the proj6 build fix

src/3d/CMakeLists.txt
```diff
@@ -165,7 +165,7 @@ IF(MSVC)
   SET_TARGET_PROPERTIES(qgis_3d PROPERTIES LINK_FLAGS "/FORCE:MULTIPLE")
 ENDIF(MSVC)
 
-SET(QGIS_3D_HDRS ${QGIS_3D_HDRS} ${CMAKE_CURRENT_BINARY_DIR}/qgis_3d.h)
+SET(QGIS_3D_HDRS ${QGIS_3D_HDRS} ${QGIS_3D_PRIVATE_HDRS} ${CMAKE_CURRENT_BINARY_DIR}/qgis_3d.h)
```
m-kuhn: Don't!!
Move the erroneous ones to the public header section instead.

m-kuhn (changes requested):
Private headers must not be installed.
If some of them need to be installed currently, they cannot be considered private and should go to the public headers instead.
Or better, included in the cpp file instead.

src/3d/CMakeLists.txt
3nids: ok, done

3nids:
private headers not installed, only qgsfield_p

m-kuhn (approved):
Thanks

PR #32643, opened 2019-11-05
Fix Qt::UniqueConnection with lambda

m-kuhn:
Qt::UniqueConnection cannot be used with lambdas


elpaso (approved):

nyalldawson:
This should be backported, right?

m-kuhn:
I guess so, yes

PR #32664, opened 2019-11-05
[travis] Fix ccache

m-kuhn:
## Checklist

### DIFF
--- a/.ci/travis/linux/scripts/docker-qgis-build.sh
+++ b/.ci/travis/linux/scripts/docker-qgis-build.sh
@@ -31,7 +31,7 @@ echo "travis_fold:start:cmake"
 echo "${bold}Running cmake...${endbold}"
 
 export CC=/usr/lib/ccache/clang
-export CXX=/usr/lib/ccache/clazy
+export CXX=/usr/lib/ccache/clang++
 
 cmake \
  -GNinja \
--- a/.docker/qgis3-build-deps.dockerfile
+++ b/.docker/qgis3-build-deps.dockerfile
@@ -152,7 +152,8 @@ RUN curl -k https://downloads.kdab.com/clazy/1.6/Clazy-x86_64-1.6.AppImage -o /t
   && mkdir /opt/clazy \
   && cd /opt/clazy \
   && /tmp/Clazy.AppImage --appimage-extract \
-  && ln -s /opt/clazy/squashfs-root/AppRun /usr/lib/ccache/clazy
+  && ln -s /opt/clazy/squashfs-root/AppRun /usr/bin/clazy \
+  && ln -s ../../bin/ccache /usr/lib/ccache/clazy
 
 ENV QT_SELECT=5
 ENV LANG=C.UTF-8


m-kuhn:
Fingers crossed. Let's retrigger a travis build once the first one finished to check if it works.

nirvn:
I've reset travis 3 times, travis always took ~2 hours, seems more is needed to fix ccache here.

m-kuhn:
Looks like the official recommendation is to disable ccache with clazy. But the reasoning on resources I find is mostly that it gives wrong results, and nowhere that it's not working

nirvn:
@m-kuhn , hmm, so what's the short term solution? Reverting the clazy stuff for now?

3nids:
I will kill previous jobs in the list so this one gets through and will restart the others afterwards

m-kuhn:
Can you restart this one once before merging to be sure it actually fixes the issues?

m-kuhn:
So clazy is disabled again. It's incompatible with ccache. I thought it would give wrong results sometimes (according to some sources) but the experience here is that ccache just does no longer work.

@3nids do you see a good possibility to run it on a nightly basis at least?
Or do we have other smart ideas how to lock in all our clazy fixes? CC @nyalldawson

3nids:
We could easily have these tests run on a daily basis using cron jobs. 

But basically we cannot use ccache? 
What is surprising is that we didn't get timeouts, that means that it probably still took advantage of some caching (no way to build full QGIS on Travis in 2.5h currently)

m-kuhn:
I think pretty much no caching involved. Maybe because the compiler behind clazy is clang-9?

ccache from a random one with ccache completely disabled (directly using clazy)
```
ccache statistics
cache directory                     /root/.ccache
primary config                      /root/.ccache/ccache.conf
secondary config      (readonly)    /etc/ccache.conf
stats zero time                     Wed Nov  6 13:56:58 2019
cache hit (direct)                    17
cache hit (preprocessed)               0
cache miss                             0
cache hit rate                    100.00 %
called for link                        9
preprocessor error                     1
cleanups performed                     0
files in cache                      9540
cache size                           1.0 GB
max cache size                       1.2 GB
```

And one with ccache and clazy enabled

```
cache directory                     /root/.ccache
primary config                      /root/.ccache/ccache.conf
secondary config      (readonly)    /etc/ccache.conf
stats zero time                     Thu Nov  7 05:24:23 2019
cache hit (direct)                     6
cache hit (preprocessed)               9
cache miss                          2534
cache hit rate                      0.59 %
called for link                      329
called for preprocessing             318
compile failed                         2
preprocessor error                     1
cleanups performed                    39
files in cache                     11190
cache size                           1.0 GB
max cache size                       1.2 GB
```

nyalldawson:
I'd be in favour of a once-a-day "code quality" check. We could include
- clazy checks
- selected clang static analysis warnings
- valgrind over selected unit tests, or enable asan on the build...

(Fwiw, even locally building with clazy slows down the compilation a lot)

PR #32677, opened 2019-11-06
Fix issues relating to features not matching graduated ranges

nyalldawson:
Caused by exact matching double values, when we need to have a little bit of tolerance. Also implements a fix in the OGR provider where certain methods (minimum/maximum/unique values) was force converting values via a string, resulting in loss of precision)

### DIFF
--- a/python/core/auto_generated/symbology/qgsgraduatedsymbolrenderer.sip.in
+++ b/python/core/auto_generated/symbology/qgsgraduatedsymbolrenderer.sip.in
@@ -499,6 +499,14 @@ Updates the labels of the ranges
 .. versionadded:: 3.10
 %End
 
+    const QgsRendererRange *rangeForValue( double value ) const;
+%Docstring
+Returns the renderer range matching the provided ``value``, or ``None`` if no
+range matches the value.
+
+.. versionadded:: 3.10.1
+%End
+
   protected:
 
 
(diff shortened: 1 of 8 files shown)


nirvn:
Oh nice one!

nyalldawson:
@elpaso have you any idea why this causes a failure in the server? https://travis-ci.org/qgis/QGIS/jobs/607990439

elpaso:
@nyalldawson this patch breaks the temporal extent calculation, that is now `null` instead of the expected:

```
"temporal": {
      "interval": [
        [
          "2010-01-01T00:00:00",
          "2022-01-01T01:01:01Z"
        ]
      ],
      "trs": "http://www.opengis.net/def/uom/ISO-8601/0/Gregorian"
    }
```

the calculation is implemented in `QgsServerApiUtils::temporalExtent` but I cannot dig into it now and see what's going wrong with this patch.

nyalldawson:
@rouault Does https://github.com/qgis/QGIS/pull/32677/commits/42a83c5891e1ea564a8674894ea4c3cb943a76d5 look correct to you? Seems that the OGR min/max aggregate results for date/datetime fields returns a string value, regardless of the original field data type. This issue is vaguely familiar and I have a feeling it was discussed sometime in the past, but can't find the relevant conversation now...

nyalldawson:
Anyone care to review this one?

nirvn (approved):
@nyalldawson , went through the PR, looking good. I've stumbled on this irritating issue before myself, glad to see it fixed.

nyalldawson:
Thanks @nirvn!

Issue #32321, opened 2019-10-19
DB Manager error: column adsrc doesn't exist any more in PostgreSQL 12


KoponenJ:
When executing raster query for PostGIS error "column adef.adsrc does not exist" is generated. This is because PostgreSQL 12 doesn't have pg_attrdef.adsrc column any more (?). From version 8 documentation:
"The adsrc field is historical, and is best not used, because it does not track outside changes that might affect the representation of the default value. Reverse-compiling the adbin field (with pg_get_expr for example) is a better way to display the default value."
From version 12 release notes: "Remove obsolete pg_attrdef.adsrc column (Peter Eisentraut)."

NyakudyaA:
I have also seen this by just clicking on a vector layer .
`column adef.adsrc does not exist
LINE 8: adef.adsrc AS default_value,
^
`

I do not think this is a QGIS issue because the logs from PostgreSQL are showing me the following


![logs](https://user-images.githubusercontent.com/2510900/67273641-251b7c80-f4bf-11e9-8bb0-50202032f3a8.png)

There has been various issues linked to it in postgresql
https://www.devexpress.com/Support/Center/Question/Details/T818996/column-d-adsrc-does-not-exist-postgresql-12
https://github.com/docker-library/postgres/issues/616

blackgis:
Any solutions to that issue? I just transfaired all from postgres 11 to postgres 12 and facing now the issue that qgis can handle  this from the browser tab but not from the db manager. Any solutions on how to fix it on the postgres side or does it need a fix in QGIS?

![Postgres12_QGIS310_issue](https://user-images.githubusercontent.com/7139491/68071115-6fcaad80-fd77-11e9-9796-9e1de3110257.jpg)

ra-mcinnes:
I'm also having the same issue. The rasters seem to be correct in the DB but QGIS views them as tables rather than a raster. Additionally QGIS doesnt recognize all of the columns in the Postgres raster tables. Genuinely have no idea what the work around would be here.

pcav:
A pretty serious issue. The PostGIS provider has to be upgraded.

pcav:
I think also the Data Browser is not capable of reading the layers.

elpaso:
Fixed by https://github.com/qgis/QGIS/pull/32684

Issue #32604, opened 2019-11-03
dxf export


tsterop:
**Describe the bug**

**How to Reproduce**

**QGIS and OS versions**

**Additional context**

when i am exporting qpx files( with wptpoints) to dxf, always missing 2 points.!

nyalldawson:
Please provide detail - this report has no content!

tsterop:
when i am exporting qpx files( with wptpoints) to dxf, always missing 2  points.!

nyalldawson:
Can you provide a sample test data? There still isn't enough here to meaningfully debug.

tsterop:
![qgis_points](https://user-images.githubusercontent.com/57331286/68092628-e78efa00-fe95-11e9-9863-fce1be5d6a61.png)
![dxf_points](https://user-images.githubusercontent.com/57331286/68092663-1f963d00-fe96-11e9-9281-d4ae865fe51d.png)

tsterop:
[dxf_export.zip](https://github.com/qgis/QGIS/files/3802196/dxf_export.zip)

tsterop:
[export_3_gps.pdf](https://github.com/qgis/QGIS/files/3802206/export_3_gps.pdf)
[dxf_exporting.pdf](https://github.com/qgis/QGIS/files/3802208/dxf_exporting.pdf)

tsterop:
with
QGIS version
2.14.10-Essen
issue does not exist, gpx points and dxf points are the same.

tsterop:
with qgis 2.14 there is no issue

m-kuhn:
> dxf_export.zip

Can you add the gpx file as well?

tsterop:
with all of this files qgis do the same (2.14.10 works fine)
[kanabia2_steno_makri.zip](https://github.com/qgis/QGIS/files/3802444/kanabia2_steno_makri.zip)

tsterop:
[kanabia2_steno_makri.zip](https://github.com/qgis/QGIS/files/3802446/kanabia2_steno_makri.zip)

with this type of file new versions og qgis do the same.!

gioman:
> with all of this files qgis do the same (2.14.10 works fine)

tried your gpx on QGIS 3.10, export > save features as > dxf

8 points in, 8 points out.

tsterop:
[qgis_issue.zip](https://github.com/qgis/QGIS/files/3803546/qgis_issue.zip)
No in my setup
[...]

PR #32679, opened 2019-11-06
Proposed fix for PostgreSQL 12

pcav:
See #32321
If accepted it should be backported to LTR and 3.10


pcav:
Further work is necessary: also column c.consrc does not exist. Simply removing `c.consrc` from the query removes the error, but further study is necessary to check whether this is correct.
See also https://github.com/OSGeo/gdal/commit/b868dbe333dd862e72cac89aa9f846a7c030f82c

pcav:
Removing `consrc` may require replacing it with `pg_get_constraintdef`, see https://github.com/typeorm/typeorm/issues/4332

nyalldawson:
Superseded by #32604 32684

PR #32684, opened 2019-11-06
Fix postgis 12 adscr -> adbin consrc -> conbin

elpaso:
See: #32679

@pcav can you test this patch (with older PG too)?

### DIFF
--- a/python/plugins/db_manager/db_plugins/postgis/connector.py
+++ b/python/plugins/db_manager/db_plugins/postgis/connector.py
@@ -489,6 +489,9 @@ class PostGisDBConnector(DBConnector):
         schema, tablename = self.getSchemaTableName(table)
         schema_where = u" AND nspname=%s " % self.quoteString(schema) if schema is not None else ""
 
+        version_number = int(self.getInfo()[0].split(' ')[1].split('.')[0])
+        ad_col_name = 'adsrc' if version_number < 12 else 'adbin'
+
         sql = u"""SELECT a.attnum AS ordinal_position,
                                 a.attname AS column_name,
                                 t.typname AS data_type,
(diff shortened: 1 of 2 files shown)


pcav:
Will do ASAP, thanks!

rouault:
Just noting that it seems we have some copy&paste pattern between dbmanager and the processing tool. Woud be cool factoring that a bit, but likely out of scope of a urgent fix PR

elpaso:
@rouault yes, I noticed it too. Ideally we should add that kind of information retrieval it to the `QgsAbstractDatabaseProviderConnection` API.

python/plugins/db_manager/db_plugins/postgis/connector.py
```diff
@@ -534,12 +537,15 @@ def getTableConstraints(self, table):
         schema, tablename = self.getSchemaTableName(table)
         schema_where = u" AND nspname=%s " % self.quoteString(schema) if schema is not None else ""
 
-        sql = u"""SELECT c.conname, c.contype, c.condeferrable, c.condeferred, array_to_string(c.conkey, ' '), c.consrc,
+        version_number = int(self.getInfo()[0].split(' ')[1].split('.')[0])
+        con_col_name = 'consrc' if version_number < 12 else 'conbin'
```
pcav: I think you can put `<11 here`

pcav:
Tested, apparently all OK.
Thanks again.

python/plugins/db_manager/db_plugins/postgis/connector.py
elpaso: I've not tested 11, I only have 8, 9, 10 and 12 here. Are you sure that 11 does not use `consrc` too? Looking here it looks like it does: https://www.postgresql.org/docs/11/catalog-pg-constraint.html

pcav:
AFAICT It was still present but deprecated. Anyway, it's irrelevant, you can drop the comment. Thanks.

pcav:
Thanks!

elpaso:
No problems!

python/plugins/db_manager/db_plugins/postgis/connector.py
GeorgeJones92: hello, i'm fairly new to GitHub however i'm encountering the same problem.  What does 'postgis 12 adscr -> adbin consrc -> conbin mean as a solution'?

agiudiceandrea:
Hi @elpaso, please see https://github.com/qgis/QGIS/issues/54379.

PR #32694, opened 2019-11-06
[feature] Server OAPIF simple transactions

elpaso:
# Simple OAPIF server transactions

Preliminary specs:
https://github.com/opengeospatial/ogcapi-features/blob/master/extensions/transactions/TX_Notes.adoc

# TODO
- [x] Permissions and access control
- [x] PUT
- [x] PATCH
- [x] DELETE


elpaso:
@pblottiere and here we are! I will eventually refactor the attributes parsing (there is some duplicated code between PUT POST and PATCH) but for now I'm happy with it.

pblottiere:
@elpaso Awesome, thanks Alessandro! I take a look asap.

src/server/qgsrequesthandler.cpp
```diff
@@ -189,70 +189,77 @@ void QgsRequestHandler::setupParameters()
...
+        // XXX Output error but continue processing request ?
+        QgsMessageLog::logMessage( QStringLiteral( "Warning: error parsing post data as XML: at line %1, column %2: %3. Assuming urlencoded query string sent in the post body." )
+                                   .arg( line ).arg( column ).arg( errorMsg ) );
+
+        // Process input string as a simple query text
+
+        typedef QPair<QString, QString> pair_t;
+        QUrlQuery query( inputString );
+        QList<pair_t> items = query.queryItems();
+        for ( pair_t pair : items )
```
pblottiere: Do you think we could use a const reference for `pair`?

src/server/qgsserverrequest.h
```diff
@@ -81,6 +85,13 @@ class SERVER_EXPORT QgsServerRequest
     //! destructor
     virtual ~QgsServerRequest() = default;
 
+    /**
+     * methodToString returns a string representation of an HTTP request \a method
```
pblottiere:
```suggestion
     * Returns a string representation of an HTTP request \a method.
```

src/server/services/wfs3/qgswfs3handlers.cpp
```diff
@@ -1072,248 +1099,398 @@ void QgsWfs3CollectionsItemsHandler::handleRequest( const QgsServerApiContext &c
...
-        else
-        {
-          expressions.push_back( QStringLiteral( "\"%1\" = '%2'" ).arg( it.key() ).arg( it.value() ) );
+          selfLink = l;
+          break;
         }
       }
-    }
+      // This should never happen!
+      Q_ASSERT( !selfLink.is_null() );
```
pblottiere: I'm not sure that a `Q_ASSERT` is a good idea within the server. What do you think about raising an exception instead?

src/server/services/wfs3/qgswfs3handlers.cpp
```diff
@@ -1390,19 +1567,331 @@ void QgsWfs3CollectionsFeatureHandler::handleRequest( const QgsServerApiContext
...
+          throw QgsServerApiBadRequestException( QStringLiteral( "Feature properties are not valid" ) );
+        }
+
+      }
+      catch ( json::exception & )
+      {
+        throw QgsServerApiBadRequestException( QStringLiteral( "Feature properties are not valid" ) );
+      }
+
+      // TODO: raise if nothing to change?
```
pblottiere: I'm not convinced about raising an exception in this case, but a log message would be welcome I think :).

pblottiere:
Thanks @elpaso, it's an impressive piece of work :+1:.

I'm currently building your branch to test the whole thing

src/server/services/wfs3/qgswfs3handlers.cpp
elpaso: Yes, agreed, even if it will never happen in practice.

src/server/qgsrequesthandler.cpp
elpaso: I just moved that code (it was already there), but I'll see if it builds with `const`.

pblottiere:
> I'm currently building your branch to test the whole thing

I didn't test all possibilities but the `API description` in the HTML landing page seems good and the DELETE method on a feature from a geopackage layer works as expected  :+1:.

BTW, what is happening if two instances of QGIS Server are updating the same layer? 

Otherwise I just noticed a minor thing on the `API description` page:
- the title is `API Description`
- but the link is named `API definition`

p

Thanks @elpaso!

elpaso:
> > I'm currently building your branch to test the whole thing
> 
> I didn't test all possibilities but the `API description` in the HTML landing page seems good and the DELETE method on a feature from a geopackage layer works as expected +1.
> 
> BTW, what is happening if two instances of QGIS Server are updating the same layer?
> 

I guess that the last wins, the same that would happen with concurrent edits in the normal WFS workflow. There is no feature locking in place, IIRC there was some discussion during the last OGC sprint about "versioning" support in complex transactions (that is another implementation, we only deal with "simple" transactions here) with an E-Tag header added to the response to tag a particular version of the feature that was modified, this would let the client know if the feature was modified by another client "before" the changes were posted, but I admit that this workflow is not completely clear to me.

> Otherwise I just noticed a minor thing on the `API description` page:
> 
>     * the title is `API Description`
> 
>     * but the link is named `API definition`
> 
> 
> p
> 
> Thanks @elpaso!

Thank you for the review! What do you think would be better here: "description" or "definition"? They are both correct IMHO.

pblottiere:
> I guess that the last wins, the same that would happen with concurrent edits in the normal WFS workflow. There is no feature locking in place, IIRC there was some discussion during the last OGC sprint about "versioning" support in complex transactions (that is another implementation, we only deal with "simple" transactions here) with an E-Tag header added to the response to tag a particular version of the feature that was modified, this would let the client know if the feature was modified by another client "before" the changes were posted, but I admit that this workflow is not completely clear to me.

OK, thanks for the explanation :+1:

> What do you think would be better here: "description" or "definition"? They are both correct IMHO.

I agree, they're both valid. I maybe have a slight preference for "description".

pblottiere (approved):

Issue #32224, opened 2019-10-12
Qgis crashing after editing vertex *Crash ID*: 150a74d9327ce5a4ceba5df0b1c73c21f96e4dec


Nikki-kuromi:
## User Feedback

Usually happens when exit from vertex editing

## Report Details

*Crash ID*: 150a74d9327ce5a4ceba5df0b1c73c21f96e4dec

*Stack Trace*
<pre>
QList<QItemSelectionRange>::size :
QAbstractTableModel::index :
QTableView::paintEvent :
QWidget::event :
QFrame::event :
QAbstractItemView::viewportEvent :
QCoreApplicationPrivate::sendThroughObjectEventFilters :
QApplicationPrivate::notify_helper :
QApplication::notify :
QgsApplication::notify :
QCoreApplication::notifyInternal2 :
QWidgetPrivate::drawWidget :
QApplication::windowIcon :
QWidget::event :
QgisApp::event :
QApplicationPrivate::notify_helper :
QApplication::notify :
QgsApplication::notify :
QCoreApplication::notifyInternal2 :
QWidgetPrivate::scrollRect :
QWidgetPrivate::invalidateBuffer_resizeHelper :
QWidget::repaint :
QWidget::repaint :
QAbstractButton::click :
QAbstractButton::mouseReleaseEvent :
QToolButton::mouseReleaseEvent :
QWidget::event :
QApplicationPrivate::notify_helper :
QApplication::notify :
QgsApplication::notify :
QCoreApplication::notifyInternal2 :
QApplicationPrivate::sendMouseEvent :
QSizePolicy::QSizePolicy :
QSizePolicy::QSizePolicy :
QApplicationPrivate::notify_helper :
QApplication::notify :
QgsApplication::notify :
QCoreApplication::notifyInternal2 :
QGuiApplicationPrivate::processMouseEvent :
QWindowSystemInterface::sendWindowSystemEvents :
QEventDispatcherWin32::processEvents :
CallWindowProcW :
DispatchMessageW :
QEventDispatcherWin32::processEvents :
qt_plugin_query_metadata :
QEventLoop::exec :
QCoreApplication::exec :
main :
BaseThreadInitThunk :
RtlUserThreadStart :
</pre>

*QGIS Info*
QGIS Version: 3.4.11-Madeira
QGIS code revision: 9a8a6d4687
Compiled against Qt: 5.11.2
Running against Qt: 5.11.2
Compiled against GDAL: 2.4.1
[...]

Issue #31809, opened 2019-09-16
QGIS crashes on pressing the Vertex tool button


himem:
## User Feedback

- added some layers from S-57 file (don't know if it relates to the crash somehow)
- added a new temporary layer with 'point' geometry
- added 2 points
- edited the 1st point
- pressed the 'vertex tool' button on the toolbar

## Report Details

*Crash ID*: 4736cd0a56fd1d2e0b86810b9ef8100c2a06e12e

*Stack Trace*
<pre>
QList<QItemSelectionRange>::size :
QTableView::paintEvent :
QWidget::event :
QFrame::event :
QAbstractItemView::viewportEvent :
QCoreApplicationPrivate::sendThroughObjectEventFilters :
QApplicationPrivate::notify_helper :
QApplication::notify :
QgsApplication::notify :
QCoreApplication::notifyInternal2 :
QWidgetPrivate::drawWidget :
QApplication::windowIcon :
QWidget::event :
QgisApp::event :
QApplicationPrivate::notify_helper :
QApplication::notify :
QgsApplication::notify :
QCoreApplication::notifyInternal2 :
QWidgetPrivate::scrollRect :
QWidgetPrivate::invalidateBuffer_resizeHelper :
QWidget::repaint :
QWidget::repaint :
QAbstractButton::click :
QAbstractButton::mouseReleaseEvent :
QToolButton::mouseReleaseEvent :
QWidget::event :
QApplicationPrivate::notify_helper :
QApplication::notify :
QgsApplication::notify :
QCoreApplication::notifyInternal2 :
QApplicationPrivate::sendMouseEvent :
QSizePolicy::QSizePolicy :
QSizePolicy::QSizePolicy :
QApplicationPrivate::notify_helper :
QApplication::notify :
QgsApplication::notify :
QCoreApplication::notifyInternal2 :
QGuiApplicationPrivate::processMouseEvent :
QWindowSystemInterface::sendWindowSystemEvents :
QEventDispatcherWin32::processEvents :
CallWindowProcW :
DispatchMessageW :
QEventDispatcherWin32::processEvents :
qt_plugin_query_metadata :
QEventLoop::exec :
QCoreApplication::exec :
main :
BaseThreadInitThunk :
RtlUserThreadStart :
</pre>

*QGIS Info*
QGIS Version: 3.4.11-Madeira
QGIS code revision: 9a8a6d4687
Compiled against Qt: 5.11.2
Running against Qt: 5.11.2
[...]

PR #32703, opened 2019-11-07
Fix crash when deactivating vertex editor (fixes #32685)

### LINKED ISSUE #32685: Crash when using vertex editor
**Describe the bug**

Getting a QGIS Crash when using the Vextex Editor Panel. My layer has Z values.

After the crash the command line shows the following:

```
Warning: QObject::connect: invalid null parameter
Warning: QObject::connect: invalid null parameter
Warning: QObject::connect: invalid null parameter
Warning: QOpenGLContext::swapBuffers() called with non-exposed window, behavior is undefined
need for update
need for update
need for update
need for update
need for update
Xlib: sequence lost (0x1160c > 0x160e) in reply type 0x23!
QGIS died on signal 11Could not attach to process.  If your uid matches the uid of the target
process, check the setting of /proc/sys/kernel/yama/ptrace_scope, or try
again as the root user.  For more details, see /etc/sysctl.d/10-ptrace.conf
ptrace: Operation not permitted.
No thread selected
No stack.
gdb returned 0
Aborted (core dumped)
```

**How to Reproduce**

1. Turn editing mode for a vector layer
2 . Using the vertex tool, select a couple of vertices
3. Right-click the vertices to open the Vertex Editor
4. Click the Select Features by area tool (I think any other tool will result in the same problem)
5. Bum!

Peek 2019-11-06 12-05

**QGIS and OS versions**

QGIS version | 3.10.0-A Coruña | QGIS code revision | 6c816b4204
-- | -- | -- | --
PROJ.4 Version | 493
OS Version | Ubuntu 18.04.3 LTS
Active python plugins | MetaSearch; db_manager; processing

**Additional context**

Tested with a clean profile

nirvn:
## Description

This PR insures that the vertex editor's model is cleared when the vertex tool is deactivated to avoid a crash under certain circumstances (demonstrated in a useful GIF attached to issue #32685 ).

## Checklist

### DIFF
--- a/src/app/vertextool/qgsvertextool.cpp
+++ b/src/app/vertextool/qgsvertextool.cpp
@@ -1413,12 +1413,13 @@ void QgsVertexTool::showVertexEditor()  //#spellok
 
 void QgsVertexTool::cleanupVertexEditor()
 {
-  mLockedFeature.reset();
-  // do not delete immediately because vertex editor
-  // can be still used in the qt event loop
-  mVertexEditor->deleteLater();
-
-  updateLockedFeatureVertices();
+  if ( mVertexEditor )
+  {
+    cleanupLockedFeature();
+    // do not delete immediately because vertex editor
+    // can be still used in the qt event loop
+    mVertexEditor->deleteLater();
+  }
 }
 
 void QgsVertexTool::cleanupLockedFeature()


nirvn:
This also fixes issues #32224 , #31809 , and likely more.

nyalldawson (approved):

PR #32708, opened 2019-11-07
[quick] fix deprecated use of QApplication::desktopWidget()->screenGeometry()

3nids:
This takes the first top level window and use it to get the screen.

The "perfect" approach would be to use a widget as argument in the method.

@PeterPetrik thoughts?

### DIFF
--- a/src/quickgui/qgsquickutils.cpp
+++ b/src/quickgui/qgsquickutils.cpp
@@ -13,9 +13,10 @@
  *                                                                         *
  ***************************************************************************/
 
-#include <QApplication>
-#include <QDesktopWidget>
+#include <QGuiApplication>
+#include <QScreen>
 #include <QString>
+#include <QWindow>
 
 #include "qgis.h"
 #include "qgscoordinatereferencesystem.h"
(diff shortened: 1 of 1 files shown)


PeterPetrik:
why we cannot use the "perfect" approach? 
let me ask @sklencar to comment

sklencar:
Outputs from simple test on Macbook pro 15", where results of the new implementation has more precious results.
Old vs new implementation:
```
screen resolution: 1440x900 px
screen DPI: 111x111
screen size: 330x206 mm
screen density: 0.69375
```
```
screen resolution: 1440x900 px
screen DPI: 110.5x110.5 //-> correct value
screen size: 331x207 mm //-> correct value
screen density: 0.690625
```

PeterPetrik (approved):

PeterPetrik:
@3nids what is the status of this PR? shall we merge it?

3nids:
yep!

PeterPetrik:
feel free to merge when tests pass

nyalldawson:

PR #32725, opened 2019-11-08
Processing - "Use filename as layer name" as default

### LINKED ISSUE #32591: Wrong layer name after processing procedures
In older QGIS versions (2.x), if I saved a layer to a file (Save to file) and I gave a name to that e.g. "airports", by running the algorithm (Add saved layer to project - option ticked) the created and loaded  layer's name was the same like the given filename. 

In newer versions this behavior disappeared, the resulted layer's name will be "Clipped", "Added geom info", "Intersected", instead of the given one.

I suppose the old solution was better. 
Please up vote this issue, if you feel the same.

nyalldawson:
Seems this option is widely desirable and should be the new default. Here I've refined it's behavior, added unit tests, and changed the setting key so that everyone gets the new enabled-by-default state.

Fixes #32591

### DIFF
--- a/src/core/processing/qgsprocessingcontext.cpp
+++ b/src/core/processing/qgsprocessingcontext.cpp
@@ -17,6 +17,8 @@
 
 #include "qgsprocessingcontext.h"
 #include "qgsprocessingutils.h"
+#include "qgsproviderregistry.h"
+#include "qgssettings.h"
 
 QgsProcessingContext::QgsProcessingContext()
   : mPreferredVectorFormat( QgsProcessingUtils::defaultVectorExtension() )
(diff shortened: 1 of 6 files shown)


alexbruy (approved):

src/core/processing/qgsprocessingcontext.cpp
```diff
@@ -119,3 +121,39 @@ void QgsProcessingContext::LayerDetails::setPostProcessor( QgsProcessingLayerPos
...
+    return;
+
+  const bool preferFilenameAsLayerName = QgsSettings().value( QStringLiteral( "Processing/Configuration/PREFER_FILENAME_AS_LAYER_NAME" ), true ).toBool();
+
+  // note - for temporary layers, we don't use the filename, regardless of user setting (it will be meaningless!)
+  if ( ( preferFilenameAsLayerName && !layer->isTemporary() ) || name.isEmpty() )
+  {
+    const QVariantMap sourceParts = QgsProviderRegistry::instance()->decodeUri( layer->providerType(), layer->source() );
+    const QString layerName = sourceParts.value( QStringLiteral( "layerName" ) ).toString();
+    // if ouput layer name exists, use that!
```
alexbruy: output

PR #32727, opened 2019-11-08
fix master build (missing include directory)

3nids:


m-kuhn:
https://github.com/qgis/QGIS/pull/32729 seems to be more complete

PR #32729, opened 2019-11-08
Fix includes

m-kuhn:
Followup 3b5f3e5

```sh
git grep -l "core\/symbology" | xargs grep -iL "core\/effects" | xargs sed -i -e '/\${CMAKE_SOURCE_DIR}\/src\/core\/symbology/a\' -e '  \${CMAKE_SOURCE_DIR}\/src\/core\/effects'
```


3nids:
This has always be a pain. I wonder if we should rather define some variables in the CMakeLists such as CORE_INCLUDE_DIRS and we would add them wherever needed. Does it make sense?

m-kuhn:
A lot!

PR #32732, opened 2019-11-08
[BUGFIX] Fix an hardcoded 'gpkg' extension in ProcessingConfig.py

lbartoletti:
## Description
GDAL must be compiled with sqlite to enable gpkg. Don't hardcode 'gpkg' but use `QgsVectorFileWriter.supportedFormatExtensions()[0]` instead that will return 'gpkg' if present else 'shp'.

## Checklist

### DIFF
--- a/python/plugins/processing/core/ProcessingConfig.py
+++ b/python/plugins/processing/core/ProcessingConfig.py
@@ -152,7 +152,7 @@ class ProcessingConfig:
             ProcessingConfig.tr('General'),
             ProcessingConfig.DEFAULT_OUTPUT_VECTOR_LAYER_EXT,
             ProcessingConfig.tr('Default output vector layer extension'),
-            'gpkg',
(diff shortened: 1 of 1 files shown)


lbartoletti:
The error:

```
ValueError: 'gpkg' is not in list 
  File "/home/lbartoletti/build/output/python/plugins/processing/gui/ConfigDialog.py", line 82, in apply
    self.config_widget.accept()
  File "/home/lbartoletti/build/output/python/plugins/processing/gui/ConfigDialog.py", line 326, in accept
    setting.save(qsettings)
  File "/home/lbartoletti/build/output/python/plugins/processing/core/ProcessingConfig.py", line 327, in save
    qsettings.setValue(self.qname, self.options.index(self.value))
ValueError: 'gpkg' is not in list
```

alexbruy (approved):

PR #32735, opened 2019-11-08
[quick] add utility method to select feature using their IDs from QML

3nids:


wonder-sk:
hi @3nids - I am curious about the use case - QgsVectorLayer is already exposed in quick library, and selectByIds() is marked as Q_INVOKABLE - so why not just call it directly from QML?

3nids:
@wonder-sk thanks for looking 

I recently made it invokable but failed to actually use it.
I wrote the issue in the dox: This method is required since QML cannot perform the conversion of a feature ID to a QgsFeatureId (i.e. a qint64)
From what I read, there is no way to use long int in javascript, and I could not find a way to create it.
Are we missing some implicit conversion which should be implemented?

wonder-sk:
Hmm not sure exactly how the implicit conversions are supposed to work in QML...
Feel free to merge as it is, once we have a better solution we can retire this function.

PR #32736, opened 2019-11-08
Server OAPIF: expose XML schema for collections

elpaso:
Forward collection `describedBy` link to `WFS` (like geoserver does)

Also make sure links to parent paths end with /


tests/testdata/qgis_server/api/test_wfs3_collection_testlayer_èé.json
```diff
@@ -48,8 +48,14 @@ Content-Type: application/json
       "rel": "items",
       "title": "A test vector layer èé",
       "type": "text/html"
+    },
+    {
+      "href": "http://server.qgis.org/?request=DescribeFeatureType&typenames=testlayer%20èé&service=WFS&version=2.0",
```
rouault: isn't it a bit risky to have a link that doesn't point to a subpath of /wfs3 ? Doesn't that depend on the server configuration ? Otherwise an alternative might be to have a /wfs3/collections/testlayer/xmlschema endpoint, that would internally do the WFS DescribeFeatureType request

elpaso: It would work either ways, we are in a transitional phase where the URL router examines the query string for `SERVICE` and routes to WFS/WMS/etc. if it finds it, otherwise it enters the new API router that will match all registered API endpoints (OAPIF registers itself at `/wfs3`).

So, if we specify `SERVICE=WFS` in the query string, it will route to WFS whatever the path is.

But I want to change this!

What I want to do (eventually) is to port WFS/WMS/etc. to the new QGIS API code, which means that they will be wrapped into the same `QgsServerApi` that `QgsServerOgcApi` (i.e. OAPIF) uses and they will register themselves to `/wfs`, `/wms` etc. and then we could use a landing page that responds to the server's root path listing all available services.

So, the idea is that, supposing that there is URL rewrite in place that reads the project name form the path as: `http://server.qgis.org/{project_base_name}/`

Calling:
- `http://server.qgis.org/{project_base_name}/` -> nice landing page listing all services (WFS, OAPIF, WMS, custom services etc.)
- `http://server.qgis.org/{project_base_name}/wfs/` -> WFS (XML error because no service was specified)
- `http://server.qgis.org/{project_base_name}/wms/` -> WMS (XML error because no service was specified)
- `http://server.qgis.org/{project_base_name}/wfs3/` -> OAPIF landing page

Does it make sense?

elpaso: > Otherwise an alternative might be to have a /wfs3/collections/testlayer/xmlschema endpoint, that would internally do the WFS DescribeFeatureType request

forgot to mention that this is not possible: wfs/wms etc. are "plugins" dynamically loaded, they are not in the QGIS Server API.

rouault: > Does it make sense?

yes. But as you say that no all plugins are necessary loaded, you'd have to make the describedBy conditioned by the presence of WFS ?

elpaso: I'm not saying the plugins are not necessarily loaded, they are all always loaded (unless someone removes them from the disk on purpose), what I'm saying is that they are not part of QGIS Server C++ API: I cannot invoke a WFS C++ method.

elpaso: In any event, this forward to the WFS XML schema is probably also transitional, I'd like to implement a JSON schema but I'd probably need some funding.

PR #32737, opened 2019-11-08
[3d] Multi-threaded loading of terrain tiles

wonder-sk:
This PR lets chunk queue to run multiple jobs at the same time. Since every job simply starts work in background threads, this enables multi-threaded loading.

Running with max. 4 active jobs compared to single job has brought my loading time of my test scene from ~25 seconds to ~13 seconds. So with my scene it's about 2x speedup.

## Event tracing

I have added a "hidden" shortcut `Ctrl+Shift+.` to toggle event tracing on and off. When turning it off, it will ask for a path where to save the recorded trace.

Few improvements were done on even tracing code too, to enable better tracing of the loading queue. Before we used just "duration" events which are good for tracing of work within a single thread, but not good for async events - the duration events from a single thread are expected to be strictly nested, which is not case when running multiple independent asynchronous actions.

This is how a trace looks like (the top part shows how the queue gets processed, the bottom part shows threads doing actual work (that is not yet(?) committed).

image

## Notes
- our raster pipeline is sloooow. For example, just turning raster values into colors (in QgsSingleBandGrayRenderer) takes us e.g. 5x more time than time GDAL needs to fetch that data. This is a debug build so the numbers will be a bit skewed, but there's a lot to improve (lots of ifs, using doubles, virtual function calls). It could be interesting to use GPU to do most of the raster processing using custom shaders to do all transforms. Or at least optimize some common cases...
- it seems that background jobs (2D rendering, 3D rendering, task manager) all share the same global thread pool. Maybe at some point we should start looking into more fine grained control - for example I am wondering if some tasks running for longer time can't block map canvas rendering for extended period of time...?

This work has been sponsored by a QGIS.org grant.

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -677,8 +677,20 @@ void QgisApp::vectorLayerStyleLoaded( QgsMapLayer::StyleCategories categories )
 
 void QgisApp::toggleEventTracing()
 {
+  QgsSettings settings;
+  if ( !settings.value( QStringLiteral( "qgis/enableEventTracing" ), false ).toBool() )
+  {
+    // make sure the setting is available in Options > Advanced
+    if ( !settings.contains( QStringLiteral( "qgis/enableEventTracing" ) ) )
+      settings.setValue( QStringLiteral( "qgis/enableEventTracing" ), false );
+
+    messageBar()->pushWarning( tr( "Event Tracing" ), tr( "Tracing is not enabled. Look for \"enableEventTracing\" in Options > Advanced." ) );
+    return;
+  }
+
   if ( !QgsEventTracing::isTracingEnabled() )
   {
+    messageBar()->pushSuccess( tr( "Event Tracing" ), tr( "Tracing started." ) );
     QgsEventTracing::startTracing();
   }
   else
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -196,6 +196,7 @@ Q_GUI_EXPORT extern int qt_defaultDpiX();
 #include "qgsencodingfiledialog.h"
 #include "qgserror.h"
 #include "qgserrordialog.h"
+#include "qgseventtracing.h"
 #include "qgsexception.h"
 #include "qgsexpressionselectiondialog.h"
 #include "qgsfeature.h"
@@ -671,6 +672,21 @@ void QgisApp::vectorLayerStyleLoaded( QgsMapLayer::StyleCategories categories )
     {
       checkVectorLayerDependencies( vl );
     }
+  }
+}
+
+void QgisApp::toggleEventTracing()
+{
+  if ( !QgsEventTracing::isTracingEnabled() )
+  {
+    QgsEventTracing::startTracing();
+  }
+  else
+  {
+    QgsEventTracing::stopTracing();
+    QString fileName = QFileDialog::getSaveFileName( this, tr( "Save Event Trace..." ), QString(), tr( "Event Traces (*.json)" ) );
+    if ( !fileName.isEmpty() )
+      QgsEventTracing::writeTrace( fileName );
   }
 }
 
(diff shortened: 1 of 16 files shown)


src/app/qgisapp.cpp
```diff
@@ -1430,6 +1446,9 @@ QgisApp::QgisApp( QSplashScreen *splash, bool restorePlugins, bool skipVersionCh
     attributeTable( QgsAttributeTableFilterModel::ShowVisible );
   } );
 
+  QShortcut *shortcutTracing = new QShortcut( QKeySequence( tr( "Ctrl+Shift+." ) ), this );
```
nyalldawson: Mind adding "alt" in here too? This really isn't a shortcut you'd expect normal users to ever use, and we would want to minimise accidental triggers

nyalldawson (approved):
Code looks good to me!

nirvn:
@wonder-sk , full disclosure: I'm just crossing all of my available fingers hoping for some raster pipeline optimization from you on top of @nyalldawson 's recent resampling work :)

nyalldawson:
> our raster pipeline is sloooow

This is strange to me. I've done a lot of profiling with this too and usually I see the underlying gdal retrieval dominating the time (as you'd want to see). Certain filters do slow things down (eg bicubic resampling up to a week ago was super slow). I also see a LOT of wasted time in debug profiles due to string handling for debug messages. This is done even if the are on high debug levels and can absolutely kill performance, and the raster filters seem to be very debug message happy. (I've just been ifdefing these out wherever I find them)

nyalldawson:
> for example I am wondering if some tasks running for longer time can't block map canvas rendering for extended period of time...?

Yes, this IS the case!

wonder-sk:
> > our raster pipeline is sloooow
>
> This is strange to me. I've done a lot of profiling with this too and usually I see the underlying gdal retrieval dominating the time (as you'd want to see).

Okay I did some tracing on a release build and realized it is not too bad :-) It was significantly skewed by the debug build - such a novice profiling error :-)

src/app/qgisapp.cpp
wonder-sk: Hmm... I have tried to add Alt in as well - but hit an issue: for people who use multiple keyboard layouts (like me: English & Slovak), Alt+Shift is a commonly used keyboard shortcut to switch keyboard layouts (all the way back to windows 95). And this global shortcut gets priority over Ctrl+Alt+Shift+. defined here :-( Also, trying to press Ctrl + Alt + Shift at the same time, I can't find a way how to press them (at least on my keyboard) without curling my fingers in a very awkward way :)

So... can we live with Ctrl+Shift+. or do you have any other ideas how to make it harder to trigger accidentally? It feels that the keys are quite far away to be pressed by accident - and fortunately there isn't any destructive action going to happen (not even slowing down QGIS measurably).

nyalldawson: What about either:
1. hiding it behind a qsettings flag ("enable tracing")? 
2. make it show in the customisation dialog but without any initial shortcut assigned, so that it's an opt-in thing?

wonder-sk: Hiding it behind a qsettings flag would be probably a better solution...

One of the use cases is that I want to be able to instruct a random bug reporter with a problem with 3D to record a trace and attach it to the ticket. If it is possible to enable tracing in Options > Advanced then that should not be too difficult for bug reporters to do...

nyalldawson: I guess another option would be to put it in iface -- so you'd enable by pasting something in the python console like

   iface.startTrace()
   iface.saveTrace('c:/....')

wonder-sk: I took the approach with a QSettings flag that has to be enabled in Options > Advanced first...

The option with iface sounds also interesting for the future if we want to run this in a script...

nyalldawson: Great!

nyalldawson: Great!

nyalldawson:
Merge away!!

PR #32766, opened 2019-11-10
Urlencode non ascii chars in User-Agent

### LINKED ISSUE #32740: QGIS 3.10 - WMS issue
When QGIS makes a WMS request it adds the current version to the User-Agent string, so in version QGIS 3.10.0 this results in "Mozilla/5.0 QGIS/3.10.0-A Coruña". This in turn causes some servers to reject requests, probably because of the "ñ" character.

In the main my WMS requests are made to the UK Environment Agency's WMS server and in QGIS 3.10 these are always rejected.

Test WMS URL: 
https://environment.data.gov.uk/spatialdata/lidar-composite-digital-surface-model-dtm-1m/wms

elpaso:
Fixes #32740

This fixes the issue on my tests, the user agent looks like this:

```
'User-Agent', 'Mozilla/5.0 QGIS/Coru%C3%B1a'
```

I tried to encode following https://tools.ietf.org/html/rfc2047 but it didn't work, if anyone has a better idea about how to encode non-ascii chars in the headers please speak out.

Also, there is no transliterate options in `toAscii()` that is just an (misnamed) alias to `toLatin1()` that encodes in `ISO-8859-15` and does not solve this issue either.

### DIFF
--- a/src/core/qgsnetworkaccessmanager.cpp
+++ b/src/core/qgsnetworkaccessmanager.cpp
@@ -221,8 +221,8 @@ QNetworkReply *QgsNetworkAccessManager::createRequest( QNetworkAccessManager::Op
   QString userAgent = s.value( QStringLiteral( "/qgis/networkAndProxy/userAgent" ), "Mozilla/5.0" ).toString();
   if ( !userAgent.isEmpty() )
     userAgent += ' ';
-  userAgent += QStringLiteral( "QGIS/%1" ).arg( Qgis::version() );
-  pReq->setRawHeader( "User-Agent", QUrlQuery( userAgent ).toString( QUrl::ComponentFormattingOption::EncodeUnicode ).toUtf8() );
+  userAgent += QStringLiteral( "QGIS/%1" ).arg( Qgis::versionInt() );
+  pReq->setRawHeader( "User-Agent", userAgent.toLatin1() );
 
 #ifndef QT_NO_SSL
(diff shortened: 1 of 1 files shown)


nyalldawson:
I wonder if it wouldn't be better just to use the version number here alone?

elpaso:
Yes, that would also solve the issue. Any objections?

nyalldawson:
I think encoding aside it's a good move. The names are much less relatable vs pure version numbers

nyalldawson (approved):

nirvn:
Good move.

elpaso:
Actually, I remember something about OSM, do we need to tell them about this change?

rduivenvoorde:
@elpaso OSM's problem was that we (still) sent "Mozilla/5.0" as default part of the "User-Agent" header, but we do NOT listen to their browser cookies etc etc. 

Their question was to  remove the "Mozilla/5.0" from it. But counter argument from the QGIS community was that using a non-browser User-Agent made 'certain services or proxies' fail.

I would be ok with removing the default  "Mozilla/5.0" from the User-Agent anyway.
You can do in the options, so make the input empty there and then you only receive the QGIS-version:

Screenshot-20191213123615-416x78

Also I think that IF the users SET's an User-Agent, that QGIS then not should put something behind it... but that is personal...

agiudiceandrea:
> Also I think that IF the users SET's an User-Agent, that QGIS then not should put something behind it... but that is personal...

@rduivenvoorde  @elpaso, see https://github.com/qgis/QGIS/issues/58258.

PR #32770, opened 2019-11-10
Make the DXF renderer ready for background threading and fix symbology

m-kuhn:
This puts the whole dxf export into one single rendering process and prefetches information related to this before starting to render.

The immediate benefit is to fix symbology. The current code writes symbol blocks for symbol layers (e.g. for dashed lines), when the lines are actually rendered, they don't reference the block. This was caused by the fact that rendering was restarted for writing blocks and for writing entities. With the result that different clones of the symbol layers were used and therefore could not be matched.

The advantage in the longer run is that we will be able to run dxf exports in the background (allow to cancel, processing algorithms, ...) because they no longer rely on the QgsVectorLayer objects but rather on feature source and other threading friendly facilities.

### DIFF
--- a/python/core/auto_generated/dxf/qgsdxfexport.sip.in
+++ b/python/core/auto_generated/dxf/qgsdxfexport.sip.in
@@ -85,6 +85,8 @@ The attribute value is used for layer names.
 Constructor for QgsDxfExport.
 %End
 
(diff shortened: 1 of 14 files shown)


nyalldawson:
Great move... This has been on my to-do list for a long time!

PR #32774, opened 2019-11-11
Update sip, fix use of old style enum

nyalldawson:


3nids:
why do you have to switch back?

PR #4894, opened 2017-07-20
use QGIS API instead of GDAL API to create heatmaps

alexbruy:
## Description
Rough attempt to port heatmap code to QGIS raster API. Based on @wonder-sk excellent work on #4022.

Heatmaps, created by new code are almost the same as original, very small difference caused by rounding and slightly different approach to define pixel size in QGIS API, probably this can be addressed somehow. As Processing test for Heatmap algorithm passes, difference should be a really small.

@nyalldawson, @wonder-sk I would be grateful for your review and opinions about this.

## Checklist

> Reviewing is a process done by project maintainers, mostly on a volunteer basis. We try to keep the overhead as small as possible and appreciate if you help us to do so by completing the following items. Feel free to ask in a comment if you have troubles with any of them.

- [x] Commit messages are descriptive and explain the rationale for changes
- [ ] Commits which fix bugs include `fixes #11111` in the commit message next to the description
- [ ] Commits which add new features are tagged with `[FEATURE]` in the commit message
- [ ] Commits which change the UI or existing user workflows are tagged with `[needs-docs]` in the commit message and containt sufficient information in the commit message to be documented
- [x] I have read the [QGIS Coding Standards](https://docs.qgis.org/testing/en/docs/developers_guide/codingstandards.html) and this PR complies with them
- [x] This PR passes all existing unit tests (test results will be reported by travis-ci after opening this PR)
- [ ] New unit tests have been added for core changes
- [x] I have run [the `scripts/prepare-commit.sh` script](https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTE.md#contributing-to-qgis) before each commit

PR #32803, opened 2019-11-12
use QGIS API instead of GDAL API to create heatmaps

alexbruy:
## Description
Use QGIS raster API in the heatmap (KDE) calculation. Main motivation is to allow creating heatmaps using not only GDAL provider but any other raster data provider (if more will be added in future).

In my tests using point datasets with 20000 and 500000 features and different raster sizes difference in performance is insignificant.

Revive #4894.

## Checklist

### DIFF
--- a/src/analysis/raster/qgskde.cpp
+++ b/src/analysis/raster/qgskde.cpp
@@ -13,10 +13,13 @@
  *                                                                         *
  ***************************************************************************/
 
+#include <QByteArray>
+
 #include "qgskde.h"
 #include "qgsfeaturesource.h"
 #include "qgsfeatureiterator.h"
 #include "qgsgeometry.h"
+#include "qgsrasterfilewriter.h"
 
 #define NO_DATA -9999
 
@@ -32,8 +35,6 @@ QgsKernelDensityEstimation::QgsKernelDensityEstimation( const QgsKernelDensityEs
   , mDecay( parameters.decayRatio )
   , mOutputValues( parameters.outputValues )
   , mBufferSize( -1 )
-  , mDatasetH( nullptr )
-  , mRasterBandH( nullptr )
 {
   if ( !parameters.radiusField.isEmpty() )
     mRadiusField = mSource->fields().lookupField( parameters.radiusField );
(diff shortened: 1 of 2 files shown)


src/analysis/raster/qgskde.cpp
```diff
@@ -86,16 +79,30 @@ QgsKernelDensityEstimation::Result QgsKernelDensityEstimation::prepare()
...
-  if ( !mDatasetH )
-    return FileCreationError;
-  mRasterBandH = GDALGetRasterBand( mDatasetH.get(), 1 );
-  if ( !mRasterBandH )
+  if ( !mProvider->setNoDataValue( 1, NO_DATA ) )
+  {
     return FileCreationError;
+  }
+
+  int dataTypeSize = QgsRasterBlock::typeSize( Qgis::Float32 );
```
nyalldawson: This could be replaced by a call to `block.setIsNoData()`

src/analysis/raster/qgskde.cpp
```diff
@@ -162,13 +169,13 @@ QgsKernelDensityEstimation::Result QgsKernelDensityEstimation::addFeature( const
...
+    QgsRectangle extent( ( *pointIt ).x() - radius, ( *pointIt ).y() - radius, ( *pointIt ).x() + radius, ( *pointIt ).y() + radius );
+
     // get the data
-    float *dataBuffer = ( float * ) CPLMalloc( sizeof( float ) * blockSize * blockSize );
-    if ( GDALRasterIO( mRasterBandH, GF_Read, xPosition, yPositionIO, blockSize, blockSize,
-                       dataBuffer, blockSize, blockSize, GDT_Float32, 0, 0 ) != CE_None )
-    {
-      result = RasterIoError;
-    }
+    QgsRasterBlock *block = mProvider->block( 1, extent, blockSize, blockSize );
```
nyalldawson:
```suggestion
    std::unique_ptr< QgsRasterBlock > block( mProvider->block( 1, extent, blockSize, blockSize ) );
```

nyalldawson (approved):

PR #32807, opened 2019-11-12
Fix compilation with llvm > 6 (and arch=i386)

lbartoletti:
Follow up https://www.mail-archive.com/qgis-developer@lists.osgeo.org/msg47062.html discussion

I will test (for build time) this method `"-mllvm -inline-threshold=128"` , and \"-O1"\ in another one.

cc @m-kuhn @rhurlin @landryb

## Checklist


landryb:
Fwiw that's what i have in https://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/geo/qgis/patches/patch-src_core_CMakeLists_txt?rev=1.23&content-type=text/x-cvsweb-markup - the qrc approach looks interesting too.

lbartoletti:
> Alternative approach: https://github.com/opengisch/OSGeo4A/blob/master/recipes/qgis/patches/0001-Use-qrc-for-crs-mapping.patch

cc @rhurlin

rhurlin:
> Alternative approach: https://github.com/opengisch/OSGeo4A/blob/master/recipes/qgis/patches/0001-Use-qrc-for-crs-mapping.patch

I just tried this patch on FreeBSD 13.0-CURRENT amd64 and it applies and builds fine. Will try i386 (11.3, 12.1) later, when at home again.

lbartoletti:
Nice @m-kuhn !
I think the next time I see you, you will receive some BSD :smiling_imp: 🐡 badges/stickers :wink:

m-kuhn:
> Nice @m-kuhn !
> I think the next time I see you, you will receive some BSD smiling_imp blowfish badges/stickers wink

Yay :tada:  :dagger: :smiling_imp:

rhurlin:
On FreeBSD with Poudriere, I have problems building for i386 with this patch (0001-Use-qrc-for-crs-mapping.patch). It takes hours and hours ...

lbartoletti:
@rhurlin @m-kuhn OK so the patch is not acceptable for i386. I'll try third solutions on my machine this week-end.

For now, best solution seems "-mllvm -inline-threshold=128" (I'll launch a build with "-O1")

lbartoletti:
@rhurlin Do you agree with "-mllvm -inline-threshold=128" solution, since  "-O1" is longer and the patch from mathias doesn't works for i386 arch (very longer)?

rhurlin:
> @rhurlin Do you agree with "-mllvm -inline-threshold=128" solution, since "-O1" is longer and the patch from mathias doesn't works for i386 arch (very longer)?

For me this patch is fine. It passes several Poudriere runs with i368 and amd64 on FreeBSD. Thank you very much!

lbartoletti:
@m-kuhn can you merge it when travis will be happy please?
@rhurlin thanks for testing.

And most of all, thanks to OpenBSD :blowfish: cc @landryb ;)

m-kuhn:
Thanks everyone!!

PR #32811, opened 2019-11-12
fix cache read on wfs request

troopa81:
## Description

The issue happen when the wfs server returns http code 304 (Document not modified since last request). The content response is empty, meaning the content is read from the cache.

The fix update `mGotNonEmptyResponse` variable as soon as there is something to read (from downloaded bytes or from cache), not only when we download bytes.

I cannot see a way to test different returned http code in `test_wfs_provider.py`, did I miss something?

## Checklist

### DIFF
--- a/src/providers/wfs/qgsbasenetworkrequest.cpp
+++ b/src/providers/wfs/qgsbasenetworkrequest.cpp
@@ -174,6 +174,7 @@ bool QgsBaseNetworkRequest::sendGET( const QUrl &url, const QString &acceptHeade
       // * or the owner thread of mReply is currently not doing anything because it's blocked in future.waitForFinished() (if it is the main thread)
       connect( mReply, &QNetworkReply::finished, this, &QgsBaseNetworkRequest::replyFinished, Qt::DirectConnection );
       connect( mReply, &QNetworkReply::downloadProgress, this, &QgsBaseNetworkRequest::replyProgress, Qt::DirectConnection );
+      connect( mReply, &QNetworkReply::readyRead, this, &QgsBaseNetworkRequest::replyReadyRead, Qt::DirectConnection );
 
       if ( synchronous )
       {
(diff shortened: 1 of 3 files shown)


rouault:
>  The content response is empty, meaning the content is read from the cache.

Just curious: how does the server know that we have a cache ? Is there something we send in the request to indicated we might have a cache ?

> I cannot see a way to test different returned http code in test_wfs_provider.py, did I miss something?

No, the testing framework used doesn't allow this. That said GDAL autotest suite has the capability to mock HTTP requests : webserver component in https://github.com/OSGeo/gdal/blob/master/autotest/pymod/webserver.py , and example of use in https://github.com/OSGeo/gdal/blob/master/autotest/ogr/ogr_oapif.py#L136 . Could likely be ported to QGIS with little modifications

nyalldawson:
these changes would likely need to be ported to QgsBlockingNetworkRequest also

troopa81:
> Just curious: how does the server know that we have a cache ? Is there something we send in the request to indicated we might have a cache ?

Yes, it's the flag [If-Modified-Since](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since). I think the flag is added by Qt when we enable request cache.

> Could likely be ported to QGIS with little modifications

I'll give it a try, thanks

troopa81:
> these changes would likely need to be ported to QgsBlockingNetworkRequest also

Indeed, I'll do it

troopa81:
@nyalldawson  
> these changes would likely need to be ported to QgsBlockingNetworkRequest also

Actually no, because the problem here is that while we make the request we can start [reading](https://github.com/troopa81/QGIS/blob/fix_wfs_cache/src/providers/wfs/qgswfsfeatureiterator.cpp#L451) the response content, and so the response content is empty when [reply is finished](https://github.com/troopa81/QGIS/blob/fix_wfs_cache/src/providers/wfs/qgsbasenetworkrequest.cpp#L428)

But QgsBlockingNetworkRequest is completely synchronous, no other program is supposed to read the content during the request, so its fine.

troopa81:
@rouault I finally end up using simple python http.server like it was done in qgsblockingnetworkrequest for unit test

rouault:
>  I finally end up using simple python http.server like it was done in qgsblockingnetworkrequest for unit test

Looks good

nyalldawson:
@rouault is that an approval to merge?

troopa81:
Thanks @rouault

PR #32822, opened 2019-11-13
Topology is unavailable when user cannot read topology.topology

### LINKED ISSUE #32726: Unable to load layers from PostgreSQL 9.6 / PostGIS using the browser
After upgrading from QGIS 3.8.2 to 3.10.0, I am no longer able to see my tables in the QGIS Browser under PostGIS:

public --> failed to get layers
topology --> failed to get layers

I am connecting to a PostgreSQL/PostGIS database hosted via a2hosting.com and am currently unable to change schema privileges using the postgres user.

qgis.log file shows the following:

PostGIS(1): Erroneous query: SELECT l.f_table_name,l.f_table_schema,l.f_geometry_column,upper(l.type),l.srid,l.coord_dimension,c.relkind,obj_description(c.oid),array_agg(a.attname), count(CASE WHEN t.typname IN ('geometry','geography','topogeometry','raster') THEN 1 ELSE NULL END) , 1  FROM geometry_columns l,pg_class c,pg_namespace n,pg_attribute a,pg_type t WHERE c.relname=l.f_table_name AND l.f_table_schema=n.nspname AND NOT a.attisdropped AND a.attrelid=c.oid AND a.atttypid=t.oid AND a.attnum>0 AND n.oid=c.relnamespace AND has_schema_privilege(n.nspname,'usage') AND has_table_privilege('"'||n.nspname||'"."'||c.relname||'"','select') AND l.f_table_schema='public' GROUP BY 1,2,3,4,5,6,7,c.oid,11 UNION SELECT l.f_table_name,l.f_table_schema,l.f_geography_column,upper(l.type),l.srid,2,c.relkind,obj_description(c.oid),array_agg(a.attname), count(CASE WHEN t.typname IN ('geometry','geography','topogeometry','raster') THEN 1 ELSE NULL END) , 1  FROM geography_columns l,pg_class c,pg_namespace n,pg_attribute a,pg_type t WHERE c.relname=l.f_table_name AND l.f_table_schema=n.nspname AND NOT a.attisdropped AND a.attrelid=c.oid AND a.atttypid=t.oid AND a.attnum>0 AND n.oid=c.relnamespace AND has_schema_privilege(n.nspname,'usage') AND has_table_privilege('"'||n.nspname||'"."'||c.relname||'"','select') AND l.f_table_schema='public' GROUP BY 1,2,3,4,5,6,7,c.oid,11 UNION SELECT l.table_name,l.schema_name,l.feature_column,CASE WHEN l.feature_type = 1 THEN 'MULTIPOINT' WHEN l.feature_type = 2 THEN 'MULTILINESTRING' WHEN l.feature_type = 3 THEN 'MULTIPOLYGON' WHEN l.feature_type = 4 THEN 'GEOMETRYCOLLECTION' END AS type,(SELECT srid FROM topology.topology t WHERE l.topology_id=t.id),2,c.relkind,obj_description(c.oid),array_agg(a.attname), count(CASE WHEN t.typname IN ('geometry','geography','topogeometry','raster') THEN 1 ELSE NULL END) , 1  FROM topology.layer l,pg_class c,pg_namespace n,pg_attribute a,pg_type t WHERE c.relname=l.table_name AND l.schema_name=n.nspname AND NOT a.attisdropped AND a.attrelid=c.oid AND a.atttypid=t.oid AND a.attnum>0 AND n.oid=c.relnamespace AND has_schema_privilege(n.nspname,'usage') AND has_table_privilege('"'||n.nspname||'"."'||c.relname||'"','select') AND l.schema_name='public' GROUP BY 1,2,3,4,5,6,7,c.oid,11 UNION SELECT l."r_table_name",l."r_table_schema",l."r_raster_column",'RASTER',l.srid,2,c.relkind,obj_description(c.oid),array_agg(a.attname), count(CASE WHEN t.typname IN ('geometry','geography','topogeometry','raster') THEN 1 ELSE NULL END) , 1  FROM raster_columns l,pg_class c,pg_namespace n,pg_attribute a,pg_type t WHERE c.relname=l."r_table_name" AND l."r_table_schema"=n.nspname AND NOT a.attisdropped AND a.attrelid=c.oid AND a.atttypid=t.oid AND a.attnum>0 AND n.oid=c.relnamespace AND has_schema_privilege(n.nspname,'usage') AND has_table_privilege('"'||n.nspname||'"."'||c.relname||'"','select') AND l."r_table_schema"='public' GROUP BY 1,2,3,4,5,6,7,c.oid,11 ORDER BY 2,1,3 returned 7 [ERROR:  permission denied for schema topology
LINE 1: ...ometry','raster') THEN 1 ELSE NULL END) , 1  FROM topology.l...
                                                             ^
]PostGIS(1): NOTICE: WARNING:  there is no transaction in progressPostGIS(1): Unable to get list of spatially enabled tables from the databasePostGIS(1): Erroneous query: SELECT l.f_table_name,l.f_table_schema,l.f_geometry_column,upper(l.type),l.srid,l.coord_dimension,c.relkind,obj_description(c.oid),array_agg(a.attname), count(CASE WHEN t.typname IN ('geometry','geography','topogeometry','raster') THEN 1 ELSE NULL END) , 1  FROM geometry_columns l,pg_class c,pg_namespace n,pg_attribute a,pg_type t WHERE c.relname=l.f_table_name AND l.f_table_schema=n.nspname AND NOT a.attisdropped AND a.attrelid=c.oid AND a.atttypid=t.oid AND a.attnum>0 AND n.oid=c.relnamespace AND has_schema_privilege(n.nspname,'usage') AND has_table_privilege('"'||n.nspname||'"."'||c.relname||'"','select') AND l.f_table_schema='topology' GROUP BY 1,2,3,4,5,6,7,c.oid,11 UNION SELECT l.f_table_name,l.f_table_schema,l.f_geography_column,upper(l.type),l.srid,2,c.relkind,obj_description(c.oid),array_agg(a.attname), count(CASE WHEN t.typname IN ('geometry','geography','topogeometry','raster') THEN 1 ELSE NULL END) , 1  FROM geography_columns l,pg_class c,pg_namespace n,pg_attribute a,pg_type t WHERE c.relname=l.f_table_name AND l.f_table_schema=n.nspname AND NOT a.attisdropped AND a.attrelid=c.oid AND a.atttypid=t.oid AND a.attnum>0 AND n.oid=c.relnamespace AND has_schema_privilege(n.nspname,'usage') AND has_table_privilege('"'||n.nspname||'"."'||c.relname||'"','select') AND l.f_table_schema='topology' GROUP BY 1,2,3,4,5,6,7,c.oid,11 UNION SELECT l.table_name,l.schema_name,l.feature_column,CASE WHEN l.feature_type = 1 THEN 'MULTIPOINT' WHEN l.feature_type = 2 THEN 'MULTILINESTRING' WHEN l.feature_type = 3 THEN 'MULTIPOLYGON' WHEN l.feature_type = 4 THEN 'GEOMETRYCOLLECTION' END AS type,(SELECT srid FROM topology.topology t WHERE l.topology_id=t.id),2,c.relkind,obj_description(c.oid),array_agg(a.attname), count(CASE WHEN t.typname IN ('geometry','geography','topogeometry','raster') THEN 1 ELSE NULL END) , 1  FROM topology.layer l,pg_class c,pg_namespace n,pg_attribute a,pg_type t WHERE c.relname=l.table_name AND l.schema_name=n.nspname AND NOT a.attisdropped AND a.attrelid=c.oid AND a.atttypid=t.oid AND a.attnum>0 AND n.oid=c.relnamespace AND has_schema_privilege(n.nspname,'usage') AND has_table_privilege('"'||n.nspname||'"."'||c.relname||'"','select') AND l.schema_name='topology' GROUP BY 1,2,3,4,5,6,7,c.oid,11 UNION SELECT l."r_table_name",l."r_table_schema",l."r_raster_column",'RASTER',l.srid,2,c.relkind,obj_description(c.oid),array_agg(a.attname), count(CASE WHEN t.typname IN ('geometry','geography','topogeometry','raster') THEN 1 ELSE NULL END) , 1  FROM raster_columns l,pg_class c,pg_namespace n,pg_attribute a,pg_type t WHERE c.relname=l."r_table_name" AND l."r_table_schema"=n.nspname AND NOT a.attisdropped AND a.attrelid=c.oid AND a.atttypid=t.oid AND a.attnum>0 AND n.oid=c.relnamespace AND has_schema_privilege(n.nspname,'usage') AND has_table_privilege('"'||n.nspname||'"."'||c.relname||'"','select') AND l."r_table_schema"='topology' GROUP BY 1,2,3,4,5,6,7,c.oid,11 ORDER BY 2,1,3 returned 7 [ERROR:  permission denied for schema topology
LINE 1: ...ometry','raster') THEN 1 ELSE NULL END) , 1  FROM topology.l...
                                                             ^
]PostGIS(1): NOTICE: WARNING:  there is no transaction in progressPostGIS(1): Unable to get list of spatially enabled tables from the database

strk:
Fixes determination of tables from metadata when user has no
privileges on topology. The bug was introduced by
commit bbdbca418c0eeb5a349fc257ad033adad7b4fc47

References #32002

Closes #32726

### DIFF
--- a/src/providers/postgres/qgspostgresconn.cpp
+++ b/src/providers/postgres/qgspostgresconn.cpp
@@ -1074,7 +1074,23 @@ QString QgsPostgresConn::postgisVersion()
   mTopologyAvailable = false;
   if ( mPostgisVersionMajor > 1 )
   {
-    QgsPostgresResult result( PQexec( QStringLiteral( "SELECT EXISTS ( SELECT c.oid FROM pg_class AS c JOIN pg_namespace AS n ON c.relnamespace=n.oid WHERE n.nspname='topology' AND c.relname='topology' )" ) ) );
+    // NOTE: CASE syntax is used to avoid an exception when
+    //       topology.topology does not exist
+    // See
+    // https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-EXPRESS-EVAL
+    QgsPostgresResult result(
+      PQexec(
+        QStringLiteral(
+          "SELECT has_schema_privilege(n.oid, 'usage')"
+          "   AND has_table_privilege(t.oid, 'select')"
+          "   AND has_table_privilege(l.oid, 'select')"
+          "  FROM pg_namespace n, pg_class t, pg_class l"
+          " WHERE n.nspname = 'topology'"
+          "   AND t.relnamespace = n.oid"
+          "   AND l.relnamespace = n.oid"
+          "   AND t.relname = 'topology'"
+          "   AND l.relname = 'layer'"
+        ) ) );
     if ( result.PQntuples() >= 1 && result.PQgetvalue( 0, 0 ) == QLatin1String( "t" ) )
     {
       mTopologyAvailable = true;
--- a/src/providers/postgres/qgspostgresconn.cpp
+++ b/src/providers/postgres/qgspostgresconn.cpp
@@ -353,10 +353,9 @@ QgsPostgresConn::QgsPostgresConn( const QString &conninfo, bool readOnly, bool s
     {
       QgsMessageLog::logMessage( tr( "Your PostGIS installation has no GEOS support. Feature selection and identification will not work properly. Please install PostGIS with GEOS support (http://geos.refractions.net)" ), tr( "PostGIS" ) );
     }
-
-    if ( hasTopology() )
+    else
     {
-      QgsDebugMsg( QStringLiteral( "Topology support available!" ) );
+      QgsDebugMsg( QStringLiteral( "GEOS support available!" ) );
     }
   }
 
@@ -1095,6 +1094,15 @@ QString QgsPostgresConn::postgisVersion()
     {
       mTopologyAvailable = true;
     }
+  }
+
+  if ( mTopologyAvailable )
+  {
+    QgsDebugMsg( QStringLiteral( "Topology support available!" ) );
+  }
+  else
+  {
+    QgsDebugMsg( QStringLiteral( "Topology support not available :(" ) );
   }
 
   mGotPostgisVersion = true;
--- a/src/providers/postgres/qgspostgresconn.cpp
+++ b/src/providers/postgres/qgspostgresconn.cpp
@@ -31,6 +31,7 @@
 #include "qgspostgresstringutils.h"
 
 #include <QApplication>
+#include <QStringList>
 #include <QThread>
 
 #include <climits>
@@ -403,13 +404,28 @@ void QgsPostgresConn::unref()
 }
 
 /* private */
+QStringList QgsPostgresConn::supportedSpatialTypes()
+{
+  QStringList supportedSpatialTypes;
+
+  supportedSpatialTypes << quotedValue( "geometry" )
+                        << quotedValue( "geography" )
+                        << quotedValue( "pcpatch" );
+  if ( hasRaster() ) supportedSpatialTypes << quotedValue( "raster" );
+  if ( hasTopology() ) supportedSpatialTypes << quotedValue( "topogeometry" );
+
+  return supportedSpatialTypes;
+}
+
+/* private */
 // TODO: deprecate this function
 void QgsPostgresConn::addColumnInfo( QgsPostgresLayerProperty &layerProperty, const QString &schemaName, const QString &viewName, bool fetchPkCandidates )
 {
   // TODO: optimize this query when pk candidates aren't needed
   //       could use array_agg() and count()
   //       array output would look like this: "{One,tWo}"
-  QString sql = QStringLiteral( "SELECT attname, CASE WHEN typname in ('geometry','geography','topogeometry','raster') THEN 1 ELSE null END AS isSpatial FROM pg_attribute JOIN pg_type ON atttypid=pg_type.oid WHERE attrelid=regclass('%1.%2') AND NOT attisdropped AND attnum>0 ORDER BY attnum" )
+  QString sql = QStringLiteral( "SELECT attname, CASE WHEN typname in (%1) THEN 1 ELSE null END AS isSpatial FROM pg_attribute JOIN pg_type ON atttypid=pg_type.oid WHERE attrelid=regclass('%2.%3') AND NOT attisdropped AND attnum>0 ORDER BY attnum" )
+                .arg( supportedSpatialTypes().join( ',' ) )
                 .arg( quotedIdentifier( schemaName ),
                       quotedIdentifier( viewName ) );
   QgsDebugMsg( "getting column info: " + sql );
@@ -532,7 +548,7 @@ bool QgsPostgresConn::getTableInfo( bool searchGeometryColumnsOnly, bool searchP
     // Can't use regclass here because table must exist, else error occurs.
     sql = QString( "SELECT %1,%2,%3,%4,%5,%6,c.relkind,obj_description(c.oid),"
                    "array_agg(a.attname), "
-                   "count(CASE WHEN t.typname IN ('geometry','geography','topogeometry','raster') THEN 1 ELSE NULL END) "
+                   "count(CASE WHEN t.typname IN (%9) THEN 1 ELSE NULL END) "
                    ", %8 "
                    " FROM %7 l,pg_class c,pg_namespace n,pg_attribute a,pg_type t"
                    " WHERE c.relname=%1"
@@ -547,6 +563,7 @@ bool QgsPostgresConn::getTableInfo( bool searchGeometryColumnsOnly, bool searchP
                  )
           .arg( tableName, schemaName, columnName, typeName, sridName, dimName, gtableName )
           .arg( 1 )
+          .arg( supportedSpatialTypes().join( ',' ) )
           ;
 
     if ( searchPublicOnly )
@@ -663,22 +680,23 @@ bool QgsPostgresConn::getTableInfo( bool searchGeometryColumnsOnly, bool searchP
   if ( !searchGeometryColumnsOnly )
   {
     // Now have a look for spatial columns that aren't in the geometry_columns table.
-    QString sql = "SELECT"
-                  " c.relname"
-                  ",n.nspname"
-                  ",a.attname"
-                  ",c.relkind"
-                  ",CASE WHEN t.typname IN ('geometry','geography','topogeometry') THEN t.typname ELSE b.typname END AS coltype"
-                  ",obj_description(c.oid)"
-                  " FROM pg_attribute a"
-                  " JOIN pg_class c ON c.oid=a.attrelid"
-                  " JOIN pg_namespace n ON n.oid=c.relnamespace"
-                  " JOIN pg_type t ON t.oid=a.atttypid"
-                  " LEFT JOIN pg_type b ON b.oid=t.typbasetype"
-                  " WHERE c.relkind IN ('v','r','m','p')"
-                  " AND has_schema_privilege( n.nspname, 'usage' )"
-                  " AND has_table_privilege( c.oid, 'select' )"
-                  " AND (t.typname IN ('geometry','geography','topogeometry') OR b.typname IN ('geometry','geography','topogeometry','pcpatch','raster'))";
+    QString sql = QStringLiteral( "SELECT"
+                                  " c.relname"
+                                  ",n.nspname"
+                                  ",a.attname"
+                                  ",c.relkind"
+                                  ",CASE WHEN t.typname IN (%1) THEN t.typname ELSE b.typname END AS coltype"
+                                  ",obj_description(c.oid)"
+                                  " FROM pg_attribute a"
+                                  " JOIN pg_class c ON c.oid=a.attrelid"
+                                  " JOIN pg_namespace n ON n.oid=c.relnamespace"
+                                  " JOIN pg_type t ON t.oid=a.atttypid"
+                                  " LEFT JOIN pg_type b ON b.oid=t.typbasetype"
+                                  " WHERE c.relkind IN ('v','r','m','p')"
+                                  " AND has_schema_privilege( n.nspname, 'usage' )"
+                                  " AND has_table_privilege( c.oid, 'select' )"
+                                  " AND (t.typname IN (%1) OR b.typname IN (%1))" )
+                  .arg( supportedSpatialTypes().join( ',' ) );
 
     // user has select privilege
     if ( searchPublicOnly )
--- a/src/providers/postgres/qgspostgresconn.cpp
+++ b/src/providers/postgres/qgspostgresconn.cpp
@@ -543,7 +543,7 @@ bool QgsPostgresConn::getTableInfo( bool searchGeometryColumnsOnly, bool searchP
                    " AND a.attnum>0"
                    " AND n.oid=c.relnamespace"
                    " AND has_schema_privilege(n.nspname,'usage')"
-                   " AND has_table_privilege('\"'||n.nspname||'\".\"'||c.relname||'\"','select')" // user has select privilege
+                   " AND has_table_privilege(c.oid,'select')" // user has select privilege
                  )
           .arg( tableName, schemaName, columnName, typeName, sridName, dimName, gtableName )
           .arg( 1 )
@@ -662,7 +662,7 @@ bool QgsPostgresConn::getTableInfo( bool searchGeometryColumnsOnly, bool searchP
   //search for geometry columns in tables that are not in the geometry_columns metatable
   if ( !searchGeometryColumnsOnly )
   {
-    // Now have a look for geometry columns that aren't in the geometry_columns table.
+    // Now have a look for spatial columns that aren't in the geometry_columns table.
     QString sql = "SELECT"
                   " c.relname"
                   ",n.nspname"
@@ -677,7 +677,7 @@ bool QgsPostgresConn::getTableInfo( bool searchGeometryColumnsOnly, bool searchP
                   " LEFT JOIN pg_type b ON b.oid=t.typbasetype"
                   " WHERE c.relkind IN ('v','r','m','p')"
                   " AND has_schema_privilege( n.nspname, 'usage' )"
-                  " AND has_table_privilege( QUOTE_IDENT(n.nspname) || '.' || QUOTE_IDENT(c.relname), 'select' )"
+                  " AND has_table_privilege( c.oid, 'select' )"
                   " AND (t.typname IN ('geometry','geography','topogeometry') OR b.typname IN ('geometry','geography','topogeometry','pcpatch','raster'))";
 
     // user has select privilege
@@ -806,7 +806,7 @@ bool QgsPostgresConn::getTableInfo( bool searchGeometryColumnsOnly, bool searchP
                   ",pg_attribute a"
                   " WHERE pg_namespace.oid=pg_class.relnamespace"
                   " AND has_schema_privilege(pg_namespace.nspname,'usage')"
-                  " AND has_table_privilege('\"' || pg_namespace.nspname || '\".\"' || pg_class.relname || '\"','select')"
+                  " AND has_table_privilege(pg_class.oid,'select')"
                   " AND pg_class.relkind IN ('v','r','m','p')"
                   " AND pg_class.oid = a.attrelid"
                   " AND NOT a.attisdropped"
--- a/src/providers/postgres/qgspostgresconn.cpp
+++ b/src/providers/postgres/qgspostgresconn.cpp
@@ -208,8 +208,6 @@ QgsPostgresConn::QgsPostgresConn( const QString &conninfo, bool readOnly, bool s
   , mPostgresqlVersion( 0 )
   , mPostgisVersionMajor( 0 )
   , mPostgisVersionMinor( 0 )
-  , mGistAvailable( false )
-  , mProjAvailable( false )
   , mPointcloudAvailable( false )
   , mRasterAvailable( false )
   , mUseWkbHex( false )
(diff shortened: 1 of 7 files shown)


strk:
@elpaso, writing a test for this, I noticed that QgsAbstractDatabaseProviderConnection doesn't have a TableFlag for TopoGeometries, should it ?

elpaso:
@strk I don't really know, but feel free to add it if you think it can be useful. The flags are mainly meant for filtering when searching through tables in a given connection.

strk:
I'm thinking we'll need the PG test scripts to create a `qgis_test_role_unprivileged` role in the database cluster

strk:
@elpaso: how can I get a db connection to execute a query from a QgsAbstractDatabaseProviderConnection ?

elpaso:
Have a look to https://github.com/qgis/QGIS/blob/master/tests/src/python/test_qgsproviderconnection_postgres.py

strk:
Figured: executeSql.
Now... creating a role is not enough, I also need to force closing a connection to obtain a new one, as each new connection will (with the code in this PR) determine if topology is available for the current user. Is there a way to ensure a QgsProviderRegistry::providerMetadata('postgres').createConnection returns a _new_ connection ?

strk:
Figured: executeSql.

Now... creating a role is not enough, I also need to force closing a connection to obtain a new one, as each new connection will (with the code in this PR) determine if topology is available for the current user. Is there a way to ensure a QgsProviderRegistry::providerMetadata('postgres').createConnection returns a _new_ connection ? @elpaso maybe we should have a QgsAbstractDatabaseProviderConnection::reconnect() ?

elpaso:
> @elpaso maybe we should have a QgsAbstractDatabaseProviderConnection::reconnect() ?

I don't think we need/want that, just destroy the connection and create a new one.

The name "Connection" in  "QgsAbstractDatabaseProviderConnection" may be misleading: it has nothing to do with low level DB connections, the term refers to the connections to the datasources stored in QGIS settings, and provides an interface to create/delete store and retrieve them from the settingsg, it also provides a unified interface for some common operations (list of schemas and tables etc. etc.).

I'd prefer to keep that interface simple, reconnect should be probably handled transparently at a lower level in the provider.

strk:
I'm afraid there's some pooling going on. Anyway I also have the other problem of switching to unprivileged user _before_ performing connection checks, so I guess we don't just need a role but we really need a whole new user, and we'd need to assume we can connect via password scheme as the new user, which starts becoming a bit dangerous for a testcase ... Not sure how to proceed. Maybe we could get the provider to support some env variable to request switching to a specific role upon connecting, not sure...

strk:
I found a way to connect as a different role: adding something like: ` options=-c\ role=qgis_test_role` to the connection string. But I'm not sure the createConnection() API of what's returned by `QgsProviderRegistry.instance().providerMetadata('postgres')` is really honouring that additional parameter, rather than returning me the already existing connection, because I don't see a new connection being created in the logs

strk:
Confirmed: md.createConnection() is not triggering creationg of a connection, how do I control when the connection is actually made ? I'm confused

strk:
I'm afraid `options` is just lost when going through QgsDataSourceUri class

strk:
Bingo:
```
      else if ( pname == QLatin1String( "options" ) )
      { 
        QgsDebugMsg( QStringLiteral( "backend debug options ignored" ) );
      }
```

The above code snippet is in `src/core/qgsdatasourceuri.cpp` in the body of method `QgsDataSourceUri::QgsDataSourceUri( const QString &u )`

Isn't it wonderful to follow the white rabbit in testsuite writing !?

 @jef-n  any reason to ignore `options` ? They'd be handy in this case

strk:
I dunno why those "options" are reported to be `backend debug options`...

jef-n:
> @jef-n any reason to ignore `options` ? They'd be handy in this case

Dates back to when `QgsDataSourceUri` (or back then `QgsDataSourceURI`) didn't even have `setParam`.  All provider specific handling of parameters could have been refactored out that way.

strk:
So, jef, do you think I should rewrite the test to use `setParam` instead of proper support for 'options' ?
In that case, I think there's some logic in the connection provider (new API?) to strip out any non-connection related parameter from the URI which would still give me the same connection instead of creating a new one, when passing a custom parameter.

strk:
Writing as setParam won't help because ultimately the connection provider is passing the URI to QgsDatasourceURI::connectionInfo() which currently does NOT include parameters in the returned string.

strk:
@elpaso in order to separate the two issues of QgsDatasourceUri deprecation and the fix for this regression.. I could use PGOPTIONS env variable, which is directly supported by libpq, then all I need to do is make sure a _new_ connection is created. You suggest to "destroy" a connection but I'm afraid the lower-level pooling is still keeping it alive, and reachable by baseKey

This is getting long to get right, I'm tempted to merge just the codefix, and backport, w/out a test :(

m-kuhn:
Or you add the login user as you suggested?
What's the risk of having a login user on a temporarily existing database?

strk:
>  What's the risk of having a login user on a temporarily existing database?

Users in PostgreSQL are per-cluster, not per database

elpaso:
@strk my first thought would be to create a test role "no_topology" and revoke access to topology, then use that user for the test. That's what I've done (interactively) locally to verify this issue before debugging and "git blamining".

elpaso:
mmm, if you need a temporary DB, look at https://github.com/qgis/QGIS/blob/master/tests/src/python/test_authmanager_ogr_postgres.py

strk:
@elpaso that's what my test does, creates a role, revokes select from topology.topology.
The problem is I cannot convince createConnection() to use the new role via `options` (it is stripped).
I didn't try via PGOPTIONS env variable. It could work, as long as the connection provider doesn't just return me a previously opened connection because the "connection parts of the uri" were not changed :(

strk:
If we can have a global test cluster that'd surely help. It takes time to create a cluster thus I'd rather have it created only once, for a testsuite run. And ideally only created if not already existent, so we save time on subsequent testsuite runs

m-kuhn:
> > What's the risk of having a login user on a temporarily existing database?
> 
> Users in PostgreSQL are per-cluster, not per database

s/temporarily existing database/temporarily existing cluster/

elpaso:
> @elpaso that's what my test does, creates a role, revokes select from topology.topology.
> The problem is I cannot convince createConnection() to use the new role via `options` (it is stripped).

Sorry, I don't understand: you can create a new connection passing a string URI and in the URI you can set username and password of the new no_topology role.
`QgsPostgresProviderMetadata::createConnection( const QString &uri, const QVariantMap &configuration )`

```python
conn = md.createConnection("your uri goes here", {})
```

strk:
> Sorry, I don't understand: you can create a new connection passing a string URI and in the URI you can set username and password of the new no_topology role.

This is what I was trying to avoid (creating a user with login
capabilities). Passing a role via `options=-c\ role=newrole`
allows that.

The way my PostgreSQL cluster is setup is to use peer authentication
for "localhost", so password does not work there:

  PostGIS(1): Connection to database failed
  FATAL:  Peer authentication failed for user "qgis_test_user"

Yes, I can change my configuration, but we're just adding more
steps to the initial setup (the `-c role=newrole` trick was meant
to make things simpler)

strk:
I just realized that my test was bogus. I was checking that the "TopoLayer1" table would not be found, but I didn't consider that _all_ tables (even non-spatial ones) are seeked for, so that table gets of course visible as well. I'll tweak the test to only check for  spatial tables and see if that helps spotting the difference between connection with a user having permissions and another NOT having permissions.

I suspect there's a lot more to test in this reguard (and I'd like to try the ROLE again later)

strk:
@elpaso I'm a bit lost with the saveConnection / deleteConnection API and in particular with the tests performed by tests/src/python/test_qgsproviderconnection_base.py which keep failing for me.

There are 2 tests failing.

First one:

  - line 261, in _test_operations:
```
    self.assertEqual(list(md.connections().values()), [])
AssertionError: Lists differ: [<qgis._core.QgsAbstractDatabaseProviderConnection object at 0x7f9504f683a8>] != []
```

I'm guessing this is due to me creating that new connection and not cleaning up afterwards ?
How is a connection removed ? I see calls to `md.deleteConnection` but they are passed a connection _name_ rather than a connection object, so how do I tell what to pass there ?
I've also tried to saveConnection(conn2, 'newname') and deleteConnection('newname') but that test keeps failing.

strk:
@elpaso, is it normal that running `ctest -R ProviderConnectionPostgres` triggers tests in base class, btw ? Does this mean we're running base class tests for each connection provider test case ?

elpaso:
> @elpaso, is it normal that running `ctest -R ProviderConnectionPostgres` triggers tests in base class, btw ? Does this mean we're running base class tests for each connection provider test case ?

Yes, that's the pattern with provider tests, we have a base test that all providers must pass and some specific configuration in sub classes.

strk:
@elpaso those tests check that the list of connections is empty after deleting a connection named 'qgis_test1'. The list `md.connections()` is found to be non-empty, but I never saved the connection I created so I don't know why it should not be empty. What's being automatically saving that connection ?

strk:
Funny, the test fails on my machine but passes on Travis ?! (the one about saving connections)

strk:
@elpaso can it be that the test (base class) is written in a way to pass when invoking multiple subclass tests but can fail when calling a single subclass test ?
I'm running the testsuite as such:

> ctest -R ProviderConnectionPostgres --output-on-failure

strk:
I'm afraid the fix for this bug is only partial, as other permission related issues exist: #32843

src/providers/postgres/qgspostgresconn.cpp
```diff
@@ -1056,13 +1073,38 @@ QString QgsPostgresConn::postgisVersion()
...
     }
   }
 
+  if ( mTopologyAvailable )
+  {
+    QgsDebugMsg( QStringLiteral( "Topology support available!" ) );
+  }
+  else
+  {
+    QgsDebugMsg( QStringLiteral( "Topology support not available :(" ) );
```
m-kuhn:
```suggestion
    QgsDebugMsg( QStringLiteral( "Topology support not available" ) );
```

src/providers/postgres/qgspostgresconn.cpp
```diff
@@ -1056,13 +1073,38 @@ QString QgsPostgresConn::postgisVersion()
...
+        ) ) );
     if ( result.PQntuples() >= 1 && result.PQgetvalue( 0, 0 ) == QLatin1String( "t" ) )
     {
       mTopologyAvailable = true;
     }
   }
 
+  if ( mTopologyAvailable )
+  {
+    QgsDebugMsg( QStringLiteral( "Topology support available!" ) );
```
m-kuhn:
```suggestion
    QgsDebugMsg( QStringLiteral( "Topology support available :)" ) );
```

src/providers/postgres/qgspostgresconn.cpp
```diff
@@ -402,14 +402,29 @@ void QgsPostgresConn::unref()
...
+/* private */
+QStringList QgsPostgresConn::supportedSpatialTypes()
+{
+  QStringList supportedSpatialTypes;
+
+  supportedSpatialTypes << quotedValue( "geometry" )
+                        << quotedValue( "geography" )
+                        << quotedValue( "pcpatch" );
+  if ( hasRaster() ) supportedSpatialTypes << quotedValue( "raster" );
+  if ( hasTopology() ) supportedSpatialTypes << quotedValue( "topogeometry" );
```
m-kuhn: If you want to follow the normally-used style, line-break after if clauses. Aka `astyle mkhun`

src/providers/postgres/qgspostgresconn.cpp
```diff
@@ -402,14 +402,29 @@ void QgsPostgresConn::unref()
   delete this;
 }
 
+/* private */
+QStringList QgsPostgresConn::supportedSpatialTypes()
```
m-kuhn:
```suggestion
QStringList QgsPostgresConn::supportedSpatialTypes() const
```

m-kuhn (approved):
Let's focus on the good things :)

Nice, just some nitpicking.

src/providers/postgres/qgspostgresconn.cpp
```diff
@@ -662,23 +678,24 @@ bool QgsPostgresConn::getTableInfo( bool searchGeometryColumnsOnly, bool searchP
...
-                  " JOIN pg_class c ON c.oid=a.attrelid"
-                  " JOIN pg_namespace n ON n.oid=c.relnamespace"
-                  " JOIN pg_type t ON t.oid=a.atttypid"
-                  " LEFT JOIN pg_type b ON b.oid=t.typbasetype"
-                  " WHERE c.relkind IN ('v','r','m','p')"
-                  " AND has_schema_privilege( n.nspname, 'usage' )"
-                  " AND has_table_privilege( QUOTE_IDENT(n.nspname) || '.' || QUOTE_IDENT(c.relname), 'select' )"
-                  " AND (t.typname IN ('geometry','geography','topogeometry') OR b.typname IN ('geometry','geography','topogeometry','pcpatch','raster'))";
+    // Now have a look for spatial columns that aren't in the geometry_columns table.
+    QString sql = QStringLiteral( "SELECT"
```
elpaso: I'd like to see raw strings on all these multi-line strings, but it's not really important, just for readability.

elpaso:
> @elpaso can it be that the test (base class) is written in a way to pass when invoking multiple subclass tests but can fail when calling a single subclass test ?
> I'm running the testsuite as such:
> 
> > ctest -R ProviderConnectionPostgres --output-on-failure

No, I don't think so, but honestly I always run python tests directly from the command line.

elpaso:
> @elpaso those tests check that the list of connections is empty after deleting a connection named 'qgis_test1'. The list `md.connections()` is found to be non-empty, but I never saved the connection I created so I don't know why it should not be empty. What's being automatically saving that connection ?

AFAIK nothing is automatically saving a connection.

elpaso (approved):
Looks good to me, thank you!

src/providers/postgres/qgspostgresconn.cpp
strk: Uhmpf, I liked that smiley so much ! It's a debug message after all, did we really get so serious ? :)

src/providers/postgres/qgspostgresconn.cpp
strk: Ah, now I see :)

src/providers/postgres/qgspostgresconn.cpp
strk: Can you give me a pointer about "raw strings" ?
I'd like things to stay within 80 cols if possible

elpaso: https://stackoverflow.com/questions/22668264/in-c11-can-raw-string-literals-have-multiple-lines

strk:
Argh, the "const" supportedSpatialTypes results in a failure due to many hasThis hasThat functions being non-const, and those in turn are non-const because the lazily call postgisVersion() which make change a cached variable containing postgis version.

Maybe we should postpone this change to a separate PR fixing the whole mess with some mutable or refactoring @m-kuhn ?

strk:
> No, I don't think so, but honestly I always run python tests directly from the command line.

How about adding a nice script to source to setup environment for making it easy to do for all ?

elpaso:
> > No, I don't think so, but honestly I always run python tests directly from the command line.
> 
> How about adding a nice script to source to setup environment for making it easy to do for all ?

I doubt this is possible because everyone uses a different dev setup, but if you are able to make such a script it would be certain useful.

Here is an example if you want to start from it, note that my build dir is in-source tree `${LOCAL_SUFFIX}` in the env script and I use a self-built Qt toolchain in `"/home/${USER}/${LOCAL_SUFFIX}`

from `tests/src/python` execute the test `test_qgsserver_api.py`
```bash
QGIS_DEBUG=2 LC_ALL=en_US.UTF-8  ../../../run_local_test.sh test_qgsserver_api.py
```

run_local_test.sh:
```bash
#!/bin/bash
# Run python3

# Store script directory
THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"

${THIS_DIR}/set_env.sh python3 $@
```

set_env.sh:
```bash
#!/bin/bash

export BUILD_SUFFIX='build-local-qt5132-release'
export LOCAL_SUFFIX='local5_13_2_release'
export LRELEASE="/home/${USER}/${LOCAL_SUFFIX}/bin/lrelease"
export PYLUPDATE="/home/${USER}/${LOCAL_SUFFIX}/bin/pylupdate5"
export QT_DEVICE_PIXEL_RATIO=1
export QGIS_AUTH_DB_DIR_PATH=/home/${USER}/tmp
export QT_QPA_FONTDIR=/usr/share/fonts/truetype/
export LD_LIBRARY_PATH=:/home/${USER}/${LOCAL_SUFFIX}/lib:/home/${USER}/dev/QGIS/${BUILD_SUFFIX}/output/lib:

export PYTHONPATH=$(pwd):/home/${USER}/${LOCAL_SUFFIX}/lib/python3/dist-packages/:\
/home/${USER}/dev/QGIS/${BUILD_SUFFIX}/output/python/:\
/home/${USER}/dev/QGIS/${BUILD_SUFFIX}/output/python/plugins:\
/home/${USER}/dev/QGIS/tests/src/python:\
/usr/lib/python3.6:\
/usr/lib/python3.6/plat-x86_64-linux-gnu:\                                                                                                                             
/usr/lib/python3.6/lib-dynload:\
/usr/local/lib/python3.6/dist-packages/:\
/home/${USER}/.local/lib/python3.6/site-packages:\
/usr/${LOCAL_SUFFIX}/lib/python3.6/dist-packages:\
/usr/lib/python3/dist-packages:\
${PYTHONPATH}

export QGIS_PREFIX_PATH=/home/${USER}/dev/QGIS/${BUILD_SUFFIX}/output

$@

```

Issue #32832, opened 2019-11-13
Cannot specify options in QgsDataSourceURI


strk:
The `options` parameter to QgsDatasourceURI is ignored, by code.
See https://github.com/qgis/QGIS/pull/32822#issuecomment-553437143

This issue is to add support for it, or discuss why it should be discarded
(I tried to use it to set current user)

\cc @jef-n

jef-n:
see https://github.com/qgis/QGIS/pull/32822#issuecomment-553447843

strk:
Closing as "ignored by design", due to:
`QgsDebugMsg( QStringLiteral( "backend debug options ignored" ) );` being found in code

PR #32833, opened 2019-11-13
WIP: Add support for "options" in DatasourceURI

strk:
See #32832

### DIFF
--- a/tests/src/core/testqgsdatasourceuri.cpp
+++ b/tests/src/core/testqgsdatasourceuri.cpp
@@ -49,6 +49,7 @@ void TestQgsDataSourceUri::checkparser_data()
   QTest::addColumn<QgsDataSourceUri::SslMode>( "sslmode" );
   QTest::addColumn<QString>( "sql" );
   QTest::addColumn<QString>( "myparam" );
+  QTest::addColumn<QString>( "options" );
 
 
   QTest::newRow( "oci" )
@@ -70,6 +71,7 @@ void TestQgsDataSourceUri::checkparser_data()
       << QgsDataSourceUri::SslPrefer // sslmode
       << "" // sql
       << "myvalue" // myparam
+      << "" // options
       ;
 
   QTest::newRow( "pgrast" )
@@ -91,6 +93,7 @@ void TestQgsDataSourceUri::checkparser_data()
       << QgsDataSourceUri::SslPrefer // sslmode
       << "" // sql
       << "" // myparam
+      << "" // options
       ;
 
   QTest::newRow( "pgmlsz" )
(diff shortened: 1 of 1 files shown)


elpaso:
If you want to store it see:  https://github.com/qgis/QGIS/blob/master/src/providers/postgres/qgspostgresproviderconnection.cpp#L455

Also, add it to: https://github.com/qgis/QGIS/blob/master/src/providers/postgres/qgspostgresprovider.cpp#L5288

And feel free to deprecate `QgsDataSourceUri`: it the source of all kind of wrong assumptions and IMO it should not be in core unless completely rewritten to something like: "providerkey://provider specific uri here" and then the decode URI from provider metadata should be used.

Btw, aren't those "options" partly redundant? I mean, are you sure that they will not override other URI parameters? And if they do, which one is supposed to take precedence?

jef-n:
I also think that this moves `QgsDataSourceUri` further in the wrong direction - there should be less provider(/postgres) specific stuff in it, instead of more.

strk:
My current problem is writing a test for the fix in #32822

The test needs to use 2 different "roles" for the connection, but I'd like NOT to create a new user with login capabilities to do this. The only way I see to do so is to use PGOPTIONS to set a role which is different from the connection user, so there's no overlap.

Alternatively I could have the database setup script create a user, with a password, but this would mean exposing the target PostgreSQL cluster to a time window in which people can connect using the hard-coded username/password of the qgis test role...

jef-n:
Why not remove all the `if`s that skip parameters and just let it fallthough to the `setParam`?

strk:
> Why not remove all the `if`s that skip parameters and just let it fallthough to the `setParam`?

Could work, I'll try that in pr #32822 and see if it sticks

strk:
Note, from Travis testing, that QgsDatasourceUri is not able to parse the escape sequence anyway:
```
../tests/src/core/testqgsdatasourceuri.cpp:144:37: error: unknown escape sequence '\ ' [-Werror,-Wunknown-escape-sequence]

      << "sql= table=\"\" options=-c\ role=guest "

                                    ^~
```

From https://travis-ci.org/qgis/QGIS/jobs/611410954#L467

So some work would be needed here anyway

strk:
Removing the special handling for `options` still "eats" the options:
```
431: XXXX: creating connection with uridbname='qgis_test' sslmode=disable options=-c\ role=qgis_test_role
XXX ::tables got uri: dbname='qgis_test' port=5432 sslmode=disable
```

somewhere those options get lost, and I guess it's wherever some class (Uri itself?) tries to separate connection-oriented parts from other parts, to determine if a pre-existing connection exists

strk:
Found the stripper: it is `QgsDataSourceUri::connectionInfo` who doesn't care about parameters, do you see that ?

In turn, QgsPostgresProviderConnection::QgsPostgresProviderConnection initializes the uri to whatever is returned from `QgsDataSourceUri::connectionInfo()` so that information is lost (which role to use)

strk:
Ok I'm closing this PR as I see you want QgsDatasourceUri deprecated

PR #32838, opened 2019-11-13
Fix Task manager waitforfinished

troopa81:
## Description

It could lead to crashes (If you access some shared memory data believing the task has finished to run)

This PR was originally proposed #32453 and reverted because it was crashing on Windows and Mac. This new version use semaphore instead of mutex (you cannot lock/unlock mutex from different thread, which is the case here). 

## Checklist

### DIFF
--- a/src/core/qgstaskmanager.cpp
+++ b/src/core/qgstaskmanager.cpp
@@ -96,6 +96,7 @@ void QgsTask::cancel()
   {
     // immediately terminate unstarted jobs
     terminated();
+    mNotStartedMutex.release();
   }
 
   if ( mStatus == Terminated )
--- a/src/core/qgstaskmanager.cpp
+++ b/src/core/qgstaskmanager.cpp
@@ -28,7 +28,9 @@
 QgsTask::QgsTask( const QString &name, Flags flags )
   : mFlags( flags )
   , mDescription( name )
+  , mNotStartedMutex( 1 )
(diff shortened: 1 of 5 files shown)


troopa81:
Is there a difference between `isTravis` and `runFlakyTest` ? `TestQgsTaskManager::managerWithSubTasks` is not always working so I make it flaky.

src/core/qgstaskmanager.cpp
```diff
@@ -41,6 +43,7 @@ QgsTask::~QgsTask()
     delete subTask.task;
   }
   mNotFinishedMutex.unlock();
+  mNotStartedMutex.release();
```
nyalldawson: The qt docs aren't clear -- what happens if there's no resources acquired when this is called?

troopa81: I tested it on Linux and Windows. if there's no resources acquired when release is called, the resources available is increased (by the number given in release call, 1 in this case) and if there is threads waiting on acquire, they wake up.

In this case, there is no thread to wake up because task has to be started or canceled before being deleted by the task manager. And both way release the semaphore and wake up a possible waitForFinished call.

nyalldawson: Ok, let's give it a shot!!

nyalldawson (approved):

Issue #32819, opened 2019-11-13
QGIS Crashing when adding fields to an existing shp file


TonyShep:
## User Feedback

Adding more than one new field to a SHP file.  I have managed to get this to replicate.  Add one field, save. OK. Add a second field QGIS crash.  Have closed, rebooted and get the same behavioue.

## Report Details

**Crash ID**: [fe1921e12f4432d0751980f7cd89aee94750f14c]('https://github.com/qgis/QGIS/search?q=fe1921e12f4432d0751980f7cd89aee94750f14c&type=Issues')

**Stack Trace**
<pre>
memcpy_repmovs memcpy.asm:115
QSortFilterProxyModel::qt_metacast :
QSortFilterProxyModel::removeRows :
QSortFilterProxyModel::sortRole :
QSortFilterProxyModel::`default constructor closure' :
QMetaObject::activate :
QAbstractItemModel::beginRemoveRows :
QgsAttributeTableModel::removeRows qgsattributetablemodel.cpp:177
QgsAttributeTableModel::loadLayer qgsattributetablemodel.cpp:449
QMetaObject::activate :
QMetaObject::activate :
QgsVectorLayer::updateFields qgsvectorlayer.cpp:3610
QgsVectorLayerUndoCommandAddAttribute::redo qgsvectorlayerundocommand.cpp:317
QUndoStack::push :
QgsVectorLayerEditBuffer::addAttribute qgsvectorlayereditbuffer.cpp:274
</pre>

**QGIS Info**
QGIS Version: 3.10.0-A Coruña
QGIS code revision: 6c816b4204
Compiled against Qt: 5.11.2
Running against Qt: 5.11.2
Compiled against GDAL: 2.4.1
Running against GDAL: 2.4.1

**System Info**
CPU Type: x86_64
Kernel Type: winnt
Kernel Version: 10.0.18362

gioman:
@TonyShep Please try with a new QGIS profile (don't install any 3rd party plugins).

gioman:
Closing for lack of feedback, please reopen if necessary.

PR #32841, opened 2019-11-13
Attempt to fix crash when adding fields to existing table

nyalldawson:
Refs #32819

Needs further testing before merge

### DIFF
--- a/src/gui/attributetable/qgsattributetablemodel.cpp
+++ b/src/gui/attributetable/qgsattributetablemodel.cpp
@@ -438,8 +438,6 @@ void QgsAttributeTableModel::loadLayer()
   // wrong number of attributes)
   loadAttributes();
 
-  beginResetModel();
-
   if ( rowCount() != 0 )
   {
     removeRows( 0, rowCount() );
@@ -468,14 +466,12 @@ void QgsAttributeTableModel::loadLayer()
 
         t.restart();
       }
-      featureAdded( mFeat.id(), true );
+      featureAdded( mFeat.id() );
     }
 
     emit finished();
     connect( mLayerCache, &QgsVectorLayerCache::invalidated, this, &QgsAttributeTableModel::loadLayer, Qt::UniqueConnection );
   }
-
-  endResetModel();
 }
 
 


m-kuhn:
What leads to the assumption that it's this call which is broken? The docs seem clear to me.

https://doc.qt.io/qt-5/qabstractitemmodel.html#beginResetModel

> You must call this function before resetting any internal data structures in your model or proxy model.

nyalldawson:
@m-kuhn 

A couple of things:

1. I reproduced the crash from #32819 and the stacktrace related to the model reset
2. the code surrounded by the begin/end reset model correctly uses begin/end inserting rows. This leads me to believe that the reset is redundant, and I'm unsure whether nesting begin reset/insert calls like this is allowed by Qt
3. These lines were added in fccd8e58cda97c2a11a1ad4c113ada3a21318870 as an attempt to fix https://github.com/qgis/QGIS/issues/21012, which was later fully closed by 6215f773e22cb55dac694ae6b213cd1f9ac39c53 . So possibly they are redundant in any case given the follow up commit.

That said, removing them has resulted in many failing tests, so there's clearly more at play here...!

roya0045:
Is this issue present on master? I tried to replicate with my partial table loader branch and could not replicate it (neither on my old 3.4.0 which should still have that behavior I assume).

nyalldawson:
@m-kuhn I think this is safe to merge now. The situation is:

1. I can sometimes reproduce https://github.com/qgis/QGIS/issues/32819 , but it's very temperamental and I couldn't write any unit tests which reliably triggered it.

2. With this commit I don't see the crash... but I'm not 100% sure it's fixed, because even without the commit I'd only see the crash sometimes.

3. There **is** an issue with this code, because we're making calls to beginRemoveRows while a beginResetModel is still active. I can't find any documentation describing this as a bad thing, but my suspicion (and the stack traces I get from the crash lead me to believe) that it is...

4. We don't need to reset the model here -- the rest of the function is (now) correctly signalling the atomic row based changes. And that's a better approach generally vs resetting the model anyway...

What do you think?

m-kuhn:
It's a public method, if it's called from outside, it won't be wrapped in begin/end modification calls?

Would it be more appropriate to add this as an internal method and have the public one trigger the internal one within reset commands?

nyalldawson:
> It's a public method, if it's called from outside, it won't be wrapped in begin/end modification calls?

No, they happen internally within the method

m-kuhn (approved):
All good then :+1: 
Thanks

PR #32847, opened 2019-11-14
Fix rendering artifacts on the edges of resampled raster tiles

### LINKED ISSUE #30152: PDF Export Errors with Raster Data
I'm using QGIS 3.6.2 and this issue occurs when working with a low-resolution raster dataset in TIF format.

When refreshing a layout, there appears to be more than one draw. In the initial draw, everything is almost perfect, but there a few anomalies toward the middle of the map frame. After another second there's an adjustment and the map is as it should be.

When exporting to PDF or choosing the option to print to PDF it appears that the export occurs before the final draw as the anomaly from part way through the refresh process is present in the export. The same thing occurs when exporting to an image.

After much struggle, the cause of the error was discovered to be having a raster dataset in New York State Plane and the rest of the data and the map frame in  WGS 1984 (4326). 

The aesthetic of the map looks much better when the raster data and the map frame are not in the same projection, so I hope this can be fixed.

nyalldawson:
We now buffer the request for the input to a resample filter by
a variable number of pixels (depending on the resampling type)
in order to fetch the neighbouring pixels to the edges of individual
raster blocks. This allows the resampling to utilise these
neighbouring pixels when resampling the edges of tiles, avoiding
discrepancies and rendering artifacts over the borders of
raster tiles.

Fixes #30152


nyalldawson:
@nirvn oversampling addressed now

nirvn:
Nice. As far as I'm concerned, this is ready to merge

PR #32857, opened 2019-11-14
Use service=qgis_test for referencing test database

strk:
Makes PostgreSQL testing setup consistent.
Updates README accordingly


strk:
I'm not sure auth tests can use a single `service` configuration, as I suspect we'd be using different usernames, so we'd need different services

m-kuhn (approved):
Looks good

PR #32877, opened 2019-11-14
Warn when we know for certain that no spatial index exists

nyalldawson:
...on a source used for the Join by Location algorithm

Advise users that performance will be severely degraded as a result. Results from a question asked at FOSS4G Oceania about why a join by location execution was taking 4-5 hours to run. (Correct answer: with a spatial index present, it takes only 3-4 minutes :laughing: )

Also improve UI around creating spatial indexes in the vector layer properties. 

Note that only the memory provider and delimited provider support determining spatial index presence -- GDAL has no api to test for this presence unfortunately.

### DIFF
--- a/python/core/auto_generated/processing/qgsprocessingutils.sip.in
+++ b/python/core/auto_generated/processing/qgsprocessingutils.sip.in
@@ -412,6 +412,8 @@ iterator, eg by restricting the returned attributes or geometry.
 
     virtual QgsFeatureIds allFeatureIds() const;
 
+    virtual SpatialIndexPresence hasSpatialIndex() const;
+
 
     QgsExpressionContextScope *createExpressionContextScope() const /Factory/;
 %Docstring
--- a/python/core/auto_generated/qgsfeaturesource.sip.in
+++ b/python/core/auto_generated/qgsfeaturesource.sip.in
@@ -159,6 +159,24 @@ for its ownership.
 
 .. versionadded:: 3.0
 %End
+
+    enum SpatialIndexPresence
+    {
+      SpatialIndexUnknown,
+      SpatialIndexNotPresent,
+      SpatialIndexPresent,
+    };
+
+    virtual SpatialIndexPresence hasSpatialIndex() const;
+%Docstring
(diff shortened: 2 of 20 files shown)


nirvn:
@nyalldawson , could we not come up with some basic support for the OGR provider, by coding in checks for the most wildly used formats (i.e. for shapefile, check the present of an .shx, for geopackage, check for the relevant table in the sqlite db, etc.)?

nyalldawson:
We potentially could! (It should be possible to add for postgres provider too at some stage, ideally first by adding to the db connection API).

Before we start doing this in qgis for ogr, I'd rather explore if it could be added directly in for itself first (where it rightly belongs)

nyalldawson:
I also think we could take this further in processing, eg creating the index in memory when we know for certain it doesn't exist.

nirvn:
@nyalldawson , definitively big +1 to add to the ogr/gdal library itself, but that's not likely to be backported to gdal 2.4.X, which remains the default gdal version for most linux distribution that will carry on our 3.10 LTR for the foreseeable future.

roya0045:
> @nyalldawson , definitively big +1 to add to the ogr/gdal library itself, but that's not likely to be backported to gdal 2.4.X, which remains the default gdal version for most linux distribution that will carry on our 3.10 LTR for the foreseeable future.

Do you know which version of QGIS should be based on GDAL3 and PROj6? I thought that doing the next release with those and dropping python 2 would be easy but I may be mistaken.

nyalldawson:
> GDAL3 and PROj6

Honestly, it won't be for a loooong time. Gdal 3 hasn't even hit many of the fast moving Linux distros yet. Maybe qgis 3.18 or something...

nyalldawson:
@nirvn can we merge as is? I've no plans to implement for the other providers, so it's a "you get what you get" type thing 😉

nirvn:
Sure, implementation looks good.

nirvn:
@nyalldawson , long hanging fruit: adding the same warning to select by location algorithm.

PR #32890, opened 2019-11-15
[feature] Processing raster calc: add missing btns and validate

elpaso:
Add missing buttons for ABS/MIN/MAX and validate expression.

processing-raster-calc-validator

### DIFF
--- a/python/plugins/processing/algs/qgis/ui/RasterCalculatorWidget.ui
+++ b/python/plugins/processing/algs/qgis/ui/RasterCalculatorWidget.ui
@@ -34,38 +34,17 @@
        <string>Operators</string>
       </property>
       <layout class="QGridLayout" name="gridLayout">
-       <item row="1" column="2">
-        <widget class="QPushButton" name="mACosButton">
+       <item row="2" column="4">
+        <widget class="QPushButton" name="mOpenBracketPushButton">
          <property name="text">
-          <string>acos</string>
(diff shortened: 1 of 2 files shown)


roya0045:
Do the 3 functions comes with parenthesis by default or does the user need to specify? Also auto englobing (select "str", hit 'Max', "str" -> Max("str") ), I'm not sure what it's called.

elpaso:
@roya0045 there is nothing really new here, these buttons are already available since 3.10 in the main menu -> Raster -> Raster Calculator C++ implementation, I just ported them to the Python side and added the validation (which is also already available in the C++ raster calculator dialog).

I f you are looking for a pet project, refactoring a bit the C++ implementation in a way that we don't have to duplicate things in the Python side could be a nice one :)

roya0045:
> @roya0045 there is nothing really new here, these buttons are already available since 3.10 in the main menu -> Raster -> Raster Calculator C++ implementation, I just ported them to the Python side and added the validation (which is also already available in the C++ raster calculator dialog).

Yeah I saw the code afterward.

> I f you are looking for a pet project, refactoring a bit the C++ implementation in a way that we don't have to duplicate things in the Python side could be a nice one :)

I don't think this endeavor would benefit my employer/myself, my PRs do but the delays are unpleasant.

nyalldawson (approved):

PR #32891, opened 2019-11-15
[cmake] find SSH lib

3nids:
When building on mac, it fails since it cannot find libssh.h which is included in exiv2 (depending if exiv2 has been built with SSH or not).
In the past, I had fixed this by manually setting exiv2 to be compiled withtou SSH but this is not the default case on Mac/Homebrew.
This PR allows defining `SSH_INCLUDE_DIR`, fixing the issue.


jef-n:
> The SSH library might be required by EXIV2

Shouldn't `cmake/FindEXIV2.cmake` just include it in `EXIV2_INCLUDE_DIR`(s) if required?

3nids:
Is it supposed to find it automatically? Or do you mean move the content of findssh.cmake in findexiv.cmake

jef-n:
> Is it supposed to find it automatically? Or do you mean move the content of findssh.cmake in findexiv.cmake

Yes, not sure if it's worth it, but even verify whether it's required or not.

3nids:
using `/usr/local/include` instead of `/usr/local/opt/exiv2/include` as include dir for exiv2 seems to fix the issue for me on mac.
sorry for the noise.

PR #32900, opened 2019-11-16
[feature] is_valid expression

pkinglinz:
## Description
Add an is_valid function to the expressions, which reuses GEOS is valid and returns true if a geometry is valid; if it is well-formed in 2D according to the OGC rules.

Based of @NathanW2  workshop at FOSS4g Oceania.
I have tested the function manually and confirmed the tests pass, however I can't get my build to produce the help files to check if this or correct. I"m not sure if "Uses GEOS" is needed in the help file.

I'm a python dev not C++ so any help will definitely be appreciated
Thanks

## Checklist


NathanW2:
nice work @pkinglinz.  Thanks for adding this.  

Make sure you run `scripts/prepare-commit.sh` before committing in order up the code style.  The tests all pass except for the code layout one.

If you just make the `qgsexpressionfunction.cpp` have a change so git picks it up as modified and run  prepare-commmit.sh it should fix it up.

pkinglinz:
I wasn't able to get prepare-commit.sh to run, it was missing autopep8 and flip. The astyle flag is checked, I added autopep8 but I'm unsure how to add flip. I will continue to try to get prepare-commit working but in the mean time I mad the changes to the style. Thanks again for your help

nyalldawson:
Flip should be available from the repos - "apt-get install flip"

resources/function_help/json/is_valid
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "is_valid",
+  "type": "function",
+  "description": "Returns true if a geometry is valid. Uses GEOS.",
```
nyalldawson: I'd drop the "uses geos" part -- it won't mean anything for most users. Maybe you could work in postgis' description of this function: "Tests if a geometry value is well-formed in 2D according to the OGC rules for geometry validity."

tests/src/core/testqgsexpression.cpp
```diff
@@ -971,6 +971,11 @@ class TestQgsExpression: public QObject
       QTest::newRow( "pole_of_inaccessibility not poly" ) << "geom_to_wkt(pole_of_inaccessibility( geomFromWKT('POINT (1.5 0.5)'), 0.1 ))" << false << QVariant();
       QTest::newRow( "pole_of_inaccessibility not geom" ) << "pole_of_inaccessibility('g',0.1)" << true << QVariant();
       QTest::newRow( "pole_of_inaccessibility null" ) << "pole_of_inaccessibility(NULL,0.1)" << false << QVariant();
+      QTest::newRow( "is_valid not geom" ) << "is_valid('g')" << true << QVariant();
```
DelazJ: This means that it's not only about geometry, doesn't it?

DelazJ:
Thanks @pkinglinz Also, for changelog and doc, can you add [feature] to the PR's title or a commit?

tests/src/core/testqgsexpression.cpp
nyalldawson: No, the test is checking that you can ONLY call the function with a geometry input, and other things eg strings cause errors to be thrown

NathanW2:
@pkinglinz make sure you use `[feature]` in the title as it will get picked up by the changelog and auto docs like @DelazJ said.

nyalldawson:
@nathanw2 let's just add that in a squash merge...

pkinglinz:
@NathanW2 @nyalldawson @DelazJ thanks for all the help. I have imported flip and run the prepare-commit.sh script, I know it is run by travis just wanted to check i could do it locally. Updated the description in the help docs, its pretty much exactly what was suggested.

nyalldawson (approved):

NathanW2:
nice work @pkinglinz

pkinglinz:
thanks @NathanW2 can you think of any other simple,expressions I could add to get some practice in?

NathanW2:
Not off the top of my head but if you have a look at the PostGIS functions
and see if there is anything that is cool that we are missing is a good
start normally.

nyalldawson:
@pkinglinz I'd say a "make_valid" function would be a great companion, and very similar to implement.

NathanW2:
Oh yes! That one!

PR #32905, opened 2019-11-17
Weight field should be numeric in count points in polygon

DelazJ:
alg if we want to sum them up
Using a string field the returned count is 0 while no field or numeric field returned count.


nyalldawson:
Are you sure about this? Looking at the code a string field with value '4' would work ok...

DelazJ:
Ah Indeed. I was using a field with no numerical content...

DelazJ:
My issue was that any field in the layer could be picked, regardless the content. So to test the docs, I used a field with full text and the result was 0.

nyalldawson:
> So to test the docs, I used a field with full text and the result was 0.

So in that case we should try to convert this weight value to a number, and failing that raise an exception. Would you agree?

DelazJ:
I do. Any pointers, unless it takes you less time to update the PR than finding an example I could pick? Thanks.

nyalldawson:
This is redundant now -- it's handled in the c++ port

PR #32906, opened 2019-11-17
Only load plotly library on demand

nyalldawson:
(i.e. at time of algorithm execution)

Because:
1. It's nicer to show all algorithms on all installs, and give a descriptive
error message to users when they try to run algorithms which depend on Plotly
if the library is missing. Otherwise on some installs these algorithms are
just missing for no apparent reason.

2. The plotly library takes a long time to load on windows (3-4 seconds),
so by moving this load to an on-demand load at time of algorithm execution
we can shave a few seconds off the QGIS startup time for ALL users.

(As discovered by @NathanW2 )


alexbruy (approved):
Looks good.

nyalldawson:
Thanks @alexbruy!

PR #32917, opened 2019-11-18
Fix evalution of letter/word spacing when font size is map units

### LINKED ISSUE #32825: Label letter and word spacing broken in QGIS 3.10
Looks like introduction of kerning in 3.10 has broken letter and word spacing in labels.

In QGIS 3.10 if letter rspacing is set to something other than 0, then letter's interval shrinks or expands depending on zoom level.
Label size is set to map units.
Map projection is EPSG;3857
Layer projection id EPSG:3857

The attached gif-video shows difference in v3.4 and v3.10
https://yadi.sk/d/A4GRvsg6AJgG8A

Also notice, that in 3.4 the label is shown even if it is partially visible, but in 3.10 it disappears.
"Allow truncated labels" option is enabled in both projects

nyalldawson:
Fixes #32825


nirvn (approved):

nirvn:
Looking good.

Issue #29269, opened 2019-03-03
r.series broken in Processing due to wrong "range=0,0" pre-definition


qgib:
Author Name: **Markus Neteler** (Markus Neteler)
Original Redmine Issue: [21452](https://issues.qgis.org/issues/21452)
Affected QGIS version: 3.7(master)
Redmine category:processing/grass

---
In QGIS 3.4, 3.6 r.series is broken in Processing due to wrong "range=0,0" pre-definition. It should be empty but is pre-populated with 0,0.

That leads to an *empty result* (basically a NO DATA raster map):

[...]
g.proj -c proj4="+proj=lcc +lat_1=36.16666666666666 +lat_2=34.33333333333334 +lat_0=33.75 +lon_0=-79 +x_0=609601.22 +y_0=0 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs"
r.external input="/home/mneteler/lsat7_2000.tif" band=1 output="rast_5c7bdac0189302" --overwrite -o
g.region n=228513.0 s=214975.5 e=645012.0 w=629992.5 res=28.5
r.series input=rast_5c7bdac0189302 method="average" range="0,0" output=output28566faed0e94fdaa03a9cfb57fb94a9 --overwrite

--> here range="0,0" must not be there unless the user enters values in the Processing form under "Advanced".

*How to reproduce:*

- test dataset (multilayer Landsat-7): lsat7_2000.tif attached (2MB)
- open dataset
- open r.series dialog from Processing
- select lsat7_2000.tif
- click "run" (default is "average" method which is fine to demonstrate the problem)

I tried to fix that but probably QgsProcessingParameterRange() is somehow broken or wrongly used in Processing?

It is likely that the suboptimal use of QgsProcessingParameterRange() affects also other GRASS GIS modules present in Processing.

---

- [lsat7_2000.tif](https://issues.qgis.org/attachments/download/14506/lsat7_2000.tif) (Markus Neteler)

---

Related issue(s): #29374 (relates)
Redmine related issue(s): [21558](https://issues.qgis.org/issues/21558)

---

PR #32922, opened 2019-11-18
[processing] add "not set" support for range parameter (fix #29374)

### LINKED ISSUE #29374: Allow none/empty as possible values for Processing QgsProcessingParameterRange
Original Redmine Issue: [21558](https://issues.qgis.org/issues/21558)

Redmine category:processing/grass

---
Now in QgsProcessingParameterRange "0" is always used as default.

This has two downsides:

1) 0 could not be an appropriate default and can led to errors/confusions
2) sometimes the parameter defined by QgsProcessingParameterRange is not mandatory, not being able to set it to "none" it is always used, even if is not necessary.

See for example #29269

---

Related issue(s): #29269 (relates)
Redmine related issue(s): [21452](https://issues.qgis.org/issues/21452)

---

alexbruy:
## Description
Range parameter behaves slightly differently from other parameters: even if it is optional, it returns a value `0,0` instead of `None`. This leads to issues with algorithms where range parameter should be optional, for example #29269.

This is an attempt to fix behavior of range parameter and its wrapper. Passing `None` value as default for optional parameter will set its wrapper to "not set" state and an empty list will be returned as parameter value. Fixes #29374.

Another option will be adding handlers for range parameter in affected providers.

## Checklist

### DIFF
--- a/src/core/processing/qgsprocessingparameters.cpp
+++ b/src/core/processing/qgsprocessingparameters.cpp
@@ -1447,6 +1447,7 @@ QList<double> QgsProcessingParameters::parameterAsRange( const QgsProcessingPara
 
   QStringList resultStringList;
   QVariant val = value;
+
   if ( val.canConvert<QgsProperty>() )
     resultStringList << val.value< QgsProperty >().valueAsString( context.expressionContext(), definition->defaultValue().toString() );
   else if ( val.type() == QVariant::List )
@@ -1478,9 +1479,23 @@ QList<double> QgsProcessingParameters::parameterAsRange( const QgsProcessingPara
   }
 
   if ( resultStringList.size() < 2 )
-    return QList< double >() << 0.0 << 0.0;
+    return QList< double >() << NAN << NAN;
 
-  return QList< double >() << resultStringList.at( 0 ).toDouble() << resultStringList.at( 1 ).toDouble();
+  QList< double > result;
+  bool ok = false;
+  double n = resultStringList.at( 0 ).toDouble( &ok );
+  if ( ok )
+    result << n;
+  else
+    result << NAN;
+  ok = false;
+  n = resultStringList.at( 1 ).toDouble( &ok );
+  if ( ok )
+    result << n;
+  else
+    result << NAN;
+
+  return result;
 }
 
 QStringList QgsProcessingParameters::parameterAsFields( const QgsProcessingParameterDefinition *definition, const QVariantMap &parameters, QgsProcessingContext &context )
@@ -3127,7 +3142,8 @@ bool QgsProcessingParameterRange::fromVariantMap( const QVariantMap &map )
 
 QgsProcessingParameterRange *QgsProcessingParameterRange::fromScriptCode( const QString &name, const QString &description, bool isOptional, const QString &definition )
 {
-  return new QgsProcessingParameterRange( name, description, QgsProcessingParameterNumber::Double, definition.isEmpty() ? QVariant() : definition, isOptional );
+  return new QgsProcessingParameterRange( name, description, QgsProcessingParameterNumber::Double, definition.isEmpty() ? QVariant()
+                                          : ( definition.toLower().trimmed() == QStringLiteral( "none" ) ? QVariant() : definition ), isOptional );
 }
 
 QgsProcessingParameterRasterLayer::QgsProcessingParameterRasterLayer( const QString &name, const QString &description, const QVariant &defaultValue, bool optional )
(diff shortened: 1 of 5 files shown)


nyalldawson:
Looking good! Something I wonder is whether we need to expose this as a separate option for the lower and upper values, e.g. allow a not set upper limit while requiring a numeric lower limit value. I'/m 90% sure we'll hit this requirement sometime in the future...

alexbruy:
I'm not against such improvement, although right now I don't see any use cases for it.

Should we return min and max values from `std::numeric_limits` when corresponding value is not set or better to change API a bit and return `QVariantList` with empty item(s) instead of `QList<double>` as parameter value?

nyalldawson:
> Should we return min and max values from std::numeric_limits when corresponding value is not set

I'd say a NaN value would be ideal for this

(and sorry for the delay in replying -- for some reason I forgot about this PR)

alexbruy:
Thinking about it a bit more I don't see why we may need an open-ended range parameter, as it can be easily replaced by single number. Or maybe I miss some points here?

As range parameter represented by two spinboxes what will mean "Not set" for any of them: user can not change that value or just spinbox will have no min/max limits (currently they are limited to the--99999999.999999 - 99999999.999999 interval).

nyalldawson:
> As range parameter represented by two spinboxes what will mean "Not set" for any of them: user can not change that value or just spinbox will have no min/max limits (currently they are limited to the--99999999.999999 - 99999999.999999 interval).

I was thinking something more along the lines of how the numeric parameter works when it's set to optional. In that case, you get the "clear" button shown in the spin box which you can use to set its state to the "not set" option. But for ranges, this would be possible for either the min or the max value individually.

alexbruy:
> I was thinking something more along the lines of how the numeric parameter works when it's set to optional. In that case, you get the "clear" button shown in the spin box which you can use to set its state to the "not set" option. But for ranges, this would be possible for either the min or the max value individually.

This is how it works now. Either min, max or both values can be cleared with the button and their state will be set to "not set" option. If at least one value has state "not set" then range considired as not set too.

alexbruy:
When range parameter is optional and no default value it set, widget will look like
optional-nodefaults
User can adjust lower and upper values or again reset them to "not set" using clear button
optional-defaults
If at least one boundary is "not set" then parameter considered as "not set" and should not be used.

nyalldawson:
> If at least one boundary is "not set" then parameter considered as "not set" and should not be used.

I guess that's the part which I'm unsure about. Couldn't we set the value for this situation as `[5, None]`/ `[None, 5]` and leave it up to the algorithm to decide if that means the whole parameter is "not set", or just that that particular bound is not set?

alexbruy:
@nyalldawson I tried to implement it.

neteler:
I hope the work done here will not be lost...

alexbruy:
Not stale, awaits review.

alexbruy:
@nyalldawson sorry for asking to look at this again, but I'd like to hear your opinion on this is

src/core/processing/qgsprocessingparameters.cpp
```diff
@@ -1534,9 +1535,23 @@ QList<double> QgsProcessingParameters::parameterAsRange( const QgsProcessingPara
   }
 
   if ( resultStringList.size() < 2 )
-    return QList< double >() << 0.0 << 0.0;
+    return QList< double >() << NAN << NAN;
```
nyalldawson:
```suggestion
    return QList< double >() << std::numeric_limits<double>::quiet_NaN() << std::numeric_limits<double>::quiet_NaN();
```

src/core/processing/qgsprocessingparameters.cpp
```diff
@@ -1534,9 +1535,23 @@ QList<double> QgsProcessingParameters::parameterAsRange( const QgsProcessingPara
...
 
-  return QList< double >() << resultStringList.at( 0 ).toDouble() << resultStringList.at( 1 ).toDouble();
+  QList< double > result;
+  bool ok = false;
+  double n = resultStringList.at( 0 ).toDouble( &ok );
+  if ( ok )
+    result << n;
+  else
+    result << NAN;
```
nyalldawson:
```suggestion
    result << std::numeric_limits<double>::quiet_NaN();
```

src/core/processing/qgsprocessingparameters.cpp
```diff
@@ -1534,9 +1535,23 @@ QList<double> QgsProcessingParameters::parameterAsRange( const QgsProcessingPara
...
+  ok = false;
+  n = resultStringList.at( 1 ).toDouble( &ok );
+  if ( ok )
+    result << n;
+  else
+    result << NAN;
```
nyalldawson:
```suggestion
    result << std::numeric_limits<double>::quiet_NaN();
```

nyalldawson (approved):
Looks good to me!

nyalldawson:
(and sorry about the slow review :( )

neteler:
Thanks for your hard work!

A backport to "stable" would be much appreciated, how to initiate that?

PR #32937, opened 2019-11-18
Add FlatGeobuf as an exportable format

bjornharrtell:
Initial attempt to add FlatGeobuf as an exportable vector format. FlatGeobuf is already readable in QGIS when compiled against GDAL master.

Short story on FlatGeobuf is that it tries to fill a potential void between Shapefile and GPKG by being simple as a Shapefile, better performing and suitable for streaming. Long story can be found at https://github.com/bjornharrtell/flatgeobuf. FlatGeobuf is in GDAL master and expected to be in GDAL 3.1 as per https://gdal.org/drivers/vector/flatgeobuf.html.

## Checklist

- [x] I have read the [QGIS Coding Standards]

### DIFF
--- a/src/core/qgsvectorfilewriter.cpp
+++ b/src/core/qgsvectorfilewriter.cpp
@@ -3187,10 +3187,6 @@ QList< QgsVectorFileWriter::FilterFormatDetails > QgsVectorFileWriter::supported
         return true; // Make https://twitter.com/shapefiIe a sad little fellow
       else if ( b.driverName == QLatin1String( "GPKG" ) )
         return false;
-      else if ( a.driverName == QLatin1String( "FlatGeobuf" ) )
-        return true;
-      else if ( b.driverName == QLatin1String( "FlatGeobuf" ) )
-        return false;
       else if ( a.driverName == QLatin1String( "ESRI Shapefile" ) )
         return true;
       else if ( b.driverName == QLatin1String( "ESRI Shapefile" ) )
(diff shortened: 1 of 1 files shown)


NathanW2:
Nifty!

src/core/qgsvectorfilewriter.cpp
```diff
@@ -3172,6 +3187,10 @@ QList< QgsVectorFileWriter::FilterFormatDetails > QgsVectorFileWriter::supported
         return true; // Make https://twitter.com/shapefiIe a sad little fellow
       else if ( b.driverName == QLatin1String( "GPKG" ) )
         return false;
+      else if ( a.driverName == QLatin1String( "FlatGeobuf" ) )
```
nyalldawson: Let's drop these sorting tweaks for now and revisit when we have a better idea of the format's adoption -- I think it's premature to push users to this format at this stage.

nyalldawson:
I've been looking forward to this one for a long time -- thanks for the PR!

src/core/qgsvectorfilewriter.cpp
bjornharrtell: Agreed :)

bjornharrtell: And done.

src/core/qgsvectorfilewriter.cpp
```diff
@@ -962,6 +962,21 @@ class QgsVectorFileWriterMetadataContainer
                              )
                            );
 
+      // FlatGeobuf
```
nyalldawson: I missed it earlier, but this should be ifdefed in only for GDAL versions with support for the format...

bjornharrtell: Alright, fixed. I see now that also the reading side (ogrprovider) should get some attention but I'll do that in a separate PR.

bjornharrtell:
I've now also run `./scripts/prepare-commit.sh` and was lucky enough to get `nothing was modified` as the result. :)

nyalldawson (approved):

PR #32939, opened 2019-11-19
[FEATURE][layouts] Add search box to layout manager

nyalldawson:
Allows filtering the list of layouts in a project by name, handy when a project has many layouts..!


nirvn:
Wondering whether it'd be worth having the layout manager be upgraded to a QgsDockPanel.

nyalldawson:
> Wondering whether it'd be worth having the layout manager be upgraded to a QgsDockPanel.

I'm -0 on that. I think it's quite handy having some dialogs like this one and the style manager as real top-level windows, and letting the operating window manager handle them. If we made this one a floating dock widget then you'd only ever be able to see it when the main application window is focused, and not when a layout designer is focused...

nirvn:
Far enough.

nirvn (approved):

PR #32952, opened 2019-11-19
Attribute table: please don't delete randomly!

### LINKED ISSUE #32933: In form view wrong features/records are deleted
Latest QGIS 3.4 on Linux

An image is worth a 1000 words:

Peek 2019-11-18 16-08

elpaso:
Fixes #32933

Pretty big fat bug isn't it ?

Funded by **Spare Time (is) LTD**

### DIFF
--- a/src/gui/attributetable/qgsfeaturelistview.cpp
+++ b/src/gui/attributetable/qgsfeaturelistview.cpp
@@ -325,11 +325,7 @@ void QgsFeatureListView::contextMenuEvent( QContextMenuEvent *event )
 
     QgsActionMenu *menu = new QgsActionMenu( mModel->layerCache()->layer(), feature, QStringLiteral( "Feature" ), this );
 
-    // Index is from feature list model, but we need an index from the
-    // filter model to be passed to listeners, using fid instead would
-    // have been much better in term of bugs (and headaches) but this
-    // belongs to the API unfortunately.
-    emit willShowContextMenu( menu, mModel->mapToSource( index ) );
+    emit willShowContextMenu( menu, feature.id() );
 
     menu->exec( event->globalPos() );
   }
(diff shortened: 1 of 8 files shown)


gioman:
> Funded by **Spare Time (is) LTD**

@elpaso I'm sincerely thankful for all the effort, thanks.

> Pretty big fat bug isn't it ?

yes, horrifying.

src/gui/attributetable/qgsfeaturelistview.cpp
```diff
@@ -325,7 +325,11 @@ void QgsFeatureListView::contextMenuEvent( QContextMenuEvent *event )
 
     QgsActionMenu *menu = new QgsActionMenu( mModel->layerCache()->layer(), feature, QStringLiteral( "Feature" ), this );
 
-    emit willShowContextMenu( menu, index );
+    // Index is from feature list model, but we need an index from the
+    // filter model to be passed to listeners, using fid instead would
+    // have been much better in term of bugs (and headaches) but this
+    // belongs to the API unfortunately.
+    emit willShowContextMenu( menu, mModel->mapToSource( index ) );
```
nyalldawson: This is not nice API :P I think you should remove willShowContextMenu and replace it with your suggestion of using the fid. The existing API is a bug, and should be fixed...

elpaso: I investigated that a bit deeper and there are some actions that require the full selected area (so the full index), I'm afraid the single fid would not be sufficient. 
I'll amend the comment.

elpaso: mmm correcting myself, apparently they get only get item index under the cursor when the context menu is opened, I guess we can retrieve the full selected extent in case we need it anyway.

nyalldawson (approved):

PR #32956, opened 2019-11-19
Update MXE build recipes

elpaso:
This was required by upstream updates to newer QT and CMake

### DIFF
--- a/ms-windows/mxe/build-mxe.sh
+++ b/ms-windows/mxe/build-mxe.sh
@@ -47,6 +47,16 @@ RELEASE_DIR=$(pwd)/qgis-mxe-release
 # TARGET=${TARGET}
 TARGET=i686-w64-mingw32.shared.posix
 
+# Set base path for all tools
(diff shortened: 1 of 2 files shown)


ms-windows/mxe/build-mxe.sh
```diff
@@ -24,6 +24,8 @@ set -e
 #        The artifact will be saved as a zip package in the directory
 #        from which this script is launched.
 
+# shellcheck disable=SC2155
```
nyalldawson: Why not fix this instead? https://github.com/koalaman/shellcheck/wiki/SC2155

elpaso: Because it's nonsense, pwd in that case will never return false

elpaso: But I guess I have no choice than wasting my time :(

jef-n: $PWD?

elpaso:
Superceeded by #33064

PR #32965, opened 2019-11-20
Fix broken QMap<QVariantList,...> finding

nyalldawson:
...which causes case-insensitive comparisons to be made when resolving primary keys in the Oracle and Postgres providers

qt's built in qMapLessThanKey for QVariantList is broken and does a case-insensitive operation, so we replace it with a working version instead! Ouch!


nyalldawson:
I'd love a review of this so I can merge and backport before tomorrow's 3.10.1/3.4 releases....

m-kuhn (approved):
It would be great to bring this all upstream for Qt 6, but no idea how.

PR #32967, opened 2019-11-20
Disable flaky svg cache test

m-kuhn:
This test is randomly failing.
See e.g. https://github.com/qgis/QGIS/pull/31930#issuecomment-548837574

CC @rldhont

### DIFF
--- a/tests/src/python/test_qgssvgcache.py
+++ b/tests/src/python/test_qgssvgcache.py
@@ -69,6 +69,7 @@ class TestQgsSvgCache(unittest.TestCase):
         while not self.fetched:
             QCoreApplication.processEvents()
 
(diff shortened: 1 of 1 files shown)


rldhont:
@m-kuhn `testRemoteSVG` can also failed on travis

m-kuhn:
Thanks for the feedback @rldhont , updated.
Does anyone know if there are plans to look into why these tests are failing sometimes?

rldhont:
The `testRemoteSVG` sometimes fails on travis because the content is retrieved immediately even if the the HTTP request handler waits 1 second before providing the file.

The `testRemoteSvgAsText` sometimes failed on travis because after waiting for content the SVG with text mime-type is not well rendered.

m-kuhn:
Is it possible to fix that?

nyalldawson (approved):

PR #32971, opened 2019-11-20
Add FlatGeobuf as known input format

bjornharrtell:
## Description

Add FlatGeobuf as known input format from GDAL 3.1 and forward.

## Checklist

- [x] I have read the [QGIS Coding Standards]
- [x] I have run [the `scripts/prepare-commit.sh` script]


bjornharrtell:
Rationale for also adding to sProtocolDrivers is because one of the use cases for FlatGeobuf is as a streaming serialization format fx. as a custom WFS output format (which already exists as an experimental module for GeoServer).

PR #32983, opened 2019-11-20
Massive 3.10 backport

nyalldawson:
Contains a lot of backports which didn't work because of the broken backport bot...


src/core/providers/ogr/qgsogrprovider.cpp
```diff
@@ -4454,14 +4462,23 @@ void QgsOgrProvider::open( OpenMode mode )
       QgsDebugMsg( QStringLiteral( "OGR failed to opened in update mode, trying in read-only mode" ) );
     }
 
+    QStringList options;
+#if GDAL_VERSION_NUM >= GDAL_COMPUTE_VERSION(3,1,0)
+    // assume trusted data to get more speed
+    if ( mGDALDriverName == QLatin1String( "FlatGeobuf" ) )
+    {
+      options << "VERIFY_BUFFERS=NO"
```
jef-n: `;` missing

rouault: Do we need the #ifdef ?. If the driver is not present at runtime, then this branch will not be run, and that's about all :-)
The less #ifdef, the better

Issue #32968, opened 2019-11-20
QGIS 3.10 Plugin install fails because of (hyphen in) folder name


Jannik-Schilling:
I found this problem with the plugins "ESTIMAP Recreation" and "WaterNetAnalyzer". 

Both are available in the plugin repository. 
Both can be installed without any problem from a zip file.
Both work without any problems in QGIS 3.2, 3.4 ... 3.8. But installing these plugins **from the official plugin repository** fails in QGIS 3.10
When installed from the plugin repository you get an error: 
> Plugin has disappeared: The plugin seems to have been installed but I don't know where. Probably the plugin package contained a wrong named directory. Please search the list of installed plugins. I'm nearly sure you'll find the plugin there, but I just can't determine which of them it is. It also means that I won't be able to determine if this plugin is installed and inform you about available updates. However the plugin may work. Please contact the plugin author and submit this issue.

In the folder "...\QGIS3\profiles\default\python\plugins" they are correctly installed with the folder name "[pluginname]-master" e.g "waternetanalyzer-master"

It seems as the hyphen is a problem, because other plugins, which have "_" in the folder name, work fine.

I tried to fix this by uploading a new version to the official qgis plugin repository without the hyphen. I get: "Plugin name mismatch: the plugin main folder name in the compressed file (WaterNetAnalyzer_master) is different from the original plugin package name (WaterNetAnalyzer-master)

elpaso:
Please report the issue to the plugin author.

Jannik-Schilling:
> 
> 
> Please report the issue to the plugin author.

I am the author.

elpaso:
This is the problem with your plugin: "Plugin name mismatch: the plugin main folder name in the compressed file (WaterNetAnalyzer_master) is different from the original plugin package name (WaterNetAnalyzer-master)"

The plugin package name cannot be changed after a plugin has been created.

Jannik-Schilling:
> The plugin package name cannot be changed after a plugin has been created.
yes, I understand that. This was just my first attempt to fix the problem. 

But is there any reason, why a the version with the hyphen "WaterNetAnalyzer-master" , which is already located in the qgis plugin repostory cannot be installed?

borysiasty:
It's because from the commit 5c1cc4a40d84, the hyphen is interpreted as a delimiter of a suffix in the zip file name. So WaterNetAnalyzer-master.zip is expected to contain a "WaterNetAnalyzer" directory.

[...]

PR #32991, opened 2019-11-21
Backport #32978 to 3.10: Revert 5c1cc4a40d84 in order to allow dashes in plugin directory names

borysiasty:
A manual backport from master, as the automagic doesn't work last days:

---------------
A few weeks ago I pushed the 5c1cc4a40d84 commit, so QGIS started interpreting dashes in zip file names as a version suffix delimiter (e.g. MyPlugin-0.0.1.zip should contain "MyPlugin" directory). The purpose was to end a frequent confusion: plugin authors was naming zip files like that and they were reporting those zips are not installable.

However, now it turned out (see #32968) that dashes are also used in the directory names for some time (40 plugins for QGIS 3.x vs zero for 2.x), they are even officially allowed: https://docs.qgis.org/3.4/en/docs/pyqgis_developer_cookbook/plugins/releasing.html#validation . And yes, QGIS is able to import packages with dashes in names.

In this situation I guess we have to revert that commit (although personally I'm obviously against playing with fire and allowing dashes in Python packages).

## Checklist

### DIFF
--- a/python/pyplugin_installer/installer_data.py
+++ b/python/pyplugin_installer/installer_data.py
@@ -397,11 +397,7 @@ class Repositories(QObject):
                     fileName = pluginNodes.item(i).firstChildElement("file_name").text().strip()
                     if not fileName:
                         fileName = QFileInfo(pluginNodes.item(i).firstChildElement("download_url").text().strip().split("?")[0]).fileName()
-                    match = re.match('(.*?)[.-]', fileName)
-                    if match:
-                        name = match.groups()[0]
-                    else:
-                        name = fileName
+                    name = fileName.partition(".")[0]
                     experimental = False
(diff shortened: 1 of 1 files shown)


nyalldawson:
This is included in https://github.com/qgis/QGIS/pull/32983

borysiasty:
Ah thanks!

elpaso:
@borysiasty for what it worth I agree that would be better to avoid `-` in package names, but I think that if you want to do it "right" you'd better wait for QGIS 4 (you can file an issue for that) and make sure to update the plugins web application and the documentation accordingly, and also probably take care of warning all current authors that they will need to update their packages.

borysiasty:
@elpaso: sure, it's why I reverted to the previous state ASAP, so nothing changes during 3.x. I just had to clearly cut myself off from allowing dashes in the names and from potential troubles they may lead to :)

PR #32992, opened 2019-11-21
Port Processing Rasterize to C++

elpaso:
Port processing rasterize to C++
According to my tests it's ~20% faster, I also added an optional buffer extent to avoid tiles artifacts.


m-kuhn:
Thanks @elpaso!

Gustry:
Nice addition!

You can remove the help string in the YAML file.

PR #32999, opened 2019-11-21
Rasterize: QtConcurrent version

elpaso:
Well, this is becoming interesting: this version is unsurprisingly 8 times faster (on my 8 cores machine) than the C++ port one, what do we do?

There is still room for further optimizations, but I believe that my funder (**Spare Time is LTD**) is already happy.

### DIFF
--- a/src/analysis/processing/qgsalgorithmrasterize.cpp
+++ b/src/analysis/processing/qgsalgorithmrasterize.cpp
@@ -319,6 +319,7 @@ QVariantMap QgsRasterizeAlgorithm::processAlgorithm( const QVariantMap &paramete
 
 bool QgsRasterizeAlgorithm::prepareAlgorithm( const QVariantMap &parameters, QgsProcessingContext &context, QgsProcessingFeedback *feedback )
 {
+  Q_UNUSED( feedback )
   // Retrieve and clone layers
   const QString mapTheme { parameterAsString( parameters, QStringLiteral( "MAP_THEME" ), context ) };
   const QList<QgsMapLayer *> mapLayers { parameterAsLayerList( parameters, QStringLiteral( "LAYERS" ), context ) };
--- a/src/analysis/processing/qgsalgorithmrasterize.cpp
+++ b/src/analysis/processing/qgsalgorithmrasterize.cpp
@@ -145,9 +145,7 @@ QgsRasterizeAlgorithm *QgsRasterizeAlgorithm::createInstance() const
 
 QVariantMap QgsRasterizeAlgorithm::processAlgorithm( const QVariantMap &parameters, QgsProcessingContext &context, QgsProcessingFeedback *feedback )
 {
-
-  const QString mapTheme { parameterAsString( parameters, QStringLiteral( "MAP_THEME" ), context ) };
-  const QList<QgsMapLayer *> mapLayers { parameterAsLayerList( parameters, QStringLiteral( "LAYERS" ), context ) };
+  // Note: MAP_THEME and LAYERS are handled and cloned in prepareAlgorithm
   const QgsRectangle extent { parameterAsExtent( parameters, QStringLiteral( "EXTENT" ), context, context.project()->crs() ) };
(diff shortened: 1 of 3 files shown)


Saijin-Naib:
Does this also impact rendering out data for map books and print layouts, or do those use another rasterization pathway?

elpaso:
This affects only the rasterize alg, other parts are already multithreaded.

src/analysis/processing/qgsalgorithmrasterize.cpp
```diff
@@ -247,59 +244,85 @@ QVariantMap QgsRasterizeAlgorithm::processAlgorithm( const QVariantMap &paramete
...
+      return;
+    }
+    image.fill( transparent ? bgColor.rgba() : bgColor.rgb() );
+    mapSettings.setExtent( QgsRectangle(
+                             extent.xMinimum() + x * extentRatio,
+                             extent.yMaximum() - ( y + 1 ) * extentRatio,
+                             extent.xMinimum() + ( x + 1 ) * extentRatio,
+                             extent.yMaximum() - y * extentRatio
+                           ) );
+    QgsMapRendererCustomPainterJob job( mapSettings, &painter );
```
nyalldawson: I don't think it's safe to create jobs in threads like this - rather they need to be created on the main thread and then started in the other thread.

nyalldawson:
Nice work!!

src/analysis/processing/qgsalgorithmrasterize.cpp
elpaso: Can you please elaborate? I'd like to understand **why** it is not safe.

nyalldawson: Sorry - I'm wrong. What's not thread safe is the map layer handling earlier in this algorithm (and I believe that was also the case with the earlier python version). E.g. project()->mapLayers() should not be called from a background thread, and we also run the risk of any of these layers being deleted before the renderer job starts and has had time to prepare the sources...

elpaso: What do you suggest? Wrap layers with QPointers in mapsettings and check for them in the renderer job?

roya0045:
Out of curiosity, do you think that qtConcurrent could also be applied to the join by location alg?

src/analysis/processing/qgsalgorithmrasterize.cpp
elpaso: Anyway, given the upstream issue, I have the feeling this implementation is gonna be safer because it's faster and keeps the CPU cores busy reducing the likelihood that some other thread kicks in and deletes a layer.

elpaso: Or, just and idea, how about cloning the layers before entering starting the process?

elpaso:
> Out of curiosity, do you think that qtConcurrent could also be applied to the join by location alg?

I just had a quick look, but I would say yes. 
There are a lot of possible optimizations in processing algs, but first we need to make sure the foundations are solid (thread safe).

roya0045:
> > Out of curiosity, do you think that qtConcurrent could also be applied to the join by location alg?
> 
> I just had a quick look, but I would say yes.
> There are a lot of possible optimizations in processing algs, but first we need to make sure the foundations are solid (thread safe).

Excellent news! I had a quick look also but I was not too sure how much reworking was needed to use multiprocessing optimally in that algorithm.

src/analysis/processing/qgsalgorithmrasterize.cpp
nyalldawson: @elpaso that would work if it's done in the prepareAlgorithm step (and sounds like the best solution possible!)

wonder-sk: Agreed with Nyall that starting of map rendering jobs in worker threads is potentially unsafe (all objects needed for rendering that "live" in main thread should be copied in main thread to be safe).

That said, the algorithm was starting map rendering in worker thread already before, so it's not making things actually worse here. (The same issue is in Tiles XYZ algorithm.)

PR #33004, opened 2019-11-21
Fix broken QMap<QVariantList,...> finding (3.10 backport)

nyalldawson:
which causes case-insensitive comparisons to be made when resolving primary keys in the Oracle and
Postgres providers

qt's built in qMapLessThanKey for QVariantList is broken and does a
case-insensitive operation, so we replace it with a working version instead...

(cherry picked from commit 701ea057b1908d74666a8e842445706947fe3e56)


nirvn (approved):

nyalldawson:
Thanks @nirvn!

PR #33010, opened 2019-11-22
[canvas] Fix zoom to selection for single point selection when layer is reprojected

nirvn:
## Description

There was a (long-standing) bug with the canvas zooming to selection code that resulted in a failure to zoom to selection when a single point was selected from a reprojected layer. This PR fixes the situation.

Discovered while fooling around with the latest DataPlotty.

## Checklist

### DIFF
--- a/src/gui/qgsmapcanvas.cpp
+++ b/src/gui/qgsmapcanvas.cpp
@@ -1029,8 +1029,9 @@ void QgsMapCanvas::zoomToSelected( QgsVectorLayer *layer )
   if ( layer->geometryType() == QgsWkbTypes::PointGeometry && rect.isEmpty() )
   {
     int scaleFactor = 5;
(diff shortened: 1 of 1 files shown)


src/gui/qgsmapcanvas.cpp
```diff
@@ -1023,14 +1023,15 @@ void QgsMapCanvas::zoomToSelected( QgsVectorLayer *layer )
   }
 
   rect = mapSettings().layerExtentToOutputExtent( layer, rect );
+  QgsPointXY center = rect.center();
```
m-kuhn: Shouldn't this stay inside the if clause on line 1030?
That looks like an unnecessary calculation in other cases.

m-kuhn:
Good catch!

src/gui/qgsmapcanvas.cpp
nirvn: You are correct.

PR #33037, opened 2019-11-23
[backport-3_10]Fix QGIS crashing when loading a recent project after saving dirty project

nirvn:


nirvn:
The reason why the crash occured: when the user saves a dirty project (via saveDirty()) while opening another project, the recent projects sub menu is repopulated, deleting the pre-existing QAction(s). When saveDirty() is finished and the code returns to openProject( QAction *action ), the pointer refers to a deleted QAction object.

Hence why we need to fetch the project path string prior to calling saveDirty().

nyalldawson (approved):

PR #33039, opened 2019-11-23
[feature][processing] Port and optimization of Random points in extent algorithm

root676:
## Description
This PR proposes a C++ port and rework of the _**Random points in extent**_ algorithm - please feel free to review/edit/comment! Which enhancements where done:

- I decoupled the simple random point generation from the more sophisticated random point generation which enforces minimum distances between the generated points. 
- The minimum distance parameters are now optional which allows the unrestricted random point generation to work way faster than before. 
- The algorithm now also imposes stricter checks on minimum distance between points: If a point cannot be added even after a predefined number of attempts due to distance restrictions, the algorithm stops thowing a `QgsProcessingException`. This keeps the algorithm from handing an uncomplete output to users. 
- Additionally, I exposed a new optional attempts parameter in order to give the user more control over the number of attempts when searching a suitable random point under imposed distance restrictions.

Overall the algorithm is now more efficient in terms of runtime and memory usage: 

1.  unnecessary caching of points for minimum distance evaluations is now directly handled by a `QgsSpatialIndex`
2.  the minimum distance evaluation itself is now handled entirely by the `QgsSpatialIndex`
3.  Avoidance of unnecessary extent geometry checks due to coordinates being generated only inside extent coordinate ranges.

Runtime for 1 million points (without checking for minimum distances between points):

| Language | Runtime |
| ------------- | ------------- |
| Python | 42.85 sec  |
| C++ | 3.2 sec  |
 
_Question on the side_:
While porting the algorithm, I realized that the Random points in layer bounds algorithm becomes (and has been) some kind of obsolete due to the new QGIS3 processing API: The `QgsProcessingParameterExtent` of the _Random points in extent_ algorithm is also capable of passing already loaded layer exents to the algorithm. The only difference here is that the `QgsProcessingParameterFeatureSource` parameter in the _Random points in layer bounds_ algorithm offers using layers from disk which have not been yet loaded. While dropping the _Random points in layer bounds_ alg seems a viable option, it could also affect existing models using this algorithm. Also, the algorithm could easily be ported using the current algorithms logic (...resulting in some kind of code duplication). What do you think about these functional parallels?

## Checklist


alexbruy:
"Random points in layer bounds" behaves a bit different from "Random points in extent" algorithm. While "Random points in extent" generates random points inside layer extent, "Random points in layer bounds" generates points which fall only inside **polygons** of the layer, giving you polygons with random amount of random points in each. It is also different from "Random points inside polygons" which generates **predefined** numer of random points inside each polygon.

Hope now things a bit clearer.

src/analysis/processing/qgsalgorithmrandompointsextent.cpp
```diff
@@ -0,0 +1,180 @@
...
+QString QgsRandomPointsExtentAlgorithm::groupId() const
+{
+  return QStringLiteral( "vectorcreation" );
+}
+
+void QgsRandomPointsExtentAlgorithm::initAlgorithm( const QVariantMap & )
+{
+  addParameter( new QgsProcessingParameterExtent( QStringLiteral( "EXTENT" ), QObject::tr( "Input extent" ) ) );
+  addParameter( new QgsProcessingParameterNumber( QStringLiteral( "NUM_POINTS" ), QStringLiteral( "Number of points" ), QgsProcessingParameterNumber::Integer, 1, false, 1 ) );
+  addParameter( new QgsProcessingParameterCrs( QStringLiteral( "CRS" ), QObject::tr( "Target CRS" ), QStringLiteral( "ProjectCrs" ) ) );
```
alexbruy: Changing parameter name is an API break, as it will break existing scripts and models.

alexbruy (changes requested):
Looking good.

src/analysis/processing/qgsalgorithmrandompointsextent.cpp
nyalldawson: We could possibly make this an optional hidden parameter now, and try to take the crs from the "extent" parameter instead.

src/analysis/processing/qgsalgorithmrandompointsextent.cpp
```diff
@@ -0,0 +1,180 @@
...
+  addParameter( new QgsProcessingParameterFeatureSink( QStringLiteral( "OUTPUT" ), QObject::tr( "Random points" ), QgsProcessing::TypeVectorPoint ) );
+}
+
+QString QgsRandomPointsExtentAlgorithm::shortHelpString() const
+{
+  return QObject::tr( "This algorithm creates a new point layer with a given "
+                      "number of random points, all of them within a given extent. "
+                      "A distance factor can be specified, to avoid points being "
+                      "too close to each other. If the minimum distance between points "
+                      " makes it impossible to create new points, either "
```
nyalldawson: double spaces here - at the start end end of consecutive lines

src/analysis/processing/qgsalgorithmrandompointsextent.cpp
```diff
@@ -0,0 +1,180 @@
...
+      f.setGeometry( QgsGeometry( new QgsPoint( rx, ry ) ) );
+      f.setAttributes( QgsAttributes() << i );
+      sink->addFeature( f, QgsFeatureSink::FastInsert );
+      i++;
+      feedback->setProgress( static_cast<int>( static_cast<double>( i ) / static_cast<double>( mNumPoints ) * 100 ) );
+    }
+  }
+  else
+  {
+    std::unique_ptr<QgsSpatialIndex> index( new QgsSpatialIndex() );
```
nyalldawson: No need to make this on the heap:

```suggestion
    QgsSpatialIndex index;
```

src/analysis/processing/qgsalgorithmrandompointsextent.cpp
```diff
@@ -0,0 +1,180 @@
...
+        break;
+
+      double rx = x_distribution( random_device );
+      double ry = y_distribution( random_device );
+
+      QgsGeometry randomPointGeom = QgsGeometry( new QgsPoint( rx, ry ) );
+
+      //check if new random point is inside searching distance to existing points
+      QList<QgsFeatureId> neighbors = index->nearestNeighbor( randomPointGeom, 1, mDistance );
+      if ( neighbors.size() == 0 )
```
nyalldawson:
```suggestion
      if ( neighbors.empty() )
```

src/analysis/processing/qgsalgorithmrandompointsextent.cpp
```diff
@@ -0,0 +1,180 @@
...
+  return QStringLiteral( "vectorcreation" );
+}
+
+void QgsRandomPointsExtentAlgorithm::initAlgorithm( const QVariantMap & )
+{
+  addParameter( new QgsProcessingParameterExtent( QStringLiteral( "EXTENT" ), QObject::tr( "Input extent" ) ) );
+  addParameter( new QgsProcessingParameterNumber( QStringLiteral( "NUM_POINTS" ), QStringLiteral( "Number of points" ), QgsProcessingParameterNumber::Integer, 1, false, 1 ) );
+  addParameter( new QgsProcessingParameterCrs( QStringLiteral( "CRS" ), QObject::tr( "Target CRS" ), QStringLiteral( "ProjectCrs" ) ) );
+  addParameter( new QgsProcessingParameterDistance( QStringLiteral( "MIN_DISTANCE" ), QObject::tr( "Minimum distance between points" ), 0, QStringLiteral( "CRS" ), true, 0 ) );
+  addParameter( new QgsProcessingParameterNumber( QStringLiteral( "MAX_ATTEMPTS" ), QStringLiteral( "Maximum number of attempts given the minimum distance" ), QgsProcessingParameterNumber::Integer, 200, true, 1 ) );
```
nyalldawson: This should probably be flagged as an advanced parameter

src/analysis/processing/qgsalgorithmrandompointsextent.cpp
```diff
@@ -0,0 +1,180 @@
...
+        i++;
+        distCheckIterations = 0; //reset distCheckIterations if a point is added
+        feedback->setProgress( static_cast<int>( static_cast<double>( i ) / static_cast<double>( mNumPoints ) * 100 ) );
+      }
+      else
+      {
+        if ( distCheckIterations == mMaxAttempts )
+        {
+          throw QgsProcessingException( QObject::tr( "%1 of %2 points have been successfully created, but no more random points could be found "
+                                        " due to the given minimum distance between points. Either choose a larger extent, "
```
nyalldawson: Some double spaces here too

root676:
@alexbruy Thanks for the explanation - I just had a quick look at the algorithm yesterday and must have missed the point generation in polygons. It makes sense to me now.

src/analysis/processing/qgsalgorithmrandompointsextent.cpp
root676: @nyalldawson I tried hiding the parameter from the user with the `FlagHidden` enum but this seems to make the crs parameter also invisible for the `QgsProcessingParameterDistance` (resulting in unknown units for this parameter). This is due to `QgsProcessingParameterDistance` not being able to get CRSs from extent parameters.  I declared it now as an advanced parameter as workaround.

nyalldawson: Ok, let's leave it for now. I'd move it back from being an advanced parameter if we're still actively using it.

src/analysis/processing/qgsalgorithmrandompointsextent.cpp
```diff
@@ -0,0 +1,195 @@
...
+      if ( feedback && feedback->isCanceled() )
+        break;
+
+      double rx = x_distribution( random_device );
+      double ry = y_distribution( random_device );
+
+      QgsGeometry randomPointGeom = QgsGeometry( new QgsPoint( rx, ry ) );
+
+      //check if new random point is inside searching distance to existing points
+      QList<QgsFeatureId> neighbors = index.nearestNeighbor( randomPointGeom, 1, mDistance );
```
nyalldawson: Very minor optimisation, but worthwhile:

```suggestion
      QList<QgsFeatureId> neighbors = index.nearestNeighbor( QgsPointXY( rx, ry), 1, mDistance );
```

and then move `QgsGeometry randomPointGeom = QgsGeometry( new QgsPoint( rx, ry ) );`
inside the `if ( neighbors.empty() )` branch. This avoids creating the QgsPoint on the heap whenever it violates the min distance constraint.

root676: Thanks @nyalldawson, I just added your remarks.

nyalldawson (approved):

nyalldawson:
Thanks @root676!

Do you have plans to port the random points in polygons algorithm in future? If you do, let me know before you touch this one because there's some big changes required to fully optimise that one.

src/analysis/processing/qgsalgorithmrandompointsextent.cpp
nyalldawson: There's just this point remaining ^^ and we can merge

root676:
Thanks @nyalldawson I will let you know if I start porting this algorithm!

src/analysis/processing/qgsalgorithmrandompointsextent.cpp
root676: Now we can merge ;)

python/plugins/processing/tests/testdata/qgis_algorithm_tests3.yaml
```diff
@@ -575,13 +575,14 @@ tests:
             __all__:
               precision: 4
 
-  - algorithm: qgis:randompointsinextent
+  - algorithm: native:randompointsinextent
     name: Random point in extent, don't check result
     params:
-      TARGET_CRS: EPSG:4326
+      CRS: EPSG:4326
```
nyalldawson: You'll need to revert the changes to parameter names here, that's what's causing the Travis failure

root676: Thank you for being so patient with me - I totally missed that parameter.

DelazJ:
@nyalldawson @root676 from a user documentation point of view, is it "only" a port to c++ or are there changes that would need doc update (parameter names, order of parameters...)? The point is to know whether it's worth keeping https://github.com/qgis/QGIS-Documentation/issues/4492 open in docs repo. Thanks

nyalldawson:
There was a new "max iterations" parameter added

DelazJ:
Thanks

agiudiceandrea:
> While "Random points in extent" generates random points inside layer extent, "Random points in layer bounds" generates points which fall only inside **polygons** of the layer, giving you polygons with random amount of random points in each. 

Hi @alexbruy, thank you for the clarification.

By the way, it seems to me that the description of the algorithm "Random points in layer bounds" in the [user manual](https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/vectorcreation.html#random-points-in-layer-bounds) and in the [algorithm help](https://github.com/qgis/QGIS/blob/master/python/plugins/processing/algs/help/qgis.yaml#L295-L296) are not so clear and could be improved.
It states that the tool "creates a new point layer with a given number of random points, all of them within the extent of a given layer...".

PR #33043, opened 2019-11-24
[FEATURE][diagrams] New diagram type "stacked bars"

nyalldawson:
Stacks bars of varying colors for each attribute on top of each other
vertically or horizontally.

Designed to match the "Stacked" chart renderer option available in ArcGIS.

Sponsored by SLYR

image


src/core/diagram/qgsdiagram.h
```diff
@@ -39,6 +39,21 @@ class CORE_EXPORT QgsDiagram
 {
   public:
 
+#ifdef SIP_RUN
+    SIP_CONVERT_TO_SUBCLASS_CODE
+    if ( sipCpp->diagramName() == QStringLiteral( "Pie" ) )
```
nirvn: Nice; that sip conversion could probably make its way into 3.10 (minus the stacked bit)

nirvn (approved):

src/core/diagram/qgsdiagram.h
nyalldawson: Yep, definitely. (that said, it's actually not that vital given that the subclasses don't really expose anything useful)

roya0045:
Hi there, just tested it when checking the diagram issue and I noticed there was always a gap between categories. And when checking the validation images of this PR the gap is also there. I wanted to know if it was intended as I thought it was a bug. Thanks!

PR #33044, opened 2019-11-24
[diagrams] Paint effect support for diagram renderer

nyalldawson:
Allows for diagrams to use paint effects, including drop shadows, outer glows, etc...

Sponsored by SLYR


nirvn (approved):
I love that it only takes ~200 lines to add effects here :) code looks good.

PR #33045, opened 2019-11-24
[wip] Esri mapserver connections api

elpaso:
## UNFINISHED for comments only

Implementation of the new connections API for web services (OGC and ESRI), the base abstract class is designed to handle a generic web service (OGC or ESRI) connection for both rasters and features, providers will need to implement their specific parts.

This initial implementation only handles AMS and AFS and it provides a partial implementation limited to connections storage and retrieval.


nyalldawson (commented):
Looking good to me! One thing is that we'll need to merge the afs and ams connection classes, since a single server mixes and matches these two services in a free-form way. But I'm happy to tackle this part of it!

PR #33048, opened 2019-11-24
QgsDebugMsgLevel optimisations

nyalldawson:
Rework QgsDebugMsgLevel to avoid construction of strings which won't be logged at the current debug level

Instead of always constructing debug strings, and then potentially ignoring them if they fall outside the current debug level, we instead rework the QgsDebugMsgLevel macro so that strings are only ever constructed when they WILL be logged.

This avoids the (often very expensive) string construction for debug messages whenever the results won't be used. It allows low level (i.e. level 3 or 4) debug messages to be safely used without
incurring huge slowdowns in debug builds.

TODO: ensure we only ever use QgsDebugMsg() for ERROR reporting, and move all other debugging calls to QgsDebugMsgLevel instead.

Credit for original idea goes to @wonder-sk!

### DIFF
--- a/python/core/auto_generated/qgsfield.sip.in
+++ b/python/core/auto_generated/qgsfield.sip.in
@@ -299,9 +299,7 @@ Converts the provided variant to a compatible format
         Py_BEGIN_ALLOW_THREADS
         try
         {
-          QgsDebugMsg( a0->toString() );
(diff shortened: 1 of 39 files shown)


jef-n (approved):

nyalldawson:
Unrelated test failure

Issue #30569, opened 2019-07-05
SRS transformation failures with PROJ 6.1.0


marisn:
Steps to reproduce:
add multiple data sources (in my example – three ESRI Shapefiles) with different SRS'es.

Actual result:
> No transform available between EPSG:3059 and EPSG:4326
> No transform available between EPSG:3059 and EPSG:25884

Expected result:
no error.

Valgrind output contains usual chatter about Python, still there is something interesting during exiting from QGIS, (Invalid write of size 4 in pj_ctx_set_errno) still it has nothing to do with this issue.

Plain debug output is a bit more interesting (from 3.9, still the same issue is also observable in 3.8.0):

> src/core/providers/ogr/qgsogrprovider.cpp:1277 : (extent) [9ms] Starting get extent
> src/core/providers/ogr/qgsogrprovider.cpp:1321 : (extent) [7ms] Finished get extent
> src/core/providers/ogr/qgsogrprovider.cpp:1277 : (extent) [176ms] Starting get extent
> src/core/providers/ogr/qgsogrprovider.cpp:1321 : (extent) [1ms] Finished get extent
> src/core/qgscoordinatetransform_p.cpp:234 : (initialize) [410ms] Coordinate transformation failed to initialize!
> src/core/providers/ogr/qgsogrprovider.cpp:1277 : (extent) [3ms] Starting get extent
> src/core/providers/ogr/qgsogrprovider.cpp:1321 : (extent) [1ms] Finished get extent
> src/gui/qgsmapcanvas.cpp:506 : (refresh) [96ms] CANVAS refresh already scheduled
> src/core/qgscoordinatetransform.cpp:749 : (transformCoords) [901ms] Projection failed emitting invalid transform signal: forward transform of
> (inf, inf)
> (inf, inf)
> .....
> (inf, inf)
> (inf, inf)
> PROJ: +proj=pipeline +step +proj=unitconvert +xy_in=deg +xy_out=rad +step +proj=tmerc +lat_0=0 +lon_0=24 +k=0.9996 +x_0=500000 +y_0=0 +ellps=GRS80
> Error: latitude or longitude exceeded limits
> src/core/qgscoordinatetransform.cpp:750 : (transformCoords) [5ms] throwing exception
> src/core/qgscoordinatetransform.cpp:571 : (transformBoundingBox) [32ms] rethrowing exception
> src/gui/qgsmapcanvas.cpp:393 : (setDestinationCrs) [36ms] Transform error caught: forward transform of
[...]

PR #33049, opened 2019-11-25
Hide coordinate operations which utilise a LAS/LOS grid shift file

nyalldawson:
which are unsupported by proj 6

E.g. some candidate operations between EPSG:3035 and EPSG:5514

Since these can NEVER (as of now) be instantiated by proj, even if
the grid shift files are present, it's misleading to present them
as options to users.

Refs #30569


nirvn:
@nyalldawson , for the record, I assume those transformations were never supported within QGIS, correct? I.e., our GDAL2/PROJ4 was also unable to deal with those correct?

nyalldawson:
@nirvn -- yes, PROJ has never had support for them

PR #33064, opened 2019-11-25
MXE build workflow

elpaso:
Cross build for windows using MXE and store artifact for testing.

### DIFF
--- a/.github/workflows/mxe.yml
+++ b/.github/workflows/mxe.yml
@@ -1,0 +1,21 @@
+name: Windows cross build with MXE
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    
+    steps:
+    - uses: actions/checkout@v1
+    - name: Cross build with MXE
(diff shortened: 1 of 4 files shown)


nyalldawson:
Is there anything of use here? https://github.com/kadas-albireo/QGIS/commit/ce3862f23fbe1d26d509613a14e3e11ac09a0947

elpaso:
> Is there anything of use here? [kadas-albireo@ce3862f](https://github.com/kadas-albireo/QGIS/commit/ce3862f23fbe1d26d509613a14e3e11ac09a0947)

Definitely!

elpaso:
@nyalldawson I spend some time on the kadas-albireo cross build, the advantage is that it has Python but the problem is that it doesn't work :) 

It is also much slower, bigger and complex. 

I'd stick to MXE for now, but I'll make a PR with the work I've done on the the other build system.

elpaso:
See also: https://github.com/qgis/QGIS/pull/33085

elpaso:
Ok, let's see how this behaves, if it's too slow we can revert it.

PR #33066, opened 2019-11-25
Port "Count points in polygon" algorithm to c++

nyalldawson:
Joining in on the recent Processing c++ port frenzy, I've resurrected work I started a long time ago on porting the "Count points in polygon" algorithm to c++.

On my tests this version is around 2x faster on a debug build (likely more on a release build):

With a 2.5 million feature point layer and a 78 feature polygon layer containing the points:

Previous python alg: 118.83 seconds
New c++ port: 61.47 seconds


root676:
So nice! QGIS 3.12 will set a new record in processing alg speed. ;)

nyalldawson:
> QGIS 3.12 will set a new record in processing alg speed. ;)

YES!

@root676 @alexbruy so that we don't overlap efforts, what algorithms are you both potentially looking at porting in the future? I would like to port the spatial join algorithms asap, as a profile revealed that at least 30% of the processing time of these algs is wasted in converting variants (feature attributes) to and from python values... ouch!

nirvn (approved):
Nice, I use this algorithm _a lot_, will enjoy the speed boost :)

nyalldawson:
Thanks @nirvn!

root676:
> what algorithms are you both potentially looking at porting in the future?

@nyalldawson Good idea, I would like to take a look at reworking **voronoi creation**... but this will surely cost me some time  because my time resources are somewhat limited next month and the algorithm imports this interesting `voronoi.py` module... The **concave hull** alg also looks interesting.

alexbruy:
@nyalldawson I plan to look at some vector creation tools like rectangles/ovals/diamonds and probably some general algortihms too.

roya0045:
I would like to port the spatialjoin alg if no one is willing to do it, thought I can't promise anything as I don't have much time to program extra stuff these days.

PR #33073, opened 2019-11-26
[layouts] Legend widget UI/UX improvements

nirvn:
## Description

Screenshot of PR:
image

I'm trying to improve the messy situation with regards to tool buttons (i.e. eEE) usage for the legend item widget.

Changes done to improve UI/UX:
- make the filter content by map extent switch a check box, away from being an icon-only tool button. This harmonize the legend widget with the attribute table item widget, which also shows this option as a checkbox, and reduces the current visual clusterf***
- move the filter by expression icon away from the "show count" and "add/edit expression" tool buttons, as those three buttons are otherwise visually very similar in nature and can get confusing at first glance
- change the "add/edit expression" tool button icon to remove the green plus sign, it's confusing in this context, and add visual complexity to an already busy area

## Checklist


nyalldawson (approved):
Good changes!

roya0045:
Somehow the icon that I made has changed, the + symbol is gone. :/

roya0045:
@nirvn was the change to my icon intended? I'm just curious if having the expression symbol without the + is better or not. Possibly is, I'm on the fence about that. But I felt like having an epsilon alone was kinda dull and the green + helped. But maybe I'm wrong in terms of UX/UI.

PR #33075, opened 2019-11-26
3.10 UI/UX fix backports

nirvn:


DelazJ:
@nirvn GUI changes backport should be notified to docs writers because they [break doc](https://docs.qgis.org/testing/en/docs/user_manual/print_composer/composer_items/composer_legend.html#legend-items). This is the [needs-docs] purpose.

nirvn:
@DelazJ , right, missed that (again). Is there a PR label trigger these days?

DelazJ:
Yes. Never gone. You can still add the tag in the PR title or when merging, and the issue will be created in the docs but this PR title actually means nothing and writers will still have to browse this repo to find the concerned changes. Not helpful for them.
The best scenario IMO is to manually open reports in the doc repo pointing directly to the concerned commits (or their original PR) when they are fresh.

nyalldawson (approved):

PR #33086, opened 2019-11-26
Processing temp output path configurable

signedav:
"Temporary output folder path" on Processing Settings.
On default it's still the ordinary temp path.

temp_path

### DIFF
--- a/python/plugins/processing/core/ProcessingConfig.py
+++ b/python/plugins/processing/core/ProcessingConfig.py
@@ -28,7 +28,8 @@ from qgis.core import (NULL,
                        QgsApplication,
                        QgsSettings,
                        QgsVectorFileWriter,
-                       QgsRasterFileWriter)
+                       QgsRasterFileWriter,
+                       QgsProcessingUtils)
 from processing.tools.system import defaultOutputFolder
 import processing.tools.dataobjects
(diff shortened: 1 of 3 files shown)


nyalldawson:
This still seems like an odd requirement -- why not just set the temp folder location on the account level instead? QGIS isn't the only application that writes files to a temp folder...

andreasneumann:
@nyalldawson - this is mainly for restricted requirements where users have no influence on system settings (Citrix on Windows application servers). They are not allowed to set or change env vars.

andreasneumann:
Or for organizations where the central IT services set the restricted "TEMP" env var to the C: drive and then complain that the C: drive is getting full ...

signedav:
I admit there could have been more explanation in the PR description. Thanks for the clarification @andreasneumann

Since there is often a big amount of data written and often *[Save to temporary file]* selected per default, it can happen that the user fills up the storage. So we can set an alternative temp folder location for this case, while other applications can still use the ordinary temp location.

nyalldawson:
Can you add some unit tests for this too? (Including one where you change the path after first calculating the static variable)

nyalldawson (approved):

signedav:
Thanks @nyalldawson

PR #33091, opened 2019-11-27
Make PAL problem solver a bit more memory safe, very slightly less... confusing...

nyalldawson:
Modernises a bit of code, renames some cryptic variable

(While it doesn't fix any bugs, I plan to backport this to make future backports easier)

### DIFF
--- a/src/core/pal/pal.cpp
+++ b/src/core/pal/pal.cpp
@@ -96,11 +96,11 @@ struct FeatCallBackCtx
 {
   Layer *layer = nullptr;
 
-  std::list< std::unique_ptr< Feats > > *fFeats = nullptr;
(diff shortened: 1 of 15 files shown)


src/core/pal/pal.cpp
```diff
@@ -133,7 +133,7 @@ bool extractFeatCallback( FeaturePart *ft_ptr, void *ctx )
     Feats *ft = new Feats();
     ft->feature = ft_ptr;
     ft->shape = nullptr;
-    ft->lPos = lPos;
+    ft->candidates = lPos;
```
nirvn: Suggestion: also rename the const QList< LabelPosition * > lPos variable declared above to candidates.

nirvn (approved):

m-kuhn (approved):
Big +1

PR #33092, opened 2019-11-27
[pal] Don't sort final label placements in pal

nyalldawson:
We have a (better) sorting method in QGIS which accounts for other things like the label z orders, so this sorting in the pal library is completely redundant and is immediately overwritten by QGIS label sorting routines.

### DIFF
--- a/src/core/pal/problem.cpp
+++ b/src/core/pal/problem.cpp
@@ -816,11 +816,6 @@ void Problem::chain_search()
   delete[] ok;
 }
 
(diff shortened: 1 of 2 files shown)


nirvn (approved):
Is there a noticeable speed boost attached to this change for label-heavy canvases?

nyalldawson:
@nirvn 
I didn't test, but I doubt it -- the container only holds labels which will be actually rendered, so at worst there's only gonna be 1000 or so...

PR #33093, opened 2019-11-27
Expose a project labeling option for "labeling engine version"

nyalldawson:
Adds the underlying API (currently unused) for specifying a labeling engine placement version to use for a project, and the corresponding setting in the project labeling setting dialog.

This allows users to upgrade existing projects to newer labeling engine versions. A warning will show advising users that change the version will alter the placement of labels within the project.

While currently unused, this is the framework which allows us the flexibility to tweak the labeling engine costs and obstacle avoidance strategies without altering the label placement in existing projects.

### DIFF
--- a/src/ui/qgslabelengineconfigdialog.ui
+++ b/src/ui/qgslabelengineconfigdialog.ui
@@ -160,19 +160,13 @@
    </item>
    <item>
     <layout class="QGridLayout" name="gridLayout_2" columnstretch="0,0,0">
-     <item row="2" column="0" colspan="3">
-      <widget class="QCheckBox" name="chkShowAllLabels">
-       <property name="sizePolicy">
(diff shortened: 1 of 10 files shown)


src/ui/qgslabelengineconfigdialog.ui
```diff
@@ -11,7 +11,7 @@
    </rect>
   </property>
   <property name="windowTitle">
-   <string>Automated Placement Engine</string>
+   <string>Project Label Settings</string>
```
nirvn: Isn't it the case that the dialog is also linked to diagram placement? If so, might be good to have a more generic title. The dialog _can_ be opened via the diagram's panel / dialog too.

nyalldawson: Ok, reverted

nirvn (approved):

PR #33094, opened 2019-11-27
When exporting CRS information from GDAL >= 3.0, use WKT2 format

nyalldawson:
...to avoid loss of CRS information

### DIFF
--- a/src/core/providers/gdal/qgsgdalprovider.cpp
+++ b/src/core/providers/gdal/qgsgdalprovider.cpp
@@ -293,6 +293,7 @@ QgsGdalProvider *QgsGdalProvider::clone() const
   return new QgsGdalProvider( *this );
 }
 
+#if GDAL_VERSION_NUM < GDAL_COMPUTE_VERSION(3,0,0)
 bool QgsGdalProvider::crsFromWkt( const char *wkt )
 {
 
@@ -330,6 +331,7 @@ bool QgsGdalProvider::crsFromWkt( const char *wkt )
 
   return mCrs.isValid();
 }
+#endif
 
 bool QgsGdalProvider::getCachedGdalHandles( QgsGdalProvider *provider,
     GDALDatasetH &gdalBaseDataset,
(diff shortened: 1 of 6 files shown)


nyalldawson:
Look ok to you @rouault ?

src/core/providers/gdal/qgsgdalprovider.cpp
```diff
@@ -317,7 +317,15 @@ bool QgsGdalProvider::crsFromWkt( const char *wkt )
       CPLFree( pszProj4 );
 
       char *pszWkt = nullptr;
+#if GDAL_VERSION_NUM >= GDAL_COMPUTE_VERSION(3,0,0)
+      const QByteArray multiLineOption = QStringLiteral( "MULTILINE=NO" ).toLocal8Bit();
```
rouault: looking at the logic of the whole function, I don't get it. So we have as input... wkt, that we import as a OGRSpatialReferenceH object, then export it as WKT again, to finally pass it to QgsCoordinateReferenceSystem::fromWkt() ... What's the point :-) ? I suspect this might have to do with some normalization, but this looks really odd. Not even sure this is needed at all

rouault:
> Look ok to you @rouault ?

yeah mostly... Well there are a few cases where WKT2 export will not work properly. Like how we deal with rotated lon-lat grids in the netCDF driver due to using WKT1 hacks... Ultimately we should try to get rid of that, but in the mean time, I've done things like https://github.com/OSGeo/gdal/commit/79abc711dde3cdabf14723b67eb96b8c551d33c3#diff-bda4d1c4f188e87825d8f6c38f7166b9R894 . Maybe we would need some utility function in QGIS, exportToMostSuitableWKT() ?, to hide such potential hacks.

src/core/providers/gdal/qgsgdalprovider.cpp
nyalldawson: Ehh yeah 🤢 , I should have scrolled up and looked at the wider context here (I was originally just grepping for OSRExportToWkt). I'll refactor this stuff, it's gross.

nyalldawson:
Ok, how's the revised version @rouault ?

rouault:
LGTM

PR #33095, opened 2019-11-27
Keep order of value map items when importing CSV

m-kuhn:
Backport of #33060

### DIFF
--- a/src/gui/editorwidgets/qgsvaluemapconfigdlg.cpp
+++ b/src/gui/editorwidgets/qgsvaluemapconfigdlg.cpp
@@ -173,7 +173,7 @@ void QgsValueMapConfigDlg::updateMap( const QList<QPair<QString, QVariant>> &lis
     ++row;
   }
 
-  for ( const auto &pair : list )
+  for ( const auto &pair : qgis::as_const( list ) )
   {
     if ( pair.second.isNull() )
       setRow( row, pair.first, QString() );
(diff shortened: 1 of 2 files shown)


src/gui/editorwidgets/qgsvaluemapconfigdlg.cpp
```diff
@@ -147,6 +147,18 @@ void QgsValueMapConfigDlg::removeSelectedButtonPushed()
 }
 
 void QgsValueMapConfigDlg::updateMap( const QMap<QString, QVariant> &map, bool insertNull )
+{
+  QList<QPair<QString, QVariant>> orderedMap;
+  auto end = map.constEnd();
```
elpaso: Love all these autos :car: !

src/gui/editorwidgets/qgsvaluemapconfigdlg.cpp
```diff
@@ -161,12 +173,12 @@ void QgsValueMapConfigDlg::updateMap( const QMap<QString, QVariant> &map, bool i
     ++row;
   }
 
-  for ( QMap<QString, QVariant>::const_iterator mit = map.begin(); mit != map.end(); ++mit, ++row )
+  for ( const auto &pair : list )
```
elpaso: I know it's perfectly pointless in this case but someone seemed to prefer `qgis::as_const` for readability in the past, is there a reason for not using it here?

elpaso (approved):

src/gui/editorwidgets/qgsvaluemapconfigdlg.cpp
m-kuhn:
```suggestion
  for ( const auto &pair : qgis::as_const( list ) )
```

PR #33099, opened 2019-11-27
Visible status for embedded layers in embedded group

### LINKED ISSUE #33097: Visible status for embedded layers in embedded group
**Describe the bug**
Concern embedded layers when they are in embedded group.

For project, visible status for embedded layers is not independent of their embedded group.

In the .qgs file , if the visible status is unchecked for an embedded group, all layers of this group are listed in the `embedded-invisible-layers` element, what ever if they are themselves check or uncheck. Consequently, if user saves a project with the group uncheck but some layers within checked, all layers of the group are unchecked when user open again the project.

For theme, the embedded group has no "visibility state" information when user saves a theme. Consequently, when user saves a theme, for example with group unchecked, the state doesn't change when applying theme, even if the group is checked.

**How to Reproduce**

1. Open a new project
2. Add layers, for example the world_map.gpkg, present in qgis\data\resources\data
3. Add these layers to a group
4. Save the project
5. Open a new project and load the group with `Layer => Embed Layers and Groups`
6. Check all layers and uncheck the group, save the project.
7. Re-open the project, all checks on layers are gone.

For theme

8. Check layers and uncheck group, save a theme
9. Apply the theme, all checks on layers are gone, you can also see that applying theme has no effect on group visibility state

**QGIS and OS versions**

QGIS version | 3.11.0-Master | QGIS code revision | d0348127bf
-- | -- | -- | --
OS Version | Ubuntu 19.04 | This copy of QGIS writes debugging output.
Active python plugins | redLayer; go2streetview; coordinator; processing; db_manager

Ubuntu 19.04

**Additional context**

SebastienPeillet:
- For project : check visible state for embedded layers inside an unchecked group, instead of putting all layers in embedded-invisible-layers
- For theme : Add an 'checked-group-node' to save group visible state independently to layers in it.
Fixes #33097

Hi,
My name is Sébastien Peillet and I'm a new member of Oslandia since September. It's been a few years that I began to write python plugin for Qgis and now that I am part of Oslandia, I hope I could contribute to the Qgis project in a more broadly way. All tips are welcome !

## Description
Issue explained in #33097 

To fix theses issues, some changes has been made : 

For project issue, all embbeded layers in a embedded group were listed as `embedded-invisible-layers` if the group visible state was unchecked. It has been changed with a if statement on layers visible state : 

```
      if ( child->itemVisibilityChecked() == Qt::Unchecked )
      {
        list << invisibleLayerList( child );
      }
```

For theme issue, layers are saved in the `visibility-preset` element. In the same way as before, if the group is unchecked, all layers checked within are not saved in the `visibility-preset` element, because the test is : 
```
     if ( nodeLayer->isVisible() )
```
I replaced it with:
```
     if ( node->itemVisibilityChecked() != Qt::Unchecked )
```
Like this, checked layers are saved, even if group are unchecked

And futhermore, groups have no visibility state. So I add a `has-checked-group-info` tag (similar to the `has-expanded-info` tag) in the `visibility-preset` element. It allows to list the checked group and apply the theme correctly (before there was no change applied on the group)

## Checklist

### DIFF
--- a/src/core/qgsmapthemecollection.h
+++ b/src/core/qgsmapthemecollection.h
@@ -155,7 +155,7 @@ class CORE_EXPORT QgsMapThemeCollection : public QObject
          * Returns whether information about checked/unchecked state of groups has been recorded
          * and thus whether checkedGroupNodes() is valid.
          */
-        bool hasCheckedStateInfo() const { return mHasCheckedStateInfo; }
+        bool hasCheckedStateInfo() const { return mHasCheckedStateInfo; } const SIP_SKIP;
 
         /**
          * Sets whether the map theme contains valid expanded/collapsed state of nodes
@@ -166,7 +166,7 @@ class CORE_EXPORT QgsMapThemeCollection : public QObject
         /**
          * Sets whether the map theme contains valid checked/unchecked state of group nodes
          */
-        void setHasCheckedStateInfo( bool hasInfo ) { mHasCheckedStateInfo = hasInfo; }
+        void setHasCheckedStateInfo( bool hasInfo ) { mHasCheckedStateInfo = hasInfo; } const SIP_SKIP;
 
         /**
          * Returns a set of group identifiers for group nodes that should have expanded state (other group nodes should be collapsed).
@@ -182,6 +182,7 @@ class CORE_EXPORT QgsMapThemeCollection : public QObject
          * The returned value is valid only when hasCheckedStateInfo() returns TRUE.
          * Group identifiers are built using group names, a sub-group name is prepended by parent group's identifier
          * and a forward slash, e.g. "level1/level2"
+         * \since QGIS 3.10.1
          */
         QSet<QString> checkedGroupNodes() const { return mCheckedGroupNodes; }
 
@@ -193,6 +194,7 @@ class CORE_EXPORT QgsMapThemeCollection : public QObject
 
         /**
          * Sets a set of group identifiers for group nodes that should have checked state. See checkedGroupNodes().
+         * \since QGIS 3.10.1
          */
         void setCheckedGroupNodes( const QSet<QString> &checkedGroupNodes ) { mCheckedGroupNodes = checkedGroupNodes; }
 
--- a/src/core/qgsmapthemecollection.h
+++ b/src/core/qgsmapthemecollection.h
@@ -140,7 +140,6 @@ class CORE_EXPORT QgsMapThemeCollection : public QObject
 
         /**
          * Returns set with only records for valid layers
-         * \note not available in Python bindings
          */
         QHash<QgsMapLayer *, QgsMapThemeCollection::MapThemeLayerRecord> validLayerRecords() const SIP_SKIP;
 
@@ -154,8 +153,10 @@ class CORE_EXPORT QgsMapThemeCollection : public QObject
         /**
          * Returns whether information about checked/unchecked state of groups has been recorded
          * and thus whether checkedGroupNodes() is valid.
+         * \note Not available in Python bindings
+         * \since QGIS 3.10.1
          */
-        bool hasCheckedStateInfo() const { return mHasCheckedStateInfo; } const SIP_SKIP;
+        bool hasCheckedStateInfo() const { return mHasCheckedStateInfo; } SIP_SKIP;
 
         /**
          * Sets whether the map theme contains valid expanded/collapsed state of nodes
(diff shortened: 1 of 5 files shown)


src/core/qgsmapthemecollection.h
```diff
@@ -150,12 +151,23 @@ class CORE_EXPORT QgsMapThemeCollection : public QObject
...
         /**
          * Sets whether the map theme contains valid expanded/collapsed state of nodes
          * \since QGIS 3.2
          */
         void setHasExpandedStateInfo( bool hasInfo ) { mHasExpandedStateInfo = hasInfo; }
 
+        /**
+         * Sets whether the map theme contains valid checked/unchecked state of group nodes
+         */
+        void setHasCheckedStateInfo( bool hasInfo ) { mHasCheckedStateInfo = hasInfo; }
```
nyalldawson: I'd suggest hiding this from the sip bindings -- it's not something we really want in the public stable API

src/core/qgsmapthemecollection.h
```diff
@@ -150,12 +151,23 @@ class CORE_EXPORT QgsMapThemeCollection : public QObject
          */
         bool hasExpandedStateInfo() const { return mHasExpandedStateInfo; }
 
+        /**
+         * Returns whether information about checked/unchecked state of groups has been recorded
+         * and thus whether checkedGroupNodes() is valid.
+         */
+        bool hasCheckedStateInfo() const { return mHasCheckedStateInfo; }
```
nyalldawson: See below -- I'd also suggest hiding this from the sip python bindings, as it's just an internal detail rather then something we want to expose

src/core/qgsmapthemecollection.h
```diff
@@ -165,25 +177,46 @@ class CORE_EXPORT QgsMapThemeCollection : public QObject
          */
         QSet<QString> expandedGroupNodes() const { return mExpandedGroupNodes; }
 
+        /**
+         * Returns a set of group identifiers for group nodes that should have checked state (other group nodes should be unchecked).
+         * The returned value is valid only when hasCheckedStateInfo() returns TRUE.
+         * Group identifiers are built using group names, a sub-group name is prepended by parent group's identifier
+         * and a forward slash, e.g. "level1/level2"
```
nyalldawson: \since QGIS 3.10.1

src/core/qgsmapthemecollection.h
```diff
@@ -165,25 +177,46 @@ class CORE_EXPORT QgsMapThemeCollection : public QObject
...
+
         /**
          * Sets a set of group identifiers for group nodes that should have expanded state. See expandedGroupNodes().
          * \since QGIS 3.2
          */
         void setExpandedGroupNodes( const QSet<QString> &expandedGroupNodes ) { mExpandedGroupNodes = expandedGroupNodes; }
 
+        /**
+         * Sets a set of group identifiers for group nodes that should have checked state. See checkedGroupNodes().
+         */
```
nyalldawson: \since QGIS 3.10.1

nyalldawson:
Welcome aboard @SebastienPeillet ! It's great to have you here!

SebastienPeillet:
Hi ! I made the changes you suggested.

To be sure because I'm not used to sip : to hide function, I had to add `const SIP_SKIP;` after function definition ?

nyalldawson:
yes - just adding SIP_SKIP after the function declaration should work

We also usually add 

    \note Not available in Python bindings

to the doxygen comment for the method

nyalldawson:
@SebastienPeillet this change has caused a serious regression -- when creating a theme the parent group's status is ignored. This means that the theme contains any checked layers from this group, even though they aren't visible on the canvas. 

You can see this regression in action if you make a group with some checked layers, uncheck the group, and then create a theme. Switch to a print layout and set a map to follow the theme, and the result will look very different to the canvas rendering of this theme...

Can you fix please?

SebastienPeillet:
Sorry for the late reply @nyalldawson , I just saw you fix the regression as I was at the point to propose my fix ([here](https://github.com/SebastienPeillet/QGIS/tree/fix_theme_visibility_in_print_layout)).

And from what I read, I found your fix better (adding an attribute) than mine (searching for parent visibility at each call of mapThemeVisibleLayers function).

src/core/qgsmapthemecollection.h
signedav: Hi @nyalldawson 
Sorry to come back on such an old comment.

On creating a project with PyQGIS (loading layers, apply QML files etc.) we would like to create the map themes as well. Every bit is supported except this. It's not possible to control the checked state of the group. Why's that? What have been the risks to decide that this should not be in the sip bindings?

Another way to apply map themes to a project would be to create the layer tree completely and then store the current state. But this is much less elegant.

So I would love to enable it in the API. Thanks a lot.

PR #33101, opened 2019-11-27
[field calc] do not propose to update field for read-only layers

3nids:
following discussion in https://lists.osgeo.org/pipermail/qgis-developer/2019-November/059337.html


src/app/qgsfieldcalculator.h
```diff
@@ -57,6 +57,9 @@ class APP_EXPORT QgsFieldCalculator: public QDialog, private Ui::QgsFieldCalcula
 
     QgsVectorLayer *mVectorLayer = nullptr;
 
+    bool mCanAddAttribute;
```
nyalldawson: Initalize

nyalldawson (approved):

PR #33106, opened 2019-11-27
Move the vertical spacer to the bottom of the raster image marker

DelazJ:
symbol dialog and fix tabulation order to avoid this:
image


nirvn:
@DelazJ , nice catch.

PR #33125, opened 2019-11-28
rotate() expression function

raymondnijssen:
## Checklist

### DIFF
--- a/resources/function_help/json/rotate
+++ b/resources/function_help/json/rotate
@@ -1,0 +1,10 @@
+{
+  "name": "rotate",
+  "type": "function",
+  "description": "Returns a rotated version of a geometry. Calculations are in the Spatial Reference System of this geometry.",
+  "arguments": [ {"arg":"geom","description":"a geometry"},
+                 {"arg":"rotation","description":"clockwise rotation in degrees"},
+                 {"arg":"point","description":"rotation center"}
+               ],
+  "examples": [ { "expression":"translate($geometry, 45, centroid($geometry))", "returns":"a geometry of the same type like the original one"}]
+}
--- a/resources/function_help/json/rotate
+++ b/resources/function_help/json/rotate
@@ -6,5 +6,5 @@
                  {"arg":"rotation","description":"clockwise rotation in degrees"},
                  {"arg":"point","description":"rotation center"}
                ],
-  "examples": [ { "expression":"translate($geometry, 45, centroid($geometry))", "returns":"a geometry of the same type like the original one"}]
+  "examples": [ { "expression":"rotate($geometry, 45, centroid($geometry))", "returns":"a rotated geometry of the same type as the original one"}]
 }
--- a/tests/src/core/testqgsexpression.cpp
+++ b/tests/src/core/testqgsexpression.cpp
@@ -2875,6 +2875,15 @@ class TestQgsExpression: public QObject
       QTest::newRow( "translate" ) << "translate( $geometry, -1, 2)" << geom << false << true << QgsGeometry::fromWkt( QStringLiteral( "LINESTRING (-1 2, 9 12)" ) );
       geom = QgsGeometry::fromPointXY( point );
       QTest::newRow( "translate" ) << "translate( $geometry, 1, -2)" << geom << false << true << QgsGeometry::fromWkt( QStringLiteral( "POINT(1 -2)" ) );
+
+      geom = QgsGeometry::fromPolygonXY( polygon );
+      QTest::newRow( "rotate" ) << "rotate( $geometry, 90, geomFromWKT( 'POINT(10 10)' ) )" << geom << false << true << QgsGeometry::fromWkt( QStringLiteral( "Polygon ((0 20, 10 10, 0 10, 0 0))" ) );
+      geom = QgsGeometry::fromPolygonXY( line );
+      QTest::newRow( "rotate" ) << "rotate( $geometry, -90, geomFromWKT( 'POINT(5 5)' ) )" << geom << false << true << QgsGeometry::fromWkt( QStringLiteral( "LineString (10 0, 10 10)" ) );
+      geom = QgsGeometry::fromPolygonXY( point );
+      QTest::newRow( "rotate" ) << "rotate( $geometry, 180, geomFromWKT( 'POINT(-5 -3)' ) )" << geom << false << true << QgsGeometry::fromWkt( QStringLiteral( "Point (-10 -6)" ) );
+      QTest::newRow( "rotate" ) << "rotate( $geometry, 180, geomFromWKT( 'MULTIPOINT((-5 -3))' ) )" << geom << false << true << QgsGeometry::fromWkt( QStringLiteral( "Point (-10 -6)" ) );
+      //QTest::newRow( "rotate" ) << "rotate( $geometry, 180, geomFromWKT( 'MULTIPOINT((-5 -3, 7 7, 4 4))' ) )" << geom << false << true << QgsGeometry::fromWkt( QStringLiteral( "Point (-10 -6)" ) ); // should not work
     }
 
     void eval_geometry_method()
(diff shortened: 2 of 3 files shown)


tests/src/core/testqgsexpression.cpp
```diff
@@ -2875,6 +2875,15 @@ class TestQgsExpression: public QObject
...
+
+      geom = QgsGeometry::fromPolygonXY( polygon );
+      QTest::newRow( "rotate" ) << "rotate( $geometry, 90, geomFromWKT( 'POINT(10 10)' ) )" << geom << false << true << QgsGeometry::fromWkt( QStringLiteral( "Polygon ((0 20, 10 10, 0 10, 0 20))" ) );
+      geom = QgsGeometry::fromPolylineXY( line );
+      QTest::newRow( "rotate" ) << "rotate( $geometry, -90, geomFromWKT( 'POINT(5 5)' ) )" << geom << false << true << QgsGeometry::fromWkt( QStringLiteral( "LineString (10 0, 0 10)'" ) );
+      geom = QgsGeometry::fromPointXY( point );
+      QTest::newRow( "rotate" ) << "rotate( $geometry, 180, geomFromWKT( 'POINT(-5 -3)' ) )" << geom << false << true << QgsGeometry::fromWkt( QStringLiteral( "Point (-10 -6)" ) );
+      QTest::newRow( "rotate" ) << "rotate( $geometry, 180, geomFromWKT( 'MULTIPOINT((-5 -3))' ) )" << geom << false << true << QgsGeometry::fromWkt( QStringLiteral( "Point (-10 -6)" ) );
+      //QTest::newRow( "rotate" ) << "rotate( $geometry, 180, geomFromWKT( 'MULTIPOINT((-5 -3, 7 7, 4 4))' ) )" << geom << true << true << QgsVariant();
```
raymondnijssen: I'd love to get this last (should fail) test working but I don't know how. Hope somebody can help me. Or I could just remove it.

tests/src/core/testqgsexpression.cpp
```diff
@@ -2875,6 +2875,15 @@ class TestQgsExpression: public QObject
       QTest::newRow( "translate" ) << "translate( $geometry, -1, 2)" << geom << false << true << QgsGeometry::fromWkt( QStringLiteral( "LINESTRING (-1 2, 9 12)" ) );
       geom = QgsGeometry::fromPointXY( point );
       QTest::newRow( "translate" ) << "translate( $geometry, 1, -2)" << geom << false << true << QgsGeometry::fromWkt( QStringLiteral( "POINT(1 -2)" ) );
+
+      geom = QgsGeometry::fromPolygonXY( polygon );
```
nyalldawson: I'd suggest adapting these tests to use geom_from_wkt instead of $geometry -- it means the tests will be stand alone and can be posted directly into the qgis app to help diagnose the results

nyalldawson (commented):
Looking good!

tests/src/core/testqgsexpression.cpp
```diff
@@ -2875,6 +2875,15 @@ class TestQgsExpression: public QObject
...
+      QTest::newRow( "rotate" ) << "rotate( $geometry, 90, geomFromWKT( 'POINT(10 10)' ) )" << geom << false << true << QgsGeometry::fromWkt( QStringLiteral( "Polygon ((0 20, 10 10, 0 10, 0 20))" ) );
+      geom = QgsGeometry::fromPolylineXY( line );
+      QTest::newRow( "rotate" ) << "rotate( $geometry, -90, geomFromWKT( 'POINT(5 5)' ) )" << geom << false << true << QgsGeometry::fromWkt( QStringLiteral( "LineString (10 0, 0 10)'" ) );
+      geom = QgsGeometry::fromPointXY( point );
+      QTest::newRow( "rotate" ) << "rotate( $geometry, 180, geomFromWKT( 'POINT(-5 -3)' ) )" << geom << false << true << QgsGeometry::fromWkt( QStringLiteral( "Point (-10 -6)" ) );
+      QTest::newRow( "rotate" ) << "rotate( $geometry, 180, geomFromWKT( 'MULTIPOINT((-5 -3))' ) )" << geom << false << true << QgsGeometry::fromWkt( QStringLiteral( "Point (-10 -6)" ) );
+      QTest::newRow( "rotate" ) << "rotate( $geometry, 180, geomFromWKT( 'MULTIPOINT((-5 -3, 7 7, 4 4))' ) )" << geom << true << true << QgsVariant();
```
m-kuhn:
```suggestion
      QTest::newRow( "rotate" ) << "rotate( $geometry, 180, geomFromWKT( 'MULTIPOINT((-5 -3, 7 7, 4 4))' ) )" << geom << true << true << QVariant();
```

resources/function_help/json/rotate
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "rotate",
+  "type": "function",
+  "description": "Returns a rotated version of a geometry. Calculations are in the Spatial Reference System of this geometry.",
+  "arguments": [ {"arg":"geom","description":"a geometry"},
+                 {"arg":"rotation","description":"clockwise rotation in degrees"},
```
nyalldawson: Generally we try to follow to approach used by the correponsing postgis function. In this case that would be st_rotate, which uses "radians counter-clockwise about the origin point".

nirvn:
Nice one. Would it make sense for the 3rd argument to be optional and have a default value being used when not provided? 

The default value could be the top left coordinate of the geometry's bounding box.

nyalldawson:
> The default value could be the top left coordinate of the geometry's bounding box.

I'd suggest the centroid or centroid of bounding box

nirvn:
+1 to centroid of bounding box.

tests/src/core/testqgsexpression.cpp
raymondnijssen: That is nice, and actually the way I tested the expressions manually. So it would be 

> +1 to centroid of bounding box.

Will try to make it optional. The centroid of the bbox is the same as the centroid of the geometry, right?

tests/src/core/testqgsexpression.cpp
raymondnijssen: Tnx! I tried but still get an error cause the test seems to expect a geometry:

QDEBUG : TestQgsExpression::eval_geometry_method() expected data of type 'QgsGeometry', got 'QVariant' for element 4 of data with tag 'rotate'
QFATAL : TestQgsExpression::eval_geometry_method() ASSERT: "false" in file qtestdata.cpp, line 92
FAIL!  : TestQgsExpression::eval_geometry_method() Received a fatal error.

raymondnijssen:
> > The default value could be the top left coordinate of the geometry's bounding box.
> 
> I'd suggest the centroid or centroid of bounding box

Will try to make it optional. The centroid of the bbox is the same as the centroid of the geometry, right?

tests/src/core/testqgsexpression.cpp
m-kuhn: Try QgsGeometry() (an invalid geometry) instead of a QVariant() in this case

resources/function_help/json/rotate
raymondnijssen: -1
Personally I'm in favour of using degrees instead of radians. As a GIS guy I'm expect directions and angles to be in degrees and clockwise with 0 = north. I also think most users do, and their data is likely to contain values in degrees.

Also, QgsGeometry().rotate() expects values in degrees and I would prefer being consistent  within QGIS than following the approach used in other software (postgis). And also the angle_at_vertex() expression function returns values in degrees

lbartoletti: It's a recurring debate. One day, I hope that we can add the option to choose the unit (degree, rad, grad) and the orientation (clockwise, counter, geographic).

When I develop, I am always embarrassed to use the geographical north as point 0.
But indeed, GIS users (especially QGIS?), have this habit.

So +0.5 with @raymondnijssen

lbartoletti:
Why the centroid of the bbox instead of the centroid of the geometry?

resources/function_help/json/rotate
m-kuhn: No strong preference, I can live with both.
Just pretty please let's keep expressions portable and not add an option. I remember a spreadsheet application that chose to go for translated function names depending on the system language ...

DelazJ:
```suggestion
                 {"arg":"rotation","description":"clockwise rotation from north in degrees"},
```
? though not the exact wording in the angle_at_vertex description.

raymondnijssen: @DelazJ The "from north" would not mean anything here, since the angle defines a rotation (= relative) and not a direction (= absolute).

roya0045:
> Why the centroid of the bbox instead of the centroid of the geometry?

Lets say you have a crescent shaped geometry, you would have distinct results.

resources/function_help/json/rotate
nyalldawson: Ok, fair enough!

raymondnijssen:
I didn't forget about this PR but I got stuck in making the 3rd argument optional.. :(

nyalldawson:
@raymondnijssen I didn't want this great work to miss 3.12, so I've finished off this pr at #33575!

PR #33132, opened 2019-11-28
PG provider connection API: convert to bools

### LINKED ISSUE #33130: QgsPostgresProviderConnection::executeSqlPrivate doesn't  properly handle boolean values
It looks to me that QgsPostgresProviderConnection::executeSqlPrivate doesn't honour boolean values correctly.

I'm doing this, in python:
> res = conn.executeSql("SELECT FALSE")

And the returned value looks like this:
> [[True]]

\cc @elpaso

elpaso:
Fixes #33130


strk:
Thanks @elpaso that was very fast! :)
I guess now I could tweak the testcase included in #33118 NOT to run pointcloud/topology tests when pointcloud/topology is not available on the system !

strk:
would this be ok to backport, btw ?

elpaso:
> would this be ok to backport, btw ?

Yes, absolutely.

PR #33136, opened 2019-11-28
When dropping a picture to a layout, place the center of the picture item at the cursor drop location

nyalldawson:
...instead of the top left

We want the drop location to be the center of the placed item,
because drag thumbnails are usually centered on the mouse cursor


nirvn (approved):
Makes sense. I think that also matches inkscape's behavior.

nyalldawson:
@nirvn it certainly feels much more natural!

PR #33142, opened 2019-11-29
Show "Open Document..." action when right clicking certain

nyalldawson:
files in the browser, allowing them to be opened with the default
application for that file type

E.g. PDFs will open with the default PDF viewer.

Works with PDF, ODS, XLS(X), CSV, TXT, PNG, JPEG, TIFF, SVG
(other types will likely need more work, since they aren't
currently shown in the browser)

Peek 2019-11-29 12-55


nirvn (approved):

NathanW2:
Nice! That is a handy shortcut.  Unsure about the naming here though "Open"
seems to read like we will open in QGIS maybe "Open PNG image in
external..."

nyalldawson:
> maybe "Open PNG image in external..."

Fixed

PR #33151, opened 2019-11-29
log message when adding feature fails

3nids:
due to invalid number of fields (when using joins, provider fields != layer fields)


Gustry (approved):
Not logging in a specific tab in the log panel?

Thanks, this is needed. I often got `False` when adding feature without having any kind of error message I think.

PR #33161, opened 2019-11-29
Respect placement quadrant in DXF export

m-kuhn:
If quadrant based placement is activated for labeling, this was up to now ignored on dxf export, leading to misplaced labels.
With this PR, the dxf alignment is set based on the quadrant.

PAL changes the anchor point for quadrant placed labels during candidate creation. In order to "undo" this change, the delta is now preserved in the PAL label.


nyalldawson (approved):
Can't comment on the dxf specifics, but the code looks good

PR #33162, opened 2019-11-29
DXF export preserve dashed line style

m-kuhn:
We do not (yet) support an option to guarantee stroked parts at corners.
So also do not set this flag in the DXF export.


nyalldawson (approved):
Again, can't comment on the dxf specifics, but code looks good

lbartoletti (approved):

PR #33165, opened 2019-11-30
[FEATURE] Static particle traces for rendering mesh vector dataset

vcloarec:
This PR permits to display directly in QGIS static particle traces for vector datasets in mesh layer without any plugin.
The user can choose in the mesh layer properties window :
- the color
- the size of the traces (line width)
- the count of particles
- the maximum length of the particle's tail

Screenshot_20191129_200625


nirvn:
Can you add a screenshot of the UI elements introduced in this PR?

vcloarec:
> Can you add a screenshot of the UI elements introduced in this PR?

Screenshot_20191129_225924

saberraz:
That looks great. A couple of issues I am experiencing not sure if it is related to this PR:

1. I have lost the tails from arrows:
image

1. When I use Style Panel (@vcloarec layer properties to change style is so QGIS 1.x :) ), changing Particle count does not take effect and I have to click Apply:
Peek 2019-11-30 10-06

nirvn:
Thanks for the screenshot. I think you'll have to change the maximum tail length unit to be dpi dependent, ie mm. You could hard code millimeters (I think/hope that's what's used in other parts of the mesh rendering UI), or Do The Right Thing (TM) and use the unit widget to unlock all supported unit types.

saberraz:
I agree, using unit widget is a much better option. Currently, I often end up with unexpected results in the print/animation outputs.

PeterPetrik:
can you choose different frame?

vcloarec:
>    layer properties to change style is so QGIS 1.x :)

@saberraz , I have always been a old fashion guy!
 
Changing particles count with Style panel is now fixed.

Loosing arrow tails maybe come from the line width parameter that is shared for all the three symbology. When width line is too big for arrow, the head hides the tail. As the particle traces may needs a bigger line width, when returning to arrow symbology, the heads of arrow could mask the tail.

I will look at the unit widget.

vcloarec:
> can you choose different frame?

@PeterPetrik 
Frames are generated with random particles for the current vector dataset and the frame is updated on each map refreshing. What do you mean by choosing different frame ?

saberraz:
Brilliant, the last update looking good and I can confirm both issues are now resolved.

PeterPetrik (approved):
@vcloarec all good.

i have just one question; The filter by magnitude and on user grid options that are present on arrow and streamlines renderers aren't applicable here?

PeterPetrik:
if no objections, would like to merge tomorrow.

vcloarec:
> i have just one question; The filter by magnitude and on user grid options that are present on arrow and streamlines renderers aren't applicable here?

As particle traces represents a traveled distance and directly a magnitude, it is difficult to apply filter magnitude.
On mesh or on user grid option, may can be applicable, but not implemented now...

nirvn:
Thanks for implementing the unit widget here. As a follow up PR, the mesh symbology would benefit from using that widget on all size/length settings.

PR #33183, opened 2019-12-02
Fix confusing/broken label obstacle avoidance handling

nyalldawson:
When a project is running in label engine v2 mode, discard any candidates where they collide with an obstacle feature of greater weight when compared to the label's priority

Previously, obstacle weight was used ONLY to rank a features' label candidates relative to each other, but was never used to actually prune candidates completely. This meant that the labeling obstacle functionality was confusing and frustrating for users to work with -- because despite setting layers as the maximum possible blocking weight, you'd still see labels being placed over these features (e.g. where the labeling engine had no other choice).

Now, (when a project is set to v2 labeling engine mode), labels will NEVER be placed over obstacles of greater weight. This means that labels will potentially be omitted if the only choice is to place
them over a high weighting obstacle. But ultimately, that's much more understandable for users -- they've manually set a particular layer to a high obstacle factor, so we should respect that and
never place labels on these features.

In the end, this change makes the labeling placement much simpler to understand for users, and should give power users a much nicer experience all round.

Because this can dramatically alter the placement of labels in an existing project, the new logic is only active when a project is set to the labeling engine v2 mode (via the project's label settings dialog). This is currently opt-in even for new projects, while we gather feedback from users and potentially further refine the logic. Eventually it will be made the default for new projects (QGIS >= 3.12 only)

Funded by the QGIS grants program


nirvn:
@nyalldawson , testing this out, few comments:
- since labels can now be removed and not only discourage, IMHO the obstacle slider widget should have a 0 to 10 range to match that of the label priority slider. It's much easier to get a sense of whether you're matching, above, or below the priority weight;
- we likely need to refine the default obstacle type / behavior for polygon layers. ATM, it's using the feature interior as an obstacle, which means that as soon as the obstacle weight > than any label priority, it kills those labels from the canvas. It's a bit too harsh a default behavior
- this PR raised an interesting issue with obstacles. ATM, if you have a scale-dependent visibility settings for labels, it'll also make obstacles scale-dependent. I'm wondering if this should change. For e.g., I have a road layer which I'd like to have _always_ as obstacle, but have labels only show at a given scale range.

nyalldawson:
> since labels can now be removed and not only discourage, IMHO the obstacle slider widget should have a 0 to 10 range to match that of the label priority slider. It's much easier to get a sense of whether you're matching, above, or below the priority weight;

Do you mean change it to a "int" style slider instead of the current continuous one?

> we likely need to refine the default obstacle type / behavior for polygon layers. ATM, it's using the feature interior as an obstacle, which means that as soon as the obstacle weight > than any label priority, it kills those labels from the canvas. It's a bit too harsh a default behavior

Ok, let's address in a follow up

> this PR raised an interesting issue with obstacles. ATM, if you have a scale-dependent visibility settings for labels, it'll also make obstacles scale-dependent. I'm wondering if this should change. For e.g., I have a road layer which I'd like to have always as obstacle, but have labels only show at a given scale range.

Hm, you could work around that somewhat using rule based labeling (and using a zero width '' label when you only want obstacles)

nirvn:
@nyalldawson , re slider, we need to change the QSlider maximum value to 10 instead of 100.

Here's the label priority slider: https://github.com/qgis/QGIS/blob/master/src/ui/qgstextformatwidgetbase.ui#L5838

Here's the obstacle weight slider: https://github.com/qgis/QGIS/blob/master/src/ui/qgstextformatwidgetbase.ui#L6709

nirvn:
> Ok, let's address in a follow up

Sounds reasonable.

> Hm, you could work around that somewhat using rule based labeling (and using a zero width '' label when you only want obstacles)

That's a good hack. Still wondering whether it'd be ultimately better behavior to have drawn features always be obstacles (since the absence of labels != absence of features).

nyalldawson:
>  Still wondering whether it'd be ultimately better behavior to have drawn features always be obstacles (since the absence of labels != absence of features).

It's rather expensive to calculate though -- that'd definitely hurt rendering times

nirvn:
@nyalldawson , code looking good, feel free to merge.

nyalldawson:
Thanks @nirvn!

PR #33192, opened 2019-12-02
spatialjoin/ join attributes by location c++ port

roya0045:
C++ Port of spatial join

## Description
An attempt at a c++ port of spatial join.

 Using Blockingfiltered doesn't seem to be possible at the moment.

Still looking for a way to implement good multithread.

Suggestions are welcome.
## Checklist


roya0045:
@elpaso I have tried to do multiprocess but to do so I would need multiple inputs or find a way to use multiprocessing within a class, which is not possible with qt.

Help would be appreciated. I thought about adding QReadWriteLock to share the geometry engine across thread but I'm not too sure how.

nyalldawson:
Nice work! Can I suggest you start with a direct 1:1 port of the python alg to C++, and then add the multiprocess support as a secondary step? It'll make reviewing and merging much quicker

roya0045:
> Nice work! Can I suggest you start with a direct 1:1 port of the python alg to C++, and then add the multiprocess support as a secondary step? It'll make reviewing and merging much quicker

Other than the spatial indexing step, its nearly a 1:1, code, it might be set up slightly differently but the behavior is the same as no parallel processing is implemented yet, but the current structure should help with parallel processing.

Speaking of reviewing, any ETA for the ones at the bottom of the barrel? I know they are just extra fancy features but I feel like many other  extra fancy features have gotten through in the meantime. :/

nyalldawson:
> Other than the spatial indexing step, its nearly a 1:1, code, it might be set up slightly differently but the behavior is the same as no parallel processing is implemented yet, but the current structure should help with parallel processing.

The index should not be required here -- the alg was written to take advantage of spatial indexes on the provider backend.

That's why we need a 1:1 port to start with. Then we can discuss the improvements on an atomic basis. At the moment, this can't be merged due to the changes and the multiprocessing code, so it's an unfortunate stalemate situation...

roya0045:
> That's why we need a 1:1 port to start with. Then we can discuss the improvements on an atomic basis. At the moment, this can't be merged due to the changes and the multiprocessing code, so it's an unfortunate stalemate situation...

Removed multiproc and indexing.

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+
+void QgsSpatialJoinAlgorithm::initAlgorithm( const QVariantMap & )
+{
+  addParameter( new QgsProcessingParameterFeatureSource( QStringLiteral( "INPUT" ),
+                QObject::tr( "Base Layer" ), QList< int > () << QgsProcessing::QgsProcessing::TypeVectorAnyGeometry ) );
+  addParameter( new QgsProcessingParameterFeatureSource( QStringLiteral( "JOIN" ),
+                QObject::tr( "Join Layer" ), QList< int > () << QgsProcessing::QgsProcessing::TypeVectorAnyGeometry ) );
+  addParameter( new QgsProcessingParameterEnum( QStringLiteral( "PREDICATE" ), QObject::tr("Geometric predicate"),
+                tags(), true, 0 ) );
```
nyalldawson:
```suggestion
                QStringList() << QObject::tr("intersects") << .... etc, true, 0 ) );
```

nyalldawson: This parameter needs the widget_wrapper metadata set from the original python implementation

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+  addParameter( new QgsProcessingParameterField( QStringLiteral( "FIELDS_TO_JOIN" ),
+                QObject::tr( "Fields to join to the base layer (leave empty to copy all fields)" ),
```
nyalldawson:
```suggestion
                QObject::tr( "Fields to add (leave empty to use all fields)" ),
```

(otherwise we'd need to replicate this wording change in other places too -- do that in a separate PR)

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+void QgsSpatialJoinAlgorithm::initAlgorithm( const QVariantMap & )
+{
+  addParameter( new QgsProcessingParameterFeatureSource( QStringLiteral( "INPUT" ),
+                QObject::tr( "Base Layer" ), QList< int > () << QgsProcessing::QgsProcessing::TypeVectorAnyGeometry ) );
+  addParameter( new QgsProcessingParameterFeatureSource( QStringLiteral( "JOIN" ),
+                QObject::tr( "Join Layer" ), QList< int > () << QgsProcessing::QgsProcessing::TypeVectorAnyGeometry ) );
+  addParameter( new QgsProcessingParameterEnum( QStringLiteral( "PREDICATE" ), QObject::tr("Geometric predicate"),
+                tags(), true, 0 ) );
+  addParameter( new QgsProcessingParameterField( QStringLiteral( "FIELDS_TO_JOIN" ),
```
nyalldawson:
```suggestion
  addParameter( new QgsProcessingParameterField( QStringLiteral( "JOIN_FIELDS" ),
```

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+                QObject::tr( "Fields to join to the base layer (leave empty to copy all fields)" ),
+                QVariant(), QStringLiteral( "FIELDS" ), QgsProcessingParameterField::Any, true, true ) );
```
nyalldawson:
```suggestion
                QVariant(), QStringLiteral( "JOIN" ), QgsProcessingParameterField::Any, true, true ) );
```

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+  QStringList joinMethods;
+  joinMethods << QObject::tr( "Create separate feature for each matching feature (one-to-many)" )
+          << QObject::tr( "Take attributes of the first matching feature only (one-to-one)" );
+  addParameter( new QgsProcessingParameterEnum( QStringLiteral( "METHOD" ),
+                QObject::tr( "Join type" ),
+                joinMethods, false, 1 ) );
```
nyalldawson:
```suggestion
                joinMethods, false, 0 ) );
```

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+  addParameter( new QgsProcessingParameterBoolean( QStringLiteral( "DISCARD_NONMATCHING" ),
+                QObject::tr( "Discard records which could not be joined" ),
+                false ) );
+  addParameter( new QgsProcessingParameterString( QStringLiteral( "PREFIX" ),
+                QObject::tr( "Joined field prefix" ), QVariant(), false, true ) );
+  addParameter( new QgsProcessingParameterFeatureSink( QStringLiteral( "JOINED" ), QObject::tr( "Joined layer" ), QgsProcessing::TypeVectorAnyGeometry, QVariant(), true, true ) );
```
nyalldawson:
```suggestion
  addParameter( new QgsProcessingParameterFeatureSink( QStringLiteral( "OUTPUT" ), QObject::tr( "Joined layer" ), QgsProcessing::TypeVectorAnyGeometry, QVariant(), true, true ) );
```

We **can't** change any of these names -- they are locked in stable API contract

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+  addParameter( new QgsProcessingParameterFeatureSink( QStringLiteral( "UNJOINED" ), QObject::tr( "unjoined layer" ), QgsProcessing::TypeVectorAnyGeometry, QVariant(), true, true ) );
```
nyalldawson:
```suggestion
  addParameter( new QgsProcessingParameterFeatureSink( QStringLiteral( "NON_MATCHING" ), QObject::tr( "Unjoinable features from first layer" ), QgsProcessing::TypeVectorAnyGeometry, QVariant(), true, false ) );
```

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+}
```
nyalldawson: missing the JOINED_COUNT output here

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+QString QgsSpatialJoinAlgorithm::name() const
+{
+  return QStringLiteral( "spatialjoin" );
```
nyalldawson:
```suggestion
  return QStringLiteral( "joinattributesbylocation" );
```

(locked in stable API)

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+}
+
+QString QgsSpatialJoinAlgorithm::displayName() const
+{
+  return QObject::tr( "Join attributes by their spatial relation" );
```
nyalldawson:
```suggestion
  return QObject::tr( "Join attributes by location" );
```

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+}
+
+QStringList QgsSpatialJoinAlgorithm::tags() const
+{
+  return QObject::tr( "intersect,contain,isEqual,touch,overlap,within,cross" ).split( ',' );
```
nyalldawson: Why drop off the other existing tags?

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+}
+
+QString QgsSpatialJoinAlgorithm::svgIconPath() const
+{
+  return QgsApplication::iconPath( QStringLiteral( "/algorithms/... .svg" ) );
```
nyalldawson: Fix

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+}
+
+QIcon QgsSpatialJoinAlgorithm::icon() const
+{
+  return QgsApplication::getThemeIcon( QStringLiteral( "/algorithms/... .svg" ) );
```
nyalldawson: Fix

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+}
+
+QString QgsSpatialJoinAlgorithm::group() const
+{
+  return QObject::tr( "Vector analysis" );
```
nyalldawson:
```suggestion
  return QObject::tr( "Vector general" );
```

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+}
+
+QString QgsSpatialJoinAlgorithm::groupId() const
+{
+  return QStringLiteral( "vectoranalysis" );
```
nyalldawson:
```suggestion
  return QStringLiteral( "vectorgeneral" );
```

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+{
+  return QObject::tr( "Join attributes from one vector layer to another by location." );
+}
+
+QgsSpatialJoinAlgorithm *QgsSpatialJoinAlgorithm::createInstance() const
+{
+  return new QgsSpatialJoinAlgorithm();
+}
+
+QList<int> QgsSpatialJoinAlgorithm::inputLayerTypes() const
```
nyalldawson: not needed

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+{
+  return QList< int >() << QgsProcessing::TypeVectorAnyGeometry;
+}
+
+QgsProcessing::SourceType QgsSpatialJoinAlgorithm::outputLayerType() const
```
nyalldawson: not needed

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+{
+  return QgsProcessing::TypeVectorAnyGeometry;
+}
+
+QgsCoordinateReferenceSystem QgsSpatialJoinAlgorithm::outputCrs( const QgsCoordinateReferenceSystem &inputCrs ) const
```
nyalldawson: not needed

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+{
+  mCrs = inputCrs;
+  return mCrs;
+}
+
+QString QgsSpatialJoinAlgorithm::inputParameterName() const
```
nyalldawson: not needed

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+{
+  return QStringLiteral( "Base layer" );
+}
+
+QString QgsSpatialJoinAlgorithm::inputParameterDescription() const
```
nyalldawson: not needed

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+{
+  return QObject::tr( "Base layer to join attribute to." );
+}
+
+QString QgsSpatialJoinAlgorithm::outputName() const
```
nyalldawson: not needed

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+{
+  return QObject::tr( "Joined layer" );
+}
+
+bool QgsSpatialJoinAlgorithm::prepareAlgorithm( const QVariantMap &parameters, QgsProcessingContext &context, QgsProcessingFeedback *feedback )
+{
+
+  std::unique_ptr<QgsProcessingFeatureSource> mBaseSource( parameterAsSource( parameters, QStringLiteral( "INPUT" ), context ) );
```
nyalldawson: The definition of mBaseSource and mJoinSource needs to be in the header, not here (otherwise they'll go out of scope as soon as this method exits) 

(did this even build?)

src/analysis/processing/qgsalgorithmspatialjoin.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+  return true;
+}
+
+
+QVariantMap QgsSpatialJoinAlgorithm::processAlgorithm( const QVariantMap &parameters, QgsProcessingContext &context, QgsProcessingFeedback *feedback )
+{
+
+  // make spatial index
+  QgsFeatureIterator joinIter = mJoinSource->getFeatures( QgsFeatureRequest().setDestinationCrs( mBaseSource->sourceCrs(), context.transformContext() ).setSubsetOfAttributes( fields2Fetch ) );
```
nyalldawson: Missing many parts from the original python algorithm -- please restore/port these too

src/analysis/processing/qgsalgorithmspatialjoin.cpp
roya0045: Yes it did build but I edited the code in two distinct environment and that made some weird partial changes to the code.

src/analysis/processing/qgsalgorithmspatialjoin.cpp
roya0045: As I use them as options and the original implementation only had those choices.

Otherwise if this must not be changed. I will simply make another list on the side just like the method.

src/analysis/processing/qgsalgorithmspatialjoin.cpp
roya0045: They are declared in the header also.

src/analysis/processing/qgsalgorithmspatialjoin.cpp
roya0045: Should be done.

roya0045:
@nyalldawson thanks for reviewing the preliminary code.

src/analysis/processing/qgsalgorithmspatialjoin.cpp
nyalldawson: Was that pushed yet?

src/analysis/processing/qgsalgorithmspatialjoin.cpp
nyalldawson: Right, but by re-declaring them here, you're using the local variables only -- as soon as this function exits they'll be deleted.

src/analysis/processing/qgsalgorithmspatialjoin.cpp
roya0045: Were you referring to the counting function? That should be in now.

src/analysis/processing/qgsalgorithmspatialjoin.cpp
roya0045: Yes, see line 37.

roya0045:
@nyalldawson in my test I have no clue why but the base output is labeled as NON_MATCHING, just like the non matching output. I have no clue why this happens. And the output is not automatically added to the map but the joined output pointer is ok. I'll try to debug more but this is strange.

src/analysis/processing/qgsalgorithmspatialjoin.cpp
roya0045: I forgot that I redeclared many things and forgot to fix them, this lead to many issues, should be fixed now though. thanks for reminding me!

roya0045:
@alexbruy a check from you would be appreciated to validate if I didn't miss anything or set up something badly.

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,368 @@
+/***************************************************************************
+                         qgsalgorithmjoinbylocation.cpp
+                         ---------------------
+    begin                : November 2019
+    copyright            : (C) 2019 by
```
nyalldawson: You can fill this in with your details and take credit for your work :smile:

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,368 @@
...
+{
+  addParameter( new QgsProcessingParameterFeatureSource( QStringLiteral( "INPUT" ),
+                QObject::tr( "Base Layer" ), QList< int > () << QgsProcessing::QgsProcessing::TypeVectorAnyGeometry ) );
+  addParameter( new QgsProcessingParameterFeatureSource( QStringLiteral( "JOIN" ),
+                QObject::tr( "Join Layer" ), QList< int > () << QgsProcessing::QgsProcessing::TypeVectorAnyGeometry ) );
+
+  QStringList predicates;
+  predicates << QObject::tr( "intersects" )
+             << QObject::tr( "contains" )
+             << QObject::tr( "isEqual" )
```
nyalldawson:
```suggestion
             << QObject::tr( "equals" )
```

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,368 @@
...
+             << QObject::tr( "touches" )
+             << QObject::tr( "overlaps" )
+             << QObject::tr( "within" )
+             << QObject::tr( "crosses" );
+
+  addParameter( new QgsProcessingParameterEnum( QStringLiteral( "PREDICATE" ), QObject::tr( "Geometric predicate" ),
```
nyalldawson:
```suggestion
  std::unique_ptr< QgsProcessingParameterEnum > predicateParam = qgis::make_unique< QgsProcessingParameterEnum >( QStringLiteral( "PREDICATE" ), QObject::tr( "Geometric predicate" ), predicates, true, 0 );
  QVariantMap predicateMetadata;
  QVariantMap widgetMetadata;
  widgetMetadata.insert( QStringLiteral( "useCheckBoxes" ), true );
  widgetMetadata.insert( QStringLiteral( "columns" ), 2 );
  predicateMetadata.insert( QStringLiteral( "widget_wrapper" ), widgetMetadata );
  predicateParam->setMetadata( predicateMetadata );
  addParameter( predicateParam.release() );
```
( yep, much more concise in the original Python version!!)

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+  if ( !prefix.isEmpty() )
+  {
+    for ( int i = 0; i < outFields2.count(); ++i )
+    {
+      outFields2[ i ].setName( prefix + outFields2[ i ].name() );
+    }
+  }
+
+  if ( feedback->isCanceled() )
```
nyalldawson: You can drop this -- the prepareAlgorithm step is all cheap code, so we don't need to support cancelation at this stage

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,368 @@
...
+    return false;
+
+  for ( int i = 0; i < fields2Indices.count(); ++i )
+    emptyAttrs << QVariant();
+
+  fields2Fetch = fields2Indices;
+  fields2Fetch << fields2Indices;
```
nyalldawson: This looks wrong -- it's duplicating the list

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+  mDiscardNonMatching = parameterAsBoolean( parameters, QStringLiteral( "DISCARD_NONMATCHING" ), context );
+
+  if ( !mDiscardNonMatching )
+    {
+    unjoinedFeatures.reset( parameterAsSink( parameters, QStringLiteral( "NON_MATCHING" ), context, unjoinedFeaturesId, mBaseSource->fields(),
+                            mBaseSource->wkbType(), mBaseSource->sourceCrs(), QgsFeatureSink::RegeneratePrimaryKey ) );
+    if ( parameters.value( QStringLiteral( "NON_MATCHING" ) ).isValid() && !unjoinedFeatures )
+      throw QgsProcessingException( invalidSinkError( parameters, QStringLiteral( "NON_MATCHING" ) ) );
+    }
```
nyalldawson: This code can be moved to processAlgorithm, it's all safe to do in a background thread

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,368 @@
...
+QVariantMap QgsJoinByLocationAlgorithm::processAlgorithm( const QVariantMap &parameters, QgsProcessingContext &context, QgsProcessingFeedback *feedback )
+{
+
+  Q_UNUSED( parameters );
+  int joinedCount = 0;
+  QgsFeatureIterator joinIter = mJoinSource->getFeatures( QgsFeatureRequest().setDestinationCrs( mBaseSource->sourceCrs(), context.transformContext() ).setSubsetOfAttributes( fields2Fetch ) );
+  static QHash< QgsFeatureId, QgsAttributes > input2AttributeCache;
+  QgsFeature f;
+  double step = mJoinSource->featureCount() > 0 ? 50.0 / mJoinSource->featureCount() : 1;
+  int i = 0;
```
nyalldawson: From here the c++ differs from the original python version -- can you adapt the python version as is for now?

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
roya0045: I know, I'm still on the fence about what to write.

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
roya0045: Yes I took this from another implementation, join by nearest or attribute if I recall correctly.

roya0045:
Getting something weird, when I open a project and do the join, columns are ok, but when I do it a second time, the columns are the same columns as the first join even if I use distinct layers/prefix.

roya0045:
Ok static was the culprit, I thought it would be removed with the instance but somehow it remained.

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
roya0045: Should be done

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
roya0045: This loop was removed, remaining should be the same in function.

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,340 @@
...
+    if ( parameters.value( QStringLiteral( "NON_MATCHING" ) ).isValid() && !unjoinedFeatures )
+      throw QgsProcessingException( invalidSinkError( parameters, QStringLiteral( "NON_MATCHING" ) ) );
+    }
+
+
+  Q_UNUSED( parameters );
+  int joinedCount = 0;
+  QgsFeatureIterator joinIter = mJoinSource->getFeatures( QgsFeatureRequest().setDestinationCrs( mBaseSource->sourceCrs(), context.transformContext() ).setSubsetOfAttributes( fields2Fetch ) );
+  QgsFeature f;
+
```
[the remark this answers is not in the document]
nyalldawson: (sorry, this is only a partial review -- i'm just posting bits as I'm finding time to digest this!)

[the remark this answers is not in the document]
roya0045: No worries thanks for taking the time to review! (Though I would like to have my old PRs reviewed beforehand :( )

roya0045:
@nyalldawson you were right, I mixed up the input first and join first method, hopefully my latest push will have solved it, seemed to work for the tests I did.

roya0045:
Ok somehow, the data in the column floatval becomes 'Invalid', I'm not sure if this is due to the fact that I get a feature with an invalid value and its simply the iteration order that is causing this issue or the column data itself [(see here for the debug)](https://travis-ci.org/qgis/QGIS/jobs/626841482#L1949) @nyalldawson any clue?

Otherwise locally it works fine. :/

roya0045:
@nyalldawson ok fixed the lasts issues (I hope). @alexbruy does this looks fine to you?

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,350 @@
...
+  mJoinedFeatures.reset( parameterAsSink( parameters, QStringLiteral( "OUTPUT" ), context, mJoinedFeaturesId, mOutFields,
+                                          mBaseSource->wkbType(), mBaseSource->sourceCrs(), QgsFeatureSink::RegeneratePrimaryKey ) );
+
+  if ( parameters.value( QStringLiteral( "OUTPUT" ) ).isValid() && !mJoinedFeatures )
+    throw QgsProcessingException( invalidSinkError( parameters, QStringLiteral( "OUTPUT" ) ) );
+
+  mDiscardNonMatching = parameterAsBoolean( parameters, QStringLiteral( "DISCARD_NONMATCHING" ), context );
+
+  if ( !mDiscardNonMatching )
```
nyalldawson: These two options are slightly different but independant -- you'll need to remove the above check and create the non_matching sink regardless of the value of DISCARD_NONMATCHING

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,350 @@
...
+  {
+    mUnjoinedFeatures.reset( parameterAsSink( parameters, QStringLiteral( "NON_MATCHING" ), context, mUnjoinedFeaturesId, mBaseSource->fields(),
+                             mBaseSource->wkbType(), mBaseSource->sourceCrs(), QgsFeatureSink::RegeneratePrimaryKey ) );
+    if ( parameters.value( QStringLiteral( "NON_MATCHING" ) ).isValid() && !mUnjoinedFeatures )
+      throw QgsProcessingException( invalidSinkError( parameters, QStringLiteral( "NON_MATCHING" ) ) );
+  }
+
+  mUnjoinedIds = mBaseSource->allFeatureIds();
+
+  Q_UNUSED( parameters );
```
nyalldawson: not required

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,350 @@
...
+  int joinedCount = 0;
```
nyalldawson:
```suggestion
  long joinedCount = 0;
```

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,350 @@
...
+  QgsFeatureIterator joinIter = mJoinSource->getFeatures( QgsFeatureRequest().setDestinationCrs( mBaseSource->sourceCrs(), context.transformContext() ) );
```
nyalldawson:
```suggestion
  QgsFeatureIterator joinIter = mJoinSource->getFeatures( QgsFeatureRequest().setDestinationCrs( mBaseSource->sourceCrs(), context.transformContext() ).setSubsetOfAttributes(mFields2Indices) );
```

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,350 @@
...
+  QgsFeature f;
+
+  // Create output vector layer with additional attributes
+  int step = mBaseSource->featureCount() > 0 ? 100.0 / mBaseSource->featureCount() : 1;
```
nyalldawson:
```suggestion
  const double step = mBaseSource->featureCount() > 0 ? 100.0 / mBaseSource->featureCount() : 1;
```

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,350 @@
...
+  QgsFeatureIterator joinIterator = mJoinSource->getFeatures();
```
nyalldawson: remove this line -- you've already got the iterator a few lines above

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,350 @@
...
+  int i = 0;
```
nyalldawson:
```suggestion
  long i = 0;
```

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,350 @@
...
+  return ( ok );
+}
+
+bool QgsJoinByLocationAlgorithm::processFeatures( QgsFeature &joinFeature, QgsProcessingContext &context, QgsProcessingFeedback *feedback )
+{
+
+  if ( !joinFeature.hasGeometry() )
+    return false;
+  const QgsGeometry featGeom = joinFeature.geometry();
+  QgsGeometryEngine *engine = QgsGeometry::createGeometryEngine( featGeom.constGet() );
```
nyalldawson:
```suggestion
  std::unique_ptr< QgsGeometryEngine > engine( QgsGeometry::createGeometryEngine( featGeom.constGet() ) );
```

but engine should be lazy constructed only on getting the first matching feature -- otherwise we are doing unnecessary work when in the case that there's no matching features based on the bounding box request alone

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,350 @@
...
+  engine->prepareGeometry();
+  QgsFeatureRequest req = QgsFeatureRequest().setFilterRect( featGeom.boundingBox() ).setDestinationCrs( mCrs, context.transformContext() );
```
nyalldawson:
```suggestion
  QgsFeatureRequest req = QgsFeatureRequest().setFilterRect( featGeom.boundingBox() );
```

not required here, featGeom will always be in the required crs at this stage

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,350 @@
...
+
+  if ( !mDiscardNonMatching )
+  {
+    mUnjoinedFeatures.reset( parameterAsSink( parameters, QStringLiteral( "NON_MATCHING" ), context, mUnjoinedFeaturesId, mBaseSource->fields(),
+                             mBaseSource->wkbType(), mBaseSource->sourceCrs(), QgsFeatureSink::RegeneratePrimaryKey ) );
+    if ( parameters.value( QStringLiteral( "NON_MATCHING" ) ).isValid() && !mUnjoinedFeatures )
+      throw QgsProcessingException( invalidSinkError( parameters, QStringLiteral( "NON_MATCHING" ) ) );
+  }
+
+  mUnjoinedIds = mBaseSource->allFeatureIds();
```
nyalldawson: This is sloooooow to calculate, that's why the python version only builds it if required. Can you copy the python logic?

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,350 @@
...
+        if ( engine->crosses( geom ) )
+        {
+          ok = true;
+        }
+        break;
+    }
+    if ( ok )
+      return ( ok );
+  }
+  return ( ok );
```
nyalldawson:
```suggestion
  return ok;
```

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,350 @@
...
+        break;
+      case 6:
+        if ( engine->crosses( geom ) )
+        {
+          ok = true;
+        }
+        break;
+    }
+    if ( ok )
+      return ( ok );
```
nyalldawson:
```suggestion
      return ok;
```

nyalldawson (commented):
Nearly there!

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
roya0045: I hate unique ptrs, they are so finnicky, I have no clue why I get the error that its deleted during the travis build.

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
roya0045: went for qlonglong, as long cannot be converted to qvariant.

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,345 @@
...
+    throw QgsProcessingException( invalidSinkError( parameters, QStringLiteral( "OUTPUT" ) ) );
+
+  mDiscardNonMatching = parameterAsBoolean( parameters, QStringLiteral( "DISCARD_NONMATCHING" ), context );
+
+  mUnjoinedFeatures.reset( parameterAsSink( parameters, QStringLiteral( "NON_MATCHING" ), context, mUnjoinedFeaturesId, mBaseSource->fields(),
+                           mBaseSource->wkbType(), mBaseSource->sourceCrs(), QgsFeatureSink::RegeneratePrimaryKey ) );
+  if ( parameters.value( QStringLiteral( "NON_MATCHING" ) ).isValid() && !mUnjoinedFeatures )
+    throw QgsProcessingException( invalidSinkError( parameters, QStringLiteral( "NON_MATCHING" ) ) );
+
+  if ( !mDiscardNonMatching && mUnjoinedFeatures )
```
nyalldawson:
```suggestion
  if ( !mDiscardNonMatching || mUnjoinedFeatures )
```

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,345 @@
...
+  {
+    if ( feedback->isCanceled() )
+      break;
+    feedback->setProgress( i * step );
+    if ( processFeatures( f, feedback ) )
+      joinedCount++;
+    i++;
+  }
+
+  if ( !mDiscardNonMatching )
```
nyalldawson:
```suggestion
  if ( !mDiscardNonMatching ||  mUnjoinedFeatures )
```

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,345 @@
...
+  {
+    QgsFeature f2;
+    QgsFeatureRequest remainings = QgsFeatureRequest().setFilterFids( mUnjoinedIds );
+    QgsFeatureIterator remainIter = mBaseSource->getFeatures( remainings );
+    while ( remainIter.nextFeature( f2 ) )
+    {
+      if ( mJoinedFeatures )
```
nyalldawson: add cancelation check

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,345 @@
...
+  return ok;
+}
+
+bool QgsJoinByLocationAlgorithm::processFeatures( QgsFeature &joinFeature, QgsProcessingFeedback *feedback )
+{
+
+  if ( !joinFeature.hasGeometry() )
+    return false;
+  const QgsGeometry featGeom = joinFeature.geometry();
+  std::unique_ptr< QgsGeometryEngine > engine = qgis::make_unique< QgsGeometryEngine >( QgsGeometry::createGeometryEngine( featGeom.constGet() ) );
```
nyalldawson:
```suggestion
  std::unique_ptr< QgsGeometryEngine > engine;
```

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,345 @@
...
+    joinAttributes.append( joinFeature.attribute( ix ) );
+  }
+
+  while ( it.nextFeature( baseFeature ) )
+  {
+    if ( feedback->isCanceled() )
+      break;
+    if ( mJoinMethod == 1 && !mUnjoinedIds.contains( baseFeature.id() ) )
+      continue;
+    if ( featureFilter( baseFeature, engine ) )
```
nyalldawson:
```suggestion
    if ( featureFilter( baseFeature, engine.get() ) )
```

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,345 @@
...
+  QgsFeatureRequest req = QgsFeatureRequest().setFilterRect( featGeom.boundingBox() );
+  QgsFeatureIterator it = mBaseSource->getFeatures( req );
+  QList<QgsFeature> filtered;
+  QgsFeature baseFeature;
+  bool ok = false;
+  QgsAttributes joinAttributes;
+  for ( int ix : qgis::as_const( mFields2Indices ) )
+  {
+    joinAttributes.append( joinFeature.attribute( ix ) );
+  }
```
nyalldawson: lazy evaluation (See below)

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,345 @@
...
+
+  while ( it.nextFeature( baseFeature ) )
+  {
+    if ( feedback->isCanceled() )
+      break;
+    if ( mJoinMethod == 1 && !mUnjoinedIds.contains( baseFeature.id() ) )
+      continue;
```
nyalldawson:
```suggestion
      continue;
      
    if ( !engine )
    {
      engine.reset( QgsGeometry::createGeometryEngine( featGeom.constGet() ) );
      engine->prepareGeometry();
      for ( int ix : qgis::as_const( mFields2Indices ) )
  {
    joinAttributes.append( joinFeature.attribute( ix ) );
  }
    }
```

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,345 @@
...
+  }
+  if ( mUnjoinedFeatures )
+  {
+    outputs.insert( QStringLiteral( "NON_MATCHING" ), mUnjoinedFeaturesId );
+  }
+  outputs.insert( QStringLiteral( "JOINED_COUNT" ), joinedCount );
+  return outputs;
+}
+
+bool QgsJoinByLocationAlgorithm::featureFilter( const QgsFeature &feature, std::unique_ptr< QgsGeometryEngine > engine ) const
```
nyalldawson:
```suggestion
bool QgsJoinByLocationAlgorithm::featureFilter( const QgsFeature &feature, QgsGeometryEngine* engine ) const
```

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,345 @@
...
+}
+
+bool QgsJoinByLocationAlgorithm::processFeatures( QgsFeature &joinFeature, QgsProcessingFeedback *feedback )
+{
+
+  if ( !joinFeature.hasGeometry() )
+    return false;
+  const QgsGeometry featGeom = joinFeature.geometry();
+  std::unique_ptr< QgsGeometryEngine > engine = qgis::make_unique< QgsGeometryEngine >( QgsGeometry::createGeometryEngine( featGeom.constGet() ) );
+  engine->prepareGeometry();
```
nyalldawson: move (See below)

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,345 @@
...
+    {
+      if ( mJoinedFeatures )
+      {
+        QgsFeature outputFeature( baseFeature );
+        outputFeature.setAttributes( baseFeature.attributes() + joinAttributes );
+        mJoinedFeatures->addFeature( outputFeature, QgsFeatureSink::FastInsert );
+      }
+      if ( !ok )
+        ok = true;
+      if ( !mDiscardNonMatching && mUnjoinedFeatures )
```
nyalldawson:
```suggestion
      if ( !mDiscardNonMatching || mUnjoinedFeatures )
```

roya0045: Are you sure about this? Since there is no more check to make the unjoined output I would assume that there  OR conditions should always be true as the unjoined sink is always created.

nyalldawson:
Superseded by #33476 33579

PR #33193, opened 2019-12-03
Add QgsEmfRenderer class for rendering EMF files to QPainter devices

nyalldawson:
(based on a libqemf backend, with upstreams at https://iondev.ro/qemf/ and https://github.com/KDE/calligra/tree/master/libs/vectorimage/libemf)

(Opened for comment only -- do not merge. If this is deemed fit for inclusion I'll create a fork of libqemf and collate some fixes floating around in various other forks first, and then modernise and cleanup that code)

This adds a new class QgsEmfRenderer which allows for rendering EMF picture files to a QPainter device. This is API only, added to give plugins and scripts the capability to read EMF files (e.g. to allow interoperability with windows clipboard and windows applications which heavily use this format).

I've searched widely but can't find any alternative to this approach -- while there's some pure python emf parsing libraries, they don't provide any hooks into qt's painter classes. So by the time those are added, I'd have ended up just rewriting libqemf anyway...!

### DIFF
--- a/.github/workflows/mxe.yml
+++ b/.github/workflows/mxe.yml
@@ -16,6 +16,6 @@ jobs:
     - name: Upload build
       uses: actions/upload-artifact@v1
       with:
-        name: Store artifact
+        name: QGIS for Windows 64bit
         path: qgis-mxe-release.zip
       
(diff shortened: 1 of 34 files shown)


nyalldawson:
Closing for now -- the results from libqemf have been underwhelming so far.

PR #33195, opened 2019-12-03
Default polygon obstacle mode is now Boundary avoidance 

nyalldawson:
See logic in #33183

### DIFF
--- a/src/core/qgspallabeling.cpp
+++ b/src/core/qgspallabeling.cpp
@@ -256,69 +256,10 @@ void QgsPalLayerSettings::initPropertyDefinitions()
 
 Q_NOWARN_DEPRECATED_PUSH // because of deprecated members
 QgsPalLayerSettings::QgsPalLayerSettings()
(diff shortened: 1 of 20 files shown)


nirvn (approved):

src/core/qgspallabeling.cpp
```diff
@@ -256,69 +256,10 @@ void QgsPalLayerSettings::initPropertyDefinitions()
...
-  leftDirectionSymbol = QStringLiteral( "<" );
-  rightDirectionSymbol = QStringLiteral( ">" );
-  reverseDirectionSymbol = false;
-  placeDirectionSymbol = SymbolLeftRight;
-  formatNumbers = false;
-  decimals = 3;
-  plusSign = false;
-
-  // placement
-  placement = AroundPoint;
```
nirvn: Nice cleanup of default value being moved to header.

nirvn (approved):
Nice! Thanks for addressing this so quickly after your last PR.

Issue #20630, opened 2015-03-31
Add random function with seed to expressions


qgib:
Author Name: **Olivier Dalang** (@olivierdalang)
Original Redmine Issue: [12482](https://issues.qgis.org/issues/12482)

Redmine category:expressions

---
Hi !

When using data defined values, it's hard to use the rand() function, because a new random number is generated all the time.

A symbol with this as polygon fill color:
@color_hsl(rand(0,360),100,50)@
will be changing color at each redraw, making the map look like a disco party (not saying it's not nice ;) ).

Why not add another random function where one could specify a seed, so that the random number is always the same. Typically, one would use the feature's ID as the seed.
@color_hsl(rand_seed($id,0,360),100,50)@
This way, the random color would always be the same, hopefully even across different setups.

Thanks !

Olivier

suricactus:
This was fixed and merged as part of 3.12 release. The issue should be closed.

PR #33204, opened 2019-12-03
[FEATURE][expressions] allow to seed random functions

olivierdalang:
## Description

This PR adds an optional seed parameter to `rand()` and `randf()` functions, as per issue #20630

This is very useful if you want the result to be deterministic, for instance to assign random but fixed colors to features. Using `color_hsb(rand(0,360,$id),50,50)` for instance yields always the same color for the same feature.

By the way, the PR also improves the `rand()` function, which didn't work for high values (over 32000) by using Qt's QRandomGenerator instead of qrand (which it seems was deprecated in Qt 5.11).

The PR is backwards compatible as the new parameter is optional, and the previous values were random anyway.

## Checklist

### DIFF
--- a/src/core/expression/qgsexpressionfunction.cpp
+++ b/src/core/expression/qgsexpressionfunction.cpp
@@ -368,22 +368,68 @@ static QVariant fcnRndF( const QVariantList &values, const QgsExpressionContext 
 {
   double min = QgsExpressionUtils::getDoubleValue( values.at( 0 ), parent );
   double max = QgsExpressionUtils::getDoubleValue( values.at( 1 ), parent );
+
   if ( max < min )
     return QVariant();
 
+  QRandomGenerator *generator;
+  if ( QgsExpressionUtils::isNull( values.at( 2 ) ) )
+  {
+    generator = QRandomGenerator::global();
+  }
+  else
+  {
+    quint32 seed;
+    if ( QgsExpressionUtils::isIntSafe( values.at( 2 ) ) )
+    {
+      // if seed can be converted to int, we use as is
+      seed = QgsExpressionUtils::getIntValue( values.at( 2 ), parent );
+    }
+    else
+    {
+      // if not, we hash string representation to int
+      QString seedStr = QgsExpressionUtils::getStringValue( values.at( 2 ), parent );
+      std::hash<std::string> hasher;
+      seed = hasher( seedStr.toStdString() );
+    }
+    generator = &QRandomGenerator::QRandomGenerator( seed );
+  }
+
   // Return a random double in the range [min, max] (inclusive)
-  double f = static_cast< double >( qrand() ) / RAND_MAX;
-  return QVariant( min + f * ( max - min ) );
+  return QVariant( min + generator->generateDouble() * ( max - min ) );
 }
 static QVariant fcnRnd( const QVariantList &values, const QgsExpressionContext *, QgsExpression *parent, const QgsExpressionNodeFunction * )
 {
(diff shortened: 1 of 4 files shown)


src/core/expression/qgsexpressionfunction.cpp
```diff
@@ -368,22 +368,68 @@ static QVariant fcnRndF( const QVariantList &values, const QgsExpressionContext
...
+  }
+
   // Return a random double in the range [min, max] (inclusive)
-  double f = static_cast< double >( qrand() ) / RAND_MAX;
-  return QVariant( min + f * ( max - min ) );
+  return QVariant( min + generator->generateDouble() * ( max - min ) );
 }
 static QVariant fcnRnd( const QVariantList &values, const QgsExpressionContext *, QgsExpression *parent, const QgsExpressionNodeFunction * )
 {
-  qlonglong min = QgsExpressionUtils::getIntValue( values.at( 0 ), parent );
```
m-kuhn: Is it not possible to keep that 64 bit?

olivierdalang:
Oops, just saw that QRandomGenerator was only added in Qt 5.10, so I guess this will have to wait anyways ? If so, is there a tag we can apply to freeze until we push Qt version ?

m-kuhn:
`#ifdef` it

nyalldawson:
I'd replace the qt generator with the std C++ ones

src/core/expression/qgsexpressionfunction.cpp
olivierdalang: Yes I made the change

olivierdalang:
Thanks for the review !

@nyalldawson 
> I'd replace the qt generator with the std C++ ones

I gave it another shot. Were you thinking of something like that ?
From what I read, these functions are available since C++ 11, is that available on all supported platform ? (sorry again for newbies question)

@m-kuhn 
> `#ifdef` it

Not sure what you mean ? Wouldn't we get portability issues if functions signature depends of the Qt version ? Anyway, if above commit works, that not necessary.

m-kuhn:
>  From what I read, these functions are available since C++ 11, is that available on all supported platform ?

Yes

> Not sure what you mean ? Wouldn't we get portability issues if functions signature depends of the Qt version ? Anyway, if above commit works, that not necessary.

I would have proposed to keep the signature with reduced functionality (and a warning), but even better if we can avoid the problem altogether :)

tests/src/core/testqgsexpression.cpp
```diff
@@ -2229,14 +2229,14 @@ class TestQgsExpression: public QObject
       // Two calls with the same seed always return the same number
       QgsExpression exp8( QStringLiteral( "randf(seed:=1)" ) );
       QVariant v8 = exp8.evaluate();
-      QCOMPARE( v8.toFloat() == 0.13387664401253274, true );
+      QCOMPARE( v8.toDouble() == 0.13387664401253274, true );
```
m-kuhn: QCOMPARE can be used to actually compare values, so in general it's better to do `QCOMPARE( val1, val2 )` than `QCOMPARE(val1==val2, true)`.
But since double comparison is special you are better off with `QVERIFY( qgsDoubleNear( v8.toDouble(), value ) )` in this particular case

olivierdalang: Thanks Matthias :-) I thought I'd never get the test pass...

olivierdalang:
@nyalldawson @m-kuhn Ok I think this is ready to be squashed&merged, let me know if it looks good.

nyalldawson:
this looks great to me, well done!

m-kuhn:
Good job @olivierdalang

PR #33212, opened 2019-12-03
Skip right-hand side of AND/OR expression if we already know result

wonder-sk:
In "A and B" expression, if A evaluates to false, we skip evaluation of B.
In "A or B" expression, if A evaluates to true, we skip evaluation of B.

This is commonly used compilers, it's nearly embarrassing we don't use it in expressions yet :-)

On a test map that more heavily uses expressions I could measure ~5% improvement (on release build).

### DIFF
--- a/src/core/expression/qgsexpressionnodeimpl.cpp
+++ b/src/core/expression/qgsexpressionnodeimpl.cpp
@@ -298,14 +298,22 @@ QVariant QgsExpressionNodeBinaryOperator::evalNode( QgsExpression *parent, const
 
     case boAnd:
     {
-      QgsExpressionUtils::TVL tvlL = QgsExpressionUtils::getTVLValue( vL, parent ), tvlR = QgsExpressionUtils::getTVLValue( vR, parent );
+      QgsExpressionUtils::TVL tvlL = QgsExpressionUtils::getTVLValue( vL, parent );
+      ENSURE_NO_EVAL_ERROR;
+      if ( tvlL == QgsExpressionUtils::False )
+        return TVL_False;  // shortcut -- no need to evaluate right-hand side
+      QgsExpressionUtils::TVL tvlR = QgsExpressionUtils::getTVLValue( vR, parent );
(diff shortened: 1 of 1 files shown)


src/core/expression/qgsexpressionnodeimpl.cpp
```diff
@@ -298,14 +298,22 @@ QVariant QgsExpressionNodeBinaryOperator::evalNode( QgsExpression *parent, const
 
     case boAnd:
     {
-      QgsExpressionUtils::TVL tvlL = QgsExpressionUtils::getTVLValue( vL, parent ), tvlR = QgsExpressionUtils::getTVLValue( vR, parent );
+      QgsExpressionUtils::TVL tvlL = QgsExpressionUtils::getTVLValue( vL, parent );
+      ENSURE_NO_EVAL_ERROR;
+      if ( tvlL == QgsExpressionUtils::False )
+        return TVL_False;  // shortcut -- no need to evaluate right-hand side
```
nyalldawson: Could the same shortcut apply for nulls?

nyalldawson:
Unless I'm missing something -- isn't this only shortcuting the conversion of the node's evaluation to a boolean value (i.e. it doesn't avoid actually evaluating the node itself)

wonder-sk:
> Unless I'm missing something -- isn't this only shortcuting the conversion of the node's evaluation to a boolean value (i.e. it doesn't avoid actually evaluating the node itself)

:man_facepalming: you're right... Fixed that now - and my test map speedup went from ~5% to ~10% :-)

src/core/expression/qgsexpressionnodeimpl.cpp
wonder-sk: Unfortunately not if we want to be 100% correct with three-value boolean logic. If left value is null/unknown, there are still two options: if right-hand side is FALSE, the result will be FALSE, or null/unknown otherwise.

nyalldawson (approved):

PR #33214, opened 2019-12-03
Move label obstacle settings to "placement" tab and show detailed settings in a stacked panel widget

nyalldawson:
Because:

1. The settings belong next to the placement settings, because they
affect label placement directly (and are closely associated with the
label priority setting in this tab)

2. The label settings widget is ridiculously complex and overwhelming
even for experienced QGIS users. By moving detailed settings into
stacked panels we can avoid the initial complexity of the label settings
whilst also allowing us more flexibility to expose additional settings
in future without adding to the mess. So in future, I propose we move
all detailed settings to sub panels, and leave only the initial "enable"
setting at the top level (e.g. [x] "Repeat levels" -> click settings -> see
all repeat settings).

Peek 2019-12-04 09-58


src/ui/qgslabelobstaclesettingswidgetbase.ui
```diff
@@ -0,0 +1,133 @@
...
+   <item row="0" column="0" colspan="2">
+    <widget class="QGroupBox" name="groupBox">
+     <property name="title">
+      <string>Obstacle Weight</string>
+     </property>
+     <layout class="QVBoxLayout" name="verticalLayout">
+      <item>
+       <widget class="QLabel" name="label">
+        <property name="text">
+         <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;&lt;span style=&quot; font-weight:600;&quot;&gt;Setting a higher weight for obstacles means that labels are less likely to cover features in this layer.&lt;/span&gt;&lt;/p&gt;&lt;p&gt;Only labels with a higher priority then this obstacle weight will be placed over these obstacles.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
nirvn: More verbose description most welcome.

I'm wondering whether part of this description should be conditional to the labelling engine version selected.

nirvn (approved):

nyalldawson:
Bonus fix: we can now actually change the obstacle settings when a layer is set to blocking mode (Previously these settings were stuck inside the disabled label settings widget, which meant that to edit them you had to temporarily enable simple labels, tweak the setting, and then set back to obstacle mode... woot.)

image

NathanW2:
On this note: How do we feel about moving the placement tab to be the
second tab?  I don't like it's a primary thing you normally do but is so
far away from the user.

nirvn:
@NathanW2 , I'm not super warm to the idea of moving the tab.

NathanW2:
Why is that?

nirvn:
@NathanW2 , mainly because this tab isn't always shown in the overall text formatting widget (i.e., outside of labels, the last tabs are hidden). I think it's more effective to have the tabs that always show first, helping muscle memory, etc. And yes, on that front, it does ultimately mean we should move the masking tab.

Also, I feel the text tab and formatting tab are so closely linked to one another it would be a shame to break those brothers apart ;)

PR #33215, opened 2019-12-04
Huge speedup when labels are set to "boundary" obstacle mode

nyalldawson:
Because GEOS prepared predicates are "stubbed out" for many relation types,
such as overlaps and touches, we can get a HUGE speedup by reworking
the obstacle boundary check to utilise an intersects and within check instead
(with the same results)


nyalldawson:
(by huge I mean HUGE. ~ 10 seconds per map refresh down to < 1 second for my test project)

NathanW2:
wow that is impressive!

nirvn:
A two-line change that leads to a 10x speed boost... fantastic :)

PR #33218, opened 2019-12-04
New algorithm "Repair Shapefile"

nyalldawson:
Uses GDAL to repair shapefiles which have a broken or missing .SHX file


alexbruy (approved):
Great!

PR #33219, opened 2019-12-04
Show html files in browser panel

nyalldawson:
Allows .htm(l) files to be shown and opened from the browser panel. These are often used to document data files or mapping projects


nyalldawson:
Any chance of a nice icon @nirvn?

nirvn:
@nyalldawson , what about mLayoutItemHtml.svg ?

PR #33220, opened 2019-12-04
iface.newProject() should return False if a new project could not be created

nyalldawson:
(e.g. if the user opts to cancel when prompted to save the current project)


nirvn (approved):
Good move.

Issue #31994, opened 2019-09-25
Add query reconnect to DBManager 


palmerj:
**Feature description.**
At the moment if a query tab window is open and the connection to a PG database is dropped and you try to execute a query you get the following error: `connection already closed` The workaround at the moment is you need to close the query window and open a new one and restore the SQL text. It would be better to reconnect the database connection with the query tab window, much like you can do with the database node browser. 

**Additional context**
See the current error. 
<img width="418" alt="Screen Shot 2019-09-25 at 2 01 24 PM" src="https://user-images.githubusercontent.com/580022/65568417-04820480-df9d-11e9-8bfe-822d16d0d672.png">

strk:
Interesting enough, a "reconnect" menu item exists if you right-click on the database name, but that connection seems to be different from the one used in the SQL window

strk:
@elpaso I'm not sure if this is related to recent changes in the connection API but when you reconnect from the DBManager right-click menu, you end up with the DlgSqlWindow window pointing to a deleted object:

> wrapped C/C++ object of type PGDatabase has been deleted

elpaso:
@strk I don't think it's related. 
The old API seems to be used for that kind of connection. You can quickly try with 3.4 if it has the same behavior.

strk:
Does indeed happen with 3.4 as well:
> RuntimeError: wrapped C/C++ object of type QAction has been deleted

Way to reproduce:

   1. Open an Sql Query Window from DBManager (F2)
   2. Restart PostgreSQL
   3. Execute a query (twice?)

strk:
With #33225 merged we can close this.

PR #33225, opened 2019-12-04
DBManager PostgreSQL backend using core APIs instead of psycopg2

strk:
The DBManager's SQL window gets stuck when the connection to the (PostgreSQL) database is interrupted (for example on backend restart). This PR is aimed at providing a solution, either as automatic-reconnect or user-requested reconnect action.

See #31994

### DIFF
--- a/python/plugins/db_manager/db_plugins/postgis/connector.py
+++ b/python/plugins/db_manager/db_plugins/postgis/connector.py
@@ -73,14 +73,11 @@ class CursorAdapter():
             self._execute()
 
     def _toStrResultSet(self, res):
-        #print("XXX type of QVariant(None) is " + str(type(QVariant(None))))
         newres = []
         for rec in res:
             newrec = []
             for col in rec:
-                #print("XXX col of rec of resultset valued " + str(col)+ " is typed " + str(type(col)))
                 if type(col) == type(QVariant(None)):
-                    #print("XXX qvariant type of " + str(col)+ " is " + str(col.type))
                     if (str(col) == 'NULL'):
                         col = None
                     else:
@@ -97,7 +94,7 @@ class CursorAdapter():
         if (self.sql == None):
             return
         self._debug("execute called with sql " + self.sql)
-        self.result = self._toStrResultSet(self.connection._executeSql(self.sql))
+        self.result = self._toStrResultSet(self.connection.executeSql(self.sql))
         self._debug("execute returned " + str(len(self.result)) + " rows")
         self.cursor = 0
         self.description = []
@@ -1121,7 +1118,7 @@ class PostGisDBConnector(DBConnector):
         if cursor != None:
             cursor._execute(sql)
             return cursor
-        return CursorAdapter(self, sql)
+        return CursorAdapter(self.core_connection, sql)
 
     def _executeSql(self, sql):
         return self.core_connection.executeSql(sql)
@@ -1129,7 +1126,7 @@ class PostGisDBConnector(DBConnector):
     def _get_cursor(self, name=None):
         #if name is not None:
         #   print("XXX _get_cursor called with a Name: " + name)
-        return CursorAdapter(self, name)
+        return CursorAdapter(self.core_connection, name)
 
     def _commit(self):
         pass
--- a/python/plugins/db_manager/db_plugins/postgis/connector.py
+++ b/python/plugins/db_manager/db_plugins/postgis/connector.py
@@ -35,6 +35,7 @@ from qgis.PyQt.QtCore import (
 from qgis.core import (
     Qgis,
     QgsCredentials,
+    QgsVectorLayer,
     QgsDataSourceUri,
     QgsProviderRegistry,
     QgsProviderConnectionException,
@@ -97,18 +98,30 @@ class CursorAdapter():
         self.result = self._toStrResultSet(self.connection.executeSql(self.sql))
         self._debug("execute returned " + str(len(self.result)) + " rows")
         self.cursor = 0
+
         self.description = []
-        if len(self.result):
-            for i in range(len(self.result[0])):
-                self.description.append([
-                    'column' + str(i),                         # name
-                    str,                                      # type_code
-                    10,                                       # display_size
-                    10,                                       # internal_size
-                    0,                                        # precision
-                    None,                                     # scale
-                    True                                      # null_ok
-                ])
+
+        uri = QgsDataSourceUri(self.connection.uri())
+
+        # TODO: make this part provider-agnostic
+        uri.setTable('(SELECT row_number() OVER () AS __rid__, * FROM (' + self.sql + ') as foo)')
+        uri.setKeyColumn('__rid__')
+        # TODO: fetch provider name from connection (QgsAbstractConnectionProvider)
+        # TODO: re-use the VectorLayer for fetching rows in batch mode
+        vl = QgsVectorLayer(uri.uri(False), 'dbmanager_cursor', 'postgres')
+
+        fields = vl.fields()
+        for i in range(1, len(fields)): # skip first field (__rid__)
+            f = fields[i]
+            self.description.append([
+                f.name(),                         # name
+                str,                              # type_code
+                10,                               # display_size
+                10,                               # internal_size
+                0,                                # precision
+                None,                             # scale
+                True                              # null_ok
+            ])
         self._debug("execute returned " + str(len(self.description)) + " cols")
 
     def fetchone(self):
--- a/python/plugins/db_manager/db_plugins/postgis/connector.py
+++ b/python/plugins/db_manager/db_plugins/postgis/connector.py
@@ -166,12 +166,12 @@ class PostGisDBConnector(DBConnector):
 
         :param uri: data source URI
         :type uri: QgsDataSourceUri
-        :param dbplugin: the PostGisDBPlugin parent instance
-        :type dbplugin: PostGisDbPlugin
         """
         DBConnector.__init__(self, uri)
 
-        self.dbname = uri.database()
+        username = uri.username() or os.environ.get('PGUSER')
+        self.dbname = uri.database() or os.environ.get('PGDATABASE') or username
+        uri.setDatabase(self.dbname)
 
         #self.connName = connName
         #self.user = uri.username() or os.environ.get('USER')
--- a/python/plugins/db_manager/db_plugins/postgis/connector.py
+++ b/python/plugins/db_manager/db_plugins/postgis/connector.py
@@ -37,7 +37,6 @@ from qgis.core import (
     QgsCredentials,
     QgsDataSourceUri,
     QgsProviderRegistry,
-    QgsAbstractDatabaseProviderConnection,
     QgsProviderConnectionException,
 )
 
@@ -59,7 +58,7 @@ def classFactory():
 class CursorAdapter():
 
     def _debug(self, msg):
-        print("XXX CursorAdapter[" + hex(id(self)) + "]: " + msg)
+        #print("XXX CursorAdapter[" + hex(id(self)) + "]: " + msg)
 
     def __init__(self, connection, sql=None):
         self._debug("Created with sql: " + str(sql))
@@ -1127,8 +1126,8 @@ class PostGisDBConnector(DBConnector):
         return self.core_connection.executeSql(sql)
 
     def _get_cursor(self, name=None):
-        if name is not None:
-            print("XXX _get_cursor called with a Name: " + name)
+        #if name is not None:
+        #   print("XXX _get_cursor called with a Name: " + name)
         return CursorAdapter(self, name)
 
     def _commit(self):
--- a/python/plugins/db_manager/db_plugins/postgis/connector.py
+++ b/python/plugins/db_manager/db_plugins/postgis/connector.py
@@ -95,7 +95,10 @@ class CursorAdapter():
         if (self.sql == None):
             return
         self._debug("execute called with sql " + self.sql)
-        self.result = self._toStrResultSet(self.connection.executeSql(self.sql))
+        try:
+            self.result = self._toStrResultSet(self.connection.executeSql(self.sql))
+        except QgsProviderConnectionException as e:
+            raise DbError(e, self.sql)
         self._debug("execute returned " + str(len(self.result)) + " rows")
         self.cursor = 0
 
--- a/python/plugins/db_manager/db_plugins/postgis/connector.py
+++ b/python/plugins/db_manager/db_plugins/postgis/connector.py
@@ -26,7 +26,12 @@ from builtins import range
 
 from functools import cmp_to_key
 
-from qgis.PyQt.QtCore import QRegExp, QFile, QCoreApplication
+from qgis.PyQt.QtCore import (
+    QRegExp,
+    QFile,
+    QCoreApplication,
+    QVariant
+)
 from qgis.core import (
     Qgis,
     QgsCredentials,
@@ -62,6 +67,21 @@ class CursorProxy():
         if (self.sql != None):
             self._execute()
 
+    def _toStrResultSet(self, res):
+        #print("XXX type of QVariant(None) is " + str(type(QVariant(None))))
+        newres = []
+        for rec in res:
+            newrec = []
+            for col in rec:
+                #print("XXX col of rec of resultset valued " + str(col)+ " is typed " + str(type(col)))
+                if type(col) == type(QVariant(None)):
+                    #print("XXX qvariant type of " + str(col)+ " is " + str(col.type))
+                    if (str(col) == 'NULL'):
+                        col = None
+                newrec.append(col)
+            newres.append(newrec)
+        return newres
+
     def _execute(self, sql=None):
         if self.sql == sql and self.result != None:
             print ("XXX CursorProxy execute called with sql " + sql)
@@ -70,7 +90,7 @@ class CursorProxy():
             self.sql = sql
         if (self.sql == None):
             return
-        self.result = self.connection._executeSql(self.sql)
+        self.result = self._toStrResultSet(self.connection._executeSql(self.sql))
         self.description = []
         if len(self.result):
             for i in range(len(self.result)):
--- a/python/plugins/db_manager/db_plugins/postgis/connector.py
+++ b/python/plugins/db_manager/db_plugins/postgis/connector.py
@@ -27,7 +27,14 @@ from builtins import range
 from functools import cmp_to_key
 
 from qgis.PyQt.QtCore import QRegExp, QFile, QCoreApplication
-from qgis.core import Qgis, QgsCredentials, QgsDataSourceUri, QgsCoordinateReferenceSystem
+from qgis.core import (
+    Qgis,
+    QgsCredentials,
+    QgsDataSourceUri,
+    QgsProviderRegistry,
+    QgsAbstractDatabaseProviderConnection,
+    QgsProviderConnectionException,
+)
 
 from ..connector import DBConnector
 from ..plugin import ConnectionError, DbError, Table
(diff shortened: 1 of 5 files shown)


elpaso:
@strk please keep in mind that the roadmap is to eventually get rid of all provider specific connectors implemented in python and use the `QgsAbstractDatabaseProviderConnection` C++ implementation for all DB manager operations. This is not yet possible because the `QgsAbstractDatabaseProviderConnection` is not complete and because not all providers implement it but any PR that will make this switch more difficult should not be accepted.

So, just to remind you that the connection (low level DB connection) status should be IMHO handled in the provider and not in this plugin and not in QGIS core either, what I think could be an acceptable approach is for the `QgsAbstractDatabaseProviderConnection` to throw a specific exception in case of (low-level) connection errors and let client code act on those.

@nyalldawson any opinion on this?

nyalldawson:
@elpaso agreed!

strk:
You mean we should stop using PsycoPg as a whole ?

elpaso:
> You mean we should stop using PsycoPg as a whole ?

Yes. This topic has been discussed a lot, and the ultimate goal is port DB manager to C++ (or better, drop it altogether and replace with the C++  implementation we are already using in the QGIS browser panel), of course this won't happen overnight until we do not reach feature parity but at least let's try to make the steps in the right direction.

The fact the we (you included) are spending a large share of our time (and the donor's money) fixing and patching a constantly broken DB manager is a living proof of the rightfulness of the decision, a great part of the functionality is now available, well tested and well maintained in C++.

strk:
Ok, thanks for the heads up.
I'm trying to see how to replace psycopg2 with calls to https://qgis.org/api/classQgsAbstractDatabaseProviderConnection.html

@elpas It looks like you did some work in that direction for the gpkg dbplugin  so is it a good idea to try at following that path ? One thing I see is that the "model" of the DBManager DBPlugin is to have a concept of a "cursor", which is what you call .execute() on (and .close). This concept is not present in the abstract connection provider, just a plan single-run executeSQL (no transaction?).

elpaso:
> Ok, thanks for the heads up.
> I'm trying to see how to replace psycopg2 with calls to https://qgis.org/api/classQgsAbstractDatabaseProviderConnection.html
> 
> @Elpas It looks like you did some work in that direction for the gpkg dbplugin so is it a good idea to try at following that path ? 

Yes, that was just an initial migration to the new API in order to test its concepts but you can certainly start from there.

> One thing I see is that the "model" of the DBManager DBPlugin is to have a concept of a "cursor", which is what you call .execute() on (and .close). This concept is not present in the abstract connection provider, just a plan single-run executeSQL (no transaction?).

Yeah, transactions are missing, we briefly thought about that and we decided it wasn't a priority for the class scope. 

Do you see a solid use case for that?

strk:
I don't yet see a solid use case, other than migrating incrementally (users might be already using that cursor concept). Is actually not just transactions, but also partial fetches from a cursor (_fetchone) that are missing. Maybe those could be implemented by creating an actual cursor in the database by subclasses...

elpaso:
> I don't yet see a solid use case, other than migrating incrementally (users might be already using that cursor concept). Is actually not just transactions, but also partial fetches from a cursor (_fetchone) that are missing. Maybe those could be implemented by creating an actual cursor in the database by subclasses...

fetchone is just a normal execSql where you just pick the first entry from the first row, no need to implement that.

strk:
it still needs to be implemented, in DBConnector subclasses, unless DBConnector base class as a whole implements that (and drops the "cursor" concept)

strk:
I hate python. The ability to use any member/method drives to such a mess!
The problem is that DBConnector, which is supposedly an abstract class, is using psycopg2 specific models, like:

```
    def _execute(self, cursor, sql):
        if cursor is None:
            cursor = self._get_cursor()
        try:
            cursor.execute(sql)
```

The PostgreSQL specific subclass probably used to have that kind of code, but it was removed at some point in time:
```
    # moved into the parent class: DbConnector._execute()
    # def _execute(self, cursor, sql):
    #       pass
```

Following git history it looks like such "move" was already present in the very first inclusion of code in core, as it is found in 3c5b3bbea180c3b51634f005359d9c36c5f6e78f (2012 commit from @brushtyler )

nyalldawson:
> I hate python. The ability to use any member/method drives to such a mess!

That's exactly why the plan is to move all this code outside of python. We had the same issue with processing too.

Python is just a bad choice for large, complex projects (unless you soak them with near 100% test coverage, including all GUI functionality)

strk:
@elpaso another thing missing in the new API is obtaining the names of returned columns, for a generic query executor (as needed by the SQL window of DBManager). In psycopg2 terms this is cursor's `description` method: http://initd.org/psycopg/docs/cursor.html

The `executeSQL` method of the new API  only rerutns `QList< QList< QVariant > > ` which do not include names and types of returned values

elpaso:
True, any idea? We could add an argument to return column names in the first row?

strk:
> True, any idea? We could add an argument to return column names in the first row?

it's not just names, here's the list of currently expected information:
 - name
 - type_code
 - display_size
 - internal_size
 - precision
 - scale
 - null_ok

strk:
I've faked those records for the moment, pretending all fields are strings with a length of 10 (both internal and display). It seems to work. SQL Window can query the database, even upon restart of backend. This is as of 8d4469901327dcb9bedce6c708ebd7046dcb3921. I'm sure there are lots of places that are broken with this change though, so it'll take some time to adapt more users (for sure topoviewer will need adaptation)

strk:
Getting name in first row could be good enough for a start, but I wonder if we really need cursors

elpaso:
How about a getTableInformation?

strk:
> How about a getTableInformation?

It's not only about table, but about a query result.
Could be a complex select.

In most cases it'd take running the query, or at least
_preparing_ it, or defining a cursor, in order to obtain
that information, so a separate function might be problematic
from a performance point of view, unless the separate
function returns some handle to subsequently fetch rows
from the thing (that's what a "cursor" is in psycopg2).

elpaso:
I would say that if this is a PG specific implementation it's probably acceptable to have a custom implementation in the PG plugin connector.

But I think that getting column information from a query is a valid use case for the abstract connection API, always thinking about **all** supported DB backends, not limited to PG.

python/plugins/db_manager/db_plugins/postgis/connector.py
```diff
@@ -27,7 +27,14 @@
 from functools import cmp_to_key
 
 from qgis.PyQt.QtCore import QRegExp, QFile, QCoreApplication
-from qgis.core import Qgis, QgsCredentials, QgsDataSourceUri, QgsCoordinateReferenceSystem
+from qgis.core import (
+    Qgis,
+    QgsCredentials,
+    QgsDataSourceUri,
+    QgsProviderRegistry,
+    QgsAbstractDatabaseProviderConnection,
```
elpaso: Do you need to import this? It's abstract, you get the concrete instance out of the factory.

strk:
Is there a python function to map the QVariant resultset values into typed python variables ?

elpaso:
> Is there a python function to map the QVariant resultset values into typed python variables ?

You don't need it: SIP automatically takes care of the conversion, if it doesn't then you must write native conversion code like we've done in code for many `Qgs` types.

strk:
It looks like I'm getting some QVariant typed values in the resultset. Debugging further to find out which ones, could be just some missing conversion we have to add. Probably for NULL ?

XXX col of rec of resultset valued NULL is typed <class 'PyQt5.QtCore.QVariant'>
TypeError: setText(self, str): argument 1 has unexpected type 'QVariant'

elpaso:
I'm missing the context :) But there is some conversion code from DB returned values in several places, take a look to the provider implementation of execSql (possibly the private implementation)

strk:
I confirm my problem was with NULL. By adding special code in my CursorProxy to deal with NULL (converting it to None) fixes my case. I guess this should be done at lower levels, will try to find the culprit.

strk:
@elpaso see 73f22499e8ebfa0a22dab213831ff9b57de91ee6 for context.
Basically the value returned by `QgsAbstractDatabaseProviderConnection.executeSql()` (PostgreSQL impmlementation, in this case) contains some values that are typed "QVariant", and seem to be NULL values. The actual database-side datatype of that NULL I still don't know, because that info, as we said, does not get down out of executeSQL

elpaso:
@strk maybe I forgot to check for QVariant NULL here? https://github.com/qgis/QGIS/blob/master/src/providers/postgres/qgspostgresproviderconnection.cpp#L246

strk:
I've added some debugging lines in the PostgreSQL connection code doing the conversion (where you recently added conversion for the `bool` type) and got these:

> src/providers/postgres/qgspostgresproviderconnection.cpp:287 : (executeSqlPrivate) [0ms] Unhandled PostgreSQL type name
> src/providers/postgres/qgspostgresproviderconnection.cpp:287 : (executeSqlPrivate) [0ms] Unhandled PostgreSQL type text
> src/providers/postgres/qgspostgresproviderconnection.cpp:287 : (executeSqlPrivate) [0ms] Unhandled PostgreSQL type _aclitem

So.. I'm guessing I'm hitting the `text` case

strk:
Uhm, but that code returns a 'String` typed variant if I read it correctly ?

strk:
@elpaso the tests/src/python/test_qgsproviderconnection_postgres.py file doesn't seem to have a test for null values, so maybe that'd be a good idea to try out

elpaso:
Yes please!

strk:
So, more issues with current approach of using new API: code exists that _does_ make use of a "scrollable" cursor:
```
src/core/qgsmessagelog.cpp:29 : (logMessage) [2ms] 2019-12-11T10:35:47 Python error[1] Traceback (most recent call last):
  File "/usr/src/qgis/build/dbmanager-reconnect-button/output/python/plugins/db_manager/db_plugins/data_model.py", line 76, in data
    val = self.getData(index.row(), index.column())
  File "/usr/src/qgis/build/dbmanager-reconnect-button/output/python/plugins/db_manager/db_plugins/data_model.py", line 140, in getData
    self.fetchMoreData(start)
  File "/usr/src/qgis/build/dbmanager-reconnect-button/output/python/plugins/db_manager/db_plugins/postgis/data_model.py", line 77, in fetchMoreData
    self.cursor.scroll(row_start, mode='absolute')
AttributeError: 'CursorProxy' object has no attribute 'scroll'
```

I guess for big tables, to avoid fetching all data at once, we want to be able to create a cursor and fetch partials from it. Don't you think the abstract connection provider should provide this ?

strk:
For the state of this PR I could keep building code inside my CursorProxy but it sounds like being at the wrong level (python)...

python/plugins/db_manager/db_plugins/postgis/connector.py
strk: I removed the import completely, with c01e34bf12

strk:
Now that Travis is happy it'd be time to decide on the field names, and the Cursor.
The Cursor class I wrote for this PR is hidden in the PostGIS db_manager db_plugins python class, but could as well be moved up to the base class, so that all subclasses could use the same "policy" (the abstract connector provider in core).  A limitation of such "Cursor" class is that it doesn't really implement a cursor but rather always fetches the whole resultset from queries. And it lacks field names for the returned subset. I think such support is important (proper cursor) because otherwise big tables could never be inspected with DBManager (too much data to transfer from server to client)

elpaso:
Can you please specify better what exactly (and why) is required? 
Because if just need a few rows you could always use LIMIT, and if you need column information you could always use the `QgsVectorLayer` API and check the `fields()`.

Maybe an alternative would be to have generic methods in the core class that just proxy to `executeSql` and implement super-efficient-cursor inside the PG concrete implementation for PG only.

I honestly don't have a clear picture about your use case for this cursor but what I would like to avoid is:
1. a PG specific implementation in core (we have too many already)
2. premature optimization
3. complex core API for corner super-specific-used-only-by-three-people-in-the-world cases

so, I hope you understand my resistance here :)

strk:
Opening a large table is surely not a 3-people niche task, first of
all. You mention LIMIT, but LIMIT/OFFSET is not guaranteed to scan
always the same set of records (records may change between calls,
especially since the new API does NOT support transactions). Also,
LIMIT/OFFSET still retrive all the records until the offset, making
it much slower than an actual cursor.

elpaso:
I still don't understand where do you want to use it and I'm a bit in the blind since you seem to ignore my questions, if you are fetching features why don't you use `QgsVectorLayer` API directly?

strk:
> I still don't understand where do you want to use it and I'm a bit in the blind since you seem to ignore my questions, if you are fetching features why don't you use `QgsVectorLayer` API directly? 

Users are fetching records, not "features". It could be tables without
geometries, just plain data tables, with possibly millions of rows.
That's what the DBManager allows you to do right now.

What other questions have I been ignoring ?

elpaso:
> I still don't understand where do you want to use it and I'm a bit in the blind since you seem to ignore my questions, if you are fetching features why don't you use `QgsVectorLayer` API directly?
> Users are fetching records, not "features". It could be tables without geometries, just plain data tables, with possibly millions of rows. That's what the DBManager allows you to do right now. What other questions have I been ignoring ?

QgsFeature does not need to have a geometry, QgsVectorLayer API handles aspatial tables too and also allows the user to fetch million of rows, it's used through all QGIS.

strk:
Using the vector layer seems a bit abuse to me (there's no _vector_ involved).
For example, I cannot figure out what QgsDataSourceUri should I pass a QgsVectorLayer in order for it to be a no-geometry, data-only result. Note we're talking about queries here, not tables.

There seem to be a required _geometry_ field to be used:

```
src/core/qgsmessagelog.cpp:29 : (logMessage) [0ms] 2019-12-17T15:25:17 PostGIS[1] Erroneous query: SELECT "" FROM (SELECT current_user,current_database()) AS "subQuery_0" LIMIT 0 returned 7 [ERROR:  zero-length delimited identifier at or near """"
LINE 1: SELECT "" FROM (SELECT current_user,current_database()) AS "...
               ^
]
src/core/qgsmessagelog.cpp:29 : (logMessage) [1ms] 2019-12-17T15:25:17 PostGIS[1] Invalid PostgreSQL layer
Fatal: ASSERT: "!key.isNull()" in file /usr/src/qgis/qgis-dbmanager-reconnect-button/src/providers/postgres/qgspostgresconn.cpp, line 394
QGIS died on signal -1Could not attach to process.  If your uid matches the uid of the target
process, check the setting of /proc/sys/kernel/yama/ptrace_scope, or try
again as the root user.  For more details, see /etc/sysctl.d/10-ptrace.conf
ptrace: Operation not permitted.
No thread selected
No stack.
gdb returned 0
```

Not to mention that using a QgsVectorLayer would imply continuously creating and destroying layers for every run query. This is the DBManager SQL window I'm talking about, where a user interactively runs queries, do you imagine the traffic generated by creating and destroying QgsVectorLayers ?

m-kuhn:
Whoo, a lot of different topics going on in here. Let me add to the noise :stuck_out_tongue: 

1. I'm in favor of shipping band aids in the current db manager implementation until we have something new and shiny and solid. Let's be pragmatic
2. @strk could you add a short abstract of what you did to the pull request comment on top? A link to a problem / issue report is very helpful to get started with, but a little prose on the approach taken to assess this problem would make the review easier.
3. there is quite a bit of debug statements in there, could you clean them up (remove the unneeded ones and send the ones needed to the proper logging facilities)
4. message to someone who regularly participates in qgis bugfixing sprints: even though QgsVectorLayer comes with the term "vector" in it, it's very well possible to load a "vector" layer without geometries. Also from postgres.
5. let's be nice and welcoming to each other

strk:
@m-kuhn thanks for the partecipation. I've updated the first comment description with the _intentions_ of this PR. Describing the actual implementation I'd rather do here becuase I've a feeling implementation is going to change over time.

My initial thought was to use a band-aid as you suggest in your first bullet, but at this point, given the amount of time spent on "adapting" the DBManager code to the core classes, and the frustration derived by trying to find the right spot for such a band-aid, I think it makes sense to continue in that way.

What I have at this point, in this PR, is a modification to the PostGisDBConnector class to make it use the core API instead of psycopg2. The PostGisDBConnector class is expected, by DBManager architecture, to implement a DBConnector class which includes methods such as `_get_cursor`, `_fetchone`, `_commit`, `_fetchall`, and requires also being able to return column names and types from queries. The core API I've used (`QgsAbstractDatabaseProviderConnection`) does not have this kind of support (cursors and introspection), so I faked those ones (arbitrary names for columns, locally implemented cursor after fetching all rows) by writing a CursorAdapter class to take care of that.

If the implementation of such abstract CursorAdapter is not PostgreSQL specific so could eventually become a top-level class for the DBManager sake, allowing to drop lots of psycopg2 specific code, but it should probably provide the same level of functionality before going there.

m-kuhn:
So right now we are in limbo between a band aid and something that can be considered a base for a more solid implementation?

There is a conflict of aims with the "more solid implementation" hopefully being in C++.

But if you are comfortable with seeing this disappear in a year or two and meanwhile this helps for a cleaner band aid, I don't see a reason not to merge this (after some cleanup and technical review).

strk:
I'm ok with this work disappearing any time in the future, but given I'm not sure it can be considered ready for merge (lack of SQL query result column names and lack of performant access to big tables) I'm really looking for advice from other devs: do you think it's useful to make this improved use of the core ABstractConnection C++ API from here ? Does it help in finding out the limits of such class (I think it did already). Should those limits be lifted for the sake of future redesign (also a C++ implementation of DBManager will end up needing that Cursor-based access and instrospection for arbitrary query results)

m-kuhn:
Those are important topics, but to be honest, I'm a bit at loss why they are discussed in a pull request with the title "recovery of connection loss".

If there is a proposal, that would deserve its own QEP
If there is more need for discussion, relevant issues and possibly a discussion on the mailing might be better suited.

strk:
Thanks @m-kuhn , I've sent an [email](https://lists.osgeo.org/pipermail/qgis-developer/2020-January/059842.html) to qgis-dev

strk:
Ok, so as anticipated, creating a QgsVectorLayer makes operations far slower as a new layer has to be created for _each_ and _every_ query being executed, including "service" ones. For instance, this I have in the logs, with current state of this PR:

```
XXX CursorAdapter[0x7f59bb3f26a0]: execute called with sql SELECT has_database_privilege(current_database(), 'CREATE'), has_database_privilege(current_database(), 'TEMP')
XXX CursorAdapter[0x7f59bb3f26a0]: execute returned 1 rows
src/providers/postgres/qgspostgresprovider.cpp:100 : (QgsPostgresProvider) [0ms] URI: port=5432 key='__rid__' checkPrimaryKeyUnicity='1' table="(SELECT row_number() OVER () AS __rid__, * FROM (SELECT has_database_privilege(current_database(), 'CREATE'), has_database_privilege(current_database(), 'TEMP')) as foo)" 
```

Basically, some code part of DBManager wants to check for a capability, and runs a query, the query executor creates a VectorLayer to find out names of fields, which triggers lots of other queries.

strk:
NOTE: lazily fetching field names when needed might not be possible as current code does things like:
`return [x[0] for x in c.description]` with `c` being the Cursor, so unless Python allows me to deifine a `getter` I won't be able to leave the current calling code untouched.

roya0045:
> NOTE: lazily fetching field names when needed might not be possible as current code does things like:
> `return [x[0] for x in c.description]` with `c` being the Cursor, so unless Python allows me to deifine a `getter` I won't be able to leave the current calling code untouched.

Do you mean a generator (using 'yield' ) ?

strk:
Thanks @roya0045 for the suggestion but I found a simpler way (`@property`). Now speed seems much better :)

strk:
> I think you can get all this information from the QgsField instance instead of hardcoding.

@elpaso with 8ca08ec798 I've got precision and lenght information from the QgsField instance, but I'm not sure about "displaySize" vs. "internalSize" (I guess displaySize should be left hard-coded or somehow lower?) and I dunno how to extract a mapping between the data type returned by QgsField::type and the python type name...

strk:
Well, I've used the `.type()` method and didn't get errors so I think we're fine to go.

The only missing piece here would be the chunk-based access, but now I'm not sure it is worth the trouble because I could confirm that creating a VectorLayer for each and every query is really killer for performance. 

@palmerj what do you think ? Is the SQL Window of DBManager used for inspecting data in big tables ?

palmerj:
> Is the SQL Window of DBManager used for inspecting data in big tables ?

How do you mean? Showing many rows in the grid or running a query against a large table (hence the longer time for setup if using QgsVectorLayer)? We often do both (or would like to)

strk:
> Showing many rows in the grid or running a query against a large table (hence the longer time for setup if using QgsVectorLayer)? We often do both (or would like to)

I suspect running a query against a large table could still be fast (QgsVectorLayer supposedly fetching rows only on demand, and current code _not_ fetching rows at all using it). But the current code _also_ runs the query via the abstract connection new core API which returns _all_ rows at once.

elpaso (approved):
LGTM

PR #33227, opened 2019-12-04
[postgres] Speed up array parsing (fixes #33226)

### LINKED ISSUE #33226: [postgres] Slow parsing of arrays
**Describe the bug**

It is a performance issue, not really a bug.

Lists of values returned from a PostgreSQL layer are very slow to extract.

**How to Reproduce**

1/ create a database with the 3 following tables:
```
create table t_array as select 1 as id, array_agg(random()) as values from generate_series(1,100000);
create table t_string as select 1 as id, array_to_string(array_agg(random()),',') as values from generate_series(1,100000);
-- simulate a bytea of 100000 8-bytes floating point numbers
create table t_bytea as select 1 as id, decode(repeat('0102030405060708', 100000), 'hex') as values;
```

2/ Load them in QGIS and run:
```
import time
import array

l_array = QgsProject.instance().mapLayersByName("t_array")[0]
l_string = QgsProject.instance().mapLayersByName("t_string")[0]
l_bytea = QgsProject.instance().mapLayersByName("t_bytea")[0]
start = time.time()
for f in l_array.getFeatures():
    assert len(f["values"])==100000
print("array", time.time() - start)
start = time.time()
for f in l_string.getFeatures():
    tt = [None if x == 'NULL' else float(x) for x in f["values"].split(",")]
    assert len(tt) == 100000
print("string", time.time() - start)
start = time.time()
for f in l_bytea.getFeatures():
    tt = array.array("d", f["values"].data())
    assert len(tt) == 100000
print("bytea", time.time() - start)
```

I have the following results (in seconds):
```
array 9.508360624313354
string 0.04486846923828125
bytea 0.0058481693267822266
```

**QGIS and OS versions**

3.11-master 7f415ca7a7

mhugo:
## Description

Speed up the way the PostgreSQL provider parses arrays (of numbers).

With the small benchmark detailed [here](https://github.com/qgis/QGIS/issues/33226), array parsing now takes around 0.2s, which is still slower than raw strings or bytea handling, but way better than before (it was >10s for the same test)

## Checklist

### DIFF
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -4412,11 +4412,10 @@ static void jumpSpace( const QString &txt, int &i )
 QString QgsPostgresProvider::getNextString( const QString &txt, int &i, const QString &sep )
 {
   jumpSpace( txt, i );
-  QString cur = txt.mid( i );
-  if ( cur.startsWith( '"' ) )
+  if ( i < txt.length() && txt.at( i ) == '"' )
   {
     QRegExp stringRe( "^\"((?:\\\\.|[^\"\\\\])*)\".*" );
-    if ( !stringRe.exactMatch( cur ) )
+    if ( !stringRe.exactMatch( txt.mid( i ) ) )
     {
       QgsMessageLog::logMessage( tr( "Cannot find end of double quoted string: %1" ).arg( txt ), tr( "PostGIS" ) );
       return QString();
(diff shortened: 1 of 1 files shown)


src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -4412,11 +4412,10 @@ static void jumpSpace( const QString &txt, int &i )
 QString QgsPostgresProvider::getNextString( const QString &txt, int &i, const QString &sep )
 {
   jumpSpace( txt, i );
-  QString cur = txt.mid( i );
-  if ( cur.startsWith( '"' ) )
+  if ( i < txt.length() && txt.at( i ) == '"' )
   {
     QRegExp stringRe( "^\"((?:\\\\.|[^\"\\\\])*)\".*" );
```
nyalldawson: Would be interesting to see if upgrading this to QRegularExpression would also help

nyalldawson (approved):

src/providers/postgres/qgspostgresprovider.cpp
mhugo: Good point, I was not aware of this new class. I also wonder if storing the regexp statically would be better than constructing it each time ... but I am not sure.

nyalldawson: It gets a bit tricky there - neither of the qt regex class are thread safe, so I don't think a static one can be used here

elpaso: That's a pattern I've used (static regexp) until Nyall pointed out that `QgsRegularExpression` is not thread safe, so using a static may not be a good idea.

mhugo: Ah good point ! Thanks

PR #33233, opened 2019-12-04
Option to filter negative values in logarithmic classification method

3nids:
This adds an option to filter negative and 0 values in logarithmic classification method.

To achieve this, processing parameters are introduced in the classification methods. Reasons for using them:
* classification methods are in core and it would be very complex to add a counter part in gui
* reuse existing code


src/core/classification/qgsclassificationequalinterval.h
```diff
@@ -35,13 +35,14 @@ class CORE_EXPORT QgsClassificationEqualInterval : public QgsClassificationMetho
     QgsClassificationMethod *clone() const override;
     QIcon icon() const override;
 
+    bool valuesRequired() const override {return false;}
```
nyalldawson:
```suggestion
    bool valuesRequired() const override { return false; }
```

src/core/classification/qgsclassificationlogarithmic.cpp
```diff
@@ -13,14 +13,17 @@
...
 #include "qgsclassificationlogarithmic.h"
 #include "qgssymbollayerutils.h"
 #include "qgsapplication.h"
 
 QgsClassificationLogarithmic::QgsClassificationLogarithmic()
-  : QgsClassificationMethod( ValuesNotRequired, 0 )
+  : QgsClassificationMethod( NoFlag, 0 )
 {
-
+  QgsProcessingParameterBoolean *param = new QgsProcessingParameterBoolean( QStringLiteral( "FILTER_ZERO_NEG_VALUES" ), QObject::tr( "Filter values <= 0" ), false );
```
nyalldawson:
```suggestion
  QgsProcessingParameterBoolean *param = new QgsProcessingParameterBoolean( QStringLiteral( "FILTER_ZERO_NEG_VALUES" ), QObject::tr( "Filter values ≤ 0" ), false );
```

src/core/classification/qgsclassificationmethod.cpp
```diff
@@ -161,6 +161,42 @@ QString QgsClassificationMethod::formatNumber( double value ) const
...
+    if ( def->name() == parameterName )
+      return def;
+  }
+  return nullptr;
+}
+
+void QgsClassificationMethod::setParameterValues( const QVariantMap &values )
+{
+  mParameterValues = values;
+  for ( const QString &paramName : mParameterValues.keys() ) // todo is this ok?
```
nyalldawson: no, better to iterate over the map in place and avoid the double-iteration required if you call .keys()
```suggestion
  for ( auto it =mParameterValues.begin(); it != mParameterValue.end(); ++it  )
```

src/core/classification/qgsclassificationmethod.h
```diff
@@ -270,6 +271,24 @@ class CORE_EXPORT QgsClassificationMethod SIP_ABSTRACT
...
+     * Get parameter from its id
+     * \since QGIS 3.12
+     */
+    const QgsProcessingParameterDefinition *parameterDefinition( const QString &parameterName ) const;
+
+    /**
+     * List of parameters
+     * \since QGIS 3.12
+     */
+    const QgsProcessingParameterDefinitions parameterDefinitions() const {return mParameters;}
```
nyalldawson:
```suggestion
    QgsProcessingParameterDefinitions parameterDefinitions() const {return mParameters;}
```

src/core/classification/qgsclassificationmethod.h
```diff
@@ -283,20 +302,32 @@ class CORE_EXPORT QgsClassificationMethod SIP_ABSTRACT
     //! Format the number according to label properties
     QString formatNumber( double value ) const;
 
+    /**
+     * Add parameter
+     * \since QGIS 3.12
+     */
+    void addParameter( QgsProcessingParameterDefinition *definition );
```
nyalldawson:
```suggestion
    void addParameter( QgsProcessingParameterDefinition *definition SIP_TRANSFER );
```

Dox could use a lot of work on this one :wink:

src/core/classification/qgsclassificationmethod.h
```diff
@@ -283,20 +302,32 @@ class CORE_EXPORT QgsClassificationMethod SIP_ABSTRACT
...
   private:
 
     /**
-     * Calculate the breaks, should be reimplemented, values might be an empty list
+     * Calculate the breaks, should be reimplemented, values might be an empty list if they are not required
      * If the symmetric mode is available, the implementation is responsible of applying the symmetry
      * The maximum value is expected to be added at the end of the list, but not the minimum
      */
-    virtual QList<double> calculateBreaks( double minimum, double maximum,
+    virtual QList<double> calculateBreaks( double &minimum, double &maximum,
```
nyalldawson: Dox describing why/when minimum/maximum would be changed

src/core/classification/qgsclassificationmethod.h
nyalldawson: Better also note in the dox that only certain parameter types are supported -- i.e. those which have had their widgets ported to c++.

Pure python widgets obviously won't work here.

src/gui/symbology/qgsgraduatedsymbolrendererwidget.cpp
```diff
@@ -631,6 +642,11 @@ void QgsGraduatedSymbolRendererWidget::connectUpdateHandlers()
   connect( cbxAstride, &QAbstractButton::toggled, this, &QgsGraduatedSymbolRendererWidget::classifyGraduated );
   connect( cboSymmetryPoint, qgis::overload<int>::of( &QComboBox::currentIndexChanged ), this, &QgsGraduatedSymbolRendererWidget::classifyGraduated );
   connect( cboSymmetryPoint->lineEdit(), &QLineEdit::editingFinished, this, &QgsGraduatedSymbolRendererWidget::symmetryPointEditingFinished );
+
+  for ( const auto *ppww : qgis::as_const( mParameterWidgetWrappers ) )
```
nyalldawson: No auto

src/gui/symbology/qgsgraduatedsymbolrendererwidget.cpp
```diff
@@ -798,6 +819,30 @@ void QgsGraduatedSymbolRendererWidget::methodComboBox_currentIndexChanged( int )
   }
 }
 
+void QgsGraduatedSymbolRendererWidget::updateMethodParameters()
+{
+  while ( mParametersLayout->rowCount() )
+    mParametersLayout->removeRow( 0 );
```
nyalldawson: I have a feeling this leaks. Can you double check?

src/gui/symbology/qgsgraduatedsymbolrendererwidget.cpp
```diff
@@ -798,6 +819,30 @@ void QgsGraduatedSymbolRendererWidget::methodComboBox_currentIndexChanged( int )
...
+  QgsClassificationMethod *method = QgsApplication::classificationMethodRegistry()->method( methodId );
+  Q_ASSERT( method );
+
+  // todo need more?
+  QgsProcessingContext context;
+
+  for ( const auto *def : method->parameterDefinitions() )
+  {
+    QgsAbstractProcessingParameterWidgetWrapper *ppww = QgsGui::processingGuiRegistry()->createParameterWidgetWrapper( def, QgsProcessingGui::Standard );
+    mParametersLayout->addRow( ppww->createWrappedLabel(), ppww->createWrappedWidget( context ) );
```
nyalldawson: ppww->createWrappedLabel() may return null, e.g. with checkboxes. Is that an issue here?

src/ui/qgsgraduatedsymbolrendererwidget.ui
```diff
@@ -6,15 +6,15 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>498</width>
-    <height>641</height>
+    <width>477</width>
```
nyalldawson: I wonder -- instead of adding the widgets directly in this (already very complex) widget, could we instead have a "..." or "settings" button next to the classification method combobox which opens a new pane allowing these method-specific widgets to be set?

src/core/classification/qgsclassificationmethod.cpp
```diff
@@ -161,6 +161,42 @@ QString QgsClassificationMethod::formatNumber( double value ) const
...
 
+void QgsClassificationMethod::addParameter( QgsProcessingParameterDefinition *definition )
+{
+  mParameters.append( definition );
+}
+
+QVariant QgsClassificationMethod::parameterValue( const QString &name ) const
+{
+  const QgsProcessingParameterDefinition *def = parameterDefinition( name );
+  if ( def )
```
nyalldawson: Hm.. I'm unsure on this part -- you aren't using the regular processing "parameterAs..." methods to evalute the parameter value to a variant. While this works fine with just the boolean parameter used here, I'm concerned that you'll eventually end up needing to re-implement all the complex logic which sits in the parameterAs... methods.

I think we could avoid this by creating a QgsProcessingFeedback and QgsProcessingContext object and then using the standard evaluation methods here.

nyalldawson (commented):
Exciting to see this progress -- it's a great re-use of the processing framework!

nyalldawson:
Can't wait till you make an adapter so that all registered classification methods automatically get exposed as processing algorithms too :wink:

src/gui/symbology/qgsgraduatedsymbolrendererwidget.cpp
3nids: yes, for the log method I used a boolean.

src/ui/qgsgraduatedsymbolrendererwidget.ui
3nids: The UI is indeed crowded but I have the feeling it should not be the parameter of the method which should be hidden in a secondary UI.
Also, I would like to avoid extra dialog popping out, I love the flattening experience here (kudos to @NathanW2 for this awesome improvement).

From my experience, it's rather the legend I almost never touch (legend format, precision, trim trailing 0). Also, I always leave connecting classes checked.

src/core/classification/qgsclassificationmethod.cpp
3nids: I switched to using QgsProcessingParameters::paramAsXxxxx

src/core/classification/qgsclassificationequalinterval.h
```diff
@@ -35,13 +35,14 @@ class CORE_EXPORT QgsClassificationEqualInterval : public QgsClassificationMetho
     QgsClassificationMethod *clone() const override;
     QIcon icon() const override;
 
+    bool valuesRequired() const override { return false; }
+
     static const QString METHOD_ID;
 
   private:
-    QList<double> calculateBreaks( double minimum, double maximum,
+    QList<double> calculateBreaks( double &minimum, double &maximum,
```
elpaso: Can it be const?

src/core/classification/qgsclassificationlogarithmic.h
```diff
@@ -34,10 +34,13 @@ class CORE_EXPORT QgsClassificationLogarithmic : public QgsClassificationMethod
     QString id() const override;
     QIcon icon() const override;
     QString labelForRange( double lowerValue, double upperValue, ClassPosition position ) const override;
+    bool valuesRequired() const override;
 
   private:
-    QList<double> calculateBreaks( double minimum, double maximum, const QList<double> &values, int nclasses ) override;
+    QList<double> calculateBreaks( double &minimum, double &maximum, const QList<double> &values, int nclasses ) override;
```
elpaso:
```suggestion
    QList<double> calculateBreaks( const double &minimum, const double &maximum, const QList<double> &values, int nclasses ) override;
```
maybe const?

src/core/classification/qgsclassificationequalinterval.h
3nids: nope.
The point is that the method might update min/max depending on config.
And the methods were returning classes with upper bound only, so the base class needs to add the minimum once again.
I have chosen to not modify the behavior of the methods (by return the lower bound too) as it might have causes unexpected issues in QGIS (they are used in other places).
So to make it const we have to modify this.

nyalldawson:
I don't think this is working quite right:

Peek 2019-12-06 11-09

3nids:
What makes you think it's not working?
If you have negative values, the log is undefined and you cannot calculate the classes.
The issue is that you need the smallest strictly positive value to get the minimum bound in the log scale, therefore not being able to use the min/max only.
I will modify the checkbox by a combobox

image


is this clearer?

nyalldawson:
I don't have any negative values -- so I'd expect the results to be the same regardless of the state of that checkbox??

3nids:
you have 0 values?

nyalldawson:
Ah, that's it!

3nids:
I will modify the label with "negative or 0"

PR #33237, opened 2019-12-05
Fix broken authentication/username/password support in WMS and other providers

nyalldawson:
Followup 89de7579

These providers rely on the QgsDataSourceUri param methods for storing and retrieving
authentication details, so the change made to use the dedicated username/password/auth
setters in 89de7579 broke all auth support for these providers.

Ensure that the param system is fully compatible and aliased to the dedicated username/password/authcfg getters and setters

### DIFF
--- a/src/core/qgsdatasourceuri.cpp
+++ b/src/core/qgsdatasourceuri.cpp
@@ -598,19 +598,40 @@ QByteArray QgsDataSourceUri::encodedUri() const
   {
     url.addQueryItem( it.key(), it.value() );
   }
+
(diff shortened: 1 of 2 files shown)


nyalldawson:
(urgent review needed so that we can get the backport in for 3.10.1)

elpaso (approved):
LGTM, if I get the intention this should now work the same regardless how username/password/authcf are set/gotten, correct?

nyalldawson:
@elpaso that's correct -- they are effectively aliases for each other

PR #33240, opened 2019-12-05
Add option to auto-rotate canvas to GPS bearing, show GPS bearing as a line over map

nyalldawson:
**Add option to auto-rotate canvas to GPS bearing**:

When enabled, the map canvas will automatically rotate so that it's oriented in the same direction as the GPS bearing

**Show GPS bearing as a line over map**:

Allow showings a bearing line from the GPS location pointed in the GPS's direction. Allows users to view a "current path" directional line as they navigate using a GPS

### DIFF
--- a/src/app/qgspointmarkeritem.h
+++ b/src/app/qgspointmarkeritem.h
@@ -99,7 +99,7 @@ class APP_EXPORT QgsMapCanvasSymbolItem: public QgsMapCanvasItem
 
 /**
  * \ingroup app
- * \class QgsPointMarkerItem
+ * \class QgsMapCanvasMarkerSymbolItem
  * \brief An item that shows a point marker symbol centered on a map location.
  */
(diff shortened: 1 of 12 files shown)


src/app/qgspointmarkeritem.h
```diff
@@ -27,86 +27,133 @@
...
+  private:
+
     std::unique_ptr< QgsDrawSourceEffect > mOpacityEffect;
 
-    QgsRenderContext renderContext( QPainter *painter );
+};
+
+/**
+ * \ingroup app
+ * \class QgsPointMarkerItem
```
nirvn: Should QgsPointMarkerItem be renamed to QgsMapCanvasMarkerSymbolItem here?

nirvn (approved):
Code looking good, I think there's only an issue with the dox.

nyalldawson:
Unrelated test failure

PR #33241, opened 2019-12-05
Show the total pan distance and bearing in the status bar 

nyalldawson:
...during canvas pan operations

Allows users to know exactly how far (and in what direction) they've dragged the map.


nirvn (approved):
Nice little feature.

PR #33246, opened 2019-12-05
Support for NaN in text edit widget

m-kuhn:
If one currently snaps to a geometry with undefined z coordinate (`NaN`), and uses a default expression for an attribute `z($geometry)`, the text will be `nan` and QGIS will try to save the attribute as text nan, which won't work on a numeric field.

This will work around the issue by mapping `nan` to `NULL` in the text edit widget.

There would be alternative approaches like mapping nan to NULL at the end of `QgsExpression::evaluate()`.

### DIFF
--- a/src/gui/editorwidgets/qgstexteditwrapper.cpp
+++ b/src/gui/editorwidgets/qgstexteditwrapper.cpp
@@ -228,10 +228,10 @@ void QgsTextEditWrapper::setWidgetValue( const QVariant &val )
     if ( !( field().type() == QVariant::Int || field().type() == QVariant::Double || field().type() == QVariant::LongLong || field().type() == QVariant::Date ) )
       v = QgsApplication::nullRepresentation();
   }
+  else if ( val.type() == QVariant::Double && std::isnan( val.toDouble() ) )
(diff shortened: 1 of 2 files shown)


lbartoletti (approved):

m-kuhn:
@elpaso I'm always hesitant to test hacks

elpaso:
I didn't know it was a hack :) 

My reason for asking was that the NULL/None/NaN/std::min/max/isEmpty() etc, handling in these widgets is "fragile" at best (when it's not broken already, see relation ref), I'd really like to see tests here or you hack won't survive long.

m-kuhn:
I'm really not too fond of doing a test like this. I'd actually even be happy to not have my hack live for long. I could very well imagine a future implementation where nan is preserved and not mapped to null. Let me see if I can come up with something ...

m-kuhn:
Hmmm... And what's wrong with relation ref that's worth noting here?

m-kuhn:
Other approaches seem to be poorly backed by real world use cases. Let's stick to this one for now.

m-kuhn:
@elpaso your review would be welcome

elpaso (approved):

PR #33247, opened 2019-12-05
remove invokable because it is already used on another method with same name

3nids:
This was introduced in cd5642aca0dab70b104e2325ca0e7448dddbd9a5

You should not define two methods as invokable if they have the same name as the call is uncertain.
Since it's javascript there is no type checking.

If the method must be called, I would suggest to add an extra helper in QgsQuickUtils.

### DIFF
--- a/src/core/qgsunittypes.h
+++ b/src/core/qgsunittypes.h
@@ -564,7 +564,7 @@ class CORE_EXPORT QgsUnitTypes
      *
      * \since QGIS 3.8
      */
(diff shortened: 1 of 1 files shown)


nirvn (approved):
It's fine by me, I must have copy pasted the difference version of toAbbreviatedString(...) variants and left the Q_INVOKABLE in there.

If Qml needs it one day, the solution should handle all five variants anyways.

PR #33252, opened 2019-12-05
[postgres] make createEmptyLayer respect the state of the "overwrite" flag and succeed also if the table already exists (when the "overwrite" flag is not set)

### LINKED ISSUE #27039: Allow importing data into existing tables with processing
Author Name: **Robert Collins** (@rbtcollins)
Original Redmine Issue: [19210](https://issues.qgis.org/issues/19210)
Affected QGIS version: 3.0.3
Redmine category:data_provider/postgis

---
When importing a layer that needed some schema tweaks (specifically to change the geom to a plain 'geometry' to avoid Polygon vs MultiPolygon errors... I wanted to turn off layer overwriting.

The ImportIntoPostGIS processing tool passes the overwrite option to 
qgsvectorlayerexporter which passes it to 
QgsPostgresProvider::createEmptyLayer

but as you can see here https://github.com/qgis/QGIS/blob/0a325f1eda4054418e8d31bfc1d33d87cbdc2c0f/src/providers/postgres/qgspostgresprovider.cpp#L3878 and then https://github.com/qgis/QGIS/blob/0a325f1eda4054418e8d31bfc1d33d87cbdc2c0f/src/providers/postgres/qgspostgresprovider.cpp#L3893 the PostGIS driver only uses the overwrite flag to skip dropping the existing table (good), but not to avoid creating a new table (bad) - and this naturally fails.

This seems to be the case even very far back in history, so perhaps this should be considered a feature request rather than a bug?

havatv:
## Description

The `Export to PostgreSQL` QGIS Database algorithm fails if the specified output table exists and the "overwrite" checkbox is not checked.
An exception is thrown: `ERROR: relation "test" already exists`

Issue #27039 is a request for QGIS to try to append to the existing table instead of just throwing this exception (the issue has been given the **Bug** label).

The aim of the PR is to fix this.
The fix will also allow appending to an existing PostgreSQL table (as long as the table structure is "compatible" with the features appended).
Other RDBMS providers (MSSQL and DB2) deny appending (in a slightly "cleaner" way), as do OGR and Spatialite.

The changes are reasonably simple (essentially protecting the creation of the table with an extra `if` statement, to avoid creating an existing table).

I am not building from source, so I have not run the code. Therefore careful review is required (and testing is welcome).

Since the changes to the code are very limited, backporting to 3.4 should be trivial.

## Checklist

Fixes #27039 (hopefully)


havatv:
The failing of the checks seems does not seem to be related to this PR.

roya0045:
I had a similar issue, I would suggest to rebase with master to see if that solves it.

havatv:
Thanks for the input, @roya0045.
I updated the branch yesterday, but the that did not seem to solve anything.  But I guess I have to try again...

havatv:
All checks passed, @m-kuhn.

m-kuhn:
I think the current behavior is correct. All algorithms we have will create a new table. And optionally drop an existing one.

As far as I can see, what you are looking for is an algorithm to insert/append to an existing table. There is a [plugin that does that](https://github.com/gacarrillor/AppendFeaturesToLayer) and an implementation in core would be really nice, but will be provider independent.

Btw: You can just close and reopen the PR (without merging or rebasing to master) if you happen to have an unrelated error, that avoids extra merge commits.

gioman:
> is an algorithm to insert/append to an existing table

the core Processing tool based on ogr (to import into PostGIS) also has an append option.

havatv:
> I think the current behavior is correct. All algorithms we have will create a new table. And optionally drop an existing one.

@m-kuhn, it would be strange if the current behaviour is correct. It may be that the behaviour is
intended to be different from what my PR proposes, but it is hard to believe that it is correct.

The problem here is that `QgsPostgresProvider::createEmptyLayer` accepts
and uses the "overwrite" option.
When the overwrite option is `False`, the algorithm will (correctly, in my opinion) not drop
the existing table, but still try to create the table. This will, not unexpectedly, result
in an exception (`ERROR: relation "..." already exists`).

> As far as I can see, what you are looking for is an algorithm to insert/append to an existing table. There is a [plugin that does that](https://github.com/gacarrillor/AppendFeaturesToLayer) and an implementation in core would be really nice, but will be provider independent.

@m-kuhn, I am not looking for "an algorithm to insert/append to an existing table".
What I observed (while documenting) was that the "The Export to PostgreSQL QGIS Database"
QGIS algorithm fails if the specified output table exists and the "overwrite" checkbox is not checked.

This PR will avoid the `ERROR: relation "..." already exists` exception, and also allow appending
to an existing table if the table has the same columns and geometry types (if not, and exception
would be thrown, but that would be expected behaviour).

m-kuhn:
To me the current behavior of `createEmptyLayer` sounds safe and sane:

 - Create a new table if not existant
 - Overwrite existing table if the overwrite flag has been specified
 - **Fail if the user tries to create an empty table when one exists already and overwrite has not been specified**

Some dummy code to explain:

```
provider = createEmptyLayer('schema.existing_table')
provider.addSomeData()
```

I would not expect this code to silently ignore that the table exists already and append to it. I think it's a good behavior to fail and be forced to deal with this unexpected situation.

In reality this means likely something like: the user has started a processing alg twice and will now have each record in there duplicated.
I'd rather fail and let the user figure out if he wants to specify "overwrite" (clean and restart) or explicitly use an append and with this make it clear that the algorithm is intended to be started multiple times and used for incremental import.

m-kuhn:
What could be done instead is:

- Adjust the docstring of `QgsProviderMetadata::createEmptyLayer` to be explicit about the "this method will fail if the layer already exists".
- Add a new return value `LayerAlreadyExists` which can be used by the caller to e.g. show a dialog box to the user "The layer already exists. [Append], [Overwrite], [Cancel]"

CC @elpaso who has a very good overview of the provider infrastructure.

havatv:
>     * Adjust the docstring of `QgsProviderMetadata::createEmptyLayer` to be explicit about the "this method will fail if the layer already exists".

Yes, that is a possiblity.

>     * Add a new return value `LayerAlreadyExists` which can be used by the caller to e.g. show a dialog box to the user "The layer already exists. [Append], [Overwrite], [Cancel]"

Also a possibility.

> CC @elpaso who has a very good overview of the provider infrastructure.

My impression was that allowing append for PostgreSQL would be consistent with
what the spatialite (src/providers/spatialite/qgsspatialiteprovider.cpp) and the OGR/GDAL (src/core/providers/ogr/qgsogrprovider.cpp) providers offer.
The MSSQL and DB2 providers both deny appending, but they seem to handle it in a "cleaner" way.

So I guess this boils down to the "append or not" issue.  GDAL/OGR and Spatialite offer append, while none of the DB providers seem to offer it.
The OGR/GDAL / Spatialite behaviour seems most intuitive to me.  
Allowing append at the provider level, also for DBMS providers, would make QGIS a more powerful (but not more "dangerous") tool. Since it is so easy to support append, I think it is worth it.
The developers can then decide whether they would like to take advantage of the append opportunity in algorithms and elsewhere.

m-kuhn:
I think we are discussing different things here.

- Append on provider level is possible already. Basically when you digitize a new point on the map canvas, it's appended after all ;)
- Append as processing functionality is available as ogr alg, plugin, but not yet as a "append to any layer core functionality"

What this PR proposes on the other hand is to "make createEmptyLayer() silently succeed even if an (empty or not) table is already in place". I didn't check what other providers do, but this approach seems dangerous to me.

m-kuhn:
> My impression was that allowing append for PostgreSQL would be consistent with
what the spatialite (src/providers/spatialite/qgsspatialiteprovider.cpp) and the OGR/GDAL (src/core/providers/ogr/qgsogrprovider.cpp) providers offer.

I just had a look and the [spatialite provider](https://github.com/qgis/QGIS/blob/master/src/providers/spatialite/qgsspatialiteprovider.cpp#L241-L249) as well as [the ogr provider](https://github.com/qgis/QGIS/blob/master/src/core/providers/ogr/qgsogrprovider.cpp#L399-L408) seems to behave equal to postgres and bail out if the table already exists.

nyalldawson:
> Append as processing functionality is available as ogr alg, plugin, but not yet as a "append to any layer core functionality"

And this one isn't as easy as just adding support for append in the data provider (as is done here) -- rather we need to ensure that field orders are correctly matched (if the feature to append and the original layer fields differ), correctly reproject features into the destination layer's crs, and automagically transform different geometry types to match the destination (e.g. segmenting curved geometry types where required, silently upgrading single->multiparts if needed, adding or dropping z/m values, etc)...

Adding this functionality to processing has been on my mind for a long time, but it's non-trivial and would require some funding

havatv:
> I just had a look and the [spatialite provider](https://github.com/qgis/QGIS/blob/master/src/providers/spatialite/qgsspatialiteprovider.cpp#L241-L249) as well as [the ogr provider](https://github.com/qgis/QGIS/blob/master/src/core/providers/ogr/qgsogrprovider.cpp#L399-L408) seems to behave equal to postgres and bail out if the table already exists.

@m-kuhn, you are correct.  They have the same behaviour. I must have misunderstood the code when reading it the first time.

havatv:
> And this one isn't as easy as just adding support for append in the data provider (as is done here) -- rather we need to ensure that field orders are correctly matched (if the feature to append and the original layer fields differ), correctly reproject features into the destination layer's crs, and automagically transform different geometry types to match the destination (e.g. segmenting curved geometry types where required, silently upgrading single->multiparts if needed, adding or dropping z/m values, etc)...
> 
> Adding this functionality to processing has been on my mind for a long time, but it's non-trivial and would require some funding

@nyalldawson , This PR addresses #27039. If that issue is a "won't fix", I think it should be marked accordingly (it currently has the labels "Bug" and "Data provider").

With this PR merged, it would be possible to append to a PostgreSQL table **if** the (schema / table definition of) the *Layer to import* in "Export to PostgreSQL" is identical to / compatible with the PostgreSQL table (*Table to import*) definition / schema. If it is not "compatible", an exception would be thrown when an INSERT was attempted.
I think that would be a first step in the right direction.

m-kuhn:
The problem with #27039 is that it mixes a problem statement with a proposed solution. I would focus on forgetting the proposed solution in a first step and get clear requirements for the *what* and then decide on the *how*.

If you want to hack something right now, I think the proposal https://github.com/qgis/QGIS/pull/33252#issuecomment-562938696 will be safe to do.

havatv:
Since there are no immediate plans to support appending, I have updated the QGIS database algorithm documentation accordingly.
These algorithms all take an overwrite parameter / flag, but even if the flag is set to False, overwriting will take place (at the geopackage level for GeoPackage, at the layer level for Sqlite, at the table level for PostgreSQL).
https://github.com/qgis/QGIS-Documentation/pull/4599

m-kuhn:
> These algorithms all take an overwrite parameter / flag, but even if the flag is set to False, overwriting will take place (at the geopackage level for GeoPackage, at the layer level for Sqlite, at the table level for PostgreSQL).

Does it overwrite? As far as I understood it will try to overwrite and then fail?

havatv:
@m-kuhn, the algorithms I am talking about are processing algorithms, found under QGIS-> Database in the Processing toolbox.
Example: I just tested "Package layers" (for loading into GeoPackage), and even if I don't check "Overwrite existing GeoPackage", only the layers added are present in the GeoPackage after loading.

roya0045:
> @m-kuhn, the algorithms I am talking about are processing algorithms, found under QGIS-> Database in the Processing toolbox.
> Example: I just tested "Package layers" (for loading into GeoPackage), and even if I don't check "Overwrite existing GeoPackage", only the layers added are present in the GeoPackage after loading.

It used to be ok, if this is true this could wipe out a lot of data for no reason, if you are sure that this is the case I would suggest making a ticket.

PR #33253, opened 2019-12-05
Allow saga algorithms to be run under SAGA 7.3, but with large warnings

nyalldawson:
Trying to unblock the stalemate with PROJ 6/GDAL 3 and the MacOS infrastructure, this PR adds the ability to run the SAGA algorithms under the new SAGA 7.3 LTR release. BUT..... show a huge obnoxious "NOT OFFICIALLY SUPPORTED" warning so that users know they are venturing into uncharted, not supported, dragon-filled territory.

Adding full support for SAGA 7.3 is going to be a painstaking, tedious, heart-breaking process (community, non QGIS-core-dev volunteers welcome to lead this effort!!!), but this should at least hopefully allow the mac builds to upgrade their dependencies without completely breaking SAGA support.

Ping @PeterPetrik @alexbruy

### DIFF
--- a/python/core/auto_generated/processing/qgsprocessingprovider.sip.in
+++ b/python/core/auto_generated/processing/qgsprocessingprovider.sip.in
@@ -107,6 +107,18 @@ missing external dependencies).
 .. seealso:: :py:func:`isActive`
 %End
 
+    virtual QString warningMessage() const;
+%Docstring
+Returns an optional warning message to show users when running algorithms from this provider.
+
+This can be used to return a translated warning message which should be shown to users
+of this provider. It's intended for use in cases such as a provider which relies on a 3rd-party
+backend, where the version of the backend software is not officially supported, or for
+alerting users to providers in a "beta" or "untrustworthy" state.
+
+.. versionadded:: 3.10.1
+%End
+
     virtual bool isActive() const;
 %Docstring
 Returns ``True`` if the provider is active and able to run algorithms.
--- a/python/plugins/processing/algs/saga/SagaAlgorithmProvider.py
+++ b/python/plugins/processing/algs/saga/SagaAlgorithmProvider.py
@@ -40,6 +40,7 @@ pluginPath = os.path.normpath(os.path.join(
     os.path.split(os.path.dirname(__file__))[0], os.pardir))
 
 REQUIRED_VERSION = '2.3.'
+BETA_SUPPORT_VERSION = '7.3.'
 
 
 class SagaAlgorithmProvider(QgsProcessingProvider):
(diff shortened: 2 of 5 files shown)


lbartoletti:
cc @rhurlin for FreeBSD too

rhurlin:
@lbartoletti: Thanks for cc'ing me

@nyalldawson: I like the approach to open against newer SAGA LTR version. But I am wondering, if we couldn't open it a bit wider, say '>= newest LTR version'. Then a (FreeBSD) user can also use other, newer than LTR SAGA versions, like e.g. the FreeBSD port system offers.

On FreeBSD, we allow to use newer SAGA versions with QGIS for quite some time now. Therefore, we patch the sources like this:

```
--- python/plugins/processing/algs/saga/SagaAlgorithmProvider.py.orig	2019-03-02 02:07:46 UTC
+++ python/plugins/processing/algs/saga/SagaAlgorithmProvider.py
@@ -82,7 +82,7 @@ class SagaAlgorithmProvider(QgsProcessingProvider):
 
     def canBeActivated(self):
         version = SagaUtils.getInstalledVersion(True)
-        if version is not None and version.startswith(REQUIRED_VERSION):
+        if version is not None and version >= REQUIRED_VERSION:
             return True
         return False
 
@@ -93,7 +93,7 @@ class SagaAlgorithmProvider(QgsProcessingProvider):
                                      self.tr('Processing'), Qgis.Critical)
             return
 
-        if not version.startswith(REQUIRED_VERSION):
+        if version < REQUIRED_VERSION:
             QgsMessageLog.logMessage(self.tr('Problem with SAGA installation: unsupported SAGA version (found: {}, required: {}).').format(version, REQUIRED_VERSION),
                                      self.tr('Processing'),
                                      Qgis.Critical)
```

alexbruy:
Looks good to me (though IMHO it should be disabled/removed completely).

nyalldawson:
@rhurlin 

> I like the approach to open against newer SAGA LTR version. But I am wondering, if we couldn't open it a bit wider, say '>= newest LTR version'. Then a (FreeBSD) user can also use other, newer than LTR SAGA versions, like e.g. the FreeBSD port system offers.

That's what we used to do, but it just led to a nightmare of maintenance effort (because SAGA api changes all the time, even within a single major release series). Accordingly we made the decision to only support SAGA LTR releases and nothing else...

rhurlin:
@nyalldawson Of course, this is understandable. On FreeBSD, we can widen this as before, if wanted.

Many thanks for your efforts, really appreciated.

PeterPetrik (approved):
thanks for doing this

gioman:
> Trying to unblock the stalemate with PROJ 6/GDAL 3 and the MacOS infrastructure, this PR adds the ability to run the SAGA algorithms under the new SAGA 7.3 LTR release. BUT..... show a huge obnoxious "NOT OFFICIALLY SUPPORTED" warning so that users know they are venturing into uncharted, not supported, dragon-filled territory.
> 
> Adding full support for SAGA 7.3 is going to be a painstaking, tedious, heart-breaking process (community, non QGIS-core-dev volunteers welcome to lead this effort!!!), but this should at least hopefully allow the mac builds to upgrade their dependencies without completely breaking SAGA support.

@nyalldawson thanks for this. I'm anyway a little confused by the whole SAGA situation. Was are the implications for QGIS in general (not only for macOS)? We must update the description files for the tools in order to work with SAGA ltr? @alexbruy

PR #33254, opened 2019-12-06
Show helpful status bar tooltip informing of the current select tool behavior

nyalldawson:
whenever ctrl/shift/alt modifiers are changed during a selection operation


nirvn (approved):
Nice. Can you think of other tools that'd benefit from such an approach? The label tools?

nyalldawson:
>  The label tools?

PR #33262, opened 2019-12-06
Classification methods: save/read parameters, improve negative value handling in log method

3nids:


src/core/classification/qgsclassificationlogarithmic.cpp
```diff
@@ -53,32 +53,52 @@ QIcon QgsClassificationLogarithmic::icon() const
...
+      minimum = positiveMinimum; // the minimum gets updated
+  }
+  else
+  {
+    positiveMinimum = minimum;
   }
 
   // get the min/max in log10 scale
-  double log_min = std::floor( std::log10( minimum ) );
+  double log_min = std::floor( std::log10( positiveMinimum ) );
```
nyalldawson: What's with the python style naming?

nyalldawson (commented):
LGTM!

nyalldawson (approved):

src/core/classification/qgsclassificationlogarithmic.cpp
3nids: what do you mean?

nyalldawson: it should be logMin, not log_min... (I know it's petty, but if we're not super-strict on this then our api and code will become a 2.x style mess again....)

3nids: oops, too much Python lately

PR #33265, opened 2019-12-06
Attempt to fix Windows GPKG project storage

### LINKED ISSUE #33057: projects saved inside a GPKG do not show in browser if the datasource is a network drive
I have a GPKG with a project saved in it:

if the GPKG is in "c" the browser shows correctly the project and the layers.

If the datasource is in a network drive (i.e. \\vboxsrv) mapped as a drive letter then the project does not show, but layers do (and they load ok).

Tested on 3.10 and master on Windows.

elpaso:
Fixes #33057 (hopefully)


agiudiceandrea:
Hi @elpaso, have you seen https://github.com/qgis/QGIS/issues/46969#issuecomment-2005182655 and https://gis.stackexchange.com/questions/494572/qgis-cant-open-embedded-project-from-geopackage-via-drag-and-drop-on-non-standa?

PR #33266, opened 2019-12-06
Being solid against field names variation in expressions on WFS

signedav:
When an expression had an alias or a field name written upper-/lowercase then it has not been able to map the cached spatialite field in `QgsBackgroundCachedFeatureIterator::buildRequestCache`, what resulted in a crash.

Since we get the field name used for the mapping of the cached field with `lookupField()` it works with expressions containing aliases and case intensive written field names.


m-kuhn:
@rouault does this look good to you?

m-kuhn:
Since this fixes a crash, I'll merge it.
@rouault if this finds you later, a look would still be appreciated.
@signedav can you prepare a backport?

signedav:
@m-kuhn I think this has not been an issue on 3.10 since it used `QgsWFSFeatureIterator` directly. Or do I miss something?

m-kuhn:
Oh right, sorry for the noise
https://github.com/qgis/QGIS/pull/32262#issuecomment-543416292

PR #33270, opened 2019-12-06
[UX][feature??] add progress bar to the auto layer finder

roya0045:
## Description

This simply add a progress bar to the auto layer finder feature, this should prevent some confusion and give a sense of progress to the user instead of an apparent freeze.

progress

## Checklist

### DIFF
--- a/src/app/qgshandlebadlayers.cpp
+++ b/src/app/qgshandlebadlayers.cpp
@@ -557,7 +557,7 @@ void QgsHandleBadLayers::autoFind()
     const QString fileType = mLayerList->item( i, 2 )->text();
 
     progressDialog.setValue( i );
-    QChar sentenceEnd = ( name.length() > 15) ? QChar( 0x2026 ) : '.';
+    QChar sentenceEnd = ( name.length() > 15 ) ? QChar( 0x2026 ) : '.';
     progressDialog.setLabelText( QObject::tr( "Searching for file: %1 \n [ %2 of %3 ] " ).arg( name.left( 15 ) + sentenceEnd,
                                  QString::number( i + 1 ), QString::number( layersToFind.size() ) ) );
     progressDialog.open();
--- a/src/app/qgshandlebadlayers.cpp
+++ b/src/app/qgshandlebadlayers.cpp
@@ -40,6 +40,7 @@
 #include <QDialogButtonBox>
 #include <QUrl>
 #include <QDir>
+#include <QProgressDialog>
 
 void QgsHandleBadLayersHandler::handleBadLayers( const QList<QDomNode> &layers )
 {
@@ -71,7 +72,6 @@ void QgsHandleBadLayersHandler::handleBadLayers( const QList<QDomNode> &layers )
   delete dialog;
   QApplication::restoreOverrideCursor();
 }
-
 
 QgsHandleBadLayers::QgsHandleBadLayers( const QList<QDomNode> &layers )
   : QDialog( QgisApp::instance() )
@@ -537,6 +537,9 @@ void QgsHandleBadLayers::autoFind()
   }
 
   const QList<int> constLayersToFind = layersToFind;
+
+  QProgressDialog progressDialog( QObject::tr( "Searching files" ), 0, 1, layersToFind.size(), this, Qt::Dialog );
+
   for ( int i : constLayersToFind )
   {
     int idx = mLayerList->item( i, 0 )->data( Qt::UserRole ).toInt();
@@ -552,6 +555,13 @@ void QgsHandleBadLayers::autoFind()
     const QString longName = dataInfo.fileName();
     QString provider = node.namedItem( QStringLiteral( "provider" ) ).toElement().text();
     const QString fileType = mLayerList->item( i, 2 )->text();
+
+    progressDialog.setValue( i );
+    QChar sentenceEnd = ( name.length() > 15) ? QChar( 0x2026 ) : '.';
+    progressDialog.setLabelText( QObject::tr( "Searching for file: %1 \n [ %2 of %3 ] " ).arg( name.left( 15 ) + sentenceEnd,
+                                 QString::number( i + 1 ), QString::number( layersToFind.size() ) ) );
+    progressDialog.open();
+
     if ( provider.toLower() == QStringLiteral( "none" ) )
     {
       if ( fileType == QStringLiteral( "raster" ) )
@@ -643,6 +653,7 @@ void QgsHandleBadLayers::autoFind()
         item->setForeground( QBrush( Qt::red ) );
       }
     }
+
   }
 
   QgsProject::instance()->layerTreeRegistryBridge()->setEnabled( false );
--- a/src/app/qgshandlebadlayers.h
+++ b/src/app/qgshandlebadlayers.h
@@ -43,7 +43,6 @@ class APP_EXPORT QgsHandleBadLayersHandler
 
 };
 
-
 class QPushButton;
 
 class APP_EXPORT QgsHandleBadLayers
@@ -93,6 +92,7 @@ class APP_EXPORT QgsHandleBadLayers
      * \since QGIS 3.12
      */
     QString checkBasepath( const QString &layerId, const QString &newPath, const QString &fileName );
+
 
 };
 


roya0045:
@m-kuhn @elpaso @nirvn Some feedback/review regarding this would be appreciated. I'm not too familiar with the UX UI side of things and would like to know if the proposed progress bar is ok.

Code is mostly taken from https://doc.qt.io/qt-5/qtwidgets-dialogs-findfiles-example.html and integrated in the code.

src/app/qgshandlebadlayers.cpp
```diff
@@ -526,6 +527,7 @@ void QgsHandleBadLayers::autoFind()
 {
   QDir::setCurrent( QgsProject::instance()->absolutePath() );
   QgsProject::instance()->layerTreeRegistryBridge()->setEnabled( true );
+¨
```
m-kuhn: I assume you did not compile this before opening this PR. Can I ask you to please do that?

m-kuhn (changes requested):

src/app/qgshandlebadlayers.cpp
roya0045: I'll try to get it done today, I borked my dev environment yesterday and I get the 'invalid CMAKE GENERATOR' after cloning my kit, and not every time I run cmake I get that error. I have no clue wtf is going on with qt creator and cmake but any changes just break everything and I'm forced to reinstall stuff until everything works again.

m-kuhn: Your work is really appreciated.
However, opening a PR for review without prior local testing may seem like asking for other's time for unfinished things.

* If you are mostly looking for conceptual inputs, please clearly state those questions and we can discuss those and no one needs to spend time on unfinished code.
* If you have a working implementation, open a pull request that compiled for you and if possible add a screencast to the PR so we can easily discuss on UX and implementation ends.
* If you have a broken development environment, try to fix it or send a mail to the dev list and work on fixing it before submitting code for review.

I hope this helps for the future.

roya0045: At this stage, as its building (finished fixing my environment just now). I mostly wanted to know the following elements:

- Is that dialog fine to use within QGIS or are there better option?

- Are there templates or other things that must be applied to those dialogs in order to maintain the QGIS look?

Otherwise I should be able to provide a gif as soon as it finishes building.

I didn't think about building at first as I thought those were only minor changes to the code base and were pretty standard and I could compile afterward as compiling takes some time in my VM.

src/app/qgshandlebadlayers.cpp
```diff
@@ -552,6 +559,11 @@ void QgsHandleBadLayers::autoFind()
     const QString longName = dataInfo.fileName();
     QString provider = node.namedItem( QStringLiteral( "provider" ) ).toElement().text();
     const QString fileType = mLayerList->item( i, 2 )->text();
+
+    progressDialog.setValue( i );
+    progressDialog.setLabelText( QObject::tr( "Searching for file %1. [ %2 of %n ]", nullptr, layersToFind.size() ).arg( name, i ) );
+    QCoreApplication::processEvents();
```
m-kuhn: `processEvents()` is dangerous and (unfortunately) can not be used without risking random crashes.

I would therefore leave this either as a pure feedback (I hope the bar is moving even without this call?) without the possibility to cancel.
In a future step you can also move things to a background task, and allow cancelling. But be prepared that this will be a thorough refactoring.

roya0045: Out of curiosity can you explain why processEvents is dangerous.

I did some test right now and everything seemed fine other than the cancel button was sometimes unresponsive to clicks but not to enter.

I was hoping that if this was not usable as is, though it is clearly suggested in the qt docs, we would have something in house that could work as an alternative, something like what is used in the attribute table (I forgot to check the code for that).

roya0045: [Here](https://github.com/qgis/QGIS/blob/209e9144f87654ac6dc11dadfc539d387d571ece/src/app/3d/qgs3danimationwidget.cpp#L209) the processEvents is used and the dialog cancel is linked to the feedback class. Is that usage better? Otherwise I could include a feedback class to pass in the finder function, this would also lead to a more prompt stop if the cancel button works properly.

roya0045: By doing a github search I can see that processEvents is used throughout QGIS, is it being phased out or kept in due to lack of alternatives?

elpaso:
> @m-kuhn @elpaso @nirvn Some feedback/review regarding this would be appreciated. I'm not too familiar with the UX UI side of things and would like to know if the proposed progress bar is ok.
> 

I didn't review the code but the progress bar looks ok to me from the UX/UI perspective, perhaps I would hide the `[ 5 of 12 ]` or move it into the bar but it's really just a matter of taste.

src/app/qgshandlebadlayers.cpp
m-kuhn: > Out of curiosity can you explain why processEvents is dangerous.

Very briefly: it processes events in unpredictable contexts (see e.g. this comment for a workaround https://github.com/qgis/QGIS/blob/master/src/gui/qgsmapcanvas.cpp#L566 , this discussion here ... https://github.com/qgis/QGIS-Enhancement-Proposals/issues/122#issuecomment-382887214)

> Here the processEvents is used and the dialog cancel is linked to the feedback class. Is that usage better?

No

> Otherwise I could include a feedback class to pass in the finder function, this would also lead to a more prompt stop if the cancel button works properly.

This will only work if threaded. And making this code threaded is involved and will likely prolong the review process and merge time.

> By doing a github search I can see that processEvents is used throughout QGIS, is it being phased out or kept in due to lack of alternatives?

Phased out slowly because writing safe, threaded code takest time

src/app/qgshandlebadlayers.cpp
elpaso: > At this stage, as its building (finished fixing my environment just now). I mostly wanted to know the following elements:
> 
>     * Is that dialog fine to use within QGIS or are there better option?
> 

I think it's fine, generally speaking always check first if is there any specific `Qgs` subclass, but in this case I think there isn't any,

>     * Are there templates or other things that must be applied to those dialogs in order to maintain the QGIS look?

No

> 
> 
> Otherwise I should be able to provide a gif as soon as it finishes building.
> 
> I didn't think about building at first as I thought those were only minor changes to the code base and were pretty standard and I could compile afterward as compiling takes some time in my VM.

This is generally speaking very bad practice, it's been already told you, please make sure your code builds correctly, pass it through prepare-commit hooks and try to wait until you've done all final adjustments before submitting a PR.

If you don't follow community member advice people will just stop reviewing your PRs.

roya0045:
> > @m-kuhn @elpaso @nirvn Some feedback/review regarding this would be appreciated. I'm not too familiar with the UX UI side of things and would like to know if the proposed progress bar is ok.
> 
> I didn't review the code but the progress bar looks ok to me from the UX/UI perspective, perhaps I would hide the `[ 5 of 12 ]` or move it into the bar but it's really just a matter of taste.

I just checked, also confused the progress bar and the dialog. But in the dialog you cannot access the progress bar to set custom text.

If its better to only show progress bar and not the dialog, then this is doable, otherwise I can only change the label in the dialog. Or always make a new progress bar and give it to the dialog. :/

src/app/qgshandlebadlayers.cpp
roya0045: I did some tests, without the process events, using either a progress bar in the base dialog or by simply removing it in this dialog.

I either case the UI just freezes and the progress bar doesn't even appear. In the case of the dialog only the window appear.

Other than multithread I don't think there is a way other than process events to keep the UI responsive.

roya0045:
@m-kuhn I kept digging and I think that to have something functional, I would have to move all the UI elements to QgsHandleBadLayersHandler to separate the UI and the functional code, transfer the slots, and run the bad layer handler in a thread with the proper connections to stop any process. Though I'm not sure how to handle the interactions with the QDatamodel.

Directly calling repaint was not helpful either.

roya0045:
Moved the demanding step (findFile function) to a separated thread to free the UI thread and make it responsive.

No success,
image

roya0045:
@m-kuhn Ok I found that calling open() was the way to go, its not perfect but it displays for demanding operations. Though the white box still occurs for a few moment at first. I cannot find a workaround at the moment.

I have also remove the concurrent execution, it changed nothing as the main thread was waiting for the results anyway.

@elpaso As stated previously I cannot comply with your demand as I think that recreating a new progress bar and setting it might not be worth it. Though I am limiting the width of the dialog and removed the cancel button. See image below. Disregard the grey loading, the window was simply not selected when I took the image.
image

Should be good to merge.

roya0045:
@m-kuhn are you ok with the proposed code/approach? I'm sure it can be improved later but I feel like this would be sufficient to prove to the user that QGIS did not crash and is active.

m-kuhn:
Yes, that looks safe code-wise.
The only small question is, if it's required to call `open()` on each iteration.

roya0045:
> Yes, that looks safe code-wise.
> The only small question is, if it's required to call `open()` on each iteration.

Sadly yes, this is the only way that I found to refresh the content of the dialog without any downsides. using repaint does not work and using exec waits for input and just stall everything.

Using open slow things down minimally as it forces the GUI thread to render the content of the box after some point. 

Though I think I need to change the minimum duration to enable the first drawing instead of having a blank window.

m-kuhn:
Ok, in this case all good.
Does your latest commit really work and did you test it? The docs say that the default value is 4000, so putting it to 500 seems to make it appear even faster.

roya0045:
> Ok, in this case all good.
> Does your latest commit really work and did you test it? The docs say that the default value is 4000, so putting it to 500 seems to make it appear even faster.

The way is usually works is that the dialog should appear after that time, but by calling open before 4000 ms means that the window will appear but it will be blank until open is called after 4000ms (~4 sec). By setting the delay before that the window will render at the first or second call to open.

I think a populated window is more comforting to the user than a blank window that only render after 4+ seconds.

I'm open to change on this, I could add an if to open instead and use a timer on the side to trigger a variable but I feel that is too much hassle.

m-kuhn:
If this is true, the only sensible option IMO would be to set this value to 0 ?

roya0045:
> If this is true, the only sensible option IMO would be to set this value to 0 ?

I went for 10, 10 microsecond should be sufficient, its not quite 0 but should be tighter. Otherwise this can be tweaked easily afterward.

m-kuhn:
Is there a reason you did not use 0?

roya0045:
> Is there a reason you did not use 0?

I was afraid 0 might cause some weird side effect or was the default value for infinite waiting until a specific element triggers it. I couldn't test zero as I'm building my c++ spatialjoin port to debug it and trying 0 would set me back 1 or two hour at least. This is why I went the safe way.

roya0045:
@m-kuhn did more tests, param ended up not changing anything after all, even when forcing open it seems to delay rendering until it has enough time to render no matter what.

People will have to live with a blank box for a few seconds.

roya0045:
@m-kuhn Since I cannot force rendering afterall I think this will have to do for now. Are you ok with this?

src/app/qgshandlebadlayers.cpp
```diff
@@ -552,6 +555,12 @@ void QgsHandleBadLayers::autoFind()
     const QString longName = dataInfo.fileName();
     QString provider = node.namedItem( QStringLiteral( "provider" ) ).toElement().text();
     const QString fileType = mLayerList->item( i, 2 )->text();
+
+    progressDialog.setValue( i );
+    progressDialog.setLabelText( QObject::tr( "Searching for file: %1. \n [ %2 of %3 ] " ).arg( name.left( 15 ) + "...",
```
m-kuhn: The text here seems to need some fine tuning.

- I'm not sure you wanted to add the three dots in every case (also if the file name is shorter than 15 chars)
- There will right now always be four dots (three appended to the name and one specified in the label template)
- If you stick to the ellipsis use a real ellipsis (`…` instead of `...`)
- For the ellipsis, use a single char (`'…'`)
- I'd probably drop the period `.` from the sentence

m-kuhn (commented):
Yep, looks mergable to me in this case
I left one comment for the label

src/app/qgshandlebadlayers.cpp
roya0045: I went for the distinct dots due to the issues with the ellipsis converting to weird characters, I could use the fix provided by Nyall and directly reference the character code instead of having the ellipsis. 

I also thought about not having them for layers with shorter name but I felt they did not only indicate shortening but also progress/pending action. I'm open to remove them if you feel it is more appropriate.

m-kuhn: In principle the ellipsis is good to depict truncated names, so using the character code sounds good.

roya0045: Should be fixed now. Had some issues with the layout test and whitespaces, should be all good this time.

roya0045:
Unrelated segfault in rule-based renderer test( how did that even happen?)

roya0045:
@m-kuhn I'm not sure how to check the failing test in the azure build but it works fine in my VM. Are you ok to merge this or do the azure tests need to be sorted out first?

roya0045:
@m-kuhn Thanks for the help with this PR!

PR #33276, opened 2019-12-07
More labeling obstacle refactoring, cleanups

nyalldawson:
Making it easier for future extensions to obstacle handling


nirvn (approved):
Makes sense.

PR #33278, opened 2019-12-07
Fix some warnings on invalid layers

m-kuhn:
When loading old projects that were saved without geometry type in the xml, invalid layers' geometry types cannot be resolved and some warnings are popping up on debug builds.

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -13402,7 +13402,8 @@ void QgisApp::activateDeactivateLayerRelatedActions( QgsMapLayer *layer )
         mActionAddFeature->setText( addFeatureText );
         mActionAddFeature->setToolTip( addFeatureText );
         QgsGui::shortcutsManager()->unregisterAction( mActionAddFeature );
-        QgsGui::shortcutsManager()->registerAction( mActionAddFeature, mActionAddFeature->shortcut() );
+        if ( !mActionAddFeature->text().isEmpty() ) // The text will be empty on unkown geometry type -> in this case do not create a shortcut
+          QgsGui::shortcutsManager()->registerAction( mActionAddFeature, mActionAddFeature->shortcut() );
       }
       else
       {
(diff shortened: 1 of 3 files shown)


src/app/qgisapp.cpp
```diff
@@ -13402,7 +13402,8 @@ void QgisApp::activateDeactivateLayerRelatedActions( QgsMapLayer *layer )
         mActionAddFeature->setText( addFeatureText );
         mActionAddFeature->setToolTip( addFeatureText );
         QgsGui::shortcutsManager()->unregisterAction( mActionAddFeature );
-        QgsGui::shortcutsManager()->registerAction( mActionAddFeature, mActionAddFeature->shortcut() );
+        if ( !mActionAddFeature->text().isEmpty() ) // The text will be empty on unkown geometry type -> in this case do not create a shortcut
```
m-kuhn:
```suggestion
        if ( !mActionAddFeature->text().isEmpty() ) // The text will be empty on unknown geometry type -> in this case do not create a shortcut
```

src/core/layout/qgslayoutitemattributetable.cpp
```diff
@@ -799,7 +799,9 @@ bool QgsLayoutItemAttributeTable::readPropertiesFromElement( const QDomElement &
   mVectorLayer.resolveWeakly( mLayout->project() );
 
   //connect to new layer
-  connect( sourceLayer(), &QgsVectorLayer::layerModified, this, &QgsLayoutTable::refreshAttributes );
+  QgsVectorLayer *newLayer = sourceLayer();
```
nyalldawson: Minor thing, but in cases like this I think it's nicer to condense into a single line so that the variable has a tight scope:

```
if ( QgsVectorLayer* new layer = ... )
```

nyalldawson (approved):

src/core/layout/qgslayoutitemattributetable.cpp
```diff
@@ -799,7 +798,8 @@ bool QgsLayoutItemAttributeTable::readPropertiesFromElement( const QDomElement &
   mVectorLayer.resolveWeakly( mLayout->project() );
 
   //connect to new layer
-  connect( sourceLayer(), &QgsVectorLayer::layerModified, this, &QgsLayoutTable::refreshAttributes );
+  if ( QgsVectorLayer *newLayer = sourceLayer() )
+    connect( sourceLayer(), &QgsVectorLayer::layerModified, this, &QgsLayoutTable::refreshAttributes );
```
m-kuhn:
```suggestion
    connect( newLayer, &QgsVectorLayer::layerModified, this, &QgsLayoutTable::refreshAttributes );
```

PR #33279, opened 2019-12-07
Fail expression evaluation if missing field is referenced

m-kuhn:
When evaluating an expression with a missing referenced field, this was be replaced with a string `[fieldname]`. But `expression.prepare()` would fail already, so this just worked on unprepared expressions.

When calculating data defined values, this will be counterproductive.
A field with a data defined value might work well in the expression builder (because fiels are all present) but fail in the legend, because it's rendered with no feature and hence fields. With this change, we can at least use `try("fieldname", 0)` to fallback to a default value.

### DIFF
--- a/python/core/auto_generated/qgsexpressioncontext.sip.in
+++ b/python/core/auto_generated/qgsexpressioncontext.sip.in
@@ -715,7 +715,7 @@ Convenience function for retrieving the fields for the context, if set.
 %Docstring
 Sets the original value variable value for the context.
 
-:param value: value for original value variable. This usually represents the an original widget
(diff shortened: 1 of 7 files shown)


m-kuhn:
Ouch, plenty of tests rely on this behavior.
Now the question is... On purpose?
Eager to hear other's opinion on this topic ...

nyalldawson:
I've been confused/frustrated by this too in the past. I'd be happy to see the layout test behavior changed to match what you've proposed.

nirvn:
I have no strong position here, but since we know have try(), I think the change (behavior harmonization between prepared and unprepared to be exact) sounds logical.

PR #33282, opened 2019-12-08
When DD rotation fails to evaluate, fallback to static value

m-kuhn:
# Changelog entry

The legend finally also works nicely with data defined rotation. QGIS can already apply data defined rotation to symbols and individual symbol layers for a long time already, but this resulted in a broken legend so far. The rotation of symbols (and symbol parts defined on symbol layers) on the legend can now be controlled by changing the default values next to the data defined properties.

----------------

# PR Description

The old behavior was to always fall back to 0 in case of a failing expression. Which made it hard to create a proper legend in some cases.

E.g. if you had two symbol layers with the following expressions for their rotation

    "orientation"

And

    90 + "orientation"

which would be two orthogonal symbol layers (e.g. denote a right angle), rotated by an orientation field.

Now in the legend (and symbol preview), both layers would be shown with a rotation of 0, because there is no associated feature with a field orientation.

The new behavior is to fallback to the static, configured values which makes it very intuitive to configure the legend for these cases.

@nyalldawson do you see any unintended side effects of this (I guess travis is going to tell me as well ;)

# Before
Peek 2019-12-08 09-18

# After

Peek 2019-12-08 09-16


nyalldawson:
I think your refinement is the correct desired behavior. We fallback to the value set in the gui in the case of null values, and I think invalid expressions should be treated the same. (0 is totally arbitrary in this case anyway...!)

nyalldawson (approved):

nyalldawson:
On this topic... I keep thinking we really need an option to disable the automatic rotation of offsets when symbol rotation is set 🤔

nirvn:
@nyalldawson , as in rotate symbol prior to applying offset?

nyalldawson:
@nirvn yep, keep them as totally independent settings

nirvn:
Interesting idea.

PR #33283, opened 2019-12-08
Relation style restore

elpaso:
This is a followup to https://github.com/qgis/QGIS/pull/33103 that makes it possible to automatically load related layers (referenced layers in a relation) independently from the widget configuration.

The implementation is based on a new class `QgsWeakRelation` that stored weak references to related layers, the QGIS application can use this information to automatically load related layers and resolve `QgsRelation` objects on the fly.

Relations information is stored as an additional category in layer styles.

Funded by **ARPA Piemonte**

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -678,13 +678,13 @@ void QgisApp::vectorLayerStyleLoaded( QgsMapLayer::StyleCategories categories )
     // Check broken dependencies in forms
     if ( categories.testFlag( QgsMapLayer::StyleCategory::Forms ) )
     {
-      checkVectorLayerDependencies( vl );
+      resolveVectorLayerDependencies( vl );
     }
 
     // Check broken relations and try to restore them
     if ( categories.testFlag( QgsMapLayer::StyleCategory::Relations ) )
     {
-      checkVectorLayerRelations( vl );
+      resolveVectorLayerWeakRelations( vl );
     }
 
   }
@@ -2008,73 +2008,78 @@ QgsMessageBar *QgisApp::visibleMessageBar()
   }
 }
 
-const QList<QgsVectorLayerRef> QgisApp::findBrokenLayerDependencies( QgsVectorLayer *vl ) const
+const QList<QgsVectorLayerRef> QgisApp::findBrokenLayerDependencies( QgsVectorLayer *vl, QgsMapLayer::StyleCategories categories ) const
 {
   QList<QgsVectorLayerRef> brokenDependencies;
 
-  // Check for missing layer widget dependencies
-  for ( int i = 0; i < vl->fields().count(); i++ )
+  if ( categories.testFlag( QgsMapLayer::StyleCategory::Forms ) )
   {
-    const QgsEditorWidgetSetup setup = QgsGui::editorWidgetRegistry()->findBest( vl, vl->fields().field( i ).name() );
-    QgsFieldFormatter *fieldFormatter = QgsApplication::fieldFormatterRegistry()->fieldFormatter( setup.type() );
-    if ( fieldFormatter )
+    for ( int i = 0; i < vl->fields().count(); i++ )
     {
-      const QList<QgsVectorLayerRef> constDependencies { fieldFormatter->layerDependencies( setup.config() ) };
-      for ( const QgsVectorLayerRef &dependency : constDependencies )
+      const QgsEditorWidgetSetup setup = QgsGui::editorWidgetRegistry()->findBest( vl, vl->fields().field( i ).name() );
+      QgsFieldFormatter *fieldFormatter = QgsApplication::fieldFormatterRegistry()->fieldFormatter( setup.type() );
+      if ( fieldFormatter )
       {
-        const QgsVectorLayer *depVl { QgsVectorLayerRef( dependency ).resolveWeakly(
-                                        QgsProject::instance(),
-                                        QgsVectorLayerRef::MatchType::Name ) };
-        if ( ! depVl || ! depVl->isValid() )
+        const QList<QgsVectorLayerRef> constDependencies { fieldFormatter->layerDependencies( setup.config() ) };
+        for ( const QgsVectorLayerRef &dependency : constDependencies )
         {
-          brokenDependencies.append( dependency );
+          const QgsVectorLayer *depVl { QgsVectorLayerRef( dependency ).resolveWeakly(
+                                          QgsProject::instance(),
+                                          QgsVectorLayerRef::MatchType::Name ) };
+          if ( ! depVl || ! depVl->isValid() )
+          {
+            brokenDependencies.append( dependency );
+          }
         }
       }
     }
   }
 
-  // Check for layer weak relations
-  const QList<QgsWeakRelation> constWeakRelations { vl->weakRelations() };
-  for ( const QgsWeakRelation &rel : constWeakRelations )
+  if ( categories.testFlag( QgsMapLayer::StyleCategory::Relations ) )
   {
-    QgsRelation relation { rel.resolvedRelation( QgsProject::instance(), QgsVectorLayerRef::MatchType::Name ) };
-    QgsVectorLayerRef dependency;
-    bool found = false;
-    if ( ! relation.isValid() )
+    // Check for layer weak relations
+    const QList<QgsWeakRelation> constWeakRelations { vl->weakRelations() };
+    for ( const QgsWeakRelation &rel : constWeakRelations )
     {
-      // We need just the other side of the relation
-      if ( relation.referencedLayer() == vl )
+      QgsRelation relation { rel.resolvedRelation( QgsProject::instance(), QgsVectorLayerRef::MatchType::Name ) };
+      QgsVectorLayerRef dependency;
+      bool found = false;
+      if ( ! relation.isValid() )
       {
-        dependency = rel.referencingLayer();
-        found = true;
-      }
-      else if ( relation.referencingLayer() == vl )
-      {
-        dependency = rel.referencedLayer();
-        found = true;
-      }
-      else
-      {
-        // Something wrong is going on here, maybe this relation
-        // does not really apply to this layer?
-        QgsMessageLog::logMessage( tr( "None of the layers in the relation stored in the style match the current layer, skipping relation id: %1." ).arg( relation.id() ) );
-      }
-
-      if ( found )
-      {
-        // Make sure we don't add it twice
-        bool refFound = false;
-        for ( const QgsVectorLayerRef &otherRef : qgis::as_const( brokenDependencies ) )
+        // We need just the other side of the relation
+        if ( relation.referencedLayer() == vl )
         {
-          if ( dependency.layerId == otherRef.layerId || ( dependency.source == otherRef.source && dependency.provider == otherRef.provider ) )
-          {
-            refFound = true;
-            break;
-          }
+          dependency = rel.referencingLayer();
+          found = true;
         }
-        if ( ! refFound )
+        else if ( relation.referencingLayer() == vl )
         {
-          brokenDependencies.append( dependency );
+          dependency = rel.referencedLayer();
+          found = true;
+        }
+        else
+        {
+          // Something wrong is going on here, maybe this relation
+          // does not really apply to this layer?
+          QgsMessageLog::logMessage( tr( "None of the layers in the relation stored in the style match the current layer, skipping relation id: %1." ).arg( relation.id() ) );
+        }
+
+        if ( found )
+        {
+          // Make sure we don't add it twice
+          bool refFound = false;
+          for ( const QgsVectorLayerRef &otherRef : qgis::as_const( brokenDependencies ) )
+          {
+            if ( dependency.layerId == otherRef.layerId || ( dependency.source == otherRef.source && dependency.provider == otherRef.provider ) )
+            {
+              refFound = true;
+              break;
+            }
+          }
+          if ( ! refFound )
+          {
+            brokenDependencies.append( dependency );
+          }
         }
       }
     }
@@ -2082,11 +2087,11 @@ const QList<QgsVectorLayerRef> QgisApp::findBrokenLayerDependencies( QgsVectorLa
   return brokenDependencies;
 }
 
-void QgisApp::checkVectorLayerDependencies( QgsVectorLayer *vl )
+void QgisApp::resolveVectorLayerDependencies( QgsVectorLayer *vl, QgsMapLayer::StyleCategories categories )
 {
   if ( vl && vl->isValid() )
   {
-    const auto constDependencies { findBrokenLayerDependencies( vl ) };
+    const auto constDependencies { findBrokenLayerDependencies( vl, categories ) };
     for ( const QgsVectorLayerRef &dependency : constDependencies )
     {
       // try to aggressively resolve the broken dependencies
@@ -2176,7 +2181,7 @@ void QgisApp::checkVectorLayerDependencies( QgsVectorLayer *vl )
   }
 }
 
-void QgisApp::checkVectorLayerRelations( QgsVectorLayer *vectorLayer )
+void QgisApp::resolveVectorLayerWeakRelations( QgsVectorLayer *vectorLayer )
 {
   if ( vectorLayer && vectorLayer->isValid() )
   {
@@ -6592,7 +6597,7 @@ bool QgisApp::addProject( const QString &projectFile )
     {
       if ( vl->isValid() )
       {
-        checkVectorLayerDependencies( vl );
+        resolveVectorLayerDependencies( vl );
       }
     }
 
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -2046,11 +2046,18 @@ const QList<QgsVectorLayerRef> QgisApp::findBrokenLayerDependencies( QgsVectorLa
       bool found = false;
       if ( ! relation.isValid() )
       {
-        // We need just the other side of the relation
+        // This is the big question: do we really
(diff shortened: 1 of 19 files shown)


src/app/qgisapp.cpp
```diff
@@ -1988,39 +2001,101 @@ QgsMessageBar *QgisApp::visibleMessageBar()
...
+      QgsVectorLayerRef dependency;
+      bool found = false;
+      if ( ! relation.isValid() )
       {
-        const QgsVectorLayer *depVl { QgsVectorLayerRef( dependency ).resolveWeakly(
-                                        QgsProject::instance(),
-                                        QgsVectorLayerRef::MatchType::Name ) };
-        if ( ! depVl || ! depVl->isValid() )
+        // This is the big question: do we really
+        // want to automatically load the referencing layer(s) too?
```
nyalldawson: I think yes, we do...

nyalldawson: can we avoid the confusing referencing/referenced terminology and stick with the standard parent/child ? I know the api uses the confusing versions, but we can fix that easily for 4.0... comments will be harder to find and fix.

src/app/qgisapp.cpp
```diff
@@ -1988,39 +2001,101 @@ QgsMessageBar *QgisApp::visibleMessageBar()
   }
 }
 
-QList<QgsVectorLayerRef> QgisApp::findBrokenWidgetDependencies( QgsVectorLayer *vl )
+const QList<QgsVectorLayerRef> QgisApp::findBrokenLayerDependencies( QgsVectorLayer *vl, QgsMapLayer::StyleCategories categories ) const
```
nyalldawson: if this logic grows much more, i'd suggest we move it out to a new class in app (qgisapp.cpp is already embarrassingly huge!)

src/app/qgisapp.h
```diff
@@ -2020,15 +2020,31 @@ class APP_EXPORT QgisApp : public QMainWindow, private Ui::MainWindow
     QgsMessageBar *visibleMessageBar();
 
     /**
-     * Searches for layer widget dependencies
-     * \return a list of weak references to broken widget layer dependencies
+     * Searches for layer dependencies by querying the form widgets and the
+     * \a vectorLayer itself for broken relations. Style \a categories can be
+     * used to exclude one of the currently implemented search categories
```
nyalldawson: exclude? or include?

src/core/qgsweakrelation.h
```diff
@@ -0,0 +1,113 @@
...
+                   );
+
+    /**
+     * Resolves a weak relation in the given \a project returning a possibly invalid QgsRelation
+     * and without performing any kind of validity check.
+     *
+     * \note Client code should never assume that the returned relation is valid and the
+     * layer components are not NULL.
+     */
+    QgsRelation resolvedRelation( const QgsProject *project, QgsVectorLayerRef::MatchType matchType = QgsVectorLayerRef::MatchType::All ) const;
```
nyalldawson: resolveRelation?

src/core/qgsweakrelation.h
```diff
@@ -0,0 +1,113 @@
...
+     */
+    QList<QgsRelation::FieldPair> fieldPairs() const;
+
+  private:
+
+    QgsVectorLayerRef mReferencingLayer;
+    QgsVectorLayerRef mReferencedLayer;
+    QString mRelationId;
+    QString mRelationName;
+    QgsRelation::RelationStrength mStrength;
```
nyalldawson: initialize

nyalldawson (commented):
Looks **great** to me. Nice work!

(Better get @m-kuhn's opinion too before merge)

src/app/qgisapp.cpp
m-kuhn: For the record, the terminology was chosen on purpose. Parent/child is appropriate when we talk about compositions, but is misleading on aggregations and mostly on associations.

src/app/qgisapp.h
elpaso: both :) Let's use "specify"

src/core/qgsweakrelation.h
elpaso: Actually, it's not a typo: I thought that since the method does not change the status of the object and returns a new `QgsRelation` the past participle was more appropriate, but I do not insist.

src/app/qgisapp.cpp
elpaso: Consider this (real) use case: 200 data layers in a PG DB, every data layer uses a small set of referenced aspatial layers (like categories etc.).
The use loads one data layer, the referenced layers is loaded, all other 200 referencing layers are automatically loaded. This is what scares me.

src/app/qgisapp.h
```diff
@@ -2025,15 +2025,31 @@ class APP_EXPORT QgisApp : public QMainWindow, private Ui::MainWindow
...
-     * \return a list of weak references to broken widget layer dependencies
+     * Searches for layer dependencies by querying the form widgets and the
+     * \a vectorLayer itself for broken relations. Style \a categories can be
+     * used to limit the search to one or more of the currently implemented search
+     * categories ("Forms" for the form widgets and "Relations" for layer weak relations).
+     * \return a list of weak references to broken layer dependencies
      */
-    QList< QgsVectorLayerRef > findBrokenWidgetDependencies( QgsVectorLayer *vectorLayer );
+    const QList< QgsVectorLayerRef > findBrokenLayerDependencies( QgsVectorLayer *vectorLayer,
+        QgsMapLayer::StyleCategories categories = QgsMapLayer::AllStyleCategories ) const;
```
```suggestion
        QgsMapLayer::StyleCategories categories = QgsMapLayer::StyleCategory::AllStyleCategories ) const;
```

If you modify existing code you should at least do it consistently in your new code too ;)

src/core/qgsweakrelation.h
```diff
@@ -0,0 +1,113 @@
...
+     * \param strength relation strength
+     * \param referencingLayerId ID of the referencing layer
+     * \param referencingLayerName name of the referencing layer
+     * \param referencingLayerSource source of the referencing layer
+     * \param referencingLayerProviderKey provider name of the referencing layer
+     * \param referencedLayerId ID of the referenced layer
+     * \param referencedLayerName name of the referenced layer
+     * \param referencedLayerSource name of the referenced layer
+     * \param referencedLayerProviderKey provider name of the referenced layer
+     * \param fieldPairs list of the relation field pairs
```
m-kuhn: I'd propose to remove all those "stating the obvious" param descriptions and `\a` the non obvious / important ones in the inline comment.

src/core/qgsrelationmanager.cpp
```diff
@@ -27,8 +27,11 @@ QgsRelationManager::QgsRelationManager( QgsProject *project )
 {
   if ( mProject )
   {
+    // TODO: QGIS 4 remove: relations are now stored with the layer style
```
m-kuhn: With which style, the referencing or the referenced layer?

m-kuhn (commented):
Very valuable work. I have just left some comments that jumped my eye while skimming through the code.

src/core/qgsweakrelation.h
elpaso: IMO they are all so obvious, when I wrote this doxy I was thinking: what a waste of time!

m-kuhn: Kill 'em!

src/core/qgsrelationmanager.cpp
elpaso: both: I wanted to be able to reconstruct relations in both ways, but I've only implemented automatic load of referenced layers for now, see the "big question" discussion above: https://github.com/qgis/QGIS/pull/33283#pullrequestreview-328690776

So, for the time being we have some redundancy, but I guess we can live with it, given the potential advantages.

src/core/qgsweakrelation.h
elpaso: Would code-layout agree?

m-kuhn: If you kill all of them, yes.

src/core/qgsrelationmanager.cpp
m-kuhn: Very good points there!

What about a new flag on the relation to make it "soft" or "strong". Where strong relations always try to load the dependency and soft ones are taken as a hint to which appear grayed out first/invalid, and only "fully appear" once both layers are present?

elpaso: I considered the option to implement "weakness" directly inside `QgsRelation`, turning `QgsVectorLayer`s to weak references. 

However having a wrapper looked more clean to me and first of all much simpler, because the current `QgsRelation` (and manager) implementation has a strong precondition on relation validity (if I'm not wrong invalid relations are not added to the manager and removed as soon as the layer becomes invalid).

m-kuhn: I was more talking about a user configurable option, saved as a boolean in the xml.

elpaso: I don't follow, an option for what?

m-kuhn: As you write, sometimes, one wants to only work with only a part of a huge data model. And if loading a style, possibly not resolve all relations (and load all layers).

Where to draw the line between "desired" relations (value lists, child classes, etc) and "other, currently uninteresting parts of the model" could be done by tagging some relations as strong (yes, load) and soft (only apply if other layer is already loaded).

But that's not the only way to deal with it, just thinking out loud.

elpaso: Got it, this is actually a good idea, but isn't the "strength" already meant carry that information?

Anyway my "big question" was more about the referenced layers case, I've no doubts that whenever a referencing layer is loaded its referenced counterpart is a strong dependency and must be automatically loaded or the model would be incomplete (and probably some widgets broken).

On the other hand, I'm not sure that we want to automatically load the (potentially many) referencing layers when a referenced layer is loaded.

Anyway all this machinery is controlled by a style that has been saved as a (default) style in the context of a project where relations have been explicitly set by the user, so there is already some control over what gets automatically loaded and what not.

m-kuhn: Maybe yes. Treat compositions (and aggregations?) as "auto load" and associations as "load on demand" could work. But that's very weak ground I'm basing this assumption on. Reality will probably prove me wrong.

elpaso:
@m-kuhn @nyalldawson I'd like to merge this as it is (after I make some more tests for m2m), even if I still have some doubts about what should automatically loaded and what should not.

The current approach is to (try to) automatically load what will make functionality broken (such as a layer referenced by a FK->PK relation in association or compositions) and layers used in a value-relation widget. Joins are not yet implemented but it could be added easily.

If we want the user to have full control,  adding a dialog and ask the user directly if she want to load related layers or not seems the way to go, but I'm hesitant to add add another dialog, I don't see a use case for not loading automatically layer dependencies and have a broken relation/widget.

@m-kuhn idea to have a flag is good, but again, adding another flag/enum to the relation API when we already have a "strength" looks like a bit redundant to me, on the other hand I think most of the people will not change the default strength (association) and for this reason NOT loading associations automatically could lead to a situation where this automatically-load implementation is potentially underused.

I propose to merge it now as it is and collect some feedback in time for 3.12 (hoping that someone tests this).

m-kuhn:
@elpaso that works for me :+1: 

> @m-kuhn idea to have a flag is good, but again, adding another flag/enum to the relation API when we already have a "strength" looks like a bit redundant to me, on the other hand I think most of the people will not change the default strength (association) and for this reason NOT loading associations automatically could lead to a situation where this automatically-load implementation is potentially underused.

Yes, I think we should not mix up the two types of "strength", one is on project level, the other on feature level.

nyalldawson:
> If we want the user to have full control, adding a dialog and ask the user directly if she want to load related layers or not seems the way to go, but I'm hesitant to add add another dialog, 

Agreed- let's not do a dialog. That would introduce all sorts of potentially dangerous/fragile event loop mess!

> I don't see a use case for not loading automatically layer dependencies and have a broken relation/widget.

Also agreed - that's why I think just automatically loading **everything** is the right approach, even if that's 100+ related layers.

m-kuhn:
> Agreed- let's not do a dialog. That would introduce all sorts of potentially dangerous/fragile event loop mess!

I don't think anyone was discussing a dialog on load (right?)

Off topic, but speaking about ... Every modal dialog we have is running a local event loop inside. Personally I never had trouble with local event loops and only with processEvent calls. Are local event loops really that dangerous given that any modal dialog (like layer properties etc) is running them? Or do I get something completely wrong right now.

elpaso:
> > Agreed- let's not do a dialog. That would introduce all sorts of potentially dangerous/fragile event loop mess!
> 
> I don't think anyone was discussing a dialog on load (right?)

Not really, but it's something I was thinking about loudly, forget it.

> 
> Off topic, but speaking about ... Every modal dialog we have is running a local event loop inside. Personally I never had trouble with local event loops and only with processEvent calls. Are local event loops really that dangerous given that any modal dialog (like layer properties etc) is running them? Or do I get something completely wrong right now.

I think that local event loops are not inherently dangerous in general (neither processEvents is), the signal/slot connections mechanism is designed take care of inter-thread communication, the problems arise when slots have side effects (like deleting some shared objects, like map layers and such.), my gut feeling is that in these cases there is an architectural problem somewhere else.

So, the rule of thumb for our codebase is still: "avoid them if you can", but we should eventually be more careful about shared objects life cycle and try to make the code safer and more resilient to such side effects.

m-kuhn:
Let me rephrase my question.

I do know that calling processEvents() is dangerous (in combination with the current QGIS codebase which is not gonna be sanitized in the way you describe in the foreseeable future), I have seen crashes related to this more than once.

I have my doubts a local QEventLoop is as dangerous, because then - I assume - we'd see more crashes while modal dialogs are open.

The question therefore is: are local QEventLoops really as dangerous as a call to processEvents() is. And if yes, why do our modal dialogs get away with it.

nyalldawson:
>  Are local event loops really that dangerous given that any modal dialog (like layer properties etc) is running them

Yes -- depending on the context. For instance:
1. https://github.com/qgis/QGIS/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+wfs+crash -- is caused by a local QEventLoop running on the main thread in the WFS provider. The crash occurs because when the legend item is painted it needs to fetch some features from the provider. The local event loop (there's actually no such thing -- let's call it what it really is, a main thread event loop) runs as part of the WFS feature fetching, which means that the scene painting event gets called again and when the WFS event loop returns control back to the legend renderer and it tries to paint, the paint device it was trying to paint to no longer exists. bam!

2. I've seen many crashes occur in the past (and maybe still on master?) where a dialog is triggered as part of the QgsProject loading process. This results in a similar thing -- the event loop used by the modal dialog means that things are changing underneath the project loading function, and when it gets control back things have been deleted and a crash results.

@elpaso 
> I think that local event loops are not inherently dangerous in general (neither processEvents is), the signal/slot connections mechanism is designed take care of inter-thread communication,

correct, but often the problems with event loops are extremely subtle. A modal dialog using an event loop is no problem, right up until that dialog gets shown in the middle of some other process which IS sensitive to having the event loop re-run. It's rather difficult to determine the exact consequences of inserting an event loop, so that's why I prefer to see them avoided everywhere they could be run on the main thread.

> my gut feeling is that in these cases there is an architectural problem somewhere else.

Not necessarily -- in the case of the WFS provider, the WFS provider is clearly at fault and not the legend renderer. The legend renderer correctly calls getFeatures on the main thread, and assumes that the provider is well behaved and isn't triggering a processEvents/event loop. In this case, the WFS provider needs reworking to use a semaphore based approach or similar and not rely on a local event loop.

@m-kuhn 
> I do know that calling processEvents() is dangerous (in combination with the current QGIS codebase which is not gonna be sanitized in the way you describe in the foreseeable future), I have seen crashes related to this more than once. I have my doubts a local QEventLoop is as dangerous, because then - I assume - we'd see more crashes while modal dialogs are open.

It's actually completely equivalent. A local QEventLoop running on the main thread is the **same** thing as a bunch of processEvents calls.

m-kuhn:
Thanks @nyalldawson and @elpaso 

Given

> often the problems with event loops are extremely subtle. A modal dialog using an event loop is no problem, **right up until that dialog gets shown in the middle of some other process which IS sensitive to having the event loop re-run**.

combined with the following trace from an open layer properties window.

image

I'd expect random crashes to happen more often, so do we have an answer to

> And if yes, how do our modal dialogs get away with it?

nyalldawson:
Well, that trace is fine -- the loop is running inside layerTreeViewDoubleClicked, which isn't sensitive to a processEvents call. 

Things which I know definitely ARE sensitive are:
1. During project loading
2. During rendering operations (canvas or layouts)
3. During any item painting code

(there's likely more, but those have definitely caused crashes in the past)

m-kuhn:
Let me ask the question in a different way then.

If a QEventLoop (on the main thread) should be highlighted by a pull request reviewer, as a consequence should the same apply to any call to `QDialog::exec()` or `QFieldDialog::getOpenFileName()` or `QMessageBox::warning` or plenty of other calls that act as a wrapper around a local event loop?

nyalldawson:
Well, that totally depends on the context. If there's a risk that the dialog will get shown during project loading, then yes!

m-kuhn:
And vice versa, if there is no risk of running in this context, it's also fine to approve a QEventLoop?

nyalldawson:
Yep

m-kuhn:
Thanks :+1:

roya0045:
I assume that using processEvents() would have been fine for the progressdialog used with the autofinder in this case? 

I'm fine with open at the moment but if users says that the blank box alarms them its good to know that we can use processEvents() instead.

m-kuhn:
Looks like, yes.

PR #33288, opened 2019-12-09
Death to pal::RTree

nyalldawson:
This PR drops to ancient, unmaintained, custom pal::RTree spatial index and replaces it with a backend utilising libspatialindex. This brings it into line with the rest of QGIS, which heavily uses libspatialindex, and removes a couple of thousand lines of totally maintainable code. 

(Possibly there's a speed boost because of this too, but the original rtree code never took a significant amount of time in the context of label problem solving anyway...)


NathanW2:
So long pal::RTree

nirvn:
I can't pretend do be in a position to do a proper review of all of this code, so I wont. That said, I've compiled the branch, and tested a couple of my label-heavy projects, and happy to report I can't spot regressions. Nice to see over 2,000 line of codes going away too.

nyalldawson:
@nirvn thanks for the testing -- it's certainly appreciated/valuable!

PR #33290, opened 2019-12-09
[processing] port some algorithms to C++

alexbruy:
## Description
Port several QGIS alorithms to C++:
- Extract specific vertices
- Geometry by expression
- Nearest neighbour analysis + add test for it
- Points (pixel centroids) from polygons
- Pole of inaccessibility
- Raster layer statistics
- Set raster style and Set vector style was replaced with single generic algorithm Set layer style. Old algorithms still here to keep backward compatibility but hidden from toolbox
- Truncate table
- Zonal statistics + add helper methods to `QgsZonalStatistics` to retrieve user-friendly statistics name from enum value
- Random extract. Also fixed unreported issue which caused incomplete results when generated random numbers list contains duplicated values. In such case old implementation outputs feature only once producing less features than requested.

## Checklist


root676:
@alexbruy this is **huge**! Wow!

python/plugins/processing/algs/qgis/SetRasterStyle.py
```diff
@@ -48,7 +48,7 @@ def __init__(self):
         super().__init__()
 
     def flags(self):
-        return super().flags() | QgsProcessingAlgorithm.FlagNoThreading
+        return super().flags() | QgsProcessingAlgorithm.FlagNoThreading | QgsProcessingAlgorithm.FlagHideFromToolbox
```
nyalldawson: The explicit "FlagDeprecated" would be more appropriate here (it also hides from the modeler, and would allow us to easily identify deprecated algorithms in future...)

src/analysis/processing/qgsalgorithmextractspecificvertices.h
```diff
@@ -0,0 +1,56 @@
...
+#include "qgis_sip.h"
+#include "qgsprocessingalgorithm.h"
+#include "qgsapplication.h"
+
+///@cond PRIVATE
+
+/**
+ * Native extract specific vertices algorithm.
+ */
+class QgsExtractSpecificVerticesAlgorithm : public QgsProcessingAlgorithm
```
nyalldawson: Hm, I wonder why we never made this one a feature based algorithm. Probably just an oversight. (We can address this later anyway)

src/analysis/processing/qgsalgorithmnearestneighbouranalysis.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+QString QgsNearestNeighbourAnalysisAlgorithm::groupId() const
+{
+  return QStringLiteral( "vectoranalysis" );
+}
+
+QString QgsNearestNeighbourAnalysisAlgorithm::shortHelpString() const
+{
+  return QObject::tr( "This algorithm performs nearest neighbor analysis for a point layer.\n\n"
+                      "Output is generated as an html file with the computed statistical values." );
```
nyalldawson:
```suggestion
                      "Output is generated as an HTML file with the computed statistical values." );
```

src/analysis/processing/qgsalgorithmnearestneighbouranalysis.cpp
```diff
@@ -0,0 +1,155 @@
...
+  int i = 0;
+  QgsFeature f;
+  while ( it.nextFeature( f ) )
+  {
+    if ( feedback->isCanceled() )
+    {
+      break;
+    }
+
+    QgsFeatureId neighbourId = spatialIndex.nearestNeighbor( f.geometry().asPoint(), 2 ).at( 1 );
```
nyalldawson: Hm, we can potentially give this algorithm a huge speed boost -- if we use the FlagStoreFeatureGeometries during the construction of the spatial index, then the geometries will be stored inside the algorithm and can be later retrieved using QgsSpatialIndex::geometry( ... ) . This would avoid the need for all the extra single-feature requests.

src/analysis/processing/qgsalgorithmpoleofinaccessibility.h
```diff
@@ -0,0 +1,56 @@
...
+
+#include "qgis_sip.h"
+#include "qgsprocessingalgorithm.h"
+
+///@cond PRIVATE
+
+/**
+ * Native pole of inaccessibility algorithm.
+ */
+class QgsPoleOfInaccessibilityAlgorithm : public QgsProcessingAlgorithm
```
nyalldawson: Here's another one we should have made a feature based alg.

src/analysis/processing/qgsalgorithmrandomextract.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+  QgsFeature f;
+  for ( QgsFeatureId id : fids )
+  {
+    if ( feedback->isCanceled() )
+    {
+      break;
+    }
+
+    source->getFeatures( QgsFeatureRequest( id ), QgsProcessingFeatureSource::FlagSkipGeometryValidityChecks ).nextFeature( f );
```
nyalldawson: One by one feature requests are always quite slow. Looks like the original alg looped through the whole layer to avoid this. So while the new code will be much faster for low selection counts, when the count get close to the feature count it'll be much slower. 

I think you could avoid this by creating a single feature request with a .setFilterFids on it, and iterating through that. Note that you'd need to count the number of times that feature appears in the fids list and add that many duplicate features though (or you'll reintroduce the original bug!)

On that note, I think it might be worthwhile exposing a enum choice for sampling behaviour -- i.e. "sample with replacement" or "sample without replacement".

src/analysis/processing/qgsalgorithmsumlinelength.cpp
```diff
@@ -0,0 +1,214 @@
...
+    outputFeature.setAttributes( attrs );
+    return QList< QgsFeature > () << outputFeature;
+  }
+  else
+  {
+    const QgsGeometry polyGeom = feature.geometry();
+    std::unique_ptr< QgsGeometryEngine > engine( QgsGeometry::createGeometryEngine( polyGeom.constGet() ) );
+    engine->prepareGeometry();
+
+    QgsDistanceArea da;
```
nyalldawson: I'd move this up to the prepare step, or use a unique_ptr member so that we only need to initialize once (it's not cheap to create these)

src/analysis/processing/qgsalgorithmzonalstatistics.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+QString QgsZonalStatisticsAlgorithm::shortHelpString() const
+{
+  return QObject::tr( "This algorithm calculates statistics of a raster layer for each feature "
+                      "of an overlapping polygon vector layer." );
+}
+
+QgsProcessingAlgorithm::Flags QgsZonalStatisticsAlgorithm::flags() const
+{
+  return QgsProcessingAlgorithm::flags() | QgsProcessingAlgorithm::FlagNoThreading;
```
nyalldawson: I don't suggest doing it in this PR -- but what do you think about us deprecating this alg, and adding a new one which doesn't modify layers in place? I think it's more in line with how processing works...

nyalldawson (commented):
Epic effort -- lovely work!

There's also some updates which will be required in gui/menus.py to swap qgis -> native (this doesn't happen automatically there)

src/analysis/processing/qgsalgorithmextractspecificvertices.h
alexbruy: I was thinking about making it feature based but then looked at the "Extract vertices" C++ algorithm and found that itis also layer based. So I assumed there was some reasons to make it layer based and ported it as layer based too.

Will address this later if you don't object.

src/analysis/processing/qgsalgorithmrandomextract.cpp
alexbruy: My initial idea was exactly the same - use fids filter to query only required features. Unfortunately this way we loose duplicate features if no additional logic will be added. So I decided to go for slower option for now. Will look at it again and see if I can make it better.

src/analysis/processing/qgsalgorithmzonalstatistics.cpp
alexbruy: Sounds like a plan.

src/analysis/processing/qgsalgorithmextractspecificvertices.h
nyalldawson: I'm guessing it was just an oversight -- I can't see any reason this couldn't/shouldn't be feature based.

src/analysis/processing/qgsalgorithmextractspecificvertices.cpp
```diff
@@ -0,0 +1,197 @@
...
+    outputWkbType = QgsWkbTypes::addZ( outputWkbType );
+  }
+
+  QString dest;
+  std::unique_ptr< QgsFeatureSink > sink( parameterAsSink( parameters, QStringLiteral( "OUTPUT" ), context, dest, outputFields, outputWkbType, source->sourceCrs(), QgsFeatureSink::RegeneratePrimaryKey ) );
+  if ( !sink )
+    throw QgsProcessingException( invalidSinkError( parameters, QStringLiteral( "OUTPUT" ) ) );
+
+  QString verticesString = parameterAsString( parameters, QStringLiteral( "VERTICES" ), context );
+  QStringList verticesList = verticesString.split( ',', QString::SkipEmptyParts );
```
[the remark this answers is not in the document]
alexbruy: Fixed

python/plugins/processing/algs/qgis/SetRasterStyle.py
alexbruy: Done

src/analysis/processing/qgsalgorithmnearestneighbouranalysis.cpp
alexbruy: Should be fixed now

src/analysis/processing/qgsalgorithmsumlinelength.cpp
alexbruy: Done

src/analysis/processing/qgsalgorithmapplylayerstyle.cpp
```diff
@@ -0,0 +1,90 @@
...
+}
+
+QString QgsApplyLayerStyleAlgorithm::shortHelpString() const
+{
+  return QObject::tr( "This algorithm renames a layer." );
+}
+
+QgsProcessingAlgorithm::Flags QgsApplyLayerStyleAlgorithm::flags() const
+{
+  return QgsProcessingAlgorithm::flags() | QgsProcessingAlgorithm::FlagNoThreading;
```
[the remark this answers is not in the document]
alexbruy: Should be fixed. Hope I did it correctly.

src/analysis/processing/qgsalgorithmzonalstatistics.cpp
alexbruy: Do you think we should also update `QgsZonalStatistics` class to produce new layer?

nirvn: Yes please.

src/analysis/processing/qgsalgorithmrandomextract.cpp
roya0045: Out of curiosity, would there be a way to do one request for the base set and another round for duplicates by combining the fids with an expression?

alexbruy: @roya0045 have you read comments above? This will be addressed in the follow-up pull-request

roya0045: @alexbruy yes but I was just trying to suggest a way to do things eventually.

alexbruy: this was already suggested above

> you could avoid this by creating a single feature request with a .setFilterFids on it, and iterating through that. Note that you'd need to count the number of times that feature appears in the fids list and add that many duplicate features though (or you'll reintroduce the original bug!)

roya0045: This was not quite what I had in mind, I was talking about a first phase using the fids, and then for the second pass use fids + an expression to obtain the duplicates that you need instead of doing the exact same query twice or multiple times.

By having those two lists you would not have to care about reading things multiple times.

But this hinges on the fact that there is an optimal way to do this by using an expression and I am not familiar enough with the alg to know if this is true, hence why I was asking.

nyalldawson (approved):

PR #33298, opened 2019-12-09
[processing] use active layer in the widget wrapper only for non-optional params

alexbruy:
## Description
Making @gioman (and users) a bit more happier.

Currenly layer parameters in the standard processing dialogs prepopulated with active layer. This is not very convenient in case of optional parameters as you need manually re-select empty option for them. Proposed fix prepopulates comboboxes only for non-optional parameres.

Here is how it worked before
image
and after patch
image

## Checklist

### DIFF
--- a/python/plugins/processing/gui/wrappers.py
+++ b/python/plugins/processing/gui/wrappers.py
@@ -939,7 +939,10 @@ class MapLayerWidgetWrapper(WidgetWrapper):
                 if self.parameterDefinition().flags() & QgsProcessingParameterDefinition.FlagOptional:
                     self.combo.setLayer(self.parameterDefinition().defaultValue())
                 else:
-                    self.combo.setLayer(iface.activeLayer())
+                    if self.parameterDefinition().defaultValue():
+                        self.combo.setLayer(self.parameterDefinition().defaultValue())
+                    else:
+                        self.combo.setLayer(iface.activeLayer())
             except:
                 pass
 
--- a/python/plugins/processing/gui/wrappers.py
+++ b/python/plugins/processing/gui/wrappers.py
@@ -937,10 +937,10 @@ class MapLayerWidgetWrapper(WidgetWrapper):
 
             try:
                 if self.parameterDefinition().flags() & QgsProcessingParameterDefinition.FlagOptional:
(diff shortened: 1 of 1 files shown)


python/plugins/processing/gui/wrappers.py
```diff
@@ -936,7 +936,10 @@ def createWidget(self):
             self.context = dataobjects.createContext()
 
             try:
-                self.combo.setLayer(iface.activeLayer())
+                if self.parameterDefinition().flags() & QgsProcessingParameterDefinition.FlagOptional:
+                    self.combo.setLayer(self.parameterDefinition().defaultValue())
+                else:
+                    self.combo.setLayer(iface.activeLayer())
```
m-kuhn: Should the default value also have precedence here and activeLayer only be used as fallback?

alexbruy: Maybe you are right, although I have never seen a any usage of default value for layer parameters.

Should be fixed now.

m-kuhn: I did use it recently, but now I wonder how that even worked (given that the old code was forcing `iface.activeLayer()`)

python/plugins/processing/gui/wrappers.py
```diff
@@ -936,7 +936,13 @@ def createWidget(self):
             self.context = dataobjects.createContext()
 
             try:
-                self.combo.setLayer(iface.activeLayer())
+                if self.parameterDefinition().flags() & QgsProcessingParameterDefinition.FlagOptional:
+                    self.combo.setLayer(self.parameterDefinition().defaultValue())
```
nyalldawson: This isn't going to work everywhere -- the default value may be a string pointing to a filename for instance, whereas setLayer requires a map layer value.

alexbruy: Replaced with `setValue` which should handle all of the value types

gioman:
@alexbruy thanks a lot, this is much appreciated as it fixes a UX regression in processing (and also cause of errors in some tools even if the optional input layers were not used, i.e gdal raster calc).

nyalldawson (approved):

Issue #29330, opened 2019-03-07
MDAL: crashes when zooming on EPSG:4326


qgib:
Author Name: **Paolo Cavallini** (@pcav)
Original Redmine Issue: [21513](https://issues.qgis.org/issues/21513)
Affected QGIS version: 3.6.0
Redmine category:mesh_layers
Assignee: Peter Petrik

---
* open QGIS 3.6
* type 'woorld' in the ccod box, to load a world map in 4326
* load the file: https://www.faunalia.eu/~paolo/20190307_190528_GFS_P25_.grb2
* QGIS crashes
All from Debian sid, up  to date.

---

- [Schermata del 2019-03-08 14-55-28.png](https://issues.qgis.org/attachments/download/14557/Schermata%20del%202019-03-08%2014-55-28.png) (Alberto Grava)

PeterPetrik:
probably duplicate of #28827

PR #33308, opened 2019-12-10
Fix mesh test

nyalldawson:
Fixes some flakiness in the mesh calculator test

@PeterPetrik there's some oddities happening here with memory management -- maybe you can clarify?


PeterPetrik:
map renderer is created there to generate triangular mesh. not ideal solution, should be refactored based on your comment. 
we want to (hopefully to 3.12) fix the whole generation of the triangular mesh to fix https://github.com/qgis/QGIS/issues/29330 and similar. that should fix this one too

nyalldawson:
So my fix is ok here?

PeterPetrik:
should be

PR #33313, opened 2019-12-10
Fix profile creation and absence of tbl_srs in qgis.db

nirvn:
## Description

This PR fixes related issues which leads to - among other things - a broken custom projections dialog.

Long story short, if QGIS starts and its root profiles directory is missing, it'll proceed forward with creating a default profile (and in doing so creating a profiles directory) via QDir().mkdir(). Turns out this doesn't work reliably when one needs to create *multiple* directories in the full path. We're fixing that by moving to QDir().mkpath().

This lingering issue was leading to the second problem being fixed here, whereas qgis.db was not copied properly upon profile creation when QDir().mkdir() failed. That caused an empty qgis.db to eventually be created, with the crucial tbl_srs and tbl_projection (as well as tbl_ellipsoid, which is not really a problem for GDAL > 2) missing.

Without tbl_srs, the custom projection dialog was missing. 

The second commit fixes that situation by creating a tbl_srs table when launching QGIS if we detect its missing from qgis.db.

## Checklist

### DIFF
--- a/src/core/qgsapplication.cpp
+++ b/src/core/qgsapplication.cpp
@@ -1907,41 +1907,62 @@ bool QgsApplication::createDatabase( QString *errorMessage )
         return false;
       }
     }
+
+    res = sqlite3_exec( database.get(), "SELECT acronym FROM tbl_projection LIMIT 0", nullptr, nullptr, &errmsg );
+    if ( res != SQLITE_OK )
+    {
+      sqlite3_free( errmsg );
+
+      // qgis.db is missing tbl_projection, create it
+      if ( sqlite3_exec( database.get(),
+                         "CREATE TABLE tbl_projection ("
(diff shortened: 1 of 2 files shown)


nirvn:
@nyalldawson , am I correct when thinking that the tbl_ellipsoid and tbl_projection tables are not used when QGIS is built against GDAL 3?

nirvn:
The tbl_projection is also needed as it is part of the view used to show custom projections in the projection selector widget.

Note: the reason why I'm not merely overwrite qgis.db with the shipped default when tbl_srs and/or tbl_projection are missing is to avoid any user data already present in qgis.db (unlikely, but still).

nirvn:
(Travis fails on an unrelated mesh test failure).

nirvn:
@nyalldawson , good to merge?

nyalldawson (approved):

PR #33314, opened 2019-12-10
Code shuffle: move all labeling related code into src/core/labeling

nyalldawson:
So that these files are all grouped together, making it easier to locate all the components of the labeling engine and hopefully making things easier to navigate (and also to give a tiny cleanup to the messy src/core folder)


nyalldawson:
Unrelated test failures

Issue #29632, opened 2019-04-10
Double click on selected row(s) would zoom on the feature


qgib:
Author Name: **Maxime RIOU** (Maxime RIOU)
Original Redmine Issue: [21817](https://issues.qgis.org/issues/21817)

Redmine category:attribute_table

---
It would be nice to zoom on a feature when double clicking on the row number on the left of the attribute table. I know there is a shortcut (Ctrl+J) and the button on the top of the attribute table, but what about this extra shortcut?

nyalldawson:
This is a won't fix. See the comments on the related pull request.

PR #33322, opened 2019-12-10
Zoom to feature on a map when an attribute table row is double clicked

Samweli:
For https://github.com/qgis/QGIS/issues/29632

### DIFF
--- a/python/gui/auto_generated/attributetable/qgsattributetableview.sip.in
+++ b/python/gui/auto_generated/attributetable/qgsattributetableview.sip.in
@@ -98,6 +98,15 @@ Disables selection change for these events.
 :param event: The mouse event
 %End
 
+    virtual void mouseDoubleClickEvent( QMouseEvent *event );
+
+%Docstring
+Called for mouse double-click events on a table cell.
+Disables selection change for these events.
+
+:param event: The mouse event
+%End
+
     virtual void keyPressEvent( QKeyEvent *event );
 
 %Docstring
--- a/python/gui/auto_generated/attributetable/qgsattributetableview.sip.in
+++ b/python/gui/auto_generated/attributetable/qgsattributetableview.sip.in
@@ -98,15 +98,6 @@ Disables selection change for these events.
 :param event: The mouse event
 %End
 
-    virtual void mouseDoubleClickEvent( QMouseEvent *event );
-
-%Docstring
-Called for mouse double-click events on a table cell.
-Disables selection change for these events.
-
-:param event: The mouse event
-%End
-
     virtual void keyPressEvent( QKeyEvent *event );
 
 %Docstring
--- a/src/gui/attributetable/qgsattributetablemodel.cpp
+++ b/src/gui/attributetable/qgsattributetablemodel.cpp
@@ -619,7 +619,7 @@ QVariant QgsAttributeTableModel::headerData( int section, Qt::Orientation orient
     if ( orientation == Qt::Vertical )
     {
       // TODO show DisplayExpression
-      return tr( "Feature ID: %1" ).arg( rowToId( section ) );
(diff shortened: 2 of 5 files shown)


nyalldawson:
@nirvn what's your thoughts here? I'm personally concerned that intercepting the double click inside the cell would potentially break some of the edit widget behavior.

nirvn:
I'm also sharing the same concern as Nyall raised here.

nirvn:
Why not narrowing the behavior down to clicking or double click on the vertical header (i.e. 

[ 1 ]
[ 2 ]
[ 3 ]
etc.

)?

gioman:
> Why not narrowing the behavior down to clicking or double click on the vertical header (i.e.

agree,

Samweli:
> I'm also sharing the same concern as Nyall raised here.

Currently this feature doesn't work when layer is in edit mode.

Samweli:
> Why not narrowing the behavior down to clicking or double click on the vertical header (i.e.
> 
> [ 1 ]
> [ 2 ]
> [ 3 ]
> etc.
> 
> )?

@nirvn @nyalldawson just updated the PR with @nirvn suggestion
Below a snippet on how it now works
using vertical header to zoom feature-2019-12-12 17-34

elpaso:
Is there a tooltip or something that lets a user know about this behavior or is she supposed to learn it from the manual?
 If not on the header itself, it could be added to the toolbutton that trigges the same action on the selected feature.

Samweli:
> Is there a tooltip or something that lets a user know about this behavior or is she supposed to learn it from the manual?
> If not on the header itself, it could be added to the toolbutton that trigges the same action on the selected feature.

Right @elpaso, what do you think would be the appropriate tooltip message? The toolbutton with the same action already has "**Zoom map to the selected rows (Ctrl+J)**".

timlinux:
> Right @elpaso, what do you think would be the appropriate tooltip message? The toolbutton with the same action already has "Zoom map to the selected rows (Ctrl+J)".

How about "Zoom map to the selected rows (Ctrl+J) or double click the row header."

timlinux:
Maybe we could also do:
image

Feature id: 4322 (double-click here to zoom to this feature)

Samweli:
@timlinux Okay!

Saijin-Naib:
Would this also trigger if we go down the table using the keyboard? I'd like that.

nyalldawson:
@saijin-naib that's possible since 3.8: https://qgis.org/en/site/forusers/visualchangelog38/#feature-allow-browsing-feature-list

3nids:
Rather than introducing a double click, why not creating a standard qgis action zoom to which would be created for all geometry layers.
It would then be accessible on right click.
It sounds more intuitive to me.

Samweli:
> Rather than introducing a double click, why not creating a standard qgis action zoom to which would be created for all geometry layers

What do you all think about this, should we have both functions, the context menu with a zoom action on the vertical row header (right click triggered ) and double click event on vertical row header? Or only one of these?

Samweli:
> What do you all think about this, should we have both functions, the context menu with a zoom action on the vertical row header (right click triggered ) and double click event on vertical row header? Or only one of these?

@nirvn @nyalldawson

roya0045:
Do you intend to have it enabled for edit mode? Most of the time when a user has to change some information, it is best if he can quickly see the feature and I think such a behaviour is better when editing.

Samweli:
> Do you intend to have it enabled for edit mode?

I think it should be on both view and edit modes.

nirvn (commented):
Looks good to me. I'd be fine merging this.

@3nids , are you ok?

3nids:
I am personaly -0 here:
* we already have a browsing mode when in form view
* double-click is not intuitive
* if we do need a browsing mode in the table view (vs form view), having it as an action would make it reusable in other places.

But if I'm the only here, I wouldn't mind having this merged.

PR #33326, opened 2019-12-11
CRS WKT projection fixes and user defined CRS via WKT

nyalldawson:
This PR fixes an issue where CRSes which are created from a WKT string, which does NOT match anything from the known registries, is incorrectly flagged as an invalid CRS. If proj is happy to accept the string as valid wkt, then that's sufficient for us....

It also modifies the user crs database and logic so that if a wkt representation of a CRS is available, that is stored within the custom crs definition. This allows user CRSes to be created losslessly from WKT strings, rather then forcing the CRS definition to proj and losing critical information in the process. This allows CRSes created using a WKT string to correctly work as user defined CRSes. (TODO: we need to rework the user facing Custom CRS dialog to accept WKT string definitions too, and push users towards this instead of proj strings)


nyalldawson:
@rouault QGIS specific logic aside, do you agree that pushing users towards WKT definitions for custom projections is the right move forward?

PR #33333, opened 2019-12-11
Add expressions is_empty(geom)  is_empty_or_null(geom)

lbartoletti:
## Description

Following my work on empty geometries, here is a new expression to catch them.

image

sponsored by QWAT cc @kandre @haubourg 
## Checklist

### DIFF
--- a/resources/function_help/json/is_empty
+++ b/resources/function_help/json/is_empty
@@ -4,7 +4,7 @@
   "description": "Returns true if a geometry is empty (without coordinates) or false if a is empty."
   "arguments": [ {"arg":"geom","description":"a geometry"}],
   "examples": [ { "expression":"is_empty(geom_from_wkt('LINESTRING(0 0, 1 1, 2 2)'))", "returns":"false"},
-                { "expression":"is_empty(geom_from_wkt('LINESTRING EMPTY'))", "returns":"true"}]
-                { "expression":"is_empty(geom_from_wkt('POINT(7 4)'))", "returns":"false"}]
-                { "expression":"is_empty(geom_from_wkt('POINT EMPTY'))", "returns":"true"}]
+                { "expression":"is_empty(geom_from_wkt('LINESTRING EMPTY'))", "returns":"true"},
+                { "expression":"is_empty(geom_from_wkt('POINT(7 4)'))", "returns":"false"},
+                { "expression":"is_empty(geom_from_wkt('POINT EMPTY'))", "returns":"true"} ]
 }
--- a/resources/function_help/json/is_empty
+++ b/resources/function_help/json/is_empty
@@ -1,7 +1,7 @@
 {
   "name": "is_empty",
   "type": "function",
-  "description": "Returns true if a geometry is empty (without coordinates), false if the geometry is not empty and NULL if there is no geometry.",
+  "description": "Returns true if a geometry is empty (without coordinates), false if the geometry is not empty and NULL if there is no geometry. See also `is_empty_or_null`.",
   "arguments": [ {"arg":"geom","description":"a geometry"}],
   "examples": [ { "expression":"is_empty(geom_from_wkt('LINESTRING(0 0, 1 1, 2 2)'))", "returns":"false"},
                 { "expression":"is_empty(geom_from_wkt('LINESTRING EMPTY'))", "returns":"true"},
--- a/resources/function_help/json/is_empty
+++ b/resources/function_help/json/is_empty
@@ -1,7 +1,7 @@
 {
   "name": "is_empty",
   "type": "function",
-  "description": "Returns true if a geometry is empty (without coordinates) or false if a is empty."
+  "description": "Returns true if a geometry is empty (without coordinates) or false if the geometry is empty.",
   "arguments": [ {"arg":"geom","description":"a geometry"}],
   "examples": [ { "expression":"is_empty(geom_from_wkt('LINESTRING(0 0, 1 1, 2 2)'))", "returns":"false"},
                 { "expression":"is_empty(geom_from_wkt('LINESTRING EMPTY'))", "returns":"true"},
--- a/resources/function_help/json/is_empty
+++ b/resources/function_help/json/is_empty
@@ -1,0 +1,10 @@
+{
+  "name": "is_empty",
+  "type": "function",
+  "description": "Returns true if a geometry is empty (without coordinates) or false if a is empty."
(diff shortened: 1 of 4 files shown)


m-kuhn:
Nice!
What's the preferred way to test that a geometry is neither empty nor null? I think that should be listed in the docs.

haubourg:
>  What's the preferred way to test that a geometry is neither empty nor null? I think that should be listed in the docs.

@m-kuhn well, the SQL way is `geom is Null or st_IsEmpty(geom)`, I suppose we should behave the same. 
However a all in one expression catching both NULL or empty would be nice to avoid this nasty little tricky empty notion nobody knows about. I have no clue how we could name it.

roya0045:
> @m-kuhn well, the SQL way is `geom is Null or st_IsEmpty(geom)`, I suppose we should behave the same.
> However a all in one expression catching both NULL or empty would be nice to avoid this nasty little tricky empty notion nobody knows about. I have no clue how we could name it.

But isn't a Null geometry or an empty geometry the same thing to some extent. Since geometries are defined by their position an empty container or no container should be the same to any user. 

I don't think an empty geometry gives any advantage over a Null geometry as a user, please let me know if I'm wrong.

m-kuhn:
It's an SQL thing. Just like `'' IS NOT NULL`. We shouldn't change it at this level, but make it easy for user to not get it wrong.

 - By documenting it clearly
 - And even better by adding a shorthand like `is_void(geometry)` (counter suggestions very welcome!).

resources/function_help/json/is_empty
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "is_empty",
+  "type": "function",
+  "description": "Returns true if a geometry is empty (without coordinates) or false if a is empty."
```
nyalldawson:
```suggestion
  "description": "Returns true if a geometry is empty (without coordinates) or false if the geometry is empty."
```

nyalldawson:
>, the SQL way is geom is Null or st_IsEmpty(geom), I suppose we should behave the same.

+1 to "geom is null"

m-kuhn:
> > , the SQL way is geom is Null or st_IsEmpty(geom), I suppose we should behave the same.
> 
> +1 to "geom is null"

What's that +1 for exactly? (or what's the counter proposal you are dismissing?)

nyalldawson:
> What's that +1 for exactly? (or what's the counter proposal you are dismissing?)

It was in contrast to st_isempty(geom) as a null test. A geometry could be non-null but empty...

lbartoletti:
> > What's that +1 for exactly? (or what's the counter proposal you are dismissing?)
> 
> It was in contrast to st_isempty(geom) as a null test. A geometry could be non-null but empty...

I will add another expression is_null(geom)

nyalldawson:
Why not just plugin into the existing "IS NOT NULL" support?

lbartoletti:
> Why not just plugin into the existing "IS NOT NULL" support?

I hadn't thought about that, I usually go through functions it seems clearer to me :)

m-kuhn:
Any feedback on the shorthand for null-or-empty? I think providing and recommending this could act as a help to avoid common pitfalls (for the 99% of the cases where people don't care about the difference).

lbartoletti:
> Any feedback on the shorthand for null-or-empty? I think providing and recommending this could act as a help to avoid common pitfalls (for the 99% of the cases where people don't care about the difference).

Oh sure, I can add a function grouping the is empty or null in a special function.

lbartoletti:
> > Why not just plugin into the existing "IS NOT NULL" support?
> 
> I hadn't thought about that, I usually go through functions it seems clearer to me :)

@nyalldawson In fact, no needs to add an expression "$geometry IS NULL" works

lbartoletti:
@m-kuhn FYI is_empty(NULL) (or null geometry) returns NULL too

m-kuhn:
@lbartoletti sorry, I can't follow your latest statement and it's implications. I.e. what does (or null geometry) mean, and what does the "too" refer to?

lbartoletti:
> @lbartoletti sorry, I can't follow your latest statement and it's implications. I.e. what does (or null geometry) mean, and what does the "too" refer to?

Sorry, I didn't see your message and I edited mine in the meantime.

m-kuhn:
Thanks. This sounds reasonable as test result :+1: 
I still wonder what the "too" means in the message. And if this was meant as a response to the shorthand proposal.

resources/function_help/json/is_empty
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "is_empty",
+  "type": "function",
+  "description": "Returns true if a geometry is empty (without coordinates) or false if the geometry is empty.",
```
DelazJ: Isn't there an issue in this sentence? same condition returns either true or false?

lbartoletti: Done. Thanks...

resources/function_help/json/is_empty
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "is_empty",
+  "type": "function",
+  "description": "Returns true if a geometry is empty (without coordinates), false if the geometry is not empty and NULL if there is no geometry.",
```
m-kuhn:
```suggestion
  "description": "Returns true if a geometry is empty (without coordinates), false if the geometry is not empty and NULL if there is no geometry. See also `is_empty_or_null`.",
```

m-kuhn (approved):
Thanks a lot. One minor suggestion to bring people to the right path.

DelazJ:
just to add my usual comment: `[feature] is missing` :smile:

m-kuhn:
Sorry @DelazJ , I missed your comment.
I added a label.
@lbartoletti best to add it directly to the changelog already.

DelazJ:
> Sorry @DelazJ , I missed your comment.
I added a label.

No problem. I don't think it works afterwards. Not yet. This was your grant proposal, right? Meanwhile, I think we need to manually create the report in docs repo
>@lbartoletti best to add it directly to the changelog already.

Do we have an automated changelog generated?

m-kuhn:
> No problem. I don't think it works afterwards. Not yet. This was your grant proposal, right? Meanwhile, I think we need to manually create the report in docs repo

Yep, @3nids is on it. And it should also work for previous ones, one just needs to relabel them (so the important point is to be able to find them again)

> Do we have an automated changelog generated?

Semi automated by @timlinux at some point I think

PR #33336, opened 2019-12-11
[mesh] Active flag refactoring

PeterPetrik:
update MDAL to 0.4.92 and required changes to MDAL data provider

this version introduce "active flag capability". Most of the MDAL drivers do not support active flag concept. In this case, MDAL always created dummy arrays full of 1 (true) values and returned that to QGIS. With this version, QGIS queries if active flag is supported, and if not, the data is not requested/used/fetched at all. It speeds up and saves some memory.


nyalldawson (commented):
Can't comment on the mdal specifics, but the code looks good to me

wonder-sk:
I have reviewed mdal changes on a MDAL pull request... and qgis changes look fine... +1 to merge

PR #33341, opened 2019-12-11
Add test for handling NULL values in PostgreSQL connector

strk:
See https://github.com/qgis/QGIS/pull/33225#issuecomment-564446756

### DIFF
--- a/tests/src/python/test_qgsproviderconnection_postgres.py
+++ b/tests/src/python/test_qgsproviderconnection_postgres.py
@@ -200,6 +200,15 @@ class TestPyQgsProviderConnectionPostgres(unittest.TestCase, TestPyQgsProviderCo
         self.assertEqual(conn.executeSql('SELECT FALSE'), [[False]])
         self.assertEqual(conn.executeSql('SELECT TRUE'), [[True]])
 
+    def test_nulls(self):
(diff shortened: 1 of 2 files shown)


tests/src/python/test_qgsproviderconnection_postgres.py
```diff
@@ -200,6 +200,15 @@ def test_true_false(self):
         self.assertEqual(conn.executeSql('SELECT FALSE'), [[False]])
         self.assertEqual(conn.executeSql('SELECT TRUE'), [[True]])
 
+    def test_true_false(self):
```
elpaso:
```suggestion
    def test_nulls(self):
```

elpaso (changes requested):
change the test name or it will shadow the previous one

strk:
OOps, force-pushed a fix

elpaso (approved):

strk:
@elpaso the tests indeed fail:
https://travis-ci.org/qgis/QGIS/jobs/623752437#L2455-L2471

Those NULL values are returned as False instead of None

strk:
(but only for the bool case?)

strk:
Is it ok to merge with a failing Azure pipeline ? (seems to be)

PR #33347, opened 2019-12-11
Fix another unicode string on windows

nyalldawson:


nirvn:
Oh good.

nyalldawson:
Unrelated test failures

PR #33365, opened 2019-12-12
Processing multithreaded vector

mhugent:
This PR adds code to execute vector overlay operations (intersection, union, difference, symdifference) on multiple CPU cores. The tools are integrated into the processing framework.
Besides intersectiontool, uniontool, differencetool, symdifferencetool in this PR, we also have buffertool, sliverpolygontool.h, convexhulltool, dissolvetool. Going to integrate the latter tools in a second step.


src/analysis/vector/abstracttool.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+  void AbstractTool::buildSpatialIndex( QgsSpatialIndex &index, QgsFeatureSource *layer ) const
+  {
+    QgsFeature currentFeature;
+    QgsFeatureRequest request;
+    request.setFlags( QgsFeatureRequest::SubsetOfAttributes );
+    QgsFeatureIterator it = layer->getFeatures();
+    while ( it.nextFeature( currentFeature ) )
+    {
+      index.addFeature( currentFeature );
```
alexbruy: Would not be better to bulk-load features from the iterator or source into spatial index instead of adding them one by one? It should be faster.

elpaso:
@mhugent this is great, thank you!  

Just a small request: please tell the QGIS community that you are working on a new feature in the future, it is a pure coincidence that I didn't start working on a similar improvements and we cannot afford duplicated efforts, do we?

src/analysis/vector/abstracttool.cpp
```diff
@@ -0,0 +1,177 @@
...
+  QFuture<void> AbstractTool::execute( int /*task*/ )
+  {
+    return QtConcurrent::map( mJobQueue, ProcessFeatureWrapper( this ) );
+  }
+
+  void AbstractTool::buildSpatialIndex( QgsSpatialIndex &index, QgsFeatureSource *layer ) const
+  {
+    QgsFeature currentFeature;
+    QgsFeatureRequest request;
+    request.setFlags( QgsFeatureRequest::SubsetOfAttributes );
```
alexbruy: Is this necessary? We don't fetch any attributes here

src/analysis/processing/qgsoverlayutils.cpp
```diff
@@ -452,4 +452,64 @@ void QgsOverlayUtils::resolveOverlaps( const QgsFeatureSource &source, QgsFeatur
   }
 }
 
+void QgsOverlayUtils::runVectorAnalysisTool( Vectoranalysis::AbstractTool &tool, QgsProcessingFeedback *feedback )
+{
+  QFutureWatcher<void> fWatcher;
+  QEventLoop evLoop;
+  if ( feedback )
+  {
+    QObject::connect( feedback, SIGNAL( canceled() ), &fWatcher, SLOT( cancel() ) );
```
elpaso: Please use new signal slot syntax.

src/analysis/vector/abstracttool.cpp
```diff
@@ -0,0 +1,177 @@
...
+  QVector<QgsFeature *> AbstractTool::getIntersects( const QgsRectangle &rect, QgsSpatialIndex &index, QgsFeatureSource *layer, const QgsAttributeList &attIdx )
+  {
+    mIntersectMutex.lock();
+    QList<QgsFeatureId> intersectIds = index.intersects( rect );
+    mIntersectMutex.unlock();
+
+    QVector<QgsFeature *> featureList;
+    for ( QgsFeatureId id : intersectIds )
+    {
+      QgsFeature *feature = new QgsFeature();
```
elpaso: Better use a unique_ptr here.

elpaso:
Any reason why the class name(space)s do not have QGIS `Qgs` prefix?

nyalldawson:
Please do not merge yet -- I want to do an in-depth review first

nyalldawson:
Was this based on the existing methods from QgsOverlayUtils, or ported directly from elsewhere? I'm concerned if it's the latter case, as the QgsOverlayUtils methods have been heavily refined and well tested, and are proven robust...

nyalldawson:
Actually looking at it, it's definitely not based on QgsOverlayUtils. I'm -1 for merge until this is updated to use the proven logic from QgsOverlayUtils.

mhugent:
@elpaso, @alexbruy : thank you for your constructive input

nyalldawson:
Very subtle @mhugent. My point still stands, you can't just dismiss all the hard work @wonder-sk and others have put into making the existing implementations rock solid.

I'm definitely not against you doing this work, but stability and correct results >>> speed.

mhugent:
@nyalldawson : While I agree that the existing tests should pass (PyQgsProcessingInPlace still fails, going to check why), the argument 'is not based on something' is very strange.

nyalldawson:
Look at the code! There's multiple fixes and optimisations present in master that you've completely ignored in your code.

mhugent:
>making the existing implementations rock solid

In my experience, with most current open source tools (including my PR, QGIS and others), people run into problems with real world overlay operations. Maybe it helps if we use the precision parameter of geos? I need to do some tests over the christmas days to see if that helps

mhugent:
>Look at the code! There's multiple fixes and optimisations present in master that you've completely >ignored in your code.

Coming back to the  'constructive input'. Say which fixes and optimisations you miss and I'm happy to adapt the PR code.

PeterPetrik:
@wonder-sk travels these days but I think he would like to review these changes, please wait for his review before merging too. thanks

m-kuhn:
> the argument 'is not based on something' is very strange.

I think what he is referring to is mostly code duplication which results in an additional long term maintenance overhead.

(Sorry for not bringing more constructive input in here, I lack the time for an in depth review right now).

nyalldawson:
It's not just code duplication -- it's that years of careful fixes and optimisations have been dropped by bypassing this code.

@mhugent I can't really be more constructive then that -- the multithreaded implementation should be a multithreaded version of the current code, not a completely new implementation of these algorithms. When that's done I'll be happy to review in detail and provide specific comments, but until then I can't give any guidance more specific then that.

wonder-sk:
Agreed with comments from @nyalldawson - if we need these algorithms to be multi-threaded, why not just adapt the existing implementation?

The existing implementation also has the advantage of being simple and easy to reason about. There are just two "basic" algorithms: difference and intersections. The union is just 2x difference and 1x intersection, the sym.difference is 2x difference. The implementation in this PR has all the operations mixed together (as it was the case with the older processing implementation).

To name a couple of improvements in the current QGIS implementation over the code in this PR:
- cancellation does not seem to be supported
- on the fly reprojection
- using fast insert on sink
- geometries are prepared only when needed
- difference alg: it's faster to do unary union of all geoms, then difference - rather than a series of difference operations
- layer A: using sequential reading of features instead of fetching by ID
- layer B: features are requested in groups of FIDs rather than one by one

What is the speedup that you gain with this implementation?

I am also -1 to merge this PR - I think we should go the way of making the current implementation multi-threaded rather than starting from scratch with 1K of new lines of code. It is essentially just two functions (each under 100 lines) that would need to make parallelized.

mhugent:
>What is the speedup that you gain with this implementation?

Tested if with two layers 'land coverage' and 'wood'. Both are PostGIS layers with a spatial index and the results are written to memory layer. The speedup is 2-4 times on my four-core laptop:

intersection wald<->bodenbedeckung
current: 5min 16s
multithreaded: 2min 1s

difference bodenbedeckung<->wald
current: 4min 30s
multithreaded: 2min 14s

union bodenbedeckung<->wald
current: 22min 45s
multithreaded: 5min 20s

symdiff wald<->Bodenbedeckung
current: 15min 31s
multithreaded: 4min 50s

mhugent:
I'm planning to do more tests

mhugent:
>cancellation does not seem to be supported

It is supported. In the multithreaded implementation the cancelation is done with QFuture instead of asking isCanceled() in the feature loop

mhugent:
>on the fly reprojection

Agreed, this needs to be added

>using fast insert on sink
>geometries are prepared only when needed

Thanks for the pointer. I'm going to test what the difference in performance is.

nyalldawson:
> To name a couple of improvements in the current QGIS implementation over the code in this PR:

Another **critical** improvements which the existing code has is that output results are "sanitized", automatically handling geometry collections in the correct way

Honestly, just follow the advice @wonder-sk and I are giving, and port the existing implementation rather than continuing this from-scratch rewrite. It'll be **much** faster to get to a merge-ready state than pushing ahead with this implementation.

m-kuhn:
Hi @mhugent , sorry to jump back in.

From the sidelines it looks like code deduplication and porting to existing code tools has been requested by very active core developers while I think the current development in this pull request is still continued towards the original (duplicated) tools.

I have not seen any (public) consensus on this approach, has there been any? Or am I misreading the situation?

It would be good if we could avoid a stale mate situation where a lot of effort is put into something that is not mergable - or into something that is merged without a consensus which will leave no one happy.

mhugent:
>while I think the current development in this pull request is still continued towards the original >(duplicated) tools.

Not at all. I've put the existing QgsOverlayTools code into feature-specific methods ( featureIntersection,  featureDifference). These two methods are called from the intersection, difference, union and symmetrical difference tools. Therefore it is both based on existing code and it also follows the idea from Martin to base everything on intersection and difference.

nyalldawson:
Thanks @mhugent!

Just so there's no misunderstanding: this will still need re-reviews by myself and @wonder-sk (possibly the others who have expressed concern here too) before ANYONE hits that green merge button!

m-kuhn:
Thanks @mhugent :+1: 

Glad I got the situation wrong

python/plugins/processing/tests/testdata/qgis_algorithm_tests4.yaml
```diff
@@ -364,8 +364,11 @@ tests:
         type: vector
         pk: id_a
         compare:
+          unordered: true
           fields:
             fid: skip
+          geometry:
```
nyalldawson, line 370: why are these changes required?

src/analysis/processing/qgsoverlayutils.cpp
```diff
@@ -73,189 +73,142 @@ static bool sanitizeDifferenceResult( QgsGeometry &geom )
   return true;
 }
 
-
-void QgsOverlayUtils::difference( const QgsFeatureSource &sourceA, const QgsFeatureSource &sourceB, QgsFeatureSink &sink, QgsProcessingContext &context, QgsProcessingFeedback *feedback, int &count, int totalCount, QgsOverlayUtils::DifferenceOutput outputAttrs )
+QgsFeatureList QgsOverlayUtils::featureIntersection( const QgsFeature &featA, const QgsFeatureSource &sourceA, const QgsFeatureSource &sourceB, QgsSpatialIndex &indexB, const QgsCoordinateTransformContext &transformContext, const QList<int> &fieldIndicesA, const QList<int> &fieldIndicesB )
```
nyalldawson, line 76: can you please swap back the order of these two functions? Hopefully it will make the diff cleaner and easier to review. As it stands it's basically impossible to reliably review this change.

src/analysis/vector/qgsabstracttool.cpp
```diff
@@ -0,0 +1,110 @@
...
+  {
+    return QtConcurrent::map( mJobQueue, ProcessFeatureWrapper( this ) );
+  }
+
+  void QgsAbstractTool::buildSpatialIndex( QgsSpatialIndex &index, QgsFeatureSource *layer ) const
+  {
+    QgsFeatureRequest request;
+    request.setFlags( QgsFeatureRequest::SubsetOfAttributes );
+    request.setSubsetOfAttributes( QgsAttributeList() );
+    QgsFeatureIterator it = layer->getFeatures( request );
```
nyalldawson, line 70: This isn't correctly handling the reprojection required for sources with mismatched CRS

src/analysis/vector/qgsabstracttool.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+  void QgsAbstractTool::appendToJobQueue( QgsFeatureSource *layer, int taskFlag )
+  {
+    for ( const QgsFeatureId &id : layer->allFeatureIds() )
+    {
+      mJobQueue.append( new Job( id, taskFlag ) );
+    }
+  }
+
+  bool QgsAbstractTool::getFeatureAtId( QgsFeature &feature, QgsFeatureId id, QgsFeatureSource *layer, const QgsAttributeList &attIdx )
```
nyalldawson, line 82: These one-by-one requests are **extremely** slow for many feature sources, and must be avoided.

src/analysis/vector/qgsabstracttool.h
```diff
@@ -0,0 +1,136 @@
...
+
+      /**
+       * Writes feature to output source
+       */
+      void writeFeatures( QgsFeatureList &outFeatures );
+
+      QList<Job *> mJobQueue;
+      QStringList mExceptions;
+
+      QMutex mIntersectMutex;
```
nyalldawson, line 127: seems unused

src/analysis/vector/qgsabstracttool.cpp
```diff
@@ -0,0 +1,110 @@
...
+    QgsFeatureRequest request;
+    request.setFlags( QgsFeatureRequest::SubsetOfAttributes );
+    request.setSubsetOfAttributes( QgsAttributeList() );
+    QgsFeatureIterator it = layer->getFeatures( request );
+    index = QgsSpatialIndex( it );
+  }
+
+  void QgsAbstractTool::appendToJobQueue( QgsFeatureSource *layer, int taskFlag )
+  {
+    for ( const QgsFeatureId &id : layer->allFeatureIds() )
```
nyalldawson, line 76: This causes a full iteration over the source and should be avoided - as it stands right now the code is doing one full iteration here, then requesting features one-by-one. In many cases that's going to kill any performance gains we've obtained by the multi-threading.

wonder-sk:
I wonder if there was any profiling / performance analysis done before the work started: we need to have a good understanding what is slow and when. Having multi-threaded implementation makes sense if a task can be split well into independent sub-tasks and it is heavily CPU-bound. In this case, each "job" is a mixture of I/O and CPU intensive operation (requesting features from layer + doing geometry calculations). The fact that we have I/O involved makes things difficult - things may get even slower as Nyall pointed out because suddenly the data access pattern is very different.

It would be interesting to investigate:
1. caching of layer B in memory (at least partially) and/or on local drive for fast access to data
2. using the "prepared geometry" concept for calculation of geometries (currently only available for boolean predicates like "intersects", not for calls like "intersection" that return geometries)

src/analysis/vector/qgsabstracttool.cpp
mhugent: Thank you for coming back to this PR. I agree the fetching one-by-one can be a problem, especially with slow remote DBs.
I have some ideas how to avoid the one-by-one fetching, going to  make some tests in the next days.

src/analysis/vector/qgsabstracttool.cpp
mhugent: The full iteration over the source is probably hard  to avoid. But yes, the one-by-one fetching should be avoided

mhugent:
@wonder-sk : the profiling will most likely give very different results depending on which use-case your are testing. There are definitely very CPU-intensive use-cases, especially if there are a lot of intersecting objects.
Before doing more advanced things,  I'd like to get rid of the one-by-one fetching. Then the I/O will be similar to what we have now.

nyalldawson:
Thanks @mhugent! Might be worth opening a thread on the devel list to discuss possible approaches for wider feedback before diving into more code here too -- I'd especially welcome wider expertise regarding how different file formats (e.g. gpkg) react to multi-threaded feature iteration, and e.g. whether sqlite places internal locks on the access which would prevent multi-threaded reading from benefitting.

mhugent:
Seems it is relatively straightforward to avoid the one-by-one feature fetching. I've made a quick prototype test with code that fetches the features in chunks (e.g. 100 features in one go). From this chunk, the job queue is populated and executed with multiple threads. Then the job queue is emptied and the next chunk is fetched. So the featureAtId calls are replaced  by a nextFeature() - Iteration.

I've tested it with an intersection of two layers, the same data in a local database and in a remote database on qgiscloud. The data has relatively few intersections so it should be a case where the IO is important (as opposite to a CPU-intensive use-case where we have many intersections and complex data):

Intersection on Remote PostGIS-DB:
current master: 24.97s
multithreaded (features one-by-one): 17.77s
multithreaded (chunks): 10.15s

Intersection same data local PostGIS-DB:
current master: 1.48s
multithreaded (features one-by-one): 0.76s
multithreaded (chunks): 0.87s

andreasneumann:
Interesting - thank you for sharing these results! Sounds like a good speed gain.

mhugent:
Given the promising results of the batch feature fetching, is it ok if I go on to update the PR in this direction?

python/plugins/processing/tests/testdata/qgis_algorithm_tests4.yaml
mhugent: The test was failing because in the output polygon, the start/end vertex was not the same. Don't know why this happened, however we can consider the output is correct no matter which vertex in the polygon ring is first / last.

mhugent:
So here is the implementation with single threaded batch IO and multithreaded processing. The runtime is now a bit slower on local postgres datasources compared to the first version in this PR (since IO is single threaded), but faster on remote data sources.

kikislater:
Why was it never merged into master ? Need rework ?

mhugent:
I don't know the reason it was never merged. All the feedback has been addressed and the unit tests passed

kikislater:
Stale bot closed this PR :-/ 

But now seems more complex to merge as some processing cpp files changed

rldhont:
@mhugent it has been probably lost in the mass. Can you rebase and open a new PR ?

nyalldawson:
Basically it was never merged because it's an extremely complex change, and one which we cannot afford to have any regressions at all in. No one with the required expertise (@wonder-sk , myself or @m-kuhn) had the time required to properly review, which would likely take a good half day to do.

I also have issues in the general design of the approach taken here. It cannot be extended to other processing algorithms, and I'd prefer a generic multithreading processing approach which can easily be extended to other algorithms (eg buffers, make valid, etc ) without having to have yet another multi threading implementation in qgis/processing.

gioman:
While I can't comment on the technical aspect of this matter, it seems really a pity that such important kind of improvement can't be added to QGIS considered the struggle we face since forever in having such basic operations runs at acceptable speeds with large datasets (and being bashed along the way when compared to other GIS packages). Can't you developers find some common ground and delineate a plan, eventually with the support of qgis.org? @mbernasocchi @andreasneumann

nyalldawson:
@gioman the situation is far from as straightforward as that. For instance, I strongly suspect that the recent thread on the mailing list relates to an operation being performed on a single very complex geometry. All the processing in this case is occurring in geos (which would explain a lack of progress updates - it's a single feature being processed). 

In this situation ALL software using geos will show the same lack of performance, including gdal and postgis. (Fingers crossed that the new geos overlay engine helps in cases like this!)

gioman:
> (Fingers crossed that the new geos overlay engine helps in cases like this!)

@nyalldawson is there a rough timeline about when we could have QGIS shipped with that?

nyalldawson:
@gioman it would need a new stable geos release first, and even then I'd suggest we play it safe and let the release stabilise before moving the windows builds to it. Geometry operations are just too fundamental a part of qgis, we can't use our users as a test bed for geos 😁

mhugent:
>It cannot be extended to other processing algorithms

To support other tools which have a feature-by-feature processing (e.g. buffer), it is possible to create a new subclass of QgsAbstractTool (just as it is done with QgsDifferenceTool, QgsIntersectionTool, ...

gioman:
> @gioman it would need a new stable geos release first, and even then I'd suggest we play it safe and let the release stabilise before moving the windows builds to it. Geometry operations are just too fundamental a part of qgis, we can't use our users as a test bed for geos grin

@nyalldawson this seems a polite way of saying "more than 1 year" :) Considered that we don't know how much this will improve speed I would really like to see a common push to make clip,union,intersection,difference,dissolve,buffer multi-threaded. As a user (that started to clip stuff with "ftools", so my pain comes from a long ago) I would support this effort with contribution, and I'm sure many more would do.

PR #33368, opened 2019-12-12
Processing updates

alexbruy:
## Description
Follow up #33290.

Make following algorithms feature-based:
- Extract vertices
- Extract specific vertices
- Pole of inaccessibility

Optimize "Random extract" algorithm by avoiding single-feature requests at first step and using them only when duplicated features should be added.

## Checklist

### DIFF
--- a/src/analysis/processing/qgsalgorithmrandomextract.cpp
+++ b/src/analysis/processing/qgsalgorithmrandomextract.cpp
@@ -78,7 +78,7 @@ QVariantMap QgsRandomExtractAlgorithm::processAlgorithm( const QVariantMap &para
 
   QString dest;
   std::unique_ptr< QgsFeatureSink > sink( parameterAsSink( parameters, QStringLiteral( "OUTPUT" ), context, dest, source->fields(),
-                                          source->wkbType(), source->sourceCrs() ) );
+                                          source->wkbType(), source->sourceCrs(), QgsFeatureSink::RegeneratePrimaryKey ) );
   if ( !sink )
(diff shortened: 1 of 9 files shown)


src/analysis/processing/qgsalgorithmrandomextract.cpp
```diff
@@ -110,22 +110,48 @@ QVariantMap QgsRandomExtractAlgorithm::processAlgorithm( const QVariantMap &para
...
+      {
+        break;
+      }
+
+      if ( it.value() > 1 )
+      {
+        source->getFeatures( QgsFeatureRequest( it.key() ), QgsProcessingFeatureSource::FlagSkipGeometryValidityChecks ).nextFeature( f );
+        for ( int j = 1; j < it.value(); j++ )
+        {
+          sink->addFeature( f, QgsFeatureSink::FastInsert );
```
nyalldawson: need to make sure the sink is created with the RegeneratePrimaryKey flag, otherwise you'll hit duplicate fid errors when exporting to e.g. gpkg

nyalldawson: I think all this could be simplified:
```
while ( fit.nextFeature( f ) )
{
   const int count = ids.value( f.id() );
   for ( int i = 0; i < count; ++i )
   {
       sink->addFeature( f, QgsFeatureSink::FastInsert );
   }
}
```
That should be nice and efficient, and avoids all multiple feature requests

nyalldawson:
All looks good to me!

nyalldawson (approved):

PR #33369, opened 2019-12-12
Fix compilation error in qgsdxfexport.cpp

rouault:
I'm not sure which of QT 5.9.0 or gcc 5.5.0 is not happy, but with both
of them, I get:
```
/opt/qt59/include/QtCore/qflags.h: In instantiation of ‘QFlags<T>& QFlags<T>::setFlag(Enum, bool) [with Enum = QgsDxfExport::DxfPolylineFlag]’:
/home/even/qgis/QGIS/src/core/dxf/qgsdxfexport.cpp:1102:68:   required from here
/opt/qt59/include/QtCore/qflags.h:153:46: error: no match for ‘operator~’ (operand type is ‘QgsDxfExport::DxfPolylineFlag’)
         return on ? (*this |= f) : (*this &= ~f);
```


nyalldawson:
I think an easier fix is to add:

    Q_DECLARE_OPERATORS_FOR_FLAGS( QgsDxfExport::DxfPolylineFlags )

to the end of qgsdxfexport.h

rouault:
> Q_DECLARE_OPERATORS_FOR_FLAGS( QgsDxfExport::DxfPolylineFlags )

I just tried it, but that doesn't fix the compilation error.

rouault:
Actually, the error seems to come from "enum class DxfPolylineFlag : int". If i just use an enum, that works. Is the following acceptable:
```
diff --git a/src/core/dxf/qgsdxfexport.h b/src/core/dxf/qgsdxfexport.h
index 3463e7e..e2166ec 100644
--- a/src/core/dxf/qgsdxfexport.h
+++ b/src/core/dxf/qgsdxfexport.h
@@ -148,7 +148,7 @@ class CORE_EXPORT QgsDxfExport
      *
      * \since QGIS 3.12
      */
-    enum class DxfPolylineFlag : int
+    enum DxfPolylineFlag
     {
       Closed = 1, //!< This is a closed polyline (or a polygon mesh closed in the M direction)
       Curve = 2, //!< Curve-fit vertices have been added
@@ -626,5 +626,6 @@ class CORE_EXPORT QgsDxfExport
 };
 
 Q_DECLARE_OPERATORS_FOR_FLAGS( QgsDxfExport::Flags )
+Q_DECLARE_OPERATORS_FOR_FLAGS( QgsDxfExport::DxfPolylineFlags )
 
 #endif // QGSDXFEXPORT_H
```

rouault:
> Is the following acceptable:

I guess so. I just force-pushed this change

nyalldawson (approved):

m-kuhn:
Works for me too :+1: 

For reference, I made it a scoped enum because IIRC sip produces nicer python bindings (docs for the individual values). Which is also the reason for the currently failing tests, @rouault can you run sipify / the pre commit hook?

Issue #33121, opened 2019-11-28
QGIS 3.10.0-2 (proj 6.2.1) reprojects layers incorrectly while 3.10.0-1 (proj 5.2.0) does it okey.


denchat:
**Describe the bug**

When add google XYZ tiles into a UTM project, the tiles appears to be over-shifted to the right to the UTM layers.

**How to Reproduce**

Use `QGIS-OSGeo4W-3.10.0-2-Setup-x86_64.exe` or download nightly `OSGeo4W qgis-rel-dev`. It seems both executables change their dependencies from `PROJ 5.2.0` to `PROJ 6.2.1`.

**QGIS and OS versions**

QGIS:
- QGIS `3.10.0-2` (`PROJ 6.2.1`) [6ffa89eb3e] and QGIS `3.10.0-30` Nightly (`PROJ 6.2.1`) [60555a18fc] are affected.
- QGIS `3.10.0-1` (`PROJ 5.2.0`) [6c816b4204] is **_not_** affected.

OS:
- Windows 10 x64 1909

gioman:
@denchat duplicate of https://github.com/qgis/QGIS/issues/30569 ?

denchat:
I'm not sure if this issue and #33129 are from the same cause. So I attached some files to be investigated further.

For my case, it appears that incorrect reprojections occur with on-the-fly transformation between `WGS 1984` and `Indian 1975` datum

This attachment **_[[magnolias_waterfront_epsg4240.zip](https://github.com/qgis/QGIS/files/3904084/magnolias_waterfront_epsg4240.zip)]_** contains:
- `magnolias_waterfront_epsg4240.qgs`  project xml file
- `magnolias_waterfront_epsg4240.csv`  point WKT natively in `EPSG4240` (reprojected from `WGS84` coordinates)

The project contain a landmark in `Indian 1975` and google satellite tiles in `WGS84` datum. Project CRS is in `Indian 1975` datum.

When the project is opened in `3.10.0-1` [6c816b4204], the landmark is **_[in its place](https://goo.gl/maps/u5dT4MdF1i4iK6ry5)_**. :

![correct_reprojected](https://user-images.githubusercontent.com/19730041/69850291-8b12d500-12b1-11ea-80d8-25dccb60e5f2.jpg)



But when the project is opened in `3.10.0-2` [6ffa89eb3e]  or nightly release, the landmark appears shifted. Like the `WGS 1984` datum layer are on-the-fly reprojected incorrectly :

[...]

PR #33370, opened 2019-12-13
Fixes for handling manual coordinate operations specified in projects

nyalldawson:
Refs, and probably fixes, https://github.com/qgis/QGIS/issues/33121

With this branch, loading the project attached to #33121 results in the following default transformation:

image

This has been verified upstream at proj as being the correct default transformation result to use in this situation.

If the user manually opts to use an older, deprecated transformation via:

image

then the result is:

image

(as desired by the original poster)


nirvn:
@nyalldawson , that transformation dialog's canvas is perfectly sized for Thailand :) 

Great work!

Issue #33358, opened 2019-12-12
Ellipsoid option broken in QGIS 3.10.1 and wrong area calculations


PedroVenancio:
Hi,

I'm seeing a strange behaviour in the ellipsoid selection on QGIS 3.10.1.

On QGIS 3.4.14, I've the EPSG: 3763 as default CRS, and by default, the ellipsoid selected by QGIS is GRS 1980.

On QGIS 3.10.1, I've the same EPSG: 3763 as default CRS, but in this case QGIS uses the default ellipsoid as None/Planimetric. 

Besides this incoherence, the real problem is that I can not choose the ellipsoid in 3.10.1. For instance, if I change from None/Planimetric to GRS 1980 or WGS 84, the option selected is always changed to Airy 1830 (EPSG:7001). 

Please see this behaviour in the following screencast:

![ellipsoid_3_10](https://user-images.githubusercontent.com/1331581/70722195-c1f3dc80-1cee-11ea-90f2-636a06f189c9.gif)

As a side effect (I believe), the area calculation is affected on QGIS 3.10.1.

Even if the ellipsoid option is None/Planimetric, the result given by $area is the area over the ellipsoid. 

This does not happen on 3.4.14, that is:
- $area with Ellipsoid as None/Planimetric = area($geometry)
- $area with Ellipsoid as GRS 80 != area($geometry)

So, it seems that something is broken in QGIS 3.10.1.

QGIS 3.4.14 compiled against PROJ 5.2.0 and GDAL 2.4.3.
QGIS 3.10.1 compiled against PROJ 6.3.0 and GDAL 3.1.0dev.

Tested on Windows only for now.

gioman:
@PedroVenancio confirmed here too on Windows/3.10.1

nyalldawson:
Fixed in #33373

PedroVenancio:
Thank you very much @nyalldawson !

This one was not an easy one! :)

PR #33373, opened 2019-12-13
Fix initial selection in project ellipsoid list is reset whenever the dialog is reopened

nyalldawson:
Refs #33358


nyalldawson:
For reviewers: the old behavior was quite odd, and I'm unsure how it ever properly worked. It was only comparing the first portion of the ellipsoid acronym and matching based on the first encountered match for that alone (e.g. the first "epsg" registry when the ellipsoid is any epsg:*** code)

This particular function hasn't changed in years, so it must have been something else in this dialog which was originally correcting the initial setting (and has since been removed/changed, likely around the time of f3c268262f09ce54b9ffc52bd2fe737bd7533976 / 4cd2812905a9b72abd1dc98ac7945157b88db4dc

nyalldawson:
Turns out this was a big rabbit-hole, including required fixes to the projection selection widget and more...!

nyalldawson:
(@jef-n, a nice side-effect of the azure builds is that we now have ci testing a proj 6 enabled build :smile: )

PR #33378, opened 2019-12-13
Fix the label of the output parameter of the TPI algorithm

havatv:
## Description

The only change is that the label string has been corrected. So should be trivial.

Can / should be  backported to 3.4 - I can create a PR, if you like.

## Checklist


m-kuhn (approved):

m-kuhn:
Thank you

Shouldn't this be backported to 3.10 as well?

havatv:
> Thank you
> 
> Shouldn't this be backported to 3.10 as well?

Yes (I am used to the documentation site, and 3.10 is master there). Will do.

PR #33379, opened 2019-12-13
Move layerDependencies to formatter

elpaso:
... from the widget


elpaso:
@m-kuhn is that what you had in mind (still to be checked and polished!)

src/gui/editorwidgets/qgsrelationreferencewidgetwrapper.h
```diff
@@ -19,6 +19,7 @@
 #include "qgseditorwidgetwrapper.h"
 #include "qgis_sip.h"
 #include "qgis_gui.h"
+#include "qgsvectorlayerref.h"
```
m-kuhn: That's probably a leftover?

m-kuhn (commented):
Yes, that looks exactly like what I had in mind. Thanks for the quick reaction!

src/gui/editorwidgets/qgsrelationreferencewidgetwrapper.h
elpaso: yes, thanks, I said it needed polishing (ans tests!)

m-kuhn: sure, just bumped into it.
I hope the tests can be easily adapted.

elpaso: I'm not sure, QgsVectorLayerRef is not exposed to Python so I need a C++ test and there aren't any for the formatters.

m-kuhn: just copying / renaming the ones you did for widgetwrappers won't work?

elpaso: I'm creating a new `testqgsrelationreferencefieldformatter.cpp` and moving the old test into it.

PR #33388, opened 2019-12-13
[mesh] [API] get 3d stacked raw data values

PeterPetrik:
add function to api to be able to get raw 3d values for particular point on map similarly to API for 2d values we already have

with this.change there is debugging graph in crayfish, so one can find out what values are averaged: see https://github.com/lutraconsulting/qgis-crayfish-plugin/pull/416

Screenshot 2019-12-13 at 16 29 01

### DIFF
--- a/src/core/mesh/qgsmeshlayer.h
+++ b/src/core/mesh/qgsmeshlayer.h
@@ -246,6 +246,23 @@ class CORE_EXPORT QgsMeshLayer : public QgsMapLayer
       */
     QgsMeshDatasetValue datasetValue( const QgsMeshDatasetIndex &index, const QgsPointXY &point ) const;
 
+    /**
+      * Returns the 3d values of stacked 3d mesh defined by the given point
+      *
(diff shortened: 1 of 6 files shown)


src/core/mesh/qgsmeshlayer.h
```diff
@@ -246,6 +246,23 @@ class CORE_EXPORT QgsMeshLayer : public QgsMapLayer
...
+    /**
+      * Returns the 3d values of stacked 3d mesh defined by the given point
+      *
+      * \note It uses previously cached and indexed triangular mesh
+      * and so if the layer has not been rendered previously
+      * (e.g. when used in a script) it returns NaN value
+      *
+      * \param index dataset index specifying group and dataset to extract value from
+      * \param point point to query in map coordinates
+      * \returns all 3d stacked values that belongs to face defined by given point. Returns invalid block
```
wonder-sk: s/belongs/belong/

wonder-sk (approved):

PR #33394, opened 2019-12-13
[UI] Use 'WFS / OGC API - Features' instead of WFS in a few labels

rouault:


m-kuhn:
Is there no abbreviation for `OGC API - Features`?
I am just imagining myself on the phone explaining a client to click the "WFS / OGC API - Features" button - and some almost unreadable menu entries.

elpaso:
@m-kuhn yes, there is an OAPIF ugly acronym, I think we should have stuck to WFS3 but looks like I'm alone.
@rouault I agree it's a bit long, especially for the datasource manager, does it generally fit in the tabs column (tabs) without flowing in the next line?

m-kuhn:
If WFS3 works for 95% that's good enough for me, for what it's worth ;)

rouault:
Seems a number of people in the OGC API - Features community also uses OAFeat : https://medium.com/radiant-earth-insights/ogc-api-features-stac-sprint-recap-6c876b44c9d2

m-kuhn:
If we leave it simply as WFS. Is there a risk people will not find it?

nyalldawson:
I'd be +1 on just keeping WFS too, until such time as the wider community standardises on a "common name" for these services (I suspect it'll end up being "wfs3"!). All the alternatives proposed so far are either ridiculously verbose or ugly looking acronyms.

rouault:
> All the alternatives proposed so far are either ridiculously verbose or ugly looking acronyms.

I think OAFeat is the current trend among the folks developing the standard

rouault:
Dropping that idea

Issue #33232, opened 2019-12-04
DB Manager: New SQL layer from GeoPackage has wrong geometry type declared


borysiasty:
Not sure if it's more DB Manager or OGR provider problem.

In the DB Manager, there is the "Load as new Layer" function in its SQL query tab. It works fine with PostGIS and Spatialite, but the GeoPackage plugin (i.e. the ogr provider) builds a layer with wrong CRS and geometry type declared (I bet it matches the first layer in the file).

**How to Reproduce**

1. Open the SQL query tab for a random GeoPackage connection and build a new layer from a simple query, e. g.:

```
SELECT makepoint(0,0) as geom
SELECT makeline(makepoint(0,0), makepoint(1,1)) as geom
```

2. Note the created layer's geometry type and CRS depend on the GeoPackage, not the query.

**Where's the point**

The code creating the layer is here: https://github.com/qgis/QGIS/blob/master/python/plugins/db_manager/db_plugins/gpkg/plugin.py#L181

You can also reproduce it in the Python console:

```
layer = QgsVectorLayer(filePath, 'Bad Layer', 'ogr')
layer.setSubsetString('SELECT makepoint(0,0) as geom')
QgsProject.instance().addMapLayer(layer)
```

In both cases the result is a layer with wrong geometry type (unless the first layer in the file happens to have the desired type of course). The layer uri is `fileName|layerid=0|subset=SELECT makepoint(0,0)`.

What is interesting, if you put the query directly to the layer uri instead of calling `setSubsetString`, the resulting layer is correct:

```
l = QgsVectorLayer(filePath + '|subset=SELECT makepoint(0,0) as geom', 'Good Layer', 'ogr')
QgsProject.instance().addMapLayer(l)
```

Furthermore, you can even include "layerid=0" to the uri, and it still works.

A dirty hack would be:
```
-       vl = QgsVectorLayer(self.uri().database(), layerName, 'ogr')
-       vl.setSubsetString(sql)
+       vl = QgsVectorLayer(self.uri().database() + '|subset=' + sql, layerName, 'ogr')
```

but probably it deserves a deeper fix in the ogr provider, so whoever has an idea, feel free to rename this ticket.

PR #33395, opened 2019-12-13
[DB Manager] Fix broken geometry type of GeoPackage-based SQL Layers

borysiasty:
This is a quick fix for #33232. Probably it could be better fixed in the OGR provider, but I don't have enough resources now, while this patch simply works (it's what I recently start SQL trainings from... :/).

### DIFF
--- a/python/plugins/db_manager/db_plugins/gpkg/plugin.py
+++ b/python/plugins/db_manager/db_plugins/gpkg/plugin.py
@@ -178,8 +178,7 @@ class GPKGDatabase(Database):
     def toSqlLayer(self, sql, geomCol, uniqueCol, layerName="QueryLayer", layerType=None, avoidSelectById=False, filter=""):
         from qgis.core import QgsVectorLayer
 
-        vl = QgsVectorLayer(self.uri().database(), layerName, 'ogr')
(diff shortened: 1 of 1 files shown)


nyalldawson:
This fix looks fine to me -- setSubsetString was never designed to change the geometry type of an existing layer (and adapting it to allow this would be very dangerous IMO)

@elpaso thoughts?

elpaso:
Agreed, `setSubsetString` was designed to set a `WHERE` condition to limit the features of a layer to a those passing the condition, I suppose OGR knows what to do when constructing a layer directly from a an expression starting with a `SELECT` (that is not a subset expression anymore but defines the whole layer).

The documentation of the function is not completely clear though:

```
     * Sets the string (typically sql) used to define a subset of the layer
     * \param subset The subset string. This may be the where clause of a sql statement
     *               or other definition string specific to the underlying dataprovider
     *               and data store.
```

because it suggests that the argument can be a generic "other definition string", it is clear about being a subset of the layer (an existing layer, I suppose). Perhaps @nyalldawson could help to find a better wording.

borysiasty:
Btw. the failed test is not relevant.

PR #33404, opened 2019-12-15
[needs-docs] Add precision to rotate marker checkbox label

DelazJ:
"to follow line direction" and harmonize with hash line symbol layer dialog
currently
image

the hash example 
image


roya0045:
isn't that the same as simply rotating the marker?

DelazJ:
@roya0045 not really. these are two different things. I'd say it overloads the marker own rotation. I can't (yet) explain the subtleties with words other than suggesting to give a test by yourself.

DelazJ:
@nyalldawson do you think that we should backport this label change?

PR #33405, opened 2019-12-15
Make sure a TOWGS84 parameter is added when exporting a CRS to a WKT string with Proj6

nirvn:
## Description

This PR insures that CRSes represented as WKT strings have a TOWGS84 parameter present. This ensures that our QgsCoordinateReferenceSystem::toWkt() outputs strings matching the default behavior of gdal3 and proj6's projinfo when creating projections from EPSG codes.

@rouault , is there any downside to forcing a TOWGS84 parameter onto all exported wkt strings?

## Checklist

### DIFF
--- a/src/core/qgscoordinatereferencesystem.cpp
+++ b/src/core/qgscoordinatereferencesystem.cpp
@@ -105,6 +105,20 @@ QString getFullProjString( PJ *obj )
 
   return QString( proj_as_proj_string( QgsProjContext::get(), obj, PJ_PROJ_4, nullptr ) );
 }
+
+QString getFullWktString( PJ *obj, PJ_WKT_TYPE type, const char *const options[] )
+{
+  QgsProjUtils::proj_pj_unique_ptr boundCrs( proj_crs_create_bound_crs_to_WGS84( QgsProjContext::get(), obj, nullptr ) );
+  if ( boundCrs )
+  {
+    if ( const char *wkt =  proj_as_wkt( QgsProjContext::get(), boundCrs.get(), type, options ) )
+    {
+      return QString( wkt );
+    }
+  }
+
+  return QString( proj_as_wkt( QgsProjContext::get(), obj, type, options ) );
+}
 #endif
 //--------------------------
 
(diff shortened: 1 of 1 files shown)


nirvn:
An alternative approach would be to force towgs84 parameters *only* when a CRS in created from an EPSG code.

rouault:
>  is there any downside to forcing a TOWGS84 parameter onto all exported wkt strings?

Yes. This will likely make this default TOWGS84 to be used for coordinate transformations whereas there might be other datum transformations that the user might want to select.

I'm not sure this was the right thing to do in GDAL's importFromEPSG(). I was probably afraid to break too much backward compatibility, but now I'm wondering if I shouldn't reconsider that... That said, the impact on GDAL coordinate transformation of this behaviour is limited, since it has logic in https://github.com/OSGeo/gdal/blob/a90b5e8c25abb14ba614c6f5f971d90775cc11fa/gdal/ogr/ogrct.cpp#L872 to detect if a BoundCRS of an object with an EPSG code is the same as the 'canonical' BoundCRS returned by importFromEPSG() with the default TOWGS84. If that's the case, it just uses the EPSG:XXXX representation when researching the coordinate transformation, so this basically cancels the addition of the TOWGS84[] clause.

Regarding PROJ itself / projinfo, it only triggers proj_crs_create_bound_crs_to_WGS84() if exporting to WKT1.

So perhaps restrict this change when type == PJ_WKT1_GDAL

nyalldawson:
I'm rather of the opinion that we should just "rip off this bandaid", and avoid the workarounds. I'm concerned about the potential long term impact of workarounds like this...

nirvn:
@nyalldawson , as in removing the proj4 bandaid too?

Will that create issues reopening, with a proj5 build, a project saved with a proj6 build?

nyalldawson:
> Will that create issues reopening, with a proj5 build, a project saved with a proj6 build?

Who knows. If it matches a known CRS then no, because the auth/code method has always been preferred over proj strings. If it was a custom proj string, then... possibly? I don't know how proj 5 would be impacted by the lack of this parameter. 

This whole thing is giving off the stink of a workaround that we'll be pasting band aids on top of for years in future. That's why I'd rather just take a hardline approach of "we rely on proj and don't do downstream hacks anymore" and leave everything at the defaults coming from the proj library.

nyalldawson:
@nirvn let's close this. I'm going to rip out all downstream workarounds which have snuck in to the proj 6 handling in qgis, and we'll take the stance that we implicitly trust the upstream proj results and don't do any downstream mangling.

nirvn:
Okay

PR #33409, opened 2019-12-15
Select features by value: show layer name in window title

### LINKED ISSUE #33210: Select feature by value
which layer is reported ??

image

it would be useful to indicate the layer, as for the attribute tables

image

elpaso:
Fixes #33210

### DIFF
--- a/src/app/qgsselectbyformdialog.cpp
+++ b/src/app/qgsselectbyformdialog.cpp
@@ -47,7 +47,7 @@ QgsSelectByFormDialog::QgsSelectByFormDialog( QgsVectorLayer *layer, const QgsAt
 
   QgsGui::enableAutoGeometryRestore( this );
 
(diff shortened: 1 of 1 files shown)


src/app/qgsselectbyformdialog.cpp
```diff
@@ -47,7 +47,7 @@ QgsSelectByFormDialog::QgsSelectByFormDialog( QgsVectorLayer *layer, const QgsAt
 
   QgsGui::enableAutoGeometryRestore( this );
 
-  setWindowTitle( tr( "Select Features by Value" ) );
+  setWindowTitle( tr( "%1 - Select Features by Value" ).arg( layer->name() ) );
```
nyalldawson: Maybe we should drop "by value"? It's rather redundant

nyalldawson (approved):
It's the little things which make a huge difference!!

Issue #33399, opened 2019-12-14
[mesh] Changing Reference date is broken


saberraz:
To reproduce the issue:
1- Add a mesh data (e.g. https://github.com/lutraconsulting/MDAL/raw/master/tests/data/netcdf/indonesia_nc3.nc)
2- Change the reference date <<< it is not doing anything
3- Changing time does work.
![Peek 2019-12-14 09-16](https://user-images.githubusercontent.com/2649047/70846613-e5d03300-1e52-11ea-92e7-47bd261629b9.gif)

saberraz:
Maybe related to https://github.com/qgis/QGIS/issues/31934

Issue #31933, opened 2019-09-22
[mesh] Better default for Absolute time


pcav:
The default for absolute time should be the current time rather than 2019-01-01

PR #33410, opened 2019-12-15
[mesh] Fixes time reference for mesh layer #32186 #33399 #31933

### LINKED ISSUE #32186: use default reference time from MDAL_G_referenceTime
in MDAL 0.4.x there is new function https://github.com/lutraconsulting/MDAL/issues/152 `MDAL_G_referenceTime`. The output of this function should be used as default value for absolute time 

related to #31933 and #31934

vcloarec:
fixes #32186 #33399 #31933
since 6e2ef1544f14c7a5f700447825780f5d93c3c927, MDAL provide time reference for dataset.
This PR allows QGis to read this time reference for mesh layer
If there is a valid time reference provided with dataset groups, this time reference is used to displaying time (using absolute time).
If there is no reference time provided, the time is displayed using relative time, and a time reference can be set by the user to display absolute time.
When no time reference is provided, default one is current date + time set with 00:00:00.
A push button is added to reload the reference time provided with dataset groups, if needed.

Also adds a new feature to let the user set the time unit of the provider with a combo box if thise time unit is different than hours.

image

### DIFF
--- a/src/app/mesh/qgsmeshtimeformatdialog.cpp
+++ b/src/app/mesh/qgsmeshtimeformatdialog.cpp
@@ -30,13 +30,17 @@ QgsMeshTimeFormatDialog::QgsMeshTimeFormatDialog( QgsMeshLayer *meshLayer, QWidg
 
   loadSettings();
 
+  mReloadReferenceTimeButton->setEnabled( layerHasReferenceTime() );
+
   connect( mUseTimeComboBox, qgis::overload<int>::of( &QComboBox::currentIndexChanged ), this, &QgsMeshTimeFormatDialog::saveSettings );
-  connect( mReferenceDateTimeEdit, &QDateTimeEdit::timeChanged, this, &QgsMeshTimeFormatDialog::saveSettings );
+  connect( mReferenceDateTimeEdit, &QDateTimeEdit::dateTimeChanged, this, &QgsMeshTimeFormatDialog::saveSettings );
   connect( mAbsoluteTimeFormatComboBox, qgis::overload<int>::of( &QComboBox::currentIndexChanged ), this, &QgsMeshTimeFormatDialog::saveSettings );
   connect( mUseTimeComboBox, qgis::overload<int>::of( &QComboBox::currentIndexChanged ), this, &QgsMeshTimeFormatDialog::saveSettings );
   connect( mRelativeTimeFormatComboBox, qgis::overload<int>::of( &QComboBox::currentIndexChanged ), this, &QgsMeshTimeFormatDialog::saveSettings );
   connect( mOffsetHoursSpinBox, qgis::overload<double>::of( &QDoubleSpinBox::valueChanged ), this, &QgsMeshTimeFormatDialog::saveSettings );
   connect( mPlaybackIntervalSpinBox, qgis::overload<double>::of( &QDoubleSpinBox::valueChanged ), this, &QgsMeshTimeFormatDialog::saveSettings );
+
+  connect( mReloadReferenceTimeButton, &QPushButton::clicked, this, &QgsMeshTimeFormatDialog::reloadLayerReferenceTime );
 
 }
 
@@ -54,7 +58,11 @@ void QgsMeshTimeFormatDialog::loadSettings()
     mUseTimeComboBox->setCurrentIndex( 1 );
   }
 
-  mReferenceDateTimeEdit->setDateTime( settings.absoluteTimeReferenceTime() );
+  if ( settings.absoluteTimeReferenceTime().isValid() ) //set the reference time, if not valid, set the current date + time 00:00:00
+    mReferenceDateTimeEdit->setDateTime( settings.absoluteTimeReferenceTime() );
+  else
+    mReferenceDateTimeEdit->setDateTime( QDateTime( QDate::currentDate(), QTime( 00, 00, 00 ) ) );
+
   mReferenceDateTimeEdit->setDisplayFormat( settings.absoluteTimeFormat() );
 
   int index = mAbsoluteTimeFormatComboBox->findText( settings.absoluteTimeFormat() );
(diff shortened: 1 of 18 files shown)


src/app/mesh/qgsmeshtimeformatdialog.cpp
```diff
@@ -54,7 +58,11 @@ void QgsMeshTimeFormatDialog::loadSettings()
     mUseTimeComboBox->setCurrentIndex( 1 );
   }
 
-  mReferenceDateTimeEdit->setDateTime( settings.absoluteTimeReferenceTime() );
+  if ( settings.absoluteTimeReferenceTime().isValid() ) //set the reference time, if not valid, set the current date + time 00:00:00
```
PeterPetrik: if you can move comment to separate line (line too wide), add 3rd person ("sets") and whitespace after //

src/app/mesh/qgsmeshtimeformatdialog.cpp
```diff
@@ -96,4 +104,22 @@ void QgsMeshTimeFormatDialog::enableGroups( bool useAbsoluteTime )
   mRelativeTimeGroupBox->setEnabled( ! useAbsoluteTime );
 }
 
+bool QgsMeshTimeFormatDialog::layerHasReferenceTime() const
+{
+  if ( !mLayer )
+    return false;
+  int groupCount = mLayer->dataProvider()->datasetGroupCount();
+  bool found = false;
```
PeterPetrik: you can remove this variable and simplify this to 

``` 
for ( int i=0; i< mLayer->dataProvider()->datasetGroupCount(); ++i)
  if (mLayer->dataProvider()->datasetGroupMetadata( i ).referenceTime().isValid())
    return true;

return false;
```

PeterPetrik: thinking about it, maybe you can return here the index of dataset group that HAS the reference time, and -1 if not found. that way you do not need to search second time for this group

src/app/mesh/qgsmeshtimeformatdialog.h
```diff
@@ -39,12 +39,18 @@ class APP_EXPORT QgsMeshTimeFormatDialog: public QDialog, private Ui::QgsMeshTim
     ~QgsMeshTimeFormatDialog();
 
   private slots:
-
+    void reloadLayerReferenceTime()
```
PeterPetrik: move to cpp

src/core/mesh/qgsmeshlayer.cpp
```diff
@@ -165,6 +165,33 @@ void QgsMeshLayer::setTimeSettings( const QgsMeshTimeSettings &settings )
   emit timeSettingsChanged();
 }
 
+void QgsMeshLayer::setDefaultTimeSettings()
+{
+  mTimeSettings.setRelativeTimeOffsetHours( 0 );
+
+  //search for the first valid reference time in the dataset groups
```
PeterPetrik: add whitespace after //

src/core/mesh/qgsmeshlayer.h
```diff
@@ -221,6 +221,20 @@ class CORE_EXPORT QgsMeshLayer : public QgsMapLayer
      */
     void setTimeSettings( const QgsMeshTimeSettings &settings );
 
+    /**
+     * Sets time default format settings
+     *
+     * \since QGIS 3.12
+     */
+    void setDefaultTimeSettings();
```
PeterPetrik: this function probably does not fir into qgsmeshlayer class. I would create some helper (ideally not exported out of CORE) function somewhere in mesh layer utils or the class that calls setTimeSettings. so the flow would be

1) resolve refence time
2) assign to QgsMeshTimeSettings
3) call meshlayer->setTimeSettings

src/core/mesh/qgsmeshlayer.h
```diff
@@ -221,6 +221,20 @@ class CORE_EXPORT QgsMeshLayer : public QgsMapLayer
...
+
+    /**
+     * Loads the first valid reference time of the dataset groups.
+     *
+     * \since QGIS 3.12
+     */
+    void loadReferenceTime();
```
PeterPetrik: this function probably does not fir into qgsmeshlayer class

src/providers/mdal/qgsmdalprovider.cpp
```diff
@@ -493,13 +493,17 @@ QgsMeshDatasetGroupMetadata QgsMdalProvider::datasetGroupMetadata( int groupInde
     metadata[key] = value;
   }
 
+  QString referenceTimeString( MDAL_G_referenceTime( group ) );
+  QDateTime referenceTime = QDateTime::fromString( referenceTimeString, Qt::ISODate );
```
PeterPetrik: what if referenceTimeString is empty string?

PeterPetrik (changes requested):

PeterPetrik:
very good work, just some small changes requested

src/providers/mdal/qgsmdalprovider.cpp
vcloarec: If the static method of QDateTime fails to parse the string, it returns invalid QDateTime and we considers that the dataset group don't have valid reference time.

src/core/mesh/qgsmeshlayer.h
vcloarec: ok, I simplified that with only a new method in qgsmeshutils. This method return the first valid reference time of the dataset goups.
This function is called from the QgsMeshTimeFormatDialog (only when needed)  but also from the QgsMeshLayer class just after the provider is set. I think this last call is the best place to set the default reference time for those reasons :
- default reference time is defined when the mesh layer is created. Whatever needs the reference time (time widget, active renderer widget, other future stuff), the default reference time is defined from the layer's data.
- When the mesh layer is created when opening project, the read xml method can replace the default time reference with the one saved

src/core/mesh/qgsmeshlayerutils.h
```diff
@@ -178,6 +178,15 @@ class CORE_EXPORT QgsMeshLayerUtils
      * Formats hours in human readable string based on settings
      */
     static QString formatTime( double hours, const QgsMeshTimeSettings &settings );
+
+    /**
+      * Searches and returns the first valid reference time in layer's dataset group
+      * \param meshLayer mesh layer to parse
+      *
+      * * \since QGIS 3.12
```
PeterPetrik: extra *

src/core/mesh/qgsmeshtimesettings.h
```diff
@@ -35,7 +35,19 @@
 class CORE_EXPORT QgsMeshTimeSettings
 {
   public:
-    //! Default constructor for relative time format and 0 offset
+    //! Time units used to display time
```
PeterPetrik: add `\since QGIS 3.12`

src/core/mesh/qgsmeshtimesettings.h
```diff
@@ -84,6 +96,11 @@ class CORE_EXPORT QgsMeshTimeSettings
     //! Sets format used for absolute time
     void setAbsoluteTimeFormat( const QString &absoluteTimeFormat );
 
+    //! Returns the provider time unit
```
PeterPetrik: \since QGIS 3.12

src/core/mesh/qgsmeshtimesettings.h
```diff
@@ -84,6 +96,11 @@ class CORE_EXPORT QgsMeshTimeSettings
...
+    TimeUnit providerTimeUnit() const;
+    //! Sets the provider time unit
```
PeterPetrik: \since QGIS 3.12

PeterPetrik (changes requested):
final doc changes needed...

PeterPetrik (approved):

PR #33411, opened 2019-12-15
Don't highlight whole polygon in red when using map label tools

nyalldawson:
For polygons, we don't want to fill the whole polygon itself with the rubber band as that obscures too much of the map and prevents users from getting a good view of the underlying map.

Instead, just use the boundary of the polygon for the rubber band. That's enough to give a visual indicator of the feature associated with the label without getting in the way.


nirvn:
Yes please.

nyalldawson:
Thanks @nirvn!

PR #33419, opened 2019-12-16
[DB Manager] Fix broken geometry type of GeoPackage based SQL Layers (backport from master)

borysiasty:
Cherry-picked 94e507f83ab4b7 to 3.10


borysiasty:
The failed test is proj-related and has nothing to do with that change.

PR #33423, opened 2019-12-16
List referenced layer values in Expression Builder

signedav:
When a field is an foreign key and configured as a relation reference widget, there is the possibility not only to show the values of the current layer but also the possible values in the referenced layer.

referencedfields

In the example we have the persons:
- George (in Cuba)
- Paul (in Vietnam)
- Ringo (in Venezuela)
- John (in Vietnam as well)

And the entries in the country layer are USSR, Cuba, Vietnam, Burma, Venezuela, North Korea

### DIFF
--- a/src/gui/qgsexpressionbuilderwidget.cpp
+++ b/src/gui/qgsexpressionbuilderwidget.cpp
@@ -484,8 +484,12 @@ void QgsExpressionBuilderWidget::fillFieldValues( const QString &fieldName, int 
   // if it's a request for the values of the referenced layer
   if ( cbxRelatedLayerValues->isChecked() && setup.config().contains( QStringLiteral( "Relation" ) ) )
   {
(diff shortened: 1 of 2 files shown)


signedav:
List layer values in Expression Builder

When a field is an foreign key and configured as a relation reference widget, there is the possibility not only to show the values of the current layer but also the possible values in the referenced layer.

In the example we have the persons:

George (in Cuba, in his twenties, human)
Paul (in Vietnam, in his thirties, human)
Ringo (in Venezuela, in his forties, cat)
John (in Vietnam as well, in his forties, table)
And the entries in the country layer are USSR, Cuba, Vietnam, Burma, Venezuela, North Korea

PS. Not sure about the use of `QgsProject::instance()` in the Widget-Formatter versus passing the mProject as argument. Since `QgsProject::instance()` is already used sometimes in the formatters, I used it as well. But I'm open to change.

PR #33425, opened 2019-12-16
Some label fixes in Processing dialogs

DelazJ:
## Checklist


src/analysis/processing/qgsalgorithmjoinbyattribute.cpp
```diff
@@ -79,13 +79,13 @@ void QgsJoinByAttributeAlgorithm::initAlgorithm( const QVariantMap & )
   addParameter( new QgsProcessingParameterFeatureSink( QStringLiteral( "OUTPUT" ), QObject::tr( "Joined layer" ), QgsProcessing::TypeVectorAnyGeometry, QVariant(), true, true ) );
 
   std::unique_ptr< QgsProcessingParameterFeatureSink > nonMatchingSink = qgis::make_unique< QgsProcessingParameterFeatureSink >(
-        QStringLiteral( "NON_MATCHING" ), QObject::tr( "Unjoinable features from first layer" ), QgsProcessing::TypeVectorAnyGeometry, QVariant(), true, false );
+        QStringLiteral( "NON_MATCHING" ), QObject::tr( "Unjoinable features from input layer" ), QgsProcessing::TypeVectorAnyGeometry, QVariant(), true, false );
```
alexbruy: There are two input layers here, I thinnk that's why it explicitly says which one is used

DelazJ: I see the point but the first layer is called "input layer" and the second "input layer 2". Also the first layer is later referred to as "input table(/layer)" so if it's sufficient in that case, it should be good here too. No?
my point is that we should identify the layer the same way everywhere otherwise people could get lost.

PR #33422, opened 2019-12-16
Fix selective masking regression

mhugo:
## Description

Fix for regressions discussed in https://github.com/qgis/QGIS/pull/30747 and possibly in https://github.com/qgis/QGIS/issues/32791

Some regressions also existed in unit tests, which I fixed, but I am not sure why it happened : some recent changes in label placements ?

## Checklist

- [ ] Commit messages are descriptive and explain the rationale for changes
- [ ] Commits which fix bugs include `Fixes #11111` at the bottom of the commit message
- [ ] I have read the [QGIS Coding Standards](https://docs.qgis.org/testing/en/docs/developers_guide/codingstandards.html) and this PR complies with them
- [ ] New unit tests have been added for core changes
- [ ] I have run [the `scripts/prepare-commit.sh` script](https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTING.md#contributing-to-qgis) before each commit
- [ ] I have evaluated whether it is appropriate for this PR to be backported, backport requests are left as label or comment

PR #33432, opened 2019-12-17
[labelling] Fix masked symbol layers loss when

nirvn:
## Description
Fix masked symbol layers loss when:
- opening layer properties window; or
- changing label settings in style dock

@mhugo , here's a little bit of help bullet proofing the masking feature.

Other than masked layer loss (fixed by the PR) and the canvas flashing / label blending regression (fixed by your PR #33422 ), the other follow up that'd need your attention is the generic implementation issue of force rasterizing vector layers that are not being masked. Every time we're needlessly force rasterizing symbol layers, an angel is crying somewhere :wink: that'd be worth addressing _if_ possible.

## Checklist

### DIFF
--- a/python/gui/auto_generated/qgstextformatwidget.sip.in
+++ b/python/gui/auto_generated/qgstextformatwidget.sip.in
@@ -159,6 +159,8 @@ Controls whether data defined alignment buttons are enabled.
 
 
 
+
+
   protected slots:
 
     void updateLinePlacementOptions();
--- a/src/gui/qgstextformatwidget.cpp
+++ b/src/gui/qgstextformatwidget.cpp
@@ -843,6 +843,7 @@ void QgsTextFormatWidget::updateWidgetForFormat( const QgsTextFormat &format )
   mBufferEffectWidget->setPaintEffect( mBufferEffect.get() );
 
   // mask
+  mMaskedSymbolLayers = mask.maskedSymbolLayers();
   mEnableMaskChkBx->setChecked( mask.enabled() );
   mMaskBufferSizeSpinBox->setValue( mask.size() );
   mMaskBufferUnitWidget->setUnit( mask.sizeUnit() );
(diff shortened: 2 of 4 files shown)


mhugo:
@nirvn Thanks, I was looking into the UI problem, but your solution seems cleaner than mine, I'll have a look.

mhugo:
@nirvn And about the forced rasterization ... I will think about it, but I am not sure it is easy to solve

nirvn:
@mhugo , have you ever noticed the unwanted artifacts created when masking symbol layers, for e.g.:
Screenshot from 2019-12-17 14-38-26

There's also some weirdness going on when lowering the opacity, see:
Screenshot from 2019-12-17 14-40-39

mhugo (approved):

mhugo:
@nirvn Thanks for this PR, good for me !

About the "phantom" line around the mask buffer, indeed, I can see it with non white backgrounds and a big magnification :-/ 
However, I cannot reproduce the second issue with the opacity.

mhugo:
@nirvn About the first issue, disabling anti-aliasing when rendering the mask buffer seems an easy fix, but of course edges are then a bit "sharper". What do you think ?

With antialiasing (current code):
with_antialiasing

Without:
without_antialiasing

nirvn:
@mhugo , yeah, I tried the antialiasing trick as a quick workaround, but I wasn't satisfied with the degradation resulting in switching antialiasing off on low-dpi (i.e. on-screen 96dpi) display.

mhugo:
@nirvn I meant if the antialiasing is turned off **only** for the mask buffer, not for the whole map

nirvn:
@mhugo , yeah, tried that. It's a subtle change, maybe I'm overly sensitive :)

nyalldawson:
@mhugo have you tried using QPictures for the mask storage instead of QImages? Might be a bit slower, but would likely resolve this issue and also give full vector outputs (including for the mask itself)

mhugo:
@nyalldawson Yeah, having a second look to QPicture might be worth it ...
But I am afraid it would need some non-trivial refactoring work.

In the meantime, @nirvn is a trade ever possible for the first issue ? :-) What would you accept as a temporary solution: phantom lines or aliased edges ?

And about the second issue you raised when lowering the opacity, do you have some blending modes or effects involved in the rendering stack ?

nirvn:
I'd stick with phantom lines for now. Aliasing is sooo esri ;)

PR #33433, opened 2019-12-17
Rework all QGIS boundcrs related logic

nyalldawson:
Trying to handle this on the QGIS side is dangerous -- we need to instead just trust what we get from proj and don't try to second-guess the results!

Potentially, this may cause some regressions, as boundcrs objects which previously were identified as known CRSes (e.g. EPSG:...) will no longer be, and will instead become user crses. It's both a cosmetic issue and potentially a source of projection errors (e.g. when a GDA2020 boundcrs isn't converted to the source crs then the available coordinate operations miss the grid based methods). But the alternative is that we incorrectly equate boundcrses with their underlying source crses, which also leads to incorrect projection result, and a lot of projection related logic in QGIS.

So this PR goes with the first approach, and drops all the QGIS side logic relating to these (on proj 6 builds only!!). Wherever we encounter CRSes which are incorrectly treated as user crses instead of standard auth:id CRSes, we can file tickets at proj and get the issues fixed upstream...


src/core/qgscoordinatereferencesystem.cpp
```diff
@@ -824,12 +824,19 @@ bool QgsCoordinateReferenceSystem::createFromProj4( const QString &proj4String )
   }
 #endif
 
-  // IDEALLY!!
-  // don't do any of this for proj 6 -- responsibility for all this rests in the proj library.
-  // Woohoo! we can be free of this legacy cruft FOREVER!
-  // (and if any of this has value, take it up with the proj project. That's where it belongs)
-  // I ***think*** this is safe to disable for proj 6.1.1 and above
-#if 1
+#if PROJ_VERSION_MAJOR>=6
```
nirvn: Are you sure you meant for it to be PROJ_VERSION_MAJOR>=6 and not <6 ?

nyalldawson: No the logic below that is new proj 6 specific code, the rest was moved to an #else

nirvn: Ah got it, sorry for the noise.

nyalldawson:
This has become a mega PR, which should be the bulk of the remaining changes required to fix all outstanding proj 6 related issues.

I've also:
- aliases the "proj4" members from the QgsCoordinateReferenceSystem api and deprecated the original ones (since proj4 isn't correct, it's "proj")
- deprecated most of the unsafe methods of constructing CRS objects, i.e. those which rely on internal crs ids or postgis srids. I haven't been able to completely deprecate and remove all the use of all the methods as some are still very deeply ingrained in other parts of the code. I'll keep chipping away at this over time though
- synchronized the logic to provide a unified logic with gdal, based on https://github.com/OSGeo/gdal/pull/2113
- removed all forced saving of unknown crs as user crs, and left it so that only crs which users have actually manually created themselves are saved as user crs

nirvn:
@nyalldawson , things behaving as they should under proj5, will test things out under proj6 whenever osgeo4w's qgis-dev builds include those commits. 

Fantastic efforts increasing reliability of QGIS under brand-new proj6. And thanks to @rouault for being so responsive throughout this fix sprint. I wouldn't want to run a gdal/proj environment pre all the changes and fixes done in the last 5-6 days.

On that note, would it be worth making soon-to-be-released PROJ 6.3 be the minimum version requirement?

nyalldawson:
Unrelated test failure

PR #33437, opened 2019-12-17
[feature] Stored expressions

elpaso:
- Store user expressions (with context/category, just like recent expressions)
- Clear editor icon/action

expression-storage-opt

This PR implements part of QEP https://github.com/qgis/QGIS-Enhancement-Proposals/issues/135

### DIFF
--- a/python/gui/auto_generated/qgsexpressionbuilderwidget.sip.in
+++ b/python/gui/auto_generated/qgsexpressionbuilderwidget.sip.in
@@ -217,14 +217,38 @@ Registers a node item for the expression builder.
 
     void saveToRecent( const QString &collection = "generic" );
 %Docstring
-Adds the current expression to the given collection.
+Adds the current expression to the given ``collection``.
 By default it is saved to the collection "generic".
 %End
 
     void loadRecent( const QString &collection = "generic" );
 %Docstring
-Loads the recent expressions from the given collection.
+Loads the recent expressions from the given ``collection``.
 By default it is loaded from the collection "generic".
+%End
+
+    void loadStored( const QString &collection = "generic" );
+%Docstring
+Loads the stored expressions from the given ``collection``.
+By default it is loaded from the collection "generic".
+
+.. versionadded:: 3.12
+%End
+
+    void saveToStored( const QString &label, const QString expression, const QString &helpText, const QString &collection = "generic" );
+%Docstring
+Stores the ``expression`` in the given ``collection`` with given ``label`` and ``helpText``.
+By default it is saved to the collection "generic".
+
+.. versionadded:: 3.12
+%End
+
+    void removeFromStored( const QString &name, const QString &collection = "generic" );
+%Docstring
+Removes the expression ``name`` from the stored expressions in the given ``collection``.
+By default it is removed from the collection "generic".
+
+.. versionadded:: 3.12
 %End
 
     void newFunctionFile( const QString &fileName = "scratch" );
@@ -318,6 +342,21 @@ Enabled or disable auto saving. When enabled Python scripts will be auto saved
 when text changes.
 
 :param enabled: ``True`` to enable auto saving.
+%End
+
+    void storeCurrentExpression( );
+%Docstring
+Adds the current expressions to the stored expressions.
+
+.. versionadded:: 3.12
+%End
+
+    void removeSelectedExpression( );
+%Docstring
+Removes the selected expression from the stored expressions,
+the selected expression must be a stored expression.
+
+.. versionadded:: 3.12
 %End
 
   signals:
--- a/python/gui/auto_generated/qgsexpressionbuilderwidget.sip.in
+++ b/python/gui/auto_generated/qgsexpressionbuilderwidget.sip.in
@@ -359,6 +359,18 @@ the selected expression must be a stored expression.
 .. versionadded:: 3.12
 %End
 
+    const QList<QgsExpressionItem *> findExpressions( const QString &label );
+%Docstring
+Returns the list of expression items matching a ``label``.
+
+.. note::
+
+   this function is exposed for testing purposes only
+
+.. versionadded:: 3.12
+%End
+
+
   signals:
 
     void expressionParsed( bool isValid );
--- a/python/gui/auto_generated/qgsexpressionbuilderwidget.sip.in
+++ b/python/gui/auto_generated/qgsexpressionbuilderwidget.sip.in
@@ -227,25 +227,25 @@ Loads the recent expressions from the given ``collection``.
 By default it is loaded from the collection "generic".
 %End
 
-    void loadStored( const QString &collection =  QStringLiteral( "generic" ) );
+    void loadUserExpressions( const QString &collection =  QStringLiteral( "generic" ) );
 %Docstring
-Loads the stored expressions from the given ``collection``.
+Loads the user expressions from the given ``collection``.
 By default it is loaded from the collection "generic".
 
 .. versionadded:: 3.12
 %End
 
-    void saveToStored( const QString &label, const QString expression, const QString &helpText, const QString &collection =  QStringLiteral( "generic" ) );
+    void saveToUserExpressions( const QString &label, const QString expression, const QString &helpText, const QString &collection =  QStringLiteral( "generic" ) );
 %Docstring
-Stores the ``expression`` in the given ``collection`` with given ``label`` and ``helpText``.
+Stores the user ``expression`` in the given ``collection`` with given ``label`` and ``helpText``.
 By default it is saved to the collection "generic".
 
 .. versionadded:: 3.12
 %End
 
-    void removeFromStored( const QString &name, const QString &collection =  QStringLiteral( "generic" ) );
+    void removeFromUserExpressions( const QString &name, const QString &collection =  QStringLiteral( "generic" ) );
 %Docstring
-Removes the expression ``name`` from the stored expressions in the given ``collection``.
+Removes the expression ``name`` from the user stored expressions in the given ``collection``.
 By default it is removed from the collection "generic".
 
 .. versionadded:: 3.12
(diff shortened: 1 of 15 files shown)


m-kuhn:
Nice stuff, @elpaso 

@signedav as author of https://github.com/qgis/QGIS/pull/31349, could you do a review here?

roya0045:
A very appreciated feature! Thanks @elpaso !

I've had a quick glance at the code and I couldn't quite guess where the expression were stored. I was curious to see of the expression were linked to a specific project, a project/layer combination or a specific user?

Such a feature would be even better if it could be shared across users and imported/exported to ensure that everyone in a team uses the same expressions.

elpaso:
Expressions are stored in settings, import export was planned but it is something for the future.

signedav:
Although the project scope stored expressions here #31349 are something different, they still could be used at the same places. Like e.g. here, it would make sense to have them as well maybe once. 

How how should they be identifiable for the user in future when they are both at the same places? They are both called "stored expressions". Of course we can change that in future, but maybe it makes sense to have it clear from the beginning on.

roya0045:
> Although the project scope stored expressions here #31349 are something different, they still could be used at the same places. Like e.g. here, it would make sense to have them as well maybe once.
> 
> How how should they be identifiable for the user in future when they are both at the same places? They are both called "stored expressions". Of course we can change that in future, but maybe it makes sense to have it clear from the beginning on.

Could it be possible to have an hybrid where people would use their settings to point to an expression store. This way there could be multiple storage and could be changed on a project basis? Though I assume there could be some issues with concurrent read write and using export import might be safer.

elpaso:
That's a good question and a potential source of confusion, for now I don't see a problem, because "your" stored expressions are named "Stored filter expressions" (that could be shortened to "Stored filters"). (If not yet there, you may want to add a tooltip to make it clear that are stored in the project though) .

If you plan to add a new category for the project-level expressions to the expressions tree, I think you should probably name it "Project expressions (context name)" to make it clear that they are project-level. 

You may also want to add the context (generic, selection, field-calculator, etc.) and the helptext to your implementation.

As an alternative, we could name my category "User expressions" or "Custom expressions" but I'm not really sure what's clearer, probably "User expressions" makes it clear that are stored in the user profile.

m-kuhn:
I can see the advantage for both backends.

- When it's about commonly used snippets / templates, they can be saved on the system (current implementation here)
- When it's about layer/project specific expressions (often the case as soon as field names are involved / aggregates over other layers in the project / ... ) the project backend implementation from https://github.com/qgis/QGIS/pull/31349 is more interesting

We'll probably want to have the possibility to decide on storage time if the expression should be stored on the system or in the project?

signedav:
The user needs to be able to use both at the same time. Maybe he has his locally stored "snippets" and still he receives a project from another college containing project stored expressions for filters etc.

@elpaso I like the name "user expressions" for this case here. No matter how the other ones are called.

elpaso:
@signedav I agree that there is use case for both, there is still a slight difference but the use cases mostly overlap and they are not mutually exclusive.

From the UX/UI perspective, to add a checkbox to store in the project to my dialog implementation (that will probably be used intead of yours simplified single-line name dialog) might be taken into consideration (but it's not in scope for this PR).

I'll change the category name of "my" expressions to "User expressions (context)" so that there will be place for another "Project expressions" category in the future. 

I believe you won't find difficult to use my GUI implementation to for projects expressions  management when you get to it.

elpaso:
immagine

signedav (approved):
Code looks good to me

signedav:
What do you think about the star I used in the project expressions?

When clicking the grey star the current expression is stored (store dialoge) and the star turns yellow.
Then when loading an expression there is a yellow star and on clicking it turns grey again (and deletes the expression).

But even if you like it, like I do, this could be another PR :-)

elpaso:
@signedav thank you for the review!

elpaso:
> What do you think about the star I used in the project expressions?
> 
> When clicking the grey star the current expression is stored (store dialoge) and the star turns yellow.
> Then when loading an expression there is a yellow star and on clicking it turns grey again (and deletes the expression).
> 
> But even if you like it, like I do, this could be another PR :-)

I do like it, it's like browser bookmarks! 

What I'm honestly not sure about is if it's intuitive for the majority of users, I've found out that tooltips may help, did you add any?

roya0045:
Are we forced to have the '(fieldcalc)' part to the group name? I feel like this might be more confusing than anything.

signedav:
@elpaso Yes I have there. I'm a fan of tooltips. But maybe there could be more.

SilvioGrosso:
Hello everyone,

Just tested the new options: great work indeed as usual!
The 3.12 upcoming release is going to rock for sure :-)

I may be wrong but it looks like there is a little glitch in one tooltip of an icon (the one with the dustbin): the verb Remove ends with an S which, in my view, is not totally correct and consistent with the other 2 new tooltips.
In short, *RemoveS* should be *Remove*.
I am not an Englishman, therefore feel free to ignore my comment if it is not pertinent :-)

elpaso:
> Hello everyone,
> 
> Just tested the new options: great work indeed as usual!
> The 3.12 upcoming release is going to rock for sure :-)
> 
> I may be wrong but it looks like there is a little glitch in one tooltip of an icon (the one with the dustbin): the verb Remove ends with an S which, in my view, is not totally correct and consistent with the other 2 new tooltips.
> In short, _RemoveS_ should be _Remove_.
> I am not an Englishman, therefore feel free to ignore my comment if it is not pertinent :-)

@SilvioGrosso
You are right, I'll fix it.

@roya0045 about the `(fieldcalc)` I added the context to be consistent with the "Recent expressions", I'm not sure if there is a reason for keeping it or removing, but if we decide to remove it from the user expressions (this PR implementation) I think we should remove it from the "Recent expressions" as well.

I didn't write the original  "Recent expressions" code, but I guess that the idea was to keep somewhat separate the expressions added in the field calculator context from those added in the "Selection" context and all the other expressions that are in group "generic".

@m-kuhn any opinion about keeping/removing the group for the user expressions (and for the recent)?

roya0045:
@elpaso I just feel like mentionning field calc indicate that those are only available in specific context. If those are available for every expression such as the recent expression group, (I don't recall there being mention of field calc in 3.4).
If those are available everywhere I feel like having that suffix is misleading as you are saying in your second paragraph.

I know this should be a quick fix and maybe  @DelazJ has an opinion on that.

elpaso:
> @elpaso I just feel like mentionning field calc indicate that those are only available in specific context. 

That's exactly the purpose of the context "fieldcalc", to keep expressions separated. It's the same for "Recent expressions".

m-kuhn:
I have a tendency to drop these `(fieldcalc)` parts from the labels

- As far as I can see, there is no other group right now
- These are untranslatable, technical tags/labels
- It's hard to discover what they stand for for a user
- I currently don't see a requirement to add more groups. This system is for generic snippets. Whereas the "other" system is for expressions available in a specific situation (layer/project/context)

At the moment I can't see any benefit for the `(fieldcalc)` part. This may change in the future and at this point, it can easily be readded. But as usual, it's possible I'm missing something, please prove me wrong!

elpaso:
> I have a tendency to drop these `(fieldcalc)` parts from the labels
> 
>     * As far as I can see, there is no other group right now

Don't get me wrong, I'm +0 on dropping the group, but there are other groups: "generic" and "Selection" (at least). And my implementation is not only available in field calculator but in all expression editors.

> 
>     * These are untranslatable, technical tags/labels
> 
>     * It's hard to discover what they stand for for a user

True

> 
>     * I currently don't see a requirement to add more groups. This system is for generic snippets. Whereas the "other" system is for expressions available in a specific situation (layer/project/context)
> 
> 
> At the moment I can't see any benefit for the `(fieldcalc)` part. This may change in the future and at this point, it can easily be readded. But as usual, it's possible I'm missing something, please prove me wrong!

So, you are saying to also remove the grouping for "Recent expressions" ?

If not, why?

m-kuhn:
> So, you are saying to also remove the grouping for "Recent expressions" ?

I was going for a yes but

> ... there are other groups: "generic" and "Selection" (at least).

I had missed that before. That's currently in use for the recent expressions I guess?

I wonder to what degree this partitioning is useful. I can imagine a workflow where someone will be happy to use the same expression to first test with the select by expression tool and then use it in a rule based rendering rule. Having it not available in saved expressions will be slightly disappointing. (and it can also be disappointing not having it in recent expressions).

So if you ask me again now, after having highlighted the additional sections, I'm +0 towards only disabling it in stored expression because selection of "useful" expressions is linked to a manual task whereas "recent" expressions are generated automatically and having less noise there can be handy (but I'm not totally convinced it's the right thing to do either).

elpaso:
I'm also doubtful, is there any way to collect opinions from the "power" users?


> > So, you are saying to also remove the grouping for "Recent expressions" ?
> 
> I was going for a yes but
> 
> > ... there are other groups: "generic" and "Selection" (at least).
> 
> I had missed that before. That's currently in use for the recent expressions I guess?

Exactly.

> 
> I wonder to what degree this partitioning is useful. I can imagine a workflow where someone will be happy to use the same expression to first test with the select by expression tool and then use it in a rule based rendering rule. Having it not available in saved expressions will be slightly disappointing. (and it can also be disappointing not having it in recent expressions).
> 
> So if you ask me again now, after having highlighted the additional sections, I'm +0 towards only disabling it in stored expression because selection of "useful" expressions is linked to a manual task whereas "recent" expressions are generated automatically and having less noise there can be handy (but I'm not totally convinced it's the right thing to do either).

I'm also doubtful, is there any way to collect opinions from the "power" users?

m-kuhn:
Hmmm...

- Ping them in here
- Write to the mailing list and hope for the best
- Just do what you think is best ;)

If it was me, I'd probably drop it from the new implementation until someone asks for it, in the spirit of "only add things if there is an expressed need for it". Or add it to a tooltip where you can be more verbose about what it stands for. After all you

> found out that tooltips may help, did you add any?

:stuck_out_tongue_winking_eye:

pigreco:
My experience is that the average user has no knowledge of everything.

But from now on, having the possibility of saving them, you will notice that some expressions find them and others do not.

In essence, for me every saved expression must be usable anywhere.

m2c

elpaso:
> Hmmm...
> 
>     * Ping them in here
> 
>     * Write to the mailing list and hope for the best
> 
>     * Just do what you think is best ;)
> 

I like the last one :)

nirvn:
I'm -0 on the context/category approach. I think it's a good idea in theory, but likely to create a fair amount of confusion in practice. I don't have strong enough feelings about it to make my vote a -1 though :wink:

albertograva:
I completely agree with @pigreco opinion's.

Korto19:
I'm too completely agree with @pigreco opinion's.

PR #33443, opened 2019-12-17
fix build

3nids:
on mac only?


3nids:
testing backport at the same time

PR #33451, opened 2019-12-17
Update test to reflect changes in proj library

nyalldawson:


rouault:
When PROJ 7 will be on track, this will change again (presumably one single package with all grids as GeoTIFF). So I'd suggest perhaps just checking that packageName and url are non-empty.

nyalldawson:
> presumably one single package with all grids as GeoTIFF

Won't that be a huge download?

rouault:
> Won't that be a huge download?

480 MB with the current content of https://github.com/OSGeo/proj-datumgrid-geotiff/ . People are confused by the multiple packages

rouault:
> Won't that be a huge download?

The discussion about that is in https://github.com/OSGeo/proj-datumgrid/issues/74

PR #33462, opened 2019-12-18
[feature] User expressions followup

elpaso:
Followup https://github.com/qgis/QGIS/pull/33437

- renamed the group to "User expressions"
- added help text for the group
- added QgsSettings Expressions namespace
- reworded tooltips for store/remove actions

user-expressions

### DIFF
--- a/python/core/auto_generated/qgssettings.sip.in
+++ b/python/core/auto_generated/qgssettings.sip.in
@@ -57,6 +57,7 @@ QgsSettings provides some shortcuts to get/set namespaced settings from/to a spe
       Auth,
       App,
       Providers,
+      Expressions,
       Misc
     };
 
--- a/python/gui/auto_generated/qgsexpressionbuilderwidget.sip.in
+++ b/python/gui/auto_generated/qgsexpressionbuilderwidget.sip.in
@@ -230,53 +230,50 @@ Loads the recent expressions from the given ``collection``.
 By default it is loaded from the collection "generic".
 %End
 
-    void loadUserExpressions( const QString &collection =  QStringLiteral( "generic" ) );
+    void loadUserExpressions( );
 %Docstring
-Loads the user expressions from the given ``collection``.
-By default it is loaded from the collection "generic".
+Loads the user expressions.
 
 .. versionadded:: 3.12
 %End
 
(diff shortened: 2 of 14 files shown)


SrNetoChan:
@elpaso are the custom functions created in python also stored in here? In qgis 3.10 they seem to appear in a "custom" group. I am wandering if we shouldn't keep that name only for the functions and not stored expressions.

elpaso:
What do you mean? The name here is "User" not "Custom".

SrNetoChan:
Bu the help says it stores custom functions. And then there can be a "custom" group that where the custom python functions will show.

That's why I asked. When, in that group help, you say it contains "custom functions" what do you mean with that?

elpaso:
What do you suggest? Drop "custom" to avoid confusion?

SrNetoChan:
If there's no functions in there, I wouldn't mention custom fuctions. Just "user's saved expressions"

m-kuhn:
Did you consider ["User Expressions" as discussed in the original PR](https://github.com/qgis/QGIS/pull/33437#issuecomment-566592702)?

elpaso:
Using "expressions" in an expression list seemed redundant to me.

SrNetoChan:
Actually, the list is of functions, not expressions :-p.

But the same happen with the "recent" group

elpaso:
ok, let's settle back to "User expressions"

m-kuhn:
I wouldn't mind "recent expressions", but the term *recent* has less ambiguity than the term *user*.

roya0045:
What about "Saved expressions"? It is what they are, recent implies that they are the latest, whereas saved expression is straightforward as those have been explicitly saved by the user. (Unless this is not related to the work in the other PR).

pigreco:
@roya0045 

> What about "Saved expressions"?

in the [QEP ](https://github.com/qgis/QGIS-Enhancement-Proposals/issues/135#issuecomment-558108289)it is really so

roya0045:
> @roya0045
> 
> > What about "Saved expressions"?
> 
> in the [QEP ](https://github.com/qgis/QGIS-Enhancement-Proposals/issues/135#issuecomment-558108289)it is really so

I guess we could have one named 'saved expression' and the other 'stored expression', where one would refer to expressions explicitly saved by a user whereas stored expressions would be expression stored in a project.This would ensure that people don't mix up either if we consider the other implementation discussed in the other PR.

PR #33474, opened 2019-12-19
New parameter type for coordinate operations

nyalldawson:
Allows selection of the proj coordinate operation to use when reprojecting between two CRSes

Currently not used in any algorithms, but the plan is to add it to the vector and raster reprojection algorithms (but only when running on proj 6 builds)


alexbruy (approved):
Looks good to me

nirvn:
@nyalldawson , could this also be useful to add in the save [vector/raster] layer as, since we have a CRS selector there?

PR #33479, opened 2019-12-19
BUG: scientific notation for double number between 1 and -1, fix #33475

### LINKED ISSUE #33475: Scientific notation in attribute table for double type value
**Describe the bug**

The scientific notation is not used in the attribute table for double type display. I'm aware that the use of scientific notation has already been discussed in some issues (like for labelling [here](https://github.com/qgis/QGIS/pull/32398), and [here](https://github.com/qgis/QGIS/pull/32398)).

The use of scientific notation for int or longlong type doesn't seem wished, which I agree, but at the moment it is also not used for double type. It can be annoying, especially for double number between 1 and -1, like 4.5e-09 for example that is displayed as '0.0000000' in attribute table

**How to Reproduce**

[test_decimal.zip](https://github.com/qgis/QGIS/files/3978270/test_decimal.zip)

the data test contains a gpkg file with a layer, that have a decimal field named `decimal_f`

1. Open the gpkg file in Qgis
2. Open the attribute table, `decimal_f` field displays 0,00000

image

**QGIS and OS versions**

QGIS version | 3.11.0-Master | QGIS code revision | d0348127bf
-- | -- | -- | --
OS Version | Ubuntu 19.04
Active python plugins | MetaSearch; processing; db_manager

**Additional context**

Can be fix by using 'g' mode [here](https://github.com/qgis/QGIS/blob/master/src/core/qgsfield.cpp#L255) and [here](https://github.com/qgis/QGIS/blob/master/src/core/qgsfield.cpp#L261)

image

SebastienPeillet:
## Description

Issue explained in #33475 

Double numbers in attribute table are not displayed with scientific notation which is annoying for values between 1 and -1, like 4.5e-09. After reading some issues discussion, it appears that scientific notation for high values is not necessary whished

Before :

image

After : 

image

Should be backported to 3.10 release

## Checklist

### DIFF
--- a/tests/src/core/testqgsfield.cpp
+++ b/tests/src/core/testqgsfield.cpp
@@ -412,11 +412,11 @@ void TestQgsField::displayString()
   customEnglish.setNumberOptions( QLocale::NumberOption::OmitGroupSeparator );
   QLocale::setDefault( customEnglish );
   QCOMPARE( doubleField.displayString( 5.005005 ), QString( "5.005" ) );
-  QCOMPARE( doubleField.displayString( 4.5e-09 ), QString( "4.5e-09" ) );
+  QCOMPARE( doubleField.displayString( 4.5e-09 ), QString( "4.5e-9" ) ); //for recent qt package the expected value would be "4.5e-09"
   QCOMPARE( doubleField.displayString( 1e-04 ), QString( "0.0001" ) );
   QCOMPARE( doubleFieldNoPrec.displayString( 5.005005 ), QString( "5.005005" ) );
   QCOMPARE( doubleFieldNoPrec.displayString( 5.005005005 ), QString( "5.005005005" ) );
-  QCOMPARE( doubleFieldNoPrec.displayString( 4.5e-09 ), QString( "4.5e-09" ) );
+  QCOMPARE( doubleFieldNoPrec.displayString( 4.5e-09 ), QString( "4.5e-9" ) ); //for recent qt package the expected value would be "4.5e-09"
   QCOMPARE( doubleFieldNoPrec.displayString( 1e-04 ), QString( "0.0001" ) );
   QCOMPARE( doubleFieldNoPrec.displayString( 599999898999.0 ), QString( "599999898999" ) );
   QCOMPARE( doubleFieldNoPrec.displayString( 5999.123456 ), QString( "5999.123456" ) );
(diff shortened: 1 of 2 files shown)


elpaso (approved):
LGTM, thanks!

elpaso:
@SebastienPeillet the test is failing: https://travis-ci.org/qgis/QGIS/jobs/627280613#L1295 was it passing locally on your machine?

SebastienPeillet:
My bad, I saw that the test was failing, I fixed it but I forgot to commit and push before opening the pull request...

It should be ok now (as it is on my laptop). The first time I forgot to put scientific notation in the special case where QgsField precision is mentionned ([here](https://github.com/qgis/QGIS/pull/33479/files#diff-b2de5811b366da6fc58a391b307efd38R236))

SebastienPeillet:
@elpaso It failed again even if it work through on my machine. Apparently english locale with the OmitSeparatorGroup option changes 4.5e-09 into '4.5e-9', which I didn't reproduce... I read that travis launch the build as an bionic Ubuntu version, and I use disco Ubuntu version.
So the libqt5core5a package used is not the same but I don't know if it is the cause...

elpaso:
I usually launch my local tests with `LC_ALL=en_US.UTF-8`

SebastienPeillet:
I do it also !

I did the check with a bionic Ubuntu, I obtained the same "4.5e-9" instead "4.5e-09", so I guess it's a recent change in Qt package between the bionic and the disco Ubuntu release.

I changed the value to fit with the bionic release and I let a commentary on the line for the future.

tests/src/core/testqgsfield.cpp
```diff
@@ -412,11 +412,11 @@ void TestQgsField::displayString()
   customEnglish.setNumberOptions( QLocale::NumberOption::OmitGroupSeparator );
   QLocale::setDefault( customEnglish );
   QCOMPARE( doubleField.displayString( 5.005005 ), QString( "5.005" ) );
-  QCOMPARE( doubleField.displayString( 4.5e-09 ), QString( "4.5e-9" ) ); //for recent qt package the expected value would be "4.5e-09"
+  //QCOMPARE( doubleField.displayString( 4.5e-09 ), QString( "4.5e-9" ) ); //for recent qt package the expected value would be "4.5e-09"
```
elpaso: You may also want to check for QT version at build time and act accordingly, this way the test will pass whatever QT version the developer or the CI is using.

SebastienPeillet: Yes, putting the test in comment was the lazy way to be sure it was really this one that blocked the build in azure (not very verbosed) ;)

mhugo:
Since it has been approved, I'll merge. Thanks @SebastienPeillet and thanks @elpaso for the review !

PR #33480, opened 2019-12-19
[3d] Load 3D vector layer data in background + tiling

wonder-sk:
This adds support for background loading of data from vector layers into 3D map views. Until now, loading (and tessellation) would freeze the GUI completely - this could take many seconds depending on the complexity of input data.

The basic vector layer renderer and rule-based renderer were converted to use QgsChunkedEntity which is already used for terrain rendering. There are two more improvements in addition to unlocking of GUI:
- loading process is multi-threaded instead of using just a single core
- loading is done in tiles - so it is possible to see the tiles with 3D data appearing while other data are still being loaded

There is a new configuration option in 3D tab of vector layers - it determines how deep will be the quadtree. For example, one zoom level means there will be a single tile for the whole layer. Three zoom levels means there will be 16 tiles at the leaf level (every extra zoom level multiplies that by 4, so I have limited GUI to max. 8 levels which gives ~16K tiles which is already a lot).

How a vector layer's tiling quadtree gets populated: all internal tree nodes are empty and thus the 3D map scene tries to immediately replace them with their children - this goes until leaf nodes are reached. Only nodes at the leaf level currently hold any data. This may change in the feature when we introduce more elaborate strategies - for example, internal nodes may contain a small percentage of features of the child nodes (this would allow us to show something while zoomed out a lot, not requiring to load all data).

For debugging purposes, there is also a new configuration option "show bounding boxes". This allows you to see bounding box of each tile (especially useful if there are some issues with tiles not showing up when they should).

This work has been sponsored by a QGIS.org grant.

### DIFF
--- a/src/3d/qgsrulebasedchunkloader_p.cpp
+++ b/src/3d/qgsrulebasedchunkloader_p.cpp
@@ -118,9 +118,27 @@ Qt3DCore::QEntity *QgsRuleBasedChunkLoader::createEntity( Qt3DCore::QEntity *par
     return new Qt3DCore::QEntity( parent );  // dummy entity
   }
 
+  float zMin = std::numeric_limits<float>::max();
+  float zMax = std::numeric_limits<float>::min();
+
   Qt3DCore::QEntity *entity = new Qt3DCore::QEntity( parent );
   for ( QgsFeature3DHandler *handler : mHandlers.values() )
+  {
     handler->finalize( entity, mContext );
+    if ( handler->zMinimum() < zMin )
+      zMin = handler->zMinimum();
+    if ( handler->zMaximum() > zMax )
+      zMax = handler->zMaximum();
+  }
+
+  // fix the vertical range of the node from the estimated vertical range to the true range
+  if ( zMin != std::numeric_limits<float>::max() && zMax != std::numeric_limits<float>::min() )
+  {
+    QgsAABB box = mNode->bbox();
+    box.yMin = zMin;
+    box.yMax = zMax;
+    mNode->setExactBbox( box );
+  }
 
   return entity;
 }
@@ -149,8 +167,8 @@ QgsChunkLoader *QgsRuleBasedChunkLoaderFactory::createChunkLoader( QgsChunkNode 
 
 ///////////////
 
-QgsRuleBasedChunkedEntity::QgsRuleBasedChunkedEntity( QgsVectorLayer *vl, const QgsVectorLayer3DTilingSettings &tilingSettings, QgsRuleBased3DRenderer::Rule *rootRule, const Qgs3DMapSettings &map )
-  : QgsChunkedEntity( Qgs3DUtils::layerToWorldExtent( vl->extent(), vl->crs(), map.origin(), map.crs(), map.transformContext() ),
+QgsRuleBasedChunkedEntity::QgsRuleBasedChunkedEntity( QgsVectorLayer *vl, double zMin, double zMax, const QgsVectorLayer3DTilingSettings &tilingSettings, QgsRuleBased3DRenderer::Rule *rootRule, const Qgs3DMapSettings &map )
+  : QgsChunkedEntity( Qgs3DUtils::layerToWorldExtent( vl->extent(), zMin, zMax, vl->crs(), map.origin(), map.crs(), map.transformContext() ),
                       -1,  // rootError  TODO: negative error should mean that the node does not contain anything
                       -1, // tau = max. allowed screen error. TODO: negative tau should mean that we need to go until leaves are reached
                       tilingSettings.zoomLevelsCount() - 1,
--- a/src/3d/qgsrulebasedchunkloader_p.cpp
+++ b/src/3d/qgsrulebasedchunkloader_p.cpp
@@ -17,6 +17,7 @@ QgsRuleBasedChunkLoader::QgsRuleBasedChunkLoader( const QgsRuleBasedChunkLoaderF
   : QgsChunkLoader( node )
   , mFactory( factory )
   , mContext( factory->mMap )
+  , mSource( new QgsVectorLayerFeatureSource( factory->mLayer ) )
 {
   if ( node->level() < mFactory->mLeafLevel )
   {
@@ -31,10 +32,16 @@ QgsRuleBasedChunkLoader::QgsRuleBasedChunkLoader( const QgsRuleBasedChunkLoaderF
   exprContext.setFields( layer->fields() );
   mContext.setExpressionContext( exprContext );
 
-  mFactory->mRootRule->createHandlers( layer, mHandlers );
+  // factory is shared among multiple loaders which may be run at the same time
+  // so we need a local copy of our rule tree that does not intefere with others
+  // (e.g. it happened that filter expressions with invalid syntax would cause
+  // nasty crashes when trying to simultaneously record evaluation error)
+  mRootRule.reset( mFactory->mRootRule->clone() );
+
+  mRootRule->createHandlers( layer, mHandlers );
 
   QSet<QString> attributeNames;
-  mFactory->mRootRule->prepare( mContext, attributeNames, mHandlers );
+  mRootRule->prepare( mContext, attributeNames, mHandlers );
 
   // build the feature request
   QgsFeatureRequest req;
@@ -54,20 +61,32 @@ QgsRuleBasedChunkLoader::QgsRuleBasedChunkLoader( const QgsRuleBasedChunkLoaderF
     QgsEventTracing::ScopedEvent e( "3D", "RB chunk load" );
 
     QgsFeature f;
-    QgsFeatureIterator fi = mFactory->mSource->getFeatures( req );
+    QgsFeatureIterator fi = mSource->getFeatures( req );
     while ( fi.nextFeature( f ) )
     {
       if ( mCanceled )
         break;
       mContext.expressionContext().setFeature( f );
-      mFactory->mRootRule->registerFeature( f, mContext, mHandlers );
+      mRootRule->registerFeature( f, mContext, mHandlers );
     }
   } );
 
   // emit finished() as soon as the handler is populated with features
-  QFutureWatcher<void> *fw = new QFutureWatcher<void>( nullptr );
-  fw->setFuture( future );
-  connect( fw, &QFutureWatcher<void>::finished, this, &QgsChunkQueueJob::finished );
+  mFutureWatcher = new QFutureWatcher<void>( this );
+  mFutureWatcher->setFuture( future );
+  connect( mFutureWatcher, &QFutureWatcher<void>::finished, this, &QgsChunkQueueJob::finished );
+}
+
+QgsRuleBasedChunkLoader::~QgsRuleBasedChunkLoader()
+{
+  if ( mFutureWatcher && !mFutureWatcher->isFinished() )
+  {
+    disconnect( mFutureWatcher, &QFutureWatcher<void>::finished, this, &QgsChunkQueueJob::finished );
+    mFutureWatcher->waitForFinished();
+  }
+
+  qDeleteAll( mHandlers );
+  mHandlers.clear();
 }
 
 void QgsRuleBasedChunkLoader::cancel()
@@ -86,9 +105,6 @@ Qt3DCore::QEntity *QgsRuleBasedChunkLoader::createEntity( Qt3DCore::QEntity *par
   for ( QgsFeature3DHandler *handler : mHandlers.values() )
     handler->finalize( entity, mContext );
 
-  qDeleteAll( mHandlers );
-  mHandlers.clear();
-
   return entity;
 }
 
@@ -101,7 +117,10 @@ QgsRuleBasedChunkLoaderFactory::QgsRuleBasedChunkLoaderFactory( const Qgs3DMapSe
   , mLayer( vl )
   , mRootRule( rootRule->clone() )
   , mLeafLevel( leafLevel )
-  , mSource( new QgsVectorLayerFeatureSource( vl ) )
+{
+}
+
+QgsRuleBasedChunkLoaderFactory::~QgsRuleBasedChunkLoaderFactory()
 {
 }
 
--- a/src/3d/qgsrulebasedchunkloader_p.cpp
+++ b/src/3d/qgsrulebasedchunkloader_p.cpp
@@ -1,3 +1,18 @@
+/***************************************************************************
+  qgsrulebasedchunkloader_p.cpp
+  --------------------------------------
+  Date                 : November 2019
+  Copyright            : (C) 2019 by Martin Dobias
+  Email                : wonder dot sk at gmail dot com
+ ***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
 #include "qgsrulebasedchunkloader_p.h"
 
 #include "qgs3dutils.h"
@@ -49,7 +64,7 @@ QgsRuleBasedChunkLoader::QgsRuleBasedChunkLoader( const QgsRuleBasedChunkLoaderF
   req.setSubsetOfAttributes( attributeNames, layer->fields() );
 
   // only a subset of data to be queried
-  QgsRectangle rect = Qgs3DUtils::worldToMapExtent( node->bbox(), mFactory->mLayer->crs(), map.origin(), map.crs() );
+  QgsRectangle rect = Qgs3DUtils::worldToLayerExtent( node->bbox(), mFactory->mLayer->crs(), map.origin(), map.crs(), map.transformContext() );
   req.setFilterRect( rect );
 
   //
@@ -133,7 +148,7 @@ QgsChunkLoader *QgsRuleBasedChunkLoaderFactory::createChunkLoader( QgsChunkNode 
 ///////////////
 
 QgsRuleBasedChunkedEntity::QgsRuleBasedChunkedEntity( QgsVectorLayer *vl, QgsRuleBased3DRenderer::Rule *rootRule, const Qgs3DMapSettings &map )
-  : QgsChunkedEntity( Qgs3DUtils::mapToWorldExtent( vl->extent(), vl->crs(), map.origin(), map.crs() ),
+  : QgsChunkedEntity( Qgs3DUtils::layerToWorldExtent( vl->extent(), vl->crs(), map.origin(), map.crs(), map.transformContext() ),
                       -1,  // rootError  TODO: negative error should mean that the node does not contain anything
                       -1, // tau = max. allowed screen error. TODO: negative tau should mean that we need to go until leaves are reached
                       2, // TODO: figure out from the number of features
--- a/src/3d/qgsrulebasedchunkloader_p.cpp
+++ b/src/3d/qgsrulebasedchunkloader_p.cpp
@@ -169,12 +169,12 @@ QgsChunkLoader *QgsRuleBasedChunkLoaderFactory::createChunkLoader( QgsChunkNode 
 
 QgsRuleBasedChunkedEntity::QgsRuleBasedChunkedEntity( QgsVectorLayer *vl, double zMin, double zMax, const QgsVectorLayer3DTilingSettings &tilingSettings, QgsRuleBased3DRenderer::Rule *rootRule, const Qgs3DMapSettings &map )
   : QgsChunkedEntity( Qgs3DUtils::layerToWorldExtent( vl->extent(), zMin, zMax, vl->crs(), map.origin(), map.crs(), map.transformContext() ),
-                      -1,  // rootError  TODO: negative error should mean that the node does not contain anything
-                      -1, // tau = max. allowed screen error. TODO: negative tau should mean that we need to go until leaves are reached
+                      -1, // rootError (negative error means that the node does not contain anything)
+                      -1, // max. allowed screen error (negative tau means that we need to go until leaves are reached)
                       tilingSettings.zoomLevelsCount() - 1,
                       new QgsRuleBasedChunkLoaderFactory( map, vl, rootRule, tilingSettings.zoomLevelsCount() - 1 ) )
 {
-  setShowBoundingBoxes( true );
+  setShowBoundingBoxes( tilingSettings.showBoundingBoxes() );
 }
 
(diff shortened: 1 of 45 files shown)


nirvn:
Cool!

nirvn:
@wonder-sk , just tried this PR. **Huge* improvement to the whole 3D experience, congrats.

I've stumbled on a reproducible crash with the following gdb output:

```
Thread 1 "qgis" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffe47b0e80 (LWP 30949)]
0x00007fffed33d8e4 in Qt3DCore::QNode::setEnabled(bool) () from /usr/lib/x86_64-linux-gnu/libQt53DCore.so.5
(gdb) where
#0  0x00007fffed33d8e4 in Qt3DCore::QNode::setEnabled(bool) () at /usr/lib/x86_64-linux-gnu/libQt53DCore.so.5
#1  0x00007ffff12df607 in QgsChunkedEntity::update(QgsChunkedEntity::SceneState const&) (this=0x55555bbdeb10, state=...)
    at /home/webmaster/dev/cpp/QGIS/src/3d/chunks/qgschunkedentity_p.cpp:151
#2  0x00007ffff129dec7 in Qgs3DMapScene::onFrameTriggered(float) (this=0x55555bb6c730, dt=0.0149209779) at /home/webmaster/dev/cpp/QGIS/src/3d/qgs3dmapscene.cpp:373
#3  0x00007ffff12aaa9b in QtPrivate::FunctorCall<QtPrivate::IndexesList<0>, QtPrivate::List<float>, void, void (Qgs3DMapScene::*)(float)>::call(void (Qgs3DMapScene::*)(float), Qgs3DMapScene*, void**) (f=
    (void (Qgs3DMapScene::*)(class Qgs3DMapScene * const, float)) 0x7ffff129dd2c <Qgs3DMapScene::onFrameTriggered(float)>, o=0x55555bb6c730, arg=0x7fffffffcd30)
    at /usr/include/x86_64-linux-gnu/qt5/QtCore/qobjectdefs_impl.h:152
#4  0x00007ffff12a9d60 in QtPrivate::FunctionPointer<void (Qgs3DMapScene::*)(float)>::call<QtPrivate::List<float>, void>(void (Qgs3DMapScene::*)(float), Qgs3DMapScene*, void**) (f=(void (Qgs3DMapScene::*)(class Qgs3DMapScene * const, float)) 0x7ffff129dd2c <Qgs3DMapScene::onFrameTriggered(float)>, o=0x55555bb6c730, arg=0x7fffffffcd30)
    at /usr/include/x86_64-linux-gnu/qt5/QtCore/qobjectdefs_impl.h:185
#5  0x00007ffff12a852a in QtPrivate::QSlotObject<void (Qgs3DMapScene::*)(float), QtPrivate::List<float>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) (which=1, this_=0x55555ba3fa60, r=0x55555bb6c730, a=0x7fffffffcd30, ret=0x0) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qobjectdefs_impl.h:414
#6  0x00007ffff33ef5c8 in QMetaObject::activate(QObject*, int, int, void**) () at /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#7  0x00007fffe982a3e4 in Qt3DLogic::QFrameAction::triggered(float) () at /usr/lib/x86_64-linux-gnu/libQt53DLogic.so.5
#8  0x00007fffe9829fe2 in  () at /usr/lib/x86_64-linux-gnu/libQt53DLogic.so.5
#9  0x00007fffe982a07e in  () at /usr/lib/x86_64-linux-gnu/libQt53DLogic.so.5
#10 0x00007ffff3dd7a86 in QApplicationPrivate::notify_helper(QObject*, QEvent*) () at /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#11 0x00007ffff3de0e00 in QApplication::notify(QObject*, QEvent*) () at /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#12 0x00007ffff4f8d6c1 in QgsApplication::notify(QObject*, QEvent*) (this=0x7fffffffd640, receiver=0x55555ba6bba0, event=0x7fff0800e2c0)
    at /home/webmaster/dev/cpp/QGIS/src/core/qgsapplication.cpp:417
#13 0x00007ffff33c3a9a in QCoreApplication::notifyInternal2(QObject*, QEvent*) () at /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#14 0x00007ffff33c6718 in QCoreApplicationPrivate::sendPostedEvents(QObject*, int, QThreadData*) () at /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#15 0x00007ffff341c0a7 in  () at /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#16 0x00007fffebffd84d in g_main_context_dispatch () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#17 0x00007fffebffdad0 in  () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#18 0x00007fffebffdb73 in g_main_context_iteration () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#19 0x00007ffff341b6a5 in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) () at /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#20 0x00007ffff33c263b in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) () at /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#21 0x00007ffff33ca3a6 in QCoreApplication::exec() () at /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#22 0x000055555556dbc3 in main(int, char**) (argc=1, argv=0x7fffffffdde8) at /home/webmaster/dev/cpp/QGIS/src/app/main.cpp:1597
```

Steps to reproduce the crash:
- Load this dataset into a new project: [buildings3d_fixed.zip](https://github.com/qgis/QGIS/files/4052618/buildings3d_fixed.zip)
- Open the style dock, go to the 3D symbology tab
- Set the zoom level to 7, activate single 3D symbol, and set the polygon's extrusion to the HEIGHT field
- Open a new 3D map canvas panel
- *boom* QGIS dies

wonder-sk:
@nirvn Thanks a lot for testing! Hmm that crash is caused by the fact that we currently allow up to 512 active tiles for a single entity. But here with zoom level 7 we get to a situation where the engine wants to load 4096 tiles (if zoom to the whole layer). Right now I think of just fixing it by changing the max. zoom level to 5 in order to keep max. number of tiles to 256. And later I think we will need to handle this in a more elegant way...

nirvn:
@wonder-sk , one more issue, UI/UX related:
Screenshot from 2020-01-14 14-23-40

IMHO, we should _not_ show the zoom level counts as well as bounding box (the latter being debugging-related) by default, and all the time. It's not user friendly as it can be overwhelming for new users, and it eats vertical spacing.

My suggestion:  use a collapsible group box with a title along the lines of '3D Rendering'. That'll adopt something we also have for 2D symbology:
Screenshot from 2020-01-14 14-23-49

nirvn:
@wonder-sk , thanks for embracing the collapsible group box suggestion :wink: just tested the branch again, working great.

src/3d/qgsrulebasedchunkloader_p.cpp
```diff
@@ -0,0 +1,186 @@
...
+#include <QtConcurrent>
+
+///@cond PRIVATE
+
+
+QgsRuleBasedChunkLoader::QgsRuleBasedChunkLoader( const QgsRuleBasedChunkLoaderFactory *factory, QgsChunkNode *node )
+  : QgsChunkLoader( node )
+  , mFactory( factory )
+  , mContext( factory->mMap )
+  , mSource( new QgsVectorLayerFeatureSource( factory->mLayer ) )
```
PeterPetrik: is there any possibility factory == nulptr when calling this constructor?

src/3d/qgsrulebasedchunkloader_p.cpp
```diff
@@ -0,0 +1,186 @@
...
+  QgsRectangle rect = Qgs3DUtils::worldToLayerExtent( node->bbox(), mFactory->mLayer->crs(), map.origin(), map.crs(), map.transformContext() );
+  req.setFilterRect( rect );
+
+  //
+  // this will be run in a background thread
+  //
+
+  QFuture<void> future = QtConcurrent::run( [req, this]
+  {
+    QgsEventTracing::ScopedEvent e( "3D", "RB chunk load" );
```
PeterPetrik: QStringLiteral

src/3d/qgsrulebasedchunkloader_p.cpp
```diff
@@ -0,0 +1,186 @@
...
+}
+
+
+///////////////
+
+
+QgsRuleBasedChunkLoaderFactory::QgsRuleBasedChunkLoaderFactory( const Qgs3DMapSettings &map, QgsVectorLayer *vl, QgsRuleBased3DRenderer::Rule *rootRule, int leafLevel )
+  : mMap( map )
+  , mLayer( vl )
+  , mRootRule( rootRule->clone() )
```
PeterPetrik: can rootRule be nullptr?

src/3d/qgsrulebasedchunkloader_p.cpp
```diff
@@ -0,0 +1,186 @@
...
+  , mLeafLevel( leafLevel )
+{
+}
+
+QgsRuleBasedChunkLoaderFactory::~QgsRuleBasedChunkLoaderFactory()
```
PeterPetrik: =default?

src/3d/qgsrulebasedchunkloader_p.h
```diff
@@ -0,0 +1,119 @@
...
+#include "qgsfeature3dhandler_p.h"
+#include "qgschunkedentity_p.h"
+
+class Qgs3DMapSettings;
+class QgsVectorLayer;
+class QgsVectorLayerFeatureSource;
+class QgsAbstract3DSymbol;
+class QgsFeature3DHandler;
+
+#include "qgsrulebased3drenderer.h"
```
PeterPetrik: can you move to the group of other includes (few lines up)

src/3d/qgsrulebasedchunkloader_p.h
```diff
@@ -0,0 +1,119 @@
...
+{
+  public:
+    //! Constructs the factory
+    QgsRuleBasedChunkLoaderFactory( const Qgs3DMapSettings &map, QgsVectorLayer *vl, QgsRuleBased3DRenderer::Rule *rootRule, int leafLevel );
+    ~QgsRuleBasedChunkLoaderFactory() override;
+
+    //! Creates loader for the given chunk node. Ownership of the returned is passed to the caller.
+    virtual QgsChunkLoader *createChunkLoader( QgsChunkNode *node ) const override;
+
+    const Qgs3DMapSettings &mMap;
```
PeterPetrik: should be these member variables private?

src/3d/qgsrulebasedchunkloader_p.h
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * \ingroup 3d
+ * 3D entity used for rendering of vector layers using a hierarchy of rules (just like
+ * in case of 2D rule-based rendering or labeling).
+ *
+ * It is implemented using tiling approach with QgsChunkedEntity. Internally it uses
+ * QgsRuleBasedChunkLoaderFactory and QgsRuleBasedChunkLoader to do the actual work
+ * of loading and creating 3D sub-entities for each tile.
+ *
+ * \since QGIS 3.12
```
PeterPetrik: not in public API,  mayber remove \ingroup and \since from these classes?

src/3d/qgsvectorlayerchunkloader_p.cpp
```diff
@@ -0,0 +1,172 @@
...
+  const Qgs3DMapSettings &map = mFactory->mMap;
+
+  if ( mFactory->mSymbol->type() == QLatin1String( "polygon" ) )
+    mHandler = Qgs3DSymbolImpl::handlerForPolygon3DSymbol( layer, *static_cast<QgsPolygon3DSymbol *>( mFactory->mSymbol.get() ) );
+  else if ( mFactory->mSymbol->type() == QLatin1String( "point" ) )
+    mHandler = Qgs3DSymbolImpl::handlerForPoint3DSymbol( layer, *static_cast<QgsPoint3DSymbol *>( mFactory->mSymbol.get() ) );
+  else if ( mFactory->mSymbol->type() == QLatin1String( "line" ) )
+    mHandler = Qgs3DSymbolImpl::handlerForLine3DSymbol( layer, *static_cast<QgsLine3DSymbol *>( mFactory->mSymbol.get() ) );
+  else
+    return;  // TODO: how to handle error
```
PeterPetrik: TODO

src/3d/qgsvectorlayerchunkloader_p.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+  QgsExpressionContext exprContext( Qgs3DUtils::globalProjectLayerExpressionContext( layer ) );
+  exprContext.setFields( layer->fields() );
+  mContext.setExpressionContext( exprContext );
+
+  QSet<QString> attributeNames;
+  if ( !mHandler->prepare( mContext, attributeNames ) )
+    return;  // TODO: how to handle errors
```
PeterPetrik: TODO

src/3d/qgsvectorlayerchunkloader_p.cpp
```diff
@@ -0,0 +1,172 @@
...
+  QgsRectangle rect = Qgs3DUtils::worldToLayerExtent( node->bbox(), mFactory->mLayer->crs(), map.origin(), map.crs(), map.transformContext() );
+  req.setFilterRect( rect );
+
+  //
+  // this will be run in a background thread
+  //
+
+  QFuture<void> future = QtConcurrent::run( [req, this]
+  {
+    QgsEventTracing::ScopedEvent e( "3D", "VL chunk load" );
```
PeterPetrik: QStringLiterar

src/3d/qgsvectorlayerchunkloader_p.h
```diff
@@ -0,0 +1,117 @@
...
+class QgsVectorLayerChunkLoaderFactory : public QgsChunkLoaderFactory
+{
+  public:
+    //! Constructs the factory
+    QgsVectorLayerChunkLoaderFactory( const Qgs3DMapSettings &map, QgsVectorLayer *vl, QgsAbstract3DSymbol *symbol, int leafLevel );
+
+    //! Creates loader for the given chunk node. Ownership of the returned is passed to the caller.
+    virtual QgsChunkLoader *createChunkLoader( QgsChunkNode *node ) const override;
+
+    const Qgs3DMapSettings &mMap;
```
PeterPetrik: should be private:?

src/3d/qgsvectorlayerchunkloader_p.h
```diff
@@ -0,0 +1,117 @@
...
+
+#include <QFutureWatcher>
+
+
+/**
+ * \ingroup 3d
+ * This loader factory is responsible for creation of loaders for individual tiles
+ * of QgsVectorLayerChunkedEntity whenever a new tile is requested by the entity.
+ *
+ * \since QGIS 3.12
```
PeterPetrik: maybe remove \ingroup and \since in private API?

src/3d/qgsvectorlayerchunkloader_p.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+QgsVectorLayerChunkLoader::~QgsVectorLayerChunkLoader()
+{
+  if ( mFutureWatcher && !mFutureWatcher->isFinished() )
+  {
+    disconnect( mFutureWatcher, &QFutureWatcher<void>::finished, this, &QgsChunkQueueJob::finished );
+    mFutureWatcher->waitForFinished();
+  }
+
+  delete mHandler;
```
PeterPetrik: maybe smart pointer?

src/3d/symbols/qgsline3dsymbol_p.cpp
```diff
@@ -67,10 +68,13 @@ class QgsBufferedLine3DSymbolHandler : public QgsFeature3DHandler
     //! temporary data we will pass to the tessellator
     struct LineData
     {
-      QList<QgsPolygon *> polygons;
-      QList<QgsFeatureId> fids;
+      std::unique_ptr<QgsTessellator> tessellator;
```
PeterPetrik: `#include <memory>`

src/3d/symbols/qgsline3dsymbol_p.cpp
```diff
@@ -131,26 +136,38 @@ void QgsBufferedLine3DSymbolHandler::processFeature( QgsFeature &f, const Qgs3DR
...
+void QgsBufferedLine3DSymbolHandler::processPolygon( QgsPolygon *polyBuffered, QgsFeatureId fid, float height, float extrusionHeight, const Qgs3DRenderContext &context, LineData &out )
+{
+  Qgs3DUtils::clampAltitudes( polyBuffered, mSymbol.altitudeClamping(), mSymbol.altitudeBinding(), height, context.map() );
+
+  Q_ASSERT( out.tessellator->dataVerticesCount() % 3 == 0 );
+  uint startingTriangleIndex = static_cast<uint>( out.tessellator->dataVerticesCount() / 3 );
+  out.triangleIndexStartingIndices.append( startingTriangleIndex );
+  out.triangleIndexFids.append( fid );
+  out.tessellator->addPolygon( *polyBuffered, extrusionHeight );
+  delete polyBuffered;
```
PeterPetrik: hmm, i think the polyBuffered should be deleted in the function created it... or ideally as unique_ptr

src/3d/symbols/qgsline3dsymbol_p.cpp
```diff
@@ -131,26 +136,38 @@ void QgsBufferedLine3DSymbolHandler::processFeature( QgsFeature &f, const Qgs3DR
...
+
+  mZMin = std::min( outNormal.tessellator->zMinimum(), outSelected.tessellator->zMinimum() );
+  mZMax = std::max( outNormal.tessellator->zMaximum(), outSelected.tessellator->zMaximum() );
 }
 
 
 void QgsBufferedLine3DSymbolHandler::makeEntity( Qt3DCore::QEntity *parent, const Qgs3DRenderContext &context, LineData &out, bool selected )
 {
-  if ( out.polygons.isEmpty() )
+  if ( !out.tessellator->dataVerticesCount() )
```
PeterPetrik: ! on integer ...

src/3d/symbols/qgspolygon3dsymbol_p.cpp
```diff
@@ -36,6 +37,8 @@
 
 /// @cond PRIVATE
 
+class QgsTessellator;
```
PeterPetrik: both include and forward declaration, remove one

src/3d/symbols/qgspolygon3dsymbol_p.cpp
```diff
@@ -52,12 +55,12 @@ class QgsPolygon3DSymbolHandler : public QgsFeature3DHandler
     //! temporary data we will pass to the tessellator
     struct PolygonData
     {
-      QList<QgsPolygon *> polygons;
-      QList<QgsFeatureId> fids;
-      QList<float> extrusionHeightPerPolygon;  // will stay empty if not needed per polygon
+      std::unique_ptr<QgsTessellator> tessellator;
```
PeterPetrik: `#include <memory>`

src/3d/symbols/qgspolygon3dsymbol_p.cpp
```diff
@@ -109,10 +115,13 @@ void QgsPolygon3DSymbolHandler::processPolygon( QgsPolygon *polyClone, QgsFeatur
...
-  out.fids.append( fid );
-  if ( hasDDExtrusion )
-    out.extrusionHeightPerPolygon.append( extrusionHeight );
+
+  Q_ASSERT( out.tessellator->dataVerticesCount() % 3 == 0 );
+  uint startingTriangleIndex = static_cast<uint>( out.tessellator->dataVerticesCount() / 3 );
+  out.triangleIndexStartingIndices.append( startingTriangleIndex );
+  out.triangleIndexFids.append( fid );
+  out.tessellator->addPolygon( *polyClone, extrusionHeight );
+  delete polyClone;
```
PeterPetrik: would be better to delete in caller

src/3d/symbols/qgspolygon3dsymbol_p.cpp
```diff
@@ -206,7 +218,7 @@ void QgsPolygon3DSymbolHandler::finalize( Qt3DCore::QEntity *parent, const Qgs3D
 
 void QgsPolygon3DSymbolHandler::makeEntity( Qt3DCore::QEntity *parent, const Qgs3DRenderContext &context, PolygonData &out, bool selected )
 {
-  if ( out.polygons.isEmpty() )
+  if ( !out.tessellator->dataVerticesCount() )
```
PeterPetrik: ! on int

PeterPetrik:
1. maybe unrelated to this PR: 

causing runtime warnings `Warning: QObject::connect(Qgs3DMapScene, Qgs3DMapToolIdentify): invalid null parameter`

PeterPetrik:
2. unable to test since it coredumps when opening any 3d view on MacOS 10.15 somewhere in [NSOpenGLContext setView]

Screenshot 2020-01-16 at 12 46 25

src/3d/qgsrulebasedchunkloader_p.h
wonder-sk: They are used directly from QgsRuleBasedChunkLoader so I left them public. This class is implementation detail so it's not in public api anyway...

src/3d/qgsvectorlayerchunkloader_p.h
wonder-sk: I don't have a strong opinion on this and there is no official policy about these - I would keep it as is...

src/3d/symbols/qgsline3dsymbol_p.cpp
wonder-sk: eventually yes, but for the time being I would leave it not to break something...

PeterPetrik (approved):

wonder-sk:
Something strange going on with azure pipelines - the jobs have been queued for several hours now - and when trying to look at the details of runs, I am just getting forever "Loading..." page :confused: 

I am tempted to merge without waiting for azure pipeline to do its job....

PR #33481, opened 2019-12-19
[wip] Partial revert of 9da842d4 for PG raster deadlock

### LINKED ISSUE #30392: PostGIS rasters are not loaded when opening a project
PostGIS rasters can be added using DB Manager on a new/clean project (they cannot be loaded from the browser as they do NOT show, I filed a separate ticket for it).

If you save a project with PostGIS rasters and then close/reopen it then this rasters are not loaded.

In the log:

2019-06-25T14:55:36     WARNING    RasterIO error: PostGISRasterDataset::LoadSources(): SSL error: wrong version number

Oddly enough if then go to DB Manager (while in the previously saved project) and try to add them you get an error "no connection to the server", which is odd because to double click on them (to add them) QGIS has just established the connection to the server...

If you try on a clean/new project you can add them.

QGIS master on Ubuntu.

elpaso:
**DO NOT MERGE**: this is only for testing.

This is an attempt to fix #30392, but the
original commit we are reverting was meant to fix
another issue #28567. We need some more testing
before we can say this is safe.

### DIFF
--- a/src/core/providers/gdal/qgsgdalprovider.cpp
+++ b/src/core/providers/gdal/qgsgdalprovider.cpp
@@ -203,10 +203,7 @@ QgsGdalProvider::QgsGdalProvider( const QgsGdalProvider &other )
 
   // The JP2OPENJPEG driver might consume too much memory on large datasets
   // so make sure to really use a single one.
-  // The PostGISRaster driver internally uses a per-thread connection cache.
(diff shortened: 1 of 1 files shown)


gioman:
@elpaso is this PR useless after by your recent work on the native postgis raster provider?

PR #33485, opened 2019-12-20
Allow dropping a map layer from the layer tree onto a projection selection widget 

nyalldawson:
...to set the projection to match that layer. Just a little timesaving shortcut!

Peek 2019-12-20 11-27


NathanW2:
Neat!

PR #33496, opened 2019-12-21
Rework labeling maximum candidate logic

nyalldawson:
First up, the easy one: This PR drops the setting for maximum number of point label candidates. This setting is easily replaced by automatically calculated, optimised values for each separate point placement mode individually. So that's an easy win.

Then the exciting bit: a rework of the label engine "maximum line candidates" and "maximum polygon candidates" settings and logic

The previous approach of a single fixed value which applied to ALL line and ALL polygon
features was... not ideal. It meant that all line features would be assigned the same
number of candidates, regardless of length. So a road of length 1 cm on the rendered
map would have an identical number of candidates as a 30cm road covering the length of the
whole map!! This resulted in both a lot of wasted calculations (generating a ridiculous
number of candidates for small lines at barely discernable distances from each other)
AND an insufficient number of candidates for lengthy features (resulting in worse label
placement for these features).

The situation was similar, but even worse for polygons. Checkout the screenshot below, showing how a tiny polygon is flooded with an excessive number candidates yet the larger polygon has comparatively poor coverage by its candidates:

image

Now, the setting is reworked to "Number of line candidates per cm" and "number of
polygon candidates per cm2". This means that small features get much less candidates,
and large features get much more features! Both a win for map rendering speed in many
circumstances AND good cartography... now that's a nice Christmas gift for QGIS :)

The new candidate placement is shown below. Note the reasonable candidate count for the small polygon, and better coverage of the larger polygon (a follow up PR will bring further dramatic improvements to this candidate coverage!)

image


pblottiere:
Hi Nyall,

> This PR drops the setting for maximum number of point label candidates. This setting is easily replaced by automatically calculated, optimised values for each separate point placement mode individually. So that's an easy win.

This parameter is pretty useful in case of QGIS Server. Indeed, when we have very strong production constraints in terms of rendering time (typically for the WMS service), we can manually minimize the number of candidates (it may really be significant in some cases). 

The GUI configuration is not necessary useful tough, but it would be nice to keep a way to adjust this setting through the `.ini` file. 

What do you think?

pblottiere:
And thanks, it's a good idea to update the default behavior which is clearly suboptimal :+1:

nyalldawson:
@pblottiere this was only ever a project level setting. What would you feel about allowing hard coded limits manually set through qgssettings which apply across ALL projects (defaulting to unset)?

nyalldawson:
(I'd also be very curious to hear if the **point** label candidate setting actually makes a noticeable difference in rendering times on current master. I have doubts about that - there's lots of extra optimisations in play now!)

pblottiere:
> this was only ever a project level setting.

I think we can configure the number of candidates through QGIS Desktop too:

Untitled

> What would you feel about allowing hard coded limits manually set through qgssettings which apply across ALL projects (defaulting to unset)?

Mmm, I think it would be the minimum necessary yes :).

> I'd also be very curious to hear if the point label candidate setting actually makes a noticeable difference in rendering times on current master

I have to admit that I don't know for point label candidates.

nyalldawson:
@pblottiere that particular setting is per project, it's saved inside the project and is reset whenever you make/open a new project (that's why I'm thinking a global setting key might be more appropriate for server?)

pblottiere:
> that particular setting is per project, it's saved inside the project and is reset whenever you make/open a new project 

Ah yes, sorry, I had missed your point (:sleepy:).

> that's why I'm thinking a global setting key might be more appropriate for server?

+1 for a global setting!

Thanks Nyall :+1:

nirvn (approved):
Looking good.

nyalldawson:
@pblottiere how's 5f48141da6e0c7f8a42757b2f97db6c41d7c1e73 look to you?

pblottiere:
> how's 5f48141 look to you?

It looks good to me, thanks Nyall :+1:

nyalldawson:
thanks @pblottiere !

PR #33523, opened 2019-12-26
Simplify and optimise pal label candidate costing code, fix some bugs

nyalldawson:
Fixes a logic bug which has existed in the pal polygon costing code since ... forever, and makes this code incrementally more readable and understandable.


nirvn (approved):
Looking good. The code readability work you've done in the past few weeks (months?) is really good.

PR #33525, opened 2019-12-26
[feature][processing] Add Line density algorithm

root676:
## Description
This PR proposes a _Line density algorithm_ as it is implemented in [ESRI ArcGIS Spatial Analyst](https://desktop.arcgis.com/en/arcmap/latest/tools/spatial-analyst-toolbox/how-line-density-works.htm). The algorithm basically moves a circular search geometry from cell to cell and calculates the line density measure as proposed by Silverman (1986). The measure is a weighted sum of intersecting line lengths that are intersecting with the search geometry, which is then divided by the search geometry area. The figure below gives a rough example of the algorithms functionality.

LineDensity_alg

The algorithm output for the road network of Vienna ([Data Source: Stadt Wien - data.wien.gv.at](https://www.data.gv.at/katalog/dataset/1039ed7e-97fb-435f-b6cc-f6a105ba5e09)) with randomly distributed weights between 1 and 3 looks like this:

LineDensityExample

Please feel free to review/comment/suggest improvements!

Literature:

- Silverman, B. W. Density Estimation for Statistics and Data Analysis. New York: Chapman and Hall, 1986.

## Checklist


src/analysis/processing/qgsalgorithmlinedensity.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+QString QgsLineDensityAlgorithm::groupId() const
+{
+  return QStringLiteral( "interpolation" );
+}
+
+void QgsLineDensityAlgorithm::initAlgorithm( const QVariantMap & )
+{
+  addParameter( new QgsProcessingParameterFeatureSource( QStringLiteral( "INPUT" ), QObject::tr( "Input line layer" ), QList<int>() << QgsProcessing::TypeVectorLine ) );
+  addParameter( new QgsProcessingParameterField( QStringLiteral( "WEIGHT" ), QObject::tr( "Weight field " ), QVariant(), QStringLiteral( "INPUT" ), QgsProcessingParameterField::Numeric ) );
```
nyalldawson: Should this be optional? (i.e. if not specified, all lines have weight of 1)

src/analysis/processing/qgsalgorithmlinedensity.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+bool QgsLineDensityAlgorithm::prepareAlgorithm( const QVariantMap &parameters, QgsProcessingContext &context, QgsProcessingFeedback *feedback )
+{
+  Q_UNUSED( feedback );
+  mSource.reset( parameterAsSource( parameters, QStringLiteral( "INPUT" ), context ) );
+  if ( !mSource )
+    throw QgsProcessingException( invalidSourceError( parameters, QStringLiteral( "INPUT" ) ) );
+  mWeight = parameterAsString( parameters, QStringLiteral( "WEIGHT" ), context );
+  mSearchRadius = parameterAsDouble( parameters, QStringLiteral( "RADIUS" ), context );
+  mPixelSize = parameterAsDouble( parameters, QStringLiteral( "PIXEL_SIZE" ), context );
```
nyalldawson: I wondering if an exception or warning should be raised if the pixel size is large compared to the search radius (i.e. if the search radius is completely enclosed by a pixel) -- this would mean some features could potentially fall into gaps between the search circles

src/analysis/processing/qgsalgorithmlinedensity.cpp
```diff
@@ -0,0 +1,193 @@
...
+  if ( context.project() )
+  {
+    mDa.setEllipsoid( context.project()->ellipsoid() );
+  }
+  mDa.setSourceCrs( mCrs, context.transformContext() );
+
+
+  //get cell midpoint from top left cell
+  QgsGeometry firstCellMidpoint = QgsGeometry( new QgsPoint( mExtent.xMinimum() + ( mPixelSize / 2 ), mExtent.yMaximum() - ( mPixelSize / 2 ) ) );
+  mSearchGeometry = firstCellMidpoint.buffer( mSearchRadius, 5 );
```
nyalldawson: Instead of a buffer to create a circle, use QgsCircle api instead.

src/analysis/processing/qgsalgorithmlinedensity.cpp
```diff
@@ -0,0 +1,193 @@
...
+  mSearchGeometryArea = mDa.measureArea( mSearchGeometry );
+
+  mIndex = QgsSpatialIndex( *mSource, nullptr, QgsSpatialIndex::FlagStoreFeatureGeometries );
```
nyalldawson: It's not a good idea to create and populate the index here (during the prepare step) -- it's potentially expensive, and will block the main thread.

src/analysis/processing/qgsalgorithmlinedensity.cpp
```diff
@@ -0,0 +1,193 @@
...
+  const QString outputFormat = QgsRasterFileWriter::driverForExtension( fi.suffix() );
+
+  int rows = std::max( std::ceil( mExtent.height() / mPixelSize ), 1.0 );
+  int cols = std::max( std::ceil( mExtent.width() / mPixelSize ), 1.0 );
+
+  //build new raster extent based on number of columns and cellsize
+  //this prevents output cellsize being calculated too small
+  QgsRectangle rasterExtent = QgsRectangle( mExtent.xMinimum(), mExtent.yMaximum() - ( rows * mPixelSize ), mExtent.xMinimum() + ( cols * mPixelSize ), mExtent.yMaximum() );
+
+  std::unique_ptr< QgsRasterFileWriter > writer = qgis::make_unique< QgsRasterFileWriter >( outputFile );
```
nyalldawson: No need to create this on the heap

src/analysis/processing/qgsalgorithmlinedensity.cpp
```diff
@@ -0,0 +1,193 @@
...
+  writer->setOutputFormat( outputFormat );
+  std::unique_ptr<QgsRasterDataProvider > provider( writer->createOneBandRaster( Qgis::Float32, cols, rows, rasterExtent, mCrs ) );
+  if ( !provider )
+    throw QgsProcessingException( QObject::tr( "Could not create raster output: %1" ).arg( outputFile ) );
+  if ( !provider->isValid() )
+    throw QgsProcessingException( QObject::tr( "Could not create raster output %1: %2" ).arg( outputFile, provider->error().message( QgsErrorMessage::Text ) ) );
+
+  provider->setNoDataValue( 1, -9999 );
+
+  int totalCellcnt = rows * cols;
```
nyalldawson: better go with qgssize here instead, or there's a risk of overflows:
```suggestion
  qgssize totalCellcnt = static_cast< qgssize >( rows ) * cols;
```

src/analysis/processing/qgsalgorithmlinedensity.cpp
```diff
@@ -0,0 +1,193 @@
...
+      {
+        break;
+      }
+
+      if ( col > 0 )
+        mSearchGeometry.translate( mPixelSize, 0 );
+
+      QList<QgsFeatureId> fids = mIndex.intersects( mSearchGeometry.boundingBox() );
+
+      std::unique_ptr< QgsGeometryEngine > engine( QgsGeometry::createGeometryEngine( mSearchGeometry.constGet() ) );
```
nyalldawson: I'd check first that fids isn't empty before doing any of the engine creation/preparation

src/analysis/processing/qgsalgorithmlinedensity.cpp
```diff
@@ -0,0 +1,193 @@
...
+      engine->prepareGeometry();
+
+      //remove non-intersecting fids
+      int i = 0;
+      while ( i < fids.size() )
```
nyalldawson:
```suggestion
      for ( QgsFeatureId id : ids )
```

src/analysis/processing/qgsalgorithmlinedensity.cpp
```diff
@@ -0,0 +1,193 @@
...
+    {
+      if ( feedback->isCanceled() )
+      {
+        break;
+      }
+
+      if ( col > 0 )
+        mSearchGeometry.translate( mPixelSize, 0 );
+
+      QList<QgsFeatureId> fids = mIndex.intersects( mSearchGeometry.boundingBox() );
```
nyalldawson:
```suggestion
      const QList<QgsFeatureId> fids = mIndex.intersects( mSearchGeometry.boundingBox() );
```

src/analysis/processing/qgsalgorithmlinedensity.cpp
```diff
@@ -0,0 +1,193 @@
...
+      engine->prepareGeometry();
+
+      //remove non-intersecting fids
+      int i = 0;
+      while ( i < fids.size() )
+      {
+        QgsFeatureId fid = fids.at( i );
+        QgsGeometry lineGeom = mIndex.geometry( fid );
+
+        if ( engine->intersects( lineGeom.constGet() ) == false )
```
nyalldawson:
```suggestion
        if ( !engine->intersects( lineGeom.constGet() ) )
```

src/analysis/processing/qgsalgorithmlinedensity.cpp
```diff
@@ -0,0 +1,193 @@
...
+        {
+          fids.removeAt( i );
+        }
+        i++;
+      }
+
+      QgsFeatureIds isectingFids = QSet< QgsFeatureId >::fromList( fids );
```
nyalldawson: I suggest building up this set in the loop above, rather then trimming the fids list you'd add intersecting fids to isectingFids (also, can you rename that to avoid the abbreviation? Abbreviations are confusing for non-native English speakers)

src/analysis/processing/qgsalgorithmlinedensity.cpp
```diff
@@ -0,0 +1,193 @@
...
+      QgsFeatureIterator fit = mSource->getFeatures( QgsFeatureRequest().setFilterFids( isectingFids ) );
+      QgsFeature f;
+      double abs_density = 0;
```
nyalldawson:
```suggestion
      double absDensity = 0;
```

src/analysis/processing/qgsalgorithmlinedensity.cpp
```diff
@@ -0,0 +1,193 @@
...
+      while ( fit.nextFeature( f ) )
+      {
+        double analysisLineLength =  mDa.measureLength( QgsGeometry( engine->intersection( f.geometry().constGet() ) ) );
+        double analysisWeight = f.attribute( mWeight ).toDouble();
+        abs_density += ( analysisLineLength * analysisWeight );
+      }
+
+      double lineDensity = abs_density / mSearchGeometryArea;
```
nyalldawson: To be exact -- you should be calculating the search geometry area here, and not using the earlier version. You've used an ellipsoid based calculation for the line length, and potentially the area of the buffer circle will change based on the lat/long of the actual pixel being calculated.

python/plugins/processing/tests/testdata/qgis_algorithm_tests2.yaml
```diff
@@ -982,6 +982,20 @@ tests:
...
+    params:
+      INPUT:
+        name: linedensity.gml|layername=linedensity
+        type: vector
+      PIXEL_SIZE: 1.0
+      RADIUS: 2.0
+      WEIGHT: weight
+    results:
+      OUTPUT:
+        hash: 37b153a98004a6656a27a8c5dcdb8dbefcd4af4938a512ec0cf5cd43
```
nyalldawson: These raster hashes are only good for checking whether the alg runs without errors -- they aren't any use for testing that no-one has broken the expected value.

I'd suggest adding specific checks to TestQgsProcessingAlgs instead, so that the actual output values are checked one by one.

src/analysis/processing/qgsalgorithmlinedensity.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+        if ( engine->intersects( lineGeom.constGet() ) == false )
+        {
+          fids.removeAt( i );
+        }
+        i++;
+      }
+
+      QgsFeatureIds isectingFids = QSet< QgsFeatureId >::fromList( fids );
+      QgsFeatureIterator fit = mSource->getFeatures( QgsFeatureRequest().setFilterFids( isectingFids ) );
```
nyalldawson: This is potentially quite costly (firing off a feature request for every pixel). I'd suggest you could avoid this, by building up a hash of the weights vs feature id at the same time you populate the spatial index. That means only a single full iteration of the layer is required, instead of potentially 1000's.

nyalldawson (commented):
Looking good, and a very valuable algorithm to have available! Thanks for your work here!

src/analysis/processing/qgsalgorithmlinedensity.cpp
root676: Yes, I had the same thoughts here... I think I will add an exception that the radius must be at least the square root of `mPixelSize` (=minimum bounding circle of cell).

root676:
Hi @nyalldawson, I just reworked the algorithm according to your review - thank you very much for the review and input. Could you take a look at the current test implementation - I am not sure if my implementation is sufficient here.

src/analysis/processing/qgsalgorithmlinedensity.cpp
```diff
@@ -0,0 +1,216 @@
...
+  QgsPoint firstCellMidpoint = QgsPoint( mExtent.xMinimum() + ( mPixelSize / 2 ), mExtent.yMaximum() - ( mPixelSize / 2 ) );
+  QgsCircle searchCircle = QgsCircle( firstCellMidpoint, mSearchRadius );
+  mSearchGeometry = QgsGeometry( searchCircle.toPolygon() );
+
+  return true;
+}
+
+QVariantMap QgsLineDensityAlgorithm::processAlgorithm( const QVariantMap &parameters, QgsProcessingContext &context, QgsProcessingFeedback *feedback )
+{
+  mFeatureWeights = QHash<QgsFeatureId, double>();
```
nyalldawson: Not required

src/analysis/processing/qgsalgorithmlinedensity.cpp
```diff
@@ -0,0 +1,216 @@
...
+  mIndex = QgsSpatialIndex( QgsSpatialIndex::FlagStoreFeatureGeometries );
+
+  QgsFeatureIterator fit = mSource->getFeatures();
```
nyalldawson: We could optimise this request by only requesting the weight attribute (using .setSubsetOfAttributes)

src/analysis/processing/qgsalgorithmlinedensity.cpp
```diff
@@ -0,0 +1,216 @@
...
+    throw QgsProcessingException( QObject::tr( "Could not create raster output %1: %2" ).arg( outputFile, provider->error().message( QgsErrorMessage::Text ) ) );
+
+  provider->setNoDataValue( 1, -9999 );
+
+  qgssize totalCellcnt = static_cast<qgssize>( rows ) * cols;
+  int cellcnt = 0;
+
+  for ( int row = 0; row < rows; row++ )
+  {
+    std::unique_ptr< QgsRasterBlock > rasterDataLine = qgis::make_unique< QgsRasterBlock >( Qgis::Float32, cols, 1 );
```
nyalldawson: You could optimise things further here and move rasterDataLine up out of the row loop, so that it's only created once and reused for every row. That'll avoid allocated memory after every row.

src/analysis/processing/qgsalgorithmlinedensity.cpp
```diff
@@ -0,0 +1,216 @@
...
+  QgsFeatureIterator fit = mSource->getFeatures();
+  QgsFeature f;
+
+  while ( fit.nextFeature( f ) )
+  {
+    mIndex.addFeature( f, QgsFeatureSink::FastInsert );
+
+    //default weight of lines is set to 1 if no field provided
+    double analysisWeight = 1;
+    if ( !mWeightField.isEmpty() )
```
nyalldawson: if mWeightField is empty, I'd avoid populating the feature weight hash (and also avoid looking it up in the analysis step), it'll save some redundant processing time

src/analysis/processing/qgsalgorithmlinedensity.cpp
```diff
@@ -0,0 +1,216 @@
...
+          if ( engine->intersects( lineGeom.constGet() ) )
+          {
+            double analysisLineLength =  mDa.measureLength( QgsGeometry( engine->intersection( mIndex.geometry( id ).constGet() ) ) );
+            double weight = mFeatureWeights.value( id );
+            absDensity += ( analysisLineLength *  weight );
+          }
+        }
+
+        //use ellipsoidal area
+        double analysisSearchGeometryArea = mDa.measureArea( mSearchGeometry );
```
nyalldawson: Another optimisation -- if absDensity is zero here, we could skip the area measurement calculation since the pixel value will always be 0.

nyalldawson (commented):
Beautiful, you've nailed it!

While we're squeezing out maximum performance, I've posted a few more suggestions which could further improve the processing time...

src/analysis/processing/qgsalgorithmlinedensity.cpp
root676: I am a little confused here. In the previous review you advised me to 

> build up a hash of the weights vs feature id at the same time you populate the spatial index 

This sounds very reasonable to me because only the relevant QgsFeatureIds and respective weights stay in memory and no further calls to the provider are necessary in the loops below. I understand the inefficiency of inserting the default weight (=1) if no weight field is selected and the optimization via setSubsetOfAttributes(), but why shouldn't we use the QHash here? Maybe I am missing something ;)

nyalldawson: I just mean that particular line isn't required -- you don't need to manually initialise a qhash

root676: Thx! That was the thing I was missing. ;)

nyalldawson (approved):

PR #33536, opened 2019-12-28
Fix testsqgsgdalutils on proj6

stev-0:
## Description
Fix build using Proj6 - tests were comparing std:string with QString


nyalldawson (approved):

nyalldawson:
I find this odd - the build works for me on proj 6. There's some switch or something which causes the qt macros to become super sensitive to types, which is on for Travis but not on local builds...

PR #33540, opened 2019-12-28
[Bugfix] North arrow / image should account for item rotation too.

### LINKED ISSUE #33482: North Arrow - bug when rotating Map in composer
**Describe the bug**
Working with 3.10.1 (ef24c52), Nort arrow does not rotate anymore when rotating map ...

**How to Reproduce**

1. Add map
2. Add arrow
3. Rotate map (here 15°)
4. See error, as below 

image

**QGIS and OS versions**

Version de QGIS | 3.10.1-A Coruña | Révision du code | ef24c526da
-- | -- | -- | --
Compilé avec Qt | 5.12.4 | Utilisant Qt | 5.12.4
Compilé avec GDAL/OGR | 2.4.2 | Utilisé avec GDAL/OGR | 2.4.2
Version du client PostgreSQL | 11.5 (Ubuntu 11.5-1) | Version de SpatiaLite | 4.3.0a
Version de QWT | 6.1.4 | Version de QScintilla2 | 2.10.4
OS Version | Ubuntu 19.10
Extensions Python actives | VectorBender; AnotherDXF2Shape; mask; Qgis2threejs; GeoCoding; vector_tiles_reader; ORStools; qconsolidate3; shapetools; Hqgis; DataPlotly; french_locator_filter; SRTM-Downloader; Beeline; OSMtools; QNEAT3; loadthemall; db_manager; MetaSearch; processing

roya0045:
## Description
This is my proposal to fix #33482 .

The changes implemented are the following:

Override the rotateitem method in qgslayoutitemmap to add the rotation signal.

Change the mapRotationChanged signals by adding the rotation value of the item.

Backport for 3.10 is suggested.

## Checklist

### DIFF
--- a/src/core/layout/qgslayoutitemmap.cpp
+++ b/src/core/layout/qgslayoutitemmap.cpp
@@ -2528,7 +2528,7 @@ void QgsLayoutItemMap::createStagedRenderJob( const QgsRectangle &extent, const 
 
 void QgsLayoutItemMap::rotateItem( const double angle, const QPointF transformOrigin )
 {
-  QgsLayoutItem::rotateItem( angle, tranformOrigin);
+  QgsLayoutItem::rotateItem( angle, transformOrigin);
 
   emit mapRotationChanged( mMapRotation + itemRotation() );
 }
--- a/src/core/layout/qgslayoutitemmap.cpp
+++ b/src/core/layout/qgslayoutitemmap.cpp
@@ -2528,13 +2528,7 @@ void QgsLayoutItemMap::createStagedRenderJob( const QgsRectangle &extent, const 
 
 void QgsLayoutItemMap::rotateItem( const double angle, const QPointF transformOrigin )
 {
-  double evaluatedAngle = angle + rotation();
-  evaluatedAngle = QgsLayoutUtils::normalizedAngle( evaluatedAngle, true );
-  mItemRotation = evaluatedAngle;
-
-  QPointF itemTransformOrigin = mapFromScene( transformOrigin );
-
-  refreshItemRotation( &itemTransformOrigin );
+  QgsLayoutItem::rotateItem( angle, tranformOrigin);
 
   emit mapRotationChanged( mMapRotation + itemRotation() );
 }
--- a/src/core/layout/qgslayoutitemmap.cpp
+++ b/src/core/layout/qgslayoutitemmap.cpp
@@ -447,7 +447,7 @@ void QgsLayoutItemMap::setMapRotation( double rotation )
   mMapRotation = rotation;
   mEvaluatedMapRotation = mMapRotation;
   invalidateCache();
-  emit mapRotationChanged( rotation + itemRoration() );
+  emit mapRotationChanged( rotation + itemRotation() );
   emit changed();
 }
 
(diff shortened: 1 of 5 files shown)


roya0045:
PR might not be ready for review, haven't tested, just made it so I don't forget when I get back to work.

src/core/layout/qgslayoutitemmap.cpp
```diff
@@ -447,7 +447,7 @@ void QgsLayoutItemMap::setMapRotation( double rotation )
   mMapRotation = rotation;
   mEvaluatedMapRotation = mMapRotation;
   invalidateCache();
-  emit mapRotationChanged( rotation );
+  emit mapRotationChanged( rotation + itemRotation() );
```
nyalldawson: The mapRotationChanged signal is used quite heavily in other areas of code outside of the north arrow, so I'm reluctant to change it's current behavior here. I'd suggest instead hooking into the existing QgsLayoutItem::rotationChanged signal, i.e. by making the picture item listen out for both the mapRotationChanged AND rotationChanged signals (i.e. adding new connections to rotationChanged wherever QgsLayoutItemPicture currently uses mapRotationChanged )

Also note that you'll need to update QgsLayoutItemPicture::updateMapRotation() -- it doesn't use the rotation value from the mapRotationChanged signal in any case, so it needs to be updated to base the picture orientation on mRotationMap->mapRotation() + mRotationMap->rotation().

nyalldawson (commented):
Thanks for working on this one!

src/core/layout/qgslayoutitemmap.cpp
roya0045: I'm confused as to why the value provided was useless, I saw the function and thought it was overloaded somewhere else to accept the given value and couldn't find that overloaded method. + the holiday lack of sleep doesn't help. I was debating simply changing the function at first but didn't want to alter some core component if it could be prevented.

src/core/layout/qgslayoutitemmap.cpp
```diff
@@ -2526,3 +2526,10 @@ void QgsLayoutItemMap::createStagedRenderJob( const QgsRectangle &extent, const
   mStagedRendererJob->start();
 }
 
+void QgsLayoutItemMap::rotateItem( const double angle, const QPointF transformOrigin )
```
nyalldawson: This override can be dropped now, it's not required

nyalldawson (commented):
Nearly there!

src/core/layout/qgslayoutitemmap.cpp
nyalldawson: It's not useless, it's just not used by the north arrow code. Check the QgsLayoutItemPicture::updateMapRotation() slot, it doesn't have an argument accepting the rotation value used in the signal.

nyalldawson:
There's also one connection missed in QgsLayoutItemPicture::finalizeRestoreFromXml()

src/core/layout/qgslayoutitempicture.cpp
```diff
@@ -454,7 +455,7 @@ void QgsLayoutItemPicture::updateMapRotation()
     return;
 
   // take map rotation
-  double rotation = mRotationMap->mapRotation();
+  double rotation = mRotationMap->mapRotation() + mRotationMap->itemRotation();
```
nyalldawson:
```suggestion
  double rotation = mRotationMap->mapRotation() + mRotationMap->rotation();
```

This one is subtle, but check the docs for itemRotation as there's an explanation in those

roya0045: since the only way to access the attribute in the layoutitem was through this I thought it was the only way to get this variable. I forgot to read the doc of the existing variable and only checked the output. My mistake :/

nyalldawson: Totally understandable.. I missed this one too!   It's a rather poor API, but can't be fixed because the underlying qt class method isn't virtual...

nyalldawson (approved):

PR #33552, opened 2019-12-30
Add native affine transform algorithm for vectors

### LINKED ISSUE #33550: #BUG , Scale resize from toolbox, not affine
Hi I wonder if we can have a new feature that allows you to select the layer or feature  or more importantly group and use a tool box button to drag its size like in other paint programs ? The Affine (rotation , Scale ) is good but practically difficult to use could this not be implement into the tool box. It would make things much more easier to do by dragging to position  extending one side?

Is it also possible to be able to select all your features and set the ownership publisher details please even by group to set everything?

Thanks for looking

nyalldawson:
Offers the following benefits over the GRASS/SAGA versions:
- Full support for z/m values and handling curved geometries without loss
of curves
- Works with all native data types, no need for format transformation
- Supports dynamic (data defined, per feature) translate/scale/rotate parameters
- Allows transformation and scaling of both Z and M values (if present)
- Supports in-place edit mode

Fixes #33550

### DIFF
--- a/python/plugins/processing/tests/testdata/expected/affine_transform.gml
+++ b/python/plugins/processing/tests/testdata/expected/affine_transform.gml
@@ -1,0 +1,77 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<ogr:FeatureCollection
+     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+     xsi:schemaLocation="http://ogr.maptools.org/ affine_transform.xsd"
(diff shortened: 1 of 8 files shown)


nirvn (approved):
Looking good. Very nice addition to our growing native alg. toolbox. Thanks for another Christmas gift.

MPN-stack:
Very grateful of this lovely gift, how do we get hold of it please?

alexbruy (approved):
Nice! +1 to merge

PR #33557, opened 2019-12-30
[Backport Release-3_10] Fix inconsistent dependency to project in QgsRelation/QgsRelationManager

rldhont:
Manually backport #33476


nyalldawson:
I suspect the issue is that the default constructed context, which is used here to maintain API, is created with a null QgsProject reference. That should default to the QgsProject instance to properly maintain stable API.

nyalldawson:
Actually ignore that comment -- it's obviously wrong :)

m-kuhn (approved):

rldhont:
thanks @m-kuhn

PR #33567, opened 2019-12-31
Fix validity mask filename quoting in Sieve and Fillnodata algorithms

### LINKED ISSUE #33562: Processing parameter filename needs quoting: Sieve
QGIS 3.10.1, OSGeo4W Windows install.

In the Sieve processing algorithm, QGIS needs to quote (" " and likely Windows path \ to \\ ) the mask parameter in the GDAL/OGR console call. Without it, leads to a processing error if the pathname of the mask layer has spaces, etc.  The input and output filenames are quoted properly. See problematic console call line generated in the attached picture. Thanks!

image

agiudiceandrea:
## Description
Fixes validity mask filename quoting in Sieve algorithm (Fixes #33562)

It also fixes a similar unreported bug in Fillnodata algorithm.

The patch could be backported to 3.10 and 3.4 branchs.

## Checklist

### DIFF
--- a/python/plugins/processing/algs/gdal/fillnodata.py
+++ b/python/plugins/processing/algs/gdal/fillnodata.py
@@ -133,7 +133,8 @@ class fillnodata(GdalAlgorithm):
 
         mask = self.parameterAsRasterLayer(parameters, self.MASK_LAYER, context)
         if mask:
-            arguments.append('-mask {}'.format(mask.source()))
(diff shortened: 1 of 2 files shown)


nyalldawson (approved):

nyalldawson:
(this isn't suitable for a 3.4 backport -- only crash fixes and data corruption fixes should be backported at this stage in 3.4's life)

PR #33572, opened 2020-01-01
Add constructor for QgsSpatialIndex which allows for a callback during bulk load

nyalldawson:
... and use this in some c++ processing algorithms which populate a spatial index direct from an iterator.

Quick benchmarks: cuts processing time for Delete Duplicate Geometries to 70% of original for a 3 million point input file (debug build)


nirvn (approved):
+1 to merge

PR #33575, opened 2020-01-02
Rotate expression function (with followups)

nyalldawson:
Supersedes #33125


nirvn:
@nyalldawson , thanks for implementing the optional center logic, looking good. +1 to merge.

PR #33579, opened 2020-01-02
Spatial join c++ port (with follow ups)

nyalldawson:
Supersedes #33192, implementing the final cleanups and fixes required for merge


nyalldawson:
For reference:

2.5million point file, joined against 3500 polygon file, release build:

Old python version: 208 seconds
New c++ version: 69 seconds

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,387 @@
+/***************************************************************************
+                         qgsalgorithmjoinbylocation.cpp
+                         ---------------------
+    begin                : January 2020
+    copyright            : (C) 2020 by
```
alexbruy: Probably these should not be empty

src/analysis/processing/qgsalgorithmjoinbylocation.h
```diff
@@ -0,0 +1,79 @@
+/***************************************************************************
+                         qgsalgorithmjoinbylocation.h
+                         ---------------------
+    begin                : January 2020
+    copyright            : (C) 2020 by
+    email                :
```
alexbruy: Probably these lines should not be empty

alexbruy (approved):
Looks good to me

src/analysis/processing/qgsalgorithmjoinbylocation.cpp
```diff
@@ -0,0 +1,387 @@
+/***************************************************************************
+                         qgsalgorithmjoinbylocation.cpp
+                         ---------------------
+    begin                : January 2020
+    copyright            : (C) 2020 by Alexis Roy-Lizotte
```
roya0045: I would like more something like Alex RL and alroyliz0 at gmail dot com.

roya0045:
Thanks for taking over, I knew that a more experienced programmer would implement this quicker but I wanted to help. Though I'm not sure if it helped or slowed things down. :/ Nice improvements also.

nirvn:
@roya0045 , wanting to help is a fantastic state of mind, and the QGIS project has benefited from your contributions already. 

My main suggestion to you moving forward: as a 2020 new year resolution, force yourself to stop using QGIS' github repository as your test ground in favor of a _proper_ development environment on your local machine. ATM, I have to mute most of your PRs to avoid having my inbox filling up with commits messages that would never reach the repository if you would compile locally.

VM, dual boot, whatever. Make it your resolution :)

Also, keep an open state of mind when senior developers are suggesting changes. We as a project are very, very lucky to have a couple of amazingly talented C++ / Qt gurus, and their knowledge is how we can all get better.

Thanks for your contributions in 2019; I hope you can move beyond those obstacles so you can grow as a coder and contribute even more in 2020 :smile:

roya0045:
@nirvn Thanks for the feedback, as far as VM goes yes I now have one at work and this is how I have been able to improve my PR faster, but since its the holidays I am at home and didn't have access to my VM and with limited internet I only could do things on github directly.

Though I agree that I should test things more and take more my time with things even when I think they are just a quick PR.

And you are right that the core dev of Qgis are truly experts in their craft. 

Thanks for the support!

PR #33599, opened 2020-01-04
Add option for multiple field parameters to default to selecting all available fields

nyalldawson:
For some algorithms this is better UX then defaulting to an empty list


alexbruy (approved):
Looks good to me.

Am I right that `defaultToAllFields` and `defaultValue` are mutually exclusive? Probably this should be mentioned in the doxygen comment.

nyalldawson:
> Am I right that defaultToAllFields and defaultValue are mutually exclusive? Probably this should be mentioned in the doxygen comment

That's correct -- I'll document it

Issue #33317, opened 2019-12-10
Raster calculator in Graphical Modeler Fails in Spanish, works in English


mudomau:
**Describe the bug**

I was trying to automatize a series of processing steps using the Graphical Modeler. The steps worked OK when performed one by one outside the graphical modeler, but when I tried to replicate them inside the modeler the raster calculator (final step) would return the following error: 

![imagen](https://user-images.githubusercontent.com/2564029/70540342-5df1dc80-1b65-11ea-8e22-9a034d2dab50.png)

The process is the following:
1) Calculate slope for a DEM
2) Calculate aspect for a DEM
3) Extract North-South slopes using the raster calculator with the expression: Slope*cos((3.141592/180)*Aspect)^2

After changing the locale to English_Canada and User Interface Translation to American English, the process works.

**How to Reproduce**

1) In QGIS 3.8 or 3.10, set locale to Spanish Spain and User interface translation to Español de España

2) Open Processing->Graphical Modeler

2) Create one raster layer input,

3) Calculate the Slope and Aspect of the input (using the default processing algorithms), then use the raster calculator to calculate any expression

This fails for as long as the language is set to spanish, for all expressions tried in the Raster calculator.

![imagen](https://user-images.githubusercontent.com/2564029/70540394-77932400-1b65-11ea-94e8-d15622bd9ce0.png)

**QGIS and OS versions**

Tested in Windows 10, QGIS 3.8 and 3.10

**Additional context**

gioman:
@mudomau duplicate of https://github.com/qgis/QGIS/issues/32753 ?

elpaso:
I've found the source of the issue and it's a translation problem.

I must say that the underlying implementation is extremely weak:  it relies on the fact that two strings in two different places of the code are translated exactly in the same way.

I'm checking if is there a way to make it more robust.

elpaso:
Our experienced developer @luipir has accepted the challenging task to fix this nasty bug, I'm happily passing it to him :)

btw, the offending commit is faf9716a52e (Luigi Pirelli      2018-07-21 17:28:40 +0100) and the source of the problem is in `modeler/ModelerParametersDialog.py:                return self.tr("'{0}' from algorithm '{1}` and `RasterCalculator.py:                translatedDesc = self.tr("Output '%1' from algorithm '%2'")`

[...]

PR #33601, opened 2020-01-04
fix translation problem when changing locale with not translated formula. Fixes #32753

### LINKED ISSUE #32753: Modeler and Raster Calculator fail to calc
Hi all,

context: Raster Calculator added in a Processing model + 1 rasters as input
Using the Raster Calculator module inside a Model I have an output always as float raster with [-1.7: 1.7]

This only happens if the interface is in Italian or Spanish.
Everything works correctly with the interface in English.
I have not verified with other languages

Maybe there's a related issue at
https://github.com/qgis/QGIS/issues/27130

Thanks

luipir:
## Description
This fix a #32753 locate translation problem. Probably not completly fix also #33317 on which I've to investigate more.
The fix try to do the do variable lookup inside translated formula less translation dependent => more robust

## Checklist

### DIFF
--- a/python/plugins/processing/algs/qgis/RasterCalculator.py
+++ b/python/plugins/processing/algs/qgis/RasterCalculator.py
@@ -259,8 +259,9 @@ class RasterCalculator(QgisAlgorithm):
                 #    Output 'varName' from algorithm 'algName'
                 # but var in expression is called simply
                 #    'varName' from algorithm 'algName'
-                # because variable description translation (issue #32753 and #33317) it's necessary to abstract translation 
-                # in some way. e.g. parsing var description
+                # because variable description translation (issue #32753 and #33317)
+                # it's necessary to abstract translation in some way.
+                # e.g. parsing var description
                 elements = varDescription.split("'")
                 if len(elements) == 5:
                     outVarName = elements[1]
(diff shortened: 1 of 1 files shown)


nyalldawson:
This all looks very fragile. I think we need to revisit the original fix here and try something more robust.

A possible approach could be to assign a uuid to all the ouputs. Immediately after the expression is built in the widget the strings displaying layer names in the expression are replaced with the uuid, and that's the version that is stored in the model. So we'd never be storing the translated strings as part of th expression...

luipir:
> A possible approach could be to assign a uuid to all the ouputs

hmmm... I hadn't time to think all possible use cases, but I almost agree to refactor this part. IMHO a mechanism similar to gdal raster calculator, e.g associate symbolic name to variables but not related to UUID that can change for each run.
IMHO a refactoring would maintain the following characteristic:
1) formula should visualise reference to the source algorithm (as now). this is different solution respect gdal calculator
2) formula should have a shadow version that is translation independent and it is linked with the formula (=> we could loose the ability to have a simple cut&paste other then reintroducing remapping and translation problems)
3) visible and shadow formula should be saved in the model
4) shouldn't brake model where raster calculator is already used (sadly a remapping is still necessary :/ ) but probebly there is the original error I did to avoid a deep refactoring of the alg when I reintrodced.

any suggestion how to proceed @nyalldawson?

luipir:
I'll reopen it later when I'll have time to do a refactoring of the calculator... it's a shame :/

agiudiceandrea:
Hi @luipir, was this PR refactored elsewhere?

luipir:
> Hi @luipir, was this PR refactored elsewhere?

no... I hadn't time to dig it more

PR #33608, opened 2020-01-04
New parameter type for map themes

nyalldawson:
Fixes an api break in the rasterize algorithm, and then adds a proper dedicated parameter type for map theme selection


src/core/processing/qgsprocessingparametertypeimpl.h
```diff
@@ -1729,5 +1729,58 @@ class CORE_EXPORT QgsProcessingParameterTypeCoordinateOperation : public QgsProc
     }
 
 
+};
+
+
+/**
+ * A map themeparameter for Processing algorithms.
```
alexbruy: Missed space?

alexbruy (approved):
Looking good. +1 to merge

Issue #29438, opened 2019-03-20
A ghost blinking window appears and suddenly disappears upon launch of processing tools which use QgsProcessingParameterDistance


qgib:
Author Name: **Andrea Giudiceandrea** (@agiudiceandrea)
Original Redmine Issue: [21622](https://issues.qgis.org/issues/21622)
Affected QGIS version: 3.7(master)
Redmine category:processing/gui

---
On Windows 7 64 bit with QGIS LTR 3.4.5-1 (89ee6f6e23) and QGIS master 3.7.0-10 (1205fbfa71), a ghost blinking window appears and suddenly disappears upon launch of all and only processing tools which use QgsProcessingParameterDistance for one or more parameters.

!Video_2019-03-20_035131.gif!

Some affected processing tools:

gdal:
Buffer vectors
One side buffer
Offset curve

qgis:
Create grid
Heatmap
Points along geometry
Points displacement
Pole of inaccessibility
Random points along line
Random points in extent
Random points in layer bounds
Random points inside polygons
Rectangles, ovals, diamonds (fixed)
...

native:
Array of offset (parallel) lines
Array of translated features
Extend lines
Line substring
Remove duplicate vertices
Snap points to grid
Transect
Translate
...

---

- [Video_2019-03-20_035131.gif](https://issues.qgis.org/attachments/download/14630/Video_2019-03-20_035131.gif) (Andrea Giudiceandrea)

alexbruy:
Is this still an issues? I don't see any blinking window when opening algorithm dialog.

PR #33610, opened 2020-01-05
Avoid temporary flashing dialog showing when a distance parameter widget is constructed

nyalldawson:
…Fixes a bug I can't find the report for anymore...


alexbruy:
Seems this ticket #29438

PR #33611, opened 2020-01-05
Fix effect preview picture not set when effect widget is opened in dialog mode

### LINKED ISSUE #33600: Draw effects dialog does not display the actual symbol but uses an abstract red circle
On the road to update the screenshots [in the docs](https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/vector_properties.html#draw-effects), I realize that it's not possible to display the current symbol in the effects dialog. I think this was possible in earlier versions as shown in this screenshot
image
which imho is more helpful to visualize the changes on my symbol than the abstract symbol below
image

Btw, isn't there an issue with the drop shadow preview in the second screenshot?

Using QGIS 3.10.1

nyalldawson:
Fixes #33600


nirvn (approved):

DelazJ:
Thanks @nyalldawson
Very helpful, a usual

PR #33618, opened 2020-01-06
Accept newline terminated string as temporary scratch layer

### LINKED ISSUE #33617: Pasting features as temporary scratch layer fails silently if the last character is a newline
Working with geometries built outside of QGIS, if the last line (i.e. last feature) is terminated with a newline character, using `Edit > Paste Features As > Temporary Scratch Layer...` (or Ctrl+Alt+V) fails silently.

**How to Reproduce**

Copy the tab-delimited table below, if the selection ends right after 3 on the last line, pasting works as expected. If the selection extends one more character after 3 (which is a newline character), pasting fails.

image => OK
image => Fails

```
wkt_geom	id	data	name	x	y
Point (-73.966559 40.781324)	0	954	Poi #0	-73.966559	40.781324
Point (-73.966569 40.781324)	1	2514	Poi #1	-73.966569	40.781324
Point (-73.966539 40.781324)	2	3272	Poi #2	-73.966539	40.781324
Point (-73.966589 40.781323)	3	2161	Poi #3	-73.966589	40.781323

```

**QGIS and OS versions**

Windows 10: 3.4.14, 3.10.1, ltr-dev, rel-dev, dev
Ubuntu 18.04: 3.14.11

**Note**
* This looked like an easy fix (even though I have never written c++ before), so I will submit a PR in case it helps.

i-s-o:
This PR adds code to remove the last character of the string pasted from the clipboard if it is a newline character.

Also, I noticed that `QgsClipboard::retrieveFields()` is called three times for the paste action, and, the text in the clipboard is read and split into separate lines each time. If the data in the clipboard is large, wouldn't that be inefficient?

I tested this with 50000 points, and If I am measuring it correctly, it took about 850 ms to parse all those lines (just to extract the names of the fields). Reading only the first line (i.e. the modification in this PR) took about 55 ms on my VM. Any feedback on this point is appreciated.

I haven't written any unit tests because I don't know how [yet].

Fixes #33617

### DIFF
--- a/src/app/qgsclipboard.cpp
+++ b/src/app/qgsclipboard.cpp
@@ -309,6 +309,16 @@ QgsFeatureList QgsClipboard::copyOf( const QgsFields &fields ) const
   QString text = cb->text( QClipboard::Clipboard );
 #endif
 
+  if ( text.endsWith( '\n' ) )
+  {
+    text.chop( 1 );
+    // In case Windows <EOL> marker (CRLF) makes it into the variable "text"
+    if ( text.endsWith( '\r' ) )
+    {
+      text.chop( 1 );
+    }
+  }
+
   return stringToFeatureList( text, fields );
 }
 
--- a/src/app/qgsclipboard.cpp
+++ b/src/app/qgsclipboard.cpp
@@ -270,10 +270,10 @@ QgsFields QgsClipboard::retrieveFields() const
     }
 
     //wkt?
-    QStringList lines = string.split( '\n' );
-    if ( !lines.empty() )
+    QString firstLine = string.section( '\n', 0, 0 );
+    if ( !firstLine.isEmpty() )
     {
-      QStringList fieldNames = lines.at( 0 ).split( '\t' );
+      QStringList fieldNames = firstLine.split( '\t' );
       //wkt / text always has wkt_geom as first attribute (however values can be NULL)
       if ( fieldNames.at( 0 ) != QLatin1String( "wkt_geom" ) )
       {
--- a/tests/src/app/testqgisappclipboard.cpp
+++ b/tests/src/app/testqgisappclipboard.cpp
@@ -331,6 +331,40 @@ void TestQgisAppClipboard::pasteWkt()
   QCOMPARE( features.at( 2 ).attributes().at( 0 ).toString(), QStringLiteral( "2" ) );
   QCOMPARE( features.at( 2 ).attributes().at( 1 ).toString(), QStringLiteral( "b2" ) );
   QCOMPARE( features.at( 2 ).attributes().at( 2 ).toString(), QStringLiteral( "3" ) );
+
+  // when a set of features is built outside of QGIS, last one might be terminated by newline
+  // https://github.com/qgis/QGIS/issues/33617
+  mQgisApp->clipboard()->setText( QStringLiteral( "POINT (125 10)\nPOINT (111 30)\n" ) );
+  features = mQgisApp->clipboard()->copyOf();
+  QCOMPARE( features.length(), 2 );
+  QVERIFY( features.at( 0 ).hasGeometry() && !features.at( 0 ).geometry().isNull() );
+  QCOMPARE( features.at( 0 ).geometry().constGet()->wkbType(), QgsWkbTypes::Point );
+  featureGeom = features.at( 0 ).geometry();
+  point = dynamic_cast< const QgsPoint * >( featureGeom.constGet() );
+  QCOMPARE( point->x(), 125.0 );
+  QCOMPARE( point->y(), 10.0 );
+  QVERIFY( features.at( 1 ).hasGeometry() && !features.at( 1 ).geometry().isNull() );
+  QCOMPARE( features.at( 1 ).geometry().constGet()->wkbType(), QgsWkbTypes::Point );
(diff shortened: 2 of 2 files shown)


nyalldawson:
> I tested this with 50000 points, and If I am measuring it correctly, it took about 850 ms to parse all those lines (just to extract the names of the fields). Reading only the first line (i.e. the modification in this PR) took about 55 ms on my VM. Any feedback on this point is appreciated.

Good change!

src/app/qgsclipboard.cpp
```diff
@@ -309,6 +309,11 @@ QgsFeatureList QgsClipboard::copyOf( const QgsFields &fields ) const
   QString text = cb->text( QClipboard::Clipboard );
 #endif
 
+  if (text.endsWith('\n'))
```
nyalldawson: For windows users we better check for and remove carriage returns too

nyalldawson:
@i-s-o 
Thanks for this fix! 

A couple of things to note:
- The CI will test to make sure that consistent formatting is applied to code (spacing etc). This PR will fail because some of the changes violate the formatting rules. To avoid this, you need to run the `scripts/prepare-commit.sh` script which will apply the formatting rules. This is designed to be run as a pre-commit script by git. If you need to fix formatting after making a commit, you can use `scripts/astyle-all.sh`.

- There's unit tests for this functionality at https://github.com/qgis/QGIS/blob/master/tests/src/app/testqgisappclipboard.cpp#L225 which could b e extended to add some input strings with newline terminating characters

nyalldawson:
Also, if you add the string "Fixes #33617" to your git commit message, the corresponding ticket will be auto-closed when this PR is merged

src/app/qgsclipboard.cpp
i-s-o: I was actually "watching" the Windows clipboard as I was testing (with FreeClipViewer): "\n" alone did fix the problem when the last two characters were `CRLF` in the clipboard. I assumed Qt5 was normalizing newlines to `LF` as it read from the clipboard. I will double check and modify as necessary.

i-s-o:
> To avoid this, you need to run the `scripts/prepare-commit.sh` script

Couldn't get this running on Windows: investigating...

> add the string "Fixes #33617" to your git commit message

Got it. I wasn't sure whether I needed to do this in the commit message or the PR comment: I picked the wrong one in my first attempt :-). (I think I need it in both places after all).

DelazJ:
> Got it. I wasn't sure whether I needed to do this in the commit message or the PR comment: I picked the wrong one in my first attempt :-). (I think I need it in both places after all).

Not really, as long as it's in the commit OR in the PR, it will close the issue (Eg, hover over the "fixes" term in your PR message).

nyalldawson:
> Not really, as long as it's in the commit OR in the PR, it will close the issue (Eg, hover over the "fixes" term in your PR message).

I didn't realise that. My preference would still be to put them in the commit message, because that's permanently attached to the commit and visible in the git history/blame (vs pull request content, which is lost after merge)

DelazJ:
As long as you keep the same issue tracker, don't you? ;)
Not the first time I read the advice/requirement to add to the commit message while PR is enough. And I could not let @i-s-o feel he made it the wrong way while it isn't the case
> I wasn't sure whether I needed to do this in the commit message or the PR comment: I picked the wrong one in my first attempt :-). (I think I need it in both places after all).

nyalldawson:
> As long as you keep the same issue tracker, don't you? ;)

That doesn't change anything -- if I'm git blaming a line of code, all I see is the commit message, not the corresponding PR or the PR description. Having the issue in the commit message means I can track back why the change was made... having it in the PR means I have no clues at all to work from.

We should strongly discourage including this in the PR message -- it belongs in the commit log only.

i-s-o:
@nyalldawson & @DelazJ thank you for all the feedback (and your patience)! Here is where I'm at:

> > To avoid this, you need to run the `scripts/prepare-commit.sh` script
> 
> Couldn't get this running on Windows: investigating...

Got this working in Windows following these steps:
* Rebuild the solution with `WITH_ASTYLE TRUE` option
* Install `Cygwin`
* Add git, `dos2unix`, `python`, and `pip` packages to Cygwin
* Add `autopep8` to python
* Run `scripts/prepare-commit.sh` in Cygwin Terminal

> after making a commit, you can use `scripts/astyle-all.sh`

I'm not sure you really meant this one: it took more than an hour to run and touched all the source files. Now, `git status` shows thousands of files as modified!

> There's unit tests for this functionality at ... which could be extended

I did this, but I'm not sure whether I'm running this correctly (I used `ctest -V -R clipboard` in `build-qgis-test-x86_64` dir): I got the following message:

```
1/1 Test #245: qgis_qgisappclipboard ............Exit code 0xc0000135
***Exception:   0.14 sec
```

I don't know how to proceed. Should I just push my [modified] local commits, and rely on CI for the unit tests?

luipir:
> I don't know how to proceed. Should I just push my [modified] local commits, and rely on CI for the unit tests?

just follow the forma suggestions in the test https://travis-ci.org/qgis/QGIS/jobs/633117974?utm_medium=notification&utm_source=github_status

btw would be better to have ASTYPE a nd prepare_commit hook configured when creating PRs.

nyalldawson (approved):

nyalldawson:
Looks great, thanks!

PR #33619, opened 2020-01-06
GPS fixes and rotation matching follow up

nyalldawson:
Fixes some NMEA GPS connection issues, and adds an explicit setting to allow control over the frequency of map rotation to match GPS bearing (Otherwise frequent GPS updates can cause a too many map redraws to occur quickly)


Saijin-Naib:
Oh how I wish I had access to this two years ago when I was still in precision agriculture.

I still managed to move everyone onto QGIS on Win10 tablets for field usage, but the bearing/rotation would have made things 1000x better for field data collection.

Incredible work, Nyall.

PR #33624, opened 2020-01-06
[api] Add registry and interface for QgsNumericFormats

nyalldawson:
QgsNumericFormat subclasses provide a means to format a numeric value
as a string, applying various formatting options. E.g. a default
string to value formatter (included here) includes settings for
controlling the number of decimal places, whether a thousands separator
should be shown, whether a leading + sign should be shown, whether
trailing zeros should be shown.

This PR also includes a formatter for bearings, allowing various
formats of numeric bearings to be applied (e.g. control over decimal
places, etc + control over whether direction E/W suffixes are shown,
or whether values should be limited to either the +/- 180 range or
0-360 degree range)

When formatting values, a QgsNumericFormatContext class is used
to provide context. Currently, this includes the thousands and
decimal separators to apply when formatting (which are taken by
default from the user's locale).

A registry of formatters is included to allow easy addition of
other formats in future and to allow plugin based formats.

The intention is to follow this up with gui configuration widgets
for the formats, and then expose them in various places through
the qgis ui (e.g. in the range editor widget for fields, in scalebar
numbers, as an option for formatting numeric labels, etc)


3nids:
woohoo, so nice!

src/core/numericformats/qgsbasicnumericformat.h
```diff
@@ -0,0 +1,133 @@
...
+
+    //! QgsBasicNumericFormat cannot be copied
+    QgsBasicNumericFormat( const QgsBasicNumericFormat & ) = delete;
+    //! QgsBasicNumericFormat cannot be copied
+    QgsBasicNumericFormat &operator=( const QgsBasicNumericFormat & ) = delete;
+
+    QString id() const override;
+    QString formatDouble( double value, const QgsNumericFormatContext &context ) const override;
+    QgsNumericFormat *clone() const override SIP_FACTORY;
+    QgsNumericFormat *create( const QVariantMap &configuration ) const override SIP_FACTORY;
```
3nids: isn't it static?

nyalldawson: I'm using the prototype pattern here, so it needs to be a virtual member

nyalldawson:
Thanks for the speedy review @3nids!

nyalldawson:
@elpaso would you mind looking over this one too? I know you've worked on similar issues before

nyalldawson:
Now with gui configuration widgets, and formatters for scientific notation, percentages, and currency values!

Screencast below:

Peek 2020-01-07 13-09

nirvn:
@nyalldawson , fantastic job. This'll be very useful for the project properties' coordinate display section too:
Screenshot from 2020-01-07 10-45-40

I'm wondering whether it'd be useful to have a degrees format in there, and list "decimal, degree minutes, degree minutes seconds" as format option (like your doing for bearing).

nyalldawson:
> I'm wondering whether it'd be useful to have a degrees format in there, and list "decimal, degree minutes, degree minutes seconds" as format option (like your doing for bearing).

Yes - that would be good to add in future (should be easy with the new api), and could also allow us to reuse this formatter within layout map grid annotations...

PR #33629, opened 2020-01-06
Support for Oracle curves and surfaces

troopa81:
## Description

This PR adds edition support for Oracle following geometry type: 
- CircularString(Z)
- CompoundCurve(Z)
- MultiCurve(Z)
- CurvePolygon(Z)
- MultiSurface(Z)

I remove the limitation (86269c6a72) on editing and adding geometry with Z coordinates. Is the limitation still relevant? Why it would "lead to corruption of measure"? @jef-n 

It's also fixes 3 issues : 
- 7a120eeaa7 Fix multipolygons with only one polygon
- a32dec9557 Fix read for oracle curve polygon with Z coordinates
- f2ffbec507  When reading elem info array with compound curve in multiline, we skip some elem info triplet when incrementing too much **iElem**. 
The commit also fixes the tests. Indeed when defining a multiline (DL06), it's possible to use compound line string with SDO_ELEM_INFO triplet equals to offset,4,n [where n must be greater than 1](https://docs.oracle.com/database/121/SPATL/sdo_geometry-object-type.htm#GUID-270AE39D-7B83-46D0-9DD6-E5D99C045021__BGHDGCCE). This was not always the case in test data. These instructions are not valid (regarding oracle documentation) and hide the issue.
I also change EPSG code for lines from 4326 to  3857 to avoid this message: `ORA-13035: Invalid data (arcs in geodetic data) in the SDO_GEOMETRY object`

I'm planning to backport this 3 commits once I have backported #32337

@Jean-Roc 

## Checklist


rldhont (approved):

m-kuhn:
Thanks @troopa81 !
@rldhont what does your approval mean, did you check the code, test it or just like the idea?

rldhont:
> @rldhont what does your approval mean, did you check the code, test it or just like the idea?

@m-kuhn I have check the code.

Issue #33594, opened 2020-01-03
QgsSymbol->ruleToExpression


roya0045:
**Feature description.**
We already have some ways to convert symbol rules to expression but they are independent and hard to leverage. 

The requested feature is to have a function/method that would return the full rule/expression to obtain this symbol. Meaning that the symbol itself would be taken into account but also its parent's rule (in the case of rule-based).

I've had a look at this already and was not too sure how to do this properly due to the complexity of the related elements.

roya0045:
I've looked at an other angle to achieve this and there are two points that I'm not sure about, are every QgsSymbol unique and is there a way to get the 'owner' of a given QgsSymbol?

I am hoping to be able to access the rule/category/range of a given symbol store in a QgsSymbolLegendNode. But it seems that the rendering structure and the files structures are somewhat disjointed.

nyalldawson:
>  are every QgsSymbol unique 

no - symbols are cloned in many circumstances (e.g. during preparation for background rendering)

> and is there a way to get the 'owner' of a given QgsSymbol?

no - that's not possible

> I am hoping to be able to access the rule/category/range of a given symbol store in a QgsSymbolLegendNode

calling QgsSymbolLegendNode::data and requesting the RuleKeyRole will give you the unique legend rule key, which matches to the rule key from  QgsFeatureRenderer::legendSymbolItems()
or QgsFeatureRenderer::legendKeysForFeature

roya0045:
> > are every QgsSymbol unique
> 
> no - symbols are cloned in many circumstances (e.g. during preparation for background rendering)
 What I meant was: if two layers have the same symbology, are they using the same QgsSymbol or are they distinct? I would assume the latter.

> calling QgsSymbolLegendNode::data and requesting the RuleKeyRole will give you the unique legend rule key, which matches to the rule key from QgsFeatureRenderer::legendSymbolItems()
> or QgsFeatureRenderer::legendKeysForFeature

i remember trying that but not having the desired results, I'll test it again this week.

roya0045:
@nyalldawson thanks to your input I think I have found a way using the QgsLayerTreeLayer as the bridge. I should be able to upload the PR once I fix it up a little and implement a test.

roya0045:
Proposed PR is awaiting review or merge.

PR #33632, opened 2020-01-06
[Feature] Symbol to expression

roya0045:
## Description
This PR aims to convert symbologies to expression in order to use them later. This is a step towards @m-kuhn 's approach instead of storing fids ( I need to do timing comparison between the two approaches  for layers of varying size), see [ this comment](https://github.com/qgis/QGIS/pull/30727#issuecomment-534149312.).

Otherwise in general this can be an utility thing.

This mostly leverage existing code and add minimal functions to allow the storage and transfer of information.

This is related to #33594 

The use of this expression is twofold:

- to add a way for the user to properly display the full expression (mostly with nested symbology)
- to allow the user to use said expression as a filter in functions.

## Example
symbol_exp

## Checklist


nirvn:
@roya0045 , please read the instructions (https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTING.md#contributing-to-qgis) to set up the pre commit hook in your dev. environment. As you can see with the travis failure, you have indentation issues which would have been caught when you git commit your change when the pre commit check runs locally. 

Don't check the "I have run the scripts/prepare-commit.sh" item if you have not run the script.

roya0045:
> @roya0045 , please read the instructions (https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTING.md#contributing-to-qgis) to set up the pre commit hook in your dev. environment. As you can see with the travis failure, you have indentation issues which would have been caught when you git commit your change when the pre commit check runs locally.
> 
> Don't check the "I have run the scripts/prepare-commit.sh" item if you have not run the script.

The scripts has issues on my windows VM. It tends to simply error, I'll try to debug it.

roya0045:
Unsure as to why the test fails, the expression worked when I tested it locally. Forcing caching does not seem to help either in the travis tests.

nirvn:
@roya0045 , a functional dev. environment includes having the pre-commit script run flawlessly, continue to focus on getting your dev. environment up and running as a priority.

I haven't reviewed the code of this PR; that said, my gut feeling here is that you'll have a hard time getting it reviewed as the goals and _practical uses_ of the proposed changes aren't clearly stated.

roya0045:
@nirvn I debugged the bash code and adapted it to my need in order to get it to work. One problem was that the EOL were converted to windows style instead of unix.

I have also added more details to the description, let me know if this is clearer.

roya0045:
Ok found the issue, the hash generated by the converted symbology is not the same value as the one generated by other symbologies. I'll try to find a way to deal with this.

roya0045:
@m-kuhn @nyalldawson The tests that I did locally are all good. If you have some time for a review it would be appreciated.

roya0045:
I have added code to handle else conditions.

roya0045:
Added example to showcase formatting/conversion.

roya0045:
Azure test failure does not appear to be related.

roya0045:
Symbol_expression doesn't seem to work as a filter, this is curious.

roya0045:
@m-kuhn @nyalldawson I forgot that double evaluation was not possible with expressions. To make this functionnal and as an alternative to my other approach (using fids and a new expression parameter) I would need to evaluate this expression twice, once to get the string and the other time to evaluate the condition.
I have two ways to do this:
1) evaluate only the variable first in the legend to replace in the expression before evaluation. 
2) implement a way to evaluate an element more than once in the expression engine/parser.

Both can be done sequentially, but the first might be more safe.

roya0045:
Reviews would be appreciated

roya0045:
@nyalldawson any hopes for a review when you got some time?

roya0045:
@nyalldawson simple ping if you could look at this, it shouldn't break anything.

roya0045:
Now that the freeze is over, could someone review this?

roya0045:
This should be safe to merge, can we do a dry run?

roya0045:
Not stale, awaiting review.

roya0045:
Closing this as #33819 contains the same code, will reopen if needed to merge only this part of the code.

Issue #25712, opened 2018-01-08
Conditional formatting is not displayed in an Attribute Table widget in Layout/Composer


qgib:
Author Name: **Etienne Trimaille** (@Gustry)
Original Redmine Issue: [17816](https://issues.qgis.org/issues/17816)

Redmine category:map_composer/printing

---
If we set a conditional formatting on a vector layer, we can't display it in a Layout.
It would be nice to have a checkbox in the Item properties about rendering the same rules as in the attribute table view.

PR #33638, opened 2020-01-07
Allow layout attribute tables to be styled using the foreground and background colors of matching conditional styles 

nyalldawson:
When the new "Apply layer conditional styling colors" option is enabled in the
layout attribute table settings, any conditional styling rules present in the
layer will be applied inside the layout attribute table (foreground and
background colors only, for now!).

Refs #25712

Sponsored by City of Canning


nirvn (approved):
Looking good. That makes conditional styling that much more useful.

nyalldawson:
Thanks @nirvn!

timlinux:
Ah wow - so glad to have this at last! Thanks!

PR #33642, opened 2020-01-07
Fix ZM snapping for line/polygon on 2D vertex

lbartoletti:
## Description

The snapping on a 2D layer for line/polygon with Z or M values doesn't work as expected. It returns NaN value instead of the default Z Value (or 0.0 for Measure).

before (3.10)
snap_line_2D_310 webm

after (master)
snap_line_2D_master webm

Sponsored by QWAT Group (cc @ponceta @kandre)

## Checklist

### DIFF
--- a/src/gui/qgsmaptoolcapture.cpp
+++ b/src/gui/qgsmaptoolcapture.cpp
@@ -410,6 +410,10 @@ int QgsMapToolCapture::fetchLayerPoint( const QgsPointLocator::Match &match, Qgs
       else
       {
         layerPoint = geom.constGet()->vertexAt( vId );
+        if ( QgsWkbTypes::hasZ( vlayer->wkbType() ) && !layerPoint.is3D() )
+          layerPoint.addZValue( defaultZValue() );
+        if ( QgsWkbTypes::hasM( vlayer->wkbType() ) && !layerPoint.isMeasure() )
+          layerPoint.addMValue( 0.0 );
       }
 
       // ZM support depends on the target layer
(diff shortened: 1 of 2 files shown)


haubourg:
Hi, 
concerning the M value , do I read correctly that this snaps the M value? I would rather always set 0 for measures, I really don't get what fetching a M value means in real world. 
if the case is using M for linear referencing, that would mean a side road / stream will inherit measures from another road, which is probably not what user want.

m-kuhn:
Is it possible to set the default Z value to nan?

m-kuhn:
Sorry for the rant.

I can see that with the current concept of default Z values, this pull request makes sense and I won't block it.

However I find it very troubling to see this concept being continued instead of fixing it at the heart. IMO it makes sense to have nan values.
Right now, a number of tools create nan values as Z by default. Others - like in this PR - are changing this behavior to "default Z value" here and there. I think this leads to hard to find problems down the road because "devault values" in the end are just values. They will be interpolated (a point somewhere between a real Z and a default Z will have a joker Z interpolation). They will destroy any meaningful average or other aggregate calculations...

In my opinion, "default Z" should support nan as a first measure. And then be moved as a mapping value to provider level. If a provider cannot save "unknown value" / nan / null / ... for Z, this provider should be able to be configured to map nan to a "magic value" and back. Just like a raster can have a magic no data value. In the long run this seems a safer and cleaner concept.

haubourg:
@m-kuhn totally agreed. On my side, I didn't succeed to find support for this.

lbartoletti:
@m-kuhn 
By default, the behavior of QGIS since, the introduction of the concept of default value Z, is to use it. 
As we discussed together, this PR fixes a bug. 

However, I totally agree, on the need to introduce an undefined value, just like raster snapping, etc.  

For this, we propose to write a QEP to make QGIS evolve in this direction. In the meantime, can you please merge it? :wink: 

Thank you.

m-kuhn (approved):

PR #33648, opened 2020-01-07
[windows][server] Use proj.db from proj-dev

### LINKED ISSUE #33647: QGIS Server fails to use scale dependent visibility on OSGeo4W Nightly
Features just don't appear at all, regardless of the scale (at least at reasonable scales).

Further inspection shows strange values for `@map_scale` (orders of magnitudes too big) if displayed through labels. And `@map_units` show up as `<Unknown>` for EPSG:2056 (instead of Meters).

Everything works well on QGIS desktop. On QGIS Server 3.4 as well.

m-kuhn:
It terribly fails to do any CRS related actions otherwise

Fix #33647

@nyalldawson is this the correct way to fix the proj search path?


nyalldawson:
I'm unsure -- that's a question for @jef-n. Setting PROJ_LIB is correct, but I don't know if this is the right place to do it

jef-n:
What's the point?   The regular builds don't use proj-dev - while the nightlies that do, don't include the apache config files.

m-kuhn:
The point right now is to run a master server.
But if that's not gonna be shipped in 3.12 (and if 3.10 does not suffer from that, I still needed to verify that), then let's close this.

PR #33651, opened 2020-01-07
Add Refresh action to OGC services

### LINKED ISSUE #33261: Add default setting for min/max values accuracy
Basically same as https://github.com/qgis/QGIS/issues/20181. It's possible to toggle this from Estimate to Actual on each raster, but the default is still Estimate, and there seems to be no way to configure it from settings.

### LINKED ISSUE #33621: QGIS Browser Panel item refresh cause crash
Compiling QGIS master and doing a refresh on postgres data provider connection item while a 
WMS connection item is open and there are changes ( from WMS Server ) in WMS 
layer items will cause a QGIS crash. 

This also occurs in  QGIS 3.4 and QGIS 3.10. 

Steps to reproduce

1. Add Postgres connection
2. Add WMS/WMTS connection
3. Make changes in any of the WMS/WMTS connection layers (eg. change value of the layer title), this should be done from the WMS server.
4. Expand the WMS/WMTS connection to show the layers with the expected changes.
5. Go to the Postgres connection, right click and click "Refresh".

Below is the traceback from QGIS 3.10. 

```
QgsBrowserModel::data qgsbrowsermodel.cpp:251 
QSortFilterProxyModel::data : 
QModelIndex::data : 
QStyledItemDelegate::initStyleOption : 
QStyledItemDelegate::sizeHint : 
QTreeViewPrivate::widthHintForIndex : 
QTreeView::sizeHintForColumn : 
QHeaderView::resizeSections : 
QHeaderView::sectionSize : 
QTreeViewPrivate::updateScrollBars : 
QTreeView::updateGeometries : 
QTreeView::timerEvent : 
QObject::event : 
QWidget::event : 
QFrame::event : 
QAbstractScrollArea::event : 
QApplication::notify : 
QgsApplication::notify qgsapplication.cpp:418 
QCoreApplication::notifyInternal2 : 
QEventDispatcherWin32::event : 
QApplication::notify : 
QgsApplication::notify qgsapplication.cpp:418 
QCoreApplication::notifyInternal2 : 
QCoreApplicationPrivate::sendPostedEvents : 
qt_plugin_query_metadata : 
QEventDispatcherWin32::processEvents : 
UserCallWinProcCheckWow : 
DispatchMessageWorker : 
QEventDispatcherWin32::processEvents : 
qt_plugin_query_metadata : 
QEventLoop::exec : 
QCoreApplication::exec : 
main main.cpp:1598 
BaseThreadInitThunk : 
RtlUserThreadStart : 
```

Samweli:
This PR adds refresh action to OGC Services and also fixes #33621

 For the #33621, I have override the WMSLayerItem [equal](https://github.com/Samweli/QGIS/blob/5b8759c5cd126b1da11996e19a706beff6d28ba3/src/providers/wms/qgswmsdataitems.h#L81) function from QgsLayerItem, to provide a further layer properties comparison ( leaving out path and name only comparison ), might need others efforts to increase the function layer properties coverage.

 I have also added a new class [QgsWMSLayerCollectionItem](https://github.com/Samweli/QGIS/blob/5b8759c5cd126b1da11996e19a706beff6d28ba3/src/providers/wms/qgswmsdataitems.h#L49) to handle the WMS Layers which enclose other layers. So we can handle them separately. I found it tricky to fix them especially in the #33621 case.

Below is a screenshot showing an example of how this function works in WMS/WMTS connections
OGC_refresh_on_action

### DIFF
--- a/src/providers/wms/qgswmsdataitems.cpp
+++ b/src/providers/wms/qgswmsdataitems.cpp
@@ -96,7 +96,7 @@ QVector<QgsDataItem *> QgsWMSConnectionItem::createChildren()
       // Attention, the name may be empty
       QgsDebugMsgLevel( QString::number( layerProperty.orderId ) + ' ' + layerProperty.name + ' ' + layerProperty.title, 2 );
       QString pathName = layerProperty.name.isEmpty() ? QString::number( layerProperty.orderId ) : layerProperty.name;
-      QgsDataItem *layer;
+      QgsDataItem *layer = nullptr;
 
       if ( layerProperty.name.isEmpty() )
         layer = new QgsWMSLayerCollectionItem( this, layerProperty.title, mPath + '/' + pathName, capabilitiesProperty, uri, layerProperty );
@@ -266,15 +266,13 @@ bool QgsWMSLayerCollectionItem::equal( const QgsDataItem *other )
   {
     return false;
   }
-  const QgsWMSLayerCollectionItem *otherCollectionItem = dynamic_cast<const QgsWMSLayerCollectionItem *>( other );
+  const QgsWMSLayerCollectionItem *otherCollectionItem = qobject_cast<const QgsWMSLayerCollectionItem *>( other );
   if ( !otherCollectionItem )
   {
     return false;
   }
 
   // Check if the children are not the same then they are not equal
-  if ( mChildren.isEmpty() & !otherCollectionItem->mChildren.isEmpty() )
-    return false;
   if ( mChildren.size() != otherCollectionItem->mChildren.size() )
     return false;
 
--- a/src/providers/wms/qgswmsdataitems.cpp
+++ b/src/providers/wms/qgswmsdataitems.cpp
@@ -103,7 +103,7 @@ QVector<QgsDataItem *> QgsWMSConnectionItem::createChildren()
       else
         layer = new QgsWMSLayerItem( this, layerProperty.title, mPath + '/' + pathName, capabilitiesProperty, uri, layerProperty );
 
-      children << layer ;
+      children.append( layer );
     }
   }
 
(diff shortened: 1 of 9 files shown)


src/providers/wms/qgswmsdataitems.cpp
```diff
@@ -233,31 +237,87 @@ bool QgsWMSConnectionItem::equal( const QgsDataItem *other )
...
+  setState( Populated );
+}
+
+bool QgsWMSLayerCollectionItem::equal( const QgsDataItem *other )
+{
+  if ( type() != other->type() )
+  {
+    return false;
+  }
+  const QgsWMSLayerCollectionItem *otherCollectionItem = dynamic_cast<const QgsWMSLayerCollectionItem *>( other );
```
nyalldawson:
```suggestion
  const QgsWMSLayerCollectionItem *otherCollectionItem = qobject_cast<const QgsWMSLayerCollectionItem *>( other );
```

src/providers/wms/qgswmsdataitems.cpp
```diff
@@ -233,31 +237,87 @@ bool QgsWMSConnectionItem::equal( const QgsDataItem *other )
...
+  if ( !otherCollectionItem )
+  {
+    return false;
+  }
+
+  // Check if the children are not the same then they are not equal
+  if ( mChildren.isEmpty() & !otherCollectionItem->mChildren.isEmpty() )
```
nyalldawson: This is redundant with the size check below

src/providers/wms/qgswmsdataitems.cpp
Samweli: Right, does this also apply here?https://github.com/qgis/QGIS/blob/3bdb75bda4e61dd397b0cc553baa67035fd1a25a/src/providers/wms/qgswmsdataitems.cpp#L376

nyalldawson: Yep. qobject_cast is optimised and should be used whenever casting qobject subclasses. See https://stackoverflow.com/a/43994766/1861260

Samweli: Ooh I see

nyalldawson:
This looks good to me -- is it ready for merge in your opinion @Samweli ?

nyalldawson (approved):

Samweli:
> This looks good to me -- is it ready for merge in your opinion @Samweli ?

I'm double checking its functionality, might need to improve layer comparison check, will let you know when I feel that is ready.

Samweli:
@nyalldawson that should do it, it is now ready for merge. Thanks a lot for the review.

src/providers/wms/qgswmsdataitems.cpp
```diff
@@ -226,13 +226,44 @@ bool QgsWMSConnectionItem::equal( const QgsDataItem *other )
...
   {
     return false;
   }
 
-  return ( mPath == o->mPath && mName == o->mName );
+  bool samePathAndName = ( mPath == otherConnectionItem->mPath && mName == otherConnectionItem->mName );
+
+  if ( samePathAndName )
+  {
+    // Check if the children are not the same then they are not equal
```
nyalldawson: When would this be the case?

Samweli: The current implementation has no guarantee that the wms connection path will be unique, when user edit a wms connection params eg URL only,  the connection path and name does not change but there is a possibility that children items (wms collection items) might change. 

That is why I added this check to confirm it is the same connection item, if not it should be updated.

Samweli: what do you think about @nyalldawson?

nyalldawson:
Forgot to hit merge on this yesterday, sorry!

Samweli:
> Forgot to hit merge on this yesterday, sorry!

No worries, thanks.

timlinux:
Thanks @nyalldawson

PR #33653, opened 2020-01-07
GPS - fix crashes in Python code, fix serial port connection issues

nyalldawson:
ping @NathanW2 -- as discussed yesterday.

Note that I can't test this as I don't have a serial based GPS connection device -- can you please test this and confirmation it fixes your issue?


NathanW2:
@nyalldawson  thanks very much.  I was about to do the same this morning but you beat me to it.

NathanW2:
Early testing on my laptop seem to show it's good.  Let me get remote access to the site with the GPS issue so I can test there.  Thanks again for the quick patch. Sorry for the slow testing.

NathanW2:
@nyalldawson thanks the sip transfer stuff does fix that side of the issue.  Will have a confirm on the settings override tomorrow but this should be safe to merge before that if you are happy with the code.

NathanW2:
**Note:** I also cherry-picked these over into 3.10 so it's working in there as well.

NathanW2:
+1 this looks good thanks.

PR #3617, opened 2016-10-17
[Feature] Rework snapping settings

m-kuhn:
Exposes snapping settings in a new toolbar.
This allows to quickly switch between different snapping modes / settings.
Snapping settings have also been put into a separate object so they can easily be saved / restored during runtime.

This also enables a requirement on [C++ 11 and new-style signal-slot](https://github.com/qgis/qgis3.0_api/issues/35#issuecomment-253805423) this part can be removed if someone asks for it :)
# Toolbar

snapping
# Configuration dialog

screenshot from 2016-10-18 10-06-07
# Snapping settings menu

Has been moved to the Project menu
# TODO
- [x] initial size of layer column + un-collapse tree when first shown
- [x] window has too much padding when served as a dialog
- [x] remove close button
- [ ] move tracing button to snapping toolbar
- [x] remove option to show widget in status bar
- [ ] rework icons
- [ ] button to show legend tree in status bar when in advanced mode

PR #33674, opened 2020-01-09
Add setting for format to show angular bearings to projects

nyalldawson:
The Settings - Options - Map Tools tab contains a new setting for controlling the default format to use for displaying angular bearings for newly created projects. Whenever a new project is created, it will
inherit this default settings.

The Project Properties dialog also has a new setting for the project-specific bearing format.

The intention is that whenever angular bearings are shown in QGIS, they will be formatted using the current project's bearing format settings. In this PR I've done this for the status bar pan direction message.

Also includes lots of nice API additions providing a stable, easy to discover place to set and retrieve settings like the bearing format.

Peek 2020-01-09 11-08


src/core/qgslocaldefaultsettings.h
```diff
@@ -0,0 +1,62 @@
...
+{
+
+  public:
+
+    /**
+     * Sets the default bearing \a format, which controls how angular bearings are displayed.
+     *
+     * \see bearingFormat()
+     */
+    static void setBearingFormat( const QgsBearingNumericFormat *format );
```
nirvn: Maybe at a node on ownership of format not being transferred here?

src/core/qgsproject.h
```diff
@@ -580,6 +581,21 @@ class CORE_EXPORT QgsProject : public QObject, public QgsExpressionContextGenera
...
+     * \since QGIS 3.12
+     */
+    const QgsProjectDisplaySettings *displaySettings() const SIP_SKIP;
+
+    /**
+     * Returns the project's display settings, which settings and properties relating
+     * to how a QgsProject should display values such as map coordinates and bearings.
+     * \since QGIS 3.12
+     */
+    QgsProjectDisplaySettings *displaySettings();
```
nirvn: Any philosophical reason why you've decided not to add a setter here?

src/core/qgslocaldefaultsettings.h
nyalldawson: This has been discussed in the past, with the consensus that NOT transferring ownership is the default behavior and shouldn't be documented

nirvn: @nyalldawson , TIL :smile:

src/core/qgsproject.h
nyalldawson: Yes -- partly because it's a QObject (and there may be connections setup to the existing settings object, so deleting it would break those), and also because it's following the pattern used elsewhere in QgsProject (e.g. layoutManager, annotationManager, etc)

nirvn: @nyalldawson , all good then. Hit that merge button! :)

3nids: I guess this discussion arose already in the past, but we're mixing a lot the 2 things in QGIS (modifiable QObject vs getter/setter of non-QObject based config).
ping @m-kuhn 
Wouldn't it be nice to come to a consensus and stick to it?
Or are there cases where one is better than the other?

nyalldawson: I think in this case:

1. We definitely want QgsProjectDisplaySettings to be a QObject on its own -- otherwise the signals from it would need to be moved to the already very complex QgsProject class. (The intention with classes like this one and QgsProjectViewSettings was to start breaking down the QgsProject behemoth into more maintainable smaller classes to avoid QgsProject becoming a dumping ground for everything.)

2. If QgsProjectDisplaySettings is a QObject, then we expect things to connect to its signals (in this case, QgisApp connects to it). If we make a setter for this and require api users to create a new QgsProjectDisplaySettings instance and set it on the project, any existing connections would be lost. We'd then need to add a new signal "projectDisplaySettingsChanged" to QgsProject. Instead of having granular signals in QgsProjectDisplaySettings, clients could only connect to this one all-encompassing signal and respond to it even when the actual setting they care about hasn't been the one which was changed.

3nids: This makes a lot of sense, and I recall thinking it was the way to go.
Trying to recall discussions on this, the snapping config came to my mind:
https://github.com/qgis/QGIS/pull/3617#issuecomment-254414285

My point here is really not towards your work, but rather to decide which is the way to go for the standard cases. And to me the snapping config is a really good comparison to this.

PR #33683, opened 2020-01-09
Oracle field generated always as identity

### LINKED ISSUE #33681: Oracle, impossible to add feature with an identity field
**Describe the bug**
In Oracle, there is a type of field that generated values as serial type in postgresql. This type is called identity and values are generated by default or always.

It's not possible to add feature with digitize maptool nor table attribute when the identity field is always generated because oracle does not allow a insert request where a value (even the default value with the sequential) is given for this field.

oracle_always_identity

For example in the image above, a table with a `pk` field (identity generated always) will not allow : 

`INSERT INTO test( "pk", "cnt", "name", "name2", "num_char", "geom" ) VALUES ( ?, ?, ?, ?, ?, ? )`

the request should be : 

`INSERT INTO test("cnt", "name", "name2", "num_char", "geom") VALUES ( ?, ?, ?, ?, ? )`

without the `pk` field.

Futhermore, default value like "ISEQ$$_75967".nextval (sequential for identity field) is not interpreted with the digitize maptool for both mode : 

oracle_always_identity2

when the identity value is generated by default and not always, it is possible to customize the value to do it work.

**How to Reproduce**

1. create a table with a identity field : 
`CREATE TABLE QGIS.POINT_DATA_IDENTITY ( "pk" NUMBER GENERATED ALWAYS AS IDENTITY PRIMARY KEY, GEOM SDO_GEOMETRY);`
2. add the layer to your qgis project
2. Try to add a new feature with digitize maptool or attribute table
3. It's not possible to commit the result

**QGIS and OS versions**

QGIS version | 3.10.1-A Coruña | QGIS code revision | 1d221185e1
-- | -- | -- | --
OS Version | Ubuntu 19.04 | This copy of QGIS writes debugging output.
Active python plugins | thyrsis

**Additional context**

SebastienPeillet:
## Description
Fix #33681 

It add functions (defaultValueClause and skipConstraintCheck) to evaluate default value for identity field (like "ISEQ$$_75967".nextval).

And in addFeatures function, always generated identity field are listed to be avoid in the insert request.

I add a test to check if the commit passes with an always generated identity field. 

## Checklist

### DIFF
--- a/src/providers/oracle/qgsoracleprovider.cpp
+++ b/src/providers/oracle/qgsoracleprovider.cpp
@@ -139,6 +139,14 @@ QgsOracleProvider::QgsOracleProvider( QString const &uri, const ProviderOptions 
 
   mLayerExtent.setMinimal();
 
+  // get always generated key
+  if ( !determineAlwaysGeneratedKeys() )
+  {
+    mValid = false;
+    disconnectDb();
+    return;
+  }
+
   // set the primary key
   if ( !determinePrimaryKey() )
   {
@@ -1010,6 +1018,46 @@ bool QgsOracleProvider::determinePrimaryKey()
   return mValid;
 }
 
+bool QgsOracleProvider::determineAlwaysGeneratedKeys()
+{
+  if ( !loadFields() )
+  {
+    return false;
+  }
+
+  // check to see if there is an unique index on the relation, which
+  // can be used as a key into the table. Primary keys are always
+  // unique indices, so we catch them as well.
+  QgsOracleConn *conn = connectionRO();
+  QSqlQuery qry( *conn );
+
+  QString sql = QStringLiteral( "SELECT a.column_name "
+                                "FROM all_tab_identity_cols a "
+                                "WHERE a.owner = '%1' "
+                                "AND a.table_name = '%2' "
+                                "AND a.generation_type = 'ALWAYS'" ).arg( mOwnerName ).arg( mTableName );
+
+  if ( exec( qry, sql, QVariantList() ) )
+  {
+    while ( qry.next() )
+    {
+      if ( mAttributeFields.names().contains( qry.value( 0 ).toString() ) )
+      {
+        mAlwaysGeneratedKeyAttrs.append( mAttributeFields.indexOf( qry.value( 0 ).toString() ) );
+      }
+    }
+    mValid = true;
+  }
+  else
+  {
+    QgsMessageLog::logMessage( tr( "Unable to execute the query.\nThe error message from the database was:\n%1.\nSQL: %2" )
+                               .arg( qry.lastError().text() )
+                               .arg( qry.lastQuery() ), tr( "Oracle" ) );
+    mValid = false;
+  }
+  return mValid;
+}
+
 bool QgsOracleProvider::uniqueData( QString query, QString colName )
 {
   Q_UNUSED( query )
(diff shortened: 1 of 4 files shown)


SebastienPeillet:
@troopa81 ping

src/providers/oracle/qgsoracleprovider.cpp
```diff
@@ -1184,7 +1184,43 @@ bool QgsOracleProvider::isValid() const
...
+  Q_UNUSED( constraint );
+  if ( providerProperty( EvaluateDefaultValues, false ).toBool() )
+  {
+    return !mDefaultValues.value( fieldIndex ).toString().isEmpty();
+  }
+  else
+  {
+    // stricter check - if we are evaluating default values only on commit then we can only bypass the check
+    // if the attribute values matches the original default clause
+    return mDefaultValues.value( fieldIndex ) == value.toString() && !value.isNull();
```
troopa81: Should be better to ckeck in fieldIndex is not outside `mDefaultValues` bounds

src/providers/oracle/qgsoracleprovider.cpp
```diff
@@ -1246,13 +1282,31 @@ bool QgsOracleProvider::addFeatures( QgsFeatureList &flist, QgsFeatureSink::Flag
...
       delim = ',';
     }
+    QString sql = QStringLiteral( "SELECT a.column_name "
+                                  "FROM all_tab_identity_cols a "
+                                  "WHERE a.owner = '%1' "
+                                  "AND a.table_name = '%2' "
+                                  "AND a.generation_type = 'ALWAYS'" ).arg( mOwnerName ).arg( mTableName );
+    identitytype.prepare( sql );
+
+    if ( identitytype.exec() )
```
troopa81: Is there a reason for not using `QgsOracleProvider::exec` method ? and maybe print an error message if the request failed.

tests/testdata/provider/testdata_oracle.sql
```diff
@@ -65,5 +65,16 @@ CREATE TABLE QGIS.DATE_TIMES ( "id" INTEGER PRIMARY KEY, "date_field" DATE, "dat
 
 INSERT INTO QGIS.DATE_TIMES ("id", "date_field", "datetime_field" ) VALUES (1, DATE '2004-03-04', TIMESTAMP '2004-03-04 13:41:52' );
 
+CREATE TABLE QGIS.POINT_DATA_IDENTITY ( "pk" NUMBER GENERATED ALWAYS AS IDENTITY PRIMARY KEY, GEOM SDO_GEOMETRY);
+INSERT INTO QGIS.POINT_DATA_IDENTITY (GEOM)
+ SELECT SDO_GEOMETRY( 2001,4326,SDO_POINT_TYPE(1, 2, NULL), NULL, NULL) from dual
```
troopa81: Is there a specific reason to add so much elements. It's no big deal anyway, I'm just wondering

troopa81 (commented):
Looks good to me, just a few comments to deal with

tests/testdata/provider/testdata_oracle.sql
SebastienPeillet: No, it's true, one is enough. I changed it !

src/providers/oracle/qgsoracleprovider.cpp
SebastienPeillet: No, I didn't pay attention to it. I changed this and add a else condition.

src/providers/oracle/qgsoracleprovider.cpp
SebastienPeillet: Done ! ;)

troopa81: the way you fix it, you didn't check if fieldIndex is > 0 and < mDefaultAttributes.size() but you check if fieldIndex exist in list which is not what we are looking for.

Actually, the value method return a default contructed value (QVariant here) in case fieldIndex is out of bounds so your first version was OK. You can revert your change on this one, sorry for the noise.

troopa81 (approved):
looks ok for me

src/providers/oracle/qgsoracleprovider.cpp
```diff
@@ -1246,13 +1282,35 @@ bool QgsOracleProvider::addFeatures( QgsFeatureList &flist, QgsFeatureSink::Flag
...
     QList<int> fieldId;
+    QList<int> alwaysGenerated;
 
     if ( !mGeometryColumn.isNull() )
     {
       insert += quotedIdentifier( mGeometryColumn );
       values += '?';
       delim = ',';
     }
+    QString sql = QStringLiteral( "SELECT a.column_name "
```
nyalldawson: This looks very expensive to calculate on every feature insert. I think it should be retrieved once at provider initialization and remembered

SebastienPeillet: Indeed, good idea, it could be a class attribute ! Thanks, I will do that

nyalldawson:
@troopa81 are you happy to the revised version?

troopa81:
@nyalldawson yes I am, looks ok for me

PR #33685, opened 2020-01-09
Native PostGIS raster data provider

elpaso:
## Description

This is an implementation of a PostGIS raster data provider in QGIS core.
Tiles are cached in RAM memory.

## TODO

- [x] Add support for WHERE conditions 
- [ ] ~and query layers~ (not a priority)
- [x] Add support for composite PKs (low priority)
- [x] Add support for tile PKs different than integer
- [x] Show more information in metadata properties
- [ ] ~Support endiannes when parsing WKB~ (not necessay, because PG always returns littleendian)
- [x] More tests

**Funded by: Christmas Holidays Inc.**


gioman:
Thanks Alessandro and Santa...

rouault:
Why not improving the GDAL PostgisRaster driver ?

elpaso:
> Why not improving the GDAL PostgisRaster driver ?

I tried multiple times to debug the issues without success and asked for help without receiving any: https://github.com/qgis/QGIS/issues/30392#issuecomment-561685894

Also, I'm familiar with QGIS API and not with GDAL.

There are other advantages in having it in QGIS core:
- better integration with the auth system
- re-use current low level PG connection pool implementation

rouault:
> without receiving any

Ah don't rely at pinging me like that. I might easily miss them. It is just lost in the noise of all QGIS notifications. I can't unfortunately read everything.

> better integration with the auth system

I guess some callback mechanism in GDAL could do the job

> re-use current low level PG connection pool implementation

As far as I remember the GDAL PG driver has a connection pool. See PostGISRasterDriver::GetConnection()

elpaso:
Feel free to fix the GDAL issues, complete the integration with the auth system and close this PR, I've learned some interesting things along the way and this was all volunteer work so I'll be happy whatever it is the outcome.

rouault:
> Feel free to fix the GDAL issues

I think I do more than my share on this :-) I'll not -1 this PR for sure. Just a bit sad seeing what I perceive a bit like effort duplication, but I can understand your perspective of feeling less familiar with GDAL code

elpaso:
@rouault I know you are doing a lot and I fully understand your POV.

All my initial efforts were oriented to identify the source of the issue and possibly try to fix it (in GDAL). I've spent several hours to get a full build toolchain with GDAL/PROJ from sources but when it came to debug GDAL internals and memory corruption in a multithreaded scenario I admin I was a bit lost.

rouault:
There might be some connection with https://github.com/qgis/QGIS/commit/9da842d48881fe54af734f033be91c2772fe1177 . Basically the strategy for the GDAL Postgisraster connection pool is to have a key where the id of the thread at creation time of the GDAL dataset is taken into account. Issues might arise if 2 datasets are created in the same thread, since they will share the same PG connection, but one is migrated to another thread afterwards. Above referenced commit is supposed to fix that (by using the same GDAL dataset, and a mutex around all calls)

elpaso:
I've tested this PR and it looks robust to me, I would have preferred a review but I'm merging it anyway because it fixes current issues with GDAL approach, we can revert if needed.

src/providers/postgres/raster/qgspostgresrasterprovider.cpp
```diff
@@ -0,0 +1,1600 @@
...
+  , mRequestedSrid( other.mRequestedSrid )
+  , mConnectionRO( other.mConnectionRO )
+  , mConnectionRW( other.mConnectionRW )
+  , mPrimaryKeyType( other.mPrimaryKeyType )
+  , mPrimaryKeyAttrs( other.mPrimaryKeyAttrs )
+  , mShared( other.mShared )
+{
+}
+
+QgsPostgresRasterProvider::~QgsPostgresRasterProvider()
```
nyalldawson: = default

src/providers/postgres/raster/qgspostgresrasterprovider.cpp
```diff
@@ -0,0 +1,1600 @@
...
+{
+
+}
+
+bool QgsPostgresRasterProvider::hasSufficientPermsAndCapabilities()
+{
+  QgsDebugMsg( QStringLiteral( "Checking for permissions on the relation" ) );
```
nyalldawson: QgsDebugMsgLevel( ...)

src/providers/postgres/raster/qgspostgresrasterprovider.cpp
```diff
@@ -0,0 +1,1600 @@
...
+
+      QgsPostgresResult result( connectionRO()->PQexec( sql ) );
+
+      if ( ( PGRES_TUPLES_OK == result.PQresultStatus() ) && ( result.PQntuples() > 0 ) )
+      {
+        mRasterColumn = result.PQgetvalue( 0, 0 );
+        mHasSpatialIndex = result.PQgetvalue( 0, 9 ) == 't';
+        bool ok;
+
+        mCrs = QgsCoordinateReferenceSystem::fromEpsgId( result.PQgetvalue( 0, 1 ).toLong( &ok ) );
```
nyalldawson: I think this should be createFromSrid. But keep in mind that postgis crs handling is rather broken in QGIS and needs to be rethought for 4.0 (we should be retrieving the definition from the server corresponding to the reported postgis srs, and not using any qgis internal databases for this ... which may or may not have any match to the server's definitions.)

src/providers/postgres/raster/qgspostgresrasterprovider.cpp
```diff
@@ -0,0 +1,1600 @@
...
+                                 QStringLiteral( "PostGIS" ), Qgis::Critical );
+      return false;
+    }
+
+    // Compute raster size
+    mHeight = static_cast<long>( mExtent.height() / std::abs( mScaleY ) );
+    mWidth = static_cast<long>( mExtent.width() / std::abs( mScaleX ) );
+    mIsTiled = ( mWidth != mTileWidth ) || ( mHeight != mTileHeight );
+
+    mCrs = QgsCoordinateReferenceSystem::fromEpsgId( result.PQgetvalue( 0, 9 ).toLong( &ok ) );
```
nyalldawson: again, probably createFromSrid

src/providers/postgres/raster/qgspostgresrasterprovider.cpp
```diff
@@ -0,0 +1,1600 @@
...
+  if ( mIsQuery || !connectionRO() )
+    return QgsPostgresProvider::Relkind::Unknown;
+
+  QString sql = QStringLiteral( "SELECT relkind FROM pg_class WHERE oid=regclass(%1)::oid" ).arg( quotedValue( mQuery ) );
+  QgsPostgresResult res( connectionRO()->PQexec( sql ) );
+  QString type = res.PQgetvalue( 0, 0 );
+
+  QgsPostgresProvider::Relkind kind = QgsPostgresProvider::Relkind::Unknown;
+
+  if ( type == QLatin1String( "r" ) )
```
nyalldawson: type == 'r', etc

src/providers/postgres/raster/qgspostgresrasterprovider.cpp
```diff
@@ -0,0 +1,1600 @@
...
+
+    res = connectionRO()->PQexec( sql );
+    QgsDebugMsg( QStringLiteral( "Got %1 rows." ).arg( res.PQntuples() ) );
+
+    QStringList log;
+
+    // no primary or unique indices found
+    if ( res.PQntuples() == 0 )
+    {
+      QgsDebugMsg( QStringLiteral( "Relation has no primary key -- investigating alternatives" ) );
```
nyalldawson: QgsDebugMsgLevel (QgsDebugMsg should only be used for errors)

src/providers/postgres/raster/qgspostgresrasterprovider.cpp
```diff
@@ -0,0 +1,1600 @@
...
+      // Two options here. If the relation is a table, see if there is
+      // an oid column that can be used instead.
+      // If the relation is a view try to find a suitable column to use as
+      // the primary key.
+
+      const QgsPostgresProvider::Relkind type = relkind();
+
+      if ( type == QgsPostgresProvider::Relkind::OrdinaryTable || type == QgsPostgresProvider::Relkind::PartitionedTable )
+      {
+        QgsDebugMsg( QStringLiteral( "Relation is a table. Checking to see if it has an oid column." ) );
```
nyalldawson: QgsDebugMsgLevel

src/providers/postgres/raster/qgspostgresrasterprovider.cpp
```diff
@@ -0,0 +1,1600 @@
...
+  else
+  {
+    QgsMessageLog::logMessage( tr( "Error fetching statistics for %1: %2\nSQL: %3" )
+                               .arg( mQuery )
+                               .arg( result.PQresultErrorMessage() )
+                               .arg( sql ),
+                               QStringLiteral( "PostGIS" ), Qgis::Warning );
+  }
+
+  QgsDebugMsg( QStringLiteral( "************ STATS **************" ) );
```
nyalldawson: Level

src/providers/postgres/raster/qgspostgresrasterprovider.h
```diff
@@ -0,0 +1,207 @@
...
+    int mTileHeight = 0;
+    //! Scale x
+    double mScaleX = 0;
+    //! Scale y
+    double mScaleY = 0;
+
+    QString mDetectedSrid;            //!< Spatial reference detected in the database
+    QString mRequestedSrid;           //!< Spatial reference requested in the uri
+    QgsPostgresConn *mConnectionRO = nullptr ; //!< Read-only database connection (initially)
+    QgsPostgresConn *mConnectionRW = nullptr ; //!< Read-write database connection (on update)
```
nyalldawson: Do we need this? It's all read-only, right?

src/providers/postgres/raster/qgspostgresrastershareddata.cpp
```diff
@@ -0,0 +1,397 @@
...
+}
+
+QgsPostgresRasterSharedData::TilesResponse QgsPostgresRasterSharedData::tiles( const QgsPostgresRasterSharedData::TilesRequest &request )
+{
+  QMutexLocker locker( &mMutex );
+
+  QgsPostgresRasterSharedData::TilesResponse result;
+
+  // First check for index existence
+  if ( mSpatialIndexes.find( request.overviewFactor ) == mSpatialIndexes.end() )
```
nyalldawson: constFind?

nyalldawson (commented):
A belated review! :tongue:

src/providers/postgres/raster/qgspostgresrasterprovider.h
elpaso: For now, yes. But I think editing will be possible in the future, I'd keep it even if it's currently unused.

elpaso:
@nyalldawson thank you for the review, I'll make the changes.

src/providers/postgres/raster/qgspostgresrastershareddata.cpp
elpaso: there is no such method in std::map

jgrocha:
@elpaso I've just compiled QGIS this Postgis raster support, but I'm not able to add Postgis rasters anymore (using the browser or DB Manager). I get the following error:
```
src/core/qgsmessagelog.cpp:29 : (logMessage) [0ms] [thread:0x561a671a9a60] 2020-01-25T17:56:10 PostGIS[2] Unsupported pixel type: 0
```
I saw your tests (great work, by the way) and everythng seems fine to me.

My rasters were added with `raster2pgsql`. Is it possible that the raster2pgsql encoding is different? I really don't know.

I'll investigate this issue further and I'll report back.

elpaso:
I'm happy to test your case. Just attach a DB dump.

jgrocha:
@elpaso I've created a small database with two tables, for the same image (with and without overviews). Both are stored in db. I've also tested the same image out-of-db.

The commends to load the raters were:
```bash
raster2pgsql -s 3763 -d -I -C -M `pwd`/carta_1990_ACEL.tif -F -t 1000x1000 public.carta_1900_gazeta | psql raster
```
```bash
raster2pgsql -s 3763 -d -I -C -M -l 4,8,16,32,64,128,256,512 `pwd`/carta_1990_ACEL.tif -F -t 1000x1000 public.carta_1900_gazeta_ovr | psql raster
```
The image is available at: http://webgis.di.uminho.pt/~jgr/QGIS/carta_1900_Gazeta.tif

The database backup is available at: http://webgis.di.uminho.pt/~jgr/QGIS/raster.backup

The database dump was create with:
```bash
pg_dump -Fc --file raster.backup raster
```

I'm really looking forward to see this working, but I'm missing something in my QGIS setup/compilation.

elpaso:
@jgrocha I'll have a look now, do you mind opening a bur report for the record?

elpaso:
@jgrocha nevermind, I've found the issue.

Issue #33641, opened 2020-01-07
Temporary files (processing output) are not discarded after closing QGIS.


pathmapper:
**Describe the bug**
On Windows, temporary files (processing outputs when using tools from the processing toolbox in QGIS) are not discarded after closing QGIS.

**How to Reproduce**

1. Use the GDAL tool _Clip Vector by Extent_ from the processing toolbox on a vector layer and save the output to a temporary file like on the following screenshot:

![Unbenannt](https://user-images.githubusercontent.com/20856381/71899626-6cc7b100-315c-11ea-9b51-aa4d201499b9.PNG)

2. Copy the specified output path

3. Close QGIS

4. Browse to the specified output path -> the temp files are still there

**QGIS and OS versions**

QGIS version | 3.10.1-A Coruña | QGIS code revision | ef24c526da
-- | -- | -- | --
Compiled against Qt | 5.11.2 | Running against Qt | 5.11.2
Compiled against GDAL/OGR | 3.0.2 | Running against GDAL/OGR | 3.0.2
Compiled against GEOS | 3.8.0-CAPI-1.13.1 | Running against GEOS | 3.8.0-CAPI-1.13.1
Compiled against SQLite | 3.29.0 | Running against SQLite | 3.29.0
PostgreSQL Client Version | 11.5 | SpatiaLite Version | 4.3.0
QWT Version | 6.1.3 | QScintilla2 Version | 2.10.8
Compiled against PROJ | 6.2.1 | Running against PROJ | Rel. 6.2.1, November 1st, 2019
OS Version | Windows 10 (10.0)
Active python plugins | canvas_clipper; db_manager; MetaSearch; processing

**Additional context**

On Linux this issue doesn't exist.
Please note, that on Linux there is only one directory prefixed with `processing_` created while on Windows there is a `processing_` directory in another `processing_` directory created.

For comparision here the output path on Linux:
![3](https://user-images.githubusercontent.com/20856381/71901658-b3b7a580-3160-11ea-9442-2d9422ed9ff4.PNG)

Maybe https://github.com/qgis/QGIS/issues/28222 is related.

Edit:
This issue exists also when using `QgsProcessing.TEMPORARY_OUTPUT` in PyQGIS ([context](https://gis.stackexchange.com/questions/346646/pyqgis-save-processing-output-to-temporary-file-which-is-discarded-when-qgis-i#comment566786_346688)).

alexbruy:
Partially related to changes from #33086.

[...]

PR #33686, opened 2020-01-09
[processing] use system temporary directory as default temp dir for Processing outputs

alexbruy:
## Description
#33086 has added support for configurable temporary directory for Processing outputs. As default option it uses directory, generated by Processing, e.g. `/tmp/processing_601abb2fa9664e5eab9a1ea02370fda5`. IMHO better to use system temporary directory as top-level location and create Processing folders inside it. Also current approach leads to creating extra nested directory once user have saved any change in Processing settings and path looks like `/tmp/processing_601abb2fa9664e5eab9a1ea02370fda5/processing_ee8d7407ab684e66bfd7d1555bd3db49` (see #33641).

## Checklist

### DIFF
--- a/python/plugins/processing/core/ProcessingConfig.py
+++ b/python/plugins/processing/core/ProcessingConfig.py
@@ -22,6 +22,7 @@ __date__ = 'August 2012'
 __copyright__ = '(C) 2012, Victor Olaya'
 
 import os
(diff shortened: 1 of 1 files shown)


luipir (approved):
good catch

PR #33688, opened 2020-01-09
Fix invalid attributes dialog on copy to another layer

signedav:
It's possible to copy features from one layer to another.
If there are the same fields in the destination layer, then the attributes for them are taken from the original feature. If not, the default value is taken. Otherwise the new attribute is null.

If the destination layer has constraints on the fields, these should be fulfilled now or disregarded on purpose. But not just copied invalid like it used to do. 

That's why now the attributes are checked against the constraints. And for all the invalid features a dialog pops up. 
copy-dialog_2

And on pasting only one feature, the options are reduced:
copy_paste_single

Maybe there could be added in future a multi edit function.


src/app/qgisapp.cpp
```diff
@@ -9588,6 +9589,47 @@ void QgisApp::pasteFromClipboard( QgsMapLayer *destinationLayer )
...
+  QgsFeatureList invalidFeatures;
+  for ( const QgsFeature &f : qgis::as_const( validFeatures ) )
+  {
+    for ( int idx = 0; idx < pasteVectorLayer->fields().count(); ++idx )
+    {
+      QStringList errors;
+      if ( !QgsVectorLayerUtils::validateAttribute( pasteVectorLayer, f, idx, errors, QgsFieldConstraints::ConstraintStrengthHard, QgsFieldConstraints::ConstraintOriginNotSet ) )
+      {
+        invalidFeatures << f;
+        validFeatures.removeOne( f );
```
nyalldawson: You can't modify a qlist while iterating over it like this. You'd need to use a modifiable iterator instead and .erase the current entry.

src/app/qgisapp.cpp
```diff
@@ -9588,6 +9589,47 @@ void QgisApp::pasteFromClipboard( QgsMapLayer *destinationLayer )
   // now create new feature using pasted feature as a template. This automatically handles default
   // values and field constraints
   QgsFeatureList newFeatures {QgsVectorLayerUtils::createFeatures( pasteVectorLayer, newFeaturesDataList, &context )};
+
+  // check constraints
+  QgsFeatureList validFeatures = newFeatures;
+  QgsFeatureList invalidFeatures;
+  for ( const QgsFeature &f : qgis::as_const( validFeatures ) )
```
nyalldawson: Could we skip this whole loop if no constraints are present? I'm concerned about a regression in paste time for huge numbers of features.

src/app/qgisapp.cpp
```diff
@@ -9588,6 +9589,47 @@ void QgisApp::pasteFromClipboard( QgsMapLayer *destinationLayer )
...
+  }
+
+  if ( !invalidFeatures.isEmpty() )
+  {
+    newFeatures.clear();
+
+    QgsFixAttributeDialog *dialog = new QgsFixAttributeDialog( pasteVectorLayer, invalidFeatures, this );
+    int feedback = dialog->exec();
+
+    if ( feedback == QgsFixAttributeDialog::VanishAll )
```
nyalldawson: Switch instead of if

src/gui/qgsfixattributedialog.cpp
```diff
@@ -0,0 +1,134 @@
...
+  : QDialog( parent )
+  , mFeatures( features )
+{
+  init( vl );
+}
+
+void QgsFixAttributeDialog::init( QgsVectorLayer *layer )
+{
+  QgsAttributeEditorContext context;
+  setWindowTitle( tr( "%1 - Fix Pastet Features" ).arg( layer->name() ) );
```
nyalldawson: "pasted"

src/gui/qgsfixattributedialog.cpp
```diff
@@ -0,0 +1,134 @@
...
+  mProgressBar->setVisible( mFeatures.count() > 1 );
+  infoLayout->addWidget( mProgressBar );
+  QgsFeature feature;
+  mAttributeForm = new QgsAttributeForm( layer, *mCurrentFeature, context, this );
+  mAttributeForm->setMode( QgsAttributeEditorContext::SingleEditMode );
+  mAttributeForm->disconnectButtonBox();
+  layout()->addWidget( mAttributeForm );
+
+  QDialogButtonBox *buttonBox = mAttributeForm->findChild<QDialogButtonBox *>();
+  QPushButton *cancelAllBtn = new QPushButton( tr( "Cancel all" ) );
```
nyalldawson: Fix capitalisation for buttons (should be title case)

nyalldawson: Actually these strings are a bit confusing. I'd go with "discard all",  "discard all invalid features" and "paste all (including invalid)".

nyalldawson: @nirvn these buttons are very long - got any ideas on alternative ui?

src/gui/qgsfixattributedialog.h
```diff
@@ -0,0 +1,87 @@
...
+#include "qgis_gui.h"
+
+/**
+ * \ingroup gui
+ * \class QgsFixAttributeDialog
+ * \brief Dialog to fix a list of invalid feature regarding constraints
+ * \since 3.12
+ */
+
+class GUI_EXPORT QgsFixAttributeDialog : public QDialog
```
nyalldawson: Maybe leave in app for now?

src/gui/qgsfixattributedialog.h
```diff
@@ -0,0 +1,87 @@
...
+    Q_OBJECT
+
+  public:
+
+    /**
+     * Feedback code on closing the dialog
+     */
+    enum Feedback
+    {
+      VanishAll,  //!< Feedback to cancel copy of all features (even valid ones)
```
nyalldawson: DIscardAll

nyalldawson:
Looks good to me

signedav:
Thanks @nyalldawson

PR #33697, opened 2020-01-09
[processing] Improve temporary directory handling

nyalldawson:
- Make the new Temp Folder setting optional, default to an empty string, and note that empty = use default
- Move responsibility for cleanup of temporary folders to c++ (if Processing Python part crashes, or doesn't exist, then we still want these cleaned up correctly)

### DIFF
--- a/src/core/processing/qgsprocessingutils.cpp
+++ b/src/core/processing/qgsprocessingutils.cpp
@@ -746,33 +746,35 @@ QVariant QgsProcessingUtils::generateIteratingDestination( const QVariant &input
 
 QString QgsProcessingUtils::tempFolder()
 {
-  static std::unique_ptr< QTemporaryDir > sTempFolder;
+  // we maintain a list of temporary folders -- this allows us to append additional
+  // folders when a setting change causes the base temp folder to change, while deferring
+  // cleanup of ALL these temp folders until session end (we can't cleanup older folders immediately,
+  // because we don't know whether they have data in them which is still wanted)
+  static std::vector< std::unique_ptr< QTemporaryDir > > sTempFolders;
   static QString sFolder;
   static QMutex sMutex;
   QMutexLocker locker( &sMutex );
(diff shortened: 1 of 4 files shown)


nyalldawson:
@alexbruy what do you think?

alexbruy (commented):
Looks good. I was about to look into moving cleanup steps to core code too, but you are faster :-).
+1 to merge

alexbruy (approved):

nyalldawson:
@alexbruy @elpaso @nirvn @m-kuhn do you mind re-reviewing this update? This code makes me very nervous, as I'm concerned that a bad setting as the temp folder could potentially result in data being wiped (e.g. if a user set the path as their documents folder). This change **should** be safe, because we append the /processing_XXXXXX subpath to whatever the user sets, but I'd really value extra sets of eyes over the logic to make sure I haven't missed anything.

alexbruy:
Looks good to me, as far as I can see it should be safe.

src/core/processing/qgsprocessingutils.cpp
```diff
@@ -746,17 +746,34 @@ QVariant QgsProcessingUtils::generateIteratingDestination( const QVariant &input
...
+  }
+  else if ( sFolder.isEmpty() || !sFolder.startsWith( basePath ) || !sTempFolder )
+  {
+    if ( !QDir().exists( basePath ) )
+      QDir().mkpath( basePath );
+
+    const QString templatePath = QStringLiteral( "%1/processing_XXXXXX" ).arg( basePath );
+    // leak the previous folder -- we don't want it to be cleaned up, we don't know what was in it that may still
+    // be required for this session!
+    sTempFolder.release();
```
m-kuhn:
```suggestion
    sTempFolder.setAutoRemove( false );
```

If I read this correctly, this currently leaks memory next to a physical folder.

m-kuhn: Alternatively, keep a static list of previous temp folders somewhere to be destroyed on application exit.

nyalldawson: Good idea! Can you check the new version?

PR #33643, opened 2020-01-07
[GUI] fix identify panel size when long localized strings used (fix #29546)

alexbruy:
## Description
In some locales "Auto open form" text and some items in the "Mode" combobox in the identify panel translated as very long strings. As result panel takes too much screen space and can not be reduced.

Proposed PR replaces `QCheckBox` with `QCheckBox` and `QLabel` with enabled word-wrapping and changes width of the "Mode" combobox. Fixes #29546.

Before
no-wrap
After
wrap

## Checklist
- [x] Commit messages are descriptive and explain the rationale for changes
- [x] Commits which fix bugs include `Fixes #11111` at the bottom of the commit message
- [x] I have read the [QGIS Coding Standards](https://docs.qgis.org/testing/en/docs/developers_guide/codingstandards.html) and this PR complies with them
- [ ] New unit tests have been added for core changes
- [x] I have run [the `scripts/prepare-commit.sh` script](https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTING.md#contributing-to-qgis) before each commit
- [x] I have evaluated whether it is appropriate for this PR to be backported, backport requests are left as label or comment

PR #33704, opened 2020-01-10
[needs-docs][ui] Better placement for the method combobox and auto open action

nirvn:
## Description

@alexbruy , this is a follow up to your PR #33643. I noticed a few UI regressions after the commit was merged:
image
- the combo box text gets clipped even though there's space for it to display;
- there's a wider-than-normal space between the checkbox and its label, which if the above-mentioned combo box is expanded ends up being closer to the checkbox than the label is; and
- clicking on the label doesn't toggle the checkbox, breaking well established UI expectations.

This PR fixes the regressions by removing the spacer and allowing the combo box to expand to fill available space, and moves the auto open form checkbox to a settings menu accessible via the panel toolbar. The latter is a more disrupting change, but IMHO is the right thing to do. It gives us the space to have a more self-explanatory string, works for all locale, and essentially moves a setting that doesn't need to always be part of the panel into a pop up menu.

It looks like that now:
image

## Checklist


nirvn:
@alexbruy , let me know if you are fine with this alternative solution.

alexbruy:
@nirvn I'm fine with proposed solution.

Another option, if we want to keep checkbox on the panel, is to implement custom checkbox widget with word-wrapping. I was thinking about it and even made a prototype for it. But after looking in various dialogs seems there are no other places where it can be used, so I decide to took simpler path.

alexbruy (approved):

PR #33722, opened 2020-01-10
Respect OVERWRITE parameter in package layers algorithm

### LINKED ISSUE #33721: Overwrite parameter in package layers algorithm not functional
The package layers algorithm offers a parameter "Overwrite".

Regardless if it's true or not, the algorithm will overwrite any existing geopackage at the output location (on linux and mac at least).

Expected behavior IMO:

- if set to true, an existing file should be deleted and recreated / completely replaced
- if set to false, an existing file should be left in place and layers inside added

m-kuhn:
Change in behavior: if the overwrite is set to false and the target gpkg exists, open the gpkg and add new layers.

Old behavior: the target gpkg was recreated in any case

## Documentation changes

Being explicit about the parameter OVERWRITE:

- When it's set to true: the file will be deleted and recreated
- When it's set to false: layers will be added and existing layers left in place
- No guarantee is made with respect to new layers that have a collision with files that already exist inside the file

I would say this also needs to mention that this was fixed in version 3.10.2

------------------

Fix #33721

### DIFF
--- a/src/analysis/processing/qgsalgorithmpackage.cpp
+++ b/src/analysis/processing/qgsalgorithmpackage.cpp
@@ -106,7 +106,13 @@ QVariantMap QgsPackageAlgorithm::processAlgorithm( const QVariantMap &parameters
     throw QgsProcessingException( QObject::tr( "GeoPackage driver not found." ) );
   }
 
-  gdal::ogr_datasource_unique_ptr hDS( OGR_Dr_CreateDataSource( hGpkgDriver, packagePath.toUtf8().constData(), nullptr ) );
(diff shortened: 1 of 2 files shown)


nyalldawson (approved):

nirvn:
@m-kuhn , since the merging of this PR, travis azure has the qgis_processingalgstest test failing systematically. Related?

m-kuhn:
Possibly yes. I just cannot find a way to get any more details from the failing tests, neither via otb dash nor via the test runs itself or the the tests panel on azure.

PR #33725, opened 2020-01-10
[BUGFIX] Read SLD TextSymbolizer set units to pixels

rldhont:
## Description
The SLD unit is pixels so font size unit, buffer unit, offset unit and distance unit have to be set to render pixels unit.

## Checklist

### DIFF
--- a/python/core/auto_generated/symbology/qgssymbollayerutils.sip.in
+++ b/python/core/auto_generated/symbology/qgssymbollayerutils.sip.in
@@ -171,7 +171,7 @@ Encodes a render unit into an SLD unit of measure string.
 .. seealso:: :py:func:`decodeSldUom`
 %End
 
-    static QgsUnitTypes::RenderUnit decodeSldUom( const QString &str, double *scaleFactor );
+    static QgsUnitTypes::RenderUnit decodeSldUom( const QString &str, double *scaleFactor = 0 );
 %Docstring
 Decodes a SLD unit of measure string to a render unit.
 
--- a/src/core/qgsvectorlayer.cpp
+++ b/src/core/qgsvectorlayer.cpp
@@ -4616,7 +4616,7 @@ bool QgsVectorLayer::readSldTextSymbolizer( const QDomNode &node, QgsPalLayerSet
           if ( ok )
           {
             settings.xOffset = xOffset;
-            settings.distUnits = QgsUnitTypes::RenderPixels;
+            settings.offsetUnits = QgsUnitTypes::RenderPixels;
           }
         }
         QDomElement anchorPointYElem = anchorPointElem.firstChildElement( QStringLiteral( "AnchorPointY" ) );
@@ -4627,7 +4627,7 @@ bool QgsVectorLayer::readSldTextSymbolizer( const QDomNode &node, QgsPalLayerSet
           if ( ok )
           {
             settings.yOffset = yOffset;
-            settings.distUnits = QgsUnitTypes::RenderPixels;
+            settings.offsetUnits = QgsUnitTypes::RenderPixels;
           }
         }
       }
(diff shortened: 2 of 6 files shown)


luipir:
LGTM... BTW @rldhont did you check if during SLD output of the font, it is always converted to pixel units? (could be a possible bug)

luipir:
@rldhont waiting for test fix to accept the PR

rldhont:
@luipir I have check that every units are converted to pixels.

rldhont:
@luipir code fixed

nyalldawson:
One consideration here is that pixel unit sizes have very limited use in QGIS. If you set the font size and parameters to a pixel sized unit, then any outputs made from that map in hi resolutions (such as from a print layout) will have **tiny** sizes. So it depends what the intention is here... if it's for exact 1:1 match of the SLD then the change is good, but if it's to aid users with making usable maps from an SLD style, then you're much better doing a conversion to a mm based size!

rldhont:
The intention is to do the same as createSymbolizerFromSld functions, so it's for exact 1:1 match of SLD.

Some enhancements have to be done in the QGIS SLD readers:
* reorganize the API: functions are in QgsVectorLayer, QgsSymbolLayerUtils and QgsOgcUtils
* translate from pixel unit to mm ot pt

luipir (approved):

luipir:
@nyalldawson do you agree to merge postponing enhancement listed by @rldhont ?

luipir:
@aaime FYI in case affect your SLD based workflows (I suppose not)

nyalldawson:
I don't have a strong opinion either way about this - I just wanted to point out the inherent issue with pixel based units

aaime:
SLD indeed works in pixels by default. There are no other screen units, if one wants, it is possible to use ground units, meters or feet, but that has an explicit attribute at the symbolizer level (check the Symbology Encoding extension, look for the "uom" attribute).
Looks good to me.

luipir:
> but that has an explicit attribute at the symbolizer level

@rldhont so we can't assume pixel as default without checking the context!
e..g if no Unit is specified at symbolizer level => assume pixel as default

rldhont:
@aaime `QgsSymbolLayerUtils::decodeSldUom` is not used and tested.

```
QgsUnitTypes::RenderUnit QgsSymbolLayerUtils::decodeSldUom( const QString &str, double *scaleFactor )
{
  if ( str == QLatin1String( "http://www.opengeospatial.org/se/units/metre" ) )
  {
    if ( scaleFactor )
      *scaleFactor = 1000.0;  // from meters to millimeters
    return QgsUnitTypes::RenderMapUnits;
  }
  else if ( str == QLatin1String( "http://www.opengeospatial.org/se/units/foot" ) )
  {
    if ( scaleFactor )
      *scaleFactor = 304.8; // from feet to meters
    return QgsUnitTypes::RenderMapUnits;
  }

  // pixel is the SLD default uom. The "standardized rendering pixel
  // size" is defined to be 0.28mm x 0.28mm (millimeters).
  if ( scaleFactor )
    *scaleFactor = 1 / 0.00028; // from pixels to millimeters
  return QgsUnitTypes::RenderMillimeters;
}
```

What are the available `uom` values ?

aaime:
Quoting from the Symbology Encoding 1.1 specification:

> The following uom definitions are recommended to be used:
> ``<PolygonSymbolizer uom="http://www.opengeospatial.org/se/units/metre"/>``
> ``<PolygonSymbolizer uom="http://www.opengeospatial.org/se/units/foot"/>``
> ``<PolygonSymbolizer uom="http://www.opengeospatial.org/se/units/pixel"/>``

rldhont:
@aaime can you review the code ?

luipir:
@rldhont test seems does not cover in case of "uom" but yust default (pixels) https://github.com/qgis/QGIS/pull/33725/files#diff-106be592650b55ccae79c83eecd15cebR4477
Am I  right?

luipir (approved):

PR #33734, opened 2020-01-11
New layout item type: manually created fixed tables

nyalldawson:
This new item type allows for creation of tables with contents manually entered by users (i.e. spreadsheet style), so that users can create completely custom tables.

Supports control custom cell contents, foreground and background colors (and soon, preset row and column heights).

A brand new table designer dialog has been added which allows for customization of these tables.

TODO:
[x] More unit tests for designer dialog
[ ] Add icons to designer (post merge)
[x] Final tweaks for designer UX (likely to be more post merge, pending user feedback)
[x] Expose row/column height settings to designer

Should be easy to expand in future, and there's tons of possibilities for what we could add here:
- Expression based cells!
- Control over cell text alignment
- Merged cells
- Import and export table contents (e.g. from map layer, CSV, ...)
- Clipboard functionality.

Sponsored by City of Canning


nyalldawson:
(I should also add - all the non-vector-layer specific options which exist in attribute tables are exposed here, like cell margins and alternating row/column colors, styling of grid lines, etc)

Screenshot
image

nyalldawson:
This is ready for review and merge now

nyalldawson:
Here's a little screencast demonstrating the table in action:

Peek 2020-01-13 16-16

nirvn:
Testing your table PR, nice work as always.

Comments:
- IMHO the item should be renamed to "Fixed Table" (i.e. when we hover the toolbar item button's tooltip etc.) so it'd be Attribute Table vs. Fixed Table;
- We need a way to add new lines when editing cells, I'd vote for good old Ctrl+Enter keyboard shortcut;
- The "insert rows/columns before/after" labels would be better served in the singular form when your about to insert only one row / column.

Everything worked smoothly here. 

Eventual phase 2 in the future: this needs some form of importing (via clipboard, via CSV, whatever :) ).

Saijin-Naib:
This is incredible. This will allow for some really easy-to-make revision history tables for engineering layouts, for instance.

Brilliant, Nyall!

nyalldawson:
Thanks for the review @nirvn! I've addressed all your suggestions now, and also improved the table cell editing behavior considerably

nirvn:
@nyalldawson , changes are looking good. Nice work with the "weak" edit mode too.

nyalldawson:
@nirvn so are you happy to approve for merge?

nirvn:
Yep. I haven't done a line by line review but most it is GUI stuff, which tested fine.

Thanks for addressing the comments.

nirvn (approved):

nyalldawson:
Unrelated test failure

PR #33748, opened 2020-01-13
Fix some Qt 5.13 css parsing warnings

nyalldawson:
Qt 5.13 throws qt warnings relating to use of rgb with an alpha value


nyalldawson:
Unrelated test failure

PR #33751, opened 2020-01-13
Remove use of deprecated qInstallMsgHandler

nyalldawson:
Qt 5.13 seems to have disabled the flag for hiding qt 4 compatibility code deprecation warnings, so now building QGIS floods the compiler with deprecation warnings. (Which we'll need to fix before Qt6)

This fixes 1 of them.


nyalldawson:
Unrelated test failure

PR #33753, opened 2020-01-13
azure pipelines: output on failure

m-kuhn:
## Checklist


nyalldawson:
Thank you!!

m-kuhn:
Hmmm... that doesn't help anything...

@jef-n any idea how to get to this output? Is windows also overwriting environment variables during tests?

nyalldawson:
I suspect we also need the verbose mode switch passed to ctest -- but I can't see how that can be done (there's no variable for this one)

jef-n:
> I suspect we also need the verbose mode switch passed to ctest -- but I can't see how that can be done (there's no variable for this one)

Not sure why it works on the nightlies - configuration should be almost the same and doesn't need CTEST_OUTPUT_ON_FAILURE either.   Trying CTEST_USE_LAUNCHERS.

PR #33754, opened 2020-01-13
Add new mode to "Join Attributes by Location" to take attributes from matching feature with largest area of overlap only

nyalldawson:
This allows for easy polygon->polygon joins, where you expect there to be
only a single matching feature and don't want to include features which
are just touching or have just tiny sliver polygon overlaps.

Sponsored by SMEC/SJ

### DIFF
--- a/python/plugins/processing/tests/testdata/custom/join_lines_crossing_multiple_lines.gml
+++ b/python/plugins/processing/tests/testdata/custom/join_lines_crossing_multiple_lines.gml
@@ -1,0 +1,48 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<ogr:FeatureCollection
+     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+     xsi:schemaLocation="http://ogr.maptools.org/ join_lines_crossing_multiple_lines.xsd"
+     xmlns:ogr="http://ogr.maptools.org/"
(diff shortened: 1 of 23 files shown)


nyalldawson:
(I'm considering also adding a mode "Join to most similar feature", which would use the hausdorff distance of candidates to pick the feature with most similar geometry... it's a trivial addition with this commit, and sounds useful to me!).

@alexbruy thoughts on this in general?

alexbruy:
Nice addition and code looks good to me.

Not related to this directly, but maybe we should also provide a choice on how geometry predicates are handled, e.g. using "AND" or "OR" condition?

nyalldawson:
>(I'm considering also adding a mode "Join to most similar feature", which would use the hausdorff distance of candidates to pick the feature with most similar geometry... it's a trivial addition with this commit, and sounds useful to me!).

Dropping this idea for now -- to do it properly we shouldn't use spatial predicates, and rather should do a distance within type scan for candidate features for comparison

nyalldawson:
@alexbruy do you mind taking another look at this? I implemented some of the TODO heuristics and optimised the iteration when joining few feature to many.

E.g. when joining a 7 feature polygon layer to a 3.2 million feature polygon layer, processing time (on a debug build) has dropped from 696 seconds to 1.3 seconds

alexbruy (approved):

nyalldawson:
Unrelated test failure

PR #33756, opened 2020-01-13
add field name when reporting an error in refactorfield algo

Gustry:
## Description

Just to have more debug when one expression is failing and you have a lot of fields in the algorithm.

I also added `fatal=True`, is-it correct?
Can-it be backported to 3.10?

**Side question related to this PR**, a few lines below this, there is also:

```python
                    raise QgsProcessingException(
                        self.tr(u'Evaluation error in expression "{}": {}')
                            .format(expression.expression(),
                                    expression.evalErrorString()))
```

When should we raise an exception *or* a fatal message like in the PR? When using the algo with PyQGIS, does-it mean I need both methods at the same time to have **all** errors (try/catch AND custom feedback)?

```python

class RefactorFeedBack(QgsProcessingFeedback):
    def __init__(self, bar):
        self.bar = bar
        super().__init__()
    def reportError(self, error: str, fatalError: bool = ...) -> None:
        if fatalError:
            # To have fatal error message
            LOGGER.critical(error)
            self.bar.pushCritical('Refactor field feedback', error)

try:
    results = runAndLoadResults(alg_name, params, feedback, context)
except QgsProcessingException as e:
    LOGGER.critical(e)
    # To have exceptions from the algo
    self.bar.pushCritical('Refactor field', str(e))
    return
```

## Checklist

### DIFF
--- a/python/plugins/processing/algs/qgis/FieldsMapper.py
+++ b/python/plugins/processing/algs/qgis/FieldsMapper.py
@@ -108,9 +108,12 @@ class FieldsMapper(QgisFeatureBasedAlgorithm):
                 expression.setAreaUnits(context.project().areaUnits())
                 if expression.hasParserError():
                     feedback.reportError(
-                        self.tr(u'Parser error in expression "{}": {}')
-                        .format(expression.expression(),
-                                expression.parserErrorString()))
+                        self.tr('Parser error for field "{}" with expression "{}": {}')
+                        .format(
+                            field_def['name'],
+                            expression.expression(),
+                            expression.parserErrorString()),
+                        True)
                     return False
                 self.expressions.append(expression)
             else:


luipir (approved):

luipir:
> I also added `fatal=True`, is-it correct?

IMHO yes

luipir:
> Side question related to this PR

IMHO better to have public quiestion in dev list to share discussion and knowledge

> Can-it be backported to 3.10?

Should we backport not critical stuffs also? remember that this could trigger documentation update too!

Gustry:
> btw seems that flags is not used at all! https://qgis.org/api/qgsprocessingfeedback_8cpp_source.html#l00035

I saw also QgsProcessingAlgorithmDialogFeedback, not only QgsProcessingFeedBack.

nyalldawson:
@luipir 
> btw seems that flags is not used at all! https://qgis.org/api/qgsprocessingfeedback_8cpp_source.html#l00035

As per the dox, the QgsProcessingFeedback class is basically a null logger. It outputs to the message log only. Usually the QgsProcessingAlgorithmDialogFeedback subclass is the one actually in use when running algorithms (unless you're doing it from a standalone script)

That said, the fatalError flag was a mistake -- I shouldn't have added that. It *is* currently used by the dialog to know whether the error should also cause the dialog to be reset, but that should have been handled in a cleaner manner. I suggest we drop this for 4.0

For now, it's correct to use the flag here.

nyalldawson:
Unrelated test failure

PR #33668, opened 2020-01-08
[Server] Throw exception if bad layers are not restricted

rldhont:
## Description
QGIS Server throw an exception if the project has bad layers, but the user can defined restricted layers which are unpublished layers.

So restricted layers can be bad layers server side, it is not necessary an error.

This code verified that the bad layers are not restricted. If the project contains unrestricted bad layers, the server throw an exception.

## Checklist

- [x] Commit messages are descriptive and explain the rationale for changes
- [ ] Commits which fix bugs include `Fixes #11111` at the bottom of the commit message
- [x] I have read the [QGIS Coding Standards](https://docs.qgis.org/testing/en/docs/developers_guide/codingstandards.html) and this PR complies with them
- [x] New unit tests have been added for core changes
- [x] I have run [the `scripts/prepare-commit.sh` script](https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTING.md#contributing-to-qgis) before each commit
- [x] I have evaluated whether it is appropriate for this PR to be backported, backport requests are left as label or comment

PR #33759, opened 2020-01-13
[Server] Throw exception if bad layers are not restricted

rldhont:
Manually backport #33668 

## Description
QGIS Server throw an exception if the project has bad layers, but the user can defined restricted layers which are unpublished layers.

So restricted layers can be bad layers server side, it is not necessary an error.

This code verified that the bad layers are not restricted. If the project contains unrestricted bad layers, the server throw an exception.

## Checklist


elpaso (approved):

rldhont:
thanks @elpaso

PR #33760, opened 2020-01-13
Install various missing headers and fix some includes to work with installed headers

manisandro:


src/core/CMakeLists.txt
```diff
@@ -1134,6 +1134,12 @@ SET(QGIS_CORE_HDRS
   mesh/qgstriangularmesh.h
   mesh/qgsmeshtracerenderer.h
 
+  pal/pal.h
```
nyalldawson: Do we really need these ones? pal should be completely internal and not exposed anywhere

manisandro: It is included by `qgslabelfeature.h` [1].

https://github.com/qgis/QGIS/blob/master/src/core/labeling/qgslabelfeature.h#L26

nyalldawson: Ok, I should be able to fix that

nyalldawson: @manisandro see https://github.com/qgis/QGIS/pull/33803

manisandro: @nyalldawson Thanks!

nyalldawson (approved):

manisandro:
Uhm so is there a way to get a more useful explanation than `Cmd.exe exited with code '1'.
` for the Azure CI build failures?

jef-n:
> Uhm so is there a way to get a more useful explanation than `Cmd.exe exited with code '1'. ` for the Azure CI build failures?

check dash - link in "Build QGIS step"  - points at [build](https://cdash.orfeo-toolbox.org/index.php?project=QGIS&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercombine=and&filtercount=2&showfilters=1&filtercombine=and&field1=buildname&compare1=61&value1=PR+33760+/+master+(1137)+(5a5a4f46e1)+(qgis-dev+Experimental+x86)&field2=site&compare2=61&value2=azure-pipelines) and in turn [error](https://cdash.orfeo-toolbox.org/viewBuildError.php?buildid=26718).

manisandro:
Thanks, failure is unrelated (fixed by 69a53d945c28f2439246fee05be62baa68b2375b), merging.

PR #33763, opened 2020-01-13
Move resources to libqgisgui

manisandro:
Move resources to libqgisgui, so that they are available to applications using libqgisgui


nyalldawson:
The new test results here are correct following this change, so you'll just need to update the reference images with them. (i.e. the change isn't faulty, the tests just need updating)

manisandro:
@nyalldawson Can you elaborate what you mean by updating the reference images?

nyalldawson:
@manisandro 

If you look at https://cdash.orfeo-toolbox.org/testDetails.php?test=14837655&build=26837 and https://cdash.orfeo-toolbox.org/testDetails.php?test=14837608&build=26837 you'll see that this change has caused some of the render test results to change.

But the NEW images are acceptable given the change made here, so you'll need to update the versions in the qgis code to reflect the results rendered with this change applied.

manisandro:
I have the tests pass locally, but travis still fails with this result

https://cdash.orfeo-toolbox.org/testDetails.php?test=16809435&build=32615
https://cdash.orfeo-toolbox.org/testDetails.php?test=16809547&build=32615

Am I missing something or is the difference computed between the wrong images? After all, the expected and rendered images are the same.

m-kuhn:
What is the plan with applications only using core but not gui?

See e.g. https://github.com/qgis/QGIS/pull/34144

nyalldawson:
@m-kuhn we should just take the chance and move them to core. There's *"lots** of things in core which reference them

m-kuhn:
@nyalldawson :+1: 
Ideally we'd split them, but I'm afraid that this will be a nightmare to do.

nyalldawson:
@manisandro this is done now by https://github.com/qgis/QGIS/pull/35969

PR #33764, opened 2020-01-13
Install source sip files

manisandro:
To allow external applications based on QGIS to provide SIP bindings.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -404,4 +404,9 @@ FOREACH(module ${PY_MODULES})
     )
   ENDFOREACH(pyfile)
   PY_COMPILE(py${module} "${QGIS_PYTHON_OUTPUT_DIRECTORY}/${module}")
+
+  # install source sip files
+  FILE(GLOB sip_files ${CMAKE_CURRENT_BINARY_DIR}/${module}/*.sip)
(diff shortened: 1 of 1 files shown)


nirvn:
@manisandro , this broke make install with custom CMAKE_INSTALL_PREFIX.

CMake Error at python/cmake_install.cmake:168 (file):
  file cannot create directory: /usr/share/sip/qgis/core.  Maybe need
  administrative privileges.
Call Stack (most recent call first):
  cmake_install.cmake:63 (include)

nirvn:
The SIP bindings should likely only be installed onto the user system if the CMAKE_INSTALL_PREFIX string is empty/null.

manisandro:
> The SIP bindings should likely only be installed onto the user system if the CMAKE_INSTALL_PREFIX string is empty/null.

@nirvn Hmm that won't really work as prefix is `/usr/local` by default (so with `CMAKE_INSTALL_PREFIX` unset) but i.e. when packaging it is almost always set to  `/usr/`.

Either I need to tweak `SIP_DEFAULT_SIP_DIR` to be located under `CMAKE_INSTALL_PREFIX` for that install (which I kinda doubt is feasible in a robust way`), or I suppose better I just make the install conditional on something like `SIP_GLOBAL_INSTALL`, which can be off by default.

PR #33765, opened 2020-01-13
Fix NULL set of fkeys in QgsFeatureListModel

signedav:
## Description
The function `setExtraIdentifierValuesToNull` did not set the indexes. Leaded to not updating widgets like RelationReference:
bug_relation_reference

Now `setExtraIdentifierValuesToNull` calls `setExtraIdentifierValues( QVariantList() )` what makes it obsolete. The relation reference widgets calls `setExtraIdentifierValues( QVariantList() )` directly.

The tests are improved for this case.

### DIFF
--- a/src/core/qgsfeaturefiltermodel.h
+++ b/src/core/qgsfeaturefiltermodel.h
@@ -196,9 +196,8 @@ class CORE_EXPORT QgsFeatureFilterModel : public QAbstractItemModel
      * Allows specifying one value that does not need to match the filter criteria but will
      * still be available in the model as NULL value(s).
      * \since QGIS 3.10
-     * \deprecated since QGIS 3.12 use setExtraIdentifierValues( QVariantList() ) instead
      */
-    Q_DECL_DEPRECATED void setExtraIdentifierValuesToNull() SIP_DEPRECATED;
+    void setExtraIdentifierValuesToNull();
 
     /**
      * The index at which the extra identifier value is available within the model.
--- a/src/core/qgsfeaturefiltermodel.h
+++ b/src/core/qgsfeaturefiltermodel.h
@@ -196,8 +196,9 @@ class CORE_EXPORT QgsFeatureFilterModel : public QAbstractItemModel
      * Allows specifying one value that does not need to match the filter criteria but will
      * still be available in the model as NULL value(s).
      * \since QGIS 3.10
+     * \deprecated since QGIS 3.12 use setExtraIdentifierValues( QVariantList() ) instead
      */
-    void setExtraIdentifierValuesToNull();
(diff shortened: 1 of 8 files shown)


m-kuhn:
@troopa81 you should be fairly familiar with this code IIRC, can I ask you a review?

nyalldawson:
@signedav 

> "using of setIdentifierValues( QVariantList() ) when setting to null to set the index to null index"

Can I suggest that this commit message could be made more useful by adding a description of the bug its fixing? Right now the only way to determine what this has fixed is by the screencast in the PR message, but that's lost as soon as the PR is merged and the change just becomes a single line entry in the git log...

src/core/qgsfeaturefiltermodel.h
```diff
@@ -196,8 +196,9 @@ class CORE_EXPORT QgsFeatureFilterModel : public QAbstractItemModel
      * Allows specifying one value that does not need to match the filter criteria but will
      * still be available in the model as NULL value(s).
      * \since QGIS 3.10
+     * \deprecated since QGIS 3.12 use setExtraIdentifierValues( QVariantList() ) instead
      */
-    void setExtraIdentifierValuesToNull();
+    Q_DECL_DEPRECATED void setExtraIdentifierValuesToNull() SIP_DEPRECATED;
```
troopa81: Is it really necessary to deprecate this method? I mean, the way you fix it should work now and it doesn't harm to keep it.

Same comment worth for setIdentifierValuesToNull

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
```diff
@@ -391,12 +391,12 @@ void QgsRelationReferenceWidget::deleteForeignKeys()
   }
   else
   {
-    mComboBox->setIdentifierValuesToNull();
+    mComboBox->setIdentifierValues( QVariantList() );
```
troopa81: We could keep calling the setIdentifierValuesToNull since you fixed it no?

troopa81:
> @troopa81 you should be fairly familiar with this code IIRC, can I ask you a review?

Sure. Not completly aware of this part of the code, but I take a look

troopa81:
@signedav I made some minor comments but it seems OK to me

src/core/qgsfeaturefiltermodel.h
signedav: Okay, I removed it again. And used always this function on setting to null.

troopa81 (approved):
Failed test is unrelated to this PR (pointlocatortest, I'll take a look). It looks fine to me

signedav:
Thanks @troopa81

PR #33766, opened 2020-01-13
Move QgsGpsMarker to GUI, add QgsGpsMarker::setDirection

manisandro:


src/gui/CMakeLists.txt
```diff
@@ -161,6 +161,8 @@ SET(QGIS_GUI_SRCS
   editorwidgets/qgsvaluerelationsearchwidgetwrapper.cpp
   editorwidgets/qgsvaluerelationwidgetfactory.cpp
 
+  gps/qgsgpsmarker.cpp
```
nyalldawson: If you're adding a new directory here, you'll also need to update:

https://github.com/qgis/QGIS/blob/master/doc/CMakeLists.txt#L102
https://github.com/qgis/QGIS/blob/master/python/CMakeLists.txt#L138

And then go through the:

scripts/sip_include.sh
scripts/sipify_all.sh

process

src/gui/gps/qgsgpsmarker.h
```diff
@@ -24,10 +24,10 @@
...
 /**
- * \ingroup app
+ * \ingroup gui
  * A class for marking the position of a gps pointer.
  */
-class QgsGpsMarker : public QgsMapCanvasItem
+class GUI_EXPORT QgsGpsMarker : public QgsMapCanvasItem
 {
   public:
     explicit QgsGpsMarker( QgsMapCanvas *mapCanvas );
```
nyalldawson: If this is getting exposed to Python, you'll need to add SIP_TRANSFERTHIS 
```suggestion
    explicit QgsGpsMarker( QgsMapCanvas *mapCanvas SIP_TRANSFERTHIS );
```

src/gui/gps/qgsgpsmarker.h
```diff
@@ -24,10 +24,10 @@
 class QPainter;
 
 /**
- * \ingroup app
+ * \ingroup gui
  * A class for marking the position of a gps pointer.
  */
```
nyalldawson: Since it's newly added to the api, you can mark the since version as 3.12:

```suggestion
 * \since QGIS 3.12
 */
```

src/gui/gps/qgsgpsmarker.h
```diff
@@ -37,6 +37,12 @@ class QgsGpsMarker : public QgsMapCanvasItem
      */
     void setGpsPosition( const QgsPointXY &position );
 
+    /**
+     * Sets the current GPS heading \a direction (in degrees). Rotates the marker accordingly.
+     * \since QGIS 3.12
```
nyalldawson:
```suggestion
```

(given that it's being added on the class level)

src/gui/gps/qgsgpsmarker.cpp
```diff
@@ -63,6 +63,11 @@ void QgsGpsMarker::setGpsPosition( const QgsPointXY &point )
   setPos( pt );
 }
 
+void QgsGpsMarker::setDirection( double direction )
+{
+  mDirection = direction;
+}
```
nyalldawson: Does this need a "update()" call to redraw?

PR #33780, opened 2020-01-14
Show distance from GPS lock position to current cursor 

nyalldawson:
When a GPS device is connected, whenever the user moves the cursor over the canvas a live status bar message displays the distance and bearing from the cursor to the GPS fix position.

Project distance and bearing settings are respected in this display.

Peek 2020-01-14 16-06


src/app/gps/qgsgpsinformationwidget.cpp
```diff
@@ -1461,6 +1487,38 @@ void QgsGpsInformationWidget::timestampFormatChanged( int )
...
+
+  }
+}
+
+void QgsGpsInformationWidget::updateGpsDistanceStatusMessage()
+{
+  if ( !mNmea )
+    return;
+
+  QgsDistanceArea da;
```
nirvn: Out of curiosity, is that QgsDistanceArea expensive to create (setEllipsoid, setSourceCrs, etc.)? If the answer is yes, should it instead be a member of the QgsGpsInformationWidget class, and update the ellipsoid / source CRS through project signals?

nyalldawson: Good point, I'll do that

nyalldawson: Done

nirvn:
@nyalldawson , merge away! :)

PR #33785, opened 2020-01-14
Fix out-of-sync vector layer extent after a dataChanged was emitted

nirvn:
## Description

Stumbled on this issue while dissecting why a delimited text provider dataset's extent wouldn't reflect text file modification.

When a provider indicates changes to its dataset, we need to invalidate the cached QgsVectorLayer extent so that subsequent calls to that function will reflect the extent based on the updated dataset.

## Checklist

### DIFF
--- a/src/core/qgsvectorlayer.cpp
+++ b/src/core/qgsvectorlayer.cpp
@@ -5069,6 +5069,8 @@ void QgsVectorLayer::emitDataChanged()
   if ( mDataChangedFired )
     return;
 
+  updateExtents(); // reset cached extent to reflect data changes
+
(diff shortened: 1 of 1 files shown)


src/core/qgsvectorlayer.cpp
```diff
@@ -5069,6 +5069,8 @@ void QgsVectorLayer::emitDataChanged()
   if ( mDataChangedFired )
     return;
 
+  mValidExtent = false; // reset cached extent to reflect data changes
```
nyalldawson:
```suggestion
  updateExtents(); // reset cached extent to reflect data changes
```

PR #33790, opened 2020-01-14
fix exporting of the mesh contours for polygons

PeterPetrik:
fix exporting of the mesh contours for polygons. now the exported polygons does not match triangular mesh

there was copy-paste error when creating polygon from ring, where instead of ring points the triangular mesh coordinates were taken. also fixed bug  for one special case of values on the triangular mesh vertices that are not in min-max range of exported polygon

fixes issue https://github.com/lutraconsulting/qgis-crayfish-plugin/issues/419

before fix (see exported polygons match the triangular mesh)

Screenshot 2020-01-14 at 15 44 21

with fix (exported polygons match the exported lines)
Screenshot 2020-01-14 at 15 43 00

### DIFF
--- a/src/analysis/mesh/qgsmeshcontours.cpp
+++ b/src/analysis/mesh/qgsmeshcontours.cpp
@@ -196,8 +196,36 @@ QgsGeometry QgsMeshContours::exportPolygons(
         }
         else
         {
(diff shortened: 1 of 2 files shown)


wonder-sk (approved):
Looks good to me

PR #33792, opened 2020-01-14
Preserve boolean in offline editing

m-kuhn:
Preserve boolean fields when offline editing to gpkg.
This keeps expression based on boolean fields working without artistic stunts.


signedav (approved):
Code looks good. Haven't tested if the values can be copied without problems but I think there won't be any regressions, since before they had been strings.

PR #33794, opened 2020-01-14
[processing][needs-docs] port Rectangle,ovals,diamonds algorithm to C++

alexbruy:
## Description
Port "Rectangle, ovals, diamonds" algorithm to C++. Also make width, height and rotation parameters dynamic. 

Old Python implementations (using fixed and variable distance) marked as deprecated to maintain API compatibility.

## Checklist


src/analysis/processing/qgsalgorithmrectanglesovalsdiamonds.cpp
```diff
@@ -0,0 +1,209 @@
...
+}
+
+QString QgsRectanglesOvalsDiamondsAlgorithm::groupId() const
+{
+  return QStringLiteral( "vectorgeometry" );
+}
+
+QString QgsRectanglesOvalsDiamondsAlgorithm::shortHelpString() const
+{
+  return QObject::tr( "Creates a rectangle, oval or diamond-shaped polygons from the input point layer using "
```
nyalldawson:
```suggestion
  return QObject::tr( "Creates rectangle, oval or diamond-shaped polygons from the input point layer using "
```

src/analysis/processing/qgsalgorithmrectanglesovalsdiamonds.cpp
```diff
@@ -0,0 +1,209 @@
...
+  return new QgsRectanglesOvalsDiamondsAlgorithm();
+}
+
+void QgsRectanglesOvalsDiamondsAlgorithm::initParameters( const QVariantMap & )
+{
+  addParameter( new QgsProcessingParameterEnum( QStringLiteral( "SHAPE" ), QObject::tr( "Shape" ), QStringList() << QObject::tr( "Rectangle" ) << QObject::tr( "Diamond" ) << QObject::tr( "Oval" ), false, 0 ) );
+
+  auto widthParam = qgis::make_unique < QgsProcessingParameterDistance >( QStringLiteral( "WIDTH" ), QObject::tr( "Width" ), 1.0, QStringLiteral( "INPUT" ), false, 0.0 );
+  widthParam->setIsDynamic( true );
+  widthParam->setDynamicPropertyDefinition( QgsPropertyDefinition( QStringLiteral( "Width" ), QObject::tr( "Width" ), QgsPropertyDefinition::Double ) );
```
nyalldawson:
```suggestion
  widthParam->setDynamicPropertyDefinition( QgsPropertyDefinition( QStringLiteral( "Width" ), QObject::tr( "Width" ), QgsPropertyDefinition::DoublePositive ) );
```

src/analysis/processing/qgsalgorithmrectanglesovalsdiamonds.cpp
```diff
@@ -0,0 +1,209 @@
...
+  widthParam->setDynamicLayerParameterName( QStringLiteral( "INPUT" ) );
+  addParameter( widthParam.release() );
+
+  auto heightParam = qgis::make_unique < QgsProcessingParameterDistance >( QStringLiteral( "HEIGHT" ), QObject::tr( "Height" ), 1.0, QStringLiteral( "INPUT" ), false, 0.0 );
+  heightParam->setIsDynamic( true );
+  heightParam->setDynamicPropertyDefinition( QgsPropertyDefinition( QStringLiteral( "Height" ), QObject::tr( "Height" ), QgsPropertyDefinition::Double ) );
```
nyalldawson:
```suggestion
  heightParam->setDynamicPropertyDefinition( QgsPropertyDefinition( QStringLiteral( "Height" ), QObject::tr( "Height" ), QgsPropertyDefinition::DoublePositive ) );
```

src/analysis/processing/qgsalgorithmrectanglesovalsdiamonds.cpp
```diff
@@ -0,0 +1,209 @@
...
+  heightParam->setDynamicLayerParameterName( QStringLiteral( "INPUT" ) );
+  addParameter( heightParam.release() );
+
+  auto rotationParam = qgis::make_unique < QgsProcessingParameterNumber >( QStringLiteral( "ROTATION" ), QObject::tr( "Rotation" ), QgsProcessingParameterNumber::Double, 0.0, true, 0.0, 360.0 );
```
nyalldawson:
```suggestion
  auto rotationParam = qgis::make_unique < QgsProcessingParameterNumber >( QStringLiteral( "ROTATION" ), QObject::tr( "Rotation" ), QgsProcessingParameterNumber::Double, 0.0, true, -360.0, 360.0 );
```

(just in case someone would prefer to enter -45?)

src/analysis/processing/qgsalgorithmrectanglesovalsdiamonds.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+    if ( width == 0 || height == 0 )
+    {
+      throw QgsProcessingException( QObject::tr( "Width and height should be greater than 0." ) );
+    }
+
+    double phi = rotation * M_PI / 180;
+    double xOffset = width / 2.0;
+    double yOffset = height / 2.0;
+    QgsPointXY point = outFeature.geometry().asPoint();
```
nyalldawson: Possibly an issue here with multipoint inputs -- I'd suggest we raise an exception if they are encountered

src/analysis/processing/qgsalgorithmrectanglesovalsdiamonds.cpp
```diff
@@ -0,0 +1,209 @@
...
+    double x = point.x();
+    double y = point.y();
+
+    QgsPolylineXY ring;
```
nyalldawson: Instead of using the QgsPolylineXY class (which is slow), you could use the optimised method of creating polygons which is to construct QVector<double> for x and y, and use the QgsLineString( const QVector<double>& x, const QVector<double>& y ) constructor (and then set this as the polygon exterior ring). That's the **fastest** way to create polygons in qgis.

src/analysis/processing/qgsalgorithmrectanglesovalsdiamonds.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+    for ( int i = 0; i < ring.size(); ++i )
+    {
+      QgsPointXY p = ring[ i ];
+      double px = p.x();
+      double py = p.y();
+      ring[ i ] = QgsPointXY( px * cos( phi ) + py * sin( phi ) + x,
+                              -px * sin( phi ) + py * cos( phi ) + y );
+    }
+
```
nyalldawson: If input point has z or m, call addZValue or addMValue on the created geometry too

python/plugins/processing/algs/qgis/RectanglesOvalsDiamondsFixed.py
```diff
@@ -58,6 +59,9 @@ def groupId(self):
     def __init__(self):
         super().__init__()
 
+    def flags(self):
```
nyalldawson: If we register an alias (via QgsApplication::rprocessingRegistry()->addAlgorithmAlias( "qgis:rectanglesovalsdiamondsfixed", "native:rectanglesovalsdiamonds" ) we can just drop this whole file. (I've double checked, and the implementations should give identical results)

src/analysis/processing/qgsalgorithmrectanglesovalsdiamonds.cpp
alexbruy: Fixed

src/analysis/processing/qgsalgorithmrectanglesovalsdiamonds.cpp
alexbruy: Fixed

src/analysis/processing/qgsalgorithmrectanglesovalsdiamonds.cpp
alexbruy: Done

src/analysis/processing/qgsalgorithmrectanglesovalsdiamonds.cpp
alexbruy: Done

src/analysis/processing/qgsalgorithmrectanglesovalsdiamonds.cpp
alexbruy: Should be fixed

python/plugins/processing/algs/qgis/RectanglesOvalsDiamondsFixed.py
alexbruy: Done. I'm wondering if we should do the same with "Set raster/vector layer style" algorithms

src/analysis/processing/qgsalgorithmrectanglesovalsdiamonds.cpp
alexbruy: Done

src/analysis/processing/qgsalgorithmrectanglesovalsdiamonds.cpp
alexbruy: Good catch. Changed

python/plugins/processing/algs/qgis/QgisAlgorithmProvider.py
```diff
@@ -75,7 +75,6 @@
 from .RasterCalculator import RasterCalculator
 from .RasterLayerHistogram import RasterLayerHistogram
 from .RasterSampling import RasterSampling
-from .RectanglesOvalsDiamondsFixed import RectanglesOvalsDiamondsFixed
```
Gustry: Nice move!
You need to remove the help from the YML too.

alexbruy: Which help?

Gustry: This file https://github.com/qgis/QGIS/blob/master/python/plugins/processing/algs/help/qgis.yaml
But indeed, I can't see the alg, so it means that the help was missing before?

alexbruy: I know about this file, but it contains nothing related to these two algs, that's why I asked.

Gustry: Ok!  I can't resolve the comment.

src/analysis/processing/qgsalgorithmrectanglesovalsdiamonds.cpp
```diff
@@ -0,0 +1,232 @@
...
+QString QgsRectanglesOvalsDiamondsAlgorithm::groupId() const
+{
+  return QStringLiteral( "vectorgeometry" );
+}
+
+QString QgsRectanglesOvalsDiamondsAlgorithm::shortHelpString() const
+{
+  return QObject::tr( "Creates rectangle, oval or diamond-shaped polygons from the input point layer using "
+                      "specified width, height and (optional) rotation values. Multipart inputs should be promoted"
```
nyalldawson: Missing space at the end of this line

src/analysis/processing/qgsalgorithmrectanglesovalsdiamonds.cpp
```diff
@@ -0,0 +1,232 @@
...
+          ringX[ i ] = xOffset * cos( t );
+          ringY[ i ] = yOffset * sin( t );
+        }
+        ringX[ mSegments ] = ringX.at( 0 );
+        ringY[ mSegments ] = ringY.at( 0 );
+        break;
+    }
+
+    for ( int i = 0; i < ringX.size(); ++i )
+    {
```
nyalldawson: Micro optimisation: if Phi is zero, we should skip this loop

src/analysis/processing/qgsalgorithmrectanglesovalsdiamonds.cpp
```diff
@@ -0,0 +1,232 @@
...
+      ringX[ i ] = px * cos( phi ) + py * sin( phi ) + x;
+      ringY[ i ] = -px * sin( phi ) + py * cos( phi ) + y;
+    }
+
+    std::unique_ptr< QgsPolygon > poly = qgis::make_unique< QgsPolygon >();
+    poly->setExteriorRing( new QgsLineString( ringX, ringY ) );
+
+    if ( geometry.constGet()->is3D() )
+    {
+      poly->addZValue( 0 );
```
nyalldawson: This should be the input point z (or m), not zero

src/analysis/processing/qgsnativealgorithms.cpp
```diff
@@ -306,6 +307,7 @@ void QgsNativeAlgorithms::loadAlgorithms()
   addAlgorithm( new QgsAlgorithmRemoveDuplicateVertices() );
   addAlgorithm( new QgsReclassifyByLayerAlgorithm() );
   addAlgorithm( new QgsReclassifyByTableAlgorithm() );
+  addAlgorithm( new QgsRectanglesOvalsDiamondsAlgorithm() );
   addAlgorithm( new QgsRemoveDuplicatesByAttributeAlgorithm() );
```
nyalldawson: Don't forget to alias the old name in here!

nyalldawson (approved):
Just a few minor things left to fix

src/analysis/processing/qgsnativealgorithms.cpp
nyalldawson: Ah, I missed that

src/analysis/processing/qgsalgorithmrectanglesovalsdiamonds.cpp
alexbruy: Done

PR #33803, opened 2020-01-14
Avoid inclusion of pal header in public header

nyalldawson:
Create QgsLabeling class for labeling related enums and constants, and avoid inclusion of pal header in a public QGIS header

pal is an internal implementation detail -- it should not be exposed publicly

### DIFF
--- a/src/core/labeling/qgslabelfeature.h
+++ b/src/core/labeling/qgslabelfeature.h
@@ -23,11 +23,12 @@
 #include "qgsgeos.h"
 #include "qgsmargins.h"
 #include "qgslabelobstaclesettings.h"
-#include "pal.h"
+#include "qgslabeling.h"
 
 namespace pal
 {
   class LabelInfo;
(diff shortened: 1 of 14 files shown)


src/core/labeling/qgslabelfeature.h
```diff
@@ -470,33 +471,33 @@ class CORE_EXPORT QgsLabelFeature
...
-    double mZIndex;
+    double mZIndex = 0;
     //! whether mFixedPosition should be respected
-    bool mHasFixedPosition;
+    bool mHasFixedPosition = false;
     //! fixed position for the label (instead of automatic placement)
     QgsPointXY mFixedPosition;
     //! whether mFixedAngle should be respected
-    bool mHasFixedAngle;
+    bool mHasFixedAngle = false;
```
nirvn: Nice bonus cleanups.

nirvn (approved):

PR #33807, opened 2020-01-15
New algorithm "Rename table field"

nyalldawson:
Takes an input layer, existing field and a new name for the field, and
outputs a new layer with the selected field renamed.

While this result could also be achieved with the Refactor Fields
algorithm, Refactor Fields isn't particularly model friendly. It
relies on a constant, fixed table structure, and can't adapt to
input tables with different field structures.

In constrast, this simple Rename Field algorithm adapts nicely for
model use, because it operates on a single field only and leaves
all the other fields untouched.

### DIFF
--- a/src/analysis/processing/qgsalgorithmrenametablefield.cpp
+++ b/src/analysis/processing/qgsalgorithmrenametablefield.cpp
@@ -1,0 +1,115 @@
+/***************************************************************************
+                         qgsalgorithmrenametablefield.cpp
+                         -----------------------------------
+    begin                : January 2020
+    copyright            : (C) 2020 by Nyall Dawson
+    email                : nyall dot dawson at gmail dot com
+ ***************************************************************************/
+
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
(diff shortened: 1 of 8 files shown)


nirvn:
Nice, I've often wished for exactly this.

src/analysis/processing/qgsalgorithmrenametablefield.cpp
```diff
@@ -0,0 +1,115 @@
+/***************************************************************************
+                         qgsalgorithmrenametablefield.cpp
+                         -----------------------------------
+    begin                : November 2019
+    copyright            : (C) 2019 by Alexander Bruy
```
nirvn: Header copyright and email needs updating here?

src/analysis/processing/qgsalgorithmrenametablefield.h
```diff
@@ -0,0 +1,66 @@
+/***************************************************************************
+                         qgsalgorithmrenametablefield.h
+                         ---------------------
+    begin                : December 2019
```
nirvn: January 2020?

src/analysis/processing/qgsalgorithmrenametablefield.cpp
nyalldawson: Thanks, fixed

alexbruy (approved):
Nice addition

saberraz:
A bit out of context and in my wishlist, but would have been great to have a similar algorithm for In-place editing.

Issue #33312, opened 2019-12-10
Unable to use aggregate fucntions in Text on Symbol expressions


toasthall:
When trying to use aggregate in expressions to set labels for Text on Symbols I get the following error:

![Capture](https://user-images.githubusercontent.com/5704624/70494238-1c0b6b00-1b52-11ea-9fb4-8045747a8d97.PNG)

**QGIS and OS versions**

QGIS version | 3.10.0-A CoruÃ±a | QGIS code revision | 6c816b4204
-- | -- | -- | --
Compiled against Qt | 5.11.2 | Running against Qt | 5.11.2
Compiled against GDAL/OGR | 2.4.1 | Running against GDAL/OGR | 2.4.1
Compiled against GEOS | 3.8.0-CAPI-1.13.1 | Running against GEOS | 3.8.0-CAPI-1.13.1
Compiled against SQLite | 3.29.0 | Running against SQLite | 3.29.0
PostgreSQL Client Version | 11.5 | SpatiaLite Version | 4.3.0
QWT Version | 6.1.3 | QScintilla2 Version | 2.10.8
Compiled against PROJ | 5.2.0 | Running against PROJ | Rel. 5.2.0, September 15th, 2018
OS Version | Windows 10 (10.0)
Active python plugins | AttributeAssignment; autoSaver; CalcArea; CalculateGeometry; DigitizingTools; FreehandRasterGeoreferencer; latlontools; qgis2web; quick_map_services; scriptrunner3; SemiAutomaticClassificationPlugin; SpreadsheetLayers; track_profile_2_web; db_manager; MetaSearch; processing

gioman:
Maybe it is by design, but not 100% sure at the moment.

roya0045:
That's normal, I have PRs that would solve this but they are in limbo for the moment. See #30727

PR #33819, opened 2020-01-15
[Feature][Expression] convert legend symbols to expression

### LINKED ISSUE #137: Symbol aware legend expression
# QGIS Enhancement: Title

**Date** 2019/01/09

**Author** Myself, Nyall Dawson (@nyalldawson ) and whoever else want to participate.

**Contact** alroyliz0 at gmail dot com

**maintainer** unknown

**Version** QGIS 3.6 [edit 3.8?]

# Summary

The goal of this proposition is to replace the FeatureCounter in the legend by a QgsExpression in order to allow more complex elements to be displayed dynamically in the legend (such as field statistics grouped by legend and visible extent or any strings). This would evaluate the expression using the visible map extent (if filtering visible element is activated) and be done at for each symbology of a layer.

The goal would be to obtain statistics such as the total area(or any other variable) of visible features for each legend symbol with a static string at the end. Or any other type of expression.

Currently the design and implementation is still mostly in the planing stage. 

In order to make this possible some of these elements would need to be implemented:

- Symbol scope or any other mean to filter out legend symbol/keys and the associated features

-  Implement the base class to have the necessary filtering and behavior depending on the operation (separate behaviors for aggregate expressions, strings only, or field based)

- Change the UI to use the expression instead of the counter

- Refactor the related classes 

- ? implementing feature filtering based on extent and caching it instead of using requests?

## My Section
 Core implementation
Possibly UI
Unit test?
To be determined/whatever I can do.

## Proposed Solution

To be determined

### Example(s)

None

### Affected Files

qgslayertreemodellegendnode.cpp (line 509)
QgsLayoutItemLegend
QgsAggregatecalculator
Maybe every vector/raster renderer if the symbol scope or it's alternative needs to obtain a list of feature per key/symbol.
To be determined

## Performance Implications

Unknown

## Further Considerations/Improvements

-Optionnal: Caching visible layers per extent. This could be used to filter out QgsFeatureRequest or at least replace the option to filter out feature by Atlas extent in the legend .

## Backwards Compatibility

None?

## Issue Tracking ID(s)

*(optional)*

## Votes

(required)

N.B.: The project is still in elaboration, contributions/suggestions are very welcome.

roya0045:
## Description
This PR contains the #33632 and some other changes to allow the user to use aggregate functions in the legend.

This add the legend_expression variable that return an expression representing the symbology.

Enabling proper aggregation is done so by checking replacing the legend_expression in the aggregate filter by the evaluated version that will be used to add necessary conditions to filter out features.

Comments and feedback is appreciated.

closes #21988  #33312

And closes https://github.com/qgis/QGIS-Enhancement-Proposals/issues/137

## Checklist


roya0045:
The other alternative is to replace symbol_expression in `QgsVectorLayer::aggregate`
 by using
```c++
   if ( parameters.filter.contains( "var('symbol_expression')" ) )
    parameters.filter = parameters.filter.replace( "var('symbol_expression')",  QgsExpression::replaceExpressionText( "[% ( @symbol_expression ) %]", context ) );
```

This option seems safer and more reliable.

roya0045:
@nirvn from a UX point of view, is it better to have a dedicated slot in the aggregate expression to provide the symbol ID or simply insert the `@symbol_expression` in the filter parameter.
Though I guess having an extra parameter and not being able to use it when the context is not present would be more confusing than anything else.

roya0045:
Unsure why the pipeline fails to find some find and fails to find cmake.

roya0045:
@nyalldawson I did a test and somehow the pdf that was generated contained the symbol label but the aggregate expression and the other concatenated text was not displayed. I feel like the pdf could be generated before the expression is re evaluated since it appeared ok in the layout. Would you know of a way to stall export until the proper text is displayed and ensure it is exported properly?

roya0045:
@nyalldawson Another weird observation is that the values are properly displayed in the legend item area but in the map if the legend is selected it displays only the label. The values update once the legend is deselected. Those labels are the same that are exported via png or pdf.

roya0045:
@nirvn Since you are familiar with the map rendering/exporting process, would you know of a way to refresh part of a map or force the export to wait until everything is evaluated before generating an output?

roya0045:
@nyalldawson Ijust did a test with a fairly big dataset and the old approach (using fids) is far quicker than using a purely expression based approach. I am aware that my old PRs are staling out. Let me know what you think.

roya0045:
Tweaked my test setup,expressions are not that bad, what slowed things down was the counting step induced by the fids approach. Should be fine to merge then.

roya0045:
New thing, when caching the context, only one value is returned (which is normal since cache uses the expression and not the values), but when the context is not stored it just keeps looping and always count. I need to find out what is querying the data so much.

roya0045:
Ok the skip is implemented, it could be optimized but there seems to be something constantly looping and breaking the count, forcing the expression to be reevaluated again.
The present skip will only skip when match are found as a way to speed up things. Logic need to be improved.

roya0045:
I found a part of the issue. During the if check in drawLayerTitleInternal it forces the evaluation of every item in the legend when it checked only for the title. Data is called twice, I'll try to reduce the number of calls.

src/core/qgsvectorlayer.cpp
```diff
@@ -4167,6 +4167,11 @@ QVariant QgsVectorLayer::aggregate( QgsAggregateCalculator::Aggregate aggregate,
   if ( ok )
     *ok = false;
 
+  //make a copy or make the param not const
```
roya0045: @nyalldawson if you are fine with this approach I think it would be the best and more robust as opposed to parsing text and hoping that you evaluate/replace the right ones.

src/core/layertree/qgslayertreemodellegendnode.cpp
```diff
@@ -697,25 +698,31 @@ QString QgsSymbolLegendNode::evaluateLabel( const QgsExpressionContext &context,
     QgsExpressionContextScope *symbolScope = createSymbolScope();
     contextCopy.appendScope( symbolScope );
     contextCopy.appendScope( vl->createExpressionContextScope() );
-
+    QString nodeExpression = mLayerNode->labelExpression();
+    if ( nodeExpression.contains( "filter:=" ) && nodeExpression.contains( "@symbol_expression" ) )
+      nodeExpression = nodeExpression.replace( "@symbol_expression", QgsExpression::replaceExpressionText( "[% @symbol_expression %]", &contextCopy ) );
```
roya0045: Direct Parsing approach, less optimal and more prone to errors.

roya0045:
Unrelated test failure.

roya0045:
32 bit reference relation widget test fails, unrelated to the code changes proposed.

roya0045:
What's with the cygwin install error ?

roya0045:
Could someone add Feature and Expressions tags? Thanks.

gioman:
> Could someone add Feature and Expressions tags? Thanks.

done.

roya0045:
Sill waiting for a review on this, any feedback would be appreciated. 😞

roya0045:
Travis seem to be having issues with the logs.

roya0045:
By the looks of it the layout failure is not in the code I touched, https://cdash.orfeo-toolbox.org/testDetails.php?test=18426584&build=38146

nyalldawson:
@roya0045 care to rebase off master so I can review?

roya0045:
@nyalldawson Should be rebased and squashed.

roya0045:
@nyalldawson I expect the review to be relatively straightforward as this code leverages the rule based converter to generate an expression It then. converts any ELSE clause to a standalone version and evaluate the associated variable as a filter clause to perform the final evaluation on the symbol expression to properly filter out the data. This removes any need to cache the symbol fids as this was frowned upon.

I don't plan to make the cpp test pass until needed.

roya0045:
@nyalldawson Good old ping

roya0045:
@nyalldawson Any hopes to get a quick review on this ancestral PR?

roya0045:
@nyalldawson review?

roya0045:
eh

roya0045:
@nyalldawson ping in case you ever have some time to review this.

src/core/qgsvectorlayer.cpp
```diff
@@ -4286,6 +4286,15 @@ QVariant QgsVectorLayer::aggregate( QgsAggregateCalculator::Aggregate aggregate,
...
+  QgsAggregateCalculator::AggregateParameters paramCopy = QgsAggregateCalculator::AggregateParameters( parameters );
+  if ( paramCopy.filter.contains( "var('symbol_expression')" ) )
+  {
+    QString symbEx = QgsExpression::replaceExpressionText( "[% ( @symbol_expression ) %]", context );
```
roya0045, line 4293: @nyalldawson If you think that users understand that they need to add 'eval(' to have this variable as an expression and not as text in the query this may be removed as I understand this may be the most contentious part of this PR.

roya0045:
@nyalldawson As for the legend rendering one, was this one also on your short list?

roya0045:
@nyalldawson I have removed the potential contention change in this PR. The rest of the code should not introduce significant behaviour changes.

roya0045:
@nyalldawson rebased due to weird segfaults in tests in unrelated part of the tests.

roya0045:
@nyalldawson @m-kuhn I would like your wisdom regarding the functions `QgsLayerTreeLayer::ruleIterator` and `QgsLayerTreeLayer::elseFormatter` I would like to know if I should move them directly in the rulebasedrenderer? This could be used to make symbols keys that don't confuse the counter when copy pasting symbols( see #40206 ). But that may be over the top and would need re-evaluating from time to time to invalidate...

roya0045:
@nyalldawson @andreasneumann do you have suggestions to get this grandpa PR over the line or is this simply a matter of lack of time to review?

roya0045:
> @nyalldawson @m-kuhn I would like your wisdom regarding the functions `QgsLayerTreeLayer::ruleIterator` and `QgsLayerTreeLayer::elseFormatter` I would like to know if I should move them directly in the rulebasedrenderer? This could be used to make symbols keys that don't confuse the counter when copy pasting symbols( see #40206 ). But that may be over the top and would need re-evaluating from time to time to invalidate...

@nyalldawson @m-kuhn guidance regarding this point would be appreciated, I'm reposting in case you missed the first message or simply forgot about it.

roya0045:
@nyalldawson I am assuming that you have no preferance regarding the location of the code as the pr couldn't be reviewed yet.

roya0045:
@m-kuhn if you have some extra time on hand and are in lockdown guidance would be appreciated regarding the point raise previously (if that helps this pr).

roya0045:
Keeping active until review.

roya0045:
Status quo

roya0045:
Keeping alive.

roya0045:
@nyalldawson  I did some tests and it seems that the current implementation of this PR is what mandated #34476, should the proposed expression be made static to prevent the expensive overhead? Thanks,

src/core/layertree/qgslayertreemodellegendnode.cpp
```diff
@@ -309,8 +309,12 @@ QgsSymbolLegendNode::QgsSymbolLegendNode( QgsLayerTreeLayer *nodeLayer, const Qg
   }
 
   updateLabel();
+
   if ( QgsVectorLayer *vl = qobject_cast<QgsVectorLayer *>( nodeLayer->layer() ) )
+  {
     connect( vl, &QgsVectorLayer::symbolFeatureCountMapChanged, this, &QgsSymbolLegendNode::updateLabel );
+    connect( vl, &QgsVectorLayer::rendererChanged, nodeLayer, &QgsLayerTreeLayer::updateSymbolExpressions );
```
roya0045, line 316: I have no clue why but this fires CONSTANTLY with rule-based symbology. This explain a good part of the lag. I'll see if I can find a better way to track symbology changes.

roya0045:
@nyalldawson I have performed some optimization, I am not sure if the caching is needed in the layertreelayer. A review at your convenience would be appreciated.

roya0045:
Not stale

roya0045:
:(

roya0045:
same old

roya0045:
ok

roya0045:
staying alive

roya0045:
Frozen

roya0045:
ok

roya0045:
fine

nyalldawson:
@roya0045 I'm closing this PR, as I've just implemented a QgsFeatureRenderer::legendKeyToExpression() method in https://github.com/qgis/QGIS/pull/47811. 

This change should be reworked accordingly to use that method. (It will considerably simplify things for you!)

PR #33829, opened 2020-01-15
[Feature] Add support for displaying WMTS legend graphics in layer tree

manisandro:
Adds support for displaying WMTS legend graphics directly in the layer tree, as is already the case with WMS legend graphics.

Example:
image

Sample use case: https://wmts10.geo.admin.ch/EPSG/2056/1.0.0/WMTSCapabilities.xml

### DIFF
--- a/src/providers/wms/qgswmscapabilities.cpp
+++ b/src/providers/wms/qgswmscapabilities.cpp
@@ -1569,6 +1569,20 @@ void QgsWmsCapabilities::parseWMTSContents( const QDomElement &element )
 
         style.legendURLs << legendURL;
       }
+      QDomElement thirdChildElement = secondChildElement.firstChildElement( QStringLiteral( "LegendURL" ) );
+      if ( !thirdChildElement.isNull() )
+      {
+        QgsWmtsLegendURL legendURL;
+
+        legendURL.format   = thirdChildElement.attribute( QStringLiteral( "format" ) );
+        legendURL.minScale = thirdChildElement.attribute( QStringLiteral( "minScaleDenominator" ) ).toDouble();
+        legendURL.maxScale = thirdChildElement.attribute( QStringLiteral( "maxScaleDenominator" ) ).toDouble();
+        legendURL.href     = thirdChildElement.attribute( QStringLiteral( "xlink:href" ) );
+        legendURL.width    = thirdChildElement.attribute( QStringLiteral( "width" ) ).toInt();
+        legendURL.height   = thirdChildElement.attribute( QStringLiteral( "height" ) ).toInt();
+
+        style.legendURLs << legendURL;
+      }
 
       style.isDefault = secondChildElement.attribute( QStringLiteral( "isDefault" ) ) == QLatin1String( "true" );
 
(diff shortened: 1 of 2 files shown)


src/providers/wms/qgswmscapabilities.cpp
```diff
@@ -1569,6 +1569,20 @@ void QgsWmsCapabilities::parseWMTSContents( const QDomElement &element )
 
         style.legendURLs << legendURL;
       }
+      QDomElement thirdChildElement = secondChildElement.firstChildElement( QStringLiteral( "LegendURL" ) );
+      if ( !thirdChildElement.isNull() )
+      {
+        QgsWmtsLegendURL legendURL;
+
+        legendURL.format   = thirdChildElement.attribute( QStringLiteral( "format" ) );
+        legendURL.minScale = thirdChildElement.attribute( QStringLiteral( "minScale" ) ).toDouble();
```
nyalldawson: Shouldn't this be minScaleDenominator and maxScaleDenominator?

From http://schemas.opengis.net/wmts/1.0/wmtsGetCapabilities_response.xsd:

```
<element name="LegendURL">
<annotation>
<documentation>
Zero or more LegendURL elements may be provided, providing an image(s) of a legend relevant to each Style of a Layer. The Format element indicates the MIME type of the legend. minScaleDenominator and maxScaleDenominator attributes may be provided to indicate to the client which scale(s) (inclusive) the legend image is appropriate for. (If provided, these values must exactly match the scale denominators of available TileMatrixes.) width and height attributes may be provided to assist client applications in laying out space to display the legend.
</documentation>
</annotation>
<complexType>
<complexContent>
<extension base="ows:OnlineResourceType">
<annotation>
<documentation>
The URL from which the legend image can be retrieved
</documentation>
</annotation>
<attribute name="format" type="ows:MimeType">
<annotation>
<documentation>A supported output format for the legend image</documentation>
</annotation>
</attribute>
<attribute name="minScaleDenominator" type="double">
<annotation>
<documentation>
Denominator of the minimum scale (inclusive) for which this legend image is valid
</documentation>
</annotation>
</attribute>
<attribute name="maxScaleDenominator" type="double">
<annotation>
<documentation>
Denominator of the maximum scale (exclusive) for which this legend image is valid
</documentation>
</annotation>
</attribute>
```

src/providers/wms/qgswmsprovider.cpp
```diff
@@ -303,6 +303,30 @@ QString QgsWmsProvider::getLegendGraphicUrl() const
...
   }
 
+  if ( url.isEmpty() )
+  {
+    for ( const QgsWmtsTileLayer &l : mCaps.mTileLayersSupported )
+    {
+      if ( l.identifier != mSettings.mActiveSubLayers[0] )
+        continue;
+
+      QHash<QString, QgsWmtsStyle>::const_iterator it = l.styles.find( mSettings.mActiveSubStyles[0] );
```
nyalldawson:
```suggestion
      QHash<QString, QgsWmtsStyle>::const_iterator it = l.styles.constFind( mSettings.mActiveSubStyles[0] );
      if ( it == l.styles.constEnd() )
```
(you can't mix const iterators with non const ones)

nyalldawson (approved):

PR #33831, opened 2020-01-15
Make locator prefixes case insensitive

uclaros:
## Description
All searches made with the locator widget are case insensitive but the prefix itself has to be written in lower case. This commit makes the prefix checking case insensitive, allowing for example searching with either f or F, as is also the case (no pun intended) with the similar locator in qtcreator.

## Checklist

I guess backporting depends on whether one sees this as a bug fix or as a new feature.

### DIFF
--- a/src/core/locator/qgslocator.cpp
+++ b/src/core/locator/qgslocator.cpp
@@ -57,7 +57,7 @@ QList<QgsLocatorFilter *> QgsLocator::filters( const QString &prefix )
     QList<QgsLocatorFilter *> filters =  QList<QgsLocatorFilter *>();
     for ( QgsLocatorFilter *filter : mFilters )
     {
-      if ( !filter->activePrefix().isEmpty() && filter->activePrefix() == prefix )
+      if ( !filter->activePrefix().isEmpty() && filter->activePrefix().compare( prefix, Qt::CaseInsensitive ) == 0 )
       {
         filters << filter;
       }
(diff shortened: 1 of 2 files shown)


nyalldawson (commented):
Could you add a unit test to https://github.com/qgis/QGIS/blob/master/tests/src/python/test_qgslocator.py#L188 covering this?

uclaros:
Added tests for both prefix and custom prefix.
The existing test for custom prefix:
```
        # filter with custom prefix
        QgsSettings().setValue("locator_filters/prefix_test_custom", 'xyz', QgsSettings.Gui)
        filter_c = test_filter('custom', 'abc')
        l.registerFilter(filter_c)
        self.assertEqual(filter_c.prefix(), 'abc')
        self.assertEqual(filter_c.activePrefix(), 'xyz')
        got_hit._results_ = []
        l.fetchResults('b', context)
        for i in range(100):
            sleep(0.002)
            QCoreApplication.processEvents()
        self.assertEqual(set(got_hit._results_), {'custom0', 'custom1', 'custom2'})
```
 was not covering the actual prefix since it's run with `filter_c.useWithoutPrefix == True`
Two test could be added that check 
1. that the custom prefix works (maybe redundant since it's covered by the new case insensitive one)
2. the original prefix does no longer work, like:
```
        filter_c.setUseWithoutPrefix(False)
        got_hit._results_ = []
        l.fetchResults('xyz b', context)
        for i in range(100):
            sleep(0.002)
            QCoreApplication.processEvents()
        self.assertEqual(set(got_hit._results_), {'custom0', 'custom1', 'custom2'})
        got_hit._results_ = []
        l.fetchResults('abc b', context)
        for i in range(100):
            sleep(0.002)
            QCoreApplication.processEvents()
        self.assertEqual(got_hit._results_, [])
```
If you agree, should I commit it to this PR or issue a new?
Thanks.

nyalldawson:
> If you agree, should I commit it to this PR or issue a new?

Can you issue a new one please?

PR #33832, opened 2020-01-15
Add new algorithm "Detect Dataset Changes"

nyalldawson:
This algorithm compares two vector layers, and determines which features
are unchanged, added or deleted between the two. It is designed for comparing
two different versions of the same dataset.

When comparing features, the original and revised feature geometries will be
compared against each other. Depending on the Geometry Comparison Behavior setting,
the comparison will either be made using an exact comparison (where geometries must
be an exact match for each other, including the order and count of vertices) or a
topological comparison only (where are geometries area considered equal if all of
the their component edges overlap. E.g. lines with the same vertex locations but
opposite direction will be considered equal by this method). If the topological
comparison is selected then any z or m values present in the geometries will not
be compared.

By default, the algorithm compares all attributes from the original and revised
features. If the Attributes to Consider for Match parameter is changed, then only
the selected attributes will be compared (e.g. allowing users to ignore a timestamp
or ID field which is expected to change between the revisions).

If any features in the original or revised layers do not have an associated geometry,
then care must be taken to ensure that these features have a unique set of
attributes selected for comparison. If this condition is not met, warnings will be
raised and the resultant outputs may be misleading.

The algorithm outputs three layers, one containing all features which are considered
to be unchanged between the revisions, one containing features deleted from the
original layer which are not present in the revised layer, and one containing features
add to the revised layer which are not present in the original layer.

### DIFF
--- a/src/analysis/processing/qgsalgorithmdetectdatasetchanges.cpp
+++ b/src/analysis/processing/qgsalgorithmdetectdatasetchanges.cpp
@@ -1,0 +1,432 @@
+/***************************************************************************
+                         qgsalgorithmdetectdatasetchanges.cpp
+                         -----------------------------------------
+    begin                : December 2019
+    copyright            : (C) 2019 by Nyall Dawson
+    email                : nyall dot dawson at gmail dot com
+ ***************************************************************************/
+
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+#include "qgsalgorithmdetectdatasetchanges.h"
+#include "qgsvectorlayer.h"
+#include "qgsgeometryengine.h"
+
+///@cond PRIVATE
+
+QString QgsDetectVectorChangesAlgorithm::name() const
+{
+  return QStringLiteral( "detectvectorchanges" );
+}
+
+QString QgsDetectVectorChangesAlgorithm::displayName() const
+{
+  return QObject::tr( "Detect dataset changes" );
+}
+
+QStringList QgsDetectVectorChangesAlgorithm::tags() const
+{
+  return QObject::tr( "added,dropped,new,deleted,features,geometries,difference,delta,revised,original,version" ).split( ',' );
+}
+
+QString QgsDetectVectorChangesAlgorithm::group() const
+{
+  return QObject::tr( "Vector general" );
+}
+
+QString QgsDetectVectorChangesAlgorithm::groupId() const
+{
+  return QStringLiteral( "vectorgeneral" );
+}
+
+void QgsDetectVectorChangesAlgorithm::initAlgorithm( const QVariantMap & )
+{
+  addParameter( new QgsProcessingParameterFeatureSource( QStringLiteral( "ORIGINAL" ), QObject::tr( "Original layer" ) ) );
+  addParameter( new QgsProcessingParameterFeatureSource( QStringLiteral( "REVISED" ), QObject::tr( "Revised layer" ) ) );
+
+  std::unique_ptr< QgsProcessingParameterField > compareAttributesParam = qgis::make_unique< QgsProcessingParameterField >( QStringLiteral( "COMPARE_ATTRIBUTES" ),
+      QObject::tr( "Attributes to consider for match (or none to compare geometry only)" ), QVariant(),
+      QStringLiteral( "ORIGINAL" ), QgsProcessingParameterField::Any, true, true );
+  compareAttributesParam->setDefaultToAllFields( true );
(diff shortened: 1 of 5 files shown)


nyalldawson:
(basically a 1:1 port of the FME "changedetector" transformer!)

src/analysis/processing/qgsalgorithmdetectdatasetchanges.cpp
```diff
@@ -0,0 +1,432 @@
+/***************************************************************************
+                         qgsalgorithmdetectdatasetchanges.cpp
+                         -----------------------------------------
+    begin                : December 2019
```
nirvn: January 2020

src/analysis/processing/qgsalgorithmdetectdatasetchanges.cpp
```diff
@@ -0,0 +1,432 @@
...
+    case Topological:
+      if ( QgsWkbTypes::geometryType( mOriginal->wkbType() ) != QgsWkbTypes::geometryType( mRevised->wkbType() ) )
+        throw QgsProcessingException( QObject::tr( "Geometry type of revised layer (%1) does not match the original layer (%2)" ).arg( QgsWkbTypes::geometryDisplayString( QgsWkbTypes::geometryType( mRevised->wkbType() ) ),
+                                      QgsWkbTypes::geometryDisplayString( QgsWkbTypes::geometryType( mOriginal->wkbType() ) ) ) );
+      break;
+
+  }
+
+  if ( mOriginal->sourceCrs() != mRevised->sourceCrs() )
+    feedback->reportError( QObject::tr( "CRS for revised layer (%1) does not match the original layer (%2) - reprojection accuracy may affect geometry matching" ).arg( mOriginal->sourceCrs().userFriendlyIdentifier(),
```
nirvn: That reminds me we should consider a reportWarning function that would, in terms of semantic priority, sit in between exciting pushInfo and reportError.

src/analysis/processing/qgsalgorithmdetectdatasetchanges.cpp
```diff
@@ -0,0 +1,432 @@
...
+            continue;
+          }
+        }
+
+        QgsGeometry original = originalGeometries.value( candidateId );
+        // lazy evaluation
+        if ( revised.isNull() )
+        {
+          revised = revisedFeature.geometry();
+          // drop z/m if not wanted for match
```
nirvn: Do we really want to unconditionally drop z here? I.e. is a topological match only based on a 2D representation? Wondering if it'd be better to have Topological2D and Topological3D. Or that might simply be overdoing it :)

src/analysis/processing/qgsalgorithmdetectdatasetchanges.cpp
```diff
@@ -0,0 +1,432 @@
...
+    {
+      // unchanged
+      if ( unchangedSink )
+        unchangedSink->addFeature( f, QgsFeatureSink::FastInsert );
+    }
+    else
+    {
+      // deleted feature
+      if ( deletedSink )
+        deletedSink->addFeature( f, QgsFeatureSink::FastInsert );
```
nirvn: ... stupid github won't let me delete comments.

src/analysis/processing/qgsalgorithmdetectdatasetchanges.h
```diff
@@ -0,0 +1,78 @@
+/***************************************************************************
+                         qgsalgorithmdetectdatasetchanges.h
+                         ---------------------------------------
+    begin                : December 2019
```
nirvn: January 2020

src/analysis/processing/qgsalgorithmdetectdatasetchanges.cpp
nyalldawson: Well, I started this last year so technically it did begin in December!

src/analysis/processing/qgsalgorithmdetectdatasetchanges.cpp
nyalldawson: That would depend on support in geos. Currently it doesn't handle these, so we drop them (and this limitation is mentioned in the algorithm documentation)

src/analysis/processing/qgsalgorithmdetectdatasetchanges.cpp
nyalldawson: Yeah, I agree

src/analysis/processing/qgsalgorithmdetectdatasetchanges.cpp
nirvn: @nyalldawson , okeydokey

alexbruy (approved):

nyalldawson:
Thanks @alexbruy and @nirvn!

Saijin-Naib:
This is killer. A topological geo-diff that can be integrated into a processing model? Iterative updating of data from State sources just got a lot better!

PedroVenancio:
Awesome @nyalldawson !

PR #33835, opened 2020-01-15
Allow hiding layer tree legend icons

manisandro:


nyalldawson:
If this setting is on the model level, why not just subclass the model and return QVariant for DecorationRoles in the subclass?

manisandro:
Sure, can also handle it completely KADAS-side that way, good point. Abandoning.

nyalldawson:
Thanks @manisandro! (funnily enough - i had the same requirement a year or so ago, and also initially went down the track of changing core code!)

PR #33836, opened 2020-01-15
Fix server mingw build

manisandro:


src/server/qgsserver.cpp
```diff
@@ -290,7 +290,7 @@ bool QgsServer::init()
 
 void QgsServer::putenv( const QString &var, const QString &val )
 {
-#ifdef _MSC_VER
+#ifdef Q_OS_WIN32
```
nyalldawson: this function should be using qputenv -- that's already got all the cross platform handling done for us!

nyalldawson:
@pblottiere ?

pblottiere:
Hi @manisandro,

Sorry for the latency :/.

It looks good to me, but can you use `qunsetenv()` when the value is empty instead? Otherwise we'll have a different behavior between platforms (empty vs deleted).

Thanks :+1:

manisandro:
Sure, done, thanks for the feedback.

pblottiere:
Thanks @manisandro

PR #6527, opened 2018-03-05
[needs-doc] Add stable id, save id, save count to project spec

NathanW2:
Adds some new metadata about the project itself for future use.

```
<qgis version="2.99.0-Master" save-counter="7" projectname="" save-user-full="Nathan" id="3ed6e416-b659-455d-a870-99971197890" save-id="64e96782-d2aa-4350-b2f6-7e77f31eece" save-user="nathan">
```
Adds the following information:

- id() -> Stable ID that never changes (can be used to track the project even if the file name changes)
- saveId() -> Changes on each save to give you a ID number for that version
- saveCounter() -> A int with a counter how many times this project has been saved. Same kind of deal as saveId() but we will have both anyway
- saveUser() -> Last saved user
- saveUserFull -> Last saved user full name.

**NOTE:** I was considering backporting this into 3.0 so that we can start getting projects including those ids for future release. If you save with a older version of QGIS it will nuke these values.

PR #33843, opened 2020-01-16
add userLoginName and userFullName to QGIS project when writing the p…

thbaumann:
## Description
In order to track changes in project files used by more than one user it is important to know who saved certain changes to these project files. To accomplish this I would like to add the user loginname and the user full name to the QGIS project file while writing the project.
Tracking project files can be done with git and in addition with the username as a part of the QGIS project it allows you to reconstruct any changes of these project files.

http://osgeo-org.1560.x6.nabble.com/QGIS-Developer-saving-username-in-QGIS-projectfile-for-tracking-changes-td5425750.html

I picked up the neccessary parts from the pullrequest 6527 (https://github.com/qgis/QGIS/pull/6527) which contained also changes to add stable id, save id and save count. It turned out that adding stable id's /save id's is quite tricky so this old pull request was closed. The parts to save the username were not mentioned as problematic.

In combination with the "Stable QGIS Projects" plugin from opengis this pullrequest allows better tracking of qgis projects ( https://www.opengis.ch/de/2019/04/09/plugin-for-tracking-qgis-project-files-in-git/ ) 

I don't think that it is neccessary to backport this pullrequest but for upcoming versions it would be great to have this functionality added to QGIS.

## Checklist

### DIFF
--- a/src/core/qgsproject.cpp
+++ b/src/core/qgsproject.cpp
@@ -473,6 +473,16 @@ QString QgsProject::title() const
   return mMetadata.title();
 }
 
+QString QgsProject::saveUser() const
+{
+	return mSaveUser;
+}
+
+QString QgsProject::saveUserFullname() const
+{
+	return mSaveUserFull;
+}
+
 bool QgsProject::isDirty() const
 {
   return mDirty;
@@ -725,6 +735,8 @@ void QgsProject::clear()
   mProjectScope.reset();
   mFile.setFileName( QString() );
   mProperties.clearKeys();
+  mSaveUser.clear();
+  mSaveUserFull.clear();
   mHomePath.clear();
   mCachedHomePath.clear();
   mAutoTransaction = false;
@@ -899,6 +911,24 @@ static void _getTitle( const QDomDocument &doc, QString &title )
   title = titleText.data();
 
 }
+
+static void getProjectMetadata(const QDomDocument &doc, QString &lastUser, QString &lastUserFull)
+{
+	QDomNodeList nl = doc.elementsByTagName(QStringLiteral("qgis"));
+
+	if (!nl.count())
+	{
+		QgsDebugMsg("unable to find qgis element");
+		return;
+	}
+
+	QDomNode qgisNode = nl.item(0);  // there should only be one, so zeroth element OK
+
+	QDomElement qgisElement = qgisNode.toElement(); // qgis node should be element
+	lastUser = qgisElement.attribute(QStringLiteral("save-user"), QString());
+	lastUserFull = qgisElement.attribute(QStringLiteral("save-user-full"), QString());
+}
+
 
 QgsProjectVersion getVersion( const QDomDocument &doc )
 {
@@ -1235,6 +1265,8 @@ bool QgsProject::readProjectFile( const QString &filename, QgsProject::ReadFlags
   QString oldTitle;
   _getTitle( *doc, oldTitle );
 
+  getProjectMetadata(*doc, mSaveUser, mSaveUserFull);
+
   QDomNodeList homePathNl = doc->elementsByTagName( QStringLiteral( "homePath" ) );
   if ( homePathNl.count() > 0 )
   {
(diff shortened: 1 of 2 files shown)


src/core/qgsproject.cpp
```diff
@@ -1951,7 +1986,10 @@ bool QgsProject::writeProjectFile( const QString &filename )
   QDomElement qgisNode = doc->createElement( QStringLiteral( "qgis" ) );
   qgisNode.setAttribute( QStringLiteral( "projectname" ), title() );
   qgisNode.setAttribute( QStringLiteral( "version" ), QStringLiteral( "%1" ).arg( Qgis::version() ) );
-
+  QString newSaveUser = QgsApplication::userLoginName();
+  QString newSaveUserFull = QgsApplication::userFullName();
+  qgisNode.setAttribute(QStringLiteral("save-user"), newSaveUser);
```
elpaso: Just a question: is that dashed style consistent with the majority of tag/attrs name in the project XML? I see there are many camel-case items too.

src/core/qgsproject.cpp
```diff
@@ -900,6 +912,24 @@ static void _getTitle( const QDomDocument &doc, QString &title )
 
 }
 
+static void getProjectMetadata(const QDomDocument &doc, QString &lastUser, QString &lastUserFull)
```
elpaso: Wouldn't be better named getProjectUserInformation ? We actually do have other metadata in the project file.

elpaso:
LGTM codewise, I just wonder if disclosing the username wouldn't be possibly seen as an undesired information disclosure, but perhaps I'm just too paranoid.

NathanW2:
Hold off on this please. This could leak info about a user that might not
be good to leak.

NathanW2:
And yes I am aware of that PR being mine.

thbaumann:
> 
> Hold off on this please. This could leak info about a user that might not be good to leak.

Are you concerned about the user full name or the login name?

One option could be to write the username to the projectfile only if an option is set like it is done for the author of new projects. This information is already written to the project file:

`if ( !settings.value( QStringLiteral( "projects/anonymize_new_projects" ), false, QgsSettings::Core ).toBool() )
  { ....}`

src/core/qgsproject.cpp
thbaumann: There is a mix of different styles. Some have the dashed style, some camelcase and some are just written in lowercase like <customproperties>. Are perhaps the tags with the dashed style like <layer-tree-group>,<layer-tree-layer>,<snapping-settings> older ones and the camelcase items like <ProjectViewSettings>,<projectCrs>,<homePath> mainly newer ones?

src/core/qgsproject.cpp
elpaso: Sorry, forgot to mention: QT and QGIS policy is to not use `get` for getters, just `projectUserInformation()` will do.

thbaumann: ok. good to know

Gustry:
The full name is already written in the project metadata tab by default.

nirvn:
I'll provide some comments on this tomorrow

Saijin-Naib:
Possibly related to/enhances:
Issue #29575 - New project variables
Issue ##29576 - Tie Project Properties -> Metadata -> History into a project save history function

nyalldawson:
@nathanw2 
> Hold off on this please. This could leak info about a user that might not
be good to leak.

There's lots of precendent here -- as @gustry has pointed out we already store the original author in the metadata section (and have for many releases). We also include this by default when exporting images and PDFs (in the export metadata). The export behavior can be disabled via the `projects\anonymize_new_projects` setting - so I think we could either reuse this setting to control the new behavior, or add a similar key to allow it to be manually disabled.

NathanW2:
Ok in that case I'm fine with it provided it provides ways to be disabled
at a system level for all projects.

nyalldawson:
Superseded by #33870

pigreco:
"projects \ anonymize_new_projects" and "projects \ anonymize_saved_projects"
The two variables should be set to `true` and not to` false` so that the change is voluntary and not enforced by others.
Otherwise you need a message that warns you every time you save a project.
m2c

PR #33844, opened 2020-01-16
Move layout classes to GUI, except for QgsLayoutDesignerDialog

manisandro:
So that external applications can provide a layout designer GUI.


nyalldawson:
I'm not a fan of exposing all these internals to the stable API. What would you think about moving the classes to GUI, but not exposing them, and instead requiring API users to QgsLayoutItemGuiMetadata to obtain the actual widgets?

manisandro:
Sure I'd be ok with that! So adding `#define SIP_NO_FILE` to all files not to be exposed?

nyalldawson:
Yes. I can't tell from this diff, but were files only moved here? I'd have thought some of these classes referenced other parts of app?

manisandro:
Most were just moved, only in rare cases I had to actually change something (like passing a QgsMapCanvas instance).

src/gui/layout/qgslayoutappmenuprovider.h
```diff
@@ -16,28 +16,29 @@
...
 #include <QObject>
 
-class QgsLayoutDesignerDialog;
+class QgsLayoutDesignerInterface;
 
 /**
  * A menu provider for QgsLayoutView
  */
-class QgsLayoutAppMenuProvider : public QObject, public QgsLayoutViewMenuProvider
+class GUI_EXPORT QgsLayoutAppMenuProvider : public QObject, public QgsLayoutViewMenuProvider
```
nyalldawson: Instead of moving this one (which should be strongly tied to the designer dialog), couldn't you use the QgsLayoutViewMenuProvider interface instead to implement your own provider?

src/gui/CMakeLists.txt
```diff
@@ -199,6 +227,11 @@ SET(QGIS_GUI_SRCS
   layout/qgslayoutviewtooltemporarykeyzoom.cpp
   layout/qgslayoutviewtooltemporarymousepan.cpp
   layout/qgslayoutviewtoolzoom.cpp
+  layout/qgsreportfieldgroupsectionwidget.cpp
```
nyalldawson: Do you have a specific need for these ones to be in gui? I'd say they are still very immature and I'd rather have full flexibility to tweak them in future without the restrictions of being in the gui library.

src/gui/layout/qgslayoutapputils.h
```diff
@@ -16,18 +16,22 @@
...
 
+#include "qgis_gui.h"
+
+class QgsMapCanvas;
+
 /**
  * Utils for layout handling from app.
  */
-class QgsLayoutAppUtils
+class GUI_EXPORT QgsLayoutAppUtils
```
nyalldawson: This class code be renamed

src/gui/layout/qgslayoutlegendwidget.cpp
```diff
@@ -63,9 +64,9 @@ static int _originalLegendNodeIndex( QgsLayerTreeModelLegendNode *legendNode )
 }
 
 
-QgsLayoutLegendWidget::QgsLayoutLegendWidget( QgsLayoutItemLegend *legend )
+QgsLayoutLegendWidget::QgsLayoutLegendWidget( QgsLayoutItemLegend *legend, QgsMapCanvas *mapCanvas )
   : QgsLayoutItemBaseWidget( nullptr, legend )
-  , mLegend( legend )
+  , mLegend( legend ), mMapCanvas( mapCanvas )
```
nyalldawson:
```suggestion
  , mLegend( legend )
  , mMapCanvas( mapCanvas )
```

src/gui/layout/qgslayoutappmenuprovider.h
manisandro: Sure, no problem copying it to the KADAS tree. Actually though there is the `QgsLayoutDesignerInterface` which looked like an attempt to decouple this and other classes from the actual designer implementation. Looked like a good way to spare some redundant copy-pasted code?

src/gui/CMakeLists.txt
manisandro: No I think I can do without them - I'll move them back.

src/gui/layout/qgslayoutappmenuprovider.h
nyalldawson: With QgsLayoutDesignerInterface I'm taking a super-conservative approach and only exposing a minimal set of methods in it (my next comment was going to be to see if you could avoid the newly added members in it :smile:) I''d be happier if we could avoid this particular move if you're ok with that.

manisandro: Ok - so I take I should also revert the `qgsreport*` classes switching from taking a `QgsLayoutDesignerDialog` to a `QgsLayoutDesignerInterface`?

nyalldawson: yes please!

manisandro: Ok should look a bit less scary now.

nyalldawson (approved):

manisandro:
@nyalldawson Since these classes moved to gui were excluded from the python API, do the documentation requirements still apply? If no, is there a way to tell this to the doc coverage test?

nyalldawson:
Unfortunately no -- it's a requirement for code outside of app to be documented fully. 😳

manisandro:
Ok - I'll give it a go, but you might want to suggest better strings here and there.

PR #33847, opened 2020-01-16
set the first default item when no empty field name is allowed

Gustry:
## Description

When a `QgsFieldComboBox` is used with empty field **not** allowed, it should default to the first item.

Before
Capture d’écran de 2020-01-16 12-12-53

After
Peek 16-01-2020 12-17

## Checklist

### DIFF
--- a/src/gui/qgsfieldcombobox.cpp
+++ b/src/gui/qgsfieldcombobox.cpp
@@ -47,6 +47,8 @@ void QgsFieldComboBox::setLayer( QgsMapLayer *layer )
 {
   QgsVectorLayer *vl = qobject_cast<QgsVectorLayer *>( layer );
   mFieldProxyModel->sourceFieldModel()->setLayer( vl );
+
+  setCurrentIndex( allowEmptyFieldName() ? -1 : 0 );
 }
 
 QgsVectorLayer *QgsFieldComboBox::layer() const


rldhont:
@3nids I have requested your review because you are the `QgsFieldCombobox` creator and @nyalldawson because you introduced `allowEmptyFieldName`.

nyalldawson:
Repeat after me "unit tests! Unit tests!" 😜

3nids:
I am not sure if it makes sense to select the first field randomly. While it's unset, it's very visible and would allow having a visual indication it has to be set.
I'd like to get others opinions here.

roya0045:
@3nids I agree with you, it is rare that the IDs are the desired field for a layer. If we had some form of statistics of use per field name (globally or on a layer basis) this might be of some use to use the most frequently used field. 
Otherwise a user might think that the default value is set and ok and will just go through with it regardless. 
I'm -0 as a user.

Gustry:
@nyalldawson I had a (quick) look when I made the PR but I didn't find any test file specific for this class. I will have a look ;-)

@3nids Honestly, I thought about this too. The first field is not often the correct field to choose, I agree.
But I thought it's not the case for any other QCombobox. For example, when you open a Processing algorithm, a "random" QgsMapLayer is already set for you as default.

> it's very visible and would allow having a visual indication it has to be set.

I think my GIF shows the reverse :)
The first field is compulsory while the second one is not. That's why the second one stays empty. For me, it seems better for the user, it makes it clear that the first field *only* is compulsory. 
Without this PR, in the first screenshot, they will have an error when trying to validate the form saying that the first field (the `start attribute`) mustn't be empty.

@nirvn Do you have some inputs for UX?
I can check what has been done in the drill-down forms.

nyalldawson:
@Gustry 

>I had a (quick) look when I made the PR but I didn't find any test file specific for this class. I will have a look ;-)

https://github.com/qgis/QGIS/blob/master/tests/src/python/test_qgsfieldcombobox.py

Gustry:
Ah, I was checking CPP files !
QtCreator didn't show me Python files ;-)

PR #33850, opened 2020-01-16
add flip and autopep8

Gustry:
## Description

@jef-n These packages are required to launch the prepare-commit script.

## Checklist

### DIFF
--- a/scripts/astyle.sh
+++ b/scripts/astyle.sh
@@ -52,7 +52,15 @@ if ! type -p flip >/dev/null; then
 			dos2unix -q -k "$2"
 		}
 	else
-		echo "flip not found" >&2
+    echo "flip not found" >&2
+    echo "Try:"
+    if [[ -f /etc/fedora-release ]]; then
+      echo "  dnf install dos2unix";
+    elif [[ -f /etc/debian_version ]]; then
+      echo "  apt install flip";
+    else
+      echo "  installing flip or dos2unix from your package manager";
+    fi
 		flip() {
 			:
 		}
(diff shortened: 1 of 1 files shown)


m-kuhn:
Hmm, I think they are development dependencies, not build dependencies. Or is committing a required action for building?

Gustry:
Yes, it's development dependencies. So should I just update the readme?

m-kuhn:
I think yes, that's a documentation issue (and/or the script itself could give a meaningful error if the tools are not installed)

Gustry:
Ok, I can update the INSTALL file. There is also `git` in this list so I was wondering.
https://github.com/qgis/QGIS/blob/master/debian/control.in#L75

m-kuhn:
Possibly to get the sha for the about dialog. Not sure.

scripts/astyle.sh
```diff
@@ -53,6 +53,8 @@ if ! type -p flip >/dev/null; then
 		}
 	else
 		echo "flip not found" >&2
+		echo "on debian, try:"
+		echo "apt install flip"
```
m-kuhn: How does that look?

```suggestion
echo "Try:"
if [[ -f /etc/fedora-release ]]; then
    echo "  dnf install dos2unix";
elif [[ -f debian_version ]]; then
    echo "  apt install flip";
else
    echo "  installing flip or dos2unix from your package manager";
fi
```

Gustry: Love it of course ;-)

m-kuhn (approved):
:heart:

PR #33855, opened 2020-01-16
MBTiles raster support in WMS provider

wonder-sk:
I have had this code around for a while and would like to hear what other devs think about it... (I don't mind if it does not get into 3.12.)

This PR adds MBTiles tiled raster map support to WMS provider so that it uses the same code paths
like WMTS or XYZ tiles.

Why move away from GDAL implementation? Here are the advantages of the approach through WMS provider:
- correctly scaling tiles on high dpi display
- better look when not zoomed to native resolution of the tiles. WMS provider uses smooth scaling while GDAL uses nearest neighbor by default.
- map tile showing up while rendering (with gdal it's blank map until everything is loaded)
- possible to use tile scale slider dock widget
- faster - mainly a side effect of loading fewer tiles on high dpi display


src/providers/wms/qgswmsprovider.cpp
```diff
@@ -1280,7 +1311,57 @@ void QgsWmsProvider::setupXyzCapabilities( const QString &uri )
...
+  // Here we just add extra x,y,z query items as an implementation detail (it uses TMS tiling scheme - that's why {-y})
+  mSettings.mBaseUrl += "?x={x}&y={-y}&z={z}";
+
+  QgsRectangle sourceExtent;
+  QgsRectangle sourceExtentWgs84 = mbtilesReader.extent();
+  if ( !sourceExtentWgs84.isNull() )
+  {
+    QgsCoordinateTransform ct( QgsCoordinateReferenceSystem( QStringLiteral( "EPSG:4326" ) ), QgsCoordinateReferenceSystem( mSettings.mCrsId ),
+                               transformContext() );
+    QgsPointXY customTopLeft = ct.transform( QgsPointXY( sourceExtentWgs84.xMinimum(), sourceExtentWgs84.yMaximum() ) );
```
nyalldawson: try/catch

nyalldawson (commented):
code looks good to me, and quite straightforward. I think there's a good argument for this change too. 

+0 to merge, +1 if you add a unit test.

nyalldawson:
(i'd be happy to see this merged for 3.12 too, I think it's a sufficiently low risk change)

wonder-sk:
Thanks Nyall for the fast review - I have added some basic tests and I would merge it for 3.12...

timlinux:
Any notes on how to use this @wonder-sk ? Must we enter a file:/// path to an mbtiles file in the URL

wonder-sk:
sorry @timlinux I have missed the notification....

You don't need to do anything special - it is automatic. If you simply open a mbtiles file from the browser or drag'n'drop it to qgis, the WMS provider will be now used (instead of gdal).

agiudiceandrea:
Hi @wonder-sk, it looks like there are issue when the MBTiles layer's path contains multibytes characters (after unicode U+00FF): please see https://github.com/qgis/QGIS/pull/33855.

agiudiceandrea:
Hi @wonder-sk, it looks like there are issue when the MBTiles layer's path contains multibytes characters (after unicode U+00FF): please see https://github.com/qgis/QGIS/issues/56023.

PR #33856, opened 2020-01-16
include flip and python3-autopep8 in INSTALL

jef-n:
* include flip and python3-autopep8 in INSTALL
* txt2tags 3.4 doesn't support macros (used date & modtime) anymore

superceeds #33850, followup 157a6700cac6


m-kuhn:
Does it supersede #33850?
I think this goes very well together with the current state of #33850.

PR #33868, opened 2020-01-17
Small about screen tweaks

nyalldawson:
- Avoid qt warning when about screen is opened on dev builds- consistent capitalization
- use more room for version information, which currently overflows
the available space allocated for it
- use a https url instead of http


nirvn (approved):

nirvn:
I'd love one day to get rid of the embedded leaflet-based web map in favor of a qgsmapcanvas map. It'd be an interesting experiment (maptips instead of info bubbles, etc.).

Anyways, I'm regressing, this PR's good to merge.

PR #33870, opened 2020-01-17
Save user name in project

nyalldawson:
Supersedes #33843, cleaning up the PR and adding tests

### DIFF
--- a/resources/qgis_global_settings.ini
+++ b/resources/qgis_global_settings.ini
@@ -111,6 +111,10 @@ gps\stop_bits=OneStop
 # will not be automatically populated when a new project is created.
 projects\anonymize_new_projects=false
 
+# Whether or not to anonymize projects when saving
+# If set to 1, then username information will not be stored in saved projects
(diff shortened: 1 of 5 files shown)


resources/qgis_global_settings.ini
```diff
@@ -111,6 +111,10 @@ gps\stop_bits=OneStop
 # will not be automatically populated when a new project is created.
 projects\anonymize_new_projects=false
 
+# Whether or not to anonymize projects when saving
+# If set to 1, then username information will not be stored in saved projects
```
nirvn:
```suggestion
# If set to true, then username information will not be stored in saved projects
```

nirvn (commented):
The only problem I have here is that if someone goes into the project properties, and delete the author string (which default to the user name of the project creator), that person might be left under the wrong impression that his/her user name won't be attached to the project file. 

That said, there are other ways user information leaks into the project file, for e.g. file paths, so maybe that's not a huge problem ATM.

PR #33872, opened 2020-01-17
[FEATURE] Add an option to set the default project file format (qgs/qgz)

3nids:
This adds a new global option to define the preferred project file format

image

sponsored by GEM foundation

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -6665,11 +6665,11 @@ bool QgisApp::fileSave()
     QgsProject::FileFormat defaultProjectFileFormat = settings.enumValue( QStringLiteral( "/qgis/defaultProjectFileFormat" ), QgsProject::FileFormat::Qgz );
     if ( defaultProjectFileFormat == QgsProject::FileFormat::Qgs )
     {
-      exts = qgsExt + ";;" + zipExt;
+      exts = qgsExt + QStringLiteral( ";;" ) + zipExt;
     }
     else
     {
-      exts = zipExt + ";;" + qgsExt;
+      exts = zipExt + QStringLiteral( ";;" ) + qgsExt;
     }
     QString filter;
     QString path = QFileDialog::getSaveFileName(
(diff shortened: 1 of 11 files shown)


src/app/qgisapp.cpp
```diff
@@ -6660,12 +6660,23 @@ bool QgisApp::fileSave()
 
     const QString qgsExt = tr( "QGIS files" ) + " (*.qgs)";
     const QString zipExt = tr( "QGZ files" ) + " (*.qgz)";
+
+    QString exts;
+    QgsProject::FileFormat defaultProjectFileFormat = settings.enumValue( QStringLiteral( "/qgis/defaultProjectFileFormat" ), QgsProject::FileFormat::Qgz );
+    if ( defaultProjectFileFormat == QgsProject::FileFormat::Qgs )
+    {
+      exts = qgsExt + ";;" + zipExt;
```
m-kuhn:
```suggestion
      exts = qgsExt + QStringLiteral( ";;" ) + zipExt;
```

src/app/qgisapp.cpp
```diff
@@ -6660,12 +6660,23 @@ bool QgisApp::fileSave()
...
+    }
+    else
+    {
+      exts = zipExt + ";;" + qgsExt;
```
m-kuhn:
```suggestion
      exts = zipExt + QStringLiteral( ";;" ) + qgsExt;
```

m-kuhn (approved):
Good move, I see a lot of people always switching this manually every time

elpaso:
@m-kuhn I suspect they are mostly developers, but it's a good move anyway. I wonder if the sidecar db could be loaded anyway following a same-basename convention.

3nids:
> I wonder if the sidecar db could be loaded anyway following a same-basename convention

that's a good idea!

m-kuhn:
> I suspect they are mostly developers

They are the GIS experts inside companies / organizations.
With developers I noticed they are too lazy to do extra clicks unless there is an urgent need.

> I wonder if the sidecar db could be loaded anyway following a same-basename convention.

:+1:

PR #33876, opened 2020-01-17
Make it configurable whether to install the sip source files

manisandro:
See https://github.com/qgis/QGIS/pull/33764#issuecomment-575449809


manisandro:
CC @nirvn

PR #33887, opened 2020-01-17
Server: combine opacity values from parameter with existing layer opacities

mhugent:
In QGIS 2.x, layer opacity values passed by the OPACITY - Parameter were combined with existing layer opacities. Currently, they layer opacities are just replaced, which makes it hard for webclients to present meaningful opacity sliders to the user (since the client cannot know the layer opacity in the project file). This PR restores the old behaviour.

### DIFF
--- a/src/server/services/wms/qgswmsrenderer.cpp
+++ b/src/server/services/wms/qgswmsrenderer.cpp
@@ -2708,7 +2708,7 @@ namespace QgsWms
         case QgsMapLayerType::VectorLayer:
         {
           QgsVectorLayer *vl = qobject_cast<QgsVectorLayer *>( layer );
-          vl->setOpacity( opacity / 255. );
+          vl->setOpacity( vl->opacity() * opacity / 255. );
           break;
         }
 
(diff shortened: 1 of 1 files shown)


elpaso:
@mhugent maybe I don't understand, but this will make it impossible to override opacity from a client if the layer's opacity was different than "1"?

I don't think this is the expected behavior.

elpaso:
Besides that to solve your problem, the opacity could be probably better read from the getProjectSettings.

mhugent:
@elpaso : I cannot find the opacity in the GetProjectSettings output. Maybe I'm looking for the wrong tag? Anyway, if it is in the project settings, that would also solve the client problem (the client could then set the slider to the default value).

elpaso:
@mhugent I didn't check if it's already there, but maybe the best solution would be to add it instead of this PR that IMHO is a big -1.

mhugent:
I'll add the parameter to the GetProjectSettings then.

elpaso:
@mhugent thank you!

PR #33891, opened 2020-01-17
Add layer opacity to GetProjectSettings

mhugent:
Adds opacity-Attribute to layer element in the GetProjectSettings-document


elpaso:
@mhugent LGTM, thank you! 

You will probably need to update some sever test control files.

mhugent:
That's a good idea. Added a test where two layers have opacities != 1

PR #33896, opened 2020-01-18
fixed wrong item order in add field dialog, causing wrong focused widget

uclaros:
## Description
When opening the Add Field dialog (New field ctrl+W) the focused widget is the OK button instead of the Name field (first widget in Tab Order) requiring the user to use the mouse to select the name field or press tab until the name field is selected. 
This functionality was not intended but caused by the weird layout item ordering in the ui file (maybe a QT bug?).
The diff looks more complicated than it actually is, it only puts the `<item row=...> ... </item>` blocks in ascending order, which fixes the bug.
I guess this could also be backported.

## Checklist

### DIFF
--- a/src/app/qgsaddattrdialog.cpp
+++ b/src/app/qgsaddattrdialog.cpp
@@ -58,6 +58,8 @@ QgsAddAttrDialog::QgsAddAttrDialog( QgsVectorLayer *vlayer, QWidget *parent, Qt:
 
   if ( mIsShapeFile )
     mNameEdit->setMaxLength( 10 );
+
+  mNameEdit->setFocus();
 }
 
 void QgsAddAttrDialog::mTypeBox_currentIndexChanged( int idx )
--- a/src/ui/qgsaddattrdialogbase.ui
+++ b/src/ui/qgsaddattrdialogbase.ui
@@ -17,29 +17,6 @@
    <bool>true</bool>
   </property>
   <layout class="QGridLayout" name="gridLayout">
-   <item row="2" column="0">
-    <widget class="QLabel" name="textLabel2">
-     <property name="text">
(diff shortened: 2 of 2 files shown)


nyalldawson:
Instead of rearranging the xml (which is fragile, because it's likely to regress as soon as someone edits the form), a better solution would be to explicitly set the focused widget in the dialog code.

E.g. adding something like

    mTypeName->setFocus();

to the end of QgsAddAttrDialog::QgsAddAttrDialog in qgsaddattrdialog.cpp should work.

nyalldawson:
Sorry, that should have been `mNameEdit->setFocus();`

uclaros:
That's what I thought to do at first, but after checking a few dialog constructors saw that the norm is to rely on the .ui file taborder for the initial focus. Then I figured why this dialog was misbehaving and ended up with this. I think that either qtdesigner should not output scrambled layout item order, or that the scrambled order should not matter, as it does not matter on other forms, but those are qt's issues, not qgis'.
Anyway, I could do both for completeness sake, but then should we explicitly define focus on other constructors too just in case their forms are edited and qt misbehaves?

nyalldawson:
> should we explicitly define focus on other constructors too

Yes -- I think it's a good move. Apart from avoiding accidental breakage if qt designer silently moves things around, it also would cause a compilation failure if someone removes the widget. And then they'd have to think about what the logical default widget should be and fix things right then in order to get compilation working...

uclaros:
OK, I'm commiting this now and I'll try to gather dialogs that need explicit setFocus when I got time and issue a PR.

nyalldawson:
Thanks!

nyalldawson (approved):

PR #33897, opened 2020-01-18
Add QgsMapSettings::computeExtentForScale and QgsMapSettings::computeScaleForExtent

manisandro:


src/core/qgsmapsettings.cpp
```diff
@@ -423,6 +423,35 @@ QgsCoordinateTransform QgsMapSettings::layerTransform( const QgsMapLayer *layer
...
+  // Output width in inches
+  double outWIn = outputSize().width() / double( outputDpi() );
+
+  // Desired visible width (honouring scale)
+  double scaledWIn = outWIn * scale;
+
+  if ( mapUnits() == QgsUnitTypes::DistanceDegrees )
+  {
+    // Start with an 1x1 extent around the center
+    QgsRectangle ext( center.x() - 0.5, center.y() - 0.5, center.x() + 0.5, center.y() + 0.5 );
```
nyalldawson: I don't think always using 1 x 1 is appropriate here -- better to use a fraction of the existing width/height?

src/core/qgsmapsettings.cpp
```diff
@@ -423,6 +423,35 @@ QgsCoordinateTransform QgsMapSettings::layerTransform( const QgsMapLayer *layer
...
+    // Get scale at extent, and then scale extent to the desired scale
+    double testScale = mScaleCalculator.calculate( ext, outputSize().width() );
+    ext.scale( scale / testScale );
+    return ext;
+  }
+  // Conversion from inches to mapUnits
```
nyalldawson:
```suggestion
  // Conversion from inches to mapUnits - this is safe to use, because we know here that the map units AREN'T in degrees
```

PR #33898, opened 2020-01-19
[Bugfix][3D][MESH] Fix  color ramp shader settings and flickering triangles for 3D view mesh rendering

vcloarec:
There are two bugs in the 3D view rendering for mesh : 
* an issue with the color ramp shader settings 
* when the camera moves, there are some triangles which suffer of flickering.

This PR fixes those issues.

### DIFF
--- a/src/3d/mesh/qgsmesh3dentity_p.cpp
+++ b/src/3d/mesh/qgsmesh3dentity_p.cpp
@@ -53,6 +53,7 @@ void QgsMesh3dEntity::buildGeometry()
   }
 
   mesh->setGeometry( new QgsMesh3dGeometry( mTriangularMesh,
+                     mMapSettings.origin(),
                      mExtent,
                      mSymbol.verticaleScale(),
                      mesh ) );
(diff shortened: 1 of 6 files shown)


src/3d/mesh/qgsmesh3dentity_p.cpp
```diff
@@ -53,14 +53,15 @@ void QgsMesh3dEntity::buildGeometry()
...
+                     mMapSettings.origin(),
                      mExtent,
                      mSymbol.verticaleScale(),
                      mesh ) );
 
   addComponent( mesh );
 
   Qt3DCore::QTransform *tform = new Qt3DCore::QTransform;
-  tform->setTranslation( QVector3D( float( -mMapSettings.origin().x() ), 0, float( mMapSettings.origin().y() ) ) );
+  tform->setTranslation( QVector3D( 0, 0, 0 ) ) ;
```
wonder-sk: If we don't apply any translation through the transform anymore, we could get rid of the whole qtransform component... (also in the code below)

wonder-sk (approved):

PR #33900, opened 2020-01-19
Update url to doc pages

DelazJ:
should be backported to 3.10


nyalldawson:
Unrelated test failure

PR #33912, opened 2020-01-20
remove verbose wording on layout toolbar buttons (fix #30161)

### LINKED ISSUE #30161: Remove verbose wording on layout toolbar buttons
The wording on the tool buttons in the print layout are a bit verbose IMO.   Could be shorted to `Add picture` because "Add" already implies new and we are in the layout so we already know it's going to do that.

image

alexbruy:
## Description
As discussed in #30161 remove verbose wording on layout toolbar buttons. Fixes #30161.

## Checklist


m-kuhn:
Looks good. But I don't think this should be backported, breaking translations is worse than some extra text on tooltips.

PR #33918, opened 2020-01-20
add "Organize Columns" button to the attribute table toolbar (fix #23397)

### LINKED ISSUE #23397: Hide all column in attribute table: colmun can't be displayed back
Author Name: **Yves Jacolin** (@yjacolin)
Original Redmine Issue: [15471](https://issues.qgis.org/issues/15471)
Affected QGIS version: 3.4.4

---
With the following commit [1] user can hide columns right-clicking in the header. But if the user hide all column (if there is only one column or if all column are hidden), user can't switch back the hidden column as he needs to right-click in the table header to choose the 'Organize columns' entry in the contextual menu but the table header disappears when there is no column to display.

Solution:
1. keep the header empty to allow user to right-click on it
2. allow user to hide/display columns from the Fields tab in the layer properties. 

Y.
[1] https://github.com/qgis/QGIS-Documentation/issues/1111

alexbruy:
## Description
When all columns in the attribute table are hidden (via right-click on the table header) there is no way to unhide them, as table header also gone. Proposed solution is to add "Organize Columns" button to the attribute table toolbar to provide a way to hide/unhide columns without clicking on the table header.

Fixes #23397.

Probably icon need to be changed to something different.

image

## Checklist

### DIFF
--- a/src/app/qgsattributetabledialog.cpp
+++ b/src/app/qgsattributetabledialog.cpp
@@ -54,6 +54,7 @@
 #include "qgsguiutils.h"
 #include "qgsproxyprogresstask.h"
 #include "qgisapp.h"
+#include "qgsorganizetablecolumnsdialog.h"
 
 QgsExpressionContext QgsAttributeTableDialog::createExpressionContext() const
 {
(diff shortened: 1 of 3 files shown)


roya0045:
Great move! Isn't this also a feature?

nyalldawson:
> Isn't this also a feature?

I think it's a bug -- the ability to reorganise columns was a feature added a long time ago, this just prevents a situation where a user operation results in a state that can't be fixed...

alexbruy:
As Nyall said it is more bugfix than a feature. If all columns hidden there is no way to bring them back, except restarting QGIS.

roya0045:
I thought that by adding some GUI elements to control the display of the columns, this could be more of a feature that addresses a lack in the code, rather than a fix.

Saijin-Naib:
Fixes 33466, 27250, 23397

Saijin-Naib:
Why you gotta play me like that, stalebot?

rouault:
Not in my area of expertise, but looks reasonable, so let's merge this.

rouault:
@alexbruy The automatic backport failed (probably my blunder not pushing the right button). I let you issue manual backporting PR

PR #33927, opened 2020-01-20
[BUGFIX] QgsExpression::referencedAttributeIndexes(): only report valid indices

### LINKED ISSUE #33878: Crash when applying wrong filter to OAFeat layer
QGIS master on Windows.

Create a WFS / OGC API  connection to https://pta.spatineo-devops.com/sofp/?f=json

Connect to the service

Choose a layer and double click to open a dialog that allows to create a filter with an expression

Enter a filter/expression that is based on a column name that does not actually exist in the layer, i.e. "test"=1

Click "test" or "ok" > crash

rouault:
If the expression was referencing a non-existing field, -1 was returned in the
result set, which caused later crashed in various providers, including the
Spatialite, Postgres, etc..., due to tried to dereference mFields.at(-1)

Discarding invalid indices is what is also done in
QgsFeatureRequest::OrderBy::usedAttributeIndices()

Fixes #33878

### DIFF
--- a/tests/src/python/test_qgsexpression.py
+++ b/tests/src/python/test_qgsexpression.py
@@ -15,7 +15,7 @@ import qgis  # NOQA
 from qgis.PyQt.QtCore import QVariant
 from qgis.testing import unittest
 from qgis.utils import qgsfunction
-from qgis.core import QgsExpression, QgsFeatureRequest, QgsExpressionContext, NULL
+from qgis.core import QgsExpression, QgsFeatureRequest, QgsFields, QgsExpressionContext, NULL
 
 
 class TestQgsExpressionCustomFunctions(unittest.TestCase):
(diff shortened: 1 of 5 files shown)


tests/src/python/test_qgsexpression.py
```diff
@@ -264,6 +264,12 @@ def testCreateFieldEqualityExpression(self):
         res = '"my\'field" = TRUE'
         self.assertEqual(e.createFieldEqualityExpression(field, value), res)
 
+    def testReferencedAttributeIndexesNonExistingField(self):
+        e = QgsExpression()
+        e.setExpression("foo = 1")
+        self.assertTrue(e.isValid())
+        assert not e.referencedAttributeIndexes(QgsFields())
```
m-kuhn:
```suggestion
        self.assertEqual(len(e.referencedAttributeIndexes(QgsFields())), 0)
```

m-kuhn (approved):

tests/src/python/test_qgsexpression.py
rouault: done

PR #33930, opened 2020-01-20
Oracle provider: fix compiler warnings

rouault:
- Remove use of surfaceType variable that is never read
- Avoid self assignment of nRings variable


nyalldawson (approved):

nyalldawson:
Unrelated test failure (@manisandro you merged a PR which was reporting a build failure on the Windows azure builds -- fix inbound)

manisandro:
@nyalldawson Sorry about that and thanks for fixing!

PR #33938, opened 2020-01-20
[BUGFIX] Fix import of 3D shapefile into spatialite

### LINKED ISSUE #33883: Cannot import 3D vector into Spatialite datasource (works for GPKG)
QGIS 3.10.1/master on Windows/Linux

The following affects both the QGIS Browser and DB Manager.

Create an empty Spatialite datasource add create a connection to it

Pick a 3D shapefile like https://github.com/qgis/QGIS/files/4060140/zone_omi_dati.Bari.zip (unzip it)

Try import the above vector into the Spatialite datasource by drag & drop in the QGIS browser or DB manager

The results will always be a message like "Failed to import layer! Loading of layer failed" and an empty geometryless table is created inside the Spatialite source.

The same operation using a GPKG datasource works fine.

Notes:
using OGR2OGR a direct import into an existing Spatialite datasource also fails, i.e.

```
ogr2ogr -f SQLITE test1.sqlite zone_omi.shp  -update

ERROR 1: Cannot insert feature with geometry of type MULTIPOLYGONZ in column GEOMETRY. Type POLYGONZ expected
```

what works is 

`ogr2ogr -f SQLITE teste1.sqlite zone_omi.shp -update -nlt MULTIPOLYGONZ`

or

`ogr2ogr -f SQLITE teste1.sqlite zone_omi.shp -update -nlt POLYGONZ`

rouault:
Fixes #33883

- QgsOgrProvider::wkbType(): autopromote Z/M/ZM linestring/polygon to multi
- QgsVectorLayerExporter::exportLayer(): remove unneeded hack for shapefile
- QgsSpatiaLiteProvider::createEmptyLayer(): deal with xxxZ geometry types
  (refs #https://github.com/qgis/qgis4.0_api/issues/107)

### DIFF
--- a/src/core/providers/ogr/qgsogrprovider.cpp
+++ b/src/core/providers/ogr/qgsogrprovider.cpp
@@ -1445,7 +1445,8 @@ size_t QgsOgrProvider::layerCount() const
 QgsWkbTypes::Type QgsOgrProvider::wkbType() const
 {
   QgsWkbTypes::Type wkb = QgsOgrUtils::ogrGeometryTypeToQgsWkbType( mOGRGeomType );
-  if ( mGDALDriverName == QLatin1String( "ESRI Shapefile" ) && ( wkb == QgsWkbTypes::LineString || wkb == QgsWkbTypes::Polygon ) )
+  const QgsWkbTypes::Type wkbFlat = QgsWkbTypes::flatType( wkb );
+  if ( mGDALDriverName == QLatin1String( "ESRI Shapefile" ) && ( wkbFlat == QgsWkbTypes::LineString || wkbFlat == QgsWkbTypes::Polygon ) )
   {
     wkb = QgsWkbTypes::multiType( wkb );
   }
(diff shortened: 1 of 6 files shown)


src/core/providers/ogr/qgsogrprovider.cpp
```diff
@@ -1445,7 +1445,8 @@ size_t QgsOgrProvider::layerCount() const
 QgsWkbTypes::Type QgsOgrProvider::wkbType() const
 {
   QgsWkbTypes::Type wkb = QgsOgrUtils::ogrGeometryTypeToQgsWkbType( mOGRGeomType );
-  if ( mGDALDriverName == QLatin1String( "ESRI Shapefile" ) && ( wkb == QgsWkbTypes::LineString || wkb == QgsWkbTypes::Polygon ) )
+  const auto wkbFlat = QgsWkbTypes::flatType( wkb );
```
nyalldawson:
```suggestion
  const QgsWkbTypes::Type wkbFlat = QgsWkbTypes::flatType( wkb );
```

(consensus on previous discussions of this nature was that in QGIS code auto should only be used for complex types such as iterators)

nyalldawson (approved):

src/core/providers/ogr/qgsogrprovider.cpp
rouault: done

PR #33939, opened 2020-01-21
Fix crash in rendering on proj 6 builds

### LINKED ISSUE #33902: QGIS 3.10.x crashes on zoom in/out
QGIS 3.10.2 on Windows
Zooming in and out (fast) a few times on the attached _qgis_crash.qgs_ project closes QGIS instantly, without the crash report screen.  Doing the same on the other project file (_qgis_no_crash.qgs_) on the other hand works fine. The only difference between the two is the fill type of the included two layers. On previous QGIS versions (3.4.5, 3.8.3) both projects worked fine.  Now I'm aware the vector datasets have some validity errors, but this still shouldn't be happening.

[qgis_crash_on_zoom.zip](https://github.com/qgis/QGIS/files/4083795/qgis_crash_on_zoom.zip)

nyalldawson:
Ensure PJ* objects are never reused across threads.

Fixes #33902

### DIFF
--- a/src/core/qgscoordinatereferencesystem_p.h
+++ b/src/core/qgscoordinatereferencesystem_p.h
@@ -98,11 +98,24 @@ class QgsCoordinateReferenceSystemPrivate : public QSharedData
 #if PROJ_VERSION_MAJOR>=6
       QgsReadWriteLocker locker( mProjLock, QgsReadWriteLocker::Write );
 
+      // During destruction of PJ* objects, the errno is set in the underlying
+      // context. Consequently the context attached to the PJ* must still exist !
+      // Which is not necessarily the case currently unfortunately. So
+      // create a temporary dummy context, and attach it to the PJ* before destroying
+      // it
+      PJ_CONTEXT *tmpContext = proj_context_create();
       for ( auto it = mProjObjects.begin(); it != mProjObjects.end(); ++it )
       {
+        proj_assign_context( it.value(), tmpContext );
         proj_destroy( it.value() );
       }
       mProjObjects.clear();
+      if ( mPj )
+      {
+        proj_assign_context( mPj.get(), tmpContext );
+        mPj.reset();
+      }
+      proj_context_destroy( tmpContext );
 #else
       OSRDestroySpatialReference( mCRS );
 #endif
--- a/src/core/qgscoordinatereferencesystem_p.h
+++ b/src/core/qgscoordinatereferencesystem_p.h
@@ -173,7 +173,6 @@ class QgsCoordinateReferenceSystemPrivate : public QSharedData
     OGRSpatialReferenceH mCRS;
 #endif
 
-    QString mValidationHint;
     mutable QString mProj4;
 
     //! True if presence of an inverted axis needs to be recalculated
--- a/src/core/qgscoordinatereferencesystem_p.h
+++ b/src/core/qgscoordinatereferencesystem_p.h
@@ -155,7 +155,7 @@ class QgsCoordinateReferenceSystemPrivate : public QSharedData
 
 #if PROJ_VERSION_MAJOR>=6
     mutable QReadWriteLock mProjLock;
-    mutable QMap < uintptr_t, PJ * > mProjObjects;
+    mutable QMap < PJ_CONTEXT *, PJ * > mProjObjects;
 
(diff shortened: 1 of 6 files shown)


nyalldawson:
@rouault can i get your eyes on this one?

nirvn:
Oh great, this was on my machine the last in-your-face stability issue with windows builds.

src/core/qgscoordinatereferencesystem_p.h
```diff
@@ -124,6 +136,9 @@ class QgsCoordinateReferenceSystemPrivate : public QSharedData
     bool mIsValid = false;
 
 #if PROJ_VERSION_MAJOR>=6
+
+    // this is the "master" proj object, to be used as a template for new proj objects created on different threads ONLY.
+    // Always use threadLocalProjObject() instead of this.
```
rouault: are we sure that mPj will be destroyed before the PJ_CONTEXT* that owns it ?

src/core/qgscoordinatereferencesystem_p.h
```diff
@@ -138,6 +153,49 @@ class QgsCoordinateReferenceSystemPrivate : public QSharedData
     //! Whether this is a coordinate system has inverted axis
     mutable bool mAxisInverted = false;
 
+#if PROJ_VERSION_MAJOR>=6
+    mutable QReadWriteLock mProjLock;
+    mutable QMap < uintptr_t, PJ * > mProjObjects;
```
rouault: any reason to use uintptr_t instead of PJ_CONTEXT* ?

nyalldawson: Nope, good point!

src/core/qgscoordinatereferencesystem_p.h
nyalldawson: This won't have changed in this PR, but I could add mPj to removeObjectsBelongingToCurrentThread for safety?

nyalldawson: @rouault can you get the context which a PJ* belongs to? Does api exist for that?

rouault: > This won't have changed in this PR

agreed

> but I could add mPj to removeObjectsBelongingToCurrentThread for safety?

then QgsCoordinateReferenceSystemPrivate will have a nullptr mPj and be inconsistent, no ? (in the unlikely scenario that a temporary thread would create a CRS and pass it to the main thread)
But, assuming that a QgsCoordinateReferenceSystem object cannot be simultaneously used by several threads at the same time, I'm wondering if instead of that logic of having a per-thread cache of PJ* objects , you couldn't just use proj_assign_context() with the current TLS context ? And at destruction time of mPj, you create a temporary PJ_CONTEXT like done in QgsCoordinateTransformPrivate::freeProj()

rouault: > can you get the context which a PJ* belongs to?

I don't think so. But proj_assign_context() should be cheap enough to do it systematically

nyalldawson: > then QgsCoordinateReferenceSystemPrivate will have a nullptr mPj and be inconsistent, no 

That's fine -- there's already circumstances when this can be nullptr, so it's already guarded against this.

> But, assuming that a QgsCoordinateReferenceSystem object cannot be simultaneously used by several threads at the same time

it can be -- that's the root cause of the issue here.

> And at destruction time of mPj, you create a temporary PJ_CONTEXT like done in QgsCoordinateTransformPrivate::freeProj()

Will do - incoming shortly.

jef-n:
Guess it'll take to morning in my tz for this to land in 3.10…

src/core/qgscoordinatereferencesystem_p.h
nyalldawson: > I don't think so. 

No worries - I can easily workaround this by storing the owner PJ_CONTEXT in QgsCoordinateReferenceSystemPrivate

nyalldawson:
@jef-n I'm nearly done... about to push a new version (still need to backport though)

nyalldawson:
Should be good for merging...

nirvn:
I seen green everywhere :wink:

PR #33944, opened 2020-01-21
Fix geos exceptions during labeling

### LINKED ISSUE #33931: [QGIS 3.11-Master] Qt keeps complaining "WARNING GEOS exception: UnsupportedOperationException: getX called on empty Point"
**Describe the bug**

Qt keeps issuing messages about
>     WARNING    GEOS exception: UnsupportedOperationException: getX called on empty Point
>     WARNING    GEOS exception: UnsupportedOperationException: getY called on empty Point
often when there are some labels get rendered.

**How to Reproduce**

1. set labels to show
2. try zoom in/out or pan your maps
3. notice that the Qt messages keep piling up

or download this sample then try to pan map around
[labeling_bug_perhaps.zip](https://github.com/qgis/QGIS/files/4087696/labeling_bug_perhaps.zip)

**QGIS and OS versions**
QGIS 3.11.0-61 Windows 64-bit Nightly 4a7ea580e9

QGIS version | 3.11.0-Master | 
QGIS code revision | 4a7ea580e9
Compiled against Qt  5.11.2 | Running against Qt 5.11.2
Compiled against GDAL/OGR 3.1.0dev | Running against GDAL/OGR 3.1.0dev
Compiled against GEOS 3.8.0-CAPI-1.13.1 | Running against GEOS 3.8.0-CAPI-1.13.1
Compiled against SQLite 3.29.0 | Running against SQLite 3.29.0
PostgreSQL Client Version 11.5 | SpatiaLite Version 4.3.0
QWT Version 6.1.3 | QScintilla2 Version  2.10.8
Compiled against PROJ 7.0.0 | Running against PROJ Rel. 7.0.0, March 1st, 2020
OS Version Windows 10 (10.0) | This copy of QGIS writes debugging output.
Active python plugins db_manager; processing

nyalldawson:
Fixes #33931

@m-kuhn can you look over this? I've changed a commit you made which was forcing a centroid calculation for all features during labeling...

### DIFF
--- a/src/core/labeling/qgspallabeling.cpp
+++ b/src/core/labeling/qgspallabeling.cpp
@@ -2012,7 +2012,7 @@ void QgsPalLayerSettings::registerFeature( const QgsFeature &f, QgsRenderContext
   bool ddXPos = false, ddYPos = false;
   double quadOffsetX = 0.0, quadOffsetY = 0.0;
   double offsetX = 0.0, offsetY = 0.0;
-  QgsPointXY anchorPosition = geom.centroid().asPoint();
+  QgsPointXY anchorPosition;
 
   //x/y shift in case of alignment
   double xdiff = 0.0;
@@ -2033,6 +2033,7 @@ void QgsPalLayerSettings::registerFeature( const QgsFeature &f, QgsRenderContext
       {
         quadOff = static_cast< QuadrantPosition >( quadInt );
         ddFixedQuad = true;
+        anchorPosition = geom.centroid().asPoint();
       }
     }
   }
--- a/src/core/labeling/qgspallabeling.cpp
+++ b/src/core/labeling/qgspallabeling.cpp
@@ -1957,7 +1957,7 @@ void QgsPalLayerSettings::registerFeature( const QgsFeature &f, QgsRenderContext
   {
     geom = QgsPalLabeling::prepareGeometry( geom, context, ct, doClip ? extentGeom : QgsGeometry(), mergeLines );
 
-    if ( geom.isNull() )
+    if ( geom.isEmpty() )
       return;
   }
   geos_geom_clone = QgsGeos::asGeos( geom );
@@ -3577,7 +3577,7 @@ QgsGeometry QgsPalLabeling::prepareGeometry( const QgsGeometry &geometry, QgsRen
          || ( !qgsDoubleNear( m2p.mapRotation(), 0 ) && !clipGeometry.contains( geom ) ) ) )
   {
     QgsGeometry clipGeom = geom.intersection( clipGeometry ); // creates new geometry
-    if ( clipGeom.isNull() )
+    if ( clipGeom.isEmpty() )
     {
       return QgsGeometry();
     }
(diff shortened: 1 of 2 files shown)


denchat:
Excuse me. Seeing backport bot there, I wonder if this will automatically be backported into 3.10-dev next nightly after the merging?

nyalldawson:
Yes, that's correct

denchat:
QGIS 3.10.2 release 616ad4531b somehow seems not yet to be affected by #33931 at the first place.

I see no warnings on Windows.

QGIS version 3.10.2-A Coruña | QGIS code revision 616ad4531b
Compiled against Qt 5.11.2 | Running against Qt 5.11.2
Compiled against GDAL/OGR 3.0.2 | Running against GDAL/OGR 3.0.2
Compiled against GEOS 3.8.0-CAPI-1.13.1 | Running against GEOS 3.8.0-CAPI-1.13.1
Compiled against SQLite 3.29.0 | Running against SQLite 3.29.0
PostgreSQL Client Version 11.5 | SpatiaLite Version 4.3.0
QWT Version 6.1.3 | QScintilla2 Version 2.10.8
Compiled against PROJ 6.2.1 | Running against PROJ Rel. 6.2.1, November 1st, 2019
OS Version Windows 10 (10.0)
Active python plugins | db_manager; processing

src/core/labeling/qgspallabeling.cpp
```diff
@@ -2033,6 +2033,7 @@ void QgsPalLayerSettings::registerFeature( const QgsFeature &f, QgsRenderContext
       {
         quadOff = static_cast< QuadrantPosition >( quadInt );
         ddFixedQuad = true;
+        anchorPosition = geom.centroid().asPoint();
```
m-kuhn: Shouldn't that be moved out of the isValid() branch? If the expression is not valid for the current feature won't it fallback to a default and still be offsetQuad based?

m-kuhn:
Thanks for having a look @nyalldawson 
This is a new feature for 3.12, no need to backport.

src/core/labeling/qgspallabeling.cpp
nyalldawson: Possibly? I've been looking at where the anchorPosition is actually utilised (which is only in the dxf export code), and the logic is tricky to understand. It's using the anchor position IF a data defined offset quad is set, but without first checking that the data defined value is valid. I don't really understand this, and I'm struggling to understand why we would be using the anchor position in this case at all. Shouldn't the anchor position ONLY be used for fixed position labels OR for ALL labels set to a quadrant mode, regardless of whether or not these use data defined quadrant offsets?

m-kuhn: I'll come up with an improved version of this code.
I found the pal code hard to understand on my end, that's probably the reason I failed to replicate correctly.

Basically, if offset quad (dd or not) or hali/vali is used (which is always dd if I got that right, and should always have a dd fixed position, even though the enabled state on the config page that reflects that could need some :heart: ), the anchor point needs to be present.

m-kuhn: https://github.com/qgis/QGIS/pull/33974

PR #33948, opened 2020-01-21
Deprecation fixes

m-kuhn:
QString::null is almost dead.


nyalldawson:
...And good riddance to that confusing mess

m-kuhn:
Happy to join the club of those who upgraded to recent Qt versions.

PR #33949, opened 2020-01-21
Disable "Replace theme" menu when a map theme is active

DelazJ:
otherwise you end up with map themes of the same content

### DIFF
--- a/src/app/qgsmapthemes.cpp
+++ b/src/app/qgsmapthemes.cpp
@@ -202,6 +202,7 @@ void QgsMapThemes::menuAboutToShow()
   mMenu->insertActions( mMenuSeparator, mMenuPresetActions );
   mReplaceMenu->addActions( mMenuReplaceActions );
 
+  mReplaceMenu->setEnabled( !hasCurrent );
   mActionAddPreset->setEnabled( !hasCurrent );
   mActionRemoveCurrentPreset->setEnabled( hasCurrent );
 }


nirvn:
I'm -1 for this change. There might be valid scenarios where a user wants to overwrite an existing theme with another one as a starting point to then further customize things.

DelazJ:
Why not directly replace the target theme after s.he begins making the changes over the starting theme? Any little change is enough.

DelazJ:
Also if the "Add theme" button is disabled whenever a theme is active, hence forbidding creation of another theme with the same content, it does not make sense (to me) to allow replacement of a theme with that same content. 
Better enable the "Add theme" instead of relying on such a workaround. but I'm not sure it's desirable to have map themes not unique.

DelazJ:
Hi @roya0045 sorry for the late reply. I'm not sure it's desirable to have many map themes pointing to the same "values". What would be the use case and the benefit? 
And I might be wrong but I think there could be a need (or there's a plan/wish?) to find a map theme based on its content. And allowing multiple map themes with same content wouldn't help.

roya0045:
Why is creating duplicate themes not desirable? It has no side effect, its not optimal either. But I feel that there is no need to limit such thing when there are no adverse effects.

DelazJ:
What would be the use case? What would that address? I can't think of any case someone would need to have two identical map themes, unless he wants to get trouble and confusion in the long run. BTW matching from the map canvas always checks the first (alphabetically?) map theme in the drop-down menu, ignoring any duplicate.

PR #33950, opened 2020-01-21
[feature] Allow renaming of the current map theme

### LINKED ISSUE #26563: Rename Map Themes directly in Map Theme Dropdown Menu
Author Name: **Rob Willson** (Rob Willson)
Original Redmine Issue: [18675](https://issues.qgis.org/issues/18675)

Redmine category:map_tools
Assignee: Denis Rouzaud

---
Map themes (formerly presets) are one of the most valuable and useful tools in QGIS but it is hard to organize or re-organize them when you can't rename them by just right clicking them.

Request that we be able to rename them by right clicking just like you can now do in the Favorites Menu in the browser (i.e., Rename Favorite) and you can do when saving or renaming layer styles. 

a bonus would be if you could manually reorder them as it eventually becomes difficult to come up with unique starting characters to change the sort order, currently, I have to use combinations of date time stamps, exclamation points, 0.01, 0.02 etc and keep the names in a separate text editor to be able to paste into the new map themes to keep the same format when I make changes to the company startup template.

DelazJ:
This is an early PR to add ability to rename map theme from the map theme drop-down menu (fixes #26563).
I can't say that the code is the cleanest you could have since there are some coding basis I do not have (and I expect it gets improved before merge) but from a user point of view the feature works. 

renametheme

### DIFF
--- a/src/core/layout/qgslayoutitemmap.cpp
+++ b/src/core/layout/qgslayoutitemmap.cpp
@@ -1778,6 +1778,17 @@ void QgsLayoutItemMap::mapThemeChanged( const QString &theme )
     mCachedLayerStyleOverridesPresetName.clear(); // force cache regeneration at next redraw
 }
 
+void QgsLayoutItemMap::currentMapThemeRenamed( const QString &theme, const QString &newTheme )
+{
+  if ( mFollowVisibilityPreset )
+  {
+    if ( theme == mFollowVisibilityPresetName )
+    {
+      mFollowVisibilityPresetName = newTheme;
+    }
+  }
+}
+
 void QgsLayoutItemMap::connectUpdateSlot()
 {
   //connect signal from layer registry to update in case of new or deleted layers
@@ -1820,6 +1831,7 @@ void QgsLayoutItemMap::connectUpdateSlot()
   } );
 
   connect( project->mapThemeCollection(), &QgsMapThemeCollection::mapThemeChanged, this, &QgsLayoutItemMap::mapThemeChanged );
+  connect( project->mapThemeCollection(), &QgsMapThemeCollection::mapThemeRenamed, this, &QgsLayoutItemMap::currentMapThemeRenamed );
 }
 
 QTransform QgsLayoutItemMap::layoutToMapCoordsTransform() const
--- a/src/core/layout/qgslayoutitemmap.cpp
+++ b/src/core/layout/qgslayoutitemmap.cpp
@@ -1780,12 +1780,9 @@ void QgsLayoutItemMap::mapThemeChanged( const QString &theme )
 
 void QgsLayoutItemMap::currentMapThemeRenamed( const QString &theme, const QString &newTheme )
 {
-  if ( mFollowVisibilityPreset )
+  if ( theme == mFollowVisibilityPresetName )
   {
-    if ( theme == mFollowVisibilityPresetName )
-    {
-      mFollowVisibilityPresetName = newTheme;
-    }
+    mFollowVisibilityPresetName = newTheme;
   }
 }
 
--- a/src/core/layout/qgslayoutitemmap.h
+++ b/src/core/layout/qgslayoutitemmap.h
@@ -622,7 +622,7 @@ class CORE_EXPORT QgsLayoutItemMap : public QgsLayoutItem
 
     void mapThemeChanged( const QString &theme );
 
-    //! Applies current visibility presets to the renamed map theme
+    //! Renames the active map theme called \a theme to \a newTheme
     void currentMapThemeRenamed( const QString &theme, const QString &newTheme );
 
     //! Create cache image
--- a/src/core/layout/qgslayoutitemmap.h
+++ b/src/core/layout/qgslayoutitemmap.h
@@ -622,6 +622,9 @@ class CORE_EXPORT QgsLayoutItemMap : public QgsLayoutItem
 
     void mapThemeChanged( const QString &theme );
 
+    //! Applies current visibility presets to the renamed map theme
+    void currentMapThemeRenamed( const QString &theme, const QString &newTheme );
+
     //! Create cache image
     void recreateCachedImageInBackground();
 
--- a/src/core/qgsmapthemecollection.cpp
+++ b/src/core/qgsmapthemecollection.cpp
@@ -281,6 +281,19 @@ void QgsMapThemeCollection::update( const QString &name, const MapThemeRecord &s
   emit mapThemesChanged();
 }
 
+
+void QgsMapThemeCollection::renameMapTheme( const QString &name,  const QString &newName ) //, const MapThemeRecord &state )
+{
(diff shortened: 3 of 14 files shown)


src/core/layout/qgslayoutitemmap.cpp
```diff
@@ -1767,6 +1767,17 @@ void QgsLayoutItemMap::mapThemeChanged( const QString &theme )
     mCachedLayerStyleOverridesPresetName.clear(); // force cache regeneration at next redraw
 }
 
+void QgsLayoutItemMap::currentMapThemeRenamed( const QString &theme, const QString &newTheme )
+{
+  if ( mFollowVisibilityPreset )
```
nyalldawson: Not sure that this outer condition is required -- I think we'd still want to update the theme name, even if it's been temporarily disabled for the map

src/core/qgsmapthemecollection.cpp
```diff
@@ -276,6 +276,19 @@ void QgsMapThemeCollection::update( const QString &name, const MapThemeRecord &s
   emit mapThemesChanged();
 }
 
+
+void QgsMapThemeCollection::renameMapTheme( const QString &name,  const QString &newName ) //, const MapThemeRecord &state )
```
nyalldawson:
```suggestion
void QgsMapThemeCollection::renameMapTheme( const QString &name,  const QString &newName )
```

src/core/qgsmapthemecollection.cpp
```diff
@@ -276,6 +276,19 @@ void QgsMapThemeCollection::update( const QString &name, const MapThemeRecord &s
...
+{
+  if ( !mMapThemes.contains( name ) || mMapThemes.contains( newName ) )
+    return;
```
nyalldawson: I'd make this function return a bool, and return false here

src/core/qgsmapthemecollection.h
```diff
@@ -251,12 +251,18 @@ class CORE_EXPORT QgsMapThemeCollection : public QObject
...
-     * Remove an existing map theme from collection.
+     * Removes an existing map theme from collection.
      * \since QGIS 3.0
      */
     void removeMapTheme( const QString &name );
 
-    //! Remove all map themes from the collection.
+    /**
+     * Renames an existing map theme in the collection.
+     * \since QGIS 3.14
```
nyalldawson:
```suggestion
     * \since QGIS 3.12
```

src/core/qgsmapthemecollection.h
```diff
@@ -367,6 +373,12 @@ class CORE_EXPORT QgsMapThemeCollection : public QObject
      */
     void mapThemeChanged( const QString &theme );
 
+    /**
+     * Emitted when a map theme within the collection is renamed.
+     * \since QGIS 3.14
```
nyalldawson:
```suggestion
     * \since QGIS 3.12
```

nyalldawson:
Looking good so far!

nyalldawson:
Personally I'd classify this as a bug fix, and merge for 3.12.

src/core/layout/qgslayoutitemmap.cpp
DelazJ: I was pretty sure (months ago) that without this condition, the followed map theme renaming was not propagated here. Tested again (with and without) and the renaming is always done so I removed.

src/core/qgsmapthemecollection.cpp
DelazJ: Done but I'm not sure of what it implies nor whether what i did is what you expected

src/core/qgsmapthemecollection.h
```diff
@@ -251,12 +251,18 @@ class CORE_EXPORT QgsMapThemeCollection : public QObject
...
     /**
-     * Remove an existing map theme from collection.
+     * Removes an existing map theme from collection.
      * \since QGIS 3.0
      */
     void removeMapTheme( const QString &name );
 
-    //! Remove all map themes from the collection.
+    /**
+     * Retuns whether a map theme is renamed.
```
nyalldawson:
```suggestion
     * Renames the existing map theme called \a name to \a newName.
     * Returns TRUE if the rename was successful, or FALSE if it failed (e.g. due to a duplicate name for \a newName).
```

DelazJ:
Thanks for the inputs @nyalldawson

src/core/qgsmapthemecollection.cpp
nyalldawson: Looks good!

DelazJ:
Any help here, please? I don't want it to get stale.
That said, I think I can do the 3d part if #34369 is merged (but not sure of the "backportability" at that moment)

DelazJ:
May I?

src/gui/qgsmapcanvas.h
```diff
@@ -827,6 +827,7 @@ class GUI_EXPORT QgsMapCanvas : public QGraphicsView
     void refreshMap();
 
     void mapThemeChanged( const QString &theme );
+    void mapThemeRenamed( const QString &theme, const QString &newTheme );
```
m-kuhn, line 830: Can you add a docstring for this method?

src/core/layout/qgslayoutitemmap.h
```diff
@@ -622,6 +622,9 @@ class CORE_EXPORT QgsLayoutItemMap : public QgsLayoutItem
 
     void mapThemeChanged( const QString &theme );
 
+    //! Applies current visibility presets to the renamed map theme
```
m-kuhn, line 625: since 3.14

m-kuhn: Actually, I can't really follow the meaning of this docstring.

DelazJ: Hi @m-kuhn 
Sorry I only noticed your reply few days ago.
Indeed, now that i reread... From memories, this ensures that if a theme is being used and is renamed, it will not lose the index in the combobox index and break rendering. Would this be better? 

> Renames the active map theme called \a theme to \a newTheme.

The same could be added to the others (2D and 3D map views).
Do we add "since ..." in private slots?

m-kuhn: Yes, this doc string sounds perfect.

No, `since` is only for public members (and protected or virtual private, but that doesn't apply here)

alexbruy:
@DelazJ would you mind to address remaining comments?

DelazJ:
They should already be addressed. Anything left?

alexbruy:
There is one missed I think, adding docstring to `void mapThemeRenamed()`

DelazJ:
No, GitHub artefact. It's in the latest commit

alexbruy:
Ah, sorry.

DelazJ:
np

m-kuhn (approved):
So, let's go?

DelazJ:
I'd say "yes!"

PR #33952, opened 2020-01-21
[server] do not drop layers with duplicate names

### LINKED ISSUE #33928: WMS doesn't render layer with duplicate names
When several layers with equal names are added in a project, only one of them is rendered, the other ones are silently discarded by QGIS Server.

This often happens when adding the same layer multiple times with different geometries because the underlying table is of type GeometryCollection.

m-kuhn:
When multiple layers with the same name are present, still render all of them.

Fix #33928

# Documentation

Update the QGIS server documentation that by default layer names will be used to expose layers through WMS. If multiple layers have the same name, they will be merged to a single WMS layer and cannot be requested individually and project maintainers need to take care of these layers being compatible (TODO: what does compatibility mean: CRS, fields, ...?).

Alternatively there is an option to use layer ids in the server properties page which will offer mutiple layers with the same names as individual layers.


rldhont (approved):

rldhont:
@m-kuhn I think we needs to enhance the documentation to explicitly explain the way QGIS Server get layers from project.

I can be in favor of backport to 3.10. @elpaso what is your point of view ?

elpaso (approved):
ok for backport too

m-kuhn:
@3nids interesting things going on here, additional labels trigger the doc bot again.

PR #33953, opened 2020-01-21
Fix slow PG raster SRID identify

### LINKED ISSUE #33885: Qgis scans raster tables on connection to postgis
I connect to DB with huge raster tables. It takes about 5 minutes to load list of tables

It executes following queries:
`SELECT 0, array_agg(DISTINCT 'RASTER:' || ST_SRID( "rast" )) FROM "public"."hrsl_population_raster"`

image

image

elpaso:
Reading all raster rows is too slow, by querying
raster_columns we can retrieve the SRID directly.

Fixes #33885

I'm not sure this is really a bugfix or just an optimization.

### DIFF
--- a/src/providers/postgres/qgspostgresconn.cpp
+++ b/src/providers/postgres/qgspostgresconn.cpp
@@ -1686,12 +1686,14 @@ void QgsPostgresConn::retrieveLayerTypes( QVector<QgsPostgresLayerProperty *> &l
 
     if ( layerProperty.isRaster )
     {
-      QString sql = QStringLiteral( "SELECT %3, "
-                                    "array_agg(DISTINCT ST_SRID( %1 ) || ':RASTER')"
-                                    " FROM %2" )
-                    .arg( quotedIdentifier( layerProperty.geometryColName ) )
-                    .arg( table )
-                    .arg( i - 1 );
+      const QString sql = QStringLiteral( "SELECT %1, "
+                                          "array_agg( srid || ':RASTER') "
+                                          "FROM raster_columns "
+                                          "WHERE r_raster_column = %2 AND r_table_schema = %3 AND r_table_name = %4" )
+                          .arg( i - 1 )
+                          .arg( quotedValue( layerProperty.geometryColName ) )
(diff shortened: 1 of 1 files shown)


src/providers/postgres/qgspostgresconn.cpp
```diff
@@ -1686,12 +1686,14 @@ void QgsPostgresConn::retrieveLayerTypes( QVector<QgsPostgresLayerProperty *> &l
...
     if ( layerProperty.isRaster )
     {
-      QString sql = QStringLiteral( "SELECT %3, "
-                                    "array_agg(DISTINCT ST_SRID( %1 ) || ':RASTER')"
-                                    " FROM %2" )
-                    .arg( quotedIdentifier( layerProperty.geometryColName ) )
-                    .arg( table )
-                    .arg( i - 1 );
+      const QString sql = QStringLiteral( "SELECT %1, "
+                                          "array_agg( srid || ':RASTER') "
```
luipir: take into account that potencially srid value can be different from the real one of the raster

luipir (approved):

src/providers/postgres/qgspostgresconn.cpp
elpaso: I know it can happen, but fetching all the (potentially thousands) raster tiles just to get the SRID really kills performances. And I don't think we can handle the case of multiple raster SRIDs in the same table, at least not with current provider implementations (both GDAL and native).

strk:
I see the non-raster case has a few optimizaitons:

 1. useEstimatedMetadata (would do what you're doing here)
 2. Limits scan to an upper bound GEOM_TYPE_SELECT_LIMIT
 3. Adds a WHERE clause, if one is defined

Should anyone apply to raster as well ? Better be consistent...

strk:
Oh, an another optimization in the vector code is that if the SRID in the properties is already known, it is not even queried ! (see the `srid  == std::numeric_limits<int>::min()` line...)

strk (approved):
Excellent, thanks !

( I wonder if we could reduce code in general, but I guess for this round this is EXCELLENT )

PR #33955, opened 2020-01-21
Makes layer prop. info text a bit smaller

### LINKED ISSUE #33862: Layer Properties Information tab - formatting problems
In the Layer Properties/Information dialog (QGIS 3.10 and 3.11, Windows 10),
* CRS information is cut off
* Can't copy the full CRS string (related to item above)
* As the dialog is resized, labels column wraps too early (see pic)
* Text is too big (in my opinion)

image

Compare this to QGIS 3.4
image

**How to Reproduce**
1. Download one of U.S. NOAA's RNCs (geo-referenced charts) in BSB format. I used [this one](https://charts.noaa.gov/RNCs/Agreement.shtml?12323).
2. Unzip.
3. Drag/drop the file `12323_1.KAP` onto QGIS from the appropriate sub directory.
4. Double-click the layer in the `Layers` panel.
5. Click on the `Information` tab on the left.

**Additional context**
For this specific file, I can go to `Settings > Custom Projections > Format: WKT` to get the full string, but it would be nicer if I can see (and copy) the string in the Properties dialog.

elpaso:
... also makes first cell column a bit larger

Fixes #33862 (partial fix)

raster_properties

### DIFF
--- a/src/core/qgsapplication.cpp
+++ b/src/core/qgsapplication.cpp
@@ -1437,6 +1437,7 @@ QString QgsApplication::reportStyleSheet( QgsApplication::StyleSheetType styleSh
                    "   border-collapse: collapse;"
                    "   table-layout:fixed;"
                    "   width: 100% !important;"
+                   "   font-size: 90%;"
                    "}"
                    // Override
                    "h1 { "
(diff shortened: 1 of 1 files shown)


nyalldawson:
We'll need to test this on windows too - in the past when we've made font sizes smaller they can become unreadable on windows (since it uses smaller fonts by default). But we can easily test that after merge.

elpaso:
Looks good on windows too.

Issue #32913, opened 2019-11-18
Legend causing a crash


chrisodell:
I use a WFS service. https://mapprod3.environment.nsw.gov.au/arcgis/services/Planning/EPI_Primary_Planning_Layers/MapServer/WFSServer?request=GetCapabilities&service=WFS 

It works fine until I go into my layout and add a legend. Once the legend is added - the file crashes. Everytime. 

Is it the WFS or QGIS causing the problem.

QGIS version
3.8.1-Zanzibar
QGIS code revision
dcd95cc648
Compiled against Qt
5.11.2
Running against Qt
5.11.2
Compiled against GDAL/OGR
2.4.1
Running against GDAL/OGR
2.4.1
Compiled against GEOS
3.7.2-CAPI-1.11.0
Running against GEOS
3.7.2-CAPI-1.11.0 b55d2125
PostgreSQL Client Version
9.2.4
SpatiaLite Version
4.3.0
QWT Version
6.1.3
QScintilla2 Version
2.10.8
Compiled against PROJ
5.2.0
Running against PROJ
Rel. 5.2.0, September 15th, 2018
OS Version
Windows 10 (10.0)

Stack Trace

QRasterPaintEngine::transformChanged :
QPainter::translate :
QGraphicsScene::drawForeground :
QGraphicsScene::dragMoveEvent :
QGraphicsItem::dragMoveEvent :
QGraphicsEffectSource::draw :
QgsLayoutEffect::draw :
QGraphicsScene::drawItems :
QGraphicsScene::drawItems :
QGraphicsView::paintEvent :
QgsLayoutView::paintEvent :
QWidget::event :
QFrame::event :
QGraphicsView::viewportEvent :
QCoreApplicationPrivate::sendThroughObjectEventFilters :
QApplicationPrivate::notify_helper :
QApplication::notify :
QgsApplication::notify :
QCoreApplication::notifyInternal2 :
QWidgetPrivate::drawWidget :
QWidgetPrivate::paintSiblingsRecursive :
QWidgetPrivate::drawWidget :
QWidgetPrivate::paintSiblingsRecursive :
Qd OS versions**WidgetPrivate::drawWidget :
QWidgetPrivate::paintSiblingsRecursive :
QWidgetPrivate::drawWidget :
QWidgetPrivate::paintSiblingsRecursive :
QWidgetPrivate::paintSiblingsRecursive :
QWidgetPrivate::drawWidget :
QApplication::windowIcon :
[...]

PR #33983, opened 2020-01-22
OAPIF: fix support of StringList fields

### LINKED ISSUE #33758: QGIS WFS / OGC API – Features not showing list typed attributes correctly
**Describe the bug**

QGIS WFS / OGC API – Features are not showing list typed attributes (in Identify Results) correctly. Instead of a list it returns NULL.

**How to Reproduce**

1. Add a WFS / OGC API – Features: `https://pta.spatineo-devops.com/sofp/?f=json`
2. Add Administrative units 4500k (sea area excluded) (2019) -layer
3. Use filter `"levelName" = 'maakunta'`
4. Select any polygon with Identify Features tool
5. Notice how upperLevelUnit and/or lowerLevelUnit are NULL
6. To verify the data is there add the same data as Vector Layer using Layer --> Add Layer --> Add Vector Layer... (file) with Vector Dataset(s) as `https://pta.spatineo-devops.com/sofp/collections/au_kansallinen_meri_4500k_wgs84/2019/items?f=json&levelName=maakunta&limit=100` --> Add
8. Select any polygon with Identify Features tool
9. Notice how upperLevelUnit and/or lowerLevelUnit has list data

(Any other API Features data will do if there is a list typed attributes.)

**QGIS and OS versions**

QGIS version
3.11.0-Master
QGIS code revision
12c410c3f4
Compiled against Qt
5.11.2
Running against Qt
5.11.2
Compiled against GDAL/OGR
3.1.0dev
Running against GDAL/OGR
3.1.0dev
Compiled against GEOS
3.8.0-CAPI-1.13.1
Running against GEOS
3.8.0-CAPI-1.13.1 
Compiled against SQLite
3.29.0
Running against SQLite
3.29.0
PostgreSQL Client Version
11.5
SpatiaLite Version
4.3.0
QWT Version
6.1.3
QScintilla2 Version
2.10.8
Compiled against PROJ
7.0.0
Running against PROJ
Rel. 7.0.0, March 1st, 2020
OS Version
Windows 10 (10.0)
This copy of QGIS writes debugging output.
Active python plugins
buffer_clip; 
MetaSearch; 
pluginbuilder3; 
plugin_reloader; 
db_manager; 
processing

rouault:
Fixes #33758

### DIFF
--- a/src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
+++ b/src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
@@ -20,6 +20,7 @@
 #include "qgsfeedback.h"
 #include "qgslogger.h"
 #include "qgsmessagelog.h"
+#include "qgswfsutils.h" // for isCompatibleType()
 
 #include <QDataStream>
 #include <QDir>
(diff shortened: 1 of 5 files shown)


nyalldawson:
Lgtm. Are you working on qgis bug fixes currently? Because if so, there's a big issue in the wfs background feature iterator which is causing crashes (https://github.com/qgis/QGIS/issues/32913). It's caused by the use of an event loop in the iterator class, which isn't safe to do here when features are fetched on the main thread...

rouault:
>  It's caused by the use of an event loop in the iterator class, which isn't safe to do here when features are fetched on the main thread.

Any idea how to address this ?

nyalldawson:
It's not a simple change. I think you need to ensure the fetching always occurs on a background thread, and use a semaphore to block the iterator while it waits for incoming features from the fetching thread. (That's why I asked if you are doing a qgis bug fixing stint!)

rouault:
> if you are doing a qgis bug fixing stint!

yes I'm in it. Thanks for the hints. I'll try to look at that. I think there were already similar fixes in the past in the network related code so perhaps I can find some inspiration from that

nyalldawson:
Yes possibly you could adapt what's done in QgsBlockingNetworkRequest

PR #33984, opened 2020-01-22
remove duplicate F7 shortcut used to open styling dock (fix #26696, #26697)

alexbruy:
## Description
Styling dock can be opened using global `Ctrl+3` shortcut and using `F7` shortcut which works only when Layer tree is visible. This PR removes duplicate F7 shortcut.

Fixes #26696 and #26697.

Not sure about backport to 3.10, probably for LTR better to keep both shortcuts and avoid documentation changes.

## Checklist


nyalldawson:
I'm -0 on this -- I don't see the harm in the duplicate shortcut, and the F7 shortcut here is VERY widely advertised

PR #33987, opened 2020-01-22
Fix scale visibility in style read/write

### LINKED ISSUE #33840: QgsVectorLayer readStyle does not read scale based visibility
[QgsVectorLayer::readStyle](https://github.com/qgis/QGIS/blob/8c8d48bb8fe82ef8654dd65233e3265f4c3e524e/src/core/qgsvectorlayer.cpp#L2411) in `Rendering` category does not read scale based visibility setting.

Steps to reproduce:

- run [attached pyqgis file](https://github.com/qgis/QGIS/files/4069841/test_script.zip) (contains QML style data)
  - reads a QML style string
  - creates and loads a memory layer
  - read the style into the `QgsVectorLayer`
- open the created layer properties
- see that the `Scale based visibility` setting in the `Rendering` section has not been set

elpaso:
Fixes #33840


src/core/qgsvectorlayer.cpp
```diff
@@ -2422,6 +2422,20 @@ bool QgsVectorLayer::readStyle( const QDomNode &node, QString &errorMessage,
         QDomElement e = layerOpacityNode.toElement();
         setOpacity( e.text().toDouble() );
       }
+
+      const bool hasScaleBasedVisibiliy { node.attributes().namedItem( QStringLiteral( "hasScaleBasedVisibilityFlag" ) ).nodeValue() == '1' };
+      setScaleBasedVisibility( hasScaleBasedVisibiliy );
+      bool ok;
+      const double maxScale { node.attributes().namedItem( QStringLiteral( "maxScale" ) ).nodeValue().toDouble( &ok ) };
```
luipir: maybe safer to check returned namedItem isNull node instead delegate a side effect of nodeValue return to empty string?

elpaso: No need to. Converting an empty string to double sets ok to false.

luipir: I know that! but what else if the api change and nodeValue return None (ore something else) instead of string? Because node isNull it's not garanteed that nodeValue should be empty string in the future.

PR #33990, opened 2020-01-22
Fixes crash when displaying relation editor

troopa81:
## Description

This PR fixes several crashes related to relation editor when attribute table dialog is called
- From the identify results dialog
- From python iface
- From multi edits on selected feature
- From fill ring map tool

The issue is due of uncomplete QgsAttributeEditorContext, so I add a method in QgisApp to create proper one.

Relation editor is able now to create directly related feature geometry in the canvas, which means that the QgsAttributeDialog must be non modal. So I changed the QgsAttributeDialog flags. I'm not sur it's OK regarding the UI, that why the PR is WIP.

For now, creating geometry directly from relation editor when filling a ring is disable because the code should be modified (remove dialog->exec, create the ring and the new feature at the same time when dialog is closed)

## Checklist

### DIFF
--- a/src/app/qgisapp.h
+++ b/src/app/qgisapp.h
@@ -1108,6 +1108,12 @@ class APP_EXPORT QgisApp : public QMainWindow, private Ui::MainWindow
     //! Create a new spatial bookmark
     void newBookmark( bool inProject = false );
 
+    /**
+     * Creates a default attribute editor context using the main map canvas and the main edit tools and message bar
+     * \since QGIS 3.12
+     */
+    QgsAttributeEditorContext createAttributeEditorContext();
+
   protected:
 
     //! Handle state changes (WindowTitleChange)
--- a/src/app/qgsfeatureaction.cpp
+++ b/src/app/qgsfeatureaction.cpp
@@ -53,7 +53,7 @@ QgsAttributeDialog *QgsFeatureAction::newDialog( bool cloneFeature )
 {
   QgsFeature *f = cloneFeature ? new QgsFeature( *mFeature ) : mFeature;
 
-  QgsAttributeEditorContext context;
+  QgsAttributeEditorContext context( QgisApp::instance()->createAttributeEditorContext() );
 
   QgsDistanceArea myDa;
 
@@ -61,14 +61,11 @@ QgsAttributeDialog *QgsFeatureAction::newDialog( bool cloneFeature )
   myDa.setEllipsoid( QgsProject::instance()->ellipsoid() );
 
   context.setDistanceArea( myDa );
-  context.setVectorLayerTools( QgisApp::instance()->vectorLayerTools() );
-  context.setMapCanvas( QgisApp::instance()->mapCanvas() );
-  context.setCadDockWidget( QgisApp::instance()->cadDockWidget() );
-  context.setMainMessageBar( QgisApp::instance()->messageBar() );
   context.setFormMode( QgsAttributeEditorContext::StandaloneDialog );
 
   QgsAttributeDialog *dialog = new QgsAttributeDialog( mLayer, f, cloneFeature, parentWidget(), true, context );
   dialog->setWindowFlags( dialog->windowFlags() | Qt::Tool );
+
   dialog->setObjectName( QStringLiteral( "featureactiondlg:%1:%2" ).arg( mLayer->id() ).arg( f->id() ) );
 
   QList<QgsAction> actions = mLayer->actions()->actions( QStringLiteral( "Feature" ) );
--- a/python/gui/auto_generated/editorwidgets/qgsrelationreferencewidget.sip.in
+++ b/python/gui/auto_generated/editorwidgets/qgsrelationreferencewidget.sip.in
@@ -77,6 +77,14 @@ Returns the related feature foreign keys
 %End
 
     void setEditorContext( const QgsAttributeEditorContext &context, QgsMapCanvas *canvas, QgsMessageBar *messageBar );
+%Docstring
+Sets the editor ``context``
+
+.. note::
+
+   if context cadDockWidget is null, it won't be possible to digitize
+   the geometry of a referenced feature from this widget
+%End
 
     bool embedForm();
 %Docstring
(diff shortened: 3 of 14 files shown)


src/gui/qgsattributedialog.cpp
```diff
@@ -23,7 +23,7 @@
 #include "qgssettings.h"
 
 QgsAttributeDialog::QgsAttributeDialog( QgsVectorLayer *vl, QgsFeature *thepFeature, bool featureOwner, QWidget *parent, bool showDialogButtons, const QgsAttributeEditorContext &context )
-  : QDialog( parent )
+  : QDialog( parent, Qt::Window )
```
m-kuhn: Out of curiosity, what's that for?

troopa81: to make it non modal (like in [QgsAttributeTableDialog](https://github.com/troopa81/QGIS/blob/fix_relationeditor_crash_context/src/app/qgsattributetabledialog.h#L56)) so the user can still click on the canvas.

troopa81:
I removed the WIP tag.

> Relation editor is able now to create directly related feature geometry in the canvas, which means that the QgsAttributeDialog must be non modal. So I changed the QgsAttributeDialog flags. I'm not sure it's OK regarding the UI, that why the PR is WIP.

@m-kuhn Regarding this comment, do you think it's OK to make the QgsAttributeDialog non modal? This dialog is called from different place in the application. I made some tests and it looks like OK everytime.

troopa81:
@elpaso I remove the q_assert on cadDockWidget that you add for a previous fix because I enable/disable geometry feature creation from relationeditor according to wether or not caddockwidget has been set. is that OK for you?

src/app/qgisapp.h
```diff
@@ -1102,6 +1102,12 @@ class APP_EXPORT QgisApp : public QMainWindow, private Ui::MainWindow
     //! Create a new spatial bookmark
     void newBookmark( bool inProject = false );
 
+    /**
+     * Create a default attribute editor context
```
```suggestion
     * Creates a default attribute editor context using the main map canvas and the main edit tools
```

src/gui/qgsattributedialog.cpp
m-kuhn: What about using `Qt::Tool`?

Using this approach will likely make it disappear behind the main window (leaving people with 20 open attribute dialogs - just like they now tend to open the attribute table 20 times)

elpaso:
@troopa81 ok to remove the assert but we need to do something to prevent a misuse of the API, the checks that you have added are ok, perhaps you should also add a note/warning to the getters documentation to make sure a developer knows that the returned value might be a nullptr and it must be checked before use.

src/gui/qgsattributedialog.cpp
troopa81: > Using this approach will likely make it disappear behind the main window 

That's the exact reason of this modification. It's because of this [PR](https://github.com/qgis/QGIS/pull/32528). I want to be able to raise the canvas so user can digitize a feature and then raise back the attribute dialog. 

> (leaving people with 20 open attribute dialogs - just like they now tend to open the attribute table 20 times)

I understand your concern, but I can't see a better way to solve this UI issue. Maybe, we could let the window on top of the canvas but the user can still digitize in the canvas (he has to move the window on the side though) but it's not really user friendly.

troopa81:
@elpaso I add a warning message on cadDockWidget only, I don't see any reason for the other pointer to be null.

src/gui/qgsattributedialog.cpp
m-kuhn: A better approach would be

- hide the window
- set the map tool
- show instructions in the message bar
- when done re-raise the window
- when the map tool is deactivated re-raise the window

troopa81: Indeed, it seems the best approach, thanks for the hint, I made the modifications

src/app/qgsfeatureaction.cpp
```diff
@@ -53,22 +53,18 @@ QgsAttributeDialog *QgsFeatureAction::newDialog( bool cloneFeature )
...
 
   context.setDistanceArea( myDa );
-  context.setVectorLayerTools( QgisApp::instance()->vectorLayerTools() );
-  context.setMapCanvas( QgisApp::instance()->mapCanvas() );
-  context.setCadDockWidget( QgisApp::instance()->cadDockWidget() );
-  context.setMainMessageBar( QgisApp::instance()->messageBar() );
   context.setFormMode( QgsAttributeEditorContext::StandaloneDialog );
 
   QgsAttributeDialog *dialog = new QgsAttributeDialog( mLayer, f, cloneFeature, parentWidget(), true, context );
-  dialog->setWindowFlags( dialog->windowFlags() | Qt::Tool );
```
m-kuhn: Is this done somewhere else now?

src/gui/qgsmaptooladvanceddigitizing.cpp
```diff
@@ -25,7 +25,6 @@ QgsMapToolAdvancedDigitizing::QgsMapToolAdvancedDigitizing( QgsMapCanvas *canvas
   : QgsMapToolEdit( canvas )
   , mCadDockWidget( cadDockWidget )
 {
-  Q_ASSERT( cadDockWidget );
```
m-kuhn: What's the reason for removing this?

src/app/qgsfeatureaction.cpp
troopa81: No, it was from a previous modification, I forget to restore it. It's done now.

src/gui/qgsmaptooladvanceddigitizing.cpp
troopa81: I want to disable geometry creation from relation reference widget and relation editor based on the fact that _cadDockWidget_ variable from context is null because of QgsMapToolFillRing (it creates one feature and modify another, it's a mess if you start creating child at this very moment).

After a second thoughts, I change the way I did this to keep the assert. Do you think it's fine this way?

m-kuhn: Yes, that sounds good. I'd just add a docstring (to `setEditorContext`) to make this kind of behavior explicit for other places where we will want the same behavior.

troopa81: Done! I also add the same kind of behavior for the relation reference widget.

m-kuhn (approved):
Thank you for the patience @troopa81

troopa81:
@m-kuhn thanks for your review and precious help

mhugo:
Merged, since it has been approved. Thanks @troopa81 !

PR #33993, opened 2020-01-23
Improve layer properties info dialog layout

i-s-o:
Followup  #33955, backport to release-3_10 would be appreciated.

Smaller font, and no wrap for the first column. Also fixes WMS table layout:
image


elpaso:
Thank you, but the horizontal scrollbar is a no-go: is really bad UX.
Can you make sure it doesn't appear unless the window is really super-narrow?

i-s-o:
> ... the horizontal scrollbar is a no-go: is really bad UX.

Yes, I agree.

However, this is the current situation (in 3.10 nightly), which is not the best UX either:
image

i-s-o:
I will take another look at this later tonight.

DelazJ:
I don't know how they are related but just want to say that if you:
1. expand the browser panel --> xyz tiles
1. right-click the openstreetmap entry and select "Layer properties"
1. view "metadata tab

things are rendered differently (and better?) there than when you actually load the layer and view its properties.

PR #2265, opened 2015-08-21
fix crash when using opengl in plugin

vmora:
@wonder-sk 

X server is not thread safe and using opengl contexts in several threads\* makes qgis crash. It works with mesa though, but I wonder if offscreen rendering really accesses the X server, nvidia libs do.

*in a plugin: one for a PluginLayer, another for a widget, and yet another for 3D

PR #34002, opened 2020-01-23
Set up an OpenGL Context to be shared between threads beforehand

denchat:
for plugins that depend on Qt WebEngine module.

## Description

[QGIS 3.11.0-61 4a7ea580e9 ] Plugin go2streetview somehow triggers a critical message here

> `CRITICAL    Qt : Qt WebEngine seems to be initialized from a plugin. Please set Qt::AA_ShareOpenGLContexts using QCoreApplication::setAttribute before constructing QGuiApplication.`

It turns out the plugin depends on Qt WebEngine that suggests setting up a shared opengl context.

- https://doc.qt.io/qt-5/qtwebengine.html
- https://code.qt.io/cgit/qt/qtwebengine.git/plain/src/webenginewidgets

### DIFF
--- a/src/app/main.cpp
+++ b/src/app/main.cpp
@@ -884,6 +884,15 @@ int main( int argc, char *argv[] )
   QCoreApplication::setAttribute( Qt::AA_DisableWindowContextHelpButton, true );
 #endif
 
+  // Set up an OpenGL Context to be shared between threads beforehand
+  // for plugins that depend on Qt WebEngine module.
+  // As suggested by Qt documentation at
+  //   - https://doc.qt.io/qt-5/qtwebengine.html
+  //   - https://code.qt.io/cgit/qt/qtwebengine.git/plain/src/webenginewidgets/api/qtwebenginewidgetsglobal.cpp
+#if defined(QT_OS_WIN) && !defined(QT_NO_OPENGL) && (QT_VERSION >= QT_VERSION_CHECK(5, 4, 0) )
+  QCoreApplication::setAttribute( Qt::AA_ShareOpenGLContexts, true );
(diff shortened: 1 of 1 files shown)


nyalldawson:
@wonder-sk if I recall correctly this change was proposed some years ago, and you had reservations?

wonder-sk:
Hmm I am unable to find any previous PR enabling Qt::AA_ShareOpenGLContexts - maybe it was some other flag?

The documentation is a bit sparse on this flag, so I don't really know if there could be any implications. But if I understand correctly, for Qt WebEngine this may be necessary as its architecture spawns child processes for web pages.

I guess we can try to merge and even backport to 3.10 at some point later if this flag does not cause new issues especially with QGIS 3D...

nyalldawson:
Ah I looked back and it was https://github.com/qgis/QGIS/pull/2265 - so a different flag

NathanW2:
Is this still a good fix  Seems to be failing the builds but we can fix that if it's a valid fix for the issue.  Ran into this yesterday.  Seems to enable dev tools in the web page view you need the web settings which are part of web engine now and you hit this import issue 🙄

nyalldawson:
@nathanw2 it's just a indentation issue -- you could adopt this pr and fix the style?

denchat:
Unfortunately, as of 3.11-dev 51ba7e8a89 build, the issue is still persisting when loaded with go2streetview plugin.

> CRITICAL Qt : Qt WebEngine seems to be initialized from a plugin. Please set Qt::AA_ShareOpenGLContexts using QCoreApplication::setAttribute before constructing QGuiApplication.

PR #34006, opened 2020-01-23
Some tweaks to the add delimited text layer dialog

DelazJ:
- Avoid the crs widget display when there is no geometry expected in the layer
- Align the crs widget with other geometry related settings
- various cleanup of the ui file and tab order
Instead of 
image
this PR proposes
image
and 
image
I think a cleaner solution would be to display the geometric settings inside a frame as done for delimiters but i can't manage in Qt Designer to find a container that would cleanly embed all of these widgets. I think it could be done in cpp but i'll let more qualified people spend time on that


luipir (approved):
I totally agree

DelazJ:
Could this be backported, please? I think it's more like a bugfix (showing crs while selection is "no geometry"). Thanks.

PR #34009, opened 2020-01-24
[wfs] Fix background cache iterator when using with both subset string and filter on fid(s)

troopa81:
#Description

In QgsBackgroundCacheFeatureIterator we combine request and side filter expression request (when a subsetfilter is set on layer) but it removes the fid filter on input request if there is one. This PR fixes this. 

## Checklist

### DIFF
--- a/src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
+++ b/src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
@@ -257,25 +257,27 @@ void QgsThreadedFeatureDownloader::run()
 
 // -------------------------
 
-static QgsFeatureRequest addSubsetToFeatureRequest( const QgsFeatureRequest &requestIn,
-    const QgsBackgroundCachedSharedData *shared )
-{
-  if ( shared->clientSideFilterExpression().isEmpty() )
-  {
-    return requestIn;
-  }
-  QgsFeatureRequest requestOut( requestIn );
-  requestOut.combineFilterExpression( shared->clientSideFilterExpression() );
-  return requestOut;
-}
-
 QgsBackgroundCachedFeatureIterator::QgsBackgroundCachedFeatureIterator(
   QgsBackgroundCachedFeatureSource *source, bool ownSource,
   std::shared_ptr<QgsBackgroundCachedSharedData> shared,
   const QgsFeatureRequest &request )
-  : QgsAbstractFeatureIteratorFromSource<QgsBackgroundCachedFeatureSource>( source, ownSource, addSubsetToFeatureRequest( request, shared.get() ) )
+  : QgsAbstractFeatureIteratorFromSource<QgsBackgroundCachedFeatureSource>( source, ownSource, request )
   , mShared( shared )
 {
+  if ( !shared->clientSideFilterExpression().isEmpty() )
+  {
+    // backup current request because combine filter expression will remove the fid(s) filtering
+    if ( mRequest.filterType() == QgsFeatureRequest::FilterFid || mRequest.filterType() == QgsFeatureRequest::FilterFids )
+    {
+      mAdditionalRequest = mRequest;
+      mRequest = QgsFeatureRequest( shared->clientSideFilterExpression() );
+    }
+    else
+    {
+      mRequest.combineFilterExpression( shared->clientSideFilterExpression() );
+    }
+  }
+
   if ( mRequest.destinationCrs().isValid() && mRequest.destinationCrs() != mShared->sourceCrs() )
   {
     mTransform = QgsCoordinateTransform( mShared->sourceCrs(), mRequest.destinationCrs(), mRequest.transformContext() );
@@ -601,6 +603,11 @@ bool QgsBackgroundCachedFeatureIterator::fetchFeature( QgsFeature &f )
       continue;
     }
 
+    if ( !mAdditionalRequest.acceptFeature( cachedFeature ) )
+    {
+      continue;
+    }
+
     copyFeature( cachedFeature, f, true );
     geometryToDestinationCrs( f, mTransform );
 
@@ -688,6 +695,11 @@ bool QgsBackgroundCachedFeatureIterator::fetchFeature( QgsFeature &f )
         QgsGeometry constGeom = feat.geometry();
         if ( !mFilterRect.isNull() &&
              ( constGeom.isNull() || !constGeom.intersects( mFilterRect ) ) )
+        {
+          continue;
+        }
+
+        if ( !mAdditionalRequest.acceptFeature( feat ) )
         {
           continue;
         }
--- a/src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
+++ b/src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
@@ -269,8 +269,7 @@ QgsBackgroundCachedFeatureIterator::QgsBackgroundCachedFeatureIterator(
     // backup current request because combine filter expression will remove the fid(s) filtering
     if ( mRequest.filterType() == QgsFeatureRequest::FilterFid || mRequest.filterType() == QgsFeatureRequest::FilterFids )
     {
-      mAdditionalRequest = mRequest;
-      mRequest = QgsFeatureRequest( shared->clientSideFilterExpression() );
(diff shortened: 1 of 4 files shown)


roya0045:
I'm confused as to how the subset string is distinct from an expression. Though usually expressions and fids filter are exclusive (see #9915 )

src/providers/wfs/qgsbackgroundcachedfeatureiterator.h
```diff
@@ -321,6 +321,7 @@ class QgsBackgroundCachedFeatureIterator : public QObject,
 
     QgsCoordinateTransform mTransform;
     QgsRectangle mFilterRect;
+    QgsFeatureRequest mAdditionnalRequest;
```
m-kuhn:
```suggestion
    QgsFeatureRequest mAdditionalRequest;
```

m-kuhn (commented):
Looks good, thanks :+1:

Just one small thing, using variable names from inside a provider in a title for a pull request (like the "clientSideFilterExpression" mangled to sidefilterexpression) is normally not very helpful to get started. It's better to add some bigger picture topics to the title (subset filter, wfs, fid filter) to help reviewers filter what they want to look into.

nyalldawson:
@rouault how's this look to you?

src/providers/wfs/qgsbackgroundcachedfeatureiterator.h
rouault:
```suggestion
    QgsFeatureRequest mAdditionnalRequest;
```
```suggestion
    QgsFeatureRequest mAdditionalRequest; // typically to save a FilterFid/FilterFids request that will not be captured by mRequest 
```

rouault:
Pending my remark, and resolving issues raised in other providers by the added test, LGTM

src/providers/wfs/qgsbackgroundcachedfeatureiterator.h
troopa81: Done

troopa81: Done

troopa81:
> Just one small thing, using variable names from inside a provider in a title for a pull request (like the "clientSideFilterExpression" mangled to sidefilterexpression) is normally not very helpful to get started. It's better to add some bigger picture topics to the title (subset filter, wfs, fid filter) to help reviewers filter what they want to look into.

True. Renamed

troopa81:
> I'm confused as to how the subset string is distinct from an expression. Though usually expressions and fids filter are exclusive

Actually, this is not exactly the same thing than combining filter request (from what I understand), it's a first filtering on subset string, then you filter on fid based on the result of first filtering. For this provider, it could be managed by a filter AND combination (if such request combination was possible), for the others it's quite different (traduce with where clause for PG provider for instance).

troopa81:
> Pending my remark, and resolving issues raised in other providers by the added test, LGTM

I fixed it, waiting for travis now

src/providers/delimitedtext/qgsdelimitedtextfeatureiterator.cpp
```diff
@@ -104,6 +104,14 @@ QgsDelimitedTextFeatureIterator::QgsDelimitedTextFeatureIterator( QgsDelimitedTe
     mMode = FeatureIds;
     mTestSubset = false;
   }
+  else if ( request.filterType() == QgsFeatureRequest::FilterFids )
+  {
+    QgsDebugMsg( QStringLiteral( "Configuring for returning list of ids" ) );
```
m-kuhn: Is that something we want to keep for prod?

src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
```diff
@@ -255,25 +255,24 @@ void QgsThreadedFeatureDownloader::run()
...
 {
+  if ( !shared->clientSideFilterExpression().isEmpty() )
+  {
+    // backup current request because combine filter expression will remove the fid(s) filtering
+    if ( mRequest.filterType() == QgsFeatureRequest::FilterFid || mRequest.filterType() == QgsFeatureRequest::FilterFids )
+    {
+      mAdditionalRequest = mRequest;
+    }
+
+    mRequest.combineFilterExpression( shared->clientSideFilterExpression() );
```
m-kuhn: What happens if I do

```
request = QgsFeatureRequest();
request.setFilterExpression( 'area > 1000')
request.setFilterFid(1)
wfsLayer->getFeatures( request )
```

i.e. I think this needs to go into an else and the conditional part above modified accordingly

m-kuhn (commented):
Something more I wonder on the conceptual side.
Does WFS support an optimize get feature by fid(s) request?

Right now - if I understand the approach correctly - a get feature by fid (e.g. identify feature) will download the complete subset (potentially thousands of features) and filter locally for a single fid. It seems the inverse approach would be way more efficient.

rouault:
> Does WFS support an optimize get feature by fid(s) request?

It does. But... you need to provide a gml:id as input, not a QGIS feature id, and there's no universal way of creating a bijection between both. So, a QGIS get feature by fid request can indeed download the whole dataset from the server. But if you identify a feature already downloaded from the server, it will start by querying the temporary Spatialite database first, so things are not that bad in practice for interactive scenarios.

src/providers/delimitedtext/qgsdelimitedtextfeatureiterator.cpp
troopa81: Not sure to understand the question. We need to support the fids filter for tests and prod, no?

src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
troopa81: In this case, the request is no longer an expression request, its a filter fid request (even if `mFilterExpression` is not reset, `mFilter` which is the request type is equals to FilterFid)

From what I understand, This is not possible to combine a filter fid and a filter expression in the same QgsFeatureRequest.

So, I would say that provider deals only with QgsFeatureRequest type, fid in your example.

src/providers/delimitedtext/qgsdelimitedtextfeatureiterator.cpp
m-kuhn: It was only about the debug messge

src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
m-kuhn: I think mRequest.combineFilterExpression will combine to `area > 1000 and [CONDITION CLIENT SIDE]` which might unexpectedly also filter fid 1.

src/providers/delimitedtext/qgsdelimitedtextfeatureiterator.cpp
troopa81: I removed this part anyway

src/providers/wfs/qgsbackgroundcachedfeatureiterator.cpp
troopa81: You 're right. I made the modifications

troopa81:
I revert my modification on delimited provider, I had a bad understanding of the real issue (just a mix up between pk and id)

troopa81:
Don't stale

troopa81:
> Pending my remark, and resolving issues raised in other providers by the added test, LGTM

@rouault was ok for merging, is it ok for you @m-kuhn ?

m-kuhn:
Is there any drawback on inverting the approach (i.e. doing the fid request towards the backend and the subset filtering as post processing step)?

I think it would be much more future proof if the fid->pk lookup situation improves in the future.

troopa81:
> Is there any drawback on inverting the approach (i.e. doing the fid request towards the backend and the subset filtering as post processing step)?

No, I think its better indeed. I [changed it](3d3af4c)

troopa81:
@m-kuhn Does it look like ready for merging with the last modification?

m-kuhn:
Yes, sorry I must have gotten sidetracked last time I looked at it.

m-kuhn:
Thanks a lot

PR #34012, opened 2020-01-24
Dataprovider ogr spatialite respect provider defaults

### LINKED ISSUE #33383: Edit Form shows and saves raw default-values from geopackage, spatialite or sqlite
**Describe the bug**

QGIS shows default values from sqlite tables as string in the edit-form in case of creating a new row. After save, the raw default values are stored in the table.

QGIS-form with new feature and 'wrong' default-values. Feature from sql-insert with id 1 in background.
qgis-form with default values from database

QGIS-table after save_ shown with pointless default-values.
qgis-table view after editing

**How to Reproduce**

1. Create table in sqlite, spatialite or geopackage or use provided project [default_field_value.zip](https://github.com/qgis/QGIS/files/3961156/default_field_value.zip)

```sql
-- create the test-table
CREATE TABLE my_table (
    id integer primary key autoincrement,
    comment text,
    created_at_01 text DEFAULT (datetime('now','localtime')),
    created_at_02 text DEFAULT CURRENT_TIMESTAMP
);

-- Test the function of the default values
INSERT INTO my_table (comment) VALUES ('hello world');

SELECT * FROM my_table;

id          comment      created_at_01        created_at_02      
----------  -----------  -------------------  -------------------
1           hello world  2019-12-09 18:00:12  2019-12-09 17:00:12
```

2. Load this table into QGIS
3. Start editing
4. Add a Feature
5. -> Form is prefilled with default-value function

Post on [gis.stackexchange](https://gis.stackexchange.com/questions/344448/qgis-3-shows-and-saves-raw-default-values-from-spatialite-tables)

**QGIS and OS versions**
- QGIS 3.4 LTR (3.4.14 on Windows 10 and Linux / 3.4 on OSGeoLive 13)
- QGIS 3.10 (Windows 10)

**Tried Solutions**

 - with "sql-insert" everything works as exepected
 - same behaviour on different workstations
 - same behaviour for geopackage, sqlite- and spatialite-databases
 - same behaviour with or whitout geometry
 - same behaviour in QGIS 3.4.7 (LTR) and 3.10
 - same behaviour in Windows and Linux
 - in spatialite `InvalidateLayerStatistics();` and `UpdateLayerStatistics();` does not help.
 - Project Properties -> Data Source -> Check "Evaluate default values on provider side".

**Additional context**

Easiest workaround: Set the default value as well in the field widget. Layer Properties -> Attributes Forms -> field -> Default Value `now()`

maybe relevant commit https://github.com/qgis/QGIS/commit/0ae610c5e3dd364385c2fd91f901e706a04101f1

elpaso:
Fixes #33383

There is also an initial attempt to homogenize defaultValue and defaultValueClause between providers.

See also: https://lists.osgeo.org/pipermail/qgis-developer/2020-January/060007.html

### DIFF
--- a/src/core/providers/ogr/qgsogrprovider.cpp
+++ b/src/core/providers/ogr/qgsogrprovider.cpp
@@ -1385,7 +1385,7 @@ QVariant QgsOgrProvider::defaultValue( int fieldId ) const
 
   QString defaultVal = mDefaultValues.value( fieldId, QString() );
   if ( defaultVal.isEmpty() )
-    return defaultVal;
+    return QVariant();
 
   QVariant resultVar = defaultVal;
   if ( defaultVal == QStringLiteral( "CURRENT_TIMESTAMP" ) )
--- a/src/core/providers/ogr/qgsogrprovider.cpp
+++ b/src/core/providers/ogr/qgsogrprovider.cpp
@@ -1385,7 +1385,7 @@ QVariant QgsOgrProvider::defaultValue( int fieldId ) const
 
   QString defaultVal = mDefaultValues.value( fieldId, QString() );
   if ( defaultVal.isEmpty() )
-    return QVariant();
+    return defaultVal;
 
   QVariant resultVar = defaultVal;
   if ( defaultVal == QStringLiteral( "CURRENT_TIMESTAMP" ) )
--- a/src/core/providers/ogr/qgsogrprovider.cpp
+++ b/src/core/providers/ogr/qgsogrprovider.cpp
@@ -1584,7 +1584,12 @@ bool QgsOgrProvider::addFeaturePrivate( QgsFeature &f, Flags flags )
     OGRFieldType type = OGR_Fld_GetType( fldDef );
 
     QVariant attrVal = attrs.at( qgisAttId );
-    if ( attrVal.isNull() || ( type != OFTString && attrVal.toString().isEmpty() ) )
+    // The field value is equal to the default (that might be a provider-side expression)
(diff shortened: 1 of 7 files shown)


src/core/providers/ogr/qgsogrprovider.cpp
```diff
@@ -1380,7 +1385,7 @@ QVariant QgsOgrProvider::defaultValue( int fieldId ) const
 
   QString defaultVal = mDefaultValues.value( fieldId, QString() );
   if ( defaultVal.isEmpty() )
-    return QVariant();
+    return defaultVal;
```
nyalldawson: I don't think this change is correct -- there's lots of existing code which checks if the returned value isValid(), and now that will return true instead of false (yet the default value **doesn't** exist)

m-kuhn: And wouldn't the check for `mDefaultValues.hasKey` be correct here?
Not that an empty string as default value is common, but it's at least in SQL terms it's valid.

elpaso: That was an attempt to homogenize the behavior, I will revert it and give up with the attempt.

But if you have a look to the tests, you'll see that the actual behavior of the two methods it is totally different between providers (I tested PG ogr and spatialite, but I expect that the others will not behave better).

What they return depends on a permutation of:
- the fact that it is a literal default or not
- the fact that it is a particular function/magic-value or not (CURRENT_TIMESTAMP for instance)
- which provider (and maybe which driver for OGR)
- if EvaluateDefaultValues is honored and its value

The least we can do is to document it in the API and possibly try to fix it for QGIS 4.

PR #34017, opened 2020-01-24
PG: fix default value for IDENTITY PKs

### LINKED ISSUE #29560: PostgreSQL identity column not recognized properly
Author Name: **Michal Zimmermann** (Michal Zimmermann)
Original Redmine Issue: [21745](https://issues.qgis.org/issues/21745)
Affected QGIS version: 3.6.1
Redmine category:data_provider/postgis

---
Having the following table definition:
```
CREATE TABLE test (id integer generated always as identity primary key, geom geometry(point, 4326));

```
features cannot be added to QGIS, because the Feature Attributes dialogue requires id column to be not null. However, when any value is filled, the saving fails with the following error:

```
Could not commit changes to layer test

Errors: ERROR: 1 feature(s) not added.
  
  Provider errors:
      PostGIS error while adding features: ERROR:  cannot insert into column "id"
    DETAIL:  Column "id" is an identity column defined as GENERATED ALWAYS.
    HINT:  Use OVERRIDING SYSTEM VALUE to override.

```

According to https://blog.2ndquadrant.com/postgresql-10-identity-columns/, the syntax of identity columns in PostgreSQL conforms to the SQL standard and it's been around for a while.

elpaso:
Fixes #29560 - PostgreSQL identity column not recognized properly

### DIFF
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -1111,6 +1111,31 @@ bool QgsPostgresProvider::loadFields()
     }
 
     mAttrPalIndexName.insert( i, fieldName );
+    // If this is an identity field with constraints and there is no default, let's look for a sequence:
+    // we might have a default value created by a sequence named <table>_<field>_seq
+    if ( ! identityMap[tableoid ][ attnum ].isEmpty()
+         && notNullMap[tableoid][ attnum ]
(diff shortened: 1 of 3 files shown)


nyalldawson:
Looks like this change broke the Travis build -- can you investigate?

elpaso:
I see Travis green here, where should I look for the error?

elpaso:
@nyalldawson looking at it...

PR #34033, opened 2020-01-26
PG raster: fix WKB parser band number

elpaso:
Fix https://github.com/qgis/QGIS/pull/33685#issuecomment-578453699

### DIFF
--- a/src/providers/postgres/raster/qgspostgresrasterutils.cpp
+++ b/src/providers/postgres/raster/qgspostgresrasterutils.cpp
@@ -18,7 +18,6 @@
 #include "qgspostgresrasterutils.h"
 #include "qgsmessagelog.h"
 
-
 QVariantMap QgsPostgresRasterUtils::parseWkb( const QByteArray &wkb, int bandNo )
 {
   QVariantMap result;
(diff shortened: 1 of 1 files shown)


jgrocha:
Amazing! Thank you!

Screenshot from 2020-01-26 17-21-35

I'll try to debug/improve this new provider. I think we need a few tweaks in `QgsRasterDataProvider::block`. I'll start from there.

elpaso:
Thank you for finding the bug!

elpaso:
... and your help is more than welcome of course 😀

jgrocha:
I was getting a lot of errors like:
```
src/core/qgsmessagelog.cpp:29 : (logMessage) [4ms] [thread:0x5617bca5b020] 2020-01-26T22:25:10 PostGIS[2] Unable to write tile to temporary raster from "indb256"."carta_1870planta": Access window out of range in RasterIO().  Requested
(1536,3328) of size 256x206 on raster of 2636x3533.
```
Why? Because (in this case) 3328+206=3534, but the raster size is just 3533.

These errors with just one pixel more than the raster size prevents the rendering. And these error happens a lot.

I saw that raster size computed here (src/providers/postgres/raster/qgspostgresrasterprovider.cpp:1061)
```cpp
        // Compute raster size
        mHeight = static_cast<long>( mExtent.height() / std::abs( mScaleY ) );
        mWidth = static_cast<long>( mExtent.width() / std::abs( mScaleX ) );
```
is always one pixel less than the real size, but I don't think that's the problem.

Rather than fix the mathematical expressions, I've just band-aided the problem in `QgsPostgresRasterProvider::readBlock`, using the following approach:

```cpp
      long int safeWidth = tile.width;
      long int safeHeight = tile.height;

      if (xOff + tile.width > mWidth) {
          safeWidth = mWidth - xOff;
      }
      if (yOff + tile.height > mHeight) {
          safeHeight = mHeight - yOff;
      }

      CPLErr err =  GDALRasterIO( GDALGetRasterBand( tmpDS.get(), 1 ),
                                  GF_Write,
                                  xOff,
                                  yOff,
                                  static_cast<int>( safeWidth ),
                                  static_cast<int>( safeHeight ),
                                  ( void * )( tile.data.constData() ),  // old-style because of const
                                  static_cast<int>( safeWidth ),
                                  static_cast<int>( safeHeight ),
                                  gdalDataType,
                                  0,
                                  0 );
```
Now it works like a charm! Really cool!

I'll submit a PR just to have this provider working smoothly. If you are able to review the tile size calculation, fell free to close the PR. But with this little fix, I can already enjoy your work. A big thank you to you @elpaso. And it is so cool to add the rasters right from the browser :-)

jgrocha:
With PR #34041 I can now open all my test rasters, with different tile sizes, different SRID, in-db and out-of-db.

The provider still have some issues, but it starts to shine.

jgrocha:
I can see a render issue on the right side of the raster/canvas. The most right tiles have render problems. Screenshot is attached.

right most tiles issue

(image from Natural Earth data)

jgrocha:
> I can see a render issue on the right side of the raster/canvas. The most right tiles have render problems. Screenshot is attached.
> 
> right most tiles issue
> 
> (image from Natural Earth data)

This issue was fixed in #34053

PR #34040, opened 2020-01-26
Temporal API

Samweli:
This PR aims to provide an API implementation for temporal support in QGIS as mentioned in https://github.com/qgis/QGIS-Enhancement-Proposals/issues/161 and https://github.com/qgis/QGIS-Enhancement-Proposals/issues/128


src/core/temporal/qgsabstracttemporal.h
```diff
@@ -0,0 +1,125 @@
...
+  private:
+
+    //! Temporal state
+    bool mTemporal = false;
+
+    //! Represents datetime range member.
+    QgsDateTimeRange mDateTimeRange;
+
+    //! Datetime member, for storing instant of time
+    QDateTime *mDateTime = nullptr;
```
rouault: does this need to be a pointer ? The current implementation currently has a leak of it

tests/src/core/testqgsabstracttemporal.cpp
```diff
@@ -0,0 +1,133 @@
...
+#include <QApplication>
+#include <QFileInfo>
+#include <QDir>
+
+//qgis includes...
+#include <qgsabstracttemporal.h>
+#include <qgsapplication.h>
+#include <qgsproviderregistry.h>
+
+class TestSignalReceiver : public QObject
```
rouault: unused

src/core/temporal/qgsabstracttemporal.h
Samweli: Sure, it doesn't need to be a pointer, updating...

src/core/qgsmaplayer.h
```diff
@@ -76,7 +76,7 @@ enum class QgsMapLayerType SIP_MONKEYPATCH_SCOPEENUM_UNNEST( QgsMapLayer, LayerT
  * Base class for all map layer types.
  * This is the base class for all map layer types (vector, raster).
  */
-class CORE_EXPORT QgsMapLayer : public QObject
+class CORE_EXPORT QgsMapLayer : public QObject, public QgsAbstractTemporal
```
nyalldawson: I don't think map layers should inherit QgsAbstractTemporal here. Each map layer subclass will have its own way of dealing with time, and a fixed time range may not be applicable for a certain subclass. Even for the layer types which could have a fixed time range, we'll need to expose much more control than is available through QgsAbstractTemporal, so I think for map layers we should use a different temporal class.

src/core/qgsrendercontext.cpp
```diff
@@ -38,7 +38,8 @@ QgsRenderContext::QgsRenderContext()
 }
 
 QgsRenderContext::QgsRenderContext( const QgsRenderContext &rh )
-  : mFlags( rh.mFlags )
+  : QgsAbstractTemporal()
```
nyalldawson:
```suggestion
  : QgsAbstractTemporal( rh )
```

nyalldawson: The operator= method will also need to copy the temporal attributes

src/core/qgsrendercontext.cpp
```diff
@@ -38,7 +38,8 @@ QgsRenderContext::QgsRenderContext()
...
+  , mFlags( rh.mFlags )
   , mPainter( rh.mPainter )
   , mMaskPainter( rh.mMaskPainter )
   , mCoordTransform( rh.mCoordTransform )
```
nyalldawson: QgsRenderContext::fromMapSettings will need to copy the temporal attributes from the QgsMapSettings object (including a unit test)

src/core/qgsrendercontext.cpp
nyalldawson: Can you also expand on the QgsRenderContext copy/assignment unit tests to include checks of the temporal attributes?

src/core/temporal/qgsabstracttemporal.cpp
```diff
@@ -0,0 +1,73 @@
...
+}
+
+bool QgsAbstractTemporal::isTemporal() const
+{
+  return mTemporal;
+}
+
+void QgsAbstractTemporal::setTemporalRange( const QgsDateTimeRange &dateTimeRange )
+{
+  if ( !isTemporal() )
```
nyalldawson: I don't think we should have this precondition -- we should still allow the temporal range to be set/retrieved even if the temporal attributes are disabled.

src/core/temporal/qgsabstracttemporal.h
```diff
@@ -0,0 +1,125 @@
...
+
+
+#ifndef QGSABSTRACTTEMPORAL_H
+#define QGSABSTRACTTEMPORAL_H
+
+#include "qgis_core.h"
+#include "qgis_sip.h"
+#include "qgslogger.h"
+#include "qgsmessagelog.h"
+#include "qgsapplication.h"
```
nyalldawson: Remove

src/core/temporal/qgsabstracttemporal.h
```diff
@@ -0,0 +1,125 @@
...
+#include "qgsrange.h"
+
+#include <QObject>
```
nyalldawson: Remove

src/core/temporal/qgsabstracttemporal.h
```diff
@@ -0,0 +1,125 @@
...
+ * Subclasses may wish to update the abstract temporal functions.
+ *
+ * \since QGIS 3.14
+ */
+class CORE_EXPORT QgsAbstractTemporal
+{
+  public:
+
+    /**
+     * Constructors for QgsAbstractTemporal.
```
nyalldawson:
```suggestion
     * Constructor for QgsAbstractTemporal.
```

(doxygen comments are per-method, so they are all singular and apply to only one method)

src/core/temporal/qgsabstracttemporal.h
```diff
@@ -0,0 +1,125 @@
...
+     */
+    QgsAbstractTemporal();
+
+    QgsAbstractTemporal( const bool enabled );
```
nyalldawson: needs docs

nyalldawson: ### ### 
```suggestion
    QgsAbstractTemporal( bool enabled );
```

src/core/temporal/qgsabstracttemporal.h
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * \class QgsAbstractTemporal
+ * \ingroup core
+ * Base class for temporal based classes.
+ *
+ * Subclasses may wish to update the abstract temporal functions.
+ *
+ * \since QGIS 3.14
+ */
+class CORE_EXPORT QgsAbstractTemporal
```
nyalldawson:
```suggestion
class CORE_EXPORT QgsAbstractTemporal
```

QGIS convention is to name classes "abstract" only when they have at least one pure virtual method. `QgsTemporalRangeObject` might be a better name.

src/core/temporal/qgsabstracttemporal.h
```diff
@@ -0,0 +1,125 @@
+/***************************************************************************
```
nyalldawson: I'd suggest moving this class out of the /temporal directory, unless you have plans for adding a few more files into that directory.

Otherwise, there's a few missing pieces needed when adding a new subdirectory:

- Update the doc\CmakeLists.txt file to include the new subdirectory in DOXYGEN_INPUT
- Update python\CMakeLists.txt to include the new subdirectory
- Update tests\src\core\CMakeLists.txt to include it
- Don't use `#include "temporal/qgsabstracttemporal.h", and instead ensure that "#include "qgsabstracttemporal.h"` works instead

src/core/temporal/qgsabstracttemporal.h
```diff
@@ -0,0 +1,125 @@
...
+
+    QgsAbstractTemporal( const bool enabled );
+
+    virtual ~QgsAbstractTemporal() = default;
+
+    /**
+     * Sets object as a temporal based one, which will be considered when rendering maps with a specific time range set.
+     *
+     * \see isTemporal()
+     * \since QGIS 3.14
```
nyalldawson: QGIS documentation convention is that we don't include \since directives for members if which duplicate the class level \since directive. (i.e. you'd only use \since for new members for this class if its "\since QGIS 3.16" or later)

src/core/temporal/qgsabstracttemporal.h
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Sets current datetime object.
+     *
+     * Can be used in map canvas, setting the map temporal instant
+     *
+     * \see currentDateTime()
+     * \since QGIS 3.14
+    */
+
+    void setCurrentDateTime( QDateTime *dateTime );
```
nyalldawson: We don't want this here -- as per the QEP and discussion, everything will be done using time ranges. A fixed zero-length time instant would be represented by setting the range to a time range with equal start and end times.

src/gui/qgsmapcanvas.h
```diff
@@ -72,7 +73,7 @@ class QgsReferencedRectangle;
  * Map canvas is a class for displaying all GIS data types on a canvas.
  */
 
-class GUI_EXPORT QgsMapCanvas : public QGraphicsView
+class GUI_EXPORT QgsMapCanvas : public QGraphicsView, public QgsAbstractTemporal
```
nyalldawson: I suspect this move will make things difficult later -- if you use inheritance here, it will be clumsy to respond to changes in the time range. (you'd need to make the setters virtual and override them here). Composition may be a better move for the map canvas class.

m-kuhn: ... and if there is a place where it's being considered to make a setter virtual it's almost always better to not make it virtual and call an internal virtual (private) method instead. Just sayin'.

src/core/qgsrendercontext.cpp
Samweli: Right, working on it

src/core/temporal/qgsabstracttemporal.cpp
Samweli: Ok

Samweli: By the way @nyalldawson, removing the precondition will allow setting render context temporal range even if it not a temporal one!!

src/core/qgsrendercontext.cpp
Samweli: @nyalldawson why should we pass rendercontext to QgsTemporalRangeObject as in your suggestion?, `QgsTemporalRangeObject( rh )`

Samweli: > @nyalldawson why should we pass rendercontext to QgsTemporalRangeObject as in your suggestion? `QgsAbstractTemporal( rh )` which will now be`QgsTemporalRangeObject( rh )`

Ooh I see @nyalldawson, so we can copy over the temporal attributes...

nyalldawson (approved):

src/core/qgstemporalrangeobject.cpp
```diff
@@ -0,0 +1,63 @@
...
+}
+
+bool QgsTemporalRangeObject::isTemporal() const
+{
+  return mTemporal;
+}
+
+void QgsTemporalRangeObject::setTemporalRange( const QgsDateTimeRange &dateTimeRange )
+{
+  if ( !isTemporal() )
```
nyalldawson: Sorry, forgot about this change when I approved -- as we discussed, can you remove this condition and replace with the automatic mTemporal=true approach (with docs)?

nyalldawson (changes requested):

src/core/qgstemporalrangeobject.cpp
Samweli: Ok, I have replaced it.

src/core/qgsmaplayertemporalproperties.h
```diff
@@ -0,0 +1,66 @@
...
+     * \see readXml()
+     */
+    virtual QDomElement writeXml( ... ) = 0;
+
+    /**
+     * Reads temporal properties from a DOM element previously written by writeXml()
+     *
+     * \see writeXml()
+     */
+    virtual bool readXml( QDomElement ... ) = 0;
```
nyalldawson:
```suggestion
    virtual bool readXml( const QDomElement &element, const QgsReadWriteContext &context ) = 0;
```

src/core/qgsmaplayertemporalproperties.h
```diff
@@ -0,0 +1,66 @@
...
+    QgsMapLayerTemporalProperties( bool enabled = false );
+
+    virtual ~QgsMapLayerTemporalProperties() = default;
+
+    /**
+     * Writes map layer temporal properties to a DOM element, to be used later with readXml()
+     *
+     * \see readXml()
+     */
+    virtual QDomElement writeXml( ... ) = 0;
```
nyalldawson:
```suggestion
    virtual QDomElement writeXml( QDomElement &element, QDomDocument &doc, const QgsReadWriteContext &context ) = 0;
```

src/core/qgstemporalproperty.h
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Constructor for QgsTemporalProperty.
+     */
+    QgsTemporalProperty( bool enabled = false );
+
+    /**
+     * Copy constructor
+     */
+    QgsTemporalProperty( const QgsTemporalProperty &temporalProperty );
```
nyalldawson: This shouldn't be needed

src/core/qgstemporalrangeobject.cpp
```diff
@@ -0,0 +1,63 @@
...
+}
+
+bool QgsTemporalRangeObject::isTemporal() const
+{
+  return mTemporal;
+}
+
+void QgsTemporalRangeObject::setTemporalRange( const QgsDateTimeRange &dateTimeRange )
+{
+  if ( dateTimeRange == mDateTimeRange )
```
nyalldawson: I'd drop this condition -- it could lead to strange behavior, because the `setIsTemporal( true );` below won't be enforced.

src/core/qgstemporalrangeobject.h
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Creates temporal range object with the specified temporal state.
+     */
+    QgsTemporalRangeObject( bool enabled );
+
+    /**
+     * Copies render context temporal members
+     */
+    QgsTemporalRangeObject( const QgsRenderContext &rh );
```
nyalldawson: Not required

src/core/qgsvectorlayertemporalproperties.h
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * \class QgsVectorLayerTemporalProperties
+ * \ingroup core
+ * Implementation of map layer temporal properties for vector layers.
+ *
+ * \since QGIS 3.14
+ */
+
+class CORE_EXPORT QgsVectorLayerTemporalProperties : public QgsMapLayerTemporalProperties
```
nyalldawson: Remove this class please -- it's out of scope

src/core/raster/qgsrasterlayer.h
```diff
@@ -549,6 +555,9 @@ class CORE_EXPORT QgsRasterLayer : public QgsMapLayer
     //! Pointer to data provider
     QgsRasterDataProvider *mDataProvider = nullptr;
 
+    //! Pointer to temporal properties
+    std::unique_ptr< QgsRasterLayerTemporalProperties > mTemporalProperties;
```
nyalldawson: Don't forget to create this in the QgsRasterLayer constructor.

src/core/raster/qgsrasterlayertemporalproperties.h
```diff
@@ -0,0 +1,122 @@
...
+  private:
+
+    //! Temporal layer mode.
+    TemporalMode mMode;
+
+    //! Represents datetime range member.
+    QgsDateTimeRange mRange;
+
+    //! Instance of the temporal properties
+    QgsRasterLayerTemporalProperties *mInstance = nullptr;
```
nyalldawson: Remove

src/core/raster/qgsrasterlayertemporalproperties.h
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Constructor for QgsRasterLayerTemporalProperties.
+     */
+    QgsRasterLayerTemporalProperties( bool enabled = false );
+
+    ~QgsRasterLayerTemporalProperties();
+
+    enum TemporalMode
+    {
+      ModeFixedTemporalRange,
```
nyalldawson: You'll need to add docstrings for each of these too:

```suggestion
      ModeFixedTemporalRange, //!< Explanation of the value
```

src/core/raster/qgsrasterlayertemporalproperties.h
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * Set the temporal properties mode
+     *
+     *\see mode()
+    **/
+    void setMode( TemporalMode mode );
+
+    /**
+     * Sets the temporal \a range to apply to the whole layer. Allfeatures from
```
nyalldawson: Can you update this description to reflect that it's refers to raster layers?

src/core/raster/qgsrasterlayertemporalproperties.h
```diff
@@ -0,0 +1,122 @@
...
+
+    QDomElement writeXml( ... ) override;
+
+    bool readXml( QDomElement ... ) override;
+
+    /**
+     * Return the temporal properties instance
+     *
+    **/
+    QgsRasterLayerTemporalProperties *get() const;
```
nyalldawson: Remove

src/core/raster/qgsrasterlayertemporalproperties.h
```diff
@@ -0,0 +1,122 @@
...
+
+  private:
+
+    //! Temporal layer mode.
+    TemporalMode mMode;
```
nyalldawson: Initialise to a default value

tests/src/core/testqgsrasterlayer.cpp
```diff
@@ -112,6 +114,8 @@ class TestQgsRasterLayer : public QObject
     QgsRasterLayer *mPngRasterLayer = nullptr;
     QgsRasterLayer *mGeoJp2RasterLayer = nullptr;
 
+    QgsRasterLayerTemporalProperties *temporalProperties = nullptr;
```
nyalldawson: Try to avoid members like this in tests, and instead try to keep each individual test self-contained. It avoids a massive overflow of failing tests when one test which modifies the member variable fails.

tests/src/core/testqgsrasterlayer.cpp
```diff
@@ -997,5 +1004,19 @@ void TestQgsRasterLayer::sample()
...
+{
+  QCOMPARE( mpRasterLayer->temporalProperties(), temporalProperties );
+  QVERIFY( !mpRasterLayer->temporalProperties()->isActive() );
+
+  QgsDateTimeRange dateTimeRange = QgsDateTimeRange( QDateTime( QDate( 2020, 1, 1 ) ),
+                                   QDateTime( QDate( 2020, 12, 31 ) ) );
+
+  temporalProperties->setFixedTemporalRange( dateTimeRange );
+  QCOMPARE( mpRasterLayer->temporalProperties()->fixedTemporalRange(), dateTimeRange );
+
```
nyalldawson: Also add a test of the raster temporal properties writeXml and readXml methods (when you've implemented these)

src/core/qgstemporalrangeobject.h
Samweli: If we remove this we wont be able to copy over the render context temporal attributes here
https://github.com/qgis/QGIS/blob/96c8c2f7d2046c6db5b9204b97ea37ab7d012ba1/src/core/qgsrendercontext.cpp#L42

src/core/qgsvectorlayertemporalproperties.h
Samweli: Yap, I thought I deleted these!!

src/core/qgstemporalrangeobject.h
nyalldawson: Can you backport this change from your other PR, and then we'll merge this one? It'll make #34590 smaller and easier to review as a result...

Samweli: Alright

nyalldawson (approved):

PR #34046, opened 2020-01-27
Fix overflow by assigning double max to float, invalid min

nyalldawson:
Raising a valid warning on msvc builds

### DIFF
--- a/src/3d/qgsfeature3dhandler_p.h
+++ b/src/3d/qgsfeature3dhandler_p.h
@@ -122,8 +122,8 @@ class QgsFeature3DHandler
     void updateZRangeFromPositions( const QVector<QVector3D> &positions );
 
   protected:
-    float mZMin = std::numeric_limits<double>::max();
-    float mZMax = std::numeric_limits<double>::min();
+    float mZMin = std::numeric_limits<float>::max();
+    float mZMax = std::numeric_limits<float>::lowest();
 };
 
 


src/3d/qgsfeature3dhandler_p.h
```diff
@@ -122,8 +122,8 @@ class QgsFeature3DHandler
     void updateZRangeFromPositions( const QVector<QVector3D> &positions );
 
   protected:
-    float mZMin = std::numeric_limits<double>::max();
-    float mZMax = std::numeric_limits<double>::min();
+    float mZMin = std::numeric_limits<float>::max();
+    float mZMax = std::numeric_limits<float>::lowest();
```
wonder-sk: Is there are reason to switch from min() to lowest() ? As far as I understand lowest() returns some very small positive number close to zero (1.17549e-38) while we ideally want very large negative number...

elpaso: @wonder-sk it's the opposite: min returns the positive closest to 0, lowest returns the smallest negative

nirvn: TIL :)

wonder-sk: Oh :man_facepalming: I must have used it wrong so many times! (very confusing IMHO)

wonder-sk (approved):

PR #34048, opened 2020-01-27
Enable another test on azure

nyalldawson:
Fix source encoding issue in a test file, enable for azure builds


m-kuhn:
Very much +1

Issue #32080, opened 2019-10-01
Extend line does not work in compoundCurvedZ layer


SrNetoChan:
**Describe the bug**

It seems that the extend line functionality (node tool) only works editing the line starting point, but doesn't when you try to extend the final node. furthermore, this only happens when the feature is saved into the gpkg file. While only on memory it works fine.

I have tried a multilineZ layer and the problem did not show up. I was only able to reproduce using a CompoundCurveZ geopackage layer.

Here's a dataset sample

https://mega.nz/#!cYlmDaTb!0OTijkGrPWYCnuYa3zUHkrEdaCk4f5wBHNgEnnB5tng

**How to Reproduce**

I hope the screenshot is self-explanatory.

![Peek 2019-10-01 14-05](https://user-images.githubusercontent.com/3607161/65964627-44d9fa80-e455-11e9-924f-648cc30c367b.gif)

**QGIS and OS versions**

QGIS version | 3.9.0-Master | QGIS code revision | QGIS code revision
-- | -- | -- | --
Compiled against Qt | 5.9.5 | Running against Qt | 5.9.5
Compiled against GDAL/OGR | 2.2.3 | Running against GDAL/OGR | 2.2.3
Compiled against GEOS | 3.6.2-CAPI-1.10.2 | Running against GEOS | 3.6.2-CAPI-1.10.2 4d2925d6
Compiled against SQLite | 3.22.0 | Running against SQLite | 3.22.0
PostgreSQL Client Version | 11.5 (Ubuntu 11.5-1.pgdg18.04+1) | SpatiaLite Version | 4.3.0a
QWT Version | 6.1.3 | QScintilla2 Version | 2.10.2
PROJ.4 Version | 493
OS Version | Ubuntu 18.04.3 LTS
Active python plugins | Qgis2threejs; valuetool; DigitizingTools; mgrstools; qgistester; trackable_project_files; changeDataSource; quick_map_services; MemoryLayerSaver; firstaid; workbench; vector_tiles_reader; QuickOSM; processing; db_manager

**Additional context**

PR #34055, opened 2020-01-27
BUG: fix add vertex from end point for CompoundCurveZ in gpkg, fix #3…

SebastienPeillet:
## Description
#32080

As described in the issue, using a compound curve layer (in gpkg) it's not possible to extend existing entity from the end point, while it's possible from start point.

This was not possible because compound curve is composed of several part and user shouldn't be able to add an extra vertex in a part that is followed by another part. Extent from start point was available because the vertex index is 0, so we are sure there is no part before.

I fixed it by allowing vertex insert if it is for the last past (``i == mCurves.size() - 1``) of the curve and the vertex index is equal to the existing vertex count + 1

## Checklist

### DIFF
--- a/src/core/geometry/qgscompoundcurve.cpp
+++ b/src/core/geometry/qgscompoundcurve.cpp
@@ -730,6 +730,16 @@ QVector< QPair<int, QgsVertexId> > QgsCompoundCurve::curveVertexId( QgsVertexId 
       }
       break;
     }
+    else if ( id.vertex >= currentVertexIndex && id.vertex == currentVertexIndex + increment + 1 && i == ( mCurves.size() - 1 ) )
+    {
+      int curveVertexId = id.vertex - currentVertexIndex;
+      QgsVertexId vid;
+      vid.part = 0;
+      vid.ring = 0;
+      vid.vertex = curveVertexId;
+      curveIds.append( qMakePair( i, vid ) );
+      break;
+    }
     currentVertexIndex += increment;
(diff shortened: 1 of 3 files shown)


nyalldawson:
Can you please add a unit test for this?

SebastienPeillet:
Sorry for the delay, I added the tests for compound curve in the vertex tools, especially the ones about adding a extra vertex to a linestring in a compound curve layer and verifying that this is not available for circular string !

nyalldawson (commented):
The new test is much appreciated, but can we get a specific test in testqgsgeometry.cpp testing the compound curved insertVertex method in this circumstance?

SebastienPeillet:
Ok, I added a new test in compoundCurve function of testqgsgeometry.cpp. It verify that a vertex can be added to a Linestring in a compoundCurve geometry. 

It's already tested that no extra vertex can be added to a CircularString ([line 10595](https://github.com/qgis/QGIS/blob/f9f16f0e8230c995e788bcbc3be9c6af66c5e394/tests/src/core/testqgsgeometry.cpp#L10595)):

```cpp
  //insert vertex at end
  QVERIFY( !c24.insertVertex( QgsVertexId( 0, 0, 9 ), QgsPoint( 31.0, 32.0 ) ) );
```

nyalldawson (approved):

nyalldawson:
Thanks @SebastienPeillet !

PR #34061, opened 2020-01-27
Add algorithm "Fill NoData cells"

root676:
## Description
_although it is feature freeze..._

This PR introduces a new processing algorithm for filling NoData cells in a raster dataset.
This algorithm scans an input raster dataset for NoData cells and fills them with the user defined fill value. The algorithm respects the data type of the input raster so that eg. floating point numbers will be dealt appropriately with an integer raster. The resulting dataset will not contain any NoData cells.

The PR also contains a unit test for the algorithm.

Please feel free to review/comment/suggest improvements!

## Checklist


src/analysis/processing/qgsalgorithmfillnodata.cpp
```diff
@@ -0,0 +1,158 @@
...
+}
+
+QString QgsFillNoDataAlgorithm::displayName() const
+{
+  return QObject::tr( "Fill NoData cells" );
+}
+
+QStringList QgsFillNoDataAlgorithm::tags() const
+{
+  return QObject::tr( "noData,NoData,data,cells,fill,set" ).split( ',' );
```
nyalldawson:
```suggestion
  return QObject::tr( "noData,NoData,data,cells,fill,set" ).split( ',' );
```
```suggestion
  return QObject::tr( "data,cells,fill,set" ).split( ',' );
```

The tag searching isn't case sensitive, and there's no need to include terms which are already part of the display name.

src/analysis/processing/qgsalgorithmfillnodata.cpp
```diff
@@ -0,0 +1,158 @@
...
+QString QgsFillNoDataAlgorithm::groupId() const
+{
+  return QStringLiteral( "rastertools" );
+}
+
+void QgsFillNoDataAlgorithm::initAlgorithm( const QVariantMap & )
+{
+  addParameter( new QgsProcessingParameterRasterLayer( QStringLiteral( "INPUT" ), QStringLiteral( "Raster input" ) ) );
+  addParameter( new QgsProcessingParameterBand( QStringLiteral( "BAND" ), QObject::tr( "Band Number" ), 1, QStringLiteral( "INPUT" ) ) );
+  addParameter( new QgsProcessingParameterNumber( QStringLiteral( "FILL_VALUE" ), QObject::tr( "Fill value" ), QgsProcessingParameterNumber::Double, 1, false, 1, 10000000 ) );
```
nyalldawson:
```suggestion
  addParameter( new QgsProcessingParameterNumber( QStringLiteral( "FILL_VALUE" ), QObject::tr( "Fill value" ), QgsProcessingParameterNumber::Double, 1, false ) );
```

I don't think we should have a fixed min/max here -- we need to allow negative numbers and potentially there's use cases where 0 or > 10000000 would be required.

src/analysis/processing/qgsalgorithmfillnodata.cpp
```diff
@@ -0,0 +1,158 @@
...
+  addParameter( new QgsProcessingParameterRasterDestination( QStringLiteral( "OUTPUT" ), QObject::tr( "Output raster" ) ) );
+}
+
+QString QgsFillNoDataAlgorithm::shortHelpString() const
+{
+  return QObject::tr( "This algorithm resets the NoData values in the input raster "
+                      "to a chosen value. This value can be set by the user using the Fill value parameter. "
```
nyalldawson:
```suggestion
                      "to a chosen value, resulting in a raster dataset with no NoData pixels. This value can be set by the user using the Fill value parameter. "
```

src/analysis/processing/qgsalgorithmfillnodata.cpp
```diff
@@ -0,0 +1,158 @@
...
+                      "The algorithm respects the input raster data type (eg. a fill value with commas gets truncated "
```
nyalldawson:
```suggestion
                      "The algorithm respects the input raster data type (eg. a floating point fill value will be truncated "
```

src/analysis/processing/qgsalgorithmfillnodata.cpp
```diff
@@ -0,0 +1,158 @@
...
+  mNbCellsXProvider = mInterface->xSize();
+  mNbCellsYProvider = mInterface->ySize();
+  return true;
+}
+
+QVariantMap QgsFillNoDataAlgorithm::processAlgorithm( const QVariantMap &parameters, QgsProcessingContext &context, QgsProcessingFeedback *feedback )
+{
+  //test if input dataset has NoData
+  if ( !mInputRaster->dataProvider()->sourceHasNoDataValue( mBand ) )
+    throw QgsProcessingException( QObject::tr( "Input raster has no NoData values. There exist no NoData cells to fill." ) );
```
nyalldawson: I'm not sure we want this to be an exception here, and instead think we should just raise a warning (`feedback->reportError( ... )`). If we raise an exception it will terminate a model when encountered, and potentially users may need a model which can gracefully handle inputs either with or without nodata values present.

src/analysis/processing/qgsalgorithmfillnodata.cpp
```diff
@@ -0,0 +1,158 @@
...
+  writer->setOutputFormat( outputFormat );
+  std::unique_ptr<QgsRasterDataProvider > provider( writer->createOneBandRaster( mInterface->dataType( mBand ), mNbCellsXProvider, mNbCellsYProvider, mExtent, mCrs ) );
+  if ( !provider )
+    throw QgsProcessingException( QObject::tr( "Could not create raster output: %1" ).arg( outputFile ) );
+  if ( !provider->isValid() )
+    throw QgsProcessingException( QObject::tr( "Could not create raster output %1: %2" ).arg( outputFile, provider->error().message( QgsErrorMessage::Text ) ) );
+
+  //prepare output provider
+  mDestinationRasterProvider = provider.get();
+  mDestinationRasterProvider->setNoDataValue( 1, -9999 );
```
nyalldawson: I don't think we want to set this -- the output is supposed to have NO nodata value, and potentially -9999 is an existing, valid value used within the input raster layer.

src/analysis/processing/qgsalgorithmfillnodata.cpp
```diff
@@ -0,0 +1,158 @@
...
+  int iterTop = 0;
+  int iterCols = 0;
+  int iterRows = 0;
+  std::unique_ptr< QgsRasterBlock > filledRasterBlock;
+  while ( iter.readNextRasterPart( mBand, iterCols, iterRows, filledRasterBlock, iterLeft, iterTop ) )
+  {
+    if ( feedback )
+      feedback->setProgress( 100 * ( ( iterTop / maxHeight * nbBlocksWidth ) + iterLeft / maxWidth ) / nbBlocks );
+    if ( !filledRasterBlock->hasNoDataValue() )
+      continue;
```
nyalldawson: Still need to write the block to the destination provider (untouched!)

src/analysis/processing/qgsalgorithmfillnodata.cpp
```diff
@@ -0,0 +1,158 @@
...
+    for ( int row = 0; row < iterRows; row++ )
+    {
+      if ( feedback && feedback->isCanceled() )
+        break;
+      for ( int column = 0; column < iterCols; column++ )
+      {
+        if ( feedback && feedback->isCanceled() )
```
nyalldawson: Could skip this -- I think in other raster algorithms we only check for cancellation per row, and not on every pixel

src/analysis/processing/qgsalgorithmfillnodata.h
```diff
@@ -0,0 +1,78 @@
...
+    QgsRectangle mExtent;
+    QgsCoordinateReferenceSystem mCrs;
+    int mLayerWidth;
+    int mLayerHeight;
+    int mNbCellsXProvider = 0;
+    int mNbCellsYProvider = 0;
+    double mInputNoDataValue;
+    QgsRasterDataProvider *mDestinationRasterProvider;
+
+    bool mDynamicFillValue = false;
```
nyalldawson: Not needed

src/analysis/processing/qgsalgorithmfillnodata.h
```diff
@@ -0,0 +1,78 @@
...
+    QgsProperty mDynamicFillValueProperty;
```
nyalldawson: not needed

src/analysis/processing/qgsalgorithmfillnodata.h
```diff
@@ -0,0 +1,78 @@
...
+    int mBand;
+    std::unique_ptr< QgsRasterInterface > mInterface;
+    QgsRectangle mExtent;
+    QgsCoordinateReferenceSystem mCrs;
+    int mLayerWidth;
+    int mLayerHeight;
+    int mNbCellsXProvider = 0;
+    int mNbCellsYProvider = 0;
+    double mInputNoDataValue;
+    QgsRasterDataProvider *mDestinationRasterProvider;
```
nyalldawson: This doesn't need to be a member variable -- it's only used locally in processAlgorithm

nyalldawson (commented):
Nice feature!

src/analysis/processing/qgsalgorithmfillnodata.h
root676: First I thought about implementing a dynamic parameter. But I saw too little use cases for it - although it would be interesting to have something like `$cellmidpoint` available in the expression variables for some dynamic geometry calculations per cell.

nyalldawson: True -- but I think it would be of limited use in this particular algorithm. A "create/update raster using expression" algorithm sounds very interesting though!

nyalldawson (approved):

PR #34064, opened 2020-01-28
allow replacement of invalid variants in TextEditWrapper

stev-0:
## Description
Post merge of #30758 I noticed some odd behaviour with editing JSON fields - upon entering an invaliid JSON field, the invalid text persisted in the Text Edit when changing records. This was caused by #33660 which overlapped the original PR. This PR attempts to return to the old behaviour without affecting the fix in #33660 

@troopa81 would be good to check this will not affect what you were doing

### DIFF
--- a/src/gui/editorwidgets/qgstexteditwrapper.cpp
+++ b/src/gui/editorwidgets/qgstexteditwrapper.cpp
@@ -341,13 +341,13 @@ void QgsTextEditWrapper::setWidgetValue( const QVariant &val )
     }
   }
 
-  if ( mPlainTextEdit )
-  {
-    if ( val != value() )
-      mPlainTextEdit->setPlainText( v );
-  }
+// allow replacement of invalid variants with
+// what was there before - e.g for JSON
+  if ( mPlainTextEdit && ( val != value() || !val.isValid() ) )
+    mPlainTextEdit->setPlainText( v );
 
-  if ( mLineEdit && val != value() )
+
+  if ( mLineEdit && ( val != value() || !val.isValid() ) )
     mLineEdit->setText( v );
 }
 


src/gui/editorwidgets/qgstexteditwrapper.cpp
```diff
@@ -347,7 +347,9 @@ void QgsTextEditWrapper::setWidgetValue( const QVariant &val )
       mPlainTextEdit->setPlainText( v );
   }
 
-  if ( mLineEdit && val != value() )
+// allow replacement of invalid variants with
+// what was there before - e.g for JSON
+  if ( mLineEdit && ( val != value() || val.typeName() == 0 ) )
```
nyalldawson:
```suggestion
  if ( mLineEdit && ( val != value() || !val.isValid() ) )
```

src/gui/editorwidgets/qgstexteditwrapper.cpp
```diff
@@ -347,7 +347,9 @@ void QgsTextEditWrapper::setWidgetValue( const QVariant &val )
       mPlainTextEdit->setPlainText( v );
   }
 
-  if ( mLineEdit && val != value() )
+// allow replacement of invalid variants with
+// what was there before - e.g for JSON
+  if ( mLineEdit && ( val != value() || !val.isValid() ) )
```
nyalldawson: Will this same fix apply to the `mPlainTextEdit` logic above?

stev-0: Yes, well caught, thanks

troopa81:
> @troopa81 would be good to check this will not affect what you were doing

Sorry to answer so lately. the idea of my fix IIRC was to prevent the modification of the line edit if the value hasn't change so I think your modification should be fine,

stev-0:
This already seems to be fixed in #34292

PR #34078, opened 2020-01-28
Update Grass7Utils.py

ilikeqgis:
With Windows, in OSGeo4W64 installation, the path to the GRASS executables is in the form: grass + version number without dash or dot.
Example: C:\OSGeo4W64\apps\grass\grass78
And in no case under the name: grass-7.8
In fact, I changed the code for this search to work by replacing "grass-7." by "grass7"

## Checklist

### DIFF
--- a/python/plugins/processing/algs/grass7/Grass7Utils.py
+++ b/python/plugins/processing/algs/grass7/Grass7Utils.py
@@ -220,7 +220,7 @@ class Grass7Utils:
                 else:
                     testfolder = os.path.join(os.path.dirname(QgsApplication.prefixPath()), 'grass')
                     if os.path.isdir(testfolder):
-                        grassfolders = sorted([f for f in os.listdir(testfolder) if f.startswith("grass-7.") and os.path.isdir(os.path.join(testfolder, f))], reverse=True, key=lambda x: [int(v) for v in x[len("grass-"):].split('.') if v != 'svn'])
+                        grassfolders = sorted([f for f in os.listdir(testfolder) if f.startswith("grass7") and os.path.isdir(os.path.join(testfolder, f))], reverse=True, key=lambda x: [int(v) for v in x[len("grass-"):].split('.') if v != 'svn'])
                         if grassfolders:
                             folder = os.path.join(testfolder, grassfolders[0])
             elif isMac():


gioman:
@ilikeqgis how this PR differs from https://github.com/qgis/QGIS/pull/34076 ?

jef-n:
> @ilikeqgis how this PR differs from #34076 ?

different target branch.

jef-n:
GRASS should be used with the "with GRASS" shortcut - that sets GISBASE and is used to determine the directory.   The changed part is in that case never triggered. The order part (key=) is also not considering the new naming.

ilikeqgis:
I propose to reproduce in QGIS 3 a solution that already existed in version 2.

GRASS is present in "Processing Toolbox" for any version of QGIS (QGIS 2.x/3.x with or without GRASS).
With the "QGIS with GRASS" version, we can launch it in addition via the Plugins menu and make it appear in a dedicated window "Grass Tools"...

"Grass Tools" stay different of Grass inside "Processing Toolbox".

It worked with QGIS 2.18 to use GRASS inside "Processing Toolbox" and my solution allows it to work also via QGIS 3.x. However, this solution does not load the dedicated part "Grass Tools" and present in the Plugins menu, only available with "QGIS with GRASS".

See the 5 slides (pics) of the following link : [https://www.flickr.com/gp/186661453@N05/EuN34K](url)

luipir:
> See the 5 slides (pics) of the following link : [https://www.flickr.com/gp/186661453@N05/EuN34K](url)

link does not work,... btw seems, from comment by @jef-n , that is "won't fix". If so, please close it.

ilikeqgis:
I hope that you can see this new link (or copy it) 
https://www.flickr.com/gp/186661453@N05/EuN34K

To clarify my explanations:
In QGIS Desktop 3.10.2 / 3.11.0 (without GRASS), I don't understand that we can see GRASS in the processing toolbox if it is forbidden to use it inside for this simple QGIS version (different from "QGIS Desktop 3.10.2 / 3.11.0 with GRASS 7.8.2").
I would prefer to display GRASS in the "Processing Toolbox" only when we use "QGIS Desktop 3.xx.x with GRASS 7.8.2".

However, what I wanted to underline and correct in the grass7Utils.py file is the name of the GRASS sub-directory  in Windows : it is written "grass78" and not "grass-7.8"...
Before the GRASS directories on Windows were written grass + version number with dash and dot (for example : "grass-7.6.1").
From now on, they are written in the following form : grass + simplified version number without dash or dot (for example : "grass78").

This is why my proposal seeks to replace the terms "-7." with "7" and prevent the code (line 223) from looking for a directory "grass-7." that doesn't exist, but preferably one "grass7" which exists...

Anyway, thank you for your involvement in this work.
Olivier ATHIMON

alexbruy:
When you start QGIS using any shortcut, Processing just loads algorithm descriptions which are always available and not part of the GRASS install. But as @jef said, to have GRASS working you should use correct shortcut.

ilikeqgis:
I drop my contribution because it seems normal to display the Grass tools in the Desktop version without being able to use these Grass tools in this environment. I would have however thought normal not to display it or if display to be able to activate them while avoiding having an error message what we have in all cases currently.

PR #34079, opened 2020-01-28
Update Grass7Utils.py

ilikeqgis:
qgis 3.4.15 is compiled with grass 7.8.2.
I declare grass 7.8 at first.
On the other hand, under windows, GRASS is saved in a directory in the form grass + version number without dash or point.

With Windows, in OSGeo4W64 installation, the path to the GRASS executables is in the form: grass + version number without dash or dot.
Example: C:\OSGeo4W64\apps\grass\grass78
And in no case under the name: grass-7.8
In fact, I changed the code for this search to work by replacing "grass-7." by "grass7"

## Checklist

### DIFF
--- a/python/plugins/processing/algs/grass7/Grass7Utils.py
+++ b/python/plugins/processing/algs/grass7/Grass7Utils.py
@@ -165,8 +165,8 @@ class Grass7Utils:
                 ]
         else:
             cmdList = [
-                "grass76", "grass74", "grass72", "grass70", "grass",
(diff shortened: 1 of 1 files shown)


ilikeqgis:
I drop my contribution because it seems normal to display the Grass tools in the Desktop version without being able to use these Grass tools in this environment. I would have however thought normal not to display it or if display to be able to activate them while avoiding having an error message what we have in all cases currently.

PR #34058, opened 2020-01-27
Remove add vertex circular

lbartoletti:
## Description
From https://github.com/qgis/QGIS/issues/29688 : Tool Add Circular String doesn't delete last vertex with "Del" or "Backspace" Key.

I introduce the backspace key to delete the last vertex

bug_del_vertex webm

## Checklist

- [X] Commit messages are descriptive and explain the rationale for changes
- [X] Commits which fix bugs include `Fixes #11111` at the bottom of the commit message
- [X] I have read the [QGIS Coding Standards](https://docs.qgis.org/testing/en/docs/developers_guide/codingstandards.html) and this PR complies with them
- [ ] New unit tests have been added for core changes
- [X] I have run [the `scripts/prepare-commit.sh` script](https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTING.md#contributing-to-qgis) before each commit
- [X] I have evaluated whether it is appropriate for this PR to be backported, backport requests are left as label or comment

PR #34080, opened 2020-01-28
Allow undo del vertex for shape tools

lbartoletti:
## Description
Same as https://github.com/qgis/QGIS/pull/34058 for shape tools

## Checklist

### DIFF
--- a/src/app/qgsmaptoolcircle2points.cpp
+++ b/src/app/qgsmaptoolcircle2points.cpp
@@ -34,10 +34,10 @@ void QgsMapToolCircle2Points::cadCanvasReleaseEvent( QgsMapMouseEvent *e )
 
   if ( e->button() == Qt::LeftButton )
   {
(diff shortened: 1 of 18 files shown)


src/app/qgsmaptoolcircle2points.cpp
```diff
@@ -34,7 +34,8 @@ void QgsMapToolCircle2Points::cadCanvasReleaseEvent( QgsMapMouseEvent *e )
 
   if ( e->button() == Qt::LeftButton )
   {
-    mPoints.append( point );
+    if ( mPoints.size() < 1 )
```
nyalldawson:
```suggestion
    if ( mPoints.empty() )
```

And doesn't this make the condition on the next if redundant?

lbartoletti: Right. Done.

nyalldawson (approved):

PR #34101, opened 2020-01-29
Fix crash in server feature info due to storing incorrect reference to temporary

nyalldawson:
Seriously, auto is just plain dangerous and makes for fragile code. Let's all just stop using it PLEASE?

### DIFF
--- a/src/analysis/mesh/qgsmeshcontours.cpp
+++ b/src/analysis/mesh/qgsmeshcontours.cpp
@@ -90,7 +90,7 @@ QgsGeometry QgsMeshContours::exportPolygons(
     if ( !mScalarActiveFaceFlagValues.active( nativeIndex ) )
       continue;
 
-    const auto &triangle = mTriangularMesh->triangles().at( i );
+    const QgsMeshFace &triangle = mTriangularMesh->triangles().at( i );
     const int indices[3] =
     {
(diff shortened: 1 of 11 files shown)


rouault:
> Seriously, auto is just plain dangerous and makes for fragile code

One can discuss the merits or non-merits of auto, but I don't think it is related at all with the issue here

Given
```
struct Foo
{
    int x = 0;
};

struct Bar
{
    Foo f;
    const Foo& getF() const { return f; }
};

static Bar getBar()
{
    Bar b;
    return b;
}

int main()
{
    const Foo& x = getBar().getF();
    const auto& y = getBar().getF();
    return x.x + y.x;
}
```
how does `const Foo& x = getBar().getF()` is obviously more wrong than `const auto& y = getBar().getF();` ? I'd claim that it doesn't change anything to detect erroneous use of reference to temporary

nyalldawson:
@rouault 

I'd say it's much much easier to pickup this particular bug --- because in this case the returned value is an implicitly shared class, and in a review I'd flag storing a reference to a shared class as something which should be fixed.

With auto I have absolutely no way of knowing what "queryLayersNickname" returns (without firing up an ide). It could easily be some large Qgs class which **should** be stored as a reference!

PR #34102, opened 2020-01-29
Fix crash in QgsGeometryDuplicateCheck due to heap use after free

nyalldawson:


nyalldawson:
Unrelated test failure

PR #34103, opened 2020-01-29
Fix crash in selective masking test due to early python garbage collection

nyalldawson:


nyalldawson:
Unrelated test failure

PR #34105, opened 2020-01-29
Remove deprecated QBuffer constructor usage

nyalldawson:
Not much info on why this is deprecated -- the only thing I can find is https://codereview.qt-project.org/c/qt/qt3d/+/205392 "Deprecate QBuffer::type. It can also be removed from the backend class, now the renderer doesn't look at it anyway."


wonder-sk (approved):
Looks good - yeah it always looked superfluous to have the type there, so it's good to see it go :-)

PR #30155, opened 2019-06-10
QObjectUniquePtr

m-kuhn:
Is basically a `QPointer` with ownership.

The encapsulated QObject will be deleted as soon as the (any) `QObjectUniquePtr` scopes out.

PR #34106, opened 2020-01-29
Fix deprecated QWeakPointer use of QObject

nyalldawson:


m-kuhn:
See https://github.com/qgis/QGIS/pull/30155#issuecomment-500747467

QPointer internally still uses QWeakPointer, so if it's safe to use in the context there it should also be safe to use by us I guess?

https://code.woboq.org/qt5/qtbase/src/corelib/kernel/qpointer.h.html#QPointer::wp

nyalldawson:
Odd. I really wish they'd give more explanations about deprecated methods in the docs, and not just "its use is discouraged". 

But in any case, it's about to be removed so we'll need to change this code...

m-kuhn:
Can we wait for removal and see how they handle it on their end (see link above) and follow their lead?

nyalldawson:
I assume they've dealt with it -- they are unlikely to have deprecated it without adapting the qt code. I assume they've avoided it by some template trickery.

Fwiw, KDE dealt with this by just using qpointer: https://phabricator.kde.org/D15103 (by the comments on that pr, they were stabbing in the dark too. Qt upstream really needs to lift their documentation game here)

m-kuhn:
This code is from Qt 5.13.1 (see bottom of the page)

https://code.woboq.org/qt5/qtbase/src/corelib/kernel/qpointer.h.html#QPointer::wp

nyalldawson:
Right - but why isn't it raising warnings in qt? 

https://development.qt-project.narkive.com/UYCFSOnI/undeprecating-qpointer seems to suggest qpointer would be the correct choice here in any case (I trust Thiago's knowledge of this...)

nyalldawson:
More clues here: https://git.sailfishos.org/faenil/qtbase/commit/ee7fa67eb0060b9c794bae4c7b06110c3b9626e0

nyalldawson:
Ok so they are totally cheating -- they added the overload with an unused bool argument, and the assign method, to avoid the deprecation warnings...

m-kuhn:
:rofl:

nyalldawson:
@m-kuhn so are you happy with the swap to QPointer?

m-kuhn:
I think we should use the cheater-overloads and wait for Qt to remove things and see how they deal with it.

nyalldawson:
> I think we should use the cheater-overloads and wait for Qt to remove things and see how they deal with it.

Compiler says no. They are private members, only usable by QPointer because it's declared a friend class.

nyalldawson:
@m-kuhn can I merge in this form? Rationale:
1. It's what Qt upstream is clearly pushing us toward -- we shouldn't be using QWeakPointer with QObject
2. We can't apply the same approach used by upstream internally
3. I want to enable QT_DEPRECATED_SINCE our codebase ASAP,  so that I can move on with life and stop dealing with these deprecation warnings!

m-kuhn:
Ok, let's go. Because of 3 and 2. :+1:

PR #33921, opened 2020-01-20
QGIS Development Server Application

elpaso:
## QGIS Development HTTP Server

This is a small independent command line application that implements a minimal HTTP development server for QGIS Server.

It can be useful when developing and testing QGIS Server projects, modules and plugins without requiring a full web server/FCGI stack.

qgis_mapserver

The application is completely independent it does not introduce any change in the existing QGIS code, for this reason I'd like to merge it for 3.12 even if we are in feature freeze.

```bash
Usage: qgis_mapserver [options] [address:port]
QGIS Development Server

Options:
  -h, --help        Displays this help.
  -v, --version     Displays version information.
  -l <logLevel>     Sets log level (default: 0)
                    0: INFO
                    1: WARNING
                    2: CRITICAL
  -p <projectPath>  Path to a QGIS project file (*.qgs or *.qgz),
                    if specified it will override the query string MAP argument
                    and the QGIS_PROJECT_FILE environment variable

Arguments:
  addressAndPort    Listen to address and port (default: "localhost:8000")
                    address and port can also be specified with the environment
                    variables QGIS_SERVER_ADDRESS and QGIS_SERVER_PORT
```

Sample output:
```bash
QGIS Development Server listening on http://localhost:8000
CTRL+C to exit
127.0.0.1 [lun gen 20 15:16:41 2020] 5140 103ms "GET /wfs3/?MAP=/home/ale/dev/QGIS/tests/testdata/qgis_server/test_project.qgs HTTP/1.1" 200
127.0.0.1 [lun gen 20 15:16:41 2020] 3298 2ms "GET /wfs3/static/jsonFormatter.min.js HTTP/1.1" 200
127.0.0.1 [lun gen 20 15:16:41 2020] 1678 3ms "GET /wfs3/static/jsonFormatter.min.css HTTP/1.1" 200
127.0.0.1 [lun gen 20 15:16:41 2020] 1310 5ms "GET /wfs3/static/style.css HTTP/1.1" 200
127.0.0.1 [lun gen 20 15:16:43 2020] 4285 13ms "GET /wfs3/collections?MAP=/home/ale/dev/QGIS/tests/testdata/qgis_server/test_project.qgs HTTP/1.1" 200
```

PR #34124, opened 2020-01-29
Revert "QGIS Development Server Application"

elpaso:
Reverts qgis/QGIS#33921

After some more complex test I noticed that the current implementation is crashing too often (it's a bit better with parallel rendering off, but it doesn't completely fix the issue). 

Better to wait for a more robust implementation.


elpaso:
Wait, maybe it's not too bad.. I have a PR that it's not too crashy.

Gustry:
Can you also remove the branch? It's on the main repo when we git fetch. Thanks

elpaso:
@Gustry done, sorry I forgot.

PR #34126, opened 2020-01-29
Server standalone lambda disconnect

elpaso:
This fixes a couple of issues with the development server under load.

There are still too many processEvents in server code that might call the connection destructor while inside the lambda, but with parallel rendering disabled the development server behaves much better.

### DIFF
--- a/src/server/qgis_mapserver.cpp
+++ b/src/server/qgis_mapserver.cpp
@@ -36,7 +36,7 @@ on the command line.
 #include <QNetworkInterface>
 #include <QCommandLineParser>
 #include <QObject>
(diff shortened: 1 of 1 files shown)


elpaso:
@pblottiere I could use your help here :) 
Maybe you find a way to make it work with parallel rendering on.

pblottiere:
Hi @elpaso,

> Maybe you find a way to make it work with parallel rendering on.

I'll give it  try, but feel free to merge this PR in the meantime :+1:

elpaso:
@pblottiere I think I've found a way around the event loop and processEvents hell.

PR #34130, opened 2020-01-29
Fix cppcheck nullPointer/nullPointerRedundantCheck warnings in src/core/

rouault:
Found with
~/cppcheck/cppcheck --enable=all --inconclusive ../src/core --library=$HOME/cppcheck/cfg/qt.cfg -j 8 -DSIP_OUT 2>&1 | grep nullPointer

Mostly false positives or indeed redundant check, except perhaps in
QgsSingleBandPseudoColorRenderer::toSld() where a null pointer dereference
could occur in case of failed dynamic_cast.

### DIFF
--- a/src/core/expression/qgsexpressionfunction.cpp
+++ b/src/core/expression/qgsexpressionfunction.cpp
@@ -743,8 +743,10 @@ static QVariant fcnAggregateRelation( const QVariantList &values, const QgsExpre
     }
   }
 
+  if ( !context->hasFeature() )
+    return QVariant();
+  QgsFeature f = context->feature();
 
-  FEAT_FROM_CONTEXT( context, f )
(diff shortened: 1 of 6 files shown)


src/core/expression/qgsexpressionfunction.cpp
```diff
@@ -743,16 +743,17 @@ static QVariant fcnAggregateRelation( const QVariantList &values, const QgsExpre
     }
   }
 
+  if ( !context->hasFeature() ) return QVariant();
```
nyalldawson:
```suggestion
  if ( !context->hasFeature() )
    return QVariant();
```

nyalldawson (approved):

src/core/expression/qgsexpressionfunction.cpp
rouault: done

Issue #34089, opened 2020-01-28
Draw effects units setting doesn't persist


Houska1:
(Vector, line) Symbology / Simple Line / Draw effects / Inner or outer glow unit selection for blur radius doesn't persist.

Example - create an outer glow for the line, set the spread to ... pixels and blur radius to ... pixels. This means changing the units for each from mm to pixels.

![image](https://user-images.githubusercontent.com/48919900/73262878-44503700-419d-11ea-8d4e-81782183cb33.png)

Click OK to leave draw effects. Reopen draw effects. The blur radius units are back to mm.
![image](https://user-images.githubusercontent.com/48919900/73262947-69dd4080-419d-11ea-988f-bbb6ac606203.png)

Also happens for inner glow but not for drop shadow. Haven't tested the rest, nor have I investigated whether the blur radius is actually rendered as pixels or mm (it's a subtle setting...).

Houska1:
Update also happens if you choose the blur radius units to something else (e.g. inches), not just pixels. Using 3.10.2 code revision d4cd3cfe5a (latest full not dev OSGeo4W release).

PR #34135, opened 2020-01-30
Fix cppcheck uninitMemberVar warnings in src/core/

rouault:
Found with
~/cppcheck/cppcheck --enable=all --inconclusive ../src/core --library=$HOME/cppcheck/cfg/qt.cfg -j 8 -DSIP_OUT 2>&1 | grep uninitMemberVar


src/core/qgsfeaturerequest.cpp
```diff
@@ -56,7 +56,7 @@ QgsFeatureRequest::QgsFeatureRequest( const QgsExpression &expr, const QgsExpres
 {
 }
 
-QgsFeatureRequest::QgsFeatureRequest( const QgsFeatureRequest &rh )
+QgsFeatureRequest::QgsFeatureRequest( const QgsFeatureRequest &rh ): mTransformContext{}
```
rouault: I'm wondering here if mTransformContext shouldn't rather be copied by the operator= ?

nyalldawson: Yes, you're right...

src/core/effects/qgsgloweffect.cpp
```diff
@@ -33,6 +33,8 @@ QgsGlowEffect::QgsGlowEffect( const QgsGlowEffect &other )
   , mSpreadUnit( other.spreadUnit() )
   , mSpreadMapUnitScale( other.spreadMapUnitScale() )
   , mBlurLevel( other.blurLevel() )
+  , mBlurUnit( other.mBlurUnit )
+  , mBlurMapUnitScale( other.mBlurMapUnitScale )
```
nyalldawson: Likely fixes https://github.com/qgis/QGIS/issues/34089

src/core/qgsfeaturerequest.cpp
rouault: done

src/core/effects/qgsgloweffect.cpp
rouault: No, it doesn't

nyalldawson: Maybe because it's missing in the operator= method?

rouault: They were indeed missing in operator=, but fixing it doesn't fix the issue. Perhaps a pure GUI issue

nyalldawson: Yep - gui issue. I'll fix after this is merged.

rouault: > I'll fix after this is merged.

I'm on it. Fix under way

rouault: Fixed by https://github.com/qgis/QGIS/pull/34135/commits/01412a7bec32d736d13a6166f3ecbc2f5230284c . The fix is only effective if combined with the copy constructor/operator= fix . So this whole PR should probably be backported

PR #34139, opened 2020-01-30
Friendlier names for input layers when defining model child algorithm inputs

nyalldawson:
When listing available layer sources for an input in the model designer, prefer to show the user's defined name for that output instead of the generic one

Makes it easier to associate inputs with the correct outputs when creating models


luipir (approved):
yes please

nyalldawson:
Unrelated test failures

PR #34140, opened 2020-01-30
Fix Points Along Geometry handling of multipart geometries

### LINKED ISSUE #34022: "Points Along Geometry" not creating points if geometry is multipart
I found the result of qgis:pointsalonglines with v3.10 and 3.2.2 are different, exactly the new version gives a wrong result. see the screenshot attatched.

nyalldawson:
Fixes #34022


nyalldawson:
Unrelated test failure

PR #34142, opened 2020-01-30
Fix QStringList values are not accepted as valid values 

nyalldawson:
…for multi field parameters


nyalldawson:
Unrelated test failure

PR #34110, opened 2020-01-29
Find PyQt5 sip files on Archlinux

pblottiere:
## Description

After my last Archlinux upgrade, QGIS compilation is failing because PyQt5 sip files cannot be found:

``` bash
[9/973] Generating analysis/sip_analysispart0.cpp, analysis/sip_analysispart1.cpp, analysis/sip_analysispart2.cpp, analysis/sip_analysispart3.cpp
FAILED: python/analysis/sip_analysispart0.cpp python/analysis/sip_analysispart1.cpp python/analysis/sip_analysispart2.cpp python/analysis/sip_analysispart3.cpp
cd /home/pblottiere/devel/qgis/qgis_pbl_master/build/python && /usr/bin/cmake -E echo && /usr/bin/cmake -E touch /home/pblottiere/devel/qgis/qgis_pbl_master/build/python/analysis/sip_analysispart0.cpp /home/pblottiere/devel/qgis/qgis_pbl_master/build/python/analysis/sip_analysispart1.cpp /home/pblottiere/devel/qgis/qgis_pbl_master/build/python/analysis/sip_analysispart2.cpp /home/pblottiere/devel/qgis/qgis_pbl_master/build/python/analysis/sip_analysispart3.cpp && /usr/bin/sip -w -e -x ANDROID -x ARM -x MOBILITY_LOCATION -n PyQt5.sip -t Qt_5_14_0 -t WS_X11 -g -o -a /home/pblottiere/devel/qgis/qgis_pbl_master/build/python/qgis.analysis.api -n PyQt5.sip -y /home/pblottiere/devel/qgis/qgis_pbl_master/build/output/python/qgis/_analysis.pyi -j 4 -c /home/pblottiere/devel/qgis/qgis_pbl_master/build/python/analysis -I /home/pblottiere/devel/qgis/qgis_pbl_master/build/python/analysis -I /usr/share/sip -I /home/pblottiere/devel/qgis/qgis_pbl_master/python /home/pblottiere/devel/qgis/qgis_pbl_master/build/python/analysis/analysis.sip
sip: Unable to find file "QtCore/QtCoremod.sip"
```

Actually the default SIP dir for these files is `/usr/share/sip`:

```
 Python 3.8.1 (default, Jan 22 2020, 06:38:00)
[GCC 9.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sipconfig
>>> sipconfig.Configuration().default_sip_dir
'/usr/share/sip'
```

[...]

PR #34146, opened 2020-01-30
Find PyQt5 sip files on Archlinux

pblottiere:
## Description

Backport of https://github.com/qgis/QGIS/pull/34110


luipir (approved):
I suppose there no way to check on CI

PR #34152, opened 2020-01-30
Fix widget wrong values 0 doubles

### LINKED ISSUE #34118: QGIS 3.10.2 replace 0 to NULL
I have a PostGIS table with numeric(9,0) field. 
In QGIS 3.10.2 in Feature attributes window NULL appears instead of 0.
The attribute table show the correct value, but when I edit a record, and place 0 to the field it saves NULL to the database.

QGIS 3.10.1 and 3.4.15 are working fine.

client:
QGIS 3.10.2 Windows 10

server:
PostgreSQL 9.6.5, compiled by Visual C++ build 1800, 64-bit
PostGIS 2.4

0toNULL

elpaso:
Fixes #34118

Because:

QVariant(0.0) == QVariant(QVariant.Double) -> True

but:

QVariant(0.0).isNull() == QVariant(QVariant.Double).isNull() -> False

### DIFF
--- a/src/gui/editorwidgets/qgstexteditwrapper.cpp
+++ b/src/gui/editorwidgets/qgstexteditwrapper.cpp
@@ -58,9 +58,14 @@ QVariant QgsTextEditWrapper::value() const
     v = mLineEdit->text();
   }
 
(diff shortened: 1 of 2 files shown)


troopa81 (approved):
looks ok to me, thanks @elpaso

elpaso:
@troopa81 this needs backporting, correct?

troopa81:
> @troopa81 this needs backporting, correct?

Yes, I think so

PR #34153, opened 2020-01-30
Activate buffer drawing when setting a buffer property through auxiliary storage

### LINKED ISSUE #28544: [Auxiliary storage] Labels Buffer DD options do not have fields set and do not render in map canvas
Original Redmine Issue: [20724](https://issues.qgis.org/issues/20724)
Affected QGIS version: 3.4.2
Redmine category:map_tools

---
Enable labeling on a layer
Use the "change label" tool from the label toolbar, creating a data storage auxiliary DB.
In the change label properties dialog, there are some buffer related options (size, color) that you can change.
Modify the buffer properties: nothing happens in map canvas.
However, looking into the Layer properties dialog --> Attributes form (or Auxiliary storage) tab, you have both auxiliary_storage_labeling_buffercolor and auxiliary_storage_labeling_buffersize that you can unhide and modify. no changes either.

Give a look to the layer properties --> Labels --> Buffer tab, the properties are grayed while they should be enabled with yellow widgets for some.

pblottiere:
## Description

Fixes #28544


nyalldawson:
Wouldn't this mean that the following occurs:

1. setup labeling for a layer, no buffer enabled or wanted
2. click the "label properties" tool, click a label, change something non-buffer related (e.g. text size)
3. the buffer is accidentally enabled for all labels in the layer

nirvn:
The proper fix is on the GUI end of things. I'd enable buffer settings widgets (ie setEnabled()) if enable buffer is checked *or* if a data defined expression is used to enable the buffer.

nyalldawson:
@nirvn we already do that -- I don't think that relates to the bug report

nirvn:
Hm OK, read the issue too quickly.

Haven't tried on my master build just now, I think the proper fix would be to add a [x] buffer checkbox in the change label properties dialog (instead of merely relying on setting buffer size and color). 

If we don't want to add a checkbox, then we'd need to come up with a expression to activate buffer instead of always activating buffer that'd rely on auxiliary field buffer size or color values not being null.

pblottiere:
Hi @nirvn and @nyalldawson,

> 3. the buffer is accidentally enabled for all labels in the layer

Indeed.

> Haven't tried on my master build just now, I think the proper fix would be to add a [x] buffer checkbox in the change label properties dialog 

Adding a dedicated buffer checkbox seems to be the simpler and the more explicit solution. I'm going to do that.

Thanks for your inputs :+1:

pblottiere:
>> Haven't tried on my master build just now, I think the proper fix would be to add a [x] buffer checkbox in the change label properties dialog

> Adding a dedicated buffer checkbox seems to be the simpler and the more explicit solution. I'm going to do that.

About the issue raised by @nyalldawson, maybe we could set the default buffer size to 0. This way, once the user has explicit checked the buffer checkbox in the label properties dialog:

1. we activate the buffer setting for the whole layer
2. we set the buffer size to 0 to avoid side effects on other labels 
3. we create corresponding auxiliary fields for the buffer
4. we set the buffer size for the current label according to the configuration of the label properties dialog

What do you think?

nirvn:
-1 to set default buffer size to 0, that's bad UX. When users are activating the buffer checkbox, it's good for them to already see some visual change on the canvas.

I really think this is about getting a dedicated buffer activation auxiliary field, that's toggled via checkbox in the individual label properties dialog.

pblottiere:
> I really think this is about getting a dedicated buffer activation auxiliary field, that's toggled via checkbox in the individual label properties dialog.

Well, it'll be more elegant, you're right :+1:. Thanks @nirvn.

pblottiere:
Hi @nirvn,

I added a dedicated checkbox in the individual label properties dialog. Let me know what you think :).

pblottiere:
I'll merge this PR in the coming days if I don't have comments in the meantime.

PR #34162, opened 2020-01-31
Port optimisations from Join by Location algorithm to Extract by Location and Select by Location

nyalldawson:
... allowing an optimised iteration when extracting few features from a reference layer containing many.

Also add warnings when no spatial index exists yet is desirable


alexbruy (approved):

nyalldawson:
Unrelated test failure

PR #34168, opened 2020-01-31
Disable PG flacky test case

elpaso:
test fails on Travis without a reason


elpaso:
CC @nyalldawson

m-kuhn:
Can you provide a link to a failing test for reference?

elpaso:
@m-kuhn like this? https://travis-ci.org/qgis/QGIS/jobs/643991374#L2987

m-kuhn:
Yes, I think this should be mandatory for flaky tests :+1:

PR #34175, opened 2020-01-31
WFS case insensitive KVP URL query arguments check

### LINKED ISSUE #34148: QGIS doesn't respect OGC guidelines for KVP parameters
**Describe the bug**

QGIS makes unwarranted case sensitive comparisons of KVP keys

**How to Reproduce**

In the latest WFS 2.0 specification (and all previous specifications):
"09-025r1_OGC_Web_Feature_Service_WFS_2.0_ISOFDIS_19142_Geographic_information_-_Web_Feature_Service.pdf"

it clearly states in section "6.2.5.2 Parameter ordering and case" (my emphasis of NOT):

"Parameter names shall NOT be case sensitive, but parameter values shall be case sensitive. In this International Standard, parameter names used in KVP encoding are typically shown in uppercase for typographical clarity, NOT as a requirement."

**QGIS and OS versions**

Any version including 3.4.15, Windows 10

**Additional context**

I provide two non-standard "Login=" and "Password=" parameters in a WFS stub url, which some software of ours changed to lowercase, which prompted QGIS to re-insert the two parameters again (because clearly "login" was deemed different than "Login"), yielding an error. A case-insensitive comparision would have prevented this, which should have been performed re: the standard.

Please fix this.

elpaso:
Fixes #34148

### DIFF
--- a/src/providers/wfs/qgswfsdatasourceuri.cpp
+++ b/src/providers/wfs/qgswfsdatasourceuri.cpp
@@ -202,11 +202,18 @@ QUrl QgsWFSDataSourceURI::requestUrl( const QString &request, const Method &meth
         // One could argue that the server should expose them in the DCP endpoint.
         url = QUrl( mGetEndpoints[ request ] );
         urlQuery = QUrlQuery( url );
+        // OGC case insensitive keys KVP: see https://github.com/qgis/QGIS/issues/34148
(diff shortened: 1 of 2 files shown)


elpaso:
@rouault mind having a look?

rouault:
LGTM

elpaso:
@rouault thank you for stepping by :+1:

PR #34182, opened 2020-02-01
Fixed vertex tool selection on a rotated map canvas

uclaros:
## Description

The vertex tool rectangular selection selects wrong vertices on a rotated map.

Peek 2020-02-01 03-27

To fix this, the simple select tool logic has been applied to the vertex tool.

## Checklist

### DIFF
--- a/src/app/vertextool/qgsvertextool.cpp
+++ b/src/app/vertextool/qgsvertextool.cpp
@@ -476,7 +476,7 @@ void QgsVertexTool::cadCanvasPressEvent( QgsMapMouseEvent *e )
 
     // the user may have started dragging a rect to select vertices
     if ( !mDraggingVertex && !mDraggingEdge )
-      mSelectionRectStartPos.reset( new QPoint( e->pos() ) );
+      mSelectionRubberBandStartPos.reset( new QPoint( e->pos() ) );
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/app/vertextool/qgsvertextool.cpp
```diff
@@ -534,7 +536,7 @@ void QgsVertexTool::cadCanvasReleaseEvent( QgsMapMouseEvent *e )
       if ( mLockedFeature && mLockedFeature->layer() != vlayer )
         continue;  // with locked feature we only allow selection of its vertices
 
-      QgsRectangle layerRect = toLayerCoordinates( vlayer, map_rect );
+      QgsRectangle layerRect = toLayerCoordinates( vlayer, rubberBandGeometry.boundingBox() );
```
nyalldawson: map_rect can be completely removed now

src/app/vertextool/qgsvertextool.cpp
```diff
@@ -2430,28 +2432,28 @@ void QgsVertexTool::highlightAdjacentVertex( double offset )
   zoomToVertex( vertex );  // make sure the vertex is visible
 }
 
-void QgsVertexTool::startSelectionRect( QPoint point0 )
+void QgsVertexTool::initSelectionRubberBand()
```
nyalldawson: These changes seem unrelated -- can you defer these to a follow-up PR?

uclaros: Sorry, could you be more specific? Do you mean renaming the methods?

nyalldawson: I mean that the actual fix is just the change in cadCanvasReleaseEvent from using the rect to using the geometry (which is a great fix, thanks!). The rest of these changes are unrelated (e.g. why have you changed the colors of the rubber band?) Those are the bits which should be left to a separate PR for discussion.

uclaros: Well, in order to get the geometry I changed from using a QRubberBand to a QgsRubberBand, but it defaults to different colors so I tried to make it look just like before so the ux doesn't change (couldn't get the exact color QRubberBand uses by default, maybe it's system specific?).
Then the start-update-stop methods had to be adapted for mSelectionRubberBand.
After that, renaming the methods to reflect that they fiddle with the SelectionRubberBand seemed logical but obviously not necessary. The rest though I think are necessary because of the transition from QRubberBand to QgsRubberBand

nyalldawson: Thanks for the explanation! I missed the QRubberBand -> QgsRubberband swap here, I'll re-review accordingly

nyalldawson (approved):

PR #34184, opened 2020-02-01
[azure] Switch osgeo mirror to osuosl

m-kuhn:
Because 🚀

## Checklist

### DIFF
--- a/.ci/azure-pipelines/azure-pipelines.yml
+++ b/.ci/azure-pipelines/azure-pipelines.yml
@@ -146,7 +146,7 @@ jobs:
 #        osgeo4w | $(OSGEO4W_ARCH)
 #    displayName: Cache OSGeo4W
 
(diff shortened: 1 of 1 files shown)


nirvn:
How much of a speed gain are we talking about here?

m-kuhn:
On QField it brought download time from 30 minutes to 6.5 minutes. So roughly a factor of 5.

nirvn:
🤜🤛

m-kuhn:
Seems to be only a factor of 1.5 to 2 here. But still 10-15 minutes down.

m-kuhn:
Heh, I forgot to update the most important path. The last run was just faster by pure chance.
After doing the real change we are down from 20-30 minutes before to 6 minutes now.

nirvn:
I'd call that an improvement.

PR #34187, opened 2020-02-01
Create test profiles in temp

### LINKED ISSUE #34185: Runs of testsuite leaves hundreds of directories in ~/.local/share
Every test run via ctest ends up creating a directory under `~/.local/share` (on my Unix setup, not sure where else on other setups). Such directory contains a `profiles` subdir with a `default` subdir and a bunch of files. Example:
```
strk@liz:~] find ~/.local/share/qgis_clippertest/
/home/strk/.local/share/qgis_clippertest/
/home/strk/.local/share/qgis_clippertest/profiles
/home/strk/.local/share/qgis_clippertest/profiles/default
/home/strk/.local/share/qgis_clippertest/profiles/default/symbology-style.db
/home/strk/.local/share/qgis_clippertest/profiles/default/qgis.db
/home/strk/.local/share/qgis_clippertest/profiles/default/Unknown Organization
/home/strk/.local/share/qgis_clippertest/profiles/default/Unknown Organization/qgis_clippertest.ini
/home/strk/.local/share/qgis_clippertest/profiles/default/palettes
/home/strk/.local/share/qgis_clippertest/profiles/default/bookmarks.xml
```

This was reported to be expected but UNDESIRED, which I agree with.
This ticket is to keep track of the problem.

\cc @m-kuhn

m-kuhn:
Partially fixes #34185

@strk I guess something similar will still need to be done for python tests

### DIFF
--- a/src/test/qgstest.h
+++ b/src/test/qgstest.h
@@ -29,7 +29,6 @@
     QTemporaryDir dir; \
     QgsApplication app(argc, argv, false); \
     app.init( dir.path() ); \
-    dir.setAutoRemove( false ); \
     app.setAttribute(Qt::AA_Use96Dpi, true); \
     QTEST_DISABLE_KEYPAD_NAVIGATION \
     QTEST_ADD_GPU_BLACKLIST_SUPPORT \
--- a/src/test/qgstest.h
+++ b/src/test/qgstest.h
@@ -26,8 +26,10 @@
   QT_END_NAMESPACE \
   int main(int argc, char *argv[]) \
   { \
+    QTemporaryDir dir; \
     QgsApplication app(argc, argv, false); \
-    app.init(); \
+    app.init( dir.path() ); \
+    dir.setAutoRemove( false ); \
     app.setAttribute(Qt::AA_Use96Dpi, true); \
     QTEST_DISABLE_KEYPAD_NAVIGATION \
     QTEST_ADD_GPU_BLACKLIST_SUPPORT \
--- a/tests/src/core/testqgstranslateproject.cpp
+++ b/tests/src/core/testqgstranslateproject.cpp
@@ -51,7 +51,7 @@ class TestQgsTranslateProject : public QObject
 
 void TestQgsTranslateProject::initTestCase()
 {
-  //start application
+  settings.clear();
   original_locale = settings.value( QStringLiteral( "locale/userLocale" ), "" ).toString() ;
 }
 
--- a/tests/src/core/testqgstranslateproject.cpp
+++ b/tests/src/core/testqgstranslateproject.cpp
@@ -52,16 +52,12 @@ class TestQgsTranslateProject : public QObject
 void TestQgsTranslateProject::initTestCase()
 {
   //start application
-  QgsApplication::init();
-  QgsApplication::initQgis();
-
   original_locale = settings.value( QStringLiteral( "locale/userLocale" ), "" ).toString() ;
 }
 
 void TestQgsTranslateProject::cleanupTestCase()
 {
   settings.setValue( QStringLiteral( "locale/userLocale" ), original_locale );
-  QgsApplication::exitQgis();
 
   //delete translated project file
   QString translatedProjectFileName( TEST_DATA_DIR );


src/test/qgstest.h
```diff
@@ -26,8 +26,10 @@
   QT_END_NAMESPACE \
   int main(int argc, char *argv[]) \
   { \
+    QTemporaryDir dir; \
     QgsApplication app(argc, argv, false); \
-    app.init(); \
+    app.init( dir.path() ); \
+    dir.setAutoRemove( false ); \
```
strk: Why not autoremoving it ? Shall it be determined by some env variable ?

m-kuhn: Good point, I didn't realize that the scope is the complete main 
```suggestion
```

strk:
I guess python side we could do this in `start_app` inside `python/testing/__init__.py` ?

m-kuhn:
Sounds like a plan

strk:
Possibly a better plan: export `QGIS_CUSTOM_CONFIG_PATH` from CMake, which should work for all implementations of tests

m-kuhn:
The variable needs to be there on runtime, not build time. Not sure if CMake can handle that.

Personally, I don't like environment variables that much and use them as a last resort if everything else fails. But : whatever works!

strk:
@m-kuhn see how you like https://github.com/m-kuhn/QGIS/pull/63

tests/src/core/testqgstranslateproject.cpp
```diff
@@ -52,16 +52,12 @@ class TestQgsTranslateProject : public QObject
 void TestQgsTranslateProject::initTestCase()
 {
   //start application
-  QgsApplication::init();
```
nyalldawson: This is done in a lot of places, mostly in tests which require a custom QSettings environment.

What do you think about modifying the QGSTEST_MAIN macro so that we always use a custom QSettings environment for each test, automatically cleared before the test is executed? 

I can't see any issue if we did this, and the benefit is that it would nicely isolate local profile settings away from the tests (tests won't be affected by local settings, and local setting profiles won't be changed by the test).

m-kuhn: QGSTEST_MAIN should already do that after this PR on a per test basis. Here the issue is about a per test case basis, not something that can be fixed in QGSTEST_MAIN AFAICT. And I wouldn't be surprised if we had tests that rely on the current behavior. I'll just try to fixup the failing ones.

m-kuhn:
Let's see if azure is any happier today

m-kuhn:
@strk I think there's still an issue with running this on windows. That's why I constantly reopened to restart them but never really found the time to dive into it and check what's wrong.

strk:
I had a feeling that one red CI never blocked merging, so that it was
not mandatory to wait for that green (wouldn't I be unable to merge
if it was mandatory, otherwise?)

Soooo, do you think I should revert the merge or can we let it go ?

m-kuhn:
Windows CI is still in "incubation phase" and therefore not yet a required test. But the goal is to make it mandatory as well once it stabilizes enough.

I'm quite sure this one here introduced problems with the win test suite. Which means one step away from the goal.

I see a couple of options:

 - Fix the tests
 - Revert and reopen a new pull request
 - Mask the windows tests (not really nice, right?)

And in general: if you see a pull request like this one, please ask what's up before hitting the merge button!!

strk:
It's 4 commits, hard to revert.

For "fix the test", do you have a link to an actual build log ? Because the one from most current commit of this PR is a 404: https://dev.azure.com/qgis/QGIS/_build/results?buildId=3206

m-kuhn:
No, unfortunately not. I did restart the test 3 times because of this issue (logs are deleted after some time aparently). But the next tests on master should also show the failure (and if they don't, this means there is no issue and everyone is happy)

If "merge commit" is used instead of "rebase", revert is easier ;)

nyalldawson:
Ok, I've reverted manually

strk:
I'm really sorry @nyalldawson, @m-kuhn will never do such a thing again.
@m-kuhn I dont see a "reopen" button here, so I guess it'll take another PR :(

m-kuhn:
Yes, a new PR sounds like the way to go :+1:

PR #34201, opened 2020-02-02
Fix leak

nyalldawson:
Very minor, but a leak is a leak!


nyalldawson:
Unrelated test failure

PR #34202, opened 2020-02-02
Handle unit type "m" in custom WKT strings on proj 6 builds

### LINKED ISSUE #34196: unit "m" not handled in custom WKT string
**Describe the bug**

When loading the attached shapefile to QGIS 3.10.2 (tested with 3.10.2-2, also with master branch release), the layer has Unknown CRS.  
QGIS 3.4.15 assigns _EPSG:31287_ CRS to the same shapefile when loaded.  

The _.prj_ file has a WKT1 string (I don't know if WKT1-ESRI or WKT1(-OGC/-GDAL)), describing the _EPSG:31287_ projected CRS, but with the _MGI to WGS 84 (8) - EPSG:1194_ transformation for its datum (not the same as the default preferred one).  

If the user has defined the _"Use CRS from first layer added"_ setting, the project remains with an unknown CRS without a warning.  
As can see in [https://gis.stackexchange.com/questions/349304/qgis-3-10-2-ltr-projection-problem](https://gis.stackexchange.com/questions/349304/qgis-3-10-2-ltr-projection-problem), this causes an erroneous definition of the scale and an erroneous behavior of the distance measurement tool in planimetric mode.  
It is striking that the measurement tool in ellipsoidal mode works well.  

**How to Reproduce**

1. Tick on the _"Use CRS from first layer added"_ setting.
2. Create a new project.
3. Load the attached shapefile.
4. The Layer Properties CRS says _Unknown CRS_. The Project Properties CRS has not a CRS defined. The scale of the canvas is allways _0:1_. Define (if not already) an ellipsoid for the project and measure an ellipsoidal distance, it seems to show a right value. Measure a planimetric distance, the value seems to be right in the units of measure of the layer, but those units are not known (all units in the tool returns the same value).

**QGIS and OS versions**

QGIS version
3.10.2-A Coruña
QGIS code revision
e28feaf5fd
Compiled against Qt
5.11.2
Running against Qt
5.11.2
Compiled against GDAL/OGR
3.1.0dev
Running against GDAL/OGR
3.1.0dev
Compiled against GEOS
3.8.0-CAPI-1.13.1
Running against GEOS
3.8.0-CAPI-1.13.1 
Compiled against SQLite
3.29.0
Running against SQLite
3.29.0
PostgreSQL Client Version
11.5
SpatiaLite Version
4.3.0
QWT Version
6.1.3
QScintilla2 Version
2.10.8
Compiled against PROJ
7.0.0
Running against PROJ
Rel. 7.0.0, March 1st, 2020
OS Version
Windows 10 (10.0)
This copy of QGIS writes debugging output.
Active python plugins
processing

**Attached files**

[shapefile.zip](https://github.com/qgis/QGIS/files/4143621/shapefile.zip)
[project.zip](https://github.com/qgis/QGIS/files/4143623/project.zip)

nyalldawson:
Fixes #34196

### DIFF
--- a/src/core/qgscoordinatereferencesystem.cpp
+++ b/src/core/qgscoordinatereferencesystem.cpp
@@ -56,9 +56,10 @@
 #include <cpl_csv.h>
 
 
-
+#if PROJ_VERSION_MAJOR<6
 //! The length of the string "+lat_1="
 const int LAT_PREFIX_LEN = 7;
+#endif
 
 CUSTOM_CRS_VALIDATION QgsCoordinateReferenceSystem::sCustomSrsValidation = nullptr;
 
@@ -1683,7 +1684,8 @@ void QgsCoordinateReferenceSystem::setMapUnits()
   }
 
   PJ_CONTEXT *context = QgsProjContext::get();
-  QgsProjUtils::proj_pj_unique_ptr coordinateSystem( proj_crs_get_coordinate_system( context, d->threadLocalProjObject() ) );
+  QgsProjUtils::proj_pj_unique_ptr crs( QgsProjUtils::crsToSingleCrs( d->threadLocalProjObject() ) );
+  QgsProjUtils::proj_pj_unique_ptr coordinateSystem( proj_crs_get_coordinate_system( context, crs.get() ) );
   if ( !coordinateSystem )
   {
     d->mMapUnits = QgsUnitTypes::DistanceUnknownUnit;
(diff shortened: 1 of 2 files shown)


src/core/qgscoordinatereferencesystem.cpp
```diff
@@ -1719,7 +1721,9 @@ void QgsCoordinateReferenceSystem::setMapUnits()
          unitName.compare( QLatin1String( "hemisphere degree minute second" ), Qt::CaseInsensitive ) == 0 ||
          unitName.compare( QLatin1String( "degree (supplier to define representation)" ), Qt::CaseInsensitive ) == 0 )
       d->mMapUnits = QgsUnitTypes::DistanceDegrees;
-    else if ( unitName.compare( QLatin1String( "metre" ), Qt::CaseInsensitive ) == 0 )
+    else if ( unitName.compare( QLatin1String( "metre" ), Qt::CaseInsensitive ) == 0
```
rouault: comparing the conversion factor (here 1.0) with let's say a 1e-10 relative error threshold would probably be more robust

nyalldawson: Is the conversion factor always to metres?

rouault: For linear units, yes. And for angular units, it is to radians. If the CRS is a projected CRS, then the type of unit of the first axis is necessarily linear units. And if the CRS is a geographic CRS, the typoe of unit of the first axis is necessarily angular units.

PR #34208, opened 2020-02-03
Fix Qt53DExtrasConfig.cmake syntax which triggers error on ubuntu 20.04 (cmake 3.13)

nirvn:
## Description

@wonder-sk , @jef-n , the cmake version shipped with upcoming Ubuntu 20.04 doesn't appear to be as forgiving as previous versions. When building with 3D enabled (using our 3D extras headers workaround), I get a syntax here which this PR fixes.


Gustry:
Can you tell us what need to be done to compile with 3D?
Updating the INSTALL file would be nice. There isn't any mention in the doc how to compile with 3D (a few days ago, I wasn't aware of this external extra header available in the repo).
I can update the doc if needed as soon as my 3D is compiling :) Thanks @nirvn

nirvn:
@Gustry , this gets 3D to compile on my machine:

cmake -DWITH_3D=True -DCMAKE_PREFIX_PATH={path to QGIS' git repo}/external/qt3dextra-headers/cmake -DQT5_3DEXTRA_INCLUDE_DIR={path to QGIS' git repo}/external/qt3dextra-headers -DQT5_3DEXTRA_LIBRARY=/usr/lib/x86_64-linux-gnu/libQt53DExtras.so {path to QGIS' git repo}

Gustry:
I will give it a try. Thanks

nyalldawson (approved):

PR #34212, opened 2020-02-03
Fix detection of bound CRSes when creating a CRS from WKT (fixes #34211)

### LINKED ISSUE #34211: QGIS fails to get CRS of inserted shapefile under gdal3/proj6
**Describe the bug**

That didn't take long. 

Using a  gdal3/proj6 QGIS build, inserting a shapefile layer fails to assign the proper CRS (wrongly flagged as Unknown CRS).

An error/warning message is printed in the console:

`src/core/qgscoordinatetransform_p.cpp:289 : (proj_logger) [40ms] proj_crs_get_coordinate_system: Object is not a SingleCRS`

**How to Reproduce**

1. Launch a gdal3/proj6 QGIS build, and create a new project
2. Add the attached dataset onto the project:
[ELCs2012.zip](https://github.com/qgis/QGIS/files/4147133/ELCs2012.zip)

3. The CRS of the shapefile is EPSG:3148, yet QGIS registers the shapefile as having an Unknown CRS

**QGIS and OS versions**

QGIS master, Ubuntu 20.04, gdal 3.0.4, proj 6.3.0.

nirvn:
## Description

@nyalldawson , this PR fixes issue #34211 . Do you think the increased flexibility when creating a CRS from a WKT is acceptable here? You know best.

### DIFF
--- a/src/core/qgscoordinatereferencesystem.cpp
+++ b/src/core/qgscoordinatereferencesystem.cpp
@@ -1587,7 +1587,7 @@ bool QgsCoordinateReferenceSystem::setWktString( const QString &wkt, bool allowP
     if ( authName.isEmpty() || authCode.isEmpty() )
     {
       // try 2, use proj's identify method and see if there's a nice candidate we can use
-      QgsProjUtils::identifyCrs( d->threadLocalProjObject(), authName, authCode );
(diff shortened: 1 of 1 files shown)


nyalldawson:
No, we shouldn't do that here. It's a different crs when it's bound.

PR #34216, opened 2020-02-03
[WIP] BUG: postgres serial id conflict during an upsert join, fix #33148

### LINKED ISSUE #33148: Unable to add a new feature when joining two PostGIS tables
**Describe the bug**

I have two Postgresql tables: parent and child, as the following:

```
CREATE SEQUENCE public.parent_id_seq;

CREATE TABLE public.parent
(
    id integer NOT NULL DEFAULT nextval('public.parent_id_seq'::regclass),
    type character varying COLLATE pg_catalog."default",
    CONSTRAINT parent_pkey PRIMARY KEY (id)
);

CREATE SEQUENCE public.child_id_seq;

CREATE TABLE public.child
(
    id integer NOT NULL DEFAULT nextval('public.child_id_seq'::regclass),
    name character varying COLLATE pg_catalog."default",
    parent_id integer,
    CONSTRAINT child_pkey PRIMARY KEY (id),
    CONSTRAINT child_parent_id_fkey FOREIGN KEY (parent_id)
        REFERENCES public.parent (id) MATCH SIMPLE
        ON UPDATE NO ACTION
        ON DELETE NO ACTION
);
```

In QGIS I have made a join between these two tables. But I am not able to add new features from the attribute table. When I do so I got the following error:

Could not commit changes to layer child

```
Errors: ERROR: 1 feature(s) not added.
  
  Provider errors:
      PostGIS error while adding features: ERROR:  invalid input syntax for integer: "nextval('public.parent_id_seq'::regclass)"
    LINE 1: ...ild"("id","name","parent_id") VALUES ($1,'name 9','nextval('...

```

**How to Reproduce**
1- Create these two tables, as described above.
2- Add them to QGIS.
3- Right click on the parent table, then select properties, then from the Joins tab create a new Join as shown in the attached screen shot.
4- Then allow editing both layers.
5- After that open the attribute table of the parent layer, then select Add Feature, and fill in the type and name fields. 
6- Save the parent layer, it will be saved with no problems. Then the child layer, it will generate the error mentioned above.

**QGIS and OS versions**

QGIS version
3.10.0-A Coruña
QGIS code revision
6ffa89eb3e
Compiled against Qt
5.11.2
Running against Qt
5.11.2
Compiled against GDAL/OGR
3.0.2
Running against GDAL/OGR
3.0.2
Compiled against GEOS
3.8.0-CAPI-1.13.1
Running against GEOS
3.8.0-CAPI-1.13.1 
Compiled against SQLite
3.29.0
Running against SQLite
3.29.0
PostgreSQL Client Version
11.5
SpatiaLite Version
4.3.0
QWT Version
6.1.3
QScintilla2 Version
2.10.8
Compiled against PROJ
6.2.1
Running against PROJ
Rel. 6.2.1, November 1st, 2019
OS Version
Windows 10 (10.0)
Active python plugins
quick_map_services; 
db_manager; 
MetaSearch; 
processing

Running on Windows 10 machine.

scr1

SebastienPeillet:
## Description

fix #33148, @devfaz74 provides data test in it.

As explained in the issue, serial type column causes broken link between parent and child table using the upsert on edit feature.

The default values `nextval('public.parent_id_seq'::regclass)` is used to create the child entity but there is no update when parent entity is commited (and `nextval('public.parent_id_seq'::regclass)` is converted to int value during this step so the link is broken).

The (not obvious for users) solution is to activate the `evaluateDefaultValues` option in the project properties menu.

Maybe we should activate this option by default when a join is added with the `upsertOnEdit` flag. That what I propose in this PR, but I am not sure it's the best solution because : 

- It doesn't keep information about previous state of `evaluateDefaultValues`, if the user remove the join in the future.
- I don't know if there is specific contraindications to the use of `evaluateDefaultValues` option...

I let the subject opened to other solution. If this one is accepted I will create the test in consequence.

## Checklist : 
 - [x] Commit messages are descriptive and explain the rationale for changes.

 - [x] Commits which fix bugs include `Fixes #11111` at the bottom of the commit message. If this is your first pull request and you forgot to do this, write the same statement into this text field with the pull request description.

 - [ ] New unit tests have been added for relevant changes

 - [x] You have run the `scripts/prepare-commit.sh` script (https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTING.md#contributing-to-qgis) before each commit.
    If you didn't do this, you can also run `./scripts/astyle-all.sh` from your source folder.

 - [x] You have read the QGIS Coding Standards (https://docs.qgis.org/testing/en/docs/developers_guide/codingstandards.html) and this PR complies with them

### DIFF
--- a/src/core/qgsvectorlayerjoinbuffer.cpp
+++ b/src/core/qgsvectorlayerjoinbuffer.cpp
@@ -82,6 +82,11 @@ bool QgsVectorLayerJoinBuffer::addJoin( const QgsVectorLayerJoinInfo &joinInfo )
     cacheJoinLayer( mVectorJoins.last() );
   }
 
+  if ( joinInfo.hasUpsertOnEdit() )
+  {
+    QgsProject::instance()->setEvaluateDefaultValues( true );
+  }
+
   // Wait for notifications about changed fields in joined layer to propagate them.
   // During project load the joined layers possibly do not exist yet so the connection will not be created,
   // but then QgsProject makes sure to call createJoinCaches() which will do the connection.


SebastienPeillet:
Does someone has an opinion on this ?

PR #34217, opened 2020-02-03
Close vertex for line when editing. fixes #32359 [needs-docs]

### LINKED ISSUE #32359: Impossible to draw a closed line as snap to first vertex is not working
Is impossible to draw a closed line (in one go) as even with snapping on, is not possible to snap to the first vertex. Line must be finished as not closed, then with the node tool the last vertex must be manually snapped to the first one.

lbartoletti:
## Description

This PR can be considered as a bugfix (by a workaround) or as a feature.

It is not possible to simply fix this problem, @troopa81 will be able to give the details.

However, as a draftsman, not being able to close a line is one of the most annoying things in QGIS. I suggest to close the line by pressing the "C" (close) key as proposed in many drawing tools to fix this problem.

I let you comment if for you it should be merged in master and backported to 3.10 if it's bugfix or not.

Thanks.

### DIFF
--- a/src/gui/qgsmaptooldigitizefeature.cpp
+++ b/src/gui/qgsmaptooldigitizefeature.cpp
@@ -92,6 +92,40 @@ void QgsMapToolDigitizeFeature::setCheckGeometryType( bool checkGeometryType )
   mCheckGeometryType = checkGeometryType;
 }
 
+void QgsMapToolDigitizeFeature::keyPressEvent( QKeyEvent *e )
+{
+  QgsVectorLayer *vlayer = qobject_cast<QgsVectorLayer *>( mLayer );
+  if ( !vlayer )
+    //if no given layer take the current from canvas
+    vlayer = currentVectorLayer();
+
+  if ( !vlayer )
+  {
+    notifyNotVectorLayer();
+    return;
+  }
+
+  QgsVectorDataProvider *provider = vlayer->dataProvider();
+
+  if ( !( provider->capabilities() & QgsVectorDataProvider::AddFeatures ) )
+  {
+    emit messageEmitted( tr( "The data provider for this layer does not support the addition of features." ), Qgis::Warning );
+    return;
+  }
+
+  if ( !vlayer->isEditable() )
+  {
+    notifyNotEditableLayer();
+    return;
+  }
+
+  if ( e && e->key() == Qt::Key_C && mode() == CaptureLine && vlayer->geometryType() == QgsWkbTypes::LineGeometry && mCheckGeometryType )
+  {
+    closePolygon();
+    QgsMapMouseEvent e2( mCanvas, QEvent::MouseButtonRelease, QPoint( ), Qt::RightButton );
+    cadCanvasReleaseEvent( &e2 );
+  }
+}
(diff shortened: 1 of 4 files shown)


troopa81:
> It is not possible to simply fix this problem, @troopa81 will be able to give the details.

The thing is that the snapping cache is updated only when a feature is added/removed. 

After a second thought, I'm not sure this is too complicated. Maybe, we could just connect `QgsVectorLayerEditBuffer::geometryChanged` to `QgsPointLocator::onGeometryChanged` and it will work. It's just an idea, I cannot confirm this will surely work.

src/gui/qgsmaptooldigitizefeature.cpp
```diff
@@ -92,6 +92,40 @@ void QgsMapToolDigitizeFeature::setCheckGeometryType( bool checkGeometryType )
   mCheckGeometryType = checkGeometryType;
 }
 
+void QgsMapToolDigitizeFeature::keyPressEvent( QKeyEvent *e )
+{
+  QgsVectorLayer *vlayer = qobject_cast<QgsVectorLayer *>( mLayer );
```
nyalldawson: This whole block should be inside the `e->key() == Qt::Key_C` condition, and if the key isn't 'c' then you need to delegate to QgsMapToolCapture::keyPressEvent (or pressing escape/backspace will be broken)

src/gui/qgsmaptooldigitizefeature.cpp
```diff
@@ -92,6 +92,40 @@ void QgsMapToolDigitizeFeature::setCheckGeometryType( bool checkGeometryType )
...
+    return;
+  }
+
+  if ( !vlayer->isEditable() )
+  {
+    notifyNotEditableLayer();
+    return;
+  }
+
+  if ( e && e->key() == Qt::Key_C && mode() == CaptureLine && vlayer->geometryType() == QgsWkbTypes::LineGeometry && mCheckGeometryType )
```
nyalldawson: shouldn't this include support for CapturePolygon mode? I realise that a right click will auto-close a polygon, but for consistency across the tools I think 'c' should also close and end polygon digitization.

nyalldawson (commented):
Nice improvement. I'm +1 to classing this as a bug fix and merging for 3.12

src/gui/qgsmaptooldigitizefeature.cpp
lbartoletti: Right. Done.

src/gui/qgsmaptooldigitizefeature.cpp
lbartoletti: I hesitated, but indeed, it is consistent like that.

nyalldawson (approved):

uclaros:
Hardcoding such a feature to a key is not a good idea. Isn't the key C already hardcoded to *construction mode* in advanced digitizing? This will break advanced digitizing! A better solution would be to close the line with a modifier to the right click pressed to end digitizing like Ctrl+right click, or enabling snapping to the digitized line as @troopa81 suggested, or both.

PR #34221, opened 2020-02-03
Fix "Random points along line" processing algorithm

### LINKED ISSUE #33156: "Random points along line" algorithm execution fails every two calls
I ran the alg multiple times with the same parameters and concluded that it works every two calls and fails as well every two calls. The error log:

> QGIS version: 3.11.0-Master
QGIS code revision: c52a6e711a
Qt version: 5.11.2
GDAL version: 3.1.0dev
GEOS version: 3.8.0-CAPI-1.13.1 
PROJ version: Rel. 7.0.0, March 1st, 2020
Processing algorithm…
Algorithm 'Random points along line' starting…
Input parameters:
{ 'INPUT' : '//serveur//Études/doss/candidatures.gpkg|layername=A_laisons_douces', 'MIN_DISTANCE' : 0, 'OUTPUT' : 'TEMPORARY_OUTPUT', 'POINTS_NUMBER' : 5 }
>
>Traceback (most recent call last):
File "C:/OSGEO4~1/apps/qgis-dev/./python/plugins\processing\algs\qgis\RandomPointsAlongLines.py", line 126, in processAlgorithm
f = next(source.getFeatures(request.setFilterFid(fid).setSubsetOfAttributes([])))
StopIteration
>
>Execution failed after 0.16 seconds
>
>Loading resulting layers
Algorithm 'Random points along line' finished

agiudiceandrea:
## Description

Fixes "Random points along line" (qgis:randompointsalongline) processing algorithm for two issues:

1) see https://github.com/qgis/QGIS/issues/33156#issuecomment-580628786: the algorithm fails to randomly pick a feature from e.g. spatialite, geojson, gpkg or memory layers (but not from e.g. shapefile or gml layers) in some circumstances;
2) see https://github.com/qgis/QGIS/issues/33156#issuecomment-581395400: the algorithm fails to properly choose a random point in the first part of a line segment (i.e. at a distance from the start vertex of the segment that is less then the minDistance).

This PR needs to be backported to 3.10 branch.

Fixes #33156

### DIFF
--- a/python/plugins/processing/algs/qgis/RandomPointsAlongLines.py
+++ b/python/plugins/processing/algs/qgis/RandomPointsAlongLines.py
@@ -146,25 +146,24 @@ class RandomPointsAlongLines(QgisAlgorithm):
             length = da.measureLine(startPoint, endPoint)
             dist = length * random.random()
 
-            if dist > minDistance:
-                d = dist / (length - dist)
-                rx = (startPoint.x() + d * endPoint.x()) / (1 + d)
-                ry = (startPoint.y() + d * endPoint.y()) / (1 + d)
+            d = dist / (length - dist)
+            rx = (startPoint.x() + d * endPoint.x()) / (1 + d)
+            ry = (startPoint.y() + d * endPoint.y()) / (1 + d)
 
-                # generate random point
-                p = QgsPointXY(rx, ry)
-                geom = QgsGeometry.fromPointXY(p)
-                if vector.checkMinDistance(p, index, minDistance, points):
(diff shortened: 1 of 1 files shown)


nyalldawson (approved):

nyalldawson:
You'll need to manually backport this one.

One other issue with this alg I just noticed -- I think the "minimum distance" parameter here is somewhat misleading, as it could easily be interpreted as the "minimum distance along the line" between points (instead of "minimum distance not-following-the-line".). I think that's a documentation fix though, explicitly stating what this parameter actually means...

PR #34223, opened 2020-02-03
[Bugfix] Read SLD TextSymbolizer for lines

rldhont:
## Description

When reading SLD TextSymbolizer the labeling is well activated but the rendering is not well configured. It seems that placement has to be `QgsPalLayerSettings.Line`.

- [ ] New unit tests have been added for relevant changes
- [x] You have run the `scripts/prepare-commit.sh` script (https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTING.md#contributing-to-qgis) before each commit.
- [x] You have read the QGIS Coding Standards (https://docs.qgis.org/testing/en/docs/developers_guide/codingstandards.html) and this PR complies with them

### DIFF
--- a/tests/src/python/test_qgssymbollayer_readsld.py
+++ b/tests/src/python/test_qgssymbollayer_readsld.py
@@ -444,7 +444,12 @@ class TestQgsSymbolLayerReadSld(unittest.TestCase):
         self.assertEqual(format.size(), 18)
         self.assertEqual(format.sizeUnit(), QgsUnitTypes.RenderPixels)
 
+        # the layer contains lines
+        # from qgis.core import QgsWkbTypes
+        # self.assertEqual(layer.geometryType(), QgsWkbTypes.LineGeometry)
+        # the placement should be QgsPalLayerSettings.Line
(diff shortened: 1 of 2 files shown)


tests/src/python/test_qgssymbollayer_readsld.py
```diff
@@ -444,7 +444,12 @@ def testLineOpacity():
         self.assertEqual(format.size(), 18)
         self.assertEqual(format.sizeUnit(), QgsUnitTypes.RenderPixels)
 
+        # the layer contains lines
+        # from qgis.core import QgsWkbTypes
+        # self.assertEqual(layer.geometryType(), QgsWkbTypes.LineGeometry)
```
nyalldawson: Why is this commented out?

nyalldawson (commented):
What's the best match for the SLD spec? Line mode == straight labels, does the spec dictate this or are they rendered as curved labels?

tests/src/python/test_qgssymbollayer_readsld.py
rldhont: Because I'd like to verify that the layer contains line and the placement is line, but the placement gets from the layer's label settings is not line.

PR #34225, opened 2020-02-03
Make minimm number of nodes to 2 in subdivide algorithm.

3nids:
was there any reason to make it 8?
use-case: I want to subdivide lines in segments.


nyalldawson:
I think you want to use the "explode lines" algorithm, not subdivide

nyalldawson:
I'm reticent to make this change -- as described above, the "explode lines" algorithm addresses your use case. If we make the minimum < 4 then these results would be totally broken for polygon inputs, so we'd at least need to detect values < 4 and raise an exception for polygon inputs. (but my suspicion is that trying to subdivide polygons to a maximum of 4 nodes will also result in errors, hence the min 8 value set by the original postgis implementation this was based on....)

3nids:
@nyalldawson indeed explode was the solution.
But can you explain the rationale for these 8 nodes at least?

nyalldawson:
The function was ported from postgis, which has that restriction. I'm unsure if it's an artificial restriction, or required because of the internal logic.

3nids:
ok, thanks for the explanation.

PR #34227, opened 2020-02-03
Fix layer dependencies double load of layers

### LINKED ISSUE #34226: Layer dependencies is leading to loading a layer twice
With a n:m relation project (pipe <- join table -> event), the pipe layer gets duplicated when opening it. 
It seems it's created by the layer dependency detection.

test project: 
[Archive.zip](https://github.com/qgis/QGIS/files/4148538/Archive.zip)

elpaso:
Fixes #34226


3nids:
successfully tested with my small project

3nids (approved):

PR #34247, opened 2020-02-04
Follow up 012ee83: delete .qpj file when overwriting pre-existing shapefile dataset

nirvn:
## Description

@nyalldawson , as you requested, here's the follow up PR which also disables the .qpj workaround in the QgsOgrProviderUtils::createEmptyDataSource function.

It also occurred to me we need to delete pre-existing .qpj when overwriting pre-existing datasets, otherwise worlds might collide :wink:


nyalldawson (approved):

nyalldawson:
Thanks!

PR #34249, opened 2020-02-04
 Don't force users to make a choice in the datum transform dialog

### LINKED ISSUE #34234: Select Transformation of "LayerName" - locked window
Using 3.10.2 I get a pop-up window when QGIS first opens requesting the selection of a NAD83-NAD27 transformation for a file stored within the project.  I can select various transformations but the OK button never becomes active.  (See image below) The only way to get out of the window is to kill QGIS.  I have installed the North American datumgrid package in "C:\Program Files\QGIS 3.10\share\proj\*" and I have verified the requested grid files are present.   I reverted back to 3.10.1 and I do not get the locked window.  

image

Running Windows 10 Pro for Workstations

nyalldawson:
It's possible to get a situation where none of the options can be
selected, because they all require a grid which isn't available on
the system. So always allow users to close the dialog to dismiss it
(which has the same effect as picking the default selection from
the dialog when possible, or giving null transforms if NO transforms
are available at all...)

Fixes #34234

Also match the UI of the coordinate operation widget with other "missing grid" widgets in QGIS, and offer a user-friendly way to install the required grid.

### DIFF
--- a/src/app/CMakeLists.txt
+++ b/src/app/CMakeLists.txt
@@ -53,7 +53,6 @@ SET(QGIS_APP_SRCS
   qgssourcefieldsproperties.cpp
   qgsattributesformproperties.cpp
   qgsidentifyresultsdialog.cpp
-  qgsinstallgridshiftdialog.cpp
   qgsfeatureaction.cpp
   qgslayercapabilitiesmodel.cpp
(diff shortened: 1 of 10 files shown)


nirvn:
Good move. That'll also help intimidated users who might prefer to bail out too.

Issue #33458, opened 2019-12-18
Error if user CRS contains non-ASCII characters in name and is used as project crs.


dr1542:
There is not displayed USER:CRS_ID in qgis status bar, but crs name displayed in tooltip. 
![CRS-screen](https://user-images.githubusercontent.com/58694010/71074677-c22f2480-2193-11ea-8a11-5c3e5f1fac04.png)

It is not only UI error - unable to switch to another user CRS with non-ASCII characters in name. If try to switch to EPSG crs or user crs with only ASCII characters in name and then to second CRS with non-ASCII characters coordinate panel in status bar continues display coordinates of first user CRS with non-ASCII name.

I checked this in Qgis 3.11.0-Master (0d84e35839) and 3.10.0-A CoruÃ±a (6c816b4204)
OS - Windows 7

Qgis 2.18.28 works fine

gioman:
> Qgis 2.18.28 works fine

tagging as regression.

Dkrav-UA:
In 3.10.1 version everything was fine. It was broken in 3.10.2 version.

gioman:
> was broken

"was" or "is"?

Dkrav-UA:
> "was" or "is"?

3.10.1 - ok
3.10.2 - broken

troopa81:
I don't reproduce with current 3.10.x
![Uploading erreur_crs.png…]()

you can install it with OSGEO4W installer [qgis-rel-dev](https://www.qgis.org/en/site/forusers/alldownloads.html#windows).

If OK, it would be delivered in 3.10.3

nyalldawson:
I can't reproduce. Can you post more detailed steps to reproduce please?

dr1542:
I can't reproduce this too. But only after I was created new user profile with empty qgis.db crs database. In new session i was created some crs definitions and qgis works as expected. When I opened this issue, I used my old qgis.db, because I use a lot of custom crs.
At first glance, I don't find any difference in the table formats in the old and new qgis.db but it seems to be there.

Dkrav-UA:
I confirm that with a new user profile everything is ok. But with the old profile problem is more than just show USER:XXXXX in the corner. For example, I can't delete or edit existing CRS's. I get an error.
![Screenshot_20200204_094515](https://user-images.githubusercontent.com/16414895/73724070-36695b80-4733-11ea-962a-85a145b74c32.png)
But validation check returns OK.

nirvn:
[...]

PR #34250, opened 2020-02-04
Show "Unknown CRS" entry in crs selector widget and status bar widget for non-standard, non-user CRSes

nyalldawson:
Fixes some UX issues when a layer or project is set to a unknown, non-standard CRS.

Refs #33458


nyalldawson:
@nirvn would love your testing/feedback here

nirvn:
Tested, all good.

PR #34256, opened 2020-02-04
Raise log level of server exception

pblottiere:
## Description

Fixes issue raised [here](https://lists.osgeo.org/pipermail/qgis-developer/2020-February/060108.html).

### DIFF
--- a/src/server/qgsserver.cpp
+++ b/src/server/qgsserver.cpp
@@ -420,7 +420,7 @@ void QgsServer::handleRequest( QgsServerRequest &request, QgsServerResponse &res
     {
       responseDecorator.write( ex );
       QString format;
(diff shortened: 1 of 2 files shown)


pblottiere:
Hi @dmarteau,

Do you have other messages raised with an inappropriate log level in mind?

dmarteau:
> Do you have other messages raised with an inappropriate log level in mind?

Hi @pblottiere,

Yes, message when loading services, they have the 'warning' level, IMHO it should be 'info': https://github.com/qgis/QGIS/blob/master/src/server/qgsserviceregistry.cpp#L131 . Each time is see these I tell myself that I should do something :-).

I think it is happening because the QgsMessageLog::logMessage has 'warning' as default level for output and many calls to logMessage do not specify the log level.

pblottiere:
Thanks for your input David :+1:

PR #34258, opened 2020-02-04
Fix user projection not identified as such

nirvn:
## Description

@nyalldawson , while looking into issue #33458 , I stumbled on a bug whereas user projections are not identified as such. 

Long story short here, the qgis.db provided @Dkrav-UA revealed that the auth_name and auth_id columns for the saved user projections weren't null (just empty strings). In turn, that led the authid column in that statement (https://github.com/qgis/QGIS/blob/master/src/core/qgscoordinatereferencesystem.cpp#L634) to become ':' instead of null.

As a result, the code would return a valid CRS with its authid() function returning ':', which created problems with the custom projection dialog (i.e. since the authid wasn't 'USER:XXXX', the code would not behave properly there).

The PR fixes the situation by adding a ':' check and handle that the same way we handle an empty string when loading a CRS from the user database.

### DIFF
--- a/src/core/qgscoordinatereferencesystem.cpp
+++ b/src/core/qgscoordinatereferencesystem.cpp
@@ -653,7 +653,7 @@ bool QgsCoordinateReferenceSystem::loadFromDatabase( const QString &db, const QS
     wkt = statement.columnAsText( 8 );
     d->mAxisInvertedDirty = true;
 
-    if ( d->mSrsId >= USER_CRS_START_ID && d->mAuthId.isEmpty() )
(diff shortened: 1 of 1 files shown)


nirvn:
Once upon a time, auth_name and auth_id were likely not null, see this comment (https://github.com/qgis/QGIS/blob/master/src/core/qgscoordinatereferencesystem.cpp#L628). So this would impact folks that have carried a user profile for a long, long time.

nyalldawson (approved):

nyalldawson:
Thanks for working this one through!

PR #34268, opened 2020-02-04
Do not replace plus by space in REQUEST_BODY

arnaud-morvan:
## Description

Actually + signs in request body are replaced by spaces when passing through QgsServerParameters (see: https://github.com/qgis/QGIS/commit/bbf7a78453cca095396f9ee2ca3d037e40f7d208).

As I would prefer not to change the behavior of QgsServerParameters, I just encode plus sign to "%2B" as we would do for an url query string parameter.

This might be backported to all active release since 3.4 (as linked commit was commited in 3.4).

Example affected REQUEST BODY:
```
<GetFeature>
    <Query typeName="feature:mytable" srsName="EPSG:3857">
        <Filter>
            <PropertyIsEqualTo matchCase="true">
                <PropertyName>etage</PropertyName>
                <Literal>+1</Literal>
            </PropertyIsEqualTo>
...
```

### DIFF
--- a/src/server/qgsrequesthandler.cpp
+++ b/src/server/qgsrequesthandler.cpp
@@ -252,7 +252,7 @@ void QgsRequestHandler::parseInput()
 
           mRequest.setParameter( attrName.toUpper(), attr.value() );
         }
-        mRequest.setParameter( QStringLiteral( "REQUEST_BODY" ), inputString );
+        mRequest.setParameter( QStringLiteral( "REQUEST_BODY" ), inputString.replace( '+', QStringLiteral( "%2B" ) ) );
       }
     }
   }
--- a/tests/src/python/test_qgsserver_request.py
+++ b/tests/src/python/test_qgsserver_request.py
@@ -194,6 +194,21 @@ class QgsServerRequestTest(QgsServerTestBase):
         _check_links(params)
         _check_links(params, 'POST')
 
+    def test_fcgiRequestBody(self):
+        """Test request body"""
+        data = '<Literal>+1</Literal>'
+        self._set_env({
+            'SERVER_NAME': 'www.myserver.com',
+            'SERVICE': 'WFS',
+            'REQUEST_BODY': data,
+            'CONTENT_LENGTH': str(len(data)),
+            'REQUEST_METHOD': 'POST',
+        })
+        request = QgsFcgiServerRequest()
+        response = QgsBufferServerResponse()
+        self.server.handleRequest(request, response)
+        self.assertEqual(request.parameter('REQUEST_BODY'), '<Literal>+1</Literal>')
+
     def test_add_parameters(self):
         request = QgsServerRequest()
         request.setParameter('FOOBAR', 'foobar')


elpaso:
@arnaud-morvan  As far as I can see the commit  bbf7a78 affects the query string, not the request body.

Are you sure that the '+' sign is converted to a space also on POST body? If that's the case it should be fixed there and not like you are doing here.

But let's hear what @pblottiere thinks.

pblottiere:
Hi @arnaud-morvan,

Can you add a unit test to highlight the underlying behavior please?

arnaud-morvan:
@pblottiere : I've added a test that highlight the problem.
Here is the output of the test (before the proposed fix): 
```
574: ======================================================================
574: FAIL: test_fcgiRequestBody (__main__.QgsServerRequestTest)
574: Test request body
574: ----------------------------------------------------------------------
574: Traceback (most recent call last):
574:   File "/home/amorvan/dev/QGIS/tests/src/python/test_qgsserver_request.py", line 210, in test_fcgiRequestBody
574:     self.assertEqual(request.parameter('REQUEST_BODY'), '<Literal>+1</Literal>')
574: AssertionError: '<Literal> 1</Literal>' != '<Literal>+1</Literal>'
574: - <Literal> 1</Literal>
574: ?          ^
574: + <Literal>+1</Literal>
574: ?          ^
574: 
```

elpaso (approved):
This is a patch and it's acceptable, but I think that the whole QgsRequestHandler::parseInput needs refactoring: what happens if a future service/API wants to upload a binary (for example an image)?

Do we really want to arbitrarily replace '+' charachters in a binary? 

I think that the right approach here would be to store the POST/PATCH/PUT request body as a QByteArray and delegate the parsing to the handlers. 

What are we doing now is limited to XML and JSON, that is far too simple for the real web world.

pblottiere:
>  I've added a test that highlight the problem.

@arnaud-morvan Nice, thanks :+1:

> Actually + signs in request body are replaced by spaces when passing through QgsServerParameters (see: bbf7a78).

> This might be backported to all active release since 3.4 (as linked commit was commited in 3.4).

@arnaud-morvan Actually the linked commit is just a refactoring of a mechanism introduced sooner (see https://github.com/qgis/QGIS/pull/6879/commits/4087521f246b20469d52e601465164f2fb0021c3), meaning that it's probably broken since 3.0.

> This is a patch and it's acceptable

@elpaso Indeed, but I think it would be better to move this logic in `QgsServerParameters` itself. We shouldn't have to hack the query to retrieve valid parameters. What do you think about adding a private `QgsServerParameters::prepare` method which would be in charge of replacing the `+` character in the `REQUEST_BODY` content? This new method could be called before loading the parameter in the `QgsServerParameters::load` method.

> I think that the right approach here would be to store the POST/PATCH/PUT request body as a QByteArray and delegate the parsing to the handlers.

+1

elpaso:
> > I've added a test that highlight the problem.
> 
> @arnaud-morvan Nice, thanks +1
> 
> > Actually + signs in request body are replaced by spaces when passing through QgsServerParameters (see: [bbf7a78](https://github.com/qgis/QGIS/commit/bbf7a78453cca095396f9ee2ca3d037e40f7d208)).
> 
> > This might be backported to all active release since 3.4 (as linked commit was commited in 3.4).
> 
> @arnaud-morvan Actually the linked commit is just a refactoring of a mechanism introduced sooner (see [4087521](https://github.com/qgis/QGIS/commit/4087521f246b20469d52e601465164f2fb0021c3)), meaning that it's probably broken since 3.0.
> 
> > This is a patch and it's acceptable
> 
> @elpaso Indeed, but I think it would be better to move this logic in `QgsServerParameters` itself. We shouldn't have to hack the input query to retrieve valid parameters. What do you think about adding a private `QgsServerParameters::prepare` method which would be in charge of replacing the `+` character in the `REQUEST_BODY` content? This new method could be called before loading the parameter in the `QgsServerParameters::load` method.

I'm not sure, what is wrong here is that we are treating the request body as if it was an URL query string parameter (see: `mRequest.setParameter( QStringLiteral( "REQUEST_BODY" ), inputString );` and `mUrl.setQuery( mParams.urlQuery() );`, this is the root of the issue.

While it is perfectly safe and sound to replace `+` with ` ` in an URL component (this is what the web has always done and it's buried in some RFC somewhere), the same it is NOT safe and should NOT be done for a POST/PUT/PATCH request body, that should be untouched.

So, what is wrong here starts with  `mRequest.setParameter( QStringLiteral( "REQUEST_BODY" ), inputString );`, IMHO the right thing to do is to create a `mRequest.setBody(QByteArray)` and do not pass the REQUEST_BODY around as a query string component.


> 
> > I think that the right approach here would be to store the POST/PATCH/PUT request body as a QByteArray and delegate the parsing to the handlers.
> 
> +1

elpaso:
Just to be clear, this PR is going to break something in the future but for now it's acceptable, unless @arnaud-morvan wants to **Do The Right Thing (TM)** and fix it for good, of course :1st_place_medal:

pblottiere:
> So, what is wrong here starts with mRequest.setParameter( QStringLiteral( "REQUEST_BODY" ), inputString );, IMHO the right thing to do is to create a mRequest.setBody(QByteArray) and do not pass the REQUEST_BODY around as a query string component.

Totally agree, it's the good long-term solution :+1:. I just wanted to have a less hacky temporary fix in the meantime :). However `QgsServerParameters` class is not aware of the type of the request (GET, POST, ...) so my proposition is not viable. 

> Just to be clear, this PR is going to break something in the future but for now it's acceptable

OK for me. I don't have more idea for a more proper temporary fix...

> unless @arnaud-morvan wants to Do The Right Thing (TM) and fix it for good, of course :1st_place_medal:

No pressure @arnaud-morvan ;)

arnaud-morvan:
Thanks @elpaso and @pblottiere for the review.

I agree that this is just a quickfix, but pretty sure it cannot break anything as in the corner use case where POST request body is supposed to be interpreted as a QUERY STRING, such body is parsed and parameters are set with query string keys.
Every service handler that use request parameter('REQUEST_BODY") should expect the original body.

Could one of you merge this in master and 3.10 (I do not have merge button) for the moment as this is blocking me to go on production with a project.

And I will take some time, maybe at the hackfest to add and use request.body() method in all request handlers.

Regards

pblottiere:
> Could one of you merge this in master and 3.10 (I do not have merge button) for the moment as this is blocking me to go on production with a project.

+1 for backporting.

> And I will take some time, maybe at the hackfest to add and use request.body() method in all request handlers.

Nice, thanks Arnaud :+1:. I'll merge this quickfix in the meantime then.

pblottiere:
Hi @arnaud-morvan,

qgis-bot failed to backport. Can you take a look?

arnaud-morvan:
Hi @pblottiere,
I just created the backport pull requests.

PR #34271, opened 2020-02-04
Add highlight features when "follow map theme" is selected in map layout

### LINKED ISSUE #34178: QGIS Server GetPrint: HIGHLIGHT_GEOM is not printed if map layers are configured to follow a map theme
**Describe the bug**

follow map theme

**How to Reproduce**

**QGIS and OS versions**

QGIS Master

jgrocha:
## Description

Users are not able to print some highlight feature(s) (like a property) on top of a map layout, if the map layout item is locked to a theme (a set of layers). 

Fix #34178


src/core/layout/qgslayoutitemmap.cpp
```diff
@@ -1893,23 +1893,27 @@ QList<QgsMapLayer *> QgsLayoutItemMap::layersToRender( const QgsExpressionContex
...
+
+  if ( !layers().isEmpty() )
+    renderLayers = layers();
+  else
+    renderLayers = mLayout->project()->mapThemeCollection()->masterVisibleLayers();
+
+  // preserve the list of highlight layers to add if presetName is set
+  for ( const auto mlayer : renderLayers )
+  {
+    if ( mlayer->name().contains( "highlight_" ) )
```
nyalldawson: This is quite hacky. I think it should be handled in server instead, by modifying the theme definition to include all desired map layers including the highlight ones.

nyalldawson (dismissed):
as above

src/core/layout/qgslayoutitemmap.cpp
jgrocha: Fully agree! But I was not able to find a more elegant solution (yet).

The same theme can be used in more than one map in the layout; and the highlight can be only on one of those maps. I think we can not change the theme.

I'm creating a more complicated combination to test this (a layout with several maps and several highlights, and some with layers following a theme).

nyalldawson: What about copying the theme, modifying it as desired, and then setting the map to follow the new theme?

jgrocha: @nyalldawson I need some help, sorry! 
The highlight layers are already created on the server logic, as the other layers passed as parameters, before we know if the map item follows a project theme.
Here (at `QgsLayoutItemMap::layersToRender`) we were discarding all server layers (including the highlight ones) and use only the ones from the theme.
I've just added a few lines to preserve the highlighted layers, before dropping them all, if the map item follows a theme. It's a kind of a hack (not a proper refactor), I agree, but it does seem dangerous.
In summary, I need your help. I'm not seeing the overall picture as you do.

nyalldawson: What I'm suggesting is something like (and I don't know the internals of the server code to give specifics, sorry!)

- create the highlight layers
- before exporting the layout, you see which maps need to include the highlight layers
- for these, if any have a theme set, you copy the theme, add the highlight layers to it, and then set the map item to follow that new duplicate theme
- drop the temporary theme

pblottiere: I agree with Nyall's comments, it should clearly be managed on server side. And the idea of creating a new temporary theme seems the way to go indeed.

pblottiere:
@jgrocha can you add a unit test please? It will be necessary to merge you PR and I'll be able to help you more easily :).

Moreover, I think we have the same problem with external layers...

jgrocha:
> @jgrocha can you add a unit test please? It will be necessary to merge you PR and I'll be able to help you more easily in the meantime :).
> 
> Moreover, I think we have the same problem with external layers...

@pblottiere Thanks for your help. I reverted my hack and created a unit test. **It is expected that the CI will fail with this additional test.**

I've created another layout on `tests/testdata/qgis_server_accesscontrol/project.qgs` with a map item that follows a project theme (the theme is called World).

jgrocha:
@nyalldawson I was able to do everything all on the server side.

I create a new theme with all the layers of the original theme plus the highlight layers. 

I set the new theme with:
```
map->setFollowVisibilityPresetName( presetName + "_higlight" );
```
and the new theme is set, but I don't get the highlight layers.

The highlight layers are discard on [`QgsLayoutItemMap::layersToRender`](https://github.com/qgis/QGIS/blob/6bffe18af3d3931395ecc0d69718c9700febd69d/src/core/layout/qgslayoutitemmap.cpp#L1889). This calls
[`QgsMapThemeCollection::mapThemeVisibleLayers`](https://github.com/qgis/QGIS/blob/6bffe18af3d3931395ecc0d69718c9700febd69d/src/core/qgsmapthemecollection.cpp#L314) which iterate over the layer order, but this layer is not there. So it is discarded.

Should I add the layer to the layer tree? The layer is created, but it is never added to the layer tree. But if it is not there, the `QgsMapThemeCollection::mapThemeVisibleLayers` will discard it.

jgrocha:
I've added the layers to the layer tree to make this work, but I not sure if this does not have side effects.

jgrocha:
Added `unConfigurePrintLayout` method to revert changes made to themes and to the layer tree.

This PR passes the test I've added. The test just covers the issue #34271 

@nyalldawson All code is on the server side. Thanks for help.

src/server/services/wms/qgswmsrenderer.cpp
```diff
@@ -546,9 +547,34 @@ namespace QgsWms
       handlePrintErrors( layout.get() );
     }
 
+    // remove temeporary themes
+    // remove temeporary highlight layers
```
jgrocha: Ping @3nids Hum... The spell checker is not working on C++ comments? Is is really useful to have the spell checker on the CI.

nyalldawson:
Given that all the changes here reside in server now (thanks!) I've no issue with the changes, but they'll need to be reviewed by @pblottiere or someone else familiar with the server code.

jgrocha:
Ping @pblottiere 
Can you take a look, please?

jgrocha:
Ping @pblottiere @elpaso

jgrocha:
...Pong

jgrocha:
@pblottiere Can you take a look, please?

src/server/services/wms/qgswmsrenderer.cpp
```diff
@@ -546,9 +547,34 @@ namespace QgsWms
...
+    unConfigurePrintLayout( layout.get() );
```
pblottiere, line 552: It seems that the `layout` parameter is a unique_ptr clone of the original layout coming from the project. So it'll be destroyed at the end of the `getPrint` method without adversely impacting the underlying project.

Am I wrong? Because in this case the `unConfigurePrintLayout` would be unnecessary.

pblottiere: Oh, it seems that the project is not itself cloned in the `QgsPrintLayout::clone` method so you're right,  we have to post process  it.

But I still have one concern about the `unConfigurePrintLayout`. Actually, if something goes wrong during the `getPrint` method before reaching the `unConfigurePrintLayout`, the project is left in an uncleaned state (because the new theme and corresponding layers won't have been removed).

What do you think @jgrocha?

jgrocha: @pblottiere If something goes wrong with `getPrint`, QGIS Server ends returning an exception or something else, but it will not use this highlight or layer configuration again. For the next request, QGIS Server will read the layout's layer configuration again from the project and the layers and highlight to print from the request.

In other words: is this `unConfigurePrintLayout` really necessary? If QGIS Server handles each request separately, if one request (like this one) changes the layer configuration (in memory), all other subsequent requests are not affected. 

To make sure, I've tested the same PR without `unConfigurePrintLayout`. All subsequent requests were not affected. But I'm probably missing something. Can you help me to find a situation in which a subsequent request can be affected by this change?

P.S. Thank you so much for taking a look into this.

jgrocha: @pblottiere I'm so sorry, but inadvertently I've pushed my repository after rebasing. I had to recreate the pull request. This same PR is now #36007.

jgrocha:
This PR as been replaced by #36007. The contribution is exactly the same, plus an additional commit to resolve a merge conflict due to more recent updates.

Issue #33929, opened 2020-01-20
Allow "fallback" non-preferred transforms if the preferred coordinate operation fails


SteveLowman:
**Describe the bug**

In QGIS 3.8 and 3.10, with the high-precision transformation file 'OSTN15_NTv2_OSGBtoETRS.gsb' installed into the OSGeo4W64\share\proj folder, the transformation is not working correctly, and the behaviour is very different to the older, less precise transformations available for EPSG:27700

**How to Reproduce**

1. Use QGIS 3.10.2, including proj 7.0.0 and gdal 3.1.0dev, and ensure that OSTN15_NTv2_OSGBtoETRS.gsb is pasted into the OSGeo4W64\share\proj folder
2. Download, unzip, and open the .qgz file with data, as attached
[27700_Transformation_issue_Jan_2020.zip](https://github.com/qgis/QGIS/files/4087123/27700_Transformation_issue_Jan_2020.zip)
3. You should see three layers in the Layers Panel; they have the same identical rectangle, covering part of the UK and its vicinity, but only the EPSG:27700 layer is visible on the map canvas, because OSTN15_NTv2_OSGBtoETRS.gsb is preventing the other two layers from rendering.
![Three_layers](https://user-images.githubusercontent.com/8450382/72746711-b4155f00-3baa-11ea-999c-609d42c6403d.png)

![One_feature](https://user-images.githubusercontent.com/8450382/72746724-bb3c6d00-3baa-11ea-8b27-6c111e0e824a.png)

4. In 'Project Properties > CRS > Datum Transformations', edit the transformations for EPSG:3857 and EPSG:4326, in each case choosing the second available transformation (it has 2 metre precision).
![Edit Datum Transformations](https://user-images.githubusercontent.com/8450382/72746752-d14a2d80-3baa-11ea-9487-66ca5b2d5d9f.png)

![Select Datum Transformations](https://user-images.githubusercontent.com/8450382/72746848-11a9ab80-3bab-11ea-9300-1046d1d29531.png)

5. Refresh the map canvas, and you should see that the two transformed layers (test_3857 & test_4326) will appear as diagonal hatched polygons with identical dimensions over the solid colour of the EPSG:27700 polygon.
![Three_features](https://user-images.githubusercontent.com/8450382/72746871-22f2b800-3bab-11ea-9705-44be38620fe6.png)

[...]

PR #34273, opened 2020-02-05
Implement "fallback" transformation when selected operation fails

nyalldawson:
When the selected coordinate operation fails to project coordinates, try falling back to a default PROJ-provided operation to transform the coordinates

This allows **some** transformation of coordinates to occur in the situation:
- user has selected a coordinate operation utilising a grid shift file
- one or more points which fall OUTSIDE the grid file are transformed

Not for merge -- likely needs some form of user-visible-warning when this occurs

Refs #33929

### DIFF
--- a/src/core/qgscoordinatetransform.cpp
+++ b/src/core/qgscoordinatetransform.cpp
@@ -49,6 +49,11 @@ QReadWriteLock QgsCoordinateTransform::sCacheLock;
 QMultiHash< QPair< QString, QString >, QgsCoordinateTransform > QgsCoordinateTransform::sTransforms; //same auth_id pairs might have different datum transformations
 bool QgsCoordinateTransform::sDisableCache = false;
 
+std::function< void( const QgsCoordinateReferenceSystem &sourceCrs,
+                     const QgsCoordinateReferenceSystem &destinationCrs,
+                     const QgsDatumTransform::TransformDetails &desiredOperation,
+                     const QgsDatumTransform::TransformDetails &usedOperation )> QgsCoordinateTransform::sFallbackOperationOccurredHandler = nullptr;
+
 QgsCoordinateTransform::QgsCoordinateTransform()
 {
   d = new QgsCoordinateTransformPrivate();
@@ -637,12 +642,14 @@ void QgsCoordinateTransform::transformCoords( int numPoints, double *x, double *
     return;
   }
 
+#if PROJ_VERSION_MAJOR>=6
   double *xprev = new double[ numPoints ];
   memcpy( xprev, x, sizeof( double ) * numPoints );
   double *yprev = new double[ numPoints ];
   memcpy( yprev, y, sizeof( double ) * numPoints );
   double *zprev = new double[ numPoints ];
   memcpy( zprev, z, sizeof( double ) * numPoints );
+#endif
 
 #ifdef COORDINATE_TRANSFORM_VERBOSE
   double xorg = *x;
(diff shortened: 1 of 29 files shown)


nyalldawson:
@rouault keen to hear your thoughts on this!

src/core/qgscoordinatetransform.cpp
```diff
@@ -637,6 +637,13 @@ void QgsCoordinateTransform::transformCoords( int numPoints, double *x, double *
     return;
   }
 
+  double *xprev = new double[ numPoints ];
```
rouault: should be restricted #if PROJ_VERSION_MAJOR>=6 to avoid mem leak

src/core/qgscoordinatetransform.cpp
```diff
@@ -637,6 +637,13 @@ void QgsCoordinateTransform::transformCoords( int numPoints, double *x, double *
...
+  memcpy( xprev, x, sizeof( double ) * numPoints );
+  double *yprev = new double[ numPoints ];
+  memcpy( yprev, y, sizeof( double ) * numPoints );
+  double *zprev = new double[ numPoints ];
+  memcpy( zprev, z, sizeof( double ) * numPoints );
```
rouault: this method cannot be called with z==NULL right ?

src/core/qgscoordinatetransform.cpp
```diff
@@ -707,6 +720,45 @@ void QgsCoordinateTransform::transformCoords( int numPoints, double *x, double *
   }
 #endif
 
+#if PROJ_VERSION_MAJOR>=6
+
+  if ( actualRes != 0 )
```
rouault: actualRes will not be reliable for numPoints > 1. you'd probably have to iterate over all transformed points to check their x != HUGE_VAL

src/core/qgscoordinatetransform.cpp
nyalldawson: right -- all code paths leading here pad out the z array with 0's if it doesn't already exist

src/core/qgscoordinatetransform.cpp
nyalldawson: Why is that? Does proj_errno get reset mid-operation when transforming multiple points?

rouault: I'm not sure. Error handling in PROJ is super confusing (to me at least)... See https://github.com/OSGeo/PROJ/issues/1856

nyalldawson:
Getting closer... here's how things look now:

image


image


@rouault I'm curious to hear your thoughts on https://github.com/qgis/QGIS/pull/34273/commits/3f2c6733337164bd4c838906b6cc3602e2c2d18b -- is this a proj bug?

nyalldawson:
Little screencast:

Peek 2020-02-14 15-16

rouault:
> I'm curious to hear your thoughts on [3f2c673](https://github.com/qgis/QGIS/commit/3f2c6733337164bd4c838906b6cc3602e2c2d18b) -- is this a proj bug?

maybe... A reproducer ?

nirvn:
Tested the branch, I can confirm it fixes the linked issues. No regressions spotted so far either.

nirvn:
@nyalldawson , the plan is to introduce this in 3.12 and backport it in 3.10.4 (ie the point release after the one coming in a few days)?

nyalldawson:
@nirvn I'm unsure... Its an important fix, but relatively high risk...

nirvn:
For what it's worth, I would backport. I think the benefits outweight the risks. Plus, it'd be good for this LTR to be as sync'ed as possible on the projection/transformation front, it'll ease backports for the year to come.

nyalldawson:
Should be ready for merge now...

PR #34276, opened 2020-02-05
Add a recenter button to the GPS information panel

nyalldawson:
There's NO way to recenter the map on the GPS location, which is a HUGE ux limitation in QGIS' GPS handling. This commit adds a new "Recenter" button to the panel to allow users to jump right to
the current GPS position.

It's a borderline feature, but I feel that without this ability the whole QGIS/GPS integration is rather crippled...


nirvn (approved):
I'm fine with this being merged. It's half way to a feature but as your pointing out it's also halfway to fixing an unfulfilled expected feature behavior.

The few added lines of actually code also all within src/app.

Approved ;)

timlinux:
I dont have a GPS on my laptop so can't see where this new feature is, but didn't this already do what you needed?

image

nyalldawson:
@timlinux that's correct -- the other setting only really works with a moving GPS, as it's only triggered when the GPS moves. If you're stationary and then move the map the new button allows a means to recenter the map on the last GPS position

PR #34281, opened 2020-02-05
Fix geometry save/restore for snapping configuration dialog

nyalldawson:
and make dialog a Tool window type


nyalldawson:
@3nids how's this for you?

PR #34287, opened 2020-02-05
Fix recent expressions in select by exp dlg

elpaso:
backport of https://github.com/qgis/QGIS/pull/33922


src/gui/qgsexpressionselectiondialog.cpp
```diff
@@ -61,6 +61,7 @@ QgsExpressionSelectionDialog::QgsExpressionSelectionDialog( QgsVectorLayer *laye
   mExpressionBuilder->loadRecent( QStringLiteral( "Selection" ) );
```
nyalldawson: Is this still needed?

nyalldawson (approved):

PR #34305, opened 2020-02-05
Fix row conditional formatting

### LINKED ISSUE #34122: Full row conditional formatting formats wrong full rows
**Describe the bug**
In the attribute table, full row conditional formatting formats wrong rows when the attribute table is sorted and then refreshed. Some rows that should be formatted are not, and some that should not be, they are.

**How to Reproduce**
Open the attribute table from the attached project, sort by some column and hit refresh.
The highlight condition is `"top" % 10 = 0 and "right" % 10 = 0` and the rows that should be highlighted are the ones with `"shouldbegreen" = 1`.

**QGIS and OS versions**
Happens on qgis master, but also on older builds like 3.8, both linux and windows. Haven't tested others. 

**Additional context**
Could be the same bug as #28352, though no edits are required to reproduce it so I opened a new one.

Here is a sample project and dataset. I added lots of rows so it's easier for the bug to appear.
[bug.zip](https://github.com/qgis/QGIS/files/4130026/bug.zip)

elpaso:
Fixes #34122 

The original implementation was storing the styles
in an hash keyed by the row number from the model
index, but that one changes when the table is reloaded.

By using feature id as a key instead, the style association
with the feature is maintained through sort/reload
operations.

### DIFF
--- a/src/gui/attributetable/qgsattributetablemodel.cpp
+++ b/src/gui/attributetable/qgsattributetablemodel.cpp
@@ -711,14 +711,14 @@ QVariant QgsAttributeTableModel::data( const QModelIndex &index, int role ) cons
     {
       mExpressionContext.setFeature( mFeat );
       QList<QgsConditionalStyle> styles;
-      if ( mRowStylesMap.contains( index.row() ) )
+      if ( mRowStylesMap.contains( mFeat.id() ) )
       {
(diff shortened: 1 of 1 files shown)


NathanW2:
Thanks heaps for the fix!

nyalldawson (approved):

PR #34311, opened 2020-02-05
Turn relative path into absolute path for reading embedded projects.

### LINKED ISSUE #34060: QGIS does not load embedded layers when opening projects
**Describe the bug**

I am having an issue with vs 3.10.n with handling embeded layers particularly on reopening - seems to save embed layers added in a session but when reopening a second time the link is dropped.  

**How to Reproduce**
Follow the following procedure:
Open a new or existing project;
Add embed layers and save project;
Close the project;
Reopen the project - the embed layers will show;
Save the project again and close;
Reopen the project again - will list the embeded layers in the Layers panel but does not link to the data and therefore does not include them on the map.

**QGIS and OS versions**
Using Windows10 and have found the issue on QGis vs3.10.0, .1 and still on .2

**Additional context**
Also does not save the embedded layers from a previous project version - displays them when opening the first time but after saving and reopening embed layers not saved.

pblottiere:
Fixes #34060


nyalldawson:
Can we get a unit test?

pblottiere:
> Can we get a unit test?

Of course, it's on its way :).

pblottiere:
This PR fixes 2 things.

The first one is an invalid project path when loading an embedded project, resulting in the next message `WARNING    Error zip file does not exist:` (as indicating in the issue).

The second one is a misleading parameter in `createEmbeddedLayer` calls:

https://github.com/qgis/QGIS/blob/82324ece0c9da91e1b3b2663cd70b0c3a8951965/src/core/qgsproject.cpp#L1008

The last `flags` parameter is actually at the fourth position instead of the `saveFlag` parameter which should be set to `true`. BTW, I don't understand why the compiler is not complaining... What do you think @nyalldawson? Am I missing something?

Moreover, considering the fact that embedded layers/groups are regularly broken, the `TestQgsProject` class is now a friend class of `QgsProject` in order to test underlying mechanisms more accurately.

PR #34312, opened 2020-02-05
[Oracle] Don't crash on getFeature(id) with an invalid id

troopa81:
## Description

In some edge cases, user call getFeature(id) with non existing id. It happens for instance with float or string column as primary key if the mapping fid <-> column value has not being built yet.

This PR fixes this and return an invalid feature in such case.

### DIFF
--- a/src/providers/oracle/qgsoracleprovider.cpp
+++ b/src/providers/oracle/qgsoracleprovider.cpp
@@ -418,11 +418,9 @@ QString QgsOracleUtils::whereClause( QgsFeatureId featureId, const QgsFields &fi
     case PktRowId:
     case PktFidMap:
     {
-      QVariant pkValsVariant = sharedData->lookupKey( featureId );
-      if ( !pkValsVariant.isNull() )
+      QVariantList pkVals = sharedData->lookupKey( featureId );
+      if ( !pkVals.isEmpty() )
       {
-        QList<QVariant> pkVals = pkValsVariant.toList();
-
         if ( primaryKeyType == PktFidMap )
         {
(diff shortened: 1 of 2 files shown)


src/providers/oracle/qgsoracleprovider.cpp
```diff
@@ -418,11 +418,9 @@ QString QgsOracleUtils::whereClause( QgsFeatureId featureId, const QgsFields &fi
     case PktRowId:
     case PktFidMap:
     {
-      QVariant pkValsVariant = sharedData->lookupKey( featureId );
-      if ( !pkValsVariant.isNull() )
+      QVariantList pkVals = sharedData->lookupKey( featureId );
```
nyalldawson: What's actually happening here? Is a non-list value getting added to the lookup somehow?

troopa81: No, the lookup map is empty, and returns en [empty QVariantList](https://github.com/troopa81/QGIS/blob/fix_wcs_blank_tile/src/providers/oracle/qgsoracleprovider.cpp#L3381) for the given featureId. Empty list is not a null variant, so the condition needs to be changed (it's pretty much the same behavior in postgres provider)

troopa81:
Travis error seems unrelated to this commit, so I close/reopen the PR

src/providers/oracle/qgsoracleprovider.cpp
nyalldawson: Ah, good catch

nyalldawson: Looks like the postgres provider already has this fix too

nyalldawson (approved):

PR #34319, opened 2020-02-06
Some very ugly ports away from deprecated Qt methods

nyalldawson:
There's no straightforward way to avoid the deprecated methods here, so we have to
copy Qt5's compatibility helpers. Ugly, but unavoidable.

### DIFF
--- a/src/core/qgsvirtuallayerdefinition.cpp
+++ b/src/core/qgsvirtuallayerdefinition.cpp
@@ -168,58 +168,141 @@ QgsVirtualLayerDefinition QgsVirtualLayerDefinition::fromUrl( const QUrl &url )
   return def;
 }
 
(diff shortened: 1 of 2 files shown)


3nids:
wouldn't it be nicer to paste them in qgis.h or maybe some other new header?
If this kind of code is centralized, it is more likely to be reused and globally replaced in the future.
But is there a requirement that it stays in the cpp files?

nyalldawson:
> wouldn't it be nicer to paste them in qgis.h or maybe some other new header?

We don't want to use these anywhere. Not even where they are used now :) I'm hoping someone else can find an alternative approach which avoids the pre-encoded requirements, but failing that I'd much rather hide these ugly details away from anyone using them elsewhere.

PR #34320, opened 2020-02-06
use scope enums for QgsMapLayerAction::Target and Flag

3nids:
because it just looks way better in the PyQGIS docs


3nids:
Qt issue preventing from using scope enums for flags.
Has been fixed in Qt 5.12.
https://bugreports.qt.io/browse/QTBUG-47652

nyalldawson:
Time to bump the qt min?

3nids:
Ubuntu 18.04 is 5.9
Ubuntu 18.10 is EOL
Ubuntu 19.04 and above are 5.12

Debian Buster (19.07) is..... 5.11 which is a pity since Qt 5.12 is LTS.

Saijin-Naib:
Windows x64 release/master both appear to still be QT 5.11.2

PR #34322, opened 2020-02-06
Fix IN/NOT IN evaluation with very number like strings

### LINKED ISSUE #34314: Select by attribute selects the wrong features
**Describe the bug**
Using the *select by attribute* tool selects the wrong features under circumstances.
In my case, it happens when using a `WHERE field IN ('values')` clause.

**How to Reproduce**

[select_in.gpkg.zip](https://github.com/qgis/QGIS/files/4161277/select_in.gpkg.zip)
Load the attached geopackage and use select by attribute:
- `g_prop_id = '010080383000187219'` selects one feature, ok
- `G_PROP_ID in ('010080383000187219')` selects one feature, ok
-  `g_prop_id in ('010080383000187219')` selects three features, not ok

Using these expressions as a filter always returns correctly one feature.

**QGIS and OS versions**
Debian with a week old master build.

**Additional context**
First noticed on Oracle layer, exported to geopackage and it still happens.

nyalldawson:
Only allow evaluation of number-like strings as numbers when
we aren't comparing two strings (matches behavior with other
expression comparison functions like equals)

Fixes #34314

### DIFF
--- a/src/core/expression/qgsexpressionnodeimpl.cpp
+++ b/src/core/expression/qgsexpressionnodeimpl.cpp
@@ -807,8 +807,11 @@ QVariant QgsExpressionNodeInOperator::evalNode( QgsExpression *parent, const Qgs
     {
       bool equal = false;
       // check whether they are equal
-      if ( QgsExpressionUtils::isDoubleSafe( v1 ) && QgsExpressionUtils::isDoubleSafe( v2 ) )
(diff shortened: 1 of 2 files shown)


nyalldawson:
Yes, it's because qgis expressions aren't strongly typed and a string field with value '5' can be used in "field"*2 without issues.

PR #34325, opened 2020-02-06
Fix KML/GPX export coordinate transformation not matching context

nirvn:
## Description

When exporting to an OGR dataset, some drivers (AFAIK only KML) enforce one specific CRS. Until now, we left it to the OGR library to do the transformation, however under GDAL3/PROJ6, we want to be in charge of transformation to insure it matches the project/layer context (i.e. user picked transformation).

### DIFF
--- a/src/core/processing/qgsprocessingutils.cpp
+++ b/src/core/processing/qgsprocessingutils.cpp
@@ -627,9 +627,14 @@ QgsFeatureSink *QgsProcessingUtils::createFeatureSink( QString &destination, Qgs
       // use QgsVectorFileWriter for OGR destinations instead of QgsVectorLayerImport, as that allows
       // us to use any OGR format which supports feature addition
       QString finalFileName;
-      std::unique_ptr< QgsVectorFileWriter > writer = qgis::make_unique< QgsVectorFileWriter >( destination, options.value( QStringLiteral( "fileEncoding" ) ).toString(), newFields, geometryType, crs, format, QgsVectorFileWriter::defaultDatasetOptions( format ),
-          QgsVectorFileWriter::defaultLayerOptions( format ), &finalFileName, QgsVectorFileWriter::NoSymbology, sinkFlags );
-
+      QgsVectorFileWriter::SaveVectorOptions saveOptions;
+      saveOptions.fileEncoding = options.value( QStringLiteral( "fileEncoding" ) ).toString();
+      saveOptions.driverName = format;
+      saveOptions.datasourceOptions = QgsVectorFileWriter::defaultDatasetOptions( format );
+      saveOptions.layerOptions = QgsVectorFileWriter::defaultLayerOptions( format );
+      saveOptions.symbologyExport = QgsVectorFileWriter::NoSymbology;
+      saveOptions.actionOnExistingFile = QgsVectorFileWriter::CreateOrOverwriteFile;
+      std::unique_ptr< QgsVectorFileWriter > writer( QgsVectorFileWriter::create( destination, newFields, geometryType, crs, context.transformContext(), saveOptions, sinkFlags, &finalFileName ) );
       if ( writer->hasError() )
       {
         throw QgsProcessingException( QObject::tr( "Could not create layer %1: %2" ).arg( destination, writer->errorMessage() ) );
@@ -892,8 +897,10 @@ QString convertToCompatibleFormatInternal( const QgsVectorLayer *vl, bool select
   {
     QString temp = QgsProcessingUtils::generateTempFilename( baseName + '.' + preferredFormat );
 
-    QgsVectorFileWriter writer( temp, context.defaultEncoding(),
-                                vl->fields(), vl->wkbType(), vl->crs(), QgsVectorFileWriter::driverForExtension( preferredFormat ) );
+    QgsVectorFileWriter::SaveVectorOptions saveOptions;
+    saveOptions.fileEncoding = context.defaultEncoding();
+    saveOptions.driverName = QgsVectorFileWriter::driverForExtension( preferredFormat );
+    std::unique_ptr< QgsVectorFileWriter > writer( QgsVectorFileWriter::create( temp, vl->fields(), vl->wkbType(), vl->crs(), context.transformContext(), saveOptions ) );
     QgsFeature f;
     QgsFeatureIterator it;
     if ( selectedFeaturesOnly )
@@ -905,7 +912,7 @@ QString convertToCompatibleFormatInternal( const QgsVectorLayer *vl, bool select
     {
       if ( feedback->isCanceled() )
         return QString();
-      writer.addFeature( f, QgsFeatureSink::FastInsert );
+      writer->addFeature( f, QgsFeatureSink::FastInsert );
     }
     return temp;
   }
--- a/src/core/qgsabstractgeopdfexporter.cpp
+++ b/src/core/qgsabstractgeopdfexporter.cpp
@@ -15,6 +15,7 @@
  ***************************************************************************/
 
 #include "qgsabstractgeopdfexporter.h"
+#include "qgscoordinatetransformcontext.h"
 #include "qgsrenderedfeaturehandlerinterface.h"
 #include "qgsfeaturerequest.h"
 #include "qgslogger.h"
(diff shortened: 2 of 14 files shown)


gioman:
GPX too maybe?

nirvn:
@gioman , I'm not sure it's appropriate here, as OGR doesn't transform the coordinates when saving. The .gpx file opens as WGS84 but the saved geometries have kept the source CRS coordinates.

nirvn:
@rouault , looking at the GPX spec, it's explicitly stated that coordinates are always in wgs84 degrees. Should GDAL's GPX driver behave like its KML driver and force convert geometries?

@gioman , on our end of things, we can force transform to WGS84.

rouault:
> Should GDAL's GPX driver behave like its KML driver and force convert geometries?

could be appropriate indeed.

src/core/qgsvectorfilewriter.cpp
```diff
@@ -393,6 +398,17 @@ void QgsVectorFileWriter::init( QString vectorFileName,
   }
 
   // consider spatial reference system of the layer
+  if ( driverName == QLatin1String( "KML" ) || driverName == QLatin1String( "GPX" ) )
+  {
+    if ( srs.authid() != QStringLiteral( "EPSG:4326" ) )
+    {
+      // Those drivers outputs WGS84 geometries, let's align our output CRS to have QGIS take charge of geometry transformation
+      QgsCoordinateReferenceSystem wgs84 = QgsCoordinateReferenceSystem::fromEpsgId( 4326 );
+      mCoordinateTransform = new QgsCoordinateTransform( srs, wgs84, transformContext );
```
nyalldawson: leaks, better make it a unique_ptr

nirvn: @nyalldawson , done.

nirvn:
@nyalldawson , OK to merge?

src/core/qgsvectorfilewriter.h
```diff
@@ -591,7 +593,8 @@ class CORE_EXPORT QgsVectorFileWriter : public QgsFeatureSink
                          QgsVectorFileWriter::FieldValueConverter *fieldValueConverter,
                          const QString &layerName,
                          QgsVectorFileWriter::ActionOnExistingFile action,
-                         QString *newLayer = nullptr
+                         QString *newLayer = nullptr,
+                         QgsCoordinateTransformContext transformContext = QgsCoordinateTransformContext()
```
nyalldawson: We really need to make these context arguments mandatory here, or we'll get unpredictable results depending on where files are created from (e.g. they are not actually set anywhere in this PR - so regardless of the user's settings you'll just get a default transform).

nirvn:
@nyalldawson , OK, I've implemented your comment and travis is (finally) green. Ready to (hopefully) approve and merge.

src/core/processing/qgsprocessingutils.cpp
```diff
@@ -892,8 +897,10 @@ QString convertToCompatibleFormatInternal( const QgsVectorLayer *vl, bool select
   {
     QString temp = QgsProcessingUtils::generateTempFilename( baseName + '.' + preferredFormat );
 
-    QgsVectorFileWriter writer( temp, context.defaultEncoding(),
-                                vl->fields(), vl->wkbType(), vl->crs(), QgsVectorFileWriter::driverForExtension( preferredFormat ) );
+    QgsVectorFileWriter::SaveVectorOptions saveOptions;
+    saveOptions.fileEncoding = context.defaultEncoding();
+    saveOptions.driverName = QgsVectorFileWriter::driverForExtension( preferredFormat );
+    std::unique_ptr< QgsVectorFileWriter > writer( QgsVectorFileWriter::create( temp, vl->fields(), vl->wkbType(), vl->crs(), vl->transformContext(), saveOptions ) );
```
nyalldawson:
```suggestion
    std::unique_ptr< QgsVectorFileWriter > writer( QgsVectorFileWriter::create( temp, vl->fields(), vl->wkbType(), vl->crs(), context.transformContext(), saveOptions ) );
```

src/core/qgsabstractgeopdfexporter.cpp
```diff
@@ -162,19 +163,22 @@ bool QgsAbstractGeoPdfExporter::saveTemporaryLayers()
       // write out features to disk
       const QgsFeatureList features = it.value();
       QString layerName;
-      QgsVectorFileWriter writer( filePath, QString(), features.first().fields(), features.first().geometry().wkbType(), QgsCoordinateReferenceSystem(), QStringLiteral( "GPKG" ), QStringList(), QStringList(), nullptr, QgsVectorFileWriter::NoSymbology, QgsFeatureSink::RegeneratePrimaryKey, &layerName );
-      if ( writer.hasError() )
+      QgsVectorFileWriter::SaveVectorOptions saveOptions;
+      saveOptions.driverName = QStringLiteral( "GPKG" );
+      saveOptions.symbologyExport = QgsVectorFileWriter::NoSymbology;
+      std::unique_ptr< QgsVectorFileWriter > writer( QgsVectorFileWriter::create( filePath, features.first().fields(), features.first().geometry().wkbType(), QgsCoordinateReferenceSystem(), QgsCoordinateTransformContext(), saveOptions, QgsFeatureSink::RegeneratePrimaryKey, nullptr, &layerName ) );
```
nyalldawson: Hm, I'll let this one pass :P

src/core/qgsvectorfilewriter.cpp
```diff
@@ -2499,6 +2534,11 @@ gdal::ogr_feature_unique_ptr QgsVectorFileWriter::createFeature( const QgsFeatur
     {
       // build geometry from WKB
       QgsGeometry geom = feature.geometry();
+      if ( mCoordinateTransform )
+      {
+        // output dataset requires coordinate transform
+        geom.transform( *mCoordinateTransform );
```
nyalldawson: Better try/catch a QgsCsException here

src/core/qgsvectorfilewriter.h
```diff
@@ -543,22 +543,26 @@ class CORE_EXPORT QgsVectorFileWriter : public QgsFeatureSink
...
-                         const QStringList &datasourceOptions = QStringList(),
-                         const QStringList &layerOptions = QStringList(),
-                         QString *newFilename = nullptr,
-                         QgsVectorFileWriter::SymbologyExport symbologyExport = QgsVectorFileWriter::NoSymbology,
-                         QgsFeatureSink::SinkFlags sinkFlags = nullptr
+    /**
+     * Create a new vector file writer
+     * \deprecated since QGIS 3.10.3, use the QgsVectorFileWriter::create function.
+     */
+    Q_DECL_DEPRECATED QgsVectorFileWriter( const QString &vectorFileName,
```
nyalldawson: SIP_DEPRECATED

src/core/qgsvectorfilewriter.h
```diff
@@ -576,6 +580,8 @@ class CORE_EXPORT QgsVectorFileWriter : public QgsFeatureSink
      * \param layerName layer name. If let empty, it will be derived from the filename (added in QGIS 3.0)
      * \param action action on existing file (added in QGIS 3.0)
      * \param newLayer potentially modified layer name (output parameter) (added in QGIS 3.4)
+     * \param transformContext transform context, needed if the output file srs is forced to specific crs (added in QGIS 3.10.3)
+     * \param sinkFlags feature sink flags (added in QGIS 3.10.3)
      * \note not available in Python bindings
      */
     QgsVectorFileWriter( const QString &vectorFileName,
```
nyalldawson: This should be deprecated too

nyalldawson (commented):
This is great work!

But...  it doesn't actually work yet :smile: 

Because there's still the QgsVectorFileWriter::writeAsVectorFormat  and prepareWriteAsVectorFormat static methods which aren't deprecated and which don't take a context argument, it means that there's parts of qgis exporting layers which won't correctly transform using the context settings. Including (importantly!) the right click layer -> save as action. :stuck_out_tongue: 

We'll need to deprecate these ones and add new ones which require a transform context argument. I'd suggest here we just go with the ugly V2 suffixes...

src/core/qgsvectorfilewriter.h
nirvn: I initially did, but deprecating that constructor leads to a warning when we *have* to use it in our static create function. Solution?

nyalldawson: Wrap the call in a NOWARN block -- I don't think we can cleanly fix this whole class until 4.0 when we can hard break it

src/core/qgsvectorfilewriter.h
```diff
@@ -325,7 +325,7 @@ class CORE_EXPORT QgsVectorFileWriter : public QgsFeatureSink
      * \param fieldValueConverter field value converter (added in QGIS 2.16)
      */
 #endif
-    static QgsVectorFileWriter::WriterError writeAsVectorFormat( QgsVectorLayer *layer,
+    Q_DECL_DEPRECATED static QgsVectorFileWriter::WriterError writeAsVectorFormat( QgsVectorLayer *layer,
```
nyalldawson: Don't forget the "\deprecated Use ... " dox too

nyalldawson (approved):
Merge away!

nirvn:
Thanks.

PR #34336, opened 2020-02-06
Fix form filtering when used in feature selection dialog

troopa81:
## Description

Filter expression was not updated on form filtering in feature selection dialog

### DIFF
--- a/src/gui/qgsfeatureselectiondlg.cpp
+++ b/src/gui/qgsfeatureselectiondlg.cpp
@@ -50,6 +50,9 @@ QgsFeatureSelectionDlg::QgsFeatureSelectionDlg( QgsVectorLayer *vl, const QgsAtt
   connect( mActionSelectedToTop, &QAction::toggled, this, [this]( bool checked ) { mDualView->setSelectedOnTop( checked ); } );
   connect( mActionZoomMapToSelectedRows, &QAction::triggered, this, &QgsFeatureSelectionDlg::mActionZoomMapToSelectedRows_triggered );
   connect( mActionPanMapToSelectedRows, &QAction::triggered, this, &QgsFeatureSelectionDlg::mActionPanMapToSelectedRows_triggered );
+
+  connect( mDualView, &QgsDualView::filterExpressionSet, this, &QgsFeatureSelectionDlg::setFilterExpression );
+  connect( mDualView, &QgsDualView::formModeChanged, this, &QgsFeatureSelectionDlg::viewModeChanged );
 }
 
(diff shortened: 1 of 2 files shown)


troopa81:
Don't stale

src/gui/qgsfeatureselectiondlg.cpp
```diff
@@ -128,3 +131,14 @@ void QgsFeatureSelectionDlg::mActionPanMapToSelectedRows_triggered()
...
+
+void QgsFeatureSelectionDlg::setFilterExpression( const QString &filter, QgsAttributeForm::FilterType type )
+{
+  mFeatureFilterWidget->setFilterExpression( filter, type, true );
+}
+
+void QgsFeatureSelectionDlg::viewModeChanged( QgsAttributeEditorContext::Mode mode )
+{
+  if ( mode != QgsAttributeEditorContext::SearchMode )
+    mActionSearchForm->setChecked( false );
```
m-kuhn: Does this not need to be changed in the other direction too?

```suggestion
    mActionSearchForm->setChecked( mode == QgsAttributeEditorContext::SearchMode );
```

troopa81: I don't think so, because mActionSearchForm is already checked. The only way to switch to SearchMode is to toggle mActionSearchForm

m-kuhn: In case someone adds a keyboard shortcut or other fancy way to switch this behavior, would it hurt to make sure the state is synchronized?

troopa81: I was afraid that it triggers another useless signal even if the state didn't change, but it's not the case according to doc and my tests. I made the modifications.

m-kuhn: For reference, when I write properties, I normally do the following (and I think I started when I read somewhere Qt handles it the same way to avoid binding loops):

```
void setProperty( const Property &property )
{
  if ( property == mProperty )
    return;

  mProperty = property;
  emit propertyChanged();
}
```

m-kuhn (approved):

src/gui/qgsfeatureselectiondlg.cpp
troopa81: Good to know. But I think it not the case for QLineEdit::setText for instance which fire a textChanged on every setText even if the text hasn't changed

troopa81:
is it OK for merging @m-kuhn ? If so, could you please merge it? Thanks

m-kuhn (approved):

m-kuhn:
Yep, thanks!

PR #34337, opened 2020-02-06
Show field type icons on attribute table headers

uclaros:
## Description
I really miss having this info on the attribute table after getting used to other database managers.

Screenshot_20200206_212933
If it is on purpose not there, in order not to clutter the view, then maybe there could be a settings entry to enable / disable it.
Any other reason not to have this?

### DIFF
--- a/src/app/qgsoptions.cpp
+++ b/src/app/qgsoptions.cpp
@@ -679,7 +679,8 @@ QgsOptions::QgsOptions( QWidget *parent, Qt::WindowFlags fl, const QList<QgsOpti
   mDataSourceManagerNonModal->setChecked( mSettings->value( QStringLiteral( "/qgis/dataSourceManagerNonModal" ), false ).toBool() );
   cbxCheckVersion->setChecked( mSettings->value( QStringLiteral( "/qgis/checkVersion" ), true ).toBool() );
   cbxCheckVersion->setVisible( mSettings->value( QStringLiteral( "/qgis/allowVersionCheck" ), true ).toBool() );
-  cbxAttributeTableDocked->setChecked( mSettings->value( QStringLiteral( "/qgis/dockAttributeTable" ), false ).toBool() );
+  cbxAttributeTableDocked->setChecked( mSettings->value( QStringLiteral( "/qgis/dockAttributeTable" ), true ).toBool() );
+  cbxAttributeTableHeaderIcons->setChecked( mSettings->value( QStringLiteral( "/qgis/showAttributeTableHeaderIcons" ), false ).toBool() );
   cbxCompileExpressions->setChecked( mSettings->value( QStringLiteral( "/qgis/compileExpressions" ), true ).toBool() );
 
   mComboCopyFeatureFormat->addItem( tr( "Plain text, no geometry" ), QgsClipboard::AttributesOnly );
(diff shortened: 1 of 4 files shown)


roya0045:
Good move @uclaros !

nyalldawson:
@nirvn thoughts?

For reference -- this information is currently available by hovering over the column header

nirvn:
-0.5, I'm not a big fan of this for two reasons:
1/ this takes up horizontal space away from actual column name, which will be problematic for datasets with a large nb of attributes;
2/ it's not dark theme friendly.

uclaros:
1. What about adding a settings entry, defaulting to false? I know I would disable it too under certain circumstances.
2. Most icons are not designed to be dark theme friendly, being either too bright, having white background, low contrast with the dark background etc. Those specific icons don't look too bad on the attribute table imo. They are almost invisible though on the field drop-down when calculating fields...

gioman:
> -0.5, I'm not a big fan of this for two reasons:
> 1/ this takes up horizontal space away from actual column name, which will be problematic for datasets with a large nb of attributes;
> 2/ it's not dark theme friendly.
> 
> Edit: it also risks unnecessarily "stealing focus". Header should not get in the way visually.

While I understand this points, this is still of great help/utility for the vast majority of common users.

roya0045:
I think adding a setting would be good, though I think it should default to true as most users would benefit from it. @uclaros I didn't check but could a right click in the header provide the option to display or hide those icons? I feel like a right click would be optimal for this as having a dedicated option would be annoying to toggle.

nirvn:
I'd challenge the assertion made here that this is something "most users" / "common users" are in need of. 

Field type is mostly something you'd need to know when modifying the field. In this case, we have the calculator bar which shows field type, as well as edit widgets hinting at type.

gioman:
> I'd challenge the assertion made here that this is something "most users" / "common users" are in need of.
> 
> Field type is mostly something you'd need to know when modifying the field. In this case, we have the calculator bar which shows field type, as well as edit widgets hinting at type.

sure, but maybe your experience of users is different from mine. Most of the users I meet/woork with have little to no idea of what data types are the columns they are working with, they usually (mostly) overlook the hover on the column header and ends up looking at data type in vector properties.

SrNetoChan:
I think it's useful for most users as well. As useful as in any other place that the icons now appear. No one remembers all field types in their data.

Besides, if there's a setting to turn it of, why not?

uclaros:
@nirvn A use case is having, among others, three fields which are a(int), b(str), c(str). You open the calculator bar, select a (you see it's int) and type *b+c* and you get `1 + 1 = 11`. 
Getting the datatypes of all fields at a glance solves this and helps you know your data better.

ps. IMO we also need those to show on delete fields and reorder fields spartan dialogs

nirvn:
Big +1 on adding icons on the delete n reorder fields dialogs. 

As for expressions, we could add the field type in the auto complete tooltip as users type in the field.

uclaros:
@nirvn I'll open a feature request for that!

Is there a simple way to disable showing the icons in the table view instead of the model? I think it would be preferred, but I couldn't find one.

3nids:
I'm also -0 on having the field type in the attribute table as a header icon. I'd rather have it as a tooltip. But no big deal to me.
But most important to me, please avoid adding a new setting. I think it is best to try to keep settings to critical things and that you have a continuum when you switch between work stations.

nyalldawson:
I'm -1 too - there's just too much happening in that screenshot. One possible solution would be to only show these icons while in edit mode for the layer, where they are arguably much more useful to see..

gioman:
> One possible solution would be to only show these icons while in edit mode for the layer, where they are arguably much more useful to see..

arguably, but I'll take it (just in edit mode rather than nothing).

uclaros:
@nyalldawson There's too many NULLs in the screenshot making it look like there's too much going on! Dbeaver database manager also has icons for filter and sort and it still does not look too congested to me, you get used to it fast.

@3nids As a user, I was thinking that settings are ideal for this kind of ux customization. After all, there are already such settings that can make switching workstations confusing, like changing the ui theme, changing the locator prefixes, or snapping on invisible features to name a few. Type icons are nowhere near that big a change.

roya0045:
@uclaros could the icons be semi transparent in the background of the header? Though this might make things unclear...

3nids:
@uclaros don't get me wrong, settings are good when they affect some substantial behaviors. They are already a lot of settings, and adding more just brings too much distraction.

More generally, I think we should have a procedure to deal with these kind of UX changes where there is no direct consensus (I don't think the QEP is the approach here). @nirvn ?

uclaros:
Did this hit a dead end by not wanting to add a new setting?
Could there be another way to enable this? In the (small) sample of 10 people involved in the conversation 7 find it useful, it's not insignificant.

gioman:
> it's not insignificant.

agree

roya0045:
It got lost in limbo, that happens.

PR #34347, opened 2020-02-07
Change "Ignore shapefile encoding declaration" default from checked to unchecked

nyalldawson:
...and resolve shapefiles opened in QGIS ignore the cpg/embedded encoding by default.

This is a messy messy situation. The default was changed some years ago to ignore shapefile .cpg encoding BY DEFAULT, because if we don't ignore it, then it becomes impossible to override the encoding on a file-by-file basis (i.e. the setting in layer properties gets disabled and the encoding setting in the data source manager has no effect).

But by ignoring it by default we are ignoring the encoding specified by the files themselves. It basically means that users ALWAYS need to manually select the encoding for shapefiles, because we ignore the logic on OGR's side to select the embedded one automatically. This has led to a lot of user pain -- see #21264 and many posts on mailing lists e.g.

http://osgeo-org.1560.x6.nabble.com/Shapefile-with-file-cpg-codepage-td5275106.html
http://osgeo-org.1560.x6.nabble.com/QGIS-ignore-the-cpg-files-when-loading-shapefiles-td5348021.html

@rouault I spent some time trying to see if we could make a solution which satisfied both requirements -- by defaulting to an "automatic" encoding option which means trust GDAL, but I cannot find a reliable way to change the SHAPE_ENCODING configuration option used by GDAL after the layer has been initially loaded (i.e. if the user selects a specific encoding we need to change SHAPE_ENCODING to "" instead of nullptr for the already opened dataset. I thought calling QgsOgrProvider::reloadProviderData() after changing SHAPE_ENCODING may work for this, but doing so doesn't disable the UTF-8 conversion on GDAL's side...

The other approach I tried was to always disable the GDAL side UTF-8 conversion for shapefiles, but then I can't find a way to still take advantage of GDAL's .cpg/LDID handling logic and retrieve the encoding detected by GDAL (and I don't want to reimplement this logic in QGIS, if that's even possible to do)

Can you think of a nicer way to handle this in general?

### DIFF
--- a/resources/qgis_global_settings.ini
+++ b/resources/qgis_global_settings.ini
@@ -152,6 +152,11 @@ rendering\label_candidates_limit_lines=0
 # that no limit is present.
 rendering\label_candidates_limit_polygons=0
 
+# When false, QGIS uses information embedded inside Shapefiles to determine the correct encoding to use
+# when opening them (e.g. encoding specified in a cpg file). If true, QGIS will ignore this embedded encoding
+# and use UTF8 encoding by default
+dataProviders\ignoreShapeEncoding=false
+
 [colors]
 # These colors are used in logs.
 default=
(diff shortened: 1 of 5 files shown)


nyalldawson:
Better approach inbound

PR #34355, opened 2020-02-07
In relation reference widget, fix chain filter when null values are not allowed 

troopa81:
## Description

In relation reference widget, chain filters are not updated when null values are not allowed.

It works well when null values are allowed
chainfilter

This PR propose to fix this.

### DIFF
--- a/src/core/qgsfeaturefiltermodel.cpp
+++ b/src/core/qgsfeaturefiltermodel.cpp
@@ -288,7 +288,7 @@ void QgsFeatureFilterModel::updateCompleter()
     int firstRow = 0;
 
     // Move the extra entry to the first position
-    if ( mExtraIdentifierValueIndex != -1 )
+    if ( mExtraIdentifierValueIndex != -1 && currentEntryInNewList != -1 )
     {
       if ( mExtraIdentifierValueIndex != 0 )
       {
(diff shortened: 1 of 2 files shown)


troopa81:
Don't stale

m-kuhn (approved):

m-kuhn:
Thank you!

elpaso:
@troopa81 I'm afraid this commit caused a regression breaking the ability to show extra identifier values (i.e. not existant features) as `(xx)` in the combo.

@m-kuhn (remember we chatted about that a few days ago?)

troopa81:
@elpaso I don't think its related to this modification because this is broken since at least 3.10 (works in 3.4).

To be sure we do talk about the same thing, here is a little demo gif, where I delete a feature referenced by another layer. The combo box switch then to next feature.

featuremissing

elpaso:
@troopa81 no, we are talking about a different issue and I checked that 3.10 behaves as expected: when you reference a not-yet-existant feature in the combo, the item was selected shown as "(xx)" where xx is the not-yet-existant feature id. 

After this commit the combo is empty (no item selected and no "(xx)" available.

elpaso:
In this image from 3.10, the parent feature (referenced layer) id is 21, the feature was not saved (not even in the editing buffer, it was just added), the dialog in the image is after adding a referencing point feature (child of feature 21).

immagine

troopa81:
@elpaso ok, I see. I'll take a look

troopa81:
@elpaso I fail to reproduce the working scenario with 3.10, could you provide me your test data and project?

elpaso:
@troopa81, you can reproduce it with 3.12 (latest release from ubuntugis) too:

- open the project stored inside the gpkg
- start editing on area
- add an area polygon (don't save!)
- in the form, add a "sales"

[relations_test_project_3_12.gpkg.zip](https://github.com/qgis/QGIS/files/4286293/relations_test_project_3_12.gpkg.zip)

elpaso:
@troopa81 gentle ping, can we revert this commit in the meantime?

troopa81:
@elpaso I'm struggling to fix this one (without breaking everything else). I'm close to a working solution. If I don't have a working solution until tuesday, I'll revert it myself. Is that OK to you?

elpaso:
@troopa81 absolutely, thank you for taking care of this. I know it's not easy stuff.

PR #34361, opened 2020-02-07
Fix identify menu when an expression is used as display name

lbartoletti:
## Description
When an expression is filled in the display field, the identification menu always returns the ID of the clicked feature rather than the result of the expression.

In many projects we use expressions to clearly identify features, it's more readable than the ID. This PR fixes this problem.

### DIFF
--- a/src/gui/qgsidentifymenu.cpp
+++ b/src/gui/qgsidentifymenu.cpp
@@ -25,6 +25,7 @@
 #include "qgslogger.h"
 #include "qgssettings.h"
 #include "qgsgui.h"
+#include "qgsexpressioncontextutils.h"
 
 //TODO 4.0 add explicitly qobject parent to constructor
 QgsIdentifyMenu::QgsIdentifyMenu( QgsMapCanvas *canvas )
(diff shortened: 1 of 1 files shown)


src/gui/qgsidentifymenu.cpp
```diff
@@ -282,7 +283,12 @@ void QgsIdentifyMenu::addVectorLayer( QgsVectorLayer *layer, const QList<QgsMapT
   if ( !createMenu )
   {
     // case 1
-    QString featureTitle = results[0].mFeature.attribute( layer->displayField() ).toString();
+
+    QgsExpressionContext context( QgsExpressionContextUtils::globalProjectLayerScopes( layer ) );
```
nyalldawson: Can't we reuse this context throughout ::addVectorLayer, instead of the code duplication?

nyalldawson (approved):

src/gui/qgsidentifymenu.cpp
lbartoletti: Yes. Thanks

PR #34367, opened 2020-02-08
Make QgsExifTools QML friendly

nirvn:
## Description

This PR tweaks the useful QgsExifTools to make it QML-friendly, and adds a shortcut hasGeoTag() function that essentially calls getGeoTag(...) (which isn't callable via QML due to the presence of reference parameters).

@3nids , all good?

### DIFF
--- a/src/analysis/raster/qgsexiftools.cpp
+++ b/src/analysis/raster/qgsexiftools.cpp
@@ -100,6 +100,13 @@ QString doubleToExifCoordinate( const double val )
   return QStringLiteral( "%1/1 %2/1 %3/1000" ).arg( degrees ).arg( minutes ).arg( seconds );
 }
 
+bool QgsExifTools::hasGeoTag( const QString &imagePath )
+{
+  bool ok = false;
+  QgsExifTools::getGeoTag( imagePath, ok );
+  return ok;
+}
+
 QgsPoint QgsExifTools::getGeoTag( const QString &imagePath, bool &ok )
 {
   ok = false;
--- a/src/analysis/raster/qgsexiftools.h
+++ b/src/analysis/raster/qgsexiftools.h
@@ -18,6 +18,8 @@
 
 #include "qgis_analysis.h"
 #include "qgspointxy.h"
+
+#include <QObject>
 #include <QString>
 #include <QVariant>
 #include <QVariantMap>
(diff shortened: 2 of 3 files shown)


src/analysis/raster/qgsexiftools.h
```diff
@@ -47,6 +51,13 @@ class ANALYSIS_EXPORT QgsExifTools
      */
     static QgsPoint getGeoTag( const QString &imagePath, bool &ok SIP_OUT );
 
+    /**
+     * Returns TRUE if the image at \a imagePath contains a valid geotag.
+     *
+     * \see getGeoTag()
+     */
+    Q_INVOKABLE static bool hasGeoTag( const QString imagePath );
```
m-kuhn:
```suggestion
    Q_INVOKABLE static bool hasGeoTag( const QString &imagePath );
```

src/analysis/raster/qgsexiftools.h
```diff
@@ -47,6 +51,13 @@ class ANALYSIS_EXPORT QgsExifTools
      */
     static QgsPoint getGeoTag( const QString &imagePath, bool &ok SIP_OUT );
 
+    /**
+     * Returns TRUE if the image at \a imagePath contains a valid geotag.
+     *
+     * \see getGeoTag()
```
m-kuhn: \since QGIS 3.12?

src/analysis/raster/qgsexiftools.cpp
```diff
@@ -100,6 +100,13 @@ QString doubleToExifCoordinate( const double val )
   return QStringLiteral( "%1/1 %2/1 %3/1000" ).arg( degrees ).arg( minutes ).arg( seconds );
 }
 
+bool QgsExifTools::hasGeoTag( const QString imagePath )
+{
+  bool ok;
```
m-kuhn:
```suggestion
  bool ok = false;
```

Just because ;P

m-kuhn (approved):

src/analysis/raster/qgsexiftools.h
nirvn: @m-kuhn , let me try; one reason why I've added this shortcut function is that QML doesn't support bool &ref. Let's see if it'll be fine with a referenced QString.

nirvn:
OK, for the record: QML is fine a QString reference parameter, it's not with a bool reference parameter. TIL.

@m-kuhn , comments addressed, will merge when travis is happy.

PR #34368, opened 2020-02-08
Fix setSubsetString return value on error

### LINKED ISSUE #34259: filter expression error returns true
Hi,

teaching QGIS on 3.4.15 today, we face a bug on MAC and windows (have to try on ubuntu) :
When we create a filter on a layer, if the expression contains an error ( for example a parenthesis not closed), the expression returns true so the layer is not filtered.
filter_error_returns_true

uclaros:
## Description
`QgsOgrProvider::setSubsetString` should return false on malformed sql.

Fixes #34259

### DIFF
--- a/src/core/providers/ogr/qgsogrprovider.cpp
+++ b/src/core/providers/ogr/qgsogrprovider.cpp
@@ -4368,7 +4368,10 @@ OGRLayerH QgsOgrProviderUtils::setSubsetString( OGRLayerH layer, GDALDatasetH ds
   }
   else
   {
-    OGR_L_SetAttributeFilter( layer, encoding->fromUnicode( subsetString ).constData() );
+    if ( OGR_L_SetAttributeFilter( layer, encoding->fromUnicode( subsetString ).constData() ) != OGRERR_NONE )
+    {
+      return nullptr;
+    }
     subsetLayer = layer;
   }
 
--- a/tests/src/python/test_provider_shapefile.py
+++ b/tests/src/python/test_provider_shapefile.py
@@ -712,6 +712,27 @@ class TestPyQgsShapefileProvider(unittest.TestCase, ProviderTestCase):
         self.assertEqual(_lessdigits(subSet_vl.extent().toString()), filtered_extent)
         self.assertNotEqual(_lessdigits(subSet_vl.extent().toString()), unfiltered_extent)
 
+    def testMalformedSubsetStrings(self):
+        """Test that invalid where clauses allways return false"""
+
+        testPath = TEST_DATA_DIR + '/' + 'lines.shp'
+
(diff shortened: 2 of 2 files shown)


elpaso (approved):

src/core/providers/ogr/qgsogrprovider.cpp
```diff
@@ -4368,7 +4368,10 @@ OGRLayerH QgsOgrProviderUtils::setSubsetString( OGRLayerH layer, GDALDatasetH ds
   }
   else
   {
-    OGR_L_SetAttributeFilter( layer, encoding->fromUnicode( subsetString ).constData() );
+    if ( OGR_L_SetAttributeFilter( layer, encoding->fromUnicode( subsetString ).constData() ) != OGRERR_NONE )
+    {
+      return nullptr;
```
m-kuhn: Any chance to also log an error message if this happens?

uclaros: `QgsOgrProviderUtils::setSubsetString` is just a utility method used by `QgsOgrProvider::_setSubsetString` that handles the error and the logging:
https://github.com/qgis/QGIS/blob/0b4cc5adaeed099f83a47a527acefa6e2ac8bd86/src/core/providers/ogr/qgsogrprovider.cpp#L2067-L2073
then
https://github.com/qgis/QGIS/blob/0b4cc5adaeed099f83a47a527acefa6e2ac8bd86/src/core/qgsvectordataprovider.cpp#L705-L710

So I don't think extra logging is needed. I could add some unit test though, I just need to figure out where!...

m-kuhn: That sounds good.

Some tests exist already in `tests/src/python/test_provider_ogr*.py`, you could build on top of that

uclaros: Yes, but shapefile and spatialite tests are in `test_provider_shapefile.py` and `test_provider_spatialite.py`, are they not using ogrdataprovider? I think I'll put them on `test_provider_ogr.py` that is more generic, though currently all tests regarding `setSubsetString` are in each format's test file.

uclaros: Changed my mind and added the test on `test_provider_shapefile.py` since most ogr tests are there.

elpaso:
@uclaros any chance to see it backported to 3.10 and 3.12?  I don't see any potential regression unless there was some code relying on a plain wrong behavior.

elpaso:
sorry, I meant backport to 3.10 only.

elpaso:
@uclaros thank you!

uclaros:
@elpaso You're welcome :)

PR #34372, opened 2020-02-08
Reduce noise from doc bot

m-kuhn:
This hopefully only triggers the doc bot when setting the needs docs label (and not when setting any label)

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -11,7 +11,7 @@ jobs:
     name: Backport
     steps:
       - name: Backport Bot
-        if: contains( join(github.event.pull_request.labels.*.name), 'backport')
+        if: github.event.pull_request.merged && ( ( github.event.action == 'closed' && contains( github.event.pull_request.labels.*.name, 'backport') ) || contains( github.event.label.name, 'backport' ) )
         uses: Gaurav0/backport@v1.0.24
         with:
           bot_username: qgis-bot
--- a/.github/workflows/pr-needs-docs-message.yml
+++ b/.github/workflows/pr-needs-docs-message.yml
@@ -7,7 +7,7 @@ on:
       - labeled
 jobs:
   test:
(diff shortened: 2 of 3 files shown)


.github/workflows/backport.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     name: Backport
     steps:
       - name: Backport Bot
-        if: contains( join(github.event.pull_request.labels.*.name), 'backport')
+        if: github.event.pull_request.merged && ( ( github.event.action == 'closed' && contains( github.event.pull_request.labels.*.name, 'backport') ) || contains( github.event.label.name, 'backport' ) )
```
3nids:
```suggestion
        if: github.event.pull_request.merged && ( ( github.event.action == 'closed' && contains( github.event.pull_request.labels.*.name, 'backport') ) || github.event.label.name == 'backport' )
```

3nids: am I right?

3nids (approved):

.github/workflows/backport.yml
m-kuhn: Shouldn't this be "startswith" or similar?

3nids: ah sorry, yes!

3nids:
Not sure, but I had to remove and readd tag for the backport to work (didn't work on merge)
Will check tomorrow

.github/workflows/backport.yml
3nids: Are we sure merged is true when you actually merge it?

m-kuhn: After your merge the doc bot created a commit. I think that's a good hint it worked for the doc bot.

m-kuhn: Actually it should be `contains an element that startsWith backport` instead of `contains an element which is backport`

PR #34381, opened 2020-02-10
Fix numerous shapefile encoding issues

nyalldawson:
This fixes the (broken by design?) handling of Shapefile encoding, which has been an ongoing issue for years in QGIS.

See discussion at

- #21264
- http://osgeo-org.1560.x6.nabble.com/Shapefile-with-file-cpg-codepage-td5275106.html
- http://osgeo-org.1560.x6.nabble.com/QGIS-ignore-the-cpg-files-when-loading-shapefiles-td5348021.html

(+ others!)

The situation was that we had two different code paths for handling GDAL side attribute decoding OR QGIS side decoding. Unfortunately they are both incompatible with each other, and due to GDAL API for this we can't unify the two approaches. (More technical detail in the commit log message!)

So, now we:
- always do the decoding on QGIS' side. This allows users to manually override a shapefile's declared encoding, because they are often incorrect!
- use a port of GDAL's shapefile detection logic (it's not exposed in GDAL API, so I had to re-implement it here) so that we default to reading shapefiles by respecting the embedded encoding information (via cpg files or DBF LDID information)
- Completely remove the confusing/broken "Ignore shapefile encoding declaration" option, as it's no longer required -- users are ALWAYS able to manually change the encoding of shapefiles layers if needed
- Always show users the detected embedded encoding in the layer properties, instead of always showing "UTF-8" when the embedded encoding information is used

This should give the best of both worlds -- a nice default behavior resulting in shapefiles being read with the correct encoding, whilst still allowing users to override this on a layer-by-layer basis as needed.


rouault:
How does that play with /vsizip/ shapefiles ? It seems that if the user doesn't override the encoding, then the new detection routine will not be able to find the encoding from the file, and will default to no recoding ? Perhaps CPLSetConfigOption( "SHAPE_ENCODING", "" ); shouldn't be defined then.

An alternative to setting this (global) configuration option is to use the ENCODING open option of the shapefile driver in GDALOpenEx()

Regarding this comment
```
  // - we can't change this setting on-the-fly. If we don't set it upfront, we can't reverse this decision later when
  //   a user does want/need to manually specify the encoding
``` 
wouldn't a solution be to re-open the shapefile with a new ENCODING setting when QgsOgrProvider::setEncoding() is called ? Wouldn't that avoid completely the need to have QGIS to have the DBF/CPG logic ?

nyalldawson:
> How does that play with /vsizip/ shapefiles 

Right, I'll need to disable it here and force use of or side decoding

> wouldn't a solution be to re-open the shapefile with a new ENCODING setting when QgsOgrProvider::setEncoding() is called 

I initially tried this with no luck (behaviour was unpredictable -- when changing from ogr side decoding to qgis I'd frequently get corrupted decoded values)

nyalldawson:
(Dammit hit close by accident!)

> Wouldn't that avoid completely the need to have QGIS to have the DBF/CPG logic ?

Yes, but at the cost of being able to see what the actual encoding is in qgis.

rouault:
https://github.com/OSGeo/gdal/pull/2230 could potentially help for the future

nyalldawson:
@rouault 

> OSGeo/gdal#2230 could potentially help for the future

Yep, looks great. I'll ifdef that in and use it instead of the qgis version for newer gdal versions. 

Thinking on the vsi issue, I think it's ok to leave as is in this PR, because:
- Loading these files you'll get the same results as qgis has given by default since forever -- GDAL side decoding is disabled and the system encoding is used by default
- Users still have the option to manually set the encoding to the correct one (same as existing behavior, so no regressions). If we switch to GDAL side decoding for vsi sources then users will lose this ability
- There's a short-term regression ONLY if users have manually changed the cryptic "Ignore shapefile encoding declaration" option from its default setting, when a non-system encoded shp is loaded via vsi (previously GDAL side decoding would have been used for this situation, respecting the cpg and ldid info). But given OSGeo/gdal#2230 gives us the ultimate fix for this, it's a short-term regression only -- when 2230 can be used we'll get the correct (cpg/ldid) encoding used by default while still relying on QGIS side decoding.

nyalldawson:
@rouault an updated version which uses the GDAL 3.1 method was just pushed -- thanks for exposing this upstream, it's working great here now.

nirvn:
Would it be worth it backporting the gdal change to 3.0.5?

rouault:
> Would it be worth it backporting the gdal change to 3.0.5?

This is a new feature from GDAL point of view. Not sure when/if there will be a 3.0.5 but it would probably be around 3.1.0

kannes:
>     * Completely remove the confusing/broken "Ignore shapefile encoding declaration" option, as it's no longer required -- users are ALWAYS able to manually change the encoding of shapefiles layers if needed

Good riddance! Thank you for this.

PR #34403, opened 2020-02-11
fix ownership issues in dualview/attributetable

### LINKED ISSUE #33120: crash when showing a relation editor
since 328f365 
I am getting crashes from here
image

3nids:
correctly fix #33120


nyalldawson (commented):
Looks good, just need to update the dox and sip bindings

3nids:
@nyalldawson done

PR #34411, opened 2020-02-11
improvement of the Drag'n'Drop designer

### LINKED ISSUE #28570: Drag and drop designer with confusing layout
Original Redmine Issue: [20750](https://issues.qgis.org/issues/20750)
Affected QGIS version: 3.5(master)
Redmine category:gui

---
If I want to change the edit widget for a field I have to highlight it from "Available widget" pane. 
I often highlight the field from "Form layout" and after some time I notice, all the changes where done under the wrong field.

My proposal: disable widget settings pane when user  changing Form layout.

3nids:
* do not show editor when several fields are selected
* synchronize selection both ways

fixes #28570

Screen-Recording-2020-02-11-at-17 08 49

### DIFF
--- a/src/app/qgsattributesformproperties.cpp
+++ b/src/app/qgsattributesformproperties.cpp
@@ -821,14 +821,14 @@ void QgsAttributesFormProperties::apply()
 
   editFormConfig.setUiForm( mEditFormLineEdit->text() );
 
-  editFormConfig.setLayout( ( QgsEditFormConfig::EditorLayout ) mEditorLayoutComboBox->currentIndex() );
+  editFormConfig.setLayout( static_cast<QgsEditFormConfig::EditorLayout>( mEditorLayoutComboBox->currentIndex() ) );
 
   editFormConfig.setInitCodeSource( mInitCodeSource );
   editFormConfig.setInitFunction( mInitFunction );
   editFormConfig.setInitFilePath( mInitFilePath );
   editFormConfig.setInitCode( mInitCode );
 
-  editFormConfig.setSuppress( ( QgsEditFormConfig::FeatureFormSuppress )mFormSuppressCmbBx->currentIndex() );
+  editFormConfig.setSuppress( static_cast<QgsEditFormConfig::FeatureFormSuppress>( mFormSuppressCmbBx->currentIndex() ) );
 
   // relations
   QTreeWidgetItem *relationContainer = mAvailableWidgetsTree->invisibleRootItem()->child( 1 );
@@ -1104,7 +1104,7 @@ void DnDTree::onItemDoubleClicked( QTreeWidgetItem *item, int column )
     case QgsAttributesFormProperties::DnDTreeItemData::Container:
     {
       QDialog dlg;
-      dlg.setObjectName( QLatin1Literal( "attributeFormPropertiesContainerDialog" ) );
+      dlg.setObjectName( QLatin1String( "attributeFormPropertiesContainerDialog" ) );
       dlg.setWindowTitle( tr( "Configure Container" ) );
       QFormLayout *layout = new QFormLayout() ;
       dlg.setLayout( layout );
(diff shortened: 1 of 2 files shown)


m-kuhn:
here, take this hero-of-the-day badge :medal_sports:

saberraz:
Nice one!

3nids:
I would flag this as a bug and merge to master. I would maybe wait a bit before backporting to LTR.

src/app/qgsattributesformproperties.cpp
```diff
@@ -509,43 +511,72 @@ QTreeWidgetItem *QgsAttributesFormProperties::loadAttributeEditorTreeItem( QgsAt
...
     }
+  }
+}
 
+void QgsAttributesFormProperties::clearAttributeTypeFrame()
+{
+  if ( mAttributeTypeDialog )
+  {
+    mAttributeTypeFrame->layout()->removeWidget( mAttributeTypeDialog );
+    mAttributeTypeDialog->deleteLater();
```
signedav: I think, since you clear the `QgsAttributeTypeDialog` (like it should) it's not available anymore here: https://github.com/qgis/QGIS/blob/d1126c3a126e93b3bc28436968e8bc79c5b4583f/src/app/qgsattributesformproperties.cpp#L87 on cancellation of the layer property here:
https://github.com/qgis/QGIS/blob/master/src/app/qgsvectorlayerproperties.cpp#L953

This leads to a crash. It would fix it if you check the `QgsAttributeTypeDialog` for existence in the `init()` but I'm not sure why it has to init the Attribute Form Properties on cancellation at all. 

I'm sure it's not your fault but mine, since I did this two years ago. Tell me if I should investigate more or we can have a call, then I can show you the situation.

signedav (changes requested):
First: I do love it. Thanks a lot.

The code looks good. But I found the crash effected by old code.

signedav:
**Additional input (not part of the review, but fits in the PR):**
This Category Containers here 
https://github.com/qgis/QGIS/blob/d1126c3a126e93b3bc28436968e8bc79c5b4583f/src/app/qgsattributesformproperties.cpp#L101
do not need to be dragable. We could unset the flags for the items here. But possibly we need a new `DnDTreeItemData::Type`. StyleContainer e.g. 

Btw. this old code here:
https://github.com/qgis/QGIS/blob/d1126c3a126e93b3bc28436968e8bc79c5b4583f/src/app/qgsattributesformproperties.cpp#L926-L947
Does not make any sence, since it's in the if clause checking for containers. The icons should be implemented properly or this can be removed (code is my own fault as well, tell me if I should care about myself).

src/app/qgsattributesformproperties.cpp
```diff
@@ -1107,7 +1138,7 @@ void DnDTree::onItemDoubleClicked( QTreeWidgetItem *item, int column )
       }
 
       QgsCollapsibleGroupBox *styleGroupBox = new QgsCollapsibleGroupBox( tr( "Style" ), layout->widget() );
-      styleGroupBox->setObjectName( QLatin1Literal( "attributeFormPropertiesContainerStyle" ) );
+      styleGroupBox->setObjectName( QLatin1String( "attributeFormPropertiesContainerStyle" ) );
```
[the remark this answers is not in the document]
3nids: it's deprecated

src/app/qgsattributesformproperties.cpp
3nids: the crash should be fixed with last commit

3nids:
regarding your last comment, I have a second part coming which will handle this!

signedav (approved):

3nids:
tagged as feature for the changelog

DelazJ:
> tagged as feature for the changelog

@3nids that means that we should close https://github.com/qgis/QGIS-Documentation/issues/4890, right?

3nids:
@DelazJ yes, I created a small PR in the docs directly  https://github.com/qgis/QGIS-Documentation/pull/4912

PR #34417, opened 2020-02-12
Clear all use of Qt API deprecated up to 5.13

nyalldawson:


3nids (approved):
thanks a lot for the tedious work!

PR #34419, opened 2020-02-12
Fix labels don't render when mode is set to "From Symbol Bounds" and features are not set to be obstacles for other labels

nyalldawson:
A lovely corner case that cost me 20 mins of debugging :yum:


nirvn:
This amounts to 10 minute per changed line! :wink:

PR #34420, opened 2020-02-12
Fix parentheses in QgsExpression::dump()

m-kuhn:
Running `QgsExpression('NOT (1 or 2 or 3)').dump()` previously returned `NOT 1 or 2 or 3` -> now `NOT ( 1 or 2 or 3 )`.

This is especially painful since QGIS as WFS client translates `attribute NOT IN (1, 2, 3)` to an ogc filter of the form `NOT ( attribute = 1 or attribute = 2 or attribute = 3 )` which would be translated back to a faulty expression on the wfs server.


elpaso (approved):
backport?

PR #34433, opened 2020-02-12
drad'n'drop improvents part 2

### LINKED ISSUE #29063: A more intuitive way to open the group/tab properties in drag and drop form designer
Author Name: **Alexandre Heurion** (Alexandre Heurion)
Original Redmine Issue: [21245](https://issues.qgis.org/issues/21245)

Redmine category:forms

---
En utilisant un formulaire, nous voulons parfois qu'un groupe n'apparaisse que sous certaines conditions. Pour ça il existe une option "Configurer le Conteneur" qui s'ouvre en double-cliquant sur le sous-groupe. On peut alors y rentrer une expression contrôlant la visibilité du groupe (cf.image).
Cependant ce principe de double-clic n'est pas intuitif (comme un "clic droit > propriétés"). L'idéal serait (à mon avis) que les informations de cette boîte de dialogue apparaissent dans la colonne de droite (cf. image) comme informations de groupe.

Using the formularies, sometimes we wish a group only shows up with some conditions. There is an option for that (called in french "Configurer le Conteneur") as a dialog box that opens by double clickling on a group name. Then we can write some expression allowing the group to shows up or not (cf.image).
However this is not an intuitive way to use this option (not as much as a "right clic > properties" weel used with Windows). For me, it should be better if the options included in this box are displayed directly at the right of the screen (cf.image) like the informations of the field when we're selecting one.

---

- [formulaire Qgis champ optionnel 2.png](https://issues.qgis.org/attachments/download/14291/formulaire%20Qgis%20champ%20optionnel%202.png) (Alexandre Heurion)

3nids:
display the container editor in the right panel
this avoids opening a modal dialog

a new type of itemData has been added as WidgetType
to distinguish them in the available widgets from the containers in the form layout

fixes #29063

Screen-Recording-2020-02-12-at-14 19 55 (1)


src/app/attributeformconfig/qgsattributeformcontaineredit.h
```diff
@@ -0,0 +1,45 @@
...
+ */
+class APP_EXPORT QgsAttributeFormContainerEdit: public QWidget, private Ui_QgsAttributeFormContainerEdit
+{
+    Q_OBJECT
+
+  public:
+    explicit QgsAttributeFormContainerEdit( QTreeWidgetItem *item, QWidget *parent = nullptr );
+
+
+    void updateItemData();
```
signedav: Doesn't it need docs?

src/app/qgsattributesformproperties.cpp
```diff
@@ -941,6 +981,9 @@ QTreeWidgetItem *DnDTree::addItem( QTreeWidgetItem *parent, QgsAttributesFormPro
       case QgsAttributesFormProperties::DnDTreeItemData::HtmlWidget:
         //no icon for HtmlWidget
         break;
+
+      case QgsAttributesFormProperties::DnDTreeItemData::WidgetType:
```
signedav: This will never happen since this is all in the other if clause checking for container (old thing I mentioned in the last pr). Maybe you are keeping it for your upcoming implementations.

Anyway, I would check here for WidgetType and add this `newItem->setBackground( 0, QBrush( Qt::lightGray ) );`

signedav (approved):
So far so awesome.

I approve - see my comments.

src/app/qgsattributesformproperties.cpp
3nids: I will take care of the icons later.
But sadly the svg are missing.

src/app/attributeformconfig/qgsattributeformcontaineredit.h
3nids: It's in app, I added a short description for the class only

src/app/qgsattributesformproperties.cpp
3nids: sorry, the fix for the light gray is coming in next round

PR #34436, opened 2020-02-12
Fix spatialite crash on field added

### LINKED ISSUE #34375: QGIS crashes adding new field to geometryless Spatialite tbale
Reproduce bug by...

1. Start new QGIS project.
2. Add or create new Spatialite layer.
3. Open attribute table.
4. Toggle editing.
5. Add new field.  (NOT add new feature, but field.)
6. Upon clicking "OK", QGIS crashes.

Bug even present when table has no existing features/records.

QGIS version | 3.11.0-Master | QGIS code revision | a0723939fa
-- | -- | -- | --
PROJ.4 Version | 493
Active python plugins | latlontools; joinmultiplelines; qdraw; DataPlotly; mmqgis; DigitizingTools; openlayers_plugin; processing; db_manager

elpaso:
Fixes #34375

Add a check for the field index validity, but the issue was really
an undesired side effect of the spaghetti-signal-slot hell.

### DIFF
--- a/src/providers/spatialite/qgsspatialiteprovider.cpp
+++ b/src/providers/spatialite/qgsspatialiteprovider.cpp
@@ -1024,6 +1024,11 @@ QVariant QgsSpatiaLiteProvider::defaultValue( int fieldId ) const
 
 QString QgsSpatiaLiteProvider::defaultValueClause( int fieldIndex ) const
 {
+  if ( ! mAttributeFields.exists( fieldIndex ) )
+  {
+    return QString();
+  }
+
   if ( mAttributeFields.at( fieldIndex ).name() == mPrimaryKey && mPrimaryKeyAutoIncrement )
   {
     return tr( "Autogenerate" );


m-kuhn:
+1 for the PR
-1 for the not really helpful comment

elpaso:
oh, I guess a 0 it's not a bad score after all :/

m-kuhn:
Let's stick to the technical side in here and consider it a :+1:

m-kuhn (approved):

nyalldawson:
> -1 for the not really helpful comment

I understand where you're coming from, but lets be honest -- the attribute table/form HAS become a really complex interwoven network of signals built on top of signals. It's screaming out for a re-write at some future stage.

m-kuhn:
> I understand where you're coming from, but lets be honest -- the attribute table/form HAS become a really complex interwoven network of signals built on top of signals. It's screaming out for a re-write at some future stage.

In this particular case here I would say the reason is more that attribute indexes (instead of field names) are stored. This is ok on background threads where they are linked to a particular feature source but not on the main thread where they can change anytime. Even if the current implementation might not be the best of its breed, just dumping shit on it won't help anyone, serves as an excuse for writing bad code on top "because it's broken anyway" -- and possibly ignores that the topic attribute table is complex by nature if it needs to work properly for different backends, table sizes, latencies, ...
I think it would be worth stopping the sideswipes and work towards a better solution in a constructive way.

elpaso:
Ok, I apologize for the pointless comment. My feeling was a bit of frustration because I know that I'm not fixing the real issue here but I don''t have the time/resource/skills for a better solution right now.

m-kuhn:
The fix in here looks good to me as it possibly also prevents crashes in other areas / plugins.

nyalldawson:
Please understand that I'm not blaming anyone here! To me it's the exact same situation that composer was in pre - 3.0, in that the feature set has grown so far beyond the original intentions and design that it's naturally strayed into code complexity and workarounds. That's why I'd advocate for a rewrite, where the **current** feature set can be incorporated into the core design.

That said, this hypothetical rewrite would be a LOT of work and would be a very complex task in itself!

gioman:
> That said, this hypothetical rewrite would be a LOT of work and would be a very complex task in itself!

I think at some point this should really be raised at a PSC level to plan a roadmap, costs, etc.

m-kuhn:
It's the very same here, no offense intended. I was just trying to keep things professional and everything seems to be sorted on this end now :)

I agree on the rewrite. I have been talking to potential clients but didn't find anyone willing to invest into this so far (and I have 0 interest in doing this as volunteer work).

PR #34446, opened 2020-02-12
When showing original path in the "Select a Data Source" dialog allow the original path text to be selected and copied

nyalldawson:
... and make the original path a hyperlink to open the corresponding folder

Makes it much easier for users to find the correct fixed paths,
e.g. by allowing them to use OS level find files functionality


nirvn:
I was coming in to suggest using the find closest path function only to see you already did. +1 to merge, nice tweak.

PR #34448, opened 2020-02-13
[Optim] Only call data once in drawLayerTitleInternal

roya0045:
## Description

This is a small and petty optimisation but when operations done in data are expansive the multiple calls can stack up.

This simply aims to only call data once.


nyalldawson (approved):

roya0045:
@nyalldawson thanks Nyall!

PR #34460, opened 2020-02-13
GDAL: make sure we are not reading outside the raster

### LINKED ISSUE #34435: Raster calculator turns one row of pixels into nodata
Discussion with code examples and screenshots in https://gis.stackexchange.com/questions/349531/qgis-zonal-statistics-and-no-data/349559

Steps to re-produce with QGIS 3.11 Master on Windows 10.

1. Download https://s3.us-east-2.amazonaws.com/earthstatdata/FertilizerBalance_Geotiff.zip
2. Use file NitrogenBalanceOnLandscape_140Crops.tif (later "source")
3. Create a new GeoTIFF by applying the following raster calculator formula
`("source@1"<0) * "source@1"`
4. Open the result into QGIS. The southmost pixel row gets, in my environment, a nodata value of -3.40282347e+38. In the source image all those pixels are zeroes.
Clicked coordinate may be for example -27.239, -89.987

The issue happens for me when using the default, non-accelerated path. If I activate the OpenCL acceleration then the lowest pixel row is not changed into nodata.

elpaso:
Fixes #34435

Unfortunately I wasn't able to write a test for this
case: it seems to be related to a very rare combination
of floating point (double) rounding issues that I could
only reproduce manually.

But since I was trying to test it and I wrote some
raster block test cases, I thought it would be good
to leave them in the PR instead of throwing them
away.

The changes in the raster calculator are not strictly related 
but they make the GPU and CPU path behave the same as
far as nodata are concerned, as was pointed out in  
#34435, the behavior was different (the CPU path was the 
good one).

### DIFF
--- a/src/analysis/raster/qgsrastercalculator.cpp
+++ b/src/analysis/raster/qgsrastercalculator.cpp
@@ -124,11 +124,18 @@ QgsRasterCalculator::Result QgsRasterCalculator::processCalculation( QgsFeedback
     }
   }
 
+  // Check if we need to read the raster as a whole (which is memory inefficient
+  // and not interruptible by the user) by checking if any raster matrix nodes are
+  // in the expression
+  bool requiresMatrix = ! calcNode->findNodes( QgsRasterCalcNode::Type::tMatrix ).isEmpty();
+
 #ifdef HAVE_OPENCL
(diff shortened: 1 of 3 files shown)


nyalldawson:
Lgtm

PR #34469, opened 2020-02-14
Fix Task manager race conditions issues

troopa81:
## Description

This PR fixes differrent race condition issues is QgsTaskManager:

c380b09945  When a task is completed, statusChanged signal is emitted in task thread. As a consequence cleanupAndDeleteTask will be called in the main thread and the task will be deleted (later). It happens sometimes that instructions after statusChanged are executed after the task had been effectively deleted, leading to crash or non emitted signals.

It cause also issue with sub task. Indeed, when a subTask is completed, you have no garantie that instructions after statusChanged are executed before the parent caught the statusChanged signal and delete itself and the subtask.

This commit propose to force  `processSubTasksForCompletion` and `processSubTasksForTermination` execution in main thread (`processSubTasksForHold` if out of the scope because it doesn't delete tasks and is executed in main thread), so all status modification and deletion are made in a determinist order. 

a79431d5ad : remove unlock from `terminated`/`completed` because unlock call from `terminated` crashes the application when called from `cancel` method on not started task. In this case, the task is canceled without being started, so the mutex isn't locked. And the Qt documentation says

> Unlocking a mutex that is not locked results in undefined behavior.

5e4c1ea6a5 : cancel is called from a different thread than the one running the task so mShouldTerminate should be protected with a mutex

956b142239 : we need to be sure that the task is really completly finished (start method has exited) before deleting the task object to avoid crash

0ed9ea93f5 : I fix the tests, slip subTask in several one to ease debug (when it fails on CI for instance) and remove the SKIP tests on CI. I run all tests on linux and Windows, [repeatdly](https://github.com/troopa81/QGIS/commit/3a2382792c) and injecting latency in task manager to try to make it crash. So far, I have no more failing tests.

### DIFF
--- a/src/core/qgstaskmanager.cpp
+++ b/src/core/qgstaskmanager.cpp
@@ -91,7 +91,9 @@ void QgsTask::cancel()
   if ( mOverallStatus == Complete || mOverallStatus == Terminated )
     return;
 
+  mShouldTerminateMutex.lock();
   mShouldTerminate = true;
+  mShouldTerminateMutex.unlock();
   if ( mStatus == Queued || mStatus == OnHold )
   {
     // immediately terminate unstarted jobs
@@ -109,6 +111,12 @@ void QgsTask::cancel()
   {
     subTask.task->cancel();
   }
+}
+
+bool QgsTask::isCanceled() const
+{
+  QMutexLocker locker( &mShouldTerminateMutex );
+  return mShouldTerminate;
 }
 
 void QgsTask::hold()
--- a/src/core/qgstaskmanager.cpp
+++ b/src/core/qgstaskmanager.cpp
@@ -264,7 +264,7 @@ void QgsTask::setProgress( double progress )
 void QgsTask::completed()
 {
   mStatus = Complete;
-  QMetaObject::invokeMethod( this, &QgsTask::processSubTasksForCompletion, Qt::AutoConnection );
+  QMetaObject::invokeMethod( this, "processSubTasksForCompletion" );
 }
 
 void QgsTask::processSubTasksForCompletion()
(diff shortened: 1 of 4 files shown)


src/core/qgstaskmanager.cpp
```diff
@@ -254,7 +264,7 @@ void QgsTask::setProgress( double progress )
 void QgsTask::completed()
 {
   mStatus = Complete;
-  processSubTasksForCompletion();
+  QMetaObject::invokeMethod( this, &QgsTask::processSubTasksForCompletion, Qt::AutoConnection );
```
elpaso: If I'm not wrong, AutoConnection is the default anyway.

elpaso:
Nice work!

src/core/qgstaskmanager.cpp
troopa81: Right! and the call with a method pointer doesn't exist yet in Qt 5.9 (failing the CI). I changed it.

troopa81:
> Nice work!

Thanks

rouault:
@troopa81 I think for automatic backport to work, I need to squash when merging ? Is it OK for you ?

m-kuhn:
@rouault it should also work with merge commits meanwhile. (https://github.com/Gaurav0/backport/pull/41)

rouault:
OK, I took the "Merge commit" road

3nids:
sadly, the bot was broken and is back to former version which doesn't work for merge commits.

rouault:
@troopa81 Please issue a manual PR for the 3.12 backport

nyalldawson:
@troopa81 there's a test intermittently failing here still: see https://travis-ci.org/github/qgis/QGIS/jobs/662189227?utm_medium=notification&utm_source=github_status

troopa81:
@nyalldawson  I'll take a look

troopa81:
@nyalldawson fixed #35108

PR #34500, opened 2020-02-17
Recover pointcloud tests

strk:
Removed with https://github.com/qgis/QGIS/pull/34495 for unstable
results, improved here to build a well-known version of pointcloud.

### DIFF
--- a/tests/testdata/Dockerfile-postgis
+++ b/tests/testdata/Dockerfile-postgis
@@ -10,12 +10,12 @@ RUN chmod 400 /etc/ssl/private/postgres_key.key
 # NOTE: release 1.2.0 would not build against PostgreSQL-11:
 # https://github.com/pgpointcloud/pointcloud/issues/248
 RUN apt-get -y update; apt-get -y install build-essential autoconf postgresql-server-dev-11 libxml2-dev zlib1g-dev
+RUN dpkg -L postgresql-server-dev-11
 RUN wget -O- \
   https://github.com/pgpointcloud/pointcloud/archive/v1.2.1.tar.gz \
   | tar xz && \
   cd pointcloud-1.2.1 && \
-  PATH=/usr/lib/postgresql/11/bin/:$PATH && \
   ./autogen.sh && \
-  ./configure && \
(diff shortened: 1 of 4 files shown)


strk:
Failure we want to address:
https://travis-ci.org/qgis/QGIS/jobs/650370964#L837

strk:
Error is:
> configure: error: the PGXS Makefile /usr/lib/postgresql/12/lib/pgxs/src/makefiles/pgxs.mk cannot be found. Please install the PostgreSQL server development packages and re-run configure.

We're installing postgresql-server-dev-11 as part of the Dockerfile instructions, so I dunno why 12 is being found, I guess he kartoza Docker is _also_ having 12 installed, so pg_config points there, and we just need to tweak the PATH variable to make sure to use 11 instead.

tests/testdata/Dockerfile-postgis
```diff
@@ -5,3 +5,16 @@ ADD auth_system/certs_keys/postgres.key /etc/ssl/private/postgres_key.key
 ADD auth_system/certs_keys/issuer_ca_cert.pem /etc/ssl/certs/issuer_ca_cert.pem
 
 RUN chmod 400 /etc/ssl/private/postgres_key.key
+
+# Compile and install PointCloud.
+# NOTE: release 1.2.0 would not build against PostgreSQL-11:
```
pblottiere: The last 1.2.1 release  should compile fine though.

strk: Trying 1.2.1 next...

strk:
Packaging issue, looks like: PostgreSQL 11 server dev package not including pg_config !
Any reason why our test uses 11 instead of 10 or 12 ?

strk:
@pblottiere if you find the time to fix https://github.com/pgpointcloud/pointcloud/issues/256 we'd be able to install pointcloud from a package instead of building it ...

pblottiere:
> @pblottiere if you find the time to fix pgpointcloud/pointcloud#256 we'd be able to install pointcloud from a package instead of building it ...

Yes, it's on my todo list. I'll probably have time to investigate by the end of this month.

PR #34520, opened 2020-02-18
[ui] Fix diagram attribute colors missing on non-default theme in the diagram properties dialog

nirvn:
## Description

Hadn't stumbled on a non-default theme issue. Our diagram properties dialog was using setBackground() on a tree view widget to indicate the color for diagram attributes. This is a big no-no as the color is _not_ visible for non-default themes. 

Before vs. fixed:
image

Bonuses:
- the color column is set to be as narrow as possible
- the UI now matches other tree widgets having color columns (color ramp shader widget, raster renderer's unique value UI, etc.)


nyalldawson (approved):

nirvn:
Travis failure unrelated.

PR #34521, opened 2020-02-18
[georeferencer] Fix wrong target CRS using GDAL3/PROJ6 builds

nirvn:
## Description

Under GDAL3/PROJ6 builds, passing the CRS to GDAL as a Proj4 string results in the wrong projection being picked. The fix is simple, pass the projection as WKT, a superior format to begin with.

### DIFF
--- a/src/plugins/georeferencer/qgsimagewarper.cpp
+++ b/src/plugins/georeferencer/qgsimagewarper.cpp
@@ -97,7 +97,11 @@ bool QgsImageWarper::createDestinationDataset( const QString &outputName, GDALDa
   if ( crs.isValid() )
   {
     OGRSpatialReference oTargetSRS;
+#if PROJ_VERSION_MAJOR>=6
+    oTargetSRS.importFromWkt( crs.toWkt( QgsCoordinateReferenceSystem::WKT2_2018 ).toLatin1().data() );
+#else
     oTargetSRS.importFromProj4( crs.toProj().toLatin1().data() );
+#endif
 
     char *wkt = nullptr;
     OGRErr err = oTargetSRS.exportToWkt( &wkt );


src/plugins/georeferencer/qgsimagewarper.cpp
```diff
@@ -97,7 +97,12 @@ bool QgsImageWarper::createDestinationDataset( const QString &outputName, GDALDa
   if ( crs.isValid() )
   {
     OGRSpatialReference oTargetSRS;
+    //oTargetSRS.importFromProj4( crs.toProj().toLatin1().data() );
+#if PROJ_VERSION_MAJOR>=6
+    oTargetSRS.importFromWkt( crs.toWkt().toLatin1().data() );
```
nyalldawson: Better export in wkt 2018 format so it's not lossy

src/plugins/georeferencer/qgsimagewarper.cpp
```diff
@@ -97,7 +97,12 @@ bool QgsImageWarper::createDestinationDataset( const QString &outputName, GDALDa
   if ( crs.isValid() )
   {
     OGRSpatialReference oTargetSRS;
+    //oTargetSRS.importFromProj4( crs.toProj().toLatin1().data() );
```
nyalldawson: Delete

nyalldawson (commented):
Is importFromProj4 used elsewhere? I thought I'd caught all of these...

src/plugins/georeferencer/qgsimagewarper.cpp
nirvn: God, could I have been more sloppy before pushing? I don't think so...

src/plugins/georeferencer/qgsimagewarper.cpp
nirvn: Done.

rouault: toUtf8() would likely be more appropriate. proj.db content is encoded in UTF-8

nirvn: @rouault , roger that.

PR #34522, opened 2020-02-18
Fix hide/show button in attribute form configuration for relations

obrix:
Should fix a bug where ticking the show label button in the attribute form configuration for a relation had no effect. Now it properly hide/show the labels in the attribute form for relation editor widget. Also propagate the show label configuration if a field widget is actually a QgsRelationReferenceWidget.

- In QgsAttributeForm::createWidgetFromDef, move setShowLabel, setShowLinkButton and setShowUnlinkButton after the instantiation of QgsAttributeFormRelationEditorWidget.
Setting those parameters before had no effect as QgsRelationEditorWidget was not yet created.

- Add a missing test of showLabel parameter in Qgsrelationeditorwidget.

- Check if field widget is actually a QgsRelationReferenceWidget by casting the widget. Show / hide the title with the show label property from the attribute form configuration.


src/gui/qgsattributeform.cpp
```diff
@@ -1824,6 +1826,14 @@ QgsAttributeForm::WidgetInfo QgsAttributeForm::createWidgetFromDef( const QgsAtt
 
         QgsEditorWidgetWrapper *eww = QgsGui::editorWidgetRegistry()->create( widgetSetup.type(), mLayer, fldIdx, widgetSetup.config(), nullptr, this, mContext );
         QgsAttributeFormEditorWidget *formWidget = new QgsAttributeFormEditorWidget( eww, widgetSetup.type(), this );
+
+        QgsRelationReferenceWidget *relRefWidget = dynamic_cast<QgsRelationReferenceWidget *>( eww->widget() );
+        if ( relRefWidget )
+        {
+          //This field widget can be a frame for a reference relation, show or hide its label accordingly
+          relRefWidget->setShowLabel( widgetDef->showLabel() );
```
troopa81: I'm not sure this is a good idea to have the same parameter to configure the visibility of two different things:
- the label in the form left column
- the groupbox label in the relation reference embedded form 

Maybe it would be better to add a specific configuration parameter

obrix: I agree. We can potentially keep the first commit from this PR which actually fix a bug. But to show/hide the groupbox label which is the title of the QgsRelationReferenceWidget we could add a new check box element in the attribute editor configuration view. That would be a new feature though. Is it okay for you ?

troopa81: > Is it okay for you ?
Yes

troopa81 (approved):
Seems ok for merging with the last modifications

obrix:
@elpaso It seems you fixed directly this issue by resolving issue #34651 . But It seems your forgot to also move setShowLabel( relDef->showLabel() ) after the instantation of QgsAttributeFormRelationEditorWidget ?

elpaso:
@obrix totally possible, sorry but I have no time to look at that now, if you feel confident please fix it.

Issue #34362, opened 2020-02-07
QGIS DB Manager fails to escape quotes in comment field when importing a shapefile into PostGIS


Nitrodist:
**Describe the bug**

User input is not escaped when using the 'comment' option in using QGIS's DB Manager to import a shapefile to a postgis instance. Having a quote in the optional comment value causes the import to crash/fail:

```
An error has occurred while executing Python code: 

db_manager.db_plugins.plugin.DbError: syntax error at or near "s" LINE 1: ...ourhoods" IS E'Imported from the City of Winnipeg's ODP http...                                                             ^ 
Traceback (most recent call last):
  File "/Applications/QGIS3.10.app/Contents/MacOS/../Resources/python/plugins/db_manager/db_plugins/connector.py", line 89, in _execute
    cursor.execute(sql)
psycopg2.errors.SyntaxError: syntax error at or near "s"
LINE 1: ...ourhoods" IS E'Imported from the City of Winnipeg's ODP http...
                                                             ^


During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Applications/QGIS3.10.app/Contents/MacOS/../Resources/python/plugins/db_manager/dlg_import_vector.py", line 382, in accept
    self.db.connector.commentTable(schema, table, com)
  File "/Applications/QGIS3.10.app/Contents/MacOS/../Resources/python/plugins/db_manager/db_plugins/postgis/connector.py", line 756, in commentTable
    self._execute(None, 'COMMENT ON TABLE "{0}"."{1}" IS E\'{2}\';'.format(schema, tablename, comment))
  File "/Applications/QGIS3.10.app/Contents/MacOS/../Resources/python/plugins/db_manager/db_plugins/connector.py", line 97, in _execute
    raise DbError(e, sql)
db_manager.db_plugins.plugin.DbError: syntax error at or near "s"
LINE 1: ...ourhoods" IS E'Imported from the City of Winnipeg's ODP http...
                                                             ^



Python version: 3.7.3 (default, Mar 27 2019, 09:23:15) [Clang 10.0.1 (clang-1001.0.46.3)] 
QGIS version: 3.10.2-A Coruña A Coruña, 616ad4531b 

Python Path:
/Applications/QGIS3.10.app/Contents/MacOS/../Resources/python
/Users/markcampbell/Library/Application Support/QGIS/QGIS3/profiles/default/python
[...]

PR #34525, opened 2020-02-18
Correct bug from postgres comment using quote or caps

Ailurupoda:
## Description
Add the possibility to comment a postgreSQL table using  quote and other special characters.

#34362

### DIFF
--- a/python/plugins/db_manager/db_plugins/postgis/connector.py
+++ b/python/plugins/db_manager/db_plugins/postgis/connector.py
@@ -878,7 +878,7 @@ class PostGisDBConnector(DBConnector):
         if comment is None:
             self._execute(None, 'COMMENT ON TABLE "{0}"."{1}" IS NULL;'.format(schema, tablename))
         else:
-            self._execute(None, 'COMMENT ON TABLE "{0}"."{1}" IS E\'{2}\';'.format(schema, tablename, comment))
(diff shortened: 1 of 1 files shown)


python/plugins/db_manager/db_plugins/postgis/connector.py
```diff
@@ -878,7 +878,7 @@ def commentTable(self, schema, tablename, comment=None):
         if comment is None:
             self._execute(None, 'COMMENT ON TABLE "{0}"."{1}" IS NULL;'.format(schema, tablename))
         else:
-            self._execute(None, 'COMMENT ON TABLE "{0}"."{1}" IS E\'{2}\';'.format(schema, tablename, comment))
+            self._execute(None, 'COMMENT ON TABLE "{0}"."{1}" IS $${2}$$;'.format(schema, tablename, comment))
```
nyalldawson:
```suggestion
            self._execute(None, 'COMMENT ON TABLE "{0}"."{1}" IS $escape${2}$escape$;'.format(schema, tablename, comment))
```

Just to play it super-safe?

Ailurupoda: Seems good to me. It's at least cleaner :)

Ailurupoda:
How to solve the conflicts with qgis.QGIS Azure Pipelines ? i dont' really understand how this works ..

nyalldawson:
@Ailurupoda it was an unrelated test failure

PR #34532, opened 2020-02-18
[FEATURE] mesh simplification

### LINKED ISSUE #31575: Adaptive derived mesh
Render speed enhancements by using resampling at different zoom resolutions –
create “overview” for grids

Right now the derived triangles are calculated once and they are kept for all zoom levels and canvas extents. This is sub-optimal, since the the number of rendered elements should be dependent on zoom level (it is not useful to have multiple triangles for one rendered pixel or on the other hand one triangle for whole canvas). 

Current implementation also assumes that the native/original vertices are always used in the derived mesh. This assumption leads into various bugs with reprojection.

related: https://github.com/qgis/QGIS/issues/29330

related: https://github.com/qgis/QGIS/issues/28827

vcloarec:
As mesh layer could have millions of triangles, rendering could be very slow, especially when all the triangles are displayed in the view whereas triangles are too small to be viewed. 

For those situations, this PR provides an option to simplify the mesh. Simplification leads to one or more simplified mesh that represents levels of detail. When rendering the mesh, the appropriate level of detail is chosen to have an adequate rendering depending on the view.

A new tab in the mesh layer properties widget allows the user to change the settings :
- The reduction factor is used to simplify mesh, each level of details has approximately a number of triangles equal the number of triangles of the previous level of details divided by the reduction factor
- Maximum mesh resolution: this is the average size (in pixels) of the triangles that is permitted to display, if this average size of the mesh is lesser than this value, a lower level of details mesh is displayed, which will have a average size just greater than the maximum mesh resolution.

meshSimplificationWidget

This feature permits to speed up rendering as in those tables (result in milliseconds) :
meshSimplicationBenchmark

Demo GIF
Before:
meshSimplificationbefore

After:
meshSimplificationAfter


PeterPetrik:
fixes #31575

PeterPetrik (approved):
already reviewed in the PR within vcloarec fork https://github.com/vcloarec/QGIS/pull/4

external/meshOptimizer/LICENSE.md
```diff
@@ -0,0 +1,21 @@
+MIT License
```
nyalldawson: Can you add external/meshOptimizer to the exceptions paths in scripts/astyle.sh?

src/analysis/mesh/qgsmeshcalcutils.cpp
```diff
@@ -699,20 +699,11 @@ void QgsMeshCalcUtils::updateMesh() const
...
-    // some dummy triangular mesh.
-    QgsMapSettings mapSettings;
-    mapSettings.setExtent( mMeshLayer->extent() );
-    mapSettings.setDestinationCrs( mMeshLayer->crs() );
-    mapSettings.setOutputDpi( 96 );
-    QgsRenderContext context = QgsRenderContext::fromMapSettings( mapSettings );
-
-    delete mMeshLayer->createMapRenderer( context );
+    QgsCoordinateTransform transform;
+    transform.setDestinationCrs( mMeshLayer->crs() );
```
nyalldawson: You **need** to use the correct transform context here

src/analysis/mesh/qgsmeshcontours.cpp
```diff
@@ -43,13 +43,11 @@ QgsMeshContours::QgsMeshContours( QgsMeshLayer *layer )
...
-  QgsMapSettings mapSettings;
-  mapSettings.setExtent( mMeshLayer->extent() );
-  mapSettings.setDestinationCrs( mMeshLayer->crs() );
-  mapSettings.setOutputDpi( 96 );
-  QgsRenderContext context = QgsRenderContext::fromMapSettings( mapSettings );
-  mTriangularMesh.reset( new QgsTriangularMesh() );
-  mTriangularMesh->update( mNativeMesh.get(), &context );
+  QgsCoordinateTransform transform;
+  transform.setDestinationCrs( mMeshLayer->crs() );
+  transform.setSourceCrs( mMeshLayer->crs() );
```
nyalldawson: Same here -- you need to use the transform context

src/app/mesh/qgsmeshlayerproperties.cpp
```diff
@@ -146,6 +146,13 @@ void QgsMeshLayerProperties::syncToLayer()
 #ifdef HAVE_3D
   mVector3DWidget->setLayer( mMeshLayer );
 #endif
+
+  QgsDebugMsg( QStringLiteral( "populate rendering tab" ) );
```
nyalldawson: Remove, or use QgsDebugMsgLevel at a lower level

src/app/mesh/qgsmeshlayerproperties.cpp
```diff
@@ -204,6 +211,22 @@ void QgsMeshLayerProperties::apply()
   mVector3DWidget->apply();
 #endif
 
+  QgsDebugMsg( QStringLiteral( "processing rendering tab" ) );
```
nyalldawson: Same

src/core/mesh/qgsmeshlayer.cpp
```diff
@@ -73,6 +73,27 @@ void QgsMeshLayer::setDefaultRendererSettings()
...
+    for ( int i = 0; i < simplifyMeshes.count() ; ++i )
+    {
+      mTriangularMeshes.emplace_back( simplifyMeshes[i] );
+    }
+  }
+}
+
+bool QgsMeshLayer::hasSimplifiedMeshes() const
+{
+  return ( mTriangularMeshes.size() > 1 );
```
nyalldawson: SHould this be > 0? If it's correct, a comment would be useful to explain why

src/core/mesh/qgsmeshlayer.h
```diff
@@ -182,18 +183,24 @@ class CORE_EXPORT QgsMeshLayer : public QgsMapLayer
...
+     * and the simplified triangular meshes, the one returned is which has the average triangle size just greater than triangleSize.
+     * The size of a triangle is the maximum between the height and the width of the triangle bounding box
+     * For default parameter (=0), it returns base triangular mesh.
+     * \param minimumTriangleSize is the average size criteria in canvas map units
+     * \returns triangular mesh
+     * \note triangular size added in QGIS 3.14
      * \note Not available in Python bindings
      */
-    QgsTriangularMesh *triangularMesh() SIP_SKIP;
+    QgsTriangularMesh *triangularMesh( double minimumTriangleSize = 0 ) const SIP_SKIP;
```
nyalldawson: What happens with owernship here? Is it a factory?

src/core/mesh/qgsmeshsimplificationsettings.h
```diff
@@ -0,0 +1,80 @@
...
+
+#include "qgis_core.h"
+#include "qgis.h"
+
+/**
+ * \ingroup core
+ *
+ * Represents a overview renderer settings
+ *
+ * \note The API is considered EXPERIMENTAL and can be changed without a notice
```
nyalldawson: If this is the case, you could consider hiding it from Python for now?

src/core/mesh/qgsmeshsimplificationsettings.h
```diff
@@ -0,0 +1,80 @@
...
+     * This reduction factor is used for simplification of each successive simplified mesh. For  example, if the base mesh has 5M faces,
+     * and the reduction factor is 10, the first simplified mesh will have approximativly 500 000 faces, the second 50 000 faces,
+     * the third 5000, ...
+     * If highter reduction factor leads to simpler meshes, it produces also fewer levels of detail.
+     * The reduction factor has to be strictly greater than 1. If not, the simplification processus will render nothing.
+     */
+    void setReductionFactor( double value );
+
+    //! Writes configuration to a new DOM element
+    QDomElement writeXml( QDomDocument &doc ) const;
```
nyalldawson: Add QgsReadWriteContext arguments to this and readXml, just in case we need them in future

src/core/mesh/qgstriangularmesh.h
```diff
@@ -118,6 +120,41 @@ class CORE_EXPORT QgsTriangularMesh
...
+     * the prededent goal of number of triangles.
+     * There are as many simplified meshes as necessary to have a the minimum triangles count on the last simplified mesh.
+     *
+     * The caller has to take the ownership of returned meshes.
+     *
+     * \param reductionFactor is the factor used to reduce the number of triangles of the mesh
+     * \param minimumTrianglesCount is the minimal faces count on simplified mesh
+     * \since QGIS 3.14
+     */
+    QVector<QgsTriangularMesh *> simplifyMesh( double reductionFactor, int minimumTrianglesCount = 10 ) const;
```
nyalldawson: SIP_FACTORY ?

nyalldawson (changes requested):

src/core/mesh/qgsmeshlayer.cpp
vcloarec: you're right...

src/core/mesh/qgstriangularmesh.h
vcloarec: hmmm, this file is SIP_NO_FILE

src/core/mesh/qgsmeshlayer.h
vcloarec: keep ownership, added in the doc

nyalldawson: thanks!

src/core/mesh/qgsmeshlayer.cpp
PeterPetrik: isn't mTriangularMeshes[0] base triangular mesh? so if mTriangularMeshes == 0, it has not base trianguler mesh, nor simplified meshed, if mTriangularMeshes == 1, it has base trinagular mesh, and if mTriangularMeshes > 1 it has simplified meshes & base mesh

vcloarec: @PeterPetrik exactly, I was a bit confusing last evening, due to a long day... As unit test made me back to the good road, only a comment is added...

PeterPetrik: testing, such an underestimated power

src/core/mesh/qgsmeshsimplificationsettings.h
vcloarec: yes no need now in Python, so I add SIP_NO_FILE

src/app/mesh/qgsmeshlayerproperties.cpp
vcloarec: done

src/analysis/mesh/qgsmeshcalcutils.cpp
```diff
@@ -699,20 +700,9 @@ void QgsMeshCalcUtils::updateMesh() const
...
-    // of the native mesh. So create
-    // some dummy triangular mesh.
-    QgsMapSettings mapSettings;
-    mapSettings.setExtent( mMeshLayer->extent() );
-    mapSettings.setDestinationCrs( mMeshLayer->crs() );
-    mapSettings.setOutputDpi( 96 );
-    QgsRenderContext context = QgsRenderContext::fromMapSettings( mapSettings );
-
-    delete mMeshLayer->createMapRenderer( context );
+    QgsCoordinateTransform transform( mMeshLayer->crs(), mMeshLayer->crs(), QgsProject::instance() );
```
nyalldawson: You need to avoid the QgsProject::instance() call here -- that will cause all sorts of issues in processing and elsewhere. You'll need to instead make the QgsTransformContext an argument for updateMesh() (or in the constructor for QgsMeshCalcUtils )

src/analysis/mesh/qgsmeshcontours.cpp
```diff
@@ -43,13 +44,10 @@ QgsMeshContours::QgsMeshContours( QgsMeshLayer *layer )
...
 
-  QgsMapSettings mapSettings;
-  mapSettings.setExtent( mMeshLayer->extent() );
-  mapSettings.setDestinationCrs( mMeshLayer->crs() );
-  mapSettings.setOutputDpi( 96 );
-  QgsRenderContext context = QgsRenderContext::fromMapSettings( mapSettings );
-  mTriangularMesh.reset( new QgsTriangularMesh() );
-  mTriangularMesh->update( mNativeMesh.get(), &context );
+
+  QgsCoordinateTransform transform( mMeshLayer->crs(), mMeshLayer->crs(), QgsProject::instance() );
```
nyalldawson: same here -- gotta make the transform context an explicit argument in the constructor

src/analysis/mesh/qgsmeshcalcutils.cpp
vcloarec: I am asking my self, here, not sure there is a real context for transform. More over, there is not a real transform as the transform is from layer crs to layer crs. It is just a workaround to create a triangular mesh in the layer crs taht will be used to mesh calculator operation. The CRS will always be the layer CRS.
Can we just used a "empty" QgsTransformContext obtained by default constructor ?

PeterPetrik: in that case maybe you can create a function that does not take transformation at all (or as default transformation is invalid) and it will not reproject vertices when creating the triangular mesh. And here in the code you will not pass there any context ,so it would be obvious what is going on here.

vcloarec: @PeterPetrik , very good idea, so no need the transform here and no workaround anymore!

src/core/mesh/qgsmeshlayer.h
```diff
@@ -198,9 +198,10 @@ class CORE_EXPORT QgsMeshLayer : public QgsMapLayer
 
     /**
      * Update the base triangular mesh, create one if it doesn't exist
+     * \param transform Transforamtion from layer CRS to destination (e.g. map) CRS. With invalid tranform, it keeps the native mesh CRS
```
PeterPetrik: typo Transforamtion

src/core/mesh/qgstriangularmesh.h
```diff
@@ -53,13 +53,12 @@ class CORE_EXPORT QgsTriangularMesh
     ~QgsTriangularMesh();
 
     /**
-     * If update is needed (not yet constructed, native mesh changed or CRS changed),
-     * constructs triangular mesh from layer's native mesh and context. Populates spatial index.
+     * Constructs triangular mesh from layer's native mesh and transform to destination CRS. Populates spatial index.
      * \param nativeMesh QgsMesh to access native vertices and faces
-     * \param transform coordinate transform from layer CRS
+     * \param transform Transforamtion from layer CRS to destination (e.g. map) CRS. With invalid tranform, it keeps the native mesh CRS
```
PeterPetrik: Transforamtion

src/core/mesh/qgstriangularmesh.h
```diff
@@ -53,13 +53,12 @@ class CORE_EXPORT QgsTriangularMesh
...
      * \returns true if the mesh is effectivly updated, and false if not
     */
-    bool update( QgsMesh *nativeMesh, const QgsCoordinateTransform &transform );
+    bool update( QgsMesh *nativeMesh, const QgsCoordinateTransform &transform = QgsCoordinateTransform() );
```
PeterPetrik: you probably need to add check when `!transform.isValid()` to the logic when to update the code?

vcloarec: done

PeterPetrik:
@nyalldawson i would like to merge this (you still have "requesting changes flag"). Do you have any other comment?

PR #34260, opened 2020-02-04
Backport #33683 oracle identity field management

SebastienPeillet:
## Description

Backport of #33683 `Oracle field generated always as identity`

Fix #33681

This PR has been forgotten by @qgib because it was disable when the merge appended.

PR #34546, opened 2020-02-19
BUG: fix invalid layer result when Oracle version is under 12

SebastienPeillet:
## Description

Relative to this [thread](
http://osgeo-org.1560.x6.nabble.com/Editing-Oracle-table-can-t-add-feature-td5423980.html)

Since the identity type is handled by oracle providers (cf. #33683), layers are considered as unvalid for users who use Oracle version under 12. Indeed *all_tab_identity_cols* table doesn't exist before 12 version. In consequence the sql request for identity columns fails.

This PR fix this problem.

### DIFF
--- a/src/providers/oracle/qgsoracleprovider.cpp
+++ b/src/providers/oracle/qgsoracleprovider.cpp
@@ -1028,28 +1028,43 @@ bool QgsOracleProvider::determineAlwaysGeneratedKeys()
     return false;
   }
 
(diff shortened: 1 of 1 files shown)


nyalldawson:
Does this need backporting?

SebastienPeillet:
In my opinion yes, but I didn't mention it because the previous PR about identity type is still waiting for backporting (#34260).

PR #34568, opened 2020-02-20
Add algorithms for raising warnings and exceptions from models

nyalldawson:
These algorithms raise either a custom warning in the processing log, OR raise
an exception which causes the model execution to terminate.

An optional condition expression can be specified to control whether or not
the warning/exception is raised, allowing logic like "if the output layer from
another algorithm contains more then 10 features, then abort the model execution"

Sponsored by Fisel + König


alexbruy (approved):

nyalldawson:
Thanks @alexbruy !

PR #34576, opened 2020-02-20
Fix unicode character errors when fetching WKT string from GDAL

nirvn:
## Description

Without this fix, wkt string passed on by GDAL's OSRExportToWktEx function have unicode character errors. For e.g.:
image

### DIFF
--- a/src/core/qgsogrutils.cpp
+++ b/src/core/qgsogrutils.cpp
@@ -736,7 +736,7 @@ QString QgsOgrUtils::OGRSpatialReferenceToWkt( OGRSpatialReferenceH srs )
   OSRExportToWkt( srs, &pszWkt );
 #endif
 
-  const QString res( pszWkt );
+  const QString res = QString::fromLatin1( pszWkt );
   CPLFree( pszWkt );
   return res;
 }


lbartoletti (approved):

nirvn:
@rouault , looks good to you?

nyalldawson:
Ping @rouault

rouault:
I don't think this is appropriate. Unfortunately WKT strings might be in Latin1 or UTF-8, so that should be conditionialized by some detection of the encoding

(note: I'm sorry I just noticed that one by chance. I'm unable to make my email @rouault filter works apparently probably because I've another filter for general qgis emails ...)

rouault:
Thinking that instead of QGIS messing with that, this should perhaps be better fixed in PROJ itself. On ingestion of WKT, PROJ would try to detect if its UTF-8 or Latin1. If the later, then it would recode to UTF-8 (which would guarantee than its output is UTF-8). PROJ 7 has 'imported' a few recoding routines from GDAL in src/filemanager.cpp, that could probably be used for that purpose.

rouault:
Hum, wait... In the screeshot, it seems the WKT comes from the SRS of a TIFF file, but TIFF don't contain WKT. The WKT is synthetized by GDAL / PROJ. How come do you get Latin1 ? PROJ >= 6 database only contains UTF-8

nirvn:
@rouault , you can check the file here: https://www.dropbox.com/s/dvleccq5ico94wz/cambodia-indian1960.tif?dl=0

nyalldawson:
@nirvn did you create that file?

nirvn:
@nyalldawson , yes, a long time ago in a far far galaxy (in mid 2017, so early QGIS 3, ~GDAL 2.2).

rouault:
The file is fine. There's no WKT in it as expected. "gdalsrsinfo cambodia-indian1960.tif.part > /tmp/out.txt" + "file /tmp/out.txt" reports "UTF-8 Unicode text" as expected. Isn't the issue that the HTML report isn't ready for UTF-8 characters ? Are you sure that the non-ASCII characters are really Latin1 ones ?

nirvn:
@rouault , I'll check again tonight but I'm pretty sure outputting the wkt string via the QGIS python console, ie iface.activeLayer().crs().toWkt(), also results in a malformed string. I remember testing this before filing the issue.

nirvn:
@rouault , just checked, my recollection was correct, the string is malformed when called from the python console too:
Screenshot from 2020-02-28 16-41-06

nyalldawson:
So what's the verdict here? PROJ issue? GDAL issue? QGIS issue?

rouault:
I feel more investigation should be done to better understand the root cause of the issue

PR #34581, opened 2020-02-20
Small rework on QgsFeatureFilterModel

### LINKED ISSUE #34580: Qgis crash when opening an attribute table on a layer
**Describe the bug**
Qgis hang and crash when opening an attribute table.

**How to Reproduce**
Here is the bug on video : [Bug Replay](https://gfycat.com/fr/yellowishcomplexdodobird)

And the stack trace : 
```
Crash ID: cc8ade6cd3fcfe46835320fdbb4c0bed32b59115


Stack Trace


QVector::QVector qvector.h:358
QgsFieldExpressionValuesGatherer::entries qgsfeaturefiltermodel_p.h:89
QgsFeatureFilterModel::updateCompleter qgsfeaturefiltermodel.cpp:228
QMetaCallEvent::placeMetaCall :
QObject::event :
QApplication::notify :
QgsApplication::notify qgsapplication.cpp:418
QCoreApplication::notifyInternal2 :
QCoreApplicationPrivate::sendPostedEvents :
qt_plugin_query_metadata :
QEventDispatcherWin32::processEvents :
UserCallWinProcCheckWow :
DispatchMessageWorker :
QEventDispatcherWin32::processEvents :
qt_plugin_query_metadata :
QEventLoop::exec :
QCoreApplication::exec :
main main.cpp:1598
BaseThreadInitThunk :
RtlUserThreadStart :




QGIS Info
QGIS Version: 3.10.2-A Coru�a
QGIS code revision: d4cd3cfe5a
Compiled against Qt: 5.11.2
Running against Qt: 5.11.2
Compiled against GDAL: 3.0.3
Running against GDAL: 3.0.3



System Info
CPU Type: x86_64
Kernel Type: winnt
Kernel Version: 10.0.18362
```

**QGIS and OS versions**
Version QGIS : 3.10.2-2
Windows 10

**Additional context**

I cannot reproduce this bug on Ubuntu 19.10 either in debug or release mode. This bug does not always happen on windows and sometime several opening closing of the attributes table are necessary to reproduce it.

I investigated the code and might have narrowed the issue on a specific part of QgsFeatureFilterModel::updateCompleter method. See pull request #34581 for a possible fix.

Ping @kandre as he is the one that found this bug first.

obrix:
Add a check for nullptr on mGatherer in QgsFeatureFilterModel::updateCompleter.

Also avoid relying on signal to slot call order conservation to make sure that gathererThreadFinished (that delete mGatherer)  is called after updateCompleter :
  - gathererThreadFinished is disconnected from finished signal emitted at the end of QgsFieldExpressionValuesGatherer::run
  - gathererThreadFinished is instead called explicitly at the end of updateCompleter
  - when QgsFieldExpressionValuesGatherer::run is stopped manually, rely on the connection to QgsFieldExpressionValuesGatherer::deleteLater to clean mGatherer.

This should fix an observed bug where a crash happened at the begining of updateCompleter because mGatherer was null.

Could fix #34580

### DIFF
--- a/src/core/qgsfeaturefiltermodel.cpp
+++ b/src/core/qgsfeaturefiltermodel.cpp
@@ -225,6 +225,12 @@ QVariant QgsFeatureFilterModel::data( const QModelIndex &index, int role ) const
 void QgsFeatureFilterModel::updateCompleter()
 {
   emit beginUpdate();
+  if( !mGatherer )
+  {
+    emit endUpdate();
+    return;
+  }
+
   QVector<Entry> entries = mGatherer->entries();
 
   if ( mExtraIdentifierValueIndex == -1 )
@@ -335,6 +341,8 @@ void QgsFeatureFilterModel::updateCompleter()
     emit filterJobCompleted();
   }
   emit endUpdate();
+
+  gathererThreadFinished();
 }
 
 void QgsFeatureFilterModel::gathererThreadFinished()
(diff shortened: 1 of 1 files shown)


lbartoletti (approved):
I can not reproduce the bug, but the fix looks good to me compared the crash report.

m-kuhn:
Good catch!
I'm pretty sure the crash is related to long roundtrips (a database which is not running locally) - and I could only ever observe it on Windows and Mac machines for some reason.

+1 to merge, but please let's refrain from merging the backports for a couple of days to leave it some time for real world tests.

PR #34607, opened 2020-02-22
3.10 backports

nyalldawson:
Backports which were delayed to after 3.10.3 (+some other fixes which weren't backported)

Notably includes the transform fallback work and shapefile encoding rework

For merge early in the 3.10.4 cycle for widest testing and feedback


nirvn (approved):

nyalldawson:
Thanks @nirvn!

PR #34617, opened 2020-02-24
New standalone console tool for running processing algorithms

nyalldawson:
**UPDATE**: the final tool is called `qgis_process`, all examples here have been updated to reflect the change

This new ~~qgis_transform~~ qgis_process tool allows users to run processing algorithms (both built-in, and those provided by plugins) directly from the console.

Running:

- `qgis_process list` will output a complete list of all available algorithms, grouped by provider.
- `qgis_process plugins` lists available and activated plugins which advertise the hasProcessingProvider metadata option (only these plugins are loaded by the tool)
- `qgis_process help algid` outputs the help and input descriptions for the specified algorithm, e.g. `qgis_process help native:centroids`

`qgis_process run`: runs an algorithm. Parameters are specified by a
`--param=value` syntax. E.g.

    qgis_process run native:centroids --INPUT="my_shapefile.shp" --OUTPUT="centroids.kml"

or

    qgis_process run native:buffer --INPUT=/home/me/my.shp --DISTANCE=20 --OUTPUT=/home/me/buffered.shp

While running an algorithm a text-based feedback bar is shown, and the operation can be cancelled via CTRL+C

Sponsored by the Swedish User Group

More detail is available in https://github.com/qgis/QGIS-Enhancement-Proposals/issues/140 

Still to come (follow up PRs):

- running a model directly from a model file
- running a script directly from an algorithm script file
- using the .env variable file approach when running under windows


roya0045:
Good to see this carried out!

nyalldawson:
Gotta update this -- looks like the consensus from the QEP discussion was to use qgis_process as the executable name.

src/transform/qgstransform.cpp
```diff
@@ -0,0 +1,440 @@
...
+  else
+    fflush( stdout );
+}
+
+
+//! load Python support if possible
+std::unique_ptr< QgsPythonUtils > loadPythonSupport()
+{
+  QString pythonlibName( QStringLiteral( "qgispython" ) );
+#if defined(Q_OS_MAC) || defined(Q_OS_LINUX)
```
lbartoletti: Q_OS_UNIX (include mac, linux and *BSD :wink:)

lbartoletti: and maybe
`#if defined(Q_OS_UNIX) and !defined(Q_OS_ANDROID)`

src/transform/main.cpp
```diff
@@ -0,0 +1,123 @@
...
+{
+#ifdef Q_OS_WIN  // Windows
+#ifdef _MSC_VER
+  _set_fmode( _O_BINARY );
+#else //MinGW
+  _fmode = _O_BINARY;
+#endif  // _MSC_VER
+#endif  // Q_OS_WIN
+
+  QgsApplication *app = new QgsApplication( argc, argv, false );
```
elpaso: Why a pointer? Cannot allocate on stack?

root676:
Very nice! Can't wait to see this in action!

src/transform/qgstransform.cpp
```diff
@@ -0,0 +1,440 @@
...
+    return execute( algId, params );
+  }
+  else
+  {
+    std::cerr << QStringLiteral( "Command %1 not known!\n" ).arg( command ).toLocal8Bit().constData();
+  }
+  return 1;
+}
+
+void QgsProcessingExec::showUsage( const QString &appName )
```
elpaso: I would recommend to use QCommandLineParser as done here: https://github.com/qgis/QGIS/blob/master/src/server/qgis_mapserver.cpp#L137

src/transform/main.cpp
nyalldawson: Fixed

src/transform/qgstransform.cpp
nyalldawson: I just tried to do this, but it's not possible as QCommandLineParser requires knowledge up front of the "--" syntax options, whereas we need to adapt these based on the algorithm selected

nyalldawson:
Now supports running models direct from the command line too!

PedroVenancio:
Wonderful @nyalldawson !

mbernasocchi:
> Now supports running models direct from the command line too!

wow so cool @nyalldawson thanks a lot, is it models in python or the xml (.model3). Can also custom python scripts be run?
Thanks again!

gioman:
@nyalldawson oh this is so great :)

first quick tests


```
giovanni@sibirica:~$ qgis_process run native:dissolve --INPUT="Downloads/distritos.gpkg" --DISSOLVE_ALL=true --OUTPUT="Downloads/dissolved.gpkg"
error loading plugin
                                                                                                                                                             
                                                                                                                                                             
----------------                                                                                                                                             
Inputs                                                                                                                                                       
----------------                                                                                                                                             
                                                                                                                                                             
DISSOLVE_ALL:   true                                                                                                                                         
INPUT:  Downloads/distritos.gpkg                                                                                                                                               
OUTPUT: Downloads/dissolved.gpkg                                                                                                                                               
                                                                                                                                                                               
0...10...20...30...40...50...60...70...80...90.
----------------
Results
----------------

OUTPUT: Downloads/dissolved.gpkg
```


not sure why the "error loading plugin" message appears. Maybe is related to the fact that I just compiled master but from the GUI I cannot load Processing ("missing osgeo module")?

gioman:
Does it work on a headless server?

src/process/qgsprocess.cpp
```diff
@@ -0,0 +1,505 @@
...
+int QgsProcessingExec::run( const QStringList &args )
+{
+  if ( args.size() == 1 )
+  {
+    showUsage( args.at( 0 ) );
+    return 0;
+  }
+
+  // core providers
+  QgsApplication::processingRegistry()->addProvider( new QgsNativeAlgorithms( QgsApplication::processingRegistry() ) );
```
luipir, line 161: sorry to arrive late... what about the other available/installed providers?

luipir: I mean, I can see that you load all processing plugin... do we already moved gdal, saga and grass as in plugins area?

nyalldawson: It should be done in https://github.com/qgis/QGIS/pull/34617/files/1e55212fd563b532fca57d8e0a8ce95cf40f9f27#diff-d8e4bd8d4ecd2dd30113b5c3a6cd2a1aR254 -- but looking at this now, I think we need to ensure that the main python processing plugin (where the gdal/grass/saga/otb initialisation occurs) is loaded before any other plugin. I'll push a follow up

nyalldawson:
@mbernasocchi 
>  is it models in python or the xml (.model3). Can also custom python scripts be run?

At the moment it's .model3 files. I'll look into allowing python scripts which provide a single algorithm too :+1:

nyalldawson:
> not sure why the "error loading plugin" message appears. Maybe is related to the fact that I just compiled master but from the GUI I cannot load Processing ("missing osgeo module")?

Hmm - possibly. I wonder why you're missing that module though, it's a dependency of QGIS. What platform is this on?

And what does `qgis_process plugins` show?

nyalldawson:
@gioman 
> Does it work on a headless server?

It's been intentionally designed to avoid all gui imports, so yes! :smile: Note that you may still need a fake x environment in order to use the rendering based algorithms (like the create XYZ tile algorithms) (for the same reasons as qgis server requires this -- something to do with font rendering in my understanding...)

nyalldawson:
@gioman https://github.com/qgis/QGIS/pull/35805 and https://github.com/qgis/QGIS/pull/35800 should give more helpful output for your error message

pigreco:
Is it possible to use it on windows ??
If yes, will it work via the osgeo shell ??

great job, thank you very much

gioman:
@nyalldawson thanks for the answers! One more question: is it possible to pipe the output of one command to another command?

alexbruy:
@gioman this is not possible. But there is a workaround - creating a model and running tool with this model.

mbernasocchi:
@nyalldawson I was trying a model that uses the qgis voronoi alg. Two things:

- the model stops without error and seems to ignore the voronoi alg even when counting steps
image
image

- qgis_process list shows no voronoi and some other algs are also missing, qgis_process plugins shows an empty list. what algs are available?

here the model in case you need it helps. if you remove the voronoi step all goes through correctly
[small.model3.txt](https://github.com/qgis/QGIS/files/4489043/small.model3.txt)

nyalldawson:
@mbernasocchi is that on master? (If not, can you try a new build?) There's better debugging output now for when a plugin fails to load

mbernasocchi:
@nyalldawson was on 3b3c7d8012407e14fb24b684bb8a623836202f4a (yesterday's master around 12 UTC IIRC. pulling new master now

mbernasocchi:
@nyalldawson I allowed myself to update the merge message to reflect the tool name change, I've also updated https://changelog.qgis.org/en/qgis/version/3.14/ . I'm wondering, is https://github.com/qgis/QGIS/blob/master/ChangeLog manually updated or somehow automatically?

agiudiceandrea:
@nyalldawson it seems these examples are not working

> ```
> qgis_process run native:centroids --INPUT="my_shapefile.shp" --OUTPUT="centroids.kml"
> ```
> or
> ```
> qgis_process run native:buffer --INPUT=/home/me/my.shp --DISTANCE=20 --OUTPUT=/home/me/buffered.shp
> ```

because qgis_process requires to specify the value of all the algorithm parameters, instead of let the algorithm use the default value if a parameter entry is missing.

Please have a look at https://github.com/qgis/QGIS/issues/37692

pigreco:
@nyalldawson 

An installation package on
Linux specific to qgis_process without the need to install all the
qgis package, for example without the GUI which is not needed in this case.

thank you so much

attualmente per utilizzare qgis senza gui sotto ubuntu 18.04 lts devo impostare la variabile:

`export QT_QPA_PLATFORM='offscreen'`

angri589:
How are multiple files/layers defined for an input for multilayers?  Example **native:mergevectorlayers**

suhendra0812:
> How are multiple files/layers defined for an input for multilayers? Example **native:mergevectorlayers**

Up. I have the same issue.

andrewharvey:
> How are multiple files/layers defined for an input for multilayers? Example native:mergevectorlayers

@angri589 @suhendra0812 you can pass the value twice, eg.

```
qgis_process run native:mergevectorlayers -- LAYERS=1.geojson LAYERS=2.geojson OUTPUT=output.geojson
```

suhendra0812:
Oh yeah, I just realize that. Thanks.

suhendra0812:
> > How are multiple files/layers defined for an input for multilayers? Example native:mergevectorlayers
> 
> @angri589 @suhendra0812 you can pass the value twice, eg.
> 
> ```shell
> qgis_process run native:mergevectorlayers -- LAYERS=1.geojson LAYERS=2.geojson OUTPUT=output.geojson
> ```

Oh, yeah. I just realize that. Thanks.

PR #34620, opened 2020-02-24
[labelling] Fix suboptimal logic for polygon's default around centroid

nirvn:
This PR improves (IMHO, greatly) the default labelling placement for polygon features. 

Long story short, our default placement for polgyons is set to 'around [centroid] point'. It takes the polygon's centroid, and following that handles things as if we were labelling a point feature. Which means it is actually skipping the _ideal_ placement: straight on top of the polygon's centroid.

This PR remedies to that by adding an 'over [centroid] point' location to the default placement.

Before vs. PR:
image

I'm not exactly sure how I lived that long without fixing this :wink: On a side note, @nyalldawson 's work demystifying and making the labelling code much more readable made it so much easier to improve labelling.


andreasneumann:
Nice - thanks for adding this!

Saijin-Naib:
Oooh wow, that's better

src/core/pal/feature.cpp
```diff
@@ -293,7 +293,9 @@ std::size_t FeaturePart::createCandidatesOverPoint( double x, double y, std::vec
   double labelW = getLabelWidth( angle );
   double labelH = getLabelHeight( angle );
 
-  double cost = 0.0001;
+  // if other candidates were created around a point prior to this (for polygons),
+  // lower the cost by half to favorize this placement
+  double cost = lPos.size() > 0 ? 0.00005 : 0.0001;
```
nyalldawson:
```suggestion
  double cost = !lPos.empty() ? 0.00005 : 0.0001;
```

src/core/pal/feature.cpp
```diff
@@ -1793,10 +1795,16 @@ std::vector< std::unique_ptr< LabelPosition > > FeaturePart::createCandidates( P
           case QgsPalLayerSettings::OverPoint:
```
nyalldawson: Can you split up this case statement to separate the Around/Over modes now? the amount of shared logic is much less than the specific logic for these two options

nyalldawson (commented):
I think some more logic is required here. createCandidatesOverPoint currently considers the quadoffset and positionOffset values from the label feature. If you had a layer set to "offfset from centroid" mode, where you had changed these settings, and then switched to "around centroid" mode, I think they will still maintain their values and be non-zero. So then the "over centroid" placement you are trying to make here will be offset from the centroid. 

There's actually not much logic from createCandidatesOverPoint that you're using here, so I'd suggest just copying that function to a new one "createCandidateOverPolygonCentroid", stripping out all the logic which applies to non-over-centroid placement, and calling that from FeaturePart::createCandidates. Make sense?

nyalldawson (approved):

nyalldawson:
Hurry up and merge... I can't stop noticing this all over my maps now :stuck_out_tongue:

Saijin-Naib:
Does this only take effect if we update our project to the labeling/placement engine V2?

nyalldawson:
Wheeeee!

PR #34631, opened 2020-02-25
Add the RandomPointsOnLines Processing algorithm 

havatv:
## Description

The ***Random points on lines*** algorithm supplements the ***Random points along line*** algorithm, and will prove to be more useful to the majority of users than the "original".

Features:
* The points are distributed randomly over the lines based on "along the line" distance, meaning that the distribution of the points will be flat over the length of the line (each place on the feature has the same probability of being "hit"). The green points in the illustration below have been generated using this method.
  The ***Random points along line***, on the other hand, uses a line segment based approach, meaning that the density will depend on the segment length (short segments will have a higher point density than longer ones). The blue points in the illustration below have been generated using this method - note the high density at the curved part near the bottom right end of the line.

  Comparison of the randomisation method (20 points generated). Green dots: proposed algorithm. Blue diamonds: The ***Random points along line*** algorithm. Both with a minimum distance of 0.
  randompoints

  Random points (20) on lines generated by the proposed algorithm (minimum distance is 0):
  random_rpol

  Random points (20) on lines generated by the the ***Random points along line*** algorithm (minimum distance is 0):
  random_rpal

* The features of the input layer are treated individually, and the algorithm will try to give each of them the specified number of points (maximum), producing a point layer with a maximum of
`#features` * `#points_per_feature` points.
  The ***Random points along line*** algorithm, on the other hand, distributes the points over all the features of the input dataset, resulting in a point dataset with the specified number of points (maximum).
* The random **seed** can be specified (not an option for the ***Random points along line***
   algorithm).

See also my proposal for updating the short documentation of the ***Random points along line*** algorithm (#34605).

This PR could be backported, as it does not depend on features in master (tested on 3.4.15 and 3.10.2).

One unit tests has been created. If the algorithm is accepted for inclusion, more tests could be added - if desired.

I will fix the documentation - expanding a bit on the short help, and including some illustrations.


havatv:
A unit test has been added.

havatv:
Unit test and code formatting OK - finally!

havatv:
Any hope for this one?
Something else that needs to be done before it can be considered for merging?

python/plugins/processing/algs/qgis/RandomPointsOnLines.py
```diff
@@ -0,0 +1,257 @@
...
+                                                       10, False, 1, 1000))
+        self.addParameter(QgsProcessingParameterNumber(self.RANDOM_SEED,
+                                                       self.tr('Integer seed (for random)'),
+                                                       QgsProcessingParameterNumber.Integer,
+                                                       None, True, 1, 1000000000))
+        self.addParameter(QgsProcessingParameterFeatureSink(self.OUTPUT,
+                                                            self.tr('Random points'),
+                                                            type=QgsProcessing.TypeVectorPoint))
+        self.addOutput(QgsProcessingOutputNumber(self.OUTPUT_POINTS,
+                                                 self.tr('Number of point generated')
```
nyalldawson:
```suggestion
                                                 self.tr('Total number of points generated')
```

python/plugins/processing/algs/qgis/RandomPointsOnLines.py
```diff
@@ -0,0 +1,257 @@
...
+                                                       1, False, 1, 1000000000))
+        self.addParameter(QgsProcessingParameterDistance(self.MIN_DISTANCE,
+                                                         self.tr('Minimum distance between points'),
+                                                         0, self.INPUT, False, 0, 1000000000))
+        self.addParameter(QgsProcessingParameterNumber(self.MAXTRIESPERPOINT,
+                                                       self.tr('Maximum number of attempts per point'),
+                                                       QgsProcessingParameterNumber.Integer,
+                                                       10, False, 1, 1000))
+        self.addParameter(QgsProcessingParameterNumber(self.RANDOM_SEED,
+                                                       self.tr('Integer seed (for random)'),
```
nyalldawson:
```suggestion
                                                       self.tr('Random number seed'),
```

python/plugins/processing/algs/qgis/RandomPointsOnLines.py
```diff
@@ -0,0 +1,257 @@
...
+        if source is None:
+            raise QgsProcessingException(self.invalidSourceError(parameters,
+                                                                 self.INPUT))
+
+        pointCount = self.parameterAsInt(parameters, self.POINTS_NUMBER,
+                                         context)
+        minDistance = self.parameterAsDouble(parameters, self.MIN_DISTANCE,
+                                             context)
+        randSeed = self.parameterAsInt(parameters, self.RANDOM_SEED, context)
+        MaxTriesPerPoint = self.parameterAsInt(parameters,
```
nyalldawson:
```suggestion
        maxTriesPerPoint = self.parameterAsInt(parameters,
```

python/plugins/processing/algs/qgis/RandomPointsOnLines.py
```diff
@@ -0,0 +1,257 @@
...
+                                               context, fields,
+                                               QgsWkbTypes.Point,
+                                               source.sourceCrs(),
+                                               QgsFeatureSink.RegeneratePrimaryKey)
+        if sink is None:
+            raise QgsProcessingException(
+                self.invalidSinkError(parameters, self.OUTPUT))
+
+        totNPoints = 0  # The total number of points generated
+        missedPoints = 0  # The number of misses (too close)
```
nyalldawson: Would be nice to make this an algorithm output too - maybe even two outputs (one for total number of misses, one for number of lines with misses)

python/plugins/processing/algs/qgis/RandomPointsOnLines.py
```diff
@@ -0,0 +1,257 @@
...
+
+        maxIterations = pointCount * MaxTriesPerPoint
+        # Go through all the features of the layer
+        for f in source.getFeatures():
+            if feedback.isCanceled():
+                break
+            lineGeoms = []  # vector of line elements (>=1 for multi)
+            fGeom = f.geometry()  # get the geometry
+            if fGeom is None:
+                feedback.reportError('Null geometry - skipping!', False)
```
nyalldawson: We usually just skip these without warning

python/plugins/processing/algs/qgis/RandomPointsOnLines.py
```diff
@@ -0,0 +1,257 @@
...
+            while nIterations < maxIterations and nPoints < pointCount:
+                # Try to generate a point
+                if feedback.isCanceled():
+                    break
+                # Get the random "position" along the line for this point
+                randomLength = random.random() * totLineLength
+                # To accumulated line length as we move along the geometry:
+                accLength = 0
+                # Go through the parts to find the random point location
+                for l in lineGeoms:
```
nyalldawson: This could all be greatly simplified -- you could directly call QgsGeometry::interpolate on the original line string feature's geometry with the desired randomLength. Basically you're re-implementing QgsGeometry::interpolate here.

python/plugins/processing/algs/qgis/RandomPointsOnLines.py
```diff
@@ -0,0 +1,257 @@
...
+        pointCount = self.parameterAsInt(parameters, self.POINTS_NUMBER,
+                                         context)
+        minDistance = self.parameterAsDouble(parameters, self.MIN_DISTANCE,
+                                             context)
+        randSeed = self.parameterAsInt(parameters, self.RANDOM_SEED, context)
+        MaxTriesPerPoint = self.parameterAsInt(parameters,
+                                               self.MAXTRIESPERPOINT,
+                                               context)
+        fields = QgsFields()
+        fields.append(QgsField('id', QVariant.Int, '', 10, 0))
```
nyalldawson: Instead of this, I'd rather keep all the input fields intact and just append a new "point_num" field on the end. That'll be more useful for users.

nyalldawson (commented):
Looking close! (and sorry for the delayed review).

python/plugins/processing/algs/qgis/RandomPointsOnLines.py
havatv: Will fix - probably tomorrow...

python/plugins/processing/algs/qgis/RandomPointsOnLines.py
havatv: I will look into that. Perhaps tomorrow.

python/plugins/processing/algs/qgis/RandomPointsOnLines.py
havatv: I think the user has the right to know. Could perhaps be more silent...

python/plugins/processing/algs/qgis/RandomPointsOnLines.py
havatv: I think I did that in an earlier iteration.  Don't remember why I went for this solution. I will look at it.

havatv (commented):
Many thanks for reviewing, @nyalldawson! And very useful input and suggestions. I will have a closer look.

python/plugins/processing/algs/qgis/RandomPointsOnLines.py
havatv: I added this as an option. The default is not to include the attributes from the line.

python/plugins/processing/algs/qgis/RandomPointsOnLines.py
havatv: Added as outputs (POINTS_MISSED and LINES_WITH_MISSED_POINTS).
I also added an output for the number of null or empty geometries (LINES_WITH_EMPTY_OR_NO_GEOMETRY) and

python/plugins/processing/algs/qgis/RandomPointsOnLines.py
havatv: Changed to pushInfo.

havatv (commented):
Thanks again for you thorough review, @nyalldawson. I have tried to address all but one of your suggestions / input. I have also added a new test with "minimum distance" and "attributes from lines" set.

havatv:
Travis failure:

> 10/10 Test  #8: qgis_sip_uptodate ................***Failed   45.59 sec
> 
>  *** SIP file not up to date: core/auto_generated/fieldformatter/qgsvaluerelationfieldformatter.sip
> 
>  28c28
> 
> <       ValueRelationItem( const QVariant &key, const QString &value, const QString &description );
> 
> ---
> 
> >       ValueRelationItem( const QVariant &key, const QString &value, const QString &description = QString() ); 

Is that my fault?

roya0045:
> Travis failure:
> 
> > 10/10 Test  #8: qgis_sip_uptodate ................***Failed   45.59 sec
> > *** SIP file not up to date: core/auto_generated/fieldformatter/qgsvaluerelationfieldformatter.sip
> > 28c28
> > <       ValueRelationItem( const QVariant &key, const QString &value, const QString &description );
> > 
> > > ```
> > >   ValueRelationItem( const QVariant &key, const QString &value, const QString &description = QString() ); 
> > > ```
> 
> Can that be my fault?

Nope, someone didn't sip their pr, don't mind it.

python/plugins/processing/algs/qgis/RandomPointsOnLines.py
nyalldawson: Can I ask you to revisit this? Making it optional deviates from the behaviour of all other similar algorithms, such as "points along geometry". We should always keep the input attributes intact and not make this optional.

python/plugins/processing/algs/qgis/RandomPointsOnLines.py
nyalldawson: How'd you go with this? I'd like to avoid the duplicate (python heavy) code

havatv:
@nyalldawson wrote (on the inclusion of attributes from the line layer):

> Can I ask you to revisit this? Making it optional deviates from the behaviour of all other similar algorithms, such as "points along geometry". We should always keep the input attributes intact and not make this optional.

In my opinion, this is not a "normal" input layer, as the lines will only be reflected in the output layer as point "samples". I therefore still mean that including the attributes of the line layer should be optional. But I agree that we can change the default to include the attributes (so that the user will have to "opt out" of what you say is the normal behaviour, instead of "opt in").

nyalldawson:
> In my opinion, this is not a "normal" input layer, as the lines will only be reflected in the output layer as point "samples".

I still see it as essential to keep the input attributes -- if I was doing a sample based on the input lines, then without the input attributes the result has no context...

havatv (commented):
> In my opinion, this is not a "normal" input layer, as the lines will only be reflected in the output layer as point "samples". I therefore still mean that including the attributes of the line layer should be optional. But I agree that we can change the default to include the attributes (so that the user will have to "opt out" of what you say is the normal behaviour, instead of "opt in").

You have convinced me that the default should be to include the attributes from the line layer. So I have now made this parameter optional, and have changed the default to True (that is, include attributes from the line layer).
It can't be a problem to give the user the option to leave the attributes out, except that it clutters the GUI (but that is not much of an issue here). And I disagree that the points have not context without the line attributes - the context is implicit in that they are all placed on lines from the same layer. The only context that is lost is which line the point is on. If you want to get random sample point on a stream/river or road network, your primary goal is most probably to get the location of the sample. You may or may not be interested in which part of the stream network the point is on.

havatv:
Seems as if there is a very slight difference in the geometries returned by the new variant. I will update the tests.

havatv:
@nyalldawson, I hope to have addressed / implemented all your comments and suggestions, and both tests seem to run OK.
Anything else you would like me to do?

nyalldawson:
Superseded by #35458 35282

havatv:
> Superseded by #35458 35282

Should be #35282

PR #34636, opened 2020-02-25
Add parameters to contour.py to enable creation of polygons

pwicks86:
## Description

`gdal_contour` supports creating polygon contours instead of lines, but QGIS doesn't expose that functionality. This commit adds the necessary bits to expose it. Specifically, it adds 3 new parameters to the plugin:

* `CREATE_POLYGON`: This controls whether polygon mode is enabled, i.e. is `-p` passed to `gdal_contour`. 

* `FIELD_NAME_MIN` and `FIELD_NAME_MAX`: These correspond to the `-amin` and `-amax` parameters of `gdal_contour` and are only added to the output if the polygon mode is enabled.

 I also modified the plugin to avoid setting the `-a` parameter when polygon mode is enabled to ensure that there aren't extraneous errors from `gdal_contour`. 

### Screenshot
image

### DIFF
--- a/python/plugins/processing/algs/gdal/contour.py
+++ b/python/plugins/processing/algs/gdal/contour.py
@@ -213,12 +213,12 @@ class contour_polygon(contour):
         self.removeParameter(contour.FIELD_NAME)
 
         self.addParameter(QgsProcessingParameterString(self.FIELD_NAME_MIN,
-                                                       self.tr('Attribute name to use for the minimum elevation of contour polygons (if not set, no elevation attribute is attached)'),
+                                                       self.tr('Attribute name for minimum elevation of contour polygon'),
                                                        defaultValue='ELEV_MIN',
                                                        optional=True))
 
         self.addParameter(QgsProcessingParameterString(self.FIELD_NAME_MAX,
-                                                       self.tr('Attribute name to use for the maximum elevation of contour polygons (if not set, no elevation attribute is attached)'),
+                                                       self.tr('Attribute name for maximum elevation of contour polygon'),
                                                        defaultValue='ELEV_MAX',
                                                        optional=True))
 
--- a/python/plugins/processing/algs/gdal/contour.py
+++ b/python/plugins/processing/algs/gdal/contour.py
@@ -47,7 +47,10 @@ class contour(GdalAlgorithm):
     BAND = 'BAND'
     INTERVAL = 'INTERVAL'
     FIELD_NAME = 'FIELD_NAME'
+    FIELD_NAME_MIN = 'FIELD_NAME_MIN'
+    FIELD_NAME_MAX = 'FIELD_NAME_MAX'
     CREATE_3D = 'CREATE_3D'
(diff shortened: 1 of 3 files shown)


nyalldawson:
Nice addition @pwicks86 !

Can you extend the existing unit tests from https://github.com/qgis/QGIS/blob/master/python/plugins/processing/tests/GdalAlgorithmsRasterTest.py#L407 to cover these new options too?

alexbruy:
I'm a big -1 for these changes.

We have already added "Additional command line parameters" exactly for such cases to allow users pass any additional parameters not exposed in the GUI. And we already rejected several similar chnages to other tools.

pwicks86:
@nyalldawson Sure, no problem.

nyalldawson:
@alexbruy I agree with you in principal, but I think that the new polygon switch is a sufficiently "high level" option that we should expose this one.

(Alternatively - what would you think about a specific "Contours (polygon)" algorithm using a subclass of the existing alg?)

nirvn:
@nyalldawson , I like the idea of adding a contours (polygon) algorithm, that'd be far more discoverable to begin with.

pwicks86:
Either way works for me, I just kept running into situations where I wanted to be able to use this easily from QGIS.

alexbruy:
@nyalldawson if we really-really need to have this exposed via GUI, IMHO better to have separate algorithm just for creating polygons without all other options.

Probably we also need to define some rules when such changes will be accepted and when rejected.

pwicks86:
If we make a `contour_polygon` that subclasses `contour`, should it appear in the menu under `Extraction` or just in the processing toolbox?

pwicks86:
I'll just leave it in the processing toolbox for now, should have an updated PR soon.

pwicks86:
Hrm, not sure why that unit test is failing, investigating...

python/plugins/processing/algs/gdal/contour.py
```diff
@@ -192,5 +192,62 @@ def getConsoleCommands(self, parameters, context, feedback, executing=True):
...
+    def __init__(self):
+        super().__init__()
+
+    def initAlgorithm(self, config=None):
+        super().initAlgorithm(config)
+        # FIELD_NAME isn't used in polygon mode
+        self.removeParameter(contour.FIELD_NAME)
+
+        self.addParameter(QgsProcessingParameterString(self.FIELD_NAME_MIN,
+                                                       self.tr('Attribute name to use for the minimum elevation of contour polygons (if not set, no elevation attribute is attached)'),
```
nyalldawson:
```suggestion
                                                       self.tr('Attribute name for minimum elevation of contour polygon'),
```

It's a bit more concise, and the existing description is longer than we generally use for parameters

nyalldawson: (same with the max attribute)

nyalldawson (commented):
Looking good to me!

nyalldawson (approved):

PR #34654, opened 2020-02-25
[feature] Relation widget force suppress popup

elpaso:
Adds an option to the relation widget configuration
to suppress form popup open when new features are
added in an embedded form context.

This option overrides the form-level option (that might
still be the desired behavior when the form is used
as a standalone form).

qgis-feature-hide-form-on-new-features

Note: the final label is **Force hide form on add feature**

### DIFF
--- a/python/core/auto_generated/qgsvectorlayertools.sip.in
+++ b/python/core/auto_generated/qgsvectorlayertools.sip.in
@@ -91,6 +91,25 @@ Copy and move features with defined translation.
          - errorMsg: If given, it will contain the error message
 %End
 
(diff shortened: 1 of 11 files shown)


elpaso:
@m-kuhn what's your feeling on this one?

m-kuhn:
Good idea.

Could we synchronize the wording between frontend and API / here and in the "global" layer properties?

elpaso:
@m-kuhn I'll change it to "Force hide form on add feature" to make it clear that it overrides all other options.

m-kuhn (approved):

PR #34662, opened 2020-02-25
fix crash when exporting check geometries gaps

### LINKED ISSUE #34661: QGIS crashes when exporting check geometries results
**Describe the bug**
QGIS crashes when trying to export the results of *Check Geometries*, specifically when the results contain Gaps.

**How to Reproduce**
-Run check geometries on a polygon layer with gaps, selecting *check for gaps smaller than...*
-Click export on the results tab

**QGIS and OS versions**
Master on Debian

uclaros:
## Description
Check geometries Gaps results do not belong to a layer, so `error->layerId()` returns an empty string. `mChecker->featurePools()[error->layerId()]->layer()` then crashes QGIS.

I guess it should be backported too.

Fixes #34661

### DIFF
--- a/src/plugins/geometry_checker/qgsgeometrycheckerresulttab.cpp
+++ b/src/plugins/geometry_checker/qgsgeometrycheckerresulttab.cpp
@@ -278,7 +278,7 @@ bool QgsGeometryCheckerResultTab::exportErrorsDo( const QString &file )
   for ( int row = 0, nRows = ui.tableWidgetErrors->rowCount(); row < nRows; ++row )
   {
     QgsGeometryCheckError *error = ui.tableWidgetErrors->item( row, 0 )->data( Qt::UserRole ).value<QgsGeometryCheckError *>();
-    QString layerName = QString( "" );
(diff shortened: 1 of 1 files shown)


src/plugins/geometry_checker/qgsgeometrycheckerresulttab.cpp
```diff
@@ -278,9 +278,16 @@ bool QgsGeometryCheckerResultTab::exportErrorsDo( const QString &file )
   for ( int row = 0, nRows = ui.tableWidgetErrors->rowCount(); row < nRows; ++row )
   {
     QgsGeometryCheckError *error = ui.tableWidgetErrors->item( row, 0 )->data( Qt::UserRole ).value<QgsGeometryCheckError *>();
-    QgsVectorLayer *srcLayer = mChecker->featurePools()[error->layerId()]->layer();
+    QString layerName = QString( "" );
```
m-kuhn:
```suggestion
    QString layerName = QString();
```

m-kuhn (approved):

PR #34674, opened 2020-02-26
Make "pre-set DPI not equal to painter's DPI" check less stringent

nyalldawson:
This assert gets tripped frequently when working with magnified maps, because there's tiny (inconsequential) differences in the dpi value.


nirvn:
Ah thanks, hit that wall quite a few times myself.

PR #34675, opened 2020-02-26
[FEATURE] Mesh dataset in 3D view

vcloarec:
Time being, mesh layer can be rendered in 3D view only based on the Z values of the mesh's vertices to render the terrain.
This PR adds the possibility to render a mesh layer in the 3D view based on values of the dataset. 
The vertical component of the vertices can be set equal to dataset values (for example, level of water surface) and the texture of the mesh can be set to render other dataset values with color shading (for example velocity).
The entry to render dataset in 3D view is the layer properties of the mesh layer. 

image

So, the principle to render the two types of rendering (Z value and dataset) is to use two entries ;
- the terrain settings are in the 3D configuration widget, as before this PR
- the dataset rendering settings are in the mesh layer properties, the user can choose:
                      - the dataset group that will be used for the vertical component of the mesh
                      - the vertical scale
                      - If the dataset has to be considered as relative to the vertices Z value
                      - The rendering color settings : unique color, or with the color ramp shader already used for 2D shading

That permits to render the terrain and the dataset of a mesh layer in the same 3D view.

image


nirvn:
@vcloarec , keep those PRs coming, fantastic work! :)

Looking at the screenshot only, one comment: we really want to avoid fixed unit type (especially pixels), it doesn't export well.

vcloarec:
@nirvn , tell me if I am wrong, but in 3D view, we can't used unit map for rendering the wireframe line because, the wireframe line width is the same for all distance from the camera.

nirvn:
@vcloarec , oh sorry, yes, you are correct for 3D wireframe line.

PeterPetrik:
I did a pre-review in the original PR https://github.com/vcloarec/QGIS/pull/5 already

wonder-sk (approved):
The code looks good to me!

PR #34689, opened 2020-02-26
[processing modeler] display "parents" links with dotted stroke

olivierdalang:
## Description

This makes the "parent" links display as dotted lines in the modeler instead of solid lines, so that they are less visually proeminent.

### DIFF
--- a/python/plugins/processing/modeler/ModelerScene.py
+++ b/python/plugins/processing/modeler/ModelerScene.py
@@ -152,6 +152,7 @@ class ModelerScene(QGraphicsScene):
             for depend in alg.dependencies():
                 arrow = ModelerArrowItem(self.algItems[depend], -1,
                                          self.algItems[alg.childId()], -1)
+                arrow.setPenStyle(Qt.DotLine)
                 self.algItems[depend].addArrow(arrow)
                 self.algItems[alg.childId()].addArrow(arrow)
                 arrow.updatePath()


nyalldawson:
Got a before/after screenshot?

olivierdalang:
before:

before

after:

after

nirvn:
I'm -1 on this. IMHO it actually makes it more confusing than not at first glance.

olivierdalang:
> I'm -1 on this. IMHO it actually makes it more confusing than not at first glance.

@nirvn I think it's good have a visual distinction between normal links (that represent data passing from one algorithm to another) and other type of links (`parent` is just a way to tell the order of execution of algorithm, no data is involved). Also it uses the same style as dependencies for inputs (such as "field" inputs depending on "vector layer" inputs), so it's not something completely new either...

nyalldawson:
I'm with @nirvn here -- I don't see the different between output based model flow and flow based on hard dependencies between algorithms. The dotted line suggests a "weak" link, yet if you've gone to the trouble of manually setting these dependencies then you've clearly got a good reason to do so...

m-kuhn:
@nyalldawson and @nirvn do you prefer *no distinction at all* between parent and data flow connection or is the current proposal not suitable and something else should be done?

CC @gacarrillor

m-kuhn:
Ping @nyalldawson @nirvn

nirvn:
@m-kuhn , on my end, I don't much of a difference between a e.g. user-inputted vector dataset vs algorithm's intermediary output vector dataset used as an input for a subsequent algorithm.

On top of which, the dotted line connection is already used to identify linkages between input parameters (for eg vector dataset <-> vector field attribute). That makes me a double-no for using the dotted line connection :)

If there is a need to identify a connection between a user inputted parameter, maybe using the yellow color of that box type to color the connector would be a better solution.

m-kuhn:
> @m-kuhn , on my end, I don't much of a difference between a e.g. user-inputted vector dataset vs algorithm's intermediary output vector dataset used as an input for a subsequent algorithm.

At least to my understanding this comparison does not correspond with the goal of this PR.

Correct me if I am wrong, but here we talk about:

 - Algorithm Output/User Input -> Algorithm Input Connections (data flow -> stroked line)
 - Parent algorithms (dependency on algorithm level, no inputs or outputs involved, no direct data flow -> dashed line)

nirvn:
@m-kuhn , doh, you are correct. I'm withdrawing myself from this discussion then :wink: 

I don't have strong feelings on how to draw connections between parent <-> child algorithms.

nyalldawson:
I still feel it's very similar. You don't manually set a parent unless there's some dependency which the modeller was unable to automatically detect. And in that case, the child is just as dependant on the parent as it is when it relies on data coming from a parent..

m-kuhn:
Nobody said they are not at least similar ;)

While the two concepts have a common denominator (dependency) they also have differences (data flow on processing level).

Currently the parent algorithm line goes into "In" which suggests that it's connected to an input and there's a data flow on processing level happening.

Would a gray line instead of a dashed line make it more acceptable?
Or connecting the line to another connector than "In"?

nyalldawson:
> Currently the parent algorithm line goes into "In" which suggests that it's connected to an input and there's a data flow on processing level happening.

@m-kuhn That's an older screenshot. Here's how it looks on master:

image

Honestly my preference would be to show an explicit "arrow" here, showing the directional flow....

olivierdalang:
@nyalldawson Ok now I'm confused :-) In your screenshot, which one is the "parent" link ? The solid line between `Basic statistics` and `Buffer` or the dotted line on the top-right ? Is that screenshot current master or current master with this PR applied ?

nyalldawson:
It's the solid line between Basic statistics and Buffer or the dotted line on the top-right . The other one is joined to a comment

gacarrillor:
I think @nyalldawson's screenshot improves readability. There are 2 algorithms connected, no longer an algorithm to an input parameter. 

> Honestly my preference would be to show an explicit "arrow" here, showing the directional flow....

I think an arrow would make algorithm dependencies crystal clear :+1:

nyalldawson:
In any case, this pr would need to be reworked following these classes bring ported to C++ - it no longer applies cleanly

PR #34697, opened 2020-02-26
[Bugfix][Server] WMS GetFeatureInfo provides bounding box when configured

rldhont:
Some providers always retrieve geometry even if the NoGeometry flags is used in request. So in WMS GetFeatureInfo the bounding box even if it as not been configured.

The `WITH_GEOMETRY` parameter has been introduced to force getting the geometry. The project can contain a parameter to add wkt geometry in the WMS GetFeatureInfo response: `wmsFeatureInfoAddWktGeometry`

For example:
* if the datasource is an ESRI ShapeFIle, the bounding box is always provided
* if the datasource is an Oracle table or a Psotgis one, the bounding box is only provided when `wmsFeatureInfoAddWktGeometry` and `WITH_GEOMETRY`

To provide something more homogeneous:
* The wkt geometry is only provided if the project has been configured to provide wkt geometry and the request contains WITH_GEOMETRY.
* The bounding box is provided if the project has been configured to provide wkt geometry
* The bounding box is not provided if the project has not been configured to provide wkt geometry

### DIFF
--- a/src/server/services/wms/qgswmsrenderer.cpp
+++ b/src/server/services/wms/qgswmsrenderer.cpp
@@ -1424,7 +1424,7 @@ namespace QgsWms
     bool segmentizeWktGeometry = QgsServerProjectUtils::wmsFeatureInfoSegmentizeWktGeometry( *mProject );
     const QSet<QString> &excludedAttributes = layer->excludeAttributesWms();
 
-    bool hasGeometry = addWktGeometry || featureBBox || layerFilterGeom;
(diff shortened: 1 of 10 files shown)


pblottiere:
> To provide something more homogeneous:
>
> - The wkt geometry is only provided if the project has been configured to provide wkt geometry and the request contains WITH_GEOMETRY.
> - The bounding box is provided if the project has been configured to provide wkt geometry
> - The bounding box is not provided if the project has not been configured to provide wkt geometry

The rationale looks good to me :+1:.

Thanks @rldhont.

pblottiere (approved):

PR #34736, opened 2020-02-27
Relation-ref widget ask the user to save the parent layer when saving…

elpaso:
… children

When the user clicks on "Save child layer edits" on an embedded
form if the following condition are met:

- the parent feature was added but not saved (feature id is 0)
- the layer is not part of a transaction group

qgis-embedd-forms-save-prent


elpaso:
I know this PR does not solve the problem in all scenarios, but perhaps make it easier to use the relation reference widget in a a few simple cases.

PR #34739, opened 2020-02-27
Add a vertical spacer at the bottom of mask label tab

DelazJ:
to avoid this 
image 
wonder if the spacer should not be right after the "Draw effects" button...


nyalldawson:
> wonder if the spacer should not be right after the "Draw effects" button...

I think that would cause the helper label to be too easy to overlook

PR #34741, opened 2020-02-27
.github/ISSUE_TEMPLATE/bug_report.md: explicit mention that questions aren't appropriate for github issues

rouault:
Directly borrowed from GDAL bug_report.md

Admittedly this doesn't stop users abusing github issues, but at least we can
close such inappropriate tickets without much discussion

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -11,7 +11,7 @@ IMPORTANT: Do NOT use GitHub to post any questions or support requests!
            They will be closed immediately and ignored.
 
 Questions should go to the qgis-user mailing list at https://lists.osgeo.org/mailman/listinfo/qgis-user
-or other support forums. GitHub issues are for bug reports and suggestions for new features.
+or other support forums such as https://gis.stackexchange.com/. GitHub issues are for bug reports and suggestions for new features.
 
 Bug fixing and feature development is a community responsibility, and not the responsibility of the QGIS project alone.
 If this bug report or feature request is high-priority for you, we suggest engaging a QGIS developer or support organisation and financially sponsoring a fix
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,6 +7,12 @@ assignees: ''
 
 ---
 <!--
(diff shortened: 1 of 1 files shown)


roya0045:
Big +1

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -7,6 +7,12 @@ assignees: ''
 
 ---
 <!--
+IMPORTANT: Do NOT use GitHub to post any questions or support requests!
+           They will be closed immediately and ignored.
+
+Questions should go to the qgis-user mailing list at https://lists.osgeo.org/mailman/listinfo/qgis-user
+or other support forums. GitHub issues are for bug reports and suggestions for new features.
```
nyalldawson:
```suggestion
or other support forums such as https://gis.stackexchange.com/. GitHub issues are for bug reports and suggestions for new features.
```

Gustry:
Side question. It seems people are not reading too much the template (narrow text input, difficult to read...). 
Why not using the same approach as on QGIS-Documenation : 
Using a template like this : 
https://raw.githubusercontent.com/qgis/QGIS-Documentation/master/.github/ISSUE_TEMPLATE/2_application_issue.md which gives https://github.com/qgis/QGIS-Documentation/issues/new/choose It seems no-one is now using the QGIS Application Bug report button. Maybe @DelazJ or @havatv can confirm?

roya0045:
> Side question. It seems people are not reading too much the template (narrow text input, difficult to read...).
> Why not using the same approach as on QGIS-Documenation :
> Using a template like this :
> https://raw.githubusercontent.com/qgis/QGIS-Documentation/master/.github/ISSUE_TEMPLATE/2_application_issue.md which gives https://github.com/qgis/QGIS-Documentation/issues/new/choose It seems no-one is now using the QGIS Application Bug report button. Maybe @DelazJ or @havatv can confirm?
> 
> It's written that users shouldn't create tickets about third party plugins, but we still have some tickets sometimes. Maybe we should do this too?

I think that more buttons would be a better option, an option could also be added for specific recurrent issues such as the mac 'malware' popup.

m-kuhn:
@Gustry :100:

Gustry:
I just made a PR with my proposition. #35017

PR #34747, opened 2020-02-28
Align 3D View properties tab rendering with common properties tabs

DelazJ:
* using groupbox for items instead of line separator
* reducing the oversized margins

before (from 3.10 but still true in 3.12 for the issues) vs after
image

image

image

image

PS: I don't know whether "shading" is the appropriate word...


nyalldawson:
MUCH nicer, thanks!

I noticed there's a capitalization error in the 3d Config dialog -- the group box title for "Terrain shading" should be "Terrain Shading"

nirvn:
@DelazJ , keep those UI fixes coming :) :+1:

PR #34748, opened 2020-02-28
Fix relation reference widget test crashes under debug builds

nyalldawson:
Because of the new assert in QgsMapToolAdvancedDigitizing requiring the cad widget

Not the cleanest fix -- the widget should instead gracefully handle having no cad widget available by disabling the digiting tool instead, but we urgently need the CI infrastructure to stop flagging everything because of this change.


nirvn:
Thanks for handling this one.

elpaso:
Thanks, on a more general note: should we start/continue adding `Q_ASSERT`s to make it clear when certain constructors (or methods) have strong preconditions?

Instead of asserts we could find another way to document preconditions but there must be **one way or another**, especially because we pass a lot of pointers around.

PR #34656, opened 2020-02-25
Drag'n'drop improvements backport to 3.10

3nids:
backport of #34411, #34433, #34479 and 34503
it will live for a month in 3.12 before reaching next release of 3.10, so this sounds safe enough to me

PR #34749, opened 2020-02-28
Remove oversized margin of the Container properties frame

DelazJ:
in drag-and-drop mode of attribute forms, harmonizing with fields properties, to look like this

image
instead of 
image

To backport to 3.10 after #34656


3nids:
thanks

PR #34750, opened 2020-02-28
Remove duplicate nodes from each part in a collection

gacarrillor:
## Description

**Expected**: Running the `Remove duplicate vertices` algorithm does remove duplicate vertices from each part in a geometry collection (e.g., a MultiPolygon layer).

**Currently**: Each run of such algorithm removes duplicate vertices in a single part at a time. So, if you have 3 parts that have duplicate vertices (each one separately), you need to run the algorithm 3 times to 'clean' the layer. :confused: 

This PR fixes the issue, adjusting the GeometryCollection.removeDuplicateNodes() method, which was causing the misbehaviour in the Processing algorithm.

New tests for:
 + MultiPolygon.removeDuplicateNodes()
 + Remove Duplicate Vertices algorithm.
 + Second call to removeDuplicateNodes() should return False for basic geometry types.


nyalldawson:
Good catch!

nyalldawson (approved):

gacarrillor:
The test for Remove Duplicate Vertices algorithm is fixed now. I was using GML 3.2 (I chose that version in Save As... dialog), but now I just selected the default version.

There is still a test failing (`217 - qgis_relationreferencewidget (Child aborted)`), which is unrelated to this PR.

PR #34765, opened 2020-02-28
[BUG][Mesh] Fix streamlines/traces issue when CRS transform fails

vcloarec:
When CRS transform fails,  trace/streamlines renderers can have issue, QGIS crash or streamlines not rendered.
This PR fix this issue.


PeterPetrik (approved):

PeterPetrik:
wondering if can add some tests

PeterPetrik:
unable to backport to 3.10 since streamlines is not available there

PR #34775, opened 2020-02-28
Fix crash when opening raster layer properties

### LINKED ISSUE #34772: Crash when opening raster layer properties
When selecting a raster layer's properties, (right-click raster layer, select properties...) QGIS crashes with a signal 11.

It looks like the problem was introduced in commit 3aaeb81df016164af6b280a7a77eb8f595d1fbe3 and is particular to my type of display setup.

I have 3 screens, arranged as:
```
         +----+
+-------+|    |+--------+
|   3   ||  2 ||    1   |
+-------+|    |+--------+
         +----+
```

When layer properties is attempting to calculate the horizontalDpi, it is asking QT for a screen at a position that doesn't exist.

PR with fixes is forthcoming

ttislerdg:
This bug was exposed on my particular screen setup (see #34772).  The raster layer properties was trying to get the screen at an x,y position where there was no screen.

When QgsRasterLayerProperties constructor was trying to calculate the DPI of the screen the widget was on, but since hasn't been displayed yet, it diesn't have a logical location on the screen. 

`QScreen *screen = QGuiApplication::screenAt( mapToGlobal( QPoint( width() / 2, 0 ) ) );`

i.e. mapToGlobal(...) was returning the same value that was sent in, which isn't a valif position on my screen, hence *screen was null, and the next usage caused a NULL pointer access crash. 

Since QT5, the QPaintDevice class contains a logicalDpiX().  QWidget is a QPaintDevice, so we can simply use that method to get the logical DPI.

NOTE:  I don't have a high DPI display, so I can't verify changes on those devices.

This change should be backported to any version(s) updated with  3aaeb81 

Fixes #34772

### DIFF
--- a/doc/CONTRIBUTORS
+++ b/doc/CONTRIBUTORS
@@ -86,6 +86,7 @@ Shiva Reddy Koti
 Stefanie Tellex
 Steven Mizuno
 Tamas Szekeres
+Tim Tisler
 Tom Russo
 Tyler Mitchell
 Vita Cizek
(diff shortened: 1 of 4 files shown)


nirvn:
Nice first(?) contribution.

PR #34777, opened 2020-02-28
Fix Python3 'TypeError' when loading Oracle metadata in DB Manager

### LINKED ISSUE #34774: Python3 'TypeError' when loading Oracle metadata in DB Manager
Error occurs when loading a single Oracle Layer in DB Manager.
Info tab doesn't load properly and a python error is thrown.

Exact error:
TypeError: '>=' not supported between instances of 'NoneType' and 'int'

File/Line: https://github.com/qgis/QGIS/blob/master/python/plugins/db_manager/db_plugins/oracle/info_model.py#L145

QGIS version: 3.4.8 (but last modification of this file was 2 years ago)
Python version: 3.7

Possible duplicate of #32086.
The same error as #25250 but using Oracle (I believe).

raulbbarbosa:
Error is related to python 2 to 3 migration.
See https://docs.python.org/3/whatsnew/3.0.html#ordering-comparisons

Adding proper check for NoneType before checking if variable is 'greater than' solves it.

Fixes #34774

### DIFF
--- a/python/plugins/db_manager/db_plugins/oracle/info_model.py
+++ b/python/plugins/db_manager/db_plugins/oracle/info_model.py
@@ -142,7 +142,7 @@ class ORTableInfo(TableInfo):
                     "DBManagerPlugin", "Rows (estimation):"),
                  self.table.estimatedRowCount)
             )
-        if self.table.rowCount or self.table.rowCount >= 0:
+        if self.table.rowCount is not None and self.table.rowCount >= 0:
             # Add a real count of rows
             tbl.append(
                 (QApplication.translate("DBManagerPlugin", "Rows (counted):"),
@@ -624,7 +624,7 @@ class ORVectorTableInfo(ORTableInfo, VectorTableInfo):
         if self.table.extent and len(self.table.extent) == 4:
             extent_str = (u"{:.9f}, {:.9f} - {:.9f}, "
                           u"{:.9f}".format(*self.table.extent))
-        elif self.table.rowCount > 0 or self.table.estimatedRowCount > 0:
+        elif (self.table.rowCount is not None and self.table.rowCount > 0) or (self.table.estimatedRowCount is not None and self.table.estimatedRowCount > 0):
             # Can't calculate an extent on empty layer
             extent_str = QApplication.translate(
                 "DBManagerPlugin",


python/plugins/db_manager/db_plugins/oracle/info_model.py
```diff
@@ -624,7 +624,7 @@ def spatialInfo(self):
         if self.table.extent and len(self.table.extent) == 4:
             extent_str = (u"{:.9f}, {:.9f} - {:.9f}, "
                           u"{:.9f}".format(*self.table.extent))
-        elif self.table.rowCount > 0 or self.table.estimatedRowCount > 0:
+        elif (self.table.rowCount is not None and self.table.rowCount > 0) or (self.table.estimatedRowCount is not None and self.table.estimatedRowCount > 0):
```
troopa81: same comments than above

python/plugins/db_manager/db_plugins/oracle/info_model.py
```diff
@@ -142,7 +142,7 @@ def generalInfo(self):
                     "DBManagerPlugin", "Rows (estimation):"),
                  self.table.estimatedRowCount)
             )
-        if self.table.rowCount or self.table.rowCount >= 0:
+        if self.table.rowCount is not None and self.table.rowCount >= 0:
```
troopa81: Just replacing `or` with `and` should fix the issue no? the `is not None` check is not necessary. 

this works in python3
```python
a=None
>>> if a and a > 100 :
       print("test")
```

python/plugins/db_manager/db_plugins/oracle/info_model.py
troopa81: same comment than above

troopa81 (changes requested):
Thanks for this work!

python/plugins/db_manager/db_plugins/oracle/info_model.py
raulbbarbosa: Not exactly. Row count can be 0 which gives a False value and will not satisfy the condition. It is 'greater than or equal' to be precise.

python/plugins/db_manager/db_plugins/oracle/info_model.py
raulbbarbosa: Here, I guess it can be replaced as you said, since you don't have the '>=', only '>'.

python/plugins/db_manager/db_plugins/oracle/info_model.py
troopa81: Indeed I miss the case where rowCount equals 0. Actually, the way you write it removes the ambiguity in both case. Forget about my comments

troopa81 (approved):
This seems ok to me, LGTM

rouault:
@raulbbarbosa  The automated backport failed. I let you issue manual backport pull requests

PR #34778, opened 2020-02-28
[Bugfix] Prevent crash when deleting invalid pointer

roya0045:
## Description

This is a quick fix to a segfault that Travis encountered during the Qgs relationreferencewidget test for one of my build..

### DIFF
--- a/src/core/qgsfeaturefiltermodel.cpp
+++ b/src/core/qgsfeaturefiltermodel.cpp
@@ -339,8 +339,11 @@ void QgsFeatureFilterModel::updateCompleter()
 
 void QgsFeatureFilterModel::gathererThreadFinished()
 {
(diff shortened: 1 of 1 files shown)


m-kuhn:
Deleting a nullptr is a safe no-op, I'd be surprised to see this fix something.

Can you open an issue with a link to your segfault and ping @troopa81 ?

roya0045:
@m-kuhn the issue occured during the travis test, I did not try to replicate it but this fix seem to have resolved it. I just though it was a potential source of error but I might be mistaked.

I estimated that the error came from there due to this https://cdash.orfeo-toolbox.org/testDetails.php?test=16814015&build=32644 but I might be totally off.

roya0045:
closing, crash still occured on the 32bit test

PR #3036, opened 2016-04-26
Support negative short integer primary keys on PostgreSQL

strk:
See http://hub.qgis.org/issues/14262

NOTE: signed int64 keys would still fail

See also https://github.com/qgis/QGIS/pull/2797

PR #34780, opened 2020-02-28
Fix pg bigint pk not text

### LINKED ISSUE #34077: QGIS casts bigint Primary key field to a text field
Using version of QGIS:
3.10.2-A Coruña
Postgres version 9.6

Although also tested on QGIS 3.4, Qgis 2.18 and postgresql 11

Steps to reproduce:
1) Engage the detailed logging on your postgres instance
2) create a table using a bigint as primary key:
```
create table if not exists
test.test_4
(id bigint primary key,
geometry geometry(POINT, 4326));
```
3) Load this table into QGIS
4) Create some points in the table
5) Identify a point using the identify tool
6) Check your postgresql logging and see:

> 2020-01-28 09:15:40 CET LOG:  statement: DECLARE qgis_350 BINARY CURSOR FOR SELECT st_asbinary("geometry",'NDR'),"id"::text FROM "test"."test_4" WHERE "id"::text='1'

The id field is cast to a text field which causes the index on the primary key to be skipped. For larger tables, this causes a huge slowdown.

This does not happen with primary keys of the type:
- int
- serial
- bigserial

espinafre:
On PostgreSQL tables whose primary keys are of type bigint/bigserial,
QGIS used to cast the primary keys to text, causing the database to do
bigint::text castas and full table scans instead of using the indices,
causing in turn slowness on updates and other queries.

This fix creates a new QgsPostgresPrimaryKeyType, PktInt64. PostgreSQL
don't know about unsigned types, so the PK type PktUint64 loses sense.
Knowing that the PK is an integer type, QGIS won't cast it to text
anymore, thus enabling the correct use of database indices.

Fixes #34077


jef-n:
QGIS uses signed 64bit feature ids.  Positive values for features from the data source and negative values for  feature ids of new unsaved features.  Where possible QGIS uses the actual key from the datasource, but for datasource that have wider keys (eg. full 64bit, bigint, strings, compound keys etc.), we remap the keys to positive features ids.  See also #3036

espinafre:
> QGIS uses signed 64bit feature ids. Positive values for features from the data source and negative values for feature ids of new unsaved features. Where possible QGIS uses the actual key from the datasource, but for datasource that have wider keys (eg. full 64bit, bigint, strings, compound keys etc.), we remap the keys to positive features ids. See also #3036

Thanks for pointing that out. So, PktUint64 stays. The problem with forcing bigint/bigserial keys to PktFidMap is that QGIS casts those to text, generating queries such as:

` UPDATE "myschema"."tb_my_big_table" SET "class_of_feature"='Swamp' WHERE "id"::text='9067684'`

This, on a table with currently 10 million records (polygons), takes 4+ seconds on a very robust PostgreSQL 12 server, with only a single user; when editing a . We chose to use bigserial for the id because we expect this table to grow larger than 2 billion records. With the changes in this PR, the generated queries are as follows:

` UPDATE "myschema"."tb_my_big_table" SET "class_of_feature"='Swamp' WHERE "id"=9067684`
and each query/update takes under 0.2 miliseconds.

elpaso:
I'm afraid that the only robust solution is to check at layer load time if the pk column contains any value <= 0 and use the pk map in that case. 

Ie the check passes, we can probably handle the signed pk as if it was an unsigned one.

m-kuhn:
It should be possible to use aa pk map locally and still send the requests as integer.
Or am I missing something?

jef-n:
> It should be possible to use aa pk map locally and still send the requests as integer.
> Or am I missing something?

Yes, I'd think the main problem is the missing support for `bigint` as type - and that's why it is handled as text. Did anyone try `int8` instead?

espinafre:
> > It should be possible to use aa pk map locally and still send the requests as integer.
> > Or am I missing something?
> Yes, I'd think the main problem is the missing support for `bigint` as type - and that's why it is handled as text. Did anyone try `int8` instead?

On my system int8 gives the same result as bigint (PktInt64 in QGIS parlance). This last PR didn't work because for some reason the user is not allowed to edit values whose PK is negative; positive values on the same table work perfectly well. I'm starting to think that maybe there is something between the interface and the provider that don't take negative int8 FIDs too well. Later this week I'll tidy up my local mess and push it here, so that better minds than mine can take a look and hopefully have better insights.

espinafre:
Well, here is what I've got so far:

1. The original problem is that the QGIS PostgreSQL provider casts bigint/bigserial/int8 primary keys to text, greatly impacting performance on large tables - especially those that by their nature (many records) require such long primary keys.
2. Sandro Santilli (@strk) submitted a PR (#3036) that enabled editing of attributes of features whose primary keys (32 bit integers) are negative, by taking advantage of the fact that FIDs in QGIS are 64 bit wide, and treating those keys as 64 bit FIDs (text).
3. This PR treats bigint/bigserial/int8 primary keys as numbers, which prevents the PostgreSQL provider from casting those to text, enabling the database to make full use of the indexes and primary keys on tables whose PK are of those types.
4. The test testSignedIdentifiers from the suite PyQgsPostgresProvider fails with this PR because the returned FID for loaded features, like strk's patch, is the original integer value, and not the count of features as returned by QgsPostgresSharedData::lookupFid(), which was the expected behavior when the test was written.
4. However, since QGIS thinks that negative FIDs are added by the user after table loading, the user cannot edit features whose PKs are negative bigint: https://github.com/qgis/QGIS/blob/9e5966cf8c079c49109254f5b8a3847e3f00c017/src/core/qgsfeatureid.h#L28 and https://github.com/qgis/QGIS/blob/9e5966cf8c079c49109254f5b8a3847e3f00c017/src/core/qgsvectorlayereditbuffer.cpp#L240 where fid is a qint64.
5. @elpaso suggests to check if there are non-positive values in the pk column at load time, and keep the current implementation of converting it to PktFidMap if that is the case, or going with treating it as an ordinary qint64 otherwise. The difficulty I have with this approach is that the primary key type is determined from the table metadata, before the provider even takes a look at the values themselves: https://github.com/qgis/QGIS/blob/9e5966cf8c079c49109254f5b8a3847e3f00c017/src/providers/postgres/qgspostgresprovider.cpp#L1548 and https://github.com/qgis/QGIS/blob/9e5966cf8c079c49109254f5b8a3847e3f00c017/src/providers/postgres/qgspostgresprovider.cpp#L74 It is not clear to me if I should check the loaded values elsewhere in the code and change the pk type accordingly, and I fear unexpected side effects. I will try and explore this idea further. @strk also had this foresight in his commit message for treating int32 keys (see https://github.com/qgis/QGIS/commit/cbcc929feed532ba916d2155393e571c3e623df3 )
6. @m-kuhn suggested to keep the current implementation of treating qint64 PKs as fidmaps but sending requests to the database as integers, skipping the cast to ::text; @jef-n corroborates the idea. The following quick-and-dirty patch against master (not this PR) results in being able to edit features with positive PKs, but not non-positive:

```
diff --git a/src/providers/postgres/qgspostgresconn.cpp b/src/providers/postgres/qgspostgresconn.cpp
index 20b5702877..6b45925cb4 100644
--- a/src/providers/postgres/qgspostgresconn.cpp
+++ b/src/providers/postgres/qgspostgresconn.cpp
@@ -1631,6 +1631,10 @@ QString QgsPostgresConn::fieldExpression( const QgsField &fld, QString expr )
            .arg( majorVersion() < 2 ? "asewkt" : "st_asewkt",
                  expr );
   }
+  else if ( type == QLatin1String( "int8" ) )
+  {
+         return expr;
+  }
   else if ( type == QLatin1String( "geography" ) )
   {
     return QStringLiteral( "st_astext(%1)" ).arg( expr );
diff --git a/src/providers/postgres/qgspostgresprovider.cpp b/src/providers/postgres/qgspostgresprovider.cpp
index b45743a1db..05d472ba01 100644
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -80,7 +80,7 @@ QgsPostgresProvider::pkType( const QgsField &f ) const
       // (in which case we could use pktUint64)
       // we'll have to use a Map type.
       // See https://github.com/qgis/QGIS/issues/22258
-      return PktFidMap; // pktUint64
+      return PktUint64;
 
     case QVariant::Int:
       return PktInt;
```
It does, however, generate UPDATE statements without casting the PKs to text where I'm able to edit features (that is, for positive values of the PK).

I'm still trying to reason with all this, and any other insights, pointers, corrections will be greatly appreciated. Thanks!

elpaso:
Regarding point 6, looking at the code it seems there is something wrong here: if I get it right, https://github.com/qgis/QGIS/blob/9e5966cf8c079c49109254f5b8a3847e3f00c017/src/providers/postgres/qgspostgresprovider.cpp#L1548 is hit whenever a PK is specified in the URI (the other branch of the `if` is for when the PK needs to be guessed. So, in case the PK is known from the URI it is always treated like a FID map and that is probably wrong, we should get the PK type in that case and act accordingly (but that would not be enough to solve your issue of course).

The other consideration is that given the way QGIS is implemented you cannot use negative keys outside the provider internals (i.e. a `QgsFeature` cannot leave the provider with a negative ID unless they are new features that do not exist outside of the current QGIS editing session).

So, option 6 may work in a few cases (when there are no negative PKs in the existing table and there is a guarantee that other applications will never add rows with negative PKs to the table) and it has the "problem" of requiring an extra query to determine the presence of negative PKs at load time.

Solution 7 may be more easy to implement and looks more robust, I think you should work on that one first, you should be able use the map even when you have negative IDs in the existing table.

In any event, if you do want to continue work on this issue (which is of course very welcome) I think you should start adding a few **failing** test data cases to 
https://github.com/qgis/QGIS/blob/9e5966cf8c079c49109254f5b8a3847e3f00c017/tests/testdata/provider/testdata_pg.sh (maybe create another .sql file for them) and implement the **failing** test cases in https://github.com/qgis/QGIS/blob/9e5966cf8c079c49109254f5b8a3847e3f00c017/tests/src/python/test_provider_postgres.py

With a decent test bed it will be much easier to test any potential solution we can come up with.

espinafre:
Following @elpaso 's recommendations, I've gone with @m-kuhn and @jef-n idea, which is to treat int8 keys as fidmaps internally, but sending those values to the DBMS as bare values, without quoting or casting them to text. My preliminary test went very well, as I could edit fields with both positive and negative IDs, and they were sent to the DBMS as they were originally; also, because we are using the existing fidmap mechanism, we don't have clashes with QGIS internal FID handling. Looks very good to me!

Now I will finish my homework (assigned by @elpaso ) and write those formal tests. Thank you so much!

espinafre:
I shouldn't have rebased my branch before merging, right?...

m-kuhn:
Thanks for the update. Promosing!

There is nothing wrong about rebasing per se. But here, something seems to have gone wrong.

What I would do:

Start a new branch from QGIS master and cherry-pick your commits onto it

E.g. 

```
git checkout master
git checkout -b pg_bigint_pk_no_cast
git cherry-pick 626a45b d10a520 cda6ce7
```

Fingers crossed

roya0045:
That rebase looks wrong.

espinafre:
@roya0045 , yes, it does. This is what I did: 
```
git fetch upstream
git checkout master # from my local repo
git merge upstream/master
git push # to my GitHub fork
git checkout fix_pg_bigint_pk_not_text # this branch
git rebase master
git push
```

and that's how it got so messed up.

roya0045:
@espinafre why didn't you just `git rebase upstream/master` directly in your branch? That always worked fine for me.

espinafre:
@roya0045 because I hadn't thought of that, dumb me :D I'll do that next time, thanks for the tip!

espinafre:
Please see #35162 instead; it contains only the changes that fixes the bug, without the whole merge/rebase mess.

PR #34783, opened 2020-02-29
Fix incorrect rowCount results in layout model

nyalldawson:
Picked up by a modeltest enabled build


nyalldawson:
Ok relation widget test, you've got 2 days to be fixed before you get blacklisted.

roya0045:
@nyalldawson isn't blacklisting the test just brushing a crash under the rug?

PR #34794, opened 2020-03-01
Fix layers are incorrectly shown when theme groups are unchecked 

nyalldawson:
Fixes a serious regression caused by #33099


nirvn:
@nyalldawson , ~~auto~~thanks :wink: 

Glad to see a test case was added too.

nyalldawson:
@nirvn was that a review? 😛

nirvn:
Ya, merge at will.

PR #34796, opened 2020-03-01
Avoid confirmation dialog from Refactor fields when not appropriate

gacarrillor:
A confirmation dialog from Refactor fields keeps showing after rejecting the algorithm dialog.

1st_case

**Suggested solution**: destroy widgets from a Processing dialog as soon as it is rejected.

**Steps to reproduce:** 

1. Open QGIS.
2. Load 2 layers.
3. Open the Refactor Fields dialog.
4. Press the Escape key.
5. Remove the layer that is currently selected in the Layers Panel.
6. You get the confirmation dialog from Refactor fields.

**Another case:**

1. Open QGIS.
2. Open the Refactor Fields dialog.
3. Press the Escape key.
4. Load  layer to QGIS and then remove it.
5. Once more, load a layer (it might be the same as in 4).
6. You get the confirmation dialog from Refactor fields.

These 2 are quite feasible scenarios.

It can be even more confusing when you have more layers in the project and run the algorithm several times: the connections may accumulate, showing several confirmation dialogs, consecutively.

--------------------------------
@arnaud-morvan, does it make sense for you?

### DIFF
--- a/src/gui/processing/qgsprocessingalgorithmdialogbase.cpp
+++ b/src/gui/processing/qgsprocessingalgorithmdialogbase.cpp
@@ -654,6 +654,15 @@ void QgsProcessingAlgorithmDialogBase::setInfo( const QString &message, bool isE
   processEvents();
 }
 
+void QgsProcessingAlgorithmDialogBase::reject()
+{
+  if ( !mAlgorithmTask )
+  {
+    setAttribute( Qt::WA_DeleteOnClose );
+  }
+  QDialog::reject();
+}
+
 //
 // QgsProcessingAlgorithmProgressDialog
 //
--- a/python/gui/auto_generated/processing/qgsprocessingalgorithmdialogbase.sip.in
+++ b/python/gui/auto_generated/processing/qgsprocessingalgorithmdialogbase.sip.in
@@ -187,6 +187,9 @@ Copies the current log contents to the clipboard.
 Switches the dialog to the parameters page.
 %End
 
+    virtual void reject();
+
+
   protected:
 
     virtual void closeEvent( QCloseEvent *e );
--- a/src/gui/processing/qgsprocessingalgorithmdialogbase.h
+++ b/src/gui/processing/qgsprocessingalgorithmdialogbase.h
@@ -234,6 +234,8 @@ class GUI_EXPORT QgsProcessingAlgorithmDialogBase : public QDialog, private Ui::
      */
     void showParameters();
 
+    void reject() override;
+
   protected:
 
     void closeEvent( QCloseEvent *e ) override;


nyalldawson (commented):
I think this is just hiding the real issue. The underlying cause is that the widget is being deleted with the dialog, which is a leak...

gacarrillor:
Should we just do this in `qgsprocessingalgorithmdialogbase.cpp`?

```
void QgsProcessingAlgorithmDialogBase::reject()
{
  if ( mMainWidget )
  {
    mMainWidget->deleteLater();
  }
  done(QDialog::Rejected);
}
```

gacarrillor:
@nyalldawson: Changed the suggested solution: now the main widget is deleted as soon as the algorithm dialog is rejected.

src/gui/processing/qgsprocessingalgorithmdialogbase.cpp
```diff
@@ -592,6 +592,15 @@ void QgsProcessingAlgorithmDialogBase::setInfo( const QString &message, bool isE
   processEvents();
 }
 
+void QgsProcessingAlgorithmDialogBase::reject()
+{
+  if ( mMainWidget )
```
nyalldawson: I'm not sure about this fix -- mMainWidget should already be getting deleted along with the dialog, because it's parented to the dialog in setMainWidget. Possibly what's actually happening here is that the whole dialog is being leaked on cancel, not just the widget.

What happens if you put a

    setAttribute( Qt::WA_DeleteOnClose );

in reject instead?

gacarrillor: It works smoothly! 
Shall I replace the previous commit by a new one, perhaps calling `setAttribute( Qt::WA_DeleteOnClose );` in the constructor?

nyalldawson: Not in the constructor. It gets complex, but if you close the dialog WHILE an algorithm is running in the background then we don't want to delete the dialog. Rather we hide it away until the algorithm execution is finished, just in case the user clicks the running task and wants to see the log and detailed progress reports.

So we should only delete the dialog on close while its not running (hence putting this in the reject override)

nyalldawson: Although -- you better double check the reject **isn't** called while the algorithm is executing...!

gacarrillor: Alright, then https://github.com/qgis/QGIS/pull/34796/commits/52bf7a07b36570e88d124316ad66d9ff7aef015c sets the `Qt::WA_DeleteOnClose` in reject(), only if there is no task running.

I've tested that this successfully solves the initial issue removing unexpected refactor fields confirmation dialogs.

src/gui/processing/qgsprocessingalgorithmdialogbase.cpp
```diff
@@ -592,6 +592,14 @@ void QgsProcessingAlgorithmDialogBase::setInfo( const QString &message, bool isE
   processEvents();
 }
 
+void QgsProcessingAlgorithmDialogBase::reject()
+{
+  if ( !mAlgorithmTask ){
+    setAttribute( Qt::WA_DeleteOnClose );
+  }
+  done( QDialog::Rejected );
```
nyalldawson: I think this should be `QDialog::reject()`

gacarrillor: Fixed, rebased and pushed in https://github.com/qgis/QGIS/pull/34796/commits/0bb5c6f0a71520fdb6844789b28ee9ea9a050ec4

gacarrillor:
Fixed conflicts with master, rebased and pushed.

nyalldawson:
@gacarrillor I'm going to hold off merging this for a bit -- i've got an incoming PR where I think I've addressed deeper memory issues in this dialog, and I'm curious to see if it avoids the need for this fix too.

gacarrillor:
@nyalldawson Ok, when it's ready, just let me know if you'd like me to test the issue which this PR stems from.

nyalldawson:
Can you please test again with master and see if the underlying issues have been fixed by recent C++ ports?

gacarrillor:
Sure, I just tested it on current master and got an exception. Here is the issue: https://github.com/qgis/QGIS/issues/35516

nyalldawson (approved):

gacarrillor:
@nyalldawson, can we backport it to QGIS 3.10.x?

PR #34807, opened 2020-03-01
DB Manager does not recognize if PostIGIS-Layer is Table, View or Materialized View

### LINKED ISSUE #34806: DB Manager does not recognize if PostIGIS-Layer is Table, View or Materialized View
Since QGIS 3.10 the DB-Manager does not recognize any more, if a PostGIS-Layer is a Table, a View or a Materialized View. In QGIS 3.4 LTR it still works.
This makes problems when i try to add some views or materialized views to the Canvas.

I figured out, that this is related to the the SQL-Query in 
*PostGISDBConnector.getVectorTables()*
from 
*qgis.python.plugins.db_manager.postgis*

To execute the query, since qgis 3.10 
*QgsAbstractDatabaseProviderConnection.ExecuteSQL(sql)*
is used, whereas before the "normal"
*self.connection.cursor().execute(sql)* 
was used.

The QgsAbstractDatabaseProviderConnection.ExecuteSQL(sql)
returns "NULL" for the column "relclass".

A simple example to check in the qgis python console:

```sql
md = QgsProviderRegistry.instance().providerMetadata('postgres')
core_connection = md.findConnection('reghan')
sql = "SELECT relname, relkind FROM pg_class c, pg_namespace n WHERE n.oid = c.relnamespace AND n.nspname = 'apl' AND c.relkind IN ('t', 'v', 'm')"
tables = core_connection.executeSql(sql)
```
this returns a list of layers in the PostGIS-Schema *apl*.
The second Column, that should represent the  *relkind* is returned as NULL.

The problem can be fixed by casting the column relkind to text in the sql-query:

```sql
sql = "SELECT relname, relkind::text FROM pg_class c, pg_namespace n WHERE n.oid = c.relnamespace AND n.nspname = 'apl' AND c.relkind IN ('t', 'v', 'm')"
tables = core_connection.executeSql(sql)
```
I'll make a pull request to fix this bug in QGIS.

MaxBo:
Fixes #34806

cast relkind to text to fix bug with QgsAbstractDatabaseProviderConnection.ExecuteSQL to return NULL instead of the relkind

### DIFF
--- a/python/plugins/db_manager/db_plugins/postgis/connector.py
+++ b/python/plugins/db_manager/db_plugins/postgis/connector.py
@@ -482,7 +482,7 @@ class PostGisDBConnector(DBConnector):
 
         # discovery of all tables and whether they contain a geometry column
         sql = u"""SELECT
-                                                cla.relname, nsp.nspname, cla.relkind,
+                                                cla.relname, nsp.nspname, cla.relkind::text,
                                                 pg_get_userbyid(relowner), cla.reltuples, cla.relpages,
                                                 pg_catalog.obj_description(cla.oid),
                                                 """ + geometry_fields_select + """
@@ -554,7 +554,7 @@ class PostGisDBConnector(DBConnector):
 
         # discovery of all tables and whether they contain a raster column
         sql = u"""SELECT
-                                                cla.relname, nsp.nspname, cla.relkind,
+                                                cla.relname, nsp.nspname, cla.relkind::text,
                                                 pg_get_userbyid(relowner), cla.reltuples, cla.relpages,
                                                 pg_catalog.obj_description(cla.oid),
                                                 """ + raster_fields_select + """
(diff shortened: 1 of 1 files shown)


nyalldawson:
@elpaso possibly the underlying cause of this is char type handling in the connection class?

MaxBo:
@nyalldawson 
char type handling actually seems to be the underlying problem.
```sql
md = QgsProviderRegistry.instance().providerMetadata('postgres')
core_connection = md.findConnection('reghan')
sql = "SELECT 't'::pg_catalog.char, 't'::text, 't'::varchar"
print(core_connection.executeSql(sql))
```
returns

```sql
[[NULL, 't', 't']]
```

elpaso:
@nyalldawson yep, PR done with test.

elpaso:
@MaxBo thank you for the PR! 

There was an issue in the types conversion that fixes the issue without requiring an explicit type cast, see:

Superceeded by: https://github.com/qgis/QGIS/pull/34815

Issue #34798, opened 2020-03-01
Available WMS layers are not displayed in QGIS Browser


pathmapper:
**Describe the bug**
Available WMS layers are not displayed in the QGIS Browser.

**How to Reproduce**
After adding a WMS with the following URL to the browser with QGIS 3.12.0 the available WMS layers are not displayed in the browser (there is only 1 layer displayed but there are more):
https://www.wms.nrw.de/geobasis/wms_nw_dop

![grafik](https://user-images.githubusercontent.com/20856381/75623020-6414b980-5ba6-11ea-965c-a20230d2b5db.png)

![grafik](https://user-images.githubusercontent.com/20856381/75622898-72aea100-5ba5-11ea-8da0-5c66b4629d68.png)

**QGIS and OS versions**

QGIS version | 3.12.0-București | QGIS code revision | 409bcf8eaa
-- | -- | -- | --
Compiled against Qt | 5.11.3 | Running against Qt | 5.11.3
Compiled against GDAL/OGR | 2.4.0 | Running against GDAL/OGR | 2.4.0
Compiled against GEOS | 3.7.1-CAPI-1.11.1 | Running against GEOS | 3.7.1-CAPI-1.11.1 27a5e771
Compiled against SQLite | 3.27.2 | Running against SQLite | 3.27.2
PostgreSQL Client Version | 11.6 (Debian 11.6-0+deb10u1) | SpatiaLite Version | 4.3.0a
QWT Version | 6.1.4 | QScintilla2 Version | 2.10.4
Compiled against PROJ | 5.2.0 | Running against PROJ | Rel. 5.2.0, September 15th, 2018
OS Version | Debian GNU/Linux 10 (buster)
Active python plugins | ClipMultipleLayers; quick_map_services; scriptrunner3; LAStools; canvas_clipper; savedialog; loadthemall; kmltools; MapsPrinter; mmqgis; pluginbuilder3; vector_tiles_reader; layerGroupFilter; plugin_reloader; realcentroid; mask; MetaSearch; db_manager; processing

**Additional context**

- With QGIS 3.10.1 the WMS layers are displayed. 

- With QGIS 3.12.0 it is possible to add the WMS layers via the "Data Source Manager":
![grafik](https://user-images.githubusercontent.com/20856381/75622864-2e230580-5ba5-11ea-8c5b-247f1b6b3d80.png)

- https://github.com/qgis/QGIS/issues/34692 could be related.

gioman:
[...]

Issue #34692, opened 2020-02-26
Traiiling "?" is removed from a WMS URL so the QGIS browser does not show anymore the content of groups


Thsi0815:
New description: https://github.com/qgis/QGIS/issues/34692#issuecomment-591878120

##############

Old description:

![Annotation 2020-02-26 133422](https://user-images.githubusercontent.com/61503307/75345348-da5ea680-589c-11ea-9d3c-77f49abbb12c.png)

The new qgis version: 3.1.2 does not accept any special characters like "?" in the URL-address of a WMS-Source. After a refresh of the working wms-source the url and the source does no more work.

gioman:
@Thsi0815 works ok here, can you provide an URL?

Thsi0815:
http://www.geoportal.rlp.de/mapbender/php/wms.php?layer_id=61680&VERSION=1.1.1&withChilds=1Am 26.02.20, 18:33 schrieb Giovanni Manghi <notifications@github.com>:

  @Thsi0815 works ok here, can you provide an URL?

gioman:
> http://www.geoportal.rlp.de/mapbender/php/wms.php?layer_id=61680&VERSION=1.1.1&withChilds=1Am 26.02.20, 18:33 schrieb Giovanni Manghi <notifications@github.com>:

@Thsi0815 works fine here (3.10.3 on Windows as also 3.12.0)

![Screenshot_20200226_190614](https://user-images.githubusercontent.com/1951107/75378151-1ad51900-58cb-11ea-9778-fc6502f0f348.png)

Thsi0815:
Sorry, but on my installation I could not insert any kind of special character in the URL field.Tomorrow I will setup and test the new version once again.I will send you the result tomorrowThank you...  :-)Am 26.02.20, 20:07 schrieb Giovanni Manghi <notifications@github.com>:

   
   http://www.geoportal.rlp.de/mapbender/php/wms.php?layer_id=61680&VERSION=1.1.1&withChilds=1Am 26.02.20, 18:33 schrieb Giovanni Manghi notifications@github.com: 
   
  @Thsi0815 works fine here (3.10.3 on Windows as also 3.12.0)

gioman:
> Sorry, but on my installation I could not insert any kind of special character in the URL field.Tomorrow I will setup and test the new version once again.I will send you the result tomorrowThank you..

I copy/pasted your URL, are you writing it manually?

Thsi0815:
Good morning

[...]

PR #34808, opened 2020-03-01
Added a fix for the missing WMS category layers

Samweli:
Provides a fix for https://github.com/qgis/QGIS/issues/34692 and https://github.com/qgis/QGIS/issues/34798, these issues occurred because of the new way of displaying wms collection layers and layers on the browser was not torelant with non-conforming wms services as the old approach was, this PR adds a fix for that.

The new way of these layers separates collection layers and data layers, this enable simple refresh of the wms layers compared to the old way of treating the collection layers and data layers as one.

I'm currently working on a fix in a different PR to enable adding to the layers panel the new collection layers.

### DIFF
--- a/src/providers/wms/qgswmsdataitems.cpp
+++ b/src/providers/wms/qgswmsdataitems.cpp
@@ -98,7 +98,7 @@ QVector<QgsDataItem *> QgsWMSConnectionItem::createChildren()
       QString pathName = layerProperty.name.isEmpty() ? QString::number( layerProperty.orderId ) : layerProperty.name;
       QgsDataItem *layer = nullptr;
 
-      if ( layerProperty.name.isEmpty() )
+      if ( layerProperty.name.isEmpty() || !layerProperty.layer.isEmpty() )
         layer = new QgsWMSLayerCollectionItem( this, layerProperty.title, mPath + '/' + pathName, capabilitiesProperty, uri, layerProperty );
       else
         layer = new QgsWMSLayerItem( this, layerProperty.title, mPath + '/' + pathName, capabilitiesProperty, uri, layerProperty );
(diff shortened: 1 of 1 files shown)


pathmapper:
@Samweli you're saying that https://www.wms.nrw.de/geobasis/wms_nw_dop is non-conforming to the spec?

If yes, could you please specify which part of this service is non-conforming?

timlinux:
@pathmapper hi

Actually we went and re-read the spec and it is a little confusing:

> 7.2.4.6.3 Name
> If, and only if, a layer has a <Name>, then it is a map layer that can be requested by using that Name in the LAYERS parameter of a GetMap request. A Layer that contains a <Name> element is referred to as a “named layer” in this International Standard. If the layer has a Title but no Name, then that layer is only a category title for all the layers nested within. A server that advertises a Layer containing a Name element shall be able to accept that Name as the value of LAYERS argument in a GetMap request and return the corresponding map. A client shall not attempt to request a layer that has a Title but no Name.
A server shall throw a service exception (code="LayerNotDefined") if an invalid layer is requested.
A containing category itself may include a Name by which a map portraying all of the nested layers can be requested at once. For example, a parent layer "Roads" may have children “Interstates” and “State Highways” and allow the user to request either child individually or both together.
The Name is not inherited by child Layers.

http://portal.opengeospatial.org/files/?artifact_id=14416

The first paragraph indicates that a collection should not have a Name, whilst the last references a containing category which can have a name, so we are a little confused....

pathmapper:
Hi and thanks @timlinux, I read it this way:

A category layer may have a name and if it has a name, this name could be used to request all nested map layers.

If a layer has no name (and only a title), then it is for sure a category layer (and not a map layer).

src/providers/wms/qgswmsdataitems.cpp
```diff
@@ -98,7 +98,7 @@ QVector<QgsDataItem *> QgsWMSConnectionItem::createChildren()
       QString pathName = layerProperty.name.isEmpty() ? QString::number( layerProperty.orderId ) : layerProperty.name;
       QgsDataItem *layer = nullptr;
 
-      if ( layerProperty.name.isEmpty() )
+      if ( layerProperty.name.isEmpty() || !layerProperty.layer.isEmpty() )
```
pathmapper: According to the spec (http://portal.opengeospatial.org/files/?artifact_id=14416, 7.2.4.6.2) 

> A <Title> is mandatory for all layers;

so `!layerProperty.layer.isEmpty()` should always be true, right?

Samweli: > so !layerProperty.layer.isEmpty() should always be true, right?
How will it always be true?

Samweli: @pathmapper if I have correctly understand you, `!layerProperty.layer.isEmpty()` checks if there any nested layers and not titles.

pathmapper: @Samweli ahh right, sorry, I was focused on names and titles...

Samweli: ok @pathmapper

nyalldawson (approved):

PR #34817, opened 2020-03-02
Warn that layers will be lost when overwriting container vector file

### LINKED ISSUE #32823: Avoid accidentally overwriting Geopackage database when creating new layer
When creating a new Geopackage layer and selecting an existing gpkg file, QGIS asks if the gpkg database should be overwritten or if the new layer should be added to the existing database. Although adding the layer is the default, it easily happens that someone accidentally clicks on the wrong button and overwrites an existing database with many layers.
It would be good if QGIS would display an additional dialog box "You are about to overwrite a database containing x layers with y features - are you shure? [Overwrite data | Cancel]" when "overwrite" is selected.

An alternative would be to remove the option to overwrite the database and replace the current dialog by a status message "New Layer is being added to existing database".
This would make overwriting an existing database an explicit action which requires deleting the old gpkg file first.

uclaros:
## Description
Removed the dangerous option to delete the whole container file (geopackage, spatialite etc) when it contains a layer with the same name as the one the user is trying to save.

Fixes  #32823

### DIFF
--- a/src/gui/ogr/qgsvectorlayersaveasdialog.cpp
+++ b/src/gui/ogr/qgsvectorlayersaveasdialog.cpp
@@ -19,6 +19,7 @@
 #include "qgsvectorlayersaveasdialog.h"
 #include "qgsprojectionselectiondialog.h"
 #include "qgsvectordataprovider.h"
+#include "qgsogrdataitems.h"
 #include "qgscoordinatereferencesystem.h"
 #include "qgseditorwidgetfactory.h"
 #include "qgseditorwidgetregistry.h"
@@ -292,7 +293,23 @@ void QgsVectorLayerSaveAsDialog::accept()
         if ( ret == QMessageBox::Cancel )
           return;
         if ( msgBox.clickedButton() == overwriteFileButton )
+        {
+          const QList<QgsOgrDbLayerInfo *> subLayers = QgsOgrLayerItem::subLayers( filename(), format() );
+          QStringList layerList = QStringList();
+          for ( const auto layer : subLayers )
+          {
+            layerList.append( layer->name() );
+          }
+          if ( layerList.length() > 1 && QMessageBox::warning( this,
+               tr( "Overwrite file" ),
+               tr( "This file contains %1 layers that will be discarded!\n"
+                   "The following layers will be permanently lost:\n"
+                   "\n%2" ).arg( QString::number( layerList.length() ), layerList.join( ", " ) ),
+               QMessageBox::Ok | QMessageBox::Cancel,
+               QMessageBox::Cancel ) == QMessageBox::Cancel )
+            return;
           mActionOnExistingFile = QgsVectorFileWriter::CreateOrOverwriteFile;
+        }
         else if ( msgBox.clickedButton() == overwriteLayerButton )
           mActionOnExistingFile = QgsVectorFileWriter::CreateOrOverwriteLayer;
       }
@@ -322,7 +339,23 @@ void QgsVectorLayerSaveAsDialog::accept()
         if ( ret == QMessageBox::Cancel )
           return;
         if ( msgBox.clickedButton() == overwriteFileButton )
+        {
+          const QList<QgsOgrDbLayerInfo *> subLayers = QgsOgrLayerItem::subLayers( filename(), format() );
+          QStringList layerList = QStringList();
+          for ( const auto layer : subLayers )
+          {
+            layerList.append( layer->name() );
+          }
+          if ( layerList.length() > 1 && QMessageBox::warning( this,
+               tr( "Overwrite file" ),
+               tr( "This file contains %1 layers that will be discarded!\n"
+                   "The following layers will be permanently lost:\n"
+                   "\n%2" ).arg( QString::number( layerList.length() ), layerList.join( ", " ) ),
+               QMessageBox::Ok | QMessageBox::Cancel,
+               QMessageBox::Cancel ) == QMessageBox::Cancel )
+            return;
           mActionOnExistingFile = QgsVectorFileWriter::CreateOrOverwriteFile;
+        }
         else if ( msgBox.clickedButton() == overwriteLayerButton )
           mActionOnExistingFile = QgsVectorFileWriter::CreateOrOverwriteLayer;
         else if ( msgBox.clickedButton() == appendToLayerButton )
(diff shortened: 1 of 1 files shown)


roya0045:
I wouldn't vouch for removing this option, the only thing that I would do is add a dialog asking for confirmation.

3nids:
To me, it's the very first message which is problematic.
When you select an existing file, it says it is going to be overwritten even when your intention is to add a layer to it.

I think we should have a radio button on top of the dialog, letting the choice between: "create/overwrite GPGK" and "add layer to existing GPKG".

That would remove any confusion, and bonus, you won't get a warning when you select an existing file in the dialog.

uclaros:
> When you select an existing file, it says it is going to be overwritten even when your intention is to add a layer to it.

No, it does not. It just saves the new layer in the existing geopackage.

Imagine the exact equivalent to the *Overwrite File* option when saving for example a spreadsheet:
You get a dialog saying *File spreadsheet.xls already exists. Do you want to overwrite it, or do you want to overwrite it and then also delete all other files (regardless of their format) that exist in the same folder?*
This does not exist in any software because it just makes no sense. If the user wants to delete the folder (or geopackage) he can do it explicitly in the file selection dialog while selecting the filename and seeing that it already exists.

An ideal solution for me would be to be able to browse the geopackage contents like it was a folder, in the file selection dialog. Then the user would already be aware of the geopackage contents and in a better position to decide the name of the layer to choose or if he wants to delete the geopackage and all it's contents. But that is probably too much work.

nirvn:
-1 to remove a feature here. If wording is not clear enough, we can update that. I personally use file overwrite quite often. 

Not offering this option also means ppl might end up adding a layer to a preexisting gpkg without realizing it, and potentially share sensitive data by mistake.

uclaros:
> Not offering this option also means ppl might end up adding a layer to a preexisting gpkg without realizing it, and potentially share sensitive data by mistake.

This may already be happening, since the dialog appears only on layer collision, not on filename collision.

3nids:
> No, it does not. It just saves the new layer in the existing geopackage.

Then this is really misleading:

Screen Recording 2020-03-02 at 15 07 57

uclaros:
Is this an OSX thing? In my build this does not pop up, neither with the *New Geopackage Layer* nor with the *Save Vector Layer As...* dialogs.

QGIS version | 3.13.0-Master | QGIS code revision | 09fb04d732
-- | -- | -- | --
PostgreSQL Client Version | 12.2 (Debian 12.2-1) | SpatiaLite Version | 4.3.0a
OS Version | Debian GNU/Linux bullseye/sid | This copy of QGIS writes debugging output.
Active python plugins | firstaid; plugin_reloader; openlayers_plugin; QuickWKT; StreetView; FreehandRasterGeoreferencer; processing; db_manager

nyalldawson:
I'm with @nirvn here, I'm -1 to removing the option and think we just need extra confirmation/warning prompts instead.

uclaros:
Added a warning with the layers that the user is about to say goodbye to. It should not spawn when the file has only one layer (ie Overwrite file and overwrite layer will have the same effect) so the user does not get used to it and clicks OK without thinking. It only spawns when other layers are going to be deleted.
I also have the following thoughts that bother me:
1. Is my use of `QgsOgrLayerItem::subLayers()` proper for the occasion or should I have looked elsewhere?
2. Am I wrong or the file has been opened three times, once for `QgsVectorFileWriter::editionCapabilities()`, once for `QgsVectorFileWriter::targetLayerExists()` and once for `QgsOgrLayerItem::subLayers()`? Should this be optimized?
3. I don't like that I had to copy paste the same code in 3 places in the same file. Wouldn't it be better to create a single message box for all capabilities cases and only add the proper buttons depending on the case?
4. This has to be applied to at least `qgsnewgeopackagelayerdialog.cpp` and `qgsnewspatialitelayerdialog.cpp`, maybe more places. There's a lot of duplicate code there that can lead to inconsistencies (like the new shapefile dialog has radio buttons for Z/M values, while the geopackage and spatialite have chackboxes). Wouldn't it be better to have a `QgsNew*LayerDialog` that handles such stuff centrally and subclass it for each format? Even more, have a base class that both `QgsVectorLayerSaveAsDialog` and `QgsNew*LayerDialog` could use?

uclaros:
This is what the new warning dialog looks like:
Screenshot_20200322_184906

roya0045:
> This is what the new warning dialog looks like:
> Screenshot_20200322_184906

I can get behind this.

roya0045:
@uclaros what happens when the layer list is too big, does the ok gets out of the screen or can there be a text box with a scroll bar in that case?

uclaros:
This is a generic `QMessageBox::warning` so it will probably grow with the contents and push the buttons out of the screen, indeed. Maybe I could put the layers list into `detailedText` which requires pressing a *show details* button and display a scrollable text box.

uclaros:
I restructured the code to avoid all the duplicate code in each capability case.

As for the overwrite whole file confirmation, it now looks like this:
Peek 2020-04-04 17-32

It still only spawns when there is more than one layer on the file to be overwritten.

nyalldawson:
@uclaros I like that!

(off topic: why are your layers named .shp? :smile: )

uclaros:
> (off topic: why are your layers named .shp? smile )

@nyalldawson They were merged in a geopackage with some ad-hoc ogr2ogr bash script but didn't bother to trim the extension! Apparently Greece still lives in the age of .shp and .mdb!

src/gui/ogr/qgsvectorlayersaveasdialog.cpp
```diff
@@ -383,6 +346,40 @@ void QgsVectorLayerSaveAsDialog::accept()
     }
   }
 
+  if ( mActionOnExistingFile == QgsVectorFileWriter::AppendToLayerNoNewFields )
+  {
+    if ( QgsVectorFileWriter::areThereNewFieldsToCreate( filename(), layername(), mLayer, selectedAttributes() ) )
+    {
+      if ( QMessageBox::question( this,
+                                  tr( "Save Vector Layer As" ),
+                                  tr( "The existing layer has different fields. Do you want to add the missing fields to the layer?" ) ) == QMessageBox::Yes )
```
nyalldawson, line 355:
```suggestion
                                  tr( "The existing layer has additional fields. Do you want to add the missing fields to the layer?" ) ) == QMessageBox::Yes )
```

src/gui/ogr/qgsvectorlayersaveasdialog.cpp
```diff
@@ -383,6 +346,40 @@ void QgsVectorLayerSaveAsDialog::accept()
...
+      {
+        mActionOnExistingFile = QgsVectorFileWriter::AppendToLayerAddFields;
+      }
+    }
+  }
+  else if ( mActionOnExistingFile == QgsVectorFileWriter::CreateOrOverwriteFile )
+  {
+    const QList<QgsOgrDbLayerInfo *> subLayers = QgsOgrLayerItem::subLayers( filename(), format() );
```
nyalldawson, line 363: This is leaked - the entries need to be deleted

src/gui/ogr/qgsvectorlayersaveasdialog.cpp
```diff
@@ -383,6 +346,40 @@ void QgsVectorLayerSaveAsDialog::accept()
...
+    QStringList layerList = QStringList();
```
nyalldawson, line 364:
```suggestion
    QStringList layerList;
```

src/gui/ogr/qgsvectorlayersaveasdialog.cpp
```diff
@@ -383,6 +346,40 @@ void QgsVectorLayerSaveAsDialog::accept()
...
+    for ( const auto layer : subLayers )
```
nyalldawson, line 365:
```suggestion
    for ( const QgsOgrDbLayerInfo* layer : subLayers )
```

src/gui/ogr/qgsvectorlayersaveasdialog.cpp
```diff
@@ -383,6 +346,40 @@ void QgsVectorLayerSaveAsDialog::accept()
...
+    {
+      layerList.append( layer->name() );
+    }
+    if ( layerList.length() > 1 )
+    {
+      layerList.sort( Qt::CaseInsensitive );
+      QMessageBox msgBox;
+      msgBox.setIcon( QMessageBox::Warning );
+      msgBox.setWindowTitle( tr( "Overwrite file" ) );
```
nyalldawson, line 374:
```suggestion
      msgBox.setWindowTitle( tr( "Overwrite File" ) );
```

nyalldawson (changes requested):
Looking good! Just some minor cleanups left...

nyalldawson:
(I think we should backport this to LTR)

roya0045:
@uclaros nice UI improvement on the dialog!

PR #34818, opened 2020-03-02
Fix crash in relation reference test

### LINKED ISSUE #34779: Crash in RelationReference test
Reported in https://github.com/qgis/QGIS/pull/34778
(Does not always happen, likely a race condition)

Trace here https://cdash.orfeo-toolbox.org/testDetails.php?test=16814015&build=32644

@troopa81 since I haven't seen crashes before on these tests I assume this is related to your recent work.

troopa81:
## Description

Fix flaky test in relation reference widget test. 

QgsFeatureFilterModel needs to wait for the `mGatherer` to be complety finished. Even if the `collectedValues` signal emitting is the last instruction, there is no garantie `run` is over when we delete it (I can reproduce the issue every time by adding a sleep just after the signal emitting).

### DIFF
--- a/src/core/qgsfeaturefiltermodel.cpp
+++ b/src/core/qgsfeaturefiltermodel.cpp
@@ -353,6 +353,8 @@ void QgsFeatureFilterModel::updateCompleter()
 
 void QgsFeatureFilterModel::gathererThreadFinished()
 {
+  // It's possible that gatherer run method is not completely over, so we wait
+  mGatherer->wait();
   delete mGatherer;
   mGatherer = nullptr;
   emit isLoadingChanged();


elpaso:
Just hit this one myself, I believe the source of the issue was https://github.com/qgis/QGIS/commit/f39f5db882cecc85524e7d1a8efb0aa92d064200#diff-bee409cb02c75d3272f877c88ea1e86eR345

troopa81:
@elpaso Thanks for noticing. 

@obrix Is the fix looks fine to you?

obrix:
@troopa81 Thanks for this one. I was trying to fix a crash in updateCompleter (see #34581 ), It made the situation a bit better but it still did not fix the issue completely. Maybe your commit will fix this once and for all. Still not convinced that everything make sense in the way the thread is handled here though.

troopa81:
We have a talk with @obrix on the subjet. 

We think the first issue was not due to a signal order but because of the latency between the emitting of `collectedValues` and `finished` signal. As a consequence, updateCompleter could have triggered another reload and created another mGatherer object while the finished signal was not emitted and received yet. This ends in mGatherer being deleted while updateCompleter needs it.

The fix of @obrix fix this and this PR adds a wait to be sure the thread is completely finished. So it should be fine now.

PR #34819, opened 2020-03-02
Add missing attributes at QgsAttributeEditorContext creation. 

obrix:
Should fix some asserts and crashs. I came across this bug when using multiple edit tool in the digitizing tool bar. An assert was raise in QgsMapToolAdvancedDigitizing::QgsMapToolAdvancedDigitizing line 28 because the  advancedDigitizingDockWidget pointer was null. The mapCanvas and vectorlayertools were also missing.

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -9268,6 +9268,9 @@ void QgisApp::modifyAttributesOfSelectedFeatures()
   QgsFeature f;
   QgsAttributeEditorContext context;
   context.setAllowCustomUi( false );
+  context.setVectorLayerTools( mVectorLayerTools );
+  context.setCadDockWidget(mAdvancedDigitizingDockWidget);
+  context.setMapCanvas(mMapCanvas);
 
   QgsAttributeDialog *dialog = new QgsAttributeDialog( vl, &f, false, this, true, context );
   dialog->setMode( QgsAttributeEditorContext::MultiEditMode );
@@ -9567,6 +9570,7 @@ void QgisApp::selectByForm()
   QgsAttributeEditorContext context;
   context.setDistanceArea( myDa );
   context.setVectorLayerTools( mVectorLayerTools );
+  context.setCadDockWidget(mAdvancedDigitizingDockWidget);
   context.setMapCanvas( mMapCanvas );
 
   QgsSelectByFormDialog *dlg = new QgsSelectByFormDialog( vlayer, context, this );
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -9269,8 +9269,8 @@ void QgisApp::modifyAttributesOfSelectedFeatures()
   QgsAttributeEditorContext context;
   context.setAllowCustomUi( false );
   context.setVectorLayerTools( mVectorLayerTools );
-  context.setCadDockWidget(mAdvancedDigitizingDockWidget);
-  context.setMapCanvas(mMapCanvas);
+  context.setCadDockWidget( mAdvancedDigitizingDockWidget );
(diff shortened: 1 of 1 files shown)


elpaso:
This is just another chapter in the saga, have you considered adding a constructor that takes the necessary elements and deprecate the default constructor?

obrix:
@elpaso Yes, I can delete the default constructor to ensure the initialization of these attributes, but it requires the modification of a lot of base classes constructors which rely on it. I am looking into it.

elpaso:
@obrix while you are working on that class,  perhaps you can have some good ideas about the `parentContext()`: it is dangerous, because it is potentially set from something allocated on the stack that goes out of scope, I would probably recommend to take a copy of it.

m-kuhn:
If we are going to run QGIS without a map canvas (as has been discussed recently) I guess it should be possible to create a context without those properties set. It might be better to adjust the widgets themselves to be able to work in a reduced mode if map canvas et al are missing?

nyalldawson:
I agree with @m-kuhn here -- the widgets themselves need to gracefully handle having no canvas available

elpaso:
Yeah, good point. I guess we not to add a lot of checks in the widgets.

src/gui/qgsattributeeditorcontext.h
```diff
@@ -103,6 +103,14 @@ class GUI_EXPORT QgsAttributeEditorContext
       Q_ASSERT( parentContext.vectorLayerTools() );
     }
 
+    ~QgsAttributeEditorContext()
+    {
+      if ( mParentContext )
+      {
+        delete mParentContext;
```
elpaso: Thank you! 
Is it possible to use a `std::unique_ptr` instead of a raw pointer? We are trying to get rid of `new` and `delete` where possible.

m-kuhn: - Are we sure we don't care about parent state changes at all and can just hold a copy?
- If yes, could we store by value instead of pointer?
- If no, should we rather switch to a QObject (guarded pointer) based approach instead?

obrix: Actually after more tests I am not sure about the first point. I've got several crashes appearing with the last commit instantiating a new QgsAttributeEditorContext as the parent so we may be stuck with the previous implementation.

Using a std::unique_ptr is not convenient here (pointer on same class type + copy constructor deletion). Also we cannot store directly a value because it's the same type (does not compile).

I may be able to spend more time to think about this later, but right now the first commit still fix the main crash.

obrix:
@elpaso About depecrating / deleting the default constructor : It leads to a chain of necessary changes and makes mandatory the addition of a new QgsAttributeEditorContext parameter to the constructor of tens of objects (widgets, widgetwrappers, widgetfactories, up to the constructor of QgsDualView). Another solution is to store the context by pointer instead of by value everywhere. I think this is a pretty major change here, may be out of the scope of this PR ?

elpaso:
@obrix yes, I didn't mean that the changes were required in this PR, please go ahead with the original commit.

I just wanted to ping you about the potential issues with parentContext() since I've also been working on this area of code during the last few days and maybe you had some good ideas about making it safer. We can move this topic to a different PR, no problem.

src/gui/qgsattributeeditorcontext.h
m-kuhn: Totally agree! :+1:

obrix:
@elpaso Okay thank you so I'll just remove the last commit which seems to cause some issues possibly related to parent state changes as discussed with @m-kuhn . I agree with you as to the fact that this topic should probably be addressed later as the architecture of QgsAttributeEditorContext and how it is handled seems a bit odd to me.

PR #34825, opened 2020-03-02
Db manager add foreign table support

MaxBo:
I wrote a Pull Request  for #28018
It adds support for Foreign Tables in Postgresql-Databases to the DB-Manager.
They are handled now as normal views.
Instead of a View-Definition, in the Table Info some information on the Foreign Data Wrapper, the Foreign Server and the Foreign Table are displayed.
To work, it requires the Bugfix #34815.

I don't know how to add a test case to the Pull Request (adding a FDW/Foreign Server and a ForeignTable to a test database, testing, if they are correctly found by the db manager). I would be happy for a hint how to add such a test.


elpaso:
Honestly any time I see a new feature in DB manager I'm not very happy: we are trying to kill it and replace it with a robust C++ implementation based on core C++ classed and the QGIS data browser.

So, the question becomes: is this a feature that only affects postgres or is there any chance to abstract the feature and integrate it in QgsAbstractDatabaseProviderConnection ?

For example, if abstracted, it maybe used for relation discovery in core?

MaxBo:
Hi @elpaso, i am only familiar with ForeignDataWrappers/Foreign Tables in Postgresql/PostGIS-Databases and use them quite often in order to reduce redundant data spreading over different databases or to integrate Data from different sources into a postgresql-database. Besides the postgres_fdw there datawrappers for oracle, mssql or any OGR-datasource (ogr_fdw). In pg_class they are marked as relkind='f' instead of 'v' for a normal view inside the same database.

I don't know how the possibility to integrate external datasources works in Oracle or MSSQL-Databases or other Datasources, and how it is implemented there, so it could then be adressed by the QgsAbstractDatabaseProviderConnection.

In the QGIS data browser, in a PostGIS-Connection, a ForeignTable with a GeometryColumn actually is displayed in the same way as Views or Materialized Views. Somehow, also a primary key attribute is derived from the data.
Not displayed in the QGIS data browser are Foreign Tables without a Geometry Column.

I am not familiar enough to assess how an abstraction could look like, but a robust implementation in the qgis core would of cause be apreciated.

nyalldawson:
> Honestly any time I see a new feature in DB manager I'm not very happy: we are trying to kill it and replace it with a robust C++ implementation based on core C++ classed and the QGIS data browser.

I tend to agree -- I think we should consider a hard block on new feature development for db manager and put in a requirement that work is done using the connections api instead. Otherwise we'll never reach the ultimate goal of replacing all this python code...

elpaso:
@MaxBo if you are interested in exploring the possibility to integrate this functionality into the connections API, (if I get it right) I guess you should work into the provider QgsPostgresConn class and add support for the wrapped data sources (the new relkind).

Something to be optionally considered is if we want to expose some wrapped datasources through native providers too, for instance: a CSV local datasource might also be exposed as a QGIS delimited text layer.

MaxBo:
Dear @elpaso and @nyalldawson , thanks for your feedback. I had a look at the QgsPostgresConn C++-Code. But as i have only very basic C++-knowledge (only reading C++-Code, no coding, compiling or debugging), i don't feel capable implementing this feature in C++. Si i would be happy if someone else would include this in the future.

elpaso:
Is that ok to close this one? the functionality is now implemented in core.

PR #34827, opened 2020-03-02
[feature] Get current parent form values in child forms

elpaso:
## Description

This PR introduces the possibility to use current values from the "parent" form in filter expressions (for  now only in value-relation widgets, to be used in drill-down filters).

A new "parentForm" scope was added as well as a new set of functions and variables to access the parent from from within an embedded child form. The new functions and variables were modeled on the existing "current_value", "current_feature" etc..

The new functions and variables are also available when the parent form was opened from a new (unsaved, unbuffered) feature, making it easier to create drill-down filters based on the parent's values when adding children from an unsaved parent form.

## Example

qgis-parent-form-value

## New functions and variables

qgis-feature-current_parent_value
qgis-feature-current_parent_geometry
qgis-feature-current_parent_feature

Funded by: **ARPA Piemonte**

### DIFF
--- a/src/gui/editorwidgets/qgshtmlwidgetwrapper.cpp
+++ b/src/gui/editorwidgets/qgshtmlwidgetwrapper.cpp
@@ -82,9 +82,13 @@ void QgsHtmlWidgetWrapper::setHtmlContext( )
   if ( !mWidget )
     return;
 
-  QgsAttributeEditorContext attributecontext = context();
+  const QgsAttributeEditorContext attributecontext = context();
   QgsExpressionContext expressionContext = layer()->createExpressionContext();
   expressionContext << QgsExpressionContextUtils::formScope( mFeature, attributecontext.attributeFormModeString() );
+  if ( attributecontext.parentFormFeature().isValid() )
+  {
+    expressionContext << QgsExpressionContextUtils::parentFormScope( attributecontext.parentFormFeature() );
+  }
   expressionContext.setFeature( mFeature );
 
   HtmlExpression *htmlExpression = new HtmlExpression();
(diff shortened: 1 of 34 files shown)


src/gui/editorwidgets/qgshtmlwidgetwrapper.cpp
```diff
@@ -82,9 +82,13 @@ void QgsHtmlWidgetWrapper::setHtmlContext( )
   if ( !mWidget )
     return;
 
-  QgsAttributeEditorContext attributecontext = context();
+  const QgsAttributeEditorContext attributecontext = context();
   QgsExpressionContext expressionContext = layer()->createExpressionContext();
   expressionContext << QgsExpressionContextUtils::formScope( mFeature, attributecontext.attributeFormModeString() );
+  if ( attributecontext.parentFormFeature().isValid() )
```
m-kuhn, line 88: Hi @elpaso 

Just stumbled upon this pull request.
I'm wondering how this works for the HTML widget. As far as I can tell it is currently the responsibility for each widget implementation to keep the `parentFormFeature` in the context up to date (e.g. https://github.com/qgis/QGIS/pull/34827/files#diff-38bcb929a48491604891779e4ea4dc0dR92-R97) and the HTML widget does not provide this code. Which would mean, the feature here is always invalid? I didn't test it, so I might have missed something.

I'm tempted to discuss an API change proposal here, but I'd first like to be sure that my interpretation of this is correct.

m-kuhn: Nevermind, I indeed missed something. Sorry for the noise.

PR #34831, opened 2020-03-02
Select-deselect layers in check geometries plugin

### LINKED ISSUE #22783: Geometry Checker: Add 'Select all'/'Deselect all' buttons
Author Name: **Reinhard Reiterer** (Reinhard Reiterer)
Original Redmine Issue: [14829](https://issues.qgis.org/issues/14829)

Redmine category:c++_plugins/geometry_checker

---
Subject says it all...

---

- [check_geometries_dialog.png](https://issues.qgis.org/attachments/download/9927/check_geometries_dialog.png) (Reinhard Reiterer)

### LINKED ISSUE #28473: Check Geometries should remember input layers
Author Name: **Lars Ellebjerg** (@Ellebjerg)
Original Redmine Issue: [20653](https://issues.qgis.org/issues/20653)

Redmine category:c++_plugins/geometry_checker

---
When running Check Geometries most of the settings are remembered from previous runs. The choice of layers however is at every new run set to all vector layers in the project, active or not. It would be nice if only the layer from the previous run was active as default. Geometry checking is often an iterative process where several checks are necessary.

### LINKED ISSUE #33380: Please add a Select /deselect all (features) in Check Geometries tool
Please make users able to select deselect all the layers present in the Check Geometries tool dialog

While using Check Geometries on a project with hundreds of layes on TOC, currently the User is not able to select/deselect all the  shapefiles. This  cause the user must deselect all the layers one by one, with potential hundreds of clicks, and at the end of the list of layers it is not at all sure to have deselected all the unnecessary layers.
The introduction of a "Select / Deselect all" (button or flag), would be much appreciated, an also  would drastically reduce the time spent on selection.

I attach an example screencast of my typical project with many layers inside

Check_Geometries_selection

best regards

Antonio

uclaros:
## Description
Added buttons to select-deselect all layers in check geometries core plugin.
Since now selecting all layers is always just one click away, I changed the default status of new layers to unchecked instead of checked.
Also now the dialog does not forget which layers where checked every time a layer is added to or removed from the project.

select_all

I also tried more time than I would like to admit to fit the layers list in a `QSplitter` in order to make it user resizable but QtDesigner had a different opinion, so I just enlarged it a bit. The QSplitter would be great though since scrolling past the end of the layer list scrolls the whole dialog and throws the cursor out of the list...

Fixes #33380
Fixes #28473
Fixes #22783

### DIFF
--- a/src/plugins/geometry_checker/qgsgeometrycheckerresulttab.cpp
+++ b/src/plugins/geometry_checker/qgsgeometrycheckerresulttab.cpp
@@ -106,10 +106,7 @@ QgsGeometryCheckerResultTab::QgsGeometryCheckerResultTab( QgisInterface *iface, 
   ui.tableWidgetErrors->horizontalHeader()->setSortIndicator( 0, Qt::AscendingOrder );
   ui.tableWidgetErrors->resizeColumnToContents( 0 );
   ui.tableWidgetErrors->resizeColumnToContents( 1 );
-  ui.tableWidgetErrors->horizontalHeader()->setSectionResizeMode( 2, QHeaderView::Stretch );
-  ui.tableWidgetErrors->horizontalHeader()->setSectionResizeMode( 3, QHeaderView::Stretch );
-  ui.tableWidgetErrors->horizontalHeader()->setSectionResizeMode( 4, QHeaderView::Stretch );
-  ui.tableWidgetErrors->horizontalHeader()->setSectionResizeMode( 5, QHeaderView::Stretch );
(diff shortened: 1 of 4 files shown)


uclaros:
Most of the result columns could not be resized, one could not read the whole error message nor the coordinates. Fixed on the second commit.

nyalldawson (approved):

nyalldawson:
Nice fix!

Antoviscomi:
@uclaros thanks a lot!

PR #34832, opened 2020-03-03
More model designer c++ ports

nyalldawson:
Ports the arrow item and other model designer code to c++, also minor improvements to the shape and connection point logic for the arrow links in models.


nyalldawson:
Here's the new logic in action:

Peek 2020-03-03 10-27

vs old logic:

Peek 2020-03-03 10-28

alexbruy (approved):
Looking good.

Side note, I'm wondering should we switch back to straight arrows instead of curved ones.

nyalldawson:
> Side note, I'm wondering should we switch back to straight arrows instead of curved ones.

I've wondered the same. I think possibly we should use curved arrows for "data flow" and straight ones to indicate parameter association?

alexbruy:
> I think possibly we should use curved arrows for "data flow" and straight ones to indicate parameter association?

Interesting idea, +1 from me

nirvn:
Blender has a great way to handle connection lines for its equivalent to our model editor.

The lines are straight. If you click on any segment of the line, it creates a node that you can move around to have segmented lines at angles. 

It allows complex models to scale really well.

PR #34834, opened 2020-03-03
codecs qt plugins not in qt5, remove from bundle check

kyngchaos:
## Description
This bundle check will fail because codecs does not exist in Qt5, thus not bundled.  Causes crash when build copy of Qt5 present.


kyngchaos:
oops

Issue #34841, opened 2020-03-03
export dxf improvement


SergeyChern:
1. If text label contains CR LF signs and these signs are replaced to spaces (qgsdxfexport.cpp, function QgsDxfExport::drawLabel(...), strings 2317-2318). Result text is of one line, but label->getHeight() is height of multiline text. 
So, my idea is to calc count of lines in label than divide label's height by this count.
Something like this:
    int lines_count = 1;
    int start_pos = txt.indexOf(QLatin1String("\r\n"));
    while (start_pos >= 0)
    {
  	  start_pos = txt.indexOf(QLatin1String("\r\n"), start_pos + 1);
  	  lines_count++;
    }
    txt.replace(QLatin1String("\r\n"), QLatin1String(" "));
    start_pos = txt.indexOf('\r');
    while (start_pos >= 0)
    {
  	  start_pos = txt.indexOf('\r', start_pos + 1);
  	  lines_count++;
    }
    txt.replace('\r', ' ');
    start_pos = txt.indexOf('\n');
    while (start_pos >= 0)
    {
  	  start_pos = txt.indexOf('\n', start_pos + 1);
  	  lines_count++;
    }
    txt.replace('\n', ' ');
    writeText(dxfLayer, txt, QgsPoint(label->getX(), label->getY()), label->getHeight() / float(lines_count), label->getAlpha() * 180.0 / M_PI, tmpLyr.format().color());

2. I've got AutoCAD error if layer name contains comma.
Please, add replacement of comma to underscore in function
QString QgsDxfExport::dxfLayerName( const QString &name ):
  layerName.replace( ',', '_');

3. Some AutoCAD versions can't process transparency group code 440.
Please, remove code below  or make it optional in function
void QgsDxfExport::writeGroup( const QColor &color, int exactMatchCode, int rgbCode, int transparencyCode ):
  if ( transparencyCode != -1 && color.alpha() < 255 )
    writeGroup( transparencyCode, 0x2000000 | color.alpha() );

gioman:
@SergeyChern could you propose patches with pull requests?

SergeyChern:
> @SergeyChern could you propose patches with pull requests?

Never did it, but tried

gacarrillor:
The second one was already solved, see https://github.com/qgis/QGIS/issues/47381

PR #34843, opened 2020-03-03
Update qgsdxfexport.cpp

SergeyChern:
Issue #34841
1. Recalc height of multiline labels
2. ACAD crashes if layer name contains comma
3. Transparency group code 440 leads to ACAD error

## Description

### DIFF
--- a/src/core/dxf/qgsdxfexport.cpp
+++ b/src/core/dxf/qgsdxfexport.cpp
@@ -150,8 +150,9 @@ void QgsDxfExport::writeGroup( const QColor &color, int exactMatchCode, int rgbC
 
   int c = ( color.red() & 0xff ) * 0x10000 + ( color.green() & 0xff ) * 0x100 + ( color.blue() & 0xff );
   writeGroup( rgbCode, c );
-  if ( transparencyCode != -1 && color.alpha() < 255 )
-    writeGroup( transparencyCode, 0x2000000 | color.alpha() );
+  // transparency group code 440 leads to ACAD error (#34841)
+  //if ( transparencyCode != -1 && color.alpha() < 255 )
+  //  writeGroup( transparencyCode, 0x2000000 | color.alpha() );
 }
 
 void QgsDxfExport::writeGroupCode( int code )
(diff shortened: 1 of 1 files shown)


src/core/dxf/qgsdxfexport.cpp
```diff
@@ -150,8 +150,9 @@ void QgsDxfExport::writeGroup( const QColor &color, int exactMatchCode, int rgbC
 
   int c = ( color.red() & 0xff ) * 0x10000 + ( color.green() & 0xff ) * 0x100 + ( color.blue() & 0xff );
   writeGroup( rgbCode, c );
-  if ( transparencyCode != -1 && color.alpha() < 255 )
-    writeGroup( transparencyCode, 0x2000000 | color.alpha() );
+  // transparency group code 440 leads to ACAD error (#34841)
```
m-kuhn: Is this a QGIS issue or an ACAD issue?

src/core/dxf/qgsdxfexport.cpp
```diff
@@ -2314,8 +2333,9 @@ void QgsDxfExport::drawLabel( const QString &layerId, QgsRenderContext &context,
 
   if ( mFlags & FlagNoMText )
   {
-    txt.replace( QChar( QChar::LineFeed ), ' ' );
-    txt.replace( QChar( QChar::CarriageReturn ), ' ' );
+    // move this code to writeText(...):
+    //txt.replace( QChar( QChar::LineFeed ), ' ' );
+    //txt.replace( QChar( QChar::CarriageReturn ), ' ' );
```
m-kuhn: Better remove old lines, don't comment them

m-kuhn:
@mhugent what is your opinion on this?

src/core/dxf/qgsdxfexport.cpp
nyalldawson: I was wondering the same. I think it's disappointing to limit this functionality because of a 3rd party bug. What about adding an option in the UI for "Safe export for maximum compatibility"?

SergeyChern: Last version of AutoDesc's software "TrueViewer..." crashes on this code. Some old versions of AutoCAD can't process it too. I think to add option to UI is best way.

m-kuhn: It's possible that this code is indeed broken (== it actually is a QGIS issue and we should solve it).
Are you able to check what autodesk itself produces for transparency?

SergeyChern:
Have to be changed

PR #34853, opened 2020-03-04
Fix deprecated method redirection

DelazJ:


src/core/qgsproviderregistry.h
```diff
@@ -187,7 +187,7 @@ class CORE_EXPORT QgsProviderRegistry
      * Returns a new widget for selecting layers from a provider.
      * Either the \a parent widget must be set or the caller becomes
      * responsible for deleting the returned widget.
-     * \deprecated QGIS 3.10 - use QgsGui::providerGuiRegistry()->createDataSourceWidget() instead
+     * \deprecated QGIS 3.10 - use QgsGui::providerGuiRegistry()->sourceSelectProviders(providerKey)[0]->createDataSourceWidget() instead
```
nyalldawson:
```suggestion
     * \deprecated QGIS 3.10 - use QgsGui::sourceSelectProviderRegistry()->createDataSourceWidget() instead
```

is a bit cleaner

nyalldawson:
Can you update the sip bindings here?

DelazJ:
Should be OK. I'm not sure the failure is related.

PR #34854, opened 2020-03-04
 Fix QgsCoordinateReferenceSystem documentation rendering

DelazJ:
with better list display, highlight related methods, add deprecation version
https://qgis.org/pyqgis/master/core/QgsCoordinateReferenceSystem.html
question: does this class create an **invalid** CRS object, as stated in the doc?
ps: there is obviously an issue with rendering of list items that expand on more than one line or paragraph (making the doc indistinct and hard to read/understand)

### DIFF
--- a/python/core/auto_generated/qgscoordinatereferencesystem.sip.in
+++ b/python/core/auto_generated/qgscoordinatereferencesystem.sip.in
@@ -15,6 +15,7 @@
 
 
 
+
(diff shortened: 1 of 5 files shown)


nyalldawson:
> question: does this class create an invalid CRS object, as stated in the doc?

Yes

nyalldawson (approved):

PR #34852, opened 2020-03-03
Port most of the remaining ModelerScene code to c++ 

nyalldawson:
The final bit of the recent Python -> c++ conversion and cleanup for model designer. Now the Python ModelerScene class is a minimal shell implementing the bare necessity required for compatibility with the older deprecated Processing Python API.

PR #34856, opened 2020-03-04
Comments in Processing Models

nyalldawson:
(includes https://github.com/qgis/QGIS/pull/34852)

This allows users to create comments attached to model components (inputs, algorithms or outputs). Comments are shown linked to the associated component, and can be freely moved around the model.

(will be more user-friendly when model components can also be resized :laughing: )

Peek 2020-03-04 16-27

 Funded by Fisel + König

### DIFF
--- a/python/core/auto_generated/processing/models/qgsprocessingmodelchildalgorithm.sip.in
+++ b/python/core/auto_generated/processing/models/qgsprocessingmodelchildalgorithm.sip.in
@@ -35,6 +35,16 @@ should be set to a QgsProcessingAlgorithm algorithm ID.
     virtual QgsProcessingModelChildAlgorithm *clone() const /Factory/;
 
 
+    void copyNonDefinitionPropertiesFromModel( QgsProcessingModelAlgorithm *model );
+%Docstring
+Copies all non-specific definition properties from the the matching component from a ``model``.
+
+This includes properties like the size and position of the component, but not properties
+like the specific algorithm or input details.
+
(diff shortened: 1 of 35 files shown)


3nids:
Nice!
Instead of putting the comment in another box, wouldn't it be possible to show it as collapsible item in the main box or as a tooltip?
Or is there a specific reason to do like this?

nyalldawson:
> Or is there a specific reason to do like this?

Well, the designer is an inherently visual approach to workflows, so I wanted the comments to be immediately and always visible when viewing the model. Otherwise you'd still need to "hunt around" with the mouse to find specific parts of a model...

3nids:
I see your points.
Maybe a global toggle for comments would be interesting?

nyalldawson:
> Maybe a global toggle for comments would be interesting?

that's a good idea, I'll see if I can squeeze that in

olivierdalang:
What would be really nice too (but arguably that's another feature) would be some free named/commented boxes, allowing to better visually organise and comment whole sections of the algorithm.

A few images of inspiration (taken from Grasshopper, the nicest node based algorithm editor I know of) :

image
image
image

nyalldawson:
That's hopefully coming soon too!

nyalldawson:
@3nids I'll defer the "hide comments" action to a follow up PR -- it's definitely coming, but triggered a refactor of some other classes and I'd rather keep that work isolated

alexbruy (approved):

PR #34861, opened 2020-03-04
Allow to open data loader dialog from metasearch

### LINKED ISSUE #34759: MetaSearch 'NoneType' object has no attribute 'addRasterLayer'
I get the python error
'NoneType' object has no attribute 'addRasterLayer'
in function add_to_ows in maindialog.py:757

```
ows_provider = QgsProviderRegistry.instance().createSelectionWidget(stype[2],
                                                                            self)
        service_type = stype[0]

        # connect dialog signals to iface slots
        if service_type == 'OGC:WMS/OGC:WMTS':
            ows_provider.addRasterLayer.connect(self.iface.addRasterLayer)
```

Open MetaSearch Dialog.
Search for service in csw
Select found wms service
Click on Button AddData
Select WMS/WMTS
Python Error

**QGIS and OS versions**
QGIS 3.10.2 and 3.12.0 on Windows.

With QGIS 3.4 I was able to select WMS/WMTS.  WMS Dialog opens without error.

The API says:
QWidget * QgsProviderRegistry::createSelectionWidget
Deprecated:
QGIS 3.10 - use QgsGui::providerGuiRegistry()->createDataSourceWidget() instead

DelazJ:
fixes #34759
Also allow to abort the message box to modify connections list

### DIFF
--- a/python/plugins/MetaSearch/dialogs/maindialog.py
+++ b/python/plugins/MetaSearch/dialogs/maindialog.py
@@ -735,9 +735,11 @@ class MetaSearchDialog(QDialog, BASE_CLASS):
         if sname in keys:  # duplicate found
             msg = self.tr('Connection {0} exists. Overwrite?').format(sname)
             res = QMessageBox.warning(self, self.tr('Saving server'), msg,
-                                      QMessageBox.Yes | QMessageBox.No)
-            if res != QMessageBox.Yes:  # assign new name with serial
+                                      QMessageBox.Yes | QMessageBox.No | QMessageBox.Cancel)
+            if res == QMessageBox.No:  # assign new name with serial
                 sname = serialize_string(sname)
+            elif res == QMessageBox.Cancel:
+                return
 
         # no dups detected or overwrite is allowed
         if caller in ['mActionAddAms', 'mActionAddAfs']:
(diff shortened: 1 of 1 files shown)


python/plugins/MetaSearch/dialogs/maindialog.py
```diff
@@ -748,8 +750,7 @@ def add_to_ows(self):
         self.settings.endGroup()
 
         # open provider window
-        ows_provider = QgsProviderRegistry.instance().createSelectionWidget(stype[2],
-                                                                            self)
+        ows_provider = QgsGui.providerGuiRegistry().sourceSelectProviders(stype[2])[0].createDataSourceWidget()
```
nyalldawson:
```suggestion
        ows_provider = QgsGui.sourceSelectProviderRegistry().createSelectionWidget(stype[2], self, Qt.Widget,  QgsProviderRegistry.WidgetMode.Embedded)
```

Does that work? Should be a bit more resilient because it has checks to ensure that the requested provider exists...

DelazJ: Yes it works. With both (setModal+show) and (setmodality+exec_). Should I revert the changes I made there ( I made them because the dialog was disappearing as soon as opened)?

nyalldawson: Yeah let's revert them. Who knows what side effects that could have..!

PR #34876, opened 2020-03-04
[plugin manager] allow to install stable or experimental

olivierdalang:
## Description

This PR improves the plugin manager to offer the possibility for users to choose between the stable or the experimental version.

This allows to push pre-releases to the repository for power-users to test, while still allowing them to switch back easily to stable versions, without having to untick the global `enable experimental` option.

For users who didn't check the `enable experimental` option, this PR makes no difference.

For users who checked the `enable experimental` options, there are now two install/upgrade/downgrade buttons when there's both a stable and an experimental release.

Below are some before/after screenshots for a typical plugin (from [this dummy repository](https://raw.githubusercontent.com/olivierdalang/debug-qgis-plugin-repository/master/plugins.xml)).

I still need to test behaviour a bit more (hence draft pull request), but would be interested in collecting feedback for both for UX and code already.

Cheers !

### Plugin not yet installed

Before PR
e_before_not_installed

After PR
e_after_not_installed

### Experimental version installed

Before PR
e_before_experimental_installed

After PR
e_after_experimental_installed

### Stable version installed

Before PR : not possible (unless you uncheck show experimental versions)

After PR
e_after_stable_installed

### DIFF
--- a/python/pyplugin_installer/installer.py
+++ b/python/pyplugin_installer/installer.py
@@ -220,14 +220,19 @@ class QgsPluginInstaller(QObject):
                 "installed": plugin["installed"] and "true" or "false",
                 "available": plugin["available"] and "true" or "false",
                 "status": plugin["status"],
+                "status_exp": plugin["status_exp"],
                 "error": plugin["error"],
                 "error_details": plugin["error_details"],
                 "experimental": plugin["experimental"] and "true" or "false",
                 "deprecated": plugin["deprecated"] and "true" or "false",
                 "trusted": plugin["trusted"] and "true" or "false",
                 "version_available": plugin["version_available"],
+                "version_available_stable": plugin["version_available_stable"] or "",
+                "version_available_experimental": plugin["version_available_experimental"] or "",
                 "zip_repository": plugin["zip_repository"],
                 "download_url": plugin["download_url"],
+                "download_url_stable": plugin["download_url_stable"],
+                "download_url_experimental": plugin["download_url_experimental"],
                 "filename": plugin["filename"],
                 "downloads": plugin["downloads"],
                 "average_vote": plugin["average_vote"],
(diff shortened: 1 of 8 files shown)


elpaso:
Nice feature!

Saijin-Naib:
What's the behavior for Upgrade/Upgrade all when "Experimental" is enabled?

Does it default to Upgrade/Upgrade all always doing Experimental?

Is the stable/experimental flag set per-plugin? Is there a way to globally over-ride/reset?

olivierdalang:
> What's the behavior for Upgrade/Upgrade all when "Experimental" is enabled?
> Is the stable/experimental flag set per-plugin? Is there a way to globally over-ride/reset?

@Saijin-Naib 
That's not implemented yet, but the idea is to use the currently installed plugin to determine if that plugin is in the "stable" or "experimental" channel. When using upgrade all, you'd get the highest stable version unless you were already using an experimental version, in which case you'd get the highest version.

I didn't think of a global override/reset. To me it seems upgrading all plugins in their respective channel should suit most of use cases, but please let me know if there's a scenario I didn't think of.

m-kuhn:
Nice!

I really like the announcement when you have experimental installed. I would probably hide the "This plugin has an experimental version available" banner on the other hand. It almost looks like an announcement for something you *really* should want to install, while I think it should be rather be a hidden option which a curious advanced user/insider can find by looking closely.

nyalldawson:
Nice work!

Really minor quibble, but could you unify the widths of those buttons along the bottom?


image

olivierdalang:
Hi !

Here's another iteration :

- replace the two buttons by just one button with a dropdown when there's both an experimental and a stable version available (the default action depends on the current "channel", so that you only need the dropdown button if you want to switch "channel"). This makes the default action more prominent and takes less screen space.
- `upgrade all` implemented (as described above) and `upgradeable tab` displays plugin accordingly
- display a banner when a plugin is installed as experimental, rather than "this is an experimental plugin", as we now more talk about experimental versions rather than experimental plugins.
- more compact display of available versions

followup

m-kuhn (approved):

borysiasty:
@olivierdalang could you please look at #36217 ?

pigreco:
Please add `Plugin` label

PR #34892, opened 2020-03-05
[MESH][3D] fix active faces not supported

vcloarec:
When active face flag values are not supported, the array of those flag are empty. The 3D dataset scalar rendering didn't take account of that case.
This PR fixes this issue

### DIFF
--- a/src/3d/mesh/qgsmesh3dgeometry_p.cpp
+++ b/src/3d/mesh/qgsmesh3dgeometry_p.cpp
@@ -144,7 +144,7 @@ static QByteArray createDatasetIndexData( const QgsTriangularMesh &mesh, const Q
   for ( int i = 0; i < trianglesCount; ++i )
   {
     int nativeFaceIndex = mesh.trianglesToNativeFaces()[i];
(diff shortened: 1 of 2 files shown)


PeterPetrik (approved):

src/3d/mesh/qgsmesh3dgeometry_p.cpp
```diff
@@ -144,7 +144,7 @@ static QByteArray createDatasetIndexData( const QgsTriangularMesh &mesh, const Q
   for ( int i = 0; i < trianglesCount; ++i )
   {
     int nativeFaceIndex = mesh.trianglesToNativeFaces()[i];
-    const bool isActive = mActiveFaceFlagValues.active( nativeFaceIndex );
+    const bool isActive = mActiveFaceFlagValues.active().isEmpty() || mActiveFaceFlagValues.active( nativeFaceIndex );
```
PeterPetrik: it does not harm, but not sure if 100% required https://github.com/qgis/QGIS/blob/master/src/core/mesh/qgsmeshdataprovider.cpp#L292

PR #34895, opened 2020-03-05
Remember parameter values between model designer runs

nyalldawson:
When designing a model, users typically will need to run the model many times as they tweak its structure.

This change causes the parameters used when running the model from the designer to be remembered and saved into the model, so that each time you run the model from the designer you don't have to re-set all the input parameter values to the desired test ones.

Makes iterative model design SO much easier!

Sponsored by Alta Ehf

Also includes a partial port of the model designer dialog from Python to c++, because c++ >> Python.

### DIFF
--- a/python/core/auto_generated/processing/models/qgsprocessingmodelalgorithm.sip.in
+++ b/python/core/auto_generated/processing/models/qgsprocessingmodelalgorithm.sip.in
@@ -443,6 +443,32 @@ These variables are added to the model's expression context scope, allowing for 
 .. versionadded:: 3.8
 %End
 
+    QVariantMap designerParameterValues() const;
+%Docstring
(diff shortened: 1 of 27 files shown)


nyalldawson:
Now also includes @3nids' requested "Show Comments" option

3nids:
Nice!
I was wondering, what do you think about doing the same for all processing algs and save the last params in the project (or in global, but it makes less sense to me)?

nyalldawson:
> I was wondering, what do you think about doing the same for all processing algs and save the last params in the project (or in global, but it makes less sense to me)?

mmm... not sure. I'd rather do it via https://github.com/qgis/QGIS-Enhancement-Proposals/issues/130

alexbruy (approved):

PR #34898, opened 2020-03-06
Add expression variables for current map temporal range

nyalldawson:
Adds:

@map_start_time: Start of the map's temporal time range (as a datetime value)
@map_end_time: End of the map's temporal time range (as a datetime value)
@map_interval: Duration of the map's temporal time range (as an interval value)


nyalldawson:
Ping @Samweli

PR #34902, opened 2020-03-06
Disable flaky tests on Travis

nyalldawson:
Travis is seriously annoying these days, so disable the culprits:

PyQgsPostgresRasterProvider: intermittently returns different results
PyQgsExpressionBuilderWidget: intermittent crash on end
PyQgsRulebasedRenderer: intermittent crash on end

### DIFF
--- a/.ci/travis/linux/scripts/test_blacklist.txt
+++ b/.ci/travis/linux/scripts/test_blacklist.txt
@@ -4,6 +4,13 @@ PyQgsPalLabelingServer
 qgis_composerutils
 PyQgsAppStartup
 
+# flaky 
+PyQgsPostgresRasterProvider
+
+# intermittent crashes
+PyQgsExpressionBuilderWidget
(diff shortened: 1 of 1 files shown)


elpaso:
Let me have a look to PyQgsPostgresRasterProvider, I hoped I had fixed it and I need the tests to be running.

nirvn:
Scary to have intermittent crashes we don't understand in the rule based renderer. But it's indeed been a source of frustration.

nyalldawson:
I tracked that one down. It's a bug in sip, where it incorrectly keeps ownership despite a transfer annotation

nirvn:
@nyalldawson , does it potentially impact users? If so, stupid question from my part (as I assume you've already answered it yourself): is there any work around that we could apply?

elpaso:
let's see if this helps: https://github.com/qgis/QGIS/pull/34904

PR #34905, opened 2020-03-06
Fixes the activation of the shape map tools when digitizing is finished

lbartoletti:
## Description
The digitizing logic for these tools is to deactivate them when digitizing is complete (right-click). However, the tool remains active, but the icon is not pushed in, and the Advanced Digitizing Tool is not active. To fix it, activate the tool again.

### DIFF
--- a/src/app/qgsmaptooladdcircle.cpp
+++ b/src/app/qgsmaptooladdcircle.cpp
@@ -139,3 +139,13 @@ void QgsMapToolAddCircle::clean()
   if ( vLayer )
     mLayerType = vLayer->geometryType();
 }
+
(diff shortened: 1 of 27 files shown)


mhugo:
Thanks for the fix and the description. Good for me. However, I noticed all the map tools share lots of code in common. At least for the right button handling, isn't there a way to factor a bit ?

lbartoletti:
> Thanks for the fix and the description. Good for me. However, I noticed all the map tools share lots of code in common. At least for the right button handling, isn't there a way to factor a bit ?

Done, with a tiny refacto. I'll look for a better replacement of these tools later.

mhugo:
@lbartoletti Thanks !

PR #34590, opened 2020-02-20
QGIS Project temporal settings

Samweli:
This PR add project time settings functionality, it includes work from this [PR](https://github.com/qgis/QGIS/pull/34497) which has added support for constrains of temporal layers from WMS-T provider.

User will be able to set the project temporal range by either using manual input or calculate it from the current projects temporal layers by clicking the provided "Calculate from Layers" button.

This implementation set the stage for adding animation feature for temporal layers in QGIS.

Example of how setting project temporal range will be.
set project time settings

Prior discussions about QGIS temporal support can be found [here](https://github.com/qgis/QGIS-Enhancement-Proposals/issues/161) and [here ](https://github.com/qgis/QGIS-Enhancement-Proposals/issues/128)

PR #34909, opened 2020-03-06
Temporal layers indicators

Samweli:
Adds support for the temporal layers indicators in the Layers panel work tree.

These indicators change depending on from where the current layers temporal range is set from, eg from the project settings or from layer temporal properties options.

This PR contains work from https://github.com/qgis/QGIS/pull/34590
Temporal_layer_indicator

Note: This is part of  the work from discussions found  [here](https://github.com/qgis/QGIS-Enhancement-Proposals/issues/161)  and [here](https://github.com/qgis/QGIS-Enhancement-Proposals/issues/128) .


src/app/qgslayertreeviewprojecttimeindicator.h
```diff
@@ -0,0 +1,53 @@
...
+#define QGSLAYERTREEVIEWPROJECTTIMEINDICATOR_H
+
+
+#include "qgslayertreeviewindicatorprovider.h"
+
+/**
+ * Adds indicators for showing temporal layers that are using project time to set their
+ * current active temporal range.
+ */
+class QgsLayerTreeViewProjectTimeIndicatorProvider : public QgsLayerTreeViewIndicatorProvider
```
nyalldawson: @Samweli could you condense this logic into QgsLayerTreeViewTemporalIndicatorProvider ? It'll just be a matter of changing the return icon and tooltip text based on the layer's temporal source, so I don't think there's a strong need for two separate classes here.

src/app/qgslayertreeviewtemporalindicator.cpp
```diff
@@ -0,0 +1,93 @@
...
+}
+
+void QgsLayerTreeViewTemporalIndicatorProvider::connectSignals( QgsMapLayer *layer )
+{
+  if ( !( qobject_cast<QgsVectorLayer *>( layer ) || qobject_cast<QgsRasterLayer *>( layer ) ) )
+    return;
+  QgsMapLayer *mapLayer = layer;
+  connect( mapLayer, &QgsMapLayer::dataSourceChanged, this, &QgsLayerTreeViewTemporalIndicatorProvider::onLayerChanged );
+
+  if ( mapLayer && mapLayer->dataProvider() )
```
nyalldawson: This is not the right signal to connect to. Instead we should:

1. Make QgsMapLayerTemporalProperties a QObject, and add a "changed" signal to it
2. Connect that signal to onLayerChanged here
3. Make sure the changed signal is correctly emitted whenever the QgsMapLayerTemporalProperties settings change, including when QgsTemporalProperties::setIsActive is called (you'd need to make that virtual in QgsTemporalProperties in order for this to work)

src/app/qgslayertreeviewtemporalindicator.cpp
```diff
@@ -0,0 +1,93 @@
...
+    mLayer = mapLayer;
+    QgsRasterDataProvider *provider = qobject_cast<QgsRasterDataProvider *>( mapLayer->dataProvider() );
+    connect( provider, &QgsRasterDataProvider::statusChanged,
+             this, &QgsLayerTreeViewTemporalIndicatorProvider::onLayerChanged );
+  }
+}
+
+void QgsLayerTreeViewTemporalIndicatorProvider::onIndicatorClicked( const QModelIndex &index )
+{
+  Q_UNUSED( index )
```
nyalldawson: Here we should open the layer properties, and switch to the temporal tab

src/app/qgslayertreeviewtemporalindicator.cpp
```diff
@@ -0,0 +1,93 @@
...
+}
+
+bool QgsLayerTreeViewTemporalIndicatorProvider::acceptLayer( QgsMapLayer *layer )
+{
+  if ( !layer )
+    return false;
+  if ( layer->temporalProperties() )
```
nyalldawson:
```suggestion
  return layer && layer->temporalProperties() && layer->temporalProperties()->isActive();
```

src/app/qgslayertreeviewtemporalindicator.cpp
```diff
@@ -0,0 +1,93 @@
...
+    return QStringLiteral( "/mIndicatorTimeFromProject.svg" );
+
+  return QStringLiteral( "/mIndicatorTemporal.svg" );
+}
+
+QString QgsLayerTreeViewTemporalIndicatorProvider::tooltipText( QgsMapLayer *layer )
+{
+  if ( layer->temporalProperties()->temporalSource() ==
+       QgsMapLayerTemporalProperties::TemporalSource::Project )
+    return tr( "<b>Temporal layer using Project time </b>" );
```
nyalldawson:
```suggestion
    return tr( "<b>Temporal layer using project's time range</b>" );
```

src/app/qgslayertreeviewtemporalindicator.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+  if ( !mapLayer )
+  {
+    if ( !mapLayerFromIndicator )
+      return;
+    else
+      updateLayerIndicator( mapLayerFromIndicator );
+  }
+  else
+    updateLayerIndicator( mapLayer );
```
nyalldawson: This looks quite complex. I think this should work instead:

```suggestion
  QgsMapLayer *layer = qobject_cast<QgsMapLayer *>( sender() );
  if ( !layer )
    return;

  updateLayerIndicator( layer );
```

src/app/qgslayertreeviewtemporalindicator.cpp
```diff
@@ -0,0 +1,93 @@
...
+void QgsLayerTreeViewTemporalIndicatorProvider::connectSignals( QgsMapLayer *layer )
+{
+  if ( !( qobject_cast<QgsVectorLayer *>( layer ) || qobject_cast<QgsRasterLayer *>( layer ) ) )
+    return;
+  QgsMapLayer *mapLayer = layer;
+  connect( mapLayer, &QgsMapLayer::dataSourceChanged, this, &QgsLayerTreeViewTemporalIndicatorProvider::onLayerChanged );
+
+  if ( mapLayer && mapLayer->dataProvider() )
+  {
+    mLayer = mapLayer;
```
nyalldawson: Don't store the mapLayer here as a member -- this indicator provider will be used for ALL map layers in the project. (i.e. only one instance of the provider is ever created, it's not layer-specific)

src/app/qgslayertreeviewtemporalindicator.h
```diff
@@ -0,0 +1,49 @@
...
+
+    //! Adds/removes indicator of a layer
+    void onLayerChanged();
+
+  private:
+    bool acceptLayer( QgsMapLayer *layer ) override;
+    QString iconName( QgsMapLayer *layer ) override;
+    QString tooltipText( QgsMapLayer *layer ) override;
+
+    QgsMapLayer *mLayer = nullptr;
```
nyalldawson:
```suggestion

```

src/core/qgsmaplayertemporalproperties.h
```diff
@@ -38,6 +38,7 @@
  */
 class CORE_EXPORT QgsMapLayerTemporalProperties : public QgsTemporalProperty
 {
+
```
nyalldawson: add Q_OBJECT (also to QgsRasterLayerTemporalProperties)

src/core/qgstemporalproperty.h
```diff
@@ -58,8 +59,16 @@ class CORE_EXPORT QgsTemporalProperty
     */
     bool isActive() const;
 
+  signals:
+
+    /**
+     * Emitted when the temporal properties have changed.
+     */
+    void changed();
```
nyalldawson: Can you add unit tests checking that this is correctly emitted when the various settings change (and not re-emitted when the existing value is re-set)

src/app/qgslayertreeviewtemporalindicator.cpp
```diff
@@ -0,0 +1,86 @@
...
+QgsLayerTreeViewTemporalIndicatorProvider::QgsLayerTreeViewTemporalIndicatorProvider( QgsLayerTreeView *view )
+  : QgsLayerTreeViewIndicatorProvider( view )
+{
+}
+
+void QgsLayerTreeViewTemporalIndicatorProvider::connectSignals( QgsMapLayer *layer )
+{
+  if ( !( qobject_cast<QgsVectorLayer *>( layer ) || qobject_cast<QgsRasterLayer *>( layer ) ) )
+    return;
+  QgsMapLayer *mapLayer = layer;
```
nyalldawson: no need for this copy here - just use "layer" instead.

nyalldawson (commented):
Looking close!

Now that QgsTemporalProperty/QgsMapLayerTemporalProperties/QgsRasterLayerTemporalProperties are QObjects, you'll need to ensure these all have the Q_OBJECT macro inclusion, and also accept a parent QObject* argument in their constructors.

Then, we need to make sure that the QgsRasterLayerTemporalProperties are correctly parented to the raster layer when constructed. So in qgsrasterlayer.cpp

    mTemporalProperties = qgis::make_unique<QgsRasterLayerTemporalProperties>();

becomes

    mTemporalProperties =  new QgsRasterLayerTemporalProperties( this );

You'll also need to change

    std::unique_ptr< QgsRasterLayerTemporalProperties > mTemporalProperties;
 
to 

    QgsRasterLayerTemporalProperties * mTemporalProperties = nullptr;

in the header, since we'll be using Qt's parent memory management instead of c++ unique_ptr

nyalldawson (approved):

nyalldawson:
Looks great, thanks!

nyalldawson:
Can you rebase this branch off master? There's a merge conflict here.

Samweli:
> Can you rebase this branch off master? There's a merge conflict here.

Sure, done!

nyalldawson:
Thanks -- I'll merge as soon as the CI results are green!

Samweli:
Thanks @nyalldawson

PR #34924, opened 2020-03-06
Add model for connections from a provider

nyalldawson:
Allows for easy creation of comboboxes/lists which list the registered connections for a specified data provider (the provider must implement the connections API, so postgres/gpkg/spatialite for now)


nyalldawson:
Also includes a schema model now!

nyalldawson:
And lastly with a table model too...

nyalldawson:
A totally finally with a set of chainable widgets exposing this functionality.

As seen in action here:
Peek 2020-03-08 11-36

nirvn:
That'll be useful for our create new gpkg/spatialite layer dialogs.

PR #34928, opened 2020-03-07
[processing][needs-docs] Adjust status of controls executing an algorithm dialog 

### LINKED ISSUE #34484: Usability: Avoid running a Processing algorithm twice by accident
**Feature description.**
After you run a Processing algorithm, the `Log` tab is visible and the `Run` button is enabled, so you can run the algorithm again with a single click. From a user perspective, it shouldn't be the case. If you want to run the algorithm again, you could at least have a look at the `Parameters` tab once more.

**1st proposal**
1. Run the algorithm.
2. Disable `Run` button.
3. Enable `Run` button as soon as the user activates the `Parameters` tab.

1st_proposal

**2nd proposal**

Only run if `Parameters` tab is active. That is, avoid running the algorithm if `Log` tab is active.

2nd_proposal

**3rd proposal**
1. Run the algorithm.
2. Change `Run` button by `Prepare next run` button.
3. A click on `Prepare next run` button activates `Parameters` tab, which changes `Prepare next run` button by `Run` button. 

3rd_proposal

----------------------------------------
Let me know which one you all prefer, then I'll create a PR. I prefer the 2nd proposal. 
If you have another idea, please comment it.

*Note: I didn't write a QEP because this is a minor change in UI.*

gacarrillor:
## Description

Currently, after you run a Processing algorithm, the `Log` tab is visible and the `Run` button is enabled, so you can run the algorithm again with a single click. From a user perspective, it shouldn't be the case. If you want to run the algorithm again, you should at least have a look at the `Parameters` tab once more.

This PR adds the following changes to such behaviour:

  + `Run` button is not shown anymore in the `Log` tab, therefore, you can only run algorithms from the `Parameters` tab.
  + While running an algorithm, the `Parameters` panel is now disabled, as well as `Run as Batch Process...` and `Run as Single Process...` buttons.
  + When an algorithm execution finishes (either successfully or not), a new button `Change Parameters` is shown in the `Log` tab.
  + There is now consistency among algorithm dialogs: Before this PR, `Batch Algorithm Dialog` was the only dialog blocking parameter widgets.

These changes were applied to the `Algorithm Dialog` and `Batch Algorithm Dialog`, and work on `Edit in place` dialogs as well. This PR also takes into account cancelling an algorithm execution or obtaining failures in the execution.

(More details and preliminary discussion in issue https://github.com/qgis/QGIS/issues/34484)

### Algorithm dialog

pr_algorithm_dialog_v3

### Batch algorithm dialog

pr_batch_algorithm_dialog_v3

### Cancelling algorithm execution

pr_cancelling_algorithm_dialog_v3

----------------------------
Fix #34484 
Since this PR implies updating QGIS docs, I could give a hand adjusting the corresponding section.

### DIFF
--- a/python/gui/auto_generated/processing/qgsprocessingalgorithmdialogbase.sip.in
+++ b/python/gui/auto_generated/processing/qgsprocessingalgorithmdialogbase.sip.in
@@ -75,6 +75,11 @@ Returns the main widget for the dialog, usually a panel for configuring algorith
 Switches the dialog to the log page.
 %End
 
+    void showParameters();
+%Docstring
+Switches the dialog to the parameters page.
+%End
+
     bool wasExecuted() const;
 %Docstring
 Returns ``True`` if an algorithm was executed in the dialog.
@@ -197,6 +202,11 @@ Returns the dialog's run button.
 Returns the dialog's cancel button.
 %End
 
+    QPushButton *changeParametersButton();
+%Docstring
+Returns the dialog's change parameters button.
+%End
+
     QDialogButtonBox *buttonBox();
 %Docstring
 Returns the dialog's button box.
@@ -221,6 +231,12 @@ Sets whether the algorithm was executed through the dialog.
 .. seealso:: :py:func:`setResults`
 %End
 
+    void setExecutedAnyResult( bool executedAnyResult );
+%Docstring
+Sets whether the algorithm was executed through the dialog (no matter the result).
+%End
+
+
     void setResults( const QVariantMap &results );
 %Docstring
 Sets the algorithm results.
@@ -238,6 +254,17 @@ Displays an info ``message`` in the dialog's log.
     void resetGui();
 %Docstring
 Resets the dialog's gui, ready for another algorithm execution.
+%End
+
+    void updateRunButtonVisibility();
+%Docstring
+Sets visibility for mutually exclusive buttons Run and Change Parameters.
+%End
+
+    void blockControlsWhileRunning();
+%Docstring
+Blocks run and changeParameters buttons and parameters tab while the
+algorithm is running.
 %End
 
     QgsMessageBar *messageBar();
--- a/python/gui/auto_generated/processing/qgsprocessingalgorithmdialogbase.sip.in
+++ b/python/gui/auto_generated/processing/qgsprocessingalgorithmdialogbase.sip.in
@@ -256,6 +256,12 @@ Displays an info ``message`` in the dialog's log.
 Resets the dialog's gui, ready for another algorithm execution.
 %End
 
+    virtual void resetAdditionalGui();
+%Docstring
+For subclasses to register their own GUI controls to be reset, ready
+for another algorithm execution.
+%End
+
     void updateRunButtonVisibility();
 %Docstring
 Sets visibility for mutually exclusive buttons Run and Change Parameters.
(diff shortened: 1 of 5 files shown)


roya0045:
I get where you're coming from but must of what I see is putting restrain and annoyances to the UX just because someone ran the algorithm another time by accident. I know that what I say doesn't matter but as a user I'm never a fan of PR that block things and impede my freedom. :/

alexbruy:
I'm -1 with disabling "Parameters" tab during algorithm execution. Why I can't take a look at the parameters while algorithm executes? What if I have multiple dialogs open and running and want to cancel some specific instance?

alexbruy:
Also I know, I'm late but from my point of view adding "Change parameters" button makes very little sense it is enough just to hide "Run" button from the "Log" tab.

gacarrillor:
@alexbruy Even though parameters are also displayed in the Log tab, I understand that disabling Parameters tab is too much. Therefore, I am now following the Batch Algorithm Dialog's behaviour: leave tabs enabled, but disable the Parameters panel, so that you can check the parameters but cannot change them during execution. I just updated the screencasts.

Concerning the "Change Parameters" button, well, I tend to agree with you. It's something we discussed in https://github.com/qgis/QGIS/issues/34484 and some core devs agreed on adding the button as the best option for users. I hope we all can come to an agreement.

m-kuhn:
UX is always hard to get right for everyone. I think that all the proposals discussed in #34484 are better than the current situation and it would be nice to get out of this stealth limbo here.

@alexbruy could you leave a comment in the original discussion since that's where the decision was taken. If there is no change in the voting results by Friday, March 20, I'd like to merge this PR as is to move ahead.

alexbruy:
@m-kuhn done

m-kuhn:
thanks everyone !
I think we have quite a good consensus with this. If someone still thinks it's awkward in 2 months time, feel free to open up a follow up pull request :)

nyalldawson:
Having used this change over the last 24 hours, there's something which needs fixing.

Previously the button order was always:

    [ close]   [ run ]

Now things change depending on the state. Initially we have:

    [close ]     [ run]

after running and algorithm, the order changes to

    [ change parameters ] [ close] 

i.e. the position of the close button jumps around. I've hit the wrong button many many times over the last day as a result. 

Similarly, if I hit Change Parameters, the button order changes back to:

   [ close ] [ run ]

and it's really easy to hit close now by accident, cos you were expecting "run" to appear under your cursor.

So basically, we shouldn't mess with the button order, and it needs to be:

   [ close ] [ change parameters]

with close **always** on the same side.

gacarrillor:
@nyalldawson Thanks for the report. It has been addressed in https://github.com/qgis/QGIS/pull/35227

gacarrillor:
Two more questions from my side:

 + Could somebody please add the `Needs Documentation` label to this PR?

 + Creating #35227, I realized the `python/plugins/processing/ui/DlgAlgorithmBase.ui` seems not to be used anymore. Shall I remove it from QGIS code?

m-kuhn:
> Could somebody please add the Needs Documentation label to this PR?

@gacarrillor you should be able to do it yourself, @mbernasocchi granted you the rights

gacarrillor:
> > Could somebody please add the Needs Documentation label to this PR?
> 
> @gacarrillor you should be able to do it yourself, @mbernasocchi granted you the rights

Unfortunately, I still can't do it.

labels

mbernasocchi:
@gacarrillor You havent accepted the invitation yet to the QGIS community team, I just reinvited you

gacarrillor:
Thank you @mbernasocchi, I wasn't aware of an invitation, now it has appeared in my e-mail. 
Happy to be part of this (GitHub) community! :tada:

PR #34936, opened 2020-03-09
[FEATURE][callouts] Add control over anchor point for callout on label

nyalldawson:
This gives users control over where a callout should join to the label
text. (Previously, you only had control over where the callout would
join to the corresponding feature geometry).

Choices include:
- Closest point (previous behavior)
- Label Centroid
- Fixed corners: Top left/top right/bottom left/bottom right/etc

Data defined control over the label anchor is also possible

Helps fix some callout ugliness!

Previous result (closest point mode):

image

With new "centroid" placement:

image


nyalldawson:
@nirvn care to review this one?

nirvn:
@nyalldawson , I'll do more than that. Not only am I going to review this, I'll add the settings is the label propreties map tool dialog too :)

nyalldawson:
> I'll add the settings is the label propreties map tool dialog too :)

wheeeee :smiley:

src/ui/callouts/widget_simplelinecallout.ui
```diff
@@ -203,39 +216,43 @@
...
        <property name="text">
-        <string>Anchor point</string>
+        <string>Label anchor point</string>
        </property>
       </widget>
      </item>
-     <item row="5" column="1" colspan="2">
-      <widget class="QComboBox" name="mAnchorPointComboBox"></widget>
+     <item row="6" column="1" colspan="2">
+      <widget class="QComboBox" name="mLabelAnchorPointComboBox"/>
```
nirvn: Needs to be added in the <tabstops> section.

tests/src/core/testqgscallout.cpp
```diff
@@ -939,6 +949,527 @@ void TestQgsCallout::calloutDataDefinedOffsetFromLabel()
...
+  callout->dataDefinedProperties().setProperty( QgsCallout::LabelAnchorPointPosition, QgsProperty::fromExpression( QStringLiteral( "'TL'" ) ) );
+  settings.setCallout( callout );
+
+  vl->setLabeling( new QgsVectorLayerSimpleLabeling( settings ) );  // TODO: this should not be necessary!
+  vl->setLabelsEnabled( true );
+
+  QgsDefaultLabelingEngine engine;
+  engine.setMapSettings( mapSettings );
+  engine.addProvider( new QgsVectorLayerLabelProvider( vl, QString(), true, &settings ) );
+  //engine.setFlags( QgsLabelingEngine::RenderOutlineLabels | QgsLabelingEngine::DrawLabelRectOnly );
```
nirvn: Those commented out lines are meant to stay for some practical debugging use or should those be removed?

nyalldawson: I honestly have no idea -- they came from tests in the label engine test suite which have been used as a template everywhere else. I don't know if it's still a valid comment or not, so I've just been leaving it in place...

nirvn: Ok, well +1 for consistency, let's leave those in :)

src/ui/callouts/widget_simplelinecallout.ui
nyalldawson: Fixed

nirvn (approved):

nyalldawson:
Thanks @nirvn!

PR #34938, opened 2020-03-09
Add undo/redo support to model designer

nyalldawson:
Makes QGIS more forgiving for users!

Works just like you'd expect:

Peek 2020-03-09 11-54

Note that we save the whole model definition in the undo stack, not just the affect component changes. This was a lesson I learnt after the composer redesign -- it's just safest to save and restore the WHOLE document then try to deal with interconnected changes.


nyalldawson:
@alexbruy did you see this one? Can I get a review please?

alexbruy:
@nyalldawson sorry, I have missed this. Last days were busy.

alexbruy (approved):
Looking good.

nyalldawson:
Thanks @alexbruy ! Let me know if you need a break from reviewing these :wink:

Issue #34537, opened 2020-02-19
"Allow NULL value" in Relation Reference Widget is not respected in attributes form 


basilrabi:
**Describe the bug**

When ticking `Allow NULL value` in Relation Reference (many-to-one in postgis) widget and editing the feature (combobox), selecting NULL or entering NULL or empty text in the combo box is not treated as a **change** in the feature.

In order to successful set the foreign key field to `NULL` in QGIS 3.10.2, you need to tick `Use a read-only line edit instead of a combobox` then press the red `X` mark in the form.
However in master (a3c08ad667b), this work around does not work and is clearly a regression from 3.10.x.

**How to Reproduce**

To reproduce, see [screencast](https://github.com/qgis/QGIS/files/4222866/Screencast.from.Wednesday.19.February.2020.10.17.37.AM.PST.zip).

**QGIS and OS versions**

1. QGIS 3.10.2 Flatpak (OS: Fedora 31)
1. QGIS master (a3c08ad667b) (OS: Fedora 31)

**Additional context**

Although I did not test, this could also be a regression from QGIS 3.4 and might be related to [this issue](https://issues.qgis.org/issues/17505).

gioman:
> Although I did not test, this could also be a regression from QGIS 3.4 and might be related to [this issue](https://issues.qgis.org/issues/17505).

@basilrabi it would be very important to find out if is a regression.

basilrabi:
@gioman I tested in Windows using QGIS 3.4.15. It works fine there, choosing NULL in the combobox is recorded as a change and can be saved successfully. This is a regression in 3.10 and master.

basilrabi:
The fix works for me in master. I hope this fix also reaches 3.10.x and 3.12.x

PR #34943, opened 2020-03-09
Fixes relation reference widget several issues

troopa81:
# Description

This PR fixes various bug in relation reference widget combobox.
- A regression coming from PR #34355 (see this [comment](https://github.com/qgis/QGIS/pull/34355#issuecomment-593455667) ).
- https://github.com/qgis/QGIS/issues/34537 : Allow settings null in relation reference widget
  The default constructed QVariant when there is no selected values is invalid (type is QMetaType::UnknownType), which break [this](https://github.com/qgis/QGIS/blob/master/src/gui/qgsattributeform.cpp#L1039). 
- Thread access to shared memory ressource
- ComboBox entries populating

I removed the `updateIndex` method because it changes the current index even if doesn't need to. The strategy of which item to prefer between null and a possible entry is now done in `QgsFeatureFilterModel::updateCompleter`.

Once we agree on this PR, I will backport manually in 3.10 and 3.12 with #34355

### DIFF
--- a/src/core/qgsfeaturefiltermodel.cpp
+++ b/src/core/qgsfeaturefiltermodel.cpp
@@ -225,9 +225,11 @@ QVariant QgsFeatureFilterModel::data( const QModelIndex &index, int role ) const
 void QgsFeatureFilterModel::updateCompleter()
 {
   emit beginUpdate();
-  if ( !mGatherer )
+
+  QgsFieldExpressionValuesGatherer *gatherer = qobject_cast<QgsFieldExpressionValuesGatherer *>( sender() );
+  if ( gatherer->wasCanceled() )
   {
-    emit endUpdate();
+    delete gatherer;
     return;
   }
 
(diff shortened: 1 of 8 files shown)


elpaso:
@troopa81 I'm sorry but I have no time to pull and build and test your branch right now but as long as you were able to reproduce the various issues and you are confident they are fixed I would suggest to merge it to master (that will make it quicker to test).

The required condition is that if after the merge there are still issues you will be able to jump back on them again quickly.

How does it sounds?

troopa81:
@elpaso 

OK, let's merge on next monday (I'll ping you). I'm not available at the moment to fix possible regression.

FYI, CI fails because of rulebasedrenderer test, which I think has nothing to do with this PR.

troopa81:
@elpaso Let's merge? I can wait for backport if you'd like to test in the days to come.

elpaso:
@troopa81 I trust your confidence :) I'm not able to test it right now.

PR #34954, opened 2020-03-10
Remove duplicate nodes from all inner rings

gacarrillor:
## Description

**Expected**: Running the `Remove duplicate vertices` algorithm does remove duplicate vertices from each inner ring in a polygon.

**Currently**: Each run of such algorithm removes duplicate vertices in a single inner ring at a time. So, if you have 2 inner rings that have duplicate vertices (each one separately), you need to run the algorithm 2 times to 'clean' the layer. :confused:

This PR fixes the issue, adjusting the `QgsCurvePolygon.removeDuplicateNodes()` method, which was causing the misbehaviour in the Processing algorithm.

New tests for:
 + Remove Duplicate Vertices algorithm.

These are the duplicate vertices in the test data, just in case you want to validate:

image

Follow up #34750


nyalldawson (approved):

m-kuhn:
Thanks @gacarrillor 

Should this be backported too?

gacarrillor:
> Should this be backported too?

For sure! :+1:

PR #34955, opened 2020-03-10
Add dedicated parameter type for database connections

nyalldawson:
Allows selection from the registered database connections for a specific database provider type (the provider must implement the connections API)


alexbruy (approved):
Nice!

nyalldawson:
Thanks @alexbruy !

PR #34968, opened 2020-03-10
[api] New database table name widget

elpaso:
## A new widget and a small set of new API methods

A new small widget to select a new table for an existing DB connection:

qgis-new-table-widget gif

## API for hiding layers from the browser

With changes to dataitems/dataitemprovider and the the browser proxy model


src/core/qgsbrowserproxymodel.cpp
```diff
@@ -153,9 +153,28 @@ bool QgsBrowserProxyModel::filterAcceptsRow( int sourceRow, const QModelIndex &s
   if ( !filterAcceptsProviderKey( sourceIndex ) || !filterRootAcceptsProviderKey( sourceIndex ) )
     return false;
 
+  if ( ! mShowLayers )
```
nyalldawson: Can you add tests for this in TestQgsBrowserProxyModel

src/core/qgsdataitemproviderregistry.h
```diff
@@ -63,6 +65,14 @@ class CORE_EXPORT QgsDataItemProviderRegistry
      */
     void removeProvider( QgsDataItemProvider *provider );
 
+    /**
+     * Returns the (possibly blank) data provider key for a given data item provider name.
+     *
+     * \param dataItemProviderName name of the data item provider
+     * \since QGIS 3.14
+     */
+    QString dataProviderKey( const QString &dataItemProviderName );
```
nyalldawson: Add test to TestQgsDataItemProviderRegistry

src/gui/qgsnewdatabasetablenamewidget.cpp
```diff
@@ -0,0 +1,266 @@
...
+  }
+  else
+  {
+    mBrowserModel = browserModel;
+    mBrowserModel->initialize();
+  }
+
+  setupUi( this );
+
+  mValidationResults->setStyleSheet( QStringLiteral( "* { font-weight: bold; color: red; }" ) );
```
nyalldawson: @nirvn is this the safe?

src/gui/qgsnewdatabasetablenamewidget.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+  QStringList hiddenProviders
```
nyalldawson: Hm, this is not great. I can easily see this getting out of sync with the existing providers. How would you feel about adding a provider "whitelist" to QgsBrowserProxyModel instead?

src/gui/qgsnewdatabasetablenamewidget.cpp
```diff
@@ -0,0 +1,266 @@
...
+    {
+      hiddenProviders.push_back( provider->name() );
+    }
+  }
+
+  mBrowserProxyModel.setBrowserModel( mBrowserModel );
+  mBrowserProxyModel.setDataItemProviderKeyFilter( hiddenProviders );
+  mBrowserProxyModel.setShowLayers( false );
+  mBrowserTreeView->setHeaderHidden( true );
+  mBrowserTreeView->setExpandsOnDoubleClick( false );
```
nyalldawson: Why this?

src/gui/qgsnewdatabasetablenamewidget.cpp
```diff
@@ -0,0 +1,266 @@
...
+    emit tableNameChanged( mTableName );
+    validate();
+  } );
+
+  connect( mBrowserTreeView, &QgsBrowserTreeView::clicked, this, [ = ]( const QModelIndex & index )
+  {
+    if ( index.isValid() )
+    {
+      const QgsDataItem *dataItem( mBrowserProxyModel.dataItem( index ) );
+      if ( dataItem )
```
nyalldawson:
```suggestion
      if ( const QgsDataItem *dataItem = mBrowserProxyModel.dataItem( index )  )
```

src/gui/qgsnewdatabasetablenamewidget.cpp
```diff
@@ -0,0 +1,266 @@
...
+      {
+        const QgsDataCollectionItem *collectionItem = qobject_cast<const QgsDataCollectionItem *>( dataItem );
+        if ( collectionItem )
```
nyalldawson: ### 
```suggestion
        if ( const QgsDataCollectionItem *collectionItem = qobject_cast<const QgsDataCollectionItem *>( dataItem ) )
```

src/gui/qgsnewdatabasetablenamewidget.cpp
```diff
@@ -0,0 +1,266 @@
...
+        {
+          const QString providerKey { QgsApplication::dataItemProviderRegistry()->dataProviderKey( dataItem->providerKey() ) };
+          if ( mShownProviders.contains( providerKey ) )
```
nyalldawson: Will this ever be false?

src/gui/qgsnewdatabasetablenamewidget.h
```diff
@@ -0,0 +1,140 @@
...
+     * \param parent optional parent for this widget
+     */
+    explicit QgsNewDatabaseTableNameWidget( QgsBrowserGuiModel *browserModel = nullptr,
+                                            const QStringList &providersFilter = QStringList(),
+                                            QWidget *parent = nullptr );
+
+    /**
+     * Returns the currently selected schema for the new table
+     */
+    QString schema();
```
nyalldawson: const

src/gui/qgsnewdatabasetablenamewidget.h
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     * Returns the current name of the new table
+     */
+    QString table();
```
nyalldawson: const

src/gui/qgsnewdatabasetablenamewidget.h
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     * Returns the currently selected data item provider key
+     */
+    QString dataProviderKey();
```
nyalldawson: const

src/gui/qgsnewdatabasetablenamewidget.h
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     * Returns TRUE if the widget contains a valid new table name
+     */
+    bool isValid() const;
+
+    /**
+     * Returns the validation error or an empty string is the widget status is valid
+     */
+    QString validationError();
```
nyalldawson: const

nyalldawson (commented):
Looking good (+nice clean ups!) 

The data source select dialog includes a refresh button. Do you think we should add one here? I'm imagining a use case where someone is asked for a table name, realises they need to create a new schema, and then needs to refresh the tree to show this newly created schema. (Or does the view automatically expose the browser right click actions like refresh and create schema already?)

src/gui/qgsnewdatabasetablenamewidget.cpp
nirvn: It's 'safe' in that we do this in many (relatively speaking) other places.

In the future, we should have QgsApplication::color( Role ), with default colors that can be overwritten by themes. But that won't happen today :)

@elpaso , BTW, I'd request two UI changes here: 
1/ don't use bold, a change of color to stay table name collision is enough;
2/ IMHO, using red to provide a tip to the user as to what to do here (i.e. 'select a database schema, enter a unique name...') is not good UI/UX. Could you change this so that unless you're reporting an error (i.e. name collision), the string uses the default system color? Thanks

src/core/qgsbrowserproxymodel.cpp
elpaso: Yes, I thought about it but the existing test does not use real layer items that I can hide and the test isn't going to be trivial to implement.

src/gui/qgsnewdatabasetablenamewidget.cpp
elpaso: That makes sense. I would also rename setDataItemProviderKeyFilter to setDataItemProviderKeyExcludeFilter and create a new setDataItemProviderKeyIncludeFilter

src/gui/qgsnewdatabasetablenamewidget.cpp
elpaso: No, it should not.

src/gui/qgsnewdatabasetablenamewidget.cpp
elpaso: shameless copied from the change datasource dialog.

> Looking good (+nice clean ups!)
> 
> The data source select dialog includes a refresh button. Do you think we should add one here? I'm imagining a use case where someone is asked for a table name, realises they need to create a new schema, and then needs to refresh the tree to show this newly created schema. (Or does the view automatically expose the browser right click actions like refresh and create schema already?)

No, the view has no context menus. What would you prefer: add the context menus or the refresh button, or both?

nyalldawson: I'm unsure - the context menus sound useful (and less obtrusive) but I'm 50/50 on whether having actions which edit things are good practice in an "open" type dialog. Your call!

src/gui/qgsnewdatabasetablenamewidget.cpp
elpaso: BTW, keep in mind that we would need to create a white or a black list in any case: we cannot show unsupported root items in the browser tree widget.

elpaso:
I think 52 files changed for a small widget is kind of a record, did I win anything? :smiley_cat: 

(and it's not yet done)

nyalldawson:
New code looks good to me!

src/gui/qgsnewdatabasetablenamewidget.cpp
elpaso: Added refresh tool button

PR #34978, opened 2020-03-10
fix crash when a sink parameter definition is missing in a processing alg

3nids:
when running an alg with a missing parameter sink definition, I got a crash.
this solves it.

here is a crashing alg: https://gist.github.com/3nids/01d57ea4ca15535094cd51cd2728520b


nyalldawson:
I think you'll get a crash for any of these parameter types of they don't exist, not just this one.

Raising an exception here would be preferred I think -- it should be a fatal mistake to make

3nids:
It's indeed better, but the message is not shown:
Screenshot 2020-03-11 at 17 49 59

3nids:
much better, thanks @nyalldawson 

image

nyalldawson:
Looking good! I'd just add the name of the sink in that error too - "No parameter definition exists for a sink named 'xxxx'"

3nids:
@nyalldawson you don't have access to the name in `QgsProcessingParameters::parameterAsSink`, I did a try catch in `QgsProcessingAlgorithm::parameterAsSink` to send the name.

nyalldawson (approved):

PR #34997, opened 2020-03-11
Add ability to edit, import and export user expressions

### LINKED ISSUE #34690: Enable editing and sharing of user expressions
# Problem

The new save expressions in QGIS are very cool however it lacks a few options to make it awesome.

user-expression

* Add ability to edit a user expression including the help text
* Add the ability to export user expression. 
* Import user expression - Expression differ between versions so a quick check would have to be done when importing. I assume add QGIS version in the XML for the user expression

*NB* On a side note maybe it's also worth to add the export/import options for Custom functions as well

suricactus:
## Description

Add ability to edit, import and export user expressions

Fixes qgis#34690, qgis#28478

#### Edit
A new button is added that allow to edit saved expressions. It is only active when a user expression is selected in the expressions tree. When pressed, it opens a window that allows editing the currently selected user expression. Changing the label of the expression makes a copy of the currently existing one.
Demo [Edit](https://imgur.com/w3UEOoR)
#### Import/Export
A new button with menu is added, that allows to import/export user expressions from/to a JSON file. Clicking on either of the menu entries opens a file selector dialog for the expressions JSON file source/destination. In case of label conflicts, additional dialog asks how to proceed - either to overwrite or to skip the current expression.
Demo [Import/Export](https://imgur.com/0eiaFfu)

### DIFF
--- a/src/gui/qgsexpressionbuilderwidget.cpp
+++ b/src/gui/qgsexpressionbuilderwidget.cpp
@@ -1432,20 +1432,20 @@ void QgsExpressionBuilderWidget::exportUserExpressions_pressed()
 
   settings.setValue( QStringLiteral( "lastExportExpressionsDir" ), saveFileInfo.absolutePath(), QgsSettings::App );
 
-  std::unique_ptr< QJsonDocument > exportJson = exportUserExpressions();
+  QJsonDocument exportJson = exportUserExpressions();
   QFile jsonFile( saveFileName );
 
   if ( !jsonFile.open( QFile::WriteOnly | QIODevice::Truncate ) )
     QMessageBox::warning( this, tr( "Export user expressions" ), tr( "Error while creating the expressions file." ) );
 
-  if ( ! jsonFile.write( exportJson->toJson() ) )
+  if ( ! jsonFile.write( exportJson.toJson() ) )
     QMessageBox::warning( this, tr( "Export user expressions" ), tr( "Error while creating the expressions file." ) );
   else
     jsonFile.close();
 }
 
 
-std::unique_ptr< QJsonDocument > QgsExpressionBuilderWidget::exportUserExpressions()
+QJsonDocument QgsExpressionBuilderWidget::exportUserExpressions()
 {
   const QString group = QStringLiteral( "user" );
   QgsSettings settings;
@@ -1482,7 +1482,7 @@ std::unique_ptr< QJsonDocument > QgsExpressionBuilderWidget::exportUserExpressio
   }
 
   exportObject["expressions"] = exportList;
-  std::unique_ptr< QJsonDocument > exportJson = qgis::make_unique< QJsonDocument >( exportObject );
+  QJsonDocument exportJson = QJsonDocument( exportObject );
 
   return exportJson;
 }
--- a/src/gui/qgsexpressionbuilderwidget.cpp
+++ b/src/gui/qgsexpressionbuilderwidget.cpp
@@ -69,8 +69,8 @@ QgsExpressionBuilderWidget::QgsExpressionBuilderWidget( QWidget *parent )
   connect( btnSaveExpression, &QPushButton::pressed, this, &QgsExpressionBuilderWidget::storeCurrentUserExpression );
   connect( btnEditExpression, &QPushButton::pressed, this, &QgsExpressionBuilderWidget::editSelectedUserExpression );
   connect( btnRemoveExpression, &QPushButton::pressed, this, &QgsExpressionBuilderWidget::removeSelectedUserExpression );
-  connect( mActionImportUserExpressions, &QAction::triggered, this, &QgsExpressionBuilderWidget::importUserExpressions_pressed );
-  connect( mActionExportUserExpressions, &QAction::triggered, this, &QgsExpressionBuilderWidget::exportUserExpressions_pressed );
+  connect( btnImportExpressions, &QPushButton::pressed, this, &QgsExpressionBuilderWidget::importUserExpressions_pressed );
+  connect( btnExportExpressions, &QPushButton::pressed, this, &QgsExpressionBuilderWidget::exportUserExpressions_pressed );
   connect( btnClearEditor, &QPushButton::pressed, txtExpressionString, &QgsCodeEditorExpression::clear );
 
   txtHelpText->setOpenExternalLinks( true );
@@ -94,7 +94,8 @@ QgsExpressionBuilderWidget::QgsExpressionBuilderWidget( QWidget *parent )
   btnSaveExpression->setIcon( QgsApplication::getThemeIcon( QStringLiteral( "mActionFileSave.svg" ) ) );
   btnEditExpression->setIcon( QgsApplication::getThemeIcon( QStringLiteral( "mActionToggleEditing.svg" ) ) );
   btnRemoveExpression->setIcon( QgsApplication::getThemeIcon( QStringLiteral( "mActionDeleteSelected.svg" ) ) );
-  btnImportExportExpressions->setIcon( QgsApplication::getThemeIcon( QStringLiteral( "mActionFileSaveAs.svg" ) ) );
+  btnExportExpressions->setIcon( QgsApplication::getThemeIcon( QStringLiteral( "mActionSharingExport.svg" ) ) );
+  btnImportExpressions->setIcon( QgsApplication::getThemeIcon( QStringLiteral( "mActionSharingImport.svg" ) ) );
   btnClearEditor->setIcon( QgsApplication::getThemeIcon( QStringLiteral( "mActionFileNew.svg" ) ) );
 
   expressionTree->setContextMenuPolicy( Qt::CustomContextMenu );
@@ -1377,7 +1378,9 @@ void QgsExpressionBuilderWidget::editSelectedUserExpression()
        ( item->parent() && item->parent()->text() != mUserExpressionsGroupName ) )
     return;
 
-  QgsExpressionStoreDialog dlg { item->text(), item->getExpressionText(), item->getHelpText() };
+  QgsSettings settings;
+  QString helpText = settings.value( QStringLiteral( "user/%1/helpText" ).arg( item->text() ), "", QgsSettings::Section::Expressions ).toString();
+  QgsExpressionStoreDialog dlg { item->text(), item->getExpressionText(), helpText };
 
   if ( dlg.exec() == QDialog::DialogCode::Accepted )
   {
@@ -1490,11 +1493,11 @@ QJsonDocument QgsExpressionBuilderWidget::exportUserExpressions()
 void QgsExpressionBuilderWidget::importUserExpressions_pressed()
 {
   QgsSettings settings;
-  QString lastSaveDir = settings.value( QStringLiteral( "lastImportExpressionsDir" ), QDir::homePath(), QgsSettings::App ).toString();
+  QString lastImportDir = settings.value( QStringLiteral( "lastImportExpressionsDir" ), QDir::homePath(), QgsSettings::App ).toString();
   QString loadFileName = QFileDialog::getOpenFileName(
                            this,
                            tr( "Import User Expressions" ),
-                           lastSaveDir,
+                           lastImportDir,
                            tr( "User expressions" ) + " (*.json)" );
 
   if ( loadFileName.isEmpty() )
--- a/src/gui/qgsexpressionbuilderwidget.cpp
+++ b/src/gui/qgsexpressionbuilderwidget.cpp
@@ -258,7 +258,7 @@ void QgsExpressionBuilderWidget::currentChanged( const QModelIndex &index, const
 
   // Get the item
   QModelIndex idx = mProxyModel->mapToSource( index );
-  QgsExpressionItem *item = dynamic_cast<QgsExpressionItem *>( mModel->itemFromIndex( idx ) );
+  QgsExpressionItem *item = static_cast<QgsExpressionItem *>( mModel->itemFromIndex( idx ) );
   if ( !item )
     return;
 
@@ -355,8 +355,8 @@ void QgsExpressionBuilderWidget::updateFunctionFileList( const QString &path )
   {
     // Create default sample entry.
     newFunctionFile( "default" );
-    txtPython->setText( QString( "'''\n#Sample custom function file\n "
-                                 "(uncomment to use and customize or Add button to create a new file) \n%1 \n '''" ).arg( txtPython->text() ) );
+    txtPython->setText( QStringLiteral( "'''\n#Sample custom function file\n "
+                                        "(uncomment to use and customize or Add button to create a new file) \n%1 \n '''" ).arg( txtPython->text() ) );
     saveFunctionFile( "default" );
   }
 }
@@ -416,7 +416,7 @@ void QgsExpressionBuilderWidget::loadFunctionCode( const QString &code )
 void QgsExpressionBuilderWidget::expressionTree_doubleClicked( const QModelIndex &index )
 {
   QModelIndex idx = mProxyModel->mapToSource( index );
-  QgsExpressionItem *item = dynamic_cast<QgsExpressionItem *>( mModel->itemFromIndex( idx ) );
+  QgsExpressionItem *item = static_cast<QgsExpressionItem *>( mModel->itemFromIndex( idx ) );
   if ( !item )
     return;
 
@@ -1250,7 +1250,7 @@ void QgsExpressionBuilderWidget::showContextMenu( QPoint pt )
 {
   QModelIndex idx = expressionTree->indexAt( pt );
   idx = mProxyModel->mapToSource( idx );
-  QgsExpressionItem *item = dynamic_cast<QgsExpressionItem *>( mModel->itemFromIndex( idx ) );
+  QgsExpressionItem *item = static_cast<QgsExpressionItem *>( mModel->itemFromIndex( idx ) );
   if ( !item )
     return;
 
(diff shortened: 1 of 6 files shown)


elpaso:
Nice! Can you add a screencast or a couple of pictures of the new GUI?

elpaso:
Oh, sorry I didn't follow the link, you could actually use "peek" to record in GIF format and attach it to the PR description, this will be very useful for changelog and documentation.

https://github.com/phw/peek

suricactus:
I just tried peek and it's semi broken to me on KDE. How do you turn off the top bar, so it can record the whole window?

elpaso:
I'm actually using it on KDE with no issues but I never tried to record the whole window: you must choose the smallest possible region: github limit is 10 MB

suricactus:
Edit:
Peek 2020-03-11 17-35
Import/Export:
Peek 2020-03-11 17-37
Label conflict:
Peek 2020-03-11 17-45
Save as:
Peek 2020-03-11 17-47

elpaso:
The buttons row is a little crowded now, what do you think about splitting the toolbar in two rows, or moving the icon buttons to a vertical bar?

suricactus:
Clone:
Peek 2020-03-11 17-52
Edit:
Peek 2020-03-11 17-59
Import/export:
Peek 2020-03-11 18-00
Overwrite confirm:
Peek 2020-03-11 17-56

suricactus:
I think it is fine for now, maybe I can keep the save extension button on the toolbar and put the import/export as sub menu buttons (like they are now). The bigger problem is that the expressions code area is constantly resizing like in the first gif. Setting a new minimum size of the left side might help.

suricactus:
I've answered on one of the above mentioned issue pages the following:

> 2.3) Do you insist to keep the original storage implementation in the .ini file? I think working with JSON is much easier and we can store the user_expressions.json together with the other profile configuration files. Also this format mimics the one found in resources/function_help/json/*convex_hull*. Python functions are already stored outside the ~/.local/share/QGIS/QGIS3/profiles/default/QGIS/QGIS3.ini file.

What do you think?

elpaso:
I think that they must stay in the ini file, the reason is the global settings implementation that allows organizations to ship/install QGIS with pre-defined settings defaults, that includes the user expressions.

suricactus:
I've tried to combine them in a single menu button like in the screenshot, but I don't find any way to make the main button disabled, but to keep the menu drop down enabled. Any ideas, or we should keep it like this for now?
image

elpaso:
> I've tried to combine them in a single menu button like in the screenshot, but I don't find any way to make the main button disabled, but to keep the menu drop down enabled.

I don't know if that's possible.

> Any ideas, or we should keep it like this for now?

yeah, what happens if you hit save when it's supposed to be disabled?

roya0045:
Instead of overwriting save, could it just be a menu with the name "...", I can understand having export under save but import under save is counter-intuitive.

suricactus:
^ the screenshot from above has names instead icons, cause I made it from QT designer. This is my final idea, otherwise I leave it as it is: separate buttons for save and export:
image

suricactus:
Oh, @roya0045 I misunderstood you. Yes, it's a bit counter intuitive, but it should be located in the same place as the export.

src/gui/qgsexpressionbuilderwidget.cpp
```diff
@@ -1597,22 +1600,35 @@ void QgsExpressionBuilderWidget::loadExpressionsFromJson( const QJsonDocument &e
...
+  QgsLogger::warning( QString::number( skippedExpressionLabels.count() ) + QStringLiteral( __FILE__ ) + ": " + QString::number( __LINE__ ) );
+
+  if ( ! skippedExpressionLabels.isEmpty() )
+  {
+    QStringList skippedExpressionLabelsQuoted;
+    for ( const QString &skippedExpressionLabel : skippedExpressionLabels )
+      skippedExpressionLabelsQuoted.append( QString( "'%1'" ).arg( skippedExpressionLabel ) );
+
+    QMessageBox::information( this,
+                              tr( "Skipped expressions" ),
```
nyalldawson: Maybe "Invalid Expressions" would be better? (Also not title case capitalisation for dialog titles)

src/gui/qgsexpressionbuilderwidget.cpp
```diff
@@ -1320,6 +1356,306 @@ void QgsExpressionBuilderWidget::removeSelectedUserExpression()
...
+
+  QTextStream jsonStream( &jsonFile );
+  QString jsonString = jsonFile.readAll();
+  jsonFile.close();
+
+  QJsonDocument importJson = QJsonDocument::fromJson( jsonString.toUtf8() );
+
+  if ( importJson.isNull() )
+  {
+    QMessageBox::warning( this, tr( "Import user expressions" ), tr( "Error while reading the expressions file." ) );
```
nyalldawson: Title case for dialog title

src/gui/qgsexpressionbuilderwidget.cpp
```diff
@@ -1320,6 +1356,306 @@ void QgsExpressionBuilderWidget::removeSelectedUserExpression()
...
+void QgsExpressionBuilderWidget::showMessageBoxConfirmExpressionOverwrite(
+  bool &isApplyToAll,
+  bool &isOkToOverwrite,
+  const QString &label,
+  const QString &oldExpression,
+  const QString &newExpression )
+{
+  QMessageBox::StandardButtons buttons = QMessageBox::Yes | QMessageBox::YesToAll | QMessageBox::No | QMessageBox::NoToAll;
+  switch ( QMessageBox::question( this,
+                                  tr( "Expression overwrite" ),
```
nyalldawson: Title case

src/gui/qgsexpressionbuilderwidget.h
```diff
@@ -380,6 +380,55 @@ class GUI_EXPORT QgsExpressionBuilderWidget : public QWidget, private Ui::QgsExp
...
+     * \since QGIS 3.14
+     */
+    void exportUserExpressions_pressed();
+
+    /**
+     * Create the expressions JSON document storing all the user expressions to be exported.
+     * \returns the created expressions JSON file
+     * \since QGIS 3.14
+     */
+    QJsonDocument *exportUserExpressions();
```
nyalldawson: If this transfers ownership, it should return a std::unique_ptr

src/ui/qgsexpressionbuilder.ui
```diff
@@ -933,6 +961,19 @@ Saved scripts are auto loaded on QGIS startup.</string>
     </widget>
    </item>
   </layout>
+  <action name="mActionImportUserExpressions">
+   <property name="text">
+    <string>Import user expressions</string>
```
nyalldawson: Title case for action labels

src/gui/qgsexpressionbuilderwidget.cpp
suricactus: They could be perfectly valid expressions, just skipped to import because of label collision. More like "Not Imported Expressions"? "Skipped Imports"?

src/gui/qgsexpressionbuilderwidget.h
suricactus: Can you please review if I made proper use of `str::unique_ptr`.

roya0045:
> Oh, @roya0045 I misunderstood you. Yes, it's a bit counter intuitive, but it should be located in the same place as the export.

I think that the best option would be to set the buttons in two rows, one for expression operators, the other for edit, save, export import (with dedicated import and export buttons). 
This would separate the management button and the expression buttons.
Maybe the management buttons could be beside the help or below the preview.

suricactus:
There is a problem that the buttons for clearing the expressions text area and saving make perfectly sense to stay where they are, while the rest play along with the second column. While doing the screenshots bellow, I was thinking of putting the operators below? 

So we either separate them in two columns:
Optimal size:
image
Minimal size:
image

Or two rows:
Optimal size:
image
Minimal size:
image

I would love to make whatever solution is chosen, but I give up on producing mock ups for now.

roya0045:
I'd be more inclined to go with the row option but inverse the order, not because it makes sense but mostly out of habit and since the expression insert are closer to where the cursor would be when editing an expression (the top of the text zone). @nirvn @DelazJ  opinion on this?

src/gui/qgsexpressionbuilderwidget.h
```diff
@@ -380,6 +380,55 @@ class GUI_EXPORT QgsExpressionBuilderWidget : public QWidget, private Ui::QgsExp
...
+     * \since QGIS 3.14
+     */
+    void exportUserExpressions_pressed();
+
+    /**
+     * Create the expressions JSON document storing all the user expressions to be exported.
+     * \returns the created expressions JSON file
+     * \since QGIS 3.14
+     */
+    std::unique_ptr< QJsonDocument > exportUserExpressions();
```
elpaso: I don't think you need a unique ptr here: QJsonDocument is implictly shared and even if it wasn't RVO would most probably kick in ad elide the copy. 
You can  safely and efficiently return by value here.

suricactus: You can have a look now.

src/gui/qgsexpressionbuilderwidget.cpp
```diff
@@ -1296,6 +1310,28 @@ void QgsExpressionBuilderWidget::storeCurrentUserExpression()
   }
 }
 
+void QgsExpressionBuilderWidget::editSelectedUserExpression()
+{
+  // Get the item
+  QModelIndex idx = mProxyModel->mapToSource( expressionTree->currentIndex() );
+  QgsExpressionItem *item = dynamic_cast<QgsExpressionItem *>( mModel->itemFromIndex( idx ) );
```
elpaso:
```suggestion
  QgsExpressionItem *item = static_cast<QgsExpressionItem *>( mModel->itemFromIndex( idx ) );
```

src/gui/qgsexpressionbuilderwidget.cpp
```diff
@@ -1320,6 +1356,302 @@ void QgsExpressionBuilderWidget::removeSelectedUserExpression()
...
+
+    // we want to import only items of type expression for now
+    if ( expressionObj["type"].toString() != "expression" )
+    {
+      skippedExpressionLabels.append( expressionObj["name"].toString() );
+      continue;
+    }
+
+    // we want to import only items of type expression for now
+    if ( expressionObj["group"].toString() != "user" )
```
elpaso:
```suggestion
    if ( expressionObj["group"].toString() != QStringLiteral( "user" ) )
```

src/gui/qgsexpressionbuilderwidget.cpp
```diff
@@ -1320,6 +1356,302 @@ void QgsExpressionBuilderWidget::removeSelectedUserExpression()
...
+      skippedExpressionLabels.append( expressionObj["name"].toString() );
+      continue;
+    }
+
+    const QString label = expressionObj["name"].toString();
+    const QString expression = expressionObj["expression"].toString();
+    const QString helpText = expressionObj["description"].toString();
+
+    // make sure they have valid name
+    if ( label.contains( "\\" ) || label.contains( "/" ) )
```
elpaso:
```suggestion
    if ( label.contains( "\\" ) || label.contains( '/' ) )
```

src/gui/qgsexpressionbuilderwidget.cpp
```diff
@@ -1320,6 +1356,302 @@ void QgsExpressionBuilderWidget::removeSelectedUserExpression()
...
+    }
+  }
+
+  loadUserExpressions( );
+
+  if ( ! skippedExpressionLabels.isEmpty() )
+  {
+    QStringList skippedExpressionLabelsQuoted;
+    for ( const QString &skippedExpressionLabel : skippedExpressionLabels )
+      skippedExpressionLabelsQuoted.append( QString( "'%1'" ).arg( skippedExpressionLabel ) );
```
elpaso:
```suggestion
      skippedExpressionLabelsQuoted.append( QStringLiteral( "'%1'" ).arg( skippedExpressionLabel ) );
```

same for other `QString("...")` except when it's an empty string, `QString()` is preferred in that case.

src/gui/qgsexpressionbuilderwidget.cpp
suricactus: I directly copied this from other places in the same file, where they used dynamic cast. I will convert all of them

src/gui/qgsexpressionbuilderwidget.h
elpaso: looks good

src/gui/qgsexpressionbuilderwidget.cpp
suricactus: Applied on all the files

elpaso:
Code is looking good!
My vote goes to the two-rows option (but with the toolbar left- aligned) but let's wait for @nirvn opinion.

roya0045:
@elpaso are you in favor of keeping expressions insertion button above and management buttons below or the opposite as presented in the screenshot?

elpaso:
@roya0045 I prefer the management buttons on top, the keyboard-like buttons are probably better placed below the editor, like we do in SQL editor dialog, in raster calculator and like all smartphones GUI do.

roya0045:
@suricactus Did you end up making two separate icons for import and export? With the two rows there should be enough place to have them displayed and not hidden anymore.

suricactus:
Nope, what you see in the gifs is the interface in the code.
These should be the icons to be used:
https://github.com/qgis/QGIS/blob/master/images/themes/default/mActionSharingExport.svg
https://github.com/qgis/QGIS/blob/master/images/themes/default/mActionSharingImport.svg

I am waiting for @nirvn opinion how this should be changed.

roya0045:
@suricactus as long as there are dedicated icons for import and export now that there is sufficient place with the two rows I am fine with any icons.

elpaso:
@suricactus I tested your PR right now and it works well, thank you, great job here!

Before we merge there are two little things that could make the GUI/UX better:

1. split the toolbar in 2: one at the top with the `Blank, Edit, Save, Delete, Export, Import` buttons **in that order** and the other one at the bottom with the keyboard-like buttons (if it was my choice I would probably remove this second toolbar completely but I can already hear people screaming they want it back)

2. since after (1) horizontal space will not be an issue anymore, make the import/export button separate and choose better icons for them (the floppy is not appropriate for import or export IMO).

3. if you want to gain extra points (maybe in a separate PR) you should move the import/export functions into `core` library: I see a great potential in sharing the expressions through the resource sharing plugin (CC @havatv) and having the import/export login in core and accessible to Python would definitely make it possible.

suricactus:
1.1) I would suggest the order to be:
`Blank`, `Save`, `|`, `Edit`, `Delete`, `Export`, `Import` 
Blank and Save are related to the expression box alone, while the rest are related to the already existing user expressisons.
1.2) It is nice for newcomers until there is a better way to introduce them to expressions. (also nice if you are lazy and drink tee with the other hand)
2) https://github.com/qgis/QGIS/pull/34997#issuecomment-598697781
3) I will make it a separate PR, otherwise this becomes too big and unfocused.

elpaso:
> 1.1) I would suggest the order to be:
> `Blank`, `Save`, `|`, `Edit`, `Delete`, `Export`, `Import`
> Blank and Save are related to the expression box alone, while the rest are related to the already existing user expressisons.

makes sense

> 1.2) It is nice for newcomers until there is a better way to introduce them to expressions. (also nice if you are lazy and drink tee with the other hand)

ok

> 2) [#34997 (comment)](https://github.com/qgis/QGIS/pull/34997#issuecomment-598697781)

yes, those icons look perfect.

> 3) I will make it a separate PR, otherwise this becomes too big and unfocused.

ok

suricactus:
image

havatv:
>     3. if you want to gain extra points (maybe in a separate PR) you should move the import/export functions into `core` library: I see a great potential in sharing the expressions through the resource sharing plugin (CC @havatv) and having the import/export login in core and accessible to Python would definitely make it possible.

Support for expressions in QGIS resource sharing would be welcome. We could use a sub-directory with the name "expressions" for the purpose (complementing the "image", "models", "processing", "rscripts", "svg" and "symbol" directories).
"Forcing" expressions to be saved to a default file system location under the user's QGIS directory (R scripts have their `processing/rscripts` directory, models have their `processing/models` directory, ...) would make them easy to handle by the Resource Sharing plugin. This location could be overridden by the user in settings.

elpaso:
@suricactus looks great, we are almost there! 

Can you please left-align the top toolbar or (maybe better) leave the import/export on the right and put the rest on the left?

suricactus:
@elpaso let's move all the buttons to the left, most of the QGIS interfaces use this arrangement already and I don't want to bring even more visual inconsistencies.

A few directions to think for the future:
1) Add support for custom expression groups to organize the user expressions. One can later export and share only a specific group.
2) What @havatv said above is more or less what I suggested in the other thread. This would simplify how the user expressions are stored and managed within the QGIS code too.

image

Peek 2020-03-17 08-45

roya0045:
@suricactus Good work on this PR!

pigreco:
@suricactus 

wow
it was what I dreamed of for years
great job and thanks for the implementation

pigreco:
Please add Feature label

havatv:
Expressions are supported in the just released version 0.15.0 of the QGIS Resource Sharing plugin if they area placed in the *expressions* folder of a collection.
Expressions are made available in *User expressions* in the expressions dialogue.

PR #34999, opened 2020-03-11
[BUG][MESH][3D] fix enable/disable mesh 3D rendering

### LINKED ISSUE #34985: [mesh] The 3d renderer checkbox is unchecked by default with no effect
Add a mesh layer
Open the layer properties dialog
Enable the 3d view tab
The "Enable 3D renderer" is unchecked while sub items it should disable are enabled
image

### LINKED ISSUE #34986: [Mesh] Disabling the "enable 3D renderer" should not reset the user settings in the dialog
1. Add a mesh layer
2. Open layer properties dialog
3. Enable the 3d view dialog
4. Check the "Enable 3d renderer" button and modify some settings, eg set the vertical scale from 1 to 2
5. Apply (or OK, reopen the layer properties to do it the hard way)
6. Disable the "Enable 3D renderer" setting, OK.
7. Reopen: you get #34985 but you'll notice that settings you made in the dialog are wiped, reset to initial values of the dialog (ie, our vertical scale displays 1 and not 2).

Disabling the box can be a temporary action and we shouldn't reset the configuration the user made; these are supposed to be disabled anyways so shouldn't harm anything if left customized (and resetting is not consistent with the other parts of the properties dialog)

vcloarec:
This PR fixes #34985 and fixes #34986 
Thanks to @DelazJ for trying and reporting issue.

### DIFF
--- a/src/3d/symbols/qgsmesh3dsymbol.h
+++ b/src/3d/symbols/qgsmesh3dsymbol.h
@@ -81,14 +81,14 @@ class _3D_EXPORT QgsMesh3DSymbol : public QgsAbstract3DSymbol
      *
      * \since QGIS 3.14
      */
-    bool renderer3dEnabled() const;
+    bool isEnabled() const;
 
     /**
      * Sets if the 3d rendering is enabled
      *
      * \since QGIS 3.14
      */
-    void setRenderer3dEnabled( bool mRenderer3dEnabled );
+    void setEnabled( bool enabled );
 
     //! Returns method that determines altitude (whether to clamp to feature to terrain)
     Qgs3DTypes::AltitudeClamping altitudeClamping() const { return mAltClamping; }
@@ -323,7 +323,7 @@ class _3D_EXPORT QgsMesh3DSymbol : public QgsAbstract3DSymbol
     QgsPhongMaterialSettings mMaterial;  //!< Defines appearance of objects
     bool mAddBackFaces = false;
 
-    bool mRenderer3dEnabled = true;
+    bool mEnabled = true;
 
     //! Triangles settings
     bool mSmoothedTriangles = false;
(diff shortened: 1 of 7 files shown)


src/app/3d/qgsmesh3dsymbolwidget.h
```diff
@@ -57,6 +58,7 @@ class QgsMesh3dSymbolWidget : public QWidget, private Ui::QgsMesh3dPropsWidget
     void setColorRampMinMax( double min, double max );
     QgsMeshLayer *mLayer = nullptr;
     QgsMeshDatasetGroupListModel *mDatasetGroupListModel = nullptr;
+    QgsMesh3DSymbol mSymbol; // used to store and returns the settings that are not handled by the widget
```
PeterPetrik: the comment is not clear...

src/3d/symbols/qgsmesh3dsymbol.h
```diff
@@ -76,6 +76,20 @@ class _3D_EXPORT QgsMesh3DSymbol : public QgsAbstract3DSymbol
     void writeXml( QDomElement &elem, const QgsReadWriteContext &context ) const override;
     void readXml( const QDomElement &elem, const QgsReadWriteContext &context ) override;
 
+    /**
+     * Returns if the 3d rendering is enabled
+     *
+     * \since QGIS 3.14
+     */
+    bool renderer3dEnabled() const;
```
PeterPetrik: would it be possible use enabled() isEnabled() etc consistenty with e.g. https://github.com/qgis/QGIS/blob/b510ad0803383e62a3022ff84914b980df68c9fd/src/core/mesh/qgsmeshrenderersettings.h#L44

DelazJ:
Backport to LTR?

vcloarec:
No, I am very not sure, there are a lot of changes since 3.10. The new 3D mesh rendering available since 3.12.

DelazJ:
> The new 3D mesh rendering available since 3.12

image
Maybe I misunderstand but above is a 3.10.3 (old but I'm not sure the code has changed) image showing the 3d mesh already there and with the issue

vcloarec:
Indeed this bug is on 3.10, but maybe I was too lazy because it could not be backported automatically .. now is done manually with  #36596

PR #35003, opened 2020-03-11
add missing QgsProcessingAlgorithm::parameterAsDateTime

3nids:


nyalldawson:
Thanks! Can you also reference this from the data time parameter dox? (Add th "..should be evaluated using..." note which is used elsewhere)

3nids:
@nyalldawson here you go

nyalldawson (approved):

PR #35008, opened 2020-03-12
Adds support to sort the plugins by update date; Show the plugin version update time

### LINKED ISSUE #22416: Add tools to organize/tag plugins
Author Name: **Andrew McAninch** (@spaceof7)
Original Redmine Issue: [14437](https://issues.qgis.org/issues/14437)

Redmine category:unknown

---
With the hundreds of plugins available I think adding some tools to manage plugins would be really useful.  There are a good number of plugins that are regional or are not at all relevant to the sort of work I do in a GIS and being able to tag these as hidden would be helpful for reducing the number of plugins I need to sort through when I'm searching for a tool.  I also will see new plugins that I don't have an immediate use for but I know will be useful in the future, and Id like a way to be able to tag these too.  I suppose I'm looking for the ability to manage/view different lists/sets of plugins.

suricactus:
- Added a new sort option for plugins: sort by date updated.
- Next to the version one can see the timestamp when it was released.

Fixes https://github.com/qgis/QGIS/issues/22416

Peek 2020-03-12 06-44

The code also support sorting by date created, but is currently not enabled. The reason for that is bug in the plugin repository server, as both "create_date" and "update_date" fields hold the same "update_date" data. I have filled a bug report and once it is fixed, the sort option should be enabled.

https://github.com/qgis/QGIS-Django/issues/69


src/app/pluginmanager/qgspluginmanager.cpp
```diff
@@ -977,12 +1016,34 @@ void QgsPluginManager::showPluginDetails( QStandardItem *item )
       downloadUrl = downloadUrl.replace( QStringLiteral( "download/" ), QString() );
     }
 
-    html += QStringLiteral( "<tr><td class='key'>%1 </td><td title='%2'><a href='%2'>%3</a></td></tr>"
+    QString dateUpdatedStr( "" );
```
nyalldawson:
```suggestion
    QString dateUpdatedStr;
```

suricactus:
@nyalldawson ping

roya0045:
@suricactus don't forget to fix the conflicts if you aren't on the case already.

suricactus:
@roya0045 oh, I just saw there are conflicts, thanks!

suricactus:
Resolved conflicts with the newly added stable/experimental selection. Should be ready for merge now.
Peek 2020-03-30 04-55

m-kuhn (approved):

m-kuhn:
Very nice!

agiudiceandrea:
@suricactus it seems this features doesn't work anymore.
The "updated at" date is missing in the plugin details for all the plugins and the "Sort by Date Updated" sorting method doesn't work as expected.

suricactus:
@agiudiceandrea hey, thanks for noticing. I am pretty sure there is not a lot that can be done on QGIS side if the information is missing in `plugins.xml`. I will be able to have a look later next month why it got broken.

PR #35009, opened 2020-03-12
Trim (QString::simplified) the query from the locator widget

### LINKED ISSUE #33935: Ignore Spaces and Tabs in Search In the Status Bar
Feature description.
QGIS would rock even more if it would ignore spaces in the search bar in the Status Bar. When you copy something from an excel cell and paste it in the search bar, QGIS would put a space at the end and would not find anything. It would be cool if it doesn't do that and/or ignore all spaces or atleast leading and trailing spaces.

Additional context
See here it did not find anything:
image

Found a pipe after I deleted the space:
image

suricactus:
Fixes: https://github.com/qgis/QGIS/issues/33935
Peek 2020-03-12 07-11


suricactus:
I will open a new PR, as things changed

PR #35014, opened 2020-03-12
fix in german translation

volaya:
minor fix in DE translation file


arnaud-morvan:
Hello @volaya , I'm pretty sure this translations are done through Transifex.
I'm translator only for french, but I found this text here : https://www.transifex.com/qgis/QGIS/translate/#fr/qgis-application/137005888?q=text%3A'Close+Project'
So this might be overwritten by : https://github.com/qgis/QGIS/blob/master/scripts/pull_ts.sh

jef-n:
Fixed in transifex

PR #35015, opened 2020-03-12
[server] Use request.data() instead of parameter REQUEST_BODY

arnaud-morvan:
#34701  Description

Better fix than https://github.com/qgis/QGIS/pull/34268

Use `request.data()` REQUEST_BODY parameter to create `QDomDocument` in POST requests.

Parameter REQUEST_BODY is not used anymore except in tests, but I've not removed it in case some plugins use it, but it might be completely removed in future version.


arnaud-morvan:
@pblottiere @elpaso: here is the fix for POST request BODY handling, it was simpler that I thought, request already have a data() method.

rldhont:
@arnaud-morvan can you fix test ?

rldhont (approved):

rldhont:
Tests ok

Issue #32888, opened 2019-11-15
[Forms] Layer fields are not available when putting an expression for default values


ponceta:
**Describe the bug**

**How to Reproduce**

Any vector Layer
Any field in form
Default value
Use expression
Layer fields are not available

![image](https://user-images.githubusercontent.com/8614223/68944852-33ec0b00-07af-11ea-8d78-2a5ff3f88d4f.png)

When for constraint :

![image](https://user-images.githubusercontent.com/8614223/68944888-4cf4bc00-07af-11ea-90e8-2ed5c8ae7401.png)

**QGIS and OS versions**

Tested on QGIS 3.4.13, 3.10.0 

Win 7-10 x64 

**Additional context**

roya0045:
I think this was mostly to prevent people from simply using the field itself. This is probably not a bug but to prevent a bug. Though I agree that this seems like an inconvenience.

gioman:
> This is probably not a bug

I also think is by design.

nyalldawson:
Duplicate of #30034

ponceta:
The group is **fields and values.**

Fields are NULL, it is non-sense to add them, I agree.

But for values it would be nice to have the selected attribute relationnal values.

PR #35030, opened 2020-03-12
fix 'fields and values' not accessible in expression editor in default values

3nids:
even if the expression generator is set, the expression dialog needs a pointer to the vector layer to behave es expected

### DIFF
--- a/src/ui/attributeformconfig/qgsattributetypeedit.ui
+++ b/src/ui/attributeformconfig/qgsattributetypeedit.ui
@@ -235,7 +235,7 @@
          <item>
           <widget class="QLabel" name="label_7">
            <property name="text">
-            <string>The expression is using fields on insert. This behavior is not guaranteed.</string>
+            <string>Using fields in a default value expression only works if "Apply default value on update" is checked.</string>
            </property>
           </widget>
          </item>
--- a/src/ui/attributeformconfig/qgsattributetypeedit.ui
+++ b/src/ui/attributeformconfig/qgsattributetypeedit.ui
@@ -184,36 +184,65 @@
      <property name="title">
       <string>Defaults</string>
      </property>
-     <layout class="QGridLayout" name="gridLayout_2" columnstretch="0,1">
-      <item row="1" column="0">
+     <layout class="QGridLayout" name="gridLayout_2" columnstretch="0,0,0">
+      <item row="4" column="0">
        <widget class="QLabel" name="label_3">
         <property name="text">
          <string>Default value</string>
         </property>
        </widget>
       </item>
-      <item row="3" column="1">
-       <widget class="QLabel" name="mDefaultPreviewLabel">
-        <property name="text">
-         <string/>
-        </property>
-       </widget>
-      </item>
-      <item row="1" column="1">
-       <widget class="QgsExpressionLineEdit" name="mExpressionWidget" native="true">
-        <property name="focusPolicy">
-         <enum>Qt::StrongFocus</enum>
-        </property>
-       </widget>
-      </item>
-      <item row="3" column="0">
+      <item row="6" column="0">
        <widget class="QLabel" name="label_4">
         <property name="text">
          <string>Preview</string>
         </property>
        </widget>
       </item>
-      <item row="4" column="0" colspan="2">
+      <item row="4" column="1">
+       <widget class="QgsExpressionLineEdit" name="mExpressionWidget" native="true">
+        <property name="focusPolicy">
+         <enum>Qt::StrongFocus</enum>
+        </property>
+       </widget>
+      </item>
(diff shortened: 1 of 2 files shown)


3nids:
the other idea would be to get the layer from the context if the layer is null in the dialog constructor
see #35031

Gustry:
@3nids Very similar issue, I noticed in master `Field and values` is not accessible too in the visibility by expression for a container.
It's working well on QGIS 3.10.

m-kuhn:
It was left out on purpose. Field values are not available when calculating default values.

Should this be done only when "Refresh on Update" is checked? Or are there other possibilities when using values from the current form does make sense?

roya0045:
See #32888

3nids:
there is now a warning if an expression uses fields on insert:

Screen-Recording-2020-03-12-at-1

signedav (approved):

3nids:
since the gif is a bit too fast, here is the text:

> The expression is using fields on insert. This behavior is not guaranteed.

nyalldawson:
What does "is not guaranteed" mean? It should say "will be totally broken"!

m-kuhn:
I would say "Using fields in a default value expression only works if apply default values on update is checked"

This leaves open if someone finds a scenario in which it works for him for whatever reason and also avoids using the term "broken" in our beloved application.

nyalldawson:
@m-kuhn

> This leaves open if someone finds a scenario in which it works for him 

That situation is impossible! There's no feature set at expression evaluation time for this, so it will NEVER work...

nyalldawson:
Being more constructive, maybe we should only show the fields in the dialog when "on update" is checked. That at least will make it less likely for user confusion.

And then we add the strongly worded warning just in case a user changes uses "on update", creates an expression with fields, then unchecks "on update"..

m-kuhn:
I didn't revise how it works nowadays.
Some scenarios I could imagine that it works (without having groundtruthed it):

- There is a possibility to provide a feature to a new attribute form. Maybe now or in the future attributes specified in there can be taken into account.
- Maybe the order of initialization of the default value matters (which is very close to UB)

-----------

The proposal to only show it if the checkbox has been activated has been rejected in internal discussions at the laax hackfest because

- The intuitive order to fill the widgets is from top to bottom, meaning the "update" checkbox will naturally be updated after filling the expression and there is nothing that will tell a user he should do otherwise
- This makes it look like a system with random behavior which normally does not increase my trust in an application

nyalldawson:
> Some scenarios I could imagine that it works (without having groundtruthed it):

But neither of these reflect current reality. In current master, you can NEVER use field values as part of non-on-update defaults.

> This makes it look like a system with random behavior which normally does not increase my trust in an application

I agree with this. And my concern is that just saying "not guaranteed" we are lying to users and going to cause a whole lot of wasted time and frustration. A user will likely keep smashing away at their expression thinking THEY are doing something wrong, when they'll actually never be able to fix their expression...

nyalldawson:
To take an extreme example, I think this message is just as destructive as it would be if we added a button on the toolbar which did nothing but show a message box "this button action is not guaranteed. Keep pressing it, maybe it'll work sometime!"

roya0045:
I agree with Nyall on this, usually I'm in favor of letting the users do more and not baby-proofing thing but this is taking things too far as most use case will result in confusion due to a seemingly inconsistent behavior.

m-kuhn:
Now I am a bit confused. I agree with both of you that we shouldn't use "not guaranteed" and we shouldn't have a confusing behavior. Can we still try to find something that does not involve the term "broken" as I tried to come up with in https://github.com/qgis/QGIS/pull/35030#issuecomment-598343631?

src/ui/attributeformconfig/qgsattributetypeedit.ui
```diff
@@ -184,36 +184,65 @@
...
+           </property>
+           <property name="pixmap">
+            <pixmap resource="../../../images/images.qrc">:/images/themes/default/mIconWarning.svg</pixmap>
+           </property>
+          </widget>
+         </item>
+         <item>
+          <widget class="QLabel" name="label_7">
+           <property name="text">
+            <string>The expression is using fields on insert. This behavior is not guaranteed.</string>
```
3nids: following @m-kuhn's proposition
```suggestion
            <string>Using fields in a default value expression only works if apply default values on update is checked.</string>
```

nyalldawson: That's looking better:

"Using fields in a default value expression only works if "Apply default value on update" is checked."

3nids: thanks

PR #35031, opened 2020-03-12
in QgsExpressionBuilderDialog, if layer is null, try to get it from the context

3nids:
alternative approach for #35030


3nids:
is this still interesting?

m-kuhn:
I don't think so

PR #35035, opened 2020-03-12
Add description to value relation widget

m-kuhn:
Add the possibility to add a description to each value in a value relation widget.
The description will show up as tooltip

Please excuse the different languages in the demo video.

Peek 2020-03-12 16-28

### DIFF
--- a/src/core/fieldformatter/qgsvaluerelationfieldformatter.h
+++ b/src/core/fieldformatter/qgsvaluerelationfieldformatter.h
@@ -39,9 +39,10 @@ class CORE_EXPORT QgsValueRelationFieldFormatter : public QgsFieldFormatter
     struct ValueRelationItem
     {
       //! Constructor for ValueRelationItem
-      ValueRelationItem( const QVariant &key, const QString &value )
+      ValueRelationItem( const QVariant &key, const QString &value, const QString &description )
         : key( key )
         , value( value )
+        , description( description )
       {}
 
       //! Constructor for ValueRelationItem
@@ -49,6 +50,7 @@ class CORE_EXPORT QgsValueRelationFieldFormatter : public QgsFieldFormatter
 
       QVariant key;
       QString value;
+      QString description;
     };
 
     typedef QVector < QgsValueRelationFieldFormatter::ValueRelationItem > ValueRelationCache;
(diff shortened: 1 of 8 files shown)


src/gui/editorwidgets/qgsvaluerelationwidgetwrapper.cpp
```diff
@@ -365,6 +365,7 @@ void QgsValueRelationWidgetWrapper::populate( )
     for ( const QgsValueRelationFieldFormatter::ValueRelationItem &element : qgis::as_const( mCache ) )
     {
       whileBlocking( mComboBox )->addItem( element.value, element.key );
+      mComboBox->setItemData( mComboBox->count() - 1, element.description, Qt::ToolTipRole );
```
3nids: `element.description` might be an empty string. Better go with an invalid `QVariant` for the cases where no expression were given.

src/core/fieldformatter/qgsvaluerelationfieldformatter.h
```diff
@@ -39,16 +39,18 @@ class CORE_EXPORT QgsValueRelationFieldFormatter : public QgsFieldFormatter
     struct ValueRelationItem
     {
       //! Constructor for ValueRelationItem
-      ValueRelationItem( const QVariant &key, const QString &value )
+      ValueRelationItem( const QVariant &key, const QString &value, const QString &description )
```
3nids: I believe this is an API break
```suggestion
      ValueRelationItem( const QVariant &key, const QString &value, const QString &description = QString() )
```

3nids (approved):

src/gui/editorwidgets/qgsvaluerelationwidgetwrapper.cpp
m-kuhn:
```suggestion
      if ( !element.description.isEmpty() )
        mComboBox->setItemData( mComboBox->count() - 1, element.description, Qt::ToolTipRole );
```

ponceta:
Thank you @m-kuhn @3nids !

pigreco:
Please add `Widget` label

m-kuhn:
@mbernasocchi can you add @pigreco to the qgis/community group so he can manage issues and labels?

mbernasocchi:
> @mbernasocchi can you add @pigreco to the qgis/community group so he can manage issues and labels?

done

PR #35042, opened 2020-03-12
Temporal API followups

nyalldawson:
Some cleanups, some extra tests, and a fix to ensure that raster layers set to a fixed temporal range aren't rendered outside of that time range


nyalldawson:
Unrelated test failures

PR #35044, opened 2020-03-12
Fix for temporal layers refresh issue after temporal range update

Samweli:
This PR provide a fix for temporal layers to be refreshed when user has update the layer's temporal range through editing temporal layer properties.

### DIFF
--- a/python/core/auto_generated/qgstemporalproperty.sip.in
+++ b/python/core/auto_generated/qgstemporalproperty.sip.in
@@ -53,6 +53,11 @@ Returns ``True`` if the temporal property is active.
 Emitted when the temporal properties have changed.
 %End
 
+    void temporalRangeChanged( const QgsDateTimeRange &range );
+%Docstring
+Emitted when the temporal range in the properties have changed.
+%End
+
 };
 
 /************************************************************************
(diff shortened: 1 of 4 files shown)


nyalldawson:
Sorry, but I'm not happy with this approach. We currently have the QgsRasterLayerTemporalProperties class mixing two **very** different concepts:

1. Temporal properties relating to animations. Eg. TemporalMode, intervalHandlingMethod, fixedTemporalRange.
2. Settings which relate to which data to retrieve for a static map layer, e.g. temporalRange

So the two concepts need to be split from each other, because they have very different results in QGIS. We need to move this concept of static temporal ranges for wmst out of QgsRasterLayerTemporalProperties and keep it totally separate to the animation handling classes. 

I think what you are trying to achieve with this should instead be done through the uri of the raster layer -- what you want is to embed the static temporal range and static reference temporal range into the layer's source (i.e. its data provider uri) so that this remains unchanged regardless of what the current canvas' temporal range is.

Samweli:
>  what you want is to embed the static temporal range and static reference temporal range into the layer's source (i.e. its data provider uri) so that this remains unchanged regardless of what the current canvas' temporal range is.

Ok, if the uri remains unchanged how will the animation be able to update the layer?

nyalldawson:
Closing following discussion with @Samweli

PR #35047, opened 2020-03-13
Temporal dock UX follow ups & cleanups

nyalldawson:
Improves the UX of the new temporal dock and fixes a few UI issues.


nirvn:
Screenshot?

nyalldawson:
Peek 2020-03-13 10-37

nirvn:
Nice improvements.

That said, we need to do something about the minimum width of that panel as it is way too large ATM. See:
Peek 2020-03-13 08-16

IMHO, we should move the [ << ] [ < ] [ play/pause ] [ > ] [ >> ] controls down to sit next to the slider, there's no real need for the slider to be that wide anyways. It would cut down the minimum width by 1/4 or so.

nyalldawson:
> That said, we need to do something about the minimum width of that panel as it is way too large ATM.

I'd also suggest we could move more widgets into the settings panel, but let's do that in follow up PRs

nyalldawson:
Unrelated test failure

PR #35048, opened 2020-03-13
Fix some use-after-free issues in task manager test

nyalldawson:


nyalldawson:
Unrelated test failure

PR #35065, opened 2020-03-13
Mssql connections api

elpaso:
Implementation of the connections API for MSSQL

### DIFF
--- a/src/providers/mssql/qgsmssqlproviderconnection.cpp
+++ b/src/providers/mssql/qgsmssqlproviderconnection.cpp
@@ -391,6 +391,9 @@ QList<QgsMssqlProviderConnection::TableProperty> QgsMssqlProviderConnection::tab
     }
     else
     {
+      // Add an invalid column
+      table.addGeometryColumnType( QgsWkbTypes::Type::NoGeometry,
+                                   QgsCoordinateReferenceSystem() );
       table.setFlag( QgsMssqlProviderConnection::TableFlag::Aspatial );
     }
 
--- a/src/providers/mssql/qgsmssqlproviderconnection.cpp
+++ b/src/providers/mssql/qgsmssqlproviderconnection.cpp
@@ -22,6 +22,7 @@
 #include "qgsmssqlprovider.h"
 #include "qgsexception.h"
 #include "qgsapplication.h"
+#include "qgsmessagelog.h"
 
 
 const QStringList QgsMssqlProviderConnection::EXTRA_CONNECTION_PARAMETERS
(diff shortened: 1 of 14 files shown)


src/providers/mssql/qgsmssqlproviderconnection.cpp
```diff
@@ -0,0 +1,482 @@
...
+    table.setTableName( row[1].toString() );
+    table.setGeometryColumn( row[2].toString() );
+    //const QVariant srid { row[3] };
+    //const QVariant type { row[4] }; // GEOMETRY|GEOGRAPHY
+    if ( row[5].toBool() )
+      table.setFlag( QgsMssqlProviderConnection::TableFlag::View );
+
+    int geomColCount { 0 };
+
+    if ( ! table.geometryColumn().isEmpty() )
```
nyalldawson: How do you feel about adding a flag "GeometryColumnDetailsOptional" ="in cases where resolving the full geometry details of a table are expensive, setting this flag will skip fetching these details"

tests/src/python/test_qgsproviderconnection_base.py
```diff
@@ -184,23 +213,30 @@ def _test_operations(self, md, conn):
...
                 else:
                     self.assertEqual(res, [['QGIS Rocks - \U0001f604', 666, 1.234, 1234, QtCore.QDate(2019, 7, 8), QtCore.QDateTime(2019, 7, 8, 12, 0, 12)]])
-                sql = "SELECT time FROM %s" % table
+                sql = "SELECT time_t FROM %s" % table
                 res = conn.executeSql(sql)
-                self.assertIn(res, ([[QtCore.QTime(12, 0, 13)]], [['12:00:13.00']]))
-                sql = "DELETE FROM %s WHERE string = 'QGIS Rocks - \U0001f604'" % table
+
+                # This does not work in MSSQL and returns a QByteArray, we have no way to know that it is a time
+                # value and there is no way we can convert it.
```
nyalldawson: The provider iterator has code which does this -- but I guess the issue here is that you don't have the field type information to use to detect this situation?

nyalldawson (commented):
It's beautiful! 😭😀

nyalldawson:
Related question following on from our discussion yesterday: what do you think about adding a method to QgsProviderRegistry to retrieve a connection given a string like "postgres://my_conn_name" or "mssql://conn_name"?

elpaso:
> Related question following on from our discussion yesterday: what do you think about adding a method to QgsProviderRegistry to retrieve a connection given a string like "postgres://my_conn_name" or "mssql://conn_name"?

I think it's a good idea, just let's make it **always** lowercase, and get finally rid of **WFS**

We should also start a QEP about  `QgsDataSourceUri` refactoring.

tests/src/python/test_qgsproviderconnection_base.py
elpaso: I've investigated this issue for hours, I think there is a bug upstream in QT or in the ODBC driver, because according to QT docs `time` must be supported. And yes, we don't have a field here we can guess the type from, we could check the QByteArray size and try a reinterpret_cast conversion but it's far from being a solid approach.

src/providers/mssql/qgsmssqlproviderconnection.cpp
elpaso: That's a good idea, there are cases where you just need the table names and you don't care about other details. But I would leave this to a separate PR because it impacts all other providers.

tests/src/python/test_qgsproviderconnection_base.py
nyalldawson: Yep, leave it as is I say, it's the right decision

nyalldawson (approved):

PR #35070, opened 2020-03-13
Disable PyQgsCoordinateOperationWidget on azure

nyalldawson:
I **really** don't want to disable this one -- but it's failing on azure yet passing locally on Windows, so no way to fix for now.


jef-n:
> I **really** don't want to disable this one -- but it's failing on azure yet passing locally on Windows, so no way to fix for now.

Building against the proj-dev with proj-dev-data?  I guess azure pick up grid you don't have.

PR #35072, opened 2020-03-13
add a completer for the file widget when creating a new GPKG

3nids:
image

### DIFF
--- a/src/gui/qgsnewgeopackagelayerdialog.cpp
+++ b/src/gui/qgsnewgeopackagelayerdialog.cpp
@@ -132,7 +132,7 @@ QgsNewGeoPackageLayerDialog::QgsNewGeoPackageLayerDialog( QWidget *parent, Qt::W
     checkOk();
   } );
 
-  QgsProviderConnectionModel *ogrProviderModel = new QgsProviderConnectionModel( "ogr" );
+  QgsProviderConnectionModel *ogrProviderModel = new QgsProviderConnectionModel( QStringLiteral( "ogr" ) );
 
   QCompleter *completer = new QCompleter();
   completer->setModel( ogrProviderModel );
--- a/src/gui/qgsnewgeopackagelayerdialog.cpp
+++ b/src/gui/qgsnewgeopackagelayerdialog.cpp
@@ -132,9 +132,9 @@ QgsNewGeoPackageLayerDialog::QgsNewGeoPackageLayerDialog( QWidget *parent, Qt::W
     checkOk();
   } );
 
-  QgsProviderConnectionModel *ogrProviderModel = new QgsProviderConnectionModel( QStringLiteral( "ogr" ) );
+  QgsProviderConnectionModel *ogrProviderModel = new QgsProviderConnectionModel( QStringLiteral( "ogr" ), this );
 
-  QCompleter *completer = new QCompleter();
+  QCompleter *completer = new QCompleter( this );
   completer->setModel( ogrProviderModel );
   completer->setCompletionRole( QgsProviderConnectionModel::RoleUri );
   completer->setCompletionMode( QCompleter::PopupCompletion );
--- a/src/gui/qgsnewgeopackagelayerdialog.cpp
+++ b/src/gui/qgsnewgeopackagelayerdialog.cpp
@@ -36,6 +36,7 @@
 #include <QLineEdit>
 #include <QMessageBox>
 #include <QFileDialog>
+#include <QCompleter>
 
 #include <ogr_api.h>
 #include <ogr_srs_api.h>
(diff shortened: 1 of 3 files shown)


src/gui/qgsnewgeopackagelayerdialog.cpp
```diff
@@ -129,6 +130,17 @@ QgsNewGeoPackageLayerDialog::QgsNewGeoPackageLayerDialog( QWidget *parent, Qt::W
     }
     checkOk();
   } );
+
+  settings.beginGroup( "ogr/GPKG/connections", QgsSettings::Providers );
```
nyalldawson: Can you use the connections API instead of raw setting traversal?

3nids: do you have a pointer to the API? which class is it? cheers

nyalldawson: https://qgis.org/api/classQgsProviderMetadata.html#af560b01841d55eb53b430e92bb7aaac7

nyalldawson: Actually - you could probably just create the completer directly using QgsProviderConnectionModel.

nirvn:
Nice move. Would be great to see the same applied to the new spatialite layer dialog.

3nids:
here you go 
image

3nids:
I am just not sure if this needs filtering to gpkg.
I have an sqlite in my browser and it's not listed. Are now the spatialite always loaded using QGIS provider and not the OGR one?

src/gui/qgsnewgeopackagelayerdialog.cpp
```diff
@@ -129,6 +131,15 @@ QgsNewGeoPackageLayerDialog::QgsNewGeoPackageLayerDialog( QWidget *parent, Qt::W
     }
     checkOk();
   } );
+
+  QgsProviderConnectionModel *ogrProviderModel = new QgsProviderConnectionModel( "ogr" );
```
nyalldawson: QStringLiteral

nyalldawson: Parent to "this", same with completer

3nids:
```suggestion
  QgsProviderConnectionModel *ogrProviderModel = new QgsProviderConnectionModel( QStringLiteral( "ogr" ) );
```

elpaso:
> I am just not sure if this needs filtering to gpkg.

I think yes.

> I have an sqlite in my browser and it's not listed. Are now the spatialite always loaded using QGIS provider and not the OGR one?

For spatialite? No, you can use both `ogr` and `spatialite` providers, when using the file browser you load them with `ogr` and when you use the connections or the data source dialog you load them with `spatialite` provider (yes, I know it's silly, we should probably just drop the spatialite provider and use only OGR for it).

For GPKGs we always use OGR provider.

nirvn:
@elpaso , -1 to drop spatialite driver prior to going through cost of doing so. A little over a year ago, that cost included a speed drop that would arguably not be acceptable. 

We'd never consider dropping postgresql provider in favor of ogr either :)

elpaso:
> @elpaso , -1 to drop spatialite driver prior to going through cost of doing so. A little over a year ago, that cost included a speed drop that would arguably not be acceptable.

Interesting, is there any metric available?

> 
> We'd never consider dropping postgresql provider in favor of ogr either :)

That's not a fair comparison :)

BTW look at who's maintaining it: I'm not dropping anything (even if I'd like to).

nirvn:
@elpaso , yeah , I know you've given your fair share of sweat on this provider. 

The comparison to postgresql isn't fair but the basic argument remains: we have those providers for reasons and advantages that needs to be weighed seriously prior to taking decisions :)

No fresh metrics but easy enough to come up with using large datasets.

elpaso:
What really bothers me is that we are not running spatialite tests in travis.

3nids:
@elpaso how can I filter using QgsProviderConnectionModel? I need to create the provider or is there a way to find out (without looking at the filename extension)?

elpaso:
@3nids I don't know, Nyall wrote it, but I see there is a provider (key) in the constructor.

uclaros:
Could also be applied to `qgsnewspatialitelayerdialog.cpp`, `qgsnewvectorlayerdialog.cpp` and `qgsvectorlayersaveasdialog.cpp`?
See also my related comment on having different dialogs for similar actions https://github.com/qgis/QGIS/pull/34817#issuecomment-596206028

nyalldawson:
@3nids

> how can I filter using QgsProviderConnectionModel

Use a QSortFilterProxyModel

3nids:
@nyalldawson you wrote that ogr will strictly be gpgk, so no need to filter more, correct? if yes, ok to merge?
if not, my question was rather, am I forced to create the provider to know if it's a gpkg or a shapefile?
cheers

nyalldawson (approved):
Merge away!

nirvn:
@3nids , just to avoid benevolent coding overlap, I'm handling the new spatialite dialog now.

PR #35076, opened 2020-03-14
Data type support for QgsProcessingParameterMapLayer

alexbruy:
## Description
Allow to define accepted layer types for `QgsProcessingParameterMapLayer`. This is useful when algorithm can work with different layer types (for example, points and rasters) and allows to avoid duplicating same algorithm with different inputs.

This is an API improvement, not visible for vast majority of users.


alexbruy:
New argument `types` added to the parameter constructor. Probably to maintain consistency with other parameters, e.g. `QgsProcessingParameterFeatureSource`, better to have `types` not in the end but right after `description` argument. This will be an API break, but `QgsProcessingParameterMapLayer` used only in few algorithms changing its API should be safe.

I'm going to add native modeler dialog/wrapper in the follow-up pull-request.

src/core/processing/qgsprocessingparameters.cpp
```diff
@@ -2269,9 +2270,152 @@ QString QgsProcessingParameterMapLayer::valueAsPythonString( const QVariant &val
...
+    }
+  }
+  return QString();
+}
+
+QVariantMap QgsProcessingParameterMapLayer::toVariantMap() const
+{
+  QVariantMap map = QgsProcessingParameterDefinition::toVariantMap();
+  QVariantList types;
+  const auto constMDataTypes = mDataTypes;
```
nyalldawson: Same here, we could avoid the const copy

src/core/processing/qgsprocessingparameters.cpp
```diff
@@ -2269,9 +2270,152 @@ QString QgsProcessingParameterMapLayer::valueAsPythonString( const QVariant &val
...
+  }
+  map.insert( QStringLiteral( "data_types" ), types );
+  return map;
+}
+
+bool QgsProcessingParameterMapLayer::fromVariantMap( const QVariantMap &map )
+{
+  QgsProcessingParameterDefinition::fromVariantMap( map );
+  mDataTypes.clear();
+  QVariantList values = map.value( QStringLiteral( "data_types" ) ).toList();
```
nyalldawson:
```suggestion
  const QVariantList values = map.value( QStringLiteral( "data_types" ) ).toList();
```
Avoids the need for the extra const copy

nyalldawson (commented):
Looks good! Could you add some tests covering the gui changes too? (Eg make sure that the filtering is applied to the combo box)

src/core/processing/qgsprocessingparameters.cpp
```diff
@@ -2269,9 +2270,152 @@ QString QgsProcessingParameterMapLayer::valueAsPythonString( const QVariant &val
...
 }
 
+QString QgsProcessingParameterMapLayer::asScriptCode() const
+{
+  QString code = QStringLiteral( "##%1=" ).arg( mName );
+  if ( mFlags & FlagOptional )
+    code += QStringLiteral( "optional " );
+  code += QStringLiteral( "layer " );
+
+  const auto constMDataTypes = mDataTypes;
```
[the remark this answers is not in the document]
alexbruy: Done. Also fixed this in other places.

src/core/processing/qgsprocessingparameters.cpp
alexbruy: Done. Also fixed this in other places.

src/core/processing/qgsprocessingparameters.cpp
alexbruy: Done. Also fixed this in other places.

alexbruy:
Test added.

@nyalldawson what is your opinion about small API break: moving `types` argument in the parameter constructor right after `description` argument?

nyalldawson:
@alexbruy

> what is your opinion about small API break: moving types argument in the parameter constructor right after description argument?

I don't think we should -- there'll be more big breakage heading for processing plugins in 4.0 (due to removal of WidgetWrapper and the remaining python base classes), so I'd prefer to avoid any unnecessary disruption as far as possible..

nyalldawson (approved):

PR #35077, opened 2020-03-14
improve GTIFF JPEG compression by adding PHOTOMETRIC=YCBCR (fix #35002)

### LINKED ISSUE #35002: Improved Create Options when creating GTIFF using GDAL Translate and JPEG Compression
**Feature description.**

When creating a GTiff using GDAL Translate from processing or 'Export as...'   there is the option to select JPEG Compression. This sets create create options COMPRESS = JPEG and QUALITY = 75. 

According to the below link (which seems to work well for me) it would be worth also including the create options PHOTOMETRIC=YCBCR  and TILED=YES when jpeg compression is selected.

[http://blog.cleverelephant.ca/2015/02/geotiff-compression-for-dummies.html](http://blog.cleverelephant.ca/2015/02/geotiff-compression-for-dummies.html)

If this is a good idea, then it looks like it would be a very easy change in:
GIS/src/gui/qgsrasterformatsaveoptionswidget.cpp

alexbruy:
## Description
Improve JPEG compression for GTIFF format by using YCBCR color space. This reduces size of the image even more as JPEG compression works better with this color space.

Fixes #35002.

### DIFF
--- a/src/gui/qgsrasterformatsaveoptionswidget.cpp
+++ b/src/gui/qgsrasterformatsaveoptionswidget.cpp
@@ -71,7 +71,7 @@ QgsRasterFormatSaveOptionsWidget::QgsRasterFormatSaveOptionsWidget( QWidget *par
         << QStringLiteral( "COMPRESS=DEFLATE PREDICTOR=2 ZLEVEL=9" ) );
     sBuiltinProfiles[ QStringLiteral( "z_gtiff_4jpeg" )] =
       ( QStringList() << QStringLiteral( "GTiff" ) << tr( "JPEG Compression" )
-        << QStringLiteral( "COMPRESS=JPEG JPEG_QUALITY=75" ) );
+        << QStringLiteral( "COMPRESS=JPEG JPEG_QUALITY=75 PHOTOMETRIC=YCBCR" ) );
 
     // overview compression schemes for GTiff format, see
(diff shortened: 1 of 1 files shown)


nyalldawson:
Given https://github.com/qgis/QGIS/issues/35002#issuecomment-597804709, this lgtm!

nyalldawson (approved):

rouault:
Actually this change might not be appropriate as such: PHOTOMETRIC=YCBCR is only valid if the number of bands is 3. Otherwise GDAL will error out

nyalldawson:
Ahh, guess we better revert

angusmcb:
How about adding it as another profile? "JPEG Compression + YCBCR" for example.

There's already precedent for some of the create options breaking on particular filetypes, as the existing jpeg compression will only work on byte bands. Using YCBCR also seems like a very useful feature, and not easy to find if it's not clearly signposted; it took me a couple of years of GIS use before I discorvered it, and now I often see geotiffs not optimally compressed.

rouault:
Selecting YCBCR adds extra loss to a lossy compression scheme, mostly due to Cb and Cr channel being subsampled by a factor of 2. It is generally acceptable for images that must be "analyzed" only by the human eye which is less sensitive to colors than luminance, but not for analytics on them.
So yes a dedicated profile could make sense

PR #35084, opened 2020-03-14
[needs-docs][processing] Optionally, also remove EMPTY geometries in RemoveNullGeometries algorithm

gacarrillor:
## Description

Empty geometries do not contain coordinates. Thus, like `NULL` geometries, empty geometries are not spatial data in a strict sense. To make it easier for users to clean a vector layer, the `Remove null geometries` algorithm now has an `Also remove empty geometries` option.

The algorithm dialog looks like this:

image


src/analysis/processing/qgsalgorithmremovenullgeometry.cpp
```diff
@@ -60,7 +61,10 @@ QString QgsRemoveNullGeometryAlgorithm::shortHelpString() const
 {
   return QObject::tr( "This algorithm removes any features which do not have a geometry from a vector layer. "
                       "All other features will be copied unchanged.\n\n"
-                      "Optionally, the features with null geometries can be saved to a separate output." );
+                      "Optionally, the features with null geometries can be saved to a separate output.\n\n"
+                      "If 'Also remove empty geometries' is checked, the algorithm removes features whose geometries"
```
nyalldawson:
```suggestion
                      "If 'Also remove empty geometries' is checked, the algorithm removes features whose geometries "
```

src/analysis/processing/qgsalgorithmremovenullgeometry.cpp
```diff
@@ -74,6 +78,8 @@ QVariantMap QgsRemoveNullGeometryAlgorithm::processAlgorithm( const QVariantMap
   if ( !source )
     throw QgsProcessingException( invalidSourceError( parameters, QStringLiteral( "INPUT" ) ) );
 
+  const bool remove_empty = parameterAsBoolean( parameters, QStringLiteral( "REMOVE_EMPTY" ), context );
```
nyalldawson:
```suggestion
  const bool removeEmpty = parameterAsBoolean( parameters, QStringLiteral( "REMOVE_EMPTY" ), context );
```

nyalldawson (commented):
Nice addition!

src/analysis/processing/qgsalgorithmremovenullgeometry.cpp
gacarrillor: Applied in https://github.com/qgis/QGIS/pull/35084/commits/3b6c04f70e57c2ce3b6bb2bee93c3f643f7a88b6

src/analysis/processing/qgsalgorithmremovenullgeometry.cpp
gacarrillor: Applied in https://github.com/qgis/QGIS/pull/35084/commits/3b6c04f70e57c2ce3b6bb2bee93c3f643f7a88b6

PR #35085, opened 2020-03-15
Feature: Add toolbutton to "Deselect Features from the Current Active layer"

### LINKED ISSUE #31598: deselect features of the active layer
hello,

it would be useful, if i could deselect features of the active layer (with an new button in the select menu?). at the moment i can do this only if i open the attributetable and click the deselect button. this is not a short way and for large tables painful.

and additional it would be cool if the select menu (from edit menu) also available in the layer contextmenu (right mouseclick). this is little bit inspired from A**GIS ;-)

i hope someone can implement this !? :-)

thank you very much

suricactus:
Pretty much the title.
Peek 2020-03-15 01-55

CODE NOTE: Only the three selection tools were using the `addActions` way of adding actions, so I rewrote them to be the same as the rest of the function.
```diff
- bt->addActions( actionsList )
+ bt->addAction ( singleAction )
```
The function is still extremely ugly and needs some love one day.

Fixes: https://github.com/qgis/QGIS/issues/31598

### DIFF
--- a/images/images.qrc
+++ b/images/images.qrc
@@ -268,6 +268,7 @@
         <file>themes/default/mActionDeleteSelected.svg</file>
         <file>themes/default/mActionDeleteTable.svg</file>
         <file>themes/default/mActionDeselectAll.svg</file>
+        <file>themes/default/mActionDeselectActiveLayer.svg</file>
         <file>themes/default/mActionDuplicateLayer.svg</file>
         <file>themes/default/mActionDuplicateComposer.svg</file>
         <file>themes/default/mActionEditCopy.svg</file>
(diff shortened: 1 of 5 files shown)


suricactus:
I forgot to mention I moved the default `themes/default/mActionDeselectAll.svg` to `themes/default/mActionDeselectCurrentLayer.svg`, and created a new `themes/default/mActionDeselectAll.svg`, which resembles multiple layers. May need to change the icon at other places too.

j-lippmann:
@suricactus thank you for implementing this Feature! It is also available in the layercontextmenu?

suricactus:
I don't think it should be, as there is a dedicated toolbar button for that, as well in the menu "Edit->Select->Deselect Features from the Current Active layer".  For users who need that, they can always use a shortcut, much easier.

Which brings the discussion, as "Ctrl+A" selects all the features in the current layer, I believe "Ctrl+Shift+A" should do the opposite (deselect from the current layer only), not deselecting from all layers. This would be another breaking change, which the core developers should approve.

nyalldawson:
Looks good to me! @nirvn any UI thoughts?

nirvn:
I'm fine with it, I'm sure it'll please quite a few users out there :smile:

DelazJ:
In the attribute table, the icon used should be the deselectActiveLayer and not the deselectalllayers
image

pigreco:
Please add `GUI/UX` label

DelazJ:
@pigreco, `gui/ux`, `temporal`and others are used to build the changelog? Or what are the reasons of your requests?

pigreco:
> @pigreco, `gui/ux`, `temporal`and others are used to build the changelog? Or what are the reasons of your requests?

Exactly, to build and order changelogs faster, thanks

PR #35088, opened 2020-03-15
Move responsiblity for creation of file filter strings from Python to C++

nyalldawson:
Some cleanups standing in the way of more interesting work


alexbruy (approved):
Very useful addition!

PR #35090, opened 2020-03-15
Drop Processing PostgisTableSelector, use standard QgsNewDatabaseTableNameWidget instead

nyalldawson:
Swaps more adhoc Python code for the well tested, connections based c++ class.

If this is accepted alongside #35088 then I'll port the whole DestinationSelectionPanel to c++


alexbruy (approved):

nyalldawson:
@elpaso how's this for you?

elpaso (approved):

PR #35097, opened 2020-03-15
Make Add Group button act as Group Selected if selected layers >= 2

uclaros:
## Description
When there are more than one layers selected and the user presses the *Add Group* button, create a new group and put those layers in it too!
Peek 2020-03-15 22-51


nyalldawson:
I like this idea!! 

@nirvn what do you think?

nirvn:
Love it!

pigreco:
Please add `GUI7UX` label

PR #35101, opened 2020-03-16
Make provider key lookup case insensitive, add QgsConnectionRegistry

nyalldawson:
This PR does two things:

- Make provider key lookup case insensitive. We use a confusing mix of provider key casing (e.g. 'ogr', 'WFS', 'wms'), so make the lookups case insensitive to avoid frustration. (There's 0% chance of two different providers having the same key but different case, so should be no regressions here!)

- Adds QgsConnectionRegistry. Initially only a basic shell of a class, but has one method for retrieving a connection based on a single string of the format "provider://connection name" (e.g. "postgres://my connection"). We can further flesh out this class as required...


elpaso (approved):
Very handy!

agiudiceandrea:
Hi @nyalldawson, it seems to me this PR introduced some weird behaviour and inconsistency.
In fact, thanks to this PR, it is possible to create e.g. a Memory layer with `QgsVectorLayer(uri, name, "Memory")`, anyway such layer will not behave exactly as a Memory layer created with `QgsVectorLayer(uri, name, "memory")` because the provider key name is compared in a case sensitive way in a lot of part of the source code...

nyalldawson:
@agiudiceandrea 

> because the provider key name is compared in a case sensitive way in a lot of part of the source code...

I don't think this change is the issue here -- rather that the rest of the code should also be case insensitive. (Or we make the layer update the provider string to match the actual matched provider key after setting the data source)

agiudiceandrea:
In the rest of the code the check is usually made as `providerkey == QLatin1String( "memory" )`
I noticed the issue when creating a layer with "Memory" provider key and adding it to the project map; e.g. it doesn't display the memory layer icon in the layer tree and doesn't trigger the warning when the project is closed.

> we make the layer update the provider string to match the actual matched provider key after setting the data source

Thanks @nyalldawson. Yes, I also think this would fix the inconsistency!

PR #35102, opened 2020-03-16
Use core connections API in processing database algorithms instead of Python versions

nyalldawson:
Ports the Postgres and Spatialite algorithms to the core database connection API instead of Python versions of these.

The python classes are basically 1:1 duplicates of the core classes, without unit tests and without a common API.

Part of the ongoing Processing database handling cleanup work, but also adds nice API for creation of spatial indexes which can be later used as part of the db manager cleanup/porting efforts.


alexbruy (approved):

python/plugins/processing/algs/gdal/ogr2ogrtopostgislist.py
```diff
@@ -244,7 +255,7 @@ def getConsoleCommands(self, parameters, context, feedback, executing=True):
         arguments.append('-f')
         arguments.append('PostgreSQL')
         arguments.append('PG:"')
-        for token in uri.connectionInfo(executing).split(' '):
+        for token in QgsDataSourceUri(uri).connectionInfo(executing).split(' '):
```
elpaso: Shouldn't use `decodeUri` here?

elpaso (approved):
Maybe you can use `decodeUri` in some places instead of splitting on ` ` but I didn't check if it's really appropriate in this case.

python/plugins/processing/algs/gdal/ogr2ogrtopostgislist.py
nyalldawson: Probably, but I'm reluctant to change anything I don't have to here :smiley:

nirvn:
@nyalldawson , nice work.

PR #35103, opened 2020-03-16
Fix case of spatialite provider name

nyalldawson:
Breaks adding spatialite layers through the browser

@elpaso you recently changed this to uppercase (in 2e992c096a83b25b20ef607550361c39648557bc) - can you recall why?

### DIFF
--- a/src/gui/qgsnewdatabasetablenamewidget.cpp
+++ b/src/gui/qgsnewdatabasetablenamewidget.cpp
@@ -26,7 +26,7 @@
 
 
 // List of data item provider keys that are filesystem based
-QStringList QgsNewDatabaseTableNameWidget::FILESYSTEM_BASED_DATAITEM_PROVIDERS { QStringLiteral( "GPKG" ), QStringLiteral( "SPATIALITE" ) };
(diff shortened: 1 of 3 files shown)


elpaso:
> Breaks adding spatialite layers through the browser
> 
> @elpaso you recently changed this to uppercase (in [2e992c0](https://github.com/qgis/QGIS/commit/2e992c096a83b25b20ef607550361c39648557bc)) - can you recall why?

When I did a small refactoring and cleanup in the providers to return the `dataProviderName` after @PeterPetrik added the `providerName` (which is the **data items** provider name, **not the data provider name**), to propagate the name in the children  I took the `providerName`  and I copied in the constructors of the child items, as you can see the root item `providerName` returns upper case: https://github.com/qgis/QGIS/blob/release-3_12/src/providers/spatialite/qgsspatialitedataitems.cpp#L279

So, probably `QgsSpatiaLiteDataItemProvider::name()` was wrong in the first place and I propagated the error.

nyalldawson:
@elpaso so you're happy with this change?

elpaso:
Yes, but not knowing why  https://github.com/qgis/QGIS/blob/release-3_12/src/providers/spatialite/qgsspatialitedataitems.cpp#L279 was uppercase I'm not sure if we are missing something.

PR #35110, opened 2020-03-16
Snapping enabled on a configurable scale range

obrix:
## Description

This pull request follows the qgis enhancement proposal 167 described in details here : [Snapping enabled on a configurable scale range](https://github.com/qgis/QGIS-Enhancement-Proposals/issues/167)

The goal is to provide a way to optimizing the snapping feature by allowing the user to enable it only on specified scale range. A global setting and a per layer setting are available.

For more details please refer to the qgis enhancement proposal description and discussion.

### DIFF
--- a/src/app/qgssnappinglayertreemodel.cpp
+++ b/src/app/qgssnappinglayertreemodel.cpp
@@ -100,6 +100,31 @@ QWidget *QgsSnappingLayerDelegate::createEditor( QWidget *parent, const QStyleOp
     return w;
   }
 
+  if ( index.column() == QgsSnappingLayerTreeModel::LimitToScaleRangeColumn )
+  {
+    QCheckBox *cbUseLimit = new QCheckBox( parent );
+    cbUseLimit->setChecked( false );
+    return cbUseLimit;
+  }
+
+  if ( index.column() == QgsSnappingLayerTreeModel::MinScaleColumn )
+  {
+    QDoubleSpinBox *minLimitSp = new QDoubleSpinBox( parent );
+    minLimitSp->setDecimals( 5 );
+    minLimitSp->setMaximum( 99999999.990000 );
+    minLimitSp->setToolTip( tr( "Snapping scale range min" ) );
+    return minLimitSp;
+  }
+
+  if ( index.column() == QgsSnappingLayerTreeModel::MaxScaleColumn )
+  {
+    QDoubleSpinBox *maxLimitSp = new QDoubleSpinBox( parent );
+    maxLimitSp->setDecimals( 5 );
+    maxLimitSp->setMaximum( 99999999.990000 );
+    maxLimitSp->setToolTip( tr( "Snapping scale range max" ) );
+    return maxLimitSp;
+  }
+
   return nullptr;
 }
 
@@ -139,6 +164,23 @@ void QgsSnappingLayerDelegate::setEditorData( QWidget *editor, const QModelIndex
     if ( w )
     {
       w->setCurrentIndex( w->findData( units ) );
+    }
+  }
+  else if ( index.column() == QgsSnappingLayerTreeModel::LimitToScaleRangeColumn )
+  {
+    QCheckBox *w = qobject_cast<QCheckBox *>( editor );
+    if ( w )
+    {
+      w->setChecked( val.toBool() );
+    }
+  }
+  else if ( index.column() == QgsSnappingLayerTreeModel::MinScaleColumn ||
+            index.column() == QgsSnappingLayerTreeModel::MaxScaleColumn )
+  {
+    QDoubleSpinBox *w = qobject_cast<QDoubleSpinBox *>( editor );
+    if ( w )
+    {
+      w->setValue( val.toDouble() );
     }
   }
 }
@@ -183,6 +225,23 @@ void QgsSnappingLayerDelegate::setModelData( QWidget *editor, QAbstractItemModel
       model->setData( index, w->value(), Qt::EditRole );
     }
   }
+  else if ( index.column() == QgsSnappingLayerTreeModel::LimitToScaleRangeColumn )
+  {
+    QCheckBox *w = qobject_cast<QCheckBox *>( editor );
+    if ( w )
+    {
+      model->setData( index, w->isChecked(), Qt::EditRole );
+    }
+  }
+  else if ( index.column() == QgsSnappingLayerTreeModel::MinScaleColumn ||
+            index.column() == QgsSnappingLayerTreeModel::MaxScaleColumn )
+  {
+    QDoubleSpinBox *w = qobject_cast<QDoubleSpinBox *>( editor );
+    if ( w )
+    {
+      model->setData( index, w->value(), Qt::EditRole );
+    }
+  }
 }
 
 
@@ -200,7 +259,7 @@ QgsSnappingLayerTreeModel::QgsSnappingLayerTreeModel( QgsProject *project, QgsMa
 int QgsSnappingLayerTreeModel::columnCount( const QModelIndex &parent ) const
 {
   Q_UNUSED( parent )
-  return 5;
+  return 8;
 }
 
 Qt::ItemFlags QgsSnappingLayerTreeModel::flags( const QModelIndex &idx ) const
@@ -400,6 +459,12 @@ QVariant QgsSnappingLayerTreeModel::headerData( int section, Qt::Orientation ori
           return tr( "Units" );
         case 4:
           return tr( "Avoid overlap" );
+        case 5:
+          return tr( "Activate snapping on scale range" );
+        case 6:
+          return tr( "Range min" );
+        case 7:
+          return tr( "Range max" );
         default:
           return QVariant();
       }
@@ -579,6 +644,45 @@ QVariant QgsSnappingLayerTreeModel::data( const QModelIndex &idx, int role ) con
         }
       }
     }
+
+    if ( idx.column() == LimitToScaleRangeColumn )
+    {
+      if ( role == Qt::DisplayRole )
+      {
+        return  QString::number( ls.limitToScaleRange() );
+      }
+
+      if ( role == Qt::UserRole )
+      {
+        return ls.limitToScaleRange();
+      }
+    }
+
+    if ( idx.column() == MinScaleColumn )
+    {
+      if ( role == Qt::DisplayRole )
+      {
+        return QString::number( ls.minScale() );
+      }
+
+      if ( role == Qt::UserRole )
+      {
+        return ls.minScale();
+      }
+    }
+
+    if ( idx.column() == MaxScaleColumn )
+    {
+      if ( role == Qt::DisplayRole )
+      {
+        return QString::number( ls.maxScale() );
+      }
+
+      if ( role == Qt::UserRole )
+      {
+        return ls.maxScale();
+      }
+    }
   }
 
   return QVariant();
@@ -707,6 +811,69 @@ bool QgsSnappingLayerTreeModel::setData( const QModelIndex &index, const QVarian
         avoidIntersectionsList.removeAll( vl );
 
       mProject->setAvoidIntersectionsLayers( avoidIntersectionsList );
+      emit dataChanged( index, index );
+      return true;
+    }
+  }
+
+  if ( index.column() == LimitToScaleRangeColumn && role == Qt::EditRole )
+  {
+    QgsVectorLayer *vl = vectorLayer( index );
+    if ( vl )
+    {
+      if ( !mIndividualLayerSettings.contains( vl ) )
+        return false;
+
+      QgsSnappingConfig::IndividualLayerSettings ls = mIndividualLayerSettings.value( vl );
+      if ( !ls.valid() )
+        return false;
+
+      ls.setLimitToScaleRange( value.toBool() );
+      QgsSnappingConfig config = mProject->snappingConfig();
+      config.setIndividualLayerSettings( vl, ls );
+      mProject->setSnappingConfig( config );
+      emit dataChanged( index, index );
+      return true;
+    }
+  }
+
+  if ( index.column() == MinScaleColumn && role == Qt::EditRole )
+  {
+    QgsVectorLayer *vl = vectorLayer( index );
+    if ( vl )
+    {
+      if ( !mIndividualLayerSettings.contains( vl ) )
+        return false;
+
+      QgsSnappingConfig::IndividualLayerSettings ls = mIndividualLayerSettings.value( vl );
+      if ( !ls.valid() )
+        return false;
+
+      ls.setMinScale( value.toDouble() );
+      QgsSnappingConfig config = mProject->snappingConfig();
+      config.setIndividualLayerSettings( vl, ls );
+      mProject->setSnappingConfig( config );
+      emit dataChanged( index, index );
+      return true;
+    }
+  }
+
+  if ( index.column() == MaxScaleColumn && role == Qt::EditRole )
+  {
+    QgsVectorLayer *vl = vectorLayer( index );
+    if ( vl )
+    {
+      if ( !mIndividualLayerSettings.contains( vl ) )
+        return false;
+
+      QgsSnappingConfig::IndividualLayerSettings ls = mIndividualLayerSettings.value( vl );
+      if ( !ls.valid() )
+        return false;
+
+      ls.setMaxScale( value.toDouble() );
+      QgsSnappingConfig config = mProject->snappingConfig();
+      config.setIndividualLayerSettings( vl, ls );
+      mProject->setSnappingConfig( config );
       emit dataChanged( index, index );
       return true;
     }
--- a/src/app/qgssnappinglayertreemodel.cpp
+++ b/src/app/qgssnappinglayertreemodel.cpp
@@ -27,6 +27,7 @@
 #include "qgssnappingconfig.h"
 #include "qgsvectorlayer.h"
 #include "qgsapplication.h"
+#include "qgsscalewidget.h"
 
 QgsSnappingLayerDelegate::QgsSnappingLayerDelegate( QgsMapCanvas *canvas, QObject *parent )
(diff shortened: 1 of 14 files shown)


obrix:
@troopa81 @nyalldawson Could you review this PR ?

troopa81 (changes requested):
It's not clear to me how the scale limit apply:
- If global limit to scale is disabled, no scale range (global nor per layer) are applied
- Else we use global range only if per layer one hasn't been defined

As a user, I would have expected
- If global limit to scale is disable, per layer scale range are applied
- Else we check global scale range AND per layer scale range before snapping on this layer

But maybe, I'm a weird user :)

obrix:
@troopa81 Yes you got it the global boolean disable the feature globally (including the per layer configuration). The layer specific setting applies if specified, otherwise the global setting applies.

I think it is useful to have a simple way to completely disable the feature. Your expectation give no way to disable the optimization except resetting manually all settings which might be a pain I think.

src/core/qgssnappingutils.cpp
```diff
@@ -291,20 +291,40 @@ QgsPointLocator::Match QgsSnappingUtils::snapToMap( const QgsPointXY &pointMap,
   }
   else if ( mSnappingConfig.mode() == QgsSnappingConfig::AdvancedConfiguration )
   {
+    typedef QList<LayerConfig>::const_iterator LayerConfigIterator;
     QList<LayerAndAreaOfInterest> layers;
-    for ( const LayerConfig &layerConfig : qgis::as_const( mLayers ) )
+    QList<LayerConfigIterator> filteredConfigs;
```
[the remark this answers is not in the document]
obrix: yes

src/core/qgssnappingconfig.h
```diff
@@ -116,19 +116,23 @@ class CORE_EXPORT QgsSnappingConfig
          * \param type
          * \param tolerance
          * \param units
+         * \param minScale 0.0 disable scale limit
+         * \param maxScale 0.0 disable scale limit
          * \deprecated since QGIS 3.12 use the method with SnappingTypeFlag instead.
          */
-        Q_DECL_DEPRECATED IndividualLayerSettings( bool enabled, SnappingType type, double tolerance, QgsTolerance::UnitType units ) SIP_DEPRECATED;
+        Q_DECL_DEPRECATED IndividualLayerSettings( bool enabled, SnappingType type, double tolerance, QgsTolerance::UnitType units, double minScale, double maxScale ) SIP_DEPRECATED;
```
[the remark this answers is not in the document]
obrix: Probably not.

src/app/qgssnappinglayertreemodel.cpp
```diff
@@ -579,6 +630,46 @@ QVariant QgsSnappingLayerTreeModel::data( const QModelIndex &idx, int role ) con
...
+      }
+    }
+
+    if ( idx.column() == MaxScaleColumn )
+    {
+      if ( role == Qt::DisplayRole )
+      {
+        if ( ls.maxScale() <= 0.0 )
+        {
+          return QString( "not set" );
```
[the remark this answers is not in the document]
obrix: yes

src/app/qgssnappinglayertreemodel.cpp
```diff
@@ -579,6 +630,46 @@ QVariant QgsSnappingLayerTreeModel::data( const QModelIndex &idx, int role ) con
...
       }
     }
+
+    if ( idx.column() == MinScaleColumn )
+    {
+      if ( role == Qt::DisplayRole )
+      {
+        if ( ls.minScale() <= 0.0 )
+        {
+          return QString( "not set" );
```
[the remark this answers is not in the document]
obrix: yes

src/app/qgssnappingwidget.cpp
```diff
@@ -173,6 +175,23 @@ QgsSnappingWidget::QgsSnappingWidget( QgsProject *project, QgsMapCanvas *canvas,
   mToleranceSpinBox->setObjectName( QStringLiteral( "SnappingToleranceSpinBox" ) );
   connect( mToleranceSpinBox, static_cast < void ( QDoubleSpinBox::* )( double ) > ( &QDoubleSpinBox::valueChanged ), this, &QgsSnappingWidget::changeTolerance );
 
+  mMinScaleWidget = new QgsScaleWidget();
+  mMinScaleWidget->setToolTip( tr( "Min scale on which snapping is enabled" ) );
```
3nids:
```suggestion
  mMinScaleWidget->setToolTip( tr( "Minimum scale from which snapping is enabled" ) );
```
maybe some native speaker can come up with something better

src/app/qgssnappingwidget.cpp
```diff
@@ -173,6 +175,23 @@ QgsSnappingWidget::QgsSnappingWidget( QgsProject *project, QgsMapCanvas *canvas,
...
+  mMinScaleWidget->setObjectName( QStringLiteral( "SnappingMinScaleSpinBox" ) );
+  connect( mMinScaleWidget, &QgsScaleWidget::scaleChanged, this, &QgsSnappingWidget::changeMinScale );
+
+  mMaxScaleWidget = new QgsScaleWidget();
+  mMaxScaleWidget->setToolTip( tr( "Max scale on which snapping is enabled" ) );
```
3nids:
```suggestion
  mMaxScaleWidget->setToolTip( tr( "Maximum scale up to which snapping is enabled" ) );
```
...not so sure

src/app/qgssnappingwidget.cpp
```diff
@@ -173,6 +175,23 @@ QgsSnappingWidget::QgsSnappingWidget( QgsProject *project, QgsMapCanvas *canvas,
...
+  mMaxScaleWidget->setObjectName( QStringLiteral( "SnappingMaxScaleSpinBox" ) );
+  connect( mMaxScaleWidget, &QgsScaleWidget::scaleChanged, this, &QgsSnappingWidget::changeMaxScale );
+
+
+  mLimitToScale = new QAction( tr( "Toggle Snapping limit on scale" ), this );
```
3nids: I am not sure if it should be camel case or lower case but not a mix

3nids: "Toggle scale dependent snapping" ?

src/app/qgssnappingwidget.cpp
```diff
@@ -173,6 +175,23 @@ QgsSnappingWidget::QgsSnappingWidget( QgsProject *project, QgsMapCanvas *canvas,
...
+  mLimitToScale->setCheckable( true );
+  mLimitToScale->setIcon( QIcon( QgsApplication::getThemeIcon( "/mIconSnappingOnScale.svg" ) ) );
+  mLimitToScale->setObjectName( QStringLiteral( "EnableSnappinLimitOnScaleAction" ) );
```
3nids:
```suggestion
  mLimitToScale->setObjectName( QStringLiteral( "EnableSnappingLimitOnScaleAction" ) );
```

src/core/qgssnappingconfig.h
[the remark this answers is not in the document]
3nids: +1

src/core/qgssnappingconfig.h
```diff
@@ -180,6 +184,18 @@ class CORE_EXPORT QgsSnappingConfig
         //! Sets the type of units
         void setUnits( QgsTolerance::UnitType units );
 
+        //! Returns min scale on which snapping is limited
```
3nids: missing \since and other following new methods

src/core/qgssnappingutils.cpp
```diff
@@ -291,20 +291,40 @@ QgsPointLocator::Match QgsSnappingUtils::snapToMap( const QgsPointXY &pointMap,
...
+
+    bool inRangeGlobal = ( mSnappingConfig.minScale() <= 0.0 || mMapSettings.scale() >= mSnappingConfig.minScale() )
+                         && ( mSnappingConfig.maxScale() <= 0.0 || mMapSettings.scale() <= mSnappingConfig.maxScale() );
+
+    for ( LayerConfigIterator it = mLayers.begin(); it != mLayers.end(); ++it )
```
3nids:
```suggestion
    for ( const LayerConfig &layerConfig : qgis::as_const( mLayers ) )
```

src/core/qgssnappingutils.cpp
```diff
@@ -291,20 +291,40 @@ QgsPointLocator::Match QgsSnappingUtils::snapToMap( const QgsPointXY &pointMap,
...
+      }
     }
     prepareIndex( layers, relaxed );
 
     QgsPointLocator::Match bestMatch;
     QgsPointLocator::MatchList edges; // for snap on intersection
     double maxSnapIntTolerance = 0;
 
-    for ( const LayerConfig &layerConfig : qgis::as_const( mLayers ) )
+    for ( LayerConfigIterator &it : filteredConfigs )
```
3nids:
```suggestion
    for ( const LayerConfig &layerConfig : qgis::as_const( filteredConfigs ) )
```

tests/src/core/testqgssnappingutils.cpp
```diff
@@ -33,7 +33,7 @@ struct FilterExcludePoint : public QgsPointLocator::MatchFilter
 {
   explicit FilterExcludePoint( const QgsPointXY &p ) : mPoint( p ) {}
 
-  bool acceptMatch( const QgsPointLocator::Match &match ) override { return match.point() != mPoint; }
+  bool acceptMatch( const QgsPointLocator::Match &match ) override { return match.point() != mPoint;  }
```
3nids:
```suggestion
  bool acceptMatch( const QgsPointLocator::Match &match ) override { return match.point() != mPoint; }
```

troopa81:
> I think it is useful to have a simple way to completely disable the feature. 

I understand your point, so OK for a button which disable all limit scale range. Maybe it would be good to disable (grayed) per layer min/max configuration when the button is disable, as it is done for the global min/max. This way, the user understand it applies to both global and per layer configuration.

3nids (commented):
* could you provide a small screencast of the config widget + toolbar
* would it make sense to add some tests to this?
* small thing but rather than talking about "Limit on scale snapping" I would rather use "Scale dependent snapping". (to be approved by better english speakers)

3nids:
> Yes you got it the global boolean disable the feature globally (including the per layer configuration). The layer specific setting applies if specified, otherwise the global setting applies.

good idea!

troopa81:
@obrix As soon as you define a per layer min, both per layer min/max applies even if you define a global max and your per layer max is *not set*. I find it a little bit confusing. Wouldn't be better to separate min et max ? Or at least, make the current behavior more explicit (tooltip?)

obrix:
@troopa81 

> I understand your point, so OK for a button which disable all limit scale range. Maybe it would be good to disable (grayed) per layer min/max configuration when the button is disable, as it is done for the global min/max. This way, the user understand it applies to both global and per layer configuration.

Yes good idea to also explicitly grey out the per layer configuration when the feature is disabled.

> As soon as you define a per layer min, both per layer min/max applies even if you define a global max and your per layer max is not set. I find it a little bit confusing. Wouldn't be better to separate min et max ? Or at least, make the current behavior more explicit (tooltip?)

Yes this is a deliberate choice, from my point of view mixing some configuration from the global settings and some configuration from the per layer settings was confusing. I wanted to avoid that but I am not really satisfied by the way it is presented to the user though. If this is confusing for you I should probably separate min/max and use global setting when one is not set.

@3nids 

> would it make sense to add some tests to this?
small thing but rather than talking about "Limit on scale snapping" I would rather use "Scale dependent snapping". (to be approved by better english speakers)

Not sure about a test. I can take a look. I Agree with "Scale dependent snapping".

troopa81:
> Not sure about a test. I can take a look.

+1 for a test. at least in testqgssnappingutils.cpp to check the
different snapping behavior (with or without global min/max, with or without per
layer configuration...)

obrix:
@troopa81 If I handle separately min and max range for layer specific settings (ie if not set -> use global configuration), it makes it impossible for the user to specify a global setting and disable completely snapping optimization on a specific layer. This is also the case for the current implementation (ie if I unset both min and max, it will be replaced by global settings).

A solution I can think of is to re add a check box to specifically enable/disable the custom setting on a specific layer. I think it makes more sense and is more explicit for the user.

So we would have 
* Global checkbox button to enable/disable the feature globally, with global min max range applied to all layer
* A checkbox per layer to enable/disable a specific layer setting with min and max (with tooltip "Use layer specific snapping range")

I think this way the user will not have to guess which min max is used, it is explicit.

troopa81:
@obrix I'm concerned that it will make it even more complicated. I'd prefer to keep it simple
- a default (global) config min/max
- a per layer config min/max
- a button to disable scale dependent snapping

if both global and per layer min (resp. max) are defined (different from "not set"), we check only the per layer one.

troopa81:
> So we would have

>    Global checkbox button to enable/disable the feature globally, with global min max range applied to all layer
>    A checkbox per layer to enable/disable a specific layer setting with min and max (with tooltip "Use layer specific snapping range")

> I think this way the user will not have to guess which min max is used, it is explicit.

After a second thoughts, it may be in fact better to do it this way so it removes any ambiguity. @3nids do you agree?

3nids:
A maybe simpler alternative would be a tri-state for the main one:
* inactive
* active - global
* active - advanced / per layer

So, if we are in active global we can hide the columns in the table.

I think this feature is very very specific and even more when going per layer. So we should try not to complexify too much the UI.
I think the approach proposed above has both the advantage of the lightest possible UX and the clarity (at a small cost of not being able to fine the global setting per layer, but it sounds acceptable to me)

obrix:
The tri-state approach is okay for me as it also remove any ambiguity.

3nids:
ready to re-review / merge?

src/core/qgssnappingconfig.h
```diff
@@ -126,9 +138,11 @@ class CORE_EXPORT QgsSnappingConfig
          * \param type
          * \param tolerance
          * \param units
+         * \param minScale 0.0 disable scale limit
+         * \param maxScale 0.0 disable scale limit
          * \since QGIS 3.12
          */
-        IndividualLayerSettings( bool enabled, SnappingTypeFlag type, double tolerance, QgsTolerance::UnitType units );
+        IndividualLayerSettings( bool enabled, SnappingTypeFlag type, double tolerance, QgsTolerance::UnitType units, double minScale, double maxScale );
```
3nids, line 145: this is an API break, can you use default values?

src/core/qgssnappingconfig.h
```diff
@@ -168,18 +182,55 @@ class CORE_EXPORT QgsSnappingConfig
          */
         void setTypeFlag( QgsSnappingConfig::SnappingTypeFlag type );
 
-        //! Returns the tolerance
+        /**
+         * !
+         * Returns the tolerance
+         * \since QGIS 3.12
```
3nids, line 188: I don't think this method and the following were added in 3.12
https://github.com/qgis/QGIS/blame/master/src/core/qgssnappingconfig.h#L172

src/core/qgssnappingconfig.h
```diff
@@ -168,18 +182,55 @@ class CORE_EXPORT QgsSnappingConfig
...
+         * Sets the type of units
+         * \since QGIS 3.12
+         */
         void setUnits( QgsTolerance::UnitType units );
 
+        /**
+         * Returns min scale on which snapping is limited
+         * \since QGIS 3.14
+         */
+        double minScale() const;
```
3nids, line 214: both in the docstring and the method names can you use minimum and maximum in whole words rather than min/max?

src/core/qgssnappingconfig.h
obrix: Done

src/core/qgssnappingconfig.h
obrix: Yes I think I went too fast with some doc modification.

src/core/qgssnappingconfig.h
obrix: Done

src/app/qgssnappingwidget.h
```diff
@@ -100,6 +102,10 @@ class APP_EXPORT QgsSnappingWidget : public QWidget
 
     void changeTolerance( double tolerance );
 
+    void changeMinScale( double pMinScale );
```
3nids, line 105:
```suggestion
    void setMinimumScale( double minimumScale );
```
arguments are also part of the API due to Python, let's give them nice names
same for `changeMaxScale`

3nids: no biggie, I see now it's in app.
But better be safe before someone moves this to GUI ;)

src/app/qgssnappingwidget.h
```diff
@@ -123,7 +130,7 @@ class APP_EXPORT QgsSnappingWidget : public QWidget
     DisplayMode mDisplayMode;
 
     //! modeChanged determines if widget are visible or not based on mode
-    void modeChanged();
+    void modeChanged( );
```
3nids, line 133:
```suggestion
    void modeChanged();
```

3nids (approved):

src/app/qgssnappingwidget.h
obrix: done

src/app/qgssnappingwidget.h
obrix: done

src/app/qgssnappingwidget.cpp
3nids: if no ones comes up with something better, I would recommend to accept this one

src/app/qgssnappingwidget.cpp
3nids: +1

src/app/qgssnappingwidget.cpp
obrix: Yes, I went with it.

troopa81 (approved):

PR #35111, opened 2020-03-16
Avoid crash when opening raster layer properties for wms layers.

Samweli:
This PR adds checks for histogram and pyramid pages pointers when setting the "helpPage" property, this is to prevent a crash when opening the raster layer properties for wms based raster layers.


nyalldawson (approved):

timlinux:
@nyalldawson Out of curiosity, why are the PR's getting closed and re-opened again? I was hoping to get this into  tonight's build so our users can test. Thanks!

nyalldawson:
@timlinux it retriggers a new ci run with the pr rebased off master. We do it when a unrelated test fails or something else broke a ci run

timlinux:
@nyalldawson ah goddit thanks. And thanks for merging.

PR #35112, opened 2020-03-16
Determine primary key from Oracle view

troopa81:
## Description

From an Oracle database, you can specify a disabled primary key constraint so tools can easily determine primary key. 

> View Constraints Oracle Database does not enforce view constraints. However, you can enforce constraints on views through constraints on base tables.
> 
> You can specify only unique, primary key, and foreign key constraints on views, and they are supported only in DISABLE NOVALIDATE mode. You cannot define view constraints on attributes of an object column.

https://docs.oracle.com/cd/B19306_01/server.102/b14200/clauses002.htm

This PR use this behavior to determine primary key on view.

### DIFF
--- a/src/providers/oracle/qgsoracleprovider.cpp
+++ b/src/providers/oracle/qgsoracleprovider.cpp
@@ -895,8 +895,8 @@ bool QgsOracleProvider::determinePrimaryKey()
   if ( !mIsQuery )
   {
     if ( !exec( qry, QString( "SELECT column_name"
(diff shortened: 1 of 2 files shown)


troopa81:
Don't stale

troopa81:
Dont' stale

elpaso (approved):

troopa81:
Thanks @elpaso Do you mind merging it? I don't have the rights to do it

PR #35114, opened 2020-03-16
Copy custom properties to the offline project (offline editing)

signedav:
And having `QgsMapLayer::customProperties` as a get-method to receive all the properties of the layer.

### DIFF
--- a/src/core/qgsmaplayer.h
+++ b/src/core/qgsmaplayer.h
@@ -615,6 +615,13 @@ class CORE_EXPORT QgsMapLayer : public QObject
     void setCustomProperties( const QgsObjectCustomProperties &properties );
 
     /**
+     * Read all custom properties from layer. Properties are stored in a map and saved in project file.
+     * \see setCustomProperties
+     * \since QGIS 3.14
+     */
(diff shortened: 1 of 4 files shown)


src/core/qgsmaplayer.h
```diff
@@ -614,6 +614,12 @@ class CORE_EXPORT QgsMapLayer : public QObject
      */
     void setCustomProperties( const QgsObjectCustomProperties &properties );
 
+    /**
+     * Read all custom properties from layer. Properties are stored in a map and saved in project file.
+     * \see setCustomProperties
```
3nids: It misses since QGIS 3.14

3nids (approved):

src/core/qgsmaplayer.h
```diff
@@ -614,6 +614,12 @@ class CORE_EXPORT QgsMapLayer : public QObject
...
+     */
+    QgsObjectCustomProperties customProperties();
```
nyalldawson:
```suggestion
    const QgsObjectCustomProperties& customProperties() const;
```

Issue #35113, opened 2020-03-16
project_crs_name variable


Saijin-Naib:
**Feature description.**
A variable to expose the "name" or text of the project's CRS instead of just the authority ID and definition string would be quite valuable, as not everyone recognizes CRSs by their EPSG ID, and having "friendly" names visible, on a map layout for instance, is quite helpful.

**Additional context**
GIS.SE Question:  garth-cooper
https://gis.stackexchange.com/questions/354053/getting-text-name-of-current-project-crs-in-qgis-layout

> 
> I'm want to display the current project projection in a QGIS layout.
> 
> Right now I can get QGIS to display the current EPSG number like below:
> 
> [![Current Solution][1]][1]
> 
> But what I want to do is get QGIS layout to print the ESPG name as well like below with an expression:
> 
>  [![What I want][2]][2]
> 
>   [1]: https://i.stack.imgur.com/AgikV.png
>   [2]: https://i.stack.imgur.com/cR6Cm.png

Solution:  ian-turnton

> I think you will need a custom function to do this, the following seems to work though I haven't tested it extensively.
> 
```
>     from qgis.core import *
>     from qgis.gui import *
>     from qgis.utils import iface
>     
>     @qgsfunction(args='auto', group='Custom')
>     def getCrsDescription(feature, parent):
>         """
>         Returns the description text of the project CRS
>         <h2>Example usage:</h2>
>         <ul>
>           <li>getCrsDescription() -> WGS 84</li>
>           <li>getCrsDescription() -> OSGB 1936 / British National Grid</li>
>         </ul>
>         """
>         return iface.mapCanvas().mapSettings().destinationCrs().description()
```

roya0045:
Agreed.

roya0045:
@Saijin-Naib what about map_crs_description, this seems to be the one you want.

Saijin-Naib:
> @Saijin-Naib what about map_crs_description, this seems to be the one you want.

Does not seem to be available in 3.12.0. Is that master only?

roya0045:
Should be, if other map crs are there, I forgot I added that. https://github.com/qgis/QGIS/blob/f514516d0771d2163b0f42801dc3eb23c88e20c4/tests/src/core/testqgslayoutmap.cpp#L609

Maybe that's because its a map thing and the layout textbox is not linked to a specific map, hence why that context was not added for that.

Saijin-Naib:
Thanks so much, guys!

opfcl:
[...]

PR #35118, opened 2020-03-16
Add more project crs expressions

roya0045:
## Description

This bring crs variables that were present in layout item map to the project.

Motivated by: #35113 

Unsure if tests are warranted as those already exists for the layout item map.

### DIFF
--- a/src/core/expression/qgsexpression.cpp
+++ b/src/core/expression/qgsexpression.cpp
@@ -700,12 +700,12 @@ void QgsExpression::initVariableHelp()
   sVariableHelpTexts()->insert( QStringLiteral( "project_home" ), QCoreApplication::translate( "variable_help", "Home path of current project." ) );
   sVariableHelpTexts()->insert( QStringLiteral( "project_crs" ), QCoreApplication::translate( "variable_help", "Coordinate reference system of project (e.g., 'EPSG:4326')." ) );
   sVariableHelpTexts()->insert( QStringLiteral( "project_crs_definition" ), QCoreApplication::translate( "variable_help", "Coordinate reference system of project (full definition)." ) );
-  sVariableHelpTexts()->insert( QStringLiteral( "project_units" ), QCoreApplication::translate( "variable_help", "Unit of the projet's CRS." ) );
+  sVariableHelpTexts()->insert( QStringLiteral( "project_units" ), QCoreApplication::translate( "variable_help", "Unit of the project's CRS." ) );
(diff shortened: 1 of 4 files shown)


nyalldawson:
> Unsure if tests are warranted as those already exists for the layout item map.

Yeah, we'll still need them -- sorry :laughing: 

LGTM though!

tests/src/core/testqgsproject.cpp
```diff
@@ -597,6 +597,50 @@ void TestQgsProject::testSetGetCrs()
   ellipsoidChangedSpy.clear();
 }
 
+void
```
nyalldawson: I don't think tests run like this -- I think they need to be defined as private slots in order for the test executable to run them

nyalldawson: ACtually this test definitely isn't running - because it's testing @map_crs, etc, when it should be testing project_crs,...!

roya0045: Sorry too many windows opened and ended up commit this instead of something else.

nyalldawson:
Looks good, let's see what Travis says...

nyalldawson:
Travis says no. Did you build locally to test before pushing?

roya0045:
> Travis says no. Did you build locally to test before pushing?

I usually don't run the tests and simply do a manual test. I ended up implementing the tests in github web.

roya0045:
I'm getting an unrelated sip error, not sure what got merged with an imperfect sip binding but the tests are ok.

src/core/expression/qgsexpression.cpp
```diff
@@ -700,6 +700,12 @@ void QgsExpression::initVariableHelp()
   sVariableHelpTexts()->insert( QStringLiteral( "project_home" ), QCoreApplication::translate( "variable_help", "Home path of current project." ) );
   sVariableHelpTexts()->insert( QStringLiteral( "project_crs" ), QCoreApplication::translate( "variable_help", "Coordinate reference system of project (e.g., 'EPSG:4326')." ) );
   sVariableHelpTexts()->insert( QStringLiteral( "project_crs_definition" ), QCoreApplication::translate( "variable_help", "Coordinate reference system of project (full definition)." ) );
+  sVariableHelpTexts()->insert( QStringLiteral( "project_units" ), QCoreApplication::translate( "variable_help", "Unit of the projet's CRS." ) );
```
nyalldawson:
```suggestion
  sVariableHelpTexts()->insert( QStringLiteral( "project_units" ), QCoreApplication::translate( "variable_help", "Unit of the project's CRS." ) );
```

(also elsewhere here)

roya0045: should be fixed

nyalldawson (approved):

PR #35120, opened 2020-03-16
Don't override all cursors in bad layer handler dialog

nyalldawson:
Avoids the arrow cursor showing everywhere in this dialog, which is odd UX


roya0045:
Good catch!

PR #35130, opened 2020-03-17
Add "filter by geometry type" and "filter by layer type" algorithms to processing

nyalldawson:
This PR adds two new algorithms to processing:

- "Filter by geometry type": This algorithm filters features by their geometry type. Incoming features will be directed to different outputs based on whether they have a point, line or polygon geometry. It allows for model creation which responds to different input layer geometry types by applying different logic depending on the input geometry type.

- "Filter by layer type": This algorithm allows conditional model branching based on an input
layer type. For instance, it allows a model to adapt to the actual layer type of a generic "map layer" parameter input, and decide which branch of the model to run as a result.

It also adds in the required api to allow algorithms to "prune" model branches based on their calculated results. E.g. a model which returns the new FlagPruneModelBranchesBasedOnAlgorithmResults flag will cause any remaining parts of the model which are dependent on the outputs of that algorithm to be entirely skipped IF the algorithm does not return that particular output. (This is a prerequisite component for a future generic "conditional branching by expression" algorithm, and also used by "filter by layer type" to control the model flow based on the input layer type)

### DIFF
--- a/python/core/auto_generated/processing/qgsprocessingalgorithm.sip.in
+++ b/python/core/auto_generated/processing/qgsprocessingalgorithm.sip.in
@@ -47,6 +47,7 @@ Abstract base class for processing algorithms.
       FlagSupportsInPlaceEdits,
       FlagKnownIssues,
       FlagCustomException,
+      FlagPruneModelBranchesBasedOnAlgorithmResults,
       FlagDeprecated,
     };
     typedef QFlags<QgsProcessingAlgorithm::Flag> Flags;
--- a/python/plugins/processing/tests/testdata/expected/filter_geom_lines_line.gml
+++ b/python/plugins/processing/tests/testdata/expected/filter_geom_lines_line.gml
@@ -1,0 +1,44 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<ogr:FeatureCollection
+     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+     xsi:schemaLocation="http://ogr.maptools.org/ filter_geom_lines_line.xsd"
+     xmlns:ogr="http://ogr.maptools.org/"
+     xmlns:gml="http://www.opengis.net/gml">
+  <gml:boundedBy>
+    <gml:Box>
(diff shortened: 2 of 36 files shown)


alexbruy (approved):
Looking good. I'm only wondering if we should also handle mesh layers in the "Filter by layer type" algorithm.

nyalldawson:
>  I'm only wondering if we should also handle mesh layers in the "Filter by layer type" algorithm.

That was the original plan, but:
1. Nothing in core uses mesh layer parameter types for me to test with
2. We don't have a mesh layer output type, or any algorithms which work with them :laughing: 

Based on this I couldn't see the value in it...

saberraz:
> > I'm only wondering if we should also handle mesh layers in the "Filter by layer type" algorithm.
> 
> That was the original plan, but:
> 
> 1. Nothing in core uses mesh layer parameter types for me to test with
> 2. We don't have a mesh layer output type, or any algorithms which work with them laughing
> 
> Based on this I couldn't see the value in it...

Currently, mesh output format is not decided and depends on the mesh type. Mesh calculator is a presented as processing algorithm in Crayfish plugin...but as you said, it is note core yet.

nyalldawson:
@saberraz 
> Currently, mesh output format is not decided and depends on the mesh type. Mesh calculator is a presented as processing algorithm in Crayfish plugin...but as you said, it is note core yet.

Let me know when you want to start moving things across to QGIS -- I can help/give advice on the porting of the crayfish parameters

pigreco:
the two new algorithms are not present in the updated master today (OSGeo4W64)

roya0045:
@nyalldawson @DelazJ I noticed these algoritms are present under Layer Tools and Vector Selection, though I can understand the logic, wouldn't it make more sense to have them in Modeler Tools as they are specific to the modeler and since they are not in the processing widget, this seem confusing to have other group have more alg in the modeler.

nyalldawson:
@roya0045 I've no strong opinion either way. Feel free to send in a pr changing this!

PR #35138, opened 2020-03-17
[BUG] Fix QgsColorRampShaderWidget

### LINKED ISSUE #33882: Mesh contours color ramp sometimes not populated
1. load https://github.com/lutraconsulting/MDAL/blob/master/tests/data/ugrid/D-Flow1.2/bw_11_zonder_riviergrid_met_1dwtg_clm.nc as mesh layer. Nothing displays because Bed Evelation has all NaNs (no Z values in raw data)
2. click on "flow element center bedlevel (bl)" dataset. .. still nothing (BUG)
3. click on contrours tab, looks like color ramp was not populated (this is populated for other files). when you manually change min or max, the color ramp is created and data is drawn on canvas

### LINKED ISSUE #34962: [Mesh] Problem with styling when changing quantity in mesh tree
QGIS should remember the style when switching between quantities.

vcloarec:
This PR fixes issues related to the `QgsColorRampShaderWidget`:
fix #33882  
fix #34962

It also fixes the 'inf' issue with discrete interpolation : the color ramp tree was sometime populated with inf values.

Raster and mesh layer are concerned by this PR

### DIFF
--- a/src/gui/raster/qgscolorrampshaderwidget.cpp
+++ b/src/gui/raster/qgscolorrampshaderwidget.cpp
@@ -94,6 +94,7 @@ QgsColorRampShaderWidget::QgsColorRampShaderWidget( QWidget *parent )
   resetClassifyButton();
 
   connect( mClassificationModeComboBox, static_cast<void ( QComboBox::* )( int )>( &QComboBox::currentIndexChanged ), this, &QgsColorRampShaderWidget::classify );
+  connect( mColorInterpolationComboBox, static_cast<void ( QComboBox::* )( int )>( &QComboBox::currentIndexChanged ), this, &QgsColorRampShaderWidget::classify );
   connect( mClassifyButton, &QPushButton::clicked, this, &QgsColorRampShaderWidget::classify );
   connect( btnColorRamp, &QgsColorRampButton::colorRampChanged, this, &QgsColorRampShaderWidget::applyColorRamp );
   connect( mNumberOfEntriesSpinBox, static_cast < void ( QSpinBox::* )( int ) > ( &QSpinBox::valueChanged ), this, &QgsColorRampShaderWidget::classify );
@@ -265,7 +266,6 @@ void QgsColorRampShaderWidget::setUnitFromLabels()
   }
 }
 
-
 void QgsColorRampShaderWidget::mAddEntryButton_clicked()
 {
   QgsTreeWidgetItemObject *newItem = new QgsTreeWidgetItemObject( mColormapTreeWidget );
@@ -337,6 +337,7 @@ void QgsColorRampShaderWidget::classify()
     connect( newItem, &QgsTreeWidgetItemObject::itemEdited,
              this, &QgsColorRampShaderWidget::mColormapTreeWidget_itemEdited );
   }
+
   mClipCheckBox->setChecked( colorRampShader->clip() );
 
   autoLabel();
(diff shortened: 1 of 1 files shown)


saberraz:
Tested the PR and now it works as expected :ok_hand:

src/gui/raster/qgscolorrampshaderwidget.cpp
```diff
@@ -388,7 +388,7 @@ void QgsColorRampShaderWidget::applyColorRamp()
 
       double value = currentItem->text( ValueColumn ).toDouble();
       double position = ( value - mMin ) / ( mMax - mMin );
-      currentItem->setData( ColorColumn, Qt::EditRole, ramp->color( position ) );
+      whileBlocking( static_cast<QgsTreeWidgetItemObject *>( currentItem ) )->setData( ColorColumn, Qt::EditRole, ramp->color( position ) );
```
PeterPetrik: is `static_cast<QgsTreeWidgetItemObject *>` really needed here?

src/gui/raster/qgscolorrampshaderwidget.cpp
```diff
@@ -645,9 +644,20 @@ void QgsColorRampShaderWidget::mColormapTreeWidget_itemEdited( QTreeWidgetItem *
 
 void QgsColorRampShaderWidget::setFromShader( const QgsColorRampShader &colorRampShader )
 {
+  populateColormapTreeWidget( colorRampShader.colorRampItemList() );
+
+  // Those objects are connected to classify() the color ramp shader if they change, or call widget change
+  // need to block them to avoid to classify and to alter the color ramp, or to call duplicate widget change
+  whileBlocking( mClipCheckBox )->setChecked( colorRampShader.clip() );
+  whileBlocking( mColorInterpolationComboBox )->setCurrentIndex( mColorInterpolationComboBox->findData( colorRampShader.colorRampType() ) );
+  mColorInterpolationComboBox_currentIndexChanged( mColorInterpolationComboBox->currentIndex() );
```
PeterPetrik: why you block  the signal currentIndexChanged on line above and immediately emits it? isn't is the same as removal of `whileBlocking` line above?

src/gui/raster/qgscolorrampshaderwidget.cpp
```diff
@@ -645,9 +644,20 @@ void QgsColorRampShaderWidget::mColormapTreeWidget_itemEdited( QTreeWidgetItem *
...
+  whileBlocking( mClassificationModeComboBox )->setCurrentIndex( mClassificationModeComboBox->findData( colorRampShader.classificationMode() ) );
+  mClassificationModeComboBox_currentIndexChanged( mClassificationModeComboBox->currentIndex() );
```
PeterPetrik: why you block  the signal currentIndexChanged on line above and immediately emits it? isn't is the same as removal of `whileBlocking` line above?

src/gui/raster/qgscolorrampshaderwidget.cpp
vcloarec: yes, because `currentItem` is a `QTreeWidgetItem` that is not a `QObject` and do not provide signal/slots. Derived class `QgsTreeWidgetItemObject` is derived too from `QObject`.

src/gui/raster/qgscolorrampshaderwidget.cpp
vcloarec: Blocking to avoid to call the slots `QgsColorRampShaderWidget::classify` that is also connected.

src/gui/raster/qgscolorrampshaderwidget.cpp
vcloarec: Blocking to avoid to call the slots `QgsColorRampShaderWidget::classify` that is also connected.

PeterPetrik (approved):

PR #35142, opened 2020-03-17
QgsFieldMappingWidget & model

elpaso:
A new `QgsFieldMappingModel` class to be used with `QgsFieldMappingWidget` in the refactoring field processing alg and in other contexts.

The widget & model are designed to work with the following inputs:
- one set of destination `QgsFields`
- one set of source `QgsFields` that are made available in the expression context
- an optional list of expressions to be used as default values for the field mapping

Method to reset the model, to append and remove fields, to move selected rows of fields up and down are also available.

The widget output its current status in a list of `Field` records that contain:
- the (possibly edited) destination `QgsField`
- the original destination field name (because it might have been renamed)
- the expression to be used to generate values for the destination field

Note that a vector layer instance is not required.

# Widget is the inner table view

immagine


src/gui/qgsfieldmappingmodel.cpp
```diff
@@ -0,0 +1,356 @@
...
+  return Qt::ItemFlags();
+}
+
+bool QgsFieldMappingModel::setData( const QModelIndex &index, const QVariant &value, int role )
+{
+  if ( index.isValid() )
+  {
+    if ( role == Qt::EditRole )
+    {
+      Field &f { const_cast<Field &>( mMapping.at( index.row() ) ) };
```
m-kuhn:
```suggestion
      Field &f = mMapping[index.row()];
```

Just for ease of readability :)

m-kuhn (commented):
I like this a lot. The field mapper is such a powerful tool but still hard to use and having these fields available in the expression builder is an enormous help!

One thing, I just noticed this PR after merging https://github.com/qgis/QGIS/pull/31387 which also aimed to make refactor fields easier to use. Do you see it feasible to integrate this in here too?

src/gui/qgsfieldmappingmodel.cpp
elpaso: > One thing, I just noticed this PR after merging #31387 which also aimed to make refactor fields easier to use. Do you see it feasible to integrate this in here too?

Yes, it shouldn't be hard.

src/gui/qgsfieldmappingmodel.h
```diff
@@ -0,0 +1,157 @@
...
+{
+
+    Q_OBJECT
+
+  public:
+
+    /**
+     * The ColumnDataIndex enum represents the column index for the view
+     */
+    enum ColumnDataIndex
```
3nids:
```suggestion
    enum class ColumnDataIndex : int
```

src/gui/qgsfieldmappingmodel.h
```diff
@@ -0,0 +1,157 @@
...
+    {
+      SourceExpression,     //!< Expression
+      DestinationName,      //!< Destination field name
+      DestinationType,      //!< Destination field QVariant::Type casted to (int)
+      DestinationLength,    //!< Destination field length
+      DestinationPrecision  //!< Destination field precision
+    };
+
```
3nids:
```suggestion
    Q_ENUM( ColumnDataIndex );
```

3nids:
@elpaso awesome! can you provide a screenshot of the mapping widget?

elpaso:
> @elpaso awesome! can you provide a screenshot of the mapping widget?

it's exactly the same than the current FieldsMappingPanel.py internal table view, this is basically a C++ port of that, plus some design changes in the API to make id more adaptable to use outside of the processing context.

elpaso:
@3nids added anyway, but as I said it's the same UI as before (at least that was my intention).

src/gui/qgsfieldexpressionwidget.h
```diff
@@ -244,7 +250,7 @@ class GUI_EXPORT QgsFieldExpressionWidget : public QWidget
     QToolButton *mButton = nullptr;
     QgsFieldProxyModel *mFieldProxyModel = nullptr;
     QString mExpressionDialogTitle;
-    std::shared_ptr<const QgsDistanceArea> mDa;
+    std::shared_ptr<const QgsDistanceArea> mDistanceArea;
```
nyalldawson: I realise you didn't write this -- but there's no reason for this to be a shared ptr. A unique_ptr would better express the situation here.

src/gui/qgsfieldmappingmodel.cpp
```diff
@@ -0,0 +1,434 @@
...
+  , mExpressionContextGenerator( new ExpressionContextGenerator( &mSourceFields ) )
+{
+  setDestinationFields( destinationFields, expressions );
+}
+
+QVariant QgsFieldMappingModel::headerData( int section, Qt::Orientation orientation, int role ) const
+{
+  if ( role == Qt::DisplayRole )
+  {
+    if ( orientation == Qt::Horizontal )
```
nyalldawson: Minor nitpick, but I think a switch statement is more appropriate here

src/gui/qgsfieldmappingmodel.cpp
```diff
@@ -0,0 +1,434 @@
...
+QVariant QgsFieldMappingModel::data( const QModelIndex &index, int role ) const
+{
+  if ( index.isValid() )
+  {
+    const int col { index.column() };
+    const Field &f { mMapping.at( index.row() ) };
+
+    const QgsFieldConstraints::Constraints constraints { fieldConstraints( f.field ) };
+
+    if ( role == Qt::DisplayRole || role == Qt::EditRole )
```
nyalldawson: again, a switch is more appropriate here

src/gui/qgsfieldmappingmodel.cpp
```diff
@@ -0,0 +1,434 @@
...
+    {
+      switch ( col )
+      {
+          {
```
nyalldawson: this looks extraneous? (surprised it doesn't break the build..!)

src/gui/qgsfieldmappingmodel.cpp
```diff
@@ -0,0 +1,434 @@
...
+          case static_cast<int>( ColumnDataIndex::SourceExpression ):
```
nyalldawson: are these static_casts actually needed? Might be nicer to static_cast `col` rather then these values

src/gui/qgsfieldmappingmodel.h
```diff
@@ -0,0 +1,165 @@
...
+    int rowCount( const QModelIndex &parent = QModelIndex() ) const override;
+    int columnCount( const QModelIndex &parent = QModelIndex() ) const override;
+    QVariant data( const QModelIndex &index, int role ) const override;
+    QVariant headerData( int section, Qt::Orientation orientation, int role ) const override;
+    Qt::ItemFlags flags( const QModelIndex &index ) const override;
+    bool setData( const QModelIndex &index, const QVariant &value, int role ) override;
+
+  private:
+
+    class ExpressionContextGenerator: public QgsExpressionContextGenerator
```
nyalldawson: I don't think this is needed -- you could just make QgsFieldMappingModel itself a QgsExpressionContextGenerator and avoid this class.

src/gui/qgsfieldmappingmodel.cpp
```diff
@@ -0,0 +1,434 @@
...
+}
+
+bool QgsFieldMappingModel::setData( const QModelIndex &index, const QVariant &value, int role )
+{
+  if ( index.isValid() )
+  {
+    if ( role == Qt::EditRole )
+    {
+      Field &f = mMapping[index.row()];
+      switch ( index.column() )
```
nyalldawson: same here -- i think you could cast index.column() instead of the case values.

src/gui/qgsfieldmappingmodel.cpp
```diff
@@ -0,0 +1,434 @@
...
+
+  // Always swap down
+  const int row { up ? index.row() - 1 : index.row() };
+  // Range checking
+  if ( row < 0 || row + 1 >= rowCount( QModelIndex() ) )
+  {
+    return false;
+  }
+  beginMoveRows( QModelIndex( ), row, row, QModelIndex(), row + 2 );
+#if QT_VERSION < QT_VERSION_CHECK(5, 13, 0)
```
nyalldawson: :+1:

src/gui/qgsfieldmappingmodel.cpp
```diff
@@ -0,0 +1,434 @@
...
+#endif
+  endMoveRows();
+  return true;
+}
+
+QString QgsFieldMappingModel::bestMatchforField( const QgsFieldMappingModel::Field &f, QStringList &excludedFieldNames )
+{
+  // Search for fields in the source
+  // 1. match by name
+  for ( const auto &sf : qgis::as_const( mSourceFields ) )
```
nyalldawson: Can you avoid the auto here

src/gui/qgsfieldmappingmodel.h
```diff
@@ -0,0 +1,165 @@
...
+
+        const QgsFields *mSourceFields;
+
+    };
+
+    QgsFieldConstraints::Constraints fieldConstraints( const QgsField &field ) const;
+
+    bool moveUpOrDown( const QModelIndex &index, bool up = true );
+
+    QString bestMatchforField( const QgsFieldMappingModel::Field &field, QStringList &excludedFieldNames );
```
nyalldawson: Some docs explaining exactly what this does would be useful

src/gui/qgsfieldmappingmodel.cpp
```diff
@@ -0,0 +1,434 @@
...
+  beginResetModel();
+  for ( const Field &f : qgis::as_const( mMapping ) )
+  {
+    if ( f.expression.isField() )
+    {
+      usedFields.push_back( f.expression.expression().mid( 1, f.expression.expression().length() - 2 ) );
+    }
+  }
+  // not const on purpose
+  for ( Field &f : mMapping )
```
nyalldawson: I don't think you're allowed to edit a qt container while iterating over it this way. You'd need to do the full `for ( auto it = mMapping.begin(), .... ) ` approach.

src/gui/qgsfieldmappingmodel.cpp
```diff
@@ -0,0 +1,434 @@
...
+}
+
+void QgsFieldMappingModel::setDestinationEditable( bool destinationEditable )
+{
+  mDestinationEditable = destinationEditable;
+}
+
+const QMap<QVariant::Type, QString> QgsFieldMappingModel::dataTypes() const
+{
+  static const QMap<QVariant::Type, QString> sDataTypes
```
nyalldawson: maybe add a comment here (and in the memory provider) that the two lists needs to be synchronized. (Or better - move it to a static member in QgsVectorDataProvider? Something like `allKnownFieldTypes` ?

src/gui/qgsfieldmappingmodel.h
```diff
@@ -0,0 +1,165 @@
...
+     * The Field struct holds information about a mapped field
+     */
+    struct Field
+    {
+      //! The original name of the field
+      QString originalName;
+      //! The field in its current status (it might have been renamed)
+      QgsField field;
+      //! The expression for the mapped field from the source fields
+      QgsExpression expression;
```
nyalldawson: Does this need to be a QgsExpression? As far as I can see we could get away with a QString alone, and avoid the unnecessary expression parsing.

nyalldawson (commented):
Looks fantastic, lovely work!

src/gui/qgsfieldmappingmodel.cpp
m-kuhn: You can introduce extra `{}` wherever you want. They'll act as scope for variables. The only place where I have found that useful so far is inside case labels.

elpaso: that's just a leftover from a cut&paste, I'll remove it.

src/gui/qgsfieldmappingmodel.cpp
elpaso: they are needed after the `enum` became a `class enum` but I agree that it is better to case the `col`

src/gui/qgsfieldmappingmodel.cpp
elpaso: well, they are not the same list: in the memory provider we have multiple supported entries for the same QVariant::Type, I simplified the list here by providing a single entry for each of the duplicated QVariant::Type basic types.

src/gui/qgsfieldmappingmodel.h
elpaso: Yes, I though about it and my rationale was the perhaps having a QgsExpression would have spared additional conversions when checking for validity. But I'm ok with using a string here.

src/gui/qgsfieldexpressionwidget.h
elpaso: maybe in a separate PR

src/gui/qgsfieldmappingmodel.cpp
```diff
@@ -94,64 +94,75 @@ QVariant QgsFieldMappingModel::data( const QModelIndex &index, int role ) const
...
+          QStringList constraintDescription;
+          if ( constraints.testFlag( QgsFieldConstraints::Constraint::ConstraintUnique ) )
+          {
+            constraintDescription.push_back( tr( "Unique" ) );
+          }
+          if ( constraints.testFlag( QgsFieldConstraints::Constraint::ConstraintNotNull ) )
+          {
+            constraintDescription.push_back( tr( "Not null" ) );
+          }
+          if ( constraints.testFlag( QgsFieldConstraints::Constraint::ConstraintNotNull ) )
```
gacarrillor: `ConstraintExpression`

elpaso: what?

gacarrillor: This if should be with `ConstraintExpression`, `ConstraintNotNull` is in the previous if.

elpaso: oh, I see, thank you.

src/gui/qgsfieldmappingmodel.cpp
```diff
@@ -33,39 +33,43 @@ QVariant QgsFieldMappingModel::headerData( int section, Qt::Orientation orientat
...
+          {
+            return tr( "Length" );
+          }
+          case ColumnDataIndex::DestinationPrecision:
+          {
+            return tr( "Precision" );
+          }
+          case ColumnDataIndex::DestinationConstraints:
+          {
+            return tr( "Constraints" );
```
gacarrillor: Could you please replace "Constraints" by "Template properties" as agreed in #31387?

elpaso: mmm, I didn't do it on purpose. The widget is not mean to be used only in the refactoring alg, and "Template properties" for a column that contains information about constraints seems rather obscure to me.

@nyalldawson what do you think?

m-kuhn: In this case it would be best if it's possible to override it from the refactoring algorithm? After all the refactoring algorithm is a very prominent place for this widget.

(For the record, I can also live with "Constraints" but it would be a pity to keep this PR in stale for longer because of this label)

gacarrillor: > (For the record, I can also live with "Constraints" but it would be a pity to keep this PR in stale for longer because of this label)

Same here. Just mentioned it because it was the consensus here: https://github.com/qgis/QGIS/pull/31387#issuecomment-586851400 and because I understood that the "Template properties" column could have more than constraints, e.g., show if the current field has a relation reference or a default value. But it is up to you, of course.

nyalldawson: Can we merge for now, and I'll address this in upcoming followups?

nyalldawson: Gentle bump

nyalldawson:
Superseded,

PR #35153, opened 2020-03-17
FIX #31036 Remove color scheme doesn't work

### LINKED ISSUE #31036: Remove color scheme doesn't work
In the _settings-options-colors_ dialog  I can arrange my color palettes.
Adding and importing works fine, removing a palette howerver doesn't work. I get a confirmation dialog ("are you sure you want to remove [palette name]) , but clicking OK doesn't remove the palette.
After adding a color to the new palette, switch to a different pallette, switch back to the newly added palette I am able to remove the newly added palette,

QGis 3.8.0 / Win10 (x64)

suricactus:


src/core/qgscolorscheme.cpp
```diff
@@ -356,6 +356,12 @@ bool QgsUserColorScheme::erase()
     return false;
   }
 
+  // if file does not exist, nothing to do on the disk, so we can consider erasing done
+  if ( ! QFile::exists( filePath ) )
```
nyalldawson: How does this situation eventuate? Is the file deleted elsewhere in the code?

suricactus: Nope, seems like a "feature" of the module. Check https://github.com/suricactus/QGIS/blob/colorscheme_rm/src/core/qgscolorscheme.cpp#L319-L323 . I was afraid to `touch` the file there, as it might cause other implications.

suricactus: Maybe a little bit more context of the bug would be nice. You cannot delete a color scheme only if it is freshly created and you haven't add any colors to it. If you add at least one color, then it works fine.

nyalldawson: Thanks! That helps the review 😁

nyalldawson (approved):

PR #35157, opened 2020-03-17
After fixing a broken layer path, also auto-fix any other broken layers pointing to same folder

nyalldawson:
After fixing a broken layer path, also auto-fix any other broken layers which are pointing to a file in the same folder as the layer we just fixed, (where a matching filename exists in the newly selected folder)

Previously this logic only applied to layers where the file path was identical to the layer just fixed (i.e. multi-layer sources such as gpkg/gdb), now it also works nicely with multi-file directories such as directories full of shapefiles.

(this is SUCH as nice UI fix for broken projects that I'd like to backport it, if acceptable?)


nirvn:
Yes please!

PR #35160, opened 2020-03-18
Rework picture item UI and behavior

### LINKED ISSUE #17061: Add new SVG/Image selector to composer
Author Name: **Nathan Woodrow** (@NathanW2)
Original Redmine Issue: [8257](https://issues.qgis.org/issues/8257)

Assignee: Larry Shaffer

---
Replace image selector with new SVG picker dialog.

nyalldawson:
Adds an explicit choice between SVG or raster image sources, which allows us to clean up the configuration panel for layout pictures by hiding options which don't apply to a certain picture source. Also permits us to:
- Reuse the standard svg selector tree widget, which loads images in a background thread and fixes #17061
- Uses the standard SVG and image selector line edit, which permit drag and drop of images and expose options to embed images and link to online sources

Ultimately this is motivated by a desire to allow users to embed images in layouts and layout templates

Sponsored by SLYR

### DIFF
--- a/src/gui/layout/qgslayoutpicturewidget.cpp
+++ b/src/gui/layout/qgslayoutpicturewidget.cpp
@@ -338,6 +338,7 @@ void QgsLayoutPictureWidget::setGuiElementValues()
     {
       whileBlocking( mSvgSourceLineEdit )->setSource( mPicture->picturePath() );
 
+      mBlockSvgModelChanges++;
       QAbstractItemModel *m = viewImages->model();
       QItemSelectionModel *selModel = viewImages->selectionModel();
       for ( int i = 0; i < m->rowCount(); i++ )
@@ -350,6 +351,7 @@ void QgsLayoutPictureWidget::setGuiElementValues()
           break;
         }
       }
+      mBlockSvgModelChanges--;
     }
     else if ( mRadioRaster->isChecked() )
     {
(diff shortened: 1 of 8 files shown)


src/gui/layout/qgslayoutpicturewidget.cpp
```diff
@@ -450,14 +330,39 @@ void QgsLayoutPictureWidget::setGuiElementValues()
       mAnchorPointComboBox->setEnabled( false );
     }
 
+    whileBlocking( mRadioSVG )->setChecked( mPicture->mode() == QgsLayoutItemPicture::FormatSVG );
```
nirvn: What happens if the mode is unknown?

nirvn (approved):
Nice cleanup.

src/gui/layout/qgslayoutpicturewidget.cpp
nyalldawson: It shouldn't actually be possible -- unknown is now just a way of avoiding stable api breakage (i.e. scripts which don't explicitly set the mode)

nirvn: So when you load a pre 3.14 project, the picture items will automatically have their modes set to `QgsLayoutItemPicture::FormatSVG` ?

nyalldawson: Yep - it's done internally as soon as the picture source is set

PR #35162, opened 2020-03-18
PostgreSQL provider don't cast bigint PKs to text

### LINKED ISSUE #34077: QGIS casts bigint Primary key field to a text field

### LINKED ISSUE #34264: [postgres] Performance issue: Fields are converted to text
**Describe the bug**

It appears that values read from a PostgreSQL source are first converted to text when the SQL query is crafted. This results in unneeded extra space allocated and transferred on the network.

Sources:

- https://github.com/qgis/QGIS/blob/master/src/providers/postgres/qgspostgresfeatureiterator.cpp#L639
- https://github.com/qgis/QGIS/blob/master/src/providers/postgres/qgspostgresconn.cpp#L1622

It seems the conversion to text is here on purpose but I cannot find the rationale behind it. Maybe to ease some other parts of code ? Could anyone shed some light ?

**How to Reproduce**

You can reproduce easily by looking at queries that QGIS generates on the PostgreSQL server (like with a call to `pg_stat_statements`). The query looks like `select "field1"::text, "field2"::text from ...` whatever the types of `field1`, `field2`are ...

**QGIS and OS versions**

https://github.com/qgis/QGIS/commit/e64a6acd38

espinafre:
This PR is instead of botched #34780 (because of my mistakes with git merging/rebasing); please see the whole discussion there. In addition, a test case for updating tables with positive, zero, negative PKs; I couldn't find a way to test if the queries are being sent to the PostgreSQL backend as designed from within the testing framework. @elpaso , @m-kuhn , is this mergeable? I'm hoping to backport this in time for the next 3.10 and 3.12 point releases.

Fixes #34077
Partially fixes #34264


tests/src/python/test_provider_postgres.py
```diff
@@ -429,6 +429,89 @@ def testPktIntInsert(self):
         self.assertNotEqual(f[0]['pk'], NULL, f[0].attributes())
         vl.deleteFeatures([f[0].id()])
 
+    def testPktUpdateBigintPk(self):
+        """Test if we can update objects with positive, zero and negative bigint PKs."""
+        vl = QgsVectorLayer('{} sslmode=disable srid=4326 key="pk" table="qgis_test".{} (geom)'.format(self.dbconn, 'bigint_pk'), "bigint_pk", "postgres")
+        dp = vl.dataProvider()
+        flds = dp.fields()
```
m-kuhn:
```suggestion
        flds = vl.fields()
```

In this particular case it's completely equal to what you proposed, but let's use best practice in our code base.

tests/src/python/test_provider_postgres.py
```diff
@@ -429,6 +429,89 @@ def testPktIntInsert(self):
...
+            elif ft['value'] == '0th value':
+                statuses[2] = 1
+            elif ft['value'] == '-1th value':
+                statuses[3] = 1
+        self.assertTrue( all ( x == 1 for x in statuses ) )
+
+    def testPktUpdateBigintPkNonFirst(self):
+        """Test if we can update objects with positive, zero and negative bigint PKs in tables whose PK is not the first field"""
+        vl = QgsVectorLayer('{} sslmode=disable srid=4326 key="pk" table="qgis_test".{} (geom)'.format(self.dbconn, 'bigint_non_first_pk'), "bigint_non_first_pk", "postgres")
+        flds = vl.dataProvider().fields()
```
m-kuhn:
```suggestion
        flds = vl..fields()
```

m-kuhn (approved):
It looks generally good to me, thanks.
I can imagine that adding tests for what's sent to the db is hard to do, so I'm good with sticking to functional tests and trusting your performance tests (and hoping no regressions are introduced for the latter).

src/providers/postgres/qgspostgresconn.cpp
```diff
@@ -203,6 +203,7 @@ QgsPostgresConn::QgsPostgresConn( const QString &conninfo, bool readOnly, bool s
   , mOpenCursors( 0 )
   , mConnInfo( conninfo )
   , mGeosAvailable( false )
+  , mProjAvailable( false )
```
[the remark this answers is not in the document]
espinafre: This was because there was an omission in debugging when the original (unknown to me) programmer queried for the database capabilities; he meant to output the PostgreSQL proj version in debug output, but missed that step, which caused a Qt warning.

tests/src/python/test_provider_postgres.py
espinafre: Thank you for your suggestions, this new commit solves those issues (and does away with the dp variable altogether!)

elpaso:
LGTM, the test coverage on PG is pretty good, so I guess that if the tests are passing we should not expect regressions.

espinafre:
Travis is still choking on formatting, it is as if it hasn't seen ad8f89b . Locally running `ctest -V -R qgis_indentation` results in success.

m-kuhn:
See https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTING.md#contributing-to-qgis for some hints how to get your code formatted automatically

espinafre:
I've ran prepare-commit.sh before pushing the C++ code, but I think I missed it when adding the test code, thanks. What can I do about the failed qgis_sip_uptodate test? It passes when I run locally with `ctest -V -R qgis_sip_uptodate`

espinafre:
This is the full output of my local ctest -v -R qgis_sip_uptodate (which fails on Travis):

```
UpdateCTestConfiguration  from :/home/jose/sw/qgis/QGIS/build-master/DartConfiguration.tcl
Parse Config file:/home/jose/sw/qgis/QGIS/build-master/DartConfiguration.tcl
 Add coverage exclude regular expressions.
 Add coverage exclude: /CMakeFiles/CMakeTmp/
SetCTestConfiguration:CMakeCommand:/usr/bin/cmake
UpdateCTestConfiguration  from :/home/jose/sw/qgis/QGIS/build-master/DartConfiguration.tcl
Parse Config file:/home/jose/sw/qgis/QGIS/build-master/DartConfiguration.tcl
Test project /home/jose/sw/qgis/QGIS/build-master
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 577
    Start 577: qgis_sip_uptodate

577: Test command: /home/jose/sw/qgis/QGIS/tests/code_layout/test_sipfiles_uptodate.sh
577: Test timeout computed to be: 1500
1/1 Test #577: qgis_sip_uptodate ................   Passed   85.44 sec

The following tests passed:
        qgis_sip_uptodate

100% tests passed, 0 tests failed out of 1

Total Test time (real) =  85.46 sec
```

m-kuhn:
> I've ran prepare-commit.sh before pushing the C++ code,

You need to run it before committing (it will only work on uncommitted files to save some time).
Does `./scripts/sipify_all.sh` modify files?

espinafre:
> > I've ran prepare-commit.sh before pushing the C++ code,
> 
> You need to run it before committing (it will only work on uncommitted files to save some time).
> Does `./scripts/sipify_all.sh` modify files?

Only testdata/untracked files

m-kuhn:
Ah wait, that remaining thing seems to be something that changed on master meanwhile. The test has a weird behavior with respect to comparing "master merged into pr" with plain "pr".
Can you do a rebase on latest master locally again and force push?

```
git fetch upstream
git rebase upstream/master
git push -f
```

espinafre:
> Ah wait, that remaining thing seems to be something that changed on master meanwhile. The test has a weird behavior with respect to comparing "master merged into pr" with plain "pr".
> Can you do a rebase on latest master locally again and force push?
> 
> ```
> git fetch upstream
> git rebase upstream/master
> git push -f
> ```

Done. Let's see how Travis likes this, thank you for the tip.

espinafre:
Travis failed the same test...

espinafre:
I've done a make clean and re-ran ./scripts/sipify_all.sh, which made the change in commit 6767395; this is what I just pushed.

espinafre:
Now another test failed on Travis, but passed locally: PyQgsPostgresRasterProvider. I've got no idea why.

espinafre:
Did nothing this time, just fetched upstream/master, rebased against it, and pushed -f, according to @m-kuhn instructions. The previous failing test in Travis still passes locally (PyQgsPostgresRasterProvider).

nyalldawson:
If you really want to test this fully (and I would strongly suggest you do), then you need to implement the ProviderTestCase in a class which uses a bigint primary key.

Checkout TestPyQgsPostgresProviderCompoundKey in test_provider_postgres.py for a template on how to do this

espinafre:
> If you really want to test this fully (and I would strongly suggest you do), then you need to implement the ProviderTestCase in a class which uses a bigint primary key.
> 
> Checkout TestPyQgsPostgresProviderCompoundKey in test_provider_postgres.py for a template on how to do this

Hi, don't you think that the methods testPktUpdateBigintPk and testPktUpdateBigintPkNonFirst that I wrote in 0853029 are enough? Those exercise loading and saving features with bigint PKs, for a table with an ordinary bigint pk and another table whose PK is not the first one defined.

nyalldawson:
> Hi, don't you think that the methods testPktUpdateBigintPk and testPktUpdateBigintPkNonFirst that I wrote in 0853029 are enough? Those exercise loading and saving features with bigint PKs, for a table with an ordinary bigint pk and another table whose PK is not the first one defined.

Depends how confident you want to be. If you implement the ProviderTestCase then you can 100% sleep at night knowing you're covered in all existing conceivable situations, and will also be **automatically** covered whenever new functionality is added to the providers...

nyalldawson:
(just to clarify: the provider test class is a mega stress test. It throws thousands of edge cases at the provider to make sure they handle them correctly).

espinafre:
> (just to clarify: the provider test class is a mega stress test. It throws thousands of edge cases at the provider to make sure they handle them correctly).

Thanks for the pointer, and for enlightening me as always, I'm working on it. Just subclassing the ProviderTestCase and adjusting the data source? Is there any requirements for the data source (geometry type for example)? Also, is there a need for one subclass for each kind of table (that is, with a single ordinary PK, with a PK which is not defined as the first field, with a composite PK in which at least one member is a bigint...)? Sorry for the torrent of questions, but you provoked it :D

espinafre:
I still cannot figure out why the test PyQgsPostgresRasterProvider fails on Travis but not on my own system (PostgreSQL 12 + postgisraster 3.0), neither on Azure.

nyalldawson:
> I still cannot figure out why the test PyQgsPostgresRasterProvider fails on Travis but not on my own system (PostgreSQL 12 + postgisraster 3.0), neither on Azure.

It's unrelated -- it fails randomly regardless of the PR

nyalldawson:
> Just subclassing the ProviderTestCase and adjusting the data source? Is there any requirements for the data source (geometry type for example)?

Yes - look at how "someDataCompound" is loaded in tests/testdata/provider/testdata_pg.sql. You need to load in the same reference data rows (but with the different primary key type). Everything from the "pk" field onward needs to stay the same. 

If you want to also run the editing reference tests then you need to implement `getEditableLayer` in your test subclass. See TestPyQgsPostgresProvider.getEditableLayer for how this is done (basically it needs to drop the existing editable version of the table, then recreate it anew, since each test in the editable part of the test suite wants a clean copy of this table to work from, and it'll mangle it up during its tests.)

> Also, is there a need for one subclass for each kind of table (that is, with a single ordinary PK, with a PK which is not defined as the first field, with a composite PK in which at least one member is a bigint...)

If they aren't already covered, then yes -- this would be very valuable! It's probably the biggest (only?) remaining gap in the test coverage of postgres provider.

elpaso:
> > I still cannot figure out why the test PyQgsPostgresRasterProvider fails on Travis but not on my own system (PostgreSQL 12 + postgisraster 3.0), neither on Azure.
> 
> It's unrelated -- it fails randomly regardless of the PR

I hoped it was fixed. Is it still flacky?

roya0045:
@elpaso a rebase should fix this, but now the azure build is bjorked.

espinafre:
New fetch/rebase/push dance, added the tests that @nyalldawson asked for; are those right? They sometimes fail, sometimes succeed on my local machine, depending on whether I set LC_COLLATE to en_US or C. Let's see how Travis like them.

espinafre:
Is it possible to rerun travis-ci without a new push?

espinafre:
Hi, shall I do a new rebase, since travis looks fixed?

m-kuhn:
You can try.
But for the travis it's almost the same if you do rebase or close/reopen pull request. Is it possible that your changes caused the current travis failures?

espinafre:
> You can try.
> But for the travis it's almost the same if you do rebase or close/reopen pull request. Is it possible that your changes caused the current travis failures?

I don't believe so; Travis fails on OTB testing, which I didn't touch at all. I'll try to run some more tests locally, then rebase/push again to see what happens.

m-kuhn:
Check again after @jef-n has closed/reopened

espinafre:
I've added a FeatureSourceTestCase subclass to the test cases, as instructed by @nyalldawson , with the getSource/getEditableLayer implemented. That's why it took so long, and it was good too, because I caught an omission that I corrected in my last "real" commit ( 742454b ).

I've noted on the method overrides (with TODO annotations) where the code diverges from the original classes, and why, in order to correct those issues in the future.

All those tests pass locally, on my machine, with PostgreSQL version 12.1 and LANG=en_US.UTF-8.

Is this mergeable?

espinafre:
I think @nyalldawson should take a look at the Python test ( d4672b0 ). Also, this should be backportable to 3.12 and 3.10 without issues. The github doesn't allow me to add @nyalldawson to the list of reviewers. @m-kuhn , can you do that? Thank you very much.

elpaso:
@espinafre we don't usually assign anything to anyone, unless we know for sure that the developer agreed.

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -559,10 +567,22 @@ QString QgsPostgresUtils::whereClause( QgsFeatureId featureId, const QgsFields &
...
       Q_ASSERT( pkAttrs.size() == 1 );
-      whereClause = QStringLiteral( "%1=%2" ).arg( QgsPostgresConn::quotedIdentifier( fields.at( pkAttrs[0] ).name() ) ).arg( featureId );
-      break;
+      QVariantList pkVals = sharedData->lookupKey( featureId );
+      if ( !pkVals.isEmpty() )
+      {
+        QgsField fld = fields.at( pkAttrs[0] );
+        whereClause = conn->fieldExpression( fld );
+        if ( !pkVals[0].isNull() )
+          whereClause += "=" + pkVals[0].toString();
```
m-kuhn, line 580: This should be more efficient. Not completely sure if it works. If not, go for `QStringLiteral`

```suggestion
          whereClause +=  '=' + pkVals[0].toString();
```

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -622,14 +641,41 @@ QString QgsPostgresUtils::whereClause( const QgsFeatureIds &featureIds, const Qg
...
+      QString expr;
+
+      //simple primary key, so prefer to use an "IN (...)" query. These are much faster then multiple chained ...OR... clauses
+      if ( !featureIds.isEmpty() )
+      {
+        QString delim;
+        expr = QStringLiteral( "%1 IN (" ).arg( QgsPostgresConn::quotedIdentifier( fields.at( pkAttrs[0] ).name() ) );
+
+        const auto constFeatureIds = featureIds;
+        for ( const QgsFeatureId featureId : constFeatureIds )
```
m-kuhn, line 664: Please use this style (the other approach was introduced by an automatic conversion)

```suggestion
        for ( const QgsFeatureId featureId : qgis::as_const( featureIds ) )
```

src/providers/postgres/qgspostgresfeatureiterator.cpp
```diff
@@ -744,20 +744,37 @@ bool QgsPostgresFeatureIterator::getFeature( QgsPostgresResult &queryResult, int
...
+      // database value
+      fid = QgsPostgresUtils::int32pk_to_fid( fid );
+      break;
+
+    case PktUint64:
+    case PktInt64:
+    {
+      QVariantList pkVal;
+
+      int idx = mSource->mPrimaryKeyAttrs[0];
```
m-kuhn, line 763: Use `at` for read only access, `operator[]` has some non obvious side effects to provide a non-const reference.

```suggestion
      int idx = mSource->mPrimaryKeyAttrs.at( 0 );
```

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -80,7 +80,12 @@ QgsPostgresProvider::pkType( const QgsField &f ) const
       // (in which case we could use pktUint64)
       // we'll have to use a Map type.
```
m-kuhn, line 81: This comment is superseded?

m-kuhn (approved):
It looks good to me as far as I can tell. It's not trivial to understand each change (but that's in the nature of this code, not related to your work), so I'll trust the tests and general good look.

Some minor code style requests and we are good to ship it in my opinion.

I would like to wait for one or two months with a backport to LTR to have some time for any eventual regression to be sorted out first.

src/providers/postgres/qgspostgresprovider.cpp
espinafre: It is, I've removed the part that mentions the bug and the treatment as pktUint64.

src/providers/postgres/qgspostgresprovider.cpp
espinafre: May I change this on other parts of the code? Those that handle PktFidMap and PktInt?

src/providers/postgres/qgspostgresprovider.cpp
espinafre: Done, see 9607570

src/providers/postgres/qgspostgresfeatureiterator.cpp
espinafre: Done, see 9607570

src/providers/postgres/qgspostgresprovider.cpp
espinafre: Done, see 9607570 (also changed the other parts of the code)

src/providers/postgres/qgspostgresprovider.cpp
espinafre: Done, see 9607570

espinafre: It does work.

espinafre:
Commit 9607570 addresses and incorporates all suggestions by @m-kuhn . Local tests still pass. Is backporting to 3.12 automatic?

m-kuhn:
> Commit 9607570 addresses and incorporates all suggestions by @m-kuhn . Local tests still pass. Is backporting to 3.12 automatic?

:+1: 

Now it is.

espinafre:
> @espinafre we don't usually assign anything to anyone, unless we know for sure that the developer agreed.

I've said "assigned to me" in jest, along the lines of "the teacher assigned homework to the students" :)

espinafre:
@m-kuhn Does the MXE build failure prevent the merge?

elpaso:
NoOn Apr 17, 2020 20:15, "José de Paula Rodrigues N. Assis" <notifications@github.com> wrote:
@m-kuhn Does the MXE build failure prevent the merge?

m-kuhn:
Thanks @espinafre nice work!

roya0045:
@espinafre don't worry about MXE, the whole service seems to be having issues today.

PR #35184, opened 2020-03-18
Show data defined buttons for dynamic parameters even in batch mode

nyalldawson:
I recall I initially hid these in batch mode because it didn't seem like there'd be a use case for them, but recent discussion on the dev mailing list has revealed otherwise!

Anyway, they work fine in batch mode, we just need to show them.


nyalldawson:
@alexbruy interested in your opinion here

alexbruy (approved):

PR #35185, opened 2020-03-18
In batch mode, don't try to run rows with invalid parameter values

nyalldawson:
Previously we would show a warning about these, but then go ahead and try to run the row anyway (using an empty set of parameters), resulting in the log being filled with confusing error messages.

Instead, keep the existing warning advising about which values are invalid, but skip the affected row and don't try to run it at all.

@alexbruy the other option would be to not run ANY rows if any are invalid, and completely block the execution until all rows are fixed. What do you think is better?


nyalldawson:
Marked as backport, but NOT for this upcoming cycle of releases. This can safely wait till the next round.

alexbruy:
I think better to skip only invalid rows and proceed with other. We already show a warning, so users have all necessary information to fix things.

alexbruy (approved):

PR #35188, opened 2020-03-19
FIX #15144 Add tooltip on fields in identify window/attribute window/feature form

### LINKED ISSUE #15144: add tooltip on fields in identify window / attribute window / feature form
Author Name: **Regis Haubourg** (@haubourg)
Original Redmine Issue: [5559](https://issues.qgis.org/issues/5559)

Redmine category:gui

---
Many times, fields name are short and not explicit enough to understand their meaning. 
Comments are already available in fields properties and are perfect for that. Thoses comments are really nicely integrated when using Postgis.
 
I would love to have tooltips displaying comments, real name if alias is used and maybe field type and precision. see attached pictures for examples
Anyone interested in it?

---

- [qgis_field_comment_tooltip.jpg](https://issues.qgis.org/attachments/download/4453/qgis_field_comment_tooltip.jpg) (Regis Haubourg)

### LINKED ISSUE #15670: Show count in Identify Results dialog
Author Name: **Aren Cambre** (Aren Cambre)
Original Redmine Issue: [6424](https://issues.qgis.org/issues/6424)

Redmine category:gui

---
When I use the *Identify Features* feature, the *Identify Results* dialog should show a count of items found. Occasionally, especially with some of my bad datasets, *Identify Features* turns up more items than I can count at a glance.

suricactus:
- added new `QString QgsField::displayType( const bool showConstraints = false )` to unify the display of field types
- added new argument `expression` to `QgsFieldModel::fieldToolTip( const QgsField &field, const QString &expression = QStringLiteral() )`. 
Now the tooltip shows:
```
<alias> (<field>)\n<type>\n<comment>\n<expression>
``` 

with appropriate formatting
- added meaningful tooltips in the "Identify Results" dialog
- unified the tooltips across "Feature Attributes" form, "Attribute Table" and "Identify Tool" 

Fixes #15144

## Description
Tooltip field:
image
Tooltip virtual field:
image

### DIFF
--- a/src/core/qgsfieldmodel.h
+++ b/src/core/qgsfieldmodel.h
@@ -135,7 +135,14 @@ class CORE_EXPORT QgsFieldModel : public QAbstractItemModel
      * like the field name, alias and type.
      * \since QGIS 3.0
      */
-    static QString fieldToolTip( const QgsField &field, const QString &expression = QString() );
+    static QString fieldToolTip( const QgsField &field );
+
+    /**
+     * Returns a HTML formatted tooltip string for a \a field, containing details
+     * like the field name, alias, type and expression.
+     * \since QGIS 3.14
+     */
+    static QString fieldToolTipExtended( const QgsField &field, const QgsVectorLayer *layer );
 
     /**
      * Manually sets the \a fields to use for the model.
--- a/src/core/qgsfieldmodel.h
+++ b/src/core/qgsfieldmodel.h
@@ -135,7 +135,7 @@ class CORE_EXPORT QgsFieldModel : public QAbstractItemModel
      * like the field name, alias and type.
      * \since QGIS 3.0
      */
-    static QString fieldToolTip( const QgsField &field );
+    static QString fieldToolTip( const QgsField &field, const QString &expression = QString() );
 
     /**
      * Manually sets the \a fields to use for the model.
--- a/python/core/auto_generated/qgsfield.sip.in
+++ b/python/core/auto_generated/qgsfield.sip.in
@@ -104,6 +104,18 @@ represents.
 .. versionadded:: 3.12
 %End
 
+
+    QString displayType( bool showConstraints = false ) const;
+%Docstring
+Returns the type to use when displaying this field, including the length and precision of the datatype if applicable.
+
+This will be used when the full datatype with details has to displayed to the user.
+
+.. seealso:: :py:func:`type`
+
+.. versionadded:: 3.14
+%End
+
     QVariant::Type type() const;
 %Docstring
 Gets variant type of the field as it will be retrieved from data source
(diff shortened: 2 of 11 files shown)


nyalldawson:
Any chance you could split this to multiple PRs? There's unrelated changes here which would be nice to discuss on their own merits.

suricactus:
> added "selected features: " next to the layer name in the "Identify Results" dialog
This is the only part I can split, the rest are interdependent. Is this enough?

nyalldawson:
Yep, I would like to discuss that change seperately

suricactus:
@nyalldawson now you can have a look.

suricactus:
This is unrelated travis error, it's funny how a change in the qt ui "breaks" the postgresql raster provider. :)

src/core/qgsfieldmodel.h
```diff
@@ -135,7 +135,7 @@ class CORE_EXPORT QgsFieldModel : public QAbstractItemModel
      * like the field name, alias and type.
      * \since QGIS 3.0
      */
-    static QString fieldToolTip( const QgsField &field );
+    static QString fieldToolTip( const QgsField &field, const QString &expression = QString() );
```
nyalldawson: I don't think the expression handling belongs here -- I think it's cleaner to just add this line in QgsAttributeTableModel/QgsIdentifyResultsDialog manually

nyalldawson:
+1 to the result, just a minor code cleanup wanted

src/core/qgsfieldmodel.h
suricactus: For sure it's a dirty solution, I will remove it. Having a helper for this in the future might be handy, so the tooltips remain the same style across all widgets.

nyalldawson: I agree about the helper. What about moving it to a static method here?

   static QString appendExpressionToTooltip( const QString& tooltip, const QString& expression )

?

suricactus: Why pass the tooltip string at all? This gives too much freedom which may bring another round of unconsistancies? I would say keep the same approach as I did originally, just adding a new function for that.

```static QString fieldToolTipWithExpression( const QgsField &field, const QString &expression )```

This way everyone is forced to use tooltips that begin with the same contents, then they are free to append their stuff.

nyalldawson: Hmm, what about

    static QString expandedFieldToolTip( const QgsField &field, QgsVectorLayer* layer )

and just put ALL the logic in there?

suricactus: I like that! But are we sure we will always have the layer in advance? If we are about to create a new layer, it might be missing, so _**maybe**_ overloading will be needed in the future with second parameter being a string?

I only have comment on the name, I would prefer `fieldToolTipExpanded` over `expandedFieldToolTip`, it looks better from API points of view.

nyalldawson (approved):
LGTM!

src/core/qgsfieldmodel.h
suricactus: Have a look now, I have renamed it to `fieldToolTipExtended` as "expanded" seems strange to me? You are native speaker, you know better if the new name makes sense :)

PR #35189, opened 2020-03-19
Make "Ctrl+Shift+A" shortcut to trigger "Deselect current layer" instead of  "Deselect all layers"

suricactus:
As I said in https://github.com/qgis/QGIS/pull/35085 , the "Ctrl+Shift+A" shortcut should be exactly the opposite of "Ctrl+A" to match user expectations.

### DIFF
--- a/src/ui/qgisapp.ui
+++ b/src/ui/qgisapp.ui
@@ -1210,7 +1210,7 @@
     <string>Deselect Features from All Layers</string>
    </property>
     <property name="shortcut">
-    <string>Alt+Shift+A</string>
+    <string>Ctrl+Alt+A</string>
    </property>
   </action>
   <action name="mActionDeselectActiveLayer">
--- a/src/ui/qgisapp.ui
+++ b/src/ui/qgisapp.ui
@@ -1209,6 +1209,9 @@
    <property name="text">
     <string>Deselect Features from All Layers</string>
    </property>
+    <property name="shortcut">
+    <string>Alt+Shift+A</string>
+   </property>
   </action>
   <action name="mActionDeselectActiveLayer">
    <property name="icon">
--- a/src/ui/qgisapp.ui
+++ b/src/ui/qgisapp.ui
@@ -1209,9 +1209,6 @@
    <property name="text">
     <string>Deselect Features from All Layers</string>
    </property>
-   <property name="shortcut">
-    <string>Ctrl+Shift+A</string>
-   </property>
   </action>
   <action name="mActionDeselectActiveLayer">
    <property name="icon">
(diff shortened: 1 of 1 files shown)


uclaros:
Shouldn't there still be a default shortcut for deselect from all layers? ctrl+alt+shift+A maybe?

suricactus:
Users can always change their shortcuts. It is important to have sensible defaults, especially for newcomers.

nyalldawson:
> Shouldn't there still be a default shortcut for deselect from all layers? ctrl+alt+shift+A maybe?

+1 to this -- I think its good to still offer an out-of-the-box standard shortcut for "deselect EVERYTHING"

suricactus:
I've added a "Deselect All" shortcut, but `alt+shift+A`. Under Qt I have observed taht all the "ctrl+shift+alt+<Key>" does not work properly, or at least under Linux. I've just tried that and the shortcut never gets fired as "ctrl+shift+A" seems to steal it. 

So no "ctrl+shift+alt+A" even though Emacs guys would love this :)

uclaros:
Alt+shift is a common shortcut to change keyboard input language causing most combinations of alt+shift+... to not be captured...

suricactus:
Yes, I do suffer from the "Alt+Shift" fever too. Any other idea for combination is appreciated.

uclaros:
How about ctrl+shift+A for deselect all from all layers, since some people may have got used to it and plain shift+A for deselect all from current?

suricactus:
I disagree, `ctrl+<Key>` and `ctrl+shift+<Key>` should always do the opposite of each other. `ctrl+A` selects all features from the current layer, therefore `ctrl+shift+a` should not touch other layers. I have seen a few new users who get frustrated by this behavior. 
I admit it is a "breaking" change, but the new shortcut is safer that the previous one.
I just checked there are not "Alt+Shift+<Key>" shortcuts, so my idea is not an option.
What about "Ctrl+Alt+A", it seems free?

nyalldawson:
Ctrl+Alt+A works for me!

suricactus:
Can someone try to set custom shortcut with "Ctrl+Alt+A" before merging this and confirm it works. My KDE is stealing the shortcut and I cannot find where it goes :(

suricactus:
Okay, found it! On KDE I find it the most useless global shortcut:
`Settings -> (Global) Shortcuts -> Kwin -> Activate Window Demanding Attention`

It works on KDE, XFCE and Windows 10. Ready for merge @nyalldawson .

nyalldawson:
Thanks @suricactus !

DelazJ:
Under Layout dialog, Edit menu, there's a "Deselect All" entry that uses the new icon but still points to the `ctrl+shift+a`. For consistency, should it not also be updated? master 3ad17e526b

borysiasty:
As mentioned in the PR above, Ctrl+Alt = AltGr in windows (for some weird reason there isn't even any way to disable it), so it steals diacritical characters from some languages :( 

The simplest solution would be to just drop Ctrl+Alt+A and leave the good old action (deselect from all layers) without any shortcut. I found myself always using the toolbar button for it anyway.

PR #35190, opened 2020-03-19
 FIX #15670 Show count in Identify Results dialog

### LINKED ISSUE #15670: Show count in Identify Results dialog

suricactus:
Added "selected features: <FEATURES_COUNT>" next to the layer name in the "Identify Results" dialog. 

Fixes #15670

image

### DIFF
--- a/src/app/qgsidentifyresultsdialog.cpp
+++ b/src/app/qgsidentifyresultsdialog.cpp
@@ -481,6 +481,19 @@ QTreeWidgetItem *QgsIdentifyResultsDialog::layerItem( QObject *object )
   return nullptr;
 }
 
+QTreeWidgetItem *QgsIdentifyResultsDialog::findResultFeaturesItem( QTreeWidgetItem *layerItem )
+{
+  for ( int i = 0; i < layerItem->childCount(); i++ )
+  {
+    QTreeWidgetItem *item = layerItem->child( i );
+
+    if ( item->data( 0, Qt::UserRole ).toString() == QLatin1String( "count" ) )
+      return item;
+  }
+
+  return nullptr;
+}
+
 void QgsIdentifyResultsDialog::addFeature( const QgsMapToolIdentify::IdentifyResult &result )
 {
   switch ( result.mLayer->type() )
@@ -508,7 +521,7 @@ void QgsIdentifyResultsDialog::addFeature( QgsVectorLayer *vlayer, const QgsFeat
 
   if ( !layItem )
   {
-    layItem = new QTreeWidgetItem( QStringList() << vlayer->name() << tr( "Selected features: %1" ).arg( 0 ) );
+    layItem = new QTreeWidgetItem( QStringList() << vlayer->name() );
     layItem->setData( 0, Qt::UserRole, QVariant::fromValue( qobject_cast<QObject *>( vlayer ) ) );
     lstResults->addTopLevelItem( layItem );
 
@@ -526,7 +539,19 @@ void QgsIdentifyResultsDialog::addFeature( QgsVectorLayer *vlayer, const QgsFeat
   featItem->setData( 0, Qt::UserRole + 1, mFeatures.size() );
   mFeatures << f;
   layItem->addChild( featItem );
-  layItem->setText( 1, tr( "Selected features: %1" ).arg( layItem->childCount() ) );
+
+  QTreeWidgetItem *resultFeaturesCountItem = findResultFeaturesItem( layItem );
+  if ( ! resultFeaturesCountItem )
+  {
+    resultFeaturesCountItem = new QTreeWidgetItem( QStringList() << tr( "(Count Results)" ) );
+    resultFeaturesCountItem->setData( 0, Qt::UserRole, QLatin1String( "count" ) );
+    layItem->insertChild( 0, resultFeaturesCountItem );
+  }
+
+  if ( resultFeaturesCountItem )
+  {
+    resultFeaturesCountItem->setText( 1, QString::number( layItem->childCount() - 1 ) );
+  }
 
   if ( derivedAttributes.size() >= 0 )
   {
@@ -1106,6 +1131,9 @@ void QgsIdentifyResultsDialog::editingToggled()
   for ( i = 0; i < layItem->childCount(); i++ )
   {
     QTreeWidgetItem *featItem = layItem->child( i );
+
+    if ( featItem->data( 0, Qt::UserRole ).toString() == QLatin1String( "result features" ) )
+      continue;
 
     int j;
     for ( j = 0; j < featItem->childCount() && featItem->child( j )->data( 0, Qt::UserRole ).toString() != QLatin1String( "actions" ); j++ )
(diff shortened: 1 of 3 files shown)


uclaros:
Nice! But *Selected Features* is a little misguiding, the features are not selected. I'd prefer just: *x features*

nyalldawson:
My concern with this is that the extra text in the second column makes the header rows less easy to immediately identify -- it just becomes one big block of "table" cells.

What about instead of putting things in this second column, you instead just append a "(2)", "(3)" count suffix to the end of the layer name IF more than one feature was identified? That would give the same info to the user without affecting the tree readability.

suricactus:
@nyalldawson I do agree this makes the whole "tree widget" less focused. I have evaluated several options and chose to put the features count in the second column because:
1) **It is easy to spot the number of features at any time.** On the screenshot above, you see that layer names are extremely long and putting the number as a suffix will bring zero UX. All the space in the first column is already wasted. Tooltips are also not an option to me, they should give more additional clarification, rather than that being sole source of information. 
2) **Clearly stays away from the layer name.** It is common for quick and dirty data checks to download a bunch of layers with the same names, which browsers suffix with a number in the end.
3) @uclaros **The features are definitely "selected" in the context of the identify tool.** I don't argue is this is the best terminology, but for sure this is what is used both in the interface and the code. I would change the text to "result features: %d", as "features" is also misleading if we are talking about the totals or not.

An alternative solution to this dead end situation is to add additional row just below the layer itself, surrounded by brackets? Similarly to how "(Derived)" and "(Actions)" are shown under each feature?
```
layer_name_here			|
	(result features)	| 2
	fid					| 1
		fid				| 1
		name			| Feature Name 1
	fid					| 3
		fid				| 3
		name			| Feature Name 1
layer_name_here (1)		|
	(result features)	| 1
	fid					| 1
		fid				| 1
		name			| Feature Name 1
```

nyalldawson:
> The features are definitely "selected" in the context of the identify tool

No -- they are "identified", not "selected" :smile: 

> An alternative solution to this dead end situation is to add additional row just below the layer itself, surrounded by brackets?

This works for me! Could we make it only show if > 1 feature is identified?

suricactus:
> No -- they are "identified", not "selected" 

image

I guess I have to change the tooltip to "Copy the Identified Feature to Clipboard" too :)


> This works for me! Could we make it only show if > 1 feature is identified?

deal!

nyalldawson:
> I guess I have to change the tooltip to "Copy the Identified Feature to Clipboard" too :)

Yep, good catch!

uclaros:
> I guess I have to change the tooltip to "Copy the Identified Feature to Clipboard" too :)

No! Only the selected feature of the identified features is copied. In the context of the Identify Results Panel the feature is selected and one cannot confuse this with the actual selected features of the layer.



> An alternative solution to this dead end situation is to add additional row just below the layer itself, surrounded by brackets? Similarly to how "(Derived)" and "(Actions)" are shown under each feature?

This has the major drawback of not having the info available when the tree is collapsed, and it's very useful info to have available on a collapsed tree, like a summary.



> This works for me! Could we make it only show if > 1 feature is identified?

This is kind of inconsistent look, to sometimes have an extra row.

> it just becomes one big block of "table" cells.

This bothered me too initially, but I believe the best solution after all is to still have the info at that place but make the displayed value stand out. This could be done by making it very minimal, like for example `[42]`, or very identifiable like `#### 42 Features ####` or something.

suricactus:
> No! Only the selected feature of the identified features is copied. In the context of the Identify Results Panel the feature is selected and one cannot confuse this with the actual selected features of the layer. OTOH one could be confused and expect "Copy the Identified Feature to Clipboard" to copy all identified features.

I am not so sure how this is different from the initial discussion on "selected" vs "identified"... That's why I propose **"the Identified Feature"**. Or "the Current Identified Feature" or "the Current Feature" or "the Highlighted Feature". But it would be much better if we get rid of the "selected" word within the UI of the panel, so no discussion like this will ever arise.

> This has the major drawback of not having the info available when the tree is collapsed, and it's very useful info to have available on a collapsed tree, like a summary.

I do agree this is a drawback. But how many layers and features you usually identify? The tool is rather slow on bigger areas or many layers, so this information is always a click away.

> This is kind of inconsistent look, to sometimes have an extra row.

I tried this, it looks ugly when some layers have one feature and some don't. So in the end there is not condition to show it.

> This bothered me too initially, but I believe the best solution after all is to still have the info at that place (ie at the value column) but make the displayed value stand out. This could be done by making it very minimal, like for example [42], or very identifiable like #### 42 Features #### or something.

Having "magic" numbers here and there is a UI practice to be discouraged. We either put it on the same row as the layer with clear explanation (the original PR), or put it as a child (see below). 

image

uclaros:
> Or "the Current Identified Feature" or "the Current Feature" or "the Highlighted Feature".

^ This! Use *current* or *highlighted*



> Having "magic" numbers here and there is a UI practice to be discouraged. 

Agree. What about same row but formatted to stand out?

suricactus:
Not sure for the tooltip, let's wait for Nyall's input.

> Agree. What about same row but formatted to stand out?

Without proper theming, this can be dangerous.

nyalldawson:
Hmm...

I still think we should re-evaluate just adding the suffix to the layer name. One additional benefit of this is that because we leave the second column empty, we could correctly fix this situation:

image

where the layer name is cut off. Because if we keep the second column clear, we could use QTreeView::setFirstColumnSpanned and then the layer name will nicely overflow into the second column.

uclaros:
> and then the layer name will nicely overflow into the second column

The only problem will be with long layer names the second column will not act as a blank spacer as per your initial concern.
Also on a collapsed tree with many identified layers the count numbers will not be aligned since their positions are dependent on the layer name's length, but that's minor.
I like it better than on a second row.

suricactus:
Even though I don't like mixing user and meta data together, I agree this might look better, stay tuned.

suricactus:
I know @nyalldawson requested to hide the number when only a single feature is identified, but I changed the condition to single feature in single layer. If there are more than one identified feature in one or more layers, then the count is visible. Otherwise imagine some layers having the label and some don't IMHO it will look terrible. 

TBH for me it looks pretty ok, it matches how the feature count works in the "Layers" panel. 
image

If they add "badges" to QT, we can make it awesome, but not now.

suricactus:
Oops, ignore my previous message. Unless I do some ugly stuff, it's not easy to implement the logic I initially stated.

So final update:
- show count if the number of identified features > 1
- span first column
- make it bold, so it can be distinguished (also matches the style in the layers panel)
image

nyalldawson:
Looks beautiful to me, thanks for your flexibility!

nyalldawson (approved):

nirvn:
@suricactus , nice visual improvement, thanks.

suricactus:
The documentation guys should update https://docs.qgis.org/testing/en/docs/user_manual/introduction/general_tools.html#identify

PR #35108, opened 2020-03-16
Fix TaskManager flaky test

troopa81:
## Description

Fix TaskManager subTaskPartialComplete2 flaky test since the work done in #34469

PR #35200, opened 2020-03-19
[Backport 3.12] Fix task manager race condition issues

troopa81:
## Description

Backport #34469 et #35108 in release-3.12


nyalldawson:
Can we hold off merging this till after this weeks round of releases please? It's moderate risk and would be better to merge early in a cycle

troopa81:
OK for postponing merge

PR #35207, opened 2020-03-19
Make a proper registry for scalebar renderers

nyalldawson:
This was all previously very ad hoc, with hardcoded string values everywhere and fragile reliance on translated strings.

Make it nice and formal with a proper registry, which makes it easier to add additional scalebar renderers in future (including via plugins)

Note that while adding the new pure virtual members to QgsScaleBarRenderer would break any existing subclasses, this **isn't** a stable API break because QgsScaleBarRenderer was never previously able to be subclassed in Python and actually used anywhere in QGIS.


src/core/scalebar/qgsscalebarrendererregistry.cpp
```diff
@@ -0,0 +1,105 @@
+/***************************************************************************
+    QgsScaleBarRendererRegistry.cpp
```
nirvn: This needs lower casing.

src/core/scalebar/qgsscalebarrendererregistry.h
```diff
@@ -0,0 +1,90 @@
...
+ ***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+#ifndef QgsScaleBarRendererRegistry_H
```
nirvn: This needs upper casing (and the header's .h file name above needs lower casing)

tests/src/python/test_qgsscalebarrendererregistry.py
```diff
@@ -0,0 +1,69 @@
+# -*- coding: utf-8 -*-
+"""QGIS Unit tests for QgsScaleBarRendererRegistry
+
+.. note:: This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation; either version 2 of the License, or
+(at your option) any later version.
+"""
+__author__ = 'Nyall Dawson'
+__date__ = '6/01/2020'
```
nirvn: 20/03/2020?

nirvn (approved):
Code looks good, only some comments need changing.

PR #35208, opened 2020-03-19
Add @layers, @layer_ids project scope variables

nyalldawson:
which contain lists of map layers and map layers ids for all layers from the current project

This mimics the existing @map_layers, @map_layer_ids, but unlike the @map variants these return ALL project layers, not just those associated with the current context's map settings.

Sponsored by SLYR


nirvn (approved):

she-weeds:
Nyall, is it too late to request a @layer_names variable as well? For example, so that people using aggregate functions can test whether a standard layer name in their expression matches any of the layer names in the project (or is close to it, using fuzzy matching, etc...) and the expression remains valid even if layers are replaced (which results in different layer IDs)? I know this can be done using `array_foreach(@layers,layer_property(@element,'name'))`  but @layer_names would be so much cleaner.

PR #32630, opened 2019-11-04
[FEATURE] add maptip and expression display in expressions

Gustry:
## Description

Expose the display expression and the maptip in expressions.

This is useful if we don't want to copy/paste the maptip in the composer when we generate an atlas.

By default, these are evaluated. But we can choose to not evaluate the expression, so it can be modified on runtime (search and replace, remove first line...).
In this case, we need to use the `eval` function. For the maptip, as it's a template with `[%...%]`, I needed to add `eval_template`. I just thought I can merge in `eval` with a flag `isTemplate=False` by default? Any idea?

## Checklist

- [x] Commit messages are descriptive and explain the rationale for changes
- [x] Commits which fix bugs include `Fixes #11111` at the bottom of the commit message
- [x] I have read the [QGIS Coding Standards](https://docs.qgis.org/testing/en/docs/developers_guide/codingstandards.html) and this PR complies with them
- [x] New unit tests have been added for core changes
- [x] I have run [the `scripts/prepare-commit.sh` script](https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTING.md#contributing-to-qgis) before each commit
- [ ] I have evaluated whether it is appropriate for this PR to be backported, backport requests are left as label or comment

PR #35211, opened 2020-03-20
add maptip, expression display and eval_template expressions

Gustry:
## Description

Expose the display expression and the maptip in expressions.

This is useful if we don't want to copy/paste the maptip in the composer when we generate an atlas.

By default, these are evaluated. But we can choose to not evaluate the expression, so it can be modified on runtime (search and replace, remove first line...).
In this case, we need to use the `eval` function. For the maptip, as it's a template with `[% ... %]`, I needed to add `eval_template`. 

Expressions added:

```
eval_template
display_expression
maptip
```

Previous PR https://github.com/qgis/QGIS/pull/32630 closed by stalebot

### DIFF
--- a/resources/function_help/json/display_expression
+++ b/resources/function_help/json/display_expression
@@ -1,0 +1,39 @@
+{
+        "name": "display_expression",
+	"type": "function",
+        "description": "Returns the display expression for a given feature in a layer. If called with no parameters, it evaluates the current feature. The expression is evaluated by default.",
+        "arguments": [
+                {
+                        "arg": "feature",
+                        "optional": true,
+                        "default": "current feature",
+                        "description": "The feature which should be evaluated."
+		},
+                {
+                        "arg": "layer",
+                        "optional": true,
+                        "default": "current layer",
+                        "description": "The layer (or its id or name)."
+                },
+		{
+			"arg": "evaluate",
+                        "description": "If the expression must be evaluated. If false, the expression will be returned as a string literal only (which could potentially be later evaluated using the 'eval' function).",
+			"optional": true,
+			"default": "true"
+		}
+	],
+        "examples": [
+                {
+                        "expression": "display_expression()",
+                        "returns": "The display expression of the current feature."
+                },
+                {
+                        "expression": "display_expression($currentfeature)",
+                        "returns": "The display expression for a given feature."
+		},
+		{
+                        "expression": "display_expression('a_layer_id', $currentfeature, 'False')",
+                        "returns": "The display expression of the current feature not evaluated."
+		}
+	]
+}
--- a/resources/function_help/json/eval_template
+++ b/resources/function_help/json/eval_template
@@ -1,0 +1,13 @@
+{
+	"name": "eval_template",
+	"type": "function",
+	"description": "Evaluates a template which is passed in a string. Useful to expand dynamic parameters passed as context variables or fields.",
+	"arguments": [{
+		"arg": "template",
+		"description": "a template string"
+	}],
+	"examples": [{
+                "expression": "eval_template('QGIS [% upper(\\\\'rocks\\\\') %]')",
+                "returns": "QGIS ROCKS"
+	}]
+}
--- a/resources/function_help/json/maptip
+++ b/resources/function_help/json/maptip
@@ -1,0 +1,39 @@
+{
+	"name": "maptip",
+	"type": "function",
+        "description": "Returns the maptip for a given feature in a layer. If called with no parameters, it evaluates the current feature. The maptip is evaluated by default.",
+        "arguments": [
+            {
+                    "arg": "feature",
+                    "optional": true,
+                    "default": "current feature",
+                    "description": "The feature which should be evaluated."
+            },
+            {
+                    "arg": "layer",
+                    "optional": true,
+                    "default": "current layer",
+                    "description": "The layer (or its id or name)."
+            },
+            {
+                    "arg": "evaluate",
+                    "description": "If the expression must be evaluated. If false, the expression will be returned as a string literal only (which could potentially be later evaluated using the 'eval_template' function).",
+                    "optional": true,
+                    "default": "true"
+            }
+	],
+        "examples": [
+                {
+                        "expression": "maptip()",
+                        "returns": "The maptip of the current feature."
+		},
+                {
+                        "expression": "maptip($currentFeature)",
+                        "returns": "The maptip of the current feature."
+                },
+		{
+			"expression": "maptip('a_layer_id', $currentFeature, 'False')",
+                        "returns": "The maptip of the current feature not evaluated."
+		}
+	]
+}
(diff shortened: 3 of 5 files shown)


resources/function_help/json/display
```diff
@@ -0,0 +1,39 @@
...
+		},
+                {
+                        "arg": "layer",
+                        "optional": true,
+                        "default": "current layer",
+                        "description": "The layer (or its id or name)."
+                },
+		{
+			"arg": "evaluate",
+                        "description": "If the expression must be evaluated. If False, you should call 'eval'",
```
nyalldawson:
```suggestion
                        "description": "If the expression must be evaluated. If false, the expression will be returned as a string literal only (which could potentially be later evaluated using the 'eval' function) ",
```

resources/function_help/json/eval_template
```diff
@@ -0,0 +1,13 @@
+{
+	"name": "eval_template",
+	"type": "function",
+	"description": "Evaluates a template which is passed in a string. Useful to expand dynamic parameters passed as context variables or fields.",
+	"arguments": [{
+		"arg": "template",
+		"description": "a template string"
+	}],
+	"examples": [{
+		"expression": "eval_template(QGIS [% 'rocks' || '!' %])",
```
nyalldawson: Is this correct? Shouldn't 'QGIS...' be wrapped in `'`?

resources/function_help/json/maptip
```diff
@@ -0,0 +1,39 @@
...
+            },
+            {
+                    "arg": "layer",
+                    "optional": true,
+                    "default": "current layer",
+                    "description": "The layer (or its id or name)."
+            },
+            {
+                    "arg": "evaluate",
+                    "description": "If the expression must be evaluated. If False, you should call 'eval_template'",
```
nyalldawson: same as above

src/core/expression/qgsexpressionfunction.cpp
```diff
@@ -1513,6 +1519,97 @@ static QVariant fcnAttributes( const QVariantList &values, const QgsExpressionCo
...
+      return layer->mapTipTemplate();
+    }
+    else
+    {
+      return layer->displayExpression();
+    }
+  }
+
+  QgsExpressionContext subContext( *context );
+  // subContext.appendScopes( QgsExpressionContextUtils::globalProjectLayerScopes( layer ) );
```
nyalldawson: remove

src/core/expression/qgsexpressionfunction.cpp
```diff
@@ -5903,6 +6000,30 @@ const QList<QgsExpressionFunction *> &QgsExpression::Functions()
...
+      QStringLiteral( "Record and Attributes" ),
+      QString(),
+      false,
+      QSet<QString>()
+    );
+    maptipFunc->setIsStatic( false );
+    functions << maptipFunc;
+
+    QgsStaticExpressionFunction *displayFunc = new QgsStaticExpressionFunction(
+      QStringLiteral( "display" ),
```
nyalldawson: I'm concerned about the use of the generic "display" term here. Could we use "display_expression" or something a bit more explicit instead?

tests/src/core/testqgsexpression.cpp
```diff
@@ -2015,6 +2022,51 @@ class TestQgsExpression: public QObject
...
+      QTest::newRow( "maptip wrong feature" ) << QStringLiteral( "maptip()" ) << QgsFeature() << mPointsLayer << true << QVariant();
+      QTest::newRow( "maptip wrong layer" ) << QStringLiteral( "maptip()" ) << firstFeature << noLayer << true << QVariant();
+      QTest::newRow( "maptip not evaluated" ) << QStringLiteral( "maptip(@layer_id, $currentfeature, False)" ) << firstFeature << mPointsLayer << false << QVariant( "Maptip with class = [% \"Class\" %]" );
+
+      QTest::newRow( "display with 2 params" ) << QStringLiteral( "display(@layer_id, $currentfeature)") << firstFeature << mPointsLayer << false << QVariant( "Display expression with class = ' || Class" );
+      // QTest::newRow( "display without params" ) << QStringLiteral( "display()" ) << firstFeature << mPointsLayer << false << QVariant( "Display expression with class = Biplane" );
+      // QTest::newRow( "maptip without params" ) << QStringLiteral( "maptip()" ) << false << QVariant( "Maptip with class = Biplane" );
+      //QTest::newRow( "maptip" ) << QStringLiteral( "maptip('points', get_feature_by_id('points', '1'))" ) << false << QVariant( "Maptip with class = Biplane" );
```
nyalldawson: what's with this?

src/core/expression/qgsexpressionfunction.cpp
Gustry: Yes, I will change

tests/src/core/testqgsexpression.cpp
Gustry: No, I must enable these tests.
But it seems my tests is not working well, travis fails.
These new expressions are working in QGIS Desktop. But all these commented test and the test L2043 failed. The evaluation context seems wrong. Do you have a quick idea where the error can be.

Gustry:
Thanks for the review @nyalldawson 
I have some issue to make the test working on travis (and my computer).
All expressions are working nicely in QGIS Desktop, but when I copy/paste them in the test.
All commented lines in the test must be in the PR.

tests/src/core/testqgsexpression.cpp
nyalldawson: I think you'll need to add the layer to a project, and then add the project scope to your context

Gustry: Thanks, it should be OK now

Gustry:
Thanks for the review, I have updated the PR.
Travis should be happy.

Gustry:
>  61/615 Test  #61: qgis_fontmarkertest ...............................***Failed    0.70 sec

Unrelated test failure

nyalldawson (approved):

resources/function_help/json/display_expression
```diff
@@ -0,0 +1,39 @@
...
+		}
+	],
+        "examples": [
+                {
+                        "expression": "display_expression()",
+                        "returns": "The display expression of the current feature."
+                },
+                {
+                        "expression": "display_expression($currentfeature)",
+                        "returns": "The display expression for a given feature."
```
DelazJ, line 32: @Gustry this is still for the current feature -- ref maptip example

Gustry: Sorry @DelazJ ? Is there an issue?
I just spotted `$currentFeature` vs `$currentfeature`. I will fix it

DelazJ: Sorry I meant it would be
> returns": "The display expression for the current feature."

Which remains a "given" feature but... Also this is how this example is explained for maptip function

Gustry: Do you mean you prefer something like https://github.com/qgis/QGIS/blob/master/resources/function_help/json/is_selected#L21
I wrote "current feature", when the feature is coming from the scope, not as parameter

PR #35217, opened 2020-03-20
Fix multiple connecting slots in dual view

signedav:
Since the dual view `init()` functionality is used for re-initialization, it should not connect multiple times the same objects. ~So the connection type `UniqueConnection` is used, to avoid that e.g. in Relation Editor Widget the context menu displays the same entries several times...~
So the connections of the member objects only created once are moved to the constructor.

### DIFF
--- a/src/gui/attributetable/qgsdualview.cpp
+++ b/src/gui/attributetable/qgsdualview.cpp
@@ -52,6 +52,12 @@ QgsDualView::QgsDualView( QWidget *parent )
   connect( mFeatureListView, &QgsFeatureListView::aboutToChangeEditSelection, this, &QgsDualView::featureListAboutToChangeEditSelection );
   connect( mFeatureListView, &QgsFeatureListView::currentEditSelectionChanged, this, &QgsDualView::featureListCurrentEditSelectionChanged );
   connect( mFeatureListView, &QgsFeatureListView::currentEditSelectionProgressChanged, this, &QgsDualView::updateEditSelectionProgress );
+  connect( mFeatureListView, &QgsFeatureListView::willShowContextMenu, this, &QgsDualView::widgetWillShowContextMenu );
+
+  connect( mTableView, &QgsAttributeTableView::willShowContextMenu, this, &QgsDualView::viewWillShowContextMenu );
+  mTableView->horizontalHeader()->setContextMenuPolicy( Qt::CustomContextMenu );
+  connect( mTableView->horizontalHeader(), &QHeaderView::customContextMenuRequested, this, &QgsDualView::showViewHeaderMenu );
+  connect( mTableView, &QgsAttributeTableView::columnResized, this, &QgsDualView::tableColumnResized );
 
   mConditionalFormatWidgetStack->hide();
   mConditionalFormatWidget = new QgsFieldConditionalFormatWidget( this );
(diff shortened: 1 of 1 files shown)


src/gui/attributetable/qgsdualview.cpp
```diff
@@ -118,11 +118,11 @@ void QgsDualView::init( QgsVectorLayer *layer, QgsMapCanvas *mapCanvas, const Qg
   mLayer = layer;
   mEditorContext = context;
 
-  connect( mTableView, &QgsAttributeTableView::willShowContextMenu, this, &QgsDualView::viewWillShowContextMenu );
+  connect( mTableView, &QgsAttributeTableView::willShowContextMenu, this, &QgsDualView::viewWillShowContextMenu, Qt::UniqueConnection );
```
troopa81: Wouldn't be better to make the connect in the constructor?

src/gui/attributetable/qgsdualview.cpp
```diff
@@ -146,11 +146,11 @@ void QgsDualView::init( QgsVectorLayer *layer, QgsMapCanvas *mapCanvas, const Qg
     mAttributeEditor->layout()->addWidget( mAttributeForm );
   }
 
-  connect( mAttributeForm, &QgsAttributeForm::widgetValueChanged, this, &QgsDualView::featureFormAttributeChanged );
-  connect( mAttributeForm, &QgsAttributeForm::modeChanged, this, &QgsDualView::formModeChanged );
-  connect( mMasterModel, &QgsAttributeTableModel::modelChanged, mAttributeForm, &QgsAttributeForm::refreshFeature );
-  connect( mAttributeForm, &QgsAttributeForm::filterExpressionSet, this, &QgsDualView::filterExpressionSet );
-  connect( mFilterModel, &QgsAttributeTableFilterModel::sortColumnChanged, this, &QgsDualView::onSortColumnChanged );
+  connect( mAttributeForm, &QgsAttributeForm::widgetValueChanged, this, &QgsDualView::featureFormAttributeChanged, Qt::UniqueConnection );
```
troopa81: _mAttributeForm_ is deleted just above so the connection should be removed, same for _mMasterModel_ deleted in initModels (maybe the connect should be moved in that method)

troopa81 (changes requested):

src/gui/attributetable/qgsdualview.cpp
signedav: Hm... I will go over it again. Thanks.

src/gui/attributetable/qgsdualview.cpp
signedav: I'll check it.

src/gui/attributetable/qgsdualview.cpp
signedav: I agree that it does not need to be unique since the connection are deleted before anyway. I think we cannot move the connections with the `mMasterModel` to the `initModels` since it connects to the `mAttributeForm` created later. (same for `mFilterModel`).

src/gui/attributetable/qgsdualview.cpp
signedav: I moved the connections for `mTableView` and `mFeatureListView` to the constructor. Can you have a look again? Thanks.

src/gui/attributetable/qgsdualview.cpp
troopa81: Right

troopa81 (approved):
It seems OK then, thanks!

signedav:
Thanks @troopa81 
A backport makes sense IMHO, don't you think?

troopa81:
> A backport makes sense IMHO, don't you think?

For sure in 3.12. And I would say yes for 3.10 also.

3nids:
@signedav you'll have to do the backport manually (I think the new backport should work on new (rebased) pull requests)

signedav:
Thanks. Will do it tomo.

PR #35230, opened 2020-03-21
[UI] In labels dialog, change the tab bar for a row a buttons

3nids:
Since the elide mode of the tab bar was set to none, the stack widget was super wide, making it painful to access buttons when use in the docked mode.

before
image

after:
image

### DIFF
--- a/src/gui/qgstextformatwidget.cpp
+++ b/src/gui/qgstextformatwidget.cpp
@@ -504,15 +504,15 @@ void QgsTextFormatWidget::initWidget()
 
   // set button group for stacked widget
   QButtonGroup *tabBtuttonGroup = new QButtonGroup( this );
-  tabBtuttonGroup->addButton( mTextButton, 0 );
-  tabBtuttonGroup->addButton( mFormattingButton, 1 );
-  tabBtuttonGroup->addButton( mBufferButton, 2 );
-  tabBtuttonGroup->addButton( mMaskButton, 3 );
-  tabBtuttonGroup->addButton( mBackgroundButton, 4 );
-  tabBtuttonGroup->addButton( mShadowButton, 5 );
-  tabBtuttonGroup->addButton( mCalloutButton, 6 );
-  tabBtuttonGroup->addButton( mPlacementButton, 7 );
-  tabBtuttonGroup->addButton( mRenderButton, 8 );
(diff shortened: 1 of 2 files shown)


3nids:
now without spacing between the buttons:
image

roya0045:
> now without spacing between the buttons:
> image

Can we just have the button stretch to fill the space. No one likes tiny buttons and this option seems to cause wasted space.

3nids:
and...with slightly bigger icons
image

3nids:
and.........centered
image

nirvn:
Ok, I'm likely -1 on change (to fix an OS X theme issue) as it is likely resulting in a regression whereas the style dock won't be able to reduce its width beyond that of the width of  that row of buttons.

The tab bar handles overflow of tabs on narrow widths by having a scrollable area with a visual indicator.

That said, I'm not at home and haven't checked the changes yet. I'm just guessing it'll be an issue when I review later today :)

nirvn:
Also, since our OS X builds live on the edge of Qt tech, an alternative approach could be to implement a fix within Qt itself.

3nids:
> style dock won't be able to reduce its width beyond that of the width of that row of buttons

what about putting this in a scroll area then?

> The tab bar handles overflow of tabs on narrow widths by having a scrollable area with a visual indicator.

as you can see on first print screen, for me it's not the case and the widget is barely usable at the moment.
So you get a scrollbar on windows/linux, correct? 
But are the tabs as wide as on mac?

roya0045:
On windows is seems fine 
image

nirvn:
@3nids , yep, on linux and windows you get a scroll area, and no, the tabs aren't wide. That said, I did stumble on this issue of overly wide tabs in the past, when improving our night mapping and blend of gray themes. I actually implemented a CSS syntax solution, see: https://github.com/qgis/QGIS/commit/7c50ed9b7e480863a6f46615d77d0e6e0a7247a8#diff-d942df53ebc6fa26dcb816c6c1f09694

I'd suggest an alternative approach here (if it works): add the above-linked stylesheet syntax at the application level when you detect QGIS is launched under OS X, using its OS X-specific Qt style.

3nids:
superseded by #35265

PR #35236, opened 2020-03-22
[RPM] remove python3-qscintilla as requirement

basilrabi:
python3-qscintilla does not exist in fedora 33 which makes the building and installation of rpm to fail.
Also, python3-qscintilla-qt5 is already required and provides the same according
to https://src.fedoraproject.org/rpms/qgis/c/49bfbaf1ea1738e909862a897f6849336a823372?branch=master


basilrabi:
cc @daniviga

m-kuhn (approved):
Looks like a leftover from PyQt4

PR #35233, opened 2020-03-21
Try updating downloading links to OTB 7.1

nyalldawson:

PR #35237, opened 2020-03-22
Re-enable otb tests with stable download link of 7.0

m-kuhn:
Refs https://github.com/qgis/QGIS/pull/35233


nyalldawson (approved):
Much better solution! (Although the breakage looks quite severe with 7.10... not sure what's going on there)

PR #35238, opened 2020-03-22
Add "stepped line" and "hollow" scalebar styles

nyalldawson:
Adds two new scalebar styles (which are available in ArcMap, but previously impossible to reproduce in QGIS)

Stepped line style:

image

Hollow (aka "south african") style:

image

Also includes the api cleanups suggested in https://github.com/qgis/QGIS/pull/35225

Sponsored by SLYR


src/core/layout/qgslayoutitemscalebar.cpp
```diff
@@ -96,21 +93,21 @@ void QgsLayoutItemScaleBar::draw( QgsLayoutItemRenderContext &context )
...
   {
     // compatibility code - ScalebarLineColor and ScalebarLineWidth are deprecated
     QgsExpressionContext expContext = createExpressionContext();
-    std::unique_ptr< QgsFillSymbol > sym( mSettings.fillSymbol2()->clone() );
+    std::unique_ptr< QgsFillSymbol > sym( mSettings.alternateFillSymbol()->clone() );
     Q_NOWARN_DEPRECATED_PUSH
     sym->setColor( mDataDefinedProperties.valueAsColor( QgsLayoutObject::ScalebarFillColor2, expContext, mSettings.fillColor2() ) );
     Q_NOWARN_DEPRECATED_POP
-    mSettings.setFillSymbol2( sym.release() );
+    mSettings.setAlternateFillSymbol( sym.release() );
```
nirvn: Thanks, it reads much nicer now.

src/core/scalebar/qgsnumericscalebarrenderer.cpp
```diff
@@ -33,7 +33,12 @@ QString QgsNumericScaleBarRenderer::visibleName() const
 
 int QgsNumericScaleBarRenderer::sortKey() const
 {
-  return 7;
+  return 9;
```
nirvn: I'm wondering whether it'd be worth changing it to eg. 100; then we'd have visual scale bar 0-99 and non visual bars 100-inf :)

Or we can keep increasing it as need arise, no big deal.

nirvn (approved):
Great!

src/core/scalebar/qgsnumericscalebarrenderer.cpp
nyalldawson: Done

nirvn: Thanks.

agiudiceandrea:
These bug reports https://github.com/qgis/QGIS/issues/38369 and https://github.com/qgis/QGIS/issues/37413 belongs to this PR.

PR #35243, opened 2020-03-22
Fix #32759, Don't zoom out when selecting vertices on the vertex editor

### LINKED ISSUE #32759: Maps zooms out on editing vertex by vertex tool
Canvas magnifier is set to other than 100%. I use vertex tool to manually editing coordinates, filling values in the coordinate table. On every coordinate update canvas zooms out.

Qgis 3.10
Win 7 x64 pro

uclaros:
## Description
Selecting vertices on the vertex editor while magnification is not set to 100% no longer messes with the zoom level.
Also changed to using the `visibleExtent()` since it might be smaller than the `extent()`.
Also added a canvas refresh. If selecting vertices outside the canvas, the view was zoomed but features were not drawn until zoomed or panned.

Fixes #32759

### DIFF
--- a/src/app/vertextool/qgsvertexeditor.cpp
+++ b/src/app/vertextool/qgsvertexeditor.cpp
@@ -433,9 +433,10 @@ void QgsVertexEditor::updateVertexSelection( const QItemSelection &, const QItem
     try
     {
       QgsRectangle transformedBbox = t.transform( *bbox );
-      QgsRectangle canvasExtent = mCanvas->mapSettings().extent();
+      const QgsRectangle canvasExtent = mCanvas->mapSettings().visibleExtent();
       transformedBbox.combineExtentWith( canvasExtent );
-      mCanvas->setExtent( transformedBbox );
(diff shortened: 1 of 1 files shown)


nyalldawson:
Thank you!! This has bugged me for a long time, and assumed the fix was complex. Didn't realise it was a matter of setting the "magnified" argument. 

Definitely some documentation improvements needed here... they are inbound now!

nyalldawson (approved):

uclaros:
@nyalldawson  NP! Why does the magnify argument default to false though? Don't we most of the times want to set the true visible extent on the user's screen?

nyalldawson:
@uclaros yeah I think it's the opposite of what we want...

uclaros:
A quick check did not reveal any other cases of using `setExtent()` and explicitly setting the magnify argument so it's either the default `false` was preffered or magnification was ignored.
Spatial bookmarks and zoom to feature for example ignore current magnification while they should not.
Is changing the default argument considered API break?

PR #35244, opened 2020-03-23
Add numeric formatter "fraction" style

### LINKED ISSUE #21289: Display fractions on scale bar
Author Name: **Spencer Gardner** (@spencerrecneps)
Original Redmine Issue: [13228](https://issues.qgis.org/issues/13228)

---
There should be an option to display labels on the scale bar as fractions. I often display scales at fractions of a mile, which looks cluttered. Instead of 0.250 the scale should show 1/8 and so on, ideally taking advantage of fractional characters in the source font if available.

I'm surprised this hasn't already been raised, but I couldn't see any previous discussion about this in my searching. Apologies if this duplicates another request.

nyalldawson:
This style represents decimal numbers as vulgar fractions, e.g. "3/4" instead of 0.75.

Options include using Unicode superscript and subscript characters for nicer typography, e.g. ¹⁷/₂₃ (this is the default mode, disabling this option uses the "17/23" format). An option also exists for
using dedicated unicode characters for specific fractions (where a unicode character exists), e.g. ½ or ¾

Ultimately this allows for creation of scalebars with fractional representations of distances, e.g. 0 ----- ½ ----- 1 km (instead of 0 ------ 0.5 ------ 1km)

Fixes #21289

Sponsored by SLYR

image


nirvn (approved):

nyalldawson:
Thanks @nirvn!

PR #35247, opened 2020-03-23
Fix "zoom to" actions fail to correctly set canvas extent when canvas has a locked scale

nyalldawson:
In these particular actions, we should ignore the canvas scale lock so that the canvas correctly zooms to the specified extent and matches user expectations

In particular this fixes:
- zooming to bookmarks
- matching canvas extent to layout map extents
- zooming to layer and feature extents
- pushing layout map scales from layout maps to canvas

Also improve dox

Follows up #35243


nirvn (approved):
Nice; I'm not using magnification much because every time I do, I find a little tiny wrong here and there. You're increasing the chance that will change soon! :wink;

PR #35251, opened 2020-03-23
Enable snapping to grid for models in designer

nyalldawson:
This PR implements two "snapping" features in the Processing model designer:

1. Users can enable a new "Enable Snapping" option from the view menu, which will cause all component moving or resizing operations to automatically snap to grids

2. After selecting some components, users can select Edit -> Snap Selected Components to Grid to manually snap just those selected components.

Additionally, I've added a helpful "select all" action for quickly selecting all components in a model.

### DIFF
--- a/python/gui/auto_generated/processing/models/qgsmodelcomponentgraphicitem.sip.in
+++ b/python/gui/auto_generated/processing/models/qgsmodelcomponentgraphicitem.sip.in
@@ -108,11 +108,6 @@ Shows a preview of moving the item from its stored position by ``dx``, ``dy``.
 Sets a new scene ``rect`` for the item.
 %End
 
-    QRectF previewItemRectChange( QRectF rect );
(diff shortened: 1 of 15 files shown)


nirvn:
Cool.

I'm not sure we should switch off snap to grid. I feel that, contrary to the layout designer, the model designer should have it on by default.

nyalldawson:
@nirvn I originally had this thought too, but in practice it just "felt" wrong. Would you mind revisiting this after merge? (It's an easy true->false change)

nirvn:
@nyalldawson , sure. I think for it to work well, we'd need default item size to correlate with grid, ie item width is 8 × grid spacing, height is 2 × grid spacing.

nyalldawson:
Yeah, that's a good point!

alexbruy (approved):

PR #35255, opened 2020-03-23
[Symbology][Feature] Allow centroid fill markers to be clipped within polygon

suricactus:
Allow centroid fill markers to be clipped within polygon or polygon part. 

Added two options to the centroid fill widget:
- **Clip markers to polygon boundary** - works exactly how the `Random points` clip works
- **Clip markers to current part boundary only** - clips the marker symbol only within the current part in case of multipart geometry

### Demo:
Peek 2020-03-23 07-20

### Applications:

The only way to properly create this kind of maps.
Screenshot_20200323_062016

### TODOs:
- add some tests
- add expressions support for centroid fill

### DIFF
--- a/python/core/auto_generated/symbology/qgsfillsymbollayer.sip.in
+++ b/python/core/auto_generated/symbology/qgsfillsymbollayer.sip.in
@@ -2183,6 +2183,8 @@ Caller takes ownership of the returned symbol layer.
 %Docstring
 Sets whether a point is drawn for all parts or only on the biggest part of multi-part features.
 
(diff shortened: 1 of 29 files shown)


suricactus:
Any help with the SIP error will be appreciated, I have no clue how to fix it. Tried to be more explicit about which "Part" struct, didn't work. Obviously I managed to compile on my machine to create the gif, but I cannot anymore.

suricactus:
Found it, I had to define my newly added struct private. Sorry for the noise.

Issue #35183, opened 2020-03-18
[processing] Auto geometry fixer


roya0045:
**Feature description.**
Sometimes when you try to process some data you get the invalid geometry, you never know which layer and you end up running fix geometry on the two inputs just to be safe.

Though this is a hassle. 

We have an option to skip invalid geometry but you end up missing element.

The best option would be something that intercept the invalid geometry and make it valid by fixing it. 

Though with the current codebase this is hard to do as the validation steps is far remote from the part that output the feature. Here is what I can think that might work:

- rework the internal to now only provide true or false on check but to handle a geometry (would require a lot of refactoring).
- output the feature with a flag or trigger to intercept it on the way out and fix it there
- use a signal to intercept the feature and fix it.
- when encountering an error, fix the faulty input and restart automatically with the temporary fixed input.

uclaros:
Keep in mind that *fix geometries* does not always provide the fix the user wants to apply to the data so it should not be applied automatically and silently. For example fixing a self intersection might generate a new feature, while the user would prefer to delete a couple of vertices, and that may cause unexpected results in the processing algorithm.

roya0045:
@uclaros I don't think there can be a magic solution to geometry errors, I simply feel that since fix geometry would be the function that is applied anyway to a geometry when the user receive this error, then doing it within the process is not that bad.
And yes a message/popup would be beneficial. Its simply that to implement this efficiently its not that easy with the current flow.

uclaros:
@roya0045 For me, *Check validity* is the next step when I receive this error on data that I want to have control over, not an automatic fix, so more info on the geometry error would me welcome!

The current setting *Invalid features filtering* has the following options:
- Do not filter (better performance)
- Skip (ignore) features with invalid geometries
- Stop algorithm execution when a geometry is invalid

Maybe there could be a fourth option to do as you propose like *fix invalid geometries and retry*.

roya0045:
@uclaros yes this is what I'm proposing.

nyalldawson:
@roya0045 some rough notes on how you'd do this:

- the guts all happens in qgsfeaturerequest.cpp and qgsvectorlayerfeatureiterator.cpp
[...]

PR #35257, opened 2020-03-23
Per feature source properties (feature limit, invalid geometry handling)

nyalldawson:
Hookup backend API allowing features sources to be limited to a specific maximum number of features, and allowing per-source overriding of the default "invalid geometry" handling behavior


alexbruy (approved):

roya0045:
I was hoping that this PR would help me decide for a better approach in #35183 . Though I'm not sure if this complexities the per feature approach that I was hoping to make or if I should go for the fix-all method.

PR #35265, opened 2020-03-23
[UI] fix tab widget on mac

3nids:
image

superseds #35230

### DIFF
--- a/src/app/qgisappstylesheet.cpp
+++ b/src/app/qgisappstylesheet.cpp
@@ -16,14 +16,16 @@
  *                                                                         *
  ***************************************************************************/
 
+#include <QFont>
+#include <QStyle>
+
 #include "qgisappstylesheet.h"
 #include "qgsapplication.h"
 #include "qgisapp.h"
+#include "qgsproxystyle.h"
 #include "qgslogger.h"
 #include "qgssettings.h"
 
-#include <QFont>
-#include <QStyle>
 
 QgisAppStyleSheet::QgisAppStyleSheet( QObject *parent )
   : QObject( parent )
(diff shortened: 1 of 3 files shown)


src/gui/qgsproxystyle.h
```diff
@@ -57,6 +57,8 @@ class GUI_EXPORT QgsAppStyle : public QProxyStyle
     explicit QgsAppStyle( const QString &base );
     QPixmap generatedIconPixmap( QIcon::Mode iconMode, const QPixmap &pixmap, const QStyleOption *opt ) const override;
 
+    QString baseStyle() const {return mBaseStyle;}
```
nirvn:
```suggestion
    QString baseStyle() const { return mBaseStyle; }
```

src/app/qgisappstylesheet.cpp
```diff
@@ -106,18 +108,24 @@ void QgisAppStyleSheet::buildStyleSheet( const QMap<QString, QVariant> &opts )
...
   // Specifically setting a custom close button temporarily works around issue.
-  // TODO: Remove when regression is fixed (Qt 5.9.3 or 5.10?); though hard to tell,
-  //       since we are overriding the default close button image now.
   if ( mMacStyle )
   {
     ss += QLatin1String( "QTabBar::close-button{ image: url(:/images/themes/default/mIconCloseTab.svg); }" );
     ss += QLatin1String( "QTabBar::close-button:hover{ image: url(:/images/themes/default/mIconCloseTabHover.svg); }" );
   }
+#endif
+  if ( mMacStyle )
```
nirvn: @3nids , have you tried non-default themes (night mapping and/or blend of gray) to make sure this unconditional stylesheet addition doesn't mess those up. 99.9% it won't, but I don't have a mac at-hand to confirm.

nirvn (approved):
Much better approach, glad it works out fine.

nirvn:
@3nids , BTW, would be good to backport this to LTR (and 3.12).

src/app/qgisappstylesheet.cpp
3nids: yes it does look good

PR #35272, opened 2020-03-23
take another try to get a valid relation in the relation reference widget

3nids:
If the relation name is invalid, try to get the relation using the field name.

I had the case with a project where relations were apparently badly configured (bad relation name).

For the referenced layer, the relation editor was saying the relation was invalid although it was correctly configured in the layer properties.
BUT the referencing layer had its referencing field configured as a relation reference widget but with a bad relation ID.

This fixes this scenario.

Who is the current master of the relation reference widget? @m-kuhn @signedav @pblottiere ?


m-kuhn (approved):
Looks like a reasonable change to me

Issue #35107, opened 2020-03-16
Qgis server 3.10.3 issues


loic74:
With qgis server 3.10.3 and 3.10.3 projects (on Ubuntu bionic), Qgis server crashes after the first request :

`
2020/03/16 11:39:04 [error] 1277#1277: *33 upstream prematurely closed connection while reading response header from upstream, client: 192.168.0.11, server: 192.168.0.11, request: "GET /qgisserver/qgis_mapserv.fcgi?map=%2Fhome%2Fsig%2Fwebmapping%2Ddata%2Fprojets%2Feaux%2Feau%5Fassainissement%2Eqgs&request=GetCapabilities&service=WMS&version=1%2E3%2E0 HTTP/1.1", upstream: "fastcgi://unix:/var/run/qgisserver.socket:", host: "192.168.0.11"
`

`
2020/03/16 11:39:04 [error] 1277#1277: *36 connect() to unix:/var/run/qgisserver.socket failed (111: Connection refused) while connecting to upstream, client: 192.168.0.11, server: 192.168.0.11, request: "GET /qgisserver/qgis_mapserv.fcgi?map=%2Fhome%2Fsig%2Fwebmapping%2Ddata%2Fprojets%2Feaux%2Feau%5Fassainissement%2Eqgs&request=GetCapabilities&service=WMS HTTP/1.1", upstream: "fastcgi://unix:/var/run/qgisserver.socket:", host: "192.168.0.11"
`

Qgis server 3.10.3 does not crash with 3.4.13 projects. But, in this case, getprint shows the wrong layers if any group layer is specified.

Thanks for your help.

rldhont:
About crashes, how did you configure QGIS Server ?
* apache with mod_fcgi
* nginx with spawn-fcgi
* nginx with fcgiwrap
 
what are the environment variables you have set ?

loic74:
nginx with spawn-fcgi

jgrocha:
@loic74 Do you have QGIS Server log enabled? Can you check the output of QGIS Server log? Maybe with the log enabled, we can see where the QQGIS Server crashes.

I usually set:
```
QGIS_SERVER_LOG_FILE=/var/log/qgis/qgisserver.log
QGIS_SERVER_LOG_LEVEL=1
```

Check the available log options from https://docs.qgis.org/testing/en/docs/user_manual/working_with_ogc/server/config.html

gisgogo:
[...]

Issue #35203, opened 2020-03-19
Print composer - bug when using theme and undisplayed group


simogeo:
Having layers in group (unchecked so invisible) and using Theme in Print composer cause unwated displayed layers.

**See example below (invisible group) :**

![image](https://user-images.githubusercontent.com/154323/77095513-4d77ba80-6a0e-11ea-8846-2a57dbd88a4b.png)

**But layers are visible on print composer when map is set to specified Theme - see result :** 

![image](https://user-images.githubusercontent.com/154323/77095568-62ece480-6a0e-11ea-9678-6ec810f48fba.png)

**How to Reproduce**

1. Create a Group in layers manager
2. Add layers in group
3. Make group unckeched
4. Add theme
5. Create print composer 
6. Add map set content based on created Theme
7. See bug 

**QGIS and OS versions**

QGIS version | 3.12.0-București | QGIS code revision | 409bcf8eaa
-- | -- | -- | --
Compiled against Qt | 5.12.4 | Running against Qt | 5.12.4
Compiled against GDAL/OGR | 2.4.2 | Running against GDAL/OGR | 2.4.2
Compiled against GEOS | 3.7.2-CAPI-1.11.2 | Running against GEOS | 3.7.2-CAPI-1.11.2 b55d2125
Compiled against SQLite | 3.29.0 | Running against SQLite | 3.29.0
PostgreSQL Client Version | 11.7 (Ubuntu 11.7-0ubuntu0.19.10.1) | SpatiaLite Version | 4.3.0a
QWT Version | 6.1.4 | QScintilla2 Version | 2.10.4
Compiled against PROJ | 5.2.0 | Running against PROJ | Rel. 5.2.0, September 15th, 2018
OS Version | Ubuntu 19.10
Active python plugins | VectorBender; AnotherDXF2Shape; mask; Qgis2threejs; GeoCoding; vector_tiles_reader; ORStools; qconsolidate3; shapetools; Hqgis; DataPlotly; french_locator_filter; SRTM-Downloader; Beeline; OSMtools; QNEAT3; loadthemall; db_manager; MetaSearch; processing

roya0045:
pretty sure there were similar tickets regarding themes and group or complex symbology. I might be wrong.

nyalldawson:
This is fixed already

simogeo:
Sorry for duplicate ! And thanks for the good work

PR #35277, opened 2020-03-23
[Tests][Server] WMS GetPrint group

rldhont:
## Description

Add WMS GetPrint group test.


rldhont:
Try to fix #35107

rldhont:
I have ran the test on branch release-3_4 and it passes. So it is a regression.

rldhont:
WMS_GetPrint_Group_Rendered

The rendered image seems to display all layers and not only the layers in the group

nyalldawson:
@rldhont sounds like https://github.com/qgis/QGIS/issues/35203 ?

rldhont:
@nyalldawson how #35203 has been fixed ?

nyalldawson:
Yes. But it requires projects to be re-saved to apply the fix

rldhont:
@nyalldawson you mean the fix is in the QGIS XML Project ?

nyalldawson:
@rldhont 
> you mean the fix is in the QGIS XML Project ?

yes - the issue was caused in projects saved in a limited window of QGIS versions, after #33099 was applied and before https://github.com/qgis/QGIS/pull/34794 was merged. The only solution is to load the project post https://github.com/qgis/QGIS/pull/34794, fix the theme, and resave.

rldhont:
@nyalldawson There are no themes in the test project, and I found something was missing in the code.

rldhont:
@dmarteau @elpaso @pblottiere can you help me to fix this issue. I have probably found where the issue comes from but my fix segfault.

rldhont:
@pblottiere I have found the commit that create the regression https://github.com/qgis/QGIS/commit/5c874ca554309ed6b8cbead84d79bc9b6756ac60

This commit removes `QgsRenderer::stylizedLayers` in which groups nickname is transformed into a `QgsMapLayer` list. Can you take a look at my fix ?

pblottiere:
Hi @rldhont,

> Can you take a look at my fix ?

The rationale looks good. Waiting for Travis then :crossed_fingers:

rldhont:
@pblottiere thanks for the review

rldhont:
Yeah travis is happy

PR #35286, opened 2020-03-24
[Symbology][Feature] Allow centroid fill markers to be clipped within polygon

suricactus:
Allow centroid fill markers to be clipped within polygon or polygon part. 

Added two options to the centroid fill widget:
- **Clip markers to polygon boundary** - works exactly how the `Random points` clip works
- **Clip markers to current part boundary only** - clips the marker symbol only within the current part in case of multipart geometry

### Demo:
Peek 2020-03-23 07-20

### Applications:

The only way to properly create this kind of maps.
Screenshot_20200323_062016

My apologies for the second PR, I messed up the previous branch with accidental merge instead of rebase. Git is life, git is hard.


src/core/symbology/qgsfillsymbollayer.cpp
```diff
@@ -3509,41 +3513,118 @@ void QgsCentroidFillSymbolLayer::stopRender( QgsSymbolRenderContext &context )
 
 void QgsCentroidFillSymbolLayer::renderPolygon( const QPolygonF &points, QList<QPolygonF> *rings, QgsSymbolRenderContext &context )
 {
-  if ( !mPointOnAllParts )
+  Part part;
```
nyalldawson, line 3516: There's a bug here (also in the random marker fill symbol layer) which causes issues in multi layer symbols. I can't merge this until that bug is fixed, because this will break any existing multi-layer symbols using the centroid fill (the severity of the issue with random marker fills isn't so bad, since the bug was present ever since the random marker symbol layer type was introduced... hence no regressions possible!)

suricactus: Should I close the PR then? Or just standby until the bug is fixed? I am afraid I am not qualified to track the bug (yet), so the only option for me is to wait or to change the implementation. Is there an open issue for the bug you mentioned?

nyalldawson: Standby for now -- what you have here looks good, it's deeper architectural changes required

suricactus:
ping

nyalldawson:
I'd like to unfreeze this for 3.14 -- the code has been approved and ready to merge for months, yet was blocked by an unrelated bug. I've a fix inbound for that bug, after which I think it's only fair on @suricactus if we merge for 3.14.

nyalldawson:
@suricactus can you please rebase off master now? The blocker has been fixed so we are good to go...

suricactus:
Great! I will rebase and check if all good a bit later today.

nyalldawson:
Thanks! One thing which would be great to include is a unit test where the centroid fill with the new option is "sandwiched" between two semi - transparent simple fills

suricactus:
Not sure if I understand you correctly, you mean something like this:
image

nyalldawson:
Yes but with another semi transparent fill on top of the marker as well. So simple - marker - simple. A sandwich!

suricactus:
Clear! Will have a sandwich! 🥪

suricactus:
@nyalldawson sorry for the delay, it took me some time to remember how to work with the symbology API. Hope all is good. Thanks!

nyalldawson (approved):

PR #35291, opened 2020-03-24
Improve offline editing layer detection and code

m-kuhn:
Offline editing currently relies on

- "layername" being the exactly the first part of the URL
- `QgsDataSourceUri::param( "{file_path_to_gpkg}|layername" )` to return the layer name

This solidifies this behavior with the regex approach from the gpkg provider (still not perfect but *much* more stable).

Also comes with some general modernization of the code.

### DIFF
--- a/src/core/qgsofflineediting.cpp
+++ b/src/core/qgsofflineediting.cpp
@@ -38,6 +38,8 @@
 #include "qgsogrutils.h"
 #include "qgsvectorfilewriter.h"
 #include "qgsvectorlayer.h"
+#include "qgsproviderregistry.h"
+#include "qgsprovidermetadata.h"
 
 #include <QDir>
 #include <QDomDocument>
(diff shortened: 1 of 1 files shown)


src/core/qgsofflineediting.cpp
```diff
@@ -1492,7 +1492,15 @@ void QgsOfflineEditing::committedFeaturesAdded( const QString &qgisLayerId, cons
   }
   else
   {
-    tableName = uri.param( offlinePath + "|layername" );
+    if ( dataSourceString.indexOf( QLatin1String( "|layername=" ) ) != -1 )
```
nyalldawson: Why not use QgsDataProviderRegistry::decodeUri instead?

m-kuhn: I was looking for something like `layer->dataProvider().staticMetaData.decodeUri()` but didn't find any.

nyalldawson: Nope, you do it through the registry using the layer's provider key

m-kuhn: Ok. If this is the preferred approach I can do that.

nyalldawson: Yep, it's definitely preferred. If nothing else it's covered by unit tests...

signedav (approved):

m-kuhn:
Manually tested and works

src/core/qgsofflineediting.cpp
m-kuhn: Done :+1:

PR #35292, opened 2020-03-24
Expose per-feature-source advanced options for processing inputs

nyalldawson:
Among a ton of cleanups, this PR exposes per-feature-source advanced options, such as:
- The ability to limit the number of features read from the source. (Useful in model development, you can easily test run models using a small subset of an input layer) 
- The ability to override on a per-input basis the global processing option for how to handle invalid geometries. Now you can safely leave the default processing setting at the conservative "Stop algorithm execution when a geometry is invalid" setting while easily temporarily overriding this for one particular input!

Peek 2020-03-24 16-00


nyalldawson:
Other exciting stuff here includes:
- more Python -> c++ conversion
- more unit tests
- setting up the basis for processing widgets to utilise stacked panels (more use of this is inbound!)
- a cleanup of the "iterate over input" option, including a new hi-dpi friendly svg icon

alexbruy (approved):
Wow! Very nice feature

agiudiceandrea:
It looks like setting the CreateIndividualOutputPerInputFeature flag doesn't work. See: https://github.com/qgis/QGIS/issues/56610.

PR #35295, opened 2020-03-24
Extract guts of QgsDataSourceSelectDialog to a QWidget

nyalldawson:
...so that this widget logic can be used outside of a dialog without code duplication


elpaso (approved):
LGTM!

PR #35296, opened 2020-03-24
By default, do *not* force all fields to become strings when exporting to XLS/XLSX/ODS/CSV

nirvn:
## Description

This PR reverts a change made in 2016 (https://github.com/qgis/QGIS/commit/8ee697bf2dcb7cb6329d2d6aefc009350878375d) which transformed all field types to string when saving as XLS/XLSX/ODS/CSV. That default behavior leads to data type loss and other regressions. For e.g., when you save a temporary memory output (say from a processing algorithm) as XLSX, all values become strings, and field _names_ are also lost.

@rouault , is there a technical limitation on GDAL's side that led you to force all fields to be strings for these drivers? If not, I'm begging for a reversal of default behavior here.

### DIFF
--- a/src/gui/ogr/qgsvectorlayersaveasdialog.cpp
+++ b/src/gui/ogr/qgsvectorlayersaveasdialog.cpp
@@ -424,7 +424,6 @@ void QgsVectorLayerSaveAsDialog::mFormatComboBox_currentIndexChanged( int idx )
   {
     if ( mOptions & Fields )
       mAttributesSelection->setVisible( true );
-    fieldsAsDisplayedValues = ( sFormat == QLatin1String( "CSV" ) || sFormat == QLatin1String( "XLS" ) || sFormat == QLatin1String( "XLSX" ) || sFormat == QLatin1String( "ODS" ) );
   }
 
   // Show symbology options only for some formats


rouault:
>  is there a technical limitation on GDAL's side that led you to force all fields to be strings for these drivers?

No, I believe it was per request of @andreasneumann (if my memories are not wrong...)

nirvn:
Okay, I'm going to merge this now to plug a default behavior that leads to data loss.

@andreasneumann , do you remember why you'd have asked for this?

andreasneumann:
> @andreasneumann , do you remember why you'd have asked for this?

I didn't ask all values to be strings on export, but I wanted to have the ability to export presentation values instead of the internal values, e.g. a date should be exported with the format defined in the widget or for the widget types relation reference widget, value map widget, and value relation widget one would get the human-readable values and no meaningless internal numeric codes.

@nirvn - I don't know exactly what you reverted here, but please reconsider. Spreadsheets are often presentation formats and not databases, where one needs the internal codes.

nirvn:
@andreasneumann , okay. The good news is, you still have that ability :) the only change I did was the revert the behavior that silently switched on auto-stringification of all field types (including integers simple integers etc) into strings.

If you need your spreadsheet exports to use the string representation for any given field, you can be determinate about it and check the relevant checkboxes.

andreasneumann:
I guess that sounds fine. Most people probably need representation values and not internal values in spreadsheets, but I agree that numbers should survive as numbers, when there are no widgets defined that have other presentation values.

nyalldawson:
I think a good compromise would be to auto select only certain fields with widget types of relations & value maps, and leave others unselected (by default)

PR #35300, opened 2020-03-24
fix fields properties table background on dark mode (macos)

3nids:
before:
image

after:
image

### DIFF
--- a/src/gui/vector/qgssourcefieldsproperties.cpp
+++ b/src/gui/vector/qgssourcefieldsproperties.cpp
@@ -20,6 +20,9 @@
 #include "qgsproject.h"
 #include "qgsapplication.h"
 #include "qgsexpressioncontextutils.h"
+#include "qgsgui.h"
+#include "qgsnative.h"
+
 
 QgsSourceFieldsProperties::QgsSourceFieldsProperties( QgsVectorLayer *layer, QWidget *parent )
   : QWidget( parent )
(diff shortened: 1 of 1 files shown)


nirvn:
@3nids , this makes it nice on dark themes (really nice actually :) ) but there's a fundamental problem that still needs addressing: we really shouldn't set a background color without setting the foreground color too.

For e.g., here's how night mapping looks like:
Screenshot from 2020-03-24 17-37-40

Now it's not as bad as the text color is mid gray, but it hits the same issue you had. 

Long story short, can you add a setForeground( color ) to items too for dark and non-dark themes?

3nids:
> Long story short, can you add a setForeground( color ) to items too for dark and non-dark themes?

I don't think it's going to work: the Night mapping theme is not detected as a dark mode (dark mode is OS not app).
Correct?

nirvn:
@3nids , it's not going to switch to dark mode color palette, but at least it could have a black foreground (ie text) against those very bright colors.

3nids:
> but at least it could have a black foreground (ie text) against those very bright colors.

it's just that I would prefer to minimize the hacking around the default styles. I was even wondering about removing the lines background for the dark theme. 
I would rather propose to modify the foreground using qss.

If it's OK for you I'd like to merge as is.

nirvn (approved):
Merge at will. The other issue can be dealt with separately.

For the record, stylesheet doesn't mix well with manually setting background n foreground.

3nids:
manually backport to 3.12 in #35300, backport to 3.10 is skipped (I would argue that macos users are more likely to use last version rather than LTR)

PR #35303, opened 2020-03-24
fix group background color of groups in expression builder tree

### LINKED ISSUE #35299: the field calculator panel in the default Macintosh style has white font on a white background that does not allow you to see the functions.
the field calculator panel in the default Macintosh style has white font on a white background that does not allow you to see the functions.

**QGIS 3.12/3.10 and OSx Mojave**

**Additional context**

Schermata 2020-03-24 alle 11 22 40

3nids:
fixes #35299

before 
image

after
image

### DIFF
--- a/src/gui/qgsexpressionbuilderwidget.cpp
+++ b/src/gui/qgsexpressionbuilderwidget.cpp
@@ -13,6 +13,22 @@
  *                                                                         *
  ***************************************************************************/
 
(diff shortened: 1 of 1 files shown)


3nids:
with single transparent color (good catch @nirvn ) 

image

image

nirvn (approved):

PR #35307, opened 2020-03-24
Check that plugin dir exists before linking to it

strk:
Reduces frustration for plugin developers willing to test
the plugin..

I can think of more improvements but I'm not sure how many people
use the tool and thus how much backward compatibility should be
retained.


nyalldawson (approved):
LGTM

PR #35319, opened 2020-03-25
Interface and API for unified development/debugging tools

nyalldawson:
Adds interface hooks to allow plugins (and c++) to register "development/debugging" tools

These tools appear in a new Development/Debugging Tools dock, and provide a unified handling and place for these tools in the UI.

The intention here is that specialised development/debugging tools will be moved to this common interface, e.g.
- network logger
- first aid style Python local variables inspector
- startup time debugging tools
- layer load and rendering time debugging tools
... ?

Here's an example demo showing a "Python locals" dev tool:

image

Based heavily off @NathanW2's style dock work!

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -103,6 +103,7 @@ IF(WITH_APIDOC)
       ${CMAKE_SOURCE_DIR}/src/gui/auth
       ${CMAKE_SOURCE_DIR}/src/gui/attributetable
       ${CMAKE_SOURCE_DIR}/src/gui/callouts
+      ${CMAKE_SOURCE_DIR}/src/gui/devtools
       ${CMAKE_SOURCE_DIR}/src/gui/editorwidgets
       ${CMAKE_SOURCE_DIR}/src/gui/editorwidgets/core
       ${CMAKE_SOURCE_DIR}/src/gui/effects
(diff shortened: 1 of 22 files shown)


nyalldawson:
@nathanw2 would love your thoughts

NathanW2:
Nice work.  This will be a good feature when building a plugin

nirvn:
Really cool stuff @nyalldawson .

alexbruy:
Nice addition!

ymoisan:
@nyalldawson quick question : I thought network traffic visible in the development/debugging tools panel would also show /vsicurl/http:/ requests e.g. to a COG.  I guess QGIS can't trap HTTP get range requests because there aren't any to the extent that it is GDAL underneath that does the heavy lifting and QGIS just shows the result.  

I managed to see what I wanted, the actual range of bytes asked, using Wireshark but it would be nice if that were accessible in the development/debugging tools panel.  Or am I missing something ?

nyalldawson:
You're not missing anything. We need to add hooks into gdal to allow custom callbacks on network activity, but that's a non trivial piece of work

ymoisan:
Right.  Thanx Nyall.

drnextgis:
@nyalldawson is there an open issue to keep track the status of that feature?

PR #35321, opened 2020-03-25
Port processing extent parameter to new api, many other improvements

nyalldawson:
This PR ultimately ports the extent parameter to the new c++ api, but along the way it:

- refactors the existing gui widget "QgsExtentGroupBox" to move the guts out to "QgsExtentWidget" (allowing it to be used in places where a group box is inappropriate/looks bad). It also adds a new "condensed" mode for the widget, because the default appearance is very space heavy.
- The new parameter widget utilises this common QgsExtentWidget, so that we remove all the duplicate code which was previously used by the old python wrapper
- allows drag and drop of project map layers from the layer tree to a QgsExtentWidget to auto-set the extent to match
- allows QgsExtentWidget to be optionally nullable
- Allows QgsGeometry values to be set for extent parameter values. This is done to aid use of extent parameters with the "precalculated value" mode in models -- you can now use all the expression geometry functions to calculate a geometry and the bounding box of this is used for the parameter value in the child algorithm


nirvn:
@nyalldawson , can you post a screenshot of the condensed UI?

nyalldawson:
> can you post a screenshot of the condensed UI?

sorry, I've switched branches -- but it's just a line edit with toolbutton next to it

alexbruy (approved):

nyalldawson:
Unrelated test failure

PR #35322, opened 2020-03-25
[ui] Further improve source fields properties colors to play well with any light/gray/dark themes

nirvn:
## Description

@3nids , a small follow up to your source fields properties coloring fix. Instead of using different colors when we detect a native dark mode, IMHO it's much better to use semi-transparent colors to work under all conditions.

This is how night mapping looked before vs. this PR applied:
image


3nids (approved):

3nids:
backport will work on 3.12 but not on 3.10 (I did not backport original PR to 3.10)

PR #35324, opened 2020-03-25
Allow running algorithms directly on database (and other non-disk) sources without loading into projects first

nyalldawson:
This change allows users to directly browse to non disk-based layer sources for any processing feature source inputs. It allows these inputs to be taken direct from postgres, sql server, oracle, wfs, afs, etc layers directly without having to first load them into a project!

We take full advantage of the QGIS browser to enable this:

Peek 2020-03-25 16-37

### DIFF
--- a/python/core/auto_generated/processing/qgsprocessingutils.sip.in
+++ b/python/core/auto_generated/processing/qgsprocessingutils.sip.in
@@ -92,6 +92,28 @@ value.
 .. seealso:: :py:func:`compatibleVectorLayers`
 %End
 
+    static QString encodeProviderKeyAndUri( const QString &providerKey, const QString &uri );
+%Docstring
+Encodes a provider key and layer ``uri`` to a single string, for use with
+decodeProviderKeyAndUri()
+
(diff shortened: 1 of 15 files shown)


nirvn:
Oh...my...god!

NathanW2:
Hot dam!

alexbruy (approved):
Awesome!

Jean-Roc:
Great ! Could this be extended to relation forms ? This would allow the use of tables for relationnal values without loading them in the project

nyalldawson:
@Jean-Roc 

> Great ! Could this be extended to relation forms ? This would allow the use of tables for relationnal values without loading them in the project

Behind the scenes it's quite a different approach vs relations. Processing always "re-loads" layers from disk sources before running the alg, so this was a way of extending this and exposing the ability in the ui to set non-project layer paths directly. Relation layers however are required to be constantly available and are very closely tied to their parent layer. That's not to say what you want isn't possible, it's just that it couldn't take advantage of anything here...

PR #35332, opened 2020-03-25
when adding new container in form designer, preselect the current group as parent

3nids:
ezgif-3-045ada41ec57


3nids:
unrelated test failure

signedav (approved):
Looks good

PR #35333, opened 2020-03-25
Make QgsNewDatabaseTableNameWidget a QgsPanelWidget, tweak toolbar size

nyalldawson:
Minor tweaks


nirvn (approved):
I was going to do the icon sizing bit as soon as all that magic stops landing :smile:

PR #35335, opened 2020-03-25
[ogr] Mark OpenStreetMap file extensions (.osm, .pbf) as featuring layers

nirvn:
## Description

Because thanks to @nyalldawson , we'll be able to run processing against those OpenStreetMap files without having to add layers onto our projects to process them. Woupidou.

Backporting to 3.12 and 3.10 because it's also just useful to have the browser panel acknowledge those formats are layered.

Peek 2020-03-25 15-53


nyalldawson (approved):

nirvn:
Unrelated travis failure (postgis raster), merging.

PR #35361, opened 2020-03-26
Fix use of attributes() function in field calculator

### LINKED ISSUE #35358: Cannot update field using Field Calculator with attributes() expression
When using an expression that has `attributes()` in it, although the expression resolves correctly in the preview in Field Calculator, whenever I hit OK to update fields, it always returns `NULL`. 

* I am able to update fields using Field Calculator with any other expression using strings, `attribute()`, or map functions. 

* This happens regardless of whether I'm updating an existing field, creating a new field, or creating a virtual field; whether the layer is open for edits or not.

* I have tested this in 3.10.4 and 3.12.0 as well as 3.12.1. Installed using the Network Installer.

The screenshots attached show the simple example of `array_to_string(map_akeys(attributes()))`. which should return a list of field names, and do in Field Calculator preview (red line), but not when applied as an update.

att_issue_1
att_issue_2
att_issue_3

nyalldawson:
Fixes #35358


m-kuhn (approved):

m-kuhn:
Did I see this correctly that it disables the isStatic flag? (the diff is hard to read)

nyalldawson:
Yep the default is to be static -- this changes explicilty sets it as non-static.

Previously it was getting compiled away to nothing in the prepare step whenever it was being used for multi-feature evaluations

m-kuhn:
Yep, deleting this flag totally does make sense.

PR #35364, opened 2020-03-26
Allowing saving outputs direct to more database formats (and other nice stuff)

nyalldawson:
This PR allows users to save processing vector outputs direct to more database formats.

Previously outputs could only be written direct to postgres databases. With this change, this functionality has been made more flexible and now supports direct writing to any database provider which implements the connections API (currently postgres, geopackage, spatialite and sql server)

Ultimately this exposes the new ability to directly save outputs to SQL Server or Spatialite databases (alongside the previous GPKG+Postgres options which already existed)

(As soon as oracle, db2, ... have the connections API implemented we'll instantly gain direct write support for those too!)

We do this via a nice inline version of @elpaso's new "new database table name" widget.

Peek 2020-03-26 13-57

Other nice stuff includes:
- you can now drag and drop file or folders from explorer or the qgis browser to output parameters in order to easily overwrite these files (or save to the folder), matching the behavior possible with inputs
- you can actually write scripts which output direct to ANY qgis data provider (including oracle and db2)... it's just that these don't get exposed in the UI because they don't yet support the connections API

### DIFF
--- a/python/gui/auto_generated/processing/qgsprocessingoutputdestinationwidget.sip.in
+++ b/python/gui/auto_generated/processing/qgsprocessingoutputdestinationwidget.sip.in
@@ -65,6 +65,14 @@ Emitted whenever the "skip output" option is toggled in the widget.
 %Docstring
 Emitted whenever the destination value is changed in the widget.
 %End
+  protected:
+
+    virtual void dragEnterEvent( QDragEnterEvent *event );
+
+    virtual void dragLeaveEvent( QDragLeaveEvent *event );
+
+    virtual void dropEvent( QDropEvent *event );
+
 
 };
 
--- a/python/gui/auto_generated/qgsnewdatabasetablenamewidget.sip.in
+++ b/python/gui/auto_generated/qgsnewdatabasetablenamewidget.sip.in
@@ -46,6 +46,13 @@ Constructs a new QgsNewDatabaseTableNameWidget
 :param parent: optional parent for this widget
 %End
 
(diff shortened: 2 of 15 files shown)


alexbruy (approved):

Jean-Roc:
@nyalldawson could you point me to an explanation of this connection api ? This feature added to the previous direct read would be great to work with, as we are using Oracle we could have a look at the possibility of adding this to a future order.

elpaso:
@Jean-Roc the connections API is a new API for data providers that exposes to the QGIS core a set of abstracted methods to manage stored connections (i.e. the various connections to DBs that you can see in the browser) and to perform common operations on DBs.

It was developed mainly by myself with the help and peer-review of @nyalldawson, initially as a grant proposal with implementations for postgres and GPKG, afterwards it was added to MSSQL and spatialite, we are still missing Oracle and MySQL (not to mention DB2, if anybody cares about it).

You can find some more details in the first implementation PR:
https://github.com/qgis/QGIS/pull/31190

and here:
https://www.itopen.it/qgis-abstract-connections-api/

The main classes:
https://qgis.org/api/classQgsAbstractProviderConnection.html
https://qgis.org/api/classQgsAbstractDatabaseProviderConnection.html#details

The final goal of this API is to replace a huge amount of very old and barely tested provider-specific implementations (in C++ and Python) with a solid and tested core implementation and to expose these methods to Python in an unified and consistent way.

Jean-Roc:
thanks Alessandro !

PR #35367, opened 2020-03-26
[postgres] Do not overwrite detected type with requested type

### LINKED ISSUE #34629: Postgis mixed geometry data table causes QGIS > 3.8 issues as more recent QGIS versions see all records instead of those of the right type
**Describe the bug**

In QGIS more recent than 3.8 a postgis database that has (multi)points, lines and (multi)polygons into one table causes issues as QGIS sees all rows in the database in each of the layers.
In QGIS 3.8 (and earlier), this problem was not present. QGIS 3.10 and 3.12 both have the issue.

FWIW: the database is not used by QGIS alone, so splitting the table into 3 tables it is simply not an option (need for unique identifiers and other tables not even used in QGIS that refer to this data)

This causes problems when viewing the table data in QGIS, when rendering labels ( a layer will generate labels for data of the wrong geometry type), and when exporting shape files for that layer.

**How to Reproduce**

Have a postgis database table containing in a single table multipoint, point, multipolygon, 
multilinestring, and linestring data.

e.g.:
```
=> \d eo
                                          Table "our.eo"
     Column      |          Type           | Collation | Nullable |            Default             
-----------------+-------------------------+-----------+----------+--------------------------------
 id              | integer                 |           | not null | nextval('eo_id_seq'::regclass)
 mapname         | character varying(50)   |           |          | 
 geom            | geometry(Geometry,4326) |           |          | 
[....]
```

```
=> select count(*), geometrytype(geom) from eo group by geometrytype(geom);
 count |  geometrytype   
-------+-----------------
     3 | 
  8674 | MULTIPOINT
   140 | MULTIPOLYGON
   229 | POINT
     3 | MULTILINESTRING
    25 | LINESTRING
(6 rows)
```

Create a new project in QGIS 3.10 , connect to the postgis database to add a vector layer and see the database there 3 times (as expected), once for (multi)points, once for lines and once for (multi)polygons:

image

Add either the 3 layers to the project (one by one or all in one go doesn't matter)

Click open Attributes table for each of the layers and see the counts:

E.g. for the points one:
image
compare that 9075 to the rows in the database above, and confirm data for known polygon data is included as well.

e.g. for the polygon layer:
image

Doing the same in QGIS 3.12 yields exactly the same results as in 3.10

Doing the same in QGIS 3.8 works without a problem:
polygons:
Screenshot 2020-02-24 at 17 55 01
lines:
Screenshot 2020-02-24 at 17 55 30
points:
Screenshot 2020-02-24 at 17 55 45

Once these rows get seen in multiple layers, at least the following additional problems get triggered:
- Labels get shown for elements that have the wrong Geometry type. (to 
Green labels on the points, red labels on the polygons, placement set so they don't overlap:
zoomed in to a single point, no overlapping polygon anywhere near:
image

- exporting to shapefiles fails with errors
e.g. exporting to an ESRI shapefile of the points layer gives
image

Instructions don't work on a mac.

image

Tested versions:
QGIS 3.8.2  - macOS Catalina 10.15.3 - works as expected
QGIS 3.10.2  - macOS Catalina 10.15.3 - bug
QGIS 3.12.0 - macOS Catalina 10.15.3 - bug

**Additional context**

m-kuhn:
Since https://github.com/qgis/QGIS/pull/32292 the "detected geometry type" was overwritten by the "requested geometry type", especially in "GEOMETRY" type situations.

The postgis provider contains logic to [add to the WHERE clause a filter on geometry type if "detected type" and "requested type" are equal](https://github.com/qgis/QGIS/blob/final-3_12_0/src/providers/postgres/qgspostgresfeatureiterator.cpp#L486-L489). This check no longer worked ever since.

Fix #34629

### DIFF
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -3659,8 +3659,7 @@ bool QgsPostgresProvider::getGeometryDetails()
 
     if ( result.PQntuples() == 1 )
     {
-      QString dt = result.PQgetvalue( 0, 0 );
-      detectedType = dt;
+      detectedType = result.PQgetvalue( 0, 0 );
 
       QString dim = result.PQgetvalue( 0, 2 );
       if ( dim == QLatin1String( "3" ) && !detectedType.endsWith( 'M' ) )
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -3643,7 +3643,7 @@ bool QgsPostgresProvider::getGeometryDetails()
     }
   }
 
-  QString detectedType = mRequestedGeomType == QgsWkbTypes::Unknown ? QString() : QgsPostgresConn::postgisWkbTypeName( mRequestedGeomType );
+  QString detectedType;
   QString detectedSrid = mRequestedSrid;
   if ( !schemaName.isEmpty() )
   {
@@ -3660,7 +3660,7 @@ bool QgsPostgresProvider::getGeometryDetails()
     if ( result.PQntuples() == 1 )
     {
       QString dt = result.PQgetvalue( 0, 0 );
-      if ( dt != "GEOMETRY" ) detectedType = dt;
+      detectedType = dt;
 
       QString dim = result.PQgetvalue( 0, 2 );
       if ( dim == QLatin1String( "3" ) && !detectedType.endsWith( 'M' ) )
--- a/tests/src/python/test_qgsproviderconnection_postgres.py
+++ b/tests/src/python/test_qgsproviderconnection_postgres.py
@@ -69,6 +69,13 @@ class TestPyQgsProviderConnectionPostgres(unittest.TestCase, TestPyQgsProviderCo
         rl = QgsRasterLayer(conn.tableUri('qgis_test', 'Raster1'), 'r1', 'postgresraster')
         self.assertTrue(rl.isValid())
 
+    def test_postgis_geometry_filter(self):
+        """Make sure the postgres provider only returns one matching geometry record and no polygons etc."""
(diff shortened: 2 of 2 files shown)


strk (approved):
I love the code simplification and the test. Not sure the PR title really matches the code change (is it the contrary ? NOT overriding requested with detected?) but I'll trust you on this.

If you have requested my review it is probably because I was the last one touching this code but I'm pretty sure I tried as much as possible to retain what was already happening in previous versions so please take a look to see if older versions also need the change.

Thank you !

m-kuhn:
> Not sure the PR title really matches the code change (is it the contrary ? NOT overriding requested with detected?) but I'll trust you on this.

I think there is a "not" in the title :)

> If you have requested my review it is probably because I was the last one touching this code

Mainly because it partially reverts #32292 (which also is in 3.10 and 3.12 --> yes I'd opt for a backport)

strk:
> > Not sure the PR title really matches the code change (is it the contrary ? NOT overriding requested with detected?) but I'll trust you on this.
> 
> I think there is a "not" in the title :)

The title tells me you want to avoid that the DETECTED type gets overwritten by the REQUESTED type. Is this what the code does ?

strk:
> Mainly because it partially reverts #32292 (which also is in 3.10 and 3.12 --> yes I'd opt for a backport)

Uhm, and indeed it looks like I did change that behaviour:
https://github.com/qgis/QGIS/pull/32292/files#diff-374a3d3ebe5430cec252d7f2c7d5f457R3532
Sorry for that, and thanks again for writing more tests !

m-kuhn:
Thanks for the review @strk

m-kuhn:
> The title tells me you want to avoid that the DETECTED type gets overwritten by the REQUESTED type. Is this what the code does ?

Previous behavior: The detected type (db) was overwritten by the requested type (uri) if the detected type was geometry
New behavior: The detected type (db) is preserved in the provider next the requested type (uri) so we can check at any time, if the two match (or if we have to assume that other types will be present in the db and need to filter them by type if we want to only get matching features)

PR #35384, opened 2020-03-26
[processing] fixed detection of grass folder in Windows

volaya:
## Description

Looks like installation of grass in windows, both with the standalone installer and with OSGEO4W, has changed how the grass folder is named. The method that detects it fails now, and grass algorithms cannot be run from Processing.

This PR fixes it

### DIFF
--- a/python/plugins/processing/algs/grass7/Grass7Utils.py
+++ b/python/plugins/processing/algs/grass7/Grass7Utils.py
@@ -221,7 +221,8 @@ class Grass7Utils:
                 else:
                     testfolder = os.path.join(os.path.dirname(QgsApplication.prefixPath()), 'grass')
                     if os.path.isdir(testfolder):
-                        grassfolders = sorted([f for f in os.listdir(testfolder) if f.startswith("grass-7.") and os.path.isdir(os.path.join(testfolder, f))], reverse=True, key=lambda x: [int(v) for v in x[len("grass-"):].split('.') if v != 'svn'])
+                        grassfolders = sorted([f for f in os.listdir(testfolder) if f.startswith("grass") and os.path.isdir(os.path.join(testfolder, f))], 
+                                                reverse=True, key=lambda x: [int(v) for v in x[len("grass"):] if v.isnumeric()])
                         if grassfolders:
                             folder = os.path.join(testfolder, grassfolders[0])
             elif isMac():


gioman:
> ##
> 
> Looks like installation of grass in windows, both with the standalone installer and with OSGEO4W, has changed how the grass folder is named. The method that detects it fails now, and grass algorithms cannot be run from Processing.

@volaya what QGIS versions are affected?

volaya:
My installation of 3.4 already has a folder named "grass78", but I don't know when we started naming grass folders like that. Current 3.13 also uses that schema (the code searches for grass-x.x)

Maybe @jef-n can provide more detail on this

jef-n:
@volaya GRASS renamed the directory.  QGIS relies on GISBASE to be set (which is the case if you use the "with GRASS" shortcut - that also allows the GRASS plugin to load).

volaya:
> @volaya GRASS renamed the directory. QGIS relies on GISBASE to be set (which is the case if you use the "with GRASS" shortcut - that also allows the GRASS plugin to load).

I don't know if i'ts an issue in my machine or in the installers, but GISBASE is not set, even if running with the "with GRASS" shortcut. The GRASS plugin fails to load. It happens in both the standalone installer and the OSGEO4W one

In any case, I think this PR is needed, because it fixes the code use when GISBASE is *not* defined. If it is defined, the code fixed in this PR will be ignored.

arnaud-morvan:
Same problem here.

I just tested on QGIS 12 from standalone installer with no "not core" plugin loaded.

When I try to run GRASS algorithm, I get this message:
```
This algorithm cannot be run :-( 
The specified GRASS 7 folder "C:\PROGRA~1\QGIS3~1.12\bin\bin" does not contain a valid set of GRASS 7 modules. Please, go to the Processing settings dialog, and check that the GRASS 7 folder is correctly configured
```

But I do not find any setting for GRASS folder.

In Python console, I've tried to get the GISBASE:
```
import os 
os.environ["GISBASE"]
  File "C:\PROGRA~1\QGIS3~1.12\apps\Python37\lib\code.py", line 90, in runcode
    exec(code, self.locals)
  File "<input>", line 1, in <module>
  File "C:\PROGRA~1\QGIS3~1.12\apps\Python37\lib\os.py", line 678, in __getitem__
    raise KeyError(key) from None
KeyError: 'GISBASE'
```

GRASS is located at: `C:\Program Files\QGIS 3.12\apps\grass\grass78`

@volaya: could you fix the indentation for the CI to be happy ?

gioman:
@volaya reopening?

PR #35388, opened 2020-03-26
Refresh WMS-T layers when static temporal range is updated

Samweli:
Enables wmst layers to be refreshed when user has updated the layer's temporal range through editing layer static temporal range in the raster layer properties source tab.

This PR also contains ui updates on raster layer properties temporal tab and source tab.


nyalldawson:
@samweli looking good!

Can you cherry-pick the commits I added to https://github.com/nyalldawson/QGIS/commits/Samweli-fix_temporal_wmst_layer_refresh  please?

After that, the last change needed is to move the layer's temporal source to the data source uri and the QgsRasterDataProviderTemporalCapabilities::setEnableTime and QgsRasterDataProviderTemporalCapabilities::isTimeEnabled also need to be moved to the layer's uri.

Samweli:
> @Samweli looking good!
> 
> Can you cherry-pick the commits I added to https://github.com/nyalldawson/QGIS/commits/Samweli-fix_temporal_wmst_layer_refresh please?
> 
> After that, the last change needed is to move the layer's temporal source to the data source uri and the QgsRasterDataProviderTemporalCapabilities::setEnableTime and QgsRasterDataProviderTemporalCapabilities::isTimeEnabled also need to be moved to the layer's uri.

Alright

nyalldawson (approved):

nyalldawson:
Thanks @Samweli !

timlinux:
Thanks @nyalldawson and @Samweli !

src/gui/raster/qgsrasterlayerproperties.cpp
```diff
@@ -1146,6 +1159,168 @@ void QgsRasterLayerProperties::apply()
...
 
+void QgsRasterLayerProperties::updateSourceStaticTime()
+{
+  if ( mWmstGroup->isEnabled() &&
+       mRasterLayer->dataProvider() &&
+       mRasterLayer->dataProvider()->temporalCapabilities()->hasTemporalCapabilities() )
+  {
+    QgsDataSourceUri uri;
+    QString uriString = mRasterLayer->dataProvider()->dataSourceUri();
+    uri.setEncodedUri( uriString );
```
elpaso, line 1170: This doesn't work and it's a bug: you should just use `QgsDataSourceUri uri { mRasterLayer->dataProvider()->dataSourceUri() };`
Setting an encoded URI from a normal URI string representation leads to wrong results.

elpaso: Please check your other usages of `setEncodedUri()`

elpaso: @Samweli sorry to push, but this is blocking my development of temporal support for PG raster provider, can you fix if **now** or should I do that?

Samweli: @elapso I did try to use `QgsDataSourceUri uri ( mRasterLayer->dataProvider()->dataSourceUri() );` but it was giving wrong results instead! I believe this is because the QgsDataSourceUri constuctor does not consider URI from wms provider, as URI from wms provider uses '&' to separate params.

Samweli: Here in this constructor https://github.com/qgis/QGIS/blob/eef58e6e4fd25c07ff582ad0cbc7bdabaafdbbb9/src/core/qgsdatasourceuri.cpp#L35

elpaso: yes, sorry, I meant just `mRasterLayer->dataProvider()->uri()`, did you try that?

`QgsDataSourceUri uri { mRasterLayer->dataProvider()->uri() };`

you make a copy of the existing uri and you can change params as you like.

elpaso: @Samweli this works for me, can you please check it out?  https://github.com/qgis/QGIS/pull/35511/files#diff-148e681e9188c937e373574247e3dd0fR1170

Samweli: I have just tried it now, it also result into the same issue. It does not parse correctly the params from the provider uri.

elpaso: do you have a project/WMS URI I can test with?

Samweli: I think this only happens for the wms provider uri. I did look around at how the currently wms code address this and I found this https://github.com/qgis/QGIS/blob/9c857ecb94dc35842234351c481896130b0551ba/src/providers/wms/qgswmsdataitems.cpp#L61, which also use `setEncodeUri()` as a workaround.

elpaso: anyway, we need a solution that works with **any** provider, not just WMS.

Samweli: > do you have a project/WMS URI I can test with?

Yes, https://geo.weather.gc.ca/geomet?layer=GDPS.ETA_TT

Samweli: > anyway, we need a solution that works with **any** provider, not just WMS.

For now, may be we can add a check if the provider is not WMS, then to use the preferred way of initializing the uri.

elpaso: That would be my last resource: we already have too many  provider-specific checks in core, it's just throwing some more dust under the carpet. But don't worry, I can fix this.

Samweli: Okay, thanks @elpaso

PR #35396, opened 2020-03-27
Inbuilt network logging tool

nyalldawson:
This tool, which is available from the new F12 "dev tools" panel, is a native port of @rduivenvoorde's network logger plugin.

It shows a list of ongoing and completed network requests, along with a whole load of useful detail like request and reply status, headers, errors, SSL configuration errors, timeouts, cache status,
etc.

Also has loads of polish and useful capabilities, such as the ability to filter requests by URL substrings and status, and you can right-click requests to open the URL in a browser or copy them as a cURL command.

Full credit to @rduivenvoorde for all the heavy lifting here!

Why do we want this as a native tool instead of a plugin? Well:
- It's extremely useful, so is nice to have available out-of-the-box
- By moving it to c++, we can startup the logging process much earlier than a plugin gets the opportunity to. This allows the log to include additional information, such as requests fired off by other plugins which occur before the network logger plugin has been loaded (also other in-built requests, e.g. news feed)
- Performance has been an ongoing struggle with the Python plugin. While this is about as close as a 1:1 port as you can get, the performance issues have all been magically solved. The classes have some very intensive list operations, which is likely the cause of the poor performance under Python.

Peek 2020-03-27 11-39


nyalldawson:
Ping @timlinux

src/app/qgisapp.cpp
```diff
@@ -1690,6 +1699,9 @@ QgisApp::~QgisApp()
   // shouldn't be needed, but from this stage on, we don't want/need ANY map canvas refreshes to take place
   mFreezeCount = 1000000;
 
+  unregisterDevToolFactory( mNetworkLoggerWidgetFactory.get() );
```
m-kuhn: <hint you didn't ask for>
  Wouldn't that be better a custom deleter
</hint you didn't ask for>

src/app/devtools/networklogger/qgsnetworkloggerpanelwidget.cpp
```diff
@@ -0,0 +1,144 @@
...
+  connect( mActionClear, &QAction::triggered, mLogger, &QgsNetworkLogger::clear );
+  connect( mActionRecord, &QAction::toggled, this, [ = ]( bool enabled )
+  {
+    QgsSettings().setValue( QStringLiteral( "logNetworkRequests" ), enabled, QgsSettings::App );
+    mLogger->enableLogging( enabled );
+  } );
+
+  connect( mLogger, &QAbstractItemModel::rowsInserted, this, [ = ]
+  {
+    // we may want to make this optional?
```
m-kuhn: connect ( mTreeView, &TreeView::scrolled, this, [] () {
  if ( mTreeView->scrollLocation == Bottom )
    mAutoScroll = true;
  else
    mAutoScroll = false;
} );

src/app/qgisapp.cpp
nyalldawson: Possibly... but I want to control the order of destruction here... so, maybe?

m-kuhn:
Nice work !

src/app/devtools/networklogger/qgsnetworkloggerpanelwidget.cpp
nyalldawson: i can't find a reference to scrollLocation?

src/app/qgisapp.cpp
m-kuhn: My first thought was that a std::unique_ptr is quite an advanced tool for the level of control you want to have here. The second one that if you use it, you should connect unregister to it's deleter. The third one that it's not really an issue ;)

src/app/devtools/networklogger/qgsnetworkloggerpanelwidget.cpp
m-kuhn: Sorry

```cpp
connect( mTreeView->verticalScrollBar(), &QAbstractSlider::sliderMoved, this, [this] ( int value ) {
  if ( value == mTreeView->verticalScrollBar()->maximum() )
    mAutoScroll = true;
  else
    mAutoScroll = false;
} );
```

elpaso:
Very nice! 

Sorry to be the party breaker here, bu I'm honestly thinking if this is a feature that belongs to the core. 

I mean that this is a dream tool for developers and to debug connections but I don't now how much such a tool will be useful for the average user (whatever it means), it seemed to me a feature that should have better remained a plugin.

But I guess it's already decided, so let it be.

nyalldawson:
@elpaso did you see my "faq" in the PR? I've outlined reasons there why this move was warranted...

elpaso:
@nyalldawson yes I did. But I'm still in doubts about how this feature will be useful for the general case.

If this is a feature that is going to be used by a tiny fraction of users in a tiny fraction of their usage time (just in case of developers and/or in case of connection issues) I'm not sure it pays off.

Plus, there are other networking monitoring tools that can be used.

On a more general note, we do have inside QGIS a lot of code that is probably used by 1 user in one million and just once a year or so. I'm not saying this is the case but I'm sure you get what I mean.

But of course, since I am (as a developer) one of those users that will greatly benefit of this feature I should probably better shut up and join the party.

src/app/qgisapp.cpp
nyalldawson: Fixed -- I couldn't get templatey goodness working direct with unique_ptr (can't find a clean way to specialize make_unique to apply custom logic in the constructor) so I've just gone with a custom RAII class instead.

src/app/devtools/networklogger/qgsnetworkloggerpanelwidget.cpp
nyalldawson: Thanks!

nyalldawson:
@elpaso I understand your concerns here. Some more factors which I'm hoping can alleviate them:

- this port was done with @rduivenvoorde's blessing and support
- the code is all very self-contained. It's a handful of simple classes in app alone, which are almost 100% isolated from all the rest of the app code (there's just a tiny change in qgisapp.cpp/.h required for registering the provider), so the maintenance requirements should be extremely low
- by default this is completely hidden from users. There's no user-visible actions or UI changes associated with this work. The **only** way to see the dev tools panel currently is by pressing the 
un-advertised "F12" shortcut

I'm firmly of the belief that the more we support QGIS plugin developers and power users, and the more tools with give this (admittedly very small) percentage of our users is a strong benefit to the whole community. The easier we make plugin development the more likely we'll see more, better quality plugins, and (fingers crossed!) more developers ultimately transitioning from plugin development to core development and maintenance. While this particular tool isn't strictly relating to plugin development, I do see it as the first step of a full set of nicely integrated out-of-the-box developer tools. I'd love to see things like first aid and the plugin reloader moved from 3rd party plugins to completely integrated tools. I'd also love to see the PyQGIS documentation (and cookbook?) added as tabs in the dev tool panel as another way of making PyQGIS development as easy as possible. 

That said, I do hear your concerns, and would love to know if others share them.

nirvn:
Tested the branch, working really well, nice to see the sluggishness gone.

I'm a big +1 on this one. Much like the python console, this is not for all users, but very important tool to have for those interested in digging in.

This is very much in line with browsers like firefox and chrome offering a set of dev tools as part of their browsers.

Can't wait to see what else we can add there.

timlinux:
CC @tomkralidis

timlinux:
Funded by Meteorological Service of Canada

@nyalldawson did you figure out any clever way to log GDAL network requests too?

NathanW2:
Good feature!  

I have had to debug a number of user issues when it comes to external web requests from inside QGIS and the question normally starts with "can I see web requests that QGIS is calling" to which the answer is "........... let's do a screen share so we can set up Fiddler" at that point things normally take a turn for the worse.  It's not hard but adds an extra step which isn't nice. add IT restrictions, other installed proxies and you are in for a world of pain. 

Having these tools will be a great help when people need to debug plugins they are building and I don't see any major downsides to having this kind of thing in core. 

It's a bit like web developer tools, most people don't even know they are there but not sure we could live without them now.

+1 on a merge for this one.

nyalldawson:
Thanks for the feedback @NathanW2 

> I have had to debug a number of user issues when it comes to external web requests from inside QGIS and the question normally starts with "can I see web requests that QGIS is calling" to which the answer is "........... let's do a screen share so we can set up Fiddler" at that point things normally take a turn for the worse. It's not hard but adds an extra step which isn't nice. add IT restrictions, other installed proxies and you are in for a world of pain.

That makes me think -- a "save log to file" option would be REALLY handy for this kind of thing!

nirvn:
@nyalldawson , you mean a "save log to file and send to Nathan's email" button? :wink:

NathanW2:
> Thanks for the feedback @NathanW2
> 
> > I have had to debug a number of user issues when it comes to external web requests from inside QGIS and the question normally starts with "can I see web requests that QGIS is calling" to which the answer is "........... let's do a screen share so we can set up Fiddler" at that point things normally take a turn for the worse. It's not hard but adds an extra step which isn't nice. add IT restrictions, other installed proxies and you are in for a world of pain.
> 
> That makes me think -- a "save log to file" option would be REALLY handy for this kind of thing!

Yes, it would just be mindful of data leaks here.  Maybe look at what other tools do before we implement that kind of thing.  It might have to remove stuff on the export.

nirvn:
@NathanW2 , @nyalldawson , a message box warning of possible sensitive content when saving log to file should suffice.

timlinux:
+1 from me to merge this too. I think it would be more broadly useful than just developers - anyone who is trying to figure out why the W*S connections aren't working as expected (or even just what requests are being made) is going to benefit. And as @nyalldawson mentions, it fixes various issues with the python implementation.

rduivenvoorde:
Thanks Nyall for porting! 

Just build this PR, and looks nice ;-)
I like the idea of saving to a file to.
Other things on my wish list:
- copy url (mainly usefull for the GET requests)
- or maybe instead of saving to file: "copy this request/response + the next 10". In combi with the filter functionality that would suffice for most server issues investigations?

When this is merged I will create a new version  of the python version which will show a message informing the user, when QGIS version >= 3.14, that the user should remove the plugin and use this core functionality.

nyalldawson:
Thanks @rduivenvoorde ,

> copy url (mainly usefull for the GET requests)

Yeah, good idea -- I'll push that after merge

Let's merge, seems like there's sufficient interest here to warrant out-of-the-box inclusion

tomkralidis:
Thanks for the merge of this functionality.  In our case this greatly enables user testing and reporting of our APIs -- great work!

PR #35400, opened 2020-03-27
in QgsFeatureFilterModel init the display expression when setting the layer

3nids:


m-kuhn:
Can you provide a description or link to an issue?

3nids:
When using a QgsFeatureListComboBox, I expect that the display expression is automatically set whenever I set the layer

m-kuhn:
Can you move it up before the call to `reload()` and the signal?

3nids:
> Can you move it up before the call to reload() and the signal?

there is already a reload (but it's a timer) in the setDisplayExpression.
Also, I would expect that the layer change signal is emitted before the display expression change signal.

m-kuhn (approved):
ok

PR #35401, opened 2020-03-27
Improve join dialog layout

m-kuhn:
Makes the group box in "Editable join layer" use the whole width

## Before
Screenshot from 2020-03-27 07-50-47

## After
Screenshot from 2020-03-27 07-57-07

### DIFF
--- a/src/ui/qgsjoindialogbase.ui
+++ b/src/ui/qgsjoindialogbase.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>505</width>
-    <height>487</height>
+    <height>576</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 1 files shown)


nirvn:
While you're cleaning up this part of our house, can you regularize the word casing of labels in this dialog? :)

m-kuhn:
"All Title Case" or "The other way around"?

suricactus:
"The other way around", it seems this is the convention within QGIS. Can also add a little bit of fresh air for these labels with `<layout class="QGridLayout" name="gridLayout" columnstretch="1,3">`?

m-kuhn:
Group box titles seem to be title case elsewhere. But those here look like checkboxes, so let's go "the other way round".

image

m-kuhn:
Screenshot from 2020-03-27 11-01-15

m-kuhn:
Also:

Does anyone know what the "virtual" part of "in virtual memory" refers to?
Could we just shorten that to "Cache join layer"? I think the meaning would not change but readability improve.

nirvn (approved):
I'd drop virtual and have "cache... in memory"

PR #35402, opened 2020-03-27
Port map layer widget wrapper to new c++ api

nyalldawson:
This is a test run, after it's merged I'll use the same template and port the other similar map layer parameters (feature source, raster layer, vector layer, etc)


alexbruy (approved):
Looking good.

Issue #35403, opened 2020-03-27
Vector joins do not work with cache and update


m-kuhn:
If a vector join is added and:

 - cache is activated
 - and editable is activated (update)

the values are only updated on the next edit just after saving the joined layer.

The video below demonstrates this with a joined layer used to store label positions.

![Peek 2020-03-27 09-13](https://user-images.githubusercontent.com/588407/77735670-44d54480-700b-11ea-9129-55672a773143.gif)

PR #35404, opened 2020-03-27
[joins] Disable caching if editing is enabled

m-kuhn:
This disables caching of joins if editing is enabled.

Gui fix for https://github.com/qgis/QGIS/issues/35403

Peek 2020-03-27 10-58


m-kuhn:
Does anyone know what the "virtual" part of "in virtual memory" refers to?
Could we just shorten that to "Cache join layer"? I think the meaning would not change but readability improve.

PR #35407, opened 2020-03-27
MacOS CI 

PeterPetrik:
Due to the update of QGIS MacOS all-in-one bundle to GDAL3, we decided to create alternative way to build all dependencies for the bundler that is not dependent on the homebrew/conda/macports. 

The development of the new system started on https://github.com/qgis/QGIS-Mac-Packager/tree/master/qgis_deps

The new system requires QT from official open-source installer in specific version and few basic homebrew utilities (flex, bison, cmake), for list see [this link](https://github.com/qgis/QGIS-Mac-Packager/blob/qgis-deps-0.1.0/scripts/install_brew_dev_packages.bash)

The new system then builds all the requires FOSS and OSGeo dependencies, starting from python (and modules like PyQt5, h5py, ...), and others (sqlite, pg, scintilla, qwt, gdal). The following steps are add missing libraries to the deps package and update bundler too to use these deps.

To try new system, see this link for [instructions](https://github.com/qgis/QGIS-Mac-Packager/blob/qgis-deps-0.1.0/qgis_deps/README.md). (note: I haven't tested with full homebrew-osgeo4mac suite installed)

The PR introduces the new variable `QGIS_MAC_DEPS_DIR` and fixes few CMake's find packages to look in the folder for all deps.

Also it sets CI to run build of QGIS for official PR in this repo. 

For running tests it would probably need some work to fix failures and so on.


PeterPetrik:
@nirvn it is building on my fork: see for example https://github.com/PeterPetrik/QGIS/runs/539739361?check_suite_focus=true

PeterPetrik:
looks like unrelated travis linux test in postgis test... can someone restart it please?

Spartacus1:
Peter, feel free to send me copies for testing. 
Best Regards

CMakeLists.txt
```diff
@@ -105,6 +105,17 @@ ENDIF (FORCE_STATIC_PROVIDERS)
 SET (WITH_CORE TRUE CACHE BOOL "Determines whether QGIS core should be built.")
 MARK_AS_ADVANCED(WITH_CORE)
 
+IF (APPLE)
+  # QGIS custom dependencies common location
+  # they can be downloaded from https://qgis.org/downloads/macos/qgis-deps
+  # and unpacked to any location
+  SET (QGIS_MAC_DEPS_DIR "" CACHE PATH "Path to QGIS-Mac-Packager custom dependencies directory")
```
3nids:
```suggestion
  SET (MAC_DEPS_DIR "" CACHE PATH "Path to QGIS-Mac-Packager custom dependencies directory")
```
so we group everything MAC related at the same place?
or MAC_QGIS_DEPS_DIR ?

.github/workflows/macos-build.yml
```diff
@@ -0,0 +1,57 @@
...
+#    branches:
+#      - master
+
+env:
+  QT_VERSION: 5.14.1
+  QGIS_DEPS_VERSION: 0.1.0
+
+jobs:
+  build:
+    # if: github.repository == 'qgis/QGIS'
```
3nids:
```suggestion
    if: github.repository == 'qgis/QGIS'
```

3nids (approved):
For what I am concerned, it's all good.
For the cmake part, I do not feel perfectly shaped to review.

CMakeLists.txt
PeterPetrik: It is consistent with existing QGIS_MACAPP_* variables we have.

3nids: let's keep it as is as it's more consistent

3nids:
> do you know how to set multiple cores build?

if you use make just use -j2, with ninja nothing to specify

Spartacus1:
Just Tested the latest NB and the raster processing problem is still there. A simple hillshade, for instance, returns:


dyld: Symbol not found: _iconv
Referenced from: /usr/lib/libarchive.2.dylib
Expected in: /Applications/QGIS3.13.app/Contents/MacOS/lib/libiconv.2.dylib
in /usr/lib/libarchive.2.dylib


Best regards

Spartacus1:
Latest build from yesterday is working great but, as in the last ones, only with the "libiconv.2.dylib medicine".  Is GDAL 3 being used?

PeterPetrik:
@Spartacus1 not the nightlies are still on GDAL2. When I will finish the new build system and bundler I will ping you for testing, but probably later next week if everything goes well

Spartacus1:
No problem Peter. I'll be paying attention and eager to test. Despite the already well known and annoying problem,. the NB seems to be very solid.

PR #35408, opened 2020-03-27
Server project settings, add 'expanded' attribute

mhugent:
Additionally to 'visible' and 'mutuallyExclusive', this PR adds 'expanded' to the GetProjectSettings response of the server. Like this, a web client can see which layer tree elements are expanded / collapsed and present it in the web map layer list the same way


Gustry:
Thanks for the addition.
@mhugent Is-it expected all these GPKG in the PR?

mhugent:
@Gustry I also noticed it and I'm already in the process of reverting them.
Seems they automatically get changed once someone runs the unit tests

mhugent:
Done, test gpkg files are reverted

PR #35412, opened 2020-03-27
[BUG][3D] fix 3D crash with measure line

vcloarec:
Recent works on 3D measure line seems to lead to crash with mesh terrain.
Adding several guards solves the issue.
Also fix unwanted widget opening (identify and measurement) when tools are not activate.
@ismailsunni : maybe a more proper fix or do you think that does the job?

To replicate the crash:
- load a raster or mesh layer
- open a 3D map view
- open the 3D config widget
- close the 3D config widget
- click on visible entity
--> crash

With settings:
- load a raster or mesh layer
- open a 3D map view
- open in the menu : Settings->Options
- close the options window
--> crash

### DIFF
--- a/src/app/3d/qgs3dmaptoolidentify.cpp
+++ b/src/app/3d/qgs3dmaptoolidentify.cpp
@@ -87,6 +87,7 @@ void Qgs3DMapToolIdentify::activate()
   }
 
   mCanvas->scene()->registerPickHandler( mPickHandler.get() );
+  mIsActivate = true;
 }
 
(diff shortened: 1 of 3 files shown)


saberraz:
I can confirm the crash...it is indeed very annoying.

ismailsunni:
Hi @vcloarec 
Thanks for the fix. I think it should be enough. Sorry, I didn't test for those two use cases and my 3d measure line tool is always activated before updating something.
Thanks!

wonder-sk (approved):

PR #35413, opened 2020-03-27
 FIX #33874 Turn on/off ALL selected layers with "Space" button

### LINKED ISSUE #33874: Turn on/off ALL selected layers with "Space" button
**Feature description.**
It would be convenient to turn on/off **all** selected layers by pressing the space bar.

In the layers panel, you can select and deselect a single layer using the "Space" button, but if you select multiple layers, pressing the Space bar turns on / off only one of them.

You can enable / disable selected layers using the “Manage Map Theme” and even create a keyboard shortcut for this, but you need to assign two keys (1 - turn on, 2 - turn off) to perform this action, which is inconvenient.

**Additional context**
Current behavior while pressing "Space" while few layers are selected:

GIS SE question about this issue:
https://gis.stackexchange.com/questions/347754/how-to-turn-on-off-all-selected-layers-in-qgis-3

suricactus:
Toggles the visibility of the currently selected layers or groups:
Peek 2020-03-27 23-32

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -2651,6 +2651,7 @@ void QgisApp::createActions()
   connect( mActionShowSelectedLayers, &QAction::triggered, this, &QgisApp::showSelectedLayers );
   connect( mActionHideSelectedLayers, &QAction::triggered, this, &QgisApp::hideSelectedLayers );
   connect( mActionToggleSelectedLayers, &QAction::triggered, this, &QgisApp::toggleSelectedLayers );
+  connect( mActionToggleSelectedLayersIndependently, &QAction::triggered, this, &QgisApp::toggleSelectedLayersIndependently );
   connect( mActionHideDeselectedLayers, &QAction::triggered, this, &QgisApp::hideDeselectedLayers );
 
   // Plugin Menu Items
@@ -7561,9 +7562,28 @@ void QgisApp::toggleSelectedLayers()
   QgsDebugMsg( QStringLiteral( "toggling selected layers!" ) );
 
   const auto constSelectedNodes = mLayerTreeView->selectedNodes();
-  for ( QgsLayerTreeNode *node : constSelectedNodes )
+  if ( ! constSelectedNodes.isEmpty() )
   {
-    node->setItemVisibilityChecked( ! node->isVisible() );
+    bool isFirstNodeChecked = constSelectedNodes[0]->itemVisibilityChecked();
+    for ( QgsLayerTreeNode *node : constSelectedNodes )
+    {
+      node->setItemVisibilityChecked( ! isFirstNodeChecked );
+    }
+  }
+}
+
+//reimplements method from base (gui) class
+void QgisApp::toggleSelectedLayersIndependently()
+{
+  QgsDebugMsg( QStringLiteral( "toggling selected layers independently!" ) );
+
+  const auto constSelectedNodes = mLayerTreeView->selectedNodes();
+  if ( ! constSelectedNodes.isEmpty() )
+  {
+    for ( QgsLayerTreeNode *node : constSelectedNodes )
+    {
+      node->setItemVisibilityChecked( ! node->itemVisibilityChecked() );
+    }
   }
 }
 
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -2650,6 +2650,7 @@ void QgisApp::createActions()
   connect( mActionHideAllLayers, &QAction::triggered, this, &QgisApp::hideAllLayers );
   connect( mActionShowSelectedLayers, &QAction::triggered, this, &QgisApp::showSelectedLayers );
   connect( mActionHideSelectedLayers, &QAction::triggered, this, &QgisApp::hideSelectedLayers );
+  connect( mActionToggleSelectedLayers, &QAction::triggered, this, &QgisApp::toggleSelectedLayers );
   connect( mActionHideDeselectedLayers, &QAction::triggered, this, &QgisApp::hideDeselectedLayers );
 
   // Plugin Menu Items
@@ -7551,6 +7552,18 @@ void QgisApp::hideSelectedLayers()
   for ( QgsLayerTreeNode *node : constSelectedNodes )
   {
     node->setItemVisibilityChecked( false );
+  }
+}
+
+//reimplements method from base (gui) class
+void QgisApp::toggleSelectedLayers()
+{
+  QgsDebugMsg( QStringLiteral( "toggling selected layers!" ) );
+
+  const auto constSelectedNodes = mLayerTreeView->selectedNodes();
+  for ( QgsLayerTreeNode *node : constSelectedNodes )
+  {
+    node->setItemVisibilityChecked( ! node->isVisible() );
   }
 }
 
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -7556,7 +7556,6 @@ void QgisApp::hideSelectedLayers()
   }
 }
 
-//reimplements method from base (gui) class
 void QgisApp::toggleSelectedLayers()
 {
   QgsDebugMsg( QStringLiteral( "toggling selected layers!" ) );
@@ -7572,7 +7571,6 @@ void QgisApp::toggleSelectedLayers()
   }
 }
 
-//reimplements method from base (gui) class
 void QgisApp::toggleSelectedLayersIndependently()
 {
   QgsDebugMsg( QStringLiteral( "toggling selected layers independently!" ) );
--- a/src/gui/layertree/qgslayertreeview.cpp
+++ b/src/gui/layertree/qgslayertreeview.cpp
@@ -507,6 +507,22 @@ void QgsLayerTreeView::mouseReleaseEvent( QMouseEvent *event )
 
 void QgsLayerTreeView::keyPressEvent( QKeyEvent *event )
 {
+  if ( event->key() == Qt::Key_Space )
+  {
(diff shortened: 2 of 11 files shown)


m-kuhn:
Nice.

Did you consider the alternative to enable all / disable all (toggle whole selection instead of each selected item individually)?

suricactus:
> Did you consider the alternative to enable all / disable all (toggle whole selection instead of each selected item individually)?

I am not sure if I understood you correctly, so please correct me, but this functionality is already present in the "eye" menu button. Just above my addition there are "Show Selected Layers" and "Hide Selected Layers"? It is only missing a shortcut, maybe "Ctrl+Space" is a good option?

uclaros:
This functionality is missing sooo much from QGIS!
I too had implemented it in my fork, (https://github.com/uclaros/QGIS/tree/toggle-visibility) feel free to grab the icon I made too that combines existing *toggle* icon with existing *show selected* icon.
I did not issue a PR though, because there are other things that need to be sorted with layer visibility and groups.

[QEP 86](https://github.com/qgis/QGIS-Enhancement-Proposals/issues/86) was a *major* step forward in usability of groups but some things are still counter-intuitive.
The tri-state checkbox of all layers gets out of sync with reality when ctrl-clicking a checked layer and checking an unchecked layer in groups automatically checks all required groups to make it visible.
Part of the issue can be seen on the first toggle of you video, where all layers are selected and toggled, but the layer in the group stays greyed while group gets checked!...
I'll issue a feature request regarding needed changes.

Also, the space functionality is OK in the layertreeview context but a different global shortcut is needed.

m-kuhn:
Yes, exactly like it's there in the eye menu. :+1: 
Space for one and Ctrl+Space for the other sounds good. I think I'll need the "toggle all" more often than the "toggle individual", so putting that one directly on Space without modifier would be my preference.

suricactus:
@uclaros wow, I am sorry to "steal" your PR. Maybe I should open draft PRs before doing something like this, so other may know in advance what I am currently doing? The maintainers should say what is the best practice here.

Amazing icon! I tried to do something like this in InkScape, but it took me more than half an hour and give up. I will take it for sure.

> I did not issue a PR though, because there are other things that need to be sorted with layer visibility and groups.

Yup, I also saw that sometimes the child remains greyed out, but I consider this as a different bug and which is out of scope for this PR.

> Also, the space functionality is OK in the layertreeview context but a different global shortcut is needed.

Do we really need a global shortcut for this? 99% of the time a user goes to the widget, selects their layers and that's it. The remaining 1% is nice to have but not required, as the "eye menu" action is always there. I think QGIS is already polluted with tons of useless shortcuts, that one can easily bind by the user if needed. I don't know if this philosophy comes from emacs or KDE, but we should not be that generous for new global shortcuts.

For example "Ctrl+Shift+T", which might be a sensible shortcut for toggling, but it is already bind to "Add CSV". Like adding CSV is really something you should do keyboard only and is that often action... I hope for QGIS 4 we get rid of the shortcut noise.

@m-kuhn Okay. It will apply the inverted state of the first item to the whole selection. I do agree that this functionality will be more popular than individual toggle, but shortcuts should remain as is IMHO. My user intuition is whenever I select something and "toggle it", as "Space" is the default shortcut for toggling, it should invert the state of all selected items individually. 

Once I switch branches yet again, I will add the other shortcut.

uclaros:
@suricactus Great minds think alike, they say! :)

Be careful, we're talking two different things here
1. Toggling the selected layers with space when layertreeview has focus via `QgsLayerTreeView::keyPressEvent`. In this case, you *could* toggle all to the state of the first selected (though which is the first?) kind of like the super practical ctrl+click of ArcMap that toggles the clicked layer and also all selected layers to the same state. (I really miss this functionality for expand - collapse selected too but that's another feature request)
2. Toggling the selected layers with the new `mActionToggleSelectedLayers`. This must toggle each selected layer's check state individually. This does not need a default shortcut (like space was bound in the initial commit) but it definitely needs to exist as an action so the user can bind a shortcut to it. It is amazingly useful to be able to A-B preview specific layers!

The second case may conflict with the current functionality regarding groups and grouped layers (the first one too if it toggles each layer individually after all).
Right now, checking a grouped layer implicitly checks all parent groups in order to make it visible. IMHO this should change to something more similar to the initial QEP 86 as follows:
- checked layers are visible
- unchecked layers are not visible
- checked layers that have at least one parent group unchecked are invisible and checkbox is greyed
- toggling *any* checkbox only toggles that checkbox, but might turn some checkboxes grey (including the checked one if it has an unchecked ancestor)
- offer a modifier (alt+click?) for *force visible* aka check this and all ancestors (also available in the context menu)
- offer a modifier (alt+click?) for goups to check/uncheck this and all children (also available in the context menu)

now the ctrl-click is freed to be used for toggle the clicked one and also set to the same state all selected (or all layers in current group level, if only one is selected), which will be great for ArcMap converts.

suricactus:
> Be careful, we're talking two different things here
@uclaros I believe these two should be two separate actions, both present in the menu and available for a global shortcut.

> though which is the first?

The first layer in the selection is the first that became active (clicked).

I have made the default shortcut to toggle as you both requested.

> Right now, checking a grouped layer implicitly checks all parent groups in order to make it visible. IMHO this should change to something more similar to the initial QEP 86 as follows:

I was using the wrong method, using the layer visibility, but not the layer check state. Fixed that.
Peek 2020-03-30 07-18

src/ui/qgisapp.ui
```diff
@@ -151,6 +151,8 @@
     <addaction name="mActionHideAllLayers"/>
     <addaction name="mActionShowSelectedLayers"/>
     <addaction name="mActionHideSelectedLayers"/>
+    <addaction name="mActionToggleSelectedLayers"/>
+    <addaction name="mActionIndependentlyToggleSelectedLayers"/>
```
uclaros:
```suggestion
    <addaction name="mActionToggleSelectedLayersIndependently"/>
```

src/app/qgisapp.cpp
```diff
@@ -7554,6 +7556,37 @@ void QgisApp::hideSelectedLayers()
   }
 }
 
+//reimplements method from base (gui) class
```
uclaros: this comment belonged to the method below, void QgisApp::hideDeselectedLayers()

src/app/qgisapp.cpp
```diff
@@ -7554,6 +7556,37 @@ void QgisApp::hideSelectedLayers()
...
+  {
+    bool isFirstNodeChecked = constSelectedNodes[0]->itemVisibilityChecked();
+    for ( QgsLayerTreeNode *node : constSelectedNodes )
+    {
+      node->setItemVisibilityChecked( ! isFirstNodeChecked );
+    }
+  }
+}
+
+//reimplements method from base (gui) class
```
uclaros: same for this comment

src/ui/qgisapp.ui
```diff
@@ -2673,6 +2675,26 @@ Acts on the currently active layer only.</string>
...
+   <property name="icon">
+    <iconset resource="../../images/images.qrc">
+     <normaloff>:/images/themes/default/mActionToggleSelectedLayers.svg</normaloff>:/images/themes/default/mActionToggleSelectedLayers.svg</iconset>
+   </property>
+   <property name="text">
+    <string>Toggle Selected Layers</string>
+   </property>
+   <property name="shortcut">
+    <string>Space</string>
+   </property>
```
uclaros: Space should be removed from a global shortcut.

src/gui/layertree/qgslayertreeview.cpp
```diff
@@ -507,6 +507,23 @@ void QgsLayerTreeView::mouseReleaseEvent( QMouseEvent *event )
...
+    if ( ! constSelectedNodes.isEmpty() )
+    {
+      bool isFirstNodeChecked = constSelectedNodes[0]->itemVisibilityChecked();
+      for ( QgsLayerTreeNode *node : constSelectedNodes )
+      {
+        node->setItemVisibilityChecked( ! isFirstNodeChecked );
+      }
+
+      // if we call the original keyPress handler, the current item will be checked to the original state yet again
+      return;
```
uclaros: If space is removed from the global shortcut this should not be necessary

uclaros:
> The first layer in the selection is the first that became active (clicked).

I'm afraid this is only true when using ctrl+click to select. When using shift+click the layertreeview order is used.

src/ui/qgisapp.ui
suricactus: Word dyslexia, thanks

src/ui/qgisapp.ui
suricactus: This is not a global shortcut, but a widget shortcut (see the next property). I use it only to hive hint to the user that this exists. If have any idea about this, it would be appreciated.

src/gui/layertree/qgslayertreeview.cpp
suricactus: Space is not a global shortcut. Unfortunately it is still needed.

src/ui/qgisapp.ui
uclaros: This is confusing. You are using an action's shortcut (`mActionToggleSelectedLayers`) to advertise functionality that does not involve the action itself, since the space button is handled directly by capturing the event in `QgsLayerTreeView`. Everywhere else in the ui the shortcut displayed is the global shortcut for the action, we should stick to that.

nirvn:
+1 to merge, UX win.

uclaros:
The code in `QgsLayerTreeView::keyPressEvent` should be triggering the action, not reimplement it. Could we avoid the duplicate code to make future edits safer?

PR #35414, opened 2020-03-27
 FIX #19903 Greatly improve font selection UX

### LINKED ISSUE #19903: Limit amount of fonts used in QGIS
Author Name: **Tobias Schneider** (Tobias Schneider)
Original Redmine Issue: [11651](https://issues.qgis.org/issues/11651)

Redmine category:unknown

---
I would like it very much if I could choose a subset from all the fonts that are installed on my OS to be used in QGIS. On Linux the list always fills the screen and it is difficult to select the correct one. Why not have a dialogue on the options menu where users can select their usual mapping fonts, and only those are shown in all other dialogues of QGIS (for example the layer styles, label menu, ...).

suricactus:
Trying to select a font on a system with thousands of installed fonts is huge pain under Linux. It causes the whole interface to freeze and does not allow searching without opening the huge list of fonts. On the gifs below the pause is caused by the font dropdown itself, it is not me. Also there is a little flash, which also feels buggy.
In addition, the font selector was inconsistent between different interfaces, now it is always an editable combobox.
Final note - in case of invalid font name, the widget uses the last selected one.
|before|after|
|------|-----|
|old|new|

### DIFF
--- a/src/ui/labeling/qgslabelpropertydialogbase.ui
+++ b/src/ui/labeling/qgslabelpropertydialogbase.ui
@@ -113,11 +113,7 @@
          </property>
          <layout class="QGridLayout" name="gridLayout">
           <item row="0" column="0" colspan="3">
-           <widget class="QFontComboBox" name="mFontFamilyCmbBx">
-            <property name="editable">
-             <bool>false</bool>
-            </property>
-           </widget>
+           <widget class="QFontComboBox" name="mFontFamilyCmbBx" />
           </item>
           <item row="1" column="0">
            <widget class="QLabel" name="label_3">
--- a/src/ui/qgseditconditionalformatrulewidget.ui
+++ b/src/ui/qgseditconditionalformatrulewidget.ui
@@ -317,11 +317,7 @@
          </widget>
         </item>
         <item>
-         <widget class="QFontComboBox" name="mFontFamilyCmbBx">
-          <property name="editable">
-           <bool>false</bool>
-          </property>
-         </widget>
+         <widget class="QFontComboBox" name="mFontFamilyCmbBx" />
         </item>
        </layout>
       </item>
--- a/src/ui/qgsoptionsbase.ui
+++ b/src/ui/qgsoptionsbase.ui
@@ -606,11 +606,7 @@
                      </widget>
                     </item>
                     <item>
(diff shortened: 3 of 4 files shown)


nyalldawson:
Looks good to me. This widget has been on my mind too -- it's got quite a lot of other issues, and there's lots of scope for improvement.

E.g.
- It would be nice to have quick shortcuts to recently used fonts and "favorite fonts" (or even some automatically built list of "your commonly used fonts"
- (like a lot of things, it's ESRI's fault, but...) many ESRI symbol fonts are badly constructed and aren't correctly tagged in the font files themselves as a symbol font. The consequence of this is that users can't see the font names in the list, just a string of symbol characters:

image

the Qt widget requires the symbol tag in order to change to the "show font name in standard font first, then in symbols" behavior, which explains why many of the fonts in that screenshot are impossible to pick from (again - ESRI's fault. You see the same behavior in other applications like MS word with these fonts). But we could easily "fix" this in a custom widget by always treating fonts with a family starting with "ESRI" as a symbol font.

suricactus:
I was also thinking about "recent" and "frequent" fonts, however I am not so sure how put it in a nice UI. Probably something like locator's grouping can do the job, but practically it means to implement the whole widget from scratch. I think this PR is enough for now, because it feels very buggy.

On the ESRI frontier... wow, never knew that. It seems that custom widget is the way to go.

nyalldawson:
> but practically it means to implement the whole widget from scratch

That was the conclusion I came to also. We can't extend the qt widgets themselves to do what we need, so we'd be re-writing from scratch (or copying qt code + adapting)

nirvn:
Or submitting an upstream patch?

nyalldawson:
I doubt they'd accept a hack workaround for ESRI's mistake (and rightly not)

nirvn:
Oh so you think the missing font name here is that narrow (ie only esri fonts to your knowledge)? Well then yeah, I see your point :)

nyalldawson:
Well, you'd need to know in advance which fonts are incorrectly tagged in order to handle them specially... And I'm only aware of the ESRI ones

nirvn:
Btw, this improvement needs to be added to the font marker widget ui too.

suricactus:
For me it is already fine without any change? Current master and 3.12:
image

Issue #35258, opened 2020-03-23
Copy/paste all styles of a layer


SGroe:
**Feature description.**

I often have to copy multiple styles from one layer to another. With the current implementation, only one style can be copied at once. I would like to see the option to copy and paste all or maybe selected styles.

**Additional context**

3nids:
you can copy all styles from the layer tree context menu
![image](https://user-images.githubusercontent.com/127259/77291457-98c7ec80-6cde-11ea-97e5-e4c3610aa1af.png)

you can fine tune what to save when you save to a file:
![image](https://user-images.githubusercontent.com/127259/77291549-c6ad3100-6cde-11ea-848e-426d758d8820.png)

adding the clipboard destination to the fine tuning is not that easy. The clipboard is in app and the vector layer properties was recently move to gui.

SGroe:
I mean something different. You can add a second, third, ... style to a layer (layer context menu > Styles > Add...). I want to to copy all styles and not all style categories from the currently active style.

roya0045:
I agree that this would be nice, this seems relatively straightforward.

NyakudyaA:
Would this also extent to exporting layer styles to qml or sld. Would it be another feature request.

roya0045:
I can try come up with something, if I add an 'import from layer' and 'override style from layer' with a menu that shows available layer from the same style, is that good for everyone?

SGroe:
It should certainly be possible to paste the styles to multiple layers at once (e.g. select layers and "Paste styles".

NyakudyaA:
> I can try come up with something, if I add an 'import from layer' and 'override style from layer' with a menu that shows available layer from the same style, is that good for everyone?

I think you just need to do as is done for the exporting of layer styles.
![Selection_003](https://user-images.githubusercontent.com/2510900/77396584-b6fc1e00-6dac-11ea-84e7-5339d569ef18.png)

roya0045:
@NyakudyaA I wasn't looking at the export route more at how to import and replicate all styles for a given layer as proposed in the main ticket.

NyakudyaA:
> @NyakudyaA I wasn't looking at the export route more at how to import and replicate all styles for a given layer as proposed in the main ticket.

Yes, understood. I also meant that for the copy styles we could employ the same GUI logic as export. Where a user can choose all styles or just a single style to copy

PR #35416, opened 2020-03-28
[Feature] import and replace all styles of a layer

roya0045:
## Description

This is my proposal to address #35258 . 

With this the user can import all the styles from another layer in the same project. I also added the option to override all styles of the current layer with another layer.

This adds two items in the style menu and the styling widget. One clicked a menu will pop up based on the layout legend selector widget listing layers with compatible symbology.

The fact that the current style of both the target and the host are unstored is somewhat annoying to deal with and could leak to some funky stuff. Summary tests are fine but I might be missing a signal or step to refresh the symbology when overriding.

style


roya0045:
Added screenshot

PR #35420, opened 2020-03-28
[locator] Allow searching for short identifiers

### LINKED ISSUE #35418: Search for short feature identifiers in the locator bar
Currently, the locator feature filter only shows results once 3 or more characters have been typed.
This makes it impossible to find ids like 1, 20, 45, ab, etc.
By using a prefix, these are included but can be hidden because of the maximum result count of 6. (ie. 100, 1000, 2100, 111, 110, 3310 shown instead of 1).

Ideas:
 - order by relevance (ideally provider side for performance reasons)
 - for < 3 characters search for exact match only
 - introduce double quotes for exact match search `"1"` (not that intuitive)

m-kuhn:
When searching for 1 or 2 character strings, execute an exact match search.
This allows to search for very short identifiers which are otherwise hard to impossible to find.

This also adds forcing exact match search by quoting the search string.

Fix #35418

# Documentation

To search for very short identifiers and values with the "all features" or "active layer features" locator filters it is required to use the prefix (`f`, `af`) by default. When used with a prefix, these filters will perform an exact match search for 1 or 2 characters. For example `af 1` will find a layer with display string `1` on any layer, but it will not find `100`. `af 100` will find `100` and `1000`.

Searching for quoted strings will also perform an exact match search. E.g. `"Wil"` will find `Wil` but not `Wattenwil`.

Peek 2020-03-28 07-39

### DIFF
--- a/src/app/locator/qgsinbuiltlocatorfilters.cpp
+++ b/src/app/locator/qgsinbuiltlocatorfilters.cpp
@@ -242,8 +242,10 @@ void QgsActiveLayerFeaturesLocatorFilter::prepare( const QString &string, const 
   if ( string.length() < 3 && !context.usingPrefix )
     return;
 
+  QString searchString = string;
+
   bool allowNumeric = false;
-  double numericalValue = string.toDouble( &allowNumeric );
+  double numericalValue = searchString.toDouble( &allowNumeric );
 
   QgsVectorLayer *layer = qobject_cast< QgsVectorLayer *>( QgisApp::instance()->activeLayer() );
   if ( !layer )
@@ -254,8 +256,15 @@ void QgsActiveLayerFeaturesLocatorFilter::prepare( const QString &string, const 
   mDispExpression.prepare( &mContext );
 
   bool exactMatch = false;
-  if ( string.length() < 3 )
+  if ( searchString.length() < 3 )
     exactMatch = true;
+
+  if ( searchString.startsWith( '"' ) && searchString.endsWith( '"' ) )
+  {
+    exactMatch = true;
+    searchString.remove( 0, 1 );
+    searchString.chop( 1 );
+  }
 
   // build up request expression
   QStringList expressionParts;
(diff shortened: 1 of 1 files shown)


3nids:
Nice! It would be nice to add the quotes for exact match too. It would be quite handy

m-kuhn:
On it already ;)

3nids (approved):
Nice.
It would have been even nicer with some regex though 😉

m-kuhn:
> It would have been even nicer with some regex though wink

Apart from reduced performance, what's the advantage ;)

3nids:
Code gets harder to read!

roya0045:
@3nids what about #35083 ?

uclaros:
So how are we going to search for features containing one or two character strings?
The problem of searching exactly for one or two character strings gets solved with the quoted for exact match mode. Imho you are breaking functionality, or I didn't get this right!

m-kuhn:
Old behavior: typing 1 or 2 characters with prefix would not do anything.
New behavior: typing 1 or 2 characters with prefix will search for an exact match.

Can you clarify on what's broken?

uclaros:
Sure, maybe this makes it clearer.

locator

m-kuhn:
Sorry, my comment was indeed wrong. Search happened on short identifiers too with prefix.
But the issue is shown in the below video:

Peek 2020-03-28 12-17

uclaros:
NP, but what you describe can be solved with the double quotes

m-kuhn:
Technically yes. But I assume that in practice the intuitiveness of this approach is better
(quoting will require documentation which no one reads. I can only estimate the percentage of people using quotes on google but the figures I imagine are a very small portion of all searches).

m-kuhn:
> ~NP, but what you describe can be solved with the double quotes~
> whoops, rushed to answer!.. What you show here is that it fails to get all matches with the single "2" character, it only gets some ("12")

What happens is that it fails to order by relevance and trims the results to a max of 6.
There would be a better solution to support "order by relevance" but this needs to be done provider side for efficiency and that's out of scope for a Saturday morning quarantine fun project ;)

uclaros:
I see, but it's not good enough reason to not show my single result matching a one or two character string!

m-kuhn:
> I see, but it's not good enough reason to not show my single result matching a one or two character string!

It fixes a real world problem, I find this reason quite compelling ;)

uclaros:
I don't agree with your logic here, sorry.
Nobody will be able to do one or two character string 'contains' searches (a real world problem) because you don't want to reach for the double quotes (not a real world problem tbh).

m-kuhn:
So sorry I have to close this.

It was for a fun project for someone who works a lot with small municipalities where double quotes are not an option.

PR #35422, opened 2020-03-28
Save vector layer: for CSV/XLS/XLSX/ODS, only use by default displayed values of a restrict set of widget types

rouault:
This is a follow-up of https://github.com/qgis/QGIS/pull/35296

This implements the suggestion of
https://github.com/qgis/QGIS/pull/35296#issuecomment-604645943

That is the 'Use displayed value' column is checked by default only when
the field has a widget of type ValueMap, ValueRelation, CheckBox and
RelationReference

And when the Attribute table of the dialog box is not displayed (e.g. when
saving a temporary scratch layer), disable those substitutions.

CC @nirvn @andreasneumann

### DIFF
--- a/src/gui/ogr/qgsvectorlayersaveasdialog.cpp
+++ b/src/gui/ogr/qgsvectorlayersaveasdialog.cpp
@@ -412,7 +412,11 @@ void QgsVectorLayerSaveAsDialog::mFormatComboBox_currentIndexChanged( int idx )
   }
 
   bool selectAllFields = true;
-  bool fieldsAsDisplayedValues = false;
+
+  // Is it a format for which fields that have attached widgets of types
+  // ValueMap, ValueRelation, etc. should be by default exported with their displayed
+  // values
+  bool isFormatForFieldsAsDisplayedValues = false;
 
   const QString sFormat( format() );
   if ( sFormat == QLatin1String( "DXF" ) || sFormat == QLatin1String( "DGN" ) )
@@ -423,7 +427,13 @@ void QgsVectorLayerSaveAsDialog::mFormatComboBox_currentIndexChanged( int idx )
   else
   {
     if ( mOptions & Fields )
+    {
       mAttributesSelection->setVisible( true );
+      isFormatForFieldsAsDisplayedValues = ( sFormat == QLatin1String( "CSV" ) ||
+                                             sFormat == QLatin1String( "XLS" ) ||
+                                             sFormat == QLatin1String( "XLSX" ) ||
+                                             sFormat == QLatin1String( "ODS" ) );
+    }
   }
 
   // Show symbology options only for some formats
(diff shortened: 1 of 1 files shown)


nirvn:
I'm feeling better with this less intrusive approach :) one thing needs adding: when the attribute table isn't shown (eg when someone saves a temporary scratch layer via its "make permanent" indicator) we need to insure data being saved stays indentical, not changed to its displayed value.

andreasneumann:
Thanks @rouault 
I would also add the relation reference, date and checkbox widget to enable presentation values by default. If someone really needs the internal values, he can check off the presentation value

@nirvn: txt/csv/xlsx/odf are usually (I'd say 90% or more) an export format for presentation, and not a storage format for original data. These formats are also not editable in QGIS, so there is a very tiny chance that someone would use these formats as a storage format.

@m-kuhn as the widget specialist - do you have any thoughts here?

m-kuhn:
I like it :+1: 

The additional widget types to add sound good to me too. Not sure about date (that can be exported spread sheet compatible, can't it?)

nirvn:
I'd be -1 for dates. This data type is handled by spreadsheet programs, you have ample flexibility to format strings as you want in excel, libreoffice calc, etc. 

Transforming dates to strings can also break sorting by those fields in your favorite spreadsheet software. 

IMHO, killing date fields *by default* isn't a good behavior.

I can see your argument being valid for valuemap and valuerelation fields.

rouault:
> when the attribute table isn't shown (eg when someone saves a temporary scratch layer via its "make permanent" indicator) we need to insure data being saved stays indentical, not changed to its displayed value.

I've just amended my commit to add this additional check

andreasneumann:
> I'd be -1 for dates. This data type is handled by spreadsheet programs, you have ample flexibility to format strings as you want in excel, libreoffice calc, etc.

Ok - let's have dates as internal format. They user can format it again in the Spreadsheet - unless there is a way to tell XLSX/ODF to export as a native date format and apply the date formatting at the same time ... but I can live with no presentation value by default.

nirvn:
@rouault ,  merci.

@andreasneumann , thanks, I think this little brainstorming is landing us on a pretty sane default :) cheers

rouault:
> unless there is a way to tell XLSX/ODF to export as a native date format and apply the date formatting at the same time

I guess this should in theory be doable, but this isn't implemented in the OGR XLSX & ODS drivers currently

rouault:
So the consensus is to add checkbox and relation reference as well ?

andreasneumann:
> I guess this should in theory be doable, but this isn't implemented in the OGR XLSX & ODS drivers currently

Never mind - not a problem.

andreasneumann:
> So the consensus is to add checkbox and relation reference as well ?

Yes, please

rouault:
> Yes, please

done

andreasneumann:
Thanks a lot, @rouault

m-kuhn (approved):

PR #35423, opened 2020-03-28
[locator] Prefer exact match

### LINKED ISSUE #35418: Search for short feature identifiers in the locator bar

m-kuhn:
This tackles issue #35418 . When a search string is entered for the "all layers locator filter" it will perform two searches: priority goes to exact matches, non-exact matches are also listed but lower in the results.

Fix https://github.com/qgis/QGIS/issues/35418

Peek 2020-03-28 14-35

### DIFF
--- a/src/app/locator/qgsinbuiltlocatorfilters.cpp
+++ b/src/app/locator/qgsinbuiltlocatorfilters.cpp
@@ -386,7 +386,12 @@ void QgsAllLayersFeaturesLocatorFilter::prepare( const QString &string, const Qg
     enhancedSearch.replace( ' ', '%' );
     req.setFilterExpression( QStringLiteral( "%1 ILIKE '%%2%'" )
                              .arg( layer->displayExpression(), enhancedSearch ) );
-    req.setLimit( 30 );
+    req.setLimit( 6 );
+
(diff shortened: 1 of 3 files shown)


nirvn (approved):
Really nice approach. All gains, no pains ;)

uclaros:
@m-kuhn You rock! Sorry if I was pushy on the other thread!
The option for exact matches only was great though, it could indeed be implemented by using the double quotes or a new filter altogether, like `ef` for exact match find!

m-kuhn:
Yes, there is still a lot of potential on this end, but let's focus on getting this one locked in first.

nyalldawson:
Can you add a test for this to testqgsapplocatorfilters.cpp please?

PR #35432, opened 2020-03-28
[Bugfix] Fix table sorting

### LINKED ISSUE #35427: [print layouts ] table, field sorting does not work
I have a table with over **4k rows**, I add the table in the print composer and I sort it according to the `data` field (datetime), the sorting does not work.

I notice that to work I have to write a number greater than **4k** in the main properties of the table (default is 30)

- first (30 lines)

image

- after (5k lines)

image

I attach vrt files:

[dpc-covid19-ita-province.zip](https://github.com/qgis/QGIS/files/4397674/dpc-covid19-ita-province.zip)

## how to replicate it:

1. download .zip file;
2. unzip it;
3. dragAndDrop vrt file in the TOC;
4. start the composer and add table;
5. sort the table using the date field;

## OSGeo4W64 Win 10

image

roya0045:
## Description
Fixes #35427 

I'll add the comment when I'll squash my commit, if I don't forget.

PR was tested, all is fine. Artefact is available as usual.

### DIFF
--- a/src/core/layout/qgslayoutitemattributetable.cpp
+++ b/src/core/layout/qgslayoutitemattributetable.cpp
@@ -519,6 +519,12 @@ bool QgsLayoutItemAttributeTable::getTableContents( QgsLayoutTableContents &cont
     req.setFilterFid( atlasFeature.id() );
   }
 
+  QVector< QPair<int, bool> > sortColumns = sortAttributes();
+  for ( int i = sortColumns.size() - 1; i >= 0; --i )
+  {
+    req = req.addOrderBy( mColumns.at( sortColumns.at( i ).first )->attribute(), sortColumns.at( i ).second );
+  }
+
   QgsFeature f;
   int counter = 0;
   QgsFeatureIterator fit = layer->getFeatures( req );
@@ -626,16 +632,6 @@ bool QgsLayoutItemAttributeTable::getTableContents( QgsLayoutTableContents &cont
     tempContents << currentRow;
     existingContents << rowContents;
     ++counter;
-  }
-
-  //sort the list, starting with the last attribute
-  QgsLayoutAttributeTableCompare c;
-  QVector< QPair<int, bool> > sortColumns = sortAttributes();
-  for ( int i = sortColumns.size() - 1; i >= 0; --i )
-  {
-    c.setSortColumn( sortColumns.at( i ).first );
-    c.setAscending( sortColumns.at( i ).second );
-    std::stable_sort( tempContents.begin(), tempContents.end(), c );
   }
 
   // build final table contents
--- a/src/core/layout/qgslayoutitemattributetable.cpp
+++ b/src/core/layout/qgslayoutitemattributetable.cpp
@@ -32,45 +32,6 @@
 #include "qgsgeometryengine.h"
 #include "qgsconditionalstyle.h"
 
-//QgsLayoutAttributeTableCompare
-
-///@cond PRIVATE
-
-/**
- * Helper class for sorting tables, takes into account sorting column and ascending / descending
-*/
-class CORE_EXPORT QgsLayoutAttributeTableCompare
-{
-  public:
-
-    /**
-     * Constructor for QgsLayoutAttributeTableCompare.
-     */
-    QgsLayoutAttributeTableCompare() = default;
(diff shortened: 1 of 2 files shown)


src/core/layout/qgslayoutitemattributetable.cpp
```diff
@@ -518,6 +518,12 @@ bool QgsLayoutItemAttributeTable::getTableContents( QgsLayoutTableContents &cont
     QgsFeature atlasFeature = mLayout->reportContext().feature();
     req.setFilterFid( atlasFeature.id() );
   }
+  
+  QVector< QPair<int, bool> > sortColumns = sortAttributes();
+  for ( int i = sortColumns.size() - 1; i >= 0; --i )
+  {
+     req = req.addOrderBy( mColumns.at( sortColumns.at( i ).first )->heading(), sortColumns.at( i ).second );
```
nyalldawson: `heading` isn't appropriate here -- that's a free text field, unrelated to the column's content

roya0045: Ah right, I see what you mean, I'll get this sorted out tomorrow with the indentation and the commit comment. Thanks!

nyalldawson: Thanks! and yes, a test would be valuable for this too (better do a range of tests -- one with a attribute only column, with with expression based column, and one with descending sort order)

roya0045: I'm looking at the code and they way to set sort columns in QgsLayoutItemAttributeTable seem to be done in the GUI side of things as there are no ways to set this in the base class. The columns to handle sorting is in qgslayoutattributeselectiondialog.cpp I would think that implementing functional test would require a revisit of the current handling or a new suite of test relying on gui elements. I think that this is out of the scope of a simple bugfix. Let me know what you think.

roya0045:
Test failure is in qgs font maker test, 100% unrelated.

roya0045:
I guess a backport would be needed.

src/core/layout/qgslayoutitemattributetable.cpp
roya0045: Maybe I'm bad at this but classes are so convoluted I can't even find a way to make an new column with an expression with the exposed api. Sorting might be doable but I can't access the columns to perform sorting, again an api rework would be needed. Currently my changes work with minimal impact. I think we'll have to leave it at that until tests are easier to perform. :/

nyalldawson: It's not that hard. No API work necessary full stop. Look at TestQgsLayoutTable::attributeTableSetAttributes() and do something like

    table->columns().at( 0 )->setAttribute( "some expression which messes with a sort order like taking 0 - an int column value" );

roya0045: Ah right, I was expecting a function to expose things and create a new column, instead of replacing a column but this will do.

roya0045: Tests are done

roya0045:
Test failure is postgres raster, unrelated.

roya0045:
Still confused as to why  PyQgsPostgresRasterProvider failed.

src/core/layout/qgslayoutitemattributetable.cpp
```diff
@@ -628,16 +634,6 @@ bool QgsLayoutItemAttributeTable::getTableContents( QgsLayoutTableContents &cont
     ++counter;
   }
 
-  //sort the list, starting with the last attribute
-  QgsLayoutAttributeTableCompare c;
```
nyalldawson, line 632: You can completely remove QgsLayoutAttributeTableCompare now -- it's private and unused.

nyalldawson (commented):
LGTM! Can you just remove the QgsLayoutAttributeTableCompare class as a final cleanup?

src/core/layout/qgslayoutitemattributetable.cpp
roya0045: The full class?

roya0045: Done

nyalldawson:
Great work, thanks!

Issue #27230, opened 2018-07-12
[Aggregates] Allow majority and minority statistics on non-numeric fields


qgib:
Author Name: **Harrissou Santanna** (@DelazJ)
Original Redmine Issue: [19402](https://issues.qgis.org/issues/19402)

Redmine category:expressions

---
I wonder if it's intentional or if it's a missing feature but currently, there's no way to know the most/less present (aka majority/minority) value in a field if that field is not of numeric type. For example, you may want to know which species is most present in an area but it does not look possible using the majority function if species are not coded as numeric.
This concerns the statistical panel as well as expression functions or aggregate algorithm.

PedroVenancio:
This issue is related to https://github.com/qgis/QGIS/issues/27271

I also think that majority/minority aggregators should work with string fields. Tested in QGIS 3.10.1, and majority/minority just work for numeric fields.

suricactus:
The problem is with the aggregate expression in general, this is not working:
```
 aggregate( @layer_id , 'majority',  "string_column" )
```
Both on ogr and postgis layers.

alexbruy:
Fixed in master.

Issue #35433, opened 2020-03-29
String aggregate-related functions treat NULL and empty strings interchanchebly


suricactus:
**Describe the bug**
Pretty much the title. I've attached the data layer I am using for calculation, but this screenshot summarizes the issue.
![image](https://user-images.githubusercontent.com/2820439/77839253-c2bc5b80-7183-11ea-8af0-b8a8796d93ac.png)
The results are scary when "concatenate" with delimiter is used, for example`concatenate(str, NULL, NULL, ' |  ')` or `count_missing(str)`.
**Expectations**
I was expecting that `NULL` values are going to be omitted as they do not represent countable values. This is already the case for aggregates for numeric types. 
Reading the documentation, especially on the `count_missing`, it states only `NULL` values should be counted.
https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/expression.html#aggregates-functions

> Function | Description
> -- | --
> count | Returns the count of matching features
> count_distinct | Returns the count of distinct values
> count_missing | Returns the count of missing (null) values
> 

**How to Reproduce**
1) load the attached layer
2) this is the expression I'm using to display the result. Any other aggregate function is also affected.
```
'\ncount(int) ' || count("int") || 
'\ncounr(str) ' || count("str") ||
'\nmissing(int) ' || count_missing("int") || 
'\nmissing(str) ' || count_missing("str")
```
**QGIS and OS versions**
All QGIS3 versions on all platforms, it is in the very core of how statistics for strings work.
**Additional context**
Applying the fix breaks a few existing unit test, but also might break user expressions dependent on this bug.

m-kuhn:
I fully agree :+1:

overkamp:
In QGIS 3.12.0 (cd141490ec) I see empty strings in text fields (attribute table) as NULL value (postgis datasource). IN QGIS 3.10.1 (ef24c526da) I see NULLs (in text fileds) as empty string. I think there is something wrong.

PR #35434, opened 2020-03-29
FEAT #27230 [Aggregates] Allow majority and minority statistics on string fields

### LINKED ISSUE #27271: majority/minority
Author Name: **Alexandre Heurion** (Alexandre Heurion)
Original Redmine Issue: [19443](https://issues.qgis.org/issues/19443)

Redmine category:expressions

---
Bonjour,

les fonctions majority et minority (notamment disponible dans l'outil agrégation) ne fonctionnent pas pour champs de chaînes de caractère. lors d'une agrégation il est donc impossible de choisir de grouper un champs de type string suivant le nombre de valeurs rencontrées, sans passer par la création d'un nouveau champs attribuant un code en format nombres entiers à chaque valeur de la colonne initiale.

---

- [erreur majority.PNG](https://issues.qgis.org/attachments/download/12982/erreur%20majority.PNG) (Alexandre Heurion)
- [paramètres agrégation.png](https://issues.qgis.org/attachments/download/12983/param%C3%A8tres%20agr%C3%A9gation.png) (Alexandre Heurion)
- [absence d_information sur le format de l_expression en entrée.PNG](https://issues.qgis.org/attachments/download/12984/absence%20d_information%20sur%20le%20format%20de%20l_expression%20en%20entr%C3%A9e.PNG) (Alexandre Heurion)

suricactus:
This PR includes:
- allow majority and minority statistics on string fields
- add string MeanLength statistic as an option in the statistical summary widget
image

Fixes #27271 #27230


src/core/qgsstringstatisticalsummary.h
```diff
@@ -206,14 +220,16 @@ class CORE_EXPORT QgsStringStatisticalSummary
     Statistics mStatistics;
 
     int mCount;
-    QSet< QString > mValues;
+    QMap< QString, int > mValues;
```
suricactus: Note the datatype change here as an implementation detail. Since strings may be quite large, use a single QMap that stores both the unique keys and the counts. This differs from the numeric implementation in `QgsStatisticalSummary`, where two separate objects are used.

suricactus: Just to be clear, this is an implementation detail, other than little performance change due to `QSet` and `QMap` implementation, there should be no other changes for the users of the API.

m-kuhn: Would that better be a QHash?

src/core/qgsstringstatisticalsummary.h
```diff
@@ -195,6 +197,18 @@ class CORE_EXPORT QgsStringStatisticalSummary
...
     double meanLength() const { return mMeanLength; }
 
+    /**
+      * Returns the most common string.
+      * \since QGIS 3.14
+      */
+    QString minority() const { return mMinority; }
+
+    /**
+      * Returns the least common string.
```
m-kuhn: Are the two docstrings here swapped?

src/core/qgsstringstatisticalsummary.h
suricactus: Nope, we need them sorted.

src/core/qgsstringstatisticalsummary.h
suricactus: Thanks!

src/core/qgsstringstatisticalsummary.h
suricactus: Since you are there, have a look on this too https://github.com/qgis/QGIS/issues/35433

suricactus:
Since you are there, have a look on this too #35433

src/core/qgsstringstatisticalsummary.h
m-kuhn: Where exactly?

suricactus: `distinctValues()` is not required, but really nice to return ordered results. The performance gain from `QHash` would be negligible.

m-kuhn: `distinctVaues()` returns a `QSet` which does not preserve the order.

src/core/qgsstringstatisticalsummary.cpp
```diff
@@ -75,6 +77,21 @@ void QgsStringStatisticalSummary::addValue( const QVariant &value )
 void QgsStringStatisticalSummary::finalize()
 {
   mMeanLength = mSumLengths / static_cast< double >( mCount );
+
+  if ( mStatistics & Minority || mStatistics & Majority )
+  {
+    QList<int> valueCounts = mValues.values();
+    std::sort( valueCounts.begin(), valueCounts.end() );
```
m-kuhn: This is quite an overhead with a complexity of `n log(n)` instead of `n`.

```
double min = *std::min_element(valueCounts.begin(), valueCounts.end());
double max = *std::max_element(valueCounts.begin(), valueCounts.end());
```

src/core/qgsstringstatisticalsummary.h
suricactus: You are right, I got confused with ECMA implementation of `Set` object. I sill believe it would be nice to keep the order for the user, but if you say I would change that.

m-kuhn: Please do. Values can still be sorted after (even with better control over locale etc. details).

src/core/qgsstringstatisticalsummary.cpp
suricactus: It was shameless copy-paste from here https://github.com/qgis/QGIS/blob/master/src/core/qgsstatisticalsummary.cpp#L231-L240, I guess this optimization is applicable there as well?

m-kuhn: Pretty much. If you can do that too, that's appreciated :+1:

suricactus: Done! Thanks a lot for the feedback! Every time I learn a new bit.

src/core/qgsstringstatisticalsummary.h
suricactus: I meant keeping the order as they appear in the features. Anyway, it's a `QHash` now.

m-kuhn (approved):
Good work

src/core/qgsstringstatisticalsummary.h
suricactus: In case of multiple values that are correct answer for `minority` or `majority` statistic, implementation based on `QHash` results in completely random choice.
```
for i in range(10):
  s = QgsStringStatisticalSummary()
  s.calculate(["one", "two", "one", "two"])
  print(s.minority(), s.majority())
```
Using `QMap` is already the solution to this problem in `QgsStatisticalSummary`, so I believe other solutions would bring more noise than benefit. 
Now it is certain, that `minority` and `majority` will return the first occurrence. ("one" is the example above)

Saijin-Naib:
Could Mode be added for this and numeric types? It can be a very important metric to summarize a dataset.

suricactus:
Isn't the current implementation of `majority` statistical mode? The documentation states:

> The majority is the value with most occurrences in the list.

https://qgis.org/api/classQgsStatisticalSummary.html#a345c1ef1b364b0db11023c3c9c3e699a

suricactus:
There is a difference though, what to do if the set has multiple modes.

src/core/qgsstringstatisticalsummary.h
nyalldawson: @m-kuhn it's not always the case that a QHash is a better choice than a QMap for a QString key anyway -- see https://thesmithfam.org/blog/2011/02/12/when-faster-is-actually-slower/. ( I assume that's unchanged for qt 5)

PR #34869, opened 2020-03-04
[FEATURE][Symbology] Add percantage size unit for Raster Image Marker Layer symbology.

beketata:
## Description

According to [Feature Request](https://github.com/qgis/QGIS/issues/30057) this patch adds new percentage (percents of original image size) size unit for Raster Image Marker Layer.

Issue #30057, opened 2019-06-03
Display Raster Image Marker at 100% of original size.


beketata:
Display Raster Image Marker at 100% of original size.

There is no option to choose original size of raster file image.
The combo box with measuring units of size group contains only:
"Millimeters, Points, Pixels, Meters at Scale, Map Units, Inches"
The good idea to add an extra item something like: "Scale of Original Size"

![RasterImageMarker](https://user-images.githubusercontent.com/7413903/58785458-51fdee80-8607-11e9-8677-8bbb8568b148.jpg)

PR #35438, opened 2020-03-29
Add percentage size unit for Raster Image Marker and Raster fill layers symbology.

beketata:
## Description

According to the [Feature Request](https://github.com/qgis/QGIS/issues/30057) this patch adds new percentage (percents of original image size) value of size units for **Raster Image Marker** and **Raster fill** (by [nirvn](https://github.com/qgis/QGIS/pull/34869#issuecomment-595090465) suggestion) layers.

For example a layer contains raster images of various sizes and aspect ratios:

crocodile giraffe

If we use an existing Raster image marker **size unit** values the layer will look like this (with normal croc, but huge giraffe):

crocGir

or like this (with normal giraffe but tiny croc):

GirCroc

The new **percentage** value of size units is intended to fix this issue.

===========================================

Both unit tests "qgis_rasterfilltest" and "qgis_rastermarkertest" were rewritten with a set of new control images for testing proposed **percentage** value of size units.

<!--
  IMPORTANT NOTES FOR FIRST TIME CONTRIBUTORS
  ===========================================

  Congratulations, you are about to make a pull request to QGIS! To make this as easy and pleasurable for everyone, please take the time to read these lines before opening the pull request.

  Include a few sentences describing the overall goals for this pull request (PR). If applicable also add screenshots or - even better - screencasts.
  Include both: *what* you changed and *why* you changed it.

  If this is a pull request that adds new functionality which needs documentation, give an especially detailed explanation.
  In this case, start with a short abstract and then write some text that can be copied 1:1 to the documentation in the best case.

  Also mention if you think this PR needs to be backported. And list relevant or fixed issues.


beketata:
How to know which unsuccessful tests require my participation and which do not?

src/core/symbology/qgsfillsymbollayer.cpp
```diff
@@ -3906,6 +3906,21 @@ void QgsRasterFillSymbolLayer::applyPattern( QBrush &brush, const QString &image
   if ( width > 0 )
   {
     size.setWidth( context.renderContext().convertToPainterUnits( width, mWidthUnit, mWidthMapUnitScale ) );
+
+    // RenderPercentage Unit Type takes original image size
+    if ( mWidthUnit == QgsUnitTypes::RenderPercentage )
```
nyalldawson, line 3911: this should be in an "else" block, avoiding the unused setWidth call from above. E.g.

    if ( mWidthUnit != QgsUnitTypes::RenderPercentage )
    {
      size.setWidth( context.renderContext().convertToPainterUnits( width, mWidthUnit, mWidthMapUnitScale ) );
    }
    else
    {
       QSize originalSize = QgsApplication::imageCache()->originalSize( imageFilePath );
       ...
    }

tests/src/core/testqgsrasterfill.cpp
```diff
@@ -96,11 +100,18 @@ void TestQgsRasterFill::initTestCase()
...
-  //setup raster fill
+  // Setup raster fill
   mRasterFill = new QgsRasterFillSymbolLayer();
+  mRasterFill->setImageFilePath( mTestDataDir + QStringLiteral( "sample_image.png" ) );
+  mRasterFill->setWidthUnit( QgsUnitTypes::RenderPixels );
+  mRasterFill->setWidth( 30.0 );
+  mRasterFill->setCoordinateMode( QgsRasterFillSymbolLayer::Feature );
+  mRasterFill->setOpacity( 1.0 );
+  mRasterFill->setOffset( QPointF( 0, 0 ) );
+  mRasterFill->setOffsetUnit( QgsUnitTypes::RenderPixels );
```
nyalldawson, line 113: Can you leave out these changes? The intention with putting them in the ::init() method is that that method gets called before every test, so if an individual test messes with the mRasterFill definition then any changes it makes are reset when the next test method is called.

tests/src/core/testqgsrastermarker.cpp
```diff
@@ -68,35 +76,44 @@ class TestQgsRasterMarker : public QObject
...
 
-  //setup the raster marker symbol
+  // Setup the raster marker symbol
   mRasterMarker = new QgsRasterMarkerSymbolLayer();
+  mRasterMarker->setPath( mTestDataDir + QStringLiteral( "sample_image.png" ) );
+  mRasterMarker->setSizeUnit( QgsUnitTypes::RenderPixels );
+  mRasterMarker->setSize( 30.0 );
+  mRasterMarker->setHorizontalAnchorPoint( QgsMarkerSymbolLayer::HCenter );
+  mRasterMarker->setVerticalAnchorPoint( QgsMarkerSymbolLayer::VCenter );
+  mRasterMarker->setOffsetUnit( QgsUnitTypes::RenderPixels );
```
nyalldawson, line 112: Same here -- the original version was more robust to failing tests

nyalldawson (commented):
The symbology changes look good!

You'll need to revert most of the changes to the test suite though -- as noted above, they cause the tests to become less robust, and have actually broken some of the existing tests (as you can see by the need to change the reference images. You shouldn't have to change any of the existing reference images in this PR).

nyalldawson:
Nice work @beketata , and sorry about the delayed review! If you can fix the tests as noted above then this is good to merge.

tests/src/core/testqgsrasterfill.cpp
beketata: Do you mean that each new test should start with exactly the same source image defined in the ::init() ?

src/core/symbology/qgsfillsymbollayer.cpp
beketata: Did it.

tests/src/core/testqgsrasterfill.cpp
nyalldawson: It should start with new instances of the symbols. Just revert your changes, add the new test alone, and it should all work fine 😁

beketata: Done.

tests/src/core/testqgsrastermarker.cpp
beketata: Done.

nyalldawson (approved):

nyalldawson:
Nailed it -- thanks!

PR #35446, opened 2020-03-29
[FEATURE] Raster Layer Contour Renderer

wonder-sk:
This new renderer draws contour lines that are calculated on the fly from the source raster band. It is possible to set interval of contour lines and symbol used for drawing.

In addition there is support for "index contours" - contour lines with higher interval, typically drawn with a wider line symbol.

If we generate contour lines on input raster block with the same size as our output raster block, the generated lines would contain too much detail. This detail can be reduced by the "downscale" factor - this will request lower resolution of the source raster.

image


roya0045:
Looks awesome and will be quite useful!

 Though maybe the name  need some rework, they are not intuitive. Is the index contour the 0? Or is the offset specified otherwise? What is the distinction between contour and index contour? Functionally they just seem to be interchangeable.

Instead could there simply be contours that can be stacked? This would lead to a default one contour and more variance and spacing interval could be added with distinct offset.

src/core/raster/qgsrastercontourrenderer.cpp
```diff
@@ -0,0 +1,226 @@
...
+  std::unique_ptr< QgsRasterBlock > outputBlock( new QgsRasterBlock() );
+  if ( !mInput || !mContourSymbol )
+  {
+    return outputBlock.release();
+  }
+
+  int inputWidth = static_cast<int>( round( width / mDownscale ) );
+  int inputHeight = static_cast<int>( round( height / mDownscale ) );
+
+  std::shared_ptr< QgsRasterBlock > inputBlock( mInput->block( mInputBand, extent, inputWidth, inputHeight, feedback ) );
```
nyalldawson: unique_ptr would be more appropriate

src/core/raster/qgsrastercontourrenderer.cpp
```diff
@@ -0,0 +1,226 @@
...
+  {
+    QgsDebugMsg( QStringLiteral( "No raster data!" ) );
+    return outputBlock.release();
+  }
+
+  if ( !inputBlock->convert( Qgis::Float64 ) ) // contouring algorithm requires double
+    return outputBlock.release();
+  double *scanline = reinterpret_cast<double *>( inputBlock->bits() );
+
+  QImage img( width, height, QImage::Format_ARGB32_Premultiplied );
```
nyalldawson: Hmm this is kinda ugly -- it's revealing a shortcoming in the raster renderer. Does it work ok with print layouts and hi-dpi outputs? 

(A pretty-gross-but-api-conserving-fix might be to add the original render context to QgsRasterBlockFeedback)

nyalldawson (commented):
Nice!!

Great to see the re-use of QgsLineSymbol and the symbol button here!

nyalldawson:
> Though maybe the name need some rework, they are not intuitive. Is the index contour the 0? Or is the offset specified otherwise? What is the distinction between contour and index contour? Functionally they just seem to be interchangeable.

It's standard terminology:
https://www.google.com/search?q=index+contour

roya0045:
> It's standard terminology:
> https://www.google.com/search?q=index+contour
But in this case I would assume that the index contour might only be a suggestion as the index could have an interval of two and the other an interval of 10.

I would still think that having the option to stack the symbology(like rule based, class, etc) might be better if it can be implemented and would help to make figures similar to this if needed 
image

And this would provide some more flexibility. Though that might be a follow up.

nyalldawson:
That's non-trivial though. I agree it would be useful in a lot of circumstances, but would need a significant re-work of the raster renderer to allow "stacked" renderers.

I'd say for now your use case could be handled by duplicating the layer in the project multiple times and applying the symbology that way...

roya0045:
Possibly. I was just trying to provide useful feedback or suggestion for an improvement. The proposed work is useful enough as is!

 Though I still think that an offset option is important. If I have a raster that start at 2.5 and I want an isoline at 10m interval I feel like the current proposal would make lines at 2.5, 12.5, 22.5, etc. I would happily be wrong.

PeterPetrik:
nice, wondering if we can do the same for meshes...

src/core/raster/qgsrastercontourrenderer.cpp
```diff
@@ -0,0 +1,226 @@
...
+  Q_UNUSED( dfLevel )
+  ContourWriterData *crData = static_cast<ContourWriterData *>( ptr );
+  QPolygonF polygon( nPoints );
+  QPointF *d = polygon.data();
+  for ( int i = 0; i < nPoints; ++i )
+  {
+    d[i] = QPointF( padfX[i] * crData->scaleX, padfY[i] * crData->scaleY );
+  }
+
+  if ( crData->indexSymbol && !qgsDoubleNear( crData->indexInterval, 0 ) && qgsDoubleNear( fmod( dfLevel, crData->indexInterval ), 0 ) )
```
nirvn: @wonder-sk , any chance you could add an expression context variable that'd contain the height the line being drawn represents? That'd open the door to a lot of extra nice possibilities via data-defined properties.

nyalldawson: @nirvn it won't be much use unless we have access to the wider QgsRenderContext -- that's where the expression context is taken from

src/core/raster/qgsrastercontourrenderer.cpp
wonder-sk: Are you referring to the fact that all the rendered lines get rasterized? That is sub-optimal for print output unfortunately, but I can't see a way around it without rewriting a chunk of the raster pipeline :-) It would be a good exercise to make a list of things that are not ideal in the current raster pipeline implementation and see how we could change the design to fix them...

I have tried with a print layout and things look okay. I have seen some small discontinuities in contours at the edges of blocks created by QgsRasterDrawer, but that can be fixed at some point later I guess.

src/core/raster/qgsrastercontourrenderer.cpp
wonder-sk: true, not sure how that shared_ptr got in there ... :-)

src/core/raster/qgsrastercontourrenderer.cpp
wonder-sk: Adding height as a variable would be possible... As Nyall says you still wouldn't have access to the main render context, only a local one, but maybe that wouldn't stop you from drawing some fancy looking lines :) In any case, I would prefer to keep the scope of this PR limited, so maybe if you'd like to do a follow-up PR to add that? ;-)

wonder-sk:
@roya0045 Adding offset should be pretty easy - in the code that generates contours in qgsrastercontourrenderer.cpp there is "contourBase" variable which I left set to zero because I didn't have much use to set the offset - it's just a matter of adding it to GUI and reading/writing it in XML...

src/core/raster/qgsrastercontourrenderer.cpp
nirvn: Sure, if that's not something you're interested in implementing, someone else can step up. My schedule is hyper busy so likely not me. 

That said, you've walked us that far already, thanks :wink:

src/core/raster/qgsrastercontourrenderer.cpp
nyalldawson: Are you against my temperorary hack of adding a render context to the feedback?

wonder-sk: Do you mean to add render context to QgsRasterBlockFeedback? I guess I won't mind...

nyalldawson: Yeah. It's an ugly hack but avoids API breakage. For 4.0 we can make a QgsRasterBlockContext or similar to clean this up...

wonder-sk: Sounds good.

Raster block context would make a lot of sense, especially for raster renderers which needlessly create new temporary QImages just to be copied to the final render target. The world would become a better place if we won't do that :grin:  (easier to show intermediate results of raster rendering etc.)

roya0045:
@wonder-sk did you plan to add offset control in this PR ? Either a shared or distinct offset would be fine at the moment, as long as its available.

wonder-sk:
@roya0045 no plans for it currently... My preference is to get this initial work merged and let others add some more bells and whistles :-)

pigreco:
Please add `Symbology` label

PR #35447, opened 2020-03-29
[Tiles XYZ] Skip blank tiles

### LINKED ISSUE #35174: Skipping blank tiles when creating XYZ tiles
An option to skip blank tiles when generating xyz/mbtiles:
https://github.com/lutraconsulting/qgis-xyz-tiles/issues/4

javicasnov:
## Description

Allow to skip those tiles where all pixels match the defined background color. Based on @nurenda work in QTiles plugin.

Fixes qgis#35174


javicasnov:
Any comments / questions?

roya0045:
Is this supposed to make the xyztile test fail?

javicasnov:
The xyztile test is failing because the background color is not included in the test parameters and the default background color is not well interpreted. 

Only happens when the algorithm is executed from the console but works perfectly when the default background color is opaque and when it has the same value for R, G, B and A (ie: QColor(0,0,0,0) or QColor(Qt.transparent) or QColor(100,100,100,100)).

In this PR I changed the default background color to QColor(255,255,255,0) because QColor(Qt.transparent) caused a black background if you have a map with transparencies and select JPG tiles.

A workaround to make the test pass would be to define a background color (a transparent one) in the test, but I think that could be a bug, so I will report it

roya0045:
@javicasnov you could just replace the result of the test in this case.

PfisterFactor:
Why hasn’t this been considered for merge? It seems like it it fixes #35174 completely.

roya0045:
@PfisterFactor Because it was stale-out, probably due to the test not being fixed.

PR #35452, opened 2020-03-30
Port last remaining input parameters to new API

nyalldawson:
This PR ports the Raster Band input parameter and Multi Layer input parameter to the new C++ API.

Benefits include:
- expression based band/layer lists in models
- inline (non-blocking) lists for multi selection parameters (fields, bands, layer lists, enums)
- rearrangement of inputs for multi-layer parameters in models, to set specific layer orders


nyalldawson:
(This leaves only the "output" style parameters to be ported and then the epic journey is over)

alexbruy (approved):

PR #35458, opened 2020-03-30
Update temporal controller widget project time button tooltip

Samweli:


src/gui/qgstemporalcontrollerwidget.cpp
```diff
@@ -74,7 +74,10 @@ QgsTemporalControllerWidget::QgsTemporalControllerWidget( QWidget *parent )
     mEndDateTime->setDateTime( range.end() );
   }
 
-  mSetToProjectTimeButton->setToolTip( tr( "Set datetimes inputs to match project time" ) );
+  mSetToProjectTimeButton->setToolTip( tr( "Set datetimes inputs to match project time. \n"
```
nyalldawson: Hm - I'm concerned this string is a little complex for users. How about

    Match time range to project.
    If a project has no explicit time range set, then the range will be calculated based on the minimum and maximum dates from any temporal-enabled layers.

Samweli: Sure.

nyalldawson (approved):

PR #30227, opened 2019-06-14
[needs-docs] Ensure layer legend graphics are horizontally scrollable (fixes #28050)

dakcarto:
## Description
This fixes #28050 by allowing the Layers panel QTreeView to horizontally scroll.

layer-panel_horiz-scroll2

It also offers the following:
- Layer titles can be fully viewed (no elide).
- The width of the layer panel no longer needs resized via its splitter as often, reducing canvas redraws.
- When the tree view has many groups or nesting, it is simpler to scroll right than resize panel.

Even though the changes are small, the result may be confusing for some users, because users can end up with the following, even though layers are loaded:

layers-panel_scrolled-blank

To solve this, I tinkered with the following:
- Auto-reset horizontal scrollbar to position 0 after a delay, but not while user is still scrolling or has the input device pressed. While the reset worked, the timer could trigger at inopportune times.
- Add a scroll area to _just_ the WMS graphic. This proved to be tricky and not as simple/clean as scrolling the whole tree.

**Proposed horizontal scrollbar reset solution**
Assuming that just leaving it as is (user in control of scrolling) is not acceptable UX, a small button in toolbar or tall/thin button to the left of tree view could offer a quick means of reseting the horizontal scroll bar on one click.

However, I have left a reset button out of this PR, so it can be backported to `release-3_4` branch as an initial fix.

---

The other commit, to limit the max width of embedded legend widgets could also be tied to a user setting, though this would be a 'feature'.

## Checklist

> Reviewing is a process done by project maintainers, mostly on a volunteer basis. We try to keep the overhead as small as possible and appreciate if you help us to do so by completing the following items. Feel free to ask in a comment if you have troubles with any of them.

- [x] Commit messages are descriptive and explain the rationale for changes
- [x] Commits which fix bugs include `fixes #11111` in the commit message next to the description
- [ ] Commits which add new features are tagged with `[FEATURE]` in the commit message
- [ ] Commits which change the UI or existing user workflows are tagged with `[needs-docs]` in the commit message and contain sufficient information in the commit message to be documented
- [x] I have read the [QGIS Coding Standards](https://docs.qgis.org/testing/en/docs/developers_guide/codingstandards.html) and this PR complies with them
- [x] This PR passes all existing unit tests (test results will be reported by travis-ci after opening this PR)
- [ ] New unit tests have been added for core changes
[...]

PR #35461, opened 2020-03-30
[layers-panel] Changes to layer mark in layer tree view

### LINKED ISSUE #32029: layers panel shows strange artifact
**Describe the bug**

There's a strange grey artifact on the Layers panel.

image

First I thought it might be some new feature but it kinda shows also mixed with other layer symbols.

**How to Reproduce**

Just open some layer on QGIS.

**QGIS and OS versions**

QGIS version | 3.9.0-Master | QGIS code revision | 880de7f364
-- | -- | -- | --
PostgreSQL Client Version | 11.5 (Ubuntu 11.5-1.pgdg18.04+1) | SpatiaLite Version | 4.3.0a
PROJ.4 Version | 493
OS Version | Ubuntu 18.04.3 LTS

**Additional context**

gacarrillor:
## Description

PR to suggest 2 changes to layer mark in layers panel:

1. Update the viewport when scrolling horizontally to avoid the flicker.

2. Avoid showing the layer mark always and instead show it only when a layer/group icon is not visible anymore. 

**Before / After** 

before      ..........          after

**Before:** 
image 

**After:**
image

There are other options for suggestion number 2, like showing the layer mark: 
 
2.1 only when the horizontal scroll value is greater than the minimum (0); or
2.2 only when the layer/group text (including feature count) is not visible anymore. This one is a bit trickier!

Fix #32029

### DIFF
--- a/src/gui/layertree/qgslayertreeview.h
+++ b/src/gui/layertree/qgslayertreeview.h
@@ -237,6 +237,8 @@ class GUI_EXPORT QgsLayerTreeView : public QTreeView
 
   private slots:
     void onCustomPropertyChanged( QgsLayerTreeNode *node, const QString &key );
+    //! Handles updating the viewport to avoid flicker
+    void onHorizontalScroll( int value );
 
   protected:
     //! helper class with default actions. Lazily initialized.
--- a/src/gui/layertree/qgslayertreeview.cpp
+++ b/src/gui/layertree/qgslayertreeview.cpp
@@ -64,7 +64,7 @@ QgsLayerTreeView::QgsLayerTreeView( QWidget *parent )
   connect( this, &QTreeView::collapsed, this, &QgsLayerTreeView::updateExpandedStateToNode );
   connect( this, &QTreeView::expanded, this, &QgsLayerTreeView::updateExpandedStateToNode );
 
-  connect( this->horizontalScrollBar(), &QScrollBar::valueChanged, this, &QgsLayerTreeView::onHorizontalScroll );
+  connect( horizontalScrollBar(), &QScrollBar::valueChanged, this, &QgsLayerTreeView::onHorizontalScroll );
 }
 
 QgsLayerTreeView::~QgsLayerTreeView()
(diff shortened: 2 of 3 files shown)


nyalldawson:
@dakcarto?

src/gui/layertree/qgslayertreeview.h
```diff
@@ -234,6 +234,8 @@ class GUI_EXPORT QgsLayerTreeView : public QTreeView
     void onCurrentChanged();
     void onExpandedChanged( QgsLayerTreeNode *node, bool expanded );
     void onModelReset();
+    //! Handles updating the viewport to avoid flicker
+    void onHorizontalScroll( int value );
```
3nids, line 238: I think this should be in the private slots

3nids (commented):
1. nice!

2. I am not 100% due to misalignment then. No strong opinion here. Might be worth 2 distinct PRs.

gacarrillor:
@3nids Well, we could also implement it in this way to avoid misalignment: 

no_misalignment

src/gui/layertree/qgslayertreeview.h
gacarrillor: Fixed in https://github.com/qgis/QGIS/pull/35461/commits/6d1b8d6d063084e12168778f4b50e1df4c685c9d

gacarrillor:
Just rebased and applied @3nids comments.

m-kuhn (commented):
I'd like to reopen this one.
Resolving the graphical glitches is a good thing, no questions asked. About the layer mark I am not completely sure. To be honest only since I looked at this pull request right now (for about the fifth time) I think I really understood what it means :D -- and with "it" I refer to the layer mark, not the change in this PR.

gacarrillor:
Yes, the layer mark was not the easiest thing to understand for me neither, but I personally think it's a good idea, once all concerns from #32029 are addressed, which this PR does.

Just let me know how should we proceed (I'll apply the suggestion you already left above). 

It would be great if @nirvn, @nyalldawson and @dakcarto can give us an opinion on this PR, since they were involved in #30227, which this PR is attempting to polish.

nyalldawson (approved):

PR #35464, opened 2020-03-30
Sipify the project validator

Gustry:
## Description

In the project properties, QGIS Server tab, there is a QGIS Server check. It gives a HTML output about problems that you may have when publishing with QGIS Server.

We want to be able to run these checks from a Python plugin (Lizmap) to check the configuration before users send the project on the server.

After this PR, we will add more checks to this validator.

PS, I can also add later a `setShortName` on a group.

Please use the squash button when it will be merged.

CC @rldhont @dmarteau


src/core/qgsprojectservervalidator.cpp
```diff
@@ -0,0 +1,129 @@
...
+#include "qgsprojectservervalidator.h"
+#include "qgsvectorlayer.h"
+
+
+QString QgsProjectServerValidator::displayValidationError( QgsProjectServerValidator::ValidationError error )
+{
+  switch ( error )
+  {
+    case QgsProjectServerValidator::Encoding:
+      return QObject::tr( "Encoding is not set properly. It shouldn't be 'System'" );
```
nyalldawson: "It shouldn't be 'System'" doesn't sound professional. Maybe "An explicit, non "System" encoding is required"

src/core/qgsprojectservervalidator.cpp
```diff
@@ -0,0 +1,129 @@
...
+    case QgsProjectServerValidator::ShortNames:
+      return QObject::tr( "Layer short name needs to be updated" );
```
nyalldawson: What exactly does this mean? Does it mean it wasn't set originally? Or clashes? A more descriptive string will help users know exactly what this means and how to rectify

src/core/qgsprojectservervalidator.cpp
```diff
@@ -0,0 +1,129 @@
...
+    case QgsProjectServerValidator::DuplicatedNames:
+      return QObject::tr( "Layers/groups have the same name or short name" );
```
nyalldawson:
```suggestion
      return QObject::tr( "One or more layers or groups have the same name or short name. Both the 'name' and 'short name' for layers and groups needs to be unique." );
```

src/core/qgsprojectservervalidator.h
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Validates a layer tree to avoid some problems on QGIS Server, and returns TRUE if it's considered valid.
+     * If validation fails, the \a results list will be filled with a list of
+     * items describing why the validation failed and what needs to be rectified
+     * \param layerTree input layer tree
+     * \param results results of the validation
+     * \returns bool
+     */
+    bool validate( QgsLayerTree *layerTree, QList< QgsProjectServerValidator::ValidationResult > &results SIP_OUT ) const;
```
nyalldawson: can be made static

Gustry:
@nyalldawson Thanks for your inputs. I have fixed them locally. I will come back shortly on this PR after talking with my colleagues.

PR #35467, opened 2020-03-30
Add "protoc" as a build dependency to support MVT vector tiles

wonder-sk:
It turns out it is not a good idea to keep generated files from protobuf compiler (protoc) in the repository because different versions of generated cpp/h files may not be compatible with different versions of runtime library. So we will use "protoc" tool to generate the files during build time for the source .proto file - like this one: https://github.com/mapbox/vector-tile-spec/blob/master/2.1/vector_tile.proto

### DIFF
--- a/.docker/qgis3-build-deps.dockerfile
+++ b/.docker/qgis3-build-deps.dockerfile
@@ -64,6 +64,7 @@ RUN  apt-get update \
     pkg-config \
     poppler-utils \
     postgresql-client \
+    protobuf-compiler \
     pyqt5-dev \
     pyqt5-dev-tools \
     pyqt5.qsci-dev \


nyalldawson (approved):

roya0045:
@wonder-sk is it possible that this or the other one related to protobuf breaks azure pipelines? I just rebased one of my PR and now it wont build due to the missing protobuf library.

roya0045:
See https://github.com/roya0045/QGIS_1/runs/557952245?check_suite_focus=true

elpaso:
@roya0045 in case someone is so nice to fix it, the recipe is: https://github.com/qgis/QGIS/blob/master/ms-windows/mxe/mxe.Dockerfile , apparently protobuf package is already available in mxe repos.

roya0045:
@elpaso I'm on it

roya0045:
Ok maybe I'm just bad at this but I tried to add what was added in the apt-get install line, nothing worked,(BTW MXE only has protobuf instead of protobuf-dev and protobuf-compiler) and I also can't build on windows because installing protobuf-devel with cygwin doesn't help.

Even with  'protobuf libprotobuf protobuf-dev protobuf-compiler', it doesn't change a thing.

roya0045:
@elpaso is it possible that by fetching your docker image it doesn't execute my modified file but yours? I tried to add an echo to debug and I didn't see it in the log.

elpaso:
> @elpaso is it possible that by fetching your docker image it doesn't execute my modified file but yours? I tried to add an echo to debug and I didn't see it in the log.

That's for sure, here is how you should proceed:

1. add protobuf to the docker recipe, you can find the available packages at https://mxe.cc/#packages
2. build the docker locally and make sure that it build correctly
3. cross your fingers and test the QGIS build scripts using your locally built docker 
4. either docker pull your image and change the build scripts to your new image or ping me and I can build and push the image for you

roya0045:
@elpaso I have added the package in the script but it did not change a thing when building. Though I have to admit that I am not equipped to build a docker locally with my work equipment. If you could 
 build and update the image this would be appreciated. I don't have enough space or rights on the machines available to me to set up Docker.

PR #35468, opened 2020-03-30
Vertex tool should not select invisible features

### LINKED ISSUE #25259: Node Tool Selects Invisible Objects
Author Name: **Ryan Desmond** (@rddesmond)
Original Redmine Issue: [17362](https://issues.qgis.org/issues/17362)
Affected QGIS version: master
Redmine category:editing

---
The node tool, selection happens without regard to the visibility of the object.  In the simple case (procedure 1), you can simply set the layer to invisible and show that object selection is still possible.  In a more complex case (procedure 2), a portion of a layer is made invisible (with categorized style rules).

With the existing node tool, procedure one is a bit unnatural, because you can only select features on the currently selected layer.  The second procedure is a much more natural use case. 
 One or both of these cases may be OBE by the new node tool in 3.0.

## Procedure 1:

1. Create a polygon scratch layer
1. Edit the layer
1. Select add feature and insert an object
1. Select the "Node Tool"
1. Toggle layer visibility to hide it
1. Click where a node would be on the newly hidden object

## Procedure 2:

1. Create a polygon scratch layer
1. Edit the layer
1. Select add feature and insert two objects
1. Select "Open Attribute table"
1. Add a new column and categorize each of the objects uniquely (e.g. 1, 2, ..)
1. Select the "Node Tool"
1. In Properties / Style, set the style to be categorized by the new column and hit "classify"
1. Unselect one of the categories
1. Click where a node would be on the newly hidden object

## Expected

Visible objects are selectable, but invisible objects are not.

## Actual

In both cases, the objects which aren't visible are selected to edit.

uclaros:
## Description
This is how it's done in `QgsMapToolSelectUtils` for `QgsMapToolSelect`.
Is this a good way to handle this?

Fixes: #25259

### DIFF
--- a/src/app/vertextool/qgsvertextool.cpp
+++ b/src/app/vertextool/qgsvertextool.cpp
@@ -516,7 +516,7 @@ void QgsVertexTool::cadCanvasReleaseEvent( QgsMapMouseEvent *e )
     // only handling of selection rect being dragged
     QList<Vertex> vertices;
     QList<Vertex> selectedVertices;
-    bool invisibleVertexSelected = false;
+    bool showInvisibleFeatureWarning = false;
 
     QgsGeometry rubberBandGeometry = mSelectionRubberBand->asGeometry();
 
@@ -577,9 +577,11 @@ void QgsVertexTool::cadCanvasReleaseEvent( QgsMapMouseEvent *e )
           continue;  // with locked feature we only allow selection of its vertices
 
         context.expressionContext().setFeature( f );
-        // make sure to only use features that are visible
         bool isFeatureInvisible = ( r && !r->willRenderFeature( f, context ) );
-        if ( isFeatureInvisible && invisibleVertexSelected )
+        // If we've already determined that we have to show users a warning about selecting invisible vertices,
+        // then we don't need to check through the vertices for other invisible features.
+        // We'll be showing the warning anyway regardless of the outcome!
+        if ( isFeatureInvisible && showInvisibleFeatureWarning )
           continue;
 
         bool isFeatureSelected = vlayer->selectedFeatureIds().contains( f.id() );
@@ -593,7 +595,7 @@ void QgsVertexTool::cadCanvasReleaseEvent( QgsMapMouseEvent *e )
             // but if the user tried to selected one we want to warn him
             if ( isFeatureInvisible )
             {
-              invisibleVertexSelected = true;
+              showInvisibleFeatureWarning = true;
               break;
             }
             vertices << Vertex( vlayer, f.id(), i );
@@ -606,7 +608,7 @@ void QgsVertexTool::cadCanvasReleaseEvent( QgsMapMouseEvent *e )
       if ( r )
         r->stopRender( context );
     }
-    if ( invisibleVertexSelected )
+    if ( showInvisibleFeatureWarning )
     {
       QgisApp::instance()->messageBar()->pushMessage(
         tr( "Invisible vertices were not selected" ),
--- a/src/app/vertextool/qgsvertextool.cpp
+++ b/src/app/vertextool/qgsvertextool.cpp
@@ -39,6 +39,7 @@
 #include "qgsvertexeditor.h"
 #include "qgsmapmouseevent.h"
 #include "qgsexpressioncontextutils.h"
+#include "qgsmessagebar.h"
 
 #include <QMenu>
 #include <QRubberBand>
@@ -515,6 +516,7 @@ void QgsVertexTool::cadCanvasReleaseEvent( QgsMapMouseEvent *e )
     // only handling of selection rect being dragged
     QList<Vertex> vertices;
     QList<Vertex> selectedVertices;
+    bool invisibleVertexSelected = false;
 
     QgsGeometry rubberBandGeometry = mSelectionRubberBand->asGeometry();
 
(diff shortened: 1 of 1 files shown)


roya0045:
If this work this will be great! How does it deal with rule based features?

uclaros:
Since it's using a copy of the layer's renderer, everything that's not rendered should not be selectable.

roya0045:
I would think that a gif of the modification (since a test seems unlikely) would be beneficial since there are no way to see this in action other than to build ourselves. But the changes seems reasonable.

saberraz:
I have tested your improvements but I am not 100% sure with this use-case:
Imagine, you are in topological edit mode, if a feature is hidden, and you try to use the node tool, it does not edit the hidden tool. I think for topological edit mode, this should work on all features and not only visible ones:

You can see in the screenshot below, the footpath is hidden (but visible in the 2nd map view). When I do node move for the main road, it works with all vertices. But, for footpath, it will leave it untouched.  I understand, your new logic to only apply to the visible features is more handy, but this might mess up a lot of data model, where they need to have the correct topology but only show the subset to user to edit:
edit

uclaros:
> they need to have the correct topology but only show the subset to user to edit:

I hear your concern, but is it really safe to edit invisible data? Moving vertices with topological editing is kind of safe (you may introduce unwanted intersections), but deleting is not. What if the user deletes the seemingly unneeded vertex (a vertex in the middle of a straight segment) you moved? You also lose a segment of your footpath! With polygons it can get worse!
A possible solution would be to use an invisible style, eg.transparent, for the invisible features.

roya0045:
I have to agree with ucarlos, what I dislike the most is having features from other layer act as editable feature when the layer is not displayed (though this may be fixed). With overlapping feature is can be nearly impossible to edit the desired feature.

uclaros:
@saberraz I just tested on master and while the invisible vertices can be selected and moved, they are *not* moved by topological editing. This PR should affect only the drag-select functionality, ie. the invisible line's vertices should not be selected:

Peek 2020-04-16 23-29

uclaros:
This PR is ready for review :)

src/app/vertextool/qgsvertextool.cpp
```diff
@@ -545,17 +546,39 @@ void QgsVertexTool::cadCanvasReleaseEvent( QgsMapMouseEvent *e )
...
+
       QgsRectangle layerRect = layerRubberBandGeometry.boundingBox();
       std::unique_ptr< QgsGeometryEngine > layerRubberBandEngine( QgsGeometry::createGeometryEngine( layerRubberBandGeometry.constGet() ) );
       layerRubberBandEngine->prepareGeometry();
 
+      QgsFeatureRequest request;
+      request.setFilterRect( layerRect );
+      request.setFlags( QgsFeatureRequest::ExactIntersect );
+      if ( r )
+        request.setSubsetOfAttributes( r->usedAttributes( context ), vlayer->fields() );
```
nyalldawson, line 566:
```suggestion
    {
        request.setSubsetOfAttributes( r->usedAttributes( context ), vlayer->fields() );
        request.setFilterExpression( r->filter( vlayer->fields() ) );
     }
```

nyalldawson:
We could also consider adding a warning message if invisible features were selected which would have otherwise been modified. (If we do that, you'll need to ignore the above suggestion!)

src/app/vertextool/qgsvertextool.cpp
```diff
@@ -545,31 +548,71 @@ void QgsVertexTool::cadCanvasReleaseEvent( QgsMapMouseEvent *e )
...
       while ( fi.nextFeature( f ) )
       {
         if ( mLockedFeature && mLockedFeature->featureId() != f.id() )
           continue;  // with locked feature we only allow selection of its vertices
 
+        context.expressionContext().setFeature( f );
+        // make sure to only use features that are visible
+        bool isFeatureInvisible = ( r && !r->willRenderFeature( f, context ) );
+        if ( isFeatureInvisible && invisibleVertexSelected )
+          continue;
```
nyalldawson, line 583: It took me a while to understand this logic. Maybe a comment like:

 // If this feature is not visible, and we've already determined that we have to show users a warning about selecting
 // invisible vertices, then we don't need to check through the vertices for this feature. We'll be showing the warning anyway regardless of the outcome!

uclaros: OK! Also renamed `invisibleVertexSelected' to `showInvisibleFeatureWarning`. I think the intention is clearer now!

uclaros:
Any suggestion for the warning message wording as a native speaker?

src/app/vertextool/qgsvertextool.cpp
nyalldawson: Much better, thanks!

nyalldawson:
Hmm. maybe "One or more selected vertices are not currently visible on the map. These will not be modified."

uclaros:
It's not that they will not be modified, it's actually that they will not be selected at all though...
I think the concept of *symbology classes* should be in the message somehow.

roya0045:
@uclaros does that mean that if you selected vertices on a layer, zoom in a move them, they will be deselected or unmoved?

uclaros:
@roya0045 no, the logic is simple and intuitive and is as follows:
When your symbology excludes some features, either because there is no matching class for it (eg delete the *all other values* class), or because a class's visibility is unchecked from the layers panel, then those features' vertices cannot be selected when dragging a rectangle with the vertex tool (just like they already cannot be selected by hover and click with the vertex tool). If however the drag rectangle contains (topologically) vertices from such features, a warning is issued stating that invisible vertices were not selected. Once the selection is done, everything works as before.

PR #35470, opened 2020-03-30
Add missing temporal properties from wmst layers when using data source manager

Samweli:
Enables temporal properties and temporal capabilities for the wmst layers which has been added to the layers panel by using the data source manager.

Example
added_missing_wmst_temporal_properties

### DIFF
--- a/src/providers/wms/qgswmssourceselect.cpp
+++ b/src/providers/wms/qgswmssourceselect.cpp
@@ -1027,7 +1027,7 @@ void QgsWMSSourceSelect::collectDimensions( QStringList &layers, QgsDataSourceUr
 {
   for ( const QgsWmsLayerProperty layerProperty : mLayerProperties )
   {
-    if ( layerProperty.name == layers.join( QStringLiteral( "," ) ) )
+    if ( layerProperty.name == layers.join( ',' ) )
     {
       // Check for layer dimensions
       for ( const QgsWmsDimensionProperty &dimension : qgis::as_const( layerProperty.dimensions ) )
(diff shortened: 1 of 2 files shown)


src/providers/wms/qgswmssourceselect.cpp
```diff
@@ -1020,6 +1023,39 @@ void QgsWMSSourceSelect::collectSelectedLayers( QStringList &layers, QStringList
   }
 }
 
+void QgsWMSSourceSelect::collectDimensions( QStringList &layers, QgsDataSourceUri &uri )
+{
+  for ( const QgsWmsLayerProperty layerProperty : mLayerProperties )
+  {
+    if ( layerProperty.name == layers.join( QStringLiteral( "," ) ) )
```
nyalldawson:
```suggestion
    if ( layerProperty.name == layers.join( ',' ) )
```

The character based overrides for string operations are MUCH MUCH faster then the QString versions

src/providers/wms/qgswmssourceselect.cpp
```diff
@@ -1020,6 +1023,39 @@ void QgsWMSSourceSelect::collectSelectedLayers( QStringList &layers, QStringList
...
+    {
+      // Check for layer dimensions
+      for ( const QgsWmsDimensionProperty &dimension : qgis::as_const( layerProperty.dimensions ) )
+      {
+        // add temporal dimensions only
+        if ( dimension.name == QLatin1String( "time" ) ||
+             dimension.name == QLatin1String( "reference_time" ) )
+        {
+          QString name = dimension.name == QLatin1String( "time" ) ?
+                         QString( "timeDimensionExtent" ) : QString( "referenceTimeDimensionExtent" );
```
nyalldawson:
```suggestion
                         QStringLiteral( "timeDimensionExtent" ) : QStringLiteral( "referenceTimeDimensionExtent" );
```

(apparently qt 6 is going to fix this mess and avoid the confusing need for QStringLiteral and QLatin1String. Fingers crossed!!)

Samweli: That will help for sure!

nyalldawson (approved):

PR #35478, opened 2020-03-31
[layouts] Allow multiline header strings

nirvn:
## Description

Once upon a time, layout tables' headers were attribute field titles. With manual/fixed tables, the user is in control of the header string, and he/she might want to have a multiline header string. This PR unlocks that.


nyalldawson:
Looks good! Would it be possible to also put th wrap code in there so longer strings are auto wrapped just like the contents?

PR #35486, opened 2020-03-31
Install jinja and pygments python3 modules in qgis docker

### LINKED ISSUE #35480: Docker image qgis:final-3_10_4_disco: Couldn't load plugin 'MetaSearch' (various python modules missing)
The docker image for qgis:final-3_10_4_disco, when started, fails to load MetaSearch plugin due to lack of `jinja2` python module on the system. Sounds like something the official image should include ?

strk:
They are needed for default-enabled MetaSearch plugin.
Fixes #35480

### DIFF
--- a/.docker/qgis.dockerfile
+++ b/.docker/qgis.dockerfile
@@ -1,7 +1,7 @@
 
 # see https://docs.docker.com/docker-cloud/builds/advanced/
 # using ARG in FROM requires min v17.05.0-ce
-ARG DOCKER_TAG=latest
+ARG DOCKER_TAG=release-3_10
 
 FROM  qgis/qgis3-build-deps:${DOCKER_TAG} AS BUILDER
 MAINTAINER Denis Rouzaud <denis@opengis.ch>
--- a/.docker/qgis.dockerfile
+++ b/.docker/qgis.dockerfile
@@ -82,3 +82,7 @@ WORKDIR /
 
 # Run supervisor
 CMD ["/usr/bin/supervisord", "-c", "/etc/supervisor/supervisord.conf"]
+
+# Install python modules required to load automatically loaded
+# python plugins
+RUN pip3 install jinja2 pygments


.docker/qgis.dockerfile
```diff
@@ -1,7 +1,7 @@
 
 # see https://docs.docker.com/docker-cloud/builds/advanced/
 # using ARG in FROM requires min v17.05.0-ce
-ARG DOCKER_TAG=latest
+ARG DOCKER_TAG=release-3_10
```
3nids, line 4: why?

.docker/qgis.dockerfile
```diff
@@ -82,3 +82,7 @@ WORKDIR /
 
 # Run supervisor
 CMD ["/usr/bin/supervisord", "-c", "/etc/supervisor/supervisord.conf"]
+
+# Install python modules required to load automatically loaded
+# python plugins
+RUN pip3 install jinja2 pygments
```
3nids, line 88: up to now other Python dependencies were installed in the dependency image.
Are you suggesting to split the dependencies used for building than the ones used at run-time?
if so, can you add a TODO in the deps images?

.docker/qgis.dockerfile
strk: Just because this is in the release-3_10 branch so it's odd that it should ever be "latest" ? Can omit this commit if you prefer.

.docker/qgis.dockerfile
strk: I didn't think about it at all, but as the "dependencies image" is indeed called "build-deps" and documented to contain all dependencies required to build, it doesn't make sense to me to add there anything which is NOT needed for building qgis.

I'm not familiar with the deps dockerfile (I see there are 2 of them?) nor I know which of those deps are runtime rather than build-time deps. How about just filing an issue, if you know which deps are in the wrong place, to  move them ?

3nids: OK fair enough.
Just the RUN command must be be placed before the CMD command.

strk:
When doing docker I took the habit to always append, to avoid useless rebuilds..(we might need to add more in the future). As you're the maintainer, how about moving it wherever you prefer ? I would not want to further have it to change it to be before WORKDIR for example...

3nids:
> how about moving it wherever you prefer ?

ok, done here #35632

strk:
Thanks Denis!

PR #35499, opened 2020-04-01
Set a better size adjustment policy for our map layer combo box widget

nirvn:
## Description

By default, Qt combo boxes aren't allowed to get smaller than the largest item text width, which can create overly wide areas.

Among other things, this translates into very wide processing algorithm panels when a map layer parameter is present and a given project has a very long layer name.

This PR updates the sizing adjustment policy of our map layer combo box widget to allow for smaller-than-widest-text resizing.


nirvn:
Before vs. PR:
image

Here's how the combo box crops text out:
Screenshot from 2020-04-01 13-45-23

nyalldawson:
Nice -- that'll help the style dock too!

PR #35515, opened 2020-04-01
Fix use of enum flags in older python/sip versions

nyalldawson:


3nids:
this is such a pity, the docs of scoped enums just look much much nicer.

nyalldawson:
We'll have to defer till we can bump the sip/python version requirement...

m-kuhn:
> We'll have to defer till we can bump the sip/python version requirement...

to?

m-kuhn:
@nyalldawson do you have more specific information about the system on which it failed?

nyalldawson:
It fails on travis, works fine locally (fedora 31). I'll post version numbers tomorrow

Issue #28925, opened 2019-01-26
Print Composer: Lock layers + style and Follow map theme


qgib:
Author Name: **Casper Bertelsen** (Casper Bertelsen)
Original Redmine Issue: [21107](https://issues.qgis.org/issues/21107)
Affected QGIS version: 3.4.4
Redmine category:map_composer/printing

---
I have two issues that sort of revolves around the same thing, that is when I want to show specific layers in my map windows in the print composer.
I've tried both using the setting *Follow map theme* and *Lock layers* combined with *Lock styles for layers*, because I usually have more than one style linked to one layer.

Example:
Style 1: Showing points as stars
         No scale-dependent visibility

Style 2: Showing points as circles
         Scale-dependent visibility (1:1000)

My composer contains a map and a legend. The composer is set up to build and atlas (Two polygons. One showing at 1:8000, One at 1:500)

## Issue using *Follow map theme*:

My map follows map theme Style 1, showing stars at any scale. Works fine until I change the theme in QGIS main window to Style 2, showing circles up to 1:1000.
Now my print composer show stars on the polygon in scale 1:500, but no stars on the polygon 1:8000.
The legend has also updated to show points as circles, even though the map shows stars.

## Issue using *Lock layers*:

After that I tried to lock layers and styles instead. Now the legend will show points as stars even though I change to theme (and style) in the main windows to show circles. Just as I would expect. However when I change to style 2 in the main windows, no points will be shown in 1:8000, thus using the scale-dependency of the current style instead of what the style locked on i set to.

I've attached a project and data following the example.

---

- [Print_composer_diff_style.zip](https://issues.qgis.org/attachments/download/14183/Print_composer_diff_style.zip) (Casper Bertelsen)

Issue #28919, opened 2019-01-25
Legend content issue when using atlas and themes


qgib:
Author Name: **Alexis Roy-L** (Alexis Roy-L)
Original Redmine Issue: [21101](https://issues.qgis.org/issues/21101)
Affected QGIS version: 3.4.4
Redmine category:map_legend

---
Greetings,

Here is the behavior I noticed:

When using themes and an atlas to generate automatically maps, the current selected sub item in the map will be used rather than the setting of the theme.

IE: If I have a layer with 7 classes and I want one of them to be displayed for 2 themes and all the classes for 1 other theme.

The legend will display the classes that are currently activated in the map regardless of what is displayed by the theme on the canvas.

This mean that I must set my map to match the theme of the canvas in order to display the proper categories in the legend, otherwise only one, none or all the classes for a given layer will be displayed depending on the current map theme and regardless of the canvas theme.

Thanks and have a nice day!

klakar:
Probably the same problem:

When creating themes with different styles and adding several map items to a layout. The legends does not follow the map theme. Even if the correct map is set as source for the legend.
![inflyttning](https://user-images.githubusercontent.com/6375959/63250339-b4ac7100-c26b-11e9-98bc-02b654af8d4a.jpg)
In the picture above is looks ok, but the interval for the colors are different between the maps.
There's no way to create a product with the correct legend from layout. It's possible to show it on screen, if you add the legend with the correct theme shown in the canvas. But as soon as you move any legend, print or make an export. All legends is reset to follow the current theme in the canvas.

Issue #24694, opened 2017-07-05
[Composer Legend] Symbology of layer always follows the map canvas and breaks the composer legend


qgib:
Author Name: **Harrissou Santanna** (@DelazJ)
Original Redmine Issue: [16795](https://issues.qgis.org/issues/16795)
Affected QGIS version: 3.5(master)
Redmine category:map_composer/printing

---
Create two different styles for a single layer.
Now in print composer, do a layout and add the layer to a composer legend.
Go to the map canvas and activate the other style of the layer. In your composer legend, you'll notice that the layer legend has changed and follows what is shown in the map canvas and not the style you have set at the beginning (which should be in line with your map).
Note that if you switch the map canvas to a preset visibility that contains the layer with the first style, the style is well rendered in the legend.
If you had switched to a preset visibility that doesn't use the layer, no change is done in the legend either.

This state can easily corrupt the legend of the layout unless you check each print composer (actually update map canvas with correct preset visibility) before exporting (to ensure the legend shown is in line with the maps in composer) --> time consuming, prone to error (and discourages to use multiple styles for a single layer).

I think that legend of the layer should be bound to the style it has been added with to the composer, and follow the changes in that style. But it shouldn't switch between styles, according to what's enabled or not in the map canvas.

======================
Ouch! I'm rewording given that the description above is just ununderstandable. Sorry, devs. Let's do it simple :-)

In few words, the symbol shown in a legend item follows the style of the layer in the map canvas and not the styled used in the map item: it uses the style applied to the layer the last time it was visible.

In a project where multiple styles are used for a single layer, this becomes problematic.

Still applies to 2.18 and master 3.1

---

Related issue(s): #26620 (duplicates), #27542 (duplicates), #27877 (relates)
Redmine related issue(s): [18733](https://issues.qgis.org/issues/18733), [19717](https://issues.qgis.org/issues/19717), [20055](https://issues.qgis.org/issues/20055)

---

PR #35519, opened 2020-04-02
Correctly render legends linked to a map with a theme set

### LINKED ISSUE #27542: Legend to reflect map theme in layouts/ reports
Author Name: **Verity Miles** (Verity Miles)
Original Redmine Issue: [19717](https://issues.qgis.org/issues/19717)

---
It would be great to have a map legend reflect a map theme in the same way that the map itself does in a print layout/ report.

I am trying to set up a report in QGIS 3.2 to export a whole series of maps. At the moment, I cannot alter the style of the legend (we have different styles for the one layer depending on the map requirements) without manually changing the style/ map theme in the map viewer window. This effectively means that I have to manually export the series of maps.

Thanks.

---

Related issue(s): #24694 (duplicates)
Redmine related issue(s): [16795](https://issues.qgis.org/issues/16795)

---

nyalldawson:
If a legend is linked to a map which is set to show a particular theme, then correctly follow that same theme's styling when rendering the legend.

Fixes layout legends always render using the canvas' visible theme styling, instead of that of the linked map.

Fixes #27542, #24694, #28919, #28925

Fix sponsored by the Victorian Planning Authority

Peek 2020-04-02 14-23


nirvn:
Thanks good.

DelazJ:
> Fixes #27542, #24694, #28919, #28925

Thanks Nyall. Afaik each issue report needs its own "fixes" mention, otherwise only the first one (#27452) will be closed when this PR is merged. See eg
image
VS
image

nyalldawson:
@roya0045 that should be working -- the unit test includes a case where a style has a different set of nodes in a theme. But if you want to pull this branch and test, it'd be very welcome!

src/core/layout/qgslayoutitemlegend.cpp
```diff
@@ -821,6 +875,11 @@ void QgsLayoutItemLegend::doUpdateFilterByMap()
   mForceResize = true;
 }
 
+QString QgsLayoutItemLegend::themeName() const
+{
+  return mThemeName;
```
nirvn, line 880: @nyalldawson , any reason why this isn't simply in the header file?

nirvn (approved):

src/core/layout/qgslayoutitemlegend.cpp
nyalldawson: There's no real benefit to it being there -- this isn't performance critical code

nirvn: Okeydokey. Merge away! :)

roya0045:
@nyalldawson In this case, I'll rebase my branch with my PRs in limbo to see of that solves the issue.

roya0045:
OK MXE and my window build is bjorked with the protobuf requirement. I won't be able to test soon.

PR #35547, opened 2020-04-03
[virtual] Fix creation of "_search_frame_" column (fixes #31356)

### LINKED ISSUE #31356: Virtual Layer SQLite Natural Join - Instant Crash
**Describe the bug**
My understanding is that QGIS is SQLite3 compliant, and as such, should be compatible with SQLite3 syntax and functions.

**How to Reproduce**
image
Create new Virtual Layer
Import needed layers (One MultiLineString, one Table)
Write SQLite Code:
SELECT * FROM "CAMP-RS_JobsList_XLS"
NATURAL JOIN "Reconciled"

Test & get valid SQLite code confirmation
Click Add and QGIS crashes instantly to desktop

**QGIS and OS versions**

QGIS version | 3.8.2-Zanzibar | QGIS code revision | 4470baa1a3
-- | -- | -- | --
Compiled against GEOS | 3.7.2-CAPI-1.11.0 | Running against GEOS | 3.7.2-CAPI-1.11.0 b55d2125
PostgreSQL Client Version | 10.8 | SpatiaLite Version | 4.3.0
OS Version | Windows 10 (10.0) |   |

mhugo:
The "_search_frame_" hidden column should only be added for layers with geometries.
It was the first column of the virtual table, it is now the last

Should fix #31356

### DIFF
--- a/src/providers/virtual/qgsvirtuallayersqlitemodule.cpp
+++ b/src/providers/virtual/qgsvirtuallayersqlitemodule.cpp
@@ -195,9 +195,6 @@ struct VTable
       mFields = mLayer ? mLayer->fields() : mProvider->fields();
       QStringList sqlFields;
 
-      // add a hidden field for rtree filtering
-      sqlFields << QStringLiteral( "_search_frame_ HIDDEN BLOB" );
-
       const auto constMFields = mFields;
       for ( const QgsField &field : constMFields )
       {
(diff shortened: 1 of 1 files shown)


mhugo:
More explanations about the bug encountered with natural joins: natural joins are implemented by looking for common column names between two tables. The "_search_frame_" hidden column was added for all virtual tables, even for geometry-less layers, which did not really make sense.
And so "_search_frame_" was used in the join columns, with a computation of a bbox for a null geometry (which was not supposed to be null)

PR #35554, opened 2020-04-03
Filter out field with readonly in the widget. Fix #34331

### LINKED ISSUE #34331: Attribute table allows editing non-editable fields through quick field calculator
If a field is set as not-editable (by unchecking the corresponding check box in the field widget configuration), values in that field can still be modified by using the quick field calculator in the attributes table dialog.

To avoid this, fields that are not editable should not be shown in the dropdown list of the quick field calculator.

Some hints:

Looks like the attributes table takes into account if the fields are set as editable or not, but the dropdown list for the quick field calculator is ignoring that.

The dropdown list uses the QgsFieldComboBox custom widget, and is correctly passing the HideReadOnly filter. However, the QgsFieldProxyModel class just classifies as read-only those fields that are hidden join fields, and for all fields in the original data, it considers them editable

See: https://github.com/qgis/QGIS/blob/master/src/core/qgsfieldproxymodel.cpp#L66

Looks like, instead of returning always false in this case, it should use the `editFormConfig().readOnly(idx)` method from the layer to decide if the cell is read-only, just like the attributes table delegate does.

ismailsunni:
## Description

I made another flag to make a field that is not editable through widget not showing up in the attribute table to prevent it gets updated. Example:
field `nama` is not editable
Selection_360
field `nama` is not showing up in the combo box.
gedung :: Features Total: 1, Filtered: 1, Selected: 0_361

### DIFF
--- a/python/core/auto_generated/qgsfieldmodel.sip.in
+++ b/python/core/auto_generated/qgsfieldmodel.sip.in
@@ -38,6 +38,7 @@ It can be associated with a QgsMapLayerModel to dynamically display a layer and 
       IsEmptyRole,
       EditorWidgetType,
       JoinedFieldIsEditable,
+      FieldIsWidgetEditable,
     };
 
     explicit QgsFieldModel( QObject *parent /TransferThis/ = 0 );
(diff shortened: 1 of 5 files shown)


m-kuhn (approved):

ismailsunni:
@Saijin-Naib perhaps just write it, if I had some extra time, I will check :)

PR #35557, opened 2020-04-03
Defer mask source widget population (fixes #34942)

### LINKED ISSUE #34942: In project with many category symbols in many layers, Layer Properties dialog takes very long time to open for vector layers
**Describe the bug**

In QGIS 3.12 and beyond, in a project with many symbology categories in many vector layers, where the mathematical product of 'No. of Layers' * 'No. of categories' = a very high number, the Layer Properties dialog for vector layers is taking a very long time to open.

This does not seem to affect the Layer Properties dialog for raster layers, but it affects all of the vector layers. So far (this description edited 25th June 2020), the bug fixing discussion and process has found that the issue is due to the introduction of Selective Masking in QGIS 3.12. A partial fix has been implemented, involving the 'Edit Mask' checkbox, but there is still a substantial delay in the sample file, even with this box unchecked.

**How to Reproduce**

[Dummy_Data.zip](https://github.com/qgis/QGIS/files/4830777/Dummy_Data.zip)

1. Open the dummy_project.qgz file which includes 64 duplicate layers, each with 235 categories (= 15040).
2. In Layers Panel, double-click on a vector layer, or right-click and select 'Properties'.
3. Record the time to open the Layer Properties dialog
expected times:
QGIS 3.10 < 2 seconds
QGIS 3.12 37 seconds with Edit Mask unchecked
QGIS 3.12 83 seconds with Edit Mask checked
(comparison of Edit Mask checked and unchecked was on different machines, so it may be inaccurate)
4. Open a project file that contains only a few vector layers with not many symbology categories.
5. Repeat steps 2 and 3 for the smaller project file.
6. Compare the times recorded to open the dialog in the smaller and larger project files.
7. If the Layer Properties dialog takes much longer to open in the larger project than the smaller one in QGIS 3.12 (but no difference in QGIS 3.10), then you will have reproduced this bug.

**QGIS and OS versions**

QGIS version
3.12.0-București
QGIS code revision
cd141490ec
Compiled against Qt
5.11.2
Running against Qt
5.11.2
Compiled against GDAL/OGR
3.0.4
Running against GDAL/OGR
3.0.4
Compiled against GEOS
3.8.0-CAPI-1.13.1
Running against GEOS
3.8.0-CAPI-1.13.1 
Compiled against SQLite
3.29.0
Running against SQLite
3.29.0
PostgreSQL Client Version
11.5
SpatiaLite Version
4.3.0
QWT Version
6.1.3
QScintilla2 Version
2.10.8
Compiled against PROJ
6.3.1
Running against PROJ
OS Version
Windows 10 (10.0)
Active python plugins
db_manager; 
MetaSearch; 
processing

**Additional context**

Before we were aware that the problem involved the number of symbology categories, we had been unable to create a functional test file.

A short-term fix was implemented, which involved the Edit Mask checkbox. Although this reduced the delay duration, it can still be very long.

This is a new bug in 3.12, and it seems to be due to the introduction of the Selective Masking feature.

I think it is probably a regression.

mhugo:
## Description

Defer the loading of masking widgets (mask sources and targets) in the layer properties.

They are loaded only on explicit user request, when the user ticks a checkbox.

Capture d’écran de 2020-04-03 16-22-53

Should fix #34942 

### DIFF
--- a/src/gui/qgsmaskingwidget.cpp
+++ b/src/gui/qgsmaskingwidget.cpp
@@ -14,6 +14,7 @@
  ***************************************************************************/
 
 #include <QSet>
+#include <QCheckBox>
 
 #include "qgsmaskingwidget.h"
 #include "qgsmasksourceselectionwidget.h"
@@ -41,6 +42,14 @@ QgsMaskingWidget::QgsMaskingWidget( QWidget *parent ) :
   connect( mMaskSourcesWidget, &QgsMaskSourceSelectionWidget::changed, this, [&]()
   {
     emit widgetChanged();
+  } );
+
+  connect( mEditMaskSettingsGroup, &QGroupBox::toggled, this, [&]( bool on )
+  {
+    if ( on && mLayer )
+    {
+      populate();
+    }
   } );
 }
 
@@ -133,8 +142,16 @@ QList<QPair<QgsSymbolLayerId, QList<QgsSymbolLayerReference>>> symbolLayerMasks(
 void QgsMaskingWidget::setLayer( QgsVectorLayer *layer )
 {
   mLayer = layer;
+  if ( mEditMaskSettingsGroup->isChecked() )
+  {
+    populate();
+  }
+}
+
+void QgsMaskingWidget::populate()
+{
   mMaskSourcesWidget->update();
-  mMaskTargetsWidget->setLayer( layer );
+  mMaskTargetsWidget->setLayer( mLayer );
 
   // collect masks and filter on those which have the current layer as destination
   QSet<QgsSymbolLayerId> maskedSymbolLayers;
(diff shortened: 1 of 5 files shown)


nyalldawson:
Fix looks good! Can you fix the widget alignment in the panel though? (label vs treeview). The layout would look nicer (and more consistent with the other tabs) if you used a checkable group box here instead of the seperate qcheckbox/qframe approach.

SteveLowman:
Does this mean the merge has been completed?

mhugo:
@nyalldawson Yes indeed, thanks for the reminder. That should be better now.

@SteveLowman Not yet, I'll merge once the automatic tests are ok and will tell you.

SteveLowman:
@mhugo I downloaded and installed Friday's nightly build, but the fix does not seem to be there. Should I still be waiting?

mhugo:
@SteveLowman Well it should be there in the "qgis-dev" version (3.13.0-27) and it should now also be in the "qgis-rel-dev" (3.12.1-10) that can be installed through the OSGeo4W installer

PR #35559, opened 2020-04-03
Expression builder : add layer_property(...,'distance_units')

jmonticolo:
## Description

For complete the existing function `layer_property(...)`, this change add a new argument 'distance_units' for return a string with the layer distance units (see [QgsUnitTypes::DistanceUnit](https://qgis.org/api/classQgsUnitTypes.html#a0602e8a06afd37c267e9e532bb27f50d))

This function can be used for display units for labels, in layouts or for access to more layer properties in the expression builder for algorithms.


nyalldawson (approved):

nyalldawson:
Thanks @jmonticolo !

dubravat:
Thank you Julien

PR #35576, opened 2020-04-04
[FEATURE][layouts] New item type for marker symbols

nyalldawson:
Allows marker symbols to be placed directly on layouts, eg. for manually placing markers over a map item or for creation of advanced custom legends.

(previously you had to hack support for these, e.g. by drawing rectangles using centroid fill symbols. Much nicer to have a dedicated item type for it, allowing more specific customisation options (coming soon!))

Sponsored by SLYR

image


nirvn:
Instantly felt in love with the item's icon :) 

This will greatly help in making custom legends, thanks.

nyalldawson:
This is ready for merge now

nirvn (approved):
Code looks good, merge away.

PR #35584, opened 2020-04-04
ctrl-k should not perform a new search in locator

uclaros:
## Description
We want ctrl-k to have the same effect as clicking the locator widget with the mouse.
#35119 only tackled half the problem by keeping the last locator string. We also don't want to perform a new search with the same string, rather than just popup the previous search results. 
That way the last selected result is still selected!


nyalldawson:
Great fix -- keep up the fantastic work @uclaros !

PR #35491, opened 2020-03-31
Random points on lines update

havatv:
## Description

1. Update the progress reporting for the *Random Points on Lines* algorithm.
   * Follow the same logic as @nyalldawson suggested.
   * Use addition instead of multiplication.

   Seems to work correctly.

PR #35589, opened 2020-04-04
Random points on lines update2

havatv:
## Description

New try, after #35491 was messed up.
Update the progress reporting for the Random Points on Lines algorithm.
Follow the same logic as @nyalldawson suggested.
Use addition instead of multiplication.

Seems to work correctly.


havatv:
Thanks, @nyalldawson!

Issue #35594, opened 2020-04-05
Set the correct temporal extent


saberraz:
I have seen that in the mesh provider, but I think it might be applicable to other dataset with temporal extent:

When loading a layer spanning over a time period, the logical thing is:

- If it is the first data with the time component, set the project/canvas time to the first time step of the newly added dataset
- If there are other datasets already loaded with time component, show the time step set in the time manager widget, if available.

The above logic follows similar one for spatial extent: QGIS zooms to the extent of the first spatial layer. The subsequent layers only shown within the current map extent (if available).

saberraz:
Could an admin please add a new label for Time Manager....now that we have the feature, there will be bugs with it too :)

gioman:
> Could an admin please add a new label for Time Manager....now that we have the feature, there will be bugs with it too :) 

@saberraz 

done

saberraz:
Thanks Gio!

PR #35599, opened 2020-04-05
Set time of the temporal controller widget

vcloarec:
This PR purposes a way to handle time when a temporal layer is loaded for the first time in a project.
Not sure it is the better way to do this, but it permits to start discussion about and every suggestion is of course welcomed.
The principle is :
Sets the time of the temporal controller widget accordingly with layer time extent (or project time extent) when loading temporal layer for the first time.
Set the time of the temporal controller to current date when the project is cleared.

### DIFF
--- a/src/gui/qgstemporalcontrollerwidget.cpp
+++ b/src/gui/qgstemporalcontrollerwidget.cpp
@@ -22,6 +22,7 @@
 #include "qgstemporalnavigationobject.h"
 #include "qgstemporalmapsettingswidget.h"
 #include "qgstemporalutils.h"
+#include "qgsmaplayertemporalproperties.h"
 
 QgsTemporalControllerWidget::QgsTemporalControllerWidget( QWidget *parent )
   : QgsPanelWidget( parent )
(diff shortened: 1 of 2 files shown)


saberraz:
Resolves https://github.com/qgis/QGIS/issues/35594

nyalldawson:
I'd suggest copying the exact approach used for the crs handling, i.e. put all the logic in QgsLayerTreeMapCanvasBridge (or cleaner, make a new QgsTemporalControllerMapCanvasBridge version of this!)

nyalldawson:
But I'm definitely +1 on the logic itself

vcloarec:
@nyalldawson thanks for your suggestion.

I had thought about putting the logic in QgsLayerTreeMapCanvasBridge but, as we can not access directly (without cast) to the temporal controller widget from that bridge class, I given up and founs it easier to connect the temporal controller widget with the project directly, even more that a signal/slot connection exists already (signal readProject).

For a new class QgsTemporalControllerMapCanvasBridge, maybe I don't understand, the QgsTemporalNavigationObject (derived from  QgsTemporalController) and the canvas are already directly connected. Do you suggest to introduce a new class to make a more complete connection, for example update the temporal controller if data source of one layer changes? In that case, the name shouldn't be QgsLayerTreeTemporalControllerBridge?

nyalldawson:
Ok, fair enough points. Let's go with this approach.

PR #35604, opened 2020-04-06
Temporal navigation tests

Samweli:
Adds tests for the temporal navigation object, which is used by the temporal controller in handling all non-gui temporal animation logic.

### DIFF
--- a/tests/src/core/testqgstemporalnavigationobject.cpp
+++ b/tests/src/core/testqgstemporalnavigationobject.cpp
@@ -1,0 +1,159 @@
+/***************************************************************************
+                         testqgstemporalnavigationobject.cpp
+                         ---------------
+    begin                : April 2020
+    copyright            : (C) 2020 by Samweli Mwakisambwe
+    email                : samweli at kartoza dot com
+ ***************************************************************************/
+
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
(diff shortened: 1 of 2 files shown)


tests/src/core/testqgstemporalnavigationobject.cpp
```diff
@@ -0,0 +1,159 @@
...
+  QCOMPARE( stateSignal.count(), 4 );
+
+  navigationObject->next();
+  QCOMPARE( navigationObject->currentFrameNumber(), 1 );
+
+  navigationObject->previous();
+  QCOMPARE( navigationObject->currentFrameNumber(), 0 );
+
+  navigationObject->skipToEnd();
+  QCOMPARE( navigationObject->currentFrameNumber(), navigationObject->totalFrameCount() - 1 );
```
nyalldawson, line 108: I don't think this is testing anything at the moment, because the default range of a navigation object is 0 length. I think you should manually set a range and time step at the start of this test and check a static value here.

Samweli: Sure

nyalldawson: That's better, thanks!

nyalldawson (approved):

tests/src/core/testqgstemporalnavigationobject.cpp
Samweli: Cool!

PR #35606, opened 2020-04-06
Allow plugins to register custom "Project Open" handlers

nyalldawson:
These allow plugins to extend the "Open Project" dialog by adding in support for new file filters, which appear in the formats drop down list alongside the existing "QGS Projects" entry.

Custom project open handlers then get first chance at loading project files.

This allows plugins to extend QGIS support by adding integrated support for opening projects from non QGS/QGZ formats, e.g. allowing users to open ArcGIS MXD documents or MapInfo WOR Workspaces direct from the project open dialog.

These non-native projects are also added to the recent projects list and welcome screen, giving them a truly first-class experience within QGIS.

Sponsored by SLYR

Peek 2020-04-06 10-47


NathanW2:
image

PR #35613, opened 2020-04-06
[vector] Fix bogus assert when adding polygons with topological edit toggled on

nirvn:
## Description

Jumping from one crash to another.

QGIS crashes due to a bogus assert in the add feature tool while adding polygons when the topological editing is toggled on. This PR fixes that.

### DIFF
--- a/src/app/qgsmaptooladdfeature.cpp
+++ b/src/app/qgsmaptooladdfeature.cpp
@@ -34,6 +34,7 @@
 #include "qgisapp.h"
 #include "qgsexpressioncontextutils.h"
 #include "qgsrubberband.h"
+
 #include <QSettings>
 
 QgsMapToolAddFeature::QgsMapToolAddFeature( QgsMapCanvas *canvas, CaptureMode mode )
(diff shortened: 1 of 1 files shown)


src/app/qgsmaptooladdfeature.cpp
```diff
@@ -91,7 +92,7 @@ void QgsMapToolAddFeature::digitized( const QgsFeature &f )
     if ( topologicalEditing )
     {
       QList<QgsPointLocator::Match> sm = snappingMatches();
-      Q_ASSERT( f.geometry().constGet()->vertexCount() == sm.size() );
+      Q_ASSERT( f.geometry().constGet()->vertexCount() == ( vlayer->geometryType() == QgsWkbTypes::PolygonGeometry ? sm.size() + 1 : sm.size() ) );
```
m-kuhn, line 95: Could the same also happen on a ring line geometry? Maybe just a `vertexCount > sm.size()`?
Disclosure: I don't have the complete picture.

nirvn: @m-kuhn , define a ring line geometry? To my knowledge, only polygons end up with vertices entered + 1 when its geometry is created.

m-kuhn: It's also possible to draw a line that has multiple vertices on top of each other. I wonder if snapping kills such duplicates. If not, please ignore my comment.

nirvn: @m-kuhn , let's go with this for now -- if I stumble on more bogus assert here, I'll fix along the road. I.e., I'm also missing a bigger picture at this moment in time, therefore I'm inclined to stick as closely as possible to the original assert's intent.

m-kuhn (approved):

PR #35614, opened 2020-04-06
 [vector] Fix bogus assert when adding polygons with topological edit toggled on

nirvn:
## Description

Themes failed to offer a visual hint of whether a given tool button, when checked, was disabled or not. This backported commit addresses that.


nyalldawson (approved):

nirvn:
@m-kuhn , ... just hit more bogus assert there, I think this needs to go altogether.

PR #32474, opened 2019-10-29
Add topological points on background layers for qgsmaptooladdfeature

lbartoletti:
## Description

Unlike other map tools I fixed, this one doesn't add topological points on background layers / snapped layers. 

before:

before webm

after:

topological_addfeature_background_layers webm

## Checklist

- [X] Commit messages are descriptive and explain the rationale for changes
- [ ] Commits which fix bugs include `Fixes #11111` at the bottom of the commit message
- [X] I have read the [QGIS Coding Standards](https://docs.qgis.org/testing/en/docs/developers_guide/codingstandards.html) and this PR complies with them
- [ ] New unit tests have been added for core changes
- [X] I have run [the `scripts/prepare-commit.sh` script](https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTING.md#contributing-to-qgis) before each commit
- [X] I have evaluated whether it is appropriate for this PR to be backported, backport requests are left as label or comment

PR #35622, opened 2020-04-06
[vector] Remove problematic topological editing related assert when adding features

nirvn:
## Description

So it turns out there are more scenarios where the assert check is bogus. Other than a newly-created polygon having nb of vertices entered + 1 as its final nb of geometry vertices, the count also doesn't match when a newly added polygon as clipped when [x] avoid overlap is active.

The more I've gotten to familiarize myself with the code here, the more I think this assert should simply go away. 

@m-kuhn , all good?


m-kuhn:
For me, yes.

@lbartoletti you seem to have added these checks in https://github.com/qgis/QGIS/pull/32474 , how do you feel about it?

nirvn:
@lbartoletti , all good with removing the assert?

lbartoletti:
LGTM
Thanks @nirvn

PR #35643, opened 2020-04-07
[snapping] Fix broken version detection which leads to broken reading of snapping type flags

nirvn:
## Description

The regular expression used to extract the project file's version was wrong, and made it impossible to properly restore snapping type flags. This PR fixes that.

Were snapping flags adding in 3.12 or only master? Not sure this needs backporting.

### DIFF
--- a/src/core/qgssnappingconfig.cpp
+++ b/src/core/qgssnappingconfig.cpp
@@ -372,7 +372,7 @@ void QgsSnappingConfig::readProject( const QDomDocument &doc )
     if ( versionElem.hasAttribute( QStringLiteral( "version" ) ) )
     {
       version = versionElem.attribute( QStringLiteral( "version" ) );
-      QRegularExpression re( "([\\d]*)\\.([\\d]*)" );
+      QRegularExpression re( "([\\d]+)\\.([\\d]+)" );
       QRegularExpressionMatch match = re.match( version );
       if ( match.hasMatch() )
       {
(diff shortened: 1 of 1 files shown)


lbartoletti:
Thanks @nirvn 
It's only master.

nirvn:
@lbartoletti , OK, noted. 

BTW, I've also noticed there's something wrong with the label in the advanced configuration widget's type column:
Screenshot from 2020-04-07 14-06-01
_Notice the absence of actual words in between the commas._

If you could look into that one, it'd be great.

src/core/qgssnappingconfig.cpp
```diff
@@ -372,7 +372,7 @@ void QgsSnappingConfig::readProject( const QDomDocument &doc )
     if ( versionElem.hasAttribute( QStringLiteral( "version" ) ) )
     {
       version = versionElem.attribute( QStringLiteral( "version" ) );
-      QRegularExpression re( "(\\d).(\\d)" );
+      QRegularExpression re( "([\\d]*)\\.([\\d]*)" );
```
3nids, line 375:
```suggestion
      QRegularExpression re( "([\\d]+)\\.([\\d]+)" );
```
?

3nids (approved):

lbartoletti:
> If you could look into that one, it'd be great.

@nirvn I'll take a look, next week. Thanks

PR #35644, opened 2020-04-07
Allow copying/cut/paste of model components

nyalldawson:
This commit allows users to copy and paste model components, both within the same model and between different models

Peek 2020-04-07 17-24

### DIFF
--- a/python/core/auto_generated/processing/models/qgsprocessingmodelgroupbox.sip.in
+++ b/python/core/auto_generated/processing/models/qgsprocessingmodelgroupbox.sip.in
@@ -38,7 +38,7 @@ Saves this group box to a QVariant.
 .. seealso:: :py:func:`loadVariant`
 %End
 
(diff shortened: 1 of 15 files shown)


alexbruy (approved):
Looks good.

But I'm no sure if we should paste also all child items by default, in some cases this is not necesary and user will need to remove unused items after pasting. Maybe we should have two paste actions: one for pasting selected items and another one for pasting selected items and their childs

nyalldawson:
What about if we only copy the comments/outputs if they are manually selected?

alexbruy:
Sounds good to me. But I'm not an UX expert, probably we should also ask @nirvn.

nirvn:
Sorry, I've not seen my name mentioned here until now. 

I think @nyalldawson 's proposal is good.

nyalldawson:
Ok, implemented. This is much nicer logic, thanks!

PR #35648, opened 2020-04-07
[vector] Remove problematic topological editing related assert when adding features

nirvn:
Assert crash backport


nyalldawson (approved):

nirvn:
:bowing_man:

PR #35669, opened 2020-04-08
[MESH][FEATURE] update to MDAL 0.5.91 Multi meshes in one file

vcloarec:
In addition of updating to MDAL 0.5.91, this PR allows to load several meshes from one file.
The file is parsed and a sub layer dialog is launched as the same logic than vector layers or raster layers.
multimesh

### DIFF
--- a/src/gui/qgssublayersdialog.h
+++ b/src/gui/qgssublayersdialog.h
@@ -36,7 +36,7 @@ class GUI_EXPORT QgsSublayersDialog : public QDialog, private Ui::QgsSublayersDi
       Ogr,
       Gdal,
       Vsifile,
-      Mdal
+      Mdal //!< \since QGIS 3.14
     };
 
     /**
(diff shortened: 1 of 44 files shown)


PeterPetrik (commented):
already reviewed in https://github.com/vcloarec/QGIS/pull/13 and happy with the implementation

src/gui/qgssublayersdialog.h
```diff
@@ -35,7 +35,8 @@ class GUI_EXPORT QgsSublayersDialog : public QDialog, private Ui::QgsSublayersDi
     {
       Ogr,
       Gdal,
-      Vsifile
+      Vsifile,
+      Mdal
```
PeterPetrik, line 39: can you please add `//!< \since QGIS 3.14` here?

PeterPetrik (approved):

PR #35677, opened 2020-04-09
Add api to override legend patch shape for legend nodes

nyalldawson:
API only for now, but ultimately allows this kind of thing:

image

Sponsored by SLYR


roya0045:
If I understand correctly to make the substitute symbol you make a geometry in a 'mini-map'. How would that be handled with auto-update?

nyalldawson:
@roya0045 there's no linkage to map objects here. We just use QgsGeometry because it's a convenient container for "drawing objects" like lines and polygons. 

The patch shapes are all static, and the plan is to ship with a handful of nice choices out of the box.

roya0045:
Ah I see that you store it in the nodes and override the preview icon if present. I was just curious since the legend tends to simply delete and repopulate the nodes on some events such as refresh. This is why I made the optimization PR to lessen that.

 But as long as the legend gets some love I'm fine with it!

Though I'm curious what is the best way to implement this and how it could be propagated( when making categories and ranges for instance) to a layers' symbology easily and in a convenient way.

nirvn (approved):
Nice!

PR #35678, opened 2020-04-09
Enable per-pixel horizontal scrolling on attribute table

nyalldawson:
Instead of the previous per-cell scrolling mode, which is very frustrating to work with when you have wide columns (e.g. columns wider than the the attribute table window)


nirvn (approved):
Yes please.

PR #35694, opened 2020-04-09
[feature] add native Round raster algorithm

root676:
## Description
This PR proposes a new processing algorithm to **round Raster values** (thanks @nyalldawson, your [algorithm list](https://github.com/nyalldawson/qgis-processing-hitlist/blob/master/fme.md
) triggered me ;) ). The algorithm is a reimplementation of the FME counterpart [RasterCellValueRounder](https://docs.safe.com/fme/html/FME_Desktop_Documentation/FME_Transformers/Transformers/rastercellvaluerounder.htm). 

The main functionality of the algorithm is either standard up/nearest/down-rounding of floating point cell values or rounding to powers of a user specified base _n_. By exposing the base as parameter to the user, the algorithm now is even a bit more versatile than the FME counterpart and allows for some interesting raster generalization (see example). The algorithm can be used with rasters of all types (even byte/integer rasters when rounding to powers of n) and outputs the same raster data type, a standard floating point rounding of an integer raster will copy it and raise a warning. A detailed description of the parameters interplay can be found in the algorithm helpstring.
The PR also contains a test of the algorithm.

**Example:**
input
_Input Raster file_
generalized
Output raster rounded to multiples of 10

Please feel free to review/comment!


roya0045:
Isn't that smooth?

And also looking at the code, this made me think. 
Isn't there a way to convert the images to an array and then reproject them? Having a complete raster as an array would be far better to control like in numpy. Though this might not be easy to do with the type of data we are dealing with.

images/images.qrc
```diff
@@ -116,6 +116,7 @@
         <file>themes/default/algorithms/mAlgorithmRandomPointsWithinPolygon.svg</file>
         <file>themes/default/algorithms/mAlgorithmRandomPointsWithinExtent.svg</file>
         <file>themes/default/algorithms/mAlgorithmRegularPoints.svg</file>
+	<file>themes/default/algorithms/mAlgorithmRoundRastervalues.svg</file>
```
nyalldawson, line 119: indentation

src/analysis/processing/qgsalgorithmroundrastervalues.cpp
```diff
@@ -0,0 +1,262 @@
...
+}
+
+QStringList QgsRoundRasterValuesAlgorithm::tags() const
+{
+  return QObject::tr( "data,cells,round,truncate" ).split( ',' );
+}
+
+QString QgsRoundRasterValuesAlgorithm::group() const
+{
+  return QObject::tr( "Raster tools" );
```
nyalldawson, line 40: I'm not sure Raster Tools is the right group for this. I think it fits better in the current Raster Analysis group, but ideally we'd split that group into purely analysis tools vs "manipulation" tools.

src/analysis/processing/qgsalgorithmroundrastervalues.cpp
```diff
@@ -0,0 +1,262 @@
...
+{
+  return QStringLiteral( "rastertools" );
+}
+
+void QgsRoundRasterValuesAlgorithm::initAlgorithm( const QVariantMap & )
+{
+  addParameter( new QgsProcessingParameterRasterLayer( QStringLiteral( "INPUT" ), QStringLiteral( "Input raster" ) ) );
+  addParameter( new QgsProcessingParameterBand( QStringLiteral( "BAND" ), QObject::tr( "Band number" ), 1, QStringLiteral( "INPUT" ) ) );
+  addParameter( new QgsProcessingParameterEnum( QStringLiteral( "ROUNDING_DIRECTION" ), QObject::tr( "Rounding direction" ), QStringList() << QObject::tr( "Round up" ) << QObject::tr( "Round to nearest" ) << QObject::tr( "Round down" ), false, 1 ) );
+  addParameter( new QgsProcessingParameterNumber( QStringLiteral( "DECIMAL_PLACES" ), QObject::tr( "Number of decimals places (use negative values to round cell values to a multiple of a base n, see advanced parameters)" ), QgsProcessingParameterNumber::Integer, 2 ) );
```
nyalldawson, line 53: This description is very long compared with the usual parameter description. I'd shorten to just "number of decimal places" and leave the full explanation for the help string.

I keep thinking we need api to support extended descriptions for parameters, which we could show in a tooltip...

src/analysis/processing/qgsalgorithmroundrastervalues.cpp
```diff
@@ -0,0 +1,262 @@
...
+  addParameter( new QgsProcessingParameterRasterDestination( QStringLiteral( "OUTPUT" ), QObject::tr( "Output raster" ) ) );
+  std::unique_ptr< QgsProcessingParameterDefinition > baseParameter = qgis::make_unique< QgsProcessingParameterNumber >( QStringLiteral( "BASE_N" ), QObject::tr( "Base n (provides the base rounding raster values near/up/down to multiples of n when Decimal parameter is negative)" ), QgsProcessingParameterNumber::Integer, 10, true, 1 );
+  baseParameter->setFlags( QgsProcessingParameterDefinition::FlagAdvanced );
+  addParameter( baseParameter.release() );
+}
+
+QString QgsRoundRasterValuesAlgorithm::shortHelpString() const
+{
+  return QObject::tr( "This algorithm rounds the cell values of a rasterdataset according to the specified number of decimals.\n "
```
nyalldawson, line 62:
```suggestion
  return QObject::tr( "This algorithm rounds the cell values of a raster dataset according to the specified number of decimals.\n "
```

src/analysis/processing/qgsalgorithmroundrastervalues.cpp
```diff
@@ -0,0 +1,262 @@
...
+                      "Alternatively, a negative number of decimal places may be used to round values to powers of a base n "
+                      "(specified in the advanced parameter Base n). For example, with a Base value n of 10 and Decimal places of -1 "
+                      "the algorithm rounds cell values to multiples of 10, -2 rounds to multiples of 100, and so on. Arbitrary base values "
+                      "may be chosen, the algorithm applies the same multiplicative principle.Rounding cell values to multiples of "
```
nyalldawson, line 66:
```suggestion
                      "may be chosen, the algorithm applies the same multiplicative principle. Rounding cell values to multiples of "
```

src/analysis/processing/qgsalgorithmroundrastervalues.h
```diff
@@ -0,0 +1,85 @@
...
+
+  private:
+    double roundNearest( double &value, int &decimals );
+    double roundUp( double &value, int &decimals );
+    double roundDown( double &value, int &decimals );
+    double roundNearestBaseN( double &value, int &multipleOfBaseN );
+    double roundUpBaseN( double &value, int &multipleOfBaseN );
+    double roundDownBaseN( double &value, int &multipleOfBaseN );
+
+    QgsRasterLayer *mInputRaster;
```
nyalldawson, line 65: Better not to store this as a member -- it potentially risks someone using it from processAlgorithm later.

src/analysis/processing/qgsalgorithmroundrastervalues.cpp
```diff
@@ -0,0 +1,262 @@
...
+    }
+    else
+    {
+      for ( int row = 0; row < iterRows; row++ )
+      {
+        if ( feedback && feedback->isCanceled() )
+          break;
+        for ( int column = 0; column < iterCols; column++ )
+        {
+          if ( analysisRasterBlock->isNoData( row, column ) )
```
nyalldawson, line 175: Calling valueAndNoData once here instead of separately calling isNoData and the value is more efficient.

src/analysis/processing/qgsalgorithmroundrastervalues.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+  QVariantMap outputs;
+  outputs.insert( QStringLiteral( "OUTPUT" ), outputFile );
+  return outputs;
+}
+
+double QgsRoundRasterValuesAlgorithm::roundNearest( double &value, int &decimals )
+{
+  double m = ( value < 0.0 ) ? -1.0 : 1.0;
+  double scaleFactor = std::pow( 10.0, decimals );
```
nyalldawson, line 225: This scalefactor is getting calculated once per pixel -- it'd be preferable to calculate it once in processAlgorithm and pass it to this function.

nyalldawson (commented):
Nice addition! You've nailed the implementation too (like usual :wink: )

nyalldawson:
@roya0045 
>Isn't there a way to convert the images to an array

That's exactly what QgsRasterBlock is

src/analysis/processing/qgsalgorithmroundrastervalues.cpp
root676: I agree. One question on adding new processing algorithm groups between major releases: Is this considered as an API break. I plan to add more raster creation algorithms and would put them in one new group. Is it problematic to regroup algorithms/add new groups between major releases?

src/analysis/processing/qgsalgorithmroundrastervalues.cpp
root676: Agreed - this would be very nice!

src/analysis/processing/qgsalgorithmroundrastervalues.cpp
nyalldawson: It's not an API break, I think it's perfectly acceptable for a minor release!

images/images.qrc
root676: fixed

src/analysis/processing/qgsalgorithmroundrastervalues.cpp
root676: fixed

src/analysis/processing/qgsalgorithmroundrastervalues.cpp
root676: fixed

src/analysis/processing/qgsalgorithmroundrastervalues.cpp
root676: fixed

src/analysis/processing/qgsalgorithmroundrastervalues.h
root676: Do you mean the QgsRasterLayer member variable? If so, it is fixed.

src/analysis/processing/qgsalgorithmroundrastervalues.cpp
root676: fixed

src/analysis/processing/qgsalgorithmroundrastervalues.cpp
root676: fixed

root676:
@nyalldawson thank you for the fast review :)

roya0045:
@nyalldawson I just find it strange that values must be handled and set one at the time instead of using broadcasting. Though it still might me more efficient in c++ than numpy.

src/analysis/processing/qgsalgorithmroundrastervalues.cpp
```diff
@@ -0,0 +1,263 @@
...
+{
+  addParameter( new QgsProcessingParameterRasterLayer( QStringLiteral( "INPUT" ), QStringLiteral( "Input raster" ) ) );
+  addParameter( new QgsProcessingParameterBand( QStringLiteral( "BAND" ), QObject::tr( "Band number" ), 1, QStringLiteral( "INPUT" ) ) );
+  addParameter( new QgsProcessingParameterEnum( QStringLiteral( "ROUNDING_DIRECTION" ), QObject::tr( "Rounding direction" ), QStringList() << QObject::tr( "Round up" ) << QObject::tr( "Round to nearest" ) << QObject::tr( "Round down" ), false, 1 ) );
+  addParameter( new QgsProcessingParameterNumber( QStringLiteral( "DECIMAL_PLACES" ), QObject::tr( "Number of decimals places" ), QgsProcessingParameterNumber::Integer, 2 ) );
+  addParameter( new QgsProcessingParameterRasterDestination( QStringLiteral( "OUTPUT" ), QObject::tr( "Output raster" ) ) );
+  std::unique_ptr< QgsProcessingParameterDefinition > baseParameter = qgis::make_unique< QgsProcessingParameterNumber >( QStringLiteral( "BASE_N" ), QObject::tr( "Base n (provides the base rounding raster values near/up/down to multiples of n when Decimal parameter is negative)" ), QgsProcessingParameterNumber::Integer, 10, true, 1 );
```
nyalldawson, line 55: Better shorten this one too

src/analysis/processing/qgsalgorithmroundrastervalues.cpp
nyalldawson: Sorry, I think you misunderstood me -- it's "scale factor" which is a constant value and should only be calculated once

src/analysis/processing/qgsalgorithmroundrastervalues.cpp
```diff
@@ -0,0 +1,263 @@
...
+  return outputs;
+}
+
+double QgsRoundRasterValuesAlgorithm::roundNearest( double &value, double &m, int &decimals )
+{
+  double scaleFactor = std::pow( 10.0, decimals );
+  return ( std::round( value * m * scaleFactor ) / scaleFactor ) * m;
+}
+
+double QgsRoundRasterValuesAlgorithm::roundUp( double &value, double &m, int &decimals )
```
nyalldawson, line 232: These arguments should be passed by value, not reference (by reference is slower then by value for small trivially copied types like int and double)

nyalldawson (commented):
Nearly there!

src/analysis/processing/qgsalgorithmroundrastervalues.cpp
root676: I condensed the parameter description a little.

src/analysis/processing/qgsalgorithmroundrastervalues.cpp
root676: fixed

nyalldawson (approved):

nyalldawson:
Perfect!

PR #35699, opened 2020-04-10
AttribueForm: No python init when no custom UI

Djedouas:
I have an issue when using the multi edition tool: despite the fact that the custom UI is not used (expected) the python custom form code is executed resulting in errors because I always use python initialization for my custom UIs (which is not found in this case obviously).

I don't see a usecase where python init form code should be executed if no custom UI is loaded.

As a result, I submit this change.

### DIFF
--- a/src/gui/qgsattributeform.cpp
+++ b/src/gui/qgsattributeform.cpp
@@ -1709,9 +1709,10 @@ void QgsAttributeForm::initPython()
   cleanPython();
 
   // Init Python, if init function is not empty and the combo indicates
-  // the source for the function code
+  // the source for the function code and the customUi should be used
   if ( !mLayer->editFormConfig().initFunction().isEmpty()
-       && mLayer->editFormConfig().initCodeSource() != QgsEditFormConfig::CodeSourceNone )
+       && mLayer->editFormConfig().initCodeSource() != QgsEditFormConfig::CodeSourceNone 
+       && mContext.allowCustomUi())
   {
 
     QString initFunction = mLayer->editFormConfig().initFunction();
--- a/src/gui/qgsattributeform.cpp
+++ b/src/gui/qgsattributeform.cpp
@@ -1712,7 +1712,7 @@ void QgsAttributeForm::initPython()
   // the source for the function code and the customUi should be used
   if ( !mLayer->editFormConfig().initFunction().isEmpty()
        && mLayer->editFormConfig().initCodeSource() != QgsEditFormConfig::CodeSourceNone 
-       && mContext.allowCustomUi())
+       && mContext.allowCustomUi() )
   {
 
     QString initFunction = mLayer->editFormConfig().initFunction();


3nids:
> I don't see a usecase where python init form code should be executed if no custom UI is loaded

There are certainly usecases for this, to work on the widgets depending on data or context.

Now, the question of showing this or not in the multi edit tool is another question, for which I have no strong opinion.

For you scenario you can maybe add a simple test to see in which context you are.

3nids (changes requested):
> I don't see a usecase where python init form code should be executed if no custom UI is loaded

There are certainly usecases for this, to work on the widgets depending on data or context.

Now, the question of showing this or not in the multi edit tool is another question, for which I have no strong opinion.

For you scenario you can maybe add a simple test to see in which context you are.

Djedouas:
Thanks a lot for your quick answer!

Ok, I see what you mean...

However, in my scenario, I can't check the context.

It might be another issue...

The python init code is executed before, in the QgsAttributeForm constructor, and the MultiEdit context is set after, in the QgsAttributeDialog constructor.

As a result I can't check the context with python because it's not relevant...

1 https://github.com/qgis/QGIS/blob/master/src/app/qgisapp.cpp#L9507
2 https://github.com/qgis/QGIS/blob/master/src/gui/qgsattributedialog.cpp#L116
3 https://github.com/qgis/QGIS/blob/master/src/gui/qgsattributeform.cpp#L79

Maybe the mode should be set in the context prior to the QgsAttributeDialog instanciation, but there is no such method in QgsAttributeEditorContext.

3nids:
what I was suggesting is rather a hack for now: using Python test the UI to see if you get the widgets of your custom UI, if not, then do nothing.

But talking about context, I just remember that there is a mode:
https://qgis.org/pyqgis/master/gui/QgsAttributeForm.html#qgis.gui.qgsattributeform.mode

You can check if it equals `SingleEditMode`

Djedouas:
Yes I know about the context mode, and I explained in my last comment why I can't use that information.

The mode is not up to date when python code is executed, therefore I can not trust it.

(Follow the links in my previous comment)

That's why for now I use the hack of checking the validity of the feature, assuming that if the feature is not valid, then I'm in MultiEditMode. But that assumption is not satisfying.

3nids:
but `ìnitPython` defines the Python code to be run but i's not actually run until the the dialog is shown, right?
So at the time the Python code is run, the context and the feature are set (see in the code, the feature is also set after initPython).

Djedouas:
Have you found where the python code is run?

Djedouas:
Found it.

https://github.com/qgis/QGIS/blob/master/src/gui/qgsattributeform.cpp#L80

`QgsAttributeForm::setFeature()` executes the python code with the correct feature.

But this method is called in the QgsAttributeForm constructor, so in my case before the context is set.

Therefore there is an issue:

1. The multi edition is triggered and [the QgsAttributeDialog is instanciated](https://github.com/qgis/QGIS/blob/master/src/app/qgisapp.cpp#L9507) with a **default editor context mode**
2. [The QgsAttributeForm is instanciated](https://github.com/qgis/QGIS/blob/master/src/gui/qgsattributedialog.cpp#L116), still with a **default editor context mode**
3. [The python code is initialized](https://github.com/qgis/QGIS/blob/master/src/gui/qgsattributeform.cpp#L79) and [run with the feature](https://github.com/qgis/QGIS/blob/master/src/gui/qgsattributeform.cpp#L80) which is a wrong feature because the `QgsAttributeForm::setFeature()` method depends on the editor context mode, which is wrong here (still default)
4. [The editor context mode is set](https://github.com/qgis/QGIS/blob/master/src/app/qgisapp.cpp#L9508) but too late ! my python code has already been initialized and run without being able to check the context mode

3nids:
Well then the only approach I see is to modify the constructor of attribute form and dialog to set the mode directly. 
It shouldn't be too hard to do this.

gacarrillor:
FYI, issue documented here: https://github.com/qgis/QGIS/issues/53519

PR #35711, opened 2020-04-11
Don't prompt to cancel the news feed task on shutdown

nyalldawson:
If it hasn't completed, just cancel it immediately without prompting users -- it's not essential that it completes and the prompt is just annoying for this task.


nirvn:
Yes, yes, and yes.

m-kuhn (approved):

PR #35718, opened 2020-04-11
Increase default bin size for raster histograms

### LINKED ISSUE #35465: Raster Singleband Pseudocolor with Quantile mode inconsistencies
I'm seeing some inconsistencies in classes breaks, when using Quantile mode on a raster with singleband pseudocolor symbology.

To reproduce:
1. Download this sample raster and load it on QGIS:
https://cld.pt/dl/download/7161caab-5ebe-4e94-8c4a-574a80821a97/perigosidade.tif

2. Properties -> Symbology -> Render type: Singleband pseudocolor -> Interpolation: Discrete -> Mode: Quantile -> Classes: 5. You will get:
imagem

3. Run GRASS r.quantile from Processing on this raster, with Number of Quantiles: 5. Leave all other options as default. You will get:
```
0:20.000000:144.000000
1:40.000000:720.000000
2:60.000000:2400.000000
3:80.000000:4500.000000 
```
These are the correct classes breaks, and so the symbology should be:
imagem

I have done some tests to be sure about the correct classification.

[A] I've run Raster Pixels to Points to get the center point of all pixels. If you want to check, here is the output geopackage:
https://cld.pt/dl/download/3764537c-2504-4326-be77-74e26c0d2a49/perigosidade.gpkg

Load it in QGIS and go to Properties -> Symbology -> Graduated -> Mode: Equal Count (Quantile) -> Classes: 5 -> Classify. You will get:
imagem

[B] I've run GRASS r.stats to get all pixel values. Here is the CSV with all values:
https://cld.pt/dl/download/b730c614-40b3-40b8-887f-2b985e753d96/perigosidade_r_stats.csv

I've tested this csv in R:
```
> table_perigosidade <- read.csv("D:\\Testes\\quantile\\perigosidade_r_stats.csv")
> quantile(table_perigosidade$raster_value, probs = c(0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0))
   0%   10%   20%   30%   40%   50%   60%   70%   80%   90%  100% 
   40    96   144   288   720  1632  2400  3000  4500  6000 46721
```

[C] Then I extracted the absolute frequencies of raster values:
```
write.csv(count(table_perigosidade, 'raster_value'), file = "D:\\Testes\\quantile\\perigosidade_r_stats_abs_freq.csv")
```
https://cld.pt/dl/download/af265a0a-024f-40c4-a04c-77eaa9f09173/perigosidade_r_stats_abs_freq.csv

and did the manual analysis of it, getting the cumulative relative frequencies and the result is the same:
https://cld.pt/dl/download/9b014dbd-f991-42e9-880e-baec555a6d18/perigosidade_r_stats_abs_freq_cum.ods

So, it seems that Singleband Pseudocolor Quantile is not returning the correct breaks.

elpaso:
Fixes #35465 
but it is still an arbitrary value,
better approaches would require to calculate other
dispersion indexes and they seem impractical (inefficient)
in this case.

### DIFF
--- a/src/core/raster/qgscolorrampshader.cpp
+++ b/src/core/raster/qgscolorrampshader.cpp
@@ -208,7 +208,7 @@ void QgsColorRampShader::classifyColorRamp( const int classes, const int band, c
 
       double cut1 = std::numeric_limits<double>::quiet_NaN();
       double cut2 = std::numeric_limits<double>::quiet_NaN();
-      int sampleSize = 250000;
+      const int sampleSize = 250000;
 
       // set min and max from histogram, used later to calculate number of decimals to display
       input->cumulativeCut( band, 0.0, 1.0, min, max, extent, sampleSize );
--- a/src/core/raster/qgsrasterinterface.cpp
+++ b/src/core/raster/qgsrasterinterface.cpp
@@ -347,9 +347,12 @@ void QgsRasterInterface::initHistogram( QgsRasterHistogram &histogram,
     }
     else
     {
-      // There is no best default value, to display something reasonable in histogram chart, binCount should be small, OTOH, to get precise data for cumulative cut, the number should be big. Because it is easier to define fixed lower value for the chart, we calc optimum binCount for higher resolution (to avoid calculating that where histogram() is used. In any any case, it does not make sense to use more than width*height;
-      myBinCount = histogram.width * histogram.height;
-      if ( myBinCount > 1000 )  myBinCount = 1000;
+      // There is no best default value, to display something reasonable in histogram chart,
+      // binCount should be small, OTOH, to get precise data for cumulative cut, the number should be big.
+      // Because it is easier to define fixed lower value for the chart, we calc optimum binCount
+      // for higher resolution (to avoid calculating that where histogram() is used. In any any case,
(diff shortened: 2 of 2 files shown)


elpaso:
@alexbruy @rouault I'm not totally happy with my approach (I just doubled the max bin count), I was wondering if a better approach for whole numbers would be to just use the value range size for the bin count. I'm only worried that it could be too big in certain situations and slow things down.

Any hint?

alexbruy:
I'm not very familiar with this part of code, but maybe we can use mixed approach: pick value range if it is not very big and in other cases use current approach with arbitrary number of bins?

elpaso:
@alexbruy thanks for having a look, that is exactly the current approach, but the max size is capped to 2000 (it was 1000 before this PR), the code is ~10 years old, I guess we could probably make it even larger (65536?), a max size must be set because we need to handle the case of floats anyway.

elpaso:
Merging as it is, while we think about a better solution.

PedroVenancio:
Hi @elpaso 
Thank you very much for the fix!
Do you think it can be backported to 3.10 and 3.12?

PR #35724, opened 2020-04-12
fix handling of checked items in checkable combobox (fix #28658)

### LINKED ISSUE #28658: QgsCheckableComboBox single item selection bug
Author Name: **Andreas Rabe** (Andreas Rabe)
Original Redmine Issue: [20839](https://issues.qgis.org/issues/20839)
Affected QGIS version: 3.4.2
Redmine category:gui

---
Given a QgsCheckableComboBox with all items checked. Deselecting all items, but a single one. After clicking somewhere else in the GUI, the item that was still selected is also deselected most of the times.

Bug was observed in a small PyQt GUI programm and is reproducible inside the QtDesigner.

alexbruy:
## Description
Improve handling of toggling item state in the checkable combobox, so single checked item does not become unchecked. Fixes #28658.

### DIFF
--- a/src/gui/qgscheckablecombobox.cpp
+++ b/src/gui/qgscheckablecombobox.cpp
@@ -52,10 +52,10 @@ bool QgsCheckableItemModel::setData( const QModelIndex &index, const QVariant &v
 
   if ( ok && role == Qt::CheckStateRole )
   {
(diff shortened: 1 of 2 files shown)


Gustry:
Thanks @alexbruy !

PR #35737, opened 2020-04-13
Fix msvc utf-8

longhuan2018:
Solve the errors caused by utf-8 encoding under Chinese windows operating system, using msvc compilation


nyalldawson:
This change has broken some tests -- see https://dev.azure.com/qgis/QGIS/_build/results?buildId=4098&view=ms.vss-test-web.build-test-results-tab. Are you able to investigate please?

longhuan2018:
> This change has broken some tests -- see https://dev.azure.com/qgis/QGIS/_build/results?buildId=4098&view=ms.vss-test-web.build-test-results-tab. Are you able to investigate please?

https://doc.qt.io/qt-5/qstring.html#QStringLiteral
```
Note: Some compilers have bugs encoding strings containing characters outside the US-ASCII character set. Make sure you prefix your string with u in those cases. It is optional otherwise.
```

nyalldawson:
Thanks! Let's see if this round passes

PR #8922, opened 2019-01-21
Server: catch datasource error

mhugent:
If a datasource is not available (e.g. an external DB is temporarily down), QGIS Server currently silently draws the map without this information. For the desktop, this behaviour (render the map with the information available and writing a warning/error to the message log) is certainly what users want. For the server however this behaviour can lead to delivering invalid maps (e.g. public administration maps missing critical information). 

Other servers (e.g. UMN) return exceptions if a layer is not available. Therefore this PR tries to propagate rendering / DB connection error message upwards to catch them in GetPrint and GetMap. It furthermore prevents caching of projects with missing datasources (a datasource may become available later on).

PR #35738, opened 2020-04-13
Server: add QGIS_SERVER_IGNORE_BAD_LAYERS config option

elpaso:
Allow to override the default behavior in case of
bad layers (which is to invalidate the whole project).

When set to TRUE, the bad layers are skipped and the
project is considered available.

Followup #8922

Funded by: Geoinformatikbüro Dassau GmbH


rldhont:
@dmarteau to find out about it

elpaso:
unrelated test failure

PR #3593, opened 2016-10-11
Last Redo [BUGFIX][QGIS Server] Joins was not reloaded if the layer i…

rldhont:
…s in cache

Add an accessor to QgsVectorLayer join buffer, to not duplicate QgsVectorLayerJoinBuffer::readXml code

PR #35741, opened 2020-04-13
Update WMS-T layers project temporal range

Samweli:
Auto updates on WMS-T layers that are using project temporal range when the project temporal range has changed.

project_range_auto_update

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -11068,11 +11068,13 @@ void QgisApp::projectTemporalRangeChanged()
 
     if ( currentLayer->dataProvider() )
     {
-      QgsDataSourceUri uri;
-      uri.setEncodedUri( currentLayer->dataProvider()->dataSourceUri() );
+      QgsProviderMetadata *metadata = QgsProviderRegistry::instance()->providerMetadata(
+                                        currentLayer->providerType() );
 
-      if ( uri.hasParam( QStringLiteral( "temporalSource" ) ) &&
-           uri.param( QStringLiteral( "temporalSource" ) ) == QStringLiteral( "project" ) )
+      QVariantMap uri = metadata->decodeUri( currentLayer->dataProvider()->dataSourceUri() );
+
+      if ( uri.contains( QStringLiteral( "temporalSource" ) ) &&
+           uri.value( QStringLiteral( "temporalSource" ) ).toString() == QStringLiteral( "project" ) )
       {
         QgsDateTimeRange range = QgsProject::instance()->timeSettings()->temporalRange();
         if ( range.begin().isValid() && range.end().isValid() )
@@ -11080,12 +11082,9 @@ void QgisApp::projectTemporalRangeChanged()
           QString time = range.begin().toString( Qt::ISODateWithMs ) + "/" +
                          range.end().toString( Qt::ISODateWithMs );
 
-          uri.removeParam( QStringLiteral( "time" ) );
-          uri.setParam( QStringLiteral( "time" ), time );
+          uri["time"] = time;
 
-          currentLayer->dataProvider()->setDataSourceUri( uri.encodedUri() );
-          currentLayer->setDataSource( currentLayer->dataProvider()->dataSourceUri(), currentLayer->name(), currentLayer->providerType(), QgsDataProvider::ProviderOptions() );
-
+          currentLayer->setDataSource( metadata->encodeUri( uri ), currentLayer->name(), currentLayer->providerType(), QgsDataProvider::ProviderOptions() );
         }
       }
     }
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -11082,7 +11082,7 @@ void QgisApp::projectTemporalRangeChanged()
           QString time = range.begin().toString( Qt::ISODateWithMs ) + "/" +
                          range.end().toString( Qt::ISODateWithMs );
 
-          uri["time"] = time;
+          uri[ QStringLiteral( "time" ) ] = time;
 
           currentLayer->setDataSource( metadata->encodeUri( uri ), currentLayer->name(), currentLayer->providerType(), QgsDataProvider::ProviderOptions() );
         }
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -11079,7 +11079,7 @@ void QgisApp::projectTemporalRangeChanged()
         QgsDateTimeRange range = QgsProject::instance()->timeSettings()->temporalRange();
         if ( range.begin().isValid() && range.end().isValid() )
         {
-          QString time = range.begin().toString( Qt::ISODateWithMs ) + "/" +
+          QString time = range.begin().toString( Qt::ISODateWithMs ) + '/' +
                          range.end().toString( Qt::ISODateWithMs );
 
           uri[ QStringLiteral( "time" ) ] = time;
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -1189,7 +1189,6 @@ QgisApp::QgisApp( QSplashScreen *splash, bool restorePlugins, bool skipVersionCh
 
   mMapCanvas->setTemporalController( mTemporalControllerWidget->temporalController() );
 
-
   QgsGui::instance()->dataItemGuiProviderRegistry()->addProvider( new QgsAppDirectoryItemGuiProvider() );
   QgsGui::instance()->dataItemGuiProviderRegistry()->addProvider( new QgsProjectHomeItemGuiProvider() );
   QgsGui::instance()->dataItemGuiProviderRegistry()->addProvider( new QgsProjectItemGuiProvider() );
@@ -4079,6 +4078,8 @@ void QgisApp::setupConnections()
       canvas->setSelectionColor( selectionColor );
     }
   } );
+
+  connect( QgsProject::instance()->timeSettings(), &QgsProjectTimeSettings::temporalRangeChanged, this, &QgisApp::projectTemporalRangeChanged );
 
   // connect legend signals
   connect( this, &QgisApp::activeLayerChanged,
@@ -11052,6 +11053,41 @@ void QgisApp::projectCrsChanged()
       alreadyAsked.append( it.value()->crs() );
       askUserForDatumTransform( it.value()->crs(),
                                 QgsProject::instance()->crs(), it.value() );
+    }
+  }
+}
+
+void QgisApp::projectTemporalRangeChanged()
+{
+  QMap<QString, QgsMapLayer *> layers = QgsProject::instance()->mapLayers();
+  QgsMapLayer *currentLayer = nullptr;
+
+  for ( QMap<QString, QgsMapLayer *>::const_iterator it = layers.constBegin(); it != layers.constEnd(); ++it )
+  {
+    currentLayer = it.value();
+
+    if ( currentLayer->dataProvider() )
+    {
+      QgsDataSourceUri uri;
+      uri.setEncodedUri( currentLayer->dataProvider()->dataSourceUri() );
+
+      if ( uri.hasParam( QStringLiteral( "temporalSource" ) ) &&
+           uri.param( QStringLiteral( "temporalSource" ) ) == QStringLiteral( "project" ) )
+      {
+        QgsDateTimeRange range = QgsProject::instance()->timeSettings()->temporalRange();
+        if ( range.begin().isValid() && range.end().isValid() )
+        {
+          QString time = range.begin().toString( Qt::ISODateWithMs ) + "/" +
+                         range.end().toString( Qt::ISODateWithMs );
+
+          uri.removeParam( QStringLiteral( "time" ) );
+          uri.setParam( QStringLiteral( "time" ), time );
+
+          currentLayer->dataProvider()->setDataSourceUri( uri.encodedUri() );
+          currentLayer->setDataSource( currentLayer->dataProvider()->dataSourceUri(), currentLayer->name(), currentLayer->providerType(), QgsDataProvider::ProviderOptions() );
+
+        }
+      }
     }
   }
 }
--- a/src/core/raster/qgsrasterlayertemporalproperties.cpp
+++ b/src/core/raster/qgsrasterlayertemporalproperties.cpp
@@ -133,6 +133,7 @@ void QgsRasterLayerTemporalProperties::setDefaultsFromDataProviderTemporalCapabi
 {
   if ( const QgsRasterDataProviderTemporalCapabilities *rasterCaps = dynamic_cast< const QgsRasterDataProviderTemporalCapabilities *>( capabilities ) )
   {
+    setIsActive( rasterCaps->hasTemporalCapabilities() );
     setFixedTemporalRange( rasterCaps->availableTemporalRange() );
 
     if ( rasterCaps->hasTemporalCapabilities() )
--- a/src/core/raster/qgsrasterlayertemporalproperties.cpp
+++ b/src/core/raster/qgsrasterlayertemporalproperties.cpp
@@ -133,7 +133,6 @@ void QgsRasterLayerTemporalProperties::setDefaultsFromDataProviderTemporalCapabi
 {
   if ( const QgsRasterDataProviderTemporalCapabilities *rasterCaps = dynamic_cast< const QgsRasterDataProviderTemporalCapabilities *>( capabilities ) )
   {
-    setIsActive( rasterCaps->hasTemporalCapabilities() );
     setFixedTemporalRange( rasterCaps->availableTemporalRange() );
 
     if ( rasterCaps->hasTemporalCapabilities() )
--- a/src/gui/raster/qgsrasterlayerproperties.cpp
+++ b/src/gui/raster/qgsrasterlayerproperties.cpp
@@ -1273,7 +1273,7 @@ void QgsRasterLayerProperties::updateSourceStaticTime()
                        range.end().toString( Qt::ISODateWithMs );
 
         uri[ QStringLiteral( "time" ) ] = time;
-        uri[ QStringLiteral( "temporalSource" ) ] = QLatin1String( "project" );
+        uri[ QStringLiteral( "temporalSource" ) ] = QLatin1String( "project" )
       }
     }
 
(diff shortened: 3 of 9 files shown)


src/app/qgisapp.cpp
```diff
@@ -10837,6 +10838,41 @@ void QgisApp::projectCrsChanged()
...
+        QgsDateTimeRange range = QgsProject::instance()->timeSettings()->temporalRange();
+        if ( range.begin().isValid() && range.end().isValid() )
+        {
+          QString time = range.begin().toString( Qt::ISODateWithMs ) + "/" +
+                         range.end().toString( Qt::ISODateWithMs );
+
+          uri.removeParam( QStringLiteral( "time" ) );
+          uri.setParam( QStringLiteral( "time" ), time );
+
+          currentLayer->dataProvider()->setDataSourceUri( uri.encodedUri() );
```
elpaso, line 10867: Please do not use encodedUri!
... and keep in mind aware that WMST is not the only provider with temporal capabilities anymore.

src/app/qgisapp.cpp
```diff
@@ -10837,6 +10838,41 @@ void QgisApp::projectCrsChanged()
...
+  QgsMapLayer *currentLayer = nullptr;
+
+  for ( QMap<QString, QgsMapLayer *>::const_iterator it = layers.constBegin(); it != layers.constEnd(); ++it )
+  {
+    currentLayer = it.value();
+
+    if ( currentLayer->dataProvider() )
+    {
+      QgsDataSourceUri uri;
+      uri.setEncodedUri( currentLayer->dataProvider()->dataSourceUri() );
```
elpaso, line 10853: Please do not use setEncodedUri! Instead:
QgsDataSourceUri uri { currentLayer->dataProvider()->uri() };

... then:

currentLayer->setDataSource( uri.uri(), currentLayer->name(), currentLayer->providerType(), QgsDataProvider::ProviderOptions() );

src/app/qgisapp.cpp
```diff
@@ -10837,6 +10838,41 @@ void QgisApp::projectCrsChanged()
...
+
+      if ( uri.hasParam( QStringLiteral( "temporalSource" ) ) &&
```
elpaso, line 10855: Where can I find some documentation about "temporalSource" ? From where is that set? (I'm wondering how to support that in PG rasters)

elpaso (changes requested):

src/app/qgisapp.cpp
Samweli: Hi @elpaso, thanks, the issue is if I don't use setEncodedUri, the URI is not initialized properly. I have tried the way you have suggested many times, but still it doesn't work for WMS URI params parsing. I'll like to also use the QgsDataSourceUri constructor for initialization as long as it parse all the params from WMS.

src/app/qgisapp.cpp
Samweli: I have used "temporalSource" to store where the WMS-T layers get the static temporal range from, can either be "project" or "provider". This is set at first from the wms provider when loading the wms-t layer for the first time and from the raster layer properties when changing static temporal range options.

src/app/qgisapp.cpp
elpaso: Encoded URI won't work in all cases. So if there is a bug in the `QgsDataSourceUri` constructor you should try to fix it with high priority.

If you can write a test case to demonstrate your issue with WMS URI params parsing, that would be helpful.

Samweli: Ok @elpaso thanks, also I have realized that WMS provider uses Encoded URI everywhere it initialize the provider URI, this means I have to update all those other uses to use the QgsDataSourceURI constructor right?
As of now the updates on WMS-T layers temporal properties is not possible, eg. in this issue https://github.com/qgis/QGIS/issues/3593

elpaso: > Ok @elpaso thanks, also I have realized that WMS provider uses Encoded URI everywhere it initialize the provider URI, this means I have to update all those other uses to use the QgsDataSourceURI constructor right?

If you mean internally in the WMS provider code, I don't see why. 

The thing here is that you must write your code (outside a specific provider) in a way that is not bound to a particular provider (beware that it's not your fault, we all know `QgsDataSourceUri` is a mess).

So, either use the way I recommended you a few lines above (never use encoded uri!) or fix `QgsDataSourceUri` encode/decode to support cases where the provider source string is **not** in encoded uri form (such as postgres and almost all providers except for OGC services).

> As of now the updates on WMS-T layers temporal properties is not possible, eg. in this issue #3593

I could not reproduce that one, sorry.

Btw, I 'm still waiting for a python code snippet to reproduce your claim that "... the issue is if I don't use setEncodedUri, the URI is not initialized properly. ".

Samweli: About the python code snippet, I'm attaching one below, using this WMS Layer https://geo.weather.gc.ca/geomet?layer=GDPS.ETA_TT

`layer = iface.activeLayer()`
`encoded_uri = QgsDataSourceUri()`
`encoded_uri.setEncodedUri(layer.dataProvider().dataSourceUri())`

`# This below returns the right value 'EPSG:4326'`
`print(encoded_uri.param("crs") )`

`normal_uri = QgsDataSourceUri(layer.dataProvider().uri())`

`# This below returns a string of params with their values instead of a crs string,`
`# 'EPSG:4326&dpiMode=7&enableTime=yes&format=image/png&layers=GDPS.ETA_TT&...'`
`print(normal_uri.param("crs"))`

Samweli: > If you mean internally in the WMS provider code, I don't see why.

So as to have only one way of initializing the QgsDataSourceURI, and not mixing two different ways (using constructor externally while using setEncodedUri function internally).

elpaso: I see, QgsDataSourceUri is totally broken we should definitely bypass it and use encode/decodeUri from provider metadata.

I can take care of the implementation for PG raster, can you make it for WMS?

Here is a quick and dirty sample implementation (need to be tested!)

```c++
QVariantMap QgsWmsProviderMetadata::decodeUri(const QString& uri)
{
  const QUrlQuery query { uri };
  const auto constItems { query.queryItems() };
  QVariantMap decoded;
  for (const auto &item: constItems )
  {
    decoded[ item.first ] = item.second;
  }
  return decoded;
}

QString QgsWmsProviderMetadata::encodeUri(const QVariantMap& parts)
{
  QUrlQuery query;
  QList<QPair<QString, QString> > items;
  for (auto it = parts.constBegin(); it != parts.constEnd(); ++it )
  {
    items.push_back({it.key(), it.value().toString() });
  }
  query.setQueryItems( items );
  return query.toString();
}
```

This is how it is supposed to work in a provider agnostic way:

```python
>> md = QgsProviderRegistry.instance().providerMetadata('postgresraster')
# The following uri comes directly from ayer.dataProvider().dataSourceUri().
>> uri = 'dbname=\'qgis_tests\' host=localhost port=5432 user=\'ale\' password=\'ale\' sslmode=disable estimatedmetadata=true srid=0 temporalDefaultTime=\'2020-04-23T13:51:00\' temporalFieldIndex=\'2\' table="idro"."hn_giornaliera" (rast)'
>> decoded = md.decodeUri(uri)
>> print(decoded)
decoded = {'dbname': 'qgis_tests',
 'estimatedmetadata': True,
 'geometrycolumn': 'rast',
 'host': 'localhost',
 'password': 'ale',
 'port': '5432',
 'schema': 'idro',
 'selectatid': False,
 'srid': '0',
 'sslmode': 1,
 'table': 'hn_giornaliera',
 'username': 'ale'}
```
Now, suppose you want to set `enableTime`:

```python
>> decoded['enableTime'] = True  # be careful here: better use a bool than yes/no!
>> print(md.encodeUri(decoded))
'dbname=\'qgis_tests\' user=\'ale\' password=\'ale\' srid=0 enableTime=\'true\' estimatedmetadata=\'true\' host=\'localhost\' key=\'\' port=\'5432\' selectatid=\'false\' service=\'\' sslmode=\'disable\' temporalDefaultTime=\'2020-04-23T13:51:00\' temporalFieldIndex=\'2\' table="idro"."hn_giornaliera" (rast)'
```

The idea is to delegate to the provider the way it manages the uri.


So, for WMS, it would work the same:

```python
In [1]: layer = iface.activeLayer()

In [2]: md = QgsProviderRegistry.instance().providerMetadata('wms')

In [3]: md.decodeUri(layer.dataProvider().dataSourceUri())
Out[3]: 
{'contextualWMSLegend': '0',
 'crs': 'EPSG:4326',
 'dpiMode': '7',
 'enableTime': 'yes',
 'featureCount': '10',
 'format': 'image/png',
 'layers': 'GDPS.ETA_TT',
 'referenceTimeDimensionExtent': '2020-04-21T12:00:00Z/2020-04-23T00:00:00Z/PT12H',
 'styles': NULL,
 'temporalSource': 'provider',
 'timeDimensionExtent': '2020-04-23T00:00:00Z/2020-05-03T00:00:00Z/PT3H',
 'type': 'wmst',
 'url': 'https://geo.weather.gc.ca/geomet?layer=GDPS.ETA_TT'}
In [4]: decoded = md.decodeUri(layer.dataProvider().dataSourceUri())
In [5]: decoded['enableTime'] = False

In [7]: md.encodeUri(decoded)
Out[7]: 'contextualWMSLegend=0&crs=EPSG:4326&dpiMode=7&enableTime=false&featureCount=10&format=image/png&layers=GDPS.ETA_TT&referenceTimeDimensionExtent=2020-04-21T12:00:00Z/2020-04-23T00:00:00Z/PT12H&styles&temporalSource=provider&timeDimensionExtent=2020-04-23T00:00:00Z/2020-05-03T00:00:00Z/PT3H&type=wmst&url=https://geo.weather.gc.ca/geomet?layer%3DGDPS.ETA_TT'

```

elpaso: @Samweli  see: https://github.com/qgis/QGIS/pull/35952

Samweli: Cool!, How about saving the provider URI after adding/updating its params? Won't it affect the past structure of the URI when saving new changes using this approach?

Samweli: > @Samweli see: #35952

Ok

elpaso: > Cool!, How about saving the provider URI after adding/updating its params? Won't it affect the past structure of the URI when saving new changes using this approach?

Not sure if I understand what you mean, but if you want to call `setDataSource()` on the layer, it takes a string representation of the uri and it should work without any conversion.

In any event, the encode/decode is controlled by the provider through provider's metadata, it's the provider responsibility to decode/encode correctly, a round trip test it's probably the best check.

Samweli: Ok, so with these updates how will the following code change to..

`QgsDataSourceUri uri { currentLayer->dataProvider()->uri() };`

... then:

`currentLayer->setDataSource( uri.uri(), currentLayer->name(), currentLayer->providerType(), QgsDataProvider::ProviderOptions() );`

src/core/raster/qgsrasterlayertemporalproperties.cpp
```diff
@@ -133,7 +133,6 @@ void QgsRasterLayerTemporalProperties::setDefaultsFromDataProviderTemporalCapabi
 {
   if ( const QgsRasterDataProviderTemporalCapabilities *rasterCaps = dynamic_cast< const QgsRasterDataProviderTemporalCapabilities *>( capabilities ) )
   {
-    setIsActive( rasterCaps->hasTemporalCapabilities() );
```
nyalldawson, line 136: Why this change?

Samweli: When updating the WMS layer uri, the setDataSourceUri function is called to save changes, if the temporal properties where set to be inactive  by user, the above line will cause to reset them to be active.

src/app/qgisapp.cpp
timlinux: Thanks very much for all this great input @elpaso !

src/app/qgisapp.cpp
```diff
@@ -11056,6 +11057,40 @@ void QgisApp::projectCrsChanged()
...
+
+      QVariantMap uri = metadata->decodeUri( currentLayer->dataProvider()->dataSourceUri() );
+
+      if ( uri.contains( QStringLiteral( "temporalSource" ) ) &&
+           uri.value( QStringLiteral( "temporalSource" ) ).toString() == QStringLiteral( "project" ) )
+      {
+        QgsDateTimeRange range = QgsProject::instance()->timeSettings()->temporalRange();
+        if ( range.begin().isValid() && range.end().isValid() )
+        {
+          QString time = range.begin().toString( Qt::ISODateWithMs ) + "/" +
```
elpaso, line 11082:
```suggestion
          QString time = range.begin().toString( Qt::ISODateWithMs ) + '/' +
```

src/gui/raster/qgsrasterlayerproperties.cpp
```diff
@@ -1246,17 +1248,17 @@ void QgsRasterLayerProperties::updateSourceStaticTime()
...
     if ( mStaticTemporalRange->isChecked() )
     {
       QString time = mStartStaticDateTimeEdit->dateTime().toString( Qt::ISODateWithMs ) + '/' +
                      mEndStaticDateTimeEdit->dateTime().toString( Qt::ISODateWithMs );
-      uri.removeParam( QStringLiteral( "time" ) );
-      uri.setParam( QStringLiteral( "time" ), time );
-
-      uri.removeParam( QStringLiteral( "temporalSource" ) );
-      uri.setParam( QStringLiteral( "temporalSource" ), QLatin1String( "provider" ) );
+      uri["time"] = time;
```
elpaso, line 1260: QStringLiteral(s)

src/gui/raster/qgsrasterlayerproperties.cpp
```diff
@@ -1270,26 +1272,26 @@ void QgsRasterLayerProperties::updateSourceStaticTime()
         QString time = range.begin().toString( Qt::ISODateWithMs ) + "/" +
```
elpaso, line 1272:
```suggestion
        QString time = range.begin().toString( Qt::ISODateWithMs ) + '/' +
```

src/providers/wms/qgswmsprovider.cpp
```diff
@@ -1100,7 +1103,8 @@ void QgsWmsProvider::addWmstParameters( QUrlQuery &query )
     }
   }
 
-  if ( uri.param( QStringLiteral( "enableTime" ) ) == QLatin1String( "no" ) )
+  if ( uri.contains( QStringLiteral( "enableTime" ) ) &&
```
elpaso, line 1106: As I mentioned in another occasion I think it's a source of potential errors and confusion the fact that we are not consistent in the usage of:
- yes/no
- true/false
- 0/1
in uris.

I would strongly recommend to be tolerant and create a function that can accept any of the above mentioned values (case-insensitive), and stick to boolean true/false when you set a value.

Note: this is absolutely not your fault, but since you are doing an amazing job here I hope you can make an extra step and pave the path for a better and consistent handling of URI parameters  (for other providers too!).

Samweli: Thanks @elpaso, I have updated the PR, now using true/false values only.

Samweli: Also @elpaso, how would want the function you have recommend to look like? something like this
```c++
void QgsWmsProviderMetadata::setBoolParam( QVariantMap uri, const QString& parameter, const QVariant& value)
{
  if ( value.toString() == QStringLiteral( "yes" ) || value.toString() == QStringLiteral( "1" ) || value.toBool() ) {
  	uri[ parameter ] = true;
  }
  else if ( value.toString() == QStringLiteral( "no" ) || value.toString() == QStringLiteral( "0" ) || !value.toBool() ) {
  	uri[ parameter ] = false;
  }
}
```

elpaso: Yes! 

but maybe better if:
1. case insensitive, you may use https://doc.qt.io/qt-5/qstring.html#compare-1 case insensitive.
2. also check for true/false (case insensitive) string values
3. rename to `setBoolParameter` (for readability)

I think we also need one for `getBoolParameter`, maybe with a default optional argument, something like:
```c++
static bool QgsProviderMetadata::boolParameter( QVariantMap uri, const QString& parameter, bool defaultValue = false )
```

Samweli: Ok @elpaso can I make these on a different PR?

elpaso: Sure, do what's more convenient to you.

src/providers/wms/qgswmsprovider.cpp
```diff
@@ -1149,10 +1157,10 @@ void QgsWmsProvider::addWmstParameters( QUrlQuery &query )
   }
 
   // If the data provider has bi-temporal properties and they are enabled
-  if ( uri.hasParam( QStringLiteral( "reference_time" ) ) &&
-       !uri.param( QStringLiteral( "reference_time" ) ).isEmpty() )
+  if ( uri.contains( QStringLiteral( "reference_time" ) ) &&
+       !uri.value( QStringLiteral( "reference_time" ) ).toString().isEmpty() )
   {
-    QString time = uri.param( QStringLiteral( "reference_time" ) );
+    QString time = uri.value( QStringLiteral( "reference_time" ) ).toString();
```
elpaso, line 1163: Is there  a reson why "reference_time" contains an underscore and the other parameters are camel case?

src/providers/wms/qgswmsprovider.cpp
Samweli: Great

Samweli: @elpaso here https://github.com/qgis/QGIS/pull/36087

src/providers/wms/qgswmsprovider.cpp
Samweli: There isn't, I think I took that from WMS-T spec, will update it.

elpaso: Yes, better, thank you: just to be consistent.

Samweli:
@elpaso @nyalldawson I think this is ready for merge. thank you.

nyalldawson:
It looks good to me - @elpaso if you're happy then lets merge

elpaso (approved):

PR #35745, opened 2020-04-13
Random points in polygons C++ algorithm

havatv:
## Description

Could replace the Python algorithm *Random points inside polygons*, but
this C++ implementation only supports "Points count" ("number of points
for each feature") directly, while "Points density" is supported by allowing the
point count to be data defined (using the polygon area to calculate the
number of points).

Offers the same functionality as the *Random points inside polygons*
algorithm, and in addition it includes (possible to opt-out) the attributes
from the containing polygon and the ability to specify a seed for the random
number generator.
As a bonus, several of the parameters can be data defined.

I will provide the algorithm documentation for the user manual.


havatv:
Travis is not happy, but it does not seem related to this PR.

nyalldawson:
This change has broken some tests -- see https://dev.azure.com/qgis/QGIS/_build/results?buildId=4098&view=ms.vss-test-web.build-test-results-tab. Can you investigate please?

nyalldawson:
ah sorry, wrong tab!

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
```diff
@@ -0,0 +1,348 @@
...
+      }
+      // Try to add a point (mMaxAttempts attempts)
+      int distCheckIterations = 0;
+      while ( distCheckIterations < maxAttemptsForThisFeature )
+      {
+        if ( feedback->isCanceled() )
+        {
+          break;
+        }
+        // Generate a random point
```
nyalldawson, line 271: This should instead use the existing QgsGeometry::randomPointsInPolygon() function, with the "acceptPoint" callback function testing the spatial index distance when required. (This is exactly the use case that randomPointsInPolygon was designed with the callback function for :smile: )

havatv:
> This change has broken some tests -- see https://dev.azure.com/qgis/QGIS/_build/results?buildId=4098&view=ms.vss-test-web.build-test-results-tab. Can you investigate please?

Does your next message mean that I don't have to investigate this (I could not find anything related to my PR at the first look)?

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
havatv: I saw the function, but was not aware of the callback.  Will have a look.

nyalldawson:
> Does your next message mean that I don't have to investigate this (I could not find anything related to my PR at the first look)?

Yep, ignore it!

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
havatv: As far as I can tell, the *QgsGeometry::randomPointsInPolygon()* function **with the callback** will not work for this. It does not have a parameter that can be used to specify the "maximum number of attempts", so it will go into an infinite loop if the specified number of points can not be created due to the requirements of the callback function.
The *QgsGeometry::randomPointsInPolygon()* could be used without the callback to generate the points. I can try that, if you like, @nyalldawson.

havatv:
@nyalldawson, I have now updated the algorithm to use the QgsGeometry.randomPointsInPolygon function when generating points for the case when both the local and global minimum distance is zero. That should even improve performance.
Are you happy with this, or would you like me to also use that function (without the callback) for cases where the minimum distance is non-zero, or something else?

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
```diff
@@ -0,0 +1,416 @@
...
+
+    const double pointProgressIncrement = featureProgressStep / ( numberPointsForThisFeature * maxAttemptsForThisFeature );
+
+    double pointProgress = 0.0;
+
+    if ( ( minDistanceForThisFeature == 0 ) && ( mMinDistanceGlobal == 0 ) )
+    {
+      unsigned long fseed = uniformIntDist( mt );
+      QVector< QgsPointXY > newPoints = polyGeom.randomPointsInPolygon( numberPointsForThisFeature, fseed );
+      for ( int i=0; i < newPoints.length(); i++ )
```
havatv, line 277:
```suggestion
      for ( int i = 0; i < newPoints.length(); i++ )
```

havatv:
Anything else I should do, @nyalldawson?

havatv:
I think this *Random points in polygons* algorithm is ready for merging, or at least a new review.
The code should be low risk, and all checks except the windows cross build failure are successful, and I don't think the cross build failure is related to this code.
The code is close to a copy of the random points on lines algorithm when it comes to parameters, and runs fine on master. I also think that this C++ algorithm is a significant improvement compared to the Python algorithm (*Random points inside polygons* - `python/plugins/processing/algs/qgis/RandomPointsPolygons.py`) that now covers this task.
The input from the previous review has been addressed.
It would great if you could find the time for this, @nyalldawson.

nyalldawson:
I'll do it this week -- sorry for the delay!

havatv:
> I'll do it this week -- sorry for the delay!

Since "this week" probably was last week, I send you a ***ping***, @nyalldawson. Sorry for disturbing... :-)

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
nyalldawson: @havatv 

> It does not have a parameter that can be used to specify the "maximum number of attempts"

Let's add it to that method then. It's not exposed to python, so isn't subject to the stable api constraint.

Regarding your benchmarks -- indeed, running randomPointsInPolygon one by one is very inefficient. randomPointsInPolygon uses an initial tesselation of the geometry which pays off for multiple point generation. 

Could you try adding the max iterations logic to randomPointsInPolygon and test with the algorithm using that method?

havatv: OK.

havatv: @nyalldawson, how do use the "acceptPoint" callback function to test the distances?
When I try to use (global) algorithm variables (the minimum distances) in the "acceptPoint" function, they seem to get lost when the function is used from the QgsGeometry class.

Here is how I tried to declare the "acceptPoint" callback function (in the algorithm cpp file, I have not declared it in the header file).

```
bool pointOK( const QgsPointXY &newPoint )
{
  // May have to check minimum distance to existing points
  // The first point can always be added
  // Global first
  if ( mMinDistanceGlobal != 0 && mIndexPoints > 0 )
  {
    QList<QgsFeatureId> neighbors = globalIndex2.nearestNeighbor( newPoint, 1, mMinDistanceGlobal );
    if ( mTotNPoints > 0 && !neighbors.empty() )
    {
      return false;
    }
  }
  // Then local (if larger than global)
  if ( minDistanceForThisFeature != 0 && mMinDistanceGlobal < minDistanceForThisFeature && mLocalIndexPoints > 0 )
  {
    QList<QgsFeatureId> neighbors = localIndex2.nearestNeighbor( newPoint, 1, minDistanceForThisFeature );
    if ( mTotNPoints > 0 && !neighbors.empty() )
    {
      return false;
    }
  }
  QgsFeature f = QgsFeature( mTotNPoints );
  QgsAttributes pAttrs = QgsAttributes();
  pAttrs.append( mTotNPoints );
  f.setAttributes( pAttrs );
  QgsGeometry newGeom = QgsGeometry::fromPointXY( newPoint );

  f.setGeometry( newGeom );
  mTotNPoints++;

  if ( minDistanceForThisFeature != 0 )
  {
    localIndex2.addFeature( f );
    mLocalIndexPoints++;
  }
  if ( mMinDistanceGlobal != 0 )
  {
    globalIndex2.addFeature( f );
    mIndexPoints++;
  }
  return true;
}
```
mMindistanceGlobal and minDistanceForThisFeature should be global variables, set in the algorithm, but they are always 0 when the function is called from QgsGeometry.randomPoinsInPolygon. mTotNPoints is also a global variable, but only set in this function - it behaves as expected.

And I used the function like this (in the algorithm code):

`      QVector< QgsPointXY > newPoints = polyGeom.randomPointsInPolygon( numberPointsForThisFeature, pointOK, fseed, feedback, maxAttemptsForThisFeature );`

(the last parameter was introduced when I modified the randomPointsInPolygon method to accept a maximum number of attempts).

havatv: @nyalldawson, I have now tried to update the QgsGeometry randomPointsInPolygon function (adding a new parameter for maximum number of attempts).
With the updated QgsGeometry randomPointsInPolygon function, my numbers for the python/plugins/processing/tests/testdata/polys.gml dataset are:
| Parameters         | geom randompoints | bbox-based    |
| ------------------ | ------------------ | ------------- |
| 10000, 0.0, 10, 1  |  0.57 seconds      | 2.81 seconds  |
| 100, 1.0, 100, 1   |  0.69 seconds      | 1.83 seconds  |
| 10, 1.0, 1000, 1   |  0.29 seconds      | 0.60 seconds  |
| 10, 1.0, 10000, 1  |  1.64 seconds     | 3.87 seconds  |

The parameters are still listed in the order they appear in the GUI: _Number of points for each feature_, _Minimum distance between points_, _Maximum number of search attempts_ and _Random seed_.

I will try to update this PR with the above mentioned changes.

havatv:
Tests are not updated yet.

havatv:
Test output is now updated, so this should be ready for inspection, @nyalldawson.
Updating QgsGeometry is a bit scaring, and my knowledge of the class and C++ is limited, so those updates need to be checked carefully.

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
havatv: This PR has now been updated, and currently only use the new version of the QgsGeometry randomPointsInPolygon function to generate the random points.
Ready for inspection, @nyalldawson.

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
```diff
@@ -0,0 +1,399 @@
...
+        pointProgress += pointProgressIncrement * ( maxAttemptsForThisFeature );
+      }
+      feedback->setProgress( baseFeatureProgress + pointProgress );
+      continue;
+    }
+    else
+    {
+      // Have to check for minimum distance
+      unsigned long fseed = uniformIntDist( mt );
+      QVector< QgsPointXY > newPoints = polyGeom.randomPointsInPolygon( numberPointsForThisFeature, pointOK, fseed, feedback, maxAttemptsForThisFeature );
```
nyalldawson, line 353: What we should do here is use a lambda function and remove all the global variables/ methods. 

E.g.

```suggestion
      QVector< QgsPointXY > newPoints = polyGeom.randomPointsInPolygon( numberPointsForThisFeature, [=]( const QgsPointXY &newPoint ) -> bool {
       
       .... contents of existing pointOK function sit here      
      
      }, fseed, feedback, maxAttemptsForThisFeature );
```

nyalldawson:
@havatv the changes to QgsGeometry look good, nice work there! 

Now you just need to modify the algorithm to use a lambda function and avoid all the global variables and I'll re-review.

havatv:
I am not familiar with lamda functions, and am still struggling with variable scope, but I have given it a shot, @nyalldawson. It runs fast and fine now. I have updated the test outputs.

havatv:
Only Mac OS build fails now, and I cant find any problems related to the PR in the Mac OS build logs, so I guess what needed to be fixed has been fixed now.

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
```diff
@@ -0,0 +1,394 @@
...
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+
+#include "qgsalgorithmrandompointsinpolygons.h"
+#include "random"
+#include <iostream>
+using namespace std;
```
nyalldawson, line 22: drop

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
```diff
@@ -0,0 +1,394 @@
...
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+
+#include "qgsalgorithmrandompointsinpolygons.h"
+#include "random"
```
nyalldawson, line 20:
```suggestion
#include <random>
```

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
```diff
@@ -0,0 +1,394 @@
...
+    double minDistanceForThisFeature = mMinDistance;
+    if ( mDynamicMinDistance )
+      minDistanceForThisFeature = mMinDistanceProperty.valueAsDouble( expressionContext, minDistanceForThisFeature );
+    const double pointProgressIncrement = featureProgressStep / ( numberPointsForThisFeature * maxAttemptsForThisFeature );
+    double pointProgress = 0.0;
+    unsigned long fseed = uniformIntDist( mt );
+    // Check if we can avoid using the acceptPoint function
+    if ( ( minDistanceForThisFeature == 0 ) && ( mMinDistanceGlobal == 0 ) )
+    {
+      QVector< QgsPointXY > newPoints = polyGeom.randomPointsInPolygon( numberPointsForThisFeature, fseed );
```
nyalldawson, line 275:
```suggestion
      QVector< QgsPointXY > newPoints = polyGeom.randomPointsInPolygon( numberPointsForThisFeature, mUseRandomSeed ? mRandSeed : 0 );
```

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
```diff
@@ -0,0 +1,394 @@
...
+                                          context, ldest, fields, QgsWkbTypes::Point, polygonSource->sourceCrs() ) );
+  if ( !sink )
+    throw QgsProcessingException( invalidSinkError( parameters, OUTPUT ) );
+
+  QgsExpressionContext expressionContext = createExpressionContext( parameters, context, polygonSource.get() );
+
+  // Initialize random engine
+  std::random_device rd;
+  std::mt19937 mt( !mUseRandomSeed ? rd() : mRandSeed );
+  std::uniform_real_distribution<> uniformDist( 0, 1 );
```
nyalldawson, line 204: I don't think we require this -- if an explicit seed isn't set, then we just pass 0 as the seed to randomPointsInPolygon and let it take control of the random selection. If a seed is set, we pass that to randomPointsInPolygon directly

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
```diff
@@ -0,0 +1,394 @@
...
+  mUseRandomSeed = parameters.value( SEED ).isValid();
+  mRandSeed = parameterAsInt( parameters, SEED, context );
+  mIncludePolygonAttr = parameterAsBoolean( parameters, INCLUDE_POLYGON_ATTRIBUTES, context );
+  return true;
+}
+
+QVariantMap QgsRandomPointsInPolygonsAlgorithm::processAlgorithm( const QVariantMap &parameters,
+    QgsProcessingContext &context, QgsProcessingFeedback *feedback )
+{
+  //::mTotNPoints = 0;
```
nyalldawson, line 183: drop

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
```diff
@@ -0,0 +1,394 @@
...
+        pointsAddedForThisFeature++;
+        pointProgress += pointProgressIncrement * ( maxAttemptsForThisFeature );
+      }
+      feedback->setProgress( baseFeatureProgress + pointProgress );
+      continue;
+    }
+    else
+    {
+      // Have to check for minimum distance, provide the acceptPoints function
+      QVector< QgsPointXY > newPoints = polyGeom.randomPointsInPolygon( numberPointsForThisFeature, [ =, &globalIndex, &indexPoints ]( const QgsPointXY & newPoint ) mutable -> bool
```
nyalldawson, line 301:
```suggestion
      QVector< QgsPointXY > newPoints = polyGeom.randomPointsInPolygon( numberPointsForThisFeature, [ & ]( const QgsPointXY & newPoint ) -> bool
```
That will capture all available variables as references (and I think you can drop the mutable)

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
```diff
@@ -0,0 +1,394 @@
...
+        if ( mMinDistanceGlobal != 0.0 && indexPoints > 0 )
+        {
+          QList<QgsFeatureId> neighbors = globalIndex.nearestNeighbor( newPoint, 1, mMinDistanceGlobal );
+          if ( totNPoints > 0 && !neighbors.empty() )
+          {
+            return false;
+          }
+        }
+        // Then local (if larger than global)
+        if ( minDistanceForThisFeature != 0 && mMinDistanceGlobal < minDistanceForThisFeature && localIndexPoints > 0 )
```
nyalldawson, line 315: Possible optimisation opportunity: the local index is always going to be smaller than the global index, so checks against the local index will always be much faster than those made against the global index. If you moved this check to sit before the global index check you may squeeze out some more speed...

nyalldawson:
I can't see immediately what's causing the macos build failure, but my suspicion is that it's the "mutable" keyword on the lambda. Can you try dropping that?

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
```diff
@@ -0,0 +1,394 @@
...
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+
+#include "qgsalgorithmrandompointsinpolygons.h"
+#include "random"
+#include <iostream>
```
havatv, line 21:
```suggestion
```

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
havatv, line 22:
```suggestion
```

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
havatv, line 183:
```suggestion
```

havatv:
I tried to move all the random stuff to geometry.randomPointsInPolygon - the result turned out to not always look very random. The screenshots below shows the result of adding one point and two points per polygon respectively, with 0 local and global minimum distance, and a random seed of 1:

image image

* One point: always placed very close to the edge of the polygon.
* Two points: always placed very close to the edge of the polygon, and also quite close together (compared to the available space).

Could we wait with removing the randomisation performed in the algorithm until we have sorted out possible issues with geometry.randomPointsInPolygon? It is quite easy to remove the randomisation parts, and I will do it as soon as geometry.randomPointsInPolygon produces acceptable results.

havatv:
Tests fail because I accepted some suggestions here, and have not committed the changes I have made locally. I may, as a first step, try to do as I suggested earlier today - keep the random logic and fix the rest.

havatv:
Here is the output when our "extra" randomisation is retained.  The screenshots below shows the result of adding one point and two points per polygon respectively, with 0 local and global minimum distance, and a random seed of 1:
image image

havatv:
@nyalldawson, I noticed recently that you had labelled this as "Frozen". I am not sure what that means.
It should not be necessary to delay because of possible issues with the geometry.randomspointsinpolygons.
And I think that generating a seed for each polygon is not a bad thing. Using the same seed for all polygons leads to patterns, especially when there are many polygons with the same shape.
The failing of the flake8 test seems to be unrelated, and the rest of the checks are now successful.

havatv:
>     * In case you should have any uncertainty, please leave a comment and we will be happy to help you proceed with this pull request.

I am not sure what to do here.
Unit tests are in place, and all tests are successful (except for the flake8 one, which seems to be unrelated, but please correct me if I am wrong).

As far as I can tell, the only objection to the current version of the code is that it uses a second layer of randomisation on top of the *geometry.randomspointsinpolygons* function. That second layer avoids some strange effects when using a seed, as documented above.

roya0045:
2 questions here:
1) Will the point count and distance variable have the options to be data-defined?
2) Would it be possible to add an edge buffer in the process? With the python versions sometimes points tend to be close to the border in some cases and it would be nice to reduce this effect, though I agree that it could be done with a buffer as a preliminary step.

havatv:
> 2 questions here:
> 
>     1. Will the point count and distance variable have the options to be data-defined?

The *Number of points for each feature*, *Minimum distance between points* and *Maximum number of attempts per point* can be data defined.

>     2. Would it be possible to add an edge buffer in the process? With the python versions sometimes points tend to be close to the border in some cases and it would be nice to reduce this effect, though I agree that it could be done with a buffer as a preliminary step.

I thought the general approach was to not duplicate code by re-implementing functionality that is provided by other algorithms. Users will then have to "chain" algorithms to achieve what they want (do a *buffer* / *setback* (buffer with negative distance) or other kind of preparation first, and then use the output as input to the random points in polygons algorithm).

roya0045:
> I thought the general approach was to not duplicate code by re-implementing functionality that is provided by other algorithms. Users should "chain" algorithms to achieve what they want (do a _buffer_ / _setback_ (buffer with negative distance) or other kind of preparation first, and then use the output as input to the random points in polygons algorithm).

This is what I thought and it is reasonable. Thanks for the reply!

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
```diff
@@ -0,0 +1,394 @@
...
+
+  QString ldest;
+  std::unique_ptr< QgsFeatureSink > sink( parameterAsSink( parameters, OUTPUT,
+                                          context, ldest, fields, QgsWkbTypes::Point, polygonSource->sourceCrs() ) );
+  if ( !sink )
+    throw QgsProcessingException( invalidSinkError( parameters, OUTPUT ) );
+
+  QgsExpressionContext expressionContext = createExpressionContext( parameters, context, polygonSource.get() );
+
+  // Initialize random engine
```
nyalldawson, line 198:
```suggestion
  // Initialize random engine -- note that we only use this if the user has specified a fixed seed
```

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
```diff
@@ -0,0 +1,394 @@
...
+    desiredNumberOfPoints += numberPointsForThisFeature;
+    int maxAttemptsForThisFeature = mMaxAttempts;
+    if ( mDynamicMaxAttempts )
+      maxAttemptsForThisFeature = mMaxAttemptsProperty.valueAsInt( expressionContext, maxAttemptsForThisFeature );
+    double minDistanceForThisFeature = mMinDistance;
+    if ( mDynamicMinDistance )
+      minDistanceForThisFeature = mMinDistanceProperty.valueAsDouble( expressionContext, minDistanceForThisFeature );
+    const double pointProgressIncrement = featureProgressStep / ( numberPointsForThisFeature * maxAttemptsForThisFeature );
+    double pointProgress = 0.0;
+    unsigned long fseed = uniformIntDist( mt );
```
nyalldawson, line 269:
```suggestion
```

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
```diff
@@ -0,0 +1,394 @@
...
+    // Check if we can avoid using the acceptPoint function
+    if ( ( minDistanceForThisFeature == 0 ) && ( mMinDistanceGlobal == 0 ) )
+    {
+      QVector< QgsPointXY > newPoints = polyGeom.randomPointsInPolygon( numberPointsForThisFeature, fseed );
```
nyalldawson, line 273:
```suggestion
      QVector< QgsPointXY > newPoints = polyGeom.randomPointsInPolygon( numberPointsForThisFeature, mUseRandomSeed ? uniformIntDist( mt ) : 0 );
```

src/analysis/processing/qgsalgorithmrandompointsinpolygons.cpp
```diff
@@ -0,0 +1,394 @@
...
+          localIndex.addFeature( f );
+          localIndexPoints++;
+        }
+        if ( mMinDistanceGlobal != 0.0 )
+        {
+          globalIndex.addFeature( f );
+          indexPoints++;
+        }
+        return true;
+      }, fseed, feedback, maxAttemptsForThisFeature );
```
nyalldawson, line 345:
```suggestion
      },  mUseRandomSeed ? uniformIntDist( mt ) : 0, feedback, maxAttemptsForThisFeature );
```

nyalldawson:
@havatv thanks for the clarification about the seed handling! Can you test the above suggested changes? That should resolve the issue with repeat random sequences used across multiple features when a fixed seed value is set

havatv:
Thanks, @nyalldawson! Your updates look good to me (all have been accepted), and (as expected) the tests do not seem to be affected.

PR #35748, opened 2020-04-13
Vector tile layer - part 5 (identify tool)

wonder-sk:

This piece adds support for identify map tool, so it is now possible to inspect attributes of features in vector tiles.

image

## Thanks

### DIFF
--- a/src/core/vectortile/qgsvectortileutils.cpp
+++ b/src/core/vectortile/qgsvectortileutils.cpp
@@ -51,7 +51,9 @@ QPolygon QgsVectorTileUtils::tilePolygon( QgsTileXYZ id, const QgsCoordinateTran
 QgsFields QgsVectorTileUtils::makeQgisFields( QSet<QString> flds )
 {
   QgsFields fields;
-  for ( QString fieldName : flds )
+  QStringList fieldsSorted = flds.toList();
+  std::sort( fieldsSorted.begin(), fieldsSorted.end() );
+  for ( QString fieldName : qgis::as_const( fieldsSorted ) )
   {
     fields.append( QgsField( fieldName, QVariant::String ) );
   }
--- a/src/core/vectortile/qgsvectortileutils.cpp
+++ b/src/core/vectortile/qgsvectortileutils.cpp
@@ -53,7 +53,7 @@ QgsFields QgsVectorTileUtils::makeQgisFields( QSet<QString> flds )
   QgsFields fields;
   QStringList fieldsSorted = flds.toList();
   std::sort( fieldsSorted.begin(), fieldsSorted.end() );
-  for ( QString fieldName : qgis::as_const( fieldsSorted ) )
+  for ( const QString &fieldName : qgis::as_const( fieldsSorted ) )
   {
     fields.append( QgsField( fieldName, QVariant::String ) );
   }
--- a/src/gui/qgsmaptoolidentify.cpp
+++ b/src/gui/qgsmaptoolidentify.cpp
@@ -329,11 +329,11 @@ bool QgsMapToolIdentify::identifyVectorTileLayer( QList<QgsMapToolIdentify::Iden
 
   if ( !layer->isInScaleRange( mCanvas->mapSettings().scale() ) )
   {
-    QgsDebugMsg( QStringLiteral( "Out of scale limits" ) );
+    QgsDebugMsgLevel( QStringLiteral( "Out of scale limits" ), 2 );
     return false;
   }
 
-  QApplication::setOverrideCursor( Qt::WaitCursor );
+  QgsTemporaryCursorOverride waitCursor( Qt::WaitCursor );
 
   QMap< QString, QString > commonDerivedAttributes;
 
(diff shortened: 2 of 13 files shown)


src/core/vectortile/qgsvectortileutils.cpp
```diff
@@ -51,7 +51,9 @@ QPolygon QgsVectorTileUtils::tilePolygon( QgsTileXYZ id, const QgsCoordinateTran
 QgsFields QgsVectorTileUtils::makeQgisFields( QSet<QString> flds )
 {
   QgsFields fields;
-  for ( QString fieldName : flds )
+  QStringList fieldsSorted = flds.toList();
+  std::sort( fieldsSorted.begin(), fieldsSorted.end() );
+  for ( QString fieldName : qgis::as_const( fieldsSorted ) )
```
nyalldawson, line 56:
```suggestion
  for ( const QString &fieldName : qgis::as_const( fieldsSorted ) )
```

src/gui/qgsmaptoolidentify.cpp
```diff
@@ -314,6 +322,126 @@ bool QgsMapToolIdentify::identifyMeshLayer( QList<QgsMapToolIdentify::IdentifyRe
...
 }
 
+bool QgsMapToolIdentify::identifyVectorTileLayer( QList<QgsMapToolIdentify::IdentifyResult> *results, QgsVectorTileLayer *layer, const QgsGeometry &geometry )
+{
+  if ( !layer || !layer->isSpatial() )
+    return false;
+
+  if ( !layer->isInScaleRange( mCanvas->mapSettings().scale() ) )
+  {
+    QgsDebugMsg( QStringLiteral( "Out of scale limits" ) );
```
nyalldawson, line 332:
```suggestion
    QgsDebugMsgLevel( QStringLiteral( "Out of scale limits" ), 2 );
```

src/gui/qgsmaptoolidentify.cpp
```diff
@@ -314,6 +322,126 @@ bool QgsMapToolIdentify::identifyMeshLayer( QList<QgsMapToolIdentify::IdentifyRe
...
+    return false;
+  }
+
+  QApplication::setOverrideCursor( Qt::WaitCursor );
```
nyalldawson, line 336: use QgsTemporaryCursorOverride instead of raw cursor override api

src/gui/qgsmaptoolidentify.cpp
```diff
@@ -314,6 +322,126 @@ bool QgsMapToolIdentify::identifyMeshLayer( QList<QgsMapToolIdentify::IdentifyRe
...
+        QByteArray data = layer->getRawTile( tileID );
+        if ( data.isEmpty() )
+          continue;  // failed to get data
+
+        QgsVectorTileMVTDecoder decoder;
+        if ( !decoder.decode( tileID, data ) )
+          continue;  // failed to decode
+
+        QMap<QString, QgsFields> perLayerFields;
+        for ( QString layerName : decoder.layers() )
```
nyalldawson, line 406:
```suggestion
        for ( const QString &layerName : decoder.layers() )
```
(rationale: clazy will raise warnings otherwise)

also need to constify decoder.layers()

src/gui/qgsmaptoolidentify.cpp
```diff
@@ -314,6 +322,126 @@ bool QgsMapToolIdentify::identifyMeshLayer( QList<QgsMapToolIdentify::IdentifyRe
...
+        {
+          QSet<QString> fieldNames = QSet<QString>::fromList( decoder.layerFieldNames( layerName ) );
```
nyalldawson, line 408: QSet::fromList is deprecated

src/gui/qgsmaptoolidentify.cpp
```diff
@@ -314,6 +322,126 @@ bool QgsMapToolIdentify::identifyMeshLayer( QList<QgsMapToolIdentify::IdentifyRe
...
+          perLayerFields[layerName] = QgsVectorTileUtils::makeQgisFields( fieldNames );
+        }
+
+        const QgsVectorTileFeatures features = decoder.layerFeatures( perLayerFields, QgsCoordinateTransform() );
+        for ( QString layerName : features.keys() )
```
nyalldawson, line 413:
```suggestion
        for ( const QString &layerName : features.keys() )
```

nyalldawson (commented):
There's also existing unit test for the identify tool/dialog, which should be extended with a vector tiles test too!

src/gui/qgsmaptoolidentify.cpp
wonder-sk: I am looking into qt docs but I can't see a note it would be deprecated - and how to replace it...?
https://doc.qt.io/qt-5/qset.html#fromList

nyalldawson: This is their guidance:
"Since Qt 5.14, range constructors are available for Qt's generic container classes and should be used in place of this method."

But given that that's 5.14 on, I guess you can ignore it!

wonder-sk: okay cool... yeah I am still at 5.9 :zany_face:

wonder-sk:
Thanks for the review - everything should be addressed now...

wonder-sk:
Hi @jef-n do you have any idea how to get test output from azure runs? I have a failing test now but there is no output whatsover - but it seems that all previous runs on Azure pipelines had no output too (only Travis builds and qgis.norbit.de have test output captured)

https://cdash.orfeo-toolbox.org/testSummary.php?project=3&name=qgis_maptoolidentifyaction&date=2020-04-14

nyalldawson (approved):

PR #35750, opened 2020-04-13
Rename scripts to swap "-" for "_" for consistency

nyalldawson:
We had a confusing and unpredictable mix of script names with - vs _, e.g. `scripts/prepare-commit.sh` vs `scripts/sipify_all.sh`

This unifies with the preferred _ convention for script file names


NathanW2:
+1

.github/CONTRIBUTING.md
```diff
@@ -9,11 +9,11 @@ Proper formatting
 Before making a pull request, please make sure your code is properly formatted
 by running the prepare commit script **before** issuing `git commit`.
 
-    ./scripts/prepare-commit.sh
+    ./scripts/prepare_commit.sh
 
 This can be automated by setting up the pre-commit hook properly.
 
-    ln -s ../../scripts/prepare-commit.sh .git/hooks/pre-commit
+    ln -s ../../scripts/prepare_commit.sh .git/hooks/pre-commit
```
m-kuhn, line 16: This change will make it hard to work on different branches. I would be in favor of leaving a sym link of prepare-commit.sh in place for a transition period at least (i.e. until 3.10 is eol).

nyalldawson: Good point, I've added the link

scripts/update_ts.sh
```diff
@@ -194,7 +194,8 @@ EOF
 
 	QT_INSTALL_HEADERS=$(qmake -query QT_INSTALL_HEADERS)
 
-	echo "TR_EXCLUDE = ${QT_INSTALL_HEADERS%}/*" >>qgis_ts.pro
+	echo "TR_EXCLUDE = ${QT_INSTALL_HEADERS%
```
m-kuhn, line 197: Was this change on purpose? I can't really tell the purpose ...

nyalldawson: Nope, accident

nyalldawson: Fixed

NathanW2:
Nice work.

>

PR #35753, opened 2020-04-13
Fix crash when clicking + in drag and drop form designer and no existing item is selected

nyalldawson:
@3nids i think this needs backporting too?


3nids (approved):
thanks.
yes this needs backporting.

PR #35754, opened 2020-04-14
Fix travis, broken due to some failure to convert struct QVariant property

nirvn:
## Description

This PR fixes broken state of travis. I have *no* idea why the python conversion of the struct ValidationResult's QVariant identifier variable, but well, it fails now.

It also happens to fail on my machine locally.

@nyalldawson , since you pretty much walked me through this, care to review? :)

### DIFF
--- a/python/core/auto_generated/metadata/qgslayermetadatavalidator.sip.in
+++ b/python/core/auto_generated/metadata/qgslayermetadatavalidator.sip.in
@@ -24,19 +24,47 @@ Abstract base class for metadata validators.
 %End
   public:
 
-    struct ValidationResult
-    {
+    class ValidationResult
+{
+%Docstring
+Contains the parameters describing a metadata validation failure.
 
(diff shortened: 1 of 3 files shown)


nyalldawson (approved):

nyalldawson:
LGTM

nirvn:
For those interested in this mystery.

Package used in a successful travis run (from ~2 days ago):
-- Found Qt version: 5.9.5
-- Found Python version: 3.6.9
-- Found PyQt5 version: 5.10.1
-- Found SIP version: 4.19.7

Packages used in a failing travis run (from today):
-- Found Qt version: 5.9.5
-- Found Python version: 3.6.9
-- Found PyQt5 version: 5.10.1
-- Found SIP version: 4.19.7

I.e., nothing has changed there that could cause a regression. 

On top of which, QGIS itself hasn't seen a code change in that part of its code for a while.

So, we'll send this one to the X Files and move on.

PR #35771, opened 2020-04-14
old Qgis2.18 slot in QgsAttributeFormlegacyInterface 

SebastienPeillet:
## Description

Working on an old project that used python script action for opening attribute form. This kind of action use the QgsAttributeFormLegacyInterface class. 

And apparently QgsAttributeFormLegacyInterface is not very used as there is still Qgis 2.18 slot in QgsAttibuteform class.

accept() / reject() slots were deprecated and has been changed to save() / resetValues()

doc 2.18 : [https://qgis.org/api/2.18/classQgsAttributeForm.html](https://qgis.org/api/2.18/classQgsAttributeForm.html)

doc 3.10 : [https://qgis.org/api/3.10/classQgsAttributeForm.html](https://qgis.org/api/3.10/classQgsAttributeForm.html)

Should be backported, even if it is not a very used feature.


nyalldawson:
Does this still work correctly if you upgrade it to a new style connection?

SebastienPeillet:
Yes. Actually, in the current state, it doesn't break the attribute form, it only raises a critical warning : 

```
2020-04-17T10:00:34     CRITICAL    Qt : QObject::disconnect: No such slot QgsAttributeForm::accept() in ~/release_3.10/qgis/src/gui/qgsattributeformlegacyinterface.cpp:54
2020-04-17T10:00:34     CRITICAL    Qt : QObject::disconnect: (sender name: 'buttonBox')
2020-04-17T10:00:34     CRITICAL    Qt : QObject::connect: No such slot QgsAttributeForm::accept() in ~/release_3.10/qgis/src/gui/qgsattributeformlegacyinterface.cpp:55
2020-04-17T10:00:34     CRITICAL    Qt : QObject::connect: (sender name: 'buttonBox')
```

I don't know why we disconnect/reconnect `accepted` signal in the first place, but as it doesn't disconnect save slot, there is no effect on the attribute form.

Changing `accept` slot to `save` slot removes the warning and the attribute form works well too.

nyalldawson:
In that case can you upgrade this to a new style connection so we get the compile time errors if it breaks in future?

nyalldawson (approved):

Issue #35773, opened 2020-04-14
[processing modeler] Small graphical glitch: links connect in the wrong direction


olivierdalang:
**Describe the bug**

Since a few nightlies, there's a small graphical glitch in the graphical modeler, where some of the beziers connect in the wrong direction.

![image](https://user-images.githubusercontent.com/1894106/79255762-21255180-7e87-11ea-9d13-5db3d17fecf2.png)

I think it comes from this change: 

**How to Reproduce**

- in a new model, create an input and an algorithm, and connect the input to a parameter of the algorithm
- move the algorithm box to the bottom left of the input

(the same happens in other circumstances, for example with parent/child links)

**QGIS and OS versions**

master (probably since this : https://github.com/qgis/QGIS/commit/3976e14ae85879e6f056379405bc96b84e2497a8#diff-eae583913502c32a9b22a1ad9dcdcf30L172-R184 )

PR #35774, opened 2020-04-14
[processing modeler] Show arrowheads for parent/child relations

olivierdalang:
## Description

This is a semi-revival of https://github.com/qgis/QGIS/pull/34689.

This PR adds an arrow head to the "parent" links in the modeler, so on can tell the direction of the relation.

Before :
image

After :
image

(note the screenshot includes a quick fix for unrelated issue https://github.com/qgis/QGIS/issues/35773)

This also goes one (small) step towards the goal of the previous PR in better distinguishing the parent/child links from the normal links. I'd go even further (maybe a ligther line, now that dotted lines are taken for comments), but am OK if you're sure it's enough like this.

By the way, great work Nyall on the new modeler code :-)


nyalldawson (commented):
I like this, thanks!

I think for API flexibility it would be nicer to add an enum to QgsModelArrowItem, something like:
   
    //! Marker styles for start and end of line markers
    enum Marker
    {
        MarkerEllipse, //!< Circle marker
        MarkerArrowHead, //!< Arrowheader marker
    };

and then modify the constructor to add `Marker startMarker, .... Marker endMarker` instead of the `bool showArrowHead` approach.

This would make it simpler to extend in future if we decide we need additional marker types. What do you think?

olivierdalang:
Yes I agree it's better. Something like this ?

(don't merge right away as I haven't tested yet, I'm having some unrelated build issues)

src/gui/processing/models/qgsmodelarrowitem.cpp
```diff
@@ -83,9 +87,38 @@ void QgsModelArrowItem::paint( QPainter *painter, const QStyleOptionGraphicsItem
   painter->setBrush( color );
   painter->setRenderHint( QPainter::Antialiasing );
 
-  for ( const QPointF &point : qgis::as_const( mNodePoints ) )
+  if ( mStartMarker == Marker::Circle )
```
nyalldawson, line 90: use a switch statement here instead (if you omit the "default:" branch, then the compiler will raise warnings if someone ever adds another marker type, giving them a direct pointer to all the code paths which need to be updated as a result!)

src/gui/processing/models/qgsmodelarrowitem.cpp
```diff
@@ -83,9 +87,38 @@ void QgsModelArrowItem::paint( QPainter *painter, const QStyleOptionGraphicsItem
...
-    painter->drawEllipse( point, 3.0, 3.0 );
+    QPointF delta = path().pointAtPercent(1.0) - path().pointAtPercent(0.95);
+    float angle = atan2( delta.y(), delta.x() ) * 180.0 / M_PI;
+    painter->translate(mEndPoint);
+    painter->rotate(angle);
+    QPolygonF arrowHead;
+    arrowHead << QPointF(0, 0) << QPointF(-6, 4) << QPointF(-6, -4) << QPointF(0, 0);
+    painter->drawPolygon( arrowHead );
+    painter->rotate(-angle);
+    painter->translate(-mEndPoint);
```
nyalldawson, line 121: There's a lot of duplicate code here -- could you extract this to a function?

nyalldawson (commented):
Looks great -- this is a very nice addition!

alexbruy:
Not directly related to this PR, but I'm wondering if we should make at least some connection lines straight instead of curly.

PR #35778, opened 2020-04-15
Allow reordering model inputs

nyalldawson:
Instead of forcing a quasi-random ordering of inputs for models, this commit exposes a new "Reorder Model Inputs" option in the model designer which allows users control over the exact order of inputs to show users for their model.

No more illogical ordering like showing a field choice before the layer choice it's based on!

Sponsored by NaturalGIS


alexbruy (approved):
Cool!

Should we allow selecting and moving multiple items at once?

nyalldawson:
> Should we allow selecting and moving multiple items at once?

The model logic for this gets really complex very quick -- so I'd prefer not unless it's absolutely required!

PR #35793, opened 2020-04-15
Offer a per feature, as well as a global minimum distance parameter, for the Random points on lines algorithm

havatv:
## Description

"Splits" the *Minimum distance between points* parameter into *Minimum distance between points (per feature)* and *Global minimum distance between points*. The last one can not be data-defined, since it would be strange to have a global parameter change on a per feature basis.

The split is important, because it makes it possible to apply the minimum distance only "locally" (for each feature), meaning that points added to features close by are not considered when checking minimum distance (only the points added for the current line feature are checked).

It is meaningful to combine the two minimum distance parameters, but if the global one is larger than the local one, the local one will have no effect.

New tests have been added, and the existing ones have been updated to reflect the change.


src/analysis/processing/qgsalgorithmrandompointsonlines.cpp
```diff
@@ -94,38 +99,54 @@ void QgsRandomPointsOnLinesAlgorithm::initAlgorithm( const QVariantMap & )
...
-                      "<ul><li>For each feature in the <i><b>Input line layer</b></i>, the algorithm attempts to add "
-                      "the specified <i><b>Number of points for each feature</b></i> to the output layer.</li> "
-                      "<li>A <i><b>Minimum distance between points</b></i> can be specified.<br> "
-                      "A point will not be generated if there is an already generated point "
-                      "(on any line feature) within this (Euclidean) distance from "
-                      "the generated location. "
-                      "If the <i><b>Minimum distance between points</b></i> is too large, it may not be possible to generate "
-                      "the specified <i><b>Number of points for each feature</b></i>.</li> "
+                      "on the lines of the <i>Input line layer</i>. "
+                      "The default behaviour is that the generated point features inherit "
```
havatv, line 110:
```suggestion
                      "The default behavior is that the generated point features inherit "
```

nyalldawson (approved):

havatv:
Thanks, @nyalldawson! I will update the documentation.

PR #35794, opened 2020-04-15
Fix csv import sort order

### LINKED ISSUE #32250: Value map: Sort order & CSV import
**Keep sort order of original CSV File**
Under "Layer properties" > "Attributes Form" there is the "value map" widget. It is possible to load value-description data from CSV. Currently the original ordering (as it is in the csv) is not kept but changed upon import. Imported data are sorted by value. It would be nice if the original ordering (as it is in the csv) would be kept. The following picture shows the current (not ideal situation):
grafik

m-kuhn:
Fixes

- do not inverse the order of csv values in the value map widget configuration
- do not skip first line of the first line in csv imports
- improves the title of the csv import dialog

Fixes https://github.com/qgis/QGIS/issues/32250

### DIFF
--- a/src/gui/editorwidgets/qgsvaluemapconfigdlg.cpp
+++ b/src/gui/editorwidgets/qgsvaluemapconfigdlg.cpp
@@ -179,6 +179,7 @@ void QgsValueMapConfigDlg::updateMap( const QList<QPair<QString, QVariant>> &lis
       setRow( row, pair.first, QString() );
     else
       setRow( row, pair.first, pair.second.toString() );
+    ++row;
   }
 }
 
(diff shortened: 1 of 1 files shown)


nyalldawson (approved):

jpnunn:
Hi - i have noticed that this has re-appeared.  I have been attempting to import a CSV that is in a non - alphabetcial order (which i require) but it imports it in an alphabetcial order.

m-kuhn:
> Hi - i have noticed that this has re-appeared. I have been attempting to import a CSV that is in a non - alphabetical order (which i require) but it imports it in an alphabetical order.

@jpnunn can you open a new issue with details, version and reproduction steps? Thank you

Issue #35786, opened 2020-04-15
Scale dependent snapping has min and max inverted


uclaros:
**Describe the bug**
Since scale is a fraction `1:500 > 1:1000000`, min scale should be the most zoomed out scale, max scale should be the most zoomed in scale, just like scale dependent layer visibility does it.

Also the tooltip should state "zoomed-in" and "zoomed-out" as in scale dependent visibility to avoid any further confusion.

roya0045:
Sometime people don't think of the '1:' part and just focus on the number. Bigger number = max smaller number = min. So this can also be understood, especially when only the divider is provided by the user.

lbartoletti:
cc @obrix @troopa81

obrix:
Hello and thank you @uclaros. Yep you are right, in the code only the denominator is handled, this is where the error comes from. The behavior should be coherent with the other qgis features so I will change it.
Labeling should also be like the one from the scale dependent layer visibility feature to avoid any confusion for the user.

uclaros:
This is fixed in #35798 so I guess it can close.

roya0045:
@uclaros you have the power to close this if you want.

uclaros:
Doh!

Issue #35789, opened 2020-04-15
Scale dependent snapping disabled drop downs


uclaros:
**Describe the bug**
When switching *Scale dependent snapping* from *per layer* to *disabled* while the min or max drop down is focused, the widget is not disabled and its values can be be changed.
Additionally, when switching to *per layer* sometimes the min and max drop down widgets are not enabled and can't be clicked.

See the following screencast for both (clicking is not visible, but you can presume when I clicked)
![Peek 2020-04-15 13-24](https://user-images.githubusercontent.com/11358178/79328171-49ac5a80-7f1e-11ea-8694-b7927b545e28.gif)

obrix:
I can reproduce the first bug but cannot reproduce the second one yet.
Anyway I rewrote the way the view is refreshed by resetting the tree view when the snapping scale mode is changed. This fix the first one for sure, it may fix the second one also. I will open a pull request.

uclaros:
The second one appeared while recording the screencast for the first, I hadn't seen it till then. I can't reproduce it either!

PR #35798, opened 2020-04-15
Scale dependent snapping bugfix

obrix:
Should fix issues #35786 and #35789.

For the first issue I adapted the behavior of min and max scale parameters to match what is done in other places in qgis : min scale is the most zoomed out scale, max scale the most zoomed in. The tool tip have also been improved.

The second bug was an issue with the way the tree view for the snapping config per layer is refreshed when the snapping scale mode is changed. I now call reset on the tree view when the mode is changed, the strange combo box behavior disappeared.

### DIFF
--- a/python/core/auto_generated/qgssnappingconfig.sip.in
+++ b/python/core/auto_generated/qgssnappingconfig.sip.in
@@ -285,7 +285,7 @@ Sets the tolerance
 
     double minimumScale() const;
 %Docstring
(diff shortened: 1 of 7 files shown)


obrix:
Hello @3nids , would you have the time to look at these two small bugfixes ?

mhugo:
Thanks for the fixes.

PR #35799, opened 2020-04-15
Try to fix mxe (again)

elpaso:


roya0045:
Hey it worked! Nice work on the follow up. Though I'm not sure why the logs were buggy in qgis.

.github/workflows/mxe.yml
```diff
@@ -1,12 +1,12 @@
 name: Windows cross build with MXE
 
-on: [push]
+on: [push, pull_request]
```
nyalldawson, line 3: good move!

nyalldawson (approved):

nirvn:
@elpaso , our 3.12 branch seems to show MXE build broken there too, do we need to backport?

elpaso:
@nirvn probably yes, I didn't check if it broke for the same reason (python dependency in WITH_QGIS_PROCESS)

roya0045:
Welp now MXE is running out of space when getting dependencies...

elpaso:
Maybe temporary, I'm experiencing various failures today from other CI systems too.

roya0045:
@elpaso possible, usually there should more room and no big dependency was added recently.

PR #35806, opened 2020-04-16
Snapping type menu fixes

uclaros:
## Description
Keep the snapping type menu open when a type is checked / unchecked so that they can be easily toggled.

Also, in advanced configuration, display the type initials when more than one snapping type is selected, instead of the bugged ", , ...".

before:

Peek 2020-04-16 02-59

after:

Peek 2020-04-16 02-55


nirvn:
Thanks for fixing the advanced config issue. 

Is there anything we could do here to fix the lack of any string on the combo box itself? Even a placeholder "click to set items" would look better than a void :)

src/app/qgssnappingwidget.cpp
```diff
@@ -44,6 +44,25 @@
 #include "qgsscalewidget.h"
 
 
+class SnapTypeMenu: public QMenu
+{
+  public:
+    SnapTypeMenu( const QString &title, QWidget *parent = nullptr )
+    {
+      QMenu( title, parent );
+    }
```
3nids, line 53:
```suggestion
      : QMenu( title, parent ) {}
```

src/app/qgssnappingwidget.cpp
```diff
@@ -44,6 +44,25 @@
...
+    void mouseReleaseEvent( QMouseEvent *e )
+    {
+      QAction *action = activeAction();
+      if ( action )
+      {
+        action->trigger();
+      }
```
3nids, line 60:
```suggestion
        action->trigger();
```

3nids (commented):
good work!
some small code layout fixes.
Thanks.

uclaros:
Fixed the code layout issues on both files.

The second commit is an attempt to remedy the "empty combobox" that @nirvn mentioned.
I tried setting a placeholder text but it's ugly because it also appears when the button is pressed, so I tried to skip showing it at all. Now when the menu closes, focus goes directly to its parent so the list of checked snapping types is immediately displayed instead of the empty toolbutton.

Peek 2020-04-17 03-18

I also tried to intercept the toolbutton's show event to trigger automatically the menu so that the button is never displayed blank. It worked, but it broke the previous focus on hide trick so I didn't keep it.

Now, I wonder how the comma separated initials appear in translations. In English it's a perfect fit, all initials are unique and one can easily understand the snapping type it describes, but in other languages it might not be the same...

uclaros:
Is translation a deal breaker here? Maybe add a method `QgsSnappingConfig::snappingTypeFlagToChar` that returns one of `V,S,A,C,M`, or some unicode glyph?

m-kuhn:
@nirvn @3nids is there anything left to do here?

nirvn:
I think it's ready to merge.

PR #35807, opened 2020-04-16
[TEMPORAL] allow temporal layout item map

vcloarec:
This PR allows the possibility, in the API (only for now), to set temporal the layout map item (2D and 3D).
If the layout map is temporal, then temporal layer can be rendered depending on the time range.
For now, this can be done with python but in the future, this maybe could be done through layout map widget settings or other GUI way.

### DIFF
--- a/python/core/auto_generated/layout/qgslayoutitemmap.sip.in
+++ b/python/core/auto_generated/layout/qgslayoutitemmap.sip.in
@@ -9,7 +9,7 @@
 
 
 
-class QgsLayoutItemMap : QgsLayoutItem
+class QgsLayoutItemMap : QgsLayoutItem, QgsTemporalRangeObject
 {
 %Docstring
 Layout graphical items for displaying a map.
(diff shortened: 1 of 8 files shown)


nyalldawson:
Nice!

Won't we also need to save/restore the temporal settings in the layout item?

vcloarec:
@nyalldawson good question!
For now, the implementation is minimal to allow Crayfish plugin to render animation, the layout are not kept after processing. Not sure of the following direction in the future... But I guess yes, we do

vcloarec:
done, but I wonder if we not have to write/read xml in the `QgsTemporalRangeObject` class

nyalldawson:
Looks good to me! Can we get some tests too?

PeterPetrik:
@nyalldawson for tests, in which file they should be added?

vcloarec:
test added and small issues fixed.

nyalldawson (approved):

PR #35808, opened 2020-04-16
New modeler algorithm for creating conditional branches

nyalldawson:
This algorithm allows useres to setup multiple conditions (via qgis expressions), which cause their corresponding branch of the model to be run or skipped depending on the result of the expression evaluation.

Sponsored by Andreas Neumann
Peek 2020-04-16 12-56


alexbruy (approved):
Great!

PR #35809, opened 2020-04-16
Add a super-basic way to pick legend patch shapes

nyalldawson:
And make legend node properties (such as renaming legend nodes) in layouts open in an inline panel.

This provides us with a place to insert more useful per-node properties for greater legend customisation power. For now, I've added a super-basic choice of legend patch shapes (With the only choices so far being the default or oval for fills, and default or "zig zag" for lines)

### DIFF
--- a/src/core/layertree/qgslegendpatchshape.cpp
+++ b/src/core/layertree/qgslegendpatchshape.cpp
@@ -67,6 +67,19 @@ QPolygonF lineStringToQPolygonF( const QgsLineString *line )
   return thisRes;
 }
 
+QPolygonF curveToPolygonF( const QgsCurve *curve )
+{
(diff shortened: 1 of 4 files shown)


roya0045:
Nice! Should this be flagged as feature and need docs since it will affect the UI?

nyalldawson:
> Nice! Should this be flagged as feature and need docs since it will affect the UI?

I consider this just a "toy" for now, and don't really want it flagged by the documentation team. I plan on following this up by exposing patch shapes to the style manager and improving this ui, so at that stage i'll tag it :)

roya0045:
@nyalldawson Ah very well, I was curious about the "omissions"!
 I am also curious if you plan to implement a way to rescale the box. In most case this wont be needed with relative graphic elements but with symbols using map units as the scale, i could come in handy.

PR #35811, opened 2020-04-16
[Backport 3.10] Value map: Sort order & CSV import

m-kuhn:
Backport #32250


nyalldawson:
Should we delay this by one cycle? I'm thinking a new policy of delaying non critical, non trivial fixes by one cycle could be a good idea for ltr. What do you think?

nyalldawson:
That said, this quite clearly **is** trivial 😁

m-kuhn:
> I'm thinking a new policy of delaying non critical, non trivial fixes by one cycle could be a good idea for ltr.

+1 - For reference: I already asked for this on several recent backports even without an explicit policy

> Should we delay this by one cycle?

Happy to do so and wait for feedback from master tests

nirvn (approved):

PR #35819, opened 2020-04-16
[TEMPORAL][MESH] Update temporal extent when mesh layer becomes temporal for first time

vcloarec:
Before this PR, when a non temporal mesh layer was loaded, the temporal extent was logically unchanged. But, after, when a temporal dataset was loaded, the layer stay non temporal and the temporal extent was not updated, event if it was the first temporal datasets.
With this PR, the temporal extent is updated when the first temporal dataset is loaded (if the mesh layer is non temporal before dataset loading).

### DIFF
--- a/src/app/mesh/qgsmeshlayerproperties.cpp
+++ b/src/app/mesh/qgsmeshlayerproperties.cpp
@@ -34,6 +34,7 @@
 #include "qgsprojectionselectiondialog.h"
 #include "qgsrenderermeshpropertieswidget.h"
 #include "qgssettings.h"
+#include "qgsprojecttimesettings.h"
 #include "qgsproviderregistry.h"
 #ifdef HAVE_3D
 #include "qgsmeshlayer3drendererwidget.h"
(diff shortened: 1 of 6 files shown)


src/gui/qgstemporalcontrollerwidget.cpp
```diff
@@ -160,7 +160,23 @@ void QgsTemporalControllerWidget::onLayersAdded()
     for ( QgsMapLayer *layer : layers )
     {
       if ( layer->temporalProperties() )
+      {
         mHasTemporalLayersLoaded |= layer->temporalProperties()->isActive();
+
+        if ( !mHasTemporalLayersLoaded )
+        {
+          //inspired by  QgsLayerTreeMapCanvasBridge::layersAdded
```
PeterPetrik, line 168: remove the comment

src/app/mesh/qgsmeshlayerproperties.cpp
```diff
@@ -229,11 +230,27 @@ void QgsMeshLayerProperties::addDataset()
...
+    {
+      mMeshLayer->temporalProperties()->setDefaultsFromDataProviderTemporalCapabilities(
+        mMeshLayer->dataProvider()->temporalCapabilities() );
+
+      if ( ! mMeshLayer->temporalProperties()->referenceTime().isValid() )
+      {
+        QDateTime referenceTime = QgsProject::instance()->timeSettings()->temporalRange().begin();
+        if ( !referenceTime.isValid() ) // If project reference time is invalid, use current date
+          referenceTime = QDateTime( QDate::currentDate(), QTime( 0, 0, 0, Qt::UTC ) );
+        mMeshLayer->temporalProperties()->setReferenceTime( referenceTime, mMeshLayer->dataProvider()->temporalCapabilities() );
```
PeterPetrik, line 247: ins't this code snipped already used somewhere else so it can be a new private function?

src/core/mesh/qgsmeshlayer.cpp
```diff
@@ -72,12 +72,15 @@ void QgsMeshLayer::setDefaultRendererSettings()
...
+    // If the first dataset group has nan min/max, display the mesh to avoid nothing displayed
     QgsMeshDatasetGroupMetadata meta = mDataProvider->datasetGroupMetadata( 0 );
-    if ( ( meta.maximum() == std::numeric_limits<double>::quiet_NaN() &&
-           meta.minimum() == std::numeric_limits<double>::quiet_NaN() ) ||
-         meta.isTemporal() )
+    if ( meta.maximum() == std::numeric_limits<double>::quiet_NaN() &&
+         meta.minimum() == std::numeric_limits<double>::quiet_NaN() )
       meshSettings.setEnabled( true );
+
+    //if the mesh is non temporal, set the static scalar dataset
```
PeterPetrik, line 81: missing space in the comment

PeterPetrik:
looks good to me

src/app/mesh/qgsmeshlayerproperties.cpp
vcloarec: here : https://github.com/qgis/QGIS/blob/e9dd602da65cdd4c658e4e9b829f8f3fbd59ac40/src/app/qgisapp.cpp#L5487
but the not exactly the same ans I think it is difficult to have an efficient refactoring.

PeterPetrik: ok

PeterPetrik (approved):

PR #35821, opened 2020-04-17
Support pasting pictures directly into layouts

nyalldawson:
Embeds the pasted picture into the layout.

Note that Qt image clipboard handling seems a bit... crap. So success may vary.


nirvn (approved):
Cool stuff.

PR #35835, opened 2020-04-17
[feature] Processing add random and normal raster algorithms

root676:
## Description

This PR introduces two processing algorithms for **random raster creation**:

- **Create random raster**: Creates uniformly distributed random raster datasets in the desired extent, raster data type and cell size including parameters for upper and lower bounds to define a range for random values.
- **Create normal raster**: Creates normally distributed random raster datasets in the desired extent, raster data type and cell size including parameters for mean and standard deviation for the normally distributed random values.

In general, the PR enhances the raster creation capabilities of QGIS so that we can offer the same raster creation functionalities as ArcGIS with the Spatial Analyst extension.

The PR also moves the **Create constant raster** algorithm into the new alg-group _Raster creation_. 

I am looking forward to comments/reviews!

For reference: Screenshot of a 250m cellsize random raster with uniformly distributed values between 1 and 100 (integer32) - Disclaimer: Could make a little dizzy ;)
randomRaster_1_100


src/analysis/processing/qgsalgorithmnormalraster.cpp
```diff
@@ -0,0 +1,187 @@
...
+      break;
+    }
+    //prepare raw data depending on raster data type
+    QgsRasterBlock block( mRasterDataType, cols, 1 );
+    switch ( mTypeId )
+    {
+      case ( 0 ):
+      {
+        std::vector<float> float32Row( cols );
+        for ( int i = 0; i < cols; i++ )
```
nyalldawson, line 158: Don't reuse `i` here -- it makes for confusing code. Better to explicitly use `row`/`col` instead

src/analysis/processing/qgsalgorithmnormalraster.cpp
```diff
@@ -0,0 +1,187 @@
...
+        {
+          float32Row[i] = static_cast<float>( mNormalDoubleDistribution( mRandomDevice ) );
```
nyalldawson, line 160: I don't think this is quite correct use of the random classes. I think you should be using mMersenneTwister here, having first seeded mMersenneTwister with mRandomDevice...

nyalldawson: See the example from https://en.cppreference.com/w/cpp/numeric/random/normal_distribution:

```
std::random_device rd{};
    std::mt19937 gen{rd()};
 
    // values near the mean are the most likely
    // standard deviation affects the dispersion of generated values from the mean
    std::normal_distribution<> d{5,2};
 
    std::map<int, int> hist{};
    for(int n=0; n<10000; ++n) {
        ++hist[std::round(d(gen))];
    }
```

src/analysis/processing/qgsalgorithmnormalraster.cpp
```diff
@@ -0,0 +1,187 @@
...
+  }
+  else
+  {
+    mRasterDataType = Qgis::Float64;
+  }
+
+  mMean = parameterAsDouble( parameters, QStringLiteral( "MEAN" ), context );
+  mStdev = parameterAsDouble( parameters, QStringLiteral( "STDEV" ), context );
+
+  mNormalDoubleDistribution = std::normal_distribution<double>( mMean, mStdev );
```
nyalldawson, line 116: hard to see if there's issues using these classes across threads -- I'd play it safe and only create them in processAlgorithm

src/analysis/processing/qgsalgorithmrandomraster.cpp
```diff
@@ -0,0 +1,313 @@
...
+    //prepare raw data depending on raster data type
+    QgsRasterBlock block( mRasterDataType, cols, 1 );
+    switch ( mTypeId )
+    {
+      case ( 0 ):
+      {
+        std::vector<quint8> byteRow( cols );
+        for ( int i = 0; i < cols; i++ )
+        {
+          byteRow[i] = static_cast<quint8>( mRandomIntDistribution( mRandomDevice ) );
```
nyalldawson, line 236: see above note about use of the c++ random classes

nyalldawson (commented):
Cool feature! While in depth unit tests obviously aren't possible here, I'd still like to see some tests which just verify that the output has been created and is of the desired type.

nyalldawson:
Looking at https://pro.arcgis.com/en/pro-app/tool-reference/data-management/create-random-raster.htm - it seems arc offers choices of other distributions too. Some of these look trivial to add here -- e.g. exponential would be just a matter of using https://en.cppreference.com/w/cpp/numeric/random/exponential_distribution instead of normal_distribution. Do you think it would be worth exposing this as an advanced parameter for the alg, or would it be better to expose other distributions as unique dedicated algorithms?

src/analysis/processing/qgsalgorithmnormalraster.cpp
root676: Ah yes, I go this mixed up with all the abbreviations in the example code ;)

root676:
Thanks for the review! Very odd... I thought that ArcMap 10.x + Spatial Analyst feature the same algorithms as ArcGIS Pro (there are only random and normal raster in [spatial analyst](https://desktop.arcgis.com/en/arcmap/latest/tools/spatial-analyst-toolbox/an-overview-of-the-raster-creation-tools.htm) . But thanks for the hint - we should definitely add more of the available C++ random number distribution classes! I think it would be best to add them as separate algorithms as some of them only allow integer/floating point numbers (this is one of the reasons why I split the random and normal raster creation into two algs). Some distributions use a different number/type of arguments too --> separate algorithm :+1: . I think, I'll address them in a separate PR.

src/analysis/processing/qgsalgorithmnormalraster.cpp
root676: fixed

src/analysis/processing/qgsalgorithmnormalraster.cpp
root676: fixed

src/analysis/processing/qgsalgorithmnormalraster.cpp
root676: fixed

src/analysis/processing/qgsalgorithmrandomraster.cpp
root676: fixed

root676:
@nyalldawson  I just added the tests for both algorithms. The random raster test now also tests if the random values are inside the specified range. The other values are not compared, they are... well, just random... :laughing:

nyalldawson:
> I just added the tests for both algorithms. The random raster test now also tests if the random values are inside the specified range. T

Perfect, nice work!

nyalldawson (approved):

nyalldawson:
>  think it would be best to add them as separate algorithms 

I agree - that sounds best to me too

Issue #28861, opened 2019-01-20
Display layer name/title/abstract/keyword clearly for WMS


qgib:
Author Name: **Jonathan Moules** (Jonathan Moules)
Original Redmine Issue: [21042](https://issues.qgis.org/issues/21042)

Redmine category:web_services_clients/wms

---
This WMS has taken the trouble to write a long Abstract:
http://ogc.geo-ide.developpement-durable.gouv.fr/wxs?map=%2Fopt%2Fdata%2Fcarto%2Fgeoide-catalogue%2F1.4%2Forg_38102%2F2f3f11f0-91c9-479a-b748-ed1098603e91.internet.map&version=1.3.0

But there's no way to see it in QGIS. Similarly the long name and long title are also hidden, despite the fact I'm viewing this on a 1600pixel wide monitor.
Similarly, as far as I can see, layer keywords are not displayed at all (although the ones on that layer are fairly meaningless), nor is any service level info (service name, title, abstract, keywords).
Could not a dialog on the right side of the WMS panel be used to display the given information for the selected layer?

alexbruy:
There are tooltips for abstract and title. Is this not enough?

PR #35838, opened 2020-04-18
various fixes to connections dialogs in Datasource Manager

### LINKED ISSUE #26038: Do not enable OK buttons if connection dialogs are missing required fields
Author Name: **Johannes Kroeger** (Johannes Kroeger)
Original Redmine Issue: [18145](https://issues.qgis.org/issues/18145)
Affected QGIS version: 3.6.0
Redmine category:gui

---
I named this bug generally although I only have two specific examples.

#26037 mentions a dialog presenting OK and Cancel, both allowing to quit doing nothing.

The XYZ Connection dialog also let's the user proceed without entering a name. Without a name no connection will get saved so the entered detail is lost if the user clicks OK.

This is also possible with the MSSQL, PostGIS, DB2 dialogs.

It is not working (i.e. set up great) for example in the ArcGISFeature- and -MapServer as well as WCS, WMS, WFS.

Please only enable OK buttons if things are OK. Otherwise do not let the user click OK.

Not sure if this style of issue is good, close at will (or if the examples are fixed). ;)

---

- [should not be able to click ok just now.png](https://issues.qgis.org/attachments/download/12197/should%20not%20be%20able%20to%20click%20ok%20just%20now.png) (Johannes Kroeger)

### LINKED ISSUE #29873: WMS "Add Default Servers" button removal
Author Name: **Jonathan Moules** (Jonathan Moules)
Original Redmine Issue: [22059](https://issues.qgis.org/issues/22059)
Affected QGIS version: 3.4.3
Redmine category:web_services_clients/wms

---
Per my enquiry to the list, the Add Default Servers button adds servers that are likely useless to the end user (software Demo servers with no meaningful data).
A couple folks suggested removing it.

alexbruy:
## Description

- take into account configured pixel ratio for XYZ connections when performing connections import/export. Previously this setting was ignored and configured value lost.
- Ok button in new connection dialog for OGR, PostGIS, Oracle, MSSQL, DB2, XYZ and Vector tile providers enabled only if all required fields are not empty preventing loosing data. Fixes #26038.
- disable editing and export buttons in the datasource select dialogs if no connections are available, having these buttons enabled makes no sense in such case.
- remove "Add default servers" button from the WMS page in the Datasource Manager. Most of the default servers are not working, the rest have very limited usage during last discussion in the mailing list couple of people suggested to remove it. Fixes #29873.
- use already available "Referer" field from the base widget in the new WMS connection dialog instead of custom field.
- add tooltip to the WMS layer title to allow displaying long titles (refs #28861)

### DIFF
--- a/src/gui/ogr/qgsnewogrconnection.h
+++ b/src/gui/ogr/qgsnewogrconnection.h
@@ -52,6 +52,7 @@ class GUI_EXPORT QgsNewOgrConnection : public QDialog, private Ui::QgsNewOgrConn
 
   public slots:
     void accept() override;
+    //! Updates state of the OK button depending of the filled fields
     void updateOkButtonState();
 
   private slots:
--- a/src/gui/ogr/qgsnewogrconnection.h
+++ b/src/gui/ogr/qgsnewogrconnection.h
@@ -52,11 +52,11 @@ class GUI_EXPORT QgsNewOgrConnection : public QDialog, private Ui::QgsNewOgrConn
 
   public slots:
     void accept() override;
-    //! Updates state of the OK button depending of the filled fields
-    void updateOkButtonState();
 
   private slots:
     void btnConnect_clicked();
+    //! Updates state of the OK button depending of the filled fields
+    void updateOkButtonState();
 
   private:
     QString mOriginalConnName;
--- a/src/gui/ogr/qgsnewogrconnection.h
+++ b/src/gui/ogr/qgsnewogrconnection.h
@@ -52,6 +52,7 @@ class GUI_EXPORT QgsNewOgrConnection : public QDialog, private Ui::QgsNewOgrConn
 
   public slots:
     void accept() override;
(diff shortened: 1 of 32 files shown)


src/gui/ogr/qgsnewogrconnection.h
```diff
@@ -52,6 +52,8 @@ class GUI_EXPORT QgsNewOgrConnection : public QDialog, private Ui::QgsNewOgrConn
 
   public slots:
     void accept() override;
+    //! Updates state of the OK button depending of the filled fields
```
m-kuhn, line 55: Is this public on purpose?
If yes, add a `\since`

m-kuhn (approved):
Nice !!

src/gui/ogr/qgsnewogrconnection.h
alexbruy: No, it should be private. Fixed now.

Thanks!

PR #35839, opened 2020-04-18
ascii expression

JanCaha:
Adds ascii expression, which returns unicode code for the first character in a string. Contains tests and documentation.

Follows PostgreSQL definition of the function.


resources/function_help/json/ascii
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "ascii",
+  "type": "function",
+  "description": "Returns the unicode code associated with a first character of the string.",
```
nyalldawson, line 4: ### 
```suggestion
  "description": "Returns the unicode code associated with the first character of a string.",
```

src/core/expression/qgsexpressionfunction.cpp
```diff
@@ -5579,6 +5592,7 @@ const QList<QgsExpressionFunction *> &QgsExpression::Functions()
         << new QgsStaticExpressionFunction( QStringLiteral( "hamming_distance" ), QgsExpressionFunction::ParameterList() << QgsExpressionFunction::Parameter( QStringLiteral( "string1" ) ) << QgsExpressionFunction::Parameter( QStringLiteral( "string2" ) ), fcnHamming, QStringLiteral( "Fuzzy Matching" ) )
         << new QgsStaticExpressionFunction( QStringLiteral( "soundex" ), QgsExpressionFunction::ParameterList() << QgsExpressionFunction::Parameter( QStringLiteral( "string" ) ), fcnSoundex, QStringLiteral( "Fuzzy Matching" ) )
         << new QgsStaticExpressionFunction( QStringLiteral( "char" ), QgsExpressionFunction::ParameterList() << QgsExpressionFunction::Parameter( QStringLiteral( "code" ) ), fcnChar, QStringLiteral( "String" ) )
+        << new QgsStaticExpressionFunction( QStringLiteral( "ascii" ), QgsExpressionFunction::ParameterList() << QgsExpressionFunction::Parameter( QStringLiteral( "code" ) ), fcnAscii, QStringLiteral( "String" ) )
```
nyalldawson, line 5595:
```suggestion
        << new QgsStaticExpressionFunction( QStringLiteral( "ascii" ), QgsExpressionFunction::ParameterList() << QgsExpressionFunction::Parameter( QStringLiteral( "string" ) ), fcnAscii, QStringLiteral( "String" ) )
```

nyalldawson: This needs to match the argument name specified in the json help file

nyalldawson (commented):
Looking good! This is a handy addition to have.

nyalldawson:
Nice work @JanCaha ! Better send in a PR updating https://github.com/qgis/QGIS/blob/master/resources/data/contributors.json now :wink:

havatv:
The name of this expression and its description do not match.

resources/function_help/json/ascii:
`"description": "Returns the unicode code associated with the first character of a string."`

If the behaviour of this expression is identical to the PostgreSQL ascii string function, the same definition should be used, to get the link to "ASCII" (PG 10.0 documentation):
`ASCII code of the first character of the argument. For UTF8 returns the Unicode code point of the character. For other multibyte encodings, the argument must be an ASCII character.`
If the behaviour is not identical to the PG one, the name "ascii" is misleading, and should be changed (to "unicode"?).

PR #35842, opened 2020-04-18
Add more 3D-related classes to PyQGIS

wonder-sk:
Add more Python wrappers for QGIS 3D classes:

- Qgs3DMapSettings
- QgsCameraPose
- QgsLayoutItem3DMap
- QgsPointLightSettings


nyalldawson (approved):

3nids:
@wonder-sk it would be cool if you could add "3d" to the modules in https://github.com/qgis/QGIS/blob/master/scripts/sip_include.sh#L37

This would correctly report any missing bindings.

I gave it a run but many files are added. 
They would require a `#define SIP_NO_FILE` if we don't want them.

PR #35843, opened 2020-04-18
[browser] React when a custom data item provider is added/removed

wonder-sk:
Until now, if a newly activated plugin had a custom data item provider that added a root item to the browser model, the new root data item would not get added and a restart of QGIS was necessary.

### DIFF
--- a/python/core/auto_generated/qgsbrowsermodel.sip.in
+++ b/python/core/auto_generated/qgsbrowsermodel.sip.in
@@ -9,6 +9,7 @@
 
 
 
+
 class QgsBrowserModel : QAbstractItemModel
 {
 %Docstring
--- a/python/core/auto_generated/qgsdataitemproviderregistry.sip.in
+++ b/python/core/auto_generated/qgsdataitemproviderregistry.sip.in
@@ -11,7 +11,7 @@
 
 
 
-class QgsDataItemProviderRegistry
+class QgsDataItemProviderRegistry : QObject
 {
 %Docstring
 This class keeps a list of data item providers that may add items to the browser tree.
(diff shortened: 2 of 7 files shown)


nyalldawson:
Does this crash if you remove a provider while the browser is populating? I've tried a similar thing in the past (based on @nathanw2's work), yet it was always crashy...

wonder-sk:
> Does this crash if you remove a provider while the browser is populating?

Hmm... I haven't tried such a scenario. What was your use case for that? Here what I am trying to do is to just avoid the situation that a freshly installed plugin needs restart of QGIS, so that users do not get confused...

nyalldawson:
>  What was your use case for that?

Basically just reloading a plugin using the plugin reloader.

wonder-sk:
> > What was your use case for that?
> 
> Basically just reloading a plugin using the plugin reloader.

Just tried reloading my test plugin with the plugin reloader and everything went fine.

On a second thought, I don't see how plugin reloader would trigger removal of a provider while browser was being initialized - with plugin reloader, the provider's root item would first get removed, then provider gets deleted, and finally a new instance of the provider gets added and a new root item gets added...

nyalldawson:
> On a second thought, I don't see how plugin reloader would trigger removal of a provider while browser was being initialized - with plugin reloader, the provider's root item would first get removed, then provider gets deleted, and finally a new instance of the provider gets added and a new root item gets added...

I **think** I found that it was caused by this sequence:

- provider's root item starts populating children in a background thread
- provider's root item would gets removed while this is happening
- plugin gets reloaded, causing provider to be deleted while background population is occuring

(i think -- it was a while back now). But in any case, it's rather a corner case and I don't think should block this.

nyalldawson (approved):

PR #35853, opened 2020-04-19
Make autocomplete in expression builder case insensitive

uclaros:
## Description
This will make me a lot happier working with both oracle all caps tables and postgres all lower-case!
I believe the initial intention was for it to be case insensitive and this line is a bug, judging by the code in `QgsCodeEditorExpression`.


nyalldawson (approved):

nyalldawson:
I can't see any reason we'd want to restrict auto complete to case sensitive matches -- every IDE i've tested is insensitive.

PR #35856, opened 2020-04-19
Vector tile layer - part 6 (writer)

wonder-sk:

Initial work on vector tile writer. Currently supporting output to a directory based on XYZ template, using Mapbox vector tiles encoding. So far no user-facing way how to use this (only Python API).

New classes:
- QgsVectorTileMVTEncoder - low-level class that operates on a single tile, converts vector features to raw tile byte array, for internal use
- QgsVectorTileWriter - higher level class that manages generation of multiple tiles, for use by clients
- QgsVectorTileMVTUtils - assorted helper functions for MVT encoding/decoding

Sample code to generate vector tiles from a single layer:

```
uri = QgsDataSourceUri()
uri.setParam("type", "xyz")
uri.setParam("url", "file:///tmp/test/{z}/{x}-{y}.pbf")
lyr = QgsVectorTileWriter.Layer(iface.activeLayer())
writer = QgsVectorTileWriter()
writer.setDestinationUri(bytes(uri.encodedUri()).decode('utf-8'))
writer.setLayers([lyr])
writer.setMinZoom(0)
writer.setMaxZoom(3)
# writer.setExtent(QgsRectangle(...))
writer.writeTiles()
```

## Thanks

### DIFF
--- a/src/core/vectortile/qgsvectortilemvtencoder.cpp
+++ b/src/core/vectortile/qgsvectortilemvtencoder.cpp
@@ -1,0 +1,357 @@
+/***************************************************************************
+  qgsvectortilemvtencoder.cpp
+  --------------------------------------
+  Date                 : April 2020
+  Copyright            : (C) 2020 by Martin Dobias
+  Email                : wonder dot sk at gmail dot com
+ ***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+#include "qgsvectortilemvtencoder.h"
+
+#include "qgslinestring.h"
+#include "qgslogger.h"
+#include "qgsmultilinestring.h"
+#include "qgsmultipoint.h"
+#include "qgsmultipolygon.h"
+#include "qgspolygon.h"
+#include "qgsvectorlayer.h"
+#include "qgsvectortilemvtutils.h"
+
+
+//! Helper class for writing of geometry commands
+struct MVTGeometryWriter
+{
+  vector_tile::Tile_Feature *feature = nullptr;
+  int resolution;
+  double tileXMin, tileYMax, tileDX, tileDY;
+  QPoint cursor;
+
+  MVTGeometryWriter( vector_tile::Tile_Feature *f, int res, const QgsRectangle &tileExtent )
+    : feature( f )
+    , resolution( res )
+    , tileXMin( tileExtent.xMinimum() )
+    , tileYMax( tileExtent.yMaximum() )
+    , tileDX( tileExtent.width() )
+    , tileDY( tileExtent.height() )
+  {
+  }
+
+  void addMoveTo( int count )
+  {
+    feature->add_geometry( 1 | ( count << 3 ) );
+  }
+  void addLineTo( int count )
+  {
+    feature->add_geometry( 2 | ( count << 3 ) );
+  }
+  void addClosePath()
+  {
+    feature->add_geometry( 7 | ( 1 << 3 ) );
+  }
+
+  void addPoint( const QgsPoint &pt )
+  {
+    addPoint( mapToTileCoordinates( pt ) );
+  }
+
+  void addPoint( const QPoint &pt )
+  {
+    qint32 vx = pt.x() - cursor.x();
+    qint32 vy = pt.y() - cursor.y();
+
+    feature->add_geometry( ( vx << 1 ) ^ ( vx >> 31 ) );
+    feature->add_geometry( ( vy << 1 ) ^ ( vy >> 31 ) );
+
+    cursor = pt;
+  }
+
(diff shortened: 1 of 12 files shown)


nyalldawson:
Just for reference: what's the plan for these classes? Will creation be exposed as a new processing algorithm?

src/core/vectortile/qgsvectortilemvtencoder.cpp
```diff
@@ -0,0 +1,357 @@
...
+
+  QPoint mapToTileCoordinates( const QgsPoint &p )
+  {
+    return QPoint( static_cast<int>( round( ( p.x() - tileXMin ) * resolution / tileDX ) ),
+                   static_cast<int>( round( ( tileYMax - p.y() ) * resolution / tileDY ) ) );
+  }
+};
+
+
+static QVector<QgsPoint> pointsFromLineString( const QgsLineString *ls )
```
nyalldawson, line 83: This is rather inefficient -- you're better accessing QgsLineString::xData and yData direct and iterating over that, building the encoded line string directly during iteration

src/core/vectortile/qgsvectortilemvtencoder.cpp
```diff
@@ -0,0 +1,357 @@
...
+
+
+QgsVectorTileMVTEncoder::QgsVectorTileMVTEncoder( QgsTileXYZ tileID )
+  : mTileID( tileID )
+{
+  QgsTileMatrix tm = QgsTileMatrix::fromWebMercator( mTileID.zoomLevel() );
+  mTileExtent = tm.tileExtent( mTileID );
+}
+
+void QgsVectorTileMVTEncoder::addLayer( QgsVectorLayer *layer, QString filterExpression, QString layerName )
```
nyalldawson, line 152: I'd add a QgsFeedback argument here to support cancelation/progress reports

src/core/vectortile/qgsvectortilemvtencoder.cpp
```diff
@@ -0,0 +1,357 @@
...
+      valueIndex = tileLayer->values_size() - 1;
+      mKnownValues[v] = valueIndex;
+
+      if ( v.type() == QVariant::Double )
+        value->set_double_value( v.toDouble() );
+      else if ( v.type() == QVariant::Int )
+        value->set_int_value( v.toInt() );
+      else if ( v.type() == QVariant::Bool )
+        value->set_bool_value( v.toBool() );
+      else
```
nyalldawson, line 264: Are other types allowed in the spec? dates/times?

src/core/vectortile/qgsvectortilewriter.cpp
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  QgsDataSourceUri dsUri;
+  dsUri.setEncodedUri( mDestinationUri );
+
+  QString sourceType = dsUri.param( QStringLiteral( "type" ) );
+  QString sourcePath = dsUri.param( QStringLiteral( "url" ) );
+  if ( sourceType != QStringLiteral( "xyz" ) )
+  {
+    mErrorMessage = "Unsupported source type for writing: " + sourceType;
```
nyalldawson, line 59: translate?

src/core/vectortile/qgsvectortilewriter.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include <QtDebug>
+#include "qgsvectorlayer.h"
+
+QgsVectorTileWriter::QgsVectorTileWriter()
+{
+  mExtent = QgsTileMatrix::fromWebMercator( 0 ).tileExtent( QgsTileXYZ( 0, 0, 0 ) );
+}
+
+
+bool QgsVectorTileWriter::writeTiles()
```
nyalldawson, line 39: Maybe add QgsFeedback argument here also for progress/cancelation?

src/core/vectortile/qgsvectortilemvtencoder.cpp
wonder-sk: Not really, just strings, floats/doubles, ints, bools

https://github.com/mapbox/vector-tile-spec/blob/master/2.1/vector_tile.proto#L17

wonder-sk:
> Just for reference: what's the plan for these classes? Will creation be exposed as a new processing algorithm?

Yeah the plan is to have a new processing algorithm, maybe two - one to write files/directories with {x},{y},{z} template, another one to write MBTiles.

src/core/vectortile/qgsvectortilemvtencoder.cpp
nyalldawson: I'd go one step further here, and not return QVector<QgsPoint> here but instead directly build the encoded line string. otherwise you're constructing a QVector of QgsPoints which aren't really needed for every linestring, which will definitely impact performance.

wonder-sk: Okay, that was just lazy myself - fixed in d025079 now :-)

There is still one intermediate vector of points created (instead of two) while encoding - the reason is that we need to do de-duplication of points in the transformed integer coordinates (spec does not allow duplicate points) and the point count needs to be written before encoding points. But I am not really worried about performance of the encoder (at least not at this stage :-) )

nyalldawson:
Lgtm!

PR #35869, opened 2020-04-20
fix deprecation warning in qgis.utils (fix #32786)

### LINKED ISSUE #32786: Deprecated python module usage: imp module is deprecated in favour of importlib
```
/python/qgis/utils.py:737: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
```

alexbruy:
## Description
The`imp.new_module()` call was deprecated in Python 3.4, as result a deprecation warning shown in `qgis.utils` module. Fixes #32786.

### DIFF
--- a/python/utils.py
+++ b/python/utils.py
@@ -545,9 +545,12 @@ def reloadProjectMacros():
         return
 
     # create a new empty python module
-    import imp
-
-    mod = imp.new_module("proj_macros_mod")
+    if sys.version_info >= (3, 4):
+        import importlib
+        mod = importlib.util.module_from_spec(importlib.machinery.ModuleSpec("proj_macros_mod", None))
+    else:
+        import imp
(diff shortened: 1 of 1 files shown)


python/utils.py
```diff
@@ -545,9 +545,12 @@ def reloadProjectMacros():
         return
 
     # create a new empty python module
-    import imp
-
-    mod = imp.new_module("proj_macros_mod")
+    if sys.version_info >= (3, 4):
+        import importlib
```
m-kuhn, line 549: Could we use that unconditionally? It should be around since 3.1

alexbruy: Sure, I just wanted to be safe in case someone still using old Python.

agiudiceandrea:
It seems the warning was not fixed: https://github.com/qgis/QGIS/issues/32786#issuecomment-618646622

Issue #35872, opened 2020-04-20
Embedded sub-group checked state aren't saved in project


SebastienPeillet:
**Describe the bug**

Relative to #33099,  that fixed embedded layer checked state. I didn't see at the moment that sub-groups inside an embedded group were not handled too.

The embedded sub-group checked state is determined by the embedded project and not by the current project. So if the embedded project has an sub-group unchecked, there is no way to save it as checked in the current project.

**How to Reproduce**

1. Open a new project, open several layer to reproduce a project like following 

![image](https://user-images.githubusercontent.com/18545440/79760866-f6247d00-8320-11ea-8ec3-2011ceab4df4.png)

2. Save the project
3. Open a new project and use "Embedded layers and groups..." functionnality to import all the layers and groups from the first project.

4. Unchek the sub-group and save the project
5. Reload the project, sub-group will be checked 

**QGIS and OS versions**

QGIS version | 3.13.0-Master | QGIS code revision | 4c22cdc6a6
-- | -- | -- | --
Compiled against Qt | 5.12.2 | Running against Qt | 5.12.2
Compiled against GDAL/OGR | 2.4.0 | Running against GDAL/OGR | 2.4.0
Compiled against GEOS | 3.7.1-CAPI-1.11.1 | Running against GEOS | 3.7.1-CAPI-1.11.1 27a5e771
Compiled against SQLite | 3.27.2 | Running against SQLite | 3.27.2
PostgreSQL Client Version | 11.5 (Ubuntu 11.5-0ubuntu0.19.04.1) | SpatiaLite Version | 5.0.0-beta0
QWT Version | 6.1.4 | QScintilla2 Version | 2.10.4
Compiled against PROJ | 5.2.0 | Running against PROJ | Rel. 5.2.0, September 15th, 2018
OS Version | Ubuntu 19.04 | This copy of QGIS writes debugging output.
Active python plugins | processing; db_manager

Pedro-Murteira:
Can still replicate this issue with QGIS 3.16.14 and 3.22.1.

SebastienPeillet:
@Pedro-Murteira indeed. I re-opened the PR and hope it will be merged.

lucamanga:
Confirming this in latest version of QGIS.

PR #35874, opened 2020-04-20
Fix embedded sub-group

SebastienPeillet:
## Description

See #35872 

The embedded sub-group checked state is determined by the embedded project and not by the current project. So if the embedded project has an sub-group unchecked, there is no way to save it as checked in the current project.

As we handle layers visibility with the `embedded-invisible-layers` tag, I added a `embedded-unchecked-groups` to handle group checked state.

```xml
  <layer-tree-group>
    <customproperties/>
    <layer-tree-group name="group1" expanded="1" checked="Qt::Checked">
      <customproperties>
        <property key="embedded" value="1"/>
        <property key="embedded-invisible-layers"/>
        <property key="embedded-unchecked-groups" value="sub-group"/>
        <property key="embedded_project" value="./projet.qgs"/>
      </customproperties>
    </layer-tree-group>
```

I added elements in testEmbeddedGroup ([testqgslayertree.cpp](https://github.com/SebastienPeillet/QGIS/blob/9020f0dd97fcf2ef7c8e0f7ac5b5c0fd978a6067b/tests/src/core/testqgslayertree.cpp#L538)) to verify the sub-group checked state works.

### DIFF
--- a/src/core/layertree/qgslayertreeutils.cpp
+++ b/src/core/layertree/qgslayertreeutils.cpp
@@ -386,7 +386,7 @@ QStringList QgsLayerTreeUtils::invisibleLayerList( QgsLayerTreeNode *node )
   return list;
 }
 
-QStringList QgsLayerTreeUtils::uncheckedGroupList( QgsLayerTreeNode *node )
+QStringList QgsLayerTreeUtils::uncheckedGroupList( QgsLayerTreeNode *node, int index )
 {
   QStringList list;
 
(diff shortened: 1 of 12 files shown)


src/core/layertree/qgslayertreeutils.cpp
```diff
@@ -396,6 +416,7 @@ void QgsLayerTreeUtils::replaceChildrenOfEmbeddedGroups( QgsLayerTreeGroup *grou
       if ( child->customProperty( QStringLiteral( "embedded" ) ).toInt() )
       {
         child->setCustomProperty( QStringLiteral( "embedded-invisible-layers" ), invisibleLayerList( child ) );
+        child->setCustomProperty( QStringLiteral( "embedded-unchecked-groups" ), uncheckedGroupList( child ) );
```
nyalldawson, line 419: This doesn't look particularly stable -- two equally named groups will mess up the result

SebastienPeillet: Thanks for the review !

Indeed, I noticed that but I wouldn't add an id to group because it seems to me that this subject has already been talked a while ago, and furthermore switching for id group will need some refactoring I guess. But I agree this is not satisfying enough.

I also though about using the node index in each embedded group (I pushed the changes). This works too, and it accepts groups with the same name, but in case of changes in the project that is embedded (a layer/group added/removed), index may change and the checked state will be lost.

So after all group id seems to be needed. I change the PR title with WIP and I will try test to add group id

SebastienPeillet: I fixed it, adding a group's id in the tree layer !

gotjoshua:
Whats going on with this PR? 

It would be really great if the project that embeds a group full of subgroups, can decide/override which sub-groups and sub items to check...

SebastienPeillet:
Some conflict appears with new commits in master. I fixed it, it should be possible to merge again.

gotjoshua:
I think this should be merged... @SebastienPeillet fixed the confilicts and all checks have passed!

gotjoshua:
Again, I'd love to prevent this from goin stale and going unmerged... @nyalldawson  any news?

gotjoshua:
Bringing back the stale once again...
stayin alive, stayin alive, woo whooo hoo hoo...

gotjoshua:
stayin aliiiiiiive!

nyalldawson:

PR #35875, opened 2020-04-20
[feature] Curved geometries support for trace tool

olivierdalang:
This PR adds curved geometries support for the trace tool. Currently, when tracing existing curved geometries, the results are segmentized. This PR fixes this by "de-segmentizing" traced curves when working on a layer that supports geometries.

pr

The de-segmentation is made by `OGRGeometry->getCurveGeometry()`.

The whole tracing logic is left unchanged, so it still makes use of segmentized geometries, as currently there's no support for true curves intersection in the libraries we use. This means there's a small approximation if snapping on curved intersections, but this was the case before already.

Note this other PR https://github.com/qgis/QGIS/pull/35877 which fixes multi-geometries not being correctly cast to linear, preventing the trace tool from working on MultiCurves (with or without this PR).

As commented in the code, I had to declare  `int OGRWktOptions::getDefaultPrecision(void)` and `bool OGRWktOptions::getDefaultRound(void)` to fix linking errors... Not familiar with OGR at all, but it doesn't look exactly right to have this there... Any idea on how to better avoid this issue ?

### DIFF
--- a/src/gui/qgsmaptoolcapture.cpp
+++ b/src/gui/qgsmaptoolcapture.cpp
@@ -290,8 +290,9 @@ bool QgsMapToolCapture::tracingAddVertex( const QgsPointXY &point )
   }
 
   // If the layer supports curves, we de-approximate curves
+  QgsSettings settings;
   QgsVectorLayer *vlayer = qobject_cast<QgsVectorLayer *>( mCanvas->currentLayer() );
-  if ( vlayer->dataProvider()->capabilities().testFlag( QgsVectorDataProvider::Capability::CircularGeometries ) )
+  if ( vlayer->dataProvider()->capabilities().testFlag( QgsVectorDataProvider::Capability::CircularGeometries ) && settings.value( QStringLiteral( "/qgis/digitizing/convert_to_curve" ), false ).toBool() )
   {
     QgsGeometry linear = QgsGeometry( mCaptureCurve.segmentize() );
     QgsGeometry curved = linear.convertToCurves();
--- a/src/gui/qgsmaptoolcapture.cpp
+++ b/src/gui/qgsmaptoolcapture.cpp
@@ -13,8 +13,7 @@
  *                                                                         *
  ***************************************************************************/
 
-#include <ogr_api.h>
-#include <ogr_geometry.h>
+#include "ogr_srs_api.h"
 
 #include "qgsmaptoolcapture.h"
 #include "qgsexception.h"
@@ -26,6 +25,7 @@
 #include "qgsmapcanvas.h"
 #include "qgsmapcanvastracer.h"
 #include "qgsmapmouseevent.h"
+#include "qgsogrutils.h"
 #include "qgspolygon.h"
 #include "qgsrubberband.h"
 #include "qgssnapindicator.h"
@@ -296,34 +296,22 @@ bool QgsMapToolCapture::tracingAddVertex( const QgsPointXY &point )
   if ( vlayer->dataProvider()->capabilities().testFlag( QgsVectorDataProvider::Capability::CircularGeometries ) )
   {
     // We convert the capture curve to OGR geometry
-    OGRGeometry *ogrCaptureCurve;
-    OGRGeometryFactory::createFromWkt( mCaptureCurve.asWkt().toLocal8Bit().data(), nullptr, &ogrCaptureCurve );
+    QByteArray wkb = mCaptureCurve.asWkb();
+    OGRGeometryH ogrCaptureCurve;
+    OGR_G_CreateFromWkb( ( unsigned char *) wkb.constData(), nullptr, &ogrCaptureCurve, wkb.size() );
 
     // We de-approximate the curves using OGR (the geometry must be forced to linear first, as a mCaptureCurve is a CompoundCurve)
-    OGRGeometry *ogrLinestring = ogrCaptureCurve->getLinearGeometry(); // it must be forced to linear, as captuecurve is a compound curve
-    OGRGeometry *ogrRecurved = ogrLinestring->getCurveGeometry();
+    OGRGeometryH ogrLinestring = OGR_G_GetLinearGeometry( ogrCaptureCurve, 0, nullptr );
+    OGRGeometryH ogrRecurved = OGR_G_GetCurveGeometry( ogrLinestring, nullptr );
 
     // We save back to mCaptureCurve
-    mCaptureCurve.clear();
-    mCaptureCurve.fromWkt( QString::fromStdString( ogrRecurved->exportToWkt() ) );
+    mCaptureCurve = *qgsgeometry_cast<QgsCompoundCurve *>( QgsOgrUtils::ogrGeometryToQgsGeometry( ogrRecurved ).constGet() );
   }
 
   tracer->reportError( QgsTracer::ErrNone, true ); // clear messagebar if there was any error
   return true;
 }
 
-// Required to avoid this (took values from https://github.com/OSGeo/gdal/blob/master/gdal/ogr/ogrgeometry.cpp):
-// qgsmaptoolcapture.cpp.obj : error LNK2019: unresolved external symbol "private: static int __cdecl OGRWktOptions::getDefaultPrecision(void)" (?getDefaultPrecision@OGRWktOptions@@CAHXZ) referenced in function "public: __cdecl OGRWktOptions::OGRWktOptions(void)" (??0OGRWktOptions@@QEAA@XZ)
-// qgsmaptoolcapture.cpp.obj : error LNK2019: unresolved external symbol "private: static bool __cdecl OGRWktOptions::getDefaultRound(void)" (?getDefaultRound@OGRWktOptions@@CA_NXZ) referenced in function "public: __cdecl OGRWktOptions::OGRWktOptions(void)" (??0OGRWktOptions@@QEAA@XZ)
-// output\bin\qgis_gui.dll : fatal error LNK1120: 2 unresolved externals
-int OGRWktOptions::getDefaultPrecision( void )
-{
-  return 15;
-}
-bool OGRWktOptions::getDefaultRound( void )
-{
-  return true;
-}
 
 QgsRubberBand *QgsMapToolCapture::takeRubberBand()
 {
--- a/src/gui/qgsmaptoolcapture.cpp
+++ b/src/gui/qgsmaptoolcapture.cpp
@@ -13,8 +13,10 @@
  *                                                                         *
  ***************************************************************************/
 
-#include "qgsmaptoolcapture.h"
+#include <ogr_api.h>
+#include <ogr_geometry.h>
 
+#include "qgsmaptoolcapture.h"
 #include "qgsexception.h"
 #include "qgsfeatureiterator.h"
 #include "qgsgeometryvalidator.h"
@@ -289,7 +291,37 @@ bool QgsMapToolCapture::tracingAddVertex( const QgsPointXY &point )
     mSnappingMatches.append( QgsPointLocator::Match() );
   }
 
+  // If the layer supports curves, we de-approximate curves
+  QgsVectorLayer *vlayer = qobject_cast<QgsVectorLayer *>( mCanvas->currentLayer() );
+  if ( vlayer->dataProvider()->capabilities().testFlag( QgsVectorDataProvider::Capability::CircularGeometries ) )
+  {
+    // We convert the capture curve to OGR geometry
+    OGRGeometry *ogrCaptureCurve;
+    OGRGeometryFactory::createFromWkt( mCaptureCurve.asWkt().toLocal8Bit().data(), nullptr, &ogrCaptureCurve );
+
+    // We de-approximate the curves using OGR (the geometry must be forced to linear first, as a mCaptureCurve is a CompoundCurve)
+    OGRGeometry *ogrLinestring = ogrCaptureCurve->getLinearGeometry(); // it must be forced to linear, as captuecurve is a compound curve
+    OGRGeometry *ogrRecurved = ogrLinestring->getCurveGeometry();
+
+    // We save back to mCaptureCurve
+    mCaptureCurve.clear();
+    mCaptureCurve.fromWkt( QString::fromStdString( ogrRecurved->exportToWkt() ) );
+  }
+
   tracer->reportError( QgsTracer::ErrNone, true ); // clear messagebar if there was any error
+  return true;
+}
+
+// Required to avoid this (took values from https://github.com/OSGeo/gdal/blob/master/gdal/ogr/ogrgeometry.cpp):
+// qgsmaptoolcapture.cpp.obj : error LNK2019: unresolved external symbol "private: static int __cdecl OGRWktOptions::getDefaultPrecision(void)" (?getDefaultPrecision@OGRWktOptions@@CAHXZ) referenced in function "public: __cdecl OGRWktOptions::OGRWktOptions(void)" (??0OGRWktOptions@@QEAA@XZ)
+// qgsmaptoolcapture.cpp.obj : error LNK2019: unresolved external symbol "private: static bool __cdecl OGRWktOptions::getDefaultRound(void)" (?getDefaultRound@OGRWktOptions@@CA_NXZ) referenced in function "public: __cdecl OGRWktOptions::OGRWktOptions(void)" (??0OGRWktOptions@@QEAA@XZ)
+// output\bin\qgis_gui.dll : fatal error LNK1120: 2 unresolved externals
+int OGRWktOptions::getDefaultPrecision( void )
+{
+  return 15;
+}
+bool OGRWktOptions::getDefaultRound( void )
+{
   return true;
 }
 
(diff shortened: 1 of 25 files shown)


src/gui/qgsmaptoolcapture.cpp
```diff
@@ -289,10 +291,40 @@ bool QgsMapToolCapture::tracingAddVertex( const QgsPointXY &point )
...
+    OGRGeometry *ogrCaptureCurve;
+    OGRGeometryFactory::createFromWkt( mCaptureCurve.asWkt().toLocal8Bit().data(), nullptr, &ogrCaptureCurve );
+
+    // We de-approximate the curves using OGR (the geometry must be forced to linear first, as a mCaptureCurve is a CompoundCurve)
+    OGRGeometry *ogrLinestring = ogrCaptureCurve->getLinearGeometry(); // it must be forced to linear, as captuecurve is a compound curve
+    OGRGeometry *ogrRecurved = ogrLinestring->getCurveGeometry();
+
+    // We save back to mCaptureCurve
+    mCaptureCurve.clear();
+    mCaptureCurve.fromWkt( QString::fromStdString( ogrRecurved->exportToWkt() ) );
```
m-kuhn, line 308: I wonder if we should encapsulate this code in qgsgeometryutils.h so it can be used from everywhere. And potentially replaced with a more stable version in one single place - or removed in one single place once things have been "properly fixed" in underlying libraries.

andreasneumann:
Cool!

src/gui/qgsmaptoolcapture.cpp
olivierdalang: Yes I wasn't really sure where to implement this... There's also `QgsAbstractGeometry.segmentize()`, so it would be make sense to add`QgsAbstractGeometry::desegmentize()` !

m-kuhn: Good point.
There's quite some overhead to it though at the moment. So I wonder to what extent we want to bring it "that deep into core" in its current shape. Very open to others opinions !

src/gui/qgsmaptoolcapture.cpp
```diff
@@ -289,10 +291,40 @@ bool QgsMapToolCapture::tracingAddVertex( const QgsPointXY &point )
     mSnappingMatches.append( QgsPointLocator::Match() );
   }
 
+  // If the layer supports curves, we de-approximate curves
+  QgsVectorLayer *vlayer = qobject_cast<QgsVectorLayer *>( mCanvas->currentLayer() );
+  if ( vlayer->dataProvider()->capabilities().testFlag( QgsVectorDataProvider::Capability::CircularGeometries ) )
+  {
+    // We convert the capture curve to OGR geometry
+    OGRGeometry *ogrCaptureCurve;
+    OGRGeometryFactory::createFromWkt( mCaptureCurve.asWkt().toLocal8Bit().data(), nullptr, &ogrCaptureCurve );
```
nyalldawson, line 300: we shouldn't use wkt as the inbetween here -- wkb should be used instead. And there's already QgsOgrUtils::ogrGeometryToQgsGeometry which is an optimised method to convert back from OGR to QGIS.

src/gui/qgsmaptoolcapture.cpp
```diff
@@ -289,10 +291,40 @@ bool QgsMapToolCapture::tracingAddVertex( const QgsPointXY &point )
...
+
+    // We de-approximate the curves using OGR (the geometry must be forced to linear first, as a mCaptureCurve is a CompoundCurve)
+    OGRGeometry *ogrLinestring = ogrCaptureCurve->getLinearGeometry(); // it must be forced to linear, as captuecurve is a compound curve
```
nyalldawson, line 303: You can't use the gdal c++ api -- you need to use the c api equivalents.

src/gui/qgsmaptoolcapture.cpp
```diff
@@ -289,10 +291,40 @@ bool QgsMapToolCapture::tracingAddVertex( const QgsPointXY &point )
...
+    OGRGeometry *ogrRecurved = ogrLinestring->getCurveGeometry();
```
nyalldawson, line 304: @rouault is OGR_G_GetCurveGeometry the same algorithm as Postgis' ST_LineToCurve ? Or is it a different algorithm? I've previously tried porting Postgis' ST_LineToCurve to a native QGIS function (See https://github.com/nyalldawson/QGIS/commit/ac8061ef6e0556fb3befbfca95d13f15c4d2e2f2), but found a few issues with the results from that function in Postgis (and never got the time to fix them unfortunately)

uclaros:
@olivierdalang Nice! On your *after* video you start tracing directly at the curve's endpoint and not at a previous straight linestring. Is the transition from straight to curved handled? Also, how is starting the trace with a *segment* snap on the curve handled?

src/gui/qgsmaptoolcapture.cpp
olivierdalang: Ok (done)

src/gui/qgsmaptoolcapture.cpp
olivierdalang: Ok (done) but just out of interest, why ? (to me the C++ api definitely looks easier to read)

src/gui/qgsmaptoolcapture.cpp
olivierdalang: I noticed some glitches on certain geometries : in certain cases, some arcs get completely removed... not sure why... is this the type or issues you encountered ?

As said above (https://github.com/qgis/QGIS/pull/35875#discussion_r411488123) having this available in QGIS would definitely make sense, it would be awesome if we could solve the issues you were mentionning.

olivierdalang:
@uclaros the tracing code wasn't changed, meaning that everything that worked before still works (transition from curves to lines), and everything that didn't still won't (tracing from/to the arc itself, unless there is an intersection). The de-approximation is done afterwards.

It would be nice to allow tracing from/to the arc though, will see if this can be added.

olivierdalang:
I was mistaken, actually tracing from/to arcs worked already (and still works) more or less... but not reliably. Here it seems to work only when near the origin, so there may be something about rounding.

I saw an `epsilon=1e-6` in the snapping code, tweaking it seems to affect the issue. I have to investigate a bit more...

lbartoletti:
>  Here it seems to work only when near the origin, so there may be something about rounding.

I also noticed this kind of problems when I writed some tests on snapping.

> I saw an `epsilon=1e-6` in the snapping code, tweaking it seems to affect the issue.

`epsilon` is "everywhere" and some times with different values.
 
> It would be good to fix that too, but IMO it's for another PR, as it is a separate (and pre-existing) issue.

+1, let me know if you want tests/help.

src/gui/qgsmaptoolcapture.cpp
lbartoletti: C API is stable (same for GEOS for example: provides long-term ABI stability)

src/gui/qgsmaptoolcapture.cpp
rouault: > OGR_G_GetCurveGeometry the same algorithm as Postgis' ST_LineToCurve ?

Much likely not. My own creation. Not something fully bullet proof. Definitely not as I discussed with @m-kuhn  some time ago
```
Some years ago I did for fun a hack. When using OGR geometry methods that ultimately rely on GEOS methods, I first linearize geometries, but hide in the LSB of some X&Y coordinates
a hidden information (see the ADD_INTERMEDIATE_POINT=STEALTH
option of https://gdal.org/api/ogrgeometry_cpp.html#_CPPv4N18OGRGeometryFactory17curveToLineStringEdddddddddidPPCKc ),
so that when GEOS is done, I can re-use to, possibly, rebuild circular portions that
weren't affected by the geometry operation. This is of course a super gross hack, not robust at all...

Little demo though on a simple case with 2 circles overlapping:

>>> from osgeo import ogr
>>> g1 = ogr.CreateGeometryFromWkt('CurvePolygon(CircularString(0 0,1 1,0 0))')
>>> g2 = ogr.CreateGeometryFromWkt('CurvePolygon(CircularString(0.5 0,1.5 1,0.5 0))')
>>> print(g1.Union(g2))
CURVEPOLYGON (COMPOUNDCURVE ((0.749706234848466 -0.16111841890272,0.721987566409211 -0.171357967376366),CIRCULARSTRING (0.721987566409211 -0.171357967376366,-0.206869769243333 0.481693462099955,0.733542564392451 1.16742630351001),(0.733542564392451 1.16742630351001,0.749699897508666 1.16113860048096,0.754972022079414 1.16329577869618),CIRCULARSTRING (0.754972022079414 1.16329577869618,1.70581670320598 0.542694044965174,0.789633573195289 -0.175089598848494),(0.789633573195289 -0.175089598848494,0.749706234848466 -0.16111841890272)))

But I wouldn't recommend relying on that. Due to the linearization, and how reconstruction
is done, there are certainly inaccuracies that would make it unfit when topological
correctness is needed. Like in the above example where there are 2 small segments, that
shouldn't be there: the union should be just 2 circular strings.
```
Code at https://github.com/OSGeo/gdal/blob/master/gdal/ogr/ogrgeometryfactory.cpp#L5858 , and core of it at https://github.com/OSGeo/gdal/blob/master/gdal/ogr/ogrgeometryfactory.cpp#L5289

olivierdalang:
@nyalldawson I tested with your implementation, and am getting the same glitches than with @rouault 's from GDAL. Some arcs are lost. But most of the times, it works as expected.

image

It would be awesome to have that branch merged in QGIS though !! Having the convertToCurve method and the processing algorithm is a big deal in terms of QGIS's better support of curves.

@nyalldawson You were mentioning issues with the results, is this what you were referring to ? Do you remember if there was something particularly nasty about that issue, or should it be fixable ?

olivierdalang:
@lbartoletti I opened https://github.com/qgis/QGIS/issues/36168 about the tracing, feel free to add more info there if you encounter the same issue

saberraz:
Nice feature!

I probably can answer my own question by compiling your branch...but off-hand, is this also available for the advanced digitizing tools? The current tracing allows users to follow existing shapes for some of the advanced digitizing tools (e.g. splitting a polygon along the existing line by tracing).

olivierdalang:
@saberraz This PR is theoretically available to all subclasses of QgsMapToolCapture, which is most if not all advanced digitizing tools (as it just converts the capture curve in circular geometries). But it seems most of the advanced digitizing tools do no support curved capture, as they iterate on points, not expecting curves (which are just 3 points), so that this PR actually breaks them... So the answer is no, support for advanced tools would have to be done separately in subsequent PRs.

I had not thought of that ! Thanks for raising it.

To deal with this, I'd suggest adding a flag on QgsMapToolCapture subclasses telling whether the tool supports circular geometries, so that we only convert to curves when supported both by the layer and by the tool.

m-kuhn:
Something else to consider, as soon as geometries are sent through geos for splitting, merging etc curves will be gone anyway.

This will need to be fixed as well for full support for all these tools. We are currently looking into what's required there, but that's still in an early stage.

olivierdalang:
Here's the progress on this :

- [x] rebase @nyalldawson's work and use it instead of OGR's, meaning we have native line to curve support in QGIS (yay!!) 
- [x] add an user setting to enable this feature (defaults to false, if we can fix all issues below we may want to default to true ?)
- [x] add a capability flag to QgsMapToolCapture, allowing each subtool to tell if it supports curved input (so far, only `QgsMapToolDigitizeFeature` and `QgsMapToolAddPart` do,), which fixes issue raised by @saberraz 
- [ ] see if we can adapt QgsMapToolAddRing too (for now, the result is segmented)
- [ ] isolate and fix the disappearing arcs issue
- [ ] see if we can relax convertToCurves's algorithm to allow some irregular points distribution (which happen if there are intersections)
- [ ] add tests
- [ ] squash commits (distinguishing Nyall's branch and this PR's feature, as they are different in scope)

olivierdalang:
@nyalldawson 
Looking into the `convertToCurves` logic, I think we should replace the concept of "regularly spaced points" with the concept of "maximum arc angle". This means that instead of testing for regular spacing, we would just ensure the arc angle doesn't exceed a certain limit (after which the feature is considered angular).

Rationale :
- an octogon, no matter how regular, isn't a circle (at least not always)
- spatial operations may alter spacing, so that regular spacing isn't a good indicator for circularness (see the issue around an intersection below)
- as a parameter, it's intuitive, as it's really the reciprocal of parameters such as the `segments` parameter of the buffer algorithm (arguably, we could even expose it in the same way in the processing algorithm, it's probably easier to understand for users than maximum arc angle)

Does this make sense ? Or is there something more to regular spacing I didn't think of ?

image

andreasneumann:
> I think we should replace the concept of "regularly spaced points" with the concept of "maximum arc angle".

I think you should also get @strk into the discussion here. He worked on this topic, trying to align behaviour of Postgis/GEOS/QGIS

olivierdalang:
Actually there is some logic in the code to deal with the octogon issue (`minQuadEdges` in `QgsInternalGeometryEngine::lineToCurve`), but it's hardcoded to 2.

So, to stay aligned with postgis, we could maybe stick with the initial version, and add another optional parameter to set `minQuadEdges` ? (which I'd suggest to default to something like twice the default of `QgsGeometry::segmentize()`)

nyalldawson:
@olivierdalang 

I remember now what I'd decided when I was originally working on this. 

One issue is that the algorithm only moves forward through the vertices. So if you get some vertices like this:

image

Then the postgis approach is to iterate through the vertices a->e. It sees that a-b-c is a potential curve candidate, so starts a curve using those initial three points. Then it continues on the d and e, sees that they are within tolerance of the a-b-c based curve, so considers them redundant (brown line above).

A better approach needs to be able to backtrack. It would start by checking a-b-c (then d and e, as above) and confirming that it's a curve candidate. It would then proceed to consider b-c-d alone, and find that a curve based on these three points (ignoring a) results in a better fit for b-c-d-e, with smaller squared distance between the points and the curve. So it would discard the previous curve candidate based on a-b-c, and instead use a straight segment for a-b, followed by the b-c-d based curve through to e (green line above)

The curve fits should use a distance based measure when comparing the curve candidate against the original lines, so that the function can accept a "maximum deviation distance" which dictates how closely the curve needs to fit to the existing straight line segments.

olivierdalang:
@nyalldawson Ok I see what you mean. In the most common usage scenario, (where it's about converting a segmentized geometry to curve), I guess it shouldn't make a big difference, as alignement will be withing the tolerance either way.

What do you think, can we merge this as is, and improve the implementation in next versions ?
 IMO if this is how postgis does it, it should be good enough for QGIS too. 

Maybe worth to flag these feature as experimental though, so that we have a little more room for changing the API in case usage reveals some other required changes...

src/gui/qgsmaptoolcapture.h
```diff
@@ -52,10 +52,11 @@ class GUI_EXPORT QgsMapToolCapture : public QgsMapToolAdvancedDigitizing
       CapturePolygon  //!< Capture polygons
     };
 
+    //! Specific capbilities of the tool
```
m-kuhn, line 55:
```suggestion
    //! Specific capabilities of the tool
```

src/gui/qgsmaptoolcapture.cpp
```diff
@@ -24,6 +23,7 @@
 #include "qgsmapcanvas.h"
 #include "qgsmapcanvastracer.h"
 #include "qgsmapmouseevent.h"
+#include "qgsogrutils.h"
```
m-kuhn, line 26:
```suggestion
```

No longer needed

m-kuhn:
Do I get the status of this PR right?

 - It implements the (ported from postgis) approach by @nyalldawson 
 - It ships a processing algorithm
 - It has tracing tool support, but opt in through an experimental option
 - This means unchanged behavior if the user doesn't opt in or trigger new actions, so no chance for data corruption
 - The algorithm has potential for improvements still to get even better curves. Partially because a "real solution" will need to be done on a lower level (geometry engine/geos) partially because some cases could be covered better

If possible, I would like to merge it, unless there is a big objection somewhere.

andreasneumann:
Just a quick motivator: it would be really nice to have some circular arc improvements in QGIS 3.14 Pi ;-) good luck on this work!

olivierdalang:
> Do I get the status of this PR right?
> 
> * It implements the (ported from postgis) approach by @nyalldawson
> * It ships a processing algorithm
> * It has tracing tool support, but opt in through an experimental option
> * This means unchanged behavior if the user doesn't opt in or trigger new actions, so no chance for data corruption
> * The algorithm has potential for improvements still to get even better curves. Partially because a "real solution" will need to be done on a lower level (geometry engine/geos) partially because some cases could be covered better
> 
> If possible, I would like to merge it, unless there is a big objection somewhere.

Yes that's it. The option isn't labeled as experimental though, should this be changed ?

And about the new API, should we add the note `The API is considered EXPERIMENTAL and can be changed without a notice` ?

m-kuhn:
> > 
> > If possible, I would like to merge it, unless there is a big objection somewhere.
> 
> Yes that's it. The option isn't labeled as experimental though, should this be changed ?
> 
> And about the new API, should we add the note `The API is considered EXPERIMENTAL and can be changed without a notice` ?

I would prefer to do it this way, yes !

m-kuhn (approved):

PR #35877, opened 2020-04-20
QgsAbstractGeometry->segmentize() not returning correct WkbType for multi-geometries

olivierdalang:
## Description

This fixes an issue where `QgsAbstractGeometry->segmentize()` did not return a linear WkbTypes when used on collections geometries.

```python
cpd_curve = QgsGeometry.fromWkt('COMPOUNDCURVE(CIRCULARSTRING(0 0, 1 1, 2 0), CIRCULARSTRING(2 0, 3 -1, 4 0))')
mlt_curve = QgsGeometry.fromWkt('MULTICURVE(COMPOUNDCURVE(CIRCULARSTRING(0 0, 1 1, 2 0)), COMPOUNDCURVE(CIRCULARSTRING(2 0, 3 -1, 4 0)))')

assert QgsWkbTypes.displayString(cpd_curve.wkbType()) == 'CompoundCurve'  # just checking...
assert QgsWkbTypes.displayString(mlt_curve.wkbType()) == 'MultiCurve'  # just checking...

cpd_curve_seg = cpd_curve.constGet().segmentize()
mlt_curve_seg = mlt_curve.constGet().segmentize()

assert QgsWkbTypes.displayString(cpd_curve_seg.wkbType()) == 'LineString'  # Works !
assert QgsWkbTypes.displayString(mlt_curve_seg.wkbType()) == 'MultiLineString'  # Assertion error : 'MultiCurve' != 'MultiLineString'
```

I added the `QgsWkbTypes::linearType( Type type )` static method, which is the 
symmetric of ``QgsWkbTypes::curveType( Type type )`.

The test could be expanded a little bit, to test also for other polygons, contents of GeometryCollection, etc. But wanted to make sure the approach is OK first.


src/core/geometry/qgswkbtypes.h
```diff
@@ -566,6 +567,147 @@ class CORE_EXPORT QgsWkbTypes
...
+     * \see curveType()
+     * \see isMultiType()
+     * \see isCurvedType()
+     * \see singleType()
+     * \see flatType()
+     * \see multiType()
+     *
+     * \since QGIS 3.14
+     */
+    static Type linearType( Type type )
```
nyalldawson, line 582: this is definitely a useful method to have. You'll need to add unit tests like this though: https://github.com/qgis/QGIS/blob/297ef133a35c6f41f18a5cecbecc537cb712abfa/tests/src/python/test_qgsgeometry.py#L2751

src/core/geometry/qgswkbtypes.h
```diff
@@ -566,6 +567,147 @@ class CORE_EXPORT QgsWkbTypes
...
+    {
+      switch ( type )
+      {
+        case Unknown:
+        case Triangle:
+        case TriangleZ:
+        case TriangleM:
+        case TriangleZM:
+          return Unknown;
```
nyalldawson, line 591: triangle types should return the same value -- those are already linear

src/core/geometry/qgswkbtypes.h
```diff
@@ -566,6 +567,147 @@ class CORE_EXPORT QgsWkbTypes
...
+          return LineString;
+
+        case MultiLineString:
+        case MultiCurve:
+          return MultiLineString;
+
+        case LineStringZ:
+        case CompoundCurveZ:
+        case CircularStringZ:
+        case LineString25D:
```
nyalldawson, line 641: better return LineString25D for LineString25D

src/core/geometry/qgswkbtypes.h
```diff
@@ -566,6 +567,147 @@ class CORE_EXPORT QgsWkbTypes
...
+          return LineStringZ;
+
+        case MultiLineStringZ:
+        case MultiCurveZ:
+        case MultiLineString25D:
```
nyalldawson, line 646: same here -- better return MultiLineString25D

src/core/geometry/qgswkbtypes.h
```diff
@@ -566,6 +567,147 @@ class CORE_EXPORT QgsWkbTypes
...
+        case CurvePolygon:
+          return Polygon;
+
+        case MultiPolygon:
+        case MultiSurface:
+          return MultiPolygon;
+
+        case PolygonZ:
+        case CurvePolygonZ:
+        case Polygon25D:
```
nyalldawson, line 677: ditto

src/core/geometry/qgswkbtypes.h
```diff
@@ -566,6 +567,147 @@ class CORE_EXPORT QgsWkbTypes
...
+          return PolygonZ;
+
+        case MultiPolygonZ:
+        case MultiSurfaceZ:
+        case MultiPolygon25D:
```
nyalldawson, line 682: and this one

nyalldawson:
Approach looks good to me!

lbartoletti (approved):

olivierdalang:
@nyalldawson thanks for the review ! I went a bit quickly over the types indeed :-) Will make the suggested changes and write some tests. 

About the implementation of `linearType`, I took it from `curveType`, but it could be much less verbose and hence a bit less error prone if by default we return the same type, and only treated the curved geometries, like this :
```
switch ( type )
{
    ...
    case CurvePolygonM:
        return PolygonM;

    case CurvePolygonZM:
        return PolygonZM;
    ...
    default:
        return type;
}
```

The same applies to `curveTypes`, but I wouldn't change curveType as part of this PR as it may break quite a few existing things...

It also makes the output a bit more strict, as currently, some already linear types would change. For instance (in curveTypes) there is
```
        case LineString:
        case CompoundCurve:
        case CircularString:
          return CompoundCurve;
```
I don't see why the curveType of CircularString is not CircularString.

What do you think ? In the long term, whatever the implementation, we need to think of update this in case there's a new geometry type appearing. It may pass through a bit more silently with my implementation, but new geometry types don't happen everyday.

nyalldawson:
> About the implementation of linearType, I took it from curveType, but it could be much less verbose and hence a bit less error prone if by default we return the same type, and only treated the curved geometries, like this :

In general I'm  not a fan of adding "default" statements to switches. I think the big advantage of using a switch with an enum is that we get explicit compile time warnings when a switch doesn't explicitly cover a newly added value. That makes it much easier to extend an enum and be confident that you've correctly addressed all the code paths which need to handle the new option.

> I don't see why the curveType of CircularString is not CircularString.

This seems odd to me too

nyalldawson (approved):

nyalldawson:
Looks good! I think this is ready for merge. We should backport too.

olivierdalang:
Great, thanks for the detailed review Nyall :-)

PR #35878, opened 2020-04-20
always return correct field-alias mapping from attributeAliases() call (fix #31870)

### LINKED ISSUE #31870: strage behaviour of attributeAliases() function
The function attributeAliases() returns an empty dictionary if the layer has been loaded in a QGIS project. Restarting QGIS or adding even only one alias, it returns the right results. 

image

image

image

Tested on Windows and Linux - QGIS version 3.4 and 3.8

alexbruy:
## Description
Right after adding vector layer to the project calling `attributeAliases()` on this layer returns an empty dictionary instead of mapping field-alias. But if at least one alias was set via "Layer properties - Attributes form" then `attributeAliases()` returns non-empty dictionary. After clearing previously set alias `attributeAliases()` continues to return non-empty dictionary. This is a bit inconsistent and confusing.

Proposed PR fixes this by initializing field-alias mapping on layer construction. Fixes #31870.

### DIFF
--- a/src/core/qgsfieldmodel.cpp
+++ b/src/core/qgsfieldmodel.cpp
@@ -36,7 +36,7 @@ QModelIndex QgsFieldModel::indexFromName( const QString &fieldName )
   // only non-empty names should be used here, as by default all fields
   // have no aliases set and calling key() fill return just first value
   // from the aliases map
-  if ( mLayer && !fldName.isNull() )
+  if ( mLayer && !fldName.isEmpty() )
   {
     // the name could be an alias
     // it would be better to have "display name" directly in QgsFields
--- a/src/core/qgsfieldmodel.cpp
+++ b/src/core/qgsfieldmodel.cpp
@@ -33,7 +33,10 @@ QModelIndex QgsFieldModel::indexFromName( const QString &fieldName )
 {
   QString fldName( fieldName ); // we may need a copy
 
-  if ( mLayer )
+  // only non-empty names should be used here, as by default all fields
+  // have no aliases set and calling key() fill return just first value
+  // from the aliases map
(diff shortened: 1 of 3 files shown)


src/core/qgsvectorlayer.cpp
```diff
@@ -178,6 +178,11 @@ QgsVectorLayer::QgsVectorLayer( const QString &vectorLayerPath,
     setDataSource( vectorLayerPath, baseName, providerKey, providerOptions, options.loadDefaultStyle );
   }
 
+  for ( const QgsField &field : mFields )
```
m-kuhn, line 181:
```suggestion
  for ( const QgsField &field : qgis::as_const( mFields ) )
```

alexbruy: Does it make sense to update other occurences?

m-kuhn: If they are in non const methods, absolutely. If they are in const methods it's not required (but I think it would not hurt and definitely ease reviewing)

src/core/qgsfieldmodel.cpp
```diff
@@ -33,7 +33,10 @@ QModelIndex QgsFieldModel::indexFromName( const QString &fieldName )
 {
   QString fldName( fieldName ); // we may need a copy
 
-  if ( mLayer )
+  // only non-empty names should be used here, as by default all fields
+  // have no aliases set and calling key() fill return just first value
+  // from the aliases map
+  if ( mLayer && !fldName.isNull() )
```
nyalldawson, line 39: .isEmpty()

alexbruy: Fixed

PR #35881, opened 2020-04-20
Correctly utilise expression contexts for attribute form container visibility

### LINKED ISSUE #35558: Problem with qgis_version_no qgis_short_version and qgis_version expressions controling forms visibility
There is an issue with @qgis_version_no, @qgis_short_version and @qgis_version expressions, when used in "Control visibility by expression", in Attributes Form (Drag and drop designer).

To reproduce:

1. Please open the sample shapefile provided here in any QGIS version (3.10.4, 3.12.1 or master 3.13):
https://cld.pt/dl/download/8ac48104-0d45-41fb-9dbd-85bd1d7d39e1/qgis_version_sample.zip

2. Identify any feature. You will see something like the following image, with 2 tabs:

imagem

3. However, the form has 3 tabs, as you can see if you open the Layer Properties -> Attributes Form:

imagem

Tab 'ID' visibility is controlled by this expression (using QGIS 3.10.4).

```
CASE 
  WHEN ( @qgis_version_no = 31004 ) THEN 1
  ELSE 0
END
```

If you are using another QGIS version, change @qgis_version_no value accordingly.

If you look at the Output preview, it is correct, as it shows 1. But it is not followed by the result in the form.

4. Changing the expression to always drop in the first condition of Case When, it works as expected:

imagem

imagem

The issue is only with @qgis_version_no, @qgis_short_version and @qgis_version expressions. 
And it happens the same way in 3.10.4, 3.12.1 and master. Tested in Windows.

nyalldawson:
-  we shouldn't use a single member instance of the context here, because continually adding new scopes to that single instance will eventually cause the context to become massive and slow
- correctly populate the context with the global/project/layer scopes

Fixes #35558

### DIFF
--- a/src/gui/attributeformconfig/qgsattributeformcontaineredit.cpp
+++ b/src/gui/attributeformconfig/qgsattributeformcontaineredit.cpp
@@ -48,6 +48,11 @@ QgsAttributeFormContainerEdit::QgsAttributeFormContainerEdit( QTreeWidgetItem *i
   connect( mShowAsGroupBoxCheckBox, &QCheckBox::stateChanged, mShowLabelCheckBox, &QCheckBox::setEnabled );
 }
 
+void QgsAttributeFormContainerEdit::registerExpressionContextGenerator( QgsExpressionContextGenerator *generator )
+{
+  mVisibilityExpressionWidget->registerExpressionContextGenerator( generator );
+}
+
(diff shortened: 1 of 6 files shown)


m-kuhn (approved):

PedroVenancio:
Hi @nyalldawson 

This fix cannot be backported? 

Thank you very much!!

PR #35886, opened 2020-04-21
Fix double-rendering of layout legends on some circumstances

nyalldawson:


nirvn:
thanks

nyalldawson:
@3nids this is the bug you encountered a couple of months ago

PR #35889, opened 2020-04-21
add is_multipart expression for geometry

JanCaha:
Adds expression is_multipart which exposes function isMultipart from QgsGeometry.


nyalldawson (approved):
Nice!

PR #35890, opened 2020-04-21
Changed filter for Android provider registry due to changes in naming

sklencar:
Providers for Android are now with arch postfix which causes that some providers have been filter out in provider registry.

Not related to any reported issue.


wonder-sk:
To extend on this a bit - with recent Android NDKs the default is that shared objects are built with a suffix containing platform name - so e.g. WMS provider normally compiled as `libwmsprovider.so` is named  `libwmsprovider_arm64-v8a.so`. The filter for provider modules thus needed a small adjustment to allow that.

wonder-sk (approved):

PR #35891, opened 2020-04-21
include missed WFS connection settings into XML exchange file (fix #35866)

### LINKED ISSUE #35866: Exporting WFS connections list will not include WFS version
The seting of WFS version in not exported when you export to xml-file

alexbruy:
## Description
WFS connection settings now have much more information then just connection name and URL: WFS version, axis orientation settings, pagination etc. All these are missed from the XML exchange file when connections are exported for sharing.

This PR adds these missed WFS settings to the exchange file. Fixes #35866

### DIFF
--- a/src/gui/qgsmanageconnectionsdialog.cpp
+++ b/src/gui/qgsmanageconnectionsdialog.cpp
@@ -475,7 +475,7 @@ QDomDocument QgsManageConnectionsDialog::saveWfsConnections( const QStringList &
 {
   QDomDocument doc( QStringLiteral( "connections" ) );
   QDomElement root = doc.createElement( QStringLiteral( "qgsWFSConnections" ) );
-  root.setAttribute( QStringLiteral( "version" ), QStringLiteral( "1.0" ) );
+  root.setAttribute( QStringLiteral( "version" ), QStringLiteral( "1.1" ) );
   doc.appendChild( root );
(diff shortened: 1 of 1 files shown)


src/gui/qgsmanageconnectionsdialog.cpp
```diff
@@ -487,7 +487,12 @@ QDomDocument QgsManageConnectionsDialog::saveWfsConnections( const QStringList &
     el.setAttribute( QStringLiteral( "name" ), connections[ i ] );
     el.setAttribute( QStringLiteral( "url" ), settings.value( path + connections[ i ] + "/url", "" ).toString() );
 
-    el.setAttribute( QStringLiteral( "referer" ), settings.value( path + connections[ i ] + "/referer", "" ).toString() );
+    el.setAttribute( QStringLiteral( "version" ), settings.value( path + connections[ i ] + "/version", "" ).toString() );
```
nyalldawson, line 490:
```suggestion
    el.setAttribute( QStringLiteral( "version" ), settings.value( path + connections[ i ] + "/version" ).toString() );
```

(elsewhere too -- it's not needed here, an empty string will be the default if you don't specify anything)

alexbruy: Fixed.

PR #35893, opened 2020-04-21
assign correct geometry calculator to expression builder widget in the Select by Expression dialog (fix #35628)

### LINKED ISSUE #35628: Wrong preview values for $area and $perimeter in the "Select By Expression"
**Describe the bug**
Very similar issue to [issue #17334](https://github.com/qgis/QGIS/issues/17334), I noticed in the Select By Expression dialog there is strange values for `$area` and `$perimeter`.

**How to Reproduce**

1. Create a new QGIS project and select the 2154 CRS (I think it's not the source of the bug but it's a projected system I'm used too (french reference CRS))
2. Create a 100 m square grid with the extent `400000,400200,7000000,7000200 [EPSG:2154]` but the bug appears for all polygon layers I tested, just it's easier with squares.
3. You have 4 square features with a logical projected perimeter of 400 m and area of 10 000 m2
4. Open the Calculate Field Dialog and verify :
  - OK: `perimeter($geometry)` : 400
  - OK: `area($geometry)` : 10000
  - OK: `$perimeter` : 399.6146268034432 (around 400, ellipsoidal distance)
  - OK: `$area` : 9980.74059100748 (around 10000, ellipsoidal area)
5. Same but with the Select By Expression Dialog :
  - OK: `perimeter($geometry)` : 400
  - OK: `area($geometry)` : 10000
  - KO: `$perimeter` : **44527796.31730943**
  - KO: `$area` : **123920290305000**

**QGIS and OS versions**

Version de QGIS | 3.12.1-București | Révision du code | 121cc00ff0
-- | -- | -- | --
Compilé avec Qt | 5.12.4 | Utilisant Qt | 5.12.4
Compilé avec GDAL/OGR | 2.4.2 | Utilisé avec GDAL/OGR | 2.4.2
Version du client PostgreSQL | 11.7 (Ubuntu 11.7-0ubuntu0.19.10.1) | Version de SpatiaLite | 4.3.0a
Version de QWT | 6.1.4 | Version de QScintilla2 | 2.10.4
OS Version | Ubuntu 19.10
Extensions Python actives | french_locator_filter; plugin_reloader; pluginbuilder3; firstaid; quick_map_services; processing; db_manager; MetaSearch

alexbruy:
## Description
When projected CRS is used preview for `$perimeter` and `$area` expressions gives wrong results, although selection itself works fine. At the same time preview for same layer/expression in the Field Calculator is correct. Fixes #35628.

### DIFF
--- a/src/gui/qgsexpressionselectiondialog.cpp
+++ b/src/gui/qgsexpressionselectiondialog.cpp
@@ -59,6 +59,11 @@ QgsExpressionSelectionDialog::QgsExpressionSelectionDialog( QgsVectorLayer *laye
   mExpressionBuilder->initWithLayer( layer, context, QStringLiteral( "selection" ) );
   mExpressionBuilder->setExpressionText( startText );
 
+  QgsDistanceArea da;
+  da.setSourceCrs( layer->crs(), QgsProject::instance()->transformContext() );
+  da.setEllipsoid( QgsProject::instance()->ellipsoid() );
+  mExpressionBuilder->setGeomCalculator( da );
+
   // by default, zoom to features is hidden, shown only if canvas is set
   mButtonZoomToFeatures->setVisible( false );
 


nyalldawson:
This is rather subtle -- but there's a better way to handle this which is by removing ALL the calls to setGeomCalculator in this class.

setGeomCalculator is mostly deprecated now, as it's handled automatically by the expression context WHEN the expression context has the project scope set.

alexbruy:
Maybe I misundertstand code, but seems expression context is set and includes project scope https://github.com/qgis/QGIS/blob/a270618408fc016e126bb4b017de5b0d5f48b74d/src/gui/qgsexpressionselectiondialog.cpp#L58-L59

BTW, same with field calculator, it also has expression context set, but additionaly uses `setGeomCalculator`.

nyalldawson:
Any call to "setGeomCalculator" cancels the auto selection of this from the project scope. That's why you just need to remove ALL calls to setGeomCalculator from this class and it work correctly from the existing project scope.

3nids:
> Any call to "setGeomCalculator" cancels the auto selection of this from the project scope. That's why you just need to remove ALL calls to setGeomCalculator from this class and it work correctly from the existing project scope.

Any chance to place an assertion to avoid this?

nyalldawson:
@3nids what would you propose? maybe we should just deprecate setGeomCalculator and force this to always be done via the project scope. But that may potentially be limiting...

3nids:
I was wondering if you could check in `setGeomCalculator` if the project scope was already added, and do a `Q_ASSERT` there.

alexbruy:
Ok, closing this for now. Will try to investigate why getting projection settings from context does not work.

nyalldawson:
I told you why 😁 (and tested it) . Just remove all calls to setGeomCalculator

alexbruy:
Probably I don't get it right. There were no calls to `setGeomCalculator` in the `qgsexpressionselectiondialog.cpp` before this PR.

PR #35897, opened 2020-04-21
use focal to build Docker images

3nids:


nirvn:
Nice. 

Understanding this was triggered by the EOL of Ubuntu Disco, the one downside of us moving to Focal: we lose testing of QGIS PRs/commits under gdal 2/proj 5. It used to be travis -> gdal2/proj5 and azure -> gdal3/proj6.

We might be OK with it, I'm just verbalizing it so we're aware of that fact.

nirvn:
I just saw the only other alternative is Eoan, which reaches EOL in July 2020. Focal it is then :)

3nids:
no, the PR are still tested against bionic.
Disco was introduced to have newer version of sip for the documentation.
I failed at upgrading the tests to disco on that time, so I kept the 2 in parallel, which will continue a bit.

3nids:
merging to test the build on cron jobs.

3nids:
if there is any issue, this can be reverted later on

nirvn:
@3nids , oh great then. Sorry for the noise.

PR #35898, opened 2020-04-21
Close 3D measurement on 3D canvas close

### LINKED ISSUE #34632: Crashing when closing 3D measurement tool window
**Describe the bug**
When trying to close 3d Measurement Tool window after closing 3D Map Viewer, QGIS crashes

**How to Reproduce**
- Open 3D Map Viewer
- Click 3D measurement tool button
- Close 3D Map Viewer window
- Try to close 3D measurement tool window

3d2

**QGIS and OS versions**

QGIS version
3.12.0-București
QGIS code revision
cd141490ec
Compiled against Qt
5.11.2
Running against Qt
5.11.2
Compiled against GDAL/OGR
3.0.4
Running against GDAL/OGR
3.0.4
Compiled against GEOS
3.8.0-CAPI-1.13.1
Running against GEOS
3.8.0-CAPI-1.13.1 
Compiled against SQLite
3.29.0
Running against SQLite
3.29.0
PostgreSQL Client Version
11.5
SpatiaLite Version
4.3.0
QWT Version
6.1.3
QScintilla2 Version
2.10.8
Compiled against PROJ
6.3.1
Running against PROJ
OS Version
Windows 10 (10.0)
Active python plugins
openlayers_plugin; 
pluginbuilder3; 
QuickOSM; 
QuickWKT; 
db_manager; 
MetaSearch; 
processing

roya0045:
## Description

Proposal to Fix #34632 

This changes ensure that the measurement dialog is closed when the 3D canvas is also closed to prevent crash and other inconsistencies.

### DIFF
--- a/src/app/3d/qgs3dmapcanvas.cpp
+++ b/src/app/3d/qgs3dmapcanvas.cpp
@@ -58,10 +58,11 @@ Qgs3DMapCanvas::Qgs3DMapCanvas( QWidget *parent )
 
 Qgs3DMapCanvas::~Qgs3DMapCanvas()
 {
+  if ( mMapTool )
+    mMapTool->deactivate();
   // make sure the scene is deleted while map settings object is still alive
   delete mScene;
   mScene = nullptr;
-
   delete mMap;
   mMap = nullptr;
 }
--- a/src/app/3d/qgs3dmaptoolmeasureline.cpp
+++ b/src/app/3d/qgs3dmaptoolmeasureline.cpp
@@ -57,6 +57,8 @@ Qgs3DMapToolMeasureLine::Qgs3DMapToolMeasureLine( Qgs3DMapCanvas *canvas )
   mDialog->setWindowFlags( mDialog->windowFlags() | Qt::Tool );
   mDialog->restorePosition();
 
+  // Update scale if the terrain vertical scale changed
+  connect( mCanvas->map(), &Qgs3DMapSettings::terrainVerticalScaleChanged, this, &Qgs3DMapToolMeasureLine::updateMeasurementLayer );
   connect( canvas, &Qgs3DMapCanvas::mapSettingsChanged, this, &Qgs3DMapToolMeasureLine::onMapSettingsChanged );
 }
 
@@ -97,7 +99,6 @@ void Qgs3DMapToolMeasureLine::activate()
     updateSettings();
     mIsAlreadyActivated = true;
   }
-
   // Show dialog
   mDialog->updateSettings();
   mDialog->show();
@@ -281,3 +282,4 @@ QVector<QgsPoint> Qgs3DMapToolMeasureLine::points() const
 {
   return mPoints;
 }
+
--- a/src/app/3d/qgs3dmaptoolmeasureline.cpp
+++ b/src/app/3d/qgs3dmaptoolmeasureline.cpp
@@ -58,7 +58,6 @@ Qgs3DMapToolMeasureLine::Qgs3DMapToolMeasureLine( Qgs3DMapCanvas *canvas )
   mDialog->restorePosition();
 
   // Update scale if the terrain vertical scale changed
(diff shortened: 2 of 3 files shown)


roya0045:
@ismailsunni @wonder-sk Are you ok with this proposal?

ismailsunni:
Hi @roya0045 
Thanks for the PR! The logic looks good to me.

wonder-sk:
I would suggest a different approach: in the destructor of Qgs3DMapCanvas the canvas should deactivate the current map tool (if any). This deactivation would also close the dialog. I think this way it the code would be cleaner (no extra signals/slots to follow) and also it would also mirror how the 2D QgsMapCanvas works already...

roya0045:
@wonder-sk good point, I was hesitating between the two approaches. I'll try to rework this when I get the chance.

roya0045:
@wonder-sk I made the changes, let me know if this is what you had in mind. Sadly I still need one connection.

src/app/3d/qgs3dmaptoolmeasureline.cpp
```diff
@@ -52,11 +52,20 @@ Qgs3DMapToolMeasureLine::Qgs3DMapToolMeasureLine( Qgs3DMapCanvas *canvas )
 {
   mPickHandler.reset( new Qgs3DMapToolMeasureLinePickHandler( this ) );
 
-  // Dialog
-  mDialog = new Qgs3DMeasureDialog( this );
-  mDialog->setWindowFlags( mDialog->windowFlags() | Qt::Tool );
-  mDialog->restorePosition();
+  if ( mDialog )
+    activate();
```
wonder-sk, line 56: I don't understand why this change is necessary - mDialog should be always nullptr in the constructor at this point...

src/app/3d/qgs3dmapcanvas.cpp
```diff
@@ -61,7 +61,7 @@ Qgs3DMapCanvas::~Qgs3DMapCanvas()
   // make sure the scene is deleted while map settings object is still alive
   delete mScene;
   mScene = nullptr;
-
+  emit canvasClosing();
```
wonder-sk, line 64: I think if you put something like
```
if ( mMapTool )
  mMapTool->deactivate();
```
at the top of the destructor body (before the scene is deleted) you don't need any signal about canvas getting closed...

src/app/3d/qgs3dmaptoolmeasureline.cpp
```diff
@@ -106,12 +115,15 @@ void Qgs3DMapToolMeasureLine::activate()
 void Qgs3DMapToolMeasureLine::deactivate()
 {
   restart();
-  if ( QgsTerrainEntity *terrainEntity = mCanvas->scene()->terrainEntity() )
+  if ( !mCanvasClosed )
```
wonder-sk, line 118: These changes also should not be necessary if the tool would be deactivated before the scene is deleted.

wonder-sk (commented):
Thanks for the update - good improvements, but the code could be further simplified to avoid any new signals - see my comments...

src/app/3d/qgs3dmaptoolmeasureline.cpp
roya0045: I thought that since the dialog was deactivated the aim was to reactivate it at some point. I'm not sure if by always deactivating we end up with plenty of hidden dialogs or if it get cleaned up at some point.

And so I thought that by moving to deactivate instead of a destructor was to reuse the dialog and I thought that this was the way to do so.

src/app/3d/qgs3dmapcanvas.cpp
roya0045: You're right, this was an oversight on my end.

wonder-sk (approved):
Thanks for the updates, now it looks really good :-)

roya0045:
@wonder-sk I just completely missed some elements when doing the code review and over-engineered  things 😞 Thanks for taking the time to guide me back on the right path.

roya0045:
@nyalldawson thank you Nyall.

PR #35900, opened 2020-04-21
move the QGIS Server validator from the project properties to its own class

Gustry:
## Description

Follow up of https://github.com/qgis/QGIS/pull/35464

In the project properties, QGIS Server tab, there is a QGIS Server check. It gives a HTML output about problems that you may have when publishing with QGIS Server.

We want to be able to run these checks from a Python plugin (Lizmap) to check the configuration before users send the project on the server.

After this PR, we will add more checks to this validator.


src/core/qgsprojectservervalidator.h
```diff
@@ -0,0 +1,99 @@
...
+      QgsProjectServerValidator::ValidationError error;
+
+      /**
+       * Identifier related to the error.
+       */
+      QVariant identifier;
+    };
+
+    /**
+     * Validates a project to avoid some problems on QGIS Server, and returns TRUE if it's considered valid.
```
m-kuhn, line 85:
```suggestion
     * Validates a project to detect problems on QGIS Server, and returns TRUE if it's considered valid.
```

src/core/qgsprojectservervalidator.h
```diff
@@ -0,0 +1,99 @@
...
+        , identifier( identifier )
+      {}
+
+      /**
+       * Error which occurred during the validation process.
+       */
+      QgsProjectServerValidator::ValidationError error;
+
+      /**
+       * Identifier related to the error.
```
m-kuhn, line 79: What's that? Any chance to be a bit more verbose?

src/core/qgsprojectservervalidator.h
```diff
@@ -0,0 +1,99 @@
...
+
+#include "qgis_core.h"
+#include "qgslayertreegroup.h"
+#include "qgslayertree.h"
+
+
+/**
+ * \ingroup core
+ * \class QgsProjectServerValidator
+ * \brief Project server validator.
```
m-kuhn, line 29:
```suggestion
 * \brief Validates the server specific parts of the configuration of a QGIS project.
```

src/app/qgsprojectproperties.cpp
```diff
@@ -1907,60 +1908,28 @@ void QgsProjectProperties::pbnWCSLayersDeselectAll_clicked()
...
+    for ( const QgsProjectServerValidator::ValidationResult &result : qgis::as_const( validationResults ) )
+    {
+      errors += QLatin1String( "<b>" ) % QgsProjectServerValidator::displayValidationError( result.error ) % QLatin1String( " :</b> " );
+      errors += result.identifier.toString();
+    }
   }
   else
   {
-    teOWSChecker->setHtml( teOWSChecker->toHtml() + "<h1>" + tr( "All layer encodings are set" ) + "</h1>" );
+    errors += QString( tr( "Ok, it seems valid." ) );
```
m-kuhn, line 1925:
```suggestion
    errors += tr( "Ok, it seems valid." );
```

src/core/qgsprojectservervalidator.cpp
```diff
@@ -0,0 +1,132 @@
...
+        QString shortName = layer->shortName();
+        if ( shortName.isEmpty() )
+          owsNames << layer->name();
+        else
+          owsNames << shortName;
+
+        if ( layer->type() == QgsMapLayerType::VectorLayer )
+        {
+          QgsVectorLayer *vl = static_cast<QgsVectorLayer *>( layer );
+          if ( vl->dataProvider()->encoding() == QLatin1String( "System" ) )
```
m-kuhn, line 70: Not sure if that's somehow prevented by the context already

```suggestion
          if ( vl->dataProvider() && vl->dataProvider()->encoding() == QLatin1String( "System" ) )
```

src/app/qgsprojectproperties.cpp
nyalldawson: Or more professionally: "Project is valid" / "Project validated successfully"

src/core/qgsprojectservervalidator.cpp
```diff
@@ -0,0 +1,132 @@
...
+#include "qgsprojectservervalidator.h"
+#include "qgsvectorlayer.h"
+
+
+QString QgsProjectServerValidator::displayValidationError( QgsProjectServerValidator::ValidationError error )
+{
+  switch ( error )
+  {
+    case QgsProjectServerValidator::Encoding:
+      return QObject::tr( "Encoding is not set properly. A non 'System' encoding is required" );
```
nyalldawson, line 30:
```suggestion
      return QObject::tr( "Encoding is not correctly set. A non 'System' encoding is required" );
```

src/core/qgsprojectservervalidator.cpp
```diff
@@ -0,0 +1,132 @@
...
+    case QgsProjectServerValidator::ShortNames:
+      return QObject::tr( "Layer short name is not valid. It must start with an unaccented alphabetical letter, followed by any alphanumeric letters, dot, dash or underscore" );
+    case QgsProjectServerValidator::DuplicatedNames:
+      return QObject::tr( "One or more layers or groups have the same name or short name. Both the 'name' and 'short name' for layers and groups needs to be unique" );
```
nyalldawson, line 34:
```suggestion
      return QObject::tr( "One or more layers or groups have the same name or short name. Both the 'name' and 'short name' for layers and groups must be unique" );
```

Gustry:
Thanks for your review!

It seems the code layout doesn't like my sip out or api doc.
https://cdash.orfeo-toolbox.org/viewBuildError.php?type=1&buildid=39189
Any hint?

Side note, I have run `scripts/sipify_all.sh` to be sure that my SIP were updated with the error above but I have another file updated `python/3d/auto_generated/qgslayoutitem3dmap.sip.in`. Is that normal?

m-kuhn:
`#include qgis_sip.h`

Gustry:
Ok thanks, indeed it fixes travis now!

rldhont (approved):

PR #35903, opened 2020-04-21
[win dev] Make build directory of package-nightly.cmd configurable

olivierdalang:
`package-nightly.cmd` is the easiest and most up to date way to make a build on Windows, but currently it hard codes the build directory to `%CD%\build-%PACKAGENAME%-%ARCH%`.

This has a few issues :
- if the QGIS source is a bit deep in the folder hierarchy, one can get a `The specified path, file name, or both are too long. [...]` error (even with long paths enabled in Windows it seems).
- the build output is very deep in the hierarchy
- the build directory is not .gitignored

This PR adds the ability to provide a BUILDDIR env var, to customize it. If unset, behaviour is unchanged. By the way, the PR also .gitignores the default build dir.

I know this script's main use if for packaging, and not for developing, but it's a reasonably simple modification :-) Following this I'll try to propose some other commits to improve dev experience on Windows (looking at you, prepare-commit.sh).

### DIFF
--- a/ms-windows/osgeo4w/package-nightly.cmd
+++ b/ms-windows/osgeo4w/package-nightly.cmd
@@ -31,7 +31,7 @@ if "%SITE%"=="" set SITE=qgis.org
 if "%TARGET%"=="" set TARGET=Nightly
 if "%BUILDNAME%"=="" set BUILDNAME=%PACKAGENAME%-%VERSION%%SHA%-%TARGET%-VC14-%ARCH%
 
-set BUILDDIR=%CD%\build-%PACKAGENAME%-%ARCH%
+if "%BUILDDIR%"=="" set BUILDDIR=%CD%\build-%PACKAGENAME%-%ARCH%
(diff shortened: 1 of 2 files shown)


ms-windows/osgeo4w/package-nightly.cmd
```diff
@@ -31,7 +31,7 @@ if "%SITE%"=="" set SITE=qgis.org
 if "%TARGET%"=="" set TARGET=Nightly
 if "%BUILDNAME%"=="" set BUILDNAME=%PACKAGENAME%-%VERSION%%SHA%-%TARGET%-VC14-%ARCH%
 
-set BUILDDIR=%CD%\build-%PACKAGENAME%-%ARCH%
+if "%BUILDDIR%"=="" set BUILDDIR=%CD%\build-%PACKAGENAME%-%ARCH%
```
m-kuhn, line 34: Can you document this somewhere?

Issue #33516, opened 2019-12-25
Build failure with txt2tags 3.3 (dropped --css-sugar & --css-inside)


sebastic:
QGIS fails to build on Debian unstable where `txt2tags` was recently updated to 3.3 which dropped the `--css-sugar` & `--css-inside` options, see:

 https://github.com/txt2tags/txt2tags/blob/v3/CHANGELOG.md#version-33-2019-11-04

The following patch fixes the issue for me, but likely won't suffice for older `txt2tags` versions where the options are still supported:
```diff
Description: Drop options removed from txt2tags.
 See: https://github.com/txt2tags/txt2tags/blob/v3/CHANGELOG.md#version-33-2019-11-04
Author: Bas Couwenberg <sebastic@debian.org>

--- a/doc/INSTALL.t2t
+++ b/doc/INSTALL.t2t
@@ -4,7 +4,7 @@ Building QGIS from source - step by step
 
 %! target       : html
 %! style        : style.css
-%! Options      : --toc --toc-level 2 --enum-title --css-sugar --css-inside
+%! Options      : --toc --toc-level 2 --enum-title
 %! preproc      : TUT_URL   https://qgis.org
 %! PostProc(html): '(?i)(<pre>)' '<div class="code">\1'
 %! PostProc(html): '(?i)(</pre>)' '\1</div>'
--- a/doc/news.t2t
+++ b/doc/news.t2t
@@ -4,7 +4,7 @@ Change history for the QGIS Project
 
 %! target       : html
 %! style        : style.css
-%! Options      : --toc --toc-level 1 --enum-title --css-sugar --css-inside
+%! Options      : --toc --toc-level 1 --enum-title
 %! preproc      : TUT_URL   https://qgis.org
 %! PostProc(html): '(?i)(<pre>)' '<div class="code">\1'
 %! PostProc(html): '(?i)(</pre>)' '\1</div>'
```

espinafre:
Hi, doc/news.t2t should be doc/NEWS.t2t because of case sensitivity in Linux filesystems. With this minor change, your patch works for me.

sebastic:
> doc/news.t2t should be doc/NEWS.t2t

Not on the 3.4 branch:

 https://github.com/qgis/QGIS/blob/release-3_4/doc/news.t2t

PR #35914, opened 2020-04-22
Fix build errors for txt2tags on macos

timlinux:
## Description

On macOS brew, we only get txt2tags 3.7 which deprecates the ``--toc-level`` option for txt2tags

This PR fixes the doc build though I think creates some small formatting issues in the generated HTML. Similar issue to https://github.com/qgis/QGIS/issues/33516

### DIFF
--- a/doc/INSTALL.t2t
+++ b/doc/INSTALL.t2t
@@ -4,7 +4,7 @@ Building QGIS from source - step by step
 
 %! target       : html
 %! style        : style.css
-%! Options      : --toc --toc-level 2 --enum-title
+%! Options      : --enum-title
 %! preproc      : TUT_URL   https://qgis.org
 %! PostProc(html): '(?i)(<pre>)' '<div class="code">\1'
 %! PostProc(html): '(?i)(</pre>)' '\1</div>'
--- a/doc/NEWS.t2t
+++ b/doc/NEWS.t2t
@@ -4,7 +4,7 @@ Change history for the QGIS Project
 
 %! target       : html
 %! style        : style.css
-%! Options      : --toc --toc-level 1 --enum-title
+%! Options      : --enum-title
 %! preproc      : TUT_URL   https://qgis.org
 %! PostProc(html): '(?i)(<pre>)' '<div class="code">\1'
 %! PostProc(html): '(?i)(</pre>)' '\1</div>'


timlinux:
See also https://github.com/txt2tags/txt2tags/releases/tag/3.7

They don't seem to give any explanation as to why they are removed or what to use instead,

PeterPetrik:
it works for me but @NOW@ is not expanded and also the additional level of items is added to the table of contents, is it OK for you Tim? 

related to https://github.com/qgis/QGIS/pull/35915

PeterPetrik:
ok, looks good, I was testing without removal of --toc

PeterPetrik:
with this change for me the  @NOW@ and @MODTIME@ are not expanded and TOC is missing at all

timlinux:
@PeterPetrik 
>with this change for me the @now@ and @MODTIME@ are not expanded and TOC is missing at all

Yeah for me too - I am still trying to find out the new replacements for those directives

timlinux:
@PeterPetrik so it seems a bunch of things are deprecated in python 3 supported version of txt2tags includeing the toc levels and date expressions:

https://github.com/txt2tags/txt2tags#feature-differences-between-versions

PeterPetrik:
maybe it would easier to just use different technology for building the INSTALL doc? markdown?

wonder-sk:
> maybe it would easier to just use different technology for building the INSTALL doc? markdown?

Yeah I would vote for getting rid of txt2tags and just use Markdown everywhere - it is automatically converted to rich text by Github/Gitlab/many other tools, and there would one step and one dependency less...

timlinux:
@wonder-sk agreed, moving it to Markdown would be better for the long term.....just need to find some time to do that

Codym48:
> @wonder-sk agreed, moving it to Markdown would be better for the long term.....just need to find some time to do that

Is there an open issue for this work?

lbartoletti:
cc @rhurlin for information

lbartoletti:
> > maybe it would easier to just use different technology for building the INSTALL doc? markdown?
> 
> Yeah I would vote for getting rid of txt2tags and just use Markdown everywhere - it is automatically converted to rich text by Github/Gitlab/many other tools, and there would one step and one dependency less...

Every release is a new change for us... +1

PeterPetrik:
closing and lets fix it with markdown in #36966

PR #35926, opened 2020-04-22
[FEATURE][MESH] 1D mesh width/color varying

vcloarec:
This PR provide a new way to render 1D mesh with edge width that can vary depending on value. So now, width and color vary depending on value on edges or on vertices.
The user can define min/max value and min/max width to set the rendering.
For color, the user can define a color ramp shader (as for 2D mesh and raster).

For this new rendering, new generic classes are introduced that could be used outside the mesh framework.

strokeWidthVarying


src/core/symbology/qgsinterpolatedlinerenderer.cpp
```diff
@@ -0,0 +1,676 @@
...
+
+    QBrush brush( Qt::SolidPattern );
+    QPen pen;
+
+    brush.setColor( mStrokeColoring.color( value1 ) );
+    painter->setBrush( brush );
+    pen.setBrush( brush );
+    painter->setPen( pen );
+
+    painter->drawPie( capBox1, ( startAngle - 1 ) * 16, 182 * 16 );
```
nyalldawson, line 101: For reference-- if you are successful and want to implement this for the standard linestring renderer, you can use QgsGeometry::taperedBuffer which works nicely with multi segment line strings.

vcloarec: Thanks, I will look at!

PeterPetrik:
already review in fork and happy with the implementation.

PeterPetrik (approved):

Issue #29098, opened 2019-02-15
[Expression] Add z_min and z_max functions to extract those values from a geometry


qgib:
Author Name: **Harrissou Santanna** (@DelazJ)
Original Redmine Issue: [21280](https://issues.qgis.org/issues/21280)

Redmine category:expressions

---
QGIS provides x_min, x_max as well as y_min and y_max functions to extract those coordinates from a geometry (particularly, lines and polygons).
It could be nice to have a way to also easily extract z min and max coordinates from geometries, without resorting to processing and aggregates expressions.

phidrho:
Also there is missing m_max, m_min. Please update title as this problem is almost the same.

alexbruy:
See also #28069.

phidrho:
@m-kuhn this one can be closed as https://github.com/qgis/QGIS/pull/35928#issue-407498479 is commited

PR #35928, opened 2020-04-22
Expressions for geometry Z and M minimum and maximum

JanCaha:
Adds expressions that support extraction of minimal and maximal Z and M from geometry - including tests.

Also adds my info to contributors.json.

### DIFF
--- a/resources/function_help/json/m_max
+++ b/resources/function_help/json/m_max
@@ -2,7 +2,7 @@
   "name": "m_max",
   "type": "function",
   "description": "Returns the maximum m (measure) value of a geometry.",
-  "arguments": [ {"arg":"geom","description":"a geometry containing m values"}],
+  "arguments": [ {"arg":"geometry","description":"a geometry containing m values"}],
   "examples": [ { "expression":"m_max( make_point_m( 0,0,1 ) )", "returns":"1"},
                 { "expression":"m_max(make_line( make_point_m( 0,0,1 ), make_point_m( -1,-1,2 ), make_point_m( -2,-2,0 ) ) )", "returns":"2"}
   ]
--- a/resources/function_help/json/m_max
+++ b/resources/function_help/json/m_max
@@ -1,8 +1,8 @@
 {
   "name": "m_max",
   "type": "function",
-  "description": "Returns the maximum m (measure) coordinate of a geometry.",
-  "arguments": [ {"arg":"geom","description":"a geometry"}],
+  "description": "Returns the maximum m (measure) value of a geometry.",
+  "arguments": [ {"arg":"geom","description":"a geometry containing m values"}],
   "examples": [ { "expression":"m_max( make_point_m( 0,0,1 ) )", "returns":"1"},
                 { "expression":"m_max(make_line( make_point_m( 0,0,1 ), make_point_m( -1,-1,2 ), make_point_m( -2,-2,0 ) ) )", "returns":"2"}
   ]
--- a/resources/function_help/json/m_max
+++ b/resources/function_help/json/m_max
@@ -1,0 +1,9 @@
+{
+  "name": "m_max",
+  "type": "function",
+  "description": "Returns the maximum m (measure) coordinate of a geometry.",
+  "arguments": [ {"arg":"geom","description":"a geometry"}],
+  "examples": [ { "expression":"m_max( make_point_m( 0,0,1 ) )", "returns":"1"},
+                { "expression":"m_max(make_line( make_point_m( 0,0,1 ), make_point_m( -1,-1,2 ), make_point_m( -2,-2,0 ) ) )", "returns":"2"}
+  ]
+}
--- a/src/core/expression/qgsexpressionfunction.cpp
+++ b/src/core/expression/qgsexpressionfunction.cpp
@@ -3253,6 +3253,110 @@ static QVariant fcnYMax( const QVariantList &values, const QgsExpressionContext 
   return QVariant::fromValue( geom.boundingBox().yMaximum() );
 }
 
+static QVariant fcnZMax( const QVariantList &values, const QgsExpressionContext *, QgsExpression *parent, const QgsExpressionNodeFunction * )
+{
+  QgsGeometry geom = QgsExpressionUtils::getGeometry( values.at( 0 ), parent );
+
+  if ( geom.isNull() )
+    return QVariant();
+
+  if ( !geom.constGet()->is3D() )
+    return QVariant();
+
+  QgsVertexId vId;
+  geom.vertexIdFromVertexNr( 0, vId );
+
+  double max = geom.constGet()->vertexAt( vId ).z();
+
+  for ( auto it = geom.vertices_begin(); it != geom.vertices_end(); ++it )
+  {
+    double z = ( *it ).z();
+
+    if ( max < z )
+      max = z;
+  }
+
+  return QVariant( max );
+}
+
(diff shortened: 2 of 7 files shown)


phidrho:
this seems to fill feature request [#29098](https://github.com/qgis/QGIS/issues/29098)
@lbartoletti you are assigned so I'm mentioning you

src/core/expression/qgsexpressionfunction.cpp
```diff
@@ -3253,6 +3253,110 @@ static QVariant fcnYMax( const QVariantList &values, const QgsExpressionContext
   return QVariant::fromValue( geom.boundingBox().yMaximum() );
 }
 
+static QVariant fcnZMax( const QVariantList &values, const QgsExpressionContext *, QgsExpression *parent, const QgsExpressionNodeFunction * )
+{
+  QgsGeometry geom = QgsExpressionUtils::getGeometry( values.at( 0 ), parent );
+
+  if ( geom.isNull() )
```
nyalldawson, line 3260:
```suggestion
  if ( geom.isNull() || geom.isEmpty() )
```
(same suggestion applies to other functions too)

src/core/expression/qgsexpressionfunction.cpp
```diff
@@ -3253,6 +3253,110 @@ static QVariant fcnYMax( const QVariantList &values, const QgsExpressionContext
...
+    return QVariant();
+
+  if ( !geom.constGet()->is3D() )
+    return QVariant();
+
+  QgsVertexId vId;
+  geom.vertexIdFromVertexNr( 0, vId );
+
+  double min = geom.constGet()->vertexAt( vId ).z();
```
nyalldawson, line 3295:
```suggestion
  double min = std::numeric_limits< double >::max();
```

(same applies to m functions)

resources/function_help/json/m_max
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "m_max",
+  "type": "function",
+  "description": "Returns the maximum m (measure) coordinate of a geometry.",
+  "arguments": [ {"arg":"geom","description":"a geometry"}],
```
nyalldawson, line 5:
```suggestion
  "arguments": [ {"arg":"geometry","description":"a geometry"}],
```

nyalldawson: (also applies to others -- I realise there's inconsistencies here across different functions, but we should standardise on non-abbreviated forms)

src/core/expression/qgsexpressionfunction.cpp
```diff
@@ -5938,7 +6042,15 @@ const QList<QgsExpressionFunction *> &QgsExpression::Functions()
                                             << QgsExpressionFunction::Parameter( QStringLiteral( "y" ) ),
                                             fcnExtrude, QStringLiteral( "GeometryGroup" ), QString() )
         << new QgsStaticExpressionFunction( QStringLiteral( "is_multipart" ), QgsExpressionFunction::ParameterList() << QgsExpressionFunction::Parameter( QStringLiteral( "geometry" ) ),
-                                            fcnGeomIsMultipart, QStringLiteral( "GeometryGroup" ) );
+                                            fcnGeomIsMultipart, QStringLiteral( "GeometryGroup" ) )
+        << new QgsStaticExpressionFunction( QStringLiteral( "z_max" ), QgsExpressionFunction::ParameterList() << QgsExpressionFunction::Parameter( QStringLiteral( "geom" ) ),
```
nyalldawson, line 6046:
```suggestion
        << new QgsStaticExpressionFunction( QStringLiteral( "z_max" ), QgsExpressionFunction::ParameterList() << QgsExpressionFunction::Parameter( QStringLiteral( "geometry" ) ),
```

(see note about this in the function json documentation)

nyalldawson (commented):
Looking good -- keep these great contributions coming :smile:

nyalldawson:
Pro tip: you'll notice that there's a quasi convention followed in QGIS dev where a new feature commit has the "[feature]" prefix at the start of the commit log message, e.g.

    [feature][expressions] add expressions for min and max M and Z - including tests

tests/src/core/testqgsexpression.cpp
```diff
@@ -1184,6 +1184,22 @@ class TestQgsExpression: public QObject
...
+      QTest::newRow( "z_min point" ) << "z_min(geom_from_wkt('POINT (0 0 1)'))" << false << QVariant( 1.0 );
+      QTest::newRow( "z_min line" ) << "z_min(make_line(make_point(0,0,0),make_point(-1,-1,-2)))" << false << QVariant( -2.0 );
+      QTest::newRow( "m_max no measure" ) << "m_max(geom_from_wkt('POINT (0 0)'))" << false << QVariant();
+      QTest::newRow( "m_max NULL" ) << "m_max(geom_from_wkt(NULL))" << false << QVariant();
+      QTest::newRow( "m_max point" ) << "m_max(make_point_m(0,0,1))" << false << QVariant( 1.0 );
+      QTest::newRow( "m_max line" ) << "m_max(make_line(make_point_m(0,0,1),make_point_m(-1,-1,2),make_point_m(-2,-2,0)))" << false << QVariant( 2.0 );
+      QTest::newRow( "m_min no measure" ) << "m_min(geom_from_wkt('POINT (0 0)'))" << false << QVariant();
+      QTest::newRow( "m_min NULL" ) << "m_min(geom_from_wkt(NULL))" << false << QVariant();
+      QTest::newRow( "m_min point" ) << "m_min(make_point_m(0,0,1))" << false << QVariant( 1.0 );
+      QTest::newRow( "m_min line" ) << "m_min(make_line(make_point_m(0,0,1),make_point_m(-1,-1,2),make_point_m(-2,-2,0)))" << false << QVariant( 0.0 );
```
m-kuhn, line 1202: Can you add a test with `nan` values for z/m?

JanCaha: Sure. 

But the question is, what should the function return when there is only the `nan` value for Z/M? For example this call: `z_max(geom_from_wkt('PointZ (1 1 nan)'))`.
Looking at expression `max()` it returns `QVariant result( QVariant::Double )` if there are no values or only `null`. But another reasonable option seems to be to return `std::numeric_limits<double>::quiet_NaN()`. 

Any ideas which solution is more reasonable?

m-kuhn: I would say go for NULL, this is much better supported by the expression engine.

JanCaha: So returning `QVariant()`, right?

m-kuhn: `QVariant( QVariant::Double )`

JanCaha:
All the suggestions should be incorporated now.

m-kuhn:
There are some code style / indentation issues (see https://travis-ci.org/github/qgis/QGIS/jobs/678631782#L980). Can you run astyle on those files and to simplify this for the future have a look at https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTING.md#contributing-to-qgis

lbartoletti:
> this seems to fill feature request [#29098](https://github.com/qgis/QGIS/issues/29098)
> @lbartoletti you are assigned so I'm mentioning you

No worries, my code was ready but not the doc and tests :)

Good job @JanCaha !

JanCaha:
I am sorry, I tend to forget to run prepare_commit.sh script before commiting. Should be all fixed now.

PR #35932, opened 2020-04-23
Add some helpful scoped RAII classes for use with QgsRenderContext

nyalldawson:
- QgsScopedRenderContextPainterSwap: allows for temporarily swapping the destination painter object for a QgsRenderContext for the lifetime of the object

- QgsScopedRenderContextScaleToMm: temporarily rescales a render context destination painter device to use millimeter based units for the lifetime of the object

- QgsScopedRenderContextScaleToPixels: temporarily rescales a render context (which has been scaled to millimeter based units) back to pixel based units, for the lifetime of the object


nyalldawson:
bah - i've got this backward :stuck_out_tongue:

Issue #33893, opened 2020-01-17
Include Layer Type Icon in Attribute Table Header


Saijin-Naib:
**Feature description.**
A small tweak to include the layer type icon in the header of the attribute table to help disambiguate between similar/same named objects of different types.

A follow-on might be to hide/disable buttons that don't make sense in the given context (spatial tools like zoom/pan/flash to feature for non-geometry tables does not make much sense, for instance).

**Additional context**
A challenge:  
Which table is for the non-spatial table in the database, and which is for the geometry table?

![image](https://user-images.githubusercontent.com/19295950/72628819-1f4a0180-391d-11ea-93bd-881204136b93.png)

PR #35933, opened 2020-04-23
Support for adding dock widgets as tabs: addTabifyDockWidget()

### LINKED ISSUE #11953: opening data tables widens the main window
Author Name: **Paolo Cavallini** (@pcav)
Original Redmine Issue: [1893](https://issues.qgis.org/issues/1893)
Affected QGIS version: 3.6.0
Redmine category:gui

---
When opening >=3 tables while in docked mode, the main window becomes very wide, instead of shrinking the space allocated to each table

gacarrillor:
## Description
This PR has three related objectives: 
 1. **Add support for adding dock widgets as tabs.**
    QGIS has a number of panels that easily clutter dock widget areas making users to manually rearrange them. Some dock widgets seem very related, e.g. `Layers` and `Layer Order` or the dock widgets created when opening attribute tables. QGIS could take advantage of a new `addTabifiedDockWidget()` to 'group' some panels by default.

  2. **Offer such capability to Python plugin developers.**
    Dock widgets are great for displaying complete interfaces that extend QGIS functionality. However, if a plugin uses a dock widget and other dock widgets exist in the same area, chances are users get lost and have to realize that the plugin's GUI is shrinked to distribute the area among all existing dock widgets. Should plugin devs had an option to tabify their dock widgets on top of existing ones, their plugin users would benefit, by avoiding to rearrange GUI controls (after realizing that the GUI is actually there).

 3. **Fix #11953**
    A possible solution to avoid shrinking the `Layers` panel when opening more than 3 attribute tables (as dock widgets), is to group attribute tables into tabs. 

## Screencasts

**Before:**
attribute_table_before

**After:** 
attribute_table_after

## Demo
tabified_dock_widget_demo


src/app/qgisapp.cpp
```diff
@@ -16046,6 +16046,61 @@ void QgisApp::triggerCrashHandler()
...
+        }
+      }
+    }
+
+    if ( !onTopOfDockWidget )
+    {
+      onTopOfDockWidget = dockWidgetsInArea.at( 0 );  // Last resort
+    }
+    tabifyDockWidget( onTopOfDockWidget, dockWidget );
+    qApp->processEvents();  // Give the app a while to refresh controls
```
nyalldawson, line 16082: This is dangerous to call. I'd suggest making the auto focus behaviour a manual step and don't perform it automatically here.

nyalldawson: Note that if you use QgsDockWidget it's easy to do via setUserVisible

src/app/qgisapp.cpp
```diff
@@ -16046,6 +16046,61 @@ void QgisApp::triggerCrashHandler()
...
+
+  addDockWidget( area, dockWidget );  // First add the dock widget, then attempt to tabify
+  if ( dockWidgetsInArea.length() > 0 )
+  {
+    QDockWidget *onTopOfDockWidget = nullptr;
+    if ( !onTopOf.isEmpty() )
+    {
+      for ( QDockWidget *cw : dockWidgetsInArea )
+      {
+        if ( cw->windowTitle() == onTopOf )
```
nyalldawson, line 16069: This should use objectName and not the window title. Otherwise code will break in translated versions

src/gui/qgisinterface.h
```diff
@@ -889,10 +889,22 @@ class GUI_EXPORT QgisInterface : public QObject
...
+    /**
+     * Adds a \a dock widget to the given \a area and tabify it (if other dock widgets
+     * exist in the same \a area). The new tab will be selected.
+     *
+     * If \a onTopOf is given, the \a dock widget will be placed on top of a dock
+     * widget with such name (if it exists).
+     *
+     * \see addDockWidget()
+     */
+    virtual void addTabifiedDockWidget( Qt::DockWidgetArea area, QDockWidget *dockwidget, const QString &onTopOf = QString() ) = 0;
```
nyalldawson, line 906: I wonder if onTopOf should be an list (ordered by priority). It would make this more adaptable to customised dock placement...

src/app/qgisapp.cpp
gacarrillor: Wow! It was easier than I was suggesting, and also effective. Thanks!

src/app/qgisapp.cpp
gacarrillor: Done!

src/gui/qgisinterface.h
gacarrillor: Yes, I'm fine with a list of object names here. 
Implemented in https://github.com/qgis/QGIS/pull/35933/commits/14ba75517310d185c1195b7027d674e9ff4b5889!

src/app/qgisapp.cpp
```diff
@@ -16046,6 +16046,55 @@ void QgisApp::triggerCrashHandler()
...
+      QStringListIterator iter( onTopOf );
+      while ( iter.hasNext() )
+      {
+        const QString &objectName = iter.next();
+        for ( QDockWidget *cw : dockWidgetsInArea )
+        {
+          if ( cw->objectName() == objectName )
+          {
+            onTopOfDockWidget = cw;
+            iter.toBack();  // The task is done, exit the iterator
```
gacarrillor, line 16076: Not sure if that's the best way to break two iterations (i.e., `iter.toBack()` and `break`).
Please let me know if we should do it differently.

gacarrillor:
I just added tests for what I was showing off in the "Demo" screencast.

src/app/qgisapp.cpp
nyalldawson: I'd just use a standard range based for loop instead of QStringListIterator, and have a "bool found" variable in order to break out of the second loop

src/app/qgisapp.cpp
```diff
@@ -16046,6 +16046,55 @@ void QgisApp::triggerCrashHandler()
...
+    }
+
+    if ( !onTopOfDockWidget )
+    {
+      onTopOfDockWidget = dockWidgetsInArea.at( 0 );  // Last resort
+    }
+    tabifyDockWidget( onTopOfDockWidget, dockWidget );
+
+    // Select our newly added tab
+    QgsDockWidget *qdw = dynamic_cast< QgsDockWidget * >( dockWidget );
```
nyalldawson, line 16090: I still think we should leave this to be a manual step done by the caller when needed. I don't think we always want to force raise newly added docks.

gacarrillor: What about a boolean parameter `raiseTab` that is `true` by default?

nyalldawson: I'm just not sure how often we want to encourage developers to raise by default. I'm concerned that this could be (unintentionally) abused by plugins  -- if they call the method without setting raise to false, then on startup a user will constantly see their dock tab order reset with the plugin's docks moving to front...

gacarrillor: Well, it could be `false` by default to avoid that. 
I would just like to let this method be in charge of raising the tab.

If you still think it's not its business, let me know. I'll implement it as it better fits with QGIS as a whole.

nyalldawson: I'm ok if you make it false by default!

gacarrillor: Implemented in https://github.com/qgis/QGIS/pull/35933/commits/3c9a375173a1f529d3af1cdc54d741502ab6366c

src/app/qgisapp.cpp
gacarrillor: The iterator was removed in favor of a basic for loop. Thanks for the suggestion.

nyalldawson (approved):

gacarrillor:
Thank you @nyalldawson for your review. 
Should I add the `Feature` label to this PR?

pigreco:
@gacarrillor 
> Should I add the `Feature` label to this PR?

Yes please
otherwise it will be difficult to trace it

NathanW2:
Nice work on this one. We should look at getting some nice icons for the tab headers if we can in another PR.  I find the tabs get a bit lost when they are tabbed

gacarrillor:
Sounds like a good indicator for users.

Saijin-Naib:
> Nice work on this one. We should look at getting some nice icons for the tab headers if we can in another PR.  I find the tabs get a bit lost when they are tabbed

Similar to #33893?

gacarrillor:
> Nice work on this one. We should look at getting some nice icons for the tab headers if we can in another PR. I find the tabs get a bit lost when they are tabbed

Like this?
image

BTW, were you talking about attribute tables or in general for every tab from a tabBar in QGIS?

PR #35935, opened 2020-04-23
[layouts] Add explicit button to insert expressions in legend text

nyalldawson:
While this has been possible for a while by manually entering a "[% expression %]" string as part of a legend item's text, this is completely hidden from users and rather useless.

Now that we have a dedicated widget to allow configuration of individual legend items, we've got the capacity to show a dedicated "insert expression" button, just like we do for layout labels.

This makes it immediately clear to users that expressions CAN be used in legend item text, and also helps them construct valid expressions using the available expression context.

Peek 2020-04-23 16-25

### DIFF
--- a/src/gui/layout/qgslayoutlegendwidget.cpp
+++ b/src/gui/layout/qgslayoutlegendwidget.cpp
@@ -1434,15 +1434,17 @@ QgsLayoutLegendNodeWidget::QgsLayoutLegendNodeWidget( QgsLayoutItemLegend *legen
     mPatchShapeButton->hide();
   }
 
-  mLabelEdit->setText( currentLabel );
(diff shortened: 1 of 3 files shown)


nyalldawson:
(hands up any qgis users who actually knew this was possible! :stuck_out_tongue: )

nirvn (approved):
Big +1 ~~; a nice bonus is how easy this makes for the user to input multiline label (i.e., without having to rely on the character to wrap option).~~ oups, it's not working :)

nirvn:
Side note: your line symbol there is good looking :)

roya0045:
> (hands up any qgis users who actually knew this was possible! 😛 )

Since we're providing some love to the legend, any plan to finally close the symbol expression saga with the last two PRs 😟

PR #35937, opened 2020-04-23
[Tests][Server] WMS GetPrint group Mask file

rldhont:
To complete PR #35277


nyalldawson:
This is still seeing failures: https://travis-ci.org/github/qgis/QGIS/jobs/678631781

nyalldawson:
https://cdash.orfeo-toolbox.org/testDetails.php?test=18793780&build=39473

rldhont:
Can we compare images and accept some differences ?

In the test, we can deactivate the strict assert between individuals and group image results because the test also contains a control image.

rldhont:
It seems to be a segfault in image checker not in QGIS Server isn't it ?

PR #35939, opened 2020-04-23
check project root name in the server validator

Gustry:
## Description

check project root name in the server validator

### DIFF
--- a/src/core/qgsprojectservervalidator.h
+++ b/src/core/qgsprojectservervalidator.h
@@ -49,7 +49,7 @@ class CORE_EXPORT QgsProjectServerValidator
       LayerShortName = 1, //!< Layer/group short name is not valid.
       LayerEncoding = 2,  //!< Encoding is not correctly set on a vector layer.
       ProjectShortName = 3,  //!< The project short name is not valid.
-      ProjectRootNameConflict = 4,  //!< The project root name isady used by a layer or a group.
+      ProjectRootNameConflict = 4,  //!< The project root name is already used by a layer or a group.
     };
 
     /**
--- a/src/core/qgsprojectservervalidator.h
+++ b/src/core/qgsprojectservervalidator.h
@@ -45,9 +45,11 @@ class CORE_EXPORT QgsProjectServerValidator
      */
     enum ValidationError
     {
-      DuplicatedNames = 0, //!< Error related to a duplicated layer name in the layer tree.
-      ShortNames = 1, //!< Layer short name is well formatted.
-      Encoding = 2  //!< Encoding is not set correctly.
+      DuplicatedNames = 0, //!< A duplicated layer/group name in the layer tree.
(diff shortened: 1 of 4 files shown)


rldhont (approved):

src/core/qgsprojectservervalidator.h
```diff
@@ -45,9 +45,11 @@ class CORE_EXPORT QgsProjectServerValidator
...
     enum ValidationError
     {
-      DuplicatedNames = 0, //!< Error related to a duplicated layer name in the layer tree.
-      ShortNames = 1, //!< Layer short name is well formatted.
-      Encoding = 2  //!< Encoding is not set correctly.
+      DuplicatedNames = 0, //!< A duplicated layer/group name in the layer tree.
+      LayerShortName = 1, //!< Layer/group short name is not valid.
+      LayerEncoding = 2,  //!< Encoding is not correctly set on a vector layer.
+      ProjectShortName = 3,  //!< The project short name is not valid.
+      ProjectRootNameConflict = 4,  //!< The project root name isady used by a layer or a group.
```
elpaso, line 52:
```suggestion
      ProjectRootNameConflict = 4,  //!< The project root name is already used by a layer or a group.
```

elpaso:
I'm wondering: is this for legacy OGC services (W?S) only? 

Did you consider if this checks are also valid for the new generation of OGC services such as OAPIF?
(because I think they doesn't).

Perhaps you should make it clear to what service(s) the warnings apply when printing the validation results.

As a side note, I honestly never found the validator particularly useful, I would rather prefer if there were automatic healing procedures instead (when possible of course), for example by automatically replacing forbidden characters.

Gustry:
Thanks @elpaso for your feedbacks.

> Did you consider if this checks are also valid for the new generation of OGC services such as OAPIF?

Sorry, I haven't played that much with OAPIF yet. We are still using a lot WMS (and WFS1.X when needed) in our projects.

CMIIW, but there is only WFS3 for now for the new generation of OGC services. WMS is still used for publishing maps, so this rule in the validator is still useful.

> Perhaps you should make it clear to what service(s) the warnings apply when printing the validation results.

I think it's a good idea.
I was planning, before the feature freeze, to split the QGIS Server panel in the Project properties. It tends to be very long to scroll down with `Capabilities`, `WMS`, `WMTS`, `WFS`, etc. To do like the labels/metadata panel.
So I think we will have more space to add a text explaining the differences. What do you think?

> As a side note, I honestly never found the validator particularly useful, I would rather prefer if there were automatic healing procedures instead (when possible of course), for example by automatically replacing forbidden characters.

I follow you. I just hit these issues sometimes and I'm just trying to have some warnings when publishing a project. I don't have a strong background in the QGIS Server code base and the OGC standards ;-)

rldhont (approved):

rldhont:
@elpaso any objections  about merge ?

PR #35948, opened 2020-04-23
Fix arguments in QgisApp::about() for exported source code

agiudiceandrea:
## Description

Fixes arguments of QGIS code branch string and link in QgisApp::about() for exported source code.

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -5111,8 +5111,16 @@ void QgisApp::about()
 
     if ( QString( Qgis::devVersion() ) == QLatin1String( "exported" ) )
     {
-      versionString += QStringLiteral( "%1</td><td><a href=\"https://github.com/qgis/QGIS/tree/release-%1_%2\">Release %1.%2</a></td>" )
-                       .arg( tr( "QGIS code branch" ) ).arg( Qgis::versionInt() / 10000 ).arg( Qgis::versionInt() / 100 % 100 );
+      versionString += tr( "QGIS code branch" );
+      if ( Qgis::version().endsWith( QLatin1String( "Master" ) ) )
+      {
+        versionString += QLatin1String( "</td><td><a href=\"https://github.com/qgis/QGIS/tree/master\">master</a></td>" );
+      }
+      else
+      {
+        versionString += QStringLiteral( "</td><td><a href=\"https://github.com/qgis/QGIS/tree/release-%1_%2\">Release %1.%2</a></td>" )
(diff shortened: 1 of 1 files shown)


src/app/qgisapp.cpp
```diff
@@ -5111,8 +5111,15 @@ void QgisApp::about()
 
     if ( QString( Qgis::devVersion() ) == QLatin1String( "exported" ) )
     {
-      versionString += QStringLiteral( "%1</td><td><a href=\"https://github.com/qgis/QGIS/tree/release-%1_%2\">Release %1.%2</a></td>" )
-                       .arg( tr( "QGIS code branch" ) ).arg( Qgis::versionInt() / 10000 ).arg( Qgis::versionInt() / 100 % 100 );
+      if ( Qgis::version().endsWith( QLatin1String( "Master" ) ) )
+      {
+        versionString += QStringLiteral( "%1</td><td><a href=\"https://github.com/qgis/QGIS/tree/master\">master</a></td>" ).arg( tr( "QGIS code branch" ) );
```
rouault, line 5116: the code would be even clearer / simpler if versionString += tr( "QGIS code branch" ) was done before the if()

agiudiceandrea: You are right, thank you for pointing that out. Now it should be better, I think.

agiudiceandrea:
The bug affects 3.10 and 3.12 branches, along with master branch. @rouault do you think it's worth backporting the fix?

rouault:
hum if I've well understood your fix, it adds a special case for "master", so I don't see any benefits for backporting to branches, which by definition are not master. Or did I miss something ?

agiudiceandrea:
The patch was originally intended to only fix the arguments in
```
QStringLiteral( "%1</td><td><a href=\"https://github.com/qgis/QGIS/tree/release-%1_%2\">Release %1.%2</a></td>" )
                       .arg( tr( "QGIS code branch" ) ).arg( Qgis::versionInt() / 10000 ).arg( Qgis::versionInt() / 100 % 100 );
```

as you can see, there are %1 and again %1 and %2, but tree arg(), creating the incorrect string
QGIS code branch | Release QGIS code branch.3
-- | --

with also incorrect link to the code branch on GitHub.

Fixing this, I also noticed that a special case for master branch was needed.

rouault:
> there were %1 and again %1 and %2, but three arg(),

ah ok, missed that ! As I've already merged, I don't think the added labels will work. Can you issue manual backport pull requests ?

rouault:
Nevermind: the backport bot did the job

PR #35954, opened 2020-04-23
remove Server Search tab from the WMS connections dialog (fix #25660)

### LINKED ISSUE #25660: WMS Server Search completely outdated
Author Name: **Bernd Vogelgesang** (Bernd Vogelgesang)
Original Redmine Issue: [17764](https://issues.qgis.org/issues/17764)
Affected QGIS version: 3.7(master)
Redmine category:web_services_clients/wms

---
In the dialog for adding WMS layers, there is a server search tab.
The search is performed trough a service by geopole.org which is run (or not) by sourcepole.
For years it seems, nothing happend to this search mechanism.

Try yourself: Fill in "bayern" into the field for searching layers connected to my home state Bavaria.
The search finds 10 layers. 2 of them are duplicates, so 8 remain.
For these 8 layers, only a single WMS layer is in fact loadable (And of no use at all)

As this service is not usable, it is quite useless, so I would suggest to remove the tab to put an end to this shameful situation.

alexbruy:
## Description
As discussed in https://lists.osgeo.org/pipermail/qgis-developer/2020-January/059941.html.

Remove Server Search tab from the WMS connections dialog as used catalog does not receive updates for years and we already have more feature rich search with MetaSearch core plugin.

Fixes #25660.


nyalldawson:
You can also remove this setting from the options dialog:

image

alexbruy:
Also removed Server Search URL from QGIS settings and Server Search tab from OWS connections dialog.

nyalldawson (approved):

PR #35956, opened 2020-04-23
look for qgis_global_settings.ini also in the AppDataLocation (fix #31288)

### LINKED ISSUE #31288: Allow qgis_global_settings.ini in standard locations
Currently the qgis_global_settings.ini file is searched for by

- App Parameter
- Env Variable
- Install directory

The first two require startup configuration (param, env) or system manipulation (env) whereas the last one is overwritten on updates.

It would be nice to include additional search paths between install and env, based on QStandardPaths. This would allow to provide the files in default locations, managed by the sys admin (or user) which are untouched by the installer and will be taken without any manual application startup configuration.

I'd propose [`QStandardPaths::AppDataLocation`](https://doc.qt.io/qt-5/qstandardpaths.html#StandardLocation-enum), where also startup.py and templates are stored.

alexbruy:
## Description
Allow to put global settings file also in the default location managed by the user or system administrator which is not touched by installer and does not require any additional setup like passing commandline parameters or settings environment variable. Only first found file will be used. On Linux AppData location usually is `$HOME/.local/share/QGIS/QGIS3`, on Windows `%AppData%\Roaming\QGIS\QGIS3`.

With this PR QGIS will look for global settins file (`qgis_global_settings.ini`) in the following places

1. path specified by commandline parameter
2. path defined by environment variable
3. AppData location (see [QStandardPaths](https://doc.qt.io/qt-5/qstandardpaths.html))
4. installation directory

### DIFF
--- a/src/app/main.cpp
+++ b/src/app/main.cpp
@@ -896,10 +896,20 @@ int main( int argc, char *argv[] )
   // Set up the QgsSettings Global Settings:
   // - use the path specified with --globalsettingsfile path,
   // - use the environment if not found
+  // - use the AppDataLocation ($HOME/.local/share/QGIS/QGIS3 on Linux, roaming path on Windows)
   // - use a default location as a fallback
   if ( globalsettingsfile.isEmpty() )
   {
     globalsettingsfile = getenv( "QGIS_GLOBAL_SETTINGS_FILE" );
+  }
+
+  if ( globalsettingsfile.isEmpty() )
+  {
+    QStringList startupPaths = QStandardPaths::locateAll( QStandardPaths::AppDataLocation, QStringLiteral( "qgis_global_settings.ini" ) );
+    if ( !startupPaths.isEmpty() )
+    {
+      globalsettingsfile = startupPaths.at( 0 );
+    }
   }
 
   if ( globalsettingsfile.isEmpty() )


src/app/main.cpp
```diff
@@ -896,12 +896,22 @@ int main( int argc, char *argv[] )
...
   {
     globalsettingsfile = getenv( "QGIS_GLOBAL_SETTINGS_FILE" );
   }
 
+  if ( globalsettingsfile.isEmpty() )
+  {
+    QStringList startupPaths = QStandardPaths::locateAll( QStandardPaths::AppDataLocation, QStringLiteral( "qgis_global_settings.ini" ) );
+    if ( !startupPaths.isEmpty() )
+    {
+      globalsettingsfile = startupPaths.at( 0 );
```
m-kuhn, line 911: Is it possible for the user to see which file has been picked up already?
How about adding a log (info) message here for system debug purpose?

m-kuhn:
Thanks @alexbruy 
I added a link to QStandardPaths to the description message for documentation purpose

src/app/main.cpp
alexbruy: This already handled below in the code. If global settings file found a message will be added to QGIS Message Log in the "QGIS" tab, see https://github.com/qgis/QGIS/blob/master/src/app/main.cpp#L922.

m-kuhn:
Merge away !

m-kuhn (approved):

PR #35968, opened 2020-04-23
Vector tile layer - part 7 (writing to MBTiles)

wonder-sk:

Added support to write vector tiles to MBTiles files. These are convenient because it is just a single file with all tiles stored as gzipped blobs in SQLite database.

Sample code to generate vector tiles to MBTiles:

```
uri = QgsDataSourceUri()
uri.setParam("type", "mbtiles")
uri.setParam("url", "/home/qgis/output.mbtiles")
lyr = QgsVectorTileWriter.Layer(iface.activeLayer())
writer = QgsVectorTileWriter()
writer.setDestinationUri(bytes(uri.encodedUri()).decode('utf-8'))
writer.setLayers([lyr])
writer.setMinZoom(0)
writer.setMaxZoom(3)
# writer.setExtent(QgsRectangle(...))
writer.writeTiles()
```

## Thanks

### DIFF
--- a/src/core/qgsmbtilesreader.cpp
+++ b/src/core/qgsmbtilesreader.cpp
@@ -18,7 +18,10 @@
 #include "qgslogger.h"
 #include "qgsrectangle.h"
 
+#include <QFile>
 #include <QImage>
+
+#include <zlib.h>
 
 
 QgsMBTilesReader::QgsMBTilesReader( const QString &filename )
@@ -46,6 +49,37 @@ bool QgsMBTilesReader::isOpen() const
   return bool( mDatabase );
 }
 
+bool QgsMBTilesReader::create()
+{
+  if ( mDatabase )
+    return false;
+
+  if ( QFile::exists( mFilename ) )
+    return false;
+
+  sqlite3_database_unique_ptr database;
+  int result = mDatabase.open_v2( mFilename, SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE, nullptr );
+  if ( result != SQLITE_OK )
+  {
+    QgsDebugMsg( QStringLiteral( "Can't create MBTiles database: %1" ).arg( database.errorMessage() ) );
+    return false;
+  }
+
+  QString sql = \
+                "CREATE TABLE metadata (name text, value text);" \
+                "CREATE TABLE tiles (zoom_level integer, tile_column integer, tile_row integer, tile_data blob);" \
+                "CREATE UNIQUE INDEX tile_index on tiles (zoom_level, tile_column, tile_row);";
+  QString errorMessage;
+  result = mDatabase.exec( sql, errorMessage );
+  if ( result != SQLITE_OK )
+  {
+    QgsDebugMsg( QStringLiteral( "Failed to initialize MBTiles database: " ) + errorMessage );
+    return false;
+  }
+
+  return true;
+}
+
 QString QgsMBTilesReader::metadataValue( const QString &key )
 {
   if ( !mDatabase )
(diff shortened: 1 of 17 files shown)


wonder-sk:
Few random thoughts:
- maybe we should rename QgsMBTilesReader to just QgsMBTiles when it is not just a reader anymore :-)
- maybe we should have in the writer extra backend specific options - e.g. mbtiles also allows adding metadata like dataset name, description, attribution, version, ...

The next step would be to create a processing algorithm... I am wondering what would be the best approach - whether to have one alg that is completely generic - or have several algs that are easier to use. Things to consider:
- support XYZ url template and MBTiles in a single alg? Or one for each kind of output?
- have one variant for a single vector layer and one variant for multiple vector layers? (the issue is that it's not just a list of layers - each layer may have extra options like filter expression, zoom range and so on). I guess with multiple layers the main issue will be the need to create some kind of custom GUI widget for layer configuration for Processing? Or shall we avoid using auto-generated processing GUI and create a custom dialog for the configuration?

src/core/qgsmbtilesreader.cpp
```diff
@@ -46,6 +49,37 @@ bool QgsMBTilesReader::isOpen() const
   return bool( mDatabase );
 }
 
+bool QgsMBTilesReader::create()
```
nyalldawson, line 52: I missed this on earlier reviews, but "MB" should be "Mb" for consistency with the rest of the API

src/core/qgsmbtilesreader.cpp
```diff
@@ -122,3 +180,127 @@ QImage QgsMBTilesReader::tileDataAsImage( int z, int x, int y )
...
+  sqlite3_bind_blob( preparedStatement.get(), 1, data.constData(), data.size(), SQLITE_TRANSIENT );
+
+  if ( preparedStatement.step() != SQLITE_DONE )
+  {
+    QgsDebugMsg( QStringLiteral( "MBTile tile failed to be set: %1,%2,%3" ).arg( z ).arg( x ).arg( y ) );
+    return;
+  }
+}
+
+bool QgsMBTilesReader::decodeGzip( const QByteArray &bytesIn, QByteArray &bytesOut )
```
nyalldawson, line 210: @rouault I'm guessing there's nothing in gdal c api we could reuse here?

src/core/vectortile/qgsvectortilewriter.cpp
```diff
@@ -115,32 +149,85 @@ bool QgsVectorTileWriter::writeTiles( QgsFeedback *feedback )
...
+  {
+    QgsVectorLayer *vl = layer.layer();
+    const QgsFields fields = vl->fields();
+
+    QJsonObject fieldsObj;
+    for ( const QgsField &field : fields )
+    {
+      QString fieldTypeStr;
+      if ( field.type() == QVariant::Bool )
+        fieldTypeStr = "Boolean";
```
nyalldawson, line 212: QStringLiterals

src/core/vectortile/qgsvectortilewriter.cpp
```diff
@@ -115,32 +149,85 @@ bool QgsVectorTileWriter::writeTiles( QgsFeedback *feedback )
...
+  }
+
+  f.write( tileData );
+  f.close();
+  return true;
+}
+
+QString QgsVectorTileWriter::mbtilesJsonSchema()
+{
+  QJsonArray arrayLayers;
```
nyalldawson, line 201: I think the general preference in the code seems to be moving away from QJson -> nlohmann. You could replace this with QgsJsonUtils::jsonFromVariant , e.g.

    return QString::fromStdString( QgsJsonUtils::jsonFromVariant( value ).dump( 2 ) );

nyalldawson:
> The next step would be to create a processing algorithm... I am wondering what would be the best approach - whether to have one alg that is completely generic - or have several algs that are easier to use. Things to consider:

My preference would be to go for several algorithms which can be specialised to the corresponding format. Otherwise we end up with confusing "grass style" algorithms with a ton of mutually exclusive options/options which only apply when other combinations of settings are applied. It's not a user-friendly result.

> I guess with multiple layers the main issue will be the need to create some kind of custom GUI widget for layer configuration for Processing

That's totally possible to do, and what I'd recommend! Just **don't** inherit from the Python WidgetWrapper class and use the c++ stable API instead.

nyalldawson:
Lgtm!

PR #35971, opened 2020-04-24
Respect newline characters when rendering layout legends

nyalldawson:
If a newline character is entered as part of a legend item's text, respect this while rendering the legend


nirvn (approved):
Nice.

PR #35974, opened 2020-04-24
Allow control over the horizontal spacing before legend group/subgroup/symbols

nyalldawson:
Gives flexibility to allow "nesting" legend groups/subgroups/symbols and much greater control over legend item placement

Sponsored by SLYR


nirvn (approved):

roya0045:
I'm a little late to the party but I was curious if you think this could possibly affect right-to-left text.

Question regarding good practice: Should we also make tests with right-to-left tests as good practice with the legend or should we assume that no error will occur until a ticket is submitted?

nyalldawson:
I did test that already, and there's a unit test here checking that the spacing works correctly on right aligned text.

@nirvn and I debated whether to expose seperate controls for left and right spacing, but decided that this would complicate the UI too much so deferred that until there's a proven need for that level of control.

roya0045:
@nyalldawson  I saw the image but I was not sure if language such as arabic are handled the same as english in this case. I guess it shouldn't have an effect.

Issue #35970, opened 2020-04-23
Processing warning: Duplicate output OUTPUT registered for alg contour_polygon


agiudiceandrea:
**Describe the bug**
In the Processing Log Messages is present the following wanring just after QGIS starts.
`WARNING    Duplicate output OUTPUT registered for alg contour_polygon`
The same warning is added to the log when the GDAL processing algorithm "Contour Polygons" (gdal:contour_polygon) is opened.

**QGIS and OS versions**

QGIS Master (97814517da) on Windows. Probably since #34636.

agiudiceandrea:
Fixed by #35976

PR #35976, opened 2020-04-24
[processing] When removing a parameter registered as an output, remove it from outputs list

nirvn:
## Description

Until now, when removing a processing parameter that's also registered as an output, the parameter definition was removed but its linked output was not. This PR fixes that.

Practically, the PR also gets rid of the following warning thrown every time QGIS is launched:
`WARNING    Duplicate output OUTPUT registered for alg contour_polygon`

### DIFF
--- a/src/core/processing/qgsprocessingalgorithm.cpp
+++ b/src/core/processing/qgsprocessingalgorithm.cpp
@@ -329,6 +329,12 @@ void QgsProcessingAlgorithm::removeParameter( const QString &name )
     delete def;
     mParameters.removeAll( def );
   }
+  const QgsProcessingOutputDefinition *outputDef = outputDefinition( name );
+  if ( outputDef )
+  {
+    delete outputDef;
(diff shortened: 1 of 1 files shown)


alexbruy (approved):

nyalldawson:
I don't think we should do this -- it's potentially dangerous and undocumented, which makes it a future source of frustration! I'd rather see the warning fixed by just manually removing the extra output in the offending algorithm alongside the parameter.

If you HAVE to go with the automatic removal, it should only be done if the parameter had the auto-create-corresponding-output setting set on it at creation time. And it would need a unit test :P

nirvn:
I think we need to implement some form of auto removal, we can limit that to auto created one. It's otherwise behaving against expectations (ie if I create an output via parameter addition, I expect that output to go when parameter is removed) I'll do a follow up PR.

agiudiceandrea:
This fixes #35970

PR #35991, opened 2020-04-24
Group postgres DELETE statements to a single query

### LINKED ISSUE #35381: Group statements when deleting from PostgreSQL
**Feature description.**

Deleting 40000 features from a PostgreSQL layer results in 40000 different `delete from ... where id = ...` statements. On remote connections this causes 40000 times the overhead and connection latency.
A single `delete from ... where id in (... ...)` would probably be much faster for remote and slow connections.
**Additional context**

uclaros:
...instead of one for each deleted feature
Deleting lots of features over remote connections should be faster.
Is there a query length limitation we should consider? I successfully ~~test deleted 10m features.~~ realized I'm tired... I tested with 1k though.

Fixes #35381

### DIFF
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -2536,6 +2536,9 @@ bool QgsPostgresProvider::addFeatures( QgsFeatureList &flist, Flags flags )
 
 bool QgsPostgresProvider::deleteFeatures( const QgsFeatureIds &ids )
 {
+  if ( ids.isEmpty() )
+    return true;
+
   bool returnvalue = true;
 
   if ( mIsQuery )
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -541,6 +541,11 @@ QString QgsPostgresProvider::whereClause( QgsFeatureId featureId ) const
   return QgsPostgresUtils::whereClause( featureId, mAttributeFields, connectionRO(), mPrimaryKeyType, mPrimaryKeyAttrs, mShared );
 }
 
+QString QgsPostgresProvider::whereClause( QgsFeatureIds featureIds ) const
+{
+  return QgsPostgresUtils::whereClause( featureIds, mAttributeFields, connectionRO(), mPrimaryKeyType, mPrimaryKeyAttrs, mShared );
+}
+
 
 QString QgsPostgresUtils::whereClause( QgsFeatureId featureId, const QgsFields &fields, QgsPostgresConn *conn, QgsPostgresPrimaryKeyType pkType, const QList<int> &pkAttrs, const std::shared_ptr<QgsPostgresSharedData> &sharedData )
 {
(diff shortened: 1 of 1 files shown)


espinafre:
@uclaros , hi, your code seems to be missing the following snippet: https://github.com/qgis/QGIS/blob/81911c1afcb092cf35ed1a675fd9e56d082c1ee2/src/providers/postgres/qgspostgresprovider.cpp#L566-L581 which was recently merged into master, and changes how int64 primary keys are treated.

uclaros:
@espinafre thanks! I did a fetch-merge-rebase-forcepush

espinafre:
Maybe checking if ids is not empty on src/providers/postgres/qgspostgresprovider.cpp:2558... if not, whereClause will generate something like "DELETE FROM table WHERE pk IN ()", which is invalid SQL syntax.

espinafre:
@uclaros , any luck pestering someone with merge privileges? Maybe @nyalldawson ? I'd love to see this in the next 3.12 release.

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -2550,17 +2558,17 @@ bool QgsPostgresProvider::deleteFeatures( const QgsFeatureIds &id )
   {
     conn->begin();
 
-    for ( QgsFeatureIds::const_iterator it = id.begin(); it != id.end(); ++it )
-    {
-      QString sql = QStringLiteral( "DELETE FROM %1 WHERE %2" )
-                    .arg( mQuery, whereClause( *it ) );
-      QgsDebugMsgLevel( "delete sql: " + sql, 2 );
+    QString sql = QStringLiteral( "DELETE FROM %1 WHERE %2" )
```
elpaso, line 2561:
```suggestion
    const QString sql = QStringLiteral( "DELETE FROM %1 WHERE %2" )
```

elpaso (approved):
LGTM, I assume that the base provider tests cover this path (I'm 99.99% sure they do but I didn't really check).

uclaros:
There's a potential issue with this: when trying to delete 10 million features the sql string got so big that postgres gave an out of memory error `ERROR:  invalid memory alloc request size 1073741824`. 
However, QGIS gets very unresponsive trying to select and delete 1 million features, let alone 10 million so it's not that probable a case. Would it make sense to make a compromise and break the request to smaller chunks, like one request for each 100,000 features? If yes, what would be a reasonable compromise of request size/number of requests?

elpaso:
@uclaros I agree that this is an unlikely scenario, but since QGIS algorithms can be run in headless mode I think it's worth setting an upper limit to the number of fids and chunk them when they are more.
I've no idea about what could be a reasonable value though. 
@strk , any idea?

andreasneumann:
I believe FME has a default bulk size of 5000 features, but it is configurable. And I am not sure if this is for reading only, or also for deleting.

strk:
@elpaso how about rounding it up to "below" 16777216 (2^24) bytes ?

uclaros:
@strk I think the generated sql has a where clause `"id" in ( pkvalue, pkvalue ...)` so the sql string size depends on both the number of features and the character length of key values.

nyalldawson:
@elpaso are you happy with the new approach?

elpaso (approved):

uclaros:
Maybe put the magic number 5000 in a const or setting?

m-kuhn:
Thanks, it's fine to ship it like this for now.
If you think it's worth adding a setting for the magic number, consider the same for the feature iterator (see postgresfeatureiterator.h --> mFeatureQueue size is hardcoded to 2000)

PR #35992, opened 2020-04-24
Use QgsTextRenderer to render legend text

nyalldawson:
Allows for legend text with buffers, shadows, etc!

Draft only for now, just to see what impact it has on existing tests...


nyalldawson:
Will revisit for 3.16

Issue #35993, opened 2020-04-25
Master password credential dialog freezes QGIS when expanding browser node


nirvn:
While I've spotted this while attempting to expand an ArcGIS Feature Service node in the browser, I assume this is a broader issue.

*Steps to reproduce*
0. Make sure your authentification manager has a master password set that's *not* stored in your OS's keyring/wallet.
1. Add a new ArcGIS Feature Service connection that requires an authentication token
2. Restart QGIS to close the authentification manager session
3. Open the browser panel, try to expand the newly saved AFS connection
4. You'll have the credential dialog pop up asking for the master password
5. Enter the master password, click OK, *boom* QGIS freezes

*More information*
I've dissected what's happening to the point where QGIS hits a locked mutex.

When hitting OK on the credential dialog, it calls `QgsApplication::authManager()->verifyMasterPassword( leMasterPass->text() )`. In that `verifyMasterPassword()` function, there's a call to `authDatabaseConnection()` via `masterPasswordRowsInDb()`. 

QGIS freezes when hitting the mutex in `authDatabaseConnection()` on this line: https://github.com/qgis/QGIS/blob/c76c3904050ae3660ee42435380ae479958023bd/src/core/auth/qgsauthmanager.cpp#L114

PR #35994, opened 2020-04-25
[authentication manager] Avoid crossing mutex wires when checking for master password

nirvn:
## Description

This PR fixes issue #35993 . We were both locking the authentication manager's mutex in a function while trying to reach another function locked by the same mutex.

@nyalldawson , I'm really not that intimidate with the code, review much appreciated.

This likely created all sorts of problem for people with master passwords that were not locked in the system's keyring/wallet.

### DIFF
--- a/src/core/auth/qgsauthmanager.cpp
+++ b/src/core/auth/qgsauthmanager.cpp
@@ -99,6 +99,7 @@ QgsAuthManager *QgsAuthManager::instance()
 QgsAuthManager::QgsAuthManager()
 {
   mMutex = new QMutex( QMutex::Recursive );
+  mMasterPasswordMutex = new QMutex( QMutex::Recursive );
   connect( this, &QgsAuthManager::messageOut,
            this, &QgsAuthManager::writeToConsole );
 }
@@ -487,7 +488,7 @@ const QString QgsAuthManager::disabledMessage() const
 
 bool QgsAuthManager::setMasterPassword( bool verify )
 {
-  QMutexLocker locker( mMutex );
+  QMutexLocker locker( mMasterPasswordMutex );
   if ( isDisabled() )
     return false;
 
@@ -497,13 +498,11 @@ bool QgsAuthManager::setMasterPassword( bool verify )
   if ( mMasterPass.isEmpty() )
   {
     QgsDebugMsg( QStringLiteral( "Master password is not yet set by user" ) );
-    locker.unlock();
     if ( !masterPasswordInput() )
     {
       QgsDebugMsg( QStringLiteral( "Master password input canceled by user" ) );
       return false;
     }
-    locker.relock();
   }
   else
   {
--- a/src/core/auth/qgsauthmanager.cpp
+++ b/src/core/auth/qgsauthmanager.cpp
@@ -497,11 +497,13 @@ bool QgsAuthManager::setMasterPassword( bool verify )
   if ( mMasterPass.isEmpty() )
   {
     QgsDebugMsg( QStringLiteral( "Master password is not yet set by user" ) );
+    locker.unlock();
     if ( !masterPasswordInput() )
     {
       QgsDebugMsg( QStringLiteral( "Master password input canceled by user" ) );
       return false;
     }
+    locker.relock();
   }
   else
   {
@@ -1200,12 +1202,13 @@ bool QgsAuthManager::updateAuthenticationConfig( const QgsAuthMethodConfig &conf
 
 bool QgsAuthManager::loadAuthenticationConfig( const QString &authcfg, QgsAuthMethodConfig &mconfig, bool full )
 {
-  QMutexLocker locker( mMutex );
   if ( isDisabled() )
     return false;
 
   if ( full && !setMasterPassword( true ) )
     return false;
+
+  QMutexLocker locker( mMutex );
 
   QSqlQuery query( authDatabaseConnection() );
   if ( full )
--- a/src/core/auth/qgsauthmanager.h
+++ b/src/core/auth/qgsauthmanager.h
@@ -863,6 +863,7 @@ class CORE_EXPORT QgsAuthManager : public QObject
     bool mScheduledDbEraseRequestEmitted = false;
     int mScheduledDbEraseRequestCount = 0;
     QMutex *mMutex = nullptr;
+    QMutex *mMasterPasswordMutex = nullptr;
 
 #ifndef QT_NO_SSL
     // mapping of sha1 digest and cert source and cert


nirvn:
For the reviewer, here's the chain of events that lead to QGIS freezing forever, awaiting a mutex unlock that'll never happen.

- When expanding a ArcGIS Feature Service connection associated to a token (again, it'd work on other connection types relying on a saved authentication method too), QgsArcGisRestUtils::queryService sends a QgsBlockingNetworkRequest in which QgsApplication::authManager()->updateNetworkRequest is called (this is happening off the main thread)
- In turn, QgsAuthManager::updateNetworkRequest finds the authentication method (in this case, ESRI token), and call QgsAuthEsriTokenMethod::updateNetworkRequest, which in turn calls QgsAuthEsriTokenMethod::getMethodConfig (still off the main thread)
- That last method, getMethodConfig, makes a call back to the authentication manager, QgsApplication::authManager()->loadAuthenticationConfig , which locks the authentication manager's mutex (still of the main thread)

That mutex lock happens too early and is problematic. The loadAuthenticationConfig() function checks whether the master password has been provided already via setMasterPassword(), which in turn will trigger a credential dialog requesting the password in QgsAuthManager::masterPasswordInput(). 

When the user enters his/her password in the credential dialog, a check is done against the authentication manager by calling QgsApplication::authManager()->verifyMasterPassword(). That function _requires_ for the mutex to be unlocked to go forward.

So in the above case, we've locked the mutex off the main thread, then while the mutex is locked launched the credential dialog to get the master password, which will end up hitting a locked mutex and wait forever.

@elpaso , you might be in the best position to review, it'd be most welcome.

nirvn:
Little python script to get QGIS to freeze on locked mutex:

```
import threading

def test():
    config = QgsAuthMethodConfig()
    QgsApplication.authManager().loadAuthenticationConfig('zzz',config,True)

thread = threading.Thread(target=test, args=())
thread.daemon = True
thread.start()
```

You have to make sure the authentication manager's master password is *not* saved in your system's keyring/wallet, otherwise the credential dialog will not pop up.

elpaso:
> Little python script to get QGIS to freeze on locked mutex:
> 
> ```
> import threading
> 
> def test():
>     config = QgsAuthMethodConfig()
>     QgsApplication.authManager().loadAuthenticationConfig('zzz',config,True)
> 
> thread = threading.Thread(target=test, args=())
> thread.daemon = True
> thread.start()
> ```
> 
> You have to make sure the authentication manager's master password is _not_ saved in your system's keyring/wallet, otherwise the credential dialog will not pop up.

Hi it's been a long time I didn't touch that code, IIRC @nyalldawson did some changes in recent times regarding mutex lockers.

In any event, I'm not totally sure about your proposed changes, what I can tell you for sure is that on of the most common issues that we had in the past was with WMS tiles (especially with multiple layers accessing to different authenticated endpoints) where the WMS provider attempts parallel downloads triggering concurrent credentials requests by QNAM.

I'll leave other comments in the code.

src/core/auth/qgsauthmanager.cpp
```diff
@@ -1200,13 +1202,14 @@ bool QgsAuthManager::updateAuthenticationConfig( const QgsAuthMethodConfig &conf
...
 bool QgsAuthManager::loadAuthenticationConfig( const QString &authcfg, QgsAuthMethodConfig &mconfig, bool full )
 {
-  QMutexLocker locker( mMutex );
   if ( isDisabled() )
     return false;
 
   if ( full && !setMasterPassword( true ) )
     return false;
 
+  QMutexLocker locker( mMutex );
```
elpaso, line 1211: This looks fine to me (unless we decide to move the lockers down one level), and I would say that it prevents a deadlock because setMasterPassword locks the mutext again.

Did you check if this change is enough to fix your issue?

src/core/auth/qgsauthmanager.cpp
```diff
@@ -497,11 +497,13 @@ bool QgsAuthManager::setMasterPassword( bool verify )
   if ( mMasterPass.isEmpty() )
   {
     QgsDebugMsg( QStringLiteral( "Master password is not yet set by user" ) );
+    locker.unlock();
```
elpaso, line 500: I think masterPasswordInput() must be protected from concurrent accesses. Maybe the best solution would be to move the lockers down one level into the methods but I understand that it would be a major rewrite of the locking system in the manager.

src/core/auth/qgsauthmanager.cpp
nirvn: Yep, this PR fixes the issue I stumbled on expanding connections in the browser panel, as well as the above-pasted mini script.

src/core/auth/qgsauthmanager.cpp
nirvn: @elpaso , a lock could take place inside qgscredential, but I can't see a way it can be done in the authentication manager itself that'd work across threads.

In any case, I wouldn't undertake that unless there's an actual bug to fix. ATM, this PR fixes a QGIS freeze/crash and doesn't come with regressions on travis / azure nor my workflows here.

nirvn:
@elpaso , are you OK with merging this freeze/crash fix as is for now?

src/core/auth/qgsauthmanager.cpp
elpaso: Sorry I wasn't clear, my question was if this change **alone** is sufficient to fix your issue without the need of the other change.

src/core/auth/qgsauthmanager.cpp
nirvn: @elpaso , or, there could be a master password mutex, independent of the overall authentication manager mutex, to prevent multiple master password credential requests.

src/core/auth/qgsauthmanager.cpp
nirvn: @elpaso , ah, let me test this out. I doubt, but who knows.

nirvn: @elpaso , nope, it's not sufficient.

saberraz:
Possible related issue we are having with our plugin: https://github.com/lutraconsulting/qgis-mergin-plugin/issues/118

nirvn:
@elpaso , I've implemented a dedicated mutex for the QgsAuthManager::setMasterPassword function, which will prevent multiple credential dialog from popping while allowing the credential process to access needed authentication manager.

nyalldawson:
@saberraz I think your issue is the opposite - you have

nyalldawson:
Oops wrong button

elpaso (approved):
LGTM now!

nirvn:
@elpaso , woupidou, thanks for the review.

PR #9735, opened 2019-04-07
postgres provider

jef-n:

PR #35995, opened 2020-04-25
Save features into PostgreSQL 12+ tables with GENERATED columns.

### LINKED ISSUE #32898: Save fails on PostGIS tables containing a generated ("GENERATED AS") column
**Describe the bug**
When a PostGIS layer contains a generated column, on a PostgreSQL 12 server, saving edits to the table fails.

**How to Reproduce**
1. In a clean PostGIS database on PostgreSQL 12 ( I use a docker container with [PostgreSQL12.0 and PostGIS 3.0](https://hub.docker.com/layers/kartoza/postgis/12.0/images/sha256-3e7a5c46a1b496eab21d0867df6ac92d0d5c542f94e4505dcf729fa6ea9aa539) ), create a new spatial table with a generated column:
```SQL
CREATE TABLE test_gen_col (
  id SERIAL PRIMARY KEY,
  geom GEOMETRY('Polygon', 4326) NOT NULL,
  cent GEOMETRY('Point') GENERATED AS ( ST_Centroid(geom) ) STORED
);
```
2. In QGIS Desktop, add the new table as a PostGIS layer
3. toggle editing
4. create a new polygon
5. save edits
6. see "Commit error"

image

**QGIS and OS versions**
QGIS 3.10 on Windows 10

**Additional context**
In the example provided, the generated column uses a spatial function, but this problem also exists for generated functions that refer to any other column in the table.

espinafre:
Fixes #32898 . GENERATED columns is a feature introduced by PostgreSQL
12, which allows column values to be generated from other columns in the
same table; this replaces the creation of triggers to, for instance,
populate a column of centroids for the stored polygons.

Quoting the PostgreSQL 12 [documentation](https://www.postgresql.org/docs/12/ddl-generated-columns.html), "A generated column cannot be written to directly. In INSERT or UPDATE commands, a value cannot be specified for a generated column, but the keyword DEFAULT may be specified."

With this commit, QGIS would use the keyword "DEFAULT". The minor drawback that I see is that users will see the string "DEFAULT" on the form field for GENERATED columns when inserting/editing a new feature, instead of the expression used to generate the table; maybe a TODO? The original expression is certainly fetched by the current code.

I've used the same table as the original reporter of the issue to test this fix; it works for me:

```sql
CREATE TABLE test_gen_col (
  id SERIAL PRIMARY KEY,
  geom GEOMETRY('Polygon', 4326) NOT NULL,
  cent GEOMETRY('Point') GENERATED ALWAYS AS ( st_centroid(geom) ) STORED
);
```

Should this be backported to 3.12 and 3.10?

### DIFF
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -920,7 +920,7 @@ bool QgsPostgresProvider::loadFields()
         notNullMap[attrelid][attnum] = attNotNull;
         uniqueMap[attrelid][attnum] = uniqueConstraint;
         identityMap[attrelid][attnum] = attIdentity.isEmpty() ? " " : attIdentity;
-        generatedMap[attrelid][attnum] = attGenerated == "s" ? defVal : "";
+        generatedMap[attrelid][attnum] = attGenerated.isEmpty() ? "" : defVal;
       }
     }
   }
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -920,7 +920,7 @@ bool QgsPostgresProvider::loadFields()
         notNullMap[attrelid][attnum] = attNotNull;
         uniqueMap[attrelid][attnum] = uniqueConstraint;
         identityMap[attrelid][attnum] = attIdentity.isEmpty() ? " " : attIdentity;
-        generatedMap[attrelid][attnum] = attGenerated;
+        generatedMap[attrelid][attnum] = attGenerated == "s" ? defVal : "";
       }
     }
   }
@@ -2077,10 +2077,14 @@ QString QgsPostgresProvider::defaultValueClause( int fieldId ) const
 
   // with generated columns (PostgreSQL 12+), the provider will ALWAYS evaluate the default values.
   // The only acceptable value for such columns on INSERT or UPDATE clauses is the keyword "DEFAULT".
+  // Here, we return the expression used to generate the field value, so the
+  // user can see what is happening when inserting a new feature.
+  // On inserting a new feature or updating a generated field, this is
+  // ommited from the generated queries.
   // See https://www.postgresql.org/docs/12/ddl-generated-columns.html
   if ( !genVal.isEmpty() )
   {
-    return "DEFAULT";
+    return defVal;
   }
 
   if ( !providerProperty( EvaluateDefaultValues, false ).toBool() && !defVal.isEmpty() )
@@ -2334,6 +2338,13 @@ bool QgsPostgresProvider::addFeatures( QgsFeatureList &flist, Flags flags )
         continue;
 
       QString fieldname = mAttributeFields.at( idx ).name();
+
+      if ( mGeneratedValues.contains( idx ) && !mGeneratedValues.value( idx, QString() ).isEmpty() )
+      {
+        QgsDebugMsg( QStringLiteral( "Skipping field %1 (idx %2) which is GENERATED." ).arg( fieldname, idx ) );
+        continue;
+      }
+
       QString fieldTypeName = mAttributeFields.at( idx ).typeName();
 
       QgsDebugMsgLevel( "Checking field against: " + fieldname, 2 );
@@ -2937,6 +2948,7 @@ bool QgsPostgresProvider::changeAttributeValues( const QgsChangedAttributesMap &
 
       // cycle through the changed attributes of the feature
       QString delim;
+      int numChangedFields = 0;
       for ( QgsAttributeMap::const_iterator siter = attrs.constBegin(); siter != attrs.constEnd(); ++siter )
       {
         try
@@ -2944,6 +2956,14 @@ bool QgsPostgresProvider::changeAttributeValues( const QgsChangedAttributesMap &
           QgsField fld = field( siter.key() );
 
           pkChanged = pkChanged || mPrimaryKeyAttrs.contains( siter.key() );
+
+          if ( mGeneratedValues.contains( siter.key() ) && !mGeneratedValues.value( siter.key(), QString() ).isEmpty() )
+          {
+            pushError( tr( "Changing the value of GENERATED field %1 is not allowed." ).arg( fld.name() ) );
+            continue;
+          }
+
+          numChangedFields++;
 
           sql += delim + QStringLiteral( "%1=" ).arg( quotedIdentifier( fld.name() ) );
           delim = ',';
@@ -2986,9 +3006,14 @@ bool QgsPostgresProvider::changeAttributeValues( const QgsChangedAttributesMap &
 
       sql += QStringLiteral( " WHERE %1" ).arg( whereClause( fid ) );
 
-      QgsPostgresResult result( conn->PQexec( sql ) );
-      if ( result.PQresultStatus() != PGRES_COMMAND_OK && result.PQresultStatus() != PGRES_TUPLES_OK )
-        throw PGException( result );
+      // Don't try to UPDATE an empty set of values (might happen if the table only has GENERATED fields,
+      // or if the user only changed GENERATED fields in the form/attribute table.
+      if ( numChangedFields > 0 )
+      {
+        QgsPostgresResult result( conn->PQexec( sql ) );
+        if ( result.PQresultStatus() != PGRES_COMMAND_OK && result.PQresultStatus() != PGRES_TUPLES_OK )
+          throw PGException( result );
+      }
 
       // update feature id map if key was changed
       if ( pkChanged && mPrimaryKeyType == PktFidMap )
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -2341,7 +2341,7 @@ bool QgsPostgresProvider::addFeatures( QgsFeatureList &flist, Flags flags )
 
       if ( !mGeneratedValues.value( idx, QString() ).isEmpty() )
       {
-        QgsDebugMsg( QStringLiteral( "Skipping field %1 (idx %2) which is GENERATED." ).arg( fieldname, idx ) );
+        QgsDebugMsg( QStringLiteral( "Skipping field %1 (idx %2) which is GENERATED." ).arg( fieldname, QString::number( idx ) ) );
         continue;
       }
 
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -2339,7 +2339,7 @@ bool QgsPostgresProvider::addFeatures( QgsFeatureList &flist, Flags flags )
 
       QString fieldname = mAttributeFields.at( idx ).name();
 
-      if ( mGeneratedValues.contains( idx ) && !mGeneratedValues.value( idx, QString() ).isEmpty() )
+      if ( !mGeneratedValues.value( idx, QString() ).isEmpty() )
       {
         QgsDebugMsg( QStringLiteral( "Skipping field %1 (idx %2) which is GENERATED." ).arg( fieldname, idx ) );
         continue;
@@ -2371,12 +2371,7 @@ bool QgsPostgresProvider::addFeatures( QgsFeatureList &flist, Flags flags )
 
       if ( i == flist.size() )
       {
-        // generated values override everything.
-        if ( !mGeneratedValues[idx].isEmpty() )
-        {
-          values += delim + defVal;
-        }
-        else if ( qgsVariantEqual( v, defVal ) )
+        if ( qgsVariantEqual( v, defVal ) )
         {
           if ( defVal.isNull() )
           {
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -889,9 +889,9 @@ bool QgsPostgresProvider::loadFields()
               " LEFT OUTER JOIN ( SELECT DISTINCT indrelid, indkey, indisunique FROM pg_index WHERE indisunique ) uniq ON attrelid=indrelid AND attnum::text=indkey::text "
 
               " WHERE attrelid IN %3"
-            ).arg( connectionRO()->pgVersion() >= 100000 ? QStringLiteral( ", attidentity" ) : QString() )
-            .arg( connectionRO()->pgVersion() >= 120000 ? QStringLiteral( ", attgenerated" ) : QString() )
-            .arg( tableoidsFilter );
+            ).arg( connectionRO()->pgVersion() >= 100000 ? QStringLiteral( ", attidentity" ) : QString(),
+                   connectionRO()->pgVersion() >= 120000 ? QStringLiteral( ", attgenerated" ) : QString(),
+                   tableoidsFilter );
 
       QgsPostgresResult fmtFieldTypeResult( connectionRO()->PQexec( sql ) );
       for ( int i = 0; i < fmtFieldTypeResult.PQntuples(); ++i )
@@ -2341,7 +2341,7 @@ bool QgsPostgresProvider::addFeatures( QgsFeatureList &flist, Flags flags )
 
       if ( !mGeneratedValues.value( idx, QString() ).isEmpty() )
       {
-        QgsDebugMsg( QStringLiteral( "Skipping field %1 (idx %2) which is GENERATED." ).arg( fieldname, idx ) );
+        QgsLogger::warning( tr( "Skipping field %1 which is GENERATED." ).arg( fieldname ) );
         continue;
       }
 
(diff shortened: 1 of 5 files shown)


espinafre:
The tests run correctly on my local system (PostgreSQL 12.0); since Travis seems to use PostgreSQL 11, I'm setting up a new environment with that version and test to figure what went wrong.

espinafre:
@jef-n , since you worked on handling identities (#9735), can you take a look at this, as it uses some of your ideas? Also, for some reason the github interface doesn't allow me to request a review; would you please step in? Thank you very much.

espinafre:
Is it possible to manually trigger the Azure build again?

espinafre:
WIP because the local test with PostgreSQL 12 fails (this won't be caught by current Travis config, which still uses PostgreSQL 11).

espinafre:
All green here. @nyalldawson , @m-kuhn , @jef-n , could you please take a look? Is there a way for the CI systems to check the changes I've made on the testing scripts to support PostgreSQL 12+-only features? Also, any chance to get this on next 3.12 release? Thank you all very much.

espinafre:
Hi, @nyalldawson , what do you think of this?

nyalldawson:
I'm not the best person to review this sorry - maybe @m-kuhn or @elpaso could take a look?

m-kuhn:
Thanks for looking into this. GENERATED columns are a really nice concept.

I wonder if this already handles updates? I see it only takes care of the `addFeatures()` method.
In general, could these values just be skipped everywhere (rather than using the `DEFAULT` keyword)? This would (slightly) reduce network I/O.

This way, the default value would be completely free to be used for something else than a static `DEFAULT` value and could instead be used to communicate to the user what calculation will be performed.

Ideally the field would also be set to read only for the attribute forms, but I don't think there's an API for that already.

espinafre:
Hi, @m-kuhn , it does handle updates, the trick is on the defaultValueClause() method, we use the DEFAULT keyword in all cases. 
The PostgreSQL docs says we must either use DEFAULT or omit the column, which I didn't find a way to do.

m-kuhn:
Let's say you have

```sql
CREATE TABLE test_gen_col (
  id SERIAL PRIMARY KEY,
  geom GEOMETRY('Polygon', 4326) NOT NULL,
  x integer GENERATED ALWAYS AS ( st_x(geom) ) STORED
);
```

```
feature = next(layer.getFeatures())
feature.setGeometry(QgsGeometry.fromWkt('Point(1 2)'))
feature['x'] = 5
layer.updateFeature(feature)
```

Is `DEFAULT` used? I might be missing something, but I thought the defaultValueClause() is mostly used by the attribute form to initialize new features.

espinafre:
Yes, DEFAULT is not used on UPDATE, of course. The PostgreSQL definition says, however, that the column must not be manually written in any way, either by ommiting it or by using the DEFAULT keyword. For INSERTs (addFeatures()), the defaultValueClause() generates the keyword DEFAULT; for UPDATEs (changeAttributeValues()), nothing is changed.

 Changing your test table definition a little (st_x/st_y applies to point):

```sql
CREATE TABLE test_gen_col3 (
  id SERIAL PRIMARY KEY,
  geom GEOMETRY('Point', 4326) NOT NULL,
  x integer GENERATED ALWAYS AS ( st_x(geom) ) STORED
);
```

Inserting a point generates the following:

```sql
INSERT INTO "public"."test_gen_col3" ("geom
","x") VALUES (st_geomfromwkb($1::bytea,4326),DEFAULT) RETURNING "id"
```

Trying to change the X value yields the following error:

```
src/core/qgsmessagelog.cpp:29 : (logMessage) [2ms] 2020-05-23T17:13:45 Messages[2] Layer test_gen_col3 : PostGIS error while changing attributes: ERROR:  column "x" can only be updated to DEFAULT
DETAIL:  Column "x" is a generated column.
```
This is the query that gets sent to the DBMS, which generates the above error:

```sql
UPDATE "public"."test_gen_col3" SET "x"=31 WHERE "id"=1
```
This is also written to the user console as a critical error.

The correct solution, I think, is to make generated fields (and other fields with DEFAULT values, perhaps) read-only, but I'd have to study the code more deeply in order to better understand how we can send this information from the provider to the UI.

m-kuhn:
I think there would be a preferable initial implementation.

IMHO, the first thing to do to implement generated columns is that the postgres provider will need to check with every insert or update if any of the involved attributes is generated and replace (or omit them). This can be done without touching the defaultValueClause(), directly where the insert or update statement is created. This should be a bit more robust.

Making the attributes read only would be an additional nice move as you say, but is not required for a minimal implementation.

Another nice move would be to consider that the geometry might also be generated (e.g. from x and y attributes), but that's also nice to have.

espinafre:
> Another nice move would be to consider that the geometry might also be generated (e.g. from x and y attributes), but that's also "nice to have".

I was going to reply that generated geometry is what motivated this (centroid from polygon) and is what I've put on the test cases, but it has just dawned on me that you mean that the main geometry of the table might be generated. I think it can be done with the current approach, but the user interface needs some polishing; he must have a way to insert a feature without a geometry.

> IMHO, the first thing to do to implement generated columns is that the postgres provider will need to check with every insert or update if any of the involved attributes is generated and replace (or omit them). 

Give me a couple of days and I will try this approach. Also, I've just realized that on my test cases there is nothing that covers UPDATEs, I'll do that as well.

m-kuhn:
> but the user interface needs some polishing; he must have a way to insert a feature without a geometry.

There's already an "add row" button in the attribute table for this purpose.

espinafre:
Okay, here ( daa3bba ) are the changes according to @m-kuhn's suggestion, that is, checking whether each field is generated, and omit those that are in case of INSERT or UPDATEs. defaultValueClause() now returns the expression used to generate the field, instead of the keyword "DEFAULT"; this way, the user can see what is happening.

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -2313,6 +2338,13 @@ bool QgsPostgresProvider::addFeatures( QgsFeatureList &flist, Flags flags )
         continue;
 
       QString fieldname = mAttributeFields.at( idx ).name();
+
+      if ( mGeneratedValues.contains( idx ) && !mGeneratedValues.value( idx, QString() ).isEmpty() )
```
m-kuhn, line 2342:
```suggestion
      if ( !mGeneratedValues.value( idx ).isEmpty() )
```

That should be sufficient

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -2339,7 +2371,12 @@ bool QgsPostgresProvider::addFeatures( QgsFeatureList &flist, Flags flags )
 
       if ( i == flist.size() )
       {
-        if ( qgsVariantEqual( v, defVal ) )
+        // generated values override everything.
+        if ( !mGeneratedValues[idx].isEmpty() )
```
m-kuhn, line 2375:
```suggestion
        if ( !mGeneratedValues.value( idx ).isEmpty() )
```

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -2339,7 +2371,12 @@ bool QgsPostgresProvider::addFeatures( QgsFeatureList &flist, Flags flags )
...
+        {
+          values += delim + defVal;
```
m-kuhn, line 2377:
```suggestion
          values += delim + QStringLiteral( "DEFAULT" );
```

We need to specify `DEFAULT` here, not the calculation expression, right?

src/providers/postgres/qgspostgresprovider.cpp
espinafre: Sure, resolved with commit 793409c.

src/providers/postgres/qgspostgresprovider.cpp
espinafre: Actually this is dead code, since we don't use DEFAULT anymore, and skip the GENERATED field entirely (line 2342)

src/providers/postgres/qgspostgresprovider.cpp
espinafre: Resolved with 793409c (condition was deleted).

espinafre:
Hi, 6374338 adds the tests for changing of attributes and geometries for tables with GENERATED columns.
@m-kuhn , is pushError() on src/providers/postgres/qgspostgresprovider.cpp#L2955-L2959 
 
 the right way to warn the user that he is trying to do something wrong?

espinafre:
Hi, @m-kuhn , are you okay with these changes? Can we merge?

I'd like to call your attention to the fact that this new code path is not yet tested on Travis, because it still uses PostgreSQL 11.

I've put a PostgreSQL version check on the loading script, and also on the test method itself, so that if it detects we are running against PostgreSQL older than 12, the method immediately returns. That's because the new syntax is incompatible with older PostgreSQL versions. However, as I've stated earlier, the tests run correctly on my local system, which has PostgreSQL 12.3.

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -881,15 +881,17 @@ bool QgsPostgresProvider::loadFields()
...
               " FROM pg_attribute"
               " LEFT OUTER JOIN pg_attrdef ON attrelid=adrelid AND attnum=adnum"
 
               // find unique constraints if present. Text cast required to handle int2vector comparison. Distinct required as multiple unique constraints may exist
               " LEFT OUTER JOIN ( SELECT DISTINCT indrelid, indkey, indisunique FROM pg_index WHERE indisunique ) uniq ON attrelid=indrelid AND attnum::text=indkey::text "
 
-              " WHERE attrelid IN %2"
-            ).arg( connectionRO()->pgVersion() >= 100000 ? QStringLiteral( ", attidentity" ) : QString() ).arg( tableoidsFilter );
+              " WHERE attrelid IN %3"
+            ).arg( connectionRO()->pgVersion() >= 100000 ? QStringLiteral( ", attidentity" ) : QString() )
```
m-kuhn, line 892:
```suggestion
            ).arg( connectionRO()->pgVersion() >= 100000 ? QStringLiteral( ", attidentity" ) : QString(),
```

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -881,15 +881,17 @@ bool QgsPostgresProvider::loadFields()
...
+            .arg( connectionRO()->pgVersion() >= 120000 ? QStringLiteral( ", attgenerated" ) : QString() )
```
m-kuhn, line 893:
```suggestion
            connectionRO()->pgVersion() >= 120000 ? QStringLiteral( ", attgenerated" ) : QString(),
```

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -881,15 +881,17 @@ bool QgsPostgresProvider::loadFields()
...
+            .arg( tableoidsFilter );
```
m-kuhn, line 894:
```suggestion
            tableoidsFilter );
```

A single `.arg` is more performant.

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -2313,6 +2338,13 @@ bool QgsPostgresProvider::addFeatures( QgsFeatureList &flist, Flags flags )
         continue;
 
       QString fieldname = mAttributeFields.at( idx ).name();
+
+      if ( !mGeneratedValues.value( idx, QString() ).isEmpty() )
+      {
+        QgsDebugMsg( QStringLiteral( "Skipping field %1 (idx %2) which is GENERATED." ).arg( fieldname, idx ) );
```
m-kuhn, line 2344:
```suggestion
        QgsDebugMsg( QStringLiteral( "Skipping field %1 (idx %2) which is GENERATED." ).arg( fieldname, QString::number( idx ) ) );
```

This was calling the wrong overload. https://doc.qt.io/qt-5/qstring.html#arg

m-kuhn (commented):
Looks good to me now, thanks. Just some minor string formatting left.

For the tests, we could consider to upgrade postgres on travis to 12, I'm not sure what other side effects that will have.

Concerning error reporting. Is making this message as prominent as putting it on the message bar justified? Or would it be better in the message log?
Basically, is it something which is the result of a user mistake and he needs to be informed (message bar) or more a code problem / unimportant side effect (message log).
I'm fine with both.

src/providers/postgres/qgspostgresprovider.cpp
espinafre: Addressed by 5a56586

src/providers/postgres/qgspostgresprovider.cpp
espinafre: Addressed by 5a56586

src/providers/postgres/qgspostgresprovider.cpp
espinafre: Addressed by 5a56586

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -2341,7 +2341,7 @@ bool QgsPostgresProvider::addFeatures( QgsFeatureList &flist, Flags flags )
 
       if ( !mGeneratedValues.value( idx, QString() ).isEmpty() )
       {
-        QgsDebugMsg( QStringLiteral( "Skipping field %1 (idx %2) which is GENERATED." ).arg( fieldname, idx ) );
+        QgsLogger::warning( tr( "Skipping field %1 which is GENERATED." ).arg( fieldname ) );
```
m-kuhn, line 2344: This will lead to warnings on every "add feature" with generated columns, correct?

espinafre:
Commit 5a56586 addresses issues pointed out by @m-kuhn , and also replaces pushError() with QgsLogger::warning. I've done this because the user has no way, in principle, to detect that a field is auto-generated by the database, so I deem it is not a user error if he tries to change a generated field. However, the user has to know that a field was not updated, and we warn him so.

Below are the last few lines of the output of `ctest -V -R PyQgsPostgresProvider` locally against PostgreSQL 12.3 (I have the full logs if needed):

```
591: PostGIS(1): FAILURE: Field -1 not found.
591: PostGIS(1): FAILURE: Field 1000 not found.
591: PostGIS(1): FAILURE: Field -1 not found..src/core/qgsmessagelog.cpp:29 : (logMessage) [2ms] 2020-06-02T10:00:20 PostGIS[1] FAILURE: Field -1 not found.
591: src/core/qgsmessagelog.cpp:29 : (logMessage) [0ms] 2020-06-02T10:00:20 PostGIS[1] FAILURE: Field 100001 not found.
591: ...
591: ----------------------------------------------------------------------
591: Ran 233 tests in 57.896s
591: 
591: OK
591: 
591: PostGIS(1): FAILURE: Field 1000 not found.
591: PostGIS(1): FAILURE: Field -1 not found.
591: PostGIS(1): FAILURE: Field 100001 not found.
1/1 Test #591: PyQgsPostgresProvider ............   Passed   69.19 sec

The following tests passed:
        PyQgsPostgresProvider

100% tests passed, 0 tests failed out of 1

Total Test time (real) =  69.64 sec
```

src/providers/postgres/qgspostgresprovider.cpp
espinafre: Yes, correct.

espinafre: I've removed the "idx" from the message because that shouldn't matter to the user, and developers can easily find out the field index if needed.

m-kuhn: I would completely remove this (or make it a debug message as before), as it currently is, the log will be spammed and there is no way out.

espinafre: Of course, this hadn't ocurred to me; I was thinking about inserting features only with the user-editable fields. Reverted to debug message in 0d1a4c5.

m-kuhn (approved):
Everything has been addressed and code is looking good.
Let's give headway to PG12 goodness :+1: 

Thanks a lot @espinafre for the pleasurable work

espinafre:
It is I who have to thank you @m-kuhn for your patience and guidance.

Speaking of which, is there something already in place so we can start working on marking read-only fields as such to the user?

Do you think it would be good to backport to LTR? Or label this as a feature so my name can get included in the QGIS hall of fame? :smile: 

Almost everything green here, except for the 32 bit Azure build (which is failing for everyone else too).

m-kuhn:
> Speaking of which, is there something already in place so we can start working on marking read-only fields as such to the user?

Not that I was aware of it, but possibly ...

> Or label this as a feature so my name can get included in the QGIS hall of fame?

If you can paste the text for the changelog including the name that should appear and ideally a screenshot, I'll be happy to put it to [the visual changelog](https://changelog.qgis.org/en/qgis/version/3.14/) for you.

Regarding backport, it would be nice to support this, but I'd give it at least one month to cool down and see if it has any unexpected side effects.

espinafre:
@m-kuhn , for the visual changelog, in the "Data Providers" section, right after Pasotti's "PG raster expose set filter to app", something like:

> ### PG: save features into tables with generated fields
> 
> Allow updating/inserting features in PostgreSQL tables with generated fields. GENERATED columns is a feature introduced by PostgreSQL 12, which allows column values to be generated from other columns in the same table; this replaces the creation of triggers to, for instance,
> automatically populate and keep up-to-date a column of centroids or areas for the polygons in the main geometry column.
> By [José de Paula Rodrigues N. Assis](https://api.github.com/users/espinafre)

Unfortunately, I couldn't come up with good screenshots; all I could think about is something showing the error before the change, and filled fields after it, but I don't think these are good enough material.

Might my name be added to the Contributors list in the About dialog? Thank you very much.

m-kuhn:
Here you go:

https://changelog.qgis.org/en/qgis/version/3.14/#postgres-save-features-into-tables-with-generated-

Help yourself into the about dialog by creating a pull request for resources/data/contributors.json

PR #36000, opened 2020-04-25
Improvements to the vector tile writer

wonder-sk:
Random improvements to the writer to make it more useful:

- filtering of input layers by expressions and min/max zoom level
- custom layer names in the output
- writing of custom metadata for MBTiles output
- auto-calculate output extent (instead of defaulting to the whole world's extent)
- passing transform context to the encoder

### DIFF
--- a/src/core/vectortile/qgsvectortilewriter.cpp
+++ b/src/core/vectortile/qgsvectortilewriter.cpp
@@ -18,6 +18,7 @@
 #include "qgsdatasourceuri.h"
 #include "qgsfeedback.h"
 #include "qgsjsonutils.h"
+#include "qgslogger.h"
 #include "qgsmbtiles.h"
 #include "qgstiles.h"
 #include "qgsvectorlayer.h"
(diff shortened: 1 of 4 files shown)


src/core/vectortile/qgsvectortilewriter.cpp
```diff
@@ -99,35 +110,62 @@ bool QgsVectorTileWriter::writeTiles( QgsFeedback *feedback )
...
+        QgsCoordinateTransform ct( QgsCoordinateReferenceSystem( "EPSG:3857" ), QgsCoordinateReferenceSystem( "EPSG:4326" ), mTransformContext );
+        QgsRectangle wgsExtent = ct.transform( outputExtent );
+        boundsStr = QString( "%1,%2,%3,%4" )
+                    .arg( wgsExtent.xMinimum() ).arg( wgsExtent.yMinimum() )
+                    .arg( wgsExtent.xMaximum() ).arg( wgsExtent.yMaximum() );
+      }
+      catch ( const QgsCsException & )
+      {
+      }
+      mbtiles->setMetadataValue( "bounds", boundsStr );
```
nyalldawson, line 144: Should this be moved up to sit inside the try block? (Ie if transformation fails then don't set the metadata key)

nyalldawson (approved):

src/core/vectortile/qgsvectortilewriter.cpp
wonder-sk: That's probably better than adding default bounds. Thanks for the suggestion, added in 745a537

PR #36005, opened 2020-04-26
Ensure 3D rendering options are disabled when adding a mesh layer

### LINKED ISSUE #34985: [mesh] The 3d renderer checkbox is unchecked by default with no effect

DelazJ:
Fixes #34985 (seems that #34999 did not)

### DIFF
--- a/src/app/3d/qgsmeshlayer3drendererwidget.cpp
+++ b/src/app/3d/qgsmeshlayer3drendererwidget.cpp
@@ -37,7 +37,6 @@ QgsMeshLayer3DRendererWidget::QgsMeshLayer3DRendererWidget( QgsMeshLayer *layer,
   mWidgetMesh = new QgsMesh3dSymbolWidget( layer, this );
   mWidgetMesh->configureForDataset();
   layout->addWidget( mWidgetMesh );
-  mWidgetMesh->setEnabled( false );
 
   connect( mChkEnabled, &QCheckBox::clicked, this, &QgsMeshLayer3DRendererWidget::onEnabledClicked );
   connect( mWidgetMesh, &QgsMesh3dSymbolWidget::changed, this, &QgsMeshLayer3DRendererWidget::widgetChanged );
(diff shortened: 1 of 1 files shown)


src/app/3d/qgsmeshlayer3drendererwidget.cpp
```diff
@@ -37,6 +37,7 @@ QgsMeshLayer3DRendererWidget::QgsMeshLayer3DRendererWidget( QgsMeshLayer *layer,
   mWidgetMesh = new QgsMesh3dSymbolWidget( layer, this );
   mWidgetMesh->configureForDataset();
   layout->addWidget( mWidgetMesh );
+  mWidgetMesh->setEnabled( false );
```
DelazJ, line 40: An alternative could be to call [onEnabledClicked()](https://github.com/qgis/QGIS/blob/master/src/app/3d/qgsmeshlayer3drendererwidget.cpp#L84). Don't know which one is better/safer/less expensive...

vcloarec, line 40: thanks for reporting (again) this issue. Indeed, #34999 didn't fix the issue... 
Instead, I would add this line here at line 60:
https://github.com/DelazJ/QGIS/blob/e9f1f71451a6b90c6f872dec3583f44f8945f32b/src/app
The widget was not set disable if  3D renderer was not provided...
Thanks again.

vcloarec (approved):

DelazJ:
@vcloarec is anything missing?

vcloarec:
@DelazJ  approved!

DelazJ:
Oh sorry. I forgot you do not have the commit right, yet. Let's wait for a committer. Thanks

DelazJ:
I Think this one should be backported to ltr also but it might require #34999 backport first.

PR #36006, opened 2020-04-26
Server standalone: read hostname and port from HOST header (if any)

elpaso:
A tiny change: use the default hostname and port from the incoming request HOST header (if defined).

### DIFF
--- a/src/server/qgis_mapserver.cpp
+++ b/src/server/qgis_mapserver.cpp
@@ -322,15 +322,6 @@ int main( int argc, char *argv[] )
             throw HttpException( QStringLiteral( "HTTP error unsupported method: %1" ).arg( methodString ) );
           }
 
-          // Build URL from env ...
-          QString url { qgetenv( "REQUEST_URI" ) };
-          // ... or from server ip/port and request path
-          if ( url.isEmpty() )
-          {
-            const QString path { firstLinePieces.at( 1 )};
-            url = QStringLiteral( "http://%1:%2%3" ).arg( ipAddress ).arg( port ).arg( path );
-          }
-
           const QString protocol { firstLinePieces.at( 2 )};
           if ( protocol != QStringLiteral( "HTTP/1.0" ) && protocol != QStringLiteral( "HTTP/1.1" ) )
           {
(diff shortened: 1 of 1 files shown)


src/server/qgis_mapserver.cpp
```diff
@@ -357,6 +348,23 @@ int main( int argc, char *argv[] )
...
+          {
+            const QString path { firstLinePieces.at( 1 )};
+            // Take Host header if defined
+            if ( headers.contains( QStringLiteral( "Host" ) ) )
+            {
+              url = QStringLiteral( "http://%1%2" ).arg( headers.value( QStringLiteral( "Host" ) ) ).arg( path );
+            }
+            else
+            {
+              url = QStringLiteral( "http://%2:%3%4" ).arg( ipAddress ).arg( port ).arg( path );
```
rouault, line 364: shouldn't that be "http://%1:%2%3"  ?

elpaso: oh, yes of course, thanks

rouault: would be good if there was a tool to detect this. gcc can detect bad arguments passed to good old printf() :-)

elpaso, line 364:
```suggestion
              url = QStringLiteral( "http://%1:%2%3" ).arg( ipAddress ).arg( port ).arg( path );
```

elpaso: Yeah, I think all we get from Qt is a runtime warning.

m-kuhn: Not even, it will just work as expected. After "fixing" it you'll have

`QString( "http://%1:%2%3" ).arg( ipAddress )` -> returns `QString( "http://127.0.0.1:%2%3" )`
`QString( "http://127.0.0.1:%2%3" ).arg( port )` -> ...

see https://doc.qt.io/qt-5/qstring.html#arg

> Returns a copy of this string with the lowest numbered place marker replaced by string a, i.e., %1, %2, ..., %99.

/smart-arse mode off

sbrunner (approved):

src/server/qgis_mapserver.cpp
```diff
@@ -357,6 +348,23 @@ int main( int argc, char *argv[] )
...
+          // Build URL from env ...
+          QString url { qgetenv( "REQUEST_URI" ) };
+          // ... or from server ip/port and request path
+          if ( url.isEmpty() )
+          {
+            const QString path { firstLinePieces.at( 1 )};
+            // Take Host header if defined
+            if ( headers.contains( QStringLiteral( "Host" ) ) )
+            {
+              url = QStringLiteral( "http://%1%2" ).arg( headers.value( QStringLiteral( "Host" ) ) ).arg( path );
```
m-kuhn, line 360: What about https?

elpaso: Nope, I though about it and since HTTPS is not supported by the 15 LOC web server I wrote it does not make sense here, but if you really want to use this standalone thing behind something that handles the certificates (which I would strongly advise against) your reverse proxy (nginx?) would need to take care take care of setting the `HOST` header.

PR #36014, opened 2020-04-27
Allow custom QgsDataItem types a chance to create a info widget

nyalldawson:
This allows plugin provided browser items a chance to create their own information panel widgets

Sponsored by SLYR:
image


nirvn (approved):

NathanW2:
Neat! That would be quite handy for any custom data formats and providers.

PR #36031, opened 2020-04-27
[backport release-3_10] [authentication manager] fix QGIS freeze when asking for master password

nirvn:
Backport of #35994

### DIFF
--- a/resources/themes/Blend of Gray/style.qss
+++ b/resources/themes/Blend of Gray/style.qss
@@ -561,11 +561,15 @@ QTabBar::tab:selected
     margin-bottom: 0px;
 }
 
-QTabBar::tab:!selected:hover
+QTabBar::tab:bottom:!selected
 {
-    /*border-top: 2px solid @selection;
-    padding-bottom: 3px;*/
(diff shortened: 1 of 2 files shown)


src/core/auth/qgsauthmanager.cpp
```diff
@@ -103,6 +103,7 @@ QgsAuthManager *QgsAuthManager::instance()
 QgsAuthManager::QgsAuthManager()
 {
   mMutex = new QMutex( QMutex::Recursive );
+  mMasterPasswordMutex = new QMutex( QMutex::Recursive );
```
m-kuhn, line 106: I think this is leaked?
Use a unique_ptr instead?

nirvn: It leaks, like the pre-existing mMutex, but assumed it was fine since we're only running one instance until app is closed.

m-kuhn: We won't crash the system, but it's better to clean things up so tools like valgrind give cleaner reports.

nirvn: If you want, I can push an extra commit moving both the new and pre existing mutex to use unique_ptr.

m-kuhn: Yes please!

nyalldawson (approved):

PR #36032, opened 2020-04-27
[processing] use hours as cost units for service area algorithms (fix #30464)

### LINKED ISSUE #30464: Homogenize travel cost units within network analysis algorithms
When running shortest path algorithm (eg, point to point) with the "fastest" type of travel, the cost field in the output layer is in "hours".
When running the service area algorithms with the "fastest" travel cost, "seconds" are expected as input. Since the latter is not obviously mentioned, this may lead to errors.

It could be nice we use the same units across those algorithms and explicitly expose the units in use.

alexbruy:
## Description
Use hours as units for "travel cost" parameter in the service area algorithms (when "fastest" strategy is used). This makes them consistent with the output of shortest path algorithms. Fixes #30464.


nyalldawson:
This could potentially break existing models, couldn't it?

alexbruy:
The only breakega is that default value (if exists) for cost parameter should be updated.

nyalldawson:
> The only breakage is that default value (if exists) for cost parameter should be updated.

That's what I'm concerned about. What do you think about adding a new parameter for default in hours, and hiding the old one and only using it when the new one isn't specified?

alexbruy:
Probably then better to leave it as is to avoid another breakage in future when old hidden parameter will be removed and hours parameter will be renamed to the old consistent name?

alexbruy:
Ok, I have added new parameted and hide the old one. Still not sure if this is a right way, as we will need to remove and rename duplicated parameters at some stage.

src/analysis/processing/qgsalgorithmserviceareafromlayer.cpp
```diff
@@ -78,6 +87,12 @@ QVariantMap QgsServiceAreaFromLayerAlgorithm::processAlgorithm( const QVariantMa
     throw QgsProcessingException( invalidSourceError( parameters, QStringLiteral( "START_POINTS" ) ) );
 
   double travelCost = parameterAsDouble( parameters, QStringLiteral( "TRAVEL_COST" ), context );
+  if ( !travelCost )
+    travelCost = parameterAsDouble( parameters, QStringLiteral( "TRAVEL_COST2" ), context );
```
nyalldawson, line 91:
```suggestion
  // use older deprecated travel cost style if specified, to maintain old api
  const bool useOldTravelCost = parameters.value( QStringLiteral("TRAVEL_COST" ) ).isValid();
  double travelCost = parameterAsDouble( parameters, useOldTravelCost ? QStringLiteral( "TRAVEL_COST" ) : QStringLiteral( "TRAVEL_COST2" ), context );
```

src/analysis/processing/qgsalgorithmserviceareafromlayer.cpp
```diff
@@ -78,6 +87,12 @@ QVariantMap QgsServiceAreaFromLayerAlgorithm::processAlgorithm( const QVariantMa
...
+
+  int strategy = parameterAsInt( parameters, QStringLiteral( "STRATEGY" ), context );
+  if ( strategy )
```
nyalldawson, line 94:
```suggestion
  if ( strategy && !useOldTravelCost )
```

src/analysis/processing/qgsalgorithmserviceareafromlayer.cpp
nyalldawson: Same changes apply to qgsalgorithmserviceareafrompoint.cpp

nyalldawson (approved):

alexbruy:
Don't know why Processing test fails of Travis, it passes locally.

PR #36039, opened 2020-04-27
[FEATURE][MESH] Identify tool for 1d mesh

vcloarec:
This PR enables the identify tool for 1D mesh.
The identify tool displays dataset scalar and vector values on the edge the closest of the clicked point.
For dataset on vertex, the dataset are interpolated depending on the projected clicked point on the edge.

identifyTool1D

### DIFF
--- a/src/core/mesh/qgsmeshlayer.h
+++ b/src/core/mesh/qgsmeshlayer.h
@@ -259,8 +259,8 @@ class CORE_EXPORT QgsMeshLayer : public QgsMapLayer
 
     /**
       * Interpolates the value on the given point from given dataset.
-      * For 3D datasets, it uses #dataset3dValue \n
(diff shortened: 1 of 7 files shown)


src/core/mesh/qgsmeshlayer.h
```diff
@@ -257,27 +259,32 @@ class CORE_EXPORT QgsMeshLayer : public QgsMapLayer
 
     /**
       * Interpolates the value on the given point from given dataset.
+      * For 3D datasets, it uses #dataset3dValue \n
```
PeterPetrik, line 262: instead of `#dataset3dValue`, just use  `dataset3dValue()` it is autoexpand

PeterPetrik (approved):

PR #36042, opened 2020-04-27
Deprecate QgsLegendRenderer methods which use raw QPainter and force use of QgsRenderContext

nyalldawson:
TODO QGIS 4.0 -- make the use of QgsRenderContext mandatory all throughout the legend renderer code -- this should not be optional


src/core/qgslegendsettings.h
```diff
@@ -272,57 +347,57 @@ class CORE_EXPORT QgsLegendSettings
      */
     void drawText( QPainter *p, const QRectF &rect, const QString &text, const QFont &font, Qt::AlignmentFlag halignment = Qt::AlignLeft, Qt::AlignmentFlag valignment = Qt::AlignTop, int flags = Qt::TextWordWrap ) const;
 
-    //! Returns a font where size is in pixel and font size is upscaled with FONT_WORKAROUND_SCALE
+//! Returns a font where size is in pixel and font size is upscaled with FONT_WORKAROUND_SCALE
```
nirvn, line 350: @nyalldawson , what's with the dox on-liners losing their intent?

nirvn (approved):

src/core/qgslegendsettings.h
nyalldawson: Not sure how that happened, fixed

PR #36047, opened 2020-04-28
Add support for retrieving temporal extent from WMS-T version 1.1

nyalldawson:
e.g. allows correct use of temporal capabilities from https://ies-ows.jrc.ec.europa.eu/effis

@samweli as discussed

### DIFF
--- a/src/providers/wms/qgswmscapabilities.cpp
+++ b/src/providers/wms/qgswmscapabilities.cpp
@@ -1049,8 +1049,6 @@ void QgsWmsCapabilities::parseLegendUrl( const QDomElement &element, QgsWmsLegen
 
 void QgsWmsCapabilities::parseDimension( const QDomElement &element, QgsWmsDimensionProperty &dimensionProperty )
 {
-
-  // TODO, check for the name value if it is time, implement WMS-T support
   dimensionProperty.name = element.attribute( QStringLiteral( "name" ) );
   dimensionProperty.units = element.attribute( QStringLiteral( "units" ) );
   dimensionProperty.unitSymbol = element.attribute( QStringLiteral( "unitSymbol" ) );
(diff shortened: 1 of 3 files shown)


src/providers/wms/qgswmscapabilities.cpp
```diff
@@ -1077,6 +1075,39 @@ void QgsWmsCapabilities::parseDimension( const QDomElement &element, QgsWmsDimen
   dimensionProperty.extent = element.text();
 }
 
+void QgsWmsCapabilities::parseExtent( const QDomElement &element, QVector<QgsWmsDimensionProperty> &dimensionProperties )
+{
+  const QString name = element.attribute( QStringLiteral( "name" ) );
+  // try to find correponsding dimension property -- i.e. we upgrade the WMS 1.1 split of Dimension and Extent to 1.3 style where Dimension holds the extent information
```
elpaso, line 1081:
```suggestion
  // try to find corresponding dimension property -- i.e. we upgrade the WMS 1.1 split of Dimension and Extent to 1.3 style where Dimension holds the extent information
```

src/providers/wms/qgswmscapabilities.cpp
```diff
@@ -1077,6 +1075,39 @@ void QgsWmsCapabilities::parseDimension( const QDomElement &element, QgsWmsDimen
...
+    if ( it->name == name )
+    {
+      it->extent = element.text();
+
+      it->defaultValue = element.attribute( QStringLiteral( "default" ) );
+
+      if ( !element.attribute( QStringLiteral( "multipleValues" ) ).isNull() )
+      {
+        QString multipleValuesAttribute = element.attribute( QStringLiteral( "multipleValues" ) );
+        it->multipleValues = ( multipleValuesAttribute == QLatin1String( "1" ) || multipleValuesAttribute == QLatin1String( "true" ) );
```
elpaso, line 1093:
```suggestion
        it->multipleValues = ( multipleValuesAttribute == '1' ) || multipleValuesAttribute == QLatin1String( "true" ) );
```

elpaso (approved):

Samweli:
Looks good @nyalldawson

src/providers/wms/qgswmscapabilities.cpp
nyalldawson: Actually there's no QString operator==( QChar ) overload, so this would use the `QString( QChar() )` implicit constructor and end up calling `operator==( QString )`. Eeek - we're actually better off leaving it at QLatin1String to utilise the optimised equality operator.

PR #36051, opened 2020-04-28
[QgsQuick] Fix overlaped variable name with QT widget's variable

sklencar:
Qt 5.14 introduced currentValue variable for combobox widget which clashes with QgsQuick variable. Note that EditorWidgetComboBox is extended Combobox from QtQuick.
ref: https://doc.qt.io/qt-5/qml-qtquick-controls2-combobox.html#currentValue-prop


PeterPetrik:
can you check in QGIS codebase we do not use it?

wonder-sk (approved):

PR #36064, opened 2020-04-28
Set maximum time step to max int. Fix #35887.

### LINKED ISSUE #35887: Not possible to set the time step to more than 99.990 units
In the new Time Manager panel, I cannot set the time to more than 99.990 units. I know you can jump to the next time unit (e.g. instead of 300 seconds, using 5 minutes), but users should be able to change to whatever value they want.
Peek 2020-04-21 07-34

ismailsunni:
## Description
Set maximum time step to max int. Fix #35887.


saberraz:
Thanks @ismailsunni !!

nyalldawson (approved):

PR #36065, opened 2020-04-28
[feature][processing] Complete random raster algorithm collection

root676:
## Description

_Some random stuff again..._

This PR refactors and completes the recently added work on **random number raster layer creation algorithms** (see [#35835.](https://github.com/qgis/QGIS/pull/35835)) With the new algorithms aimed towards distribution based random number creation, QGIS reaches to the same level of functionality as current ArcGIS random raster creation tools. In total, the PR adds the following algorithms (normal and uniform raster layer creation algs are refactored to be in line the new naming scheme):

- Create random raster layer (binomial distribution)
- Create random raster layer (exponential distribution)
- Create random raster layer (gamma distribution)
- Create random raster layer (negative binomial distribution)
- Create random raster layer (normal distribution)
- Create random raster layer (poisson distribution)
- Create random raster layer (uniform distribution)

Tests have been added for the algorithms data type choices.
I'm looking forward to comments/questions/reviews/etc.!


nyalldawson:
@root676 this is GREAT! :smile: 

Looking at the algorithms, there's a lot of common code between them all (basically everything after the creation of the random distribution is identical). I think we could avoid a lot of the duplicate code by making a common base class for the algorithms. Are you comfortable doing this?

root676:
Thanks, Yes I thought about this too... the first time I wanted to implement the algorithms I actually tried to do it with a base class (similar to the Fuzzify algs). But then I ran into some troubles because the random number distribution objects and available raster data types vary (also in type) among all algorithms. I saw no chance in using them as an argument in a pure virtual `obtainRandomNumber(distribution)` method). Do you have any hint on how to solve this with a design pattern (maybe strategy pattern: encapsulate the things that vary...)? I'll sure give it a try! :+1:

root676:
I think I have got a solution to the problem, I think I'll push the changes in a new PR and close this one. What do you think?

PR #36069, opened 2020-04-28
add expression close_line

JanCaha:
Adds expression `close_line` which returns closed LineString geometry if the LineString was not previously closed. Includes tests and documentation.

I am not completely sure how to handle multipart lines. Right now these return `QVariant()` as any geometry that is not `QgsLineString`.

### DIFF
--- a/resources/function_help/json/close_line
+++ b/resources/function_help/json/close_line
@@ -1,0 +1,8 @@
+{
+  "name": "close_line",
+  "type": "function",
+  "description": "Returns closed line string of the input line string by appending the first point to the end of the line, if it is not already closed. If the geometry is not a line string or if it is a multi line string then the result will be null.",
+  "arguments": [ {"arg":"geometry","description":"a line string geometry"}],
+  "examples": [ { "expression":"geom_to_wkt(close_line(geom_from_wkt('LINESTRING(0 0, 1 0, 1 1)')))", "returns":"LineString (0 0, 1 0, 1 1, 0 0)"},
+                { "expression":"geom_to_wkt(close_line(geom_from_wkt('LINESTRING(0 0, 1 0, 1 1, 0 0)')))", "returns":"LineString (0 0, 1 0, 1 1, 0 0)"}]
+}
--- a/resources/function_help/json/close_line
+++ b/resources/function_help/json/close_line
@@ -1,7 +1,7 @@
 {
   "name": "close_line",
   "type": "function",
-  "description": "Returns closed line string of the input line string by appending the first point to the end of the line, if it is not already closed. If the geometry is not a line string or if it is a multi line string then the result will be null.",
+  "description": "Returns a closed line string of the input line string by appending the first point to the end of the line, if it is not already closed. If the geometry is not a line string or multi line string then the result will be null.",
   "arguments": [ {"arg":"geometry","description":"a line string geometry"}],
   "examples": [ { "expression":"geom_to_wkt(close_line(geom_from_wkt('LINESTRING(0 0, 1 0, 1 1)')))", "returns":"LineString (0 0, 1 0, 1 1, 0 0)"},
                 { "expression":"geom_to_wkt(close_line(geom_from_wkt('LINESTRING(0 0, 1 0, 1 1, 0 0)')))", "returns":"LineString (0 0, 1 0, 1 1, 0 0)"}]
--- a/src/core/expression/qgsexpressionfunction.cpp
+++ b/src/core/expression/qgsexpressionfunction.cpp
@@ -3392,6 +3392,24 @@ static QVariant fcnIsClosed( const QVariantList &values, const QgsExpressionCont
   return QVariant::fromValue( curve->isClosed() );
 }
 
+static QVariant fcnCloseLine( const QVariantList &values, const QgsExpressionContext *, QgsExpression *parent, const QgsExpressionNodeFunction * )
+{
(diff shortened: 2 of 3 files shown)


nyalldawson:
> I am not completely sure how to handle multipart lines. Right now these return QVariant() as any geometry that is not QgsLineString

Maybe we should close ALL the parts of a multilinestring? I agree that it's not an obvious answer, but I think that would be more predictable then skipping these. (There's also the situation that linestring shapefiles always import as multilinestrings)

src/core/expression/qgsexpressionfunction.cpp
```diff
@@ -3392,6 +3392,24 @@ static QVariant fcnIsClosed( const QVariantList &values, const QgsExpressionCont
...
+
+  if ( geom.isNull() || geom.isMultipart() )
+    return QVariant();
+
+  const QgsLineString *line = qgsgeometry_cast< const QgsLineString * >( geom.constGet() );
+
+  if ( !line )
+    return QVariant();
+
+  QgsLineString *closed_line = line->clone();
```
nyalldawson, line 3407:
```suggestion
  std::unique_ptr< QgsLineString > closedLine( line->clone() );
```

nyalldawson:
Nice addition!!

src/core/expression/qgsexpressionfunction.cpp
JanCaha: I was wondering how to do this correctly. My solution worked, but I was thinking that there is probably a more correct solution. Is there a specific reason why to handle it like this?

JanCaha:
@nyalldawson may I ask for some advice? I spend quite some on it and I am not able to move forward.

I have something like this, which is roughly inspired by `fcnReverse`. 

```
const QgsLineString *line = qgsgeometry_cast< const QgsLineString * >( geom.constGet() );

  if ( !line && geom.isMultipart())
  {
    if ( const QgsGeometryCollection *collection = qgsgeometry_cast< const QgsGeometryCollection * >( geom.constGet() ) )
    {
      std::unique_ptr< QgsGeometryCollection > closed( collection->createEmptyWithSameType() );

      for ( int i = 0; i < collection->numGeometries(); ++i )
      {
        line = qgsgeometry_cast< const QgsLineString * >( collection->geometryN( i ) );

        std::unique_ptr< QgsLineString > closedLine( line->clone());
        closedLine->close();

        closed->addGeometry( closedLine );
      }

      return QVariant::fromValue( closed );
    }    
  }
```
However the line ` closed->addGeometry( closedLine );` throws an error (QgsLineString cannot be used in addGeometry()). I understand the error but have no idea how to get around it. It seems that I quite dont get the conversions amongst all these geometry types.

nyalldawson:
I think the issue comes from here:

```
        std::unique_ptr< QgsLineString > closedLine( line->clone());
        closedLine->close();
        closed->addGeometry( closedLine ); <<<<<<
```

Because closedLine is a unique pointer, you need to release ownership of the associated value, which you'd do via:

```
        closed->addGeometry( closedLine.release() );
```

But more generally, I think you're correct in that we should adapt the close_line function to follow the general structure from fcnReverse. So I think the result would look something like this:

```
  QVariant result;
  if ( !fGeom.isMultipart() )
  {
    const QgsLineString *line = qgsgeometry_cast<const QgsLineString * >( fGeom.constGet() );
    if ( !line )
      return QVariant();

    std::unique_ptr< QgsLineString > closedLine( line->clone() );
    closedLine->close();
    result = QVariant::fromValue( QgsGeometry( std::move( closedLine ) ) );
  }
  else
  {
    const QgsGeometryCollection *collection = qgsgeometry_cast< const QgsGeometryCollection *>( fGeom.constGet() );
    std::unique_ptr< QgsGeometryCollection > closed( collection->createEmptyWithSameType() );
    for ( int i = 0; i < collection->numGeometries(); ++i )
    {
      if ( const QgsLineString *line = qgsgeometry_cast<const QgsLineString * >( collection->geometryN( i ) ) )
      {
        std::unique_ptr< QgsLineString > closedLine( line->clone() );
        closedLine->close();
        closed->addGeometry( closedLine.release() );
      }
    }
    result = QVariant::fromValue( QgsGeometry( std::move( closed ) ) );
  }
  return result;
```

JanCaha:
Perfect. I integrated your solution into what was already done and it seems to be working well.

Btw: Can you suggest me something to read regarding the pointers? It seems that I dont quite get it. Could be generally C++ or QGIS related.

resources/function_help/json/close_line
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "close_line",
+  "type": "function",
+  "description": "Returns closed line string of the input line string by appending the first point to the end of the line, if it is not already closed. If the geometry is not a line string or if it is a multi line string then the result will be null.",
```
nyalldawson, line 4:
```suggestion
  "description": "Returns a closed line string of the input line string by appending the first point to the end of the line, if it is not already closed. If the geometry is not a line string or multi line string then the result will be null.",
```

nyalldawson (approved):

nyalldawson:
> Can you suggest me something to read regarding the pointers? It seems that I dont quite get it. Could be generally C++ or QGIS related.

std::unique_ptr is a new(ish) c++11 thing. https://www.internalpointers.com/post/beginner-s-look-smart-pointers-modern-c is a decent introduction to them...

nyalldawson:
Thanks @JanCaha !

PR #36072, opened 2020-04-28
Backport two crash fixes to 3.10

nyalldawson:
@m-kuhn given that these are crash fixes, I think we should merge immediately. What do you think?


nirvn (approved):

m-kuhn:
> @m-kuhn given that these are crash fixes, I think we should merge immediately. What do you think?

Agreed, what's worse than a crash?

m-kuhn (approved):

PR #36079, opened 2020-04-29
Add mechanism to allow new symbols to be added to the default style library to be updated when QGIS is upgraded

nyalldawson:
Temporarily adds some new patch styles from https://github.com/kartoza/QGIS-Legend-Patches

### DIFF
--- a/resources/symbology-style.xml
+++ b/resources/symbology-style.xml
@@ -1,5 +1,5 @@
 <!DOCTYPE qgis_style>
-<qgis_style version="1">
+<qgis_style version="2">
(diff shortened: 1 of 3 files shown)


nirvn (approved):
Simple, efficient solution. Fantastic.

nyalldawson:
Ok thanks! I'll revert the style inclusions and leave those to follow ups

nirvn:
Good idea. 

We could do with a few default text formats too (we'd probably need to add the possibility to declare fallback font family names to make it useful for the masses though).

PR #36081, opened 2020-04-29
Reduce margins in vector and mesh layer properties dialog

DelazJ:
digitizingtabmixed
3dviewconfigmixed


PeterPetrik:
thanks, looks good

PR #36082, opened 2020-04-29
Improving mesh symbology dialogs (part 1)

DelazJ:
Trying to fix some issues in the mesh layer properties and make them look like other dialogs:
* Use gridLayout instead of combination of horizontal + vertical layouts with no particular setting
* Reduce vertical spacer size and margins to remove unnecessary scrollbar/space, fix alignment
* Add label to widgets to mimic other others in QGIS
* Add tabstops

meshaveragemixed
meshcontoursmixed
meshgridmixed

I think we can get rid of the bold text on top of the tabs being a groupbox label. It can simply be a label/checkbox (as in labels dialog), more coherent with other dialogs with advantages of giving back more spacing for actual widgets. Opinion?
Btw other suggestions welcome. I did not tackle the vector tab, yet.

### DIFF
--- a/src/ui/mesh/qgsmeshrenderer3daveragingwidgetbase.ui
+++ b/src/ui/mesh/qgsmeshrenderer3daveragingwidgetbase.ui
@@ -6,14 +6,14 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>852</width>
-    <height>552</height>
+    <width>454</width>
+    <height>179</height>
    </rect>
   </property>
   <property name="windowTitle">
    <string notr="true">Form</string>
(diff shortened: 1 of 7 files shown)


PeterPetrik:
thanks, looks good

nyalldawson:
Ah, good move! These mesh widgets have always looked rather "3rd party", so it's nice to see them being improved :smile:

nyalldawson:
> I think we can get rid of the bold text on top of the tabs being a groupbox label. It can simply be a label/checkbox (as in labels dialog)

Yes, I agree. I dislike the nested group box approach which is currently used

PR #36087, opened 2020-04-29
Functions for adding a variant of bool params in provider metadata

Samweli:
From this discussion https://github.com/qgis/QGIS/pull/35741#discussion_r416396465.

This PR adds a consistent way of handling bool values in providers uri parameters.

### DIFF
--- a/tests/src/core/testqgsprovidermetadata.cpp
+++ b/tests/src/core/testqgsprovidermetadata.cpp
@@ -1,0 +1,128 @@
+/***************************************************************************
+                         testqgsprovidermetadata.cpp
+                         ---------------
+    begin                : April 2020
+    copyright            : (C) 2020 by Samweli Mwakisambwe
+    email                : samweli at kartoza dot com
+ ***************************************************************************/
+
+/***************************************************************************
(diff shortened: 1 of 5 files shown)


Samweli:
ping @elpaso

tests/src/core/testqgsprovidermetadata.cpp
```diff
@@ -0,0 +1,128 @@
...
+void TestQgsProviderMetadata::cleanup()
+{
+}
+
+void TestQgsProviderMetadata::cleanupTestCase()
+{
+  QgsApplication::exitQgis();
+}
+
+void TestQgsProviderMetadata::checkBoolParameterSetting()
```
elpaso, line 70: Perfect! 
If you want, you can also use https://doc.qt.io/qt-5/qttestlib-tutorial2-example.html

elpaso (approved):

tests/src/core/testqgsprovidermetadata.cpp
Samweli: ooh that is so cool!

PR #36089, opened 2020-04-29
Feature browser for preview in expression builder

3nids:
This adds a feature browser to the expression builder so one can browse features to see the result of the expression
image

A new widget is brought: `QgsFeaturePickerWidget`. For the moment, it's a simple combobox (but inherits widget so we can add a map picker button later on). It behaves the same way than the one in the relation reference widget (an editable combobox).

It relies on the same base model (`QgsFeatureFilterModel`). A new abstract class `QgsFeaturePickerModelBase` regroups the logic of the, and the new model is called `QgsFeaturePickerModel`.
Basically the `QgsFeaturePickerModel` used the feature id to identify the features while `QgsFeatureFilterModel` use a list of attributes (the foreign keys of a relation).
It declutters a bit the former model which was quite...cluttered.

As a side note, I tried to use templates for `QgsFeaturePickerModelBase`, but since templates cannot be mixed with Q_OBJECT, the alternatives sounded like the code would be less readable.

Tests have been added.

Sponsored by the **QGIS Swiss User Group**!


scripts/rename_cpp.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
```
m-kuhn, line 1:
```suggestion
#!/usr/bin/env bash

# This scripts renames the name of a class as well as its header and cpp file
# (assuming they are the lowercase version of the class name).
# 
# Usage: ./scripts/rename_cpp.sh src/core QgsMyClassName QgsMyNewClassName
```

src/core/qgsfeatureexpressionvaluesgatherer.h
```diff
@@ -32,22 +33,60 @@
  *
  * \since QGIS 3.0
```
m-kuhn, line 34:
```suggestion
 * \since QGIS 3.14
```

src/core/qgsfeaturefiltermodel.h
```diff
@@ -28,111 +25,28 @@ class QgsFieldExpressionValuesGatherer;
...
 
-    Q_PROPERTY( QgsVectorLayer *sourceLayer READ sourceLayer WRITE setSourceLayer NOTIFY sourceLayerChanged )
-    Q_PROPERTY( QString displayExpression READ displayExpression WRITE setDisplayExpression NOTIFY displayExpressionChanged )
-    Q_PROPERTY( QString filterValue READ filterValue WRITE setFilterValue NOTIFY filterValueChanged )
-    Q_PROPERTY( QString filterExpression READ filterExpression WRITE setFilterExpression NOTIFY filterExpressionChanged )
-    Q_PROPERTY( bool allowNull READ allowNull WRITE setAllowNull NOTIFY allowNullChanged )
-    Q_PROPERTY( bool isLoading READ isLoading NOTIFY isLoadingChanged )
-
     /**
      * A field of sourceLayer that is unique and should be used to identify features.
```
m-kuhn, line 33:
```suggestion
     * A set of fields of sourceLayer that is unique and should be used to identify features.
```

?

src/core/qgsfeaturefiltermodel.h
```diff
@@ -28,111 +25,28 @@ class QgsFieldExpressionValuesGatherer;
...
      * Normally the primary key field.
      * Needs to match the identifierValue.
      */
-    Q_PROPERTY( QString identifierField READ identifierField WRITE setIdentifierField NOTIFY identifierFieldChanged )
+    Q_PROPERTY( QStringList identifierFields READ identifierFields WRITE setIdentifierFields NOTIFY identifierFieldsChanged )
 
     /**
      * The value that identifies the current feature.
```
m-kuhn, line 40:
```suggestion
     * The values that identifies the current feature.
```

src/gui/qgsexpressionpreviewwidget.cpp
```diff
@@ -0,0 +1,166 @@
...
+      emit expressionParsed( true );
+      setParserError( false );
+      setEvalError( false );
+    }
+  }
+}
+
+void QgsExpressionPreviewWidget::linkActivated( const QString & )
+{
+  QgsMessageViewer *mv = new QgsMessageViewer( this );
```
m-kuhn, line 124: Can we have that scoped only?
This (temporarily) leaks.

```suggestion
  QgsMessageViewer mv( this );
```

m-kuhn (approved):

nyalldawson:
Can I review before merge please?

src/gui/qgsexpressionpreviewwidget.cpp
```diff
@@ -0,0 +1,166 @@
...
+  mv->setWindowTitle( tr( "More Info on Expression Error" ) );
+  mv->setMessageAsHtml( mToolTip );
+  mv->exec();
```
3nids, line 127:
```suggestion
  mv.setWindowTitle( tr( "More Info on Expression Error" ) );
  mv.setMessageAsHtml( mToolTip );
  mv.exec();
```

src/gui/qgsfeaturepickerwidget.cpp
```diff
@@ -0,0 +1,266 @@
...
+  mLineEdit->setSelectOnFocus( true );
+  mLineEdit->setShowClearButton( allowNull() );
+
+  mComboBox->setEditable( true );
+  mComboBox->setLineEdit( mLineEdit );
+  mComboBox->setModel( mModel );
+
+  connect( mLineEdit, &QgsFilterLineEdit::textEdited, this, &QgsFeaturePickerWidget::onCurrentTextChanged );
+
+  setToolTip( tr( "Just start typing what you are looking for." ) );
```
nyalldawson, line 67: Remove this

nyalldawson:
Is there a particular reason why QgsFeaturePickerWidget is a QWidget and not direct QComboBox?

3nids:
> why QgsFeaturePickerWidget is a QWidget

so we have room for improvements (a button for a map tool picker, browsing buttons)

3nids:
now with browsing buttons
image

nirvn:
@3nids , man, that OS X dark theme with black text on dark background, god. Use night mapping instead! :wink:

PR #36091, opened 2020-04-29
Fix unreported binding issue with legend json export

elpaso:
Now you can do:

```python
model = QgsLegendModel(project.layerTreeRoot())
ctx = QgsRenderContext()
settings = QgsLegendSettings()
renderer = QgsLegendRenderer(model, settings)
nodes = renderer.exportLegendToJson(ctx)['nodes'].toVariant()  # List of dicts
```

Without this patch there is no way to obtain the result in Python.


pblottiere:
Oh nice fix! Thanks @elpaso :+1:

What do you think about adding your example as a unit test?

m-kuhn (approved):

src/core/qgslegendrenderer.h
```diff
@@ -98,7 +98,7 @@ class CORE_EXPORT QgsLegendRenderer
      *
      * \since QGIS 3.8
```
m-kuhn, line 99:
```suggestion
     * \since QGIS 3.8
     * \since QGIS 3.14 the python signature has changed to ``QJSoneObject exportLegendToJson( context )``
```

pblottiere:
Thanks for the test :+1:

pblottiere (approved):

nyalldawson:
My question would be why this uses a reference in the first place -- it would be simpler just to return a QJsonObject directly.

nyalldawson:
(and given it hasn't worked from python previously, we've got the opportunity to fix the api now)

elpaso:
> My question would be why this uses a reference in the first place -- it would be simpler just to return a QJsonObject directly.

I asked myself the same question. But since this is one the things I stumbled upon while I was working on something totally different I just wanted to made a quick fix, that has turned in "hours" of work.

I guess this has to wait :(

pblottiere:
> My question would be why this uses a reference in the first place -- it would be simpler just to return a QJsonObject directly.

Well, I agree. I don't know who did that, but it was a bad call (sorry)...

PR #36094, opened 2020-04-29
Enable attribute table save button only when there are uncommited changes

### LINKED ISSUE #18975: Attribute table save layer edits doesn't get inactive after successful save
Author Name: **Alexandre Neto** (@SrNetoChan)
Original Redmine Issue: [10568](https://issues.qgis.org/issues/10568)
Affected QGIS version: 3.3(master)

---
When you are editing a layer, if you press the save layer edits button, after a successful commit the button get inactive. This is helpful to see that the commit was ok and to remind you to save changes.

Trying to do the same in the attribute table the icon never get's inactive (unless you stop editing). Would be nice to have the same behavior in both buttons.

uclaros:
## Description
Make the save button on the attribute table have the same behavior as the *Save Layer Edits* button: Enabled when there are uncommitted changes, disabled when no changes have been made since last save.

Fixes #18975


nyalldawson (approved):

nyalldawson:
@uclaros are you ok if we skip the backport here? given the age of the issue I don't think there's enough reward in the risk of a backport

uclaros:
No problem at all, I see it more as a feature than as a bugfix!

PR #36099, opened 2020-04-29
Vector tile layer - part 8 (writer in Processing)

wonder-sk:
This PR adds Processing integration for vector tile writing. It is just a wrapper around the existing QgsVectorTileWriter class introduced in previous pull requests.

New algorithms in the Processing toolbox:

image

Algorithm's parameters for XYZ:

image

Algorithm's parameters for MBTiles:

image

List of input layers:

image

Input layer details:

image


src/analysis/processing/qgsalgorithmwritevectortiles.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+      if ( !QgsProcessingUtils::mapLayerFromString( inputItemLayer.toString(), *context ) )
+        return false;
+    }
+
+    return true;
+  }
+
+  // TODO: anything else?
+  // - valueAsPythonString()
```
nyalldawson, line 69: yes - it's useful for the history panel

src/analysis/processing/qgsalgorithmwritevectortiles.cpp
```diff
@@ -0,0 +1,195 @@
...
+  // - asScriptCode()
```
nyalldawson, line 70: not required

src/analysis/processing/qgsalgorithmwritevectortiles.cpp
```diff
@@ -0,0 +1,195 @@
...
+  // - asPythonString()
```
nyalldawson, line 71: yes - used when converting models to python

src/analysis/processing/qgsalgorithmwritevectortiles.cpp
```diff
@@ -0,0 +1,195 @@
...
+        // good
+      }
+      else if ( ( inputLayer = qobject_cast< QgsVectorLayer * >( QgsProcessingUtils::mapLayerFromString( layerVariant.toString(), context ) ) ) )
+      {
+        // good
+      }
+      else
+      {
+        // bad
+        throw QgsProcessingException( "unknown input layer" );
```
nyalldawson, line 94: don't forget to tr these

src/analysis/processing/qgsalgorithmwritevectortiles.cpp
```diff
@@ -0,0 +1,195 @@
...
+  int maxZoom = parameterAsInt( parameters, QStringLiteral( "MAX_ZOOM" ), context );
+
+  // prepare output URI
+  QString xyzTemplate = parameterAsString( parameters, QStringLiteral( "XYZ_TEMPLATE" ), context );
+  QgsDataSourceUri dsUri;
+  dsUri.setParam( "type", "xyz" );
+  dsUri.setParam( "url", QUrl::fromLocalFile( xyzTemplate ).toString() );
+  QString uri = dsUri.encodedUri();
+
+  QVariant layersVariant = parameters.value( parameterDefinition( QStringLiteral( "LAYERS" ) )->name() );
```
nyalldawson, line 172:
```suggestion
  QVariant layersVariant = parameters.value( QStringLiteral( "LAYERS" ) );
```

save a little bit of unecessary work :laughing:

nyalldawson:
@wonder-sk in reponse to your questions:

> have a new parameter type, where value would be a QVariantList, and each item in that list would be a QVariantMap in this format:
  {
   'layer':  <string> or <QgsMapLayer>,
   // key-value pairs below are optional
   'layerName': <string>,
   'filterExpression': <string>,
   'minZoom': <int>,
   'maxZoom': <int>
  }

Perfect approach

> have a new widget wrapper in qgis_gui library, which would show a line edit + tool button to open panel to set details.
> new internal panel widget derived from QgsProcessingMultipleSelectionPanelWidget that is similar to QgsProcessingMultipleInputPanelWidget, but allows opening extra sub-panel for each vector layer to configure the extra attributes like filterExpression.
> @nyalldawson am I on the right track here or wandering into some deep dark forest? :-)

Looks good to me. A possible alternative would be to go with a table/model style approach (kind of like how the interpolation algorithms expose this functionality), but to be honest I think your approach will ultimately lead to a more usable result (for users) .

> implementation of the new parameter type should go together with algorithm implementation or move it to qgsprocessingparameters.[cpp|h] ? This kind of parameter is unlikely to be used in other algorithms... and probably it should not be in public API either

This is a good question. It'd be easy to answer if it was a plugin provided parameter type, but given that it's not and the algorithm is included by default there's actually very little difference between these two choices (apart from code location). If you went with the include-alongside-algorithm approach then you'll still likely want to expose this parameter type to the parameter type registry, as otherwise users won't be able to use it as a possible input type for their own models. And as soon as you add it to the registry, it will be identical user-facing results as if it was included alongside the other parameter types in qgsprocessingparameters.h

So at this stage it doesn't really matter, and I would go with putting it wherever you are most comfortable. The qgsprocessingparameters.* files are getting very large, so that's possible one argument which would sway putting this code alongside the algorithm.

We may need to think in future about hiding some of these specific input types in the model designer list behind a "show advanced input choices" switch (@alexbruy what's your thoughts?)

> similar question for the implementation in qgis_gui library - have the widget wrapper mixed with other (generic) widgets or move it somewhere else?

Again -- it won't have any impact on the results, so it's really about just the location of the code itself. I'd be swayed by the same argument that putting this in a new file will help avoid the existing generic files becoming even larger though!

> is it a good idea to derive the internal panel widget with list of layers from QgsProcessingMultipleSelectionPanelWidget ?

I think it's a good approach. Given that this list is no longer blocking, we could add some nice UI touches in future and do things like allowing drag and dropped selections of layers from the layer tree or file system if you do this!

>returning things like number of generated tiles - return in QVariantMap from processAlgorithm() ? 

Yes please! 
> Does it need to be defined anywhere what will be returned?

Yes - you need to defined and add outputs to your algorithm in advance, via QgsProcessingAlgorithm::addOutput(). 

The philosphy is to expose as many outputs as is reasonable for an algorithm. If you don't output these values, they can't be used elsewhere in a model, which could be a real pain point for someone if they DO need logic which depends on this. On the other hand, if you expose a LOT of outputs which never get used... there's no downside! :laughing: 

> XYZ file output template - shall I use QgsProcessingParameterString or QgsProcessingParameterFileDestination or something else? It is not a proper file name really, so I have doubts...

Elsewhere we use a QgsProcessingParameterFileDestination in folder mode to allow the user to pick the main destination folder, alongside a QgsProcessingParameterExpression which lets users pick the "template" for output file names inside this folder.

> as we will be supporting writing to either MBTiles or "raw" files with XYZ template, should there be a base algorithm that does most of the work and then two derived algorithms that handles specifics of MBTiles and XYZ template?

Yes, exactly! Try to avoid the situation where you have parameters which only make sense when certain combinations of other parameter values are present. This leads to the grass provider UX issue, where there's a bunch of interrelated and mutually exclusive choices which make the algorithms very difficult for users to understand.

nyalldawson:
(PS looking forward to your newly-obtained processing skills being put to use in some sweet integrated mesh processing algorithms :wink: )

alexbruy:
> @alexbruy what's your thoughts?

I agree. We already have several specific parameter types and corresponding widgets which are not available in the modeler right now. In future it would be good to have them exposed, but probably as "advanced" section.

wonder-sk:
Thanks a lot for the hints!

> (PS looking forward to your newly-obtained processing skills being put to use in some sweet integrated mesh processing algorithms wink )

hehe, anything in particular that you are waiting for? :-)

src/analysis/processing/qgsalgorithmwritevectortiles.cpp
wonder-sk: Actually this is what I had initially. But then I have realized that parameterDefinition() uses also case insensitive matching of parameter names, and all other algorithms seem to use that function indirectly, so I thought that should be supported here too?

nyalldawson: Hm - good point!

wonder-sk:
This should be now ready for review... I didn't expect I would end up with over 1000 new lines of code... :-)

Thanks again Nyall for your earlier feedback - I hope I have incorporated your suggestions correctly!

src/gui/processing/qgsprocessingvectortilewriterlayerswidgetwrapper.h
```diff
@@ -0,0 +1,152 @@
...
+        QgsProcessingGui::WidgetType type = QgsProcessingGui::Standard, QWidget *parent = nullptr );
+
+    // QgsProcessingParameterWidgetFactoryInterface
+    QString parameterType() const override;
+    QgsAbstractProcessingParameterWidgetWrapper *createWidgetWrapper( const QgsProcessingParameterDefinition *parameter, QgsProcessingGui::WidgetType type ) override;
+//    QgsProcessingAbstractParameterDefinitionWidget *createParameterDefinitionWidget(
+//      QgsProcessingContext &context,
+//      const QgsProcessingParameterWidgetContext &widgetContext,
+//      const QgsProcessingParameterDefinition *definition = nullptr,
+//      const QgsProcessingAlgorithm *algorithm = nullptr ) override;
```
nyalldawson, line 127: remove

nyalldawson (commented):
LGTM! Just needs some tests...

src/analysis/processing/qgsalgorithmwritevectortiles.cpp
```diff
@@ -0,0 +1,200 @@
...
+  addParameter( new QgsProcessingParameterFolderDestination( QStringLiteral( "OUTPUT_DIR" ), QObject::tr( "Output directory" ) ) );
+  addParameter( new QgsProcessingParameterString( QStringLiteral( "XYZ_TEMPLATE" ), QObject::tr( "File template" ), QStringLiteral( "{z}/{x}/{y}.pbf" ) ) );
+
+  addBaseParameters();
+}
+
+void QgsWriteVectorTilesXyzAlgorithm::prepareWriter( QgsVectorTileWriter &writer, const QVariantMap &parameters, QgsProcessingContext &context, QVariantMap &outputs )
+{
+  QString outputDir = parameterAsString( parameters, QStringLiteral( "OUTPUT_DIR" ), context );
+  QString xyzTemplate = parameterAsString( parameters, QStringLiteral( "XYZ_TEMPLATE" ), context );
```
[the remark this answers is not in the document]
wonder-sk: Good idea - added checks for {x}, {y}, {z}.

wonder-sk:
The PR should be ready to go...

nyalldawson:
LGTM

PR #36105, opened 2020-04-30
Visual improvements in sublayers selection dialog

suricactus:
An annoying lack of context when adding multiple layers.

### Changes

1. Current file name is shown in the dialog title
2. Full file name (selectable!) is present in the window
3. Select/Deselect all buttons

Peek 2020-03-25 07-36

PS I was a bit slow to apply the feedback changes on this branch so the stale bot closed my previous PR, sorry for that. https://github.com/qgis/QGIS/pull/35071

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -5619,56 +5619,57 @@ bool QgisApp::askUserForZipItemLayers( const QString &path )
     return false;
   }
 
-  // if promptLayers=Load all, load all layers without prompting
-  if ( promptLayers == QgsSublayersDialog::PromptLoadAll )
+  switch ( promptLayers )
   {
-    childItems = zipItem->children();
-  }
-  // exit if promptLayers=Never
-  else if ( promptLayers == QgsSublayersDialog::PromptNever )
-  {
-    delete zipItem;
-    return false;
-  }
-  else
-  {
-    // We initialize a selection dialog and display it.
-    QgsSublayersDialog chooseSublayersDialog( QgsSublayersDialog::Vsifile, QStringLiteral( "vsi" ), this, nullptr, path );
-    QgsSublayersDialog::LayerDefinitionList layers;
+    // load all layers without prompting
+    case QgsSublayersDialog::PromptLoadAll:
+      childItems = zipItem->children();
+      break;
+    // return because we should not prompt at all
+    case QgsSublayersDialog::PromptNever:
+      delete zipItem;
+      return false;
+    // initialize a selection dialog and display it.
+    case QgsSublayersDialog::PromptAlways:
+    case QgsSublayersDialog::PromptIfNeeded:
+      QgsSublayersDialog chooseSublayersDialog( QgsSublayersDialog::Vsifile, QStringLiteral( "vsi" ), this, nullptr, path );
+      QgsSublayersDialog::LayerDefinitionList layers;
 
-    for ( int i = 0; i < zipItem->children().size(); i++ )
-    {
-      QgsDataItem *item = zipItem->children().at( i );
-      QgsLayerItem *layerItem = qobject_cast<QgsLayerItem *>( item );
-      if ( !layerItem )
-        continue;
-
-      QgsDebugMsgLevel( QStringLiteral( "item path=%1 provider=%2" ).arg( item->path(), layerItem->providerKey() ), 2 );
-
-      QgsSublayersDialog::LayerDefinition def;
-      def.layerId = i;
-      def.layerName = item->name();
-      if ( layerItem->providerKey() == QLatin1String( "gdal" ) )
+      for ( int i = 0; i < zipItem->children().size(); i++ )
       {
-        def.type = tr( "Raster" );
-      }
-      else if ( layerItem->providerKey() == QLatin1String( "ogr" ) )
-      {
-        def.type = tr( "Vector" );
-      }
-      layers << def;
-    }
+        QgsDataItem *item = zipItem->children().at( i );
+        QgsLayerItem *layerItem = qobject_cast<QgsLayerItem *>( item );
+        if ( !layerItem )
+          continue;
 
-    chooseSublayersDialog.populateLayerTable( layers );
+        QgsDebugMsgLevel( QStringLiteral( "item path=%1 provider=%2" ).arg( item->path(), layerItem->providerKey() ), 2 );
 
-    if ( chooseSublayersDialog.exec() )
-    {
-      const auto constSelection = chooseSublayersDialog.selection();
-      for ( const QgsSublayersDialog::LayerDefinition &def : constSelection )
-      {
-        childItems << zipItem->children().at( def.layerId );
+        QgsSublayersDialog::LayerDefinition def;
+        def.layerId = i;
+        def.layerName = item->name();
+        if ( layerItem->providerKey() == QLatin1String( "gdal" ) )
+        {
+          def.type = tr( "Raster" );
+        }
+        else if ( layerItem->providerKey() == QLatin1String( "ogr" ) )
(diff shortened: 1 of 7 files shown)


nyalldawson:
Nice improvements!

I'd suggest that instead of a disabled line edit, use a standard QLabel and call

setTextInteractionFlags(Qt::TextSelectableByMouse)

On it to allow selection instead.

suricactus:
Thanks Nyall, I didn't know this flag exists. My solution was copy-paste from Layer Properties -> Source.

uclaros:
@nyalldawson  How does a QLabel handle long paths that don't fit in the dialog width?

suricactus:
The dialog gets resized in width to fit the path. I guess this should be fine even for low resolution screens.

uclaros:
> The dialog gets resized in width to fit the path. I guess this should be fine even for low resolution screens.

I think you underestimate peoples' ability to craft deep and long paths!
Resizing the dialog is a bit awkward, and resizing it past the screen width should not be an option; I'd stay with the QLineEdit. What are the drawbacks?

suricactus:
> I think you underestimate peoples' ability to craft deep and long paths!
:) It's is healthy to keep them short.

Adding "wordWrap" solves the potential problem:
image

> What are the drawbacks?

Some custom not themed QSS and a bit awkward visual style.

I vote for QLabel, it just looks better.

uclaros:
Beauty is in the eye of the beholder, but I haven't seen this approach on any file manager path/location widget.
Long paths are a thing. It's not always your data and some companies try to organize theirs with large descriptive paths.

src/gui/qgssublayersdialog.h
```diff
@@ -59,10 +88,21 @@ class GUI_EXPORT QgsSublayersDialog : public QDialog, private Ui::QgsSublayersDi
     typedef QList<QgsSublayersDialog::LayerDefinition> LayerDefinitionList;
 
     //! Constructor for QgsSublayersDialog
+
+    /**
+     * Construct a new QgsSublayersDialog object - a dialog to select which sub layers to be imported from a file (e.g. from geopackage or zipfile)
+     *
+     * @param providerType provider type
```
nyalldawson, line 95: You need to use "\param" syntax instead of "@param", so that the sipify script can do its magic and create the correct Python docstring.

src/gui/qgssublayersdialog.h
```diff
@@ -59,10 +88,21 @@ class GUI_EXPORT QgsSublayersDialog : public QDialog, private Ui::QgsSublayersDi
     typedef QList<QgsSublayersDialog::LayerDefinition> LayerDefinitionList;
 
     //! Constructor for QgsSublayersDialog
+
+    /**
+     * Construct a new QgsSublayersDialog object - a dialog to select which sub layers to be imported from a file (e.g. from geopackage or zipfile)
```
nyalldawson, line 93: Note that this dialog can be used for non-file based sources too.
```suggestion
     * Construct a new QgsSublayersDialog object - a dialog to select which sub layers to be imported from a data source (e.g. from geopackage or zipfile)
```

src/gui/qgssublayersdialog.h
```diff
@@ -59,10 +88,21 @@ class GUI_EXPORT QgsSublayersDialog : public QDialog, private Ui::QgsSublayersDi
...
+     *
+     * @param providerType provider type
+     * @param name provider type name
+     * @param parent parent widget of the dialog
+     * @param fl window flags
+     * @param providerSource full file name
```
nyalldawson, line 99:
```suggestion
     * \param uri data source URI
```

src/app/qgisapp.cpp
```diff
@@ -5620,20 +5620,20 @@ bool QgisApp::askUserForZipItemLayers( const QString &path )
   }
 
   // if promptLayers=Load all, load all layers without prompting
-  if ( promptLayers == 3 )
+  if ( promptLayers == QgsSublayersDialog::PromptLoadAll )
```
nyalldawson, line 5623: using a "switch" statement here would be cleaner

src/app/qgsoptions.cpp
```diff
@@ -419,17 +420,12 @@ QgsOptions::QgsOptions( QWidget *parent, Qt::WindowFlags fl, const QList<QgsOpti
...
-  // 3 = Load all -> never prompt, but load all sublayers
-  cmbPromptRasterSublayers->clear();
-  cmbPromptRasterSublayers->addItem( tr( "Always" ) );
-  cmbPromptRasterSublayers->addItem( tr( "If needed" ) ); //this means, prompt if there are sublayers but no band in the main dataset
-  cmbPromptRasterSublayers->addItem( tr( "Never" ) );
-  cmbPromptRasterSublayers->addItem( tr( "Load all" ) );
-  cmbPromptRasterSublayers->setCurrentIndex( mSettings->value( QStringLiteral( "/qgis/promptForRasterSublayers" ), 0 ).toInt() );
+  cmbPromptSublayers->clear();
+  cmbPromptSublayers->addItem( tr( "Always" ), QgsSublayersDialog::PromptAlways );
+  cmbPromptSublayers->addItem( tr( "If needed" ), QgsSublayersDialog::PromptIfNeeded ); //this means, prompt if there are sublayers but no band in the main dataset
```
nyalldawson, line 425:
```suggestion
  cmbPromptSublayers->addItem( tr( "If Needed" ), QgsSublayersDialog::PromptIfNeeded ); //this means, prompt if there are sublayers but no band in the main dataset
```

src/app/qgsoptions.cpp
```diff
@@ -419,17 +420,12 @@ QgsOptions::QgsOptions( QWidget *parent, Qt::WindowFlags fl, const QList<QgsOpti
...
+  cmbPromptSublayers->addItem( tr( "Never" ), QgsSublayersDialog::PromptNever );
+  cmbPromptSublayers->addItem( tr( "Load all" ), QgsSublayersDialog::PromptLoadAll ); // check if this is true
```
nyalldawson, line 427:
```suggestion
  cmbPromptSublayers->addItem( tr( "Load All" ), QgsSublayersDialog::PromptLoadAll ); // check if this is true
```

src/gui/qgssublayersdialog.cpp
```diff
@@ -42,50 +43,62 @@ class SubLayerItem : public QTreeWidgetItem
...
     default:
-      setWindowTitle( tr( "Select Layers to Add…" ) );
+      title = tr( "Select Layers to Add…" );
       layersTable->setHeaderLabels( QStringList() << tr( "Layer ID" ) << tr( "Layer name" )
                                     << tr( "Type" ) );
       mShowType = true;
-      break;
   }
 
+  QString fileFullPath = providerType == QgsSublayersDialog::Vsifile
```
nyalldawson, line 83: Given that the argument is actually the dataSourceUri, I would call decodeUri here regardless of the providerType and if if contains "path" then use this.

nyalldawson:
Regarding the label length debate: my vote would go to the word wrap approach instead of the lineedit. A disabled line edit for this pupose is really abusing the widget and does cause a lot of theming and interaction issues.

src/gui/qgssublayersdialog.h
suricactus: Done!

src/app/qgisapp.cpp
suricactus: Agree!

src/gui/qgssublayersdialog.cpp
suricactus: I agree and did it the way you said.

nyalldawson (approved):
Let's give it a shot!

Issue #21858, opened 2015-11-15
Labelling - place label horizontally outside the polygon


qgib:
Author Name: **Darek Bobak** (Darek Bobak)
Original Redmine Issue: [13834](https://issues.qgis.org/issues/13834)

Redmine category:labelling

---
I'm working in archaeology. And I often have to label polygon layer which has polygons not connected to each other. The best way to label it would be to place labels horizontally, just by the edge of the polygon but OUTSIDE of the polygon. I'm attaching the sample from other software.

---

- [labelling.png](https://issues.qgis.org/attachments/download/9317/labelling.png) (Darek Bobak)

basilrabi:
@gioman , I suggest to close this issue since #36106 seems to address this.

PR #36106, opened 2020-04-30
Automatic placement of labels outside polygons

nyalldawson:
This PR adds the ability to automatically place labels outside of polygon features. 

This is exposed in a number of ways:

1. A new placement mode which always places polygon labels for the layer outside the features:
image

image

2. When using other polygon placement modes, a new checkbox "Allow placing labels outside of polygons" can be checked to allow labels to be placed outside when it wasn't possible to place them inside the polygon:

image

image

The "allow placing labels outside" checkbox can be data defined, allowing users to either allow outside labels, prevent outside labels, or force outside labels on a feature-by-feature basis.

The placement algorithm is based on a modification of Rylov & Reimer (2016) "*A practical algorithm for the external annotation of area features*". While Rylov & Reimer propose a horizontal sweep-line based approach for selection of candidate label points, in practice I found that this delivered inferior results (and was slower) then just traversing the exterior ring of the polygon and generating candidate point at regular intervals (especially for narrow, nearly horizontal polygons).

As shown in the islands screenshot above, the outside placement mode works as expected with the "Follow label placement" multiline alignment mode.

Sponsored by the Swiss QGIS user group


src/core/labeling/qgspallabeling.cpp
```diff
@@ -1966,10 +1971,54 @@ void QgsPalLayerSettings::registerFeature( const QgsFeature &f, QgsRenderContext
...
+        }
+        else if ( value.compare( QLatin1String( "no" ), Qt::CaseInsensitive ) == 0 )
+        {
+          // block outside placement
+          polygonPlacement &= ~static_cast< int >( QgsLabeling::PolygonPlacementFlag::AllowPlacementOutsideOfPolygon );
+        }
+      }
+      else
+      {
+        if ( dataDefinedOutside.toBool() )
```
nirvn, line 2003: I like the dual handling of string and bool here, nice.

src/core/labeling/qgspallabeling.h
```diff
@@ -441,6 +443,7 @@ class CORE_EXPORT QgsPalLayerSettings
       LinePlacementOptions = 99, //!< Line placement flags
       OverrunDistance = 102, //!< Distance which labels can extend past either end of linear features
       LabelAllParts = 103, //!< Whether all parts of multi-part features should be labeled
+      PolygonLabelOutside = 109, //!< Whether labels outside a polygon feature are permitted, or should be forced
```
nirvn, line 446: Do we want to add a since 3.14 here too or explicit enough already?

src/core/labeling/qgspallabeling.h
```diff
@@ -645,6 +648,20 @@ class CORE_EXPORT QgsPalLayerSettings
...
+     * Returns the polygon placement flags, which dictate how polygon labels can be placed.
+     *
+     * \see setPolygonPlacementFlags()
+     */
+    QgsLabeling::PolygonPlacementFlags polygonPlacementFlags() const { return mPolygonPlacementFlags; }
+
+    /**
+     * Sets the polygon placement \a flags, which dictate how polygon labels can be placed.
+     *
+     * \see polygonPlacementFlags()
```
nirvn, line 661: \since QGIS 3.14

src/gui/labeling/qgslabelinggui.cpp
```diff
@@ -793,15 +811,15 @@ void QgsLabelingGui::updateGeometryTypeBasedWidgets()
       qFatal( "unknown geometry type unexpected" );
   }
 
-  if ( geometryType == QgsWkbTypes::PointGeometry )
+  if ( geometryType == QgsWkbTypes::PointGeometry || geometryType == QgsWkbTypes::PolygonGeometry )
   {
-    // follow placement alignment is only valid for point layers
-    if ( mFontMultiLineAlignComboBox->findText( tr( "Follow label placement" ) ) == -1 )
-      mFontMultiLineAlignComboBox->addItem( tr( "Follow label placement" ) );
+    // follow placement alignment is only valid for point or polygon layers
```
nirvn, line 816: Woupidou! I was always too lazy to enable this for polygons, thanks for doing it.

src/gui/labeling/qgslabelinggui.cpp
```diff
@@ -793,15 +811,15 @@ void QgsLabelingGui::updateGeometryTypeBasedWidgets()
...
+    if ( mFontMultiLineAlignComboBox->findData( QgsPalLayerSettings::MultiFollowPlacement ) == -1 )
+      mFontMultiLineAlignComboBox->addItem( tr( "Follow Label Placement" ), QgsPalLayerSettings::MultiFollowPlacement );
   }
   else
   {
-    int idx = mFontMultiLineAlignComboBox->findText( tr( "Follow label placement" ) );
+    int idx = mFontMultiLineAlignComboBox->findData( QgsPalLayerSettings::MultiFollowPlacement );
```
nirvn, line 822: Nice cleanup.

src/ui/qgstextformatwidgetbase.ui
```diff
@@ -4547,22 +4548,61 @@ font-style: italic;</string>
                           <number>0</number>
                          </property>
                          <item row="0" column="0">
-                          <widget class="QLabel" name="mCentroidLabel">
+                          <widget class="QLabel" name="label_14">
```
nirvn, line 4551: Remove that label and just have the checkbox speak for itself. It'll help avoid creating to large a minimum width situation.

src/ui/qgstextformatwidgetbase.ui
```diff
@@ -4336,6 +4321,33 @@ font-style: italic;</string>
...
+                              <widget class="QRadioButton" name="radPolygonFree">
+                               <property name="text">
+                                <string>Free (angled)</string>
+                               </property>
+                              </widget>
+                             </item>
+                             <item row="0" column="2">
+                              <widget class="QRadioButton" name="radPolygonOutside">
+                               <property name="text">
+                                <string>Outside polygons</string>
```
nirvn, line 4334: If possible, I think it'd be better for this to site on a 4th row (i.e. row="3" column="0"). Right now, the minimum width of a three column layout with long strings (offset from centroid + using perimeter (curved) + outside polygons) makes it  too wide to play nicely with the style dock.

src/ui/qgstextformatwidgetbase.ui
```diff
@@ -4547,22 +4548,61 @@ font-style: italic;</string>
...
-                           <property name="alignment">
-                            <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
+                            <string>Allowed positions</string>
                            </property>
                           </widget>
                          </item>
                          <item row="0" column="1">
+                          <layout class="QHBoxLayout" name="horizontalLayout_11">
+                           <item>
+                            <widget class="QCheckBox" name="mCheckAllowLabelsOutsidePolygons">
```
nirvn, line 4566: +tabstop

nyalldawson:
Thanks @nirvn! all addressed...

3nids:
nice!!!

would it make sense to get an expression variable which determines if the label is inside or outside the polygon? that could be used to set the label font color.

nyalldawson:
@3nids possibly... But it gets tricky. Because if you used that variable to try to control label size, then you end up with a nice circular reference 😳

3nids:
> a nice circular reference  

indeed! Too bad

nyalldawson:
It's definitely something to consider in future. We'd just have to correctly control the expression contexts for different widgets in the labelling panel to selectively expose it only for options which won't impact the placement decisions. And that's non trivial

andreasneumann:
> nice!!!
> 
> would it make sense to get an expression variable which determines if the label is inside or outside the polygon? that could be used to set the label font color.

Good idea! This would a nice enhancement. If it can be restricted to certain properties as @nyalldawson suggested.

alexbruy:
This will fix #21858.

nirvn (approved):
Looking good, thanks for addressing comments.

Travis failure seems to be related to the change from enum class to enum.

PR #36108, opened 2020-04-30
Avoid divide by zero number error 

Samweli:
Removes the possibly of divide by zero error in one of WMS-T temporal range match modes logic.


nyalldawson (approved):

Samweli:
thanks @nyalldawson

PR #36123, opened 2020-05-01
remove relation editor on double clik in form editor

3nids:
there is now the panel on the right for this


signedav (approved):
Looks good. Approve.

Btw. I think the "Show Label" has no effect on the form. But I don't know if it used to work or has even been integrated once. Maybe it's not something with this configuration part but on the widget side instead. I mention it just in case it's an easy fix for you. Should not be part of this PR.

3nids:
> Btw. I think the "Show Label" has no effect on the form. 

indeed, since the label is placed as the groupbox title.
what do you suggest? just drop the config of the label?

PR #36129, opened 2020-05-01
Enables Postgresql service based connections without database name

### LINKED ISSUE #36127: Unable to add or change Postgresql connection with empty database
**Describe the bug**

When using [lippq services](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS), there is no need to enter a database name. Right now, I'm not able to create database connections or edit existing ones without a database name.

edit_postgis_connection

**QGIS and OS versions**

Current master (compiled on 2020-04-29)

jgrocha:
## Description

Postgresql service based connections can have an empty database name, if a service name is provided. 

Fix #36127

### Additional tiny UX improvement

When the user edits the connection, the dialogue title is set accordingly. The connection can be changed from the **Browser** or from the **Data Source Manager**.

Dialogue title when editing:

improved_connection_title

### DIFF
--- a/src/providers/postgres/qgspgnewconnection.cpp
+++ b/src/providers/postgres/qgspgnewconnection.cpp
@@ -249,7 +249,7 @@ void QgsPgNewConnection::showHelp()
 void QgsPgNewConnection::updateOkButtonState()
 {
   bool enabled = !txtName->text().isEmpty() && (
-                   ( !txtService->text().isEmpty() && txtDatabase->text().isEmpty() ) ||
+                   !txtService->text().isEmpty() ||
                    ( !txtHost->text().isEmpty() && !txtPort->text().isEmpty() && !txtDatabase->text().isEmpty() ) );
   buttonBox->button( QDialogButtonBox::Ok )->setEnabled( enabled );
 }
--- a/src/providers/postgres/qgspgnewconnection.cpp
+++ b/src/providers/postgres/qgspgnewconnection.cpp
@@ -249,7 +249,7 @@ void QgsPgNewConnection::showHelp()
 void QgsPgNewConnection::updateOkButtonState()
 {
   bool enabled = !txtName->text().isEmpty() && (
-                   ( !txtService->text().isEmpty() && !txtDatabase->text().isEmpty() ) ||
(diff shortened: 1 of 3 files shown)


src/providers/postgres/qgspgnewconnection.cpp
```diff
@@ -249,7 +249,7 @@ void QgsPgNewConnection::showHelp()
 void QgsPgNewConnection::updateOkButtonState()
 {
   bool enabled = !txtName->text().isEmpty() && (
-                   ( !txtService->text().isEmpty() && !txtDatabase->text().isEmpty() ) ||
+                   ( !txtService->text().isEmpty() && txtDatabase->text().isEmpty() ) ||
```
alexbruy, line 252: If database is not required we can skip checking `txtDatabase` completely.

alexbruy (changes requested):
Looks good, only minor comment to address.

src/providers/postgres/qgspgnewconnection.cpp
jgrocha: Yes @alexbruy, you could skip the `txtDatabase` check. 

I would vote to keep the condition as it is, instead of a "don't care" about the database name when the users uses libpq services. It can be misleading to see in the documentation or in web tutorials both the service and the database filled even if we don't care about the database name.

With this condition in place, the user knows that it makes no sense to enter a database name. Since this database name is not used by libqp, we force the user to keep it empty.

But I don't have a strong opinion on this. Dropping the condition as you suggested is our current behaviour. What do you prefer?

alexbruy: As *all* parameters saved to the settings, even if they are not used, I think it is safe to drop database name from the condition.

jgrocha: Done!

alexbruy (approved):

nyalldawson:
This isn't quite right still -- I have services locally which I used to be able to connect to by only entering the database name. No host, port or service name required.

nyalldawson:
I'll fix in a follow up PR

jgrocha:
Wow! Sorry, I didn't know it was possible. It is, in fact. Thanks for fixing.

PR #36130, opened 2020-05-01
[feature][processing] Complete random raster algorithm collection #2

root676:
## Description

This PR refactors and completes the recently added work on **random number raster layer creation algorithms** (see #35835) and reworks the single algorithm implementation proposed in [#36065](https://github.com/qgis/QGIS/pull/36065) to a **base-algorithm solution** which avoids duplicate code.

With the new algorithms aimed towards distribution based random number creation, QGIS reaches to the same level of functionality as current ArcGIS random raster creation tools. In total, the PR adds the following algorithms (normal and uniform raster layer creation algs are refactored to be in line the new naming scheme):

1. Create random raster layer (binomial distribution)
2. Create random raster layer (exponential distribution)
3. Create random raster layer (gamma distribution)
4. Create random raster layer (negative binomial distribution)
5. Create random raster layer (normal distribution)
6. Create random raster layer (poisson distribution)
7. Create random raster layer (uniform distribution)

Tests have been added for the algorithms data type choices.
I'm looking forward to comments/questions/reviews/etc.!


src/analysis/processing/qgsalgorithmrandomraster.h
```diff
@@ -27,36 +27,238 @@
...
+    QStringList tags() const override;
+    QString shortHelpString() const override;
+    QgsRandomUniformRasterAlgorithm *createInstance() const override SIP_FACTORY;
+
+  protected:
+    void addAlgorithmParams() override;
+    Qgis::DataType getRasterDataType( int typeId ) override;
+    bool prepareRandomParameters( const QVariantMap &parameters, QgsProcessingContext &context ) override;
+    long generateRandomLongValue( std::mt19937 &mersenneTwister ) override;
+    double generateRandomDoubleValue( std::mt19937 &mersenneTwister ) override;
```
nyalldawson, line 89: There may potentially be a small performance boost by swapping "override" for "final" in these classes.

src/analysis/processing/qgsalgorithmrandomraster.cpp
```diff
@@ -24,34 +24,184 @@
...
+{
+  Q_UNUSED( feedback );
+  mCrs = parameterAsCrs( parameters, QStringLiteral( "TARGET_CRS" ), context );
+  mExtent = parameterAsExtent( parameters, QStringLiteral( "EXTENT" ), context, mCrs );
+  mPixelSize = parameterAsDouble( parameters, QStringLiteral( "PIXEL_SIZE" ), context );
+
+  int typeId = parameterAsInt( parameters, QStringLiteral( "OUTPUT_TYPE" ), context );
+  //prepare specific parameters
+  mRasterDataType = getRasterDataType( typeId );
+  prepareRandomParameters( parameters, context );
```
nyalldawson, line 62: I think this should be moved to processAlgorithm, to ensure it's done in the same thread as the distributions are used from.

nyalldawson (commented):
Fantastic! Thanks for the rework, this is much less maintenance now :smile:

src/analysis/processing/qgsalgorithmrandomraster.cpp
root676: Ok, np! The `prepareAlgorithm()` method is a mystery to me - on the one hand it is used to prepare the algorithm execution, on the other hand it is barely used to do so in most algorithms (except the ones written by me so far^^). I know that `processAlgorithm()` runs in a separate thread - but why is `prepareAlgorithm()` even needed if all parameter reading is done within `processAlgorithm()`?

root676: Haha - on the second read - ignore my question. I understand your concern: Thread safety and distribution objects. :smile:

nyalldawson (approved):

nyalldawson:
Non-obvious fix for the spelling failure: change https://github.com/qgis/QGIS/blob/master/scripts/spell_check/spelling.dat#L3259 to

    generater:generator:*

PR #36136, opened 2020-05-02
Allow placing manual column breaks in legends

nyalldawson:
Adds the option when configuring a legend item to place a column break before the item, causing it to be placed into a new column

This allows user control over the column content, for cases when the automatic column generation doesn't result in the desired results


nirvn (approved):
I see your putting your new legend item sub-panel to more go use :)

PR #36137, opened 2020-05-02
[processing] Set distanceArea CRS properly in SumLineLengths algorithm

### LINKED ISSUE #36135: [processing] SumLineLengths algorithm returns a wrong (!) length sum
Sample GPKG data: [sumlinelenghts.zip](https://github.com/qgis/QGIS/files/4566765/sumlinelenghts.zip)
 + CRS: EPSG:3117 (projected CRS)
 + Contains a polygon and a line layer.

Expected length sum: **51.168**,291572545866 [m]
Obtained length sum: **1.243'634.138**,1295323 [m] !!!

gacarrillor:
Currently, the algorithm sets a private `mCrs` object in `outputCrs()` and uses it in `prepareAlgorithm()`. However, `prepareAlgorithm()` is called before than `outputCrs()`, so the `mCrs` is not correctly set when used to set the `QgsDistanceArea` CRS for the calculations.

This PR fixes it.

Fix #36135


src/analysis/processing/qgsalgorithmsumlinelength.cpp
```diff
@@ -131,6 +125,7 @@ bool QgsSumLineLengthAlgorithm::prepareAlgorithm( const QVariantMap &parameters,
   {
     mDa.setEllipsoid( context.project()->ellipsoid() );
   }
+  mCrs = mLinesSource->sourceCrs();
```
nyalldawson, line 128: This isn't correct -- we need to use the polygon's source crs here, which may differ from the line's source crs. 

I think the solution will be to defer creation of mDa until the first call to processFeature, at which stage we can be certain that mCrs has correctly been set to match the input polygon's crs.

nyalldawson (changes requested):
see above

src/analysis/processing/qgsalgorithmsumlinelength.cpp
gacarrillor: Oops! That's right, it should be polygons' Crs. 

Well, a couple of possible solutions here:

1. Inside `processFeature()`, check if `mDa` is set already, if not, set it. (Is this what you suggested in your comment?)
2. Inside `prepareAlgorithm()`, get the polygon (`INPUT`) parameter and use its Crs to set mDa's source Crs. I think it's more natural to have everything set in this method.

nyalldawson (approved):

PR #36138, opened 2020-05-02
Improve temporal controller frame count

Samweli:
Improves logic for calculating frame count in temporal controller.

Currently if the start and end times are (date) 12:00 hrs to (date) 14:00 hrs with 1.5 hrs time step the temporal controller will generate following frames,

1st frame  12:00 hrs - 13:30 hrs
2nd frame 13:30 hrs - 14:00 hrs
**3rd(last) frame 15:00 hrs - 14:00 hrs**

The 3rd frame should haven't been generated, this PR provides fix for this issue, the resulting frames will be 
1st frame 12:00 hrs - 13:30 hrs
2nd(last) frame 13:30 hrs - 14:00 hrs

### DIFF
--- a/src/core/qgstemporalnavigationobject.cpp
+++ b/src/core/qgstemporalnavigationobject.cpp
@@ -191,7 +191,7 @@ void QgsTemporalNavigationObject::skipToEnd()
 long long QgsTemporalNavigationObject::totalFrameCount()
 {
   QgsInterval totalAnimationLength = mTemporalExtents.end() - mTemporalExtents.begin();
-  return std::ceil( totalAnimationLength.seconds() / mFrameDuration.seconds() ) + 1;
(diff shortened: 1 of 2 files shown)


nyalldawson:
Could we get a test covering this situation?

Samweli:
> Could we get a test covering this situation?

sure

Samweli:
> Could we get a test covering this situation?

Actually @nyalldawson , I think this test here https://github.com/qgis/QGIS/blob/89fe6347ae764fa6e7597d148a43b30b8747f583/tests/src/core/testqgstemporalnavigationobject.cpp#L116 is enough, I just need to update with a right frame number .

Samweli:
Thanks @nyalldawson!

PR #36141, opened 2020-05-02
Add a broader restorer mechanism for WMS 

pblottiere:
## Description

This PR adds a broader restorer mechanism for the WMS service (cf the discussion on https://github.com/qgis/QGIS/pull/36007).

This way we can easily add specific restorers without updating the renderer itself.

### DIFF
--- a/src/server/services/wms/qgswmsrestorer.h
+++ b/src/server/services/wms/qgswmsrestorer.h
@@ -65,12 +65,27 @@ class QgsLayerRestorer
     QMap<QgsMapLayer *, QgsLayerSettings> mLayerSettings;
 };
 
-namespace QgsWms {
+namespace QgsWms
+{
+  /**
+   * \ingroup server
+   * RAII class to restore the rendering context configuration on destruction
+   * \since QGIS 3.14
+   */
   class QgsWmsRestorer
   {
     public:
 
+      /**
+       * Constructor for QgsWmsRestorer.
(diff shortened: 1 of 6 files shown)


src/server/services/wms/qgswmsrestorer.h
```diff
@@ -64,4 +65,32 @@ class QgsLayerRestorer
...
+      QgsWmsRestorer( const QgsWmsRenderContext &context );
+
+      /**
+       * Default destructor.
+       */
+      ~QgsWmsRestorer() = default;
+
+    private:
+
+      std::unique_ptr<QgsLayerRestorer> mLayerRestorer;
```
nyalldawson, line 92: Does this need to be a unique ptr? It could just be a value instead.

pblottiere: The `QgsLayerRestorer` constructor needs a list of layers.

nyalldawson: Yes but if you could initialise it:

QgsWmsRestorer::QgsWmsRestorer( const QgsWmsRenderContext &context )
    : mLayerRestorer( context.layers() )
{}

pblottiere: Oh yes of course, silly me :upside_down_face:.

Thanks Nyall, I'm going to do that.

PR #36143, opened 2020-05-02
Add expression variables for layout item position and size.

jdugge:
Closes #25768

## Description

Add the following expression variables:

* `item_left`: horizontal position of left edge of item
* `item_top`: horizontal position of top edge of item
* `item_width`: width of item
* `item_height`: height of item
* `item_x`: horizontal position of item reference point
* `item_y`: vertical position of item reference point
* `item_reference_point`: integer representing item reference point


nyalldawson:
There's a reason these have been previously omitted -- there's no dependency resolution happening with evaluation of layout properties. So if you attempt to use these variables to move or resize other items then the results will be unpredictable. They'll work or not randomly, based on the order in which the other item's positions are calculated. 

I'm personally -1 to including these before we get dependency resolution, as we just won't be able to deliver stable results until then.

jdugge:
Ah yes, good point.

PR #36144, opened 2020-05-02
Allow relative paths to vector tile layers in projects

### LINKED ISSUE #36023: File based vector tiles and relative path
The path for file based vector tiles are always absolute. To reproduce:

1- Start a new QGIS
2- Drag and drop an vector tiles mbtiles
3- Save the project (relative path) to the same location as your vector tile
4- Move the project and vector tile to another location
5- Delete the vector tile and project from the original location
6- Try to open the project
7- It shows you missing vector tile layer

wonder-sk:
This fixes #36023 and adds tests.


wonder-sk:
Unrelated test failure in PyQgsServerWMSGetPrint

PR #36148, opened 2020-05-03
Constrain the expression feature preview widget size and avoid it expanding

DelazJ:
moving from
image

to 
image

### DIFF
--- a/src/core/qgsproperty.cpp
+++ b/src/core/qgsproperty.cpp
@@ -88,7 +88,7 @@ QgsPropertyDefinition::QgsPropertyDefinition( const QString &name, const QString
 
     case ColorWithAlpha:
       mTypes = DataTypeString;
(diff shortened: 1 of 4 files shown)


nyalldawson:
Looking at the second screenshot, there's still a lot of wasted space:
- to the right of the < and > buttons
- alignment of "feature" and "preview" labels isn't in line with the widgets above it
- wasted space above "feature" label
- lots of wasted space below the preview

DelazJ:
@nyalldawson Should all be fixed now
![expressionBuilderPreview](https://user-images.githubusercontent.com/7983394/80947410-6b4b8400-8df0-11ea-8ec8-9bf7ca3ced24.g
I also replace tooltips: previously, preview behavior was shown when using data-defined button. Moved instead to the preview label. worth a check...

nyalldawson:
That looks much better, thanks!

PR #36153, opened 2020-05-03
Server GetLegendGraphics add RULELABEL AUTO option

### LINKED ISSUE #36140: QGIS Server GetLegendGraphics forcing RULELABEL=TRUE where not needed (set)
**Describe the bug**
PR #32383 introduced displaying legend labels for all cases even if RULELABEL parameter was not set. 
For single symbols that means duplicated legend label with layer title. If you know your layer style type then you can call its legend with RULELABEL on/off and that's fine. Problem are web clients which from GetProjectSettings do not have information about that and do not know how to adjust RULELABEL parameter. 
That was not necessary in 3.4 and can be seen on pictures below (examples from Extended QWC):
legenda 3 4 legenda 3 10

**QGIS and OS versions**

QGIS Server 3.10.5 and current master

**Additional context**

Working on PR myself, will commit it shortly.

uprel:
## Description

This pull fixes #36140 and is done as a respond to pull  #32383.

It adds option AUTO to RULELABEL parameter which handles display of legend labels regarding legend type. For single symbols labels are removed, because label is duplication of layer title and for other legend types, labels are kept.

This is useful in situations where you do not want to get duplicated layer title and label on legend and you do not know legend type (like web clients).

In Server 3.4 this logic was working if RULELABEL parameter was not set and LAYERTITLE was FALSE. With adding AUTO value you get same behaviour as before in 3.4 without RULELABEL parameter.

## Backporting
Should be backported to 3.10 branch!


m-kuhn:
@signedav @elpaso @rldhont can you have a look at this?

@uprel can you adjust the test (rather than removing it) to make sure we are protected against a regression.

uprel:
Instead of removed test I added two additional in case of rulelabel not set for single symbol and for rule based symbol as suggested by @m-kuhn

signedav:
Thanks  @uprel 
I understand that it's not nice when existing clients do not work like they did before. PR #32383 fixed a wrong behavior, and this leaded to a new behavior on unset parameters that diverts from the behavior the web clients are "used to."

But I still think it's not a nice behavior, that on an unset `LAYERTITLE` it takes it as true and on unset `RULELABLE` it should take it as false.

uprel:
I agree @signedav, I will remove check for LAYERTITLE in the part for removing labels. That will be more clear, meaning that default legend without LAYERTITLE and RULELABEL set will be for single symbols:
helloworld
and for other types (example of rule based legend):
lights_status
Will commit changes and adjust tests if necessary.

signedav:
But doesn't this mean that it takes in one situation `true` on an unset `RULELABEL` (rule based legend) and on the other situation `false` (single symbol)? This is more confusing that the original suggestion, no?

Hm, I don't know what to suggest. It seems, that we have four unsatisfying approaches:
1. take `true` for unset `LAYERTITLE` and `RULELABEL` -> what it did before this PR. This is a consistent behavior but old clients expected it else (because back then `RULELABEL` did not work well)
2. take `false` for unset  `LAYERTITLE` and `RULELABEL` -> what would be worse, since unset `LAYERTITLE` has always taken `true` before. I think we all agree, that this is the wrong approach.
3. take `true` on unset `LAYERTITLE` and `false` on unset `RULELABEL` -> what this PR originally suggested, but I think it's confusing
4. take `true` on unset `LAYERTITLE` and `false` on unset `RULELABEL` when it's a single symbol but when it's a rule based symbology take `true` -> what is confusing as well.

What do you think @m-kuhn @elpaso @rldhont?

uprel:
> But doesn't this mean that it takes in one situation `true` on an unset `RULELABEL` (rule based legend) and on the other situation `false` (single symbol)? This is more confusing that the original suggestion, no?
> 
Exactly, that is whole idea of this PR. Keep in mind that this is how it worked in 3.4 and earlier.

I think this looks now more complicated then necessary. If I sum it up:

1. If you ask for RULELABEL/LAYERTITLE you will get it in all cases independent of each other (that still needs to be commited upon first review comment)
2. We are defining here how default legend will look if this parameters are not present.
3. If LAYERTITLE is not present it is always TRUE in all cases.
4. So main (and only) question here is how to set RULELABEL if it is not set by request:

A (what I suggest and how it worked already in 3.4 and before):
helloworld

B (now)
helloworld2

Which is better?

signedav:
Thanks for summarizing again @uprel 

I prefer B, because it's consistent and in my opinion what the user expects: it's the same behavior for `RULELABEL` like it is for `LAYERTITLE` -> when unset, it's displayed (true).

I understand the request to have it like in 3.4, but consider that on 3.4 the behavior have only been like this, because `RULELABEL=true` did not work.

What do the others think?

m-kuhn:
@uprel what you are proposing is that `RULELABEL` unset does not match neither directly to true or false, but rather to a different `AUTO` or `SMART` mode? Maybe it would be clearer to add this as an explicit new parameter value and document that it "defaults to AUTO if not set"?

uprel:
> @uprel what you are proposing is that `RULELABEL` unset does not match neither directly to true or false, but rather to a different `AUTO` or `SMART` mode? Maybe it would be clearer to add this as an explicit new parameter value and document that it "defaults to AUTO if not set"?

OK, that sounds fine. Having third option to set labels will work. Then we can leave default option to TRUE if not set, also lot of tests are expecting this now, so it's less to modify.

I will prepare and commit this way.

m-kuhn:
Thanks for the feedback.
Let's wait for @signedav 's (or others) opinion before investing the work, but good to see a path forward.

signedav:
You mean that unset is like `RULELABEL=auto`? And then it takes it normally as `true`, except if it's a single symbol, then it takes it as `false`? If so, then fair enough.

uprel:
If we add AUTO option then we can use it in web client so issue on this will be fixed. In that case I think I rather leave unset behaviour to TRUE and not AUTO, because that's how it is now (image B above). I know I wanted to change that in first place, but like said having third option is enough.
But it needs to go to 3.10 and into documentation.

uprel:
According to comments above I have modified code, tests and also changed title and description of this pr.

src/server/services/wms/qgswmsgetlegendgraphics.cpp
```diff
@@ -284,10 +285,11 @@ namespace QgsWms
             legendNode->setUserLabel( QStringLiteral( " " ) );
           }
         }
-        else if ( !parameters.layerTitleAsBool() && model->layerLegendNodes( nodeLayer, true ).count() )
+        else if ( ruleLabel.compare( QStringLiteral( "AUTO" ), Qt::CaseInsensitive ) == 0 )
```
m-kuhn, line 288: shouldn't this be added as a new condition in here ( `( !parameters.layer ... ) || ruleLabel.compare( "AUTO" ... )`?

signedav, line 288: Does this work when `LAYERTITLE=true&RULELABEL=auto` like it should or does it remove the rule labels as well? Shouldn't it check the LAYERTITLE parameter, like `!parameters.layerTitleAsBool() && ( model->layerLegendNodes( nodeLayer, true ).count() || ruleLabel.compare( QStringLiteral( "AUTO" ), Qt::CaseInsensitive ) == 0  ) )`?

signedav: Oh. Sorry. I haven't received @m-kuhn comment on writing this one. If you care about his input @uprel you can ignore this one...

uprel: LAYERTITLE is already handled in code, here and it's not needed again.
```
// layer titles - hidden or not
QgsLegendRenderer::setNodeLegendStyle( nodeLayer, parameters.layerTitleAsBool() ? QgsLegendStyle::Subgroup : QgsLegendStyle::Hidden );
```
        

RULELABEL is independent of LAYERTITLE parameter, so it's the case with AUTO value. Logic if labels are visible or not, does, not depend on value of LAYERTITLE parameter. I think this is more clear now.

uprel: OK, I answered above.

signedav: I'm confused. Maybe you can illuminate me. Otherwise I need more time to check it out deeper.

Currently I don't understand when a node is embedded in parent and when it's not. Apparently a single symbol node is embedded and the categories are not.

On the check I did in the mentioned PR, I checked if there are **un**embedded nodes and if so I unembedded the embedded ones (like the single symbol node). I don't really remember everything, but I guess I wanted to avoid that anything gets unembedded, if there is no other child node in there but the single symbol child node.  So when does this happen anyway? What are the embedded nodes beside the single symbol. Are there any at all?

If there are no cases of other embedded nodes, then this unembedding never happens, since there is the possibility of a single symbol (and then there are no unembedded nodes) or of not single symbol (then there might be unembedded nodes but no embedded ones). So there is nothing to unembedd anyway. What would make this clause to unneeded. And with your integration you would reintroduce it for the `RULELABEL=AUTO`, what would be fair enough then.

But if there are other embedded nodes than the single symbol, we need to check what happens to them, when they get unembedded now, and before they would not have been unembedded because it checked `LAYERTITLE=false`.

uprel: Don't know exactly about embedding nodes, that code was there before and was removing labels for single symbols. I just changed condition when this happens. Before (in 3.4) it happened if LAYERTITLE=FALSE, after your change I believe did not happen at all, so you would have same results just by removing complete `else... `sentence.
Tests are fine, so I really don't see any problems here, specially since we left default behaviour same. Change is only applied when using new value AUTO.

signedav: Okay, since my last implementation looks weird to me now, I did some more research.

I think, the only point where a nod is embedded is here:
https://github.com/qgis/QGIS/blob/master/src/core/qgsmaplayerlegend.cpp#L338-L339

This means that it embedds a child-node when there is exactly one of 'em, and when there is no label. This is the situation on single symbol or on category / rule with only one section without a label.

So the case of the unembedding never occurs before (after my implementation). So I'm good with your implementation, since it does not change the current behavior.

signedav: One last thing: When you remove the check for layername, with your `RULELABEL=AUTO` implementation, the label of the child is removed, no matter if `LAYERTITLE` is true or false or unset. When you want to have back the status of 3.4 with `RULELABEL=AUTO`, then it should still check it.

On 3.4: 
- `LAYERTITLE=FALSE` leads to display no Layertitle and no Rulelabel
- `LAYERTITLE=TRUE` or unset leads to display Layertitle and Rulelabel

After your implementation and `RULELABEL=AUTO`: 
- `LAYERTITLE=FALSE` leads to display no Layertitle and no Rulelabel
- `LAYERTITLE=TRUE` or unset leads to display Layertitle and **no** Rulelabel

Is this really what you want? If yes, then fair enough. If not, then leave the `!parameters.layerTitleAsBool() &&` there.

uprel: I did another test to try categorized legend with 1 label and AUTO parameter. It does not remove labels which is correct, so that embedding part is only for single symbols.

After some discussion in the beginning I removed LAYERTITLE check with RULELABEL=AUTO because I think it's more clear that way. So I'm correcting your part that after my implementation there is no dependency between RULELABEL and LAYERTITLE.

signedav: The part checking if LAYERTITLE is false, is not mine. It's the one that has been implemented in 3.4. as far as I know. I thought you want to have this behavior back with the AUTO, no?

uprel: I know this is older, bit I see no need for it now and I think is more clear logic

signedav: Don't you think it could be possible that an old client relies on that it's only removed when `LAYERTITLE` is false and needs this behavior? I know that I created this regression, but I think if we give em the possibility to have the old behavior back, why introduce another one that is neither like the new nor the old behavior?

But I'll approve the request still. Since it does not make any difference to the master behavior without setting `AUTO` parameter, it's okay.

signedav (approved):

src/server/services/wms/qgswmsgetlegendgraphics.cpp
uprel: Thanks, only change for old clients to get back used behaviour from 3.4 will be to add fixed AUTO value which is not a problem. This solution is better now, I will also add it into documentation once issue is opened there.

signedav: So the check for LAYERTITLE has not been in 3.4? Okay, then it's fine. I thought in 3.4 it did display the text of the singlesymbols when LAYERTITLE was true and it did not when LAYERTITLE was false.

uprel: No, sorry if I wasn't clear, you thought correct. There was a check for LAYERTITLE=false and only in that case labels were removed for single symbols. That's why with this change to get old behaviour client must apply AUTO value. But this is not a problem and like said logic to behave LAYERTITLE and RULELABEL independently is more clear. It does not have to be strict that only case to get RULELABEL=AUTO is when LAYERTITLE is FALSE.

uprel:
@signedav can you look at the backport I provided for 3.10. I had to do it manually, because bot did not work

PR #36162, opened 2020-05-04
Allow overriding the default symbol for a legend node

### LINKED ISSUE #14077: Enhancement: specify background for (transparent) items in composer legend
Original Redmine Issue: [4094](https://issues.qgis.org/issues/4094)

---
When, for example, using transparent polygons with an aerial photo underneath (I imagine this is a common thing to do), the composer legend entries can look a very different colour to the polygons in the map.  See the attached screenshot, particularly the red, purple and dark blue legend entries.

It would be good if there was a way to make the colours more similar.  I'm not sure _how_ to do this - I think something like taking an actual slice out of the map to use for the legend entry would be very complicated.  Maybe there could be an option to specify a colour or an image or something to put under the legend entries... or an option to disable transparency in the legend, or something along those lines.  Does anybody know of any other software that tries to deal with this issue?

---

- [legend_transparency.JPG](https://issues.qgis.org/attachments/download/3348/legend_transparency.JPG) (Alister Hood)

nyalldawson:
This allows users to (optionally!) customise the symbol appearance
for a legend node, e.g. to tweak the colors or symbol sizes to better
provide a "representative" patch symbol compared with how those
corresponding features actually appear on the map.

It's useful for exaggerating symbol widths, or for manually tweaking
the colors of semi-transparent symbols so that the colors represent
the actual appearance of the symbols when rendered on top of the map
content. Or to tweak the marker interval/offset in marker lines so that the
markers are nicely spaced in the legend patch.

Fixes #14077


nirvn (approved):
Love is love is legend.

m-kuhn:
Nice, I've missed that more than once!

PR #36179, opened 2020-05-05
[layouts] When dropping an SVG onto layouts, treat it as such

nirvn:
## Description

@nyalldawson , stumbled on a small issue with the picture drop handler whereas SVGs dropped are wrongly treated as raster. This PR addresses that.


nyalldawson (approved):

nyalldawson:
This is master only, right?

nirvn:
@nyalldawson , possibly :) I looked at the history, saw Nov. 28, 2019 as the date the feature was introduced and assumed it made it into 3.12.

PR #36188, opened 2020-05-05
Avoid long loading in attribute table on model change

### LINKED ISSUE #35927: Attribute table loads features really slow when changing from "Show visible.." to "Show all"
**Describe the bug**

Changing the attribute table from ``Show Features Visible On Map`` to ``Show All Features`` is taking forever (approximately 40 features per second) for any layers (tested with Postgres and Shapefiles). If I open the attribute table directly to either ``Show All Features`` or ``Show Features Visible On Map`` it works instant. Only when switching from Visible to All the bug occurs. 

**How to Reproduce**

1. Go to attribute table of a vector layer
2. Change to ``Show Features Visible On Map`` then to ``Show All Features``
3. Stuck for eternity 

**QGIS and OS versions**

Tested on 3.10.5 on Ubuntu Focal and Debian Bullseye but also with Master built from source on Debian Bullseye.

QGIS version | 3.13.0-Master | QGIS code revision | 45bead2bb9
-- | -- | -- | --
PostgreSQL Client Version | 12.2 (Debian 12.2-4) | SpatiaLite Version | 4.3.0a
Compiled against PROJ | 7.0.0 | Running against PROJ | Rel. 7.0.0, March 1st, 2020
OS Version | Debian GNU/Linux bullseye/sid
Active python plugins | pluginbuilder3; plugin_reloader; scriptrunner3; trackable_project_files; db_manager; processing

and

QGIS version | 3.10.5-A Coruña | QGIS code revision | 58395a2895
-- | -- | -- | --
PostgreSQL Client Version | 12.2 (Debian 12.2-4) | SpatiaLite Version | 4.3.0a
Compiled against PROJ | 7.0.0 | Running against PROJ | Rel. 7.0.0, March 1st, 2020
OS Version | Debian GNU/Linux bullseye/sid
Active python plugins | pluginbuilder3; qgepplugin; plugin_reloader; scriptrunner3; trackable_project_files; db_manager; processing

**Additional context**

signedav:
The filtered features in the attribute table have been reloaded when the model changed, to be able to display added features etc. in the filter mode as well. Same for only-show-visible-on-map.

Since the reload listened on the models signal `dataChanged` this leaded to very long reloads since on a complete model reload the signal was fired multiple times.

This PR fixes #35927 because the filter model is reloaded now on `featureAdded` etc. signals on the layer instead on the signal `dataChanged` of the main model (and on the `finished` signal of the model, what is called when the complete main model finished reloading).

Additionally there is a short timer to avoid long reloading times e.g. when a huge amount of features are added at the same time. 

And there has been some ugly connect/disconnect logic I improved (first commit) as well.

### DIFF
--- a/python/gui/auto_generated/attributetable/qgsattributetablefiltermodel.sip.in
+++ b/python/gui/auto_generated/attributetable/qgsattributetablefiltermodel.sip.in
@@ -94,6 +94,16 @@ Set the filter mode the filter will use.
 :param filterMode: Sets the current mode of the filter
 %End
 
+    void disconnectFilterModeConnections();
+%Docstring
+Disconnect the connections set for the current filterMode
+%End
+
+    void connectFilterModeConnections( FilterMode filterMode );
+%Docstring
+Disconnect the connections set for the new ``filterMode``
+%End
+
     FilterMode filterMode();
 %Docstring
 The current filterModel
--- a/python/gui/auto_generated/attributetable/qgsattributetablefiltermodel.sip.in
+++ b/python/gui/auto_generated/attributetable/qgsattributetablefiltermodel.sip.in
@@ -216,6 +216,16 @@ Emitted whenever the sort column is changed
 :param order: The sort order
 %End
 
+    void featuresFiltered();
+%Docstring
+Emitted when the filtering of the features has been done
+%End
+
+    void visibleReloaded();
+%Docstring
+Emitted when the the visible features on extend are reloaded (the list is created)
+%End
+
   protected:
 
     virtual bool filterAcceptsRow( int sourceRow, const QModelIndex &sourceParent ) const;
--- a/src/gui/attributetable/qgsattributetablefiltermodel.cpp
+++ b/src/gui/attributetable/qgsattributetablefiltermodel.cpp
@@ -318,25 +318,37 @@ void QgsAttributeTableFilterModel::setFilterMode( FilterMode filterMode )
 {
   if ( filterMode != mFilterMode )
   {
-    if ( filterMode == ShowVisible )
+    // cleanup existing connections
+    switch ( mFilterMode )
     {
-      connect( mCanvas, &QgsMapCanvas::extentsChanged, this, &QgsAttributeTableFilterModel::reloadVisible );
-      connect( mTableModel, &QgsAttributeTableModel::dataChanged, this, &QgsAttributeTableFilterModel::reloadVisible );
-      generateListOfVisibleFeatures();
-    }
-    else
-    {
-      disconnect( mCanvas, &QgsMapCanvas::extentsChanged, this, &QgsAttributeTableFilterModel::reloadVisible );
-      disconnect( mTableModel, &QgsAttributeTableModel::dataChanged, this, &QgsAttributeTableFilterModel::reloadVisible );
+      case ShowVisible:
+        disconnect( mCanvas, &QgsMapCanvas::extentsChanged, this, &QgsAttributeTableFilterModel::reloadVisible );
+        disconnect( mTableModel, &QgsAttributeTableModel::dataChanged, this, &QgsAttributeTableFilterModel::reloadVisible );
+        break;
+      case ShowAll:
+      case ShowEdited:
+      case ShowSelected:
+        break;
+      case ShowFilteredList:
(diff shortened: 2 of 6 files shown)


nyalldawson:
Any chance of a minimal changeset for the backports? As we all know attribute table is held together with chewing gum and it makes me nervous to see changes like this backported.

signedav:
Ok, I'll try to make a minimal bugfix changeset for the the backports separately.

signedav:
At @nyalldawson maybe you can have a look at this #36194

tudorbarascu:
@signedav Just tested this. When changing from all features to visible ones the count of filtered features remains 0 although they're plenty in the attribute table. Only by moving the map the signal is there and the filtered count is ok.
On my map, if I zoom in zoom out..several times with "visible features" activated, I get a stalled rendering.
I'll try to retest before your commits and see if that's also the case.

signedav:
I encountered issues before my commits that the title is not updated sometimes and decided to care about them later. But of course if these commits made it worse, I will care about them here. Tell me what you find out. I'll check it tomorrow.

signedav:
@tudorbarascu I've checked again and have seen that in your described case it's consequently not updating the title. I guess because I set the filterMode in `QgsAttributeTableFilterModel` before I reload the data to avoid that signals of the "old" filterMode are called on reload. So I connected the `filterChanged` to the signals of `QgsAttributeTableFilterModel`, and now the title should be updated properly.

signedav:
Hi @tudorbarascu Did you had the possibility to test it again? Is it now okay for you?

tudorbarascu:
Hi @signedav! Sorry for not being snappier. I rebuild against master. The title is now set correctly but the slowness reapearred:
- Opening an attribute table with show all features takes about 2 seconds for 25k features (lines) on my machine (PostGIS)
- Show features visible on map (around 450) is about 0.5 seconds
- Changing back to Show all features takes 28 seconds (and I have a very good setup)

Thank you for working on this.

signedav:
Thanks @tudorbarascu and sorry for the effort you had. Maybe the title update reintroduced the problem. I'll have a look again and trigger you when I have news.

tudorbarascu:
@signedav No effort.. glad to do it.. not always finding time to synchronise well though.

signedav:
Hi @tudorbarascu I have problem in reproducing the delay. On switching from "Show visible on map" to "Show all features" I have the expected 2-3 seconds (like we have it on QGIS < 3.10). I have a Geopackage with 20k features.

So I hope you can help me locating the remaining issue.

> Changing back to Show all features takes 28 seconds (and I have a very good setup) instead of about 2 seconds..

This happened *again* after I made the commit for the title update? Or was it the issue all the time on testing this PR and the behavior regarding the loading time has never been better?

tudorbarascu:
The situation was improved in the beginning but after the commit with the title update in was bad again. I'm testing with Postgres layers.

signedav:
Hi @tudorbarascu 
Happy that it actually seemed to work. But to be sure, I made some more tests with PG. And you where right, there has been a long reload sometimes. I found out that this has been effected by a call of the filtermodel `invalidateFilter()`, on the previous filterMode, before the layer is reloaded. I'm not sure, why this takes that long, but actually it's not needed to do it at all. So I splitted the methods and now it does it in the correct order, means on changing the mode it does:
- disconnect the connections of the current (old) filterMode
- reload the layer
- connect the connections of the new filterMode
- set the filterMode and invalidateFilter

So this should work now. I would be glad, if you could doublecheck. And thanks a lot.

signedav:
PS. Could you make a test on the LTR branch as well #36194?
There it only fixes the first reason of the long reload on the implementation effecting this #35927
The second issue with the `invalidateFilter()` never happened there, since I never touched the "not nice but still stable" behavior of setting the FilterMode after the reload of the layer.

tudorbarascu:
@signedav

> PS. Could you make a test on the LTR branch as well #36194?
I've started testing but I cannot find your pull request for the LTR branch ( I can only see for 3.12). Do you want me to cherry pick some commits and test for 3.10?

signedav:
> I've started testing but I cannot find your pull request for the LTR branch ( I can only see for 3.12). Do you want me to cherry pick some commits and test for 3.10?

I haven't one opened before yet, sorry. I thought about the 3.12 branch with automated back port. But I  I'll test it locally again for the 3.10 tomorrow morning, reopen the PR for 3.10 and trigger you again. Thanks.

3nids (approved):

PR #36189, opened 2020-05-05
[docker] Avoid assuming /dev/tty is usable

### LINKED ISSUE #36187: [docker] qgis_testrunner.sh assumes tty available
You cannot use qgis_testrunner.sh from a github-action, becuase github-action runs are not allocating ttys and thus the `| tee /dev/tty` code in qgis_testrunner.sh fails.

See https://github.com/linz/gazetteer/runs/645039116#step:3:1 for a debugging of github-action environment tty state, and see https://github.com/actions/runner/issues/241 for other people complaining about this on the github-action side.

In any case, using `| tee /dev/tty` in a script does sound fragile to me (at least the script should check if the terminal IS a tty?)

strk:
Closes #36187

### DIFF
--- a/.docker/qgis_resources/test_runner/qgis_testrunner.sh
+++ b/.docker/qgis_resources/test_runner/qgis_testrunner.sh
@@ -13,10 +13,19 @@ TEST_NAME=$1
 TEST_RUNNER_PATH=${TEST_RUNNER_PATH:-/usr/bin/qgis_testrunner.py}
 QGIS_BUILD_PATH=${QGIS_BUILD_PATH:-qgis}
 
-echo "Running test $1 ..."
-OUTPUT=$(QGIS_TEST_MODULE=${TEST_NAME} unbuffer ${QGIS_BUILD_PATH} --version-migration --nologo --code ${TEST_RUNNER_PATH} "$TEST_NAME"  2>/dev/null | tee /dev/tty)
+LOGFILE=/tmp/qgis_testrunner_$$
 
+echo "Running test $1 ..."
+QGIS_TEST_MODULE=${TEST_NAME} unbuffer ${QGIS_BUILD_PATH} \
+  --version-migration --nologo --code \
+  ${TEST_RUNNER_PATH} "$TEST_NAME"  \
+  2>/dev/null | \
+  tee ${LOGFILE}
+
+# NOTE: EXIT_CODE will always be 0 if "tee" works,
+#       we could `set -o pipefail` to change this
 EXIT_CODE="$?"
+OUTPUT=$(cat $LOGFILE) # quick hack to avoid changing too many lines
 if [ -z "$OUTPUT" ]; then
     echo "ERROR: no output from the test runner! (exit code: ${EXIT_CODE})"
     exit 1


strk:
I'm not sure why we want to hide stderr, btw...

3nids:
@elpaso could you have a look at this one?

elpaso:
Honestly I don't recall why the `tee` was added. But of one things I'm sure: the system was designed to run on Travis CI. (<- period)

So, whatever you do to enhance it just make sure it continues working on the environment it was designed for.

elpaso (commented):
I'm -1 to change something that has been working for years in the environment it was designed for, unless you are 100% sure that it won't break.

But if you are sure, no objections.

strk:
> So, whatever you do to enhance it just make sure it continues working on the environment it was designed for.

Is the script used by the Travis run triggered by this PR ? Or how else could we try it ?
All my change does is replace a `| tee` with a subsequent `echo`. It's a real simple change. Even improves the error handling, because `| tee` pipe would return exit code of `tee`, not of the command.

The only "downside" would be that output will come out all at once rather than while being produced, but on Travis this does not matter, right ?

elpaso:
> > So, whatever you do to enhance it just make sure it continues working on the environment it was designed for.
> 
> Is the script used by the Travis run triggered by this PR ? Or how else could we try it ?

It should be running on all PRs.  You can easily check for "docker_test_runners" in the travis logs.

> All my change does is replace a `| tee` with a subsequent `echo`. It's a real simple change. Even improves the error handling, because `| tee` pipe would return exit code of `tee`, not of the command.
> 
> The only "downside" would be that output will come out all at once rather than while being produced, but on Travis this does not matter, right ?

No idea, sorry. The only memory I have is that it was tricky and delicate to get it working (see unbuffer, tee etc.) but maybe you find a simpler way or the environment is now simpler.

Gustry:
> The only "downside" would be that output will come out all at once rather than while being produced, but on Travis this does not matter, right ?

From Travis documentation
https://docs.travis-ci.com/user/customizing-the-build/#build-timeouts

> When a job produces no log output for 10 minutes.

This was the reason. Tests on InaSAFE were taking more than 10 minutes at that time.
So we needed continuous logs from QGIS to keep travis happy.

strk:
Thanks @Gustry, then I'll change the code to use a temporary file instead.
See how you like 4427ab479d

strk:
Ok, CI is happy so I'm merging this

vot4anto:
Hi,
 I'm using the ltr image of qgis to do test with GitHub actions, but I have the problem with the tty and docker cause by the using of tee in qgis_testrunner.sh, but I have issue also with the latest release.
Following you can see the error that we have using ltr and latest image:

**ltr-3.10:**
docker exec qgis sh -c 'cd /tests_directory && qgis_testrunner.sh svir.test.integration.test_drive_oq_engine'
Running test svir.test.integration.test_drive_oq_engine ...
tee: /dev/tty: No such device or address
Finished running test svir.test.integration.test_drive_oq_engine (codes: IS_DEAD=1 IS_FAILED=0 IS_PASSED=1).
Error: Process completed with exit code 1.

**latest:**
docker exec qgis sh -c 'cd /tests_directory && qgis_testrunner.sh svir.test.integration.test_drive_oq_engine'
Running test svir.test.integration.test_drive_oq_engine ...
ERROR: no output from the test runner! (exit code: 0)
Error: Process completed with exit code 1.

Gustry:
> Is it possible to have the same scripts on ltr image?

I use the same script.
Do not use `-i` when you launch your container.

I don't have any issue for me.

vot4anto:
I don't use either -i and also -t option to launch container, but simple:
docker exec qgis sh -c "cd /tests_directory && qgis_testrunner.sh svir.test.integration.test_drive_oq_engine"
It works if I use qgis/qgis:final-3_10_8 but don't work if I use  qgis/qgis:ltr-3_10

PR #36190, opened 2020-05-05
allow editing of links in file widget

3nids:
In a form, links can now be edited (when widget is not readonly) in the file widget.

ezgif-6-c4737ab9cb1f

Sponsored by the QGIS Swiss user group

### DIFF
--- a/src/gui/qgsfilewidget.cpp
+++ b/src/gui/qgsfilewidget.cpp
@@ -51,6 +51,9 @@ QgsFileWidget::QgsFileWidget( QWidget *parent )
   mLinkLabel->setSizePolicy( QSizePolicy::Expanding, QSizePolicy::Preferred );
   mLinkLabel->setTextFormat( Qt::RichText );
   mLinkLabel->hide(); // do not show by default
+  mLinkEditButton = new QToolButton( this );
+  mLinkEditButton->setIcon( QgsApplication::getThemeIcon( QStringLiteral( "/mActionToggleEditing.svg" ) ) );
+  connect( mLinkEditButton, &QToolButton::clicked, this, &QgsFileWidget::editLink );
 
   // otherwise, use the traditional QLineEdit subclass
   mLineEdit = new QgsFileDropEdit( this );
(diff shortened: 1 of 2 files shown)


src/gui/qgsfilewidget.cpp
```diff
@@ -51,6 +51,9 @@ QgsFileWidget::QgsFileWidget( QWidget *parent )
   mLinkLabel->setSizePolicy( QSizePolicy::Expanding, QSizePolicy::Preferred );
   mLinkLabel->setTextFormat( Qt::RichText );
   mLinkLabel->hide(); // do not show by default
+  mLinkEditButton = new QToolButton( this );
```
nyalldawson, line 54: To confirm -- this is only shown when links are shown, right? not in the default line edit mode?

3nids: only shown with the label (links shown) and when not in read only.

nyalldawson: thanks!

Saijin-Naib:
This will make life way easier for me.

src/gui/qgsfilewidget.cpp
```diff
@@ -220,6 +228,43 @@ QgsFilterLineEdit *QgsFileWidget::lineEdit()
   return mLineEdit;
 }
 
+void QgsFileWidget::updateLayout()
+{
+  mLayout->removeWidget( mLineEdit );
```
nyalldawson, line 233: I'm not 100% sure (and the documentation is unclear), but I *think* this results in a leak of the QLayoutItem associated with the widget.

nyalldawson:
As noted above, I think there **may** be a leak here. If not, LGTM.

src/gui/qgsfilewidget.cpp
3nids: I think this is OK: 

> Note: The ownership of widget remains the same as when it was added.

https://doc.qt.io/qt-5/qlayout.html#removeWidget

PR #36202, opened 2020-05-05
allow creating geometryless DBF tables from the New Shapefile dialog (fix #15654)

### LINKED ISSUE #15654: Add ability to create a new dbf table
Original Redmine Issue: [6400](https://issues.qgis.org/issues/6400)

Redmine category:vectors

---
It would be good to have a menu entry "Layer>New>New DBF Table"
I think I remember seeing people ask for this feature on the list in the past...

alexbruy:
## Description
Allow creating geometryless (plain) DBF files from the New Shapefile dialog.
geometryless
Creating DBF tables can be useful in some cases and need to use 3rd party tools for this instead of doing it from QGIS is not convinient. Also this makes New Shapefile dialog more cosistent with the New Geopackage dialog which already allows creating geometryless tables.

Fixes #15654.

### DIFF
--- a/src/gui/qgsnewvectorlayerdialog.cpp
+++ b/src/gui/qgsnewvectorlayerdialog.cpp
@@ -56,7 +56,7 @@ QgsNewVectorLayerDialog::QgsNewVectorLayerDialog( QWidget *parent, Qt::WindowFla
   mWidth->setValidator( new QIntValidator( 1, 255, this ) );
   mPrecision->setValidator( new QIntValidator( 0, 15, this ) );
 
-  mGeometryTypeBox->addItem( QgsApplication::getThemeIcon( QStringLiteral( "/mIconTableLayer.svg" ) ), tr( "No geometry" ), QgsWkbTypes::NoGeometry );
+  mGeometryTypeBox->addItem( QgsApplication::getThemeIcon( QStringLiteral( "/mIconTableLayer.svg" ) ), tr( "No Geometry" ), QgsWkbTypes::NoGeometry );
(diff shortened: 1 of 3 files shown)


src/gui/qgsnewvectorlayerdialog.cpp
```diff
@@ -56,6 +56,7 @@ QgsNewVectorLayerDialog::QgsNewVectorLayerDialog( QWidget *parent, Qt::WindowFla
   mWidth->setValidator( new QIntValidator( 1, 255, this ) );
   mPrecision->setValidator( new QIntValidator( 0, 15, this ) );
 
+  mGeometryTypeBox->addItem( QgsApplication::getThemeIcon( QStringLiteral( "/mIconTableLayer.svg" ) ), tr( "No geometry" ), QgsWkbTypes::NoGeometry );
```
nyalldawson, line 59:
```suggestion
  mGeometryTypeBox->addItem( QgsApplication::getThemeIcon( QStringLiteral( "/mIconTableLayer.svg" ) ), tr( "No Geometry" ), QgsWkbTypes::NoGeometry );
```

nyalldawson (approved):

src/gui/qgsnewvectorlayerdialog.cpp
alexbruy: Thanks, fixed. Also fixed in the New Geopackage dialog.

nyalldawson: Thanks!

Issue #30023, opened 2019-05-30
Zooming with mouse wheel doesn't center on cursor when scale lock is on.


arawakarawak:
**Describe the bug**

Zooming in within the map window with mouse wheel/roller centers zoom on the mouse cursor position (and that is correct behavior). However when the scale lock is engaged, the same operation is performed relative to center of the window. 
The difference in behavior is unjustified and confusing.

**How to Reproduce**
1. Open a project
2. Try to zoom in the map window with mouse roller: the view zooms on current position of your mouse cursor
3. Engage scale lock (with the padlock icon on bottom bar) 
4. Try to zoom in: the view zooms relative to the center of the map window.

**Desktop environment:**
 - OS and version: Win 7 Home Premium 64 bit SP1

**QGIS version**
  -  3.6.3-Noosa

PR #36218, opened 2020-05-05
Re-center the map when zooming with active scale lock

jdugge:
## Description

When zooming with the mouse wheel or the Zoom map tool while the scale is locked the zoom is not relative to the mouse / zoom box position, but relative to the center of the screen (see #30023).

This PR fixes this behavior by adding a new optional `center` argument to the 
`QgsMapSettings::setMagnificationFactor` and `QgsMapCanvas::setMagnificationFactor` functions, which is then used to scale the extent rectangle.

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -3683,7 +3683,7 @@ void QgisApp::createStatusBar()
   mMagnifierWidget->setObjectName( QStringLiteral( "mMagnifierWidget" ) );
   mMagnifierWidget->setFont( statusBarFont );
   connect( mMapCanvas, &QgsMapCanvas::magnificationChanged, mMagnifierWidget, &QgsStatusBarMagnifierWidget::updateMagnification );
-  connect( mMagnifierWidget, &QgsStatusBarMagnifierWidget::magnificationChanged, mMapCanvas, &QgsMapCanvas::setMagnificationFactor );
+  connect( mMagnifierWidget, &QgsStatusBarMagnifierWidget::magnificationChanged, mMapCanvas, [ = ]( double factor ) { mMapCanvas->setMagnificationFactor( factor ); } );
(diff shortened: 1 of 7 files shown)


src/app/qgisapp.cpp
```diff
@@ -3683,7 +3683,7 @@ void QgisApp::createStatusBar()
   mMagnifierWidget->setObjectName( QStringLiteral( "mMagnifierWidget" ) );
   mMagnifierWidget->setFont( statusBarFont );
   connect( mMapCanvas, &QgsMapCanvas::magnificationChanged, mMagnifierWidget, &QgsStatusBarMagnifierWidget::updateMagnification );
-  connect( mMagnifierWidget, &QgsStatusBarMagnifierWidget::magnificationChanged, mMapCanvas, &QgsMapCanvas::setMagnificationFactor );
+  connect( mMagnifierWidget, &QgsStatusBarMagnifierWidget::magnificationChanged, mMapCanvas, [ = ]( double factor ) { mMapCanvas->setMagnificationFactor( factor ); } );
```
3nids, line 3686: why is this change required since there is a default value in the slot?

jdugge: The Qt5 `connect` syntax [does not support default arguments in slots](https://wiki.qt.io/New_Signal_Slot_Syntax#Default_arguments_in_slot), unfortunately. I used a lambda function to work around this, which seems to be the standard way to handle this ([example 1](https://artandlogic.com/2013/09/qt-5-and-c11-lambdas-are-your-friend/), [example 2](https://stackoverflow.com/a/40598273/1456857)).

nyalldawson:
@3nids what's your thoughts? Can we merge?

jdugge:
@3nids: Is there anything I should change so this PR can be merged?

nyalldawson (approved):

PR #36222, opened 2020-05-06
Allow customisation of division and subdivision symbols as distinct from scalebar tick horizontal symbol

nyalldawson:
Allows for styling division and subdivision symbols in a different
way compared with the horizontal line symbol in a tick scalebar

Sponsored by SLYR

Builds off the hard work done by @agiudiceandrea!


src/core/scalebar/qgsscalebarsettings.h
```diff
@@ -386,7 +394,7 @@ class CORE_EXPORT QgsScaleBarSettings
 
     /**
      * Sets the pen used for drawing outlines in the scalebar.
-     * \see pen()
+     * \see pen()  mSubdivisionsHeightSpinBox->setMaximum( d );
```
nirvn, line 397:
```suggestion
     * \see pen()
```

src/ui/layout/qgslayoutscalebarwidgetbase.ui
```diff
@@ -557,10 +600,15 @@
...
+  <customwidget>
+   <class>QgsSpinBox</class>
+   <extends>QSpinBox</extends>
+   <header>qgsspinbox.h</header>
+  </customwidget>
   <customwidget>
    <class>QgsCollapsibleGroupBoxBasic</class>
    <extends>QGroupBox</extends>
-   <header>qgscollapsiblegroupbox.h</header>
+   <header location="global">qgscollapsiblegroupbox.h</header>
```
nirvn, line 611: location="global" -> that's a QtCreator change?

nyalldawson: Yep -- seems to randomly turn this on or off based on ... ?

agiudiceandrea:
This is a big enhancement!

Just a note about removing the subdivisions tick height restriction: to properly handle the subdivisions tick height > segments tick height case, I think some changing should be made in how the BoxSize and the position of the horizontal bar are calculated, and how the polyline of the segments ticks is generated.
Otherwise...
image

nyalldawson:
@agiudiceandrea 

I wondered the same thing, but when testing found that you can handle this by adjusting the label distance manually, and that it seemed quite natural to do... I'd suggest we merge and to give it a test in practice and then determine if we want to automatically handle this situation in the scalebar layout code or not...

PR #36231, opened 2020-05-06
Update vertex tool to base its snapping parameters on global ones 

### LINKED ISSUE #36229: Vertex tool override global snapping parameters
The vertex tool seems to create its own local snapping parameters, overriding the global one. It might not have been an issue until now but the feature to limit snapping to some scale range is bypassed by this behavior. This can be an issue when dealing with huge volume of data.

See https://github.com/qgis/QGIS/blob/master/src/app/vertextool/qgsvertextool.cpp#L742

To reproduce : 

* Set snapping scale limit to global and specify a min and max scale (ex : [5000, 500]).
* Edit a vector layer and use the vertex tool
* Snapping is enabled at any scale

obrix:
Update vertex tool to not override some of the global snapping parameters (especially snapping enabled on a scale range).

Keep some of the specific behavior of enabling snapping in priority on an active layer (ie creation of a specific layer settings, now based an existing one if present).

Should fix #36229

### DIFF
--- a/src/app/vertextool/qgsvertextool.cpp
+++ b/src/app/vertextool/qgsvertextool.cpp
@@ -780,6 +780,9 @@ QgsPointLocator::Match QgsVertexTool::snapToEditableLayer( QgsMapMouseEvent *e )
         {
           layerSettings = existingSettings.value();
           layerSettings.setEnabled( vlayer == currentVlayer );
+          layerSettings.setTolerance( tol );
+          layerSettings.setTypeFlag( static_cast<QgsSnappingConfig::SnappingTypeFlag>( QgsSnappingConfig::VertexFlag | QgsSnappingConfig::SegmentFlag ) );
+          layerSettings.setUnits( QgsTolerance::ProjectUnits );
         }
         else
         {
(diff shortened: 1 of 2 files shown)


src/app/vertextool/qgsvertextool.cpp
```diff
@@ -768,8 +774,23 @@ QgsPointLocator::Match QgsVertexTool::snapToEditableLayer( QgsMapMouseEvent *e )
         if ( !vlayer )
           continue;
 
-        config.setIndividualLayerSettings( vlayer, QgsSnappingConfig::IndividualLayerSettings(
-                                             vlayer == currentVlayer, static_cast<QgsSnappingConfig::SnappingTypeFlag>( QgsSnappingConfig::VertexFlag | QgsSnappingConfig::SegmentFlag ), tol, QgsTolerance::ProjectUnits, 0.0, 0.0 ) );
+        QgsSnappingConfig::IndividualLayerSettings layerSettings;
+        SettingsHashMap::const_iterator existingSettings = oldLayerSettings.find( vlayer );
+        if ( existingSettings != oldLayerSettings.constEnd() )
```
troopa81, line 779: Wouldn't be simpler to initialize config with oldconfig and just changing the IndividualLayerSettings enabled boolean according to whether its the currentLayer or not (or editable layer for the following code) ?

troopa81 (changes requested):

src/app/vertextool/qgsvertextool.cpp
obrix: Yes probably if we are sure there is always an existing IndividualLayerSetting for each layer which I am not. Also tolerance, typeflag and units are overridden here for each setting.

troopa81: You're right! I still wondering if it would not be better to initialize `config` with `oldConfig` one. If someone adds a new SnapingConfig parameter, it would not be takein into count here.

But maybe it's better this way

obrix: Yes I think your are right for the initial config init with old config. I will do that.

src/core/qgssnappingutils.cpp
```diff
@@ -94,7 +94,7 @@ bool QgsSnappingUtils::isIndexPrepared( QgsPointLocator *loc, const QgsRectangle
   if ( mStrategy == IndexAlwaysFull && loc->hasIndex() )
     return true;
 
-  if ( mStrategy == IndexExtent && loc->hasIndex() && loc->extent()->intersects( areaOfInterest ) )
+  if ( mStrategy == IndexExtent && loc->hasIndex() && ( !loc->extent() || loc->extent()->intersects( areaOfInterest ) ) )
```
troopa81, line 97: It shouldn't be possible to get in a situation where the strategy is IndexExtent and the locator has no extent defined, isn't it?

obrix: Well I had to add this check to avoid snapping to crash, so yes It seems to be possible with the state of the code. I did not really understood why though. Might be worth it to spend further time to investigate.

troopa81: As we discuss, it could happen when some QgsPointLocator methods (nearestvertex for instance) are called first outside of QgsSnappingUtils (in QgsVertexTool basically). 

Then, when _prepareIndex_ is called and so _isIndexPrepared_, extent is null and strategy is IndexExtent. So, your modification make sense.

troopa81 (approved):

PR #36234, opened 2020-05-06
[processing] rename Graphics group to Plots

### LINKED ISSUE #33310: Change the name of the "Graphics" QGIS Processing group to "Plots"
**Feature description.**

"Graphics" is a strange group name for the (QGIS) Processing algorithms
that are placed in this group.
All of the algorithms in the group do statistical plots, so I suggest that we
name the group "Plots" or "Statistical plots" instead.
Any other suggestions?
If we agree on this, I can try to create a PR for it.

The following code will have to be changed in the affected algorithms:
```
def group(self):
    return self.tr('Graphics')
```
Change to:
```
def group(self):
    return self.tr('Plots')
```
For consistency, we should perhaps also change:
```
def groupId(self):
    return 'graphics'
```
to:
```
def groupId(self):
    return 'plots'
```

Affected algorithm code:

./python/plugins/processing/algs/qgis/BarPlot.py
./python/plugins/processing/algs/qgis/VectorLayerHistogram.py
./python/plugins/processing/algs/qgis/BoxPlot.py
./python/plugins/processing/algs/qgis/RasterLayerHistogram.py
./python/plugins/processing/algs/qgis/PolarPlot.py
./python/plugins/processing/algs/qgis/VectorLayerScatterplot.py
./python/plugins/processing/algs/qgis/MeanAndStdDevPlot.py
./python/plugins/processing/algs/qgis/VectorLayerScatterplot3D.py

Are there other places in the code that we need to make changes?
I could not find any with
> grep "'Graphics'" `find . -print` 2>/dev/null

I can fix the documentation (noticed this while documenting...).

**Additional context**

alexbruy:
## Description
Name "Graphics" does not reflect purpose of the algorithms (which produce statistical plots) inside this group. "Plots" is much better name.

Fixes #33310.

### DIFF
--- a/python/plugins/processing/algs/qgis/BarPlot.py
+++ b/python/plugins/processing/algs/qgis/BarPlot.py
@@ -43,10 +43,10 @@ class BarPlot(QgisAlgorithm):
     VALUE_FIELD = 'VALUE_FIELD'
 
     def group(self):
-        return self.tr('Graphics')
+        return self.tr('Plots}')
 
     def groupId(self):
-        return 'graphics'
+        return 'plots'
 
     def __init__(self):
         super().__init__()
--- a/python/plugins/processing/algs/qgis/BoxPlot.py
+++ b/python/plugins/processing/algs/qgis/BoxPlot.py
@@ -44,10 +44,10 @@ class BoxPlot(QgisAlgorithm):
     MSD = 'MSD'
 
     def group(self):
-        return self.tr('Graphics')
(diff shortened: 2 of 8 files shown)


m-kuhn:
@alexbruy you have some really good pull requests inbound recently, I love all this cleanup work a lot !

m-kuhn (approved):

alexbruy:
@m-kuhn thanks!

PR #36235, opened 2020-05-06
Expressions to convert to and from base64

JanCaha:
Expressions `to_base64` and `from_base64` including tests and docs.

### DIFF
--- a/resources/function_help/json/from_base64
+++ b/resources/function_help/json/from_base64
@@ -3,7 +3,6 @@
   "name": "from_base64",
   "type": "function",
   "description": "Decodes a provided string in Base64 into readable string.",
-  "variableLenArguments": false,
   "arguments": [
     {"arg":"string", "description": "the string to decode"}],
     "examples": [ { "expression":"from_base64('UUdJUw==')", "returns":"'QGIS'"}
--- a/resources/function_help/json/from_base64
+++ b/resources/function_help/json/from_base64
@@ -1,0 +1,11 @@
+
+{
+  "name": "from_base64",
+  "type": "function",
+  "description": "Decodes a provided string in Base64 into readable string.",
+  "variableLenArguments": false,
+  "arguments": [
+    {"arg":"string", "description": "the string to decode"}],
+    "examples": [ { "expression":"from_base64('UUdJUw==')", "returns":"'QGIS'"}
+    ]
+}
--- a/resources/function_help/json/to_base64
+++ b/resources/function_help/json/to_base64
@@ -3,7 +3,6 @@
   "name": "to_base64",
   "type": "function",
   "description": "Encodes a provided string using Base64.",
-  "variableLenArguments": false,
   "arguments": [
     {"arg":"string", "description": "the string to encode"}],
     "examples": [ { "expression":"to_base64('QGIS')", "returns":"'UUdJUw=='"}
--- a/resources/function_help/json/to_base64
+++ b/resources/function_help/json/to_base64
@@ -1,0 +1,11 @@
+
+{
+  "name": "to_base64",
+  "type": "function",
+  "description": "Encodes a provided string using Base64.",
+  "variableLenArguments": false,
+  "arguments": [
+    {"arg":"string", "description": "the string to encode"}],
+    "examples": [ { "expression":"to_base64('QGIS')", "returns":"'UUdJUw=='"}
+    ]
+}
--- a/src/core/expression/qgsexpressionfunction.cpp
+++ b/src/core/expression/qgsexpressionfunction.cpp
@@ -5545,6 +5545,25 @@ static QVariant fcnHashSha256( const QVariantList &values, const QgsExpressionCo
   return fcnHash( QgsExpressionUtils::getStringValue( values.at( 0 ), parent ), QCryptographicHash::Sha256 );
 }
 
+static QVariant fcnToBase64( const QVariantList &values, const QgsExpressionContext *, QgsExpression *parent, const QgsExpressionNodeFunction * )
+{
+  QString value = QgsExpressionUtils::getStringValue( values.at( 0 ), parent );
+
+  const QByteArray input = value.toLocal8Bit();
+
+  return QVariant( QString( input.toBase64() ) );
+}
+
+static QVariant fcnFromBase64( const QVariantList &values, const QgsExpressionContext *, QgsExpression *parent, const QgsExpressionNodeFunction * )
+{
+  QString value = QgsExpressionUtils::getStringValue( values.at( 0 ), parent );
+
+  const QByteArray base64 = value.toLocal8Bit();
+  const QByteArray decoded = QByteArray::fromBase64( base64 );
+
+  return QVariant( QString( decoded ) );
+}
+
 const QList<QgsExpressionFunction *> &QgsExpression::Functions()
 {
   // The construction of the list isn't thread-safe, and without the mutex,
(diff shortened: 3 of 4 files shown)


tests/src/core/testqgsexpression.cpp
```diff
@@ -1598,6 +1598,10 @@ class TestQgsExpression: public QObject
       QTest::newRow( "hash('QGIS', 'keccak_384')" ) << QStringLiteral( "hash('QGIS', 'keccak_384')" ) << false << QVariant( "c57a3aed9d856fa04e5eeee9b62b6e027cca81ba574116d3cc1f0d48a1ef9e5886ff463ea8d0fac772ee473bf92f810d" );
       QTest::newRow( "hash('QGIS', 'keccak_512')" ) << QStringLiteral( "hash('QGIS', 'keccak_512')" ) << false << QVariant( "6f0f751776b505e317de222508fa5d3ed7099d8f07c74fed54ccee6e7cdc6b89b4a085e309f2ee5210c942bbeb142bdfe48f84f912e0f3f41bdbf47110c2d344" );
 #endif
+      QTest::newRow( "to_base64 NULL" ) << QStringLiteral( "to_base64(NULL)" ) << false << QVariant();
+      QTest::newRow( "to_base64" ) << QStringLiteral( "to_base64('QGIS')" ) << false << QVariant( "UUdJUw==" );
+      QTest::newRow( "from_base64 NULL" ) << QStringLiteral( "from_base64(NULL)" ) << false << QVariant();
+      QTest::newRow( "from_base64" ) << QStringLiteral( "from_base64('UUdJUw==')" ) << false << QVariant( "QGIS" );
```
m-kuhn, line 1604: Can you add some tests that work on a QByteArray?

src/core/expression/qgsexpressionfunction.cpp
```diff
@@ -5545,6 +5545,25 @@ static QVariant fcnHashSha256( const QVariantList &values, const QgsExpressionCo
   return fcnHash( QgsExpressionUtils::getStringValue( values.at( 0 ), parent ), QCryptographicHash::Sha256 );
 }
 
+static QVariant fcnToBase64( const QVariantList &values, const QgsExpressionContext *, QgsExpression *parent, const QgsExpressionNodeFunction * )
+{
+  QString value = QgsExpressionUtils::getStringValue( values.at( 0 ), parent );
+
+  const QByteArray input = value.toLocal8Bit();
```
m-kuhn, line 5552: I think this should never go through a `QString` or a `toLocal8Bit` and be passed as raw bytes to tobase64.

src/core/expression/qgsexpressionfunction.cpp
```diff
@@ -5545,6 +5545,25 @@ static QVariant fcnHashSha256( const QVariantList &values, const QgsExpressionCo
...
+}
+
+static QVariant fcnFromBase64( const QVariantList &values, const QgsExpressionContext *, QgsExpression *parent, const QgsExpressionNodeFunction * )
+{
+  QString value = QgsExpressionUtils::getStringValue( values.at( 0 ), parent );
+
+  const QByteArray base64 = value.toLocal8Bit();
+  const QByteArray decoded = QByteArray::fromBase64( base64 );
+
+  return QVariant( QString( decoded ) );
```
m-kuhn, line 5564: I don't think this should be forced to a QString here and preserved as raw bytes.

m-kuhn:
Very nice functionality.
I wonder what your use case is though as you provide conversion between "strings" and "base64", whereas I would expect the conversion to be between "bytes" and "base64" to unleash the full power.

JanCaha:
Hm, I see your point. Actually, this was suggested by @nyalldawson and if I understand him correctly, then this implementation follows his suggestion. So he is more qualified to provide the use case.

Maybe the `from_base64` function could take both string and QByteArray as an input? It is possible to have the encoded data stored in both these types.

src/core/expression/qgsexpressionfunction.cpp
nyalldawson: Right - this should be:

    const QByteArray input = values.at( 0 ).toByteArray();
    return QVariant( QString( input.toBase64() ) );

src/core/expression/qgsexpressionfunction.cpp
nyalldawson: And this one should be:

     const QString value = QgsExpressionUtils::getStringValue( values.at( 0 ), parent );
     const QByteArray base64 = value.toLocal8Bit();
     const QByteArray decoded = QByteArray::fromBase64( base64 );
     return QVariant( decoded );

nyalldawson:
> Actually, this was suggested by @nyalldawson and if I understand him correctly, then this implementation follows his suggestion.

I think my email was just badly worded -- but the implementation here is flipped from what's needed.

JanCaha:
I fixed the implementation using @nyalldawson suggestions.

Had to add a new case for `run_evaluation_test` function in `testqgsexpression.cpp` as it did not supported `QByteArray`. Hopefully, it will not break anything.

resources/function_help/json/from_base64
```diff
@@ -0,0 +1,10 @@
+
+{
+  "name": "from_base64",
+  "type": "function",
+  "description": "Decodes a provided string in Base64 into readable string.",
```
nyalldawson, line 5:
```suggestion
  "description": "Decodes a provided string in Base64 into a binary value.",
```

resources/function_help/json/to_base64
```diff
@@ -0,0 +1,10 @@
+
+{
+  "name": "to_base64",
+  "type": "function",
+  "description": "Encodes a provided string using Base64.",
```
nyalldawson, line 5:
```suggestion
  "description": "Encodes a provided binary value into a string, using Base64 encoding.",
```

resources/function_help/json/to_base64
```diff
@@ -0,0 +1,10 @@
...
+  "arguments": [
+    {"arg":"string", "description": "the string to encode"}],
```
nyalldawson, line 7:
```suggestion
    {"arg":"string", "description": "the binary value to encode"}],
```

src/core/expression/qgsexpressionfunction.cpp
```diff
@@ -5851,6 +5865,12 @@ const QList<QgsExpressionFunction *> &QgsExpression::Functions()
         << new QgsStaticExpressionFunction( QStringLiteral( "sha256" ), QgsExpressionFunction::ParameterList() << QgsExpressionFunction::Parameter( QStringLiteral( "string" ) ),
                                             fcnHashSha256, QStringLiteral( "Conversions" ) )
 
+        //base64
+        << new QgsStaticExpressionFunction( QStringLiteral( "to_base64" ), QgsExpressionFunction::ParameterList() << QgsExpressionFunction::Parameter( QStringLiteral( "string" ) ),
```
nyalldawson, line 5869:
```suggestion
        << new QgsStaticExpressionFunction( QStringLiteral( "to_base64" ), QgsExpressionFunction::ParameterList() << QgsExpressionFunction::Parameter( QStringLiteral( "value" ) ),
```
(or "binary"?)

nyalldawson (commented):
Nearly there! Just some documentation improvements to go

nyalldawson:
Marking as freeze exempt given the prior approval and only doc wording left to address

nyalldawson:
Superseded by https://github.com/qgis/QGIS/pull/36563

PR #36236, opened 2020-05-06
[layout] allow sorting attribute table by field not listed in the table

### LINKED ISSUE #25671: new composer attribute table sort abilty
Author Name: **Kepes Ádám** (Kepes Ádám)
Original Redmine Issue: [17775](https://issues.qgis.org/issues/17775)

---
The next feature would be useful:

Qgis attribute tables in composer can only be sorted by visible fields. But there are cases when i don't want the field wich the sorting based on to be visible.

Qgis neither lets write random expressions at the sort table window, only written, already visible field expressions can be used for sorting table rows.

3nids:
instead of using the same data model for the displayed and the sorting columns, two data models are now used.
They use the same API / base class as they are very similar.
QgsLayoutTableColumn is no longer a QObject.

this fixes #25671

a new feature with less lines in the code base 💪

### DIFF
--- a/src/core/layout/qgscompositionconverter.cpp
+++ b/src/core/layout/qgscompositionconverter.cpp
@@ -1469,7 +1469,9 @@ bool QgsCompositionConverter::readTableXml( QgsLayoutItemAttributeTable *layoutI
 
       // sorting columns are now (QGIS 3.14+) handled in a dedicated list
       Q_NOWARN_DEPRECATED_PUSH
-      std::copy_if( layoutItem->mColumns.begin(), layoutItem->mColumns.end(), std::back_inserter( layoutItem->mSortColumns ), []( QgsLayoutTableColumn * col ) {return col->sortByRank() > 0;} );
+      for ( QgsLayoutTableColumn *col : qgis::as_const( layoutItem->mColumns ) )
+        if ( col->sortByRank() > 0 )
+          layoutItem->mSortColumns.append( col->clone() );
       std::sort( layoutItem->mSortColumns.begin(), layoutItem->mSortColumns.end(), []( QgsLayoutTableColumn * a, QgsLayoutTableColumn * b ) {return a->sortByRank() < b->sortByRank();} );
       Q_NOWARN_DEPRECATED_POP
     }
--- a/src/core/layout/qgscompositionconverter.cpp
+++ b/src/core/layout/qgscompositionconverter.cpp
@@ -1466,10 +1466,9 @@ bool QgsCompositionConverter::readTableXml( QgsLayoutItemAttributeTable *layoutI
       layoutItem->mColumns.append( column );
 
       // sorting columns are now (QGIS 3.14+) handled in a dedicated list
+      // copy the display columns if sortByRank > 0 and then, sort them by rank
       Q_NOWARN_DEPRECATED_PUSH
-      for ( const QgsLayoutTableColumn &col : qgis::as_const( layoutItem->mColumns ) )
-        if ( col.sortByRank() > 0 )
-          layoutItem->mSortColumns.append( col );
+      std::copy_if( layoutItem->mColumns.begin(), layoutItem->mColumns.end(), std::back_inserter( layoutItem->mSortColumns ), []( const QgsLayoutTableColumn & col ) {return col.sortByRank() > 0;} );
       std::sort( layoutItem->mSortColumns.begin(), layoutItem->mSortColumns.end(), []( const QgsLayoutTableColumn & a, const QgsLayoutTableColumn & b ) {return a.sortByRank() < b.sortByRank();} );
       Q_NOWARN_DEPRECATED_POP
     }
--- a/src/core/layout/qgscompositionconverter.cpp
+++ b/src/core/layout/qgscompositionconverter.cpp
@@ -1428,7 +1428,11 @@ bool QgsCompositionConverter::readTableXml( QgsLayoutItemAttributeTable *layoutI
   layoutItem->setWrapBehavior( static_cast<QgsLayoutTable::WrapBehavior>( itemElem.attribute( QStringLiteral( "wrapBehavior" ), QStringLiteral( "0" ) ).toInt() ) );
 
   //restore column specifications
+  qDeleteAll( layoutItem->mColumns );
   layoutItem->mColumns.clear();
+  qDeleteAll( layoutItem->mSortColumns );
+  layoutItem->mSortColumns.clear();
+
   QDomNodeList columnsList = itemElem.elementsByTagName( QStringLiteral( "displayColumns" ) );
   if ( !columnsList.isEmpty() )
   {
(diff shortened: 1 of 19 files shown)


src/core/layout/qgscompositionconverter.cpp
```diff
@@ -1458,10 +1460,18 @@ bool QgsCompositionConverter::readTableXml( QgsLayoutItemAttributeTable *layoutI
...
         {
-          column->mBackgroundColor = QColor( bgRed, bgGreen, bgBlue, bgAlpha );
+          column.mBackgroundColor = QColor( bgRed, bgGreen, bgBlue, bgAlpha );
         }
       }
       layoutItem->mColumns.append( column );
+
+      // sorting columns are now (QGIS 3.14+) handled in a dedicated list
+      Q_NOWARN_DEPRECATED_PUSH
+      for ( const QgsLayoutTableColumn &col : qgis::as_const( layoutItem->mColumns ) )
```
nyalldawson, line 1470: this would be easier to read/safer if you enclosed the loop contents in { / }

src/core/layout/qgslayoutitemattributetable.cpp
```diff
@@ -368,16 +367,16 @@ void QgsLayoutItemAttributeTable::restoreFieldAliasMap( const QMap<int, QString>
     return;
   }
 
-  for ( QgsLayoutTableColumn *column : qgis::as_const( mColumns ) )
+  for ( QgsLayoutTableColumn &column : mColumns )
```
nyalldawson, line 370: My understanding is that you're NOT allowed to modify a qt container inside a range based loop. Can you confirm that this is safe?

src/core/layout/qgslayoutitemattributetable.cpp
```diff
@@ -480,10 +479,9 @@ bool QgsLayoutItemAttributeTable::getTableContents( QgsLayoutTableContents &cont
     req.setFilterFid( atlasFeature.id() );
   }
 
-  QVector< QPair<int, bool> > sortColumns = sortAttributes();
-  for ( int i = sortColumns.size() - 1; i >= 0; --i )
+  for ( const QgsLayoutTableColumn &column : qgis::as_const( mSortColumns ) )
```
nyalldawson, line 482: I think there's an api break here -- if someone is using the old api to set a list of columns with their sort order contained, then that won't be respected here.

src/core/layout/qgslayouttable.cpp
```diff
@@ -758,12 +792,19 @@ void QgsLayoutTable::setWrapBehavior( QgsLayoutTable::WrapBehavior behavior )
...
-  qDeleteAll( mColumns );
   mColumns.clear();
 
   mColumns.append( columns );
 }
 
+void QgsLayoutTable::setSortColumns( const QgsLayoutTableSortColumns &sortColumns )
+{
+  //remove existing columns
+  mSortColumns.clear();
```
nyalldawson, line 803: mSortColumns = sortColumns would allow implicit sharing to optimise this

src/core/layout/qgslayouttable.h
```diff
@@ -54,7 +54,15 @@ typedef QVector< QVector< QVariant > > QgsLayoutTableContents;
...
 */
-typedef QVector< QgsLayoutTableColumn * > QgsLayoutTableColumns;
+typedef QVector<QgsLayoutTableColumn> QgsLayoutTableColumns;
+
+/**
+ * \ingroup core
+ * List of column definitions for sorting a QgsLayoutTable
+ * \since QGIS 3.14
+*/
+typedef QVector<QgsLayoutTableColumn> QgsLayoutTableSortColumns;
```
nyalldawson, line 64: Reusing QgsLayoutTableColumn seems wasteful here -- that class stores a lot more than the sort related attributes. What do you think about splitting the sort related attributes from QgsLayoutTableColumn into a new dedicated class?

src/gui/layout/qgslayoutattributeselectiondialog.cpp
```diff
@@ -211,57 +220,57 @@ QVariant QgsLayoutAttributeTableColumnModel::headerData( int section, Qt::Orient
...
+    case Heading:
+      columns()[index.row()].setHeading( value.toString() );
       emit dataChanged( index, index );
       return true;
-    case 2:
-      column->setHAlignment( Qt::AlignmentFlag( value.toInt() & Qt::AlignHorizontal_Mask ) );
-      column->setVAlignment( Qt::AlignmentFlag( value.toInt() & Qt::AlignVertical_Mask ) );
+
+    case Alignment:
+      columns()[index.row()].setHAlignment( Qt::AlignmentFlag( value.toInt() & Qt::AlignHorizontal_Mask ) );
```
nyalldawson, line 254: Lots of calls to `columns()[index.row()]` here -- could this be called once and stored for reuse in this function?

nyalldawson (commented):
Looking good! Great feature, and love the code cleanups!

src/core/layout/qgslayoutitemattributetable.cpp
3nids: Then setColumns would be used, right?
When using `setColumns`, I can check if a sorting order is provided and call `setSortColumns` at the same time.
Works for you?

src/core/layout/qgslayouttable.h
3nids: That's what I did on first run.
But it brings some nice advantage for the code
* that we can have a single model class instead of two
* less deprecation of methods (1), no need to deprecate the use of the class for another one in certain circumstances
* no need to cast in backward project compatibility (and API if we handle this in `setColumns`)

On the other hand we're talking about 3 enums and 1 double values used for nothing for a couple of columns.

I'd vote to stick as is (my agenda votes for it too ;) )

src/core/layout/qgslayoutitemattributetable.cpp
3nids: correct, I will fix this

src/core/layout/qgscompositionconverter.cpp
3nids: I will switch back to `std::copy_if` then

src/gui/layout/qgslayoutattributeselectiondialog.cpp
3nids: sadly not as we are updating a column. But I replaced `index.row()` by `r`, makes it easier to read.

src/core/layout/qgslayoutitemattributetable.cpp
nyalldawson: I thought about that too- but wouldn't it be hard to differentiate between a call to setColumns using the old API, vs correct use of the new api and calling setSortColumns and then later setColumns?

src/core/layout/qgslayouttable.h
nyalldawson: Ok, proceed....!

src/core/layout/qgslayoutitemattributetable.cpp
3nids: Well, if you correctly use the API there is no sorting info (rank) in the list when calling setColumns. If there is, then you're using the old API and we call setSortColumns.

src/gui/layout/qgslayoutattributeselectiondialog.cpp
3nids: hum, thanks for the free introduction course ;)

3nids:
review comments have been addressed.

nirvn:
@3nids , nice one!

PR #36240, opened 2020-05-06
Fix SIP QgsSnapToGridCanvasItem class declaration

### LINKED ISSUE #36033: Reproducible crash using QgsMapCanvas.items() with Oracle layers
We have found a situation that makes QGis Crash.
It didn't crash on QGis versions 3.4.12 and below.
But it does crash on Qgis versions 3.4.13 and up.
He highest version we have tested is 3.10.5 (QGIS code revision 4958ac7241).

The crash raises on invoking QgsMapCanvas.items(QRect), under that sequence of steps:
- Add at least one Oracle layer.
- Having a tool inheriting from QgsMapToolCapture, we activate it (super.activate())
That done, a crash will raise if we attempt to invoke 'iface.mapCanvas().items(QRect)'

**How to reproduce - Step 1: prepare an Oracle spatial table**

You need access to an Oracle database where you are able to create a table. Then you can use the attached ‘oracle_setup.sql’ to create a sample table:

```sql
-- Sample taken from https://docs.oracle.com/database/121/SPATL/simple-example-inserting-indexing-and-querying-spatial-data.htm#SPATL486
CREATE TABLE cola_markets (
  mkt_id NUMBER PRIMARY KEY,
  name VARCHAR2(32),
  shape SDO_GEOMETRY);
  
INSERT INTO cola_markets VALUES(
  1,
  'cola_a',
  SDO_GEOMETRY(
    2003,  -- two-dimensional polygon
    NULL,
    NULL,
    SDO_ELEM_INFO_ARRAY(1,1003,3), -- one rectangle (1003 = exterior)
    SDO_ORDINATE_ARRAY(1,1, 5,7) -- only 2 points needed to
          -- define rectangle (lower left and upper right) with
          -- Cartesian-coordinate data
  )
);

INSERT INTO user_sdo_geom_metadata
    (TABLE_NAME,
     COLUMN_NAME,
     DIMINFO,
     SRID)
  VALUES (
  'cola_markets',
  'shape',
  SDO_DIM_ARRAY(   -- 20X20 grid
    SDO_DIM_ELEMENT('X', 0, 20, 0.005),
    SDO_DIM_ELEMENT('Y', 0, 20, 0.005)
     ),
  NULL   -- SRID
);

commit;

CREATE INDEX cola_spatial_idx
   ON cola_markets(shape)
   INDEXTYPE IS MDSYS.SPATIAL_INDEX;
```

**How to reproduce - Step 2: have a QGis project with a layer pointing to this Oracle table**
Steps:

- Start QGis
- Follow the steps to add an Oracle layer (main menu > layer > add layer > add Oracle spatial layer)
1_addLayer_1
- The dialog for managing Oracle data sources will show up.
- Press ‘New’ to create a new Oracle connection. Fill in the input fields as needed.
1_addLayer_2_new
1_addLayer_3_add

**How to reproduce - Step 3: just try to get QgsMapCanvas.items() after a QgsMapToolCapture.activate()**
We have prepared a simple python script:
```python
from qgis.PyQt.QtCore import QRect
from qgis.gui import QgsMapToolCapture
from qgis.PyQt.QtWidgets import QAction
class MyTool(QgsMapToolCapture):
    def __init__(self, iface):
        self.iface=iface
        QgsMapToolCapture.__init__(self, iface.mapCanvas(), iface.cadDockWidget(), QgsMapToolCapture.CaptureLine)

    def activateTool(self):
        super(MyTool, self).activate()
    
    def crash(self):
        mc=self.iface.mapCanvas()
        r=QRect(311, -10, 635, 86)
        print("This is going to crash...")
        items = mc.items(r)

t=MyTool(iface)
t.activateTool()
t.crash()
````

After that, you will see 'This is going to crash...' on the console, and after a short moment, QGis will crash.

### PR DESCRIPTION
**Pull request by troopa81**
# Description

Fixes #36033 : missing SIP_RUN class specific declaration

### DIFF
--- a/python/gui/auto_generated/qgssnaptogridcanvasitem.sip.in
+++ b/python/gui/auto_generated/qgssnaptogridcanvasitem.sip.in
@@ -14,7 +14,7 @@
 #include <qgssnaptogridcanvasitem.h>
 %End
 
-class QgsSnapToGridCanvasItem : QObject, QgsMapCanvasItem
+class QgsSnapToGridCanvasItem : QgsMapCanvasItem
 {
 %Docstring
 
--- a/src/gui/qgssnaptogridcanvasitem.h
+++ b/src/gui/qgssnaptogridcanvasitem.h
@@ -37,8 +37,14 @@
  *
  * \since QGIS 3.4
  */
+#ifndef SIP_RUN
 class GUI_EXPORT QgsSnapToGridCanvasItem : public QObject, public QgsMapCanvasItem
 {
+#else
+class GUI_EXPORT QgsSnapToGridCanvasItem : public QgsMapCanvasItem
+{
+#endif
+
     Q_OBJECT
 
 #ifdef SIP_RUN

### CODE REVIEW

**Comment by m-kuhn** [comment]
Can you provide a detailed explanation of why this is required? This looks like one of the pull requests someone will look at in 3 years time and wonder "what was the reason for that".

**Comment by troopa81** [comment]
> Can you provide a detailed explanation of why this is required?

When you edit some vector, QgsSnapToGridItem is added to the scene and so the following Python instruction will make QGIS crash

```python
iface.mapCanvas().items()
```

It looks like SIP doesn't like double inheritance. To be honest, I find the solution in one for your previous commit #32755

m-kuhn:
Ah, thanks (for ref. most complete explanation here https://github.com/qgis/QGIS/pull/32755/commits/b648a641d4fb787b7a011739d85fcce7bf3cf0c3)

m-kuhn (approved):

PR #36243, opened 2020-05-06
Add basic temporal handling support for vector layers

nyalldawson:
This exposes some basic temporal capabilities for vector layers:

- static time range for layer (to match raster layer possibilities), this sets a single static time range which applies to the whole layer. ALL features from the layer will be shown whenever the canvas time overlaps the layer time range
- "Single field with datetime": Allows selection of a single date or datetime field from the layer. Features will be shown whenever this field value is within the canvas time range
- "Separate Fields for Start and End Date/Time": Allows selection of start and end date/datetime fields from the layer. Features will be shown whenever the time interval calculated from these fields
overlaps the canvas time range

(we should consider extending this in future, e.g. to add modes like "start time + fixed duration", "start time + duration from field", "start and end time via expressions", etc!) 

Some known limitations/inefficiencies:
- currently only date/datetime fields can be used. This was done to simplify the format handling and avoid the need to worry about string fields with different datetime formats. In future we should
allow selection of string fields and allow users to enter a custom datetime format string
- unlike the Time Manager plugin approach, the approach taken here is to rely completely on QGIS expressions and feature requests to do the filtering (Time Manager uses layer filter strings and attempts to set a native SQL filter syntax so that filtering is done on the backend). This is intentional, because it provides a unified filter approach regardless of the provider used (i.e. we don't need to worry about the different SQL syntaxes used natively by the different providers). The beauty of feature request expression compilation **should** mean that the QGIS expressions are magically turned into native backend queries, BUUUUUUUUUUUT... because we lack QGIS expression support for date time literals, we currently rely on the "to_datetime" expression function and coerce everything through strings. None of the expression compilers handle this function, so currently ALL
filtering is done on the QGIS side. We need to add functions for optimised datetime literal creation, and then ensure that the different compilers correctly map these literals across to the backend
filter syntax to allow all the filtering work to be done on the database side...

So currently, performance is much worse with large layers compared to Time Manager. (and the exposed feature set is smaller, e.g. no interpolation handling). But, the advantage is that we can use the native temporal framework and have vector layers animated alongside mesh and raster layers!


nyalldawson:
Sponsored by: Insomnia Inc

nyalldawson:
Another TODO: expose a manual "time offset" option to allow mixed time-zone layer animation.

src/core/qgstemporalutils.cpp
```diff
@@ -51,15 +53,57 @@ QgsDateTimeRange QgsTemporalUtils::calculateTemporalRangeForProject( QgsProject
...
+            const int fieldIndex = vectorLayer->fields().lookupField( properties->startField() );
+            if ( fieldIndex >= 0 )
+            {
+              layerRange = QgsDateTimeRange( vectorLayer->minimumValue( fieldIndex ).toDateTime(),
+                                             vectorLayer->maximumValue( fieldIndex ).toDateTime() );
+            }
+            break;
+          }
+
+          case QgsVectorLayerTemporalProperties::ModeFeatureDateTimeStartAndEndFromFields:
```
nirvn, line 81: I'd make the start or end field optional here. The logic would be:
- if start and end fields given, filter is start_field >= current_datetime and end_field <= current_datetime
- if only start field given, filter is start_field >= current_datetime
- if only end field given, filter is end_field >= current_datetime

As for the QgsDateTimeRange values
- if start and end fields given, range is min(start_field) max(end_field)
- if only start field given, range is min(start_field) max(start_field)
- if only end field given, range is min(end_field) max(end_field)

nirvn:
:tada:
Peek 2020-05-07 10-57

src/core/qgstemporalutils.cpp
nyalldawson: Done (but with the correct logic :wink:)

nyalldawson:
EYE CANDY!

Peek 2020-05-07 17-05

nirvn (approved):
Fantastic work.

anitagraser:
Awesome work! I've tested it with a TimeManager demo project and it works quite well. I just couldn't figure out why there was a one hour difference between what seems identical time stamps. Maybe some time zone black magic somewhere? I've fixed it for now using:

```
second(  age(  @map_end_time - to_interval('1 hour')  , "datetimetype" ) )
```

temporal1
temporal2

nyalldawson:
@anitagraser can you share the data?

anitagraser:
@nyalldawson https://drive.google.com/file/d/0Bwc-5JFVTnfITGFDTjkyakhEc1E/view?usp=sharing

nyalldawson:
@anitagraser thanks, confirmed -- there's a time zone issue happening here

nyalldawson:
@anitagraser see https://github.com/qgis/QGIS/pull/36331

dubravat:
Thank you so much!

PR #36249, opened 2020-05-07
Fix expression description HTML

### LINKED ISSUE #36191: Line breaks are ignored in the user expression help panel
As shown in the following image, in the help panel, the expression does not take into account line breaks, leading to comments and code being merged, less readable
image

troopa81:
## Description

Fixes #36191

When mixed with other html content, expression description HTML makes the rendering of carriage return in `<pre>` fails because of the following line in HTML

```html
<meta name="qrichtext" content="1" />
```
This PR proposes to remove this line from the returned HTML.

### DIFF
--- a/src/gui/qgsexpressionstoredialog.cpp
+++ b/src/gui/qgsexpressionstoredialog.cpp
@@ -59,3 +59,10 @@ QgsExpressionStoreDialog::QgsExpressionStoreDialog( const QString &label, const 
   mLabel->setText( labelFixed );
 }
 
+QString QgsExpressionStoreDialog::helpText()
+{
+  // remove meta qrichtext instruction from html. It fails rendering
+  // when mixing with other html content
+  // see issue https://github.com/qgis/QGIS/issues/36191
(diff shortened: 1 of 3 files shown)


src/gui/qgsexpressionstoredialog.h
```diff
@@ -55,7 +55,7 @@ class GUI_EXPORT QgsExpressionStoreDialog : public QDialog, private Ui::QgsExpre
     /**
      * Returns the help text
      */
-    QString helpText() { return  mHelpText->toHtml(); }
+    QString helpText();
```
m-kuhn, line 58:
```suggestion
    QString helpText() const;
```

src/gui/qgsexpressionstoredialog.cpp
```diff
@@ -59,3 +59,10 @@ QgsExpressionStoreDialog::QgsExpressionStoreDialog( const QString &label, const
   mLabel->setText( labelFixed );
 }
 
+QString QgsExpressionStoreDialog::helpText()
```
m-kuhn, line 62:
```suggestion
QString QgsExpressionStoreDialog::helpText() const
```

Issue #20350, opened 2015-02-09
Add "Select by Location" button to Attributes toolbar


qgib:
Author Name: **Corey Burger** (@CoreyBurger)
Original Redmine Issue: [12157](https://issues.qgis.org/issues/12157)

Redmine category:gui

---
The Attributes toolbar has select and select by expression, but select by location is buried under Vector>Research Tools>Analysis.

PR #36258, opened 2020-05-07
add "Selection" toolbar to the QGIS main window

### LINKED ISSUE #21045: Selection toolbar
Original Redmine Issue: [12967](https://issues.qgis.org/issues/12967)

Redmine category:gui

---
I think we should have a seprate toolbar for "Selection". Currently, there is "Attributes" toolbar which contains spatial selection (not really attribute).
Other items like "Select by location" can be moved to the "Selection" toolbar too.

alexbruy:
## Description
Currently "Attributes" toolbar contains also selection tools. While for tool like "Select by value" it is ok to be in the the "Attributes" toolbar, for spatial selection tools this is a bit strange location. So new "Selection" toolbar added and all selection tools (both spatial and attribute) was moved to this new toolbar.

In future some Processing tools like "Select by Location" can be added to this toolbar too (as asked in #20350).

Fixes #21045.

### DIFF
--- a/python/gui/auto_generated/qgisinterface.sip.in
+++ b/python/gui/auto_generated/qgisinterface.sip.in
@@ -302,6 +302,13 @@ Returns a reference to the main window "Shape Digitizing" toolbar.
 Returns a reference to the main window "Attributes" toolbar.
 %End
 
+    virtual QToolBar *selectionToolBar() = 0;
+%Docstring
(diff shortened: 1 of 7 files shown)


nyalldawson:
@nirvn what do you think? I'm +0

nirvn:
I like the idea, I'm +1. This attribute toolbar is crowded, and a bit of a potpourri, this PR helps a bit.

PR #36265, opened 2020-05-07
Abstract dxf export label provider away

m-kuhn:
This is a minimal change, moving the current dxf specific label feedback loop to a more generic `QgsLabelSink` template.

This is preparation work for a new processing algorithm to extract label positions into attributes, basically making PAL results exportable.


nyalldawson (commented):
Looking good. Can you move the new files to the core/labelling folder?

PR #36266, opened 2020-05-07
postgres: correct timestamp comparison in where clause (fix #24163)

benoitdm-oslandia:
Hi,
My name is Benoit De Mezzo and I'm a new member of Oslandia. I hope I could contribute to the Qgis project in a more broadly way. All tips are welcome !

## Description

The issue #24163 was due to sql casting to **::text** for unmanaged datatype as timestamp. The main problem was in the differents ways Qt and postgresql cast timestamp to string. Qt QDateTime cast to ISO 8601 (```yyyy-MM-ddTHH:mm:ss```) but postgresql cast to this format ```yyyy-MM-dd HH:mm:ss```. The generated sql was: 
```... WHERE ("time"::text='2017-02-20T13:41:47.000' OR ...``` 
ie. 
```... WHERE ('2017-02-20 13:41:47'='2017-02-20T13:41:47.000' OR ...``` 
so they cannot be compared as **::text**.

I solved this by forcing **QDatetime** in **quotedValue** to ISO 8601 and by adding a dedicated function **fieldExpressionForWhereClause** to avoid the cast to **::text**.

This PR could be backported to release-3_10 (I have tested it).

## Issue

Fixes #24163

### DIFF
--- a/src/providers/postgres/qgspostgresconn.cpp
+++ b/src/providers/postgres/qgspostgresconn.cpp
@@ -1212,6 +1212,7 @@ QString QgsPostgresConn::quotedValue( const QVariant &value )
   {
     case QVariant::Int:
     case QVariant::LongLong:
+    case QVariant::ULongLong:
     case QVariant::Double:
       return value.toString();
 
@@ -1596,6 +1597,42 @@ qint64 QgsPostgresConn::getBinaryInt( QgsPostgresResult &queryResult, int row, i
   }
 
   return oid;
+}
+
+QString QgsPostgresConn::fieldExpressionForWhereClause( const QgsField &fld, QVariant::Type valueType, QString expr )
+{
+  QString out;
+  const QString &type = fld.typeName();
+
+  if ( type == QLatin1String( "timestamp" ) || type == QLatin1String( "time" ) || type == QLatin1String( "date" ) )
+  {
+    out = expr.arg( quotedIdentifier( fld.name() ) );
+    // if field and value havev incompatible types, rollback to text cast
+    if ( valueType !=  QVariant::LastType && valueType != QVariant::DateTime && valueType != QVariant::Date && valueType != QVariant::Time )
+    {
+      out = out + "::text";
+    }
+  }
+
+  else if ( type == QLatin1String( "int8" ) || type == QLatin1String( "serial8" ) //
+            || type == QLatin1String( "int2" ) || type == QLatin1String( "int4" ) || type == QLatin1String( "oid" ) || type == QLatin1String( "serial" ) //
+            || type == QLatin1String( "real" ) || type == QLatin1String( "double precision" ) || type == QLatin1String( "float4" ) || type == QLatin1String( "float8" ) //
+            || type == QLatin1String( "numeric" ) )
+  {
+    out = expr.arg( quotedIdentifier( fld.name() ) );
+    // if field and value havev incompatible types, rollback to text cast
+    if ( valueType !=  QVariant::LastType && valueType != QVariant::Int && valueType != QVariant::LongLong && valueType != QVariant::ULongLong && valueType != QVariant::Double )
+    {
+      out = out + "::text";
+    }
+  }
+
+  else
+  {
+    out = fieldExpression( fld, expr ); // same as fieldExpression by default
+  }
+
+  return out;
 }
 
 QString QgsPostgresConn::fieldExpression( const QgsField &fld, QString expr )
--- a/src/providers/postgres/qgspostgresconn.cpp
+++ b/src/providers/postgres/qgspostgresconn.cpp
@@ -1212,7 +1212,6 @@ QString QgsPostgresConn::quotedValue( const QVariant &value )
   {
     case QVariant::Int:
     case QVariant::LongLong:
-    case QVariant::ULongLong:
     case QVariant::Double:
       return value.toString();
 
@@ -1624,7 +1623,7 @@ QString QgsPostgresConn::fieldExpressionForWhereClause( const QgsField &fld, QVa
   {
     out = expr.arg( quotedIdentifier( fld.name() ) );
     // if field and value havev incompatible types, rollback to text cast
-    if ( valueType !=  QVariant::LastType && valueType != QVariant::Int && valueType != QVariant::LongLong && valueType != QVariant::ULongLong && valueType != QVariant::Double )
+    if ( valueType !=  QVariant::LastType && valueType != QVariant::Int && valueType != QVariant::LongLong && valueType != QVariant::Double )
     {
       out = out + "::text";
     }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by nyalldawson** [comment]
Can you add some unit tests please?

**Comment by benoitdm-oslandia** [comment]
ok but without db mocking, it will be meaningless?!

File: tests/src/providers/testqgspostgresconn.cpp [hunk 1]
```diff
@@ -38,6 +40,36 @@ class TestQgsPostgresConn: public QObject
...
+      f2.setName( "pk" );
+      f2.setType( QVariant::ULongLong );
+      fields.append( f2 );
+
+      QList<int> pkAttrs;
+      pkAttrs.append( 0 );
+      pkAttrs.append( 1 );
+
+      QgsPostgresSharedData *shared = new QgsPostgresSharedData;
+      std::shared_ptr<QgsPostgresSharedData> sharedPtr( shared );
```
elpaso, line 63: Either allocate on stack or use a unique_ptr.

benoitdm-oslandia: is it better?

elpaso: Unfortunately not, your new version has a memory leak: you should probably allocate on stack or use a unique_ptr.

tests/src/providers/testqgspostgresconn.cpp
```diff
@@ -38,6 +40,35 @@ class TestQgsPostgresConn: public QObject
...
+      pkAttrs.append( 0 );
+      pkAttrs.append( 1 );
+
+      QgsPostgresSharedData *shared = new QgsPostgresSharedData;
+      QVariantList qvlist;
+      qvlist.append( QDateTime::fromString( "2020-05-07 17:56:00", "yyyy-MM-dd HH:mm:ss" ) );
+      qvlist.append( 123ULL );
+      shared->insertFid( 1LL, qvlist );
+
+      QCOMPARE( QgsPostgresUtils::whereClause( 1LL, fields, NULL, QgsPostgresPrimaryKeyType::PktFidMap, pkAttrs, std::shared_ptr<QgsPostgresSharedData>( shared ) ), QString( "\"ts\"='2020-05-07T17:56:00.000' AND \"pk\"=123" ) );
```
nyalldawson, line 69: Isn't `2020-05-07T17:56:00.000` incorrect here? In the PR description you state that you're trying to match Postgres' `yyyy-MM-dd HH:mm:ss` format.

benoitdm-oslandia: You are right! This is what I wrote at the beginning, I should have correct it. 

In the previous version of the code, the generated sql was: 
```... WHERE ("time"::text='2017-02-20T13:41:47.000' OR ...``` and the problem is the cast to ```::text```. 

In my various tests to check what kind of date can be handled by Postgresql, I made the mistake to believe only ```yyyy-MM-dd HH:mm:ss``` is supported but it can also handle ```yyyy-MM-ddTHH:mm:ss.SSS```. I was not able to make it read the same format but with timezone information like ```yyyy-MM-ddTHH:mm:ss.SSSZzz```. 

The simplest and most compatible format is ```Qt::ISODateWithMs``` ie. ```yyyy-MM-ddTHH:mm:ss.SSS```. I forced it to use this format because the default ```QDateTime.toString()``` differs from my pc (Qt5) and the one running the test on Github (?!?).

File: src/providers/postgres/qgspostgresconn.cpp [hunk 3]
```diff
@@ -1212,9 +1212,13 @@ QString QgsPostgresConn::quotedValue( const QVariant &value )
   {
     case QVariant::Int:
     case QVariant::LongLong:
+    case QVariant::ULongLong:
```
espinafre, line 1215: Hi, if I might add my two cents, PostgreSQL doesn't support unsigned integer types, so there is no need for ULongLong here.

This message from 2013 explains the reason: https://www.postgresql.org/message-id/CAEcSYX+Arn7y4FeYPp6ZgbiiiMfZYmsn9aUyotZB-MA1n5hTOw@mail.gmail.com

And it is still true as of 2020: https://www.postgresql.org/docs/12/datatype-numeric.html (look at the ranges); even for serial/bigserial pseudotypes, you can see that the maximum value is the same as the upper value for signed integers/big integers.

Even for date and time types, they are implemented as singed values: https://tapoueh.org/blog/2018/04/postgresql-data-types-date-timestamp-and-time-zones/

So, no need for accounting for unsigned values here and in line 1627 below.

benoitdm-oslandia: You are right, I will change it !

nyalldawson (approved):

nyalldawson:
Looking good -- you'll just need to get the tests passing again now!

tests/src/providers/testqgspostgresconn.cpp
```diff
@@ -38,6 +40,35 @@ class TestQgsPostgresConn: public QObject
...
+
+      QgsFields fields;
+      QgsField f;
+      f.setName( "ts" );
+      f.setType( QVariant::DateTime );
+      f.setTypeName( "timestamp" );
+      fields.append( f );
+      QgsField f2;
+      f2.setName( "pk" );
+      f2.setType( QVariant::ULongLong );
```
espinafre, line 55: That might be the reason your test is failing (2020-06-07 22:20:00-03:00): when telling a field type is one not recognized by QGIS, QgsPostgresConn::fieldExpression() returns a cast to text. You should just set your type to QVariant::LongLong here, and it'll be good. Thanks for your work!

benoitdm-oslandia: argg... so far so good

benoitdm-oslandia: done!

espinafre: Hi, @benoitdm-oslandia , I've cloned your repo and ran the tests locally, with the command
`ctest -V -R qgis_postgresconntest`

 I've found that if you change ULL to LL on line 66 here: https://github.com/qgis/QGIS/blob/f1bb526df8bbf5a304d5458085b7a0264235dab8/tests/src/providers/testqgspostgresconn.cpp#L63-L67 the test pass on my system.

If I may ask, what does QVariant::LastValue mean here? I couldn't find any docs about it, only the definition in /usr/include/qt5/QtCore/qvariant.h:

```c++
LastType = 0xffffffff // need this so that gcc >= 3.4 allocates 32 bits for Type
```

Is it thread safe? Thanks!

benoitdm-oslandia: sorry I was so lazy, I did not run the test to check!

PR #36267, opened 2020-05-07
Fix compilation with option WITH_GUI=FALSE

### LINKED ISSUE #29185: Fails to build with GUI disabled
Author Name: **Tom Palan** (@tpalan)
Original Redmine Issue: [21368](https://issues.qgis.org/issues/21368)
Affected QGIS version: 3.4.4
Redmine category:build/install
Assignee: Alessandro Pasotti

---
For building the QGIS server without Desktop app the following cmake command works:

cmake -G Ninja \
      -DCMAKE_VERBOSE_MAKEFILE=1 \
        -DCMAKE_INSTALL_PREFIX=/usr \
        -DBINDINGS_GLOBAL_INSTALL=TRUE \
        -DPEDANTIC=TRUE \
        -DSERVER_SKIP_ECW=TRUE \
        -DQGIS_CGIBIN_SUBDIR=/usr/lib/cgi-bin \
        -DWITH_APIDOC=TRUE \
        -DGENERATE_QHP=TRUE \
        -DWITH_CUSTOM_WIDGETS=FALSE \
        -DWITH_GLOBE=FALSE \
        -DWITH_SERVER=TRUE \
        -DWITH_SERVER_PLUGINS=TRUE \
        -DWITH_QWTPOLAR=FALSE \
        -DWITH_DESKTOP=FALSE \
        -DWITH_GUI=TRUE \
        -DDOXYGEN_ON_DEMAND=TRUE ..

But when disabling the GUI (WITH_GUI=FALSE), the build fails with error message:
ninja: error:
'../python/plugins/db_manager/db_plugins/postgis/plugins/versioning/pyg
ui', needed by
'python/plugins/db_manager/db_plugins/postgis/plugins/versioning/ui_Dlg
Versioning.py', missing and no known rule to make it

tpalan:
Fixes #29185

## Description
Python plugins are compiled even when WITH_GUI cmake option is set to FALSE, compilation then fails because no GUI components are included. This commit moves the compile of plugins into the IF (WITH_GUI) branch of the cmake file


nyalldawson:
Hmmmm - would it be possible to fix this in the plugins instead? There's a strong use case for including the processing plugin even when no gui is buillt....

tpalan:
I made a new commit, it compiles successfully, but cannot test the processing plugins in my setup.

nyalldawson:
That looks better -- and there's clearly bro l no use for db manager outside of the qgis gui!

PR #36270, opened 2020-05-07
[quick] Fix unhandled exception in QgsQuickPositionKit

wonder-sk:
If transformation of GPS position to project CRS fails, the unhandled exception would cause a crash.


nyalldawson (approved):

nyalldawson:
Backports?

wonder-sk:
> Backports?

Not sure if it is worth it - it seems that currently only Input uses qgis_quick out in the wild (which usually follows master)

PR #36273, opened 2020-05-07
Lazily initialize QgsDistanceArea in expressions

nyalldawson:
This is barely ever used (only when the $perimeter, $length or $area functions are used in an expression), and is expensive to construct.

We've been merrily creating them EVERY time we prepare an expression resulting in a lot of wasted effort. By lazily creating it only if needed we save a bunch of wasted processing and ultimately get
MUCH faster expression preparation.

(especially noticable when animating vector layers using data defined symbol settings or categorized/graduated/rule based renderers, refs #36243 , but I've also seen this hotspot when profiling other slow rendering/processing operations)

### DIFF
--- a/src/core/expression/qgsexpression.cpp
+++ b/src/core/expression/qgsexpression.cpp
@@ -272,15 +272,11 @@ void QgsExpression::initGeomCalculator( const QgsExpressionContext *context )
   // Set the geometry calculator from the context if it has not been set by setGeomCalculator()
   if ( context && ! d->mCalc )
   {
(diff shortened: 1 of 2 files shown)


nyalldawson:
I'm 50/50 on a 3.10 backport. It's a BIG speed boost... but medium risk. Opinions welcome!

nirvn:
I'd be +1 backporting to 3.10 on the basis of the de facto policy of skipping a point release.

nirvn (approved):

PR #36277, opened 2020-05-08
[canvas] Insure a refreshed canvas due to temporal range changes reflects the last provided range

nirvn:
## Description

This PR aims at improving the temporal controller's interaction with the canvas by insuring that the canvas always reflects the last selected datetime via the controller's slider.

Background: to make sure the UI isn't blocked, changing the current datetime via the controller's slider does _not_ abort an ongoing rendering (good). However, it does mean the user can end up settling on a datetime with the slider and unknowingly have a rendered canvas that does not reflect that user-picked datetime.

To fix that, a refresh is automatically triggered after a job is rendered if the temporal range changes while a rendering job is taking place.

### DIFF
--- a/src/core/qgstemporalproperty.h
+++ b/src/core/qgstemporalproperty.h
@@ -37,10 +37,14 @@ class CORE_EXPORT QgsTemporalProperty : public QObject
     Q_OBJECT
 
   public:
+
+    /**
+     * Flags attached to the temporal property.
+     */
     enum Flag
     {
       NoFlags                                         = 0,
-      FlagDontInvalidateCachedRendersWhenRangeChanges = 1  //!< Any cached rendering will not be invalidated when temporal range context is modified
+      FlagDontInvalidateCachedRendersWhenRangeChanges = 1  //!< Any cached rendering will not be invalidated when temporal range context is modified.
     };
     Q_DECLARE_FLAGS( Flags, Flag )
 
--- a/src/core/qgstemporalproperty.h
+++ b/src/core/qgstemporalproperty.h
@@ -43,7 +43,6 @@ class CORE_EXPORT QgsTemporalProperty : public QObject
      */
     enum Flag
     {
-      NoFlags                                         = 0,
       FlagDontInvalidateCachedRendersWhenRangeChanges = 1  //!< Any cached rendering will not be invalidated when temporal range context is modified.
     };
     Q_DECLARE_FLAGS( Flags, Flag )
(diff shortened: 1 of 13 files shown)


src/core/qgstemporalproperty.h
```diff
@@ -37,6 +37,12 @@ class CORE_EXPORT QgsTemporalProperty : public QObject
     Q_OBJECT
 
   public:
+    enum Flag
+    {
+      NoFlags                                         = 0,
```
nyalldawson, line 42: This can be removed

src/core/qgstemporalproperty.h
```diff
@@ -59,6 +65,11 @@ class CORE_EXPORT QgsTemporalProperty : public QObject
     */
     bool isActive() const;
 
+    /**
+     * Returns flags associated to the temporal property.
+     */
+    virtual const QgsTemporalProperty::Flags flags() const { return QgsTemporalProperty::NoFlags; }
```
nyalldawson, line 71:
```suggestion
    virtual QgsTemporalProperty::Flags flags() const { return nullptr; }
```

nyalldawson:
Regarding the first commit - the logic looks sound, but I'm wondering if you couldn't move this into the existing autoRefreshTriggered method. That should simplify the code a bit.

nirvn:
Thanks for the review.

I'm not sure I follow you on the autoRefreshTriggered suggestion. Isn't that function more generic than doing a temporal refresh? If so, I can't see how we'd add cache cleaning logic in there.

nirvn:
@nyalldawson , all green, good to merge?

nirvn:
(blah ignore for now, I still haven't changed your flag comments)

src/core/qgstemporalproperty.h
```diff
@@ -59,6 +68,11 @@ class CORE_EXPORT QgsTemporalProperty : public QObject
     */
     bool isActive() const;
 
+    /**
+     * Returns flags associated to the temporal property.
+     */
+    virtual const QgsTemporalProperty::Flags flags() const { return nullptr }
```
nyalldawson, line 74: Can you still drop the initial const here?

src/gui/qgsmapcanvas.h
```diff
@@ -177,6 +177,12 @@ class GUI_EXPORT QgsMapCanvas : public QGraphicsView
      */
     void clearCache();
 
+    /**
+     * Make sure to remove any rendered images of temporal-enabled layers from cache (does nothing if cache is not enabled)
+     * \since QGIS 3.14
+     */
+    void clearTemporalCache();
```
nyalldawson, line 184: Make this private

nyalldawson:
Nearly there 😁

src/core/qgstemporalproperty.h
nirvn: @nyalldawson , should I keep the const on overridden functions?

nyalldawson: Leave the second const, drop the first on all

nyalldawson (approved):

PR #36283, opened 2020-05-08
Respect HTML colors in labels

nyalldawson:
When enabled, this option will treat label contents as HTML, and
any FOREGROUND COLOR RELATED html formatting options will be
respected in the rendered label.

I repeat: ONLY HTML COLOR TAGS ARE RESPECTED. This is NOT a bug,
rather this feature has been designed as a "test of the waters"
with HTML formatting in labels, and accordingly only formatting
options which do not alter the font shape were considered.

On the plus side, it works correctly with ALL other label settings,
including shadows, buffers, curved labels, etc!

Sponsored by geoProRegio AG (www.geoproregio.ch)

Now some pretty pictures:

image

image


src/core/qgsrendercontext.cpp
```diff
@@ -98,7 +98,8 @@ QgsRenderContext &QgsRenderContext::operator=( const QgsRenderContext &rh )
   mHasRenderedFeatureHandlers = rh.mHasRenderedFeatureHandlers;
   mCustomRenderingFlags = rh.mCustomRenderingFlags;
   setIsTemporal( rh.isTemporal() );
-  setTemporalRange( rh.temporalRange() );
+  if ( rh.isTemporal() )
```
m-kuhn, line 101: Unrelated change?

nyalldawson:
Now also supports underline, strikethrough, and overline:

image

NathanW2:
;) hello html labels my old friend. Good to see you are back from the dead

m-kuhn (approved):
I have reviewed this with respect to

 - Basic code style sanity checks
 - Feature awesomeness
 - API looking reasonable
 - Tests are in place

And can give a +1 for all of that.

nyalldawson:
Ok, this is ready for merge (if CI is happy)

m-kuhn:
Thanks for the test work

timlinux:
I wonder more generally what the use case of this is - are there any maps in the wild you can show that use mixed colours etc in the labels?

Just curious...more options are always nice :-)

andreasneumann:
> I wonder more generally what the use case of this is - are there any maps in the wild you can show that use mixed colours etc in the labels?
> 
> Just curious...more options are always nice :-)

Yes - consider maps that show the state of industrial devices or infrastructure - e.g. waste-water structures with pump, valves, different levels. Color and bold can express the state of different sensors.

PR #36285, opened 2020-05-08
Delete cascade on composition relations

signedav:
## Description
On deleting a feature having children defined by a relation with the strength `composition` (especially used for a reference table on many-to-many relations), the children are deleted as well.

- It's cascade over all levels - means childrenchildren and childrenchildrenchildren are deleted as well. But an endless loop is avoided.
- When the child-layer is not editable, editable is triggered.
- Message reporting how many features on what layers are deleted
- Warning on delete for layers with composition relations and joins (joined features are deleted on cascade delete - this is an existing implementation)

## Looks like this
One parent-child relation (forest to tree) and one many-to-many creating a recursion on the same feature.

image

Warning message:
cascade1

Feedback message:
cascade2

## Technical details
The main part is implemented in the `QgsVectorLayer`. I thought about it to have everything in `QgsVectorLayerUtils` but since there are several methods affected, needing to call each other (like e.g. `deleteFeature` on `QgsVectorLayerJoinBuffer`), this would have lead to a lot of duplicate code.

Not every call of `deleteFeature` / `deleteFeatures` / `deleteSelectedFeatures` makes a cascade delete:

Doing cascade delete:
- `QgsAttributeTableDialog::deleteFeature`
- `QgisApp::deleteSelected`
- `QgsRelationEditorWidget::deleteFeatures`

Not doing cascade delete:
- `QgisApp::mergeSelectedFeatures`
- `QgsOfflineEditing::applyFeaturesRemoved`
- `QgsRelationEditorWidget::unlinkFeatures`
- `TopolError::fixUnion`
- `TopolError::deleteFirst`
- `TopolError::deleteSecond`
- `QgsAuxiliaryLayer::clear`
- `QgisApp::cutSelectionToClipboard`
- `QgsTrackedVectorLayerTools::rollback`
- `performTransaction` in `QgsWfs`

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -9003,7 +9003,8 @@ void QgisApp::deleteSelected( QgsMapLayer *layer, QWidget *parent, bool checkFea
 
   vlayer->beginEditCommand( tr( "Features deleted" ) );
   int deletedCount = 0;
-  if ( !vlayer->deleteSelectedFeatures( &deletedCount, true ) )
+  QgsVectorLayer::DeleteContext context { true };
+  if ( !vlayer->deleteSelectedFeatures( &deletedCount, &context ) )
   {
     visibleMessageBar()->pushMessage( tr( "Problem deleting features" ),
                                       tr( "A problem occurred during deletion from layer \"%1\". %n feature(s) not deleted.", nullptr, numberOfSelectedFeatures - deletedCount ).arg( vlayer->name() ),
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -9001,10 +9001,10 @@ void QgisApp::deleteSelected( QgsMapLayer *layer, QWidget *parent, bool checkFea
     }
   }
 
-  if ( QgsVectorLayerUtils::impactsCascadeFeatures(vlayer, QgsProject::instance() ) )
+  if ( QgsVectorLayerUtils::impactsCascadeFeatures( vlayer, QgsProject::instance() ) )
   {
     // for extra safety to make sure we know that the delete can have impact on children and joins
-    int res = QMessageBox::warning( mMapCanvas, tr( "Possible impact on descendants of layer \"%1\"").arg( vlayer->name() ),
+    int res = QMessageBox::warning( mMapCanvas, tr( "Possible impact on descendants of layer \"%1\"" ).arg( vlayer->name() ),
                                     tr( "A delete on this layer could have an impact on referencing or joined layers and their descendants. Would you still like to continue?" ),
                                     QMessageBox::Yes | QMessageBox::No );
     if ( res != QMessageBox::Yes )
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -9040,7 +9040,7 @@ void QgisApp::deleteSelected( QgsMapLayer *layer, QWidget *parent, bool checkFea
       QString feedbackMessage;
       for ( QgsVectorLayer *contextLayer : contextLayers )
       {
-        feedbackMessage += tr( " %1 on layer %2." ).arg( context.handledFeatures( contextLayer ).size() ).arg( contextLayer->name() );
+        feedbackMessage += tr( "%1 on layer %2. " ).arg( context.handledFeatures( contextLayer ).size() ).arg( contextLayer->name() );
         deletedCount += context.handledFeatures( contextLayer ).size();
       }
       visibleMessageBar()->pushMessage( tr( "%1 features deleted: %2" ).arg( deletedCount ).arg( feedbackMessage ), Qgis::Success );
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -9023,7 +9023,7 @@ void QgisApp::deleteSelected( QgsMapLayer *layer, QWidget *parent, bool checkFea
 
   vlayer->beginEditCommand( tr( "Features deleted" ) );
   int deletedCount = 0;
-  QgsVectorLayer::DeleteContext context { true };
+  QgsVectorLayer::DeleteContext context( true, QgsProject::instance() );
   if ( !vlayer->deleteSelectedFeatures( &deletedCount, &context ) )
   {
     visibleMessageBar()->pushMessage( tr( "Problem deleting features" ),
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -9033,12 +9033,12 @@ void QgisApp::deleteSelected( QgsMapLayer *layer, QWidget *parent, bool checkFea
   else
   {
     //if it effected more than one layer, print feedback for all descendants
-    if ( context.handledFeatures().size() > 1 )
+    if ( context.mHandledFeatures.size() > 1 )
     {
       deletedCount = 0;
       QString feedbackMessage;
       QMap<QgsVectorLayer *, QgsFeatureIds>::const_iterator i;
-      for ( i = context.handledFeatures().begin(); i != context.handledFeatures().end(); ++i )
+      for ( i = context.mHandledFeatures.begin(); i != context.mHandledFeatures.end(); ++i )
       {
         feedbackMessage += tr( " %1 on layer %2." ).arg( i.value().size() ).arg( i.key()->name() );
         deletedCount += i.value().size();
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -9012,7 +9012,21 @@ void QgisApp::deleteSelected( QgsMapLayer *layer, QWidget *parent, bool checkFea
   }
   else
   {
-    showStatusMessage( tr( "%n feature(s) deleted.", "number of features deleted", numberOfSelectedFeatures ) );
+    //if it effected more than one layer, print feedback for all descendants
+    if ( context.handledFeatures.count() > 1 )
+    {
+      deletedCount = 0;
+      QString feedbackMessage;
+      QMap<QgsVectorLayer *, QgsFeatureIds>::const_iterator i;
+      for ( i = context.handledFeatures.begin(); i != context.handledFeatures.end(); ++i )
+      {
+        feedbackMessage += tr( " %1 on layer %2." ).arg( i.value().count() ).arg( i.key()->name() );
+        deletedCount += i.value().count();
+      }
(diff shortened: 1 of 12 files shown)


src/core/qgsvectorlayerutils.h
```diff
@@ -319,6 +319,16 @@ class CORE_EXPORT QgsVectorLayerUtils
...
     static QString getFeatureDisplayString( const QgsVectorLayer *layer, const QgsFeature &feature );
+
+    /**
+     * \returns true if at least one feature of the \a fids on \a layer is connected as parent in at
+     * least one composition relation of the \a project or contains joins, where cascade delete is set.
+     * The \a context is used to get more details.
+     *
+     * \since QGIS 3.14
+     */
+    static bool impactsCascadeFeatures( const QgsVectorLayer *layer, const QgsFeatureIds &fids, const QgsProject *project, QgsDuplicateFeatureContext &context );
```
m-kuhn, line 330: Can you flag the context as SIP_OUT (or SIP_INOUT whatever is more appropriate)

src/gui/qgsrelationeditorwidget.cpp
```diff
@@ -730,9 +728,43 @@ void QgsRelationEditorWidget::deleteFeatures( const QgsFeatureIds &featureids )
...
+    int childrenCount = 0;
+    const auto infoContextLayers = infoContext.layers();
+    for ( QgsVectorLayer *chl : infoContextLayers )
+    {
+      childrenCount += infoContext.duplicatedFeatures( chl ).size();
+      childrenInfo += ( tr( "%1 feature(s) on layer \"%2\", " ).arg( infoContext.duplicatedFeatures( chl ).size() ).arg( chl->name() ) );
+    }
+
+    // for extra safety to make sure we know that the delete can have impact on children and joins
+    int res = QMessageBox::question( this, tr( "Delete at least %3 feature(s) on other layer(s)" ).arg( childrenCount ),
```
m-kuhn, line 744: %3?

src/gui/qgsrelationeditorwidget.cpp
```diff
@@ -730,9 +728,43 @@ void QgsRelationEditorWidget::deleteFeatures( const QgsFeatureIds &featureids )
...
+    if ( contextLayers.size() > 1 )
+    {
+      int deletedCount = 0;
+      QString feedbackMessage;
+      for ( QgsVectorLayer *contextLayer : contextLayers )
+      {
+        feedbackMessage += tr( " %1 on layer %2." ).arg( context.handledFeatures( contextLayer ).size() ).arg( contextLayer->name() );
+        deletedCount += context.handledFeatures( contextLayer ).size();
+      }
+      mEditorContext.mainMessageBar()->pushMessage( tr( "%1 features deleted: %2" ).arg( deletedCount ).arg( feedbackMessage ), Qgis::Success );
```
m-kuhn, line 765:
```suggestion
      mEditorContext.mainMessageBar()->pushMessage( tr( "%1 features deleted:%2" ).arg( deletedCount ).arg( feedbackMessage ), Qgis::Success );
```

src/core/qgsvectorlayerjoinbuffer.h
```diff
@@ -197,24 +197,26 @@ class CORE_EXPORT QgsVectorLayerJoinBuffer : public QObject, public QgsFeatureSi
      * parameter is the one coming from the target layer.
      *
      * \param fid The feature id from the target layer to delete
+     * \param context The chain of features who will be deleted for feedback and to avoid endless recursions
```
m-kuhn, line 200:
```suggestion
     * \param context The chain of features which will be deleted for feedback and to avoid endless recursions
```

src/core/qgsvectorlayer.h
```diff
@@ -504,6 +504,25 @@ class CORE_EXPORT QgsVectorLayer : public QgsMapLayer, public QgsExpressionConte
 
     };
 
+    /**
+     * Context for deleting features
+     * \since QGIS 3.14
+     */
+    struct  DeleteContext
```
m-kuhn, line 511:
```suggestion
    struct CORE_EXPORT DeleteContext
```

m-kuhn: @jef-n compensation for the blood, sweat and tears, nightly builds instead of a "fix windows build" commit tomorrow.

src/app/qgisapp.cpp
```diff
@@ -9001,17 +9001,52 @@ void QgisApp::deleteSelected( QgsMapLayer *layer, QWidget *parent, bool checkFea
...
   {
     visibleMessageBar()->pushMessage( tr( "Problem deleting features" ),
                                       tr( "A problem occurred during deletion from layer \"%1\". %n feature(s) not deleted.", nullptr, numberOfSelectedFeatures - deletedCount ).arg( vlayer->name() ),
                                       Qgis::Warning );
   }
   else
   {
-    showStatusMessage( tr( "%n feature(s) deleted.", "number of features deleted", numberOfSelectedFeatures ) );
+    const auto contextLayers = context.handledLayers();
+    //if it effected more than one layer, print feedback for all descendants
```
m-kuhn, line 9036:
```suggestion
    // if it affects more than one layer, print feedback for all descendants
```

src/core/qgsvectorlayerutils.h
```diff
@@ -319,6 +319,16 @@ class CORE_EXPORT QgsVectorLayerUtils
      * \since QGIS 3.12
      */
     static QString getFeatureDisplayString( const QgsVectorLayer *layer, const QgsFeature &feature );
+
+    /**
+     * \returns true if at least one feature of the \a fids on \a layer is connected as parent in at
```
m-kuhn, line 324:
```suggestion
     * \returns TRUE if at least one feature of the \a fids on \a layer is connected as parent in at
```

Will be replaced in api docs with the correct values for for Python/C++

src/core/qgsvectorlayerutils.cpp
```diff
@@ -934,3 +934,83 @@ QString QgsVectorLayerUtils::getFeatureDisplayString( const QgsVectorLayer *laye
...
+        else
+        {
+          // add layer and feature id
+          context.setDuplicatedFeatures( relation.referencingLayer(), childFeatureIds );
+        }
+      }
+    }
+  }
+
+  if ( layer->joinBuffer()->containsJoins() )
```
m-kuhn, line 980: Does that always exit or do we need to have a safety check for `joinBuffer()` being nullptr?

src/core/qgsvectorlayerutils.h
```diff
@@ -319,6 +319,16 @@ class CORE_EXPORT QgsVectorLayerUtils
...
+     * least one composition relation of the \a project or contains joins, where cascade delete is set.
+     * The \a context is used to get more details.
```
m-kuhn, line 326:
```suggestion
     * Details about cascading effects will be written to \a context.
```
How does that sound?

src/core/qgsvectorlayer.cpp
```diff
@@ -3164,36 +3164,88 @@ bool QgsVectorLayer::deleteAttributes( const QList<int> &attrs )
...
-    mJoinBuffer->deleteFeatures( fids );
+  return res;
+}
 
-  bool res = mEditBuffer->deleteFeatures( fids );
+bool QgsVectorLayer::deleteFeatures( const QgsFeatureIds fids, QgsVectorLayer::DeleteContext *context )
+{
+  bool res = true;
+  const auto constFids = fids;
+  for ( QgsFeatureId fid : constFids )
```
m-kuhn, line 3247:
```suggestion
  for ( QgsFeatureId fid : fids )
```
No reason to take a const copy of an already const list.

src/core/qgsvectorlayer.cpp
```diff
@@ -3164,36 +3164,88 @@ bool QgsVectorLayer::deleteAttributes( const QList<int> &attrs )
...
+    updateExtents();
   }
 
-  if ( mJoinBuffer->containsJoins() )
-    mJoinBuffer->deleteFeatures( fids );
+  return res;
+}
 
-  bool res = mEditBuffer->deleteFeatures( fids );
+bool QgsVectorLayer::deleteFeatures( const QgsFeatureIds fids, QgsVectorLayer::DeleteContext *context )
```
m-kuhn, line 3243:
```suggestion
bool QgsVectorLayer::deleteFeatures( const QgsFeatureIds &fids, QgsVectorLayer::DeleteContext *context )
```

src/core/qgsvectorlayer.cpp
```diff
@@ -3164,36 +3164,88 @@ bool QgsVectorLayer::deleteAttributes( const QList<int> &attrs )
...
+        context->mHandledFeatures.insert( this, handledFeatureIds );
+      }
+    }
+    else
+    {
+      // add layer and feature id
+      context->mHandledFeatures.insert( this, QgsFeatureIds() << fid );
+    }
+
+    const QList<QgsRelation> relations = QgsProject::instance()->relationManager()->referencedRelations( this );
```
m-kuhn, line 3195: Can we get the `QgsProject` from the context instead to avoid accessing the infamous singleton `QgsProject::instance()`.

m-kuhn (approved):
It's looking good overall, a couple of comments still remain in place but should be straightforward to solve.

src/core/qgsvectorlayer.cpp
signedav: In case of a recursion where the child-child of a feature is the parent of the currently deleted feature itself, it can be that the `QgsFeatureIds` referenced object does not exist anymore (since it's the selected features in e.g. the RelationEditorWidget). That's why it should not be a reference. It could be avoided by the implementation below as well, but I prefer that the fids are not lost.

m-kuhn: I can't really follow how this would happen, but if it is required, better take the copy explicitly in the function body to avoid that this is changed in a future cleanup. Like you are doing that anyway already on line 3246.

signedav: Yeah I did it double-save because of that. It's still possible that someone continues using the fids instead of the const one. So would you still keep the reference? Or is it fine like it is?

src/core/qgsvectorlayerutils.h
signedav: Pretty good :-)

src/core/qgsvectorlayerutils.cpp
signedav: `joinBuffer()` returns `mJoinBuffer` of `QgsVectorLayer` what is created in it's constructor. Should I make a safety check anyway?

m-kuhn: No, in this case it's fine

PR #36289, opened 2020-05-08
allow adding attributes in the New Scratch Layer dialog (fix #24397)

### LINKED ISSUE #24397: make the 'create scratch layer dialog' more like the 'create new shape file' dialog'
Author Name: **Richard Duivenvoorde** (@rduivenvoorde)
Original Redmine Issue: [16490](https://issues.qgis.org/issues/16490)

Redmine category:gui

---

Using scratch layers pretty often, I wonder why during the first creation dialog I'm not presented with (about) the same options as when I create a new shape file. That is: it would be cool to be able to add some attributes immediately. 

Maybe this can be part of the grant proposal 'ADD CONSISTENCY TO UI CONTROLS' ?

Ah... and please also add this to the python api :-)

alexbruy:
## Description
Allows to add attributes directly in the New Scratch Layer dialog making it more consistent with other dialogs like New Shapefile/Geopackage.
scratch
Possibility to create scratch layer without any attributes is still here, just don't add any fields to the list.

Fixes #24397.

### DIFF
--- a/python/gui/auto_generated/qgsnewmemorylayerdialog.sip.in
+++ b/python/gui/auto_generated/qgsnewmemorylayerdialog.sip.in
@@ -58,6 +58,13 @@ Returns the selected CRS for the new layer.
 Returns the layer name
 %End
 
+    QgsFields fields() const;
+%Docstring
+Returns attributes for the new layer.
+
+.. versionadded:: 3.14
+%End
+
 };
 
 /************************************************************************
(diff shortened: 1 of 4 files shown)


nirvn:
Oh man, @alexbruy , if we ever meet face to face, drinks re on me! I can't count the number of times *per month* I've hoped for this.

alexbruy:
Glad you find it useful, @nirvn. This bugged me a lot too.

uclaros:
Nice!!
Field types could be pulled from `QgsMemoryProvider`'s `nativeTypes()` so it's not missing types, like a true text or blob etc. 
Length and precision validators and enable/disable should be updated on type change.

This is a nice feature, but also a reminder that we have four different dialogs and code paths for creating a new layer (geopackage, shapefile, spatialite and memory layer).

PR #36131, opened 2020-05-01
Add subdivisions in ticksscalebar right segments

agiudiceandrea:
## Description

Adds the ability to create subdivisions for segments included in the right part of the ticksscalebar.

Line Ticks Up
image

Line Ticks Middle
image

Line Ticks Down
image

Scalebar item properties - Segments groupbox
image

I have not changed the behaviour related to the segments on the left side of the scalebar.

PR #36296, opened 2020-05-08
Fix tick-style scalebar when subdivisions ticks height > segments ticks height

agiudiceandrea:
## Description

Properly handle segments/subdivisions ticks drawing, box size calculation and horizontal line and labels position, when subdivisions ticks (for right segments) height is greater than segments ticks height in tick-style scalebar, enhancing the features introduced with https://github.com/qgis/QGIS/pull/36131 and https://github.com/qgis/QGIS/pull/36222.

Fixes https://github.com/qgis/QGIS/pull/36222#issuecomment-624455448.

See also https://github.com/nyalldawson/QGIS/pull/1.


nyalldawson:
Thanks!

PR #36297, opened 2020-05-08
Try to auto-populate reasonable initial choices for vector layer temporal settings

nyalldawson:
When a new vector layer is loaded, we scan through the fields for likely candidates for the start and end field choices. If found, we initially setup the temporal properties to these guessed fields, but we DON'T automatically enable temporal properties for the layer.

It's just a nice little shortcut, much like how we auto-guess the appropriate title field for a newly added layer

### DIFF
--- a/python/core/auto_generated/qgsvectorlayertemporalproperties.sip.in
+++ b/python/core/auto_generated/qgsvectorlayertemporalproperties.sip.in
@@ -150,6 +150,12 @@ settings (such as startField()) when building the filter string.
    ModeFixedTemporalRange should ALL be filtered out.
 %End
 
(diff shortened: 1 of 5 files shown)


elpaso:
Nice! I think this would also be useful for DB rasters, how about moving it up to QgsMapLayer to avoid duplication?

nyalldawson:
@elpaso
> Nice! I think this would also be useful for DB rasters, how about moving it up to QgsMapLayer to avoid duplication?

I don't think that's possible - it's heavily tied to the members of QgsVectorLayerTemporalProperties and QgsVectorLayerTemporalProperties::TemporalMode values. Unless I'm missing something?

nirvn (approved):

PR #36301, opened 2020-05-09
Clear ongoing drawing on map canvas when closing layer's edit session

### LINKED ISSUE #36292: Incomplete feature drawing should be removed when stopping edit session
If a line or polygon layer goes to non-editing mode and there is an ongoing (never finished) drawing on canvas for that layer, the drawing should be removed.

Screencast:

incomplete_digitized_feature

gacarrillor:
When a layer goes to non-edit mode, call `clean()` methods for map tools that could have temporary drawings on the map canvas.

Before: 

After: 
solution_for_incomplete_digitized_feature

Fix #36292

### DIFF
--- a/src/app/vertextool/qgsvertextool.cpp
+++ b/src/app/vertextool/qgsvertextool.cpp
@@ -2861,3 +2861,16 @@ void QgsVertexTool::cleanEditor( QgsFeatureId id )
     cleanupVertexEditor();
   };
 }
+
+void QgsVertexTool::clean()
+{
+  if ( mDraggingVertex || mDraggingEdge )
+  {
+    stopDragging();
+  }
+  if ( mSelectionRubberBand )
+  {
+    stopSelectionRubberBand();
+    mSelectionRubberBandStartPos.reset();
+  }
+}
--- a/src/gui/qgsmapcanvas.cpp
+++ b/src/gui/qgsmapcanvas.cpp
@@ -2030,6 +2030,11 @@ void QgsMapCanvas::unsetMapTool( QgsMapTool *tool )
   }
 }
 
+void QgsMapCanvas::setProject( QgsProject *project )
+{
+  mProject = project;
+}
+
 void QgsMapCanvas::setCanvasColor( const QColor &color )
 {
   if ( canvasColor() == color )
@@ -2245,6 +2250,11 @@ void QgsMapCanvas::projectThemesChanged()
 QgsMapTool *QgsMapCanvas::mapTool()
 {
   return mMapTool;
+}
+
+QgsProject *QgsMapCanvas::project()
+{
+  return mProject;
 }
 
 void QgsMapCanvas::panActionEnd( QPoint releasePoint )
--- a/src/gui/qgsmapcanvas.h
+++ b/src/gui/qgsmapcanvas.h
@@ -28,6 +28,7 @@
 #include "qgscustomdrophandler.h"
 #include "qgstemporalrangeobject.h"
 #include "qgsmapcanvasinteractionblocker.h"
+#include "qgsproject.h"
 
 #include <QDomDocument>
 #include <QGraphicsView>
@@ -354,6 +355,21 @@ class GUI_EXPORT QgsMapCanvas : public QGraphicsView
 
     //! Returns the currently active tool
     QgsMapTool *mapTool();
+
+    /**
+     * Sets the \a project linked to this canvas.
+     *
+     * \since QGIS 3.14
+     */
+    void setProject( QgsProject *project );
+
+    /**
+     * Returns the project linked to this canvas.
+     * The returned value may be NULLPTR.
+     *
+     * \since QGIS 3.14
+     */
+    QgsProject *project();
 
     //! Write property of QColor bgColor.
     void setCanvasColor( const QColor &_newVal );
@@ -1110,6 +1126,9 @@ class GUI_EXPORT QgsMapCanvas : public QGraphicsView
 
     //! previous tool if current is for zooming/panning
     QgsMapTool *mLastNonZoomMapTool = nullptr;
+
+    //! Pointer to project linked to this canvas
+    QgsProject *mProject = nullptr;
 
     //! Context menu
     QMenu *mMenu = nullptr;
--- a/src/gui/qgsmaptooledit.cpp
+++ b/src/gui/qgsmaptooledit.cpp
@@ -27,6 +27,11 @@
 QgsMapToolEdit::QgsMapToolEdit( QgsMapCanvas *canvas )
   : QgsMapTool( canvas )
 {
+  if ( mCanvas->project() )
+  {
+    connect( mCanvas->project(), &QgsProject::layersAdded, this, &QgsMapToolEdit::connectLayers );
+    connectLayers( mCanvas->project()->mapLayers().values() );  // Connect existing layers
+  }
 }
 
 double QgsMapToolEdit::defaultZValue() const
@@ -170,4 +175,40 @@ void QgsMapToolEdit::notifyNotVectorLayer()
 void QgsMapToolEdit::notifyNotEditableLayer()
 {
   emit messageEmitted( tr( "Layer not editable" ) );
+}
+
+void QgsMapToolEdit::connectLayers( const QList<QgsMapLayer *> &layers )
+{
+  for ( QgsMapLayer *layer : layers )
+  {
+    QgsVectorLayer *vlayer = qobject_cast<QgsVectorLayer *>( layer );
+    if ( vlayer )
+    {
+      connect( vlayer, &QgsVectorLayer::editingStopped, this, &QgsMapToolEdit::cleanCanvas );
+    }
+  }
+}
+
+void QgsMapToolEdit::cleanCanvas()
+{
+  if ( editableVectorLayers().isEmpty() )
+  {
+    clean();
+  }
+}
+
+QList<QgsVectorLayer *> QgsMapToolEdit::editableVectorLayers()
+{
+  QList<QgsVectorLayer *> editableLayers;
+  if ( mCanvas->project() )
+  {
+    const auto layers = mCanvas->project()->mapLayers().values();
+    for ( QgsMapLayer *layer : layers )
+    {
+      QgsVectorLayer *vlayer = qobject_cast<QgsVectorLayer *>( layer );
+      if ( vlayer && vlayer->isEditable() && vlayer->isSpatial() )
+        editableLayers << vlayer;
+    }
+  }
+  return editableLayers;
 }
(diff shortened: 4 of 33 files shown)


nyalldawson:
I'm pretty sure the behaviour described in #36292 is by design, because it allows users to pan/zoom and then continue digitising. @3nids can you confirm?

gacarrillor:
You can change the current map tool without problem while editing (this PR does not touch that behavior), but after you've closed the edit session no more rubber bands should be kept on the canvas. 

This is possible in master (edit session closed):  

image

This PR avoids it.

nyalldawson:
Ah, thanks for the clarification! This does seem like a good move

m-kuhn:
Good move.
Should this reside inside qgisapp.cpp? Or be connected to a signal of the layer?

gacarrillor:
Yes, that's another option. Then I would connect `layer->editingStopped` SIGNAL to several map tools' `clean()` SLOT.

gacarrillor:
> Good move.
Should this reside inside qgisapp.cpp? Or be connected to a signal of the layer?

Yes, that's another option.

I would connect `vectorLayer.editingStopped` SIGNAL to `mapToolCapture.clean()` SLOT (for several map tools). Connections would be set inside `QgisApp::layersWereAdded()`.

What do you think?

uclaros:
This will have the side effect of killing all rubberbands regardless of the layer that has editing stopped.
When editing a single layer it totally makes sense; having leftover rubberbands is confusing!
When editing many layers simultaneously though it's different.
Since the digitizing tools are not aware of the layer they are digitizing on, and the user has the flexibility to change the active layer at any time before right-clicking, or even change tool and start a second digitization in parallel, this will be a little against the logic of this functionality because stop editing on *any* layer will also mean stop digitizing on any other layer.
Maybe it would be better to kill the rubberbands if editing is stopped to all supported layers?

src/app/vertextool/qgsvertextool.cpp
```diff
@@ -2802,3 +2802,11 @@ void QgsVertexTool::cleanEditor( QgsFeatureId id )
     cleanupVertexEditor();
   };
 }
+
+void QgsVertexTool::clean()
+{
+  if ( mDraggingVertex )
```
uclaros, line 2808: Should also check against `mDraggingEdge`. There is also a third case of a rubberband, that of the vertex selection rectangle.

gacarrillor:
>  Maybe it would be better to kill the rubberbands if editing is stopped to all supported layers?

It makes sense!

src/app/vertextool/qgsvertextool.cpp
gacarrillor: `mDraggingEdge` --> right.

On the other hand, `cadCanvasReleaseEvent()` calls `stopSelectionRubberBand()`. So, the selection rectangle rubber band is being correctly removed. I'd say we don't need to check for that one, but let me know if (and why) you consider we should.

m-kuhn:
I still think this should reside out of app if possible, in order not to forget it if new tools are added / make using these tools in plugins easier, keeping things modular by keeping track of all layers inside the map tools through QgsProject?

And/or: the rubberband could be hidden if "current layer" is not in edit mode and cleared if all layer edits are stopped?

src/app/vertextool/qgsvertextool.cpp
uclaros: There's an edge case, if one stops editing by using a key bind while still dragging for a selection with the mouse. It's not easy to happen, but a `clean()` method should clean everything.

gacarrillor: Just replicated it. Agree, it should be removed as well, even if it's an edge case.

gacarrillor: Solved in https://github.com/qgis/QGIS/pull/36301/commits/1c55cb9ab03579a0879a43d5fba8b1d30356ccfe

gacarrillor:
> I still think this should reside out of app if possible, in order not to forget it if new tools are added / make using these tools in plugins easier, keeping things modular by keeping track of all layers inside the map tools through QgsProject?

I agree, that is cleaner. Implemented in QgsMapToolEdit. Now QgisApp is not touched by this PR.

src/gui/qgsmaptooledit.cpp
```diff
@@ -27,6 +27,8 @@
 QgsMapToolEdit::QgsMapToolEdit( QgsMapCanvas *canvas )
   : QgsMapTool( canvas )
 {
+  connect( QgsProject::instance(), &QgsProject::layersAdded, this, &QgsMapToolEdit::connectLayers );
```
m-kuhn, line 30: Thinking about it again, how about using `QgsMapCanvas::layersChanged`? Using `QgsProject::instance()` is discouraged and actually the tools always work on canvas layers?

gacarrillor: Then things get complicated. 

In QGIS I can edit a layer even if it's not visible on the map canvas. Also, the map canvas only gives us access to layers that are currently being displayed. 

For this PR (remove rubber bands when no editable layer is left in the QGIS project), that would mean I would have to keep a list of layers that once were added to canvas (layer registry?), and once the user stops the edits on a layer, go to that "layer registry" (not to the canvas) to ask if any layer there is still being edited. 

If we don't use such "layer registry", we would miss invisible layers' `editingStopped` signals, so still rubber bands could be on the map canvas even if no layers are being edited in the project.

As you see, using `QgsProject::layersAdded` is very convenient here.

uclaros:
How about checking whether `mActionAllEdits` is disabled when a `QgsMapToolEdit` is deactivated. Then do the cleanup, else skip it!

src/gui/qgsmaptooledit.cpp
nyalldawson: @m-kuhn looks like this PR stalled on this question. Can you reply please?

m-kuhn: Thanks for the heads up @nyalldawson 

Fair point, @gacarrillor , good explanation why the map canvas layer set is not sufficient here. 

Do we have any chance for a setter for the project, so we don't have to use the global singleton? The most straightforward thing would be to add `QgsMapCanvas::mProject`.

gacarrillor:
> How about checking whether `mActionAllEdits` is disabled when a `QgsMapToolEdit` is deactivated. Then do the cleanup, else skip it!

Thanks for the suggestion, but then we would be relying again on QgisApp, which is not desirable here as map tools are in the public API.

src/gui/qgsmaptooledit.cpp
gacarrillor: @m-kuhn do you mean to use a public method `project()` from `QgsMapCanvas`? If so, are you OK if we set the main canvas' project [in this block](https://github.com/qgis/QGIS/blob/master/src/app/qgisapp.cpp#L922-L943)?

m-kuhn: Thant would make perfect sense for me.
@nyalldawson are you supportive of this approach?

nyalldawson: Yes, but for consistency make sure you also set the project for additional map canvases in QgisApp::createNewMapCanvasDock()

nyalldawson (commented):
Looking really good! Nearly there.

src/gui/qgsmapcanvas.cpp
```diff
@@ -2030,6 +2030,14 @@ void QgsMapCanvas::unsetMapTool( QgsMapTool *tool )
   }
 }
 
+void QgsMapCanvas::setProject( QgsProject *project )
+{
+  if ( !project )
```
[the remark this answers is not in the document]
gacarrillor: Yes, I had doubts with that one. Fixed in the new commits.

src/gui/qgsmapcanvas.h
```diff
@@ -355,6 +356,20 @@ class GUI_EXPORT QgsMapCanvas : public QGraphicsView
...
 
+    /**
+     * Sets the project linked to this canvas.
+     *
+     * \since QGIS 3.14
+     */
+    void setProject( QgsProject *project );
+
+    /**
+     * Gets access to the project linked to this canvas.
```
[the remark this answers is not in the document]
gacarrillor: Adjusted!

src/gui/qgsmapcanvas.h
```diff
@@ -355,6 +356,20 @@ class GUI_EXPORT QgsMapCanvas : public QGraphicsView
     //! Returns the currently active tool
     QgsMapTool *mapTool();
 
+    /**
+     * Sets the project linked to this canvas.
```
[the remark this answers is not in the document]
gacarrillor: Done.

src/gui/qgsmaptooledit.cpp
```diff
@@ -171,3 +176,36 @@ void QgsMapToolEdit::notifyNotEditableLayer()
...
+  if ( editableVectorLayers().isEmpty() )
+  {
+    clean();
+  }
+}
+
+QList<QgsVectorLayer *> QgsMapToolEdit::editableVectorLayers()
+{
+  QList<QgsVectorLayer *> editableLayers;
+  const auto layers = mCanvas->project()->mapLayers().values();
```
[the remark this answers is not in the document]
gacarrillor: For sure!

nyalldawson:
Looks good to me! Let's let @m-kuhn have the final say here though...

src/gui/qgsmaptooledit.cpp
```diff
@@ -27,6 +27,11 @@
 QgsMapToolEdit::QgsMapToolEdit( QgsMapCanvas *canvas )
   : QgsMapTool( canvas )
 {
+  if ( mCanvas->project() )
+  {
+    connect( mCanvas->project(), &QgsProject::layersAdded, this, &QgsMapToolEdit::connectLayers );
+    connectLayers( mCanvas->project()->mapLayers().values() );  // Connect existing layers
```
m-kuhn, line 33: Should this also be executed when new layers get added to the project?

m-kuhn:
Very good !
Once that final doubt is sorted out I'm good with merging. Since it's a rather big one in a central piece of QGIS let's play it safe and merge it directly after release of 3.14.0 to have a full month of testing.

src/gui/qgsmaptooledit.cpp
gacarrillor: Sure, in that couple of lines we cover both scenarios: first, we keep listening for new vector layers added to the project, so that as soon as they get added we connect their `editingStopped` SIGNAL; and second, if we create a new map tool edit and the project already has vector layers, we want to make sure these existing layers have their `editingStopped` SIGNALs connected.

m-kuhn: :see_no_evil:

gacarrillor:
I'm OK with that plan. 

Thank you Matthias and Nyall for your suggestions!

PR #36302, opened 2020-05-09
Set DPI of rendered images

m-kuhn:
The DPI of rendered images has been hard set to 100 so far, regardless of MapSettings.

E.g. the following code

```
img = renderer.renderedImage()
img.save(buf, 'PNG')
```

would save the PNG with a DPI of 100.


m-kuhn:
@rldhont should we backport that too? https://github.com/qgis/QGIS/pull/37036 does not work without it

rldhont:
@m-kuhn yes you can

PR #36304, opened 2020-05-09
[FEATURE][layout] Expose temporal UI for map items

nirvn:
## Description

There was no way we'd ship _QGIS 3.14 "Temporal edition"_ without temporal support to layouts:
Peek 2020-05-09 14-45

This PR adds UI to the layout map item widget to enable/disable temporal range, as well as defining the temporal range start and end.

The start and end datetime values can be data-defined. I.e., this PR unlocks temporal atlas export. :tada:

### DIFF
--- a/src/core/qgspropertycollection.h
+++ b/src/core/qgspropertycollection.h
@@ -120,19 +120,20 @@ class CORE_EXPORT QgsAbstractPropertyCollection
     virtual QVariant value( int key, const QgsExpressionContext &context, const QVariant &defaultValue = QVariant() ) const = 0;
 
     /**
-     * Calculates the current value of the property with the specified key and interprets it as a string.
+     * Calculates the current value of the property with the specified key and interprets it as a datetime.
      * \param key integer key for property to return. The intended use case is that a context specific enum is cast to
      * int and used for the key value.
      * \param context QgsExpressionContext to evaluate the property for.
-     * \param defaultString default string to return if the property cannot be calculated as a string
+     * \param defaultDateTime default datetime to return if the property cannot be calculated as a datetime
      * \param ok if specified, will be set to TRUE if conversion was successful
-     * \returns value parsed to string
+     * \returns value parsed to datetime
      * \see value()
      * \see valueAsString()
      * \see valueAsColor()
      * \see valueAsDouble()
      * \see valueAsInt()
      * \see valueAsBool()
+     * \since QGIS 3.14
      */
     QDateTime valueAsDateTime( int key, const QgsExpressionContext &context, const QDateTime &defaultDateTime = QDateTime(), bool *ok SIP_OUT = nullptr ) const;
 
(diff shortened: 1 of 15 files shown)


nyalldawson (commented):
Can you also add a unit test for data defined temporal range? (I think there's existing tests which check the map settings with other DD properties which can be updated, no need for a render based test here)

Otherwise, GREAT feature!! This also opens the possibility of "small multiples" style layouts showing a bunch of maps showing different time ranges, which is really cool.

saberraz:
What do you think of having time support in Map Themes? User can then set up themes and just reference to the themes in the layout?

saberraz:
One other idea: extending atlas to also cover temporal in addition to spatial extend? So, you can then export maps of the same (or different extents) for a time range?

nirvn:
@saberraz , layouts would definitely benefit from a temporal export panel. But that's beyond the scope of this PR and my available free time.

As for atlases, you can already output temporal atlas by using the data defined temporal start n end properties. For eg, you could have a csv file with a start n end pair of fields and a number of rows representing date ranges. Hit atlas export, et voila :)

-1 to map theme getting temporal range, but that might just be me not getting your vision.

src/core/qgspropertycollection.h
```diff
@@ -117,6 +119,24 @@ class CORE_EXPORT QgsAbstractPropertyCollection
      */
     virtual QVariant value( int key, const QgsExpressionContext &context, const QVariant &defaultValue = QVariant() ) const = 0;
 
+    /**
+     * Calculates the current value of the property with the specified key and interprets it as a datestring.
```
nyalldawson, line 123:
```suggestion
     * Calculates the current value of the property with the specified key and interprets it as a datetime.
```

src/core/qgspropertycollection.h
```diff
@@ -117,6 +119,24 @@ class CORE_EXPORT QgsAbstractPropertyCollection
...
+     * \param key integer key for property to return. The intended use case is that a context specific enum is cast to
+     * int and used for the key value.
+     * \param context QgsExpressionContext to evaluate the property for.
+     * \param defaultDateTime default string to return if the property cannot be calculated as a datestring
```
nyalldawson, line 127:
```suggestion
     * \param defaultDateTime default datetime to return if the property cannot be calculated as a datetime
```

src/core/qgspropertycollection.h
```diff
@@ -117,6 +119,24 @@ class CORE_EXPORT QgsAbstractPropertyCollection
...
+     * \param ok if specified, will be set to TRUE if conversion was successful
+     * \returns value parsed to string
```
nyalldawson, line 129:
```suggestion
     * \returns value parsed to datetime
```

nyalldawson:
> -1 to map theme getting temporal range, but that might just be me not getting your vision.

I agree - I think it's jamming two different concepts together. I think rather we'd want to extend the spatial bookmarks concepts to add temporal bookmarks (or spatio-temporal bookmarks!)

nyalldawson (approved):

PR #36308, opened 2020-05-09
Fix building with WITH_ANALYSIS=FALSE

### LINKED ISSUE #36306: Compile fails with WITH_ANALYSIS=FALSE
Building qgis with cmake options:
cmake .. -DWITH_ANALYSIS=FALSE -DWITH_3D=FALSE -DWITH_DESKTOP=FALSE -DWITH_SERVER=TRUE -DWITH_GUI=FALSE -DWITH_QGIS_PROCESS=FALSE

fails with error message:
In file included from /home/build/QGIS/src/test/qgstest.h:20,
                 from /home/build/QGIS/tests/src/geometry_checker/testqgsvectorlayerfeaturepool.cpp:16:
/home/build/QGIS/src/core/geometry/qgsrectangle.h:24:10: fatal error: QDomDocument: No such file or directory
 #include <QDomDocument>
          ^~~~~~~~~~~~~~

tpalan:
Disable building of qgis_process with the analysis module, and disable analysis tests & python modules if not enabled.
Fixes #36306


nyalldawson:
We could **probably** still allow the qgis_process tool to be built without analysis... but it's an unlikely situation

PR #36309, opened 2020-05-09
Add cumulative temporal range setting in temporal controller

Samweli:
Adds a setting in temporal controller to set the animation temporal range as cumulative, that is all animation frames will have the same start date time but different end date time when the temporal range is cumulative.

eg.
cumulative_animation_range

### DIFF
--- a/src/core/qgsprojecttimesettings.cpp
+++ b/src/core/qgsprojecttimesettings.cpp
@@ -63,7 +63,7 @@ bool QgsProjectTimeSettings::readXml( const QDomElement &element, const QgsReadW
   mTimeStepUnit = QgsUnitTypes::decodeTemporalUnit( element.attribute( QStringLiteral( "timeStepUnit" ), QgsUnitTypes::encodeUnit( QgsUnitTypes::TemporalHours ) ) );
   mTimeStep = element.attribute( QStringLiteral( "timeStep" ), "1" ).toDouble();
   mFrameRate = element.attribute( QStringLiteral( "frameRate" ), "1" ).toDouble();
-  mCumulativeTemporalRange = element.attribute( QStringLiteral( "cumulativeTemporalRange" ), "0" ).toInt() == 1 ? true : false;
+  mCumulativeTemporalRange = element.attribute( QStringLiteral( "cumulativeTemporalRange" ), "0" ).toInt();
 
   return true;
 }
--- a/src/core/qgsprojecttimesettings.cpp
+++ b/src/core/qgsprojecttimesettings.cpp
@@ -63,6 +63,7 @@ bool QgsProjectTimeSettings::readXml( const QDomElement &element, const QgsReadW
   mTimeStepUnit = QgsUnitTypes::decodeTemporalUnit( element.attribute( QStringLiteral( "timeStepUnit" ), QgsUnitTypes::encodeUnit( QgsUnitTypes::TemporalHours ) ) );
   mTimeStep = element.attribute( QStringLiteral( "timeStep" ), "1" ).toDouble();
   mFrameRate = element.attribute( QStringLiteral( "frameRate" ), "1" ).toDouble();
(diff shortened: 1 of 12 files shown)


src/core/qgsprojecttimesettings.cpp
```diff
@@ -63,6 +63,7 @@ bool QgsProjectTimeSettings::readXml( const QDomElement &element, const QgsReadW
   mTimeStepUnit = QgsUnitTypes::decodeTemporalUnit( element.attribute( QStringLiteral( "timeStepUnit" ), QgsUnitTypes::encodeUnit( QgsUnitTypes::TemporalHours ) ) );
   mTimeStep = element.attribute( QStringLiteral( "timeStep" ), "1" ).toDouble();
   mFrameRate = element.attribute( QStringLiteral( "frameRate" ), "1" ).toDouble();
+  mCumulativeTemporalRange = element.attribute( QStringLiteral( "cumulativeTemporalRange" ), "0" ).toInt() == 1 ? true : false;
```
nyalldawson, line 66:
```suggestion
  mCumulativeTemporalRange = element.attribute( QStringLiteral( "cumulativeTemporalRange" ), "0" ).toInt();
```

src/core/qgstemporalnavigationobject.cpp
```diff
@@ -90,6 +90,9 @@ QgsDateTimeRange QgsTemporalNavigationObject::dateTimeRangeForFrameNumber( long
   const QDateTime begin = start.addSecs( frame * mFrameDuration.seconds() );
   const QDateTime end = start.addSecs( nextFrame * mFrameDuration.seconds() );
 
+  if ( mCumulativeTemporalRange )
+    return QgsDateTimeRange( start, begin, true, false );
```
nyalldawson, line 94:
```suggestion
    return QgsDateTimeRange( start, end, true, false );
```

nyalldawson (commented):
Nice feature!

As well as this global setting, I think there's an argument to be made for us to also expose this option on a layer by layer basis. I can see use cases where one layer should be cumulative, but others sliding range-based. (I'll do this for vector shortly)

src/core/qgstemporalnavigationobject.cpp
Samweli: This will cause the last frame date time range to go out of the animation temporal extent bounds.

Samweli:
> As well as this global setting, I think there's an argument to be made for us to also expose this option on a layer by layer basis. I can see use cases where one layer should be cumulative, but others sliding range-based

Makes sense +1

src/core/qgstemporalnavigationobject.cpp
nyalldawson: Could you apply the end time clamping first before doing this check then?

Samweli: Sure @nyalldawson, I'm thinking of updating the logic to look like this

```c++
const QDateTime begin = start.addSecs( frame * mFrameDuration.seconds() );
const QDateTime end = start.addSecs( nextFrame * mFrameDuration.seconds() );
  
QDateTime frameStart = begin;
 if ( mCumulativeTemporalRange )
     frameStart = start;

 if ( end <= mTemporalExtents.end() )
      return QgsDateTimeRange( frameStart, end, true, false );

return QgsDateTimeRange( frameStart, mTemporalExtents.end(), true, false );
```

are you okay with this?

timlinux:
@Samweli awesome - been missing this feature, glad you got it lined up in time for feature freeze!

src/core/qgstemporalnavigationobject.cpp
nyalldawson: Yes, looks good!

Samweli: Okay

Samweli:
> @Samweli awesome - been missing this feature, glad you got it lined up in time for feature freeze!

Thanks @timlinux!

nyalldawson (approved):

PR #36313, opened 2020-05-09
[MESH] scalar color settings depending on classification

vcloarec:
In this PR, MDAL reads classified scalar values in dataset. When values are classified, the default scalar rendering settings has color ramp shader and labels according to the classified values.

image


src/app/mesh/qgsmeshrendereractivedatasetwidget.cpp
```diff
@@ -185,6 +185,11 @@ QString QgsMeshRendererActiveDatasetWidget::metadata( QgsMeshDatasetIndex datase
   const auto options = gmeta.extraOptions();
   for ( auto it = options.constBegin(); it != options.constEnd(); ++it )
   {
+    if ( it.key() == QStringLiteral( "classification" ) )
+    {
+      msg += QStringLiteral( "<tr><td>Classified values</td></tr>" );
```
nyalldawson, line 190: Translate

nyalldawson:
I'm not so familiar with the dataset capabilities - to clarify, do some mesh formats support embedded classification metadata?

vcloarec:
@nyalldawson some formats can provide values that can be compressed by categorizing values in consecutive intervals, each represent by an integer. Now MDAL has the capabilities to recognize this dataset type and store the bounds of each class an the units in the metadata.
QGIS uses this metadata to setup adapted color ramp shader.

nyalldawson:
Thanks for the clarification @vcloarec !

QGIS code looks good to me, can't comment on the mdal side :)

PeterPetrik:
MDAL already reviewed in the MDAL. repository. all looks good

PR #36327, opened 2020-05-10
geom_to_wkb - update description

havatv:
## Description

Improve the description, as done in the documentation.

### DIFF
--- a/resources/function_help/json/geom_to_wkb
+++ b/resources/function_help/json/geom_to_wkb
@@ -1,7 +1,7 @@
 {
   "name": "geom_to_wkb",
   "type": "function",
-  "description": "Returns the Well-Known Binary (WKB) representation of a geometry as a binary value (BLOB)",
+  "description": "Returns the Well-Known Binary (WKB) representation of a geometry (a binary value)",
   "arguments": [ {"arg":"geometry","description":"a geometry"}],
   "examples": [ { "expression":"geom_to_wkb( $geometry )", "returns":"binary blob containing a geometry object"}
   ]
--- a/resources/function_help/json/geom_to_wkb
+++ b/resources/function_help/json/geom_to_wkb
@@ -1,7 +1,7 @@
 {
   "name": "geom_to_wkb",
   "type": "function",
-  "description": "Returns the Well-Known Binary (WKB) representation of a geometry (a binary value)",
+  "description": "Returns the Well-Known Binary (WKB) representation of a geometry (as a binary value)",
   "arguments": [ {"arg":"geometry","description":"a geometry"}],
   "examples": [ { "expression":"geom_to_wkb( $geometry )", "returns":"binary blob containing a geometry object"}
   ]
--- a/resources/function_help/json/geom_to_wkb
+++ b/resources/function_help/json/geom_to_wkb
@@ -1,7 +1,7 @@
 {
   "name": "geom_to_wkb",
   "type": "function",
-  "description": "Returns the Well-Known Binary (WKB) representation (a BLOB) of a geometry.",
+  "description": "Returns the Well-Known Binary (WKB) representation of a geometry as a binary value (BLOB)",
   "arguments": [ {"arg":"geometry","description":"a geometry"}],
   "examples": [ { "expression":"geom_to_wkb( $geometry )", "returns":"binary blob containing a geometry object"}
   ]
--- a/resources/function_help/json/geom_to_wkb
+++ b/resources/function_help/json/geom_to_wkb
@@ -1,7 +1,7 @@
 {
   "name": "geom_to_wkb",
   "type": "function",
-  "description": "Returns the Well-Known Binary (WKB) representation of a geometry as a binary blob.",
+  "description": "Returns the Well-Known Binary (WKB) representation (a BLOB) of a geometry.",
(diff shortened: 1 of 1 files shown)


nyalldawson:
I'm -0, the original version sounds better worded to me

havatv:
> I'm -0, the original version sounds better worded to me

We must at least avoid "binary blob" (BLOB stands for Binary Large OBject -> "binary binary large object"). That was our main reason for changing.

nyalldawson:
What about "... as a binary value (BLOB)"?

resources/function_help/json/geom_to_wkb
```diff
@@ -1,7 +1,7 @@
 {
   "name": "geom_to_wkb",
   "type": "function",
-  "description": "Returns the Well-Known Binary (WKB) representation of a geometry as a binary blob.",
+  "description": "Returns the Well-Known Binary (WKB) representation (a BLOB) of a geometry.",
```
havatv, line 4:
```suggestion
  "description": "Returns the Well-Known Binary (WKB) representation of a geometry as a binary value (BLOB)",
```

havatv:
> What about "... as a binary value (BLOB)"?

I still think it is obvious that a well known ***binary*** representation is returned as a ***binary*** value, but this is a significant improvement.

m-kuhn:
Can we drop the "blob" from it?
To me this sounds like database speak ([wikipedia seems to back this up](https://en.wikipedia.org/wiki/Blob)) and adds nothing of value in this context.

resources/function_help/json/geom_to_wkb
```diff
@@ -1,7 +1,7 @@
 {
   "name": "geom_to_wkb",
   "type": "function",
-  "description": "Returns the Well-Known Binary (WKB) representation of a geometry as a binary blob.",
+  "description": "Returns the Well-Known Binary (WKB) representation of a geometry as a binary value (BLOB)",
```
havatv, line 4:
```suggestion
  "description": "Returns the Well-Known Binary (WKB) representation of a geometry (a binary value)",
```

havatv:
> Can we drop the "blob" from it?
> To me this sounds like database speak ([wikipedia seems to back this up](https://en.wikipedia.org/wiki/Blob)) and adds nothing of value in this context.

I have tried to simplify. Still keeping the word "binary" in the explanation, but now in parenthesis.

resources/function_help/json/geom_to_wkb
```diff
@@ -1,7 +1,7 @@
 {
   "name": "geom_to_wkb",
   "type": "function",
-  "description": "Returns the Well-Known Binary (WKB) representation of a geometry as a binary blob.",
+  "description": "Returns the Well-Known Binary (WKB) representation of a geometry (a binary value)",
```
m-kuhn, line 4:
```suggestion
  "description": "Returns the Well-Known Binary (WKB) representation of a geometry as a binary value",
```

I would do it this way.
I think it's an important detail (not in parenthesis), it clarifies that it's in binary format (and not as a string, as e.g. pgadmin shows it in the web interface).

havatv: But is it not evident that a Well-Known Binary representation is a binary value? Are we not insulting the readers by repeating it?

m-kuhn: I get where you are coming from.
What it can tell you is that QGIS passes it around as a binary value internally (and not hexified or base64 or whatever encoded string as e.g. pgadmin does when displaying it).

havatv, line 4:
```suggestion
  "description": "Returns the Well-Known Binary (WKB) representation of a geometry (as a binary value)",
```

havatv: If we place it within parentheses we should avoid insulting the knowledgeable reader. ;-)
I have updated - including the word "as".

m-kuhn: I'm -0 on the parenthesis, IMHO they actually *do* suggest that it's repeated information (and hence are insulting the reader).
My take on this is, the first part (WKB) refers to *content*, the second part (binary value) refers to the *data format* internally used in qgis and the expression engine.
Compare to `3` vs. `'3'` vs. `3.0 -> the *content* is the same, but the *data format* is different.
If this is stating the obvious and repeating information, we should rather drop it completely.

havatv: > If this is stating the obvious and repeating information, we should rather drop it completely.

Dropping it entirely is fine.
If the representation had been something else than binary, I agree that it would be useful to specify the representation, but when it is binary - as one would expect, it should not be necessary to mention it.

resources/function_help/json/geom_to_wkb
```diff
@@ -1,7 +1,7 @@
 {
   "name": "geom_to_wkb",
   "type": "function",
-  "description": "Returns the Well-Known Binary (WKB) representation of a geometry as a binary blob.",
+  "description": "Returns the Well-Known Binary (WKB) representation of a geometry (as a binary value)",
```
havatv, line 4:
```suggestion
  "description": "Returns the Well-Known Binary (WKB) representation of a geometry",
```

havatv:
Thanks, @nyalldawson.

PR #36328, opened 2020-05-10
Support TMS convention with vector tiles ({-y} instead of {y})

wonder-sk:
XYZ convention has zero Y at the top and increasing downwards, but various map servers use TMS convention which has zero Y at the bottom and rising upwards (within tile matrix)


timlinux:
Thanks for this! I will give it a try....

PR #36331, opened 2020-05-11
Fix (?) some more expression functions which are sensitive to time zones

nyalldawson:
Because we don't expose timezone information anywhere in QGIS, we
have to get all datetime values treated as coming from the same
time zone.

This is going to be a rabbit hole... eventually we'll need to
fully support time zone handling in field values

Refs https://github.com/qgis/QGIS/pull/36243#issuecomment-626329313


nyalldawson:
A deeper rethought is needed. I think we'll have no alternative but exposing time zones to users. 😱

PR #36335, opened 2020-05-11
[expression] Compile make_{datetime,date,time} for sqlite

nirvn:
## Description

This PR implements expression compilation of make_{datetime,date,time} for sqlite-based datasets (geopackage and spatialite). This unlocks blazingly fast temporal navigation over large file-based datasets.

Before:
Peek 2020-05-11 09-53

After:
Peek 2020-05-11 09-52

### DIFF
--- a/src/core/qgssqliteexpressioncompiler.cpp
+++ b/src/core/qgssqliteexpressioncompiler.cpp
@@ -17,7 +17,9 @@
 
 #include "qgssqliteexpressioncompiler.h"
 #include "qgssqlexpressioncompiler.h"
+#include "qgsexpressionfunction.h"
 #include "qgsexpressionnodeimpl.h"
+#include "qgsexpression.h"
 #include "qgssqliteutils.h"
 
 QgsSQLiteExpressionCompiler::QgsSQLiteExpressionCompiler( const QgsFields &fields )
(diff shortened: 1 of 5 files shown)


src/core/qgssqliteexpressioncompiler.cpp
```diff
@@ -90,11 +90,41 @@ QString QgsSQLiteExpressionCompiler::sqlFunctionFromFunctionName( const QString
...
   return FN_NAMES.value( fnName, QString() );
 }
 
+QStringList QgsSQLiteExpressionCompiler::sqlArgumentsFromFunctionName( const QString &fnName, const QStringList &fnArgs ) const
+{
+  QStringList args( fnArgs );
+  if ( fnName == QLatin1String( "make_datetime" ) )
+  {
+    args = QStringList( QStringLiteral( "DATE('%1-%2-%3 %4:%5:%6')" ).arg( args[0].rightJustified( 4, '0' ) )
+                        .arg( args[1].rightJustified( 2, '0' ) )
```
nyalldawson, line 107: There's an issue here -- make_date etc don't necessarily have numeric arguments. You could potentially call them with sub expressions or field references

nirvn: And the expression engine doesn't resolve those nodes first?

nyalldawson: Our expression engine does, but the compiler runs before all this.

I think what we need to do is check if the node values used as arguments are constant, and if they are then it's safe to use your approach. If not the compilation should fail.

nirvn: And how is that done sir? :)

nirvn:
@m-kuhn , flake8 py3 test failing here, something wrong with the setup?

nyalldawson:
I've a fix in the HTML labels branch (which is just waiting on a green status and i'll merge)

nyalldawson:
@rouault have you got any insights into how sqlite and gdal handle time zones for datetime fields? Does gdal read these? Does it write them?

PR #36336, opened 2020-05-11
[vector layer] Fix minimumValue/maximumValue for joined and virtual fields

nirvn:
## Description

On top of this PR Doing The Right Thing (TM), it fixes use of virtual fields for temporal matters.


nyalldawson (commented):
Nice fix! And happy to hear this fixes temporal layers with virtual fields, that's a good use case to support

nyalldawson (approved):

PR #36351, opened 2020-05-11
Fix QgsGeometry nearestPoint and minimalEnclosingCircle when the geometry is empty

### LINKED ISSUE #36142: QGIS crashes with QgsGeometry method on empty collection
**Describe the bug**
QGIS crashes when calling `QgsGeometry::nearestPoint()` on an empty collection `QgsGeometry`

**How to Reproduce**
On a python console run the following:
```
g=QgsGeometry.fromWkt('GeometryCollection EMPTY')
g.nearestPoint(QgsGeometry.fromWkt('Point (42  0)'))
```

**QGIS and OS versions**
81911c1afc on Debian testing (geos 3.8.1) , 3.10.2 on windows (geos 3.8.0)

**Additional context**
the empty geometry collection also gives the following strange result:
*Nanananananananan, Batman!*
```
>>> QgsGeometry.fromWkt('GeometryCollection EMPTY').minimalEnclosingCircle()

(<QgsGeometry: Polygon ((nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan, nan nan))>, <QgsPointXY: POINT EMPTY>, 0.0)
```

obrix:
* In QgsGeometry nearestPoint method : exit early if the geometry is empty to avoid creating an invalid GEOSNearestPoints_r.

* In QgsGeometry minimalEnclosingCircle method : return a default QgsGeometry instead of a default 36 segments polygon with NaN value if the geometry is empty.

fix #36142

### DIFF
--- a/src/core/geometry/qgsgeometry.cpp
+++ b/src/core/geometry/qgsgeometry.cpp
@@ -606,6 +606,10 @@ double QgsGeometry::sqrDistToVertexAt( QgsPointXY &point, int atVertex ) const
 
 QgsGeometry QgsGeometry::nearestPoint( const QgsGeometry &other ) const
 {
+  if ( isEmpty() )
+  {
+    return QgsGeometry();
+  }
   QgsGeos geos( d->geometry.get() );
   mLastError.clear();
   QgsGeometry result = geos.closestPoint( other );
@@ -1072,7 +1076,7 @@ QgsGeometry QgsGeometry::minimalEnclosingCircle( QgsPointXY &center, double &rad
   center = QgsPointXY();
   radius = 0;
 
-  if ( !d->geometry )
+  if ( isEmpty() )
   {
     return QgsGeometry();
   }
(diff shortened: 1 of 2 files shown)


src/core/geometry/qgsgeometry.cpp
```diff
@@ -606,6 +606,10 @@ double QgsGeometry::sqrDistToVertexAt( QgsPointXY &point, int atVertex ) const
 
 QgsGeometry QgsGeometry::nearestPoint( const QgsGeometry &other ) const
 {
+  if ( isEmpty() )
```
uclaros, line 609: Wouldn't it be better if this was handled in `QgsGeos::closestPoint` instead?

obrix: Yes I agree. Done.

Issue #34317, opened 2020-02-06
`RuntimeError: qgis._gui cannot import type '����' from PyQt5.QtCore` with QGIS 3.10.2 and Ubuntu 18.04


stragu:
Using Ubuntu Studio 18.04 with the ubuntugis repos, when I start QGIS 3.10.2, I run into this issue:
	
	Couldn't load PyQGIS.
	Python support will be disabled.

	Traceback (most recent call last):
	  File "", line 1, in 
	  File "/usr/lib/python3/dist-packages/qgis/gui/__init__.py", line 25, in 
	    from qgis._gui import *
	RuntimeError: qgis._gui cannot import type '����' from PyQt5.QtCore

	Python version:
	3.6.9 (default, Nov  7 2019, 10:44:02) 
	[GCC 8.3.0]

	QGIS version:
	3.10.2-A Coruña 'A Coruña', 616ad4531b

	Python path:
	['/usr/share/qgis/python', '/home/stragu/.local/share/QGIS/QGIS3/profiles/default/python', '/home/stragu/.local/share/QGIS/QGIS3/profiles/default/python/plugins', '/usr/share/qgis/python/plugins', '/usr/lib/python36.zip', '/usr/lib/python3.6', '/usr/lib/python3.6/lib-dynload', '/home/stragu/.local/lib/python3.6/site-packages', '/home/stragu/software/reaper/src/socialreaper', '/usr/local/lib/python3.6/dist-packages', '/usr/lib/python3/dist-packages', '/usr/lib/python3.6/dist-packages']

Which results in QGIS without most of the processing tools.

I could find my Qt5 and PyQt5 version by running the following:

	python3
	from PyQt5.QtCore import QT_VERSION_STR
	print("Qt version:", QT_VERSION_STR)
	from PyQt5.Qt import PYQT_VERSION_STR
	print("PyQt version:", PYQT_VERSION_STR)

I had 5.9.2 for the Qt version, and 5.9.1 for the PyQt version.

I also checked my version of `libqt5core5a`, and I had 5.9.5 installed.

So I thought I could try updating PyQt5 incrementally and see if that helped. You can check the history of releases of PyQt5 here: https://pypi.org/project/PyQt5/#history

I first tried:

	pip3 install 'pyqt5==5.10.1'

[...]

PR #36364, opened 2020-05-12
[Oracle] Force multi geometry type at layer level

rldhont:
Oracle DBMS can mix single and multi geometry type like ESRI ShapeFile.
So the geometry type has to be multi geometry type for requested one.

### DIFF
--- a/src/providers/oracle/qgsoracleprovider.cpp
+++ b/src/providers/oracle/qgsoracleprovider.cpp
@@ -78,6 +78,10 @@ QgsOracleProvider::QgsOracleProvider( QString const &uri, const ProviderOptions 
   mSqlWhereClause = mUri.sql();
   mSrid = mUri.srid().toInt();
   mRequestedGeomType = mUri.wkbType();
+  if ( QgsWkbTypes::isSingleType( mRequestedGeomType ) && QgsWkbTypes::geometryType( mRequestedGeomType ) != QgsWkbTypes::PointGeometry )
+  {
+    mRequestedGeomType = QgsWkbTypes::multiType( mRequestedGeomType );
+  }
   mUseEstimatedMetadata = mUri.useEstimatedMetadata();
   mIncludeGeoAttributes = mUri.hasParam( "includegeoattributes" ) ? mUri.param( "includegeoattributes" ) == "true" : false;
 


Jean-Roc:
@SebastienPeillet did a PR [#34358](https://github.com/qgis/QGIS/pull/34358), it filtered single and multi geometries but it was not merged

rldhont:
Hi @Jean-Roc, the idea is not to force the geometries to multi but only to change to multi the layer geometry type.

Like ESRI Shapefile, Orcale database can mix simple and multi geometry. QGIS is able to use mixed simple and multi. Instead of separating simple and multi, this commit only change the geometry type stored in the datasource URI to indicate that the layer can store multi and not just simple geometry.

nyalldawson:
> QGIS is able to use mixed simple and multi

Actually this causes a lot of complex issues. Shapefiles silently upgrade all feature to multipart to avoid these.

rldhont:
> > QGIS is able to use mixed simple and multi
> 
> Actually this causes a lot of complex issues. Shapefiles silently upgrade all feature to multipart to avoid these.

The idea here is quite the same

rldhont:
@jef-n have you any objection against my proposition ?

uclaros:
@rldhont I'm afraid this was rushed to merge, there are issues with this approach.
Exporting a polygon oracle layer to a geopackage will automatically create a multipolygon layer and all geometries will be converted to multipolygon too, as @nyalldawson said is the case with Shapefiles.

nyalldawson:
@uclaros I'll revert -- something this critical should not be merged if there dissenting views

mtreguer:
But without this correction, something is wrong in WFS, and somewhere else perhaps. 
For example 
A WFS DescribeFeatureType , could describe the geometry as gml:LineStringPropertyType :
`<element maxOccurs="1" name="geometry" type="gml:LineStringPropertyType" minOccurs="0"/>`
And the WFS GetFeature return MultiLineString feature :
`<MultiLineString xmlns="http://www.opengis.net/gml" srsName="EPSG:4326">`

It's not WFS compliant.

nyalldawson:
That may be the case, but it's a substantial change in qgis which can't be merged before being fully reviewed

uclaros:
There is totally something wrong, but I'm afraid this correction is not the correct correction...
Oracle thinks it's correct to have *any* geometry type on the same table, not just single - multi types of the same dimensions, but that doesn't mean I should!
When I load a point layer, I expect it to have points, not multilines. Same goes when I load a line layer; I expect lines, not multilines.
I think PR #34358 was on a good track; treat single and multi as a different geometry type, like points and lines and polygons, and we should clean our data and decide whether we want a single or multi type table.
When one defines a single type layer, he explicitly states "I don't want multiparts", but when defining a multi type then probably single parts are also ok. So maybe also add an option to load layer as multi, but also promote all single to multi. 
That way you will always get what you expect.
Hope that makes sense, it's late here

SebastienPeillet:
Hi,

> something this critical should not be merged if there dissenting views

I agree, this is why I had trouble to finish PR #34317 . I think that what I proposed was a viable solution, but not very popular with Oracle user that are used to have mixed geometry tables.

To my point of view, I agree with @uclaros about the fact that even if Oracle allows something, it does not mean QGIS has to do it. Mixing different geometry types and simple et multi geometries... what a mess.

> So maybe also add an option to load layer as multi, but also promote all single to multi

Indeed it could be a solution. A real problem is that if a table is created without any data, QGIS can't determined the geometry type and if it's a single or multi geometry layer. And if, at a moment there is only single geometry feature (in Oracle SDO definition) in the table, QGIS  will propose to load single geometry layer, but it's not necessarily what expected the user who wanted to have a multi geometry layer (but can't define it in Oracle). So I think QGIS should also propose to load a multi geometry layer even if there is no multi geometry feature in the table (and doing so we convert all single geometry features to multi geometry features composed of a single part).

About PR #34317 , I can modify it to fit with what is exposed above, if we agree on this :)

PR #36366, opened 2020-05-12
Save correctly selective masking parameters

### LINKED ISSUE #33277: Symbol masking for labels seems quite broken
Possibly I'm totally misusing this feature, but I can't get it to actually DO anything!

The settings are lost just by closing dialogs, and regardless of which combination of checkboxes I tick there's no visible impact on the map...

Peek 2019-12-07 19-59

troopa81:
## Description

This PR fixes #33277. It fixes several things:

- It applies masking widget settings from the vector properties

- The edit mask settings were not serialized in the project and so need to be unchecked every time we display the widget. This checkbox was used to defer widget populate which takes time #35557. I propose to delay the populate when the widget is first shown. 

- If user selects a masked symbol layer but no mask source (or vice versa) the configuration is not saved because data model associates sources with layers. I propose to display a warning message when the configuration is not consistent.

msgSelectiveMasking

### DIFF
--- a/src/gui/qgsmaskingwidget.cpp
+++ b/src/gui/qgsmaskingwidget.cpp
@@ -28,29 +28,54 @@
 #include "qgsvectorlayerutils.h"
 #include "qgsmasksymbollayer.h"
 #include "qgsvectorlayerlabeling.h"
+#include "qgsmessagebaritem.h"
 
 QgsMaskingWidget::QgsMaskingWidget( QWidget *parent ) :
   QgsPanelWidget( parent )
 {
   setupUi( this );
+}
 
-  connect( mMaskTargetsWidget, &QgsSymbolLayerSelectionWidget::changed, this, [&]()
-  {
-    mMaskSourcesWidget->setEnabled( ! mMaskTargetsWidget->selection().isEmpty() );
-    emit widgetChanged();
-  } );
-  connect( mMaskSourcesWidget, &QgsMaskSourceSelectionWidget::changed, this, [&]()
-  {
-    emit widgetChanged();
-  } );
+void QgsMaskingWidget::onSelectionChanged()
+{
+  // display message if configuration is not consistent
+  bool printMessage =
+    ( mMaskTargetsWidget->selection().empty() && !mMaskSourcesWidget->selection().empty() )
+    || ( !mMaskTargetsWidget->selection().empty() && mMaskSourcesWidget->selection().empty() );
 
-  connect( mEditMaskSettingsGroup, &QGroupBox::toggled, this, [&]( bool on )
+  if ( mMessageBarItem && !printMessage )
   {
-    if ( on && mLayer )
-    {
-      populate();
-    }
-  } );
+    mMessageBar->popWidget( mMessageBarItem );
+    delete mMessageBarItem;
+    mMessageBarItem = nullptr;
+  }
+  else if ( !mMessageBarItem && printMessage )
+  {
+    mMessageBarItem = new QgsMessageBarItem( tr( "You need to select both mask sources and masked symbol layers. If not, all configuration will be lost" ), Qgis::Warning, 0, this );
+    mMessageBar->pushItem( mMessageBarItem );
+  }
+
+  emit widgetChanged();
+}
+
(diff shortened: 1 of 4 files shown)


src/gui/qgsmaskingwidget.cpp
```diff
@@ -28,29 +28,54 @@
...
-  connect( mMaskSourcesWidget, &QgsMaskSourceSelectionWidget::changed, this, [&]()
+    mMessageBar->popWidget( mMessageBarItem );
+    delete mMessageBarItem;
+    mMessageBarItem = nullptr;
+  }
+  else if ( !mMessageBarItem && printMessage )
   {
-    emit widgetChanged();
-  } );
+    mMessageBarItem = new QgsMessageBarItem( tr( "You need to select both mask sources and masked symbol layers. If not, all configuration will be lost" ), Qgis::Warning, 0, this );
```
mhugo, line 54: "If not, all configuration will be lost" => ", otherwise the configuration will be lost" ?

mhugo:
Loooks good to me, thanks !

src/gui/qgsmaskingwidget.cpp
nyalldawson: This message is way too long for the message bar anyway. It needs to be made much more concise

src/gui/qgsmaskingwidget.h
```diff
@@ -47,10 +49,21 @@ class GUI_EXPORT QgsMaskingWidget: public QgsPanelWidget, private Ui::QgsMasking
...
+    void onSelectionChanged();
+
   private:
     QgsVectorLayer *mLayer = nullptr;
     //! Populate the mask source and target widgets
     void populate();
+
+    QgsMessageBarItem *mMessageBarItem = nullptr;
+    bool _mustPopulate = false;
 };
```
nyalldawson, line 67: mMustPopulate

src/gui/qgsmaskingwidget.cpp
troopa81: I changed it to "Select both sources and symbol layers or your configuration will be lost"

src/gui/qgsmaskingwidget.h
```diff
@@ -47,10 +49,24 @@ class GUI_EXPORT QgsMaskingWidget: public QgsPanelWidget, private Ui::QgsMasking
     //! Emitted when a change is performed
     void widgetChanged();
 
+  protected:
+
+    void showEvent( QShowEvent * ) override;
+
+  protected slots:
```
m-kuhn, line 56:
```suggestion
  private slots:
```

src/gui/qgsmaskingwidget.cpp
```diff
@@ -28,29 +28,54 @@
...
   QgsPanelWidget( parent )
 {
   setupUi( this );
+}
+
+void QgsMaskingWidget::onSelectionChanged()
+{
+  // display message if configuration is not consistent
+  bool printMessage =
+    ( mMaskTargetsWidget->selection().empty() && !mMaskSourcesWidget->selection().empty() )
```
m-kuhn, line 43: Or simply `mMaskTargetsWidget->selection().empty() != mMaskSourcesWidget->selection().empty()`?

src/gui/qgsmaskingwidget.cpp
```diff
@@ -28,29 +28,54 @@
...
 
-  connect( mMaskTargetsWidget, &QgsSymbolLayerSelectionWidget::changed, this, [&]()
+  if ( mMessageBarItem && !printMessage )
   {
-    mMaskSourcesWidget->setEnabled( ! mMaskTargetsWidget->selection().isEmpty() );
-    emit widgetChanged();
-  } );
-  connect( mMaskSourcesWidget, &QgsMaskSourceSelectionWidget::changed, this, [&]()
+    mMessageBar->popWidget( mMessageBarItem );
+    delete mMessageBarItem;
```
m-kuhn, line 49: What happens if the user clicks the `X` on the item? is there a risk of a crash?

troopa81: yes. I added a QPointer

PR #36367, opened 2020-05-12
Add additional temporal modes for vector layers

nyalldawson:
Allows temporal filtering of layers which store a start time in one
field and a duration in a different numeric field. Users are given
a choice of temporal units for the duration (e.g. seconds, days, etc)


nyalldawson:
Also includes many more modes now :)

nirvn (approved):
Nice!

PR #36368, opened 2020-05-12
[plugin manager] only show install/experimental button when needed

### LINKED ISSUE #36217: [Plugin Manager] The Install Experimental Plugin button is shown if there is no version available
The new *Install Experimental Plugin* is hidden if the only available version is stable. However, if there is **no** available version (a C++ plugin or no Internet connection), both buttons are visible, what doesn't look good (especially if the user disabled experimental plugins ;) )

experimentals

olivierdalang:
Fixes #36217 where install experimental button/message appeared when they shouldn't, following PR #34876


m-kuhn:
@borysiasty how does this look to you?

borysiasty:
@m-kuhn  Just tested - looks and works fine!

m-kuhn:
Thank you @borysiasty

PR #36372, opened 2020-05-12
port more Processing algorithms to C++

alexbruy:
## Description
Port some Processing algorithms to C++:

- Split Vector Layer
- PostGIS Execute SQL
- SpatiaLite Execute SQL
- Polygonize
- Snap Geometries


src/analysis/processing/qgsalgorithmexecutespatialitequery.cpp
```diff
@@ -0,0 +1,96 @@
...
+}
+
+QgsExecuteSpatialiteQueryAlgorithm *QgsExecuteSpatialiteQueryAlgorithm::createInstance() const
+{
+  return new QgsExecuteSpatialiteQueryAlgorithm();
+}
+
+void QgsExecuteSpatialiteQueryAlgorithm::initAlgorithm( const QVariantMap & )
+{
+  addParameter( new QgsProcessingParameterProviderConnection( QStringLiteral( "DATABASE" ), QObject::tr( "Database (connection name)" ), QStringLiteral( "spatialite" ) ) );
```
nyalldawson, line 62: This is a welcome move, but it's also an api break from the old algorithm, because the old approach allows executing commands on a non-connected db by file path.

Not sure what the best approach forward is here. @nirvn I believe we spoke about this recently too?

src/analysis/processing/qgsalgorithmsplitvectorlayer.cpp
```diff
@@ -0,0 +1,134 @@
...
+}
+
+QString QgsSplitVectorLayerAlgorithm::groupId() const
+{
+  return QStringLiteral( "vectorgeneral" );
+}
+
+QString QgsSplitVectorLayerAlgorithm::shortHelpString() const
+{
+  return QObject::tr( "Splits input vector layer into multiple layers by specified unique ID field."
```
nyalldawson, line 50:
```suggestion
  return QObject::tr( "Splits input vector layer into multiple layers by specified unique ID field.\n\n"
```

src/analysis/processing/qgsalgorithmsplitvectorlayer.cpp
```diff
@@ -0,0 +1,134 @@
...
+                      "Each of the layers created in the output folder contains all features from "
+                      "the input layer with the same value for the specified attribute. The number "
+                      ":of files generated is equal to the number of different values found for the "
```
nyalldawson, line 53:
```suggestion
                      "of files generated is equal to the number of different values found for the "
```

src/analysis/processing/qgsalgorithmsplitvectorlayer.cpp
```diff
@@ -0,0 +1,134 @@
...
+}
+
+void QgsSplitVectorLayerAlgorithm::initAlgorithm( const QVariantMap & )
+{
+  addParameter( new QgsProcessingParameterFeatureSource( QStringLiteral( "INPUT" ), QObject::tr( "Input layer" ) ) );
+  addParameter( new QgsProcessingParameterField( QStringLiteral( "FIELD" ), QObject::tr( "Unique ID field" ),
+                QVariant(), QStringLiteral( "INPUT" ) ) );
+  addParameter( new QgsProcessingParameterFolderDestination( QStringLiteral( "OUTPUT" ), QObject::tr( "Output directory" ) ) );
+  addOutput( new QgsProcessingOutputMultipleLayers( QStringLiteral( "OUTPUT_LAYERS" ), QObject::tr( "Output layers" ) ) );
+}
```
nyalldawson, line 69: What do you feel about adding an optional advanced string parameter for "output file type"? (Not essential)

src/analysis/processing/qgsalgorithmsplitvectorlayer.cpp
```diff
@@ -0,0 +1,134 @@
...
+  for ( auto it = uniqueValues.constBegin(); it != uniqueValues.constEnd(); ++it )
+  {
+    if ( feedback->isCanceled() )
+      break;
+
+    QString fileName = QStringLiteral( "%1_%2.%3" ).arg( baseName ).arg( current ).arg( outputFormat );
+    feedback->pushInfo( QObject::tr( "Creating layer: %1" ).arg( fileName ) );
+
+    sink.reset( QgsProcessingUtils::createFeatureSink( fileName, context, fields, geometryType, crs ) );
+    QString expr = QStringLiteral( "%1 = %2" ).arg( QgsExpression::quotedColumnRef( fieldName ), QgsExpression::quotedValue( *it ) );
```
nyalldawson, line 110:
```suggestion
    QString expr = QgsExpression::createFieldEqualityExpression( fieldName, *it );
``

nyalldawson (commented):
Nice! looking good!

src/analysis/processing/qgsalgorithmexecutespatialitequery.cpp
alexbruy: My intention was to make it consistent with PostGIS algorithm which works only for registered connections. Also not sure which approach better.

nyalldawson: Me either. I think we possibly need two variants like some other algs have ("registered connections")...

alexbruy: Ok, I made another algorithm for registered connections and removed API break.

src/analysis/processing/qgsalgorithmsplitvectorlayer.cpp
alexbruy: Fixed

src/analysis/processing/qgsalgorithmsplitvectorlayer.cpp
alexbruy: Done. New parameter with available extensions added, I will turn it into more human-friendly driver names later.

src/analysis/processing/qgsalgorithmsplitvectorlayer.cpp
alexbruy: Fixed.

alexbruy:
@nyalldawson any objections to merge?

PR #36373, opened 2020-05-12
Commit changes from read only fields on the attribute form

### LINKED ISSUE #36360: Expression for default values not correctly evaluated if target field is read only
QGIS master.

Take the attached sample project:

[trees.zip](https://github.com/qgis/QGIS/files/4612352/trees.zip)

it has a layer with a column "scientific" that has an expression (based on another field called "common") for its default value:

```
CASE
WHEN
"common" = 'Cork Oak'
THEN
'Quercus suber'
WHEN
"common" = 'Eucalypt'
THEN
'Eucalyptus globulus'
ELSE
'no match'
END
```

if "scientific" editable option is checked, then it all work as expected, both on insert as also on update.

If the "editable" property is unchecked then strange things happen: the above expression is correctly evaluated until the feature form is open. When confirming the data entry in the feature form the computed value is lost and the result is always "no match".

m-kuhn:
Read only fields can be calculated from default value expressions and in this case should be committed too.

Fixes #36360

Does someone know why we should *not* do this - and why this check was in place?
Read only fields are not user editable, I assume this is the main thing to make a field read only, the read only configuration comes from the attribute form configuration.
I would prefer not to backport this commit, it makes me uneasy to remove a check that looks like it has been put there on purpose. But I just can't find a good reason for it.

### DIFF
--- a/src/gui/qgsattributeform.cpp
+++ b/src/gui/qgsattributeform.cpp
@@ -356,7 +356,7 @@ bool QgsAttributeForm::saveEdits()
         for ( int i = 0; i < dstVars.count(); i++ )
         {
 
-          if ( !qgsVariantEqual( dstVars[i], srcVars[i] ) && srcVars[i].isValid() && fieldIsEditable( fieldIndexes[i] ) )
+          if ( !qgsVariantEqual( dstVars[i], srcVars[i] ) && srcVars[i].isValid() )
           {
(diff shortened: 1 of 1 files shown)


gioman:
@m-kuhn many thanks Matthias.

nirvn:
@m-kuhn , logic sounds good. Merge?

m-kuhn:
Ok, plug and pray!

PR #36374, opened 2020-05-12
New Temporal 'goto end/start' icons

rduivenvoorde:
After some discussion on the dev mailinglist:

https://lists.osgeo.org/pipermail/qgis-developer/2020-May/061161.html

@Samweli proposed to add a pipe symbol to the end/start of goto end/start icons.

So with limited Inkscape experience I copied that line and pasted it to those two icons.

So old situation:

Screenshot-20200512144711-393x115

New situation:

Screenshot-20200512144741-385x104

As said: I just copied the line from the gotoNext icon and pasted in the icons, an aligned all to the page.
If this needs more advanced graphic handling, please others step in. Thanks.

### DIFF
--- a/images/themes/default/temporal_navigation/rewindToStart.svg
+++ b/images/themes/default/temporal_navigation/rewindToStart.svg
@@ -1,6 +1,4 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<!-- Created with Inkscape (http://www.inkscape.org/) -->
-
 <svg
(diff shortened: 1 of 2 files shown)


Samweli:
Thanks @rduivenvoorde!

nyalldawson:
+1. The existing icons are in need of a final revision in any case (in order to snap components to the pixel grid + ensure compatibility with darker themes), so it's the overall shape that is important at this stage.

PR #36385, opened 2020-05-12
Fix misalignment of rasters with RPC (fixes #35796)

### LINKED ISSUE #35796: Rasters with RPC georeferencing information are misaligned
Raster files with airborne imagery may use rational polynomial coefficients (RPC) with georeferencing information. See here for an introduction on the topic:
- https://fwarmerdam.blogspot.com/2013/09/exploring-rpcs.html
- https://gdal.org/development/rfc/rfc22_rpc.html

The problem is that in order to georeference the imagery correctly, one of the input parameters of the transformation is elevation of the terrain of the image. Without that, image will be scaled incorrectly. Right now, GDAL sets the assumed elevation to zero, so the images get misaligned.

We need a way to manually set the assumed elevation (it can be passed to GDAL using RPC_HEIGHT variable).

In addition to that, we should do a guess on the good initial value of RPC_HEIGHT. From what I understand, other pieces of software use HEIGHT_OFF parameter from image's RPC and it seems to be a better default than using zero elevation.

wonder-sk:
By default GDAL would assume that the imagery has zero elevation - if that is not the case, the image would not be shown in the correct location. Fortunately we can pass RPC_HEIGHT to GDAL to use given value as the fixed elevation for the whole image.

We try to use HEIGHT_OFF coefficient ("Geodetic Height Offset") as an estimate for elevation (it seems that ENVI software does that as well). In the future we may want to use also RPC_DEM for an even more precise georeferencing.

https://gdal.org/development/rfc/rfc22_rpc.html

@rouault Would you mind having a look at this?

- I have not found a way how to pass RPC_HEIGHT to GDALAutoCreateWarpedVRT so I ended up creating a RPC-aware copy of it within QGIS code
- I do not have much experience with RPC-based rasters - I am only fixing code where QGIS does not behave correctly according to user's expectations on some sample data I got. Is it "safe" to always use HEIGHT_OFF for RPC_HEIGHT? Or do we need some way for users to manually override RPC_HEIGHT? (I am now ignoring the possibility to have DEM specified for even better accuracy)

### DIFF
--- a/src/core/qgsgdalutils.cpp
+++ b/src/core/qgsgdalutils.cpp
@@ -325,13 +325,10 @@ static void GDALWarpInitNoDataReal( GDALWarpOptions *psOptionsIn, double  dNoDat
 }
 #endif
 
-GDALDatasetH QgsGdalUtils::rpcAwareAutoCreateWarpedVrt(
-  GDALDatasetH hSrcDS,
-  const char *pszSrcWKT,
-  const char *pszDstWKT,
-  GDALResampleAlg eResampleAlg,
-  double dfMaxError,
-  const GDALWarpOptions *psOptionsIn )
+#if GDAL_VERSION_NUM < GDAL_COMPUTE_VERSION(3,2,0)
+
+GDALDatasetH GDALAutoCreateWarpedVRTEx( GDALDatasetH hSrcDS, const char *pszSrcWKT, const char *pszDstWKT, GDALResampleAlg eResampleAlg,
+                                        double dfMaxError, const GDALWarpOptions *psOptionsIn, char **papszTransformerOptions )
 {
   VALIDATE_POINTER1( hSrcDS, "GDALAutoCreateWarpedVRT", nullptr );
 
@@ -368,7 +365,6 @@ GDALDatasetH QgsGdalUtils::rpcAwareAutoCreateWarpedVrt(
       CPL_IGNORE_RET_VAL(
         GDALAdjustValueToDataType( GDALGetRasterDataType( rasterBand ),
                                    noDataValue, &bClamped, &bRounded ) );
-
       if ( !bClamped )
       {
         GDALWarpInitNoDataReal( psWO, -1e10 );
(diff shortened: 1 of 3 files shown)


src/core/qgsgdalutils.cpp
```diff
@@ -277,3 +278,208 @@ QString QgsGdalUtils::validateCreationOptionsFormat( const QStringList &createOp
...
+{
+  InitNoData( psOptionsIn->nBandCount, &psOptionsIn->padfDstNoDataReal, dNoDataReal );
+  InitNoData( psOptionsIn->nBandCount, &psOptionsIn->padfSrcNoDataReal, dNoDataReal );
+  // older GDAL also requires imaginary values to be set
+  InitNoData( psOptionsIn->nBandCount, &psOptionsIn->padfDstNoDataImag, 0 );
+  InitNoData( psOptionsIn->nBandCount, &psOptionsIn->padfSrcNoDataImag, 0 );
+}
+#endif
+
+GDALDatasetH QgsGdalUtils::rpcAwareAutoCreateWarpedVrt(
```
rouault, line 328: would you mind contributing to GDAL a GDALAutoCreateWarpedVRTEx() function that would accept a CSLConstList papszTransformerOptions additional argument ?
You could copy it here for GDAL < 3.2 and rpcAwareAutoCreateWarpedVrt() would implement only the logic to set the transformer options for the RPC case

rouault:
>     * Is it "safe" to always use HEIGHT_OFF for RPC_HEIGHT

That should be a relatively safe assumption given http://geotiff.maptools.org/rpc_prop.html which mentions "The transformation between row and column values (r,c), and normalized row and column values (rn, cn), and between the geodetic latitude, longitude, and height ( φ, λ, h ), and normalized geodetic latitude, longitude, and height (P, L, H), is defined by a set of normalizing translations (offsets) and scales that ensure all values are contained i the range -1 to +1 : H = (Height       - HEIGHT_OFF) / HEIGHT_SCALE". So for well conditioned RPC, HEIGHT_OFF should be close to some "average" height in the area of the image (or at least heights in the scene should be normally in the [HEIGHT_OFF - HEIGHT_SCALE, HEIGHT_OFF + HEIGHT_SCALE] range)

I guess it could even be helpful for GDAL to adopt this as a default behaviour instead of using 0.
@warmedam Any thoughts regarding this ? Any recollection if 0 was chosen on purpose for RPC_HEIGHT, rather than using RPC_HEIGHT ?

but using a real DEM (and adding geoid undulation correction to it !) is the must for accurate results

src/core/qgsgdalutils.cpp
wonder-sk: That sounds like a good idea - let me try to do that...

wonder-sk: Adapted to the suggested approach in 330bb17

wonder-sk:
Hmm azure build is failing because GDAL is in version 3.2.0dev in osgeo4w, but apparently this morning the package did not include https://github.com/OSGeo/gdal/pull/2565 yet - will try again tomorrow.

wonder-sk:
Oh, a missing dllexport causing failed windows builds... this should fix it: https://github.com/OSGeo/gdal/pull/2584

pl-nfwarmerdam:
Using height_off as a default height seems quite reasonable to me.  Sorry for replying late.  I assume this is only in the context of QGIS now?  I'd be a bit more nervous about that being the default within GDAL just because it will change results for lots of people.

wonder-sk:
Yes this is only when opening rasters in QGIS, the default GDAL behavior is unchanged...

wonder-sk:
Unrelated test failure in PyQgsLocalizedDataPathRegistry

PR #36399, opened 2020-05-13
Fix xyz OSM tiles too many downloads

### LINKED ISSUE #34813: QGIS loads too many XYZ tiles on first view
**How to Reproduce**

1. Load a shapefile with a small area (e.g. buildings from geofabrik's prince-edward-island-latest-free.shp)
2. Turn on a network debugger or install QGIS network logger
3. Add OpenStreetMap XYZ tiles

242 tile requests are made on my display, but only 25-50 or so are needed to cover the screen at my resolution

image

**QGIS and OS versions**
3.12 on Windows 10

```
QGIS version
3.12.0-București
QGIS code revision
cd141490ec
Compiled against Qt
5.11.2
Running against Qt
5.11.2
Compiled against GDAL/OGR
3.0.4
Running against GDAL/OGR
3.0.4
Compiled against GEOS
3.8.0-CAPI-1.13.1
Running against GEOS
3.8.0-CAPI-1.13.1 
Compiled against SQLite
3.29.0
Running against SQLite
3.29.0
PostgreSQL Client Version
11.5
SpatiaLite Version
4.3.0
QWT Version
6.1.3
QScintilla2 Version
2.10.8
Compiled against PROJ
6.3.1
Running against PROJ
OS Version
Windows 10 (10.0)
Active python plugins
qgisnetworklogger; 
db_manager; 
MetaSearch; 
processing
```

Loading 10x the tiles required is putting heavy load on the OSMF tile servers

cc @firefishy

elpaso:
Fixes #34813 by introducing a Prefetch rendering flag
and raster provider capability that can be set by
a provider when prefetching of out-of-viewport tiles
is allowed. For now it is only denied (hardcoded) for
OSM but the implementation allows per-layer setting when
a GUI element to set the flag will been created.

### DIFF
--- a/src/core/qgsmapsettings.h
+++ b/src/core/qgsmapsettings.h
@@ -310,7 +310,6 @@ class CORE_EXPORT QgsMapSettings : public QgsTemporalRangeObject
       RenderPartialOutput      = 0x200, //!< Whether to make extra effort to update map image with partially rendered layers (better for interactive map canvas). Added in QGIS 3.0
       RenderPreviewJob         = 0x400, //!< Render is a 'canvas preview' render, and shortcuts should be taken to ensure fast rendering
       RenderBlocking           = 0x800, //!< Render and load remote sources in the same thread to ensure rendering remote sources (svg and images). WARNING: this flag must NEVER be used from GUI based applications (like the main QGIS application) or crashes will result. Only for use in external scripts or QGIS server.
-      Prefetch                 = 0x1000, //!< The render is a prefetched image, not visible in the current view extent
       // TODO: ignore scale-based visibility (overview)
     };
     Q_DECLARE_FLAGS( Flags, Flag )
--- a/src/core/qgsmapsettings.h
+++ b/src/core/qgsmapsettings.h
@@ -310,6 +310,7 @@ class CORE_EXPORT QgsMapSettings : public QgsTemporalRangeObject
       RenderPartialOutput      = 0x200, //!< Whether to make extra effort to update map image with partially rendered layers (better for interactive map canvas). Added in QGIS 3.0
       RenderPreviewJob         = 0x400, //!< Render is a 'canvas preview' render, and shortcuts should be taken to ensure fast rendering
       RenderBlocking           = 0x800, //!< Render and load remote sources in the same thread to ensure rendering remote sources (svg and images). WARNING: this flag must NEVER be used from GUI based applications (like the main QGIS application) or crashes will result. Only for use in external scripts or QGIS server.
+      Prefetch                 = 0x1000, //!< The render is a prefetched image, not visible in the current view extent
       // TODO: ignore scale-based visibility (overview)
     };
     Q_DECLARE_FLAGS( Flags, Flag )
--- a/src/core/raster/qgsrasterlayerrenderer.h
+++ b/src/core/raster/qgsrasterlayerrenderer.h
@@ -80,6 +80,8 @@ class CORE_EXPORT QgsRasterLayerRenderer : public QgsMapLayerRenderer
 
     QgsRasterPipe *mPipe = nullptr;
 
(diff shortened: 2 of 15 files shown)


src/core/qgsmapsettings.h
```diff
@@ -310,6 +310,7 @@ class CORE_EXPORT QgsMapSettings : public QgsTemporalRangeObject
       RenderPartialOutput      = 0x200, //!< Whether to make extra effort to update map image with partially rendered layers (better for interactive map canvas). Added in QGIS 3.0
       RenderPreviewJob         = 0x400, //!< Render is a 'canvas preview' render, and shortcuts should be taken to ensure fast rendering
       RenderBlocking           = 0x800, //!< Render and load remote sources in the same thread to ensure rendering remote sources (svg and images). WARNING: this flag must NEVER be used from GUI based applications (like the main QGIS application) or crashes will result. Only for use in external scripts or QGIS server.
+      Prefetch                 = 0x1000, //!< The render is a prefetched image, not visible in the current view extent
```
nyalldawson, line 313: This is the same as the existing "RenderPreviewJob" flag

elpaso: I though it was for something else. I will re-use that flag then.

timlinux:
Thanks for doing this @elpaso !

src/core/raster/qgsrasterlayerrenderer.h
```diff
@@ -80,6 +80,8 @@ class CORE_EXPORT QgsRasterLayerRenderer : public QgsMapLayerRenderer
 
     QgsRasterPipe *mPipe = nullptr;
 
+    QgsRasterLayer *mLayer;
```
wonder-sk, line 83: hi @elpaso sorry only now had a chance to look at your PR...

Having a map layer as a member of QgsRasterLayerRenderer is something that we have to strictly avoid - the problem is that render() function runs in a worker thread and it is unsafe to use any methods from the map layer (for example, the layer may have been deleted in the meanwhile). The correct way is to fetch everything we need from the map layer in constructor of QgsRasterLayerRenderer - so there should be a new flag (e.g. mProviderAllowsPrefetch) instead of a full QgsRasterLayer pointer...

elpaso: Thank you for looking at this, I'll make a follow up.

src/providers/wms/qgswmsprovider.cpp
```diff
@@ -743,20 +743,24 @@ QImage *QgsWmsProvider::draw( QgsRectangle const &viewExtent, int pixelWidth, in
       return image;
     }
 
-    QgsDebugMsgLevel( QStringLiteral( "layer extent: %1,%2 %3x%4" )
-                      .arg( qgsDoubleToString( mLayerExtent.xMinimum() ),
-                            qgsDoubleToString( mLayerExtent.yMinimum() ) )
-                      .arg( mLayerExtent.width() )
-                      .arg( mLayerExtent.height() ), 3
-                    );
+    QgsDebugMsg( QStringLiteral( "layer extent: %1,%2,%3,%4 %5x%6" )
```
nyalldawson, line 746: I also didn't review fully :) Can you put these back to QgsDebugMsgLevel? it's way to noisy to output here.

elpaso: Sure but please merge the backports and I'll make a followup to be backported after

src/core/raster/qgsrasterlayerrenderer.h
elpaso: https://github.com/qgis/QGIS/pull/36422

src/providers/wms/qgswmsprovider.cpp
elpaso: https://github.com/qgis/QGIS/pull/36422

PR #36410, opened 2020-05-13
Vector tile layer - part 9 (labeling gui)

wonder-sk:
This adds a new tab in the layer styling dock widget for vector tile layers.

The new QgsVectorTileBasicLabelingWidget class is based on the code used for QgsVectorTileBasicRendererWidget

image

## Thanks

### DIFF
--- a/src/core/vectortile/qgsvectortilebasiclabeling.cpp
+++ b/src/core/vectortile/qgsvectortilebasiclabeling.cpp
@@ -16,6 +16,7 @@
 #include "qgsvectortilebasiclabeling.h"
 
 #include "qgsexpressioncontextutils.h"
+#include "qgslogger.h"
 #include "qgsvectortilelayer.h"
 #include "qgsvectortilerenderer.h"
 #include "qgsvectortileutils.h"
@@ -151,7 +152,10 @@ QList<QgsAbstractLabelProvider *> QgsVectorTileBasicLabelProvider::subProviders(
 {
   QList<QgsAbstractLabelProvider *> lst;
   for ( QgsVectorLayerLabelProvider *subprovider : qgis::as_const( mSubProviders ) )
-    lst << subprovider;
+  {
+    if ( subprovider )  // sub-providers that failed to initialize are set to null
+      lst << subprovider;
+  }
   return lst;
 }
 
@@ -170,7 +174,11 @@ bool QgsVectorTileBasicLabelProvider::prepare( QgsRenderContext &context, QSet<Q
     QgsExpressionContextScopePopper popper( context.expressionContext(), scope );
 
     mSubProviders[i]->setFields( fields );
-    mSubProviders[i]->prepare( context, attributeNames );
+    if ( !mSubProviders[i]->prepare( context, attributeNames ) )
+    {
+      QgsDebugMsg( QStringLiteral( "Failed to prepare labeling for style index" ) + QString::number( i ) );
+      mSubProviders[i] = nullptr;
+    }
   }
   return true;
 }
@@ -196,6 +204,8 @@ void QgsVectorTileBasicLabelProvider::registerTileFeatures( const QgsVectorTileR
     filterExpression.prepare( &context.expressionContext() );
 
     QgsVectorLayerLabelProvider *subProvider = mSubProviders[i];
+    if ( !subProvider )
+      continue;  // sub-providers that failed to initialize are set to null
 
     if ( layerStyle.layerName().isEmpty() )
     {
--- a/src/gui/vectortile/qgsvectortilebasiclabelingwidget.cpp
+++ b/src/gui/vectortile/qgsvectortilebasiclabelingwidget.cpp
@@ -1,0 +1,458 @@
+/***************************************************************************
+  qgsvectortilebasiclabelingwidget.cpp
+  --------------------------------------
+  Date                 : May 2020
+  Copyright            : (C) 2020 by Martin Dobias
+  Email                : wonder dot sk at gmail dot com
+ ***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+#include "qgsvectortilebasiclabelingwidget.h"
+
+#include "qgsvectortilebasiclabeling.h"
+#include "qgsvectortilelayer.h"
+
+#include "qgslabelinggui.h"
+
+#include <QMenu>
+
+///@cond PRIVATE
+
+
+class QgsVectorTileBasicLabelingListModel : public QAbstractListModel
+{
+  public:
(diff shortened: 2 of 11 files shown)


src/gui/vectortile/qgsvectortilebasiclabelingwidget.cpp
```diff
@@ -0,0 +1,458 @@
...
+#include "qgslabelinggui.h"
+
+#include <QMenu>
+
+///@cond PRIVATE
+
+
+class QgsVectorTileBasicLabelingListModel : public QAbstractListModel
+{
+  public:
```
nyalldawson, line 30: Q_OBJECT

src/gui/vectortile/qgsvectortilebasiclabelingwidget.cpp
```diff
@@ -0,0 +1,458 @@
...
+  }
+  return QVariant();
+}
+
+QVariant QgsVectorTileBasicLabelingListModel::headerData( int section, Qt::Orientation orientation, int role ) const
+{
+  if ( orientation == Qt::Horizontal && role == Qt::DisplayRole && section >= 0 && section < 5 )
+  {
+    QStringList lst;
+    lst << tr( "Label" ) << tr( "Layer" ) << tr( "Min. zoom" ) << tr( "Max. zoom" ) << tr( "Filter" );
```
nyalldawson, line 132: What's our convention re capitalization in headers? I'm guessing it's Title Case, but not 100% sure....

src/gui/vectortile/qgsvectortilebasiclabelingwidget.cpp
```diff
@@ -0,0 +1,458 @@
...
+
+  QgsPalLayerSettings labelSettings = style.labelSettings();
+  if ( labelSettings.layerType == QgsWkbTypes::UnknownGeometry )
+    labelSettings.layerType = style.geometryType();
+
+  QgsSymbolWidgetContext context;
+  context.setMapCanvas( mMapCanvas );
+  context.setMessageBar( mMessageBar );
+
+  QgsVectorLayer *vectorLayer = nullptr;  // TODO: have a temporary vector layer with sub-layer's fields?
```
nyalldawson, line 382: I think you'll need this in order to get the expression context correctly setup for the labeling widget

nyalldawson (commented):
Looking good! One thing I'm unsure about is whether you'll need to specifically handle passing the suitable expression context to the widget -- this may be taken care of already by passing the canvas to it, but you'll at least need to add the correct fields to the context.

src/core/vectortile/qgsvectortilebasiclabeling.cpp
```diff
@@ -170,7 +174,11 @@ bool QgsVectorTileBasicLabelProvider::prepare( QgsRenderContext &context, QSet<Q
     QgsExpressionContextScopePopper popper( context.expressionContext(), scope );
 
     mSubProviders[i]->setFields( fields );
-    mSubProviders[i]->prepare( context, attributeNames );
+    if ( !mSubProviders[i]->prepare( context, attributeNames ) )
+    {
+      QgsDebugMsg( QStringLiteral( "Failed to prepare labeling for style index" ) + QString::number( i ) );
```
m-kuhn, line 179: Is that something to send to the message log?

src/gui/vectortile/qgsvectortilebasiclabelingwidget.cpp
wonder-sk: According to point 3 in https://docs.qgis.org/3.10/en/docs/developers_guide/hig.html it should be Title Case, but I just copied the convention from what is already being used in rule-based renderer/labeling widget. If you prefer Title Case then I guess I should also update the existing texts there (which will break translations)

src/gui/vectortile/qgsvectortilebasiclabelingwidget.cpp
wonder-sk: Yeah, I have a similar to-do item for the "basic renderer" widget, and a bunch of other bits to make the GUI more user friendly. If you don't mind I would address those things afterwards during feature freeze.

src/core/vectortile/qgsvectortilebasiclabeling.cpp
wonder-sk: I don't think so - prepare() fails e.g. if the labeling expression is invalid, which is something that happens normally while editing labeling expression...

src/gui/vectortile/qgsvectortilebasiclabelingwidget.cpp
wonder-sk: Just wondering - why add it if it is not needed here? It's a private class, there are no signals/slots, there's no qobject_cast in use...

src/core/vectortile/qgsvectortilebasiclabeling.cpp
m-kuhn: Isn't editing labeling expressions something that's done in released versions where debug messages are not visible?

src/gui/vectortile/qgsvectortilebasiclabelingwidget.cpp
nyalldawson: I've looked into this before, and advice from qt (and kdab) is that it's required regardless of whether the subclass signals or slots.

src/core/vectortile/qgsvectortilebasiclabeling.cpp
wonder-sk: Correct. This is just a message for developers, meaning that something may have gone wrong... Or maybe I am missing your point :-)

m-kuhn: Let's see if I can be more precise :)
How will "I" as someone configuring a project realize that my expression has an error did not work? I guess often this will be triggered by faulty expressions (and not by source code problems a developer needs to take care of)?

wonder-sk: Your labels won't show up - and the expression will be red in the GUI... just like with labeling of vector layers...

src/gui/vectortile/qgsvectortilebasiclabelingwidget.cpp
wonder-sk: Okay cool - fixed in 2cd15ea - I see it in the Qt docs now: "We strongly recommend the use of this macro in all subclasses of QObject regardless of whether or not they actually use signals, slots and properties, since failure to do so may lead certain functions to exhibit strange behavior."

src/core/vectortile/qgsvectortilebasiclabeling.cpp
m-kuhn: Yeath, true, that's the same amount of information as in this debug message :+1:

stev-0:
I think one enhancement to the GUI would be the ability to see what type of layer (point, line or polygon) the label has been created for - at the moment this isn't exposed in the interface so could confuse people

wonder-sk:
@stev-0 good suggestion... how to best do that? maybe to have a point/line/polygon icon by the layer name?

stev-0:
I think that would work yes. Maybe at least initially a new column with the read-only type in there, but I guess ultimately it would be good to be able to change this on the fly.

PR #36426, opened 2020-05-14
[FEATURE][temporal] Add off and fixed range modes to temporal navigation

nirvn:
## Description

This PR adds two new navigation mode to the temporal navigation object: an off mode (not so exiting) and a fixed range mode.

GIF time:
Peek 2020-05-14 15-37

The fixed range mode is quite useful when users are interested in being precise with their start and end range.


src/core/qgstemporalnavigationobject.h
```diff
@@ -255,6 +281,8 @@ class CORE_EXPORT QgsTemporalNavigationObject : public QgsTemporalController, pu
     //! The controller temporal navigation extent range.
     QgsDateTimeRange mTemporalExtents;
 
+    NavigationMode mNavigationMode = Animated;
```
nyalldawson, line 284: I think we should default to off here

src/gui/qgstemporalcontrollerwidget.cpp
```diff
@@ -167,6 +178,58 @@ void QgsTemporalControllerWidget::setWidgetStateFromProject()
...
+  QgsProject::instance()->writeEntry( QStringLiteral( "TemporalControllerWidget" ), QStringLiteral( "/NavigationMode" ),
+                                      static_cast<int>( QgsTemporalNavigationObject::Animated ) );
+
+  mNavigationObject->setNavigationMode( QgsTemporalNavigationObject::Animated );
+  setWidgetStateFromNavigationMode( QgsTemporalNavigationObject::Animated );
+}
+
+void QgsTemporalControllerWidget::setWidgetStateFromNavigationMode( const QgsTemporalNavigationObject::NavigationMode mode )
+{
+  mNavigationOff->setChecked( mode == QgsTemporalNavigationObject::NavigationOff );
```
nyalldawson, line 210: This would be nicer if you used a stacked widget -- then you'd just set the current index of the stacked widget here. It would help avoid the extra padding which shows up when you hide elements like this (Because qt still leaves the layout padding for them)

src/core/qgstemporalnavigationobject.h
nirvn: We could -- I was aiming at default behavior parity, open to having it off by default.

src/gui/qgstemporalcontrollerwidget.cpp
nirvn: True -- I thought it was better to go down that path to avoid duplicating the QDateTimeEdit widgets. 

If we opt-in stacked widgets, we'll have to maintain a pair of QDateTimeEdit widgets for the start datetime and end datetime.

src/core/qgstemporalnavigationobject.h
nyalldawson: Yeah - but it's bothered me a little that we default to a temporal filter immediately on adding a temporal source. It could potentially be confusing to users who would expect to see all the features/layer content show up immediately on adding a new layer

src/gui/qgstemporalcontrollerwidget.cpp
nyalldawson: I think duplication is fine, and is outweighed by the benefits of the stacked widget approach

src/core/qgstemporalnavigationobject.h
nirvn: @nyalldawson , OK, let's change the default mode.

src/gui/qgstemporalcontrollerwidget.cpp
nyalldawson: Also, if we go stacked it'll be easier to add the double ranged slider in future ;)

nyalldawson (approved):

src/core/qgstemporalnavigationobject.h
vcloarec: as with off by default, mesh layer could be not visible, I purpose this PR #36494

PR #36427, opened 2020-05-14
[Server] Expression override for online resource in service capabilities

### LINKED ISSUE #36425: Allow setting "OnlineResource" for QGIS server dynamically
**Feature description.**

It would be nice to be able to set certain server variables as environment variables to have different values in development and production. For example "OnlineResource".

Probably it's most flexible if they are defined as expressions/data defined properties, so they can use environment variables through `env()` or other variables and logic.

signedav:
In ***Project Properties > QGIS Server > Service Capabilities*** values can be defined that are send back on capabilities requests. One of them is ***Online resource*** and this can have an override expression with the global and the projet scope.

BEES

Returns on `...bees_in_ticino.qgs&SERVICE=WMS&REQUEST=GetCapabilities`
```
<OnlineResource xmlns:xlink="http://www.w3.org/1999/xlink" xlink:type="simple" xlink:href="www.mayasbees.ch/bees_in_ticino"/>
```

resolves #36425

## Additional ideas:
- Are there any inputs on what properties it would make sense to have an override as well in future?
- Maybe in future it would make sense to be able to use the input request properties as kind of `serverRequestContext` appended to the expression evaluation.

### DIFF
--- a/python/core/auto_generated/qgsproject.sip.in
+++ b/python/core/auto_generated/qgsproject.sip.in
@@ -406,6 +406,18 @@ a hierarchy of keys and corresponding values
    The key string must be valid xml tag names in order to be saved to the file.
 %End
 
+    bool writeEntry( const QString &scope, const QString &key, const QgsProperty &value );
+%Docstring
+Write a QgsProperty entry to the project file.
+
+Keys are '/'-delimited entries, implying
+a hierarchy of keys and corresponding values
+
+.. note::
+
+   The key string must be valid xml tag names in order to be saved to the file.
+%End
+
     QStringList readListEntry( const QString &scope, const QString &key, const QStringList &def = QStringList(), bool *ok = 0 ) const;
 %Docstring
 Key value accessors
@@ -418,6 +430,7 @@ implying a hierarchy of keys and corresponding values
     int readNumEntry( const QString &scope, const QString &key, int def = 0, bool *ok = 0 ) const;
     double readDoubleEntry( const QString &scope, const QString &key, double def = 0, bool *ok = 0 ) const;
     bool readBoolEntry( const QString &scope, const QString &key, bool def = false, bool *ok = 0 ) const;
+    QgsProperty readPropertyEntry( const QString &scope, const QString &key, QgsProperty def = QgsProperty(), bool *ok = 0 ) const;
 
 
     bool removeEntry( const QString &scope, const QString &key );
--- a/python/core/auto_generated/qgsproject.sip.in
+++ b/python/core/auto_generated/qgsproject.sip.in
@@ -430,7 +430,7 @@ implying a hierarchy of keys and corresponding values
     int readNumEntry( const QString &scope, const QString &key, int def = 0, bool *ok = 0 ) const;
     double readDoubleEntry( const QString &scope, const QString &key, double def = 0, bool *ok = 0 ) const;
     bool readBoolEntry( const QString &scope, const QString &key, bool def = false, bool *ok = 0 ) const;
-    QgsProperty readPropertyEntry( const QString &scope, const QString &key, QgsProperty def = QgsProperty(), bool *ok = 0 ) const;
+    QgsProperty readPropertyEntry( const QString &scope, const QString &key, const QgsProperty &def = QgsProperty(), bool *ok = 0 ) const;
 
 
     bool removeEntry( const QString &scope, const QString &key );
--- a/python/core/auto_generated/qgsproject.sip.in
+++ b/python/core/auto_generated/qgsproject.sip.in
@@ -55,6 +55,13 @@ open within the main QGIS application.
       AvoidIntersectionsLayers,
     };
 
+    enum DataDefinedServerProperty
+    {
+      NoProperty,
+      AllProperties,
+      WMSOnlineResource,
+    };
+
     static QgsProject *instance();
 %Docstring
 Returns the QgsProject singleton instance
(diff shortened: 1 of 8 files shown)


nyalldawson (commented):
I'm -1 to the approach used here -- this should use a proper QgsPropertyCollection attached on a QgsProject level, so that we can re-use it for other similar project level settings.

signedav:
Thanks for your input @nyalldawson 
Do you mean storing a `QgsPropertyCollection` to the `QgsProject` aside the current `QgsProjectPropertyKey` or writing the `QgsPropertyCollection` into a `QgsProjectProperty`?

I thought it's more clear to just write a `QgsProperty` to a `QgsProjectProperty`, because here we don't have two key-value mappings. But I'm open to make it different, since maybe I don't see some advantages the QgsPropertyCollection has.

elpaso:
I'm -1 on allowing the environment to implicitly change the server configuration without following the configuration pattern that we currently use for the server.

See my comment here:  https://github.com/qgis/QGIS/issues/36425#issuecomment-628543109

@pblottiere @rldhont I'd like to hear your opinion.

m-kuhn:
> I'm -1 on allowing the environment to implicitly change the server configuration without following the configuration pattern that we currently use for the server.

I don't think this comment reflects what this pull request does. It offers to override with an expression (which could potentially involve an env variable, but that's at the discretion of the server admin, just like a label could use an expression to show an env variable). Or am I missing a point here?

pblottiere:
Hello :),

>> I'm -1 on allowing the environment to implicitly change the server configuration without following the configuration pattern that we currently use for the server.

> I don't think this comment reflects what this pull request does. It offers to override with an expression (which could potentially involve an env variable, but that's at the discretion of the server admin, just like a label could use an expression to show an env variable). Or am I missing a point here?

I'm not totally sure what @elpaso had in mind but I think I get it.`Online Resource` is not just a label, but instead contains "something" which is returned in the `GetCapabilities` document. Therefore, it has to follow strict rules for being OGC compliant and we have to be careful regarding security aspects.

The possibility to customize the `Online Resource` field with a static string through QGIS Desktop is one thing, but allowing this same field to be customized thanks to an environment variable is another. An environment variable is easily forgotten but can lead in this particular case to serious issues.

So, generally speaking, I'd say that I share the concern of @elpaso. 

However, it seems acceptable to me to allow a server admin (which is generally a reasonable person) to override this parameter with an environment variable (which is possible with this PR with `env()`) **as soon as  it's explicitly stated that it can lead to security and OGC compliancy issues** (in the documentation and maybe with a `danger` icon in the gui?). 

BTW, I think we can compare this situation to the `Advanced Setting Editor` where a "I will be careful, I promise!" is displayed...

elpaso:
@pblottiere my main concern is that the standard pattern for server configuration is to use the documented `QGS_SERVER_...` vars through environment or settings.ini.

By taking an `env()` var from a data-defined expression for a particular aspect of OGC SERVICE capabilities we are introducing a new way of configuring the server behavior that could lead to side-effects.

Let aside that I don't see this particular use case: onlineresource is automatically calculated from the request URL and it's fine 99.99% of the times, when it's not  it can be easily rewritten by nginx or apache or a server plugin and if that's still not enough it can be statically overridden in the project.

But speaking with @m-kuhn looks like he has some good ideas we could mitigate the dangerous effects of `env()` in a server expression. Let's wait and see.

pblottiere:
> But speaking with @m-kuhn looks like he has some good ideas we could mitigate the dangerous effects of env() in a server expression. Let's wait and see.

Perfect :+1:

m-kuhn:
@elpaso is it okay to open this as an issue/bug report? Given that the discussed ideas have security implications (unrelated to this pull request here), I don't know if I should document this publicly.

If it's ok, I would like to stop discussing environment variables since this pull request does not imply anything with respect to these. It's really just "dynamic configuration" of what GetCapabilities returns.

So, to move on with this pull request here:

@pblottiere you seem to be ok with it if it's clearly documented. I would propose a placeholder text that mentions 1. the default value if left empty, 2. potential security implications (with more details in a tooltip). Is this acceptable?

pblottiere:
> If it's ok, I would like to stop discussing environment variables since this pull request does not imply anything with respect to these. It's really just "dynamic configuration" of what GetCapabilities returns.

It's ok for me.

> @pblottiere you seem to be ok with it if it's clearly documented. I would propose a placeholder text that mentions 1. the default value if left empty, 2. potential security implications (with more details in a tooltip). Is this acceptable?

It sounds good to me, thanks Matthias :+1:

m-kuhn:
@pblottiere can you come up with a good example regarding security implications?

Something like "This value is used in the GetCapabilities request. Be careful what you put here."? It reads a bit odd, since that would apply to pretty much any of the settings on this page though :)

pblottiere:
> Something like "This value is used in the GetCapabilities request. Be careful what you put here."? It reads a bit odd, since that would apply to pretty much any of the settings on this page though :)

Indeed :). So instead of adding a dedicated message for the `Online Resource` parameter, we could maybe add a broader warning for the whole `Service Capabilities` section. Something like "These parameters are used to generate the GetCapabilities document and shall be chosen carefully to avoid interoperability and security issues".

What do you think?

m-kuhn:
That's definitely better :+1: 
But no strong opinion, I would expect a sysadmin to know that already :smile: 

And the placeholder text should indicate what happens when it's left empty.

pblottiere:
> But no strong opinion, I would expect a sysadmin to know that already

Yes but these parameters can be configured by the map maker through QGIS Desktop and be totally missed by the underlying sysadmin (even if it should not).

So an explicit message will probably dissuade end-users to play with these parameters (unless they know what they're doing).

pblottiere:
> And the placeholder text should indicate what happens when it's left empty.

You're right.

elpaso:
@m-kuhn yes, better to not disclose the env issue for now... Let's fix it ASAP.

signedav:
Thanks for the discussion, guys.

The GUI would look like this:
image

signedav:
@nyalldawson Is the integration of the `QgsPropertyCollection` in `QgsProject` here what you had in mind?

m-kuhn:
@pblottiere does this look good to you now?

pblottiere:
> does this look good to you now?

Yes, thanks @m-kuhn and @signedav :+1:

Allowing implicit environment variable (without `QGIS_SERVER_` prefix) to configure QGIS Server is an open-door to potential issues, so I'm happy to see the big orange warning panel :D.

m-kuhn:
Thanks for the feedback!
I'll make sure the customer who ordered this will use `QGIS_SERVER_` prefixed variables there.

PR #36438, opened 2020-05-14
Minor typo corrected

DiGro:
## Description

Line 1768 : "some parameter value" should probably be plural: "some parameter values", otherwise better to change "some"with "a"

Should probably be backported to 3.10. Not sure though

### DIFF
--- a/src/providers/wms/qgswmsprovider.cpp
+++ b/src/providers/wms/qgswmsprovider.cpp
@@ -1765,7 +1765,7 @@ void QgsWmsProvider::parseServiceException( QDomElement const &e, QString &error
     if ( locator == QLatin1String( "time" ) )
       errorText = tr( "Request contains a time value that does not match any available layer in the server." );
     else
-      errorText = tr( "Request contains some parameter value that does not match any available layer in the server" );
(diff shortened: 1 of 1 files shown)


src/providers/wms/qgswmsprovider.cpp
```diff
@@ -1765,7 +1765,7 @@ void QgsWmsProvider::parseServiceException( QDomElement const &e, QString &error
     if ( locator == QLatin1String( "time" ) )
       errorText = tr( "Request contains a time value that does not match any available layer in the server." );
     else
-      errorText = tr( "Request contains some parameter value that does not match any available layer in the server" );
+      errorText = tr( "Request contains some parameter values that does not match any available layer in the server" );
```
m-kuhn, line 1768:
```suggestion
      errorText = tr( "Request contains some parameter values that do not match any available layer in the server" );
```

If it's plural, then let's go the full way

DiGro:
Matthias,

obviously you are right. I missed that. Thanks. Added the suggestion

m-kuhn (approved):

PR #36447, opened 2020-05-14
[postgres] Compile make_date, make_datetime, make_time functions

nyalldawson:
Allows (most) vector temporal filters to be performed server-side,
utilising indexes on the temporal fields. Ultimately makes it possible
to animate huge postgres tables using the new temporal framework.


3nids (approved):

nyalldawson:
Thanks!

PR #36452, opened 2020-05-14
Cloud Optimized Geotiff (COG)  GDAL algorithms

olivierdalang:
## Description

This adds :

- **CreateCog** algorithm : converts a raster to a Cloud Optimized Geotiff using `gdal_translate -of COG -co TILING_SCHEME=GoogleMapsCompatible -co COMPRESS=JPEG -co BIGTIFF=yes input.tif output.tif`.

- **ValidateCog** algorithm : checks whether a tiff is a valid COG using [this implementation]( https://github.com/OSGeo/gdal/blob/master/gdal/swig/python/samples/validate_cloud_optimized_geotiff.py).

- A small modification to `GdalUtils.py` for GDAL algorithms to properly report failures (exit codes != 0 still were reported as successes). This is needed so ValidateCog actually looks like it fails if the COG isn't valid. It will probably also improve UX for other algorihtms where failure would not be raised properly (e.g. that "the results couldn't be loaded" instead of "the algorithm failed" message).

#### Notes 

- **this requires GDAL >= 3.1**, I couldn't find what GDAL versions we will support on future versions of QGIS ?

- no worries if this doesn't make it to 3.14, especially if we don't have GDAL 3.1 in 3.14.

- for those not familiar with COG, it's an awesome format : a regular GeoTiff, but with pyramids carefully crafted so that it's possible to retrieve tiles through HTTP request using byte-ranges. Allows you to "stream" a multi gigabytes raster into QGIS directly from a webserver, with a file that also works perfectly locally. More info [here](https://www.cogeo.org/).

### DIFF
--- a/python/plugins/processing/algs/gdal/CreateCog.py
+++ b/python/plugins/processing/algs/gdal/CreateCog.py
@@ -1,0 +1,113 @@
+# -*- coding: utf-8 -*-
+
+"""
+***************************************************************************
+    translate.py
+    ---------------------
+    Date                 : August 2012
+    Copyright            : (C) 2012 by Victor Olaya
+    Email                : volayaf at gmail dot com
+***************************************************************************
+*                                                                         *
+*   This program is free software; you can redistribute it and/or modify  *
+*   it under the terms of the GNU General Public License as published by  *
+*   the Free Software Foundation; either version 2 of the License, or     *
+*   (at your option) any later version.                                   *
+*                                                                         *
+***************************************************************************
+"""
+
+__author__ = 'Victor Olaya'
+__date__ = 'August 2012'
+__copyright__ = '(C) 2012, Victor Olaya'
+
+import os
+
+from qgis.PyQt.QtGui import QIcon
+
+from qgis.core import (QgsRasterFileWriter,
+                       QgsProcessingException,
+                       QgsProcessingParameterDefinition,
+                       QgsProcessingParameterRasterLayer,
+                       QgsProcessingParameterString,
+                       QgsProcessingParameterRasterDestination)
+from processing.algs.gdal.GdalAlgorithm import GdalAlgorithm
+from processing.algs.gdal.GdalUtils import GdalUtils
+
+pluginPath = os.path.split(os.path.split(os.path.dirname(__file__))[0])[0]
+
+
+class CreateCog(GdalAlgorithm):
+
+    INPUT = 'INPUT'
+    OPTIONS = 'OPTIONS'
+    EXTRA = 'EXTRA'
+    OUTPUT = 'OUTPUT'
+
+    def __init__(self):
+        super().__init__()
+
+    def initAlgorithm(self, config=None):
+
+        self.addParameter(QgsProcessingParameterRasterLayer(self.INPUT, self.tr('Input layer')))
+
+        options_param = QgsProcessingParameterString(self.OPTIONS,
+                                                     self.tr('Additional creation options'),
+                                                     defaultValue='',
+                                                     optional=True)
+        options_param.setFlags(options_param.flags() | QgsProcessingParameterDefinition.FlagAdvanced)
+        options_param.setMetadata({
+            'widget_wrapper': {
+                'class': 'processing.algs.gdal.ui.RasterOptionsWidget.RasterOptionsWidgetWrapper'}})
+        self.addParameter(options_param)
+
+        extra_param = QgsProcessingParameterString(self.EXTRA,
+                                                   self.tr('Additional command-line parameters'),
+                                                   defaultValue=None,
+                                                   optional=True)
+        extra_param.setFlags(extra_param.flags() | QgsProcessingParameterDefinition.FlagAdvanced)
+        self.addParameter(extra_param)
+
+        self.addParameter(QgsProcessingParameterRasterDestination(self.OUTPUT,
+                                                                  self.tr('Converted')))
+
+    def name(self):
+        return 'create_cog'
+
+    def displayName(self):
+        return self.tr('Create COG (Cloud Optimized Geotiff)')
+
+    def group(self):
+        return self.tr('Raster conversion')
+
+    def groupId(self):
+        return 'rasterconversion'
+
+    def icon(self):
+        return QIcon(os.path.join(pluginPath, 'images', 'gdaltools', 'translate.png'))
+
+    def commandName(self):
+        return 'gdal_translate'
+
+    def getConsoleCommands(self, parameters, context, feedback, executing=True):
+        inLayer = self.parameterAsRasterLayer(parameters, self.INPUT, context)
+        if inLayer is None:
+            raise QgsProcessingException(self.invalidRasterError(parameters, self.INPUT))
+
+        out = self.parameterAsOutputLayer(parameters, self.OUTPUT, context)
+        self.setOutputValue(self.OUTPUT, out)
+
+        arguments = ['-of', 'COG', '-co', 'TILING_SCHEME=GoogleMapsCompatible', '-co', 'COMPRESS=JPEG', '-co', 'BIGTIFF=yes']
+
+        options = self.parameterAsString(parameters, self.OPTIONS, context)
+        if options:
+            arguments.extend(GdalUtils.parseCreationOptions(options))
+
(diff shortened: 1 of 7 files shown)


blacha:
This is really cool, this would make it easier for a lot of people to generate COGs!

I have been working with the GDAL COG driver for quite a while, here are a few useful settings you may want to add options for or alter the defaults: 

- `COMPRESS` We alternate between jpeg/webp for our compression options depending on what we are targeting with the COG (anything with alpha we use webp) so it would be useful to be able to choose the compression format.
- `NUM_THREADS` by default I believe this limits the compression/reprojection to only the main thread
- `RESAMPLING` defaults to `nearest` which often results in not great looking imagery COGs 
- `QUALITY` defaults to 75 which generally results in jpeg artifacts inside of COGs, possibly up this to 85-95 (?) 

Also in GDAL 3.2 (?) there will be a option to provide your own `TILING_SCHEME` !

python/plugins/processing/algs/gdal/CreateCog.py
```diff
@@ -0,0 +1,113 @@
+# -*- coding: utf-8 -*-
+
+"""
+***************************************************************************
+    translate.py
+    ---------------------
+    Date                 : August 2012
+    Copyright            : (C) 2012 by Victor Olaya
+    Email                : volayaf at gmail dot com
+***************************************************************************
```
alexbruy, line 10: Header should be updated.

python/plugins/processing/algs/gdal/CreateCog.py
```diff
@@ -0,0 +1,113 @@
...
+*   the Free Software Foundation; either version 2 of the License, or     *
+*   (at your option) any later version.                                   *
+*                                                                         *
+***************************************************************************
+"""
+
+__author__ = 'Victor Olaya'
+__date__ = 'August 2012'
+__copyright__ = '(C) 2012, Victor Olaya'
+
```
alexbruy, line 23: Also need to be updated.

python/plugins/processing/algs/gdal/ValidateCog.py
```diff
@@ -0,0 +1,85 @@
+# -*- coding: utf-8 -*-
+
+"""
+***************************************************************************
+    gdal2tiles.py
+    ---------------------
+    Date                 : January 2016
+    Copyright            : (C) 2016 by Médéric Ribreux
+    Email                : mederic dot ribreux at medspx dot fr
+***************************************************************************
```
alexbruy, line 10: Update header

python/plugins/processing/algs/gdal/ValidateCog.py
```diff
@@ -0,0 +1,85 @@
...
+*   it under the terms of the GNU General Public License as published by  *
+*   the Free Software Foundation; either version 2 of the License, or     *
+*   (at your option) any later version.                                   *
+*                                                                         *
+***************************************************************************
+"""
+
+__author__ = 'Médéric Ribreux'
+__date__ = 'January 2016'
+__copyright__ = '(C) 2016, Médéric Ribreux'
```
alexbruy, line 22: Update

python/plugins/processing/algs/gdal/ValidateCog.py
```diff
@@ -0,0 +1,85 @@
...
+        return 'rasterconversion'
+
+    def icon(self):
+        return QIcon(os.path.join(pluginPath, 'images', 'gdaltools', 'tiles.png'))
+
+    def commandName(self):
+        return os.path.join(os.path.dirname(__file__), 'helpers', 'validate_cloud_optimized_geotiff')
+
+    def flags(self):
+        return super().flags() | QgsProcessingAlgorithm.FlagDisplayNameIsLiteral
```
alexbruy, line 67: Display name is translatable string, not literal.

python/plugins/processing/algs/gdal/helpers/validate_cloud_optimized_geotiff.py
```diff
@@ -0,0 +1,382 @@
+#!/usr/bin/env python3
```
alexbruy, line 1: Do we really need to copy this into QGIS source? If this script is part of GDAL distribution better to keep it outside of QGIS tree to reduce maintenance overhead.

alexbruy (changes requested):
Not sure if need a separate algorithm to create COGs as this is just a `gdal_translate` with specific creation/command-line options.

python/plugins/processing/algs/gdal/helpers/validate_cloud_optimized_geotiff.py
olivierdalang: As far as I understood, it's not part of the distribution, but only a sample file ? At least the file doens't seem present when installed with OSGeo4W.

olivierdalang:
@alexbruy 
> Not sure if need a separate algorithm to create COGs as this is just a `gdal_translate` with specific creation/command-line options.

Yes, the only issue is that it requires both setting `-of COG` instead of `-of GTiff` and adding the appropriate `-co` settings (which we could provide as an "additional creation options preset"). So we'd need to add an "override output format [optional]" to the gdal_translate algorithm, and then it would be 2 steps for the user, which is still good enough.

Initially I wanted to add this to the export raster dialog, but it fails because of `GDALDriver::Create() ... no create method implemented for this format.`.

python/plugins/processing/algs/gdal/CreateCog.py
olivierdalang: Even if it's a copy/paste ? (I update just the dates ?)

palmerj:
> Yes, the only issue is that it requires both setting -of COG instead of -of GTiff and adding the appropriate -co settings (which we could provide as an "additional creation options preset"). So we'd need to add an "override output format [optional]" to the gdal_translate algorithm, and then it would be 2 steps for the user, which is still good enough.

Seems like a good idea to re-use the existing algorithm. But for me the UI of the current Translate algorithm is not the best, so it will need some careful design to allow for more advanced creation options.  

> But the main question is whether we can assume GDAL 3.1 is available for everybody (I guess not) ? How to deal with that ? Load the new logic only conditionally ?

That seems like the best idea. You can query GDAL for the available drivers and then query the driver metadata to see what creation options are available (which will be good for further advanced COG options that are coming in GDAL 3.2)

python/plugins/processing/algs/gdal/GdalAlgorithmProvider.py
```diff
@@ -146,6 +148,7 @@ def loadAlgorithms(self):
             ClipRasterByExtent(),
             ClipRasterByMask(),
             ColorRelief(),
+            CreateCog(),
```
m-kuhn, line 151: To answer your main question.
Why not doing something like this, so it'd be available wherever possible?

```
if int(osgeo.gdal.VersionInfo()) >= 3010000:
    self.algs+= [CreateCog()]
```

python/plugins/processing/algs/gdal/helpers/validate_cloud_optimized_geotiff.py
m-kuhn: @rouault can you shed some light here?

m-kuhn:
Good stuff.

IMHO the user benefits from a dedicated algorithm on the GUI. This will help to make it discoverable and come with relevant and preconfigured options.

Code wise, it *might* be good to reuse existing algorithms, e.g. by simply providing a model that acts as a proxy to gdal_translate. Or by subclassing the existing alg. I don't have any strong opinion here.

python/plugins/processing/algs/gdal/helpers/validate_cloud_optimized_geotiff.py
rouault: yes, the script is not installed by GDAL. It just lives in its git repository. In principle, the output of the GDAL COG driver should be conformant :-) (the script is used in GDAL autotest suite to check that)

python/plugins/processing/algs/gdal/GdalAlgorithmProvider.py
rouault: or `if osgeo.gdal.GetDriverByName('COG') is not None:`

PR #36453, opened 2020-05-15
Add ability to export animation frames

nyalldawson:
Allows exporting of temporal animation frames to successive images,
for later stitching together in an external application.

Users have precise control over the image size and map extent.

There's LOTS of scope for improvements here, which range from simple UX tweaks (using the message bar instead of message boxes, showing links to exported images after export, better warning messages) through to enhancements (exporting in a background thread! multithreaded exports!!) 

But I thought it was important functionality to have for 3.14, so even in it's current unrefined state I would like to see it merged prior to freeze


3nids (approved):

nirvn (approved):

nyalldawson:
Thank you!

PR #36464, opened 2020-05-15
Fix drag and dropping of layer tree items

nirvn:
## Description

@3nids , this is a follow up to a PR you recently merged.


3nids:
thanks for the fix, could you check if it still works across several QGIS instances?

3nids (approved):

PR #36472, opened 2020-05-15
Turn off editing and dragging for "Current environment variables" table widget in System settings window

agiudiceandrea:
## Description

Edit triggers are wrongly set on for the Current environment variables table widget in Options | System settings window, while the variables values are read only in that table.

This affects also 3.10 branch.

The patch sets to QAbstractItemView::NoEditTriggers the editTriggers property of the mCurrentVariablesTable QTableWidget in order to make it not editable.


nyalldawson:
This would be better fixed by removing the Qt::ItemIsEditable and Qt::ItemIsDragEnabled flags from qgsoptions.cpp:258

agiudiceandrea:
Thanks @nyalldawson for the tip.
Removed Qt::ItemIsEditable and Qt::ItemIsDragEnabled flags in from mCurrentVariablesTable items in qgsoptions.cpp and editTriggers, dragEnabled and dragDropMode from mCurrentVariablesTable widget in qgsoptionsbase.ui.

nyalldawson (approved):

PR #36487, opened 2020-05-16
[Backport release-3_12] Server GetLegendGraphics add RULELABEL AUTO option

uprel:
Manual backporting #36153


nirvn:
3.12 has reached its end of life. 3.10 LTR and master (what will become 3.14) are they only two branches alive.

uprel:
OK, that's why label was gone and I didn't know reason for it.

PR #36494, opened 2020-05-17
[TEMPORAL] set mesh layer visible when animated mode is off

vcloarec:
With #36426, the default temporal mode is "navigation off". 
But with this mode when mesh has only temporal datasets the mesh is not visible after loading it, and that could be disturbing for user.
With this new PR, when a temporal layer is loaded for the first time, if it is a mesh layer the temporal mode is set to "animated".
As I don't know very well how other layer type work with different navigation mode, this behavior is only for mesh layer, maybe it could be too for other types.

Some refactoring is also done.

### DIFF
--- a/src/core/mesh/qgsmeshlayer.cpp
+++ b/src/core/mesh/qgsmeshlayer.cpp
@@ -82,9 +82,7 @@ void QgsMeshLayer::setDefaultRendererSettings()
          meta.minimum() == std::numeric_limits<double>::quiet_NaN() )
       meshSettings.setEnabled( true );
 
-    // If the mesh is non temporal, set the static scalar dataset
-    if ( !mDataProvider->temporalCapabilities()->hasTemporalCapabilities() )
-      setStaticScalarDatasetIndex( QgsMeshDatasetIndex( 0, 0 ) );
+    setStaticScalarDatasetIndex( QgsMeshDatasetIndex( 0, 0 ) );
   }
   else
   {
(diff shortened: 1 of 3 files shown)


nyalldawson:
If the layer is showing no content when no temporal range is set, won't that also be an issue with layout maps (which default to no time range)

nirvn:
I'm not sure this is the right approach here, and I feel it walks us back towards the lack of an off mode.

Wouldn't it make sense to instead to work out a better behavior for temporal mesh lawyer when temporal navigation is off? Like showing the first frame instead of nothing?

vcloarec:
@nyalldawson  Indeed, I didn't think about the layout.

@nirvn : Indeed too ... much more there is only one line to remove in the mesh layer code...

Do we keep refactoring?

nirvn:
Good, much better approach IMHO. As for refactoring, I'm fine with it.

nirvn:
Btw, what happens if you provide a range that extends beyond the start and end of the mesh temporal range? It shows the last frame?

vcloarec:
It renders the first dataset included int the range, so for a time extent larger the time extent of the mesh, this would be the first dataset of the mesh.

nirvn:
Ok

vcloarec:
@nirvn about the answer at your question, I was wrong, if the start of the range is before the time extent of the mesh layer, nothing is rendered.
But after thinking about, I am now thinking that the rendered dataset is not well selected and this could be better (see https://github.com/qgis/QGIS/pull/36495#issuecomment-629903038)

vcloarec:
Finally the little change on the mesh side is done in the PR  #36495 .
So PR is now off topic... But I leave here the refactoring, I will close the PR if not wanted

PeterPetrik:
I am confused now @vcloarec . Should I review this or close?

PR #36495, opened 2020-05-17
[MESH] fix #36463 : display last dataset at the end of the time extent

### LINKED ISSUE #36463: Time controller for mesh 1D usability
1. Timestep of time slider should be equal or plural of the output time step, so not smaller
2. No data is given at the end of the simulation instead of the last output timestep.

vcloarec:
fix #36463 (point 2)
Depending of datasets, when the temporal  slider is at the end of the time extent, no datasets was displayed, and the last datasets could never be displayed.
This PR fix this issue.

### DIFF
--- a/src/app/mesh/qgsmeshlayerproperties.cpp
+++ b/src/app/mesh/qgsmeshlayerproperties.cpp
@@ -95,9 +95,9 @@ QgsMeshLayerProperties::QgsMeshLayerProperties( QgsMapLayer *lyr, QgsMapCanvas *
   delete mOptsPage_3DView;  // removes both the "3d view" list item and its page
 #endif
 
-  mComboBoxTemporalDatasetMatchingMethod->addItem( tr( "Find closest dataset before requested time" ),
+  mComboBoxTemporalDatasetMatchingMethod->addItem( tr( "Find Closest Dataset Before Requested Time" ),
       QgsMeshDataProviderTemporalCapabilities::FindClosestDatasetBeforeStartRangeTime );
-  mComboBoxTemporalDatasetMatchingMethod->addItem( tr( "Find closest dataset from requested time (after or before)" ),
+  mComboBoxTemporalDatasetMatchingMethod->addItem( tr( "Find Closest Dataset From Requested Time (After or Before)" ),
       QgsMeshDataProviderTemporalCapabilities::FindClosestDatasetFromStartRangeTime );
 
   // update based on lyr's current state
--- a/src/app/mesh/qgsmeshlayerproperties.cpp
+++ b/src/app/mesh/qgsmeshlayerproperties.cpp
@@ -95,6 +95,11 @@ QgsMeshLayerProperties::QgsMeshLayerProperties( QgsMapLayer *lyr, QgsMapCanvas *
   delete mOptsPage_3DView;  // removes both the "3d view" list item and its page
 #endif
 
+  mComboBoxTemporalDatasetMatchingMethod->addItem( tr( "Find closest dataset before requested time" ),
+      QgsMeshDataProviderTemporalCapabilities::FindClosestDatasetBeforeStartRangeTime );
+  mComboBoxTemporalDatasetMatchingMethod->addItem( tr( "Find closest dataset from requested time (after or before)" ),
+      QgsMeshDataProviderTemporalCapabilities::FindClosestDatasetFromStartRangeTime );
+
   // update based on lyr's current state
   syncToLayer();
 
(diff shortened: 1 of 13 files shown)


vcloarec:
About how the dataset has to be selected depending on the temporal controller range, after thinking about, I am not sure the actual implementation is satisfying. 
I see two other approaches than can be better. Both use only the start time of the temporal controller range, has pros/cons, but I don't know which one is the better:
- select the last dataset with time less than or equal to the start of the range.
pros : on canvas, dataset will change when the time passed the last dataset time
cons : if time of dataset have some few miliseconds or seconds more than the start of the range, the dataset will changed only for the next time step.
- select the closest dataset from the the start of the range.
pros : this is always the closest dataset that will be rendered;
cons : on canvas, dataset change will appears between to dataset time (on the middle) and not when time fo the last dataset will be passed.

vcloarec:
To avoid this issue and to have consistent behavior, it is better to let the user choose the matching method. So, this new approach let the user choose between the two matching methods described above.
image

src/app/mesh/qgsmeshlayerproperties.cpp
```diff
@@ -95,6 +95,11 @@ QgsMeshLayerProperties::QgsMeshLayerProperties( QgsMapLayer *lyr, QgsMapCanvas *
   delete mOptsPage_3DView;  // removes both the "3d view" list item and its page
 #endif
 
+  mComboBoxTemporalDatasetMatchingMethod->addItem( tr( "Find closest dataset before requested time" ),
```
nyalldawson, line 98: HIG convention is title case for combo box items (admittedly we aren't 100% on this, but are moving closer to it)

src/ui/mesh/qgsmeshlayerpropertiesbase.ui
```diff
@@ -671,26 +677,30 @@ border-radius: 2px;</string>
...
-            </spacer>
+            </widget>
            </item>
-           <item row="1" column="2">
-            <widget class="QLabel" name="label_6">
+           <item row="2" column="2">
+            <widget class="QLabel" name="label_10">
              <property name="text">
-              <string>Layer Start Time</string>
+              <string>Matching dataset Method</string>
```
PeterPetrik, line 697: Dataset Matching Method

src/core/mesh/qgsmeshdataprovidertemporalcapabilities.h
```diff
@@ -34,21 +34,37 @@ class CORE_EXPORT QgsMeshDataProviderTemporalCapabilities: public QgsDataProvide
 {
   public:
 
+    /**
+     * Method to use when requesting a tempral dataset from a time
```
PeterPetrik, line 38: typo: temporal dataset

PeterPetrik: I would clarify the docs, for example: Method for selection of mesh dataset index for time interval from temporal framework

src/core/mesh/qgsmeshdataprovidertemporalcapabilities.h
```diff
@@ -34,21 +34,37 @@ class CORE_EXPORT QgsMeshDataProviderTemporalCapabilities: public QgsDataProvide
...
+     **/
+    enum MatchingTemporalDatasetMethod
+    {
+      FindClosestDatasetBeforeStartRangeTime, //! Find the closest dataset which have its time before the requested start range time
```
PeterPetrik, line 42: docs: Finds

src/core/mesh/qgsmeshdataprovidertemporalcapabilities.h
```diff
@@ -34,21 +34,37 @@ class CORE_EXPORT QgsMeshDataProviderTemporalCapabilities: public QgsDataProvide
...
+      FindClosestDatasetFromStartRangeTime //! Find the closest dataset before or after the requested start range time
```
PeterPetrik, line 43: Finds

PeterPetrik:
the windows test failure seems unrelated, let me restart it

PeterPetrik:
i have absolutely no idea how that windows failure https://cdash.orfeo-toolbox.org/testDetails.php?test=19522666&build=42496 is related to this change 
@nyalldawson do you know if this  PyQgsLocalizedDataPathRegistry (Failed)  is some random failure/broken test?

PeterPetrik (approved):

nyalldawson:
It's unrelated

PR #36496, opened 2020-05-17
[eVis] Display the icons in the dialog

DelazJ:
To move from 
image
to 
image

I think that for consistency with other parts of QGIS (fwiw regarding eVis), we should instead use [mActionFileOpen](https://github.com/qgis/QGIS/blob/master/images/themes/default/mActionFileOpen.svg) to open file. Right?


nyalldawson:
> I think that for consistency with other parts of QGIS (fwiw regarding eVis), we should instead use mActionFileOpen to open file. Right?

Yes, that would be a good change to make.

On a related note... eww. We need to kill evis ASAP. We definitely should NOT be implementing another database connection level here.

PR #36499, opened 2020-05-17
Fix cppcheck warnings about uninitialized member variables

rouault:


src/providers/wcs/qgswcscapabilities.h
```diff
@@ -359,17 +359,17 @@ class QgsWcsCapabilities : public QObject
     /**
      * The error caption associated with the last WCS error.
      */
-    QString mErrorTitle;
+    QString mErrorTitle{};
```
nyalldawson, line 362: These look wrong -- there's no need to initialise QString members.

rouault: yes, this is a bit extra pedantism from cppcheck. It does warn about those members, because they are not explicitly set in the copy constructor (which is on purpose due to a human comment). The alternative would be to set them to the default value in the copy constructor

nyalldawson: > The alternative would be to set them to the default value in the copy constructor

I'd do that -- it's more reflective of the intended results

rouault: done

nyalldawson:
Thanks!

nyalldawson (approved):

PR #36505, opened 2020-05-18
Expose to public API the option to create separate files for each logical layer in a layout

nyalldawson:
The backend for this was previously used only when creating a GeoPDF export, where we create an individual PDF per layout logical layer which are then composited by GDAL, and was not exposed via the PyQGIS api.

By publicly exposing this functionality via the QgsLayoutExporter API we allow for plugins which can export layouts to seperate logical PDF files, allowing for other (non-GDAL) tools to be used to later
composite the layers (e.g. Adobe Illustrator (which stupidly cannot handle multi-layer single PDF files!))

Sponsored by SMEC/SJ

### DIFF
--- a/python/core/auto_generated/layout/qgslayoutexporter.sip.in
+++ b/python/core/auto_generated/layout/qgslayoutexporter.sip.in
@@ -198,6 +198,8 @@ Constructor for PdfExportSettings
 
       bool writeGeoPdf;
 
+      bool exportLayersAsSeperateFiles;
+
       bool useIso32000ExtensionFormatGeoreferencing;
 
       bool useOgcBestPracticeFormatGeoreferencing;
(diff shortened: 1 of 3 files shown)


nyalldawson:
For reference, the PyQGIS code used to use this flag looks like:

```
layout=QgsProject.instance().layoutManager().printLayouts()[0]
exporter = QgsLayoutExporter(layout)
settings =QgsLayoutExporter.PdfExportSettings()
settings.exportLayersAsSeperateFiles = True
result = exporter.exportToPdf('/home/nyall/Temporary/layers/layered.pdf', settings)
```

This will create `/home/nyall/Temporary/layers/layered_0001.pdf`, `/home/nyall/Temporary/layers/layered_0002.pdf` etc, which can then be stitched together in Illustrator via this script: https://community.adobe.com/t5/illustrator/illustrator-quot-place-quot-script-that-places-the-file-into-its-corresponding-layer-name/td-p/9788371?page=1

Issue #36465, opened 2020-05-15
Mesh Vectors cannot be switched off through clicking-on the arrow. 


PeterPetrik:
Vectors can only be switched off through unselecting “show vector”, so not by clicking-on the arrow.

PR #36506, opened 2020-05-18
disable  dataset from activedatasetwidget

vcloarec:
Enabling dataset is possible from the active dataset widget. But disabling the dataset is not possible from this same place. This PR fix this issue (#36465)


PeterPetrik (approved):

DelazJ:
Should be backported to LTR, right?

PeterPetrik:
i do not know. it has been like this since 1st mesh layer release, so it is more of the usability bug and not anything that you cannot live with. I would prefer not port things that are not really required.

PR #36509, opened 2020-05-18
Expose startup profile times in developer/debugging tools dock

nyalldawson:
This is part of an ongoing attempt to diagnose (and fix) occasional extremely long startup times of QGIS.

I've run into occasions where QGIS startup can take a couple of minutes. We have the great QgsRuntimeProfiler class which @NathanW2 introduced some years ago, so in an attempt to track down these startup delays I've extended the logging capabilities of this class and added more runtime profiles during the startup sequence.

I've also exposed these startup times in the developer/debugging tools dock, because previously the only way to see them was via the debug message log (which is difficult to access on Windows, especially for locked down enterprise machines)

Here's how this looks:

image

And here's a nice hotspot during startup we could try to target:

image


NathanW2:
Very nice!

PR #36519, opened 2020-05-18
Fix crash when masking widget is not initialized

troopa81:
## Description

Follows #36366

There is no masking widget for non spatial data.


nyalldawson:
Unrelated Travis guff

PR #36521, opened 2020-05-18
[Expression] Do not transform NULL integer as 0 when concat

### LINKED ISSUE #36112: Using concat(field1, field2) concatenates 0 instead of empty string (``''``) if any of the fields is NULL
**Describe the bug**

**How to Reproduce**
Using concat(field1, field2) concatenates ``0`` instead of empty string (``''``) if field2 is NULL.
The field2 is integer.

For a ``field1`` of text type having value ``test``, ``field2`` type integer and NULL (no value).
A label using ``concat(field1,field2)`` expression returns ``test0`` instead of the expected ``test`` (adds a ``0`` instead of an empty string)

I can expand the expression with an IF statement to solve it but.. it's a bug, right?

**QGIS and OS versions**

QGIS 3.10.5 Windows and Debian Bullseye

troopa81:
## Description

Fixes #36112

Providers returns QVariant ( QVariant::Int ) for null integer, which converts to "0" as a string. This PR proposes to not concat null values. 

The issue doesn't occured for OGR data (GPKG for instance) because this provider [returns a null QString](https://github.com/qgis/QGIS/blob/master/src/core/qgsogrutils.cpp#L318) for any null values while others provider don't ([spatialite](https://github.com/qgis/QGIS/blob/master/src/providers/spatialite/qgsspatialitefeatureiterator.cpp#L609), [postgres](https://github.com/qgis/QGIS/blob/master/src/providers/postgres/qgspostgresprovider.cpp#L4679), [delimitedtext](https://github.com/qgis/QGIS/blob/master/src/providers/delimitedtext/qgsdelimitedtextfeatureiterator.cpp#L476))

### DIFF
--- a/tests/src/core/testqgsexpression.cpp
+++ b/tests/src/core/testqgsexpression.cpp
@@ -3893,7 +3893,7 @@ class TestQgsExpression: public QObject
 
     void testConcatNULLAttributeValue()
     {
-      // Test that null integer values comming from provider are not transformed as 0
+      // Test that null integer values coming from provider are not transformed as 0
       // https://github.com/qgis/QGIS/issues/36112
 
       QgsFields fields;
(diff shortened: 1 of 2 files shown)


tests/src/core/testqgsexpression.cpp
```diff
@@ -3891,6 +3891,26 @@ class TestQgsExpression: public QObject
       QCOMPARE( QgsExpression::replaceExpressionText( input, &context ), expected );
     }
 
+    void testConcatNULLAttributeValue()
+    {
+      // Test that null integer values comming from provider are not transformed as 0
```
elpaso, line 3896:
```suggestion
      // Test that null integer values coming from provider are not transformed as 0
```

tests/src/core/testqgsexpression.cpp
```diff
@@ -3891,6 +3891,26 @@ class TestQgsExpression: public QObject
...
+      // https://github.com/qgis/QGIS/issues/36112
+
+      QgsFields fields;
+      fields.append( QgsField( QStringLiteral( "foo" ), QVariant::Int ) );
+
+      QgsFeature f;
+      f.initAttributes( 1 );
+      f.setAttribute( 0, QVariant( QVariant::Int ) );
```
elpaso, line 3904: can you also test `QVariant(  )` ? (it should give the same result)

elpaso (approved):

tests/src/core/testqgsexpression.cpp
troopa81: Yes, done

nyalldawson:
Travis is just being vindictive here -- he knows Github Actions is gunning for his job, so he's being petty

Issue #35671, opened 2020-04-08
Shape Digitizing 3-Point Rectangle Tool always creates 3D Polygon


ScottWerkema:
**Describe the bug**

When using the Shape Digitizing toolbar to create a 3-Point Rectangle (distance and projected), the resultant polygon is always 3D even on a 2D layer. 

**How to Reproduce**

1. Add a layer to a QGIS where the geometry type is a 2D polygon. In my case, this is a PostGIS layer. `geom geometry(Polygon,3857) NOT NULL`
Created using `SELECT AddGeometryColumn ('my_schema','my_spatial_table','geom',3857,'POLYGON',2);`
2. Use the "Add rectangle from 3 points" (either variant) to digitize a rectangle. 
3. Note the resultant rectangle is a 3D polygon and you are unable to save edits to this layer now.

Using other shape tools on the toolbar behave as expected. 

**Expected Result**
The 3-Point rectangle tool should create a 2D polygon for a 2D layer and a 3D polygon for a 3D layer.

**QGIS and OS versions**
macOS 10.15.4
QGIS
3.12.0
3.13 nightly (2020-04-07)

**Additional context**

This appears to be a regression in 3.12 as 3.10 and lower behave as expected.

gioman:
> This appears to be a regression in 3.12 as 3.10 and lower behave as expected.

I confirm the issue but I see the same problem on 3.10.4 I have not tested 3.4

gioman:
Working ok on 3.8, confirmed regression.

haubourg:
Cc @lbartoletti

Issue #35043, opened 2020-03-12
Can not create "Rectangle from 3 Points" in Scratch Layer or GPKG


thillyard:
**Describe the bug**
When attempting to "Create a rectangle from 3 Points" with either of the two methods given. The user is unable to commit changes to the layer. This seems to occur for GPKG, spatialite and Scratch layers, but works as expected with Shapefile Layers

**How to Reproduce**
1. Start with a blank map canvas add a scratch polygon layer
2. choose either of the "Add Rectangle by 3 Points options
![image](https://user-images.githubusercontent.com/44821733/76572225-69a8c280-6509-11ea-9720-2c0aa5a282de.png)
3. create a polygon
4. attempt to commit the edits
5. See error -->
![image](https://user-images.githubusercontent.com/44821733/76572085-fef78700-6508-11ea-8f6a-d5797ee167ce.png)

**QGIS and OS versions**

QGIS version
3.12.0-București
QGIS code revision
cd141490ec
Compiled against Qt
5.11.2
Running against Qt
5.11.2
Compiled against GDAL/OGR
3.0.4
Running against GDAL/OGR
3.0.4
Compiled against GEOS
3.8.0-CAPI-1.13.1
Running against GEOS
3.8.0-CAPI-1.13.1 
Compiled against SQLite
3.29.0
Running against SQLite
3.29.0
PostgreSQL Client Version
11.5
SpatiaLite Version
4.3.0
QWT Version
6.1.3
QScintilla2 Version
2.10.8
Compiled against PROJ
6.3.1
Running against PROJ
Rel. 6.3.1, February 10th, 2020
OS Version
Windows 10 (10.0)
Active python plugins
crayfish; 
DigitizingTools; 
go2streetview; 
icsm_ntv2_transformer; 
kmltools; 
LAStools; 
latlontools; 
mapswipetool_plugin; 
mmqgis; 
NNJoin; 
openlayers_plugin; 
processing_trajectory; 
qconsolidate3; 
qNote; 
SDEllipse; 
SRTM-Downloader; 
StreetView; 
timemanager; 
valuetool; 
db_manager; 
MetaSearch; 
processing

**Additional context**

gioman:
@thillyard if you create your layer as multipolygon then saving the feature created with that tool is ok. I'm not sure if this can be considered a bug or if it is by design.

lbartoletti:
[...]

PR #36523, opened 2020-05-18
Fix rectangle from 3 points

lbartoletti:
## Description

This PR will fix two bugs [1][2]. I don't remember why, it was assumed that the provider automatically applied the 2D/3D conversion.

[1] https://github.com/qgis/QGIS/issues/35043
[2] https://github.com/qgis/QGIS/issues/35671

### DIFF
--- a/src/app/qgsmaptooladdrectangle.cpp
+++ b/src/app/qgsmaptooladdrectangle.cpp
@@ -93,16 +93,25 @@ void QgsMapToolAddRectangle::deactivate( )
   }
 
   mParentTool->clearCurve( );
+
+  QgsVectorLayer *vLayer = static_cast<QgsVectorLayer *>( QgisApp::instance()->activeLayer() );
+  bool is3D = false;
+  if ( vLayer )
+    is3D = QgsWkbTypes::hasZ(vLayer->wkbType());
+
   // keep z value from the first snapped point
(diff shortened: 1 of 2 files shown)


src/app/qgsmaptooladdrectangle.cpp
```diff
@@ -93,16 +93,25 @@ void QgsMapToolAddRectangle::deactivate( )
...
+
+  QgsVectorLayer *vLayer = static_cast<QgsVectorLayer *>( QgisApp::instance()->activeLayer() );
+  bool is3D = false;
+  if ( vLayer )
+    is3D = QgsWkbTypes::hasZ(vLayer->wkbType());
+
   // keep z value from the first snapped point
-  std::unique_ptr<QgsLineString> lineString( mRectangle.toLineString() );
-  for ( const QgsPoint &point : qgis::as_const( mPoints ) )
+  std::unique_ptr<QgsLineString> lineString( mRectangle.toLineString( not is3D ) );
```
nyalldawson, line 103: this isn't python ;)

lbartoletti: Roooh :)
It's a valid C++ keyword :p

nyalldawson (approved):

PR #36524, opened 2020-05-18
Sensible default field precision for new fields

### LINKED ISSUE #16581: Sensible default width and precision for real columns
Author Name: **Nathanael Boehm** (Nathanael Boehm)
Original Redmine Issue: [7662](https://issues.qgis.org/issues/7662)

Redmine category:gui

---
In current dev 1.9 the default values for Width and Precision when adding a column to a table are 1 and 0 respectively for decimal/real type. It's unlikely that people would use real numbers and use this configuration, so I propose changing the defaults to perhaps Width=10 and Precision=5.

This was discovered when I was adding columns so I could manually move labels and wondered why my labels where disappearing when I saved the layer … to discover they had jumped to near Africa because my coordinates were being saved as 1,0 instead of 149.1178,-35.7456.

---

- [add_column.jpg](https://issues.qgis.org/attachments/download/5582/add_column.jpg) (Nathanael Boehm)

alexbruy:
## Description

Couple of UX fixes related to fields length and precision:
- don't pre-select default geometry type in the new SpatiaLite dialog (consistent with other new layer dialogs)
- field length and precision validators in the New Memory Layer dialog now respect native provider types
- sensible default field precision in the Add Field dialog (same as in Field Calculator, fixes #16581)

### DIFF
--- a/src/app/qgsnewspatialitelayerdialog.cpp
+++ b/src/app/qgsnewspatialitelayerdialog.cpp
@@ -50,15 +50,6 @@ QgsNewSpatialiteLayerDialog::QgsNewSpatialiteLayerDialog( QWidget *parent, Qt::W
   setupUi( this );
   QgsGui::enableAutoGeometryRestore( this );
 
-  connect( mAddAttributeButton, &QToolButton::clicked, this, &QgsNewSpatialiteLayerDialog::mAddAttributeButton_clicked );
-  connect( mRemoveAttributeButton, &QToolButton::clicked, this, &QgsNewSpatialiteLayerDialog::mRemoveAttributeButton_clicked );
-  connect( mGeometryTypeBox, static_cast<void ( QComboBox::* )( int )>( &QComboBox::currentIndexChanged ), this, &QgsNewSpatialiteLayerDialog::mGeometryTypeBox_currentIndexChanged );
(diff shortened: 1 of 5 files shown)


nyalldawson (approved):

nyalldawson:
Unrelated Travis nonsense

uclaros:
@nyalldawson @alexbruy I think the newmemorydialog stuff should be done just like in qgsaddattrdialog. Read the native types for the provider, then fill the ui accordingly.
That way if we want to add a new type tomorrow or correct a type we do it in one place ( in the provider ), and also it will be safer from mistakes like setting the mWidth for the decimal number to 30 but its validator to 20, like in this PR.

elpaso:
@alexbruy this commit causes a regression: try to add a new field to a PG layer, leave the default smallint type and save commits:

Could not commit changes to layer value

Errors: SUCCESS: 1 attribute(s) added.
  ERROR: field with index 6 is not the same!
  Provider: postgres
  Storage: PostgreSQL database with PostGIS extension
  expected field: name=ooo type=int typeName=int2 len=-1 precision=3
  retrieved field: name=ooo type=int typeName=int2 len=-1 precision=0

The field is actually created but you cannot leave the editing state.

alexbruy:
@elpaso should be fixed by #37697

PR #36526, opened 2020-05-18
Revert "[docker] Avoid assuming /dev/tty is usable"

nyalldawson:
This reverts commit ef0bfff02272fe461eab9a7719872859dcb0b4b4.

An attempt to see if this fixes Travis builds...


nyalldawson:
@strk looks like ef0bfff is definitely behind the recent travis breakage. can you find an alternate way around this? (Or maybe make the behaviour dependant on some environment variable so that we can leave the current approach for Travis builds?)

nyalldawson:
Hmmm.. maybe it's not this after all...!

strk:
I'm happy it was not.

PR #36527, opened 2020-05-18
Fix QgsGeometry::closestVertex() to return empty point

uclaros:
## Description
`QgsGeometry::closestVertex()` returns `QgsPointXY( 0, 0 )` instead of an empty point when there is no closest vertex. This was probably a leftover from before the introduction of `QgsPointXY::isEmpty`.
Also fixed a couple of docstrings with the same issue.


nyalldawson:
Thanks! Can we get a unit test too?

uclaros:
Sure! I should also change it to return an empty point if the parameter point is empty. Now it seems to return the closest point to 0,0

nyalldawson (approved):

PR #36532, opened 2020-05-19
Fix symbol layers rendering out of order when using Random Marker Fill

nyalldawson:
When rendering features, split the geometry fetching and preparation stage from the symbol layer rendering stage, and ensure that QgsSymbolLayer::startFeatureRender and ::stopFeatureRender
is correctly called in the right sequence when rendering multi-layer symbols

This fixes issues with symbol layers which rely on startFeatureRender and stopFeatureRender to correctly render, e.g. the Random Marker Fill symbol layer.

Before this fix, the logic looked like:

```
- for every symbol layer in the symbol, call startFeatureRender
- for each part in polygon, prepare the part geometry and then render each symbol layer
- for every symbol layer in the symbol, call stopFeatureRender
```

The issue with this approach is that symbol layers which defer rendering to the stopFeatureRender stage are always rendered after ALL other symbol layers in the symbol, regardless of the actual
order of the symbol layers. Ultimately this causes Random Marker Fill layers to always render on the top of symbols.

The new logic is:
```
- for each part in polygon, prepare the geometry and store the result
- for each symbol layer in the symbol:
   - call startFeatureRender
   - render the layer using each of the previously prepared parts
   - call stopFeatureRender
```

This results in correct stacking of the random marker fill in multi layer symbols, because the stopFeatureRender call is correctly called before the next layer's startFeatureRender and renderPolygon calls

Also, use QVector instead of QList for rings for improved efficiency


nyalldawson:
This one could use an in-depth review. @wonder-sk are you able to take a look?

nyalldawson:
Also refs https://github.com/qgis/QGIS/pull/35286

wonder-sk (approved):
Looks good to me!

nyalldawson:
Thanks @wonder-sk!

PR #36541, opened 2020-05-19
restore last used attribute table view when exiting multi-edit mode (fix #25099)

### LINKED ISSUE #25099: Turning Multi edit mode off does not go back to table view
Author Name: **Alexandre Neto** (@SrNetoChan)
Original Redmine Issue: [17200](https://issues.qgis.org/issues/17200)
Affected QGIS version: master

---
In the attribute table, If the user starts a multi edit mode session while in table view, when he turns it off, the attribute table will end up in Form view instead.

I think the right UX behaviour would be for the multi edit mode to remember the view mode the user was using before.

!anim.gif!

This affects both master and QGIS 2.18.

---

- [anim.gif](https://issues.qgis.org/attachments/download/11455/anim.gif) (Alexandre Neto)

alexbruy:
## Description
When attribute table is in table mode and user starts multi-edit session, turning multi-edit mode off does not return into table view, instead a Form view is used.

Fixes #25099.

### DIFF
--- a/src/gui/attributetable/qgsdualview.h
+++ b/src/gui/attributetable/qgsdualview.h
@@ -431,7 +431,7 @@ class GUI_EXPORT QgsDualView : public QStackedWidget, private Ui::QgsDualViewBas
     QgsFeature mTempAttributeFormFeature;
     QgsFeatureIds mLastFeatureSet;
     bool mBrowsingAutoPanScaleAllowed = true;
-    ViewMode mPreviousView;
+    ViewMode mPreviousView = AttributeTable;
 
     friend class TestQgsDualView;
     friend class TestQgsAttributeTable;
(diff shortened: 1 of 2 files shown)


src/gui/attributetable/qgsdualview.h
```diff
@@ -431,6 +431,7 @@ class GUI_EXPORT QgsDualView : public QStackedWidget, private Ui::QgsDualViewBas
     QgsFeature mTempAttributeFormFeature;
     QgsFeatureIds mLastFeatureSet;
     bool mBrowsingAutoPanScaleAllowed = true;
+    ViewMode mPreviousView;
```
nyalldawson, line 434: Can you initialise this value?

alexbruy: Done

PR #36563, opened 2020-05-19
Expressions to convert to and from base64

nyalldawson:
Adds the required documentation changes to #36235 so that we can merge for release


nyalldawson:
Unrelated test failure (Ping @3nids -- you broke the windows ci!)

PR #36570, opened 2020-05-20
Implement spatial index detection for OGR provider

### LINKED ISSUE #30530: Improve handling of missing spatial indexes in Processing
**Feature description.**
Many GIS algorithms heavily benefit from spatial indexes. If there is no spatial index, any operations that rely on spatial relationships between geometries will be unnecessarily slow.

For example a **Count points in polygon** comparison:
- 200,000 points in a Shapefile without spatial index: 30 seconds
- Same Shapefile but now with a .qix spatial index: 3 seconds

Processing should warn, like it nowadays does for CRS mismatches, if there is no spatial index for operations that would benefit from it.

Alternatively Processing could generate indexes internally but I don't know anything about how QGIS handles this for its internal layer structure. It would be a waste if an index was created every time an algorithm ran, maybe they could be stored in memory.

This could greatly benefit many users and lead to QGIS being faster in random comparisons *magically*.

**Additional context.**
I just realised, after years of QGIS usage, that QGIS does not write a spatial index when exporting shapefiles (see #30528). This means I wasted countless hours over the years waiting for algorithms to finish just because there was no spatial index being used. :o)

nyalldawson:
Provides warnings in Processing algorithms when running on layers which don't have a spatial index present (e.g. geojson files) and where performance will be severely hurt as a result

Fixes #30530

### DIFF
--- a/src/core/providers/ogr/qgsogrprovider.cpp
+++ b/src/core/providers/ogr/qgsogrprovider.cpp
@@ -3863,6 +3863,16 @@ QStringList QgsOgrProvider::uniqueStringsMatching( int index, const QString &sub
   return results;
 }
 
+QgsFeatureSource::SpatialIndexPresence QgsOgrProvider::hasSpatialIndex() const
+{
+  if ( mOgrLayer && mOgrLayer->TestCapability( OLCFastSpatialFilter ) )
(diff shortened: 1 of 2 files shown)


kannes:
Awesome, this will save thousands of hours of burning CPUs. Thank you!

PR #36573, opened 2020-05-20
Don't try to update raster layer data source's if they haven't changed

nyalldawson:
Aside from doing unnecessary work, I am seeing a crash in this code which is avoided by skipping the datasource reset.


nyalldawson:
@samweli do you see the same crash if you change a static time source?

Samweli:
> @Samweli do you see the same crash if you change a static time source?

@nyalldawson I can't reproduce on my side here, may be it is specific to the WMS service that you are using!

PR #36578, opened 2020-05-20
Fix WKT parser

lbartoletti:
## Description

This PR will fix WKT parser.  This should fix these errors:

```g = QgsGeometry.fromWkt('MULTILINESTRING((((1 2, 2 4, 4 5))')```
returns 
```<QgsGeometry: MultiLineString ((0 2, 2 4, 4 5))>```

and 
```
QgsGeometry.fromWkt('Point ((5, 5))')
```
returns
```
<QgsGeometry: Point (0 0)>
```
instead of null (or empty?) geometry

from #34496

### CODE REVIEW

**Comment by lbartoletti** [comment]
@elpaso I'm unsure about my fix on [this part](https://github.com/qgis/QGIS/pull/36578/commits/8d73fa7686fe6c275066436dc74adc35561f5d5e). Was that intended?

**Comment by nyalldawson** [comment]
Question: why do we want to allow these extremely wrong wkt strings?

**Comment by lbartoletti** [comment]
> Question: why do we want to allow these extremely wrong wkt strings?

I don't allow it. If it's wrong, will return a null geometry.

Today:
```
g = QgsGeometry.fromWkt('MULTILINESTRING((((1 2, 2 4, 4 5))')
```

returns
```
<QgsGeometry: MultiLineString ((0 2, 2 4, 4 5))>
```
with my pr, will return
```
<QgsGeometry: null>
```

Now I have to check that the extra parentheses

QgsGeometry.fromWkt('Point ((5, 5))') should returns null and not point(0,0)

**Comment by nyalldawson** [comment]
Ah -- i completely mis read this!!

**Comment by lbartoletti** [comment]
Here we go! All tests are fixed, there was a lot of tests with wrong WKT and also ... some wrong tests.
I also added new tests.

So, what's happened:

- some tests used the "wktype ()" form to create an "empty" wkt (eg `linestring ()`)
- a lot of wkt contained extra parenthesis eg `linestring (( 0 0, 1 1, 2 2 ))` or unbalanced ` polygon (( 0 0, 0 1 , 1 1, 1 0, 0 0 )) )`
- Our parsers doesn't tests if the coordinates are numeric values (including scientific number) and returns 0. `"POINT (a, 5)"`  returned "POINT (0, 5)". Same with extra parenthesis, `"LineString (( 1, 1, 2 2)"` returned `"LineString (0 1, 2 2)"`

With this PR, `fromWkt` methods will return False.

File: src/core/geometry/qgscircularstring.cpp [hunk 1]
```diff
@@ -317,7 +317,14 @@ bool QgsCircularString::fromWkt( const QString &wkt )
   if ( parts.second.compare( QLatin1String( "EMPTY" ), Qt::CaseInsensitive ) == 0 )
     return true;
 
-  setPoints( QgsGeometryUtils::pointsFromWKT( parts.second, is3D(), isMeasure() ) );
+  if ( parts.second.compare( QLatin1String( "NULL" ), Qt::CaseInsensitive ) == 0 )
+    return false;
+
+  QgsPointSequence points = QgsGeometryUtils::pointsFromWKT( parts.second, is3D(), isMeasure() );
+  if ( points.isEmpty() )
```
nyalldawson, line 324: Unless I'm reading this wrong, this would prevent creation of geometries from `CircularString ()`. We should be forgiving here and treat things like `CircularString ()` as `CircularString( EMPTY )` (same with NULL). 

The QGIS WKT parser was designed to be as forgiving as possible, rather then strictly follow the standards. It should consume any standard WKT string, but if we can intrepret a malformed string then we need to allow this.

lbartoletti: yes with my PR `CircularString ()` is unallowed you have to write `circularstring EMPTY` (as on postgis).

The only "non standard" is `LINESTRING Z ...` instead of `LINESTRINGZ` (as on postgis too).

nyalldawson: Can you revert that change? It was an intentional choice to be forgiving...

lbartoletti: For me, a malformed string sould not be allowed. But `CircularString ( )` could be an exception, and can return an empty geometry.

nyalldawson: Why would we want to be restrictive though? If you get a dataset with malformed wkt, would you prefer to be able to import directly in qgis or spend ages mucking around and manually fixing it? We should definitely be as permissive as possible on import, and as strict as possible on output...

lbartoletti: Personally, I prefer to have a strict behavior indicating an error in the formatting of my string than having a tool assuming a result for me (especially when that result is wrong, like today.)

But I agree with you that it's not pleasant to have to manually fix strings that have errors.

So, what can we do to fix this parser?

We've already agreed on those (even if I haven't pushed my commit yet):
LineString () => LineString EMPTY
LineString ( EMPTY ) = > LineString EMPTY

But what about this WKT
Point (( 9, 9)) => POINT (9 9) or NULL ?
LINESTRING (( 1 1, 22, 3 3) => ?
LINESTRING (( 1 1, 2 2, 3 3)) => ?

Thanks.

nyalldawson: > Personally, I prefer to have a strict behavior indicating an error in the formatting of my string than having a tool assuming a result for me (especially when that result is wrong, like today.)

Yeah, but you're coming from a developer's point of view :laughing:  End users don't care about this -- they just want the data imported with as little messing around as possible!

> Point (( 9, 9)) => POINT (9 9) or NULL

I'd say POINT (9 9) -- the intention is clear here

> LINESTRING (( 1 1, 22, 3 3)

This one should error -- it's completely unclear what the intention was

> LINESTRING (( 1 1, 2 2, 3 3)) 

We should just ignore the extraneous brackets, and treat it as 
LINESTRING ( 1 1, 2 2, 3 3)

lbartoletti: > Yeah, but you're coming from a developer's point of view laughing End users don't care about this -- they just want the data imported with as little messing around as possible!

I was (and I'm still) an user before being a developper :wink: but I've always liked when software returns results as strict as possible -- maybe I'm crazy :laughing:  

Whatever I've no objection to do these enhancements. I'm just confused since I'll have not the time to do this before the release. 

As you want, you merge it now [1] and I add modifications next days or we postpone this bugfix and it will backported.

[1] ATM, except "geometry()" which returns an empty geometry and which I constrained, others return false geometry "Point( (1 1)" returns "Point( 0 1)

nyalldawson: We definitely should defer -- if we merge in the current strict approach then it's quite likely to break some existing workflows for users

lbartoletti:
Dear Stale Bot,

I finish my PR next week.

Thanks.

tests/src/core/testqgsexpression.cpp
```diff
@@ -1218,26 +1218,18 @@ class TestQgsExpression: public QObject
       QTest::newRow( "is_multipart null" ) << "is_multipart(NULL)" << false << QVariant();
       QTest::newRow( "z_max no 3D" ) << "z_max(geom_from_wkt('POINT (0 0)'))" << false << QVariant();
       QTest::newRow( "z_max NULL" ) << "z_max(geom_from_wkt(NULL))" << false << QVariant();
-      QTest::newRow( "z_max point Z NaN" ) << "z_max(geom_from_wkt('PointZ (1 1 nan)'))" << false << QVariant( QVariant::Double );
```
nyalldawson, line 1221: Can you restore these tests?

lbartoletti: Yes, but it will not return a double since NaN in a WKT is a mistake. It will returns a NULL (QVariant::Invalid)

PR #36587, opened 2020-05-20
Vertex tool, fix escape to stop dragging edge

uclaros:
## Description
In vertex tool, dragging a vertex can be stopped by pressing Escape, but dragging an edge cannot. This PR fixes that, so escape can stop both dragging a vertex or an edge.


nyalldawson (approved):

nyalldawson (changes requested):
Actually there's additional code paths which need updating here now - eg the non-escape key presses should only apply when mDraggingEdge is false

uclaros:
I restructured it into a switch statement, the code paths are less cryptic now.
Also fixes the case where one could `highlightAdjacentVertex()` while dragging vertices.
`highlightAdjacentVertex()` handles the case of `mSelectedVertices.isEmpty()` that's why I omitted it.

nyalldawson (approved):

PR #36604, opened 2020-05-21
Drop Globe plugin

espinafre:
Dropping Globe plugin from the QGIS sources. It depends on osgearth no
newer than version 2.8, released on 2016-09-05, and which is impossible
to build against modern (Qt 5.13+) versions of Qt. With this, the
optional QGIS dependency on OpenSceneGraph and osgearth is also removed.

I've asked the qgis-developers mailing list, and some of the original authors agreed with the change: 

https://lists.osgeo.org/pipermail/qgis-developer/2020-May/061298.html

There, @wonder-sk pointed out that most of its functionality is provided by the 3D view.

This removes all references to the Globe plugin, openscenegraph and osgearth, including CMake macros and options.

I believe this shoult NOT be backported to LTR, as there might be someone still using it (but it is disabled by default on all platform configurations).


nyalldawson:
+1

If we do it now it gives users who require the missing "globe" mode in the native 3d framework time to find this last feature gap before the next ltr.

nirvn:
Seems we're all in agreement here. Who wants to push the button?

PR #36607, opened 2020-05-21
Drop the evis plugin

nyalldawson:
It's unmaintained, and duplicates functionality which is exposed elsewhere in QGIS

See https://github.com/qgis/QGIS-Enhancement-Proposals/issues/176 and https://lists.osgeo.org/pipermail/qgis-developer/2020-May/061269.html


DelazJ:
Label added in order to get it removed from docs

PR #36618, opened 2020-05-21
[processing] fix handling for stdout and file outputs in GRASS

alexbruy:
## Description
Found this when working on another issue in Processing GRASS provider.

"Fake" (redirected from stdout) outputs in GRASS algorithms does not handled correctly if destination is temporary file. Algorithms just fail with Python error saying "File 'TEMPORARY_OUTPUT' does not exists". Normal file outputs works almost fine, except that in the log instead of valid output path we got also `TEMPORARY_OUTPUT` literal.

Proposed PR fixes this by maintaining mapping between output name and destination path. This mapping then used to produce results map and also for converting text outputs into HTML.

### DIFF
--- a/python/plugins/processing/algs/grass7/Grass7Algorithm.py
+++ b/python/plugins/processing/algs/grass7/Grass7Algorithm.py
@@ -116,6 +116,7 @@ class Grass7Algorithm(QgsProcessingAlgorithm):
         self.commands = []
         self.outputCommands = []
         self.exportedLayers = {}
+        self.fileOutputs = {}
         self.descriptionFile = descriptionfile
 
         # Default GRASS parameters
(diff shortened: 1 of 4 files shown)


alexbruy:
@nyalldawson may I ask you to take a look?

nyalldawson:
I've got to admit, I don't fully understand the logic here (the original logic & the new logic :smile: )!

But I do know that since 3.0 there's been a number of disabled grass tests which relied on html outputs. I wonder if these relate to the same underlying cause and could now be re-enabled?

alexbruy:
Probably I have missed when we disabled some GRASS tests but there are some broken algs and I'm about to submit another PR which fixes them and adds tests. Will check if there are any disabled tests and try to re-enable them.

PR #36625, opened 2020-05-21
[Backport release-3_12] Fix issues when content cache is used in blocking mode

rldhont:
## Description

Manually backport #35158

Fixes numerous issues when the content cache is used in blocking mode, ultimately resulting in null entries being stored in the cache for remote requests


elpaso (approved):

nyalldawson:
3.12 is eoled

Issue #35888, opened 2020-04-21
[Mesh] issue when using trace styling with color ramps


saberraz:
When I try traces using color ramp, it works fine to start with. But eventually (moving the time slider a couple of times), QGIS fails to render the image and the CPU usage jumps to 100%

The only thing I found on the backtrace:
```
Warning: QPainter::begin: Paint device returned engine == 0, type: 3
[Detaching after fork from child process 4010]
Stacktrace (piped through c++filt):
/usr/local/src/QGIS_master/build/output/bin/qgis(+0x117e2)[0x5555555657e2]
/usr/local/src/QGIS_master/build/output/bin/qgis(+0x11f45)[0x555555565f45]
/usr/lib/x86_64-linux-gnu/libQt5Core.so.5(+0xb63b7)[0x7ffff2cc73b7]
/usr/lib/x86_64-linux-gnu/libQt5Core.so.5(+0xb6569)[0x7ffff2cc7569]
/usr/lib/x86_64-linux-gnu/libQt5Core.so.5(QMessageLogger::warning(char const*, ...) const+0xba)[0x7ffff2c962ee]
/usr/lib/x86_64-linux-gnu/libQt5Gui.so.5(QPainter::begin(QPaintDevice*)+0x587)[0x7ffff35112c7]
/usr/local/src/QGIS_master/build/output/lib/libqgis_core.so.3.13.0(+0x123be6e)[0x7ffff524ae6e]
/usr/local/src/QGIS_master/build/output/lib/libqgis_core.so.3.13.0(+0x123d8e5)[0x7ffff524c8e5]
/usr/local/src/QGIS_master/build/output/lib/libqgis_core.so.3.13.0(+0x1239616)[0x7ffff5248616]
/usr/local/src/QGIS_master/build/output/lib/libqgis_core.so.3.13.0(+0x123f435)[0x7ffff524e435]
/usr/local/src/QGIS_master/build/output/lib/libqgis_core.so.3.13.0(+0x1230dc0)[0x7ffff523fdc0]
/usr/local/src/QGIS_master/build/output/lib/libqgis_core.so.3.13.0(+0x121c9b9)[0x7ffff522b9b9]
/usr/local/src/QGIS_master/build/output/lib/libqgis_core.so.3.13.0(+0x121a03c)[0x7ffff522903c]
[...]

PR #36633, opened 2020-05-22
[MESH] fix streamlines and traces renderer when vector is null

vcloarec:
In mesh rendering, the streamlines and the trace renderers didn't stop when vector magnitude was null.
This PR fix this issue

### DIFF
--- a/src/core/mesh/qgsmeshlayerrenderer.cpp
+++ b/src/core/mesh/qgsmeshlayerrenderer.cpp
@@ -550,6 +550,9 @@ void QgsMeshLayerRenderer::renderVectorDataset()
   if ( std::isnan( mVectorDatasetMagMinimum ) || std::isnan( mVectorDatasetMagMaximum ) )
     return; // only NODATA values
 
+  if ( !( mVectorDatasetMagMaximum > 0 ) )
+    return; //all vector are null vector
+
   std::unique_ptr<QgsMeshVectorRenderer> renderer( QgsMeshVectorRenderer::makeVectorRenderer(
         mTriangularMesh,
         mVectorDatasetValues,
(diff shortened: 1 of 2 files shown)


src/core/mesh/qgsmeshlayerrenderer.cpp
```diff
@@ -550,6 +550,9 @@ void QgsMeshLayerRenderer::renderVectorDataset()
   if ( std::isnan( mVectorDatasetMagMinimum ) || std::isnan( mVectorDatasetMagMaximum ) )
     return; // only NODATA values
 
+  if ( !( mVectorDatasetMagMaximum > 0 ) )
```
PeterPetrik, line 553: `mVectorDatasetMagMaximum <= 0`

src/core/mesh/qgsmeshtracerenderer.cpp
```diff
@@ -429,6 +430,9 @@ void QgsMeshStreamField::addTrace( QPoint startPixel )
   if ( !mVectorValueInterpolator )
     return;
 
+  if ( !( mMaximumMagnitude > 0 ) )
```
PeterPetrik, line 433: mVectorDatasetMagMaximum <= 0

PeterPetrik (commented):
it fixes which ticket? i think there is missing issue # in the description

src/core/mesh/qgsmeshlayerrenderer.cpp
vcloarec: Indeed, I hesitated, but I found more explicit >0

src/core/mesh/qgsmeshtracerenderer.cpp
vcloarec: I found this issue without ticket. Maybe fix #35888, but very not sure.

vcloarec:
I found this issue without ticket during temporal tries. Maybe fix #35888, but very not sure.

PeterPetrik:
related #35888

PeterPetrik:
Merged the ticket, I have

PR #36640, opened 2020-05-22
Tweak temporal playback behavior

nyalldawson:
If a non-looping animation is paused at the end and the user hits the play
button again, auto rewind and play the animation from the start


nyalldawson:
@nirvn as discussed, a nice UX tweak which improves the feel of playback

nirvn (approved):
Ahh, yes please.

PR #36650, opened 2020-05-22
[ui] Make temporal controller's forward, backward and pause buttons behave as animation state toggle.

nirvn:
This PR improves the animation UI/UX of the temporal controller panel by having the play forward, play backward, and pause buttons act as toggles. I.e., if you hit play while the temporal animation is playing, it'll pause the animation, and vice versa. 

It's quite useful to be able to quickly toggle the animation state (via keyboard or mouse).

The PR also makes a space bar key press event starts / pauses the animation state when the user has the slider focused. This means users can pause the animation, move through temporal frames via the arrow left and right keys, and resume the animation easily:

Peek 2020-05-22 16-14

### DIFF
--- a/python/gui/auto_generated/qgstemporalcontrollerwidget.sip.in
+++ b/python/gui/auto_generated/qgstemporalcontrollerwidget.sip.in
@@ -37,6 +37,11 @@ The dock widget retains ownership of the returned object.
 %End
 
 
+  protected:
+
+    virtual void keyPressEvent( QKeyEvent *e );
+
+
 };
 
 /************************************************************************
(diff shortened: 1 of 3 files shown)


nyalldawson (approved):
We should consider adding some global shortcuts for play/pause too. @anitagraser I believe Time Manager is using super+space for this?

anitagraser:
@nyalldawson  There's Ctrl+Space but nothing else afaik:

     self.focusSC = QShortcut(QKeySequence("Ctrl+Space"), self.dock)


https://github.com/anitagraser/TimeManager/blob/9cf7858a1180b41029000edcf327e359e3e71e2a/timemanagerguicontrol.py#L130

PR #36624, opened 2020-05-21
[Backport release-3_10] Fix issues when content cache is used in blocking mode

rldhont:
## Description

Manually backport #35158

Fixes numerous issues when the content cache is used in blocking mode, ultimately resulting in null entries being stored in the cache for remote requests

PR #36651, opened 2020-05-22
[Backport release-3_10] Fix Task manager waitforfinished

rldhont:
## Description

Manually backport #32838 needed for #36624

It could lead to crashes (If you access some shared memory data believing the task has finished to run)

### DIFF
--- a/src/core/qgstaskmanager.cpp
+++ b/src/core/qgstaskmanager.cpp
@@ -29,7 +29,9 @@
 QgsTask::QgsTask( const QString &name, Flags flags )
   : mFlags( flags )
   , mDescription( name )
+  , mNotStartedMutex( 1 )
 {
+  mNotStartedMutex.acquire();
 }
 
 QgsTask::~QgsTask()
@@ -42,6 +44,7 @@ QgsTask::~QgsTask()
     delete subTask.task;
   }
   mNotFinishedMutex.unlock();
+  mNotStartedMutex.release();
 }
 
 void QgsTask::setDescription( const QString &description )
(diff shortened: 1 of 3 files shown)


troopa81:
Just for the record, there is other race conditions issues fixed in here  #35200. But the one you backport could be enough to fix your issue.

I decided on purpose to not backport those because it was a lot of critical modifications in LTR. So far, no one has ever complained in master and 3.12 but I'm still unsure about backporting it. @nyalldawson any opinion on this

rldhont:
Thanks @troopa81 for your comments.

@elpaso can I request your point of you about this backport needed to fix an issue in QGIS Server #36624 ?

nyalldawson:
I'm -0 to backporting these, simply because the consequences of a regression are extreme. 

If we do decide to backport them, then you'll need to backport ALL the task manager fixes, as one of the follow up commits may also be required.

rldhont:
Thanks @nyalldawson for your feedback.

If I cannot merge this backport, I will re-introduced and fix an event loop to wait for task starting in `QgsAbstractContentCache`.

elpaso:
@rldhont I trust @nyalldawson 's opinion on this one, I've been bitten by the event loop recently and the side effects of an event loop are really dangerous but on the other hand we've lived with that for quite a long time.

The choice between backporting all the qgstask patches or reintroduce the event loop is ultimately up to you, I'm sorry but I can't tell which one is the less risky here.

rldhont:
If I can't backport only this bugfix, I don't want to take the risk to backport all the QgsTask if @nyalldawson think it's risky.

I have reintroduced the event loop in https://github.com/qgis/QGIS/pull/36624. If I can merge #36624 with the event loop I will close this one.

nyalldawson:
I'm -1 to introducing an event loop here. That's an extremely dangerous change, totally unsuitable for LTR.

rldhont:
@nyalldawson I have not introduced a new event loop in LTR. I have just fix an already existing event loop in LTR.

rldhont:
If no reject #36651, I will merge it and close this one

PR #36657, opened 2020-05-22
Apply proxy settings both in ogr and gdal providers

mhugent:
Currently, the OGR provider reads the QGIS proxy settings and applies it to GDAL. However the GDAL provider does not yet do it.
This PR moves the function which sets the GDAL proxy to QgsGdalUtils and calls it from both OGR and GDAL providers.

### DIFF
--- a/src/core/qgsgdalutils.cpp
+++ b/src/core/qgsgdalutils.cpp
@@ -15,12 +15,15 @@
 
 #include "qgsgdalutils.h"
 #include "qgslogger.h"
+#include "qgsnetworkaccessmanager.h"
+#include "qgssettings.h"
 
 #define CPL_SUPRESS_CPLUSPLUS  //#spellok
 #include "gdal.h"
 #include "gdalwarper.h"
 #include "cpl_string.h"
 
+#include <QNetworkProxy>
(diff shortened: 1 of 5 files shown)


src/core/qgsgdalutils.cpp
```diff
@@ -486,3 +489,53 @@ GDALDatasetH QgsGdalUtils::rpcAwareAutoCreateWarpedVrt(
...
+  // instead of adding an API and complex signal/slot connections
+  // given the limited cost of checking them on every provider instantiation
+  // we can do it here so that new settings are applied whenever a new layer
+  // is created.
+  QgsSettings settings;
+  // Check that proxy is enabled
+  if ( settings.value( QStringLiteral( "proxy/proxyEnabled" ), false ).toBool() )
+  {
+    // Get the first configured proxy
+    QList<QNetworkProxy> proxyes( QgsNetworkAccessManager::instance()->proxyFactory()->queryProxy( ) );
```
nyalldawson, line 506: Proxies

nyalldawson (approved):

rouault (approved):

PR #36660, opened 2020-05-22
Support OGR GPKG in datasource uri connection

### LINKED ISSUE #36525: GPKG transaction groups are stored with empty second pair of the key
in other words, the connection string is always an empty string and there is no way to use transaction groups when you have more than one GPKG in the same project.

That happens because `QgsDataSourceUri( layer.source() ).connectionInfo()` on a GPKG returns an empty string.

The sad reality is that QgsDataSourceUri is not appropriate anymore to handle all kind of data provider connections and URI encodings that we have.

elpaso:
Makes QgsDataSourceUri::connectionInfo usable with
GPKG and OGR/Spatialite file-based DBs.

With tests.

Fixes #36525


elpaso:
This approach won't work. Looking for another one.

PR #36668, opened 2020-05-23
[temporal] Make the temporal controller's set range button more useable

nirvn:
## Description

OK, this PR is the last temporal controller UI/UX corner I felt wasn't working super well: the set range to project "refresh" button. 

This PR improves the experience by splitting the "full [project layers] range" vs. the "preset project range" and offering those through a tool button drop down menu:
image
_Ignore the auto raised temporal mode buttons, this was just a test._

### DIFF
--- a/src/gui/qgstemporalcontrollerwidget.cpp
+++ b/src/gui/qgstemporalcontrollerwidget.cpp
@@ -15,13 +15,18 @@
  *                                                                         *
  ***************************************************************************/
 
+#include "qgsapplication.h"
 #include "qgstemporalcontrollerwidget.h"
 #include "qgsgui.h"
+#include "qgsmaplayermodel.h"
 #include "qgsproject.h"
 #include "qgsprojecttimesettings.h"
 #include "qgstemporalmapsettingswidget.h"
 #include "qgstemporalutils.h"
 #include "qgsmaplayertemporalproperties.h"
+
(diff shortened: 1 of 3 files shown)


src/gui/qgstemporalcontrollerwidget.cpp
```diff
@@ -145,6 +167,38 @@ void QgsTemporalControllerWidget::keyPressEvent( QKeyEvent *e )
...
+  if ( QgsProject::instance()->timeSettings() )
+    projectRange = QgsProject::instance()->timeSettings()->temporalRange();
+  mRangeSetToProjectAction->setEnabled( projectRange.begin().isValid() && projectRange.end().isValid() );
+
+  mRangeLayersSubMenu->clear();
+  for ( int i = 0; i < mMapLayerModel->rowCount(); ++i )
+  {
+    QModelIndex index = mMapLayerModel->index( i, 0 );
+    QgsMapLayer *currentLayer = mMapLayerModel->data( index, QgsMapLayerModel::LayerRole ).value<QgsMapLayer *>();
+    if ( !currentLayer->temporalProperties() || !currentLayer->temporalProperties()->isActive() )
```
nyalldawson, line 182: We should probably add a filter for temporal layers to the map layer proxy model at some stage

nyalldawson (approved):

PR #36673, opened 2020-05-23
Remove scrollbar frame stroke and add help button to labels propertie…

DelazJ:
…s dialog
labelsProps


nirvn:
@DelazJ , this triggers a question (not directly related to your change): can the dialog's height be resized smaller than the minimum height required and show scrollbars? If not, we need to fix that, this dialog is getting quite tall.

DelazJ:
@nirvn you mean if we can get this?
image

nirvn:
@DelazJ , yep, thanks for confirming we can :)

nirvn (approved):

PR #4842, opened 2017-07-12
changes to QgsDateTimeEdit to support the minimal Date supported by Q…

mj10777:
…Changes to QgsDateTimeEdit

## Description
Changes to QgsDateTimeEdit to support the minimal Date supported by QDateTimeEdit (0100-01-01)

## Checklist

> Reviewing is a process done by project maintainers, mostly on a volunteer basis. We try to keep the overhead as small as possible and appreciate if you help us to do so by completing the following items. Feel free to ask in a comment if you have troubles with any of them.

- [x] Commit messages are descriptive and explain the rationale for changes
- [ ] Commits which fix bugs include `fixes #11111` in the commit message next to the description
- [ ] Commits which add new features are tagged with `[FEATURE]` in the commit message
- [ ] Commits which change the UI or existing user workflows are tagged with `[needs-docs]` in the commit message and containt sufficient information in the commit message to be documented
- [ x] I have read the [QGIS Coding Standards](https://docs.qgis.org/testing/en/docs/developers_guide/codingstandards.html) and this PR complies with them
- [ x] This PR passes all existing unit tests (test results will be reported by travis-ci after opening this PR)
- [ ] New unit tests have been added for core changes
- [ x] I have run [the `scripts/prepare-commit.sh` script](https://github.com/qgis/QGIS/blob/master/.github/CONTRIBUTE.md#contributing-to-qgis) before each commit

The default setting for QDateTimeEdit for the MinimalDate is '1752-09-14', based on the Act of Parliament of 1751 (Calendar Act) - which is a disaster for any Historical Database for any event before 1752-09-14.

The MinimalDate supported by QDateTimeEdit is: '0100-01-01' (as defined in  qdatetimeparser_p.h)
* both for the functions setMinimumDateTime and setMinimumDate
   * even though the QT-Documention implies at any valid QDateTime can be used with setMinimumDateTime

Both function have existed since QT 4.4, so this could be backported to the QGIS 2.

Since QgsDateTimeEdit is derived from QDateTimeEdit, this pull request adds a function that
* sets a date that both QDateTimeEdit supports and can be formatted as Qt::ISODate
   * and is called during the construction of QgsDateTimeEdit (called setMinimumEditDateTime())

This value can be reset to the default value ('1752-09-14') by calling 
* QDateTimeEdit::clearMinimumDateTime().

[...]

PR #36683, opened 2020-05-23
Allow QgsDateTimeEdit to support dates since Jan 1st 1 CE (instead of 100 CE) and avoid QDate and QDateTime field value triple conversion

### LINKED ISSUE #35574: QGIS 3.12 - Date Field does not go earlier than 1/1/100 AD in geopackage or shape file
QGIS3.12.0 (Mac Mojave 10.14.6, iMac - QGIS Website Download) - I am mapping radiometric dates from archaeological sites that date from 60000BC to 1950AD, but it appears that I can not set a date using the date field type in a geopackage or shapefile earlier than 1/1/100 AD.

I have searched the previous bugs, but did not find a similar issue. Note attached image shows the 'click earlier year' triangle button as being greyed out.

image

agiudiceandrea:
## Description

With PR #4842 the lower limit for values that can be entered in a field using the Date/Time widget was set to Jan 1st 100 CE (previously it was not possible to enter dates prior to Sep 14th 1752 CE).

The date of Jan 1st 100 CE was used only because QDateTimeEdit::setMinimumDateTime() doesn't support dates earlier than that.
https://github.com/qgis/QGIS/blob/d73eaa1d58ca6e7429d338ef42bcd16dcae6742a/src/gui/editorwidgets/qgsdatetimeedit.h#L196-L204

This limitation doesn't affect QDateTimeEdit::setDateTimeRange(), so it is possible to use it to allow entering dates also from 1 CE to 99 CE which previously where not allowed***.
```
    void setMinimumEditDateTime()
    {
      setDateTimeRange( QDateTime( QDate( 1, 1, 1 ) ), maximumDateTime() );
    }
```

Fixes #35574 (as far as possible).

This PR also avoids a triple conversion of QDate and QDateTime field value in QgsDateTimeFieldFormatter::representValue():
- QVariant-[toString]->QString
- QString-[fromString(format)]->QDateTime
- QDateTime-[toString(format)]->QString

while 
- QVariant-(toDateTime)->QDateTime
- QDateTime-[toString(format()]->QString

is enough for QDate and QDateTime fields.

#    
***It could be possible to go further and allow dates with negative years since they are supported by setDateTimeRange() and by QDate and QDateTime type fields for memory layers, but it seems they are not allowed in Date/Time field by some providers (e.g. Shapefile and GeoPackage).

Video_2020-05-23_234612

Dates with negative years could also be stored in text fields but they are not properly handled by the Date/Time widget until [QTBUG-84334](https://bugreports.qt.io/browse/QTBUG-84334) is fixed (QDateTime::toString() handles negative years while QDateTime::fromString() cannot parse them).

### DIFF
--- a/src/core/fieldformatter/qgsdatetimefieldformatter.cpp
+++ b/src/core/fieldformatter/qgsdatetimefieldformatter.cpp
@@ -51,13 +51,20 @@ QString QgsDateTimeFieldFormatter::representValue( QgsVectorLayer *layer, int fi
   const QString displayFormat = config.value( QStringLiteral( "display_format" ), defaultFormat( field.type() ) ).toString();
 
   QDateTime date;
-  if ( fieldIsoFormat )
+  if ( static_cast<QMetaType::Type>( value.type() ) == QMetaType::QDate || static_cast<QMetaType::Type>( value.type() ) == QMetaType::QDateTime )
   {
-    date = QDateTime::fromString( value.toString(), Qt::ISODate );
+    date = value.toDateTime();
   }
   else
   {
-    date = QDateTime::fromString( value.toString(), fieldFormat );
+    if ( fieldIsoFormat )
+    {
+      date = QDateTime::fromString( value.toString(), Qt::ISODate );
+    }
+    else
+    {
+      date = QDateTime::fromString( value.toString(), fieldFormat );
+    }
   }
 
   if ( date.isValid() )
--- a/src/gui/editorwidgets/qgsdatetimeedit.h
+++ b/src/gui/editorwidgets/qgsdatetimeedit.h
@@ -188,19 +188,17 @@ class GUI_EXPORT QgsDateTimeEdit : public QDateTimeEdit
     void resetBeforeChange( int delta );
 
     /**
-     * Set the lowest Date that can be displayed with the Qt::ISODate format
-     *  - uses QDateTimeEdit::setMinimumDateTime (since Qt 4.4)
+     * Set the lowest Date that can be stored in a Shapefile or Geopackage Date field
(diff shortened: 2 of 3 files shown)


nyalldawson:
Nice! Can you add some tests for this too?

agiudiceandrea:
Hi @nyalldawson, I don't know if the added test is adequate. I think there is nothing special to test on dates between Jan 1st CE and Dec 31th 99 CE as they are not special...

What do you think about the possibility to allow also dates with negative year, although they can be used in memory layers, but not in Shapefiles or GeoPackages, for example?

nyalldawson:
The test looks good to me!

nyalldawson:
> What do you think about the possibility to allow also dates with negative year, although they can be used in memory layers, but not in Shapefiles or GeoPackages, for example?

Yes, I think we should support these. Ideally we'd do add a new QgsVectorDataProvider capability flag for "Supports negative year values", and correctly return this flag for memory/other (postgis?) providers. Then you'd restrict the date range for the widget based on whether the associated layer's provider supports negative years or not.

agiudiceandrea:
Thanks. I shall have a try.
In the meanwhile I've found a bug #36715 about the Date/Time edit widget (it allows to store text into a numeric field) whose cause I can't find for now...

nyalldawson (approved):

agiudiceandrea:
The QT bug https://bugreports.qt.io/browse/QTBUG-84334 that prevented to correctly parse negative years has been fixed in Qt6 (https://codereview.qt-project.org/c/qt/qtbase/+/315029, https://github.com/qt/qtbase/commit/91140eb7268bc36efceef0459e7b124ce333829e).

PR #36685, opened 2020-05-24
Move georeferencer from plugins to app

nyalldawson:
Georefencer shouldn't be a plugin. It's core functionality which is of widespread need. Move it from plugins -> app.

FAQ
-----

**Won't this make QGIS startup slower?**

No, it'll actually make it faster. :laughing:


nirvn (approved):
Nice cleanup and micro-optimization all in one.

Issue #33555, opened 2019-12-30
Coordinate capture: update coordinates when CRS to use for coordinate display is changed


JBosm:
Hello,
when changing the CRS to use for coordinate display, the coordinates in the display box are not corrected. It would seem better if they were updated to the new CRS.
![image](https://user-images.githubusercontent.com/5849171/71579829-8d5d8e80-2afe-11ea-900d-bf1485c9e703.png)

gioman:
@JBosm how this is different from https://github.com/qgis/QGIS/issues/33554 ?

JBosm:
@gioman The two tickets are different, I'm sorry if I was not clear enough (english is not my mother tongue, feel free to ask again if still not clear). 
This one, 33555 is that when changing CRS for coordinate capture, the coordinates in the display are not immediately updated, so you have to click once again (on the map) to get new coordinates, this time with the right CRS. But when clicking, the coordinates will never be exactly the same, so updating when changing the CRS would seem cleaner and more logical to me. 
#33554  is about loosing the chosen CRS when quitting QGIS. On opening QGIS, the CRS for coordinate capture will always be 4326, even if changed earlier. Not sure if the CRS for coordinate capture should be saved in the project or in QGIS preferences.

nyalldawson:
No longer an issue with the merged "copy coordinates" approach

Issue #33554, opened 2019-12-30
Coordinate capture: save CRS to use for coordinate display


JBosm:
Hello,
when closing and re-opening QGIS, the CRS of the first coordinates is always 4326, even when changed. Would it be possible/desirable to save the last CRS used for coordinates display?
![image](https://user-images.githubusercontent.com/5849171/71579719-20e28f80-2afe-11ea-8232-9faf4821c17d.png)

nyalldawson:
No longer an issue with the merged "copy coordinates" approach

Issue #27712, opened 2018-09-19
Coordinate Capture does not use datum transformation


qgib:
Author Name: **Bernhard Ströbl** (Bernhard Ströbl)
Original Redmine Issue: [19888](https://issues.qgis.org/issues/19888)
Affected QGIS version: 3.3(master)
Redmine category:c++_plugins

---
if a datum transformation is defined (with a transformation grid file) to transform from the map CRS to the coordinate capture's CRS, this transformation is not applied, nor is the user asked if or which transformation should be applied.

nyalldawson:
No longer an issue with the merged "copy coordinates" approach

PR #36693, opened 2020-05-25
Move 'coordinate capture' functionality to map canvas

nyalldawson:
Instead of requiring users to enable the core c++ plugin 'Coordinate Capture'
in order to copy coordinates for a map point, instead add this functionality
to a map canvas right click context menu.

Now, when compatible map tools are activated (pan, zoom, select by rect), right clicking
on the canvas shows a context menu with a "Copy Coordinate" submenu. The
submenu shows options for copying the coordinate in the map CRS, WGS84 or
a custom preset CRS.

Also adds API to allow 3rd party QgsMapTool subclasses to implement their own
context menus which include this Copy Coordinate action.

Rationale
------------

Since 3.14 is removing the evis and globe plugins, and moving the georeferencer away from a plugin, I think it's sensible to "rip off the band-aid in one go" and also move the Coordinate Capture functionality away from a plugin. In any case, this is very important functionality which should be available out-of-the-box without having to manually enable a plugin.


nirvn:
+1 to ripping off the band aid in one go
+1,000 to insure that the basic functionality of copying a given coordinate under the mouse pointer is there out of the box.

nyalldawson:
Fixes #33555, fixes #33554, fixes #27712

uclaros:
Sorry for being a party-pooper here, but we lost functionality with this.
- Lost the ability to switch to previous tool with right click: use case was while using for example a digitizing tool, select pan or zoom tool, use once, right click to continue digitizing.
- Lost the ability to monitor pointer position simultaneously in wgs84 and in local projected crs.
- Lost visual feedback of the point clicked (no red dot and mouse must be moved for coordinates to actually be copied).

Other objections:
- The Pan tool hand mouse pointer has no obvious pointing edge so is a bad choice for a coordinate picker.
- Right clicking **on the map canvas** is generally considered a *done* button and not a *context menu* button. Only exception is the *Identify tool*.

I think a separate tool would be more appropriate for this functionality; click to copy coordinates and leave a marker until tool is changed, right click for context menu.

3nids:
What is the scope of this new context menu?
Would it make sense to integrate the is identification menu?

Regarding @uclaros remark, could the option of switching back to the previous tool be integrated in the menu too?

nyalldawson:
@uclaros 

My intention was to provide a streamlined, out-of-the-box alternative for the most common use of the plugin, and then for a community driven Python plugin to step up and deliver an advanced version of the original functionality.

So these features:

> Lost the ability to monitor pointer position simultaneously in wgs84 and in local projected crs.
> Lost visual feedback of the point clicked (no red dot and mouse must be moved for coordinates to actually be copied).

would belong in the community plugin.


> Lost the ability to switch to previous tool with right click: use case was while using for example a digitizing tool, select pan or zoom tool, use once, right click to continue digitizing.

I'll have to look into that one 

> Right clicking on the map canvas is generally considered a done button and not a context menu button. Only exception is the Identify tool.

I disagree -- I think that's the case for digitizing tools only, but for other tools it's nice to expose a context menu instead of doing nothing. User expectation is definitely that a right click on something shows a context menu in the absence of any other dedicated functionality...


@3nids

> Would it make sense to integrate the is identification menu?

Yes - I would have liked to do that also, but the code for the identification menu handling would need quite a lot of rework to support this.

CMKMS-LINZ:
I installed 3.14. After getting frustrated trying to find where the Coordinate Capture plugin had gone, I've found my way here.

I agree with all of @uclaros ' comments, especially that using the hand tool to get a coordinate is imprecise - the icon has no edge, I have no idea where the coordinate is, also because the plugin left a red dot at the spot clicked. The new right click function built into 3.14 doesn't.

 This loss of one of functionality from a most fundamental plugins in QGIS is very unhelpful.

uclaros:
I have ported the plug-in to python and plan to upload it to the plugins repo. My thought was to set 3.14 as the minimum required version and name the plug-in Coordinate Capture, just like the core one. That way most users needing it will hardly notice it is no longer a core plugin, and will easily find and install it. Would that be ok @nyalldawson?

nyalldawson:
@uclaros nice work, that sounds like a perfect approach!

NelsonMinar:
For any other users looking for this function: install the Coordinate Capture plugin (version 0.2 or greater). Then select the hand tool and right click on your map; a popup for "Copy Coordinate" should appear. At least right now, it sounds like this functionality may change again.

src/gui/qgsmapcanvas.cpp
```diff
@@ -795,6 +799,116 @@ void QgsMapCanvas::clearTemporalCache()
...
+    }
+  } );
+  copyCoordinateMenu->addAction( setCustomCrsAction );
+
+  mMenu->addMenu( copyCoordinateMenu );
+
+  if ( mMapTool )
+    mMapTool->populateContextMenu( mMenu );
+
+  mMenu->exec( event->globalPos() );
```
vcloarec, line 909: @nyalldawson I wonder if you see any inconvenient to make this QMenu alive only it this scope? It could be convenient to embed QObject with this QMenu, and to delete this object after QMenu::exec().

The difficulty is maybe to bring the menu in the lambda...

Issue #34352, opened 2020-02-07
Label masking doesn't work with line pattern symbol sub layers


nirvn:
While testing out the label masking feature some more today, I stumble on an implementation issue, whereas line pattern symbol's sub layer aren't masked.

Test project here:
[mask.zip](https://github.com/qgis/QGIS/files/4170252/mask.zip)

You can see here the hashed lines are checked to be masked but
![Screenshot from 2020-02-07 17-35-47](https://user-images.githubusercontent.com/1728657/74022527-4d78a980-49d0-11ea-94c1-aa92c3f60129.png)

nirvn:
@mhugo , gentle ping. It'd be great if you could look into this issue before masking is officially released as part of 3.12. Cheers

mhugo:
@nirvn Thanks for your feedback. However, I am struggling to find some available time to work on it before 3.12 ...

nirvn:
@mhugo , OK, it's a shame though, it means QGIS will ship masking without sub-symbol support (hashed lines, point pattern, geometry generator, etc.). Nobody can force you to find that hour before now and release date to insure the implementation is complete, but I'll keep hoping you find time until then.

mhugo:
After some time spent on a debugging session, it seems to come from `QgsLinePatternFillSymbolLayer::applyPattern`
https://github.com/qgis/QGIS/blob/master/src/core/symbology/qgsfillsymbollayer.cpp#L2517
where the fill line sub symbol is cloned for some reasons.

The symbol layer pointers that have to be disabled in the second pass and are stored in the rendering context do not correspond anymore ...
So I have to find a more robust way to disable symbol layers during the second pass ...

Not easy :-(

Pedro-Murteira:
Still valid on QGIS 3.22.3.

thymaro:
I don't understand _how_, but I understand _that_ it is complicated to fix. So, this is not a rant, just another sitrep:

I can confirm this is still valid for QGIS 3.22.6. Line pattern fills don't respect masks.
![grafik](https://user-images.githubusercontent.com/24420636/165026816-863f040f-02c6-44d4-8fd5-f87e57b80598.png)

PR #36697, opened 2020-05-25
Fix selective masking composition stage 

### LINKED ISSUE #34650: Selective masking: phantom lines on dark background
**Describe the bug**

Selective masking generates lines or "halos" around the masked area. They are visible on dark backgrounds.

Some examples:

image

image

image

When turning off anti-aliasing, the problem disappears, but this is not a solution.
Something is wrong in the way images are composed and leads to this rendering artifacts

### LINKED ISSUE #34947: Selective masking: issue with mask opacity
**Describe the bug**

Playing with the opacity of selective masking sometimes give strange results.

Initially reported by @nirvn https://github.com/qgis/QGIS/pull/33432#issuecomment-566422039 and on Twitter : https://twitter.com/klaskarlsson/status/1231518713804992512

image

image

**How to Reproduce**

I fail to reproduce the error on my system. But since two people reported it, there might be something ...

**QGIS and OS versions**

3.12

obrix:
At the composition stage of the second pass introduced with selective masking, the unmasked elements located under the mask are copied into a temporary image buffer and then re added later to the final image.

This copy was done using the masks from selective masking with their full opacity informations. Unmasked informations were thus combined with opacity from the mask leading to the introduction of several artefacts : 
* The "phantom lines" #34650 (opacity information introduced by the antialising on mask drawing)
* Opacity applied on non masked surfaces in #34947

To fix this issue a temporary mask is created for the copy of unmasked pixels, in this temporary mask the alpha channel is "binarized" (fully opaque on masking pixel, fully transparent on non masking pixel).

Should fix #34650 and fix #34947

Here are two examples of two issues fixed by this PR

Phantom lines : 
phantom_lines

Phantom lines fixed : 
phantom_lines_fix

Opacity applied on yellow line that is not masked : 
opacity_issue

Opacity fixed (applied only on masked black line) : 
opacity_issue_fix


nyalldawson (commented):
This is going to be sloooow. Could the original image be constructed using the QImage::Format_Mono to avoid this requirement?

nirvn:
@obrix , great to see somehow attempting to fix this issue :+1:

obrix:
@nyalldawson Not sure about Format_Mono, I have to look how this would interact with QPainter CompositionMode. I made sure to avoid using setPixel to build my image but there is probably a faster way to do this, I'm not an expert on QImage. @mhugo Do you have any idea ?

@nirvn Thanks : )

mhugo:
@nyalldawson 's suggestion seems good: it would be better to already construct a 1-bit mask rather than trying to convert it afterwards.

obrix:
@nyalldawson @mhugo Masks with an alpha channel must be kept to handle the transparency features of selective masking correctly. So this means duplicating all masks (LayerRenderJob::maskImage and LabelRenderJob::maskImages) with 1-bit versions of the same image. Duplicating painter draw calls too. It introduces some more complexity to the code and I am not sure the performances will be better this way (but maybe I am wrong) ?

After some tests it seems converting the final mask to 1-bpp mask using QImage::createAlphaMask instead of iterating on the pixel is way better performance wise. Just need to find a way to compose this 1-bpp mask in the qpainter to get the same result as the current "CompositionMode_DestinationIn" composition.

nyalldawson:
@obrix another option could be calling QImage::convertToFormat on the mask image and convert it to mono using the optimised qt format conversion method instead of the handcoded version.

On another note -- please tell me your TODO list includes refactoring masking to make it vector - export friendly! The forced rasterisation is such a huge limitation which prevents this feature being used for any serious work.

obrix:
@nyalldawson I implemented a faster method to compute my mask, this should be better performance wise. I also slightly modified the composition again to fix some issues I had with the composition of effects. The tests are now good.
Concerning the refactoring to make the feature vector friendly, as you saw this is the subject of a recent QEP by @mhugo  that I may realize !

nyalldawson:
> Concerning the refactoring to make the feature vector friendly, as you saw this is the subject of a recent QEP by @mhugo that I may realize !

Yep, saw that :+1: 

If you're redesigning how this feature works, I think you should also consider https://github.com/qgis/QGIS/issues/34352 -- that one also points to an architectural limitation in the current approach...

andreasneumann:
> If you're redesigning how this feature works, I think you should also consider #34352 -- that one also points to an architectural limitation in the current approach...

@nyalldawson - yes this is also on the radar and I asked the original funders to finance solving this issue. Slowly we should be getting there ;-) selective masking is a complex new feature and it turned out that it needs additional work. Thank you also for helping us getting there!

src/core/qgsmaprendererjob.cpp
```diff
@@ -817,7 +817,17 @@ void QgsMapRendererJob::composeSecondPass( LayerRenderJobs &secondPassJobs, Labe
       // Only retain parts of the second rendering that are "inside" the mask image
       QPainter *painter = job.context.painter();
       painter->setCompositionMode( QPainter::CompositionMode_DestinationIn );
-      painter->drawImage( 0, 0, *maskImage );
+
+      //Create an "alpha binarized" image of the maskImage to :
+      //* Eliminate antialiasing artefact
+      //* Avoid applying mask opacity to elements under the mask but not masked
+      QImage maskBinAlpha = maskImage->createMaskFromColor( 0 );
+      maskBinAlpha.setColorCount( 0 );
```
nyalldawson, line 825: Is this necessary? Seems like it would be redundant given the call to setColorTable below...

nyalldawson:
Looking good!

src/core/qgsmaprendererjob.cpp
obrix: I had some issues at one point with color count being incremented but you are right it seems it is now useless. I removed it.

nyalldawson (approved):

PR #36707, opened 2020-05-25
[FEATURE] Statistical aggregates: no more null strings as empty, add datetime stats, add mode statistic

### LINKED ISSUE #27230: [Aggregates] Allow majority and minority statistics on non-numeric fields

### LINKED ISSUE #35442: Mode(s) for Statistical Summary
**Feature description.**
I think an addition to the QGIS Statistical Summary tool to show the Mode(s) of the dataset would be quite helpful to give a summary of which values are occurring most often in the dataset. This would be in addition to the current Majority value which reports only the first? mode.

**Additional context**
As discussed in this PR:  
https://github.com/qgis/QGIS/pull/35434#issuecomment-605662607

suricactus:
First, apologies as this PR grew a bit too big. It is addressing several issues in the statistical aggregates.

Most importantly, it contains a bugfix for #35433 String aggregate-related functions treat NULL and empty strings interchangeably.

There are number of features that are also included:
- calculate "Mode" statistic for any datatype FIX https://github.com/qgis/QGIS/issues/35442
- calculate First and Last for `String` datatype
- calculate Mean, Median, StDev, StDevSample,  Minority, Majority, FirstQuartile, ThirdQuartile, InterQuartileRange, First and Last for `DateTime` and `Time` datatypes FIX https://github.com/qgis/QGIS/issues/27230

### Dev notes:
- The calculation of Mode for numeric columns is a bit hacky. I believe there should be a breaking change in QGIS 4:
https://github.com/qgis/QGIS/compare/master...suricactus:aggr_nulls?expand=1#diff-22430bd6ea504a657383ef4a1acd0ceaR275

```diff
- double QgsStatisticalSummary::statistic( QgsStatisticalSummary::Statistic stat ) const
+ QVariant QgsStatisticalSummary::statistic( QgsStatisticalSummary::Statistic stat ) const
```

This way one may calculate statistics, that do not produce a value, representable with double (e.g. the mode statistic).

- there is little risk that `signed long long` will overflow when calculating the `QgsDateTimeStatisticalSummary` statistics if someone abuses the datetime fields. I assume that all values will be relatively close to each other, so I use the first value as time reference and work with datetime deltas of the rest of the values, relative to the first value. I hope noone is storing geological date times :).


suricactus:
No stale. I have to fix some of the tests, later these days.

m-kuhn:
Hey stale-bot, don't you want to unstale?

Issue #36398, opened 2020-05-13
QGIS Server: WFS Request does not use SrsName on the geometry


ismailsunni:
**Describe the bug**

I wanted to do GetFeature request using WITHIN. This is the request:
```
<GetFeature xmlns="http://www.opengis.net/wfs" service="WFS" version="1.1.0" outputFormat="GML2" maxFeatures="1000" xsi:schemaLocation="http://www.opengis.net/wfs http://schemas.opengis.net/wfs/1.1.0/wfs.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <Query typeName="mosque" srsName="EPSG:4326" xmlns:feature="http://www.qgis.org/gml">
        <Filter xmlns="http://www.opengis.net/ogc">
            <Within>
                <PropertyName>geometry</PropertyName>
                <Polygon xmlns="http://www.opengis.net/gml" srsName="EPSG:4326">
                    <exterior>
                        <LinearRing srsName="EPSG:4326">
                            <posList srsDimension="2">
                        		110 -8
                        		111 -8
                        		111 -7
                        		110 -7
                        		110 -8
                        		</posList>
                        </LinearRing>
                    </exterior>
                </Polygon>
            </Within>
        </Filter>
    </Query>
</GetFeature>
```

it works fine.

But, when I changed the srsName to 21781 (I reproject the points also), like this:

```
<GetFeature xmlns="http://www.opengis.net/wfs" service="WFS" version="1.1.0" outputFormat="GML2" maxFeatures="1000" xsi:schemaLocation="http://www.opengis.net/wfs http://schemas.opengis.net/wfs/1.1.0/wfs.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <Query typeName="mosque" srsName="EPSG:21781" xmlns:feature="http://www.qgis.org/gml">
        <Filter xmlns="http://www.opengis.net/ogc">
            <Within>
                <PropertyName>geometry</PropertyName>
                <Polygon xmlns="http://www.opengis.net/gml" srsName="EPSG:21781">
                    <exterior>
                        <LinearRing srsName="EPSG:21781">
                            <posList srsDimension="2">532546 157825 532546 153725 541046 153725 541046 157825 532546 157825</posList>
                        </LinearRing>
                    </exterior>
                </Polygon>
            </Within>
        </Filter>
    </Query>
</GetFeature>
```

It doesn't give me the same result. It gives me:
```
[...]

PR #36713, opened 2020-05-25
Add unit test for WFS Post GetFeature Within

ismailsunni:
## Description

Unit test to replicate #36398

I am adding these 3 unit tests:
1. WFS GetFeature Within with coordinate in EPSG:4326 and srsName EPSG:4326 (passed)
1. WFS GetFeature Within with coordinate in EPSG:4326 and srsName EPSG:3857 (failed, it should return different response compared to the unit test 1)
1. WFS GetFeature Within with coordinate in EPSG:3857 and srsName EPSG:3857 (failed, failed, it should return the same response as the unit test 1)

cc @elpaso


tests/src/python/test_qgsserver_wfs.py
```diff
@@ -189,7 +189,7 @@ def test_wfs_getcapabilities_100_url(self):
                 self.assertEqual(
                     "onlineResource=\"my_wfs_advertised_url\"" in item, True)
 
-    def result_compare(self, file_name, error_msg_header, header, body):
+    def result_compare(self, file_name, error_msg_header, header, body, expectEqual=True):
```
elpaso, line 192: I don't really understand the `expectEqual` here, what is it for?

ismailsunni: I want to test if two requests with the same coordinate but different EPSG return different (unit test 1 and 2 from the description). But we don't have an `assertXMLNotEqual`, so I added `expectEqual` to the result compare. By default, it will pass if the XML is equal. If it is set to False, the XML shouldn't be the same.

It's only used for this one `withinFake3857FilterTemplate_post`

elpaso: Cannot use assertRaise for testing an expected failure?

ismailsunni: Ah, yes, I think it's a cleaner way to do it. Let me update it.

nyalldawson:
@ismailsunni can you mark the new tests as having expected failures? That way we can merge this...

elpaso:
unrelated test failure

PR #36718, opened 2020-05-25
Approved pending 3.10.8 backports

nyalldawson:
Collated set of approved 3.10 backports, which are in the pending state until 3.10.8.

(designed to cleanup the many "Not for merge" requests in the PR queue)


haubourg:
@nyalldawson If I may, the commit https://github.com/qgis/QGIS/pull/36718/commits/175e4bfeb91fe316e06d0b3cb5835b4a4d5512c9 solves an issue which blocks a massive deployement. May I ask why the backport can't make it into 3.0.7 when there still remains 14 days? I fail to get the point of blocking those backports. Do you need some reviewing force?

haubourg:
>  I fail to get the point of blocking those backports

Sorry, I meant postponing, not blocking

nyalldawson:
@haubourg 
>  If I may, the commit 175e4bf solves an issue which blocks a massive deployement. May I ask why the backport can't make it into 3.0.7 when there still remains 14 days? I fail to get the point of blocking those backports. Do you need some reviewing force?

Because that's a very high risk commit! It
- uses locale settings, which have a history of revealing unwanted side effects in corner cases
- affects an absolutely top priority component of QGIS -- we simply **cannot** risk having a regression in the scale selector in the middle of an LTR release

Given this, and the fact that it's not a new issue at all (it's an issue which afaics has been around since the dawn of QGIS) I think it's entirely acceptable to delay this fix so that it has a month wider testing & feedback period after landing in the stable (non-LTR) release.

To put it totally bluntly: we need to put avoiding the risk of regressions for **all** our existing LTR users over the needs of one enterprise!

haubourg:
> Because that's a very high risk commit! It
> 
>     * uses locale settings, which have a history of revealing unwanted side effects in corner cases
> 
>     * affects an absolutely top priority component of QGIS -- we simply **cannot** risk having a regression in the scale selector in the middle of an LTR release
> 
> 
> Given this, and the fact that it's not a new issue at all (it's an issue which afaics has been around since the dawn of QGIS) I think it's entirely acceptable to delay this fix so that it has a month wider testing & feedback period after landing in the stable (non-LTR) release.
> 
> To put it totally bluntly: we need to put avoiding the risk of regressions for **all** our existing LTR users over the needs of one enterprise!

Ok thanks for the inputs. I just have no such a vision of how risky a backport is without your insights.  

I think I also forgot  that we rely on release version to widden the user test and consolidate the fix.  My apologies if that sounded rude or misplaced.  I'll stress our customer to jump on 3.14 to check intensively those features.

ponceta:
Thanks @nyalldawson !

PR #36724, opened 2020-05-26
avoid 3D crash if terrain layer is null

vcloarec:
On 3D map settings, when no layer is chosen for terrain (mesh or DEM), QGIS can crash when the settings are applied.
This PR fix this issue.

### DIFF
--- a/src/3d/chunks/qgschunkedentity_p.cpp
+++ b/src/3d/chunks/qgschunkedentity_p.cpp
@@ -402,8 +402,7 @@ void QgsChunkedEntity::startJobs()
     delete entry;
 
     QgsChunkQueueJob *job = startJob( node );
-    if ( job )
-      mActiveJobs.append( job );
+    mActiveJobs.append( job );
   }
 }
 
@@ -415,8 +414,6 @@ QgsChunkQueueJob *QgsChunkedEntity::startJob( QgsChunkNode *node )
     QgsEventTracing::addEvent( QgsEventTracing::AsyncBegin, QStringLiteral( "3D" ), QStringLiteral( "Load " ) + node->tileId().text(), node->tileId().text() );
 
     QgsChunkLoader *loader = mChunkLoaderFactory->createChunkLoader( node );
-    if ( !loader )
-      return nullptr;
     connect( loader, &QgsChunkQueueJob::finished, this, &QgsChunkedEntity::onActiveJobFinished );
     node->setLoading( loader );
     return loader;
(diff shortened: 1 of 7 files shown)


src/3d/chunks/qgschunkedentity_p.cpp
```diff
@@ -411,6 +415,8 @@ QgsChunkQueueJob *QgsChunkedEntity::startJob( QgsChunkNode *node )
     QgsEventTracing::addEvent( QgsEventTracing::AsyncBegin, QStringLiteral( "3D" ), QStringLiteral( "Load " ) + node->tileId().text(), node->tileId().text() );
 
     QgsChunkLoader *loader = mChunkLoaderFactory->createChunkLoader( node );
+    if ( !loader )
+      return nullptr;
```
wonder-sk, line 419: Is this addition necessary?

The returned loader must be valid - otherwise things can break elsewhere (we can get to invalid state in the state machine and loading queue). If I understand correctly, with your other additions, startJob() should not even get called when the entity is not valid, so this check should not be needed...

src/3d/chunks/qgschunkedentity_p.cpp
```diff
@@ -399,7 +402,8 @@ void QgsChunkedEntity::startJobs()
     delete entry;
 
     QgsChunkQueueJob *job = startJob( node );
-    mActiveJobs.append( job );
+    if ( job )
+      mActiveJobs.append( job );
```
wonder-sk, line 406: Is this addition necessary?

There never should be null jobs coming from startJob()... We should make sure elsewhere that we do not create jobs for any entities that are not valid...

vcloarec: > Is this addition necessary?

I am not very sure they are necessary, if the but I was thinking that those guards could not hurt, and help easily avoid crash. I like guards for that!
A case where a job can be null it is when the mesh layer is null here : https://github.com/qgis/QGIS/blob/9aafc08bbabdd670adaf8e7704021987c677dbc2/src/3d/mesh/qgsmeshterraingenerator.cpp#L58
That could happen if the mesh layer in the mesh terrain generator is not valid anymore (removed by user for example). But, as the unique (for now) tile terrain entity is created directly with mesh layer information in the main thread (even if implicitly shared triangular mesh is used in other thread), there should not be problem...
So I remove those extra guards and cross finger that it always will be ok!

wonder-sk: Thanks... In the mesh terrain generator we could have `assert( mLayer )` so that we never return null pointer and make it clear that the function should not return null pointer - and that there should be code that ensures the function is not called with invalid terrain generator...

wonder-sk (approved):

PR #36736, opened 2020-05-26
quote project file path when passing it to the crash handler (fix #25532)

### LINKED ISSUE #25532: "Reloading QGIS" from the crash report dialog fails to reopen the project
Original Redmine Issue: [17635](https://issues.qgis.org/issues/17635)
Affected QGIS version: 3.1(master)
Redmine category:project_loading/saving
Assignee: Nathan Woodrow

---
I'm working on a project stored on the network and i get a crash
I click on the "Reload QGIS" button which is supposed to relaunch QGIS and reopen the project. Btw, this label is not really obvious that it would reopen the project. Maybe something like "Reload Project" instead?
Anyway... QGIS is relaunched but it fails to reopen the project as he tries to find it in the c:\osgeo4w\bin folder.
And in case of a path with spaces, it takes pieces of text for layers, returning "The layer is invalid: the layer xxx is invalid and could not be added to the project" in the messagebar.

---

- [ReloadFails.PNG](https://issues.qgis.org/attachments/download/11814/ReloadFails.PNG) (Harrissou Santanna)

alexbruy:
## Description
If project saved in the path with spaces and QGIS crashed, crash handler fails to reload project as unquoted path with spaces considered as multiple command-line arguments.

Fixes #25532.

### DIFF
--- a/src/app/qgscrashhandler.cpp
+++ b/src/app/qgscrashhandler.cpp
@@ -62,7 +62,7 @@ LONG WINAPI QgsCrashHandler::handle( LPEXCEPTION_POINTERS exception )
   QString projectFile = QgsProject::instance()->fileName();
   if ( !projectFile.isEmpty() )
     // quote project file path to avoid issues if it has spaces
-    arguments << '"' << projectFile << '"';
+    arguments << QStringLiteral( "\"%1\"" ).arg( projectFile );
   else
     arguments << projectFile;
 
--- a/src/app/qgscrashhandler.cpp
+++ b/src/app/qgscrashhandler.cpp
@@ -59,7 +59,12 @@ LONG WINAPI QgsCrashHandler::handle( LPEXCEPTION_POINTERS exception )
   arguments = QCoreApplication::arguments();
   // TODO In future this needs to be moved out into a "session state" file because we can't trust this is valid in
   // a crash.
-  arguments << QgsProject::instance()->fileName();
+  QString projectFile = QgsProject::instance()->fileName();
+  if ( !projectFile.isEmpty() )
+    // quote project file path to avoid issues if it has spaces
+    arguments << '"' << projectFile << '"';
+  else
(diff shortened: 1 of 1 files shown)


src/app/qgscrashhandler.cpp
```diff
@@ -59,7 +59,12 @@ LONG WINAPI QgsCrashHandler::handle( LPEXCEPTION_POINTERS exception )
   arguments = QCoreApplication::arguments();
   // TODO In future this needs to be moved out into a "session state" file because we can't trust this is valid in
   // a crash.
-  arguments << QgsProject::instance()->fileName();
+  QString projectFile = QgsProject::instance()->fileName();
+  if ( !projectFile.isEmpty() )
+    // quote project file path to avoid issues if it has spaces
+    arguments << '"' << projectFile << '"';
```
nyalldawson, line 65:
```suggestion
    arguments << QStringLiteral( "\"%1\"" ).arg( QgsProject::instance()->fileName() );

otherwise it's creating multiple entries in the list instead of a single string...

File: src/app/qgscrashhandler.cpp [hunk 2]
```diff
@@ -59,7 +59,12 @@ LONG WINAPI QgsCrashHandler::handle( LPEXCEPTION_POINTERS exception )
...
+  else
+    arguments << projectFile;
```
nyalldawson, line 67: shouldn't this be removed? it's appending an empty string?

alexbruy: Oh, my bad. Fixed.

src/app/qgscrashhandler.cpp
alexbruy: Oh, my bad. Fixed.

src/app/qgscrashhandler.cpp
alexbruy: Fixed.

Issue #30210, opened 2019-06-14
QGIS crashes when adding file to geopackage -> layer -> field(BLOB)


APOS80:
QGIS crach when:
1. Open geopackage layer
2. In field for binary obect(BLOB), add file.

Tested adding  Excel, Png, Jpg

[stacktrace.txt](https://github.com/qgis/QGIS/files/3290204/stacktrace.txt)

rouault:
I've reproduced this issue. The stack trace looks weird :
```
(gdb) bt
#0  0x00007ffff1fd40e0 in operator==(QString const&, QString const&) () from /opt/qt59/lib/libQt5Core.so.5
#1  0x00007ffff2e6f857 in QLabel::setText(QString const&) () from /opt/qt59/lib/libQt5Widgets.so.5
#2  0x00007ffff5e0d348 in QgsBinaryWidgetWrapper::updateValues (this=0x5154270, value=...) at /home/even/qgis/QGIS/src/gui/editorwidgets/qgsbinarywidgetwrapper.cpp:122
#3  0x00007ffff5e0dd27 in QgsBinaryWidgetWrapper::setContent (this=0x5154270) at /home/even/qgis/QGIS/src/gui/editorwidgets/qgsbinarywidgetwrapper.cpp:181
#4  0x00007ffff5e0e67c in QtPrivate::FunctorCall<QtPrivate::IndexesList<>, QtPrivate::List<>, void, void (QgsBinaryWidgetWrapper::*)()>::call(void (QgsBinaryWidgetWrapper::*)(), QgsBinaryWidgetWrapper*, void**)
    (f=(void (QgsBinaryWidgetWrapper::*)(class QgsBinaryWidgetWrapper * const)) 0x7ffff5e0da74 <QgsBinaryWidgetWrapper::setContent()>, o=0x5154270, arg=0x7fffffffa590)
    at /opt/qt59/include/QtCore/qobjectdefs_impl.h:136
#5  0x00007ffff5e0e5ff in QtPrivate::FunctionPointer<void (QgsBinaryWidgetWrapper::*)()>::call<QtPrivate::List<>, void>(void (QgsBinaryWidgetWrapper::*)(), QgsBinaryWidgetWrapper*, void**) (
[...]

PR #36738, opened 2020-05-26
Fix crash on "Open Directory" entry of contextual menu of recent projects

### LINKED ISSUE #31630: QGIS crashes when "open directory" is selected from the recent projects contextual menu
## User Feedback

I open QGIS: no project, no layer loaded
I right-clicked on a recent project and chose to open its directory --> Kaboum

I tried to reproduce and succeeded many times. I got 2-3 crashes with load qgis, right-click --> "Open directory" on different valid projects (btw the folder always opened). Then I tried other options in the contextual menu and failed to have the crash. After some reloads and trials I again got a crash but with another ID [b2e8752cd39468b3be83e51f1c34e92bfdd48cc3](https://github.com/qgis/QGIS/search?q=b2e8752cd39468b3be83e51f1c34e92bfdd48cc3&type=Issues)

Not sure this report would be of any help but... Sorry if it does not help.
Question or bug (to file)?: I used to have a stack trace (i don't know if it was useful) and now I can see I have nothing generated. I don't know why and do not understand what "missing symbols" means nor what I could do to enable it. I have changed nothing in my update process. So I think it could be nice to have more details for testers that are not developers or enable these symbols in testing for us.

## Report Details

**Crash ID**: [58b3dcd0ef8f854269f6e54f486e4d541d3baf5a](https://github.com/qgis/QGIS/search?q=58b3dcd0ef8f854269f6e54f486e4d541d3baf5a&type=Issues)

**Stack Trace**
Stack trace could not be generated due to missing symbols.

**QGIS Info**
QGIS Version: 3.9.0-Master
QGIS code revision: 1b65427fdf
Compiled against Qt: 5.11.2
Running against Qt: 5.11.2
Compiled against GDAL: 3.1.0dev
Running against GDAL: 3.1.0dev

**System Info**
CPU Type: x86_64
Kernel Type: winnt
Kernel Version: 10.0.18362

rouault:
Fixes #31630

I couldn't reproduce a crash on my platform (Linux, Ubuntu 16.04, gnome classic),
but this smells pretty much like the issue of #30210 fixed per PR #30210

So using QgsFocusKeeper hoping that it would solve the issue for people affected

### DIFF
--- a/src/app/qgswelcomepage.cpp
+++ b/src/app/qgswelcomepage.cpp
@@ -18,6 +18,7 @@
 #include "qgisapp.h"
 #include "qgsversioninfo.h"
 #include "qgsapplication.h"
+#include "qgsfocuskeeper.h"
 #include "qgssettings.h"
 #include "qgsgui.h"
 #include "qgsnative.h"
@@ -291,6 +292,7 @@ void QgsWelcomePage::showContextMenuForProjects( QPoint point )
       QAction *openFolderAction = new QAction( tr( "Open Directory…" ), menu );
       connect( openFolderAction, &QAction::triggered, this, [path]
       {
+        QgsFocusKeeper focusKeeper;
         QgsGui::instance()->nativePlatformInterface()->openFileExplorerAndSelectFile( path );
       } );
       menu->addAction( openFolderAction );


nyalldawson:
I can't see any harm in this fix - so worst case it doesn't fix the crash and isn't doing anything, but it's worth a try!

nyalldawson (approved):

PeterPetrik:
unfortunately it isn't doing anything

Issue #32034, opened 2019-09-27
Apparent deadlock caused by network use of GeoPackage in QGIS 3.4.7


paulwittle:
**Describe the bug**
We have had a number of situations where multiple users have opened the same GeoPackage file over our Windows network. In most situations this seems to work but if one or more of the users attempts to edit the GeoPackage file it has led to an apparent deadlock. 

QGIS continues to operate in most case until one user attempts to stop edits or close the application. At this point all machines connected to the GeoPackage file experience errors but none of the PCs (we've had 3 on one occassion) are able to clear the lock error. 

The only solution we have found is to identify the editing machines and hard reset them (push the power button for 5+ seconds till it turns off immediately). This frees the file lock and resolves the issue but may result in data loss if the end user has not been saving regularly. 

**How to Reproduce**

We have not been able to document the exact steps to replicate so on request I will see what I can put together. It is not linked to a command so much as just opening a GeoPackage over a Windows network but we are unsure how consistent the issue is. 

We have currently tried to encourage users to source their data from our Oracle spatial database instead but we are raising the issue as we can't stop users loading GeoPackage files and so we want to limit ICT help calls and potential data loss associated with QGIS. We are actively promoting migration to the application. 

**QGIS and OS versions**

We currently use QGIS 3.4.7 (long term release) but we don't think the issue is version specific.

**Additional context**

Message trail for this ticket does exist on the QGIS users mailing list and we believe others have also experienced this issue. They may be able to provide further context.

elpaso:
Related to https://github.com/qgis/QGIS/issues/23991

Pedro-Murteira:
@paulwittle Hello, since the time this ticket was submitted, are there any news on the subject? Did the issue stop, or is it still valid?

paulwittle:
Hello, I'm not aware this issue has been fixed and as we do not have commercial support in place for QGIS most of my reports do not receive direct updates or fixes. I think they are still useful for reference and it may have been noted / fixed by others but I'm not aware of it being resolved.

If others have a more useful update or are aware of development work that has taken place then that would be great?

Thanks,
Paul

Get Outlook for Android<https://aka.ms/ghei36>
________________________________

PR #36741, opened 2020-05-26
improve geopackage support when opened through networks drives

olivierdalang:
This PR aims to improve support of geopackages when accessed from network drives (esp. windows shares).

Initial testing seems to fix the deadlock when exiting QGIS with two clients (readonly), fixing https://github.com/qgis/QGIS/issues/32034. It may also help with https://github.com/qgis/QGIS/issues/27899.

- [x] 1. fix detection of network paths 
- [x] 2. explicitly set journal_mode=delete if journal_mode=wal is unwanted
~~- [ ] 3. deal with local/remote setup~~
~~- [ ] 4. what about spatialite ?~~

1. `QFileInfo::absolutePath()` is [documented ](https://doc.qt.io/qt-5/qfileinfo.html#absolutePath)to return forward slashes on Windows too. I still left the detection for backwards slashes as I guess it must have worked on some setups at some point ?

2. `journal_mode=wal` is persisted in the database, which means we have to set `journal_model=delete` explicitly if wal is unwanted (as it's enough for a file to have been opened once with wal enabled to stay this way). Worth noting that QGIS tries to restore `journal_mode=delete` on closing the datasource, but this isn't executed in case of a crash.

3. I guess one user working on a local file (in a shared folder) with another user accessing the same file over the network is common. In such a case, in one case it's a local database (and wal will be used), in the other it's a network file (and wal won't be used). We could try to detect such cases (by checking if a `-wal` sidecar exists while the database was supposed to be openeded without wal) and show a warning and open the dataset in readonly mode to reduce risk of corruption. I guess this would only work in one way (the when the local user opens the file before the remote user), so it's definitely not fail proof... but maybe better than nothing ?

4. In the code, the logic seems to apply only for `.gpkg` files, but I guess this all also applies to `.sqlite`, right ?

### DIFF
--- a/src/core/providers/ogr/qgsogrprovider.cpp
+++ b/src/core/providers/ogr/qgsogrprovider.cpp
@@ -4047,7 +4047,7 @@ static bool IsLocalFile( const QString &path )
   // Start with the OS specific methods since the QT >= 5.4 method just
   // return a string and not an enumerated type.
 #if defined(Q_OS_WIN)
-  if ( dirName.startsWith( "\\\\" ) )
+  if ( dirName.startsWith( "\\\\" ) || dirName.startsWith( "//" ) )
     return false;
   if ( dirName.length() >= 3 && dirName[1] == ':' &&
        ( dirName[2] == '\\' || dirName[2] == '/' ) )
--- a/src/core/providers/ogr/qgsogrprovider.cpp
+++ b/src/core/providers/ogr/qgsogrprovider.cpp
@@ -3991,8 +3991,10 @@ GDALDatasetH QgsOgrProviderUtils::GDALOpenWrapper( const char *pszPath, bool bUp
   }
 
   QString filePath( QString::fromUtf8( pszPath ) );
+
+  bool bIsGpkg = QFileInfo( filePath ).suffix().compare( QLatin1String( "gpkg" ), Qt::CaseInsensitive ) == 0;
   bool bIsLocalGpkg = false;
-  if ( QFileInfo( filePath ).suffix().compare( QLatin1String( "gpkg" ), Qt::CaseInsensitive ) == 0 &&
+  if ( bIsGpkg &&
        IsLocalFile( filePath ) &&
        !CPLGetConfigOption( "OGR_SQLITE_JOURNAL", nullptr ) &&
        QgsSettings().value( QStringLiteral( "qgis/walForSqlite3" ), true ).toBool() )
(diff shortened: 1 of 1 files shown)


olivierdalang:
Shouldn't we explicitly disable WAL when working on shares, rather than explicitly enabling it when working on local files, so that it's possible to completely disable WAL if needed ?

m-kuhn:
Thank you @olivierdalang 

1. I agree, leave the `\\` they don't hurt but might help (unless a win dev here knows better)
2. :+1: 
3. Users need to be aware, that opening files locally that are shared with others is not a safe mode of operating a gpkg (with the current implementation). Given the race condition you describe still would exist with the proposed workaround, I'm not convinced it justifies the code complexity. We could rather have a look at if it's possible to further reduce wal mode.
4. ... and possibly spatialite too. I don't know about the journal mode used by those (@rouault ?), all reports I am aware of concerned gpkg.

rouault:
Both the QGIS Spatialite provider and OGR Spatialite driver should use normal DELETE journaling mode (thus potentially exposed to the same issues that lead to using WAL for GPKG, that is that readers block writers and vice versa)

olivierdalang:
Ok so let's abandon 3, which was indeed a bit desperate :-)

For 4, I suggest also not to add it to this PR, so we avoid introducing new issues and we can backport it. For reference, here's a branch with some initial work to apply wal to sqlite layers too : https://github.com/olivierdalang/QGIS/tree/geopackages_network_also_sqlite (not tested, and only for the OGR spatialite driver so far). So that would be for another issue/PR if current behaviour actually bothers anyone.

BTW, I fail to reproduce the original issue with DELETE journaling (https://github.com/qgis/QGIS/issues/23283) here on master (Win10). Will discuss this on the ML.

So, for me, this is ready to merge & backport.

m-kuhn:
@Saijin-Naib since you seem to be excited I conclude you are affected by this issue. Can you confirm it's solved and report back so we can confidently merge the backport to 3.10?

nyalldawson:
@m-kuhn Ooooooo this is dangerous to merge so close to release!! Couldn't we delay till 3.14.1?

m-kuhn:
This is a very annoying issue, the most important thing is that this is fixed in 3.10.veryverysoon, if veryverysoon can be in a month, no issues for delaying this.

Saijin-Naib:
> @Saijin-Naib since you seem to be excited I conclude you are affected by this issue. Can you confirm it's solved and report back so we can confidently merge the backport to 3.10?

Very excited about this, haha. It removes one of the few remaining limitations of GPKG when deployed on a network share when compared to ESRI File GeoDatabase.

I'm out of office indefinitely, so I can't test as robustly as I'd prefer, but I can try my best.

PR #36742, opened 2020-05-26
use completeBaseName for world file when exporting layouts (fix #34523)

### LINKED ISSUE #34523: Exporting image with atlas option "Save World file" lead to incorrect world file naming (text after dot is missing)
**Describe the bug**
Exporting image with atlas option "Save World file" lead to incorrect world file naming - text after dot is missing. However exported filename is normal.

For example: 
"2013.08.16 © Google Earth, CNES, Airbus.tif" - export filename.
"2013.tfw" - world filename.

**How to Reproduce**

1. Open Layout view.
2. Set  "map" object option "Save World file" to "True" (checkbox).

3. Use "Export as image" with filename **"2013.08.16 © Google Earth, CNES, Airbus.tif"**
4. World file should be named incorrectly: **"2013.tfw"**

**QGIS and OS versions**
3.10.2-A Coruña
Windows 10

alexbruy:
## Description
If file name of the exported layout contains multiple dots (like "Export.2020.04.03.png") worldfile will be generated with incorrect name, stripped at the first dot.

Fixes #34523.

### DIFF
--- a/src/core/layout/qgslayoutexporter.cpp
+++ b/src/core/layout/qgslayoutexporter.cpp
@@ -445,7 +445,7 @@ QgsLayoutExporter::ExportResult QgsLayoutExporter::exportToImage( const QString 
         QFileInfo fi( outputFilePath );
         // build the world file name
         QString outputSuffix = fi.suffix();
-        QString worldFileName = fi.absolutePath() + '/' + fi.baseName() + '.'
+        QString worldFileName = fi.absolutePath() + '/' + fi.completeBaseName() + '.'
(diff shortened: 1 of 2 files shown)


nirvn:
@alexbruy , is this an issue for the save map canvas as image function?

alexbruy:
@nirvn I will check

alexbruy:
Also fixed in the `saveAsImage()` method of the `QgsMapCanvas` class. Exporting with `QgsMapRendererTask` already uses `completeBaseName`.

nirvn:
Thanks.

PR #36743, opened 2020-05-26
Avoid leaking sqlite3_stmt on errors

audun:
## Description

Fixes two small leaks that could happen if an error occurred.


m-kuhn (approved):
Looks good to me.

For reference, it would be good to move those to the RAII implementations in QgsSqliteUtils which will make sure things like this don't happen.

PR #36744, opened 2020-05-26
SpatiaLite support for importing layers with binary fields

### LINKED ISSUE #36705: Import layer to Spatialite/PostGis does not support binary data
**Describe the bug**

The QGIS Spatialite provider does not support binary data fields.

**How to Reproduce**

1. Create a new scratch layer

image

2. Add a binary field

image

Result looks good so far:
image
The attached project file is saved at this point: [minimal_binary_field.zip](https://github.com/qgis/QGIS/files/4677779/minimal_binary_field.zip)

3. Create a new Spatialite db
4. Import the layer to the new DB

5. Observe, with the appropriate amount of sadness, the error message:

image

**QGIS and OS versions**

3.12 and latest master

**Additional context**

We have some prepared binary data that is attached to each point in the layer. The current method is to encode the data in base64 and add it as a TEXT field, then later run a separate program that reads each string, decodes it and reinserts it as a proper binary blob.

I have created a patch/PR based on the PostGis implementation of bytea since binary data has worked there since @nyalldawson's commits last year. However, to my surprise, importing didn't work there either.

image

However, adding the field using the add field dialog works.

image

image

This does not currently work in Spatialite, since it doesn't have any QVariant::ByteArray support today.

I can make a separate PR with the few lines that is necessary to make importing work in PostGIS too:

image

audun:
## Description

Adds support for binary fields to SpatiaLite like in PostGIS. Fixes #36705.
prepare_commit.sh has been run.

A few thoughts about the changes:

~Since there are currently no unit tests (that I am aware of) for Spatialite~ There are plenty of unit tests run during the Python tests, but since I wasn't aware of them, I wanted to make the changes minimal and easy to verify.  There should be absolutely no functional changes in anything except for supporting ByteArray fields. I have tested all the changed code paths manually, including the error paths (by intentionally introducing errors in the SQL).

In the UPDATE statement I use sqlite3_bind_blob instead of using a binary literal. It's possible in SQLite to write blob literals by first encoding it as hexadecimal, but I think it's more difficult to verify the correctness and would have much worse performance than simply binding the value directly.

By taking that approach further and using bind instead of string literals in the entire statement I believe it's possible to move the binding of values in both the INSERT and the UPDATE statement to a common function. Should result in less code and better performance for UPDATE, but doing that would cause changes to the current functionality.

Also as @m-kuhn mentioned, using RAII would simplify the necessity of using finalize properly everywhere.

That must be in another PR or three though.

### DIFF
--- a/src/providers/spatialite/qgsspatialitefeatureiterator.cpp
+++ b/src/providers/spatialite/qgsspatialitefeatureiterator.cpp
@@ -588,6 +588,14 @@ QVariant QgsSpatiaLiteFeatureIterator::getFeatureAttribute( sqlite3_stmt *stmt, 
     return sqlite3_column_double( stmt, ic );
   }
 
+  if ( sqlite3_column_type( stmt, ic ) == SQLITE_BLOB )
+  {
+    // BLOB value
+    int blob_size = sqlite3_column_bytes( stmt, ic );
+    const char *blob = static_cast<const char *>( sqlite3_column_blob( stmt, ic ) );
+    return QByteArray( blob, blob_size );
+  }
+
   if ( sqlite3_column_type( stmt, ic ) == SQLITE_TEXT )
   {
     // TEXT value
--- a/src/providers/spatialite/qgsspatialiteprovider.cpp
+++ b/src/providers/spatialite/qgsspatialiteprovider.cpp
@@ -116,6 +116,12 @@ bool QgsSpatiaLiteProvider::convertField( QgsField &field )
       break;
     }
 
+    case QVariant::ByteArray:
+      fieldType = QStringLiteral( "BLOB" );
+      fieldSize = -1;
+      fieldPrec = -1;
+      break;
+
     default:
       return false;
   }
@@ -688,6 +694,10 @@ static TypeSubType getVariantType( const QString &type )
   {
     return TypeSubType( QVariant::List, QVariant::Invalid );
   }
+  else if ( type == QLatin1String( "blob" ) )
+  {
+    return TypeSubType( QVariant::ByteArray, QVariant::Invalid );
+  }
   else if ( type == QLatin1String( "timestamp" ) ||
             type == QLatin1String( "datetime" ) )
   {
@@ -738,6 +748,11 @@ void QgsSpatiaLiteProvider::loadFieldsAbstractInterface( gaiaVectorLayerPtr lyr 
       {
         fieldType = QVariant::Double;
         type = "DOUBLE";
+      }
+      if ( fld->BlobValuesCount != 0 )
+      {
+        fieldType = QVariant::ByteArray;
+        type = "BLOB";
       }
       mAttributeFields.append( QgsField( name, fieldType, type, 0, 0, QString() ) );
     }
@@ -4217,6 +4232,12 @@ bool QgsSpatiaLiteProvider::addFeatures( QgsFeatureList &flist, Flags flags )
             QByteArray ba = stringVal.toUtf8();
             sqlite3_bind_text( stmt, ++ia, ba.constData(), ba.size(), SQLITE_TRANSIENT );
           }
+          else if ( type == QVariant::ByteArray )
+          {
+            // binding a BLOB value
+            const QByteArray ba = v.toByteArray();
+            sqlite3_bind_blob( stmt, ++ia, ba.constData(), ba.size(), SQLITE_TRANSIENT );
+          }
           else if ( type == QVariant::StringList || type == QVariant::List )
           {
             const QByteArray ba = QgsJsonUtils::encodeValue( v ).toUtf8();
@@ -4538,6 +4559,10 @@ bool QgsSpatiaLiteProvider::changeAttributeValues( const QgsChangedAttributesMap
   for ( QgsChangedAttributesMap::const_iterator iter = attr_map.begin(); iter != attr_map.end(); ++iter )
   {
     // Loop over all changed features
+    //
+    // For each changed feature, create an update string like
+    // "UPDATE table SET simple_column=23.5, complex_column=? WHERE primary_key=fid"
+    // On any update failure, changes to all features will be rolled back
 
     QgsFeatureId fid = iter.key();
 
@@ -4552,10 +4577,13 @@ bool QgsSpatiaLiteProvider::changeAttributeValues( const QgsChangedAttributesMap
     QString sql = QStringLiteral( "UPDATE %1 SET " ).arg( QgsSqliteUtils::quotedIdentifier( mTableName ) );
     bool first = true;
 
+    // keep track of map of parameter index to value
+    QMap<int, QVariant> bindings;
+    int bind_parameter_idx = 1;
+
     // cycle through the changed attributes of the feature
     for ( QgsAttributeMap::const_iterator siter = attrs.begin(); siter != attrs.end(); ++siter )
     {
-      // Loop over all changed attributes
       try
       {
         QgsField fld = field( siter.key() );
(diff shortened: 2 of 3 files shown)


src/providers/spatialite/qgsspatialiteprovider.cpp
```diff
@@ -4564,9 +4591,61 @@ bool QgsSpatiaLiteProvider::changeAttributeValues( const QgsChangedAttributesMap
...
+          const QByteArray ba = val.toByteArray();
+          sqlite3_bind_blob( stmt, parameter_idx, ba.constData(), ba.size(), SQLITE_TRANSIENT );
+          break;
+        }
+
+        default:
+          // This will only happen if the above code is changed to bind more types,
+          // but the programmer has forgotten to handle the type here. Fatal error.
+          sqlite3_finalize( stmt );
+          throw std::runtime_error( "Attempting to bind unhandled type." );
```
nyalldawson, line 4626: Better to Q_ASSERT( false ) here

nyalldawson (commented):
Can you add some new unit tests covering this functionality?

audun:
Yes, absolutely. I was wondering how to approach writing tests, but incorporating some new ones in Python should be easy. This test suite will give me confidence enough in refactoring out the duplicate code eventually too. Thanks for reviewing the PR so far.

I have a question: It appears that the test for the Spatialite provider, `tests/src/python/test_provider_spatialite.py` is not run during the build. Seems like that might be the case for all the DB providers, but perhaps Spatialite should be an exception as it does not need a server?

Locally the test fails in an unrelated test, when it tries to remove the database file in testPKNotInt test. It seems like that test does the cleanup differently than all the rest, for some reason. Commenting out that `shutil.rmtree(basepath)` call makes the entire test suite pass. I'll commit a fix for that too.

nyalldawson:
> I have a question: It appears that the test for the Spatialite provider, tests/src/python/test_provider_spatialite.py is not run during the build.

It is run on Travis CI. None of the tests are run during the build, you need to manually run all of them post-build. See https://docs.qgis.org/testing/en/docs/developers_guide/unittesting.html

src/providers/spatialite/qgsspatialiteprovider.cpp
```diff
@@ -44,6 +44,8 @@ email                : a.furieri@lqt.it
 #include <QDir>
 #include <QRegularExpression>
 
+#include <stdexcept>
```
elpaso, line 47: not needed anymore?

src/providers/spatialite/qgsspatialiteprovider.cpp
audun: It seems like the Q_ASSERT is not checked in release mode, so if the programmer forgets to handle the type correctly here it could then silently cause user data loss. Since this error can never occur because of user input or data I think it's better to visibly let it bubble up to the "Caught unhandled std::exception" handler.

What do you think?

audun:
I have now added test cases exercising the binary layers in the same manner that is done in TestPyQgsPostgresProvider (exposing a bug 🐛) and rebased it on latest master.

src/providers/spatialite/qgsspatialiteprovider.cpp
audun: Depends on the resolution of the other comment, but I have followed the advice there in the current version so it's gone at the moment.

audun:
> > I have a question: It appears that the test for the Spatialite provider, tests/src/python/test_provider_spatialite.py is not run during the build.
> 
> It is run on Travis CI. None of the tests are run during the build, you need to manually run all of them post-build. See https://docs.qgis.org/testing/en/docs/developers_guide/unittesting.html

Sorry for being unclear. I meant on the build server during CI, but I was looking only on the Azure build and not on Travis. 
The only test I can see on Azure with Spatialite in the name is [PyQgsDBManagerSpatialite](https://dev.azure.com/qgis/QGIS/_build/results?buildId=5821&view=logs&j=c55c25a0-2401-5b7f-f792-650180689337&t=e313f7aa-4452-5f33-1018-5aea7c0f02b7&l=1069).  I expected to see PyQgsSpatialiteProvider and PyQgsProviderConnectionSpatialite too, like when I am running tests locally.

Anyway, on Travis it is there as expected:

`        Start 524: PyQgsSpatialiteProvider
524/644 Test #524: PyQgsSpatialiteProvider ...........................   Passed    4.26 sec`

m-kuhn:
The azure ones were integrated only recently and we just enabled to avoid regressions. This one was probably failing but it was not investigated why.
It would be great to enable more tests on Azure. If you want to give it a go, remove it from [this list](https://github.com/qgis/QGIS/blob/master/.ci/azure-pipelines/azure-pipelines.yml#L4) and see what's required to stabilize it.

audun:
Perhaps it had the same problem that I had to fix to get it running locally on Windows? I can try to remove it later, but currently it seems like there is a very long queue of builds so might not the best time to experiment.

nirvn (approved):
Looking good. +1 to merge when travis is green.

audun:
Perfect, thanks! 

Tested to use the conflict resolution tool in GitHub, but it created a merge commit. Did a rebase onto latest master to remove the merge commit as recommended in the contributor guide.

src/providers/spatialite/qgsspatialiteprovider.cpp
```diff
@@ -4623,9 +4657,61 @@ bool QgsSpatiaLiteProvider::changeAttributeValues( const QgsChangedAttributesMap
...
+      // some unexpected error occurred during preparation
+      const char *err = sqlite3_errmsg( sqliteHandle( ) );
+      errMsg = ( char * ) sqlite3_malloc( ( int ) strlen( err ) + 1 );
+      strcpy( errMsg, err );
+      handleError( sql, errMsg, savepointId );
+      return false;
+    }
+
+    // bind variables not handled directly in the string
+    QMap<int, QVariant>::iterator i;
```
m-kuhn, line 4674: Can you use a const iterator here?

src/providers/spatialite/qgsspatialitefeatureiterator.cpp
```diff
@@ -588,6 +588,14 @@ QVariant QgsSpatiaLiteFeatureIterator::getFeatureAttribute( sqlite3_stmt *stmt,
     return sqlite3_column_double( stmt, ic );
   }
 
+  if ( sqlite3_column_type( stmt, ic ) == SQLITE_BLOB )
+  {
+    // BLOB value
+    int blob_size = sqlite3_column_bytes( stmt, ic );
+    const void *blob = sqlite3_column_blob( stmt, ic );
+    return QByteArray( ( const char * ) blob, blob_size );
```
m-kuhn, line 596:
```suggestion
    return QByteArray( static_cast<const char *>( blob ), blob_size );
```

Super minor nitpicking, static_casts are the preferred way to handle this. (also in qgsspatialiteprovider.cpp )

audun: I totally agree, I actually wrote it like that originally, but I try to follow the style close to the code, and a few lines down there's another C-style cast.

```
    // TEXT value
    const QString txt = QString::fromUtf8( ( const char * ) sqlite3_column_text( stmt, ic ) );
```

I'll follow your suggestion here, and we can perhaps make a separate PR making it consistently C++-style casts elsewhere.

src/providers/spatialite/qgsspatialiteprovider.cpp
audun: Good idea. I see auto used above, so that makes it even easier.

nyalldawson:
Superseded by https://github.com/qgis/QGIS/pull/36876

Issue #36545, opened 2020-05-19
WFS Provider WARNING - Cannot create temporary SpatiaLite cache when using flatpak install


palmerj:
**Describe the bug**
When using the flatpak version of QGIS 3.12.2 on Ubuntu and adding any WFS layer I get the following error:

```
WARNING    Cannot create temporary SpatiaLite cache.
```

I have set the networking cache directory to `~/qgis/cache` and ensure the directory actually has 0777 right to ensure it's not some permissions issue. I note the the following empty directory gets created: `~/.qgis/cache/wfsprovider/pid_2`

I'm wondering if this issue is caused by permissions sandboxing as implemented by flatpak.

**QGIS and OS versions**

QGIS version
3.12.2-Bucuresti
QGIS code revision
8a1fb336
Compiled against Qt
5.14.2
Running against Qt
5.14.2
Compiled against GDAL/OGR
3.0.4
Running against GDAL/OGR
3.0.4
Compiled against GEOS
3.8.1-CAPI-1.13.3
Running against GEOS
3.8.1-CAPI-1.13.3
Compiled against SQLite
3.29.0
Running against SQLite
3.29.0
PostgreSQL Client Version
12.2
SpatiaLite Version
4.3.0a
QWT Version
6.1.4
QScintilla2 Version
2.10.8
Compiled against PROJ
6.3.1
Running against PROJ
Rel. 6.3.1, February 10th, 2020
OS Version
KDE Flatpak runtime
Active python plugins
MetaSearch; 
db_manager; 
processing

nyalldawson:
The flatpak install isn't officially supported. You'll need to file this with the maintainers of that package.

palmerj:
It's on the download pages as is mentioned as "There is an experimental QGIS flatpak for QGIS Stable available." Where do you state this is not officially supported?

Also I suspect a change in the core of QGIS can fix this, or at least given better information as to why it has failed

nyalldawson:
Ah - i wasn't aware we were pointing users towards it. Not sure who here has experience with that particular package though -- I still think you'd have more luck asking over at flathub.

palmerj:
I'm wondering if it's got anything to do with the GDAL vsimem file handler and how it's implemented. The flatpak QGIS configuration is very permissive and allow access to pretty much all of the host operating system files for read as write.

palmerj:
> I still think you'd have more luck asking over at flathub.

I've also done that. Thanks

rouault:
[...]

PR #36745, opened 2020-05-26
WFS provider: add more details to log message when creation of temporary SpatiaLite cache fails

rouault:
Should help in the diagnosis of issue #36545


nyalldawson (approved):

palmerj:
Thank you @rouault

PR #36754, opened 2020-05-26
[Server, WCS] Fix several non compliance errors in CoverageDescription output

### LINKED ISSUE #36504: WCS DescribeCoverage response origin always empty
When publishing a WCS services with QGIS Server, the "origin" element in the RectifiedGrid node of the spatial domain of a coverage in the DescribeCoverage response appears to always be empty.
In all my attempts, it renders as:
`<gml:origin/>`
It should instead contain a position element, like:
```
<gml:origin>
   <gml:pos>0 0</gml:pos>
</gml:origin>
```
I believe I found the culprit: the position element is never added into the origin element before the latter is added into the response; see source https://github.com/qgis/QGIS/blob/8a7feeddbb8ce3b0b85fc71101c718e4586db1d4/src/server/services/wcs/qgswcsutils.cpp#L164

The fix is likely to add:
`originElem.appendChild( originPosElem );`
between lines 163 and 164

Environment: QGIS Server 3.12 on Ubuntu 19.10

rouault:
There were several non-compliance in the CoverageDescription output to a
DescribeCoverage request. The corrected result has been validated against the
OGC WCS 1.0.0 schema.

Fixes #36504

### DIFF
--- a/src/server/services/wcs/qgswcsutils.cpp
+++ b/src/server/services/wcs/qgswcsutils.cpp
@@ -150,28 +150,29 @@ namespace QgsWcs
     QDomElement xAxisElem = doc.createElement( QStringLiteral( "gml:axisName" ) );
     QDomText xAxisText = doc.createTextNode( QStringLiteral( "x" ) );
     xAxisElem.appendChild( xAxisText );
-    spatialDomainElem.appendChild( xAxisElem );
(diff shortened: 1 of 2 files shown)


rldhont (approved):

rldhont:
Thanks @rouault

Issue #34471, opened 2020-02-14
Projection not recognised from Mapinfo converstion


untoldstory2:
When I try to open a shape file that has been translated from Mapinfo discover the projection isn't recognised. The grid is a local grid. Output.zip is the attached file that i'm trying to open

[Output.zip](https://github.com/qgis/QGIS/files/4203313/Output.zip)

I'm currently using Qgis version 3.10.2

**Additional context**

Here are some control points:

Mine X | Mine Y| MGA 94(50) X| MGA 94(50) Y
2453.122|3210.002|563053.406|7431461.771
-1735.225|-853.24|557798.872|7428929.256
5663.648|7386.58|567416.171|7434410.368
12607.859|-1438.839|571218.306|7423848.605
8502.84|2620.24|568605.287|7428993.832
-2500.032|3457.767|558433.331|7433259.449

**Here is the Mapinfo .prj file entry:**

"Mine to MGA", 3008, 33,7, 117, 0, 0.9996, 500000, 10000000 , 7, 0.949288, -0.315744, 1814394.28, 0.315737, 0.94926, -7228957.16, -100000, -100000, 100000, 100000

Asratum:
Possibly related: a custom Mercator projection is recognized in 3.10.2 as invalid, but in 3.4.9 is recognized correctly as custom CRS. Even if 3.10.2 calls it invalid, it projects it correctly.
.shp below loads correctly in 3.4.9 and has "invalid projection" in 3.10.2, when I choose Save as... after importing.
[Merc_25S_25W.zip](https://github.com/qgis/QGIS/files/4203829/Merc_25S_25W.zip)

nyalldawson:
What's the issue here? The projection isn't recognised because it **isn't** a standard projection.

Asratum:
The issue is that in earlier versions it would be recognized as a user defined CRS and given a 10000n number, and in 3.10.2 it just says it's invalid, not assigning it a CRS ID. Applies to both files uploaded so far.

nyalldawson:
Can you test with a nightly 3.10.3 release or master? I don't see the same behavior here.

Asratum:
I tested this with a 3.11.06 release from the weekly downloads (not sure if that's correct?). The problem persists; in addition it seems that a .prj string is not properly recognized - the following .prj files are nearly identical, yet one of the shapefiles gets an unknown CRS. 
3.4.9 recognizes both files correctly as UTM32S.
[...]

PR #36758, opened 2020-05-26
QgsVectorLayerSaveAsDialog: make it work with custom CRS

rouault:
The current approach was based on a srsid, that that might be unset for custom
CRS. So store a full QgsCoordinateReferenceSystem object internally and use
that for exchanges with QgsProjectionSelectionWidget

Consequence: we need to deprecate the "long crs()" method, and replace it
by a "QgsCoordinateReferenceSystem crsObject()" one. I'm not particularly
happy with the name but the existing one was not very well named...

Mark also the QgsVectorLayerSaveAsDialog( long srsid, ... ) constructor as
deprecated.

Fixes the use case of https://github.com/qgis/QGIS/issues/34471#issuecomment-586181977

### DIFF
--- a/src/gui/ogr/qgsvectorlayersaveasdialog.cpp
+++ b/src/gui/ogr/qgsvectorlayersaveasdialog.cpp
@@ -39,7 +39,7 @@ static const int COLUMN_IDX_EXPORT_AS_DISPLAYED_VALUE = 2;
 
 QgsVectorLayerSaveAsDialog::QgsVectorLayerSaveAsDialog( long srsid, QWidget *parent, Qt::WindowFlags fl )
   : QDialog( parent, fl )
-  , mCRS( srsid )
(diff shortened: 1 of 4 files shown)


src/gui/ogr/qgsvectorlayersaveasdialog.cpp
```diff
@@ -39,7 +39,7 @@ static const int COLUMN_IDX_EXPORT_AS_DISPLAYED_VALUE = 2;
 
 QgsVectorLayerSaveAsDialog::QgsVectorLayerSaveAsDialog( long srsid, QWidget *parent, Qt::WindowFlags fl )
```
nyalldawson, line 40: Can you deprecate this constructor? I want to completely eliminate srs ids in qgis 4.0...

rouault: done

nyalldawson (approved):

Issue #16100, opened 2013-01-14
Line pattern fill is not rendered to vector in pdf export


qgib:
Author Name: **Victor Axbom** (Victor Axbom)
Original Redmine Issue: [6996](https://issues.qgis.org/issues/6996)
Affected QGIS version: master
Redmine category:symbology

---
I've experienced some problems when I export maps to vector-pdf from the composer. All features in the exported pdf are vectors except for the layers which have the relative new line pattern fill set as symbology type. These are rendered as rasters, and this become a problem when trying to print to larger formats as A1 as some features from these layers disappear in the exported pdf.

---

- [pdf_export.pdf](https://issues.qgis.org/attachments/download/5256/pdf_export.pdf) (Victor Axbom)

---

Related issue(s): #27676 (duplicates)
Redmine related issue(s): [19852](https://issues.qgis.org/issues/19852)

---

T-Dunne:
Hi, is there any movement on this issue? Thanks.

jmonticolo:
Here is my workaround : a symbol layer, MultiLineString, pattern fill : [https://gis.stackexchange.com/a/354581/93097](https://gis.stackexchange.com/a/354581/93097)

pjshelton:
Still seems to be an issue in V3.16.3

I created a polygon scratch layer, symbolised it with a standard fill, and a line fill.
![image](https://user-images.githubusercontent.com/39154868/106686060-6cc30a80-6615-11eb-8364-6531e9375042.png)

When I create a composer and export to PDF, the line fill is still a raster.
![image](https://user-images.githubusercontent.com/39154868/106695122-b0be0b80-6625-11eb-9ca8-d91ed1a1d047.png)
as can be seen in this image.
![image](https://user-images.githubusercontent.com/39154868/106695174-c92e2600-6625-11eb-8e98-2d7bed7783c1.png)
The simple fill is bounded by a vector but the line fill is a raster.

Gwylohm:
Still seems to be an issue in V3.18 too.

PR #36759, opened 2020-05-27
Don't force rasterized output when exporting point pattern fills

nyalldawson:
This avoids the (very long standing!) issue of forced conversion to a raster based pattern which currently occurs when exporting maps/layouts to a vector format (e.g. PDF). The raster pattern
results in considerable quality loss, and the tiling edges of the raster brush can sometimes be seen in outputs.

Additionally, fixes render corrupt when marker subsymbols have data defined properties which affect the marker shape, such as data defined rotation or sizes

Refs #16100 (still needs fixing for line fill symbols)


nyalldawson:
Example of the data defined raster brush tiling issue:

Using a random marker size with the fill pattern previously gave this:

image

Now you get the desired result:

image

Also note the dramatic quality improvement in the rendering quality - antialiasing is properly applied!

This fix allows creation of symbols which were previously corrupted, such as scree slopes:

image

image

image

nyalldawson:
ping @andreasneumann

src/core/symbology/qgsfillsymbollayer.cpp
```diff
@@ -3427,13 +3570,13 @@ void QgsPointPatternFillSymbolLayer::applyDataDefinedSettings( QgsSymbolRenderCo
   double offsetX = mOffsetX;
   if ( mDataDefinedProperties.isActive( QgsSymbolLayer::PropertyOffsetX ) )
   {
-    context.setOriginalValueVariable( mDisplacementX );
+    context.setOriginalValueVariable( mOffsetX );
```
nirvn, line 3573: Ouch -- thanks for fixing.

nirvn (approved):
Great job.

nyalldawson:
Thanks @nirvn!

PR #36768, opened 2020-05-27
Optimized PNG files using optipng

lhirlimann:
Reduce file size of *png in the source tree.  Ran the optipng program with the highest value 7.
This reduces the overall size of png files by 80% : 
```shell
time find . -name "*png" |xargs optipng -o7
```
I calculated the size of the png images using : 
```shell
find . -name "*.png" -print0 | du -sb --files0-from=-  | awk '{ total += $1} END { print total/1024/1024/1024 }'
```
value returned before running the script : **0.157557**
value returned after running the script : **0.0338007**

I timed my efforts : 
```shell
find . -name "*png"  0.04s user 0.05s system 97% cpu 0.087 total
xargs optipng -o7  5977.41s user 21.35s system 99% cpu 1:40:27.69 total
```

This doesn't change any code, just png files reducing their size.


nyalldawson:
Hmm - maybe skip the tests folder for now. Something in conversion has broken a lot of tests.

lhirlimann:
So it's still fails on some test - way less :(

nyalldawson:
@lhirlimann I'd suggest closing this branch (it has merge conflicts which will be very tricky to resolve), and opening separate PRs where you optimise each directory of pngs in turn. There's likely quite a few which could be easily merged quickly.

PR #36654, opened 2020-05-22
[Backport release-3_10] oracle identity field #33683 #34546

SebastienPeillet:
## Description

It is a backport of PR #33683 and #34546 (that fixes an exception in the previous one). It fixes the use of identity field in Oracle DB.

The first attempt to backport #33683 didn't work due to a merge conflict (see #34260 ) so I rebase it.

PR #36769, opened 2020-05-27
Oracle - Wrong default value FIX #34482

### LINKED ISSUE #32401: Invalid query for default value of date column of an oracle layer
**Error description**
I added an Oracle spatial layer to my QGIS project and added a new polygon. When saving the changes, I get an error "FROM keyword not found where expected, ... Select 0 sysdate from dual".

**How to reproduce**
The error occurs if I do not insert any value for a date field in the attribute table of the new polygon. In the attribute settings of the QGIS layer the date field is set to editable=false, and no default value is set. 

The Oracle database date column is defined as type=DATE, nullable=YES, default=SYSDATE. QGIS probably tries to get a default value for the emtpy field, but the SQL query "Select 0 sysdate from dual" doesn´t work.

**Temporary workaround**
I found a temporary workaround for this topic: when setting the date attribut for my QGIS layer on "editable" (I don´t know how this option is called in english edition - in the german edition it is called "änderbar"), then saving the polygon works. But I do not want to allow the user to change the attribute (the empty value should be set by the database "default valuie trigger").

_System settings: QGIS 3.8.3, Oracle 12c_

### LINKED ISSUE #34482: [Oracle]Error on save geometries because wrong default values
hello,
i think there is an issue with default values for attributes, if i use oracle provider and create or paste new geometries. details in the gif's. the behaivior depends on the datatype. in my cases varchar and number. create statements are in the txt file. 

the error occur with and without showing attribute form

affected versions: 3.4.15;3.10.2 actual master

hope someone can reproduce this issue.

thanks
jan

[create_gis_gruen_tables.txt](https://github.com/qgis/QGIS/files/4208409/create_gis_gruen_tables.txt)
oracle_save_error_varchar
oracle_save_error_number

SebastienPeillet:
FIX #34482
FIX #32401

## Description

Default values are misreaded from `all_tab_columns` table.

Default values are retrieving from the `data_default` field in `all_tab_columns` table (see [here](https://github.com/qgis/QGIS/blob/master/src/providers/oracle/qgsoracleprovider.cpp#L604)).

And long story short : 
- `data_default` is a long type column
- long is a type that can contains a lot of data, so data is readed in a "piecewise" mode (see [qsql_ocispatial.cpp](https://github.com/qgis/QGIS/blob/master/src/providers/oracle/ocispatial/qsql_ocispatial.cpp#L1535) and [oracle documentation](https://docs.oracle.com/database/121/HETER/GUID-9F828F0D-60ED-4F7D-92D5-0F3DFE1C9AA7.htm#HETER4069))
- piecewise read mode concatenates the value piece by piece (for a single row), but after the first row the value is not cleared, so it concatenates the current default value with the previous one, etc

The solution is to clear values if piecewise mode is used.

We already test default values in [test_provider_oracle.py](https://github.com/qgis/QGIS/blob/master/tests/src/python/test_provider_oracle.py#L234) :

```python
    def testDefaultValue(self):
        self.assertEqual(self.source.defaultValue(1), NULL)
        self.assertEqual(self.source.defaultValue(2), "'qgis'")
```

I add `self.assertEqual(self.source.defaultValue(3), "'qgis'")` that is not tested yet. Without the PR changes, `self.source.defaultValue(3)` is equal to "qgisqgis" and not "qgis"

No backport before backporting #36654

### DIFF
--- a/src/providers/oracle/ocispatial/qsql_ocispatial.cpp
+++ b/src/providers/oracle/ocispatial/qsql_ocispatial.cpp
@@ -3432,7 +3432,10 @@ bool QOCISpatialResult::gotoNext( QSqlCachedResult::ValueCache &values, int inde
 
   // need to read piecewise before assigning values
   if ( r == OCI_SUCCESS && piecewise )
+  {
+    values.clear();
     r = d->cols->readPiecewise( values, index );
+  }
 
   if ( r == OCI_SUCCESS )
     d->cols->getValues( values, index );
--- a/tests/src/python/test_provider_oracle.py
+++ b/tests/src/python/test_provider_oracle.py
@@ -234,6 +234,7 @@ class TestPyQgsOracleProvider(unittest.TestCase, ProviderTestCase):
     def testDefaultValue(self):
         self.assertEqual(self.source.defaultValue(1), NULL)
         self.assertEqual(self.source.defaultValue(2), "'qgis'")
+        self.assertEqual(self.source.defaultValue(3), "'qgis'")
 
     def testPoints(self):
         vl = QgsVectorLayer('%s table="QGIS"."POINT_DATA" (GEOM) srid=4326 type=POINT sql=' %


nyalldawson (approved):

uclaros:
@SebastienPeillet @nyalldawson I am not quite sure what this change fixed or what *piecewise* is, but I am quite confident it is responsible for #42857 that I stumbled upon recently.

Right after clearing `values`:
https://github.com/qgis/QGIS/blob/642deb194d7f94bd47343b4f68fb1dce265d28a7/src/providers/oracle/ocispatial/qsql_ocispatial.cpp#L3416-L3417
`values` is accessed by index within `readPiecewise()` causing a crash:
https://github.com/qgis/QGIS/blob/642deb194d7f94bd47343b4f68fb1dce265d28a7/src/providers/oracle/ocispatial/qsql_ocispatial.cpp#L1590-L1605


I don't know what changed to force me to the *piecewise* codepath though, I've never had issues until this week!

PR #36771, opened 2020-05-27
Try to reactivate tests for legend and json export

pblottiere:
## Description

Unit tests are green on Arch. Let's take a look on Travis...

### DIFF
--- a/tests/src/core/testqgslegendrenderer.cpp
+++ b/tests/src/core/testqgslegendrenderer.cpp
@@ -1440,10 +1440,6 @@ void TestQgsLegendRenderer::testOpacityJson()
   const QJsonObject point_layer = root[1].toObject();
   const QJsonArray point_layer_symbols = point_layer["symbols"].toArray();
 
-#if 0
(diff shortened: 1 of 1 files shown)


nyalldawson:
Is the rendered result here correct? When I disabled the test the rendered result wasn't taking into account the opacity. Has it since been fixed?

pblottiere:
> Is the rendered result here correct? When I disabled the test the rendered result wasn't taking into account the opacity. Has it since been fixed?

Yes it seems so. Here is what I got for the blue icon and the `0.5` opacity generated in `/tmp/point_layer_icon_blue_opacity.png` by `ctest`:

blue

Same results for the red and green icons.

nyalldawson:
Hmm ok - guess something has fixed it in the meantime!

nyalldawson (approved):

pblottiere:
> guess something has fixed it in the meantime!

Well, I'm curious to know if it's still happening on your environment (if you have time of course). Because I didn't see any bugfix related to that since you deactivated the test, so it could be a false positive.

PR #36774, opened 2020-05-27
improve logic for choosing style type when exporting raster style (fix #35944)

### LINKED ISSUE #35944: Exporting raster style to SLD only works when output file has .sld suffix
# Problem

I have symbolized a raster layer and when I try to export it as an SLD it always saves it as QML.

qgis

alexbruy:
## Description
Now, when user tries to save raster style to SLD and entered filename without `.sld` extension in the save dialog, style will be exported as QML even if SLD filter is selected. 

Proposed PR changes logic to rely only on the selected filter to determine style type (QML or SLD). So if for example SLD filter is selected but file name has `.qml` extension style will be exported as SLD and file will have suffix `.qml.sld`.

Fixes #35944.

### DIFF
--- a/src/gui/raster/qgsrasterlayerproperties.cpp
+++ b/src/gui/raster/qgsrasterlayerproperties.cpp
@@ -2169,21 +2169,28 @@ void QgsRasterLayerProperties::saveStyleAs_clicked()
   QgsSettings settings;
   QString lastUsedDir = settings.value( QStringLiteral( "style/lastStyleDir" ), QDir::homePath() ).toString();
 
+  QString selectedFilter;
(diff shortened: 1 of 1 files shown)


alexbruy:
Probably we even should not rely on the manually entered extension and only take into account selected file filter.

nyalldawson (approved):

nyalldawson:
> Probably we even should not rely on the manually entered extension and only take into account selected file filter.

That sounds good!

alexbruy:
Ok, chnaged to use only file filter to determine style type. First comment also updated to reflect chnages.

PR #36778, opened 2020-05-27
 QgsDelimitedTextFile: fix parsing of files with CR end of line 

### LINKED ISSUE #36392: QGIS crashes/freezes when adding a large CSV file as delimited text layer
Operating system: Windows 10 Pro (Version 1909)
QGIS Version: 3.12
Attaching the xlsx version of the csv file as github doesnt support csv files.
I was not having this issue until today. Started using QGIS version 3.10 (more stable) last week and this morning was working fine until noon where it started crashing when uploading this delimited text layer. I restarted my pc, nothing, unistalled version 3.10 and installed version 3.12 but the problem persisted. Thanks for any type of help in advance.
[ATT.xlsx](https://github.com/qgis/QGIS/files/4619070/ATT.xlsx)

rouault:
*  QgsDelimitedTextFile: fix parsing of files with CR end of line 

Fixes #36392
Fixes #21976 
Fixes #17190

We are obliged to do 'at hand' parsing due to QT not handling CR-only
end of lines.
As we are at it, also limit each line to 1 MB to avoid potential denial
of service (which was what close to what happened here before the CR-only
parsing fix)

Add tests for parsing CR-only end of lines, and exercising the at-hand
buffering logic

*  QgsDelimitedTextSourceSelect: add a limitation to the number of fields
9d85148

Set to 10000 by default. Can be overriden through settings.

This is only for the preview in the source select dialog. (Attribute
table performs badly for much less columns)

---------------------

While being a bug fix, I'm not completely sure this is material for 3.10. At least not without being exercised a bit in master. The logic to parse lines at hand isn't completely trivial.

### DIFF
--- a/src/providers/delimitedtext/qgsdelimitedtextfile.cpp
+++ b/src/providers/delimitedtext/qgsdelimitedtextfile.cpp
@@ -39,6 +39,10 @@ QgsDelimitedTextFile::QgsDelimitedTextFile( const QString &url )
   // The default type is CSV
   setTypeCSV();
   if ( ! url.isNull() ) setFromUrl( url );
+
+  // For tests
+  QString bufferSizeStr( getenv( "QGIS_DELIMITED_TEXT_FILE_BUFFER_SIZE" ) );
+  mMaxBufferSize = bufferSizeStr.isEmpty() ? 1024 * 1024 : bufferSizeStr.toInt();
 }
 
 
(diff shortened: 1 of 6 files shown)


nyalldawson:
I believe this also fixes https://github.com/qgis/QGIS/issues/21976 ?

nyalldawson:
Also https://github.com/qgis/QGIS/issues/17190

nyalldawson (approved):

rouault:
@nyalldawson any opinion if we should tag this for 3.10 backport ?

nyalldawson:
I think it's fine to backport (with the time delay). It fixes a crash, so would be good to have in ltr. We could even delay two releases if you're concerned.

PR #36804, opened 2020-05-28
ValueMap widget: show unmatched items in ()

### LINKED ISSUE #32756: no value in the attribute form if value missing in the value map
hello
If there a value in an attribute table datafield which is not in value map, then the value not visible in the attribute form if i click the feature with the info tool.

tested with qgis 3.11 an 3.4

value_map

elpaso:
Fixes #32756

### DIFF
--- a/src/gui/editorwidgets/qgsvaluemapwidgetwrapper.cpp
+++ b/src/gui/editorwidgets/qgsvaluemapwidgetwrapper.cpp
@@ -79,7 +79,7 @@ void QgsValueMapWidgetWrapper::updateValues( const QVariant &value, const QVaria
 
   if ( mComboBox )
   {
-    if ( mComboBox->findData( v ) == -1 )
+    if ( mComboBox->findData( v ) == -1 && !( v.startsWith( '(' ) && v.endsWith( ')' ) ) )
     {
       mComboBox->addItem( v.prepend( '(' ).append( ')' ), v );
     }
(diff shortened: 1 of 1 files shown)


nyalldawson (approved):

src/gui/editorwidgets/qgsvaluemapwidgetwrapper.cpp
```diff
@@ -78,5 +78,11 @@ void QgsValueMapWidgetWrapper::updateValues( const QVariant &value, const QVaria
     v = value.toString();
 
   if ( mComboBox )
+  {
+    if ( mComboBox->findData( v ) == -1 && !( v.startsWith( '(' ) && v.endsWith( ')' ) ) )
```
m-kuhn, line 82: @elpaso 
Looking at your latest ValueMap pr I stumbled upon this.
What is this check for?

If I'm reading this correctly, `v` refers to the *stored* value / userData, while parentheses are added to the *represented* value.

If I'm reading this correctly, an unmatched entry `(none)` will currently not show up in the list. I think it should show up as `((none))` (or similar).

elpaso: IIRC this was added to avoid double addition of `()`, and yes, I know it's not robust.

m-kuhn: This code smells.
It's probably covering up an issue somewhere else where value/key are mixed up.

elpaso: I agree ... I'll see if I can reproduce the double `()` issue without that check.

m-kuhn: Thanks !

elpaso: https://github.com/qgis/QGIS/pull/37047

PR #36813, opened 2020-05-28
Several cppcheck related fixes

rouault:
* if (this != &other) in copy constructors
* missing initializations of member variables
* missing copy of member variables in copy constructor
* a couple other issues


src/analysis/interpolation/DualEdgeTriangulation.cpp
```diff
@@ -2915,25 +2915,20 @@ bool DualEdgeTriangulation::swapEdge( double x, double y )
 QList<int> *DualEdgeTriangulation::getPointsAroundEdge( double x, double y )
 {
   QgsPoint p( x, y, 0 );
-  int p1, p2, p3, p4;
-  int edge1 = baseEdgeOfTriangle( p );
+  const int int edge1 = baseEdgeOfTriangle( p );
```
nyalldawson, line 2918:
```suggestion
  const int edge1 = baseEdgeOfTriangle( p );
```

src/core/geometry/qgsgeometrycollection.cpp
```diff
@@ -36,7 +36,8 @@ QgsGeometryCollection::QgsGeometryCollection()
   mWkbType = QgsWkbTypes::GeometryCollection;
 }
 
-QgsGeometryCollection::QgsGeometryCollection( const QgsGeometryCollection &c ): QgsAbstractGeometry( c )
+QgsGeometryCollection::QgsGeometryCollection( const QgsGeometryCollection &c ):
+  QgsAbstractGeometry( c ), mHasCachedValidity( false ), mValidityFailureReason()
```
nyalldawson, line 40: we could copy the cached validity flag/reason here too

src/analysis/interpolation/DualEdgeTriangulation.cpp
rouault: fixed

src/core/geometry/qgsgeometrycollection.cpp
rouault: makes sense. Done

nyalldawson (approved):

PR #36817, opened 2020-05-29
Some tweaks to the Obstacle Settings dialog

DelazJ:
Add title, a help button and fix typo

obstaclesSettinsg
Not sure whether that was the most appropriate place to add the help button

Also I found that there's an issue in the dialog in that the second paragraph can be totally hidden whie resizing the dialog (not fixed here)


nyalldawson (approved):

nyalldawson (changes requested):
Actually the widget applies to much more then just the obstacle settings, so it's not safe to link in the help direct there. 

We'll need a method in the base widget class to access the button box, and then the help code can be moved to the correct subclass. Let me know if you need assistance with that!

nyalldawson:
Superseded by https://github.com/qgis/QGIS/pull/36875

PR #36819, opened 2020-05-29
Reduce number of expression evaluations in legend rendering

nyalldawson:
...by caching previous results and using them, until the layout is refreshed

alternative approach to https://github.com/qgis/QGIS/pull/34476


nyalldawson:
@roya0045 this is an alternative approach to https://github.com/qgis/QGIS/pull/34476 -- I think it achieves the same results but in a considerably simpler way. Can you test this and confirm? Thanks!

roya0045:
@nyalldawson this seems fine. I think that the simplest way to test this is to use an aggregate expression in the legend with a big layer. If you can simply perform this type of test I am fine with it. 

I'll try to merge this with an old branch with a functionnal MXE action to asses this. Can't promise anything these days.

nyalldawson:
@roya0045 ok we now discard cached values on map extent changes too!

roya0045:
@nyalldawson great thanks! If I could I would approve this.

roya0045:
Was there also a plan to propose an alternative to my PR enabling symbol level aggregate operations?

nyalldawson:
> Was there also a plan to propose an alternative to my PR enabling symbol level aggregate operations?

It's on my list! This PR queue is just so long and so few working actively on it... :sob:

roya0045:
@nyalldawson Greetings Nyall, since MXE is alive again I had the chance to test out this PR with my PR in limbo in it. And sadly this just seems to break things. 

Here are my observations:

- Adding a legend expression doesn't change anything when cycling through the atlas, the original text is still displayed.
- Manually changing a label seems to cause a back and forth between the expression and the current label.
- Upon refresh the labels are displayed and the expression ignored.

The goal of my PR was to reduce the cost of multiple evaluations but now the expressions are sometimes evaluated and displayed intermittently in the item properties before reverting to the basic label.

This is the case with basic labels and costly aggregate labels.

Sorry it took me so long to give proper feedback on this, I can open a ticket and make a gif if needed.

roya0045:
Did you had the time to test with expressions? I think that clearing the cache when an expression is provided or modified. And possibly on atlas change since that did not seem to have any effect as the expressions were never displayed.

nyalldawson:
@roya0045 yes, I tested with expressions. Do you have a sample project you could share?

roya0045:
Just did a minimal test at home with the same mxe build and a minimal project, and it works fine. The values in the map update properly but the values in the legend item have some delays. I'll have to restest things further tomorrow to see if its not a side effect of something.

roya0045:
I rechecked and it seems that only my old projects are problematic when using the new thing where old expressions were set.

I'll try to make something with non-confidential data to validate this effect.

roya0045:
[curved_polys.zip](https://github.com/qgis/QGIS/files/4788591/curved_polys.zip)
I could reproduce the behavior with this, data defined legend seem to be causing this. Take note that the MXE build that I am using contains my PR, but I don't see how this would affect the legend as the variable isn't even used.

To validate you should see the wrong text (the default value) when changing the expression for a layer in the legend. 

Hope this helps.

roya0045:
Ok I'm having a more thurrough look at it now that I have some time and I feel like this approach doesn't fully replicate some changes and leave some elements behind.

With a multi-symbol layer, there is no caching, only when there is a single symbol apparently, as stated in my other posts, this doesn't delay exports either, meaning that the evaluation will still occur after the file might be exported. This is why I feel like my proposed PR or something similar is still needed to properly control exports minimally. As stated I'd be glad to reopen the PR.

Issue #36205, opened 2020-05-05
DB Manager does not show anymore query results when there are duplicated field names


gioman:
QGIS master, Ubuntu 18.04

Seems to happen when in the SELECT statement there are >1 table.

![Peek 2020-05-05 14-37](https://user-images.githubusercontent.com/1951107/81072395-ffdfe000-8edd-11ea-9f9c-8472c1d87548.gif)

Gustry:
Remove the `;` and it should work. It' a duplicate of https://github.com/qgis/QGIS/issues/34731 ?

gioman:
@Gustry with or without ";" does not make any difference.

GIS-enthusiast:
I noticed something similar two weeks ago; the SELECT statement results show when running a query in Geopackage or SpatiaLite in DB Manager but not for PostGIS (running Q3.12 on Windows).

espinafre:
This is what I get on a debug build (to reproduce with @gioman 's example query, the geometry column must be of Polygon type, and there must be overlapping polygons):

```
src/core/qgsmessagelog.cpp:29 : (logMessage) [0ms] [thread:0x7f9ba0005790] 2020-05-11T20:39:49 PostGIS[1] Duplicate field id found
src/core/qgsmessagelog.cpp:29 : (logMessage) [0ms] [thread:0x7f9ba0005790] 2020-05-11T20:39:49 PostGIS[1] PostgreSQL layer has no primary key.
```
I was going to suggest adding `row_number() over() as ctid`  to your SELECT fields as a workaround, but it doesn't work either.

What gets sent to the database is:

```sql
SELECT * FROM (SELECT row_number() OVER () AS __rid__, * FROM (select row_number() over() as ctid, * from test_gen_col g1, test_gen_col g2 where st_overlaps(g1.geom, g2.geom) and g1.id <> g2.id) as foo) AS "subQuery_0" LIMIT 0;

```
(my testing table is called "test_gen_col") even with the row_number(). This is what I caught on the PostgreSQL log. The key here is the LIMIT 0. This is worth investigating.

espinafre:
Ha! @gioman , try changin your query a little:

```sql
select
	g1.*
from
	test_gen_col g1,
	test_gen_col g2
where
	st_overlaps(g1.geom, g2.geom)
	and g1.id <> g2.id
```

That is, use g1.* instead of simply * on your list of SELECTed fields, otherwise the returned dataset will have repeated duplicated field names, which will make the provider choke.

[...]

PR #36823, opened 2020-05-29
DB Manager: remove semicolon from SQL window queries

elpaso:
Partial fix for #36205

### DIFF
--- a/python/plugins/db_manager/dlg_sql_window.py
+++ b/python/plugins/db_manager/dlg_sql_window.py
@@ -653,7 +653,9 @@ class DlgSqlWindow(QWidget, Ui_Dialog):
         return sql
 
     def _getExecutableSqlQuery(self):
-        sql = self._getSqlQuery()
+        sql = self._getSqlQuery().strip()
+        if sql.endswith(';'):
+            sql = sql[:-1]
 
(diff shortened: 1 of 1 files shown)


python/plugins/db_manager/dlg_sql_window.py
```diff
@@ -653,7 +653,9 @@ def _getSqlQuery(self):
         return sql
 
     def _getExecutableSqlQuery(self):
-        sql = self._getSqlQuery()
+        sql = self._getSqlQuery().strip()
```
m-kuhn, line 656:
```suggestion
        sql = self._getSqlQuery().strip(' \n\r;')
```

Would that help when there's a trailing space/newline?

e.g.

```
select *
from x;

```

or do the check on the uncommented sql (which removes newlines)

elpaso: I'm fixing another issue here. Please file another ticket.

m-kuhn: My proposal aimed at a more generic approach for *exactly* this issue and nothing else.
I'll open a follow up pull request.

elpaso: Ok, thank you!

PR #36827, opened 2020-05-29
Fix windows newlines replacement in db manager query layer

### LINKED ISSUE #36825: gpkg query layers not working with line break after select
When a query is written in db manager against a GeoPackage and there is a line break directly after the `select`, the preview of the query loads fine. But if it's loaded as a query layer it's empty (no rows) and the column headers do not correspond to the selected fields (they seem to be taken from a different table in the same gpkg).

Broken query:

```
select
*
from
boflaeche;
```

Working query:

```
select *
from
boflaeche;
```

Affects Windows, could not reproduce on Linux.
QGIS 3.10.6 and 3.12.3

m-kuhn:
Fixes #36825

Tentative fix, awaiting confirmation from affected windows user.

### DIFF
--- a/python/plugins/db_manager/dlg_sql_window.py
+++ b/python/plugins/db_manager/dlg_sql_window.py
@@ -77,7 +77,7 @@ import re
 
 def check_comments_in_sql(raw_sql_input):
     lines = []
-    for line in raw_sql_input.split('\n'):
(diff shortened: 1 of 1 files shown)


m-kuhn:
Confirmed

PR #36828, opened 2020-05-29
Address review comment from #36823

m-kuhn:
## Description

A smaller and slighty more generic version of #36823 .


elpaso (approved):
That's more elegant than mine but basically the same functionality.

And what about the strip ( \n\r ) of your original proposal (that I didn't find related), did you find a better way?

m-kuhn:
It's more generic than the original version, as it also handles semicolons that are followed by whitespace or comments.

It implements the following part of the original comment 

> or do the check on the uncommented sql (which removes newlines)

PR #36835, opened 2020-05-29
[feature] Add native Cell statistics algorithm

root676:
## Description

_Something for 3.16..._

This PR adds the  **Cell statistics** raster analysis algorithm. It performs cell-wise statistics on a stack of raster layers while accounting for NoData. The functions available in the current implementation are:

- Sum
- Mean
- Median
- Standard Deviation
- Variance
- Maximum
- Minimum
- Majority (most frequent value)
- Minority (least frequent value)
- Range (max - min) 
- Variety (count of unique values)

The current implementation adds the statistic functions to the `QgsRasterAnalysisUtils` namespace so that they become available for future algorithms that might compute statistics on a vector of cell values. 

The algorithm is a 1:1 port of the ArcGIS Spatial Analyst tool [Cellstatistics](https://desktop.arcgis.com/en/arcmap/10.7/tools/spatial-analyst-toolbox/cell-statistics.htm). 

As always, I am looking forward to reviews/comments/suggestions/improvements!


nyalldawson (commented):
Wow, great addition! This will be very useful!

src/analysis/processing/qgsalgorithmcellstatistics.cpp
```diff
@@ -0,0 +1,309 @@
...
+             << QStringLiteral( "Mean" )
+             << QStringLiteral( "Median" )
+             << QStringLiteral( "Standard deviation" )
+             << QStringLiteral( "Variance" )
+             << QStringLiteral( "Minimum" )
+             << QStringLiteral( "Maximum" )
+             << QStringLiteral( "Minority" )
+             << QStringLiteral( "Majority" )
+             << QStringLiteral( "Range" )
+             << QStringLiteral( "Variety" );
```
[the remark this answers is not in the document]
root676: This is a very good idea! I'll add the count option.

root676:
@nyalldawson could you have a quick glance at the short help string and check if my statement on nearest neighbor resampling is correct? When I did an in depth code reading of the `QgsRasterDataProvider::block()` method, I was not sure whether the method applies nearest neighbor resampling when retrieving a block that doesn't perfectly fit the given extent, width and height arguments. This method was some of the heavier ones to read... :sweat_smile: If it is nearest neighbor resampling applies, I would suggest adding this detail in the API docs.

nirvn:
Oh yeah! This was one crucial raster tool we were missing.

Saijin-Naib:
This is AMAZING. You are just knocking these out of the park, one after the other!

root676:
Thank you all! @Saijin-Naib there is more to come! :wink:

src/analysis/processing/qgsalgorithmcellstatistics.cpp
```diff
@@ -0,0 +1,309 @@
...
+  int statisticMethodIdx = parameterAsInt( parameters, QStringLiteral( "STATISTIC" ), context );
+
+  const QString outputFile = parameterAsOutputLayer( parameters, QStringLiteral( "OUTPUT" ), context );
+  QFileInfo fi( outputFile );
+  const QString outputFormat = QgsRasterFileWriter::driverForExtension( fi.suffix() );
+
+  std::unique_ptr< QgsRasterFileWriter > writer = qgis::make_unique< QgsRasterFileWriter >( outputFile );
+  writer->setOutputProviderKey( QStringLiteral( "gdal" ) );
+  writer->setOutputFormat( outputFormat );
+  std::unique_ptr<QgsRasterDataProvider > provider( writer->createOneBandRaster( mDataType, mLayerWidth, mLayerHeight, mExtent, mCrs ) );
```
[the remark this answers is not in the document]
root676: Now, the ArcGIS logic is implemented here too.

src/analysis/processing/qgsalgorithmcellstatistics.cpp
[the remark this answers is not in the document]
root676: Implemented

src/analysis/processing/qgsalgorithmcellstatistics.cpp
```diff
@@ -0,0 +1,309 @@
...
+  if ( !provider )
+    throw QgsProcessingException( QObject::tr( "Could not create raster output: %1" ).arg( outputFile ) );
+  if ( !provider->isValid() )
+    throw QgsProcessingException( QObject::tr( "Could not create raster output %1: %2" ).arg( outputFile, provider->error().message( QgsErrorMessage::Text ) ) );
+
+  provider->setNoDataValue( 1, mNoDataValue );
```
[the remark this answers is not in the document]
root676: The NoData value gets set in the header file.

src/analysis/processing/qgsalgorithmcellstatistics.cpp
```diff
@@ -0,0 +1,309 @@
...
+  int iterCols = 0;
+  int iterRows = 0;
+  QgsRectangle blockExtent;
+  std::unique_ptr< QgsRasterBlock > outputBlock;
+  while ( outputIter.readNextRasterPart( 1, iterCols, iterRows, outputBlock, iterLeft, iterTop, &blockExtent ) )
+  {
+    std::vector< std::unique_ptr< QgsRasterBlock > > inputBlocks;
+    for ( const QgsRasterAnalysisUtils::RasterLogicInput &i : mInputs )
+    {
+      for ( int band : i.bands )
```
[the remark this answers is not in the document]
root676: Implemented

src/analysis/processing/qgsalgorithmcellstatistics.cpp
```diff
@@ -0,0 +1,309 @@
...
+      for ( int col = 0; col < iterCols; col++ )
+      {
+        double result = 0;
+        bool hasNoData = false;
+        std::vector<double> cellValues = getCellValues( inputBlocks, row, col, hasNoData );
+        int cellValueStackSize = cellValues.size();
+
+        if ( hasNoData )
+        {
+          //output cell will always be NoData if NoData occurs in cellValueStack
```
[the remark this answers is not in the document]
root676: Implemented

src/analysis/processing/qgsalgorithmcellstatistics.cpp
```diff
@@ -0,0 +1,309 @@
...
+  outputs.insert( QStringLiteral( "TOTAL_PIXEL_COUNT" ), layerSize );
+  outputs.insert( QStringLiteral( "OUTPUT" ), outputFile );
+
+  return outputs;
+}
+
+std::vector<double> QgsCellStatisticsAlgorithm::getCellValues( const std::vector< std::unique_ptr< QgsRasterBlock > > &inputBlocks, int &row, int &col, bool &hasNoData )
+{
+  //get all values from inputBlocks
+  std::vector<double> cellValues;
```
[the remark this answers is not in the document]
root676: Implemented

src/analysis/processing/qgsrasteranalysisutils.cpp
```diff
@@ -285,5 +287,166 @@ void QgsRasterAnalysisUtils::applyRasterLogicOperator( const std::vector< QgsRas
...
+  {
+    accum += std::pow( ( cellValues.at( i ) - mean ), 2.0 );
+  }
+  double stddev = std::sqrt( accum / static_cast<double>( stackSize ) );
+  return stddev;
+}
+
+double QgsRasterAnalysisUtils::varianceFromCellValues( std::vector<double> cellValues, int stackSize )
+{
+  double variance = std::pow( stddevFromCellValues( cellValues, stackSize ), 2.0 );
```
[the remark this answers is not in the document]
root676: Implemented

root676:
I'll take some more time to refactor the unit test and check if everything works as expected. I'll keep you posted.

src/analysis/processing/qgsalgorithmcellstatistics.cpp
```diff
@@ -0,0 +1,333 @@
...
+                      "   <li>Majority (least frequent value)</li>"
+                      "   <li>Range (max-min)</li>"
+                      "   <li>Variety (count of unique values)</li>"
+                      "</ul> "
+                      "Input raster layers that do not match the cell size of the reference raster layer will be "
+                      "resampled using nearest neighbor resampling. The output raster data type will be set to "
+                      "the most complex data type present in the input datasets except when using the functions "
+                      "Mean and Standard deviation (data type is always Float32) or Count and Variety (data type is always Int32).\n"
+                      "<i>Calculation details - general:</i> NoData values in any of the input layers will result in a NoData cell output if the Ignore NoData parameter is not set.\n"
+                      "<i>Calculation details - Count:</i> Count will always result in the number of cells without NoData values at the current cell location."
```
nyalldawson, line 78:
```suggestion
                      "<i>Calculation details - Count:</i> Count will always result in the number of cells without NoData values at the current cell location.\n"
```

src/analysis/processing/qgsalgorithmcellstatistics.cpp
```diff
@@ -0,0 +1,333 @@
...
+{
+  return QStringLiteral( "rasteranalysis" );
+}
+
+QString QgsCellStatisticsAlgorithm::shortHelpString() const
+{
+  return QObject::tr( "The Cell statistics algorithm computes a value for each cell of the "
+                      "output raster. At each cell location, "
+                      "the output value is defined as a function of all overlaid cell values of the "
+                      "input rasters.\n "
```
nyalldawson, line 55:
```suggestion
                      "input rasters.\n\n"
```

src/analysis/processing/qgsalgorithmcellstatistics.cpp
```diff
@@ -0,0 +1,333 @@
...
+                      "The output rasters extent and resolution is defined by a reference "
+                      "raster. The following functions can be applied on the input "
+                      "raster cells per output raster cell location: "
```
nyalldawson, line 58:
```suggestion
                      "raster cells per output raster cell location:\n"
```

src/analysis/processing/qgsalgorithmcellstatistics.cpp
```diff
@@ -0,0 +1,333 @@
...
+  return new QgsCellStatisticsAlgorithm();
+}
+
+void QgsCellStatisticsAlgorithm::initAlgorithm( const QVariantMap & )
+{
+  addParameter( new QgsProcessingParameterMultipleLayers( QStringLiteral( "INPUTS" ),
+                QObject::tr( "Input layers" ), QgsProcessing::TypeRaster ) );
+
+  QStringList statistics = QStringList();
+  statistics << QStringLiteral( "Sum" )
```
nyalldawson, line 96:
```suggestion
  statistics << QObject::tr( "Sum" )
```
(also for the others)

src/analysis/processing/qgsalgorithmcellstatistics.cpp
```diff
@@ -0,0 +1,333 @@
...
+    for ( int band : i.bands )
+    {
+      Qgis::DataType inputDataType = i.interface->dataType( band );
+      if( static_cast<int>(mDataType) < static_cast<int>(inputDataType) )
+        mDataType = inputDataType; //if raster data type is more potent, set it as new data type
+    }
+  }
+
+  //force data types on specific functions if input data types don't match
+  if( statisticMethodIdx == 2 || statisticMethodIdx == 4 ) //mean, stddev
```
nyalldawson, line 189: Would be cleaner to make an enum for the statistic calculation types

src/analysis/processing/qgsalgorithmcellstatistics.cpp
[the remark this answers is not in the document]
nyalldawson: Shouldn't we expose this choice to users though? (as an advanced parameter). There's always a chance that -9999 isn't appropriate in particular circumstances.

src/analysis/processing/qgsalgorithmcellstatistics.cpp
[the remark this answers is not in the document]
nyalldawson: Was this change pushed? I can't see it here

src/analysis/processing/qgsrasteranalysisutils.cpp
```diff
@@ -285,5 +287,196 @@ void QgsRasterAnalysisUtils::applyRasterLogicOperator( const std::vector< QgsRas
...
+
+
+double QgsRasterAnalysisUtils::stddevFromCellValues( std::vector<double> cellValues, int stackSize )
+{
+  double mean = meanFromCellValues( cellValues, stackSize );
+  double accum = 0.0;
+  for ( int i = 0; i < stackSize; i++ )
+  {
+    accum += std::pow( ( cellValues.at( i ) - mean ), 2.0 );
+  }
```
nyalldawson, line 344: You could replace this with a call to varianceFromCellValues(...)

nyalldawson:
@root676 
> could you have a quick glance at the short help string and check if my statement on nearest neighbor resampling is correct? When I did an in depth code reading of the QgsRasterDataProvider::block() method, I was not sure whether the method applies nearest neighbor resampling when retrieving a block that doesn't perfectly fit the given extent, width and height arguments. This method was some of the heavier ones to read... sweat_smile If it is nearest neighbor resampling applies, I would suggest adding this detail in the API docs.

It's correct-ish. :laughing: It depends on the actual provider of the raster layer -- some will do different resampling internal on the backend (e.g. a wms layer will be resampled on the server, so we don't really know what's going on there!). 

For gdal based layers then yes -- it's a quite poorly written nearest neighbour calculation. See the thread at https://lists.osgeo.org/pipermail/qgis-developer/2020-April/060926.html for all the gory details.

src/analysis/processing/qgsalgorithmcellstatistics.cpp
root676: Done

src/analysis/processing/qgsalgorithmcellstatistics.cpp
root676: Done

src/analysis/processing/qgsalgorithmcellstatistics.cpp
root676: Done

src/analysis/processing/qgsalgorithmcellstatistics.cpp
root676: Implemented

src/analysis/processing/qgsalgorithmcellstatistics.cpp
root676: Implemented

src/analysis/processing/qgsrasteranalysisutils.cpp
root676: Done

root676:
@nyalldawson I think I could adress all comments in the latest commits. I must have missed to push the cancel check last time.

nirvn:
@root676 , is this ready to merge?

root676:
@nirvn form my side, everything is implemented and polished. @nyalldawson would you like to do a last review round or should we merge?

nyalldawson:
@root676 the new cpp check is flagging this PR -- can you address this?

root676:
@nyalldawson the failing test should be fixed now. I don't know why the macOS build fails right now but it doesn't seem related to the PR.

src/analysis/processing/qgsalgorithmcellstatistics.cpp
```diff
@@ -0,0 +1,337 @@
...
+}
+
+QgsCellStatisticsAlgorithm *QgsCellStatisticsAlgorithm::createInstance() const
+{
+  return new QgsCellStatisticsAlgorithm();
+}
+
+void QgsCellStatisticsAlgorithm::initAlgorithm( const QVariantMap & )
+{
+  addParameter( new QgsProcessingParameterMultipleLayers( QStringLiteral( "INPUTS" ),
```
nyalldawson, line 92:
```suggestion
  addParameter( new QgsProcessingParameterMultipleLayers( QStringLiteral( "INPUT" ),
```

Can you rename this parameter as "INPUT" to match the quasi-standard from other algorithms?

src/analysis/processing/qgsalgorithmcellstatistics.cpp
```diff
@@ -0,0 +1,337 @@
...
+             << QObject::tr( "Maximum" )
+             << QObject::tr( "Minority" )
+             << QObject::tr( "Majority" )
+             << QObject::tr( "Range" )
+             << QObject::tr( "Variety" );
+
+  addParameter( new QgsProcessingParameterEnum( QStringLiteral( "STATISTIC" ), QObject::tr( "Statistic" ),  statistics, false, 0, false ) );
+  addParameter( new QgsProcessingParameterBoolean( QStringLiteral( "IGNORE_NODATA" ), QObject::tr( "Ignore NoData values" ), true ) );
+
+  addParameter( new QgsProcessingParameterRasterLayer( QStringLiteral( "REF_LAYER" ), QObject::tr( "Reference layer" ) ) );
```
nyalldawson, line 112:
```suggestion
  addParameter( new QgsProcessingParameterRasterLayer( QStringLiteral( "REFERENCE_LAYER" ), QObject::tr( "Reference layer" ) ) );
```

And let's avoid the abbreviation here -- it makes the api trickier for non english native speakers to understand

src/analysis/processing/qgsalgorithmcellstatistics.cpp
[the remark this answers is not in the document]
nyalldawson: This comment is still outstanding

nyalldawson:
Just some minor changes left to do!

src/analysis/processing/qgsalgorithmcellstatistics.cpp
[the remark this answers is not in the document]
root676: Now it should be fixed. :)

root676:
Thank you for the review @nyalldawson - all issues should be fixed by now. Lets see what the tests say.

nyalldawson (approved):

nyalldawson:
Thanks @root676 -- this is a GREAT feature!

PedroVenancio:
@root676 Really great feature Clemens!

Do you think it is feasible Cell Statistics to calculate the nth percentile of each cell? I think it would be a great add for the future! But it would require another parameter (nth percentile) in the interface. Maybe another tool for that?

Thank you very much for Cell Statistics Clemens!

root676:
@PedroVenancio Thanks! :smile: 
I think that this is a great Idea but as you mentioned it, we maybe should put it into a separate algorithm. This indeed would require another parameter which is exclusively dedicated to calculating percentiles. The other cellstatistics functions wouldn't make any use of it. 

Any alg name suggestions? :wink:

PedroVenancio:
Hi @root676 ! 

Hum, now I was thinking that the new algorithm could be two approaches! :)

- One is that we were talking about, that is, we give a stack of raster layers and the nth percentile to be calculated for each cell, and the output is the value for each cell that cuts off the first n percent of the data values when it is sorted in ascending order;

- Another is the opposite, we give a stack of raster layers and, in this case, a "reference" layer, and the output is, for each cell, the nth percentile of each cell of "reference" layer, compared to stack of raster layers. 

So, in the first option, the output is, for each cell, the value of a given nth percentile; in the second option, the output is the nth percentile of each cell of the reference layer.

I really don't know if this fit only in one algorithm, or it should be two different! To make thinks easier, maybe two?! :)

root676:
Hi @PedroVenancio ! 
Good idea but in this case I would boil it down to your algorithm example that uses a "percentile raster" for calculating the percentile of a raster stack. If you want to use a constant value for calculation you only have to create a constant raster which is very easy to do. This keeps the alg list compact and you can use the algorithm in both ways. What do you think?

PedroVenancio:
Hi @root676 !

The idea was to get something like the LibreOffice Calc / MS Excel functions:

- **PERCENTILE**: https://help.libreoffice.org/latest/en-US/text/scalc/01/04060184.html?DbPAR=CALC#bm_id3153100

percentile

The output is, for each cell, the value of the given percentile. The reference layer would be exactly as in actual Cell Statistics, giving the extent and resolution of the output.


- **PERCENTRANK**: https://help.libreoffice.org/latest/en-US/text/scalc/01/04060184.html?DbPAR=CALC#bm_id3148807

percentrank

The output is the percentage rank of each pixel value of reference layer, given the correspondent pixel values in the raster stack. The reference layer here is the base for the calculations, giving the VALUE parameter of PERCENTRANK algorithm, besides the extent and resolution of the output.

Do you think this makes sense @root676 ?

Best regards!

root676:
@PedroVenancio thanks for the examples! Now I understood you! Lets see what I can do. :wink:

PedroVenancio:
Great @root676 ! Thank you very much!!

PedroVenancio:
Hi @root676 

Sorry to be posting this here, it's just to confirm if you think this is a bug. Actually they are two things:

 1. The output of Cell Statistics does not save the Coordinate System in the header of the file, and also does not create a world file:

```
gdalinfo Cell_Statistics_Median.tif
Driver: GTiff/GeoTIFF
Files: Cell_Statistics_Median.tif
Size is 17, 14
Origin = (79784.960900000005495,133511.328100000013364)
Pixel Size = (1000.000000000000000,-1000.000000000000000)
Image Structure Metadata:
  INTERLEAVE=BAND
Corner Coordinates:
Upper Left  (   79784.961,  133511.328)
Lower Left  (   79784.961,  119511.328)
Upper Right (   96784.961,  133511.328)
Lower Right (   96784.961,  119511.328)
Center      (   88284.961,  126511.328)
Band 1 Block=17x14 Type=Int16, ColorInterp=Gray
  NoData Value=-9999
```

2. "Median: If the number of input layers is even, the median will be calculated as the arithmetic mean of the two middle values of the ordered cell input values." 
However, I've tested with 10 layers and Median creates an Integer raster (Int16), with wrong results. Take a look for instance for this pixel:

imagem

Median should be 14.5. 

Here is a sample dataset (simply 10 rasters with 1000m spatial resolutin and random values between 0 and 25): 
[Cell_Statistics.zip](https://github.com/qgis/QGIS/files/5491103/Cell_Statistics.zip)

If you confirm, I can open two separate tickets.

Thank you very much @root676 !

root676:
@PedroVenancio thanks for letting me know - I will have a look into it after work.

root676:
@PedroVenancio Alright - while I cannot confirm the issue with the output CRS there is definitely something wrong with the median function. You have to make sure that the reference raster you use has a proper CRS so it can be passed on to the output layer. I tested this for 3 random rasters in `EPSG:31255 - MGI / Austria GK Central - Projected`. This is the output of `gdalinfo`: 

```
gdalinfo root676_median_test_output.tif 
Driver: GTiff/GeoTIFF
Files: root676_median_test_output.tif
Size is 35, 31
Coordinate System is:
PROJCRS["MGI / Austria GK Central",
    BASEGEOGCRS["MGI",
        DATUM["Militar-Geographische Institut",
            ELLIPSOID["Bessel 1841",6377397.155,299.1528128,
                LENGTHUNIT["metre",1]]],
        PRIMEM["Greenwich",0,
            ANGLEUNIT["degree",0.0174532925199433]],
        ID["EPSG",4312]],
    CONVERSION["Austria Gauss-Kruger Central",
        METHOD["Transverse Mercator",
            ID["EPSG",9807]],
        PARAMETER["Latitude of natural origin",0,
            ANGLEUNIT["degree",0.0174532925199433],
            ID["EPSG",8801]],
        PARAMETER["Longitude of natural origin",13.3333333333333,
            ANGLEUNIT["degree",0.0174532925199433],
            ID["EPSG",8802]],
        PARAMETER["Scale factor at natural origin",1,
            SCALEUNIT["unity",1],
            ID["EPSG",8805]],
        PARAMETER["False easting",0,
            LENGTHUNIT["metre",1],
            ID["EPSG",8806]],
        PARAMETER["False northing",-5000000,
            LENGTHUNIT["metre",1],
            ID["EPSG",8807]]],
    CS[Cartesian,2],
        AXIS["northing (X)",north,
            ORDER[1],
            LENGTHUNIT["metre",1]],
        AXIS["easting (Y)",east,
            ORDER[2],
            LENGTHUNIT["metre",1]],
    USAGE[
        SCOPE["unknown"],
        AREA["Austria - 11°50'E to 14°50'E"],
        BBOX[46.4,11.83,48.79,14.84]],
    ID["EPSG",31255]]
Data axis to CRS axis mapping: 2,1
Origin = (-165968.898900000000140,249038.294599999993807)
Pixel Size = (1000.000000000000000,-1000.000000000000000)
Metadata:
  AREA_OR_POINT=Area
Image Structure Metadata:
  INTERLEAVE=BAND
Corner Coordinates:
Upper Left  ( -165968.899,  249038.295) ( 11d 8' 9.76"E, 47d21'34.75"N)
Lower Left  ( -165968.899,  218038.295) ( 11d 8'51.12"E, 47d 4'51.55"N)
Upper Right ( -130968.899,  249038.295) ( 11d35'57.00"E, 47d22' 3.35"N)
Lower Right ( -130968.899,  218038.295) ( 11d36'29.66"E, 47d 5'19.87"N)
Center      ( -148468.899,  233538.295) ( 11d22'21.86"E, 47d13'28.22"N)
Band 1 Block=35x31 Type=Float32, ColorInterp=Gray
```
I also checked with your data - I get the output for `EPSG:3763 - ETRS89 / Portugal TM06 - Projected` - so this should be ok. 

Regarding the median issue - pls copy the report to a new Issue and reference this PR conversation. I'll have a look into the median method to fix it. Thanks for reporting!! :wink:

PedroVenancio:
> Alright - while I cannot confirm the issue with the output CRS there is definitely something wrong with the median function. You have to make sure that the reference raster you use has a proper CRS so it can be passed on to the output layer. 

You're right @root676 I double checked now and it is fine. I can't remember exactly the reference layers I've used yesterday, so it's possible that could be some GRASS alg outputs without CRS info. Sorry for the noise. 

> Regarding the median issue - pls copy the report to a new Issue and reference this PR conversation. I'll have a look into the median method to fix it. Thanks for reporting!! 

I will do it ASAP! :)

I will also open a feature request for the PERCENTILE / PERCENTRANK ideas for the future, also linking to this PR!

Thank you very much Clemens!

PR #36836, opened 2020-05-29
Another batch of cppcheck fixes

rouault:
A number of the "declare deleted copy constructor and assignment operators as the class has pointer member variables" are not strictly needed, as they are for classes that derived from QObject that deletes those operators, but cppcheck isn't smart enough to detect that.

### DIFF
--- a/src/core/auth/qgsauthmanager.cpp
+++ b/src/core/auth/qgsauthmanager.cpp
@@ -84,6 +84,10 @@ const QString QgsAuthManager::AUTH_PASSWORD_HELPER_DISPLAY_NAME( "Password Manag
 
 QgsAuthManager *QgsAuthManager::instance()
 {
+  // NOTE cppcheck complains about identicalInnerCondition. Pedantically
+  // the below construct is not totally thread-safe.
+  // See https://preshing.com/20130930/double-checked-locking-is-fixed-in-cpp11/
+  // regarding double-checked locking pattern (DCLP)
   if ( !sInstance )
   {
     static QMutex sMutex;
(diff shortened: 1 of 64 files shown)


src/core/auth/qgsauthmanager.cpp
```diff
@@ -84,6 +84,10 @@ const QString QgsAuthManager::AUTH_PASSWORD_HELPER_DISPLAY_NAME( "Password Manag
 
 QgsAuthManager *QgsAuthManager::instance()
 {
+  // NOTE cppcheck complains about identicalInnerCondition. Pedantically
```
nyalldawson, line 87: Could we replace this with a std::once?

src/core/geometry/qgsgeometry.cpp
```diff
@@ -80,9 +80,8 @@ QgsGeometry::QgsGeometry( std::unique_ptr<QgsAbstractGeometry> geom )
   d->ref = QAtomicInt( 1 );
 }
 
-QgsGeometry::QgsGeometry( const QgsGeometry &other )
+QgsGeometry::QgsGeometry( const QgsGeometry &other ): d( other.d )
```
nyalldawson, line 83: Pedantic, but qgis convention is to put the : on a new line

nyalldawson (approved):

src/core/auth/qgsauthmanager.cpp
rouault: That would certainly be the most simple solution to make that correct compared to the other alternatives with the memory ordering hints. There's a small overhead though when using std::once . Not sure how it would compared to just removing the initial check and going through the mutex systematically

src/core/geometry/qgsgeometry.cpp
rouault: done

src/core/auth/qgsauthmanager.cpp
nyalldawson: Fair point! In this case no performance critical code should be interacting with the auth manager, so it would be safe to use here. But whatever's easiest!

rouault: changed to systematically going through the mutex

nyalldawson:
Looks good!

PR #36853, opened 2020-05-31
disabled contour/vector checkbox when rendered type is disabled

vcloarec:
For mesh layer, when contour or vector rendering are disabled, in the styling dialog, entire group boxes are disabled (event the check box to enable/disable). With this PR, only inside the group box is disabled.

### DIFF
--- a/src/app/mesh/qgsrenderermeshpropertieswidget.cpp
+++ b/src/app/mesh/qgsrenderermeshpropertieswidget.cpp
@@ -168,7 +168,6 @@ void QgsRendererMeshPropertiesWidget::onActiveScalarGroupChanged( int groupIndex
   mMeshRendererScalarSettingsWidget->setActiveDatasetGroup( groupIndex );
   mMeshRendererScalarSettingsWidget->syncToLayer();
   mContoursGroupBox->setChecked( groupIndex >= 0 );
-  mContoursGroupBox->setEnabled( groupIndex >= 0 );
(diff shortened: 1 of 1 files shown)


src/app/mesh/qgsrenderermeshpropertieswidget.cpp
```diff
@@ -168,7 +168,6 @@ void QgsRendererMeshPropertiesWidget::onActiveScalarGroupChanged( int groupIndex
   mMeshRendererScalarSettingsWidget->setActiveDatasetGroup( groupIndex );
   mMeshRendererScalarSettingsWidget->syncToLayer();
   mContoursGroupBox->setChecked( groupIndex >= 0 );
-  mContoursGroupBox->setEnabled( groupIndex >= 0 );
```
PeterPetrik, line 171: if `groupIndex >= 0` it means that the user does not selected any group in the group tree widget. So I do not understand what it help if the checkbox for enabling/disabling is available. Or what it does when user enables the vector renderer for groupIndex == -1?

vcloarec: True... I have made this PR because I found inconsistent that vector/contour can be disabled from those checkbox, but could not be enable in some cases.
But since disable/enable contour or vector from the group tree widget is fully functional, maybe those check boxes were not necessary anymore?

PeterPetrik: I think it is fine as it is.... User can disable it now from both places ...

PR #36860, opened 2020-06-01
fix scientific notation in scale notation, fix #36859

### LINKED ISSUE #36859: Project custom scales are displayed with scientific notation
**Describe the bug**

Using the Project Predefined Scales feature, it appears that denominator are displayed with scientific notation ( like 1:4.5e+06). Apart the fact that I believe that scales never use scientific notation, it also causes bugs because internationalization is not fully used in this part of the code.

Consequence :
- project scales combo box is not working for scientific value
- importing new predefined scales in project properties, it will not recognized scales that are already defined to avoid to import it twice ("1:1000000" will not be equal to "1:1e+06")

**How to Reproduce**

1. Go to Project Properties > General > Project Predefined Scales
2. Import or add scale with high denominator (like 4500000), scales are displayed right.
3. Close and go back to Project Properties, 4500000 are now displayed as 4.5e+06
4. Try adding again 4500000 as denominator, it works.

And futhermore the scale combo box doesn't work in the project

**QGIS and OS versions**

QGIS version | 3.12.3-București | QGIS code revision | 8234261527
-- | -- | -- | --
Compiled against Qt | 5.12.4 | Running against Qt | 5.12.4
Compiled against GDAL/OGR | 2.4.2 | Running against GDAL/OGR | 2.4.2
Compiled against GEOS | 3.7.2-CAPI-1.11.2 | Running against GEOS | 3.7.2-CAPI-1.11.2 b55d2125
PostgreSQL Client Version | 12.3 (Ubuntu 12.3-1.pgdg19.10+1) | SpatiaLite Version | 5.0.0-beta0
OS Version | Ubuntu 19.10 | This copy of QGIS writes debugging output.
Active python plugins | processing; db_manager

SebastienPeillet:
## Description

This PR fixes bugs with project predefined scales. Using this feature, smal scale are displayed with scientific notation (see #36859). It break the use of the QgsScaleComboBox, and alter the use of the Predefined scales configuration in Project properties.

 Internationalization was not fully used which causes troubles to convert int to string and vice versa.

### DIFF
--- a/tests/src/gui/testqgsscalecombobox.cpp
+++ b/tests/src/gui/testqgsscalecombobox.cpp
@@ -41,6 +41,7 @@ class TestQgsScaleComboBox : public QObject
     void toString();
     void toDouble();
     void allowNull();
+    void testLocale();
 
   private:
     void enterScale( const QString &scale );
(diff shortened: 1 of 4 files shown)


elpaso (approved):

tests/src/gui/testqgsscalecombobox.cpp
```diff
@@ -256,5 +257,33 @@ void TestQgsScaleComboBox::cleanup()
   delete s;
 }
 
+void TestQgsScaleComboBox::testLocale()
+{
+  QLocale::setDefault( QLocale::English );
+  QVERIFY( s->toString( 1e8 ) == "1:100,000,000" );
```
nyalldawson, line 263: Would be better to use QCOMPARE for these tests -- then you'd see exactly what the actual returned string is

nyalldawson:
This is failing on azure (it's not related to the current failure of the 32 bit builds either)

SebastienPeillet:
> This is failing on azure (it's not related to the current failure of the 32 bit builds either)

Yes, results are different according to Qt version. I fixed it !

Issue #36843, opened 2020-05-30
Aggregate tool not working


RNanclares:
**Describe the bug**
Aggregate tool doesn't work

**How to Reproduce**

1. Go to Toolbox
2. Click on Aggregate tool
3. Python error message pops and the tools fails to "start"
4. See error:

`Traceback (most recent call last):
              File "/usr/share/qgis/python/plugins/processing/gui/ProcessingToolbox.py", line 262, in executeAlgorithm
              dlg = AlgorithmDialog(alg, self.in_place_mode, iface.mainWindow())
              File "/usr/share/qgis/python/plugins/processing/gui/AlgorithmDialog.py", line 67, in __init__
              self.setMainWidget(self.getParametersPanel(alg, self))
              File "/usr/share/qgis/python/plugins/processing/gui/AlgorithmDialog.py", line 87, in getParametersPanel
              panel = ParametersPanel(parent, alg, self.in_place)
              File "/usr/share/qgis/python/plugins/processing/gui/ParametersPanel.py", line 70, in __init__
              self.initWidgets()
              File "/usr/share/qgis/python/plugins/processing/gui/ParametersPanel.py", line 104, in initWidgets
              wrapper = WidgetWrapperFactory.create_wrapper(param, self.parent())
              File "/usr/share/qgis/python/plugins/processing/gui/wrappers.py", line 1845, in create_wrapper
              return WidgetWrapperFactory.create_wrapper_from_metadata(param, dialog, row, col)
              File "/usr/share/qgis/python/plugins/processing/gui/wrappers.py", line 1876, in create_wrapper_from_metadata
              mod = __import__('.'.join(tokens[:-1]), fromlist=[tokens[-1]])
              File "/usr/lib/python3/dist-packages/qgis/utils.py", line 743, in _import
              mod = _builtin_import(name, globals, locals, fromlist, level)
              File "/usr/share/qgis/python/plugins/processing/algs/qgis/ui/AggregatesPanel.py", line 45, in 
              from processing.algs.qgis.ui.FieldsMappingPanel import (
             ImportError: cannot import name 'ExpressionDelegate' from 'processing.algs.qgis.ui.FieldsMappingPanel' (/usr/share/qgis/python/plugins/processing/algs/qgis/ui/FieldsMappingPanel.py)
             `

**QGIS and OS versions**
QGIS 3.13 Master
PopOS 20.04 / Ubuntu 20.04

Issue #36706, opened 2020-05-25
Using Fields Mapper parameter in Modeler throws "NotImplementedError"


gacarrillor:
Steps to reproduce: 

1. Open Processing Modeler. 
2. From the Inputs panel, drag and drop a Fields Mapper parameter to the model. 
3. Once you give the parameter a name and you click on OK, you get the error:

```
NotImplementedError
QgsProcessingParameterDefinition.type() is abstract and must be overridden

```

Using a just compiled master (2020.05.25).

nyalldawson:
Ehh... this is horrible. It's sip "forgetting" the Python subclass overrides again. I honestly can't see any obvious fix here except for porting that whole algorithm to c++.

gacarrillor:
Great! Thank you @nyalldawson!

jakimowb:
@nyalldawson do you think this means that it is generally impossible to specify a new `QgsProcessingParameterDefinition `in python?
I have noticed that the `NotImplementedError` error dissapears when keeping a python reference to a `QgsProcessingParameterDefinition`. However, closing the parameter definition widget in the graphical modeler will then raise a `wrapped C/C++ object ... has been deleted ` error in `ModelerDialog.addInput` (defined in `processing\modeler\ModelerDialog.py`) 

### How to reproduce

1. Call this from QGIS python shell:

```
from qgis.core import QgsProcessingParameterDefinition, QgsApplication, QgsProcessingRegistry, QgsProcessingParameterType
from qgis.PyQt.QtCore import QObject

REFS = []

class MyParameter(QgsProcessingParameterDefinition):
    """
    Definition of my parameter
    """
    def __init__(self, name='MyParameter', description='My Parameter', optional:bool=False):
        super().__init__(name, description=description, optional=optional)
        self.mMyValue: str = 'my_value'

        if True: # keep a python reference on the MyParameter instance
            global REFS
            REFS.append([self, super(MyParameter, self)])


    def isDestination(self):
        return False

    def type(self):
        return 'my_type'

    def clone(self):
        p = MyParameter()
        return p

    def description(self):
        return 'My Parameter Description'

    def toolTip(self):
        return 'My Parameter Tooltip'

    def toVariantMap(self):
        result = super().toVariantMap()
        result['my_value'] = self.mMyValue
        return result

    def fromVariantMap(self, map:dict):
        super().fromVariantMap(map)
        self.mMyValue = map.get('my_value', '')

        return True


[...]

PR #36864, opened 2020-06-01
Port refactor fields to C++, fix exceptions due to sip casting issues

nyalldawson:
This PR ports the Refactor Fields algorithm and widgets to C++, in order to resolve the sip subclassing issues encountered in #36706, which cause exceptions to be raised when a field mapping input is used in a model.

It also fixes other outstanding issues with the Refactor Fields algorithm/Fields Mapping parameter type, and adds many unit tests.

As a result, this high priority algorithm should be much more stable.

Also a step toward fixing https://github.com/qgis/QGIS/issues/36843, which will probably require porting the Aggregate algorithm to c++ too.


alexbruy (approved):
Looks good to me.

PR #36865, opened 2020-06-01
make QgsLocalizedPathRegistry thread safe

3nids:
this was failing the pull request, trying to reintroduce


src/core/qgslocalizeddatapathregistry.h
```diff
@@ -65,10 +66,16 @@ class CORE_EXPORT QgsLocalizedDataPathRegistry
     void unregisterPath( const QString &path );
 
   private:
+#ifdef SIP_RUN
+    QgsLocalizedDataPathRegistry( const QgsLocalizedDataPathRegistry &other )
```
m-kuhn, line 70:
```suggestion
    QgsLocalizedDataPathRegistry( const QgsLocalizedDataPathRegistry &other ) = default;
```

src/core/qgslocalizeddatapathregistry.cpp
```diff
@@ -82,6 +94,8 @@ void QgsLocalizedDataPathRegistry::registerPath( const QString &path, int positi
 
 void QgsLocalizedDataPathRegistry::unregisterPath( const QString &path )
 {
+  QgsReadWriteLocker locker( mLock, QgsReadWriteLocker::Write );
+
   mPaths.removeAll( QDir( path ) );
   writeToSettings();
```
m-kuhn, line 100: This is a potential dead lock.

1. Acquires a write lock
2. Calls `writeToSettings()`
3. Calls `paths()`
4. Tries to acquire a read lock which fails because there's a write lock in place

3nids:
thanks Matthias :)

PR #36873, opened 2020-06-01
Fix remaining cppcheck warnings, add scripts/cppcheck.sh and corresponding GH workflow jobs

rouault:
* Fix remaining cppcheck issues in the "warning" category, as spotted by cppcheck of Ubuntu 16.04 and 18.04. There are a lot in other categories that should deserve fixing...
* Add scripts/cppcheck.sh that fails if issues in the "warning"  category are raised, and reports other categories in non failing way
* Add 2 GH workflow with cppcheck of Ubuntu 16.04 and 18.04

### DIFF
--- a/.github/workflows/code_layout.yml
+++ b/.github/workflows/code_layout.yml
@@ -158,3 +158,29 @@ jobs:
         run: ./tests/code_layout/test_sip_include.sh
       - name: Sip Files Up To Date
         run: ./tests/code_layout/test_sipfiles_uptodate.sh
+
+  cppcheck_16_04:
+    runs-on: ubuntu-16.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Install Requirements
+        run: |
+          sudo apt install -y cppcheck
+
+      - name: Run cppcheck test
+        run: ./scripts/cppcheck.sh
+
+  cppcheck_18_04:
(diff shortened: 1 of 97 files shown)


nyalldawson (approved):

nyalldawson:
Question: what's the benefit in running the older cpp check from 16.04?

rouault:
> what's the benefit in running the older cpp check from 16.04?

I found cppcheck versions are not always monotonically increasing in quality. Some older versions detect issues that some later won't, and vice verca. Some might have quite annoying bugs too, but for the 'warning' category 16.04 and 18.04 are reasonable enough

nyalldawson:
> I found cppcheck versions are not always monotonically increasing in quality. Some older versions detect issues that some later won't, and vice verca. Some might have quite annoying bugs too, but for the 'warning' category 16.04 and 18.04 are reasonable enough

Ok, thanks for the clarification!

This looks great, by the way!

m-kuhn:
I've implemented these checks on QField too (thanks for the workflow files, @rouault )

FYI:
My impression was, that 18.04 (cppcheck 1.80) reports everything that 16.04 (cppcheck 1.70) reports as well - and more.
On the other hand 20.04 (cppcheck 1.90) seemed to perform worse. There we'll stick to 1.80 (18.04) only for now.

PR #36880, opened 2020-06-02
Use an exact test when checking if a curve is closed 

### LINKED ISSUE #36346: QGIS crashes while zooming in (too far...) with Labels on.
**Describe the bug**
Hi,
See points 1,2 3  etc below.
If you don't get the crash, try zooming faster.
Mine crashes at zoom level 2:1 and 4:1 (on second attempt to reproduce)

Hope this helps.
Kind regards,
Zoltan

**How to Reproduce**

1. Import 2 points with same coord value from a CSV file EPSG:4326
2. Roll mousewheel to zoom in on them, past 4:1  (see: No crash)
3. Zoom back out to a respectable level
4. Turn on labelling
5. Zoom in again, past 1:1,
3. My system caused an Uh-Oh crash at zoom scale 2:1

**QGIS and OS versions**

QGIS Info
QGIS Version: 3.12.2-Bucure?ti
QGIS code revision: 8a1fb33634
Compiled against Qt: 5.11.2
Running against Qt: 5.11.2
Compiled against GDAL: 3.0.4
Running against GDAL: 3.0.4

**Additional context**

Crash ID: e1d226ea3eb384d54cf43ce719faa0b7aacf3f2d

Stack Trace

GEOSEquals_r :
pal::Layer::registerFeature layer.cpp:211
QgsLabelingEngine::processProvider qgslabelingengine.cpp:248
QgsLabelingEngine::registerLabels qgslabelingengine.cpp:287
QgsDefaultLabelingEngine::run qgslabelingengine.cpp:549
QgsMapRendererJob::drawLabeling qgsmaprendererjob.cpp:932
QgsMapRendererParallelJob::renderLabelsStatic qgsmaprendererparalleljob.cpp:399
QtConcurrent::RunFunctionTask::run qtconcurrentrunbase.h:136
QThreadPoolPrivate::reset :
QThread::start :
BaseThreadInitThunk :
RtlUserThreadStart :

QGIS Info
QGIS Version: 3.12.2-Bucure?ti
QGIS code revision: 8a1fb33634
Compiled against Qt: 5.11.2
Running against Qt: 5.11.2
Compiled against GDAL: 3.0.4
Running against GDAL: 3.0.4

System Info
CPU Type: x86_64
Kernel Type: winnt
Kernel Version: 10.0.18362

nyalldawson:
Otherwise we get cases where calling isClosed() on a curve with small
(<1E-08) differences between the first and last vertex incorrectly
returns true.

This is an issue when converting rings to GEOS, as it rejects them
as being unclosed, yet calling QgsCurve::close on them has no effect...

This is the underlying root cause of #36346. In this particular case
the calculation of the bounds of the marker symbol (used for the
label obstacle calculation) results in a very small rectangle (since
it's in map (geographic) units, and the map is very zoomed in) -- so
the 1E-08 tolerance here is completely unsuitable and calling close()
on the marker bounds had no effect, ultimately resulting in an unclosed
ring being passed to GEOS.

Fixes #36346

### DIFF
--- a/src/core/geometry/qgscurve.cpp
+++ b/src/core/geometry/qgscurve.cpp
@@ -46,10 +46,10 @@ bool QgsCurve::isClosed() const
   QgsPoint start = startPoint();
   QgsPoint end = endPoint();
 
-  bool closed = qgsDoubleNear( start.x(), end.x(), 1E-8 ) &&
-                qgsDoubleNear( start.y(), end.y(), 1E-8 );
+  bool closed = qgsDoubleNear( start.x(), end.x() ) &&
+                qgsDoubleNear( start.y(), end.y() );
   if ( is3D() && closed )
-    closed &= qgsDoubleNear( start.z(), end.z(), 1E-8 ) || ( std::isnan( start.z() ) && std::isnan( end.z() ) );
+    closed &= qgsDoubleNear( start.z(), end.z() ) || ( std::isnan( start.z() ) && std::isnan( end.z() ) );
   return closed;
 }
 
--- a/src/core/pal/layer.cpp
+++ b/src/core/pal/layer.cpp
@@ -35,6 +35,7 @@
 #include "geomfunction.h"
 #include "util.h"
 #include "qgslabelingengine.h"
+#include "qgslogger.h"
 
 #include <cmath>
 #include <vector>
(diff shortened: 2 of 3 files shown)


m-kuhn:
Good catch!

nyalldawson:
@m-kuhn is that an approval to merge?

m-kuhn (approved):

m-kuhn:
Yes, sir!

nyalldawson:
Thanks!

PR #36887, opened 2020-06-02
Fixes #36114 : NULL value relation sort

### LINKED ISSUE #36114: sorting on a column that has a value relation is inconsistent
I've just noticed that sorting with a value relation as a widget gives bad results as I have NULL values in the middle.. instead of at the beginning or at the end:

image

As you can see from the image above, for the ``strada_id`` field which is coming from a joined layer and the widget is a value relation sorting has NULL values in the middle.
This is very bad as this disrupts the following workflow:
- Order by column... 
- Select some values... based on alphabetical order (if you see NULL.. you stop.. as you would expect thee aren't values after it..)
- Edit the features using field calculator (or some other stuff)
 
**How to Reproduce**

I'll put tomorrow data + project to reproduce as I don't have the time now :(.

If I use the simple text widget instead of the value relation one the sorting is ok.

**Additional context**

nyalldawson:
Fixed version of broken backport https://github.com/qgis/QGIS/pull/36885


nyalldawson:
Moved to https://github.com/qgis/QGIS/pull/36718

PR #36901, opened 2020-06-02
add the possibility to use absolute value for varying line of 1D mesh

### LINKED ISSUE #36858: 1D mesh scaling with negative values
1D mesh renderer that renders width of line based on the value of some dataset does not work with negative values in the dataset.

Use the absolute value of the dataset (by default) or implement checkbox for user to select absolute value (for vector data is already work, but not for scalar data)

vcloarec:
With the new width varying 1D mesh, when scalar value are negative, this could leads to inconsistent displaying depending on user data.
This PR introduce the possibility to use absolute value.

image

fix #36858


PeterPetrik (approved):
This is a small addition to the new feature introduce in 3.14. Without this fix, the scalar data with negative values could be difficult to visualize correctly in many real-world example. Therefore it would be nice to have it in the 3.14 release.

PR #36902, opened 2020-06-02
Reduce siz of png file in ./doc

lhirlimann:
using optipng -o7 reduce the siez of png files in the doc directory.


nyalldawson (approved):

rduivenvoorde:
@lhirlimann in the documentation (and especially the visual changelogs) we also have a lot of images (sometimes really big). Is this an easy way to reduce the size there too? Note that we nowadays also have a lot of animated gifs, probably harder to reduce in size...

lhirlimann:
> @lhirlimann in the documentation (and especially the visual changelogs) we also have a lot of images (sometimes really big). Is this an easy way to reduce the size there too? Note that we nowadays also have a lot of animated gifs, probably harder to reduce in size...

It's about processes. I don't know for .gif yet, I'll finish png and I'll do svgs. Point to the documentation if it's not ./docs in this repo and I'll have a look.

PR #36909, opened 2020-06-02
Fix styles are not copied in offline editing

### LINKED ISSUE #111: Layerstyles are lost in offlineproject
I have a project with multiple layerstyles for one layer. They are used in different visibility presets (used like in demo project ['offline_bees.qgs'](https://github.com/opengisch/QField/blob/master/resources/demo_projects/offline_bees.qgs)).

If creating an offline project only the default style is exported.

Is this related to core offline project plugin or qfieldsync?

Layerstyles in combination with visibility presets are very usefull, specially in qfield with limited possibilities.

m-kuhn:
## Description

When an offline version of a layer is created, the styles are lost. This fixes this issue.

Fixes https://github.com/opengisch/qfieldsync/issues/111

### DIFF
--- a/src/core/qgsofflineediting.cpp
+++ b/src/core/qgsofflineediting.cpp
@@ -40,6 +40,7 @@
 #include "qgsvectorlayer.h"
 #include "qgsproviderregistry.h"
 #include "qgsprovidermetadata.h"
+#include "qgsmaplayerstylemanager.h"
 
 #include <QDir>
 #include <QDomDocument>
@@ -1096,6 +1097,14 @@ void QgsOfflineEditing::updateFidLookup( QgsVectorLayer *remoteLayer, sqlite3 *d
 
 void QgsOfflineEditing::copySymbology( QgsVectorLayer *sourceLayer, QgsVectorLayer *targetLayer )
 {
+  const QMap<QString, QgsMapLayerStyle> styles = sourceLayer->styleManager()->mapLayerStyles();
+  const QStringList styleNames = styles.keys();
+
+  for ( const QString &styleName : styleNames )
+  {
+    targetLayer->styleManager()->addStyle( styleName, styles.value( styleName ) );
+  }
+
   QString error;
   QDomDocument doc;
   QgsReadWriteContext context;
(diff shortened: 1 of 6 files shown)


src/core/qgsofflineediting.cpp
```diff
@@ -1096,6 +1097,14 @@ void QgsOfflineEditing::updateFidLookup( QgsVectorLayer *remoteLayer, sqlite3 *d
 
 void QgsOfflineEditing::copySymbology( QgsVectorLayer *sourceLayer, QgsVectorLayer *targetLayer )
 {
+  const QMap<QString, QgsMapLayerStyle> styles = sourceLayer->styleManager()->mapLayerStyles();
```
nyalldawson, line 1100: Would it be cleaner for all this code to be moved to QgsVectorLayer? Eg if it was renamed to copyStyleFromLayer and accepted the style categories enum flag, I think there's likely existing code in qgsvectorlayer which could be reused for the guts of this function....

m-kuhn: Moved to QgsMapLayerStyleManager.
Adding the style categories was not straightforward, as this class heavily depends on (de-)serialized styles and all the handling of this flag is done on a different level. To do this cleanly it would require a quite profound refactoring of how styles are handled or copying a lot of code from Qgs*Layer to here, both of which are clearly out of scope for this fix.

nyalldawson: What you've done looks good!

I was actually meaning something a bit more generic, and moving copySymbology across to QgsVectorLayer (or QgsMapLayer)... but it's up to you

m-kuhn: I'm not sure I understand where you are going. Is it just moving the method to QgsMapLayer directly? That shouldn't make any difference IMO?
I could also see benefit for something like this in the UI, but this would be new functionality and could also use the current implementation.

nyalldawson: It wasn't about exposing this anywhere in UI -- just about API convenience. If I was looking for an existing API method to copy symbology settings from one layer to another, I definitely wouldn't think to look in the QgsOfflineEditing class :laughing: !

m-kuhn: It's in map layer style manager since your proposal.

PR #36910, opened 2020-06-02
Fix segfault when adding a layer

pblottiere:
Fixes a segmentation fault happening when I try to add a vector layer on Archlinux.

The string returned by `sqlite3_mprintf` has to be freed.

### DIFF
--- a/src/core/qgssqliteutils.cpp
+++ b/src/core/qgssqliteutils.cpp
@@ -126,7 +126,7 @@ QSet<QString> QgsSqliteUtils::uniqueFields( sqlite3 *connection, const QString &
   QSet<QString> uniqueFieldsResults;
   char *zErrMsg = 0;
   std::vector<std::string> rows;
-  QString sql = sqlite3_mprintf( "select sql from sqlite_master where type='table' and name=%q", quotedIdentifier( tableName ).toStdString().c_str() );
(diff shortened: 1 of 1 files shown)


pblottiere:
@rduivenvoorde I think you've the same issue (cf https://lists.osgeo.org/pipermail/qgis-developer/2020-June/061421.html).

rduivenvoorde:
Will build and test it

rduivenvoorde:
Seems to work with me, no more segfault on opening vector data dialog. 
I did have to clean first though.

pblottiere:
> Seems to work with me, no more segfault on opening vector data dialog.

Good. Thanks for testing @rduivenvoorde :+1:

3nids:
@pblottiere as mentionned in the commit, I am getting a crash on mac os.
Reverting this fixes the crash...

pblottiere:
Hi @3nids,

Mmmm... So if we use `sqlite3_mprintf`, we have a crash on Debian (and probably other Debian based dsitrib like Ubuntu) and at least Archlinux. But if we use `QgsSqlite3Mprintf`, we have a crash on mac os.

And `QgsSqlite3Mprintf` is used everywhere in QGIS source code instead of `sqlite3_mprintf`, so I don't think there's an issue with `QgsSqlite3Mprintf` itself...

A stupid idea: can you try to use a fresh `QString` for the second call to `QgsSqlite3Mprintf` instead of reusing `sql`?

PR #36911, opened 2020-06-02
never hide mesh static dataset widget

vcloarec:
The new temporal framework led to the concept of static dataset for mesh layer, that is the datasets that are displayed if the temporal navigation is off or if the mesh layer is set to "Static Dataset".

Before, the possibility to set the map canvas with temporal navigation off, the choice of the dataset (with comboboxes) was possible only if the checkbox "Static dataset" was checked in the mesh layer properties widget.
Furthermore, the default QGIS map canvas is set with temporal navigation off. So after creating a new document and loading a mesh layer, it is the static dataset that is displayed even if the mesh layer is not forced to be static.
So, now, hiding the static dataset comboboxes has fewer sens.
With this PR those comboboxes are not hidden anymore.

image

### DIFF
--- a/src/app/mesh/qgsmeshlayerproperties.h
+++ b/src/app/mesh/qgsmeshlayerproperties.h
@@ -74,7 +74,6 @@ class APP_EXPORT QgsMeshLayerProperties : public QgsOptionsDialogBase, private U
 
     void onTimeReferenceChange();
 
-    void onStaticDatasetCheckBoxChanged();
   private:
     //! Pointer to the mesh styling widget
     QgsRendererMeshPropertiesWidget *mRendererMeshPropertiesWidget = nullptr;
--- a/src/app/mesh/qgsmeshlayerproperties.h
+++ b/src/app/mesh/qgsmeshlayerproperties.h
@@ -74,6 +74,7 @@ class APP_EXPORT QgsMeshLayerProperties : public QgsOptionsDialogBase, private U
 
     void onTimeReferenceChange();
 
+    void onStaticDatasetCheckBoxChanged();
   private:
     //! Pointer to the mesh styling widget
     QgsRendererMeshPropertiesWidget *mRendererMeshPropertiesWidget = nullptr;
@@ -88,6 +89,8 @@ class APP_EXPORT QgsMeshLayerProperties : public QgsOptionsDialogBase, private U
      * Previous layer style. Used to reset style to previous state if new style
      * was loaded but dialog is canceled */
     QgsMapLayerStyle mOldStyle;
+
+    bool mIsMapSettingsTemporal = false;
 
     friend class TestQgsMeshLayerPropertiesDialog;
 
--- a/src/app/mesh/qgsmeshstaticdatasetwidget.cpp
+++ b/src/app/mesh/qgsmeshstaticdatasetwidget.cpp
@@ -37,25 +37,10 @@ void QgsMeshStaticDatasetWidget::syncToLayer()
   if ( !mLayer || !mLayer->dataProvider() )
     return;
 
-  mScalarDatasetGroup = mLayer->rendererSettings().activeScalarDatasetGroup();
-  mVectorDatasetGroup = mLayer->rendererSettings().activeVectorDatasetGroup();
   mDatasetScalarModel->setMeshLayer( mLayer );
-  mDatasetScalarModel->setDatasetGroup( mScalarDatasetGroup );
   mDatasetVectorModel->setMeshLayer( mLayer );
-  mDatasetVectorModel->setDatasetGroup( mVectorDatasetGroup );
-
-  QgsMeshDatasetIndex scalarStaticIndex = mLayer->staticScalarDatasetIndex();
-  QgsMeshDatasetIndex vectorStaticIndex = mLayer->staticScalarDatasetIndex();
-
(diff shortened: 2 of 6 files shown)


PeterPetrik:
Before (on master)
1. if the temporal navigation is **disabled**
   a) if mesh layer has force static dataset off, it shows data for timestep 0
   b) if mesh layer has force static dataset on, the extra combobox with available timesteps is shown and the selected time is shown
2. if the temporal navigation is **enabled**
   a) if mesh layer has force static dataset off, it shows data for timestep from temporal navigation
   b) if mesh layer has force static dataset on, the extra combobox with available timesteps is shown and the selected time is shown

This PR changes it that the extra combobox with available timesteps is shown and the selected time is shown, regardless of whether temporal navigation is enabled or disabled
- I am not sure how the information is used when temporal navigation is enabled? (which I believe is what users will do most of the time?)
- If the problem is that by default the temporal controller is disabled, wouldn't it be the better alternative to just enable the temporal controller anytime the mesh layer is added to the canvas?

PeterPetrik:
Another question, if https://github.com/qgis/QGIS/pull/36911 is merged, is the static dataset group  selectors even necessary at all? (since you need to choose only the time?)

PeterPetrik:
ah ok,.. I do not have strong opinion for either as it is now on master of in this PR..

wonder-sk:
One suggestion - how about making it a collapsible checkable group box (see QgsCollapsibleGroupBox)

It could be unchecked and collapsed by default
image

When checked, it auto-expands:
image

And it is also possible to expand it from collapsed state when unchecked.

This kind of group box is used in QGIS in various places so users are used to it, and I think it helps to keep the GUI clean by hiding extra options users don't need if not using some more advanced workflow...

vcloarec:
Static dataset combo boxes are needed if temporal navigation is off OR if the layer is forced to be static.
As temporal navigation can be off and layer not forced to be static, I don't think we can make the group box checkable and collapse it if layer is not forced to be static. 
But we can make it collapsable, and collapse it when temporal navigation is not off (if the map settings, as temporal range object, is not set as temporal) AND the layer not forced static when opening widget properties.
What about that?

vcloarec:
So the new version :
static

src/app/mesh/qgsmeshstaticdatasetwidget.cpp
```diff
@@ -37,25 +37,10 @@ void QgsMeshStaticDatasetWidget::syncToLayer()
   if ( !mLayer || !mLayer->dataProvider() )
     return;
 
-  mScalarDatasetGroup = mLayer->rendererSettings().activeScalarDatasetGroup();
-  mVectorDatasetGroup = mLayer->rendererSettings().activeVectorDatasetGroup();
   mDatasetScalarModel->setMeshLayer( mLayer );
-  mDatasetScalarModel->setDatasetGroup( mScalarDatasetGroup );
```
PeterPetrik, line 43: this is not required anymore?

src/app/mesh/qgsmeshstaticdatasetwidget.cpp
```diff
@@ -37,25 +37,10 @@ void QgsMeshStaticDatasetWidget::syncToLayer()
...
   mDatasetVectorModel->setMeshLayer( mLayer );
-  mDatasetVectorModel->setDatasetGroup( mVectorDatasetGroup );
```
PeterPetrik, line 45: this is not required anymore?

src/app/mesh/qgsmeshlayerproperties.h
```diff
@@ -90,6 +90,8 @@ class APP_EXPORT QgsMeshLayerProperties : public QgsOptionsDialogBase, private U
      * was loaded but dialog is canceled */
     QgsMapLayerStyle mOldStyle;
 
+    bool mIsMapSettingsTemporal = false;
```
PeterPetrik, line 93: this is assigned in constructor of QgsMeshLayerProperties but I do not see where else it is used or changed afterwards? maybe we do not need the member variable at all ?

PeterPetrik (commented):
I like the new collapsible group

src/app/mesh/qgsmeshstaticdatasetwidget.cpp
vcloarec: refactored in setScalarDatasetGroup()

src/app/mesh/qgsmeshstaticdatasetwidget.cpp
vcloarec: refactored in setVectorDatasetGroup()

src/app/mesh/qgsmeshlayerproperties.h
vcloarec: here : https://github.com/qgis/QGIS/blob/3fc1cb126d2767021a47309718ab91180e437464/src/app/mesh/qgsmeshlayerproperties.cpp#L223

and here : 

https://github.com/qgis/QGIS/blob/3fc1cb126d2767021a47309718ab91180e437464/src/app/mesh/qgsmeshlayerproperties.cpp#L474

PeterPetrik: this is where is it used, but where it is changed except contructor?

vcloarec: nowhere. I don't see the problem...

PeterPetrik (approved):

PR #36915, opened 2020-06-03
Python console improvements

nirvn:
## Description

How many times per week are you finding yourself swearing at the console after a failed copy to or pasting from clipboard after selecting all text via the ctrl+a shortcut? Swear no more.

This PR fixes _a bunch of_ issues with the console input area by moving the >>> prompt characters _into the margin_ instead of appending it into the text input area. This simplifies the code a lot and makes a bunch of keyboard shortcuts behave as you'd expect (hitting the home key, copy/pasting, etc.)

I've also taken the time to transform the help button into a help menu that provides shortcuts to the python console documentation, the pyqgis API, and the pyqgis cookbook.

PR selfie:
image

### DIFF
--- a/python/console/console.py
+++ b/python/console/console.py
@@ -29,11 +29,12 @@ from .console_sci import ShellScintilla
 from .console_output import ShellOutputScintilla
 from .console_editor import EditorTabWidget
 from .console_settings import optionsDialog
-from qgis.core import QgsApplication, QgsSettings
+from qgis.core import Qgis, QgsApplication, QgsSettings
 from qgis.gui import QgsFilterLineEdit, QgsHelp, QgsDockWidget
 from functools import partial
 
 import sys
+import re
 
 _console = None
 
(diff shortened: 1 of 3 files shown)


nyalldawson (approved):
Woohoo! I've always hated this too..!

rduivenvoorde:
@nirvn fyi on gnome with me it looks like this:

image

Would it not be nicer to give the >>> the same background color as the margin? 
Or is there no way to know what color that is?

Mmm, that is funny, in 3.12 the background of the margin WAS white?

Screenshot-20200603083855-208x353

Ok, so then probably the margin has now a wrong color instead of the cursor?

Another thing I was wondering: if you copy code (including the cursor) from the console itself (so not the input) the >>> is removed? That is another code-path? So not usable?

nirvn:
@rduivenvoorde , fix inbound.

PR #36916, opened 2020-06-03
Add processing algorithms to export a print layout as PDF/image

nyalldawson:
This allows for models which export print layouts from the current project. One
use case for this is allowing users to create an in-project model which exports
a particular set of layouts from the project to certain folders, so that they
can easily re-export the current project in a single operation instead of
having to manually open multiple layouts and export one-by-one.

Additionally, with the new capabilities to have expression based output files
inside models, you can automatically export the layouts to a folder with
the current date tag and include this in the exported file names!

### DIFF
--- a/src/analysis/CMakeLists.txt
+++ b/src/analysis/CMakeLists.txt
@@ -88,6 +88,8 @@ SET(QGIS_ANALYSIS_SRCS
   processing/qgsalgorithmjoinbynearest.cpp
   processing/qgsalgorithmjoinwithlines.cpp
   processing/qgsalgorithmkmeansclustering.cpp
+  processing/qgsalgorithmlayouttoimage.cpp
+  processing/qgsalgorithmlayouttopdf.cpp
   processing/qgsalgorithmlinedensity.cpp
   processing/qgsalgorithmlineintersection.cpp
   processing/qgsalgorithmlinesubstring.cpp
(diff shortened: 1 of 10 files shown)


nyalldawson:
(Lots more we could potentially do here: export as GeoPDF, export as SVG, export atlas/report as PDF/image/svg, set print layout map extent, set print layout map scale, ...! )

alexbruy (approved):

roya0045:
> (Lots more we could potentially do here: export as GeoPDF, export as SVG, export atlas/report as PDF/image/svg, set print layout map extent, set print layout map scale, ...! )

I would assume that data defined variables would still be valid for such exports. 

Would it be simpler in the future to have two export functions: export atlas layout and export layout? I would imagine that setting the output in the function would be simpler than having specific dedicated functions.

nyalldawson:
>I would assume that data defined variables would still be valid for such exports.

Correct

> Would it be simpler in the future to have two export functions: export atlas layout and export layout? I would imagine that setting the output in the function would be simpler than having specific dedicated functions.

Not really -- the advanced options are very different depending on the output format (svg/pdf/image), so we'd end up needing to expose a ton of options, of which only some apply depending on the format selected. The end result is a confusing set of parameters and poor UX (the grass algorithms suffer from this -- by necessity they expose a ton of different options, some of which are muttually exclusive or interdependant, and making the wrong selection of options results in errors or no output. To avoid this, the native QGIS algorithms have all been designed around the principal of keeping them as "atomic" as possible, and focused on doing one single task)

roya0045:
@nyalldawson @DelazJ The algs are in the cartography group, but are exclusive to the modeler. Would it make more sense to have them in the modeler group, or have them be available for qgis in general?

DelazJ:
@roya0045 I think you're being misled by the pr message. I never use these algs but I'm pretty sure (no computer around to check) that they can be used directly from the toolbox. Not exclusive to models. So "cartography" section it's.
Also modeler only tools are available only in modeler context.

roya0045:
@DelazJ Yeah you're right, they are also available in standalone. Doc was fixed.

PR #36919, opened 2020-06-03
[python console] Improve key up/down press for multiline command string

nirvn:
## Description

Another console behavior that drives most crazy: pressing up or down on a long, multiline string overwrites that string with the previous/next entry in the commands history.

The PR fixes this broken UX by using the following logic:

- if you are pressing key up, it'll *only* jump to previous command if your cursor's location is at the very beginning of the line, otherwise it'll allow navigating through the string (which eventually gets you to the beginning of the line, upon which you'd then go to previous command)

- if you are pressing key down, same improved behavior applies, except it's when the cursor is at the very end of the string.


nyalldawson (approved):
nice!

PR #36921, opened 2020-06-03
Limit scope of Ctrl+C shortcut 

### LINKED ISSUE #27035: Ctrl+C doesn't copy, despite dialog prompt
Author Name: **Loren Amelang** (@LorenAmelang)
Original Redmine Issue: [19206](https://issues.qgis.org/issues/19206)
Affected QGIS version: 3.0.3
Redmine category:gui

---
As the Subject says, In almost every place I've tried it, Ctrl+C is ignored but using the dialog Copy command works. The screenshot shows the Python Console, but it is definitely not confined to Python. Ctrl+C doesn't work in Log Messages, but at least that dialog doesn't show the Ctrl+C hotkey. 

I'd much rather the hotkey worked everywhere, but if there is some reason why it can't, is there some way to warn us we're in one of those places before we assume it did and paste some previous buffer we don't want? Maybe a failed Ctrl+C causes an error sound? Or a successful one causes a confirmation sound? 

Right now the only practical strategy is to assume the hotkey won't work and use the dialog Copy click. 

---

- [Ctrl+C doesn_t Copy.jpg](https://issues.qgis.org/attachments/download/12846/Ctrl+C%20doesn_t%20Copy.jpg) (Loren Amelang)

### LINKED ISSUE #31392: Command-C in About dialog copies canvas geometries instead of dialog content
**Describe the bug**

Command-C to copy About dialog content copies canvas geometries instead. Command-A selects the geometries instead of the About content. Only right-click->Select right-click->Copy work

**How to Reproduce**

1. Open About dialog and click in it to focus
2. Command-A to select (nothing highlights)
3. Command-C to copy
4. Paste somewhere and records from the active layer in the canvas are pasted 

**QGIS and OS versions**

QGIS version | 3.4.11-Madeira | QGIS code revision | 9a8a6d4687
-- | -- | -- | --
Compiled against Qt | 5.12.3 | Running against Qt | 5.12.3
Compiled against GEOS | 3.7.2-CAPI-1.11.2 | Running against GEOS | 3.7.2-CAPI-1.11.2 b55d2125
PostgreSQL Client Version | 11.3 | SpatiaLite Version | 4.3.0a
QWT Version | 6.1.4 | QScintilla2 Version | 2.11.1
Compiled against PROJ | 520 | Running against PROJ | 5.2.0

### LINKED ISSUE #31914: Coying messages in logs throws python error if python console is open
3.4.12
While reporting https://github.com/qgis/QGIS/issues/31913 - I selected the text of the error from the top error bar. This worked, But when I pressed `Ctrl-C` (copy), this raised a different error (and the content was *not* added to the copy buffer!).

`Python error: An error has occured while executing Python code. See log`.
It appears that QGIS is considering my `Ctrl-C` to try and copy the error as a `KeyboardInterrupt` for a Python script.

This makes it impossible to copy bugs/exceptions for issue reports.

**Edit**: This may have been because the Python Console was open but it doesn't have focus - the focus is on the selected error text.

### LINKED ISSUE #31918: Unable to copy message - copies attribute table instead
3.4.12 - windows
(Similar to but different from https://github.com/qgis/QGIS/issues/31914)

I'm trying to copy an error from the top error bar to paste it here for a ticket. I have selected the text and ctrl-c (copy).

Rather than the error text, QGIS is adding 345 lines of stuff from the attribute take of the WFS layer I have added (but not enabled) in the view (345 = all the features).

I've cleared by copy buffer and done it again and I get the same thing.

I've never even opened the attribute table of the WFS except after this to confirm the data was coming from it. Nor have I selected any of the WFS features. In fact the WFS didn't even connect properly (due to a bug in the WFS) and no WFS features have been rendered.

This is the WFS I added first:
http://si.icnf.pt:80/geoserver/BDG/atlas_bivalves_2007/wfs - and this was the error it came up with: https://github.com/qgis/QGIS/issues/31917

nyalldawson:
for copy features to canvas 
and for python console to console widget

Avoids ambiguous shortcut warnings/unexpected behavior when pressing Ctrl+C

Fixes #27035
Fixes #31918
Fixes #31914
Fixes #31392

### DIFF
--- a/python/console/console_output.py
+++ b/python/console/console_output.py
@@ -171,6 +171,7 @@ class ShellOutputScintilla(QsciScintilla):
         self.copyShortcut.setContext(Qt.WidgetWithChildrenShortcut)
         self.copyShortcut.activated.connect(self.copy)
         self.selectAllShortcut = QShortcut(QKeySequence.SelectAll, self)
+        self.selectAllShortcut.setContext(Qt.WidgetWithChildrenShortcut)
         self.selectAllShortcut.activated.connect(self.selectAll)
 
     def insertInitText(self):
--- a/python/console/console_output.py
+++ b/python/console/console_output.py
@@ -168,6 +168,7 @@ class ShellOutputScintilla(QsciScintilla):
         self.runScut.activated.connect(self.enteredSelected)
         # Reimplemented copy action to prevent paste prompt (>>>,...) in command view
         self.copyShortcut = QShortcut(QKeySequence.Copy, self)
+        self.copyShortcut.setContext(Qt.WidgetWithChildrenShortcut)
         self.copyShortcut.activated.connect(self.copy)
         self.selectAllShortcut = QShortcut(QKeySequence.SelectAll, self)
         self.selectAllShortcut.activated.connect(self.selectAll)
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -2816,10 +2816,14 @@ void QgisApp::createActions()
 
   connect( mActionDiagramProperties, &QAction::triggered, this, &QgisApp::diagramProperties );
 
-  // we can't set the shortcut on the copy action, because we need to restrict it's context to the canvas and it's children..
+  // we can't set the shortcut these actions, because we need to restrict their context to the canvas and it's children..
   QShortcut *copyShortcut = new QShortcut( QKeySequence::Copy, mMapCanvas );
   copyShortcut->setContext( Qt::WidgetWithChildrenShortcut );
   connect( copyShortcut, &QShortcut::activated, this, [ = ] { copySelectionToClipboard(); } );
+
+  QShortcut *selectAllShortcut = new QShortcut( QKeySequence::SelectAll, mMapCanvas );
+  selectAllShortcut->setContext( Qt::WidgetWithChildrenShortcut );
+  connect( selectAllShortcut, &QShortcut::activated, this, &QgisApp::selectAll );
 
 #ifndef HAVE_POSTGRESQL
   delete mActionAddPgLayer;
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -2816,6 +2816,11 @@ void QgisApp::createActions()
 
   connect( mActionDiagramProperties, &QAction::triggered, this, &QgisApp::diagramProperties );
 
+  // we can't set the shortcut on the copy action, because we need to restrict it's context to the canvas and it's children..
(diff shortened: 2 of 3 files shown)


nirvn:
@nyalldawson , shouldn't this be done for other shortcuts like ctrl+a for e.g.?

nirvn:
Just tested, can't ctrl + a the map canvas when the python console is open, and vice versa.

nyalldawson:
Ok - I guess there's a few more shortcuts we'll need to catch then!

One potential issue here is that by removing the shortcut from the action, we lose the display of the shortcut in the main menu. I'm going to see if there's anyway around that... I think we might need to reparent the offending action to the canvas to get around this.

nirvn:
Re ctrl+a, I've been hit by this nightmare scenario a few times:
- a save file dialog opens, I hit ctrl+a to select the filename there to overwrite quickly;
- instead of being quick, the canvas goes ahead and tries to select all points from a large remotely stored layer, freezing my app for ~5-10sec.

nyalldawson:
@nirvn Select All fixed now too (one nice side effect of this is that Ctrl+A works elsewhere too -- e.g. if the layer tree is focused then it'll select all layers!)

So the only downside is that we still lose the display of the shortcut key in the menu, but I can't find anyway to avoid that.

nirvn (approved):
Great!

uclaros:
> @nirvn Select All fixed now too (one nice side effect of this is that Ctrl+A works elsewhere too -- e.g. if the layer tree is focused then it'll select all layers!)

@nyalldawson Do you have a use case for this behavior? Because it has broken the workflow: click on layer, ctrl-a to select all features ...

nirvn:
@uclaros , here's a use case: click on a layer, ctrl a to select all layers ;)

Joking aside, it was impossible to select all items from any widget connected to the mainwindow. Eg, selecting all items in the categorized renderer list via ctrl a, etc.

nyalldawson:
We could possibly set the ctrl-A shortcut to apply within the layer tree too. It shouldn't be made global again, but we could manually add it to widgets which make sense...

nirvn:
@nyalldawson , agreed -- if we want ctrl-A to select all features when not focused on the canvas, it should be re-added selectively for narrow scenarios.

nyalldawson:
@nirvn so I'm +1 to making select all/unselect all on the layer tree apply to canvas -- are you ok with this?

nirvn:
Ya

uclaros:
I think ctrl-[xcv] on the layer tree panel should also be forwarded to the canvas, see  #38849

PR #36922, opened 2020-06-03
QGS vs QGZ is not about supporting auxiliary data

DelazJ:
but whether these data are embedded in the file or not. No?

### DIFF
--- a/src/ui/qgsoptionsbase.ui
+++ b/src/ui/qgsoptionsbase.ui
@@ -1050,7 +1050,7 @@
                     <item row="0" column="1">
                      <widget class="QRadioButton" name="mFileFormatQgzButton">
                       <property name="text">
-                       <string>QGZ Archive file format, supports auxiliary data</string>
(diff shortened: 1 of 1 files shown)


src/ui/qgsoptionsbase.ui
```diff
@@ -1060,7 +1060,10 @@
                     <item row="1" column="1">
                      <widget class="QRadioButton" name="mFileFormatQgsButton">
                       <property name="text">
-                       <string>QGS Project saved in a clear text, does not support auxiliary data</string>
+                       <string>QGS Project saved in a clear text, does not embed auxiliary data</string>
+                      </property>
+                      <property name="toolTip">
+                       <string>The auxiliary data will be kept in a separate file</string>
```
nyalldawson, line 1066:
```suggestion
                       <string>The auxiliary data will be kept in a separate .qgd data file which must be distributed along with the .qgs project file.</string>
```

nyalldawson:
Good catch!

nyalldawson (approved):

PR #36926, opened 2020-06-03
QgsOgcUtils GML: handle srsName on exterior rings

### LINKED ISSUE #36398: QGIS Server: WFS Request does not use SrsName on the geometry

elpaso:
Fixes #36398

### DIFF
--- a/src/core/qgsogcutils.cpp
+++ b/src/core/qgsogcutils.cpp
@@ -24,6 +24,7 @@
 #include "qgsrectangle.h"
 #include "qgsvectorlayer.h"
 #include "qgsexpressioncontextutils.h"
+#include "qgsproject.h"
 
 #include <QColor>
 #include <QStringList>
@@ -266,6 +267,14 @@ QgsGeometry QgsOgcUtils::geometryFromGMLPolygon( const QDomElement &geometryElem
   //read all the coordinates (as QgsPoint) into memory. Each linear ring has an entry in the vector
   QgsMultiPolylineXY ringCoordinates;
 
+  // Geometry SRS name
+  QgsCoordinateReferenceSystem geomSrs;
+
+  if ( geometryElement.hasAttribute( QStringLiteral( "srsName" ) ) )
+  {
+    geomSrs.createFromString( geometryElement.attribute( QStringLiteral( "srsName" ) ) );
+  }
+
   //read coordinates for outer boundary
   QgsPolylineXY exteriorPointList;
   QDomNodeList outerBoundaryList = geometryElement.elementsByTagNameNS( GML_NAMESPACE, QStringLiteral( "outerBoundaryIs" ) );
@@ -312,10 +321,29 @@ QgsGeometry QgsOgcUtils::geometryFromGMLPolygon( const QDomElement &geometryElem
     {
       return QgsGeometry();
     }
-    if ( readGMLPositions( exteriorPointList, posElement ) != 0 )
+    if ( readGMLPositions( exteriorPointList, posElement ) != false )
     {
       return QgsGeometry();
     }
+
+    if ( geomSrs.isValid() )
+    {
+      QgsCoordinateReferenceSystem exteriorSrs;
+      const QDomElement ringElement {  exteriorList.at( 0 ).firstChild().toElement() };
+      if ( ringElement.hasAttribute( QStringLiteral( "srsName" ) ) )
+      {
+        exteriorSrs.createFromString( ringElement.attribute( QStringLiteral( "srsName" ) ) );
+      }
+      if ( exteriorSrs != geomSrs )
+      {
+        const QgsCoordinateTransform transformer { exteriorSrs, geomSrs, QgsProject::instance()->transformContext() };
+        for ( auto it = exteriorPointList.begin(); it != exteriorPointList.end(); ++it )
+        {
+          *it = transformer.transform( *it );
+        }
+      }
+    }
+
     ringCoordinates.push_back( exteriorPointList );
 
     //read coordinates for inner boundary
--- a/tests/src/core/testqgsogcutils.cpp
+++ b/tests/src/core/testqgsogcutils.cpp
@@ -98,6 +98,20 @@ void TestQgsOgcUtils::testGeometryFromGML()
   geomBox = QgsOgcUtils::geometryFromGML( QStringLiteral( "<gml:Envelope srsName=\"foo\"><gml:lowerCorner>135.2239 34.4879</gml:lowerCorner><gml:upperCorner>135.8578 34.8471</gml:upperCorner></gml:Envelope>" ) );
   QVERIFY( !geomBox.isNull() );
   QVERIFY( geomBox.wkbType() == QgsWkbTypes::Polygon );
+
+  // Test polygon with different srsName on exterior ring
+  geom = QgsOgcUtils::geometryFromGML( QStringLiteral( R"raw(
+    <Polygon srsName="EPSG:4326">
+      <exterior>
+        <LinearRing srsName="EPSG:3857">
+          <posList srsDimension="2">890555.93 5465442.18 1001875.42 5465442.18 1001875.42 5621521.49 890555.93 5621521.49 890555.93 5465442.18</posList>
+        </LinearRing>
+      </exterior>
+    </Polygon>
+  )raw" ) );
+  QVERIFY( !geom.isNull() );
+  QVERIFY( geom.wkbType() == QgsWkbTypes::Polygon );
+  QGSCOMPARENEARRECTANGLE( geom.boundingBox(), QgsRectangle( 8, 44, 9, 45 ), 0.1 );
 }
 
 static bool compareElements( QDomElement &element1, QDomElement &element2 )
--- a/tests/src/python/test_qgsserver_wfs.py
+++ b/tests/src/python/test_qgsserver_wfs.py
@@ -217,8 +217,9 @@ class TestQgsServerWFS(QgsServerTestBase):
             body
         )
 
-    @unittest.expectedFailure("Issue #36398: QGIS Server: WFS Request does not use SrsName on the geometry")
     def test_getfeature_post(self):
+        """Issue #36398: QGIS Server: WFS Request does not use SrsName on the geometry"""
+
         tests = []
         expectNotEqualTests = []
 
@@ -328,9 +329,9 @@ class TestQgsServerWFS(QgsServerTestBase):
             <LinearRing srsName="EPSG:3857">
               <posList srsDimension="2">
                 890555.93 5465442.18
-                1001875.42, 5465442.18
-                1001875.42, 5621521.49
-                890555.93, 5621521.49
+                1001875.42 5465442.18
+                1001875.42 5621521.49
+                890555.93 5621521.49
                 890555.93 5465442.18
               </posList>
             </LinearRing>
--- a/tests/src/python/test_qgsserver_wfs.py
+++ b/tests/src/python/test_qgsserver_wfs.py
@@ -344,6 +344,33 @@ class TestQgsServerWFS(QgsServerTestBase):
 """
         tests.append(('within3857FilterTemplate_post', within3857FilterTemplate.format("")))
 
+        # Requesting features in 3857 and the filter is in 4326
+        within4326FilterTemplate3857 = """<?xml version="1.0" encoding="UTF-8"?>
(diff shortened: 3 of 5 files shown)


src/core/qgsogcutils.cpp
```diff
@@ -312,10 +321,29 @@ QgsGeometry QgsOgcUtils::geometryFromGMLPolygon( const QDomElement &geometryElem
     {
       return QgsGeometry();
     }
-    if ( readGMLPositions( exteriorPointList, posElement ) != 0 )
+    if ( readGMLPositions( exteriorPointList, posElement ) != false )
```
nyalldawson, line 324:
```suggestion
    if ( readGMLPositions( exteriorPointList, posElement ) )
```

src/core/qgsogcutils.cpp
```diff
@@ -312,10 +321,29 @@ QgsGeometry QgsOgcUtils::geometryFromGMLPolygon( const QDomElement &geometryElem
...
+    {
+      QgsCoordinateReferenceSystem exteriorSrs;
+      const QDomElement ringElement {  exteriorList.at( 0 ).firstChild().toElement() };
+      if ( ringElement.hasAttribute( QStringLiteral( "srsName" ) ) )
+      {
+        exteriorSrs.createFromString( ringElement.attribute( QStringLiteral( "srsName" ) ) );
+      }
+      if ( exteriorSrs != geomSrs )
+      {
+        const QgsCoordinateTransform transformer { exteriorSrs, geomSrs, QgsProject::instance()->transformContext() };
```
nyalldawson, line 339: Using QgsProject::instance() is definitely not server friendly here. You'll need to instead add a transform context argument to the method and use that instead (or better, add a "GML Conversion Context" struct/class containing the transform context member)

tests/src/core/testqgsogcutils.cpp
```diff
@@ -98,6 +98,20 @@ void TestQgsOgcUtils::testGeometryFromGML()
   geomBox = QgsOgcUtils::geometryFromGML( QStringLiteral( "<gml:Envelope srsName=\"foo\"><gml:lowerCorner>135.2239 34.4879</gml:lowerCorner><gml:upperCorner>135.8578 34.8471</gml:upperCorner></gml:Envelope>" ) );
   QVERIFY( !geomBox.isNull() );
   QVERIFY( geomBox.wkbType() == QgsWkbTypes::Polygon );
+
+  // Test polygon with different srsName on exterior ring
+  geom = QgsOgcUtils::geometryFromGML( QStringLiteral( R"raw(
+    <Polygon srsName="EPSG:4326">
+      <exterior>
+        <LinearRing srsName="EPSG:3857">
```
rouault, line 106: OMG ! Is it really something in active use ???? I've never seen that before. Looking quickly at the bug report of https://github.com/qgis/QGIS/issues/36398, it doesn't seem to be one of the situations of the GML snippets (but I didn't read more). If use of such crazy constructs in the wild isn't confirmed, I strongly question the need for QGIS to try to be smart and reproject. 
FYI OGR GML parsing won't do that

src/core/qgsogcutils.cpp
elpaso: I asked myself the same question: why the original author used ` != 0` ?
I think that he wanted to capture reader's attention to the counter-intuitive fact that `readGMLPositions` result `true` is the error condition and `false` means that all was ok.

I chose to leave that (and just use `false` instead of `0` for readability), I think that the compiler would be smart enough to generate the same code if omitting the comparison.

src/core/qgsogcutils.cpp
elpaso: yeah, I know. I'll do if we decide to go further with this PR (see comments below)

tests/src/core/testqgsogcutils.cpp
elpaso: @rouault the test failing test cases originally provided by @ismailsunni dis not clearly bring the issue into the light but this patch makes all the test cases provided by him to pass, I just added another test in the CPP side for clarity, but the same scenario is indirectly tested in the wfs python server tests originally provided in the bug report followup.

I myself have expressed doubts about real life use case, I couldn't find an example in the wild of such a wierd use case but I also could not find a clear indication that is would be an invalid GML. In any event while I was running the failing test cases I saw the solution and I wrote it to check if that was in fact the solution of the issue, this PR was the logical consequence of that debug session.

Additionally, if we decided to go on with the implementation, we should probably support interior rings too (I just applied the transformation to exterior).

So, I'm asking you (@rouault, @nyalldawson and expecially @ismailsunni who've found the test case presumably in the wild) what should we do here? Finish the implementation and support this wierd GMLs or close with a won't fix?

rouault: @cportele Sorry to bother you, but quick GML question. Is a construct like
``` 
<Polygon srsName="EPSG:4326">
      <exterior>
        <LinearRing srsName="EPSG:3857"> 
```
supposed to be valid ?   I read §9.10 "Spatial reference system used in a feature or feature collection" of GML 3..2.2 spec which mentions "Inheritance of the coordinate reference system continues to any depth of nesting, but if overruled by a local srsName declaration, then the new coordinate reference system is inherited by all its children in turn", but I'm not clear if the overruling mentionnes applies inside a geometry, or that a complex feature with several geometries might rather have different SRS for each of them.

cportele: @rouault - Technically, GML does not prohibit this (based on ISO 19107). There are some use cases that could exploit such a capability (although not what we see in the example, I don't understand that use case; more something like having a multi point geometry with points in different UTM zones). 

In practice, I would expect that most software will require that only a single CRS is used per feature geometry and reject everything else. The GML Simple Features profile makes this explicit, i.e. the example above would not conform to the simple feature profile.

elpaso: @cportele, thank you for the clarification, I guess we are heading towards a won't fix.

@ismailsunni can you tell us more about you use and test cases?

ismailsunni: Hi @elpaso @rouault 
Sorry, it was my mistake. The bug originally has the same srs (e.g. 3857) for all the attributes in the request. I was trying to make a simpler unit test to show the problem, but it seems ended up with something very strange and very uncommon in real life.

So, the bug originally comes from this kind of request (see that all srsName is 3857):
```xml
<?xml version="1.0" encoding="UTF-8"?>
<wfs:GetFeature service="WFS" version="1.0.0" {} xmlns:wfs="http://www.opengis.net/wfs" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/wfs http://schemas.opengis.net/wfs/1.1.0/wfs.xsd">
  <wfs:Query typeName="testlayer" srsName="EPSG:3857" xmlns:feature="http://www.qgis.org/gml">
    <ogc:Filter xmlns:ogc="http://www.opengis.net/ogc">
      <Within>
        <PropertyName>geometry</PropertyName>
        <Polygon xmlns="http://www.opengis.net/gml" srsName="EPSG:3857">
          <exterior>
            <LinearRing srsName="EPSG:3857">
              <posList srsDimension="2">
                8 44
                9 44
                9 45
                8 45
                8 44
              </posList>
            </LinearRing>
          </exterior>
        </Polygon>
      </Within>
    </ogc:Filter>
  </wfs:Query>
</wfs:GetFeature>
```

I just want to show that 4326 is always used for the filtering.

Let me know if you need more input or I can update the unit test to have a better representation  of the original bug (i.e. only needs to change the srsName in the Polygon)

elpaso: @ismailsunni this PR fixes all your test cases and an additional one that I've added (https://github.com/qgis/QGIS/pull/36926/files#diff-61a1cac6f0799a1d603fac5b80d0f5ddR347) but I'm non convinced we should support the use case and I find the tests you have provided redundant and confusing (and incorrect GML in some cases such as https://github.com/qgis/QGIS/pull/36713/files#diff-61a1cac6f0799a1d603fac5b80d0f5ddR331) 

So this is what we should do: 

1. revert your tests PR https://github.com/qgis/QGIS/pull/36713
2. **you** create another failing test with a single test case that has a correct input and a wrong output: the snippet you pasted above appears to be a **wrong** input unless `8 44` are really meant to be  meters in 3857 and the test dataset has features in that BBOX (which is **not**). Before doing that, please check if the unique tests that I have added (https://github.com/qgis/QGIS/pull/36926/files#diff-61a1cac6f0799a1d603fac5b80d0f5ddR347) represents your use case, if it does we can just use that one
3. trash this PR
4. verify if we do have a real issue
5. fix the issue if we have one

elpaso:
Waiting for a new failing test case from @ismailsunni , closing this PR as won't fix for the multiple srsName GML issue.

PR #36928, opened 2020-06-03
Clarify location of Qt for Mac OSX build

citizenfish:
Clarify location of Qt for Mac OSX build

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -53,18 +53,18 @@ Last Change : 2020-05-08
   1. Introduction
   ===============
 
-This document is the original installation guide of the described software 
-QGIS. The software and hardware descriptions named in this 
-document are in most cases registered trademarks and are therefore subject 
-to the legal requirements. QGIS is subject to the GNU General Public 
+This document is the original installation guide of the described software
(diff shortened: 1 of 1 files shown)


INSTALL
```diff
@@ -74,15 +74,15 @@ http://htmlpreview.github.io/?https://raw.github.com/qgis/QGIS/master/doc/INSTAL
...
-place for describing build procedures. Please do not remove this notice. 
+place for describing build procedures. Please do not remove this notice.
 
-/!\ Note to document writers: This document is generated from 
-doc/INSTALL.t2t - if you need to edit this document, be sure to edit that 
-file rather than the generated INSTALL document found in the root of the 
+/!\ Note to document writers: This document is generated from
+doc/INSTALL.t2t - if you need to edit this document, be sure to edit that
+file rather than the generated INSTALL document found in the root of the
 source directory.
```
DelazJ, line 86: @citizenfish As mentioned here, this file is a generated one so you probably want to edit the doc/osx.t2t file

citizenfish: My bad! I'll get that done

PR #36929, opened 2020-06-03
Clarify location of Qt install for OSX build

citizenfish:
Clarify location of Qt install for OSX build

### DIFF
--- a/doc/osx.t2t
+++ b/doc/osx.t2t
@@ -90,7 +90,7 @@ Alternatively you can download and install Qt Open Source for MacOS from the
 
 https://www.qt.io/
 
-in the same version as referenced in the install script.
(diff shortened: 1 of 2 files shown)


doc/osx.t2t
```diff
@@ -90,7 +90,7 @@ Alternatively you can download and install Qt Open Source for MacOS from the
 
 https://www.qt.io/
 
-in the same version as referenced in the install script.
+in the same version as referenced in the install script. It must be installed in /opt/Qt
```
PeterPetrik, line 93: I do not see this change in the INSTALL ... I think easiest approach for such small change it to manually add it to both INSTALL and osx.t2t file (and revert all other changes in INSTALL file)

PR #36937, opened 2020-06-03
Fix GPKG delete from browser stored connection

### LINKED ISSUE #36930: Unable to delete a stored geopackage connection from browser
On Linux and Windows

qgis-bug-delete-gpkg

Related to https://github.com/qgis/QGIS/issues/36832

elpaso:
Fixes #36930 - also delete from stored connections
before refresh.

### DIFF
--- a/src/gui/providers/ogr/qgsgeopackageitemguiprovider.cpp
+++ b/src/gui/providers/ogr/qgsgeopackageitemguiprovider.cpp
@@ -177,8 +177,14 @@ void QgsGeoPackageItemGuiProvider::deleteGpkg()
         if ( ! name.isEmpty() )
         {
           QgsProviderMetadata *md { QgsProviderRegistry::instance()->providerMetadata( QStringLiteral( "ogr" ) ) };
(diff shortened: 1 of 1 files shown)


src/gui/providers/ogr/qgsgeopackageitemguiprovider.cpp
```diff
@@ -171,6 +173,13 @@ void QgsGeoPackageItemGuiProvider::deleteGpkg()
       else
       {
         QMessageBox::information( nullptr, title, tr( "GeoPackage deleted successfully." ) );
+        // If the deleted file was a stored connection, remove it too
+        if ( ! name.isEmpty() )
+        {
+          QgsProviderMetadata *md { QgsProviderRegistry::instance()->providerMetadata( QStringLiteral( "ogr" ) ) };
+          std::unique_ptr<QgsAbstractProviderConnection> conn { md->createConnection( name ) };
+          conn->remove( name );
```
nyalldawson, line 181: Can't we just do

    md->deleteConnection( name );

instead?

elpaso: yeah, cleaner.

PR #36945, opened 2020-06-03
Ensure that fields required by marker symbol backgrounds for labels are fetched when rendering

### LINKED ISSUE #36944: Setting a database field in font symbol for complex labelling won't render
**Describe the bug**
For complex labelling feature I need to use symbols markers.
They have the advantage to be able to combine multiple items altogether.
Also, they are currently the only "easy" solution to associate many SVG symbols, with different markers, fonts, texts and colours of any kind. 

Capture d’écran 2020-06-03 à 23 29 40

Capture d’écran 2020-06-03 à 23 29 23

In one of the latest nightly, it is even possible to select the font style like this (Thin Italic).

Capture d’écran 2020-06-03 à 23 25 22

Unfortunately, there is a bug in the Characters Expressions editor for the Font marker making almost impossible to use any of the fields of the current database. It may work sometimes, but most of the time it won't display anything. In the examples above, the route numbers did work.I don't know why. I tried with String values, Integer values, converted Integers into String in calculation fields and couldn't get any result. Text is universal UTF 8. I tried by creating a virtual field with calculation, keeping the same king and just using my working field as an entry and then used this virtual field, the route numbers disappeared!
As a result, You need to write text manually in the text box entry. It is acceptable for a single feature, but for a couple ones it won't. The only way will be by creating a SVG including all the necessary items for each feature. But all the benefices of a GIS are lost in this case and it's time consuming.

**How to Reproduce**

Quite easy to reproduce.

Select a feature, whatever the kind, it doesn't matter.

Create a new label feature and use ' ' as value (to avoid text rendering).

Then in Background create a new symbol marker.
Use a Font Marker and in the text box write a space (or it won't do anything) and then select any database filed in the Characters Expressions editor at the very bottom of the window.

Capture d’écran 2020-06-03 à 21 20 44

Valid your changes and refresh your map, nothing should be displayed unless you write text by hand. The length has no importance.

**QGIS and OS versions**

I use QGIS 3.10.6, but tried in nightly and with Windows version with no better results.
Mac OS 10.14.6 'Mojave'

**Additional context**

In order to save time, I join a pre-made labelling style.
[Test labels.qml.zip](https://github.com/qgis/QGIS/files/4726440/Test.labels.qml.zip)

nyalldawson:
Fixes #36944

### DIFF
--- a/python/core/auto_generated/textrenderer/qgstextbackgroundsettings.sip.in
+++ b/python/core/auto_generated/textrenderer/qgstextbackgroundsettings.sip.in
@@ -601,6 +601,13 @@ Updates the format by evaluating current values of data defined properties.
 .. versionadded:: 3.10
 %End
 
+    QSet<QString> referencedFields( const QgsRenderContext &context ) const;
+%Docstring
+Returns all field names referenced by the configuration (e.g. from data defined properties).
+
+.. versionadded:: 3.14
+%End
+
 };
 
 /************************************************************************
--- a/python/core/auto_generated/textrenderer/qgstextbuffersettings.sip.in
+++ b/python/core/auto_generated/textrenderer/qgstextbuffersettings.sip.in
@@ -248,6 +248,13 @@ Updates the format by evaluating current values of data defined properties.
 .. versionadded:: 3.10
 %End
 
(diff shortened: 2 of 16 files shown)


nyalldawson:
(not suitable for backport)

PR #36948, opened 2020-06-04
Auto focus search box when opening "add layer to legend" dialog

### LINKED ISSUE #36905: Cursor focus in "Add Layer to Legend" dialog should be in search input
**Feature description.**
When adding items to a legend in a layout, the "Add Layer to Legend" by default has no useful focus for keyboard input.

Please put the focus on the search input, so one can start typing a layer name right away. Alternatively set it to the list, so one can use the cursor keys to browse. I guess the preference is down to individual preference, I would love the former option right now ;)

This is a tiny UX thing but would be great to have.

QGIS 3.10 on Windows, sorry if this is changed already in newer versions or if it is a OS dependant thing.

Thank you!

### LINKED ISSUE #36906: Layouts: draw coordinates should allow background rectangles or buffers
**Feature description.**
In print layouts, when "draw coordinates" is activated, there should be an option to add background rectangles or buffers (oriented automatically with the "draw coordinate" labels --> see also #36904 ), to make the coordinates more legible against the map content.

In order to avoid situations like here:
grafik

Ultimately, it would be nice to enable selective masking to mask only dark symbol levels of the map content, but as a starter, background rectangles or buffers would be good enough. Selective masking in layouts interacting with map content might be too complex ...

nyalldawson:
Fixes #36905


DelazJ:
@nyalldawson is it the right issue report?

nyalldawson:
Eh - no :) Should be #36905

PR #36949, opened 2020-06-04
set time step when first loading meshlayer

vcloarec:
Before the new temporal framework, when loading a mesh layer the user could navigate through time with a time step adapted to the dataset, which often has a constant time step.
Now with the new temporal framework, the default time step is 1 hour, that is very often not adapted to mesh dataset. So user can be confused when loading a mesh layer and navigating with the temporal controller. Furthermore, the user has to find a way to know the time step to set the temporal controller (the more straightforward way is to open mesh layer properties and look at the time related with dataset through the static dataset options...)
So this default time step independent from the mesh layer can be seen as a regression for mesh layer.
To fix that, this PR set the time step equal to the first time step of the mesh layer when this mesh layer is the first temporal layer loaded.

### DIFF
--- a/src/gui/qgstemporalcontrollerwidget.cpp
+++ b/src/gui/qgstemporalcontrollerwidget.cpp
@@ -24,6 +24,7 @@
 #include "qgstemporalmapsettingswidget.h"
 #include "qgstemporalutils.h"
 #include "qgsmaplayertemporalproperties.h"
+#include "qgsmeshlayer.h"
 
 #include <QAction>
 #include <QMenu>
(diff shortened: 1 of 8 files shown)


src/gui/qgstemporalcontrollerwidget.cpp
```diff
@@ -498,6 +508,34 @@ void QgsTemporalControllerWidget::mRangeSetToAllLayersAction_triggered()
...
+    QgsUnitTypes::TemporalUnit unit = static_cast<QgsUnitTypes::TemporalUnit>( mTimeStepsComboBox->itemData( indexUnit ).toInt() );
+    value = timeStep.seconds() * QgsUnitTypes::fromUnitToUnitFactor( QgsUnitTypes::TemporalSeconds, unit );
+  }
+  while ( value > 10 ||  indexUnit == mTimeStepsComboBox->count() );
+
+  if ( fabs( log( value ) ) > fabs( log( previousValue ) ) )
+  {
+    indexUnit--;
+    value = previousValue;
+  }
```
nyalldawson, line 531: Can you move this logic out and sit it in QgsInterval?

vcloarec: The whole logic? Not sure because that depends on the units present in the combo box (not sure all unit are present).
Maybe with providing the list of unit.

PeterPetrik:
great, this is the last piece missing for the new temporal controller for the MeshLayer users!

src/gui/qgstemporalcontrollerwidget.cpp
nyalldawson: Ah - I missed that. Could we at least get some comments in here explaining the logic?

vcloarec: Finally, the approach to select the time step unit has changed. Indeed, first approach was wrong and only around numerical value. Best to search with value AND significant digits.

nyalldawson:
Ok, merge away!

nyalldawson (approved):

PR #36954, opened 2020-06-04
warn if there is no style associated with a rule (fix #23048)

### LINKED ISSUE #23048: Layer panel: "Edit Symbol" does nothing if no symbol defined for a rule (in rule-based rendering)
Original Redmine Issue: [15106](https://issues.qgis.org/issues/15106)
Affected QGIS version: 3.0.0
Redmine category:unknown

---
Create a rule-based rendering mode and set no symbol to one of the rules
In the Layers panel, right-click on the "no-symbolized" rule and select "Edit Symbol"
Nothing happens.

I think it should pop-up a message that there's no symbol associated to the rule (or, more advanced, ask if you want to associate one symbol and then open the corresponding dialog)

alexbruy:
## Description
If layer styled with rule-based renderer and some rules have no symbol associated with them user still have "Edit Symbol" entry in the legend entry context menu. Clicking on "Edit Symbol" in this case does nothing (which is correct as there is no symbol) and can be confusing.

Proposed PR adds a popup message explaining that corresponding rule does not have any symbol associated with it. Fixes #23048.

Probably can be backported to 3.10.

### DIFF
--- a/src/app/qgsapplayertreeviewmenuprovider.cpp
+++ b/src/app/qgsapplayertreeviewmenuprovider.cpp
@@ -13,7 +13,6 @@
  *                                                                         *
  ***************************************************************************/
 #include <QClipboard>
-#include <QMessageBox>
 
 #include "qgsapplayertreeviewmenuprovider.h"
 
@@ -43,6 +42,7 @@
 #include "qgsmaplayerstylecategoriesmodel.h"
 #include "qgssymbollayerutils.h"
 #include "qgsxmlutils.h"
+#include "qgsmessagebar.h"
 
 
 QgsAppLayerTreeViewMenuProvider::QgsAppLayerTreeViewMenuProvider( QgsLayerTreeView *view, QgsMapCanvas *canvas )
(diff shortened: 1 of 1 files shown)


src/app/qgsapplayertreeviewmenuprovider.cpp
```diff
@@ -897,7 +896,10 @@ void QgsAppLayerTreeViewMenuProvider::editSymbolLegendNodeSymbol( const QString
 
   const QgsSymbol *originalSymbol = node->symbol();
   if ( !originalSymbol )
+  {
+    QMessageBox::information( nullptr, tr( "No Symbol" ), tr( "There is no symbol associated with the rule." ) );
```
nyalldawson, line 900: Could we use the QgisApp::instance()->messageBar() instead?

alexbruy: Done.

src/app/qgsapplayertreeviewmenuprovider.cpp
```diff
@@ -897,7 +896,10 @@ void QgsAppLayerTreeViewMenuProvider::editSymbolLegendNodeSymbol( const QString
 
   const QgsSymbol *originalSymbol = node->symbol();
   if ( !originalSymbol )
+  {
+    QgisApp::instance()->messageBar()->pushInfo( tr( "No Symbol" ), tr( "There is no symbol associated with the rule." ) );
```
nyalldawson, line 900:
```suggestion
    QgisApp::instance()->messageBar()->pushWarning( tr( "No Symbol" ), tr( "There is no symbol associated with the rule." ) );
```

alexbruy: Done

PR #36958, opened 2020-06-04
Fix capitalization/spacing in ArcGIS MapServer/FeatureServer strings

nyalldawson:
This should be "MapServer"/"FeatureServer", not "Map Server"/"Feature Server"

### DIFF
--- a/src/ui/qgisapp.ui
+++ b/src/ui/qgisapp.ui
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>847</width>
+    <width>1064</width>
     <height>506</height>
    </rect>
   </property>
@@ -16,7 +16,7 @@
     <rect>
      <x>0</x>
      <y>0</y>
-     <width>847</width>
+     <width>1064</width>
      <height>24</height>
     </rect>
    </property>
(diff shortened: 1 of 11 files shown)


nirvn:
For the record, I think its called ArcGIS Feature *Service*, not server (same goes for ArcGIS Map Service.

nyalldawson:
> For the record, I think its called ArcGIS Feature Service, not server (same goes for ArcGIS Map Service.

I'm not sure. I think **maybe** the protocol is called "feature service", but an implementation of the protocol is a "feature server". Googling leads to a mix of the two terms used within the ArcGIS docs...

nyalldawson:
Ok, so I think we should technically use "ArcGIS FeatureService Server". But that's incredibly clunky!

nyalldawson:
Let's revise that: technically I think it should be "ArcGIS Server Feature Service", but then there IS 3rd party implementations of the protocol which should be supported in qgis already. So let's go with "ArcGIS Feature Service"...

nirvn:
+1, you settled for my favorite option.

src/ui/qgisapp.ui
```diff
@@ -2828,10 +2828,10 @@ Acts on the currently active layer only.</string>
      <normaloff>:/images/themes/default/mActionAddAfsLayer.svg</normaloff>:/images/themes/default/mActionAddAfsLayer.svg</iconset>
    </property>
    <property name="text">
-    <string>Add Ar&amp;cGIS FeatureServer Layer…</string>
+    <string>Add Ar&amp;cGIS Feature Service Layer…</string>
```
havatv, line 2831: What is the `&amp;` doing here?

DelazJ: it's a shortcut

PR #36967, opened 2020-06-04
QgsFeature default constructor set fid to invalid

### LINKED ISSUE #36962: Constraints: unique constraint not correctly evaluated when 0
Using the same dataset from https://github.com/qgis/QGIS/issues/36167 with a slightly different join configuration (see the second image) inserting a new feature "ogc_fid" `0` is not recognized as a constraint violation even if a feature with "ogc_fid" `0` is already present in the database table, the expected behavior is to see the orange cross with the unique constraint violation tooltip.

immagine

immagine

elpaso:
## I know it can be a dangerous move, let's see how the test suite reacts

The QgsFeature default constructors initialized
feature id to 0, which is a valid feature id
instead of initializing it fo FID_NULL.

This was just wrong and broke the validator for
UNIQUE constraints in case a feature with fid 0
existed in the data provider.

Fixes #36962


m-kuhn:
This is the right thing to do.
Fingers crossed the test suite can be happified.

elpaso:
@m-kuhn I think this change will make surface a few subtle related issues, I've just found one in the json exporter where feature id is set to 0 where it should really be `null`.

m-kuhn:
Totally agree.

Sorry to play with the description, I found it pretty confusing at first (as I expect the PR description to say what it does right after, and the Issue description to say what it does wrong before)

elpaso:
> Totally agree.
> 
> Sorry to play with the description, I found it pretty confusing at first (as I expect the PR description to say what it does right after, and the Issue description to say what it does wrong before)

That make sense :) 
Sorry if I keep forgetting all the unwritten rules, I'm getting short or RAM lately ...

elpaso:
@nyalldawson I know this last commit is not right but I couldn't dig deeper in the test to check if it is a real issue in the importer code or is it a test artifact, you are more into layouts and atlases than everyone, mind having a look?

Also, we should define what a decent string representation for a NULL feature id would be, I though of a blank, but then you won't see the output without a feature, maybe `NULL` or `None` or `undefined` would be better.

src/core/expression/qgsexpressioncontextutils.cpp
```diff
@@ -525,7 +526,7 @@ QgsExpressionContextScope *QgsExpressionContextUtils::layoutScope( const QgsLayo
     QgsFeature atlasFeature = layout->reportContext().feature();
     scope->setFeature( atlasFeature );
     scope->addVariable( QgsExpressionContextScope::StaticVariable( QStringLiteral( "atlas_feature" ), QVariant::fromValue( atlasFeature ), true ) );
-    scope->addVariable( QgsExpressionContextScope::StaticVariable( QStringLiteral( "atlas_featureid" ), atlasFeature.id(), true ) );
+    scope->addVariable( QgsExpressionContextScope::StaticVariable( QStringLiteral( "atlas_featureid" ), FID_IS_NULL( atlasFeature.id() ) ? 0 : atlasFeature.id(), true ) );
```
nyalldawson, line 529:
```suggestion
    scope->addVariable( QgsExpressionContextScope::StaticVariable( QStringLiteral( "atlas_featureid" ), FID_IS_NULL( atlasFeature.id() ) ? QVariant() : atlasFeature.id(), true ) );
```

src/core/expression/qgsexpressioncontextutils.cpp
```diff
@@ -599,7 +600,7 @@ QgsExpressionContextScope *QgsExpressionContextUtils::atlasScope( const QgsLayou
     QgsFeature atlasFeature = atlas->layout()->reportContext().feature();
     scope->setFeature( atlasFeature );
     scope->addVariable( QgsExpressionContextScope::StaticVariable( QStringLiteral( "atlas_feature" ), QVariant::fromValue( atlasFeature ), true ) );
-    scope->addVariable( QgsExpressionContextScope::StaticVariable( QStringLiteral( "atlas_featureid" ), atlasFeature.id(), true ) );
+    scope->addVariable( QgsExpressionContextScope::StaticVariable( QStringLiteral( "atlas_featureid" ), FID_IS_NULL( atlasFeature.id() ) ? 0 : atlasFeature.id(), true ) );
```
nyalldawson, line 603:
```suggestion
    scope->addVariable( QgsExpressionContextScope::StaticVariable( QStringLiteral( "atlas_featureid" ), FID_IS_NULL( atlasFeature.id() ) ? QVariant() : atlasFeature.id(), true ) );
```

nyalldawson:
>  are more into layouts and atlases than everyone, mind having a look?

I'd go with setting a QVariant() (null) value in this case.

nyalldawson:
Wow... surprisingly little breakage here! :smile:

nyalldawson:
@m-kuhn what's your thoughts on the risk level of this one? I'm wondering whether it should be deferred to 3.16 or not...

m-kuhn:
14 days left for 3.14.0. That's not a lot to detect and fix additional untested issues caused by this. If reasonably possible, this should be postponed IMHO.

nyalldawson:
Ok, agreed -- let's defer to 3.16

nyalldawson:
I think this is nearly good to go! The only change I want to see is an invalid QVariant used for the Atlas feature id variable when the feature isn't set..

olivierdalang:
> I think this is nearly good to go! The only change I want to see is an invalid QVariant used for the Atlas feature id variable when the feature isn't set..

For consistency, we may then also want other atlas_* variables to return a null QVariant ? (`@atlas_feature`, `@atlas_geometry`, etc.).

nyalldawson:
@olivierdalang agreed

elpaso:
@olivierdalang it wasn't completely clear to me what to do regarding `@atlas_feature`, `@atlas_geometry`, I think they are returning invalid feature and empty geometry for now, anyway this can be addressed in a follow-up PR.

PR #36971, opened 2020-06-04
[themes] Avoid styling tree view item's hover state by default

nirvn:
## Description

Styling QTreeview::item:hover simply doesn't play well with tree views having custom item delegates, let's only do that for a select number of tree views if desired (e.g. night mapping's layer tree view).

The PR also contains a fix whereas we didn't provide default styling for tree view checkboxes, resulting in a real bad black-on-black situation with Night Mapping.

Before vs. PR:
image


nirvn:
Turns out we also didn't style the table view checkboxes, again creating invisible unchecked checkboxes:
image

nirvn:
Other area improved by this fix:
image

m-kuhn (approved):
:lipstick:

PR #36973, opened 2020-06-04
Add missing NMEA patterns

m-kuhn:
Support for:

- GNVTG
- GNGST
- GNHDT

## Description

Variables like HDOP etc. are missing depending on the positioning device in use.

On a more general note, we should likely ignore the second character `N` or `P` in the matching, it is part of the talker ID and ignored by the nmea library as well. It might even be better to completely ignore the first two characters `GN` given the talker id list on https://gpsd.gitlab.io/gpsd/NMEA.html#_gga_global_positioning_system_fix_data , there are some that don't have `G` as first character.

### DIFF
--- a/external/nmea/parse.c
+++ b/external/nmea/parse.c
@@ -283,19 +283,28 @@ int nmea_parse_GPGST( const char *buff, int buff_sz, nmeaGPGST *pack )
 
   nmea_trace_buff( buff, buff_sz );
 
+  char type;
+
   if ( 8 != nmea_scanf( buff, buff_sz,
-                        "$GPGST,%s,%f,%f,%f,%f,%f,%f,%f*",
+                        "$G%CGST,%s,%f,%f,%f,%f,%f,%f,%f*",
+                        &( type ),
                         &( time_buff[0] ),
                         &( pack->rms_pr ), &( pack->err_major ), &( pack->err_minor ), &( pack->err_ori ),
                         &( pack->sig_lat ), &( pack->sig_lon ), &( pack->sig_alt ) ) )
   {
-    nmea_error( "GPGST parse error!" );
+    nmea_error( "G?GST parse error!" );
+    return 0;
+  }
+
+  if ( type != 'P' && type != 'N' )
+  {
+    nmea_error( "G?GST invalid type " );
     return 0;
   }
 
   if ( 0 != _nmea_parse_time( &time_buff[0], ( int )strlen( &time_buff[0] ), &( pack->utc ) ) )
   {
-    nmea_error( "GPGST time parse error!" );
+    nmea_error( "G?GST time parse error!" );
     return 0;
   }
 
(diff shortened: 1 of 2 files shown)


src/core/gps/qgsnmeaconnection.cpp
```diff
@@ -115,7 +115,7 @@ void QgsNmeaConnection::processStringBuffer()
           mStatus = GPSDataReceived;
           QgsDebugMsgLevel( QStringLiteral( "*******************GPS data received****************" ), 2 );
         }
-        else if ( substring.startsWith( QLatin1String( "$GPVTG" ) ) )
+        else if ( substring.startsWith( QLatin1String( "$GPVTG" ) ) || substring.startsWith( QLatin1String( "$GNVTG" ) ) )
```
nyalldawson, line 118: This won't have any impact unless you also modify `nmea_parse_GPVTG`. See e.g. 299c61292b58602c4d6f59628b3bafd6125abd19 for an example on what you'd need to change in parse.c here!

m-kuhn: Thanks @nyalldawson 
I have only read the G?GGA parser and extrapolated that the others would have been patched also.
As noted in earlier comments, not completely convinced for failing on `type not in 'PN'` but kept the style for consistency.

external/nmea/parse.c
```diff
@@ -283,19 +283,28 @@ int nmea_parse_GPGST( const char *buff, int buff_sz, nmeaGPGST *pack )
 
   nmea_trace_buff( buff, buff_sz );
 
+  char type;
+
   if ( 8 != nmea_scanf( buff, buff_sz,
```
nyalldawson, line 288: you need to bump these expected result numbers too... have you tested this?

m-kuhn: not at all. I don't have any means to test this, so the idea was to ship and let a user with the possibilities test the nightlies. Sorry for not being more helpful.

m-kuhn: ... I hope to be able to work on unit tests for this as part of something that could hit our pipeline.

nyalldawson: :smile: (just don't let @roya0045 hear that... shhhhhh :shushing_face: )

nyalldawson (approved):

external/nmea/parse.c
roya0045: @nyalldawson I'm not sure if I shoul be insulted by this, but its fine. We can all be bumblefucks at times.

m-kuhn: I think it was a friendly pat to mostly remind me of shortcomings I've imputed to others 😜

nyalldawson: Don't be... I'm insulting @m-kuhn and calling him out for hypocrisy, not insulting you! 😁 (And @m-kuhn knows me well enough to know that 90% of what I say is firmly tongue in cheek!!)

roya0045: This is what I was thinking. No offence was taken ;)  I am also well aware of my shortcoming and always find refreshing to see someone else do things blindly from time to time.

m-kuhn:
I received confirmation that it's functional, let's backport !

PR #36975, opened 2020-06-04
Warn user when raster or vector renderere does not have any classes defined (fix #21075)  

alexbruy:
## Description
If graduated/categorized/rule-based vector renderer does not have any classes defined (e.g. user removed all classes or never pressed "Classify" button) layer won't be rendered on the canvas and labels won't be drawn. Same happens with paletted/pseudocolor raster renderers.

Prosed PR adds a warning before closing dialog/applying changes, so user can either refine settings or continue without classes.

Fixes #21075.

### DIFF
--- a/src/gui/raster/qgsrasterlayerproperties.cpp
+++ b/src/gui/raster/qgsrasterlayerproperties.cpp
@@ -49,8 +49,10 @@
 #include "qgsrasterrenderer.h"
 #include "qgsrasterrendererregistry.h"
 #include "qgsrasterresamplefilter.h"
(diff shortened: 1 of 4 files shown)


nyalldawson:
This approach only helps when setting symbology through the dialog, right?

I'm wondering if a more flexible approach (which would also help with the styling dock) would be to add a messagebar (or just a warning label) inside the associated renderer widgets and just show the warning directly in the renderer widget itself...

alexbruy:
Ok, I will try to look at this.

PR #36981, opened 2020-06-04
[Feature] texturing support for vector layer

NEDJIMAbelgacem:
###  Added texturing support for QGIS 3D
- Now from the symbol properties you can select an image to be displayed on the surfaces of buildings as diffuse map.
- The user can also change the scale of textures from the symbol properties as you can see in these screenshots:

Screenshot_20200604_182756
Screenshot_20200604_182723
- Some visual issue may appear when using textures. these issues relates to the normals of surfaces that you may be able to fix by checking "add back faces" and playing around with the invert normals and culling mode in the symbol properties.
- I modified a bit of the QgsTessellatedPolygonGeometry constructor so that every Qt3DRender::QAttribute of the geometry will be constructed properly because setting them after constructing the object can be confusing and we don't want it to construct objects that we won't need later on. I also provided default parameters for the constructor so we don't have problems with existing code.


m-kuhn:
Nice start!

Welcome Belgacem.

src/3d/symbols/qgspolygon3dsymbol_p.cpp
```diff
@@ -262,28 +260,81 @@ static Qt3DRender::QCullFace::CullingMode _qt3DcullingMode( Qgs3DTypes::CullingM
...
+      auto renderPasses = ( *tit )->renderPasses();
+      for ( auto rpit = renderPasses.begin(); rpit != renderPasses.end(); ++rpit )
+      {
+        Qt3DRender::QCullFace *cullFace = new Qt3DRender::QCullFace;
+        cullFace->setMode( _qt3DcullingMode( symbol.cullingMode() ) );
+        ( *rpit )->addRenderState( cullFace );
+      }
+    }
+    Qt3DRender::QTextureImage *textureImage = new Qt3DRender::QTextureImage;
+    textureImage->setSource( QUrl::fromLocalFile( textureFilePath ) );
```
nyalldawson, line 284: Looking at https://stackoverflow.com/questions/50944249/create-qtextureimage-from-qimage , it seems there's a means to directly construct textures from a QImage object. If this is the case, then you could use `QgsImageCache::pathAsImage() ` (via `QgsApplication::imageCache()->pathAsImage(...)`) here instead of directly loading from a local file.

There's two benefits to this approach

1. If the image has been previously loaded then it'll save a bunch of work
2. QgsImageCache supports base64 encoded image paths. This allows for embedding images directly inside a QgsProject, so that the project becomes portable and there's no path issues when you share the project (or QML style) with others. To enable this you'd also need to chance the `textureFile` widget in the ui file to a `QgsImageSourceLineEdit` widget (which adds the UI for users to embed pictures).

nyalldawson:
Wow, you've **really** "hit the ground running"!! Great start to GSOC @NEDJIMAbelgacem !

src/3d/symbols/qgspolygon3dsymbol_p.cpp
NEDJIMAbelgacem: I didn't know about that. I will change the textureFile widget to QgsImageSourceLineEdit and use QgsImageCache and make another commit.
Thanks for the suggestion!

ismailsunni:
Awesome @NEDJIMAbelgacem 
You can also add some gifs to show the 3D or how the setting works  (nice to have/see but not mandatory for sure)

timlinux:
Looking awesome @NEDJIMAbelgacem !

src/3d/symbols/qgspolygon3dsymbol_p.cpp
```diff
@@ -262,28 +260,81 @@ static Qt3DRender::QCullFace::CullingMode _qt3DcullingMode( Qgs3DTypes::CullingM
...
+    material->setSpecular( symbol.material().specular() );
+    material->setShininess( symbol.material().shininess() );
+    material->setTextureScale( symbol.material().textureScale() );
+
+    if ( isSelected )
+    {
+      // update the material with selection colors
+      // TODO : dampen the color of diffuse texture
+//      mat->setDiffuse( context.map().selectionColor() );
+      material->setAmbient( context.map().selectionColor().darker() );
```
wonder-sk, line 299: How does it look with just modified ambient color? maybe that's enough and we don't need to worry about modification of diffuse texture? (or just make the ambient color lighter if needed to make it stand out more?)

src/3d/symbols/qgspolygon3dsymbol_p.cpp
```diff
@@ -262,28 +260,81 @@ static Qt3DRender::QCullFace::CullingMode _qt3DcullingMode( Qgs3DTypes::CullingM
...
+    }
 
-  // front/back side culling
-  auto techniques = material->effect()->techniques();
-  for ( auto tit = techniques.constBegin(); tit != techniques.constEnd(); ++tit )
+    retMaterial = material;
```
wonder-sk, line 302: how about we just `return material;` here? (and in the "else" branch as well)

src/3d/symbols/qgspolygon3dsymbol_p.cpp
```diff
@@ -262,28 +260,81 @@ static Qt3DRender::QCullFace::CullingMode _qt3DcullingMode( Qgs3DTypes::CullingM
...
+    for ( auto tit = techniques.constBegin(); tit != techniques.constEnd(); ++tit )
+    {
+      auto renderPasses = ( *tit )->renderPasses();
+      for ( auto rpit = renderPasses.begin(); rpit != renderPasses.end(); ++rpit )
+      {
+        Qt3DRender::QCullFace *cullFace = new Qt3DRender::QCullFace;
+        cullFace->setMode( _qt3DcullingMode( symbol.cullingMode() ) );
+        ( *rpit )->addRenderState( cullFace );
+      }
+    }
```
wonder-sk, line 321: can we move this code to set culling mode to a separate function - that would avoid having two copies of it...

src/core/qgstessellator.cpp
```diff
@@ -436,21 +459,66 @@ void QgsTessellator::addPolygon( const QgsPolygon &polygon, float extrusionHeigh
...
+      const float scaleY = !mBounds.isNull() ? 10000.0 / mBounds.height() : 1.0;
+
+      // subtract ptFirst from geometry for better numerical stability in triangulation
+      // and apply new 3D vector base if the polygon is not horizontal
+
+      std::unique_ptr<QgsPolygon> polygonNew( _transform_polygon_to_new_base( polygon, pt0, toNewBase.get(), scaleX, scaleY ) );
+      QList< std::vector<p2t::Point *> > polylinesToDelete;
+      QHash<p2t::Point *, float> z;
+
+      _ringToPoly2tri( qgsgeometry_cast< const QgsLineString * >( polygonNew->exteriorRing() ), polyline, mNoZ ? nullptr : &z );
```
wonder-sk, line 500: We are working with a single triangle here, so do we need to run triangulation here? Can't we just use the coordinates from `polygonNew` ?

src/core/qgstessellator.cpp
```diff
@@ -436,21 +459,66 @@ void QgsTessellator::addPolygon( const QgsPolygon &polygon, float extrusionHeigh
...
+        // but Qt does not have good support for it, so using 4x4 matrix instead.
+        toNewBase.reset( new QMatrix4x4(
+                           pXVector.x(), pXVector.y(), pXVector.z(), 0,
+                           pYVector.x(), pYVector.y(), pYVector.z(), 0,
+                           pNormal.x(), pNormal.y(), pNormal.z(), 0,
+                           0, 0, 0, 0 ) );
+
+        // our 3x3 matrix is orthogonal, so for inverse we only need to transpose it
+        toOldBase.reset( new QMatrix4x4( toNewBase->transposed() ) );
+      }
```
wonder-sk, line 485: It would be good to move the rotation code to a new function to avoid code duplication...

wonder-sk (commented):
It's looking great - nice work!

I have left a couple of suggestions below on how to further improve the code...

src/3d/symbols/qgspolygon3dsymbol_p.cpp
NEDJIMAbelgacem: Actually when I tried to test this, but in no 3D entities are selected :(

src/3d/symbols/qgspolygon3dsymbol_p.cpp
NEDJIMAbelgacem: I always like seeing a function returns something at the end of it. It's just a code style preference  though.

NEDJIMAbelgacem:
Hi,
The separation between roofs and walls in now complete although it's not very good now due to technical limitations of how we classify a vector layer feature's geometry as part of wall or roof.
I made a new combo box that selects which facade we want to render (wall, roofs or walls and roofs). To know if a triangle is a wall or a roof I calculate the degrees between the triangle's normal and the horizontal plane, if the angle is less than 10 degrees the triangle is a part of a roof, otherwise it's part of a wall.
To take advantage of the walls and roofs separation, the user needs to create 2 rules in the rule based rendering each one having it's own symbol.
The user can also change the scale and rotation of textures.
I noticed that the Phong material gets reused in parts where there is no need to use textures. I hide the texture UI widgets every time except when it is needed for the vector layer.
Screenshot_2020-06-27_14-19-47

wonder-sk (approved):
Very nice! We will probably still need to do some further work on texturing, but this one is good to go...

PR #36988, opened 2020-06-05
[expression] Add optional language parameter to format_number() to control locale across different systems

nirvn:
## Description

The `format_number()` expression function relies on the user system (or overwritten) locale to format a given string. While it's mostly fine for single users, sharing project(s) across multiple users having different system locale can lead to broken expectations (i.e. variance in thousands and decimals separators). In worse case scenarios, it can lead to broken expressions expecting a given separator.

The fix is easy: add an optional language parameter for the function, just like we do with `format_date()`. It also has the nice side effect of harmonizing those functions.

### DIFF
--- a/tests/src/core/testqgsexpression.cpp
+++ b/tests/src/core/testqgsexpression.cpp
@@ -1246,6 +1246,11 @@ class TestQgsExpression: public QObject
       QTest::newRow( "format_number large" ) << "format_number(9000000.0,0)" << false << QVariant( "9,000,000" );
       QTest::newRow( "format_number many decimals" ) << "format_number(123.45600,4)" << false << QVariant( "123.4560" );
       QTest::newRow( "format_number no decimals" ) << "format_number(1999.567,0)" << false << QVariant( "2,000" );
+#if QT_VERSION >= QT_VERSION_CHECK( 5, 12, 0 )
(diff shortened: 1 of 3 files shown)


tests/src/core/testqgsexpression.cpp
```diff
@@ -1246,6 +1246,7 @@ class TestQgsExpression: public QObject
       QTest::newRow( "format_number large" ) << "format_number(9000000.0,0)" << false << QVariant( "9,000,000" );
       QTest::newRow( "format_number many decimals" ) << "format_number(123.45600,4)" << false << QVariant( "123.4560" );
       QTest::newRow( "format_number no decimals" ) << "format_number(1999.567,0)" << false << QVariant( "2,000" );
+      QTest::newRow( "format_number language parameter" ) << "format_number(123456.789,2,'fr')" << false << QVariant( "123 456,79" );
```
nyalldawson, line 1249: You may experience similar difficulties as https://github.com/qgis/QGIS/pull/36860 due to fr locale. Might be easiest to just use german locale instead.

nyalldawson (approved):

tests/src/core/testqgsexpression.cpp
nirvn: Same problem with de.

PR #36998, opened 2020-06-05
Homogenize window title separator

### LINKED ISSUE #35761: Omogenize window title
## OSGeo4W64 Win 10 - QGIS master

In my opinion, all hyphens should be the same, for example `em-dash` (Field calculator):

image

cc @nyalldawson

elpaso:
Fixes #35761

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -547,7 +547,7 @@ static void setTitleBarText_( QWidget &qgisApp )
   }
   if ( !caption.isEmpty() )
   {
-    caption += QLatin1String( " - " );
+    caption += QStringLiteral( " — " );
   }
   if ( QgsProject::instance()->isDirty() )
     caption.prepend( '*' );
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -547,7 +547,7 @@ static void setTitleBarText_( QWidget &qgisApp )
   }
   if ( !caption.isEmpty() )
   {
-    caption += QStringLiteral( " — " );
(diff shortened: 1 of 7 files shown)


src/app/qgisapp.cpp
```diff
@@ -547,7 +547,7 @@ static void setTitleBarText_( QWidget &qgisApp )
   }
   if ( !caption.isEmpty() )
   {
-    caption += QLatin1String( " - " );
+    caption += QStringLiteral( " — " );
```
nyalldawson, line 550: This results in corrupt strings on windows builds... You need to use the character code instead

elpaso: Sure (like for the three dots) I guess. But let me ask you a questions, is there anything particular in this line of code or should we change the other 27 occurrences of that character, because I just copy-pasted that character from one of the many other another pre-existing string in the code and I don't understand why it suddenly became a problem after my changes if it wasn't before.
```
$ grep -r '—' src|egrep 'cpp:'|wc -l
27
```
... or maybe you are just saying that it **is always been a problem**  but this is an opportunity to fix it?

elpaso: I see it's been used mostly inside tr:
```
QObject::tr( "Warning — this algorithm modifies......
```
does that need replaing too?

or is it just when it is inside a `QStringLiteral` ?

nyalldawson: It's fine when it's inside tr, and it's fine from Python -- it's only QStringLiteral and raw characters strings which are an issue. That said, this **may** have changed since https://github.com/qgis/QGIS/pull/35737. I'd probably still err on the side of caution though!

nyalldawson (approved):

PR #37006, opened 2020-06-05
Fix labeling aligment DD activation signals

### LINKED ISSUE #37003: Labeling: data-defined alignment grey-out activation issue
**Describe the bug**
In the label settings --> Tab "Placement" --> Section "Data Defined" there is the "Alignment" setting (horizontal and vertical). These two properties are greyed out as long as no attribute or expression is assigned at Coordinate X and Y.

However, once you assign these values, the Alignment properties are still greyed out. One has to close the layer properties dialogue with "OK" and then reopen the Layer properties dialog. Only then, the alignment properties "horizontal" and "vertical" can be set.

image

Issue: I would expect the alignment properties to be made available as soon as both "X" and "Y" is defined without having to close and reopen the dialog. The same issue appears in the "Layer Styling" side panel. But there it is even worse. Opening and closing of the panel doesn't help. The alignment properties are out of reach in the "Layer Styling" panel.

**How to Reproduce**

1. Load any Point layer
2. Open Layer Properties Dialogue
3. Go to Label settings --> Tab "Placement" --> Section "Data Defined" and set expressions "$x" for "X" and "$y" for Y
4. --> the Alignment properties are still greyed out
5. Close the layer properties dialog by using "OK"
6. Reopen the Layer properties and check Label settings --> Tab "Placement" --> Section "Data Defined". Now the "horizontal" an "vertical" alignment properties are available.

In older QGIS version (don't ask me when) this used to work fine, but since several versions now (at least 1-2 years) this seems to be broken. It is quite annoying if you work a lot with data-defined labeling.

elpaso:
Fixes #37003


m-kuhn (approved):
Super annoying one, thanks !

andreasneumann:
> Super annoying one, thanks !

Yes - thanks a lot for fixing this!

nyalldawson (approved):

PR #37013, opened 2020-06-06
[dbmanager] Qgis -> QGIS

nirvn:
## Description

Show some respect DB Manager! :wink:


nirvn:
Went for a broader hunt looking for Qgis/QGis.

m-kuhn (approved):
How could that go unnoticed for so long

PR #37021, opened 2020-06-07
reduce svg file size using scour

lhirlimann:
using scour to reduce svg file size not as nice as using optipng but it does the job.


andreasneumann:
@lhirlimann thanks for trying this.

But in my experience scour can potentially destroy SVG (not according to the specification), but because qt5 only support a subset of the SVG tiny profile.

Before merging this, each file - one by one - needs to be controlled if it still displays fine in QGIS (qt5 engine).

I've seen it (more than once) that an SVG displayed fine in the web browsers, but not in qt5's crippled SVG renderer.

lhirlimann:
> @lhirlimann thanks for trying this.
> 
> But in my experience scour can potentially destroy SVG (not according to the specification), but because qt5 only support a subset of the SVG tiny profile.

:( Is there any easy way to do that ?

andreasneumann:
> FWIW I've been using default settings and didn't try to be has hard as one can be.

Probably 99% are fine. But it is still a bit dangerous.

What scour options did you use?

> :( Is there any easy way to do that ?
don't know. You'd have to render the old and new version side by side and compare it visually. Or compare it automatically pixel by pixel. But I don't have a script for that, unfortunately.

nyalldawson:
> But in my experience scour can potentially destroy SVG (not according to the specification), but because qt5 only support a subset of the SVG tiny profile.

nyalldawson:
Some things in particular we need to look out for
- new line breaks in the xml. While these should be harmless, the positioning of newline characters inside the xml DOES sometimes break Qt's rendering of them (especially in text elements)
- make sure that "param" style attributes don't get broken. While QGIS supports these, they were never made an official part of the SVG specification (:cry: ), so 3rd party tools may not be aware of them and consequently break them/strip them out

lhirlimann:
> > FWIW I've been using default settings and didn't try to be has hard as one can be.
> 
> Probably 99% are fine. But it is still a bit dangerous.
> 
> What scour options did you use?
None. Just scour -i xxx.svg -o .svg.new >>scour log, then took the files that had been compressed and mloved the new on top of the old .svg.

> > :( Is there any easy way to do that ?
> > don't know. You'd have to render the old and new version side by side and compare it visually. Or compare it automatically pixel by pixel. But I don't have a script for that, unfortunately.

I'll try to take some time this week to see how things changed.

alexbruy:
Please note that UI icons were already optimized with `svgcleaner` which follows specifications and does not break rendering.

lhirlimann:
> Please note that UI icons were already optimized with `svgcleaner` which follows specifications and does not break rendering.

I can try svgcleaner too do you remember how you proceeded with these ?

andreasneumann:
I guess we should discuss first, what is "safe" to clean and what not - what svgcleaner options would be good to use (see https://github.com/RazrFalcon/svgcleaner/blob/master/docs/svgcleaner.adoc)

Also, if we work on this, we should make sure that every file has
- a proper XML header
- a viewBox attribute so it properly scales

I wouldn't change the geometry precision, as this seems to depend a lot on the viewBox and some of the files have very small and some very large viewBoxes - a general rules doesn't seem to apply to all of them. If we change geometry precision, it needs to be done file by file and evaluated.

But certainly a lot of whitespace can be removed and duplicate definitions in <defs/> should be reemoved (see the svgcleaner options on removing duplicates. Unused stuff in the <defs> section could also be removed.

If you want, I can have a better look tonight which svgcleaner options should be safe and which not.

lhirlimann:
> If you want, I can have a better look tonight which svgcleaner options should be safe and which not.


Sure, that would help.

andreasneumann:
I did a short review of the svgcleaner options:

Options that I would use, consider harmless and useful:
--remove-unused-defs
--remove-dupl-lineargradient
--remove-dupl-fegaussianblur
--ungroup-groups
--remove-text-attributes
--remove-needless-attributes
--append-newline
--simplify-transforms

useful, but have to be tested if they work with QtSVG:
--merge-gradients 
--regroup-gradient-stops
--remove-invalid-stops

I would refrain from removing metadata, desc elements and comments.

Path and numeric precision manipulations would have to properly tested before using them. I know for sure that QtSVG isn't conformant with all of the more exotic rules in the SVG path syntax. The problem with numeric precision is that some SVG files use very large, same very tiny viewBoxes. Thus a general rule for reducing numeric precision would first have to analyze the coordinate space and viewBoxes of the files and would have to use individual precisions per file.

lhirlimann:
> Options that I would use, consider harmless and useful:
> --remove-unused-defs
> --remove-dupl-lineargradient
> --remove-dupl-fegaussianblur
> --ungroup-groups
> --remove-text-attributes
> --remove-needless-attributes
> --append-newline
> --simplify-transforms

I'll start with that ! thanks.

andreasneumann:
> I'll start with that ! thanks.

Great.

If you have time, investigating optimizations of the path string (https://github.com/RazrFalcon/svgcleaner/blob/master/docs/svgcleaner.adoc#paths) (e.g. stripping out whitespace, removing repeated path commands, using relative instead of absolute coordinates, etc.) would have a lot of potential to reduce file size. Definitely worth looking into.

We just have to verify that these are QtSVG compatible. My guess is that the majority of the optimizations would be compatible, but probably not all of them. I vaguely remember that relative "m" commands failed 

It is also interesting to see that svgcleaner specifically mentions some optimizations that aren't compatible with QtSVG. If I had a lot of funds, I would definitely invest into better SVG and PDF support in Qt ... but that's another story.

@lhirlimann If you want, then I can have a look into the path optimizations to investigate if they are compatible with QtSVG. I can do that until the coming weekend if you can be patient.

lhirlimann:
I can be very patient if it saves a few more kb. I'll wait.

lhirlimann:
>@lhirlimann If you want, then I can have a look into the path optimizations to investigate if they are compatible with QtSVG. I can do that until the coming weekend if you can be patient.

Did you  have time to look?

PR #37022, opened 2020-06-07
Fix #36983 arguments of is_selected are flipped

### LINKED ISSUE #36983: Documentation for function is_selected is wrong
As discovered in [this stackexchange question](https://gis.stackexchange.com/questions/363873/qgis-expression-get-feature-will-not-evaluate-feature-when-called-within-is?noredirect=1#comment595413_363873), the documentation for the _is_selected_ function appears to be incorrect:
image

The order of parameters is listed incorrectly.
`is_selected([feature=current feature][,layer=current layer])`
should be
`is_selected([layer=current layer][,feature=current feature])`

And accordingly, the example given should be changed:
`is_selected(get_feature('streets', 'name', "street_name"), 'streets')`
should be
`is_selected('streets', get_feature('streets', 'name', "street_name"))`

As seen in QGIS 3.10.3

rduivenvoorde:
Fixes https://github.com/qgis/QGIS/issues/36983

I think this could be backported to 3.12 and 3.10 too?

In the help text the arguments for the is_selected function were flipped. I checked, also the example as shown was not working.

Also added a 'True' in the description, which I think was missing.

Old situation:

oldsituation

New situation:

newsituation

### DIFF
--- a/resources/function_help/json/is_selected
+++ b/resources/function_help/json/is_selected
@@ -7,7 +7,7 @@
             "variant": "No parameters",
             "variant_description": "If called with no parameters, the function will return true if the current feature in the current layer is selected.",
             "arguments": [ { "arg": "", "description": "No parameters" } ],
-            "examples": [ { "expression": "is_selected()", "returns" : "True if the current feature in current layer is selected." } ]
+            "examples": [ { "expression": "is_selected()", "returns" : "True if the current feature in the current layer is selected." } ]
         },
         {
             "variant": "One 'feature' parameter",
@@ -23,5 +23,4 @@
         }
   ]
  }
-
 
--- a/resources/function_help/json/is_selected
+++ b/resources/function_help/json/is_selected
@@ -19,7 +19,7 @@
             "variant" : "Two parameters",
             "variant_description": "If the function is called with both a layer and a feature, it will return true if the specified feature from the specified layer is selected.",
             "arguments": [ { "arg": "layer", "description": "The layer (or its ID or name) on which the selection will be checked." }, { "arg": "feature", "description": "The feature which should be checked for selection." } ],
-            "examples": [ { "expression": "is_selected( 'streets', get_feature('streets', 'name', \"street_name\"))", "returns": "True if the current building's street is selected." } ]
+            "examples": [ { "expression": "is_selected( 'streets', get_feature('streets', 'name', \"street_name\"))", "returns": "True if the current building's street is selected (assuming the building layer has a field named 'street_name' and the 'streets' layer has a field called 'name')." } ]
         }
   ]
  }
--- a/resources/function_help/json/is_selected
+++ b/resources/function_help/json/is_selected
@@ -5,7 +5,7 @@
   "variants": [
         {
             "variant": "No parameters",
-            "variant_description": "If called with no parameters, it checks if current feature in current layer is selected.",
+            "variant_description": "If called with no parameters, the function will return true if the current feature in the current layer is selected.",
             "arguments": [ { "arg": "", "description": "No parameters" } ],
             "examples": [ { "expression": "is_selected()", "returns" : "True if the current feature in current layer is selected." } ]
         },
@@ -23,6 +23,5 @@
         }
   ]
  }
-
 
 
--- a/resources/function_help/json/is_selected
+++ b/resources/function_help/json/is_selected
@@ -1,23 +1,29 @@
 {
   "name": "is_selected",
   "type": "function",
-  "description": "Returns True if a feature is selected. If called with no parameters checks the current feature.",
-  "arguments": [
+  "description": "Returns True if a feature is selected. Can be used with zero, one or two parameters, see below for different variants.",
+  "variants": [
         {
-          "arg": "layer",
-          "optional": true,
-          "default": "current layer",
-          "description": "The layer (or its id or name) on which the selection will be checked."
+            "variant": "No parameters",
+            "variant_description": "If called with no parameters, it checks if current feature in current layer is selected.",
+            "arguments": [ { "arg": "", "description": "No parameters" } ],
+            "examples": [ { "expression": "is_selected()", "returns" : "True if the current feature in current layer is selected." } ]
         },
         {
-          "arg":"feature",
-          "optional": true,
-          "default": "current feature",
-          "description":"The feature which should be checked for selection."
+            "variant": "One 'feature' parameter",
+            "variant_description": "If called with a 'feature' parameter, it returns True if that feature is selected.",
+            "arguments": [ { "arg": "feature", "description": "The feature which should be checked for selection." } ],
+            "examples": [ { "expression": "is_selected( get_feature('streets', 'name', \"street_name\")) )", "returns": "True if that feature is selected." } ]
+        },
+        {
+            "variant" : "Two parameters",
+            "variant_description": "If called with a layer and a feature, it checks if that feature in that layer is selected.",
+            "arguments": [ { "arg": "layer", "description": "The layer (or its id or name) on which the selection will be checked." }, { "arg": "feature", "description": "The feature which should be checked for selection." } ],
+            "examples": [ { "expression": "is_selected( 'streets', get_feature('streets', 'name', \"street_name\"))", "returns": "True if the current building's street is selected." } ]
         }
-  ],
-  "examples": [
-        { "expression":"is_selected()", "returns" : "True if the current feature is selected."},
-        { "expression":"is_selected('streets', get_feature('streets', 'name', \"street_name\"))", "returns":"True if the current building's street is selected."}
   ]
-}
+ }
+
+
+
+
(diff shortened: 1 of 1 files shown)


nyalldawson:
Actually this function is complex, because it changes argument order depending on the number of arguments passed:

- no arguments: current feature, current layer
- one argument: current layer, only argument is the feature
- two arguments: first argument is the layer, second is the feature (order has swapped!!)

To handle this you'll need to use the "variant" approach in the expression function help. See the "length" function help for a template in how to do this. We should list explicitly all three variants (i.e. 0/1/2 arguments) separately with an explanation and examples of each.

rduivenvoorde:
@nyalldawson mmm, I was already afraid for this :-) It is never as easy as it looks ;-)

I will update the examples to reflect all four cases, AND I will try to add an example for the 'variant' approach...
You are looking at?

Screenshot-20200608082223-869x649

This 'length' example, looks to me more like function overloading? Because 'geom_from_wkt' does not return a variant is it? It returns an actual geometry?
OR is it an implementation detail, and the implementation actually receives a variant and has to check which type it actually is?


Just so I understand correct: the internal details are not relevant for the user-help is it?
If a user uses zero arguments, she does not care about the order it is used in the implementation?
If a user uses one argument (2 cases), she also 'just uses one'
And in the last example (2 arguments) you really have to use the layer-feature order?

nyalldawson:
variant === argument overloading. Have a look at the json for the length function -- it's named "variants" in the json, but it doesn't relate at all to "qvariants"!

nyalldawson:
> Just so I understand correct: the internal details are not relevant for the user-help is it?

I think they are in this case -- otherwise it's confusing.

> If a user uses zero arguments, she does not care about the order it is used in the implementation?

For the zero argument case, we don't list any arguments in this "variant" ... so the order won't apply!

> And in the last example (2 arguments) you really have to use the layer-feature order?

Yep (unless you use named arguments, but that's a different beast)

rduivenvoorde:
Ah sorry for my confusion, after rereading I already was thinking you were not talking about 'qvariant'-type... but you were to fast :-)
Will do now.

rduivenvoorde:
Ok, probably I'm not making things easier, as I thought there would be four options (reading from the original help tekst, that both feature and layer were optional)...
But I could not make `is_selected( @layer )` or  `is_selected( 'streets' )` work...
Which seems logical then, as we ask if a FEATURE is being selected, and a FEATURE is always in it's own LAYER?

So I came up with this:

Screenshot-20200608110356-737x850

But then I wondered if the third variant (with both layer and feature) is actually possible in other ways then testing if a feature is selected in it's own layer.
But probably the idea is to make both parameters dynamic, in that way open more possibilities?

Maybe we could add some extra styling in the dialog? As the distinction between different variants could maybe me more clear?
(same in the length example you point to):

Screenshot-20200608111412-709x631


Another(technical) observation: I kept the original 'arguments' part at first because I thought it would be nice to have a condensed view of the arguments in top. But IF you have that in combination with variants, it is ignored/not shown? Which is ok off course, it is a choice.

resources/function_help/json/is_selected
```diff
@@ -1,23 +1,29 @@
 {
   "name": "is_selected",
   "type": "function",
-  "description": "Returns if a feature is selected. If called with no parameters checks the current feature.",
-  "arguments": [
+  "description": "Returns True if a feature is selected. Can be used with zero, one or two parameters, see below for different variants.",
```
nyalldawson, line 4:
```suggestion
  "description": "Returns True if a feature is selected. Can be used with zero, one or two arguments, see below for details.",
```

resources/function_help/json/is_selected
```diff
@@ -1,23 +1,29 @@
...
+  "variants": [
         {
-          "arg":"feature",
-          "optional": true,
-          "default": "current feature",
-          "description":"The feature which should be checked for selection."
+            "variant": "No parameters",
+            "variant_description": "If called with no parameters, it checks if current feature in current layer is selected.",
```
nyalldawson, line 8:
```suggestion
            "variant_description": "If called with no parameters, the function will return true if the current feature in the current layer is selected.",
```

resources/function_help/json/is_selected
```diff
@@ -1,23 +1,29 @@
...
+            "arguments": [ { "arg": "", "description": "No parameters" } ],
```
nyalldawson, line 9:
```suggestion
            "arguments": [],
```

resources/function_help/json/is_selected
```diff
@@ -1,23 +1,29 @@
...
+            "examples": [ { "expression": "is_selected()", "returns" : "True if the current feature in current layer is selected." } ]
```
nyalldawson, line 10:
```suggestion
            "examples": [ { "expression": "is_selected()", "returns" : "True if the current feature in the current layer is selected." } ]
```

resources/function_help/json/is_selected
```diff
@@ -1,23 +1,29 @@
...
         },
         {
-          "arg": "layer",
-          "optional": true,
-          "default": "current layer",
-          "description": "The layer (or its id or name) on which the selection will be checked."
+            "variant": "One 'feature' parameter",
+            "variant_description": "If called with a 'feature' parameter, it returns True if that feature is selected.",
```
nyalldawson, line 14:
```suggestion
            "variant_description": "If called with a 'feature' parameter only, the function returns true if the specified feature from the current layer is selected.",
```

resources/function_help/json/is_selected
```diff
@@ -1,23 +1,29 @@
...
+            "arguments": [ { "arg": "feature", "description": "The feature which should be checked for selection." } ],
+            "examples": [ { "expression": "is_selected( get_feature('streets', 'name', \"street_name\")) )", "returns": "True if that feature is selected." } ]
```
nyalldawson, line 16:
```suggestion
            "examples": [ { "expression": "is_selected(@atlas_feature)", "returns": "True if the current atlas feature is selected." } ]
```

The use of get_feature here is confusing, and likely misleading (unless "streets" is the current layer!)

resources/function_help/json/is_selected
```diff
@@ -1,23 +1,29 @@
...
+        },
+        {
+            "variant" : "Two parameters",
+            "variant_description": "If called with a layer and a feature, it checks if that feature in that layer is selected.",
```
nyalldawson, line 20:
```suggestion
            "variant_description": "If the function is called with both a layer and a feature, it will return true if the specified feature from the specified layer is selected.",
```

resources/function_help/json/is_selected
```diff
@@ -1,23 +1,29 @@
...
+            "arguments": [ { "arg": "layer", "description": "The layer (or its id or name) on which the selection will be checked." }, { "arg": "feature", "description": "The feature which should be checked for selection." } ],
```
nyalldawson, line 21:
```suggestion
            "arguments": [ { "arg": "layer", "description": "The layer (or its ID or name) on which the selection will be checked." }, { "arg": "feature", "description": "The feature which should be checked for selection." } ],
```

resources/function_help/json/is_selected
```diff
@@ -1,23 +1,29 @@
...
+            "examples": [ { "expression": "is_selected( 'streets', get_feature('streets', 'name', \"street_name\"))", "returns": "True if the current building's street is selected." } ]
```
nyalldawson, line 22:
```suggestion
            "examples": [ { "expression": "is_selected( 'streets', get_feature('streets', 'name', \"street_name\"))", "returns": "True if the current building's street is selected (assuming the building layer has a field named 'street_name' and the 'streets' layer has a field called 'name')." } ]
```

nyalldawson:
I've just suggested some wording improvements which may help clarify the different use cases.

> Another(technical) observation: I kept the original 'arguments' part at first because I thought it would be nice to have a condensed view of the arguments in top. But IF you have that in combination with variants, it is ignored/not shown? Which is ok off course, it is a choice.

Right, the "variant" system is designed for formatting the help when the different uses of the function are sufficiently different to make this undesirable (as in this case).

> Maybe we could add some extra styling in the dialog? As the distinction between different variants could maybe me more clear?

Yeah - agreed.

rduivenvoorde:
@nyalldawson thanks for the textual stuff, that is not my strongest point. Hope this is still usefull.

I can try to fiddle with styles, just have to see where those are, and is there a way to peep into the html(?) or is it totally not css based...?

But we can also leave it as is for now.

rduivenvoorde:
His is only raising the H elements of the descriptions (from H2 -> H1) 
Note there is also a different type of green used here... 
I cannot find where this is defined (Qt?)

I will override the green and am thinking about the type in the first hear eigher make CamelCase of UPPERCASE...

Screenshot-20200609105644-652x903

nyalldawson (approved):

nyalldawson:
> This is only raising the H elements of the descriptions (from H2 -> H1)

I think we need to make sure the page heading is still the most prominent ("function is_selected"), but otherwise it's an improvement!

> Note there is also a different type of green used here...
I cannot find where this is defined (Qt?)

See https://github.com/qgis/QGIS/blob/master/src/core/qgsapplication.cpp#L1344

But let's make those changes in a separated PR...

PR #37025, opened 2020-06-08
Add metadata setting for file destination parameters to avoid file overwrite confirmation prompt

### LINKED ISSUE #37011: Expose QFileDialog's DontConfirmOverwrite option in QgsProcessingParameterFileDestination
I have a PyQGIS Processing script which opens a CSV file in append mode. When a user runs my script, if they select an existing CSV file as the output, they will get a warning about overwriting the file, but the file will not actually be overwritten.

`QFileDialog` has a [`DontConfirmOverwrite`](https://doc.qt.io/qt-5/qfiledialog.html#Option-enum) option which I could use to suppress this misleading warning. Could this option be exposed as a parameter for [`QgsProcessingParameterFileDestination`](https://qgis.org/pyqgis/master/core/QgsProcessingParameterFileDestination.html)?

nyalldawson:
Fixes #37011


alexbruy (approved):

dericke:
Wow, that was quick! Thank you!

PR #37027, opened 2020-06-08
[3d] When showing edges, data defined polygon height was not accounted for

nyalldawson:
I'm not 100% confident this is the correct fix. But without it, the combination of data defined polygon height + showing edges results in this:

image


nyalldawson:
@wonder-sk would appreciate your review of this!

wonder-sk (approved):
Looks good - thanks!

PR #37035, opened 2020-06-08
Unittest for wfs with SRSname

ismailsunni:
## Description

This PR is addressing this discussion: https://github.com/qgis/QGIS/pull/36926#discussion_r435711557

There are two unit test added:
1. WFS request with the same SRSname as in the project (i.e. 4326)
    This unit test is used to make sure that the request is working fine.
2. WFS request with the different SRSname (3857) to 4326 project.
    This request doesn't return any feature, although the coordinates are converted/projected from the 4326's coordinates. The expected/reference files can be wrong since I am not sure what's is the correct format would be. I only know that it shouldn't be empty.

Note: I removed the previous unit test that uses 4326's coordinates but SRS 3857. I use it to show that the SRSname is not considered in the request and it still returns features although the coordinates do not make sense in 3857.

cc @elpaso 

I will add the expected error after Travis finished checking (to see the failed test first.)

### DIFF
--- a/tests/src/python/test_qgsserver_wfs.py
+++ b/tests/src/python/test_qgsserver_wfs.py
@@ -285,6 +285,34 @@ class TestQgsServerWFS(QgsServerTestBase):
 """
         tests.append(('within4326FilterTemplate_post', within4326FilterTemplate.format("")))
 
+        # Check get feature within linear ring with srsName=EPSG:3857 (different from the project/layer)
+        # The coordinates are converted from the one in 4326
+        within3857FilterTemplate = """<?xml version="1.0" encoding="UTF-8"?>
+<wfs:GetFeature service="WFS" version="1.0.0" {} xmlns:wfs="http://www.opengis.net/wfs" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/wfs http://schemas.opengis.net/wfs/1.1.0/wfs.xsd">
+  <wfs:Query typeName="testlayer" srsName="EPSG:3857" xmlns:feature="http://www.qgis.org/gml">
+    <ogc:Filter xmlns:ogc="http://www.opengis.net/ogc">
+      <Within>
+        <PropertyName>geometry</PropertyName>
+        <Polygon xmlns="http://www.opengis.net/gml" srsName="EPSG:3857">
+          <exterior>
+            <LinearRing srsName="EPSG:3857">
+              <posList srsDimension="2">
+                890555.93 5465442.18
+                1001875.42, 5465442.18
+                1001875.42, 5621521.49
+                890555.93, 5621521.49
+                890555.93 5465442.18
+              </posList>
+            </LinearRing>
+          </exterior>
+        </Polygon>
+      </Within>
+    </ogc:Filter>
+  </wfs:Query>
+</wfs:GetFeature>
+"""
+        tests.append(('within3857FilterTemplate_post', within3857FilterTemplate.format("")))
+
         srsTwoLayersTemplate = """<?xml version="1.0" encoding="UTF-8"?>
 <wfs:GetFeature service="WFS" version="1.0.0" {} xmlns:wfs="http://www.opengis.net/wfs" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/wfs http://schemas.opengis.net/wfs/1.1.0/wfs.xsd">
   <wfs:Query typeName="testlayer" srsName="EPSG:3857" xmlns:feature="http://www.qgis.org/gml">
--- a/tests/src/python/test_qgsserver_wfs.py
+++ b/tests/src/python/test_qgsserver_wfs.py
@@ -268,7 +268,7 @@ class TestQgsServerWFS(QgsServerTestBase):
         <PropertyName>geometry</PropertyName>
         <Polygon xmlns="http://www.opengis.net/gml" srsName="EPSG:4326">
           <exterior>
-            <LinearRing srsName="EPSG:4326">
+            <LinearRing>
               <posList srsDimension="2">
                 8 44
                 9 44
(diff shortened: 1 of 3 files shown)


tests/src/python/test_qgsserver_wfs.py
```diff
@@ -257,6 +257,62 @@ def test_getfeature_post(self):
...
+  <wfs:Query typeName="testlayer" srsName="EPSG:3857" xmlns:feature="http://www.qgis.org/gml">
+    <ogc:Filter xmlns:ogc="http://www.opengis.net/ogc">
+      <Within>
+        <PropertyName>geometry</PropertyName>
+        <Polygon xmlns="http://www.opengis.net/gml" srsName="EPSG:3857">
+          <exterior>
+            <LinearRing srsName="EPSG:3857">
+              <posList srsDimension="2">
+                890555.93 5465442.18
+                1001875.42, 5465442.18
```
elpaso, line 301: This does not look correct to me: remove commas.

elpaso (changes requested):

tests/src/python/test_qgsserver_wfs.py
```diff
@@ -285,6 +285,34 @@ def test_getfeature_post(self):
...
+        # The coordinates are converted from the one in 4326
+        within3857FilterTemplate = """<?xml version="1.0" encoding="UTF-8"?>
+<wfs:GetFeature service="WFS" version="1.0.0" {} xmlns:wfs="http://www.opengis.net/wfs" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/wfs http://schemas.opengis.net/wfs/1.1.0/wfs.xsd">
+  <wfs:Query typeName="testlayer" srsName="EPSG:3857" xmlns:feature="http://www.qgis.org/gml">
+    <ogc:Filter xmlns:ogc="http://www.opengis.net/ogc">
+      <Within>
+        <PropertyName>geometry</PropertyName>
+        <Polygon xmlns="http://www.opengis.net/gml" srsName="EPSG:3857">
+          <exterior>
+            <LinearRing srsName="EPSG:3857">
```
elpaso, line 299: can you remove srsName here (and on line https://github.com/qgis/QGIS/pull/37035/commits/9370d94addf06f45d469733a7458222ead25309d#diff-61a1cac6f0799a1d603fac5b80d0f5ddR268 ) ? 
It's the same as the polygon anyway (and it's not supported).

tests/src/python/test_qgsserver_wfs.py
```diff
@@ -285,6 +285,34 @@ def test_getfeature_post(self):
 """
         tests.append(('within4326FilterTemplate_post', within4326FilterTemplate.format("")))
 
+        # Check get feature within linear ring with srsName=EPSG:3857 (different from the project/layer)
```
elpaso, line 289:
```suggestion
        # Check get feature within polygon having srsName=EPSG:3857 (different from the project/layer)
```

tests/testdata/qgis_server/wfs_getfeature_within3857FilterTemplate_post.txt
```diff
@@ -0,0 +1,60 @@
+Content-Type: text/xml; subtype=gml/2.1.2; charset=utf-8
```
elpaso, line 1: Before I dig into this again, would you please confirm that this is the expected output for the case in which you are asking for  3857 features and a 3857 within filter? 

Because it seems to me that you are expecting `8.20349634,44.90148253 8.20349634,44.90148253` 4326 coordinates, should it not be in 3857?

ismailsunni: Hi @elpaso 
No, this file is not the expected output. 

I only know that the output is wrong since there is no feature returned although the polygon is the same as in 4326. Since I don't know the exact expected output, I just put the same file as in the test 4326 request. 

Or is there a way to get the correct output besides doing the request?

elpaso: I would keep things as simple as possible: you can forge a request around a single feature, just make a tiny buffer around the point and take the BBOX.
Then you will expect just that feature back with its coordinates (I'm sure you know how to convert them from 4326 to 3857). You can probably take the 4326 and change the coords and srs.

ismailsunni: I have updated the test, but I am still not sure if the converted polygon falls correctly on the expected point, I will check again tomorrow with a more clear head.

elpaso: Thank you @ismailsunni and sorry if I wasn't clear. I see that this PR still contains a set of test cases that compare output calls to verify that they are **not** equal to a reference file.

What I am asking you to provide is a singe failing test case (one: not many) that given a request (if I understand what you are trying to demonstrate, a request with a GML 3857 encoded within filter and 3857 features result) produces an expected reference output. 

In other words a "positive" (but failing) test with an expected output (it does not matter if it's forged by hand).

Is that clear?

If you want you can leave the other (just one) positive test that already passes (the one with 4326) for reference, but that is not really required.

ismailsunni: Hi @elpaso

> I see that this PR still contains a set of test cases that compare output calls to verify that they are not equal to a reference file.

Which one is it? I only add 2 unit tests in this PR, first for 4326 references, and the other one for 3857 with output result created by hand as you suggested yesterday.

I understand what you mean actually, that's why I removed the third unit test from the reverted PR on this PR.

Now, I think I have good unit test, yesterday I was doing silly thing and messed up the polygon.

As a reference, this is the polygon (as 4 blue points) with the coordinates on 4326 (red) and 3857 (black).
image

> If you want you can leave the other (just one) positive test that already passes (the one with 4326) for reference, but that is not really required.

If it's ok, I want to keep it, for reference 

Thanks @elpaso! Let me know if you need anything else.

elpaso:
Thank you. I've added your test with the patch that fixes the issue: https://github.com/qgis/QGIS/pull/37071

Issue #36904, opened 2020-06-02
Layout: tickmarks and labels in map frame should be rotated


andreasneumann:
**Bug description.**
In a layout, if a frame is activated and uses tickmarks (interior, exterior or both) and the map is rotated, the tick marks and the "draw coordinate" labels should be rotated in sync with the rotation of the coordinate crosses or grid lines.

![grafik](https://user-images.githubusercontent.com/884476/83517934-0ca22480-a4da-11ea-9965-5ad9bf50ed53.png)

Not sure if this can be regarded as a bug or feature request.

Questions: are there cases where the tick marks shouldn't be rotated? If yes, then we'll need an option, but the rotation should be enabled by default.

For the "draw coordinate" labels there should be an option to rotate them or not.

olivierdalang:
@andreasneumann 
I started working on this.

A few questions/comments :

1. IMO it's better to have an option to not rotate ticks, as in some cases it doesn't look good (e.g.  with solid grid lines)

2. There's two ways to draw rotated ticks, either using an "orthogonal size" (e.g. ticks get longer when with a low angle to the frame, to respect orthogonal alignment) or using a "normalized size" (all ticks have the same length, so they don't align orthogonaly). Code-wise it's trivial, but I'm not sure if this deserves yet another option for already complex map grid dialog or if we can just pick the one that we think is best (in which case : which one do we think is best :-) ?).

3. Currently, it's possible to choose "latitude only" or "longitude only" for each border. However, with rotated maps, it would be even better to be able to hide ticks below a certain angle. For instance, displaying ticks for grid lines that intersects the frame with a very low angle (almost parallel to the frame) isn't very helpful nor beautiful. By default, I'd hide ticks at < 30°. This is not really part of this issue, but would improve rotated maps.


### screenshots

"normalized tick length"
![image](https://user-images.githubusercontent.com/1894106/83553265-417a9f80-a50b-11ea-8e66-7bbb51938380.png)

"orthogonal tick length"
![image](https://user-images.githubusercontent.com/1894106/83553376-65d67c00-a50b-11ea-83e7-b93300080829.png)

andreasneumann:
Good question. In general I like the second variation (orthogonal tick length) because all tickmarks end at one line, but in case of very flat (small) angles), the very long lines may also look a bit ugly. But as I understand your comment 3, such very long ticks at flat angles could be suppressed. That would look prettier, I agree.

[...]

PR #37041, opened 2020-06-08
Support for rotated ticks/annotation

olivierdalang:
## Description

This PR implements support for rotating ticks/annotations in the print composer as described here https://github.com/qgis/QGIS/issues/36904. It works for both rotated maps and reprojected grids.

pr_rotated_ticks

For ticks, it adds following options :
- rotate ticks (if unchecked, ticks get drawn as before, not taking angle into account)
- rotated ticks alignement (whether ticks should have a constant length, or be aligned orthogonaly)
- skip low angled ticks (an angle threshold below which ticks aren't drawn, so you can skip ticks for grid lines that are too parralel the the map edge which don't look good and are useless)

For annotations, it adds following options :
- Parallel to Tick (next to Horizontal / Vertical Ascending / Vertical Descending)
- draw low angled (if checked, skipped ticks above will still get an annotation)

I don't like the duplication of settings per border and between annotation and ticks too much. IMO, most of the time, you want the same settings for annotations and ticks, and on all borders. The only setting that's really useful per border is the enabled/disabled per annotation/tick, so that for complex cases, you can just overlay multiple grids (this workflow could be improved with a simple duplicate grid button). This would also allow for a common case where you want many ticks, but coordinates only on major ticks.

Changing this wouldn't be backwards compatible, so it's maybe something to keep in mind for 4.0 ? Or is it a small detail enough so that we would allow breakage ?

Also in terms of code, maybe we could use the same logic I added for rotated ticks (using a QLineF representation) and compute positions using vector arithmetic in the same way for all cases, instead of having dozens of conditions with duplicated code (https://github.com/qgis/QGIS/blob/master/src/core/layout/qgslayoutitemmapgrid.cpp#L1108-L1399). I didn't go that way to avoid changing the existing code too much, but I think it would be necessary if we want more control (e.g. annotation above/over/under line).

Also it wasn't decided on the issue whether this is considered a bug fix or a new feature (I'd be more comfortable calling this a new feature as there's a risk of breaking things :-) ).

@gislars @andreasneumann @edigonzales
There's really a lot of different cases, so it would be nice if you could help test it out before we merge this :-)

Also did manage to run the python tests locally yet, so I'll have to add tests in another pass.

Cheers,

Olivier


andreasneumann:
Looks nice. Good work! Will have to test it.

@olivierdalang - I think there is a "typo" in "Parralel to Tick": "Parralel" --> should be "parallel".

While we are at it:
Would it make sense to change the label "Frame size" to "Frame size / tick length" ? Or depending on the choice of frame vs ticks, adapt the label accordingly? Same for everything "frame" vs "tick" in all the other labels ...

> Also it wasn't decided on the issue whether this is considered a bug fix or a new feature (I'd be more comfortable calling this a new feature as there's a risk of breaking things :-) ).

Let's regard it as a new feature without backports. Looking at it now, there are new options and everything. It can hardly be justified as a bug fix.

nyalldawson:
I think there's a bug shown in your screencast:

when you change the labels on the right to parallel mode, this shouldn't change the horizontal placement of them (i.e. you shouldn't have to change the frame size in order to avoid the overlap)

andreasneumann:
> when you change the labels on the right to parallel mode, this shouldn't change the horizontal placement of them (i.e. you shouldn't have to change the frame size in order to avoid the overlap)

Yes, I would agree with @nyalldawson

gislars:
Thank you @olivierdalang for your work. It took me a little to give feedback, because I had to first build QGis for the first time. Following the instruction it was straight forward and finally I got QGis running.

I focused on the label options and not the ticks, these are my findings
* the labels are centered on the ticks
* I couldn't place the label completely inside or outside the map frame with the options given in "Draw coordinates"
* the "distance to map frame" option doesn't seem to have an effect
* when enabling a frame style (interior ticks) I could move the label using frame size and margin, that might not be intuitive to the user, but works for me
* labels can overlap each other and they are hard to read 
* since overlapping happens on the corners, an offset for the axes could help. e.g. 20mm from each corner the labeling starts
* placing a label above or below a grid line 
* increasing the frame size, the tick lines increase and reach out of the map frame (1:05 in screen cast) at the corner
* setting frame size to 0 the labels are switching to horizontal placement (1:40)
* I could find out what "Draw low angled annotations" is for. Is it linked to the angle in "Skip low angled ticks"? Maybe angles didn't match to see an effect.

Notes: 
* I had to move the canvas to update the screen, but that might be something completely unrelated
* I recorded a mp4 but had to convert it to gif, which is not good quality. If it helps I could provide the mp4 somehow, let me know
* In the screencast there there are other grid configs which show a light gray grid. I added a new grid which only has the label

So far I'm happy, I think it is a valuable feature. Please let me know if you need more feedback or what else I should test. 

qgis_rotatet_tick_and_labels

olivierdalang:
Progress on this :
- [x] adapted Nyall's commit to use QgsTextFormat for annotation (which makes it possible to add buffers, shadows, etc. to annotations)
- [x] refactored the annotation's placement code (using vector arithmetic instead - this is now used for all type of annotation placement instead of the dozens of nested conditions)
- [x] added a new test for label placement variations (for non-rotated grids, to ensure this PR doesn't change existing rendering)
- [x] added a new test for rotated labels variations
- [ ] probably need to fiddle a bit until test pass (I have failures due to differences of font rendering)
- [ ] double check the API changes (deprecation of annotationFont/annotationFontColor & new annotationTextFormat getter and setter)
- [ ] (maybe) add an option to skip labels near corners to help with overshooting ticks
- [ ] cleanup commit history (squash followup commits and credit Nyall for his commit)

@gislars thanks for the detailed testing ! Unfortunately I only see the first few seconds of the gif ? But as I completely rewrote the placement logic, most of the issues you were mentioning should be fixed. Also I like your idea of skipping annotations near corners, it would definitely help with rotated grids.

pr_rotated_ticks_2

olivierdalang:
A question about the API :

Before this PR, we had following setters/getters :
    - `setAnnotationFont(QFont font)`
    - `QFont annotationFont()`
    - `setAnnotationFontColor(QColor color)`
    - `QColor annotationFontColor()`

After this PR, now that we use a QgsTextFormat, we'll have : 
    - `setAnnotationTextFormat(QgsTextFormat format)`
    - `QgsTextFormat* annotationTextFormat()` 

Do we want to keep the old methods, or deprecate them for removal in 4.0 ? The new ones are not exactly equivalent (for the font, one needs to define the font size units, and for the color, QgsTextFormat doesn't use QColor's alpha channel for opacity but a separate setOpacity())... Currently, I left the old getter/setters (non-deprecated) as shortcuts, as they are less verbose and nicer to use.

```python
# before
grid.setColor(QColor(255,0,0,127)
# after (if we deprecate)
grid.annotationTextFormat().setColor(QColor(255,0,0))
grid.annotationTextFormat().setOpacity(0.5)
```

gislars:
Today I pulled in the latest changes on this and did some more tests. 
I really like the results!

Focusing on the labels, I'll attach a PDF which is the result of my tests. I included a screenshot of my settings. 
I used different map rotations to see how the labels work. So far so good. For some rotation values, label reach out of the map frame or overlap each other. I played around with the "skip low angled" settings, but in some cases it simply can't help.

I'll attach my settings as screenshot and the PDF of my test case. 

Bildschirmfoto vom 2020-07-03 19-19-37
[grid_label_test.pdf](https://github.com/qgis/QGIS/files/4870818/grid_label_test.pdf)

nyalldawson:
@olivierdalang can i suggest deferring the move to QgsTextRenderer to a follow up PR? I've hit a number of issues which result in inferior text placement when using QgsTextRenderer in layouts, which is why the original commit was never merged and why https://github.com/qgis/QGIS/pull/35992 was never merged.

I'm hoping https://github.com/qgis/QGIS/pull/37635 may help here -- but in the meantime I suggest reverting the inclusion of my text renderer commit in order to avoid this becoming a road block to your other work.

nyalldawson:
@olivierdalang one last question -- how hard do you think it would be to add an api method to return the rects used for the labels as a multipolygon geometry? If we had this, then we could easily inject it into the labeling run for the map and avoid placing map labels under grid annotation labels...

olivierdalang:
> @olivierdalang can i suggest deferring the move to QgsTextRenderer to a follow up PR? I've hit a number of issues which result in inferior text placement when using QgsTextRenderer in layouts, which is why the original commit was never merged and why #35992 was never merged.
> 
> I'm hoping #37635 may help here -- but in the meantime I suggest reverting the inclusion of my text renderer commit in order to avoid this becoming a road block to your other work.

@nyalldawson I haven't noticed issues with text placement ? The rendering is slightly different (+/- 1 pixel) with VS without the QgsTextRenderer, but nothing humanly noticeable I think. Do you have some specific cases in mind ? Did you check @gislars extensive testing (in the PDF linked above) and the changes in expected images from the test suite ?

(of course tests will probably fail again if #37635 is merged first (or the other way around), but at this stage, one more round of failing tests won't scare me :-) )


> @olivierdalang one last question -- how hard do you think it would be to add an api method to return the rects used for the labels as a multipolygon geometry? If we had this, then we could easily inject it into the labeling run for the map and avoid placing map labels under grid annotation labels...

That should be relatively easy, the transform here https://github.com/qgis/QGIS/pull/37041/files#diff-16e4fb7840b3ec7b9c0c375f50e34524R1303-R1304 can be used to get the points of the rectangles.

nyalldawson:
> I haven't noticed issues with text placement ? The rendering is slightly different (+/- 1 pixel) with VS without the QgsTextRenderer, but nothing humanly noticeable I think. Do you have some specific cases in mind ? Did you check @gislars extensive testing (in the PDF linked above) and the changes in expected images from the test suite ?

I just double checked, and there's definitely issues :smile: Some are resolved with the scaling workaround PR, but there's some more metric based calculations that need fixing. 

I've worked out what's required here, so I'll open a separate PR with my changes. (It's also helped me work out the issue holding back the legend text renderer branch!!)

nyalldawson:
@m-kuhn i suspect there's been a bad merge in the ui file here -- there's some duplicate labels now:

image

(also a typo in alignment)

and a missing label here:

image

But more seriously, somehow there's two combo boxes with a duplicate name present in the ui now...

nyalldawson:
(mAnnotationPositionLeftComboBox)

PR #37049, opened 2020-06-09
[geopdf] Allow users to reorder layers in the generated layer tree

### LINKED ISSUE #36535: GeoPDF export: Control Layer order
**Ability to control the order of layers as they appear in Acrobat's _Layers Navigation Pane_.**
Currently, the order of layers in Acrobat's navigation pane appears to be random--at least I have not been able to figure this out. This means that to make these PDFs understandable to my clients I need to go into Acrobat and reorder the layers after the export.

nyalldawson:
Fixes #36535


nirvn (approved):

3nids:
it seems this layer tree would be a very good candidate for QgsLayerTreeFilterProxyModel :)

nyalldawson:
> QgsLayerTreeFilterProxyModel

I originally used a layer tree model subclass for this, but then dropped it back to the layer model when the reordering was needed. I couldn't find a nice way to handle both layer tree groups + arbitrary rearrangement of layers which may or may not be in groups!

PR #37051, opened 2020-06-09
Some fixes for labels missing from layers even when "show all" labels is checked

nyalldawson:
One unhandled situation dating from forever, and one regression caused by the newer placement/obstacle logic


nirvn (approved):
Nice, that comes in at a perfect time over here, about to print a label-heavy A0 topo map and was negatively impacted by the obstacle regression.

PR #37053, opened 2020-06-09
Refine refreshConnections method to avoid triggering a full refresh of ALL browser content

### LINKED ISSUE #37007: QGIS Browser: Oracle connection refreshs when adding/removing gpkg connection
Description:
With an active connection to an Oracle database in the QGIS Browser I make add or remove a geopackage connection (right-click gpkg add/reomove connection). This action triggers a refresh of the table of contents for the oracle connection (Our Oracle db takes about 190 seconds to scan :| ). From what I can tell this doesn't effect a PostGIS connection but the refresh is so much faster than Oracle its hard to see.

Steps to reproduce
1.  Start QGIS
2. Connect to an Oracle Database in QGIS Browser
3. After the catalog is refreshed navigate to a geopackage
4. Right-click the gpkg and click 'add connection'

I have checked that this bug exists in Windows versions of 3.10.5 and 3.8.3 on different machines. Version info below:

`
QGIS version
3.10.5-A Coruña
QGIS code revision
984615fe1e
Compiled against Qt
5.11.2
Running against Qt
5.11.2
Compiled against GDAL/OGR
3.0.4
Running against GDAL/OGR
3.0.4
Compiled against GEOS
3.8.1-CAPI-1.13.3
Running against GEOS
3.8.1-CAPI-1.13.3
Compiled against SQLite
3.29.0
Running against SQLite
3.29.0
PostgreSQL Client Version
11.5
SpatiaLite Version
4.3.0
QWT Version
6.1.3
QScintilla2 Version
2.10.8
Compiled against PROJ
6.3.1
Running against PROJ
OS Version
Windows Server 2016 (10.0)
Active python plugins
debug_vs; 
processing_r; 
QuickOSM; 
db_manager; 
processing
`

nyalldawson:
Instead limit refresh to the provider associated with the item only (and provide a means for items to refresh a different provider, e.g. to allow the geopackage connection item to be refreshed when a
new connection is added through a directory item)

Fixes #37007

(I don't see this as suitable for a backport)

### DIFF
--- a/src/core/qgsdataitem.h
+++ b/src/core/qgsdataitem.h
@@ -443,8 +443,13 @@ class CORE_EXPORT QgsDataItem : public QObject
 
     virtual void refresh();
 
-    //! Refresh connections: update GUI and emit signal
-    virtual void refreshConnections();
+    /**
+     * Causes a data item provider to refresh all registered connections.
+     *
+     * If \a provider is specified then only the matching provider will be refreshed. Otherwise,
+     * all providers will be refreshed (which is potentially very expensive!).
+     */
+    virtual void refreshConnections( const QString &providerKey = QString() );
 
     virtual void childrenCreated();
 
@@ -457,12 +462,13 @@ class CORE_EXPORT QgsDataItem : public QObject
     void stateChanged( QgsDataItem *item, QgsDataItem::State oldState );
 
     /**
-     * Emitted when the provider's connections of the child items have changed
+     * Emitted when the connections of the provider with the specified key have changed.
+     *
      * This signal is normally forwarded to the app in order to refresh the connection
      * item in the provider dialogs and to refresh the connection items in the other
-     * open browsers
+     * open browsers.
      */
-    void connectionsChanged();
+    void connectionsChanged( const QString &providerKey = QString() );
 
   protected slots:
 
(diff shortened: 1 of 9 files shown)


src/core/qgsdataitem.h
```diff
@@ -443,8 +443,13 @@ class CORE_EXPORT QgsDataItem : public QObject
 
     virtual void refresh();
 
-    //! Refresh connections: update GUI and emit signal
-    virtual void refreshConnections();
+    /**
+     * Causes a data item provider to refresh all registered connections.
+     *
+     * If \a provider is specified then only the matching provider will be refreshed. Otherwise,
```
elpaso, line 449:
```suggestion
     * If \a providerKey is specified then only the matching provider will be refreshed. Otherwise,
```

src/core/qgsdataitem.h
```diff
@@ -457,12 +462,13 @@ class CORE_EXPORT QgsDataItem : public QObject
     void stateChanged( QgsDataItem *item, QgsDataItem::State oldState );
 
     /**
-     * Emitted when the provider's connections of the child items have changed
+     * Emitted when the connections of the provider with the specified key have changed.
```
elpaso, line 465:
```suggestion
     * Emitted when the connections of the provider with the specified \a providerKey have changed.
```

elpaso (approved):

PR #37056, opened 2020-06-09
Add missing country attributes in `world_map.gpkg`

ghtmtt:
## Description

I noticed that in the `world_map.gpkg` file there where some attributes missing (France, Norway). Just added that in the attribute table taking the values from Natural Earth. 

Code hasn't been touched.


m-kuhn:
@arongergely can I ask you to check this PR? Thank you.

nyalldawson:
I've compared (using the "Detect Dataset Changes" algorithm), and it looks good to me!

arongergely:
Ah, already merged. Thanks @ghtmtt @m-kuhn, I just checked the added values anyway, they are correct.

PR #37058, opened 2020-06-09
fix missing table name

3nids:
this was causing a crash spotted here: https://github.com/qgis/QGIS/commit/07eca3e6702f16785f64686dcbf568de1564eb4f#r39692642
also discussed here: #36910

The table name was actually missing.


m-kuhn (approved):
Thanks !

pblottiere:
> The table name was actually missing.

Thanks @3nids!

PR #37067, opened 2020-06-09
[processing] Make gdaladdo optional parameters truly optional

nirvn:
## Description

This PR improves the gdal gdaladdo algorithm by making the overview levels optional (made optional in gdal since 2.3) as well as making the resampling method truly optional (i.e., won't show up in the cli string outputted by the algorithm).


m-kuhn (approved):

nyalldawson:
> made optional in gdal since 2.3

We still (unfortunately) support GDAL 2.2. You'll need to put the new behavior inside a GDAL version check

nyalldawson (changes requested):
needs version check

nirvn:
@nyalldawson , version check added.

python/plugins/processing/algs/gdal/gdaladdo.py
```diff
@@ -125,8 +132,9 @@ def getConsoleCommands(self, parameters, context, feedback, executing=True):
         arguments = []
         arguments.append(fileName)
 
-        arguments.append('-r')
-        arguments.append(self.methods[self.parameterAsEnum(parameters, self.RESAMPLING, context)][1])
+        if self.RESAMPLING in parameters and parameters[self.RESAMPLING] is not None:
+            arguments.append('-r')
```
nyalldawson, line 136: Does this also need a version check?

nirvn: Not AFAIK, document doesn't link this parameter's default value to a specific gdal version. In fact, even the algorithm itself meant for it to be optional, it was just not reflected in the resulting command line.

nirvn: @nyalldawson , here, gdal 2.2's gdaladdo does indeed have a default value and doesn't need the presence of -r:

https://github.com/OSGeo/gdal/blob/release/2.2/gdal/apps/gdaladdo.cpp#L111

PR #37068, opened 2020-06-09
WEBP for QGIS Server

m-kuhn:
## Description

[Because](https://www.opengis.ch/2020/06/09/offline-wms-benchmarking-raster-formats-for-qfield/) [why not](https://developers.google.com/speed/webp/).

### DIFF
--- a/src/server/services/wms/qgswmsutils.cpp
+++ b/src/server/services/wms/qgswmsutils.cpp
@@ -79,6 +79,11 @@ namespace QgsWms
     {
       return JPEG;
     }
+    else if ( format.compare( QLatin1String( "webp" ), Qt::CaseInsensitive ) == 0  ||
+              format.compare( QLatin1String( "image/webp" ), Qt::CaseInsensitive ) == 0 )
+    {
+      return WEBP;
+    }
     else
     {
       // lookup for png with mode
@@ -146,6 +151,9 @@ namespace QgsWms
         contentType = "image/jpeg";
         saveFormat = "JPEG";
         break;
+      case WEBP:
+        contentType = QStringLiteral( "image/webp" );
+        saveFormat = QStringLiteral( "WEBP" );
       default:
         QgsMessageLog::logMessage( QString( "Unsupported format string %1" ).arg( formatStr ) );
         saveFormat = UNKN;
@@ -159,7 +167,7 @@ namespace QgsWms
     if ( outputFormat != UNKN )
     {
       response.setHeader( "Content-Type", contentType );
-      if ( saveFormat == "JPEG" )
+      if ( saveFormat == QLatin1String( "JPEG" ) || saveFormat == QLatin1String( "WEBP" ) )
       {
         result.save( response.io(), qPrintable( saveFormat ), imageQuality );
       }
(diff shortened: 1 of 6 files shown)


andreasneumann:
Nice - how about the opposite? Consuming WebP as a WMS client in QGIS desktop?

Would that be equally easy to implement like enabling it in QGIS server?

m-kuhn:
> Consuming WebP as a WMS client in QGIS desktop?

Good question. Likely yes. But I guess answering the question is best done through an implementation (which is not on my radar today)

elpaso (approved):
Nice!

rldhont (approved):

src/server/services/wms/qgswmsutils.cpp
```diff
@@ -74,11 +74,16 @@ namespace QgsWms
     {
       return PNG;
     }
-    else if ( format.compare( QLatin1String( "jpg " ), Qt::CaseInsensitive ) == 0  ||
+    else if ( format.compare( QLatin1String( "jpg" ), Qt::CaseInsensitive ) == 0  ||
```
m-kuhn, line 77: @elpaso @rldhont do you know what's this for? I wonder if that's actually ever used. It wasn't functional before (?) and nobody complained.

pblottiere: I think that most people use `image/jpeg` (as well as unit tests) so it probably slipped under the radar.

elpaso: I think it's been there for ages, It's not a standard mime type and untested. I believe the idea was just to be tolerant in what we accept.

rldhont:
@m-kuhn thanks, just said just done, you have been faster then me.

andreasneumann:
> which is not on my radar today

fair enough ;-)

sbrunner (approved):

src/server/services/wms/qgswmsutils.cpp
m-kuhn: What's your opinion, fix it, leave it or ditch it?
I think `image/jpg` or `jpeg` would be just as likely to be used, rendering this nothing but a random string.

rldhont: fix it, isn't it ?

m-kuhn: The code is a) broken b) non standard c) almost random and d) very likely unused. Four reasons for removing. I find it hard to find a good one for keeping.

m-kuhn: What does "fixing" mean though? Simply removing the `' '` at the end of the string? Or harmonizing with the content type and changing `'jpg'` to `'jpeg'`? Or adding both? Isn't any of that breaking the interface, someone could already be sending requests with `jpg ` right now?

If it was really "broken code" I'd be all in for a fix. But I'm having a hard time seeing this as broken code. It looks more like a desperate attempt, pretending to be tolerant (as confirmed by @elpaso above).

elpaso: @m-kuhn I think that in the old times it was more common for clients to use "jpg" "jpeg" with or without "image/", lower or upper case, I have no idea if that's still the situation today, I guess it's not.

So, if fixing does not cost too much I would leave that possibility, but if we do, we should document and put it under test. 

I don't really have a strong opinion here.

m-kuhn:
Shall we still ship that with 3.14?
It's a super small patch, very low risk but hard to label "bugfix".

elpaso:
Ship it!

pblottiere:
+1 to merge :+1:

m-kuhn:
:tada:

PR #37084, opened 2020-06-10
Use a proper ellipsoidal length calculation when calculating the size of the scalebar decoration

nyalldawson:
Avoids a misleading Cartesian-based scalebar (unless project is
set to always use cartesian measurements, that is!)

Fixes #28407

### DIFF
--- a/src/app/decorations/qgsdecorationscalebar.cpp
+++ b/src/app/decorations/qgsdecorationscalebar.cpp
@@ -206,6 +206,32 @@ void QgsDecorationScaleBar::setupScaleBar()
   }
   mSettings.setLabelBarSpace( 1.8 );
 }
+
+double QgsDecorationScaleBar::mapWidth( const QgsMapSettings &settings ) const
+{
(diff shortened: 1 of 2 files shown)


agiudiceandrea:
Hi @nyalldawson, is it possible this be the cause of the https://github.com/qgis/QGIS/issues/40748 issue (which occurs for QGIS ltr >= 3.10,8 and QGIS pr >= 3.14.0)?

nyalldawson:
@agiudiceandrea it's possible -- but prior to this the scale bar results were misleading anyway, so it would be tricky to test. That issue description makes it sound like there's a missing connection or signal causing the bar to update...

agiudiceandrea:
Not so misleading...

with 3.10.7
image

with 3.10.8
image

with 3.10.8 after zoom in and zoom last
image

agiudiceandrea:
https://github.com/qgis/qgis/compare/final-3_10_7...final-3_10_8#diff-e7cc4955553676e489c9dd13346c81957323993e398984b0d569365995b32be0

PR #37087, opened 2020-06-10
fix qgis INSTALL step

PeterPetrik:
fix https://github.com/qgis/QGIS/commit/dcd017a655f9583bb759c7d00cdc164de0a0dfa7#commitcomment-39794207


nirvn:
@PeterPetrik , thanks for the quick follow up.

PeterPetrik:
lets merge it right away?

nirvn:
@PeterPetrik , have you tried make install against those changes? If so, yeah, merge away, no real point in waiting for travis and azure both not testing make install.

PR #37105, opened 2020-06-10
Add safety guard for DD attribute alias

m-kuhn:
Fixes a crash if a widget is configured for an inexistent field.

Check diff [without whitespace](https://github.com/qgis/QGIS/pull/37105/files?diff=split&w=1) for easier review.


elpaso (approved):
LGTM thanks!

Issue #37020, opened 2020-06-07
Inconsistent styling with vector tiles


saberraz:
This could be related to https://github.com/qgis/QGIS/issues/36995

In the project below, I cannot see the countries boarder when zoomed out:
https://public.cloudmergin.com/projects/saber/natural-earth

![natural-earth](https://user-images.githubusercontent.com/2649047/83978085-ef67be80-a8fc-11ea-8584-29030320df0b.gif)

nyalldawson:
@saberraz do you have a public QGS project which demonstrates this?

Issue #36995, opened 2020-06-05
[vector tiles] inconsistent display of labels


saberraz:
I have created labels for `names`  field. You should be able to view labels for all the features in the Points layer (brown circles). But it only displays Scotland.

![label](https://user-images.githubusercontent.com/2649047/83860224-e9b08400-a716-11ea-93fe-0f0796aaa606.png)

When you zoom in and out, the choice of displaying label appears to be very random.
https://public.cloudmergin.com/projects/saber/zoomstack

PR #37112, opened 2020-06-10
[vector tile] Fix issue with disappearing features (fixes #36982)

### LINKED ISSUE #36982: [vector tiles] Adding label makes feature disappear
When I add the same rule for symbology to label, the feature disappears:
vectortile

[zoomstack.qgs.zip](https://github.com/qgis/QGIS/files/4732425/zoomstack.qgs.zip)

wonder-sk:
The issue was that renderer/labeling were only keeping "their" required fields, but the final fields may have been extended by the other, making the field indices cached in expressions invalid.

The fix is to keep the final QgsFields around and avoid creating those QgsFields on-the-fly from previously requested fields.

### DIFF
--- a/python/core/auto_generated/vectortile/qgsvectortilerenderer.sip.in
+++ b/python/core/auto_generated/vectortile/qgsvectortilerenderer.sip.in
@@ -44,6 +44,15 @@ Sets polygon of the tile
 Returns polygon (made out of four corners of the tile) in screen coordinates calculated from render context
 %End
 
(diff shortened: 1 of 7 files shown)


saberraz:
Fixes https://github.com/qgis/QGIS/issues/36982 and https://github.com/qgis/QGIS/issues/36995 and https://github.com/qgis/QGIS/issues/37020 ?

wonder-sk:
#36995 is a separate issue
#37020 ... maybe, not sure, would need to investigate

PR #37113, opened 2020-06-10
Fix list display in dox

DelazJ:
should fix the display of the end of this list 

image

I have no idea how list multiline item can be handled so make it a single long line.


nyalldawson:
Superseded by https://github.com/qgis/QGIS/pull/37118

PR #37115, opened 2020-06-10
Tests to check correctness of WHERE clauses for PostgreSQL

espinafre:
When I submitted the improvement for handling PostgreSQL bigint/bigserial primary keys, preventing QGIS from casting those to text (PR #35162) I wrote some functional testcases in Python. Those tests only assured that my changes didn't break anything, but they didn't prove that the changes really worked as intended. I've checked that by looking at debug logs and the database logs. I didn't know then a way to formally check the generated SQL queries from inside the testing framework.

Now, while reviewing #36266, I saw that @benoitdm-oslandia came up with the idea of calling the QgsPostgresUtils::whereClause() method directly, while using an instance of QgsPostgresSharedData to pass the fields to the provider.

So, I stole his idea and wrote these tests, and I've gained some more insights about the QGIS provider in the process. For instance, I've found that in the case of compound keys, the original problem of QGIS casting integers to text persist; that's not a common use case, but still, I think it is worth a shot (thats what the FIXME comment on the last test is for).

Also, with these tests we cases we may rest more assured that there won't be regressions in this case.

### DIFF
--- a/tests/src/providers/testqgspostgresprovider.cpp
+++ b/tests/src/providers/testqgspostgresprovider.cpp
@@ -199,7 +199,8 @@ void TestQgsPostgresProvider::testQuotedValueBigInt()
   QgsFields fields;
   QList<int> pkAttrs;
   QVariantList vlst;
-  QgsPostgresSharedData *sdata;
+
+  std::shared_ptr< QgsPostgresSharedData > sdata( new QgsPostgresSharedData() );
 
   QgsField f0, f1, f2, f3;
 
@@ -208,15 +209,10 @@ void TestQgsPostgresProvider::testQuotedValueBigInt()
   f0.setType( QVariant::Int );
   f0.setTypeName( "int" );
 
-  fields.clear();
-  pkAttrs.clear();
-  vlst.clear();
-
   fields.append( f0 );
   pkAttrs.append( 0 );
   vlst.append( 42 );
 
-  sdata = new QgsPostgresSharedData;
   // for positive integers, fid  == the value, there is no map.
   sdata->insertFid( 42, vlst );
 
@@ -235,7 +231,7 @@ void TestQgsPostgresProvider::testQuotedValueBigInt()
   pkAttrs.append( 0 );
   vlst.append( -9223372036854775800LL ); // way outside int4 range
 
-  sdata = new QgsPostgresSharedData;
+  sdata->clear();
   sdata->insertFid( 1LL, vlst );
 
   QCOMPARE( QgsPostgresUtils::whereClause( 1LL, fields, NULL, QgsPostgresPrimaryKeyType::PktInt64, pkAttrs, std::shared_ptr<QgsPostgresSharedData>( sdata ) ), QString( "\"fld_bigint\"=-9223372036854775800" ) );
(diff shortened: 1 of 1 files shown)


espinafre:
Hi, @nyalldawson , what do you think of this? @elpaso , can you comment, since you've worked o this recently? Travis, at least, likes it.

tests/src/providers/testqgspostgresprovider.cpp
```diff
@@ -190,5 +194,112 @@ void TestQgsPostgresProvider::testDecodeDateTimes()
...
+  QgsField f0, f1, f2, f3;
+
+  // 4 byte integer
+  f0.setName( "fld_integer" );
+  f0.setType( QVariant::Int );
+  f0.setTypeName( "int" );
+
+  fields.clear();
+  pkAttrs.clear();
+  vlst.clear();
```
nyalldawson, line 213:
```suggestion
```

Not needed

tests/src/providers/testqgspostgresprovider.cpp
```diff
@@ -190,5 +194,112 @@ void TestQgsPostgresProvider::testDecodeDateTimes()
...
+
+  fields.append( f0 );
+  pkAttrs.append( 0 );
+  vlst.append( 42 );
+
+  sdata = new QgsPostgresSharedData;
```
nyalldawson, line 219:
```suggestion
  std::shared_ptr< QgsPostgreSharedData > sdata( new QgsPostgresSharedData() );
```

nyalldawson (commented):
More tests are always appreciated! :smile:

tests/src/providers/testqgspostgresprovider.cpp
espinafre: Resolved by commit 07358a0 .

tests/src/providers/testqgspostgresprovider.cpp
espinafre: Resolved by commit 07358a0 . Thanks for the hint!

tests/src/providers/testqgspostgresprovider.cpp
```diff
@@ -190,5 +194,108 @@ void TestQgsPostgresProvider::testDecodeDateTimes()
...
+  f1.setType( QVariant::LongLong );
+  f1.setTypeName( "int8" );
+
+  fields.clear();
+  pkAttrs.clear();
+  vlst.clear();
+
+  fields.append( f1 );
+  pkAttrs.append( 0 );
+  vlst.append( -9223372036854775800LL ); // way outside int4 range
```
elpaso, line 232: did you mean way outside int8 ?

tests/src/providers/testqgspostgresprovider.cpp
```diff
@@ -190,5 +194,108 @@ void TestQgsPostgresProvider::testDecodeDateTimes()
...
+  f2.setType( QVariant::Double );
+  f2.setTypeName( "float8" );
+
+  fields.clear();
+  pkAttrs.clear();
+  vlst.clear();
+
+  fields.append( f2 );
+  pkAttrs.append( 0 );
+  vlst.append( 3.141592741 );
```
elpaso, line 250: I like this one :+1:

elpaso (approved):
LGTM Thanks!

tests/src/providers/testqgspostgresprovider.cpp
espinafre: Outside int4, really, the value doesn't fit a 4-byte (32 bit) integer.

tests/src/providers/testqgspostgresprovider.cpp
espinafre: A little homage to the most well rounded release of QGIS yet 😁

espinafre:
Are we good to merge? Thanks!

Issue #37072, opened 2020-06-09
QGIS 3.10.6 does not build with GCC 10 and qt 5.15


buzo-ffm:
**Describe the bug**

QGIS 3.10.6 does not build with GCC 10 and qt 5.15. See attached output of `make -k` (second try).

**How to Reproduce**
Build with gcc 10.

**QGIS and OS versions**
QGIS 3.10.6 on Arch Linux (rolling release)

**Additional context**

I am trying to build [this package](https://aur.archlinux.org/packages/qgis-ltr/) for Arch Linux.

buzo-ffm:
[make.log.gz](https://github.com/qgis/QGIS/files/4753738/make.log.gz)

pblottiere:
For now I tried to compile QGIS master with GCC 10 and qt 5.15 and I have some errors to fix:

```
[ 79%] Built target translations
[ 79%] Generating gui/sip_guipart0.cpp, gui/sip_guipart1.cpp, gui/sip_guipart2.cpp, gui/sip_guipart3.cpp, gui/sip_guipart4.cpp, gui/sip_guipart5.cpp

sip: core/conversions.sip:1440: %MappedType template for this type has already been defined
make[2]: *** [python/CMakeFiles/python_module_qgis__gui_autogen.dir/build.make:1068: python/gui/sip_guipart0.cpp] Error 1
make[2]: *** Deleting file 'python/gui/sip_guipart0.cpp'
make[2]: Target 'python/CMakeFiles/python_module_qgis__gui_autogen.dir/build' not remade because of errors.
make[1]: *** [CMakeFiles/Makefile2:7364: python/CMakeFiles/python_module_qgis__gui_autogen.dir/all] Error 2
[ 79%] Generating analysis/sip_analysispart0.cpp, analysis/sip_analysispart1.cpp, analysis/sip_analysispart2.cpp, analysis/sip_analysispart3.cpp, analysis/sip_analysispart4.cpp, analysis/sip_analysispart5.cpp

sip: core/conversions.sip:1440: %MappedType template for this type has already been defined
make[2]: *** [python/CMakeFiles/python_module_qgis__analysis_autogen.dir/build.make:730: python/analysis/sip_analysispart0.cpp] Error 1
make[2]: *** Deleting file 'python/analysis/sip_analysispart0.cpp'
make[2]: Target 'python/CMakeFiles/python_module_qgis__analysis_autogen.dir/build' not remade because of errors.
[...]

PR #37116, opened 2020-06-10
Fix compilation with QT 5.15

pblottiere:
## Description

This PR fixes a bunch of issues raised during compilation with GCC 10, Qt/PyQt 5.15.0 and  SIP 4.19.23. Some of these have been raised in #37072.

It's mainly binding issues so a SIP expert should take a look to tell me if I did something wrong (@3nids?), especially about the `%MappedType template for this type has already been defined` issue.


3nids (approved):
good job.
If Travis is happy, I'm happy ;)

PR #37118, opened 2020-06-11
Automatically correct formatting of lists in doxygen and fix list appearance in PyQGIS docs

### LINKED ISSUE #37114: Make the prepare-commit script smart enough to catch and format list items for pyqgis doc
I think code formatting script should be improved so that it catches list items in the PyQGIS documentation, when changes are committed.

[Version in 3.10](https://qgis.org/pyqgis/3.10/server/QgsServerOgcApiHandler.html#qgis.server.QgsServerOgcApiHandler.write)
image

[Version in master](https://qgis.org/pyqgis/master/server/QgsServerOgcApiHandler.html#qgis.server.QgsServerOgcApiHandler.write)
image

More legible in master, right? This is a recurring fix I've been involved in lately but I think it's probably better and efficiently fixed programmatically than manually.
The fix is to add an empty line before the first item of the list, as you can see at
https://github.com/qgis/QGIS/commit/ea4f2bbfb962f309791cdf4468248061f8163089#diff-eeccb14026a2c077e0ab60f38986e7c8R193-R195

I have no idea what it takes to implement but we could consider catching these formatting errors with the prepare-commit script. A list item generally begins with `*  -` so maybe we can look for the first occurrence inside a method dox, and precede it with an empty `*` line.

There are still issues with the list item on multiple lines. i don't know how they should be formatted so can't suggest a fix.

nyalldawson:
Fixes #37114 -- all the required formatting rules are now applied in the prepare commit step.


DelazJ:
Thanks @nyalldawson for such a quick fix

3nids:
good job.
You'll soon be entitled as a perl scripts master :)

DelazJ:
@nyalldawson looks like I overlooked another situation of list items, using `* *`. See https://github.com/qgis/QGIS/blob/master/src/server/qgsaccesscontrolfilter.h#L41 <-> https://qgis.org/pyqgis/master/server/QgsAccessControlFilter.html#qgis.server.QgsAccessControlFilter

Issue #37070, opened 2020-06-09
Obtrusive warning in the auth dialog


m-kuhn:
On the authentication dialog, there is a warning if the password is saved in plain text.

The warning itself is important, but the visual attention it requests right now is slightly exaggerated.

![picture](https://user-images.githubusercontent.com/588407/84184412-94b19c80-aa8d-11ea-8a40-5365e3a01ca5.png)

Ideas:

 - De-bold (at least the content, maybe the title "Warning")
 - Switch colors away from red
 - Instead add a :warning: icon

elpaso:
Agreed, do we have anything about warnings in the HIG ? I think something in the lines of the message bar warning should probably looks better.

nirvn:
I don't think a message bar works there. It's a constant state/warning.

+1 for a warning sign and a short non red text. I'll give it a try this week.

elpaso:
> I don't think a message bar works there. It's a constant state/warning.
> 
> +1 for a warning sign and a short non red text. I'll give it a try this week.

I didn't mean to use a message bar, I was suggesting to use the same visual of the warning messages that appear in the message bar, i.e. same color, fonts etc..

nyalldawson:
This is fixed now, right?

nirvn:
Ya.

PR #37124, opened 2020-06-11
[ui] Reduce visual attention grab of the authentication settings wiget's warning label

nirvn:
## Description

This PR addresses the UI issue #37070 .

Before vs. PR:
image

image

### DIFF
--- a/src/gui/auth/qgsauthsettingswidget.cpp
+++ b/src/gui/auth/qgsauthsettingswidget.cpp
@@ -104,7 +104,7 @@ const QString QgsAuthSettingsWidget::dataprovider() const
 
 const QString QgsAuthSettingsWidget::formattedWarning( WarningType warning )
 {
-  QString out = tr( "<div>Warning: credentials stored as plain text in %1.</div>" );
+  QString out = tr( "Warning: credentials stored as plain text in %1." );
   switch ( warning )
   {
     case ProjectFile:
--- a/src/ui/auth/qgsauthsettingswidget.ui
+++ b/src/ui/auth/qgsauthsettingswidget.ui
@@ -173,21 +173,35 @@
          </property>
         </widget>
        </item>
-       <item row="3" column="0" colspan="3">
-        <widget class="QLabel" name="lblWarning">
-         <property name="styleSheet">
-          <string notr="true">QLabel{color: rgb(255, 0, 0);font-weight: bold;}</string>
-         </property>
-         <property name="text">
-          <string>Warning text!</string>
-         </property>
-         <property name="alignment">
-          <set>Qt::AlignCenter</set>
-         </property>
-         <property name="wordWrap">
-          <bool>true</bool>
-         </property>
-        </widget>
+       <item row="3" column="1" colspan="2">
+        <layout class="QHBoxLayout" name="horizontalLayout_2">
+         <item>
+           <widget class="QLabel" name="lblWarningIcon">
+           <property name="pixmap">
+            <pixmap>:/images/themes/default/mIconWarning.svg</pixmap>
+           </property>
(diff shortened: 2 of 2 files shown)


elpaso:
Looks good, but from bold red to light grey it's a big jump, did you try to make it black (fully saturated)?

nirvn:
@elpaso , the color is left to the OS or theme to decide. That said, I've removed the disabled state for the label, so it'll be black on default theme. Glad you like the change overall.

@elpaso , BTW, there's another UI/UX change which I would like to push. I think the tabs' order should be reversed, i.e. the "basic" tab comes before the more advanced "configurations" tab. It's an order that's to me a bit more logic IMHO. Also UX-wise, we can move from the basic -> to -> the configuration via the [ convert to configuration ] button, and not the other way around, which means we're better with the basic panel showing first.

elpaso:
> @elpaso , the color is left to the OS or theme to decide. That said, I've removed the disabled state for the label, so it'll be black on default theme. Glad you like the change overall.
> 

Thank you.

> @elpaso , BTW, there's another UI/UX change which I would like to push. I think the tabs' order should be reversed, i.e. the "basic" tab comes before the more advanced "configurations" tab. It's an order that's to me a bit more logic IMHO. Also UX-wise, we can move from the basic -> to -> the configuration via the [ convert to configuration ] button, and not the other way around, which means we're better with the basic panel showing first.

That change was deliberate and I''m -1 on changing that.

What we were trying to do was to avoid/prevent/discourage as hard as we could the use of plain text unencrypted credentials stored in the project file XML (people tend to share these .qg[sz] files).

That request came explicitly from large (mainly government) organizations.

The big red warning was part of that effort, but I agree that a big red warning was a punch in the eye from a UX/UI point of view.

m-kuhn (approved):
Much better

m-kuhn:
Regarding the advance vs. basic discussion.

I think whatever 80% of the people use should come first.
Is this authentication configuration meanwhile?
I find that many people still struggle to get the concept and set it up or that they explicitly want to bundle credentials in the project.
My feeling is that auth manager is still lacking some UX / quick start / functionality to deserve being the number one choice for everyone.

elpaso:
> Regarding the advance vs. basic discussion.
> 
> I think whatever 80% of the people use should come first.

Mind sharing the source of your statistics?

> Is this authentication configuration meanwhile?
> I find that many people still struggle to get the concept and set it up or that they explicitly want to bundle credentials in the project.
> My feeling is that auth manager is still lacking some UX / quick start / functionality to deserve being the number one choice for everyone.

May be true, there is always margin for a better UX. Do you have anything in mind to make the process smoother?

The point here is that the default setting that allows people to store credentials in plain text in the project was a blocker for QGIS adoption in these large government organizations (we are talking about hundreds or thousands of users). 

Maybe the current warning is now sufficient but I doubt it.

m-kuhn:
> Do you have anything in mind to make the process smoother?

- Initial configuration of an authid requires going through the authmanager dialog
  - If a project is opened with an unconfigured authid, the system asks for the password every time. Current situation is, the user has to copy the authid, go the the authentication manager, and create a new matching configuration. It would be good if it was possible to directly create this authid when asked for login credentials.
  - For an organisation wide setup of authids, the current possibilities are:
    - Copying an auth database to every user
    - Providing users with installation instructions including entering a cryptic id
    - It would be good if the authid could be specified in global_settings and the user guided in setting up login credentials for this authid.
    - Introduction of a global "default" authid, so generally this one would be used everywhere unless someone explicitly wants a "special" authid.
- It's currently not possible to share authentication credentials with a project (which usually ends up with the fact that it's sent by mail instead)

PR #37129, opened 2020-06-11
Matrix configuration dialog should open as an inline panel, not a dialog

nyalldawson:
Since all the other processing widgets have abandoned modal dialogs, this should do the same...

### DIFF
--- a/src/gui/processing/qgsprocessingmatrixparameterdialog.cpp
+++ b/src/gui/processing/qgsprocessingmatrixparameterdialog.cpp
@@ -23,8 +23,8 @@
 
 ///@cond NOT_STABLE
 
-QgsProcessingMatrixParameterDialog::QgsProcessingMatrixParameterDialog( QWidget *parent, Qt::WindowFlags flags, const QgsProcessingParameterMatrix *param, const QVariantList &initialTable )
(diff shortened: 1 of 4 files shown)


roya0045:
@nyalldawson I know this is the wrong place to ask, but it felt it fit with the PR: in your changes to the modeler, was there any changes made that allowed to have data defined output (both the output file path/name and the QGIS alias )?

nyalldawson:
@roya0045 

You can now data define output paths when making models -- is that what you mean?

roya0045:
@nyalldawson Yes but also the name of the temporary/memory output.
Let's say I want to output 5 memory layers from my model, validate then and then swing them in a DB, it just slow things down to rename them.

But having the DD output file is great!

nyalldawson:
@roya0045 
>  Yes but also the name of the temporary/memory output.

Ahh - I think in that case you want to use the "rename layer" algorithm. That accepts data defined strings as the new layer name. You'd insert this algorithm after the main algorithm, and it would rename the output coming from the other algorithm...

roya0045:
@nyalldawson I completely overlooked that function... Thanks for reminding me!

Issue #37028, opened 2020-06-08
Font colour in grid decoration


hrdrckx:
QGIS users would greatly benifit from the inclusion of choice of font colour in the
view>decorations>grid & grid properties menu

Currently the Grid line symbol colour can be chose but the text which labels these grid lines cannot and remains the default black which often renders the labels invisible on dark backgrounds.

rduivenvoorde:
@hrdrckx (personal observation) I think if you add an image with such an ugly rendering, it will encourage more people to invest time in this :-) 
At this moment I personally cannot imagine exactly what you mean (but probably others can).

hrdrckx:
![screenshot_20200611-141954](https://user-images.githubusercontent.com/40512017/84352104-49af9a80-abef-11ea-98f3-0e79c32893cc.png)

something like this that just came up on my screen?  I'm sure i could make something much worse if i tried :P

rduivenvoorde:
@hrdrckx ok NOW it is clear :-)

For others: talking about this dialog:

![Screenshot-20200611085534-410x603](https://user-images.githubusercontent.com/731673/84354543-5c12df80-abc1-11ea-8eee-45fe47f902d3.png)

The 'decoration'-items are very old functions, often other options are available.
I was thinking about: there is a 'processing' function called 'Create Grid'.. that creates a grid LAYER which you can style in whatever style you want...

![Screenshot-20200611085005-951x674](https://user-images.githubusercontent.com/731673/84354080-9039d080-abc0-11ea-95a3-9233a07a6e5c.png)

Mmm, trying that out now, I realize it is a hassle to get labels on the edge of your mapcanvas...

So the old decoration thing has it's value apparently. I'll try on the discussion list what others think: add a color button (or text styling?) to the decorations/grid dialog. 
OR maybe create some labeling regime in which the labels always try to end up on the edges of the canvas. In that way it would be possible to create such labels in all other ways, AND we could add it to the Create Grid dialog and maybe get rid of the old decorations stuff?

rduivenvoorde:
Ah, and I also realized now, off course creating a Layout also has the possibilities to create a grid AND you can set the font color for it :-)

![Screenshot-20200611091339-1690x919](https://user-images.githubusercontent.com/731673/84356091-fa07a980-abc3-11ea-844c-309b5fd9a886.png)

hrdrckx:
@rduivenvoorde - great overview thanks!

Regarding your other suggestions:
[...]

PR #37133, opened 2020-06-11
[decorations] Fix hardcoded black color for the grid decoration

nirvn:
## Description

This PR fixes an issue with the grid decoration (issue #37028) whereas the hardcoded black color used to render grid annotations made those unreadable when hovering dark parts of the canvas.

~~Before vs~~ PR:
Screenshot from 2020-06-11 15-42-12

The fix involved using our QgsTextRenderer class, which as a collateral benefit allows us to get rid of a bunch of unmaintained code in the grid decoration class. Win-win.

### DIFF
--- a/src/app/decorations/qgsdecorationgrid.cpp
+++ b/src/app/decorations/qgsdecorationgrid.cpp
@@ -23,16 +23,17 @@
 #include "qgslogger.h"
 #include "qgsmaplayer.h"
 #include "qgsrasterlayer.h"
+#include "qgsmapcanvas.h"
 #include "qgsmaptopixel.h"
 #include "qgspathresolver.h"
 #include "qgspointxy.h"
 #include "qgsproject.h"
-#include "qgssymbollayerutils.h" //for pointOnLineWithDistance
-#include "qgssymbol.h" //for symbology
+#include "qgssymbollayerutils.h"
+#include "qgssymbol.h"
 #include "qgsmarkersymbollayer.h"
 #include "qgsreadwritecontext.h"
 #include "qgsrendercontext.h"
-#include "qgsmapcanvas.h"
+#include "qgstextrenderer.h"
 
 #include <QPainter>
 #include <QAction>
@@ -50,9 +51,6 @@
 
 //non qt includes
 #include <cmath>
-
-
-#define FONT_WORKAROUND_SCALE 10 //scale factor for upscaling fontsize and downscaling painter
 
 
 QgsDecorationGrid::QgsDecorationGrid( QObject *parent )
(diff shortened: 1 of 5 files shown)


Saijin-Naib:
@nirvin, is it possible with this PR to set a color value of "inverted" to ensure optimal visibility no matter the underlying subject matter?

nirvn:
@Saijin-Naib , sure:
Screenshot from 2020-06-11 21-51-14

The fix involved switching from raw Qt text rendering to QGIS' powerful QgsTextRenderer class, comes in with all of its features and goodies.

nirvn:
Yeah, that's been possible for a long while.

nyalldawson (approved):

rduivenvoorde:
@nirvn SUPERcool fix!!!! I was thinking about trying to add a button or so, but this is so much better! Thanks!

One thing I see is that the text 'Font' in the button in the dialog dialog is partly hidden (also in the earlier dialog, see https://github.com/qgis/QGIS/issues/37028 for a screenshot):

Screenshot-20200612090643-402x577

Thought I had some strange font (Cantarell Regular 10pt) but even if I change the font it stays like this. Is the button smaller then usual? Or the caption by default a too large text-size?

Another observation: apparently there is a difference in using the little arrow (you are able to change some basic props) or 'click' the font button... in which you can change all AND the basic props. Would it not be better to only have a Font-button? (I missed the dialog at first because I am used to using the dropdown arrow if I see it... Ah.. duh.. now I see it is always like this Argh. never-known-QGIS-option :-)

I understand that it has always like this, so leave it as is. I'm very gratefull for this work already!

nirvn:
@rduivenvoorde , The font size of the preview drawn on the button matches the text format settings, that's the case for all font buttons.

I love and use the down arrow menu _all the time_. It's a quick way to change color, font family, and font size without opening a dialog. If you don't like it, just don't click on the down arrow :wink: That menu is also useful when copy pasting text formatting across several buttons.

rduivenvoorde:
You are right, the dropdown IS a good shortcut, it is just that intuitively that was my first try to look at your fix... and I missed it ... and just on the second one I tried to push the button. But.. . I will learn, thanks :-)

About the font in the button, I first thought: I never see this problem... 
Not that I have to set the font size to 8pt to at least be able to read the characters (see screenie).

But apparently we use different ways to show a font. Below you see font in the label stuff, there we only show the name in a fixed font and without preview or size:

Screenshot-20200612100100-766x236

But upon further searching indeed in the layouts we do the same as in the grid dialog: see the Font button here

Screenshot-20200612100458-739x550

Will check windows also

rduivenvoorde:
Ah, on windows apparently the 'button'(?) behaves differently.
Indeed below Font-size is 16pt(!)

Screenshot-20200612101455-292x126

And this is 10 pt (both on windows, via Remote desktop that is why lower img quality)

Screenshot-20200612101527-274x112

PR #37135, opened 2020-06-11
Fix editor vertex with "non-dot" locales

### LINKED ISSUE #29682: can't change value in vertex editor if "locale" settings use commas as decimal separator
Original Redmine Issue: [21867](https://issues.qgis.org/issues/21867)
Affected QGIS version: 3.6.1
Redmine category:digitising

---
if you have a locale that uses commas as decimal separator then manually editing values in the vertex editor does not work. Workaround: set QGIS locale to English.

---

Related issue(s): #29811 (duplicates)
Redmine related issue(s): [21996](https://issues.qgis.org/issues/21996)

---

elpaso:
Fixes #29682

### DIFF
--- a/src/app/vertextool/qgsvertexeditor.cpp
+++ b/src/app/vertextool/qgsvertexeditor.cpp
@@ -504,6 +504,6 @@ void CoordinateItemDelegate::setEditorData( QWidget *editor, const QModelIndex &
   QLineEdit *lineEdit = qobject_cast<QLineEdit *>( editor );
   if ( lineEdit && index.isValid() )
   {
-    lineEdit->setText( QLocale().toString( index.data( ).toDouble( ) ) );
+    lineEdit->setText( QLocale().toString( index.data( ).toDouble( ), 'f', 4 ) );
   }
 }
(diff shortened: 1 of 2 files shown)


src/app/vertextool/qgsvertexeditor.cpp
```diff
@@ -499,4 +499,11 @@ void CoordinateItemDelegate::setModelData( QWidget *editor, QAbstractItemModel *
   }
 }
 
-
+void CoordinateItemDelegate::setEditorData( QWidget *editor, const QModelIndex &index ) const
+{
+  QLineEdit *lineEdit = qobject_cast<QLineEdit *>( editor );
+  if ( lineEdit && index.isValid() )
+  {
+    lineEdit->setText( QLocale().toString( index.data( ).toDouble( ) ) );
```
nyalldawson, line 507: Should we use a fixed precision here to avoid scientific notation values?

elpaso: Yes, good point, I'll use the same format as in displayText.

PR #37136, opened 2020-06-11
Fix decimal point issues on raster properties forms,  Fixes #33859

### LINKED ISSUE #33859: Can't set min/max values to decimal in raster symbology with QGis Linux versions
When setting the properties of a raster layer, I encounter an issue related to Min/Max values form fields (with Singleband pseudo color render type) : default precalculated value can be decimal, but I can't manually set the value to decimal.

Strange thing is : form field accepts dot symbol `.` if such symbol is already present, but the key doesn't work if the default value is an integer. I also cannot copy/paste a value in those fields.

In my case, as a matter of example, default max value is `30.8921` and default min value is `-100`. If I suppress only decimal values in the max value field, I can achieve to set to `0.5`, but I clean the field, I am not able to add a dot. In the min value field, I cannot add a dot at all.

This issue occurred on QGis 3.4 LTS, 3.8, 3.10 with Ubuntu 18.04, Debian 10 and MacOS 10 Catalina. Doesn't occur on Windows versions (tested only on QGis 3.4 with Windows 10).

How to reproduce :
- Add raster layer in QGis (3.4 for example) under Ubuntu 18.04 (didn't test with other versions)
- Select the layer and open the Properties window
- In symbology, choose `Singleband pseudocolor` for `Render type`
- Try to set default Min value to `-0.5` and Max value to `0.5`

Please tell me if you need additional infos, this bug looks quite surprising in my opinion (or is it ?).

SebastienPeillet:
## Description

Issue explained and discussed here : #33859

QGIS use QDoubleValidator on QLineEdit in several forms, including raster layer forms.

As QValidator is initialized with the current locale, the expected decimal point is determined by the locale.

For exemple in french, it' is a comma and not a point.

Sometimes the string value is converted to double value, but not with the QLocale().toDouble() method. In this case the string value can't be converted.

I fixed the bug on raster properties form (but I expect it exists in some other form). The affected QLineEdit  are : 
- the min and max range QLineEdits in gray single band, pseudo color single band, color multi band forms
- the transparency QLineEdit

I also fixed values display in the same forms plus palette form

To fix it , I create a new validator, called QgsDoubleValidator (as the Qgslonglongvalidator) using the same principle that use [QgsFieldValidator](https://github.com/qgis/QGIS/blob/master/src/gui/qgsfieldvalidator.cpp) for double type.

The regular expression allows to use the point and the locale decimal point as decimal point

Futhermore a static method toDouble is created to convert the string to double whatever if it is the point or the locale decimal point that is used.

I started as I will fix a bug, but now maybe it's more a feature...


src/gui/qgsdoublevalidator.h
```diff
@@ -0,0 +1,144 @@
...
+    Q_OBJECT
+
+  public:
+    explicit QgsDoubleValidator( QObject *parent )
+      : QRegExpValidator( parent )
+      , b( std::numeric_limits<qreal>::min() )
+      , t( std::numeric_limits<qreal>::max() )
+    {
+      // The regular expression accept double with point as decimal point but also the locale decimal point
+      QRegExp reg( QStringLiteral( "-?[\\d]{0,1000}([\\.%1][\\d]{0,1000})?([\\d]e[+-]?[\\d]{0,1000})?" ).arg( locale().decimalPoint() ) );
```
nyalldawson, line 44: please use QRegularExpression instead -- QRegExp is deprecated

elpaso, line 44: Can you use raw strings for readability?
      `QRegExp reg( QStringLiteral( R"(.....)" ...`

src/gui/qgsdoublevalidator.h
```diff
@@ -0,0 +1,144 @@
...
+      this->setRegExp( reg );
+    }
+
+    QgsDoubleValidator( double bottom, double top, QObject *parent )
+      : QRegExpValidator( parent )
+      , b( bottom )
+      , t( top )
+    {
+      // The regular expression accept double with point as decimal point but also the locale decimal point
+      QRegExp reg( QStringLiteral( "-?[\\d]{0,1000}([\\.%1][\\d]{0,1000})?([\\d]e[+-]?[\\d]{0,1000})?" ).arg( locale().decimalPoint() ) );
```
elpaso, line 62: If the expression is the same you should turn it into a constant (uppercase) and place it at the top of the file.

tests/src/gui/testqgsdoublevalidator.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+void TestQgsDoubleValidator::init()
+{
+}
+
+void TestQgsDoubleValidator::cleanup()
+{
+}
+
+void TestQgsDoubleValidator::validate()
```
elpaso, line 55: this is fine but I would recommend to use _data driven testing here, it will be easier to read the test and add new cases: https://doc.qt.io/qt-5/qttestlib-tutorial2-example.html

tests/src/gui/testqgsdoublevalidator.cpp
```diff
@@ -0,0 +1,138 @@
...
+{
+    QLocale().setDefault(QLocale::English);
```
elpaso, line 57: Can you please add test cases with group (thousands) separators?

I think we should test and accept all valid inputs inputs the user locale but also accept the C locale valid inputs as a fallback, for example assuming locale is it_IT, these would be all valid:

1,234  ( interpreted as 1.234 )
1.234 (interpreted as 1234 ... this is ambiguous but the user locale takes precedence)
1'234,0000001 ( 1234.0000001 )
1234,0000001 ( 1234.0000001 )
the following are interpreted using the fallback C locale:
1234.0000001  ( 1234.0000001  because it's valid in the fallback C locale)
1,234.0000001 ( 1234.0000001 because it's valid in the fallback C locale and it's not ambiguous)
1,234.000  ( 1234.0000001 because it's valid in the fallback C locale and it's not ambiguous)

elpaso (changes requested):
Please:
- add a cpp implementation file
- tests with group separator and C locale fallback
- in the new header file: class doxygen documentation for the class and for the methods

src/gui/qgsdoublevalidator.h
SebastienPeillet: :+1: Done, I placed it in the new cpp file

tests/src/gui/testqgsdoublevalidator.cpp
SebastienPeillet: I didn't know the trick, thanks for the documentation. I made the changes

src/gui/qgsdoublevalidator.h
SebastienPeillet: I changed it with the constant declaration !

SebastienPeillet: Ok, I made the changes

tests/src/gui/testqgsdoublevalidator.cpp
SebastienPeillet: I add tests with group separator, as you said there is some ambiguous cases, in particular when C locale and user locale share the same decimalpoint and/or group separator. To handle it, the test [changes the expected value](https://github.com/SebastienPeillet/QGIS/commit/a52df08fc944e0563c5525fc2ba9b2422a349909#diff-cb725c98171908938554db1067e951f9R136) (validate method will return QValidator::Intermediate and not QValidator::Invalid).

Tell me if you think there is still missing tests

SebastienPeillet:
@elpaso @nyalldawson thanks for your reviews ! i'm way more satisfy to this PR with your advices.

src/gui/qgsdoublevalidator.cpp
```diff
@@ -0,0 +1,131 @@
...
+const QString PERMISSIVE_DOUBLE = R"(-?[\d]{0,1000}([\.%1][\d]{0,1000})?(e[+-]?[\d]{0,%2})?)";
+
+QgsDoubleValidator::QgsDoubleValidator( QObject *parent )
+  : QRegularExpressionValidator( parent )
+  , b( std::numeric_limits<qreal>::min() )
+  , t( std::numeric_limits<qreal>::max() )
+{
+  // The regular expression accept double with point as decimal point but also the locale decimal point
+  QRegularExpression reg( PERMISSIVE_DOUBLE.arg( locale().decimalPoint() ).arg( 1000 ) );
+  this->setRegularExpression( reg );
```
nyalldawson, line 36:
```suggestion
  setRegularExpression( reg );
```

src/gui/qgsdoublevalidator.cpp
```diff
@@ -0,0 +1,131 @@
...
+}
+
+QgsDoubleValidator::QgsDoubleValidator( QRegularExpression reg, double bottom, double top, QObject *parent )
```
nyalldawson, line 39:
```suggestion
QgsDoubleValidator::QgsDoubleValidator( const QRegularExpression& reg, double bottom, double top, QObject *parent )
```

src/gui/qgsdoublevalidator.cpp
```diff
@@ -0,0 +1,131 @@
...
+  : QRegularExpressionValidator( parent )
+  , b( bottom )
+  , t( top )
+{
+  this->setRegularExpression( reg );
```
nyalldawson, line 44:
```suggestion
  setRegularExpression( reg );
```

src/gui/qgsdoublevalidator.cpp
```diff
@@ -0,0 +1,131 @@
...
+}
+
+QgsDoubleValidator::QgsDoubleValidator( double bottom, double top, QObject *parent )
+  : QRegularExpressionValidator( parent )
+  , b( bottom )
+  , t( top )
+{
+  // The regular expression accept double with point as decimal point but also the locale decimal point
+  QRegularExpression reg( PERMISSIVE_DOUBLE.arg( locale().decimalPoint() ).arg( 1000 ) );
+  this->setRegularExpression( reg );
```
nyalldawson, line 54:
```suggestion
  setRegularExpression( reg );
```

src/gui/qgsdoublevalidator.cpp
```diff
@@ -0,0 +1,131 @@
...
+}
+
+QgsDoubleValidator::QgsDoubleValidator( double bottom, double top, int dec, QObject *parent )
+  : QRegularExpressionValidator( parent )
+  , b( bottom )
+  , t( top )
+{
+  // The regular expression accept double with point as decimal point but also the locale decimal point
+  QRegularExpression reg( PERMISSIVE_DOUBLE.arg( locale().decimalPoint() ).arg( QString::number( dec ) ) );
+  this->setRegularExpression( reg );
```
nyalldawson, line 64:
```suggestion
  setRegularExpression( reg );
```

src/gui/qgsdoublevalidator.h
```diff
@@ -0,0 +1,155 @@
+#ifndef QGSDOUBLEVALIDATOR_H
```
nyalldawson, line 1: convention is copyright at immediate top of file, followed by header guard

src/gui/qgsdoublevalidator.h
```diff
@@ -0,0 +1,155 @@
...
+     * Constructor for QgsDoubleValidator.
+     * \since QGIS 3.14
+     */
+    explicit QgsDoubleValidator( QObject *parent );
+
+    /**
+     * Constructor for QgsDoubleValidator.
+     * \since QGIS 3.14
+     */
+    QgsDoubleValidator( QRegularExpression reg, double bottom, double top, QObject *parent );
```
nyalldawson, line 54:
```suggestion
    QgsDoubleValidator( const QRegularExpression &expression, double bottom, double top, QObject *parent );
```

Avoid abbreviations in public API

nyalldawson: Also add explanation of what bottom/top means

src/gui/qgsdoublevalidator.h
```diff
@@ -0,0 +1,155 @@
...
+     * Constructor for QgsDoubleValidator.
+     * \since QGIS 3.14
+     */
+    QgsDoubleValidator( double bottom, double top, QObject *parent );
+
+    /**
+     * Constructor for QgsDoubleValidator.
+     * \since QGIS 3.14
+     */
+    QgsDoubleValidator( double bottom, double top, int dec, QObject *parent );
```
nyalldawson, line 66: Explain "dec" (and expand the abbreviation)

src/gui/qgsdoublevalidator.h
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
```
nyalldawson, line 68: we don't usually include dox for overriden functions

src/gui/qgsdoublevalidator.h
```diff
@@ -0,0 +1,155 @@
...
+     * \since QGIS 3.14
+     */
+    QValidator::State validate( QString &input, int & ) const override;
+
+    /**
+     * Evaluates input QString validity according to QRegularExpression
+     * and ability to be converted in double value.
+     * \since QGIS 3.14
+     */
+    QValidator::State validate( QString input ) const;
```
nyalldawson, line 80:
```suggestion
    QValidator::State validate( QString input ) const;
```

Shouldn't this be a reference to input?

src/gui/qgsdoublevalidator.h
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * Converts QString to double value.
+     * It used locale interpretation first
+     * and C locale interpretation as fallback
+     * \since QGIS 3.14
+     */
+    static double toDouble( QString input, bool *ok ) SIP_SKIP;
```
nyalldawson, line 88:
```suggestion
    static double toDouble( const QString& input, bool *ok ) SIP_SKIP;
```

src/gui/qgsdoublevalidator.h
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * Converts QString to double value.
+     * It used locale interpretation first
+     * and C locale interpretation as fallback
+     * \since QGIS 3.14
+     */
+    static double toDouble( QString input );
```
nyalldawson, line 96:
```suggestion
    static double toDouble( const QString &input );
```

src/gui/qgsdoublevalidator.h
```diff
@@ -0,0 +1,155 @@
...
+    double top() const { return t; }
+
+  private:
+    // Disables copy constructing
+    Q_DISABLE_COPY( QgsDoubleValidator )
+
+    /**
+     * Bottom range limit
+     */
+    double b;
```
nyalldawson, line 147:
```suggestion
    double mMinimum;
```

src/gui/qgsdoublevalidator.h
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * Top range limit
+     */
+    double t;
```
nyalldawson, line 152:
```suggestion
    double mMaximum;
```

src/gui/qgsdoublevalidator.cpp
```diff
@@ -0,0 +1,131 @@
...
+}
+
+QValidator::State QgsDoubleValidator::validate( QString &input, int & ) const
+{
+  if ( input.isEmpty() )
+    return Intermediate;
+
+
+  bool ok = false;
+  double entered = QgsDoubleValidator::toDouble( input, &ok );
```
elpaso, line 74:
```suggestion
  const double entered = QgsDoubleValidator::toDouble( input, &ok );
```

src/gui/qgsdoublevalidator.cpp
```diff
@@ -0,0 +1,131 @@
...
+}
+
+QValidator::State QgsDoubleValidator::validate( QString input ) const
+{
+  if ( input.isEmpty() )
+    return Intermediate;
+
+
+  bool ok = false;
+  double entered = QgsDoubleValidator::toDouble( input, &ok );
```
elpaso, line 96:
```suggestion
  const double entered = QgsDoubleValidator::toDouble( input, &ok );
```

src/gui/qgsdoublevalidator.h
```diff
@@ -0,0 +1,155 @@
...
+#include <QLocale>
+#include "qgis_gui.h"
+#include "qgis_sip.h"
+
+/**
+ * \ingroup gui
+ *
+ * QgsDoubleValidator is a QLineEdit Validator that combines QDoubleValidator
+ * and QRegularExpressionValidator to allow user to enter double with both
+ * local and C interpretation.
```
elpaso, line 34:
```suggestion
 * local and C interpretation as a fallback.
```

src/gui/qgsdoublevalidator.h
```diff
@@ -0,0 +1,155 @@
...
+ */
+class GUI_EXPORT QgsDoubleValidator : public QRegularExpressionValidator
+{
+    Q_OBJECT
+
+  public:
+
+    /**
+     * Constructor for QgsDoubleValidator.
+     * \since QGIS 3.14
```
elpaso, line 46: I don't think you need to add \since unless it's different fom the whole class.

tests/src/gui/testqgsdoublevalidator.cpp
```diff
@@ -0,0 +1,185 @@
...
+void TestQgsDoubleValidator::validate_data()
+{
+  QTest::addColumn<QString>( "actualState" );
+  QTest::addColumn<int>( "expState" );
+
+  QTest::newRow( "C decimal" ) << QString( "4cd6" ) << int( QValidator::Acceptable );
+  QTest::newRow( "locale decimal" ) << QString( "4ld6" ) << int( QValidator::Acceptable );
+  QTest::newRow( "locale decimal" ) << QString( "4444ld6" ) << int( QValidator::Acceptable );
+
+  // QgsDoubleValidator doesn't expect group separator but it tolerates it,
```
elpaso, line 66: why is that considered non acceptble? 

If the user enters the numbr `1234.56789` in it_IT locale as `1.234,567899` this would be correctly converted by `QLocale::toDouble()` without any required editing by the user, so I think it should be valid.

nyalldawson:
Too risky for 3.14.0 -- deferred

tests/src/gui/testqgsdoublevalidator.cpp
SebastienPeillet: Actually it's not "acceptable" because of the regular expression constraint,  but the value will be converted. The actual regexp doesn't include group separator. The user can't enter a number using group separator, except when he pastes the whole number.

Intermediate validation is based on the `QLocale::toDouble()` result (so Intermediate result can be converted) and to get Acceptable result, it needs to fit regexp condition and range limit.

I choose to put aside group separator to ensure that the entered value will be converted no matter what. For example, I wish to enter the number 1234567.89 in it_IT locale as 1.234.567,89 from the left to the right : 

|String value | Result | comment|
|-----------------|----------|-------------|
|1 | 1 |it_IT locale |
|1. | 1 |C locale |
|1.2 | 1.2 |C locale |
|1.23 | 1.23 |C locale |
|1.234 | 1234 |it_IT locale |
|1.234. & 1.234.5 | 0 | can't be converted as it excepted a number like x.xxx.xxx in it_IT locale or x.xxxxxxxx in C locale. At this point the entered value it can't be converted and does not fit regexp. To avoid that user enters wrong value the validator does not allow to add the second '.' without the 3 digits at the same time as it will be possible to convert it.
|1.234.56 | 0 | can't be converted and does not fit regexp but allow to add an extra digit as it will be possible to convert the string.|
1.234.567 | 1234567 |it_IT locale|
1.234.567,8 |1234567.8 | it_IT locale|
1.234.567,89 |1234567.89 | it_IT locale|

So in conclusion @elpaso, you can paste value with group separator and it will be converted in double value if you validate the form. Maybe we should be more flexible on the group separator but it could lead to wrong value and on my opinion QValidator exists to avoid these mistakes

SebastienPeillet:
> Too risky for 3.14.0 -- deferred

Yes, this is what I wanted to say by writing `I started as I will fix a bug, but now maybe it's more a feature...`

tests/src/gui/testqgsdoublevalidator.cpp
elpaso: That makes sense, thank you for clarifying.

src/gui/qgsdoublevalidator.cpp
SebastienPeillet: Done !

src/gui/qgsdoublevalidator.cpp
SebastienPeillet: Done !

src/gui/qgsdoublevalidator.cpp
SebastienPeillet: Done !

src/gui/qgsdoublevalidator.h
SebastienPeillet: Done !

src/gui/qgsdoublevalidator.h
SebastienPeillet: Done !

src/gui/qgsdoublevalidator.h
SebastienPeillet: Done !

src/gui/qgsdoublevalidator.h
SebastienPeillet: Done !

src/gui/qgsdoublevalidator.cpp
SebastienPeillet: Done !

src/gui/qgsdoublevalidator.cpp
SebastienPeillet: Done !

src/gui/qgsdoublevalidator.cpp
SebastienPeillet: Done !

src/gui/qgsdoublevalidator.cpp
SebastienPeillet: Done !

src/gui/qgsdoublevalidator.h
SebastienPeillet: Done !

src/gui/qgsdoublevalidator.h
SebastienPeillet: ok, I wasn't sure !

src/gui/qgsdoublevalidator.h
SebastienPeillet: You're right, I changed it

elpaso:
@SebastienPeillet can you please make the requested changes so that we can finally merge?

src/gui/qgsdoublevalidator.h
SebastienPeillet: Done !

src/gui/qgsdoublevalidator.h
SebastienPeillet: Done !

src/gui/qgsdoublevalidator.h
SebastienPeillet: Done !

src/gui/qgsdoublevalidator.h
SebastienPeillet: Done !

SebastienPeillet:
> @SebastienPeillet can you please make the requested changes so that we can finally merge?

@elpaso It's done, sorry I didn't see the remained requested changes before (hidden in the conversation by default). Thanks for the reminder !

src/gui/qgsdoublevalidator.h
```diff
@@ -0,0 +1,149 @@
...
+
+    /**
+     * Constructor for QgsDoubleValidator.
+     */
+    explicit QgsDoubleValidator( QObject *parent );
+
+    /**
+     * Constructor for QgsDoubleValidator.
+     *
+     * bottom is the minimal range limit accepted by the validator
```
elpaso, line 52:
```suggestion
     * \param bottom the minimal range limit accepted by the validator
```
can you please use the `\param` sintax for arguments? (here and everywhere)

src/gui/qgsdoublevalidator.h
```diff
@@ -0,0 +1,149 @@
...
+     * top is the maximal range limit accepted by the validator
+     * decimal is the number of decimal accepted by the validator
+     */
+    QgsDoubleValidator( double bottom, double top, int decimal, QObject *parent );
+
+
+    QValidator::State validate( QString &input, int & ) const override SIP_SKIP;
+
+    /**
+     * Evaluates input QString validity according to QRegularExpression
```
elpaso, line 78:
```suggestion
     * Evaluates \a input string validity according to QRegularExpression
```

elpaso: and all other occurrences too...

SebastienPeillet: It's done ! To avoid new corrections on this next time, is there a tacit rule for italic in documentation, can't find reference of it in [Qgis coding standard](https://docs.qgis.org/3.10/en/docs/developers_guide/codingstandards.html#api-documentation) ?

src/gui/qgsdoublevalidator.h
SebastienPeillet: My bad... I add it !

src/gui/qgsdoublevalidator.h
elpaso: It's not about italic, it's for "Use this command to refer to member arguments in the running text." instead of using `\param` you can use `\a` for an inline description of the argument.

elpaso (approved):

PR #37137, opened 2020-06-11
fixes invalid relation in search form

### LINKED ISSUE #34410: 'creating a relation sets wrong expression in "display expression" and breaks the "select features by value" functionality
I created a `1:m` relationship between two polygonal vector layers (_I attach data and project_), 

image

doing a search by value using the **F3** function key (selection by value / select feature) on the parent layer, I get a red error message.

```
The relationship is invalid. Check that the relationship definitions are correct
```

image

and strange, because the **COD_REG** field is `NULL`

In reality, the relationship is correct because by opening the attribute table (parent layer) in module mode, everything works well:

image

---

## DATA AND PROJECT

[BUG_selectbyrelationship.zip](https://github.com/qgis/QGIS/files/4186674/BUG_selectbyrelationship.zip)

---

## the problem occurs in QGIS 3.4.15, 3.10.2-2 and master ` with clean profile !!!`

---

```
Versione di QGIS 3.10.2-A Coruña
Revisione codice QGIS d4cd3cfe5a
Compilato con Qt 5.11.2 
Esecuzione con Qt 5.11.2
Compilato con GDAL/OGR 3.0.3
Esecuzione con GDAL/OGR 3.0.4
Compilato con GEOS 3.8.0-CAPI-1.13.1
Esecuzione con GEOS 3.8.0-CAPI-1.13.1 
Compiled against SQLite 3.29.0
Running against SQLite 3.29.0
Versione client PostgreSQL 11.5
Versione SpatiaLite 4.3.0
Versione QWT 6.1.3
Versione QScintilla2 2.10.8
Compiled against PROJ 6.3.0
Running against PROJ Rel. 6.3.0, January 1st, 2020
OS Version Windows 10 (10.0)
```

### OSGeo4W64

3nids:
fixes #34410

If the relation is not defined in the config, guess it from the layer.
The config is empty while we are initalizing the attribute form of the referencing layer for the search widget wrapper.
There might be a better (more correct) fix by correctly initializing the form but I could not easily find out.
This approach sounds reasonable to me.

### DIFF
--- a/src/gui/editorwidgets/qgsrelationreferencesearchwidgetwrapper.cpp
+++ b/src/gui/editorwidgets/qgsrelationreferencesearchwidgetwrapper.cpp
@@ -207,6 +207,8 @@ void QgsRelationReferenceSearchWidgetWrapper::initWidget( QWidget *editor )
   }
 
   QgsRelation relation = QgsProject::instance()->relationManager()->relation( config( QStringLiteral( "Relation" ) ).toString() );
+  if ( !relation.isValid() && !layer()->referencingRelations( mFieldIdx ).isEmpty() )
+    relation = layer()->referencingRelations( mFieldIdx )[0];
   mWidget->setRelation( relation, false );
 
   mWidget->showIndeterminateState();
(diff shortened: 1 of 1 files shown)


agiudiceandrea:
Could this be merged?

m-kuhn:
> There might be a better (more correct) fix by correctly initializing the form but I could not easily find out.

Agreed, that's why I postponed for so long.
In short, while it's possible to have multiple relations defined for a field, this will always take the first one. In practice this will very often be sufficient. So let's merge.
However, the first one who will run into this limitation will be very surprised.

m-kuhn (approved):

3nids:
or we do this, only if there is a single relation existing?

m-kuhn:
> or we do this, only if there is a single relation existing?

That would make it more transparent

PR #37139, opened 2020-06-11
feature picker model: correctly initalize fields for invalid feature

### LINKED ISSUE #37100: Expression editor: syntax check fails if table is empty
**Describe the bug**
When the table is empty - naturally the preview fails - but the syntax check should not fail:
image

Table has at least one record (syntax check of the exact same expression as before is fine):
image

I would expect the syntax check to also work fine on an empty table.

**QGIS and OS versions**
Tested in QGIS master (3.13) on Linux Ubuntu 18.04

3nids:
and force it to be valid in the expression preview widget, so expression can be evaluated even when layer has no feature

fixes #37100

### DIFF
--- a/src/gui/qgsexpressionpreviewwidget.cpp
+++ b/src/gui/qgsexpressionpreviewwidget.cpp
@@ -48,7 +48,18 @@ void QgsExpressionPreviewWidget::setExpressionText( const QString &expression )
 void QgsExpressionPreviewWidget::setCurrentFeature( const QgsFeature &feature )
 {
   // todo: update the combo box if it has been set externaly?
-  mExpressionContext.setFeature( feature );
+
+  // force the feature to be valid, so it can evaluate an invalid feature but having its fields set
+  if ( !feature.isValid() )
+  {
+    QgsFeature validFeature( feature );
+    validFeature.setValid( true );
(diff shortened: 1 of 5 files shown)


nyalldawson (approved):

src/gui/qgsexpressionpreviewwidget.cpp
```diff
@@ -48,7 +48,18 @@ void QgsExpressionPreviewWidget::setExpressionText( const QString &expression )
...
 {
   // todo: update the combo box if it has been set externaly?
-  mExpressionContext.setFeature( feature );
+
+  // force the feature to be valid, so it can evaluate an invalid feature but having its fields set
+  if ( !feature.isValid() )
+  {
+    QgsFeature validFeature( feature );
+    validFeature.setValid( true );
+    mExpressionContext.setFeature( validFeature );
```
m-kuhn, line 57: Should this also show a warning that the preview is not reliable because no feature is available?

nyalldawson: I think we should also totally hide the picker in this situation

3nids: That's a good idea. I'll do that

PR #37142, opened 2020-06-11
[RPM] fix rpm build by replacing NEWS with NEWS.md in spec file

basilrabi:
## Description

RPM building fails due to removal of `NEWS` in dcd017a655f9583bb759c7d00cdc164de0a0dfa7. Removing `NEWS` in the spec template fixes this.


nyalldawson:
It's not possible/worth it to use the new NEWS.md file instead?

basilrabi:
> It's not possible/worth it to use the new NEWS.md file instead?

Sounds better. Will try to build locally first.

basilrabi:
@nyalldawson , I was able to build the rpms using NEWS.md.

PR #37143, opened 2020-06-11
Use "markers" terminology consistently in centroid fill widget

### LINKED ISSUE #37106: Harmonize terms used in the centroid fill symbol dialog
We use both point and markers while they both refer to the same item. Might be worth using one to avoid confusion. And should it be markers or marker?
image

cb681e9

nyalldawson:
Fixes #37106

### DIFF
--- a/src/ui/symbollayer/widget_centroidfill.ui
+++ b/src/ui/symbollayer/widget_centroidfill.ui
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>368</width>
+    <width>382</width>
     <height>242</height>
    </rect>
   </property>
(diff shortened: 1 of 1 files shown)


src/ui/symbollayer/widget_centroidfill.ui
```diff
@@ -15,66 +15,56 @@
...
-       </property>
-       <property name="text">
-        <string>Draw point on every part of multi-part features</string>
-       </property>
-      </widget>
-     </item>
-    </layout>
+    <widget class="QCheckBox" name="mDrawAllPartsCheckBox">
+     <property name="toolTip">
+      <string>When unchecked, a single point will be drawn on the biggest part of multi-part features</string>
```
nirvn, line 33: When unchecked, a single point -> a single marker?

nirvn (approved):

Issue #30573, opened 2019-07-06
Incorrect Turkish locale behavior (on Windows)


i-s-o:
- [x] Search through existing issue reports and gis.stackexchange.com to check whether the issue already exists
- [x] Test with a [clean new user profile](https://docs.qgis.org/testing/en/docs/user_manual/introduction/qgis_configuration.html?highlight=profile#working-with-user-profiles).
- [x] Create a light and self-contained sample dataset and project file which demonstrates the issue

**Describe the bug**
Turkish locale behavior is incorrect on Windows:
1. Sorting mixes characters with and without diacritics together.
2. Uppercase/Lowercase functions don't produce expected outcomes:
    a. upper('i') -> 'I' (instead of 'İ')
    b. lower('I') -> 'i'  (instead of 'ı')
    See: https://en.wikipedia.org/wiki/Dotted_and_dotless_I

**How to Reproduce**
[turkish-uppercase.zip](https://github.com/qgis/QGIS/files/3364315/turkish-uppercase.zip)
1. Use the attached project, data, and the custom Python function file.
2. Open the print layout, and run the atlas.
3. Observe the errors:
    a. E.g. all countries starting with "C" should be listed before the counties starting with "Ç".
    b. Capitalized version of Israel in Turkish should be "İSRAİL".

Windows: wrong placement of the currency symbol
![qgis-ltr-bin_2019-07-02_14-27-17](https://user-images.githubusercontent.com/1661549/60750093-a29f9900-9f71-11e9-85b7-aa1bdf66e536.png)

Ubuntu: correct placement of the currency symbol
![VirtualBoxVM_2019-07-02_14-25-58](https://user-images.githubusercontent.com/1661549/60750107-ca8efc80-9f71-11e9-93ab-152f8045e581.png)

Windows: wrong sort order & capitalization
![qgis-dev-bin-g7_2019-07-05_11-53-26](https://user-images.githubusercontent.com/1661549/60750125-f5795080-9f71-11e9-80a0-a751c4bfb630.png)

Ubuntu: correct sort order & capitalization
![VirtualBoxVM_2019-07-05_18-47-46](https://user-images.githubusercontent.com/1661549/60750133-104bc500-9f72-11e9-9776-118f5741ffe0.png)

**QGIS and OS versions**
QGIS version: 3.9.0-Master; code revision: accbb2d7f1
Compiled/Running against Qt: 5.11.2
Compiled/Running against GDAL/OGR: 2.4.1
[...]

PR #37147, opened 2020-06-12
When required text codecs aren't available on the system, proxy the conversion over to GDAL's API

### LINKED ISSUE #36871: encoding CP852 missing in windows instalation
**There is no possibility to set encoding CP 852 to layer in windows instalation. **
Code CP 852 is not in the list of encodings. But is available in ubuntu instalation. If I have shapefile with this encoding defined in cpg file the encoding pre-set to System. QGIS seems like ignore cpg file. There is no more possibility so set this by settings ->datasource->ignore shapefile encoding declaration in this version.

**QGIS and OS versions**
QGIS 3.10.6 on win 10 and ubuntu 18.04

nyalldawson:
Resolves missing text codecs like CP852 on windows builds. These
were previously available, but then Qt upstream dropped the ICU
library from their windows builds, and accordingly a whole bunch
of older text codecs are no longer available by default on the
windows builds.

Fixes #36871


nyalldawson:
@rouault I'd love your eyes on this one!

agiudiceandrea:
Does this fix also allow to **create** a shapefile with a CP852 encoding (or another one not available in Qt for Windows)?

src/core/qgsogrproxytextcodec.cpp
```diff
@@ -0,0 +1,98 @@
...
+{
+  static QStringList codecs;
+  static std::once_flag initialized;
+  std::call_once( initialized, [&]( )
+  {
+    // there's likely others that are supported by GDAL, but we're primarily concerned here
+    // with codecs used by the shapefile driver, and which are no longer supported on the
+    // windows Qt builds (due to removal of ICU support in windows Qt builds)
+    // see https://github.com/qgis/QGIS/issues/36871
+    for ( int i = 437; i <= 950; ++i )
```
rouault, line 91: don't know if that matters in the usage, but not all codes in that range will be valid. On Unix, the list could probably be retrieved from iconv API. On Windows, WideCharToMultiByte() API is used

nyalldawson: Any clue why Travis is segfaulting with the new test? I can't reproduce locally, and valgrind is happy...

src/core/qgsogrproxytextcodec.cpp
```diff
@@ -0,0 +1,98 @@
...
+  return result;
+}
+
+QByteArray QgsOgrProxyTextCodec::convertFromUnicode( const QChar *unicode, int length, ConverterState * ) const
+{
+  if ( !unicode )
+    return QByteArray();
+
+  const QString src = QString( unicode, length );
+  char *res = CPLRecode( src.toLocal8Bit().constData(), CPL_ENC_UTF8, mName.constData() );
```
rouault, line 50: shouldn't that rather be something like (untested) ?
```suggestion
  char *res = CPLRecode( src.toUtf8().constData(), CPL_ENC_UTF8, mName.constData() );
```

src/core/qgsogrproxytextcodec.cpp
rouault: I could reproduce it in a Docker container mimicking Travis
The issue is the one pointed at https://doc.qt.io/qt-5/qtextcodec.html#QTextCodec-1

I've confirmed the following change in TestQgsOgrUtils::textCodec() fixes the crash
```
+  // The QTextCodec should always be constructed on the heap. Qt takes ownership and will delete it when the application terminates.
+  QgsOgrProxyTextCodec* codec = new QgsOgrProxyTextCodec( "CP852" );
+  QCOMPARE( codec->toUnicode( codec->fromUnicode( "abcŐ" ) ), QStringLiteral( "abcŐ" ) );
+  QCOMPARE( codec->toUnicode( codec->fromUnicode( "" ) ), QString() );
+  // cppcheck-suppress memleak
```

That also means that qgsvectordataprovider.h must be changed to not free mProxyCodec

nyalldawson:
Let's give this a shot now...

i-s-o:
In addition to the codepage problem, the lack of ICU library has at least two other consequences in some languages/locales:
* Incorrect capitalization, e.g.: https://en.wikipedia.org/wiki/Dotted_and_dotless_I
  `tr_locale = QLocale(QLocale.Turkish)`
  `tr_locale.toUpper("Diyarbakır") # => DIYARBAKIR (wrong, should be DİYARBAKIR)`
* Incorrect collation, e.g.: https://github.com/qgis/QGIS/issues/30573

Is there any chance of compiling Qt with ICU for QGIS distributions? I will search the Web to figure out what it'll take to compile Qt but I don't think I will be to make any contributions in this area with my level of understanding on how compilers work.

i-s-o:
> lack of ICU library has at least two other consequences

@jef-n could you provide some insight on whether/how to include ICU in Qt for QGIS releases?

jef-n:
@i-s-o patience.  the next round of packages have icu support

i-s-o:
@jef-n that's great news, thanks!

When you say "next round", is it with 3.16+Qt 5.11, or with 3.18+Qt 5.15? I am just curious about the timing, no urgency: this issue only came up a year ago while helping a friend. I handled the missing CP857 by using ogr2ogr, and the string case change by installing PyICU, and writing the upper case function in Python.

jef-n:
There's only one Qt in OSGeo4W (not counting the qt4 cruft left) - currently 5.11 and next 5.15.  And all QGIS' will make use of that once the switch is done (ie. 3.10.x and 3.16 and above).

PR #37148, opened 2020-06-12
Let's try this

nirvn:
## Description

Are we pass 32-bit bad luck?


nirvn:
Azure 32 bit passed once, let's try a few more times.

nirvn:
Second success in a row.

nirvn:
Three green azure 32 bit in a row. Should we re-enable it?

PR #37164, opened 2020-06-12
FIX virtual layers with CSV layers opened with "delimitedtext" provider

### LINKED ISSUE #32259: Cannot restore virtual layer that uses delimited text layer at project load
**Describe the bug**

A virtual layer that uses a delimited text layer has a bad source after the project is loaded.

Bad virtual layer source after project load

`?layer=delimitedtext:D%3A%2Fdaten%2Fdigi%2Faccuracy%2Fmatched_trackpoints_773.csv%3Ftype%3Dcsv%26delimiter%3D%3B%26detectTypes%3Dyes%26xField%3Dlon%26yField%3Dlat%26crs%3DEPSG%3A4326%26spatialIndex%3Dno%26subsetIndex%3Dno%26watchFile%3Dno:matched_trackpoints:UTF-8&query=SELECT%20DISTINCT%20t.track_id%0D%0AFROM%20matched_trackpoints%20t&nogeometry`

Correct virtual layer source

`?layer=delimitedtext:file%3A%2F%2F%2FD%3A%2Fdaten%2Fdigibus%2Faccuracy%2Fmatched_trackpoints_773.csv%3Ftype%3Dcsv%26delimiter%3D%3B%26detectTypes%3Dyes%26xField%3Dlon%26yField%3Dlat%26crs%3DEPSG%3A4326%26spatialIndex%3Dno%26subsetIndex%3Dno%26watchFile%3Dno:matched_trackpoints:UTF-8&query=SELECT%20DISTINCT%20t.track_id%0D%0AFROM%20matched_trackpoints%20t&nogeometry`

Difference is following string after 'delimitedtext:'
`file%3A%2F%2F%2F`

**How to Reproduce**

1. Add a delimited text layer to the map
2. Create a virtual layer that uses this delimited text layer
3. Close project
4. Open project
5. Layer is bad

**QGIS and OS versions**

3.8.3 and Win10

**Additional context**

suricactus:
cWhen using a CSV layer opened via "delimitedtext" provider (e.g. the delimited text dialog) in virtual layer, everything work fine until the project is reloaded.

In that case, some weird behabiour is observed - the virtual layer becomes invalid, because the CSV file name is not prefixed with `file://` and this breaks the layer itself, the layer properties etc.

This PR makes sure that the "delimitedtext" provider is always working with paths, prefixed with `file://`, which prevents the odd behaviour of occuring and makes sure the project with virtual layers can be successfully re-opened.

Fixes #32259

### DIFF
--- a/src/providers/virtual/qgsvirtuallayerprovider.cpp
+++ b/src/providers/virtual/qgsvirtuallayerprovider.cpp
@@ -562,9 +562,11 @@ void QgsVirtualLayerProvider::updateStatistics() const
   Q_ASSERT( ! mTableName.isNull() );
 
   QString sql = QStringLiteral( "SELECT Count(*)%1 FROM %2%3" )
-                .arg( hasGeometry ? QStringLiteral( ",Min(MbrMinX(%1)),Min(MbrMinY(%1)),Max(MbrMaxX(%1)),Max(MbrMaxY(%1))" ).arg( quotedColumn( mDefinition.geometryField() ) ) : QString(),
-                      mTableName,
-                      subset );
+                .arg(
+                  hasGeometry ? QStringLiteral( ",Min(MbrMinX(%1)),Min(MbrMinY(%1)),Max(MbrMaxX(%1)),Max(MbrMaxY(%1))" ).arg( quotedColumn( mDefinition.geometryField() ) ) : QString(),
+                  mTableName,
+                  subset
+                );
   Sqlite::Query q( mSqlite.get(), sql );
   if ( q.step() == SQLITE_ROW )
   {
--- a/src/providers/virtual/qgsvirtuallayerprovider.cpp
+++ b/src/providers/virtual/qgsvirtuallayerprovider.cpp
@@ -317,6 +317,7 @@ bool QgsVirtualLayerProvider::createIt()
       provider.replace( QLatin1String( "'" ), QLatin1String( "''" ) );
       QString source = mLayers.at( i ).source;
       source.replace( QLatin1String( "'" ), QLatin1String( "''" ) );
+      // the encoding might be an empty string, which breaks the SQL query below
       QString encoding = mLayers.at( i ).encoding.isEmpty()
                          ? QStringLiteral( "System" )
                          : mLayers.at( i ).encoding;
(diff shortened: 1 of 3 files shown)


src/providers/virtual/qgsvirtuallayerprovider.cpp
```diff
@@ -317,7 +317,9 @@ bool QgsVirtualLayerProvider::createIt()
       provider.replace( QLatin1String( "'" ), QLatin1String( "''" ) );
       QString source = mLayers.at( i ).source;
       source.replace( QLatin1String( "'" ), QLatin1String( "''" ) );
-      QString encoding = mLayers.at( i ).encoding;
+      QString encoding = mLayers.at( i ).encoding.isEmpty()
+                         ? QStringLiteral( "System" )
+                         : mLayers.at( i ).encoding;
```
suricactus, line 322: When the virtual layer is broken, the encoding becomes an empty string, which causes a broken SQL below in the `QgsVLayer('%2','%4',%3)` part, as `%3` is nothing and the trailing comma is wrong (and dangerous).

src/providers/virtual/qgsvirtuallayerprovider.cpp
```diff
@@ -555,11 +557,16 @@ QgsRectangle QgsVirtualLayerProvider::extent() const
 void QgsVirtualLayerProvider::updateStatistics() const
 {
   bool hasGeometry = mDefinition.geometryWkbType() != QgsWkbTypes::NoGeometry;
-  QString subset = mSubset.isEmpty() ? QString() : " WHERE " + mSubset;
+  QString subset = mSubset.isEmpty() ? QString() : QStringLiteral( " WHERE %1" ).arg( mSubset );
+
+  Q_ASSERT( ! mTableName.isNull() );
```
suricactus, line 562: `mTableName` might be an empty string when the layer is broken, because `createIt` is never called. This ASSERT should catch situations like the one that this PR is targeting.

SGroe:
Fixes #32259

nyalldawson:
Too risky for 3.14.0 -- let's defer till after release

nyalldawson:
@suricactus this has likely lead to https://github.com/qgis/QGIS/issues/38163 . Can you investigate please?

suricactus:
@nyalldawson looking at the issue, enforcing the "file://" prefix broke the things. However, I don't have access to any windows machine to verify and test a solution. I will reach back as soon as possible.

m-kuhn:
@suricactus in your original message you noted that re-saving breaks the file:// prefix. Would it make sense to switch to re-save using the original uri?

nyalldawson:
@suricactus any chance of a fix before this week's point release? If not, I propose we revert this pr as the regression is more serious then the problem fixed here.

PR #37165, opened 2020-06-12
Try to re-enable MXE

elpaso:
I didn't change anything but it seems to be working now (it was an issue with the workflow, not with the build itself),  let's try to enable it again.


roya0045:
I always found these type of build to be finnicky, glad it seems fine once again!

elpaso:
@roya0045 the MXE build is always been rock solid reliable and robust. It's the github workflow machinery that doesn't work reliably (just as it's good old pal Travis).

PR #37168, opened 2020-06-12
[backport] Fix image size ignored in image widget

### LINKED ISSUE #33682: Wrong image preview and image preview size in form
I have problems with the integrated document viewer when an attribute form is created with the drag and drop designer.

I am working with a gpkg layer (polygon) where some attributes are paths to images/photos.
For fill in data, I use the Attributes Form -> for better structure I use the form Layout drag and drop designer. (Layout see attachment as sshot_1.png)
and for images/photos the Widget Type „Attachment“.

I have defined the image width/height in the Integrated Document Viewer to 150X150px. 
When visualizing the table in the form view, some strange things in displaying the former inserted pictures appear: The images start as a 1x1px image and grows automatically to the full size as wide as the attribute form. Completely ignoring the defined 150x150px. Until every single image reaches the full size, the system is blocked so I can only watch the images growing.

The second issue when working with this layer is following:
(This issue could not be reproduced in the attached sample– it only appeared in the original project but I suggest it could have the same origin)

When switching between different features, I would expect that:
* If an image was selected and saved before, it should be displayed.
* If no image has been selected before, the image path and the preview should be empty

What on my system appears is following:
* If an image was selected and saved before, the image grows as described in the first part.
* If no image has been selected before, the image path is empty but the image from the feature I selected before is displayed by growing.

**How to Reproduce**

1.	create a layer (.shp or .gpkg layer) with attribute fields for the photo path. Type Text, Length: 250 (Or use the sample data attached)
2.	in the layer properties choose Attribute form
3.	Use the Drag and Drop Designer and structure the form as in the sshot_1.png
Tab1 (create Category Tab1 as a Tab, foto1 as a group in a container and so on)
Define the widget type "Attachment" for each of these fields
In the Integrated Document Viewer set the Type to Image and try to set a preview size of 150x150px 
4.	Change to edit mode and digitize some objects and add on some features fotos per path.
5.	Save and stop editing
6.	in the attribute table switch to Form View and select different features.
7.	Voila - !
**QGIS and OS versions**

QGIS version | 3.4.13-Madeira | QGIS code revision | QGIS code revision
-- | -- | -- | --
Active python plugins | attributePainter; latlontools; mapswipetool_plugin; openlayers_plugin; Qgis2threejs; qgis2web; quick_map_services; SelectWithin; shapetools; db_manager; MetaSearch; processing
[problem.zip](https://github.com/qgis/QGIS/files/4039667/problem.zip)
sshot-1

elpaso:
Fixes #33682  (maybe)

Cherry-pick from master commit 72c0ae73ba6822cec66

### DIFF
--- a/src/gui/editorwidgets/qgsexternalresourcewidgetwrapper.cpp
+++ b/src/gui/editorwidgets/qgsexternalresourcewidgetwrapper.cpp
@@ -185,6 +185,14 @@ void QgsExternalResourceWidgetWrapper::initWidget( QWidget *editor )
     {
       mQgsWidget->fileWidget()->setFilter( cfg.value( QStringLiteral( "FileWidgetFilter" ) ).toString() );
     }
+    if ( cfg.contains( QStringLiteral( "DocumentViewerHeight" ) ) )
+    {
+      mQgsWidget->setDocumentViewerHeight( cfg.value( QStringLiteral( "DocumentViewerHeight" ) ).toInt( ) );
+    }
+    if ( cfg.contains( QStringLiteral( "DocumentViewerWidth" ) ) )
+    {
+      mQgsWidget->setDocumentViewerWidth( cfg.value( QStringLiteral( "DocumentViewerWidth" ) ).toInt( ) );
+    }
   }
 
   if ( mLineEdit )


m-kuhn:
How can the "maybe" be interpreted?

elpaso:
See the explanations in the original PR https://github.com/qgis/QGIS/pull/37093

m-kuhn:
Note to self:

> I don't know if this really solves the issue because I cannot reproduce it exactly but I'm sure that it solves the bug that the image sizes were completely ignored.

To be sure I interpret that correctly, this PR is scheduled to be merged after confirmation of the master fix?

elpaso:
@m-kuhn no: this is a bug fix I'm 100% sure of it. The only thing I'm not sure about is if this solves completely or partially the bug #33682.

It must be merged.

m-kuhn:
So "tentative fix", let's close, hope for the best; and reopen in case someone still bumps into it?

elpaso:
> So "tentative fix", let's close, hope for the best; and reopen in case someone still bumps into it?

Yeah, that's the best we can do, I'm pretty confident this fixes the whole issue but I haven't been to reproduce the "growing" effect described in the bug.

Anyway, the widget was completely ignoring the image size parameters, that is the part of the bug that I'm fixing, whether that was the root cause of the "growing effect" or not, I cannot tell.

m-kuhn:
Sounds good, looks safe.

m-kuhn (approved):

PR #37172, opened 2020-06-12
PostgreSQL composite PK with bigint component

### LINKED ISSUE #37126: PgSQL bigint values shown as NULL if they are part of composite PK
PostgreSQL bigint values are shown as NULL on the attribute table if the bigint column is part of a composite primary key.

As an example, with the following table and data:

```sql
CREATE TABLE tb_test_compound_pk
(
        pk1 integer,
        pk2 bigint,
        pk3 real,
        value varchar(16),
        geom geometry(Point, 4326),
        PRIMARY KEY (pk1, pk2, pk3)
);

INSERT INTO tb_test_compound_pk (pk1, pk2, pk3, value, geom) VALUES
        (1, 1, 1.0, 'test 1', ST_SetSRID(ST_Point(-47.930, -15.818), 4326)),
        (1, 2, 3.141592741, 'test 2', ST_SetSRID(ST_Point(-47.887, -15.864), 4326)),
        (2, 2, 2.718281828, 'test 3', ST_SetSRID(ST_Point(-47.902, -15.763), 4326)),
        (2, 2, 1.0, 'test 4', ST_SetSRID(ST_Point(-47.952, -15.781), 4326));
```

This is what the attribute table looks like on master qgis:

image

In such cases, the existing data cannot be updated, but new features can be added. Tables with a single bigint or bigserial field as the primary key are displayed correctly. Same goes for tables with bigint fields which are not part of a primary key.

espinafre:
Bigint PostgreSQL fields are handled correctly if they are part of a
composite primary key. As a bonus, we always pass double values as
quoted to the DBMS (but not cast to text). With this one can work with
tables with real/double primary keys without the penalty of casting
these values to text. Fixes #37126.

Also added test cases for working with tables with composite primary keys, and tables whose primary key is of a floating point type.

### DIFF
--- a/tests/src/python/test_provider_postgres.py
+++ b/tests/src/python/test_provider_postgres.py
@@ -776,20 +776,16 @@ class TestPyQgsPostgresProvider(unittest.TestCase, ProviderTestCase):
         """
         Check that tables with PKs composed of many fields of different types are correctly read and written to
         """
-        vl = QgsVectorLayer('{} sslmode=disable srid=4326 key=\'"pk1","pk2","pk3"\' table="qgis_test"."tb_test_compound_pk" (geom)'.format(self.dbconn), "test_compound", "postgres")
+        vl = QgsVectorLayer('{} sslmode=disable srid=4326 key=\'"pk1","pk2"\' table="qgis_test"."tb_test_compound_pk" (geom)'.format(self.dbconn), "test_compound", "postgres")
         self.assertTrue(vl.isValid())
 
         fields = vl.fields()
 
-        # Only 6 decimals for PostgreSQL 11.
-        f = next(vl.getFeatures(QgsFeatureRequest().setFilterExpression('pk3 = 3.14159')))
+        f = next(vl.getFeatures(QgsFeatureRequest().setFilterExpression('pk1 = 1 AND pk2 = 2')))
         # first of all: we must be able to fetch a valid feature
         self.assertTrue(f.isValid())
         self.assertEqual(f['pk1'], 1)
         self.assertEqual(f['pk2'], 2)
-
-        # Only 6 decimals for PostgreSQL 11.
-        self.assertEqual(f['pk3'], 3.14159)
         self.assertEqual(f['value'], 'test 2')
 
         # can we edit a field?
@@ -798,14 +794,10 @@ class TestPyQgsPostgresProvider(unittest.TestCase, ProviderTestCase):
         self.assertTrue(vl.commitChanges())
 
         # Did we get it right? Let's create a new QgsVectorLayer and try to read back our changes:
-        vl2 = QgsVectorLayer('{} sslmode=disable srid=4326 table="qgis_test"."tb_test_compound_pk" (geom) key=\'"pk1","pk2","pk3"\' '.format(self.dbconn), "test_compound2", "postgres")
+        vl2 = QgsVectorLayer('{} sslmode=disable srid=4326 table="qgis_test"."tb_test_compound_pk" (geom) key=\'"pk1","pk2"\' '.format(self.dbconn), "test_compound2", "postgres")
         self.assertTrue(vl2.isValid())
-        f2 = next(vl.getFeatures(QgsFeatureRequest().setFilterExpression('pk3 = 3.14159')))
+        f2 = next(vl2.getFeatures(QgsFeatureRequest().setFilterExpression('pk1 = 1 AND pk2 = 2')))
         self.assertTrue(f2.isValid())
-
-        # just making sure we have the correct feature
-        # Only 6 decimals for PostgreSQL 11.
-        self.assertEqual(f2['pk3'], 3.14159)
 
         # Then, making sure we really did change our value.
         self.assertEqual(f2['value'], 'Edited Test 2')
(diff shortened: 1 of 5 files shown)


espinafre:
Hi, @nyalldawson , all green here. I had to change the Python tests in 0143fbf because they ran correctly on my local PostgreSQL 12, but not on Travis' PostgreSQL 11.

This is because of the way PostgreSQL changed the way it outputs floating point numbers: with 11, the output was limited (by default) to 6 digits, while in 12, it outputs as many digits as necessary to reflect the underlying value. Here: 
https://www.postgresql.org/docs/12/release-12.html#id-1.11.6.8.4
they document what changed.

This makes a bit dangerous to use floating point values in QGIS as primary keys to tables.  Well, the original issue is solved at least.

Works best if the original source doesn't have more than 6 digits on a floating point PK component, or no floating point PK component at all.

tests/testdata/provider/testdata_pg_bigint_pk.sql
```diff
@@ -104,20 +104,29 @@ INSERT INTO qgis_test.provider_bigint_nonfirst_pk  (zeroth_field, key1, key2, pr
 (1, 2, 3, 4,  400, 'Honey', 'Honey', '4', TIMESTAMP '2021-05-04 13:13:14', '2021-05-04', '13:13:14', '0101000020E610000014AE47E17A5450C03333333333935340')
 ;
 
-/* -- PostgreSQL 12 or later
```
nyalldawson, line 107: please don't drop these!

nyalldawson:
> As a bonus, we always pass double values as
quoted to the DBMS (but not cast to text).

tests/testdata/provider/testdata_pg_bigint_pk.sql
espinafre: Hi, I've included these in a previous PR because I had started working with tests against partitioned tables, but since working with partitions is transparent to the user, I saw no reason to test against these. These tables are, as of now, dead code, they are not used in any of the tests.

Also, the comment was wrong: these are not for PostgreSQL 12, but also work with 11 (except that the values in the clause FROM x TO y must be in parenthesis).

Do you still want to keep it?

espinafre:
> > As a bonus, we always pass double values as
> > quoted to the DBMS (but not cast to text).
> 
> I don't understand this -- why do we need to quote them if they are being treated as numeric values?

Hi, @nyalldawson ,

Because of this:

```
jose=# select * from qgis_test.tb_test_compound_pk where pk3 = 3.14159;
 pk1 | pk2 | pk3 | value | geom 
-----+-----+-----+-------+------
(0 rows)

jose=# select * from qgis_test.tb_test_compound_pk where pk3 = '3.14159';
 pk1 | pk2 |   pk3   | value  |                        geom                        
-----+-----+---------+--------+----------------------------------------------------
   1 |   2 | 3.14159 | test 2 | 0101000020E6100000A8C64B3789F147C0EE7C3F355EBA2FC0
(1 row)
```

Both PostgreSQL 11 and 12 yield the same result (and with LANG=en_US.UTF-8 or LANG=C.UTF-8), which prevents QGIS from being able to update features that have floating point values as primary key/primary key components.

tests/src/python/test_provider_postgres.py
```diff
@@ -772,6 +772,142 @@ def testPktUpdateBigintPkNonFirst(self):
...
+
+        # can we edit a field?
+        vl.startEditing()
+        vl.changeAttributeValue(f.id(), fields.indexOf('value'), 'Edited Test 2')
+        self.assertTrue(vl.commitChanges())
+
+        # Did we get it right? Let's create a new QgsVectorLayer and try to read back our changes:
+        vl2 = QgsVectorLayer('{} sslmode=disable srid=4326 table="qgis_test"."tb_test_compound_pk" (geom) key=\'"pk1","pk2","pk3"\' '.format(self.dbconn), "test_compound2", "postgres")
+        self.assertTrue(vl2.isValid())
+        f2 = next(vl.getFeatures(QgsFeatureRequest().setFilterExpression('pk3 = 3.14159')))
```
espinafre, line 803: it should be vl2.getFeatures here, not vl

nyalldawson:
Hmm - I'm still unsure about that one. Can you move it to a separate PR for individual discussion?

espinafre:
> Hmm - I'm still unsure about that one. Can you move it to a separate PR for individual discussion?

About quoting floats? Or about deleting unused test tables? Yes, I can sure do either!

nyalldawson:
quoting floats -- no issue with deleting the unused test!

tests/testdata/provider/testdata_pg_bigint_pk.sql
espinafre: Commit 6578c98 addresses the issue of quoting floating point values.

espinafre:
Hi, @nyalldawson , all green, with the float quoting removed as requested (this will be handled in issue #37245 and its upcoming fix).

I've further illustrated the originating issue https://github.com/qgis/QGIS/issues/37126#issuecomment-644508756 with a couple of screenshots that show the before/after.

May we backport to the LTR (3.10) branch? Thank you very much!

espinafre:
ping @nyalldawson

nyalldawson (approved):

PR #37180, opened 2020-06-12
Modernize language, avoid insensitive words

nyalldawson:
Modernize language and avoid insensitive/trigger words (as far as possible -- some are coming from Qt or other external APIs)

Because anything we can do to make the QGIS codebase more culturally inclusive is a small step in the right direction.

Tagging as needs-docs, because a similar update should be made to the documentation repo.


rduivenvoorde:
@nyalldawson I agree with your goal, but also wonder if such terms are not more or less 'industry standard', but let's not argue, everything we can do to reach that goal is for the good cause.

Googling around a little, one of the other alternatives for whitelist is 'safelist' (instead 'allowlist'). That sound as a sharper difference: block/safe vs block/allow.

Others proposed allow/deny.

rduivenvoorde:
seeing https://twitter.com/natfriedman/status/1271253144442253312 now... Github thinking about   finding alternatives for “master’ (master/slave?)

nyalldawson:
@rduivenvoorde I've followed the guidance from the chromium project in the use of allowList - see https://chromium.googlesource.com/chromium/src/+/master/styleguide/inclusive_code.md#racially-neutral

PR #37181, opened 2020-06-12
Fix encoding of ellipsis in Windows hopefully #37174

### LINKED ISSUE #37174: Map canvas, right mouse button
how to reproduce the problem

1. open any project
2. right-click anywhere on the map canvas
3. special characters appear in the context menu

OSGeo4W64b win 10 QGIS 3.13 master

image

image

rduivenvoorde:
See #37174, this will hopefull fix the encoding of the ellipsis in Windows. 

I'm myself not able to see if it works untill the nightly of Windows is build.

Also: I'm not sure if

`QString( "Set Custom CRS" ).append( QChar( 0x2026 )`

instead of

`QStringLiteral( "Set Custom CRS…" )`

is the proper way to do this (or if there are other better string concat options).

(The QStringLiteral (marco?) did not accept an inline concat of string+QChar)

Feel free to ignore or edit for better ways...

Fixes #37174

### DIFF
--- a/src/gui/qgsmapcanvas.cpp
+++ b/src/gui/qgsmapcanvas.cpp
@@ -889,7 +889,7 @@ void QgsMapCanvas::showContextMenu( QgsMapMouseEvent *event )
     }
   }
   copyCoordinateMenu->addSeparator();
-  QAction *setCustomCrsAction = new QAction( QStringLiteral( "Set Custom CRS…" ), mMenu );
+  QAction *setCustomCrsAction = new QAction( tr( "Set Custom CRS…" ), mMenu );
   connect( setCustomCrsAction, &QAction::triggered, this, [ = ]
   {
     QgsProjectionSelectionDialog selector( this );


src/gui/qgsmapcanvas.cpp
```diff
@@ -889,7 +889,7 @@ void QgsMapCanvas::showContextMenu( QgsMapMouseEvent *event )
     }
   }
   copyCoordinateMenu->addSeparator();
-  QAction *setCustomCrsAction = new QAction( QStringLiteral( "Set Custom CRS…" ), mMenu );
+  QAction *setCustomCrsAction = new QAction( QString( tr( "Set Custom CRS" ) ).append( QChar( 0x2026 ) ), mMenu );
```
m-kuhn, line 892:
```suggestion
  QAction *setCustomCrsAction = new QAction( tr( "Set Custom CRS…" ) ), mMenu );
```

That should be sufficient (and make sure it can be properly handled in every translation; e.g. not sure about RTL languages).

m-kuhn (approved):
Thank you

rduivenvoorde:
Sigh, sorry for the mess...

rduivenvoorde:
squashed

nyalldawson (approved):

PR #37186, opened 2020-06-13
Fix compiler warnings found during Docker-based compilation

rouault:
Most, if not all, of them false-positive or non-critical, but still good to avoid warnings when possible.

external/mdal changes have been submitted for upstream inclusion per https://github.com/lutraconsulting/MDAL/pull/267

### DIFF
--- a/src/core/effects/qgsimageoperation.cpp
+++ b/src/core/effects/qgsimageoperation.cpp
@@ -829,6 +829,7 @@ QRect QgsImageOperation::nonTransparentImageRect( const QImage &image, QSize min
       if ( qAlpha( imgScanline[x] ) )
       {
         xmin = x;
+        break;
       }
     }
   }
@@ -842,6 +843,7 @@ QRect QgsImageOperation::nonTransparentImageRect( const QImage &image, QSize min
       if ( qAlpha( imgScanline[x] ) )
       {
         xmax = x;
+        break;
       }
     }
   }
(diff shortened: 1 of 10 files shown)


src/core/effects/qgsimageoperation.cpp
```diff
@@ -829,6 +829,7 @@ QRect QgsImageOperation::nonTransparentImageRect( const QImage &image, QSize min
       if ( qAlpha( imgScanline[x] ) )
       {
         xmin = x;
+        break;
```
nyalldawson, line 832: Isn't this whole block wrong? Reading it, the X iteration should be on the outside, right?

rouault: I'm seeing this code for the first time, but I believe it is correct (it would have worked OK even before the break addition). I guess there would be issues later in the function if the image was fully transparent, but perhaps it cannot happen when this function is called

nyalldawson: Ah, I incorrectly read it as breaking out of the outer loop, sorry.

I always see the function as a huge culprit when profiling UI slowdown in large projects. Have you got any clever ideas to speed it up?

rouault: nothing strikes me at particularly inefficient. I can't think of a smarter algorithm. I can see that on a fully transparent image a full scan of the image will be done, but that's expected. I guess  breaking in with a debugger would be needed to have a better idea of what might be costly.
qAlpha(x) does (x >> 24). If one assumes that a transparent pixel has also r,g,b to 0 in most cases, then one could do ( x && qAlpha(x) ) to avoid doing the bitshift in most cases, but I'm not sure that would improve things. Might perhaps degrade things...

nyalldawson: Ok - I guess the real solution is to avoid calling this function in the first place

nyalldawson (approved):

PR #37195, opened 2020-06-14
fix bad alloc when mesh's arrows of 3D rendering has too small spacing

### LINKED ISSUE #37184: Memory bad allocation when trying to render mesh vectors in 3d
Here is the stack trace:
```
ERROR: Status 4: Requested index is not valid: -1
ERROR: Status 4: Requested index is not valid: -1
../src/core/qgis.cpp:90 : (qgsMalloc) [31194ms] [thread:0x7fffac02f280] Negative or zero size 0.
../src/core/raster/qgsrasterblock.cpp:78 : (reset) [0ms] [thread:0x7fffac02f280] Couldn't allocate data memory of 0 bytes
../src/core/raster/qgsrasterblock.cpp:287 : (setIsNoData) [0ms] [thread:0x7fffac02f280] Image not allocated
need for update
../src/core/raster/qgssinglebandpseudocolorrenderer.cpp:216 : (block) [657ms] [thread:0x7fffac02f280] No raster data!
../src/core/qgsapplication.cpp:447 : (notify) [13335ms] Caught unhandled std::exception: std::bad_alloc
[Thread 0x7fff588d8700 (LWP 2191821) exited]
[Thread 0x7fffb77fe700 (LWP 2191765) exited]
[Thread 0x7fffb7fff700 (LWP 2191764) exited]

```

image

vcloarec:
fix #37184 
When the arrow grids for 3D rendering has a spacing too small, the vector containing the grid is too big and that can lead to a bad alloc. This too small resolution appear when using a project that have been created before the new feature to render arrows with mesh in 3D view (3.14). 
This PR avoid this bad alloc with a try/catch and prevent to have such too small resolution.

### DIFF
--- a/src/core/mesh/qgsmeshlayerutils.cpp
+++ b/src/core/mesh/qgsmeshlayerutils.cpp
@@ -136,7 +136,7 @@ QVector<QgsVector> QgsMeshLayerUtils::griddedVectorValues( const QgsMeshLayer *m
   }
   catch ( ... )
   {
-    QgsDebugMsgLevel( "Unable to store the arrow grid in memmory", 1 );
+    QgsDebugMsgLevel( "Unable to store the arrow grid in memory", 1 );
     return QVector<QgsVector>();
   }
 
--- a/src/core/mesh/qgsmeshlayerutils.cpp
+++ b/src/core/mesh/qgsmeshlayerutils.cpp
@@ -22,6 +22,7 @@
 #include "qgsmeshlayerutils.h"
 #include "qgsmeshtimesettings.h"
 #include "qgstriangularmesh.h"
+#include "qgslogger.h"
 #include "qgsmeshdataprovider.h"
 #include "qgsmesh3daveraging.h"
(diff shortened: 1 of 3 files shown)


src/core/mesh/qgsmeshlayerutils.cpp
```diff
@@ -129,7 +130,16 @@ QVector<QgsVector> QgsMeshLayerUtils::griddedVectorValues( const QgsMeshLayer *m
...
     return vectors;
 
-  vectors.reserve( size.height()*size.width() );
+  try
+  {
+    vectors.reserve( size.height()*size.width() );
+  }
+  catch ( ... )
+  {
+    QgsDebugMsgLevel( "Unable to store the arrow grid in memmory", 1 );
```
DelazJ, line 139:
```suggestion
    QgsDebugMsgLevel( "Unable to store the arrow grid in memory", 1 );
```

PeterPetrik (approved):

PR #37196, opened 2020-06-14
emit mesh dataset group changed after it effectively changed

vcloarec:
The mesh layer emits signal when active dataset group changed. Before this PR, the signal was emitted just before it effectively changed and outside layers method. 
That's could lead to unwanted behavior, especially with plugins that listen to this change.
This PR fix this issue.


PeterPetrik (approved):

PeterPetrik:
unrelated win build error

PR #37199, opened 2020-06-14
Fix state handling in the layer properties WMS-T options 

Samweli:
The WMS-T static temporal options in the layer properties should be disabled when the temporal properties are in play, the changes in this PR will help notify user that the temporal controller is in control of the WMS-T layer, user will have to deactivate layer temporal properties in order to use the static temporal options.

Screenshot 
wmst_state_update

PR #37200, opened 2020-06-14
Disable project time usage in WMS-T if the project temporal range is invalid

Samweli:
Includes updates from https://github.com/qgis/QGIS/pull/37199,
The static temporal options enables usage of project temporal range in setting layer static temporal range, when the project temporal range is invalid the option to use the project temporal range should be disabled, and notify user to update the project temporal range first before using it in the WMS-T option.

Screenshot
update_project_time_to_use_it


nyalldawson:
Looks good, but there's changes needed for #37199

PR #37203, opened 2020-06-14
Make PostgreSQL use VALUES in IN clauses.

### LINKED ISSUE #37088: Unable to process treatment with large number of selected features
### Describe the bug
Working on a layer with more than 120 000 selected features.
When I'm trying to save these selected features in a new layer, or to process a buffer treatment only on these selected features, it's always result with an empty layer.

_**Note that working with full layer's features (more than 150 000) everything is fine.
So it's seems that the bug come from the way QGIS is dealing with a huge features selection.**_

### How to Reproduce

1. Load two layers from a postgres Database, both are MultiPolygon and in same CRS. One containing thousands of buildings on other containing some administrative boundaries.
2. Apply a filter in QGIS on these two layers
3. Make a select by location to get buidlings intersect and within administrative boundaries
4. Make a buffer on buildings layer with "only selected features" or Save selected features in a new layer.
5. Every time result is an empty layer.

### QGIS and OS versions

- Tried with : 

Version de QGIS : 3.4.13-Madeira
Compilé avec Qt : 5.11.2
Utilisant Qt : 5.11.2
Compilé avec GDAL/OGR : 2.4.1
Utilisé avec GDAL/OGR : 2.4.1
Compilé avec GEOS : 3.8.0-CAPI-1.13.1
Utilisé avec GEOS : 3.8.0-CAPI-1.13.1 
Version du client PostgreSQL : 11.5
Version de SpatiaLite : 4.3.0
Version de QWT : 6.1.3
Version de QScintilla2 : 2.10.8
Compilé avec PROJ : 5.2.0
Fonctionne avec PROJ : Rel. 5.2.0, September 15th, 2018
Extensions Python actives : QuickOSM; db_manager; processing

- And also with : 

Version de QGIS : 3.10.2-A Coruña
Compilé avec Qt : 5.11.2
Utilisant Qt : 5.11.2
Compilé avec GDAL/OGR : 3.0.3
Utilisé avec GDAL/OGR : 3.0.3
Compilé avec GEOS : 3.8.0-CAPI-1.13.1
Utilisé avec GEOS : 3.8.0-CAPI-1.13.1 
Compiled against SQLite : 3.29.0
Running against SQLite : 3.29.0
Version du client PostgreSQL : 11.5
Version de SpatiaLite : 4.3.0
Version de QWT : 6.1.3
Version de QScintilla2 : 2.10.8
Compilé avec PROJ : 6.3.0
Fonctionne avec PROJ : Rel. 6.3.0, January 1st, 2020
OS Version : Windows 10 (10.0)
Extensions Python actives : QuickOSM; db_manager; processing

espinafre:
Make PostgreSQL use VALUES in IN clauses, which is supposed to have
better performance when updating/deleting large datasets.

Fixes #37088.

Replicating my comment on the originating issue report:

Okay, while researching this issue, I've found the following:

https://dba.stackexchange.com/a/91539

In summary, IN clauses with raw values, such as those currently generated by qgis:

```sql
(...) WHERE pk_field IN ( 21, 42, 84 )
```

are much slower than using the following construct:

```sql
(...) WHERE pk_field IN ( VALUES (21), (42), (84) )
```

For small datasets (that is, when changing few features), the performance difference isn't detectable by the user, but for larger datasets, it could have a big impact.

Here: https://stackoverflow.com/a/42273163 there is a comparison of execution times and query plans between using IN () with a list of values and =ANY (VALUES (...), ...) clause; the latter is much faster. This other answer: https://stackoverflow.com/a/30263897 shows    that IN (VALUES) is equivalent to =ANY (VALUES) in query plan and execution time; the difference is syntactic only.

### DIFF
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -639,15 +639,17 @@ QString QgsPostgresUtils::whereClause( const QgsFeatureIds &featureIds, const Qg
     {
       QString expr;
 
-      //simple primary key, so prefer to use an "IN (...)" query. These are much faster then multiple chained ...OR... clauses
+      // simple primary key, so prefer to use an "IN (...)" query. These are much faster then multiple chained ...OR... clauses
+      // The VALUES clause causes PostgreSQL to generate a virtual table against which the original table is joined, allowing
+      // the DBMS to run faster query plans.
       if ( !featureIds.isEmpty() )
       {
         QString delim;
-        expr = QStringLiteral( "%1 IN (" ).arg( ( pkType == PktOid ? QStringLiteral( "oid" ) : QgsPostgresConn::quotedIdentifier( fields.at( pkAttrs[0] ).name() ) ) );
+        expr = QStringLiteral( "%1 IN ( VALUES " ).arg( ( pkType == PktOid ? QStringLiteral( "oid" ) : QgsPostgresConn::quotedIdentifier( fields.at( pkAttrs[0] ).name() ) ) );
 
         for ( const QgsFeatureId featureId : qgis::as_const( featureIds ) )
         {
-          expr += delim + FID_TO_STRING( ( pkType == PktOid ? featureId : FID2PKINT( featureId ) ) );
+          expr += delim + QStringLiteral( "(%1)" ).arg( FID_TO_STRING( ( pkType == PktOid ? featureId : FID2PKINT( featureId ) ) ) );
           delim = ',';
         }
         expr += ')';
@@ -660,11 +662,13 @@ QString QgsPostgresUtils::whereClause( const QgsFeatureIds &featureIds, const Qg
     {
       QString expr;
 
-      //simple primary key, so prefer to use an "IN (...)" query. These are much faster then multiple chained ...OR... clauses
+      // simple primary key, so prefer to use an "IN (...)" query. These are much faster then multiple chained ...OR... clauses
+      // The VALUES clause causes PostgreSQL to generate a virtual table against which the original table is joined, allowing
+      // the DBMS to run faster query plans.
       if ( !featureIds.isEmpty() )
       {
         QString delim;
-        expr = QStringLiteral( "%1 IN (" ).arg( QgsPostgresConn::quotedIdentifier( fields.at( pkAttrs[0] ).name() ) );
+        expr = QStringLiteral( "%1 IN ( VALUES " ).arg( QgsPostgresConn::quotedIdentifier( fields.at( pkAttrs[0] ).name() ) );
 
         for ( const QgsFeatureId featureId : qgis::as_const( featureIds ) )
         {
@@ -672,7 +676,7 @@ QString QgsPostgresUtils::whereClause( const QgsFeatureIds &featureIds, const Qg
           if ( !pkVals.isEmpty() )
           {
             QgsField fld = fields.at( pkAttrs[0] );
-            expr += delim + pkVals[0].toString();
+            expr += delim + QStringLiteral( "(%1)" ).arg( pkVals[0].toString() );
             delim = ',';
           }
         }
--- a/tests/src/providers/testqgspostgresprovider.cpp
+++ b/tests/src/providers/testqgspostgresprovider.cpp
@@ -38,6 +38,7 @@ class TestQgsPostgresProvider: public QObject
     void decodeJsonbMap();
     void testDecodeDateTimes();
     void testQuotedValueBigInt();
+    void testWhereClauseWithInValues();
 };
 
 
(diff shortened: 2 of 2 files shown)


espinafre:
Hi, @nyalldawson , all green here, except for Azure (which seems to be failing for everyone). Should I make a backport to the 3.10 branch?

nyalldawson:
I believe this change will actually make #37088 worse, because the created SQL string is longer in all cases. So we'd hit the character limit with a smaller feature selection after this change.

It may be a win for performance, but it definitely doesn't fix #37088!

nyalldawson:
Let's target this for 3.14.1 -- it's too risky a change to make at this stage of the release cycle

espinafre:
Okay, tomorrow I'll test it with a large database that I have at work. Let's see how it fares.

uclaros:
Doesn't this optimization only apply to non indexed fields? Our pk_field should be indexed anyhow.
Maybe instead of complicating the query with all those parentheses we should warn the user and try to build an index.

espinafre:
> Doesn't this optimization only apply to non indexed fields? Our pk_field should be indexed anyhow.
> Maybe instead of complicating the query with all those parentheses we should warn the user and try to build an index.

Of course you are right. TLDR: 1. PostgreSQL 12 DOES run queries with many values in the IN clause; 2. This patch is slower than what is already current in QGIS; 3. I will retract this patch.

Below is how I arrived at this decision.

I've run some tests against PostgreSQL 12.

My test table is defined as follows:

```sql
CREATE TABLE tb_large 
(
    id BIGINT PRIMARY KEY,
    nome_carta VARCHAR(32),
    geom GEOMETRY(Polygon, 4674));
CREATE INDEX ON tb_large USING GIST(geom);
```
I have a table with more than 10 million rows from which I extracted these three fields and imported into this test table. Then, I've selected 80000 random ids from that table with this:

```shell
$ psql -P tuples_only --csv -c 'select id from tb_large order by random() limit 80000;' > ids_grandes
```

Then, with some creative shell scripting, I've generated two large SQL queries with these IDs, one with the current method used by QGIS:

```sql
SELECT id, nome_carta, geom FROM tb_large WHERE id IN ( 9911446, 5360210, 1546558, 5027279, <... list of 80k values ...>);
```
This resulted in a SQL file with 697 KiB in length.

The other query, as generated by this PR, using the VALUES clause:

```sql
SELECT id, nome_carta, geom FROM tb_large WHERE id IN ( VALUES (9911446), (5360210), (1546558), (5027279), (3270372), <... list of 80k values between parenthesis ..>);
```
It resulted in a file with 854 KiB in length.

The original query consistently runs faster than this one on PostgreSQL 12:

with VALUES (this patch):

```
$ time psql -f sql_tot_values.sql > output_values.txt
real    0m3.159s
user    0m1.787s
sys     0m0.091s
```

without VALUES (currently in QGIS):

```
$ time psql -f sql_tot_bare.sql > output_bare.txt
real    0m2.436s
user    0m1.708s
sys     0m0.081s
```

In PostgreSQL 11, the times are consistently similar:

with VALUES (this patch):

```
$ time psql -p 5433 -f sql_tot_values.sql > output_values.txt
real    0m2.655s
user    0m1.731s
sys     0m0.091s
```

without VALUES (currently in QGIS):

```
$ time psql -p 5433 -f sql_tot_bare.sql > output_bare.txt
real    0m2.646s
user    0m1.713s
sys     0m0.086s
```

espinafre:
<snip>

> 
> It may be a win for performance, but it definitely doesn't fix #37088!

@nyalldawson , it is NOT a win for performance, sorry, see my testing results in this thread.

espinafre:
@nyalldawson , @uclaros , 

TLDR: 1. This patch offers no performance gain at all; 2. Both version of PostgreSQL (11 and 12) can run queries with large quantities of values in the IN clause, even if those values are of type "text".

The original reporter showed that his table's primary key is of type text (varchar(14) to be more accurate). So, I've changed my testing table to reflect this. It is now defined as:

```sql
CREATE TABLE tb_large
(
    id VARCHAR(14) PRIMARY KEY,
    idint BIGINT,
    nome_carta VARCHAR(32),
    geom GEOMETRY(Polygon, 4674));
CREATE INDEX ON tb_large USING GIST(geom);
```

The data is the same as the version describe above, that is, more than 10 million rows, with the exception that the field id is now the same as idint, only converted to text. For the queries themselves, they are just like the ones on the previous test (with integer values), except that I quote the values in all cases, since we are dealing with strings now.

Here are the execution times on PostgreSQL 12 (they are consistent across many runs):

With VALUES:
```
$ time psql -f sql_tot_values.sql > output_values.txt
real    0m2.643s
user    0m1.644s
sys     0m0.098s
```

Without VALUES:

```
$ time psql -f sql_tot_bare.sql > output_bare.txt
real    0m2.628s
user    0m1.652s
sys     0m0.073s
```

For PostgreSQL 11 (the same version as the initial reporter):

with VALUES:

```
$ time psql -p 5433 -f sql_tot_values.sql > output_values.txt
real    0m2.854s
user    0m1.645s
sys     0m0.091s
```

without VALUES:

```
$ time psql -p 5433 -f sql_tot_bare.sql > output_bare.txt
real    0m2.847s
user    0m1.672s
sys     0m0.082s
```

I've run these tests with much larger WHERE clauses too (280 000 ids in the IN clause), and the bare version run in about 1 minute, while the version with VALUES run in about 2 minutes on PostgreSQL 12.

Chunking the UPDATE may help on the QGIS side, but it seems to be innocuous on the PostgreSQL side.

Sorry to have made this mistake, I should have tested more extensively before sending this PR.

nyalldawson:
> Sorry to have made this mistake, I should have tested more extensively before sending this PR.

No need to apologise -- you've done good research here!

Shall we close this now?

espinafre:
> > Sorry to have made this mistake, I should have tested more extensively before sending this PR.
> 
> No need to apologise -- you've done good research here!
> 
> Shall we close this now?

Yes, sure. I would like to know if @uclaros is working on a fix for the original issue (#37088), or if I can start doing it myself. Thanks!

PR #37213, opened 2020-06-15
Make "show feature count" button apply to all selected indexes

### LINKED ISSUE #37194: Count of features is displayed only for the last selected layer of the layout legend
Layout legend items properties --> Layout items section
Select many layers and press the "show feature count" button
Only the last selected layer gets the count

nosumLegendItem

Actually afaics, the "move up"/"move down" as well as the "edit properties" tools behave the same: they are available when multiselection is done but only act on the last selected item.

nyalldawson:
Fixes #37194


roya0045:
Since the same behaviour also applies to move, should it also be fixed or will this cause too much trouble when moving many items with the button rather with drag & drop?

nyalldawson:
Move is definitely trickier. I figured it deserved its own ticket 😁

roya0045:
Want me to make one just to have it logged?

PR #37216, opened 2020-06-15
Fix rendering of multipart features using geometry generators

### LINKED ISSUE #23730: Geometry generator creates too many geometries for multipart input
Author Name: **Anita Graser** (@anitagraser)
Original Redmine Issue: [15810](https://issues.qgis.org/issues/15810)
Affected QGIS version: 3.4.5
Redmine category:symbology
Assignee: Nyall Dawson

---
I'm using geometry generator to buffer island polygons and style them with a dashed outline. The style looks wrong if applied to multipart features. It seems like the geometry generator computes the buffer around all features for every part of the multipart feature. So if there are three parts, all buffers are drawn three times on top of each other. 

I can work around it with a more elaborate expression, but I think that's not how it is supposed to be:

```
if (
  @geometry_part_count > 1,
  buffer( geometry_n( $geometry, @geometry_part_num), 0.5 ),
  buffer( $geometry, 0.5 )
)

```

### LINKED ISSUE #35356: Intersection (Geom. Generator) not rendering when polygons intersect in one part and abut in another
This bug is a little bit difficult to explain, but it took me the better part of a full day to figure out this was at issue in one of my own projects.

The bug seems to happen when using Geometry Generator: Intersection...when a portion of polygons do in fact intersect (A), then have a void area (B), then have a portion that abuts (C).  The actual intersect area does not render.

Screenshot from 2020-03-25 16-41-11

By moving any point from the abut portion (D), the intersect area does render (E).

Screenshot from 2020-03-25 16-51-29

**How to Reproduce**

1. Load provided testA.gpkg
2. Apply testA.qml for styles, including intersection code:

`intersection(
	geometry(
		get_feature('testA', 'id', '1')
	),
	geometry(
		get_feature('testA', 'id', '2')
	)
)`

3. Geometry Generator does NOT render intersect area.
4. Now, edit any points from abut portion and Geometry Generator does render intersect area(s).

**QGIS and OS versions**

QGIS version | 3.13.0-Master | QGIS code revision | 5a3b1ced84
-- | -- | -- | --
PostgreSQL Client Version | 10.12 (Ubuntu 10.12-0ubuntu0.18.04.1) | SpatiaLite Version | 4.3.0a
PROJ.4 Version | 493
Active python plugins | latlontools; joinmultiplelines; processing; db_manager

[testA.zip](https://github.com/qgis/QGIS/files/4383771/testA.zip)

nyalldawson:
When rendering a multipart geometry, ensure that geometry generator symbol is only rendered once, instead of once per part

Fixes #23730


nirvn:
Very useful fix.

PR #37217, opened 2020-06-15
Fix rendering of geometry generator expressions which return geometry collection results

### LINKED ISSUE #35356: Intersection (Geom. Generator) not rendering when polygons intersect in one part and abut in another

nyalldawson:
Fixes #35356

But in general, this is exciting because it allows QgsSymbol to correctly render any
input geometry types, and handle collections nicely. In future releases we could quite
easily add a symbol layer type for rendering collections, where the symbol layer has
a marker/line/fill symbol set, and the appropriate selection of these is used to
render each geometry part (it's basically identical to a geometry generator!)

Geometry collection handling at a QgsVectorLayer level, now that's a totally different beast....


nyalldawson:
Also note for reviewers: I think QgsSymbol::renderFeature is overdue for splitting up into smaller functions, but I was conscious of avoiding too much code churn in this commit. So that can come later...

nyalldawson:
Another note for reviewers:

What has changed here (the diff is more confusing then it needs to be) is:

1. instead of basing all decisions off the geometry type, we now base them off the symbol type. Previously, the different parts of this function would test the geometry type to select between handling points, lines or polygons. Now we use the actual symbol type to do this.
2. During the first stage, we dive into collections/multipart geometries with a recursive function which filters and collects only the geometry parts which are compatible with the symbol type. 
3. The remaining stages all use the filtered part lists to render and postprocess geometry parts, instead of reusing the original geometry. 

So basically, if the function gets a mixed collection geometry (as it was with the geometry generator result), the first pass collects all the parts of this collection which match the symbol type, and then we only consider these in the rendering stage.

nirvn (approved):

PR #37220, opened 2020-06-15
Correctly expose date time inputs (and a bunch of others) as possible sources for in-model expressions

### LINKED ISSUE #37219: Datetime Input Parameter in Processing Graphical Modeler not working
Not sure where exactly the problem is, anyway. The `datetime` field of the `test` layer contains some random well formatted `datetime` values.

Using the algorithm `Extract by expression` with a simple expression like:

`to_date( "datetime" ) = make_date(2020, 06, 15)`

works perfectly (1 feature extracted). Putting this in a Processing model with a `Datetime` input seems not working.

The model is very simple (and included in the project attached with the name `Datetime extraction`):

image

`enterdate` is the `Datetime` parameter but it seems not available in the Expression Builder in the Model and also when running the model:

1. Within the model settings (no `@enterdate` variable)

   image

2. During the model running (no `@enterdate` variable)

   image

in both cases also the expression `parameter('@enterdate')` is not evaluated.

If the parameter `enterdate` is a String input then the parameters is correctly exposed in the Expression Builder within the model settings and also when running it.

This is the second model with the name `String extraction` always attached with the project itself:

image

and the input parameter `enterdate_string` exposed correctly as variable in all the expression builder dialogs:

1. Within the model settings (@enterdate_string` variable)

   image

2. During the model running (`@enterdate_string` variable) used with the expression `to_date("datetime") = to_date(@enterdatestring)`:
  
   image

and the result is correctly generated

Here the attached project with data and both models saved as project models [datetime_test.zip](https://github.com/qgis/QGIS/files/4778411/datetime_test.zip)

nyalldawson:
Fixes #37219

### DIFF
--- a/src/core/processing/models/qgsprocessingmodelalgorithm.cpp
+++ b/src/core/processing/models/qgsprocessingmodelalgorithm.cpp
@@ -838,7 +838,22 @@ QMap<QString, QgsProcessingModelAlgorithm::VariableDefinition> QgsProcessingMode
       << QgsProcessingParameterExpression::typeName()
       << QgsProcessingParameterField::typeName()
       << QgsProcessingParameterString::typeName()
-      << QgsProcessingParameterAuthConfig::typeName(),
+      << QgsProcessingParameterAuthConfig::typeName()
+      << QgsProcessingParameterCrs::typeName()
+      << QgsProcessingParameterRange::typeName()
+      << QgsProcessingParameterPoint::typeName()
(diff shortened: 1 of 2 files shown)


DelazJ:
@nyalldawson wrong issue report I'm afraid ;)

nyalldawson:
Thanks, fixed

ghtmtt:
many thanks @nyalldawson for the quick fix! I was stuck in a simple GPS points extraction ;)

PR #37233, opened 2020-06-15
Add MXE badge

elpaso:


agiudiceandrea:
Hi @elpaso, it seems the badge is no longer available

image

elpaso:
You can remove it, we aren't building MXE since years.

Issue #36396, opened 2020-05-13
range widget breaks serial/nextval column


gioman:
QGIS 3.10/12/master

take a postgresql table with a column defined as "serial", if the "range" widget is applied to this column then it stops to be automatically filled with the autoincrementing value.

Worked with no issues in 2.18.

lucamanga:
As you can see, in QGIS 2.18 the **ogc_fid** field (serial type)  is correctly set as "Text Edit", as in this image:

![immagine](https://user-images.githubusercontent.com/23235179/81899577-7efaa580-95bb-11ea-97b4-a6112122d102.png)

gioman:
> As you can see, in QGIS 2.18 the **ogc_fid** field (serial type) is correctly set as "Text Edit", as in this image:

@lucamanga I'm confused, are you saying that QGIS 3 applies automatically the range widget to such columns (when a layer is loaded in a project)?

lucamanga:
the problem is that, after importing the layer from postgres (with the style saved in the DB):
- QGIS 2.18 threats the ogc_fid field as Text Edit, **correctly**
- QGIS 3 threats the ogc_fid field as Range, **incorrectly**

With range, QGIS 3 cannot set the value as `nextval(...)`

gioman:
> the problem is that, after importing the layer from postgres (with the style saved in the DB):
> 
>     * QGIS 2.18 threats the ogc_fid field as Text Edit, **correctly**
> 
>     * QGIS 3 threats the ogc_fid field as Range, **incorrectly**

if your style is set to give the ogc_fid column a range widget then is normal that you get that...
are you sure that your saved style **is not** set to give that column that widget?

lucamanga:
OK. Removing the old style and re-saving the updated style with Text Edit, **this fixes the bug.**
You can close it.
Thank you.

gioman:
Possibly it could make sense that giving (by mistake, I can't see any other reason) the range widget to a serial column should not break the autoincrement... anyway seems a very low priority issue.

m-kuhn:
Alternatively, enabling "evaluate default values server side" should also help.

lucamanga:
> Alternatively, enabling "evaluate default values server side" should also help.

where is this setting?

m-kuhn:
project properties -> data source

elpaso:
[...]

PR #37249, opened 2020-06-16
Evaluate defaults when adding new features in transaction group

### LINKED ISSUE #37222: Using "Add feature" to create feature and child feature with transaction group causes foreign key error
The QGIS documentation includes an example of creating [many-to-many relationships](https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/attribute_table.html#introducing-many-to-many-n-m-relations) with PostGIS and Geopackage databases, using transaction groups and two one-to-many relations on a junction table.

After working through the example using Geopackage in QGIS 3.12 a "foreign key error" occurs when trying to simultaneously create a new feature with a new child feature using the "Add record (Ctrl + .)" menu, see the GIF below:

add_record

This might be due to the fact that the `id` field of the new feature is not automatically filled when going through "Add record" (instead it reads "Autogenerate"). When adding a new feature using the attribute table the `id` is already filled in, and creating a new feature with a new child feature works without problems, see the GIF below:

add_new_feature

A QGIS project and Geopackage file for testing is attached:

[many_to_many.zip](https://github.com/qgis/QGIS/files/4778751/many_to_many.zip)

elpaso:
Fixes #37222

### DIFF
--- a/src/core/providers/ogr/qgsogrprovider.cpp
+++ b/src/core/providers/ogr/qgsogrprovider.cpp
@@ -1452,13 +1452,72 @@ QVariant QgsOgrProvider::defaultValue( int fieldId ) const
   else if ( defaultVal == QStringLiteral( "CURRENT_TIME" ) )
     resultVar = QTime::currentTime();
 
+  // Get next sequence value for sqlite in case we are inside a transaction
+  if ( mOgrOrigLayer &&
+       mTransaction &&
+       mDefaultValues.value( fieldId, QString() ) == tr( "Autogenerate" ) &&
+       providerProperty( EvaluateDefaultValues, false ).toBool() &&
+       ( mGDALDriverName == QLatin1String( "GPKG" ) ||
+         mGDALDriverName == QLatin1String( "SQLite" ) ) &&
+       mFirstFieldIsFid &&
+       fieldId == 0 )
+  {
+    QgsOgrLayerUniquePtr resultLayer = mOgrOrigLayer->ExecuteSQL( QByteArray( "SELECT seq FROM sqlite_sequence WHERE name = " ) +  QgsSqliteUtils::quotedValue( mOgrOrigLayer->name() ).toUtf8() );
+    if ( resultLayer )
+    {
+      gdal::ogr_feature_unique_ptr f;
+      if ( f.reset( resultLayer->GetNextFeature() ), f )
+      {
+        bool ok { true };
+        const QVariant res = QgsOgrUtils::getOgrFeatureAttribute( f.get(),
+                             fields().at( 0 ),
+                             0, textEncoding(), &ok );
+        if ( ok )
+        {
+          long long nextVal { res.toLongLong( &ok ) };
+          if ( ok )
+          {
+            // Increment
+            resultVar = ++nextVal;
+            mOgrOrigLayer->ExecuteSQLNoReturn( QByteArray( "UPDATE sqlite_sequence SET seq = seq + 1 WHERE name = " ) +  QgsSqliteUtils::quotedValue( mOgrOrigLayer->name() ).toUtf8() );
+          }
+        }
+
+        if ( ! ok )
+        {
+          QgsMessageLog::logMessage( tr( "Error retrieving next sequence value for %1" ).arg( QString( mOgrOrigLayer->name() ) ), tr( "OGR" ) );
+        }
+      }
+      else  // no sequence!
+      {
+        resultVar = 1;
+        mOgrOrigLayer->ExecuteSQLNoReturn( QByteArray( "INSERT INTO sqlite_sequence (name, seq) VALUES( " +
+                                           QgsSqliteUtils::quotedValue( mOgrOrigLayer->name() ).toUtf8() ) + ", 1)" );
+      }
+    }
+    else
+    {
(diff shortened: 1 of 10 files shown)


m-kuhn:
This is only required for sqlite based formats. Postgresql supports evaluating default values without an insert. Can we have an isolated patch?

elpaso:
@m-kuhn sure, what do you suggest to do? An ugly provider key switch?

m-kuhn:
Or fix the default value functions in the gpkg provider to have it fixed at the source?
Or let the provider report capabilities and switch based on them?

elpaso:
> Or fix the default value functions in the gpkg provider to have it fixed at the source?

I'm picking for this one, with spatialite support too.

m-kuhn:
Good choice :)

src/core/providers/ogr/qgsogrprovider.cpp
```diff
@@ -1452,13 +1452,72 @@ QVariant QgsOgrProvider::defaultValue( int fieldId ) const
   else if ( defaultVal == QStringLiteral( "CURRENT_TIME" ) )
     resultVar = QTime::currentTime();
 
+  // Get next sequence value for sqlite in case we are inside a transaction
+  if ( mOgrOrigLayer &&
+       mTransaction &&
+       mDefaultValues.value( fieldId, QString() ) == tr( "Autogenerate" ) &&
```
m-kuhn, line 1458: Is "Autogenerate" translated? Isn't that coming from sqlite itself?

src/core/providers/ogr/qgsogrprovider.cpp
```diff
@@ -1452,13 +1452,72 @@ QVariant QgsOgrProvider::defaultValue( int fieldId ) const
...
+          {
+            // Increment
+            resultVar = ++nextVal;
+            mOgrOrigLayer->ExecuteSQLNoReturn( QByteArray( "UPDATE sqlite_sequence SET seq = seq + 1 WHERE name = " ) +  QgsSqliteUtils::quotedValue( mOgrOrigLayer->name() ).toUtf8() );
+          }
+        }
+
+        if ( ! ok )
+        {
+          QgsMessageLog::logMessage( tr( "Error retrieving next sequence value for %1" ).arg( QString( mOgrOrigLayer->name() ) ), tr( "OGR" ) );
```
m-kuhn, line 1488:
```suggestion
          QgsMessageLog::logMessage( tr( "Error retrieving next sequence value for %1" ).arg( QString::fromUtf8( mOgrOrigLayer->name() ) ), tr( "OGR" ) );
```

Does the same, just explicit

src/core/qgssqliteutils.cpp
```diff
@@ -203,6 +203,49 @@ QSet<QString> QgsSqliteUtils::uniqueFields( sqlite3 *connection, const QString &
...
+      else
+      {
+        errorMessage = QObject::tr( "Error retrieving default value for %1" ).arg( tableName );
+      }
+    }
+    else // increment
+    {
+      if ( dsPtr.exec( QStringLiteral( "UPDATE sqlite_sequence SET seq = %1 WHERE name = %2" )
+                       .arg( QString::number( ++result ) )
+                       .arg( quotedTableName ), errorMessage ) != SQLITE_OK )
```
m-kuhn, line 237:
```suggestion
                       quotedTableName ), errorMessage ) != SQLITE_OK )
```

src/core/qgssqliteutils.cpp
```diff
@@ -203,6 +203,49 @@ QSet<QString> QgsSqliteUtils::uniqueFields( sqlite3 *connection, const QString &
...
+      }
+      else
+      {
+        errorMessage = QObject::tr( "Error retrieving default value for %1" ).arg( tableName );
+      }
+    }
+    else // increment
+    {
+      if ( dsPtr.exec( QStringLiteral( "UPDATE sqlite_sequence SET seq = %1 WHERE name = %2" )
+                       .arg( QString::number( ++result ) )
```
m-kuhn, line 236:
```suggestion
                       .arg( QString::number( ++result ),
```

Optimization

m-kuhn:
Looks good. I don't know the sequence handling of sqlite enough to judge it low level.

src/core/providers/ogr/qgsogrprovider.cpp
elpaso: nope

m-kuhn: fingers crossed we have a consistent translation of this magic value everywhere.

elpaso: it's the same for OGR provider, I think I homogenized spatialite provider taking OGR as the source of truth a while go ... I think it's ok to translate it because it's exposed to the users but yeah: we definitely need to pay attention to translate it all the times we refer to it in the code.

m-kuhn: I agree, let's keep it like this for now.

To make the link: https://github.com/qgis/QGIS/issues/36396

nyalldawson:
This looks potentially dangerous -- can we delay till after 3.14.0?

elpaso:
> This looks potentially dangerous -- can we delay till after 3.14.0?

OK, IMO it's not high risk but I'm not in a hurry.

PR #37251, opened 2020-06-16
Fix registerMapLayerConfigWidgetFactory docs

suricactus:
`QgsInterface::registerMapLayerConfigWidgetFactory` and `QgsInterface::registerMapLayerConfigWidget` work on any type of layer, not only vector layers.


m-kuhn:
@suricactus thanks for the update. Can you run sipify.pl on this file to update the sip files? And ideally activate the pre commit hook to avoid such issues in the future.

suricactus:
@m-kuhn or I should not commit with `-n` option.  Sorry for the noise, all should be good now.

nyalldawson (approved):

PR #37255, opened 2020-06-16
fix georeferencer sticky move gcp button

### LINKED ISSUE #37248: Georeferencer sticky Move GCP point button
**Describe the bug**
In the georeferencer, if one clicks the *Move GCP point* button and moves a GCP, then clicks *Add point*, after each new GCP is added the *Move GCP point* button gets pressed. 

**How to Reproduce**
- Open a raster in the georeferencer
- Add a few GCPs
- Click the *Move GCP point* button and move a GCP
- Click *Add point* button and add a new GCP
- The *Add point* button becomes visually un-pressed and the *Move GCP point* button becomes visually pressed. The actual action when clicking on the canvas remain that of *Add point* though.

**QGIS and OS versions**
b2da49b7 on Debian testing

**Additional context**
*Move GCP Point* actually means *Move Ground Control Point Point*

uclaros:
## Description
Store a pointer to the active map tool so the georeferencer does not stick with the move GCP tool

Fixes #37248

### DIFF
--- a/src/app/georeferencer/qgsgeorefmainwindow.cpp
+++ b/src/app/georeferencer/qgsgeorefmainwindow.cpp
@@ -401,16 +401,22 @@ void QgsGeoreferencerMainWindow::generateGDALScript()
 void QgsGeoreferencerMainWindow::setAddPointTool()
 {
   mCanvas->setMapTool( mToolAddPoint );
+  QgisApp::instance()->mapCanvas()->setMapTool( mPrevQgisMapTool );
 }
 
 void QgsGeoreferencerMainWindow::setDeletePointTool()
 {
   mCanvas->setMapTool( mToolDeletePoint );
+  QgisApp::instance()->mapCanvas()->setMapTool( mPrevQgisMapTool );
 }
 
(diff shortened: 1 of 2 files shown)


Saijin-Naib:
These fixes are so great for UX. Thank you!

uclaros:
This PR is ready for a review

src/app/georeferencer/qgsgeorefmainwindow.cpp
```diff
@@ -401,16 +401,22 @@ void QgsGeoreferencerMainWindow::generateGDALScript()
 void QgsGeoreferencerMainWindow::setAddPointTool()
 {
   mCanvas->setMapTool( mToolAddPoint );
+  QgisApp::instance()->mapCanvas()->setMapTool( mPrevQgisMapTool );
```
nyalldawson, line 404: I think we should check here that mPrevQgisMapTool is not nullptr. Otherwise I believe that on creation QgsGeoreferencerMainWindow will call setAddPointTool with mPrevQgisMapTool as null, and cause the qgis window to lose its current map tool immediately

nyalldawson (approved):

PR #37256, opened 2020-06-16
In georeferencer, allow to move and delete GCPs without perfect aiming

uclaros:
## Description
In the georeferencer, one has to aim frustratingly good in order to use the *Delete GCP* and *Move GCP point* tool.
This PR fixes this by allowing a manhattan distance of 10 between the mouse cursor and the actual GCP, similarly to the way the *vertex tool* works.

`QgsGeorefDataPoint::contains` is no longer used; should I remove it?

### DIFF
--- a/src/app/georeferencer/qgsgeorefmainwindow.cpp
+++ b/src/app/georeferencer/qgsgeorefmainwindow.cpp
@@ -519,7 +519,7 @@ void QgsGeoreferencerMainWindow::deleteDataPoint( QPoint coords )
   // This is the manhattan length from mouse cursor point to GCP
   // We will use this distance to pick the closest GCP to the cursor
   // GCPs farther than the initial minDistance are too far and are ignored
-  qreal minDistance = 10;
+  qreal minDistance = mCanvas->physicalDpiX() / 10;
   for ( QgsGCPList::iterator it = mPoints.begin(); it != mPoints.end(); ++it )
   {
     qreal distance = ( *it )->distance( coords, true );
@@ -555,7 +555,7 @@ void QgsGeoreferencerMainWindow::selectPoint( QPoint p )
   // This is the manhattan length from mouse cursor point to GCP
   // We will use this distance to pick the closest GCP to the cursor
   // GCPs farther than the initial minDistance are too far and are ignored
-  qreal minDistance = 10;
+  qreal minDistance = mCanvas->physicalDpiX() / 10;
   for ( QgsGCPList::const_iterator it = mPoints.constBegin(); it != mPoints.constEnd(); ++it )
   {
     qreal distance = ( *it )->distance( p, isMapPlugin );
--- a/src/app/georeferencer/qgsgeorefmainwindow.cpp
+++ b/src/app/georeferencer/qgsgeorefmainwindow.cpp
@@ -117,6 +117,7 @@ QgsGeoreferencerMainWindow::QgsGeoreferencerMainWindow( QWidget *parent, Qt::Win
   {
     dockThisWindow( true );
   }
+  mPointSelectionDistance = settings.value( QStringLiteral( "/Plugin-GeoReferencer/Config/PointSelectionDistance" ), "10" ).toInt();
 }
 
 void QgsGeoreferencerMainWindow::dockThisWindow( bool dock )
@@ -519,7 +520,7 @@ void QgsGeoreferencerMainWindow::deleteDataPoint( QPoint coords )
   // This is the manhattan length from mouse cursor point to GCP
   // We will use this distance to pick the closest GCP to the cursor
   // GCPs farther than the initial minDistance are too far and are ignored
-  qreal minDistance = mCanvas->physicalDpiX() / 10;
+  qreal minDistance = mPointSelectionDistance * mCanvas->devicePixelRatioF();
   for ( QgsGCPList::iterator it = mPoints.begin(); it != mPoints.end(); ++it )
   {
     qreal distance = ( *it )->distance( coords, true );
@@ -555,7 +556,7 @@ void QgsGeoreferencerMainWindow::selectPoint( QPoint p )
   // This is the manhattan length from mouse cursor point to GCP
   // We will use this distance to pick the closest GCP to the cursor
   // GCPs farther than the initial minDistance are too far and are ignored
-  qreal minDistance = mCanvas->physicalDpiX() / 10;
+  qreal minDistance = mPointSelectionDistance * mCanvas->devicePixelRatioF();
   for ( QgsGCPList::const_iterator it = mPoints.constBegin(); it != mPoints.constEnd(); ++it )
   {
     qreal distance = ( *it )->distance( p, isMapPlugin );
@@ -668,6 +669,7 @@ void QgsGeoreferencerMainWindow::showGeorefConfigDialog()
     mCanvas->refresh();
     QgisApp::instance()->mapCanvas()->refresh();
     QgsSettings s;
+    mPointSelectionDistance = s.value( QStringLiteral( "/Plugin-GeoReferencer/Config/PointSelectionDistance" ) ).toInt();
     //update dock state
     bool dock = s.value( QStringLiteral( "/Plugin-GeoReferencer/Config/ShowDocked" ) ).toBool();
     if ( dock && !mDock )
--- a/src/app/georeferencer/qgsgeorefmainwindow.cpp
+++ b/src/app/georeferencer/qgsgeorefmainwindow.cpp
@@ -515,19 +515,27 @@ void QgsGeoreferencerMainWindow::addPoint( const QgsPointXY &pixelCoords, const 
 
 void QgsGeoreferencerMainWindow::deleteDataPoint( QPoint coords )
 {
+  QgsGeorefDataPoint *pt = nullptr;
+  // This is the manhattan length from mouse cursor point to GCP
+  // We will use this distance to pick the closest GCP to the cursor
+  // GCPs farther than the initial minDistance are too far and are ignored
+  qreal minDistance = 10;
   for ( QgsGCPList::iterator it = mPoints.begin(); it != mPoints.end(); ++it )
   {
-    QgsGeorefDataPoint *pt = *it;
-    if ( /*pt->pixelCoords() == coords ||*/ pt->contains( coords, true ) ) // first operand for removing from GCP table
+    qreal distance = ( *it )->distance( coords, true );
+    if ( distance < minDistance )
     {
-      delete *it;
-      mPoints.erase( it );
-      mGCPListWidget->updateGCPList();
-
-      mCanvas->refresh();
-      break;
+      pt = *it;
+      minDistance = distance;
     }
   }
+  if ( !pt )
+    return;
+  mPoints.removeOne( pt );
+  delete pt;
+  mGCPListWidget->updateGCPList();
+
+  mCanvas->refresh();
   updateGeorefTransform();
 }
 
@@ -544,13 +552,17 @@ void QgsGeoreferencerMainWindow::selectPoint( QPoint p )
   // Get Map Sender
   bool isMapPlugin = sender() == mToolMovePoint;
   QgsGeorefDataPoint *&mvPoint = isMapPlugin ? mMovingPoint : mMovingPointQgis;
-
+  // This is the manhattan length from mouse cursor point to GCP
+  // We will use this distance to pick the closest GCP to the cursor
+  // GCPs farther than the initial minDistance are too far and are ignored
+  qreal minDistance = 10;
   for ( QgsGCPList::const_iterator it = mPoints.constBegin(); it != mPoints.constEnd(); ++it )
   {
-    if ( ( *it )->contains( p, isMapPlugin ) )
+    qreal distance = ( *it )->distance( p, isMapPlugin );
+    if ( distance < minDistance )
     {
       mvPoint = *it;
-      break;
+      minDistance = distance;
     }
   }
 }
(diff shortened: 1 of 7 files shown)


nirvn:
Nice, that's a nice PITA taken care of.

src/app/georeferencer/qgsgeorefmainwindow.cpp
```diff
@@ -505,19 +505,27 @@ void QgsGeoreferencerMainWindow::addPoint( const QgsPointXY &pixelCoords, const
 
 void QgsGeoreferencerMainWindow::deleteDataPoint( QPoint coords )
 {
+  QgsGeorefDataPoint *pt = nullptr;
+  // This is the manhattan length from mouse cursor point to GCP
+  // We will use this distance to pick the closest GCP to the cursor
+  // GCPs farther than the initial minDistance are too far and are ignored
+  qreal minDistance = 10;
```
nyalldawson, line 512: instead of the hardcoded value -- could we use QgsMapTool::searchRadiusMU/QgsMapTool::searchRadiusMM() instead?

uclaros: I guess we could, but is there a real benefit from it?
AFAICT, `searchRadiusMM` is pretty much tied to the identify tool, so the options and code comment descriptions will have to be adapted too.
I can imagine someone raising this value to identify multiple features at once, or minimizing it for better identify precision, and then struggle with the georeferencer and his strange `searchRadiusMM` value choice!

nyalldawson: Hmm - I think we need an equivalent central method for map editing tools then. My concern with a hardcoded pixel value is that it's going to be completely inappropriate for certain setups (e.g. hidpi screens). We need something which accounts for dpi also, which is why a mm based size would be preferable.

uclaros: There is also the *Search radius for vertex edits* setting (`/qgis/digitizing/search_radius_vertex_edit`) that looks more appropriate, but it can also be set for map units which makes no sense for the georeferencer case...

uclaros: @nyalldawson I think the easiest solution is to just set the hardcoded value to a larger number if 10 is too small for hidpi displays (I cannot test).
A too small value will make the tool function like it does now, needing better aiming from the user.
A slightly large value will cause no real issue, since the GCP closest to the click point will be moved, the user will still expect this GCP to be moved.
A huge value could be a little strange as the closest GCP might be very far from the click position and the user might not make the connection.

nyalldawson: I think there's a better approach :)

We should:
1. Add a method in the QgsMapTool base class for "selection tolerance pixels" and use that here
2. The base class method would use a font metric based size (e.g. the width of a "-" character, so that it correctly returns a useful tolerance for hidpi setups

uclaros: I am not sure I completely understood your suggestion. The method should return 10 for low dpi and for example 40 on hidpi?
Given that I am not familiar with hidpi, how would the above work on a mixed dual screen setup? One window on a hidpi screen, one window on a low? Or even half a window on one screen and half on another?
I think I'll re-suggest to find a ballpark value good enough for both setups (I don't have a hidpi screen) and leave your approach for another PR.

nyalldawson: > Given that I am not familiar with hidpi, how would the above work on a mixed dual screen setup? One window on a hidpi screen, one window on a low? Or even half a window on one screen and half on another?

No, you're right -- it wouldn't. To handle this situation you'd need to add the "selectionTolerancePixels" api into QgsMapCanvas, because that will be aware of the screen that it's visible on.

> I think I'll re-suggest to find a ballpark value good enough for both setups (I don't have a hidpi screen) and leave your approach for another PR.

Unfortunately you won't find one -- it'll be either too big for use on lowdpi, or too small for usability on high dpi. The only approach is to take the dpi into account. I'm aware that many of our map tools are unusable on hidpi, which is why I'd like to see us start addressing this properly moving forward...

uclaros: > Unfortunately you won't find one -- it'll be either too big for use on lowdpi, or too small for usability on high dpi

If we had let's say a value of 10 effective hidpi pixels, what would that translate to normal pixels in lowdip displays? Even at 4 times the pixels, 40 is a perfectly usable number for lowdpi displays, way better than the one or two pixels that is the equivalent with the current functionality! At least for the georeferencer tools I don't think that too big can be an issue.

m-kuhn:
```
>>> iface.mapCanvas().physicalDpiX()
210
```

Should be HiDPI

uclaros: Thanks @m-kuhn ,
my laptop's 1366x768 returns 118 and my desktop's 1920x1080 returns 72.

Would it make sense to just do `qreal minDistance = mCanvas->physicalDpiX() / 10;` ?

m-kuhn: That's what I'd propose.

Then, if possible, I would multiply with a conversion factor to convert inches to mm and make that configurable through the identify radius.

uclaros: The identify radius is pretty much tied to the identify tool and doesn't really make sense to be used here. If one wants a tighter or looser identify radius, he probably would not want that to affect vertex picking in the georeferencer.

m-kuhn: Possibly. But a common implementation for picking nodes for the node tool and gcp's here would make sense?

uclaros: Yes, that would make sense! But, the vertex tool uses for this purpose the setting `/qgis/digitizing/search_radius_vertex_edit`, which may be set in pixels but also in map units. And if set in map units it would not make sense for use with the georeferencer. That is why I think we should settle for a static value, maybe a percentage of the dpi, like in the previous comment.
Another approach would be to add a georeferencer only setting, but I know that adding new settings is frowned upon so I never proposed that.

m-kuhn: For search radius in pixels, we could convert this to reasonable values in mm `search_radius_map_tools` and reuse.

I can't actually think of a use case where the search radius in map units makes sense. Can you? If it's still required to have that for a good reason we could have two settings in parallel (one in mm, one in mu) -- or we just drop it.

uclaros: The issue is we now have `search_radius_vertex_edit` and `search_radius_vertex_edit_units` that defines if the former is in px or mu.

Truth is I cannot think of a use case for mu either, but it seems that the mu option preexisted and px was added later, plus I 'd hate to brake someones workflow just because I couldn't think of his use case!

m-kuhn: What about adding a new UnitType in QgsTolerance?

uclaros: I gave `qreal minDistance = mCanvas->physicalDpiX() / 10;` a shot and it works nice though slightly different on each of my two displays. Could you please test it on a hidpi setup?

m-kuhn: Gives around 7mm radius here

uclaros: I also get around 7mm on a 24" 1920x1080 display and around 3mm on a 13" 1366x768 display.
Both values are very usable, way better than the current 0.1mm imho! =p
What do you think?

m-kuhn: I think having involving dpi is a very good thing. And the number `10` should be configurable.

uclaros: Well, if it's configurable then there is no need for it to be dpi dependent, is it?

uclaros:
Do you think this is on the right track? Should I un-stale it or should I let it go?

m-kuhn:
I think this is on a good track. But we need the factor (10) to be configurable (visual impairment or other specific environment). Still for a good default value or multi screen setups the dpi needs to be taken into account. I.e. the config should reflect distance in mm, not pixels. Ideally this is shared with another setting as discussed earlier.

uclaros:
Giving this a last look before abandoning and noticed there could be a much simpler approach:
Right now, the gcp-clicked logic uses `QgsGeorefDataPoint::contains()` which in turn calls `mGCPSourceItem->shape().contains();` . Contrary to what I thought at first, `mGCPSourceItem->shape()` is **not** the QPainterPath used to draw the point on the canvas, but a temporary QPainterPath with the same ellipse size, used only to perform this `contains()` query.
That means we could enlarge the ellipse returned by `mGCPSourceItem->shape()` by doubling or quadrupling its radii, without affecting the displayed point size while doubling or quadrupling the click distance uniformly for low dpi and hidpi layouts.
What do you think?

nirvn:
@uclaros , I'd love to see this pushed through the finish line. Are you able to take @m-kuhn 's comments into account and adjust the PR?

uclaros:
@nirvn 
- Regarding configurability, I can add a georeferencer specific setting.
- Regarding mm instead of pixels, I am not sure how I can achieve that if not with `mCanvas.physicalDpiX() or Y`. Any input is welcome.
- Regarding reusing other setting, there are two candidate settings but none of them is appropriate as discussed.

m-kuhn:
Can you remind me what's wrong with physicalDpi?

uclaros:
One problem is it does not work as I expected, I might be misunderstanding something:
Right now, I am using `qreal minDistance = mCanvas->physicalDpiX() / 10;` expecting a distance of 0.1inch or 2.5mm but, as I mentioned in the discussion, I get 7mm on a 24" 1920x1080 display and around 3mm on a 13" 1366x768 display.

Furthermore, I came to realize that we *do not* want this distance to be in mm. We want this distance to be measured in the smallest unit of the feedback of the user's interaction, ie the smallest distance one can move the mouse and see it moving on the screen.
For normal displays that unit is the pixel. Not sure what that is for hidpi, it could be `pixel * mCanvas->devicePixelRatioF()`?

uclaros:
Update:
1. Added a georeferencer specific setting for the minimum point selection distance, with a default value of 10, minimum 4.
2. Multiplied the distance by `mCanvas->devicePixelRatioF()`, to increase the  distance proportionally for hidpi
3. Compare with the actual distance instead of the manhattan length so we have an theoretical circle instead of a rhombus.
4. Added ctrl+e shortcut to mov**e** GCP. Add is ctrl+a, delete is ctrl+d. Ctrl+m was not used because it is not a one hand shortcut.
5. Rebased

uclaros:
@nirvn or @m-kuhn could you please evaluate the latest changes on a hidpi setup?

m-kuhn:
devicePixelRatio is 1.0 here despite the HiDpi screen. This might be a local configuration issue though.
Can you give me your numbers for:

```
iface.mapCanvas().devicePixelRatioF()
1.0
iface.mapCanvas().physicalDpiX()
210
```

uclaros:
`iface.mapCanvas().devicePixelRatioF()` is `1.0` for all my screens.
`iface.mapCanvas().physicalDpiX()` is `118` for my laptop's 1366x768 13" screen and `72` for my desktop's 1920x1080 24" screen.


> I dare to challenge this statement. When a user is aiming for a point, it's his eye and hand which are the limitation. These sensors don't work in pixels, they work in real world distances. So we want the setting and default value to be in mm and the computer calculate the equivalent in pixels for us.

I do not agree. The hand and mouse sensor works in real world distances indeed, but it's all pixels once the pointer is displayed on the screen! You can see the mouse move less than a mm but you cannot see the mouse pointer move less than one pixel.
Here's a sample case with distance in mm to highlight the issue:
Let's say we wanted the default distance to be 10mm. That translates to 28 pixels on my 24" screen, not too big, not too small, nice ux.
On a 3.5" 1080p portable monitor 10mm would be like 130 pixels, allowing the mouse to move 5 times the distance on the mousepad than with the bigger screen.
Then on a 75" tv with the same resolution that would be almost 10 pixels, 3 times harder to point with the mouse than on the normal screen.

m-kuhn:
> On a 3.5" 1080p portable monitor 10mm would be like 130 pixels, allowing the mouse to move 5 times the distance on the mousepad than with the bigger screen.

Are you sure that's how things work? I.e. the higher the physical dpi the slower your mouse moves on the screen?

uclaros:
AFAIK, same distance traveled by the mouse on the mouse pad results in same pixels traveled by the pointer on the screen, regardless of screen size or resolution (granted that no mouse acceleration setting is enabled of course).
You can test this by lowering your screen resolution, then checking that you can reach the screen bounds with smaller mouse movements.
Merry X-mas!

m-kuhn:
Even if this is true, the main reason for me stays visual recognition. What I perceive as "close enogh to snap" is not related to pixels.

uclaros:
> What I perceive as "close enogh to snap" is not related to pixels.

It definitely is. If 10mm was your close enough to snap, would you be able to operate on a concert video wall where pixels are 10mm apart?

m-kuhn:
You are right,it's not perfect. But still better than px for the average everyday gis scenario I guess?

uclaros:
I am sorry, but I completely disagree. If it was in any way better than px then snapping tolerance would have been implemented in screen mm instead of px. Same for qt buttons and menus and widgets in general.

nyalldawson:
@uclaros that's not quite right -- a lot of this stuff was in place from way before hidpi was even a concept, so we can't take past decisions as correct here.

For reference, all the qt docs describe best practice as basing screen sizes on font metrics (ie real world units like mm) instead of pixels.

uclaros:
> font metrics (ie real world units like mm)

I don't think that is so. Font metrics actually mean pixels, after any scaling done by the os or the display driver. Hence, `QFontMetrics` and `QFontMetricsF` methods return values are in **pixels**

nyalldawson:
Font metrics return pixels, but they are converted from fixed values in terms of character sizes (in real world units). That's why they are used as a conversion route from "how big should things actually appear on the screen" to "what size does this correspond to in pixels"

uclaros:
Wouldn't then a `QFont font("times", 24);` need to be the same size in mm on my phone, pc and videowall regardless of screen resolution?

nyalldawson:
@uclaros you omit the font size -- ie.

    QFontMetrics fm( QFont() );

that uses the default application font sizing then -- so basically ui elements and measurements are effectively specified in "multiple of menu/standard widget height" units. The thinking there is that for a videowall the user will have already set the default font size to a suitable size, using whatever scaling factors/dpi/font size settings their particular operating system offers.

uclaros:
Maybe I'm tired, but I still do not see where real world units are. Isn't the default application font defined by the application itself, regardless of display?

PR #37257, opened 2020-06-16
Fix leak in mesh 3d renderer

nyalldawson:


vcloarec:
Thanks, I missed that one!

PR #37260, opened 2020-06-17
Tweak legend expression cache invalidation

nyalldawson:
@roya0045 this fixes the issue you describe in https://github.com/qgis/QGIS/pull/36819#issuecomment-644953883


roya0045:
@nyalldawson Thanks Nyall!

roya0045:
I'll update my branch the old way and retest at work. Thanks for the quick fix!

roya0045:
Ok it works properly. Though this make QGIS quite sluggish as opposed to my implementation. I feel like there is always something going on in the background but its not stuttering. 

I'm not sure if the invalidation is too intense or anything else.

roya0045:
Just for clarification, the loop is not too noticeable with a single layer with an expression but the delay seem to increase when there are many expressions set even if the legend item is not visible. Though it might just be my job laptop doing other stuff.

PR #37277, opened 2020-06-17
Fix spatialite wrong feature count

### LINKED ISSUE #29264: Vectorlayer from spatialite missing features and inconsistent feature count
Author Name: **Henrik Spångmyr** (@HenrikSpa)
Original Redmine Issue: [21447](https://issues.qgis.org/issues/21447)
Affected QGIS version: 3.6.0
Redmine category:data_provider/spatialite

---
QGIS show different feature count in layer list and in attribute table and some features are missing in attribute table.

I have database A with 1622 features in table obs_points.
I have a view of table obs_points, also with 1622 features.

I created a new database B and transfered all rows from A.obs_points to B.obs_points. Database B also has a view of B.obs_points.

I've attached an image which shows the issue.

There should be 1622 rows in all tables and views.

Layer obs_points_A show 1480 features in the layerlist, but 1580 in the attribute table and 1622 in spatialite_gui. << WRONG
Layer obs_points_A_view show 1565 features in the layerlist, but 1622 in the attribute table and 1622 in spatialite_gui. << WRONG
Layer obs_points_B show 1622 features in the layerlist, but 1622 in the attribute table and 1622 in spatialite_gui. << Correct
Layer obs_points_B_view show 1622 features in the layerlist, but 1622 in the attribute table and 1622 in spatialite_gui. << Correct

When I look in the attribute table for obs_points_A, the rows with rowids 1-42 (the rowid in the sqlite table) seems to be the missing rows.

Also, loading a spatialite table to QgsVectorLayer using qgis, i get vlayer.featureCount()=0 for some reason while I get lots of features from vlayer.getFeatures(). Not sure if it's related or not.

So, my transfer of rows fixed the issue, but I'm not sure why the first database isn't working well with QGIS 3.6.

---

- [featurecount_bug.png](https://issues.qgis.org/attachments/download/14497/featurecount_bug.png) (Henrik Spångmyr)
- [qgis_redmine_21447_2.sqlite](https://issues.qgis.org/attachments/download/14501/qgis_redmine_21447_2.sqlite) (Henrik Spångmyr)

elpaso:
Fixes #29264

### DIFF
--- a/src/providers/spatialite/qgsspatialiteprovider.cpp
+++ b/src/providers/spatialite/qgsspatialiteprovider.cpp
@@ -5744,7 +5744,6 @@ bool QgsSpatiaLiteProvider::getTableSummaryAbstractInterface( gaiaVectorLayerPtr
                       lyr->ExtentInfos->MaxX, lyr->ExtentInfos->MaxY );
     // This can be wrong! see: GH #29264
     // mNumberFeatures = lyr->ExtentInfos->Count;
-    
     // Note: the unique ptr here does not own the handle, it is just used for the convenience
     //       methods available within the class.
     sqlite3_database_unique_ptr slPtr;
@@ -5756,7 +5755,7 @@ bool QgsSpatiaLiteProvider::getTableSummaryAbstractInterface( gaiaVectorLayerPtr
       stmt.step();
       mNumberFeatures = sqlite3_column_int64( stmt.get(), 0 );
     }
-    // Note: the pointer handle is ownd by the provider, releasing it 
+    // Note: the pointer handle is owned by the provider, releasing it
     slPtr.release();
   }
   else
--- a/src/providers/spatialite/qgsspatialiteprovider.cpp
+++ b/src/providers/spatialite/qgsspatialiteprovider.cpp
@@ -5744,6 +5744,9 @@ bool QgsSpatiaLiteProvider::getTableSummaryAbstractInterface( gaiaVectorLayerPtr
                       lyr->ExtentInfos->MaxX, lyr->ExtentInfos->MaxY );
     // This can be wrong! see: GH #29264
     // mNumberFeatures = lyr->ExtentInfos->Count;
+    
+    // Note: the unique ptr here does not own the handle, it is just used for the convenience
+    //       methods available within the class.
     sqlite3_database_unique_ptr slPtr;
     slPtr.reset( sqliteHandle() );
     int resultCode;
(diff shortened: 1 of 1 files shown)


src/providers/spatialite/qgsspatialiteprovider.cpp
```diff
@@ -5742,7 +5742,18 @@ bool QgsSpatiaLiteProvider::getTableSummaryAbstractInterface( gaiaVectorLayerPtr
...
+    sqlite3_database_unique_ptr slPtr;
+    slPtr.reset( sqliteHandle() );
+    int resultCode;
+    sqlite3_statement_unique_ptr stmt { slPtr.prepare( QStringLiteral( "SELECT COUNT(1) FROM %2" ).arg( mQuery ), resultCode )};
+    if ( resultCode == SQLITE_OK )
+    {
+      stmt.step();
+      mNumberFeatures = sqlite3_column_int64( stmt.get(), 0 );
+    }
+    slPtr.release();
```
nyalldawson, line 5756: Doesn't this leak?

elpaso: I don't think so

nyalldawson: What takes ownership of the released pointer?

elpaso: Nothing takes it because the provider owns the pointer in the first place: https://github.com/qgis/QGIS/blob/9d040bb917407e3c936f3a04162eff2b1dcaaad1/src/providers/spatialite/qgsspatialiteprovider.cpp#L1124

nyalldawson: In that case I don't think it's suitable for storing in a unique_ptr -- why not just use a raw pointer?

elpaso: because of the convenience method `prepare` that come with `sqlite3_database_unique_ptr` and because it returns `sqlite3_statement_unique_ptr` (that is used as a real unique ptr in this code).

elpaso: I've added two comments to clarify ownership.

nyalldawson: Ok -- I think in future we should consider making mSqliteHandle a sqlite3_database_unique_ptr, but it looks like that gets complicated due to the QgsSqliteHandle class used here. ...

nyalldawson (approved):

nyalldawson:
@elpaso defer till 3.14.1?

elpaso:
@nyalldawson sorry: I merged before seeing your defer comment.

Issue #37266, opened 2020-06-17
Lookups in tables with more than 100 entries show wrong value in Relation Reference form parts


Tugark:
**Describe the bug**

**In short:** Having lookup tables with more than 100 entries which are specified as Relation Reference in Object Forms default to the first entry in the lookup table when the form is opened. This happens for existing objects, so it essentially displays an incorrect value. This only happens when the Reference attribute is declared as not allowing for NULL values.

**Detailed description:** We have lookup tables with more than 100 entries. These are defined as relations and used in forms as Relation Reference:
![image](https://user-images.githubusercontent.com/16555423/84866594-33d02880-b07a-11ea-9d73-1cef031ffea0.png)

For objects whose lookup ID is within the first 100 entries of the lookup table, everything works fine:
![image](https://user-images.githubusercontent.com/16555423/84866484-1307d300-b07a-11ea-8ae7-fc0c7086680a.png)

However, if the lookup ID is above the first 100 entries, the value is defaulted to the first entry in the lookup table when the form is opened. The selection results window shows the correct ID, however.
![image](https://user-images.githubusercontent.com/16555423/84866678-55311480-b07a-11ea-93c7-77593c20908e.png)

This bug is problematic, because if you open the form in Edit mode and save the object, the lookup ID is overwritten (in the above case, the tree with ID 2625 would turn to an Abies alba).

A temporary workaround is to allow NULL-values on the form:
![image](https://user-images.githubusercontent.com/16555423/84866809-86a9e000-b07a-11ea-84da-fbb61159c9f7.png)
 
This leads to the correct display of the ID. In that case, the above example shows the correct ID:
![image](https://user-images.githubusercontent.com/16555423/84866854-99bcb000-b07a-11ea-8bf3-2eae4fae7124.png)

**How to Reproduce**

I attached a sample geopackage, consisting of an object table with geometries (tree) and an attribute table (tree_types). tree has two objects, one with a lookup ID < 100, and one with a lookup ID > 100 (the data shown in the screenshots above).

[tree.zip](https://github.com/qgis/QGIS/files/4790931/tree.zip)

**Steps:**
1. Load both Geopackage layers
2. Define the relation in project settings as follows:
![image](https://user-images.githubusercontent.com/16555423/84867254-25364100-b07b-11ea-8792-965c474711ad.png)
3. Configure the form for the "tree" object as shown in the screenshots above (default values are fine, except the NULL configuration)
[...]

PR #37280, opened 2020-06-17
fix feature list combo box having identifier value further than fetching limit

3nids:
see discussion in #37266

I believe the issue was introduced here: https://github.com/qgis/QGIS/pull/34943/files#diff-bee409cb02c75d3272f877c88ea1e86eR277

### DIFF
--- a/src/core/qgsfeaturepickermodelbase.cpp
+++ b/src/core/qgsfeaturepickermodelbase.cpp
@@ -269,7 +269,7 @@ void QgsFeaturePickerModelBase::updateCompleter()
     const int newEntriesSize = entries.size();
 
     // fixed entry is either NULL or extra value
-    const int nbFixedEntry = ( mExtraValueDoesNotExist ? 1 : 0 ) + ( mAllowNull ? 1 : 0 );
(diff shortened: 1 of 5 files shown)


andreasneumann:
I confirm that the issue is fixed with this branch in master. Just compiled and tested.

3nids:
After discussion with @m-kuhn, we will merge this in for 3.14, but wait for 1 cycle for 3.10.
Reasoning is that data corruption is involved but the risks involved with that piece of code are not negligible.

A better fix might follow-up (avoiding the addition of the new member), but some side effects with the chain filters in the relation reference widget are expected.

Waiting for Travis for the green light, and this will be merged before the release.

nyalldawson:
> After discussion with @m-kuhn, we will merge this in for 3.14, but wait for 1 cycle for 3.10.
Reasoning is that data corruption is involved but the risks involved with that piece of code are not negligible.

:+1:

troopa81:
@3nids Thank you for dealing with this and sorry to not answer sooner, I was afk. 

The fix looks good to me, and I think it's a wise decision to wait before backporting it in 3.10.

PR #30137, opened 2019-06-08
Fix json hanlding of bools and complete json(b) PG support

elpaso:
Fixes #30131 and add tests for jsonutils

PR #37281, opened 2020-06-17
Improvements in QgsPostgresProvider::changeFeatures

espinafre:
Included support for json/jsonb in this method, as there already exists
for QgsPostgresProvider::changeAttributeValues; fixed how the provider
deals with bigint primary keys in this case; added a few tests that
exercise changing of attributes, geometries, and primary keys of
features.

### DIFF
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -3013,9 +3013,14 @@ bool QgsPostgresProvider::changeAttributeValues( const QgsChangedAttributesMap &
         if ( result.PQresultStatus() != PGRES_COMMAND_OK && result.PQresultStatus() != PGRES_TUPLES_OK )
           throw PGException( result );
       }
+      else // let the user know that no field was actually changed
+      {
+        QgsLogger::warning( tr( "No fields were updated on the database." ) );
+      }
 
       // update feature id map if key was changed
-      if ( pkChanged && mPrimaryKeyType == PktFidMap )
+      // PktInt64 also uses a fid map even if it is a stand alone field.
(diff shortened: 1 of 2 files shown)


espinafre:
Hi, @elpaso , would like to take a look at this? Since you merged #30137, this, among other things, adds json/jsonb handling in the changeFeatures() method. Also, I've tried to cover all possibilities in the test cases. Thank you!

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -3013,9 +3013,14 @@ bool QgsPostgresProvider::changeAttributeValues( const QgsChangedAttributesMap &
...
           throw PGException( result );
       }
+      else // let the user know that no field was actually changed
+      {
+        QgsLogger::warning( tr( "No fields were updated on the database." ) );
+      }
 
       // update feature id map if key was changed
-      if ( pkChanged && mPrimaryKeyType == PktFidMap )
+      // PktInt64 also uses a fid map even if it is a stand alone field.
```
elpaso, line 3022: Maybe, for clarity, you could add a boolean member like `mUseFidMap` or `mFidMapEnabled` to store this information, it would be easier than remembering the exceptional behavior of `PkInt64`.

elpaso (approved):
LGTM: great job, thank you!

src/providers/postgres/qgspostgresprovider.cpp
espinafre: I was going to implement it, but it was merged before I got around it :smile: Yes, it may be useful to have such a flag, because other data types might be handled the same way (I'm looking at real and double PostgreSQL types), but I will leave it for later.

PR #37286, opened 2020-06-18
[backport] fix feature list combo box having identifier value further than fetching limit

### LINKED ISSUE #37266: Lookups in tables with more than 100 entries show wrong value in Relation Reference form parts

3nids:
fixes #37266

backport of #37280

### DIFF
--- a/src/core/qgsfeaturefiltermodel.cpp
+++ b/src/core/qgsfeaturefiltermodel.cpp
@@ -275,7 +275,7 @@ void QgsFeatureFilterModel::updateCompleter()
     const int newEntriesSize = entries.size();
 
     // fixed entry is either NULL or extra value
-    const int nbFixedEntry = ( mExtraValueDoesNotExist ? 1 : 0 ) + ( mAllowNull ? 1 : 0 );
+    const int nbFixedEntry = ( mExtraValueDoesNotExist ? 0 : 1 ) + ( mAllowNull ? 1 : 0 );
 
     // Find the index of the current entry in the new list
     int currentEntryInNewList = -1;
--- a/src/core/qgsfeaturefiltermodel.cpp
+++ b/src/core/qgsfeaturefiltermodel.cpp
@@ -257,6 +257,7 @@ void QgsFeatureFilterModel::updateCompleter()
       setExtraValueDoesNotExist( true );
     }
 
+    mKeepCurrentEntry = true;
     mShouldReloadCurrentFeature = false;
 
     if ( mFilterValue.isEmpty() )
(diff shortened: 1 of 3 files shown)


3nids:
Since nothing bad rose about this 3.14, I am proposing to merge this backport.

3nids:
too risky / not worth the effort for 3.10

you can do `QgsSettings().setValue("maxEntriesRelationWidget", 1000, QgsSettings.Gui)` to increase the fetching limit (100 by default).

PR #37292, opened 2020-06-18
fix Qt 5.15 warning for flags initalization with nullptr

3nids:
for reference, here is the script
```
for f in $(ag -c -G '\.(h|cpp)$' '\w+::\w+ flags = nullptr' | cut -d: -f1 ); do gsed -i -r 's/(\w+::\w+) flags = nullptr/\1 flags = \1()/' $f; done
```

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -7195,7 +7195,7 @@ void QgisApp::dxfExport()
     dxfExport.setDestinationCrs( d.crs() );
     dxfExport.setForce2d( d.force2d() );
 
-    QgsDxfExport::Flags flags = nullptr;
+    QgsDxfExport::Flags flags = QgsDxfExport::Flags();
     if ( !d.useMText() )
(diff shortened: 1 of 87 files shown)


src/app/qgisapp.cpp
```diff
@@ -7195,7 +7195,7 @@ void QgisApp::dxfExport()
     dxfExport.setDestinationCrs( d.crs() );
     dxfExport.setForce2d( d.force2d() );
 
-    QgsDxfExport::Flags flags = nullptr;
+    QgsDxfExport::Flags flags = QgsDxfExport::Flags();
```
nyalldawson, line 7198: I guess this is redundant now, and 

```suggestion
    QgsDxfExport::Flags flags;
```

would have the same meaning?

nyalldawson (approved):

nyalldawson:
Let's definitely merge before release -- otherwise backporting will be painful...

PR #37295, opened 2020-06-18
Fix performance issue when splitting polygons

### LINKED ISSUE #34326: "split features" of PostGIS layers become slow since 3.6
Since version 3.6 of Qgis, the digitalization tools “split Features” , is very slow to perform the cutting (several seconds). The more polygons the cutting line crosses, the longer the cutting execution time takes.
Configuration:
Qgis 3.6 or 3.8 or 3.10
PostgreSQL 10.6 database /POSTGIS="2.5.2 r17328 "
OS Windows 10 or Opensuse 15.1
This bug did not exist on Qgis 3.4 (Windows or Linux) for which cutting is very fast.

wonder-sk:
Fixes #34326

Using test data with a HUGE polygon with several hundreds of thousands
of vertices and nearly a thousand of holes, split geometry was taking
really long time. On my laptop ~67 seconds for a simple split line.

With this fix, the time went down to ~10 seconds for the test polygon.

The issue was that split tool was doing an expensive exact intersection
calculation for each polygon part returned from polygonize, and then
calculating areas to see whether the original polygon part and intersected
part were more or less the same. I think that can be replaced by a much
simpler point in polygon test to figure out whether the particular
polygon part from polygonize is falls inside the original geometry
or whether it falls into a hole or exterior.

In my tests the time could be further improved to ~3 seconds by disabling
mergeGeometriesMultiTypeSplit() function - it is a bit of a mistery to me
what problem does it solve, however it has N^2 running time with N being
number of rings - but I did not feel adventurous enough to poke into it.
(mainly because there are no unit tests that would verify its behavior)

### DIFF
--- a/src/core/geometry/qgsgeos.cpp
+++ b/src/core/geometry/qgsgeos.cpp
@@ -896,32 +896,24 @@ QgsGeometryEngine::EngineOperationResult QgsGeos::splitPolygonGeometry( GEOSGeom
     return InvalidBaseGeometry;
   }
 
+  // we will need prepared geometry for intersection tests
+  const_cast<QgsGeos *>( this )->prepareGeometry();
+  if ( !mGeosPrepared )
+    return EngineError;
+
   //test every polygon if contained in original geometry
   //include in result if yes
   QVector<GEOSGeometry *> testedGeometries;
-  geos::unique_ptr intersectGeometry;
 
-  //ratio intersect geometry / geometry. This should be close to 1
-  //if the polygon belongs to the input geometry
-
+  // test whether the polygon parts returned from polygonize algorithm actually
+  // belong to the source polygon geometry (if the source polygon contains some holes,
+  // those would be also returned by polygonize and we need to skip them)
   for ( int i = 0; i < numberOfGeometries( polygons.get() ); i++ )
   {
     const GEOSGeometry *polygon = GEOSGetGeometryN_r( geosinit()->ctxt, polygons.get(), i );
-    intersectGeometry.reset( GEOSIntersection_r( geosinit()->ctxt, mGeos.get(), polygon ) );
-    if ( !intersectGeometry )
-    {
-      QgsDebugMsg( QStringLiteral( "intersectGeometry is nullptr" ) );
-      continue;
-    }
 
-    double intersectionArea;
-    GEOSArea_r( geosinit()->ctxt, intersectGeometry.get(), &intersectionArea );
-
-    double polygonArea;
-    GEOSArea_r( geosinit()->ctxt, polygon, &polygonArea );
-
-    const double areaRatio = intersectionArea / polygonArea;
-    if ( areaRatio > 0.99 && areaRatio < 1.01 )
+    geos::unique_ptr pointOnSurface( GEOSPointOnSurface_r( geosinit()->ctxt, polygon ) );
+    if ( pointOnSurface && GEOSPreparedIntersects_r( geosinit()->ctxt, mGeosPrepared.get(), pointOnSurface.get() ) )
       testedGeometries << GEOSGeom_clone_r( geosinit()->ctxt, polygon );
   }
 
--- a/src/core/geometry/qgsgeos.cpp
+++ b/src/core/geometry/qgsgeos.cpp
@@ -880,8 +880,13 @@ QgsGeometryEngine::EngineOperationResult QgsGeos::splitPolygonGeometry( GEOSGeom
   if ( !mGeos )
     return InvalidBaseGeometry;
 
+  // we will need prepared geometry for intersection tests
+  const_cast<QgsGeos *>( this )->prepareGeometry();
+  if ( !mGeosPrepared )
+    return EngineError;
+
   //first test if linestring intersects geometry. If not, return straight away
-  if ( !GEOSIntersects_r( geosinit()->ctxt, splitLine, mGeos.get() ) )
+  if ( !GEOSPreparedIntersects_r( geosinit()->ctxt, mGeosPrepared.get(), splitLine ) )
     return NothingHappened;
 
   //first union all the polygon rings together (to get them noded, see JTS developer guide)
@@ -896,12 +901,7 @@ QgsGeometryEngine::EngineOperationResult QgsGeos::splitPolygonGeometry( GEOSGeom
     return InvalidBaseGeometry;
   }
 
-  // we will need prepared geometry for intersection tests
-  const_cast<QgsGeos *>( this )->prepareGeometry();
-  if ( !mGeosPrepared )
-    return EngineError;
-
-  //test every polygon if contained in original geometry
+  //test every polygon is contained in original geometry
   //include in result if yes
   QVector<GEOSGeometry *> testedGeometries;
 
@@ -928,6 +928,10 @@ QgsGeometryEngine::EngineOperationResult QgsGeos::splitPolygonGeometry( GEOSGeom
     return NothingHappened;
   }
 
+  // For multi-part geometries, try to identify parts that have been unchanged and try to merge them back
+  // to a single multi-part geometry. For example, if there's a multi-polygon with three parts, but only
+  // one part is being split, this function makes sure that the other two parts will be kept in a multi-part
+  // geometry rather than being separated into two single-part geometries.
   mergeGeometriesMultiTypeSplit( testedGeometries );
 
   int i;


src/core/geometry/qgsgeos.cpp
```diff
@@ -896,32 +896,24 @@ QgsGeometryEngine::EngineOperationResult QgsGeos::splitPolygonGeometry( GEOSGeom
     return InvalidBaseGeometry;
   }
 
+  // we will need prepared geometry for intersection tests
+  const_cast<QgsGeos *>( this )->prepareGeometry();
```
nyalldawson, line 900: might be a bit of a performance win if you move this up to about line 882, so that the prepared geometry could be used for the splitline/polygon intersects test which happens on line 884.

src/core/geometry/qgsgeos.cpp
```diff
@@ -896,32 +896,24 @@ QgsGeometryEngine::EngineOperationResult QgsGeos::splitPolygonGeometry( GEOSGeom
...
+  if ( !mGeosPrepared )
+    return EngineError;
+
   //test every polygon if contained in original geometry
```
nyalldawson, line 904:
```suggestion
  //test every polygon is contained in original geometry
```

nyalldawson (approved):

nyalldawson:
Gosh -- the comments in this function are all over the place... it's very opaque!!

nyalldawson:
> In my tests the time could be further improved to ~3 seconds by disabling
mergeGeometriesMultiTypeSplit() function - it is a bit of a mistery to me
what problem does it solve

Arghhh -- I agree -- this function is completely indecipherable!! Best I can tell is that it's creating a multi* geometry type from a filtered list of the input geometries, but in an incredibly weird and completely inefficient way...

Possible the intention with in the block with the GEOSEquals test is to ensure that parts which are topologically unchanged by the split are left completely unchanged, i.e. the rings aren't reversed or densified in any way... but if that's the case, then it's broken anyway because the changed part is appended to the unionGeom list instead of the original part...

nyalldawson:
Looks like it originated here: https://github.com/qgis/QGIS/commit/4635c89077650b2111836e1bfb7cfd51427d3c25 , but no helpful clues there. 

Honestly -- I'd kill it. It's 12 year old code with no unit tests, no clear meaning, and it's causing issues. And if neither you or I can comprehend it's meaning, then we're unlikely to see anyone else step forward with the understanding required to fix/update this block.

wonder-sk:
Thanks for the review @nyalldawson 

In the end I have figured out what mergeGeometriesMultiTypeSplit() does - added a comment to it:
```
For multi-part geometries, try to identify parts that have been unchanged and try to merge them back
to a single multi-part geometry. For example, if there's a multi-polygon with three parts, but only
one part is being split, this function makes sure that the other two parts will be kept in a multi-part
geometry rather than being separated into two single-part geometries.
```
So I would keep it - it does a useful thing, just in a rather inefficient way, but that could be addressed separately...

nyalldawson:
> So I would keep it - it does a useful thing, just in a rather inefficient way, but that could be addressed separately...

Ok, makes sense. I wonder if GEOS_Equals has shortcuts in place like a bounding box check before it does the actual topological comparison.... (@dbaston?)

wonder-sk:
Unrelated PyQgsWFSProvider test failure on win 32-bit build

dbaston:
> I wonder if GEOS_Equals has shortcuts in place like a bounding box check before it does the actual topological comparison....

Yes, It does a bounding box check. Without totally understanding what the function in question is doing, you might be able to try adding a short-circuit with `GEOSEqualsExact` 

```
if ( ((GEOSEqualsExact(a, b, 0) == 1) || (GEOSEquals(a, b) == 1)) )
```

That checks "exact equality" (should be relatively cheap) instead of topological equality (relatively expensive). The `== 1` is because these functions return `2` on exception...

dbaston:
As an example, I checked the GADM boudnary of Australia. `GEOSEquals` takes 6.4 seconds and `GEOSEqualsExact` takes 6ms, around 1000x faster.

In between the case of `GEOSEqualsExact` passing and `GEOSEquals` passing is the case of `GEOSEqualsExact` passing with normalized inputs. Maybe worth considering some kind of equality testing fuction that implements these various levels of short-circuits, at least for complex geometries.

Kereboraskal:
Good morning

I am writing to confirm the problem on the split.
Working on layers with many features (be it lines or polygons), even the simplest of splits on one of them, takes much more time than what happened with previous versions of qgis (<= 3.4)
On a layer of lines we use for work it takes 2-3 sec for a simple split. I can't share this layer but I tried it on an osm lines layer and here it takes even longer. The reason seems to be the greater number of features.
I also tried on polygon layers and as the features increase, the split time increases.
It is true that as the complexity of the split increases, the time increases but it seemed proportional, therefore 1 split 2 sec, 2 simultaneous splits 4 sec.
On osm lines layer it takes 9 sec for 1 split and 18 sec for 2 simultaneous splits (no filter 4591764 features).

simple1split

simple2split

The problem is precisely the single split that to do something that previously it took virtually no time, now takes seconds. If I then filter the data, the situation gets worse and from 2 seconds it can take up to 6 seconds for a simple split (on layer we use for work).

If I filter osm lines layer with - "highway" is not null – for a simple split it takes 25 sec (3476591 features).

filter1splitIsNotNull

If I filter osm lines layer with - "highway" = 'motorway' – for a simple split it takes 6 sec (only 18471 features).

QGIS version:
3.10.10 long stable.

gioman:
>  3.10.10 long stable.

@Kereboraskal if I'm not wrong the fix for this was added only in 3.14 and above, not 3.10. Please try 3.14.

Kereboraskal:
I tried on QGIS 3.14.16-Pi and the problem is the same.

albertodeluca:
I confirm too the problem with the split tool, both on polygons and on linestrings. It started with QGIS 3.6: up to 3.4 (including 2.x), the tool was flawless. Still there in 3.15.

Just adding my two cents to the discussion:
- when the split tool is first activated on a layer with lots of features, I see an increase in network traffic from the PostGIS server where the data are stored. I assume QGIS is building a cache. That takes several seconds;
- on subsequent splits, there is no network traffic, but I see a peak in CPU usage for the QGIS app (I'm using Win 10). That usually takes at least 2-3 seconds for a single split. Same layer on QGIS 3.4 takes no time.

gioman:
> I tried on QGIS 3.14.16-Pi and the problem is the same.

@Kereboraskal please file a ticket.

Kereboraskal:
Ok. And from my calculations I found this:
time in sec; n features; nFeatures/time
1.85; 860778; 860788/1.85=465285/1
9; 4519764; 4519764/9=502169/1
Every 500000 features it takes approximately 1 sec for 1 simple split

Kereboraskal:
There was already an issue (https://github.com/qgis/QGIS/issues/34326) that has been closed and merged with this discussion.
Do I need to create another one?

gioman:
>  Do I need to create another one?

no. I'll reopen the old one.

wonder-sk:
@albertodeluca @Kereboraskal could you please provide a dataset which we could use for testing of the performance issue you are having?

gioman:
> @albertodeluca @Kereboraskal could you please provide a dataset which we could use for testing of the performance issue you are having?

@wonder-sk see this ticket https://github.com/qgis/QGIS/issues/34326#issuecomment-585328814 were there are (were?) datasets. Anyway the see the slowness (compared to 3.4) you must test on a remote server, on a local one you likely won't see any.

Doctor-Who:
@gioman on a local server, it will be slow too, I've test with local PostGIS docker instance too.
@wonder-sk if need, i can send again some sample data

albertodeluca:
Thank you @gioman Unfortunately I'm not able to grant you access to the PostGIS DB where the data are stored. But if it can be of any help, I can tell you that when I split even a 2-vertices linestring (part of a 800000+ features EPSG 32632 PostGIS table) I see network data spiking, QGIS freezes and the split operation is not completed until the data download finishes:

Capture

wonder-sk:
@Doctor-Who yes please having access to a slow dataset would be very useful - I believe the issue will be data-specific.

Looking at screenshots from @Kereboraskal, it looks like he/she is splitting a line layer... but *performance fixes on this PR only affect splits of polygon layers* - so this is likely a completely unrelated problem.

@albertodeluca hmm iteresting... Is it polygon layer or line layer that you are splitting? And if you would export your layer to a geopackage (or a shapefile) on your local drive, is it going to be still slow? If yes, it would be great if you could share such file (or its subset) - possibly privately via email or Mergin if the data is sensitive.

So to summarize - to be able to replicate and fix the issues we need:
1. some input layer
2. sample cutline which takes too long
3. approximate time the split takes in the latest QGIS release (optionally compared to timing with an older release)

albertodeluca:
@wonder-sk I'm working on a linestring layer. I cannot share it with you, but a test on an OSM roads dataset showed the same problem, even on a local geopackage: I can share that geopackage with you. It's a pretty big file though (200 MB 7zipped), because the problem does not affect small datasets. Is it ok if I post a download link here?

The cutline is not that relevant: a single split takes a few seconds on my laptop, and time increases proportionally with the number of intersections between linestring and cutline. On QGIS 3.4 it took virtually no time.

Doctor-Who:
Here a sample data ( https://cloud.sirs-fr.com/index.php/s/25cYCpoPBta5KS2 ), try to draw a line cutting several polygons and here some benchmark on this extract : 

- QGIS 3.4 it take 1-2 seconds
- QGIS 3.14, 3.16 it take nearly 10 seconds

Same action but on the entire layer, 3-4 seconds QGIS 3.4 versus 20 seconds QGIS 3.14, 3.16, etc.
Same time observed with GPKG or a local PostGreSQL even on a remote PostGreSQL (10 Gbits very efficient internal network)

wonder-sk:
@Doctor-Who thanks - could you please also post a line layer with a sample cutline feature where you see the slow performance? Just to make sure we are using the same benchmark. I have done few trials and have experienced quite reasonable response (under 1-2 seconds)

dbaston:
@Doctor-Who Are the GEOS versions the same between 3.4 and 3.14 / 3.16?

Doctor-Who:
@wonder-sk Unfortunately I'm working only on polygon layer.
@dbaston On my openSUSE I'm with 3.8.1 so same version with QGIS 3.4 or 3.14. On Windows I've try the standalone package so I've suppose it's not the same GEOS version.

saberraz:
@Doctor-Who I think what @wonder-sk meant is to provide a line showing the trace of where you split the polygon.

Doctor-Who:
@saberraz Thanks for your advice.
Here a screen recording example with 3.14 : 
Peek 05-01-2021 18-32

Same GPKG on 3.4 :
Peek 05-01-2021 18-41

albertodeluca:
Here's a linestring dataset for testing: [https://drive.google.com/file/d/167mqdxOwEB9jgJvgTw-MPSfYddI9lKZs/view?usp=sharing](url)

It's a geopackage with two layers: roads and cutline. Both have a spatial index. Using the split tool on the road layer along the only cutline feature (roughly) takes:

- 1 second on QGIS 3.4
- 125 seconds on QGIS 3.10. During those 125 seconds one CPU core was at 100%.

saberraz:
@Doctor-Who it takes around 5 seconds for me in QGIS 3.16 for the same dataset and the split line.

albertodeluca:
I've tried on a different machine (a 4 years old laptop with 16 GB of RAM and i7 processor, running win 10 64 bit). It took 135 seconds with QGIS 3.16.2. Are we talking about the same split tool (just to be sure):

image

wonder-sk:
I can replicate the issue of @albertodeluca - I have raised a new ticket for it: https://github.com/qgis/QGIS/issues/40860

It seems to be the same problem that affects also @Doctor-Who (although the slowness is less drastic)

As the issue is not relevant to this pull request, so let's stop the discussion and move it to #40860.

saberraz:
Just for future reference, I have tried the fix by  @wonder-sk  on @Doctor-Who data:
Peek 2021-01-07 11-06

Doctor-Who:
Many thanks @saberraz to make some test on our sample data.

PR #37298, opened 2020-06-18
use attribute value in the output file name for Split vector layer algorithm (fix #37296)

### LINKED ISSUE #37296: Split vector layers returns incorrect layer names
**Describe the bug**
When using the split vector layer tool the resulting layers have incorrect names, like attribute_name + index[0,1,2,3,4,5,...] but it should be attribute name + attribute value. It happens with text and numeric fields and using different output formats (gpkg, shp).

**How to Reproduce**

1. Go to Split Vector layer tool in processing
2. Split layer using Unique ID Field
3. Save layers to folder
4. Resulting layers have incorrect names

**QGIS and OS versions**

QGIS 3.13
Ubuntu 20.04

**Additional context**
Test gpkg
[split_test.zip](https://github.com/qgis/QGIS/files/4800294/split_test.zip)

alexbruy:
## Description
Split vector layer creates files with names containing field name + index, while it should create files with name containing field name + unique attribute value.

Fixes #37296. Master only.

### DIFF
--- a/src/analysis/processing/qgsalgorithmsplitvectorlayer.cpp
+++ b/src/analysis/processing/qgsalgorithmsplitvectorlayer.cpp
@@ -119,7 +119,7 @@ QVariantMap QgsSplitVectorLayerAlgorithm::processAlgorithm( const QVariantMap &p
     if ( feedback->isCanceled() )
       break;
 
-    QString fileName = QStringLiteral( "%1_%2.%3" ).arg( baseName ).arg( current ).arg( outputFormat );
+    QString fileName = QStringLiteral( "%1_%2.%3" ).arg( baseName ).arg( ( *it ).toString() ).arg( outputFormat );
     feedback->pushInfo( QObject::tr( "Creating layer: %1" ).arg( fileName ) );
 
     sink.reset( QgsProcessingUtils::createFeatureSink( fileName, context, fields, geometryType, crs ) );


uclaros:
This is a little dangerous imho, at least without some basic checking.
What if the unique attribute value contains backslashes on windows? or is a blob? or ...

nyalldawson:
I'd propose a different approach -- add an expression parameter (advanced) for the output file name, and make it default to the current @row_number style approach....

alexbruy:
@uclaros this is how it used to work from the very begginning. Most of the users are not stupid and understand what they are doing.

alexbruy:
@nyalldawson while idea of expression-based file name is interesting, I'm afraid we can not use @row_number here, as we don't operate on features sequentially instead we work on the sets of features with same field value. If I'm not wrong there is no expression which can be used here.

uclaros:
> Most of the users are not stupid and understand what they are doing.

I can't say I am a big fan of that way of thinking...

Another problem, apart from characters that will mess the filename like slashes or *, is that if there are both null and empty values, they will both be exported to the same filename, overwriting one another.

Also, a notice that output filenames will be silently overwritten would be nice.
Also, exporting to multiple geopackages with a single layer each is a little weird.

alexbruy:
Not stale, fixes regression.

alexbruy:
Not stale.

alexbruy:
Merging now to restore behaviour of the original Python tool. Then we can try to find another solution.

RNanclares:
Thanks!

PR #37307, opened 2020-06-19
Fix some more Qt deprecation warnings

nyalldawson:


src/core/labeling/qgstextlabelfeature.cpp
```diff
@@ -121,14 +125,23 @@ void QgsTextLabelFeature::calculateInfo( bool curvedLabeling, QFontMetricsF *fm,
       }
       // this workaround only works for clusters with a single character. Not sure how it should be handled
       // with multi-character clusters.
+#if QT_VERSION < QT_VERSION_CHECK(5, 13, 0)
```
rouault, line 128: shouldn't that be 5.11 (in other places of this PR too) ?

nyalldawson: Yes -- you're right!

PR #37311, opened 2020-06-19
[wms] Fix broken WMS layers from servers using relative OnlineResource paths

nirvn:
## Description

WMS servers in the wild can server OnlineResource as relative paths ( e.g. https://data.opendevelopmentcambodia.net/geoserver/ODCambodia/wms? ). QGIS' WMS provider would produce broken layers for those servers.

This PR fixes the provider by checking whether a given path is relative, and if so resolve it.

_Note: I'm not familiar with the WMS spec, I suspect relative paths might be an explicit or implicit breach of the spec. That said, it's in the wild already and it costs us nothing to be tolerant here._

### DIFF
--- a/src/providers/wms/qgswmscapabilities.cpp
+++ b/src/providers/wms/qgswmscapabilities.cpp
@@ -452,8 +452,9 @@ QDateTime QgsWmsSettings::parseWmstDateTimes( QString item )
 // ----------------------
 
 
-QgsWmsCapabilities::QgsWmsCapabilities( const QgsCoordinateTransformContext &coordinateTransformContext ):
-  mCoordinateTransformContext( coordinateTransformContext )
+QgsWmsCapabilities::QgsWmsCapabilities( const QgsCoordinateTransformContext &coordinateTransformContext, const QString &baseUrl ):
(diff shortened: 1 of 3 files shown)


src/providers/wms/qgswmscapabilities.cpp
```diff
@@ -691,8 +692,13 @@ void QgsWmsCapabilities::parseService( const QDomElement &element, QgsWmsService
 
 void QgsWmsCapabilities::parseOnlineResource( const QDomElement &element, QgsWmsOnlineResourceAttribute &onlineResourceAttribute )
 {
-
-  onlineResourceAttribute.xlinkHref = QUrl::fromEncoded( element.attribute( QStringLiteral( "xlink:href" ) ).toUtf8() ).toString();
+  QUrl url = QUrl::fromEncoded( element.attribute( QStringLiteral( "xlink:href" ) ).toUtf8() );
+  if ( url.isRelative() )
+  {
+    QUrl baseUrl = QUrl( mBaseUrl );
```
elpaso, line 698:
```suggestion
    const QUrl baseUrl = QUrl( mBaseUrl );
```

elpaso (approved):

PR #36812, opened 2020-05-28
Approved pending 3.10.9 backports 

nyalldawson:
Backports approved for the 3.10.9 release

PR #37318, opened 2020-06-19
[Backport 3.10] Fix performance issue when splitting polygons

wonder-sk:
Backport of #37295 to 3.10.x

### DIFF
--- a/src/core/geometry/qgsgeos.cpp
+++ b/src/core/geometry/qgsgeos.cpp
@@ -896,32 +896,24 @@ QgsGeometryEngine::EngineOperationResult QgsGeos::splitPolygonGeometry( GEOSGeom
     return InvalidBaseGeometry;
   }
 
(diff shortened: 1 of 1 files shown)


wonder-sk:
How does it work with the 'not for merge' label? (I haven't seen it used before)

nyalldawson:
@wonder-sk 

We're now delaying all non-critical backports (non crash, non data corruption fix) to LTR by 1 point release, so that the change has at least one month's worth of testing in a stable release prior to inclusion in LTR. It's to reduce risk of introducing regressions mid-LTR cycle.

nyalldawson (approved):

nyalldawson:
Moved to https://github.com/qgis/QGIS/pull/36812

wonder-sk:
Ah ok cool that makes sense :+1:

PR #37339, opened 2020-06-22
PG: fix "name" field length (63 instead of 64)

elpaso:


src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -1123,8 +1123,8 @@ bool QgsPostgresProvider::loadFields()
       else if ( fieldTypeName == QLatin1String( "name" ) )
       {
         fieldSubType = QVariant::String;
-        // "name" type lenght is 64 according to: SELECT typlen FROM pg_type WHERE typname = 'name';
-        fieldSize = 64;
+        // "name" type lenght is 63
```
stev-0, line 1126: typo - switch to **length**

stev-0, line 1126: typo - switch to length

pblottiere:
```suggestion
        // "name" type length is 63
```

pblottiere: @stev-0 FYI you can use GitHub "Suggestions" for this kind of comment ^.

And thanks for raising this :+1:

elpaso, line 1126:
```suggestion
```

PR #37353, opened 2020-06-23
use different names for QgsProject::layer(s)WillBeRemoved signals in Python

3nids:


m-kuhn:
Can you add some reasoning, isn't this an API break?

3nids:
you cannot have two signals with same signature in Python.
I tested, and it's the signals with layer IDs which are coming.
So, not an API break.

src/core/qgsproject.h
```diff
@@ -1631,7 +1631,7 @@ class CORE_EXPORT QgsProject : public QObject, public QgsExpressionContextGenera
      * \see layerWillBeRemoved()
      * \see layersRemoved()
      */
-    void layersWillBeRemoved( const QList<QgsMapLayer *> &layers );
+    void layersWillBeRemoved( const QList<QgsMapLayer *> &layers ) SIP_PYNAME( layersWillRemovedLayers );
```
m-kuhn, line 1634: The proposed names miss a `Be`
What about just giving them a completely different name (and adapting some api docs to guide people to these signals).

```suggestion
    void layersWillBeRemoved( const QList<QgsMapLayer *> &layers ) SIP_PYNAME( aboutToRemoveLayers );
```

3nids: in such case, we also change the cpp names, right?
Advantages:
* we can specifically write the since annotation for PyQGIS
* that avoids the overload in cpp

m-kuhn: Why not.
I'd like to check with @nyalldawson what he thinks about this API change

nyalldawson: Yeah I'm +1 to rename the cpp method too -- overloaded signals are just a pain.

3nids:
closing in favor of #37543

Issue #37359, opened 2020-06-23
Default values using @snapping_result delete attributes


m-kuhn:
Given a point layer with an attribute `name` and an attribute `time`.
Configure the default value for the field `name` using the expression below. Disable (or enable) apply on update.

Expected behavior: when I add a point on top of an existing one, it will copy the name and I'm able to fine tune the fields.
Observed behavior: when I add a point on top of an existing one, it will copy the name but as soon as I edit the time, the field `name` is cleared.

```
with_variable(
  'first_snapped_point',
  array_first( @snapping_results ),
  attribute(
    get_feature_by_id(
      map_get( @first_snapped_point, 'layer' ),
      map_get( @first_snapped_point, 'feature_id' )
    ),
    'name'
  )
)
```

m-kuhn:
This issue is composed of a couple of interleaved issues:

1. The variable `@snapping_result` is only populated to pre-evaluate default values before they are sent to the attribute form. If the attribute form later on updates / recalculates default values, map tool capture expression scope is lost.
2. The functionality to update depending fields as they are filled. This functionality is expected only to work when "apply on update" is toggled (at least that's what the warning on the layer properties -> attribute form dialog suggests). This functionality is still enabled also when apply on update is disabled.
3. Bonus issue: the dependencies on the layer are calculated based on the required fields. The expression function `attribute()` propagates that it depends on all fields. So if *any* other field changes, this field will depend on it and be recalculated.

PR #37361, opened 2020-06-23
Only update depending fields if "apply on update" is specified

m-kuhn:
References https://github.com/qgis/QGIS/issues/37359

### DIFF
--- a/src/gui/qgsattributeform.cpp
+++ b/src/gui/qgsattributeform.cpp
@@ -2483,22 +2483,26 @@ void QgsAttributeForm::updateDefaultValueDependencies()
     QgsEditorWidgetWrapper *eww = qobject_cast<QgsEditorWidgetWrapper *>( ww );
     if ( eww )
     {
-      QgsExpression exp( eww->field().defaultValueDefinition().expression() );
-      const QSet<QString> referencedColumns = exp.referencedColumns();
-      for ( const QString &referencedColumn : referencedColumns )
+      const QgsDefaultValue defaultValueDefinition = eww->field().defaultValueDefinition();
+      if ( defaultValueDefinition.applyOnUpdate() )
       {
-        if ( referencedColumn == QgsFeatureRequest::ALL_ATTRIBUTES )
+        QgsExpression exp( defaultValueDefinition.expression() );
+        const QSet<QString> referencedColumns = exp.referencedColumns();
+        for ( const QString &referencedColumn : referencedColumns )
         {
-          const QList<int> allAttributeIds( mLayer->fields().allAttributesList() );
-
-          for ( const int id : allAttributeIds )
+          if ( referencedColumn == QgsFeatureRequest::ALL_ATTRIBUTES )
           {
-            mDefaultValueDependencies.insertMulti( id, eww );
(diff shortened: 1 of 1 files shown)


m-kuhn:
@signedav can you have a look at this?

src/gui/qgsattributeform.cpp
```diff
@@ -2483,23 +2483,27 @@ void QgsAttributeForm::updateDefaultValueDependencies()
     QgsEditorWidgetWrapper *eww = qobject_cast<QgsEditorWidgetWrapper *>( ww );
     if ( eww )
     {
-      QgsExpression exp( eww->field().defaultValueDefinition().expression() );
-      const QSet<QString> referencedColumns = exp.referencedColumns();
-      for ( const QString &referencedColumn : referencedColumns )
+      const QgsDefaultValue defaultValueDefinition = eww->field().defaultValueDefinition();
+      if ( defaultValueDefinition.applyOnUpdate() )
```
signedav, line 2487: I'm not sure if this is what the user expects: What does "apply on update" mean? 1. Possibility: Apply on update of the feature (means editing and not adding it) or 2. Possibilty: on update another field (on adding or editing the feature)? 

Currently the first possibility is integrated. Means it checks this only on adding a new feature here https://github.com/qgis/QGIS/blob/master/src/gui/qgsattributeform.cpp#L510 

It's what it should do in my opinion. But if I'm wrong, it should anyway handle it on the line 510, no?

m-kuhn: > I'm not sure if this is what the user expects: What does "apply on update" mean? 1. Possibility: Apply on update of the feature (means editing and not adding it) or 2. Possibilty: on update another field (on adding or editing the feature)?

I agree this could be clearer. The original idea was clearly 1.

But, @nyalldawson had the opposite perception in the following discussion https://github.com/qgis/QGIS/pull/35030#issuecomment-598343631, so currently we have a warning on the dialog (see below) that indicates this is not supposed to be working.

image

signedav: I remember that discussion. Fair enough but wouldn't it be clearer with an option like: "Apply default value on attribute updates"? Since the values should be updated when this option is checked.

And then catch it here instead https://github.com/qgis/QGIS/blob/master/src/gui/qgsattributeform.cpp#L510

When then the "Apply default value on update" is checked we would have the issue  #37359 still, no? So we'll need a follow up fix.

m-kuhn: > but wouldn't it be clearer with an option like: "Apply default value on attribute updates"? Since the values should be updated when this option is checked.

That would be a possibility but would add a new option and therefore not be very backport friendly?
This patch aims to fix a inconsistency between the warning and the behavior. The question is if the warning is wrong or the behavior is wrong. If we change the behavior, we will get a band aid for the particular issue #37359. But I'm open to kill the warning instead as well.

> When then the "Apply default value on update" is checked we would have the issue #37359 still, no? So we'll need a follow up fix.

Correct, but this is a completely separate fix, where the snapping scope should be preserved in the attribute form.

m-kuhn:
See https://github.com/qgis/QGIS/pull/37769

PR #37367, opened 2020-06-23
Quoting floating point values on PostgreSQL provider.

### LINKED ISSUE #37245: Cannot edit attributes from PostgreSQL tables with floating point PKs
Given the following PostgreSQL/PostGIS table:

```sql
CREATE TABLE qgis_test.tb_test_float_pk
(
    pk REAL PRIMARY KEY,
    value VARCHAR(16),
    geom geometry(Point, 4326)
);
INSERT INTO qgis_test.tb_test_float_pk (pk, value, geom) VALUES
    (3.14159, 'first test', ST_SetSRID(ST_Point(-47.887, -15.864), 4326)),
    (2.71828, 'second test', ST_SetSRID(ST_Point(-47.902, -15.763), 4326));
```

The attribute table is correctly shown, but the attributes cannot be edited/saved with QGIS master:

qgis_edit_float_pk

This was discovered while preparing the test cases for #37172

espinafre:
This quotes floating point values on the PostgreSQL data provider, which
allows such values to be used either as stand-alone primary keys or as
components of composite primary keys on underlying tables. The
PostgreSQL session parameter extra_float_digits is explicitly set to 3
(the default on PostgreSQL 11 and older; in PostgreSQL 12 this defaults
to 1), to keep compatibility with all database versions and to avoid
round-trip errors.

In this PR, NUMERIC/DECIMAL values are no longer cast to
QVariant::Double in order to avoid loss of precision.

Fixes #37245 . 

This PR originated from @nyalldawson 's request to discuss this change in more depth, here: https://github.com/qgis/QGIS/pull/37172#issuecomment-644445379

### DIFF
--- a/tests/testdata/provider/testdata_pg_bigint_pk.sql
+++ b/tests/testdata/provider/testdata_pg_bigint_pk.sql
@@ -119,13 +119,58 @@ INSERT INTO qgis_test.tb_test_compound_pk (pk1, pk2, value, geom) VALUES
     (2, 1, 'test 3', ST_SetSRID(ST_Point(-47.902, -15.763), 4326)),
     (2, 2, 'test 4', ST_SetSRID(ST_Point(-47.952, -15.781), 4326));
 
+CREATE TABLE qgis_test.tb_test_composite_float_pk
+(
+    pk1 INTEGER,
+    pk2 BIGINT,
+    pk3 REAL,
+    value VARCHAR(16),
+    geom geometry(Point, 4326),
+    PRIMARY KEY (pk1, pk2, pk3)
+);
+
+INSERT INTO qgis_test.tb_test_composite_float_pk (pk1, pk2, pk3, value, geom) VALUES
+    (1, 1, 1.0,         'test 1', ST_SetSRID(ST_Point(-47.930, -15.818), 4326)),
+    (1, 2, 3.141592741, 'test 2', ST_SetSRID(ST_Point(-47.887, -15.864), 4326)),
+    (2, 2, 2.718281828, 'test 3', ST_SetSRID(ST_Point(-47.902, -15.763), 4326)),
+    (2, 2, 1.0,         'test 4', ST_SetSRID(ST_Point(-47.952, -15.781), 4326));
+
 CREATE TABLE qgis_test.tb_test_float_pk
 (
     pk REAL PRIMARY KEY,
     value VARCHAR(16),
     geom geometry(Point, 4326)
 );
-
+-- those values (pi, Euler's, and a third) will be truncated/rounded to fit
+-- PostgreSQL's internal type size. REAL is IEEE-754 4 bytes (32 bit).
 INSERT INTO qgis_test.tb_test_float_pk (pk, value, geom) VALUES
-    (3.14159, 'first test',  ST_SetSRID(ST_Point(-47.887, -15.864), 4326)),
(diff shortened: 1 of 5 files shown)


espinafre:
@nyalldawson and @elpaso , everything green here. The pgsql provider tests also pass on my local system against PostgreSQL 12.

With this implementation, I'm very comfortable and confident about the way this handles floating point types (REAL and DOUBLE PRECISION in PostgreSQL), but I'm not so sure about the consequences that this brings to NUMERIC/DECIMAL types, and your inputs will be invaluable. I've expressed most of my concerns, with examples and further references, on the originating issue, here: https://github.com/qgis/QGIS/issues/37245#issuecomment-647780860

I'm aware of the discussion in #23127 . My main concern here is the loss of precision for fixed point PostgreSQL values when QGIS casts those to double. So, what do you think?

espinafre:
@elpaso Yes, the first issue is allowing a floating point column to be used as a PK. Or, in a more plausible case, to be used as a component of a composite PK (that's how I caught this).

The NUMERIC/DECIMAL thing is another issue, and this is what worries me the most, because QGIS casts those to double, with a potential and silent loss of precision, and I believe that we, as the tool developers, have no way to know if such a precision loss is important or not for our end user. I suspect that most PostGIS tables with NUMERIC fields result from importing shapefiles, but there might be other uses for it in the wild.

I didn't see QGIS dealing with fixed-point or arbitrary precision numbers anywhere in the code. I like the idea of incorporating such a library into QGIS. The library that you pointed out seems to be unmaintained; perhaps complete? That code is small, and it could be very useful if merged into the QGIS tree, but I'm not into the legalese side of things.

I've incorporated the string conversion for NUMERICs in order to avoid the precision loss, and also to be able to use them as PKs/PK components.

I will revert this shortly, but NUMERICs won't be able to function as PKs correctly.

espinafre:
Hi, @elpaso , sorry for the delay, sometimes life gets in the way. NUMERIC values removed from this PR; they will continue to be handle as Doubles, as they have always been. Can you review? Thank you very much!

tests/testdata/provider/testdata_pg_bigint_pk.sql
```diff
@@ -119,13 +119,58 @@ INSERT INTO qgis_test.tb_test_compound_pk (pk1, pk2, value, geom) VALUES
...
+    (2.718281828459045235360287471352662497757247093699959574966967627724076630353, 'second test', ST_SetSRID(ST_Point(-47.902, -15.763), 4326)),
+    (1.333333333333333333333333333333333333333333333333333333333333333333333333333, 'third teste', ST_SetSRID(ST_Point(-47.751, -15.644), 4326));
+
+CREATE TABLE qgis_test.tb_test_numeric_pk
+(
+    pk NUMERIC(41, 40) PRIMARY KEY,
+    value VARCHAR(16),
+    geom geometry(Point, 4326)
+);
+-- those values (pi, Euler's, and a third) will be truncated to fit
```
elpaso, line 170: is this still checked?

espinafre: It is not, I removed the check for NUMERIC in the Python code, but forgot to delete the testdata. c682440 removes it, thanks!

espinafre:
Hi, @elpaso , all green here :smiley:

elpaso (approved):

espinafre:
@elpaso , do you think we can merge? Shall I make a 3.14 backport? Thanks!

Issue #37336, opened 2020-06-22
Feature request's ordering fails for {join,virtual,auxiliary} fields


nirvn:
The QgsFeatureRequest's ordering feature fails for join, virtual, and auxiliary fields as it assumes fields are all attached to the main dataset. Most noticeably, this leads to symbology's feature ordering as well as layout attribute table item's ordering failures.

Here's a small sample project to test the issue
[ordering.zip](https://github.com/qgis/QGIS/files/4812464/ordering.zip)

- To see the symbology's feature ordering bug, try ordering by the dataset's "nb" field vs. the joined "join_pos" field. Both fields have same values and yet the ordering clearly fails for the joined field.

- To see the layout attribute table item's ordering bug, open the layout embedded into the above test project.

nirvn:
Hmm, not sure why this won't fix label was added. 

There's a related UI issue here, which I'm hoping it will *not* be fixed in favor of fixing the underlying issue :) ATM, the combo boxes for feature ordering (symbology, layout attribute table, etc.) shows all fields, even the ones that don't work.

gioman:
> Hmm, not sure why this won't fix label was added.

not sure too why it happens... but it is very annoying.

m-kuhn:
It looks like the order by is currently triggered on the provider iterator. The provider iterator does not have additional fields available, so it can not handle this kind of order by. For other ordered requests it would be good to do it on provider level still, in particular where we can forward these to the provider and let the provider already order features (==> to prevent pre-fetching all features for sorting and to benefit from indexes on the db).
So, the provider iterator should receive the order by, decide if it can handle it and communicate that back to the vector layer feature iterator which should jump in if the provider iterator failed to handle it.

Gustry:
> not sure too why it happens... but it is very annoying.

It's in the issue template, when you use the first two templates : https://github.com/qgis/QGIS/issues/new/choose

m-kuhn:
I think we could improve the UX of the new issue page. The first two items are visually most prominent (They are on top and have an icon). Intuitively people will click there - also those that want to file a valid bug. I'd suggest to reorder, add icons to the other ones and move the term "plugin" to earlier in the description.

PR #37372, opened 2020-06-24
[vector layer] Fix feature request order by not working against non-provider fields

nirvn:
## Description

This PR fixes issue #37336 , whereas a feature request's order by would not work if clause(s) would rely on non-provider fields.

### DIFF
--- a/src/core/qgsvectorlayerfeatureiterator.cpp
+++ b/src/core/qgsvectorlayerfeatureiterator.cpp
@@ -140,6 +140,27 @@ QgsVectorLayerFeatureIterator::QgsVectorLayerFeatureIterator( QgsVectorLayerFeat
     mRequest.setFilterRect( mFilterRect );
   }
 
+  // check whether the order by clause(s) can be delegated to the provider
+  mDelegatedOrderByToProvider = !mSource->mHasEditBuffer;
+  if ( !mRequest.orderBy().isEmpty() )
+  {
+    QSet<int> attributeIndexes;
+    const auto usedAttributeIndices = mRequest.orderBy().usedAttributeIndices( mSource->mFields );
+    for ( int attrIndex : usedAttributeIndices )
+    {
+      if ( mSource->mFields.fieldOrigin( attrIndex ) != QgsFields::OriginProvider )
+        mDelegatedOrderByToProvider = false;
+
+      attributeIndexes << attrIndex;
+    }
+
+    if ( mRequest.flags() & QgsFeatureRequest::SubsetOfAttributes && !mDelegatedOrderByToProvider )
+    {
+      attributeIndexes += qgis::listToSet( mRequest.subsetOfAttributes() );
+      mRequest.setSubsetOfAttributes( qgis::setToList( attributeIndexes ) );
+    }
+  }
+
   if ( mRequest.filterType() == QgsFeatureRequest::FilterExpression )
   {
     mRequest.expressionContext()->setFields( mSource->mFields );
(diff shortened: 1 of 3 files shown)


m-kuhn:
Cool, thanks for tackling this!

I think the better approach would be to let the provider decide if he is able to respond to the request or wants the vector layer iterator do it for him.
This would be more future proof compared to the approach proposed in here. The provider could inspect the expression for supported functions and decide based on that if he is able to execute the order by or not. Or even (in the far future ;) ) resolve joins directly on the database.

If that's too much for now, I'd probably just execute the order by in the vector layer in any case and never forward to the provider iterator (i.e. not copy the orderBy to the provider request and execute locally).

nirvn:
@m-kuhn , I've only got time to tackle the non-provider field issue ATM. I agree it'd be a win to do provider-level (OGR et cie) order by eventually. 

So, if I understand you well, you want me to never pass on orderBy to any provider for now?

nyalldawson:
> , if I understand you well, you want me to never pass on orderBy to any provider for now?

That'd be a regression - mssql and Oracle currently handle provider side order by correctly...

nirvn:
OK, so on the basis of doing no harm, I think we shouldn't change the behavior of other providers for provider fields for the time being.

nyalldawson:
> I think the better approach would be to let the provider decide if he is able to respond to the request or wants the vector layer iterator do it for him.

I disagree - I think we'll always need something at the vector iterator level to determine in advance whether or not it's even possible to handover to the provider. Ie if the edit buffers is empty. Otherwise, every provider would have to somehow implement this vector layer level check too..

src/core/qgsvectorlayerfeatureiterator.cpp
```diff
@@ -140,20 +140,38 @@ QgsVectorLayerFeatureIterator::QgsVectorLayerFeatureIterator( QgsVectorLayerFeat
...
+  mDelegatedOrderByToProvider = !mSource->mHasEditBuffer;
+  if ( !mRequest.orderBy().isEmpty() )
+  {
+    const auto usedAttributeIndices = mRequest.orderBy().usedAttributeIndices( mSource->mFields );
+    for ( int attrIndex : usedAttributeIndices )
+    {
+      if ( mSource->mFields.fieldOrigin( attrIndex ) != QgsFields::OriginProvider )
+        mDelegatedOrderByToProvider = false;
+
+      attributeIndexes << attrIndex;
```
nyalldawson, line 155: We also need to check the various added/edited/deleted feature maps are empty. If not, we can't delegate to provider...

nirvn: This is happening on line 146, i.e. `mDelegatedOrderByToProvider = !mSource->mHasEditBuffer;`

m-kuhn:
Ok, convinced. Please proceed!

src/core/qgsvectorlayerfeatureiterator.cpp
```diff
@@ -140,14 +140,35 @@ QgsVectorLayerFeatureIterator::QgsVectorLayerFeatureIterator( QgsVectorLayerFeat
...
+    }
+
+    if ( mRequest.flags() & QgsFeatureRequest::SubsetOfAttributes )
```
m-kuhn, line 157:
```suggestion
    if ( mRequest.flags() & QgsFeatureRequest::SubsetOfAttributes && !mDelegatedOrderByToProvider )
```

m-kuhn (approved):

src/core/qgsvectorlayerfeatureiterator.cpp
nirvn: Hm, are you sure about that?

nirvn: I guess it makes sense since we're adding those order by fields to the mProviderRequest down the line.

m-kuhn: Is there a chance left that we will do the order by on this level even if we delegate to provider?

PR #37380, opened 2020-06-24
PG styles: add type column

### LINKED ISSUE #37338: Unable to see styles stored in the database
Since the update to 3.14 it is impossible to see the styles stored in the database, this makes this error:
```
2020-06-22T08:21:29     WARNING    Requête erronée : SELECT styleQML FROM layer_styles WHERE f_table_catalog='fibrea' AND f_table_schema='pit' AND f_table_name='iti' AND f_geometry_column ='geom' AND type='Line' ORDER BY CASE WHEN useAsDefault THEN 1 ELSE 2 END,update_time DESC LIMIT 1 a retourné 7 [ERROR: column "type" does not exist
             LIGNE 1 : ...able_name='iti' AND f_geometry_column ='geom' AND type='Line...
              ^
```

elpaso:
Fixes #37338

### DIFF
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -357,6 +357,22 @@ void QgsPostgresProvider::reloadProviderData()
   mLayerExtent.setMinimal();
 }
 
+bool QgsPostgresProviderMetadata::ensureTypeColumn( QgsPostgresConn *conn, const QgsDataSourceUri &uri, QString &errCause ) const
+{
+  if ( !columnExists( *conn, QStringLiteral( "layer_styles" ), QStringLiteral( "type" ) ) )
+  {
+    QgsPostgresResult res( conn->PQexec( "ALTER TABLE layer_styles ADD COLUMN type varchar NULL" ) );
+    if ( res.PQresultStatus() != PGRES_COMMAND_OK )
+    {
+      errCause = QObject::tr( "Unable to add column type to layer_styles table. "
+                              "Maybe this is due to table permissions (user=%1). "
+                              "Please contact your database admin" ).arg( uri.username() );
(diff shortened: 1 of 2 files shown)


jgrocha:
Thank you @elpaso for this fix. I did almost the same locally, but I still get an issue when applying old styles (reporting bad type mismatch, which might not be related). I'll switch to your PR and use my test setup.

jgrocha:
Two issues with this fix, that can be improved:
- If one style with the same name already exists (but with null), the new one is written, instead of updating the existing one,
- If there is an old default style, the default flag is not removed.

elpaso:
@jgrocha cool, I leave it to you :)

jgrocha (changes requested):
As I've suggested, we can improve this fix but preventing the existence of the same style name and making sure that we remove the `useasdefault` flag for all other available styles.

elpaso:
> As I've suggested, we can improve this fix but preventing the existence of the same style name and making sure that we remove the `useasdefault` flag for all other available styles.

I got it, please go ahead.

jgrocha:
Ok, thanks.

elpaso:
Anyway, the main issue here was that your original implementation was not patching the table (adding the type col) on style load but only on style write.

jgrocha:
> Anyway, the main issue here was that your original implementation was not patching the table (adding the type col) on style load but only on style write.

Yes, but we can live with that. My proposal is to keep the old table as it is, if we are just reading the project. My new PR ensures that we can read the old table and add only the new column on the first write.

I've tested it carefully by opening old projects and trying all possible interactions.

This regression was introduced by me and I'm very sorry for the noise. I hope this PR makes the transition as transparent as possible.

A big thank you @elpaso to keep an eye and the pressure to fix this.

PR #37387, opened 2020-06-24
Fix missing type column in layer_styles in Postgresql

### LINKED ISSUE #37338: Unable to see styles stored in the database

jgrocha:
## Description

This fix #37338, providing a transparent transition from previous QGIS versions, regarding the `layer_styles` table stored on the Postgresql provider. This issue was introduced by #32138.

`layer_styles` is only changed when styles are written. Reading only old projects and retrieving styles does not change the database.

### DIFF
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -5082,7 +5082,7 @@ bool QgsPostgresProviderMetadata::saveStyle( const QString &uri, const QString &
                                 " AND f_table_schema=%2"
                                 " AND f_table_name=%3"
                                 " AND f_geometry_column=%4"
-                                " AND type=%5"
+                                " AND (type=%5 OR type IS NULL)"
                                 " AND styleName=%6" )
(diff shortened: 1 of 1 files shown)


elpaso (approved):
LGTM, it would be nice to have a test so that it won't break in the future but we can do it in a separate PR.

CoryAlbrecht:
Regression #39850 .

PR #37389, opened 2020-06-24
Memory dataset group and possibility to persist it

vcloarec:
This PR introduce memory dataset groups for mesh layer. These dataset groups are temporary and are not kept when the project is closed.

Memory dataset groups can be created from the mesh calculator with a new option.

This PR also allows the possibility to remove or save these memory dataset groups to a file with specified driver.

memory

### DIFF
--- a/src/app/mesh/qgsmeshdatasetgrouptreeview.cpp
+++ b/src/app/mesh/qgsmeshdatasetgrouptreeview.cpp
@@ -596,7 +596,8 @@ void QgsMeshDatasetGroupTreeView::contextMenuEvent( QContextMenuEvent *event )
 
 void QgsMeshDatasetGroupTreeView::removeCurrentItem()
 {
-  mModel->removeItem( currentIndex() );
+  if ( QMessageBox::question( this, tr( "Remove Dataset Group" ), tr( "Remove dataset group?" ) ) == QMessageBox::Ok )
+    mModel->removeItem( currentIndex() );
 }
 
 void QgsMeshDatasetGroupTreeView::onDatasetGroupSaved()
--- a/src/app/mesh/qgsmeshdatasetgrouptreeview.cpp
+++ b/src/app/mesh/qgsmeshdatasetgrouptreeview.cpp
@@ -627,7 +627,7 @@ QMenu *QgsMeshDatasetGroupTreeView::createContextMenu()
       break;
     case QgsMeshDatasetGroupTreeItem::Memory:
     case QgsMeshDatasetGroupTreeItem::OnTheFly:
-      contextMenu->addAction( tr( "Remove dataset group" ), this, &QgsMeshDatasetGroupTreeView::removeCurrentItem );
+      contextMenu->addAction( tr( "Remove Dataset Group" ), this, &QgsMeshDatasetGroupTreeView::removeCurrentItem );
       mSaveMenu->createSaveMenu( groupIndex, contextMenu );
       break;
   }
@@ -655,7 +655,7 @@ QMenu *QgsMeshDatasetGroupSaveMenu::createSaveMenu( int groupIndex, QMenu *paren
     return  nullptr;
 
   QMenu *menu = new QMenu( parentMenu );
-  menu->setTitle( QObject::tr( "Save Datasets group as..." ) );
+  menu->setTitle( QObject::tr( "Save Datasets Group as..." ) );
   QgsMeshDatasetGroupMetadata groupMeta = mMeshLayer->datasetGroupMetadata( groupIndex );
 
   QgsProviderMetadata *providerMetadata = QgsProviderRegistry::instance()->providerMetadata( mMeshLayer->dataProvider()->name() );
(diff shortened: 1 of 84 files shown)


src/app/qgisapp.cpp
```diff
@@ -12854,31 +12856,36 @@ bool QgisApp::checkMemoryLayers()
     }
     else if ( it.value() && it.value()->isTemporary() )
       hasTemporaryLayers = true;
+    else if ( it.value() && it.value()->type() == QgsMapLayerType::MeshLayer )
```
nyalldawson, line 12859: Instead of this, you should override QgsMapLayer::isTemporary

src/app/mesh/qgsmeshdatasetgrouptreeview.cpp
```diff
@@ -504,6 +583,57 @@ void QgsMeshDatasetGroupTreeView::deselectAllGroups()
...
+
+  switch ( item->storageType() )
+  {
+    case QgsMeshDatasetGroupTreeItem::None:
+      break;
+    case QgsMeshDatasetGroupTreeItem::File:
+      break;
+    case QgsMeshDatasetGroupTreeItem::Memory:
+    case QgsMeshDatasetGroupTreeItem::OnTheFly:
+      contextMenu->addAction( tr( "Remove dataset group" ), this, &QgsMeshDatasetGroupTreeView::removeCurrentItem );
```
nyalldawson, line 630: Convention is Title Case for menu actions

src/app/mesh/qgsmeshdatasetgrouptreeview.cpp
```diff
@@ -518,3 +648,86 @@ void QgsMeshDatasetGroupTreeView::selectAllItem( bool isChecked )
...
+  if ( !mMeshLayer )
+    return;
+
+  QgsSettings settings;
+  QString filter;
+  if ( !fileSuffix.isEmpty() )
+    filter = QStringLiteral( "%1 (*.%2)" ).arg( driver ).arg( fileSuffix );
+  QString exportFileDir = settings.value( QStringLiteral( "lastMeshDatasetDir" ), QDir::homePath(), QgsSettings::App ).toString();
+  QString saveFileName = QFileDialog::getSaveFileName( nullptr,
+                         QObject::tr( "Save mesh datasets" ),
```
nyalldawson, line 712: Title Case for dialog titles

nyalldawson:
@nirvn mind doing a UI review of this? 

One issue I can see immediately is that the calculator isn't using the correct font in the editor, so we're seeing the qt 5.14 issue with the ridiculous font spacing...

src/app/qgisapp.cpp
vcloarec: done, but, here I prefer keep the distinguish between mesh layer and other type because the logic is not exactly the same : for other, all the mesh is temporary, for mesh layer, only some dataset groups.

vcloarec:
> One issue I can see immediately is that the calculator isn't using the correct font in the editor, so we're seeing the qt 5.14 issue with the ridiculous font spacing...

In the GIF, It is built with Qt 5.12.8. Maybe is it the space with a variable, for example "vitesse     ms". In that case, it is due to the data, the name of the dataset group has the unity at the end, separated with several spaces.

PeterPetrik:
I have already reviewed this privately in https://github.com/vcloarec/QGIS/pull/22

nyalldawson:
Personally it looks odd to me to show the memory indicator icon in the mesh calculator dialog. Why is it needed there?

vcloarec:
Memory dataset groups are not persitent and will be removed after closing the project. User could need to know this dataset group is not persistent, especially if this dataset is used to calculate another one. For now, there is no consequences because there is no dependence after the creation of the new one, but there will be soon...

But I agree, it is not the best UI design, especially in the calculator dialog. Maybe it will be better with icon on the right?

nirvn:
Please give me some time to review the UI.

One question: why are those virtual datasets have to be non-persistent?

vcloarec:
> One question: why are those virtual datasets have to be non-persistent?

Those "virtual" dataset group are stored in memory, not and file, so can't be retrieved after closing the project.

For information, there is an upcoming PR with "on the fly" dataset group, and those new dataset groups will be calculated on real time, and some (those that not depend on temporary dataset) could be persistent.

nirvn:
@vcloarec , couldn't you simply store the formula and re-compute it upon re-opening the project?

vcloarec:
> @vcloarec , couldn't you simply store the formula and re-compute it upon re-opening the project?

The "on the fly"  will store the formula and calculate when needed. For the "memory", the formula is not stored, it is not the concept for now. But it could be later... For now, "memory" are just temporary dataset that could be persist in file if needed (as "on the fly")

nirvn:
@vcloarec , I'm wondering whether this temporary vs persistent virtual dataset adds a level of UI/UX complexity without much to gain from. I.e., if those virtual datasets would all be persistent, you wouldn't need to have the memory icon anywhere, etc.

PeterPetrik:
I propose to merge this as it is, and once the on-the-fly PR lands, we can clean the UI in one go?

nirvn:
Could we at least remove all memory icons from this PR? That can't stay in any case.

vcloarec:
With the last commit, the memory icon is removed from the dataset active widget and from the mesh calculator.
However, it is still in the available dataset group tree view in the source tab of mesh properties. But I put it on the right of the tree view row.
I think it is necessary to distinguish somewhere dataset group type because "memory" items have not the same behavior (they can be removed or saved), and the user needs to know if there is data that will not be saved if the project is closed.

@nirvn , let me know if it is ok, or do I need to remove ALL?

PeterPetrik:
talked with @nirvn , agreed on the plan forward, so we can merge this one and improve in the on-the-fly dataset followup PR.

PR #37393, opened 2020-06-25
force reference time (current) to be UTC for mesh layer

vcloarec:
Force the reference time for mesh layer to be in UTC in any cases

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -5552,7 +5552,7 @@ QgsMeshLayer *QgisApp::addMeshLayerPrivate( const QString &url, const QString &b
 
   QDateTime referenceTime = QgsProject::instance()->timeSettings()->temporalRange().begin();
   if ( !referenceTime.isValid() ) // If project reference time is invalid, use current date
-    referenceTime = QDateTime( QDate::currentDate(), QTime( 0, 0, 0, Qt::UTC ) );
+    referenceTime = QDateTime( QDate::currentDate(), QTime( 0, 0, 0, Qt::UTC ), Qt::UTC );
 
   if ( ! layer || ( !layer->isValid() && layer->subLayers().isEmpty() ) )
   {
--- a/src/core/mesh/qgsmeshlayer.cpp
+++ b/src/core/mesh/qgsmeshlayer.cpp
@@ -201,7 +201,7 @@ bool QgsMeshLayer::addDatasets( const QString &path, const QDateTime &defaultRef
       {
         QDateTime referenceTime = defaultReferenceTime;
         if ( !defaultReferenceTime.isValid() ) // If project reference time is invalid, use current date
-          referenceTime = QDateTime( QDate::currentDate(), QTime( 0, 0, 0, Qt::UTC ) );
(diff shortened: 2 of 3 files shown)


nyalldawson:
This will likely break the temporal controller, which assumes local times are used everywhere...

vcloarec:
> This will likely break the temporal controller, which assumes local times are used everywhere...

hmmm, I believed it was UTC. I had seen in the temporal controler widget, the time specification is UTC :
https://github.com/qgis/QGIS/blob/5936249b722b514a741adb22ffab8639a943c4a6/src/ui/qgstemporalcontrollerwidgetbase.ui#L416

Maybe I miss something...

nyalldawson:
Eh... this is getting too complex/confusing. I think we're going to have to "rip off the bandaid" and just move to always showing timezone information for datetime values everywhere, and then make sure the expression engine handles this correctly. Complicated... :(

vcloarec:
I don't know exactly how other layers work with temporal, but for mesh layer, it is quite simple as time zone is not supported by MDAL. So whatever if local or UTC, as long as we know which one is it...  
For now and before this PR, for mesh layer, it seems everything is ok for now except when the reference time of the data is invalid and the the reference time is set to current date with time 00:00. As this time is set to local, there is an offset depending of the local time zone that some time leads to a rendering issue.
This PR only force this default time to avoid this issue and there isn't consequences for other part of QGIS.

vcloarec:
@nyalldawson , Does this PR is really problematic?

PeterPetrik:
If I understand it correctly...

1. time slider is in local time (implemented)
2. user should see time zone (not-implemented) everywhere where he sees time
3. data providers should give time to time controllers in local time (right now every provider does differently, mess?)

so if this is way forward, we should rework this PR so mesh data provider return time in local timezone? @nyalldawson

vcloarec:
>     1. time slider is in local time (implemented)

For now, time slider seems to be in UTC... 
But, before QGIS could support time zone, can we implement it in local time? In other words, set the time edit in local specification could break anything for other layer format? If we can implement temporal controller to be local, this is a good point for mesh layer because we don't need anymore the workaround that consist to force the time to be in UTC.
What about a combo box to simply choose between UTC and local for the time slider?
In this case, for mesh layer, no offset if the temporal controller is set to local, and an offset depending of the time zone of the user if the temporal controller is set to UTC (because for MDAL, the time is supposed to be local).

vcloarec:
not stale

vcloarec:
@nyalldawson 
Can we do something for this?
Without that, formats that do not have native reference time have always displaying issue when just loaded : the datasets is not displayed or with the wrong dataset.
I understand that the choice between UTC or local can make troubles, but for mesh layer, we have to now which spec is it in the time controller to accord the date/time for the loaded mesh and avoiding offset between displayed time and dataset time.
If for now, time controller keep the UTC spec, we have to force the UTC for mesh layer.
It will be really good to have this issue fixed for the next 3.14 release.

nyalldawson:
Ok - let's merge and deal with the fallout when it happens :)

nyalldawson:
(can we skip the backport for now though?)

nyalldawson (approved):

vcloarec:
> can we skip the backport for now though?

that is for the backport I insisted heavily...

PR #37403, opened 2020-06-25
Reactivate WMS GetPrint unit tests

rldhont:
## Description

Since #35277 [Tests][Server] WMS GetPrint group, the QGIS Server WMS GetPrint tests segfault because it takes too long time. The max duration of a test is 55 sec https://github.com/qgis/QGIS/commit/fc561d0375a4df8b38bcc0c1036dc27e644984c0 and the QGIS Server WMS GetPrint takes more than 55 sec to finish. The test was also too big.

The solution is to split the QGIS Server WMS GetPrint tests.


nyalldawson:
It's probably too annoying to fix now, but the reason these tests are so slow is that they use 300 DPI outputs. All the main layout checks (non server) drop the output DPI to keep the reference images small, and it greatly speeds up the reference/rendered comparison (simply due to much lower pixel count to check through!)

nyalldawson (approved):
Lgtm, let's give it a try!

rldhont:
Travis is green !

pblottiere:
> Travis is green !

Awesome :)

elpaso:
@rldhont excellent! What was exactly the problem?

Gustry:
Great news!

> 611/648 Test #611: PyQgsServerWMSGetPrint ............................   Passed   51.32 sec

Any chance you can move another test? It's pretty close to 55 seconds...

rldhont:
@Gustry I have added an other split

PR #37411, opened 2020-06-25
Name role in expression item to be stable on fields in the expression builder

signedav:
Issue has been that on using alias for the fields the label of the expression item are `<name> (<alias>)` and this label has been used to get the values listed as examples. By introducing the new role `ITEM_NAME_ROLE` we work with the field name instead of the label. This role is currently only used for the expression items for the fields. But I think it's the proper way. Maybe it can be used for other expression items as well in future.


signedav:
@NathanW2 Could you have a look at this? Thanks.

nyalldawson (approved):

PR #37429, opened 2020-06-26
[feature] Server WFS3 api sorting

elpaso:
For documentation: the API is really self-documented thanks to OPENAPI, but here is some information about the new parameters:

     "sortby" : {
        "name" : "sortby",
        "in" : "path",
        "description" : "Sort results by the specified field name",
        "required" : false,
        "schema" : {
          "type" : "string"
        }
      },
      "sortdesc" : {
        "name" : "sortdesc",
        "in" : "path",
        "description" : "Sort results in descending order, field name must be specified with 'sortby' parameter",
        "required" : false,
        "schema" : {
          "type" : "bool"
        }
      },

### DIFF
--- a/resources/server/api/ogc/schema.json
+++ b/resources/server/api/ogc/schema.json
@@ -374,7 +374,7 @@
           "type" : "string"
         }
       },
-      "sortBy" : {
+      "sortby" : {
         "name" : "sortby",
         "in" : "path",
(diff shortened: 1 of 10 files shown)


pblottiere:
It looks good to me.

Thanks @elpaso :+1:

PR #37430, opened 2020-06-26
Add project load profile times to debugging tools dock

nyalldawson:
This PR reworks the QgsRuntimeProfiler class, to:
1. make the profiling thread safe, so that it's possible to record times across multiple threads
2. allows different profiling "groups"

The new groups api is used here to expose a breakdown of project load times in the debugging dock, alongside the existing QGIS startup profiling. Now it's possible to get a breakdown of the various stages of project load, in order to identify the causes of slow project load times.

TODO: in 99% of cases this will be caused by one or more layers, so it would be nice to further break down the individual layer load times, e.g. so that for a WFS layer we see the time taken for the initial server capabilities queries, etc...
For now, you just get the overall load time of each individual layer.

image

image

ALSO TODO: Add a group with profile times for map rendering, so you can see exactly which layers in your project are causing slow map redraw times...


nyalldawson:
TODO 3: expose via a QGIS server administration console :wink:

PR #37439, opened 2020-06-27
3D map view / network logger icon fix and harmonization

nirvn:
## Description

Before vs. PR:
image

I've also taken the time to harmonize the play button of the 3D map panels:
Screenshot from 2020-06-27 11-30-51


nirvn:
This little cleanup exercise made me realize we have a glaring inconsistency in our temporal controller vs 3D map animation loop widget. I've updated the 3D map animation loop widget to match that of the temporal controller (the refresh icon used to loop wasn't very explicit
Screenshot from 2020-06-27 12-24-22
 anyways):

Haaa, harmony.

nyalldawson:
What's your thoughts on the play icon for network logger? I think a "record" icon would probably be more suitable...

nirvn:
@nyalldawson , true.

nirvn:
Done:
image

nyalldawson:
Much nicer, thanks!

nirvn:
Last UI harmonization commit for this PR: move icon-less toolbar actions (something we do _not_ do anywhere in the app) under a toolbar menu:
image
 
This UI/UX matches what we have elsewhere and is more future-proof.

nirvn:
OK, last one I swear. I just realized the 3D map view's play button didn't reflect the playback state (i.e. like the temporal controller, the button maintains its pressed state when playback is active).

The last commit fixes that:
Peek 2020-06-27 16-01

nirvn:
Unrelated azure failure, merging.

PR #37451, opened 2020-06-27
[afs] Read field aliases from layer definition

nyalldawson:
Respect provider side field aliases from ArcGIS feature services


nirvn (approved):

nyalldawson:
Unrelated test failure

PR #37457, opened 2020-06-28
Fix raster quantile with float

### LINKED ISSUE #37449: Raster quantile symbology does not work for float64 rasters
Sample data: https://github.com/qgis/QGIS/files/4840624/RISCO.zip

Must be tested on 3.10.7 because on 3.14/master quantile symbology is broken, see: https://github.com/qgis/QGIS/issues/37448

From https://github.com/qgis/QGIS/issues/35465#issuecomment-650565325 :

With this raster, quantiles are not correctly calculated (in this case, in QGIS 3.10.7, as it does not work in 3.14).

Rounding that raster and converting to a Int32, it already calculate quantiles correctly (image on the right):

quantile_float64

Using Raster Pixels to Points algorithm to export Float64 pixel values to points, and calculating quantiles in the resulting vector layer, the output is the correct (as seen on the left image).

Running r.quantile in the Float64 raster, the output is correct:

0:20.000000:120.000000
1:40.000000:1320.000000
2:60.000000:2400.000000
3:80.000000:4500.000000 

Also correct exporting to CSV and calculating in R:

> tabela_risco <- read.csv("D:\\RISCO_RASTER_PIXELS_TO_POINTS.csv")
> quantile(tabela_risco$VALUE, probs = c(0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0))
      0%      10%      20%      30%      40%      50%      60%      70%      80%      90%     100% 
    36.0     72.0    120.0    792.0   1320.0   1920.0   2400.0   3000.0   4500.0   6000.0 218030.4

So, it seems that quantile classification is still failing with some kind of data.

elpaso:
Fixes #37449

### DIFF
--- a/src/core/raster/qgsrasterinterface.cpp
+++ b/src/core/raster/qgsrasterinterface.cpp
@@ -363,8 +363,8 @@ void QgsRasterInterface::initHistogram( QgsRasterHistogram &histogram,
       }
       else
       {
-        // This is for not integer types:
-        myBinCount = std::min( 2000, histogram.width * histogram.height );
+        // This is for not integer types where we cannot limit the bin size
+        myBinCount = histogram.width * histogram.height;
       }
     }
   }
--- a/src/core/raster/qgsrasterinterface.cpp
+++ b/src/core/raster/qgsrasterinterface.cpp
@@ -306,7 +306,7 @@ void QgsRasterInterface::initHistogram( QgsRasterHistogram &histogram,
   {
     // Calc resolution from theSampleSize
     double xRes, yRes;
-    xRes = yRes = std::sqrt( ( finalExtent.width() * finalExtent.height() ) / sampleSize );
+    xRes = yRes = std::sqrt( ( static_cast<double>( finalExtent.width( ) ) * finalExtent.height() ) / sampleSize );
 
     // But limit by physical resolution
     if ( capabilities() & Size )
@@ -336,7 +336,7 @@ void QgsRasterInterface::initHistogram( QgsRasterHistogram &histogram,
   }
   QgsDebugMsgLevel( QStringLiteral( "theHistogram.width = %1 histogram.height = %2" ).arg( histogram.width ).arg( histogram.height ), 4 );
 
-  long myBinCount = binCount;
+  qint64 myBinCount = binCount;
   if ( myBinCount == 0 )
   {
     // TODO: this was OK when stats/histogram were calced in provider,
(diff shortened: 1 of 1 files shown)


src/core/raster/qgsrasterinterface.cpp
```diff
@@ -363,8 +363,8 @@ void QgsRasterInterface::initHistogram( QgsRasterHistogram &histogram,
       }
       else
       {
-        // This is for not integer types:
-        myBinCount = std::min( 2000, histogram.width * histogram.height );
+        // This is for not integer types where we cannot limit the bin size
+        myBinCount = histogram.width * histogram.height;
```
rouault, line 367: not related to this change, but I see we could potentially have a integer overflow on multiplication on sufficiently large raster when sampleSize == 0 and ( capabilities() & Size ). Same just above at line 362. At line 309 too

rouault: `myHistogram.histogramVector.resize( myBinCount );` at line 430 could throw a out-of-memory exception if myBinCount is too large. Perhaps 2000 is too small, but unbounded bincount could be a memory & performance killer. Maybe 1 million or 10 million upper limit ?

elpaso: > not related to this change, but I see we could potentially have a integer overflow on multiplication on sufficiently large raster when sampleSize == 0 and ( capabilities() & Size ). Same just above at line 362. At line 309 too

Are you sure? https://github.com/qgis/QGIS/pull/37457/files/d49e140cf66468f9743fe2cb05b4f155898e37f8#diff-1a8a563373bea4e36bc554b9ff1c0067R309 these are all already `double`.

Ok for line 362.

Please see my last commit for the limits.

src/core/raster/qgsrasterinterface.cpp
```diff
@@ -336,7 +336,7 @@ void QgsRasterInterface::initHistogram( QgsRasterHistogram &histogram,
   }
   QgsDebugMsgLevel( QStringLiteral( "theHistogram.width = %1 histogram.height = %2" ).arg( histogram.width ).arg( histogram.height ), 4 );
 
-  int myBinCount = binCount;
+  long myBinCount = binCount;
```
rouault, line 339: perhaps use long long or qint64, since long will be 32 bit on Linux 32 bit, and Windows 32 & 64 bit.

src/core/raster/qgsrasterinterface.cpp
rouault: > Are you sure? https://github.com/qgis/QGIS/pull/37457/files/d49e140cf66468f9743fe2cb05b4f155898e37f8#diff-1a8a563373bea4e36bc554b9ff1c0067R309 these are all already `double`

at line 309, ( finalExtent.width() * finalExtent.height() ) / sampleSize is evaluated as a integer. std::sqrt() transforms to double only after. Casting finalExtent.width() to double should avoid any int overflow

PR #37459, opened 2020-06-28
Fix restoring dialog geometry after minimize when enableAutoGeometryRestore is used

### LINKED ISSUE #37230: "Automatically hide georeferencer window" option does not correctly restore georeferencer dialog position
I'm georeferencing rasters, using two monitors: QGIS main canvas in the first monitor and the georeferencer dialog in the second one.
1. Select "Add point" button
1. Click the place in the georeferencer window. The "Enter Map coordinates" dialog opens.
1. Select "From map canvas" in that dialog. The georeferencer dialog is reduced
1. Click the actual placement in the main window. The georeferencer and the "Enter map coordinates" dialogs reopen BUT... the georeferencer dialog is not placed at its previous position. I often get it, overlapping the two monitors (somewhat at the middle of the two screens width). This means that I have to move it fully to the second monitor before adding the next gcp point.

This happened until I decided to report and, writing down the steps one by one made me realize that the culprit is the "Automatically hide georeferencer window" option in the "Enter Map coordinates" dialog. Once I unchecked it, I no longer reproduce the issue (normal, the dialog does not hide/show anymore).
But whether this option is checked or not, I think the georeferencer dialog should honor its previous placement (as long as the monitor exists)

Using master 3.13 e9a0c01b21 on Windows 10

uclaros:
## Description
There is a bug in how `QgsGui::enableAutoGeometryRestore` functions. A minimized window does not store its geometry before minimizing but recalls the geometry saved the last time it was closed when it is restored. An example case is #37230 
Storing the geometry also on hide should fix that. Are the `Close` and `Destroy` cases now redundant?

Fixes #37230

### DIFF
--- a/src/gui/qgswidgetstatehelper_p.cpp
+++ b/src/gui/qgswidgetstatehelper_p.cpp
@@ -28,7 +28,9 @@ QgsWidgetStateHelper::QgsWidgetStateHelper( QObject *parent ) : QObject( parent 
 
 bool QgsWidgetStateHelper::eventFilter( QObject *object, QEvent *event )
 {
-  if ( event->type() == QEvent::Close || event->type() == QEvent::Destroy )
+  if ( event->type() == QEvent::Close ||
+       event->type() == QEvent::Destroy ||
+       event->type() == QEvent::Hide )
   {
     QWidget *widget = qobject_cast<QWidget *>( object );
 


nyalldawson (approved):
LGTM - let's give it a shot!

uclaros:
Isn't it safe to backport at least to 3.14? The georeferencer bug is really annoying!

PR #37464, opened 2020-06-28
Add modeler algorithm to create directories

nyalldawson:
This algorithm allows models to create new file directories. Useful for
models which have to output results to certain dynamically created paths
(e.g. a folder containing the current date)


nirvn (approved):

Saijin-Naib:
Super useful. This would have helped a ton working with Agricultural data!

Thanks, Nyall!

roya0045:
Same point as for filter, as this is specific to models, do you have any opposition if I change the group of this alg?

PR #37474, opened 2020-06-29
Layout legend maximum marker size

mhugent:
If there are many point layers with size in map units, the size of the point markers in the layout legend can get quite big. Therefore it would be handy to have the  possibility to restrict the maximum marker symbol size while the automatic legend update is still active. This PR provides the possibility to enter the maximum marker size (in mm) in the layout legend widget. What do you think about it?


andreasneumann:
So - if the size would grow beyond the maximum size it would still display with the maximum - right?

How about the minimum size: with sizes/widths in map units symbols would also be illegible when things to too small.

andreasneumann:
@mhugent Do you have a screenshot how/where the configuration would be done in the layer/symbology properties?

mhugent:
>So - if the size would grow beyond the maximum size it would still display with the maximum - right?
Yes, correct
>How about the minimum size: with sizes/widths in map units symbols would also be illegible when things to too small
Good point, going to add that as well

andreasneumann:
> If there are many point layers with size in map units, the size of the point markers in the layout legend can get quite big. Therefore it would be handy to have the possibility to restrict the maximum marker symbol size while the automatic legend update is still active. This PR provides the possibility to enter the maximum marker size (in mm) in the layout legend widget. What do you think about it?

I am glad you work on it. Incidentally, we ran into exactly the same issue today with our team. Was about to create an issue report, but then I saw your PR ;-)

andreasneumann:
> If there are many point layers with size in map units, the size of the point markers in the layout legend can get quite big.

The same issue applies to stroke-width legends for linestrings/polygon layers, and potentially other places (like font sizes for font-markers, etc.). But we are not the client/contractor here, so the original client needs to decide whether they want to invest fixing the stroke-width issues as well ...

mhugent:
It would be straightforward to implement it also for line width. Polygon outline is more difficult, since it is not possible to set outline width at QgsFillSymbol - level

mhugent:
max_marker_size

Here is the screenshot

andreasneumann:
One thing I wonder:

We have the same issue on both legends: the layer tree widget in the main canvas window and also in the layout legend.

Because the same/similar issue is on both legend types, I wonder if the setting could/should reside in the layer symbology properties instead? I'm not sure yet - just brainstorming ...

mhugent:
I've now added a minimum symbol size. Also, minimum / maximum size in mm are now valid for both marker size and line width. Or should min/max marker size and min/max line width two separate settings?

src/core/layertree/qgslayertreemodellegendnode.cpp
```diff
@@ -557,10 +557,25 @@ QSizeF QgsSymbolLegendNode::drawSymbol( const QgsLegendSettings &settings, ItemC
...
+
+    //If marker size exceeds max marker size, clone symbol and set max size in mm
+    if ( ( maxSymbolSize > 0 && size > maxSymbolSize ) || ( minSymbolSize > 0 && size < minSymbolSize ) )
+    {
+      minMaxSizeMarkerSymbol.reset( dynamic_cast<QgsMarkerSymbol *>( s->clone() ) );
+      minMaxSizeMarkerSymbol->setSize( size > maxSymbolSize ? maxSymbolSize : minSymbolSize );
+      minMaxSizeMarkerSymbol->setSizeUnit( QgsUnitTypes::RenderMillimeters );
+      s = minMaxSizeMarkerSymbol.get();
+      size = ( size > maxSymbolSize ) ? maxSymbolSize : minSymbolSize;
+    }
```
nyalldawson, line 577: Could you move this out to a function in QgsSymbolLayerUtils? (same with line symbol scaling)

andreasneumann:
> One thing I wonder:
> 
> We have the same issue on both legends: the layer tree widget in the main canvas window and also in the layout legend.
> 
> Because the same/similar issue is on both legend types, I wonder if the setting could/should reside in the layer symbology properties instead? I'm not sure yet - just brainstorming ...

@nyalldawson @wonder-sk - do you have any thoughts here? We have two legend types: the layer tree legends and the print layout legends. Both legends suffer from this issue. Couldn't we have a common setting for both legends or do you think we would need to handle these two legend types separately?

nyalldawson:
> do you have any thoughts here? We have two legend types: the layer tree legends and the print layout legends. Both legends suffer from this issue. Couldn't we have a common setting for both legends or do you think we would need to handle these two legend types separately?

Well -- i'd probably just hardcode in some reasonable ranges for the main layer tree, and apply them only when symbols have map unit based sizes...

andreasneumann:
> > do you have any thoughts here? We have two legend types: the layer tree legends and the print layout legends. Both legends suffer from this issue. Couldn't we have a common setting for both legends or do you think we would need to handle these two legend types separately?
> 
> Well -- i'd probably just hardcode in some reasonable ranges for the main layer tree, and apply them only when symbols have map unit based sizes...

So hard-coding for the layer legend tree and user-defined values for the print layout legends like in this PR? Yes - would make sense to me.

We just had the issues recently that also the layer legend tree grows and shrinks to insane values ... but hard-coded values would probably work well there.

nyalldawson:
Can you add some tests for this too please?

PR #37494, opened 2020-06-30
Minimum year possible in layout map item temporal extent should be 1

nirvn:
## Description

@nyalldawson , one more.


nyalldawson (approved):

nyalldawson:
Thanks!

PR #37498, opened 2020-06-30
Add api to QgsMapSettings to specify clipping regions to apply while rendering maps

nyalldawson:
This is all API only for now, but this PR adds the backend capabilities to specify clipping regions for use during map render operations.

The initial use case is for a follow up PR which will add an option to automatically clip layout map renders to the current atlas feature geometry, but I've deliberately kept the API here very flexible as there's many other potential use cases for clipping regions during render (e.g. to allow for non-square map renders!)

For non-vector layers, the clipping is applied through a QPainter clipping path. For vector layers, the API exposes various options in how the clipping should be applied:

1. As an "intersects" style test only -- i.e. all vector features which intersect the path are rendered, without actually clipping the feature's geometry

2. By modifying the feature's geometry and clipping the geometry to the region. This means that any stroke/outlines set for the symbol will be rendered along the boundary of the clipping region, even if the feature actually partially falls outside this region.

or 

3. Using a painter clip path, so that features are clipped to the region BUT at painter time only -- so boundaries of the feature which fall outside the clip region WON'T be visible

It remains to be determine how much of this we'd actually expose to users, but it's good to have the flexibility in the API!

Sponsored by City of Canning


nyalldawson:
... and if I do a bit of hackery to expose this in layouts, here's the result :wink: 
image

image

image

image

src/core/qgsmapclippingregion.h
```diff
@@ -0,0 +1,125 @@
...
+class CORE_EXPORT QgsMapClippingRegion
+{
+  public:
+
+    /**
+     * Feature clipping behavior.
+     */
+    enum class FeatureClippingType : int
+    {
+      Intersect, //!< Clip the geometry of these features to the region prior to rendering (i.e. feature boundaries will follow the clip region)
```
nirvn, line 40: Wondering if there's a better name for it (to try and avoid the one-letter Intersect vs Intersects difference).

Maybe OverlayClip or IntersectClip?

src/core/qgsmapsettings.cpp
```diff
@@ -682,6 +682,21 @@ void QgsMapSettings::setLabelBoundaryGeometry( const QgsGeometry &boundary )
   mLabelBoundaryGeometry = boundary;
 }
 
+void QgsMapSettings::addClippingRegion( const QgsMapClippingRegion &region )
```
nirvn, line 685: Background question: have we ever discussed whether and when we should use add vs append? I have absolutely no strong feeling either way, just wondering.

nirvn (commented):
@nyalldawson , I've quickly looking through the code. Logic looks sound and clean. Great feature addition.

src/core/qgsmapclippingregion.h
nirvn: ... or:
- IntersectClip
- IntersectFilter

src/core/qgsmapsettings.cpp
m-kuhn: add for "add to an unordered set", append for "append to end of an ordered list"?

m-kuhn:
Thank you, City of Canning. And Nyall.

src/core/qgsmapclippingregion.h
nyalldawson: Agreed, thanks - I've renamed all these enum values for clarity

src/core/qgsmapsettings.cpp
nyalldawson: @m-kuhn -- I like that. I'll leave this one as "add", because even though value is internally stored in a list, the order isn't important and we may change that in future.

nirvn:
🟢

PR #37518, opened 2020-07-01
disable feature picker when no feature is available in expression preview widget

3nids:
following discussion in https://github.com/qgis/QGIS/pull/37139#discussion_r439374026

### DIFF
--- a/src/gui/qgsexpressionpreviewwidget.cpp
+++ b/src/gui/qgsexpressionpreviewwidget.cpp
@@ -55,10 +55,12 @@ void QgsExpressionPreviewWidget::setCurrentFeature( const QgsFeature &feature )
     QgsFeature validFeature( feature );
     validFeature.setValid( true );
     mExpressionContext.setFeature( validFeature );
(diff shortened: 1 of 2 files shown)


nyalldawson:
Why not hide it also?

3nids:
I tend to prefer to disable rather than hiding but no strong opinion.

src/gui/qgsexpressionpreviewwidget.cpp
```diff
@@ -55,10 +55,12 @@ void QgsExpressionPreviewWidget::setCurrentFeature( const QgsFeature &feature )
     QgsFeature validFeature( feature );
     validFeature.setValid( true );
     mExpressionContext.setFeature( validFeature );
+    mFeaturePickerWidget->setEnabled( false );
```
m-kuhn, line 58:
```suggestion
    mFeaturePickerWidget->setTooltip( tr( "No feature was found on this layer to test the expression." ) );
```

Add that and we've got a good reason for not hiding the widget ;)

PR #37538, opened 2020-07-02
invalidate cache / reload data on commit for layer dependencies

3nids:
fixes the issue of WFS layers not being reloaded

supersedes #37475


3nids:
I would like to get some eyes on this one…
ping @troopa81 @nyalldawson @m-kuhn @rouault

3nids:
after discussion, #37475 makes a better job

PR #37543, opened 2020-07-02
rename QgsProject::layer(s)WillBeRemoved to aboutToRemoveLayer(s)

3nids:
the overloading was not used in Python bindings
renaming is preferred since it avoids the usage of overloads

supersedes #37353


nyalldawson (approved):

nyalldawson:
Actually -- looking at the test failure, it seems this isn't possible without breaking API

3nids:
I was not aware of the possibility of using the overloads from PyQt.
I wonder then which signature is used as default. The first implemented? By some ordering of types?
It's quite a bad approach that you are not forced to specify the signature if overloads exist.

So, indeed I don't see any solution except recplicating + deprecating the signals which is a bit too much here.
Pity!

PR #37556, opened 2020-07-03
Show python code sample in PyQGIS doc

DelazJ:
to adjust doc at https://qgis.org/pyqgis/master/core/QgsVectorLayer.html?highlight=code%20sample

### DIFF
--- a/python/core/auto_generated/qgsvectorlayer.sip.in
+++ b/python/core/auto_generated/qgsvectorlayer.sip.in
@@ -34,6 +34,11 @@ manages editing transactions.
 
 Sample usage of the QgsVectorLayer class:
 
+.. code-block:: python
+
+         uri = "point?crs=epsg:4326&field=id:integer"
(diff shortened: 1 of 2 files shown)


3nids:
I think that we should rather show Python only in such case. Python code is show in Doxygen and I guess most C++ devs can translate the code.
But no strong opinion

DelazJ:
I can remove the cpp ones if it's better. I saw #37553 and thought that the code was shown depending on the context and it was better to write both.

3nids:
yes, I was hesitating a bit.
In the other one, the one line is a bit less explicit to me.
But as said, no strong opinion. It might be worth having some stricter guidelines.

3nids:
(sorry, I didn't meant to close)

nyalldawson:
I'm for removing the C++ examples and translating them all to python examples. I agree with @3nids that C++ devs will easily be able to translate python examples to C++ "on the fly".

Would be a nice little project to upgrade all code examples in the docs to python.

3nids:
superseded by #37642

Issue #37567, opened 2020-07-03
LayoutItem's positionning should always be relative to page


olivierdalang:
**Describe the bug**

In the UI, LayoutItem's position is relative to a page (which you can choose using a spinbox). Internally however, the position is stored in absolute coordinates (relative to the whole layout).

A/ For the UI, as LayoutItem have no reference/index to their page, the page is determined dynamically, based on the item's position and the pages geometries. This leads to many non deterministic behaviors, (when page sizes change or are deleted, when saving/importing templates, etc.).

B/ Additionally, while coordinates are page-relative in the spinboxes, they seem so be absolute when using expressions (so that a "10" in the spinbox doesn't equal a "10" in expression unless on the first page).

Instead, I suggest we :
- add a page index variable to LayoutItems, that would act as a hard reference to a page index
- keep the absolute position private, and only expose the relative position in the API and for xml exports.

For backwards compatibility, the index would be 0 (first page) for all items, meaning it should be backwards compatible, as position relative to page 0 is the same as absolute position.

The page index would be changed automatically when moving items around in the canvas, but besides that, it would not be changed dynamically (so that theoretically, you can have an item linked to page 1 that appears on page 2).

(As a further step, we could even imaging allowing to position items relative to other map items instead of just pages.)

**How to Reproduce**

A/ 
- create a new layout with two pages of 200x200mm
- create any MapItem on the first page, roughly in the midde
- change the first page's height to 300mm by typing `300` in the height spinbox
- see that the MapItem ends up on the second page
=> while typing 300, the first page's height goes through 3, so that the MapItem temporarily is on the second page, then gets attached to the second page
(this gets even worse in dynamic setups, for instance in an atlas with a data-defined page height)

B/
- create a new layout with two pages
- create a MapItem at 10;10 on the second page
- set a data-defined value with expression `10` for the Y position
- see that the MapItem gets moved to the first page

**QGIS and OS versions**

3.15 master, Windows

olivierdalang:
See PR https://github.com/qgis/QGIS/pull/37568 that adds an expression to work around this issue

DelazJ:
> => while typing 300, the first page's height goes through 3, so that the MapItem temporarily is on the second page, then gets attached to the second page
(this gets even worse in dynamic setups, for instance in an atlas with a data-defined page height)

This looks quite bad to me. If we cannot attach items to their page, can we not fix this wrong assignment by firing the signal only after the user leaves the widget?

[...]

PR #37568, opened 2020-07-03
add Y page offset expression

olivierdalang:
# Description

This PR adds the `layout_pageoffsets` expression, to be used in the layout composer. The expression returns an array of Y coordinates of the top of each page in the layout, allowing to dynamically position items on pages in a context where page sizes may change.

Example to set an item at 2.5mm from the top of page 2 (for the data-defined position Y of a map item) :
```
@layout_pageoffsets[1] + 2.5
```

This constitutes a workaround for https://github.com/qgis/QGIS/issues/37567


m-kuhn:
@nyalldawson how does that look to you?

olivierdalang:
Not stale

m-kuhn:
this looks good to me, I'd like to merge end of the week if there are no objections.

nyalldawson:
LGTM - but can you add a unit test please?

olivierdalang:
@nyalldawson Done, is it ok like this ?

nyalldawson (approved):

PR #37570, opened 2020-07-03
Fix attributes on copy paste while QGIS stays responsive

signedav:
By using `show()` instead of `exec()` the `QgsFixAttributeDialog` to fix features on Copy/Paste if the constraints are not fulfilled, the user is able to do stuff meanwhile. Like e.g. create a feature on the parent layer by using the + in the Relation Reference Widget.

Since this fixes a behavior that is not nice I like to backport it as well.

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -10222,21 +10222,33 @@ void QgisApp::pasteFromClipboard( QgsMapLayer *destinationLayer )
       context.setMapCanvas( mMapCanvas );
 
       QgsFixAttributeDialog *dialog = new QgsFixAttributeDialog( pasteVectorLayer, invalidFeatures, this, context );
-      int feedback = dialog->exec();
 
-      switch ( feedback )
+      connect( dialog, &QgsFixAttributeDialog::finished, this, [ = ]( int feedback )
       {
-        case QgsFixAttributeDialog::PasteValid:
-          //paste valid and fixed, vanish unfixed
-          newFeatures << validFeatures << dialog->fixedFeatures();
-          break;
-        case QgsFixAttributeDialog::PasteAll:
-          //paste all, even unfixed
-          newFeatures << validFeatures << dialog->fixedFeatures() << dialog->unfixedFeatures();
-          break;
-      }
+        QgsFeatureList features = newFeatures;
+        switch ( feedback )
+        {
+          case QgsFixAttributeDialog::PasteValid:
+            //paste valid and fixed, vanish unfixed
+            features << validFeatures << dialog->fixedFeatures();
+            break;
+          case QgsFixAttributeDialog::PasteAll:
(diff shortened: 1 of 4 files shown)


andreasneumann:
That's an important fix! People have been tracing existing geometries instead of copy/paste because of this behaviour.

src/app/qgisapp.cpp
```diff
@@ -10214,26 +10214,43 @@ void QgisApp::pasteFromClipboard( QgsMapLayer *destinationLayer )
...
-      QgsFixAttributeDialog *dialog = new QgsFixAttributeDialog( pasteVectorLayer, invalidFeatures, this );
-      int feedback = dialog->exec();
+      QgsAttributeEditorContext context( createAttributeEditorContext() );
+      context.setAllowCustomUi( false );
+      context.setFormMode( QgsAttributeEditorContext::StandaloneDialog );
 
-      switch ( feedback )
+      QgsFixAttributeDialog *dialog = new QgsFixAttributeDialog( pasteVectorLayer, invalidFeatures, this, context );
+
+      connect( dialog, &QgsFixAttributeDialog::finished, this, [ = ]( int feedback )
```
nyalldawson, line 10223: This will leak if the dialog is cancelled

m-kuhn: Out of interest.

The [QDialog docs](https://doc.qt.io/qt-5/qdialog.html) mention

> e.g. an OK button to the accept() slot and a Cancel button to the reject() slot.
> [...]
> If the user presses the Esc key in a dialog, QDialog::reject() will be called.

and

> QDialog::finished(int result) : This signal is emitted when the dialog's result code has been set, either by the user or by calling done(), accept(), or reject().

... but I cannot exclude that there's a way to get out of the dialog without deleting it. What would be the proper solution?

nyalldawson: So does pressing cancel in this dialog still result in the paste happening?

m-kuhn: I assumed that's catched in the switch below, but @signedav is in the better position to answer.

signedav: On canceling the dialog, `features` stays empty. Means we call `pasteFeatures` with 0 features and raise there the warning message that no features have  been copied. We could catch this case earlier, but it means to have a duplicate code for the message.

nyalldawson: Ok, fair enough!

m-kuhn (approved):

src/app/qgisapp.cpp
signedav: Thanks @nyalldawson @m-kuhn

andreasneumann:
Any chance we can backport this to 3.10 as well? It is a very annoying problem for us.

PR #37573, opened 2020-07-03
WIP Context menu 'Search selected in PyQGIS docs' in Console input and output

rduivenvoorde:
Was trying out some things in the python console recently and was 'missing' the context help to search for the docs of a certain class or term in qgis.org/pyqgis..

This PR add a 'Search selected in PyQGIS docs' context menu, both in the input and output pane of the Python Console

The selected term (for example 'qgsMapCanvas') will sent you to:

https://qgis.org/pyqgis/master/search.html?q=

So in this case:

https://qgis.org/pyqgis/master/search.html?q=qgsMapCanvas

I started with the output pane (see below) which has an icon in the context menu:

Screenshot-20200703201050-587x296

Then I did the input pane, which nowhere has icons, so skipped it there also:

Screenshot-20200703201005-474x230

Thinking about adding it in the 'Editor' pane too, IF others are positive about this idea.

This is a quick hack so feel free to discuss or throw away the idea :-)

### DIFF
--- a/python/console/console_output.py
+++ b/python/console/console_output.py
@@ -19,8 +19,8 @@ email                : lrssvtml (at) gmail (dot) com
 Some portions of code were taken from https://code.google.com/p/pydee/
 """
 
-from qgis.PyQt.QtCore import Qt, QCoreApplication, QThread, QMetaObject, Q_RETURN_ARG, Q_ARG, QObject, pyqtSlot
-from qgis.PyQt.QtGui import QColor, QFont, QKeySequence, QFontDatabase
+from qgis.PyQt.QtCore import Qt, QCoreApplication, QThread, QMetaObject, Q_RETURN_ARG, Q_ARG, QObject, pyqtSlot, QUrl
+from qgis.PyQt.QtGui import QColor, QFont, QKeySequence, QFontDatabase, QDesktopServices
 from qgis.PyQt.QtWidgets import QGridLayout, QSpacerItem, QSizePolicy, QShortcut, QMenu, QApplication
 from qgis.PyQt.Qsci import QsciScintilla, QsciLexerPython
 from qgis.core import Qgis, QgsApplication, QgsSettings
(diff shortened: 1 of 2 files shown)


python/console/console_output.py
```diff
@@ -332,6 +338,13 @@ def enteredSelected(self):
         self.shell.insertFromDropPaste(cmd)
         self.shell.entered()
 
+    def searchPyQGIS(self):
+        if self.hasSelectedText():
```
nyalldawson, line 342: there is actually a way to get the actual token corresponding to a cursor position from QScintilla -- specifically "GetWordFromPosition"

m-kuhn:
Nice.
Shouldn't we add the icon on the input pane too? I don't think it's a deliberate choice to not showing icons there.

rduivenvoorde:
Looking at the code, it looks like it could have some cleanup... As you can see, the additions to both classes are almost idential/duplicates... And adding this to the editor-class would add a third copy :-(

Had a small email chat with @slarosa and he agreed that it would be nice to add some base class (for I think console_sci console_output and console_editor) as they already share some code which now look like copy/pastes... 
I will probably have a stab into that first.

rduivenvoorde:
See https://github.com/qgis/QGIS/pull/37589

3nids:
isn't it possible to try to point directly to the exact link in the documentation? 
and if not, propose to search.

rduivenvoorde:
@3nids the direct link is (I think) hard, because you selecting a part of text is more general (select 'Point' or 'MapCanvas') and search makes it easy to click to relevant pages.
Or are you proposing a round trip? First check if there is some class, attribute or module with that name, and then 'guess'? which module? I think becomes pretty complex.

Note this PR is now moved to https://github.com/qgis/QGIS/pull/37602 as part of creating a BaseClass

3nids:
I was thinking to look at classes only, and default to text search if it doesn't find anything.
That could stay simple. 

The other thing I had in mind was to hack the terminal, so that whenever you type help(MyClass) it would print the link to the docs.

But both can be implemented :)

m-kuhn:
Also take into account https://github.com/qgis/QGIS/pull/37573#discussion_r449735922

> there is actually a way to get the actual token corresponding to a cursor position from QScintilla -- specifically "GetWordFromPosition"

That will return a name from the cursor position instead of a string

PR #37577, opened 2020-07-03
Random Points in Polygons algorithm - Fix two wrong parameter names in shortHelpString and sort the parameters

havatv:
## Description

Fix two wrong parameter names in the shortHelpString function of the *Random Points in Polygons* algorithm (and sort the parameters).
The updates are only to strings, so should not affect the build.


havatv:
@nyalldawson, sorry about this. Should be an easy merge, though.

nyalldawson (approved):

PR #37586, opened 2020-07-04
Simplify some painter related operations and make safer

nyalldawson:
Use scoped painter state RAII class wherever possible, and move context painter related flag handling to a central place


nirvn (approved):
Nice

PR #37587, opened 2020-07-04
Add unit test for maximum symbol size in legend

mhugent:
Unit test for legend max symbol size (https://github.com/qgis/QGIS/pull/37474)


nyalldawson (approved):
Thanks!!

mhugent:
Should be setMaximumSymbolSize now. I cannot push, it says master is a protected branch...

PR #37588, opened 2020-07-04
[Feature] Export 3D scene feature

NEDJIMAbelgacem:
## Description

This PR is still under development and a part of my summer of code project (sorry if it's too big for reviews :( ).
I started working on exporting 3D map scenes to ,obj files.
Currently in my implementation, the user writes scene name (which will be the .obj file name later) and chooses a folder to export the 3D scene files to, I did it like this because later I will add textures and .mtl files that will contain material information.
The user can also give an integer to specify the resolution of the terrain and specify whether the triangles should look smooth in an app like blender.
Here is my initial UI for the export settings:
 map_export_ui
I made a separate class to manage the export settings because we will have more settings later on.
Here is a screenshot of a terrain and how it is exported to blender with the smoothing option on and off:
Original 3D view:
image
smoothing option off:
image
smoothing option on:
image
The exporter can also handle exporting vector layers and flat terrains:
vector_layer_and_flat_terrain

I tried to make well written code focusing on extensibility.
I want to know if the architecture of classes is good enough, if anyone have suggestion to improve the quality please let me know. Any feedback or suggestion are welcome.

TODO:
- [x] Find good icon for scene exporting action.
- [x] Add normals
- [x] Add texture coordinates
- [x] Manage copying the textures from the scene to a file.
- [x] Handle .mtl specification.
- [x] Add phong material settings handling.
- [ ] Exporting of other types of layers.
- [ ] Permit the user to choose which layers he wants to export and configure them individually.


src/3d/qgs3dmapexportsettings.cpp
```diff
@@ -0,0 +1,26 @@
...
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+#include "qgs3dmapexportsettings.h"
+
+Qgs3DMapExportSettings::Qgs3DMapExportSettings( QObject *parent )
+  : QObject( parent )
+  , mSceneName( "Scene" )
+  , mSceneFolderPath( "~/" )
```
nyalldawson, line 21: You can use QDir::homePath() for this, and it will work nicely on windows

src/3d/qgs3dmapexportsettings.cpp
```diff
@@ -0,0 +1,26 @@
...
+  , mTerrainResolution( 128 )
```
nyalldawson, line 22: These initializers should be moved to the header (C++11) style

src/3d/qgs3dmapscene.cpp
```diff
@@ -763,3 +766,23 @@ void Qgs3DMapScene::updateSceneState()
...
+
+  for ( QgsMapLayer *layer : mLayerEntities.keys() )
+  {
+    exporter.parseEntity( mLayerEntities[layer] );
+  }
+  exporter.parseEntity( mTerrain );
+
+  // TODO: make the scene .obj file created using a function in Qgs3DMapExportSettings
+  QString filePath = exportSettings.sceneFolderPath();
+  if ( !filePath.endsWith( '/' ) ) filePath.push_back( '/' );
```
nyalldawson, line 785: Use the QDir API to construct the file path

src/3d/qgs3dsceneexporter.cpp
```diff
@@ -0,0 +1,493 @@
...
+void Qgs3DSceneExporter::process( QgsTessellatedPolygonGeometry *geom )
+{
+  // Just use position attributes for now
+  processAttribute( geom->mPositionAttribute );
+}
+
+void Qgs3DSceneExporter::saveToFile( const QString &filePath )
+{
+  QFile file( filePath );
+  if ( !file.open( QIODevice::WriteOnly | QIODevice::Text ) )
```
nyalldawson, line 473: Add the truncate flag here too, so that fine overwrites work correctly

src/3d/qgsexportobject.h
```diff
@@ -0,0 +1,71 @@
...
+#include <QVector>
+#include <QVector3D>
+
+/**
+ * @brief The QgsExportObject class
+ * Manages the data of each object of the scene (positions, normals, texture coordinates ...) since each object
+ * \ingroup 3d
+ * \since QGIS 3.16
+ */
+class QgsExportObject : public QObject
```
nyalldawson, line 30: This class name is quite generic -- would Qgs3DExportObject work?

src/ui/3d/map3dexportwidget.ui
```diff
@@ -0,0 +1,89 @@
...
+     </item>
+     <item row="1" column="0">
+      <widget class="QLabel" name="folderLAbel">
+       <property name="text">
+        <string>Folder</string>
+       </property>
+      </widget>
+     </item>
+     <item row="0" column="1" colspan="2">
+      <widget class="QLineEdit" name="sceneNameLineEdit"/>
```
nyalldawson, line 34: You should use QgsFileWidget here, that's got lots of built in logic for selection of folders

nyalldawson (commented):
Nice work! Some minor comments on the general coding here (not a review of the 3d component, I'll let Martin do that!), but nothing major

NEDJIMAbelgacem:
> Nice work! Some minor comments on the general coding here (not a review of the 3d component, I'll let Martin do that!), but nothing major

Thanks!
feel free to give me your thoughts of the code structure even if that's not the a full review it will help me fix things before the code gets even bigger

src/3d/CMakeLists.txt
```diff
@@ -110,6 +111,9 @@ SET(QGIS_3D_HDRS
   chunks/qgschunkloader_p.h
   chunks/qgschunkqueuejob_p.h
   mesh/qgsmeshterraingenerator.h
+  qgs3dsceneexporter.h
+  qgs3dmapexportsettings.cpp
```
wonder-sk, line 115: cpp file within headers section?

src/3d/qgs3dexportobject.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+  for ( int i = 0; i < positionsBuffer.size() / 3; ++i )
+  {
+    mIndexes << i + 1;
+  }
+}
+
+void Qgs3DExportObject::setupPositionCoordinates( const QVector<float> &positionsBuffer, const QVector<unsigned int> &faceIndex, float scale, const QVector3D translation )
+{
+  // TODO: delete vertices that are not used
```
wonder-sk, line 49: Does it happen that we have vertices that do not get used?

src/3d/qgs3dexportobject.cpp
```diff
@@ -0,0 +1,159 @@
...
+}
+
+void Qgs3DExportObject::saveTo( QTextStream &out, float scale, const QVector3D &center )
+{
+  // Set groups
+  // turns out grouping doest work as expected in blender
+
+  // smoothen edges
+  if ( mSmoothEdges ) out << "s on\n";
+  else out << "s off\n";
```
wonder-sk, line 97: code style: it's strongly preferred in qgis code to have if/else statements on separate lines:
```
if ( a )
  do_a();
else
  do_b();
```

src/3d/qgs3dexportobject.cpp
```diff
@@ -0,0 +1,159 @@
...
+}
+
+void Qgs3DExportObject::setupPositionCoordinates( const QVector<float> &positionsBuffer, const QVector<unsigned int> &faceIndex, float scale, const QVector3D translation )
+{
+  // TODO: delete vertices that are not used
+  for ( int i = 0; i < positionsBuffer.size(); i += 3 )
+  {
+    for ( int j = 0; j < 3; ++j )
+    {
+      mVertxPosition << positionsBuffer[i + j] * scale + translation[j];
```
wonder-sk, line 54: looks like a typo mVertxPosition -> mVertexPosition

src/3d/qgs3dexportobject.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+  if ( !hasNormals && !mNormals.empty() )
+  {
+    qDebug() << "WARNING: vertex normals count and vertex positions count are different";
+  }
+  int verticesCount = mVertxPosition.size() / 3;
+
+  auto getVertexIndex = [&]( int i ) -> QString
+  {
+    int negativeIndex = -1 - ( verticesCount - i );
```
wonder-sk, line 131: why are the indices negative here? something that .obj format requires?

src/3d/qgs3dmapexportsettings.h.bom
```diff
@@ -0,0 +1,45 @@
+﻿/***************************************************************************
```
wonder-sk, line 1: looks like this file has been added by mistake?

src/3d/qgs3dsceneexporter.cpp
```diff
@@ -0,0 +1,524 @@
...
+        vArr[k] = data.at( i + j + k );
+      }
+      result.push_back( v );
+    }
+  }
+  return result;
+}
+
+template<>
+QVector<unsigned int> getAttributeData<unsigned int>( Qt3DRender::QAttribute *attribute )
```
wonder-sk, line 81: Maybe it would be better to rename this function? From what I understand it has different semantics that the original template function getAttributeData() which also iterates over vertices, while this one does not ... assuming it is used for index buffer?

src/3d/qgs3dsceneexporter.cpp
```diff
@@ -0,0 +1,524 @@
...
+    {
+      vArr[k] = data.at( i + k );
+    }
+    result.push_back( v );
+  }
+
+  return result;
+}
+
+QVector<float> createPlaneVertexData( float w, float h, const QSize &resolution )
```
wonder-sk, line 101: do we need these createPlane*Data() methods? Can't we just extract the contents of attributes of the plane geometry created for terrain tiles?

src/3d/qgs3dsceneexporter.cpp
```diff
@@ -0,0 +1,524 @@
...
+  return tileEntity;
+}
+
+QgsTerrainTileEntity *Qgs3DSceneExporter::getDemTerrainEntity( QgsTerrainEntity *terrain, QgsChunkNode *node )
+{
+  QgsTerrainTileEntity *tileEntity = nullptr;
+  // Just create a new tile (we don't need to export exact level of details as in the scene)
+  // create the entity synchronously and then it will be deleted once our scene exporter instance is deallocated
+  const Qgs3DMapSettings &map = terrain->map3D();
+  QgsDemTerrainGenerator *generator = static_cast<QgsDemTerrainGenerator *>( map.terrainGenerator() );
```
wonder-sk, line 333: I think it would be safer to make your own copy of the terrain generator... imagine that during the export there may be loading of terrain tiles in a background thread - and we may modify the resolution of that generator while it is doing some work.

src/3d/qgs3dsceneexporter.cpp
```diff
@@ -0,0 +1,524 @@
...
+      }
+    }
+  }
+
+  QgsTerrainGenerator *generator = settings.terrainGenerator();
+  QgsTerrainTileEntity *terrainTile = nullptr;
+  QgsTerrainTextureGenerator *textureGenerator = terrain->textureGenerator();
+  QSize oldResolution = textureGenerator->textureSize();
+  textureGenerator->setTextureSize( QSize( mTerrainTextureResolution, mTerrainTextureResolution ) );
+  switch ( generator->type() )
```
wonder-sk, line 290: I think it may be a cleaner solution to add a new virtual function to QgsTerrainGenerator class, which run the export code specific to each generator. That way, all the details would stay together with the other code of that generator...

src/3d/qgs3dsceneexporter.cpp
```diff
@@ -0,0 +1,524 @@
...
+    {
+      mesh = meshTyped;
+      continue;
+    }
+    Qt3DCore::QTransform *transformTyped = qobject_cast<Qt3DCore::QTransform *>( component );
+    if ( transformTyped != nullptr )
+    {
+      transform = transformTyped;
+    }
+  }
```
wonder-sk, line 359: It would be good to have some utility functions e.g. findGeometryRendererComponent(), findTransformComponent() to make code shorter here and in parseDemTile()...

src/3d/qgs3dsceneexporter.cpp.bom
```diff
@@ -0,0 +1,522 @@
+﻿/***************************************************************************
```
wonder-sk, line 1: This file was probably checked in by mistake

src/3d/qgstessellatedpolygongeometry.h
```diff
@@ -79,12 +81,15 @@ class QgsTessellatedPolygonGeometry : public Qt3DRender::QGeometry
...
     QgsFeatureId triangleIndexToFeatureId( uint triangleIndex ) const;
 
+    friend class Qgs3DSceneExporter;
   private:
 
     Qt3DRender::QAttribute *mPositionAttribute = nullptr;
     Qt3DRender::QAttribute *mNormalAttribute = nullptr;
     Qt3DRender::QBuffer *mVertexBuffer = nullptr;
 
+    QVector<float> mVertexData;
```
wonder-sk, line 91: do we need this new member?

src/3d/terrain/qgsdemterraintileloader_p.cpp
```diff
@@ -52,7 +52,7 @@ static void _heightMapMinMax( const QByteArray &heightMap, float &zMin, float &z
 }
 
 
-QgsDemTerrainTileLoader::QgsDemTerrainTileLoader( QgsTerrainEntity *terrain, QgsChunkNode *node )
+QgsDemTerrainTileLoader::QgsDemTerrainTileLoader( QgsTerrainEntity *terrain, QgsChunkNode *node, bool loadSynchronously )
```
wonder-sk, line 55: Hmm I would like to avoid having a new `loadSynchornously` flag in terrain tile loaders. I would suggest that instead we have a virtual method `waitForFinished()` that would block until the loading has not finished. This should keep the code cleaner and easier to test because there will be fewer possible code paths.

src/3d/terrain/qgsterraintextureimage_p.h
```diff
@@ -62,6 +62,8 @@ class QgsTerrainTextureImage : public Qt3DRender::QAbstractTextureImage
     //! Returns debug information (normally map tile coordinates)
     QString imageDebugText() const { return mDebugText; }
 
+    QImage *getImage() { return &mImage; }
```
wonder-sk, line 65: is this used anywhere?

src/3d/terrain/qgsterraintileloader_p.cpp
```diff
@@ -64,6 +64,11 @@ void QgsTerrainTileLoader::loadTexture()
   mTextureJobId = mTerrain->textureGenerator()->render( mExtentMapCrs, mNode->tileId(), mTileDebugText );
 }
 
+void QgsTerrainTileLoader::loadTextureSynchronously()
```
wonder-sk, line 67: As suggested above - I would like to avoid this synchronous mode altogether and have waitForFinished() instead which would handle that.

src/app/3d/qgs3dmapcanvasdockwidget.cpp
```diff
@@ -291,6 +297,34 @@ void Qgs3DMapCanvasDockWidget::configure()
...
+  QgsMap3DExportWidget *w = new QgsMap3DExportWidget( mCanvas->scene(), exportSettings, &dlg );
+
+  QDialogButtonBox *buttons = new QDialogButtonBox( QDialogButtonBox::Ok | QDialogButtonBox::Cancel, &dlg );
+
+  connect( buttons, &QDialogButtonBox::accepted, &dlg, &QDialog::accept );
+  connect( buttons, &QDialogButtonBox::rejected, &dlg, &QDialog::reject );
+  connect( &dlg, &QDialog::accepted, [ = ]()
+  {
+    w->exportScene();
+  } );
```
wonder-sk, line 319: It would be easier to read if instead of having this slot with a lambda function the `w->exportScene()` call would be on the last line of this function... because if dlg.exec() returns true, that means the dialog has been accepted.

src/app/3d/qgs3dmapcanvasdockwidget.cpp
```diff
@@ -291,6 +297,34 @@ void Qgs3DMapCanvasDockWidget::configure()
...
+void Qgs3DMapCanvasDockWidget::exportScene()
+{
+
+  QDialog dlg;
+  dlg.setWindowTitle( tr( "Export 3D Scene" ) );
+  dlg.setObjectName( QStringLiteral( "3DSceneExportDialog" ) );
+  dlg.setMinimumSize( 380, 460 );
+//  QgsGui::instance()->enableAutoGeometryRestore( &dlg );
+
+  Qgs3DMapExportSettings *exportSettings = new Qgs3DMapExportSettings( &dlg );
```
wonder-sk, line 309: Is this deleted anywhere? Maybe better to just use stack-based local variable: `Qgs3DMapExportSettings exportSettings( &dlg );`

src/3d/qgs3dsceneexporter.cpp
```diff
@@ -0,0 +1,524 @@
...
+  QgsTerrainTextureGenerator *textureGenerator = terrain->textureGenerator();
+  QSize oldResolution = textureGenerator->textureSize();
+  textureGenerator->setTextureSize( QSize( mTerrainTextureResolution, mTerrainTextureResolution ) );
+  switch ( generator->type() )
+  {
+    case QgsTerrainGenerator::Dem:
+      for ( QgsChunkNode *node : leafs )
+      {
+        terrainTile = getDemTerrainEntity( terrain, node );
+        this->parseDemTile( terrainTile, textureGenerator );
```
wonder-sk, line 296: Are these tiles getting deleted anywhere?

wonder-sk (commented):
Nice work!

src/3d/CMakeLists.txt
NEDJIMAbelgacem: Sorry I added it by mistake and forgot it there :facepalm:

src/3d/qgs3dexportobject.cpp
NEDJIMAbelgacem: I don't think that happens actually, I wrote this when I was looking for why invalid vertices are returned. I think that could happen if the height map is not square and we end up storing vertex coordinates for invalid faces even though we don't actually use it.
This code here detects invalid faces:
```
    if ( faceIndex[i] == faceIndex[i + 1] && faceIndex[i + 1] == faceIndex[i + 2] ) continue;
```

src/3d/qgs3dmapexportsettings.h.bom
NEDJIMAbelgacem: Yes I deleted the file

src/3d/qgs3dexportobject.cpp
NEDJIMAbelgacem: In the .obj spec negative indices refer to the last defined vertices (-1 for the last vertice).
This helps keeping each object data separated

src/3d/qgs3dsceneexporter.cpp
NEDJIMAbelgacem: Yes I renamed it to getIndexData

src/3d/qgs3dsceneexporter.cpp
NEDJIMAbelgacem: I didn't think about that. I added ->clone()

src/app/3d/qgs3dmapcanvasdockwidget.cpp
NEDJIMAbelgacem: I made the dialog object a parent of that class, so it is deleted when the dlg object is deallocated. I guess I was over engineering here.

src/3d/qgs3dsceneexporter.cpp
NEDJIMAbelgacem: Yes I made the Qgs3DSceneExporter class derive from a QObject so I can make every generated object a child of the scene exporter and it will be deleted automatically after.
It's like I made it into a policy in my mind whenever I make a new object, it should be a child of the object where it is created. That seems like a good idea for our use case.

src/app/3d/qgs3dmapcanvasdockwidget.cpp
NEDJIMAbelgacem: For some reason I thought that we will want the dialog to not block the whole app in the future so I made it like that

src/3d/qgs3dsceneexporter.cpp
NEDJIMAbelgacem: At first I carelessly added ->clone() but that actually doesn't work because the height map generator gets its terrain resolution parameter from map.terrainGenerator()
This is the code from QgsDemTerrainTileLoader constructor:
```
QgsDemTerrainGenerator *generator = static_cast<QgsDemTerrainGenerator *>( map.terrainGenerator() );
heightMapGenerator = generator->heightMapGenerator();
mSkirtHeight = generator->skirtHeight();
```

src/3d/qgs3dsceneexporter.cpp
NEDJIMAbelgacem: I made them into one function template named findTypedComponent

src/3d/qgs3dexportobject.cpp
```diff
@@ -92,6 +80,18 @@ void Qgs3DExportObject::setupTextureCoordinates( const QVector<float> &texturesB
   mTexturesUV << texturesBuffer;
 }
 
+void Qgs3DExportObject::setupPhongMaterial( const QgsPhongMaterialSettings &material )
+{
+  QColor diffuse = material.diffuse();
+  QColor specular = material.specular();
+  QColor ambient = material.ambient();
+  float shininess = material.shininess();
+  setMaterialParameter( QString( "Kd" ), QString( "%1 %2 %3" ).arg( diffuse.redF() ).arg( diffuse.greenF() ).arg( diffuse.blueF() ) );
```
elpaso, line 89:
```suggestion
  setMaterialParameter( QStringLiteral( "Kd" ), QStringLiteral( "%1 %2 %3" ).arg( diffuse.redF() ).arg( diffuse.greenF() ).arg( diffuse.blueF() ) );
```

For const rvalue strings it's better and faster to use `QStringLiteral` instead of `QString`, can you please check other occurrences too?

NEDJIMAbelgacem: Okay, thanks for the suggestion. I will try changing that whenever I notice myself doing it.
By the way should QStringLiteral be used for QString and const char* comparison too?

src/3d/qgs3dexportobject.cpp
```diff
@@ -0,0 +1,205 @@
...
+    for ( int j = 0; j < 3; ++j )
+      vertexIndex << faceIndex[i + j] + 1;
+  }
+}
+
+Qgs3DExportObject::Qgs3DExportObject( const QString &name, const QString &parentName, QObject *parent )
+  : QObject( parent )
+  , mName( name )
+  , mParentName( parentName )
+  , mSmoothEdges( false )
```
nyalldawson, line 49: Move initialization to header (c++11 style)

src/3d/qgs3dexportobject.cpp
```diff
@@ -0,0 +1,205 @@
...
+{
+}
+
+void Qgs3DExportObject::setupPositionCoordinates( const QVector<float> &positionsBuffer, float scale, const QVector3D translation )
```
nyalldawson, line 53:
```suggestion
void Qgs3DExportObject::setupPositionCoordinates( const QVector<float> &positionsBuffer, float scale, const QVector3D& translation )
```

src/3d/qgs3dexportobject.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+void Qgs3DExportObject::setupTextureCoordinates( const QVector<float> &texturesBuffer )
+{
+  mTexturesUV << texturesBuffer;
+}
+
+void Qgs3DExportObject::setupMaterial( QgsAbstractMaterialSettings *material )
+{
+  QMap<QString, QString> parameters = material->toExportParameters();
+  for ( QString &key : parameters.keys() ) setMaterialParameter( key, parameters[key] );
```
nyalldawson, line 88:
```suggestion
  const QMap<QString, QString> parameters = material->toExportParameters();
  for ( auto it = parameters.begin(); it != parameters.end(); ++it )
  {
     setMaterialParameter( it.key(), it.value() );
  }
```

(this is considerable more efficient -- it avoids the initial loop over the container to create the keys list, and then the n x map lookups for each key in the map)

src/3d/qgs3dexportobject.cpp
```diff
@@ -0,0 +1,205 @@
...
+    }
+  }
+
+  bool hasTextures = mTexturesUV.size() == mVertexPosition.size() / 3 * 2;
+  // if the object has normals then the normals and positions buffers should be the same size
+  bool hasNormals = mNormals.size() == mVertexPosition.size();
+
+  if ( !hasNormals && !mNormals.empty() )
+  {
+    qDebug() << "WARNING: vertex normals count and vertex positions count are different";
```
nyalldawson, line 143: Use QgsDebugMsg instead of qDebug()

src/3d/qgs3dexportobject.cpp
```diff
@@ -0,0 +1,205 @@
...
+  if ( mTexturesUV.size() != 0 && !mTextureImage.isNull() )
+  {
+    QString filePath = QDir( folderPath ).filePath( materialName + ".jpg" );
+    mTextureImage.save( filePath, "JPG" );
+    mtlOut << "\tmap_Kd " << materialName << ".jpg" << "\n";
+  }
+  for ( QString key : mMaterialParameters.keys() )
+  {
+    mtlOut << "\t" << key << " " << mMaterialParameters[key] << "\n";
+    qDebug() << "\t" << key << " " << mMaterialParameters[key] << "\n";
```
nyalldawson, line 201:
```suggestion
    QgsDebugMsgLevel( QStringLiteral("\t%1 %2").arg( key, mMaterialParameters[key] ), 3)";
```

src/3d/qgs3dexportobject.h
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * \brief The Qgs3DExportObject class
+ * Manages the data of each object of the scene (positions, normals, texture coordinates ...) since each object
+ * \ingroup 3d
+ * \since QGIS 3.16
+ */
+class Qgs3DExportObject : public QObject
+{
+    Q_OBJECT
```
nyalldawson, line 37: Why is this a QObject?

src/3d/qgs3dmapexportsettings.h
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * \brief The Qgs3DMapExportSettings class
+ * Manages the various settings the user can choose from when exorting a 3D scene
+ * \ingroup 3d
+ * \since QGIS 3.16
+ */
+class _3D_EXPORT Qgs3DMapExportSettings : public QObject
+{
+    Q_OBJECT
```
nyalldawson, line 33: This shouldn't be a QObject either (unless I'm missing something?)

src/3d/qgs3dmapexportsettings.h
```diff
@@ -0,0 +1,86 @@
...
+
+#include "qgis_3d.h"
+
+#include <QString>
+#include <QObject>
+#include <QDir>
+
+/**
+ * \brief The Qgs3DMapExportSettings class
+ * Manages the various settings the user can choose from when exorting a 3D scene
```
nyalldawson, line 27:
```suggestion
 * Manages the various settings the user can choose from when exporting a 3D scene
```

src/3d/qgs3dmapscene.cpp
```diff
@@ -856,3 +860,54 @@ void Qgs3DMapScene::updateSceneState()
...
+  exporter.setSmoothEdges( exportSettings.smoothEdges() );
+  exporter.setExportNormals( exportSettings.exportNormals() );
+  exporter.setExportTextures( exportSettings.exportTextures() );
+  exporter.setTerrainTextureResolution( exportSettings.terrainTextureResolution() );
+  exporter.setScale( exportSettings.scale() );
+
+  for ( QgsMapLayer *layer : mLayerEntities.keys() )
+  {
+    QgsMapLayerType layerType =  layer->type();
+    Qt3DCore::QEntity *rootEntity = mLayerEntities[layer];
```
nyalldawson, line 879:
```suggestion
  for ( auto it = mLayerEntities.constBegin(); it != mLayerEntities.constEnd(); ++it )
  {
    QgsMapLayer* layer = it.key();
    QgsMapLayerType layerType =  layer->type();
    Qt3DCore::QEntity *rootEntity = it.value();
```

src/3d/qgs3dmapscene.cpp
```diff
@@ -856,3 +860,54 @@ void Qgs3DMapScene::updateSceneState()
...
+        if ( !exporter.parseVectorLayerEntity( rootEntity, qobject_cast<QgsVectorLayer *>( layer ) ) )
+          notParsedLayers.push_back( layer->name() );
+        break;
+      case QgsMapLayerType::RasterLayer:
+        notParsedLayers.push_back( layer->name() );
+        break;
+      case QgsMapLayerType::PluginLayer:
+        notParsedLayers.push_back( layer->name() );
+        break;
+      case QgsMapLayerType::MeshLayer:      //!< Added in 3.2
```
nyalldawson, line 892:
```suggestion
      case QgsMapLayerType::MeshLayer:
```

src/3d/qgs3dmapscene.cpp
```diff
@@ -856,3 +860,54 @@ void Qgs3DMapScene::updateSceneState()
...
+        notParsedLayers.push_back( layer->name() );
+        break;
+      case QgsMapLayerType::VectorTileLayer: //!< Added in 3.14
```
nyalldawson, line 895:
```suggestion
      case QgsMapLayerType::VectorTileLayer:
```

src/3d/qgs3dmapscene.cpp
```diff
@@ -856,3 +860,54 @@ void Qgs3DMapScene::updateSceneState()
...
+        notParsedLayers.push_back( layer->name() );
+        break;
+    }
+  }
+
+  if ( mTerrain != nullptr )
```
nyalldawson, line 901:
```suggestion
  if ( mTerrain )
```

src/3d/qgs3dmapscene.cpp
```diff
@@ -856,3 +860,54 @@ void Qgs3DMapScene::updateSceneState()
...
+    exporter.parseTerrain( mTerrain, "Terrain" );
+
+  exporter.save( exportSettings.sceneName(), exportSettings.sceneFolderPath() );
+
+  if ( notParsedLayers.size() != 0 )
```
nyalldawson, line 906:
```suggestion
  if ( !notParsedLayers.empty() )
```

src/3d/qgs3dmapscene.cpp
```diff
@@ -856,3 +860,54 @@ void Qgs3DMapScene::updateSceneState()
...
+  {
+    QString message = "The following layers were not exported: \n";
```
nyalldawson, line 908:
```suggestion
    QString message = QObject::tr("The following layers were not exported: ") + '\n';
```

src/3d/qgs3dmapscene.cpp
```diff
@@ -856,3 +860,54 @@ void Qgs3DMapScene::updateSceneState()
...
+    for ( QString layerName : notParsedLayers )
```
nyalldawson, line 909:
```suggestion
    for ( const QString &layerName : notParsedLayers )
```

src/3d/qgs3dmapscene.cpp
```diff
@@ -856,3 +860,54 @@ void Qgs3DMapScene::updateSceneState()
...
+      message += layerName + "\n";
+    QgsMessageOutput::showMessage( "3D exporter warning", message, QgsMessageOutput::MessageText );
```
nyalldawson, line 911:
```suggestion
    QgsMessageOutput::showMessage( tr("3D exporter warning"), message, QgsMessageOutput::MessageText );
```

src/3d/qgs3dsceneexporter.cpp
```diff
@@ -0,0 +1,804 @@
...
+QVector<T> getAttributeData( Qt3DRender::QAttribute *attribute, QByteArray data )
+{
+  uint bytesOffset = attribute->byteOffset();
+  uint bytesStride = attribute->byteStride();
+  uint vertexSize = attribute->vertexSize();
+  QVector<T> result;
+
+  if ( bytesStride == 0 )
+  {
+    qDebug() << "WARNING: bytesStride==0 at " << __FILE__ << ":" << __LINE__;
```
nyalldawson, line 92: qDebug() -> QgsDebugMsg

src/3d/qgs3dsceneexporter.cpp
```diff
@@ -0,0 +1,804 @@
...
+    return result;
+  }
+
+  for ( int i = bytesOffset; i < data.size(); i += bytesStride )
+  {
+    for ( unsigned int j = 0; j < vertexSize * sizeof( T ); j += sizeof( T ) )
+    {
+      // maybe a problem with indienness can happen?
```
nyalldawson, line 100:
```suggestion
      // maybe a problem with endianess can happen?
```

src/3d/qgs3dsceneexporter.cpp
```diff
@@ -0,0 +1,804 @@
...
+    case Qt3DRender::QAttribute::VertexBaseType::Short:
+      return _getIndexDataImplementation<short>( data );
+    case Qt3DRender::QAttribute::VertexBaseType::UnsignedShort:
+      return _getIndexDataImplementation<ushort>( data );
+    case Qt3DRender::QAttribute::VertexBaseType::Byte:
+      return _getIndexDataImplementation<char>( data );
+    case Qt3DRender::QAttribute::VertexBaseType::UnsignedByte:
+      return _getIndexDataImplementation<uchar>( data );
+    default:
+      qDebug() << "WARNING: Probably trying to get index data using an attribute that has vertex data at " << __FILE__ << ":" << __LINE__;
```
nyalldawson, line 146: QgsDebugMsg (elsewhere here too)

src/3d/qgs3dsceneexporter.cpp
```diff
@@ -0,0 +1,804 @@
...
+}
+
+Qgs3DSceneExporter::Qgs3DSceneExporter( Qt3DCore::QNode *parent )
+  : Qt3DCore::QEntity( parent )
+  , mSmoothEdges( false )
+  , mTerrainResolution( 128 )
+  , mExportNormals( true )
+  , mExportTextures( false )
+  , mTerrainTextureResolution( 512 )
+  , mScale( 1.0f )
```
nyalldawson, line 173: Move initialization to header

src/3d/qgs3dsceneexporter.cpp
```diff
@@ -0,0 +1,804 @@
...
+        mObjects.push_back( object );
+      }
+      return true;
+    }
+    else
+    {
+      QgsVectorLayer3DRenderer *vectorLayerRenderer = dynamic_cast< QgsVectorLayer3DRenderer *>( abstractVectorRenderer );
+      const QgsAbstract3DSymbol *symbol = vectorLayerRenderer->symbol();
+      QString symbolType = symbol->type();
+      if ( symbolType == "polygon" )
```
nyalldawson, line 234: This logic should probably be moved to the symbol subclasses themselves

src/3d/qgs3dsceneexporter.h
```diff
@@ -0,0 +1,138 @@
...
+    void parseDemTile( QgsTerrainTileEntity *tileEntity, const QString &layerName );
+    //! Constructs a Qgs3DExportObject from the flat tile entity
+    void parseFlatTile( QgsTerrainTileEntity *tileEntity, const QString &layerName );
+    //! Constructs a Qgs3DExportObject from the mesh terrain entity
+    void parseMeshTile( QgsTerrainTileEntity *meshEntity, const QString &layerName );
+
+    QString getObjectName( const QString &name );
+  private:
+    QMap<QString, int> usedObjectNamesCounter;
+    QVector<Qgs3DExportObject *> mObjects;
```
nyalldawson, line 128: I can't see where mObjects is cleaned up -- I believe the contents of this vector are leaked. Perhaps a destructor should be added which deletes all these objects?

src/3d/qgsabstractmaterialsettings.h
```diff
@@ -115,6 +115,11 @@ class _3D_EXPORT QgsAbstractMaterialSettings SIP_ABSTRACT
      */
     virtual QgsLineMaterial *toLineMaterial( const QgsMaterialContext &context ) const = 0 SIP_FACTORY;
 
+    /**
+     * Returns the parameters to be exported to .mtl file
+     */
+    virtual QMap<QString, QString> toExportParameters() const = 0 SIP_FACTORY;
```
nyalldawson, line 121:
```suggestion
    virtual QMap<QString, QString> toExportParameters() const = 0;
```

The factory annotation is only required for methods which return a pointer along with ownership (or a list of pointers). This function is returning values directly, so it's not required here.

src/3d/qgsphongmaterialsettings.cpp
```diff
@@ -141,6 +143,16 @@ QgsLineMaterial *QgsPhongMaterialSettings::toLineMaterial( const QgsMaterialCont
   return mat;
 }
 
+QMap<QString, QString> QgsPhongMaterialSettings::toExportParameters() const
+{
+  QMap<QString, QString> parameters;
+  parameters[ QString( "Kd" ) ] = QStringLiteral( "%1 %2 %3" ).arg( mDiffuse.redF() ).arg( mDiffuse.greenF() ).arg( mDiffuse.blueF() );
+  parameters[ QString( "Ka" ) ] = QStringLiteral( "%1 %2 %3" ).arg( mAmbient.redF() ).arg( mAmbient.greenF() ).arg( mAmbient.blueF() );
+  parameters[ QString( "Ks" ) ] = QStringLiteral( "%1 %2 %3" ).arg( mSpecular.redF() ).arg( mSpecular.greenF() ).arg( mSpecular.blueF() );
+  parameters[ QString( "Ns" ) ] = QStringLiteral( "%1" ).arg( mShininess );
```
nyalldawson, line 152:
```suggestion
  parameters[ QStringLiteral( "Kd" ) ] = QStringLiteral( "%1 %2 %3" ).arg( mDiffuse.redF() ).arg( mDiffuse.greenF() ).arg( mDiffuse.blueF() );
  parameters[ QStringLiteral( "Ka" ) ] = QStringLiteral( "%1 %2 %3" ).arg( mAmbient.redF() ).arg( mAmbient.greenF() ).arg( mAmbient.blueF() );
  parameters[ QStringLiteral( "Ks" ) ] = QStringLiteral( "%1 %2 %3" ).arg( mSpecular.redF() ).arg( mSpecular.greenF() ).arg( mSpecular.blueF() );
  parameters[ QStringLiteral( "Ns" ) ] = QStringLiteral( "%1" ).arg( mShininess );
```

nyalldawson (commented):
Some minor code convention notes and qt related coding tips, otherwise looks good!

src/3d/qgs3dmapexportsettings.h
NEDJIMAbelgacem: I made the Qgs3DMapExportSettings behave as a parent of  Qgs3DExportObject so that every Qgs3DExportObject instantiated using a Qgs3DMapExportSettings will be deallocated once the parent Qgs3DMapExportSettings is deleted.

src/3d/qgs3dsceneexporter.cpp
NEDJIMAbelgacem: I forgot I wrote this and was like "why did I write this" :laughing:

src/3d/qgs3dmapexportsettings.h
nyalldawson: That's using a sledgehammer approach :laugh: If you're not using signals/slots/some complex shared ownership, it's much clearer to just use plain old pointers and clean them up in the destructor (or, the modern approach of a std::vector containing std::unique_ptr, so that they'll be cleaned up automatically)

NEDJIMAbelgacem: Yes this class shouldn't be QObject. It is only necessary for the Qgs3DSceneExporter to inherit from Qt3DCore::QEntity so that every QEntity we create dynamically will be deallocated properly

src/3d/qgs3dsceneexporter.h
NEDJIMAbelgacem: I made each object inside mObjects a child of QgsSceneExporter (that's why I made the class inherit QObject).
Now I changed that so that mObjects vector are deleted inside the destructor

wonder-sk (approved):
I am happy with the current implementation - great work @NEDJIMAbelgacem  :tada:

nirvn:
Wouhou!

PR #37589, opened 2020-07-04
WIP Add simple BaseClass to Python Console classes

rduivenvoorde:
This is a Work In Progress to do some cleanup in the python code for the Python Console (output, (line)-input and editor.
It appears there is some duplicate code in those files, so I start here to add a BaseClass and hopefully are able to move some of that duplicate code to this class.

Starting with those lists of constant colors :-)


rduivenvoorde:
I think setLexers is the next candidate:

Screenshot-20200704164205-2561x1169

If you are into this code, feel free to propose other parts/candidates...

rduivenvoorde:
Sorry, should not be a branch here... but in my personal repo. 

Will cleanup

rduivenvoorde:
Moved to: https://github.com/qgis/QGIS/pull/37602

PR #37594, opened 2020-07-05
Expose option to control PDF image compression method when exporting layouts to PDF

nyalldawson:
Options are for Lossy compression, which is the default JPEG compression
used, and Lossless compression (which creates bigger files in most
cases, but is much more suitable for professional printing outputs
or for post-production in Illustrator/etc)

The Bad news: this option is available in builds based on Qt 5.13 or later.

I've personally wanted this option for a LONG LONG time (ever since I started using QGIS), since the default JPEG compression used in the outputs is really unsuitable for high quality production work. It's only been supported since Qt 5.13, but even then Qt didn't announce this new functionality in any way at all (it's completely absent on the official qt changelogs!). To me this is a much more exciting Qt enhancement then all the Qt embedded stuff upstream focuses on these days :wink:

### DIFF
--- a/python/core/auto_generated/layout/qgslayoutrendercontext.sip.in
+++ b/python/core/auto_generated/layout/qgslayoutrendercontext.sip.in
@@ -32,6 +32,7 @@ Stores information relating to the current rendering settings for a layout.
       FlagDrawSelection,
       FlagDisableTiledRasterLayerRenders,
       FlagRenderLabelsByMapLayer,
+      FlagLosslessImageRendering,
     };
     typedef QFlags<QgsLayoutRenderContext::Flag> Flags;
 
--- a/python/core/auto_generated/qgsmapsettings.sip.in
+++ b/python/core/auto_generated/qgsmapsettings.sip.in
@@ -345,6 +345,7 @@ Gets color that is used for drawing of selected vector features
       RenderPartialOutput,
       RenderPreviewJob,
       RenderBlocking,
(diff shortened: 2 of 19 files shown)


nirvn:
Since 5.14 or 5.13? Your description says the former while your code says the latter :)

nirvn:
That'll have to be in the save map as PDF dialog too, really cool.

nyalldawson:
Ah, you're correct. 5.13.

nirvn (approved):

agiudiceandrea:
Hi @nyalldawson, I had a look at an issue about the layer rendering opacity not taken into account when exporting a layout to PDF https://github.com/qgis/QGIS/issues/42698.
It seems to me the regression was introduced in 3.15 and not backported to 3.10 and 3.14 and is Qt version dependant.
QGIS 3.16.0 is the first version with which the issue occurs. The issue occurs with QGIS 3.16 and 3.18 using Qt 5.15.2, while it doesn't occur using Qt 5.11.2.
So, may this new feature (introduced in 3.15, not backported, Qt version dependant) have triggered the regression?

rouault:
> So, may this new feature (introduced in 3.15, not backported, Qt version dependant) have triggered the regression?

I doubt so. My understanding of that PR is that it only changes behaviour with QT >= 5.13

agiudiceandrea:
@rouault you are right... please see https://github.com/qgis/QGIS/issues/42698

PR #37595, opened 2020-07-05
Load processing results to layer group (optional)

gacarrillor:
## Description

When running several Processing algorithms (with `Open output file after running algorithm` checked), it is common to end up with an untidy QGIS project. Manual re-arrangement of layers is then required to preserve the original layer tree configuration, which may take some time depending on user expertise.

This PR adds the possibility to set a group name in Processing options to contain all Processing result layers, so that they can be easily located in the layer tree and then turned on/off, removed, exported, among others. 

Screencast:

processing_results2

### DIFF
--- a/python/plugins/processing/core/ProcessingConfig.py
+++ b/python/plugins/processing/core/ProcessingConfig.py
@@ -62,6 +62,7 @@ class ProcessingConfig:
     DEFAULT_OUTPUT_RASTER_LAYER_EXT = 'DefaultOutputRasterLayerExt'
     DEFAULT_OUTPUT_VECTOR_LAYER_EXT = 'DefaultOutputVectorLayerExt'
     TEMP_PATH = 'TEMP_PATH2'
+    RESULTS_GROUP_NAME = 'RESULTS_GROUP_NAME'
 
     settings = {}
     settingIcons = {}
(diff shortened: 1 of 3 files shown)


alexbruy:
Probably better to have only one setting "Group name" instead of two settings? If this setting is not defined layers will be loaded as now and if some value is entered then group with such name will be used.

gacarrillor:
@alexbruy Yes, I also considered that option. But, then I gave it a second thought attempting to make it more explicit for users, therefore the checkbox.

I'm fine with any of the two approaches here. I propose that we wait for another core dev's opinion, so that we can make a decision.

m-kuhn:
Nice functionality.
I would also go for a single option as proposed by @alexbruy with a placeholder text that makes this behavior explicit.
Otherwise this should go with a bigger redesign of the processing settings where we can have visually grouped conditional options like checkable group boxes.

gacarrillor:
Good, I'm fine with that approach as well. 

Just one question, what should be the default value, an empty string or something like "Processing results"?

nirvn:
@gacarrillor , nice contribution. I'm -1 on having it by default. 

On a side note: it'd be nice to have groups created for algorithms that output more than one layer. But I'm straying away from the PR here, sorry :wink:

gacarrillor:
Thanks for your comments, I just implemented the suggested approach:

:heavy_check_mark: Single setting.
:heavy_check_mark: Default value is empty, which won't load results to any predetermined group.
:heavy_check_mark: Show a placeholder explaining the behavior.

> On a side note: it'd be nice to have groups created for algorithms that output more than one layer. But I'm straying away from the PR here, sorry :wink: 

Yes, that would be good. In fact, one could also think in a new group per Processing execution. Something like "Processing results 2020.07.20 15:00". Just to brainstorm :smiley:

m-kuhn (approved):

PR #37599, opened 2020-07-05
Minor api cleanups (use Maximum/Minimum instead of min/max), dox

nyalldawson:


m-kuhn (approved):

troopa81:
Looks like testqgslegendrenderer is still [using the old API](https://github.com/qgis/QGIS/blob/master/tests/src/core/testqgslegendrenderer.cpp#L512). It doesn't compile anymore, but I failed to understand why everything is fine with the CI.

elpaso:
@troopa81 I don't have an explanation, but see this: https://github.com/qgis/QGIS/pull/37625

troopa81:
@elpaso Thanks

PR #37602, opened 2020-07-05
Add/use simple BaseClass to Python Console classes

rduivenvoorde:
This is a Work In Progress to do some cleanup in the python code for the Python Console (output, (line)-input and editor.
It appears there is some duplicate code in those files, so I start here to add a BaseClass and hopefully are able to move some of that duplicate code to this class.

Starting with those lists of constant colors :-)


nyalldawson:
This is a good move!

Something to keep in mind here -- we have a bit of code duplication with three separate Python script editors. One is the console version which you're looking at here. The other is QgsCodeEditorPython, which is a c++ widget and is used everywhere else we expose Python editing capabilities (e.g. project macros, custom expression functions). The last is the Processing script editor. 

It would be nice to start to condense all these into a single class. I think the way forward would be to port functionality bit by bit from the console editor into QgsCodeEditorPython, with the goal of one day making QgsCodeEditorPython the base class for the console editor.

rduivenvoorde:
I'm proposing to use just one code color scheme for python (instead of the two now used in the console settings.

Mathieu said: "Regarding the editor vs console color schemes, I also had the feeling it was overkill. So +1 to remove one. Be aware that the processing [python] script editor reads one set of colors, make sure you keep that one if you go ahead and remove one of the two sets."

Salvatore: "Regarding the editor vs console color schemes, I also had the feeling it was overkill. So +1 to remove one. Be aware that the processing [python] script editor reads one set of colors, make sure you keep that one if you go ahead and remove one of the two sets."

rduivenvoorde:
I could get some hints to further work in this. I think I'm more or less where I wanted to be at first...

Only TODO's:
- either create a 'color' tab in the console settings dialog (and move that block from the console settings tab)
OR create a python-code-color tab in the main settings (as current colors are shared between python console elements and processing python editors
- move the font/size from the base class for the editor (as that one has a separate font/size?) OR also iron that out, so all panes have the same size...

Any hints/idea's appreciated

@nyalldawson I do not think I have enought sip/python/cpp knowledge to merge this into QgsCodeEditorPython
But happy to try if some more info is passed :-)

python/console/console_base.py
```diff
@@ -0,0 +1,154 @@
...
+"""
+
+from qgis.PyQt.QtCore import Qt, QUrl
+from qgis.PyQt.QtGui import QColor, QFontDatabase, QDesktopServices
+from qgis.PyQt.Qsci import QsciScintilla, QsciLexerPython, QsciAPIs
+from qgis.core import QgsApplication, Qgis
+import os
+
+
+class QgsQsciScintillaBase(QsciScintilla):
```
nirvn, line 29: Could we come up with a better class name here? QgsQsci isn't super great. But I have no alternative to offer so if no better suggestion emerges, that can stick :)

python/console/console_base.py
```diff
@@ -0,0 +1,154 @@
...
+    MARGIN_BACKGROUND_COLOR = "#efefef"
+    MARGIN_FOREGROUND_COLOR = "#636363"
+    SELECTION_BACKGROUND_COLOR = "#d7d7d7"
+    SELECTION_FOREGROUND_COLOR = "#303030"
+    MATCHED_BRACE_BACKGROUND_COLOR = "#b7f907"
+    MATCHED_BRACE_FOREGROUND_COLOR = "#303030"
+    EDGE_COLOR = "#efefef"
+    FOLD_COLOR = "#efefef"
+
+    def __init__(self, parent=None):
```
nirvn, line 56: There's other stuff you can move in there. One crucial bit:

```
        # Set the default font
        font = QFontDatabase.systemFont(QFontDatabase.FixedFont)
        self.setFont(font)
        self.setMarginsFont(font)
```

That's needed everywhere we use QsciScintillaBase otherwise we end up with font glitches.

nirvn (commented):
Nice cleanup work. Left a few comments to consider.

python/console/console_base.py
rduivenvoorde: Happy to rename, just wanted to follow the normal naming conventions...

python/console/console_base.py
rduivenvoorde: Done, see all commits of today

rduivenvoorde:
Sorry for taking so long...

If all tests are happy, it would be nice if someone could have a look again.

I think in general the editors also look more consistent (because of all the coloring/margin/etcetc in the base Class)

I think there has been a nice cleanup, and having the base Class makes it easier to add functionality to one of the three scintilla panes...
It should also make it easier to move stuff to cpp maybe

I just hope I did not break something....

rduivenvoorde:
Hopefully to draw some attention, I will add some screenies :-)

The upper part (this PR) as you can see is a little more consitent in margins etc (because all the three editors: output, input and editor) share the same base.
The 'Search Selected in PyQGIS docs' is added

Screenshot-20200902111832-1445x801

The screenie left is this PR: the Console and Editor settings are merged (so ALL editors use the same python colors and settings), the OK/Cancel bar has some air, etc.

Screenshot-20200902111950-2123x758

slarosa:
Hi @rduivenvoorde, +1 for merging in the current form as this PR is a good code cleanup. Next we can try to inherit the base class on ```QgsCodeEditorPython``` one. 
I would just like to propose to rename ```QgsQsciScintillaBase``` to (maybe) ```QgsPythonConsole``` or just ```QgsConsole``` ??

rduivenvoorde:
> Hi @rduivenvoorde, +1 for merging in the current form as this PR is a good code cleanup. Next we can try to inherit the base class on `QgsCodeEditorPython` one.
> I would just like to propose to rename `QgsQsciScintillaBase` to (maybe) `QgsPythonConsole` or just `QgsConsole` ??

Can I leave the 'Base' into it? So QgsPythonConsoleBase or maybe even better: QgsPythonEditorBase?

slarosa:
@rduivenvoorde +1 to keep the Base word. ```QgsPythonConsoleBase``` sounds good to me!
Anyway, I done a quick test on MacOS and it looks good!

nyalldawson:
@nirvn can you take a re-look at this now?

python/console/console_base.py
nirvn: Fine.

python/console/console_base.py
nirvn: Thanks

python/console/console_base.py
```diff
@@ -0,0 +1,174 @@
...
+
+        # output and console
+        loadFont = self.settings.value("pythonConsole/fontfamilytext")
+        if loadFont:
+            font.setFamily(loadFont)
+        fontSize = self.settings.value("pythonConsole/fontsize", type=int)
+        if fontSize:
+            font.setPointSize(fontSize)
+
+        # TODO: editor has it's own font and size...
```
nirvn, line 114: What are we doing here, getting rid of it as we're trying to remove the dual settings situation?

nirvn (commented):
Nice to see this being pushed through the finish line. 

I think that might be ready to merge and test out. Just one comment on a lingering TODO.

python/console/console_base.py
rduivenvoorde: Thanks @nirvn Mathieu

I will remove the TODO it is not valid anymore because the editor does NOT have different settings anymore...

And I will rename QgsQsciScintillaBase to QgsPythonConsoleBase as Salvatore proposed

rduivenvoorde:
@nirvn anything else we want to add? Or can it be merged?

And IF it can be merged, should I do it myself (I do have the rights, I think), or is the general convention that somebody else does it after an approval of a core committer?

m-kuhn:
Thanks @rduivenvoorde !

nirvn:
:bow:

PR #37605, opened 2020-07-05
Virtual dataset groups for mesh layer

vcloarec:
This PR introduces "virtual" data set group for mesh layer that replaces the "memory" dataset group of the previous PR #37389 .

With the mesh calculator the user can choose to create those "virtual" dataset groups that will be added to the layer. Then, for these dataset groups, values are not stored in memory but each dataset is calculated when needed whit the formula entered in the mesh calculator.

Those virtual dataset groups are saved with the project.

If needed, the user can remove them or can persist them on files to make them persistent.

Persistent and virtual dataset groups are distinguished in source tab of the mesh layer properties by the background of the item, yellow for persistent and purple for virtual (as virtual field for vector layer).

virtual dataset


nirvn:
@vcloarec , the UI improvements in here are a good step in the right direction, thanks for that :+1:

nirvn:
@vcloarec , it would be great if this UI would allow for editing of the formula (vs having to delete and re-create a virtual dataset merely to tweak the formula).

src/analysis/mesh/qgsmeshvirtualdatasetgroup.h
```diff
@@ -26,9 +26,19 @@
...
 #define SIP_NO_FILE
 
+/**
+ * \ingroup analysis
+ * \class QgsMeshVirtualDatasetGroup
+ * Represents a dataset group calculated from a formula string.
+ * The calculation is done by the QgsMeshCalcUtils class from a QgsMeshCalcNode created from the formula
+ * Each dataset is calculted when needed, so there is only one dataset store in this class all the time.
+ *
+ * \since QGIS 3.6
```
Gustry, line 36:
```suggestion
 * \since QGIS 3.16
```

src/analysis/mesh/qgsmeshvirtualdatasetgroup.h
```diff
@@ -68,7 +78,12 @@ class ANALYSIS_EXPORT QgsMeshVirtualDatasetGroup: public QgsMeshDatasetGroup
     bool calculateDataset() const;
 };
 
-
+/**
+ * \ingroup analysis
+ * \class QgsMeshVirtualDatasetGroupGenerator
+ * Represents a factory that can be used to create virtual dataset group
+ * \since QGIS 3.6
```
Gustry, line 85:
```suggestion
 * \since QGIS 3.16
```

vcloarec:
> it would be great if this UI would allow for editing of the formula (vs having to delete and re-create a virtual dataset merely to tweak the formula).
@nirvn , it was my intention to add this, but still there is no control for circular reference in the mesh calculation algorithm, I prefer not. May be for next round.

src/analysis/mesh/qgsmeshcalcutils.cpp
```diff
@@ -153,6 +77,99 @@ std::shared_ptr<QgsMeshMemoryDataset> QgsMeshCalcUtils::create( const QgsMeshMem
   return create( grp.dataType() );
 }
 
+std::shared_ptr<QgsMeshMemoryDataset> QgsMeshCalcUtils::create( const QgsMeshDatasetIndex &datasetIndex ) const
```
wonder-sk, line 80: Can we rename this to e.g. createMemoryDataset() to make it more clear?

src/app/mesh/qgsmeshdatasetgrouptreeview.cpp
```diff
@@ -588,13 +593,36 @@ void QgsMeshDatasetGroupTreeView::contextMenuEvent( QContextMenuEvent *event )
...
+      for ( int dependentGroupIndex : dependencies )
+      {
+        QgsMeshDatasetGroupTreeItem *item = mModel->datasetGroupTreeItem( dependentGroupIndex );
+        if ( item )
+        {
+          varList.append( item->name() );
+          varList.append( QStringLiteral( "\n" ) );
+        }
+      }
+      QMessageBox::information( this, tr( "Remove Dataset Group" ), tr( "This dataset group can be removed because it has the following dependencies :\n%1" )
```
wonder-sk, line 613: can -> can't  ?

Also, is it correct that `groupIndexDependencies()` returns other groups that depend on this group? (or is it the other way around: that this group depends on some other groups?). In case of the former, I would maybe reformulate "because it has the following dependencies:" to something like "because other dataset groups depend on it:"

src/app/mesh/qgsmeshdatasetgrouptreeview.cpp
```diff
@@ -719,19 +747,117 @@ void QgsMeshDatasetGroupSaveMenu::saveDatasetGroup( int datasetGroup, const QStr
...
+}
+
+QVariant QgsMeshAvailableDatasetGroupTreeModel::headerData( int section, Qt::Orientation orientation, int role ) const
+{
+  Q_UNUSED( section )
+
+  if ( orientation == Qt::Horizontal && role == Qt::DisplayRole )
+  {
+    if ( section == 0 )
+      return tr( "Groups" );
```
wonder-sk, line 818: Maybe better to use singular `Group` ?

src/core/mesh/qgsmeshdatagenerator.h
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * \ingroup core
+ *
+ * Abstract class that is an interface for mesh data generator.
+ *
+ * Mesh generator can be used to create mesh data as virtual dataset group.
+ *
+ * \since QGIS 3.16
+ */
+class CORE_EXPORT QgsMeshDataGeneratorInterface
```
wonder-sk, line 40: If I understand correctly, we need this QgsMeshDataGeneratorInterface and QgsMeshDataGeneratorRegistry only because the mesh calculator implementation is inside qgis_analysis library?

I am wondering if it is worth adding the extra complexity with these new classes. And for example, any 3rd party apps would have issues using virtual dataset groups generator as it is only registered within QgisApp... What do you think if we would rather move the mesh calculator from qgis_analysis to qgis_core? Would it be a big problem? I guess the only reason the mesh calculator is inside qgis_analysis is because the raster calculator was already there - but even with raster calculator I think sooner or later we will want to have "virtual" raster bands calculated on the fly...

src/core/mesh/qgsmeshdataset.cpp
```diff
@@ -608,31 +636,149 @@ QString QgsMeshDatasetGroupTreeItem::defaultName() const
...
+  for ( int index : mDatasetGroupDependentOn )
+  {
+    QDomElement dependOnElement = doc.createElement( QStringLiteral( "dependent-on-item" ) );
+    dependOnElement.setAttribute( QStringLiteral( "dataset-index" ), index );
+    itemElement.appendChild( dependOnElement );
+  }
+
+  for ( int index : mDatasetGroupDependencies )
+  {
+    QDomElement dependencieElement = doc.createElement( QStringLiteral( "dependencie-item" ) );
```
wonder-sk, line 695: typo? "dependencie-item" -> "dependencies-item"

src/core/mesh/qgsmeshdataset.h
```diff
@@ -530,6 +529,234 @@ class CORE_EXPORT QgsMeshDatasetMetadata
...
+
+    /**
+     * Type of the dataset group
+     *
+     * \since QGIS 3.16
+     */
+    enum Type
+    {
+      None, //! Generic type used for non typed dataset group
+      Persistent, //! Dataset group store on file
```
wonder-sk, line 587:
```suggestion
      Persistent, //! Dataset group stored in a file
```

src/core/mesh/qgsmeshdataset.h
```diff
@@ -530,6 +529,234 @@ class CORE_EXPORT QgsMeshDatasetMetadata
...
+  public:
+
+    /**
+     * Type of the dataset group
+     *
+     * \since QGIS 3.16
+     */
+    enum Type
+    {
+      None, //! Generic type used for non typed dataset group
```
wonder-sk, line 586: Do we actually need this type at all?

src/core/mesh/qgsmeshdatasetgroupstore.cpp
```diff
@@ -257,13 +279,37 @@ void QgsMeshDatasetGroupStore::readXml( const QDomElement &storeElem, const QgsR
...
       source = mPersistentProvider;
+      sourceIndex = datasetElem.attribute( QStringLiteral( "source-index" ) ).toInt();
+    }
+    else
+    {
+      QString keyGenerator = datasetElem.attribute( QStringLiteral( "source-type" ) );
+      source = mExtraDatasets.get();
+      QString name = datasetElem.attribute( QStringLiteral( "name" ) );
+      QString formula = datasetElem.attribute( QStringLiteral( "formula" ) );
+      qint64 startTime = datasetElem.attribute( QStringLiteral( "startTime" ) ).toLongLong();
```
wonder-sk, line 299: use `start-time` and `end-time` attribute names to keep the same convention?

wonder-sk (commented):
Great work!

src/analysis/mesh/qgsmeshcalculator.cpp
```diff
@@ -267,12 +293,12 @@ QgsMeshCalculator::Result QgsMeshCalculator::processCalculation( QgsFeedback *fe
...
   // calculate statistics
-  outputGroup->calculateStatistic();
+  outputGroup->initialize();
+  outputGroup->setReferenceTime( static_cast<QgsMeshLayerTemporalProperties *>( mMeshLayer->temporalProperties() )->referenceTime() );
 
   const QgsMeshDatasetGroupMetadata meta = outputGroup->groupMetadata();
-  bool err;
 
-  if ( mResultInMemory )
+  if ( mDestination == QgsMeshDatasetGroup::Memory )
```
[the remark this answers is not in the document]
vcloarec: The result will be the same, but if one day there is other enum...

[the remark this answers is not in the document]
vcloarec: Not exactly the same logic. Here we want this if and only if `mDestination == QgsMeshDatasetGroup::Memory`

src/analysis/mesh/qgsmeshcalculator.cpp
```diff
@@ -170,7 +175,7 @@ QgsMeshCalculator::Result QgsMeshCalculator::expressionIsValid(
 QgsMeshCalculator::Result QgsMeshCalculator::processCalculation( QgsFeedback *feedback )
 {
   // check input
-  if ( mOutputFile.isEmpty() && !mResultInMemory )
+  if ( mOutputFile.isEmpty() && !( mDestination != QgsMeshDatasetGroup::Persistent ) )
```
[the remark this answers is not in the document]
vcloarec: sure...

src/app/mesh/qgsmeshdatasetgrouptreeview.cpp
vcloarec: first case.

src/core/mesh/qgsmeshdataset.cpp
vcloarec: there is only one item per element, so dependec**y**

src/core/mesh/qgsmeshdataset.h
vcloarec: yes, it is the default one used for QgsMeshDatasetGroupTreeItem, used for example for the root item.

PeterPetrik:
hi @vcloarec, would it be possible to move mesh calculator to qgis core as well as address other Martin's comments?

PeterPetrik (approved):

PR #37619, opened 2020-07-06
Manage Postgres parallel plans when estimating row count

### LINKED ISSUE #37342: [postgres] Wrong feature counts when using estimated metadata
Feature count returned by postgres provider may get wrong. In a case I was debugging, that was a PostgreSQL table with nearly 2M polygons. The estimated number of rows (shown in DB manager, using PostgreSQL query planner statistics) is +/- the same as the true number of rows.

When the layer is loaded with "use estimated metadata" = false, things work fine and featureCount() reports the correct count. But when **"use estimated metadata" = true**, then featureCount() only reports 2 (two!) features.

As a result, that completely breaks the default hybrid indexing strategy for snapping and QGIS tries to load all 2M geometries - which takes a lot of time and memory...

Tested with QGIS 3.10.7 and PostgreSQL 10.3.

Everything works fine in QGIS 3.4. It looks like this got broken in https://github.com/qgis/QGIS/pull/32114 (3.12+) and https://github.com/qgis/QGIS/pull/33019 (backport to 3.10.x)... so it probably affects also 3.14 and master (but I have not tested).

Further testing shows that this is due to the fact that the PostgreSQL uses **parallel query execution** and the EXPLAIN command says at the top-level that it will gather results from two workers:

Screenshot from 2020-06-22 19-34-32

So... it looks like it is quite fragile to depend on the EXPLAIN command output. My suggestion would be to stick to using `reltuples` from `pg_catalog.pg_class` for tables - and probably just for views we would use the EXPLAIN command which may get broken in the future again.

cc @troopa81

troopa81:
## Description

Fixes #37342, follows #32114

Previous PR wasn't dealing with possible parallel plan when estimating count. This one use a simpler SQL request where the number of rows is contained in the top node whether it is parallel (sum up the children gather node count) or not

### DIFF
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -3559,14 +3559,13 @@ long QgsPostgresProvider::featureCount() const
     {
       // parse explain output to estimate feature count
       // we don't use pg_class reltuples because it returns 0 for view
-      sql = QStringLiteral( "EXPLAIN (FORMAT JSON) SELECT count(*) FROM %1%2" ).arg( mQuery, filterWhereClause() );
+      sql = QStringLiteral( "EXPLAIN (FORMAT JSON) SELECT 1 FROM %1%2" ).arg( mQuery, filterWhereClause() );
       QgsPostgresResult result( connectionRO()->PQexec( sql ) );
 
       const QString json = result.PQgetvalue( 0, 0 );
       const QVariantList explain = QgsJsonUtils::parseJson( json ).toList();
-      const QVariantMap countPlan = explain.count() ? explain[0].toMap().value( "Plan" ).toMap() : QVariantMap();
-      const QVariantList queryPlan = countPlan.value( "Plans" ).toList();
-      const QVariant nbRows = queryPlan.count() ? queryPlan[0].toMap().value( "Plan Rows" ) : QVariant();
+      const QVariantMap countPlan = !explain.isEmpty() ? explain[0].toMap().value( "Plan" ).toMap() : QVariantMap();
+      const QVariant nbRows = countPlan.value( "Plan Rows" );
 
       if ( nbRows.isValid() )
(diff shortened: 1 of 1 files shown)


saberraz:
Does this need to be backported?

saberraz:
@wonder-sk would it be possible to review this PR?

troopa81:
> Does this need to be backported?

for 3.14 yes, for 3.10 I believe that it was proposed to wait & see if everything is fine in 3.14 before backporting.

wonder-sk:
Hmm... what do you think of using this method (parsing execution plan from JSON) _only_ with views - and for tables use the old method of reading the estimated counts directly from pg_class? The reliance on execution plan and its JSON is a bit scary to me and I would expect it can break when there are changes to the planner or to the output format...

troopa81:
> Hmm... what do you think of using this method (parsing execution plan from JSON) only with views - and for tables use the old method of reading the estimated counts directly from pg_class? 

It would cost an extra request to check if *mQuery* is a view or a table (calling [relkind](https://github.com/qgis/QGIS/blob/e91abdc23f944318db5e3559d47b38e36a81ae9e/src/providers/postgres/qgspostgresprovider.cpp#L4893))

> The reliance on execution plan and its JSON is a bit scary to me and I would expect it can break when there are changes to the planner or to the output format...

I understand your concerns but the proposed version relies on the number of rows of plan top node, so the planner could change but you'll always get an estimated row count in top node. As for the format, it could indeed change but as it's kind of a API breaking change, I would say it shouldn't happen often (never?). And if it happens we would have to fix the view part anyway.

wonder-sk:
Okay - let's leave it like this and hope PostgreSQL won't change the syntax in the future... or make the texts translatable :-)

wonder-sk (approved):

troopa81:
@wonder-sk Would you mind merging it?

PR #37635, opened 2020-07-07
Apply a font scale workaround factor to QgsTextRenderer

nyalldawson:
This is a hack which has been in place in layouts/composer for a long time in order to provide more consistent and reliable text rendering at small font sizes or when "zoomed out". Without this hack the font rendering and metric based calculations for these small font sizes can be unstable, leading to font sizes "jumping" around as a layout is zoomed into or out of.


nyalldawson:
(fingers crossed this is the magical unicorn fix to the text renderer layout issues...)

nyalldawson:
Merged in #37644

PR #37637, opened 2020-07-07
[MetaSearch] fix Python error when adding resources (#35560)

### LINKED ISSUE #35560: metasearch python error when trying to add a resource that is available as both WMS and WFS
it is about the metasearch catalog client which is a core plugin of QGIS.

i added the following service: http://nationaalgeoregister.nl/geonetwork/srv/dut/csw 
use keyword "bag" to search for

image

and then i select add data --> wfs

image

and then i got an error

image

tomkralidis:
## Description

Fixes #35560 in 3.10 branch.


nyalldawson (approved):

gioman:
@tomkralidis tested the patch, seems fine to me! Thanks!

PR #37640, opened 2020-07-07
fix adding relation with multiple field pairs from dialog

3nids:


src/app/qgsrelationmanagerdialog.cpp
```diff
@@ -138,7 +138,9 @@ void QgsRelationManagerDialog::mBtnAddRelation_clicked()
       ++suffix;
     }
     relation.setId( relationId );
-    relation.addFieldPair( addDlg.references().at( 0 ).first, addDlg.references().at( 0 ).second );
+    const auto references = addDlg.references();
+    for ( int p = 0; p < references.count(); p++ )
```
m-kuhn, line 142: Could this be a range based for loop?

```suggestion
    for ( const auto &ref : references )
```

3nids: obviously, done.

m-kuhn (approved):

PR #37662, opened 2020-07-08
Use QgsTextRenderer for drawing map grid text in layouts

nyalldawson:
Allows for grid annotations which use buffers, shadows, background shapes, etc!

Peek 2020-07-08 11-35


nirvn:
Sweet -- nice to see layout grids catching up to decoration grids wink: fantastic stuff.

nirvn (approved):
Looking good!

PR #37671, opened 2020-07-08
Rescale raster algorithm for Processing

### LINKED ISSUE #26099: Add a Rescale rasters alg
Author Name: **Paolo Cavallini** (@pcav)
Original Redmine Issue: [18208](https://issues.qgis.org/issues/18208)

Redmine category:analysis_library

---
It would be useful to have a new alg rescaling raster values to percent of max.
This would be useful e.g. for probability models, so that we can easily draw lines encompassing a certain percentage of total probability.
See #17761

alexbruy:
## Description
Add Rescale raster algorithm to change raster value range preserving shape of the raster's histogram. Useful when rasters from different sources should be compared or processed together and their pixel values should be in the same range. By default algorithm preserves NODATA value of the input raster but there is an option to override it.

Here is histogram of the DEM (value range is 85-243)
histogram-1
And histogram of same DEM rescaled to the range 100-1000
histogram-2

Fixes #26099.

### DIFF
--- a/src/analysis/processing/qgsalgorithmrescaleraster.cpp
+++ b/src/analysis/processing/qgsalgorithmrescaleraster.cpp
@@ -144,12 +144,11 @@ QVariantMap QgsRescaleRasterAlgorithm::processAlgorithm( const QVariantMap &para
   std::unique_ptr< QgsRasterBlock > block;
   while ( iter.readNextRasterPart( mBand, iterCols, iterRows, block, iterLeft, iterTop ) )
   {
-    if ( feedback )
-      feedback->setProgress( 100 * ( ( iterTop / blockHeight * numBlocksX ) + iterLeft / blockWidth ) / numBlocks );
+    feedback->setProgress( 100 * ( ( iterTop / blockHeight * numBlocksX ) + iterLeft / blockWidth ) / numBlocks );
 
     for ( int row = 0; row < iterRows; row++ )
     {
-      if ( feedback && feedback->isCanceled() )
+      if ( feedback->isCanceled() )
         break;
 
       for ( int col = 0; col < iterCols; col++ )
--- a/src/analysis/processing/qgsalgorithmrescaleraster.cpp
+++ b/src/analysis/processing/qgsalgorithmrescaleraster.cpp
@@ -1,0 +1,168 @@
+/***************************************************************************
+                         qgsalgorithmrescaleraster.cpp
+                         ---------------------
+    begin                : July 2020
+    copyright            : (C) 2020 by Alexander Bruy
+    email                : alexander dot bruy at gmail dot com
+ ***************************************************************************/
+
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+#include <limits>
+#include "math.h"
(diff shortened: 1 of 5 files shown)


src/analysis/processing/qgsalgorithmrescaleraster.cpp
```diff
@@ -0,0 +1,178 @@
...
+}
+
+QString QgsRescaleRasterAlgorithm::displayName() const
+{
+  return QObject::tr( "Rescale raster" );
+}
+
+QStringList QgsRescaleRasterAlgorithm::tags() const
+{
+  return QObject::tr( "raster,rescale,min,max" ).split( ',' );
```
nyalldawson, line 37:
```suggestion
  return QObject::tr( "raster,rescale,minimum,maximum,range" ).split( ',' );
```

src/analysis/processing/qgsalgorithmrescaleraster.cpp
```diff
@@ -0,0 +1,178 @@
...
+}
+
+QString QgsRescaleRasterAlgorithm::groupId() const
+{
+  return QStringLiteral( "rasteranalysis" );
+}
+
+QString QgsRescaleRasterAlgorithm::shortHelpString() const
+{
+  return QObject::tr( "Rescales raster layer to the new values range preserving shape "
```
nyalldawson, line 52:
```suggestion
  return QObject::tr( "Rescales raster layer to a new value range, while preserving the shape "
```

src/analysis/processing/qgsalgorithmrescaleraster.cpp
```diff
@@ -0,0 +1,178 @@
...
+                      "(distribution) of the raster's histogram (pixel values).\n\n"
+                      "By default algorithm preserves original NODATA value, but there is "
```
nyalldawson, line 54:
```suggestion
                      "By default the algorithm preserves original the NODATA value, but there is "
```

nyalldawson (approved):

src/analysis/processing/qgsalgorithmrescaleraster.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+QString QgsRescaleRasterAlgorithm::groupId() const
+{
+  return QStringLiteral( "rasteranalysis" );
+}
+
+QString QgsRescaleRasterAlgorithm::shortHelpString() const
+{
+  return QObject::tr( "Rescales raster layer to the new values range preserving shape "
+                      "(distribution) of the raster's histogram (pixel values).\n\n"
```
nyalldawson, line 53:
```suggestion
                      "(distribution) of the raster's histogram (pixel values) Input values are mapped using a linear interpolation from the source raster's minimum and maximum pixel values to the destination minimum and maximum pixel range.\n\n"
```

src/analysis/processing/qgsalgorithmrescaleraster.cpp
alexbruy: Fixed

src/analysis/processing/qgsalgorithmrescaleraster.cpp
alexbruy: Fixed, thanks!

src/analysis/processing/qgsalgorithmrescaleraster.cpp
alexbruy: Applied, thanks!

src/analysis/processing/qgsalgorithmrescaleraster.cpp
alexbruy: Fixed.

Saijin-Naib:
This has some awesome Remote Sensing applications!

PR #37672, opened 2020-07-08
Return on shortcut paste when layer not editable

signedav:
On Copy / Paste pasting with the shortcut (Ctrl-V) to a layer that is not editable, it did nothing but give the message "x features pasted" as feedback. Now on shortcut paste (and on shortcut cut (Ctrl-X)) we give the feedback that the layer is not editable (like on shortcut delete).

This issue has been mentioned here https://github.com/qgis/QGIS/pull/37617#discussion_r450365151 but is not directly related to the the attribute-fix implementation.

Additionally I improved some feedback messages.

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -10270,32 +10270,38 @@ void QgisApp::pasteFromClipboard( QgsMapLayer *destinationLayer )
 
 void QgisApp::pasteFeatures( QgsVectorLayer *pasteVectorLayer, int invalidGeometriesCount, int nTotalFeatures, QgsFeatureList &features )
 {
-  pasteVectorLayer->addFeatures( features );
-  QgsFeatureIds newIds;
-  newIds.reserve( features.size() );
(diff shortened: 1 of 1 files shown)


m-kuhn:
Thanks a lot.
Looks good to me.
Can you prepare a manual backport without the translation changes?

DelazJ:
>  Can you prepare a manual backport **without the translation changes**?

Just to mention that all the three active releases of QGIS are translatable till they are EOLed. (in the case you were worrying to break translations for the version life)

m-kuhn:
> Just to mention that all the three active releases of QGIS are translatable till they are EOLed. (in the case you were worrying to break translations for the version life)

It's my worry that translations for backports are worse maintained than translations for master and it's a shorter time frame. If we change source strings today we'll need someone to provide updated translations in the next 14 days or we have a "translation regression" in the next patch release.

signedav:
...and the changes on the existing message texts are not that important. Fine for me to leave them there as they are...

m-kuhn (approved):

PR #37697, opened 2020-07-09
when adding new field apply precision only when field type supports it

alexbruy:
## Description
Fixes [regression](https://github.com/qgis/QGIS/pull/36524#issuecomment-655614909) in the add attribute dialog (field precision is used when adding field even if field type does not support precision) which prevents user from leaving editing session.


src/gui/qgsaddattrdialog.cpp
```diff
@@ -138,7 +138,7 @@ QgsField QgsAddAttrDialog::field() const
            ( QVariant::Type ) mTypeBox->currentData( Qt::UserRole ).toInt(),
            mTypeBox->currentData( Qt::UserRole + 1 ).toString(),
            mLength->value(),
-           mPrec->value(),
+           mPrec->isVisible() ? mPrec->value() : 0,
```
nyalldawson, line 141: This approach is a bit dangerous, because if the method is called after the dialog has been accepted then the widget is no longer visible in any case. Better to also disable it when you hide it, and use isEnabled here instead

alexbruy: Fixed.

nyalldawson (approved):

PR #37698, opened 2020-07-09
Disable oracle tests

m-kuhn:
Oracle tests randomly fail at the moment

See e.g. https://cdash.orfeo-toolbox.org/testDetails.php?test=21120313&build=47048


m-kuhn:
Checking the trends on cdash again, I don't see it fail that often. My memory must have failed me and this was an exception that it failed.

PR #37717, opened 2020-07-09
Fix label of button to load a 3D map view in layout

### LINKED ISSUE #37711: [layout][3D] Change the label of "copy settings from 3d view" button
1. Add two 3D map views to your project
2. In a print layout, add a 3d map item
3. To assign it a 3d map view, in the item properties, push the "copy settings from 3D view" button and pick a 3d view
You get the 3D map item render the 3d view you selected and in the "camera pose" section, values are updated
5. Now select the other 3d map view. You'll notice that while the image is updated, the camera pose section is still the same.
 Camera pose settings are copied only the first time.

I'm fine with these values being independent but I think that the label of the button is misleading (it does not always copy all the settings of the view I select). Also this looks confusing to me considering that there's another button dedicated to populate/update/assign those camera settings.

The "copy settings from 3d view" could instead label something like (tooltip oriented because of #37710) "select 3D map view to render", "Assign 3D map view"...

DelazJ:
fixes #37711

### DIFF
--- a/src/ui/layout/qgslayout3dmapwidgetbase.ui
+++ b/src/ui/layout/qgslayout3dmapwidgetbase.ui
@@ -69,7 +69,7 @@
           <item>
            <widget class="QPushButton" name="mCopySettingsButton">
             <property name="text">
-             <string>Copy Settings from a 3D View…</string>
+             <string>Assign 3D Map View…</string>
             </property>
            </widget>
           </item>


DelazJ:
Am I unlucky that nobody could review this PR or the only one feeling that the label of this button does not do what it promises?

nyalldawson:
I'm not 100% on the new wording, that's the only reason I haven't commented yet! To me "assign" implies an ongoing link between the layout item and the map, rather then the "copy" settings which clearly reflects that there's NO live link between the item and map.

nirvn:
I share the same opinion as @nyalldawson , I looked at this PR a while back and seems my comment then never got published.

DelazJ:
Sorry for the late reply; I had broken my 3D build and didn't have access to one until now.
> Am I unlucky that nobody could review this PR or the only one feeling that the label of this button does not do what it promises?

Seems the answer is "both" 😞 
> To me "assign" implies an ongoing link between the layout item and the map, rather then the "copy" settings which clearly reflects that there's NO live link between the item and map.

Indeed. I thought there was a link between the 3D map item and the 3d map view but when this button is used the first time, you only copy the camera position and what the 3d map view is currently showing. You can manually change the camera position or (re)set it using the "set from 3D view" button.
But what happens to the rendering? If I update/change the map theme assigned to the 3D Map view, this is not reflected in the 3d map item until I reselect the view to use. Exporting the layout before I "reassign" the 3D map view results in a file with an outdated 3D map.
What is the advantage of this behavior?

I think this button should link the rendering of the 3D map view and 3D map item: at the creation it adds rendering + camera position and whenever the 3D map view is updated, it also updates the 3D map item rendering, leaving camera position as is.

PR #37734, opened 2020-07-10
Avoid texture jump when panning maps

nyalldawson:
We don't need the whole expense of the "Avoid artifacts when rendering as map tiles"
option, because that disables to default clipping of geometry parts which fall
outside the visible canvas extent and accordingly results in a lot of wasted rendering
calculations. Instead, we can just always use the upper left corner of the original
feature geometry before the optimisations like clipping to visible regions are applied,
and use this as the texture origin.

This means we can safely make this handling always applied and not an opt-in setting


andreasneumann:
Thanks, Nyall!

PR #37745, opened 2020-07-10
fix formula in the brightnesscontrast filter for rasters with opacity (fix #37742)

### LINKED ISSUE #37742: Regression: non-default {brightness,saturation,contrast,gamma} values and a global opacity < 100% breaks raster rendering
@alexbruy , I'm 99.999% sure this is a regression from your cool gamma feature. Under QGIS master, reducing the global opacity of a singleband gray raster when any of the {brightness,saturation,contrast,gamma} values have been tweaked break the rendering.

E.g., this is a simple DEM raster with 50% global opacity:
image

As soon as I change the brightness (e.g. setting brightness to _1_), the rendering is totally broken:
image

alexbruy:
## Description

Fixes #37742.

### DIFF
--- a/src/core/raster/qgsbrightnesscontrastfilter.cpp
+++ b/src/core/raster/qgsbrightnesscontrastfilter.cpp
@@ -183,7 +183,7 @@ int QgsBrightnessContrastFilter::adjustColorComponent( int colorComponent, int a
     double adjustedColor = colorComponent / alphaFactor;
 
     // Make sure to return a premultiplied color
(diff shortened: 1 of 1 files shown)


nirvn:
@alexbruy , thanks, that was fast :+1: I haven't tried other filters (gamma, etc.), might be worth doing a quick check on those too.

alexbruy:
@nirvn only brightness/contrast/gamma was affected. There were missed multiplier in the formula when code was merged from two independent filters in to single pipe.

nirvn:
@alexbruy , OK.

nyalldawson (approved):

PR #37756, opened 2020-07-11
Add option to show light source origins

### LINKED ISSUE #37726: Option to show light sources
It's currently very difficult to setup light sources in a 3d scene, because there's no easy way of seeing exactly where the light source is located. An option to temporarily show a marker sphere (or directional cone) showing the origin and direction of light sources would be very helpful!

nyalldawson:
If enabled, shows a sphere at light source origins, allowing easier repositioning
and placement of light sources relative to the scene contents

Fixes #37726

### DIFF
--- a/python/3d/auto_generated/qgs3dmapsettings.sip.in
+++ b/python/3d/auto_generated/qgs3dmapsettings.sip.in
@@ -352,6 +352,21 @@ Returns whether to show camera's view center as a sphere (for debugging)
 
 .. versionadded:: 3.4
 %End
+
+    void setShowLightSourceOrigins( bool enabled );
+%Docstring
+Sets whether to show light source origins as a sphere (for debugging)
+
+.. versionadded:: 3.16
+%End
+
+    bool showLightSourceOrigins() const;
+%Docstring
+Returns whether to show light source origins as a sphere (for debugging)
+
+.. versionadded:: 3.16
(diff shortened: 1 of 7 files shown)


nyalldawson:
@wonder-sk mind looking this over?

wonder-sk (approved):
Nice - very useful.

I wonder if we can also visualize directional lights somehow :-)   Maybe a cylinder + cone to create a 3d arrow pointing somewhere to the middle of the scene? :-)

(after Belgacem has added support for directional lights, I am wondering if we should not default to a one or two directional lights instead of point lights as the directional lights better approximate light from sun, and the intensity calculation does not involve some awkward numerical terms which are a bit difficult to grasp)

nyalldawson:
> I wonder if we can also visualize directional lights somehow :-) Maybe a cylinder + cone to create a 3d arrow pointing somewhere to the middle of the scene? :-)

I think we could -- I was going to check what blender/etc does to represent these, and see if we can copy this!

> (after Belgacem has added support for directional lights, I am wondering if we should not default to a one or two directional lights instead of point lights as the directional lights better approximate light from sun, and the intensity calculation does not involve some awkward numerical terms which are a bit difficult to grasp)

I'm +1 to that -- getting nice lighting setup seems to be the biggest hassle I have with 3d scenes

PR #37764, opened 2020-07-11
[python console] Match key up / down behavior to that of bash and allow for a 'soft' history to be editable

nirvn:
This PR reverts the recent behavior change of a key up / down press event which meant to prevent accidental loss of content of a command being edited when e.g. pressing the up key on the second line of a reflowed text.

This new approach has the python console input line behave like bash, whereas a key up / down will always move to the next / previous history item wherever the cursor is located. _However_ until the user hits enter, any edits to historical items or the new line will be remembered.

This insures that an accidental move back in history doesn't lead to unwanted code loss.

### DIFF
--- a/python/console/console_sci.py
+++ b/python/console/console_sci.py
@@ -104,7 +104,8 @@ class ShellScintilla(QsciScintilla, code.InteractiveInterpreter):
             self.runsource(line)
 
         self.history = []
-        self.historyIndex = 0
+        self.softHistory = []
+        self.softHistoryIndex = 0
         # Read history command file
         self.readHistoryFile()
 
@@ -342,7 +343,15 @@ class ShellScintilla(QsciScintilla, code.InteractiveInterpreter):
     def displayPrompt(self, more=False):
         self.SendScintilla(QsciScintilla.SCI_MARGINSETTEXT, 0, str.encode("..." if more else ">>>"))
 
-    def updateHistory(self, command):
+    def syncSoftHistory(self):
+        self.softHistory = self.history[:]
+        self.softHistory.append('')
+        self.softHistoryIndex = len(self.softHistory) - 1
+
+    def updateSoftHistory(self):
+        self.softHistory[self.softHistoryIndex] = self.text()
+
+    def updateHistory(self, command, skipSoftHistory=False):
         if isinstance(command, list):
             for line in command:
                 self.history.append(line)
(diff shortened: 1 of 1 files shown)


nirvn:
This came out of a discussion with @m-kuhn who rightfully pointed out that it was slightly aggravating to push key up to go back in history and have the cursor be placed at the beginning of the line.

uclaros:
> This insures that an accidental move back in history doesn't lead to unwanted code loss.

This is really great! However, the fact that all soft history entries are editable and edits are preserved when using up/down but lost when using enter can be confusing. For example, when browsing history with up/down the user decides to start typing on a clean line. Instead of pressing down until he reaches the end of history he clears the line and start typing. With soft history he can use up/down again to take glimpses of previous commands, however the one he is overwriting is missing and in the middle of the history stack making it harder to find again.
There should be some visual indication that soft history has been edited so the user is aware. Or maybe when soft history has been edited, a new entry could be appended so both original history and edited is preserved.

The history editing concept feels a little strange to me since my bash (debian) defaulted to using *history-search-backward* / *history-search-forward* with up/down keys and that now feels natural to me (ie only browse history entries that starts with current text).

nyalldawson:
> With soft history he can use up/down again to take glimpses of previous commands, however the one he is overwriting is missing and in the middle of the history stack making it harder to find again.

I agree with this -- it's always annoyed me how bash does this too

nyalldawson:
After chatting with @nirvn I realise that I misinterpreted the new behaviour, and I'm +1 to follow it as is!

nirvn:
@uclaros , I'm pretty sure this PR does exactly what you said at the end of your second paragraph:

>  Or maybe when soft history has been edited, a new entry could be appended so both original history and edited is preserved.

The edits to the 'soft' history are not permanent. Once the user executes an edited command in the history (i.e. hitting <return>), that edited command is appended to become the latest item of the permanent history and the soft history is reset to include both the latest item (the edited command) as well as the historical one (reset to its original value).

Let's merge now and have it run on dev builds for a while, I'm pretty sure it'll stick :smile:

uclaros:
I actually meant to add an entry to the soft history. So that the soft history always has all of the hard history entries and the edited entries.

m-kuhn:
Finally got around to test this. Works like a charm, thank you !

Issue #37757, opened 2020-07-11
Broken layers auto-find function freezes QGIS


elpaso:
The (relatively) new feature "Auto-find" in the broken layer dialog freezes QGIS completely in some cases.

I've been able to reproduce this in a few cases, try opening the attached project and click on "Auto-Find" in the broken layers dialog.

[MS_itopen.zip](https://github.com/qgis/QGIS/files/4906354/MS_itopen.zip)

elpaso:
CC @roya0045

roya0045:
Its not a freeze per say, its just searching  layer, I have tried to give feedback by showing a progress bar but it seems that the process doesn't allow the bar to pop up sometimes. I'm wondering if it could be done in a background thread.

But yeah, it demands a lot of ressource if the possible paths are deep to explore.

Suggestions are welcomed.

elpaso:
> Its not a freeze per say, 

Well, I waited ~ 10 minutes with QGIS completely unresponsive ... I'd call it a freeze/hang.

Anyway, this is a serious bug, are you sure there is no infinite loop somewhere?

> its just searching layer,

Where exactly? In the whole filesystem?

> I have tried to give feedback by showing a progress bar but it seems that the process doesn't allow the bar to pop up sometimes. 

The process must be user-interruptible.

> I'm wondering if it could be done in a background thread.

I think it's the right choice.

roya0045:
@elpaso the only issue that I have with background thread is users trying to use a project when the sources haven't all been found.

Yes the layers are searched in an outward spiral pattern until the search directory reaches a point where its too close to the root.

roya0045:
I'll try what failed previously as I used the progress dialog and process all events. I'll see if I can use a background thread instead.

PR #37766, opened 2020-07-11
Background layer finding

roya0045:
## Description
 improvement related to #37757 

I'm not familiar with background thread and the task manager. I'm making this as a draft for the moment in case someone has suggestions as I test things.

### DIFF
--- a/src/app/qgshandlebadlayers.cpp
+++ b/src/app/qgshandlebadlayers.cpp
@@ -562,7 +562,8 @@ void QgsHandleBadLayers::autoFind()
 
   const QList<int> layersToFind = fileBasedRows( !mLayerList->selectedItems().isEmpty() );
 
-  QProgressDialog progressDialog( QObject::tr( "Searching files" ), QObject::tr( "Cancel" ), 1, layersToFind.size(), this, Qt::Dialog );
+  QProgressDialog progressDialog( QObject::tr( "Searching files" ), 0, 1, layersToFind.size(), this, Qt::Dialog );
+  QgsTaskManager *manager = QgsApplication::taskManager();
 
   for ( int i : std::as_const( layersToFind ) )
   {
@@ -614,7 +615,20 @@ void QgsHandleBadLayers::autoFind()
 
     if ( !dataSourceChanged )
     {
-      QStringList filesFound = QgsFileUtils::findFile( fileName, basepath, 5 );
+      QStringList filesFound;
+      QgsFileSearchTask *fileutil = new QgsFileSearchTask( fileName,  basepath, 4, 4, QgsProject::instance()->absolutePath() );
+      fileutil->setDescription( "Searching for " + fileName );
+      manager->addTask( fileutil );
+      while ( !( ( fileutil->status() == QgsTask::Complete ) || ( fileutil->status() == QgsTask::Terminated ) ) )
+      {
+        QCoreApplication::processEvents();
+        if ( progressDialog.wasCanceled() )
+          fileutil->cancel();
+      }
+      // fileutil->waitForFinished();
+      if ( !( fileutil->isActive() ) )
+        filesFound = fileutil->results();
+
       if ( filesFound.length() > 1 )
       {
         bool ok;
@@ -668,10 +682,13 @@ void QgsHandleBadLayers::autoFind()
         item->setForeground( QBrush( Qt::red ) );
       }
     }
-
+    if ( progressDialog.wasCanceled() )
+      break;
   }
 
   QgsProject::instance()->layerTreeRegistryBridge()->setEnabled( false );
 
 }
+
+
 
--- a/src/core/qgsfileutils.cpp
+++ b/src/core/qgsfileutils.cpp
@@ -25,6 +25,7 @@
 #include <QDir>
 #include <QSet>
 #include <QDirIterator>
+#include "qgsfeedback.h"
 
 #ifdef Q_OS_UNIX
 // For getrlimit()
@@ -182,7 +183,7 @@ QString QgsFileUtils::findClosestExistingPath( const QString &path )
   return res == QLatin1String( "." ) ? QString() : res;
 }
 
-QStringList QgsFileUtils::findFile( const QString &file, const QString &basePath, int maxClimbs, int searchCeilling, const QString &currentDir )
+QStringList QgsFileUtils::findFile( const QString file, const QString basePath, int maxClimbs, int searchCeilling, const QString currentDir, QgsFeedback *feedback )
 {
   int depth = 0;
   QString originalFolder;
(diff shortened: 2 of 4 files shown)


src/core/qgsfileutils.cpp
```diff
@@ -228,3 +244,41 @@ QStringList QgsFileUtils::findFile( const QString &file, const QString &basePath
...
+  : QgsTask( tr( "searching for %1" ).arg( file ), QgsTask::CanCancel )
+  , mFile( file )
+  , mBasePath( basePath )
+  , mMaxClimbs( maxClimbs )
+  , mSearchCeil( searchCeilling )
+  , mCurrentDir( currentDir )
+  , mDialog( dialog )
+  , mResults( container )
+{
+  connect( dialog, QProgessDialog::canceled, this, QgsFileSearchTask::cancel );
```
nyalldawson, line 259: You can drop the dialog handling here -- QgsTask /QgsTaskManager does its own progress reporting (and it's not safe to access any gui component like a dialog from a background thread, this crashes Qt)

If you want to show the progress in a specific dialog outside of the standard task manager widget, you'd connect the task's progressChanged signal to your dialog's setProgress slot (and similarly with the cancelation -- connect the dialog's canceled signal to the tasks' cancel slot), but setup this in the main thread only.

src/core/qgsfileutils.cpp
```diff
@@ -228,3 +244,41 @@ QStringList QgsFileUtils::findFile( const QString &file, const QString &basePath
...
+}
+
+void QgsFileSearchTask::cancel()
```
nyalldawson, line 262: You can drop all the custom cancellation code here -- the base class versions are sufficient

src/app/qgshandlebadlayers.cpp
```diff
@@ -524,6 +524,7 @@ QString QgsHandleBadLayers::checkBasepath( const QString &layerId, const QString
 
 void QgsHandleBadLayers::autoFind()
 {
+  QgsTaskManager *manager = new QgsTaskManager();
```
nyalldawson, line 527: Don't create a new task manager, instead use the global application one via QgsApplication::taskManager()

src/core/qgsfileutils.h
```diff
@@ -98,8 +100,60 @@ class CORE_EXPORT QgsFileUtils
      * \returns List of strings of the first matching path in unix format.
      * \since QGIS 3.12
      */
-    static QStringList findFile( const QString &file, const QString &basepath = QString(), int maxClimbs = 4, int searchCeiling = 4, const QString &currentDir = QString() );
+    static QStringList findFile( const QString &file, const QString &basepath = QString(), int maxClimbs = 4, int searchCeiling = 4, const QString &currentDir = QString(), QgsFileSearchTask *task = nullptr );
```
nyalldawson, line 103: instead of passing the task as an argument here, the correct approach is to pass an optional "QgsFeedback" object to functions like this. Then in your task you create a new QgsFeedback object for passing to this function, and proxy the QgsFeedback objects' cancelation/progress reporting to the task via connections.

It keeps utility functions like findFile nice and generic and available for in a wider variety of contexts (e.g. we could easily then add a processing algorithm for finding files... which... actually sounds like a great idea :laughing: )

nyalldawson:
I should say that generally this is heading in the right direction, and is a good move!!

roya0045:
@nyalldawson Thanks for the review, I have my doubts about creating a task manager dependency in core files as it makes a core file dependent on the app. And I'm not sure if having a worker thread would be the best option or if using the process events with the dialog sufficient as it doesn't mix code but creates some potential issues...

src/app/qgshandlebadlayers.cpp
roya0045: Forgot about that, Thanks,

src/core/qgsfileutils.cpp
roya0045: Done

src/core/qgsfileutils.cpp
roya0045: Since using process events was clunky, I want to be able to provide a way to stop the whole process from the gui. Not sure what is the best option and if the task manager will only stop one lookup and not the whole process.

src/core/qgsfileutils.h
roya0045: Done, I wasn't sure if the feedback was mostly for processing or had any other dependencies as I wanted something linked to the dialog if possible.

roya0045:
Unsure what is causing the crash. I removed the container in the hopes that it was a pointer story but this does not seem to be the cause. Would have to debug but its hard to do when I have to always fix a reseting VM and nonsensical build issues on windows.

roya0045:
With the while loop it does the first item, then crash, but the dialog pop up. I could connect the cancel dialog and the cancel of the task. Still unsure what is the issue. Meanwhile the new QT and c++ don't like some element in the code and getting back the 'old' stuff is annoying.

roya0045:
@nyalldawson still no clue as to what might cause the crash...

roya0045:
Waiting for vs2019 patch to be included before re-attempting to debug

roya0045:
Still looking for the cause of crash, free time is rare these days

roya0045:
image
Here is the valgrind of the crash. Unsure what is the cause...

roya0045:
@nyalldawson have you ever had this issue with the manager and tasks? I thought I replicated how other tasks as handled but from the stack trace I'm not sure what could be the issue or if its related to the searching functions and task handling...

roya0045:
Using non ref for most parameters still result in CTD without trace. on win.

roya0045:
After removing refs I get a crash on wait, I'm not sure if its an issue with the manager and how it handle threads/tasks...
image

roya0045:
I removed all refs, no more crashes, though I also removed some const, I don't think they hurt, could reimplement them.

roya0045:
@nyalldawson Code is fine and stable now, not a prototype anymore. I would guess this is a feature.

roya0045:
Found something wonky with the existing implementation while using it. I'll try to fix it also in this PR until this is merged.

Current issue seems to be caused by multiple sources found. Will look into it in the next few days.

roya0045:
Fix is up waiting review/merge on both pr.

roya0045:
Not stale, waiting for review.

roya0045:
Same as before

roya0045:
Yep.

roya0045:
still fine

roya0045:
same

roya0045:
Not stale

roya0045:
It's fine.

roya0045:
waitin'

roya0045:
same old

roya0045:
fine

roya0045:
staying alive

roya0045:
ok

roya0045:
ok

src/core/qgsfileutils.h
```diff
@@ -182,4 +186,49 @@ class CORE_EXPORT QgsFileUtils
 
 };
 
+
+/**
+ * \ingroup core
+ * \class QgsFileSearchTask
+ * \brief Background thread handler for findFile.
+ * \since QGIS 3.18
```
lbartoletti, line 194: Can you rebase your PR and update the "since" part? Thanks

roya0045: Done as requested, thanks!

roya0045:
still ok

roya0045:
still ok

roya0045:
unstale

PR #37769, opened 2020-07-11
Cache snapping results in feature dialog for subsequent recalculation

### LINKED ISSUE #37359: Default values using @snapping_result delete attributes

m-kuhn:
Snapping results are used for initial calculation of default values on an attribute form.

If a default value has to be recalculated (because it depends on another attribute) the snapping results are no longer available.

Fixes #37359

### DIFF
--- a/src/gui/qgsattributedialog.h
+++ b/src/gui/qgsattributedialog.h
@@ -95,6 +95,14 @@ class GUI_EXPORT QgsAttributeDialog : public QDialog
      */
     bool event( QEvent *e ) override;
 
+    /**
+     * Sets an additional expression context scope to be used
+     * for calculations in this form.
+     *
+     * \since QGIS 3.16
+     */
+    void setExtraContextScope( QgsExpressionContextScope *extraScope SIP_TRANSFER );
+
   public slots:
     void accept() override;
     void reject() override;
--- a/src/gui/qgsattributedialog.h
+++ b/src/gui/qgsattributedialog.h
@@ -96,7 +96,7 @@ class GUI_EXPORT QgsAttributeDialog : public QDialog
     bool event( QEvent *e ) override;
 
     /**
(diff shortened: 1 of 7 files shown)


m-kuhn:
@signedav can I get your eyes on this?

src/gui/qgsattributedialog.h
```diff
@@ -95,6 +95,14 @@ class GUI_EXPORT QgsAttributeDialog : public QDialog
      */
     bool event( QEvent *e ) override;
 
+    /**
+     * Sets an additional expression context scope to be used
+     * for calculations in this form.
```
signedav, line 100: Is it an "additional"/"extra" scope or is it just "the" scope used in the expression context for the calculations?
Anyway, maybe add the param like: `Sets \a extraScope as the additional expression...`

signedav (approved):
Code looks good :+1:

src/gui/qgsattributedialog.h
m-kuhn: Other scopes are already there (global, project, layer, form) and this one can be provided from the exterior (e.g. map tool capture)
I'll adjust

PR #37776, opened 2020-07-12
Abstract out material settings

nyalldawson:
Opens the possibilities for non-phong based materials for 3d symbols

### DIFF
--- a/python/3d/3d_auto.sip
+++ b/python/3d/3d_auto.sip
@@ -3,6 +3,7 @@
 %Include auto_generated/qgs3d.sip
 %Include auto_generated/qgs3dmapsettings.sip
 %Include auto_generated/qgs3dtypes.sip
+%Include auto_generated/qgsabstractmaterialsettings.sip
 %Include auto_generated/qgsabstractvectorlayer3drenderer.sip
 %Include auto_generated/qgscamerapose.sip
 %Include auto_generated/qgslayoutitem3dmap.sip
--- a/python/3d/auto_generated/qgsabstractmaterialsettings.sip.in
+++ b/python/3d/auto_generated/qgsabstractmaterialsettings.sip.in
@@ -1,0 +1,108 @@
+/************************************************************************
+ * This file has been generated automatically from                      *
+ *                                                                      *
+ * src/3d/qgsabstractmaterialsettings.h                                 *
+ *                                                                      *
+ * Do not edit manually ! Edit header and run scripts/sipify.pl again   *
+ ************************************************************************/
+
+
+
+
+
+class QgsMaterialContext
+{
+%Docstring
+Context settings for a material.
+
+.. warning::
+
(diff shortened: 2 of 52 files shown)


nyalldawson:
@wonder-sk another one for you :)

q
```diff
@@ -0,0 +1,521 @@
+[1mdiff --git a/src/core/layertree/qgslayertreelayer.cpp b/src/core/layertree/qgslayertreelayer.cpp[m
```
wonder-sk, line 1: Probably unintended for check in :-)

wonder-sk (commented):
I really like where this work is going!

We just need to be careful about the kind of data on which a particular material is applied. For example, we use phong material settings class for all 3D symbols now, but it is a bit of cheating there anyway... because we support different things with different symbols, and when starting to use abstract base class, we should be probably more explicit about these differences. To summarize what we have:
- polygons - use phong material, expecting triangles
- buffered lines - use phong material, expecting triangles
- simple lines - use custom line material, expecting line data
- instanced points - use custom phong material capable of instancing, expecting triangles + points
- billboard points - use custom material, expecting points

We probably need to define rendering techniques, and any materials would register to a particular technique? So for example, for Phong shading we would have two materials - one from Qt for polygons and buffered lines, another one would be our custom for instanced rendering...

q
nyalldawson: Oops, yep!

nyalldawson:
> We just need to be careful about the kind of data on which a particular material is applied. For example, we use phong material settings class for all 3D symbols now, but it is a bit of cheating there anyway... because we support different things with different symbols, and when starting to use abstract base class, we should be probably more explicit about these differences. To summarize what we have:

> polygons - use phong material, expecting triangles
buffered lines - use phong material, expecting triangles

FYI - something has broken this in recent builds -- buffered lines now just show corrupted. I tried to track down what's happening, but can't work it out.

> simple lines - use custom line material, expecting line data
instanced points - use custom phong material capable of instancing, expecting triangles + points
billboard points - use custom material, expecting points

> We probably need to define rendering techniques, and any materials would register to a particular technique? So for example, for Phong shading we would have two materials - one from Qt for polygons and buffered lines, another one would be our custom for instanced rendering...

Yeah, that's a good point. Given that there's like to be a larger set of materials vs rendering techniques, I'm thinking of adding a flag to the abstract material class for "supported techniques". Then the symbol class itself can have a new "required techniques" method, which we can use at run time to restrict the choice of available materials to those which are compatible for the symbol. What do you think?

nyalldawson:
> We probably need to define rendering techniques, and any materials would register to a particular technique? So for example, for Phong shading we would have two materials - one from Qt for polygons and buffered lines, another one would be our custom for instanced rendering...

@wonder-sk I'll address this part in the next PR (after #37952 is merged)

PR #37780, opened 2020-07-12
Swap display_expression parameters order

DelazJ:
Because the commented expression below instead works
image

### DIFF
--- a/resources/function_help/json/display_expression
+++ b/resources/function_help/json/display_expression
@@ -2,39 +2,36 @@
   "name": "display_expression",
   "type": "function",
   "groups": ["Record and Attributes"],
-  "description": "Returns the display expression for a given feature in a layer. If called with no parameters, it evaluates the current feature. The expression is evaluated by default.",
-  "arguments": [
+  "description": "Returns the display expression for a given feature in a layer. The expression is evaluated by default. Can be used with zero, one or more arguments, see below for details.",
+  "variants": [
     {
-      "arg": "layer",
-      "optional": true,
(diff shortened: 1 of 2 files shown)


Gustry:
According to the test, it depends on the number of arguments.
https://github.com/qgis/QGIS/blob/master/tests/src/core/testqgsexpression.cpp#L2146

If a single argument is provided, it's a feature.
If two arguments, first it's the layer and then the feature.

It's the same for `is_selected` https://github.com/qgis/QGIS/blob/master/resources/function_help/json/is_selected and `maptip` https://github.com/qgis/QGIS/blob/master/resources/function_help/json/maptip

DelazJ:
So only the example is wrong? I should keep the parameters order.
Having to switch parameters order depending on their number looks confusing, that said.

Gustry:
> Having to switch parameters order depending on their number looks confusing, that said.

I agree now. I took example on the `is_selected` expression.
We could add the logic to manage both order in each expression.

nyalldawson:
The is_selected help was recently updated to fix a similar situation. I suggest modelling the help for display_expression on the new is_selected help format.

DelazJ:
I did something, and it renders as below (feeback welcome as i'm not quite comfortable wit this function)
image

nyalldawson:
Looks good to me! @gustry?

Gustry:
Yes, LGTM, thanks @DelazJ !

DelazJ:
Great; maptip is now also aligned with them

PR #37785, opened 2020-07-13
Add invalid state to QgsTextFormat, and improve UX for setting manual table text format overrides

nyalldawson:
Remove the checkbox for overridding text format and use the new "not set" state from the font button instead. Makes for a much nicer API and UX when setting format overrides for table cells.

### DIFF
--- a/src/core/textrenderer/qgstextshadowsettings.cpp
+++ b/src/core/textrenderer/qgstextshadowsettings.cpp
@@ -43,6 +43,33 @@ QgsTextShadowSettings::~QgsTextShadowSettings() //NOLINT
 
 }
 
+bool QgsTextShadowSettings::operator==( const QgsTextShadowSettings &other ) const
+{
+  if ( d->enabled != other.enabled()
+       || d->shadowUnder != other.shadowPlacement()
+       || d->offsetAngle != other.offsetAngle()
+       || d->offsetDist != other.offsetDistance()
+       || d->offsetUnits != other.offsetUnit()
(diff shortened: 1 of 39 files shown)


src/core/textrenderer/qgstextshadowsettings.cpp
```diff
@@ -43,6 +43,33 @@ QgsTextShadowSettings::~QgsTextShadowSettings() //NOLINT
...
+bool QgsTextShadowSettings::operator==( const QgsTextShadowSettings &other ) const
+{
+  if ( d->enabled != other.enabled()
+       || d->shadowUnder != other.shadowPlacement()
+       || d->offsetAngle != other.offsetAngle()
+       || d->offsetDist != other.offsetDistance()
+       || d->offsetUnits != other.offsetUnit()
+       || d->offsetMapUnitScale != other.offsetMapUnitScale()
+       || d->offsetGlobal != other.offsetGlobal()
+       || d->radius != other.blurRadius()
```
nirvn, line 55: radius == blurRadius? I'm just asking to make sure, we did change quite a few variable / function names during the 3.X cycle.

nirvn (approved):
Nice; I'm pretty sure these new {==,!=} operators might come in handy in other scenarios too.

src/core/textrenderer/qgstextshadowsettings.cpp
nyalldawson: Yep -- it's just the internal member name which wasn't updated

PR #37787, opened 2020-07-13
Allow cells in manual text tables to have expression based contents

nyalldawson:
Allows individual cells from a manual text table to take their contents
from a preset expression. Expressions have access to the full layout
item expression context, allowing cells to calculate and display
metadata style values or aggregate based calculations.

Sponsored by City of Canning

### DIFF
--- a/src/core/layout/qgslayouttable.cpp
+++ b/src/core/layout/qgslayouttable.cpp
@@ -989,6 +989,14 @@ QMap<int, QString> QgsLayoutTable::headerLabels() const
   return headers;
 }
 
+QgsExpressionContextScope *QgsLayoutTable::scopeForCell( int row, int column ) const
+{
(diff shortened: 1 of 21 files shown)


src/core/layout/qgslayouttable.cpp
```diff
@@ -989,6 +989,14 @@ QMap<int, QString> QgsLayoutTable::headerLabels() const
   return headers;
 }
 
+QgsExpressionContextScope *QgsLayoutTable::scopeForCell( int row, int column ) const
+{
+  std::unique_ptr< QgsExpressionContextScope > cellScope = qgis::make_unique< QgsExpressionContextScope >();
+  cellScope->setVariable( QStringLiteral( "row_number" ), row + 1, true );
```
nirvn, line 995: I really love you for that! :)

nirvn (approved):
:tada:

PR #37788, opened 2020-07-13
Add a Qgs3D singleton object, designed to collect other singleton-ish objects in a central place

nyalldawson:
Copying the pattern setup by the QgsGui singleton container

### DIFF
--- a/src/3d/qgs3d.cpp
+++ b/src/3d/qgs3d.cpp
@@ -1,0 +1,32 @@
+/***************************************************************************
+                         qgs3d.cpp
+                         ----------
+    begin                : July 2020
+    copyright            : (C) 2020 by Nyall Dawson
+    email                : nyall dot dawson at gmail dot com
+ ***************************************************************************/
+
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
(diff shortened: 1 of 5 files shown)


src/3d/qgs3d.cpp
```diff
@@ -0,0 +1,32 @@
...
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+
+#include "qgs3d.h"
+
+Qgs3D *Qgs3D::instance()
+{
+  static Qgs3D *sInstance( new Qgs3D() );
+  return sInstance;
```
m-kuhn, line 23: Should we use `Q_GLOBAL_STATIC` so it's cleaned up on exit?

nyalldawson: Possibly? If we apply it to QgsGui too...

m-kuhn: Worth a try, with a potential revert if we have crashes on exit because of interdependencies.

nyalldawson: Done

nyalldawson:
@m-kuhn Windows says no: https://cdash.orfeo-toolbox.org/testDetails.php?test=21270597&build=47599

m-kuhn (approved):

PR #37789, opened 2020-07-13
[FEATURE][3d] Enable embedded and remote 3D models for 3D point symbols

nirvn:
## Description

This PR unlocks project-embedded and remote 3D models for 3D point symbols. 

Being able to embed 3D models in the project file can be super useful when sharing projects.

Obligatory screenshot (take note of the 'embedded file' in the 3D symbol configuration widget at the right):
image

### DIFF
--- a/src/3d/qgs3dmapscene.cpp
+++ b/src/3d/qgs3dmapscene.cpp
@@ -35,6 +35,7 @@
 #include <QOpenGLFunctions>
 #include <QTimer>
 
+#include "qgsapplication.h"
 #include "qgsaabb.h"
 #include "qgsabstract3dengine.h"
 #include "qgs3dmapscenepickhandler.h"
@@ -47,7 +48,9 @@
 #include "qgseventtracing.h"
 #include "qgsmeshlayer.h"
 #include "qgsmeshlayer3drenderer.h"
+#include "qgspoint3dsymbol.h"
 #include "qgsrulebased3drenderer.h"
+#include "qgssourcecache.h"
 #include "qgsterrainentity_p.h"
 #include "qgsterraingenerator.h"
 #include "qgstessellatedpolygongeometry.h"
@@ -109,6 +112,41 @@ Qgs3DMapScene::Qgs3DMapScene( const Qgs3DMapSettings &map, QgsAbstract3DEngine *
   connect( &map, &Qgs3DMapSettings::showLightSourceOriginsChanged, this, &Qgs3DMapScene::updateLights );
   connect( &map, &Qgs3DMapSettings::fieldOfViewChanged, this, &Qgs3DMapScene::updateCameraLens );
   connect( &map, &Qgs3DMapSettings::renderersChanged, this, &Qgs3DMapScene::onRenderersChanged );
+
+  connect( QgsApplication::instance()->sourceCache(), &QgsSourceCache::remoteSourceFetched, this, [ = ]( const QString & url )
+  {
+    const QList<QgsMapLayer *> modelVectorLayers = mModelVectorLayers;
+    for ( QgsMapLayer *layer : modelVectorLayers )
+    {
+      QgsAbstract3DRenderer *renderer = layer->renderer3D();
+      if ( renderer )
+      {
+        if ( renderer->type() == QLatin1String( "vector" ) )
+        {
+          const QgsPoint3DSymbol *pointSymbol = static_cast< const QgsPoint3DSymbol * >( static_cast< QgsVectorLayer3DRenderer *>( renderer )->symbol() );
+          if ( pointSymbol->shapeProperties()[QStringLiteral( "model" )].toString() == url )
+          {
+            removeLayerEntity( layer );
+            addLayerEntity( layer );
+          }
+        }
+        else if ( renderer->type() == QLatin1String( "rulebased" ) )
+        {
+          const QgsRuleBased3DRenderer::RuleList rules = static_cast< QgsRuleBased3DRenderer *>( renderer )->rootRule()->descendants();
+          for ( auto rule : rules )
+          {
+            const QgsPoint3DSymbol *pointSymbol = dynamic_cast< const QgsPoint3DSymbol * >( rule->symbol() );
+            if ( pointSymbol->shapeProperties()[QStringLiteral( "model" )].toString() == url )
+            {
+              removeLayerEntity( layer );
+              addLayerEntity( layer );
+              break;
+            }
+          }
+        }
+      }
+    }
+  } );
 
   // create entities of renderers
 
@@ -617,6 +655,31 @@ void Qgs3DMapScene::addLayerEntity( QgsMapLayer *layer )
          ( renderer->type() == QLatin1String( "vector" ) || renderer->type() == QLatin1String( "rulebased" ) ) )
     {
       static_cast<QgsAbstractVectorLayer3DRenderer *>( renderer )->setLayer( static_cast<QgsVectorLayer *>( layer ) );
+      if ( renderer->type() == QLatin1String( "vector" ) )
+      {
+        QgsVectorLayer *vlayer = qobject_cast<QgsVectorLayer *>( layer );
+        if ( vlayer->geometryType() == QgsWkbTypes::PointGeometry )
+        {
+          const QgsPoint3DSymbol *pointSymbol = static_cast< const QgsPoint3DSymbol * >( static_cast< QgsVectorLayer3DRenderer *>( renderer )->symbol() );
+          if ( pointSymbol->shape() == QgsPoint3DSymbol::Model )
+          {
+            mModelVectorLayers.append( layer );
+          }
+        }
+      }
+      else if ( renderer->type() == QLatin1String( "rulebased" ) )
+      {
+        const QgsRuleBased3DRenderer::RuleList rules = static_cast< QgsRuleBased3DRenderer *>( renderer )->rootRule()->descendants();
+        for ( auto rule : rules )
+        {
+          const QgsPoint3DSymbol *pointSymbol = dynamic_cast< const QgsPoint3DSymbol * >( rule->symbol() );
+          if ( pointSymbol && pointSymbol->shape() == QgsPoint3DSymbol::Model )
+          {
+            mModelVectorLayers.append( layer );
+            break;
+          }
+        }
+      }
     }
     else if ( layer->type() == QgsMapLayerType::MeshLayer && renderer->type() == QLatin1String( "mesh" ) )
     {
@@ -692,6 +755,7 @@ void Qgs3DMapScene::removeLayerEntity( QgsMapLayer *layer )
   {
     QgsVectorLayer *vlayer = qobject_cast<QgsVectorLayer *>( layer );
     disconnect( vlayer, &QgsVectorLayer::selectionChanged, this, &Qgs3DMapScene::onLayerRenderer3DChanged );
+    mModelVectorLayers.removeAll( layer );
   }
 
   if ( layer->type() == QgsMapLayerType::MeshLayer )
--- a/src/3d/symbols/qgspoint3dsymbol_p.cpp
+++ b/src/3d/symbols/qgspoint3dsymbol_p.cpp
@@ -44,12 +44,14 @@
 #include "qgspoint3dsymbol.h"
 #include "qgs3dmapsettings.h"
 
+#include "qgsapplication.h"
 #include "qgsvectorlayer.h"
 #include "qgspoint.h"
 #include "qgs3dutils.h"
 #include "qgsbillboardgeometry.h"
 #include "qgspoint3dbillboardmaterial.h"
 #include "qgslogger.h"
+#include "qgssourcecache.h"
 #include "qgssymbol.h"
 #include "qgssymbollayerutils.h"
 #include "qgssymbollayer.h"
@@ -439,19 +441,24 @@ void QgsModelPoint3DSymbolHandler::addSceneEntities( const Qgs3DMapSettings &map
   Q_UNUSED( map )
   for ( const QVector3D &position : positions )
   {
-    // build the entity
-    Qt3DCore::QEntity *entity = new Qt3DCore::QEntity;
+    const QString source = QgsApplication::instance()->sourceCache()->localFilePath( symbol.shapeProperties()[QStringLiteral( "model" )].toString() );
+    // if the source is remote, the Qgs3DMapScene will take care of refreshing this 3D symbol when the source is fetched
+    if ( !source.isEmpty() )
+    {
+      // build the entity
+      Qt3DCore::QEntity *entity = new Qt3DCore::QEntity;
 
-    QUrl url = QUrl::fromLocalFile( symbol.shapeProperties()[QStringLiteral( "model" )].toString() );
-    Qt3DRender::QSceneLoader *modelLoader = new Qt3DRender::QSceneLoader;
-    modelLoader->setSource( url );
+      QUrl url = QUrl::fromLocalFile( source );
+      Qt3DRender::QSceneLoader *modelLoader = new Qt3DRender::QSceneLoader;
+      modelLoader->setSource( url );
 
-    entity->addComponent( modelLoader );
-    entity->addComponent( transform( position, symbol ) );
-    entity->setParent( parent );
+      entity->addComponent( modelLoader );
+      entity->addComponent( transform( position, symbol ) );
+      entity->setParent( parent );
 
 // cppcheck wrongly believes entity will leak
 // cppcheck-suppress memleak
+    }
   }
 }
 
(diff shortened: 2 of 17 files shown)


src/3d/qgs3dmodelcache.cpp
```diff
@@ -0,0 +1,147 @@
...
+  connect( this, &QgsAbstractContentCacheBase::remoteContentFetched, this, &Qgs3DModelCache::remote3DModelFetched );
+}
+
+Qt3DRender::QMesh *Qgs3DModelCache::pathAsMesh( const QString &path, bool blocking )
+{
+  Qt3DRender::QMesh *mesh = new Qt3DRender::QMesh;
+
+  const QString file = path.trimmed();
+  if ( file.isEmpty() )
+    return mesh;
```
nyalldawson, line 70: should this be nullptr?

src/3d/qgs3dmodelcache.cpp
```diff
@@ -0,0 +1,147 @@
...
+  return mesh;
+}
+
+Qt3DRender::QSceneLoader *Qgs3DModelCache::pathAsSceneLoader( const QString &path, bool blocking )
+{
+  Qt3DRender::QSceneLoader *sceneLoader = new Qt3DRender::QSceneLoader;
+
+  const QString file = path.trimmed();
+  if ( file.isEmpty() )
+    return sceneLoader;
```
nyalldawson, line 94: again, should this be nullptr?

src/3d/qgs3dmodelcache.h
```diff
@@ -0,0 +1,119 @@
...
+     * Constructor for Qgs3DModelCache, with the specified \a parent object.
+     */
+    Qgs3DModelCache( QObject *parent SIP_TRANSFERTHIS = nullptr );
+
+    /**
+     * Returns a QMesh object with its source set to the specified \a path
+     *
+     * \a path may be a local file, remote (HTTP) url, or a base 64 encoded string (with a "base64:" prefix).
+     */
+    Qt3DRender::QMesh *pathAsMesh( const QString &path, bool blocking = false );
```
nyalldawson, line 93: I'd suggest adding SIP_FACTORY to this and pathAsSceneLoader -- even though it's currently not exposed to sip, it may save hassle in the long term if this class is ever exposed. Either that or make it return a std::unique_ptr so that it's clear that ownership is returned

nyalldawson (commented):
I think you missed the line edit widget files?

src/3d/qgs3dmodelcache.cpp
nirvn: Na, it returns an empty mesh object. It's IMHO better than nullptr, it gives you the certainty that you get an object. You can always is the source().isEmpty() afterwards.

nirvn:
I did a bit of testing with base64 URLs. In a test project of 14,000 points, using base64 URLs as source for models was _much_ slower and QGIS' memory usage saw an increase of 3gb. 

Instead of relying on costly base64 URLs, the 3d model cache now uses a QTemporaryDir to copy embedded and remote models for QGIS to use. No additional time or memory usage cost :tada:

nyalldawson:
LGTM -- let's give @wonder-sk a chance to review before merge

src/3d/qgs3dmodelcache.h
```diff
@@ -0,0 +1,112 @@
...
+#include <Qt3DRender/QMesh>
+
+
+///@cond PRIVATE
+
+/**
+ * \ingroup 3d
+ * \class Qgs3DModelCacheEntry
+ * An entry for a Qgs3DModelCache, representing a 3D model filename
+ * \since QGIS 3.6
```
nyalldawson, line 34:
```suggestion
 * \since QGIS 3.16
```

nirvn: No interest in backporting to RIP 3.6 branch? :wink:

nirvn:
@jef-n , any idea why windows won't compile this? Azure throws the following errors:

```
Build Log Line 	4265
Error 	
mocs_compilation.cpp.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) public: virtual __cdecl QgsAbstractContentCache<class Qgs3DModelCacheEntry>::~QgsAbstractContentCache<class Qgs3DModelCacheEntry>(void)" (__imp_??1?$QgsAbstractContentCache@VQgs3DModelCacheEntry@@@@UEAA@XZ) referenced in function "public: virtual __cdecl Qgs3DModelCache::~Qgs3DModelCache(void)" (??1Qgs3DModelCache@@UEAA@XZ)

Build Log Line 	4266
Error 	
qgs3dmodelcache.cpp.obj : error LNK2001: unresolved external symbol "__declspec(dllimport) public: virtual __cdecl QgsAbstractContentCache<class Qgs3DModelCacheEntry>::~QgsAbstractContentCache<class Qgs3DModelCacheEntry>(void)" (__imp_??1?$QgsAbstractContentCache@VQgs3DModelCacheEntry@@@@UEAA@XZ)

Build Log Line 	4267
Error 	
qgs3dmapscene.cpp.obj : error LNK2001: unresolved external symbol "__declspec(dllimport) public: virtual __cdecl QgsAbstractContentCache<class Qgs3DModelCacheEntry>::~QgsAbstractContentCache<class Qgs3DModelCacheEntry>(void)" (__imp_??1?$QgsAbstractContentCache@VQgs3DModelCacheEntry@@@@UEAA@XZ)

Build Log Line 	4268
Error 	
qgspoint3dsymbol_p.cpp.obj : error LNK2001: unresolved external symbol "__declspec(dllimport) public: virtual __cdecl QgsAbstractContentCache<class Qgs3DModelCacheEntry>::~QgsAbstractContentCache<class Qgs3DModelCacheEntry>(void)" (__imp_??1?$QgsAbstractContentCache@VQgs3DModelCacheEntry@@@@UEAA@XZ)

Build Log Line 	4269
Error 	
qgs3dmodelcache.cpp.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) public: __cdecl QgsAbstractContentCache<class Qgs3DModelCacheEntry>::QgsAbstractContentCache<class Qgs3DModelCacheEntry>(class QObject *,class QString const &,long,int)" (__imp_??0?$QgsAbstractContentCache@VQgs3DModelCacheEntry@@@@QEAA@PEAVQObject@@AEBVQString@@JH@Z) referenced in function "public: __cdecl Qgs3DModelCache::Qgs3DModelCache(class QObject *)" (??0Qgs3DModelCache@@QEAA@PEAVQObject@@@Z)

Build Log Line 	4270
Error 	
qgs3dmodelcache.cpp.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) protected: class QByteArray __cdecl QgsAbstractContentCache<class Qgs3DModelCacheEntry>::getContent(class QString const &,class QByteArray const &,class QByteArray const &,bool)const " (__imp_?getContent@?$QgsAbstractContentCache@VQgs3DModelCacheEntry@@@@IEBA?AVQByteArray@@AEBVQString@@AEBV2@1_N@Z) referenced in function "private: class QString __cdecl Qgs3DModelCache::fetch3DModel(class QString const &,bool &,bool)const " (?fetch3DModel@Qgs3DModelCache@@AEBA?AVQString@@AEBV2@AEA_N_N@Z)

Build Log Line 	4271
Error 	
qgs3dmodelcache.cpp.obj : error LNK2019: unresolved external symbol "__declspec(dllimport) protected: class Qgs3DModelCacheEntry * __cdecl QgsAbstractContentCache<class Qgs3DModelCacheEntry>::findExistingEntry(class Qgs3DModelCacheEntry *)" (__imp_?findExistingEntry@?$QgsAbstractContentCache@VQgs3DModelCacheEntry@@@@IEAAPEAVQgs3DModelCacheEntry@@PEAV2@@Z) referenced in function "public: class Qt3DRender::QMesh * __cdecl Qgs3DModelCache::pathAsMesh(class QString const &,bool)" (?pathAsMesh@Qgs3DModelCache@@QEAAPEAVQMesh@Qt3DRender@@AEBVQString@@_N@Z)

Build Log Line 	4272
Error 	
qgs3dmodelcache.cpp.obj : error LNK2001: unresolved external symbol "protected: virtual void __cdecl QgsAbstractContentCache<class Qgs3DModelCacheEntry>::onRemoteContentFetched(class QString const &,bool)" (?onRemoteContentFetched@?$QgsAbstractContentCache@VQgs3DModelCacheEntry@@@@MEAAXAEBVQString@@_N@Z)
```

src/3d/qgs3dmodelcache.h
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * \class Qgs3DModelCache
+ * \ingroup 3d
+ * A cache for 3D models.
+ *
+ * Qgs3DModelCache delivers pre-fetched 3D models.
+ *
+ * Qgs3DModelCache is not usually directly created, but rather accessed through
+ * QgsApplication::3dModelCache().
```
wonder-sk, line 64: should this read Qgs3D::qgs3DModelCache() instead?

src/3d/qgs3dmodelcache.h
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * Returns a QSceneLoader object with its source set to the specified \a path
+     *
+     * \a path may be a local file, remote (HTTP) url, or a base 64 encoded string (with a "base64:" prefix).
+     */
+    Qt3DRender::QSceneLoader *pathAsSceneLoader( const QString &path, bool blocking = false );
+
+    //! Returns the Qgs3DModelCache singleton instance
+    static Qgs3DModelCache *instance();
```
wonder-sk, line 94: how about this instance() - shall we use the one from Qgs3D or this one?

src/3d/qgs3dmodelcache.h
```diff
@@ -0,0 +1,112 @@
...
+     * \a path may be a local file, remote (HTTP) url, or a base 64 encoded string (with a "base64:" prefix).
+     */
+    Qt3DRender::QMesh *pathAsMesh( const QString &path, bool blocking = false );
+
+    /**
+     * Returns a QSceneLoader object with its source set to the specified \a path
+     *
+     * \a path may be a local file, remote (HTTP) url, or a base 64 encoded string (with a "base64:" prefix).
+     */
+    Qt3DRender::QSceneLoader *pathAsSceneLoader( const QString &path, bool blocking = false );
```
wonder-sk, line 91: Hmm... to me whether a model is loaded as QMesh or as a QSceneLoader should be an implementation detail of the 3d point symbol implementation... could we have API that would just return URI and leave the rest to qgspoint3dsymbol_p.cpp ?

src/3d/qgs3dmodelcache.h
```diff
@@ -0,0 +1,112 @@
...
+
+///@cond PRIVATE
+
+/**
+ * \ingroup 3d
+ * \class Qgs3DModelCacheEntry
+ * An entry for a Qgs3DModelCache, representing a 3D model filename
+ * \since QGIS 3.16
+ */
+class _3D_EXPORT Qgs3DModelCacheEntry : public QgsAbstractContentCacheEntry
```
wonder-sk, line 36: do we need to export this class?

src/3d/qgs3dmapscene.cpp
```diff
@@ -108,6 +109,18 @@ Qgs3DMapScene::Qgs3DMapScene( const Qgs3DMapSettings &map, QgsAbstract3DEngine *
...
+  {
+    const QList<QgsMapLayer *> pointVectorLayers = mPointVectorLayers;
+    for ( QgsMapLayer *layer : pointVectorLayers )
+    {
+      // remove old entity - if any
+      removeLayerEntity( layer );
+      // add new entity - if any 3D renderer
+      addLayerEntity( layer );
+    }
+  } );
```
wonder-sk, line 122: Would it be possible to put handling of the "model fetched" signal to qgspoint3dsymbol_p.cpp? I would like to avoid too many implementation details leaking into 3D map scene code...

Also, it looks like it would rebuild all point layers - even those that are not using that model at all...

src/3d/qgs3dmodelcache.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+Qt3DRender::QMesh *Qgs3DModelCache::pathAsMesh( const QString &path, bool blocking )
+{
+  Qt3DRender::QMesh *mesh = new Qt3DRender::QMesh;
+
+  const QString file = path.trimmed();
+  if ( file.isEmpty() )
+    return mesh;
+
+  QMutexLocker locker( &mMutex );
```
wonder-sk, line 75: Is it necessary to use mutex here? All creation of QEntity objects is done in the main thread (only loading of actual data to provide geometries/textures is done in worker threads)

wonder-sk (commented):
This will be a nice addition!

src/3d/qgs3dmodelcache.cpp
nirvn: @wonder-sk , is there no foreseeable scenario where that'll happen off the main thread? I thought it wouldn't hurt to have it there for such a day. If it's not going to happen, we can take it off.

src/3d/qgs3dmapscene.cpp
nirvn: @wonder-sk , I'm open to suggestions as to where you think we can plug in the model fetched signal in there. I couldn't see a way, but that definitively doesn't mean it can't be done :wink:

src/3d/qgs3dmodelcache.h
nirvn: @wonder-sk , initially, I thought we could actually cache the QMesh and QSceneLoader and actually gain some speed boost out of it. Turns out Qt doesn't allow copying of QMesh and QSCeneLoader objects at the moment. I'm trying to see if we can return copies of a cached QEntity. If that can't be done, I don't mind returning QUrl URIs, unless you think Qt will eventually allow copying of the above-mentioned objects, then it might make sense to keep things in its current form.

src/3d/qgs3dmapscene.cpp
nirvn: (actually on that, if we leave it in 3D map scene, we can narrow down the list of layers to be rebuild by only adding point layers with models as 3D symbols.

src/3d/qgs3dmodelcache.h
wonder-sk: I would be very cautious with caching/sharing of some Qt3D instances - some classes must not be even used multiple times as components of different entities. And especially with QSceneLoader-based point symbols I think we will need to change implementation eventually as it does not scale with the number of points. Currently for N points of a layer we create N QSceneLoader objects which is not optimal and I have seen it crash with higher number of points (at least on older version of Qt). The correct solution would be to load QSceneLoader once, traverse its internals for sub-entities (meshes with materials and transforms) and then apply instanced rendering like we do for simple 3d point shapes where the geometry is only stored in memory once, but rendered multiple times for single frame. Based on all this, my preference is not to expose QSceneLoader and just return URIs :-)

src/3d/qgs3dmapscene.cpp
wonder-sk: I would probably create a new class, e.g. QgsModelPoint3DSymbolEntity derived from QEntity. It could listen to the signal and refresh its internal QSceneLoader if URI matches. Then QgsModelPoint3DSymbolHandler::addSceneEntities() would just create instances of that class...

src/3d/qgs3dmodelcache.cpp
wonder-sk: I don't see that coming anytime soon - I don't see a reason why we would be going that way. As QEntity objects are derived from QObject, it is in my opinion also cleaner/simpler design to make them always in main thread rather than creating them in worker threads and then moving them to the main thread where they belong (because the scene root entity lives in main thread too). Another reason is that creation of entities should be always fairly quick as all loading should be done in workers run before QEntity gets created.

src/3d/symbols/qgspoint3dsymbol_p.cpp
```diff
@@ -469,20 +475,24 @@ void QgsModelPoint3DSymbolHandler::addMeshEntities( const Qgs3DMapSettings &map,
   // get nodes
   for ( const QVector3D &position : positions )
   {
-    // build the entity
-    Qt3DCore::QEntity *entity = new Qt3DCore::QEntity;
+    const QString source = QgsApplication::instance()->sourceCache()->localFilePath( symbol.shapeProperties()[QStringLiteral( "model" )].toString() );
+    if ( !source.isEmpty() )
```
wonder-sk, line 479: Could you please add a comment around here that if the source is remote, the Qgs3DMapScene will take care of refreshing this 3D symbol when the source is fetched? (so that it's clear we didn't give up, just passively waiting...)

src/3d/qgs3dmapscene.cpp
```diff
@@ -110,6 +113,41 @@ Qgs3DMapScene::Qgs3DMapScene( const Qgs3DMapSettings &map, QgsAbstract3DEngine *
...
 
+  connect( QgsApplication::instance()->sourceCache(), &QgsSourceCache::remoteSourceFetched, this, [ = ]( const QString & url )
+  {
+    const QList<QgsMapLayer *> modelVectorLayers = mModelVectorLayers;
+    for ( QgsMapLayer *layer : modelVectorLayers )
+    {
+      QgsAbstract3DRenderer *renderer = layer->renderer3D();
+      if ( renderer )
+      {
+        if ( renderer->type() == QLatin1String( "vector" ) ) //|| renderer->type() == QLatin1String( "rulebased" ) ) )
```
wonder-sk, line 124: commented out code

wonder-sk (approved):
Nice - thanks Mathieu for all the cleanups!

nirvn:
@wonder-sk , thanks, I'll adjust accordingly.

Do you have any idea as to why the windows build is *still* failing?

src/app/3d/qgs3dmodelsourcelineedit.h
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * \ingroup 3d
+ * \class Qgs3DModelSourceLineEdit
+ * A line edit widget with toolbutton for setting a 3D model source path.
+ *
+ * Designed for use with QgsSourceCache.
+ *
+ * \since QGIS 3.16
+ */
+class _3D_EXPORT Qgs3DModelSourceLineEdit : public QgsAbstractFileContentSourceLineEdit
```
wonder-sk, line 34: @nirvn this would be my guess why windows build fails... the superfluous _3D_EXPORT may be trying to force DLL import annotation and causing havoc

PR #37802, opened 2020-07-14
[layouts] Allow cell-based data-defined text format in the attribute table

nirvn:
## Description

This PR merely unlocks all the underlying work already done to enable data defined text format properties for every cell in layout attribute table items.

For the record, it allows to do cool stuff like:
Screenshot from 2020-07-14 14-38-58


tests/src/core/testqgslayouttable.cpp
```diff
@@ -1052,6 +1054,34 @@ void TestQgsLayoutTable::verticalGrid()
...
+  frame->setFrameEnabled( true );
+  l.addLayoutItem( frame );
+  table->addFrame( frame );
+
+  QgsTextFormat textFormat = QgsTextFormat::fromQFont( QgsFontUtils::getStandardTestFont( QStringLiteral( "Bold" ) ) );
+  table->setHeaderTextFormat( textFormat );
+  table->setContentTextFormat( textFormat );
+  table->refresh();
+
+  QCOMPARE( table->totalHeight(), 150.0 );
```
nyalldawson, line 1076: This is likely to be fragile cross platform -- could we get a render test instead?

nyalldawson:
Looks good!

tests/src/core/testqgslayouttable.cpp
nirvn: I was trying to avoid a render test altogether. You think there's no way around it?

nyalldawson: I think it's preferable in this case -- it would also cover the actual rendering code then

PR #37813, opened 2020-07-14
 Move the "Discourage diagrams from covering features" setting

DelazJ:
from Rendering tab to Placement tab.
Makes it inline with labels dialog, and next to the related priority rank setting

image

### DIFF
--- a/src/ui/qgsdiagrampropertiesbase.ui
+++ b/src/ui/qgsdiagrampropertiesbase.ui
@@ -968,37 +968,6 @@
                          </item>
                         </layout>
                        </item>
-                       <item row="4" column="0" colspan="3">
-                        <layout class="QHBoxLayout" name="horizontalLayout_12">
-                         <item>
-                          <widget class="QLabel" name="label_8">
-                           <property name="text">
-                            <string>Discourage diagrams and labels from covering features</string>
-                           </property>
-                          </widget>
-                         </item>
-                         <item>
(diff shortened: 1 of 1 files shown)


src/ui/qgsdiagrampropertiesbase.ui
```diff
@@ -1619,6 +1578,51 @@
...
+                     <property name="leftMargin">
+                      <number>8</number>
+                     </property>
+                     <property name="rightMargin">
+                      <number>8</number>
+                     </property>
+                     <item>
+                      <widget class="QLabel" name="label_8">
+                       <property name="text">
+                         <string>Discourage diagrams and labels from covering features</string>
```
DelazJ, line 1599: @nyalldawson why are labels mentioned here? Do they not have their settings in the labels --> Placement tab already?

nyalldawson: There's no distinction between the two when placing labels -- so if you had diagrams switched on for a layer and have this setting enabled for the diagrams, then it will also block labels from covering the features, not just diagrams.

DelazJ: And if I don't set this one but set the "discourage.." in the labels tab, would it prevent diagrams from covering the feature? IOW, should we add _diagrams_ to the corresponding label in Labels tab?

nyalldawson: Yes, that's correct

PR #37814, opened 2020-07-14
Fix FindPyQt5 cmake and python scripts

dminor:
## Description

As written, the FindPyQt5.cmake and FindPyQt5.py scripts will succeed even if
PyQt5 is not installed. In FindPyQt5.cmake, we have PYQT5_FIND_REQUIRED rather
than PyQt5_FIND_REQUIRED, so the package is always treated as though it is
optional. In FindPyQt5.py, the script will succeed with a default location for
the sip directory, even if the PyQt5 directory is not located.

For some systems, if enough other dependencies are installed, the build can
get about halfway complete without the main PyQt5 dependency, and then fail
with a message like:

sip: Unable to find file "QtXml/QtXmlmod.sip"

I hit this building on Ubuntu, and there have been a few similar bug reports in
the past.


m-kuhn (approved):
Very good move!

PR #37826, opened 2020-07-15
allow to customize convertToCurves parameters for tracing curves

olivierdalang:
## Description

This mitigates an issue with the new (3.14) tracing curves feature, where the default tolerance is too low for curves to be correctly detected (especially when working far from the origin).

This PR sets the default angle/distance tolerance to 1e-6 when tracing curves (this matches the default of the processing algorithm, the default of `QgsGeometry.convertToCurves()` being 1e-8).

Of course there's no one value fits all when it comes to tolerance, so it is now customizable via a QgsSettings.

Open questions :
- do we want to these settings to the configuration panel (under Digitizing/Tracing), or is it too technical ?
- would it make sense to change the defaults of `QgsGeometry.convertToCurves()` to 1e-6 too ?

Cheers !


m-kuhn:
Thanks,
I would like to ship that with 3.14.1 so we can run some tests and band aid problematic scenarios.

And in parallel assess if

a) we can come up with an approach that chooses the tolerance based on the precision of input coordinates
b) we can moves geometries towards the origin
c) we should expose this configuration in the UI

m-kuhn (approved):

PR #37830, opened 2020-07-15
Fix pasting features - only a single insert operation, no unnecessary (failing) updates

suricactus:
When pasting features from another layer, the current implementation calls both `QgsVectorLayer::addFeatures` and `QgsVectorLayer::changeAttributeValue`. This results in both `INSERT` and `UPDATE` operations, but the latter is unnecessary, especially updating a feature with not yet known feature id. 

The error has not been caught yet, because in case of invalid feature ID for `UPDATE`, it results in a query `UPDATE ... WHERE fid = 0`, which is executes and does nothing (unless someone is unlucky enough). However, the error was caught with the PostGIS provider and `int64` id column, because the `WHERE` clause remains empty string, which causes a SQL error.

I am currently investigating the possibility to have a test case, but I am done with the fix itself.

### DIFF
--- a/python/gui/auto_generated/qgsattributeeditorcontext.sip.in
+++ b/python/gui/auto_generated/qgsattributeeditorcontext.sip.in
@@ -31,6 +31,7 @@ showing an embedded form due to relations)
     {
       SingleEditMode,
       AddFeatureMode,
+      FixAttributeMode,
       MultiEditMode,
       SearchMode,
       AggregateSearchMode,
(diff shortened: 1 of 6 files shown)


gioman:
@suricactus was this also the cause of the slowness when copy/pasting features between tables/layers?

suricactus:
@gioman there is a fair performance improvement, but it still feels slow to me.

@signedav can you have a look on this fix?

signedav (approved):
Looks reasonable. Thanks a lot.

m-kuhn:
@suricactus can you run astyle on the changed files?

PR #37832, opened 2020-07-15
Browser create database table

elpaso:
This is the second (and last) part of QEP https://github.com/qgis/QGIS-Enhancement-Proposals/issues/171

By providing a generic dialog to create new DB vector (and aspatial) tables, this PR enables "New Table..." context menu item for all providers that implements the connections API and have the required capabilities.

The PR is not yet ready for merge because it needs some minor polishing and some more testing but the implementation is feature complete.

qgis-browser-new-table


src/gui/qgsnewvectortabledialog.cpp
```diff
@@ -0,0 +1,753 @@
...
+  }
+  // Check for field names and geom col name
+  if ( isSpatial && fields().names().contains( mGeomColumn->text(), Qt::CaseSensitivity::CaseInsensitive ) )
+  {
+    mValidationErrors.push_back( tr( "Geometry column name cannot be equal to an existing field name!" ) );
+  }
+  // No geometry and no fields? No party!
+  if ( ! isSpatial && mFieldModel->fields().count() == 0 )
+  {
+    mValidationErrors.push_back( tr( "The table has no geometry column and no fields: cannot create and empty table!" ) );
```
DelazJ, line 325: Spotted from your cast ;)
```suggestion
    mValidationErrors.push_back( tr( "The table has no geometry column and no fields: cannot create an empty table!" ) );
```

src/core/qgsabstractdatabaseproviderconnection.cpp
```diff
@@ -226,6 +227,101 @@ QgsAbstractDatabaseProviderConnection::TableProperty QgsAbstractDatabaseProvider
                                         .arg( name, schema ) );
 }
 
+QList<QgsVectorDataProvider::NativeType> QgsAbstractDatabaseProviderConnection::nativeTypes() const
```
nyalldawson, line 230: This method is rather ugly -- wouldn't it be much less work/more reliable to move the native types methods from the various providers to a static function and use that from the connection implementations?

src/ui/qgsnewvectortabledialogbase.ui
```diff
@@ -0,0 +1,416 @@
...
+      <widget class="QgsProjectionSelectionWidget" name="mCrs" native="true"/>
+     </item>
+    </layout>
+   </item>
+   <item row="1" column="1">
+    <layout class="QVBoxLayout" name="mButtonsLayout">
+     <item>
+      <widget class="QPushButton" name="mAddFieldBtn">
+       <property name="text">
+        <string>Add field</string>
```
nyalldawson, line 247:
```suggestion
        <string>Add Feld</string>
```

src/ui/qgsnewvectortabledialogbase.ui
```diff
@@ -0,0 +1,416 @@
...
+       <property name="icon">
+        <iconset resource="../../images/images.qrc">
+         <normaloff>:/images/themes/default/mActionNewAttribute.svg</normaloff>:/images/themes/default/mActionNewAttribute.svg</iconset>
+       </property>
+      </widget>
+     </item>
+     <item>
+      <widget class="QPushButton" name="mDeleteFieldBtn">
+       <property name="text">
+        <string>Delete field</string>
```
nyalldawson, line 258:
```suggestion
        <string>Remove Field</string>
```

src/gui/qgsnewvectortabledialog.cpp
```diff
@@ -0,0 +1,753 @@
...
+    mCrs->setEnabled( hasGeom );
+    mCrsLabel->setEnabled( hasGeom );
+    mDimensionsLabel->setEnabled( hasGeom );
+    mHasMChk->setEnabled( hasGeom );
+    mHasZChk->setEnabled( hasGeom );
+    validate();
+  } );
+
+  // Hardcode geometry types
+  // TODO: this information should really come from the connection through the provider but there is no API for that yet
```
nyalldawson, line 126: Good idea -- definitely something we should consider adding in future. We should also consider curved types support too (e.g. via a capability flag for "supportscurves"), and z / m support.

nyalldawson: Also a flag for "provider is not sensitive to multi/single polygon/line distinction"

src/gui/qgsnewvectortabledialog.cpp
```diff
@@ -0,0 +1,753 @@
...
+        {
+          fieldList.append( fields().at( i ) );
+        }
+      }
+      setFields( fieldList );
+      selectRow( mCurrentRow - 1 );
+    }
+  } );
+
+  connect( mFieldDownBtn, &QPushButton::clicked, this, [ = ]
```
nyalldawson, line 192: I like this approach of putting the lambdas directly in connections for button actions like this, good move

nyalldawson (commented):
Great work!

src/core/qgsabstractdatabaseproviderconnection.cpp
elpaso: @nyalldawson I've thought about this a lot, the reason why it needs to be dinamic is the OGR provider: being it a multi-storage provider the nativeTypes function can only be executed on a layer instance because it needs to know the GDAL/OGR driver.

The alternative is to implement it in the metadata/connections API but in that case there would be a duplication of code to return the hardcoded list of native types, and I'd like to avoid that.

The assumption here is that the user is going to create a new table so he must have permissions to do so or it will fail post-mortem when the dialog is accepted anyway so we should be on the safe side (functionally speaking).

src/gui/qgsnewvectortabledialog.cpp
elpaso: Yeah, I was asking myself if I was abusing it but I like to have the action code close to the connection, I guess the only reason to switch the lambda with an independent function is if it's called from multiple places.

src/gui/qgsnewvectortabledialog.cpp
elpaso: Is there any database provider in the connections API that would have this flag right now?
IIRC we currently support:
- postgis
- gpkg
- spatialite
- mssql

src/core/qgsabstractdatabaseproviderconnection.cpp
nyalldawson: Hm - good point. But I still think it's a generally ugly approach for something like postgres. The postgres situation is rather simple -- we just extract the lump of code from qgspostgresprovider.cpp lines 228->270 and move that to a static function, which we call from both qgspostgresprovider.cpp and qgspostgresproviderconnection.cpp

Similar applies to mssql and spatialite. Really OGR is the odd one out here, so I'd only do the fallback approach for that provider.

elpaso: If you are ok with duplicating code, I'll go that way. Another point I was also thinking at is the possibility that different endpoints of the same provider support different data types for example if there is a particular postgres extension installed. That would be another reason to keep it dynamic.

elpaso: I've just checked pg provider and there is already code that expose different fields for different versions of postgres.

nyalldawson: Why is it duplicating code? It's moving the existing code to a re-usable standalone function...

src/gui/qgsnewvectortabledialog.cpp
nyalldawson: SQL server doesn't care, that's one...

elpaso: How about in QgsAbstractDatabaseProviderConnection:

```
    /**
     * The GeometryColumnCapability enum represents the geomery column features supported by the connection
     * \since QGIS 3.16
     */
    enum GeometryColumnCapability
    {
      Z = 1 << 1,                    //! Supports Z dimension
      M = 1 << 2,                    //! Supports M dimension
      SinglePart = 1 << 3            //! Multi and single part types are distinct geometry types
    };
```

Same question as above: can it be static or we'd expect different flags from different backends?

src/core/qgsabstractdatabaseproviderconnection.cpp
elpaso: 1. we cannot make `QList<QgsVectorDataProvider::NativeType> QgsVectorDataProvider::nativeTypes() const` static because it returns different values depending on the driver for OGR and depending on PG version and extensions for postgres
2. we cannot make it static in `QgsAbstractDatabaseProviderConnection` either for the same reason, unless ...

... if we settled for a subset of supported field types we could make it static in `QgsAbstractDatabaseProviderConnection` but that would mean to hardcode the list of native types in the function implementation, this is duplication of code because the "sister" non-static function `QgsVectorDataProvider::nativeTypes()` would need to return a similar (but dynamic) list.

This is not only duplication of code but we also risk that they eventually get out of sync.

src/gui/qgsnewvectortabledialog.cpp
nyalldawson: It's going to be static -- it depends on the actual backend's capabilities and QGIS' provider functionality. But it won't vary connection-by-connection for one particular provider.

We'll also need a "supports curves" flag

src/core/qgsabstractdatabaseproviderconnection.cpp
elpaso: static methods done!

nyalldawson:
New changes are looking good!

nirvn:
@elpaso , a warning needs addressing:

/home/webmaster/dev/cpp/QGIS/src/gui/qgsnewvectortabledialog.cpp: In member function ‘virtual QVariant QgsNewVectorTableFieldModel::data(const QModelIndex&, int) const’:
/home/webmaster/dev/cpp/QGIS/src/gui/qgsnewvectortabledialog.cpp:542:9: warning: this statement may fall through [-Wimplicit-fallthrough=]
  542 |         switch ( static_cast<ColumnHeaders>( index.column() ) )
      |         ^~~~~~
/home/webmaster/dev/cpp/QGIS/src/gui/qgsnewvectortabledialog.cpp:570:7: note: here
  570 |       case Qt::ItemDataRole::TextAlignmentRole:
      |       ^~~~

elpaso:
I'm on holidays until the end of the week, I'll take care of it when I'm back.

src/core/providers/ogr/qgsgeopackageproviderconnection.cpp
```diff
@@ -380,3 +408,14 @@ QList<QVariantList> QgsGeoPackageProviderConnection::executeGdalSqlPrivate( cons
...
 
+
+QList<QgsVectorDataProvider::NativeType> QgsGeoPackageProviderConnection::nativeTypes() const
+{
+  QList<QgsVectorDataProvider::NativeType> types;
+  QgsVectorLayer vl { uri(), QStringLiteral( "temp_layer" ), QStringLiteral( "ogr" ) };
+  if ( ! vl.isValid() || ! vl.dataProvider() )
+  {
+    throw QgsProviderConnectionException( QObject::tr( "Error retrieving native types for %1: %2" ).arg( uri() ).arg( vl.dataProvider()->errors().join( '\n' ) ) );
+  }
```
uclaros, line 419: @elpaso in the case of `! vl.dataProvider()` you're still trying to access `vl.dataProvider()->errors()` which causes QGIS to crash.

elpaso: Yeah, of course :/ I'll fix it. Thank you for notifying.

PR #3532, opened 2016-09-27
Add auto-discovery of relations for PostgresQL and SpatiaLite

pvalsecc:
Fixed the add relation functionnality: the table is sorted. When the code
was setting the sorted column, the row was sorted and the other columns it was
setting were set on the wrong row.

PR #37833, opened 2020-07-15
Fix discover relations

lbartoletti:
## Description

I don't know why [this query](https://github.com/qgis/QGIS/pull/3532) was used rather than another, but we found a case where not all relations are discovered. I propose this query (maybe there are others doing even better?) to discover all relations.

Context:

[test_relations.zip](https://github.com/qgis/QGIS/files/4931663/test_relations.zip)

Relations discovered by QGIS 3.14:
discover_relations_before

With this query:
after

cc @sylvainbeo


m-kuhn:
@lbartoletti Would you be able to provide details about the schema in question? Or if possible even a test case?
I'm not in the situation to review the SQL on a purely textual basis.

lbartoletti:
> @lbartoletti Would you be able to provide details about the schema in question? Or if possible even a test case?
> I'm not in the situation to review the SQL on a purely textual basis.

Ah ah... I uploaded the database in .sql file format but forgot than Github doesn't allow it... It's fixed with the zip attached.

m-kuhn:
Would it be possible to add this as a test?

lbartoletti:
> Would it be possible to add this as a test?

Sure I will, I waited your point of view before since as mentioned, I really don't know if there's a reason I missed where Qgis doesn't have to list all the relationships or if it was a mistake in the query.

m-kuhn:
If that was the question: I have no idea and improving this sounds good even more if the detection is backed up by unit tests.

This was originally added by @pvalsecc IIRC.

tests/src/python/test_qgsrelationpostgres.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+	cls.mapCanvas = QgsMapCanvas()
+        cls.dbconn = 'service=\'qgis_test\''
+        if 'QGIS_PGTEST_DB' in os.environ:
+            cls.dbconn = os.environ['QGIS_PGTEST_DB']
+        # Create test layer
+	tables = [ 'c_amgmt_amgmt_lot', 'c_batiment_bat_lot', 'c_ens_immo_amgmt', 'c_ens_immo_bat', 'c_terrain_ens_immo', 't_actes', 't_adresse', 't_amgmt', 't_amgmt_lot', 't_bat', 't_bat_lot', 't_ens_immo', 't_terrain' ]
+	vl_tables = ['vl_c_amgmt_amgmt_lot','vl_c_batiment_bat_lot','vl_c_ens_immo_amgmt','vl_c_ens_immo_bat','vl_c_terrain_ens_immo','vl_t_actes','vl_t_adresse','vl_t_amgmt','vl_t_amgmt_lot','vl_t_bat','vl_t_bat_lot','vl_t_ens_immo','vl_t_terrain' ]
+
+	for i in range(len(tables)):
```
m-kuhn: Does this still belong to `setUpClass`? I think the indentation is not properly aligned.

tests/src/python/test_qgsrelationpostgres.py
```diff
@@ -0,0 +1,61 @@
...
+            QgsProject.instance().addMapLayer(cls.vl_tables[i])
+
+        cls.relMgr = QgsProject.instance().relationManager()
+
+    def test_discover_relations(self):
+        """
+        Test the automatic discovery of relations
+        """
+        relations = self.relMgr.discoverRelations([], self.vl_tables)
+        assert(len(relations) == 18)
```
m-kuhn, line 57:
```suggestion
        self.assertEqual(len(relations), 18)
```

m-kuhn (approved):
I trust a) you thoroughly tested and b) the unit tests

PR #37837, opened 2020-07-16
Functions Help improvements

SrNetoChan:
## Description

This PR does:
- removes obsolete help files
- adds missing groups help files
- fix some files to work better with https://github.com/qgis/QGIS-Documentation/pull/5811


nyalldawson (approved):

resources/function_help/json/format_date
```diff
@@ -5,7 +5,7 @@
...
  {"arg":"language","optional":true,"description":"language (lowercase, two- or three-letter, ISO 639 language code) used to format the date into a custom string"}
```
DelazJ, line 9: It could be nice to know the default language used (system, settings dialogue or en?), since it's optional

resources/function_help/json/CASE
```diff
@@ -1,5 +1,14 @@
...
+  "groups": ["Conditionals"],
+  "description": "The CASE expression evaluates several conditions and returns a result when the first condition is met. When a condition is true, it will stop reading and return the corresponding result. If none of the conditions is true, it returns the value in the ELSE clause. Furthermore, if no ELSE clause is set and none of the conditions is met, it returns NULL.<br><pre>CASE<br>WHEN <i>condition</i> THEN <i>result</i><br>[ …n ]<br>[ ELSE <i>result</i> ]<br>END</pre>[ ] marks optional components<br>",
+  "arguments": [
+   {"arg":"WHEN condition","description":"The condition expression to evaluate"},
+   {"arg":"THEN result","description":"If <i>condition</i> evaluates to True then <i>result</i> is evaluated and returned."},
+   {"arg":"ELSE result","description":"If none of the above conditions evaluated to True then <i>result</i> is evaluated and returned."}
+  ],
+  "examples": [
+   { "expression":"CASE WHEN \"column\" IS NULL THEN 'None' END", "returns":" Returns the string 'none' if the \"column\" field is NULL"}
```
DelazJ, line 12:
```suggestion
   { "expression":"CASE WHEN \"name\" IS NULL THEN 'None' END", "returns":" Returns the string 'none' if the \"name\" field is NULL"},
   { "expression":"CASE WHEN $area > 10000 THEN 'Big property' WHEN $area > 5000 THEN 'Medium property' ELSE 'Small property' END", "returns":" Returns the string 'Big property' if the area is bigger than 10000, 'Medium property' if the area is between 5000 and 10000, and 'Small property' for others"}
```
I find "column field" a bit ugly.
And suggest an ELSE use (could be helpful - feel free to adjust)

DelazJ (commented):
I was wondering: 
- is there functions for modeler contexts (no idea, just asking!) or are they also processing ones?
- (not necessarily related to this PR) when working with colors or date, the components such as red, green, hour, month to retrieve particular value should be in English or can be translated. Eg could color_part('200,10,30','green') --> color_part('200,10,30','vert') ? I have no date example on top of my head although that would better show my concerns, particularly because they handle language. No qgis around to test so asking.  And I'm asking because would it be necessary to mention these caution/information to translators/users?

resources/function_help/json/age
```diff
@@ -2,10 +2,9 @@
   "name": "age",
   "type": "function",
   "groups": ["Date and Time"],
-  "description": "Returns the difference between two dates or datetimes.<br>The difference is returned as a <code>Interval</code> and needs to be used with one of the following functions in order to extract useful information:<br /><ul><li><code>year</code><li><code>month</code><li><code>week</code><li><code>day</code><li><code>hour</code><li><code>minute</code><li><code>second</code></ul>",
+  "description": "Returns the difference between two dates or datetimes.<br>The difference is returned as an <code>Interval</code> and needs to be used with one of the following functions in order to extract useful information:<br /><ul><li><code>year</code></li><li><code>month</code></li><li><code>week</code></li><li><code>day</code></li><li><code>hour</code></li><li><code>minute</code></li><li><code>second</code></ul>",
```
havatv, line 5:
```suggestion
  "description": "Returns the difference between two dates or datetimes.<br>The difference is returned as an <code>Interval</code> and needs to be used with one of the following functions in order to extract useful information:<br /><ul><li><code>year</code></li><li><code>month</code></li><li><code>week</code></li><li><code>day</code></li><li><code>hour</code></li><li><code>minute</code></li><li><code>second</code></li></ul>",
```

resources/function_help/json/Form
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Form",
+  "type": "group",
+  "description": "This group contains functions that operate exclusively under the attribute form context. For example, in field's widgets settings."
```
havatv, line 4:
```suggestion
  "description": "This group contains functions that are only available in the attribute form context. For example, in field widget settings."
```

resources/function_help/json/format_date
SrNetoChan: This is a bit out of scope for this PR, unless anyone knows.

resources/function_help/json/Processing
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Processing",
+  "type": "group",
+  "description": "This group contains functions that operate exclusively under the processing algorithms context."
```
havatv, line 4:
```suggestion
  "description": "This group contains functions that are only available in the processing algorithms context."
```

havatv (commented):
I would like to know why these function help files ($atlasfeature, $atlasfeatureid, $atlasgeometry, $feature, $numfeatures, $numpages) are being removed. What they seem to have in common is that they operate on atlases / reports.
They still exist don't they (I have used them quite recently, and found the current help essential)? If they still exist - could we perhaps add them to the new "Layout" group?

    "groups": ["Layout"]

SrNetoChan:
> I would like to know why these function help files ($atlasfeature, $atlasfeatureid, $atlasgeometry, $feature, $numfeatures, $numpages) are being removed. What they seem to have in common is that they operate on atlases / reports.
> They still exist don't they (I have used them quite recently, and found the current help essential)? If they still exist - could we perhaps add them to the new "Layout" group?
> 
>     "groups": ["Layout"]

@havatv I believe those functions still work for backwards compatibility, but they have been replaced by variables. If you check, they are no longer listed on on the GUI. Makes no sense to keep the in the docs.

resources/function_help/json/CASE
```diff
@@ -1,5 +1,15 @@
 {
   "name": "CASE",
   "type": "expression",
-  "description": "<pre><br>CASE<br>WHEN <i>condition</i> THEN <i>result</i><br>[ …n ]<br>[ ELSE <i>result</i> ]<br>END<br></pre>[ ] marks optional components<br><h4>Arguments</h4><br><i>  WHEN condition</i> - The condition expression to evaluate. <br><br><i>  THEN result</i> - If <i>condition</i> evaluates to True then <i>result</i> is evaluated and returned. <br><br><i>  ELSE result</i> - If none of the above conditions evaluated to True then <i>result</i> is evaluated and returned.<br><br><h4>Example</h4><pre>CASE<br>WHEN <i>\"column\" IS NULL</i> THEN <i>'None'</i><br>END</pre>"
+  "groups": ["Conditionals"],
+  "description": "The CASE expression evaluates several conditions and returns a result when the first condition is met. When a condition is true, it will stop reading and return the corresponding result. If none of the conditions is true, it returns the value in the ELSE clause. Furthermore, if no ELSE clause is set and none of the conditions is met, it returns NULL.<br><pre>CASE<br>WHEN <i>condition</i> THEN <i>result</i><br>[ …n ]<br>[ ELSE <i>result</i> ]<br>END</pre>[ ] marks optional components<br>",
```
havatv, line 5:
```suggestion
  "description": "CASE is used to evaluate a series of conditions and return a result for the first condition that is met. The conditions are evaluated sequentially, and if a condition is true, the evaluation will stop, and the corresponding result will be returned. If none of the conditions are true, the value in the ELSE clause is returned. Furthermore, if no ELSE clause is set and none of the conditions are met, NULL is returned.<br><pre>CASE<br>WHEN <i>condition</i> THEN <i>result</i><br>[ …n ]<br>[ ELSE <i>result</i> ]<br>END</pre>[ ] marks optional components<br>",
```

resources/function_help/json/CASE
```diff
@@ -1,5 +1,15 @@
...
+  "arguments": [
+   {"arg":"WHEN condition","description":"The condition expression to evaluate"},
```
havatv, line 7:
```suggestion
   {"arg":"WHEN condition","description":"A condition expression to evaluate"},
```

resources/function_help/json/Form
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Form",
+  "type": "group",
+  "description": "This group contains functions that are only available in the attribute form context. For example, in field widget settings."
```
havatv, line 4:
```suggestion
  "description": "This group contains functions that are available in the attribute form context. For example, in field widget settings."
```

resources/function_help/json/Processing
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Processing",
+  "type": "group",
+  "description": "This group contains functions that are only available in the processing algorithms context."
```
havatv, line 4:
```suggestion
  "description": "This group contains functions that are available in the processing algorithms context."
```

havatv (approved):
I have a few suggestions.

It would be nice if this could be merged soon, since a part of the documentation is generated from these files, and without these fixes, we get errors when building the docs.

nyalldawson:
@DelazJ 
>  the components such as red, green, hour, month to retrieve particular value should be in English or can be translated

No, they cannot be

nyalldawson:
> No, they cannot be

To expand on this -- if we translated them, then it would potentially break projects which are moved between different locales

PR #37840, opened 2020-07-16
Add options to dynamically tweak dash pattern in simple line symbol layers

nyalldawson:
New options are:

- Align dash pattern to line length: If checked, the dash pattern lengths
will be subtely adjusted in order to ensure that when a line is rendered
it will end with a complete dash element, instead of a gap element or
partial dash element

- Tweak dash pattern at sharp corners: If checked, this option dynamically
adjusts the dash pattern placement so that sharp corners are represented
by a full dash element coming into and out of the sharp corner. It's designed
to better represent the underlying geometry while rendering dashed lines,
especially for jagged lines

Screenshots:

Before, showing yellow lines at actual start and end of features -- the dash pattern ends with a gap, which hides the true length of the line:

image

With the new "Align dash pattern to line length" option checked:

image

Here's an animation showing the new "tweak dash pattern at sharp corners" option being toggled:

Peek 2020-07-16 15-25


src/core/symbology/qgslinesymbollayer.cpp
```diff
@@ -615,6 +633,293 @@ void QgsSimpleLineSymbolLayer::applyDataDefinedSymbology( QgsSymbolRenderContext
...
+void QgsSimpleLineSymbolLayer::drawPathWithDashPatternTweaks( QPainter *painter, const QPolygonF &points, QPen pen ) const
+{
+  if ( pen.dashPattern().empty() || points.size() < 2 )
+    return;
+
+  QVector< qreal > sourcePattern = pen.dashPattern();
+  const double dashWidthDiv = std::max( 1.0, pen.widthF() );
+  // back to painter units
+  for ( int i = 0; i < sourcePattern.size(); ++ i )
+    sourcePattern[i ] *= dashWidthDiv;
```
nirvn, line 645:
```suggestion
    sourcePattern[i] *= dashWidthDiv;
```

src/core/symbology/qgssymbollayerutils.h
```diff
@@ -648,6 +648,36 @@ class CORE_EXPORT QgsSymbolLayerUtils
...
+     * from the end of the line.
+     *
+     * May return an empty linestring if the substring is zero length.
+     *
+     * \since QGIS 3.16
+     */
+    static QPolygonF polylineSubstring( const QPolygonF &polyline, double startOffset, double endOffset );
+
+    /**
+     * Returns TRUE if the angle formed by the line \a p1 - \a p2 - \a p3 forms a "sharp" corner.
```
nirvn, line 666: Might be worth adding in the dox what minimum angle is considered to be a "sharp" corner.

src/gui/symbology/qgssymbollayerwidget.cpp
```diff
@@ -323,6 +344,12 @@ void QgsSimpleLineSymbolLayerWidget::setSymbolLayer( QgsSymbolLayer *layer )
 
   whileBlocking( mRingFilterComboBox )->setCurrentIndex( mRingFilterComboBox->findData( mLayer->ringFilter() ) );
 
+  whileBlocking( mCheckAlignDash )->setChecked( mLayer->alignDashPattern() );
+  mCheckDashCorners->setEnabled( mLayer->alignDashPattern() );
+  whileBlocking( mCheckDashCorners )->setChecked( mLayer->tweakDashPatternOnCorners() );
+  if ( !mLayer->alignDashPattern() )
+    mCheckDashCorners->setChecked( false );
```
nirvn, line 351: Why not adding this condition in the whileBlocking( mCheckDashCorners )->setChecked( ... ) line above?

`whileBlocking( mCheckDashCorners )->setChecked( mLayer->tweakDashPatternOnCorners() && !mLayer->alignDashPattern() ) );`

nirvn (commented):
That'll come in handy.

nyalldawson:
Thanks for the review @nirvn!

Issue #37542, opened 2020-07-02
Please switch from sip4 to sip5


sebastic:
As reported by Dmitry Shachnev in [Debian Bug #964132](https://bugs.debian.org/964132):
> Recently I have updated pyqt5 and related packages to use sip5 instead of
> sip4 for build. This is in experimental for now, but I want to upload it
> to unstable in a month or two.
>
> This change breaks PyQt5 extension ABI, so all packages build-depending on
> pyqt5-dev need to build with sip5 too.
>
> Quick migration guide:
>
> 1) Build-depend on python3-sipbuild(-dbg) and sip5-tools.
>
>    - Inspect pyproject.toml (if it exists) for additional dependencies, maybe
>      you will need also python3-pyqtbuild.
>
> 2) If your package has a project.py file, call sip-build to build it.
>    For separation of configure and build phases, pass --no-make to it.
>
> 3) If your package does not have a project.py file, but has a configure.py,
>    update that file to the latest version of the boilerplate (v2.16), which
>    you can find in the experimental version of qscintilla2 package.
>
>    (Qscintilla2's configure.py used to be a recommended way to build sip
>    projects.)
>
> 4) If your package called /usr/bin/sip directly, call /usr/bin/sip5 instead.
>
> 5) If your package had a runtime dependency on python3-sip or ${sip3:Depends},
>    depend on python3-pyqt5.sip instead. There is no analogue for dh_sip3
>    helper yet.
>
>    (Upstream now recommends each project to use its own private copy of sip.
>    In Debian, I recommend you to use pyqt5's copy which is python3-pyqt5.sip).
>
> 6) If you build a -dev package with .sip files, they are now installed into
>    a new location (in dist-packages). Also .toml files now should be present.
>
> You can see what I have done in pyqt5, pyqt5chart, pyqt5webengine and
> qscintilla2 packages in experimental (see the relevant commits in git).
>
> The official documentation for sip5 is available online [1] or in sip5-doc
> package. For help you can ask me or pyqt/sip mailing list [2] (subscription
> needed to post messages).
>
> Please stage your changes in experimental (or in a VCS) for now. They will
> need to be uploaded to unstable together with pyqt5.
>
> [1]: https://www.riverbankcomputing.com/static/Docs/sip/
> [2]: https://www.riverbankcomputing.com/mailman/listinfo/pyqt

PR #37842, opened 2020-07-16
Get rid of SIP deprecated functions

mitya57:
- `sipMappedType` struct was deprecated in SIP v4.8 in favour of `sipTypeDef`. The related functions should be changed as follows:
    - `sipFindMappedType` → `sipFindType`;
    - `sipCanConvertToMappedType` → `sipCanConvertToType`;
    - `sipConvertToMappedType` → `sipConvertToType`;
    - `sipReleaseMappedType` → `sipReleaseType`.

- `sipCanConvertToEnum` was deprecated in SIP v4.19.4. The documentation suggests to use `sipConvertToEnum` directly. In QGIS case, that function was used to raise a `TypeError` exception when the type does not match, however `sipConvertToEnum` [raises a `TypeError` itself][1] so that code is not needed.

- `sipTransferBreak` was deprecated in SIP v4.14. The documentation says to [use `sipTransferTo(obj, NULL);` instead][2] (available since v4.3), which is what I have done here.

All deprecated functions were removed in SIP v5.0, so this is needed in preparation for SIP v5 support (#37542). I plan to add actual support in the next pull request.

This does not change minimal SIP dependency. The code is already using `sipConvertToEnum` which means that SIP v4.19.4 is required. (In fact, `sipConvertToEnum` was added together with deprecating `sipCanConvertToEnum`.)

[1]: https://riverbankcomputing.com/hg/sip/file/853c4a8b4592/siplib/siplib.c#l7705
[2]: https://riverbankcomputing.com/hg/sip/file/853c4a8b4592/sphinx/c_api.rst#l2453

### DIFF
--- a/python/core/auto_generated/geometry/qgsgeometry.sip.in
+++ b/python/core/auto_generated/geometry/qgsgeometry.sip.in
@@ -1533,7 +1533,7 @@ is null, a ValueError will be raised.
     }
     else
     {
-      const sipMappedType *qvector_type = sipFindMappedType( "QVector<QgsPointXY>" );
+      const sipTypeDef *qvector_type = sipFindType( "QVector<QgsPointXY>" );
       sipRes = sipConvertFromNewType( new QVector< QgsPointXY >( sipCpp->randomPointsInPolygon( a0, a1 ) ), qvector_type, Py_None );
     }
 %End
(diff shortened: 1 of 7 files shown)


nyalldawson (approved):

nyalldawson:
Thanks -- this is much appreciated!

lbartoletti:
Theses commits should be backported since are needed by backports #37901 (tested locally on FreeBSD with qgis 3.14.15)

PR #37869, opened 2020-07-18
createUserProfile: ensure qgis.db file is user writable

lsix:
## Description

In some package managers (such as NIX[1], used in NixOS[2]), packaged
data is installed read-only.

When creating a new user profile, QGis copies the `qgis.db` file from
the packages and the user ends having a read-only `qgis.db` database.

This commit ensures that once copied when creating a ne profile, the
`qgis.db` file is made user writable.

Since we are running QGIS LTS in nixpkgs (currently QGIS-3.10.7), it would be greatly appreciated if this could be back ported.

[1] https://nixos.org/nix/
[2] https://nixos.org/

### DIFF
--- a/src/core/qgsuserprofilemanager.cpp
+++ b/src/core/qgsuserprofilemanager.cpp
@@ -156,6 +156,17 @@ QgsError QgsUserProfileManager::createUserProfile( const QString &name )
 
     //now copy the master file into the users .qgis dir
     masterFile.copy( qgisPrivateDbFile.fileName() );
+
+    // In some packaging systems, the master can be read-only. Make sure to make
+    // the copy user writable.
+    const QFile::Permissions perms = QFile( qgisPrivateDbFile.fileName() ).permissions();
+    if ( !( perms & QFile::WriteOwner ) )
+    {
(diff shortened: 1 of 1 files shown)


src/core/qgsuserprofilemanager.cpp
```diff
@@ -156,6 +156,17 @@ QgsError QgsUserProfileManager::createUserProfile( const QString &name )
 
     //now copy the master file into the users .qgis dir
     masterFile.copy( qgisPrivateDbFile.fileName() );
+
+    // In some packaging systems, the master can be read-only. Make sure to make
+    // the copy user writable.
+    QFile::Permissions const perms = QFile( qgisPrivateDbFile.fileName() ).permissions();
```
m-kuhn, line 162:
```suggestion
    const QFile::Permissions perms = QFile( qgisPrivateDbFile.fileName() ).permissions();
```

m-kuhn:
Looks good to me, just a small coding style suggestion.
Backporting sounds fine to me too.
Thank you.

src/core/qgsuserprofilemanager.cpp
lsix: Changed.

Thanks for the suggestion.

m-kuhn (approved):

PR #37883, opened 2020-07-20
Don't load GRASS algs if GRASS is not installed

### LINKED ISSUE #37877: qgis_process incorrectly lists GRASS algorithms even if GRASS is not installed
**Describe the bug**

The `qgis_process list` command incorrectly lists the GRASS algorithms among the available processing algorithms even when GRASS is not installed in the system.

Conversely, SAGA algorithms are correctly not listed if SAGA is not installed.

**QGIS and OS versions**

QGIS 3.15.0 e91abdc23f (qgis-dev 3.15.0-18) on Windows 10.

**Additional context**

PR https://github.com/qgis/QGIS/pull/34617 New standalone console tool for running processing algorithms

agiudiceandrea:
## Description

This patch prevents the `qgis_process list` command from listing the GRASS algorithms among the available processing algs when GRASS is not installed in the system.

See also the same logic used for SAGA algorithms
https://github.com/qgis/QGIS/blob/d0dda27afa2cda4c7a21e1ed89adbb427788b7a2/python/plugins/processing/algs/saga/SagaAlgorithmProvider.py#L95-L100

and a similar behavior for OTB algorithms
https://github.com/qgis/QGIS/blob/9fedd2c581e23b0eaac024a9b62530d2428f2f16/python/plugins/processing/algs/otb/OtbAlgorithmProvider.py#L128-L130

Ref: https://github.com/qgis/QGIS/pull/34617

Fixes https://github.com/qgis/QGIS/issues/37877

### DIFF
--- a/python/plugins/processing/algs/grass7/Grass7AlgorithmProvider.py
+++ b/python/plugins/processing/algs/grass7/Grass7AlgorithmProvider.py
@@ -128,6 +128,12 @@ class Grass7AlgorithmProvider(QgsProcessingProvider):
         return algs
 
     def loadAlgorithms(self):
+        version = Grass7Utils.installedVersion(True)
+        if version is None:
+            QgsMessageLog.logMessage(self.tr('Problem with GRASS installation: GRASS was not found or is not correctly installed'),
+                                     self.tr('Processing'), Qgis.Critical)
+            return
+
         self.algs = self.createAlgsList()
         for a in self.algs:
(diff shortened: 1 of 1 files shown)


nyalldawson:
LGTM -- @alexbruy what do you think?

alexbruy:
I'm fine with this approach, although my personal preference is to show all algorithms from all enabled providers and leave backends installation/setup as user responsibility.

PR #37371, opened 2020-06-24
Addition of Helicopter svg marker

nikc22:
The organisation I work for operates helicopters and I've built this svg file to allow me to use this as a marker. I've named the file in the convention of other files and placed into the appropriate folder, while testing to ensure the marker can be colourised in the same way as other svg markers.

I'm adding this as others might benefit from the work I've done.

## Description

As this is not a code change and the addition of a single svg file, I've not run the pre-commit scripts.

PR #37887, opened 2020-07-21
Addition of Helicopter svg marker

nyalldawson:
Supersedes https://github.com/qgis/QGIS/pull/37371


nikc22:
Thanks for resolving this, I didn't quite understand how I needed to edit the view box.

I'll take a look at what I need to do to get these right for the other markers I'm thinking of adding.

PR #37895, opened 2020-07-21
[vector file writer] LIBKML driver also requires coordinates to be in wgs84

nirvn:
## Description

While we do not use the LIBKML driver in our save vector as file dialog, calling QgsVectorFileWriter::driverForExtension( "kml" ) will return LIBKML. We need to make sure this driver also forces coordinates to be in wgs84 to produce a valid output.

@rouault , out of curiosity, in 2020, would you recommend that QGIS relies on the KML driver or the LIBKML driver to write kml files?

### DIFF
--- a/src/core/qgsvectorfilewriter.cpp
+++ b/src/core/qgsvectorfilewriter.cpp
@@ -419,7 +419,7 @@ void QgsVectorFileWriter::init( QString vectorFileName,
   }
 
   // consider spatial reference system of the layer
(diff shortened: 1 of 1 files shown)


nyalldawson (approved):

rouault:
> would you recommend that QGIS relies on the KML driver or the LIBKML driver to write kml files?

For basic / "normal" needs, both should be OK. The LIBKML one has more options to use "all" KML features, but they need to be explicitly asked for.

PR #37919, opened 2020-07-22
[Processing] Fix GDAL Assign Projection: update QgsRasterLayer crs

rldhont:
## Description

Even if the projection is assign, the QgsRasterLayer and the QgsRasterDataProvider was not updated.

The fix reloads the QgsRasterDataProvider's data and updates the QgsRasterLayer's crs.

### DIFF
--- a/python/plugins/processing/algs/gdal/AssignProjection.py
+++ b/python/plugins/processing/algs/gdal/AssignProjection.py
@@ -28,7 +28,8 @@ from qgis.PyQt.QtGui import QIcon
 from qgis.core import (QgsProcessingException,
                        QgsProcessingParameterRasterLayer,
                        QgsProcessingParameterCrs,
-                       QgsProcessingOutputRasterLayer)
+                       QgsProcessingOutputRasterLayer,
+                       QgsProcessingContext)
 from processing.algs.gdal.GdalAlgorithm import GdalAlgorithm
 from processing.algs.gdal.GdalUtils import GdalUtils
 
@@ -97,3 +98,37 @@ class AssignProjection(GdalAlgorithm):
         self.setOutputValue(self.OUTPUT, fileName)
 
         return commands
+
+    def postProcessAlgorithm(self, context, feedback):
+        # get output value
+        fileName = self.output_values.get(self.OUTPUT)
+        if not fileName:
+            return {}
+
+        # search in context project's layers
+        if context.project():
+
+            for l in context.project().mapLayers().values():
+
+                # check the source
+                if l.source() != fileName:
+                    continue
+
+                # reload provider's data
+                l.dataProvider().reloadData()
+                l.setCrs(l.dataProvider().crs())
+                l.triggerRepaint()
+
+        # search in context temporary layer store
+        for l in context.temporaryLayerStore().mapLayers().values():
+
+            # check the source
+            if l.source() != fileName:
+                continue
+
+            # reload provider's data
+            l.dataProvider().reloadData()
+            l.setCrs(l.dataProvider().crs())
+            context.temporaryLayerStore().addMapLayer(l)
+
+        return {}
--- a/python/plugins/processing/tests/GdalAlgorithmsRasterTest.py
+++ b/python/plugins/processing/tests/GdalAlgorithmsRasterTest.py
@@ -28,7 +28,9 @@ import tempfile
 
 from qgis.core import (QgsProcessingContext,
                        QgsProcessingFeedback,
-                       QgsRectangle)
+                       QgsRectangle,
+                       QgsRasterLayer,
+                       QgsProject)
 
 from qgis.testing import (start_app,
                           unittest)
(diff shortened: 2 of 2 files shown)


python/plugins/processing/algs/gdal/AssignProjection.py
```diff
@@ -28,7 +28,9 @@
 from qgis.core import (QgsProcessingException,
                        QgsProcessingParameterRasterLayer,
                        QgsProcessingParameterCrs,
-                       QgsProcessingOutputRasterLayer)
+                       QgsProcessingOutputRasterLayer
```
Gustry, line 31:
```suggestion
                       QgsProcessingOutputRasterLayer,
```

alexbruy (approved):
Looks good to me.

python/plugins/processing/algs/gdal/AssignProjection.py
```diff
@@ -94,6 +112,24 @@ def getConsoleCommands(self, parameters, context, feedback, executing=True):
 
         commands.append(GdalUtils.escapeAndJoin(arguments))
 
+        # if the input layer is a project's layer, the algorithm can
+        # update the QgsMapLayer CRS
+        if context.project() and context.project().mapLayer(inLayer.id()):
```
nyalldawson, line 117: This isn't safe to call from a background thread -- you'd need to make the algorithm run in the main thread to be able to access QgsProject.mapLayer()

rldhont: I don't want to run the algorithm in the main thread. I would like to be sure that the input which is the output will be updated at the end of the algorithm.

* How to know that the input layer is already loaded in the project context ?
* If the algorithm is run in an other thread than the main one, is the `context.project()` not null even if it is not safe to access `mapLayer()` ? 
* How to reload or update the input QgsRasterLayer, that is also the output, at the end of the algorithm ? I try to solve https://github.com/qgis/QGIS/issues/37920

nyalldawson: I'd suggest overriding the postProcessAlgorithm method (which always runs in the main thread) and do the update there.

rldhont: thanks Nyall, I will try

nyalldawson: Sorry I missed off the rest of my advice! I was going to suggest that the post processor runs through all the layers in the project, and updates all of them where the layers source matches the changed layer's source. Then it'll nicely handle the case that the layer is loaded multiple times in the prpject

rldhont: Thanks @nyalldawson, I have overrided postProcessAlgorithm method and I was able to get the QgsRasterLayer's crs updated in the project and the processing context.

rldhont:
Fixed #37920

python/plugins/processing/algs/gdal/AssignProjection.py
```diff
@@ -28,7 +28,9 @@
 from qgis.core import (QgsProcessingException,
                        QgsProcessingParameterRasterLayer,
                        QgsProcessingParameterCrs,
-                       QgsProcessingOutputRasterLayer)
+                       QgsProcessingOutputRasterLayer,
+                       QgsProcessingContext,
```
rldhont, line 32:
```suggestion
                       QgsProcessingContext,
                       QgsProcessingUtils,
```

python/plugins/processing/algs/gdal/AssignProjection.py
```diff
@@ -97,3 +115,40 @@ def getConsoleCommands(self, parameters, context, feedback, executing=True):
         self.setOutputValue(self.OUTPUT, fileName)
 
         return commands
+
+    def postProcessAlgorithm(self, context, feedback):
+        # get output value
+        fileName = self.outputValue(self.OUTPUT)
```
nyalldawson, line 121: I'd approach this differently:
1. scan through the map layers in the project, and if any match the layer source then update that
2. do the same for the context layers

That will avoid the need for SetCrsPostProcessor and should be more robust

rldhont: Thnaks @nyalldawson, I have removed SetCrsPostProcessor

python/plugins/processing/algs/gdal/AssignProjection.py
```diff
@@ -97,3 +99,29 @@ def getConsoleCommands(self, parameters, context, feedback, executing=True):
...
+        if not fileName:
+            return {}
+
+        # get layer from output
+        layer = QgsProcessingUtils.mapLayerFromString(fileName, context)
```
nyalldawson, line 110: You shouldn't need this - rather loop through ALL layers in the project/context, and update all matching ones you find

python/plugins/processing/algs/gdal/GdalAlgorithm.py
```diff
@@ -130,6 +130,11 @@ def getOgrCompatibleSource(self, parameter_name, parameters, context, feedback,
             ogr_layer_name = GdalUtils.ogrLayerName(input_layer.dataProvider().dataSourceUri())
         return ogr_data_path, ogr_layer_name
```
nyalldawson, line 132: You can avoid this by storing the layer's path as a member in the algorithm during the processAlgorithm step, and then retrieving it in the postProcess step

python/plugins/processing/algs/gdal/AssignProjection.py
```diff
@@ -97,3 +98,42 @@ def getConsoleCommands(self, parameters, context, feedback, executing=True):
...
+                if l.source() != fileName:
+                    continue
+
+                # reload provider's data
+                l.dataProvider().reloadData()
+                l.setCrs(l.dataProvider().crs())
+                l.triggerRepaint()
+
+                # the input layer which is the output one has been found
+                return {}
```
nyalldawson, line 122: The same raster file may be present multiple times in a project or context, so we should continue the loop here and not break out early

nyalldawson (commented):
Getting close! Can you add a unit test too please?

nyalldawson (commented):
Lgtm! Can you add a unit test too please?

nyalldawson (approved):

rldhont:
The test does not work because gdal_edit.py is not installed
```
Processing(0): GDAL execution console output
/bin/sh: 1: gdal_edit.py: not found
```

PR #37921, opened 2020-07-22
Fix default value expression on double field

troopa81:
## Description

When a default value expression is defined on provider side (Postgres or Oracle  for instance) for a double field, the value always default to 0 and the expression is lost when you create a feature 

For instance, on Postgres
```sql
create table test_double( id serial primary key, mydouble double precision DEFAULT 6+3, geom geometry(Point,4326)
```
And Oracle
```sql
CREATE TABLE QGIS.POINT_DATA_IDENTITY ( "pk" NUMBER GENERATED ALWAYS AS IDENTITY PRIMARY KEY, GEOM SDO_GEOMETRY)
```
Both *pk* and *mydouble* always defaults to 0

### DIFF
--- a/src/core/qgsfield.cpp
+++ b/src/core/qgsfield.cpp
@@ -252,6 +252,13 @@ QString QgsField::displayString( const QVariant &v ) const
   // Special treatment for numeric types if group separator is set or decimalPoint is not a dot
   if ( d->type == QVariant::Double )
   {
+    // if value doesn't contain a double (a default value expression for instance),
+    // apply no transformation
+    bool ok;
+    v.toDouble( &ok );
+    if ( !ok )
+      return v.toString();
+
     // Locales with decimal point != '.' or that require group separator: use QLocale
     if ( QLocale().decimalPoint() != '.' ||
          !( QLocale().numberOptions() & QLocale::NumberOption::OmitGroupSeparator ) )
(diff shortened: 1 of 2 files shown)


troopa81:
cc @Jean-Roc maybe you already meet this issue on number identity field

m-kuhn:
For reference, this is yet another band aid for the current default value implementation (see https://github.com/qgis/QGIS/issues/36396#issuecomment-639382894)

troopa81:
> For reference, this is yet another band aid for the current default value implementation (see #36396 (comment))

I'm OK that it will be way better to have a dedicated default value QVariant, but in the mean time, do you agree with this fix ? Anyway, it doesn't hurt to check if a value is truly a double before going on formatting it, no?

m-kuhn (approved):
IMHO representation of a double which is not a double is undefined behavior. In general I prefer UB to fail since it might highlight implementation problems.

Randomly quoting a python snippet here to highlight the point:

```
>>> float('(1+1)')
  File "<stdin>", line 1, in <module>
ValueError: could not convert string to float: '(1+1)'

```

However, this bandaids a nasty issue, so let's go for it. And envision an improved solution in the longer run.

troopa81:
> In general I prefer UB to fail since it might highlight implementation problems.

I agree, but for now _displayString_ method is permissive (e.g. other numeric conversion fails, it falls back to a classic toString call)

PR #37952, opened 2020-07-24
Manage 3D symbols through style manager

### LINKED ISSUE #31479: Expose 3D symbols to style manager
Something is missing here:

image

That's right!... QGIS has no ability to manage and handle 3d symbols via style manager!

nyalldawson:
This PR adds support for managing 3d symbols through the QGIS style manager, including support for importing and exporting libraries of 3d symbols, and completes the recent work in fully abstracting out 3d symbols and their widget handlers.

Peek 2020-07-24 10-20

One notable omission is that there's no thumbnails generated for 3d symbols in the library, and everything gets a generic icon for now. Generating thumbnails is no trivial matter, so it's deffered for now...

Fixes #31479


nirvn:
@nyalldawson , fantastic work!

What about displaying some basic information - such as colors - with the preview icon. We could still use the 3D symbol, but tweak it to reflect part of the saved symbol.

E.g.:
Screenshot from 2020-07-24 12-11-40

nyalldawson:
@nirvn it's an option I guess - still not a trivial inclusion, so I'd rather explore taking a snapshot of the real symbol  first

timlinux:
Nice!

PR #37959, opened 2020-07-25
Add a registry for 3d materials

nyalldawson:
Adds a new QgsMaterialRegistry class for managing 3d material types, and reworks the current phong material widgets and symbol serialization to make it generic. This opens the door for future addition of new 3d materials other than the current phong material (e.g. CAD style gooch materials, physically based metallic materials, unlit materials, etc)


nyalldawson:
Note that there's one missing component here, which is filtering the selection of available registered materials for a given symbol type to just those which are compatible with a particular symbol (refs https://github.com/qgis/QGIS/pull/37776#pullrequestreview-450226481)

Saijin-Naib:
Are materials with normal/specular in addition to diffuse possible with these changes?

nyalldawson:
@saijin-naib

Eventually, yes. This is a step toward that.

wonder-sk (approved):
Great stuff

PR #37961, opened 2020-07-26
Use correct case for the first letter of words in GUI tooltips - qgisapp.ui

havatv:
While documenting, I found some inconsistencies in the first letter case for words in GUI tooltips.
I have tried to use the "correct" case for the words in the tooltips ("important" words have an uppercase first letter, and "insignificant" words are lowercase).

## Description

Use correct case for the first letter of words in GUI tooltips - qgisapp.ui.
Should/could be backported to 3.10.

### DIFF
--- a/src/ui/qgisapp.ui
+++ b/src/ui/qgisapp.ui
@@ -1175,7 +1175,7 @@
     <string>Select Feature(s)</string>
    </property>
    <property name="toolTip">
-    <string>Select Features by area or single click</string>
+    <string>Select Features by Area or Single Click</string>
    </property>
   </action>
   <action name="mActionSelectPolygon">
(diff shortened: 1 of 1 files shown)


src/ui/qgisapp.ui
```diff
@@ -2768,7 +2768,7 @@ Acts on the currently active layer only.</string>
     <string>Add Circular String by Radius</string>
    </property>
    <property name="toolTip">
-    <string>Add circular string by radius</string>
+    <string>Add Circular String by radius</string>
```
nyalldawson, line 2771: Radius?

nyalldawson (approved):
Great move! The tooltip inconsistencies have annoyed me too!

src/ui/qgisapp.ui
havatv, line 2771:
```suggestion
    <string>Add Circular String by Radius</string>
```

PR #37972, opened 2020-07-27
Fix loading batch settings with date/time values

### LINKED ISSUE #37838: Cannot import saved batch process of Model
**Describe the bug**
I have created a model using Graphical Modeler in QGIS 3.14. The works fine as a single process.
I then created a Batch Process. The batch process also works fine.
Then I saved the Batch Process.
Now when open the saved Batch Process JSON file, it does not load and gives the following error 
  'An error has occurred while executing Python code: 

NameError: name 'QDateTime' is not defined 
  File "C:/PROGRA~1/QGIS3~1.14/apps/qgis/./python/plugins\processing\gui\BatchPanel.py", line 543, in load
    value = eval(params[param.name()])
  File "", line 1, in 
NameError: name 'QDateTime' is not defined

Python version: 3.7.0 (v3.7.0:1bf9cc5093, Jun 27 2018, 04:59:51) [MSC v.1914 64 bit (AMD64)] 
QGIS version: 3.14.0-Pi Pi, 9f7028fd23 '

Second issue is that one of the input fields does not show up in the columns when we are creating the Batch process.

The saved batch file opens in QGIS 3.10, but the 'Datetime' input of the Graphical Modeler shows an error and does not work. I guess there is not 'Datetime' input in 3.10

**How to Reproduce**

1. Go to Graphical Modeler
2. Create a model
3. Run the model
4. Select 'Run as Batch Process'
5. Create the batch Process and save it.
6. Close the Batch Process and open it again
7. Load the same saved batch Process
8. See the error --> NameError: name 'QDateTime' is not defined

**QGIS and OS versions**

QGIS version
3.14.0-Pi
QGIS code revision
9f7028fd23
Compiled against Qt
5.11.2
Running against Qt
5.11.2
Compiled against GDAL/OGR
3.0.4
Running against GDAL/OGR
3.0.4
Compiled against GEOS
3.8.1-CAPI-1.13.3
Running against GEOS
3.8.1-CAPI-1.13.3
Compiled against SQLite
3.29.0
Running against SQLite
3.29.0
PostgreSQL Client Version
11.5
SpatiaLite Version
4.3.0
QWT Version
6.1.3
QScintilla2 Version
2.10.8
Compiled against PROJ
6.3.2
Running against PROJ
Rel. 6.3.2, May 1st, 2020
OS Version
Windows 10 (10.0)
Active python plugins
db_manager; 
MetaSearch; 
processing
QDateTime Error

**Additional context**

nyalldawson:
Fixes #37838

### DIFF
--- a/python/plugins/processing/gui/BatchPanel.py
+++ b/python/plugins/processing/gui/BatchPanel.py
@@ -37,7 +37,19 @@ from qgis.PyQt.QtWidgets import (
     QMenu,
     QAction
 )
-from qgis.PyQt.QtGui import QPalette
+
+# adding to this list? also update the HistoryDialog.py executeAlgorithm imports!!
+
+from qgis.PyQt.QtCore import (
+    QTime,  # NOQA - must be here for saved file evaluation
+    QDate,  # NOQA - must be here for saved file evaluation
+    QDateTime  # NOQA - must be here for saved file evaluation
+)
+from qgis.PyQt.QtGui import (
+    QPalette,
+    QColor,  # NOQA - must be here for saved file evaluation
+)
+
 from qgis.PyQt.QtCore import (
     QDir,
     QFileInfo,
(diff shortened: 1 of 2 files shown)


nirvn:
@nyalldawson , you might as well add Q{DateTime,Date,Time} imports to the history dialog. We don't have any temporal parameter yet but when we do, we are likely to forget that part :smile:

nyalldawson:
> We don't have any temporal parameter yet but when we do

Actually we do!

nirvn:
Oh, then, the suggestion becomes a requirement! :wink:

nyalldawson:
@nirvn done!

PR #37974, opened 2020-07-27
Add mechanism for algorithms to alter their core behavior when run in-place mode

nyalldawson:
This allows algorithms to dynamically adapt their behavior to make them compatible
with in-place mode. Previously, some useful algorithms could not be
run in-place because they alter a layer's structure (e.g. adding new
fields).

Now, these algorithms have a means to detect that they are being
run in-place and change their input parameters accordingly. E.g.
an algorithm which usually adds new fields to store calculated
values (such as "add xy fields to layer") could instead expose
field parameter choices to ask the user to pick from existing
fields in which to store the calculated values, thereby avoiding
the need to change the table structure and making them eligable
for running in-place mode.

Note that this needs to be handled algorithm-by-algorithm, it's
not automatic! It's just the raw api to allow this... Here I've just updated the newly added "Align points to features" algorithm to add support for in place mode, as a bit of an demonstration of how an algorithm can be adapted.

Peek 2020-07-27 13-46

### DIFF
--- a/src/analysis/processing/qgsalgorithmangletonearest.cpp
+++ b/src/analysis/processing/qgsalgorithmangletonearest.cpp
@@ -102,8 +102,10 @@ QString QgsAngleToNearestAlgorithm::groupId() const
 
 QgsAngleToNearestAlgorithm::~QgsAngleToNearestAlgorithm() = default;
 
-void QgsAngleToNearestAlgorithm::initAlgorithm( const QVariantMap & )
+void QgsAngleToNearestAlgorithm::initAlgorithm( const QVariantMap &configuration )
 {
+  mIsInPlace = configuration.value( QStringLiteral( "IN_PLACE" ) ).toBool();
+
   addParameter( new QgsProcessingParameterFeatureSource( QStringLiteral( "INPUT" ),
                 QObject::tr( "Input layer" ), QList< int >() << QgsProcessing::TypeVectorPoint ) );
   addParameter( new QgsProcessingParameterFeatureSource( QStringLiteral( "REFERENCE_LAYER" ),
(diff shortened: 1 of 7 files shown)


src/analysis/processing/qgsalgorithmangletonearest.cpp
```diff
@@ -112,13 +114,23 @@ void QgsAngleToNearestAlgorithm::initAlgorithm( const QVariantMap & )
   addParameter( new QgsProcessingParameterDistance( QStringLiteral( "MAX_DISTANCE" ),
                 QObject::tr( "Maximum distance to consider" ), QVariant(), QStringLiteral( "INPUT" ), true, 0 ) );
 
-  addParameter( new QgsProcessingParameterString( QStringLiteral( "FIELD_NAME" ), QObject::tr( "Angle field name" ), QStringLiteral( "rotation" ) ) );
+  if ( !mIsInPlace )
+    addParameter( new QgsProcessingParameterString( QStringLiteral( "FIELD_NAME" ), QObject::tr( "Angle field name" ), QStringLiteral( "rotation" ) ) );
+  else
+    addParameter( new QgsProcessingParameterField( QStringLiteral( "FIELD_NAME" ), QObject::tr( "Angle field name" ), QStringLiteral( "rotation" ), QStringLiteral( "INPUT" ) ) );
```
m-kuhn, line 120: Would it make sense to also allow updating an existing field when *not* run in place? (Or add a field when run in place?)

Or am I missing something?

m-kuhn:
Great new API. Just wondering if it often makes sense to modify parameters or more often to modify the processing of the alg itself.

alexbruy (approved):
Nice feature

nyalldawson:
@m-kuhn 
> Great new API. Just wondering if it often makes sense to modify parameters or more often to modify the processing of the alg itself.

I think the current approach with adding new fields by default is correct. Otherwise, if you're wanting to create a new layer from an algorithm you'd be forced to do a lot of annoying steps:
1. copy the layer
2. manually create the fields for the new attributes to be stored in
3. run the algorithm on the modified layer, picking the fields you just created, and get another new layer output

vs the current approach:
1. just run the alg :wink:

So I'd like us to continue this convention even with this new api:
- running not "in place": algorithms should always create new fields to store newly created attributes. Optionally, they can expose a choice of entering this new field name for users
- running "in place": expose field parameters to let users pick from existing fields to store the attribute. Where it makes sense, make these optional (e.g. if calculation of the fields is just a nice "Added extra", and not the core purpose of the algorithm)

I'll add an example of a simpler algorithm adaptation to this PR -- the "align markers" alg is complicated somewhat because it modifies the layer's renderer.

m-kuhn:
I think you misread me. I agree the "add fields" by default is useful and "update fields" for in-place often makes sense.

See https://github.com/qgis/QGIS/pull/37974#discussion_r460713421

> Would it make sense to **also** allow updating an existing field when not run in place? (Or add a field when run in place?)

PR #37990, opened 2020-07-28
Add API for clipping layout maps based on a shape/polygon item

nyalldawson:
Allows using shape or polygon layout items to set the overall shape of a layout map item's contents, opening the door for non-rectangular maps in layouts!

API only for now -- can't be set via the UI.

But here's a sample ugly screenshot:

image


src/core/layout/qgslayoutitemmap.cpp
```diff
@@ -836,6 +859,17 @@ bool QgsLayoutItemMap::readPropertiesFromElement( const QDomElement &itemElem, c
   return true;
 }
 
+QPainterPath QgsLayoutItemMap::framePath() const
+{
+  if ( mItemClippingSettings->isActive() )
+  {
+    const QgsGeometry g = mItemClippingSettings->clipPathInMapItemCoordinates();
+    if ( !g.isNull() )
+      return g.constGet()->asQPainterPath();
```
nirvn, line 868: That's a really nice implementation detail.

nirvn (approved):
B-e-a-u-t-i-f-u-l.

Would it ever be useful to allow for an inverted clip, i.e. a difference operation? I can't think of any scenario where it'd be useful, just thinking out-loud.

nyalldawson:
> Would it ever be useful to allow for an inverted clip, i.e. a difference operation? I can't think of any scenario where it'd be useful, just thinking out-loud.

I can't think of when we'd need that either :stuck_out_tongue:

PR #37994, opened 2020-07-28
[locator] Add user locale handling to the goto locator

nirvn:
## Description

This allows users to type in coordinate using their locale's decimal separator. E.g., for French Canadian, the following lon lat string will now work: `104,1234 11,1234`.

Locale awareness also means coordinates with a comma as thousand separator will work too. For e.g., this UTM coordinate would work just fine: `456,789.1 1,234,567.2`.


m-kuhn:
Happy to merge if it works for you. Refraining from a review since I found the only reliable way to review regexes is an adequate set of unit tests.

uclaros:
> Happy to merge if it works for you. Refraining from a review since I found the only reliable way to review regexes is an adequate set of unit tests.

Is there an easy way to unit test this with different locales?

PR #38002, opened 2020-07-28
WFS server: use exact intersection if there is a bbox parameter

mhugent:
If there is a bbox-Parameter in a WFS GetFeature request, the ExactIntersect - Flag has not  been set to the feature request object. Therefore it sometimes returned more features than it should.


rldhont (approved):

Gustry:
@mhugent Maybe the backport bot didn't work because your fork is named `Quantum-GIS`. You should go in your repository settings and rename it. It might fix the issue.

PR #38020, opened 2020-07-28
Don't try to update raster min/max settings when a raster layer is set to Updated Canvas mode and is being rendered in a 3d map (3.10)

nyalldawson:
Backport #38001


nirvn (approved):
Nice and clean.

PR #38021, opened 2020-07-29
Add API for "Annotation layers"

nyalldawson:
This PR adds the underlying API for a new QgsMapLayer subclass: "QgsAnnotationLayer".

An annotation layer is designed for freeform drawing of georeferenced objects, which don't have any of the constraints of adding objects to a standard vector layer. They don't have fields, and can contain many different types of drawing objects in a single layer (e.g. markers, lines, text objects, etc). The annotation items are all spatially referenced, so are attached to particular geographic geometries (potentially allowing for them to be converted to real vector layers in future).

Use cases for annotation layers include:
- exposing very simple and quick to use drawing tools for users who are not "GIS savvy", yet who have a need to digitize/draw geographically referenced markup
- allowing easy to use tools for cartographically touching up a map, e.g. adding specific text labels to a map which aren't associated with any particular layer or feature, and which can be moved and manipulated in a freeform way
- exposing drawing tools which don't fit into the "simple features specification" for vector layers, e.g. a curved text object which follows a bezier curve

An annotation layer consists of a set of items which are subclasses of QgsAnnotationItem. Included in this PR are three annotation item types:
- QgsAnnotationMarkerItem, for placement of a marker symbol on a map
- QgsAnnotationLineStringItem, for drawing a linestring which is styled by a line symbol
- QgsAnnotationPolygonItem, for drawing polygon regions styled by a fill symbol

(future PRs will add various text item subclasses too)

This work is loosely based off the redlining tools support from Sourcepole's Kadas fork, but with many changes:
- rendering has been made thread safe
- item classes are managed by a central registry, allowing for easy addition of new item types and for plugin based annotation item types
- the included marker/line/polygon item types all rely on QgsSymbol classes for rendering, and each individual item of these types in an annotation layer can have its own unique appearance using the full range of QGIS symbology which is available for point/line/polygon items (including draw effects)

No GUI is exposed for working with annotation layers -- funding is sought for adding this (I'll be launching a crowdfund for this shortly). For now, it's the underlying API only.

Sponsored by the QGIS Swiss user group.


Saijin-Naib:
😭🙇‍♂️🙏

This is so important!

Thanks Nyall & SUG!

nyalldawson:
@m-kuhn @nirvn any chance of a review?

src/core/annotations/qgsannotationitem.h
```diff
@@ -0,0 +1,155 @@
...
+ * Abstract base class for annotation items which are drawn with QgsAnnotationLayers.
+ *
+ * \since QGIS 3.16
+ */
+class CORE_EXPORT QgsAnnotationItem
+{
+
+#ifdef SIP_RUN
+    SIP_CONVERT_TO_SUBCLASS_CODE
+    if ( sipCpp->type() == "marker" )
```
m-kuhn, line 43: We don't have anything more performant than strings here?

```suggestion
    if ( sipCpp->type() == QLatin1String( "marker" ) )
```

src/core/annotations/qgsannotationlayer.h
```diff
@@ -0,0 +1,126 @@
...
+    QgsMapLayerRenderer *createMapRenderer( QgsRenderContext &rendererContext ) override SIP_FACTORY;
+    QgsRectangle extent() const override;
+    void setTransformContext( const QgsCoordinateTransformContext &context ) override;
+    bool readXml( const QDomNode &layerNode, QgsReadWriteContext &context ) override;
+    bool writeXml( QDomNode &layer_node, QDomDocument &doc, const QgsReadWriteContext &context ) const override;
+    bool writeSymbology( QDomNode &node, QDomDocument &doc, QString &errorMessage, const QgsReadWriteContext &, StyleCategories categories = AllStyleCategories ) const override;
+    bool readSymbology( const QDomNode &node, QString &errorMessage, QgsReadWriteContext &context, StyleCategories categories = AllStyleCategories ) override;
+
+  private:
+    QMap<QString, QgsAnnotationItem *> mItems;
```
m-kuhn, line 121: What do we gain by giving the item ids such an importance (lookup table here, remove by id)... compared to using pointers as main reference?

src/core/annotations/qgsannotationitem.h
```diff
@@ -0,0 +1,155 @@
...
+     * \see setCrs()
+     */
+    QgsCoordinateReferenceSystem crs() const { return mCrs; }
+
+    /**
+     * Sets the \a crs used for storing the location of the item.
+     *
+     * \see crs()
+     */
+    void setCrs( const QgsCoordinateReferenceSystem &crs );
```
m-kuhn, line 108: What's the reason for items having individual crs (rather than all being in layer crs)?

m-kuhn (commented):
Very cool stuff !!!!!

src/core/annotations/qgsannotationitemregistry.cpp
```diff
@@ -0,0 +1,79 @@
+/***************************************************************************
+                            qgsannotationitemregistry.cpp
+                            -------------------------
+    begin                : October 2019
```
nirvn, line 4: I assume this beautiful code was sitting on a local branch since October 2019? :smile:

src/core/annotations/qgsannotationlinestringitem.h
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * Returns the symbol used to render the item.
+     *
+     * \see setSymbol()
+     */
+    const QgsLineSymbol *symbol() const;
+
+    /**
+     * Sets the \a symbol used to render the marker item.
```
nirvn, line 82:
```suggestion
     * Sets the \a symbol used to render the line item.
```

src/core/annotations/qgsannotationpolygonitem.h
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Returns the symbol used to render the item.
+     *
+     * \see setSymbol()
+     */
+    const QgsFillSymbol *symbol() const;
+
+    /**
+     * Sets the \a symbol used to render the marker item.
```
nirvn, line 81:
```suggestion
     * Sets the \a symbol used to render the polygon item.
```

src/core/qgsapplication.h
```diff
@@ -682,6 +683,12 @@ class CORE_EXPORT QgsApplication : public QApplication
      */
     static QgsLayoutItemRegistry *layoutItemRegistry() SIP_KEEPREFERENCE;
 
+    /**
+     * Returns the application's annotation item registry, used for annotation item types.
+     * \since QGIS 3.12
```
nirvn, line 688:
```suggestion
     * \since QGIS 3.16
```

nirvn (approved):
Great stuff.

src/core/annotations/qgsannotationitem.h
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * Renders the item to the specified render \a context.
+     *
+     * The \a feedback argument can be used to detect render cancelations during expensive
+     * render operations.
+     */
+    virtual void render( QgsRenderContext &context, QgsFeedback *feedback ) = 0;
```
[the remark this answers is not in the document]
nyalldawson: Yes -- all the symbology and text rendering code use modifiable references to the context, because they are modifying the expression context, painter, and other properties of the context.

src/core/annotations/qgsannotationitemregistry.h
```diff
@@ -0,0 +1,212 @@
...
+    /**
+     * Returns a translated, user visible name for the annotation item class.
+     * \see visiblePluralName()
+     */
+    QString visibleName() const { return mVisibleName; }
+
+    /**
+     * Returns a translated, user visible name for plurals of the annotation item class (e.g. "Labels" for a "Label" item).
+     */
+    QString visiblePluralName() const { return mVisibleNamePlural; }
```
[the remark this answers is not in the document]
nyalldawson: That looks neat, but I can't see anyway to use it for the English builds...! Anyway, I'm following the pattern currently used elsewhere (e.g. QgsLayoutItemRegistry)

src/core/annotations/qgsannotationlayer.h
nyalldawson: I was aiming for a persistent ID which could be used between project reloads/layer clones/etc, and for the extra safety that storing a string uuid gives vs storing a raw pointer to an annotation (e.g. in the drawing UI classes)

src/core/annotations/qgsannotationitem.h
nyalldawson: Good question! I had my slyr hat on here, and annotations in ESRI world are stored using a referenced geometry + crs pair. So approaching it like this makes it easier to translate between the two. I don't see any problem in particular with this approach, expect a bit of extra code complexity. But I'm not set on it if you think it's a mistake...

src/core/annotations/qgsannotationitemregistry.cpp
nyalldawson: correct!

src/core/annotations/qgsannotationitem.h
nyalldawson: Again, I'm following the pattern set by symbol layers. I suppose we could use dynamic casts as the tests... but with sip I'm inclined to leave what works :laughing:

src/core/annotations/qgsannotationitem.h
m-kuhn: I was wondering what we gain (not sure if ESRI compatibility alone is convincing). I was a bit afraid of complexity: project crs vs layer crs vs item crs seems a lot of reprojection and possible room for missing one.

nyalldawson: Ah - but an annotation layer itself has no CRS. So it's just map CRS vs item CRS.

src/core/annotations/qgsannotationlayer.h
m-kuhn: I think I remember you saying some years ago that you wish we used more layer pointers rather than layer ids. Did that change, is the situation different here or is my memory distorted?

nyalldawson: I can't recall exactly, but in any case now I'd prefer pointers with the layer stored in a QPointer :wink: But that approach only works for QObjects, and I don't really want to make the annotation items qobjects themselves.

m-kuhn: Thanks for the clarification. I think shared_ptr don't play that well with sip ?! No more objections :+1:

src/core/annotations/qgsannotationitem.h
m-kuhn: Hmm, every layer has a `QgsMapLayer::crs()` by design.

nyalldawson: ...But since a while we've supported returning invalid crs objects from this as a reflection of "no crs" 😁

src/core/annotations/qgsannotationlayer.h
nyalldawson: I tried a while ago to make sip handle unique ptrs with absolutely no luck.. unfortunately shared ptrs would be even harder. Maybe qt for python will help fix this in the long run...

src/core/annotations/qgsannotationitem.h
m-kuhn: How are you going to work with this UI wise? Add it with the current project CRS? Will the CRS change if an item is moved?

nyalldawson: Ok, I'm convinced. I'll rework this!

nyalldawson:
Unrelated test failure

PR #38022, opened 2020-07-29
Translate instructions when help file is not found

DelazJ:


m-kuhn:
Does that translate?

DelazJ:
"Allow to translate"? If it's the wording, it can still be fixed at merge time (if any).
If you meant whether it works, TBH no idea. I follow what is done in both doc and website repo to have HTML contents available in Transifex. Eg https://github.com/qgis/QGIS-Website/pulls?q=is%3Apr+is%3Aclosed+localizable

m-kuhn:
> If you meant whether it works, TBH no idea. I follow what is done in both doc and website repo to have HTML contents available in Transifex. Eg https://github.com/qgis/QGIS-Website/pulls?q=is%3Apr+is%3Aclosed+localizable

The website and the doc repos have html template engines in place (might be sphinx that handles this?), this file here is plain html, not processed by anything.
It could be dynamically generated by code either on runtime or as a CMake step, but I don't think this change here will help, sorry.

DelazJ:
Don't be sorry. I've tried, and I learnt. Thanks

PR #38026, opened 2020-07-29
[Oracle] Add support for virtual (generated) column

troopa81:
## Description

This PR adds support for virtual Oracle columns defined this way
```sql
CREATE TABLE QGIS.GENERATED_COLUMNS ( "pk" INTEGER PRIMARY KEY, "generated_field" GENERATED ALWAYS AS ('test:' || "pk") VIRTUAL);
```
**Note:** It's not possible to define a generated column on an SDO_GEOMETRY field in Oracle.

For now, when creating a new feature, generated field is editable and should not, we need to add a new API method to retrieve readonly field from provider as discussed [here](https://github.com/qgis/QGIS/pull/35995#issuecomment-633055700). It will come in a different PR (If I find enough time).

I add a test in providerbase to ensure we get the same behavior in postgres and oracle. 

cc @SebastienPeillet


src/providers/oracle/qgsoracleprovider.cpp
```diff
@@ -608,9 +610,16 @@ bool QgsOracleProvider::loadFields()
...
-                 " FROM all_tab_columns t"
-                 " WHERE t.owner=? AND t.table_name=?" ) ;
+                 ",t.data_default" +
+                 QString( mOracleVersion >= 12 ?
+                          ",CASE WHEN t.virtual_column = 'YES' OR a.generation_type = 'ALWAYS' THEN 'YES' ELSE 'NO' END" :
+                          ",t.virtual_column" ) +
+                 " FROM all_tab_cols t" +
+                 QString( mOracleVersion >= 12 ?
+                          " LEFT JOIN all_tab_identity_cols a ON a.column_name = t.column_name AND a.owner = t.owner AND a.table_name = t.table_name" :
+                          "" ) +
```
SebastienPeillet, line 620: Identity column needs 12th version of Oracle but virtual column are available since the 11th version, isn't it ? I don't know if it's will concern many users.

src/providers/oracle/qgsoracleconn.h
```diff
@@ -179,6 +179,11 @@ class QgsOracleConn : public QObject
 
     bool hasSpatial();
 
+    /**
+     * \returns Oracle database version, -1 in an error occured
```
SebastienPeillet, line 183:
```suggestion
     * \returns Oracle database major version, -1 in an error occured
```

SebastienPeillet (commented):
It seem's good to me ! :+1: 

There is just my question about people using 11th version of Oracle that won't be able to use virtual column

src/providers/oracle/qgsoracleprovider.cpp
troopa81: If *mOracleVersion* is lower than 12, then I still get the *t.virtual_column* (which is already set with "YES" if it's a virtual column), so I'd say it should work with version 11.

troopa81:
Don't stale!

src/providers/oracle/qgsoracleprovider.cpp
SebastienPeillet: Indeed, sorry I misread the line between the condition in SQL inside the condition in C++...

SebastienPeillet (approved):

troopa81:
@rldhont could you review this PR and if ok merge it?

rldhont (approved):

rldhont:
LGTM!

troopa81:
@rldhont do you mind merging it?

rldhont:
@troopa81 spell check error https://github.com/qgis/QGIS/runs/1061665740?check_suite_focus=true: 
```
src/providers/oracle/qgsoracleconn.h
183:     * 
eturns Oracle database major version, -1 if an error occured
```
Can you propose a fix ?

troopa81:
> Can you propose a fix ?

Done. I wonder how it could happen since the CI was all green if I remember correctly.

PR #38030, opened 2020-07-29
WFS server: use exact instersection if there is a bbox parameter

mhugent:
Backport to 3.10


nyalldawson (approved):

nyalldawson:
Moved to https://github.com/qgis/QGIS/pull/37904

PR #38035, opened 2020-07-29
Use geometry for argument in functions instead of geom

DelazJ:
There's currently a mix of geom and geometry to name the geometry argument. Harmonizing! But I don't know if it's safe (eg if someone was using the named argument `geom:=` in an older expression)
Also tweak some descriptions


nyalldawson:
We can't change argument names without breaking projects -- we'll first need a way to transparently map the old name across (not trivial).

DelazJ:
Thanks for confirmation. I've seen such a change in a recent PR so expected that it would pass ;(
OK I'll then pick the examples part in another PR.

nyalldawson:
Actually- hold off on this. Maybe it's easy to map across..

nyalldawson:
Superseded by https://github.com/qgis/QGIS/pull/38040

PR #38040, opened 2020-07-29
[expressions] Silently alias "geom" named arguments across to "geometry"

nyalldawson:
So that was can standardize on using "geometry" as the named argument
insted of mixing "geom" and "geometry" in the public docs.


nyalldawson:
Supersedes https://github.com/qgis/QGIS/pull/38035

DelazJ:
Thanks, Nyall

PR #38041, opened 2020-07-30
Further improvements for material handling, add Gooch shading model

nyalldawson:
Further improves the QgsAbstractMaterialSettings API, by:
- adding icon support for registered material types (for use in the material selector combo box)
- adds support for handling different rendering techniques to QgsAbstractMaterialSettings. Now material classes can indicate which rendering techniques they support (e.g. Triangles, Lines, InstancedPoints, etc), and the material is only offered when it's compatible with the required technique. In addition, I've improved the technique handling in the gooch material widget to avoid showing settings which have no effect in particular contexts (e.g. hiding all but the ambient color choice for simple 2d line symbols)
- I've added a new material: "Gooch". This material gives a CAD style 3d render, where the shading is not dependent on appropriate lighting setup in a scene and where the 3d details are always visible, regardless of the lighting choice:

image


nyalldawson:
@nirvn I can't compress these svgs -- Qt chokes on the gradients if I do and shows nothing!

src/3d/qgsgoochmaterialsettings.cpp
```diff
@@ -0,0 +1,115 @@
...
+  return new QgsGoochMaterialSettings( *this );
+}
+
+void QgsGoochMaterialSettings::readXml( const QDomElement &elem, const QgsReadWriteContext & )
+{
+  mWarm = QgsSymbolLayerUtils::decodeColor( elem.attribute( QStringLiteral( "warm" ), QStringLiteral( "107,0,107" ) ) );
+  mCool = QgsSymbolLayerUtils::decodeColor( elem.attribute( QStringLiteral( "cool" ), QStringLiteral( "255,130,0" ) ) );
+  mDiffuse = QgsSymbolLayerUtils::decodeColor( elem.attribute( QStringLiteral( "diffuse" ), QStringLiteral( "178,178,178" ) ) );
+  mSpecular = QgsSymbolLayerUtils::decodeColor( elem.attribute( QStringLiteral( "specular" ) ) );
+  mShininess = elem.attribute( QStringLiteral( "shininess2" ), QStringLiteral( "100" ) ).toFloat();
```
wonder-sk, line 59: Just curious... why is it shininess2 ? :-)

wonder-sk (approved):
Awesome stuff - Gooch material has been on my wish list for some time - and as already discussed, it could even be the default material for QGIS 3D as it does not require good adjustment of lights. Now adding the physically-based rendering using QMetalRoughMaterial should be a piece of cake :grin: 

Regarding rendering techniques - that's pretty much how I imagined it. Few random notes:
- maybe for Lines we should have a separate material settings subclass (with a separate widget) - as it only uses diffuse color, it is maybe a bit surprising that its material is called "Realistic (Phong)"
- I wonder about the "terrain" rendering technique - if I understand correctly, it is currently mainly used to hide diffuse color from the list. Maybe it would be worth renaming it to something like "triangles with texture coordinates" as that's the true description of the technique...?
- regarding textures - maybe textured material should be another material settings subclass - virtual methods like textureRotation() do not seem to be great for abstract interface. I think the current texturing support is relatively basic, and we will be probably adding some more strategies on how texture coordinates are automatically assigned in tessellator, similarly at some point we may support reading of texture coordinates directly from geometry data. I do not have a clear vision how all this would be integrated - this is just a quick brain dump :-)

src/3d/qgsgoochmaterialsettings.cpp
nyalldawson: I can't work out the behaviour of the shininess parameter in the Qt materials, but it seems to be very different with the new material classes vs the older phong material class. The values which are reasonable for QPhongMaterial give silly results for gooch material. And then when switching between material types, we attempt to read the older material xml using the new material, in order to restore as many compatible properties as possible (e.g. so that the diffuse color doesn't get reset when you flip between material types). So this was a hack to avoid copying the invalid phong shininess values across the gooch, and vice-versa...

nyalldawson:
> maybe for Lines we should have a separate material settings subclass (with a separate widget) - as it only uses diffuse color, it is maybe a bit surprising that its material is called "Realistic (Phong)"
I wonder about the "terrain" rendering technique - if I understand correctly, it is currently mainly used to hide diffuse color from the list. Maybe it would be worth renaming it to something like "triangles with texture coordinates" as that's the true description of the technique...?
regarding textures - maybe textured material should be another material settings subclass - virtual methods like textureRotation() do not seem to be great for abstract interface. I think the current texturing support is relatively basic, and we will be probably adding some more strategies on how texture coordinates are automatically assigned in tessellator, similarly at some point we may support reading of texture coordinates directly from geometry data. I do not have a clear vision how all this would be integrated - this is just a quick brain dump :-)

All valid points, I'll address!

nyalldawson:
> I wonder about the "terrain" rendering technique - if I understand correctly, it is currently mainly used to hide diffuse color from the list. Maybe it would be worth renaming it to something like "triangles with texture coordinates" as that's the true description of the technique...?

Actually it's a little different, because we also hide the usual texture control settings from the widget with this technique. I added it as a way to differentiate Triangles from the terrain type, because the terrain type has a predefined texture which the user can't change. I'll tweak the naming to reflect this.

Issue #37979, opened 2020-07-27
TIN interpolation create strange triangles - Error in the interpolation


cesarcorreo:
**Describe the bug**
I'm performing a TIN interpolation following a road (12 km long and 60 m wide) with a regular layer of points in a grid of 10x10 m. 
The triangles created "look for" points or values very far.
The surface interpolated are 2x2m but also tried 0.5x0.5m 
The problem in localised in some areas (Red circles) and in others works fine (Blue circles)
I also did the interpolation in ArcGIS and it worked properly.
(Maybe limit the distance where a point look for another could solve the problem)
Last screenshot compares the results from ArcGIS (natural neighbor) in green and from QGIS (Red)

**How to Reproduce**
I'll share the point layer to the assign person (not a public file)

**QGIS and OS versions**

QGIS version
3.14.0-Pi
QGIS code revision
9f7028fd23
Compiled against Qt
5.11.2
Running against Qt
5.11.2
Compiled against GDAL/OGR
3.0.4
Running against GDAL/OGR
3.0.4
Compiled against GEOS
3.8.1-CAPI-1.13.3
Running against GEOS
3.8.1-CAPI-1.13.3
Compiled against SQLite
3.29.0
Running against SQLite
3.29.0
PostgreSQL Client Version
11.5
SpatiaLite Version
4.3.0
QWT Version
6.1.3
QScintilla2 Version
2.10.8
Compiled against PROJ
6.3.2
Running against PROJ
Rel. 6.3.2, May 1st, 2020
OS Version
Windows 10 (10.0)
Active python plugins
ale; 
AnotherDXF2Shape; 
autoSaver; 
ClosestPoint; 
contour; 
DataPlotly; 
dimensioning; 
dimensions_selector; 
EasyCustomLabeling; 
FeatureGridCreator; 
freehandEditing3; 
gdrive_provider; 
geoscience; 
GroupStats; 
HCMGIS; 
ImportPhotos; 
kmltools; 
KP_find; 
LAStools; 
latlontools; 
lines_around_point; 
LocatePoints; 
mapswipetool_plugin; 
MBTiles2img; 
Mergin; 
mmqgis; 
NNJoin; 
OSMDownloader; 
PluginLoadTimes; 
pointsamplingtool; 
polystrip; 
profiletool; 
qchainage; 
Qgis2threejs; 
qgis2web; 
qgiscloud; 
QGISSortAndNumber-master; 
qgis_resource_sharing; 
qProf; 
[...]

Issue #30549, opened 2019-07-04
TIN creation crash


IVsese:
Crash ID: 185de4a9a11eb29f05672d8b176e731c0d36ee2f

```
Stack Trace


MathUtils::lineIntersection :
DualEdgeTriangulation::insertForcedSegment :
DualEdgeTriangulation::addLine :
QgsTinInterpolator::insertData :
QgsTinInterpolator::initialize :
QgsTinInterpolator::interpolatePoint :
PyInit__analysis :
QgsGridFileWriter::writeFile :
PyInit__analysis :
PyMethodDef_RawFastCallKeywords :
PyMethodDef_RawFastCallKeywords :
PyEval_EvalFrameDefault :
PyFunction_FastCallDict :
PyMethodDef_RawFastCallDict :
PyObject_Call :
PyInit_sip :
CPLStringList::empty :
PyInit__core :
QgsProcessingAlgorithm::runPrepared :
QgsProcessingAlgRunnerTask::run :
PyInit__core :
QgsTask::start :
QThreadPoolPrivate::reset :
QThread::start :
BaseThreadInitThunk :
RtlUserThreadStart :
```

QGIS Info
QGIS Version: 3.8.0-Zanzibar
QGIS code revision: 11aff65f10
Compiled against Qt: 5.11.2
Running against Qt: 5.11.2
Compiled against GDAL: 2.4.1
Running against GDAL: 2.4.1

System Info
CPU Type: x86_64
Kernel Type: winnt
Kernel Version: 10.0.17763

gioman:
Please attach sample data and parameter used.

IVsese:
[cv05orogcurvas.ZIP](https://github.com/qgis/QGIS/files/3359318/cv05orogcurvas.ZIP)

gioman:
@IVsese zip only contains a single file (.shx). You must attach the complete shapefile and please also specify the parameters used in the TIN tool interface.

IVsese:
[oro.ZIP](https://github.com/qgis/QGIS/files/3359337/oro.ZIP)
![parameters](https://user-images.githubusercontent.com/52532674/60668483-620a2780-9e6c-11e9-95eb-afc6121876b7.png)

Thanks, I was trying to create the TIN lots of times, changings parameters but it doesn´t work

gioman:
Confirmed crash on Linux/master too, even when using a much more human "1m" as output resolution (instead of the reporter 10cm).

IVsese:
[...]

PR #38042, opened 2020-07-30
renaming and cleaning of Triangulation class and export to QgsMesh

vcloarec:
In this PR :
- renaming and cleaning of the Triangulation class and DualEdgeTriangulation class to make them more consistent with QGIS code
- add a method to export the triangulation to a QgsMesh instance

The aim of this PR is to add the possibility to construct mesh layer with triangulation from existing points and lines. This first step allows to obtain a QgsMesh that is the base of mesh layers.

### DIFF
--- a/src/analysis/CMakeLists.txt
+++ b/src/analysis/CMakeLists.txt
@@ -10,14 +10,14 @@ SET(QGIS_ANALYSIS_SRCS
   interpolation/qgstininterpolator.cpp
   interpolation/Bezier3D.cpp
   interpolation/CloughTocherInterpolator.cpp
-  interpolation/DualEdgeTriangulation.cpp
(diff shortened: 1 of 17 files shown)


nyalldawson:
@vcloarec I'm a fan of this cleanup!

Just so you're aware going into this work -- there's a serious outstanding issue with these triangulation classes. Here's a good issue report detailing the issue: https://github.com/qgis/QGIS/issues/37979 . There's a handful of other reports of similar nature (e.g. https://github.com/qgis/QGIS/issues/22486 , https://github.com/qgis/QGIS/issues/18981 ), and one of a crash: https://github.com/qgis/QGIS/issues/30549

nyalldawson (approved):

PR #38045, opened 2020-07-30
Show some more warnings when spatial indexes are missing  but desirable

nyalldawson:


src/analysis/processing/qgsalgorithmclip.cpp
```diff
@@ -96,6 +96,9 @@ QVariantMap QgsClipAlgorithm::processAlgorithm( const QVariantMap &parameters, Q
   if ( !maskSource )
     throw QgsProcessingException( invalidSourceError( parameters, QStringLiteral( "OVERLAY" ) ) );
 
+  if ( featureSource->hasSpatialIndex() == QgsFeatureSource::SpatialIndexNotPresent )
+    feedback->reportError( QObject::tr( "No spatial index exists for input layer, performance will be severely degraded" ) );
```
Gustry, line 100: Side question, would it be nice to have a feedback->reportWarning ?

nyalldawson: Yes, I'd like this too!

PR #38050, opened 2020-07-30
Do not restore geometry for maximized windows

marisn:
Geometry restore causes subsequent minimization of previously maximized windows
as observed with KWin 5.19, Qt 5.15 on X11

## Description

When a minimized window with enabled QgsGui::enabledAutoGeometryRestore is restored to its original size (by window manager), it instantly gets minimized back by call to QgsGuiUtils::restoreGeometry. It is observable as a quick flash of window and also as a removal and instant restore of _NET_WM_STATE_HIDDEN window property.
The problem is timing sensitive as fast clicking gazzillion times on window icon in task manager sometimes manages to bring out window from its minimization loop.

Generally the issue boils down to QWindow in QgsWidgetStateHelper::eventFilter [1] having state Qt::WindowMinimized when (in mean time while?) it has been already restored to maximized state by window manager. Call to widget->restoreGeometry() will not completely reset WindowStates for maximized state (as requested to be restored into) but just OR them [2] thus leading to a WindowMinimized|WindowMaximized state. At the end of restoreGeometry, WindowState is updated and propagates back to WM thus restoring of _NET_WM_STATE_HIDDEN window property and effectively minimizing the window.
I am not an expert in XCB, Qt or C++ and thus have no idea if more elegant solution is possible.
1. https://github.com/qgis/QGIS/blob/836666e782a6dcf75d3497ebd7563fda816619a9/src/gui/qgswidgetstatehelper_p.cpp
2. https://code.qt.io/cgit/qt/qtbase.git/tree/src/widgets/kernel/qwidget.cpp#n7327

## How to Reproduce original issue

1. Start a new QGIS instance;
2. Open Georeferencer window;
3. Maximize the Georeferencer window (if it is not maximized);
4. In Georeferencer window hit minimize button;
5. In the task manager click on Georeferencer window icon to restore it;
6. Observe how Georeferencer window flashes and stays minimized.

### DIFF
--- a/src/gui/qgswidgetstatehelper_p.cpp
+++ b/src/gui/qgswidgetstatehelper_p.cpp
@@ -14,6 +14,7 @@
  *                                                                         *
  ***************************************************************************/
 #include "qgswidgetstatehelper_p.h"
+#include <QWindow>
 #include <QWidget>
 #include <QEvent>
 #include <QObject>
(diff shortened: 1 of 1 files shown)


marisn:
@nyalldawson could you, please, take look at this?

nyalldawson:
It's really hard to review this one! I think we should merge into master for wider testing and to see if there's any identified regressions prior to release. @m-kuhn do you agree?

PR #38059, opened 2020-07-31
Split api from specifying 3d map layers to render from api for specifying terrain layers to render

nyalldawson:
Previously the one method was used for both, which made it very
difficult to have a different set of layers used for terrain
generation vs 3d entity generation (you had to manually create a map
theme, which is very messy)

The previous api also made it impossible to have a shaded terrain which just used the specified phong material settings, as it always overwrote the diffuse color setting.


wonder-sk (approved):
:+1: Nice!

Maybe it would be worth adding terrainLayersChanged() signal as well?

This also brings us closer to the point where map layers would be (by default?) excluded from rendering on terrain in case they have 3D renderer assigned - it seems most of the time it is unwanted to render such layers on the terrain as well...

nyalldawson:
@wonder-sk 

> This also brings us closer to the point where map layers would be (by default?) excluded from rendering on terrain in case they have 3D renderer assigned - it seems most of the time it is unwanted to render such layers on the terrain as well...

That's my thinking too!

PR #38067, opened 2020-07-31
[FEATURE][processing] Export atlas layout as image algorithm

nirvn:
## Description

This PR adds a new native processing algorithm to export layout atlases. Obligatory screenshot:
Screenshot from 2020-07-31 16-07-27

This is _super_ useful in models where remote datasets are fetched, processed, with the goal of being used as atlas coverage layers.


src/analysis/processing/qgsalgorithmlayoutatlastoimage.cpp
```diff
@@ -0,0 +1,250 @@
...
+}
+
+QString QgsLayoutAtlasToImageAlgorithm::groupId() const
+{
+  return QStringLiteral( "cartography" );
+}
+
+QString QgsLayoutAtlasToImageAlgorithm::shortDescription() const
+{
+  return QObject::tr( "Exports an atlas layout atlas as image." );
```
nyalldawson, line 56:
```suggestion
  return QObject::tr( "Exports an atlas layout as a set of images." );
```

src/analysis/processing/qgsalgorithmlayoutatlastoimage.cpp
```diff
@@ -0,0 +1,250 @@
...
+}
+
+QString QgsLayoutAtlasToImageAlgorithm::shortHelpString() const
+{
+  return QObject::tr( "This algorithm outputs an atlas layout image file (e.g. PNG or JPEG images)." );
```
nyalldawson, line 61:
```suggestion
  return QObject::tr( "This algorithm outputs an atlas layout to a set of image files (e.g. PNG or JPEG images)." );
```

I'd also suggest expanding a little here an explaining that the coverage layer and other atlas parameters can be overridden.

src/analysis/processing/qgsalgorithmlayoutatlastoimage.cpp
```diff
@@ -0,0 +1,250 @@
...
+    atlas->setEnabled( previousEnabled );
+    atlas->setCoverageLayer( previousCoverageLayer );
+    atlas->setFilenameExpression( previousFilenameExpression, error );
+    atlas->setFilterExpression( previousFilterExpression, error );
+    atlas->setSortFeatures( previousSortFeatures );
+    atlas->setSortAscending( previousSortAscending );
+    atlas->setSortExpression( previousSortExpression );
+  };
+
+  QgsVectorLayer *layer = parameterAsVectorLayer( parameters, QStringLiteral( "COVERAGE_LAYER" ), context );
```
nyalldawson, line 155: What about making the coverage layer parameter (and other atlas settings) optional, and using their existing values if not set? That way you could use this algorithm to export an existing atlas which is already fully setup in the layout designer.

src/analysis/processing/qgsalgorithmlayoutatlastoimage.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+QVariantMap QgsLayoutAtlasToImageAlgorithm::processAlgorithm( const QVariantMap &parameters, QgsProcessingContext &context, QgsProcessingFeedback *feedback )
+{
+  // this needs to be done in main thread, layouts are not thread safe
+  QgsPrintLayout *layout = parameterAsLayout( parameters, QStringLiteral( "LAYOUT" ), context );
+  if ( !layout )
+    throw QgsProcessingException( QObject::tr( "Cannot find layout with name \"%1\"" ).arg( parameters.value( QStringLiteral( "LAYOUT" ) ).toString() ) );
+
+  QString error;
+  QgsLayoutAtlas *atlas = layout->atlas();
```
nyalldawson, line 126: This is fine for now, but we should add a way to clone a layout to avoid this in future.

src/analysis/processing/qgsalgorithmlayoutatlastoimage.cpp
```diff
@@ -0,0 +1,250 @@
...
+  atlas->setFilenameExpression( expression, error );
+  if ( !error.isEmpty() )
+  {
+    restoreAtlas();
+    throw QgsProcessingException( QObject::tr( "Error setting atlas filename expression" ) );
+  }
+
+  expression = parameterAsString( parameters, QStringLiteral( "FILTER_EXPRESSION" ), context );
+  atlas->setFilterExpression( expression, error );
+
```
nyalldawson, line 169: Also raise this error?

nyalldawson (commented):
Looking good! Can you add some tests too?

src/analysis/processing/qgsalgorithmlayoutatlastoimage.cpp
nirvn: How about this: if a coverage layer is not provided, the algorithm exports whatever is already setup in the designer (and throws an error if a layout doesn't have an enabled atlas). If a coverage layer is provided, then all settings are overwritten using the parameters. 

The need to overwrite everything arises from the fact that we currently thread empty parameters (eg sort by expression) as a way to tell the algorithm to deactivate sorting.

src/analysis/processing/qgsalgorithmlayoutatlastoimage.cpp
nirvn: Yeah; I also was thinking it could be useful to have a QgsPrintLayout::setAtlas( QgsLayoutAtlas *atlas ) function too in the future.

src/analysis/processing/qgsalgorithmlayoutatlastoimage.cpp
nyalldawson: Sounds good! Just make sure the docs explain this behaviour

nirvn:
Review addressed, test case added.

nyalldawson:
Can you extend the test to check for the new layers parameter too? Specifically that if not specified, the original layers are used, and if specified then the layers are applied

nirvn:
@nyalldawson , is there a way to do that without image comparison?

nirvn:
(Travis failure unrelated)

PR #38070, opened 2020-07-31
show current layer CRS in the layer context menu and allow to change it using recent CRSs

### LINKED ISSUE #13882: make it possible to see & change layer CRS directly from layer menu
Author Name: **Mathieu Pellerin - nIRV** (Mathieu Pellerin - nIRV)
Original Redmine Issue: [3824](https://issues.qgis.org/issues/3824)

Redmine category:gui

---
It'd be nice for QGIS to display the 3 most recently used CRS items directly from the layer menu and allow the user to click on them to set the layer's CRS to the selected item.

Accompanying this change, QGIS should also display the current layer CRS directly in the menu.

Resulting layer menu would look like this:

```
/---------------------\\
| Zoom to layer...    |
| Show in overview    |
| Remove              |
| ------------------- |/------------------------------\\
| Layer CRS         > || Current CRS (disabled state) |
| ------------------- || Set project to Layer CRS     |
| [...]               || ---------------------------- |
|                     || Recent #13929                    |
|                     || Recent #13930                    |
|                     || Recent #13931                    |
|                     || ---------------------------- |
|                     || Open CRS selector...         |
|                     |\\------------------------------/
|                     |
\\---------------------/

```

IMO, positive aspects of this change are:
* A good way to increase efficiency with regards to layers CRS management. 
* Allow for the user to skip the time QGIS needs to compile the CRS list & create the CRS selector window in cases he/she only wants to select most recently used CRS. (as it currently is, even on high-end windows machines, the user has to wait a couple of seconds for the window to show up) 
* Remove the 'set project to layer CRS' item from the root menu, making the root menu less busy by removing an item rarely used (in theory, the user would use this function only once per project)
* Allow for the user to quickly see the layer CRS without opening the property window, also useful in deciding whether to use layer CRS as project CRS

alexbruy:
## Description
Adds current layer CRS directly to the layer's context menu so it can be easily discoverable. Also allows to quickly set layer CRS to one of the three recently used CRSs without opening CRS selector.

crs

Fixes #13882.

### DIFF
--- a/src/app/qgsapplayertreeviewmenuprovider.cpp
+++ b/src/app/qgsapplayertreeviewmenuprovider.cpp
@@ -346,7 +346,7 @@ QMenu *QgsAppLayerTreeViewMenuProvider::createContextMenu()
           {
             QAction *action = menuSetCRS->addAction( tr( "Set to %1" ).arg( crs.userFriendlyIdentifier( QgsCoordinateReferenceSystem::ShortString ) ) );
             action->setProperty( "layerId", layer->id() );
-            action->setProperty( "crs", crs.toWkt() );
+            action->setProperty( "crs", crs.toWkt( QgsCoordinateReferenceSystem::WKT_PREFERRED ) );
             i++;
             if ( i == 2 )
               break;
--- a/src/app/qgsapplayertreeviewmenuprovider.cpp
+++ b/src/app/qgsapplayertreeviewmenuprovider.cpp
@@ -328,15 +328,38 @@ QMenu *QgsAppLayerTreeViewMenuProvider::createContextMenu()
         if ( !layer->isInScaleRange( mCanvas->scale() ) )
           menu->addAction( tr( "Zoom to &Visible Scale" ), QgisApp::instance(), &QgisApp::zoomToLayerScale );
 
-        QMenu *menuSetCRS = new QMenu( tr( "Set CRS" ), menu );
-        // set layer crs
-        QAction *actionSetLayerCrs = new QAction( tr( "Set Layer CRS…" ), menuSetCRS );
-        connect( actionSetLayerCrs, &QAction::triggered, QgisApp::instance(), &QgisApp::setLayerCrs );
-        menuSetCRS->addAction( actionSetLayerCrs );
+        QMenu *menuSetCRS = new QMenu( tr( "Layer CRS" ), menu );
+        QAction *actionCurrentCrs = new QAction( layer->crs().authid(), menuSetCRS );
+        actionCurrentCrs->setEnabled( false );
+        menuSetCRS->addAction( actionCurrentCrs );
         // assign layer crs to project
         QAction *actionSetProjectCrs = new QAction( tr( "Set &Project CRS from Layer" ), menuSetCRS );
         connect( actionSetProjectCrs, &QAction::triggered, QgisApp::instance(), &QgisApp::setProjectCrsFromLayer );
         menuSetCRS->addAction( actionSetProjectCrs );
+
+        const QList< QgsCoordinateReferenceSystem> recentProjections = QgsCoordinateReferenceSystem::recentCoordinateReferenceSystems();
+        if ( !recentProjections.isEmpty() )
+        {
+          menuSetCRS->addSeparator();
+          int i = 0;
+          for ( const QgsCoordinateReferenceSystem &crs : recentProjections )
+          {
+            QAction *action = menuSetCRS->addAction( crs.authid() );
+            action->setProperty( "layerId", layer->id() );
+            action->setProperty( "crs", crs.authid() );
+            i++;
+            if ( i == 2 )
+              break;
+          }
+        }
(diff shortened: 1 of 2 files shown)


src/app/qgsapplayertreeviewmenuprovider.cpp
```diff
@@ -328,16 +328,39 @@ QMenu *QgsAppLayerTreeViewMenuProvider::createContextMenu()
         if ( !layer->isInScaleRange( mCanvas->scale() ) )
           menu->addAction( tr( "Zoom to &Visible Scale" ), QgisApp::instance(), &QgisApp::zoomToLayerScale );
 
-        QMenu *menuSetCRS = new QMenu( tr( "Set CRS" ), menu );
-        // set layer crs
-        QAction *actionSetLayerCrs = new QAction( tr( "Set Layer CRS…" ), menuSetCRS );
-        connect( actionSetLayerCrs, &QAction::triggered, QgisApp::instance(), &QgisApp::setLayerCrs );
-        menuSetCRS->addAction( actionSetLayerCrs );
+        QMenu *menuSetCRS = new QMenu( tr( "Layer CRS" ), menu );
+        QAction *actionCurrentCrs = new QAction( layer->crs().authid(), menuSetCRS );
```
nirvn, line 332:
```suggestion
        QAction *actionCurrentCrs = new QAction( layer->crs().userFriendlyIdentifier(), menuSetCRS );
```

src/app/qgsapplayertreeviewmenuprovider.cpp
```diff
@@ -328,16 +328,39 @@ QMenu *QgsAppLayerTreeViewMenuProvider::createContextMenu()
...
         menuSetCRS->addAction( actionSetProjectCrs );
 
+        const QList< QgsCoordinateReferenceSystem> recentProjections = QgsCoordinateReferenceSystem::recentCoordinateReferenceSystems();
+        if ( !recentProjections.isEmpty() )
+        {
+          menuSetCRS->addSeparator();
+          int i = 0;
+          for ( const QgsCoordinateReferenceSystem &crs : recentProjections )
+          {
+            QAction *action = menuSetCRS->addAction( crs.userFriendlyIdentifier() );
```
nyalldawson, line 347: Could we make this entry say "Set to xxx..." instead of just the crs name?

src/app/qgsapplayertreeviewmenuprovider.cpp
```diff
@@ -328,16 +328,39 @@ QMenu *QgsAppLayerTreeViewMenuProvider::createContextMenu()
...
+            action->setProperty( "layerId", layer->id() );
+            action->setProperty( "crs", crs.authid() );
```
nyalldawson, line 349: Not all crs have an auth id. The only reliable way to handle this would be to export the crs to wkt and use this as the data here.

src/app/qgsapplayertreeviewmenuprovider.cpp
```diff
@@ -328,16 +328,39 @@ QMenu *QgsAppLayerTreeViewMenuProvider::createContextMenu()
...
+        const QList< QgsCoordinateReferenceSystem> recentProjections = QgsCoordinateReferenceSystem::recentCoordinateReferenceSystems();
+        if ( !recentProjections.isEmpty() )
+        {
+          menuSetCRS->addSeparator();
+          int i = 0;
+          for ( const QgsCoordinateReferenceSystem &crs : recentProjections )
+          {
+            QAction *action = menuSetCRS->addAction( tr( "Set to %1" ).arg( crs.userFriendlyIdentifier( QgsCoordinateReferenceSystem::ShortString ) ) );
+            action->setProperty( "layerId", layer->id() );
+            action->setProperty( "crs", crs.toWkt() );
```
nyalldawson, line 349:
```suggestion
            action->setProperty( "crs", crs.toWkt( QgsCoordinateReferenceSystem::WKT_PREFERRED ) );
```

nyalldawson (approved):

roya0045:
I never understood the need to change a crs of a given layer when it is initially properly projected.

I think people will mistake this for a quick in-place transform because" Who would think that exposing a way to change the CRS without reprojection is a good idea."

gioman:
> I never understood the need to change a crs of a given layer when it is initially properly projected.

@roya0045 there isn't, but the problem that there are a number of datasources that need to have the CRS manually set up. I.e. by choice we add KML and GPX always in WGS84, as they should be, but the truth is that frequently such datasources have a different CRS. Also shapefiles with a wrong CRS are not unsual. Also there is the case (less problematic) where a datasource has a CRS that is written in  way is formally correct but do not match any entry in QGIS db, so the layer is added with a custom/unknown CRS.

roya0045:
@gioman I never had to deal with bad CRS other than on old files with missing projection, so doing it through the property menu was fine for me. If such cases are frequent enough to warrant a new menu item then it is fine. Though they were always very rare for me. Thank for the reply.

Issue #32981, opened 2019-11-20
3D View Skybox


Saijin-Naib:
**Feature description.**
3D game engines oftentimes feature support for a Skybox, which is a 3D cube with inverted normals rendered at a sufficiently large size as to encompass the entire scene. Each face of the cube has an image applied to it so that when the camera points in any direction, the image is rendered onto the screen, giving the impression of a fully realized sky/horizon.

Having a skybox in QGIS's 3D View would allow for more photorealistic rendering, as well as possibly more stylized rendering if wanted/needed.

**Additional context**
The implementation I am most familiar with comes from the iDTech family of 3D game engines. In this implementation, six images of the same format are put into a directory, and the engine applies them to the face of the cube by reading a suffix on the filename.

The suffixes are:  
bk [back]
ft [front]
lf [left face]
rt [right face]
up [up]
dn [down]

The engine is expected to find the images either with or without a prepended underscore before the suffix, for instance:  
sky_bk.png
sky_ft.png
skylf.png
sky_rt.png
skyup.png
skydn.png

Documentation from the QuakeWiki:
https://quakewiki.org/wiki/Skybox_Support

This format for skyboxes is very mature and has seen adoption as a de-facto standard in many games/engines due to the adoption of the iDTech engines, so there are numerous pre-made skyboxes that could be dropped in as-is, which is a big benefit.

alexbruy:
Seems implemented in #38087.

Saijin-Naib:
Yes and no. His implementation doesn't work with the huge library of existing sky boxes from game engines, so would this still be able to be kept open as an expansion of the current implementation?

kannes:
Just ran into this when I tried to set a skybox using this good-old format. Would be a great feature!

Re-open please.

Saijin-Naib:
Yeah, that was my thinking when I asked for this format. We have something like 30 years of skyboxes created in this format to pick through... Should be pretty awesome for end-users.

kannes:
To be honest I did not get any panoramic sky to work either, nothing ever shows up for me. Are you getting anything displayed at all when using the bk, ft, lf, rt, up, dn format?

Saijin-Naib:
I've not tried in a long while, but last I tried, it didn't work for me either. It has been quite a few versions, though. Something to check this week!

I'll pull a skybox from Quake2 :D

Saijin-Naib:
[...]

PR #38087, opened 2020-08-02
[3D feature] Skybox feature

NEDJIMAbelgacem:
## Skybox rendering
QGIS 3D views support rendering skyboxes.
From the the 3D configuration dialog you can enable the skybox rendering and choose from 2 different ways of providing skyboxes.
### Different textures for each face:
distinct_faces
Specify each texture seperately according to which face it belongs to.
### High resolution 360 image:
360Image
You can use 360 images like .hdr ones from [hdrihaven](https://hdrihaven.com/)
### GIFs of the skybox rendering:
skybox1
skybox2


src/3d/qgs3dmapscene.cpp
```diff
@@ -325,7 +320,6 @@ void Qgs3DMapScene::onCameraChanged()
 
 void Qgs3DMapScene::updateScene()
 {
-  QgsEventTracing::addEvent( QgsEventTracing::Instant, QStringLiteral( "3D" ), QStringLiteral( "Update Scene" ) );
```
nyalldawson, line 328: What's the rationale behind removing this?

src/3d/qgs3dmapscene.cpp
```diff
@@ -831,6 +825,10 @@ void Qgs3DMapScene::addCameraViewCenterEntity( Qt3DRender::QCamera *camera )
   {
     mEntityCameraViewCenter->setEnabled( mMap.showCameraViewCenter() );
   } );
+//  connect( &mMap, &Qgs3DMapSettings, this, [this]
```
nyalldawson, line 828: remove

src/3d/qgs3dmapscene.cpp
```diff
@@ -861,6 +859,40 @@ void Qgs3DMapScene::updateSceneState()
   setSceneState( Ready );
 }
 
+void Qgs3DMapScene::onSkyboxSettingsChanged()
+{
+  qDebug() << __FUNCTION__;
```
nyalldawson, line 864: remove

src/3d/qgsskyboxentity.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+  mMaterial->setEffect( mEffect );
+  mMaterial->addParameter( mGammaStrengthParameter );
+  mMaterial->addParameter( mTextureParameter );
+
+  mMesh->setXYMeshResolution( QSize( 2, 2 ) );
+  mMesh->setXZMeshResolution( QSize( 2, 2 ) );
+  mMesh->setYZMeshResolution( QSize( 2, 2 ) );
+
+//   TODO: change the kybox position according to camera
```
nyalldawson, line 85:
```suggestion
//   TODO: change the skybox position according to camera
```

src/3d/qgsskyboxentity.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+  mTextureParameter->setName( "skyboxTexture" );
+  mTextureParameter->setValue( QVariant::fromValue( mLoadedTexture ) );
+
+  reloadTexture();
+}
+
+void QgsHDRSkyboxEntity::reloadTexture()
+{
+  mLoadedTexture->setSource( QUrl::fromUserInput( mHDRTexturePath ) );
```
nyalldawson, line 116: This will need to use QgsImageCache, just like the textures in the phong material (otherwise embedded textures won't work)

src/3d/qgsskyboxentity.h
```diff
@@ -0,0 +1,106 @@
...
+ */
+class _3D_EXPORT QgsSkyboxEntity : public Qt3DCore::QEntity
+{
+    Q_OBJECT
+  public:
+    //! Constructor
+    QgsSkyboxEntity( QNode *parent = nullptr );
+
+  protected:
+    Qt3DRender::QEffect *mEffect;
```
nyalldawson, line 47: initialize all these to nullptr here

src/3d/qgsskyboxentity.h
```diff
@@ -0,0 +1,106 @@
...
+ * \since QGIS 3.16
+ */
+class _3D_EXPORT QgsHDRSkyboxEntity : public QgsSkyboxEntity
+{
+  public:
+    //! Construct a skybox from a high resolution 360 image
+    QgsHDRSkyboxEntity( const QString &hdrTexturePath, Qt3DCore::QNode *parent = nullptr );
+
+    //! Returns the path of the current texture in use
+    QString getHDRTexturePath() const { return mHDRTexturePath; }
```
nyalldawson, line 69:
```suggestion
    QString hdrTexturePath() const { return mHDRTexturePath; }
```

src/3d/qgsskyboxsettings.h
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * Returns the type of the skybox
+     * This can be one of 3 types: "Textures collection", "HDR texture", "Distinct Faces"
+     */
+    QString skyboxType() const { return mSkyboxType; }
+
+    /**
+     * Sets the type of the skybox
+     * the type can be one of 3 types: "Textures collection", "HDR texture", "Distinct Faces"
```
nyalldawson, line 84: It would be cleaner to use an enum for these types, instead of the string values

src/3d/qgsskyboxsettings.h
```diff
@@ -0,0 +1,127 @@
...
+
+    //! Returns the extension of a skybox of type "Textures collection"
+    QString skyboxExtension() const { return mSkyboxExt; }
+    //! Sets the extension of a skybox of type "Textures collection"
+    void setSkyboxExtension( const QString &extension ) { mSkyboxExt = extension; }
+
+    //! Returns the HDR texture path of a skybox of type "HDR texture"
+    QString hdrTexturePath() const { return mHDRTexturePath; }
+    //! Sets the HDR texture path of a skybox of type "HDR texture"
+    void setHDRTexturePath( const QString &texturePath ) { mHDRTexturePath = texturePath; }
```
nyalldawson, line 101:
```suggestion
    void setHdrTexturePath( const QString &texturePath ) { mHDRTexturePath = texturePath; }
```

nyalldawson (commented):
Nice!

src/3d/qgs3dmapscene.cpp
NEDJIMAbelgacem: Ah that was just a random mistake. Sorry about that

src/3d/qgsskyboxentity.cpp
NEDJIMAbelgacem: I tried to do this for HDR images but apparently that's not supported with QImage.

NEDJIMAbelgacem:
Thanks for the reviews @nyalldawson !

NEDJIMAbelgacem:
Hi @DelazJ , what documentations I need to add ?

DelazJ:
Hi @NEDJIMAbelgacem 

QGIS has a user-oriented documentation at https://docs.qgis.org, and some 3D stuffs at https://docs.qgis.org/testing/en/docs/user_manual/introduction/qgis_gui.html#d-map-view. The "Needs Documentation" label is meant to generate, once this PR is merged, an issue report in the [docs repository](www.github.com/qgis/QGIS-Documentation/issues) with full content of your first message, so that writers are aware of this new feature and document it. So the clearer and up to date the first message of the PR is (and i feel it is the case here), the most helpful it's for docs team (which btw welcomes anyone willing to help write about _their_ features... :wink: ).
Thanks for the contributions.

NEDJIMAbelgacem:
Ah I see, I wrote the PR message from "what I changed" point of view. I will write more user oriented message for this PR and my other PRs as well. Thanks @DelazJ

wonder-sk (commented):
Great job - the code is looking very good!

I have a few suggestions:
- we can probably safely get rid of TexturesCollectionSkybox type... it is essentially the same thing as DistinctTexturesSkybox just with fixed naming of files, and I have not really seen that particular naming scheme (chosen in the original qt3dextras implementation) used anywhere else - that's why I would prefer to get rid of it and make both GUI and code cleaner
- can we rename HDRSkybox to "Panoramic" skybox (both in the code and in GUI)? I think in this case it does not really matter whether the images are HDR (high dynamic range) or not - what really matters that they should be panoramic. This would be also closer to how this format is called elsewhere (https://docs.unity3d.com/Manual/skybox-shaders.html). I guess it is just a coincidence that the skybox panoramas you were using were also HDR images :-)
- it may be interesting to have support for "6 sides" type skybox (as called in Unity) as it seems that some people/engines use this format - e.g. http://scmapdb.com/wad:skybox-skies
- naming of files in GUI for Distinct Textures Skybox (Neg Y, Pos X, ...) could be a bit more user friendly - e.g. Right (+X), Left (-X), Top (+Y), ...

src/3d/qgs3dmapsettings.h
```diff
@@ -426,6 +412,16 @@ class _3D_EXPORT Qgs3DMapSettings : public QObject, public QgsTemporalRangeObjec
      */
     double outputDpi() const { return mDpi; }
 
+    /**
+     * Returns the current configuration of the skybox
+     */
+    QgsSkyboxSettings skyboxSettings() const SIP_SKIP { return mSkyboxSettings; }
```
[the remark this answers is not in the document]
NEDJIMAbelgacem: It's because when I changed the skybox type to being an enum I both didn't know how to correctly export it and the enum was inside QgsSkyboxEntity class that inherits from Qt3DCore::QEntity which wasn't available for python bindings.

src/3d/qgsskyboxentity.cpp
```diff
@@ -0,0 +1,199 @@
...
+  addComponent( mMesh );
+  addComponent( mMaterial );
+}
+
+// HDR skybox
+
+QgsHDRSkyboxEntity::QgsHDRSkyboxEntity( const QString &hdrTexturePath, QNode *parent )
+  : QgsSkyboxEntity( parent )
+  , mHDRTexturePath( hdrTexturePath )
+  , mLoadedTexture( new Qt3DRender::QTextureLoader( parent ) )
```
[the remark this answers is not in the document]
NEDJIMAbelgacem: I followed what Qt3D does. I tried using Qt3DRender::QTextureImage but it didn't work properly and opengl was using a random texture from the scene.

src/3d/qgsskyboxentity.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+  mMaterial->setEffect( mEffect );
+  mMaterial->addParameter( mGammaStrengthParameter );
+  mMaterial->addParameter( mTextureParameter );
+
+  mMesh->setXYMeshResolution( QSize( 2, 2 ) );
+  mMesh->setXZMeshResolution( QSize( 2, 2 ) );
+  mMesh->setYZMeshResolution( QSize( 2, 2 ) );
+
+  //   TODO: change the skybox position according to camera
```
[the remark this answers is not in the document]
NEDJIMAbelgacem: No it's not. The skybox cube position shouldn't matter because it gets removed with the vertex shader.

NEDJIMAbelgacem:
Thanks for the reviews @wonder-sk !

wonder-sk (approved):

DelazJ:
> Ah I see, I wrote the PR message from "what I changed" point of view. I will write more user oriented message for this PR and my other PRs as well.

@NEDJIMAbelgacem It would be nice, indeed. Thanks. And it would also help since this description will be picked to generate the release changelog.
Otherwise, I opened some 3D docs PR at https://github.com/qgis/QGIS-Documentation/pull/5994 and https://github.com/qgis/QGIS-Documentation/pull/5993 and I'm not quite confident, so if ever you want to give it a look... Thanks

Saijin-Naib:
> Great job - the code is looking very good!
> I have a few suggestions:
> it may be interesting to have support for "6 sides" type skybox (as called in Unity) as it seems that some people/engines use this format - e.g. http://scmapdb.com/wad:skybox-skies

I raised this as a Feature Request in Issue #32981

waytgibbs:
In attempting to use the new Skybox feature with a panoramic HDR background in the 9/15/2020 nightly build, I'm observing rendering issues when exporting animation frames from a keyframed flyover sequence in the 3D map view. See attached images for a four-frame sequence that illustrates the issue, which occurs at irregular intervals. These glitches are not visible when stepping through the flyover interactively within QGIS.
HiRes51383
HiRes51384
HiRes51385
HiRes51386

PR #38090, opened 2020-08-02
More functions updates

DelazJ:
fix examples and description
add/update examples to "cross-link" functions (for docs purpose)
Use 'value' parameter instead of 'val'


resources/function_help/json/aggregate
```diff
@@ -8,14 +8,15 @@
...
 	{"arg":"expression", "description":"sub expression or field name to aggregate"},
 	{"arg":"filter", "optional":true, "description":"optional filter expression to limit the features used for calculating the aggregate. Fields and geometry are from the features on the joined layer. The source feature can be accessed with the variable @parent."},
-	{"arg":"concatenator", "optional":true, "description":"optional string to use to join values for 'concatenate' aggregate"},
-	{"arg":"order_by", "optional":true, "description":"optional filter expression to order the features used for calculating the aggregate. Fields and geometry are from the features on the joined layer."}
+	{"arg":"concatenator", "optional":true, "default":"''", "description":"optional string to use to join values for 'concatenate' aggregate"},
```
DelazJ, line 11: Or is this NULL value? Or should it be `"default":"EMPTY STRING"`

resources/function_help/json/aggregate
```diff
@@ -8,14 +8,15 @@
...
+	{"arg":"order_by", "optional":true, "description":"optional filter expression to order the features used for calculating the aggregate. Fields and geometry are from the features on the joined layer. By default, no sort is done"}
```
DelazJ, line 12: Or are they pulled based on their $id? And is it worth a mention?
This would need to be applied to other aggregates, once ready

DelazJ (commented):
Added some questions

resources/function_help/json/aggregate
m-kuhn: I think an empty string is fine here, I would leave as is.

resources/function_help/json/aggregate
m-kuhn: It's correct like it is. (There is a reasonably high chance that you will get stable order but) we do not guarantee a stable order.

I would use the wording from the postgres doc "By default, the features will be returned in an unspecified order". (https://www.postgresql.org/docs/12/queries-order.html)

nyalldawson (approved):

PR #38096, opened 2020-08-03
New algorithm "Flatten Relationship"

nyalldawson:
This algorithm flattens all relationships for a vector layer,
exporting a single layer containing one master feature per
related feature. This master feature contains all the
attributes for the related features.

It's designed as a quick way to de-normalize a relation from
a project, e.g. to allow exporting to CSV

Sponsored by SMEC/SJ

### DIFF
--- a/src/analysis/processing/qgsalgorithmflattenrelationships.cpp
+++ b/src/analysis/processing/qgsalgorithmflattenrelationships.cpp
@@ -29,7 +29,7 @@ QString QgsFlattenRelationshipsAlgorithm::name() const
 
 QString QgsFlattenRelationshipsAlgorithm::displayName() const
 {
-  return QObject::tr( "Flatten relationships" );
+  return QObject::tr( "Flatten relationship" );
 }
 
 QStringList QgsFlattenRelationshipsAlgorithm::tags() const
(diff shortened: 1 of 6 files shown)


src/analysis/processing/qgsalgorithmflattenrelationships.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+  QgsVectorLayer *layer = parameterAsVectorLayer( parameters, QStringLiteral( "INPUT" ), context );
+  if ( !layer )
+    throw QgsProcessingException( invalidSourceError( parameters, QStringLiteral( "INPUT" ) ) );
+
+  const QList<QgsRelation> relations = project->relationManager()->referencedRelations( layer );
+  if ( relations.size() > 1 )
+    throw QgsProcessingException( QObject::tr( "Found %1 relations. This algorithm currently supports only a single relation." ).arg( relations.size() ) );
+
+  if ( relations.empty() )
```
3nids, line 94:
```suggestion
  else if ( relations.empty() )
```

src/analysis/processing/qgsalgorithmflattenrelationships.cpp
```diff
@@ -0,0 +1,159 @@
...
+    throw QgsProcessingException( QObject::tr( "No relations found." ) );
+
+  mRelation = relations.at( 0 );
```
3nids, line 97: While reading the description, I was expecting to get all the relations for this layer.
If this is not the case, this should be explicitly stated that the alg randomly takes the first relation in the list.

nyalldawson: Thanks, I've reworded the name/description to clarify this. I couldn't think of a clean way to handle more than one relationship for a layer -- exporting a full cross join between all relations seems very over the top, and unlikely what a user is wanting...

src/analysis/processing/qgsalgorithmflattenrelationships.cpp
```diff
@@ -0,0 +1,158 @@
...
+#include "qgsalgorithmflattenrelationships.h"
+#include "qgsrelationmanager.h"
+#include "qgsvectorlayer.h"
+#include "qgsvectorlayerfeatureiterator.h"
+
+///@cond PRIVATE
+
+QString QgsFlattenRelationshipsAlgorithm::name() const
+{
+  return QStringLiteral( "flattenrelationships" );
```
DelazJ, line 27: @nyalldawson based on the changes done in the PR due to review, should this instead be _flattenrelationship_? Or too late or you expect it addresses all relations one day?
Also the alg allows to skip the output, [optional]. Intended?

PR #38097, opened 2020-08-03
Fix strange behaviour when mix of model outputs and fixed inputs are selected as the input for a multi layer parameter

nyalldawson:
@nirvn as identified by you...


nirvn (approved):
:tada: fixes both the crash and the missing layers situation.

PR #38099, opened 2020-08-03
Be more tolerant, and expose map layer outputs as possible inputs for multi input parameters regardless of the required type

nyalldawson:
Otherwise we potentially block valid use cases. Better to be flexible
and catch errors at run time then prevent users making the model they need...


nyalldawson:
@nirvn you'll also need this one

nirvn (approved):

PR #38100, opened 2020-08-03
fix escaping of $ for Make Generator (MacOS)

PeterPetrik:
https://github.com/qgis/QGIS/runs/926569184?check_suite_focus=true#step:6:2313

not sure what has changed in last 4 days, but the escaping of `$` the "bash-way" is not working anymore on macos (both CI and my desktop)


PeterPetrik:
it looks like it compiles on macbuild.qgis.org too

3nids:
might be superseded by #38103

src/core/CMakeLists.txt
```diff
@@ -702,7 +702,11 @@ SET(QGIS_CORE_SRCS
 
 FILE(GLOB JSON_HELP_FILES "${CMAKE_SOURCE_DIR}/resources/function_help/json/*")
 IF(NOT USING_NINJA)
-  STRING(REPLACE "$" "$$" JSON_HELP_FILES "${JSON_HELP_FILES}")
+  IF (APPLE)
+    STRING(REPLACE "$" "\$" JSON_HELP_FILES "${JSON_HELP_FILES}")
```
SrNetoChan, line 706: So this was my fault? We did some changes to JSON files, was that the problem? Is there something I could have done to avoid it?

PeterPetrik: I do not think it is your problem. I do not know something changed probably with CMake or Make or whoknows...

PR #36448, opened 2020-05-14
build with VS2019

jef-n:

PR #38102, opened 2020-08-03
build with VS2019

jef-n:
revives #36448


nyalldawson:
Exciting!!!

nyalldawson:
@jef-n while you're reworking the package-nightly.cmd script, would it be possible to split the build step from the test run? From my research it's impossible to get the full test debugging output with the combined build/run approach currently used, which makes it impossible to diagnose failing windows tests via the CI. I'd love to get the blacklist trimmed here, but can't see any way to do it using the current approach...

m-kuhn:
@jef-n what is the status of this? the green tick looks promising

jef-n:
There are pending osgeo4w changes (replacement of nircmd with brspawn) that concern other packages (all qgis packages except qgis-dev and saga-ltr) and of course other osgeo4w updates (SpatialLite/GDAL/PROJ/Qt/Python etc.; elimination of old msvcrts) that probably should be done before redoing the former.

m-kuhn:
would it make sense to disable the windows azure tests meanwhile?

jef-n:
> would it make sense to disable the windows azure tests meanwhile?

Why?

m-kuhn:
All master builds are red currently.

m-kuhn:
And to counter to the potential answer "shouldn't we fix the failing things then": I have tried, but I'm failing to see what's wrong because azure doesn't allow me to have a look at it.

jef-n:
Rerunning azure pipeline for 0b9a75c https://dev.azure.com/qgis/QGIS/_build/results?buildId=8240&view=results to make sure it a qgis change and not something else.  That apparently was the last master version that built on azure.

m-kuhn:
Seems to be failing now.

And this happens when I click on "Published Test Run : https://dev.azure.com/qgis/QGIS/_TestManagement/Runs?runId=22620&_a=runCharts."

image

m-kuhn:
Yes. Interesting enough your [direct link above works](https://github.com/qgis/QGIS/pull/38102#issuecomment-674887720). It's just that I can't navigate there from a failing build. Strange, this azure place.

jef-n:
> Yes. Interesting enough your [direct link above works](https://github.com/qgis/QGIS/pull/38102#issuecomment-674887720). It's just that I can't navigate there from a failing build. Strange, this azure place.

I get to that page from the "Details" link the github's azure results. From there I can get to the tests uploaded to azure with the link below "Tests and coverage" on the right (https://dev.azure.com/qgis/QGIS/_build/results?buildId=8240&view=ms.vss-test-web.build-test-results-tab).   The "Published Test Run :" links are unreachable for me too - but 404.  But the link to the results in dash is in the "Building QGIS" step anyway.

m-kuhn:
They are disabled now, I'll try this when the next generation of windows builds go live

PR #38103, opened 2020-08-03
Fix build with CMake 3.18

adriaandegroot:
It looks like CMake 3.18 fixed escaping of `$` in generated Makefiles. It does on FreeBSD, at least, where I had to apply this patch in packaging to get existing QGis packages to build against newer CMake. I'm still checking if this doesn't break with older CMake, though -- I'm fairly confident of the syntax, but you never know until it compiles.


PeterPetrik:
I have cmake version 3.17.0 and have the same problem on MacOS, tried to fix with https://github.com/qgis/QGIS/pull/38100 but this is maybe better.

3nids:
From 3.18 [release notes](https://cmake.org/cmake/help/v3.18/release/3.18.html#id1):

> In CMake 3.18.0 the add_test() command learned to support special characters in test names. This was accidentally left out of its release notes. Unfortunately the fix breaks existing projects that were using manual quoting or escaping to work around the prior limitation. This fix has been reverted in 3.18.1, but may be re-introduced in future versions of CMake with a policy for compatibility.

Is this the issue here?
So this might be for 3.18.0 only, and "fixed" in 3.18.1 ?

PeterPetrik:
I have the problem in 3.17.0, and I build it without test...

nyalldawson:
Let's merge to get the CI fixed, and then see if @PeterPetrik fix is also required....

gdt:
As a data point, qgis 3.10.7 built fine on NetBSD with cmake 3.17.3.  It failed to build with 3.18.2 and 3.18.3, due to $$$$area.  qgis 3.10.10, which has this changes, builds ok with 3.18.3.   (I did not test 3.17.3, as pkgsrc now has 3.18.3.)

@3nids The quoting is not about test names, so the information you found in the release notes seems to be something else.  But, it seems there was  a general cleanup pass in quoting.   From a skim of the cmake sources, the entire quoting process seems complicated and difficult!

PR #38111, opened 2020-08-03
fix long reload of filter change in attribute table

### LINKED ISSUE #38018: "Show selected features" is very slow in recent versions
There seems to be a performance regression concerning "Show selected features" in the attribute table of recent QGIS versions. (See also https://twitter.com/DrChrisGeoSci/status/1288120808951312389)

I've tested this with the Natural Earth dataset ne_10m_populated_places layer. The expression "NAME" like 'A%' is close to instantaneous but switching to "Show selected features" takes forever and freezes QGIS master and 3.14 in a way that looks like it crashed (i.e. "not responding" on Windows). 

In 3.4 "Show selected features" is much faster. There's no freeze.

signedav:
Avoid to call the slot updateEditSelectionDependencies on every insertRow / removeRow of the model by having a timer.

this fixes #38018

Additionally do not emit the currentEditSelectionChanged when no selection has been changed in updateEditSelectionDependencies. Only emit the signal to achieve an update of e.g. the count of features in the attribute table form view.

The code to get the index is now in editSelectionChanged and updateEditSelectionDependencies  but the code gets better readable like this.

### DIFF
--- a/src/gui/attributetable/qgsfeaturelistview.cpp
+++ b/src/gui/attributetable/qgsfeaturelistview.cpp
@@ -162,11 +162,7 @@ void QgsFeatureListView::editSelectionChanged( const QItemSelection &deselected,
     QItemSelection localSelected = mModel->mapSelectionFromMaster( selected );
     viewport()->update( visualRegionForSelection( localDeselected ) | visualRegionForSelection( localSelected ) );
   }
-  updateEditSelectionDependencies();
-}
 
-void QgsFeatureListView::updateEditSelectionDependencies()
-{
   QItemSelection currentSelection = mCurrentEditSelectionModel->selection();
   if ( currentSelection.size() == 1 )
   {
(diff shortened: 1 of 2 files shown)


nyalldawson:
I'm sorry adopt on a "angry old school Linus Torvalds vibe", but this is a terrible fix! The attribute table is already held together by chewing gum, but fixes like this just bump the hacks up an order of magnitude! :angry: Isn't there any other approach? Couldn't we disable the whole "ensure edit selection" logic while the model is populating?

signedav:
I agree that it's not the nicest approach but I wanted to fix it with a small amount of changes. Exactly because it's so fragile I don't wanted to touch more than needed for this fix that should be backportet.

nyalldawson:
> I agree that it's not the nicest approach but I wanted to fix it with a small amount of changes. Exactly because it's so fragile I don't wanted to touch more than needed for this fix that should be backportet.

I get that, but in this case I don't think it's appropriate. I'd rather see the original changes reverted if we're concerned about regressions in the LTR, and then reintroduced with appropriate safeguards in 3.16 only...

gioman:
> The attribute table is already held together by chewing gum

@nyalldawson https://github.com/qgis/QGIS/issues/38018#issuecomment-665271554 isn't really time to start a wide discussion about this matter and how to solve it?

nyalldawson:
@gioman 

> isn't really time to start a wide discussion about this matter and how to solve it?

yes, I think we do need to start thinking seriously about rebuilding it... perhaps kick off a discussion on the ml?

gioman:
> perhaps kick off a discussion on the ml?

@nyalldawson I agree and I can do that... but "maybe" you are a more authoritative person to do so :)

nyalldawson:
This can be closed, right?

signedav:
Yes. Thanks.

PR #38115, opened 2020-08-03
[postgres] Fix superslow layer instantiation with huge databases (fixes #38114)

### LINKED ISSUE #38114: Extremely slow loading postgis layers from huge databases
With the postgres provider, the layer load performance depends heavily on number of tables in database, because QgsPostgresProvider::loadFields() fetches the whole pg_type table.

In my case of a database with hundreds of thousands tables, with poor connection, instantiation of one layer may take over one minute, what makes big projects opening **over one hour** (!).

By filtering only due records of pg_type, that time is lowered into seconds.

borysiasty:
If a postgresql database contains millions of columns in tables, instantiating any layer becomes madly slow, because QgsPostgresProvider::loadFields fetches type info of all those columns, instead of just those present in source table or query. See bug #38114.

This PR adds a filter to only fetch used columns, speeding up project openin in my case from over one hour (!) to a couple of seconds.

For best readability, I split this PR into three commits:

1) Just two code blocks swapped in order to make the third commit more readable.

2) As typeMap is generally referenced with PQftype, remove an additional reference to it, so we're sure the only keys of that map come from PQftype. AFAIK both PQftype and that old complicated solution finally lead to  pg_type.oid. @nyalldawson, you implemented that solution in #fdd68963c71714. Does the simplification look safe for you? Or was it intentional?

3) The actual fix: collect all PQftypes that will be later used (see point 2) and only fetch "their" pg_type.

### DIFF
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -817,7 +817,6 @@ bool QgsPostgresProvider::loadFields()
   QgsPostgresResult result( connectionRO()->PQexec( sql ) );
 
   QMap<Oid, QMap<int, QString> > fmtFieldTypeMap, descrMap, defValMap, identityMap, generatedMap;
-  QMap<Oid, QMap<int, Oid> > attTypeIdMap;
   QMap<Oid, QMap<int, bool> > notNullMap, uniqueMap;
   if ( result.PQnfields() > 0 )
   {
@@ -865,7 +864,6 @@ bool QgsPostgresProvider::loadFields()
         QString formatType = fmtFieldTypeResult.PQgetvalue( i, 2 );
         QString descr = fmtFieldTypeResult.PQgetvalue( i, 3 );
         QString defVal = fmtFieldTypeResult.PQgetvalue( i, 4 );
-        Oid attType = fmtFieldTypeResult.PQgetvalue( i, 5 ).toUInt();
         bool attNotNull = fmtFieldTypeResult.PQgetvalue( i, 6 ).toInt();
         bool uniqueConstraint = fmtFieldTypeResult.PQgetvalue( i, 7 ).toInt();
         QString attIdentity = connectionRO()->pgVersion() >= 100000 ? fmtFieldTypeResult.PQgetvalue( i, 8 ) : " ";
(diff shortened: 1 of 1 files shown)


nyalldawson:
>  @nyalldawson, you implemented that solution in #fdd68963c71714. Does the simplification look safe for you? Or was it intentional?

Sorry, I honestly can't recall -- it's going back too far. But it seems there's a bunch of tests I added for this which are failing now, so if you get them passing them I'm happy with the new approach!

> If a postgresql database contains millions of columns in tables

I initially misread this as "millions of columns in a table", and wondered WTF you were doing!! :stuck_out_tongue_closed_eyes:

borysiasty:
@nyalldawson wrote:
> Sorry, I honestly can't recall -- it's going back too far. But it seems there's a bunch of tests I added for this which are failing now, so if you get them passing them I'm happy with the new approach!

Ah, there they hide! I only checked for lowercase ctest -R "postgres" :)

> I initially misread this as "millions of columns in a table", and wondered WTF you were doing!! ￼

Just backing up the Internet in a single table, to have it at hand after the apocalypse ;)

gioman:
> If a postgresql database contains millions of columns in tables, instantiating any layer becomes madly slow

@borysiasty interesting. Do you have an idea about the number of columns starting from this slowdown starts to be noticeable?
I have relatively small database that has a few tens of tables each one having quite a lot of columns, and I always felt that using it QGIS was noticeably slow compared to others with tables with much less columns.

borysiasty:
@gioman At the moment I'm looking at a db with about 25k tables x 10 columns. It gives over 100k records in pg_type (and over 1M in pg_attribute). When opening a project with 50 layers *on localhost*, I'm able to read name of every layer being instantiated, so it takes a second, maybe half. With this fix, all layers just blink within a second. The same speedup I notice when using ogr provider instead of the native one.

Just replicate any schema again and again until it becomes noticeable :)

borysiasty:
@gioman a small correction: it's number of tables rather than columns. A huge number of records pg_type is unique per table, not column. Ordinary columns usually share a few simple data types (thus records in pg_type).

gioman:
> @gioman a small correction: it's number of tables rather than columns.

@borysiasty thanks, so this is not the source of the slowness I'm seeing in my use case.

borysiasty:
Superseded by #38158

PR #38116, opened 2020-08-03
Fix some leaks

nyalldawson:
Some relatively harmless memory leak fixes


src/core/qgsapplication.cpp
```diff
@@ -2427,6 +2427,7 @@ QgsApplication::ApplicationMembers::~ApplicationMembers()
   delete mRendererRegistry;
   delete mSvgCache;
   delete mImageCache;
+  delete mSourceCache;
```
nirvn, line 2430: :bow:

PR #38117, opened 2020-08-03
Fix leak of network replies when network content is fetched in a thread which is destroyed soon after

nyalldawson:
Because the thread is destroyed, the event loops never executes and
deleteLater is never called, resulting the a leak of the network reply

(Moderately risky!)


nyalldawson:
Merging for wider testing, will keep an eye out for regressions...

PR #38144, opened 2020-08-05
Treat mdb files as a database source in browser

nyalldawson:
Allows these to be expanded so that particular layers can be loaded, notably allowing projects with broken mdb layer paths to be repaired by allowing users to pick the correct target layer from an mdb file.


Saijin-Naib:
And boom, ANOTHER huge barrier to adoption goes bye-bye.

Thank you, Nyall!

PR #38151, opened 2020-08-05
Add a iface method to directly set a specific GPS connection to connect to in the GPS Information panel

nyalldawson:
Allows for plugins to create subclasses of QgsGpsConnection and then hook these into the existing GPS panel widget

### DIFF
--- a/python/gui/auto_generated/qgisinterface.sip.in
+++ b/python/gui/auto_generated/qgisinterface.sip.in
@@ -1397,6 +1397,16 @@ creating a new browser model for every use.
 .. versionadded:: 3.4
 %End
 
+    virtual void setGpsPanelConnection( QgsGpsConnection *connection ) = 0;
+%Docstring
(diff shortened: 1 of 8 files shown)


m-kuhn:
Adding mock location?

nyalldawson:
Yes - I have a need to "replay" a prerecorded log of nmea commands!

m-kuhn:
sounds to me like this should be an api in core (not gui), but I don't know the whole gps code

nyalldawson:
There's already a registry where general connections can be handled in core - what's missing is a way to set a specific connection to be used by the GPS panel (which lives in app). That's what this does -- allows you to programmatically set the connection vs having to make a user pick it.

nyalldawson:
In general though I agree - there's a ton of useful stuff tied up specifically to the GPS panel widget which would be great to extract and move to reusable methods!

m-kuhn (approved):
Looks good to me

PR #38152, opened 2020-08-05
Fix long reload of filter change in attribute table by revert bad commit

signedav:
Revert of commit d409c65de9c31266ae00a16d1004b4856dc1f5b4 leading to long reload time of the attribute table after `setFilterMode`.

This bad commit has been a fix on the minor bug, that the counter on form view mode of the attribute table has not been updated after adding or removing a feature that did not affect the edit selection.

This PR is an alternative to the ugly fix with timer here #38111

I suggest to merge that to have #38018 fixed, and backport it to 3.10 and 3.14.
And then we go for a new and proper fix for the counter reset for the master with calling the ensure-edit-selection part after the model has populated (instead of on every add-/remove row).

### DIFF
--- a/src/gui/attributetable/qgsfeaturelistview.cpp
+++ b/src/gui/attributetable/qgsfeaturelistview.cpp
@@ -162,11 +162,7 @@ void QgsFeatureListView::editSelectionChanged( const QItemSelection &deselected,
     QItemSelection localSelected = mModel->mapSelectionFromMaster( selected );
     viewport()->update( visualRegionForSelection( localDeselected ) | visualRegionForSelection( localSelected ) );
   }
-  updateEditSelectionDependencies();
-}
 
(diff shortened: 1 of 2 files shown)


signedav:
What do you think @nyalldawson?

nyalldawson:
I think this is the right move. @m-kuhn do you agree?

@gioman can you confirm that this reversion fixes the issue?

m-kuhn (approved):

m-kuhn:
I agree, sounds safe!

signedav:
I don't know why it's failing on the MS tests qgis.QGIS. When I login over the link displayed in the failed publishing tests, it tells me I don't have access... 
Can I do anything, does it need a restart or is it fine to merge anyway?

nyalldawson:
It's unrelated -- don't worry about it !

PR #38155, opened 2020-08-05
Use full canonical path to installed plugins

nyalldawson:
Resolves symlinks which may be used for the plugin folder, e.g. by
plugin developers who symlink a development folder into their profile's
plugin folder path.

Makes it a bit easier for plugin developers to see whether a plugin
is pointing to a local development path or a plugin installed through
the repo.


Gustry:
Thanks, I guess it's going to be useful !

It would be nice if the plugin manager could detect if a plugin is under a git config and doesn't update a plugin by default if a upgrade is available. 
Who never deleted a plugin (with its symlink) when it has a git config by ugrapding it from QGIS ? :)

nyalldawson:
> Who never deleted a plugin (with its symlink) when it has a git config by ugrapding it from QGIS ? :)



uh ... me :blush:

PR #38158, opened 2020-08-05
[postgres] Fix superslow layer instantiation in huge databases by only fetching type info of used fields.

borysiasty:
If a postgresql database contains hundreds of thousands tables, instantiating any layer becomes madly slow, because QgsPostgresProvider::loadFields fetches type info for all those tables, instead of just columns of source table(s). See bug #38114.

This PR adds a filter to only fetch used columns, speeding up project opening in my case from over one hour (!) to a couple of seconds.

Supersedes #38115.

### DIFF
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -789,12 +789,15 @@ bool QgsPostgresProvider::loadFields()
   // Clear cached information about enum values support
   mShared->clearSupportsEnumValuesCache();
 
+  QString sql;
+  QString attroidsFilter;
+
   if ( !mIsQuery )
   {
     QgsDebugMsgLevel( QStringLiteral( "Loading fields for table %1" ).arg( mTableName ), 2 );
 
     // Get the relation oid for use in later queries
-    QString sql = QStringLiteral( "SELECT regclass(%1)::oid" ).arg( quotedValue( mQuery ) );
+    sql = QStringLiteral( "SELECT regclass(%1)::oid" ).arg( quotedValue( mQuery ) );
     QgsPostgresResult tresult( connectionRO()->PQexec( sql ) );
     QString tableoid = tresult.PQgetvalue( 0, 0 );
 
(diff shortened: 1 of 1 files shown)


src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -900,10 +894,42 @@ bool QgsPostgresProvider::loadFields()
...
+    // Prepare filter for fetching pg_type info
+    if ( !attroids.isEmpty() )
+    {
+      QStringList attroidsList;
+      const auto constAttroids = attroids;
+      for ( Oid attroid : constAttroids )
+      {
+        attroidsList.append( QString::number( attroid ) );
       }
+      attroidsFilter = QStringLiteral( "WHERE oid in (" ) + attroidsList.join( QStringLiteral( "," ) ) + QStringLiteral( ")" );
```
nyalldawson, line 912:
```suggestion
      attroidsFilter = QStringLiteral( "WHERE oid in (%1)" ).arg( attroidsList.join( ',' ) );
```

nyalldawson (approved):

nyalldawson:
Looks great, thanks!

src/providers/postgres/qgspostgresprovider.cpp
borysiasty: Ah, thanks! It was awful, sorry for that :)

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -900,10 +894,42 @@ bool QgsPostgresProvider::loadFields()
...
+        // Also include atttype oid from pg_attribute, because PQnfields only returns basic type for for domains
+        attroids.insert( attType );
+      }
+    }
+
+    // Prepare filter for fetching pg_type info
+    if ( !attroids.isEmpty() )
+    {
+      QStringList attroidsList;
+      const auto constAttroids = attroids;
```
m-kuhn, line 907:
```suggestion
```

m-kuhn, line 907:
```suggestion
```

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -900,10 +894,42 @@ bool QgsPostgresProvider::loadFields()
...
+      for ( Oid attroid : constAttroids )
```
m-kuhn, line 908:
```suggestion
      for ( Oid attroid : qgis::as_const( attroids ) )
```

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -789,12 +789,15 @@ bool QgsPostgresProvider::loadFields()
...
+  QString sql;
+  QString attroidsFilter;
+
   if ( !mIsQuery )
   {
     QgsDebugMsgLevel( QStringLiteral( "Loading fields for table %1" ).arg( mTableName ), 2 );
 
     // Get the relation oid for use in later queries
-    QString sql = QStringLiteral( "SELECT regclass(%1)::oid" ).arg( quotedValue( mQuery ) );
+    sql = QStringLiteral( "SELECT regclass(%1)::oid" ).arg( quotedValue( mQuery ) );
```
m-kuhn, line 800: Unrelated: for performance reasons, could we do

```sql
QString tableoid = QStringLiteral( "regclass(%1)::oid" ).arg( quotedValue( mQuery ) );
```

and completely remove one query (roundtrip)?

m-kuhn (approved):

src/providers/postgres/qgspostgresprovider.cpp
borysiasty: it's duplicated :)

src/providers/postgres/qgspostgresprovider.cpp
borysiasty: Ok, squeezed into one query (although initially I wanted to change as less as possible, to avoid any unnecessary backport risk ;) )

borysiasty:
The failed Azure build doesn't look related.

PR #38174, opened 2020-08-06
Workaround GDAL reporting that shapefiles support DateTime

### LINKED ISSUE #38011: Field calculator should not show the "date & time" datatype for shapefiles
**How to Reproduce**

1. Download for instance: https://firms.modaps.eosdis.nasa.gov/data/active_fire/c6/shapes/zips/MODIS_C6_Europe_24h.zip
2. Unzip and load in QGIS
3. This shapefile has an ACQ_DATE (Date) and an ACQ_TIME (String) fields.
4. To convert them to a single Datetime field, in Field Calculator:
Output field name: ACQ_DTTM
Output field type: Date & Time
Expression: to_datetime("ACQ_DATE" || ' ' || left("ACQ_TIME",2) || ':' || right("ACQ_TIME",2))

Preview is correct:
imagem

New field in Attribute Table also correct:
imagem

And also in Fields Properties:
imagem

4. However, when I Save Edits, it changes the new field type to Date, loosing time information:
imagem

imagem

5. If I export the Shapefile to GeoPackage and add the new field after it, everything works fine. So, this seems to only affect Shapefiles.

**QGIS and OS versions**

Tested on QGIS 3.14.1 and QGIS 3.10.8, both on Windows 10.

dminor:
This was fixed in OSGeo/gdal#2821 but we probably want to add a workaround in QGIS until we update to a version of GDAL that includes that fix.

Fixes #38011

### DIFF
--- a/src/core/providers/ogr/qgsogrprovider.cpp
+++ b/src/core/providers/ogr/qgsogrprovider.cpp
@@ -564,6 +564,15 @@ QgsOgrProvider::QgsOgrProvider( QString const &uri, const ProviderOptions &optio
     CSLDestroy( papszTokens );
   }
 
+  // Older versions of GDAL incorrectly report that shapefiles support
+  // DateTime.
+#if GDAL_VERSION_NUM <= GDAL_COMPUTE_VERSION(3,1,2)
+  if ( mGDALDriverName == QLatin1String( "ESRI Shapefile" ) )
+  {
+    supportsDateTime = false;
+  }
+#endif
+
   if ( supportsDate )
   {
     nativeTypes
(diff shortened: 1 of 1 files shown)


src/core/providers/ogr/qgsogrprovider.cpp
```diff
@@ -564,6 +564,15 @@ QgsOgrProvider::QgsOgrProvider( QString const &uri, const ProviderOptions &optio
     CSLDestroy( papszTokens );
   }
 
+  // Older versions of GDAL incorrectly report that shapefiles support
+  // DateTime.
+#if GDAL_VERSION_NUM <= GDAL_COMPUTE_VERSION(3,1,2)
```
rouault, line 569: The GDAL fix will appear in 3.2. I'm a bit hesitant to backport it in the 3.1 series, so :
```suggestion
#if GDAL_VERSION_NUM < GDAL_COMPUTE_VERSION(3,2,0)
```

PR #38187, opened 2020-08-07
Extend QgsSqlStatement to allow parsing of sql fragments and handle Microsoft SQL variants

nyalldawson:
Allows the parser to handle where or expression fragments, and also parse the Microsoft SQL variants too...


nyalldawson:
@rouault what's your thoughts here?

PR #38189, opened 2020-08-07
Server landingpage

elpaso:
# QGIS Server Landing Page Plugin

C++ port of https://github.com/elpaso/qgis-server-landing-page-plugin

qgis-browser-c++


nyalldawson:
This is great!!!

I looked briefly over the code in a general way and it looks good to me, but I can't comment on the server specifics.

elpaso:
I think we can merge, this is an independent module with no impact on existing functionalities. 

@jef-n for the packaging, this would be another server package, that depends on `wms` and `wfs3`.

dmarteau:
Hi,

Sorry to be late on the conversation but I stumbled on some issues using landing pages when behind a proxy:

* setting /ows/catalog as landing page you may reach the uri /ows/catalog/index.html but not `/catalog/` as it return an invalid unterminated response (got timeout error from proxy) .

* If you attempt to get the metadata of a layer you get an uri like /project/<some_id>/..... : this makes the link unusuable in the context of a proxified url.

It seems that the landing_page api does not consumme urls like other services: in `wfs3` service the  prefix uri is returned correctly in the handled responses: if I give `/whatever/wfs3/.....` , I will get correctly formatted urls in the response. But with the landing_page api, declaring `/catalog` as prefix and fetching` `/whatever/catalog/...` will return a 404.

PR #38196, opened 2020-08-08
[FEATURE][processing] Export layout atlas as PDF algorithm

nirvn:
## Description

PDF counterpart to the recent export atlas layout to image algorithm.


nirvn:
Azure failure unrelated.

src/analysis/processing/qgsalgorithmlayoutatlastopdf.cpp
```diff
@@ -0,0 +1,288 @@
...
+QString QgsLayoutAtlasToPdfAlgorithm::shortDescription() const
+{
+  return QObject::tr( "Exports an atlas layout as a PDF." );
+}
+
+QString QgsLayoutAtlasToPdfAlgorithm::shortHelpString() const
+{
+  return QObject::tr( "This algorithm outputs an atlas layout as a PDF file.\n\n"
+                      "If a coverage layer is set, the selected layout's atlas settings exposed in this algorithm "
+                      "will be overwritten. In this case, an empty filter or sort by expressing will turn those "
```
nyalldawson, line 64:
```suggestion
                      "will be overwritten. In this case, an empty filter or sort by expression will turn those "
```

nyalldawson (commented):
There's a LOT of duplicate code here between this algorithm and the existing atlas to image algorithm -- could that code be moved to a common base class or utility function instead?

nirvn:
@nyalldawson , test failures unrelated, OK to merge?

nyalldawson (approved):

PR #38199, opened 2020-08-08
Make the map layer type and validity a property

m-kuhn:
makes map layer type and layer validity available from QML.
Also adds a notification signal about layer validity change.


nirvn (approved):

m-kuhn:
@nirvn are you okay with the latest update as well?

nirvn (commented):
Code looks good.

PR #38202, opened 2020-08-08
Setup Qt 5.14 github workflow

m-kuhn:
## Description

Sets up a new workflow to build QGIS with Qt 5.14 (Fedora 32) in a github workflow ci.

Pros:

 - We are building about a recent Qt version
 - We are building against recent other libraries
 - We build with recent utilities that should perform better with code analysis

TODOs:

 -   ~Fix remaining warnings -- or disable parts~

TODOs (not in scope for this PR)

 - Enable and validate tests
 - Update to Qt 5.15 -- that's the mid term goal, as soon as a stable enough distro (might even be Fedora rawhide) starts shipping it.


m-kuhn:
QGIS Quick had to be disabled. It uses deprecated functionality. If updating it to the code we are using over at [QField](https://github.com/opengisch/QField/pull/702) the tests start to crash. The first assumption from top of my head would be this is triggered by no window created before quickutils are created. While I cleaned up some other things, I left this as an exercise for someone else to fix, since it's not exactly clear to me how it's supposed to work. (@PeterPetrik @vcloarec I guess one of you might be interested)

external/mdal/frmts/mdal_selafin.cpp
```diff
@@ -399,9 +399,9 @@ void MDAL::SelafinFile::populateDataset( MDAL::Mesh *mesh, std::shared_ptr<MDAL:
   }
 
   // now calculate statistics
-  for ( const std::shared_ptr<DatasetGroup> group : groupsInOrder )
+  for ( const std::shared_ptr<DatasetGroup> &group : groupsInOrder )
```
nyalldawson, line 402: ping @vcloarec

nyalldawson:
Great work @m-kuhn ! All looks good to me!

external/mdal/frmts/mdal_selafin.cpp
m-kuhn: what's the ping for? something to worry?

nirvn: To make sure it's applied upstream?

m-kuhn: Ah, good point. Shouldn't block then?

nirvn: It likely should. I'm sure that'll be applied upstream prior to the next mdal update in QGIS' source tree. In the meantime, good to have it fixed here.

nyalldawson: Yep, just warning mdal guys to port upstream.

m-kuhn: Cool, I'll merge then, this will also ensure that any future mdal update comes with this patch applied.

PeterPetrik: ported https://github.com/lutraconsulting/MDAL/pull/289 👍 

but don't be shy, your direct contributions to MDAL are welcomed too!

PeterPetrik:
> QGIS Quick had to be disabled. It uses deprecated functionality. If updating it to the code we are using over at [QField](https://github.com/opengisch/QField/pull/702) the tests start to crash. The first assumption from top of my head would be this is triggered by no window created before quickutils are created. While I cleaned up some other things, I left this as an exercise for someone else to fix, since it's not exactly clear to me how it's supposed to work. (@PeterPetrik @vcloarec I guess one of you might be interested)

Not sure if I understand, we use Qt 5.14.2 for some time already for releasing without problems?

external/mdal/frmts/mdal_selafin.cpp
m-kuhn: > ported [lutraconsulting/MDAL#289](https://github.com/lutraconsulting/MDAL/pull/289) +1
> 
> but don't be shy, your direct contributions to MDAL are welcomed too!

good call. in retrospective a "missed opportunity". I'll try to find something else to PR.

m-kuhn:
It's a deprecated function on Qt side, used in the QgisQuick implementation, see https://github.com/qgis/QGIS/runs/963296308#step:4:3334

PeterPetrik:
@m-kuhn @3nids that was PR for that https://github.com/qgis/QGIS/pull/32708

not sure why it was not merged?

m-kuhn:
also not sure.
i tried again within this pull request here, but as mentioned, it caused crashes on travis.

3nids:
Tests were failing on this pr. If it turns out to be green now, perfect.

PeterPetrik:
I am not clear how it is supposed to work either ...

PR #38207, opened 2020-08-08
Fix QGIS server getPrint when highlight is requested over a map following a theme

### LINKED ISSUE #34178: QGIS Server GetPrint: HIGHLIGHT_GEOM is not printed if map layers are configured to follow a map theme

jgrocha:
## Description

Fix #34178

To be able to print with a highlight geometry and a map following a theme, this PR creates a temporary theme.

The temporary theme is removed by `QgsWmsRestorer`. I've added a very simple `QgsProjectRestorer` class to QgsWmsRestorer. This class can be expanded later to make sure other properties of the project are preserved, if they are changed during a specific request.

This PR replaces the former #36007.

This fix was only possible with @pblottiere guidance.

### DIFF
--- a/src/server/services/wms/qgswmsrenderer.cpp
+++ b/src/server/services/wms/qgswmsrenderer.cpp
@@ -622,31 +622,32 @@ namespace QgsWms
             renderLayers = c->project()->mapThemeCollection()->masterVisibleLayers();
 
           // get highlight layers, if any
-          bool has_highlight_layer = false;
-          for ( auto layer : mapSettings.layers() )
+          bool hasHighlightLayer = false;
+          const QList<QgsMapLayer *> layers = mapSettings.layers();
+          for ( QgsMapLayer *layer : layers )
           {
 
             if ( RE_HIGHLIGHT_LAYER.match( layer->name() ).hasMatch() )
             {
-              has_highlight_layer = true;
+              hasHighlightLayer = true;
               auxHighlightLayers << layer;
               c->project()->layerTreeRoot()->insertLayer( 0, layer );
             }
 
           }
 
-          if ( has_highlight_layer )
+          if ( hasHighlightLayer )
           {
 
             // create a new theme with all layers
             QgsMapThemeCollection::MapThemeRecord rec;
             // layers from the theme
-            for ( auto layer : renderLayers )
+            for ( QgsMapLayer *layer : qgis::as_const( renderLayers ) )
             {
               rec.addLayerRecord( QgsMapThemeCollection::MapThemeLayerRecord( layer ) );
             }
             // highlight layers
-            for ( auto layer : auxHighlightLayers )
+            for ( QgsMapLayer *layer : qgis::as_const( auxHighlightLayers ) )
             {
               rec.addLayerRecord( QgsMapThemeCollection::MapThemeLayerRecord( layer ) );
             }
--- a/src/server/services/wms/qgswmsrenderer.cpp
+++ b/src/server/services/wms/qgswmsrenderer.cpp
@@ -64,6 +64,7 @@
 #include "qgsserverexception.h"
 #include "qgsexpressioncontextutils.h"
 #include "qgsfeaturestore.h"
+#include "qgsmapthemecollection.h"
 
 #include <QImage>
 #include <QPainter>
@@ -602,8 +603,62 @@ namespace QgsWms
         }
       }
 
+      QList<QgsMapLayer *> renderLayers, auxHighlightLayers;
+
       if ( !map->keepLayerSet() )
       {
+
+        static const QRegularExpression RE_HIGHLIGHT_LAYER = QRegularExpression( "^highlight_\\d+$" );
+
+        QString presetName = map->followVisibilityPresetName();
+
+        QString newPresetName;
+        if ( !presetName.isEmpty() )
+        {
+          // get layers from theme
+          if ( c->project()->mapThemeCollection()->hasMapTheme( presetName ) )
+            renderLayers = c->project()->mapThemeCollection()->mapThemeVisibleLayers( presetName );
+          else
+            renderLayers = c->project()->mapThemeCollection()->masterVisibleLayers();
+
+          // get highlight layers, if any
+          bool has_highlight_layer = false;
+          for ( auto layer : mapSettings.layers() )
+          {
+
+            if ( RE_HIGHLIGHT_LAYER.match( layer->name() ).hasMatch() )
+            {
+              has_highlight_layer = true;
+              auxHighlightLayers << layer;
+              c->project()->layerTreeRoot()->insertLayer( 0, layer );
+            }
+
+          }
+
+          if ( has_highlight_layer )
+          {
+
+            // create a new theme with all layers
+            QgsMapThemeCollection::MapThemeRecord rec;
+            // layers from the theme
+            for ( auto layer : renderLayers )
+            {
+              rec.addLayerRecord( QgsMapThemeCollection::MapThemeLayerRecord( layer ) );
+            }
+            // highlight layers
+            for ( auto layer : auxHighlightLayers )
+            {
+              rec.addLayerRecord( QgsMapThemeCollection::MapThemeLayerRecord( layer ) );
+            }
+
+            // Add the theme and set it
+            newPresetName = presetName + QStringLiteral( "__highlight" );
+            c->project()->mapThemeCollection()->insert( newPresetName, rec );
+            map->setFollowVisibilityPresetName( newPresetName );
+          }
+
+        }
+
         if ( cMapParams.mLayers.isEmpty() && cMapParams.mExternalLayers.isEmpty() )
         {
           map->setLayers( mapSettings.layers() );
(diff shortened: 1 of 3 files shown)


src/server/services/wms/qgswmsrenderer.cpp
```diff
@@ -602,8 +603,62 @@ namespace QgsWms
...
+        if ( !presetName.isEmpty() )
+        {
+          // get layers from theme
+          if ( c->project()->mapThemeCollection()->hasMapTheme( presetName ) )
+            renderLayers = c->project()->mapThemeCollection()->mapThemeVisibleLayers( presetName );
+          else
+            renderLayers = c->project()->mapThemeCollection()->masterVisibleLayers();
+
+          // get highlight layers, if any
+          bool has_highlight_layer = false;
```
m-kuhn, line 625:
```suggestion
          bool hasHighlightLayer = false;
```

Coding convention is camel case

src/server/services/wms/qgswmsrenderer.cpp
```diff
@@ -602,8 +603,62 @@ namespace QgsWms
...
+          for ( auto layer : mapSettings.layers() )
```
m-kuhn, line 626:
```suggestion
          const QList<QgsMapLayer *> layers = mapSettings.layers();
          for ( QgsMapLayer *layer : layers )
```

NEVER iterate with a for loop directly on a list retrieved through a method (it's a silly thing of Qt ... ).
Also be explicit about return values (unless they are incredibly hard to write).

src/server/services/wms/qgswmsrenderer.cpp
```diff
@@ -602,8 +603,62 @@ namespace QgsWms
...
+
+          }
+
+          if ( has_highlight_layer )
+          {
+
+            // create a new theme with all layers
+            QgsMapThemeCollection::MapThemeRecord rec;
+            // layers from the theme
+            for ( auto layer : renderLayers )
```
m-kuhn, line 644:
```suggestion
            for ( auto layer : qgis::as_const( renderLayers ) )
```

When iterating containers, make sure they are const

src/server/services/wms/qgswmsrenderer.cpp
```diff
@@ -602,8 +603,62 @@ namespace QgsWms
...
+            {
+              rec.addLayerRecord( QgsMapThemeCollection::MapThemeLayerRecord( layer ) );
+            }
+            // highlight layers
+            for ( auto layer : auxHighlightLayers )
```
m-kuhn, line 649: Same. `const` -- and for extra points explicit type

src/server/services/wms/qgswmsrestorer.cpp
```diff
@@ -138,10 +140,33 @@ QgsLayerRestorer::~QgsLayerRestorer()
   }
 }
 
+QgsProjectRestorer::QgsProjectRestorer( const QgsProject *project )
```
m-kuhn, line 143:
```suggestion
QgsProjectRestorer::QgsProjectRestorer( QgsProject *project )
```

A project restorer says in its name that it will modify the project. So why suggest it would be const? :-)

src/server/services/wms/qgswmsrestorer.cpp
```diff
@@ -138,10 +140,33 @@ QgsLayerRestorer::~QgsLayerRestorer()
...
+{
```
m-kuhn, line 144: : mThemeCollection( project->mapThemeCollection() )

src/server/services/wms/qgswmsrestorer.cpp
```diff
@@ -138,10 +140,33 @@ QgsLayerRestorer::~QgsLayerRestorer()
...
+  mThemeCollection = const_cast<QgsProject *>( project )->mapThemeCollection( );
```
m-kuhn, line 145:
```suggestion
```

src/server/services/wms/qgswmsrestorer.h
```diff
@@ -63,6 +63,35 @@ class QgsLayerRestorer
     };
 
     QMap<QgsMapLayer *, QgsLayerSettings> mLayerSettings;
+
+};
+
+/**
+ * \ingroup server
+ * RAII class to restore project configuration on destruction
+ * \since QGIS 3.0
```
m-kuhn, line 72: That long already?

src/server/services/wms/qgswmsrestorer.h
```diff
@@ -63,6 +63,35 @@ class QgsLayerRestorer
...
+ */
+class QgsProjectRestorer
+{
+  public:
+
+    /**
+     * Constructor for QgsProjectRestorer.
+     * \param project
```
m-kuhn, line 80:
```suggestion
```

m-kuhn (commented):
Hi @jgrocha thanks for the pull request.
I have reviewed it with respect to coding standards only.

Happy to leave the architectural review and judgement of the "hackyness" level of the approach to someone more familiar with server code.

jgrocha:
> Hi @jgrocha thanks for the pull request.
> I have reviewed it with respect to coding standards only.
> 
> Happy to leave the architectural review and judgement of the "hackyness" level of the approach to someone more familiar with server code.

Thank you @m-kuhn for your review.

m-kuhn:
Maybe still a question on the architectural level. Would it be possible to provide extra information about highlighting through the mapSettings instead of temporarily modifying the project? I can see that the `XyzRestorer` approach has been done with other parts too, but wouldn't it be cleaner to just leave all this in a pristine state and provide extra informations through context?

jgrocha:
> Maybe still a question on the architectural level. Would it be possible to provide extra information about highlighting through the mapSettings instead of temporarily modifying the project? I can see that the `XyzRestorer` approach has been done with other parts too, but wouldn't it be cleaner to just leave all this in a pristine state and provide extra informations through context?

If we add something like `void setHighlightGeometry( const QgsGeometry &highlight )` to `QgsMapSettings`, this have to be also managed on the desktop side. We have to check if a `highlightGeometry` is present, and add it to the output. The highlight would be always empty on the desktop side, but it would appear in the code, as another `QgsMapSettings` property.

In my initial PR #34271 (check the [discussion](https://github.com/qgis/QGIS/pull/34271#discussion_r374860739)), both @nyalldawson and @pblottiere argued that this should be handled only on the server side. What I did was really to make it possible, without changing anything on the desktop side.

nyalldawson:
I'm ok with this being handled in core if it's done in the correct way, and I think adding explicit, non-server specific api to QgsMapSettings would be acceptable.

m-kuhn:
If from server-maintainer side there is no objections, happy to follow their lead.

Otherwise it would deserve a discussion of the api first (a simple `setHighlightGeometry( QgsGeometry )` seems very narrow for a) its support for exactly one geometry and b) no explicit styling information, happy to discuss options)

nyalldawson:
> Otherwise it would deserve a discussion of the api first (a simple setHighlightGeometry( QgsGeometry ) seems very narrow for a) its support for exactly one geometry and b) no explicit styling information, happy to discuss options)

Or just use the new QgsAnnotationLayer support directly? It's designed with use cases like this in mind...

nyalldawson:
That said, I think the discussion is becoming circular -- if QgsAnnotationLayer was used you'd still need to inject it into the theme first

m-kuhn:
> That said, I think the discussion is becoming circular -- if QgsAnnotationLayer was used you'd still need to inject it into the theme first

How about a (couple of) "overlay layers" provided through the context? Or a callback?

m-kuhn:
@pblottiere @elpaso @rldhont what do you think about the approach taken in here? I think the server should aim towards using an "immutable project", especially since the project is cached between different requests.
I'm almost sure, that we will not be able to always completely undo the changes in the restorer.

nyalldawson:
@m-kuhn 

An alternative approach would be to have a QgsProject::clone() method. Then a clone of the project would be taken prior to safely messing with the layers and print layouts.

(This would be useful elsewhere too -- if we cloned a whole project, we could then safely export layouts in a background thread...)

m-kuhn:
Maybe. But might add some tricky questions -- clone layers as well? Does that have an impact on the caching and therefore defeat the purpose.

nyalldawson:
Yeah, the intention would be to create an entire new copy of the project and ALL its contents. For it to work well we'd need to avoid it becoming just a "save to XML and restore on a new instance" thing, and instead shortcut all the work from constructors by just copying calculate properties directly...

m-kuhn:
That will involve writing `clone()` methods down to all the providers.
I guess the "add annotation layer through context" will be easier to implement.

pblottiere:
Hi @m-kuhn and @nyalldawson,

Sorry for being late in the discussion, we have been victim of a home invasion by a 5 days old boy, so I tried to stay AFK for some days ;).


> That will involve writing clone() methods down to all the providers. Doesn't sound super trivial, but would be useful here and there.

Indeed a `clone()` method sounds useful but I agree with Mathias that it doesn't seem trivial.

Moreover, if we use it at each incoming request on server side, it would become a very important citizen in terms of performance. So we should really be extra careful.

> Still I guess the "add annotation layer through context" will be easier to implement.

+1 to dig in this direction then (sorry for the return trip @jgrocha)

jgrocha:
Thank you all for the valuable feedback. I think we have found a path to move forward.

PR #38210, opened 2020-08-09
Two typo's corrected

DiGro:
Line  57  :  "as a set of image."  should probably be "as a set of images."  Plural, not singular
Line  64  :  "sort by expressing"  should probably be "sort by expression"

## Description

Two typo's, see above, corrected in order to display correct documentation

Not sure if this should be backported.

------------------------


m-kuhn:
Thanks.
Nothing to backport, this is master only.

m-kuhn (approved):

PR #38215, opened 2020-08-10
Add QgsAnnotationPointTextItem annotation item class for rendering text at a specified map point

nyalldawson:
Adds the ability to insert formatted text at a specific map point in an annotation layer


tests/src/python/test_qgsannotationpointtextitem.py
```diff
@@ -0,0 +1,207 @@
...
+        image.fill(QColor(255, 255, 255))
+        painter = QPainter(image)
+        rc.setPainter(painter)
+
+        try:
+            item.render(rc, None)
+        finally:
+            painter.end()
+
+        self.assertTrue(self.imageCheck('pointtext_item', 'pointtext_item', image))
```
nirvn, line 154: Looking at the masking, there seems to be quite a difference in spacing (a difference that diminishes as the string grows, i.e. not constant). What's that all about?

nirvn (approved):
Looking good.

tests/src/python/test_qgsannotationpointtextitem.py
nyalldawson: Just more cross-platform font rendering differences. I built the test images on Fedora, which applies some different metrics for whatever reason...

Issue #12827, opened 2010-06-01
Load/Save queries in the Query builder


qgib:
Author Name: **pinux -** (pinux -)
Original Redmine Issue: [2767](https://issues.qgis.org/issues/2767)

Redmine category:data_provider
Assignee: nobody -

---
Load/Save queries in the Query builder as it is possible for the Search query builder.

PR #38218, opened 2020-08-10
restore missed "save query" functionality in the Query builder

### LINKED ISSUE #18550: Save query no more possible
Original Redmine Issue: [10082](https://issues.qgis.org/issues/10082)

Redmine category:vectors

---
On previous QGIS releases was possible to save table queries (when using the advanced query from table of attributes) in a .qqf file.

It would be good to have back such functionality.

---

Related issue(s): #18598 (duplicates)
Redmine related issue(s): [10139](https://issues.qgis.org/issues/10139)

---

alexbruy:
## Description

In previous QGIS versions it was possible to save queries used in the Search Query Builder in the *.qqf files, but at some point when Search Query Builder was replaced by Query Builder class this functionality was missed. This PR restores functionality for saving/loading queries.

FIxes #18550 and #12827.

### DIFF
--- a/src/gui/qgsquerybuilder.cpp
+++ b/src/gui/qgsquerybuilder.cpp
@@ -74,12 +74,12 @@ QgsQueryBuilder::QgsQueryBuilder( QgsVectorLayer *layer,
 
   pbn = new QPushButton( tr( "&Save…" ) );
   buttonBox->addButton( pbn, QDialogButtonBox::ActionRole );
-  pbn->setToolTip( tr( "Save query to an xml file" ) );
+  pbn->setToolTip( tr( "Save query to QQF file" ) );
   connect( pbn, &QAbstractButton::clicked, this, &QgsQueryBuilder::saveQuery );
 
   pbn = new QPushButton( tr( "&Load…" ) );
   buttonBox->addButton( pbn, QDialogButtonBox::ActionRole );
-  pbn->setToolTip( tr( "Load query from xml file" ) );
+  pbn->setToolTip( tr( "Load query from QQF file" ) );
   connect( pbn, &QAbstractButton::clicked, this, &QgsQueryBuilder::loadQuery );
 
   setupGuiViews();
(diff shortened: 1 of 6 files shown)


alexbruy:
BTW, seems `QgsSearchQueryBuilder` class is not used anymore, maybe we should remove it.

m-kuhn:
> BTW, seems `QgsSearchQueryBuilder` class is not used anymore, maybe we should remove it.

:+1:  should be deprecated and tagged for removal

src/gui/qgsquerybuilder.cpp
```diff
@@ -66,6 +72,16 @@ QgsQueryBuilder::QgsQueryBuilder( QgsVectorLayer *layer,
   buttonBox->addButton( pbn, QDialogButtonBox::ActionRole );
   connect( pbn, &QAbstractButton::clicked, this, &QgsQueryBuilder::clear );
 
+  pbn = new QPushButton( tr( "&Save…" ) );
+  buttonBox->addButton( pbn, QDialogButtonBox::ActionRole );
+  pbn->setToolTip( tr( "Save query to an xml file" ) );
```
m-kuhn, line 77:
```suggestion
  pbn->setToolTip( tr( "Save query to qqf file" ) );
```

src/gui/qgsquerybuilder.cpp
```diff
@@ -66,6 +72,16 @@ QgsQueryBuilder::QgsQueryBuilder( QgsVectorLayer *layer,
...
+  connect( pbn, &QAbstractButton::clicked, this, &QgsQueryBuilder::saveQuery );
+
+  pbn = new QPushButton( tr( "&Load…" ) );
+  buttonBox->addButton( pbn, QDialogButtonBox::ActionRole );
+  pbn->setToolTip( tr( "Load query from xml file" ) );
```
m-kuhn, line 82:
```suggestion
  pbn->setToolTip( tr( "Load query from qqf file" ) );
```

src/gui/qgsquerybuilder.cpp
```diff
@@ -436,3 +452,136 @@ void QgsQueryBuilder::showHelp()
...
+  if ( search.hasParserError() )
+  {
+    QMessageBox::critical( this, tr( "Query Result" ), search.parserErrorString() );
+    return;
+  }
+
+  QString newQueryText = query;
+
+#if 0
+  // TODO: implement with visitor pattern in QgsExpression
```
m-kuhn, line 536: What's the purpose of this section exactly?

src/gui/qgsquerybuilder.cpp
alexbruy: Fixed

src/gui/qgsquerybuilder.cpp
alexbruy: Fixed

src/gui/qgsquerybuilder.cpp
alexbruy: This potentially can be useful for cases when loaded query reference attributes which are not exists in the layer. In such case we may want to try "fix" query by asking user which attributes should be used instead of missed ones and making corresponding replacement.

On other hand we don't do this in other places, so maybe it should be removed.

PR #38222, opened 2020-08-10
Check for dimensions when comparing WMS layer properties

Samweli:
## Description
Currently when a WMS server capabilities has updated layer's dimensions, refreshing the WMS connection item does not update the corresponding layer items dimensions in the browser.

This PR adds a check for WMS dimensions when comparing WMS layer properties. This will cause WMS layer items in the browser to update to when there are new dimensions and user refreshes the WMS connection item.

cc @tomkralidis

### DIFF
--- a/src/providers/wms/qgswmscapabilities.h
+++ b/src/providers/wms/qgswmscapabilities.h
@@ -204,6 +204,15 @@ struct QgsWmsDimensionProperty
 
     return QgsDateTimeRange();
   }
+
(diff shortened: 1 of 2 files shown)


nyalldawson (commented):
Can we get a test for this too please?

Samweli:
> Can we get a test for this too please?

sure

Samweli:
ping @nyalldawson

nyalldawson (approved):

nyalldawson:
Thanks @Samweli !

PR #38225, opened 2020-08-10
New project reading flag to trust layer metadata with server settings

rldhont:
## Description

Trust layer metadata can be defined at the project level. With this new project reading flag, it is possible to force the trust layer metadata.

For the server, the trust layer metadata option can be override with `QGIS_SERVER_TRUST_LAYER_METADATA`  setting environnement .

Funded by [Ifremer](https://wwz.ifremer.fr/).


nyalldawson:
Lgtm!

elpaso (approved):
Maybe just add some more information in the flag documentation about in which cases using the flag is NOT recommended (I guess that it is in case the layer extent may vary during time and the metadata would be not correct anymore).

rldhont:
thanks @elpaso, is the description enough ?

elpaso:
Perfect! Thank you.

haubourg:
@rldhont I don't get what is the difference with the current trust flag, which is already at the project level. Can you explain a bit more the differences from a user point of view?

rldhont:
@haubourg from a QGIS Desktop user point of view, it is changing nothing, except that if the QGIS Server has been configured with `QGIS_SERVER_TRUST_LAYER_METADATA=1` he no longer needs to check the trust option.

rldhont:
@haubourg, the trust flag only skips the extent request to the provider. For you what are the request to skip with this flag ?

rldhont:
To complete the `QGIS_SERVER_TRUST_LAYER_METADATA` setting environnement: propagate it https://github.com/qgis/QGIS/pull/38464

PR #38230, opened 2020-08-10
[wms] When a legend graphic fetch fails, don't show the error in the layer tree, but instead just log it to the message log

nyalldawson:
This error indicates an issue on the server configuration and isn't
anything that a user can resolve. Showing it in the layer tree incorrectly
indicates that there's a problem with the layer rendering, which isn't
really true. Better to be more forgiving of misconfigured legend
services (there seems to be a lot!) and don't show this warning
right in the user's face...


nirvn (approved):
That was a lot of //crap cleaned up, good janitorial work :wink:

nyalldawson:
Thanks @nirvn!

gioman:
@nyalldawson thanks fot this, it was really annoying "see" those broken legends.

PR #38232, opened 2020-08-11
[processing] Allow merging layers with mixed attribute type

nirvn:
## Description

This PR updates the merge vector layers algorithm to tolerate mixed attribute types. When stumbling on a mixed attribute types scenario, the destination field type will be set to text.


nirvn:
BTW, we could have a different approach here and be tolerant by default and have a [x] do not accept mixed attribute types, which would abort the alg.

Or, we are just tolerant all the time and throw a info string in the alg output when we detect mixed types.

nirvn:
After some more thinking, I'll just have the tolerant behavior applied all the time. Updated PR coming in a couple of minutes.

nyalldawson (approved):

PR #38264, opened 2020-08-13
Harmonize geometry-based parameters name in expression

### LINKED ISSUE #38246: [expressions] named parameters and geometric predicates
In functions that have two geometries as arguments (see functions: intersect, contains, touches etc ...) the named parameters do not work, see attached screenshots:

here it works `geometry_a` and not `geometry1`
image

it doesn't work here
image

no distinction between topics:
image

image

image

arguments should be homogenized, because there are:
1. `geometry`
2. `geometry1`
3. `geometry a`

DelazJ:
using 'geometry1' (to mimic 'point1' parameter) instead of 'geometry_a' or 'geometry a'
Fixes #38246


DelazJ:
make_circle and make_ellipse registered 'geometry' in expressionfunction.cpp but expose in the json help file 'center'. center is more meaningful imo but I guess this can not be changed so simply

nyalldawson:
> make_circle and make_ellipse registered 'geometry' in expressionfunction.cpp but expose in the json help file 'center'. center is more meaningful imo but I guess this can not be changed so simply

If the "geometry" name was never publicly exposed then it's basic impossible that someone stumbled on it and is using it in a project. I'd be ok with you fixing the expression function itself to match the docs for this situation!

src/core/expression/qgsexpressionnodeimpl.cpp
```diff
@@ -91,6 +91,10 @@ QString QgsExpressionNode::NodeList::cleanNamedNodeName( const QString &name )
     cleaned = QStringLiteral( "geometry" );
   else if ( cleaned == QLatin1String( "val" ) )
     cleaned = QStringLiteral( "value" );
+  else if ( cleaned == QLatin1String( "geometry a" ) or cleaned == QLatin1String( "geometry_a" ) )
```
nyalldawson, line 94:
```suggestion
  else if (  cleaned == QLatin1String( "geometry_a" ) )
```

The versions spaces in them would never have worked...

src/core/expression/qgsexpressionnodeimpl.cpp
```diff
@@ -91,6 +91,10 @@ QString QgsExpressionNode::NodeList::cleanNamedNodeName( const QString &name )
...
+    cleaned = QStringLiteral( "geometry1" );
+  else if ( cleaned == QLatin1String( "geometry b" ) or cleaned == QLatin1String( "geometry_b" ) )
```
nyalldawson, line 96:
```suggestion
  else if ( cleaned == QLatin1String( "geometry_b" ) )
```

DelazJ:
> If the "geometry" name was never publicly exposed then it's basic impossible that someone stumbled on it and is using it in a project.

Somehow it's potentially exposed if you use the auto-completion. Typing `make_circle(` would display the parameters list with "geometry" in it. But I'm not sure of how many potential projects we are talking about (if any) and would be happy to just switch to center.

nyalldawson (approved):

PR #38277, opened 2020-08-13
Revert "FIX virtual layers with CSV layers opened with "delimitedtext " provider"

### LINKED ISSUE #38163: can't import anymore CSV files located on a network share
**Describe the bug**
In QGIS 3.14.1 I am unabe to import a UTF-8 CSV file as point data using UTM X and UTM Y coordinates like I have for quite some time now. This only occurs with CSV files hosted over my network connection on a separate computer that we use for collaboration. I made a .CSV file in Excel on my desktop that imported properly then moved it to our network storage and I was unable to import the CSV. I downloaded the long-term release 3.10 and it worked fine with both files no problem. Seems to be limited to 3.14.1. I tried a clean install but that did not work.

Returns the following "Critical Error" with some information from paths redacted.

Layer is not valid: The layer file://remote-desktop/DESK%20Documents/USA/Nevada/Nye%20Co,%20NV/%20Information/UTM%20Coordinates%20LMs.csv?type=csv&maxFields=10000&detectTypes=no&xField=Easting&yField=Northing&crs=EPSG:26711&spatialIndex=no&subsetIndex=no&watchFile=no is not a valid layer and can not be added to the map. Reason:

**How to Reproduce**

Reproduced with new files and old files. Import is unsuccessful from the start.

**QGIS and OS versions**

3.14.1 -Pi

nyalldawson:
Reverts https://github.com/qgis/QGIS/pull/37164, which has lead to
https://github.com/qgis/QGIS/issues/38163

Fixes #38163

Need to get this merged before 3.14.2 release today, as #38163 is a much more serious regression then the issue the PR was originally designed to fix

### DIFF
--- a/src/providers/delimitedtext/qgsdelimitedtextprovider.cpp
+++ b/src/providers/delimitedtext/qgsdelimitedtextprovider.cpp
@@ -62,15 +62,6 @@ QRegExp QgsDelimitedTextProvider::sCrdDmsRegexp( "^\\s*(?:([-+nsew])\\s*)?(\\d{1
 QgsDelimitedTextProvider::QgsDelimitedTextProvider( const QString &uri, const ProviderOptions &options )
   : QgsVectorDataProvider( uri, options )
 {
-  // uri should be in the form of "file:///path/to/file.csv?query=params", if not, enforce it in that format
-  // first read the already encoded url to get the query string
-  QUrl url = QUrl::fromEncoded( uri.toLatin1() );
-  // temporarily store the query string
-  const QString tmpUrlQuery = url.query();
-  // make sure that the url is actually prefixed with "file://". However, this breaks the query part ("?" char gets encoded), so discard the query string
-  url = QUrl::fromLocalFile( url.path() );
-  // finally restore the query part
-  url.setQuery( tmpUrlQuery );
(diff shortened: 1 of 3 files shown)


nyalldawson:
ping @suricactus I'm reverting your fix for 3.14.2, let's revisit when you've an alternative fix for the original issue available

m-kuhn (approved):

nyalldawson:
Thanks for looking this over @m-kuhn!

m-kuhn:
Thanks for remembering this @nyalldawson . Too bad, I hope we find a better solution within a month.

nyalldawson:
@m-kuhn 
>  Too bad, I hope we find a better solution within a month.

Still, I consider this one a great demonstration that our "protect the LTR at all costs" approach is working! :laughing: that's the good news at least!

m-kuhn:
> a great demonstration that our "protect the LTR at all costs" approach is working!

For the record: I'm 110 % convinced of this statement

suricactus:
pong @nyalldawson 
Sorry for the very late reply, I was off last week. Sorry for causing all the trouble, let's see if there is another solution for the same problem I tried to address.

PR #38279, opened 2020-08-14
Move QgsPalLayerSettings line settings api to QgsLabelLineSettings

nyalldawson:
Doing some more work to reduce the size of the QgsPalLayerSettings mega class, and provide nicer api for setting line related label settings


nyalldawson:
@3nids do you know which version of pyqt/sip/python we need to get enum classes to work correctly? It works fine locally, but fails on Travis. I'd like to bump our minimum Python/pyqt/sip version up so that these work reliably wherever QGIS is built. Right now we have a very awkward situation where the growing use of enum classes means that PyQGIS is simply broken on some platforms.

3nids:
@nyalldawson I don't recall exactly, I tried to find the issue on the [log](https://www.riverbankcomputing.com/hg/sip/shortlog) without success. 
The sure thing is that it's failing on Ubuntu Bionic (18.04).

One path would be that we only build LTR on Bionic and do not support stable anymore on it. That sounds like an acceptable solution to me (long-term OS, long-term software), but this position might not be shared among the users.

Another path would be to use an internal sip with QGIS, but I am not sure about the feasibility of this (having a different sip for QGIS than the one used in PyQt).

(and this would also lead to another recurring problem of updating the tests to make them pass on a newer CI OS)

PR #38285, opened 2020-08-14
Add topological points to all editable layers on split feature

### LINKED ISSUE #37823: Split features map tool only adds topological points to same layer
**Describe the bug**
When splitting features with the *split features* map tool and having *topological editing* enabled, topological points are added only to intersecting features of the same layer. To correctly preserve topology, all editable vector layers should have the points added, just like when using the *vertex tool*.

**QGIS and OS versions**
master on Debian testing

uclaros:
## Description
When splitting a feature, topological points are now added to all editable layers sharing the cut edge, thus preserving topology.
Previously, one would have to add the vertices beforehand with the vertex tool that correctly added the vertices to all layers.
before:
before
after:
after

Fixes #37823


src/core/qgsvectorlayer.h
```diff
@@ -1568,7 +1569,7 @@ class CORE_EXPORT QgsVectorLayer : public QgsMapLayer, public QgsExpressionConte
      * changes can be discarded by calling rollBack().
      * \since QGIS 3.16
      */
-    Q_INVOKABLE QgsGeometry::OperationResult splitFeatures( const QgsCurve *curve, bool preserveCircular = false, bool topologicalEditing = false );
+    Q_INVOKABLE QgsGeometry::OperationResult splitFeatures( const QgsCurve *curve, QgsPointSequence &topologyTestPoints SIP_OUT, bool preserveCircular = false, bool topologicalEditing = false );
```
3nids, line 1572: this is an API break as the result in Python is becoming a tuple

3nids (changes requested):
Good move!
There is just an API break.
Could this be accompanied by unit test?

src/core/qgsvectorlayer.h
uclaros: Is it an API break if it is not released yet? this method was just introduced with 24b1b3b4c0e and it will be released in 3.16

uclaros:
> Could this be accompanied by unit test?

Yes, unit test will be added!

src/core/qgsvectorlayer.h
3nids: oops, sorry, indeed.

3nids (approved):

src/core/qgsvectorlayer.h
uclaros: The truth is I wanted to add another overload to preserve the API but was getting collisions on the python signatures. Maybe we should deprecate more variants of this method for QGIS 4 or refactor topologicalPoints as a member instead of a parameter.

src/app/qgsmaptoolsplitfeatures.cpp
```diff
@@ -92,44 +92,84 @@ void QgsMapToolSplitFeatures::cadCanvasReleaseEvent( QgsMapMouseEvent *e )
...
-    else if ( returnCode == QgsGeometry::OperationResult::InvalidBaseGeometry )
-    {
-      QgisApp::instance()->messageBar()->pushMessage(
-        tr( "No feature split done" ),
-        tr( "The geometry is invalid. Please repair before trying to split it." ),
-        Qgis::Warning,
-        QgisApp::instance()->messageTimeout() );
-    }
-    else if ( returnCode != QgsGeometry::OperationResult::Success )
+    QgsGeometry::OperationResult returnCode = vlayer->splitFeatures( captureCurve(), topologyTestPoints, false, topologicalEditing );
```
uclaros, line 98: Had to change this parameter to `false`. With `true` all split polygons are silently transformed to `CurvePolygon` even if the split was done with a straight line.
@vcloarec would you mind shedding some light on this if it was intentional?

src/app/qgsmaptoolsplitfeatures.cpp
```diff
@@ -92,44 +92,84 @@ void QgsMapToolSplitFeatures::cadCanvasReleaseEvent( QgsMapMouseEvent *e )
...
+
+    switch ( returnCode )
```
uclaros, line 100: `returnCode` is `OperationResult::Success` even if no features were split. This change is a step closer to fixing #38287 but further changes are eventually needed.

src/app/qgsmaptoolsplitfeatures.cpp
vcloarec: Yes, it was intentional. As GEOS doesn't support splitting with curve, there is a workaround that is to discretize the curved elements, split with them and finally de-segmentize split features. But indeed, with that and I missed it, polygons are not polygons anymore but CurvedPolygons, even with straight split... Sorry about that... I purpose to keep the argument true, but to add this in the splitting method to deactivate the preserving if there is no curved element in the QgsCurve argument: PR here : https://github.com/qgis/QGIS/pull/38294

uclaros:
#38294 should be merged first to fix the failing test

src/app/qgsmaptoolsplitfeatures.cpp
uclaros: thanks!

uclaros:
Any chance we can have this on 3.16? It's been sitting here for some time and in 3.18 it will need further changes since it will cause an API break, further complicating the already many signatures of `QgsVectorLayer::splitFeatures` etc.
Also, it's borderline a feature and mostly a bug fix since it provides what is the expected behavior.

nyalldawson:
@3nids can you make the merge call please? You've previously reviewed this and are best place to judge the risk level!

uclaros:
Gentle ping @3nids

PR #38290, opened 2020-08-14
Fix make link to pre-commit

ismailsunni:
## Description

Small fix on command to make a link

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -13,7 +13,7 @@ by running the prepare commit script **before** issuing `git commit`.
 
 This can be automated by setting up the pre-commit hook properly.
 
-    ln -s ../../scripts/prepare_commit.sh .git/hooks/pre-commit
+    ln -s $PWD/scripts/prepare_commit.sh ./.git/hooks/pre-commit
 
 Getting your pull request merged
 --------------------------------


m-kuhn (approved):

ismailsunni:
@m-kuhn it seems so. I think my change is working unless the filesystem is mounted. Perhaps it's better to revert my change :), sorry for the noise

m-kuhn:
How is it related to a mounted filesystem? I'd first like to understand the reasoning before proceeding to ask for a revert :-)

ismailsunni:
so, by using the relative path, we can mount the QGIS source directory to somewhere, and the symlink will still work[1]. 

But, if we use the absolute path (my change), the symlink will probably not work if we mount the QGIS source directory to somewhere.

This mounting is not a common case I guess, but perhaps when someone moves the whole directory, it will still work (the absolute path will not work for sure). Example, it was `/home/ismail/dev/cpp/QGIS` to `/home/ismail/new-dev/cpp/QGIS`

[1] https://unix.stackexchange.com/questions/10370/make-a-symbolic-link-to-a-relative-pathname

m-kuhn:
What's the advantage of the absolute path?

jef-n:
> What's the advantage of the absolute path?

Good question - I don't see one either or why an extra `./` needed.

ismailsunni:
@m-kuhn I don't think it has any advantage now. Perhaps, at that time I couldn't create the symlink properly, then I thought it must be an absolute path. But after seeing your https://github.com/qgis/QGIS/pull/38290#issuecomment-678253847, I tried again with a relative path and it works. That's why I suggest this PR is reverted.

m-kuhn:
:+1: Can you do a PR?
Sorry I was a bit fast with approving that fast, I had the impression that you introduced relative paths, must have been confused :-)

PR #38294, opened 2020-08-15
fix transform Polygon to CurvePolygon when splitting

vcloarec:
When splitting with curve, resulting split features are transform to CurvePolygon to preserving curved element.
This PR deactivates this preserving when the curve does not contained curved element, so there is no transform to CurvePolygon if there are only straight segments.


nyalldawson:
Got a unit test for this one?

vcloarec:
> Got a unit test for this one?

it seems tests in https://github.com/qgis/QGIS/pull/38285 cover this fix.

uclaros:
The split tool lacked proper tests, that's why this bug was introduced. I created some tests with #38285 , so once I get it merged some curve split tests should be added there.

PR #38296, opened 2020-08-15
More updates of expression help

DelazJ:
- Expose default values of args
- Fix args name or description
- Fix/add more examples (to show various possibilities of a function or so that functions that relate have examples that relate)

### DIFF
--- a/resources/function_help/json/to_dm
+++ b/resources/function_help/json/to_dm
@@ -7,10 +7,11 @@
     {"arg":"coordinate","description":"A latitude or longitude value."},
     {"arg":"axis","description":"The axis of the coordinate. Either 'x' or 'y'."},
     {"arg":"precision", "description":"Number of decimals."},
-    {"arg":"formatting", "optional": true, "default":"", "description":"Designates the formatting type. Acceptable values are NULL, 'aligned' or 'suffix'."}
+    {"arg":"formatting", "optional": true, "default":"", "description":"Designates the formatting type. Acceptable values are NULL (default), 'aligned' or 'suffix'."}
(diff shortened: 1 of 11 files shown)


resources/function_help/json/to_dm
```diff
@@ -7,10 +7,11 @@
     {"arg":"coordinate","description":"A latitude or longitude value."},
     {"arg":"axis","description":"The axis of the coordinate. Either 'x' or 'y'."},
     {"arg":"precision", "description":"Number of decimals."},
-    {"arg":"formatting", "optional": true, "default":"", "description":"Designates the formatting type. Acceptable values are NULL, 'aligned' or 'suffix'."}
+    {"arg":"formatting", "optional": true, "default":"", "description":"Designates the formatting type. Acceptable value is NULL (default), 'aligned' or 'suffix'."}
```
nyalldawson, line 10: This reads worse to me -- why the change?

DelazJ: - the `"default" : ""` does not render in QGIS, hence the precision next to the NULL value
- with a plural form, I'd expect an "AND" list so either `Acceptable values are NULL, 'aligned' and 'suffix'` or use the singular form with "OR".

nyalldawson: Mmm... it doesn't read well to me -- it's just awkward grammar. I'm not an English teacher, but I'm 99.9% sure the original form is correct.

DelazJ: Reverted...

nyalldawson (approved):

PR #38301, opened 2020-08-15
Adjust width of relation reference config widget

m-kuhn:
Reduce minimum width of the combobox.
This has bugged me for ages. The widget configuration taking more horizontal space than required.

# Before (horizontal scroll bar visible)

image

# After (no horizontal scroll bar)

image


nirvn (approved):

m-kuhn:
UX fixes like this are backportable I guess?

nirvn:
No objection to backporting this over here

PR #38311, opened 2020-08-16
[processing] Upgrade the sample raster value algorithm to native C++

nirvn:
## Description

Another python algorithm bites the dust. 

Quick test with a 14,000 points dataset against a 1-band DEM raster:

```
Python:
Execution completed in 17.21 seconds

Native C++:
Execution completed in 8.27 seconds
```

:tada:


src/analysis/processing/qgsalgorithmrastersampling.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+bool QgsRasterSamplingAlgorithm::prepareAlgorithm( const QVariantMap &parameters, QgsProcessingContext &context, QgsProcessingFeedback * )
+{
+  QgsRasterLayer *layer = parameterAsRasterLayer( parameters, QStringLiteral( "INPUT_RASTER" ), context );
+  if ( !layer )
+    throw QgsProcessingException( invalidRasterError( parameters, QStringLiteral( "INPUT_RASTER" ) ) );
+
+  mBandCount = layer->bandCount();
+  mCrs = layer->crs();
+  mDataProvider.reset( static_cast< QgsRasterDataProvider *>( layer->dataProvider()->clone() ) );
```
nyalldawson, line 80: :+1:

src/analysis/processing/qgsalgorithmrastersampling.cpp
```diff
@@ -0,0 +1,166 @@
...
+{
+  std::unique_ptr< QgsFeatureSource > source( parameterAsSource( parameters, QStringLiteral( "INPUT" ), context ) );
+  if ( !source )
+    throw QgsProcessingException( invalidSourceError( parameters, QStringLiteral( "INPUT" ) ) );
+
+  QString fieldPrefix = parameterAsString( parameters, QStringLiteral( "COLUMN_PREFIX" ), context );
+  QgsFields newFields;
+  for ( int band = 1; band <= mBandCount; band++ )
+  {
+    newFields.append( QgsField( QStringLiteral( "%1%2" ).arg( fieldPrefix, QString::number( band ) ), QVariant::Double ) );
```
nyalldawson, line 95: If we wanted to be cleverer, we could check the raster data type and use an int field here when appropriate.

src/analysis/processing/qgsalgorithmrastersampling.cpp
```diff
@@ -0,0 +1,166 @@
...
+    feedback->setProgress( current * step );
+
+    if ( !feature.hasGeometry() )
+    {
+      current++;
+      continue;
+    }
+
+    QgsGeometry geometry = feature.geometry();
+    if ( geometry.isMultipart() && geometry.get()->partCount() > 1 )
```
nyalldawson, line 127:
```suggestion
    if ( geometry.isMultipart() && geometry.get()->partCount() != 1 )
```

src/analysis/processing/qgsalgorithmrastersampling.cpp
```diff
@@ -0,0 +1,166 @@
...
+    {
+      sink->addFeature( feature, QgsFeatureSink::FastInsert );
+      feedback->reportError( QObject::tr( "Impossible to sample data of multipart feature %1." ).arg( feature.id() ) );
+      continue;
+    }
+    QgsPointXY point = geometry.isMultipart() ? geometry.asMultiPoint()[0] : geometry.asPoint();
```
nyalldawson, line 133:
```suggestion
    QgsPointXY point( geometry.isMultipart() ? qgs_geometrycast< const QgsPoint* >( qgs_geometrycast< const QgsMultiPoint* >( geometry.constGet() )->geometryN( 0 ) ) : qgs_geometrycast< const QgsPoint* >( geometry.constGet() ) );
```

Just a bit more efficient!

TODO: we really need to improve the api for the QgsGeometryCollection subclasses. It's not at all friendly that retrieving a specific point from a multipoint geometry requires this mess of casts!

src/analysis/processing/qgsalgorithmrastersampling.cpp
```diff
@@ -0,0 +1,166 @@
...
+    try
+    {
+      point = ct.transform( point );
+    }
+    catch ( const QgsException &e )
+    {
+      sink->addFeature( feature, QgsFeatureSink::FastInsert );
```
nyalldawson, line 140: You need to populate the sample fields with NULL values here, or the memory provider will reject the add.

nyalldawson (commented):
Nice cleanup!

PR #38330, opened 2020-08-17
Server landingpage: GCC does not like uniform initialization

### LINKED ISSUE #38308: Unable to load server landingpage
**Describe the bug**
`[{"code":"Internal server error","description":"The API handler returned an error: [json.exception.type_error.305] cannot use operator[] with a string argument with array"}].`
Something around [this line](https://github.com/qgis/QGIS/blob/master/src/server/services/landingpage/qgslandingpagehandlers.cpp#L72) appears to be causing the problem - the variable `info` appears to be an json array rather than a dictionary so an`id` property cannot be assigned, With the [following line](https://github.com/qgis/QGIS/blob/master/src/server/services/landingpage/qgslandingpagehandlers.cpp#L73) commented out the error message no longer appears and there is a response in `index.json`, but no project summaries load, which I believe is a side effect of the json being incorrect as a result of the issue above.

**How to Reproduce**
Start qgis_mapserver, navigate to localhost:8000 in browser, server successfully reads projects in `QGIS_SERVER_PROJECTS_DIRECTORIES` but does not create landing page

**Environment**
Using QGIS 3.15 master 4e795316cd on Ubuntu 20.04

elpaso:
Fixes #38308

### DIFF
--- a/src/server/services/landingpage/qgslandingpagehandlers.cpp
+++ b/src/server/services/landingpage/qgslandingpagehandlers.cpp
@@ -44,7 +44,7 @@ void QgsLandingPageHandler::handleRequest( const QgsServerApiContext &context ) 
   }
   else
   {
-    const json projects { projectsData( ) };
+    const json projects = projectsData( ) ;
     json data
     {
       { "links", links( context ) },
@@ -64,7 +64,7 @@ const QString QgsLandingPageHandler::templatePath( const QgsServerApiContext &co
 
 json QgsLandingPageHandler::projectsData() const
 {
-  json j { json::array() };
+  json j = json::array();
   const auto availableProjects { QgsLandingPageUtils::projects( ) };
   const auto constProjectKeys { availableProjects.keys() };
   for ( const auto &p : constProjectKeys )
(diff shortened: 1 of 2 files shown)


stev-0:
Thanks. I seem to remember something similar with one of my PRs - see https://github.com/qgis/QGIS/pull/31293.

What is the compiler used to run the builds and the automated test suite? Should I be looking to build with that or is compatibility across compilers more important?

elpaso:
Yeah, it's the same issue, the problem is that I tend to forget it :/

CI for linux (and Mac?) is using clang as most of developers are.

m-kuhn:
Clang gives really verbose hints and build warnings.
Adding a gcc build might be a good idea too nonetheless.
Or sticking to C++ constructs that are predictable :stuck_out_tongue:

nyalldawson:
> CI for linux (and Mac?) is using clang as most of developers are.

I'm not :D

nyalldawson (approved):

nyalldawson:
(Wonder if we should add a unit test to `git grep "json.* {"` and fail if it's found (and also not use auto so that this test works;) )

PR #38332, opened 2020-08-17
Server landingpage initial extent

elpaso:
When loading a map in the map browser set initial extent to what was set in the project (map canvas)


src/server/services/landingpage/qgslandingpageutils.cpp
```diff
@@ -222,6 +223,39 @@ json QgsLandingPageUtils::projectInfo( const QString &projectUri )
   json info = json::object();
   info[ "id" ] = QCryptographicHash::hash( projectUri.toUtf8(), QCryptographicHash::Md5 ).toHex();
   QgsProject p;
+
+  // Read initial extent from map canvas
+  QObject::connect( &p, &QgsProject::readProject, qApp, [ & ]( const QDomDocument & projectDoc )
```
nyalldawson, line 228: Could you use https://github.com/qgis/QGIS/blob/master/src/core/qgsprojectviewsettings.h#L62 instead? Or extend that class to have a proper stable API for the storage of all canvas extents?

elpaso: > Could you use https://github.com/qgis/QGIS/blob/master/src/core/qgsprojectviewsettings.h#L62 instead? Or extend that class to have a proper stable API for the storage of all canvas extents?

That was indeed my first approach, but testing it in desktop with a project opened gives me:

```python
In [1]: project = QgsProject.instance()

In [2]: v =project.viewSettings()

In [3]: v.defaultViewExtent()
Out[3]: <QgsReferencedRectangle: 0 0, 0 0 ()>
```

also, `setDefaultViewExtent()` is never called so I don't really get how it's supposed to work.

PR #38340, opened 2020-08-18
Change placement mode selection widget from radio buttons to combobox, rework UI

nyalldawson:
This PR reworks the Label "Placement" tab. It's currently a very dense, widget heavy tab which is daunting for new users (and I suspect contains many options which aren't understood even by experienced users)

So in this PR I've:
- Changed the existing radio button selection for placement modes to a combo box. The combo box is more compact and avoids the messy oversized padding which is forced on the radio buttons for non-polygon layers. The more compact widget gives us space to add some nice explanatory text. Before and after screenshots below. (Some nice illustrative icons would be great to add to the combo box too!)

image

image

Some other screenshots:

image

image

image

image

image

- I've also moved the "Repeating Labels" and "Label Overrun" settings to collapsible groups, with nice explanatory text. Before/after images below. (Again, I think some nice illustrations sitting to the left of the helper text would be great to add in future)

image

image

### DIFF
--- a/python/gui/auto_generated/qgstextformatwidget.sip.in
+++ b/python/gui/auto_generated/qgstextformatwidget.sip.in
@@ -155,6 +155,15 @@ Controls whether data defined alignment buttons are enabled.
     virtual QgsExpressionContext createExpressionContext() const;
 
 
+    QgsWkbTypes::GeometryType labelGeometryType() const;
+%Docstring
+Returns the geometry type which will be used by the labeling engine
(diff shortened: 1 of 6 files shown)


NathanW2:
This looks really nice.

nirvn:
Oh man, what a difference.

BTW, I really love the descriptive UI style you've been using in recent months. We should definitively try and use that in as many parts as possible.

PR #38343, opened 2020-08-18
Fix/add tabstops to widgets for a better keyboard use

DelazJ:


m-kuhn:
So important, so often neglected.

m-kuhn (approved):

PR #38345, opened 2020-08-18
Add a scrollbar to 3D Map view rendering settings

### LINKED ISSUE #38134: Layer Properties -> 3D View needlessly forces dialog minimum height
**Describe the bug**
Switching to *3D View* on *Layer Properties* dialog messes with the dialog's minimum height, forcing it to enlarge and not entirely fit on a oldschool laptop 1366x768 screen.
I know it's a bit dated hardware, but there is no reason for 3D View tab to have a larger minimum height than the other tabs.

**How to Reproduce**
- Open layer properties on a 1366x768 screen
- Switch to 3D View tab
- The dialog is enlarged and no longer fit entirely on the screen.

DelazJ:
to allow resizing and avoid dialog getting big when enabled (should fix #38134 )
3dscrolls

I'm however seeing an issue with the shading group for line and point features (not polygons), in that it shrinks if the dialog is reduced. If any hint where I should look...
3dpointshrinks

### DIFF
--- a/src/app/3d/qgsvectorlayer3drendererwidget.cpp
+++ b/src/app/3d/qgsvectorlayer3drendererwidget.cpp
@@ -25,6 +25,7 @@
 #include "qgsapplication.h"
 #include "qgs3dsymbolregistry.h"
 #include "qgsabstractmaterialsettings.h"
+#include "qgsvscrollarea.h"
 
 #include <QBoxLayout>
 #include <QCheckBox>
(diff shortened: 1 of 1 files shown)


timlinux:
I spent the evening last night trying to resolve the squashed materials box with no luck :-( Hope someone can figure it out....

DelazJ:
The latest commit moves the whole tab in the scroll area, giving more scrolling area to rendering settings, and mimicks the symbology tab
image
but I now realize that, if you enable "single symbol", then go back to "no symbols" you also get this (the layer rendering group is hidden at the bottom of the scroll area). Doesn't happen if you first time open the widget.
image
I don't know whether I should revert. I think this layout is better than the previous but the drawback is quite huge. If anyone has an  idea...? I should stop trying to fix GUI in c++ :smile: 
> I spent the evening last night trying to resolve the squashed materials box with no luck :-( Hope someone can figure it out....

Thanks @timlinux Let's hope together

PR #38348, opened 2020-08-18
Fix delimitedtext in virtual layers

suricactus:
The following change is inspired while addressing the bug described here https://github.com/qgis/QGIS/pull/37164 TLDR:

> When using a CSV layer opened via "delimitedtext" provider (e.g. the delimited text dialog) in virtual layer, everything work fine until the project is reloaded. In that case, some weird behabiour is observed - the virtual layer becomes invalid, because the CSV file name is not prefixed with file:// and this breaks the layer itself, the layer properties etc.

So I checked the layer source in the `.qgs` project file after saving and I already spot the difference in the way how the path to file is stored for "delimitedtext" layers. 

The original CSV is stored as:
```xml
<datasource>file:./delimitedtext.csv?type=csv&amp;delimiter=%5Ct;&amp;maxFields=10000&amp;detectTypes=yes&amp;geomType=none&amp;subsetIndex=no&amp;watchFile=no</datasource>
# for readability
<datasource>file:./delimitedtext.csv?type=csv&amp;delimiter=\t;&amp;maxFields=10000&amp;detectTypes=yes&amp;geomType=none&amp;subsetIndex=no&amp;watchFile=no</datasource>
```
while the virtual layer uses:

```xml
<datasource>?layer=delimitedtext:.%2Fdelimitedtext.csv%3Ftype%3Dcsv%26delimiter%3D%5Ct%3B%26maxFields%3D10000%26detectTypes%3Dyes%26geomType%3Dnone%26subsetIndex%3Dno%26watchFile%3Dno:delimitedtext:&amp;query=SELECT%20*%20FROM%20delimitedtext</datasource>
# for readability
<datasource>?layer=delimitedtext:./delimitedtext.csv?type=csv&delimiter=\t;&maxFields=10000&detectTypes=yes&geomType=none&subsetIndex=no&watchFile=no:delimitedtext:&amp;query=SELECT * FROM delimitedtext</datasource>
```
Note the missing `file:` prefix, even though `file:` prefixed files must have an absolute path, QGIS uses this a workaround to  remember the `delimitedtext` paths relative to the project root if configured to. Everything from now on happens on writing/reading the project XML.

The prefix is added by `QgsMapLayer::encodedSource` would make it invalid and relative to the project root with `file:` prefix, and `QgsMapLayer::decodedSource` would make it valid absolute path with `file://` prefix. All of this is because `QgsDelimitedtextProvider` requires url's as data source and I see it rather complex to change it's behaviour (NOTE the unexpected side effects with Windows sharing in the previous PR). 

The methods are overloaded by `QgsVectorLayer::(d)ecodedSource` and there are long if statemets, responsible for this kind of modifications for each data source. In the case of "delimitedtext" - add/remove `file:` prefix and make relative/absolute path. However, virtual layers are a special case, that is not handled properly and this `file:` prefixing is not being done for "delimitedtext" sources. This leads to broken reloading of the csv layer.

This PR addresses this bug of virtual layers by adding additional if in virtual layers, handling the special case for delimited text. After the modification, the virtual layer datasource in the .qgs file looks like:

```xml
<datasource>?layer=delimitedtext:file%3A.%2Fdelimitedtext.csv%3Ftype%3Dcsv%26delimiter%3D%255Ct%3B%26maxFields%3D10000%26detectTypes%3Dyes%26geomType%3Dnone%26subsetIndex%3Dno%26watchFile%3Dno:delimitedtext:&amp;query=SELECT%20*%20FROM%20delimitedtext</datasource>
# for readability
"<datasource>?layer=delimitedtext:file:./delimitedtext.csv?type=csv&delimiter=%5Ct;&maxFields=10000&detectTypes=yes&geomType=none&subsetIndex=no&watchFile=no:delimitedtext:&amp;query=SELECT * FROM delimitedtext</datasource>"
```
Which results in properly prefixed filename and escsaped query.

Even though not the best solution in terms of maintainability, it fixes the original issue in a way that supports easy backporting.

Nevertheless, I would suggest and I am willing to volunteer for a solution where the `QgsVectorLayer::(d)encodedSource` is not using a huge if statement, but each provider has implemented their own `(d)encodedSource` method, so we can remove code duplication and prevent similar mistakes in the future.

@m-kuhn @nyalldawson What do you think?

### DIFF
--- a/src/core/qgsvectorlayer.cpp
+++ b/src/core/qgsvectorlayer.cpp
@@ -1966,8 +1966,20 @@ QString QgsVectorLayer::encodedSource( const QString &source, const QgsReadWrite
         // syntax: provider:url_encoded_source_URI(:name(:encoding)?)?
         theURIParts = value.split( ':' );
         theURIParts[1] = QUrl::fromPercentEncoding( theURIParts[1].toUtf8() );
-        theURIParts[1] = context.pathResolver().writePath( theURIParts[1] );
-        theURIParts[1] = QUrl::toPercentEncoding( theURIParts[1] );
+
+        if ( theURIParts[0] == QLatin1String( "delimitedtext" ) )
+        {
+          QUrl urlSource = QUrl( theURIParts[1] );
+          QUrl urlDest = QUrl::fromLocalFile( context.pathResolver().writePath( urlSource.toLocalFile() ) );
+          urlDest.setQuery( urlSource.query() );
+          theURIParts[1] = QUrl::toPercentEncoding( urlDest.toString(), QByteArray( "" ), QByteArray( ":" ) );
+        }
+        else
+        {
+          theURIParts[1] = context.pathResolver().writePath( theURIParts[1] );
+          theURIParts[1] = QUrl::toPercentEncoding( theURIParts[1] );
+        }
+
         queryItems[i].second =  theURIParts.join( QStringLiteral( ":" ) ) ;
       }
     }
@@ -2040,7 +2052,28 @@ QString QgsVectorLayer::decodedSource( const QString &source, const QString &pro
         // syntax: provider:url_encoded_source_URI(:name(:encoding)?)?
         theURIParts = value.split( ':' );
         theURIParts[1] = QUrl::fromPercentEncoding( theURIParts[1].toUtf8() );
-        theURIParts[1] = context.pathResolver().readPath( theURIParts[1] );
+
+        if ( theURIParts[0] == QStringLiteral( "delimitedtext" ) )
+        {
+          QUrl urlSource = QUrl( theURIParts[1] );
+
+          if ( !theURIParts[1].startsWith( QLatin1String( "file:" ) ) )
+          {
+            QUrl file = QUrl::fromLocalFile( theURIParts[1].left( theURIParts[1].indexOf( '?' ) ) );
+            urlSource.setScheme( QStringLiteral( "file" ) );
+            urlSource.setPath( file.path() );
+          }
+
+          QUrl urlDest = QUrl::fromLocalFile( context.pathResolver().readPath( urlSource.toLocalFile() ) );
+          urlDest.setQuery( urlSource.query() );
+
+          theURIParts[1] = urlDest.toString();
+        }
+        else
+        {
+          theURIParts[1] = context.pathResolver().readPath( theURIParts[1] );
+        }
+
         theURIParts[1] = QUrl::toPercentEncoding( theURIParts[1] );
         queryItems[i].second =  theURIParts.join( QStringLiteral( ":" ) ) ;
       }


m-kuhn:
Thanks for this update and the detailed description @suricactus 
The current approach taken for en-/decoding looks cumbersome and I am in favor of giving this a try. From my perspective it's hard to give an estimation of possible pitfalls, so let's merge master first, 3.14 in a week and follow up with 3.10 backport after the next point release.

I also think that deduplicating this code into individual dataproviders would make a lot of sense. Ideally, qgsvectorlayer.cpp should not have *any* code that handles provider specific tasks.

nyalldawson:
Have you double checked that this approach works fine with csvs stored on a windows network share?

suricactus:
To be tested, don't merge yet.

suricactus:
Tested, it saves the map layer data source without modifications:

```
BEFORE SAVING
CSV
file://shared/c/Users/myuser/Desktop/test.csv?type=csv&delimiter=;&maxFields=10000&detectTypes=yes&xField=lon&yField=lat&crs=EPSG:4326&spatialIndex=no&subsetIndex=no&watchFile=no
VIRTUAL
?layer=delimitedtext:file%3A%2F%2Fshared%2Fc%2FUsers%2Fmyuser%2FDesktop%2Ftest.csv%3Ftype%3Dcsv%26delimiter%3D%3B%26maxFields%3D10000%26detectTypes%3Dyes%26xField%3Dlon%26yField%3Dlat%26crs%3DEPSG%3A4326%26spatialIndex%3Dno%26subsetIndex%3Dno%26watchFile%3Dno:test:&query=SELECT%20*%20FROM%20test

AFTER SAVING
CSV
file://shared/c/Users/myuser/Desktop/test.csv?type=csv&delimiter=;&maxFields=10000&detectTypes=yes&xField=lon&yField=lat&crs=EPSG:4326&spatialIndex=no&subsetIndex=no&watchFile=no
VIRTUAL
?layer=delimitedtext:file%3A%2F%2Fshared%2Fc%2FUsers%2Fmyuser%2FDesktop%2Ftest.csv%3Ftype%3Dcsv%26delimiter%3D%3B%26maxFields%3D10000%26detectTypes%3Dyes%26xField%3Dlon%26yField%3Dlat%26crs%3DEPSG%3A4326%26spatialIndex%3Dno%26subsetIndex%3Dno%26watchFile%3Dno:test:&query=SELECT%20*%20FROM%20test
```

suricactus:
Just to clarify, I have tested both local and network shared CSV as a source for a virtual layer. Tested loading and using before and after project being saved and reloaded. In any of the tested situations it works as expected, solving the original issue and not having the side effects from the previous fix attempt.

nyalldawson:
Ok, great! Thanks for the confirmation

m-kuhn (approved):

PR #38349, opened 2020-08-18
WMS Server: display visibilityChecked property of a layer in GetProjectSettings

mhugent:
In QGIS, a layer can be checked but still invisible if the parent group is not visible (the layer is then kind of greyed out). To allow the webclient mimic this behaviour, it needs additional information in the GetProjectSettings response. This PR adds 'visiilityChecked' in addition to 'visible'.


elpaso (approved):

elpaso:
@mhugent LGTM, just update the test reference files.

Also, you might want to check out the new landing page API, it provides a full TOC tree in json format, it's ideal for web clients.

PR #38350, opened 2020-08-18
Fix transformCoordinates() for rasters with RPC georeferencing

wonder-sk:
Unfortunately for raster with RPC georeferencing, the transform used in GDAL provider's transformCoordinates() was using RPC_HEIGHT at zero elevation, while rendering pipeline was using RPC_HEIGHT at the correct elevation from HEIGHT_OFF metadata. This was causing shift in what was rendered in map canvas vs what transform function returned. This is now fixed by also the transform function using the correct elevation.

### DIFF
--- a/src/core/qgsgdalutils.cpp
+++ b/src/core/qgsgdalutils.cpp
@@ -490,6 +490,18 @@ GDALDatasetH QgsGdalUtils::rpcAwareAutoCreateWarpedVrt(
   return GDALAutoCreateWarpedVRTEx( hSrcDS, pszSrcWKT, pszDstWKT, eResampleAlg, dfMaxError, psOptionsIn, opts );
 }
 
+void *QgsGdalUtils::rpcAwareCreateTransformer( GDALDatasetH hSrcDS, GDALDatasetH hDstDS, char **papszOptions )
+{
+  if ( GDALGetMetadata( hSrcDS, "RPC" ) )
+  {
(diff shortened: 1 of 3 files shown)


src/core/qgsgdalutils.cpp
```diff
@@ -490,6 +490,18 @@ GDALDatasetH QgsGdalUtils::rpcAwareAutoCreateWarpedVrt(
...
 }
 
+void *QgsGdalUtils::rpcAwareCreateTransformer( GDALDatasetH hSrcDS, GDALDatasetH hDstDS, char **papszOptions )
+{
+  if ( GDALGetMetadata( hSrcDS, "RPC" ) )
+  {
+    // well-behaved RPC should have height offset a good value for RPC_HEIGHT
+    const char *heightOffStr = GDALGetMetadataItem( hSrcDS, "HEIGHT_OFF", "RPC" );
+    if ( heightOffStr )
+      papszOptions = CSLAddNameValue( papszOptions, "RPC_HEIGHT", heightOffStr );
```
rouault, line 500: you should CSLDuplicate() the input papszOptions, and CSLDestroy() it after use. Currently, the caller of rpcAwareCreateTransformer() might CSLDestroy() something that has been destroyed by CSLAddNameValue()

wonder-sk: Ah good point, thanks - fixed in 75965e5

PR #38354, opened 2020-08-18
Refactor annotation item subclasses so that they work correctly with curved geometries

nyalldawson:
Better to be more flexible!

Also fix some bugs found along the way, improve some code...


nirvn (approved):

nyalldawson:
Thanks @nirvn!

PR #38355, opened 2020-08-19
Fix bad handling of non-square svg images in svg marker symbol layers

nyalldawson:
A bunch of fixes which collectively fix a number of issues when handling non-square svg images in svg marker symbol layers


nyalldawson:
Sponsored by SLYR

PR #38361, opened 2020-08-19
Expose option to allow users to control whether anchor point for line labels is a hint or a strict requirement

nyalldawson:
Strict: Labels are placed exactly on the label anchor only, and no
other fallback placements are permitted.

Hint: The label anchor is treated as a hint for the preferred label
placement, but other placements close to the anchor point are permitted.

Peek 2020-08-19 17-29


roya0045:
This seems to be only valid for lines, Was there a plan to apply similar things to other geometry?

I find that label placement with polygons is hit or miss as the line placement options are provided and it tends to chose a random contour line, thus it can be inside or outside randomly. Thought this issue might have been solved already.

nyalldawson:
It's only for lines -- for polygons/points you'd change the labelling mode to "offset from centroid" or "offset from point" if you want a strict, single candidate label

roya0045:
> It's only for lines -- for polygons/points you'd change the labelling mode to "offset from centroid" or "offset from point" if you want a strict, single candidate label

Could we eventually have an option to force the label inside outside as on its own or in combination with the line placement option for polygons? My main gripe is that it provide the best angle but the line chosen is inconsistent

nyalldawson:
> Could we eventually have an option to force the label inside outside as on its own or in combination with the line placement option for polygons? My main gripe is that it provide the best angle but the line chosen is inconsistent

Can you show an illustration? I'm having trouble following...

roya0045:
[label_test.zip](https://github.com/qgis/QGIS/files/5102783/label_test.zip)
@nyalldawson here is a simple use case when I usually want to have the label within the box. Sometimes to get them to place properly you must tick many placement option, sometimes 'left of line' place it outside of the box to the right for some of them. Using a mix of settings provide varying and inconsistent results as it seems to thing that the outside lines have directions, which seems kinda strange in some ways conceptually.

nyalldawson:
@roya0045 I'd suggest you could achieve what you're after by:

- using the "free" placement mode
- using a perimenter placement mode, and select "line orientation dependent position", "left of line", "right of line", then "draw labels which fit completely within feature only" from the "rendering" tab (right at the bottom)

I agree that this could be made nicer, but it's unrelated to the current PR.

PR #38370, opened 2020-08-19
[virtual layers] support curved geometries (segmentized)

olivierdalang:
## Description

So far, it wasn't possible to input curved geometries in virtual layers as they aren't supported by Spatialite. The curved geometries would be treated as empty geometries.

This PR fixes this by segmentizing geometries prior to feeding them to Spatialite.

(internal issue : https://github.com/opengisch/QGIS-Issues-SH/issues/67)

### DIFF
--- a/tests/src/python/test_provider_virtual.py
+++ b/tests/src/python/test_provider_virtual.py
@@ -357,7 +357,11 @@ class TestQgsVirtualLayerProvider(unittest.TestCase, ProviderTestCase):
                (3, "POLYGON", "((0 0,1 0,1 1,0 0))"),
                (4, "MULTIPOINT", "((1 1))"),
                (5, "MULTILINESTRING", "((0 0,1 0),(0 1,1 1))"),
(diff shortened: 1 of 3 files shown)


tests/src/python/test_provider_virtual.py
```diff
@@ -372,7 +376,13 @@ def test_geometryTypes(self):
             l2 = QgsVectorLayer("?layer_ref=%s" % l.id(), "vtab", "virtual", QgsVectorLayer.LayerOptions(False))
             self.assertEqual(l2.isValid(), True)
             self.assertEqual(l2.dataProvider().featureCount(), 1)
-            self.assertEqual(l2.dataProvider().wkbType(), wkb_type)
+            # we ensure the geom type matches the original (semgentized) type
```
nyalldawson, line 379: "Segmentized"

nyalldawson:
LGTM. I guess there's no option to upgrade the backend of virtual layers here?

olivierdalang:
> LGTM. I guess there's no option to upgrade the backend of virtual layers here?
Indeed I don't really see an option besides spatialite ?

PR #38383, opened 2020-08-20
Allow adding new features via attribute form

### LINKED ISSUE #37846: Adding a new feature from attribute table fails, but works from digitizing toolbar
On tables with constraints, adding a new feature from the attribute table fails (with errors about violated constraints, the feature form does not open), but adding a new feature on the same table from the digitizing toolbar seems to work fine, opening the form and letting the user fill in some data in order to fulfill the constraints.

See this [video here](https://youtu.be/U6aSFfVU8OQ).

Tested on master and QGIS 3.10.

suricactus:
Addresses to fix #37846. See also the proposed 

The attribute table is already a complex enough, so a possible solution changing there is hard enough. Also little to no change for backporting. Note the behaviour in the described bug is only present when in "Project properties" the checkbox "Automatically create transaction groups where possible" is checked.

The proposed solution is:
> Make the current "Add feature" button to have a little arrow with two actions: "Add feature via attribute form" and "Add feature via attribute table". By default we have the latter, but when transaction mode is enabled, it is disabled and only the "Add feature via attribute form" is enabled. This will be minimal and backportable code and UI layout change which will overcome the attribute table limitations, as well as adding the convinience of using the attribute form when desired. As a bonus, the button remembers the last selection of action - add via form or via table.

Peek 2020-08-20 10-46

Even though it might feel as a duplication of form view in the attribute table, it is a change that touches the least amount of code and allows using the attribute table to add new features when they don't match the constraint. Which also makes the solution backportable.

### DIFF
--- a/src/app/qgsattributetabledialog.cpp
+++ b/src/app/qgsattributetabledialog.cpp
@@ -17,6 +17,7 @@
 #include <QMessageBox>
 #include <QGridLayout>
 #include <QDialogButtonBox>
+#include <QMenu>
 
 #include "qgsattributetabledialog.h"
 #include "qgsattributetablemodel.h"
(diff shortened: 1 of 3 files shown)


m-kuhn (approved):
Looks very good for a minimal invasive change

PR #38400, opened 2020-08-20
Build with Qt 5.14 only when sources have changed

3nids:


m-kuhn:
Test updates and CMakeLists.txt changes should be built as well (maybe even more?)

PR #38405, opened 2020-08-20
[feature][expression] refFunction port to core

olivierdalang:
## Description

This PR ports the [refFunction](https://plugins.qgis.org/plugins/refFunctions/) plugin to core. RefFunctions are spatial operators expressions, allowing to make spatial queries on other layers from withing expressions, opening up a lot of really powerful workflows.

The hard work was done by @enricofer (the author of the plugin) and @m-kuhn , this is a rebase and polishing of their work.

reffunc2

The following new expression functions are implemented :
```
geometry_overlay_intersects(layer[,expression][,filter][,limit][,cache])
geometry_overlay_contains(layer[,expression][,filter][,limit][,cache])
geometry_overlay_crosses(layer[,expression][,filter][,limit][,cache])
geometry_overlay_equals(layer[,expression][,filter][,limit][,cache])
geometry_overlay_touches(layer[,expression][,filter][,limit][,cache])
geometry_overlay_disjoint(layer[,expression][,filter][,limit][,cache])
geometry_overlay_within(layer[,expression][,filter][,limit][,cache])
geometry_overlay_nearest(layer[,expression][,filter][,limit][,max_distance][,cache])
```

They all work in the same way : the spatial operator is evaluated against the other layer. If an expression is provided, it returns an array of results. If no expression is provided, it returns a boolean telling whether at least one feature matched.
Features can optionally be filtered by an expression, and optionally be limited to a certain count.

The `geometry_overlay_nearest` has an additional max_distance filter.

I'll leave this PR as a draft for some time as I still need to test it a little more and there's some more optimisation that can be done.


olivierdalang:
About naming the functions, I'm not 100% sure of the choice...
Please comment if you have strong opinions on this.

Here are some alternatives :
```
geometry_overlay_intersects(...)
spatial_join_intersects(...)
spatial_operation_intersects(...)
so_intersects(...)
```

It would also be possible to have just one function (except nearest, that has a different signature), and provide the operator as an argument :
```
geometry_overlay('intersects', ...)
spatial_join('intersects', ...)
spatial_operation('intersects', ...)
```

Things to keep in mind :
- possible confusion with plain spatial operators functions (such as `intersects(...)`)
- separate functions are more auto-completion friendly, and make it easier to discover the functions to the user
- separate functions overcrowds the function list, is more verbose in the code and makes it less clear to the user that all these functions work in the same way.

bru500:
+1 for just one function, because it makes it more clear, that all the functions work the same and a better over all view.

enricofer:
@m-kuhn I'm proud to have you as great mentor! thanks for sharing with me your ideas and know-how.
@olivierdalang really a great work! Thank you for the professional "refurbishment" of that almost forgotten piece of code.

I'm very happy for taking care with you of a remote intuition and to see it landing in core.

> 
> 
> About naming the functions, I'm not 100% sure of the choice...
> Please comment if you have strong opinions on this.

What about spreading the question to seek a better naming solution among a bunch of selected developers and power users? Or do we risk to have more noise and confusion?

From my point of view I continue to found meaningful the name "geometry_overlay" even if it use the used word "geometry"
"spatial_join" could be more self explaining for data extraction but not for spatial test and in this case I would put beside "spatial_test" for boolean checking. But It would be a coming back an outdated solution, instead of the smart solution of null expression for boolean result.

At last I would leave all as is. 


> 
> Here are some alternatives :
> 
> ```
> geometry_overlay_intersects(...)
> spatial_join_intersects(...)
> spatial_operation_intersects(...)
> so_intersects(...)
> ```
> 
> It would also be possible to have just one function (except nearest, that has a different signature), and provide the operator as an argument :

Why not add anyway the generic function with a spatial predicate parameter?

> 
> ```
> geometry_overlay('intersects', ...)
> spatial_join('intersects', ...)
> spatial_operation('intersects', ...)
> ```
> 
> Things to keep in mind :
> 
>     * possible confusion with plain spatial operators functions (such as `intersects(...)`)
> 
>     * separate functions are more auto-completion friendly, and make it easier to discover the functions to the user
> 
>     * separate functions overcrowds the function list, is more verbose in the code and makes it less clear to the user that all these functions work in the same way.

nyalldawson:
I have **grave** concerns with the approaches taken here. Specifically,

- Creating a spatial index within the expression is extremely costly. E.g. for a case where you are comparing against a large layer (say, a 1 million point postgis layer), then the first expression evaluation will require a complete traversal of that entire layer before any results can be obtained. A small map pan will then require the complete traversal and index rebuild again, because expression contexts are not persistent between map renders (by design). So for anything but a trivially small layer, the performance will be very poor, even if you're only ever comparing a single polygon source feature against this large point layer. This spatial index is also storing the feature geometries, so for a large layer that's a lot of memory use.

- Even worse, it seems that there's multiple complete layer traversals occurring here in some circumstances (when the "materialize" function is called on the target layer. When this happens, you're doing a complete layer traversal of the target layer and storing all the results in a memory layer. Again, for large layers this is extremely costly both in processing time and memory usage.

I'd suggest to rethink this approach and discard the creation of the spatial index entirely. The approach taken by the processing algorithms which do a similar task is to rely on the backend spatial index, and send off individual feature requests to the provider for the provider to optimise using the source's existing spatial indices. This works really well, and avoids the need for QGIS to ever even know about target features which fall outside of the desired intersection extent. Have a look at qgsalgorithmextractbylocation.cpp for a template in how to do this.

I also think there's a lot of thread safety issues in this PR. It's not immediately obvious where it's used (maybe it's a remnant), but that `Q_DECLARE_METATYPE( std::shared_ptr<QgsVectorLayer> )` line makes me extremely nervous. Layers should never be 
accessed from background threads, and storing them in a shared_ptr doesn't solve this.

m-kuhn:
@olivierdalang @enricofer thanks a lot for your work, inputs and expertise to bring this to the current state!

@nyalldawson good questions (I'll take over since these parts have been completely started by me)

1. There is a tradeoff to be done here, a local cache/index vs. usage of remote index. A server remote index does make sense when we do have an index (often) and when roundtrips are cheap. On the other hand, this means *a query for each individual feature* so if we render 1000 features and have a query time of 0.1 seconds, this is 100 seconds already. Compare with e.g.  `get_feature()` calls, I have had people contacting me about poor performance with this multiple times in the past.
Ideally we would implement all that's required to send a "join" to the backend and let it do the hard work. But that's far away from trivial. What we could do is a `bool cached` parameter to toggle the local cache.

2. Completely agreed on expressions/background threads accessing vector layers. This could be improved by doing a `QgsExpressionUtils::getFeatureSource()` on which the materialize call is done, that would certainly help stability. On the other hand, the `shared_ptr` should be save, since the cache layer itself is only accessed from the thread, and the `shared_ptr` is there to connect its lifetime to the expression cache.

nyalldawson:
> There is a tradeoff to be done here, a local cache/index vs. usage of remote index. A server remote index does make sense when we do have an index (often) and when roundtrips are cheap. On the other hand, this means a query for each individual feature so if we render 1000 features and have a query time of 0.1 seconds, this is 100 seconds already. Compare with e.g. get_feature() calls, I have had people contacting me about poor performance with this multiple times in the past.
Ideally we would implement all that's required to send a "join" to the backend and let it do the hard work. But that's far away from trivial. What we could do is a bool cached parameter to toggle the local cache.

Hmm - I concede your point about the multiple requests... but I just can't see how a forced local spatial index, recreated at every map render, is a workable solution. At best it'll be slow, at worst it will exhaust all the available memory and just crash. 

The way I see it:
- forced local index: very poor performance for large target datasets, regardless of size of source dataset. good performance for large source dataset joining to small target dataset. Very high memory requirements, with storage of all target features in memory.
- multiple requests/remote index: poor performance for large source datasets. good performance for small source datasets joining to large (or very large) target datasets. Very low memory requirement -- everything is done via iterators. 

Regardless of the approach, that materialized layer has got to go, and be replaced by a transform on the fly during index population (this is possible via setting the target crs on the feature request used for the index population)

I think we need to be realistic and admit that these functions will only ever be usable for small datasets, and for large datasets/high performance requirements then you need to pre-process your data. I'd almost go so far as to put a hard limit on the features usuable by these expression functions, and then raise evaluation errors if its exceeded. Otherwise we're misleading users into thinking that we're officially supporting this use case.

Another thing to keep in mind: if this is being used for rendering maps, then fetching the whole remote dataset and storing in an index is overkill to the max :smile: . We'd need some way to set in the context the target extent, so that you're only ever fetching target features within the visible map area

m-kuhn:
> Hmm - I concede your point about the multiple requests... but I just can't see how a forced local spatial index, recreated at every map render, is a workable solution. At best it'll be slow, at worst it will exhaust all the available memory and just crash.

At best it will work perfectly fine. At worst it will crash. But since processing creates memory layers all the time without much trouble I wonder how often this is really the case.

> 
> The way I see it:
> 
>     * forced local index: very poor performance for large target datasets, regardless of size of source dataset. good performance for large source dataset joining to small target dataset. Very high memory requirements, with storage of all target features in memory.
> 
>     * multiple requests/remote index: poor performance for large source datasets. good performance for small source datasets joining to large (or very large) target datasets. Very low memory requirement -- everything is done via iterators.
> 

Sums it up nicely. What about a "cached" parameter. True -> use cache, False -> do not use cache, NULL (auto) -> use a heuristic based on reported feature count. I don't think we know how many times the expression is going to be evaluated? So we can only set a hard limit there, e.g. 10000 features to switch between cached and uncached.

> 
> Regardless of the approach, that materialized layer has got to go, and be replaced by a transform on the fly during index population (this is possible via setting the target crs on the feature request used for the index population)

It's used to cache attributes required later on to evaluate the expression for the results. It can't just go unless we really drop the whole caching - of which I'm still not convinced.

> 
> I think we need to be realistic and admit that these functions will only ever be usable for small datasets, and for large datasets/high performance requirements then you need to pre-process your data. I'd almost go so far as to put a hard limit on the features usuable by these expression functions, and then raise evaluation errors if its exceeded. Otherwise we're misleading users into thinking that we're officially supporting this use case.

In this case we can just stick to caching? ;-P

I think we are better off with a clear description of how things work in the function documentation. Let's not try to outsmart the users. This kind of behavior would so turn me off.

> 
> Another thing to keep in mind: if this is being used for rendering maps, then fetching the whole remote dataset and storing in an index is overkill to the max . We'd need some way to set in the context the target extent, so that you're only ever fetching target features within the visible map area

Not sure it's that easy with nearest neighbour, disjoint etc. Even the bounding box would need to be a param (which in some cases maybe could be automatically determined from the expression context). But I'd treat that as out of scope for this PR.

nyalldawson:
> At best it will work perfectly fine. At worst it will crash. But since processing creates memory layers all the time without much trouble I wonder how often this is really the case.

Has this been tested in that situation? Can you load a 10 million point + file and try joining against it?

> What about a "cached" parameter. True -> use cache, False -> do not use cache, NULL (auto) -> use a heuristic based on reported feature count. I don't think we know how many times the expression is going to be evaluated? So we can only set a hard limit there, e.g. 10000 features to switch between cached and uncached.

Maybe... but then it's starting to reek of a feature which is exposing too much of the qgis internals to users and expecting them to know what to do with all these knobs and settings. I personally **hate** features like that!

m-kuhn:
> > At best it will work perfectly fine. At worst it will crash. But since processing creates memory layers all the time without much trouble I wonder how often this is really the case.
> 
> Has this been tested in that situation? Can you load a 10 million point + file and try joining against it?

Nope, can do

> 
> > What about a "cached" parameter. True -> use cache, False -> do not use cache, NULL (auto) -> use a heuristic based on reported feature count. I don't think we know how many times the expression is going to be evaluated? So we can only set a hard limit there, e.g. 10000 features to switch between cached and uncached.
> 
> Maybe... but then it's starting to reek of a feature which is exposing too much of the qgis internals to users and expecting them to know what to do with all these knobs and settings. I personally **hate** features like that!

I would be more than happy to have a "one size suits all" solution. If we implement sql compiler join support, maybe we eventually get there. Until then I see situations where **either** approach will perform very, very badly and the only solution to that is to give the power to the user. I personally **hate** it when a system tries to outsmart me without any knob :stuck_out_tongue_winking_eye:

olivierdalang:
@m-kuhn  @Nyalldawson
Thanks for the great discussion !

I've run some benchmarks with various cases to see how it performs with large datasets, once with the code in it's current state, and once without the local spatialindex (just using `QgsFilterRequest::filterRect()`), and there are the exact trend that you guys foresaw:
- building a local spatial index is expensive when comparing few features against many features, so we'd rather use the native index (if any)
- for postgis, the database roundtrips start make a difference when there's many features in the base layer.
- there's indeed no approach is best all of the time, as there are just too many parameters (feature count, result/filtering expression evaluation cost, database roundtrip, etc...).

In general, I find it pretty usable still in real-time (symbology) even with quite a large number of features thanks to non-blocking rendering. It's more annoying in the field calculator, where the calculation is blocking and can't be aborted, leading to a functional freeze if the computation takes too long. But IMO that's a different and general QGIS issue.

One significant difference appears when zoomed in, as with the current code, the spatial index is rebuilt for the whole layer even if showing only a small subset of features. The difference can be seen clearly in the screencasts below.
Rebuilding the spatial index of the current index only would improve the situation but wouldn't be reliable (as intersection could happen outside of the current extent). 

I'd suggest :
- removing local spatial index/cache logic
- add a note in the function's help that a spatial index is required to get acceptable performance
- add a note in the function's help that for large layers and/or slow providers, it's better to do a proper spatial join
- open an issue about implementing proper cancellation of too-long running attribute calculations

That would help keep things simple both in the code and for the end user, and maybe to better profit of possible future optimisations. Also I think it's acceptable to expect a spatial index, as the layers are unusable anyways as soon as they get somewhat large without it, so even better if this leads to some users leaning about the need of spatial indexes.

For `geometry_overlay_nearest` however, I don't think we have any equivalent of `QgsSpatialIndex::nearestNeighbor(...)`. when relying on the provider's spatial index... For that case the current logic may be the only way. But it's different in logic anyways.

## benchmarks

100000 vs 100000 features with **native index**

native_index_100000

100000 vs 100000 features with **cached index**

local_index_100000

Full results :
[results-with-local-index.txt](https://github.com/qgis/QGIS/files/5150258/results-with-local-index.txt)
[results-with-native-index.txt](https://github.com/qgis/QGIS/files/5150259/results-with-native-index.txt)

olivierdalang:
And some more tests with and without postgres latency.
In this case, it's indeed a bit ridiculous with the native index.

100000 vs 100000 features with **native index** on postgis with no latency (local server)
native_index_postgis_100000

100000 vs 100000 features with **cached/local index** on postgis with no latency (local server)
local_index_postgis_100000

1000 vs 1000 features with **native index** on postgis with simulated latency (200ms)
native_index_postgis_highlatency_1000

1000 vs 1000 features with **cached/local index** on postgis with simulated latency (200ms)
local_index_postgis_highlatency_1000

m-kuhn:
Thanks for all the tests, @olivierdalang !

I think we have 3 options

 1. Only Uncached
   - [+] That's what's done already
   - [-] Horrible performance with latency
 2. Only Cached
   - [+] Good performance with latency
   - [-] That's nowhere done yet
   - [-] Horrible performance with huge datasets, low latency / zoomed in
 3. "Cached" as parameter
   - [+] Gives control to the user to decide what's best for a particular situation
   - [-] More complex code, that might eventually become legacy when caching is implemented on a different level

I would go for 3. but I can live with 1. as well if it's the general preference.

olivierdalang:
I've added a `cache` parameter that defaults to false, and documented it like this : 
```
set this to true to build a local spatial index (most of the time, this is unwanted, unless you are working with a particularly slow data provider)
```
@nyalldawson Does that work for you ? It's true that it stays a bit cryptic, but I'd rather avoid trying to auto-set this with a heuristic, as I think it's better to have a consistently slow feature than a randomly slow feature :-)

I also refactored the code to simplify a little bit.

I'm setting the PR to ready for merge.

olivierdalang:
Oops. spell check says `features from **an other** layer` is wrong, is it really ?! (not native speaker)

enricofer:
I'm testing the PR and I'm not having computational issue so far.
As largerly expected a not aware use of overlay functions even on not so heavy datasets could bring to slow down and freeze on rendering and on attributes table updates. In my opinion we should at least re-enable default memory materialization of target layers leaving only local spatial indexing as option. I would propose @olivierdalang to explore this feature extending performance tests to only memory layer materialization.

olivierdalang:
@enricofer Could you provide more detail about the case in which you experience the slow downs (what format, how many features, etc.) ? Or if possible attach it here ?

Is it maybe that the data has no spatial index ? In that case, it is expected to be really slow as soon as there's more than a few hundreds of features (to the point of being unusable).

About materializing without creating the spatial index, for the reason above, I think it would yield really poor results. Or did you mean the opposite ?

Actually, maybe we could just return an error if there's no spatial index (saying that the user must either add a spatial index, or set cache=true). Or in such a case, just force the cache parameter.

Also feature freeze being today, I suggest we merge in this state. I'll be avaible to do some further polishing/optimisation before the release.

m-kuhn:
I agree, let's merge this for wider testing.
There are a couple of scenarios where this is a/ very useful and b/ performant enough. We will *always* find situations where it doesn't live up to the users expectations. There are probably a couple of low hanging fruits to improve and others which require advanced background work.

M-Rick:
I tried these new functions in QGIS 3.16, but they don't seem to work. I am needing to keep the refFunction plugin for now.
I tried to collect an attribute from a line layer to an intersecting point layer. At the beginning I used those brand new features, but I had no results at all. I thought my layers contained errors. But I then tried with refFunction and I got the expected results.
With the following refFunction expression it does work out of the box.
`geomintersects('Line','Ref')`
But when I switch to the new overlay functions it doesn't work.
I tried this expression instead of refFunction and it doesn't return me any results.
`overlay_intersects('Line','Ref’) `


Version de QGIS | 3.16.0-Hannover | Révision du code | 4af1cbfb97
-- | -- | -- | --
Compilé avec Qt | 5.14.2 | Utilisant Qt | 5.14.2
Compilé avec GDAL/OGR | 3.1.2 | Utilisé avec GDAL/OGR | 3.1.2
Compilé avec GEOS | 3.8.1-CAPI-1.13.3 | Utilisé avec GEOS | 3.8.1-CAPI-1.13.3
Compilé avec SQLite | 3.31.1 | Fonctionne avec SQLite | 3.31.1
Version du client PostgreSQL | 12.3 | Version de SpatiaLite | 4.3.0a
Version de QWT | 6.1.4 | Version de QScintilla2 | 2.11.4
Compilé avec PROJ | 6.3.2 | Fonctionne avec PROJ | Rel. 6.3.2, May 1st, 2020
Version de l'OS | macOS 10.16
Extensions Python actives | QuickOSM; processing_js; location_lab; refFunctions; qgis2web; qfieldsync; SgmExpressionFunctions; Mergin; processing; db_manager; MetaSearch

olivierdalang:
@M-Rick Thanks for reporting your issue !

Could you please open a new issue instead, so we can better keep track of it ? Also don't forget to add data and exact steps to reproduce the issue. Ping me in the issue (`@olivierdalang`) so I'll get a notification and look into it.

DelazJ:
> overlay_intersects('Line','Ref’) 

This is not the correct syntax of the new function. It should be `overlay_intersects('Line', Ref)` assuming Ref is the field you want to collect data from. Not a QGIS issue IMHO.
See examples at https://docs.qgis.org/3.16/en/docs/user_manual/working_with_vector/functions_list.html#overlay-intersects Also keep in mind that it returns an array so either your destination field allows array type or you should add a conversion using array_to_string

gioman:
> > 
> 
> This is not the correct syntax of the new function. It should be `overlay_intersects('Line', Ref)` assuming Ref is the field you want to collect data from. Not a QGIS issue IMHO.

@DelazJ after a quick test it does not seems to work anyway (the equivalent from the plugin works).

DelazJ:
@gioman could you share the expression you tested and the context (ref comment above about array output), please?

pigreco:
from a quick test to me it works:

image

## OSGeo4W  win 10 - QGIS 3.16 Hannover

gioman:
> from a quick test to me it works:

@pigreco with double quotes preview works, but then values are not saved to the target table/column.

gioman:
> @gioman could you share the expression you tested and the context (ref comment above about array output), please?

@DelazJ here it goes

[exp_sample.zip](https://github.com/qgis/QGIS/files/5537031/exp_sample.zip)

pigreco:
@gioman 
> @pigreco with double quotes preview works, but then values are not saved to the target table/column.

it is normal because they are arrays, if you want to save them you have to define the new field as text and use the expression

```
array_to_string(
 overlay_intersects( 'Com01012020_g_WGS84',"comune")
 )
```
image

gioman:
> it is normal because they are arrays, if you want to save them you have to define the new field as text and use the expression

@pigreco ok I see. I little less convenient than the original function.

DelazJ:
> @pigreco with double quotes preview works,

@gioman Even without quotes, it works. The expression here is a field and as such can take double-quotes or not
image

gioman:
> @gioman Even without quotes, it works

@DelazJ I may be missing something, but it does not for me, unless using also "array_to_string" in the expression.

DelazJ:
@gioman I meant "no need of quotes to have the expression return results. Ie, `overlay_intersects( 'Com01012020_g_WGS84', "comune")` is the same as `overlay_intersects( 'Com01012020_g_WGS84', comune)`."
But yes, if you want to store in a string field, you will always need the array_to_string function.

gioman:
>  But yes, if you want to store in a string field, you will always need the array_to_string function.

actually I tried to extract a number and put it into a integer field, and it also do not work (also because extracting from polygons with points, so only 1 value is extracted).

m-kuhn:
Even if only 1 value is extracted it will be an array (`[1]`), it needs to be extracted with `array_first(overlay_intersects( 'Com01012020_g_WGS84', comune))`.

gioman:
> Even if only 1 value is extracted it will be an array (`[1]`), it needs to be extracted with `array_first(overlay_intersects( 'Com01012020_g_WGS84', comune))`.

ok. Should we add this in the function help?

DelazJ:
I was also about to propose to add an example that returns comma separated list, as string. I think it's more common than the array_first case, and covers your use case here

acid_simbolo

m-kuhn:
Improving the function help would be very nice :+1:

gioman:
> Improving the function help would be very nice +1

@m-kuhn after thinking about I'm not sure how is best to do that, after all is normal that any function that returns an array an array function must be used too... Maybe adding a general note if is not already there (can't find one, but docs are very large nowadays!)? @DelazJ

m-kuhn:
Maybe we could add it to the expression builder dialog, where we know it's an array (and we do some magic to convert it to a string already).

> Hint: this expression returns an array, <a href="link/to/the/expression_array_doc">learn how to use arrays</a>.

gioman:
> Maybe we could add it to the expression builder dialog, where we know it's an array (and we do some magic to convert it to a string already).

sounds good to me.

DelazJ:
@gioman array functions are referenced at https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/functions_list.html#array-functions.
Adding the note as suggested by @m-kuhn sounds a good idea. In The docs, we currently display a url to reach array functions from the overlay functions (end of https://docs.qgis.org/testing/en/docs/user_manual/working_with_vector/functions_list.html#overlay-contains eg). We will probably remove them once this note is in place to avoid redundancy.
What I actually was suggesting in my previous comment is adding in the help panel an example of the type ```array_to_string( overlay_intersects( 'regions', "name"))``` that returns a comma separated list of values (and not array, as most of the current examples)

PR #38408, opened 2020-08-21
Use zero to represent not applicable precision

### LINKED ISSUE #38360: Cannot commit changes to PostGIS layer with field calculator
One gets an error message when trying to save edits after adding a text column using the field calculator (using a PostGIS table). Here's how to reproduce the bug:

1. Create a new PostGIS table (for example with an id field and a geometry field).
2. Open the attribute table and click on the field calculator.
3. Using the field calculator, create a new field with field type "Text, unlimited length (text)" and add some text content (the same bug also occurs with char and varchar).
4. Try to save edits

The following warning will appear:

Could not commit changes to layer temp

Errors: SUCCESS: 1 attribute(s) added.
  ERROR: field with index 2 is not the same!
  Provider: postgres
  Storage: PostgreSQL database with PostGIS extension
  expected field: name=dfdd type=QString typeName=text len=-1 precision=0
  retrieved field: name=dfdd type=QString typeName=text len=-1 precision=-1

5. Try to save the edits a second time, no more warning

The bug only occurs when the field calculator creates a new field. One workaround consists in creating a new empty text field and then using the field calculator to add some content. 

QGIS versions: 3.14.15 and 3.10.9
OS: Windows 10

dminor:
The documentation for QgsField states that zero should be used for
an unset precision, but the database providers use -1 in some places
for that purpose.

Fixes #38360

## Description

It would have been less work to change the field editor to use -1 for unset values, but the documentation for QgsField is clear that 0 represents an unset value. I tested these changes manually using the field editor and spatialite and postgis, but I don't have access to DB2, MSSQL or Oracle to test the changes there.

### DIFF
--- a/src/providers/db2/qgsdb2provider.cpp
+++ b/src/providers/db2/qgsdb2provider.cpp
@@ -1646,22 +1646,22 @@ bool QgsDb2Provider::convertField( QgsField &field )
 
     case QVariant::DateTime:
       fieldType = QStringLiteral( "TIMESTAMP" );
-      fieldPrec = -1;
+      fieldPrec = 0;
       break;
 
     case QVariant::Date:
       fieldType = QStringLiteral( "DATE" );
-      fieldPrec = -1;
+      fieldPrec = 0;
       break;
 
     case QVariant::Time:
       fieldType = QStringLiteral( "TIME" );
-      fieldPrec = -1;
+      fieldPrec = 0;
       break;
 
     case QVariant::String:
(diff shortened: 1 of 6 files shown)


dminor:
@nyalldawson Hi, I was wondering if you could have a look at this or suggest another reviewer? If it's not suitable, please let me know, I'd rather move on to something new rather than let this linger. Thanks!

nyalldawson:
Let's give it a shot -- I think it's the right move

Issue #37843, opened 2020-07-16
Attribute table: Field filter fails when table column has a relation reference widget assigned


andreasneumann:
Tested on both Linux and Windows in current master and 3.10. In version 3.4 this used to work.

If you assign a relation reference widget to an attribute table column, and try to use the field filter, it fails to work. If you change the widget e.g. to a value relation or text input widget, then the field filter seems to work.

![image](https://user-images.githubusercontent.com/884476/87658888-0fd93300-c75d-11ea-8b70-2bf1cafec0db.png)

andreasneumann:
Addition: even in select/filter features using form search mode (Ctrl-F), the filter fails to work as soon as a relation reference widget is assigned to a column that should be searched.

andreasneumann:
Correction: in Ctrl-F mode it seems to work "sort of".

The user has to select a value from the relation reference widget and then has to use in addition the "Equal To" filter mode, whereas with other widgets (e.g. value relation), the "Equal To" is selected automatically.

![image](https://user-images.githubusercontent.com/884476/87660390-43b55800-c75f-11ea-8649-330deb0f3dac.png)

PR #38416, opened 2020-08-21
Allow Null in Relation Reference Search Widget

signedav:
Having NULL available in the combo box on a field filter, getting back all features containing no FK.

Same in the _Select / Filter Feature Form (Ctrl+F)_ - means a NULL + Equal To is like the Flag "IS NULL".

Additionally in the _Select / Filter Feature Form (Ctrl+F)_ the "Equal To" flag is chosen as default as soon as the value changed (like in the Value Relation Widget) like suggested here  #37843

### DIFF
--- a/src/gui/editorwidgets/qgsrelationreferencesearchwidgetwrapper.cpp
+++ b/src/gui/editorwidgets/qgsrelationreferencesearchwidgetwrapper.cpp
@@ -101,9 +101,17 @@ QString QgsRelationReferenceSearchWidgetWrapper::createExpression( QgsSearchWidg
     case QVariant::ULongLong:
     {
       if ( flags & EqualTo )
+      {
+        if ( v.isNull() )
+          return fieldName + " IS NULL";
         return fieldName + '=' + v.toString();
+      }
       else if ( flags & NotEqualTo )
+      {
+        if ( v.isNull() )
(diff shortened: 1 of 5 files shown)


signedav:
Hi @nyalldawson 
Are there any objections in supporting the null in the search? Otherwise is it okay for you to merge this?
Thanks a lot.

nyalldawson:
@signedav 
> When you introduced it, you passed false for the AllowNULL value here

From memory, getting this to work with relation widget was out of scope at the time and it was easier just to disable it then deal with the bunch of required improvements which would allow null handling for the widget which were missing at the time.

If you've tested and this works OK, then let's merge!

src/gui/editorwidgets/qgsrelationreferencesearchwidgetwrapper.cpp
```diff
@@ -96,9 +101,17 @@ QString QgsRelationReferenceSearchWidgetWrapper::createExpression( QgsSearchWidg
     case QVariant::ULongLong:
     {
       if ( flags & EqualTo )
+      {
+        if ( v.isNull() )
```
nyalldawson, line 105: This would be good to add tests for

nyalldawson (approved):

signedav:
Thanks @nyalldawson

PR #38430, opened 2020-08-22
[symbology] Improve rendering of selected SVG markers

nirvn:
## Description

A picture worth a thousand words:
image

### DIFF
--- a/python/core/auto_generated/symbology/qgsmarkersymbollayer.sip.in
+++ b/python/core/auto_generated/symbology/qgsmarkersymbollayer.sip.in
@@ -694,6 +694,7 @@ Calculates the marker aspect ratio between width and height.
 
 
 
+
 };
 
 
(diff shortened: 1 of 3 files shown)


nyalldawson (commented):
One thing that I don't think is taken into account here is data defined paths. If I had marker with the default path having fill param, but a couple of features having an override to a non-parameterised svg, then by my understanding of this change we won't see the selected state for those features.

I'd suggest we handle this simply by detecting if data defined path is set, and if so, then just use the old approach of the yellow highlight box...

nirvn:
What about rasterizing & using the image operation when data defined path is detected? I'd prefer that over the rectangle.

nyalldawson:
That works too!

nirvn:
@nyalldawson , comments addressed.

nyalldawson (approved):
Let's give it a try!

PR #38437, opened 2020-08-24
[processing] Add a cluster size attribute to the dbscan cluster algorithm

nirvn:
## Description

This PR adds a new cluster size attribute attached to output features of the DBSCAN cluster algorithm. The value represents the number of features within an identified cluster. Useful attribute in many cases.

### DIFF
--- a/src/analysis/processing/qgsalgorithmdbscanclustering.cpp
+++ b/src/analysis/processing/qgsalgorithmdbscanclustering.cpp
@@ -69,6 +69,10 @@ void QgsDbscanClusteringAlgorithm::initAlgorithm( const QVariantMap & )
                         QObject::tr( "Cluster field name" ), QStringLiteral( "CLUSTER_ID" ) );
   fieldNameParam->setFlags( fieldNameParam->flags() | QgsProcessingParameterDefinition::FlagAdvanced );
   addParameter( fieldNameParam.release() );
+  auto sizeFieldNameParam = qgis::make_unique<QgsProcessingParameterString>( QStringLiteral( "SIZE_FIELD_NAME" ),
+                            QObject::tr( "Cluster size field name" ), QStringLiteral( "CLUSTER_SIZE" ) );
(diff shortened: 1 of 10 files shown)


src/analysis/processing/qgsalgorithmdbscanclustering.cpp
```diff
@@ -133,11 +139,13 @@ QVariantMap QgsDbscanClusteringAlgorithm::processAlgorithm( const QVariantMap &p
   feedback->pushInfo( QObject::tr( "Analysing clusters" ) );
   std::unordered_map< QgsFeatureId, int> idToCluster;
   idToCluster.reserve( index.size() );
+  std::unordered_map< int, int> clusterSize;
   QgsFeatureIterator features = source->getFeatures( QgsFeatureRequest().setNoAttributes() );
   const long featureCount = source->featureCount();
+  dbscan( minSize, eps, borderPointsAreNoise, featureCount, features, index, idToCluster, clusterSize, feedback );
```
nyalldawson, line 145: passing clusterSize here shouldn't be required -- dbscan doesn't do anything useful with it

nirvn: I'm passing it on so I can initiate the map with all cluster IDs and a count of 0. That allows for a nice one line loop to compute the  cluster size afterwards. You prefer if I'd go at it another way?

nyalldawson: It shouldn't be needed -- the call to ` clusterSize[ idCluster.second ]` would automatically add the missing id anyway and initialize it to 0.

nirvn: TIL moment, thanks

nirvn:
@nyalldawson , comments addressed.

nyalldawson (approved):

PR #38442, opened 2020-08-25
Add square brackets around the saved expression

om-henners:
## Description

When saving a raster expression the generated expression didn't contain square brackets around the letters used for band identifiers. Sticking with the NDVI example:

```
("NIR@1" - "Red@1") / ("NIR@1" + "Red@1")
```

becomes

```
(a - b) / (a + b)
```

Due to the way the expression is parsed these would not be interactie parameters for the user to set as the parameters requre square brackets around the layer tags. This change simply updates the string replacement to include the square brackets so you would get instead:

```
([a] - [b]) / ([a] + [b])
```

### DIFF
--- a/python/plugins/processing/algs/qgis/ui/RasterCalculatorWidgets.py
+++ b/python/plugins/processing/algs/qgis/ui/RasterCalculatorWidgets.py
@@ -188,7 +188,7 @@ class ExpressionWidget(BASE, WIDGET):
         used = [v for v in self.options.values() if v in exp]
 
         for i, v in enumerate(used):
-            exp = exp.replace(v, chr(97 + i))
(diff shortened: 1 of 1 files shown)


m-kuhn:
@luipir @volaya @nyalldawson what's your opinion on this? looks like a super easy fix, but I'm not comfortable with this area of the code to be confident enough for a click onto merge.

nyalldawson (approved):

luipir:
> @luipir @volaya @nyalldawson what's your opinion on this? looks like a super easy fix, but I'm not comfortable with this area of the code to be confident enough for a click onto merge.

sorry for my late... I read only now!

PR #38450, opened 2020-08-25
[processing] Add -vv parameter to Gdal Viewshed algorithm

ghtmtt:
## Description

Adds the `-vv` parameter to the viewshed algorithm of GDAL. It can be added to the `Additional command line parameters` but I think that the default 255 value returned from GDAL can be a little bit confusing to normal users that expects a binary raster as 0=notVisible and 1=Visible.


nyalldawson:
Can you update the tests too please?

rouault:
This pull request should be rebased. It includes unrelated commits

PR #38461, opened 2020-08-26
List availabale GDAL vector drivers (fix #18738)

### LINKED ISSUE #18738: List of available OGR drivers
Author Name: **Andre Joost** (Andre Joost)
Original Redmine Issue: [10313](https://issues.qgis.org/issues/10313)

Redmine category:data_provider

---
Under @Settings -> Options -> GDAL@ is a list of all installed raster drivers, which can be unchecked if necessary.

I suggest to create a similar list for the installed OGR vector drivers. Many people do not know the difference between GDAL and OGR, and the list does not tell them that only raster drivers are listed. 

It would be nice to see whether exotic drivers like ESRI FileGDB are installed properly. 

---

Related issue(s): #20296 (relates)
Redmine related issue(s): [12101](https://issues.qgis.org/issues/12101)

---

alexbruy:
## Description
In QGIS settings there is a GDAL tab with list of available GDAL raster drivers and their capabilities. If several drivers can be used to open data, user can disable some drivers to make sure that data will be opened with required driver. Similar situation with vector formats. For example there are FileGDB and OpenFileGDB drivers to work with .gdb files, but if data file has version 9.x, FileGDB won't be able to open it. But there are no UI with list of available vector drivers.

This pull-requst adds "Vector Drivers" tab to the GDAL section in the QGIS options dialog. At this tab user can see all available vector drivers and their capabilities as well as uncheck drivers which should not be used to open vector data.

Raster drivers
raster

Vector drivers
vector

Fixes #18738.

### DIFF
--- a/src/app/qgsoptions.cpp
+++ b/src/app/qgsoptions.cpp
@@ -99,7 +99,7 @@ QgsOptions::QgsOptions( QWidget *parent, Qt::WindowFlags fl, const QList<QgsOpti
   setupUi( this );
   connect( cbxProjectDefaultNew, &QCheckBox::toggled, this, &QgsOptions::cbxProjectDefaultNew_toggled );
   connect( leLayerGlobalCrs, &QgsProjectionSelectionWidget::crsChanged, this, &QgsOptions::leLayerGlobalCrs_crsChanged );
-  connect( lstGdalDrivers, &QTreeWidget::itemDoubleClicked, this, &QgsOptions::lstGdalDrivers_itemDoubleClicked );
+  connect( lstRasterDrivers, &QTreeWidget::itemDoubleClicked, this, &QgsOptions::lstRasterDrivers_itemDoubleClicked );
   connect( mProjectOnLaunchCmbBx, static_cast<void ( QComboBox::* )( int )>( &QComboBox::currentIndexChanged ), this, &QgsOptions::mProjectOnLaunchCmbBx_currentIndexChanged );
   connect( spinFontSize, static_cast < void ( QSpinBox::* )( int ) > ( &QSpinBox::valueChanged ), this, &QgsOptions::spinFontSize_valueChanged );
   connect( mFontFamilyRadioQt, &QRadioButton::released, this, &QgsOptions::mFontFamilyRadioQt_released );
@@ -1890,7 +1890,7 @@ void QgsOptions::leLayerGlobalCrs_crsChanged( const QgsCoordinateReferenceSystem
   mLayerDefaultCrs = crs;
 }
 
-void QgsOptions::lstGdalDrivers_itemDoubleClicked( QTreeWidgetItem *item, int column )
+void QgsOptions::lstRasterDrivers_itemDoubleClicked( QTreeWidgetItem *item, int column )
 {
   Q_UNUSED( column )
   // edit driver if driver supports write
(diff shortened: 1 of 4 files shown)


SrNetoChan:
@alexbruy Once this PR gets merged, please assign the documentation ticket to me :-) thanks

src/app/qgsoptions.cpp
```diff
@@ -2259,6 +2259,7 @@ void QgsOptions::loadGdalDriverList()
   QStringList myDrivers;
   QStringList myGdalWriteDrivers;
   QMap<QString, QString> myDriversFlags, myDriversExt, myDriversLongName;
+  QMap<QString, bool> driversType; // true for raster, false for vector
```
nyalldawson, line 2262: Would be slightly nicer if this were QMap< QString, QgsMapLayerType >, but that's not a blocker!

src/app/qgsoptions.cpp
```diff
@@ -2396,6 +2429,31 @@ void QgsOptions::saveGdalDriverList()
       }
     }
   }
+
+  // vector drivers
+  for ( int i = 0; i < lstVectorDrivers->topLevelItemCount(); i++ )
```
nyalldawson, line 2434: you could put the existing raster loop in a lambda and then re-use it here to avoid the duplicate code

src/core/qgsapplication.cpp
```diff
@@ -1687,7 +1687,7 @@ void QgsApplication::registerGdalDriversFromSettings()
   QStringList myList;
   if ( !joinedList.isEmpty() )
   {
-    myList = joinedList.split( ' ' );
+    myList = joinedList.split( QStringLiteral( "," ) );
```
nyalldawson, line 1690: This will break existing configuration, right? Probably safer to change the setting key name, and use the old value with the space delimiter only if the newer (comma separated) key doesn't exist

src/core/qgsapplication.cpp
```diff
@@ -1707,7 +1707,7 @@ void QgsApplication::applyGdalSkippedDrivers()
     if ( !sDeferredSkippedGdalDrivers()->contains( driverName ) )
       realDisabledDriverList << driverName;
   }
-  QString myDriverList = realDisabledDriverList.join( ' ' );
+  QString myDriverList = realDisabledDriverList.join( QStringLiteral( "," ) );
```
nyalldawson, line 1710:
```suggestion
  QString myDriverList = realDisabledDriverList.join( ',' );
```

nyalldawson:
Great work -- this will be very useful!!

src/app/qgsoptions.cpp
alexbruy: Changed to use `QgsMapLayerType`.

src/app/qgsoptions.cpp
alexbruy: Should be fixed now. But not sure I did everything correctly.

src/core/qgsapplication.cpp
alexbruy: Yes, you are right. Fixed according to your suggestion.

src/core/qgsapplication.cpp
alexbruy: Fixed.

nyalldawson (approved):

PR #38464, opened 2020-08-26
Trust layer metadata propagation

rldhont:
## Description

The trust flag at the project level is only used to read vector layer extent from xml, not from provider.

This flag was not available at the vector layer and data provider level.

We propose a new QgsMapLayer reading flag to propagate the trust layer metadata prohect's read flag and a new provider flag to trust datasource config. 

Trusting the datasource config means that the provider can use estimated metadata, the primary key is unique and the detectable geometry type and srid are the same as the requested.

### DIFF
--- a/src/core/qgsdataprovider.h
+++ b/src/core/qgsdataprovider.h
@@ -103,15 +103,32 @@ class CORE_EXPORT QgsDataProvider : public QObject
      */
     struct ProviderOptions
     {
+
+      /**
+       * Coordinate transform context
+       */
       QgsCoordinateTransformContext transformContext;
+
     };
+
+    /**
+     * Flags which control dataprovider construction.
+     * \since QGIS 3.16
+     */
+    enum ReadFlag
+    {
+      FlagTrustDataSource = 1 << 0, //!< Trust datasource config (primary key unicity, geometry type and srid, etc). Improves provider load time by skipping expensive checks like primary key unicity, geometry type and srid and by using estimated metadata on data load. Since QGIS 3.16
+    };
+    Q_DECLARE_FLAGS( ReadFlags, ReadFlag )
 
     /**
      * Create a new dataprovider with the specified in the \a uri.
      *
-     * Additional creation options are specified within the \a options value.
+     * Additional creation options are specified within the \a options value and since QGIS 3.16 creation flags are specified within the \a flags value.
      */
-    QgsDataProvider( const QString &uri = QString(), const QgsDataProvider::ProviderOptions &providerOptions = QgsDataProvider::ProviderOptions() );
+    QgsDataProvider( const QString &uri = QString(),
+                     const QgsDataProvider::ProviderOptions &providerOptions = QgsDataProvider::ProviderOptions(),
+                     QgsDataProvider::ReadFlags flags = QgsDataProvider::ReadFlags() );
 
     /**
      * Returns the coordinate system for the data source.
@@ -628,6 +645,9 @@ class CORE_EXPORT QgsDataProvider : public QObject
     //! Sets error message
     void setError( const QgsError &error ) { mError = error;}
 
+    //! Read flags. It's up to the subclass to respect these when needed
+    QgsDataProvider::ReadFlags mReadFlags = QgsDataProvider::ReadFlags();
+
   private:
 
     /**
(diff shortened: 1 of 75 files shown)


rldhont:
@mdouchin @dmarteau Can you take a look at the code ? thanks

src/core/qgsdataprovider.h
```diff
@@ -104,6 +104,7 @@ class CORE_EXPORT QgsDataProvider : public QObject
     struct ProviderOptions
     {
       QgsCoordinateTransformContext transformContext;
+      bool trustDatasourceConfig; //!< Trust datasource config (primary key unicity, geometry type and srid, etc). Improves provider load time.
```
nyalldawson, line 107: It'd be more future-friendly if we also used a new flags setup here instead of bools. (Don't forget the "since" annotation too!)

src/core/qgsmaplayer.h
```diff
@@ -553,6 +553,7 @@ class CORE_EXPORT QgsMapLayer : public QObject
     enum ReadFlag
     {
       FlagDontResolveLayers = 1 << 0, //!< Don't resolve layer paths or create data providers for layers.
+      FlagTrustLayerMetadata = 1 << 1, //!< Trust layer metadata. Improves layer load time.
```
nyalldawson, line 556:
```suggestion
      FlagTrustLayerMetadata = 1 << 1, //!< Trust layer metadata. Improves layer load time by skipping expensive checks like ######### on layer load. Since QGIS 3.16
```

src/core/raster/qgsrasterlayer.cpp
```diff
@@ -125,7 +125,7 @@ QgsRasterLayer::QgsRasterLayer( const QString &uri,
   QgsDebugMsgLevel( QStringLiteral( "Entered" ), 4 );
   setProviderType( providerKey );
 
-  QgsDataProvider::ProviderOptions providerOptions { options.transformContext };
+  QgsDataProvider::ProviderOptions providerOptions { options.transformContext, ( mReadFlags & QgsMapLayer::FlagTrustLayerMetadata ) ? true : false };
```
nyalldawson, line 128: This is a fragile way of constructing ProviderOptions -- if someone reorganised the members in the future then this code would break. Better to make an explicit constructor for ProviderOptions so that this is using stable api.

src/core/qgsdataprovider.h
```diff
@@ -628,18 +629,18 @@ class CORE_EXPORT QgsDataProvider : public QObject
     //! Sets error message
     void setError( const QgsError &error ) { mError = error;}
 
-  private:
```
nyalldawson, line 631: This change doesn't seem required -- can you revert?

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -1733,7 +1734,7 @@ void QgsPostgresProvider::determinePrimaryKeyFromUriKeyColumn()
     if ( !mPrimaryKeyAttrs.isEmpty() )
     {
       bool unique = true;
-      if ( mCheckPrimaryKeyUnicity )
+      if ( mCheckPrimaryKeyUnicity || mOptions.trustDatasourceConfig )
```
mdouchin, line 1737: If I understand correctly, I would have used `|| !mOptions.trustDatasourceConfig` so that we do not check unicity if we trust the datasource ?

rldhont, line 1737:
```suggestion
      if ( mCheckPrimaryKeyUnicity || !mOptions.trustDatasourceConfig )
```

rldhont: right, fix

src/core/qgsdataprovider.h
rldhont: I have removed my changes on `struct ProviderOptions` and added `ReadFlags`

src/core/raster/qgsrasterlayer.cpp
rldhont: I have removed my changes on `struct ProviderOptions` and added `ReadFlags`

src/core/qgsdataprovider.h
rldhont: I have fixed it by created `mReadFlags`

rldhont:
Hi @nyalldawson, what do you thonk about ths flags implémentation ?

nyalldawson (approved):
Looks good to me, nice work!

rldhont:
Thanks @nyalldawson

rldhont:
@pblottiere or @elpaso do you agree to merge ?

pblottiere (approved):

rldhont:
tahnsk @pblottiere @elpaso -> Speed up QGIS Server ;-)

pblottiere:
> tahnsk @pblottiere @elpaso -> Speed up QGIS Server ;-)

Thanks to you @rldhont, very good work :+1:

PR #38475, opened 2020-08-27
Server: Enhance WMS Layer style restorer for SLD

rldhont:
## Description

The WMS Layer restorer stored the QGIS style as XML document to restored it after SLD applying.

To speed up QGIS Server WMS request, the SLD will be applied in a dedicated style so it will be not necessary to store the QGIS Style as XML and restore it.

### DIFF
--- a/src/server/services/wms/qgswmsrenderer.cpp
+++ b/src/server/services/wms/qgswmsrenderer.cpp
@@ -3182,8 +3182,17 @@ namespace QgsWms
   void QgsRenderer::setLayerSld( QgsMapLayer *layer, const QDomElement &sld ) const
   {
     QString err;
+    // Defined sld style name
+    const QStringList styles = layer->styleManager()->styles();
+    QString sldStyleName = "__sld_style";
+    while ( styles.contains( sldStyleName ) )
+    {
+      sldStyleName.append( '@' );
+    }
+    layer->styleManager()->addStyleFromLayer( sldStyleName );
+    layer->styleManager()->setCurrentStyle( sldStyleName );
     layer->readSld( sld, err );
-    layer->setCustomProperty( "readSLD", true );
+    layer->setCustomProperty( "sldStyleName", sldStyleName );
   }
 
   QgsLegendSettings QgsRenderer::legendSettings() const
--- a/src/server/services/wms/qgswmsrestorer.cpp
+++ b/src/server/services/wms/qgswmsrestorer.cpp
@@ -33,20 +33,6 @@ QgsLayerRestorer::QgsLayerRestorer( const QList<QgsMapLayer *> &layers )
 
     settings.mNamedStyle = layer->styleManager()->currentStyle();
 
-    // set a custom property allowing to keep in memory if a SLD file has
-    // been loaded for rendering
-    layer->setCustomProperty( "readSLD", false );
-
-    QString errMsg;
-    QDomDocument styleDoc( QStringLiteral( "style" ) );
-    QDomElement styleXml = styleDoc.createElement( QStringLiteral( "style" ) );
-    styleDoc.appendChild( styleXml );
-    if ( !layer->writeStyle( styleXml, styleDoc, errMsg, QgsReadWriteContext() ) )
-    {
-      QgsMessageLog::logMessage( QStringLiteral( "QGIS Style has not been added to layer restorer for layer %1: %2" ).arg( layer->name(), errMsg ) );
-    }
-    ( void )settings.mQgisStyle.setContent( styleDoc.toString() );
-
     switch ( layer->type() )
     {
       case QgsMapLayerType::VectorLayer:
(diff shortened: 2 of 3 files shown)


rldhont:
@elpaso @pblottiere @dmarteau do you think, this can be backported ?

src/server/services/wms/qgswmsrestorer.cpp
```diff
@@ -94,13 +84,9 @@ QgsLayerRestorer::~QgsLayerRestorer()
...
     if ( layer->customProperty( "readSLD", false ).toBool() )
     {
-      QString errMsg;
-      QDomElement root = settings.mQgisStyle.documentElement();
-      QgsReadWriteContext context = QgsReadWriteContext();
-      if ( !layer->readStyle( root, errMsg, context ) )
-      {
-        QgsMessageLog::logMessage( QStringLiteral( "QGIS Style has not been read from layer restorer for layer %1: %2" ).arg( layer->name(), errMsg ) );
-      }
+      QString sldStyleName = layer->customProperty( "sldStyleName" ).toString();
```
pblottiere, line 87: @rldhont What do you think about using only one custom property? Because in this case I think we can remove the `readSLD` bool property and directly check `sldStyleName` with something like:

``` cpp
const QString prop = layer->customProperty( "sldStyleName", "" ).toString();
if ( ! prop.isEmpty() )
{
    ...
}
```

rldhont: Thanks for the review. I will propose some code to remove `readSLD`.

rldhont: @pblottiere done

src/server/services/wms/qgswmsrestorer.cpp
```diff
@@ -92,17 +78,12 @@ QgsLayerRestorer::~QgsLayerRestorer()
...
+    if ( !sldStyleName.isEmpty() )
     {
-      QString errMsg;
-      QDomElement root = settings.mQgisStyle.documentElement();
-      QgsReadWriteContext context = QgsReadWriteContext();
-      if ( !layer->readStyle( root, errMsg, context ) )
-      {
-        QgsMessageLog::logMessage( QStringLiteral( "QGIS Style has not been read from layer restorer for layer %1: %2" ).arg( layer->name(), errMsg ) );
-      }
+      layer->styleManager()->removeStyle( sldStyleName );
```
pblottiere, line 84: I don't see an explicit style restoring here. The default style is reestablished as soon as the current style is removed?

src/server/services/wms/qgswmsrenderer.cpp
```diff
@@ -3182,8 +3182,17 @@ namespace QgsWms
   void QgsRenderer::setLayerSld( QgsMapLayer *layer, const QDomElement &sld ) const
   {
     QString err;
+    // Defined sld style name
+    QStringList styles = layer->styleManager()->styles();
+    QString sldStyleName = "__sld_style";
+    while ( styles.contains( sldStyleName ) )
+    {
+      sldStyleName.append( '@' );
+    }
```
pblottiere, line 3191: @elpaso Do you know if we have some kind of random string generator in QGIS?

src/server/services/wms/qgswmsrenderer.cpp
```diff
@@ -3182,8 +3182,17 @@ namespace QgsWms
   void QgsRenderer::setLayerSld( QgsMapLayer *layer, const QDomElement &sld ) const
   {
     QString err;
+    // Defined sld style name
+    QStringList styles = layer->styleManager()->styles();
```
pblottiere, line 3186:
```suggestion
    const QStringList styles = layer->styleManager()->styles();
```

pblottiere:
Hi @rldhont,

> To speed up QGIS Server WMS request, the SLD will be applied in a dedicated style so it will be not necessary to store the QGIS Style as XML and restore it.

Do you have some metrics to share? I'm curious :).

src/server/services/wms/qgswmsrestorer.cpp
rldhont: The style is restored earlier new line 77 old line 91
```py
layer->styleManager()->setCurrentStyle( settings.mNamedStyle );
```

rldhont:
Here https://github.com/qgis/QGIS/pull/32342#issuecomment-578499400 @andreassteffens mentions

> Some profiling i did in order to identify a performance problem with QGIS Server revealed that saving the original SLD for all layers in the `QgsLayerRestorer` so it can be restored after rendering actually leads to poor performance for projects with many vector layers and/or complex symbology (take a project with ALKIS layers as created by the Norbit importer/plugin as example).
>
> Wouldn't it make more sense to pass the entire `QgsWmsRenderContext` to the `QgsLayerRestorer` and check whether an alternative SLD styling has been supplied for a layer in the request BEFORE serializing the layer's style to SLD for restoration?

rldhont:
@pblottiere @elpaso gentle ping for review and for backport ?

pblottiere:
Hi @rldhont,

> gentle ping for review and for backport ?

+1 to merge.

I'm +0 for the backport (it's not a bugfix nor a critical issue, but it shouldn't cause any trouble). So feel free if you need it.

pblottiere (approved):

rldhont:
@pblottiere thanks! I do not backport it. It isnot a bugfix.

PR #38484, opened 2020-08-28
Fix 'invisible news' issue : after hiding the news it never comes back

timlinux:
Should be backported to LTR too.

## Description

There is a bug in QGIS where if you go to settings and hide the news feed, then re-enable it in settings, you never see it displayed.  This is because the restore logic first sets it to 0 when hiding it then restores it to the last size (i.e. 0) when restoring it. I added logic so that restoring the news now sets the splitter to 50% of the height. The screencast below shows it now restoring properly.

qgis-news-resizing

As a second improvement we could also force the news panel to update after changing settings...something for another time....

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -12160,6 +12160,8 @@ void QgisApp::showOptionsDialog( QWidget *parent, const QString &currentPage, in
     double factor = mySettings.value( QStringLiteral( "qgis/magnifier_factor_default" ), 1.0 ).toDouble();
     mMagnifierWidget->setDefaultFactor( factor );
     mMagnifierWidget->updateMagnification( factor );
+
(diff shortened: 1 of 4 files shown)


timlinux:
OK wait I should have dealt with the restore geometry better, another commit inbound to fix that....

timlinux:
That is all I have for this PR - it moves the needle forward for the most irritating issue (news hidden never returns).....I'll leave figuring out how to show / hide without restarts for future (it is an edge case that is only useful if the news panel is actually on display when you enter into QgsOptions).

nyalldawson:
LGTM!

BTW -- your qgis is starting extremely slowly. It may be worth checking the dev tools startup profiler (via f12) and see if there's one particular plugin causing this

PR #38488, opened 2020-08-29
Server: Call QgsProject::setInstance on project passed as parameter

dmarteau:
## Description

Category: bugfix

The `QgsConfigCache::project` method implicitely pass the project to `QgsProject:setInstance`. 

However passing explicitely a project as parameter to `QgsServerHandler` bypass the config cache lookup and the project is not passed to `QgsProject:setInstance` as the current instance. 

Because of this, errors occurs on rendering, especially with the getPrint request:  see https://github.com/3liz/py-qgis-server/issues/18.

This PR introduce a fix by moving the  `setInstance` call outside of the `QgsConfigCache` and apply it to the current project  in a consisent way.

This issue prevent using the Qgis server python to be used correctly with instanciated QgisProject outside the config cache: I would strongly advise a backport to ltr (3.10) and release (3.14)

### DIFF
--- a/src/core/qgsproject.h
+++ b/src/core/qgsproject.h
@@ -1985,7 +1985,7 @@ class CORE_EXPORT QgsProject : public QObject, public QgsExpressionContextGenera
     friend class QgsProviderRegistry;
 
     // Required by QGIS Server for switching the current project instance
-    friend class QgsConfigCache;
+    friend class QgsServer;
 
     friend class TestQgsProject;
 };
--- a/src/server/qgsconfigcache.cpp
+++ b/src/server/qgsconfigcache.cpp
@@ -112,7 +112,6 @@ const QgsProject *QgsConfigCache::project( const QString &path, QgsServerSetting
         QStringLiteral( "Server" ), Qgis::Critical );
     }
   }
-  QgsProject::setInstance( mProjectCache[ path ] );
   return mProjectCache[ path ];
 
 }
--- a/src/server/qgsserver.cpp
+++ b/src/server/qgsserver.cpp
@@ -373,6 +373,9 @@ void QgsServer::handleRequest( QgsServerRequest &request, QgsServerResponse &res
         project = mConfigCache->project( configFilePath, sServerInterface->serverSettings() );
       }
 
+      // Set the current project instance
+      QgsProject::setInstance( const_cast<QgsProject *>( project ) );
+
       if ( project )
       {
         sServerInterface->setConfigFilePath( project->fileName() );


rldhont (approved):

pblottiere:
Hi @dmarteau,

From what I read [here](https://github.com/qgis/QGIS/blob/master/src/server/qgsserver.cpp#L373), the `QgsConfigCache::project` method is called when a project is passed to the `QgsServer::handleRequest`, therefore the `QgsProject:setInstance` method too.

I'm sure I missed something obvious so can you give more details please?

Thanks David :+1:

dmarteau:
Hi @pblottiere 

Nope, the test here https://github.com/qgis/QGIS/blob/master/src/server/qgsserver.cpp#L368 state that if the project is passed as parameters `QgsServer::handleRequest`, then the `QgsConfigCache::project` is *not* called. That's the purpose of it: you dont want/have to retrieve a project from cache if you pass it explicitely.

pblottiere:
> then the QgsConfigCache::project is not called

Yes you're right, I missed the `!` :).

pblottiere (approved):

elpaso:
This breaks long established workflows and deployed applications and makes it impossible to set the current project from Python from standalone paths that do not call `handleRequest`. 

It should have never been backported to a stable release!

And I could not even find a workaround: how is it now possible to fetch a project from the configCache and set it current? `QgsProject::setInstance` is not exposed to Python. 

The only alternative is now to re-load the project completely skipping the cache or am I missing something?

dmarteau:
Hi @elpaso 

 > This breaks long established workflows and deployed applications and makes it impossible to set the current project from Python from standalone paths that do not call handleRequest.

I see your point, the problem may arise if the QgConfigCache is used outside the server context, nevertheless, this is a  real bugfix since and, as yous noted, QgsProject::setInstance is not available in python,  it  makes passing project to  handleRequest useless.  

> The only alternative is now to re-load the project completely skipping the cache or am I missing something?

Unfortunately, I'm afraid this is the case with the actuel bugfix.
We may restore the  QgsProject::setInstance in the QgConfigCache for the next bugfix release and handling the specific case when the project is passed to `handleRequest`.

My strong feeling in the real problem come from the fact that QgsProject::setInstance is not exposed and should never been tied exclusively to the QgConfigCache.

nyalldawson:
This commit never should have been merged to 3.10 without waiting for the required one month delay. It was introduced **specifically** to avoid situations like this in the LTR branch. @rldhont please be more mindful of this in future!

PR #38502, opened 2020-08-31
Add "mainAnnotationLayer" to QgsProject 

nyalldawson:
This returns a reference to a main/default annotation layer for use by the project. It forms the default location to place new annotation items which should appear above all map layers.

All projects have a main annotation layer, which is treated as an internal layer and which isn't exposed to users via the layer tree or standard api for retrieving project layers. They are free to add ADDITIONAL annotation layers (e.g. to mix in annotation items between other map layers), but this main annotation layer will always be present and available for drawings


src/core/qgsproject.h
```diff
@@ -1154,6 +1156,19 @@ class CORE_EXPORT QgsProject : public QObject, public QgsExpressionContextGenera
...
+     *
+     * This layer is always present in projects, and will always be rendered
+     * above any other map layers during map render jobs.
+     *
+     * It forms the default location to place new annotation items which
+     * should appear above all map layers.
+     *
+     * \since QGIS 3.16
+     */
+    QgsAnnotationLayer *mainAnnotationLayer();
```
nirvn, line 1170: I think I'd just go for annotationLayer() and remove the main prefix.

src/gui/qgsmapcanvas.cpp
```diff
@@ -585,13 +586,19 @@ void QgsMapCanvas::refreshMap()
     mSettings.setLayerStyleOverrides( QgsProject::instance()->mapThemeCollection()->mapThemeStyleOverrides( mTheme ) );
   }
 
+  // render main annotation layer above all other layers
+  QgsMapSettings renderSettings = mSettings;
+  QList<QgsMapLayer *> allLayers = renderSettings.layers();
+  allLayers.insert( 0, QgsProject::instance()->mainAnnotationLayer() );
```
nirvn, line 592: We'll need to apply this to the save map [canvas] as image/PDF.

src/core/qgsproject.h
nyalldawson: I'm not so keen on that -- I wanted a clear distinction between this special layer and any "normal" annotation layers which may be present in the project (via the standard QgsProject map layer storage methods). 

An alternative could be "defaultAnnotationLayer", but I'd still like something which conveys the concept that it's "special" and works different to normal layers (because it's always drawn on top)

src/gui/qgsmapcanvas.cpp
nyalldawson: Good catch, will do

src/core/qgsproject.h
nirvn: persistentAnnotationLayer , topAnnotationLayer ? I just feel that the main prefix doesn't add much context here.

In any case, I see your point and I'll get over it if you merge as is :wink: go with what you think is best.

PR #38511, opened 2020-08-31
Enable spell test

m-kuhn:
## Description

Enables the lost spelling test


.github/workflows/code_layout.yml
```diff
@@ -120,7 +120,12 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Indentation Test
+      - name: Install Requirements
+        run: |
+          sudo apt install -y \
+              unbuffer \
```
m-kuhn, line 126:
```suggestion
              expect \
```

3nids:
It looks like the test will be run over all the files which will take ages.
It would be better to perform the test on modified files only

m-kuhn:
@3nids done

- what would you propose to fix the current error?
- what's required to make it show the correction suggestion?

3nids:
your last try sounds good, I think you have to merge to see it running.

m-kuhn:
It works, it already prints out an error inside the changed file 

```
Running spell check on files: .github/workflows/code_layout.yml 
170:        run: ./tests/code_layout/test_sipfiles_uptodate.sh
some errors have been found.
##[error]Process completed with exit code 1.
```

 - This needs to be fixed before merging (`# spellok`?)
 - Do you know if it is possible to get also the correction suggestions from the spell checker

3nids:
#spellok should have been good, I renamed the script instead.

PR #38521, opened 2020-09-01
Fix filename prefix is no longer shown when adding layers from a GPX file to QGIS

### LINKED ISSUE #37551: layer name does not show anymore when loading gpx datasources
With QGIS 3.12, when loading a gpx layer, the name of that gpx layer is displayed in the explorer.
I notice that with version 3.14, only the "type" of the layer (in the gpx sense): "track" or "Waypoint" is displayed. 

Avec QGIS 3.12, lorsque l'on charge une couche gpx, le nom de cette couche gpx s'affiche dans l'explorateur.
Je constate qu'avec la version 3.14, seul le "type" de la couche (au sens gpx) : "track" ou "Waypoint" s'affiche.

nyalldawson:
Fixes #37551

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -6023,6 +6023,8 @@ QList<QgsMapLayer *> QgisApp::askUserForOGRSublayers( QgsVectorLayer *layer, con
   if ( !chooseSublayersDialog.exec() )
     return result;
 
+  const bool addToGroup = chooseSublayersDialog.addToGroupCheckbox();
+
   QString name = layer->name();
 
   auto uriParts = QgsProviderRegistry::instance()->decodeUri(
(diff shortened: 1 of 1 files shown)


luipir (approved):

nyalldawson:
This regression was caused by 8ff229e33062432bf9ad34d0cf05e6dbda2adb2b, but there's no detail in that commit to describe why the logic was changed. @suricactus can you clarify why this was changed?

suricactus:
@nyalldawson  I don't recall precisely why. I guess as they might be in a group to remove the visuall clutter of the same name appearing multiple times in the legend. However, when they are not in a group this is causing more of a problem than fixing something. Thanks for the fix.

nyalldawson:
@suricactus so I guess the ideal logic would be to skip the filename if we've opted to place them in a group...

suricactus:
Yes, either this, or adding additional checkbox "Prefix with filename", so all the combinations are present to the user.

nirvn:
I'd be careful to add too many options here. I think only skipping the filename when we add layer(s) into a parent group is good behavior as is.

nyalldawson:
@suricactus ok, new logic is implemented --now we only add the filename prefix if the layers AREN'T being added to a group. Should be good to merge.

luipir (approved):

suricactus:
Nice! Thanks!

PR #38525, opened 2020-09-01
[mssql] Don't allow creating connection only when both Host and Provider/DSN are empty

### LINKED ISSUE #38197: Browser MSSQL add connection dialogue doesn't allow DSN without entering a Host name
Using the left pane browser, MSSQL to add a connection, the OK button is greyed out until a host is added. If using a DSN a host shouldn't be needed.

QGIS 14.1
W10-64

This seems like a regressive change, pretty sure it was working in 3.12(?)

agiudiceandrea:
## Description

Adds the possibility to create a connection when Provider/DSN parameter is specified and the Host parameter is empty.

See https://github.com/qgis/QGIS/pull/37467, https://github.com/qgis/QGIS/commit/89298373e0a0be4928f163dba6c1b187faca6e4d and https://github.com/qgis/QGIS/issues/38197#issuecomment-670849551

Fixes #38197

The issue addressed also affects 3.14 branch.

### DIFF
--- a/src/providers/mssql/qgsmssqlnewconnection.cpp
+++ b/src/providers/mssql/qgsmssqlnewconnection.cpp
@@ -41,6 +41,7 @@ QgsMssqlNewConnection::QgsMssqlNewConnection( QWidget *parent, const QString &co
 
   buttonBox->button( QDialogButtonBox::Ok )->setDisabled( true );
   connect( txtName, &QLineEdit::textChanged, this, &QgsMssqlNewConnection::updateOkButtonState );
+  connect( txtService, &QLineEdit::textChanged, this, &QgsMssqlNewConnection::updateOkButtonState );
   connect( txtHost, &QLineEdit::textChanged, this, &QgsMssqlNewConnection::updateOkButtonState );
   connect( listDatabase, &QListWidget::currentItemChanged, this, &QgsMssqlNewConnection::updateOkButtonState );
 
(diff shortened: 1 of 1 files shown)


src/providers/mssql/qgsmssqlnewconnection.cpp
```diff
@@ -252,6 +253,6 @@ void QgsMssqlNewConnection::showHelp()
 void QgsMssqlNewConnection::updateOkButtonState()
 {
   QListWidgetItem *item = listDatabase->currentItem();
-  bool enabled = !txtName->text().isEmpty() && !txtHost->text().isEmpty() && item;
+  bool enabled = !txtName->text().isEmpty() && !( txtService->text().isEmpty() && txtHost->text().isEmpty() ) && item;
```
nyalldawson, line 256: Minor thing, but I think this would be more readable if we flipped the logic:

```suggestion
  bool disabled = txtName->text().isEmpty() || ( txtService->text().isEmpty() && txtHost->text().isEmpty() ) || !item;
```

agiudiceandrea: Yes, indeed. Thanks for the suggestion. I've force pushed my branch with a new commit.

nyalldawson (approved):

nyalldawson:
Thanks!

PR #38530, opened 2020-09-01
Correctly render MapRendereJob use case steps in pyqgis doc

DelazJ:
~I don't know how it will render in the c++ api website (it's [already broken! ](https://qgis.org/api/classQgsMapRendererJob.html)) but it will probably make the pyqgis [doc more readable](https://qgis.org/pyqgis/master/core/QgsMapRendererJob.html#qgis.core.QgsMapRendererJob)~ should both be good now; i changed numbering system

### DIFF
--- a/src/core/qgsmaprendererjob.h
+++ b/src/core/qgsmaprendererjob.h
@@ -167,11 +167,11 @@ struct LabelRenderJob
  *
  * Common use case:
  *
- * # prepare QgsMapSettings with rendering configuration (extent, layer, map size, ...)
- * # create QgsMapRendererJob subclass with QgsMapSettings instance
- * # connect to job's finished() signal
- * # call start(). Map rendering will start in background, the function immediately returns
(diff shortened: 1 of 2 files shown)


src/core/qgsmaprendererjob.h
```diff
@@ -167,11 +167,11 @@ struct LabelRenderJob
  *
  * Common use case:
  *
- * # prepare QgsMapSettings with rendering configuration (extent, layer, map size, ...)
- * # create QgsMapRendererJob subclass with QgsMapSettings instance
- * # connect to job's finished() signal
- * # call start(). Map rendering will start in background, the function immediately returns
- * # at some point, slot connected to finished() signal is called, map rendering is done
+ * 1. prepare QgsMapSettings with rendering configuration (extent, layer, map size, ...)
```
nyalldawson, line 170: While you're at it, could you capitalise these sentences?

PR #38537, opened 2020-09-01
Correct bash command to run ccmake for Mac

imanhn:
The path argument to ccmake corrected and the need for sudo also notified.

## Description

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1205,7 +1205,10 @@ In a Terminal cd to the qgis source folder previously downloaded, then:
     cd ..
     mkdir build
     cd build
+```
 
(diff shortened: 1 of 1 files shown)


3nids:
The ccmake argument is a valid fix thanks.
But there is no requirement to run with sudo neither to wrap commands in a bash script.

imanhn:
sudo is needed because all the folders created by cmake in the `build` directory are not readonly for everyone and only open to write for system account (at least mine is like that). regarding to wrap it with a bash file you are right it is not necessary but if SUDO is required it would be a justified practice to wrap those multiple commands in a bash file.

PR #38555, opened 2020-09-02
[feature][gui] add "close all tabs" to QgsMessageLogViewer

olivierdalang:
## Description

This adds an "Empty tab and close others" actions to the log viewer panel context menu. This saves one frustrating click when debugging and you want to clear all messages.

image

### DIFF
--- a/src/gui/qgsmessagelogviewer.cpp
+++ b/src/gui/qgsmessagelogviewer.cpp
@@ -82,6 +82,19 @@ void QgsMessageLogViewer::showContextMenuForTabBar( QPoint point )
          );
   mTabBarContextMenu->addAction( actionCloseOtherTabs );
 
+  QAction *actionCloseAllTabs = new QAction( tr( "Close All Tabs" ), mTabBarContextMenu );
+  actionCloseAllTabs->setEnabled( tabWidget->tabBar()->count() > 0 );
+  connect( actionCloseAllTabs, &QAction::triggered, this, [this]
+  {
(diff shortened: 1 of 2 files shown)


nyalldawson:
@nirvn thoughts on the UX here?

nirvn:
I don't mind it, it can be useful. The wording of the action could be improved. Maybe "clear content and close other tabs"?

uclaros:
Why not 'Close All Tabs'?

nirvn:
@uclaros , because closing all tabs isn't allowed (so far). Closing last tab merely clears its content. You can try closing all tabs manually, you'll see :)

olivierdalang:
I guess the only reason we don't support closing all tabs is that QTabWidgets look confusing without tabs (looks like a text area, as it's all white). But we could just hide it if empty ?

In terms of UX it would be better (you can close all on any tab, so that on next message you only get tabs with new content)

olivierdalang:
I changed it to "Close All Tabs", and just hide the widget when empty. It reappears again with the next message.

closeall

m-kuhn:
@olivierdalang what do you think about a label saying "Message log empty" when all tabs are closed, I guess it can be opened manually still?

olivierdalang:
> @olivierdalang what do you think about a label saying "Message log empty" when all tabs are closed, I guess it can be opened manually still?

Done :

image

m-kuhn (approved):

PR #38560, opened 2020-09-02
Fix unreported server standalone POST data handling

elpaso:
The problem was with post data sent in different frames, first the headers are sent, then the body but `readyRead` was only called once.


elpaso:
@pblottiere I'd appreciate your opinion: I'm not really convinced by the approach.

PR #38582, opened 2020-09-03
Do not query geography_columns on postgis < 1.5

### LINKED ISSUE #38580: QGIS Postgres provider does not load any postgres 8.4 postgis 1.4.2 geometry table
I know it's and old version but this is a serious regression.

DB manager loads the tables correctly.

elpaso:
Fixes #38580

### DIFF
--- a/src/providers/postgres/qgspostgresconn.cpp
+++ b/src/providers/postgres/qgspostgresconn.cpp
@@ -488,7 +488,8 @@ bool QgsPostgresConn::getTableInfo( bool searchGeometryColumnsOnly, bool searchP
       dimName    = QStringLiteral( "l.coord_dimension" );
       gtableName = QStringLiteral( "geometry_columns" );
     }
-    else if ( i == SctGeography )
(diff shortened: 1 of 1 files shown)


src/providers/postgres/qgspostgresconn.cpp
```diff
@@ -488,7 +488,7 @@ bool QgsPostgresConn::getTableInfo( bool searchGeometryColumnsOnly, bool searchP
       dimName    = QStringLiteral( "l.coord_dimension" );
       gtableName = QStringLiteral( "geometry_columns" );
     }
-    else if ( i == SctGeography )
+    else if ( i == SctGeography && mPostgresqlVersion >= 150000 )
```
rouault, line 491: testing the postgres version looks wrong to me. Should rather be against mPostgisVersionMajor and mPostgisVersionMinor

elpaso: yes, right. Thanks for noticing.

PR #38584, opened 2020-09-03
Temporary fix for missed function OGR_Fld_GetAlternativeNameRef()

imanhn:
## Description

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1227,7 +1227,7 @@ After the initial Terminal configure, you can use ccmake to make further changes
 
 ```bash
     cd build
-    ccmake ..
+    ccmake ../QGIS
 ```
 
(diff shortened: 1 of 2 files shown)


rouault:
This fix is unneeded. You just need to update your GDAL 3.2.0dev copy to the latest state of GDAL master

imanhn:
Thanks for the response. Yes, I found the missed function in gdal/gdal/ogr/ogr_feature.h

So install guide (Section 4.1.2) should be updated, telling users/contributors (like me) to clone gdal-master and compile it instead of using OSGeo4W.exe.

Right?

rouault:
> compile it instead of using OSGeo4W.exe.

If you use the gdal-dev of OSGeo4W you must be ready to install the updated version regularly.

PR #38589, opened 2020-09-03
Increase max page number item property to 9999 in print layout

### LINKED ISSUE #38585: Limitation of max. number of pages in Print Layouts.
There seems to be an indirect limitation to a maximum of 99 pages in a print layout.
For sure you can add more than 99 pages in a single print layout, and also place items (labels, maps, etc.) on pages >99.
But: in Position & Size Panel, it always shows this item is on page 99, but in reality it is on page >99. When changing the Position and Size, the item automatically jumps to page 99.

agiudiceandrea:
## Description

Adds the "maximum" property to mPageSpinBox in qgslayoutitemwidgetbase.ui and sets it to to 9999.

like for mPageSpin in [qgslayoutnewitemproperties.ui](https://github.com/qgis/QGIS/blob/master/src/ui/layout/qgslayoutnewitemproperties.ui#L378-L391)

Fixes https://github.com/qgis/QGIS/issues/38585

The issue also affects the 3.10 and the 3.14 branches.

### DIFF
--- a/src/ui/layout/qgslayoutitemwidgetbase.ui
+++ b/src/ui/layout/qgslayoutitemwidgetbase.ui
@@ -176,6 +176,9 @@
           <property name="minimum">
            <number>1</number>
           </property>
+          <property name="maximum">
+           <number>9999</number>
+          </property>
           <property name="showClearButton" stdset="0">
            <bool>false</bool>
           </property>


nyalldawson:
Sure... why not?! :stuck_out_tongue_closed_eyes:

nyalldawson (approved):

PR #38559, opened 2020-09-02
Fix unreported crash on startup

elpaso:
annotation layer could be null

PR #38593, opened 2020-09-03
Fix crash when creating project and default transform context has some entries

nyalldawson:
Alternate fix for https://github.com/qgis/QGIS/pull/38559


nyalldawson:
@elpaso I think this might be a better fix for the crash...

PR #38599, opened 2020-09-04
Port MapBox GL style to QGIS style conversion from Python plugins

nyalldawson:
This PR ports the work began by @wonder-sk and continued in the MapTiler plugin to implement a MapBox GL style to QGIS style converter to c++.

The intention here is to provide a single, definitive converter which can now be utilised by plugins like MapTiler. There's a requirement to have this code as a core QGIS c++ implementation so that we can extend the vector tile support in QGIS to automatically style layers correctly by default, without any manual user work. That's not done in this PR, but I'll shortly be pushing a follow up which adds the ability to easily configure ESRI VectorTileServices in QGIS, including automatic application of their correct styles.

(Also adds a bunch of tests for this conversion, which weren't present in the Python implementations.)


src/core/vectortile/qgsmapboxglstyleconverter.cpp
```diff
@@ -0,0 +1,1381 @@
...
+    format.buffer().setEnabled( true );
+    format.buffer().setSize( bufferSize * PIXEL_RATIO );
+    format.buffer().setSizeUnit( QgsUnitTypes::RenderPixels );
+    format.buffer().setColor( bufferColor );
+  }
+
+  QgsPalLayerSettings labelSettings;
+
+  // TODO - likely improvements required here
+  labelSettings.fieldName = QStringLiteral( "name:latin" );
```
wonder-sk, line 739: I think we don't need `name:latin` as the default field name - in the original code I used it just because I did not parse field name at all and this field was used in OpenMapTiles styles...

wonder-sk (approved):
Great stuff - looking forward to have it in master so that loading styled vector tile layers gets easier!

I have seen there are few to-dos for attributes like `fill-pattern` or `text-halo-blur` - maybe it would be good to report those as skipped in the list of warnings?

src/core/vectortile/qgsmapboxglstyleconverter.cpp
nyalldawson: Thanks -- I've gone back and revisited this text-field handling, it's much more robust now!

PR #38605, opened 2020-09-04
Added missing space

DiGro:
Line 123/124 : "generated points are considered.""If the"  should be:  "generated points are considered. ""If the"
                       to display correct in Help. 
                       Therefore added one space after the dot behind "considered" to make that happen.

## Description  ;  Display correct documentation

Adding the one space does not effect the working of the algorithm.
It is just to display correct documentation

Not sure if this should be backported (it is just a minor enhancement)


nyalldawson:
Thanks! Let's skip the backport and save the translators some work...

Issue #29021, opened 2019-02-08
']' keyboard (alt-gr + '+' ) not working on QGis 3.4.4 with Spanish or Italian keyboard - Windows


qgib:
Author Name: **Javier Becerra Elcinto** (Javier Becerra Elcinto)
Original Redmine Issue: [21203](https://issues.qgis.org/issues/21203)
Affected QGIS version: 3.4.4
Redmine category:gui

---
Running QGis 3.4.4, in a Windows 8.1 setup in Spanish, with Spanish es-es keyboard (Qwerty with '[' being alt-gr+'`' and ']' alt-gr+'+'), I can properly type '[' (alt-gr+'`') but nothing happens when I try to type ']' (alt-gr+'+').
This happens in Python Console (quite annoying actually), on Python Console's text editor or when changing a layer's name (right-click on the layer tree view on bottom left of the screen), but it works properly on dialogs shown in independent windows (i.e. open file dialog/save as...).

---

Related issue(s): #28820 (duplicates), #29800 (duplicates)
Redmine related issue(s): [21001](https://issues.qgis.org/issues/21001), [21985](https://issues.qgis.org/issues/21985)

---

Pedro-Murteira:
I can still replicate this on QGIS 3.16.14 and 3.22.1.

PR #38608, opened 2020-09-04
Drop the conflicting Ctrl+Alt+A shortcut 

borysiasty:
A few months ago (24e1a8269cf5837) the **Ctrl+Shift+A** shortcut (*Deselect all features from all layers*) was replaced with **Ctrl+Alt+A**, as the old one has been used for a new action, *Deselect features only from the current layer* (452ca911903e).

Unfortunately in Windows **Ctrl+Alt+A** is an alias to **GrAlt+A**, what is used for national characters in some localisations (Polish for example, but according to [Wikipedia](https://en.wikipedia.org/wiki/AltGr_key) also in Belgian, Latvian, Turkish, UK, Irish and Dutch), so that change broke the Locator bar and possibly other text inputs in scope of the map canvas. By the way, in Polish Debian/KDE the new shortcut seems to not work at all.

I propose to just drop that troublesome shortcut, leaving the old action (*for all layers*) without any. I can't find any other reasonable and safe shortcut for that action, and if there is only **Ctrl+Alt+A**, the new action (*current layer only*) probably deserves more for it. It's also consistent with its use in the Attribute table.

@suricactus is it ok for you?

### DIFF
--- a/src/ui/qgisapp.ui
+++ b/src/ui/qgisapp.ui
@@ -1234,9 +1234,6 @@
    <property name="text">
     <string>Deselect Features from All Layers</string>
    </property>
-   <property name="shortcut">
-    <string>Ctrl+Alt+A</string>
-   </property>
   </action>
   <action name="mActionDeselectActiveLayer">
    <property name="icon">


suricactus:
@borysiasty aah, didn't know that it is an alias to *GrAlt*. :( It is fine with me, but I will reference the discussion intiial https://github.com/qgis/QGIS/pull/35189 . Maybe the others have objections?

borysiasty:
@suricactus me too, just win users told me they can't type ą in Locator... 

I didn't know that discusssion, let's go there.

agiudiceandrea:
@borysiasty @suricactus Please see this https://github.com/qgis/QGIS/issues/29021 similar issue with ctrl+alt++ on Windows system with italian or spanish keyboard.

borysiasty:
So we have 13 Ctrl+Alt shortcuts. Almost all of them are used in one or more localisations (the last column seem to be the least troublesome):

```
Ctrl+Alt+A    Ctrl+Alt++    Ctrl+Alt+/
Ctrl+Alt+M    Ctrl+Alt+-    Ctrl+Alt+[
Ctrl+Alt+P    Ctrl+Alt+=    Ctrl+Alt+]
Ctrl+Alt+V    Ctrl+Alt+;    Ctrl+Alt+{
                            Ctrl+Alt+}
```

One solution is to drop all Ctrl+Alt shortcuts (what will hurt users from not affected languages).
Another is to only change or drop them in translation for affected languages.

Personally I'm ok with both.

agiudiceandrea:
> One solution is to drop all Ctrl+Alt shortcuts (what will hurt users from not affected languages).
> Another is to only change or drop them in translation for affected languages.

Would it be possible to remove the Ctrl+Alt+ shortcuts only when the Python console is active or, more generally, when the cursor is in an active editable field/place?

nyalldawson:
> Another is to only change or drop them in translation for affected languages.

I'm +1 to this -- I'm not too keen on loss of usability as a result of the specific locale handling.

nirvn:
+1 too.

agiudiceandrea:
> One solution is to drop all Ctrl+Alt shortcuts (what will hurt users from not affected languages).

Anyway, the discriminant is not the interface translation language, but the keyboard layout: on a system with an italian or spanish keyboard layout the issue is present non only when the italian or spanish user interface translation is set, but also when the system locale is overridden and the American English user interface is set.

borysiasty:
> Would it be possible to remove the Ctrl+Alt+ shortcuts only when the Python console is active or, more generally, when the cursor is in an active editable field/place?

Not easily. They are main window action shortcuts and work globally. Theoretically we could capture them in Py console but I guess it would be much easier to use them for something new than to revert to the OS-default action. And we could need to implement it everywhere: in Locator, Browser filter etc.

> I'm +1 to this -- I'm not too keen on loss of usability as a result of the specific locale handling.

It's probably a dozen locales affected, but only one OS. So I'd rather say: *as a result of supporting that particular OS* ;) 

Well, I'm also not too keen to hurt 80% of users for those 20 (those values are totally guessed), but it seems Ctrl+Alt shortcuts are in general not good for Windows.
 
> Anyway, the discriminant is not the interface translation language, but the keyboard layout: on a system with an italian or spanish keyboard layout the issue is present non only when the italian or spanish user interface translation is set, but also when the system locale is overridden and the American English user interface is set.

Good point. It's not a big problem here in PL as almost anyone use localised interface, but there was a discussion about translation quality recently and if I remember correctly many users in Spain or Portugal (please forgive my poor memory) use the EN version...

agiudiceandrea:
@borysiasty @suricactus @nirvn why don't go ahead with this PR? The issue is still present in addition to https://github.com/qgis/QGIS/issues/29021.

Issue #38167, opened 2020-08-06
[3D] Keep the last used tab of the Lights widget when configuring 3D map view


DelazJ:
1. Open the 3D Configuration dialog of a 3D map view
1. Expand the Lights menu and begin setting a "directional light"
1. Apply
1. Reopen the dialog in order to tweak the values: you get the "point lights" tab.
It could be nice to keep showing the directional light tab instead. More convenient.

PR #38615, opened 2020-09-06
[3D] Keep the last used tab of the Lights widget when configuring 3D map view

NEDJIMAbelgacem:
## Description
This implements the feature request mentioned in  #38167

### DIFF
--- a/src/app/3d/qgs3dmapconfigwidget.cpp
+++ b/src/app/3d/qgs3dmapconfigwidget.cpp
@@ -114,6 +114,7 @@ Qgs3DMapConfigWidget::Qgs3DMapConfigWidget( Qgs3DMapSettings *map, QgsMapCanvas 
 
   widgetLights->setPointLights( mMap->pointLights() );
   widgetLights->setDirectionalLights( mMap->directionalLights() );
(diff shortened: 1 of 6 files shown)


src/app/3d/qgs3dmapconfigwidget.cpp
```diff
@@ -114,6 +114,7 @@ Qgs3DMapConfigWidget::Qgs3DMapConfigWidget( Qgs3DMapSettings *map, QgsMapCanvas
 
   widgetLights->setPointLights( mMap->pointLights() );
   widgetLights->setDirectionalLights( mMap->directionalLights() );
+  widgetLights->setOpenLightTab( mMap->lightsWidgetTab() );
```
nyalldawson, line 117: A better approach here is to store/retrieve  the last used tab in QSettings, rather then in the map settings. That avoids the "bleed" of application/GUI related settings into the core 3d classes which is done here (the addition of lastOpenLightTab/setOpenLightTab in the base 3d map settings class).

nyalldawson:
Superseded by https://github.com/qgis/QGIS/pull/38901

Issue #35797, opened 2020-04-15
First impressions


olivierdalang:
**Describe the bug**

Just reinstalled QGIS from scratch on a new PC, and on first launch, I get this screen :

![image](https://user-images.githubusercontent.com/1894106/79366568-b2aac700-7f4c-11ea-9ed3-e723dbe6178b.png)

I had forgotten about it, but I think we should get rid of it. IMO it looks unprofessional (both the font style and the wording), and doesn't bring much besides slowing down the first launch.

The regular splash screen gives a much better first impression, and especially now that we have the "news" section, it feels redundant.

At the same time, we may want to improve the 

compared to our nice splash screen) 

I'd 

**How to Reproduce**

**QGIS and OS versions**

**Additional context**

roya0045:
I agree that the news panel is oversized and causes confusion to new and existing users when the previous project tab  is hidden under.

nyalldawson:
+1 to removal. It was mostly in place to facilitate upgrades from 2.x, but that's no longer a pressing concern.

NathanW2:
You can remove the Welcome screen if there are no other versions of QGIS installed however if this is a fresh install on a machine with 2.x still on it.  I think we should keep the one that does the settings migrations because there is no doubt people still moving...

I know we have all moved on but other people haven't and I put that there to ease the move

alexbruy:
Also this screen can be useful in future, when another breakage may occur.

andreasneumann:
I agree with Olivier, it looks so ugly to me, and despite the 2.x migration it has zero value. I would be in favor of removing it starting from 3.14. Someone who hasn't yet migrated from 2.x to 3.x will most likely also not migrate to the newest 3.x release, but probably to 3.10.

andreasneumann:
> Also this screen can be useful in future, when another breakage may occur.

Can we keep it and deactivate it?

kannes:
This is long gone and the issue can be closed :)

PR #38616, opened 2020-09-06
Aesthetic improvements to the first run dialog 

timlinux:
The first run dialog is shown for users who are either running QGIS for the first time on their system, or who are upgrading from an old 2.x installation.

In the first case, it invites the user to view the changelog (great!) and then has a big button to get started using QGIS.

In the second case, it offers to migrate the user's old settings to QGIS 3 before launching the application.

As the first impression a first time user gets, the dialog looks a little.....sad.

So this PR tries to lay out the dialog in a more pleasing way and also makes some small tweaks to the wording to hopefully make things a little more welcoming.

Before my changes:

FirstRunScreen-BeforeFixing

After my changes - with 2.x profile migration:

FirstRunScreen-WithOldQGIS

After my changes - no migration of 2.x profile:

FirstRunScreen-NoOldQGIS

### DIFF
--- a/src/ui/qgsfirstrundialog.ui
+++ b/src/ui/qgsfirstrundialog.ui
@@ -6,178 +6,101 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>659</width>
-    <height>464</height>
+    <width>742</width>
+    <height>320</height>
    </rect>
   </property>
   <property name="windowTitle">
    <string>Welcome to QGIS</string>
   </property>
-  <layout class="QGridLayout" name="gridLayout_2">
+  <layout class="QGridLayout" name="gridLayout">
+   <item row="0" column="0" rowspan="3">
+    <widget class="QLabel" name="label_3">
+     <property name="sizePolicy">
+      <sizepolicy hsizetype="Maximum" vsizetype="Maximum">
+       <horstretch>0</horstretch>
+       <verstretch>0</verstretch>
+      </sizepolicy>
+     </property>
+     <property name="text">
+      <string/>
+     </property>
+     <property name="pixmap">
+      <pixmap resource="../../images/images.qrc">:/images/icons/qgis_icon.svg</pixmap>
+     </property>
+     <property name="scaledContents">
+      <bool>false</bool>
+     </property>
+    </widget>
+   </item>
+   <item row="0" column="1">
+    <widget class="QLabel" name="mWelcomeLabel">
+     <property name="font">
+      <font>
+       <pointsize>23</pointsize>
+      </font>
+     </property>
+     <property name="text">
+      <string>Welcome to QGIS 3</string>
+     </property>
+    </widget>
+   </item>
+   <item row="1" column="1">
+    <widget class="QLabel" name="mWelcomeDevLabel">
+     <property name="text">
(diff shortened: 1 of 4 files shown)


timlinux:
CC @andreasneumann - I remember this was also something you noticed as an annoyance.....

nirvn:
Most welcome. Just a quick reaction for now: I feel the logo is a bit oversized.

timlinux:
> Most welcome. Just a quick reaction for now: I feel the logo is a bit oversized.

Awww - I like it big :-) But maybe it needs some more padding in the gutter between the text and logo ....

timlinux:
Maybe like this: 
image

nyalldawson:
@timlinux I can't recall where the conversation is (I think a bug ticket), but consensus was to disable this dialog again until qgis 4 rolls around. It's basically noise for the vast majority of new installs now!

timlinux:
@nyalldawson Ah found the ticket:

https://github.com/qgis/QGIS/issues/35797

I actually did look for a ticket relating to it before I started but missed it  when searching for 'first run' etc. Reading through the ticket I would say that feelings were mixed. I think it is OK to keep it if it looks tidy. Maybe we can just show it if we see they are migrating from 2.x, otherwise hide it....

nyalldawson:
@timlinux would it be possible to check in advance if a qgis 2 profile is actually present? If not, let's avoid the dialog entirely....

andreasneumann:
Much better than the previous one. But like @nyalldawson said, it would be great if it could be skipped, if there is no .qgis2 folder present.

timlinux:
Ok @nyalldawson and @andreasneumann thanks I will do that.

timlinux:
The above commits should now skip the dialog unless you are coming from QGIS 2.x.
Also adds more horizontal spacing in the layout like my mockup above.

nyalldawson:
Thanks @timlinux  -- the new logic looks good to me!

I'll let @nirvn handle the rest of this review (the UI layout changes)

timlinux:
Thanks @nyalldawson 

@nirvn I fixed up some alignment issues too:

image

If you still hate the large logo I will do a design with a smaller one...

src/ui/qgsfirstrundialog.ui
```diff
@@ -6,178 +6,107 @@
...
-         </widget>
-        </item>
-       </layout>
-      </widget>
-     </item>
-    </layout>
+   <item row="1" column="1">
+    <widget class="QLabel" name="mWelcomeDevLabel">
+     <property name="text">
+      <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;&lt;span style=&quot; font-style:italic;&quot;&gt;You are running a dev version. We would love your feedback and testing.&lt;/span&gt;&lt;/p&gt;&lt;p&gt;&lt;a href=&quot;http://changelog.qgis.org/en/qgis/version/3.4-LTR/&quot;&gt;&lt;span style=&quot; text-decoration: underline; color:#2a76c6;&quot;&gt;Check out &lt;/span&gt;&lt;/a&gt;the change log for all the great new features introduced with this release!&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
nirvn, line 54: Is there a reason why we link to 3.4 LTR changelog? That has already reached its end of life.

nirvn (commented):
I'm fine with the size of the logo now. It's still quite large, but better balanced overall.

Just one comment re an hyperlink to 3.4 LTR changelog

nyalldawson:
@timlinux any chance to address the final comments before release?

PR #38619, opened 2020-09-07
Vector Tile layer properties metadata improvements

nyalldawson:
- make the first tab look consistent with other layer types (here's the before/after):

Before:
image

After:
image

- Add QGIS layer metadata editing tab, because vector tile layers can also have useful metadata which users want to set (e.g. attribution and source details)


nirvn (approved):
Nice UI polish.

PR #38627, opened 2020-09-07
Add option to create connections for ArcGIS Vector Tile Services to easily load vector tile layers from these servers

nyalldawson:
(also add support for authentication configuration and custom referer string to vector tile connections)

This change adds explicit UI options for adding and configuring connections to ESRI vector tile services. While these services use the standard vector tile framework for data retrieval, they can be directly added by creating a connection using the VectorTileService API endpoint on the ArcGIS server.

After a user adds a layer from a VectorTileService, the server's default styling and labeling for the layer is automatically loaded immediately for the user:

Peek 2020-09-07 16-08

### DIFF
--- a/src/gui/vectortile/qgsarcgisvectortileconnectiondialog.h
+++ b/src/gui/vectortile/qgsarcgisvectortileconnectiondialog.h
@@ -38,9 +38,6 @@ class QgsArcgisVectorTileConnectionDialog : public QDialog, public Ui::QgsArcgis
   private slots:
     void updateOkButtonState();
 
-  private:
-    QString mBaseKey;
-    QString mCredentialsBaseKey;
 };
 
 ///@endcond
--- a/src/gui/vectortile/qgsarcgisvectortileconnectiondialog.h
+++ b/src/gui/vectortile/qgsarcgisvectortileconnectiondialog.h
@@ -1,0 +1,48 @@
+/***************************************************************************
+    qgsarcgisvectortileconnectiondialog.h
+    ---------------------
+    begin                : September 2020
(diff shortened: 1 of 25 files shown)


timlinux:
Ah wow - nice!

src/gui/vectortile/qgsarcgisvectortileconnectiondialog.h
```diff
@@ -0,0 +1,48 @@
...
+
+    QString connectionUri() const;
+    QString connectionName() const;
+
+  private slots:
+    void updateOkButtonState();
+
+  private:
+    QString mBaseKey;
+    QString mCredentialsBaseKey;
```
nirvn, line 43: `mBaseKey` and `mCredentialsBaseKey` don't appear to be used, remove?

src/ui/qgsvectortileconnectiondialog.ui
```diff
@@ -67,43 +67,71 @@
...
-         <string>http://example.com/{z}/{x}/{y}.pbf</string>
+        <property name="checked">
+         <bool>true</bool>
         </property>
        </widget>
       </item>
-      <item row="6" column="0" colspan="2">
-       <widget class="QCheckBox" name="mCheckBoxZMax">
+      <item row="7" column="0" colspan="3">
+       <widget class="QGroupBox" name="mAuthGroupBox">
```
nirvn, line 81: Glad to see you've added authentication to the 'generic' vector tile dialog too, thanks for that.

nirvn (commented):
Looking great as always.

nyalldawson:
Thanks @nirvn!

Issue #36701, opened 2020-05-25
Improvements in Locator feature  search


andreasneumann:
**Feature description.**

1. Make the nr of hits configurable (global, per layer, current layer) for the search modes "f" and "af"
2.  improve mode "f" to allow searching in certain columns only: f @my_field my_value - add autocompletion for attribute names if possible
3. In the layer properties fields tab allow to specify if an attribute should be "searchable" (next to the WMS/WFS columns)

About the third item, a rough idea how this could be handled:

![grafik](https://user-images.githubusercontent.com/884476/82812078-7bf09680-9e92-11ea-8b67-e4ef54c212bd.png)

As discussed with @3nids - will be handled as part of our support contract

andreasneumann:
Ah before I forget about it: please add in some extra time for improving the documentation as well.

I think this is right section?

https://docs.qgis.org/testing/en/docs/user_manual/introduction/qgis_configuration.html#locator-settings

uclaros:
I think we are missing a visual indicator that the max results has been reached and there are more matching results that are not displayed! Especially if the number of displayed results is configurable.

3nids:
PR for configuration of max number of results: https://github.com/qgis/QGIS/pull/38628
PR for restricting the active layer search to a specific field with auto completion: https://github.com/qgis/QGIS/pull/38633

3nids:
PRs for defining which field is searchable:

1. https://github.com/qgis/QGIS/pull/38634 (merged)
2. https://github.com/qgis/QGIS/pull/38687
3. https://github.com/qgis/QGIS/pull/38688

3nids:
PR for flashing features: https://github.com/qgis/QGIS/pull/38694

PR #38628, opened 2020-09-07
allow to define max number of results for some locator filters

3nids:
this allows to define:
*  maximum number of results for active layer
* global maximum filter + per layer maximum numer of results of all layers features filter

### DIFF
--- a/src/app/locator/qgsinbuiltlocatorfilters.cpp
+++ b/src/app/locator/qgsinbuiltlocatorfilters.cpp
@@ -17,6 +17,7 @@
 
 #include <QClipboard>
 #include <QMap>
+#include <QSpinBox>
 #include <QString>
 #include <QToolButton>
 #include <QUrl>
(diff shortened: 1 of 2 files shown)


3nids:
is this worth a FEATURE label?

nyalldawson:
When would a user themselves need to configure this? (vs the filter implementation itself...)

3nids:
This has been a feature request here: #36701
Mainly, it's because some want to see all results rather than a subset.
I still think we should keep the numbers low by default (mainly looking at how recent online tools work: https://map.geo.admin.ch/ or google maps)

nyalldawson:
Sorry -- misread the diff. I thought you where adding it for ALL locator filters, not particular ones!

3nids:
test is failing on push on this repo (then branch has been pushed here, sorry for this).
any objection?

PR #38630, opened 2020-09-07
[ui] Improve the load style dialog

nirvn:
- Remove harmful horizontal spacer to fix expanding widget issues and location of help button
- Relocate the file widget to sit on top of the category list (to harmonize with the save style dialog)

Before (left) vs. PR (right):
image


m-kuhn (approved):
So much nicer :heart:

Issue #38393, opened 2020-08-20
Dbmanager causes QGIS to crash when trying to open Postgres db


rduivenvoorde:
To reproduce:
(I'm on Debian testing here, QGIS master self compiled)

select postgis_full_version()  gives me
POSTGIS="3.0.2 r2fb2a18" [EXTENSION] PGSQL="120" GEOS="3.8.1-CAPI-1.13.3" PROJ="7.1.0" LIBXML="2.9.10" LIBJSON="0.15" LIBPROTOBUF="1.3.3" WAGYU="0.4.3 (Internal)" (core procs from "3.0.0 r17983" need upgrade)

select version() gives me
PostgreSQL 12.3 (Debian 12.3-1+b1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 9.3.0-13) 9.3.0, 64-bit

- create a fresh profiles-path (qgis  --profiles-path /tmp/foo)
- create a local Postgres db connection (plain passwords)
- test with normal QGIS Postgres connection manager if you are able to see the schema's  and tables: OK
- test with Browser: OK
- open DB Manager
- open PostGIS node (still ok)
- click 'localhost' node: CRASH

stacktrace below:

../src/core/qgsdataitem.cpp:956 : (~QgsDataCollectionItem) [20ms] mName = localhost mPath = pg:/localhost
../src/core/qgsdataitem.cpp:266 : (~QgsDataItem) [0ms] mName = localhost mPath = pg:/localhost mChildren.size() = 0
../src/core/qgsdataitem.cpp:266 : (~QgsDataItem) [0ms] mName = OpenStreetMap mPath = xyz:/OpenStreetMap mChildren.size() = 0
../src/core/qgsdataitem.cpp:956 : (~QgsDataCollectionItem) [0ms] mName = localhost mPath = pg:/localhost
../src/core/qgsdataitem.cpp:266 : (~QgsDataItem) [0ms] mName = localhost mPath = pg:/localhost mChildren.size() = 0
../src/core/qgsdataitem.cpp:266 : (~QgsDataItem) [0ms] mName = OpenStreetMap mPath = xyz:/OpenStreetMap mChildren.size() = 0
QGIS died on signal 11[New LWP 120929]
[New LWP 120930]
[New LWP 120931]
[New LWP 120932]
[New LWP 120934]
[New LWP 120935]
[New LWP 120936]
[New LWP 120937]
[New LWP 120944]
[New LWP 120945]
[New LWP 120950]
[...]

PR #38634, opened 2020-09-07
add configuration flags to QgsFields

3nids:
This is a preliminary work to allow defining if fields are searchable in active layer locator filter
(feature request: https://github.com/qgis/QGIS/issues/36701)

The idea to use flags is to make this future proof and easily extend the flags.
Saving/reading to XML will be handy using [QMetaEnum::valueToKeys](https://doc.qt.io/qt-5/qmetaenum.html#valueToKeys)

(Later on, WMS/WFS capabilities could be moved to these flags)

### DIFF
--- a/src/core/qgsfield.h
+++ b/src/core/qgsfield.h
@@ -60,7 +60,7 @@ class CORE_EXPORT QgsField
     Q_PROPERTY( QString alias READ alias WRITE setAlias )
     Q_PROPERTY( QgsDefaultValue defaultValueDefinition READ defaultValueDefinition WRITE setDefaultValueDefinition )
     Q_PROPERTY( QgsFieldConstraints constraints READ constraints WRITE setConstraints )
-    Q_PROPERTY( ConfigurationFlags flags READ flags WRITE setFlags )
+    Q_PROPERTY( ConfigurationFlags configurationFlags READ configurationFlags WRITE setConfigurationFlags )
 
 
   public:
@@ -305,13 +305,13 @@ class CORE_EXPORT QgsField
      * Returns the Flags for the field (searchable, …)
      * \since QGIS 3.16
      */
-    QgsField::ConfigurationFlag flags() const SIP_SKIP;
+    QgsField::ConfigurationFlags configurationFlags() const SIP_SKIP;
 
     /**
      * Sets the Flags for the field (searchable, …)
      * \since QGIS 3.16
      */
-    void setFlags( QgsField::ConfigurationFlag flags ) SIP_SKIP;
+    void setConfigurationFlags( QgsField::ConfigurationFlags configurationFlags ) SIP_SKIP;
 
     //! Formats string for display
     QString displayString( const QVariant &v ) const;
--- a/src/core/qgsfield.h
+++ b/src/core/qgsfield.h
@@ -60,8 +60,22 @@ class CORE_EXPORT QgsField
     Q_PROPERTY( QString alias READ alias WRITE setAlias )
     Q_PROPERTY( QgsDefaultValue defaultValueDefinition READ defaultValueDefinition WRITE setDefaultValueDefinition )
     Q_PROPERTY( QgsFieldConstraints constraints READ constraints WRITE setConstraints )
+    Q_PROPERTY( Flags flags READ flags WRITE setFlags )
+
 
   public:
+
+    /**
+       * Flags for fields
+       * \since QGIS 3.16
+       */
+    enum class Flag : int
+    {
+      Searchable = 0x1, //!< Defines if the field is searchable (used in the locator search for instance)
+      DefaultFlags = Searchable, //!< Default set of flags for a field
+    };
+    Q_ENUM( Flag )
+    Q_DECLARE_FLAGS( Flags, Flag )
 
     /**
      * Constructor. Constructs a new QgsField object.
@@ -281,6 +295,18 @@ class CORE_EXPORT QgsField
      * \since QGIS 3.0
      */
     void setAlias( const QString &alias );
+
+    /**
+     * Returns the Flags for the field (searchable, …)
+     * \since QGIS 3.16
+     */
+    QgsField::Flags flags() const;
+
+    /**
+     * Sets the Flags for the field (searchable, …)
+     * \since QGIS 3.16
+     */
+    void setFlags( QgsField::Flags flags );
 
     //! Formats string for display
     QString displayString( const QVariant &v ) const;
--- a/src/core/qgsfield.h
+++ b/src/core/qgsfield.h
@@ -60,22 +60,27 @@ class CORE_EXPORT QgsField
     Q_PROPERTY( QString alias READ alias WRITE setAlias )
     Q_PROPERTY( QgsDefaultValue defaultValueDefinition READ defaultValueDefinition WRITE setDefaultValueDefinition )
     Q_PROPERTY( QgsFieldConstraints constraints READ constraints WRITE setConstraints )
-    Q_PROPERTY( Flags flags READ flags WRITE setFlags )
+    Q_PROPERTY( ConfigurationFlags flags READ flags WRITE setFlags )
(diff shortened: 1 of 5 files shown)


3nids:
Before going any further, I'd like to know if there is any objections to adding such flags to QgsField.

elpaso:
@3nids I like the idea of the flags but we need to clear up the semantics: are these flags if directly related to the data source?
I tend to think at `QgsField` as a container for properties that come directly from the provider. So, if the "Searchable" flag comes from the provider (e.g. WMS exposing queryable) it is ok for me, but if that is a property that the user sets I have some doubts it's the right way to go.

For example, for the server flags: they are basically set by the user and are part of the layer configuration in the server context. I would keep them separate, but I like the flags approach though.

3nids:
> but if that is a property that the user sets I have some doubts it's the right way to go.

yes, it's intended to be set by the user.

> I would keep them separate

So you would add a new list as a property of the layer?

I like the idea of moving this to field because the information is where you need it, you already have iterators for the fields.
It avoids having to keep the lists in sync with possible changing fields in the layer (adding/removing/renaming).

It was a bit the same we did for the editable, searchable, required layers lists which were moved as flagd in QgsMapLayer (https://github.com/qgis/QGIS/commit/89830f614eac2f4aec17ba4bc07db885ab7c4ca7)

I can go for a more descriptive phrasing, like `ConfigurationFlags`, would it sound better?

elpaso:
> > but if that is a property that the user sets I have some doubts it's the right way to go.
> 
> yes, it's intended to be set by the user.
> 
> > I would keep them separate
> 
> So you would add a new list as a property of the layer?

No, as long as it is clear what the flags are and they are homogeneous this is probably not necessary.

Can you make a preliminary list of what settings will be stored as flags (you mentioned server configuration flags such as exposure as wms/wfs and the like).

> 
> I like the idea of moving this to field because the information is where you need it, you already have iterators for the fields.
> It avoids having to keep the lists in sync with possible changing fields in the layer (adding/removing/renaming).
> 
> It was a bit the same we did for the editable, searchable, required layers lists which were moved as flagd in QgsMapLayer ([89830f6](https://github.com/qgis/QGIS/commit/89830f614eac2f4aec17ba4bc07db885ab7c4ca7))
> 
> I can go for a more descriptive phrasing, like `ConfigurationFlags`, would it sound better?

Yes, that makes it clearer.

src/core/qgsfield.h
```diff
@@ -282,6 +296,18 @@ class CORE_EXPORT QgsField
      */
     void setAlias( const QString &alias );
 
+    /**
+     * Returns the Flags for the field (searchable, …)
+     * \since QGIS 3.16
```
nyalldawson, line 301: \warning these will be totally broken on some python versions :laughing:

src/core/qgsfield.h
```diff
@@ -282,6 +296,18 @@ class CORE_EXPORT QgsField
      */
     void setAlias( const QString &alias );
 
+    /**
+     * Returns the Flags for the field (searchable, …)
```
nyalldawson, line 300:
```suggestion
     * Returns the Flags for the field (searchable, ...)
```

I'd skip the unicode characters in dox -- it's likely to open up a world of hurt

nyalldawson:
I'm not against storing extra stuff in QgsField (e.g. I'd like to add an optional QgsNumericFormat value to it at some stage). The tricky bit is handling expectations from users who assume that changing the field's flags will automatically do ... something...! You'll need to add a massive /warning note to the dox to state that changing the flags doesn't actually change the layer settings, and point users to the correct api to use to actually set these flags on the layer/provider level.

I'm also wondering... I get the love for enum classes, but I'm thinking it's just too premature to use these widely. Could we hold back on that until post QGIS 3.16 when we do the proposed massive dependency requirement bump?

3nids:
> Can you make a preliminary list of what settings will be stored as flags (you mentioned server configuration flags such as exposure as wms/wfs and the like).

I think it would make sense to bring _WMS_, _WFS_. _Editable_ could come to mind (it can be discussed if it's related to widget or field).
But the scope will be limited to _searchable_ here.

3nids:
>  I get the love for enum classes, but I'm thinking it's just too premature to use these widely. Could we hold back on that until post QGIS 3.16 when we do the proposed massive dependency requirement bump

Since this preliminary work is required, I would propose to no publish them in the PyQGIS API, only in C++ (until 3.18). Sounds like a plan?

elpaso:
@3nids sorry if it's unrelated, but there is a pending issue where you were mentioned that might be related with the enums, would you please have a look? https://github.com/qgis/QGIS/issues/38393

elpaso:
> > Can you make a preliminary list of what settings will be stored as flags (you mentioned server configuration flags such as exposure as wms/wfs and the like).
> 
> I think it would make sense to bring _WMS_, _WFS_. _Editable_ could come to mind (it can be discussed if it's related to widget or field).

Can we separate the server specific flags from the desktop flags? 

Or maybe we need a more complex struct and use the same "origin" or "domain" concept similar to what we have for constraints?

WFS editable can be both a field property of a remote WFS that we are accessing from the desktop and a user-settable flag that we set for a field in a project that will be served from QGIS Server.

What I was trying to say with my first comment is that we should possibly avoid to mix the flags that are user-configurable and those that are "hard" coming from the provider that cannot be overridden by the API (if a remote WFS field is not editable there is nothing we can do about it but in case we are designing a WFS for QGIS Server it's up to us to decide if a field is editable or not)


> But the scope will be limited to _searchable_ here.

Is that different from _queryable_ ?

elpaso:
mmm actually _queryable_ in WMS context applies to the whole layer so it might not be a good example, I'm not sure if we want to propagate layer properties to the field flags.

3nids:
> What I was trying to say with my first comment is that we should possibly avoid to mix the flags that are user-configurable and those that are "hard" coming from the provider that cannot be overridden by the API (if a remote WFS field is not editable there is nothing we can do about it but in case we are designing a WFS for QGIS Server it's up to us to decide if a field is editable or not)

I understand your point now. Last commit renames the flags to ConfigurationFlags and I added some text in the docstring.

For searchable, this flags is here mainly for the active layer features locator filter, to determine in which field it can search by default.

jgrocha:
@3nids For Python users, we are tagging `excludeAttributesWms` and `excludeAttributesWfs` as deprecated, without an alternative. The [alternative](https://qgis.github.io/pyqgis/master/core/QgsVectorLayer.html?highlight=configurationflags#qgis.core.QgsVectorLayer.excludeAttributesWms) `fields().configurationFlags()` is not exposed yet.

Do you see any inconvenient to expose `field.configurationFlags()` or `fields().configurationFlags()` write now?

jgrocha:
> @3nids For Python users, we are tagging `excludeAttributesWms` and `excludeAttributesWfs` as deprecated, without an alternative. The [alternative](https://qgis.github.io/pyqgis/master/core/QgsVectorLayer.html?highlight=configurationflags#qgis.core.QgsVectorLayer.excludeAttributesWms) `fields().configurationFlags()` is not exposed yet.
> 
> Do you see any inconvenient to expose `field.configurationFlags()` or `fields().configurationFlags()` write now?

On the current master, `excludeAttributesWms()` and `excludeAttributesWfs()` are returning an empty set.

Captura de ecrã de 2020-12-15 00-48-40

```python
layer = iface.activeLayer()
layer.excludeAttributesWfs()
set()
layer.excludeAttributesWms()
set()
```

3nids:
I made a PR to fix the deprecated methods: https://github.com/qgis/QGIS/pull/40602

Regarding the availability in Python, I am waiting on the bump on the requirements to have a more recent sip to avoid issues with scoped enums.
The window to have this by 3.18 is getting tight, not sure when it will happen.

jgrocha:
Thank you @3nids for your prompt answer and for #40602

PR #38643, opened 2020-09-08
Add help string for parameters in Processing

Gustry:
## Description

This PR restores the previous behavior from QGIS 2 where we could set a help string at the parameter level.

This will be visible in the tooltip, in the CLI qgis_process and Python Processing API.

Selection_016

### DIFF
--- a/src/core/processing/qgsprocessingparameters.h
+++ b/src/core/processing/qgsprocessingparameters.h
@@ -427,7 +427,7 @@ class CORE_EXPORT QgsProcessingParameterDefinition
      * Constructor for QgsProcessingParameterDefinition.
      */
     QgsProcessingParameterDefinition( const QString &name, const QString &description = QString(), const QVariant &defaultValue = QVariant(),
-                                      bool optional = false );
+                                      bool optional = false, const QString &help = QString() );
 
     virtual ~QgsProcessingParameterDefinition() = default;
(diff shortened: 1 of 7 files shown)


root676:
This makes the designing complex processing algorithms so much easier! No parameter description in the algorithm helpstring anymore!

src/core/processing/qgsprocessingparameters.h
```diff
@@ -476,6 +476,20 @@ class CORE_EXPORT QgsProcessingParameterDefinition
...
 
+    /**
+     * Returns the help for the parameter.
+     * \see setHelp()
+     * \since QGIS 3.16
+     */
+    QString help() const { return mHelp; }
+
+    /**
+     * Sets the \a help for the parameter.
```
nyalldawson, line 487:
```suggestion
     * Sets the \a help for the parameter.
     *
     * The help string should be a translated, in-depth description of the parameter as how it impacts the algorithm's operation.
```

nyalldawson (commented):
Great, thanks for this!

Gustry:
Would it be OK if it's API only?
I noticed that the code goes in the model for creating parameters etc.

nyalldawson:
Superseded by https://github.com/qgis/QGIS/pull/38897

Issue #29504, opened 2019-03-28
Zonal Statistics alters zone layer, should create new layer


qgib:
Author Name: **Johannes Kroeger** (Johannes Kroeger)
Original Redmine Issue: [21688](https://issues.qgis.org/issues/21688)

Redmine category:processing/qgis

---
Zonal Statistics tries to write the statistics to the file/layer chosen in "Vector layer containing zones".

I have not tried, but I would assume that this renders the tool unusable with e.g. a read-only WFS source.

Pretty much all other tools, including Zonal Histogram, instead create a new layer or file.

Please make Zonal Statistics not try to alter the zone layer by default, instead use the same mechanism ("[Create temporary layer]"  by default) as other tools.

alexbruy:
This should be an analysis library feature request not Processing

alexbruy:
Fixed in #38650

PR #38650, opened 2020-09-08
[processing] Add Zonal statistics algorithm that creates a new output

m-kuhn:
So far, when calculating zonal statistics, these were always updating the original data source, adding additional fields.
This is often not desirable (cluttering the dataset) and sometimes not possible (read only data source). Also it integrates badly with the modeller, where we normally have immutable data sources and data pipelines.

This pull request adds a new **Zonal statistics (feature based)** algorithm, that works as expected.

Implements #29504

### DIFF
--- a/src/analysis/processing/qgsalgorithmzonalstatistics.cpp
+++ b/src/analysis/processing/qgsalgorithmzonalstatistics.cpp
@@ -64,7 +64,7 @@ QString QgsZonalStatisticsAlgorithm::groupId() const
 QString QgsZonalStatisticsAlgorithm::shortHelpString() const
 {
   return QObject::tr( "This algorithm calculates statistics of a raster layer for each feature "
-                      "of an overlapping polygon vector layer. The reults will be written in place." );
+                      "of an overlapping polygon vector layer. The results will be written in place." );
 }
 
 QgsProcessingAlgorithm::Flags QgsZonalStatisticsAlgorithm::flags() const
--- a/src/analysis/processing/qgsalgorithmzonalstatistics.cpp
+++ b/src/analysis/processing/qgsalgorithmzonalstatistics.cpp
@@ -69,7 +69,7 @@ QString QgsZonalStatisticsAlgorithm::shortHelpString() const
 
 QgsProcessingAlgorithm::Flags QgsZonalStatisticsAlgorithm::flags() const
 {
-  return QgsProcessingAlgorithm::flags() | QgsProcessingAlgorithm::FlagNoThreading;
+  return QgsProcessingAlgorithm::flags() | QgsProcessingAlgorithm::FlagNoThreading | QgsProcessingAlgorithm::FlagDeprecated;
 }
 
 QgsZonalStatisticsAlgorithm *QgsZonalStatisticsAlgorithm::createInstance() const
--- a/src/analysis/processing/qgsalgorithmzonalstatistics.cpp
+++ b/src/analysis/processing/qgsalgorithmzonalstatistics.cpp
@@ -89,14 +89,14 @@ void QgsZonalStatisticsAlgorithm::initAlgorithm( const QVariantMap & )
   addParameter( new QgsProcessingParameterRasterLayer( QStringLiteral( "INPUT_RASTER" ), QObject::tr( "Raster layer" ) ) );
   addParameter( new QgsProcessingParameterBand( QStringLiteral( "RASTER_BAND" ),
                 QObject::tr( "Raster band" ), 1, QStringLiteral( "INPUT_RASTER" ) ) );
-  addParameter( new QgsProcessingParameterVectorLayer( QStringLiteral( "INPUT_VECTOR" ), QObject::tr( "Vector layer containing zones" ),
+  addParameter( new QgsProcessingParameterFeatureSource( QStringLiteral( "INPUT_VECTOR" ), QObject::tr( "Feature source containing zones" ),
                 QList< int >() << QgsProcessing::TypeVectorPolygon ) );
   addParameter( new QgsProcessingParameterString( QStringLiteral( "COLUMN_PREFIX" ), QObject::tr( "Output column prefix" ), QStringLiteral( "_" ) ) );
 
   addParameter( new QgsProcessingParameterEnum( QStringLiteral( "STATISTICS" ), QObject::tr( "Statistics to calculate" ),
                 statChoices, true, QVariantList() << 0 << 1 << 2 ) );
 
-  addOutput( new QgsProcessingOutputVectorLayer( QStringLiteral( "INPUT_VECTOR" ), QObject::tr( "Zonal statistics" ), QgsProcessing::TypeVectorPolygon ) );
+  addParameter( new QgsProcessingParameterFeatureSink( QStringLiteral( "OUTPUT" ), QObject::tr( "Zonal statistics" ), QgsProcessing::TypeVectorPolygon, QVariant(), false, true, true ) );
 }
 
 bool QgsZonalStatisticsAlgorithm::prepareAlgorithm( const QVariantMap &parameters, QgsProcessingContext &context, QgsProcessingFeedback * )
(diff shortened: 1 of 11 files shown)


src/analysis/vector/qgszonalstatistics.h
```diff
@@ -63,6 +66,17 @@ class ANALYSIS_EXPORT QgsZonalStatistics
...
     Q_DECLARE_FLAGS( Statistics, Statistic )
 
+    enum Result
+    {
+      Success = 0,
+      LayerTypeWrong = 1,
+      LayerInvalid,
+      RasterInvalid,
+      RasterBandInvalid,
+      FailedToCreateField = 8,
```
m-kuhn, line 76: Remove, no longer exists

src/analysis/processing/qgsalgorithmzonalstatistics.cpp
```diff
@@ -89,14 +89,14 @@ void QgsZonalStatisticsAlgorithm::initAlgorithm( const QVariantMap & )
...
-  addParameter( new QgsProcessingParameterVectorLayer( QStringLiteral( "INPUT_VECTOR" ), QObject::tr( "Vector layer containing zones" ),
+  addParameter( new QgsProcessingParameterFeatureSource( QStringLiteral( "INPUT_VECTOR" ), QObject::tr( "Feature source containing zones" ),
                 QList< int >() << QgsProcessing::TypeVectorPolygon ) );
   addParameter( new QgsProcessingParameterString( QStringLiteral( "COLUMN_PREFIX" ), QObject::tr( "Output column prefix" ), QStringLiteral( "_" ) ) );
 
   addParameter( new QgsProcessingParameterEnum( QStringLiteral( "STATISTICS" ), QObject::tr( "Statistics to calculate" ),
                 statChoices, true, QVariantList() << 0 << 1 << 2 ) );
 
-  addOutput( new QgsProcessingOutputVectorLayer( QStringLiteral( "INPUT_VECTOR" ), QObject::tr( "Zonal statistics" ), QgsProcessing::TypeVectorPolygon ) );
+  addParameter( new QgsProcessingParameterFeatureSink( QStringLiteral( "OUTPUT" ), QObject::tr( "Zonal statistics" ), QgsProcessing::TypeVectorPolygon, QVariant(), false, true, true ) );
```
nyalldawson, line 99: Unfortunately we can't make this change here -- it would break all existing processing models and scripts which use this algorithm.

That said -- I do think this is a very good change. What we'd need to do is
1. split the algorithm into two (with a common base class).
2. the existing native:zonalstatistics algorithm remains unchanged, with the exception that it now returns the FlagDeprecated flag to hide it from users
3. the new subclass does your new logic and creates a new sink, instead of editing in place. (and this one can remove the FlagNoThreading flag). It has a different id, so as not to clash with the existing deprecated algorithm.

src/analysis/vector/qgszonalstatistics.cpp
```diff
@@ -305,55 +249,66 @@ int QgsZonalStatistics::calculateStatistics( QgsFeedback *feedback )
...
-    changeMap.insert( f.id(), changeAttributeMap );
+    if ( mSink )
+    {
+      QgsFeature newFeature = feature;
+      newFeature.setFields( mFields );
+      for ( const auto &attr : changeAttributeMap.toStdMap() )
+      {
+        newFeature.setAttribute( attr.first, attr.second );
+      }
+      mSink->addFeature( newFeature );
```
nyalldawson, line 290:
```suggestion
      mSink->addFeature( newFeature, QgsFeatureSink::FastInsert );
```

nirvn:
This always bugged me, glad to see it being addressed.

src/analysis/processing/qgsalgorithmzonalstatistics.cpp
m-kuhn: Fair points. I didn't like the approach that much anyway, completely different one coming up.

src/analysis/processing/qgsalgorithmzonalstatistics.cpp
```diff
@@ -64,7 +64,7 @@ QString QgsZonalStatisticsAlgorithm::groupId() const
 QString QgsZonalStatisticsAlgorithm::shortHelpString() const
 {
   return QObject::tr( "This algorithm calculates statistics of a raster layer for each feature "
-                      "of an overlapping polygon vector layer." );
+                      "of an overlapping polygon vector layer. The reults will be written in place." );
```
DelazJ, line 67:
```suggestion
                      "of an overlapping polygon vector layer. The results will be written in place." );
```

m-kuhn:
@nyalldawson wondering about your thoughts concerning this approach:

 - a new feature based algorithm with a new id
 - refactoring of the zonal stats to be directly callable from within the loop (static method)

It's clean code wise. But it does expose 2 algorithms instead of 1 with support for in place. But I didn't see a possibility for this (nor did you propose it in your comment I think)

src/analysis/processing/qgsalgorithmzonalstatisticsfeaturebased.cpp
```diff
@@ -0,0 +1,189 @@
...
+};
+
+QString QgsZonalStatisticsFeatureBasedAlgorithm::name() const
+{
+  return QStringLiteral( "zonalstatisticsfb" );
+}
+
+QString QgsZonalStatisticsFeatureBasedAlgorithm::displayName() const
+{
+  return QObject::tr( "Zonal statistics (feature based)" );
```
nyalldawson, line 45:
```suggestion
  return QObject::tr( "Zonal statistics" );
```

Lets just deprecate the old one -- I really don't want two algorithms showing in the toolbox for this one function, it's way too confusing for users!

src/analysis/processing/qgsalgorithmzonalstatisticsfeaturebased.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+QgsFields QgsZonalStatisticsFeatureBasedAlgorithm::outputFields( const QgsFields &inputFields ) const
+{
+  Q_UNUSED( inputFields )
+  return mOutputFields;
+}
+
+QgsProcessingFeatureSource::Flag QgsZonalStatisticsFeatureBasedAlgorithm::sourceFlags() const
+{
+  return QgsProcessingFeatureSource::FlagSkipGeometryValidityChecks;
```
nyalldawson, line 113: Better remove this flag -- the algorithm uses geos intersections so it needs valid inputs.

src/analysis/processing/qgsalgorithmzonalstatisticsfeaturebased.cpp
```diff
@@ -0,0 +1,189 @@
...
+                                  .arg( rasterLayer->bandCount() ) );
+
+  if ( !rasterLayer->dataProvider() )
+    throw QgsProcessingException( QObject::tr( "Invalid raster layer. Layer %1 is invalid." ).arg( rasterLayer->id() ) );
+
+  mRaster.reset( rasterLayer->dataProvider()->clone() );
+  mCrs = rasterLayer->crs();
+  mPixelSizeX = rasterLayer->rasterUnitsPerPixelX();
+  mPixelSizeY = rasterLayer->rasterUnitsPerPixelY();
+  QgsFeatureSource *source = parameterAsSource( parameters, inputParameterName(), context );
```
nyalldawson, line 143: This is leaked -- it needs storing in a unique ptr

src/analysis/processing/qgsalgorithmzonalstatisticsfeaturebased.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+QgsFeatureList QgsZonalStatisticsFeatureBasedAlgorithm::processFeature( const QgsFeature &feature, QgsProcessingContext &context, QgsProcessingFeedback *feedback )
+{
+  Q_UNUSED( context )
+  Q_UNUSED( feedback )
+  QgsAttributes attributes = feature.attributes();
+  attributes.resize( mOutputFields.size() );
+
+  QMap<QgsZonalStatistics::Statistic, QVariant> results = QgsZonalStatistics::calculateStatistics( mRaster.get(), feature.geometry(), mPixelSizeX, mPixelSizeY, mBand, mStats );
+  for ( const auto &result : results.toStdMap() )
```
nyalldawson, line 172: This looks inefficient -- why not iterate the QMap directly?

m-kuhn: Readability.
As soon as we bump to c++ 17 we'll be able to do that directly on the QMap (yay!)

DelazJ:
@m-kuhn I guess docs should be aligned https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/rasteranalysis.html#zonal-statistics ?

PR #38651, opened 2020-09-08
Allow data-defined offset for fill symbol layers

nyalldawson:
Any fill symbol layer which supports offsetting polygons (i.e. simple
fills, raster image fills, shapeburst and gradent fills) now allows
for data-defined offsets


nyalldawson:
Approval for this was already given in https://github.com/qgis/QGIS/pull/38642 -- I've just pulled one of the commits out of that PR and added tests for it.

PR #38652, opened 2020-09-08
Allow users to optionally specify the URL for default style for vector tile connections

nyalldawson:
When setting up a vector tile source connection, there's a new option
to enter a URL to a MapBox GL JSON style configuration. If one has
been entered, then that style will be applied whenever the layers
from the connection are added to QGIS.

Works also with Arcgis vectortileservice connections -- in this
case the URL overrides the default style configuration specified
in the server configuration.

Peek 2020-09-09 09-25


nirvn (approved):
All good.

PR #38654, opened 2020-09-09
UI Improvements for vector tile symbology/labeling

nyalldawson:
Some quality of life improvements for users editing complex vector tiles styles:

- show tooltips in the lists, so you can easily see the full filter rules and names without resizing columns
- show the current canvas zoom level in the widgets, and allow users to filter the list of styles to only show those which are currently visible. Makes it much easier to identify which style is being used to render the current map!
- Set the @zoom_level variable to the correct canvas zoom level when editing symbols, so that expression previews work nicely.


nirvn (approved):
Cool stuff.

Issue #37847, opened 2020-07-16
New feature is hidden when created from attribute table


andreasneumann:
When creating a new feature from the attribute table with a no-geometry table, the new feature is added, but it is added to the bottom of the form or table and not activated properly. The user has to manually scroll down to the very bottom and activate the new feature in the form manually.

I would expect either:
- QGIS would activate the new feature in the form right away in the already existing form view
- QGIS would open the new feature in a separate form window

see [video](https://youtu.be/YiaSdFL4TYs)

The issue appears in both master and QGIS 3.10.

andreasneumann:
Might be related to issue #37846

nyalldawson:
> QGIS would open the new feature in a separate form window

This is what we used to do, but it was rather odd behaviour because the attribute table was already visible containing the new row.

andreasneumann:
> This is what we used to do, but it was rather odd behaviour because the attribute table was already visible containing the new row.

Yes - but the current behavior is even more odd: you add a new feature --> then you don't see anything until you get it that you have to scroll down to the very bottom. Can't QGIS do that for me? I mean the scroll-down and focus of the newly added record?

andreasneumann:
@suricactus - I tested it - much better! It works fine if the attribute table is in "form mode", but if it is in "table mode", there is still the issue that the  newly created feature is added and not focused. Are you able to address the case of the "table mode" as well? Do you understand what I mean or should I record a video?

suricactus:
Yup, it's clear, I will fix that these days too.

roya0045:
Is this fixed then?

andreasneumann:
Seems to be fixed an backported

suricactus:
Glad to hear @andreasneumann  :)

PR #38658, opened 2020-09-09
FIX 37847 New feature is hidden when created from attribute table

suricactus:
Now when a new feature is added in the attribute table when in form view, the new feature is automatically selected and the feature list on the left side is ensured to show it (not necessarily on top). In response of #37847 .

Peek 2020-09-09 09-45

### DIFF
--- a/src/app/qgsfeatureaction.h
+++ b/src/app/qgsfeatureaction.h
@@ -61,6 +61,12 @@ class APP_EXPORT QgsFeatureAction : public QAction
      */
     void setForceSuppressFormPopup( bool force );
 
+    /**
+     * Returns the current feature. It might be nullptr feature, so it's safe to use this method only
+     * right after`addFeature()` returned true.
+     */
+    QgsFeature *feature() const { return mFeature; }
+
   signals:
 
     /**
--- a/src/app/qgsfeatureaction.h
+++ b/src/app/qgsfeatureaction.h
@@ -62,10 +62,9 @@ class APP_EXPORT QgsFeatureAction : public QAction
     void setForceSuppressFormPopup( bool force );
 
     /**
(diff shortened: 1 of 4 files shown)


src/gui/attributetable/qgsfeaturelistview.cpp
```diff
@@ -189,18 +189,27 @@ void QgsFeatureListView::selectAll()
...
   const auto constFids = fids;
   for ( QgsFeatureId fid : constFids )
   {
-    selection.append( QItemSelectionRange( mModel->mapToMaster( mModel->fidToIdx( fid ) ) ) );
+    QModelIndex modelIdx = mModel->fidToIdx( fid );
+
+    if ( ! firstModelIdx.isValid() )
+      firstModelIdx = modelIdx;
+
+    selection.append( QItemSelectionRange( mModel->mapToMaster( firstModelIdx ) ) );
```
signedav, line 202: Not sure, but shouldn't it take `modelIdx` still?

suricactus: thanks, true that

src/app/qgsfeatureaction.h
```diff
@@ -61,6 +61,12 @@ class APP_EXPORT QgsFeatureAction : public QAction
      */
     void setForceSuppressFormPopup( bool force );
 
+    /**
+     * Returns the current feature. It might be nullptr feature, so it's safe to use this method only
+     * right after`addFeature()` returned true.
+     */
+    QgsFeature *feature() const { return mFeature; }
```
nyalldawson, line 68: Can we return a invalid QgsFeature here instead of mFeature is null, and avoid the raw pointer please?

suricactus: Done!

signedav (approved):

PR #38659, opened 2020-09-09
Add support for filter expressions in relation reference widget

suricactus:
Adds an option for static filter in relation reference form widget.

Peek 2020-09-09 10-20

### DIFF
--- a/src/gui/editorwidgets/qgsrelationreferenceconfigdlg.h
+++ b/src/gui/editorwidgets/qgsrelationreferenceconfigdlg.h
@@ -39,13 +39,6 @@ class GUI_EXPORT QgsRelationReferenceConfigDlg : public QgsEditorConfigWidget, p
     QVariantMap config() override;
     void setConfig( const QVariantMap &config ) override;
 
-  public slots:
-
-    /**
-     * Opens an expression dialog and sets its value as filter expression for the relation reference.
-     */
-    void editExpression();
-
   private:
     void loadFields();
     void addFilterField( const QString &field );
@@ -58,6 +51,11 @@ class GUI_EXPORT QgsRelationReferenceConfigDlg : public QgsEditorConfigWidget, p
     void relationChanged( int idx );
     void mAddFilterButton_clicked();
     void mRemoveFilterButton_clicked();
+
+    /**
+     * Opens an expression dialog and sets its value as filter expression for the relation reference.
+     */
+    void mEditExpression_clicked();
 };
 
 #endif // QGSRELATIONREFERENCECONFIGDLGBASE_H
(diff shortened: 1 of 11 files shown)


src/gui/editorwidgets/qgsrelationreferenceconfigdlg.h
```diff
@@ -39,6 +39,13 @@ class GUI_EXPORT QgsRelationReferenceConfigDlg : public QgsEditorConfigWidget, p
     QVariantMap config() override;
     void setConfig( const QVariantMap &config ) override;
 
+  public slots:
+
+    /**
+     * Opens an expression dialog and sets its value as filter expression for the relation reference.
+     */
+    void editExpression();
```
m-kuhn, line 47: Can this be a private slot?

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
```diff
@@ -906,9 +918,13 @@ void QgsRelationReferenceWidget::filterChanged()
         {
           QMap<QString, QString> filtersAttrs = filters;
           filtersAttrs[fieldName] = QgsExpression::createFieldEqualityExpression( fieldName, txt );
-          QString expression = filtersAttrs.values().join( QStringLiteral( " AND " ) );
-
           QgsAttributeList subset = attrs;
+
+          QString expression = filterExpression;
+          if ( ! filterExpression.isEmpty() && ! filtersAttrs.values().isEmpty() )
+            filterExpression = QStringLiteral( " AND " );
```
m-kuhn, line 925: Is that correct? Should this overwrite the original filterExpression or did you mean to append?
Also should the filters be put into parentheses?

suricactus: You are right, this is wrong. Thanks for noticing.

suricactus: Since all the groupings are `AND` parentheses are not required, but probably more future proof.

src/gui/editorwidgets/qgsrelationreferenceconfigdlg.h
suricactus: Got inspired by `QgsValueRelationConfigDlg::editExpression`, which is a public slot:
https://github.com/qgis/QGIS/blob/65eed59ed0a0ff4fc27d332bd6124c0d2ad69a4f/src/gui/editorwidgets/qgsvaluerelationconfigdlg.h#L39-L40

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
suricactus: It was also using the wrong variable, should be good now.
```diff
-            filterExpression = QStringLiteral( " AND " );
+            expression += QStringLiteral( " AND " );
```

src/gui/editorwidgets/qgsrelationreferenceconfigdlg.h
m-kuhn: Looks like a leftover, only `public` if required.

suricactus: I will make it `private` then, no requirements involved.

PR #38661, opened 2020-09-09
[labelling] Allow non-zero distance from line values for on line curved placement

nirvn:
## Description

Because there are no reasons why we were not allowing a non-zero distance value for on line curved label placement and because it's needed for vector tile style import.

@nyalldawson , very little to change, but since the labeling code is a nightmare, review appreciated :)

Peek 2020-09-10 08-02


nirvn:
@nyalldawson , I don't think any of the failures are actual regressions, merely noise in the machine. I'll update the reference images when you give me the green light.

nirvn:
Vector tile style conversion for distanced curved labels implemented:
image

src/core/vectortile/qgsmapboxglstyleconverter.cpp
```diff
@@ -1040,6 +1040,36 @@ void QgsMapBoxGlStyleConverter::parseSymbolLayer( const QVariantMap &jsonLayer,
...
+    QPointF textOffset;
+    if ( jsonLayout.contains( QStringLiteral( "text-offset" ) ) )
+    {
+      const QVariant jsonTextOffset = jsonLayout.value( QStringLiteral( "text-offset" ) );
+
+      // units are ems!
+      switch ( jsonTextOffset.type() )
+      {
+        case QVariant::Map:
+          ddLabelProperties.setProperty( QgsPalLayerSettings::LabelDistance, QStringLiteral( "array_geT(%1 ,1)" ).arg( parseInterpolatePointByZoom( jsonTextOffset.toMap(), context, textSize, &textOffset ) ) );
```
nyalldawson, line 1053:
```suggestion
          ddLabelProperties.setProperty( QgsPalLayerSettings::LabelDistance, QStringLiteral( "array_get(%1,1)" ).arg( parseInterpolatePointByZoom( jsonTextOffset.toMap(), context, textSize, &textOffset ) ) );
```

tests/src/core/testqgslabelingengine.cpp
```diff
@@ -1274,6 +1274,18 @@ void TestQgsLabelingEngine::testCurvedLabelCorrectLinePlacement()
 
   img = job2.renderedImage();
   QVERIFY( imageCheck( QStringLiteral( "label_curved_label_below_1" ), img, 20 ) );
+
+  // and on line with distance value...
+  settings.lineSettings().setPlacementFlags( QgsLabeling::LinePlacementFlag::OnLine | QgsLabeling::LinePlacementFlag::MapOrientation );
+  settings.dist = 5.0;
```
nyalldawson, line 1280: Could we also get a test with negative distance, and some using the parallel mode too?

nirvn:
@nyalldawson , review addressed, travis green.

src/core/labeling/qgslabelfeature.h
```diff
@@ -274,6 +274,18 @@ class CORE_EXPORT QgsLabelFeature
      */
     void setDistLabel( double dist ) { mDistLabel = dist; }
 
+    /**
+     * Applies to on-line "curved" placement strategy for linestring features.
+     * Distance of the label from the feature (in map units)
+     */
+    double originalDistLabel() const { return mOriginalDistLabel; }
```
nyalldawson, line 281: How would you feel about flipping this, and adding "adjustedLabelDistance()" instead? Then the above/below line code would be changed to use the adjusted distance instead of distLabel(). I think that approach would be much more readable...

nirvn:
Killing this whole approach, allowing distance for the on line placement breaks mixed line placement scenarios.

PR #38662, opened 2020-09-09
Raster layer can use XML extent like vector layer

rldhont:
## Description

The trust flag at the project level is only used to read vector layer extent from xml, not from provider.

This flag was not available for raster layer.

We propose to add the same capabilities in raster layer

### DIFF
--- a/src/core/raster/qgsrasterlayer.h
+++ b/src/core/raster/qgsrasterlayer.h
@@ -472,24 +472,6 @@ class CORE_EXPORT QgsRasterLayer : public QgsMapLayer
      */
     bool ignoreExtents() const;
 
-    /**
-     * Flag allowing to indicate if the extent has to be read from the XML
-     * document when data source has no metadata or if the data provider has
-     * to determine it.
-     *
-     * \since QGIS 3.16
-     */
-    void setReadExtentFromXml( bool readExtentFromXml );
-
-    /**
-     * Returns TRUE if the extent is read from the XML document when data
-     * source has no metadata, FALSE if it's the data provider which determines
-     * it.
-     *
-     * \since QGIS 3.16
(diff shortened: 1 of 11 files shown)


m-kuhn:
Could we just move it up to QgsMapLayer?

src/core/raster/qgsrasterlayer.h
```diff
@@ -121,6 +123,14 @@ class CORE_EXPORT QgsRasterLayer : public QgsMapLayer
        */
       QgsCoordinateTransformContext transformContext = QgsCoordinateTransformContext();
 
+
+      /**
+       * If TRUE, the layer extent will be read from XML (i.e. stored in the
+       * project file). If FALSE, the extent will be determined by the provider on layer load.
+       * \since QGIS 3.16
+       */
+      bool readExtentFromXml = false;
```
nyalldawson, line 132: Can we use flags here instead? I'm not a fan of adding setReadExtentFromXml/readExtentFromXml to the already complex layer api

rldhont:
@m-kuhn and @nyalldawson the idea was in a first implementation to mimic the vector layer use xml extent for raster. It is to speed up QGIS Server project load.

I have planed to implement a way to manage the XML extent and to extend this capabilities to other layer types.

m-kuhn:
Can you elaborate on the advantage of one implementation per layer type over one single implementation? Couldn't we avoid code duplication and maintenance overhead by having one single implementation?

nyalldawson:
I'd rather see a clean API added going forward vs spreading the "messier" API to more classes...

rldhont:
With https://github.com/qgis/QGIS/pull/38464 merged, it will be easier to propose e cleaner API.

PR #38669, opened 2020-09-09
Don't blame plugins for slow loading times

uclaros:
## Description
When the *recent projects* list contains an entry located on an inaccessible/offline network location then, while QGIS is starting, Windows may spend quite some time trying to reach it but the splash screen misleadingly indicates *Restoring loaded plugins*.
This PR updates the message to *Updating recent project paths*, so the user knows the reason behind his coffee break while QGIS launches.


Saijin-Naib:
Stupid question, but could those be lazy-loaded/checked asynchronously in the background instead?

nyalldawson (approved):

PR #38672, opened 2020-09-10
Parentheses are needed in case the filter expression contains "OR"

suricactus:
This is a follow up PR to https://github.com/qgis/QGIS/pull/38659 . In case the filter expression is set to `1 OR 0` and there are filter fields, it will become `1 OR 0 AND field1 = 'something'` which is wrong. 

With this PR parentheses are ensured around the filter expression and the result would be ` ( 1 OR 0 ) AND field1 = 'something'`. I have added a test checking this situation.

### DIFF
--- a/src/gui/editorwidgets/qgsrelationreferencewidget.cpp
+++ b/src/gui/editorwidgets/qgsrelationreferencewidget.cpp
@@ -862,7 +862,7 @@ void QgsRelationReferenceWidget::filterChanged()
 
   QgsFeature f;
   QgsFeatureIds featureIds;
-  QString filterExpression = mFilterExpression;
+  QString filterExpression = mFilterExpression.isEmpty() ? mFilterExpression : QStringLiteral( " ( %1 ) " ).arg( mFilterExpression );
 
   // comboboxes have to be disabled before building filters
   if ( mChainFilters )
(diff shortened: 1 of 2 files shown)


m-kuhn (approved):
thanks

src/gui/editorwidgets/qgsrelationreferencewidget.cpp
```diff
@@ -862,7 +862,7 @@ void QgsRelationReferenceWidget::filterChanged()
 
   QgsFeature f;
   QgsFeatureIds featureIds;
-  QString filterExpression = mFilterExpression;
+  QString filterExpression = mFilterExpression.isEmpty() ? mFilterExpression : QStringLiteral( " ( %1 ) " ).arg( mFilterExpression );
```
3nids, line 865:
```suggestion
  if( !mFilterExpression.isEmpty() )
    mFilterExpression = QStringLiteral( " ( %1 ) " ).arg( mFilterExpression );
```

3nids: or `mFilterExpression.prepend( " (" ).append( ") " );`

suricactus: I guess you mean (note the missing `m` prefix):

```cpp
  QString filterExpression = mFilterExpression;
  if( !filterExpression.isEmpty() )
    filterExpression = QStringLiteral( " ( %1 ) " ).arg( filterExpression );
```

suricactus: In case you meant to change the member `mFilterExpression`, I would say it is better to keep it untouched and wrap the parentheses only in the places where they are needed.

I have applied the change as in my previous comment.

suricactus:
@3nids thanks for putting the missing `!`

PR #38678, opened 2020-09-10
Fix visible state of buttons in the Relation Editor Widget

### LINKED ISSUE #37718: Make more buttons in relation editor configurable
**Feature description.**
It is suggested to make most (all?) buttons configurable (show | don't show) in the relation editor widget:

image

image

This includes:
- Add child feature (important, as in some projects we only want to allow to link/unlink existing features)
- duplicate child feature
- delete child feature
- zoom to child feature

signedav:
`mZoomToFeatureButton` visibility is set on several places and it used to override each other. Same for `mSaveEditsButton` and `mAddFeatureGeometryButton`.

To keep it stable and not being dependent on the order of the setting, it store the configured visibility of buttons in `mButtonsVisibility` and calls `upadteButtons` from `setVisibleButtons`, `setRelation`, `initDualView` and `setEditorContext`.

For `mZoomToFeatureButton` it checks if the flag is set, the editorContexts mapCanvas exists and if the referenced layer of `relation` / the referencing layer of the `nmRelation` is spacial
For `mAddFeatureGeometryButton` it checks the same, but additionally for the existence of `mEditorContext.cadDockWidget()`.

For `mSaveEditsButton` and `mToggleEditingButton` it checks has the member `mLayerInSameTransactionGroup`. This is additionally set correctly according if the parent AND the child are in this group (or with many-to-many relations additionally the "other" parent layer as well).

fixes #37718

### DIFF
--- a/src/gui/qgsrelationeditorwidget.cpp
+++ b/src/gui/qgsrelationeditorwidget.cpp
@@ -430,12 +430,6 @@ void QgsRelationEditorWidget::updateButtons()
   mDeleteFeatureButton->setEnabled( editable && selectionNotEmpty );
   mUnlinkFeatureButton->setEnabled( linkable && selectionNotEmpty );
 
-  mZoomToFeatureButton->setVisible(
-    mButtons.testFlag( QgsAttributeEditorRelation::Button::ZoomToChildFeature ) &&
-    mEditorContext.mapCanvas() &&
-    mChildIsSpatial
-  );
-
   mZoomToFeatureButton->setEnabled( selectionNotEmpty );
 
   mToggleEditingButton->setChecked( editable );
--- a/src/gui/qgsrelationeditorwidget.cpp
+++ b/src/gui/qgsrelationeditorwidget.cpp
@@ -287,16 +287,9 @@ void QgsRelationEditorWidget::initDualView( QgsVectorLayer *layer, const QgsFeat
     text = tr( "Add Polygon Feature" );
   }
 
-  if ( text.isEmpty() || !mEditorContext.mapCanvas() || !mEditorContext.cadDockWidget() )
-  {
-    mAddFeatureGeometryButton->setVisible( false );
-  }
-  else
-  {
-    mAddFeatureGeometryButton->setIcon( icon );
-    mAddFeatureGeometryButton->setText( text );
-    mAddFeatureGeometryButton->setToolTip( text );
-  }
+  mAddFeatureGeometryButton->setIcon( icon );
+  mAddFeatureGeometryButton->setText( text );
+  mAddFeatureGeometryButton->setToolTip( text );
 
   updateButtons();
 }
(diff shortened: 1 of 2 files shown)


src/gui/qgsrelationeditorwidget.cpp
```diff
@@ -959,14 +945,15 @@ bool QgsRelationEditorWidget::showSaveChildEditsButton() const
...
-  mZoomToFeatureButton->setVisible( buttons.testFlag( QgsAttributeEditorRelation::Button::ZoomToChildFeature ) );
+  mButtons = buttons;
+  mLinkFeatureButton->setVisible( mButtons.testFlag( QgsAttributeEditorRelation::Button::Link ) );
+  mUnlinkFeatureButton->setVisible( mButtons.testFlag( QgsAttributeEditorRelation::Button::Unlink ) );
+  mSaveEditsButton->setVisible( mButtons.testFlag( QgsAttributeEditorRelation::Button::SaveChildEdits ) );
+  mAddFeatureButton->setVisible( mButtons.testFlag( QgsAttributeEditorRelation::Button::AddChildFeature ) );
+  mAddFeatureGeometryButton->setVisible( mButtons.testFlag( QgsAttributeEditorRelation::Button::AddChildFeature ) );
+  mDuplicateFeatureButton->setVisible( mButtons.testFlag( QgsAttributeEditorRelation::Button::DuplicateChildFeature ) );
+  mDeleteFeatureButton->setVisible( mButtons.testFlag( QgsAttributeEditorRelation::Button::DeleteChildFeature ) );
+  mZoomToFeatureButton->setVisible( mButtons.testFlag( QgsAttributeEditorRelation::Button::ZoomToChildFeature ) && mChildIsSpatial );
```
3nids, line 956:
```suggestion
  mZoomToFeatureButton->setVisible( mButtons.testFlag( QgsAttributeEditorRelation::Button::ZoomToChildFeature ) && mChildIsSpatial && mEditorContext.mapCanvas() );
```

3nids (approved):
Looks much cleaner!
What about the check for the map canvas, can it be introduced as proposed?

src/gui/qgsrelationeditorwidget.cpp
```diff
@@ -427,27 +420,18 @@ void QgsRelationEditorWidget::updateButtons()
...
-  );
-
   mZoomToFeatureButton->setEnabled( selectionNotEmpty );
-
   mToggleEditingButton->setChecked( editable );
   mSaveEditsButton->setEnabled( editable );
+
+  mLinkFeatureButton->setVisible( mButtonsVisibility.testFlag( QgsAttributeEditorRelation::Button::Link ) );
+  mUnlinkFeatureButton->setVisible( mButtonsVisibility.testFlag( QgsAttributeEditorRelation::Button::Unlink ) );
+  mSaveEditsButton->setVisible( mButtonsVisibility.testFlag( QgsAttributeEditorRelation::Button::SaveChildEdits ) && mToggleEditingButton->isVisible() );
```
3nids, line 429:
```suggestion
  mSaveEditsButton->setVisible( mButtonsVisibility.testFlag( QgsAttributeEditorRelation::Button::SaveChildEdits ) && editable );
```

sounds better no?

3nids, line 429: can't we use another condition that the button visibility?

3nids (approved):
nice!

src/gui/qgsrelationeditorwidget.cpp
signedav: One possibility would be to check if the referencing layer is in the `transactionGroup` but I don't like that it iterates here since it's not really something that changed.

Another possibility would be to set the status in something like `mLayerInTransactionGroup = true` or `mLayerEditable = false` here https://github.com/qgis/QGIS/pull/38678/files#diff-52237b5949ffab9a9049c00ed3a713e1L332 and then check the member variable in the `setVisible`.

What do you think?

3nids: yes, let's go for the member, it will make the logic more explicit.

3nids (approved):

signedav:
Thanks @3nids

PR #38679, opened 2020-09-10
extend qgsEnumKeyToValue to allow using the value (int) as a key (string)

3nids:


src/core/qgis.h
```diff
@@ -513,18 +513,33 @@ template<class T> QString qgsEnumValueToKey( const T &value ) SIP_SKIP
 /**
  * Returns the value corresponding to the given \a key of an enum.
  * If the key is invalid, it will return the \a defaultValue.
+ * If \a tryValueAsKey is true, it will try to use convert the string key to an enum value
```
DelazJ, line 516:
```suggestion
 * If \a tryValueAsKey is true, it will try to convert the string key to an enum value
```

3nids: thanks

PR #38681, opened 2020-09-10
move the logic of writing weak relations info away from QgsVectorLayer to QgsWeakRelation

3nids:
and also use QgsRelation::writeXml to avoid an incomplete info if the relation code gets other attributes

### DIFF
--- a/src/core/qgsvectorlayer.cpp
+++ b/src/core/qgsvectorlayer.cpp
@@ -2129,29 +2129,8 @@ bool QgsVectorLayer::readSymbology( const QDomNode &layerNode, QString &errorMes
       for ( int i = 0; i < relationNodes.length(); ++i )
       {
         const QDomElement relationElement = relationNodes.at( i ).toElement();
-        QList<QgsRelation::FieldPair> fieldPairs;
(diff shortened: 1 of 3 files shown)


elpaso (approved):
LGTM, I didn't check what https://github.com/qgis/QGIS/pull/38681/files#diff-a7da93edd7cd70922f2cdf20cf27c5c1R87 does but I guess it creates a `relation` element that we can get on the next line.

3nids:
>  it creates a relation element that we can get on the next line

yes, exactly.

Is this code tested somewhere?
I had a look in testqgsweakrelation.cpp, but there is no reading/writing of a project with broken layers. I couldn't find any other test.

elpaso:
@3nids there are some tests here https://github.com/qgis/QGIS/pull/32487/files#diff-f924ee63dae157d08a6f66fc07eac90cR311 but I don't think they cover the XML serialization, just the resolution functionality.

elpaso:
@3nids sorry, I meant this test link: https://github.com/qgis/QGIS/pull/33283/files#diff-095fa493f14d5c2ee03ea529640d700d the other one is more generic.

3nids:
Yes, indeed. thanks for the pointers.

PR #38687, opened 2020-09-10
Configure searchable flag for field in vector layer properties

3nids:
image

### DIFF
--- a/python/core/auto_generated/qgsvectorlayer.sip.in
+++ b/python/core/auto_generated/qgsvectorlayer.sip.in
@@ -2292,6 +2292,8 @@ can also be set. Setting an empty expression will clear any existing expression 
 .. versionadded:: 3.0
 %End
 
+
+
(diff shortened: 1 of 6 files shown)


nyalldawson:
Instead of the growing list of columns here, could we try to use the checkable combo box widget instead and move the server related flags in there too? (Or push them away to a server tab, where they actually belong!)

3nids:
> Instead of the growing list of columns here, could we try to use the checkable combo box widget instead and move the server related flags in there too? (Or push them away to a server tab, where they actually belong!)

Good point, I'll do that.

3nids:
manually merged with 59ff51697663c41a79d34cfa2989f3aa00c57ac6 4544c9614fa54548373fd6b3212642e370df9b2c 14ef69f8a5b3fceaf415ab4e9e9ef6389fae3631

3nids:
I will do the UI work in a following PR

PR #38691, opened 2020-09-11
Convert some more mapbox gl styles

nirvn:
## Description

This PR adds support for:
- converting text-offset values for labels set to line placement
- converting interpolated icon marker images
- converting interpolated fill pattern images

Interpolated icon markers GIF:
Peek 2020-09-11 14-33
_Look for the restaurant, bar, etc. markers going from round dots to larger icons._

### DIFF
--- a/src/core/vectortile/qgsmapboxglstyleconverter.cpp
+++ b/src/core/vectortile/qgsmapboxglstyleconverter.cpp
@@ -1992,9 +1992,31 @@ QImage QgsMapBoxGlStyleConverter::retrieveSprite( const QString &name, QgsMapBox
   return sprite;
 }
 
-QString QgsMapBoxGlStyleConverter::retrieveSpriteAsBase64( const QVariant &value, QgsMapBoxGlStyleConversionContext &context, QSize &size, QString &spriteProperty, QString &spriteSizeProperty )
+QString QgsMapBoxGlStyleConverter::retrieveSpriteAsBase64( const QVariant &value, QgsMapBoxGlStyleConversionContext &context, QSize &spriteSize, QString &spriteProperty, QString &spriteSizeProperty )
 {
-  QString returnPath;
+  QString spritePath;
+
+  auto prepareSprite = [context]( const QImage & sprite, QString & path, QSize & size )
+  {
+    if ( !sprite.isNull() )
+    {
+      size = sprite.size() * context.pixelSizeConversionFactor();
+      QByteArray blob;
+      QBuffer buffer( &blob );
+      buffer.open( QIODevice::WriteOnly );
+      sprite.save( &buffer, "PNG" );
+      buffer.close();
+      QByteArray encoded = blob.toBase64();
+      path = QString( encoded );
+      path.prepend( QLatin1String( "base64:" ) );
+    }
+    else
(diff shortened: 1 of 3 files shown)


src/core/vectortile/qgsmapboxglstyleconverter.cpp
```diff
@@ -1938,24 +1992,130 @@ QImage QgsMapBoxGlStyleConverter::retrieveSprite( const QString &name, QgsMapBox
...
+                           .arg( size.width() );
+
+      QSize s;
+      for ( int i = 0; i < stops.size() - 1; ++i )
+      {
+        ;
+        sprite = retrieveSprite( stops.value( 0 ).toList().value( 1 ).toString(), context );
+        if ( !sprite.isNull() )
+        {
+          s = sprite.size() * context.pixelSizeConversionFactor();
```
nyalldawson, line 2060: This same code is used multiple times - could you pull it out to a single function?

src/core/vectortile/qgsmapboxglstyleconverter.cpp
```diff
@@ -1039,6 +1029,56 @@ void QgsMapBoxGlStyleConverter::parseSymbolLayer( const QVariantMap &jsonLayer,
...
+        }
+
+        if ( !textOffset.isNull() )
+        {
+          labelSettings.distUnits = context.targetUnit();
+          labelSettings.dist = std::abs( textOffset.y() ) - textSize;
+          labelSettings.lineSettings().setPlacementFlags( textOffset.y() > 0.0 ? QgsLabeling::BelowLine : QgsLabeling::AboveLine );
+          if ( !textSizeProperty.isEmpty() && textOffsetProperty.isEmpty() )
+          {
+            ddLabelProperties.setProperty( QgsPalLayerSettings::LabelDistance, QStringLiteral( "with_variable('text_size',%2,%1*@text_size-@text_size)" ).arg( std::abs( textOffset.y() / textSize ) ).arg( ( textSizeProperty ) ) );
```
nyalldawson, line 1073: It is this actually needed? I can't work out the conditions where it's used

nirvn: It's needed when the text size is dynamically defined via a property.

nirvn: Ie, the offset might be static, but em being a derivative of text size, if the latter changes, we need to adjust.

src/core/vectortile/qgsmapboxglstyleconverter.cpp
nirvn: Will do.

nirvn:
@nyalldawson , review addressed.

nirvn:
@nyalldawson , one more bit of styling converted, the text-rotation-alignment:viewport mode for line labels. 

That's often used (alongside background shields) for road numbers:
Screenshot from 2020-09-13 09-29-34

As for the shield itself, I don't think we can convert it ATM.

nirvn:
Well, no, let's do shields after all:
image

:smile:

nyalldawson (approved):

PR #38695, opened 2020-09-11
Translate qgis::fieldcalculator to C++

suricactus:
The qgis:fieldcalculator algorithm is currently written in Python and consists of rather custom UI. This PR attempts to translate the algorithm to a C++ implementation, aiming at speed, compatibility.

The UI slightly changes from the previous implementation, having to custom UI widgets. There is a bit of inconvinience that both "Result in" fields are optional, but I don't find a way to make them either required. This is well noted in the algorithm description.

| before | after new field |
|----------|--------|
| Screenshot_20200909_230218 |  image |

A simpler version of https://github.com/qgis/QGIS/pull/38675 , that targets only the algorithm.


src/analysis/processing/qgsalgorithmfieldcalculator.cpp
```diff
@@ -0,0 +1,223 @@
...
+QgsFields QgsFieldCalculatorAlgorithm::outputFields( const QgsFields & ) const
+{
+  return mFields;
+}
+
+QString QgsFieldCalculatorAlgorithm::shortHelpString() const
+{
+  return QObject::tr( "This algorithm computes a new vector layer with the same features of the input layer, "
+                      "but with an additional attribute. The values of this new attribute are computed from each "
+                      "feature using a mathematical formula, based on the properties and attributes of the feature."
```
m-kuhn, line 96: - Better talk about "an expression" instead of "a math formula"?

suricactus: It was copy paste from the python implementation. Nice isuggestion.

src/analysis/processing/qgsalgorithmfieldcalculator.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+QgsFields QgsFieldCalculatorAlgorithm::outputFields( const QgsFields & ) const
+{
+  return mFields;
+}
+
+QString QgsFieldCalculatorAlgorithm::shortHelpString() const
+{
+  return QObject::tr( "This algorithm computes a new vector layer with the same features of the input layer, "
+                      "but with an additional attribute. The values of this new attribute are computed from each "
```
m-kuhn, line 95: - Maybe better "and either overwrites an existing ... or adds an additional"?

m-kuhn (approved):
Given the improvements of a stable C++ algorithm over the python one and the previous implementation in https://github.com/qgis/QGIS/pull/38675 being discouraged at the moment, I am in favor of merging this one with the minimal UI change for now and continuing the discussion for an improved user interface in parallel.

Are there any tests for this algorithm already?

suricactus:
I have added an additional test. Good you mentioned the tests, it was not exactly backwards compatible. Now all should be good.

python/plugins/processing/tests/testdata/qgis_algorithm_tests3.yaml
```diff
@@ -915,6 +915,22 @@ tests:
...
+    name: Test field calculator points (new API)
+    params:
+      FIELD_LENGTH: 10
+      FIELD_PRECISION: 3
+      FIELD_TYPE: 0
+      FORMULA: "\"id2\" *2"
+      INPUT:
+        name: points.gml
+        type: vector
+      NEW_FIELD_NAME: test
```
m-kuhn, line 928: @suricactus the old algorithm used `FIELD_NAME` as input, is there a safe way to have this backwards compatible?

suricactus: Did you encounter an issue? This should keep the backwards compatibility. The old test is passing too? 
https://github.com/qgis/QGIS/pull/38695/files#diff-53600f4cb9f762166bd6d0f4d56f555cR125-R135

PR #38701, opened 2020-09-11
Add option for Natural Sorting on Points to Paths algorithm

### LINKED ISSUE #35525: Natural sort in Points to path processing algorithm
**Feature description.**
QGIS would rock even more if a natural sort was used on the processing algorithm *Points to path* when the sort field is of type *string*. It is quite common to have points named *A1, A2, ..., A11* or even just integers in a string field so the resulting sort is *A1, A11, A2, ...*. 
In order to get the expected results now an intermediate numeric field has to be calculated first and used for sorting. 
With a natural sort the result will be the expected *A1, A2, ... A11*

**Additional context**
This might also apply to other algorithms too.

uclaros:
## Description
When using a *text* field for sorting on the *Points to Paths* algorithm the resulting order is not what most users would expect:
`a1,a11,a2,a222,a3`

This PR adds a *Sort text naturally* checkbox that when enabled will use a natural sort algorithm:
`a1,a2,a3,a11,a222`

Fixes #35525

### DIFF
--- a/python/plugins/processing/algs/qgis/PointsToPaths.py
+++ b/python/plugins/processing/algs/qgis/PointsToPaths.py
@@ -218,7 +218,7 @@ class PointsToPaths(QgisAlgorithm):
     def natural_sort_key(self, value, _nsre=re.compile('([0-9]+)')):
         if not value[0]:
             retVal = (False, None)
-        elif type(value[0]) in (str, unicode):
+        elif type(value[0]) == str:
             retVal = (True, [int(text) if text.isdecimal() else text.lower() for text in _nsre.split(value[0])])
         else:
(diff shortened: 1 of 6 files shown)


nyalldawson:
Any chance you could port this algorithm to C++ in the process? I'm concerned about adding new features like this in the python implementation which will ultimately make it harder to match the behaviour when we do eventually port it..

uclaros:
Since it's going for 3.18, I guess I'll find some time to do it.
This algorithm is severely handicapped without natural sorting, so I'll definitely implement it in c++ too.

nyalldawson:
Thanks!

alexbruy:
Superseded by #40014

Issue #38676, opened 2020-09-10
Layout: make more map-grid draw coordinates settings data-defined or controlled by QGIS


andreasneumann:
**Feature description.**
Depending on the map rotation angle (variable @map_rotation) we want to be able to control whether to "Show All | Show Latitude Only | Show Longitude Only | Disabled). Basically, the goal is to avoid labels at very flat angles in relation to their reference map edge. The second goal of this feature is to avoid the necessity for manual changes for draw coordinate settings when the map rotation changes.

![image](https://user-images.githubusercontent.com/884476/92726756-65e9e180-f36e-11ea-9aea-c57842544c83.png)

I see two options here:

1. QGIS would switch automatically for me, taking into account threshold values. It would probably mean a fifth option, something like "Automatic switch between lat/lon". E.g. when map_rotation is between >315 or <45 "Show Longitude", when >=45 and <= 135 switch to Show Latitude, and so on.
2. Make it data-defined and let the user control it by himself based on his own rules taking into account @map_rotation

For me, both options would be fine. 1. is more convenient, 2. more flexible

andreasneumann:
@olivierdalang and @nyalldawson - what do you think about the two options I listed above. Which one should we implement? 1), 2) or both?

olivierdalang:
@andreasneumann 
With a orthogonal grid (same CRS for the map and the grid), can't you achieve the desired effect by choosing "Show all" and setting a threshold angle to 45° ?

andreasneumann:
@olivierdalang right - this is exactly what I want. Thanks for the hint!

There is one small problem though: when you exactly hit the threshold value (e.g. 45 deg, or 135 deg map rotation with 45 deg skip low angled annotations), then neither latitude nor longitude are drawn. Any idea how to solve this?

olivierdalang:
@andreasneumann 
This case would require a fix in the code (put a `<=` instead of a `<`).

But then you'd end up with both latitudes and longitudes...

Though not very elegant, using 44.999° (or 45.001°) instead of an exact 45° should work without any visible difference ? That would avoid adding yet another option/setting in the already overcrowded (in part thanks to me :-) ) map grid settings widget... 

If that doesn't work for you, I'd go for option B/, as option A/ may not work well in some scenarios (e.g. when you have different CRS between the map and the grid, the grid may not be orthogonal, so that latitude/longitude may not all fall on the same map border). It would also avoid adding another option.

[...]

PR #38702, opened 2020-09-11
Rotated ticks followup

olivierdalang:
## Description

Some more polishing around the new rotated ticks feature as per https://github.com/qgis/QGIS/issues/38676:
- allow to input decimal values for the "skip low angled ticks" param
- add some tolerance to the "angle < threshold angle" comparison to deal with precision errors

(I'll leave this PR as draft for some time as it's not new features and there may be some more fixes coming before the release with more testing being done)


nyalldawson (approved):

nyalldawson:
Let's just do atomic PRs here, rather then one large collected PR

PR #38710, opened 2020-09-11
fix include for gui export

vcloarec:
fixes to allow real export of layers properties (vector and raster) from gui API


nyalldawson (approved):

m-kuhn:
Ideally we'd let the CI figure out missing installed include files ...

elpaso:
@m-kuhn ideally a pre-commit script so we spare two hours waiting for Travis.

m-kuhn:
If it's performant enough to not slow down committing, that's an option.

PR #38713, opened 2020-09-12
Show number of features affected by in place editing

### LINKED ISSUE #38680: When running in place "repair geometries" with a selection it is unclear if it only affects the selected features
Run "repair geometries" in place with a selection. The algorithm is started without a dialog and reports "successfully finished". It is unclear to the user if it was selected only or not. Checking in the processing history, the flag "selectedOnly=True" can be seen.

I see two options:

 - Offer a dialog to let the user check/uncheck "selected only"
 - Report the number of features in the confirmation dialog of the in place editing

m-kuhn:
Fixes #38680

image

### DIFF
--- a/python/plugins/processing/gui/AlgorithmExecutor.py
+++ b/python/plugins/processing/gui/AlgorithmExecutor.py
@@ -207,7 +207,7 @@ def execute_in_place_run(alg, parameters, context=None, feedback=None, raise_exc
 
                 feedback.setProgress(int((current + 1) * step))
 
-            results, ok = {'count': current + 1}, True
+            results, ok = {'__count': current + 1}, True
 
         else:  # Traditional 'run' with delete and add features cycle
 
@@ -246,7 +246,7 @@ def execute_in_place_run(alg, parameters, context=None, feedback=None, raise_exc
                     raise QgsProcessingException(tr("Error adding processed features back into the layer."))
                 new_ids = set([f.id() for f in active_layer.getFeatures(req)])
                 new_feature_ids += list(new_ids - old_ids)
-                results['count'] = len(new_feature_ids)
+                results['__count'] = len(new_feature_ids)
 
         active_layer.endEditCommand()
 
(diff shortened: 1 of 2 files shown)


m-kuhn:
@elpaso @nyalldawson  what's your opinion on the approach to return the number of features?

python/plugins/processing/gui/ProcessingToolbox.py
```diff
@@ -253,7 +253,7 @@ def executeAlgorithm(self):
                 feedback = MessageBarProgress(algname=alg.displayName())
                 ok, results = execute_in_place(alg, parameters, feedback=feedback)
                 if ok:
-                    iface.messageBar().pushSuccess('', self.tr('{} complete').format(alg.displayName()))
+                    iface.messageBar().pushSuccess('', self.tr('{algname} completed. {featurecount} feature(s) processed.').format(algname=alg.displayName(), featurecount=results['count']))
```
nirvn, line 256: What about using a %n tr count here? i.e. self.tr('{algname} completed. {featurecount}').format(algname=...,featurecount=self.tr('%n feature(s) processed','',results['count'])) ?

m-kuhn: Good point, I was trying to use `tr().arg()` which doesn't work ... But that's not actually required!

python/plugins/processing/gui/AlgorithmExecutor.py
```diff
@@ -207,7 +207,7 @@ def execute_in_place_run(alg, parameters, context=None, feedback=None, raise_exc
 
                 feedback.setProgress(int((current + 1) * step))
 
-            results, ok = {}, True
+            results, ok = {'count': current + 1}, True
```
nyalldawson, line 210: You'd also need to add this to the "#Tradition run..." block results below

nyalldawson (approved):

PR #38714, opened 2020-09-12
Speedup ogc tests

m-kuhn:
## Description

Two improvements to ogc testing ci

 - Disable ccache, [it has a 0% hit rate at the moment](https://github.com/qgis/QGIS/pull/38714/commits/6a62f70019dfb0558d590090453b5b8be01df7b9) and I'm not aware how to improve things with github workflows
 - Disable building of tests. We do that in other workflows and it just adds to the build time here.

1h straight now vs. 1h 22 minutes before = **27% faster**


pblottiere (approved):

pblottiere:
> 1h straight now vs. 1h 22 minutes before = 27% faster

Nice, thanks Matthias :+1:

PR #38722, opened 2020-09-13
Show preview icons in rule based labeling list of rules and vector tile labeling list of rules

nyalldawson:
Makes it MUCH easier to identify specific rules in the list, and matches
appearance with the rule based renderer/vector tile renderer list appearances.

image


nirvn (approved):
Beautiful. One of these things that make you go "hm, how did we ever not have that to begin with" :)

PR #38725, opened 2020-09-14
Optimise rendering of complex vector tiles

nyalldawson:
Instead of ALWAYS converting all features in a tile to QGIS representations,
now we intelligently skip over any layers which aren't required for
rendering or labeling (e.g. because the current renderer/labeling
configuration is disabling these layers or doesn't have a rule for
them).

This improves rendering speed with sources like the OS ZoomStack tiles,
which have a LOT of detail even at small map scales (e.g. building
data is present in very zoomed out tiles!!).


nyalldawson:
(For further reference, the actual issue here is that the ZoomStack  tiles have far too much detail at particular scales.


At the same zoom level:

Zoomstack
image

ESRI World Basemap

image

There's just WAY too many features in the Zoomstack tiles for the scale! (@saberraz I suggest NOT using these tiles when showcasing the vector tile support in QGIS, as the rendering speed gives the incorrect impression that QGIS is at fault)
)

PR #38733, opened 2020-09-14
Only show Select and digitizing-based menus in the Edit menu

### LINKED ISSUE #38470: Minor GUI glitch: Edit menu showing Select and Add xxxx menu items while not usable
This is a minor GUI glitch, but I happened to search for something in QGIS while I did not have any project opened (see screenshot), and then in the Edit menu there were some menu-times usable (well not sub items), while I think these should be deactivated:

- Select (there is not project, nor features so nothing to select)? In this menu-item there are sub-menu items also activated

- The Add Circle/Rectangle/Regular Polgyon/Ellipse/Annotation items are activated (though they cannot do anything as their sub-items ARE deactivated.

Screenshot-20200827101746-1141x744

DelazJ:
when appropriate (if not empty project and/or a compatible layer)

fixes #38470

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -14100,6 +14100,7 @@ void QgisApp::activateDeactivateLayerRelatedActions( QgsMapLayer *layer )
 
   if ( !layer )
   {
+    menuSelect->setEnabled( false );
     mActionSelectFeatures->setEnabled( false );
     mActionSelectPolygon->setEnabled( false );
     mActionSelectFreehand->setEnabled( false );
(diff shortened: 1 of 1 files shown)


rduivenvoorde:
@DelazJ Thanks! 

I build your PR and indeed now those menu items are invisible and can be used as soon as you have an editable polygon or line layer.

I'm not sure about the Annotation layer though. If I start QGIS, and NO project is selected, that menu item does not do anything. Not before I created (some kind of) empty project.
Not sure what others think, but I would think that that item can also be inactive untill there is a project. 
BUT if that is hard, as then it should be enabled as soon as there is a valid project, which I'm sure there is a signal for... then leave it like this.

Others?

elpaso (approved):
LGTM, can we merge for now and address the annotation layers separately?

rduivenvoorde:
@elpaso sure plz merge

DelazJ:
> If I start QGIS, and NO project is selected, that menu item does not do anything. Not before I created (some kind of) empty project.

Indeed. I forgot that "No project" case (QGIS opens in dev env with a default project).
If no project is opened at all, more entries probably need to be disabled, eg in View menu (the layers related entries, zoom..), in Layer menu (overviews and remove layer) and maybe more...
Not sure it's something I could fix faster 😞 , so I'd also agree with merging. Thanks.

PR #38741, opened 2020-09-14
[DelimitedText provider] Avoid false positive detection of some date looking content as Time (fixes #38091)

### LINKED ISSUE #38091: CSV date field being interpreted as time type
**Describe the bug**
Since 3.14.0-Pi the behaviour of CSV import has changed. Previously with "Record and Fields Options -> Detect Field Types" selected UK-format date fields would result in a QString field. Interpretation as a date would be nice, I suppose,  but this is harmless. However, in 3.14 such a field is interpreted as a time, resulting in a meaningless QTime field. Dates entered in ISO 8601 format as interpreted correctly, yielding a QDate.

**How to Reproduce**

I have provided a zip containing an example project. It loads 2 CSV files. 
* `date-issue-sample.csv` contains a single point with a field named `Valid From (dd/mm/yyyy)`. The single row fills this field with a UK-format date - `01/03/2004`
*  `date-issue-sample2.csv` contains a single point with a field named `Valid From (dd/mm/yyyy)`. The single row fills this field with the same date in ISO 8601-format - `2004-03-01`

The UK date is given a QTime value of `01:03:20.000`
The ISO date is given a QDate value of `2004-03-01`

**QGIS and OS versions**

QGIS version
3.14.0-Pi
QGIS code revision
9f7028fd23
Compiled against Qt
5.11.2
Running against Qt
5.11.2
Compiled against GDAL/OGR
3.0.4
Running against GDAL/OGR
3.0.4
Compiled against GEOS
3.8.1-CAPI-1.13.3
Running against GEOS
3.8.1-CAPI-1.13.3
Compiled against SQLite
3.29.0
Running against SQLite
3.29.0
PostgreSQL Client Version
11.5
SpatiaLite Version
4.3.0
QWT Version
6.1.3
QScintilla2 Version
2.10.8
Compiled against PROJ
6.3.2
Running against PROJ
Rel. 6.3.2, May 1st, 2020
OS Version
Windows 10 (10.0)
Active python plugins
db_manager; 
MetaSearch; 
processing

**Additional context**

[Date-Issue.zip](https://github.com/qgis/QGIS/files/5011906/Date-Issue.zip)

Thank you,
Simon Melhuish, G4TJC
[Summits On The Air](https://www.sota.org.uk/)

rouault:


src/providers/delimitedtext/qgsdelimitedtextprovider.cpp
```diff
@@ -652,8 +652,21 @@ void QgsDelimitedTextProvider::scanFile( bool buildIndexes )
 
       if ( couldBeTime[i] && !couldBeDateTime[i] )
       {
-        QTime t = QTime::fromString( value );
-        couldBeTime[i] = t.isValid();
+        // Accept 12:34, 12:34:56 or 12:34:56.789
+        // We do not use QTime::fromString() as it accepts strings like
+        // 01/03/2004 as valid times
+        couldBeTime[i] = value.length() >= 5 &&
```
nirvn, line 658: @rouault , I looked into this a white back. Around qt 5.12-5.14, QTime::fromString() became a lot saner and limits the format to those values you have in your comment. It's a problem on windows as we still ship an old Qt 5 version.

There could be an ifdef >= 5.14 here to use Qt.

rouault: ok, done

nirvn (approved):
Thanks.

PR #38742, opened 2020-09-14
[QgsQuick] Added soft and hard constraints in the form

sklencar:
In the current version of QgsQuick forms, soft and hard constraints are not distinguished and used only for a form styling.
If any hard constrain is not met, the form cannot be saved. An unmet soft constrain should only notify a user.

TODO: Notify user about unmet constraints with more than coloured label.


nyalldawson:
@PeterPetrik @wonder-sk can you take a look over this one please?

PeterPetrik (approved):

PR #38750, opened 2020-09-15
Vector tile style conversions: further refinements

nyalldawson:
Aside from fixing some more issues when converting MapBox GL styles, this PR also notifies the user when a warning is raised during the conversion:

image

It's useful to advise them if a layer was not converted, or is missing a required font...

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -2089,7 +2089,7 @@ void QgisApp::handleDropUriList( const QgsMimeDataUtils::UriList &lst )
       bool res = layer->loadDefaultStyle( error, warnings );
       if ( res && !warnings.empty() )
       {
-        QString message = QStringLiteral( "<p>%1</p>" ).arg( tr( "The following warnings were generated while converting the LYR file:" ) );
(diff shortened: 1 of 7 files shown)


src/app/qgisapp.cpp
```diff
@@ -2060,10 +2078,31 @@ void QgisApp::handleDropUriList( const QgsMimeDataUtils::UriList &lst )
...
       bool ok = false;
-      layer->loadDefaultStyle( ok );
       layer->loadDefaultMetadata( ok );
+
+      QString error;
+      QStringList warnings;
+      bool res = layer->loadDefaultStyle( error, warnings );
+      if ( res && !warnings.empty() )
+      {
+        QString message = QStringLiteral( "<p>%1</p>" ).arg( tr( "The following warnings were generated while converting the LYR file:" ) );
```
nirvn, line 2092: ... while converting the LYR? :wink:

nirvn (approved):

src/app/qgisapp.cpp
nyalldawson: lol, guess where the original inspiration for this code came from?!

PR #38751, opened 2020-09-15
Fix calculation of zoom level for vector tiles

nyalldawson:
Previously the code was calculating the exact zoom level (a double
value) and then rounding this to an integer in order to determine
which rules should be applied.

This appears to violate the vector tile styling specifications,
which are designed to "round down" the zoom level, so that styling
rules like:

- layer zoom range: 12-22
- interpolated expression:
  case when @zoom_level > 12 and @zoom_level <= 14 then 1
  when @zoom_level > 14 and @zoom_level < 18 then 2
  when @zoom_level >=18 then 4 end

work correctly when the exact zoom level is just less than 12, e.g. 11.8

So now we use floor when converting a zoom level to int so that the
styling rules work correctly.

Additionally, this adds a new @vector_tile_zoom expression variable
which contains the original double value of the calculated tile zoom
(not the integer one used for layer visibility). Many mapbox GL styling
rules rely on non-integer zoom levels for interpolation, e.g.

Case when @vector_tile_zoom >= 11.2 then 4 ...

This change allows for smooth interpolation between zoom levels which
matches the web map appearance, instead of "jumpy" fixed level interpolation
we previously had.

### DIFF
--- a/src/core/expression/qgsexpression.cpp
+++ b/src/core/expression/qgsexpression.cpp
@@ -790,6 +790,7 @@ void QgsExpression::initVariableHelp()
 
   // vector tile layer variables
   sVariableHelpTexts()->insert( QStringLiteral( "zoom_level" ), QCoreApplication::translate( "variable_help", "Zoom level of the tile that is being rendered (derived from the current map scale). Normally in interval [0, 20]." ) );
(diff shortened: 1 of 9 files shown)


nirvn (approved):
Tried it, working well.

PR #38752, opened 2020-09-15
Make GPS bearing line ellipsoid aware

nyalldawson:
...instead of just a straight line drawn from the GPS point. The straight line approach is too naive and doesn't well represent the actual GPS bearing.


nyalldawson:
(att @pcav )

NathanW2:
Got a screenshot? :)

pcav:
Thanks @nyalldawson

PR #38758, opened 2020-09-15
Bugfix gh38755 expose project setinstance

### LINKED ISSUE #38755: Server API: current project is not settable anymore
See https://github.com/qgis/QGIS/pull/38488 for the discussion.

This bug affects all maintained branches!

elpaso:
Fixes #38755

Needs forward porting to all active branches

### DIFF
--- a/python/core/auto_generated/qgsproject.sip.in
+++ b/python/core/auto_generated/qgsproject.sip.in
@@ -38,6 +38,20 @@ open within the main QGIS application.
 Returns the QgsProject singleton instance
 %End
 
+    static void setInstance( QgsProject *project );
+%Docstring
+Set the current project instance to ``project``
+
+.. note::
+
+   this is used mainly by the server, which caches the projects and (potentially) needs to switch the current instance on every request
+
+.. seealso:: :py:func:`instance`
+
+.. versionadded:: 3.10
+%End
+
+
     explicit QgsProject( QObject *parent /TransferThis/ = 0 );
 %Docstring
 Create a new QgsProject.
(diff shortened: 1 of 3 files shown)


nyalldawson:
I realise this is a regression fix, but it also makes me nervous in that a plugin calling this (outside of server) will cause major breakage to qgis app. Is there any alternative approach?

elpaso:
I expected this objection but I don't really see a big issue here, the advantage is that it adds some more flexibility to the singleton management, that IMO should eventually disappear completely (QGIS 4 maybe?).

I though about exposing it through a server specific API (`QgsServer::setProjectInstance()`?) but it doesn't really belong there: it is for server consumption but who knows: there might be other standalone scripts not related to serve that could benefit of that API.

Another alternative is expose it through the config cache with an other optional argument to `QgsConfigCache` `const QgsProject *project( const QString &path, QgsServerSettings *settings = nullptr )` or with a new method.

dmarteau:
> there might be other standalone scripts not related to serve that could benefit of that API.

It might the best alternative, since restoring the setInstance call in the QgsConfigCache only does not solve the issue:  preventing to set the singleton breaks other behaviors (like expression evaluation) and has an impact on every python scripts that might use it.

 > that IMO should eventually disappear completely (QGIS 4 maybe?).

+1

nyalldawson:
If you're going to expose it, please add a:

    \warn calling this method can have serious, unintended consequences, including instability, data loss and undefined behaviour. Use with extreme caution!

nyalldawson (approved):

nyalldawson:
Fixes regression introduced in LTR => immediate merge

PR #38769, opened 2020-09-15
[GUI] Raster histogram: restore behavior before #35465 changes

rouault:
Commits 98261bcfd4bd05948bd1b2a68c741c80e4e2e515 and ebdb546c1044b793d25097ddc04ed34b5fb6d956
for #35465 have changed the logic to compute the default number of bins
for the histogram. While appropriate in the context of #35465, that tends
to augment the number of bins, which is inappropriate for the raster histogram
chart, where beyond 1000, the chart becomes unreadable.

However I think the logic for plotting the raster histogram should probably be
revised, but I'm not clear how: ask the user for a number of bins and/or take
into account the width in pixels of the chart to determine the number of bins
(should that depend if the user has zoomed in... ?)

### DIFF
--- a/src/gui/raster/qgsrasterhistogramwidget.cpp
+++ b/src/gui/raster/qgsrasterhistogramwidget.cpp
@@ -49,9 +49,7 @@
 #include <time.h>
 #endif
 
-// this has been removed, now we let the provider/raster interface decide
-// how many bins are suitable depending on data type and range
-//#define RASTER_HISTOGRAM_BINS 256
+constexpr int SAMPLE_SIZE = 250000; // number of sample cells
 
 QgsRasterHistogramWidget::QgsRasterHistogramWidget( QgsRasterLayer *lyr, QWidget *parent )
   : QgsMapLayerConfigWidget( lyr, nullptr, parent )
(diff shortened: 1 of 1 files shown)


elpaso:
@rouault IIRC I shortly investigated the literature about how to choose the appropriate number of bins (for continuous ranges) and I didn't find a solution, it seems that it really depends on data, number of samples, range etc and there are several different approaches according to https://en.wikipedia.org/wiki/Histogram#Number_of_bins_and_width

rouault:
IMHO, showing the histogram as a continuous line is probably not ideal. This should rather be vertical bars, with a > 1 pixel width.  

(The absolute numbers in the vertical axis are also probably not significant as they depend on how many pixels have been sampled. A percentage of the total number of pixels samples would likely be better, but that doesn't impact the general issue of the readability of the plot due to choices made on the horizontal axis)

PedroVenancio:
@rouault @elpaso 

In fact, the solution implemented in https://github.com/qgis/QGIS/issues/35465 keeps failing determining quantiles with some rasters. Here is an example where, again, quantiles are wrongly calculated in raster, but correctly in vector (Raster pixels to points)

imagem

But I'm seeing some other inconsistencies with raster symbology. Please look, for instance, at this screencast:

https://cld.pt/dl/download/02852983-9f11-439d-bc81-33b58ade21bc/raster_symbology.mp4 

Or for this, with "no data" that change from transparent to coloured, just with zoom in/out. It is rendered as singleband pseudocolor with discrete interpolation (pixels with value 255 are masked as "no data value" in properties):

https://cld.pt/dl/download/d31eba1a-c6cf-446d-b151-10d8764cf909/no_data_transparency.mp4

Do you think this is all related? Or should I open separated tickets for each one?

gioman:
> In fact, the solution implemented in #35465 keeps failing determining quantiles with some rasters.

@PedroVenancio is there any ticket we should reopen?

PedroVenancio:
> > In fact, the solution implemented in #35465 keeps failing determining quantiles with some rasters.
> 
> @PedroVenancio is there any ticket we should reopen?

I think so @gioman 
https://github.com/qgis/QGIS/issues/35465#issuecomment-699636602

elpaso:
@PedroVenancio honestly I'm a bit lost in all these different issues, I'd recommend to open a separate issue for each problem you see and attach a small raster (the smallest the better) for testing.

PR #38770, opened 2020-09-15
[GUI] Identification result: fix warning 'CRITICAL Qt : QTableWidget: cannot insert an item that is already owned by another QTableWidget' (fixes #37215)

### LINKED ISSUE #37215: Switching between WMS info-output formats throws: CRITICAL  QTableWidget: cannot insert an item that is already owned by another QTableWidget
Loading a WMS layer, activating the i-tool and showing info of an object from the WMS layer shows 
a big red error in the message bar, 

Screenshot-20200615062432-748x93

and "CRITICAL    Qt : QTableWidget: cannot insert an item that is already owned by another QTableWidget"
in the Log Messages.

To reproduce

https://geodata.nationaalgeoregister.nl/bevolkingskernen2008/ows?SERVICE=WMS&
I load layer nr 2 in EPSG:28992 and click i-tool

Now click one of the object, will default to for example html or feature output, which are both fine!
NOW switch to 'Text' output. you will see the 'critical'-message.

Every time you switch to html/feature and then back to Text it will be shown.

QGIS version | 3.13.0-Master | QGIS code revision | 3ef7d89a35
-- | -- | -- | --
PostgreSQL Client Version | 12.3 (Debian 12.3-1+b1) | SpatiaLite Version | 4.3.0a
Compiled against PROJ | 7.0.1 | Running against PROJ | Rel. 7.0.1, May 1st, 2020
OS Version | Debian GNU/Linux bullseye/sid | This copy of QGIS writes debugging output.

rouault:


src/app/qgsidentifyresultsdialog.cpp
```diff
@@ -1095,7 +1095,6 @@ void QgsIdentifyResultsDialog::addFeature( QgsRasterLayer *layer,
     {
       auto valueLabel { qgis::make_unique<QLabel>( links ) };
       valueLabel->setOpenExternalLinks( true );
-      tblResults->setItem( j, 3, item );
```
nyalldawson, line 1098: I think this should be `tblResults->setItem( j, 3, new QTableWidgetItem( it.value() ) )`, just like the `!foundLinks` branch. But then we could just move the `tblResults->setItem( j, 3, new QTableWidgetItem( it.value() ) )` outside of the `if ( foundLinks )` condition entirely...

rouault: good point. done

nyalldawson (approved):

PR #38785, opened 2020-09-16
Better handling of postgres arrays with listwidget

### LINKED ISSUE #38176: QGIS inserts empty array into nullable array fields by default when creating new features in a PostGIS layer
**Describe the bug**

Given the following PostGIS table schema:

```
CREATE TABLE test
(
    id SERIAL PRIMARY KEY,
    geom GEOMETRY(LineString, 4326) NOT NULL,
    array INTEGER[]
)
```

QGIS will insert an empty PostgreSQL array (i.e. `{}`) into the `array` field by default when creating new features whilst editing the `test` PostGIS layer. However, since the `array` field is nullable (i.e. it is **not** marked `NOT NULL`), I would expect QGIS to enter `NULL` into the `array` field by default (as this would be the default value if editing the table using other software).

**How to Reproduce**

See above.

**QGIS and OS versions**

QGIS version | 3.10.7-A Coruña | QGIS code revision | 7b4ca4c8d0
-- | -- | -- | --
OS Version | Windows Server 2016 (10.0)
Active python plugins | profiletool; quick_map_services; valuetool; db_manager; MetaSearch; processing

**Additional context**

### LINKED ISSUE #38784: Postgres arrays not saved using ListWidget when adding multiple features simultaneously
This bug was noticed in master while investigating #38176. When you have a postgres table with an int[] field, and you add more than one feature at a time using the default listwidget the resulting int[] fields are null  post saving the session, no matter what information you put in them, despite saving without error.

Investigation revealed that adding a single feature did not cause this behaviour. Seems to be a bug in the code that deals with prepared statements in Postgres.

Unclear how long this bug has been present - is definitely in 3.14.

Pull request is incoming.

stev-0:
While trying to fix #38176 I came across another issue with postgres arrays and the list widget - see #38784.
Fixes #38784 #38176

Tests still in progress.

### DIFF
--- a/src/gui/editorwidgets/qgslistwidgetwrapper.cpp
+++ b/src/gui/editorwidgets/qgslistwidgetwrapper.cpp
@@ -70,16 +70,25 @@ QVariant QgsListWidgetWrapper::value() const
 {
   QVariant::Type type = field().type();
   if ( !mWidget ) return QVariant( type );
+  const QVariantList list = mWidget->list();
   if ( type == QVariant::StringList )
   {
     QStringList result;
-    const QVariantList list = mWidget->list();
     for ( QVariantList::const_iterator it = list.constBegin(); it != list.constEnd(); ++it )
       result.append( it->toString() );
     return result;
   }
   else
-    return QVariant( mWidget->list() );
+  {
+    if ( list.size() == 0 )
+    {
+      return QVariant( );
+    }
+    else
+    {
+      return list;
+    }
+  }
 }
 
 void QgsListWidgetWrapper::onValueChanged()
--- a/src/gui/editorwidgets/qgslistwidgetwrapper.cpp
+++ b/src/gui/editorwidgets/qgslistwidgetwrapper.cpp
@@ -71,6 +71,10 @@ QVariant QgsListWidgetWrapper::value() const
   QVariant::Type type = field().type();
   if ( !mWidget ) return QVariant( type );
   const QVariantList list = mWidget->list();
+  if ( list.size() == 0 && config( QStringLiteral( "EmptyIsNull" ) ).toBool() )
+  {
+    return QVariant( );
+  }
   if ( type == QVariant::StringList )
   {
     QStringList result;
@@ -80,14 +84,7 @@ QVariant QgsListWidgetWrapper::value() const
   }
   else
   {
-    if ( list.size() == 0 )
-    {
-      return QVariant( );
-    }
-    else
-    {
-      return list;
-    }
+    return list;
   }
 }
 
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -2500,7 +2500,25 @@ bool QgsPostgresProvider::addFeatures( QgsFeatureList &flist, Flags flags )
         }
         else
         {
-          v = paramValue( value.toString(), defaultValues[ i ] );
+          // the conversion functions expects the list as a string, so convert it
+          if ( value.type() == QVariant::StringList )
+          {
+            QStringList list_vals = value.toStringList();
+            // all strings need to be double quoted to allow special postgres
+            // array characters such as {, or whitespace in the string
+            // but we need to escape all double quotes and backslashes
+            list_vals.replaceInStrings( "\\", "\\\\" );
+            list_vals.replaceInStrings( "\"", "\\\"" );
+            v = QStringLiteral( "{\"" ) + value.toStringList().join( QStringLiteral( "\",\"" ) ) + QStringLiteral( "\"}" );
+          }
+          else if ( value.type() == QVariant::List )
+          {
+            v = "{" + value.toStringList().join( "," ) + "}";
+          }
+          else
+          {
+            v = paramValue( value.toString(), defaultValues[ i ] );
+          }
 
           if ( v != value.toString() )
           {
--- a/tests/src/gui/testqgslistwidget.cpp
+++ b/tests/src/gui/testqgslistwidget.cpp
@@ -20,19 +20,38 @@
 #include <qgslistwidget.h>
 #include <editorwidgets/core/qgseditorwidgetwrapper.h>
 #include <qgsapplication.h>
+#include <qgsvectorlayer.h>
+#include <editorwidgets/qgslistwidgetwrapper.h>
 
 class TestQgsListWidget : public QObject
 {
     Q_OBJECT
+  private:
+    QString dbConn;
   private slots:
     void initTestCase() // will be called before the first testfunction is executed.
     {
       QgsApplication::init();
       QgsApplication::initQgis();
+      dbConn = getenv( "QGIS_PGTEST_DB" );
+      if ( dbConn.isEmpty() )
+      {
+        dbConn = "service=\"qgis_test\"";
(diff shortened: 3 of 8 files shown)


src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -2502,6 +2502,16 @@ bool QgsPostgresProvider::addFeatures( QgsFeatureList &flist, Flags flags )
...
           v = paramValue( value.toString(), defaultValues[ i ] );
 
+          // the conversion functions expects the list as a string, so convert it
+          if (value.type() == QVariant::StringList)
+          {
+              v = "{\"" + value.toStringList().join("\",\"") + "\"}";
+          }
+          else if (value.type() == QVariant::List)
+          {
+              v = "{" + value.toStringList().join(",") + "}";
```
nyalldawson, line 2512:
```suggestion
              v = '{' + value.toStringList().join( ',' ) + '}';
```

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -2502,6 +2502,16 @@ bool QgsPostgresProvider::addFeatures( QgsFeatureList &flist, Flags flags )
         {
           v = paramValue( value.toString(), defaultValues[ i ] );
 
+          // the conversion functions expects the list as a string, so convert it
+          if (value.type() == QVariant::StringList)
+          {
+              v = "{\"" + value.toStringList().join("\",\"") + "\"}";
```
nyalldawson, line 2508:
```suggestion
              v = QStringLiteral( "{\"" ) + value.toStringList().join( QStringLiteral( "\",\"") ) + QStringLiteral( "\"}" );
```

rouault: This lacks logic to escape characters like double-quote. I'm wondering if you shouldn't use the quotedValue() method for this case and the below one.

A new test in tests/src/python/test_provider_postgres.py would be welcome

stev-0: I started the tests in c++ in testqgslistwidget.cpp as I thought most of this would relate to list widget only, so added that for now, but I can see the potential benefit of having it further up the chain so I could translate into python if it's considered worthwhile to do so.

I didn't use quotedValue() in the end as I think the array syntax is slightly different to normal strings, and so all that I think is needed is to escape double quotes within the string.

stev-0:
Re-triggering tests as OGC failed for extraneous reasons

tests/src/gui/testqgslistwidget.cpp
```diff
@@ -120,6 +137,107 @@ class TestQgsListWidget : public QObject
       QCOMPARE( wrapper->value().toList(), expected );
       QVERIFY( widget->valid() );
     }
+
+    void testPostgres()
+    {
+      //create pg layers
+      QString dbConn = getenv( "QGIS_PGTEST_DB" );
```
stev-0, line 144: Should put DB init in test setup to avoid repitition

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -2502,6 +2502,21 @@ bool QgsPostgresProvider::addFeatures( QgsFeatureList &flist, Flags flags )
...
           v = paramValue( value.toString(), defaultValues[ i ] );
 
+          // the conversion functions expects the list as a string, so convert it
+          if ( value.type() == QVariant::StringList )
+          {
+            QStringList list_vals = value.toStringList();
+            // all strings need to be double quoted to allow special postgres
+            // array characters such as {, or whitespace in the string
+            // but we need to escape all double quotes
+            list_vals.replaceInStrings( "\"", "\\\"" );
```
rouault, line 2512: Shouldn't slashes be escaped too ? (before doing the double quote escaping)

stev-0: yes, correct, thank you

src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -2502,6 +2502,22 @@ bool QgsPostgresProvider::addFeatures( QgsFeatureList &flist, Flags flags )
         {
           v = paramValue( value.toString(), defaultValues[ i ] );
```
peetw, line 2503: Could this be moved to an `else` statement below to avoid immediately reassigning it in the case where `value` is a `StringList` or `List`?

stev-0: Yup, fair point, flows better that way I think

stev-0:
It feels like the choice between an empty list widget equating to null or equating to an empty list should be a user configurable setting in the widget config as there are a number of possible configurations depending on how the underlying database / provider is set up, and no obvious default. 

I will try and set this up in the List Widget UI.

src/gui/editorwidgets/qgslistwidgetwrapper.cpp
```diff
@@ -70,16 +70,32 @@ QVariant QgsListWidgetWrapper::value() const
...
   if ( !mWidget ) return QVariant( type );
+  const QVariantList list = mWidget->list();
   if ( type == QVariant::StringList )
   {
-    QStringList result;
-    const QVariantList list = mWidget->list();
-    for ( QVariantList::const_iterator it = list.constBegin(); it != list.constEnd(); ++it )
-      result.append( it->toString() );
-    return result;
+    if ( list.size() == 0 && config( QStringLiteral( "EmptyIsNull" ) ).toBool() )
```
peetw, line 76: This `if` statement and block looks to be duplicated on L:90-93 below. If that is intentional, I think it could be de-duplicated and pulled out to be a "pre-check" outside of the current containing if statement (i.e. it can be inserted in-between L:73-74 above), which would help reduce the nesting and make the control flow a bit easier to follow.

peetw:
> It feels like the choice between an empty list widget equating to null or equating to an empty list should be a user configurable setting in the widget config as there are a number of possible configurations depending on how the underlying database / provider is set up, and no obvious default.

@stev-0 If I've interpreted the changes correctly, the default will be to treat empty as null, with the ability to switch it to treat empty as an empty array if required. Is that correct? If so, seems reasonable to me and fits with what happens by default for things like a text field, where it will insert null rather than an empty string if the field is left empty.

stev-0:
> @stev-0 If I've interpreted the changes correctly, the default will be to treat empty as null, with the ability to switch it to treat empty as an empty array if required. Is that correct? If so, seems reasonable to me and fits with what happens by default for things like a text field, where it will insert null rather than an empty string if the field is left empty.

Yes, that's right.

src/gui/editorwidgets/qgslistwidgetwrapper.cpp
stev-0: yup, good point, 👍

stev-0:
I think this is in a good state to be merged - three people have reviewed it and tests passing. Aware that there are **a lot** of PRs at the moment though!

nyalldawson:
@m-kuhn does this look ok to you? I can't see any issues and believe it's mergable..

nyalldawson:
Let's just do this :+1:

nyalldawson:
Manually backported to https://github.com/qgis/QGIS/pull/40712

PR #38788, opened 2020-09-16
[browser] Correctly save expanded node state when closing QGIS

nyalldawson:
Otherwise browser expanded nodes are never saved, and the browser
always returns to some random previous state


nirvn:
OMG YESSSS PLEASE!

elpaso:
Long awaited, thank you!

PR #38797, opened 2020-09-16
Expression preview widget: do not disable feature picker (fixes #38646)

### LINKED ISSUE #38646: Preview of Expression String Builder 'freezes' if Feature id higher then available
Little movie to show issue: [Expression.zip](https://github.com/qgis/QGIS/files/5188114/Expression.zip)

Little dataset to test: [10records.zip](https://github.com/qgis/QGIS/files/5188118/10records.zip)

See screencast: the Expression String Builder has the ability to walk through the features to see the value of the expression.
For example in the attached dataset of 10(!) records and wanting to set the size of the symbol to some datadefinied rule, you open the Expression Builder, and:
- select one field (in the movie it is 'timestamp')
- now you can go over the features by either clicking the < and > button, of by pulling down the drop down, OR (!) by filling in a number in the input
- IF you fill in a valid number (in this case between 1 and 10) all is fine
- BUT if you fill in an invalid number (eg 20) then your 'preview' is not working anymore (which is fine), BUT you cannot go back: at that stage both the < and > buttons, nor the dropdown nor the input is working to fix you mistake...
The 'Feature' input is greyed out, and you cannot delete the string or so, and the 'Preview' sticks to NULL:

Screenshot-20200908140843-668x351

Only after closing the Expression String Builder dialog and reopening it, all is working fine again.

Not sure what a good solution is here?
- before accepting the input, check if it is a valid one?
- NOT making the input work as input (not a good idea for big datasets...)
- make it possible to 're-edit' the input after a faulty input (if I could just delete the 20, and insert a good value all would be fine)

What do others think?

rouault:
In the text edit of the feature picker, if one currently enters an invalid value,
the whole feature picker becomes disabled, which requires to close and re-open
the expression string builder dialog.

So instead of doing this, use the preview label to indicate that a valid
feature is needed to evaluate the expression, for expressions that require feature
fields.

Previous related pull requests are:
https://github.com/qgis/QGIS/pull/37518
https://github.com/qgis/QGIS/pull/37139

### DIFF
--- a/src/gui/qgsexpressionpreviewwidget.cpp
+++ b/src/gui/qgsexpressionpreviewwidget.cpp
@@ -49,20 +49,7 @@ void QgsExpressionPreviewWidget::setCurrentFeature( const QgsFeature &feature )
 {
   // todo: update the combo box if it has been set externaly?
 
-  // force the feature to be valid, so it can evaluate an invalid feature but having its fields set
-  if ( !feature.isValid() )
-  {
-    QgsFeature validFeature( feature );
-    validFeature.setValid( true );
-    mExpressionContext.setFeature( validFeature );
-    mFeaturePickerWidget->setEnabled( false );
-    mFeaturePickerWidget->setToolTip( tr( "No feature was found on this layer to evaluate the expression." ) );
-  }
-  else
-  {
-    mExpressionContext.setFeature( feature );
-    mFeaturePickerWidget->setEnabled( true );
-  }
+  mExpressionContext.setFeature( feature );
   refreshPreview();
 }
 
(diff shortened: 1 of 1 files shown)


src/gui/qgsexpressionpreviewwidget.cpp
```diff
@@ -49,20 +49,7 @@ void QgsExpressionPreviewWidget::setCurrentFeature( const QgsFeature &feature )
 {
   // todo: update the combo box if it has been set externaly?
 
-  // force the feature to be valid, so it can evaluate an invalid feature but having its fields set
-  if ( !feature.isValid() )
-  {
-    QgsFeature validFeature( feature );
-    validFeature.setValid( true );
```
3nids, line 56: if you remove this, that will bring back the issue when a layer is empty and you try an expression (https://github.com/qgis/QGIS/issues/37100).
Or am I missing something?

rouault: With this pull request, if the layer is empty (or you specify an invalid FID),
- if the expression doesn't use fields, then it is evaluated
- if the expression uses fields, the preview label displays "No feature was found on this layer to evaluate the expression" (not that the expression is invalid)

3nids: ok, that makes sense.

3nids (approved):

PR #38805, opened 2020-09-16
QgsField::ConfigurationFlags: use a negative form for the flags so th…

3nids:
…at default flags is None

this will make this futureproof, so newly added flags won't cause trouble for compatibility with old projects


src/core/qgsfield.h
```diff
@@ -80,10 +81,9 @@ class CORE_EXPORT QgsField
 #endif
     {
       None = 0, //!< No flag is defined
-      Searchable = 1 << 1, //!< Defines if the field is searchable (used in the locator search for instance)
-      ExposeViaWms = 1 << 2, //!< Fields is available if layer is served as WMS from QGIS server
-      ExposeViaWfs = 1 << 3, //!< Fields is available if layer is served as WFS from QGIS server
-      DefaultFlags = Searchable | ExposeViaWms | ExposeViaWfs, //!< Default set of flags for a field
+      NotSearchable = 1 << 1, //!< Defines if the field is searchable (used in the locator search for instance)
+      DoNotExposeViaWms = 1 << 2, //!< Fields is available if layer is served as WMS from QGIS server
```
nyalldawson, line 85: HideFromWms/HideFromWfs might be slightly nicer wording?

nyalldawson (approved):

PR #38807, opened 2020-09-16
GetFeatureInfo and no data values

### LINKED ISSUE #37781: Server GetFeatureInfo returns 0 where is should return NULL
QGIS Server 3.10 on Ubuntu Server 20.04. QGIS Desktop 3.10 on Windows/Linux.

Publish a singleband raster as WMS, one that has at least some areas with NULL values.

Do a GetFeatureInfo on the WMS layer, in a place where the datasource has a NULL value, QGIS Server returns 0 instead.

pblottiere:
## Description

Fixes https://github.com/qgis/QGIS/issues/37781.

I kept the `no data` value like in QGIS Desktop, but `null` is maybe more appropriate in the QGIS Server context. Opinions?


gioman:
@pblottiere thanks!

elpaso:
for json output I would set it to `null`, unsure about the text and html, maybe `no data` is clearer.

elpaso:
@pblottiere would you mind adding a test for json too?

pblottiere:
Hi @elpaso,

> for json output I would set it to null, unsure about the text and html, maybe no data is clearer.

> would you mind adding a test for json too?

Yes I'm going to address your two comments in the coming days.

pblottiere:
@elpaso Your comments have been addressed.

Issue #37401, opened 2020-06-25
if shortcut is set for copy features then Ctrl+C not working anymore


MMR100:
In QGIS 3.14, the Ctrl+C shortcutis not working anymore. 
As far as I understand, it has been removed in this version due to issues with Python Console. 
However I am not using the console so I decided to set up again Ctrl+C as a shortcut for copying features in keyboard shortcuts. 
The shortcut is not working. 
If I set up another shortcut (e.g. Ctrl+W), it is well working. 
I don't mind to use a new shortcut instead of the classic Crtl+C, but I think the user should be informed that Ctrl+C cannot be chose as a shortcut for copying features.

nirvn:
@MMR100 , Ctrl+C is still working, but it's scope has been limited to avoid interference with panels. If you left click on the map canvas (with selected features) and click on Ctrl+C, features should be copied. Can you confirm?

MMR100:
I found the error. 
If no shortcut is set for copy feature, the Ctrl+C is working. 
If I set up Ctrl+C in the list of keyboard shortcut, then it is not working anymore. 
Strange??

MMR100:
I checked for Ctrl + X, Ctrl+V and Ctrl+A. The same problem occurs.

troopa81:
I think it's a duplicate of #37503. I close the ticket, please re-open it if the problem still persists on master after #38810 merging

thbaumann:
The problem still exists in QGIS 3.28 / 3.30 / 3.32. As soon as in the shortcut options of QGIS CTRL+C and CTRL+V is set then features won't be copied. As soon as you remove the shortcut-assignment it works again.

MieWinstrup:
It is still a problem in QGIS 3.40 / 3.44 and 3.99 (the nightly build). I have tried with a clean user profile, and I'm only able to copy features from one layer to another layer with the shortcuts CTRL+C and CTRL+V - as long as I haven't set up a shortcut. That seems weird.

troopa81:
@MieWinstrup I just tried it with current master and it works. Could you describe exactly what you are doing and share a project sample so I could reproduce ?

Let's reopen the issue in the mean time

MieWinstrup:
Before I set up shortcuts, I can copy points from one layer and copy to another pointlayer with CTRL+C and CTRL+V.

When I set up a shortcut for Copy features and Paste features - then I can not use the shortcuts.

[...]

PR #38810, opened 2020-09-16
Add actions to the attribute table view so they get triggered by shortcut

### LINKED ISSUE #37503: the keybord shortcut CTRL+C copy only the first cell, not the whole line
Hi there,
On the Attribute table, the keybord shortcut CTRL+C copy only the first cell, not the whole line.
It's works fine when using the button.
QGIS 3.14.0-Pi 64bit - release 9f7028fd23
thank's
Thomas.

troopa81:
## Description

Fixes #37503 and #37401 

According to *Qt::WidgetWithChildrenShortcut* documentation

> The shortcut is active when its parent widget, or any of its children has focus. Children which are top-level widgets, except pop-ups, are not affected by this shortcut context.

But in this case the parent is the menubar, not the dialog. So we need to add the actions to the QgsDualView in order to make the shortcut work when the QgsDualView has the focus.


troopa81:
Only backport to 3.10 if #36955 is backported

nyalldawson:
#36955  won't be backported -- it's too risky and not important enough for inclusion in LTR.

nyalldawson (approved):

PR #38814, opened 2020-09-16
Don't html-escape log message in processing panel

### LINKED ISSUE #37934: Greater-than and ampersand sign displayed as &gt; and &amp; in Log Message of Processing
**Describe the bug**
The sign `>` (greater-than) and ampersand `&` are displayed with their HTML equivalent `&gt;` and `&amp;` in the Log panel of a custom processing tool, but similar symbols like `< ≤ ≥` are displayed correctly.

**How to Reproduce**

In the `def processAlgorithm(self, parameters, context, feedback):` of a custom processing tool, add the lines:
```python
feedback.pushInfo(">")
feedback.pushInfo(u"\u003E")
feedback.pushInfo("&")
feedback.pushInfo("<")
feedback.pushInfo("& > < ≤ ≥")
```

The first two lines will show `&gt;` and the third `&amp;`, but the last two will display the characters correctly.

**QGIS and OS versions**
QGIS 3.12.2-București
QGIS code revision 8a1fb33634
Compiled against Qt 5.11.2
Windows 10 (10.0)

pblottiere:
## Description

aa

Fixes https://github.com/qgis/QGIS/issues/37934.

### DIFF
--- a/src/gui/processing/qgsprocessingalgorithmdialogbase.cpp
+++ b/src/gui/processing/qgsprocessingalgorithmdialogbase.cpp
@@ -403,7 +403,7 @@ void QgsProcessingAlgorithmDialogBase::reportError( const QString &error, bool f
 
 void QgsProcessingAlgorithmDialogBase::pushInfo( const QString &info )
 {
-  setInfo( info );
+  setInfo( info, false, false );
   processEvents();
 }
 


nyalldawson:
This fix is not quite correct -- because with it pushing a string like "<b>2</b>" incorrectly treats the string as HTML. That might seem correct for the log window, but we actually don't support HTML tags in the feedback methods as not all QgsProcessingFeedback subclasses support formatted output (e.g. when running using qgis_process).

https://github.com/qgis/QGIS/pull/38822 is an alternative approach which fixes both strings

pblottiere:
OK thanks for the explanation @nyalldawson :+1:

PR #38826, opened 2020-09-17
Fix glaring plugin integration gap, add an interface to add pages to the project properties dialog

nirvn:
## Description

This PR 'fixes' a glaring plugin integration gap by adding an interface to add page(s) to the project properties dialog. This is how it looks:
Screenshot from 2020-09-17 11-38-13

I would like to call for this PR to be merged into upcoming 3.16 (which is set to become our new LTR). Rational for doing so:
- virtually all code addition is in /src/app
- code added has been well tested as it's essentially mirroring what the interface has had for adding pages to the options dialog since 3.0
- IMHO, we really want this in this LTR to be as useful as it possibly can be for plugin developers.

### DIFF
--- a/src/app/qgsprojectproperties.cpp
+++ b/src/app/qgsprojectproperties.cpp
@@ -76,7 +76,7 @@
 //qt includes
 #include <QInputDialog>
 #include <QFileDialog>
-#include <QHeaderView>  // Qt 4.4
+#include <QHeaderView>
 #include <QMessageBox>
 #include <QDesktopServices>
 #include <QAbstractListModel>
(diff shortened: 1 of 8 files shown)


src/app/qgsprojectproperties.cpp
```diff
@@ -1525,6 +1543,12 @@ void QgsProjectProperties::apply()
 
   QgsProject::instance()->displaySettings()->setBearingFormat( mBearingFormat->clone() );
 
+  const auto constMAdditionalProjectPropertiesWidgets = mAdditionalProjectPropertiesWidgets;
+  for ( QgsOptionsPageWidget *widget : constMAdditionalProjectPropertiesWidgets )
```
nyalldawson, line 1547:
```suggestion
  for ( QgsOptionsPageWidget *widget : qgis::as_const( mAdditionalProjectPropertiesWidgets ) )
```

nyalldawson (approved):

PR #38828, opened 2020-09-17
[themes] Fix blend of gray checkbox glitches when embedded into table (et cie) widgets

nirvn:
## Description

Before (left) vs. fixed (right):
image


nyalldawson:
Ewwww

nirvn:
@nyalldawson , that's an understatement :laughing:

PR #38831, opened 2020-09-17
Check if a point is 3D before returning its z value

### LINKED ISSUE #37772: z function returns Nan when used with a 2d point geometry
In expression builder (I tested the expression preview, I didn't try to calculate values in a field):
* z_min( geom_from_wkt( 'POINT ( 0 0 )' ) ) returns Null. Same for z_max.
* z( geom_from_wkt( 'POINT ( 0 0 )' ) )  returns nan
* z( geom_from_wkt( 'MULTIPOINT ( 0 0 , 1 5)')) returns null

Not sure that nan means a lot to most of us. Could be nice to instead display null.

pblottiere:
## Description

Fixes https://github.com/qgis/QGIS/issues/37772.

A unit test is in progress.

### DIFF
--- a/src/core/expression/qgsexpressionfunction.cpp
+++ b/src/core/expression/qgsexpressionfunction.cpp
@@ -2349,6 +2349,9 @@ static QVariant fcnGeomZ( const QVariantList &values, const QgsExpressionContext
   if ( geom.isNull() )
     return QVariant(); //or 0?
 
+  if ( !geom.constGet()->is3D() )
+    return QVariant();
+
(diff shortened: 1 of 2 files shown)


lbartoletti:
BTW, there is the same problem with M point. So you may add this in `fcnGeomM`

```
if ( !geom.constGet()-> isMeasure() )
    return QVariant();
```

pblottiere:
> BTW, there is the same problem with M point.

Thanks for the input @lbartoletti, I take a look.

PR #38841, opened 2020-09-17
[Processing][GDAL] Clip Raster by extent - override CRS

rldhont:
## Description

The GDAL algorithm Clip raster by extent does not provide the capabilities to override the output CRS.

In some cases, QGIS does not read the projection defined in the input layer. The user has defined the layer's CRS, but if QGIS has not read the INPUT CRS, the GDAL algorithm does the same so the OUTPUT layer has no CRS. 

The work around is to use the GDAL Assign projection algorithm.

The solution would be to override the OUTPUT layer's CRS by using `-a_srs` in the command line.

We propose to fix it by adding a boolean parameter to override the OUTPUT layer's CRS by the INPUT layer's one.

### DIFF
--- a/python/plugins/processing/algs/gdal/ClipRasterByExtent.py
+++ b/python/plugins/processing/algs/gdal/ClipRasterByExtent.py
@@ -33,6 +33,7 @@ from qgis.core import (QgsRasterFileWriter,
                        QgsProcessingParameterExtent,
                        QgsProcessingParameterString,
                        QgsProcessingParameterNumber,
+                       QgsProcessingParameterBoolean,
                        QgsProcessingParameterRasterDestination)
 from processing.algs.gdal.GdalAlgorithm import GdalAlgorithm
 from processing.algs.gdal.GdalUtils import GdalUtils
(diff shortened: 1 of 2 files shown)


rldhont:
branch rebased and ready for review

nyalldawson:
@alexbruy do you want to review this one?

python/plugins/processing/algs/gdal/ClipRasterByExtent.py
```diff
@@ -133,6 +139,10 @@ def getConsoleCommands(self, parameters, context, feedback, executing=True):
         arguments.append(str(bbox.xMaximum()))
         arguments.append(str(bbox.yMinimum()))
 
+        crs = inLayer.crs()
+        if override_crs and crs.isValid():
+            arguments.append('-a_srs {}'.format(crs.toProj()))
```
alexbruy, line 144: This is no safe, better to use `gdal_crs_string()` method here.

alexbruy (changes requested):
Would you mind adding a small test with the new parameter?

python/plugins/processing/algs/gdal/ClipRasterByExtent.py
rldhont: Thanks for the review. I add the use of `GdalUtils.gdal_crs_string()`.

PR #38842, opened 2020-09-17
hide splash screen when loading the project

### LINKED ISSUE #27051: Splash Screen Persists While 'Handle Bad Layers' Dialog is Open
Author Name: **Brendan Heberlein** (Brendan Heberlein)
Original Redmine Issue: [19222](https://issues.qgis.org/issues/19222)
Affected QGIS version: 3.0.2
Redmine category:gui

---
If I move or delete files used by a QGIS project & then reopen the project, the splash screen does not go away until after I have closed the 'handle bad layers' dialog window. The splash screen remains on top & is rather annoying to work around.

Running Mac OS 10.13.3 on a MacBook Pro Retina (Early 2015).

### LINKED ISSUE #34784: Application splash screen makes Handle Unavailable Layers dialog unusable
**Describe the bug**

When opening a project with a missing layer on Mac OS, the application splash screen remains in the foreground obscuring the layer update window:

QGIS-foreground

**How to Reproduce**

Open a project with a missing or broken layer.

**QGIS and OS versions**

QGIS 3.4 LTS and QGIS 3.10 LTS on Mac OS 10.12.

PeterPetrik:
fix https://github.com/qgis/QGIS/issues/27051
fix https://github.com/qgis/QGIS/issues/34784

I am not sure this is correct fix, since now the splash screen is not visible during project load even there are no broken layers.
- Maybe better to not show it once hidden? 
- Or somehow pass splash object to `QgsHandleBadLayersHandler::handleBadLayers` and temporary hide it there?

### DIFF
--- a/src/app/main.cpp
+++ b/src/app/main.cpp
@@ -1394,6 +1394,9 @@ int main( int argc, char *argv[] )
   /////////////////////////////////////////////////////////////////////
   if ( ! sProjectFileName.isEmpty() )
   {
+    // in case the project contains broken layers, interactive
+    // "Handle Bad Layers" is displayed that could be blocked by splash screen
+    mypSplash->hide();
     qgis->openProject( sProjectFileName );
   }
 


nyalldawson:
I think this is OK -- I don't see an issue with closing the splash screen when the project starts loading, by that stage we will already have the application window constructed and providing feedback regarding the project load progress...

PeterPetrik:
so @nyalldawson you suggest to not show() it afterwards, just hide?

nyalldawson:
Yep. It probably made sense a while back, but now we show project load progress reports in the main window

nyalldawson (approved):

PR #38847, opened 2020-09-17
Coordinate operation widget: avoid repeating scope and remarks

rouault:
For coordinate operations that are not concatenated operations, we
currently display twice the scope and remarks. Let's strip a bit of
text to avoid confusing users even more :-)

Before:

After:
image


nyalldawson (approved):

nyalldawson:
Unrelated test failure

PR #38848, opened 2020-09-17
Hold the GIL for more geometry related methods

nyalldawson:


nyalldawson:
Unrelated test failure (@pblottiere the new OGC test is failing on a number of PRs for unrelated reasons)

PR #38854, opened 2020-09-18
Disable labels when widget is disabled in map item grid properties

DelazJ:
the aim is to fix the labels being enabled while their widget is off
image

but i think we can do better:
- [ ] use a checkbox with its own label for "Rotate ticks" instead of an additional label
- [ ] move the distance "margin to map corner" above the "rotate tick" option; I think it relates to grid being rotated and not to ticks being rotated
- [ ] I think there's a bug in that the rotation properties are enabled if tick is selected as frame while they should be enabled when the map item has a rotation applied to map AND tick is selected instead.
- [ ] extend the above logic to annotation

That said, I'm not a user of this feature so probably miss some info. 
@andreasneumann @olivierdalang

### DIFF
--- a/src/gui/layout/qgslayoutmapgridwidget.cpp
+++ b/src/gui/layout/qgslayoutmapgridwidget.cpp
@@ -344,10 +344,15 @@ void QgsLayoutMapGridWidget::toggleFrameControls( bool frameEnabled, bool frameF
   mRightDivisionsLabel->setEnabled( frameEnabled );
   mTopDivisionsLabel->setEnabled( frameEnabled );
   mBottomDivisionsLabel->setEnabled( frameEnabled );
+
   mRotatedTicksCheckBox->setEnabled( rotationEnabled );
+  mRotatedTicksLabel->setEnabled( rotationEnabled );
   mRotatedTicksLengthModeComboBox->setEnabled( rotationEnabled );
+  mRotatedTicksLengthModeLabel->setEnabled( rotationEnabled );
   mRotatedTicksThresholdSpinBox->setEnabled( rotationEnabled );
+  mRotatedTicksThresholdLabel->setEnabled( rotationEnabled );
   mRotatedTicksMarginToCornerSpinBox->setEnabled( rotationEnabled );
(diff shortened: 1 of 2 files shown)


nyalldawson:
"margin to map corner" should probably be "margin from map corner" or "distance to map corner", and it needs a unit prefix.

nyalldawson:
Actually should "rotate ticks" be a checkable group box? I believe all the settings below it apply only when "rotate ticks" is checked, so a group box would be more appropriate.

olivierdalang:
@DelazJ 
> * [ ]  use a checkbox with its own label for "Rotate ticks" instead of an additional label

Even better would be as @nyalldawson suggests (with a group box)

> * [ ]  move the distance "margin to map corner" above the "rotate tick" option; I think it relates to grid being rotated and not to ticks being rotated

Not really, this setting only applies to "outwards facing" ticks, to avoid ticks/annotation overshooting the frame. Again, would make more sense in a group box.

> * [ ]  I think there's a bug in that the rotation properties are enabled if tick is selected as frame while they should be enabled when the map item has a rotation applied to map AND tick is selected instead.

When the map has a rotation or when map CRS != grid CRS (in which case the lines may not be orthogonal). However, I think we can keep this enabled all the time, it's not too surprising that it has not much effect when the map isn't rotated.

> * [ ]  extend the above logic to annotation

Yes


@nyalldawson 

> "margin to map corner" should probably be "margin from map corner" or "distance to map corner", and it needs a unit prefix.

Agreed

> Actually should "rotate ticks" be a checkable group box? I believe all the settings below it apply only when "rotate ticks" is checked, so a group box would be more appropriate.

Agreed


@DelazJ let me know if you intend to do these above as part as this PR. If not I'll do it. Thanks for the improvements !

olivierdalang:
> > @DelazJ let me know if you intend to do these above as part as this PR. If not I'll do it. Thanks for the improvements !
> 
> Happy to let you take the lead @olivierdalang. Thanks.

Done here : https://github.com/qgis/QGIS/pull/39037

I'm closing this for now, don't hesitate to reopen/open an issue/comment on the other PR if I missed anything.

PR #38861, opened 2020-09-18
Check if feature is rendered before displaying its map tips

### LINKED ISSUE #37066: Map tips should not appear for invisible features
Style a layer with graduated symbology.
Enable map tips and disable some of the categories in the layers panel, or delete some categories, as per the two layers in the screenshot.
Mouse around in the map and in both cases you will find that map tips still appear for invisible features - in this case those with a value less than 20,000.
image

pblottiere:
## Description

Fixes https://github.com/qgis/QGIS/issues/37066

### DIFF
--- a/src/gui/qgsmaptip.cpp
+++ b/src/gui/qgsmaptip.cpp
@@ -224,6 +224,12 @@ QString QgsMapTip::fetchFeature( QgsMapLayer *layer, QgsPointXY &mapPosition, Qg
     renderer.reset( vlayer->renderer()->clone() );
     renderer->startRender( renderCtx, vlayer->fields() );
     filter = renderer->capabilities() & QgsFeatureRenderer::Filter;
+
+    const QString filterExpression = renderer->filter( vlayer->fields() );
+    if ( ! filterExpression.isEmpty() )
+    {
+      request.setFilterExpression( filterExpression );
+    }
   }
 
   QgsFeatureIterator it = vlayer->getFeatures( request );
--- a/src/gui/qgsmaptip.cpp
+++ b/src/gui/qgsmaptip.cpp
@@ -23,6 +23,7 @@
 #include "qgswebview.h"
 #include "qgswebframe.h"
 #include "qgsapplication.h"
+#include "qgsrenderer.h"
 #include "qgsexpressioncontextutils.h"
 
 // Qt includes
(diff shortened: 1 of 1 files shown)


src/gui/qgsmaptip.cpp
```diff
@@ -212,12 +213,30 @@ QString QgsMapTip::fetchFeature( QgsMapLayer *layer, QgsPointXY &mapPosition, Qg
...
+  renderCtx.expressionContext() << QgsExpressionContextUtils::layerScope( vlayer );
+
+  std::unique_ptr< QgsFeatureRenderer > renderer;
+  if ( vlayer->renderer() )
+  {
+    renderer.reset( vlayer->renderer()->clone() );
+    renderer->startRender( renderCtx, vlayer->fields() );
+  }
+
   QgsFeatureIterator it = vlayer->getFeatures( request );
```
nyalldawson, line 227: It's possible to optimise this request based on the renderer's visible features -- if the layer has a renderer, you can use QgsFeatureRenderer::filter to set a filter expression on the request to avoid fetching and testing some invisible features, which should speed this functionality up

nyalldawson (approved):

src/gui/qgsmaptip.cpp
pblottiere: Indeed! I take a look. Thanks for your input Nyall :+1:.

pblottiere: @nyalldawson  I think I took into account your comment. Let me know if it's not what you had in mind.

pblottiere:
I'll merge this PR in the coming days if I don't have more comments in the meantime.

src/gui/qgsmaptip.cpp
pblottiere: Mmm... In my use case with a graduated renderer and a SHP, the `renderer->filter( vlayer->fields() )` call returns an empty string so I get an "empty" iterator from `vlayer->getFeatures( request )`. So map tips are never displayed.

I can write a unit test to highlight this behavior if you want?

nyalldawson: It's not implemented for the graduated renderer -- so if filter() returns an empty string, you don't set a filter expression on the iterator. It's implemented for categorized, rule based renderers only.

pblottiere: OK good to know. The code is now updated.

PR #38868, opened 2020-09-19
Fix warning regarding ABC import from collections

tirkarthi:
## Description

Importing ABC from collections has been deprecated since Python 3.4 and will be removed in Python 3.10. Hence import ABC from collections in Python 2 and collections.abc in Python 3 for compatibility.


nyalldawson:
You can safely drop python <3.6 compatibility, no need to keep that around anymore

nyalldawson:
Superseded by https://github.com/qgis/QGIS/pull/39014

PR #38869, opened 2020-09-19
Bump build to Qt 5.15

m-kuhn:
## Description

Getting ready for the future :rocket:


m-kuhn:
This will need to wait until after 3.16 is released when we can raise the minimum version a lot of dependencies.
Currently it would just lead to a myriad of ifdefs to keep old and new code working.

PR #38870, opened 2020-09-19
Add feedback to executeSql

### LINKED ISSUE #38092: DB Manager "cancel" button to stop a query is not working
QGIS master.

Run a long query and try stop it by pressing "cancel" or "esc": nothing happens, the query will run  until the end... or you must kill QGIS.

elpaso:
Fixes #38092 by adding an optional QgsFeedback argument to
the executeSql method and by implementing the PQCancel
method in the PG provider internals.

While the cancellation works well for all supported provider while
fetching results in the loop, the cancellation of a running query is now
implemented for the postgres provider connection only because the GPKG
and GDAL both rely on GDALDatasetExecuteSQL which cannot be interrupted.

This PR also introduce a few optimizations in the PG DB-Manager
code that should probably fix also other "slowness" issues that
were reported after 3.x during PG query execution.

A small UX change in the SQL dialog makes it evident to the user that
a cancellation request has been sent to the backend: the button text
is changed to "Cancelling, please wait..." so that for
provider connections that are not able to interrupt the running query
and must wait for the fetching loop to exit from the executeSql call
the user knows that something is happening and that a cancellation
request has been successfully sent.

### DIFF
--- a/python/plugins/db_manager/dlg_sql_window.py
+++ b/python/plugins/db_manager/dlg_sql_window.py
@@ -381,7 +381,7 @@ class DlgSqlWindow(QWidget, Ui_Dialog):
 
     def executeSqlCanceled(self):
         self.btnCancel.setEnabled(False)
-        self.btnCancel.setText(self.tr("Canceling, please wait ..."))
+        self.btnCancel.setText(self.tr("Canceling…"))
         self.modelAsync.cancel()
 
     def executeSqlCompleted(self):
--- a/python/plugins/db_manager/dlg_sql_window.py
+++ b/python/plugins/db_manager/dlg_sql_window.py
@@ -381,7 +381,7 @@ class DlgSqlWindow(QWidget, Ui_Dialog):
 
     def executeSqlCanceled(self):
         self.btnCancel.setEnabled(False)
-        self.btnCancel.setText(self.tr("Cancelling, please wait ..."))
+        self.btnCancel.setText(self.tr("Canceling, please wait ..."))
         self.modelAsync.cancel()
 
     def executeSqlCompleted(self):
--- a/python/plugins/db_manager/dlg_sql_window.py
+++ b/python/plugins/db_manager/dlg_sql_window.py
@@ -381,6 +381,7 @@ class DlgSqlWindow(QWidget, Ui_Dialog):
 
     def executeSqlCanceled(self):
         self.btnCancel.setEnabled(False)
+        self.btnCancel.setText(self.tr("Cancelling, please wait ..."))
         self.modelAsync.cancel()
 
     def executeSqlCompleted(self):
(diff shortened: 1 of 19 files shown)


python/plugins/db_manager/dlg_sql_window.py
```diff
@@ -381,6 +381,7 @@ def updateUiWhileSqlExecution(self, status):
 
     def executeSqlCanceled(self):
         self.btnCancel.setEnabled(False)
+        self.btnCancel.setText(self.tr("Canceling, please wait ..."))
```
nyalldawson, line 384: What about just "Canceling" and leave the "please wait" bit for a tooltip? Otherwise the button will grow very large for this lengthy string

src/providers/postgres/qgspostgresproviderconnection.cpp
```diff
@@ -192,24 +192,52 @@ void QgsPostgresProviderConnection::renameSchema( const QString &name, const QSt
...
   else
   {
+
+    if ( feedback && feedback->isCanceled() )
+    {
+      return results;
+    }
+
+    // This is gross but I tried with both conn and a context QObject without success: the lambda is never called.
+    QMetaObject::Connection moConn;
```
nyalldawson, line 226: Could you rename this to remove the initial "m" prefix? Its confusing as it looks like a member variable at first glance.

nyalldawson (approved):

m-kuhn:
Nice!

I guess this can also be used to send back other information / warnings from the db directly to where the command was launched (e.g. db manager) instead of the global message bar / message log.

python/plugins/db_manager/dlg_sql_window.py
elpaso: I won't put up an argument but here are two reasons and a half for leaving as it is:

0.5: after https://github.com/OSGeo/gdal/pull/2953 the cancellation will happen so fast for GPKG/SL that most of the times nobody will notice the button change, for PG the cancellation will already be immediate after this PR
1. there is plenty of horizontal space in that part of the dialog, at the left of the button there is a huge progress bar
2. by changing the size and the phrasing of the button drastically we reduce the likelihood that the user misses the change

but, as I said it's not so important to me.

nyalldawson: Let's let @nirvn make the call?

nirvn: +1 to "Canceling" without the please wait suffix.

m-kuhn: +1 and I think we should keep the ellipsis … as an indicator that something is going on (unless we can add an animated spinner).

nirvn: For the spinner, I suspect the cursor is already set to spinner when we click on [ Execute SQL ]. If it's not, +1 to spinner too.

m-kuhn: Ideally, the cursor shouldn't spin after a click onto "Execute SQL" if it's still possible to click on "Cancel". If any, I would suggest to add spinners and ellipseses directly onto GUI elements to put them into context.

nirvn: True that.

elpaso:
@m-kuhn this PR is the GDAL part of the bugfix, we will eventually be able to use it when released https://github.com/OSGeo/gdal/pull/2953#event-3788638320

m-kuhn:
Thanks @elpaso good job!

PR #38879, opened 2020-09-19
Extend extent history to 100 entries

uclaros:
## Description
Zoom last - Zoom next history is very limited (20 entries) and is easily consumed by mouse wheel zooming.
Maybe raise it to even more than 100?


nyalldawson:
I'd suggest 100 is a good limit - there's little need for more than this as it would be unusable anyway. (Probably we should implement some logic to avoid remembering extents if they are show for < 1s to avoid cluttering the history with unwanted intermediate zoom steps)

PR #38882, opened 2020-09-19
Recalculate layout when legend item size changes

### LINKED ISSUE #38881: AMS legend overlapping existing legend entries
When adding a new AMS layer, the legend is rendered on top of existing layers below in the layer tree.
As soon as the window is switched or the user clicks on the legend, the nodes are moved down and things are corrected.

Peek 2020-09-19 22-39

Steps to reproduce:

1. Add some layers
2. Add an AMS layer on top of the legend (e.g. from https://sampleserver1.arcgisonline.com/arcgis/rest/services)

m-kuhn:
This triggers an `update()` of the legend viewport when the size of the legend item changes.

The solution is only partially satisfying. This should be done automatically by the model / view themselves. And it seems to work correctly for WMS. I couldn't find out why.

This patch works without any side effects. Happy to dump it in favor of an alternative approach.

Fixes #38881

### DIFF
--- a/src/gui/layertree/qgslayertreeview.cpp
+++ b/src/gui/layertree/qgslayertreeview.cpp
@@ -76,28 +76,31 @@ QgsLayerTreeView::~QgsLayerTreeView()
 
 void QgsLayerTreeView::setModel( QAbstractItemModel *model )
 {
-  if ( !qobject_cast<QgsLayerTreeModel *>( model ) )
+  QgsLayerTreeModel *layerTreeModel = qobject_cast<QgsLayerTreeModel *>( model );
+  if ( !layerTreeModel )
     return;
 
   connect( model, &QAbstractItemModel::rowsInserted, this, &QgsLayerTreeView::modelRowsInserted );
   connect( model, &QAbstractItemModel::rowsRemoved, this, &QgsLayerTreeView::modelRowsRemoved );
 
   if ( mMessageBar )
-    connect( layerTreeModel(), &QgsLayerTreeModel::messageEmitted,
+    connect( layerTreeModel, &QgsLayerTreeModel::messageEmitted,
              [ = ]( const QString & message, Qgis::MessageLevel level = Qgis::Info, int duration = 5 )
   {mMessageBar->pushMessage( message, level, duration );}
          );
 
   QTreeView::setModel( model );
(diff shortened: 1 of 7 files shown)


nyalldawson:
Could you test with a project involving 1000s of layers and confirm there's no regressions in project handling speed here? I've previously seen that legend update code like this can cause temporary qgis hangs on large projects (that's an ongoing issue, but I don't want it to get any worse then it currently is!)

m-kuhn:
What is the requirement exactly? Can I duplicate a vector layer 1000 times and add an AMS layer and test with that?

src/gui/layertree/qgslayertreeview.cpp
```diff
@@ -586,3 +589,16 @@ void QgsLayerTreeView::onHorizontalScroll( int value )
   Q_UNUSED( value )
   viewport()->update();
 }
+
+void QgsLayerTreeView::onDataChanged( const QModelIndex &topLeft, const QModelIndex &bottomRight, const QVector<int> &roles )
+{
+  Q_UNUSED( topLeft )
+  Q_UNUSED( bottomRight )
+
+  // If an item is resized asynchroneously (e.g. wms legend)
```
m-kuhn, line 598:
```suggestion
  // If an item is resized asynchronously (e.g. wms legend)
```

nyalldawson:
That'd be a good test. Also throw a vector layer with rule based renderer or categorised renderer and 10000 categories in there.

m-kuhn:
I'm confident this doesn't introduce any noticeable performance impact. It triggers a (deferred) repaint like any user interface interaction does (clicking somewhere, hitting a random button, scrolling, ...) and only does this under a specific condition.

I'll try to create meaningful results. It fear it won't be easy.

_Talking about... Debugging the legend code I've seen code paths triggered repeatedly without any good reason, so if you say you've observed bad performance ... I'm not surprised._

m-kuhn:
Confirmed with 10000 categories legend, 34 seconds to add the layer with and without the patch.

m-kuhn:
Followup issue https://github.com/qgis/QGIS/issues/38890

nyalldawson (approved):
Thanks, proceed at will!

PR #38884, opened 2020-09-19
Add rotation widget to the georeferencer canvas

uclaros:
## Description
You have your maps scanned with arbitrary orientations? No problem!

rotate

### DIFF
--- a/src/app/georeferencer/qgsgeorefmainwindow.cpp
+++ b/src/app/georeferencer/qgsgeorefmainwindow.cpp
@@ -771,6 +771,13 @@ void QgsGeoreferencerMainWindow::extentsChangedQGisCanvas()
   }
 }
 
+void QgsGeoreferencerMainWindow::updateCanvasRotation()
+{
+  double degrees = mRotationEdit->value();
+  mCanvas->setRotation( degrees );
+  mCanvas->refresh();
+}
+
 // Canvas info slots (copy/pasted from QGIS :) )
 void QgsGeoreferencerMainWindow::showMouseCoords( const QgsPointXY &p )
 {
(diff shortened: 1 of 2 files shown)


src/app/georeferencer/qgsgeorefmainwindow.cpp
```diff
@@ -1069,6 +1076,26 @@ QLabel *QgsGeoreferencerMainWindow::createBaseLabelStatus()
 
 void QgsGeoreferencerMainWindow::createStatusBar()
 {
+  // add a widget to show/set current rotation
+  mRotationLabel = createBaseLabelStatus();
+  mRotationLabel->setObjectName( QStringLiteral( "mRotationLabel" ) );
+  mRotationLabel->setText( tr( "Rotation" ) );
+  mRotationLabel->setToolTip( tr( "Current clockwise map rotation in degrees" ) );
```
m-kuhn, line 1083:
```suggestion
  mRotationLabel->setToolTip( tr( "Map rotation in degrees clockwise" ) );
```

I think "current" doesn't add a lot and having "map rotation" as first words helps to understand quickly

m-kuhn (commented):
Nice.

Did you consider to use a QDial? https://doc.qt.io/qt-5/qdial.html

src/app/georeferencer/qgsgeorefmainwindow.cpp
uclaros: I kept the tooltip the same as in the main canvas' rotation widget for translation economy. If I indeed change it, the word 'display' could be added so it is clear that the rotation has no effect on georeferencing!

uclaros:
I didn't. My thought was to be similar to the main canvas rotation widget so that users are familiar with its purpose - mostly the fact that it rotates only the display and does not affect georeferencing

nyalldawson (approved):

PR #38887, opened 2020-09-20
Tune build with Qt 5.14

m-kuhn:
## Description

Just some CI improvements


3nids:
why have you removed the cache? it wasn't working well?

m-kuhn:
Github CI cache is immutable. It's well to cache dependencies (package X in version Y => always serve the same binaries) but cannot be used to preserve a build cache that changes each time.

PR #38891, opened 2020-09-20
Improve legend performance

### LINKED ISSUE #38890: Layer tree performance with many nodes
When the layer tree has many nodes, interacting with it becomes cumbersome and slow.

Steps to reproduce:

 - Create a random point layer with 10000 points
 - Create a categorized renderer on the point id
 - Add a new layer to the legend
 - Wait (34 seconds here)

m-kuhn:
Only recalculate the position of the icons for affected legend nodes.
Avoids brute forcing over complex legends as soon as something is changed.

Fixes #38890


nyalldawson (approved):
Let's give it a shot...

PR #38893, opened 2020-09-20
Allow adding AMS layers even if project has no CRS

### LINKED ISSUE #38892: AMS layers not loaded if project has no CRS
1. Set project CRS to "No Projection"
2. Add an AMS layer
3. Reflect on why you chose to do 2
4. Still nothing shown

m-kuhn:
The uri for AMS layers are separated by spaces ` ` instead of the usual `&`.
No surprise that hacks are not that robust.

Fixes #38892


nyalldawson:
This is super fragile -- it breaks CRS which are defined using wkt strings alone. Oh well, that's not a regression introduced here...!

nyalldawson:
Actually this code and logic in crsAndFormatAdjustedLayerUri is horrible in general...!

m-kuhn:
I agree. Do you want to propose a better API?

PR #38900, opened 2020-09-21
Don't assume iface is available

### LINKED ISSUE #38862: qgis_process list raises exception if processing algorithm provider is activated
Set up:
Ubuntu 20.04 in docker (see Dockerfile below)
QGIS master from https://qgis.org/ubuntu-nightly, September 18th

When running `qgis_process list` , the following error appears:

```
  File "/usr/lib/python3/dist-packages/qgis/utils.py", line 334, in _startPlugin
plugins[packageName] = package.classFactory(iface)
File "/usr/share/qgis/python/plugins/processing/__init__.py", line 50, in classFactory
from processing.ProcessingPlugin import ProcessingPlugin
File "/usr/lib/python3/dist-packages/qgis/utils.py", line 794, in _import
mod = _builtin_import(name, globals, locals, fromlist, level)
File "/usr/share/qgis/python/plugins/processing/ProcessingPlugin.py", line 48, in <module>
  from processing.gui.ConfigDialog import ConfigOptionsPage
File "/usr/lib/python3/dist-packages/qgis/utils.py", line 794, in _import
mod = _builtin_import(name, globals, locals, fromlist, level)
File "/usr/share/qgis/python/plugins/processing/gui/ConfigDialog.py", line 57, in <module>
  from processing.gui.menus import defaultMenuEntries, menusSettingsGroup
File "/usr/lib/python3/dist-packages/qgis/utils.py", line 794, in _import
mod = _builtin_import(name, globals, locals, fromlist, level)
File "/usr/share/qgis/python/plugins/processing/gui/menus.py", line 45, in <module>
  vectorMenu = iface.vectorMenu().title()
AttributeError: 'NoneType' object has no attribute 'vectorMenu'

../../src/core/qgsmessagelog.cpp:29 : (logMessage) [0ms] 2020-09-18T14:54:38 Couldn't load plugin 'processing' due to an error when calling its classFactory() method[1] Couldn't load plugin 'processing' due to an error when calling its classFactory() method
```

To reproduce, use the following docker container:

````
FROM ubuntu:20.04

ENV DEBIAN_FRONTEND noninteractive

RUN apt-get update && \
  apt-get install -y wget gpg tzdata  && \
  echo "deb https://qgis.org/ubuntu-nightly/ focal main" > /etc/apt/sources.list.d/qgis-nightly.list && \
  wget -qO - https://qgis.org/downloads/qgis-2020.gpg.key | gpg --no-default-keyring --keyring gnupg-ring:/etc/apt/trusted.gpg.d/qgis-archive.gpg --import && \
  chmod a+r /etc/apt/trusted.gpg.d/qgis-archive.gpg && \
  apt-get update

RUN apt-get install -y qgis

ENV QT_QPA_PLATFORM offscreen

RUN mkdir -p ~/.local/share/QGIS/QGIS3/profiles/default/QGIS && \
  echo "[PythonPlugins]\nprocessing=true" > ~/.local/share/QGIS/QGIS3/profiles/default/QGIS/QGIS3.ini
````

The error only occurs, when the last line is included. If not, there is no error, but the Python plugins are not listed.

nyalldawson:
E.g. in qgis_process iface is None

Fixes #38862

### DIFF
--- a/python/plugins/processing/gui/menus.py
+++ b/python/plugins/processing/gui/menus.py
@@ -40,105 +40,114 @@ from processing.tools import dataobjects
 
 algorithmsToolbar = None
 menusSettingsGroup = 'Menus'
-
 defaultMenuEntries = {}
-vectorMenu = iface.vectorMenu().title()
-analysisToolsMenu = vectorMenu + "/" + Processing.tr('&Analysis Tools')
-defaultMenuEntries.update({'qgis:distancematrix': analysisToolsMenu,
-                           'native:sumlinelengths': analysisToolsMenu,
-                           'native:countpointsinpolygon': analysisToolsMenu,
-                           'qgis:listuniquevalues': analysisToolsMenu,
-                           'qgis:basicstatisticsforfields': analysisToolsMenu,
-                           'native:nearestneighbouranalysis': analysisToolsMenu,
-                           'native:meancoordinates': analysisToolsMenu,
-                           'native:lineintersections': analysisToolsMenu})
-researchToolsMenu = vectorMenu + "/" + Processing.tr('&Research Tools')
-defaultMenuEntries.update({'native:creategrid': researchToolsMenu,
-                           'qgis:randomselection': researchToolsMenu,
-                           'qgis:randomselectionwithinsubsets': researchToolsMenu,
-                           'native:randompointsinextent': researchToolsMenu,
-                           'qgis:randompointsinlayerbounds': researchToolsMenu,
-                           'native:randompointsinpolygons': researchToolsMenu,
-                           'qgis:randompointsinsidepolygons': researchToolsMenu,
-                           'native:randompointsonlines': researchToolsMenu,
-                           'qgis:regularpoints': researchToolsMenu,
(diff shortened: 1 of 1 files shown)


alexbruy:
Oh, you are faster. I was about to submit fix for this. But your approach looks better than mine.

nyalldawson:
@alexbruy it's a temporary fix -- for 3.18 i'd thinking of pulling apart processing into multiple separate components/plugins:
- qgis python algs
- gdal provider
- saga provider
-grass provider
-otb provider
- processing "core" (the remaining python bits required for processing to run, e.g. model provider + script provider)
- processing "gui" -- all the stuff which requires iface, such as the toolbox, model designer, etc.

The intention then would be that for 4.0 we'd drop processing "core" (it'd all be in c++), and if possible at that stage processing "gui" too.

What do you think?

alexbruy:
I'm fully agree that Processing should become a core component of QGIS. Only providers will remain plugin. So I'm a big +1 here.

m-kuhn:
At the same time, could we update plugin init methods to

- `initGui()` called by app
- `initCore()` called by app, process, server

PR #38908, opened 2020-09-21
Some locator fixes

3nids:
* fix display title expression not displayed anymore in the results
* active locator filter also search in display title
* optimize scale for point layers in QgsMapCanvas::zoomToFeatureIds

### DIFF
--- a/src/app/locator/qgsinbuiltlocatorfilters.cpp
+++ b/src/app/locator/qgsinbuiltlocatorfilters.cpp
@@ -282,7 +282,7 @@ QStringList QgsActiveLayerFeaturesLocatorFilter::prepare( const QString &string,
   QStringList expressionParts;
   QStringList completionList;
   const QgsFields fields = layer->fields();
-  QgsAttributeList subsetOfAttributes;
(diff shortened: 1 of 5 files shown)


nyalldawson:
https://github.com/qgis/QGIS/pull/38908/commits/2ae418da24616b1f1fb809796f17d0d9d599f566 looks like a feature to me!

3nids:
> 2ae418d looks like a feature to me!

That's a copy paste from the all layers filter.
I think it's quite an expected behavior.

But if preferred I can leave it aside.

andreasneumann:
> [2ae418d](https://github.com/qgis/QGIS/commit/2ae418da24616b1f1fb809796f17d0d9d599f566) looks like a feature to me!

Yes, it's an edge case. It is a new feature - but it makes the search much more useful. The display attribute is normally the combination of attribute columns most interesting to the user. I could only test it very close to feature freeze and we couldn't make it before the freeze.

If not acceptable, let's move it to 3.18 then.

nyalldawson:
It makes me slightly nervous, especially given there's no tests covering it (hint hint)

3nids:
test added

3nids:
I completed with test for the display expression part + added test for the completion.

nyalldawson (approved):

PR #38921, opened 2020-09-21
Don't delete QgsSymbolSelectorWidget symbol when widget is closed

### LINKED ISSUE #38920: Crash when editing style and opening a project
**Describe the bug**

In Dock mode, when editing symbol from categorized renderer, if the user opens a new project, QGIS crashes

I already have a fix for the issue, PR is coming

**How to Reproduce**

1. From *Properties > Style > Categorized renderer* click on *Symbol* button
2. Open another project
3. QGIS crashes

**QGIS and OS versions**

QGIS master
Debian bullseye

**Additional context**

troopa81:
## Description

Fixes #38920 :  Don't delete the symbol without deleting the widget. The widget is still connected to some signals (see constructor) and try to access the deleting symbol in slot method. The widget and the symbol will be deleted by QgsPanelWidgetStack::closePanel (panel has autoDelete flag)

### DIFF
--- a/src/gui/qgssymbolbutton.cpp
+++ b/src/gui/qgssymbolbutton.cpp
@@ -115,7 +115,6 @@ void QgsSymbolButton::showSettingsDialog()
     d->setPanelTitle( mDialogTitle );
     d->setContext( symbolContext );
     connect( d, &QgsPanelWidget::widgetChanged, this, &QgsSymbolButton::updateSymbolFromWidget );
-    connect( d, &QgsPanelWidget::panelAccepted, this, &QgsSymbolButton::cleanUpSymbolSelector );
     panel->openPanel( d );
   }
   else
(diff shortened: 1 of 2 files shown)


m-kuhn (approved):
Ownership documentation of the functions as well as unique_ptr-wise reflect the suggestions of this PR.
I wonder how this went by unnoticed for so long.

nyalldawson:
I don't think this is correct -- from the documentation for QgsSymbolSelectorWidget:

    "\note The ownership of the symbol is not transferred and must exist for the lifetime of the widget."

In QgsSymbolButton::showSettingsDialog() we create a new symbol via

    QgsSymbol *newSymbol = mSymbol->clone();

and then create the selector widget using this symbol (ownership is NOT transferred):

    QgsSymbolSelectorWidget *d = new QgsSymbolSelectorWidget( newSymbol, QgsStyle::defaultStyle(), mLayer, nullptr );

so the widget's symbol is STILL owned by the QgsSymbolButton. With this PR there's nothing cleaning up the symbol, so there's a leak.

I admit this is all disgusting memory management, but I can't see any alternative which doesn't break the (bad) api for QgsSymbolSelectorWidget which has been around since 2.x. QgsSymbolSelectorWidget should really take a clone of the symbol and drop the requirement that the symbol pointer exists for the lifetime of the widget... (TODO for 4.0, there's already a note in the code to this effect)

nyalldawson:
(also note that this same code logic exists in many places, e.g. QgsCategorizedSymbolRendererWidget,...)

nyalldawson:
Alternative fix https://github.com/qgis/QGIS/pull/38928

m-kuhn:
Thanks for looking closer @nyalldawson I was misled by the local std::unique_ptr and missed the case that in the dock widget scenario the memory management is ... counter intuitive.

m-kuhn (changes requested):
Leaks

PR #38929, opened 2020-09-21
Dump backtraces on uncaught exceptions

nyalldawson:
Instead of just logging that we encountered the exception, give more useful debug info by dumping the backtrace


m-kuhn (approved):
YES!

PR #38944, opened 2020-09-22
Browser fields enhancements

elpaso:
- Check field `subType` for a fallback icon for fields where type is not known (for instance "name"  for PG 12 table/schema identifiers
- Fix unreported issue with `fields()` not retrieved for certain tables due to multiple geometry column types/srid

### DIFF
--- a/src/providers/postgres/qgspostgresproviderconnection.cpp
+++ b/src/providers/postgres/qgspostgresproviderconnection.cpp
@@ -671,32 +671,32 @@ QgsFields QgsPostgresProviderConnection::fields( const QString &schema, const QS
     // This table might expose multiple geometry columns (different geom type or SRID)
     // but we are only interested in fields here, so let's pick the first one.
     TableProperty tableInfo { table( schema, tableName ) };
-    if ( tableInfo.geometryColumnTypes().count( ) > 1 )
+    try
     {
-      try
+      QgsDataSourceUri tUri { tableUri( schema, tableName ) };
+      if ( tableInfo.geometryColumnTypes().count( ) > 1 )
       {
-        QgsDataSourceUri tUri { tableUri( schema, tableName ) };
         TableProperty::GeometryColumnType geomCol { tableInfo.geometryColumnTypes().first() };
         tUri.setGeometryColumn( tableInfo.geometryColumn() );
         tUri.setWkbType( geomCol.wkbType );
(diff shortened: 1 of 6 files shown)


tests/src/python/test_qgsproviderconnection_postgres.py
```diff
@@ -329,6 +329,30 @@ def test_fields(self):
         fields = conn.fields('qgis_test', 'someData')
         self.assertEqual(fields.names(), ['pk', 'cnt', 'name', 'name2', 'num_char', 'dt', 'date', 'time', 'geom'])
 
+   def test_fields_no_pk(self):
```
elpaso, line 332:
```suggestion
    def test_fields_no_pk(self):
```

src/providers/postgres/qgspostgresproviderconnection.cpp
```diff
@@ -656,3 +656,48 @@ QList<QgsVectorDataProvider::NativeType> QgsPostgresProviderConnection::nativeTy
...
+        TableProperty::GeometryColumnType geomCol { tableInfo.geometryColumnTypes().first() };
+        tUri.setGeometryColumn( tableInfo.geometryColumn() );
+        tUri.setWkbType( geomCol.wkbType );
+        tUri.setSrid( QString::number( geomCol.crs.postgisSrid() ) );
+        if ( tableInfo.primaryKeyColumns().count() > 0 )
+        {
+          tUri.setKeyColumn( tableInfo.primaryKeyColumns().first() );
+        }
+        tUri.setParam( QStringLiteral( "checkPrimaryKeyUnicity" ), QLatin1String( "0" ) );
+        QgsVectorLayer::LayerOptions options { true, true };
```
nyalldawson, line 688:
```suggestion
        QgsVectorLayer::LayerOptions options { false, false };
```

elpaso: Why `readExtentFromXml`? I'm only interested in the fields here, we don't need any extent or style information, I thought that by constructing the options with `readExtentFromXml`=`false` we can speed things up, you are right about `loadDefaultStyle` though, I think it skips some loading operations.

nyalldawson: Ok good point -- readExtentFromXml SHOULD be true!

nyalldawson (approved):

PR #38947, opened 2020-09-22
[locator] active filter lists all matching fields as results when typing '@'

3nids:
results can be triggered to restrict the search to the given field

image

### DIFF
--- a/src/app/locator/qgsinbuiltlocatorfilters.cpp
+++ b/src/app/locator/qgsinbuiltlocatorfilters.cpp
@@ -328,7 +328,7 @@ QStringList QgsActiveLayerFeaturesLocatorFilter::prepare( const QString &string,
       mFieldsCompletion << field.name();
     }
 
-    // the completion list (returned by the current method) is used by the locator line edit directy
+    // the completion list (returned by the current method) is used by the locator line edit directly
     completionList.append( QStringLiteral( "@%1 " ).arg( field.name() ) );
 
     if ( field.type() == QVariant::String )
--- a/src/app/locator/qgsinbuiltlocatorfilters.cpp
+++ b/src/app/locator/qgsinbuiltlocatorfilters.cpp
@@ -262,7 +262,7 @@ QStringList QgsActiveLayerFeaturesLocatorFilter::prepare( const QString &string,
   mFieldsCompletion.clear();
 
   // Normally skip very short search strings, unless when specifically searching using this filter or try to match fields
-  if ( string.length() < 3 && !context.usingPrefix && !string.startsWith( QStringLiteral( "@" ) ) )
+  if ( string.length() < 3 && !context.usingPrefix && !string.startsWith( '@' ) ) )
     return QStringList();
 
   QgsSettings settings;
@@ -1133,4 +1133,3 @@ void QgsGotoLocatorFilter::triggerResult( const QgsLocatorResult &result )
 
   mapCanvas->flashGeometries( QList< QgsGeometry >() << QgsGeometry::fromPointXY( point ) );
 }
-
--- a/src/app/locator/qgsinbuiltlocatorfilters.cpp
+++ b/src/app/locator/qgsinbuiltlocatorfilters.cpp
@@ -38,6 +38,7 @@
 #include "qgsvectorlayerfeatureiterator.h"
 #include "qgsexpressioncontextutils.h"
 #include "qgssettings.h"
+#include "qgslocatorwidget.h"
 
 
 QgsLayerTreeLocatorFilter::QgsLayerTreeLocatorFilter( QObject *parent )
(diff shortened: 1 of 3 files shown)


tests/src/app/testqgsapplocatorfilters.cpp
```diff
@@ -220,6 +220,16 @@ void TestQgsAppLocatorFilters::testActiveLayerFieldRestriction()
   restr = QgsActiveLayerFeaturesLocatorFilter::fieldRestriction( search );
   QCOMPARE( restr, QStringLiteral( "home" ) );
   QCOMPARE( search, QStringLiteral( "" ) );
+
+  search = QStringLiteral( "@" );
+  restr = QgsActiveLayerFeaturesLocatorFilter::fieldRestriction( search );
+  QVERIFY( !restr.isNull() );
+  QCOMPARE( search, QStringLiteral( "" ) );
```
nyalldawson, line 227:
```suggestion
  QCOMPARE( search, QString() );
```

src/app/locator/qgsinbuiltlocatorfilters.cpp
```diff
@@ -361,6 +372,17 @@ void QgsActiveLayerFeaturesLocatorFilter::fetchResults( const QString &string, c
   QString searchString = string;
   fieldRestriction( searchString );
 
+  // propose available fields for restriction
+  for ( const QString &field : qgis::as_const( mFieldsCompletion ) )
+  {
+    QgsLocatorResult result;
+    result.displayString = QStringLiteral( "@%1" ).arg( field );
+    result.description = QStringLiteral( "Limit the search to the field '%1'" ).arg( field );
```
nyalldawson, line 380: tr

src/app/locator/qgsinbuiltlocatorfilters.cpp
```diff
@@ -258,8 +259,10 @@ QString QgsActiveLayerFeaturesLocatorFilter::fieldRestriction( QString &searchSt
 
 QStringList QgsActiveLayerFeaturesLocatorFilter::prepare( const QString &string, const QgsLocatorContext &context )
 {
-  // Normally skip very short search strings, unless when specifically searching using this filter
-  if ( string.length() < 3 && !context.usingPrefix )
+  mFieldsCompletion.clear();
+
+  // Normally skip very short search strings, unless when specifically searching using this filter or try to match fields
+  if ( string.length() < 3 && !context.usingPrefix && !string.startsWith( QStringLiteral( "@" ) ) )
```
nyalldawson, line 265:
```suggestion
  if ( string.length() < 3 && !context.usingPrefix && !string.startsWith( '@' ) ) )
```

src/app/locator/qgsinbuiltlocatorfilters.cpp
```diff
@@ -308,16 +311,24 @@ QStringList QgsActiveLayerFeaturesLocatorFilter::prepare( const QString &string,
     if ( field.configurationFlags().testFlag( QgsField::ConfigurationFlag::NotSearchable ) )
       continue;
 
-    if ( !_fieldRestriction.isEmpty() && !field.name().startsWith( _fieldRestriction ) )
+    if ( !_fieldRestriction.isNull() && !field.name().startsWith( _fieldRestriction ) )
```
nyalldawson, line 314: why this change? it's a step back, in that the concept of null QStrings is deprecated...

src/app/locator/qgsinbuiltlocatorfilters.h
```diff
@@ -123,11 +141,15 @@ class APP_EXPORT QgsActiveLayerFeaturesLocatorFilter : public QgsLocatorFilter
     QString mLayerId;
     QIcon mLayerIcon;
     QStringList mAttributeAliases;
+    QStringList mFieldsCompletion;
     int mMaxTotalResults = 30;
 
     friend class TestQgsAppLocatorFilters;
 };
 
+Q_DECLARE_METATYPE( QgsActiveLayerFeaturesLocatorFilter::Data )
```
nyalldawson, line 150: Would be cleaner to avoid the Data struct and instead use a QVariantMap value.

src/app/locator/qgsinbuiltlocatorfilters.cpp
3nids: I wanted to distinguish when '@' was typed with nothing yet. But I can replace it with a `startsWith('@')`.

nyalldawson: Why doesn't testing .isEmpty() work for the same result?

3nids: My first idea was:
search: 'something' => fieldRestriction is null
search: '@field something' => fieldRestriction is field
search: '@' => fieldRestriction is empty string, not null

I added now a pointer to a boolean in the private method, things are a bit more explicit like this.

src/app/locator/qgsinbuiltlocatorfilters.h
3nids: done

PR #38948, opened 2020-09-22
Some tweaks to 3D map view Configuration dialog

DelazJ:
Fixing labels and margins, adding units suffix to spinboxes (btw if someone can confirm that those are right) and ui cleanup

Before
image

After
image

I wonder whether to remove the "light direction" label or offset the X, Y and Z labels. This is what we do eg in labels formatting multiline properties, or somehow in layout map item grid interval property, but it's not the case in the tab next door, aka "point lights" (for attenuation)

And if someone knows how we can get rid of this blank space at the bottom of the properties...

### DIFF
--- a/src/app/3d/qgsshadowrenderingsettingswidget.cpp
+++ b/src/app/3d/qgsshadowrenderingsettingswidget.cpp
@@ -24,6 +24,10 @@ QgsShadowRenderingSettingsWidget::QgsShadowRenderingSettingsWidget( QWidget *par
   : QWidget( parent )
 {
   setupUi( this );
+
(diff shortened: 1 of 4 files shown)


nyalldawson:
While you're doing this -- could you update the various shadow setting spin boxes with clear values which reset them to the default setting?

DelazJ:
> While you're doing this -- could you update the various shadow setting spin boxes with clear values which reset them to the default setting?

Done, @nyalldawson. Do we have rules re when we should use qgsspinbox (and this clear method) over qspinbox?

nyalldawson:
We should ALWAYS use QgsSpinBox/QgsDoubleSpinBox.

nyalldawson:
Thanks!

PR #38949, opened 2020-09-22
[VirtualLayer] Catch exception while updating virtual layer stats

### LINKED ISSUE #34378: Error in SQL statement for Virtual layers may lock layers in map canvas
### Issue:
Sometimes when I do something "wrong" in a virtual layer, the test can be "No error", but the result after "Add" is an error and any visible layers in the canvas will be "locked". No matter what layers are turned on or off, added or deleted, the canvas stays the same. QGIS does not crash, but will need a restart to work.

### Repeat:
Add "world" layer to canvas (world in coordinate field), rename to "world".
Add "contributors" layer to canvas (contributors in coordinate field), rename "contributors"

Create virtual layer:

`select * from contributors where st_intersects(contributors.geometry, (select geometry from world where iso_a2='FR'))`

This should generate a layer with points only inside the France country polygon.

Now, edit the virtual layer expression to something that is wrong, like:

`select * from contributors where st_intersects(contributors.geometry, (select * from world where iso_a2='FR'))`

(Note the * where previously geometry was)

A "Test" will not throw any errors, but when updating the layer, there will be an error...

**After this the canvas is locked!**

The only solution I've found to "un-freeze" the canvas is to restart QGIS.

### System

QGIS version | 3.10.2-A Coruña | QGIS code revision | 616ad4531b
-- | -- | -- | --
PROJ.4 Version | 493
OS Version | Ubuntu 18.04.4 LTS
Active python plugins | qfieldsync; LmOpenData; qgis2web; layout_loader; plugin_reloader; DataPlotly; Qgis2threejs; latlontools; pluginbuilder3; MetaSearch; db_manager; processing

troopa81:
## Description

Fixes #34378 : Uncaught exception was breaking QGIS

### DIFF
--- a/tests/src/python/test_provider_virtual.py
+++ b/tests/src/python/test_provider_virtual.py
@@ -1271,6 +1271,27 @@ class TestQgsVirtualLayerProvider(unittest.TestCase, ProviderTestCase):
         self.assertEqual(gpkg_virtual_layer.featureCount(), 1)
         self.assertEqual(gpkg_virtual_layer.subsetString(), '"join_value" = \'twenty\'')
 
+    def test_feature_count_on_error(self):
+
+        # Test that triggered exception while getting feature count on a badly defined
+        # virtual layer is correctly caught (see https://github.com/qgis/QGIS/issues/34378)
+
+        l1 = QgsVectorLayer(os.path.join(self.testDataDir, "france_parts.shp"), "france", "ogr",
+                            QgsVectorLayer.LayerOptions(False))
+        self.assertEqual(l1.isValid(), True)
+        QgsProject.instance().addMapLayer(l1)
+
+        df = QgsVirtualLayerDefinition()
+        df.setQuery('select error')
+
+        vl = QgsVectorLayer(df.toString(), "testq", "virtual")
+        self.assertEqual(vl.isValid(), False)
+        self.assertEqual(vl.featureCount(), 0)
+        ids = [f.id() for f in vl.getFeatures()]
+        self.assertEqual(len(ids), 0)
+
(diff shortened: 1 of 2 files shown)


tests/src/python/test_provider_virtual.py
```diff
@@ -1271,6 +1271,27 @@ def test_subset_string(self):
         self.assertEqual(gpkg_virtual_layer.featureCount(), 1)
         self.assertEqual(gpkg_virtual_layer.subsetString(), '"join_value" = \'twenty\'')
 
+    def test_feature_count_on_error(self):
+
```
elpaso, line 1275:
```suggestion
        """Test that triggered exception while getting feature count on a badly defined
        virtual layer is correctly caught (see https://github.com/qgis/QGIS/issues/34378)"""
```

python function comments are supposed to be placed right after the function definition, you can remove the following two lines.

elpaso:
@troopa81 LGTM.

Not related at all but I was wondering if there a reason why this class is using its own `QSqlite` interface instead of the `QgsSqliteUtils` ?

troopa81:
> Not related at all but I was wondering if there a reason why this class is using its own QSqlite interface instead of the QgsSqliteUtils ?

`QgsSqliteUtils` (added in 2017) didn't not exist when Virtual layer feature was added (in 2015). I don't see any other reason for virtual layer to not use the same interface.

troopa81:
Thanks for review @elpaso

nyalldawson:
Can you rebase and fix the merge conflict please?

troopa81:
> Can you rebase and fix the merge conflict please?

Done

PR #38951, opened 2020-09-22
Add Layer crs authid expression

roya0045:
## Description

I feel like this is long overdue.

This is especially useful when doing transformation to compare features or simply check if the features are within the map extent without having to manually change the projection in the expression each time.


roya0045:
@nyalldawson is this even a feature? Just exposing new variable? If this is a feature I might add some more vars as I'm at it since this won't be merged soon. I'll see what I can do if I have some time.

roya0045:
bad stalebot

nyalldawson (approved):

roya0045:
YAY Thanks a ton Nyall for reviewing PRs!

PR #38952, opened 2020-09-22
[WFS/OAPIF providers] Fix freeze when feature requests issued from main thread and authentication manager triggers (fixes #37224)

### LINKED ISSUE #37224: Freeze when showing attribute table of WFS layer
How to reproduce: open a project with WFS layer, try to open attribute table

notice: in this scenario, the credentials are not saved in the project or in the auth manager, but entered at opening of the project (I did not test yet in a different manner)

Here are the different stacks

image

image

image

I tried to add a processEvents in `QgsBackgroundCachedFeatureIterator::fetchFeature`, but this brings the freeze to another place (which after a few tries seems to alternate with the first stack above):

image

rouault:


nyalldawson:
@rouault do these changes need to be copied to QgsBlockingNetworkRequest too?

rouault:
> do these changes need to be copied to QgsBlockingNetworkRequest too?

that's actually the other way round. I've taken inspiration from QgsBlockingNetworkRequest for the strategy to adopt regarding signals and threading

nyalldawson:
Ok, great. Thanks for the confirmation!

PR #38956, opened 2020-09-22
[Backport 3.10] [WFS provider] Fix freeze when feature requests issued from main thread and authentication manager triggers (fixes #37224)

### LINKED ISSUE #37224: Freeze when showing attribute table of WFS layer

rouault:
The first commit is a cherry-pick of 2eeceb5fa81b5f6133e7a883859f609f96d3543f that has been in master since 6 months, and was needed to avoid a crash when opening the reproducer project of #37224 due to a invalid layer

The actual fix of the WFS provider is the second commit, which is manual backport of PR #38952

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -13252,10 +13252,13 @@ void QgisApp::activateDeactivateLayerRelatedActions( QgsMapLayer *layer )
     case QgsMapLayerType::RasterLayer:
     {
       const QgsRasterLayer *rlayer = qobject_cast<const QgsRasterLayer *>( layer );
-      if ( rlayer->dataProvider()->dataType( 1 ) != Qgis::ARGB32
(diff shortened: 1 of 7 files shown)


nyalldawson:
I'd like to delay this one by a cycle - even though it's a hang fix, it's moderately high risk

nyalldawson:
Can you back out the changes to the world map gpkg?

rouault:
> Can you back out the changes to the world map gpkg?

Done

nyalldawson (approved):

nyalldawson:

PR #38964, opened 2020-09-22
PGSQL: making timestamptz a recognized date/time type.

### LINKED ISSUE #38962: PostgreSQL temporal layer timestamptz type not recognized as date/time
When enabling the temporal feature for a PostgreSQL layer which contains a column of type "timestamp with timezone" (timestamptz in PostgreSQL parlance), that column isn't recognized as a temporal column. For instance, a table defined thus:

```sql
CREATE TABLE test_temporal
(
        id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
        dt_event TIMESTAMP WITH TIME ZONE,
        geom Geometry(Point, 4674)
);
```

The field dt_event doesn't appear as a field with a date/time value:
image

Instead, when fiddling with the other options, that field seems to be recognized as of type string:

image

If the field is of type "TIMESTAMP WITHOUT TIME ZONE" (the PostgreSQL default), then it is correctly recognized as such, and can be used for temporal goodness. I'm preparing a PR to address this shortly, as the fix is trivial.

QGIS version | 3.15.0-Master | QGIS code revision | 1c1febc0d6
-- | -- | -- | --
Compiled against GDAL/OGR | 3.1.2 | Running against GDAL/OGR | 3.1.2
Compiled against SQLite | 3.32.3 | Running against SQLite | 3.32.3
PostgreSQL Client Version | 12.4 | SpatiaLite Version | 4.3.0a
QWT Version | 6.1.4 | QScintilla2 Version | 2.11.5
Compiled against PROJ | 7.1.0 | Running against PROJ | Rel. 7.1.1, September 1st, 2020
OS Version | Gentoo/Linux
Active python plugins | QuickWKT; MetaSearch; db_manager; processing

espinafre:
This trivial change makes the PostgreSQL type "TIMESTAMP WITH TIME ZONE", or "timestamptz" internally recognized as a date/time field, which can be used for the temporal capabilities. I believe this can be safely backported.

Fixes #38962.

### DIFF
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -1054,7 +1054,7 @@ bool QgsPostgresProvider::loadFields()
         fieldType = QVariant::Time;
         fieldSize = -1;
       }
-      else if ( fieldTypeName == QLatin1String( "timestamp" ) )
+      else if ( fieldTypeName == QLatin1String( "timestamp" ) || fieldTypeName == QLatin1String( "timestamptz" ) )
       {
         fieldType = QVariant::DateTime;
         fieldSize = -1;


nyalldawson:
Fix looks good, but be aware that QGIS has somewhat broken handling of timezones for datetime values, so the results may not behave exactly as you'd expect!

PR #38965, opened 2020-09-23
Fix numerous bugs and UI issues with CRS selection

nyalldawson:
Fixes a bunch of bugs in the new CRS options from the layer context menu, and then goes on to improve the UI regarding CRS handling and resolves many paper cut issues with those...


nirvn:
Sweeeeet stuff, ideal timing too, just in time for our LTR users to enjoy this.

Btw, there's another chunk of projection UI that needs improvement: ATM, the project properties tab is way too crowded as it vertically piles up projection & transformation widgets. IMHO, we need to split those into tabs. It'll also have the collateral benefit of stressing importance of transformations. And opens the door in the future for more UI additions.

nyalldawson:
@nirvn done, but needs a new icon for the new Transformations tab!

nirvn:
Here:
[transformation.zip](https://github.com/qgis/QGIS/files/5265070/transformation.zip)

nirvn (approved):

i-s-o:
Would you be willing to reconsider 3824b5b?

Here are a few of situations this action might be useful:

1. Working with spatial data from multiple sources (e.g. state and federal agencies in the USA):
Sources providing nationwide data will usually have a CRS suitable for a very large area. Then, each state and/or local agency (e.g. New York City) will use a different projection (usually a State Plane). For a given project, a user might use data from multiple states and overlay them on a nationwide layer. At different stages of the project, one might want to focus on a specific state, then another one, then the nation as a whole, hence the usefulness of setting the project CRS from a specific layer.

2. Using a web-based raster layer(s):
At different stages of a project, one might want to set the project's projection to the base layer to get the best rendering of the original raster, later on switching to a more appropriate projection for further analysis and/or display. Then, switching back and forth for further review.

3. This action is very useful while teaching the concept of projections.

nyalldawson:
@i-s-o I believe you've misinterpreted that commit. It only hides the menu "set project crs from layer" item when the user has selected multiple layers each with a different crs. Because in that case, what's the correct behaviour? Pick a random layer and set the project crs from that? 😁

i-s-o:
That makes perfect sense. My brain must have ignored the word "selected" in your commit message because I would never think of that option if I had multiple layers selected :smile:. Sorry for the noise.

nyalldawson:
And actually, you've made me realise there's a remaining bug because we still show the first layer's crs at the top of the menu. This should be replaced by a " mixed crs" message

PR #38968, opened 2020-09-23
fix edge case where fkey is 0 in relation reference widget 

3nids:
if a not null constraint was set, the form could not be validated.

### DIFF
--- a/src/core/qgsfeaturefiltermodel.cpp
+++ b/src/core/qgsfeaturefiltermodel.cpp
@@ -96,7 +96,7 @@ bool QgsFeatureFilterModel::identifierIsNull( const QVariant &identifier ) const
   const QVariantList values = identifier.toList();
   for ( const QVariant &value : values )
   {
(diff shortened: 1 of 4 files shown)


src/native/mac/qgsmacnative.mm
```diff
@@ -60,7 +61,10 @@ - ( BOOL )userNotificationCenter:( NSUserNotificationCenter * )center shouldPres
 
 void QgsMacNative::setIconPath( const QString &iconPath )
 {
-  mQgsUserNotificationCenter->_qgisIcon = QtMac::toNSImage( QPixmap( iconPath ) );
+  QImage image = QPixmap( iconPath ).toImage();
```
nyalldawson, line 64: unrelated change

3nids: thanks

nyalldawson (approved):

PR #38969, opened 2020-09-23
Treat GeoPDF files as databases in browser panel

nyalldawson:
Since these can have multiple sub layers, we need to allow users
to expand them out to pick a specific layer or it is not possible
to repair broken GeoPDF layer sources in projects.


nyalldawson:
Unrelated test failure

PR #38981, opened 2020-09-23
Bugfix gh38975 and other db manager issues

### LINKED ISSUE #38975: `show search_path` returns 0 rows
In DB Manager the above command returns the correct value in 3.10, but 0 rows in 3.14

### LINKED ISSUE #38976: Repeated column names cause visualization issues in DB Manager
In QGIS 3.10 the following query returns correct results:

```
SELECT a.id, b.id, ST_Distance(a.geom, b.geom)
  FROM punti a, fiumi b
```
whereas in 3.14 distance is not displayed.
```
SELECT ST_Distance(a.geom, b.geom)
  FROM punti a, fiumi b
```
returns correct results in both.

elpaso:
Fixes #38976
Fixes #38975

### DIFF
--- a/python/plugins/db_manager/db_plugins/postgis/connector.py
+++ b/python/plugins/db_manager/db_plugins/postgis/connector.py
@@ -29,7 +29,6 @@ from functools import cmp_to_key
 from qgis.PyQt.QtCore import (
     QRegExp,
     QFile,
-    QCoreApplication,
     QVariant,
     QDateTime,
     QTime,
@@ -39,7 +38,6 @@ from qgis.PyQt.QtCore import (
 from qgis.core import (
     Qgis,
     QgsCoordinateReferenceSystem,
-    QgsCredentials,
     QgsVectorLayer,
     QgsDataSourceUri,
     QgsProviderRegistry,
@@ -48,9 +46,10 @@ from qgis.core import (
 )
 
 from ..connector import DBConnector
-from ..plugin import ConnectionError, DbError, Table
+from ..plugin import DbError, Table
 
 import os
+import re
 import psycopg2
 import psycopg2.extensions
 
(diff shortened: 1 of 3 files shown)


python/plugins/db_manager/db_plugins/postgis/connector.py
```diff
@@ -118,29 +117,49 @@ def description(self):
...
+            self._description = []
 
-            # TODO: make this part provider-agnostic
-            sql = self.sql if self.sql.upper().find(' LIMIT ') >= 0 else self.sql + ' LIMIT 1 '
-            uri.setTable('(SELECT row_number() OVER () AS __rid__, * FROM (' + sql + ') as foo)')
-            uri.setKeyColumn('__rid__')
-            # TODO: fetch provider name from connection (QgsAbstractConnectionProvider)
-            # TODO: re-use the VectorLayer for fetching rows in batch mode
-            vl = QgsVectorLayer(uri.uri(False), 'dbmanager_cursor', 'postgres')
+            if re.match('^SHOW', self.sql.strip().upper()):
```
nyalldawson, line 122: Why's this a special case?

elpaso: Because we cannot create a `QgsVector(query)Layer` from that: it's not a SQL expression but an internal command, this will fail:
```sql
(SELECT row_number() OVER () AS __rid__, * FROM (SHOW search_path ) as foo);
```

nyalldawson: Ok, makes sense. I'm just conscious that this dbmanager specific code has a limited lifetime left, so this statement will break again sometime in the future...

elpaso: I was actually dreaming about a C++ QgsSqlQueryWidget ...

pcav: > Ok, makes sense. I'm just conscious that this dbmanager specific code has a limited lifetime left, so this statement will break again sometime in the future...

is there a plan to replace DB Manager functionality with something better?

elpaso: Yes: the plan is to integrate DB-manager functionalities into core until we can RIP it forever. 

We already have a growing number of functionalities ported but we aren't quite there yet.

pcav: [OT here:] Is there a list of the functionalities that will be migrated? Is it worth a QEP?

PR #38989, opened 2020-09-24
Make processing's batch process faster

### LINKED ISSUE #38987: Processing's batch process performs badly when adding a large number of files
When auto-filling the batch table with a large number of files, QGIS will freeze for a *long* time (e.g. on one machine, 3000 files = +/-20 minutes :scream: ).

Here's a hotspot graph output that indicates a couple of bottlenecks:
image

nirvn:
## Description

This and @nyalldawson 's PR #38988 makes processing's batch process usable when adding a large number of rows (from files, expressions, etc.) and therefore fixes #38987 . 

On the long run, we probably want to avoid creating widgets for all rows all all time and instead only show widgets when editing cell values. But for now, the two drastic speed boost unlocked does the job.

### DIFF
--- a/python/plugins/processing/gui/BatchPanel.py
+++ b/python/plugins/processing/gui/BatchPanel.py
@@ -391,9 +391,13 @@ class BatchPanelFillWidget(QToolButton):
                 res = [res]
 
             first_row = self.panel.batchRowCount() if self.panel.batchRowCount() > 1 else 0
+            self.panel.addRow(len(res))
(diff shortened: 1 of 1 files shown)


nyalldawson (approved):

FelisPimeja:
Wow! @nirvn, @nyalldawson - thank you so much, guys!
This was really annoying - I even had to split long batch queues into parts to make it work comfortable.

PR #38997, opened 2020-09-24
Add a semi-automated table of content for the Install file

DelazJ:
Because it's no always easy to find the section you are interested in, let's generate a more verbose table of contents using https://freelance-tech-writer.github.io/table-of-contents-generator/index.html

This PR adds the whole structure, but maybe should we limit to 2, 3 levels?


nyalldawson:
Good move!

PR #39005, opened 2020-09-24
remove Volumes (MacOS) from Browser

### LINKED ISSUE #29702: Browser cannot open Volumes under macOS
Author Name: **Chang Liao** (Chang Liao)
Original Redmine Issue: [21887](https://issues.qgis.org/issues/21887)
Affected QGIS version: 3.6.1
Redmine category:browser

---
I tried both QGIS 3.4 and 3.6, but none of them is working probably.
It cannot open the /Volumes directory on my Mac even though there are external driver connected.

P.S. I was able to open it ONLY in QGIS 3.2.

PeterPetrik:
fix https://github.com/qgis/QGIS/issues/29702

The Volumes is not displayed in any of the Applications I use, it is not intended for the regular users


nyalldawson:
How do you open USB drives then? (Not a Mac user, this might be a dumb question)

PeterPetrik:
i do not have USB connector on Mac anymore, need to find a USB key and dongle to try it out (?). It would be possible to get to the place probably through "/" item, in case Mac wants to allow user to access it

PeterPetrik:
you are right, closing this , it is wrong fix... see https://github.com/qgis/QGIS/pull/39096

PR #38101, opened 2020-08-03
Typo, URL and formatting in the build instructions

DelazJ:
The provided URL for the install instructions file was wrong (unless it's broken?)
Add some formatting to text to ease readability
fix some typo and http URL

PR #39008, opened 2020-09-24
Remove "creation of MSYS environment" instructions

DelazJ:
as discussed in #38101
Not clear to me whether the [section 4.2](https://github.com/qgis/QGIS/blob/master/INSTALL.md#42-building-using-mingw) should also be removed


nyalldawson:
Not sure about section 4.2 -- @m-kuhn @jef-n  ?

jef-n:
4.2 can go too.

PR #39009, opened 2020-09-24
[processing] fix crash in graphical modeler due to QList index invalidation

luipir:
## Description

Fixed crash in Processing Graphical Modeler due to index invalidation

### DIFF
--- a/src/gui/processing/models/qgsmodelcomponentgraphicitem.cpp
+++ b/src/gui/processing/models/qgsmodelcomponentgraphicitem.cpp
@@ -1006,6 +1006,9 @@ QString QgsModelChildAlgorithmGraphicItem::linkPointText( Qt::Edge edge, int ind
           return param->flags() & QgsProcessingParameterDefinition::FlagHidden || param->isDestination();
         } ), params.end() );
 
+        // check if index remain valid after erasing of the previous step. Otherwise qgis crash calling params.at( index )
+        if ( index >= params.size() )
+          return QString();
 
         QString title = params.at( index )->description();
         if ( !mInputs.value( params.at( index )->name() ).toString().isEmpty() )


luipir:
No issue associated to this bug

nyalldawson:
Have you got more information about how this issue is triggered? The same filtering code is used in QgsModelChildAlgorithmGraphicItem::linkPointCount, so I'm concerned that there's a deeper underlying bug at play here...

luipir:
> Have you got more information about how this issue is triggered? The same filtering code is used in QgsModelChildAlgorithmGraphicItem::linkPointCount, so I'm concerned that there's a deeper underlying bug at play here...

I've a complex model (with submodels) that trigger this... and I'm wating to have permission to attach the model to allow replication :( but client is really slow sorry. It is a model opening correctly in 3.10 but crashing opening after 3.12 and I had time to debug only now when client decided to move all to the next LTR.

If they continue to be late are you interested to have the model shared privately? I do not like to notify issue in this way sorry...

nyalldawson:
> If they continue to be late, are you interested to have the model shared privately? I do not like to notify issue in this way sorry...

Yes please!

luipir:
[epyris.zip](https://github.com/qgis/QGIS/files/5290737/epyris.zip)

luipir:
teh procedure is:
1) install plugin via zip and activate it (hopefully shoud work ;) ). All models and scripts should be installed into processing
2) try to open graphical modeler: Models->EPyRIS Severity Map->Severity map and perimeter generation
3) would crash at least with this configuration:

QGIS version | 3.15.0-Master | QGIS code revision | **e7e08fd504 !!!!! <- should be master, in my case I user this PR patch to avoid crash !!!!!**
-- | -- | -- | --
PostgreSQL Client Version | 12.4 (Ubuntu 12.4-1.pgdg18.04+1) | SpatiaLite Version | 4.3.0a
PROJ.4 Version | 493
OS Version | Ubuntu 18.04.5 LTS | This copy of QGIS writes debugging output.
Active python plugins | plugin_reloader; IPyConsole; pluginbuilder3; copernicushub; firstaid; epyris; db_manager; processing; MetaSearch

nyalldawson:
@luipir 

https://github.com/qgis/QGIS/pull/39200 should fix the underlying issue

luipir:
> #39200 should fix the underlying issue

great tnx

PR #39011, opened 2020-09-24
Fix output raster type for variance in Cell statistics algorithm

root676:
## Description
This PR fixes an the wrong output raster data type for the variance function of the **Cell statistics** algorithm. If you input only rasters of type `Int32`, the variance statistic was written to an `Int32` output raster, truncating the variance result statistics. The output data type has now been set to `Float32` as minimum and `Float64` if all inputs are `Float64` by default. 

A testcase has been added.


nyalldawson (approved):
Thanks!

PR #38985, opened 2020-09-24
Fix QgsProcessingParameterNumber docstring example

SrNetoChan:
## Description

While trying to use this example from the PyQGIS API documentation this example looked wrong. There is no ``QgsProcessingParameter``

This PR updates the docstring, which should fix the API documentation.

PR #39013, opened 2020-09-24
Fix QgsProcessingParameterNumber docstring example

nyalldawson:
Supersedes https://github.com/qgis/QGIS/pull/38985


SrNetoChan:
Ha... You were quicker than me (obviously). I was now setting up pre-commit hooks and running sipify_all, by a suggestion from @elpaso.

At least I will be ready for next time :-)

Thanks!

nyalldawson:
@SrNetoChan no worries! Normally I wouldn't be so aggressive about pushing PRs along and would only supersede abandoned ones, but during freeze I want to keep the PR queue nice and small so we can do speedy reviews and keep the fixes flowing in :+1:

PR #39033, opened 2020-09-25
Avoid accessing QgsLayoutItemsListView's model before it is defined

### LINKED ISSUE #39032: Report editor crashes when right clicking on empty Items list
**Describe the bug**
QGIS crashes when right clicking on an empty Items list while editing a report.

**How to Reproduce**
1. Project > New report > give name > ok
2. Right click inside the empty *Items* list on the *Items* panel.
3. Crash

**QGIS and OS versions**
02f5c13ea0 on Debian

uclaros:
## Description
Avoid accessing `QgsLayoutItemsListView`'s model before it is defined

Fixes #39032

### DIFF
--- a/src/gui/layout/qgslayoutitemslistview.cpp
+++ b/src/gui/layout/qgslayoutitemslistview.cpp
@@ -114,6 +114,8 @@ void QgsLayoutItemsListView::setCurrentLayout( QgsLayout *layout )
 
 void QgsLayoutItemsListView::showContextMenu( QPoint point )
 {
+  if ( !mModel )
+    return;
   QModelIndex index = indexAt( point );
   QgsLayoutItem *item = mModel->itemFromIndex( index );
   if ( !item )


src/gui/layout/qgslayoutitemslistview.cpp
```diff
@@ -115,9 +115,9 @@ void QgsLayoutItemsListView::setCurrentLayout( QgsLayout *layout )
 void QgsLayoutItemsListView::showContextMenu( QPoint point )
 {
   QModelIndex index = indexAt( point );
-  QgsLayoutItem *item = mModel->itemFromIndex( index );
-  if ( !item )
+  if ( !index.isValid() )
     return;
+  QgsLayoutItem *item = mModel->itemFromIndex( index );
```
elpaso, line 120: Why don't you leave the original check for safety?
```c++
  if ( !item )
    return;
```

uclaros: If `index.isValid()` then `mModel->itemFromIndex( index )` should return an item. I think it's unreachable code, but I can leave it there, sure!

uclaros: Maybe a single `if ( !mModel ) return;` at the beginning of the method would be better?

nyalldawson: I agree

PR #39041, opened 2020-09-25
Set up cache geometry if HTML widget needs it

### LINKED ISSUE #34791: Form fails to initialize $geometry
Evaluating expressions involving $geometry inside an HTML widget in a form fails when displaying the attribute table, unless something else forces its evaluation for each feature already.

Minimal replicable failure case attached; see picture and zip.

image

The HTML field contains (in both layers) 
```
My geometry is '<script>document.write(expression.evaluate("geom_to_wkt($geometry)"));</script>'
```

In the first layer, it fails to evaluate property. (Testing on my real project shows $geometry is NULL). In the second layer, it evaluates fine, and the only difference is that I have added to that layer (but not on the form itself!) a virtual field with expression `geom_to_wkt($geometry)`. I assume that is forcing `$geometry` to be evaluated and stored somewhere, where it is then available for the form to use in the HTML field too.

The same happens in a QML widget.

Note however there is no problem - `$geometry` is not null - when testing with the preview in the `Configure HTML widget` window in the Form settings tab. It's only when you open the form in the attribute table that `$geometry` becomes null.

QGIS 3.12, OSGeo4W.

[geomnullerror.zip](https://github.com/qgis/QGIS/files/4270339/geomnullerror.zip)

troopa81:
## Description

Fixes #34791 : It's possible to define a geometry based expression in HTML widget, but it the vector layer cache doesn't require geometry the expression will fail. This PR propose to run an html rendering to guess if the expression needs the geometry and then set up the cache geometry option on the vector layer cache. 

Pretty sure same issue happen to QML widget. I'll create an issue and a separate PR for this one.


troopa81:
@elpaso Do you mind reviewing this PR, as you were the one who initially add the HTML widget?

troopa81:
don't stale

troopa81:
I rebased, could someone review and merge this PR ?

troopa81:
Rebase and all green!

troopa81:
Could someone review and merge this long standing PR please ? Thank you in advance

Houska1:
As the reporter of the bug that @troopa81 seems to have fixed, another vote of encouragement to get this reviewed and merged! Thanks

PR #39042, opened 2020-09-25
Copy LayerTreeNode's custom properties when converting project to offline

gacarrillor:
When converting a project to offline, QGIS should copy all original layerTreeNode's custom properties to the new one. 

Note that this is already done for QgsMapLayers (see https://github.com/qgis/QGIS/pull/35114).

### DIFF
--- a/src/core/layertree/qgslayertreenode.h
+++ b/src/core/layertree/qgslayertreenode.h
@@ -227,10 +227,31 @@ class CORE_EXPORT QgsLayerTreeNode : public QObject
     QVariant customProperty( const QString &key, const QVariant &defaultValue = QVariant() ) const;
     //! Remove a custom property from layer. Properties are stored in a map and saved in project file.
     void removeCustomProperty( const QString &key );
+
+    // TODO: Harmonize this with QgsMapLayer::customPropertyKeys() for QGIS 4 API
+
     //! Returns list of keys stored in custom properties
     QStringList customProperties() const;
     //! Remove a child from a node
     bool takeChild( QgsLayerTreeNode *node );
+
+    // TODO: Harmonize this with QgsMapLayer::customProperties() for QGIS 4 API
+
+    /**
+     * Read all custom properties from layer tree node
+     * \see setCustomPropertiesMap
+     * \since QGIS 3.16
+     */
+    const QgsObjectCustomProperties &customPropertiesMap() const;
+
+    // TODO: Harmonize this with QgsMapLayer::setCustomProperties() for QGIS 4 API
+
+    /**
+     * Set custom properties for layer tree node. Current properties are dropped.
+     * \see customPropertiesMap
+     * \since QGIS 3.16
+     */
+    void setCustomPropertiesMap( const QgsObjectCustomProperties &properties );
 
   signals:
 
--- a/python/core/auto_generated/layertree/qgslayertreenode.sip.in
+++ b/python/core/auto_generated/layertree/qgslayertreenode.sip.in
@@ -240,6 +240,8 @@ Read a custom property from layer. Properties are stored in a map and saved in p
 %Docstring
 Remove a custom property from layer. Properties are stored in a map and saved in project file.
 %End
+
+
     QStringList customProperties() const;
 %Docstring
 Returns list of keys stored in custom properties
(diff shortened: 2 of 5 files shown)


src/core/layertree/qgslayertreenode.h
```diff
@@ -227,11 +227,32 @@ class CORE_EXPORT QgsLayerTreeNode : public QObject
     QVariant customProperty( const QString &key, const QVariant &defaultValue = QVariant() ) const;
     //! Remove a custom property from layer. Properties are stored in a map and saved in project file.
     void removeCustomProperty( const QString &key );
+
+    // TODO: Harmonize this with QgsMapLayer::customPropertyKeys() for QGIS 4 API
```
nyalldawson, line 231: You could do that now -- add the new method and deprecate customProperties(). That'll save work when QGIS 4.0 time rolls around...

src/core/layertree/qgslayertreenode.h
```diff
@@ -227,11 +227,32 @@ class CORE_EXPORT QgsLayerTreeNode : public QObject
...
+    const QgsObjectCustomProperties &customPropertiesMap() const;
+
+    // TODO: Harmonize this with QgsMapLayer::setCustomProperties() for QGIS 4 API
+
+    /**
+     * Set custom properties for layer tree node. Current properties are dropped.
+     * \see customPropertiesMap
+     * \since QGIS 3.16
+     */
+    void setCustomPropertiesMap( const QgsObjectCustomProperties &properties );
```
nyalldawson, line 254: I'd go with just setCustomProperties directly here -- I realise it doesn't match the getter, but that's just a temporary thing which we can fix for 4.0

m-kuhn: Is the proposal to change the return value of `customProperties()` with 4.0?

I agree it would have been nicer to do that straight from the beginning. But now that we have the current signature, I'm -1 to change it.

Such changes are a nightmare:

- you never know which parts of a plugin API have been upgraded
- even worse if you try to keep a plugin compatible with 3.0 and 4.0 in parallel
- it will also render any existing documentation (gis.stackexchange etc) not only useless but misleading and frustrating

Using a fresh and collision free name here has way more advantages

src/core/layertree/qgslayertreenode.h
gacarrillor: Good, just a question: what would be preferred in the context of this PR?

In my local branch I've deprecated the method and while compiling I get a couple of warnings. Should I also point all existent calls to the new method?

I ask this because it will bring other parts of the source code (`qgslayoutlegendwidget.cpp` and `qgsmaplayerlegend.cpp`) into this PR. Is this something acceptable before releasing v3.16?

src/core/layertree/qgslayertreenode.h
nyalldawson: @m-kuhn 

> Is the proposal to change the return value of customProperties() with 4.0?

Yes -- see discussion above

> Using a fresh and collision free name here has way more advantages

I acknowledge your points, but the result is an inconsistent API which will be in place...forever.

m-kuhn: Inconsistent where exactly?

nyalldawson: There's inconsistencies all over the place:

QgsMapLayer:

- customPropertyKeys: returns a list of keys
- setCustomProperty
- customProperty
- setCustomProperties
- customProperties (returns QgsObjectCustomProperties)
- removeCustomProperty

QgsLayout and QgsLayoutObject have:
- setCustomProperty
- customProperty
- removeCustomProperty
- customProperties (returns list of keys)

QgsLayerTreeNode has:
- setCustomProperty
- customProperty
- removeCustomProperty
- customProperties (returns list of keys)

So the inconsistencies relate to whether customProperties returns either a list of keys or a reference to the properties object itself. Fixing QgsMapLayer to bring it into line with the other classes is less work, but then we have to find a good new name for the method which returns the object itself rather than just the keys. Alternatively we leave QgsMapLayer as is and bring the other classes into line with this...

m-kuhn: Can we use `customPropertKeys` and `customPropertiesMap`, that should be doable with minimal conflict?

nyalldawson: I'm not a huge fan of customPropertiesMap though -- that implies the result is a QMap, which it isn't...

Maybe customPropertiesRef? (although I don't particularly like that either!!)

m-kuhn: I guess yiu don't prefer customPropertiesCollection or ...Set either?

nyalldawson: Or we just improve the api and deprecate the inconsistent ones? :P Anyway, I wont stand in the way -- you clearly understand the point I was raising, so happy with whatever approach you decide to take to resolve it.

m-kuhn: In this case I would propose the following to get things into shape:

1. Rename `QgsObjectCustomProperties` to `QgsCustomPropertiesStore`. Add an alias for the "old" version. Deprecate
2. Rename getters/setters to `customPropertiesStore` and `setCustomPropertiesStore`
3. Add a new `QVariantMap QgsCustomPropertiesStore::toMap()` member and add a docstring /todo to the mMap member that this will be replaced with `toMap()` in QGIS 4

I hope that suits everyone.

nyalldawson: Sounds good!!!

nyalldawson:

PR #39052, opened 2020-09-26
Improve multicolumn legend layout

### LINKED ISSUE #37417: Legends in layouts 'Split Layers' with uneven Columns
Hello this is a bug that affects legend creation with columns in QGIS 3.14 and Ubuntu 20.04
I have a theme with 27 categories. I wish for the legend to display 9 columns, with 3 categories per column.
QGIS shows 4 columns with 3 categories, 4 categories, 2 categories and 1category, instead of showing 9 columns with 3 categories (9 · 3 = 27 categories).

**How to Reproduce**
Create a polygon layer with 27 polygons, create an Integer field ("Id"), assign value of id = @row_number.
Symbolize polygons by categories, using "Id" field. 
Create Layout, add Legend. Hide the layer's name.
In columns fill Columns with 9, activate Split Layers option.

Below is a caption of the situation.

imagen

I'm also attaching a zip file with QGZ and GPKG.
[Bug_Split_Theme_Columns.zip](https://github.com/qgis/QGIS/files/4832712/Bug_Split_Theme_Columns.zip)

*** I believe it may have something to do with the "Space between symbols". When its 0 - 0.5 mm, columns have 3 categories each. As I increase its value, columns begin to scramble.

"Space between symbols" = 0,5 mm
imagen

"Space between symbols" = 1,0 mm
imagen

"Space between symbols" = 3,0 mm
imagen

Andres Bodini

**FIRST EDIT:**
Apparently the error could be in the file: QGIS-master/src/core/qgslegendrenderer.cpp
However I don't know C++..... 

I attach the file in case anyone that knows C++ can help.

[qgslegendrenderer.cpp.zip](https://github.com/qgis/QGIS/files/4973434/qgslegendrenderer.cpp.zip)

jdugge:
## Correct calculation of target column heights in multicolumn legends

Multicolumn legends sometimes have uneven column lengths even in cases where equally sized columns would be possible. This is caused by the fact that the heuristic currently calculates the target column height by summing the height of all legend items including the space between items and dividing it by the number of columns, without taking into account that the first item in each column has no space above it.

This pull request corrects this behavior by reducing the total height of all legend items by the inter-item space that is not required due to the column breaks.

Fixes #37417


jdugge:
The change leads to a different output for the `testColumnBreaks4()` unit test.

Old behaviour:
expected_legend_column_breaks4

New behaviour:
legend_column_breaks4

The new layout is less tall than the old result, while still respecting the manual column breaks. I think that the new behavior is preferable.

bru500:
Great, thank you! Missed this much better behaviour.

nyalldawson:
Nice fix!

I agree about the reference image -- can you update this so that the tests pass again?

jdugge:
@nyalldawson Done

nyalldawson (approved):

nyalldawson:
Thanks @jdugge !

Issue #38978, opened 2020-09-23
Add Vector Layer hang QGIS on complex, multi-layer project


jjk0giap:
In QGIS 3.10, very big project, ~200 vector layers with many rule-based styles, addidng new vector layer - freeze QGIS, with no crash. 
With small projects there is no problem, same project on QGIS 2.18 - layers add with no problem, so this is regression. 

**How to Reproduce**
1. Open any very big project
2. Go to Add Vector Layer dialog
3. Select multiple files, (I've tried on SHP)
4. Click Add 
5. Layers are added to layer tree, but QGIS freeze on canvas refresh. Add Layer  Dialog cannot be closed, or even moved.

**QGIS and OS versions**
Affected QGIS versions - 3.10, 3.12, 3.14, 3.15 nightly
Windows 10 64bit

**Context**
I've tried with --noplugins option and with new user profile.

gioman:
@jjk0giap can you please try with a new QGIS profile to check if it makes any difference? thanks.

jjk0giap:
Prease read **Context**

gioman:
> Prease read **Context**

@jjk0giap that was added later, I answered based on the notification I received:

![Screenshot_20200923_142525](https://user-images.githubusercontent.com/1951107/94018592-b0288500-fda8-11ea-8b4c-a6f528c6366f.png)

jjk0giap:
New clue: drag&drop with test 12 SHP files:
1. Empty project is very fast, usuable almost immediately
2. Big project is very slow, takes 1-2 minute to refresh canvas after layers are added, later works normally.
Same for drag&drop Spatialite database.

matosang:
I have this exact same issue.

matosang:
One thing I've noticed is that after a new layer is added to layer tree, and when QGIS freeze on canvas refresh, if I leave the application to grind away, it can eventually become functional, but it can take 10-20 or more minutes. Not sure why it takes so long...

petercornelissen:
I have a problem that is probably related to this. I've tried new profiles, rebuilding the project etc. and it all works fine for a while then eventually it freezes when trying to add a new lay, pan a raster, move a layer etc. for 5 or 10 minutes. If you save the project, exit and restart it all works fine again until you add something else. I'm now using GeoPackages, but I had the same problem with ShapeFiles and why I went to GeoPackages hoping that would solve the problem, but it didn't make any difference. 3.12 and 3.14 both exhibit the probelm.

[...]

PR #39066, opened 2020-09-28
[layer tree] Avoid needless layer tree model computations for scale-based layer visibility

nirvn:
## Description

This PR optimizes the layer tree model by avoiding needless computations around scale-based layer visibility. Optimizations include:
- If the map canvas scale hasn't changed, avoid triggering scale-based layer visibility check loop
- In cases scale _has_ changed, make sure the model only emits dataChanged for layers that go from hidden to visible (or vice versa)
- Restrict dataChanged roles to _only_ the ones that are impacted by visibility change

With the PR applied, I can't see any freeze when panning on the canvas with the test project in #38978 .


m-kuhn (approved):
Works and code looks good

Issue #39058, opened 2020-09-27
Wrong assignments of raster values to symbology Paletted/Unique values or Singleband Pseudocolor


PedroVenancio:
I'm seeing some inconsistencies when styling raster layers.

Here is a sample raster (Float32) that has pixels with 4 different values:

[prob0115_sample.zip](https://github.com/qgis/QGIS/files/5289216/prob0115_sample.zip)

```
 - 0;
 - 3.450000047683716;
 - 6.900000095367432;
 - 10.350000381469727.
```
 
Loading it to QGIS, the default Singleband Gray render type shows the 4 values with different levels of gray.

- If I try to apply a Paletted/Unique Values render style, to get different colours for each value, canvas shows these different colours, but legend on TOC display wrong values for each one:

```
 - 0;
 - 3.78333;
 - 6.62083;
 - 10.4042.
```

![imagem](https://user-images.githubusercontent.com/1331581/94377515-5ca49700-011a-11eb-87f4-f3f11a91ade1.png)

As you can see, value in Symbology dialogue only show 0, 3, 6 and 10, and labels are different from raster values (not even rounded).

- If I try to use Singleband Pseudocolor as render type, with Exact interpolation, to get 4 particular colours to each one of the 4 raster values, I get some classes not rendered in canvas (just transparent):

![imagem](https://user-images.githubusercontent.com/1331581/94377720-b0fc4680-011b-11eb-95b9-e6a93ee8aec2.png)

However, if I assign the exact raster values manually, with full precision (all decimal places), they are already rendered:

![imagem](https://user-images.githubusercontent.com/1331581/94377812-33850600-011c-11eb-9b40-92cab22cd7e4.png)

Here is a screencast that shows this behaviour: https://cld.pt/dl/download/02852983-9f11-439d-bc81-33b58ade21bc/raster_symbology.mp4 

So, I believe that all these problems come from the raster precision that is not captured or is rounded by symbology engine. Anyone confirm?

Tested with 

QGIS version | 3.14.16-Pi | QGIS code revision | 2b2f966720
-- | -- | -- | --
Compiled against Qt | 5.11.2 | Running against Qt | 5.11.2
Compiled against GDAL/OGR | 3.2.0dev | Running against GDAL/OGR | 3.2.0dev
Compiled against GEOS | 3.8.1-CAPI-1.13.3 | Running against GEOS | 3.8.1-CAPI-1.13.3
[...]

PR #39071, opened 2020-09-28
Raster classification: show number of digits according to data type

elpaso:
Not really a bug fix but related to #39058:

- depending on data type, show a different number of digits when
  classifying the raster
- code cleaning

Note that there will be a separate followup PR to allow the user to define the number of decimal places in the legend label.


elpaso:
@PedroVenancio can you please test this patch? It is only about the display of the right number of decimals.

PedroVenancio:
@elpaso This one is already in master! Really great! Waiting for https://github.com/qgis/QGIS/pull/39079 merge (witch seems like a new feature, but that helps solving an issue)!

PR #39073, opened 2020-09-28
Split the geometry_overlay_contains function help into variants

DelazJ:
Before I extend to the other functions, I'd appreciate feedback on the wording.
 
image

### DIFF
--- a/resources/function_help/json/geometry_overlay_contains
+++ b/resources/function_help/json/geometry_overlay_contains
@@ -2,15 +2,25 @@
   "name": "geometry_overlay_contains",
   "type": "function",
   "groups": ["GeometryGroup"],
-  "description": "Performs a spatial join of type CONTAINS. This returns an array of results of an expression evaluated on features from a different layer that CONTAINS the current feature, or, if no expression if provided, simply returns whether at least one feature from the other layer CONTAINS the current feature.",
-  "arguments": [
+  "description": "Performs a spatial join of type CONTAINS. This returns whether at least one feature from a layer contains the current feature, or an array of results of an expression evaluated using features that contain the current feature.",
+  "variants": [
+  {
+     "variant": "Boolean check",
+     "variant_description": "If called with only the layer parameter, the function evaluates whether at least one feature from the layer contains the current feature and returns True or false.",
+     "arguments": [ { "arg": "layer", "description": "the layer whose overlay is checked" } ],
+     "examples": [ { "expression": "geometry_overlay_contains('regions')", "returns": "True, if at least one feature from the \"regions\" layer contains the current feature." } ]
+  },
+  {
+    "variant": "Array of overlaid features",
+    "variant_description": "If called with an expression parameter, the function evaluates the expression on the features from the layer and returns an array of results for the features that contain the current feature.",
+    "arguments": [
     {
       "arg": "layer",
-      "description": "the other layer"
+      "description": "the layer whose overlay is checked"
     },
     {
       "arg": "expression",
-      "description": "an optional expression to evaluate on the features from the other layer (if not set, the function will just return a boolean indicating whether there is at least one match)",
+      "description": "an optional expression to return the matching features from the layer. If not set, the function will return a boolean indicating whether there is at least one match.",
       "optional": true
     },
     {
@@ -27,14 +37,10 @@
       "arg": "cache",
       "description": "set this to true to build a local spatial index (most of the time, this is unwanted, unless you are working with a particularly slow data provider)",
       "optional": true,
-      "default": false
+      "default": "false"
     }
-  ],
-  "examples": [
-    {
-      "expression": "geometry_overlay_contains('regions')",
-      "returns": "True"
-    },
+    ],
+    "examples": [
     {
       "expression": "geometry_overlay_contains('regions', name)",
       "returns": "['South Africa', 'Africa', 'World']"
@@ -47,5 +53,7 @@
       "expression": "geometry_overlay_contains('regions', name, limit:=1)",
       "returns": "['South Africa']"
     }
-  ]
+    ]
+   }
+ ]
 }


resources/function_help/json/geometry_overlay_contains
```diff
@@ -2,15 +2,25 @@
   "name": "geometry_overlay_contains",
   "type": "function",
   "groups": ["GeometryGroup"],
-  "description": "Performs a spatial join of type CONTAINS. This returns an array of results of an expression evaluated on features from a different layer that CONTAINS the current feature, or, if no expression if provided, simply returns whether at least one feature from the other layer CONTAINS the current feature.",
-  "arguments": [
+  "description": "Performs a spatial join of type CONTAINS. This returns whether at least one feature from a layer contains the current feature, or an array of results of an expression evaluated using features that contain the current feature.",
```
DelazJ, line 5: Do we have a link we can point to, explaining what a contains spatial join is? something like https://postgis.net/docs/ST_Contains.html? The GEOS website is hard to browse to me.

resources/function_help/json/geometry_overlay_contains
```diff
@@ -27,14 +37,10 @@
       "arg": "cache",
       "description": "set this to true to build a local spatial index (most of the time, this is unwanted, unless you are working with a particularly slow data provider)",
       "optional": true,
-      "default": false
+      "default": "false"
```
m-kuhn, line 40: Is "default" different from "optional" (you quoted `"false"` but not `true` one line above)

resources/function_help/json/geometry_overlay_contains
m-kuhn: Do we have something in our docs?

DelazJ: The "only" place in docs we expose this kind of predicates is at [extract by location alg](https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/vectorselection.html#extract-by-location) and alike (reminds me that we should adapt this to the "join by location"-based algs)

The "source" is available as html at https://docs.qgis.org/testing/en/docs/user_manual/processing_algs/qgis/qgis_algs_include.html but this is a hidden,  internal and "garbage" file I'm not sure we want to expose.

resources/function_help/json/geometry_overlay_contains
DelazJ: `Optional` is whether the argument can be ignored or not so I guess it expects a Boolean (and as is, it renders correctly in GUI). `Default` is the value it takes if not used, and looking to the current GUI, it's not displayed, while quoting the value makes it appear in the dialog (as you can see above). I don't know what happens if not quoted.
Sorry. I can explain only based on experience (the behavior I see in QGIS), not code wise.

resources/function_help/json/geometry_overlay_contains
m-kuhn: Too bad. I think it would be good to cover such concepts somewhere in our documentation but am not experienced enough with the structure to advice where.

m-kuhn: To follow up on this, ideally (IMHO) we would add this to our docs but linking to postgis sounds fine too, it's correct in content but weird in context.

DelazJ: Quickly "browsing" the docs, I can't see where we would fluently add this but I'm surprised no project already sketched these concepts.
Yep, I can see the weirdness you mention. How about:
```suggestion
  "description": "Performs a spatial join of type CONTAINS. This returns whether at least one feature from a layer contains the current feature, or an array of results of an expression evaluated using features that contain the current feature.<br>This function follows the PostGIS <a href='https://postgis.net/docs/ST_Contains.html'>ST_CONTAINS</a> function.",
```

m-kuhn: "...the GEOS functionality which is documented in PostGIS"? Would be more accurate but maybe too complex to read?

DelazJ:
I am wondering; should `geometry_overlay_contains(layer)` returns whether:
- the current feature **is contained in** a feature of the target layer _(current behavior)_
- or the current feature **contains** a feature of the target layer ?

Same question for geometry_overlay_within. I don't know what is more logical and whether we already have functions that take other layer as parameter and their logic.

And unrelated question: could `geometry_overlay_function` be shorten to `overlay_function`? Or do we really need the geometry_ part?

m-kuhn:
CC @olivierdalang @enricofer

enricofer:
> I am wondering; should `geometry_overlay_contains(layer)` returns whether:
> 
>     * the current feature **is contained in** a feature of the target layer _(current behavior)_
> 
>     * or the current feature **contains** a feature of the target layer ?
> 
> 
> Same question for geometry_overlay_within. I don't know what is more logical and whether we already have functions that take other layer as parameter and their logic.

The current implementations of geometry_overlay_contains and geometry_overlay_within are correct as the spatial predicate refers to current feature, as @DelazJ suggests. The issue is in the Function help where in the descriptions of the behaviour for contains and within, current feature and features contained in target layer are inverted. 

> 
> And unrelated question: could `geometry_overlay_function` be shorten to `overlay_function`? Or do we really need the geometry_ part?

For me is a reasonable shortening.

DelazJ:
Thanks for the feedback @enricofer. You are right, the help was inverting the elements.
PR for functions rename at #39174

DelazJ:
I gave this function a closer look and I'm not convinced that the doc split as suggested here is fully correct. The function can return:
* an array: `overlay_contains(layer, expression, filter, limit, cache)`
 * a boolean: `overlay_contains(layer, filter, limit, cache)`

Due to only the expression parameter making the difference, showing the variants (with parameters description and examples) would lead to a huge help, that may be confusing. 
An alternative at #39389

enricofer:
The boolean output mode is intended for a straightforward usage in expression contexts where it suffits to know if a certain feature respect or not a given geospatial predicate, i.e. in ruled symbology. so in very most cases the syntax `overlay_contains(layer)` suffits
The other parameters does not make really sense unless for _filter_ param in case we would check only certain feature of the target layer. In this case we should use the function with this syntax: `overlay_contains(layer, null, filter)` 
For sure _limit_ param has no sense in this case because the boolean output functions returns true as soon as it find a feature candidate.
I'm not sure if really helps the user including all these use cases in help. In my opinion the brief boolean output function will mostly be used only with _layer_ param and I would prefer a concise help leaving sophisticated tweaks that might be confusing.  The current help content appears to me clear https://github.com/qgis/QGIS/pull/39389.

PR #39075, opened 2020-09-29
Move widgets from Q(Double)SpinBox to Qgs(Double)SpinBox 

DelazJ:
based on https://github.com/qgis/QGIS/pull/38948#issuecomment-697958015. Anything else to do?

Couldn't find how to make some plugins and providers dialogs use it.

### DIFF
--- a/src/plugins/geometry_checker/CMakeLists.txt
+++ b/src/plugins/geometry_checker/CMakeLists.txt
@@ -45,6 +45,7 @@ INCLUDE_DIRECTORIES(
   ${CMAKE_SOURCE_DIR}/src/core/effects
   ${CMAKE_SOURCE_DIR}/src/core/providers/ogr
   ${CMAKE_SOURCE_DIR}/src/gui
+  ${CMAKE_SOURCE_DIR}/src/gui/editorwidgets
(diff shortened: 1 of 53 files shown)


nyalldawson (approved):
You'll need to add `${CMAKE_SOURCE_DIR}/src/gui/editorwidgets` to 
https://github.com/qgis/QGIS/blob/1f7b710e00ff675c88d599eb20a6392b7931cb40/src/providers/grass/CMakeLists.txt#L13 to fix the mac build failure

nyalldawson:
Lovely work!

DelazJ:
The only remaining QSpinBox widgets are available in src/auth/oauth2/qgsauthoauth2edit.ui and src/ui/qgsarcgisvectortileconnectiondialog.ui. No idea where to include the widget to avoid build fail.

nyalldawson:
@DelazJ I don't think it's possible to address the auth ones without a considerable (tricky) reworking of the auth providers.

nyalldawson:
@DelazJ one consequence of this change is that many spin boxes are now showing clear buttons even when set to their initial default values. It's painful, but we should do a sweep of the buttons changed here and ensure that their clear value is correctly set to a reasonable default value...

DelazJ:
Ok. I Will take a look...

PR #39079, opened 2020-09-29
Raster singleband legend precision

elpaso:
Followup #39058 and #39071 I don't really know if this qualifies as a bugfix or a feature.

immagine

CC: @gioman @PedroVenancio

### DIFF
--- a/src/gui/qgsguiutils.cpp
+++ b/src/gui/qgsguiutils.cpp
@@ -275,6 +275,48 @@ namespace QgsGuiUtils
     }
     return QSize( adjustedSize, adjustedSize );
   }
+
+  QString displayValue( const Qgis::DataType rasterDataType, const double value )
+  {
+
+    int precision { 9 };
+    switch ( rasterDataType )
+    {
+      case Qgis::DataType::Int16:
+      case Qgis::DataType::UInt16:
+      {
+        precision = 5;
+        break;
+      }
+      case Qgis::DataType::Int32:
+      case Qgis::DataType::UInt32:
+      {
+        precision = 10;
+        break;
+      }
+      case Qgis::DataType::Byte:
+      {
+        precision = 3;
+        break;
+      }
+      case Qgis::DataType::Float32:
+      {
+        precision = 9;
+        break;
+      }
+      case Qgis::DataType::Float64:
+      {
+        precision = 17;
+        break;
+      }
+      default:
+      {
+        precision = 9;
+      }
+    }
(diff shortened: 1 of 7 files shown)


src/gui/qgsguiutils.cpp
```diff
@@ -275,6 +275,48 @@ namespace QgsGuiUtils
...
+      {
+        precision = 9;
+        break;
+      }
+      case Qgis::DataType::Float64:
+      {
+        precision = 17;
+        break;
+      }
+      default:
```
nyalldawson, line 312: Can you drop the default clause and list all enum values explicitly here? Then we will get a build warning if new types are added and the developer knows they have to update this method accordingly

nyalldawson:
Regarding the UI: could we avoid the checkbox and instead set the clear value for the spin to use a "default" string?

I'm happy for this to be classed as a bug fix.

elpaso:
> Regarding the UI: could we avoid the checkbox and instead set the clear value for the spin to use a "default" string?

@nyalldawson  Actually, I removed the checkbox completely and set the max precision of the spinbox to the max significant digits. It looks simpler and more user friendly this way.

nyalldawson:
@elpaso got a screenshot of that?

elpaso:
> @elpaso got a screenshot of that?

Image updated

src/gui/raster/qgscolorrampshaderwidget.cpp
```diff
@@ -118,6 +120,18 @@ void QgsColorRampShaderWidget::setRasterDataProvider( QgsRasterDataProvider *dp
...
+  if ( mRasterDataProvider )
+  {
+    const int maxDigits { QgsGuiUtils::significantDigits( mRasterDataProvider->dataType( mBand ) ) };
+    // Get current value
+    const int currentPrecision { mLabelPrecisionSpinBox->value() };
+    mLabelPrecisionSpinBox->setMaximum( maxDigits );
+    if ( currentPrecision > maxDigits )
+    {
+      mLabelPrecisionSpinBox->setValue( maxDigits );
+    }
```
nyalldawson, line 133: Is this needed? Won't setting the maximum for the spin box already handle this situation?

elpaso: I'm not sure: does it? I'll check. There is also the question if valueChanged is triggered (it should).

nyalldawson (approved):

PR #39088, opened 2020-09-29
Raster paletted/unique float support

### LINKED ISSUE #39058: Wrong assignments of raster values to symbology Paletted/Unique values or Singleband Pseudocolor

elpaso:
Fixes #39058 an adds some small improvements in the
progress feedback.

This is technically an API change, even if it's small I'm not sure we can merge it now, I could create another ctor for Python API if needed.

### DIFF
--- a/src/core/raster/qgspalettedrasterrenderer.cpp
+++ b/src/core/raster/qgspalettedrasterrenderer.cpp
@@ -480,26 +480,26 @@ QgsPalettedRasterRenderer::ClassData QgsPalettedRasterRenderer::classDataFromRas
   // Collect unique values for float rasters
   if ( raster->dataType( bandNumber ) == Qgis::DataType::Float32 || raster->dataType( bandNumber ) == Qgis::DataType::Float64 )
   {
+    const QgsRectangle fullExtent { raster->extent() };
     QList<double> values;
-    if ( feedback )
-    {
-      // Start showing some progress
-      feedback->setProgress( 1 );
-    }
-    std::unique_ptr<QgsRasterBlock> block { raster->block( bandNumber, raster->extent(), raster->xSize(), raster->ySize(), feedback ) };
+
     if ( feedback && feedback->isCanceled() )
     {
       return data;
     }
+    const double yStep { fullExtent.height() / raster->ySize() };
     // Max MAX_FLOAT_CLASSES classes!
     qgssize col = 0;
     qgssize row = 0;
-    for ( ; row < static_cast<qgssize>( raster->ySize() ); ++row )
+    for ( row = raster->ySize() - 1; row >= 0; --row )
     {
       if ( presentValues >= MAX_FLOAT_CLASSES )
       {
         break;
       }
+      raster->extent();
+      const QgsRectangle rowExtent { fullExtent.xMinimum(), fullExtent.yMinimum() + yStep * row, fullExtent.xMaximum(), fullExtent.yMinimum() + yStep *( row + 1 ) };
+      std::unique_ptr<QgsRasterBlock> block { raster->block( bandNumber, rowExtent, raster->xSize(), 1, feedback ) };
       for ( col = 0; col < static_cast<qgssize>( raster->xSize() ); ++col )
       {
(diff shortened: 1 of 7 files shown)


src/core/raster/qgspalettedrasterrenderer.cpp
```diff
@@ -472,37 +475,98 @@ QgsPalettedRasterRenderer::ClassData QgsPalettedRasterRenderer::classDataFromRas
...
     {
-      data << Class( currentValue, QColor(), QLocale().toString( currentValue ) );
-      presentValues++;
+      return data;
+    }
+    const double yStep { fullExtent.height() / raster->ySize() };
+    // Max MAX_FLOAT_CLASSES classes!
+    qgssize col = 0;
+    qgssize row = 0;
+    for ( row = raster->ySize() - 1; row >= 0; --row )
```
nyalldawson, line 495: It would be more performant to iterate over the raster using QgsRasterIterator here -- row sized blocks like this will result in many data requests.

src/core/raster/qgspalettedrasterrenderer.cpp
```diff
@@ -472,37 +475,98 @@ QgsPalettedRasterRenderer::ClassData QgsPalettedRasterRenderer::classDataFromRas
...
   double presentValues = 0;
-  for ( int idx = 0; idx < histogram.binCount; ++idx )
+
+  // Collect unique values for float rasters
+  if ( raster->dataType( bandNumber ) == Qgis::DataType::Float32 || raster->dataType( bandNumber ) == Qgis::DataType::Float64 )
   {
-    int count = histogram.histogramVector.at( idx );
-    if ( count > 0 )
+    const QgsRectangle fullExtent { raster->extent() };
+    QList<double> values;
```
nyalldawson, line 485: A list isn't a great choice here -- could you use an ordered set instead? or unordered set and sort after collecting the values?

nyalldawson:
> This is technically an API change, even if it's small I'm not sure we can merge it now, I could create another ctor for Python API if needed.

I don't think that's an issue here

nyalldawson (approved):

PedroVenancio:
@elpaso This is working great with floats in 3.16.

The only thing I see is that the legend is not ordered by value:

imagem

imagem

It can be manually reordered, so, nothing problematic.

elpaso:
Can you file another ticket for the ordering issue please?

PR #39099, opened 2020-09-30
Add QgsCPLHTTPFetchOverrider class and use it in OGR provider 

rouault:
This utility class redirects GDAL's CPL HTTP calls through QgsBlockingNetworkRequest,
for GDAL >= 3.2 (see OSGeo/gdal#2991)

And also QgsBlockingNetworkRequest: add head(), put() and deleteResource() methods

### DIFF
--- a/src/core/providers/ogr/qgsogrprovider.cpp
+++ b/src/core/providers/ogr/qgsogrprovider.cpp
@@ -28,6 +28,7 @@ email                : sherman at mrcc.com
 #include "qgssettings.h"
 #include "qgsapplication.h"
 #include "qgsauthmanager.h"
+#include "qgscplhttpfetchoverrider.h"
 #include "qgsdataitem.h"
 #include "qgsdataprovider.h"
 #include "qgsfeature.h"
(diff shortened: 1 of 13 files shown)


nyalldawson:
This is great! 

One improvement I'd like to see here relates to the QgsSetRequestInitiatorClass macro which automatically sets the source location where a request is created, and it's used for transparency of network requests (it's visible in the network logger panel). I think it should be possible to create a similar macro which creates the QgsCPLHTTPFetchOverrider object while also using QgsSetRequestInitiatorClass to tag the source location in the request.

src/core/providers/ogr/qgsogrprovider.cpp
```diff
@@ -512,6 +513,17 @@ QgsOgrProvider::QgsOgrProvider( QString const &uri, const ProviderOptions &optio
                           mSubsetString,
                           mOgrGeometryTypeFilter );
 
+  if ( mFilePath.contains( QLatin1String( "authcfg" ) ) )
+  {
+    QRegularExpression authcfgRe( " authcfg='([^']+)'" );
```
nyalldawson, line 518: Is the space at the start deliberate?

src/core/qgsblockingnetworkrequest.h
```diff
@@ -106,6 +106,66 @@ class CORE_EXPORT QgsBlockingNetworkRequest : public QObject
...
+     *
+     * The optional \a feedback argument can be used to abort ongoing requests.
+     *
+     * The method will return NoError if the get operation was successful. The contents of the reply can be retrieved
+     * by calling reply().
+     *
+     * If an error was encountered then a specific ErrorCode will be returned, and a detailed error message
+     * can be retrieved by calling errorMessage().
+     *
+     * @since 3.18
```
nyalldawson, line 127: @since -> \since

src/core/providers/ogr/qgsogrprovider.cpp
rouault: yes. Taken from another occurence in that file

src/core/qgsblockingnetworkrequest.h
rouault: done

rouault:
>  I think it should be possible to create a similar macro which creates the QgsCPLHTTPFetchOverrider object while also using

done

nyalldawson (approved):

nyalldawson:
@rouault can you fix the conflicts here and then merge?

PR #39103, opened 2020-09-30
[print composer map] Followup fixes for map grids annotations/ticks

### LINKED ISSUE #39160: Feedback: QgsTextRenderer for drawing map grid text in layouts not HTML compatible
**Describe the bug**
Not really a bug, more like feedback on a new feature in Master.

Using QgsTextRenderer for grid labels in layout is great, but the option to render the text as html has limited value, since the text seems to be "placed" before it is rendered. 

This means that if you have html tags in your label string they will occupy space creating an open distance from the actual rendered text to the location where it should be placed.

Just as an example, using the following expression as a grid label source:
`'<span style="color:blue;">' || @grid_number || '</span><span style="color:red;">Degree</span>'`

Will show up as:
Skärmbild från 2020-10-03 16-04-51

**How to Reproduce**

See above.

**QGIS and OS versions**

QGIS version | 3.15.0-Master | QGIS code revision | ef7ecbf95b
-- | -- | -- | --
Compiled against Qt | 5.12.8 | Running against Qt | 5.12.8
PostgreSQL Client Version | 12.4 (Ubuntu 12.4-0ubuntu0.20.04.1) | SpatiaLite Version | 4.3.0a
OS Version | Linux Mint 20 | This copy of QGIS writes debugging output.
Active python plugins | LmOpenData; DataPlotly; latlontools; layout_loader; mmqgis; Qgis2threejs; qfieldsync; Discovery; postgis_geoprocessing; MetaSearch; processing; db_manager

**Additional context**

olivierdalang:
This PR fixes some issues following the work on rotated composer map grid ticks/annotations :

- [x] 1. Improves the UI by using group boxes (see https://github.com/qgis/QGIS/pull/39037)
- [x] 2. Adds some data-definable settings to some map grid options, as requested in #38676
- [x] 3. Fixes #39160 where layout map grid annotations went south when using HTML

## Screenshot

1.
image
image

2. 
image


andreasneumann:
Hi @olivierdalang 

I just tested this PR, but I think that something is broken currently:

I "disabled" the right and bottom coordinate labels, but both are shown on the map as "Show All".

The dd-defined overrides work though. If I set a static value of "disabled" in the expression editor, then the labels disappear as expected. So something with the comboboxes and the assignment of corresponding values seems to be broken.

image

andreasneumann:
Seems to work fine now - thanks!

Regarding the wording:
image

Would it be better to call it "Show Both" instead of "Show All"? I don't have a strong opinion though ...

src/core/layout/qgslayoutitemmapgrid.cpp
```diff
@@ -1894,6 +1895,22 @@ bool QgsLayoutItemMapGrid::shouldShowForDisplayMode( QgsLayoutItemMapGrid::Annot
          || ( mode == QgsLayoutItemMapGrid::LongitudeOnly && coordinate == QgsLayoutItemMapGrid::Longitude );
 }
 
+
+QgsLayoutItemMapGrid::DisplayMode gridAnnotationDisplayModeFromDD( QString ddValue, QgsLayoutItemMapGrid::DisplayMode defValue )
+{
+  if ( ddValue == QStringLiteral( "x_only" ) )
```
nyalldawson, line 1901: Can you make these comparisons case insensitive? Also requires some unit tests for this.

olivierdalang: done

olivierdalang:
I've just closed the other 2 related PRs and added the commits here, as it's too hard to maintain multiple PRs affecting same UI files and same expected images.

Left to do :
- [ ] determine if we're OK with nested groups in the UI
- [ ] update expected images to fix tests

src/core/layout/qgslayoutitemmapgrid.cpp
```diff
@@ -1894,6 +1899,22 @@ bool QgsLayoutItemMapGrid::shouldShowForDisplayMode( QgsLayoutItemMapGrid::Annot
...
 
+
+QgsLayoutItemMapGrid::DisplayMode gridAnnotationDisplayModeFromDD( QString ddValue, QgsLayoutItemMapGrid::DisplayMode defValue )
+{
+  if ( ddValue.toLower() == QStringLiteral( "x_only" ) )
+    return QgsLayoutItemMapGrid::LatitudeOnly;
+  else if ( ddValue.toLower() == QStringLiteral( "y_only" ) )
+    return QgsLayoutItemMapGrid::LongitudeOnly;
+  else if ( ddValue.toLower() == QStringLiteral( "disabled" ) )
+    return QgsLayoutItemMapGrid::HideAll;
```
nyalldawson, line 1910:
```suggestion
  if ( ddValue.compare( QLatin1String( "x_only" ), Qt::CaseInsensitive ) == 0 )
    return QgsLayoutItemMapGrid::LatitudeOnly;
  else if ( ddValue.compare( QLatin1String( "y_only", Qt::CaseInsensitive ) == 0 )
    return QgsLayoutItemMapGrid::LongitudeOnly;
```

etc (QString::compare is much more efficient vs multiple calculations of the entire lowercase string!)

nyalldawson:
Looking good -- you'll just need to update the reference image masks now and I'll merge

olivierdalang:
Yaay finally got parse_dash_results.py to work.

I updated the masks, wondering though how to decide whether we should update the mask VS replace expected image.

So this is ready to merge

@nyalldawson 
Just drawing your attention again about the nested groups in the UI you didn't seem to like too much, that are still in this PR. Let me know if you prefer that I remove the sub-groups.

nyalldawson (approved):

PR #39104, opened 2020-09-30
[FEATURE] [OGR provider] Add support for open options 

rouault:
* With "|option:FOO=BAR|option:FOO2=BAR2" in the URI

* [UI] OGR source select: add a dynamic layout to define open options

image

### DIFF
--- a/src/core/providers/ogr/qgsogrprovider.cpp
+++ b/src/core/providers/ogr/qgsogrprovider.cpp
@@ -222,11 +222,11 @@ void QgsOgrProvider::repack()
       QString errCause;
       if ( mLayerName.isNull() )
       {
-        mOgrOrigLayer = QgsOgrProviderUtils::getLayer( mFilePath, true, QStringList(), mLayerIndex, errCause, true );
+        mOgrOrigLayer = QgsOgrProviderUtils::getLayer( mFilePath, true, mOpenOptions, mLayerIndex, errCause, true );
       }
       else
       {
-        mOgrOrigLayer = QgsOgrProviderUtils::getLayer( mFilePath, true, QStringList(), mLayerName, errCause, true );
+        mOgrOrigLayer = QgsOgrProviderUtils::getLayer( mFilePath, true, mOpenOptions, mLayerName, errCause, true );
       }
 
       if ( !mOgrOrigLayer )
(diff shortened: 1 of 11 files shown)


nirvn:
@rouault , wouhou! This is fantastic, a long wished possibility.

I hope this will make it into 3.16 LTR. In many instances, having the possibility to set options can work around dataset issues ppl have been reporting over the years.

nirvn:
@rouault , working well here:
image

I think we need a better name than 'option options' for the collapsible groupbox header. I'd go for removing open and label it 'Options'.

For drivers that show option widgets in there, I think it'd be useful to add a QLabel with an hyperlink to the driver's help page. A list like the one CSV produces would benefit from having an hyperlink to reach https://gdal.org/drivers/vector/csv.html easily.

nirvn:
@rouault , one regression spotted: layers with a subset string that includes return carriages (i.e. '\n') aren't restored properly when projects are loaded.

nirvn:
@rouault , actually, the problem isn't directly linked to this PR, merely exposed through your use of encodeUri/decodeUri. I'll fix the underlying issue in a different PR.

rouault:
> I think we need a better name than 'option options' for the collapsible groupbox header. I'd go for removing open and label it 'Options'.
> 
> For drivers that show option widgets in there, I think it'd be useful to add a QLabel with an hyperlink to the driver's help page. A list like the one CSV produces would benefit from having an hyperlink to reach https://gdal.org/drivers/vector/csv.html easily.

Done. See updated screenshot

src/core/providers/ogr/qgsogrprovider.cpp
```diff
@@ -3516,10 +3521,17 @@ QString QgsOgrProviderMetadata::encodeUri( const QVariantMap &parts )
...
-         + ( !geometryType.isEmpty() ? QStringLiteral( "|geometrytype=%1" ).arg( geometryType ) : QString() )
-         + ( !subset.isEmpty() ? QStringLiteral( "|subset=%1" ).arg( subset ) : QString() );
+  const QStringList openOptions = parts.value( QStringLiteral( "openOptions" ) ).toStringList();
+  QString uri = path
+                + ( !layerName.isEmpty() ? QStringLiteral( "|layername=%1" ).arg( layerName ) : !layerId.isEmpty() ? QStringLiteral( "|layerid=%1" ).arg( layerId ) : QString() )
+                + ( !geometryType.isEmpty() ? QStringLiteral( "|geometrytype=%1" ).arg( geometryType ) : QString() )
+                + ( !subset.isEmpty() ? QStringLiteral( "|subset=%1" ).arg( subset ) : QString() );
+  for ( const QString &openOption : openOptions )
+  {
+    uri += QStringLiteral( "|openoption:" );
```
nyalldawson, line 3531: how would you feel about changing this to just "|option:xxxx=yyyy"? It's slightly more compact, and will help avoid layer uris growing too unwieldy...

src/gui/providers/ogr/qgsogrsourceselect.cpp
```diff
@@ -310,20 +317,46 @@ void QgsOgrSourceSelect::setProtocolWidgetsVisibility()
...
+      {
+        value = le->text();
+      }
+    }
+    if ( !value.isEmpty() )
+    {
+      openOptions << QStringLiteral( "%1=%2" ).arg( control->objectName() ).arg( value );
+    }
+  }
+
```
nyalldawson, line 347:
```suggestion
  for ( QWidget *control : mOpenOptionsWidgets )
  {
    QString value;
    if ( QComboBox *cb = qobject_cast<QComboBox *>( control ) )
    {
      value = cb->itemData( cb->currentIndex() ).toString();
    }
    else if ( QLineEdit *le = qobject_cast<QLineEdit *>( control ) )
    {
      value = le->text();
    }
    if ( !value.isEmpty() )
    {
      openOptions << QStringLiteral( "%1=%2" ).arg( control->objectName() ).arg( value );
    }
  }

```

src/gui/providers/ogr/qgsogrsourceselect.cpp
```diff
@@ -556,4 +634,156 @@ void QgsOgrSourceSelect::showHelp()
...
+             psOption->psChild == nullptr )
+        {
+          continue;
+        }
+        options << psOption->psChild->pszValue;
+      }
+    }
+
+    QLabel *label = new QLabel( pszOptionName );
+    QWidget *control;
```
nyalldawson, line 731:
```suggestion
    QWidget *control = nullptr;
```

nyalldawson:
I think we should merge this for 3.16 -- there's potentially a very large number of bug reports we can close just by exposing these options and directing users toward them, and the value exposed here (basically unlocking all GDAL functionality for the datasources, instead of just the default subset) would be highly desirable for the upcoming LTR.

@m-kuhn do you agree?

src/gui/providers/ogr/qgsogrsourceselect.cpp
rouault: done

src/core/providers/ogr/qgsogrprovider.cpp
rouault: done

m-kuhn:
I agree the benefits outweigh the risks 👍

nyalldawson (approved):

agiudiceandrea:
> > For drivers that show option widgets in there, I think it'd be useful to add a QLabel with an hyperlink to the driver's help page. A list like the one CSV produces would benefit from having an hyperlink to reach https://gdal.org/drivers/vector/csv.html easily.
> 
> Done. See updated screenshot

Hi @rouault it seems this doesn't work as expected: e.g. for an ESRI Shapefile source, the link points to a no longer existing page at https://gdal.org/drv_shape.html while the correct link should be https://gdal.org/drivers/vector/shapefile.html

https://github.com/qgis/QGIS/blob/ad01045f0f7838254ef392efdc48ee32988258ff/src/gui/providers/ogr/qgsogrsourceselect.cpp#L772

See also https://github.com/qgis/QGIS/issues/39647

rouault:
>  it seems this doesn't work as expected: e.g. for an ESRI Shapefile source, the link points to a no longer existing page at https://gdal.org/drv_shape.html while the correct link should be https://gdal.org/drivers/vector/shapefile.html

I've fixed that in GDAL 3.2

nirvn:
@rouault , is there an acceptable way to add a redirect on gdal.org to cover ppl that will run QGIS with GDAL < 3.2 ? Alternatively we can push a fix in QGIS.

rouault:
>  is there an acceptable way to add a redirect on gdal.org to cover ppl that will run QGIS with GDAL < 3.2

you could potentially add a redirect around https://github.com/OSGeo/gdal/blob/master/gdal/doc/source/_extensions/redirects.py#L33

src/gui/providers/ogr/qgsogrsourceselect.cpp
```diff
@@ -556,4 +629,156 @@ void QgsOgrSourceSelect::showHelp()
...
+      control = le;
+    }
+    control->setObjectName( pszOptionName );
+    mOpenOptionsWidgets.push_back( control );
+
+    const char *pszDescription = CPLGetXMLValue( psItem, "description", nullptr );
+    if ( pszDescription )
+    {
+      label->setToolTip( QStringLiteral( "<p>%1</p>" ).arg( pszDescription ) );
+      control->setToolTip( QStringLiteral( "<p>%1</p>" ).arg( pszDescription ) );
```
DelazJ, line 761: Hi, this PR introduces a number of strings in the dialog. These tooltips are helpful but is there a chance they could be made translatable, please? See https://lists.osgeo.org/pipermail/qgis-tr/2024-January/001387.html
Thanks.

rouault: There is no immediate straightforward way of making them translatable as that would require to extract them from GDAL first. I guess that could potentially be done, and I'm CC'ing @elpaso if he might be interested in looking that on GDAL side.
I would imagine a Python script querying the drivers, getting the XML of their creation options and extracting the "description" attribute to create a gettext .pot file
The main question to me is who would translate that then. I don't think GDAL itself has a sufficiently large i18n interested community to tackle that task (chicken & egg problem as there's no translation mechanism in GDAL currently), so maybe we should rely on QGIS community to do that.
So the process would be:
1. GDAL script creating .pot file
2. Importing that into QGIS and create/refresh associated .po file
3. Have QGIS community translate those GDAL .po files
4. Get the translated files (not sure how Transiflex works, but is it possible to get an individual file back?) and re-inject into GDAL, and add a gettext-based mechanism in GDAL to get translated strings

Step 4 is probably optional for a QGIS-only context, but would probably be needed to justify using GDAL sponsored time on that task (we'd want also other frontends to GDAL than QGIS to be able to benefit from translated strings).

Behind the technicalities, there's also a licensing issue. What is the license of translated files done in the QGIS context? Ideally, we'd want them to be usable under a liberal license, compatible of the GDAL MIT license if we re-inject them into GDAL.

PR #39109, opened 2020-10-01
[ogr] Fix decodeUri unable to handle newline characters in the subset string

nirvn:
## Description

The OGR provider's decoeUri function didn't handle newline characters in subset string properly, resulting in a broken decoded file path and missing subset part.

This was uncovered while testing PR #39104 .


src/core/providers/ogr/qgsogrprovider.cpp
```diff
@@ -3433,7 +3433,7 @@ QVariantMap QgsOgrProviderMetadata::decodeUri( const QString &uri )
     const QRegularExpression geometryTypeRegex( QStringLiteral( "\\|geometrytype=([a-zA-Z0-9]*)" ) );
     const QRegularExpression layerNameRegex( QStringLiteral( "\\|layername=([^|]*)" ) );
     const QRegularExpression layerIdRegex( QStringLiteral( "\\|layerid=([^|]*)" ) );
-    const QRegularExpression subsetRegex( QStringLiteral( "\\|subset=(.*)$" ) );
+    const QRegularExpression subsetRegex( QStringLiteral( "\\|subset=((?:.*|\n*)*)\\Z" ) );
```
nyalldawson, line 3436: Maybe add \r here too?

nirvn: @nyalldawson , done.

nyalldawson (approved):

PR #39114, opened 2020-10-01
New feature to the "Drop field(s)" algorithm

MrChebur:
## Description

Added a new parameter.
It allows you to delete all fields except the selected ones.
This feature can be very useful in the model builder as you cannot know all the field names before you start the processing.
Some details: https://gis.stackexchange.com/questions/375494/dropping-all-fields-exept-one-with-qgis-3-model-builder

EhmNlBpt5c


pathmapper:
This PR closes https://github.com/qgis/QGIS/issues/39345.

nyalldawson:
Great feature!

I'd prefer to see this exposed as a seperate tool to the existing "delete fields" algorithm. Ie. a "retain fields" algorithm. @alexbruy what do you think?

uclaros:
I love this!
I suggest we change *Fields to drop* to *Select fields* and then have the type of action selectable with a combo box *Drop selected (default)* or *Keep selected*

nyalldawson:
@MrChebur I've taken your work here and applied a modified version of it to  #40225 , where I've also ported both algorithms to c++ for speed and compile-time safety. Thanks for the inspiration!

PR #39090, opened 2020-09-29
Fix writing weak relations to XML

gacarrillor:
While hunting another bug, I came across a problem with the `QgsWeakRelation::writeXML()`.

See this: 

```xml
<!DOCTYPE qgis PUBLIC 'http://mrcc.com/qgis.dtd' 'SYSTEM'>
<qgis version="3.15.0-Master" styleCategories="Relations">
 <referencedLayers>
  <relation layerName="d_interesadodocumentotipo" referencingLayer="Usuarios_79987b60_ba7d_4764_a431_85f0395acc24" layerId="d_interesadodocumentotipo_3cdfd0b5_f3a5_47b3_8b4e_c5123b770ce7" id="usuarios_estado_fkey" name="usuarios_estado_fkey" referencedLayer="d_estadotipo_2dfc9c87_9a87_49d0_981b_bf7304ae6ccc" dataSource="dbname='fdc_model_02' host=localhost port=5432 user='postgres' key='t_id' checkPrimaryKeyUnicity='1' table=&quot;coord_cecilia&quot;.&quot;d_interesadodocumentotipo&quot;" providerKey="postgres" strength="Association">
   <fieldRef referencedField="t_id" referencingField="estado"/>
  </relation>
  <relation referencingLayer="Usuarios_79987b60_ba7d_4764_a431_85f0395acc24" id="usuarios_rol_fkey" name="usuarios_rol_fkey" referencedLayer="d_roltipo_33a4f8aa_df75_4be1_82b9_bfdd89385240" strength="Association">
   <fieldRef referencedField="t_id" referencingField="rol"/>
  </relation>
  <relation referencingLayer="Usuarios_79987b60_ba7d_4764_a431_85f0395acc24" id="usuarios_tipo_documento_fkey" name="usuarios_tipo_documento_fkey" referencedLayer="d_interesadodocumentotipo_3cdfd0b5_f3a5_47b3_8b4e_c5123b770ce7" strength="Association">
   <fieldRef referencedField="t_id" referencingField="tipo_documento"/>
  </relation>
 </referencedLayers>
 <layerGeometryType>4</layerGeometryType>
</qgis>
```

The first relation tag has mixed attributes (2 referenced layers are there involved: `d_estadotipo` and `d_interesadodocumentotipo`).

Additionally, only the first relation element gets the weak relation attributes.

With the changes of the current PR, now I get this, as expected:

```xml
[...]

PR #39126, opened 2020-10-01
Test for writing weak relation attributes to XML (followup #39090)

gacarrillor:
New test. That's it. 
See #39090.


elpaso:
Thank you @gacarrillor !

PR #39128, opened 2020-10-02
Fix all remaining Qt QFlags deprecated constructor warnings

nyalldawson:


nyalldawson:
Unrelated test failure

PR #39131, opened 2020-10-02
[Oracle] Fix new connection greyed ok button

### LINKED ISSUE #38979: Oracle - add/edit connection dialog forces hostname input - conflicts with Oracle Network Connection
**Description**

Creating or editing an Oracle connection with the data source manager is not possible with a version > 3.10.x, the validation control is greyed out

**How to Reproduce**

1. Go to Data Source manager > Oracle
2. Create or edit a connection
3. Set all the required parameters and test the connection
4. Try to click the button **"OK"** to save or update the connection

**QGIS and OS versions**

Last version tested that works -> 3.10.3
Last version tested -> 3.15.0-Master 9305e22a6c
Other versions tested 3.12 3.14

troopa81:
## Description

Fixes #38979 : When creating a new Oracle connection, database can be either the database name or a service name (described in tnsnames.ora file), so we don't have to disable the OK button when only database and name are set. 

At first, I planned to add a *Service* field like it's done in Postgres but it would break all Oracle existing configuration, so better not to.

### DIFF
--- a/src/providers/oracle/qgsoraclenewconnection.cpp
+++ b/src/providers/oracle/qgsoraclenewconnection.cpp
@@ -195,6 +195,11 @@ void QgsOracleNewConnection::showHelp()
 
 void QgsOracleNewConnection::updateOkButtonState()
 {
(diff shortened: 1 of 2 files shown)


troopa81:
@alexbruy Could you review this PR since its related to #35838

alexbruy (approved):

PR #39133, opened 2020-10-02
Commit data on scale changed for advanced snap configuration min/max scale

### LINKED ISSUE #39031: Scale dependent snapping does not activate as long as it displays the scale as "1:xxxxxx"
**Describe the bug**
When defining per layer scale dependent snapping, the scale is not considered unless focus has changed from the scale drop down to another widget of the individual layer settings and the scale display changes from `1:1000` to `1000`.

See screencast:
snap_scale

**How to Reproduce**
1. Enable advanced snapping with per layer scale dependent snapping
2. Select min scale for a layer from the drop down.
3. Close the advanced configuration widget
4. The value set is not respected

**QGIS and OS versions**
02f5c13ea0 on Debian

**Additional context**
ping @obrix

troopa81:
## Description 

Fixes #39031 : commitData signal must be sent so the item view is aware that model needs to be updated

@obrix Is there any special reason why you use a QgsScaleWidget instead of a QgsScaleComboBox. The former just adds a button to get current scale from canvas but was not set as visible.

### DIFF
--- a/src/app/qgssnappinglayertreemodel.cpp
+++ b/src/app/qgssnappinglayertreemodel.cpp
@@ -129,6 +129,7 @@ QWidget *QgsSnappingLayerDelegate::createEditor( QWidget *parent, const QStyleOp
   {
     QgsScaleWidget *minLimitSp = new QgsScaleWidget( parent );
     minLimitSp->setToolTip( tr( "Minimum scale from which snapping is enabled (i.e. most \"zoomed out\" scale)" ) );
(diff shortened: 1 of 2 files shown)


obrix:
@troopa81 I don't think there was a specific reason. If you replace it by the combobox directly with no issue, you can probably also replace it in QgsSnappingWidget for the global min/max scale.

troopa81:
@obrix I guess I was wrong, replacing *QgsScaleWidget* with *QgsScaleComboBox* breaks the advanced snapping configuration flotatting window (though it's the same widgets). So I keep the QgsScaleWidget and just add the fix for the model update.

nyalldawson (approved):

PR #39150, opened 2020-10-03
Cleanup color handling in QgsCodeEditor classes

nyalldawson:
Remove a bunch of duplicate default color definitions, and ensure
that all subclasses correctly respect any user defined color overrides,
yet that we still default to following the application theme's matching
color schemes.


nyalldawson:
@nirvn your review/testing would be appreciated!

nirvn (approved):
Most welcome cleanup.

PR #39175, opened 2020-10-04
Make it possible to use 'inbuilt:'-uri with OGR provider

rduivenvoorde:
Not sure if this is still handy, but I was looking for some code to use the world map in the PyQGIS cookbook...

Loading the easter egg 'world' actually loads an inbuild gpkg from
the (install dir) resources directory.
It ends up in the project file with the eg following uri:
 source="inbuilt:/data/world_map.gpkg|layername=countries"
This commit makes it possible to actually (re)use that uri in the
OGR provider.

So in pyqgis you can now do:
```
iface.addVectorLayer('inbuilt:/data/world_map.gpkg|layername=countries', 'Countries', 'ogr')
```

I do not think this breaks anything?
I'm aware this also works
```
iface.addVectorLayer(QgsApplication.pkgDataPath() + '/resources/data/world_map.gpkg|layername=Countries', '', 'ogr')
```

### DIFF
--- a/src/core/providers/ogr/qgsogrprovider.cpp
+++ b/src/core/providers/ogr/qgsogrprovider.cpp
@@ -4616,6 +4616,12 @@ void QgsOgrProvider::open( OpenMode mode )
   {
     mFilePath += ",tables=" + mLayerName;
   }
+  else if ( mFilePath.startsWith( QLatin1String( "inbuilt:" ) ) )
+  {
+    // be able to use an 'inbuild:'-uri to load ogr data from QGIS resources
+    // eg in python: iface.addVectorLayer('inbuilt:/data/world_map.gpkg|layername=countries', 'Countries', 'ogr')
+    mFilePath = QgsApplication::pkgDataPath() + QStringLiteral( "/resources" ) + mFilePath.mid( 8 );
+  }
 
   if ( mode == OpenModeForceReadOnly )
     openReadOnly = true;


nyalldawson (changes requested):
This needs to be done in a different way -- it's not a provider level responsibility (and infact the OGR provider already works with it for layers coming from projects).

Rather you'd need to edit QgisApp::addVectorLayerPrivate to extract the filename from the vectorLayerPath argument (via QgsProviderRegistry::instance()->decodeUri() ) and then push the returned path value through QgsPathResolver().readPath() before updating the parsed uri map with the resolved value, and then reencode this uri for use in the rest of addVectorLayerPrivate.

Let me know if you'd like to tackle this or want me to...

rduivenvoorde:
Thanks @nyalldawson ! I will try this myself first, if in trouble I'll call for help :-)
Should I just use this branch, or should I just create a new one?

nyalldawson:
Totally up to you and what git workflow you're comfortable with! :D

arongergely:
Maybe it's better to implement a convenient way to resolve the path of "inbuilt:" to absolute paths. Then the output could be used in the OGR provider call, the usual way.

Curently `QgsPathResolver` resolves the 'inbuilt:' paths when the project file is opened: https://github.com/qgis/QGIS/pull/31396


By the way there have been talks before about elevating the world layer from 'easter egg' status.
Could this be the time?

It is already used as the base layer for the projection bounds overview. (Implemented in `QgsCoordinateBoundsPreviewMapWidget`)

rduivenvoorde:
@arongergely note that 
```
iface.addVectorLayer(QgsApplication.pkgDataPath() + '/resources/data/world_map.gpkg|layername=Countries', '', 'ogr')
```
 works...
So you can use as 'inbuilt:' :  `QgsApplication.pkgDataPath() + '/resources/data/'` I think?

arongergely:
Yes that should work! It was a long while ago - but I remember I hit some issues when i was implementing code for multiplatform use - . Considering people may use to cookbook on windows..  UNIX / Windows file separators issue? backslash / forward slash?

arongergely:
But in any case one could just handle that via the `os` module:
```iface.addVectorLayer(os.path.join(QgsApplication.pkgDataPath(),'resources','data','world_map.gpkg|layername=Countries'), '', 'ogr')```

PR #39176, opened 2020-10-04
Code linting: modify "if ( foo() ) { foo()->bar }"-like constructs

rouault:
cppcheck sometimes detects this as a potential null pointer dereference.
This makes sense, although it is quite unlikely that the return a foo()
might change in between.

To avoid such warnings, and also repeated method calls which can have some
cost, I've run the following refactoring script to store the result of
foo() in a local variable.

```shell
for i in `find ../src -name "*.cpp"`; do python analyze.py $i > $i.tmp; diff $i.tmp $i >/dev/null || (echo "Paching $i"; mv $i.tmp $i); rm -f $i.tmp; done
```

with analyze.py being

```python
import re
import sys

lines = [l[0:-1] if l[-1] == '\n' else l for l in open(sys.argv[1], "rt").readlines()]

# Look for "if ( foo() )" or "if ( foo.bar() )"
if_pattern = re.compile("(^[ ]*)if \( ([a-zA-Z0-9_\.]+)\(\) \)$")

i = 0
while i < len(lines):
    line = lines[i]
    modified = False

    m = if_pattern.match(line)
    if m:
        next_line = lines[i+1]
        indent = m.group(1)
        s = m.group(2) + "()"
        tmpVar = m.group(2).split('.')[-1]
        tmpVar = 'l' + tmpVar[0].upper() + tmpVar[1:]

        if next_line == indent + '{':
            found = False

            # Look in the block after the if() for a pattern where we dereference
            # "foo()"
            j = i + 1
            while lines[j] != indent + '}':
                if lines[j].find(s + '->') >= 0 or lines[j].find('*' + s) >= 0 or lines[j].find(s + '[') >= 0:
                    found = True
                    break
                j += 1

            if found:
                print(indent + 'if ( auto *' + tmpVar + ' = ' + s +  ' )')
                j = i + 1
                while lines[j] != indent + '}':
                    print(lines[j].replace(' ' + s, ' ' + tmpVar).replace('.' + s, '.' + tmpVar).replace('*' + s, '*' + tmpVar).replace('!' + s, '!' + tmpVar))
                    j += 1
                print(lines[j])
                modified = True
                i = j

        else:
            if next_line.find(s) >= 0:
                print(indent + 'if ( auto *' + tmpVar + ' = ' + s +  ' )')
                print(next_line.replace(' ' + s, ' ' + tmpVar).replace('.' + s, '.' + tmpVar).replace('*' + s, '*' + tmpVar).replace('!' + s, '!' + tmpVar))
                modified = True
                i += 1

    if not modified:
        print(line)
    i += 1
```

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -4186,10 +4186,10 @@ void QgisApp::setupConnections()
     {
       canvas->setCanvasColor( backgroundColor );
     }
-    if ( mapOverviewCanvas() )
+    if ( auto *lMapOverviewCanvas = mapOverviewCanvas() )
     {
-      mapOverviewCanvas()->setBackgroundColor( backgroundColor );
-      mapOverviewCanvas()->refresh();
+      lMapOverviewCanvas->setBackgroundColor( backgroundColor );
+      lMapOverviewCanvas->refresh();
     }
   } );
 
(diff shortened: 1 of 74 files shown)


src/app/qgisapp.cpp
```diff
@@ -4186,10 +4186,11 @@ void QgisApp::setupConnections()
     {
       canvas->setCanvasColor( backgroundColor );
     }
-    if ( mapOverviewCanvas() )
+    auto &&lMapOverviewCanvas = mapOverviewCanvas();
+    if ( lMapOverviewCanvas )
```
nyalldawson, line 4190: Not sure if your script can be updated, but a much cleaner result would be:

```
   if ( auto* lMapOverviewCanvas = mapOverviewCanvas() )
   {
     ...
   }
```

is that possible?

rouault: > is that possible?

That's the whole point of using a script :-) Done

PR #39186, opened 2020-10-05
Remove QgsPythonConsoleBase so that we have a single definitive Python editor class (QgsCodeEditorPython)

nyalldawson:
One Python code editor widget to rule them all!

Now all our Python editors support a common appearance, behaviour and autocompletion support.


nyalldawson:
(actually there's one last piece of functionality which needs moving from console to QgsCodeEditor -- the autocompletion of brackets)

PR #39190, opened 2020-10-06
Move code editor color scheme tracking to dedicated registry

nyalldawson:
This cleans up the code, and makes it possible for creation of plugins which add new code editor color schemes!


nirvn (approved):
Nice one.

PR #39192, opened 2020-10-06
Fix message log dock causes havoc with other docks placed in the same area

### LINKED ISSUE #27868: Log Messages Panel disappears
Author Name: **Kadir S** (Kadir S)
Original Redmine Issue: [20046](https://issues.qgis.org/issues/20046)
Affected QGIS version: 3.3(master)
Redmine category:gui

---
While _Log Messages Panel_ is open, it disappears if I minimize and maximize QGIS. When switching another program and then returning QGIS, no problem, it stays open.

I use the panel extensively. I encounter that issue in both v2.18.22 and v3.2.1.

### LINKED ISSUE #31955: "Log Messages" shortcut is broken if "Python console" is enabled
While trying to replicate https://github.com/qgis/QGIS/issues/31914#issuecomment-534017343 I came across an issue that was previously reported but closed during 2 to 3 migration (#22127)
1. Open the "Log messages" panel. Do not anchor it in the dialog (keep it floating)
1. Open the Python Console. It will likely be added to the same tab as the Message log. So far so good
1. Go to the "Python Console". It comes foreground and you no longer see the tabs title. So no way to click on the other tab to reactivate it.
1. Try also to enable the "log messages" panel using the tooltip button on the status bar. Nothing!
If you however go the "View --> Panels -->" route then the "log messages" tab is back.

3.9 766db97 on Windows 10 but it is also the case in 3.4.12

### LINKED ISSUE #39171: Log Messages panel does not show when Temporal Controller Panel is visible in same dock
Opening messages with the little text balloon button does NOT show the messages panel when both temporal controller and log messages panel are docked in the same dock.

In a fresh profile, if you open the Log Messages panel it will show in the south dock
If you open the Temporal Controller Panel it will show in the north dock

As soon as you drag one of the two to the other dock so they are together in one dock, you will have troubles showing them together using the little 'text balloon button'.
Below shown correctly if both are fine:

Screenshot-20201004151200-670x266

With both panels hidden (and docked in the same dock), if you start with the time controller (via context menu!!) and then the log messages (via context menu!!) all is fine, and the tabbed interface is shown

BUT if you start with the time controller, and then use the text-balloon button in the lower right corner, it seems like you cannot make the log messages appear gain.

ONLY using the context menu (in an empty part of the toolbar), you can get it in view again...

As soon as you split them the button is working ok again.
So the issue is only when both are docked in the same dock

nyalldawson:
Fixes #31955, fixes #27868, fixes #39171

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -1298,8 +1298,8 @@ QgisApp::QgisApp( QSplashScreen *splash, bool restorePlugins, bool skipVersionCh
   addDockWidget( Qt::BottomDockWidgetArea, mLogDock );
   mLogDock->setWidget( mLogViewer );
   mLogDock->hide();
-  connect( mMessageButton, &QAbstractButton::toggled, mLogDock, &QWidget::setVisible );
-  connect( mLogDock, &QDockWidget::visibilityChanged, mMessageButton, &QAbstractButton::setChecked );
+  connect( mMessageButton, &QAbstractButton::toggled, mLogDock, &QgsDockWidget::setUserVisible );
+  connect( mLogDock, &QgsDockWidget::visibilityChanged, mMessageButton, &QAbstractButton::setChecked );
   connect( QgsApplication::messageLog(), static_cast < void ( QgsMessageLog::* )( bool ) >( &QgsMessageLog::messageReceived ), this, &QgisApp::toggleLogMessageIcon );
   connect( mMessageButton, &QAbstractButton::toggled, this, &QgisApp::toggleLogMessageIcon );
   mVectorLayerTools = new QgsGuiVectorLayerTools();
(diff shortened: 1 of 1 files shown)


nirvn (approved):

rduivenvoorde:
Thanks @nyalldawson! Build and tested.

One minor visual issue: the little bubble button becomes active/nonactive if you click it it self.
But if you use the contextmenu to show/hide the log messages panel, the bubble button active state is not updated...
I can live with it :-) But thought to let know.

rduivenvoorde:
Nice headtrick by the way: closing 3 issues \o/

nyalldawson:
@rduivenvoorde I can't reproduce that -- could you share a screencast?

rduivenvoorde:
More info (my screencast plugin just broke...):
it is because if you use the context menu to hide/show the Log Messages, it does NOT get focus. 
And the little balloon button is only 'active' if the actual Log Messages TAB is in focus.
Difference is that with the balloonbutton you show the messages AND it get's focus
with context menu you only add the messages to the tab list but (in my case) the Temporal Controller stays in focus
This more clear?

nyalldawson:
Which context menu are you using?

rduivenvoorde:
Right click in empty place of toolbar (to enable Panels and Toolbars)
Screenshot-20201006115212-469x1003

Issue #39036, opened 2020-09-25
erroneous postgis query when trying to discover relations


beerockxs:
When I try to discover relations between my tables, I get this warning in the PostGIS message log:

Erroneous query: WITH foreign_keys AS ( SELECT c.conname, c.conrelid, c.confrelid, unnest(c.conkey) AS conkey, unnest(c.confkey) AS confkey, (SELECT relname FROM pg_catalog.pg_class WHERE oid = c.conrelid) as referencing_table, (SELECT relname FROM pg_catalog.pg_class WHERE oid = c.confrelid) as referenced_table, (SELECT relnamespace::regnamespace::text FROM pg_catalog.pg_class WHERE oid = c.confrelid) as constraint_schema FROM pg_constraint c WHERE contype = 'f' AND c.conrelid::regclass = '.'::regclass ) SELECT fk.conname as constraint_name, a.attname as column_name, fk.constraint_schema, referenced_table as table_name, af.attname as column_name, fk.confkey as ordinal_position FROM foreign_keys fk JOIN pg_attribute af ON af.attnum = fk.confkey AND af.attrelid = fk.confrelid JOIN pg_attribute a ON a.attnum = conkey AND a.attrelid = fk.conrelid ORDER BY fk.confrelid, fk.conname ; returned 7 [ERROR: invalid name syntax
             LINE 1: ...ERE contype = 'f' AND c.conrelid::regclass = '.'::regcl...
              ^
             ]

QGIS version | 3.15.0-Master | QGIS code revision | 3119eb7de2
-- | -- | -- | --
Compiled against Qt | 5.11.2 | Running against Qt | 5.11.2
Compiled against GDAL/OGR | 3.2.0dev | Running against GDAL/OGR | 3.2.0dev
Compiled against GEOS | 3.8.1-CAPI-1.13.3 | Running against GEOS | 3.8.1-CAPI-1.13.3
Compiled against SQLite | 3.29.0 | Running against SQLite | 3.29.0
PostgreSQL Client Version | 11.5 | SpatiaLite Version | 4.3.0
QWT Version | 6.1.3 | QScintilla2 Version | 2.10.8
Compiled against PROJ | 7.2.0 | Running against PROJ | Rel. 7.2.0, November 1st, 2020
OS Version | Windows 10 (10.0) | This copy of QGIS writes debugging output.
Active python plugins | SpreadsheetLayers; db_manager; MetaSearch; processing

elpaso:
[...]

PR #39195, opened 2020-10-06
PG: Silence warning when discovering relations on query layers

elpaso:
No functional changes: just silence a warning.

Followup https://github.com/qgis/QGIS/issues/39036

### DIFF
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -4907,7 +4907,7 @@ QList<QgsRelation> QgsPostgresProvider::discoverRelations( const QgsVectorLayer 
   // Skip less silently if layer is not valid
   if ( !mValid )
   {
-    QgsLogger::warning( QStringLiteral( "Error discovering relations of %1: invalid layer" ).arg( mQuery ) );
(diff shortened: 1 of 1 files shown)


src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -4897,12 +4897,21 @@ QList<QgsVectorLayer *> QgsPostgresProvider::searchLayers( const QList<QgsVector
...
+  if ( mIsQuery || mTableName.isEmpty() || mSchemaName.isEmpty() )
+  {
+    return result;
+  }
+
+  // Skip less silently if layer is not valid
   if ( !mValid )
   {
-    QgsLogger::warning( "Error getting the foreign keys of " + mTableName + ": invalid connection" );
+    QgsLogger::warning( QStringLiteral( "Error discovering relations of %1: invalid layer" ).arg( mQuery ) );
```
nyalldawson, line 4910: Should we translate log messages like this?

elpaso:
```suggestion
    QgsLogger::warning( tr( "Error discovering relations of %1: invalid layer" ).arg( mQuery ) );
```

nyalldawson (approved):

PR #39217, opened 2020-10-07
Fix fold icon colors in code editors

nyalldawson:
Allow these to be set so that they can match the editor theme


nyalldawson:
@nirvn you'll want to update the dark themes defaults to include these

PR #39220, opened 2020-10-07
[gps] Ensure that the status bar GPS distance readout only shows in response to a mouse action

nyalldawson:
... and doesn't always re-show as a result of a GPS position change

Otherwise as soon as the message appears, it will endlessly
re-show as the GPS position is constantly changing. Instead
ensure that the message starts showing for a maximum of
2 seconds in response to a mouse movement, yet will update
during these 2 seconds if the GPS position is updated during
this time.

### DIFF
--- a/src/app/gps/qgsgpsinformationwidget.cpp
+++ b/src/app/gps/qgsgpsinformationwidget.cpp
@@ -1634,12 +1634,14 @@ void QgsGpsInformationWidget::updateGpsDistanceStatusMessage( bool forceDisplay 
   if ( !mNmea )
     return;
 
+  static constexpr int GPS_DISTANCE_MESSAGE_TIMEOUT_MS = 2000;
+
   if ( !forceDisplay )
   {
     // if we aren't forcing the display of the message (i.e. in direct response to a mouse cursor movement),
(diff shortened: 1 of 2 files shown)


src/app/gps/qgsgpsinformationwidget.cpp
```diff
@@ -1595,27 +1595,40 @@ void QgsGpsInformationWidget::cursorCoordinateChanged( const QgsPointXY &point )
...
 {
   if ( !mNmea )
     return;
 
+  if ( !forceDisplay )
+  {
+    // if we aren't forcing the display of the message (i.e. in direct response to a mouse cursor movement),
+    // then only show an updated message when the GPS position changes if the previous forced message occurred < 2 seconds ago.
+    // otherwise we end up showing infinite messages as the GPS position constantly changes...
+    if ( mLastForcedStatusUpdate.hasExpired( 2000 ) )
```
lbartoletti, line 1616: Can you create a constant with this 2000 instead of this magic number?
BTW, do you know why it's exactly 2 seconds?
Thanks.

nyalldawson: It's just the default timeout for timed status bar messages -- but yes, I'll make it a constant

PR #39224, opened 2020-10-07
[ui][code editor] Remove fold area for the sql code editor

nirvn:
## Description

@nyalldawson , as discussed, let's get rid of the fold area for the SQL editor, there's little point in having it there.


nyalldawson (approved):

nyalldawson:
Should we do that for the expression editor too?

nirvn:
@nyalldawson , done.

Issue #39080, opened 2020-09-29
GetLegendImage URL should respect the ignore GetMap value in GetCapabilities check box with WMS & WMTS requests


ianturton:
**Describe the bug**

If a WMTS reports an incorrect URL in it's capabilities document QGis allows users to ignore them but getLegend requests don't ignore the capabilities value. Until #21425 is implemented then make the legend graphic url dependent on if the getMap URL is being ignored.  

**How to Reproduce**

1. Install a local GeoServer and break the capabilities file by adding a non existent url in the proxy url under global settings.
2. Add the WMTS getCapabilities link from the GeoServer front page to QGIS, checking the ignore url for gettile and getfeatureinfo boxes
3. Install QGIS Network logger
4. add a layer from the GeoServer install
5. No legend is displayed - instead of generating a correct URL from the base of the getCapabilities request it will attempt to use the incorrect one from the capabilities document.

**QGIS and OS versions**

QGIS version
3.14.16-Pi
QGIS code revision
a235a14
Compiled against Qt
5.12.8
Running against Qt
5.12.8
Compiled against GDAL/OGR
3.0.4
Running against GDAL/OGR
3.0.4
Compiled against GEOS
3.8.0-CAPI-1.13.1
Running against GEOS
3.8.0-CAPI-1.13.1
Compiled against SQLite
3.31.1
Running against SQLite
3.31.1
PostgreSQL Client Version
12.4 (Ubuntu 12.4-0ubuntu0.20.04.1)
SpatiaLite Version
4.3.0a
QWT Version
6.1.4
QScintilla2 Version
2.11.2
Compiled against PROJ
6.3.1
Running against PROJ
Rel. 6.3.1, February 10th, 2020
OS Version
Ubuntu 20.04.1 LTS
Active python plugins
vrtbuilderplugin;
autoSaver;
Projestions;
QGISSortAndNumber-master;
QGIS3-getWKT;
qgis_hats;
MemoryLayerSaver;
isochrones;
firstaid;
OSMDownloader;
qchainage;
plugin_reloader;
Beeline;
profiletool;
polystrip;
TerrainShading;
cartogram3;
Hqgis;
pluginbuilder3;
quick_map_services;
osgb-plugin;
pointsamplingtool;
DataPlotly;
QuickWKT;
timemanager;
processing;
MetaSearch;
db_manager;
qgis_resource_sharing;
qgisnetworklogger

**Additional context**

agiudiceandrea:
[...]

PR #39231, opened 2020-10-07
Use mIgnoreGetMapUrl flag to adjust getLegendGraphic requests

ianturton:
## Description

If a WMS or WMTS reports an incorrect URL in it's capabilities document QGis allows users to ignore them but getLegend requests don't ignore the capabilities value. Until #21425 is implemented then make the legend graphic url dependent on if the getMap URL is being ignored. 

see #39080

### DIFF
--- a/src/providers/wms/qgswmsprovider.cpp
+++ b/src/providers/wms/qgswmsprovider.cpp
@@ -350,12 +350,13 @@ QString QgsWmsProvider::getLegendGraphicUrl() const
       break;
     }
   }
-
+  QgsDebugMsgLevel( QStringLiteral( "step 1-> getLegendGraphicUrl is %1" ).arg( url ), 6 );
   if ( url.isEmpty() && !mCaps.mCapabilities.capability.request.getLegendGraphic.dcpType.isEmpty() )
   {
     url = mCaps.mCapabilities.capability.request.getLegendGraphic.dcpType.front().http.get.onlineResource.xlinkHref;
   }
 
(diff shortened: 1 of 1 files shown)


nyalldawson:

elpaso (approved):
LGTM

src/providers/wms/qgswmsprovider.cpp
```diff
@@ -380,6 +381,34 @@ QString QgsWmsProvider::getLegendGraphicUrl() const
...
+  {
+    if ( url.isEmpty() )
+    {
+      QgsDebugMsgLevel( QStringLiteral( "converting getLegendGraphicUrl is %1" ).arg( url ), 9 );
+      // rewrite the URL if the one in the capabilities document is incorrect
+      // strip every thing after the ? from the base url
+      const QStringList parts = mSettings.mBaseUrl.split( QRegularExpression( "\\?" ) );
+      const QString base = parts.isEmpty() ? mSettings.mBaseUrl : parts.first();
+      // and strip everything before the `rest` element (at least for GeoServer)
+      const int index = url.length() - url.lastIndexOf( QStringLiteral( "rest" ) ) + 1; // +1 for the /
```
m-kuhn, line 394: What exactly is the "rest" element? Do we have a problem if we don't have it int he url?

m-kuhn: Actually, this is in a condition `url.isEmpty()`, won't that always be empty?

PR #39241, opened 2020-10-07
Fix layout order on drag drop action, fix #39136

### LINKED ISSUE #39136: Trouble ordering elements in layout's composer
**Describe the bug**

Ordering elements in layout doesn't give expected results.

**How to Reproduce**

1. Create a new composer
2. Create a bunch of element (type does not matter)
3. Try to order element as desired

**Try to insert an element "insert" between elements 1 and 2**
image

_Expected result :_
"insert" goes between 1 and 2
image

_Actual result :_
"insert" goes between 2 and 3
image

**Try to put "insert" on top : goes in second position**
image
image

**Try to put "insert" at the bottom : just can't, "insert" doesn't move**
image

**QGIS and OS versions**
Qgis 3.14.16-2 Windows

**Additional context**

None

SebastienPeillet:
## Description

Bug described by @nicogodet in #39136 

In layout list item, drag and drop to move item doesn't work as expected. In the [dropMimeData](https://github.com/qgis/QGIS/compare/master...SebastienPeillet:fix_layout_order_on_drag-drop_action?expand=1#diff-66138644cf02582ee3590cf5657517a7R308) function, the index use to insert the moved item is shifted by one.

Futhermore when user drops item on last line, input `column` variable is initiated on -1, so column is not equal to `ItemId` [here](https://github.com/qgis/QGIS/compare/master...SebastienPeillet:fix_layout_order_on_drag-drop_action?expand=1#diff-66138644cf02582ee3590cf5657517a7R311). In consequence the item is not moved at the bottom of the list.

I added a new test in [testqgslayoutmodel.cpp](https://github.com/qgis/QGIS/compare/master...SebastienPeillet:fix_layout_order_on_drag-drop_action?expand=1#diff-ac42da3851bd88ac4291deaccc824a05R565) to check that the dropMimeData works as expected

### DIFF
--- a/src/core/layout/qgslayoutmodel.cpp
+++ b/src/core/layout/qgslayoutmodel.cpp
@@ -308,7 +308,7 @@ bool zOrderDescending( QgsLayoutItem *item1, QgsLayoutItem *item2 )
 bool QgsLayoutModel::dropMimeData( const QMimeData *data,
                                    Qt::DropAction action, int row, int column, const QModelIndex &parent )
 {
-  if ( column != ItemId )
+  if ( column != ItemId & column != -1 )
   {
     return false;
   }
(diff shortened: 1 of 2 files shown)


nyalldawson (commented):
Good fix -- thanks!

nyalldawson:
This also fixes a crash I've seen while reordering items

PR #39242, opened 2020-10-07
Make Snap to common angles disabled by default

### LINKED ISSUE #39166: qgis snapping feature is flawed if  "Snapping to Common Angles" is ON
**Describe the bug**

The snapping is on for a layer and the snapping markers (pink square marker for vertex and pink hourglass marker for segment) are shown when digitizing. However, dangles get created which means the snapping does not function right. Increasing the Tolerance in "Advanced Config Dialog" does not help

**How to Reproduce**
create a simple linear layer in a Geopackage and try to digitize it. Try to snap to endpoints, vertices, and segments. At the same time with "Topology Checker" tool, define a rule "must not have dangles" on the same layer and check the topology of the digitized lines.

image

image

**QGIS and OS versions**
GIS version | 3.10.9-A Coruña on Debian unstable (sid)

**Additional context**

uclaros:
## Description
*Snap to common angles* in *Advanced digitizing* was set to right angles (90deg) by default.
This feature should be disabled by default and enabled at will only when needed.

Fixes #39166


lbartoletti (approved):
+1

Issue #39178, opened 2020-10-05
setting a field to NULL via field calculator freezes qgis when "Automatic Creation of Transaction Group" is enabled


faridcher:
**How to Reproduce**
Try to set the tsize field of labs layer to NULL via field calculator .

related:
https://issues.qgis.org/issues/11422
[subdiv.zip](https://github.com/qgis/QGIS/files/5330575/subdiv.zip)

gioman:
@faridcher cannot confirm on 3.10.10 on Ubuntu 20.04-

faridcher:
upgraded to 3.10.10 and the problem persists. Ubuntu is Debian derivative so I don't think it's OS specific. I can record a video if that helps.

faridcher:
upgraded to 3.10.10 and the problem persists. Please try with attached gpkg file in the original post. Try to set the tsize field of labs label to NULL.

thanks

elpaso:
No issues here on QGIS master/linux

gioman:
> upgraded to 3.10.10 and the problem persists. Please try with attached gpkg file in the original post. Try to set the tsize field of labs layer to NULL.
> 
> thanks

@faridcher try with a new QGIS profile.

faridcher:
A new profile did not help. However, the problem relates to the qgis project file and it occurs when "Automatic Creation of Transaction Group" is enabled in the "Project Properties > Data Sources". Un-checking that option solves it.

elpaso:
This is indeed a big bad voracious :bug:

gioman:
> This is indeed a big bad voracious bug

@elpaso 
same origin as https://github.com/qgis/QGIS/issues/39177 ?

elpaso:
Definitely related.

elpaso:
I now know the source of the bug but the fix isn't trivial.

elpaso:
See also: https://github.com/qgis/QGIS/pull/39244

elpaso:
Reproducing test case:

```python
    def testTransactionGroupIterator(self):
        """Test issue GH #39178: the bug is that this test hangs
        forever in an endless loop"""

        project = QgsProject()
        project.setAutoTransaction(True)
        tmpfile = os.path.join(
            self.basetestpath, 'tempGeoPackageTransactionGroupIterator.gpkg')
        ds = ogr.GetDriverByName('GPKG').CreateDataSource(tmpfile)
        lyr = ds.CreateLayer('test', geom_type=ogr.wkbPoint)
[...]

PR #39244, opened 2020-10-07
[OGR provider] Remove calls to ResetReading() in changeGeometryValues() and changeGeometryValues() in situations where this is safe to do

rouault:
Such calls will cause issues in the context of transaction groups where
feature iterators will share the same connection handle.

CC @elpaso

### DIFF
--- a/src/core/providers/ogr/qgsogrprovider.cpp
+++ b/src/core/providers/ogr/qgsogrprovider.cpp
@@ -2306,6 +2306,22 @@ bool QgsOgrProvider::changeAttributeValues( const QgsChangedAttributesMap &attr_
 
   const bool inTransaction = startTransaction();
 
+  // Some drivers may need to call ResetReading() after GetFeature(), such
+  // as GPKG in GDAL < 2.3.0 to avoid letting the database in a locked state.
+  // But this is undesirable in general, so don't do this when we know that
+  // we don't need to.
+  bool mayNeedResetReadingAfterGetFeature = true;
+  if ( mGDALDriverName == QLatin1String( "ESRI Shapefile" ) )
+  {
+    mayNeedResetReadingAfterGetFeature = false;
+  }
+#if GDAL_VERSION_NUM >= GDAL_COMPUTE_VERSION(2,3,0)
+  else if ( mGDALDriverName == QLatin1String( "GPKG" ) )
+  {
+    mayNeedResetReadingAfterGetFeature = false;
+  }
+#endif
+
   for ( QgsChangedAttributesMap::const_iterator it = attr_map.begin(); it != attr_map.end(); ++it )
   {
     QgsFeatureId fid = it.key();
@@ -2320,7 +2336,11 @@ bool QgsOgrProvider::changeAttributeValues( const QgsChangedAttributesMap &attr_
       pushError( tr( "Feature %1 for attribute update not found." ).arg( fid ) );
       continue;
     }
-    mOgrLayer->ResetReading(); // needed for SQLite-based to clear iterator
+
+    if ( mayNeedResetReadingAfterGetFeature )
+    {
+      mOgrLayer->ResetReading();
+    }
 
     QgsLocaleNumC l;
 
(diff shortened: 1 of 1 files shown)


src/core/providers/ogr/qgsogrprovider.cpp
```diff
@@ -2489,6 +2509,22 @@ bool QgsOgrProvider::changeGeometryValues( const QgsGeometryMap &geometry_map )
 
   const bool inTransaction = startTransaction();
 
+  // Some drivers may need to call ResetReading() after GetFeature(), such
+  // as GPKG in GDAL < 2.3.0 to avoid letting the database in a locked state.
```
nyalldawson, line 2513: Maybe move this duplicate code out to a function?

rouault: yeah, waiting for @elpaso confirmation that it actually helps for the issue we discussed by email

elpaso: Thanks, I'll try that tomorrow and it will certainly help but I'm 100% sure that's not enough: QgsOgrFeatureIterator constructor is calling reset in several places through OGRGeoPackageTableLayer::SetSpatialFilter, OGRGeoPackageTableLayer::SetSpatialFilterRect and OGRGeoPackageTableLayer::SetAttributeFilter they all call ResetReading and these methods are invoked in the QgsOgrFeatureIterator constructor.

elpaso: ... and the initial rewind call of course.

nyalldawson: Should we disable geopackage transaction support until this is sorted out?

elpaso: @nyalldawson yes, if we can't find a solution before the release.

elpaso: @rouault I can confirm that this patch does **not** fix the issue. The `QgsOgrFeatureIterator` ctor needs patches too, I'll try to work on it.

rouault: @elpaso For my information, could you explain or reexplain me what is the issue with QgsOgrFeatureIterator issuing ResetReading() at the beginning of the iteration ? Of course, it is problematic if ResetReading() is called after iteration has started, but not at its beginning. I probably miss something.

In the scenario you pointed in https://github.com/qgis/QGIS/issues/39178#issuecomment-705456951, if changeAttributeValue() no longer issues a ResetReading(), we should be safe, shouldn't we ?
```
        self.assertTrue(vl.startEditing())

        for f in vl.getFeatures():
            self.assertTrue(vl.changeAttributeValue(1, 1, 'new value'))
```

elpaso: @rouault no, the fact is that the last line of code here creates a new (nested) iterator
```c++
QgsVectorLayerUndoPassthroughCommandChangeAttribute::QgsVectorLayerUndoPassthroughCommandChangeAttribute( QgsVectorLayerEditBuffer *buffer, QgsFeatureId fid, int field, const QVariant &newValue )
  : QgsVectorLayerUndoPassthroughCommand( buffer, QObject::tr( "change attribute value" ) )
  , mFid( fid )
  , mField( field )
  , mNewValue( newValue )
  , mOldValue( mBuffer->L->getFeature( mFid ).attribute( field ) )
```

A commit worth more than a thousand words, this fixes the issue in the test:  https://github.com/qgis/QGIS/commit/505ef99328f8d1ba63919aeb89d815a6c4ce679b

elpaso: @rouault this PR fixes the issue https://github.com/qgis/QGIS/pull/39257 but it's just an ugly patch, I know it.

rouault: > no, the fact is that the last line of code here creates a new (nested) iterator

That was the part I was missing, thanks.

elpaso:
Superceeded by https://github.com/qgis/QGIS/pull/39257

PR #39266, opened 2020-10-08
Update qgsgeometry.h

tschmetzer:
This change is because in the pygis documentation it is impossible to see and distinguish what is a input parameter and what is a return value.

Please bear in mind that the file https://github.com/qgis/QGIS/blob/master/python/core/auto_generated/geometry/qgsgeometry.sip.in seems to have been edited manually (although there is a clear statement saying that shouldn't be done) as the return parameter of the %Docstring in lines 733/734 is different to the original version of qgsgeometry.h

From a pyGIS perspective even better than my already made proposal in the code for returning parameters would be to use \returns instead of \param if that is technically possible (I am no C++ programmer), e.g. for closestSegmentWithContext() :
     * \returns minDistPoint Returns the closest point on the segment with regards to the input point. The closest 
     * point is the perpendicular point if an end of the segment is not the shortest distance
     * \returns index_after_closest_segment Returns the index of the vertex after the closest segment. The vertex
     * before the closest segment is always afterVertex - 1
     * \returns left_or_right_of_line Returns if the point is located on the left or right side of the geometry ( < 0 means left, > 0 means right, 0 indicates
     * that the test was unsuccessful, e.g. for a point exactly on the line)
     * \returns The squared Cartesian distance (also returned in sqrDist) of the input point to the minDistPoint, negative number on error

## Description

### DIFF
--- a/src/core/geometry/qgsgeometry.h
+++ b/src/core/geometry/qgsgeometry.h
@@ -639,7 +639,7 @@ class CORE_EXPORT QgsGeometry
      * \param Returns the square distance (sqrDist) between the closest vertex and the specified point
      * \returns the closest point in geometry. If not found (empty geometry), returns null point and sqrDist is negative.
      */
-    QgsPointXY closestVertex( const QgsPointXY &point, int &atVertex SIP_OUT, int &index_before_closest_vertex SIP_OUT, int &index_after_closest_vertex SIP_OUT, double &sqrDist SIP_OUT ) const;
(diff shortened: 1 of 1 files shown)


nyalldawson:
> From a pyGIS perspective even better than my already made proposal in the code for returning parameters would be to use \returns

This should automatically be handled by the "sipify" scripts, which is responsible for translating the c++ headers and doxygen documentation into Python doc strings.

I think the underlying issue here is that sipify only currently handles a single "out" style parameter, so doesn't play nice with these particular methods.

Unfortunately sipify is a perl script. Have you any experience with perl?

nyalldawson:
Superseded by https://github.com/qgis/QGIS/pull/39269

PR #39267, opened 2020-10-08
Bugfix gh39230 expression fields ogr add features

### LINKED ISSUE #39230: Layer X: Feature has too many attributes (expecting 5, received 7) when "Automatic Creation of Transaction Group" is enabled
This error has the same origin as #39177 and #39178 .  It happens when I start digitizing a layer with a derived field.

image

qgis 3.10.10 on debian unstable

[test.zip](https://github.com/qgis/QGIS/files/5348327/test.zip)

elpaso:
Fixes #39230 and an unreported crash

### DIFF
--- a/src/core/providers/ogr/qgsogrprovider.cpp
+++ b/src/core/providers/ogr/qgsogrprovider.cpp
@@ -1658,7 +1658,7 @@ bool QgsOgrProvider::addFeaturePrivate( QgsFeature &f, Flags flags )
   }
 
   QgsAttributes attributes = f.attributes();
-  const QgsFields &qgisFields { f.fields() };
+  const QgsFields qgisFields { f.fields() };
 
   QgsLocaleNumC l;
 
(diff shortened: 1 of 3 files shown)


src/core/providers/ogr/qgsogrprovider.cpp
```diff
@@ -1657,15 +1657,16 @@ bool QgsOgrProvider::addFeaturePrivate( QgsFeature &f, Flags flags )
     }
   }
 
-  QgsAttributes attrs = f.attributes();
+  QgsAttributes attributes = f.attributes();
+  const QgsFields &qgisFields { f.fields() };
```
nyalldawson, line 1661:
```suggestion
  const QgsFields qgisFields { f.fields() };
```

no need for a reference here -- QgsFields are implicitly shared, so copies are inexpensive.

nyalldawson (approved):

PR #39284, opened 2020-10-09
add default UTF-8 encoding to grass provider by default

### LINKED ISSUE #37508: Exporting GRASS vector (save as...) crashes QGIS
QGIS 3.10/14/master on Windows and Linux:

Add a GRASS vector (i.e. from the GRASS sample dataset) to the project and try to export it via "save as...", QGIS crashes.

PeterPetrik:
fixes https://github.com/qgis/QGIS/issues/37508

it fixes the crash, where the text encoding is never set in the grass provider and therefore neither in qgsgrassfeatureiterator .

I checked the grass (gis.h) headers and i do not see a way to get a source encoding and I am not sure either the UTF-8 as default is good (still it is used for exact same reason in OGR provider..)


nyalldawson:
Let's try... in the worst case scenario bad encoding is ultimately better then a crash!

PR #39286, opened 2020-10-09
Transaction group: keep editing status after save

### LINKED ISSUE #39282: Transaction groups: not-current layers a set to not editable after save
1. create a project with transaction group option switched on
2. add two layers A and B (GPKG, PG or other supporting transactions) from the same connection
3. switch editing on on one layer (the other is set editable as well, so far so good)
4. make layer A the current one by clicking on its legend entry
5. digitize something on A and accept the form
6. click on save button
7. layer A is still editable, layer B is not!

elpaso:
Fixes #39282


m-kuhn (approved):
Thanks

PR #39291, opened 2020-10-09
GUI: renaming GPKG layer: allow any character but | (fixes #39288)

### LINKED ISSUE #39288: [Browser] Renaming only accepts characters and behaviour issue with unaccepted character in name
**Describe the bug**
When renaming a geopackage layer, using the rename function only allows the user to provide characters, not symbols (other than _ ) or numbers. 

If the name of the layer already contain number or special characters those cannot be interacted with other than by deleting part of the text containing them in one action.

**How to Reproduce**

In the browser with a geopackage containing layers.
1 Right click any layer
2. Rename Layer ...
3. Try to input numbers

**QGIS and OS versions**

Win 10, QGIS 3.14.16

**Additional context**

Untested for any other providers, assuming same constrains apply.

rouault:


roya0045:
@rouault Thank you Even, I did not date to propose a fix as I wasn't sure which characters were not acceptable in sqlite/geopackage names.

nyalldawson (approved):

nyalldawson:

nyalldawson:
Oops wrong tab

PR #39292, opened 2020-10-09
Use QLatin1String instead of QStringLiteral when possible

rouault:
- Added  scripts/qstringfixup.sh and .py (manually run for now)
- QStringLiteral -> QLatin1String changes generated by scripts/qstringfixup.sh --all

### DIFF
--- a/src/app/browser/qgsinbuiltdataitemproviders.cpp
+++ b/src/app/browser/qgsinbuiltdataitemproviders.cpp
@@ -189,7 +189,7 @@ void QgsAppDirectoryItemGuiProvider::populateContextMenu( QgsDataItem *item, QMe
       s.setValue( QStringLiteral( "/browser/hiddenPaths" ), pathsList );
 
       // get parent path and refresh corresponding node
-      int idx = path.lastIndexOf( QStringLiteral( "/" ) );
+      int idx = path.lastIndexOf( QLatin1Char( '/' ) );
       if ( idx != -1 && path.count( QStringLiteral( "/" ) ) > 1 )
       {
         QString parentPath = path.left( idx );
@@ -551,9 +551,9 @@ void QgsLayerItemGuiProvider::populateContextMenu( QgsDataItem *item, QMenu *men
     if ( layerItem )
     {
       // Also check for postgres layers (rasters are handled by GDAL)
-      isFile = ( layerItem->providerKey() == QStringLiteral( "ogr" ) ||
-                 layerItem->providerKey() == QStringLiteral( "gdal" ) ) &&
-               ! layerItem->uri().startsWith( QStringLiteral( "PG:" ) );
+      isFile = ( layerItem->providerKey() == QLatin1String( "ogr" ) ||
+                 layerItem->providerKey() == QLatin1String( "gdal" ) ) &&
+               ! layerItem->uri().startsWith( QLatin1String( "PG:" ) );
     }
     else
     {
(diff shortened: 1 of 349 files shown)


src/app/browser/qgsinbuiltdataitemproviders.cpp
```diff
@@ -552,7 +552,7 @@ void QgsLayerItemGuiProvider::populateContextMenu( QgsDataItem *item, QMenu *men
     {
       // Also check for postgres layers (rasters are handled by GDAL)
       isFile = ( layerItem->providerKey() == QStringLiteral( "ogr" ) ||
```
nyalldawson, line 554: Your script missed this line somehow

src/app/browser/qgsinbuiltdataitemproviders.cpp
```diff
@@ -552,7 +552,7 @@ void QgsLayerItemGuiProvider::populateContextMenu( QgsDataItem *item, QMenu *men
...
-                 layerItem->providerKey() == QStringLiteral( "gdal" ) ) &&
+                 layerItem->providerKey() == QLatin1String( "gdal" ) ) &&
                ! layerItem->uri().startsWith( QStringLiteral( "PG:" ) );
```
nyalldawson, line 556: startsWith is another candidate you could catch too

src/app/qgshandlebadlayers.cpp
```diff
@@ -127,7 +127,7 @@ QgsHandleBadLayers::QgsHandleBadLayers( const QList<QDomNode> &layers )
     QString datasource = node.namedItem( QStringLiteral( "datasource" ) ).toElement().text();
     QString provider = node.namedItem( QStringLiteral( "provider" ) ).toElement().text();
     QString vectorProvider = type == QLatin1String( "vector" ) ? provider : tr( "none" );
-    bool providerFileBased = ( provider == QStringLiteral( "gdal" ) || provider == QStringLiteral( "ogr" ) || provider == QStringLiteral( "mdal" ) );
+    bool providerFileBased = ( provider == QStringLiteral( "gdal" ) || provider == QStringLiteral( "ogr" ) || provider == QLatin1String( "mdal" ) );
```
nyalldawson, line 130: Here's some more missed opportunities

nyalldawson (approved):

src/app/browser/qgsinbuiltdataitemproviders.cpp
rouault: fixed, as well as your other remarks

rouault:
@nyalldawson Since you reviewed it, I've also extended it to deal with != QStringLiteral() and .contains(QStringLiteral()) and .indexOf(QStringLiteral()). All have QLatin1String override, so I assume this is beneficial to replace them too

nyalldawson:
Was there any false positives here? Looks like a great candidate for including in the prepare commit script!

rouault:
> Was there any false positives here?

yes, there were unfortunately 2 instances of false positive (caught by adhoc `if m and 'qgetenv' not in line and 'h.first' not in line:`) for the == QStringLiteral("foo") or != QStringLiteral("foo") where the left-hand side was a QByteArray. If comparing a QByteArray with a QLatin1String, this results in an ambiguous situation for the compiler. For those very few situations, we could possibly require people to make it obvious by adding some QByteArray mention in the line, like ``/*QByteArray*/(x) == QStringLiteral("foo")`` and an enhanced script would skip those situations

But I'm realizing that https://github.com/KDE/clazy would probably a safer approach (I've never used it though), although it would require integration in the compilation.

rouault:
I've added logic to deal with QString::replace() as well. Run the script over tests/ as well
And I've added the script under scripts/ . Must be manually run for now

rouault:
Regarding QLatin1Char('x') vs 'x', yes I've hesitated. But if one day we would want to build with QT_NO_CAST_FROM_ASCII enabled, that would save fixing those places. But up to you, I can change to plain 'x'

nyalldawson (approved):

PR #39305, opened 2020-10-11
Enable -DQT_USE_QSTRINGBUILDER and fix related build errors

rouault:
Background:
https://www.qt.io/blog/2011/06/13/string-concatenation-with-qstringbuilder

When enabling -DQT_USE_QSTRINGBUILDER, QString() concatenation with operator +
are done in a more efficient way, especially with repeated + operators.

Qt and creator themselves are built with that enabled.


src/providers/postgres/qgspostgresprovider.cpp
```diff
@@ -1214,15 +1214,15 @@ bool QgsPostgresProvider::loadFields()
     {
       const QString seqName { mTableName + '_' + fieldName + QStringLiteral( "_seq" ) };
       const QString seqSql { QStringLiteral( "SELECT c.oid "
```
nyalldawson, line 1216: Let's drop the initialiser brace use here -- it's a constant source of issues with astyle

nyalldawson (approved):

nyalldawson:
Great change! Thanks!

src/providers/postgres/qgspostgresprovider.cpp
rouault: done, but that doesn't change astyle output...

nyalldawson (approved):

PR #39307, opened 2020-10-11
level of detail for 3D mesh

vcloarec:
This PR allows the user, if mesh simplification is enabled, to choose a level of details for 3D rendering.
It is quite convenient when the mesh contains millions of triangles, slowing down the computer,  and rendering navigation in 3D view very difficult. 

lod


src/3d/symbols/qgsmesh3dsymbol.h
```diff
@@ -326,6 +326,22 @@ class _3D_EXPORT QgsMesh3DSymbol : public QgsAbstract3DSymbol
      */
     void setArrowsFixedSize( bool arrowsFixedSize );
 
+    /**
+     * Returns the level of detail index for rendering the mesh
+     * \see QgsMeshSimplificationSettings
```
nyalldawson, line 331: Some explanation of what the returned value means would be useful here

src/app/3d/qgsmesh3dsymbolwidget.cpp
```diff
@@ -111,6 +115,23 @@ void QgsMesh3dSymbolWidget::setLayer( QgsMeshLayer *meshLayer, bool updateSymbol
...
 
+  mComboBoxTriangleCount->clear();
+  if ( meshLayer && meshLayer->meshSimplificationSettings().isEnabled() )
+  {
+    mComboBoxTriangleCount->setVisible( true );
+    mLabelTriangleCount->setVisible( true );
+    int lodCount = meshLayer->triangularMeshLevelOfDetailCount();
+    for ( int i = 0; i < lodCount; ++i )
+    {
+      mComboBoxTriangleCount->addItem( QString::number( meshLayer->triangularMeshByLodIndex( i )->triangles().count() ) );
```
nyalldawson, line 126: To make this a bit clearer for users, I'd suggest that the combo box strings include the level too. E.g. instead of just the raw count "1234134", maybe something like

    Level 1 (123342 triangles)
    Level 2 (5453 triangles)

etc

Or, alternatively, if the level of detail count is calculated by just taking a fixed % of triangles, then expose that here:

   "100% (234234 triangles)"
   "75% (123123 triangles)"

etc

nyalldawson (commented):
Looks good! Got any unit tests for this too?

src/app/3d/qgsmesh3dsymbolwidget.cpp
vcloarec: this:
combo
?

nyalldawson: Whatt's the 1/10, 1/112  etc represent?

vcloarec:
tests fail because need https://github.com/qgis/QGIS/pull/39319

src/app/3d/qgsmesh3dsymbolwidget.cpp
vcloarec: the reduction factor from the original mesh. Percentage is not pertinent because the reduction is too strong.

nyalldawson: So 1/10 = 10% of the triangles?

vcloarec: > So 1/10 = 10% of the triangles?

yes, that's it

nyalldawson: Hmm that looks a bit odd to me still. @nirvn have you any ideas here?

vcloarec: IMHO, just "XXX triangles" is enough...

nirvn: I'm also not a fan of the combobox with XXXX triangles.

What about using a slider instead? I feel that could do a better UX job. 

Each step in the slider would represent 1/10, 1/137, etc. Users would slide left/right to obtain right level of simplification.

It'd behave very much like the shapebuster fill symbol layer's quality slider.

vcloarec: slider gif

nirvn: Much better IMHO.

nyalldawson:
@vcloarec can you fix the conflicts here and then merge?

vcloarec:
conflict solved and some change made considering recent change.
But I don't have this wonderful power to merge...

PeterPetrik:
I think the need of this patch as such is because we do not have tiled rendering implemented for mesh layers. Problem is that with mesh layer we load all nodes at the same time all the times. Once the tiled/chunked 3D renderer is implemented, similarly to vector 3d renderer, we wouldn't need such setting of level of details (@wonder-sk ?)

nyalldawson:
Let's merge for now -- we can always remove the setting if it proves redundant if/when a chunked loader is made for mesh layers.

Issue #39296, opened 2020-10-10
Digitizing broken when reprojecting from EPSG:32648 (project CRS) to EPSG:3148 (layer CRS)


nirvn:
When digitizing a feature on a layer (CRS EPGS:3148) within a project having a CRS set to EPGS:32648, the resulting feature vertices are a/ no located where the clicks occured on the canvas (i.e. coordinates are misreprojected) and more worryingly b/ the whole feature goes 'corrupt', extra vertices are added/modified as you digitize additional vertices.

GIF time:
![UdtnpP4FXQ](https://user-images.githubusercontent.com/1728657/95645842-1e7d7f00-0aed-11eb-8987-c7e219d9b246.gif)

Test project file:
[serious_projection_bug.zip](https://github.com/qgis/QGIS/files/5358461/serious_projection_bug.zip)

Steps to reproduce:
1. Open the above-attached test project file
2. Toggle the edit mode of the project layer 
3. Add a new polygon, see it fail live

gioman:
@nirvn which QGIS version? seems ok on 3.10.

nirvn:
Issue is on master (I assume it's a proj >= 6 issue)

gioman:
> Issue is on master

confirmed.

nirvn:
Fixed in master.

PR #39308, opened 2020-10-11
Fix scenarios under which digitizing on layer with a CRS different from the project fails

nirvn:
## Description

Partial fix for #39296. QgsMapSettings' layerToMapCoordinates( QgsVectorLayer, QgsPoint ) function wrongly assumed the provided QgsPoint always contained a valid z() value. 

@nyalldawson , review appreciated. An alternative fix would be to update the QgsCoordinateTransform transformInPlace() function to handle nan z values (by essentially skipping the z-related value checks).

### DIFF
--- a/src/core/qgsmapsettings.cpp
+++ b/src/core/qgsmapsettings.cpp
@@ -500,7 +500,7 @@ QgsPoint QgsMapSettings::layerToMapCoordinates( const QgsMapLayer *layer, const 
 {
   double x = point.x();
   double y = point.y();
-  double z = !std::isnan( point.z() ) ? point.z() : 0.0;
+  double z = point.z();
 
   try
   {
@@ -513,7 +513,7 @@ QgsPoint QgsMapSettings::layerToMapCoordinates( const QgsMapLayer *layer, const 
     QgsMessageLog::logMessage( QObject::tr( "Transform error caught: %1" ).arg( cse.what() ), QObject::tr( "CRS" ) );
   }
 
-  return QgsPoint( x, y, !std::isnan( point.z() ) ? z : point.z() );
+  return QgsPoint( x, y, z );
 }
 
 
@@ -554,7 +554,7 @@ QgsPoint QgsMapSettings::mapToLayerCoordinates( const QgsMapLayer *layer, const 
 {
   double x = point.x();
   double y = point.y();
-  double z = !std::isnan( point.z() ) ? point.z() : 0.0;
+  double z = point.z();
 
   try
   {
@@ -567,7 +567,7 @@ QgsPoint QgsMapSettings::mapToLayerCoordinates( const QgsMapLayer *layer, const 
     QgsMessageLog::logMessage( QObject::tr( "Transform error caught: %1" ).arg( cse.what() ), QObject::tr( "CRS" ) );
   }
 
-  return QgsPoint( x, y, !std::isnan( point.z() ) ? z : point.z() );
+  return QgsPoint( x, y, z );
 }
 
 
--- a/src/core/qgsmapsettings.cpp
+++ b/src/core/qgsmapsettings.cpp
@@ -496,11 +496,11 @@ QgsPointXY QgsMapSettings::layerToMapCoordinates( const QgsMapLayer *layer, QgsP
   return point;
 }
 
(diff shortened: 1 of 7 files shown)


nirvn:
This also fixes vertex symbols dislocated when using the node editor. 

Before:
image

PR:
image

nirvn:
Alright, the last commit completes the fix for #39296. Long story short, capturing reprojected lines or polygons was adding undesired vertices very close to the ones added via left mouse click due to the "round-trip" projection transformation occurring when adding a vertex (projecting from map to layer CRS), followed by the creation of a new segment which used the first and last vertex of the layer CRS geometry back into map CRS.

By the time the coordinates were transformed back into map CRS, a slight decimal change occurs, which meant that when adding an extra curve segment back to the layer CRS geometry, we were adding extra vertices.

The fix is to avoid round-trip transformation by storing first and last captured vertices as original map points.

src/core/qgsmapsettings.cpp
```diff
@@ -500,7 +500,7 @@ QgsPoint QgsMapSettings::layerToMapCoordinates( const QgsMapLayer *layer, QgsPoi
 {
```
vcloarec, line 500: There is surely a good reason, but why in these methods the `point` argument is not a const reference?
I'd seen that in a recent past, and I wondered about.

nyalldawson: It should be fixed. In the past QgsPoint was a small and trivially copyable value, but it's not anymore.

nirvn: Done.

src/core/qgsmapsettings.cpp
```diff
@@ -496,11 +496,11 @@ QgsPointXY QgsMapSettings::layerToMapCoordinates( const QgsMapLayer *layer, QgsP
   return point;
 }
 
-QgsPoint QgsMapSettings::layerToMapCoordinates( const QgsMapLayer *layer, QgsPoint point ) const
+QgsPoint QgsMapSettings::layerToMapCoordinates( const QgsMapLayer *layer, const QgsPoint &point ) const
 {
   double x = point.x();
   double y = point.y();
-  double z = point.z();
+  double z = !std::isnan( point.z() ) ? point.z() : 0.0;
```
nyalldawson, line 503: I think we might need to handle this at a lower level -- in QgsCoordinateTransform::transformCoords itself. Otherwise we're likely to have other circumstances where a QgsPoint is transformed which will exhibit similar issues.

What we'd need to do is change these tests:

```
    // proj_errno is sometimes not an accurate method to test for transform failures - so we need to
    // manually scan for nan values
    if ( std::any_of( x, x + numPoints, []( double v ) { return std::isinf( v ); } )
    || std::any_of( y, y + numPoints, []( double v ) { return std::isinf( v ); } )
    || std::any_of( z, z + numPoints, []( double v ) { return std::isinf( v ); } ) )
    {
      actualRes = 1;
    }
```

To also test whether any output nan values are NOT nan values in the xprev/yprev/zprev arrays. That should fix the ultimate underlying issue causing this behavior.

nirvn: Right, I suspected as much.

nirvn:
@nyalldawson , done.

tests/src/core/testqgscoordinatetransform.cpp
```diff
@@ -486,6 +487,22 @@ void TestQgsCoordinateTransform::transformContextNormalize()
...
+  QVERIFY( ct.isValid() );
+  QgsPoint pt( 0.0, 0.0 );
+  double x = pt.x();
+  double y = pt.y();
+  double z = pt.z();
+  ct.transformInPlace( x, y, z );
+
+  QGSCOMPARENEAR( x, 0.0, 0.01 );
+  QGSCOMPARENEAR( y, 0.0, 0.01 );
+  QGSCOMPARENEAR( z, pt.z(), 0.01 );
```
nyalldawson, line 504: not sure that QGSCOMPARENEAR will work with nan values -- better to test direct with std::isnan

nyalldawson (approved):

nyalldawson:
@nirvn In your PR description you say "partial fix for ...". I gather that's out of date now and it's a complete fix?

nirvn:
@nyalldawson , correct, I'll close the issue when this is merged.

PR #39319, opened 2020-10-12
fix 3d mesh dataset rendering

vcloarec:
Issue in 3d mesh dataset rendering...


nyalldawson:
Do we need some more test coverage here?

vcloarec:
> Do we need some more test coverage here?

Yes, sure, it will be covered by tests added in  https://github.com/qgis/QGIS/pull/39307 that need this fix to not fail.

PeterPetrik:
i think this need a backport...?

vcloarec:
Actually it was already cover by a test, but with a dataset index (group = 0, index = 0). As the issue came from that group and index was inverted...
So the test is now fixed.


> i think this need a backport...?

It is a feature introduce in 3.14 release and the issue was introduced when implementing the new temporal framework in the same release.

Issue #28853, opened 2019-01-19
QgsAction "enabledOnlyWhenEditable" not available in Python


qgib:
Author Name: **Benjamin Jakimow** (@jakimowb)
Original Redmine Issue: [21034](https://issues.qgis.org/issues/21034)
Affected QGIS version: 3.4.3
Redmine category:python_bindings_/_sipify

---
## Problem

The @QgsAction@ keyword @enabledOnlyWhenEditable@ is unavailable Python, although
the keyword is documented in both, the C++ (https://qgis.org/api/classQgsAction.html) and Python documentation (https://qgis.org/pyqgis/master/core/QgsAction.html).

the C++ and Python documentation  describe it

## Example

```
from qgis.core import QgsAction
QgsAction(QgsAction.GenericPython, 'dummy', 'print("dummy")', 'dummy', True, notificationMessage='dummy')

1. fails
QgsAction(QgsAction.GenericPython, 'dummy', 'print("dummy")', 'dummy', True, notificationMessage='dummy', enabledOnlyWhenEditable=False)

```

## Systems

Windows 10 Education, 64-Bit
macOS 10.13.6

QGIS version 3.4.3-Madeira
QGIS code revision 2f64a3c4e7

stev-0:
Looks like [this commit](https://github.com/qgis/QGIS/commit/555cdf0a20d926fde8708a2f70600cfcf5fccc35) changed that - unclear why it was done and therefore whether the code or just documentation should be changed.

@signedav might know as the author of the commit?

jakimowb:
Just checked it with QGIS 3.12.2, but "enabledOnlyWhenEditable" is not available. And I still would like to have it in PyQGIS.

```
from qgis.core import QgsAction
QgsAction(QgsAction.GenericPython, 'dummy', 'print("dummy")', 'dummy', True, notificationMessage='dummy')

# fails
QgsAction(QgsAction.GenericPython, 'dummy', 'print("dummy")', 'dummy', True, notificationMessage='dummy', enabledOnlyWhenEditable=False)
```

QGIS Version

QGIS version | 3.12.2-București | QGIS code revision | 8a1fb33634
-- | -- | -- | --
Compiled against Qt | 5.11.2 | Running against Qt | 5.11.2
Compiled against GDAL/OGR | 3.0.4 | Running against GDAL/OGR | 3.0.4
[...]

PR #39325, opened 2020-10-12
Add note about Python bindings in QgsAction

Gustry:
## Description

From : https://qgis.org/pyqgis/3.10/core/QgsAction.html

```
QgsAction(type: QgsAction.ActionType, description: str, command: str, capture: bool = False) Create a new QgsAction

Parameters

        type – The type of this action

        description – A human readable description string

        command – The action text. Its interpretation depends on the type

        capture – If this is set to True, the output will be captured when an action is run

        enabledOnlyWhenEditable – if True then action is only enable in editmode

```

The constructor is missing this parameter.

@signedav Why is it not possible to create an action enabled only when the layer is editable? 
To overpass this bindings, we need to use a QML and load it to have this kind of action.

### DIFF
--- a/python/core/auto_generated/qgsaction.sip.in
+++ b/python/core/auto_generated/qgsaction.sip.in
@@ -42,7 +42,7 @@ Create a new QgsAction
 :param description: A human readable description string
 :param command: The action text. Its interpretation depends on the type
 :param capture: If this is set to ``True``, the output will be captured when an action is run
-:param enabledOnlyWhenEditable: if ``True`` then action is only enable in editmode
+:param enabledOnlyWhenEditable: if ``True`` then action is only enable in editmode. Not available in Python bindings.
 %End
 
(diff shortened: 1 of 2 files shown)


signedav (approved):

signedav:
Hi @Gustry 

Like I mentioned here #28853 it's just not implemented.
Would make sense to make it possible.
Meanwhile thanks for updating the comments.

Gustry:
Thanks @signedav for the review and the ticket.

It's not clear to me why there is a special constructor in Python ?
If the different constructor is really needed, can I add at least a

```
QgsAction.setEnabledOnlyWhenEditable(bool enabled)
```
?

signedav:
Yes, but preferably use flags instead of a boolean. If you want to do it, then I appreciate that. It got a little forgotten, I guess.

PR #39336, opened 2020-10-13
[symbology] Fix SVG marker anchor calculation and improve bounds

nirvn:
## Description

This PR fixes a long-standing issue with SVG markers whereas its anchor calculation was wrongly assuming the SVG marker to be square in size. It resulted in wrong placement with setting the vertical anchor to top / bottom.


src/core/symbology/qgsmarkersymbollayer.cpp
```diff
@@ -2458,18 +2459,24 @@ bool QgsSvgMarkerSymbolLayer::writeDxf( QgsDxfExport &e, double mmMapUnitScaleFa
 QRectF QgsSvgMarkerSymbolLayer::bounds( QPointF point, QgsSymbolRenderContext &context )
 {
   bool hasDataDefinedSize = false;
-  double scaledSize = calculateSize( context, hasDataDefinedSize );
-  scaledSize = context.renderContext().convertToPainterUnits( scaledSize, mSizeUnit, mSizeMapUnitScale );
+  double scaledWidth = calculateSize( context, hasDataDefinedSize );
```
nyalldawson, line 2462: I don't think these changes correctly account for data defined symbol names...?

nirvn: Hmm, you are right. I guess there's no way around calling the svg cache to get height for data defined names/paths. 

This means to get proper anchoring, we'll have to do the same in renderPoint too. You're good with that?

nirvn:
@nyalldawson , code updated.

src/core/symbology/qgsmarkersymbollayer.cpp
```diff
@@ -2487,29 +2494,37 @@ QRectF QgsSvgMarkerSymbolLayer::bounds( QPointF point, QgsSymbolRenderContext &c
   }
   strokeWidth = context.renderContext().convertToPainterUnits( strokeWidth, mStrokeWidthUnit, mStrokeWidthMapUnitScale );
 
-  //need to get colors to take advantage of cached SVGs
```
nyalldawson, line 2490: can you leave this comment? I found it useful to know why the colors are being decoded here

nyalldawson (approved):

src/core/symbology/qgsmarkersymbollayer.cpp
nirvn: @nyalldawson , will re-add -- thanks for the review, the SVG marker code logic can be a bit of a PITA.

Issue #39275, opened 2020-10-09
build failed with Python 3.8.3


longhuan2018:
OS : Microsoft Windows 10 Pro (Chinese Version)
Compiler: MSVC 2017
Python Version : 3.8.3
SIP Version :  4.19.24
QT Version : 5.15.0

The following error is prompted during compilation
```
E:\vcpkg\buildtrees\qgis\x64-windows-dbg\python\core\sip_corepart3.cpp(231556): error C2248: 'QgsAuthManager::QgsAuthManager': cannot access protected member declared in class 'QgsAuthManager' [E:\vcpkg\buildtrees\qgis\x64-windows-dbg\python\python_module_qgis__core.vcxproj]
         E:\vcpkg\buildtrees\qgis\src\al-3_14_16-a174ab543c.clean\src\core\auth\qgsauthmanager.h(755): note: see declaration of 'QgsAuthManager::QgsAuthManager'
         E:\vcpkg\buildtrees\qgis\src\al-3_14_16-a174ab543c.clean\src\core\auth\qgsauthmanager.h(64): note: see declaration of 'QgsAuthManager'
```

longhuan2018:
I found that adding "explicit QgsAuthManager();" in the "python\core\auto_generated\auth\qgsauthmanager.sip.in" file can solve the bug.

```
  protected:


    explicit QgsAuthManager();
};

/************************************************************************
 * This file has been generated automatically from                      *
 *                                                                      *
 * src/core/auth/qgsauthmanager.h                                       *
 *                                                                      *
 * Do not edit manually ! Edit header and run scripts/sipify.pl again   *
 ************************************************************************/

```

lbartoletti:
Can not reproduce it (on FreeBSD). Can you open a PR with your patch?

PR #39339, opened 2020-10-13
Fix python 3.8.3 build failed

longhuan2018:
fix issues #39275


nyalldawson:
This seems to indicate that there's no issue with Python 3.8 -- the ci build and tests all pass. I suspect #39275 is a local issue.

longhuan2018:
My test environment is windows, but I found WITH_BINDINGS=OFF in Windows cross build with MXE. I will test some other environments to see if this problem can be reproduced.

longhuan2018:
My test environment is windows, but I found WITH_BINDINGS=OFF in Windows cross build with MXE. I will test some other environments to see if this problem can be reproduced.

PR #39341, opened 2020-10-13
[layer tree] Avoid blurry icons on devices with a pixel ratio != 1

nirvn:
## Description

The layer tree model's scaleIconSize() function needs to take into account the device pixel ratio to produce properly sized icons in environments where Qt's high dpi scaling (i.e. AA_EnableHighDpiScaling) is switched on.

See before (left) vs. PR (rigth):
image

### DIFF
--- a/python/core/auto_generated/qgsapplication.sip.in
+++ b/python/core/auto_generated/qgsapplication.sip.in
@@ -937,6 +937,17 @@ Set a single custom expression variable.
 .. versionadded:: 3.0
 %End
 
+    static int scaleIconSize( int standardSize );
+%Docstring
+Scales an icon size to compensate for display pixel density, making the icon
+size hi-dpi friendly, whilst still resulting in pixel-perfect sizes for low-dpi
+displays.
+
+``standardSize`` should be set to a standard icon size, e.g. 16, 24, 48, etc.
+
+.. versionadded:: 3.16
+%End
+
     int maxConcurrentConnectionsPerPool() const;
 %Docstring
 The maximum number of concurrent connections per connections pool.
--- a/src/core/layertree/qgslayertreemodel.cpp
+++ b/src/core/layertree/qgslayertreemodel.cpp
@@ -13,18 +13,16 @@
  *                                                                         *
  ***************************************************************************/
 
-#include <QApplication>
-#include <QDesktopWidget>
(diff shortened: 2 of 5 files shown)


elpaso:
@nirvn did you consider making a QgsApplication method for that, so that we can reuse it whenever we need to scale some icons? I mean: not only in the layer tree.

nirvn:
@elpaso , that's a possibility -- @nyalldawson , thoughts?

nyalldawson:
Is the smaller size of the point in the after screenshot correct? Does it match the canvas size now?

nirvn:
@nyalldawson , yes it does

m-kuhn:
Would we ideally use `layerTreeView->screen()->physicalDotsPerInch()` -- or is this not feasible because we are in the model?

nirvn:
@m-kuhn , not possible from the model itself AFAIK.

nyalldawson:
> @elpaso , that's a possibility -- @nyalldawson , thoughts?

Yes, that would be good.

There's already QgsGuiUtils::scaleIconSize, which is a near identical copy of this. Perhaps it's possible to call a QgsApplication method from the QgsGuiUtils method instead, so that we only have a single version of this conversion floating around...

nirvn:
@nyalldawson , I can't call QgsGuiUtils from src/core though right?

nyalldawson:
That's why I'm proposing moving the guts of the QgsGuiUtils function to QgsApplication

nirvn:
Ah, got it. I shall proceed.

nirvn:
@nyalldawson , done, all moved to QgsApplication::scaleIconSize

nyalldawson (approved):

nyalldawson:
@PeterPetrik @3nids can you please test on macos after this is merged and confirm that there's no regressions? I believe macos is the only other platform which defaults to a > 1 pixel ratio

3nids:
@nyalldawson from a first quick look, everything looks the same than before.

3nids:
not sure this is related but lately...
image

nirvn:
I was worried it'd do that. Qt still so messy with hidpi handling. I'll commit a follow up PR to address issue.

3nids:
and also the data defined button…
image

PR #39343, opened 2020-10-13
Fix: proj4 v6 wkt initializes an invalid CRS

rldhont:
## Description
The wkt can be a valid crs description but it can be an invalid crs. This depends on the proj4 version. A wkt provided by proj4 version >= 6 can be invalid for proj4 <= 5 but the proj4 string provides a valid crs.

For exemple this WKT is provided by proj4 version 6. It initializes an invalid CRS with proj4 version <= 5.
```
PROJCRS["unknown",
    BASEGEOGCRS["unknown",
        DATUM["World Geodetic System 1984",
            ELLIPSOID["WGS 84",
                6378137,
                298.257223563,
                LENGTHUNIT["metre",1]
            ],
            ID["EPSG",6326]
        ],
        PRIMEM["Greenwich",0,
            ANGLEUNIT["degree",0.0174532925199433],
            ID["EPSG",8901]
        ]
    ],
    CONVERSION["unknown",
        METHOD["Mercator (variant B)",
            ID["EPSG",9805]
        ],
        PARAMETER["Latitude of 1st standard parallel",
            46,
            ANGLEUNIT["degree",0.0174532925199433],
            ID["EPSG",8823]
        ],
        PARAMETER["Longitude of natural origin",
            0,
            ANGLEUNIT["degree",0.0174532925199433],
            ID["EPSG",8802]
        ],
        PARAMETER["False easting",
            0,
            LENGTHUNIT["metre",1],
            ID["EPSG",8806]
        ],
        PARAMETER["False northing",
            0,
            LENGTHUNIT["metre",1],
            ID["EPSG",8807]
        ]
    ],
    CS["Cartesian",2],
    AXIS["(E)",
        east,
        ORDER[1],
        LENGTHUNIT["metre",
            1,
            ID["EPSG",9001]
        ]
    ],
    AXIS["(N)",
        north,
        ORDER[2],
        LENGTHUNIT["metre",
            1,
            ID["EPSG",9001]
        ]
    ]
]
```

### DIFF
--- a/src/core/qgscoordinatereferencesystem.cpp
+++ b/src/core/qgscoordinatereferencesystem.cpp
@@ -2067,6 +2067,15 @@ bool QgsCoordinateReferenceSystem::readXml( const QDomNode &node )
 
       const QString wkt = srsNode.namedItem( QStringLiteral( "wkt" ) ).toElement().text();
       initialized = createFromWktInternal( wkt, description );
+
+      // the wkt can be a valid crs description but it can be an invalid crs
+      // this depends on the proj4 version
+      // a wkt provided by proj4 version >= 6 can be invalid for proj4 <= 5
+      // but the proj4 string provides a valid crs
+      if ( initialized && !isValid() )
+      {
+        initialized = false;
+      }
     }
 
     if ( !initialized )


src/core/qgscoordinatereferencesystem.cpp
```diff
@@ -2067,6 +2067,15 @@ bool QgsCoordinateReferenceSystem::readXml( const QDomNode &node )
 
       const QString wkt = srsNode.namedItem( QStringLiteral( "wkt" ) ).toElement().text();
       initialized = createFromWktInternal( wkt, description );
+
+      // the wkt can be a valid crs description but it can be an invalid crs
+      // this depends on the proj4 version
+      // a wkt provided by proj4 version >= 6 can be invalid for proj4 <= 5
+      // but the proj4 string provides a valid crs
+      if ( initialized && !isValid() )
```
nyalldawson, line 2075: I don't understand this change -- as far as I can see, createFromWktInternal always returns d->mIsValid already from the expected return paths. Which return path in createFromWktInternal is being hit here?

rldhont: I have tested this code in QGIS Python Console:
```py
wkt = 'PROJCRS["unknown", BASEGEOGCRS["unknown", ... ] ]'
crs = QgsCoordinateReferenceSystem()
QgsCoordinateReferenceSystem.createFromWkt(crs, wkt)
# Returns True
crs.isValid()
# Returns False
```
If I do the same with a crs' wkt provided by proj4 version < 6, I have:
```py
wkt = 'PROJCS["unnamed", GEOGCS["WGS 84", ... ]'
crs = QgsCoordinateReferenceSystem()
QgsCoordinateReferenceSystem.createFromWkt(crs, wkt)
# Returns True
crs.isValid()
# Returns True
```

nyalldawson: Can you use a debugger and confirm which return path is being hit?

rldhont: The issue is line 872 https://github.com/qgis/QGIS/blob/master/src/core/qgscoordinatereferencesystem.cpp#L872 and in the other createFrom methods and probably in the way CRS definition cache is managed.

nyalldawson: Ah - in that case line 872 should read `return d->mIsValid`. Can you add a test for this too? (i.e. call createFromWkt twice with the same bad wkt string and test that both times it returns false).

rldhont: I have created a new PR with this fix and more https://github.com/qgis/QGIS/pull/39419

PR #39349, opened 2020-10-13
Separate qgis_process params with --

m-kuhn:
## Description

Collision free parameters for qgis_process by separating executable args and algorithm args with `--`

```
qgis_process run algname --distanceUnits=meters -- PARAM1=23 PARAM2=/path/to/file distanceUnits=someOtherParam
```

Still supports the legacy version using `--PARAM=VALUE` syntax

@nyalldawson looking for your feedback before finetuning


nyalldawson:
Looks good! There's tests for this tool also which will need extending (also the docs)

m-kuhn:
CC @stibu81 how does this look to you?

nyalldawson (approved):

stibu81:
> CC @stibu81 how does this look to you?

@m-kuhn This looks like an excellent idea! Thanks for letting me know.

PR #39350, opened 2020-10-13
Fix deleting python console history entries

### LINKED ISSUE #38894: Deleting entries from the python console history causes python error
**Describe the bug**
Deleting entries from the python console history causes python error

**How to Reproduce**
1. Open the python console
2. ctrl+shift+space to open the command history
3. select an entry and hit del

**Additional context**
This was introduced by the changes in #37764
Gently ping @nirvn

uclaros:
## Description
Another approach at fixing #38894 while preserving the soft-history.

Fixes #38894


slarosa:
Thank you @uclaros.

PR #39354, opened 2020-10-13
Fix GPKG FID issues when running in-place mode

### LINKED ISSUE #33816: "duplicate key value violates unique constraint" error when using "Edit Feature in Place" with PostGIS Layer
in our postgis database for line(road) layer we have defined primary key field and have applied unique not null constraint when i save split line segements manually by editing toolbar it works fine, but when i split lines with "split with lines" tool by turning on edit feature in place it gives me postgis error
"""
Could not commit changes to layer pk_roads

Errors: ERROR: 4 feature(s) not added.
  
  Provider errors:
      PostGIS error while adding features: ERROR:  duplicate key value violates unique constraint "pk_roads"
    CONTEXT:  SQL statement "INSERT INTO pk_roads SELECT(pk_roads '(,,,,,,,,,,,,0102000020110F000002000000A00ABFB828A35F411E7335C4B9484D41C9CA687C3CA35F41405E7DA3BD484D41,103564332,"digi1,",,,,,,,,,,,,,,,,,,,,,,,,)').* RETURNING id;"
    PL/pgSQL function fn_rd_partition() line 9 at EXECUTE
    
"""
but from my point of view it must have to work fine for "split with lines" tool if same splitting works from "Advance Digitizing Toolbar"

### LINKED ISSUE #37761: Split with lines causes data loss
**Describe the bug**
I have a memory line layer and with this layer I want to split geoms of gpkg-file by using the edit in place mode. After finsihing the job there is not the expect result but aduplication of geoms in the gpk-file

**How to Reproduce**

1. load a poly-layer of gpkg-file
2. creaete a memory line line
3. active edit in place mode
4. start the processing tool split with lines
5. no split but a duplication of geoms

**QGIS and OS versions**

QGIS version | 3.14.0-Pi | QGIS code revision | 9f7028fd
-- | -- | -- | --
Compiled against Qt | 5.14.2 | Running against Qt | 5.14.2
PostgreSQL Client Version | 12.2 | SpatiaLite Version | 4.3.0a
QWT Version | 6.1.4 | QScintilla2 Version | 2.10.8
OS Version | KDE Flatpak runtime
Active python plugins | VectorBender; QuickMultiAttributeEdit3; polystrip; qfieldsync; MetaSearch; db_manager; processing

**Additional context**

nyalldawson:
Correctly discard fid field values when running algorithms with the RegeneratePrimaryKey flag in in-place mode

Fixes #37761, fixes #33816

### DIFF
--- a/src/core/qgsvectorlayerutils.cpp
+++ b/src/core/qgsvectorlayerutils.cpp
@@ -735,13 +735,22 @@ void QgsVectorLayerUtils::matchAttributesToFields( QgsFeature &feature, const Qg
   feature.setFields( fields );
 }
 
-QgsFeatureList QgsVectorLayerUtils::makeFeatureCompatible( const QgsFeature &feature, const QgsVectorLayer *layer )
+QgsFeatureList QgsVectorLayerUtils::makeFeatureCompatible( const QgsFeature &feature, const QgsVectorLayer *layer, QgsFeatureSink::SinkFlags sinkFlags )
 {
   QgsWkbTypes::Type inputWkbType( layer->wkbType( ) );
   QgsFeatureList resultFeatures;
   QgsFeature newF( feature );
   // Fix attributes
   QgsVectorLayerUtils::matchAttributesToFields( newF, layer->fields( ) );
+
+  if ( sinkFlags & QgsFeatureSink::RegeneratePrimaryKey && layer->storageType() == QLatin1String( "GPKG" ) )
+  {
+    // drop incoming fid value, let it be regenerated
+    const int fidIndex = layer->fields().lookupField( QStringLiteral( "fid" ) );
+    if ( fidIndex >= 0 )
+      newF.setAttribute( fidIndex, QVariant() );
+  }
+
   // Does geometry need transformations?
   QgsWkbTypes::GeometryType newFGeomType( QgsWkbTypes::geometryType( newF.geometry().wkbType() ) );
   bool newFHasGeom = newFGeomType !=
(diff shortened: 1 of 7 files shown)


src/core/qgsvectorlayerutils.cpp
```diff
@@ -734,13 +734,22 @@ void QgsVectorLayerUtils::matchAttributesToFields( QgsFeature &feature, const Qg
...
   QgsWkbTypes::Type inputWkbType( layer->wkbType( ) );
   QgsFeatureList resultFeatures;
   QgsFeature newF( feature );
   // Fix attributes
   QgsVectorLayerUtils::matchAttributesToFields( newF, layer->fields( ) );
+
+  if ( sinkFlags & QgsFeatureSink::RegeneratePrimaryKey && layer->storageType() == QLatin1String( "GPKG" ) )
+  {
+    // drop incoming fid value, let it be regenerated
+    const int fidIndex = layer->fields().lookupField( QStringLiteral( "fid" ) );
```
rouault, line 748: you should probably use the name of the first field returned by the provider instead of the hard coded "fid", in case it would have been named differently than the default.

Question (related to my last email): why do we need that hack for GPKG, but not with the Postgres provider when it exposes a integer primary key column ?

nyalldawson: Hmm - I'll need to dig deeper. My first guess at the difference proved wrong...

nyalldawson: Ok, fixed. I've made this method compatible with non-ogr/non-gpkg providers too.

tests/src/python/test_qgsprocessinginplace.py
```diff
@@ -912,6 +913,26 @@ def test_unique_constraints(self):
 
         self.assertTrue(gpkg_layer.commitChanges())
 
+    def test_regenerate_fid(self):
+        """Test RegeneratePrimaryKey flag"""
+        temp_dir = QTemporaryDir()
```
elpaso, line 918:
```suggestion

        temp_dir = QTemporaryDir()
```

nyalldawson: what am I missing? an extra line?

elpaso: yes, I though all our nice pre commit would have taken care of that but apparently not.

PR #39357, opened 2020-10-14
Fix copy/paste/select all shortcuts don't work when layer tree is selected

### LINKED ISSUE #38849: ctrl+v to paste features no longer works when focus is on layers panel
**Describe the bug**
When using ctrl-v to paste features, the focus must be on the map canvas in order to work.
A very common scenario is to want to copy features between layers and it used to work like
1. set layer A active
2. select features
3. ctrl-c
4. set layer B active
5. ctrl-v

This no longer works since step 4. sets the focus on the layers panel. In order for it to work one has to click on the map canvas between steps 4 and 5 which is kind of annoying.

This is a regression and could be caused by #36955

**QGIS and OS versions**
3.14.1 win7 and linux

nyalldawson:
Fixes #38849


nirvn (approved):
Good compromise

PR #39367, opened 2020-10-14
Don't clear document path from external resource widget when its null

### LINKED ISSUE #39063: QGIS 3.10.8 and 3.14 crashes when opening attribute table after setting up a photo field and using lay-out form with attachment widget
When adding a field for photos (text string length 150 caracters) in a point layer and indicating that the attribute form must use a attachment widget (setting the path to 'relative paths' + relative to project path) QGIS crashes when I try to open the attribute table. The problem persists when using another user profile and running the project on QGIS 3.14 'pi' within Ubuntu/Mint).

Remarks
1/ **At first this worked!** I was able to make a package of the project with QFieldSync and run different lay-out forms on the QField mobile app, making pictures in the field
2/ While trying to add maptips in the project with the goal to hoover over the points and see the corresponding photos this didn't work out as I wish (I am not een very experienced user...). I tried using HTML codes in the display tab to show the pictures based on the path in the field 'foto' of the layer "BomenEnBos" (see video https://youtu.be/6HYqDCdpJko from 6:05). This failed and I deleted the html code.
3/ When I delete the field with the photos ("foto") there is no problem opening the attribute table
4/ When adding the photo field as a text file, there is no problem
5/ When adding the photo field and setting up the form as a attachment file, QGIS crashes again when trying to open the attribute table
6/ When adding the photo field and setting up the form as a text edit file, there is no problem
7/ Strange, while playing with the point layer to find out when QGIS crashes, I could add the photo field with the form + attachment widget in two of the four lay-out forms without QGIS crashing. QGIS still crashes in the lay-out forms 'GevaarlijkeBomen' and 'TBomen_overige'.

The result is that I can't go out with the QField app for making pictures in both forms...
For your convenience I add a geopackage and project of the layer with styles (see zip file). I tried the project and it crashes.

## Report Details

**Crash ID**: [1af84a1884b8cc130c86858ec052a44d30c5f9de](https://github.com/qgis/QGIS/search?q=1af84a1884b8cc130c86858ec052a44d30c5f9de&type=Issues)

**Stack Trace**
<pre>
QGridLayoutEngine::rowCount :
QTextEngine::itemize :
QTextLayout::beginLayout :
QWidgetLineControl::redoTextLayout :
QWidgetLineControl::updateDisplayText :
QLineEdit::changeEvent :
QWidget::event :
QLineEdit::event :
QApplication::notify :
</pre>

**QGIS Info**
QGIS Version: 3.10.8-A Coru�a
QGIS code revision: d09a4f2388
Compiled against Qt: 5.11.2
Running against Qt: 5.11.2
Compiled against GDAL: 3.0.4
Running against GDAL: 3.0.4

**System Info**
CPU Type: x86_64
Kernel Type: winnt
Kernel Version: 10.0.18363

Knipsel
Knipsel2
Screenshot_20200926-181507__01
Screenshot_20200926-181443__01

[Bugreports.zip](https://github.com/qgis/QGIS/files/5291661/Bugreports.zip)

troopa81:
## Description

Fixes #39063 : We enter in an infinite loop with *updateValues* and *valueChanged* calling themselves because the widget value never stops to switch from "NULL" to "" when we clear the QFileWidget document path.


tests/src/gui/testqgsexternalresourcewidgetwrapper.cpp
```diff
@@ -0,0 +1,148 @@
...
+  QgsApplication::exitQgis();
+}
+
+void TestQgsExternalResourceWidgetWrapper::init()
+{
+  vl = qgis::make_unique<QgsVectorLayer>( QStringLiteral( "NoGeometry?field=link:string" ),
+                                          QStringLiteral( "myvl" ),
+                                          QLatin1String( "memory" ) );
+
+  // // add fields
```
nyalldawson, line 67: what's with this? is it needed?

troopa81: Oops no!

PR #39368, opened 2020-10-14
Fix WFS-T 1.1.0 support

elpaso:
This can be considered a bugfix because when the user selects WFS 1.1.0 in the WFS source select dialog (or get it selected by "Auto detect") it is expected that WFS 1.1.0 is also used for transactions but WFS-T 1.1.0 was not supported until this patch effectively making it impossible to post transactions on certain servers (ESRI WFS).

(the same stands for 2.0 actually but that version is not part of this fix)

I've manually tested this PR against:
- Geoserver
- QGIS server
- ESRI mapserver

the automated tests for CI were generated with geoserver.


src/providers/wfs/qgswfsprovider.cpp
```diff
@@ -1580,17 +1582,25 @@ bool QgsWFSProvider::sendTransactionDocument( const QDomDocument &doc, QDomDocum
...
 QDomElement QgsWFSProvider::createTransactionElement( QDomDocument &doc ) const
 {
   QDomElement transactionElem = doc.createElementNS( QgsWFSConstants::WFS_NAMESPACE, QStringLiteral( "Transaction" ) );
-  // QString WfsVersion = mShared->mWFSVersion;
-  // For now: hardcoded to 1.0.0
-  QString WfsVersion = QStringLiteral( "1.0.0" );
-  transactionElem.setAttribute( QStringLiteral( "version" ), WfsVersion );
+  const QString WfsVersion = mShared->mWFSVersion;
+  // only 1.1.0 and 1.0.0 are supported
+  if ( WfsVersion == QStringLiteral( "1.1.0" ) )
```
rouault, line 1615: what if the user has let version = "auto" and the server returns a WFS 2.0 GetCapabilites response ? It looks like we will use WFS-T 1.0 then, or am I missing something ? Probably that if mWFSVersion == "2.0.0" we should use WFS-T 1.1.0 then ? The most prudent way would probably to issue a GetCapabilties query with VERSION=1.1.0 forced and see if the server reports transaction capabilities. But that's a complication...

src/providers/wfs/qgswfsdatasourceuri.cpp
```diff
@@ -373,6 +373,12 @@ bool QgsWFSDataSourceURI::hideDownloadProgressDialog() const
   return mURI.hasParam( QgsWFSConstants::URI_PARAM_HIDEDOWNLOADPROGRESSDIALOG );
 }
 
+
+bool QgsWFSDataSourceURI::preferCoordinatesForWfst11() const
```
rouault, line 377: This doesn't seem to be used at all. I guess there will be a forth coming commit that will make use of it instead of the currently disabled GML3 code path ?

src/providers/wfs/qgswfsprovider.cpp
```diff
@@ -877,7 +877,17 @@ bool QgsWFSProvider::addFeatures( QgsFeatureList &flist, Flags flags )
       {
         the_geom.convertToMultiType();
       }
-      QDomElement gmlElem = QgsOgcUtils::geometryToGML( the_geom, transactionDoc );
+      QDomElement gmlElem;
+      // WFS 1.1.0 uses preferably GML 3, but ESRI mapserver in 2020 doesn't like it so we stick to GML2
+      /*
+      if ( mShared->mWFSVersion == QStringLiteral( "1.1.0" ) )
+      {
+        gmlElem = QgsOgcUtils::geometryToGML( the_geom, transactionDoc, QLatin1String( "GML3" ) );
```
rouault, line 885: if this gets re-enabled, I guess we might have axis order issues. On a quick look the method signature you use for QgsOgcUtils::geometryToGML( ) will not take into account EPSG axis order, which will probably be a problem with compliant servers when using urn:ogc:def:crs:EPSG::4326. You would need the extended signature where the value of invertAxisOrientation would be true when crs.hasAxisInverted()

Shouldn't the settings on the reading part that mess with axis order (mURI.ignoreAxisOrientation() / mURI.invertAxisOrientation()) be applied there too ?  See QgsWFSSharedData::computeFilter() for the logic. Basically in WFS 1.0, we only take into account mURI.invertAxisOrientation()) (the user says "my server does broken things. invert the axis order"), and in WFS 1.1, we take into account both : if ignoreAxisOrientation() is set then it means "the server behaves as WFS 1.0 and it returns long, lat for EPSG:4326", otherwise EPSG:4326 will be "lat, lon" . And invertAxisOrientation() is applied on top of that.

src/providers/wfs/qgswfsdatasourceuri.cpp
elpaso: A broken cherry pick from another branch, I'll fix it now.

src/providers/wfs/qgswfsprovider.cpp
elpaso: Sorry, I missed a commit in the cherry-pick, now it's the updated version. I agree on the full signature for geometryToGML, I'll try to apply the logic you recommend.

src/providers/wfs/qgswfsprovider.cpp
elpaso: I considered to fall-back from 2.0 to 1.1 but I chose the less impacting changes: leave the old behavior (fallback to 1.0). I can implement the fall-back from 2.0 to 1.1 if you think it's safe.

src/providers/wfs/qgswfsshareddata.cpp
```diff
@@ -27,6 +27,7 @@ QgsWFSSharedData::QgsWFSSharedData( const QString &uri )
   , mURI( uri )
 {
   mHideProgressDialog = mURI.hideDownloadProgressDialog();
+  mServerPrefersCoordinatesForTransactions_1_1 = uri.contains( QLatin1String( "/arcgis/" ), Qt::CaseSensitivity::CaseInsensitive );
```
elpaso, line 30: this is wrong and I'll remove this: it is now an option in the connection config.

src/providers/wfs/qgswfsprovider.cpp
rouault: > I can implement the fall-back from 2.0 to 1.1 if you think it's safe.

with 1.1 you'll get axis order issues :-) so I'm not completely convinced

src/providers/wfs/qgswfsprovider.cpp
elpaso: @rouault I'm not sure I get it right: 

Is this the correct interpretation for WFS-T?

## WFS-T 1.1

```c++
const bool applyAxisInversion = ( crs().hasAxisInverted() && !  ignoreAxisOrientation() ) || invertAxisOrientation();
```

## WFS-T 1.0:

```c++
const bool applyAxisInversion = crs().hasAxisInverted() || invertAxisOrientation();
```

elpaso: @rouault I've tried to apply this logic in my last commit, I also refactored it into a method.

elpaso: I have tried to follow GeoServer conventions here: https://docs.geoserver.org/stable/en/user/services/wfs/axis_order.html

For QGIS Server WFS-T 1.1 that exposes srs as `EPSG:4326` I had to check for `useEPSGColumnFormat` and skip axis inversion in that case. 

This seems to work, I added a few QGIS Server integration tests for that. 

I'm not sure about Geomedia that was mentioned in a comment.

We should be close to a working solution, at least I hope so :smile_cat:

src/providers/wfs/qgswfsprovider.cpp
elpaso: Ok, I'll leave it like it was, there is still the case when 'auto' returns 1.1, in this case 1.1 will be used for WFS-T too but I guess it's ok.

src/providers/wfs/qgswfsprovider.cpp
```diff
@@ -787,6 +787,49 @@ void QgsWFSProvider::reloadProviderData()
...
+  {
+    gmlVersion = QgsOgcUtils::GML_2_1_2;
+    applyAxisInversion = mShared->mURI.invertAxisOrientation();
+  }
+
+  gmlElem = QgsOgcUtils::geometryToGML(
+              geometry,
+              transactionDoc,
+              gmlVersion,
+              mShared->srsName(),
```
rouault, line 825: I guess it could be a bit funky if you use WFS 2.0 on the reading side that returns a urn::bla:EPSG:4326 and on the transaction side you use that when outputing the srsName in the geometry with WFS-T 1.0. Servers might have issues to interpret that correctly. They will probably assume lon, lat order because of WFTS-T 1.0, but the urn::bla:EPSG:4326 that could suggest lat, lon order.  Wouldn't we also have issues with the feature ids ? There's no guarantee that the gml:id of WFS 2.0 might be the fid of WFS-T 1.0 ...

elpaso: I used srsName here when testing geoserver: it sends `urn:...` as a default SRS and that means latitude/longitude, but if we use `EPSG:4326` it assumes longitude/latitude. 

According to geoserver docs, 
-    WFS 1.0.0: Provides geographic coordinates in east/north and may not be trusted to respect the EPSG definition axis order.
-   WFS 1.1.0: Respects the axis order defined by the EPSG definition.
-   WFS 2.0.0: Respects the axis order defined by the EPSG definition.

Anyway, besides to implement 2.0 what would you recommend? I can't find a better solution to fit all situations.

rouault: > I used srsName here when testing geoserver: it sends urn:... as a default SRS and that means latitude/longitude, but if we use EPSG:4326 it assumes longitude/latitude.

So currently, let's say with GeoServer with WFS 2.0 that returns a srsName="urn::...EPSG:4326", for geometries we emit in transactions as WFS-T 1.0, do they have srsName="urn::...EPSG:4326" and lat,lon order ?  Or srsName="EPSG:4326" and lon,lat order ?

> Anyway, besides to implement 2.0 what would you recommend?

At least for GeoServer and other well behaved servers, using WFS-T 1.1 with WFS 2.0 would seem to me to be a safer solution since they should have the same axis order behaviour and gml:id

elpaso: If GeoServer WFS 2.0 sends `urn::...EPSG:4326` (or  `urn:x-ogc:def:crs:EPSG:4326`)  in capabilities and we emit a transaction WFS-T 1.0 it will use srsName `urn::...EPSG:4326` lat/lon (inverted axis).

But if a server sends `EPSG:4326` (like QGIS does) we follow GeoServer convention which is to treat `EPSG:4326` as lon/lat.

rouault: > we emit a transaction WFS-T 1.0 it will use srsName `urn::...EPSG:4326` lat/lon (inverted axis).

OK, that's what I understood as well from the code. This is probably a "non-vetted" combination. Anyway I just wanted to point of potential issues I could think of. I let you judge of the best / least-worse conclusion :-)

elpaso: Honestly I'm a bit lost into all the combinations of implementations/specification-flaws and workarounds that we can find in the wild, I hope this is a good compromise. 

We could even add a separate option (hack) for inverted axis in transactions or to choose the WFS-T version for transactions but it looks to me that the list of options (and potentially incompatible ones) could be really overwhelming.

Anyway, thank you for the review and all the help!

Do you feel like we can merge (and backport) this PR?

rouault: > Do you feel like we can merge (and backport) this PR?

The only situation where this changes behaviour underneath is when WFS 1.1 is used for the reading side: now, we'll switch to WFS-T 1.1, which is very reasonable.
Yes, we can probably treat that as a bug fix.
I'm wondering about the new UI string added: what's our policy regarding this ? Are there translation updates in the stable branch ?

elpaso: I don't know if we have real policy, probably better to avoid changes but a new little string in advanced options is probably fine.

rouault: OK go ahead :-)

rouault (approved):

PR #39379, opened 2020-10-15
Allow selection of invalid geometries

### LINKED ISSUE #38460: 'Select by polygon' doesn't select anymore geometries with errors
**Describe the bug**
When trying to do a seleciton on a layer, not all features inside the drawn polygon are being selected. 

**How to Reproduce**

1. Add layer (mine is a polygon GeoPackage in EPSG:3879
2. Click on 'Select features by area/polygon'
3. Try to select features
4. See error in the screenshot below

**QGIS and OS versions**
QGIS version 3.14.15-Pi running on Windows 10. No plug-ins installed in this fresh profile.

**Additional context**

Here's how it **should work** and how it works in 3.10:
select_310

Here's how it "works" in the latest 3.14:
select_314

nyalldawson:
When selecting features via rect, ignore GEOS errors due to invalid geometries and allow these features to be selected based on their  bounding boxes alone.

Fixes #38460

I'm not 100% on this fix. I can't see why this behavior has changed since 3.10, unless it's caused by changes in GEOS itself. Feedback is appreciated!!

### DIFF
--- a/python/core/auto_generated/geometry/qgsgeometryengine.sip.in
+++ b/python/core/auto_generated/geometry/qgsgeometryengine.sip.in
@@ -257,10 +257,30 @@ Splits this geometry according to a given line.
 
     virtual QgsAbstractGeometry *offsetCurve( double distance, int segments, int joinStyle, double miterLimit, QString *errorMsg = 0 ) const = 0 /Factory/;
 
+    void setLogErrors( bool enabled );
+%Docstring
+Sets whether warnings and errors encountered during the geometry operations should be logged.
+
+By default these errors are logged to the console and in the QGIS UI. But for some operations errors are expected and logging
+these just results in noise. In this case setting ``enabled`` to ``False`` will avoid the automatic error reporting.
+
+.. versionadded:: 3.16
+%End
+
   protected:
+
+    void logError( const QString &engineName, const QString &message ) const;
+%Docstring
+Logs an error ``message`` encountered during an operation.
+
+.. seealso:: :py:func:`setLogErrors`
+
+.. versionadded:: 3.16
+%End
 
     QgsGeometryEngine( const QgsAbstractGeometry *geometry );
 };
+
 
 /************************************************************************
  * This file has been generated automatically from                      *
--- a/src/app/qgsmaptoolselectutils.cpp
+++ b/src/app/qgsmaptoolselectutils.cpp
@@ -30,6 +30,7 @@ email                : jpalmer at linz dot govt dot nz
 #include "qgis.h"
 #include "qgsproject.h"
 #include "qgsexpressioncontextutils.h"
+#include "qgsmessagelog.h"
 
 #include <QMouseEvent>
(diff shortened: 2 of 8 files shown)


uclaros:
This will select false positives, ie. invalid geometries that may not even be visible on the current extent but their mbr is. There will be no visual clue that this has happened, apart from the number of selected features.
The current problematic behavior though is falsely not selecting invalid geometries, which has the selection highlighting (or lack of) as an obvious visual clue.

nyalldawson:
@uclaros actually the situation is a bit more complex then that. The iterator is being requested with the Exact Intersection flag here, so we already can be confident that they do have more then just a bounding box intersection (and if not then it's a bug in the particular data provider). I suspect that the test in the non-contains branch here is actually redundant and could be removed.

uclaros:
I think that's true only for the rectangle selection where the rectangle is the same as its mbr. For polygonal selection though, the exact intersection does not suffice and the non-contains branch is also required.

nyalldawson:
Good point! How do you think we should handle this then?

uclaros:
I don't know, maybe something like this?
```
if ( !selectGeomTrans.contains( g ) &&
     !selectGeomTrans.lastError().isEmpty() &&
     !selectGeomTrans.contains( g.makeValid() ) )
  continue;
```
I think this will give a second chance to the invalid geometry.
But, is `lastError` reset after a successful `contains()`?
And what if the `selectGeomTrans` is invalid instead?

nyalldawson:
> I don't know, maybe something like this?

Yep, that should work!

> But, is lastError reset after a successful contains()?

Yep, it's reset before calling any QgsGeometry method which can potentially set it.

> And what if the selectGeomTrans is invalid instead?

We could just fix the selection geom once before testing any features -- e.g. buffer it by 0.

uclaros:
We could also check if `selectGeometry` is the same as its mbr, ie selection was made by rectangle. In that case, completely skip the redundant non-contains branch. This I believe is the most used case (the default select by rectangle without pressing alt) so avoiding the extra intersection check can only be good.

Related but slightly off topic thought: would it make sense to move the refined selection logic to `QgsFeatureRequest` by adding some kind of `setFilterGeometry()` method that would support all spatial predicates? This logic is probably repeated in various other places and could be centralized.

nyalldawson:
> Related but slightly off topic thought: would it make sense to move the refined selection logic to QgsFeatureRequest by adding some kind of setFilterGeometry() method that would support all spatial predicates? This logic is probably repeated in various other places and could be centralized.

That'd be great! (It's not a trivial piece of work though...)

nyalldawson:
@uclaros (much) improved version pushed now.

uclaros:
That's (much) better!
Maybe we should also log a failed second chance in case `makeValid()` fails?

nyalldawson:
> Maybe we should also log a failed second chance in case makeValid() fails?

Done!

PR #39382, opened 2020-10-15
Fix WMS auth basic password encoding

### LINKED ISSUE #39243: Passwords are URL-encoded before base64 encoding for HTTP Basic auth
**Describe the bug**

QGIS supports HTTP Basic authentication when adding a WMS source if the user provides a username and password. The format for an HTTP Basic auth header is "Authorization: Basic _base64-encoded-user:pass_". When the password includes a reserved URL character e.g. % QGIS is URL-encoding the character before base64 encoding. This results in the wrong password being provided to the WMS and an authentication failure.

**How to Reproduce**
1. Add a WMS to QGIS with any valid URL (does not need to be a valid WMS to satisfy step 6) and provide the username & password "user" & "pa%%word" in the service definition's Basic auth tab. HTTP and not HTTPS may be preferable for inspecting traffic
2. Open QGIS Network Logger plugin, or establish some other way of inspecting HTTP traffic
3. Expand the added WMS to initiate a GetCapabilities request (it does not matter if the request fails)
4. Inspect the HTTP headers issued with the GetCapabilities request
5. Base64-decode the encoded value after "Basic " in the "Authorization" header
6. Observe text "user:pa%25%25word" showing that % in pa%%word have been URL-encoded

**QGIS and OS versions**
QGIS 3.10.10
MacOS 10.15.6

**Additional context**

elpaso:
Fixes #39243

Manually backported from master https://github.com/qgis/QGIS/pull/39347


nyalldawson (approved):

nyalldawson:

PR #39388, opened 2020-10-15
[OGR provider / GPKG] Avoid skipConstraintCheck() return true on the GPKG fid colum

rouault:
This aligns the QgsOgrProvider::skipConstraintCheck() implementation with QgsPostgresProvider::skipConstraintCheck()

And this makes GPKG behave like Postgres for the scenario in https://lists.osgeo.org/pipermail/qgis-developer/2020-October/062401.html

### DIFF
--- a/src/core/providers/ogr/qgsogrprovider.cpp
+++ b/src/core/providers/ogr/qgsogrprovider.cpp
@@ -1519,8 +1519,7 @@ QString QgsOgrProvider::defaultValueClause( int fieldIndex ) const
 bool QgsOgrProvider::skipConstraintCheck( int fieldIndex, QgsFieldConstraints::Constraint constraint, const QVariant &value ) const
 {
   Q_UNUSED( constraint )
-  // If the field is a fid, skip in case it's the default value
-  if ( fieldIndex == 0 && mFirstFieldIsFid )
+  if ( providerProperty( EvaluateDefaultValues, false ).toBool() )
   {
     return ! mDefaultValues.value( fieldIndex ).isEmpty();
   }


rouault:
@elpaso as I see you added this method per https://github.com/qgis/QGIS/commit/a12939d380d8dfd05e45ee9f9cb464541b52f59f , maybe you can think of of potential issues of the change I propose here ?

elpaso:
@rouault it was a long time ago, but I see I added a few tests, if they still passes I guess that this PR is ok.

nyalldawson (approved):

m-kuhn:
For reference: we had to introduce a patch on QField since this pull request.

What fails: we ask the provider for a default value -> it returns "Autogenerate" -> we try to set this value on a field -> it's a numeric field it can't set it to a text -> it checks for "is null a valid value" -> ogr says: no, constraint failure.

Is this expected behavior?

elpaso:
> For reference: we had to introduce a patch on QField since this pull request.
> 
> What fails: we ask the provider for a default value -> it returns "Autogenerate" -> we try to set this value on a field -> it's a numeric field it can't set it to a text -> it checks for "is null a valid value" -> ogr says: no, constraint failure.
> 
> Is this expected behavior?

@m-kuhn I think it is. The providers (ogr and spatialite) handle this special value internally and do the right thing.

What do you mean you cannot set it to a text? Attributes are QVariants, they can hold anything.

m-kuhn:
The missing gap here is the internal widget that cannot hold a QVariant.
It's almost the same as when a range widget is used in QGIS.

Simplified code sample:

```
numericWidget.setValue(provider->getDefaultValue()); # returns "Autogenerate"
# Widget tries to convert "Autogenerate" to number, fails and sets it to NULL -- (QGIS range would set it to 0)
feature.setAttribute(0, numericWidget.value()) # "Autogenerate" is lost
```

What we do on QField side is substitute `NULL` with `defaultValue` again before the constraint check which is a bit an awkward hack but works. I could imagine we have similar issues for some edge cases in QGIS.

I could not follow what particular issue was fixed by this PR, so I thought I'd at least report this finding back.

rouault:
> I could not follow what particular issue was fixed by this PR

The scenario described in https://lists.osgeo.org/pipermail/qgis-developer/2020-October/062401.html on a GeoPackage layer :
"""One difference I see is that in the GPKG case, when creating features through 
the GUI, I can set a new feature to a fid value already taken, whereas if I 
try to do the same with a Postgres layer, the GUI prevents me from doing that. 
There must be some extra hint that the Postgres provider provides, and GPKG 
does not.
"""

m-kuhn:
Good point. Unique checks should not be skipped.

The not null check on the other hand seems to be well skippable.

If feature with fid=NULL is sent to the constraint check, it reports an invalid constraint.
If I add a feature with fid=NULL, it will happily generate a new fid for me.

PR #39391, opened 2020-10-15
correctly determine if variables are static in aggregate expression and filter, give @parent in generic aggregates

### LINKED ISSUE #33382: aggregates inside with_variable gives incorrect result
**Describe the bug**

When using any aggregate function (e.g. `minimum`) inside a `with_variable`, the result seems to be wrongly cached, even if it's not cachable.

**How to Reproduce**

- Create a layer with an `id` column,  which can contain any value (as long as it's unique per row)
- Filter on `"id" = with_variable('outer_id', "id", minimum("id", filter := "id" = @outer_id))` (guess who only just learned of `@parent`, lol)
- Notice that, while the filter expression should be a noop, the only result is the first row.

**QGIS and OS versions**

QGIS version | 3.10.0-A Coruña | QGIS code branch | Release QGIS code branch.3
-- | -- | -- | --
Compiled against GDAL/OGR | 3.0.1 | Running against GDAL/OGR | 3.0.1
Compiled against GEOS | 3.7.3-CAPI-1.11.3 | Running against GEOS | 3.7.3-CAPI-1.11.3 b50468f
Compiled against SQLite | 3.30.0 | Running against SQLite | 3.30.0
PostgreSQL Client Version | 11.5 | SpatiaLite Version | 4.3.0a
QWT Version | 6.1.2 | QScintilla2 Version | 2.11.2
Compiled against PROJ | 6.1.1 | Running against PROJ | Rel. 6.1.1, July 1st, 2019
OS Version | NixOS 20.03pre-git (Markhor)
Active python plugins | pdokservicesplugin; db_manager; processing

3nids:
fixes #33382

### DIFF
--- a/src/core/expression/qgsexpressionfunction.cpp
+++ b/src/core/expression/qgsexpressionfunction.cpp
@@ -907,7 +907,7 @@ static QVariant fcnAggregateGeneric( QgsAggregateCalculator::Aggregate aggregate
   QString cacheKey;
   if ( !isStatic )
   {
-    cacheKey = QStringLiteral( "aggfcn:%1:%2:%3:%4:%5%6:%7" ).arg( vl->id(), QString::number( aggregate ), subExpression, parameters.filter,
+    cacheKey = QStringLiteral( "agg:%1:%2:%3:%4:%5%6:%7" ).arg( vl->id(), QString::number( aggregate ), subExpression, parameters.filter,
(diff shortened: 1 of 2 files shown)


nirvn (approved):
Nice and elegant.

nyalldawson (approved):

PR #39392, opened 2020-10-15
Geometry fixer fixes

olivierdalang:
## Description

This PR adds some fixes in the topology fixer logic (that is hidden by default, but used in the plugins)  :
- fixes an issue where trying to fix gaps would have no effect (happened when the gap was defined by a non-shared vertex on the edge or inside a neighbouring polygon)
- fixed an issue where trying to fix overlaps would have no effect (happened because the tolerance used when first finding the overlap wasn't the same than when fixing the overlap, so that the initial overlap polygon wouldn't be found when fixing it)
- fixed an issue where all points of a geometry would be "snapped to grid" when fixing an overlap, even points far away from the overlap (happened because the tolerance set to detect the overlaps was also used to modify the geometry)

There's many more issues (just look at the tracker :-) ) including in the provided overlap fixer test, where a sliver is left after fixing the overlap.
It's not easy to fix if we want to keep the ability to fix each part of an overlap independently (as an overlap part has shared edges with the base geom). But I'm not really sure of the current state of the geometry/topology checker plugin VS core feature, so I suggest to treat that separately (in what we shall call a *geometry fixer fixes fix PR* :-) )


src/analysis/vector/geometry_checker/qgsgeometryoverlapcheck.cpp
```diff
@@ -153,7 +153,9 @@ void QgsGeometryOverlapCheck::fixError( const QMap<QString, QgsFeaturePool *> &f
   }
   else if ( method == Subtract )
   {
-    std::unique_ptr< QgsAbstractGeometry > diff1( geomEngineA->difference( interPart, &errMsg ) );
+    std::unique_ptr< QgsGeometryEngine > geomEngineDiffA = QgsGeometryCheckerUtils::createGeomEngine( geometryA.constGet(), 0 );
+    geomEngineDiffA->prepareGeometry();
```
nyalldawson, line 157: There's no benefit to this -- the GEOS difference method doesn't take advantage of prepared geometries, so it's just extra work to prepare them.

src/analysis/vector/geometry_checker/qgsgeometryoverlapcheck.cpp
```diff
@@ -162,9 +164,9 @@ void QgsGeometryOverlapCheck::fixError( const QMap<QString, QgsFeaturePool *> &f
     {
       QgsGeometryCheckerUtils::filter1DTypes( diff1.get() );
     }
-    std::unique_ptr< QgsGeometryEngine > geomEngineB = QgsGeometryCheckerUtils::createGeomEngine( geometryB.constGet(), mContext->reducedTolerance );
-    geomEngineB->prepareGeometry();
-    std::unique_ptr< QgsAbstractGeometry > diff2( geomEngineB->difference( interPart, &errMsg ) );
+    std::unique_ptr< QgsGeometryEngine > geomEngineDiffB = QgsGeometryCheckerUtils::createGeomEngine( geometryB.constGet(), 0 );
+    geomEngineDiffB->prepareGeometry();
```
nyalldawson, line 168: same here

tests/src/geometry_checker/testqgsgeometrychecks.cpp
```diff
@@ -1240,7 +1243,11 @@ void TestQgsGeometryChecks::testOverlapCheckToleranceBug()
   QgsFeature f;
   testContext.second[layers["overlap_layer_tolerance_bug.shp"]]->getFeature( 0, f );
   double areaOld = f.geometry().area();
+  QgsPoint pointOld_1 = f.geometry().vertexAt( 1 );
+  QgsPoint pointOld_2 = f.geometry().vertexAt( 2 );
   QCOMPARE( areaOld, 10442.710061549426 );
+  QCOMPARE( pointOld_1, QgsPoint( 2537221.53079314017668366, 1152360.02460834058001637 ) );
```
nyalldawson, line 1249: Can you swap this to QGSCOMPARENEAR and add some tolerance to the test? Testing this high precision looks fragile to me.

src/analysis/vector/geometry_checker/qgsgeometrygapcheck.cpp
```diff
@@ -367,19 +367,39 @@ bool QgsGeometryGapCheck::mergeWithNeighbor( const QMap<QString, QgsFeaturePool
...
-  QgsSpatialIndex neighbourIndex( QgsSpatialIndex::Flag::FlagStoreFeatureGeometries );
-  neighbourIndex.addFeatures( neighbours );
 
   QgsPolyline snappedRing;
   QgsVertexIterator iterator = errGeometry->vertices();
   while ( iterator.hasNext() )
   {
     QgsPoint pt = iterator.next();
     QgsVertexId id;
-    QgsGeometry closestGeom = neighbourIndex.geometry( neighbourIndex.nearestNeighbor( QgsPointXY( pt ) ).first() );
```
nyalldawson, line 379: Dropping the index is going to have large impacts on the speed of this test. Is there any way you could retain the index but make it instead store the boundaries of polygons instead, so that the nearest neighbour test will consider the distance from rings?

olivierdalang: For robust results, I think we'd have to build the spatial index on vertices, in which case I dont think there's a performance gain in using an index ? (it still somehow needs to iterate through all vertices).

The issue is that if we are checking a point that lies exactly on an other polygon's edge, the distance to the ring may very well be 0, that doesn't mean this ring is the one with the closest vertex (this false assumption was the cause of the bug being fixed).

So we'd need to check not only against the nearest feature, but against N nearest features, probably within a search radius R, and then still search the closest vertex on these features... I'm not sure what to use for N and R, and think in the particular context of the geometry checker, where it's quite uncommon to have gaps between more than 3-4 features, it would end up making no real difference at the cost of some complex logic.

What do you think ?

At the very least, I replaced `distance` with `distanceSquared` to save one operation per vertex.

nyalldawson:
Looks good to me!

@m-kuhn I'd appreciate your review here too, you know this code better than I do...

m-kuhn (approved):

PR #39397, opened 2020-10-15
Maintain range label in rule based legend

### LINKED ISSUE #28011: Rule-based symbology legend values
Author Name: **Josh Waters** (Josh Waters)
Original Redmine Issue: [20190](https://issues.qgis.org/issues/20190)

Redmine category:map_legend

---
Under rule-based symbology, if I add ranges to the rule, it auto-populates legend entries.  Those seem to get lost however once applied to the rule based symbology.  There is no field there for legend, and on the print layout screen, the legend entries take the text of the rule itself rather than those aforementioned auto-populated legend entries.  Would it make sense to have a legend field in the rule-based symbology and allow those entries to carry over?

roya0045:
## Description

Proposal to fix #28011 

Should keep the existing label if it exists, otherwise use the rule based expression as a label. 

Unsure if this should be managed by a tickbox or if this is considered a direct upgrade. I'm open to implementing a gui control to this.


nyalldawson (approved):

nyalldawson:
Looks good, thanks!

Issue #39385, opened 2020-10-15
Make 'Derived' and 'Actions' optional in the Identify Results Tree panel


rduivenvoorde:
When using the i-tool (identify) on several layers, in the 'Top Down' mode, the attribute key/values will be presented in Identify Results panel.
Besides the actual attributes of a features QGIS currently also adds (sometimes A LOT) of so called 'Derived' attributes, like clicked xy, closest xy, parts and parts number etc etc. 

I propose to create a (in profile saved) setting, with which you either show or hide this Derived Attributes.

To enable/disable, there could be:
- add an item in the little Settings menu in the toolbar of the panel (the little tool button)

![Screenshot-20201015102053-365x88](https://user-images.githubusercontent.com/731673/96096422-24e66f00-0ed0-11eb-92cd-f2aa6b88f5e5.png)

- add some kind of little button in the little button bar of the panel (like the tool button)

- there is also an 'Identify'-section in the Settings/Map Tools tab, but I do not think this should be hidden for users, it should be easy to enable/disable it as users also find this information helpful
(see https://lists.osgeo.org/pipermail/qgis-developer/2020-October/062391.html)

- we could make 2 options, one to hide the derived ones and one to hide the actions, but to start (and to me) it is ok to do this in one go. 
Another option would be to make these attributes not so major, as in put them on bottom of the list, or grey them out.

Below is a quick test, to see what happens if I disable the derived attributes.
These are two vector layers, a point and a polygon (left is with and right is without derived info):

![Screenshot-20201014172526-868x856](https://user-images.githubusercontent.com/731673/96096079-b86b7000-0ecf-11eb-8693-45d9b8faa153.png)

Note that there are 4 places in code (for vector, raster, mesh and vector tile layers) to fix this.

Below a screenie of a identify panel with all four types of layers in it:

![Screenshot-20201015101331-453x1216](https://user-images.githubusercontent.com/731673/96097301-13519700-0ed1-11eb-97ed-2427de003723.png)

jmckenna:
Personally I constantly check my settings in /Settings/Options/ and I honestly had no idea that the identify panel has its own settings that can only be accessed with the little tool icon at the top of the identify panel (I have never noticed that icon before).   

I'd prefer that this setting be available both in /Settings/Options/Map Tools/, as well as directly in the panel settings.

jmckenna:
[...]

PR #39399, opened 2020-10-15
Adding a Identify Result setting to hide the Derived attributes

rduivenvoorde:
See #39385. This adds boolean setting so the user can choose to show or hide the derived attribute values.
For certain use cases, OR when the user is not interested in all these geometry related info, this derived info is too much.

Setting looks like:

Screenshot-20201015213836-340x84

The difference (in project with all 4 types of layers that have feature info):

Screenshot-20201015214037-1135x1207

I was uncertain how to handle the Mesh layer, as it looks like it receives 2 'features' for one click: one with the actual 'data' and one with a 'Geometry'. BOTH have a 'derived attributes', but I think the one for the data (ETA in this case) is actually valuable so keeping that one, but hiding the whole geometry stuff (see screenshot layer 1/mercator_test_puff-123).

One thing that bothers me also is that raster and mesh layers do not have their layernames (in bold) as root-node of the feature.
Happy to change/fix that too in this PR. I do not think it is helpfull to see the layer number (or is it feature number) in the result list?

### DIFF
--- a/src/app/qgsidentifyresultsdialog.cpp
+++ b/src/app/qgsidentifyresultsdialog.cpp
@@ -334,6 +334,7 @@ QgsIdentifyResultsDialog::QgsIdentifyResultsDialog( QgsMapCanvas *canvas, QWidge
   connect( mCollapseAction, &QAction::triggered, this, &QgsIdentifyResultsDialog::mCollapseAction_triggered );
   connect( mActionCopy, &QAction::triggered, this, &QgsIdentifyResultsDialog::mActionCopy_triggered );
   connect( mActionAutoFeatureForm, &QAction::toggled, this, &QgsIdentifyResultsDialog::mActionAutoFeatureForm_toggled );
+  connect( mActionHideDerivedAttributes, &QAction::toggled, this, &QgsIdentifyResultsDialog::mActionHideDerivedAttributes_toggled );
 
   mOpenFormAction->setDisabled( true );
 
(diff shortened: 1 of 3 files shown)


src/app/qgsidentifyresultsdialog.cpp
```diff
@@ -1130,6 +1134,12 @@ void QgsIdentifyResultsDialog::addFeature( QgsMeshLayer *layer,
     connect( layer, &QgsMapLayer::crsChanged, this, &QgsIdentifyResultsDialog::layerDestroyed );
   }
 
+  // for Mesh layers it looks like it is best to hide the 'Geometry' feature, but keep the actual 'derived' attributes
+  if ( label == "Geometry" && QgsSettings().value( QStringLiteral( "/Map/hideDerivedAttributes" ), false ).toBool() )
```
nyalldawson, line 1138: This won't work for non-english installs. 
```suggestion
  if ( label == tr( "Geometry" ) && QgsSettings().value( QStringLiteral( "/Map/hideDerivedAttributes" ), false ).toBool() )
```

src/ui/qgsidentifyresultsbase.ui
```diff
@@ -366,12 +366,23 @@
...
    <property name="text">
     <string>Auto open form for single feature results</string>
    </property>
+  </action>
+  <action name="mActionHideDerivedAttributes">
    <property name="checkable">
     <bool>true</bool>
    </property>
+   <property name="text">
+    <string>Hide derived attributes from results</string>
```
nyalldawson, line 381:
```suggestion
    <string>Hide Derived Attributes from Results</string>
```

nyalldawson:
Looks good to me!

src/app/qgsidentifyresultsdialog.cpp
rduivenvoorde: Ah thanks. Yes I was wondering if such words were translated or not after my commit...

rduivenvoorde:
Thanks @nyalldawson !

jmckenna:
Great enhancement, thanks @rduivenvoorde !

nyalldawson (approved):

nyalldawson:
First new feature for 3.18! Congratulations @rduivenvoorde !!

rduivenvoorde:
\o/

PR #39400, opened 2020-10-15
add field icon in DnD tree

3nids:
image

Really useful when you don't want to distinguish joined fields, or more generally to show the field type.

Can this be seen as bugfix?

### DIFF
--- a/src/core/qgsfields.h
+++ b/src/core/qgsfields.h
@@ -321,9 +321,11 @@ class CORE_EXPORT  QgsFields
 
     /**
      * Returns an icon corresponding to a field index, based on the field's type and source
+     * \param fieldIdx the field index
+     * \param considerOrigin, if True the icon will the origin of the field
      * \since QGIS 2.14
      */
-    QIcon iconForField( int fieldIdx ) const SIP_FACTORY;
+    QIcon iconForField( int fieldIdx, bool considerOrigin = false ) const SIP_FACTORY;
 #ifdef SIP_RUN
     % MethodCode
     if ( a0 < 0 || a0 >= sipCpp->count() )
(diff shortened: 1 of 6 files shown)


nyalldawson:
Yes -- great idea!!

(I wonder if we could get icons from qt designer for the widget side too...)

src/core/qgsfields.h
```diff
@@ -321,9 +321,11 @@ class CORE_EXPORT  QgsFields
 
     /**
      * Returns an icon corresponding to a field index, based on the field's type and source
+     * \param fieldIdx the field index
+     * \param considerOrigin, if True the icon will the origin of the field
```
nyalldawson, line 325:
```suggestion
     * \param considerOrigin if TRUE the icon will the origin of the field
```

elpaso:
Cool! Thank you.

nyalldawson (approved):

src/gui/vector/qgsattributesformproperties.cpp
```diff
@@ -97,12 +97,12 @@ void QgsAttributesFormProperties::initAvailableWidgetsTree()
   for ( int i = 0; i < fields.size(); ++i )
   {
     const QgsField field = fields.at( i );
-    DnDTreeItemData itemData = DnDTreeItemData( DnDTreeItemData::Field, field.name(), field.name() );
+    DnDTreeItemData itemData = DnDTreeItemData( DnDTreeItemData::Field, field.name(), field.name(), QColor() );
```
3nids, line 100:
```suggestion
    DnDTreeItemData itemData = DnDTreeItemData( DnDTreeItemData::Field, field.name(), field.name() );
```

PR #39402, opened 2020-10-15
PostgreSQL 12 + PostGIS 3 + GEOS 3.8.1 + Python 3.8.6 tests

### LINKED ISSUE #39401: PostgreSQL tests failing with PostGIS 3.0.2 + GEOS 3.8.1 + Python
I've noticed that the QGIS tests have been failing lately against PostgreSQL + PostGIS 3.0.2 + GEOS 3.8.1 + Python 3.8.6:

```
======================================================================
FAIL: testGetFeaturesExp (__main__.TestPyQgsPostgresProvider)
----------------------------------------------------------------------
  File "/home/jose/software/qgis/qgis-repo/QGIS/tests/src/python/providertestbase.py", line 167, in testGetFeaturesExp
    self.runPolyGetFeatureTests(self.poly_provider)
  File "/home/jose/software/qgis/qgis-repo/QGIS/tests/src/python/providertestbase.py", line 141, in runPolyGetFeatureTests
    self.assert_query(provider,
  File "/home/jose/software/qgis/qgis-repo/QGIS/tests/src/python/providertestbase.py", line 80, in assert_query
    FeatureSourceTestCase.assert_query(self, source, expression, expected)
  File "/home/jose/software/qgis/qgis-repo/QGIS/tests/src/python/featuresourcetestbase.py", line 146, in assert_query
    assert set(expected) == result, 'Expected {} and got {} when testing expression "{}"'.format(set(expected),
AssertionError: Expected {1, 3} and got set() when testing expression "overlaps(buffer($geometry,1),geom_from_wkt( 'Polygon ((-75.1 76.1, -75.1 81.6, -68.8 81.6, -68.8 76.1, -75.1 76.1))'))"

```
Also, tests failing because of irrelevant rounding errors:

```
======================================================================
FAIL: testGeneratedFields (__main__.TestPyQgsPostgresProvider)
Test if GENERATED geometry/geography columns are correctly handled by the provider.
----------------------------------------------------------------------
  File "/home/jose/software/qgis/qgis-repo/QGIS/tests/src/python/test_provider_postgres.py", line 557, in testGeneratedFields
    self.assertEqual(f2['poly_area'], expected_area)
AssertionError: 43069568296.34386 != 43069568296.34387


```

I've asked the qgis-developer mailing list about this, but no one replied:

https://lists.osgeo.org/pipermail/qgis-developer/2020-September/062316.html

I'm preparing a simple PR that fixes these issues and will allow us to keep supporting these newer versions of the software chain.

QGIS version | 3.15.0-Master | QGIS code revision | 69faea81d9
-- | -- | -- | --
Compiled against GDAL/OGR | 3.1.2 | Running against GDAL/OGR | 3.1.2
Compiled against SQLite | 3.33.0 | Running against SQLite | 3.33.0
PostgreSQL Client Version | 13.0 | SpatiaLite Version | 4.3.0a
QWT Version | 6.1.4 | QScintilla2 Version | 2.11.5
Compiled against PROJ | 7.1.1 | Running against PROJ | Rel. 7.1.1, September 1st, 2020
OS Version | Gentoo/Linux
Active python plugins | QuickWKT; MetaSearch; db_manager; processing; quick_map_services

espinafre:
I've noticed that lately the PostgreSQL tests have failing on me, as I've explained on #39401. One of the problems is a rounding error, which I've fixed by adding round() to 4 digits on the relevant PostgreSQL test cases; the other issue was a little harder.

I've bisected it to  250ce7615293b3970cc6a98ec925e2a22b1608a0, which introduced the value of 8.0 as the default parameter "segments" for the buffer() QGIS function. Since that literal is a floating point value, the PostgreSQL provider quotes it (per #37367), which makes the failing test generate the following SQL snippet:

`ST_Buffer("geom",1,'8')`

The PostGIS [documentation](https://postgis.net/docs/ST_Buffer.html) says that ST_Buffer has two signatures for geometry types:

```
geometry ST_Buffer(geometry g1, float radius_of_buffer, text buffer_style_parameters='');
geometry ST_Buffer(geometry g1, float radius_of_buffer, integer num_seg_quarter_circle);
```

Since the value '8' is passed quoted to the DBMS, PostGIS understands that it is a string, and the first form is called, which generates the error and ultimately causes the test to fail.

With this change, the value 8.0 is now simply 8, which is treated as an integer by the provider, and therefore not quoted, which generates the following SQL snippet:

`ST_Buffer("geom",1,8)`

which is the second form of the ST_Buffer function documented above, and what we really want. Also, it makes more sense that the number of segments in a quarter circle is integer!

Fixes  #39401

### DIFF
--- a/tests/src/python/test_provider_postgres.py
+++ b/tests/src/python/test_provider_postgres.py
@@ -907,7 +907,7 @@ class TestPyQgsPostgresProvider(unittest.TestCase, ProviderTestCase):
         gotten_attrs = [f4['pk1'], f4['pk2'], f4['pk3'], f4['value']]
         self.assertEqual(gotten_attrs[0], expected_attrs[0])
         self.assertEqual(gotten_attrs[1], expected_attrs[1])
-        self.assertAlmostEqual(gotten_attrs[2], expected_attrs[2])
+        self.assertAlmostEqual(gotten_attrs[2], expected_attrs[2], places=4)
         self.assertEqual(gotten_attrs[3], expected_attrs[3])
 
         # Finally, let's delete one of the features.
--- a/tests/src/python/test_provider_postgres.py
+++ b/tests/src/python/test_provider_postgres.py
@@ -554,7 +554,7 @@ class TestPyQgsPostgresProvider(unittest.TestCase, ProviderTestCase):
         expected_area = 43069568296.34387
 
         assert compareWkt(generated_geometry, expected_geometry), "Geometry mismatch! Expected:\n{}\nGot:\n{}\n".format(expected_geometry, generated_geometry)
-        self.assertEqual(f2['poly_area'], expected_area)
+        self.assertEqual(round(f2['poly_area'], 4), round(expected_area, 4))
         self.assertEqual(f2['name'], 'QGIS-3')
 
         # Checking if we can correctly change values of an existing feature.
@@ -582,7 +582,7 @@ class TestPyQgsPostgresProvider(unittest.TestCase, ProviderTestCase):
         # reading back
         vl2 = QgsVectorLayer('{} table="qgis_test"."{}" (geom) srid=4326 type=POLYGON key="id" sql='.format(self.dbconn, "test_gen_col"), "test_gen_col", "postgres")
         f2 = next(vl2.getFeatures(QgsFeatureRequest()))
-        self.assertEqual(f2['poly_area'], expected_area)
+        self.assertEqual(round(f2['poly_area'], 4), round(expected_area, 4))
 
         # now, getting a brand new QgsVectorLayer to check if changes (UPDATE) in the geometry are reflected in the generated fields
         vl = QgsVectorLayer('{} table="qgis_test"."{}" (geom) srid=4326 type=POLYGON key="id" sql='.format(self.dbconn, "test_gen_col"), "test_gen_col", "postgres")
@@ -603,7 +603,7 @@ class TestPyQgsPostgresProvider(unittest.TestCase, ProviderTestCase):
         expected_area = 67718478405.28429
 
         assert compareWkt(generated_geometry, expected_geometry), "Geometry mismatch! Expected:\n{}\nGot:\n{}\n".format(expected_geometry, generated_geometry)
-        self.assertEqual(f2['poly_area'], expected_area)
+        self.assertEqual(round(f2['poly_area'], 4), round(expected_area, 4))
         self.assertEqual(f2['name'], 'New')
 
         # Geography columns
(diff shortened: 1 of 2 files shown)


nyalldawson:
> I've bisected it to 250ce76, which introduced the value of 8.0 as the default parameter "segments" for the buffer() QGIS function. Since that literal is a floating point value, the PostgreSQL provider quotes it (per #37367), which makes the failing test generate the following SQL snippet:

Ewww -- #37367 is completely wrong in changing this! Instead of changing the expression function we need to partially rollback #37367 and ensure that double values AREN'T quoted in the postgres expression compiler. Are you able to do this?

tests/src/python/test_provider_postgres.py
```diff
@@ -554,7 +554,7 @@ def testGeneratedFields(self):
         expected_area = 43069568296.34387
 
         assert compareWkt(generated_geometry, expected_geometry), "Geometry mismatch! Expected:\n{}\nGot:\n{}\n".format(expected_geometry, generated_geometry)
-        self.assertEqual(f2['poly_area'], expected_area)
+        self.assertEqual(round(f2['poly_area'], 4), round(expected_area, 4))
```
nyalldawson, line 557: You can use self.assertAlmostEqual here instead, which will give nicer debug messages when the test fails

espinafre: I'll try this, thanks for the pointer!

espinafre:
> > I've bisected it to [250ce76](https://github.com/qgis/QGIS/commit/250ce7615293b3970cc6a98ec925e2a22b1608a0), which introduced the value of 8.0 as the default parameter "segments" for the buffer() QGIS function. Since that literal is a floating point value, the PostgreSQL provider quotes it (per #37367), which makes the failing test generate the following SQL snippet:
> 
> Ewww -- #37367 is completely wrong in changing this! Instead of changing the expression function we need to partially rollback #37367 and ensure that double values AREN'T quoted in the postgres expression compiler. Are you able to do this?

Yes, I certainly can do that; however, we lose the ability to use fixed point (NUMERIC in PostgreSQL) columns as primary keys, or as part of a composite primary key, as I've shown in issue #37245 . Are you okay with this? Shall I proceed? Thanks!

nyalldawson:
> . Are you okay with this? Shall I proceed? Thanks!

Not if we limit the changes from the earlier PR so that they don't apply to the postgres SQL expression compiler. This is only used for conversion of QGIS filter expressions to postgres expressions, so won't affect the PK related filtering code.

tests/src/python/test_provider_postgres.py
espinafre: Fixed by 9537f48

espinafre:
Hi, @nyalldawson , I couldn't find a way to leave floating point literals unquoted just for the expression compiler; it seems to me that it is an all-or-nothing proposition. There might be a way, but I didn't find it. As it is, this PR at least allows the tests to pass against PostgreSQL 12 (and now, against PostgreSQL 13) on my local systems.

Can we merge this while we further investigate or devise a form to separate expression compiler's statements from the rest of them?

nyalldawson (approved):

nyalldawson:
@espinafre I've fixed the expression compiler in https://github.com/qgis/QGIS/pull/40713

PR #39414, opened 2020-10-16
UUID expressions function with string format parameter

signedav:
In the expressions: `uuid()` used to generate the UUID in the given format returned by `QUuid::createUuid` with curly braces and dashes: `{0bd2f60f-f157-4a6d-96af-d4ba4cb366a1}`

This needed to be formatted inside the expression with `substr` etc. to get it without braces or dashes.

`QUuid::createUuid` supports a `stringFormat` parameter https://doc.qt.io/qt-5/quuid.html#StringFormat-enum

Now you can enter a parameter to the `uuid` function in the expressions:
```
uuid() -> {0bd2f60f-f157-4a6d-96af-d4ba4cb366a1}
uuid('WithoutBraces') -> 0bd2f60f-f157-4a6d-96af-d4ba4cb366a1
uuid('Id128') -> 0bd2f60ff1574a6d96afd4ba4cb366a1
```

### DIFF
--- a/resources/function_help/json/uuid
+++ b/resources/function_help/json/uuid
@@ -2,7 +2,22 @@
   "name": "uuid",
   "type": "function",
   "groups": ["Record and Attributes"],
-  "description": "Generates a Universally Unique Identifier (UUID) for each row using the Qt <a href='https://doc.qt.io/qt-5/quuid.html#createUuid'>QUuid::createUuid</a> method. Each UUID is 38 characters long.",
-  "examples": [ { "expression":"uuid()", "returns":"'{0bd2f60f-f157-4a6d-96af-d4ba4cb366a1}'"}
+  "description": "Generates a Universally Unique Identifier (UUID) for each row using the Qt <a href='https://doc.qt.io/qt-5/quuid.html#createUuid'>QUuid::createUuid</a> method.",
+  "variants": [
+    {
+      "variant": "No parameters",
+      "variant_description": "If called with no parameters, the function will generate the UUID with the default string format 'WithBraces'. Each UUID is 38 characters long.",
+      "arguments": [],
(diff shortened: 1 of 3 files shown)


DelazJ:
@signedav I'm not sure we need to document these as variants of the function. I can't talk code wise but from the help side, instead of showing variants, how about a single function of type `uuid([stringFormat='WithBraces'])` that returns a string format?:
- it takes an optional parameter
- the default value of that parameter is WithBraces, meaning that, ie `uuid()` and `uuid('WithBraces')` will return the uuid string with braces. The current behavior if I'm not wrong.
- using another value of the parameter (WithoutBraces or WithoutDashes) you get the new formatting

Or do I miss something in the implementation?

nyalldawson (commented):
Code looks good to me! Can you add some tests too?

signedav:
@DelazJ It's both fine for me. I wrote the help with the variants like I've seen it for other expressions like e.g. `display_expression`. I thought this is the way we do it. But the way you described is done for e.g. `num_selected`. So I have no strong opinion. According to what we do it one way or another?

(Fine to change it to "format").

DelazJ:
> I wrote the help with the variants like I've seen it for other expressions like e.g. display_expression. I thought this is the way we do it. But the way you described is done for e.g. num_selected. 

Yes, to me this function behavior is more like num_selected or somehow more like [to_dm](https://docs.qgis.org/3.16/en/docs/user_manual/working_with_vector/functions_list.html#to-dm), to_dms; it's the result "format" that changes, not really what is evaluated or what is returned. IMHO, we should keep variants for situations where the parameters are different or lead to very different values.
Also adding variants double the length of the help (and, just my feeling, the current formatting does not make them easy to differentiate)
>(Fine to change it to "format").

:+1:

signedav:
Thanks for the explanation @DelazJ I changed it to the way you suggested.

nyalldawson (approved):

signedav:
Thanks @nyalldawson and @DelazJ

PR #39415, opened 2020-10-16
[processing] PointsToPath allow expression for order fields

m-kuhn:
Allows using `$id` as expression for a csv with ordered values

Adds a new parameter `ORDER_EXPRESSION` to the PointsToPath algorithm.

### DIFF
--- a/python/plugins/processing/algs/qgis/PointsToPaths.py
+++ b/python/plugins/processing/algs/qgis/PointsToPaths.py
@@ -24,7 +24,8 @@ __copyright__ = '(C) 2014, Alexander Bruy'
 import os
 from datetime import datetime
 
-from qgis.core import (QgsFeature,
+from qgis.core import (QgsExpression,
+                       QgsFeature,
                        QgsFeatureSink,
                        QgsFields,
                        QgsField,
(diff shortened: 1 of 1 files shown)


python/plugins/processing/algs/qgis/PointsToPaths.py
```diff
@@ -73,8 +76,8 @@ def initAlgorithm(self, config=None):
                                                               self.tr('Input point layer'), [QgsProcessing.TypeVectorPoint]))
         self.addParameter(QgsProcessingParameterBoolean(self.CLOSE_PATH,
                                                         self.tr('Close path'), defaultValue=False))
-        self.addParameter(QgsProcessingParameterField(self.ORDER_FIELD,
-                                                      self.tr('Order field'), parentLayerParameterName=self.INPUT))
+        self.addParameter(QgsProcessingParameterExpression(self.ORDER_FIELD,
```
nyalldawson, line 79: Unfortunately this could potentially break models/scripts -- e.g. if a script is using a hardcoded field name with a space in it, then it won't automatically become a valid QGIS expression. I think you'll need to make a new parameter here and hide the old one.

nyalldawson (approved):

roya0045:
@m-kuhn Good addition! I usually have to add the order tool in my model, this should simplify them slightly.

nyalldawson:
@m-kuhn any chance you feel like porting this one too C++ prior to merge?

m-kuhn:
It's less the "feel like" and more the "time constraint" that prevents me from putting more into this.

nyalldawson:
Fair enough. I just want the algorithm ported before https://github.com/qgis/QGIS/pull/38701 is merged, as the results from that one will likely change between a python and C++ version.

nyalldawson (approved):

PR #39419, opened 2020-10-16
FIX: QgsCoordinateReferenceSystem::createFrom* has to return CRS's validity

rldhont:
## Description

The methods `QgsCoordinateReferenceSystem::createFrom*` returned true when the CRS was found in cache instead of the CRS's validity.

This fixed it and the tests.

It replaces https://github.com/qgis/QGIS/pull/39343 where I found that `createFromWKt` can return `True` even if the WKT does not represent a valid CRS for Proj4.

### DIFF
--- a/tests/src/core/testqgscoordinatereferencesystem.cpp
+++ b/tests/src/core/testqgscoordinatereferencesystem.cpp
@@ -345,7 +345,7 @@ void TestQgsCoordinateReferenceSystem::ogcWmsCrsCache()
   QVERIFY( QgsCoordinateReferenceSystem::ogcCache().contains( QStringLiteral( "not a CRS" ) ) );
   // a second time, so invalid crs is fetched from cache
   QgsCoordinateReferenceSystem crs4;
-  QVERIFY( crs4.createFromOgcWmsCrs( QStringLiteral( "not a CRS" ) ) );
+  QVERIFY( !crs4.createFromOgcWmsCrs( QStringLiteral( "not a CRS" ) ) );
   QVERIFY( !crs4.isValid() );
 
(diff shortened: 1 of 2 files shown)


nyalldawson (approved):

tests/src/core/testqgscoordinatereferencesystem.cpp
```diff
@@ -327,21 +327,25 @@ void TestQgsCoordinateReferenceSystem::fromOgcWmsCrs()
...
   // invalid
-  QgsCoordinateReferenceSystem crs3 = QgsCoordinateReferenceSystem::fromOgcWmsCrs( QStringLiteral( "not a CRS" ) );
+  QgsCoordinateReferenceSystem crs3;
+  QVERIFY( !crs3.createFromOgcWmsCrs( QStringLiteral( "not a CRS" ) ) );
   QVERIFY( !crs3.isValid() );
   QVERIFY( QgsCoordinateReferenceSystem::ogcCache().contains( QStringLiteral( "not a CRS" ) ) );
   // a second time, so invalid crs is fetched from cache
-  QgsCoordinateReferenceSystem crs4 = QgsCoordinateReferenceSystem::fromOgcWmsCrs( QStringLiteral( "not a CRS" ) );
+  QgsCoordinateReferenceSystem crs4;
+  QVERIFY( crs4.createFromOgcWmsCrs( QStringLiteral( "not a CRS" ) ) );
```
rldhont, line 348:
```suggestion
  QVERIFY( !crs4.createFromOgcWmsCrs( QStringLiteral( "not a CRS" ) ) );
```

PR #39421, opened 2020-10-16
make the SVG selector collapsible (+ remove duplicated code)

3nids:
defer loading of SVG icons to only show them when the widget is made visible

this is mainly code moving to its own-class.
this is quite an improvement to me in the general fluidity when working in projects with many SVGs.
it also make the UI much less cluttered.

I'd consider this as a bug fix.

(It takes about 3 seconds to show the widget as the images are loaded, I get a waiting cursor on my OS which is apparently not visible in the video)
svg


nyalldawson:
The browser should be loading all the svg previews in a seperate thread. At least it was this way in the past.to avoid any delays with this widget. Is something broken with the thread?

3nids:
> The browser should be loading all the svg previews in a seperate thread. 

I realised there were some duplicate code. 
Now,

* I removed the class I created (`QgsSvgBrowserWidget`)
*  the symbol layer widget is using the dedicated `QgsSvgSelector`
* in `QgsSvgSelector`, there is now a collapsible group box for the tree
* in `QgsAbstractFileContentSourceLineEdit`, the tool button is now a `QgsPropertyOverrideButton`

> At least it was this way in the past.to avoid any delays with this widget. Is something broken with the thread?

yes this seems broken.

3nids:
> At least it was this way in the past.to avoid any delays with this widget. Is something broken with the thread 

This is actually the initial filling with all SVG icons 
https://github.com/qgis/QGIS/blob/29b76f44b11fc0ff2d3bf95a85bf36ece093b4e2/src/gui/symbology/qgssvgselectorwidget.cpp#L477-L481

I replaced the initial model with a path, and there is no blocking anymore.

I would propose to 

1. merge this PR as it's a both a UX improvement (collapsible groupbox) and code duplication removal
2. discuss the solutions for the loading (do not load all at first, specify a limit, load the images only when scrolling,…)

src/gui/symbology/qgssymbollayerwidget.cpp
```diff
@@ -2567,12 +2513,13 @@ void QgsSvgMarkerSymbolLayerWidget::setSymbolLayer( QgsSymbolLayer *layer )
   registerDataDefinedButton( mStrokeWidthDDBtn, QgsSymbolLayer::PropertyStrokeWidth );
   registerDataDefinedButton( mAngleDDBtn, QgsSymbolLayer::PropertyAngle );
   registerDataDefinedButton( mOffsetDDBtn, QgsSymbolLayer::PropertyOffset );
-  registerDataDefinedButton( mFilenameDDBtn, QgsSymbolLayer::PropertyName );
   registerDataDefinedButton( mFillColorDDBtn, QgsSymbolLayer::PropertyFillColor );
   registerDataDefinedButton( mStrokeColorDDBtn, QgsSymbolLayer::PropertyStrokeColor );
   registerDataDefinedButton( mHorizontalAnchorDDBtn, QgsSymbolLayer::PropertyHorizontalAnchor );
   registerDataDefinedButton( mVerticalAnchorDDBtn, QgsSymbolLayer::PropertyVerticalAnchor );
 
+  registerDataDefinedButton( mSvgSelectorWidget->sourceLineEdit()->fileToolButton(), QgsSymbolLayer::PropertyVerticalAnchor );
```
nirvn, line 2521:
```suggestion
  registerDataDefinedButton( mSvgSelectorWidget->sourceLineEdit()->fileToolButton(), QgsSymbolLayer::PropertyName );
```

nirvn (commented):
One suggested change. Otherwise I'm good with seeing this merged, good cleanup.

PR #39424, opened 2020-10-16
fix #39068 : geometry_overlays's limit parameter not working as expected

### LINKED ISSUE #39068: Wrong limit parameter behavior for geometry_overlay_intersects function
If I use `limit:=2`, then I get the expected value (a single feature)
image
If I use `limit:=1`,it's expected to return that same value but it instead returns an empty list (and not coherent wit the docs in the third panel (the limit is supposed to be included)
image

But if I use limit with the geometry_overlay_disjoint, it works as expected
image

QGIS version | 3.15.0-Master | QGIS code revision | 0fb7c44645
-- | -- | -- | --
Compiled against GDAL/OGR | 3.2.0dev | Running against GDAL/OGR | 3.2.0dev
Compiled against PROJ | 7.2.0 | Running against PROJ | Rel. 7.2.0, November 1st, 2020
OS Version | Windows 10 (10.0) | This copy of QGIS writes debugging output.
Active python plugins |

olivierdalang:
## Description

This fixes #39068 , making the limit parameter for geometry_overlay_intersects (and similar) functions behave as expected.

The issue happened because the limit parameter was used to limit the feature request BEFORE the spatial operator check.

By the way, this also improves behavior when `cache=true`, as before that, the limit parameter was applied to the materialized spatial index (hence being quite useless).


nyalldawson (commented):
Looks good! Could we get a test for this too?

olivierdalang:
Ok added a test, I think we can merge if travis is happy

PR #39426, opened 2020-10-16
ReadEntry ok value must be set to False when property doesn't exist

### LINKED ISSUE #37851: readEntry and readBoolEntry always report found
**Describe the bug**

readEntry and readBoolEntry alwys report found as true.    readNumEntry does not

**How to Reproduce**

Start QGIS.  Do not load a project file
Open the Python Console and type the following commands (responses in italic)
proj = QgsProject.instance() 
proj.readBoolEntry('', 'xxx', False) 
*(False, True)*
proj.readEntry('', 'xxx', 'xx') 
*('xx', True)*
proj.readNumEntry('', 'xxx', -1) 
*(-1, False)*

**QGIS and OS versions**

QGIS version | 3.10.7-A Coruña | QGIS code revision | 7b4ca4c8d0
-- | -- | -- | --
OS Version | Windows 10 (10.0)
Active python plugins | plugin_reloader; QSWAT3; QSWATPlus3_64; remotedebug; db_manager; MetaSearch; processing

**Additional context**

troopa81:
Fixes #37851 : fix readXXXEntry methods, add few tests and complete documentation

### DIFF
--- a/src/core/qgsproject.h
+++ b/src/core/qgsproject.h
@@ -498,35 +498,35 @@ class CORE_EXPORT QgsProject : public QObject, public QgsExpressionContextGenera
      * keys would be the familiar QgsSettings-like '/' delimited entries,
      * implying a hierarchy of keys and corresponding values
      */
-    QStringList readListEntry( const QString &scope, const QString &key, const QStringList &def = QStringList(), bool *ok = nullptr ) const;
+    QStringList readListEntry( const QString &scope, const QString &key, const QStringList &def = QStringList(), bool *ok = nullptr SIP_OUT ) const;
 
     /**
+     * \param def returned value if key doesn't exist
+     * \param ok set to TRUE if key exists and has been successfully retrieved
      * \returns entry value as string from \a scope given its \a key
-     * \param def returned value if key doesn't exist
-     * \param ok set to TRUE if key exists and has been successfully retrieved
      */
-    QString readEntry( const QString &scope, const QString &key, const QString &def = QString(), bool *ok = nullptr ) const;
+    QString readEntry( const QString &scope, const QString &key, const QString &def = QString(), bool *ok = nullptr SIP_OUT ) const;
 
     /**
+     * \param def returned value if key doesn't exist
+     * \param ok set to TRUE if key exists and has been successfully retrieved
      * \returns entry value as integer from \a scope given its \a key
-     * \param def returned value if key doesn't exist
-     * \param ok set to TRUE if key exists and has been successfully retrieved
      */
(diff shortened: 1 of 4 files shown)


src/core/qgsproject.h
```diff
@@ -500,9 +500,32 @@ class CORE_EXPORT QgsProject : public QObject, public QgsExpressionContextGenera
      */
     QStringList readListEntry( const QString &scope, const QString &key, const QStringList &def = QStringList(), bool *ok = nullptr ) const;
 
+    /**
+     * \returns entry value as string from \a scope given its \a key
+     * \param def returned value if key doesn't exist
+     * \param ok set to TRUE if key exists and has been successfully retrieved
+     */
     QString readEntry( const QString &scope, const QString &key, const QString &def = QString(), bool *ok = nullptr ) const;
```
nyalldawson, line 508: Can you mark ok as "SIP_OUT" in here so that the python docs are correctly generated?

src/core/qgsproject.h
```diff
@@ -500,9 +500,32 @@ class CORE_EXPORT QgsProject : public QObject, public QgsExpressionContextGenera
      */
     QStringList readListEntry( const QString &scope, const QString &key, const QStringList &def = QStringList(), bool *ok = nullptr ) const;
 
+    /**
+     * \returns entry value as string from \a scope given its \a key
```
nyalldawson, line 504: Returns needs to come after the param annotations so that sipify can do its full magic

src/core/qgsproject.h
m-kuhn: Isn't that an API change?

nyalldawson: Doesn't seem to be - check the test results from this pr, sip is already treating ok as a return argument. It's just sipify which needs the hints for the documentation.

troopa81: indeed, there is no API change. I added SIP_OUT

src/core/qgsproject.h
troopa81: Thanks for pointing that out. I did the modifications

troopa81:
I fixed readEntryList also

nyalldawson (approved):

nyalldawson:
@m-kuhn this pr shows a number of warnings in the documentation action (not all related to this pr, some have slipped in through other prs). These used to error out the builds, but don't anymore. Any ideas why?

m-kuhn:
Good question. Do you have a link to those warnings (are they on the CI)?
Otherwise we could add a new job to the workflows, dedicated to check api docs only

PR #39436, opened 2020-10-16
[Server] Fix + in parameter value getting converted to space if received via POST

manisandro:
For example, if a layer name contains a +, and you submit a GetMap request per POST (x-www-form-urlencoded), the plus will be correctly handled at [1], but then in `QgsServerParameters::add` [2] and `QgsServerParameters::urlQuery` [3] it will be treated as a space, see the note for `QUrlQuery::addQueryItem` here [4]. So i.e. a layename `hello+world` will become `hello world`

[1] https://github.com/qgis/QGIS/blob/master/src/server/qgsrequesthandler.cpp#L224
[2] https://github.com/qgis/QGIS/blob/master/src/server/qgsserverparameters.cpp#L430
[3] https://github.com/qgis/QGIS/blob/master/src/server/qgsserverparameters.cpp#L444
[4] https://doc.qt.io/qt-5/qurlquery.html#addQueryItem

### DIFF
--- a/src/server/qgsserverparameters.cpp
+++ b/src/server/qgsserverparameters.cpp
@@ -427,7 +427,7 @@ void QgsServerParameters::save( const QgsServerParameter &parameter )
 void QgsServerParameters::add( const QString &key, const QString &value )
 {
   QUrlQuery query;
-  query.addQueryItem( key, value );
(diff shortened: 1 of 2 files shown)


nyalldawson:

PR #39439, opened 2020-10-17
QgsVectorLayerExporter: defer GPKG spatial index creation at file closing

rouault:
Should fix the issue raised in
https://lists.osgeo.org/pipermail/qgis-developer/2020-October/062422.html

The difference with QgsVectorFileWriter is that QgsVectorFileWriter does not
close and re-open the layer, so the defered spatial index creation mechanism
that is built-in in the GPKG driver works. But in the case of
QgsVectorLayerExporter, we close and re-open and thus the spatial index gets
created immediately.

### DIFF
--- a/src/core/providers/ogr/qgsogrprovider.cpp
+++ b/src/core/providers/ogr/qgsogrprovider.cpp
@@ -2710,13 +2710,21 @@ bool QgsOgrProvider::deleteFeatures( const QgsFeatureIds &id )
     returnvalue = false;
   }
 
-  if ( mFeaturesCounted != QgsVectorDataProvider::Uncounted &&
-       mFeaturesCounted != QgsVectorDataProvider::UnknownCount )
+  if ( mGDALDriverName == QLatin1String( "ESRI Shapefile" ) )
   {
-    if ( returnvalue )
-      mFeaturesCounted -= id.size();
-    else
-      recalculateFeatureCount();
+    // Shapefile behaves in a special way due to possible recompaction
+    recalculateFeatureCount();
+  }
+  else
+  {
+    if ( mFeaturesCounted != QgsVectorDataProvider::Uncounted &&
+         mFeaturesCounted != QgsVectorDataProvider::UnknownCount )
+    {
+      if ( returnvalue )
+        mFeaturesCounted -= id.size();
+      else
+        recalculateFeatureCount();
+    }
   }
 
   clearMinMaxCache();
--- a/src/core/providers/ogr/qgsogrprovider.cpp
+++ b/src/core/providers/ogr/qgsogrprovider.cpp
@@ -1875,7 +1875,14 @@ bool QgsOgrProvider::addFeatures( QgsFeatureList &flist, Flags flags )
     returnvalue = false;
   }
 
-  recalculateFeatureCount();
+  if ( mFeaturesCounted != QgsVectorDataProvider::Uncounted &&
+       mFeaturesCounted != QgsVectorDataProvider::UnknownCount )
+  {
+    if ( returnvalue )
+      mFeaturesCounted += flist.size();
+    else
+      recalculateFeatureCount();
+  }
 
   if ( returnvalue )
     clearMinMaxCache();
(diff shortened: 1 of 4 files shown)


nyalldawson:
This definitely helps, thanks!

I've tracked down the remainder of the slowness to two things:

1. the call to recalculateFeatureCount() which is made after adding features in QgsOgrProvider. Commenting that out brings the speed equal to the QgsVectorFileWriter version. 

I think that we shouldn't be calling recalculateFeatureCount() here, but rather doing a shortcut version like:
 - if mFeaturesCounted ==  QgsVectorDataProvider::Uncounted || mFeaturesCounted == QgsVectorDataProvider::UnknownCount, do nothing -- we didn't have a count before adding the features, so we don't need it after either
- else just count the number of successful calls to addFeaturePrivate and add this to the existing count (which is what the postgres provider does)

2. The call to syncToDisc which occurs every 2000 features when using QgsVectorLayerExporter. Not sure immediately what the solution to this is. Possibly we could aggressively increase the buffer size for gpkg layers so that the sync occurs less often, or again just defer it to destruction of the exporter...

what do you think?

rouault:
>   if mFeaturesCounted ==  QgsVectorDataProvider::Uncounted || mFeaturesCounted == QgsVectorDataProvider::UnknownCount, do nothing -- we didn't have a count before adding the features, so we don't need it after either

sounds very reasonable

> else just count the number of successful calls to addFeaturePrivate and add this to the existing count (which is what the postgres provider does)

That wouldn't work in GPKG case since we're doing the inserts in a transaction (whose scope might be just addFeatures(), or larger if using the QGIS multi-layer transaction metchanism). So manually updating the feature count could be done only if all addFeaturePrivate() call have succeeded.

> Possibly we could aggressively increase the buffer size for gpkg layers so that the sync occurs less often

Yes for GPKG this could be put to 1 million. This doesn't affect RAM usage AFAIR: the content is written in a side car file by SQLite before being committed.

I can stack those changes here if you want

rouault:
> Yes for GPKG this could be put to 1 million. This doesn't affect RAM usage AFAIR: the content is written in a side car file by SQLite before being committed.

Hum this does affect QGIS RAM usage however. I'll try with 10 K (the current default is 200 actually, not 2000). A more involved strategy would be to try to evaluate the RAM usage of a QgsFeature

roya0045:
Is there even a need to recalculate the count? If truly needed couldn't we just remove or add to the count instead of iterating through the full dataset to update the count?

rouault:
> Is there even a need to recalculate the count? If truly needed couldn't we just remove or add to the count instead of iterating through the full dataset to update the count?

That's what I've just done in an above commit

roya0045:
Nice! My brain just didn't register the + or - in mFeaturesCounted += flist.size(); when I checked the code...

src/core/providers/ogr/qgsogrprovider.cpp
```diff
@@ -1875,7 +1875,14 @@ bool QgsOgrProvider::addFeatures( QgsFeatureList &flist, Flags flags )
...
   }
 
-  recalculateFeatureCount();
+  if ( mFeaturesCounted != QgsVectorDataProvider::Uncounted &&
+       mFeaturesCounted != QgsVectorDataProvider::UnknownCount )
+  {
+    if ( returnvalue )
+      mFeaturesCounted += flist.size();
+    else
+      recalculateFeatureCount();
```
nyalldawson, line 1884: What about just resetting the feature count to "uncounted" here, so that we defer the counting until it's actually needed?

rouault: @nyalldawson QgsOgrProvider::featureCount() just returns mFeaturesCounted . I don't want to rewrite all the provider :-) Calling recalculeFeatureCount() here isn't really an issue as it is an error code path.

nyalldawson: Fair enough - I thought we already had logic here around deferred feature counting!

nyalldawson (approved):

nyalldawson:
Tests great -- thanks very much!

timlinux:
Thanks so much for raising this @nyalldawson and for the fix @rouault !

PR #39440, opened 2020-10-17
QgsVectorLayerExporter: RAM based logic to decide when to flush features

rouault:
Flush features when we have ~ 10 MB of them in memory.

This is a continuation of https://github.com/qgis/QGIS/pull/39439 , which requires new core functions: QgsGeometry::wkbSize() and QgsFeature::approximateRAMUsage() . Not sure if this is 3.16 material

### DIFF
--- a/python/core/auto_generated/geometry/qgsabstractgeometry.sip.in
+++ b/python/core/auto_generated/geometry/qgsabstractgeometry.sip.in
@@ -196,6 +196,15 @@ Sets the geometry from a WKT string.
     typedef QFlags<QgsAbstractGeometry::WkbFlag> WkbFlags;
 
 
+    virtual int wkbSize( QgsAbstractGeometry::WkbFlags flags = QgsAbstractGeometry::WkbFlags() ) const = 0;
+%Docstring
+Returns the length of the QByteArray returned by :py:func:`~QgsAbstractGeometry.asWkb`
+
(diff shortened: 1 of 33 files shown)


nyalldawson:
Looks great! Could we make the name "approximateRamUsage" to keep capitalisation consistency?

rouault:
``approximateRamUsage`` looks weird to me, RAM being an acronym. Perhaps ``approximateMemoryUsage`` instead ?

nyalldawson:
> RAM being an acronym

We inherited qt's naming here (e.g. QDomElement), and try to keep consistency with that so that the API is more predictable.

nyalldawson:
(but obviously "memory" works too!)

rouault:
ok, renamed to ``approximateMemoryUsage``

nyalldawson:
Last thought: 10mb seems rather conservative to me. Would you be happy to bump this up? (A lot of workflows end up with memory layers of a couple of hundred mb, and I've not seen any issues reported regarding this

rouault:
> Last thought: 10mb seems rather conservative to me. Would you be happy to bump this up?

Just give me your number :-) But a transaction with 10 MB is already quite large. Not sure that increasing that will improve perfs that much.

nyalldawson:
100mb? 50 mb?

rouault:
ok, amended to use 100 MB

nyalldawson (approved):

PR #39443, opened 2020-10-17
Fix crash in GPKG connection api nativeTypes

elpaso:
See: https://github.com/qgis/QGIS/pull/37832#pullrequestreview-510987207

### DIFF
--- a/src/core/providers/ogr/qgsgeopackageproviderconnection.cpp
+++ b/src/core/providers/ogr/qgsgeopackageproviderconnection.cpp
@@ -435,11 +435,11 @@ QList<QgsVectorDataProvider::NativeType> QgsGeoPackageProviderConnection::native
   QgsVectorLayer::LayerOptions options { false, true };
   options.skipCrsValidation = true;
   const QgsVectorLayer vl { uri(), QStringLiteral( "temp_layer" ), QStringLiteral( "ogr" ), options };
-  if ( ! vl.isValid() || ! vl.dataProvider())
+  if ( ! vl.isValid() || ! vl.dataProvider() )
   {
-    const QString errorCause { vl.dataProvider() && vl.dataProvider()->hasErrors() ?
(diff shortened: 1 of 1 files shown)


src/core/providers/ogr/qgsgeopackageproviderconnection.cpp
```diff
@@ -432,10 +432,15 @@ QList<QVariantList> QgsGeoPackageProviderConnection::executeGdalSqlPrivate( cons
...
 {
   QList<QgsVectorDataProvider::NativeType> types;
-  QgsVectorLayer vl { uri(), QStringLiteral( "temp_layer" ), QStringLiteral( "ogr" ) };
+  QgsVectorLayer::LayerOptions options { false, true };
+  options.skipCrsValidation = true;
+  const QgsVectorLayer vl { uri(), QStringLiteral( "temp_layer" ), QStringLiteral( "ogr" ), options };
   if ( ! vl.isValid() || ! vl.dataProvider() )
   {
-    throw QgsProviderConnectionException( QObject::tr( "Error retrieving native types for %1: %2" ).arg( uri() ).arg( vl.dataProvider()->errors().join( '\n' ) ) );
+    const QString errorCause { vl.dataProvider() &&vl.dataProvider()->hasErrors() ?
```
uclaros, line 440: astyle seems to suggest removing the space after `&&` operator when inside a curly braces initialization. It's not that common a case, but it looks kind of ugly. Could this be mitigated in astyle?

nyalldawson: Probably easiest just to avoid the brace 😉

elpaso: Yeah, we have these nice tools that like to live their own lives and tell us what it's right and wrong. 

I think it's a kind of a human predominance thing that I insist to use `{}`.

PR #39448, opened 2020-10-18
[needs-docs][ui] UI/UX revamp of the 3D map configuration dialog

nirvn:
## Description

A GIF worth a thousand words:
Peek 2020-10-18 16-00

This PR revamps the 3D map configuration dialog to match the look and behavior of other parts of QGIS (labeling, diagrams). Over the years, the dialog has become an endless collection of widgets which the user has to scroll through, expand/collapse to reach. It was a nightmare which IMHO needed addressing prior to our upcoming 3.16 LTR.

The dialog now regroups widgets within themed panels. When reopening the dialog, the last selected panel is remembered (makes tweaking lights sooo  much easier).

### DIFF
--- a/src/ui/3d/map3dconfigwidget.ui
+++ b/src/ui/3d/map3dconfigwidget.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>555</width>
-    <height>675</height>
+    <width>593</width>
+    <height>623</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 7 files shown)


nirvn:
Note: the icons aren't great, but couldn't come with anything better as of this moment.

src/ui/3d/map3dconfigwidget.ui
```diff
@@ -473,6 +847,25 @@
...
  </tabstops>
- <resources/>
- <connections/>
+ <resources>
+  <include location="../../../images/images.qrc"/>
+ </resources>
+ <connections>
+  <connection>
+   <sender>m3DOptionsListWidget</sender>
+   <signal>currentRowChanged(int)</signal>
```
nyalldawson, line 856: Could you move this to C++ instead? Connections in UI files are too fragile.

nyalldawson (approved):
Looks beautiful!! Huge improvement

vcloarec:
Thanks you very much!!!

PR #39458, opened 2020-10-19
More 3d config dialog tweaks

nyalldawson:
Notably this includes a fix for the paper cut encountered by @timlinux where a terrain could be set to Mesh or DEM without requiring the user to actually select a layer, resulting in a broken terrain for the map:

image


nirvn (approved):
Nice!

PR #39460, opened 2020-10-19
Speed up mesh triangulation algorithm by avoiding a bunch of unnecessary QVector detachments

nyalldawson:


vcloarec:
good point! Maybe,one day, it will be good to change this way of referencing point and edge in this class...

PR #39462, opened 2020-10-19
improve ui for 3D directional light setting

vcloarec:
Make more user friendly the directional light settings for 3D view :
directional

### DIFF
--- a/src/app/3d/qgslightswidget.cpp
+++ b/src/app/3d/qgslightswidget.cpp
@@ -51,16 +51,13 @@ QgsLightsWidget::QgsLightsWidget( QWidget *parent )
   connect( btnRemoveDirectionalLight, &QToolButton::clicked, this, &QgsLightsWidget::onRemoveDirectionalLight );
 
   connect( cboDirectionalLights, qgis::overload<int>::of( &QComboBox::currentIndexChanged ), this, &QgsLightsWidget::onCurrentDirectionalLightChanged );
-  connect( spinDirectionX, qgis::overload<double>::of( &QDoubleSpinBox::valueChanged ), this, &QgsLightsWidget::onSpinBoxDirectionChanged );
-  connect( spinDirectionY, qgis::overload<double>::of( &QDoubleSpinBox::valueChanged ), this, &QgsLightsWidget::onSpinBoxDirectionChanged );
-  connect( spinDirectionZ, qgis::overload<double>::of( &QDoubleSpinBox::valueChanged ), this, &QgsLightsWidget::onSpinBoxDirectionChanged );
   connect( spinDirectionalIntensity, qgis::overload<double>::of( &QDoubleSpinBox::valueChanged ), this, &QgsLightsWidget::updateCurrentDirectionalLightParameters );
   connect( btnDirectionalColor, &QgsColorButton::colorChanged, this, &QgsLightsWidget::updateCurrentDirectionalLightParameters );
 
   connect( dialAzimuth, &QSlider::valueChanged, [this]( int value ) {spinBoxAzimuth->setValue( ( value + 180 ) % 360 );} );
-  connect( spinBoxAzimuth, qgis::overload<double>::of( &QDoubleSpinBox::valueChanged ), this, &QgsLightsWidget::onSpinBoxAzimuthChange );
-  connect( spinBoxAltitude, qgis::overload<double>::of( &QDoubleSpinBox::valueChanged ), this, &QgsLightsWidget::onSpinBoxAltitudeChange );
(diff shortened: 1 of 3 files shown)


vcloarec:
By the way, I wonder if the directional light could be better as default light (sun is directional on earth).

nirvn:
I'm honestly hesitating here. I see where you're going with that, I'm just not a big fan of the qwt widgets to begin with.

What about only keeping the spinboxes and get rid of the qwt and slider widgets?

Also, could this essentially allow us to get rid of the X / Y / Z spinners? If so, :tada:

vcloarec:
@nirvn , thanks for you opinion. 

Can you be more precise on your dislike of the qwt widget? It is the same widget  (used for the navigation widget in the 3D view, except appearance). Is it for aesthetic reasons or technical reasons? I hesitated with the QDial widget (as for raster layer hill shade), but I am not a big fan of this one and I find the qwt one looks better.

It will hurt a bit me to remove those widgets because they are very convenient to set quickly the direction and are quite explicit about what is done.

nyalldawson:
@vcloarec my thoughts:
- I strongly suspect that we'll need to rip out all of qwt from qgis at some stage in the near future. Qwt development is extremely slow paced, and I doubt we'll see any qt 6 compatibility for years (qwt master still is based on qt4, with ifdefs in place to allow it to work on qt5!). Ultimately Qwt was a huge pain point with the transition to qgis 3, and I don't want to see our reliance on it increasing.

- but probably more important: does the light direction actually correspond to a real compass direction? It's rather misleading if the UI gives this appearance, but then we just blindly treat north as "up" and don't do any handling for the map crs and actually ensure that "north" is north! 😁

vcloarec:
>     * but probably more important: does the light direction actually correspond to a real compass direction?

Indeed, nothing to ensure that it is the real North, but one thing is sure, it is the same "North" as the navigation compass. So for consistence and to help the user to not lose the north, may it is better to have cardinal directions on both compasses? Or, remove them on both?

For qwt, I understand that we will have to remove the widget one day (I saw your mail about this subject)... Here it is just a widget that sends the value when moving, nothing else, so not too difficult to remove. And changing value with only spin box is quite difficult and not very convenient. We could use QDial class, but here, for aesthetic reason, I am not sure... Maybe something exists on Qt Chart, but not sure.

In any case, we have to do something to set the direction by another way than X,Y,Z, I think it is to difficult for the user...

nyalldawson:
I'd still be inclined just to go with the qdial + spinbox combination. That gives some form of directional indication while still allowing exact control, and sidesteps the issue with compass directions. We'll need to revisit the navigation compass at some stage for the same reason, but I'd rather avoid the issue creeping elsewhere too.

Aside from the direction issue, the way I see it is:

- qwt compass: possibly has slightly more user control over the value

- qdial: no library issues, appearance is more modern and correctly respects platform/qgis theme and styling settings. Used elsewhere in qgis (eg label shadow direction, hillshade renderer, ...?)

vcloarec:
qwt removed, with default UI theme

directional

Other UI theme
image

image

nyalldawson:
Thanks! Can we kill the x/y/z spins now?

vcloarec:
I am thinking about replace the x,y,z spin box by labels to inform user of their value. What about?

nyalldawson:
Sounds great!

nirvn:
@vcloarec , much nicer already, thanks. You've also seem to have stumbled on a themes issue whereas the fill color of vertical sliders is inverted, I'll address that on my end.

As for spinboxes to a label, good move. You can kill the try rows and make it a one row information label.

nyalldawson:
@nirvn looks like blend of grey needs a fix for the dial markers too

vcloarec:
directional

nirvn:
@nyalldawson , couldn't find a way to properly style a QDial.

nyalldawson:
One last thought: is "altitude" the correct label here? Shouldn't that be declination?

vcloarec:
> One last thought: is "altitude" the correct label here? Shouldn't that be declination?

hmmmm, I learned something today : http://stellarium.sourceforge.net/wiki/index.php/Astronomical_Concepts#Altitude.2FAzimuth_Coordinates

So if I understand well it seems that _Altitude/Azimuth Coordinates_ are for position in the sky from elsewhere on earth, a direction view from earth. These coordinates change with the position of the observer
_Right Ascension/Declination Coordinates_ are like long/lat to define a position on the sky but they didn't change with the position on earth because it is position on the celestial sphere.

Here, I think it is preferable to use  _Altitude/Azimuth Coordinates_.

nyalldawson:
Ok, fair enough... getting my old astronomy classes mixed in with everything else...

PR #39464, opened 2020-10-19
fix sync between options list / stacked widget 3D config

vcloarec:
the option lists and the stacked widgets are not sync when opening


nirvn:
@vcloarec , fixed in PR #39458 .

vcloarec:
yea, I just see it...

PR #39467, opened 2020-10-19
fix height clamping on mesh terrain

### LINKED ISSUE #39455: Setting 3d map to use a mesh layer as terrain breaks symbology which relies on terrain height
If I create a 3d view and set its terrain to use a mesh layer, then it is not possible to use any 3d symbology settings which refer to the terrain height. 

E.g. setting a point layer to use "Terrain" for "Altitude clamping" always results in points with an elevation of 0. 

This works correctly for DEM and Online terrain sources.

(Ultimately, it's caused because QgsMeshTerrainGenerator doesn't implement the heightAt method.)

vcloarec:
fix #39455


src/3d/mesh/qgsmeshterraingenerator.cpp
```diff
@@ -171,6 +172,27 @@ void QgsMeshTerrainGenerator::readXml( const QDomElement &elem )
   mSymbol->readXml( elem.firstChildElement( "symbol" ), rwc );
 }
 
+float QgsMeshTerrainGenerator::heightAt( double x, double y, const Qgs3DMapSettings &map ) const
+{
+  Q_UNUSED( map );
+  if ( !meshLayer()->triangularMesh() )
```
nyalldawson, line 178: Doesn't this method need to be thread safe?

PeterPetrik: I think this just returns pointer, so should be ok.

PeterPetrik (approved):

src/3d/mesh/qgsmeshterraingenerator.cpp
nyalldawson: yes but couldn't the user be changing the mesh layer, deleting it, etc while a background thread is calling heightAt?

PeterPetrik: editing  is not allowed at  the moment, but you are right that maybe it is future-safe to be threadsafe here

nyalldawson: Well they could still delete the layer and cause a crash right now :laughing:

vcloarec:
Indeed, QGIS crashes if we remove the mesh terrain layer and clamp again on terrain.
The last commit makes the triangular mesh data implicitly shared ( triangular mesh has its data stored with QVector) while the mesh layer is set to the QgsMeshTerrainGenerator and heigthAt() used this data.
That is also an opportunity to apply the same principle for creation of 3D entity for terrain and dataset : the triangular mesh is pick up from the mesh layer as soon as posible in the main thread and propagated where it is needed with implicit sharing of data.

nyalldawson:
Thanks @vcloarec !

vcloarec:
test in Travis fails but not on my side...

nyalldawson:
@vcloarec I see this locally sometimes too. Some of these tests here have this note:

```
  // When running the test on Travis, it would initially return empty rendered image.
  // Capturing the initial image and throwing it away fixes that. Hopefully we will
  // find a better fix in the future.
```

Does calling captureSceneImage additional times help here? (i.e. more than two times). Otherwise I wonder if the test is actually revealing a real bug, which shows inconsistently. (possibly it's the cause of Tim's issue where the mesh settings would sometimes not stick and he'd have to reload the scene..)

tests/src/3d/testqgs3drendering.cpp
```diff
@@ -336,6 +336,8 @@ void TestQgs3DRendering::testMeshTerrain()
   meshTerrain->setLayer( mLayerMeshTerrain );
   map->setTerrainGenerator( meshTerrain );
 
+  QCOMPARE( meshTerrain->heightAt( 1500, 2200, *map ), 200.0 );
```
vcloarec, line 339: odd, this line does not fail...

tests/src/3d/testqgs3drendering.cpp
```diff
@@ -347,6 +349,8 @@ void TestQgs3DRendering::testMeshTerrain()
   // Capturing the initial image and throwing it away fixes that. Hopefully we will
   // find a better fix in the future.
   Qgs3DUtils::captureSceneImage( engine, scene );
+  Qgs3DUtils::captureSceneImage( engine, scene );
+  Qgs3DUtils::captureSceneImage( engine, scene );
 
   QImage img = Qgs3DUtils::captureSceneImage( engine, scene );
   QVERIFY( renderCheck( "mesh_terrain_1", img, 40 ) );
```
vcloarec, line 356: and this one fails

vcloarec:
I think the issue come from the change to make this more thread safe (but not sure with these failing renderings). But I really don't see where, and as on my side there is no problem, not easy to debug with travis...

nyalldawson:
Nice -- looks like you fixed it! Just gotta remove those debug statements now and we can merge

vcloarec:
yes, fixed!!! A long battle debugging with Travis remotely to finally find a silly issue, but good to know : if there is "uniform" in shaders that are not defined in Qt (QParameters), the shaders could not work...
But odd that work in my computer. I wonder if that depends on hardware and GPU. Maybe, some GPU could support this leak of definition of "uniform"?

nyalldawson (approved):

PR #39468, opened 2020-10-19
[ui] Fix mesh terrain single color / ramp settings out of sync

nirvn:
## Description

This fixes an out of sync scenario whereas the user selects mesh as terrain type, and gets to see rendering type set to single color yet the color ramp-related widgets aren't hidden.


nyalldawson:
Ah thanks -- I hit this too earlier!

PeterPetrik (approved):

PR #39477, opened 2020-10-19
Fix QGIS/rpm

lhirlimann:
Update the content of rpm/ so it works on modern fedoras.

* update default version to buidl using mock.
* remove 32bit builds as these will go away.
* cleanup spec file from cruft related to pre 28 versions of Fedora.


m-kuhn:
CC @daniviga

daniviga:
@m-kuhn LGTM

PR #39492, opened 2020-10-20
Avoid (offline)-suffix in offlineediting

signedav:
Fixes broken widgets and expressions in offline layers.

Offline layers created by offlineediting core (and with this by the offline editing plugin and QFieldSync) used to get the static suffix " (offline)".
This leaded to broken value relation widgets and broken expressions (using the layername).

The introduced suffix parameter in `convertToOfflineProject` allows to pass no suffix, and with this the layer names are unchanged in the offline project. 
This is what offline editing plugin does with this PR and QFieldSync will do it with the upcoming release as well.

Still the suffix parameter is by default " (offline)", what means that it will remain the same behavior for any other plugins using `convertToOfflineProject`. On `synchronize` it reads the suffix from a `customProperty` of the layer. If there is no such `customProperty` it's " (offline)" by default. This means, if the offline project has been created before this change and is synchronized back after this change, it will still work. So it's completely backwards compatible.

An indicator on the layer tree to visualize if it's an offline layer will follow up in another PR.

### DIFF
--- a/src/core/qgsofflineediting.cpp
+++ b/src/core/qgsofflineediting.cpp
@@ -61,6 +61,7 @@ extern "C"
 #define CUSTOM_PROPERTY_REMOTE_PROVIDER "remoteProvider"
 #define CUSTOM_SHOW_FEATURE_COUNT "showFeatureCount"
 #define CUSTOM_PROPERTY_ORIGINAL_LAYERID "remoteLayerId"
+#define CUSTOM_PROPERTY_LAYERNAME_SUFFIX "layerNameSuffix"
 #define PROJECT_ENTRY_SCOPE_OFFLINE "OfflineEditingPlugin"
 #define PROJECT_ENTRY_KEY_OFFLINE_DB_PATH "/OfflineDbPath"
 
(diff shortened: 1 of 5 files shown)


src/core/qgsofflineediting.cpp
```diff
@@ -249,7 +250,8 @@ void QgsOfflineEditing::synchronize()
     QString remoteSource = layer->customProperty( CUSTOM_PROPERTY_REMOTE_SOURCE, "" ).toString();
     QString remoteProvider = layer->customProperty( CUSTOM_PROPERTY_REMOTE_PROVIDER, "" ).toString();
     QString remoteName = layer->name();
-    remoteName.remove( QRegExp( " \\(offline\\)$" ) );
+    QString remoteNameSuffix = layer->customProperty( CUSTOM_PROPERTY_LAYERNAME_SUFFIX, " (offline)" ).toString();
+    remoteName.remove( remoteNameSuffix );
```
nyalldawson, line 254: This will remove the suffix from anywhere in the string, not just the end.

signedav: Thanks. I changed that.

signedav:
Hi @nyalldawson 
I expected to have this on the 3.16.1 release since it fixes issues in offline projects. Are there any problems with that?

I think it would be nice to have the indicators as well from the other PR. But there I would understand if not since it's not really a fix. Thats why I seperated the PRs.

nyalldawson:
I'm fine for this to backported!

signedav:
Cool. Thanks.

PR #39498, opened 2020-10-20
Add temporalFrameDurationChanged signal, and make view more aware of …

rduivenvoorde:
This commit adds a new signal to the QgsTemporalNavigationObject, which is
emitted when the frameDuration (of current QgsTemporalNavigationObject) is
changed.

It also fixes the issue that changing the frame in the Temporal
Navigation-WIDGET itself was not reflected in the widget itself. To see this, change the 'Step' spinbox in the widget and see that now the Frame-text (next to the little buttons in the top) is also changing.

Screenshot-20201020112923-717x119

I was a little puzzled about which signals to emit where. I even thought that this new signal was not needed at all because to me it looks like the 'updateTemporalRange( range )' signal is actually more or less used as 'modelChanged' signal?
But to be honest I always loose my way in the needed granularity of signals in model/view models...

Anyway: I really hope this will still make it in QGIS 3.16 as:
- it fixes gui glitches (like changing the Step not reflected in Frame: text
- it makes it truly possible now to register and actually use the temporal controller stuff in PyQGIS/Plugins

Sorry for being late :-(

Some example PyQGIS code I use to test it (and which I am planning to add to the cookbook after cleaning up...)
Note that this holds some old stuff:

```
l=iface.mapCanvas().currentLayer()
p=l.temporalProperties()
p.setMode(QgsVectorLayerTemporalProperties.ModeFeatureDateTimeInstantFromField)
# QgsVectorLayerTemporalProperties::setStartField( const QString &startFieldName )
#p.setStartField('t_start') # 6dots
p.setStartField('Time')   # view3
# void QgsVectorLayerTemporalProperties::setDurationUnits( QgsUnitTypes::TemporalUnit units )
# QgsUnitTypes.TemporalUnit.TemporalSeconds
p.setDurationUnits(QgsUnitTypes.TemporalUnit.TemporalSeconds)
p.setFixedDuration(1800)  # setting the LAYERS event duration
p.setIsActive(True) # OK
#p.changed.emit() # does not update the Layer Panel

# to update the legend (the temporal indicator) !!
node=QgsProject.instance().layerTreeRoot().findLayer(l) # find QgsLayerTreeLayer in QgsLayerTree
iface.layerTreeView().model().refreshLayerLegend(node)

# get a handle to current project and determine start and end range of current temporal enabled layers
project = QgsProject.instance()
range = QgsTemporalUtils.calculateTemporalRangeForProject(project)
print(f'Project Temporal Range: start: {range.begin().toString()} end: {range.end().toString()}')

# ??? get a handle to the Temporal Controller (QgsTemporalController) of this MapCanvas
c = iface.mapCanvas().temporalController()
#import sip
#sip.cast(iface.mapCanvas().temporalController(), QgsTemporalNavigationObject).setTemporalExtents(range)
#c.setTemporalExtents(range)

#c.temporalExtentsChanged.emit(range) # ?? should go vai c.setTemporalExtents(range)??
c.setTemporalExtents(range)

duration = c.frameDuration()
print(f'Player frameDuration: {duration.seconds()} seconds')  # PLAYER duration!!

#i=QgsInterval()
#i.setSeconds(3600)
#c.setFrameDuration(i)
# quicker via:
c.setFrameDuration(QgsInterval(QgsInterval.HOUR*1))

duration = c.frameDuration()
print(f'Player frameDuration: {duration.seconds()} seconds')  # PLAYER duration!!


#c = iface.mapCanvas().temporalController()
#c.skipToEnd()
#c.next()
#c.previous()
```

### DIFF
--- a/src/gui/qgstemporalcontrollerwidget.cpp
+++ b/src/gui/qgstemporalcontrollerwidget.cpp
@@ -53,6 +53,7 @@ QgsTemporalControllerWidget::QgsTemporalControllerWidget( QWidget *parent )
   setWidgetStateFromNavigationMode( mNavigationObject->navigationMode() );
   connect( mNavigationObject, &QgsTemporalNavigationObject::navigationModeChanged, this, &QgsTemporalControllerWidget::setWidgetStateFromNavigationMode );
   connect( mNavigationObject, &QgsTemporalNavigationObject::temporalExtentsChanged, this, &QgsTemporalControllerWidget::setDates );
+  connect( mNavigationObject, &QgsTemporalNavigationObject::temporalFrameDurationChanged, this, &QgsTemporalControllerWidget::setTimeStep );
   connect( mNavigationOff, &QPushButton::clicked, this, &QgsTemporalControllerWidget::mNavigationOff_clicked );
   connect( mNavigationFixedRange, &QPushButton::clicked, this, &QgsTemporalControllerWidget::mNavigationFixedRange_clicked );
   connect( mNavigationAnimated, &QPushButton::clicked, this, &QgsTemporalControllerWidget::mNavigationAnimated_clicked );
--- a/src/gui/qgstemporalcontrollerwidget.cpp
+++ b/src/gui/qgstemporalcontrollerwidget.cpp
@@ -53,7 +53,7 @@ QgsTemporalControllerWidget::QgsTemporalControllerWidget( QWidget *parent )
   setWidgetStateFromNavigationMode( mNavigationObject->navigationMode() );
   connect( mNavigationObject, &QgsTemporalNavigationObject::navigationModeChanged, this, &QgsTemporalControllerWidget::setWidgetStateFromNavigationMode );
   connect( mNavigationObject, &QgsTemporalNavigationObject::temporalExtentsChanged, this, &QgsTemporalControllerWidget::setDates );
-  connect( mNavigationObject, &QgsTemporalNavigationObject::temporalFrameDurationChanged, this, &QgsTemporalControllerWidget::setTimeStep );
+  connect( mNavigationObject, &QgsTemporalNavigationObject::temporalFrameDurationChanged, this, &QgsTemporalControllerWidget::setTimeStepWithBlocking );
   connect( mNavigationOff, &QPushButton::clicked, this, &QgsTemporalControllerWidget::mNavigationOff_clicked );
   connect( mNavigationFixedRange, &QPushButton::clicked, this, &QgsTemporalControllerWidget::mNavigationFixedRange_clicked );
   connect( mNavigationAnimated, &QPushButton::clicked, this, &QgsTemporalControllerWidget::mNavigationAnimated_clicked );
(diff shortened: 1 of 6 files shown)


src/gui/qgstemporalcontrollerwidget.cpp
```diff
@@ -53,6 +53,7 @@ QgsTemporalControllerWidget::QgsTemporalControllerWidget( QWidget *parent )
   setWidgetStateFromNavigationMode( mNavigationObject->navigationMode() );
   connect( mNavigationObject, &QgsTemporalNavigationObject::navigationModeChanged, this, &QgsTemporalControllerWidget::setWidgetStateFromNavigationMode );
   connect( mNavigationObject, &QgsTemporalNavigationObject::temporalExtentsChanged, this, &QgsTemporalControllerWidget::setDates );
+  connect( mNavigationObject, &QgsTemporalNavigationObject::temporalFrameDurationChanged, this, &QgsTemporalControllerWidget::setTimeStep );
```
nyalldawson, line 56: I'm a little worried here about a signal bouncing back and forward from the temporal navigation object to the widget, and then the logic in QgsTemporalControllerWidget::setTimeStep causing a change to mTimeStepsComboBox or mStepSpinBox and then that being sent back to the QgsTemporalNavigationObject. Can you put a blocker in here by setting a local "bool mBlockFrameDurationChangedSignal" in QgsTemporalControllerWidget which is set to true at the start of setTimeStep and set back to false at the end of setTimeStep, and which prevents the extra call back to setTimeStep here?

rduivenvoorde: @nyalldawson thanks for looking. Yes I will. This is always one of those gui/events-programming things that I'm not sure how to do this in a proper/save/clean way... I think the safest way would be a gui in which you set a TimeStep/QgsInterval in one go, that is there is no 'intermediate product' going to the controller when the user is busy with one of the 2 widgets involved in this... For now I will add the blocker.

rduivenvoorde: @nyalldawson trying about this, and thinking about it, I think that the 'check' in 

https://github.com/qgis/QGIS/pull/39498/files#diff-b4a4c7eb1b9251a97a6e0bdaad9d253fcedd249631167e366cd3f4c27f5789f6R181-R184

actually is enough to prevent bouncing of events?

Looking at this problem as a true Model View way: if  we would have a couple of views/widgets showing the state of the model (QgsTemporalNavigationObject) we want those to change immediately if the user changes something in another view isn't it?

I'm also trying/willing to implement your idea, but just don't fully get it...
What could/should I block from happening? The fact that a user can (via the widgets) be in the middle of setting a (non-valid intermediat) value, I cannot block is it?

nyalldawson: I'm not confident that that check will be sufficient, because the logic in the widget is quite complex (and I can't tell immediately how it's going to mess with the value coming from the navigation object)!

PR #39500, opened 2020-10-20
Fix backwards compatibility of QgsFieldCalculatorAlgorithm

### LINKED ISSUE #39387: Field calculator processing algorithm lost field name in existing model
**Describe the bug**
I have a model created some month ago that contains a few *Field Calculator* algorithms. The resulting field name is blank if I try to run the model with master 69faea81d9, causing the model to fail.

In older version:
image

In 69faea81d9:
image

This is caused by the c++ port of the algorithm and the UX changes in #38695.
Ping @suricactus

suricactus:
After the translation to C++, I was splitted the configuration for existing and new field. However, the little I knew that there is no easy way to keep backwards compatibility with the modeler. Fix https://github.com/qgis/QGIS/issues/39387

This changed the input UI a bit:
image

Once the model algorithm is saved with the old parameter names,  I failed to find a way to migrate the parameter names. I can imagine a method like `QVariantMap QgsProcessingAlgorithm::migrateParameters( QVariantMap loadedParams )` that would take the `loadedParams` parameters as the ones from the model file, and return the corrent params names. Something like:
```
def migrateParameters( loadedParams  ):
  param1, param2 = loadedParams['legacyParam'].split(' ')
  return { param1: param1, param2: param2 }
  
migrateParameters( { 'legacyParam': 'hello world' } )  # => { param1: 'hello', param2: 'world' }
```

### DIFF
--- a/src/analysis/processing/qgsalgorithmfieldcalculator.cpp
+++ b/src/analysis/processing/qgsalgorithmfieldcalculator.cpp
@@ -92,7 +92,7 @@ QString QgsFieldCalculatorAlgorithm::shortHelpString() const
   return QObject::tr( "This algorithm computes a new vector layer with the same features of the input layer, "
                       "but either overwriting an existing attribute or adding an additional attribute. The values of this field "
                       "are computed from each feature using an expression, based on the properties and attributes of the feature. "
-                      "Note that if \"Field name\" is existing field in the layer,all the rest of the field settings are ignored." );
+                      "Note that if \"Field name\" is an existing field in the layer then all the rest of the field settings are ignored." );
 }
 
 QgsFieldCalculatorAlgorithm *QgsFieldCalculatorAlgorithm::createInstance() const
--- a/src/analysis/processing/qgsalgorithmfieldcalculator.cpp
+++ b/src/analysis/processing/qgsalgorithmfieldcalculator.cpp
@@ -186,11 +186,10 @@ QgsFeatureList QgsFieldCalculatorAlgorithm::processFeature( const QgsFeature &fe
     }
 
     attributes[mFieldIdx] = value;
-    attributes.append( value );
   }
   else
   {
-    attributes.append( QVariant() );
+    attributes[mFieldIdx] = QVariant();
   }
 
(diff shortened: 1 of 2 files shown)


src/analysis/processing/qgsalgorithmfieldcalculator.cpp
```diff
@@ -94,8 +92,7 @@ QString QgsFieldCalculatorAlgorithm::shortHelpString() const
   return QObject::tr( "This algorithm computes a new vector layer with the same features of the input layer, "
                       "but either overwriting an existing attribute or adding an additional attribute. The values of this field "
                       "are computed from each feature using an expression, based on the properties and attributes of the feature. "
-                      "Note that selecting a value in \"Result in existing field\" will ignore all the rest of the "
-                      "field settings." );
+                      "Note that if \"Field name\" is existing field in the layer,all the rest of the field settings are ignored." );
```
nyalldawson, line 95:
```suggestion
                      "Note that if \"Field name\" is an existing field in the layer then all the rest of the field settings are ignored." );
```

nyalldawson (changes requested):
There's some bugs here:

From line 188:

```
    attributes[mFieldIdx] = value;
    attributes.append( value );
  }
  else
  {
    attributes.append( QVariant() );
  }
```

This is incorrect -- if an existing field is being replaced then you shouldn't be appending a new value to the end of attributes. And similarly if a new field is being added then you shouldn't be calling `attributes[mFieldIdx] = value;`

Can you fix this too please?

src/analysis/processing/qgsalgorithmfieldcalculator.cpp
suricactus: done

suricactus:
The append is not needed at all. Thanks for spotting this. If you look here, I always have a valid `mFieldIdx`:
https://github.com/qgis/QGIS/pull/39500/files#diff-80c9e2e480ddd1856f6dec1760f8bcaf63542f497565bad06da2aa526316a5d9R134-R141

nyalldawson (approved):

PR #39505, opened 2020-10-20
[Oracle]  Fix default value evaluation when adding features

### LINKED ISSUE #39504: [Oracle] Fail to commit default values evaluated on provider side
**Describe the bug**

String default expression would fail if the option *Evaluate default values on provider side* is checked

PR is coming

**How to Reproduce**

1. Create table with default values
```sql
CREATE TABLE "QGIS"."TEST_EVAL_EXPR" (pk INTEGER, "name" VARCHAR2(100) DEFAULT 'qgis')
```

2. Load it in QGIS
3. Activate the option *Evaluate default values on provider side*
4. Add a feature and save
5. The commit fails with the following message

```
 Provider errors:
      Oracle error while adding features: Oracle error: Evaluation of default value failed
    SQL: ORA-00904: "QGIS": invalid identifier
    Unable to execute statement
    Error: SELECT qgis FROM dual
```

**QGIS and OS versions**

QGIS master
Debian testing

**Additional context**

troopa81:
Fixes #39504 : Provider has to evaluate *defaultValueClause* when needed, not the already evaluated string.

### DIFF
--- a/src/providers/oracle/qgsoracleprovider.cpp
+++ b/src/providers/oracle/qgsoracleprovider.cpp
@@ -1353,7 +1353,7 @@ bool QgsOracleProvider::addFeatures( QgsFeatureList &flist, QgsFeatureSink::Flag
 
       insert += delim + quotedIdentifier( fld.name() );
 
-      QString defVal = defaultValueClause( idx );
+      const QString defVal = defaultValueClause( idx );
 
       values += delim + '?';
       defaultValues.append( defVal );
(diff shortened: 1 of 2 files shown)


troopa81:
I'm not in favor of backporting this in 3.10, please let me know if you think otherwise

src/providers/oracle/qgsoracleprovider.cpp
```diff
@@ -1353,7 +1353,7 @@ bool QgsOracleProvider::addFeatures( QgsFeatureList &flist, QgsFeatureSink::Flag
 
       insert += delim + quotedIdentifier( fld.name() );
 
-      QString defVal = defaultValue( idx ).toString();
+      QString defVal = defaultValueClause( idx );
```
elpaso, line 1356: may be const?
```suggestion
      const QString defVal = defaultValueClause( idx );
```

elpaso (approved):

PR #39506, opened 2020-10-20
Homogenize the look of Identify Results for different LayerTypes

rduivenvoorde:
Minimal gui fix:

The code to build the Treeview for the Identify Results is split
in four, for the four different LayerTypes (vector, raster, mesh,
vector-tile).

But although vector and vector tile showed the 'toplevel' in bold
with the layername, the others showed... layer-count and name
in normal font.

This commit changes it so if you use the identify tool in a mixed
layer environment, they look more or less the same...

@PeterPetrik or is there a special reason to show it like this?
Same question for raster-people...

In screenshot below there are 4 different layer types.
Left is current situation
Right is this PR
( minor change, I know but it bugged me :-) )

Screenshot-20201020162754-1674x619

### DIFF
--- a/src/app/qgsidentifyresultsdialog.cpp
+++ b/src/app/qgsidentifyresultsdialog.cpp
@@ -881,10 +881,12 @@ void QgsIdentifyResultsDialog::addFeature( QgsRasterLayer *layer,
 
   if ( !layItem )
   {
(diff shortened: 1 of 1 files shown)


PeterPetrik (approved):
Thanks, frankly I do not remember, but now it looks definitely better.

PR #39513, opened 2020-10-21
Improve marker rotation and point offset tools

nyalldawson:
Fixes some issues with these tools:
- point offset preview symbol was not being shown
- tools used click and hold, instead of standard click-click behavior
- preview overlay for marker rotation was super ugly and covered too much of the underlying map


nirvn:
Most welcome.

PR #39517, opened 2020-10-21
[Backport 3.10] Don't clear document path from external resource widget when its null 

troopa81:
## Description

Backport #39367 in 3.10


nyalldawson:
Manually marking as stale because stalebot has left for christmas break

troopa81:
> Manually marking as stale because stalebot has left for christmas break

Should we not merge this one in 3.10 ? Looks like nobody has complained about the fix in master since 16th of october.

nyalldawson (approved):

PR #39520, opened 2020-10-21
[ui] Add an Apply button to the 3D map configuration dialog

nirvn:
## Description

Because going in and out of the dialog to tweak settings is boooring. 

GIF:
Peek 2020-10-21 16-17

### DIFF
--- a/src/app/3d/qgs3dmapcanvasdockwidget.cpp
+++ b/src/app/3d/qgs3dmapcanvasdockwidget.cpp
@@ -260,9 +260,42 @@ void Qgs3DMapCanvasDockWidget::configure()
 
   Qgs3DMapSettings *map = mCanvas->map();
   Qgs3DMapConfigWidget *w = new Qgs3DMapConfigWidget( map, mMainCanvas, &dlg );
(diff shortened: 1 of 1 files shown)


nirvn:
Unrelated qt 5.14 check failure, travis green.

src/app/3d/qgs3dmapcanvasdockwidget.cpp
```diff
@@ -260,45 +260,53 @@ void Qgs3DMapCanvasDockWidget::configure()
...
   {
     buttons->button( QDialogButtonBox::Ok )->setEnabled( valid );
   } );
 
-  QVBoxLayout *layout = new QVBoxLayout( &dlg );
-  layout->addWidget( w, 1 );
-  layout->addWidget( buttons );
-  if ( !dlg.exec() )
-    return;
+  connect( w, &Qgs3DMapConfigWidget::configChanged, this, [ = ]()
```
nyalldawson, line 278: Unless I'm missing something obvious, configChanged is never emitted if "ok" is clicked. So changes won't be saved if "apply" is not clicked before OK

nyalldawson: (to solve this it's probably cleanest to move this lambda code out to a private slot, and both connect it to the apply button  click and also call it after the dlg is executed.

nyalldawson (approved):

PR #39524, opened 2020-10-21
also use SVG selector in SVG fill symbology

3nids:
follow up of #39421 

some duplicated code removal :)


nirvn (approved):
A net minus 131 line of code, yes please.

PR #39527, opened 2020-10-21
Indicator for offline layers

signedav:
Indicator is displayed according the custom property `isOfflineEditable` of `QgsMapLayer`.
This is introduced because with this PR  #39492 there the possibility to avoid ` (offline)` suffix in the layer names (what used to lead to troubles).
indicator

**Thoughts about the symbol**
I wanted to avoid to have a "negative" symbol like unplugged cable or no-wifi.
So I propose a box-icon. Standing for a portable layer.


nirvn:
Is there a reason/need for this to be implemented here instead of the qfieldsync plugin?

signedav:
Offline Editing Core (used by QFieldSync, Offline Editing Plugin and otherers) used to write the suffix " (offline)". Since this is fixed in core by avoiding writing the suffix, this is integrated here as well.

src/core/qgsmaplayer.h
```diff
@@ -1424,6 +1424,12 @@ class CORE_EXPORT QgsMapLayer : public QObject
      */
     void isValidChanged();
 
+    /**
+     * Emitted when a custom property of a node within the tree has been changed or removed
```
nyalldawson, line 1428:
```suggestion
     * Emitted when a custom property of the layer has been changed or removed.
```

src/core/qgsmaplayer.h
```diff
@@ -1424,6 +1424,12 @@ class CORE_EXPORT QgsMapLayer : public QObject
...
+     *
+     * \since QGIS 3.16
+     */
+    void customPropertyChanged( QgsMapLayer *layer, const QString &key );
```
nyalldawson, line 1432:
```suggestion
    void customPropertyChanged( const QString &key );
```

Let's drop the layer argument -- it's not required, since it will always be available via sender() if needed.

src/core/qgsmaplayer.cpp
```diff
@@ -1707,7 +1707,11 @@ QStringList QgsMapLayer::customPropertyKeys() const
 
 void QgsMapLayer::setCustomProperty( const QString &key, const QVariant &value )
 {
-  mCustomProperties.setValue( key, value );
+  if ( !mCustomProperties.contains( key ) || mCustomProperties.value( key ) != value )
+  {
+    mCustomProperties.setValue( key, value );
+    emit customPropertyChanged( this, key );
```
nyalldawson, line 1713: Can you add tests for these signals please?

nyalldawson (commented):
Code looks good. We'll need to revise that icon though, it doesn't match the style used by other indicator icons (faux 3d vs flat).

Can we defer merging this till after release?

signedav:
Since I find it difficult to illustrate a box in a flat icon I had another idea.

src/core/qgsmaplayer.h
```diff
@@ -1424,6 +1424,12 @@ class CORE_EXPORT QgsMapLayer : public QObject
      */
     void isValidChanged();
 
+    /**
+     * Emitted when a custom property of the layer has been changed or removed.
+     *
+     * \since QGIS 3.16
```
nyalldawson, line 1430:
```suggestion
     * \since QGIS 3.18
```

nyalldawson (approved):

nyalldawson:
I'm happy with this PR! 

I'll delegate it to @nirvn now to discuss the icon itself :wink:

nirvn:
I think we could try another attempt at an icon here. The checkbox one is better than the box, but it's so generic.

I was thinking maybe something along the lines of a wall outlet like this:
image

signedav:
Thanks @nirvn I like the idea.

Maybe it's because I'm Swiss that I have to think about an afraid goblin - it's not a positive connotation.
The European slot looks more like a robot - I like that better. Is maybe something you can connotate with "being mobile" as well.
But maybe it's bad to have to much European things in it - so there is a general slot as well. But this is too abstract again.

|US slot | general slot  | European slot |
|---------|------|--------|
|usplug|anonym|schuko|

nirvn:
I'd go for the US slot as the three holes patterns makes it more distinctive. But I'm from Canada, maybe I'm biased :)

nyalldawson:
Could you avoid the cultural issue by going for an unplugged cable instead?

signedav:
You mean like this @nyalldawson ?
image

Doesn't this look like something is not good with this layer? Like it has no connection but it should have?

nirvn:
I'm +1 on unplugged cable.

nyalldawson:
Yeah

signedav:
I don't really like it. It looks for me like something happened with the layer that shouldn't have happened.
But okay. I'll give it some more thoughts. And if I don't have another good idea I can convince you with, then let's go for unplugged.

signedav:
Well, I didn't came up with a better idea, so I think we can rely on your experience that it's a good choice :-)

Looks like this:
image

@nirvn I made the icon in Inkscape and have a lot of crap in the source. Do you know how to clean it up?

signedav:
Happy with the icon @nirvn? 
Nyall approved the code already.

nirvn:
Icon looks much better. Can you try to run it through an online svg compressor?

signedav:
Failing Mac OS build is unrelated.

PR #39529, opened 2020-10-21
Option to add field constraint hints via programatic call in "Refactor field"

suricactus:
Allows to programatically open the algorithm dialog with preset constraint hints.

image

Test with:
```
dlg = processing.createAlgorithmDialog("native:refactorfields", { 
    'FIELDS_MAPPING' : [
        {'expression': '\"id\"','length': -1,'name': 'id','precision': 0,'type': 2},
        {'expression': '4','length': -1,'name': 'name1','precision': 0,'type': 10, 'constraints': QgsFieldConstraints.ConstraintNotNull | QgsFieldConstraints.ConstraintUnique}
        ], 
    'INPUT' : 'memory://Point?crs=EPSG:4326&field=id:integer(-1,0)&field=name:string(-1,0)', 
    'OUTPUT' : '/home/suricactus/projects/tmp/asd1.gpkg' })
res = dlg.exec_()
```

### DIFF
--- a/src/gui/processing/qgsprocessingfieldmapwidgetwrapper.cpp
+++ b/src/gui/processing/qgsprocessingfieldmapwidgetwrapper.cpp
@@ -120,6 +120,7 @@ void QgsProcessingFieldMapPanelWidget::setValue( const QVariant &value )
   QgsFields destinationFields;
   QMap<QString, QString> expressions;
 
+  const QgsFields layerFields = mLayer ? mLayer->fields() : QgsFields();
   const QVariantList fields = value.toList();
   for ( const QVariant &field : fields )
   {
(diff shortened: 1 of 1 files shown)


src/gui/processing/qgsprocessingfieldmapwidgetwrapper.cpp
```diff
@@ -130,6 +131,27 @@ void QgsProcessingFieldMapPanelWidget::setValue( const QVariant &value )
...
+    {
+      const QgsFieldConstraints::Constraints constraints = static_cast<QgsFieldConstraints::Constraints>( map.value( QStringLiteral( "constraints" ), 0 ).toInt() );
+      QgsFieldConstraints fieldConstraints;
+
+      if ( constraints & QgsFieldConstraints::ConstraintNotNull )
+        fieldConstraints.setConstraint( QgsFieldConstraints::ConstraintNotNull );
+      if ( constraints & QgsFieldConstraints::ConstraintUnique )
+        fieldConstraints.setConstraint( QgsFieldConstraints::ConstraintUnique );
+      if ( constraints & QgsFieldConstraints::ConstraintExpression )
+        fieldConstraints.setConstraint( QgsFieldConstraints::ConstraintUnique );
```
m-kuhn, line 150:
```suggestion
        fieldConstraints.setConstraint( QgsFieldConstraints::ConstraintExpression );
```

suricactus: Whooops, thanks

src/gui/processing/qgsprocessingfieldmapwidgetwrapper.cpp
```diff
@@ -130,6 +131,27 @@ void QgsProcessingFieldMapPanelWidget::setValue( const QVariant &value )
                 map.value( QStringLiteral( "length" ), 0 ).toInt(),
                 map.value( QStringLiteral( "precision" ), 0 ).toInt() );
 
+    int layerFieldIdx = layerFields.indexFromName( f.name() );
+
+    if ( mLayer && layerFieldIdx >= 0 && ! map.contains( QStringLiteral( "constraints" ) ) )
+    {
+      f.setConstraints( layerFields.at( layerFieldIdx ).constraints() );
```
gacarrillor, line 138: _(This was going to be a new QGIS issue, but I know where the behavior comes from, so I prefer to comment it here.)_

I think we would not needed to set constraints from `mLayer`, because it could be the source/input layer.

Highlighting constraints is relevant only if they are obtained from the target/template layer.

-------------------------------------
**Example**

This two point tables share the field `exactitud_vertical`, but have different constraints set on it.

```
CREATE TABLE survey.lc_puntocontrol (
	t_id bigserial NOT NULL,
	tipo_punto_control int8 NULL,
	exactitud_horizontal numeric(5,3) NOT NULL,
	exactitud_vertical numeric(5,3) NOT NULL,
	CONSTRAINT lc_puntocontrol_pkey PRIMARY KEY (t_id)
);

CREATE TABLE survey.lc_puntolindero (
	t_id bigserial NOT NULL,
	puntotipo int8 NOT NULL,
	exactitud_horizontal numeric(5,3) NOT NULL,
	exactitud_vertical numeric(5,3) NULL,
	CONSTRAINT lc_puntolindero_pkey PRIMARY KEY (t_id)
);
```

Steps to reproduce:

1. Open the Refactor Fields algorithm.
2. Set the template layer first (as `lc_puntolindero`).
3. Now set the INPUT layer (as `lc_punto_control`).
4. You get:
image

As you can see, `exactitud_vertical` is highlighted, but it shouldn't, since the template layer has no constraint on such field.

-------------------------------------

Could we perhaps remove the `if ()` part above and only leave the `else` part (with its corresponding condition, of course)?

@suricactus @m-kuhn

suricactus: Yes, of course. Thank you for the detailed explanation what is expected! I will address this issue to match the expected behaviour.

suricactus: hey, @gacarrillor , sorry for the delay. This should fix the issue: https://github.com/qgis/QGIS/pull/40703

gacarrillor: Thank you @suricactus :+1:

PR #39531, opened 2020-10-21
fix 3D directional light ui

vcloarec:
altitude was always applied with value of zero


nyalldawson:
Do you think we should block altitudes <0 here? I've tried and can't obtain any usable results with upward angled lights, so I think they'll just lead to user confusion...

vcloarec:
> Do you think we should block altitudes <0 here?

I had think about, but preferred let's negative value possible to allow user to have upward lights if he wants. I do not have strong opinion on this...

nyalldawson:
What I mean is that upward lights don't seem to work in any way. They always result in a dark scene.

vcloarec:
yes, true, except with vertical or quasi vertical triangles. Let's remove negative altitude values!

nyalldawson (approved):

PR #39540, opened 2020-10-21
Test doxy layout

m-kuhn:
## Description

More documentation safety checks

### DIFF
--- a/src/analysis/processing/qgsalgorithmtinmeshcreation.cpp
+++ b/src/analysis/processing/qgsalgorithmtinmeshcreation.cpp
@@ -23,6 +23,8 @@
 #include "qgsmeshlayer.h"
 #include "qgis.h"
 
+///@cond PRIVATE
+
 QString QgsTinMeshCreationAlgorithm::group() const
 {
   return QObject::tr( "Mesh" );
@@ -206,3 +208,5 @@ bool QgsTinMeshCreationAlgorithm::canExecute( QString *errorMessage ) const
 
   return true;
 }
+
+///@endcond PRIVATE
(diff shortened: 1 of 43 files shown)


src/analysis/processing/qgsalgorithmtinmeshcreation.cpp
```diff
@@ -132,17 +134,13 @@ QVariantMap QgsTinMeshCreationAlgorithm::processAlgorithm( const QVariantMap &pa
     destinationCrs = context.project()->crs();
   triangulation.setCrs( destinationCrs );
 
-  if ( !mVerticesLayer.isEmpty() && feedback )
```
nyalldawson, line 135: What happened here? Bad rebase?

m-kuhn: ouch

m-kuhn:
I can't fix

```
/home/runner/work/QGIS/QGIS/src/core/labeling/qgspallabeling.h:272: error: documented symbol `enum Q_DECL_DEPRECATED QgsPalLayerSettings::QgsPalLayerSettings::LinePlacementFlags' was not declared or defined. (warning treated as error, aborting now)
```

What I tried is to define `Q_DECL_DEPRECATED=`. This works but triggers a new set of issues for the check that `\deprecated` is also `Q_DECL_DEPRECATED`. I tried aliasing `Q_DECL_DEPRECATED_ENUM` to `Q_DECL_DEPRECATED` and defining this to `Q_DECL_DEPRECATED_ENUM=`. This works but would require a myriad of sipify patches.
@nyalldawson ok to just `@cond PRIVATE` the two affected enums? Or drop the `Q_DECL_DEPRECATED` and go for dox only?

nyalldawson:
> Or drop the Q_DECL_DEPRECATED and go for dox only?

Yep, try this one first?

PR #39543, opened 2020-10-22
Add temporalFrameDurationChanged signal

nyalldawson:
Updated version of https://github.com/qgis/QGIS/pull/39498

@rduivenvoorde does this work for you?


rduivenvoorde:
> Updated version of #39498
> @rduivenvoorde does this work for you?

Building...

rduivenvoorde:
@nyalldawson yep, this all works for me:
- gui is/widgets are updated on stepDuration changes
- I can change the stepDuration in pyqgis and it is reflected and working!

Although I would not be able to fix it like this myself, in this way...

Thanks very much for this, really appreciated!!

PR #39544, opened 2020-10-22
Display available variables in the expression parameter dialog of a modeler

### LINKED ISSUE #39137: [Modeler] Missing variables in modeler expression when used in default expression
Another modeler Bug or improvement.

The expression builder dialogs display the available variables when initiated from a processing window (IE an expression parameter in a tool in the model or when setting an expression parameter in the model dialog itself).

But the variables of the model are not available when using the expression builder to set a default value for an expression, even though those variables are properly evaluated at runtime.

This can be misleading as it would appear that the default expression does not have access to those elements or force the user to have a separate way to use the variables that are available in the model.

The fix would be to simply provide the model expression context scope to this expression builder for the expression parameter.

roya0045:
## Description

Fixes #39137 

This is my proposal to display the variables of a model in the default expression dialog.

I tried many permutations and this is the best way that I found, use the same context creation as used in the dialogs during widget initiation, cache it in the dialog. 

This was needed since the processing context couldn't be stored safely and was deleted, so adding the parameters to the function and calling it during creation was the best way to ensure that those parameters are available during creation.

Caching the context in the existing member context of the QgsExpressionLineEdit ensure that there was no need to call the context creation later.

### DIFF
--- a/python/gui/auto_generated/qgsexpressionlineedit.sip.in
+++ b/python/gui/auto_generated/qgsexpressionlineedit.sip.in
@@ -127,6 +127,15 @@ an expression context for the widget.
                   create an expression context when required.
 %End
 
+    void setExpressionContext( QgsExpressionContext &context );
+%Docstring
+Set a default expression context. Will be used if no QgsEcpressionContextGenerator or layer is set.
+
+:param context: The QgsExpressionContext to use.
+
+.. versionadded:: 3.18
+%End
+
   signals:
 
     void expressionChanged( const QString &expression );
--- a/src/gui/processing/qgsprocessingwidgetwrapperimpl.cpp
+++ b/src/gui/processing/qgsprocessingwidgetwrapperimpl.cpp
@@ -59,6 +59,8 @@
 #include "qgsrasterbandcombobox.h"
 #include "qgsprocessingoutputdestinationwidget.h"
 #include "qgscheckablecombobox.h"
+#include "qgsexpressioncontext.h"
+#include "qgsexpressioncontextutils.h"
 #include <QToolButton>
 #include <QLabel>
 #include <QHBoxLayout>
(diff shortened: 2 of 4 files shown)


roya0045:
Unsure if I should keep the two alg scope or ditch them as they don't seem to change anything usefull.

roya0045:
@nyalldawson works as intended. ready for review.

roya0045:
@nyalldawson simple ping, let me know if you need anything else.

m-kuhn:
Please keep an eye on it, according to https://github.com/actions/stale it should


> Input | Description | Usage
> -- | -- | --
> remove-stale-when-updated | Remove stale label from issue/pr on updates or comments. Defaults to true | Optional
> -- | -- | --

roya0045:
@m-kuhn seem to have worked, we only need to wait for the next run.

roya0045:
keeping alive

nyalldawson:
@roya0045 There's a substantial rework needed here -- we should remove the new setExpressionContext method added to the widget and instead rely entirely on the existing expression context generator API. 
QgsProcessingExpressionParameterDefinitionWidget::createExpressionContext should be based off QgsProcessingModelerParameterWidget::createExpressionContext(), but with removal of all the child algorithm specific scopes as these will have no meaning for a parameter default value (which sits at the model level, not a child algorithm level). So the created expression would be just the expression context taken from the processing context + a single QgsExpressionContextUtils::processingModelAlgorithmScope. For similar reasons all the highlighted variable/function handling isn't required, as that is all specific to the child algorithm level.

roya0045:
> @roya0045 There's a substantial rework needed here -- we should remove the new setExpressionContext method added to the widget and instead rely entirely on the existing expression context generator API.
> QgsProcessingExpressionParameterDefinitionWidget::createExpressionContext should be based off QgsProcessingModelerParameterWidget::createExpressionContext(), but with removal of all the child algorithm specific scopes as these will have no meaning for a parameter default value (which sits at the model level, not a child algorithm level). So the created expression would be just the expression context taken from the processing context + a single QgsExpressionContextUtils::processingModelAlgorithmScope. For similar reasons all the highlighted variable/function handling isn't required, as that is all specific to the child algorithm level.

I initially looked into it but I couldn't get that to work properly somehow, I'll give it a second go shortly.

nyalldawson:
Superseded by https://github.com/qgis/QGIS/pull/41000

PR #39546, opened 2020-10-22
3d light default tweaks

nyalldawson:
Tweaks the configuration of newly created light sources for better (easier for users to understand) lighting, and switches newly created 3d maps to a single directional light instead of a single point light source.

### DIFF
--- a/src/ui/3d/qgslightswidget.ui
+++ b/src/ui/3d/qgslightswidget.ui
@@ -64,7 +64,7 @@
    <item>
     <widget class="QStackedWidget" name="mStackedWidget">
      <property name="currentIndex">
-      <number>2</number>
+      <number>1</number>
      </property>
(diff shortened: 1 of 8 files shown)


src/ui/3d/qgslightswidget.ui
```diff
@@ -111,11 +111,14 @@
...
         <widget class="QgsDoubleSpinBox" name="spinIntensity">
          <property name="decimals">
-          <number>1</number>
+          <number>2</number>
          </property>
          <property name="maximum">
           <double>999999.000000000000000</double>
          </property>
+         <property name="singleStep">
+          <double>0.100000000000000</double>
```
nirvn, line 120: @nyalldawson , good move.

nirvn (approved):

PR #39548, opened 2020-10-22
Fix duplicating features

3nids:
* adding child features was failing because the fields were not correctly set from the attribute table action
* the referenced feature is added before the referencing children (otherwise this leads to a foreign key constraint error if done in transaction mode)


signedav (approved):
Looks good. Thanks.

PR #39549, opened 2020-10-22
Update overlay_nearest function help and examples

### LINKED ISSUE #39069: Document that limit:=-1 is possible for geometry_overlay_* functions?
According to doc, `limit` expects an integer but using `limit:=-1` works and returns all the resulting features
image

Either fix the doc accordingly or constrain the value to integer. The value not being set already returns the whole results so IMHO -1 is not necessary.

DelazJ:
Trying to match logic and complete #39413 
Also fixes #39069  
image

### DIFF
--- a/resources/function_help/json/overlay_nearest
+++ b/resources/function_help/json/overlay_nearest
@@ -2,55 +2,85 @@
   "name": "overlay_nearest",
   "type": "function",
   "groups": ["GeometryGroup"],
-  "description": "This returns an array of results of an expression evaluated on features from a different layer ordered BY DISTANCE to the current feature, or, if no expression if provided, simply returns whether at least one feature from the other layer was found. Note : this function can be slow and consume a lot of memory for large layers.",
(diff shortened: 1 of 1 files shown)


resources/function_help/json/overlay_nearest
```diff
@@ -2,55 +2,64 @@
...
+      "description": "an optional expression to filter the target features to check. If not set, all the features in the target layer will be used.",
       "optional": true
     },
     {
       "arg": "limit",
-      "description": "an optional integer to limit the number of matching features (if not set, only the nearest feature will be returned)",
-      "optional": true
+      "description": "an optional integer to limit the number of matching features. If not set, only the nearest feature will be returned. If set to -1, returns all the matching features.",
+      "optional": true,
+      "default": "1"
```
DelazJ, line 25: I notice that limit:=0 has the same result as 1, and any value <0 returns all the matching features.
Not sure if of any relevance for users but wanted to share...

nyalldawson (approved):

PR #39553, opened 2020-10-22
PG: Fixes #39530 by handling unknown

### LINKED ISSUE #39530: DB Manager: SQL SELECT query skip fields when having fields with identical names
When executing a SELECT query in the DB Manager, if by chance two successive fields have the same name (as possibly in a JOIN) , the second name is replaced by the name of the next field whose content is in turn deleted.  Here's a simple example: 

`select 1 as a, 2 as b , 3 as b, 'xxy' as test;`

Gives :

> a        b        test
> 1        2             3

I noticed this strange behaviour in QGis 3.14.16 in Mac OS.
`

Version de QGIS | 3.14.16-Pi | Révision du code | a235a149f0
-- | -- | -- | --
Compilé avec Qt | 5.12.3 | Utilisant Qt | 5.12.3
Compilé avec GDAL/OGR | 2.4.1 | Utilisé avec GDAL/OGR | 2.4.1
Compilé avec SQLite | 3.28.0 | Fonctionne avec SQLite | 3.28.0
Version du client PostgreSQL | 11.3 | Version de SpatiaLite | 4.3.0a
Version de QWT | 6.1.4 | Version de QScintilla2 | 2.11.1
Version de l'OS | macOS High Sierra (10.13)
`
Capture d’écran 2020-10-21 à 16 28 54

elpaso:
Fixes #39530

@jef-n what do you think? Too dangerous?

### DIFF
--- a/src/providers/postgres/qgspostgresprovider.cpp
+++ b/src/providers/postgres/qgspostgresprovider.cpp
@@ -1146,6 +1146,12 @@ bool QgsPostgresProvider::loadFields()
           fieldType = QVariant::String;
           fieldSize = -1;
         }
+        else if ( fieldTypeName == QLatin1String( "unknown" ) )
+        {
+          // Assume it is convertible to text
+          fieldType = QVariant::String;
+          fieldSize = -1;
+        }
         else
         {
           QgsMessageLog::logMessage( tr( "Field %1 ignored, because of unsupported type %2" ).arg( fieldName, fieldTType ), tr( "PostGIS" ) );


jef-n:
Adding it to the "text"…fieldTypeName.startsWith( QLatin1String( "date" )  block would shorter.  Otherwise it probably nicer to have it handled as text than skipped completely.

elpaso:
> Adding it to the "text"…fieldTypeName.startsWith( QLatin1String( "date" ) block would shorter. Otherwise it probably nicer to have it handled as text than skipped completely.

Yes, I though about adding it to an existing case but I think it stands up better in the fallback `else` https://github.com/qgis/QGIS/pull/39553/files#diff-8fc1e19bdc3a51f281bf3789e7087de205ae9287ad1a30225aaaf7d92e033154R1139 where we handle unknown/exceptional cases.

PR #39559, opened 2020-10-22
Add QgsGeometryUtils::perpendicularCenterSegment

tschmetzer:
## Suggestion of a different perpendicularSegment geometry method

During my python scripting I was missing a function that I had to create
for myself in python which I needed in order to split a polyline using
QgsGeometry.splitGeometry().

The function I want to contribute will creates a perpendicular segment to
a given segment with it's center at a given point (the already available
function QgsGeometryUtils.perpendicularSegment doesn't fullfill that need)

The function has been discussed on the mailing list.


src/core/geometry/qgsgeometryutils.h
```diff
@@ -620,6 +620,40 @@ class CORE_EXPORT QgsGeometryUtils
...
+     * \param y2: y-coordinate of segmentPoint2, the segment's end point
+     *
+     * Result is a line (segment) centered in point p and perpendicular to segment [segmentPoint1, segmentPoint2]
+     *
+     * \param perpendicularSegmentPoint1x: x-coordinate of the perpendicularCenterSegment's start point
+     * \param perpendicularSegmentPoint1y: y-coordinate of the perpendicularCenterSegment's start point
+     * \param perpendicularSegmentPoint2x: x-coordinate of the perpendicularCenterSegment's end point
+     * \param perpendicularSegmentPoint2y: y-coordinate of the perpendicularCenterSegment's end point
+     *
+     * \param (optional parameter) segmentLength trims to given length. Default length is double the length of the input segment. Set to 0 for normalized length which is equal to 1.
```
nyalldawson, line 643:
```suggestion
     * May be used to split geometries. Unless \a segmentLength is specified the new centered perpendicular line segment will have double the length of the input segment.
     *
     * The result is a line (segment) centered in point p and perpendicular to segment [segmentPoint1, segmentPoint2].
     *
     * \param centerPointx x-coordinate of the point where the center of the perpendicular should be located
     * \param centerPointy y-coordinate of the point where the center of the perpendicular should be located
     * \param segmentPoint1x: x-coordinate of segmentPoint1, the segment's start point
     * \param segmentPoint1y: y-coordinate of segmentPoint1, the segment's start point
     * \param segmentPoint2x: x-coordinate of segmentPoint2, the segment's end point
     * \param y2: y-coordinate of segmentPoint2, the segment's end point
     * \param perpendicularSegmentPoint1x: x-coordinate of the perpendicularCenterSegment's start point
     * \param perpendicularSegmentPoint1y: y-coordinate of the perpendicularCenterSegment's start point
     * \param perpendicularSegmentPoint2x: x-coordinate of the perpendicularCenterSegment's end point
     * \param perpendicularSegmentPoint2y: y-coordinate of the perpendicularCenterSegment's end point
     * \param segmentLength trims to given length. Default length is double the length of the input segment. Set to 0 for normalized length which is equal to 1.
```

Gotta remove those empty lines and ensure all param annotations are consecutive, otherwise the sipify script can't do it's magic completely.

src/core/geometry/qgsgeometryutils.h
```diff
@@ -620,6 +620,40 @@ class CORE_EXPORT QgsGeometryUtils
...
+     * \since QGIS 3.17?
+     *
+     */
+
+    static void perpendicularCenterSegment(   double centerPointx, double centerPointy,
+                                              double segmentPoint1x, double segmentPoint1y,
+                                              double segmentPoint2x, double segmentPoint2y,
+                                              double &perpendicularSegmentPoint1x SIP_OUT, double &perpendicularSegmentPoint1y SIP_OUT,
+                                              double &perpendicularSegmentPoint2x SIP_OUT, double &perpendicularSegmentPoint2y SIP_OUT,
+                                              double *segmentLength=nullptr
```
nyalldawson, line 655: What do you think about changing this to

    double segmentLength = 0

where a segmentLength of 0 means "default length". If someone wants a normalized segment they can easily set this to 1 then.

tests/src/core/testqgsgeometryutils.cpp
```diff
@@ -720,6 +721,76 @@ void TestQgsGeometryUtils::testPerpendicularSegment()
...
+    double perpendicularSegmentPoint1x=0, perpendicularSegmentPoint1y=0, perpendicularSegmentPoint2x=0, perpendicularSegmentPoint2y=0, segmentLength=0;
+
+    // default case 1: centerPoint and perpendicular line on given segment without segmentLength
+    QgsPoint centerPoint(2,1.5);
+    QgsPoint segmentPoint1(2,1);
+    QgsPoint segmentPoint2(2,2);
+
+    QgsGeometryUtils::perpendicularCenterSegment( centerPoint.x(),centerPoint.y(), segmentPoint1.x(), segmentPoint1.y(), segmentPoint2.x(), segmentPoint2.y(),
+            perpendicularSegmentPoint1x, perpendicularSegmentPoint1y, perpendicularSegmentPoint2x, perpendicularSegmentPoint2y);
+    QCOMPARE( perpendicularSegmentPoint1x, (3) );
```
nyalldawson, line 736:
```suggestion
    QCOMPARE( perpendicularSegmentPoint1x, 3.0 );
```
etc

roya0045:
Wouldn't it better to use QgsPointXY instead of having each 'entity' as two inputs/outputs?

nyalldawson:
> Wouldn't it better to use QgsPointXY instead of having each 'entity' as two inputs/outputs?


See the conversation on the qgis-developer mailing list -- by using doubles instead of QgsPointXY we avoid unnecessary conversions if the caller has a different object type (e.g. QgsPoint instead of QgsPointXY)

src/core/geometry/qgsgeometryutils.h
tschmetzer: I think for python scripters it is easier the less arguments there are to fill when there's a function they need to use. Based on that usability perspective I would consider an optional argument more helpful. What do you think?
Good idea! Semantics wise I agree to your proposal.

tschmetzer:
I must admit that I haven't fully grasped the background behind using doubles instead of QgsPointXY/QgsPoint. I don't want to deep dive into that discussion (again) but I agree and have the feeling that being able to use QgsPoint and QgsPointXY would probably be very favorable for both programmers as well as scriptwriters. Would be good to have a long-term solution. The mixture of coordinates/QgsPointXY/QgsPoint is not intuitive. Unfortunately I couldn't find a search function for the qgis-developer mailing list. I can only view the archive month wise.

src/core/geometry/qgsgeometryutils.h
nyalldawson: I agree that optional is the way to go. I just find a "three option" argument confusing (nullptr, 0 or value > 0). That's why I'd prefer optional with default as 0 and skip the nullptr option...

nyalldawson:
> The mixture of coordinates/QgsPointXY/QgsPoint is not intuitive.

Here's a super-short summary:
1. QgsPointXY is a purely 2d point, no z or m value handling. It's designed to be very lightweight and memory efficient for situations where a point will NEVER have z or m values. 
2. QgsPoint can be used to optionally store z and m values. It's a much heavier class, not as memory efficient, and slower to use because of QgsAbstractGeometry inheritance. QgsPointXY is (sometimes, often?) a better class to use for performance critical code (that is, unless the input values are ALREADY QgsPoint values, and you want to avoid the cost of conversion!)
3. raw doubles: work for either classes, and are the BEST choice for code which is absolutely performance critical.

In this case 3 is the right choice -- the QgsGeometryUtils methods are low level utilities designed for optimal performance. If you want API convenience, you could add QgsPoint/QgsPointXY overrides for this function (which just calls the version with raw doubles, i.e. no code duplication). But you'd add those after getting the lowest level version merged in first.

src/core/geometry/qgsgeometryutils.h
tschmetzer: I understand we don't need two options to achieve the same objective (nullptr, segmentLength=0).
By optional I meant having the choice to add/use the segmentLength argument or not. On one hand I can see an advantage of forcing the script user to actively think about what length to choose for his task. On the other hand I guess the standard use case is probably that the script user will not need to set any length for his task and may not want to care about an additional, irrelevant argument to be set. 
I don't consider this a safety critical function and can see a benefit in user friendliness with the latter option. It can anyhow be confusing as it came to my mind that a segmentLength of 0 would have the same effect as a segmentLength of 2. 
What do you think of keeping the nullptr and eliminating the confusing 0 value option meaning the argument accepts only values >0 ?

nyalldawson: Using a pointer here is really not ideal, because it makes the function harder to call when you DO want a specific segment length. You'd need to do something like:

    double myLength = 5;
    QgsGeometryUtils::perpendicularCenterSegment( ...... , &myLength );

instead of just

    QgsGeometryUtils::perpendicularCenterSegment( ...... , 5 );

I'm all for making the argument optional, but I'd do that by making the default value 0. ie. the signature becomes:


```
    ...
     * \param segmentLength (optional) Trims to given length. A segmentLength value of 0 refers to the default length which is double the length of the input segment. Set to 1 for a normalized length.
    ...
     */
    ...
    static void perpendicularCenterSegment( double centerPointx, double centerPointy,
    ...
                                            double segmentLength = 0
                                          ) SIP_HOLDGIL;
```

Then it's easy to call by default if you don't care about the segment length, and easy for users to specify a particular desired length (including "1" if they want a normalized segment)

What do you think?

nyalldawson:

src/core/geometry/qgsgeometryutils.h
tschmetzer: Agreed

PR #39568, opened 2020-10-23
Fix #38468 NOT taking the range.begin into account

### LINKED ISSUE #38468: TimeController taking BOTH upper and lower limit into account of timeframes
See also some info in this Thread:
https://lists.osgeo.org/pipermail/qgis-developer/2020-August/062066.html

The idea is that when you time-filter data with a frame of 60 minutes, that the lower limit is taken, but the upper limit is not
Screenshot-20200827094104-971x539

This is a project including data (6 points, with start and end time, frames of one hour):
[TimeTestDataProject.zip](https://github.com/qgis/QGIS/files/5134771/TimeTestDataProject.zip)

This is a movie to show that apparently the lower AND upper limit are taken into account:
[timedots_movie.zip](https://github.com/qgis/QGIS/files/5134775/timedots_movie.zip)

Not that you can see that the upper limit is taken into account using these configurations:
- Single Field with Date/Time with only t_start and a Event duration of 60 minutes
- Separate Fields for Start and End 
- Using expressions eg:  to_time(  "t_start" ) and  to_time(  "t_end" )
I did not test 'Separate field for Start and Event Duration'... but I think that will also have the issue

rduivenvoorde:
See #38468 for more discussion.

Reasoning is: if working with events with duration of exact 1hour
you do NOT want to see 2 of them show up if your temporal view
frame is also exact 1 hour

I added some comments in the code, because for me it was pretty hard to find this out. 
If for others though this is very clear, feel free to remove the commit

Mmm, maybe this deserves a test...  what about pulling this in 3.16 and I promise I create one :-)
(creating tests takes me always so much time...)

### DIFF
--- a/src/core/qgsvectorlayertemporalproperties.cpp
+++ b/src/core/qgsvectorlayertemporalproperties.cpp
@@ -440,10 +440,10 @@ QString QgsVectorLayerTemporalProperties::createFilterString( const QgsVectorLay
       }
       else
       {
-        return QStringLiteral( "(%1 %2 %3 AND %1 %4 %5) OR %1 IS NULL" ).arg( QgsExpression::quotedColumnRef( mStartFieldName ),
-               range.includeBeginning() ? QStringLiteral( ">=" ) : QStringLiteral( ">" ),
+        // Working with features with events with a duration, so taking this duration into account (+ QgsInterval( -mFixedDuration, mDurationUnit ) ))
+        // Then we are NOT taking the range.includeBeginning() and range.includeEnd() into account (deliberatly, see #38468)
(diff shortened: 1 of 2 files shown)


nyalldawson:
There's existing tests for this you'll need to update

rduivenvoorde:
Argh... was adding a whole bunch of QgsVectorLayerTemporalProperties.checkFilterString tests in 
``tests/src/core/testqgstemporalproperty.cpp``
but now see (in Travis) that this is already covered in a python test... soo much to know...

rduivenvoorde:
Mmm, oracle test error? Seems not related?

PR #39577, opened 2020-10-23
Hidden layers

elpaso:
With API support for text filtering, thanks to @nyalldawson for the proxy model implementation.

qgis-hidden-layers

Funded by **ARPA Piemonte**

### DIFF
--- a/src/app/qgslayercapabilitiesmodel.cpp
+++ b/src/app/qgslayercapabilitiesmodel.cpp
@@ -157,46 +157,49 @@ Qt::ItemFlags QgsLayerCapabilitiesModel::flags( const QModelIndex &idx ) const
   }
   else
   {
-    if ( idx.column() == IdentifiableColumn )
+    switch ( idx.column() )
     {
-      if ( layer->isSpatial() )
+      case IdentifiableColumn:
+      {
+        if ( layer->isSpatial() )
+        {
+          return Qt::ItemIsEnabled | Qt::ItemIsSelectable | Qt::ItemIsEditable | Qt::ItemIsUserCheckable;
+        }
+        else
+        {
+          return Qt::ItemFlags();
+        }
+      }
+      case ReadOnlyColumn:
+      {
+        if ( layer->type() == QgsMapLayerType::VectorLayer )
+        {
+          return Qt::ItemIsEnabled | Qt::ItemIsSelectable | Qt::ItemIsEditable | Qt::ItemIsUserCheckable;
+        }
+        else
+        {
+          return Qt::ItemFlags();
+        }
+      }
+      case SearchableColumn:
+      {
+        if ( layer->type() == QgsMapLayerType::VectorLayer )
+        {
+          return Qt::ItemIsEnabled | Qt::ItemIsSelectable | Qt::ItemIsEditable | Qt::ItemIsUserCheckable;
+        }
+        else
+        {
+          return Qt::ItemFlags();
+        }
+      }
+      case RequiredColumn:
       {
         return Qt::ItemIsEnabled | Qt::ItemIsSelectable | Qt::ItemIsEditable | Qt::ItemIsUserCheckable;
       }
-      else
-      {
-        return Qt::ItemFlags();
-      }
-    }
-    else if ( idx.column() == ReadOnlyColumn )
-    {
-      if ( layer->type() == QgsMapLayerType::VectorLayer )
+      case HiddenColumn:
       {
         return Qt::ItemIsEnabled | Qt::ItemIsSelectable | Qt::ItemIsEditable | Qt::ItemIsUserCheckable;
       }
-      else
-      {
-        return Qt::ItemFlags();
-      }
-    }
-    else if ( idx.column() == SearchableColumn )
-    {
-      if ( layer->type() == QgsMapLayerType::VectorLayer )
-      {
-        return Qt::ItemIsEnabled | Qt::ItemIsSelectable | Qt::ItemIsEditable | Qt::ItemIsUserCheckable;
-      }
-      else
-      {
-        return Qt::ItemFlags();
-      }
-    }
-    else if ( idx.column() == RequiredColumn )
-    {
-      return Qt::ItemIsEnabled | Qt::ItemIsSelectable | Qt::ItemIsEditable | Qt::ItemIsUserCheckable;
-    }
-    else if ( idx.column() == HiddenColumn )
-    {
-      return Qt::ItemIsEnabled | Qt::ItemIsSelectable | Qt::ItemIsEditable | Qt::ItemIsUserCheckable;
     }
   }
   return Qt::ItemFlags();
--- a/src/app/qgslayercapabilitiesmodel.cpp
+++ b/src/app/qgslayercapabilitiesmodel.cpp
@@ -32,6 +32,7 @@ QgsLayerCapabilitiesModel::QgsLayerCapabilitiesModel( QgsProject *project, QObje
     mSearchableLayers.insert( it.value(), it.value()->type() == QgsMapLayerType::VectorLayer && it.value()->flags().testFlag( QgsMapLayer::Searchable ) );
     mIdentifiableLayers.insert( it.value(), it.value()->flags().testFlag( QgsMapLayer::Identifiable ) );
     mRemovableLayers.insert( it.value(), it.value()->flags().testFlag( QgsMapLayer::Removable ) );
+    mHiddenLayers.insert( it.value(), it.value()->flags().testFlag( QgsMapLayer::Hidden ) );
   }
 }
 
@@ -76,6 +77,11 @@ bool QgsLayerCapabilitiesModel::removable( QgsMapLayer *layer ) const
   return mRemovableLayers.value( layer, true );
 }
 
+bool QgsLayerCapabilitiesModel::hidden( QgsMapLayer *layer ) const
+{
+  return mHiddenLayers.value( layer, true );
+}
+
 bool QgsLayerCapabilitiesModel::readOnly( QgsMapLayer *layer ) const
 {
   return mReadOnlyLayers.value( layer, true );
@@ -89,7 +95,7 @@ bool QgsLayerCapabilitiesModel::searchable( QgsMapLayer *layer ) const
 int QgsLayerCapabilitiesModel::columnCount( const QModelIndex &parent ) const
 {
   Q_UNUSED( parent )
-  return 5;
+  return 6;
 }
 
 QVariant QgsLayerCapabilitiesModel::headerData( int section, Qt::Orientation orientation, int role ) const
@@ -110,6 +116,8 @@ QVariant QgsLayerCapabilitiesModel::headerData( int section, Qt::Orientation ori
           return tr( "Searchable" );
         case RequiredColumn:
           return tr( "Required" );
+        case HiddenColumn:
+          return tr( "Hidden" );
         default:
           return QVariant();
       }
@@ -125,6 +133,8 @@ QVariant QgsLayerCapabilitiesModel::headerData( int section, Qt::Orientation ori
           return QVariant();
         case RequiredColumn:
           return tr( "Layers which are protected from inadvertent removal from the project." );
+        case HiddenColumn:
+          return tr( "Layers which not shown in the legend." );
         default:
           return QVariant();
       }
@@ -181,6 +191,10 @@ Qt::ItemFlags QgsLayerCapabilitiesModel::flags( const QModelIndex &idx ) const
       }
     }
     else if ( idx.column() == RequiredColumn )
+    {
+      return Qt::ItemIsEnabled | Qt::ItemIsSelectable | Qt::ItemIsEditable | Qt::ItemIsUserCheckable;
+    }
+    else if ( idx.column() == HiddenColumn )
     {
       return Qt::ItemIsEnabled | Qt::ItemIsSelectable | Qt::ItemIsEditable | Qt::ItemIsUserCheckable;
     }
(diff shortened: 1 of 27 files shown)


elpaso:
@nyalldawson there are still a few warnings I couldn't fix:

```c++
QAbstractItemModel::endRemoveRows:  Invalid index ( 1 , 0 ) in model QgsLayerTreeProxyModel(0x2609340)
QAbstractItemModel::endRemoveRows:  Invalid index ( 1 , 0 ) in model QgsLayerTreeProxyModel(0x24d2970)
```

Maybe you can have a look?

src/app/qgslayercapabilitiesmodel.cpp
```diff
@@ -125,6 +133,8 @@ QVariant QgsLayerCapabilitiesModel::headerData( int section, Qt::Orientation ori
           return QVariant();
         case RequiredColumn:
           return tr( "Layers which are protected from inadvertent removal from the project." );
+        case HiddenColumn:
+          return tr( "Layers which not shown in the legend." );
```
m-kuhn, line 137:
```suggestion
          return tr( "Layers which are not shown in the legend." );
```

src/app/qgsprojectlayergroupdialog.cpp
```diff
@@ -135,7 +132,7 @@ QString QgsProjectLayerGroupDialog::selectedProjectFile() const
 
 bool QgsProjectLayerGroupDialog::isValid() const
 {
-  return nullptr != mTreeView->layerTreeModel();
+  return static_cast< bool >( mTreeView->layerTreeModel() );
```
m-kuhn, line 135: Why is this changed?

src/core/layertree/qgslayertreemodel.cpp
```diff
@@ -34,7 +34,7 @@
 #include "qgssymbollayerutils.h"
 #include "qgsvectorlayer.h"
 #include "qgslayerdefinition.h"
-
+#include "qgsmessagelog.h"
```
m-kuhn, line 37: Leftover?

m-kuhn (commented):
Very nice

m-kuhn:
cc @gacarrillor

src/app/qgsprojectlayergroupdialog.cpp
elpaso: I'm not sure (it's from a Nyall's commit), but I think it's much clearer/explicit this way.

src/core/layertree/qgslayertreemodel.cpp
elpaso: Yep.

src/app/qgslayercapabilitiesmodel.cpp
```diff
@@ -184,6 +194,10 @@ Qt::ItemFlags QgsLayerCapabilitiesModel::flags( const QModelIndex &idx ) const
     {
       return Qt::ItemIsEnabled | Qt::ItemIsSelectable | Qt::ItemIsEditable | Qt::ItemIsUserCheckable;
     }
+    else if ( idx.column() == HiddenColumn )
```
3nids, line 197: could you turn this into a switch?

3nids (commented):
I made some thinking into this in the past, and I had quite some unasnwerer questions....I'll write what I have in mind to start discussing this.

What happens in QgsMapLayerComboBox? (Does this become a choice in the API?) If layers are hidden, when does it make sense to show them to the user?
What happens in other layer tree models, mainly snapping (DXF, and maybe others)?

If this info is for the main layer tree only, wouldn't it make sense to save this as a property of the nodes rather than as a layer capability? This would allow hiding groups at once which would be very handy.

If we stick to layer capabilities, if you hide all layers of layer tree group what happens to it? You show it empty?

nyalldawson:
> What happens in QgsMapLayerComboBox? 

Personally I think they should also be hidden here. I would see this flag as being useful for "internal details", like relationship child tables. You need them in the project, but don't want to clutter up the UI with them anywhere...

elpaso:
> I made some thinking into this in the past, and I had quite some unasnwerer questions....I'll write what I have in mind to start discussing this.
> 
> What happens in QgsMapLayerComboBox? (Does this become a choice in the API?) If layers are hidden, when does it make sense to show them to the user?
> What happens in other layer tree models, mainly snapping (DXF, and maybe others)?
> 

The primary use case that I had in mind here was hiding the layer from the legend, but I agree with Nyall that the use of the proxy model could be extended for other layer selection tools, for this reason I think the flag probably belongs to the layer and not the legend node.


> If this info is for the main layer tree only, wouldn't it make sense to save this as a property of the nodes rather than as a layer capability? This would allow hiding groups at once which would be very handy.
> 

True, but I don't see a real use case for this. We are not talking about layer/group canvas visibility here (which is already available), hiding layer is meant for accessory layers (joined, lookups for value-relations etc., most probably aspatial) that you still need in a project but you don't want them to pollute the legend and other layer selection tools. 

> If we stick to layer capabilities, if you hide all layers of layer tree group what happens to it? You show it empty?

For the empty groups (when they only contain hidden layers), I don't think they should be automatically hidden, if the user created a group he may want to use it as a container for other layers he will be putting in the group or he can remove it himself. The question is what to do when a group that contains hidden layers is removed. But I think we stick to the current behavior, it would be nice to show the number of layers being removed in the confirmation dialog though.

3nids:
> The primary use case that I had in mind here was hiding the layer from the legend, but I agree with Nyall that the use of the proxy model could be extended for other layer selection tools

+1

> for this reason I think the flag probably belongs to the layer and not the legend node.

this could also be achieved by giving the layer tree as argument to the map layer combobox

> True, but I don't see a real use case for this. We are not talking about layer/group canvas visibility here (which is already available), hiding layer is meant for accessory layers (joined, lookups for value-relations etc., most probably aspatial) that you still need in a project but you don't want them to pollute the legend and other layer selection tools.

in my experience, these layers are often a few ones and grouped. 
But I don't have a strong opinion.

> For the empty groups (when they only contain hidden layers), I don't think they should be automatically hidden, if the user created a group he may want to use it as a container for other layers he will be putting in the group or he can remove it himself. The question is what to do when a group that contains hidden layers is removed. But I think we stick to the current behavior, it would be nice to show the number of layers being removed in the confirmation dialog though.

If we do this, we should warn the user about hidden layers being removed (and that it could break the project).

elpaso:
> > For the empty groups (when they only contain hidden layers), I don't think they should be automatically hidden, if the user created a group he may want to use it as a container for other layers he will be putting in the group or he can remove it himself. The question is what to do when a group that contains hidden layers is removed. But I think we stick to the current behavior, it would be nice to show the number of layers being removed in the confirmation dialog though.
> 
> If we do this, we should warn the user about hidden layers being removed (and that it could break the project).

Definitely, I'll work on that.

gacarrillor:
The discussion we had here may be of interest:

https://github.com/qgis/QGIS-Enhancement-Proposals/issues/155

elpaso:
@gacarrillor sorry I missed (or totally forget) that QEP, thanks for the pointer. 

But I see there wasn't much consensus, there was any progress from sept. 2019?

Anyway, what do we do now?

Me (and Nyall) have already spent a considerable time on this development and it would be a pity to throw it away. 
In this PR we have a nice proxy model, that can be potentially used to filter layers by name too (other criteria can be added).

AFAICS the only use case that won't be covered by my implementation is hiding groups, that I don't see a very strong use case (why would you create a group if you mean to hide everything in it in the first place?) .

gacarrillor:
> @gacarrillor sorry I missed (or totally forget) that QEP, thanks for the pointer.
> 
> But I see there wasn't much consensus, there was any progress from sept. 2019?
> 
> Anyway, what do we do now?
> 
> Me (and Nyall) have already spent a considerable time on this development and it would be a pity to throw it away.
> In this PR we have a nice proxy model, that can be potentially used to filter layers by name too (other criteria can be added).

No worries @elpaso. I've seen the progress you two have, which is substantial! Please go ahead with the implementation. After all, the goal is to benefit the community (and for sure we all will)!

I just put the link here for you to read all the concerns raised there (e.g., by @andreasneumann), so that you can keep them in mind.

From the discussion in the QEP, I see a key use case which is to filter layers in the QMapLayerComboBox and in the Processing dialogs, which you seem to be considering now.

> AFAICS the only use case that won't be covered by my implementation is hiding groups, that I don't see a very strong use case (why would you create a group if you mean to hide everything in it in the first place?) .

Regarding groups, you might want to put all your look-up tables in a group and just hide that single group, or all your base maps or whatever. However, I leave it to you to decide on whether this PR should cover such case.

elpaso:
@gacarrillor thank you for the feedback, let me think about it.

src/app/qgslayercapabilitiesmodel.cpp
elpaso: Sure, it's just funny you are asking  .. https://github.com/qgis/QGIS/commit/6b66b2048a7d06b7db9ae33054fe3b2d5314e571#diff-5e009636e6446c3a89221a5d469cf0619daade026ea52c7cea0abb66888750aeR129

3nids: sorry, probably because of the first if on top (LayerColumn).
let's move it to the switch condition too, and do `if ( !layer && idx.column() != LayerColumn )`

3nids: (as for now you'll have a warning since LayerColumn is not handled in the switch)

elpaso: no warnings, because there is no cast, I'll add it.

elpaso:
@gacarrillor I thought about it an I'm not convinced about linking "hidden" to the tree (instead of the layer) and I'm hesitant to add a group hide checkbox that would be stored as a property of the node either (it wouldn't be hard though).

For the server part (@andreasneumann) , I think we need to add the "hidden" flag information to the getProjectSettings and to the new landing page API calls, I'll do that in a followup PR.

There are already configuration options to choose which layers are published as WMS/WFS so I don't think we need to override them. Just providing that piece of information to the API calls that are meant to be consumed by web applications will do just fine.

3nids:
As far as I am concerned, I am happy with current implementation.

One remark, the denomination of "hidden" layer is counter-intuitive to me. What about "secret layer"? Not so sure, but I think we shouldn't use a word related to visibility for this.

Also, do you consider adding a switch to the layer tree to show the "secret layers"? (so that we can still easily access their properties or their map visibility).

elpaso:
> As far as I am concerned, I am happy with current implementation.
> 
> One remark, the denomination of "hidden" layer is counter-intuitive to me. What about "secret layer"? Not so sure, but I think we shouldn't use a word related to visibility for this.

Good point. I'll leave the choice to native speakers @nyalldawson , @timlinux, @saberraz (any other?)

> 
> Also, do you consider adding a switch to the layer tree to show the "secret layers"? (so that we can still easily access their properties or their map visibility).

Good idea, a bit tricky from the UI/UX POV though... we don't have much space in the TOC toolbar ... do you have anything in mind?

3nids:
> Good idea, a bit tricky from the UI/UX POV though... we don't have much space in the TOC toolbar ... do you have anything in mind?

You could turn the filter button into a menu ?

elpaso:
> > Good idea, a bit tricky from the UI/UX POV though... we don't have much space in the TOC toolbar ... do you have anything in mind?
> 
> You could turn the filter button into a menu ?

qgis-hidden-layers-menu

I'm still in doubts about the "hidden"/"secret" but if you are sure that it fits better I have no objections to change it (and rename the API accordingly), my main concern is that in my mind I associate "secret" with credentials, so I immediately think at passwords and such when I hear the words "secret", but maybe it's just me.

nyalldawson:
I'm -1 on secret. I think "hidden" is the correct terminology here.

nyalldawson:
Another option could be "private layers"

elpaso:
We settled on "Private Layers", because "hidden" is already used in the legend GUI to indicate layers that are not currently visible.

nyalldawson:
@elpaso this is working nicely thanks!

One thing I think we should consider is adding a toggle for showing private layers in the layer tree toolbar. I.e. under the "eye" icon have checkable entry for "show private layers" (defaults to off). This would let users temporarily show these so that they can make changes to them without having to edit their "private" status first. What do you think?

3nids:
@nyalldawson see https://github.com/qgis/QGIS/pull/39577#issuecomment-720470622 and discussion above.

I don't think it belongs to the "eye" menu as this is toggling visibility in the map, not in the legend. That's why I proposed to integrate this in the filter menu.

nyalldawson:
@3nids that works

nirvn:
@elpaso , this appears to have broken layer trees of pre-existing projects. I get the following errors when loading a project and half of my layer tree items are missing:

2020-11-06T14:50:14     CRITICAL    Qt : QAbstractItemModel::endInsertRows: Invalid index ( 14 , 0 ) in model QgsLayerTreeProxyModel(0x556066d4a680)
2020-11-06T14:50:14     CRITICAL    Qt : QAbstractItemModel::endInsertRows: Invalid index ( 13 , 0 ) in model QgsLayerTreeProxyModel(0x556066d4a680)
2020-11-06T14:50:14     CRITICAL    Qt : QAbstractItemModel::endInsertRows: Invalid index ( 15 , 0 ) in model QgsLayerTreeProxyModel(0x556066d4a680)
2020-11-06T14:50:14     CRITICAL    Qt : QAbstractItemModel::endInsertRows: Invalid index ( 20 , 0 ) in model QgsLayerTreeProxyModel(0x556066d4a680)

elpaso:
@nirvn can you attach a sample project?

nirvn:
@elpaso , here: [CambodiaSoldOut.zip](https://github.com/qgis/QGIS/files/5499882/CambodiaSoldOut.zip)

elpaso:
@nirvn thanks, I'll look into it soon.

borysiasty:
I have the same problem as @nirvn. A lot of layers is gone from the main layer tree and from QgsProjectLayerGroupDialog, but they all are still present in the Project Properties dialog.

elpaso:
@borysiasty I'm on it: https://github.com/qgis/QGIS/pull/39872 please be patient because I'm not yet done.

PR #39580, opened 2020-10-24
Let gdal:convertformat convert all layers at once

karatemuffin:
## Problem 

The gdal:convertformat(aka ogr2ogr.py) wrapper around ogr2ogr is only converting one layer of a source file at the moment. 
e.g.
When setting the following options:
```
INPUT: https://download.geofabrik.de/europe/austria-latest.osm.pbf
OUTPUT: somefile.sqlite
```
they are translated to the following console command
`ogr2ogr -f "SQLite" OUTPUT INPUT points`
where _points_ is describing the layer to convert, and the user isnt able to change that. 

## Changes
I added a Boolean Parameter to the Plugin that lets you convert all layers of a source file.
When active it eliminates the _layerName_ being added at the end of the console command. 
The default behavior to import only 1 automatically chosen layer has been kept. 
Added also the Translation for the languages i speak.

### DIFF
--- a/python/plugins/processing/algs/gdal/ogr2ogr.py
+++ b/python/plugins/processing/algs/gdal/ogr2ogr.py
@@ -37,7 +37,7 @@ from processing.algs.gdal.GdalUtils import GdalUtils
 class ogr2ogr(GdalAlgorithm):
     INPUT = 'INPUT'
     OPTIONS = 'OPTIONS'
-    LAYER = 'LAYER'
+    KEEP_LAYERS = 'LAYER'
     OUTPUT = 'OUTPUT'
 
     def __init__(self):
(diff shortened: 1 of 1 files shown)


python/plugins/processing/algs/gdal/ogr2ogr.py
```diff
@@ -36,6 +37,7 @@
 class ogr2ogr(GdalAlgorithm):
     INPUT = 'INPUT'
     OPTIONS = 'OPTIONS'
+    LAYER = 'LAYER'
```
roya0045, line 40: This parameter name is quite misleading. something like 'KEEP_LAYERS' seems more fiting.

karatemuffin: I renamed the parameter as you suggested.

karatemuffin (commented):
works as expected

nyalldawson:
Manually marking as stale because stalebot has left for christmas break

PR #39582, opened 2020-10-24
[RPM] Make spec file compatible with Fedora 33

daniviga:
## Description

Update the QGIS spec file to make it compile on Fedora 33. The change is backward compatible with any currently supported Fedora release (31 and 32)

Needs backport to 3.10 and 3.16


daniviga:
cc @m-kuhn

m-kuhn:
Thanks @daniviga

daniviga:
Thanks to you @m-kuhn

PR #39584, opened 2020-10-24
[Georeferencer] Reproject data points in the desired output projection

roya0045:
## Description

This is a proposal to improve the georeferencer plugin in a way that doesn't force the user to set the canvas to the same crs as the desired output.

To do this I am storing the original projection of the points and reprojecting them when needed to georeference the file.

I have noticed that moving those points is quite finnicky, I'm not too sure if this is due to these changes or just my testing VM/build.

Still need to work on the UI ti add the crs selector properly.

### DIFF
--- a/src/app/georeferencer/qgsgcplist.cpp
+++ b/src/app/georeferencer/qgsgcplist.cpp
@@ -15,6 +15,9 @@
 
 #include "qgspointxy.h"
 #include "qgsgeorefdatapoint.h"
+#include "qgscoordinatereferencesystem.h"
+#include "qgscoordinatetransform.h"
+#include "qgsproject.h"
 
 #include "qgsgcplist.h"
 
@@ -30,16 +33,33 @@ QgsGCPList::QgsGCPList( const QgsGCPList &list )
   }
 }
 
-void QgsGCPList::createGCPVectors( QVector<QgsPointXY> &mapCoords, QVector<QgsPointXY> &pixelCoords )
+void QgsGCPList::createGCPVectors( QVector<QgsPointXY> &mapCoords, QVector<QgsPointXY> &pixelCoords, const QgsCoordinateReferenceSystem targetCrs )
 {
   mapCoords   = QVector<QgsPointXY>( size() );
   pixelCoords = QVector<QgsPointXY>( size() );
+  QgsPointXY transCoords;
   for ( int i = 0, j = 0; i < sizeAll(); i++ )
   {
     QgsGeorefDataPoint *pt = at( i );
     if ( pt->isEnabled() )
     {
-      mapCoords[j] = pt->mapCoords();
+      if ( targetCrs.isValid() )
+      {
+        try
+        {
+          transCoords =  QgsCoordinateTransform( pt->crs(), targetCrs,
+                                                 QgsProject::instance() ).transform( pt->mapCoords() );
+          mapCoords[j] = transCoords;
+          pt->setTransCoords( transCoords );
+        }
+        catch ( const QgsException &e )
+        {
+          Q_UNUSED( e );
+          mapCoords[j] = pt->mapCoords();
+        }
+      }
+      else
+        mapCoords[j] = pt->mapCoords();
       pixelCoords[j] = pt->pixelCoords();
       j++;
     }
--- a/src/app/georeferencer/qgsgcplist.h
+++ b/src/app/georeferencer/qgsgcplist.h
@@ -21,6 +21,7 @@
 
 class QgsGeorefDataPoint;
 class QgsPointXY;
+class QgsCoordinateReferenceSystem;
 
 // what is better use inherid or agrigate QList?
 class QgsGCPList : public QList<QgsGeorefDataPoint *>
@@ -29,7 +30,7 @@ class QgsGCPList : public QList<QgsGeorefDataPoint *>
     QgsGCPList() = default;
     QgsGCPList( const QgsGCPList &list );
 
-    void createGCPVectors( QVector<QgsPointXY> &mapCoords, QVector<QgsPointXY> &pixelCoords );
+    void createGCPVectors( QVector<QgsPointXY> &mapCoords, QVector<QgsPointXY> &pixelCoords, const QgsCoordinateReferenceSystem targetCrs );
     int size() const;
     int sizeAll() const;
 
--- a/src/app/georeferencer/qgsgeorefdatapoint.cpp
+++ b/src/app/georeferencer/qgsgeorefdatapoint.cpp
@@ -16,22 +16,25 @@
 
 #include "qgsmapcanvas.h"
 #include "qgsgcpcanvasitem.h"
+#include "qgscoordinatereferencesystem.h"
+
 
 #include "qgsgeorefdatapoint.h"
 
 QgsGeorefDataPoint::QgsGeorefDataPoint( QgsMapCanvas *srcCanvas, QgsMapCanvas *dstCanvas,
                                         const QgsPointXY &pixelCoords, const QgsPointXY &mapCoords,
-                                        bool enable )
+                                        const QgsCoordinateReferenceSystem proj, bool enable )
   : mSrcCanvas( srcCanvas )
   , mDstCanvas( dstCanvas )
   , mPixelCoords( pixelCoords )
   , mMapCoords( mapCoords )
   , mId( -1 )
+  , mCrs( proj )
   , mEnabled( enable )
 {
+  mTransCoords = QgsPointXY( mapCoords );
   mGCPSourceItem = new QgsGCPCanvasItem( srcCanvas, this, true );
   mGCPDestinationItem = new QgsGCPCanvasItem( dstCanvas, this, false );
-
   mGCPSourceItem->setEnabled( enable );
   mGCPDestinationItem->setEnabled( enable );
   mGCPSourceItem->show();
@@ -46,8 +49,10 @@ QgsGeorefDataPoint::QgsGeorefDataPoint( const QgsGeorefDataPoint &p )
 //  mGCPDestinationItem = new QgsGCPCanvasItem(p.dstCanvas(), p.pixelCoords(), p.mapCoords(), p.isEnabled());
   mPixelCoords = p.pixelCoords();
   mMapCoords = p.mapCoords();
+  mTransCoords = p.transCoords();
   mEnabled = p.isEnabled();
   mResidual = p.residual();
+  mCrs = p.crs();
   mId = p.id();
 }
 
@@ -75,6 +80,24 @@ void QgsGeorefDataPoint::setMapCoords( const QgsPointXY &p )
   {
     mGCPDestinationItem->update();
   }
+}
+
+void QgsGeorefDataPoint::setTransCoords( const QgsPointXY &p )
+{
+  mTransCoords = p;
+  if ( mGCPSourceItem )
+  {
+    mGCPSourceItem->update();
+  }
+  if ( mGCPDestinationItem )
+  {
+    mGCPDestinationItem->update();
+  }
+}
+
+QgsPointXY QgsGeorefDataPoint::transCoords() const
+{
+  return mTransCoords.isEmpty() ? mMapCoords : mTransCoords;
 }
 
 void QgsGeorefDataPoint::setEnabled( bool enabled )
@@ -147,6 +170,10 @@ void QgsGeorefDataPoint::moveTo( QPoint p, bool isMapPlugin )
   {
     QgsPointXY pnt = mGCPDestinationItem->toMapCoordinates( p );
     mMapCoords = pnt;
+    if ( mSrcCanvas && !mSrcCanvas->mapSettings().destinationCrs().isValid() )
+      mCrs = mSrcCanvas->mapSettings().destinationCrs();
+    else
+      mCrs = mGCPDestinationItem->canvas()->mapSettings().destinationCrs() ;
   }
   mGCPSourceItem->update();
   mGCPDestinationItem->update();
--- a/src/app/georeferencer/qgsgeorefdatapoint.h
+++ b/src/app/georeferencer/qgsgeorefdatapoint.h
@@ -17,6 +17,7 @@
 #define QGSGEOREFDATAPOINT_H
 
 #include "qgsmapcanvasitem.h"
+#include "qgscoordinatereferencesystem.h"
 
 class QgsGCPCanvasItem;
 
@@ -28,7 +29,7 @@ class QgsGeorefDataPoint : public QObject
     //! constructor
     QgsGeorefDataPoint( QgsMapCanvas *srcCanvas, QgsMapCanvas *dstCanvas,
                         const QgsPointXY &pixelCoords, const QgsPointXY &mapCoords,
-                        bool enable );
+                        const QgsCoordinateReferenceSystem proj, bool enable );
     QgsGeorefDataPoint( const QgsGeorefDataPoint &p );
     ~QgsGeorefDataPoint() override;
 
@@ -38,6 +39,9 @@ class QgsGeorefDataPoint : public QObject
 
     QgsPointXY mapCoords() const { return mMapCoords; }
     void setMapCoords( const QgsPointXY &p );
+
+    QgsPointXY transCoords() const;
+    void setTransCoords( const QgsPointXY &p );
 
     bool isEnabled() const { return mEnabled; }
     void setEnabled( bool enabled );
@@ -53,6 +57,8 @@ class QgsGeorefDataPoint : public QObject
     QPointF residual() const { return mResidual; }
     void setResidual( QPointF r );
 
+    QgsCoordinateReferenceSystem crs() const { return mCrs; }
+
   public slots:
     void moveTo( QPoint, bool isMapPlugin );
     void updateCoords();
@@ -64,8 +70,10 @@ class QgsGeorefDataPoint : public QObject
     QgsGCPCanvasItem *mGCPDestinationItem = nullptr;
     QgsPointXY mPixelCoords;
     QgsPointXY mMapCoords;
+    QgsPointXY mTransCoords;
 
     int mId;
+    QgsCoordinateReferenceSystem mCrs;
     bool mEnabled;
     QPointF mResidual;
 
--- a/src/app/georeferencer/qgsgeorefmainwindow.cpp
+++ b/src/app/georeferencer/qgsgeorefmainwindow.cpp
@@ -540,11 +540,11 @@ void QgsGeoreferencerMainWindow::linkGeorefToQgis( bool link )
 }
 
 // GCPs slots
-void QgsGeoreferencerMainWindow::addPoint( const QgsPointXY &pixelCoords, const QgsPointXY &mapCoords,
+void QgsGeoreferencerMainWindow::addPoint( const QgsPointXY &pixelCoords, const QgsPointXY &mapCoords, const QgsCoordinateReferenceSystem &crs,
     bool enable, bool finalize )
 {
   QgsGeorefDataPoint *pnt = new QgsGeorefDataPoint( mCanvas, QgisApp::instance()->mapCanvas(),
-      pixelCoords, mapCoords, enable );
+      pixelCoords, mapCoords, QgsCoordinateReferenceSystem( crs ), enable );
   mPoints.append( pnt );
   mGCPsDirty = true;
   if ( finalize )
@@ -612,7 +612,6 @@ void QgsGeoreferencerMainWindow::movePoint( QPoint p )
   if ( mvPoint )
   {
     mvPoint->moveTo( p, isMapPlugin );
-    mGCPListWidget->updateGCPList();
   }
 
 }
@@ -620,6 +619,7 @@ void QgsGeoreferencerMainWindow::movePoint( QPoint p )
 void QgsGeoreferencerMainWindow::releasePoint( QPoint p )
 {
   Q_UNUSED( p )
+  mGCPListWidget->updateGCPList();
   // Get Map Sender
   if ( sender() == mToolMovePoint )
   {
@@ -637,7 +637,7 @@ void QgsGeoreferencerMainWindow::showCoordDialog( const QgsPointXY &pixelCoords 
   {
     mMapCoordsDialog = new QgsMapCoordsDialog( QgisApp::instance()->mapCanvas(), pixelCoords, this );
     connect( mMapCoordsDialog, &QgsMapCoordsDialog::pointAdded, this,
-    [this]( const QgsPointXY & a, const QgsPointXY & b ) { this->addPoint( a, b ); }
+    [this]( const QgsPointXY & a, const QgsPointXY & b, const QgsCoordinateReferenceSystem & crs ) { this->addPoint( a, b, crs ); }
            );
     mMapCoordsDialog->show();
   }
@@ -1291,13 +1291,14 @@ bool QgsGeoreferencerMainWindow::loadGCPs( /*bool verbose*/ )
 
     QgsPointXY mapCoords( ls.at( 0 ).toDouble(), ls.at( 1 ).toDouble() ); // map x,y
     QgsPointXY pixelCoords( ls.at( 2 ).toDouble(), ls.at( 3 ).toDouble() ); // pixel x,y
-    if ( ls.count() == 5 || ls.count() == 8 )
+    QgsCoordinateReferenceSystem proj( ls.at( 8 ) );
+    if ( ls.count() == 5 || ls.count() == 9 )
     {
       bool enable = ls.at( 4 ).toInt();
-      addPoint( pixelCoords, mapCoords, enable, false );
+      addPoint( pixelCoords, mapCoords, proj, enable, false );
     }
     else
-      addPoint( pixelCoords, mapCoords, true, false );
+      addPoint( pixelCoords, mapCoords, proj, true, false );
 
     ++i;
   }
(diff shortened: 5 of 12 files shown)


roya0045:
UI is fine, still need to see if move gcp is problematic and can be improved.

roya0045:
Ok the move tool works ok in the georef window and refreshes, but not in the main canvas, the points doesn't move until the canvas is refreshed. I don't think my changes should have changed anything, I'll look into it.

roya0045:
The constant reprojecting points when moving due to https://github.com/qgis/QGIS/blob/fe2faaacf17b0f6cf2aaac1ef4e10cf3d5ef45f8/src/app/georeferencer/qgsgeorefmainwindow.cpp#L567 seems to be the UI issue. I think that I should move the update on the release rather than move.

roya0045:
I have no clue why but the CRS selctor member stays as a null pointer in some places. I'm not sure if it goes out of scope or is badly initiated. But the values are passed on. QT confuses me.

roya0045:
something is really wonky with QgsProjectionSelectionWidget, it always return an empty crs and the connections/init/signals don't work. The one in the transformation dialog seems somewhat ok. The one I am trying to make work just doesn't cooperate for unknown reasons.

roya0045:
Simple use case worksa as intended, will try with varying source projction next.

src/app/georeferencer/qgsgcplist.h
```diff
@@ -29,7 +29,7 @@ class QgsGCPList : public QList<QgsGeorefDataPoint *>
     QgsGCPList() = default;
     QgsGCPList( const QgsGCPList &list );
 
-    void createGCPVectors( QVector<QgsPointXY> &mapCoords, QVector<QgsPointXY> &pixelCoords );
+    void createGCPVectors( QVector<QgsPointXY> &mapCoords, QVector<QgsPointXY> &pixelCoords, const QString &wktProj = QString() );
```
nirvn, line 32: What your passing on is the QgsCoordinateReferenceSystem's authid(), calling this variable wktProj is misleading.

src/app/georeferencer/qgsgcplist.cpp
```diff
@@ -30,16 +33,27 @@ QgsGCPList::QgsGCPList( const QgsGCPList &list )
   }
 }
 
-void QgsGCPList::createGCPVectors( QVector<QgsPointXY> &mapCoords, QVector<QgsPointXY> &pixelCoords )
+void QgsGCPList::createGCPVectors( QVector<QgsPointXY> &mapCoords, QVector<QgsPointXY> &pixelCoords, const QString &wktProj )
 {
+  QgsCoordinateReferenceSystem projTarget( wktProj );
+  QgsCoordinateTransform transformer;
```
nirvn, line 39: Not used.

src/app/georeferencer/qgsgcplist.cpp
```diff
@@ -30,16 +33,27 @@ QgsGCPList::QgsGCPList( const QgsGCPList &list )
   }
 }
 
-void QgsGCPList::createGCPVectors( QVector<QgsPointXY> &mapCoords, QVector<QgsPointXY> &pixelCoords )
+void QgsGCPList::createGCPVectors( QVector<QgsPointXY> &mapCoords, QVector<QgsPointXY> &pixelCoords, const QString &wktProj )
 {
+  QgsCoordinateReferenceSystem projTarget( wktProj );
```
nirvn, line 38: targetCrs might be a better name?

src/app/georeferencer/qgsgcplist.cpp
```diff
@@ -30,16 +33,27 @@ QgsGCPList::QgsGCPList( const QgsGCPList &list )
...
   for ( int i = 0, j = 0; i < sizeAll(); i++ )
   {
     QgsGeorefDataPoint *pt = at( i );
     if ( pt->isEnabled() )
     {
-      mapCoords[j] = pt->mapCoords();
+      if ( projTarget.isValid() )
+      {
+        transCoords =  QgsCoordinateTransform( pt->crs(), projTarget,
+                                               QgsProject::instance() ).transform( pt->mapCoords() );
```
nirvn, line 51: You'll have to wrap the transform call within a try{} and catch exceptions (e.g. https://github.com/qgis/QGIS/blob/master/src/app/locator/qgsinbuiltlocatorfilters.cpp#L1007 )

src/app/georeferencer/qgsgeorefdatapoint.cpp
```diff
@@ -16,17 +16,21 @@
...
                                         const QgsPointXY &pixelCoords, const QgsPointXY &mapCoords,
-                                        bool enable )
+                                        const QString &wktProj, bool enable )
   : mSrcCanvas( srcCanvas )
   , mDstCanvas( dstCanvas )
   , mPixelCoords( pixelCoords )
   , mMapCoords( mapCoords )
+  , mTransCoords( mapCoords )
   , mId( -1 )
+  , mProj( wktProj )
```
nirvn, line 33: That just seems to be a recipe for disaster. I'd either create a QgsCoordinateReferenceSystem in the constructor, or even better, instead of passing on a string reference as function parameter, pass on a QgsCoordinateReferenceSystem.

Also, rename the variable to mCrs to ease readability.

src/app/georeferencer/qgsgeorefdatapoint.cpp
```diff
@@ -77,6 +82,26 @@ void QgsGeorefDataPoint::setMapCoords( const QgsPointXY &p )
...
+  }
+  if ( mGCPDestinationItem )
+  {
+    mGCPDestinationItem->update();
+  }
+}
+
+QgsPointXY QgsGeorefDataPoint::transCoords() const
+{
+  if ( mTransCoords!=mMapCoords )
```
nirvn, line 100: Try to set up your pre commit git hook, it would have revealed a formatting issue here. 

Code-wise, why the if here? Simply return mTransCoords.

src/app/georeferencer/qgsgeorefdatapoint.h
```diff
@@ -28,7 +29,7 @@ class QgsGeorefDataPoint : public QObject
     //! constructor
     QgsGeorefDataPoint( QgsMapCanvas *srcCanvas, QgsMapCanvas *dstCanvas,
                         const QgsPointXY &pixelCoords, const QgsPointXY &mapCoords,
-                        bool enable );
+                        const QString &wktProj, bool enable );
```
nirvn, line 32: Same as above, avoid passing on a string, you can pass on a QgsCoordinateReferenceSystem object instead.

src/app/georeferencer/qgsgeorefmainwindow.cpp
```diff
@@ -238,6 +239,8 @@ void QgsGeoreferencerMainWindow::openRaster()
   QString otherFiles = tr( "All other files (*)" );
   QString lastUsedFilter = s.value( QStringLiteral( "/Plugin-GeoReferencer/lastusedfilter" ), otherFiles ).toString();
 
+  qDebug()<<lastUsedFilter;
```
nirvn, line 242: qDebug()s need to go.

src/ui/georeferencer/qgsmapcoordsdialogbase.ui
```diff
@@ -16,67 +16,96 @@
...
+       <widget class="QLabel" name="textLabel2">
+        <property name="text">
+         <string>Y / North</string>
+        </property>
+       </widget>
+       <widget class="QLineEdit" name="leYCoord"/>
+      </widget>
+     </item>
+     <item>
+      <widget class="QgsProjectionSelectionWidget" name="mProjSelect" native="true">
```
nirvn, line 58: Why having a projection selection widget here? We should deal with only two CRSes in the georeferencer context: the CRS set for the raster being georeferenced, and the main window's map canvas.

nirvn (changes requested):
@roya0045 , I had a quick look at it. I haven't compiled / tried it, but you addressing the comments here and cleaning up the PR (i.e. squashing commits into one) will increase the chance of getting a green light.

src/ui/georeferencer/qgsmapcoordsdialogbase.ui
roya0045: I'd have to disagree, if you have a map with coordinates drawn on it and you need to type them manually, having the option to set the crs of the coordinate source is critical. This is why I have added this widget, though it can be wonky at time.

src/app/georeferencer/qgsgeorefdatapoint.h
roya0045: I wasn't sure if it was better to store a string or the crs object, as this meant more imports but no conversion needed.

src/app/georeferencer/qgsgeorefdatapoint.cpp
roya0045: TransCoords is not always set, I wasn't sure how to check an empty/unset qgspointxy as I wrote this.

src/ui/georeferencer/qgsmapcoordsdialogbase.ui
nirvn: I'd split your effort into two then. First, come up with a clean PR to do map canvas to georeferencing raster's projection transformation, that's a gap that's undeniably worth fixing.

In a second PR, work on polishing the code and UI to propose this additional feature.

src/app/georeferencer/qgsgcplist.h
roya0045: should I pass a CRS directly instead here also? I was on the fence, initially it was a wkt that was passed but changed to authid to debug the faulty crs widget.

nirvn: Yes, CRS object here would be advised here too.

src/app/georeferencer/qgsgcplist.cpp
roya0045: Unsure if its a good idea to keep the transform when projections haven't changed or if better to make a new one each time.

src/app/georeferencer/qgsgeorefdatapoint.cpp
nirvn: return mTransCoords.isEmpty() ? mMapCoords : mTransCoords ?

src/ui/georeferencer/qgsmapcoordsdialogbase.ui
roya0045: Welp, it works fine now, I just had to not declare the projection selector in the header. All now works as intended, projections are properly assigned and coords are converted. I'll be making a gif to display.

src/app/georeferencer/qgsgeorefmainwindow.cpp
```diff
@@ -482,11 +482,11 @@ void QgsGeoreferencerMainWindow::linkGeorefToQgis( bool link )
 }
 
 // GCPs slots
-void QgsGeoreferencerMainWindow::addPoint( const QgsPointXY &pixelCoords, const QgsPointXY &mapCoords,
+void QgsGeoreferencerMainWindow::addPoint( const QgsPointXY &pixelCoords, const QgsPointXY &mapCoords, const QString &proj,
```
nyalldawson, line 485:
```suggestion
void QgsGeoreferencerMainWindow::addPoint( const QgsPointXY &pixelCoords, const QgsPointXY &mapCoords, const QgsCoordinateReferenceSystem &crs,
```

src/app/georeferencer/qgsgeorefmainwindow.cpp
```diff
@@ -579,7 +579,7 @@ void QgsGeoreferencerMainWindow::showCoordDialog( const QgsPointXY &pixelCoords
   {
     mMapCoordsDialog = new QgsMapCoordsDialog( QgisApp::instance()->mapCanvas(), pixelCoords, this );
     connect( mMapCoordsDialog, &QgsMapCoordsDialog::pointAdded, this,
-    [this]( const QgsPointXY & a, const QgsPointXY & b ) { this->addPoint( a, b ); }
+    [this]( const QgsPointXY & a, const QgsPointXY & b, const QString & crs ) { this->addPoint( a, b, crs ); }
```
nyalldawson, line 582:
```suggestion
    [this]( const QgsPointXY & a, const QgsPointXY & b, const QgsCoordinateReferenceSystem & crs ) { this->addPoint( a, b, crs ); }
```

src/app/georeferencer/qgsgeorefmainwindow.cpp
```diff
@@ -1245,15 +1246,16 @@ void QgsGeoreferencerMainWindow::saveGCPs()
...
              .arg( qgsDoubleToString( pt->mapCoords().x() ),
                    qgsDoubleToString( pt->mapCoords().y() ),
                    qgsDoubleToString( pt->pixelCoords().x() ),
                    qgsDoubleToString( pt->pixelCoords().y() ) )
              .arg( pt->isEnabled() )
              .arg( qgsDoubleToString( pt->residual().x() ),
                    qgsDoubleToString( pt->residual().y() ),
-                   qgsDoubleToString( std::sqrt( pt->residual().x() * pt->residual().x() + pt->residual().y() * pt->residual().y() ) ) )
+                   qgsDoubleToString( std::sqrt( pt->residual().x() * pt->residual().x() + pt->residual().y() * pt->residual().y() ) ),
+                   pt->crs().toWkt() )
```
nyalldawson, line 1258:
```suggestion
                   pt->crs().toWkt( QgsCoordinateReferenceSystem::WKT_PREFERRED ) )
```

src/app/georeferencer/qgsgeorefmainwindow.h
```diff
@@ -85,7 +85,7 @@ class QgsGeoreferencerMainWindow : public QMainWindow, private Ui::QgsGeorefPlug
 
     // gcps
     void addPoint( const QgsPointXY &pixelCoords, const QgsPointXY &mapCoords,
-                   bool enable = true, bool finalize = true );
+                   const QString &proj, bool enable = true, bool finalize = true );
```
nyalldawson, line 88:
```suggestion
                   const QgsCoordinateReferenceSystem &crs, bool enable = true, bool finalize = true );
```

src/app/georeferencer/qgsmapcoordsdialog.cpp
```diff
@@ -92,7 +96,14 @@ void QgsMapCoordsDialog::buttonBox_accepted()
...
 
-  emit pointAdded( mPixelCoords, QgsPointXY( x, y ) );
+  if ( !mCrs.isValid() )
+  {
+    if ( mProjSelect->crs().isValid() )
+      mCrs = mProjSelect->crs();
+    else
+      mCrs = mQgisCanvas->mapSettings().destinationCrs();
+  }
+  emit pointAdded( mPixelCoords, QgsPointXY( x, y ), mCrs.toWkt() );
```
nyalldawson, line 106:
```suggestion
  emit pointAdded( mPixelCoords, QgsPointXY( x, y ), mCrs );
```

src/app/georeferencer/qgsmapcoordsdialog.cpp
```diff
@@ -161,6 +175,12 @@ double QgsMapCoordsDialog::dmsToDD( const QString &dms )
     return res;
 }
 
+void QgsMapCoordsDialog::updateWkt( const QgsCoordinateReferenceSystem &crs )
```
nyalldawson, line 178: wkt?

src/app/georeferencer/qgsmapcoordsdialog.h
```diff
@@ -68,19 +70,24 @@ class QgsMapCoordsDialog : public QDialog, private Ui::QgsMapCoordsDialogBase
     void maybeSetXY( const QgsPointXY &, Qt::MouseButton );
     void updateOK();
     void setPrevTool();
+    void updateWkt( const QgsCoordinateReferenceSystem &crs );
 
   signals:
-    void pointAdded( const QgsPointXY &, const QgsPointXY & );
+    void pointAdded( const QgsPointXY &, const QgsPointXY &, const QString & );
```
nyalldawson, line 76:
```suggestion
    void pointAdded( const QgsPointXY &, const QgsPointXY &, const QgsCoordinateReferenceSystem & );
```

roya0045:
georef_upgrade


raster: cartogram by Max Roser for Our World in Data. Basemap: Bing satellite.

roya0045:
@nyalldawson Thanks Nyall, I tried that initially and got annoyed with the vague debug statement when trying to figure out what was wrong. I'll retry it in a few moments.

src/app/georeferencer/qgsgeorefdatapoint.h
roya0045: done

src/app/georeferencer/qgsgcplist.h
roya0045: done

src/app/georeferencer/qgsgcplist.cpp
roya0045: yep

src/app/georeferencer/qgsgeorefdatapoint.cpp
roya0045: changed

src/app/georeferencer/qgsgeorefmainwindow.cpp
roya0045: Thanks Nyall

src/app/georeferencer/qgsmapcoordsdialog.cpp
roya0045: remnant of string approach.

src/ui/georeferencer/qgsmapcoordsdialogbase.ui
roya0045: @nirvn I'd like to keep this in if possible, it works far better this way, see the gif below to see it in action.

roya0045:
@nirvn With regards to the gif, do you still suggest doing this in two  part or keeping the widget, as it is pretty important and fully functionnal.

nirvn (approved):

nirvn:
@roya0045 , yeah, I've come around, the widget is fine there.

nirvn:
@roya0045 , can you rebase the PR to fix the conflicts, I'll merge afterward. Thanks for pushing this through the finish line, it's a useful feature addition.

src/app/georeferencer/qgsgeorefdatapoint.cpp
```diff
@@ -147,6 +170,10 @@ void QgsGeorefDataPoint::moveTo( QPoint p, bool isMapPlugin )
   {
     QgsPointXY pnt = mGCPDestinationItem->toMapCoordinates( p );
     mMapCoords = pnt;
+    if ( mSrcCanvas && !mSrcCanvas->mapSettings().destinationCrs().toWkt().isEmpty() )
```
nyalldawson, line 173:
```suggestion
    if ( mSrcCanvas && mSrcCanvas->mapSettings().destinationCrs().isValid() )
```

nyalldawson:
(That's an approval from me after the above change is committed!)

roya0045:
@nirvn @nyalldawson I have rebased, squashed and made the change, all looks uncorrupted.

Thank you for taking the time to review this PR!

roya0045:
@zacharlie could you add the changelog tag to this PR, thanks.

roya0045:
@nirvn Thanks a ton Mathieu!  If you have some free time I also have other PR that could benefit from your feedback.

uclaros:
Admittedly a bit late, but I have a few objections regarding this.
1. QGIS crashes when trying to load a file with an old .points in the georeferencer. (list index out of range in `src/app/georeferencer/qgsgeorefmainwindow.cpp:1294`
2. Storing the whole crs wkt in each row in the .points file makes the file big and not as human readable. Also riskier for corruption when hand edited.
3. The crs selector in the coordinate picker dialog should default to the map canvas' crs and not to *invalid projection* for users who want to just type in coordinates.
4. The x and y lineedits in the coordinate picker dialog shrink when dialog's y size is reduced to minimum.
5. The user cannot know what crs each of the points in the gcp table is in, unless he opens the .points file and tries to read the wkt!
6. When typing in coordinates and selecting a different crs, the point is displayed in the map canvas at the typed coordinates in the map crs, not at the transformed value.

Do we really need per gcp crs definition? I think this has complicated things for little gain...

roya0045:
@uclaros 
1. Is it common to load back old points? I can try to look into a fix
2. Why would you want to hand edit the file? Should it even be human readable?
3. What is a 'just type coordinates' are people just wanting to type random coordinates without any crs? The previous behaviour is to estimate that all points are in the output crs by default. I see no wrong in this.
4. Is this a critical point? Even after the changes?
5. I thought about displaying it but you can move the point in the canvas if you want to correct them otherwise.
6. Are they at the right spot in the canvas? It was fine in my test but I can look into it.
7. Would need to look into it as I don't see why people would want to change the file directly rather than importing it.

The per gcp crs is useful when you need to use gcp in multiple coordinates from sources with multiple gcp, as opposed to having to convert all the coordinates to a single projection before inputing them.

uclaros:
1. Yes, it is quite common and that's the reason the .points file is saved. Re-loading the .points allows to fine-tune a georeference, change the transformation type or even just to better visualize the residuals. This is an easy fix though.
2. The only advantage of having a flat text file for .points is that it is human readable and easily modifiable. Transposing a grid is probably the most common reason I hand-modify .points files. 
3. The most common case is project crs = output crs. User sets crs on project, loads raster on georef, starts adding points before opening the *transformation settings*, now the *enter map coordinates* dialog states *invalid projection*.
4. This is just a minor cosmetic issue.

I would suggest a slightly different approach that allows less room for misunderstanding by users:

- On raster load, output crs defaults to project crs
- When adding gcps a crs selector is offered. Default value is project crs, but if changed the value is also stored for the next gcp.
- If the user selects a different crs for a gcp, the coords are transformed to project crs and stored as such in the editor and in the .points file.
- Choosing a different output crs actually re-projects the output file to the desired output crs.

The only disadvantage I see is that since coords will be shown and stored in project crs then users will not see the number they typed but its projected equivalent. However, with the current solution things are more confusing, since on the same table coords in different crs's are shown and the user cannot distinguish what point is in what crs

uclaros:
8. Generate GDAL script is broken when used with per gcp crs.

roya0045:
Do you have any exemple of what 'broken' means? Is the destination crs set? I had a quick look and couldn't find what else may be problematic.

uclaros:
> Do you have any exemple of what 'broken' means? Is the destination crs set? I had a quick look and couldn't find what else may be problematic.

- load a raster, set epsg:2100 crs both project and output
- add 2 gcps in 2100 eg.1,2 and 2,1
- add 1 gcp in 4326, eg 3,3
- the gcp table correctly displays the reprojected coordinates for the last point
- the map canvas displays the 3 points at 2100 coords (1,2), (2,1), **(3,3)** (two correct, one wrong, while the label displays the reprojected value)
- click *Generate gdal script* : all seems fine
- save the .points and open in a text editor: all points are at the correct coords and crs, ok
- load the .points on georeferencer
- the gcp table incorrectly displays **(3,3)** for the last point instead of the reprojected value
- click *Generate gdal script* : the last point is **(3,3)**


Another confusing bit is that the user may click *From map canvas* and pick a point, then he can change the crs and really mess things up! Changing the crs should automatically replace the lineedit values with the reprojected ones.

What do you think about my alternative approach proposed above?

roya0045:
Ah so the script is problematic after loading the table.

uclaros:
> I would suggest a slightly different approach that allows less room for misunderstanding by users:
> 
>     * On raster load, output crs defaults to project crs
> 
>     * When adding gcps a crs selector is offered. Default value is project crs, but if changed the value is also stored for the next gcp.
> 
>     * If the user selects a different crs for a gcp, the coords are transformed to project crs and stored as such in the editor and in the .points file.
> 
>     * Choosing a different output crs actually re-projects the output file to the desired output crs.
> 
> 
> The only disadvantage I see is that since coords will be shown and stored in project crs then users will not see the number they typed but its projected equivalent. However, with the current solution things are more confusing, since on the same table coords in different crs's are shown and the user cannot distinguish what point is in what crs

@nyalldawson @nirvn @roya0045  could you evaluate my proposed alternative approach and completely avoid the aforementioned issues?

roya0045:
@uclaros 
1) this is done in the fix PR
2) I want to look into this as I get the chance
3) The files are not linked, the user has to save again
4) This is done but you have to reproject/georeference again

uclaros:
The bullets describe the proposed workflow logic. The main difference I propose is that while points may be entered on differenct crss *all points are handled and stored reprojected on a single crs*

roya0045:
Ok but if you use 2 or 3 crs to establish your reference and you want to tweak one point by changing its values, you would need to recall the crs of the original coordinates otherwise you'll display reprojected coordinates and this may be event more confusing if the provided coordinates by the user are not what they inputted and can't fix them other than delete and recreate a point.

roya0045:
BTW I think I solved the crs recall thing, though I don't have the time to test soon.

uclaros:
> Ok but if you use 2 or 3 crs to establish your reference and you want to tweak one point by changing its values, you would need to recall the crs of the original coordinates otherwise you'll display reprojected coordinates and this may be event more confusing if the provided coordinates by the user are not what they inputted and can't fix them other than delete and recreate a point.

This should only be a problem when directly modifying coords in the gcp table. In that situation however it is way more confusing the fact that the user *does not know what crs was chosen for each point*!

Modifying points from a different crs could be aided by a *pick new point for this gcp* where the pick coords dialog would be redisplayed and a crs could be re-chosen. This would also benefit other cases where moving gcps is not practical like using the same grid for different rasters.

roya0045:
crs recall is fixed, display issue is inconveinient, solution is nearly done.

uclaros:
> crs recall is fixed, display issue is inconveinient, solution is nearly done.

The case of different crs per gcp is still handled incorrectly.

PR #39596, opened 2020-10-25
The order_expression parameter in Points to Path algorithm is not optional

DelazJ:


nyalldawson:
It needs to be optional, or older scripts will break

DelazJ:
but if I run it in the GUI without filling a value (I assume this is what is meant by optional), it asks me a value
https://github.com/qgis/QGIS/blob/788f71fc424765c27ee63e82a246e8bcb70f6a28/python/plugins/processing/algs/qgis/PointsToPaths.py#L125-L126

roya0045:
Can we use rowid/fid as the default? Or just the first field available, if its not truly optional.

nyalldawson:
Hmmmm, I think we need to remove that exception and just allow it to run without any order set

DelazJ:
> Hmmmm, I think we need to remove that exception and just allow it to run without any order set

Would this not lead to some unpredictable results? Something that would be useless and people will anyway have to give an order if they want to control the output?

Is it not possible to have it mandatory and, given that the issue resides with old scripts, have a check that there's a order_field value provided and transfer that value to order_expression? Afaict, order_field was also mandatory so the scripts should have a field assigned. Not sure if I'm clear enough but the idea is to have some logic to assign order_field value to order_expression if the script runs an old call.

nyalldawson:
No, that's not possible in current API. There's logic outside of the algorithm itself which tests that all mandatory parameters are filled. so making this parameter mandatory will break execution eg with the qgis_process tool. 

@m-kuhn any ideas here?

uclaros:
How about deprecating the algorithm and creating a tidier c++ one with a proper *Points to lines* name?

nyalldawson:
@uclaros win! Sounds like an ideal solution

m-kuhn:
> but if I run it in the GUI without filling a value (I assume this is what is meant by optional), it asks me a value
> 
> https://github.com/qgis/QGIS/blob/788f71fc424765c27ee63e82a246e8bcb70f6a28/python/plugins/processing/algs/qgis/PointsToPaths.py#L125-L126

It's not intended to be run from the gui without this param.

But allows legacy mode
https://github.com/qgis/QGIS/blob/788f71fc424765c27ee63e82a246e8bcb70f6a28/python/plugins/processing/algs/qgis/PointsToPaths.py#L122-L126

DelazJ:
> It's not intended to be run from the gui without this param.

But for an end user seeing optional, it is tempting... and then you realize it's a wrong optional
> But allows for legacy mode

Sorry. I miss these lines.
> How about deprecating the algorithm and creating a tidier c++ one with a proper Points to lines name?

Beyond my skills ;( so if someone wants to take the lead on this, feel free.
Thanks.

PR #39601, opened 2020-10-25
update unfinished Polish translation

hexmind:
## Description

Unfinished Polish translation has been updated.


Gustry:
I think you need to update strings only in https://www.transifex.com/qgis/
Because these TS files are pulled from Transifex.

hexmind:
Good to know, thanks.

PR #39602, opened 2020-10-25
Fix mssql connections not added/removed predictably when not from ma…

domi4484:
Fix mssql connections not added/removed predictably when not from main thread.

## Description


nyalldawson (approved):

nyalldawson:
For reference -- this brings the sql provider in line with the similar handling done for the auth db in QgsAuthManager::authDatabaseConnection(). I'm confident it's the right fix given that the auth db connection handling has been in place for many releases now and is used by EVERY qgis user.

PR #39632, opened 2020-10-26
[3d] Fixes for point cloud 3D renderer

wonder-sk:
This makes 3D renderer to behave correctly:
1. add data of child nodes to the existing parent nodes's data rather than replacing it
2. fix bounding box calculation that was causing incorrect culling of nodes

qgis-3d-point-cloud-trencin

### DIFF
--- a/src/3d/qgsaabb.cpp
+++ b/src/3d/qgsaabb.cpp
@@ -92,3 +92,8 @@ QList<QVector3D> QgsAABB::verticesForLines() const
   }
   return vertices;
 }
+
+QString QgsAABB::toString() const
+{
+  return QString( "X %1 - %2  Y %3 - %4  Z %5 - %6" ).arg( xMin ).arg( xMax ).arg( yMin ).arg( yMax ).arg( zMin ).arg( zMax );
+}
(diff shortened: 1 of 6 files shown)


nyalldawson:
@wonder-sk would the bounding box calculations fixed as part of this/the earlier pr also affect non point cloud sources? I've often seen issues where boxes close to the camera disappear too early, causing terrain segments to disappear

src/3d/qgsaabb.cpp
```diff
@@ -92,3 +92,8 @@ QList<QVector3D> QgsAABB::verticesForLines() const
   }
   return vertices;
 }
+
+QString QgsAABB::toString() const
+{
+  return QString( "X %1 - %2  Y %3 - %4  Z %5 - %6" ).arg( xMin ).arg( xMax ).arg( yMin ).arg( yMax ).arg( zMin ).arg( zMax );
```
nyalldawson, line 98:
```suggestion
  return QStringLiteral( "X %1 - %2  Y %3 - %4  Z %5 - %6" ).arg( xMin ).arg( xMax ).arg( yMin ).arg( yMax ).arg( zMin ).arg( zMax );
```

wonder-sk:
> would the bounding box calculations fixed as part of this/the earlier pr also affect non point cloud sources? I've often seen issues where boxes close to the camera disappear too early, causing terrain segments to disappear

This fix is really specific to the point clouds.

I have seen issues with data disappearing too early in other occasions as well - that will need case by case investigation... if we have a case that is easy to replicate, I can have a look...

nyalldawson (approved):

PR #39634, opened 2020-10-26
fix map setting temporal be active when updating time controller

vcloarec:
When a mesh layer is loaded and it is the first layer, temporal controller is updated accordingly  with the mesh temporal data even temporal navigation is off. This update sends signal that finally sets the temporal map settings as active while the temporal navigation is off.
This PR avoids this behavior.


nyalldawson:
can you fix the commit message ? it seems truncated

Issue #38863, opened 2020-09-18
Compress/deflate raster algorithm


roya0045:
**Feature description.**
Outputs of some tools are files without any optimisation. It can be quite cumbersome to deal with 800mb files with a single band.

It would be nice to have a way to compress rasters directly in a **model**, either in place or not in order to not end up with a considerable amount of overly sized rasters.

uclaros:
You can use gdal translate in the model with custom creation options but it's not super user friendly

baswein:
I end up running GDAL Warp as my last step for most of my raster products in the model with the advanced>creation options. But it is a manual process and I often have to look up the options and being bad a spelling end up with a typo. It would be nicer if the model algos could use the preset raster creation option profiles.

alexbruy:
Fixed in #57902.

Please note that some algorithms still don't have support for creation options as they use different API/code path to create output.

roya0045:
> Fixed in #57902.
> 
> Please note that some algorithms still don't have support for creation options as they use different API/code path to create output.

Thanks Alex, this is a great improvement!

PR #39645, opened 2020-10-27
[Processing][GDal] Compressed Gtiff raster tool

roya0045:
## Description

This is a tool proposal to make compressed gtiff with rasters.

As the available creation options are limited in the existing dialogs and since models produce uncompressed rasters this should help optimising things.

This help expose some more compression variables and methods. I have restrained the output to a tiff as they have the most creation/compression options.

i am aware of the extra parameter that are available but I feel like an official tool help expose settings.

I still need to make the tests and find out the unique name() and instance() python errors.

I would like to hear if anyone think this is a bad idea or uneeded before going too far with this.

Thanks


alexbruy:
Can you elaborate why creation options are limited? You can pass exactly the same options and their combinations as you use in your tool. It is even possible to create presets for various purposes.

As for models, it is up to user to specify desired creation options for outputs. Creating separate tool won't make model outputs compressed automatically, will it? User still will need to add algorithm. How this is different from specifying creation options? Not to mention than with additional algorithm this will be additional model step and more processing time and temporary files.

roya0045:
@alexbruy I agree, but I felt like adding an official tool would make things clearer to user that aren't too familiar with the Gdal creation options. Also not all raster algorithms provide a way to use gdal creation options.

As stated in other issues/PR having the options is one thing but most users won't search for gdal creation option and know what they are. By having a dedicated tool would make things easier for most users.

baswein:
@roya0045 As I mentioned in https://github.com/qgis/QGIS/issues/38863 I am excited to have an easier to use compression tool in the model builder. Does it use the GDAL Create Option Profiles?  Could you post some screenshots?

roya0045:
@baswein Its still a work in progress, I'm getting some weird error on load but the UI work. If you want to try the easiest way is to get the modified files and replace the provider in C:\OSGeo4W64\apps\qgis\python\plugins\processing\algs\gdal and add the file to this folder and reload the provider. It should work but I didn't get the chance to test on a sample dataset yet to see the compression rate.

alexbruy:
> most users won't search for gdal creation option and know what they are

Wouldn't be better to improve core widget to make them more obvious? This way whole QGIS will benefit as widget is used not only in the Processing?

> not all raster algorithms provide a way to use gdal creation options

Wouldn't be better to improve existing algorithms instead?

roya0045:
@alexbruy Current widget has limited space to display information, I don't want to rework everything in it.

As for improving other alg this would mean creating a temporary file and then using the settings on it when gdal is not used. This is the same as a standalone tool.

Reworking everything would also take more time rather than having this as a distinct tool.

alexbruy:
> Current widget has limited space to display information

@roya0045 may I ask you to open feature request of bug report about it and describe what is missed in your opinion? Personally I think that updating widget makes much more sense.

roya0045:
@alexbruy But integrating the widget with other providers for each tool that outputs a raster seems excessive, don't you think?

This is not only about the UI but also usability.

uclaros:
Sorry @roya0045, I too believe that this effort could be better spent making *translate (convert format)* user friendlier. The tool is already a cut down UI version of gdal_translate made specifically for changing formats, so compressing a tiff is also what it is supposed to do, but the average user may not know that!
Few things that should be done imho:
1. **Add "compress" to the tools name or make it searchable with "compress" as a keyword.** The average user does not know what gdal_translate is, he just knows he wants to compress his raster.
2. **Make the creation options names and values selectable from lists.** The average user does not know what the creation options are for their format of choice. The help button scares the average user away and it only helps the advanced user to remember some -co he forgot and only for GTiff. Picking `PHOTOMETRIC` from a list of available creation options and then `YCBCR` from another would be so much easier.
3. **Name the default preset "default - uncompressed"** or similar, to reflect that if the user wants compression, this is a way to achieve it!
4. **Add more presets for common use cases.** We should have more presets baked in, like `-co COMPRESS=JPEG -co JPEG_QUALITY=100 -co PHOTOMETRIC=YCBCR -co INTERLEAVE=PIXEL -co TILED=YES`

roya0045:
@alexbruy @uclaros Very well, I'll try to see how the current widget can be improved. I'll close this in the meantime but I'll probably use it some of the code, mostly for quality settings.

nyalldawson:
@roya0045 don't forget that it's possible to extend processing through plugins! If you feel there's a strong need for this algorithm you could make your own plugin which adds it, and then you'll be free to make any decisions yourself so that the tool fits your needs exactly.

roya0045:
@nyalldawson I'd rather contribute to core as I feel it has more potential to help everyone.

PR #39651, opened 2020-10-27
[pointclouds] QgsPointCloudBlock

PeterPetrik:
Adds a data storage classes for raw data storage for point clouds.

### DIFF
--- a/src/core/pointcloud/qgspointcloudblock.h
+++ b/src/core/pointcloud/qgspointcloudblock.h
@@ -27,56 +27,7 @@
 
 #define SIP_NO_FILE
 
-/**
- * \ingroup core
- *
- * Attribute for point cloud data
- * pair of name and size in bytes
- */
-class CORE_EXPORT QgsPointCloudAttribute
-{
-  public:
-    //! Ctor
-    QgsPointCloudAttribute( const QString &name, int size );
-
-    //! Returns name of the attribute
-    QString name() const;
-
-    //! Returns size of the attribute in bytes
-    int size() const;
-
-  private:
-    QString mName;
-    int mSize;
-};
-
-/**
- * \ingroup core
- *
- * Collection of point cloud attributes
- */
-class CORE_EXPORT QgsPointCloudAttributeCollection
-{
-  public:
-    //! Ctor
-    QgsPointCloudAttributeCollection();
-    //! Ctor with given attributes
-    QgsPointCloudAttributeCollection( const QVector<QgsPointCloudAttribute> &attributes );
-    //! Adds extra attribute
-    void push_back( const QgsPointCloudAttribute &attribute );
-
-    //! Returns all attributes
-    QVector<QgsPointCloudAttribute> attributes() const;
-
-    //! Returns -1 if no such attribute exists
-    int offset( const QString &attributeName, const QgsPointCloudAttribute *foundAttribute ) const;
-
-    //! Returns total size of record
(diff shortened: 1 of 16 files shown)


src/core/providers/ept/qgseptdecoder.cpp
```diff
@@ -34,52 +36,39 @@
...
-    // qint32 y = *(double*)(ptData+8);
-    // qint32 z = *(double*)(ptData+16);
+    qint32 x = *( double * )( s );
+    qint32 y = *( double * )( s + 8 );
+    qint32 z = *( double * )( s + 16 );
     char cls = bytes[30];
-    // vertices.push_back( Point3D( x, y, z ) );
-    classes[i] = cls;
+
+    stream << x << y << z << cls;
```
wonder-sk, line 63: QDataStream has its own way of serializing data - it is not raw data stream

src/core/pointcloud/qgspointcloudindex.h
```diff
@@ -156,29 +157,20 @@ class CORE_EXPORT QgsPointCloudIndex: public QObject
...
+     * Returns node data block
+     *
+     * e.g. positions (needs to be scaled and offset applied to get coordinates) or
+     * classification, intensity or custom attributes
+     *
+     * It is caller responsibility to free the block.
+     *
+     * May return nullptr in case the node is not present or any other problem with loading
+     */
+    virtual QgsPointCloudBlock *nodeData( const IndexedPointCloudNode &n, const QgsPointCloudAttributeCollection &requestedAttributes ) = 0;
```
nyalldawson, line 173: I'm not a fan of passing the attribute collection directly here. I'd rather see this added to a "QgsPointCloudRequest"  (modeled off QgsFeatureRequest) class instead, and the request class used here. That would give us more flexibility to add extra filtering options in future.

src/core/pointcloud/qgspointcloudblock.h
```diff
@@ -0,0 +1,108 @@
...
+                        const QgsPointCloudAttributeCollection &attributes,
+                        const QByteArray &data );
+    //! Dtor
+    ~QgsPointCloudBlock();
+
+    //! Returns raw pointer to data
+    const char *data() const;
+
+    //! Returns number of points that are stored in the block
+    int pointCount() const;
```
nyalldawson, line 97: is int always going to be large enough for this? long long would be safer.

nyalldawson:
Request looks good -- thanks!

src/core/pointcloud/qgspointcloudblock.h
PeterPetrik: I think it should be ok with `int`, since the point cloud block would have only data from one data file at the moment, which is around ~10k points..

nyalldawson: I'm just thinking ahead for if/when we get providers for other formats which aren't nicely partitioned

PeterPetrik: I do not think it is required now, with hitting limit of int, you would explode your memory with the stored data in bytestream

nyalldawson: Ok, fair enough!

nyalldawson (approved):

nyalldawson:
Looking good!

(We better get some tests in place soon for this. Ping me if you'd like me to add these)

Issue #23988, opened 2017-01-13
Add functions to convert color format


qgib:
Author Name: **Harrissou Santanna** (@DelazJ)
Original Redmine Issue: [16073](https://issues.qgis.org/issues/16073)

Redmine category:expressions

---
It could be nice to have functions to convert color from one format to another ( rgb(a), hsl, hex...)
For example, it could help to display @symbol_color value in another format than hex.

suricactus:
Not sure if this feature is really needed, as the main way of working with colors is in hex. There is already the `color_part` function, that provides flexibility to extract whatever one wants from a given color.
```
color_part('#ff00ff', 'hue') || ', ' || color_part('#ff00ff', 'saturation') || ', ' || color_part('#ff00ff', 'value')
'300, 100, 100'
```

PR #39655, opened 2020-10-27
Add two new expression functions: rgb_to_hex, to_base 

hamiltoncj:
## Description

[FEATURE] This adds two new expression functions:

Function rgb_to_hex - Returns a hex string representation of a color based on its red, green, blue, and optional alpha components. The output is either '#RRGGBB' or '#RRGGBBAA' where RR, GG, BB, and AA are the red, green, blue and alpha components.

Function to_base - Convert an integer value to a string representation in a different base or number system. Binary is base 2, octal base 8, and hexadecimal base 16.

rgb_to_hex

to_base

These are both function that I have found missing and wanted them for my use. Since this is my first code commit I only worked on these two, but plan on other related functions.

### DIFF
--- a/resources/function_help/json/rgb_to_hex
+++ b/resources/function_help/json/rgb_to_hex
@@ -1,0 +1,12 @@
+{
+  "name": "rgb_to_hex",
+  "type": "function",
+  "groups": ["Color"],
+  "description": "Returns a hex string representation of a color based on its red, green, blue, and optional alpha components. The output is either '#RRGGBB' or '#RRGGBBAA' where RR, GG, BB, and AA are the red, green, blue and alpha components.",
+  "arguments": [ {"arg":"red","description":"red component as an integer value from 0 to 255"},
+                 {"arg":"green","description":"green component as an integer value from 0 to 255"},
+                 {"arg":"blue","description":"blue component as an integer value from 0 to 255"},
+                 {"arg":"alpha","optional":true,"description":"alpha component as an integer value from 0 to 255"}],
+  "examples": [ { "expression":"rgb_to_hex(255,127,0)", "returns":"'#FF7F00'"} ]
+}
+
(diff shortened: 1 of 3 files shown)


nyalldawson:
@hamiltoncj Code looks good -- nice work!

I'll leave the discussion about the formatting scripts to the mailing list thread in order to keep things centralized. Apart from that, the only other change required here is some new tests for the new functions.

The color ones would sit somewhere around here:
https://github.com/qgis/QGIS/blob/master/tests/src/core/testqgsexpression.cpp#L1490

each line looks like this:

    QTest::newRow( "ramp color" ) << "ramp_color('Spectral',0.3)" << false << QVariant( "254,190,116,255" );

where the "ramp color" string is a unique name identifying the particular test. It should have enough detail to identify exactly which tests fails, so a good name is something like "rgb_to_hex red greater than 255".
The second argument ("ramp_color('....')) is the expression to test. The next "false" is whether the expression is invalid and should raise an error (e.g. when an input is invalid), and finally the last is the expected result from the expression.

You add as many lines as necessary to give your functions a good work out. Include ones which test the function with bad inputs, e.g. string values and components <0 or > 255.

The to_base tests could go somewhere around here: https://github.com/qgis/QGIS/blob/master/tests/src/core/testqgsexpression.cpp#L838

nyalldawson:
BTW -- welcome to the wonderful world of core QGIS development! I realise it's a steep learning curve, but after you get past this I'm sure you'll find it's a pleasure to contribute and the possibilities your enhancements can take will really explode. It's great to have you onboard!

hamiltoncj:
So I am not sure about the process of correcting the styling given that I am currently doing this on windows and once I correct it how do I update the pull request code? Sorry for not knowing this, but this is completely new to me.

nyalldawson:
@hamiltoncj I've replied via the qgis Dev mailing list thread, can we keep the conversation centred there?

alexbruy:
There is related ticket #23988 with good example how to extract color components using existing functions.

nyalldawson:

PR #39656, opened 2020-10-27
Handle 'inbuilt:' vector uri

rduivenvoorde:
This is take 2 to make it possible to use the inbuilt-vector data (world map etc) to be used in PyQGIS etc:
```
iface.addVectorLayer('inbuilt:/data/world_map.gpkg|layername=countries', 'Countries', 'ogr')
```
First take was: #39175 There @nyalldawson proposed to do it not ogr-specific but on a higher level. This is my first attempt to do it as he described there: https://github.com/qgis/QGIS/pull/39175#pullrequestreview-501689631

Not sure if that was what he meant. I think more code has to get rewritten here (as for example now a inbuilt-uri WITHOUT layername-part will not load AND not popup the 'choose layer dialog', because the vectorLayerPath is const... 
Should I create a copy of it and use that in the whole function?
Or am I totally wrong anyway.

One argument why I like #39175 actually better is that (untill now) inbuild is only used with OGR provider (and it is/was a oneline change). But feel free to disagree/comment or stop me from trying :-)

Also note that my proposed plan B in #39175 
```
iface.addVectorLayer(QgsApplication.pkgDataPath() + '/resources/data/world_map.gpkg|layername=Countries', '', 'ogr')
```
actually does not work on Linux distro installs, because the gpkg in the resource is not readable by the average user apparently.

### DIFF
--- a/src/app/qgisapp.cpp
+++ b/src/app/qgisapp.cpp
@@ -12900,7 +12900,27 @@ QgsVectorLayer *QgisApp::addVectorLayerPrivate( const QString &vectorLayerPath, 
   QgsVectorLayer::LayerOptions options { QgsProject::instance()->transformContext() };
   // Default style is loaded later in this method
   options.loadDefaultStyle = false;
-  QgsVectorLayer *layer = new QgsVectorLayer( vectorLayerPath, baseName, providerKey, options );
+
+  QVariantMap uriElements = QgsProviderRegistry::instance()->decodeUri( providerKey, vectorLayerPath );
+  // eg: QMap(("layerId", QVariant(Invalid))("layerName", QVariant(QString, "countries"))("path", QVariant(QString, "inbuilt:/data/world_map.gpkg")))
+  QgsVectorLayer *layer;
+  if ( uriElements.contains( "path" ) && uriElements.value( "path" ).toString().startsWith( QLatin1String( "inbuilt:" ) ) )
+  {
+    QString resourcePath;
+    QgsPathResolver resolver = QgsPathResolver( baseName );
+    if ( uriElements.contains( "layerName" ) )
+    {
+      resourcePath = resolver.readPath( uriElements.value( "path" ).toString() ) + "|layername=" + uriElements.value( "layerName" ).toString();
+    }
+    /*else{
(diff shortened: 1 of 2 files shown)


src/app/qgisapp.cpp
```diff
@@ -12900,7 +12900,27 @@ QgsVectorLayer *QgisApp::addVectorLayerPrivate( const QString &vectorLayerPath,
...
+    }
+    /*else{
+        resourcePath = resolver.readPath(path);  // mmm vectorLayerPath is const.... should remove the layerid= part??
+    }*/
+    layer = new QgsVectorLayer( resourcePath, baseName, providerKey, options );
+  }
+  else
+  {
+    layer = new QgsVectorLayer( vectorLayerPath, baseName, providerKey, options );
+  }
```
nyalldawson, line 12923:
```suggestion

  QVariantMap uriElements = QgsProviderRegistry::instance()->decodeUri( providerKey, vectorLayerPath );
  if ( uriElements.contains( QStringLiteral( "path"  ) ) )
  {
    // run layer path through QgsPathResolver so that all inbuilt paths and other localised paths are correctly expanded
    uriElements[ QStringLiteral( "path" ) ] = QgsPathResolver().readPath( uriElements.value( QStringLiteral( "path" ) ).toString() );
  }
  const QString updatedUri = QgsProviderRegistry::instance()->encodeUri( providerKey, uriElements );
  QgsVectorLayer *layer = new QgsVectorLayer( updatedUri, baseName, providerKey, options );
```

It's a bit cleaner this way -- we always run through the path resolver so that any plugin registered path modifiers will also get a chance to update the path.

rduivenvoorde:
Thanks @nyalldawson this is what it should be. Now I understand better your reasoning. And also removing the "|layername=countries" from the uri results in he 'choose layer' popup!
If tests pass, is this something to be backported into 3.16?

rduivenvoorde:
I'm looking into adding a python test (for the inbuilt: uri), but wondering where to put it...

nyalldawson:
> I'm looking into adding a python test (for the inbuilt: uri), but wondering where to put it...

There's no real infrastructure in place for unit test QgisApp methods.

PR #39664, opened 2020-10-28
Mesh export processes, first part : infrastructure and export vertices

vcloarec:
First part of porting the export mesh processes (actually in crayfish plugin) to QGIS.

Here the infrastructure with dataset group paramater and dataset time parameter to allow the user to easily choose a dataset.

And the first process that allow the user to export to a vector layer the mesh vertices with dataset value as attribute.

exportVertices


src/core/processing/qgsprocessingparametermeshdataset.h
```diff
@@ -0,0 +1,215 @@
...
+#define QGSPROCESSINGPARAMETERMESHDATASET_H
+
+#include "qgsprocessingparameters.h"
+#include "qgsprocessingparametertype.h"
+#include "qgsmeshdataset.h"
+
+/**
+ * A parameter for processing algorithms that need a list of mesh dataset groups
+ * A valid value for this parameter is a list (QVariantList) of dataset groups index in the mesh layer scope
+ *
```
nyalldawson, line 26: You should describe here how you expect algorithms to evaluate the parameter value. Ideally add a "valueAsDatasetGroupIndexes" method to this class too, to make that easier for callers. Same applies to QgsProcessingParameterTypeMeshDatasetGroups

src/gui/processing/qgsprocessingmeshdatasetwidget.cpp
```diff
@@ -0,0 +1,568 @@
...
+  mToolButton->setText( QString( QChar( 0x2026 ) ) );
+  hl->addWidget( mToolButton );
+
+  setLayout( hl );
+
+  mLineEdit->setText( tr( "%1 dataset groups selected" ).arg( 0 ) );
+
+  mToolButton->setPopupMode( QToolButton::InstantPopup );
+  QMenu *toolButtonMenu = new QMenu( this );
+  connect( toolButtonMenu->addAction( tr( "Current active dataset group" ) ),
```
nyalldawson, line 53:
```suggestion
  connect( toolButtonMenu->addAction( tr( "Current Active Dataset Group" ) ),
```

src/gui/processing/qgsprocessingmeshdatasetwidget.cpp
```diff
@@ -0,0 +1,568 @@
...
+           &QAction::triggered, this, &QgsProcessingMeshDatasetGroupsWidget::selectCurrentActiveDatasetGroup );
+  connect( toolButtonMenu->addAction( tr( "Select in available dataset groups" ) ),
```
nyalldawson, line 55:
```suggestion
  connect( toolButtonMenu->addAction( tr( "Select in Available Dataset Groups" ) ),
```

although this wording sounds strange to me -- should it be "select FROM available data groups"?

src/gui/processing/qgsprocessingmeshdatasetwidget.cpp
```diff
@@ -0,0 +1,568 @@
...
+  mMeshLayer = layer;
+  mValue.clear();
+  updateSummaryText();
+  emit changed();
+}
+
+void QgsProcessingMeshDatasetGroupsWidget::setValue( const QVariant &value )
+{
+  if ( value.isValid() )
+    mValue = value.type() == QVariant::List ? value.toList() : QVariantList() << value;
```
nyalldawson, line 74: here you should use the QgsProcessingParameterMeshDatasetGroups::valueAsDatasetGroupsIndexes method described above

src/gui/processing/qgsprocessingmeshdatasetwidget.h
```diff
@@ -0,0 +1,161 @@
...
+    void showDialog();
+    void selectCurrentActiveDatasetGroup();
+
+  private:
+    const QgsProcessingParameterMeshDatasetGroups *mParam = nullptr;
+    QVariantList mValue;
+
+    QLineEdit *mLineEdit = nullptr;
+    QToolButton *mToolButton = nullptr;
+    QgsMeshLayer *mMeshLayer = nullptr;
```
nyalldawson, line 51: For safety I'd make this a QPointer

src/gui/processing/qgsprocessingmeshdatasetwidget.cpp
```diff
@@ -0,0 +1,568 @@
...
+  if ( mProcessingContextGenerator )
+    context = mProcessingContextGenerator->processingContext();
+
+  if ( !context )
+  {
+    tmpContext = qgis::make_unique< QgsProcessingContext >();
+    context = tmpContext.get();
+  }
+
+  QgsMeshLayer *meshLayer = QgsProcessingParameters::parameterAsMeshLayer( wrapper->parameterDefinition(), wrapper->parameterValue(), *context );
```
nyalldawson, line 220: See logic in QgsProcessingFieldWidgetWrapper::setParentLayerWrapperValue -- you need to take ownership of the layer here.

src/gui/processing/qgsprocessingmeshdatasetwidget.cpp
```diff
@@ -0,0 +1,568 @@
...
```
nyalldawson, line 325: Also need ownership here

src/gui/processing/qgsprocessingmeshdatasetwidget.cpp
```diff
@@ -0,0 +1,568 @@
...
+
+QgsProcessingMeshDatasetTimeWidget::QgsProcessingMeshDatasetTimeWidget( QWidget *parent,
+    const QgsProcessingParameterMeshDatasetTime *param,
+    const QgsProcessingParameterWidgetContext &context ):
+  QWidget( parent ),
+  mParam( param )
+{
+  setupUi( this );
+
+  dateTimeEdit->setDisplayFormat( "yyyy-MM-dd HH:mm:ss" );
```
nyalldawson, line 390: Set the range for dateTimeEdit to the maximum possible -- e.g. see QgsTemporalControllerWidget

tests/src/gui/testprocessinggui.cpp
```diff
@@ -8869,6 +8872,169 @@ void TestProcessingGui::testTinInputLayerWrapper()
...
+  settings = layer->rendererSettings();
+  settings.setActiveScalarDatasetGroup( 0 );
+  settings.setActiveVectorDatasetGroup( -1 );
+  layer->setRendererSettings( settings );
+  datasetGroupWidget->selectCurrentActiveDatasetGroup();
+  QVERIFY( !datasetTimeWidget->isEnabled() );
+  pythonString = timeDefinition.valueAsPythonString( timeWrapper.widgetValue(), context );
+  QCOMPARE( pythonString, QStringLiteral( "{'type': 'static'}" ) );
+}
+
```
nyalldawson, line 9037: Also needs tests for the underlying parameter API too

nyalldawson:
Great work! I'm really happy to see these algorithms being ported across to core processing -- I think we'll see a lot of enhancements coming to them as a result!

src/ui/processing/qgsprocessingmeshdatasettimewidget.ui
```diff
@@ -0,0 +1,72 @@
...
+  <property name="windowTitle">
+   <string notr="true">Form</string>
+  </property>
+  <layout class="QVBoxLayout" name="verticalLayout">
+   <item>
+    <layout class="QHBoxLayout" name="horizontalLayout">
+     <item>
+      <widget class="QRadioButton" name="radioButtonCurrentCanvasTime">
+       <property name="text">
+        <string>Current Canvas Time</string>
```
nyalldawson, line 22:
```suggestion
        <string>Current canvas time</string>
```
Check box labels should be capitalized just like standard labels -- i.e. sentence case

src/ui/processing/qgsprocessingmeshdatasettimewidget.ui
```diff
@@ -0,0 +1,72 @@
...
+       </property>
+       <property name="checked">
+        <bool>true</bool>
+       </property>
+      </widget>
+     </item>
+     <item>
+      <widget class="QRadioButton" name="radioButtonDefinedDateTime">
+       <property name="text">
+        <string>Defined Date/Time</string>
```
nyalldawson, line 32:
```suggestion
        <string>Predefined date/time</string>
```

src/ui/processing/qgsprocessingmeshdatasettimewidget.ui
```diff
@@ -0,0 +1,72 @@
...
+       </property>
+       <property name="checked">
+        <bool>false</bool>
+       </property>
+      </widget>
+     </item>
+     <item>
+      <widget class="QRadioButton" name="radioButtonDatasetGroupTimeStep">
+       <property name="text">
+        <string>Dataset Group Time Step</string>
```
nyalldawson, line 42:
```suggestion
        <string>Dataset group time step</string>
```

src/ui/processing/qgsprocessingmeshdatasettimewidget.ui
```diff
@@ -0,0 +1,72 @@
...
+     </property>
+     <property name="timeSpec">
+      <enum>Qt::UTC</enum>
+     </property>
+    </widget>
+   </item>
+   <item>
+    <widget class="QLabel" name="labelCurrentCanvasTime">
+     <property name="text">
+      <string>Current Canvas Time</string>
```
nyalldawson, line 64:
```suggestion
      <string>Current canvas time</string>
```

src/core/processing/qgsprocessingparametermeshdataset.cpp
```diff
@@ -0,0 +1,201 @@
...
+    return true;
+
+  if ( type == QStringLiteral( "dataset-time-step" ) )
+  {
+    if ( value.type() != QVariant::List )
+      return false;
+    if ( value.toList().count() != 2 )
+      return false;
+  }
+  else if ( type == QStringLiteral( "current-canvas-time" ) || type == QStringLiteral( "defined-date-time" ) )
```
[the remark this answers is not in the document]
vcloarec: it is set here : https://github.com/qgis/QGIS/blob/48be634a43bdd4f790d962c71a37800c8fe90665/src/gui/processing/qgsprocessingmeshdatasetwidget.cpp#L560
I was thinking it was enough ... this option is set only if there is a map canvas in the `QgsProcessingParameterWidgetContext`.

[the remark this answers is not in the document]
nyalldawson: > But maybe in the QgsProcessingParameter no need to have this option available. Just consider this option as date/time defined. The choice is only done in the widget.

Yeah - there's confusion here because the implementation looks like it should accept a value of "current-canvas-time", but in fact that has no meaning outside of the widget.

I'd still be inclined to push for adding the time range to the context. I think that would be a nice flexible approach which will pay off for us later. It also means that an execution using "current-context-time" will work correctly when re-run from the processing history dialog, because it would correctly reflect the current canvas time instead of the time at the point of last execution...

src/analysis/processing/qgsalgorithmexportmeshvertices.cpp
```diff
@@ -0,0 +1,301 @@
...
+  addParameter( new QgsProcessingParameterFeatureSink( QStringLiteral( "OUTPUT_LAYER" ), QObject::tr( "Output Vector Layer" ), QgsProcessing::TypeVectorPoint ) );
+}
+
+static QVector<double> vectorValue( const QgsMeshDatasetValue &value, int exportOption )
+{
+  QVector<double> ret;
+  if ( exportOption == 0 || exportOption == 2 )
+  {
+    ret.append( value.x() );
+    ret.append( value.y() );
```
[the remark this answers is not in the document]
vcloarec: hmmm, not sure, there could be more values added after if `exportOption==2`. But indeed, it is possible to optimize by other way

src/analysis/processing/qgsalgorithmexportmeshvertices.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+  for ( int i = 0; i < datasetGroupsList.count(); ++i )
+  {
+    int  groupIndex = datasetGroupsList.at( i ).toInt();
+    QgsMeshDatasetIndex datasetIndex = meshLayer->datasetIndexAtRelativeTime( relativeTime, groupIndex );
+
+    DataGroup dataGroup;
+    dataGroup.metadata = meshLayer->datasetGroupMetadata( datasetIndex );
+    int dataCount = dataGroup.metadata.dataType() == QgsMeshDatasetGroupMetadata::DataOnVertices ? mNativeMesh.vertexCount() : mNativeMesh.faceCount();
+    dataGroup.datasetValues = meshLayer->datasetValues( datasetIndex, 0, dataCount );
```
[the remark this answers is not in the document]
vcloarec: It could be quite expensive with very big meshes, but it is quite reasonable  comparing with other operation. As in the processAlgorithm we have to be thread safe, I think we don't have a lot of choice, we prepare here the data that are thread safe.
I think other solutions will ask deeper modification in the data provider to make it thread safe.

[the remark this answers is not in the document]
nyalldawson: Could you potentially clone the mesh layer in prepare, and then use the clone in the process step?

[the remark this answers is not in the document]
vcloarec: The problem is to bring the data provider, especially MDAL provider...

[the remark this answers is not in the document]
nyalldawson: Is there any "in memory" mesh layer concept? If not, then couldn't you re-load a new mesh layer using the same uri?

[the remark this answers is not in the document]
vcloarec: Actually, now cloning the mesh is possible, and the data provider is recreated from the same uri. But, for now it is not a real cloning because there are some data that are not cloned as extra persistent dataset loaded in the provider and other extra dataset like virtual dataset. I was thinking about make a real cloning by writing a dom element with writeXml and then readXml with the new mesh. It will need some refactorization of the write/read xml but seems possible.

But for now I am not completly convince of this way :
We could have a new mesh layer that could be used only in the process with a thread safe way, but this new layer needs :
- recreate a new data provider, actually MDAL needs to load the mesh on its side, and with a lot of driver, it needs to load all the dataset groups too.
- to reload completly a new native mesh in QGIS side (btw make change in the mesh layer implementation to allow reloading mesh without make the triangular mesh that consume ressources)
- if simplification is enable, restart a mesh simplification that could consume a lot of ressources. But here, maybe we could avoid this simplification  for the new mesh.
Even all that could be done in the process thread, it will consume quite much more memory, especially for big meshes.

With the actual implementation, in the algorithm preparation, what is the more expensive is the transfer of the data from the data provider side to QGIS side, and just for the selected dataset. Afer all, it is "just" a copy of data without calculation, and it is not so long for ordinary mesh (it is what is done for the renderer in the main thread).

So cloning the mesh layer could reduce the preparation to almost nothing but will need more resources (time and memory) on the process thread and that will need also more or less complex modification in the mesh layer implementation.

With the actual implementation of the algorithm, it spends as much time in the main thread as the renderer and it consume just the memory needed to store the needed dataset.

src/analysis/processing/qgsalgorithmexportmeshvertices.cpp
```diff
@@ -0,0 +1,301 @@
...
+  QgsFields fields;
+  for ( int i = 0; i < mDataPerGroup.count(); ++i )
+  {
+    const DataGroup &dataGroup = mDataPerGroup.at( i );
+    if ( dataGroup.metadata.isVector() )
+    {
+      if ( mExportVectorOption == 0 or mExportVectorOption == 2 )
+      {
+        fields.append( QString( "%1_x" ).arg( dataGroup.metadata.name() ) );
+        fields.append( QString( "%1_y" ).arg( dataGroup.metadata.name() ) );
```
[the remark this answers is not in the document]
vcloarec: here, we can have several vector dataset, so names have to be distinguished. And name are predictable as dataset group name can be known.

src/gui/processing/qgsprocessingmeshdatasetwidget.cpp
vcloarec: hmmm, it is set to max in the ui file

src/core/processing/qgsprocessingparametermeshdataset.cpp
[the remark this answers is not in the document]
vcloarec: > You'd also need to add a way to set the context time via the qgis_process standalone tool, just like how the context's area/distance units can be set.
Maybe a string with ISO8601 format could be entered after --CURRENT_TIME. and we could have a function like this : https://github.com/lutraconsulting/MDAL/blob/f65103aab2df2ba7676ffadd0b98caef0a75cbef/mdal/mdal_datetime.hpp#L71

Any idea where this function could be implemented in QGIS code ?

src/gui/processing/qgsprocessingmeshdatasetwidget.cpp
vcloarec: If I understand well the logic in QgsProcessingFieldWidgetWrapper::setParentLayerWrapperValue, we need ownership because if context come from `context=tmpContext.get();` the layer will be destruct at the end of this scope.

If the context come from mProcessingContextGenerator, we also take it and if needed in other scope, the layer will produce again another one that will be in the temporary store.

If I am not wrong, for mesh layer, there will be several instance of the same layer, one more for each wrapper (here the datasetGroup and the datasetTime wrappers). What I don't like is the memory consuming of this approach which the mesh layer, as each layer have all the data (mesh , dataset,).

Does the life time of the context in the context generator is greater that the life time of the wrapper?
In that case, maybe taking the ownership will be useful only if the context generator provide a nullptr?

vcloarec:
@nyalldawson changes after your comments done. There still points to clarify (see above)

src/core/mesh/qgsmeshlayerutils.h
```diff
@@ -264,6 +276,25 @@ class CORE_EXPORT QgsMeshLayerUtils
       QgsMeshDataBlock *activeFaceFlagValues,
       const QgsMeshRendererScalarSettings::DataResamplingMethod method = QgsMeshRendererScalarSettings::NeighbourAverage );
 
+    /**
+     * Calculates magnitude values ont vertices from the given QgsMeshDataBlock.
```
PeterPetrik, line 280: typo `ont`

src/core/mesh/qgsmeshtimesettings.h
```diff
@@ -72,7 +72,7 @@ class CORE_EXPORT QgsMeshTimeSettings
   private:
 
     QString mRelativeTimeFormat = QStringLiteral( "d hh:mm:ss" );
-    QString mAbsoluteTimeFormat = QStringLiteral( "dd.MM.yyyy hh:mm:ss" );
+    QString mAbsoluteTimeFormat = QStringLiteral( "yyyy-MM-dd HH:mm:ss" );
```
PeterPetrik, line 75: why this change? this causes to fail travis test.

vcloarec: to accord date time format with other part of QGIS. This had to be done for a moment...

src/gui/processing/qgsprocessingmeshdatasetwidget.cpp
vcloarec: Finally the approach has been changed as this logic is not very adapted for mesh layer.
Now the logic is 

- when setting the mesh layer to the widget, if the mesh layer is in the project, the widget keeps a pointer to the layer to update value if needed (mesh layer can change : current active dataset groups, extra dataset groups, virtual dataset groups, ...)

- If the mesh layer is not in the project, the widget do not keep a pointer but extract all what is needed, here as the layer is not in the project, it could not change.

src/analysis/processing/qgsalgorithmexportmeshvertices.cpp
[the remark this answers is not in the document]
vcloarec: Much clearer with tracing :

- a coarse mesh (2.9 millions of faces) with only one dataset group exported:
image
Preparing (little pink one) : 18.6 ms.

- a less big mesh (730 000 faces), but with a lot of dataset group exported (20 groups):
image
Preparing (little pink one) : 1.1 s.

20 dataset groups is quite big, and I think it is a extreme case

vcloarec:
Tracing:

- a coarse mesh (2.9 millions of faces) with only one dataset group exported:
image
Preparing (little pink one) : 18.6 ms.

- a smaller mesh (730 000 faces), but with a lot of dataset groups exported (20 groups):
image
Preparing: 1.1 s.

20 dataset groups is quite big, and I think it is a extreme case.

@nyalldawson : are you happy with the last changes?

I would like to merge this one  because there are some new algorithms waiting to be pushed.

src/core/processing/qgsprocessingparametermeshdataset.h
PeterPetrik: I think this is fixed by valueAsDatasetGroup

src/analysis/processing/qgsalgorithmexportmeshvertices.cpp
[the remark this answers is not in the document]
nyalldawson: Ok, let's give it a try...

src/core/processing/qgsprocessingparametermeshdataset.cpp
```diff
@@ -0,0 +1,282 @@
...
+  return typeName();
+}
+
+bool QgsProcessingParameterMeshDatasetGroups::checkValueIsAcceptable( const QVariant &input, QgsProcessingContext *context ) const
+{
+  Q_UNUSED( context );
+  return valueIsAcceptable( input );
+}
+
+QString QgsProcessingParameterMeshDatasetGroups::valueAsPythonString( const QVariant &value, QgsProcessingContext &context ) const
```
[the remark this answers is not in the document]
vcloarec: https://github.com/qgis/QGIS/blob/38348119fe7cd3dabfbf52a81054042d62ae8c8d/tests/src/gui/testprocessinggui.cpp#L8997
https://github.com/qgis/QGIS/blob/38348119fe7cd3dabfbf52a81054042d62ae8c8d/tests/src/gui/testprocessinggui.cpp#L8998
https://github.com/qgis/QGIS/blob/38348119fe7cd3dabfbf52a81054042d62ae8c8d/tests/src/gui/testprocessinggui.cpp#L9012
https://github.com/qgis/QGIS/blob/38348119fe7cd3dabfbf52a81054042d62ae8c8d/tests/src/gui/testprocessinggui.cpp#L9013
https://github.com/qgis/QGIS/blob/38348119fe7cd3dabfbf52a81054042d62ae8c8d/tests/src/gui/testprocessinggui.cpp#L9014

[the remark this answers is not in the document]
vcloarec: ok, I got it... also in testqgsprocessing.cpp ...

PeterPetrik:
Looks like Vincent fixed all the items from review. If no objections (@nyalldawson), I would like to merge this. Thank Vincent for great work and Nyall for the in-depth reviews!

nyalldawson:
Yep, looks good to me! Thanks @vcloarec

PR #39673, opened 2020-10-28
Server wms getlegendgraphics json symbolscale

elpaso:
Add min/max scale denominators to json symbol export.


zacharlie:
@elpaso Is there any chance you could add more detail for how this is feature is implemented for end users?

elpaso:
@zacharlie this is API only. It will be used by the server landing page to hide/show symbols based on the map scale.

PR #39683, opened 2020-10-28
3D material color data defined

vcloarec:
In the latest version of QGIS (3.16), it is possible to configure color of features in 3D rendering, however this configuration is fixed and all features get the same color. In 2D rendering, it is however possible to also set up data-defined color - in such case, color of each feature is determined by a value of its attribute or through an expression that may take one or more attributes of the feature into account. The idea is to provide this functionality for 3D features.
Rendering of polygons may be currently done with two different kinds of materials, each having distinct appearance and configuration of colors:
Realistic (Phong) material - calculated using 3 colors (diffuse, ambient, specular) and taking lights configuration into account
CAD (Gooch) material - calculated using 4 colors (diffuse, warm, cool, specular), ignoring lights configuration
We are not considering the recently added textured material into account as it is still experimental and the appearance of objects is influenced by its texture image anyway.
After this work, each color selector will have an additional button on the right side to allow specification of an attribute for the color (or an expression), just like elsewhere in QGIS.

Screenshot_20201028_103215

hotSpot

categorized

Funded by Kristianstad (www.kristianstad.se)


src/app/3d/qgsgoochmaterialwidget.cpp
```diff
@@ -40,14 +40,18 @@ QgsGoochMaterialWidget::QgsGoochMaterialWidget( QWidget *parent )
   connect( spinShininess, static_cast<void ( QDoubleSpinBox::* )( double )>( &QDoubleSpinBox::valueChanged ), this, &QgsGoochMaterialWidget::changed );
   connect( spinAlpha, static_cast<void ( QDoubleSpinBox::* )( double )>( &QDoubleSpinBox::valueChanged ), this, &QgsGoochMaterialWidget::changed );
   connect( spinBeta, static_cast<void ( QDoubleSpinBox::* )( double )>( &QDoubleSpinBox::valueChanged ), this, &QgsGoochMaterialWidget::changed );
+  connect( mDiffuseDataDefinedButton, &QgsPropertyOverrideButton::changed, this, &QgsGoochMaterialWidget::changed );
```
nyalldawson, line 43: Other data defined settings use a function for this to avoid all the duplicate code -- see e.g. QgsCalloutWidget::registerDataDefinedButton. Can you follow this approach here? It ultimately  makes adding new properties a one line change for the widget.

src/app/3d/qgsgoochmaterialwidget.cpp
```diff
@@ -71,5 +80,13 @@ QgsAbstractMaterialSettings *QgsGoochMaterialWidget::settings()
   m->setShininess( spinShininess->value() );
   m->setAlpha( spinAlpha->value() );
   m->setBeta( spinBeta->value() );
+
+  QgsPropertyCollection dataDefinedPropertie;
```
nyalldawson, line 84: Here you could use

    const QList< QgsPropertyOverrideButton * > buttons = findChildren< QgsPropertyOverrideButton * >();

and loop over the found widgets to set their properties. That avoids more duplicate code.

nyalldawson, line 84:
```suggestion
  QgsPropertyCollection dataDefinedProperties;
```

nyalldawson:
Great stuff @vcloarec  -- this has been on my wishlist for a long time!

Code looks good here too. One concern I do have is with the introduction of new shaders which are only used in certain circumstances. I'm concerned that we're seeing a slow explosion of shaders, which are all only partially compatible with the range of settings available in 3d maps. (We already see this in the point shaders where some material settings don't work with certain lighting settings). Would it be possible to ALWAYS use the new shaders here, regardless of whether data defined colors are active or not?

wonder-sk:
> One concern I do have is with the introduction of new shaders which are only used in certain circumstances. I'm concerned that we're seeing a slow explosion of shaders, which are all only partially compatible with the range of settings available in 3d maps. (We already see this in the point shaders where some material settings don't work with certain lighting settings). Would it be possible to ALWAYS use the new shaders here, regardless of whether data defined colors are active or not?

Yeah this is a common concern in 3D apps - the number of shaders and the possible number of combinations can get out of hand easily. In some apps/frameworks they use conditional building of shaders to keep the number of shaders low, but at the expense of their complexity. We have discussed this with Vincent and thought that the current approach is a good compromise. The issue is that data-defined colors for phong material adds 12 bytes, for gooch it adds 16 bytes. Right now we use 12 bytes for position and 12 bytes for normals (this could be optimized though to fewer bytes), so the data-defined shaders add 50% or more data per vertex - which is quite significant extra baggage. That's why we wanted to keep the original shaders unchanged and highly efficient, and only pay the extra cost where it is needed...

nyalldawson:
@wonder-sk 

Good point.

thought: what about instead of using the qt shader for the non data defined path, we instead use a copy of the new shader introduced here but tweaked for constant colouring? That way we'd have an easier path to implement new functionality in both shaders, as the diff between them is likely to be relatively small...

roya0045:
@vcloarec Great addition!

wonder-sk:
> thought: what about instead of using the qt shader for the non data defined path, we instead use a copy of the new shader introduced here but tweaked for constant colouring? That way we'd have an easier path to implement new functionality in both shaders, as the diff between them is likely to be relatively small...

I am not sure how you mean tweaking a shader for constant colouring... Do you mean it as having the body something like this?
```
main()
{
  if (hasDataDefinedColors)
  { /* do more expensive work */ }
  else
  { /* do the orignal constant shading */ }
}
```
The issue here is that the shader also needs to have the inputs declared - for constant shading these are "uniforms" (basically constant parameters passed to the GPU before execution), for data-defined shading these are "in/out" parameters that are set in vertex shader and consumed in fragment shader... in shader programs these are fundamentally different and I am not sure how we would combine them... Or did you mean something else?

nyalldawson:
No what I was meaning was copy the whole shader and then make the colors constant in the copy. We'd still have two shaders, but instead of the non data defined material using the qt shader, it would instead be using the constantified qgis shader. The benefit is that we'd be able to keep these two as "in sync" as possible, vs one completely custom shader and one which is coming from the depths of qt.

DelazJ:
@vcloarec Nice. Would this cover a case that I applied a categorized rendering to my layer, and when I enable the 3D rendering, I see my features with their corresponding 2D color. Kind of using `@symbol_color` variable?

vcloarec:
@DelazJ , Unfortunately, this variable is not present in this context.

src/3d/materials/qgsabstractmaterialsettings.h
```diff
@@ -170,8 +178,60 @@ class _3D_EXPORT QgsAbstractMaterialSettings SIP_ABSTRACT
...
+    * Returns the symbol material property collection, used for data defined overrides.
+    * \since QGIS 3.18
+    */
+    QgsPropertyCollection dataDefinedProperties() const;
+
+    /**
+    * Returns a reference to the material properties definition, used for data defined overrides.
+    * \since QGIS 3.18
+    */
+    const QgsPropertiesDefinition  &propertyDefinition() const;
```
nyalldawson, line 208:
```suggestion
    const QgsPropertiesDefinition  &propertyDefinitions() const;
```

:wink:

needs to match all the other methods which return property definitions

vcloarec:
hmmm again a failing test with 3D rendering and Travis. On my side, the test is ok... 
I have no idea where the issue is... I wonder if it is not due to hardware config. Maybe the rendering could vary with GPU type?
By the way, in my computer I had never seen the testExtrudedPolygon be sucessful.

roya0045:
It's strange how the edges are different https://cdash.orfeo-toolbox.org/testDetails.php?test=24496536&build=55062

roya0045:
Could it be just a slight shade of grey off?

vcloarec:
> It's strange how the edges are different https://cdash.orfeo-toolbox.org/testDetails.php?test=24496536&build=55062

Frankly, visualy, I don't see any difference. Here that is the new test image I create today on my computer.
And when I run the other old test (not data defined), it fail with this difference image:
polygon3d_extrusion_result_diff

That's look as the same difference image but with not the same test.
It is like if my computer do not generate exactly the same black...

vcloarec:
ok, positioning are not the same, 

Peek 30-10-2020 22-04

Could be fixed (like other one control image) by adding to the image control the file generated by Travis

nyalldawson:
@vcloarec I see the same locally, and can't explain it either. That's why I just started adding the alternative render images. 

What operating system do you use?

vcloarec:
@nyalldawson : kubuntu 20.04

nyalldawson:
Ok. My suspicion is that something has subtely changed in the camera handling in qt, which causes the difference

vcloarec:
ahh, yes ... possible, but maybe also elsewhere... here, it seems more positioning on the "floor"

tests/src/3d/testqgs3drendering.cpp
```diff
@@ -398,6 +399,59 @@ void TestQgs3DRendering::testExtrudedPolygons()
   QVERIFY( renderCheck( "polygon3d_extrusion", img, 40 ) );
 }
 
+void TestQgs3DRendering::testExtrudedPolygonsDataDefined()
+{
+  QgsPropertyCollection propertyColection;
+  QgsProperty diffuseColor;
+  QgsProperty ambientColor;
+  QgsProperty specularColor;
+  diffuseColor.setExpressionString( QStringLiteral( "'blue'" ) );
```
nyalldawson, line 408: This test would be more useful if we included a conditional statement in here, so that we're certain that not all features are getting the same result.

nyalldawson:
@vcloarec new code looks good to me.

I'm still interested in hearing your thoughts on the option of copying & trimming your new shader and using it instead of the qt one

vcloarec:
> I'm still interested in hearing your thoughts on the option of copying & trimming your new shader and using it instead of the qt one

Here new shaders are introduced because color parameters are depending on vertices, and that is not supported by native Qt shaders. So no other choice to rewrite our proper shaders... The formulas for this shaders are quite simple, and original shaders are relatively easy to reproduce. I don't think there is a reason that the formula will diverge in the future... Why not write our proper shaders and use them for not datadefined rendering? Well, that's need to reconstruct the material with parameters, and all other kinds of stuff and Qt does that quite well without anything to do. Recently, I see the danger of proper shaders, spent long hours to find why the shaders work on my computer and not with Travis... It seems shader could be quite sensitive to uncontrolled parameters, maybe OS or hardware... I think letting Qt handles those problems is quite a good idea. If Qt has shaders that work as we want, let's keep them! 
Another argument is the future and possible variation of shaders language or/and GPU hardware. Qt seems able to handle a variety of hardware and shaders could be handle with shaders graph that will allow having custom shaders. I do not explore those shaders graph and doc about are relatively non-existent... Using those shaders graph to make custom shaders seems to be cross-platform (OS/GPU hardware) and appears the way where to go in the future.
In this case, we will not need to write the shaders in openGL SL, Qt will handle the conversion of graph to adapted language for the platform. So, today, maybe we have to limit the count of proper shaders written in openGL SL.

tests/src/3d/testqgs3drendering.cpp
vcloarec: polygon3d_extrusion_data_defined

nyalldawson: Thanks!

wonder-sk:
@nyalldawson regarding shaders - if I understand correctly, you don't want to be dependent on built-in Qt shaders if we have some customized versions to avoid possible issues in the future if Qt shader implementation gets modified, right? Would you like to address that in this PR or shall we defer it for later? My preference would be to defer it because:
- since Qt 5.10 there is QShaderProgramBuilder class that can be passed a shader graph and it will build the shaders programs. It should allow what we need (variations of inputs). I still need to get my head around it, the documentation is sparse as usual. The shader graphs probably need to be created with KDAB's Kuesa tool. Some slides here: https://www.kdab.com/wp-content/uploads/stories/qt3d_node_editor_and_shader_generator.pdf
- Qt includes also shader implementations for other backends (e.g. OpenGL ES) which we don't support yet

Jean-Roc:
FYI, a [Qt blog post ](https://www.qt.io/blog/graphics-in-qt-6.0-qrhi-qt-quick-qt-quick-3d)was published last week mentioning the shader management/pipeline of Qt 6.0

nyalldawson:
@wonder-sk 
> regarding shaders - if I understand correctly, you don't want to be dependent on built-in Qt shaders if we have some customized versions to avoid possible issues in the future if Qt shader implementation gets modified, right?

Actually, it's the opposite -- I'm concerned that in future some new lighting type or other scene level setting will be introduced in qgis and either the data defined shader or non-data defined, default qt shader won't end up with support for it. And then the developer at that time will brush it off with an "out of scope for current work" type comment.

Ultimately we risk ending up with a bunch of semi-broken functionality in the 3d framework which only work for users under limited circumstances. 

So I am keen to see the differences between various shaders used for different code paths minimized, so that at least it's easy for developers adding functionality in future to copy/paste their changes across shaders.

Let me know if I'm being pedantic though! :smile:

wonder-sk:
> Actually, it's the opposite -- I'm concerned that in future some new lighting type or other scene level setting will be introduced in qgis and either the data defined shader or non-data defined, default qt shader won't end up with support for it. And then the developer at that time will brush it off with an "out of scope for current work" type comment.

Hmm for me that is not really a concern - as long we are using the common "forward" renderer approach where each triangle is rendered in one step, adding new features to rendering should not be affecting individual materials. That's why I would prefer to use Qt materials/shaders as much as we can - less code to maintain for us. For example, we can take QMetalRoughMaterial and start using it with minimal effort to get more realistic looking shading.

If someone would be introducing some changes to the 3D renderer that would require changes to multiple materials, that would be a pretty serious change and we would be forced to forget about any Qt material implementations and have all materials/shaders on our own. Such a situation would happen e.g. if we would switch to a "deferred" renderer (https://learnopengl.com/Advanced-Lighting/Deferred-Shading) but that's a kind of change I would like to avoid unless really necessary...

To me, dealing with shader code in QGIS code is a necessary evil and I am trying to optimize towards having less of it rather than more. Ideally I would prefer to have zero shaders in QGIS code and delegate all that to Qt3D. Also my general approach is that if something can be done cheaper in a shader or more expensive outside of a shader, I prefer the latter, as maintenance of shaders is a pain (especially if we are not a team of dedicated 3D graphics engineers). And with fewer shaders it will be also easier to move to Qt 6 where we may need to provide shaders for multiple backends (Vulkan, DirectX, Metal).

nyalldawson:
@wonder-sk 

Ok -- I'm happy to follow your guidance here!

wonder-sk:
Cool - hopefully I am not misguiding all of us on this topic :-)

It would be great to have access to some experienced 3D graphics developers one day to get some guidance on these things...

nyalldawson:
@vcloarec 
this has introduced a UI issue, where the data defined settings appear as available for the phong material in non-polygon contexts. eg they show as an option for line symbols but can't be used there (they result in broken rendering if set).

I can see two ways to handle this -- one would be to subclass the QgsPhongMaterialSettings and QgsGoochMaterialSettings classes to new "QgsPhongDataDefinedMaterialSettings" / "QgsGoochDataDefinedMaterialSettings" classes and refine the subclasses' supportsTechnique method (with a new technique added for polygons with data defined colors). Or add some additional api into QgsAbstractMaterialSettings and the material widgets to selectively show the data defined options only in certain circumstances.

vcloarec:
@nyalldawson 
Indeed...
I think the second one is simpler, I make a PR to fix that

PR #39686, opened 2020-10-28
few fixes in custom processing parameters

alexbruy:
## Description
Fixes some issues in custom Processing parameters (based on the review comments in #39575)

- Make `QgsProcessingParameterVectorTileWriterLayers` public so it can be used in other algorithms.
- Mark `QgsProcessingParameterTypeTinInputLayers` and `QgsProcessingParameterTypeVectorTileWriterLayers` as private API
- Use member context in vector tile widget wrapper to avoid dangling pointers.

### DIFF
--- a/src/core/processing/qgsprocessingparametertininputlayers.h
+++ b/src/core/processing/qgsprocessingparametertininputlayers.h
@@ -65,10 +65,13 @@ class CORE_EXPORT QgsProcessingParameterTinInputLayers: public QgsProcessingPara
     static QString typeName() { return QStringLiteral( "tininputlayers" ); }
 };
 
+///@cond PRIVATE
+
 /**
  * Parameter type definition for QgsProcessingParameterTinInputLayers.
  *
  * \ingroup core
+ * \note This class is not a part of public API.
  * \since QGIS 3.16
  */
 class CORE_EXPORT QgsProcessingParameterTypeTinInputLayers : public QgsProcessingParameterType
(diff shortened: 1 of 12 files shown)


src/core/processing/qgsprocessingparametertininputlayers.h
```diff
@@ -65,10 +65,13 @@ class CORE_EXPORT QgsProcessingParameterTinInputLayers: public QgsProcessingPara
     static QString typeName() { return QStringLiteral( "tininputlayers" ); }
 };
 
+///@cond PRIVATE
```
nyalldawson, line 68: I'd sip skip this too -- it was never supposed to be public api, and there's no need for python code to be using it so no danger in removing it now.

src/core/processing/qgsprocessingparametervectortilewriterlayers.h
```diff
@@ -66,6 +65,7 @@ class CORE_EXPORT QgsProcessingParameterVectorTileWriterLayers : public QgsProce
 
 };
 
+///@cond PRIVATE
```
nyalldawson, line 68: Same here, sip skip this one too

nyalldawson:
Thanks!

src/core/processing/qgsprocessingparametertininputlayers.h
alexbruy: Done.

src/core/processing/qgsprocessingparametervectortilewriterlayers.h
alexbruy: Done.

alexbruy:
I also removed from SIP some other parameter types in addition to vector tiles and tin inputs.

nyalldawson (approved):

PR #39695, opened 2020-10-29
[processing] add parameterAsEnumString() method to retrieve the string value of the enum

alexbruy:
# Description
Allow enum parameter accept static (non-translatable) strings as enum values. This can be useful when enum values are platform-dependent and consequently retrieving them by index is not reliable, for example this can be the case of available encodings or formats.

New argument `usesStaticStrings` was added to the enum parameter constructor as well as helper methods `usesStatisStrings()` and `setUsesStaticStrings()`. If `usesStaticStrings` is set to `true` all enum values are treated as non-translatable strings and `parameterAsEnumString()` or `parameterAsEnumStrings()` methods should be used to evaluate enum parameter value to a static string value.

DXF Export algorithm was updated to use static strings for its encoding parameter.

Related to #39575


nirvn:
Nice, that could already be used in the export atlas/print layout to images alg for the file format param.

src/core/processing/qgsprocessingparameters.cpp
```diff
@@ -511,6 +511,28 @@ QList<int> QgsProcessingParameters::parameterAsEnums( const QgsProcessingParamet
...
+
+  return parameterAsEnumString( definition, parameters.value( definition->name() ), context );
+}
+
+QString QgsProcessingParameters::parameterAsEnumString( const QgsProcessingParameterDefinition *definition, const QVariant &value, const QgsProcessingContext &context )
+{
+  if ( !definition )
+    return QString();
+
+  int index = parameterAsInt( definition, value, context );
```
nyalldawson, line 527: I think there's actually two different requirements we've got at play here:

1. A method which converts an integer parameter value back to a string (which you've done here).
and 
2. A method which allows the parameter value to be specified as a string value, but restricting it to a list of available options

For the dxf encoding option we actually need number 2 -- a way of having the encoding option be specified as a string value, so that it's not an integer at all and is not going to change if a particular platform has a different set of encodings available (or a different ordering of encodings).

So the API included here is useful, but it's not a solution to the DXF encoding type issue, which would need to be handled differently.

alexbruy:
@nyalldawson sorry for long delay with this. Hopefully now implementation addresses the issue as we have discussed.

nyalldawson (approved):
Looks good!

The last missing part would be to modify the widget wrapper so that it returns string values for the parameter whenever the parameter definition indicates that it uses static string values...

alexbruy:
If there are no objections I'm going to merge this.

nyalldawson:
Looks great to me, merge away!

PR #39704, opened 2020-10-29
Fix logic in QgsUserInputWidget which incorrectly deletes other widgets whenever one existing widget is deleted (3.10 backport)

nyalldawson:
Fixes calls to iface.addUserInputWidget() does not show any widget if
an existing user input widget is already shown

(cherry picked from commit d8351c0a211588c7eed9bc3a887ca8c38fe950c0)


nirvn (approved):

nyalldawson:
Moved to https://github.com/qgis/QGIS/pull/39604

PR #39709, opened 2020-10-30
[layer tree] Fix corrupted tree when filter (by extent) is on with legend widget(s) enabled

nirvn:
## Description

Without that fix, legend widgets are corrupting the layer tree when users have filter by extent switched on.

### DIFF
--- a/src/core/layertree/qgslayertreemodel.cpp
+++ b/src/core/layertree/qgslayertreemodel.cpp
@@ -1230,22 +1230,38 @@ QList<QgsLayerTreeModelLegendNode *> QgsLayerTreeModel::filterLegendNodes( const
     {
       for ( QgsLayerTreeModelLegendNode *node : qgis::as_const( nodes ) )
       {
-        QString ruleKey = node->data( QgsSymbolLegendNode::RuleKeyRole ).toString();
-        bool checked = mLegendFilterUsesExtent || node->data( Qt::CheckStateRole ).toInt() == Qt::Checked;
-        if ( checked )
+        switch ( node->data( QgsSymbolLegendNode::NodeTypeRole ).value<QgsLayerTreeModelLegendNode::NodeTypes>() )
         {
-          if ( QgsVectorLayer *vl = qobject_cast<QgsVectorLayer *>( node->layerNode()->layer() ) )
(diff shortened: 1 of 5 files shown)


nyalldawson (approved):
Fix is fine for what it is -- but it could be done a little cleaner:

1. Add a new role to QgsLayerTreeModelLegendNode::LegendNodeRoles for "NodeTypeRole"
2. Add a new enum to QgsLayerTreeModelLegendNode for "NodeTypes", with values "EmbeddedWidget", "DataDefinedSizeLegend", "ImageLegend", "RasterSymbolLegend", "SimpleLegend", "SymbolLegend", "WmsLegend", and make the corresponding QgsLayerTreeModelLegendNode subclasses return the correct value in their data implementations for the NodeTypeRole role
3. Retrieve the node type from the node's data here, and use a switch here to explicitly control whether various types should be subject to the filtering

nyalldawson:
Nice work -- thanks!

src/core/layertree/qgslayertreemodel.cpp
```diff
@@ -1216,22 +1216,36 @@ QList<QgsLayerTreeModelLegendNode *> QgsLayerTreeModel::filterLegendNodes( const
...
+                  filtered << node;
+              }
+              else
+              {
+                filtered << node;
+              }
+            }
+            else  // unknown node type or unchecked
+              filtered << node;
+            break;
```
nyalldawson, line 1247:
```suggestion
            {
            const QString ruleKey = node->data( QgsSymbolLegendNode::RuleKeyRole ).toString();
            bool checked = mLegendFilterUsesExtent || node->data( Qt::CheckStateRole ).toInt() == Qt::Checked;
            if ( checked )
            {
              if ( QgsVectorLayer *vl = qobject_cast<QgsVectorLayer *>( node->layerNode()->layer() ) )
              {
                if ( mLegendFilterHitTest->legendKeyVisible( ruleKey, vl ) )
                  filtered << node;
              }
              else
              {
                filtered << node;
              }
            }
            else  // unknown node type or unchecked
              filtered << node;
            break;
            }
```

I'm not sure how this compiles right now :P

PR #39710, opened 2020-10-30
Add adapter class QgsGeocoderLocatorFilter

nyalldawson:
This class implements the required logic to bridge a
class which implements the QgsGeocoderInterface interface to a
QgsLocatorFilter. It allows easy creation of a locator filter
from a geocoder.

E.g.

    class GoogleGeocoder(QgsGeocoderInterface):
      ... class which implements QgsGeocoderInterface...

    my_google_geocoder = GoogleGeocoder(api_key)
    my_google_locator_filter = QgsGeocoderLocatorFilter('google', 'Google Maps', 'google', my_google_geocoder, iface.mapCanvas())
    iface.registerLocatorFilter(my_google_locator_filter)

Note:

There's an abstract base class QgsAbstractGeocoderLocatorFilter which
lives in core. This base class has the guts of the geocoder to filter
adapter logic, but it does NOT implement the part which actually zooms
the canvas to the geocode results (because it's in core!)

This abstract base class is in place for applications like QField,
where the GUI library and QgsMapCanvas is NOT used. Those applications
can then implement their own concrete class based on QgsAbstractGeocoderLocatorFilter
with the correct logic to zoom their interface to a region. Ultimately,
a single QgsGeocoderInterface could be usable across QGIS desktop, QField,
with almost all logic shared via the common abstract adapter class.

## Example

Adding a google geocoder to the locator bar (insert your own Google Maps API key!!)

```
api_key = 'not mine'
import urllib.parse
import json


class GoogleGeocoder(QgsGeocoderInterface):
   
    def __init__(self, api_key):
        super().__init__()
        self.api_key = api_key
        
        self.endpoint = "https://maps.googleapis.com/maps/api/geocode/json?"
        
    def flags(self):
        return QgsGeocoderInterface.Flag.GeocodesStrings
        
    def get_request_url(self, address, bounds):
        bounds_string = ''
        if bounds:
            bounds_string = f'&bounds={bounds.yMinimum()},{bounds.xMinimum()}|{bounds.yMaximum()},{bounds.xMaximum()}'
        
        return self.endpoint + f"sensor=false{bounds_string}&address={urllib.parse.quote(address)}&key={self.api_key}"
        
    def geocodeString(self, string, context, feedback):
        all_results = []

        bounds = None
        if context.areaOfInterest():
            g = context.areaOfInterest()
            ct = QgsCoordinateTransform(context.areaOfInterestCrs(), QgsCoordinateReferenceSystem('EPSG:4326'), context.transformContext())
            g.transform(ct)
            bounds = g.boundingBox()
            
        
        url = self.get_request_url(string, bounds)
        
        request = QNetworkRequest(QUrl(url))
        new_req = QgsBlockingNetworkRequest()
        error_code = new_req.get(request, feedback=feedback)
        
        if error_code != QgsBlockingNetworkRequest.NoError:            
            return [QgsGeocoderResult.errorResult(new_req.errorMessage())]

        try:
            json_array = json.loads(new_req.reply().content().data().decode("utf-8"))
        except:
            return []

		# Interpret the response

        # https://developers.google.com/maps/documentation/geocoding/intro#GeocodingRequests
        if (not "status" in json_array) or \
           (not "results" in json_array) or \
           (len(json_array['results']) <= 0):
            return [QgsGeocoderResult.errorResult('no results')]
        
        if "error_message" in json_array:
            error_message = json_array["error_message"]
            return [QgsGeocoderResult.errorResult(error_message)]

        if (json_array["status"] == "REQUEST_DENIED") or \
           (json_array["status"] == "OVER_QUERY_LIMIT"):
           return [QgsGeocoderResult.errorResult('over limit')]

        if json_array["status"] != "OK":
            return [QgsGeocoderResult.errorResult(json_array["status"])]

        for result_num, result in enumerate(json_array["results"]):
            latitude = float(result["geometry"]["location"]["lat"])
            longitude = float(result["geometry"]["location"]["lng"])
            
            geom = QgsGeometry.fromPointXY(QgsPointXY(longitude, latitude))
            res = QgsGeocoderResult(result.get('formatted_address'), geom, QgsCoordinateReferenceSystem('EPSG:4326'))
            
            attributes = {}
            attributes['result_num'] = result_num
            attributes['status'] = str(json_array["status"])
            
            if "formatted_address" in result:
                attributes['formatted_address'] = str(result["formatted_address"])

            if "place_id" in result:
                attributes['place_id'] = str(result["place_id"])
            
            if "location_type" in result["geometry"]:
                attributes['location_type'] = str(result["geometry"]["location_type"])
                
            res.setAdditionalAttributes(attributes)

            all_results.append(res)

        return all_results

        
google_geocoder = GoogleGeocoder(api_key)
google_locator_filter = QgsGeocoderLocatorFilter('Google', 'Google', 'addr', google_geocoder, iface.mapCanvas())
iface.registerLocatorFilter(google_locator_filter)
```

Peek 2020-10-30 14-41


nyalldawson:
@3nids @rduivenvoorde 

This is part 2 -- there's more clues as to how to use the new API here.

@3nids I'd especially like your opinion on the QField compatibleness of this approach!

src/core/geocoding/qgsabstractgeocoderlocatorfilter.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include "qgsabstractgeocoderlocatorfilter.h"
+#include "qgsgeocoder.h"
+#include "qgsgeocodercontext.h"
+
+QgsAbstractGeocoderLocatorFilter::QgsAbstractGeocoderLocatorFilter( const QString &name, const QString &displayName, const QString &prefix, QgsGeocoderInterface *geocoder )
+  : mName( name )
+  , mDisplayName( displayName )
+  , mPrefix( prefix )
+  , mGeocoder( geocoder )
+{
```
3nids, line 25: why don't you inherit QgsLocatorFilter here?

3nids: and have the clone method declared as pure virtual?

3nids (approved):
This looks perfect to me!

src/core/geocoding/qgsabstractgeocoderlocatorfilter.cpp
nyalldawson: > why don't you inherit QgsLocatorFilter here?

Ok, done -- I had to add the parent argument to the geocoder filter classes too

> and have the clone method declared as pure virtual?

nyalldawson: > and have the clone method declared as pure virtual?

It's already pure virtual -- it gets inherited from QgsLocatorFilter as a pure virtual method

3nids: perfect, thanks

PR #39712, opened 2020-10-30
[processing] Fix add (single) row tool button broken

### LINKED ISSUE #39696: Add layer to batch processing not working anymore
**Describe the bug**
Adding new layers to batch processing is not working in the latest release

**How to Reproduce**

1. Open processing toolbox
2. Select a tool where batch processing is possible (e.g. Export to PostgreSQL)
3. Try to add a new layer to the batch with the green ➕ icon.
4. A new row does not appear where it is supposed to. 

**QGIS and OS versions**

QGIS 3.16 on Windows 10. Fresh profile with no plugins. 

**Additional context**

Below is an animation of the feature not working. 

batc_process

nirvn:
## Description

Fixes #39696.

### DIFF
--- a/python/plugins/processing/gui/BatchPanel.py
+++ b/python/plugins/processing/gui/BatchPanel.py
@@ -444,7 +444,7 @@ class BatchPanel(QgsPanelWidget, WIDGET):
         self.alg = alg
         self.parent = parent
 
-        self.btnAdd.clicked.connect(self.addRow)
+        self.btnAdd.clicked.connect(lambda: self.addRow(1))
         self.btnRemove.clicked.connect(self.removeRows)
         self.btnOpen.clicked.connect(self.load)
         self.btnSave.clicked.connect(self.save)


roya0045:
This seems strange that the default argument is ignored to be honest. If this is the case, would it be better to emit 1 when the signal is sent or to add a new function to circumvent the issue? I'm just asking to see what would be best practice.

nirvn:
It's actually not ignored, but set to false due to the clicked signal being connected as one checked parameter, which it passes onto nb.

A if nb is False conditional check at the beginning of addRow could be an alternative approach, I just felt it was a bit more hackish.

roya0045:
@nirvn Thank you for the info! I didn't know how qt handled those case.

nyalldawson (approved):

nyalldawson:
(insert usual grumbles about fragility of Python code here!)

PR #8154, opened 2018-10-09
use Ubuntu cosmic for base image

3nids:
This will easily allow newer version of Qt but mainly SIP for auto build of the Python API docs.

WIP!

PR #39718, opened 2020-10-30
enable test on Ubuntu Focal

3nids:
This is a preliminary work to dependencies bump.

Then, the question is how (who) to handle the failing tests.

There was a former try some time ago: https://github.com/qgis/QGIS/pull/8154/files


3nids:
coming back with a better approach, moving directly to github workflows

Issue #39097, opened 2020-09-30
Add more contextualized and meaningful examples to geometry_overlay_* functions


DelazJ:
Currently, the examples are all the same (copy-paste?)
```
geometry_overlay_*('regions') → True
geometry_overlay_*('regions', name) → ['South Africa', 'Africa', 'World']
geometry_overlay_*('regions', name, name != 'World') → ['South Africa', 'Africa']
geometry_overlay_*('regions', name, limit:=1) → ['South Africa']
```
leading to:
- `geometry_overlay_disjoint ('regions', name) → ['South Africa', 'Africa', 'World']`: how could something be disjoint to World in GIS context (assuming World is a feature)?
- `geometry_overlay_equals('regions', name) → ['South Africa', 'Africa', 'World']`: these results are not equal
- It's quite _unusual_ to have a layer in which you have countries, continents... and World?

I think it's good that we provide actual results in the examples instead of a text explaining what the results would be. And also that we keep using the same dataset among these related functions (though this should not be a requirement). But i'm afraid our examples are not yet good.

DelazJ:
Fixed by #39389 #39413 and #39549

PR #39726, opened 2020-10-31
[3.16] Update expression overlay functions

DelazJ:
backports #39413 and #39549 because they fix actual issues with function description (eg st_within is reversed) and examples (#39097, #39069)

### DIFF
--- a/resources/function_help/json/overlay_contains
+++ b/resources/function_help/json/overlay_contains
@@ -2,7 +2,7 @@
   "name": "overlay_contains",
   "type": "function",
   "groups": ["GeometryGroup"],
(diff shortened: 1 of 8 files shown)


m-kuhn:
How is that playing with translations?

DelazJ:
As usual... 3.16 continues to be translated so 3.16.x will show English unless translators have done the update in the meantime. TBH, I have serious concerns with the current examples and keeping translation (at .0 for an LTR) would not be a valid argument to me.
Also note that these are master's text so in transifex they can translate the strings once and pick them for the other version.

m-kuhn:
Ok, let's proceed, it looks like valid fixes.
It would be very interesting to see some statistics about how much translation is being done on release branches at some points.

DelazJ:
>  It would be very interesting to see some statistics about how much translation is being done on release branches at some points.

@m-kuhn not statistics, and OK, line numbering changes contribute a lot(?) to the diff but looking at https://github.com/qgis/QGIS/commit/264f8c56cc4d4cc46c941033304bec5ca4d73ed3 there are many language communities continuing to update 3.16 translation. And [3.10](https://github.com/qgis/QGIS/commit/f47a575f4bab82878aeb7dacb532b7a8b321ec82)

PR #39734, opened 2020-11-01
[processing] swap wrongly set default and minimum value for pixel size in the create constant raster algorithm (fix #39732)

### LINKED ISSUE #39732: "Create constant raster layer" has hardcoded limit for pixel size
**Describe the bug**
If you enter a pixel size smaller than 0.01 in the "Create constant raster layer" processing tool, it will automatically change it to 0.01.

When passing a smaller pixel size to it via pyqgis, it will fail with a `QgsProcessingException`: "Incorrect parameter value for PIXEL_SIZE"

This makes it impossible to use for detailed rasters in coordinates systems with big values like EPSG:4326.

**How to Reproduce**

1. Set your project CRS to EPSG:4326
2. Launch the "Create constant raster layer" processing tool
3. Enter 0.001 as "Pixel size"
4. The tool will automatically set the value to 0.01

**QGIS and OS versions**
QGIS version | 3.17.0-Master | QGIS code revision | 7a14973234
-- | -- | -- | --
Compiled against GDAL/OGR | 3.3.0dev-a9ff452a51-dirty | Running against GDAL/OGR | 3.3.0dev-a9ff452a51-dirty
Compiled against SQLite | 3.33.0 | Running against SQLite | 3.33.0
PostgreSQL Client Version | 12.4 | SpatiaLite Version | 4.3.0a
QWT Version | 6.1.5 | QScintilla2 Version | 2.11.5
OS Version | Arch Linux

alexbruy:
## Description

Seems default and minimum values in the Create Constant Raster algorithm were mixed by mistake, so default was smaller than allowed minimum value.

Fixes #39732

### DIFF
--- a/src/analysis/processing/qgsalgorithmconstantraster.cpp
+++ b/src/analysis/processing/qgsalgorithmconstantraster.cpp
@@ -66,7 +66,7 @@ void QgsConstantRasterAlgorithm::initAlgorithm( const QVariantMap & )
   addParameter( new QgsProcessingParameterExtent( QStringLiteral( "EXTENT" ), QObject::tr( "Desired extent" ) ) );
   addParameter( new QgsProcessingParameterCrs( QStringLiteral( "TARGET_CRS" ), QObject::tr( "Target CRS" ), QStringLiteral( "ProjectCrs" ) ) );
   addParameter( new QgsProcessingParameterNumber( QStringLiteral( "PIXEL_SIZE" ), QObject::tr( "Pixel size" ),
-                QgsProcessingParameterNumber::Double, 0.00001, false, 0.01 ) );
+                QgsProcessingParameterNumber::Double, 0.01, false, 0.00001 ) );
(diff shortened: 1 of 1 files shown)


kannes:
0.00001 would still be about 1 meter! https://gis.stackexchange.com/questions/8650/measuring-accuracy-of-latitude-and-longitude/8674#8674

I can easily imagine people wanting to map with millimeter accuracy (ignoring geodesy ;) ) so it should be 7 zeros at least.

Is setting a minimum needed at all?

nyalldawson:
@alexbruy are you able to set the minimum as zero and then raise exceptions in the actual algorithm execution if the pixel size isn't > 0? That should address @kannes concerns.

alexbruy:
@nyalldawson I'd rather close this, as it addressed in another PR I'm working on.

agiudiceandrea:
> I'd rather close this, as it addressed in another PR I'm working on.

Hi @alexbruy, have you managed to fix the issue in another PR?

PR #39742, opened 2020-11-02
[api] Add new class QgsGoogleMapsGeocoder

nyalldawson:
This geocoder utilises the Google Maps "geocoding" API in order to geocode
strings. The Google Maps service is not publicly available, and accordingly
an API key must be first obtained from Google and specified when constructing
this class.

(The user is responsible for managing their Google Maps API key, and ensuring
that the use of this geocoder does not exceed their usage limits! Excessive use
of the Google Maps geocoder API can result in charges being applied to the API key
holder.)

This is raw underlying API only. It is intended to be exposed to QGIS users only
via plugins, and does not result in any out-of-the-box Google Maps functionality.

In order for a plugin to use it, they must do something like this:

    # create a google maps geocoder
    api_key = 'my api key'
    coder = QgsGoogleMapsGeocoder(api_key)

    # add it to the locator bar
    filter = QgsGeocoderLocatorFilter('Google', 'Google', 'addr', coder, iface.mapCanvas())
    iface.registerLocatorFilter(filter)

Mini FAQ:

Q: Why is this being added to core, instead of a plugin?
A: While a plugin would be perfectly acceptable if we are only targetting QGIS desktop,
the intention here is to allow this underlying, low level class to be reused outside
of QGIS desktop (e.g. on QField) with a minimal amount of duplicate effort.


nyalldawson:
Ping @3nids @rduivenvoorde @nirvn

src/core/geocoding/qgsgooglemapsgeocoder.h
```diff
@@ -0,0 +1,83 @@
...
+    QgsGeocoderResult jsonToResult( const QVariantMap &json ) const;
+
+    /**
+     * Sets a specific API \a endpoint to use for requests. This is for internal testing purposes only.
+     */
+    void setEndpoint( const QString &endpoint );
+
+  private:
+
+    QString mApiKey;
```
3nids, line 77: can we have setter/getter for API key and region?

src/core/geocoding/qgsgooglemapsgeocoder.cpp
```diff
@@ -0,0 +1,253 @@
...
+QgsGeocoderResult QgsGoogleMapsGeocoder::jsonToResult( const QVariantMap &json ) const
+{
+  const QVariantMap geometry = json.value( QStringLiteral( "geometry" ) ).toMap();
+  const QVariantMap location = geometry.value( QStringLiteral( "location" ) ).toMap();
+  const double latitude = location.value( QStringLiteral( "lat" ) ).toDouble();
+  const double longitude = location.value( QStringLiteral( "lng" ) ).toDouble();
+
+  const QgsGeometry geom = QgsGeometry::fromPointXY( QgsPointXY( longitude, latitude ) );
+
+  QgsGeocoderResult res( json.value( QStringLiteral( "formatted_address" ) ).toString(),
```
3nids, line 211: could we have a way to be a bit finer in the results and have a group and description properties as string? If given they would be used in the locator filter result.

3nids (approved):
Nice!

src/core/geocoding/qgsgooglemapsgeocoder.h
nyalldawson: Done

src/core/geocoding/qgsgooglemapsgeocoder.cpp
nyalldawson: I've added group, and the api for description. There's nothing returned by the google maps geocoder which lends itself to the description field though (rather that applies more to the Google places api, which could be a potential other geocoder class!)

3nids: perfect, thanks a lot!

PR #39751, opened 2020-11-02
Vertex editor: make displayed coordinate precision configurable

mhugent:
In the vertex editor, coordinates are displayed with 4 decimal places after the dot. This is a problem e.g. with lat/lon data. This PR sets the decimal places to the value configured in the project.

### DIFF
--- a/src/app/vertextool/qgsvertexeditor.cpp
+++ b/src/app/vertextool/qgsvertexeditor.cpp
@@ -477,7 +477,8 @@ CoordinateItemDelegate::CoordinateItemDelegate( QObject *parent )
 
 QString CoordinateItemDelegate::displayText( const QVariant &value, const QLocale &locale ) const
 {
-  return locale.toString( value.toDouble(), 'f', 4 );
+  int dp = QgsProject::instance()->readNumEntry( QStringLiteral( "PositionPrecision" ), QStringLiteral( "/DecimalPlaces" ), 4 );
+  return locale.toString( value.toDouble(), 'f', dp );
 }
 
 QWidget *CoordinateItemDelegate::createEditor( QWidget *parent, const QStyleOptionViewItem &, const QModelIndex &index ) const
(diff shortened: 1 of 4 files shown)


src/app/vertextool/qgsvertexeditor.cpp
```diff
@@ -477,7 +477,8 @@ CoordinateItemDelegate::CoordinateItemDelegate( QObject *parent )
 
 QString CoordinateItemDelegate::displayText( const QVariant &value, const QLocale &locale ) const
 {
-  return locale.toString( value.toDouble(), 'f', 4 );
+  int dp = QgsProject::instance()->readNumEntry( QStringLiteral( "PositionPrecision" ), QStringLiteral( "/DecimalPlaces" ), 4 );
```
nyalldawson, line 480: instead of directly accessing the setting, I think calling QgsCoordinateUtils::calculateCoordinatePrecision would be more appropriate. Is that an option?

mhugent: Yes, it  is an option, selection of precision is also possible in 'manual' mode. For 'automatic' mode, I'm a bit concerned that users who don't know about this setting might see unexpected results. E.g. they zoom out and only see meter precision (in a meter system) so they probably will think it is a bug.

Maybe something like this?:
manual mode: precision is taken from the setting
automatic mode: hardcode precision depending on layer units (3 decimal places for meter, feet, etc and 8 decimal places for degrees)

That said, I don't have a very strong opinion about it. If you still think that QgsCoordinateUtils::calculateCoordinate is better, I'll call that method.

nyalldawson: @mhugent 
> For 'automatic' mode, I'm a bit concerned that users who don't know about this setting might see unexpected results. E.g. they zoom out and only see meter precision (in a meter system) so they probably will think it is a bug.

Good point -- I agree with your concern

> manual mode: precision is taken from the setting
automatic mode: hardcode precision depending on layer units (3 decimal places for meter, feet, etc and 8 decimal places for degrees)

+1 

> That said, I don't have a very strong opinion about it. If you still think that QgsCoordinateUtils::calculateCoordinate is better, I'll call that method.

I can't see a clean way to integrate your proposed logic into the existing function. What do you think creating `QgsCoordinateUtils::calculateCoordinatePrecisionForCrs( const QgsCoordinateReferenceSystem& crs, QgsProject* project = nullptr )` and putting your new logic in there? I suspect there's elsewhere we should apply this same kind of logic to outside of the node tool (e.g. the offset/buffer distance spin box for the interactive offset tool comes to mind).

mhugent: That's a good suggestion. I'm going to update the PR into that direction.

agiudiceandrea:
The issue fixed by this PR still affects QGIS 3.16. See https://github.com/qgis/QGIS/issues/44931.

I think it could be useful and safe to backport this PR to the 3.16 branch.

agiudiceandrea:
Manual backported to release-3_16 https://github.com/qgis/QGIS/pull/44961

PR #39760, opened 2020-11-02
[point clouds] [3D] Eye dome lighting

NEDJIMAbelgacem:
## Description
This adds support for eye dome lighting, eye dome lighting works as a post processing effect and compares each pixel's depth (distance off the camera) with neighbor pixels depth and according to that depth difference edges are highlighted.
The UI for EDL configuration is added to the advanced settings tab of the 3D configuration dialog
Imgur
Eye dome lighting strength has the effect of increasing the contrast, allowing for better depth perception (needs to be adjusted when zooming in).
Eye dome lighting distance is the distance of the used pixels off the center pixel and it has the effect of making edges thicker.
The eye dome lighting is by principle an image based rendering simular to SSAO (screen space ambiant occlusion). Therefore it works on the whole scene and doesn't differ between point clouds or regular 3D objects (that's why the edges are thick and hollow in the screenshot).
Obligatory screenshot of point clouds :)
Imgur


src/3d/qgs3dmapsettings.h
```diff
@@ -362,6 +362,21 @@ class _3D_EXPORT Qgs3DMapSettings : public QObject, public QgsTemporalRangeObjec
     //! Returns whether to display labels on terrain tiles
     bool showLabels() const { return mShowLabels; }
 
+    //! Sets whether eye dome lighting will be used
```
nyalldawson, line 365:
```suggestion
    /**
     * Sets whether eye dome lighting will be used
     * \see eyeDomeLightingEnabled()
     * \since QGIS 3.18
     */
```
etc
(i.e. add the missing \since annotations and link setter to getter with a \see tag)

src/3d/qgs3dmapsettings.h
```diff
@@ -595,6 +628,10 @@ class _3D_EXPORT Qgs3DMapSettings : public QObject, public QgsTemporalRangeObjec
     bool mIsSkyboxEnabled = false;  //!< Whether the skybox is enabled
     QgsSkyboxSettings mSkyboxSettings; //!< Skybox related configuration
     QgsShadowSettings mShadowSettings; //!< Shadow rendering related settings
+
+    bool mEyeDomeLightingEnabled = false;
+    double mEyeDomeLightingStrength = 1000.0f;
```
nyalldawson, line 633:
```suggestion
    double mEyeDomeLightingStrength = 1000.0;
```

src/3d/qgsshadowrenderingframegraph.h
```diff
@@ -153,6 +155,10 @@ class QgsShadowRenderingFrameGraph
     float mShadowBias = 0.00001f;
     int mShadowMapResolution = 2048;
 
+    bool mEyeDomeLightingEnabled = false;
+    double mEyeDomeLightingStrength = 1000.0f;
```
nyalldawson, line 159:
```suggestion
    double mEyeDomeLightingStrength = 1000.0;
```

src/3d/qgsshadowrenderingframegraph.h
```diff
@@ -116,6 +116,8 @@ class QgsShadowRenderingFrameGraph
     void addTexturePreviewOverlay( Qt3DRender::QTexture2D *texture, const QPointF &centerNDC, const QSizeF &size, QVector<Qt3DRender::QParameter *> additionalShaderParameters = QVector<Qt3DRender::QParameter *>() );
     //! Sets shadow rendering to use a directional light
     void setupDirectionalLight( const QgsDirectionalLightSettings &light, float maximumShadowRenderingDistance );
+    //! Sets eye dome lighting shading related settings
+    void setupEyeDomeLighting( bool enabled, double strength, double distance );
```
nyalldawson, line 120: Is distance a double or int? In Qgs3DMapSettings you've used ints.

nyalldawson (commented):
Nice -- code looks good to me!

Can you add tests too? (I can if you're not confident with this)

NEDJIMAbelgacem:
@nyalldawson To make the tests for the eye dome lighting work I need to change somethings in the offscreen renderer so that it uses the post processing effect. I will make the tests but I want to add them in a different PR that will change the offscreen renderer.

nyalldawson:
@NEDJIMAbelgacem ok, makes sense!

This looks good to merge. @wonder-sk were you wanting to review also?

NEDJIMAbelgacem:
> This looks good to merge. @wonder-sk were you wanting to review also?

There is some weird bug where the EDL shading works for me but didn't work for Martin. We tried investigating the issue and it appears that the depth buffer is not filled properly. I will add a way to visualize these buffers and try debuging the problem. The good news out of this is that the bug might be the cause of the shadow rendering problem from before.
I hope we can fix it soon

nyalldawson:
Let me build and test here...

nyalldawson:
For me it kind of works... I see it nicely at some positions, then moving the camera slightly results in it disappearing:


Peek 2020-11-04 11-24

NEDJIMAbelgacem:
> For me it kind of works... I see it nicely at some positions, then moving the camera slightly results in it disappearing:

Yes, that's actually normal since the depth information depends on the position of the camera so if you zoom in and out you need to adjust the strength value. I need to look and see if there is some magic formula to adjust the strength value dynamically but I'm not sure if there is one.

NEDJIMAbelgacem:
One more thing, can you try using shadows with point clouds? I want to see if it's working for you. Here is what it looks like for me

nyalldawson:
> Yes, that's actually normal since the depth information depends on the position of the camera so if you zoom in and out you need to adjust the strength value. 

Hmm - I don't think it's clear in the video, but I'm panning across the scene and the effect is cutting in and out. It's not a constant relationship to the zoom.

Shadows seem to work ok:

image

vs

image

nyalldawson:
Wow, the shadow + point cloud combination is great, it really helps with visibliity! I don't think i've seen this supported elsewhere...

wonder-sk:
> @wonder-sk were you wanting to review also?

The code looks good to me, but as Belgacem said, it does not work on my laptop at all at this point :cry: - hopefully the addition of debug views will help to find the problem (also for development of more post-processing effects in the future).

PeterPetrik:
@wonder-sk so we wait for fix for merging?

wonder-sk:
Thanks for addition of the debug helpers @NEDJIMAbelgacem 

After a couple of hours of debugging, these two lines are causing all the issues for me:
```
mForwardDepthTexture->setComparisonFunction( Qt3DRender::QTexture2D::ComparisonFunction::CompareLessEqual );
mForwardDepthTexture->setComparisonMode( Qt3DRender::QTexture2D::ComparisonMode::CompareRefToTexture );
```
When I remove these two lines, things start to work correctly, even casting of shadows from a directional light.

I don't understand why we should have comparison to a reference texture enabled - this seems wrong. It's funny though that my opengl drivers behave differently than from other people's drivers (I am using official nvidia drivers on linux).

@NEDJIMAbelgacem could you please check if things still work correctly for you if you remove those two lines? If yes, then great and I am happy to get all this merged

NEDJIMAbelgacem:
Thanks a lot for figuering out where the problem is @wonder-sk ! 

> I don't understand why we should have comparison to a reference texture enabled - this seems wrong. It's funny though that my opengl drivers behave differently than from other people's drivers (I am using official nvidia drivers on linux).

I think I was trying to make the shadows better by setting polygon offset or something and forgot to remove the 2 lines. I will keep in mind that this can cause problems in the future.

> @NEDJIMAbelgacem could you please check if things still work correctly for you if you remove those two lines? If yes, then great and I am happy to get all this merged

Yes, it works correctly for me. Again, thanks a lot for your help!!

src/app/3d/qgs3dmapconfigwidget.cpp
```diff
@@ -161,6 +161,31 @@ Qgs3DMapConfigWidget::Qgs3DMapConfigWidget( Qgs3DMapSettings *map, QgsMapCanvas
...
+
+  mDebugShadowMapCornerComboBox->addItem( QStringLiteral( "Top Right" ) );
+  mDebugShadowMapCornerComboBox->addItem( QStringLiteral( "Top Left" ) );
+  mDebugShadowMapCornerComboBox->addItem( QStringLiteral( "Bottom Left" ) );
+  mDebugShadowMapCornerComboBox->addItem( QStringLiteral( "Bottom Right" ) );
+
+  mDebugDepthMapCornerComboBox->addItem( QStringLiteral( "Top Right" ) );
+  mDebugDepthMapCornerComboBox->addItem( QStringLiteral( "Top Left" ) );
+  mDebugDepthMapCornerComboBox->addItem( QStringLiteral( "Bottom Left" ) );
+  mDebugDepthMapCornerComboBox->addItem( QStringLiteral( "Bottom Right" ) );
```
wonder-sk, line 177: all these strings should be translatable - i.e. wrapped in tr()

also best to avoid code duplication...

src/app/3d/qgs3dmapconfigwidget.cpp
```diff
@@ -161,6 +161,31 @@ Qgs3DMapConfigWidget::Qgs3DMapConfigWidget( Qgs3DMapSettings *map, QgsMapCanvas
...
+
+//  mDebugShadowMapCornerComboBox->setCurrentText( QStringLiteral( "Top Left" ) );
+//  mDebugDepthMapCornerComboBox->setCurrentText( QStringLiteral( "Top Right" ) );
```
wonder-sk, line 180: remove commented code?

src/3d/qgs3dmapsettings.h
```diff
@@ -595,6 +678,18 @@ class _3D_EXPORT Qgs3DMapSettings : public QObject, public QgsTemporalRangeObjec
     bool mIsSkyboxEnabled = false;  //!< Whether the skybox is enabled
     QgsSkyboxSettings mSkyboxSettings; //!< Skybox related configuration
     QgsShadowSettings mShadowSettings; //!< Shadow rendering related settings
+
+    bool mEyeDomeLightingEnabled = false;
+    double mEyeDomeLightingStrength = 1000.0;
+    int mEyeDomeLightingDistance = 1;
+
+    bool mDebugShadowMapEnabled = false;
+    QString mDebugShadowMapCorner = "Top Left";
```
wonder-sk, line 687: Rather than setting and returning the position as a string, it would be better to have an enumeration with the list of recognized positions - more robust to errors...

src/3d/qgsshadowrenderingframegraph.cpp
```diff
@@ -185,20 +183,28 @@ QgsShadowRenderingFrameGraph::QgsShadowRenderingFrameGraph( QWindow *window, Qt3
...
   // textures preview pass
   Qt3DRender::QFrameGraphNode *previewPass = constructTexturesPreviewPass();
   previewPass->setParent( mMainViewPort );
+
+
+  mDebugDepthMapPreviewQuad = this->addTexturePreviewOverlay( mForwardDepthTexture, QPointF( 0.8f, 0.8f ), QSizeF( 0.2f, 0.2f ) );
+  mDebugShadowMapPreviewQuad = this->addTexturePreviewOverlay( mShadowMapTexture, QPointF( -0.8f, -0.8f ), QSizeF( 0.2f, 0.2f ) );
+  mDebugDepthMapPreviewQuad->setEnabled( false );
+  mDebugShadowMapPreviewQuad->setEnabled( false );
+//  this->addTexturePreviewOverlay( mForwardDepthTexture, QPointF( 0.8f, 0.8f ), QSizeF( 0.2f, 0.2f ) );
```
wonder-sk, line 198: remove commented code?

src/3d/qgs3dmapsettings.cpp
```diff
@@ -237,6 +237,20 @@ void Qgs3DMapSettings::readXml( const QDomElement &elem, const QgsReadWriteConte
...
   mShadowSettings.readXml( elemShadows, context );
 
+  QDomElement elemEyeDomeLighting = elem.firstChildElement( QStringLiteral( "eye-dome-lighting" ) );
+  mEyeDomeLightingEnabled = elemEyeDomeLighting.attribute( "enabled", QStringLiteral( "0" ) ).toInt();
+  mEyeDomeLightingStrength = elemEyeDomeLighting.attribute( "eye-dome-lighting-strength", QStringLiteral( "1000.0" ) ).toDouble();
+  mEyeDomeLightingDistance = elemEyeDomeLighting.attribute( "eye-dome-lighting-distance", QStringLiteral( "1" ) ).toInt();
+
+  QDomElement elemDebugSettings = elem.firstChildElement( QStringLiteral( "debug-settings" ) );
+  mDebugShadowMapEnabled = elemDebugSettings.attribute( QStringLiteral( "shadowmap-enabled" ), QStringLiteral( "0" ) ).toInt();
+  mDebugShadowMapCorner = elemDebugSettings.attribute( QStringLiteral( "shadowmap-corner" ), QStringLiteral( "Top Left" ) );
```
wonder-sk, line 247: In XML best to either save the corner as the enumeration value typed to int - or have a function to translate between strings e.g. "top-left" and enumeration values. (These strings would not be translatable)

src/3d/qgs3dmapsettings.h
nyalldawson: There's a Qt enum we can reuse here -- Qt::Corner

NEDJIMAbelgacem: > There's a Qt enum we can reuse here -- Qt::Corner

Yes, I'm using that :)

nyalldawson:
@NEDJIMAbelgacem 
I've just built your latest branch for testing. What's the "Debug depth map" and shadow map options supposed to look like? 

If I enable debug depth map, this is what I see:

image

The position and size option has no effect.  Similar results with "Debug shadow map", except that it's completely red.

I still see an issue with the eye dom effect vanishing as I move towards the edges of the cloud. It looks like its triggered by some condition like when the camera center falls outside of the point cloud bounds (or possibly when the camera center moves off the underlying flat terrain).

Here's a screencast showing this happening as I pan towards the edge of the cloud

Peek 2020-11-06 11-16

wonder-sk:
Hi @nyalldawson 

your depth map looks fine - it is supposed to be (at this point) red-black image with nearer pixels being darker. It should show up in a corner though, not overlaid on top of the whole 3D view. Shadow map should be a similar depth map, but rendered from the point of view of the light used for shadows. I guess if it is all red that's fine if you have not enabled shadows.

Regarding the vanishing EDL shading:
- I can also replicate slowly vanishing shading as I move towards the "end" of the map (but not a sudden change). A quick look suggests that this has to do with the way how near and far planes of the 3D view frustum (https://en.wikipedia.org/wiki/Viewing_frustum) are configured in Qgs3DMapScene::updateCameraNearFarPlanes()
- another issue is that our EDL shading is not invariant from the setup of near/far planes. In the shader we work with "raw" depth buffer values which are in range [0..1] but those value are not growing linearly with increasing depth: Objects close to the near plane get much more depth range than those close to the far plane, and so the effect depends on the distance from camera. Probably we could linearize the depth values to make the shading more robust.
- the sudden change will be indeed caused by a chunk of the terrain underneath getting culled and thus affecting near/far camera plane calculation

I would suggest that we merge the the changes as they are and then continue fixing things in follow up PRs...

nyalldawson:
> It should show up in a corner though, not overlaid on top of the whole 3D view

Ok, I'll look into what's happening here

> I would suggest that we merge the the changes as they are and then continue fixing things in follow up PRs...

Fine by me!

PR #39763, opened 2020-11-02
suggest qt5-image-formats-plugins for server

mbernasocchi:
## Description
This allows using webp in the server, see https://github.com/gem/oq-qgis-server/pull/45 for more details, @jef-n I'm not sure at all if this is how it is done, and if it should recommend or suggest, feel free to kill this PR if it is wrong/unwanted.
It would also need to be backported to 3.16


mbernasocchi:
thanks @jef-n

PR #39766, opened 2020-11-02
Speedup label placement when building against GEOS-3.9.0

strk:
Uses GEOSPreparedNearestPoints signature
See https://trac.osgeo.org/geos/ticket/1007


nyalldawson:
LGTM! Is this safe to backport in your opinion? (ie is the GEOS indexed distance code sufficiently reliable?)

Out of curiosity, do you have any benchtests of the speed bump?

strk:
I've not performed performance tests, was counting on you, @nyalldawson , as the reporter of https://trac.osgeo.org/geos/ticket/1007

nyalldawson:
@strk 
Unfortunately I don't have a QGIS build setup against GEOS master to test with at the moment.

Anyway, changes look good to me!

nyalldawson (approved):

PR #39769, opened 2020-11-03
fix geometry validator

vcloarec:
fix QgsGeometryVaidatot that can make Qgis crash when manipulating vertices

### DIFF
--- a/src/core/qgsgeometryvalidator.cpp
+++ b/src/core/qgsgeometryvalidator.cpp
@@ -65,8 +65,8 @@ void QgsGeometryValidator::checkRingIntersections( int p0, int i0, const QgsLine
         {
           d = -distLine2Point( ring1XAtj, ring1YAtj, w.perpVector(), sX, sY );
           if ( d > 0 && d < w.length() &&
-               ring0->pointN( i + 1 ) != ring1->pointN( j + 1 ) && ring1->pointN( i + 1 ) != ring1->pointN( j ) &&
-               ring1->pointN( i + 0 ) != ring1->pointN( j + 1 ) && ring1->pointN( i + 0 ) != ring1->pointN( j ) )
+               ring0->pointN( i + 1 ) != ring1->pointN( j + 1 ) && ring0->pointN( i + 1 ) != ring1->pointN( j ) &&
+               ring0->pointN( i + 0 ) != ring1->pointN( j + 1 ) && ring0->pointN( i + 0 ) != ring1->pointN( j ) )
           {
             QString msg = QObject::tr( "segment %1 of ring %2 of polygon %3 intersects segment %4 of ring %5 of polygon %6 at %7, %8" )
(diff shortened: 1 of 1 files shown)


nyalldawson:
Ouch -- annoying that this could compile :( 

Can you add a test which triggers this? QgsGeometryValidator unfortunately has very poor test coverage at the moment

vcloarec:
I found this one during own GIS work, and I do not have much time for now to add test...

src/core/qgsgeometryvalidator.cpp
```diff
@@ -65,8 +65,8 @@ void QgsGeometryValidator::checkRingIntersections( int p0, int i0, const QgsLine
         {
           d = -distLine2Point( ring1XAtj, ring1YAtj, w.perpVector(), sX, sY );
           if ( d > 0 && d < w.length() &&
-               ring0[i + 1] != ring1[j + 1] && ring0[i + 1] != ring1[j] &&
-               ring0[i + 0] != ring1[j + 1] && ring0[i + 0] != ring1[j] )
+               ring0->pointN( i + 1 ) != ring1->pointN( j + 1 ) && ring1->pointN( i + 1 ) != ring1->pointN( j ) &&
```
uclaros, line 68: Seems like you've mixed the rings. Ring0's counter is i and ring1's counter is j

vcloarec: oups... you are right, it is a mess!

vcloarec: fixed

src/core/qgsgeometryvalidator.cpp
```diff
@@ -65,8 +65,8 @@ void QgsGeometryValidator::checkRingIntersections( int p0, int i0, const QgsLine
         {
           d = -distLine2Point( ring1XAtj, ring1YAtj, w.perpVector(), sX, sY );
           if ( d > 0 && d < w.length() &&
-               ring0[i + 1] != ring1[j + 1] && ring0[i + 1] != ring1[j] &&
-               ring0[i + 0] != ring1[j + 1] && ring0[i + 0] != ring1[j] )
+               ring0->pointN( i + 1 ) != ring1->pointN( j + 1 ) && ring0->pointN( i + 1 ) != ring1->pointN( j ) &&
+               ring0->pointN( i + 0 ) != ring1->pointN( j + 1 ) && ring0->pointN( i + 0 ) != ring1->pointN( j ) )
           {
             QString msg = QObject::tr( "segment %1 of ring %2 of polygon %3 intersects segment %4 of ring %5 of polygon %6 at %7, %8" )
                           .arg( i0 ).arg( i ).arg( p0 )
```
uclaros, line 72: Seems like those args are mixed too. i is the vertex index and i0 should be the ring index

nyalldawson:
I've added a test in https://github.com/qgis/QGIS/pull/39790 (doing my penance)

Issue #36037, opened 2020-04-27
Wrong size of legend symbols when using Geometry generator symbology


VxTedxV:
I was not able to reopen a closed ticket : #22638

This issue remains in QGis 3.12.1 !

Tested with line geometry : in legend, the line is half screen long before we can read the label !

gioman:
Duplicate of https://github.com/qgis/QGIS/issues/33755

Issue #33755, opened 2020-01-13
Wrong size of legend symbols when using Geometry generator symbology


gwenandres:
Hello, 
I get really big symbols in the legend on a point layer with surface or line geometry generator when I choose categorized symbols.

For example : 
- I create a new point memory layer 
- I add a column "type"  (integer)
- For style, I choose categorized symbols on "type" column
- For symbols, I choose surface geometry generator (`buffer ($ geometry, 2)`)
- Then in legend I get really big symbols for each type.

I am working with QGIS 3.4.14 LTR on Windows 7 or 10.
I try with QGIS 3.10.1, I have the same behaviour.
![Qgis_issue_categorized_symbols](https://user-images.githubusercontent.com/15909387/72250057-0c70bd80-35fb-11ea-9be8-b7ab989b59ff.png)

gioman:
@gwenandres your layer has a CRS that has degrees as units?

gwenandres:
The CRS is 2154 so it has meters as units.

gioman:
@gwenandres Can't confirm here on 3.10.1 on Ubuntu 18.04 but I'm not totally sure I followed the exact steps that lead you to see this problem.

![Screenshot_20200116_134557](https://user-images.githubusercontent.com/1951107/72530155-a7b99b00-3866-11ea-81f7-3689a1587ce8.png)

gwenandres:
@gioman  Sorry if I am not clear. I have the problem on a point layer, not on a surface layer. 
![QGIS3-pointlayer](https://user-images.githubusercontent.com/15909387/72532397-a2168200-3873-11ea-9f9a-f32a62183c96.png)

Yes it can be strange but I have point layer with surface generator.

gioman:
> I have the problem on a point layer

@gwenandres oh I see... moreover not only the legend symbol does not make sense, but also the generated area on map is not visible at all.

SrNetoChan:
I was able to replicate this. It only happens on Rule-based (maybe categorized too) Symbol, but not in single symbol, where it works just fine.

jonnyforestGIS:
@SrNetoChan I confirm as well this behaviour when using the geometry generator and happens with categorized symbol. Cheers

jonnyforestGIS:
@gioman having this #39777 as a possible resolution for this issue, this ticket can be close?

gioman:
> @gioman having this #39777 as a possible resolution for this issue, this ticket can be close?

[...]

PR #39777, opened 2020-11-03
[layer tree][ui] Allow users to set a minimum/maximum legend size

nirvn:
## Description

This PR fixes a long standing UI/UX issue with the layer tree panel whereas legend icons for symbols using map units can end up taking an undesirably large amount of space in the tree widget.

The PR introduces a new pair of settings, a minimum legend size value and a maximum legend size value. Those are tweaked in the settings dialog's canvas and legend panel:
image

By default, we now max out symbol sized above 32mm by default. Users can set this to 0 to allow symbols to grow infinitely as it was before. But really, why? :wink: 

Here's an example, with symbols clipped to 25 mm:
image

Funded by SO!GIS
CC @andreasneumann

### DIFF
--- a/src/ui/qgsoptionsbase.ui
+++ b/src/ui/qgsoptionsbase.ui
@@ -3343,7 +3343,7 @@
                     <item>
                      <widget class="QLabel" name="label_58">
                       <property name="text">
-                       <string>WMS getLegendGraphic Resolution</string>
+                       <string>WMS getLegendGraphic resolution</string>
                       </property>
                      </widget>
                     </item>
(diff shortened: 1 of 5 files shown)


jonnyforestGIS:
@nirvn this is related with this issue #33755 . Cheers

roya0045:
Nice work Mathieu!

troopa81:
Another possibly related ticket #36037. Will it fix this issue?

src/ui/qgsoptionsbase.ui
```diff
@@ -3377,6 +3377,125 @@
...
+                       <double>0.200000000000000</double>
+                      </property>
+                      <property name="value">
+                       <double>0.10</double>
+                      </property>
+                      <property name="showClearButton" stdset="0">
+                       <bool>true</bool>
+                      </property>
+                      <property name="toolTip">
+                       <string extracomment="(set value to 0 to skip minimum size)"/>
```
nyalldawson, line 3439: You could make this clearer by setting the clear value text to "Not set"

nyalldawson (commented):
Nice fix! Can we get some tests here too please?

nirvn:
@nyalldawson , comments addressed, thanks for the review.

andreasneumann:
Will this be backportable to older versions? Specifically 3.16 and 3.10? I hope that at least 3.16 will be possible.

nirvn:
I'd be +1 to backport to 3.16. It's definitely fixing a long-standing UI issue we've had with the layer tree panel.

tests/src/core/testqgslayertree.cpp
```diff
@@ -291,6 +293,46 @@ void TestQgsLayerTree::testCheckStateMutuallyExclusiveEdgeCases()
   delete root3;
 }
 
+void TestQgsLayerTree::testRestrictedSymbolSize()
+{
+  QgsSettings settings;
```
nyalldawson, line 298: Can you add the QCoreApplication::setOrganizationName/QCoreApplication::setOrganizationDomain/QCoreApplication::setApplicationName calls to initTestCase in this class so that running the test doesn't mess up your local settings?

nyalldawson:
I'm +1 to backport to 3.16, given that 3.16 isn't the LTR yet and a certain amount of flux is still allowable in the release.

jonnyforestGIS:
Hi @nirvn , I tested this in the last master in OSGeo4W64 and I don't have change. Please correct if I'm doing something wrong.

QGIS Environment:


QGIS   version | 3.17.0-Master | QGIS code revision | ccdff23dbe
-- | -- | -- | --
Compiled   against Qt | 5.11.2 | Running against Qt | 5.11.2
Compiled   against GDAL/OGR | 3.3.0dev | Running against GDAL/OGR | 3.3.0dev
Compiled   against GEOS | 3.8.1-CAPI-1.13.3 | Running against GEOS | 3.8.1-CAPI-1.13.3
Compiled   against SQLite | 3.29.0 | Running against SQLite | 3.29.0
PostgreSQL   Client Version | 11.5 | SpatiaLite Version | 4.3.0
QWT   Version | 6.1.3 | QScintilla2 Version | 2.10.8
Compiled   against PROJ | 8.0.0 | Running against PROJ | Rel. 8.0.0, March 1st, 2021
OS Version | Windows 10 (10.0) | This copy of QGIS   writes debugging output.
Active   python plugins | DataPlotly;
processing_r;
Qgis2threejs;
quick_map_services;
MetaSearch;
processing

Example:
legend_size

DelazJ:
@jonnyforestGIS did you finally get it work? I've also tried, using geometry generator, single, categorized, rule-based, proportional and still a big symbol in the layers panel. Closed and reopened the project. Still no change. I'm on 3.16.2.
@nirvn is there a particular prerequisite/method to get this work? Thanks

SrNetoChan:
@nirvn is this supposed to work with geometry generator layers? I see the option on 3.16.3, but I still get huge symbols in the legend.

SrNetoChan:
Some more information, it works fine on single symbol renderer with a geometry generator, but not in categorized, graduated or rule-based using geometry generator symbols.

SrNetoChan:
Peek 2021-01-20 01-01

gioman:
> Some more information, it works fine on single symbol renderer with a geometry generator, but not in categorized, graduated or rule-based using geometry generator symbols.

@SrNetoChan better file a separate ticket.

SrNetoChan:
@gioman  there's already #33755

roya0045:
Is this an issue with multi-symbols and the geometry generator or multi-symbols in general?

SrNetoChan:
Just when you use geometry generators. Haven't tested with proportional symbols in multi symbols

PR #39778, opened 2020-11-03
Bugfix gh39757 shared locale sort

### LINKED ISSUE #39757: Raster Legend not sorting correct when edited due to decimal separation
Version QGis 3.16.0

bug not in QGIS 3.14.15

Language settings: Dutch
Country settings: Dutch

I have a style saved in QGis 3.14.15. When using this style in QGis 3.16 it works good.
But changing the value with a decimal precision the legend items moves to the bottem.

Closing and opening the legend settings, shows the correct order again.

See attachments for explanation.
Step 1
Knipsel 1
Step 2
Knipsel2
Step 3
Knipsel3

elpaso:
Try to fix #39757

### DIFF
--- a/python/core/auto_additions/qgis.py
+++ b/python/core/auto_additions/qgis.py
@@ -1,2 +1,3 @@
 # The following has been generated automatically from src/core/qgis.h
+Qgis.DataType.baseClass = Qgis
 Qgis.PythonMacroMode.baseClass = Qgis
--- a/python/core/auto_generated/raster/qgsrastershaderfunction.sip.in
+++ b/python/core/auto_generated/raster/qgsrastershaderfunction.sip.in
@@ -115,7 +115,23 @@ Returns the minimum value for the raster shader.
 Returns legend symbology items if provided by renderer.
 %End
 
+    int labelPrecision() const;
+%Docstring
+Returns label precision
+
+.. versionadded:: 3.16
+%End
+
+    void setLabelPrecision( int labelPrecision );
+%Docstring
+Sets label precision to ``labelPrecision``
+
+.. versionadded:: 3.16
+%End
+
   protected:
+
+
 
 
 };
--- a/python/gui/auto_generated/qgsdoublevalidator.sip.in
+++ b/python/gui/auto_generated/qgsdoublevalidator.sip.in
@@ -55,10 +55,19 @@ Constructor for QgsDoubleValidator.
 
 :param bottom: the minimal range limit accepted by the validator
 :param top: the maximal range limit accepted by the validator
-:param decimal: the number of decimal accepted by the validator
+:param decimal: the number of decimals accepted by the validator
 :param parent: parent object
 %End
 
+    QgsDoubleValidator( int decimal, QObject *parent );
+%Docstring
+Constructor for QgsDoubleValidator.
+
+:param decimal: the number of decimals accepted by the validator
+:param parent: parent object
+
+.. versionadded:: 3.16
+%End
 
 
     QValidator::State validate( QString &input ) const;
--- a/python/gui/auto_generated/raster/qgscolorrampshaderwidget.sip.in
+++ b/python/gui/auto_generated/raster/qgscolorrampshaderwidget.sip.in
@@ -10,6 +10,9 @@
 
 
 
+
+
+
 class QgsColorRampShaderWidget: QWidget, protected Ui::QgsColorRampShaderWidgetBase
 {
 %Docstring
@@ -81,6 +84,8 @@ Returns shared function used in the renderer
 %Docstring
 Sets widget state from the color ramp shader
 %End
+
+    static int VALUE_ROLE;
 
   signals:
     void minimumMaximumChangedFromTree( double minimum, double maximum );
--- a/python/gui/auto_generated/raster/qgscolorrampshaderwidget.sip.in
+++ b/python/gui/auto_generated/raster/qgscolorrampshaderwidget.sip.in
@@ -11,8 +11,6 @@
 
 
 
-
-
 class QgsColorRampShaderWidget: QWidget, protected Ui::QgsColorRampShaderWidgetBase
 {
 %Docstring
@@ -84,8 +82,6 @@ Returns shared function used in the renderer
 %Docstring
 Sets widget state from the color ramp shader
 %End
-
-    static int VALUE_ROLE;
 
   signals:
     void minimumMaximumChangedFromTree( double minimum, double maximum );
--- a/python/gui/auto_generated/raster/qgscolorrampshaderwidget.sip.in
+++ b/python/gui/auto_generated/raster/qgscolorrampshaderwidget.sip.in
@@ -112,6 +112,7 @@ Loads min and max values from color ramp tree
 %End
 
   protected:
+
     void populateColormapTreeWidget( const QList<QgsColorRampShader::ColorRampItem> &colorRampItems );
 %Docstring
 Populates color ramp tree from ramp items
--- a/src/core/qgis.h
+++ b/src/core/qgis.h
@@ -115,6 +115,7 @@ class CORE_EXPORT Qgis
       ARGB32 = 12, //!< Color, alpha, red, green, blue, 4 bytes the same as QImage::Format_ARGB32
       ARGB32_Premultiplied = 13 //!< Color, alpha, red, green, blue, 4 bytes  the same as QImage::Format_ARGB32_Premultiplied
     };
+    Q_ENUM( DataType )
 
     /**
      * Authorisation to run Python Macros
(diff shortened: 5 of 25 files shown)


elpaso:
@PedroVenancio and @jkr-wrk the binary will be available shortly at https://github.com/qgis/QGIS/actions/runs/343473956

### What it's in the PR:

- fix localized input: accepts your locale decimal separator first and fallback on `.` if that doesn't produce a valid number
- for the values, try to show the maximum number of decimals allowed by the data type (float, double, long etc.)
- fix min/max wrong values when discrete interpolation
- fix sorting of classes after editing

### What to test:
- different data types (int, long, float, double)
- different interpolations
- editing of values and check if sorting of classes don't break

Happy testing!

PedroVenancio:
Hi @elpaso 

I still see those issues here:

 - with float32 and 64 (ok with int16);
imagem

 - editing the values break the value itself, and I can only enter an integer number. If I enter a decimal, it rounds it to the next integer: 
qgis_editing_values_2

Here is the sample project I'm testing: https://cld.pt/dl/download/61894ebd-ade0-4f12-8aa3-1405ab2690cc/Float_Issue_Project.zip

elpaso:
https://github.com/qgis/QGIS/issues/39757 is not about paletted but singleband pseudocolor. I don't think this patch applies to paletted (not yet at least).

nyalldawson:
Looking good to me!

elpaso:
@nyalldawson thank you for the review, I will add some more bugfixes in this PR soon, I want to address the float/double paletted too which still has some legacy int cast that breaks the classes.

I'm thinking that the custom delegate `QgsLocaleAwareNumericLineEditDelegate` should be used in the paletted model too, so that we have consistency in how the values are shown and edited and validated, I will use the `QgsDoubleValidator` for that.

PedroVenancio:
> #39757 is not about paletted but singleband pseudocolor. I don't think this patch applies to paletted (not yet at least).

Hi @elpaso 

Sorry, I was testing Paletted.

Now looking at singleband pseudocolor, I'm still seeing two issues:
- It does not change the labels according to the value;
- The ordering only happens after do Ok and then open Properties again.

Here is the old behaviour, and the new:

qgis_single_band_pseudo

Thanks @elpaso

elpaso:
@PedroVenancio the last commit should fix most of the remaining issues.

For the paletted/unique, there is no auto-update of the labels if the value is changed, it was never implemented, for singleband there is logic for that.

Can you please test both paletted and singleband with different data types?

elpaso:
@PeterPetrik @vcloarec I don't know if you are the right guys to ask this but maybe you can have a look if this has an impact on meshes symbology? In particular, the new widget tree delegate assumes that the models store the `double` value in `DisplayRole` and it takes care of proper text formatting according to the underlying provider data type (`Qgis::DataType`) I don't know enough about meshes to know how this will have an impact on those.

PeterPetrik:
I think if test passes we should be fine. if you want to test, you can grab https://github.com/lutraconsulting/MDAL/blob/master/tests/data/grib/Madagascar.wave.7days.grb and see if color widget works. Thanks!

vcloarec:
as mesh use only `double` for scalar value, I think it will be ok.

elpaso:
@vcloarec @PeterPetrik I see no related issues with the test mesh. 

Unrelated to this PR: the precision spinbox and the prefix do not work out of the box unless you reclassify. The reason is that the default labels use limited precision and they are seen by the implementation as if they were changed by the user on purpose. So, the prefix and precision spinbox do not override user-edited labels. 

I'm not going to fix this though.

PedroVenancio:
Hi @elpaso 

- Paletted / Unique Values

1. The ordering is correct now after clicking 'Classify', but doing 'Ok' the ordering, for some reason, is lost.
2. The raster is float64 and does not show so much decimal places as in 'Value' column. For instance, pixel with 1.88 becomes 1,8799999999999999.

qgis_paletted_unique


- Singleband Pseudocolor - Linear

1. Now it works as expected until I do the 'Ok'. After I do the 'Ok', label does not change with precision (returns to the default 4) and does not change in sync with Value. Also, the label suffix appears and is shown in the last class.

qgis_singleband_pseudocolor_linear


- Singleband Pseudocolor - Discrete

1. Label of the first (lower) class does not change automatically, but change for instance after the change of label precision.
2. After the 'Ok' all becomes unresponsive in sync with value.
3. Label precision return to 4 and it is added a label suffix that is shown in the last class.

qgis_singleband_pseudocolor_discrete

elpaso:
@PedroVenancio let's discuss one by one:


> Hi @elpaso
> 
>     * Paletted / Unique Values
> 
> 
>     1. The ordering is correct now after clicking 'Classify', but doing 'Ok' the ordering, for some reason, is lost.

Fixed in 58ce77f

> 
>     2. The raster is float64 and does not show so much decimal places as in 'Value' column. For instance, pixel with 1.88 becomes 1,8799999999999999.
> 

Won't fix: the idea is that we show the real value with full precision in the value column, in case of `1.88` it is 1.87999999999999989341858963598E0  (see: http://www.binaryconvert.com/result_double.html?decimal=049046056056), but C++ implementation supports 15 digits so we get 1,8799999999999999

This is the value used internally to match the value in the renderer and I think we must show it unaltered with full precision (omitting trailing zeroes) because there is no way to know which precision is meaningful for the user's use case.

> 
>     * Singleband Pseudocolor - Linear
> 
> 
>     1. Now it works as expected until I do the 'Ok'. After I do the 'Ok', label does not change with precision (returns to the default 4) and does not change in sync with Value. Also, the label suffix appears and is shown in the last class.
> 
> 
>     * Singleband Pseudocolor - Discrete
> 
> 
>     1. Label of the first (lower) class does not change automatically, but change for instance after the change of label precision.
> 
>     2. After the 'Ok' all becomes unresponsive in sync with value.
> 
>     3. Label precision return to 4 and it is added a label suffix that is shown in the last class.
> 
> 

Please have a look to my latest commits and test it again, thank you for your help!

PedroVenancio:
Hi @elpaso 

> > * Paletted / Unique Values
> > 
> > 
> > 1. The ordering is correct now after clicking 'Classify', but doing 'Ok' the ordering, for some reason, is lost.
> 
> Fixed in 58ce77f

Paletted / Unique Values seems fixed now! Thank you very much!



> > * Singleband Pseudocolor - Linear
> > 

Here it seems everything fine, until a class value be manually changed. By doing that, the label precision becomes irresponsive and label unit suffix is automatically filled with a comma and a sequence of zeros corresponding to the precision choosed at that moment:

qgis_singleband_pseudocolor_linear_2



> > * Singleband Pseudocolor - Discrete
> >

With Discrete I see:

1. The lower value of each class label is always the same of minimum;

2. It seems that the color ramp is not applied in its full range;

3. Label precision becomes irresponsive but just when the manually changing of range values leads to a reordering; however, if the precision is changed before do the "ok", that does not happen and the label is updated.

Label precision field can be negative, maybe block it at zero should be better. 

qgis_singleband_pseudocolor_discrete_2


Again, thank you very much @elpaso !!

elpaso:
> Label precision field can be negative, maybe block it at zero should be better.

This is by design, to reduce precision, try it and check the tooltip for explanations.

I should have fixed the other issues, please have a look.

PedroVenancio:
Hi @elpaso ,

It's seems that almost everything was tackle down!

Now I've just this in Discrete:

1. Color ramp have a strange behaviour. After be changed, it seems that is not applied in its full range. However, if I update something (number of classes for instance), it is applied in it's full range again. Anyway, this does not seem a great concern at this moment.

2. Label still have some challenges. It still becomes unresponsive when the manually changing of range values leads to a reordering and do "Ok". The change of value does not update the label, as it happens when the change does not involve a reordering, and leads to the unresponsive behaviour after the "Ok".

qgis_singleband_pseudocolor_discrete_4

Thanks @elpaso !

elpaso:
@PedroVenancio I'll have a look, FYI when the label color is grey it means that they have the default text and when they are black it means that they have been edited by the user. In the second case the precision and suffix do not update the "edited" values automatically, that's why they become (using your words) "unresponsive".

elpaso:
@PedroVenancio I'm looking to the first issue, I'm not sure that's a bug: the logic behind is that any time there are already values in the list (like when you change ramp) the only thing that is changed is the color which is calculated from the new band. If you want them to be recalculated you need to press "Classify" or change the number of classes (which triggers a new classification). I'm hesitant to change that behavior because it seems it was done on purpose, even if the comment in the code are not very clear about that (in fact the comment suggests the opposite), @nirvn may know more about the expected behavior.

elpaso:
@PedroVenancio can you check if my last commit fixes the second issue?

PedroVenancio:
@elpaso It seems that everything is fixed now!!

Thank you very much for the hard work! Really impressive!

elpaso:
@PedroVenancio thank you for testing and for your high quality issue reporting, this bugfix wouldn't have been possible without that!

PedroVenancio:
Thank you very much @elpaso !! Testing is the easier part, fixing is the serious part!

PR #39780, opened 2020-11-03
[pointclouds] add basic structure for PDAL layer

PeterPetrik:
Add basic structure of PDAL layer see https://github.com/wonder-sk/point-cloud-experiments/issues/17

we read the CRS and extent from PDAL. nothing is drawn on any canvas.

Screenshot 2020-11-03 at 14 03 45


PeterPetrik:
not sure how to add exception to spell_check for "entwine"...

src/providers/pdal/qgspdaldataitems.cpp
```diff
@@ -0,0 +1,79 @@
...
+  QFileInfo info( path );
+  info.setFile( path );
+
+  // allow only normal files
+  if ( !info.isFile() )
+    return nullptr;
+
+  // Filter files by extension
+  // TODO get file filter list from PDAL library
+  if ( !path.endsWith( QStringLiteral( "laz" ) ) )
```
wonder-sk, line 73: - better to check suffix from `info`
- can we also add .las

src/providers/pdal/qgspdaldataitems.cpp
```diff
@@ -0,0 +1,79 @@
...
+    return nullptr;
+
+  QString name = info.baseName() + QStringLiteral( ".laz" );
```
wonder-sk, line 76: better to use fileName() ?

src/core/providers/ept/qgseptdataitems.cpp
```diff
@@ -29,6 +29,7 @@ QgsEptLayerItem::QgsEptLayerItem( QgsDataItem *parent,
   : QgsLayerItem( parent, name, path, uri, QgsLayerItem::PointCloud, QStringLiteral( "ept" ) )
 {
   mToolTip = uri;
+  mIconName = QStringLiteral( "mIconEntwineLayer.svg" );
```
wonder-sk, line 32: Not sure how others feel about icons, I would probably use a single icon for point cloud layers... and do not make difference whether it is EPT or PDAL - for users that's not really important...

src/providers/pdal/qgspdaldataitems.h
```diff
@@ -0,0 +1,49 @@
...
+ ***************************************************************************/
+
+#ifndef QGSPDALDATAITEMS_H
+#define QGSPDALDATAITEMS_H
+
+#include "qgsdataitem.h"
+#include "qgsdataitemprovider.h"
+#include "qgsprovidermetadata.h"
+
+class CORE_EXPORT QgsPdalLayerItem : public QgsLayerItem
```
wonder-sk, line 25: CORE_EXPORT in a provider?

src/providers/pdal/qgspdalsourceselect.cpp
```diff
@@ -0,0 +1,54 @@
...
+    QMessageBox::information( this,
+                              tr( "Add mesh layer" ),
+                              tr( "No layers selected." ) );
+    return;
+  }
+
+  for ( const QString &path : QgsFileWidget::splitFilePaths( mPath ) )
+  {
+    emit addMeshLayer( path, QFileInfo( path ).baseName(), QStringLiteral( "mdal" ) );
+  }
```
wonder-sk, line 53: mesh / mdal leftovers

wonder-sk (commented):
Looking great.

We will have to think a bit more about the GUI in data sources manager - my preference would be to have a single tab for point clouds whether they are coming from pdal or ept... But I guess that's easy to change when we have a plan.

src/core/providers/ept/qgseptdataitems.cpp
nyalldawson: Yeah, I agree with this. (Just like we don't use a different icon for gpkg vs shp)

src/providers/pdal/qgspdaldataitems.cpp
nyalldawson, line 73:
```suggestion
  if ( !path.endsWith( QLatin1String( "laz" ), Qt::CaseInsensitive ) )
```

nyalldawson: Alternatively -- if the pdal provider metadata implemented the filters() method (with a new FilterType of Point Cloud) then we could use QgsFileUtils::extensionsFromFilter here to automatically generate a list of supported extensions directly from the provider metadata.

src/providers/pdal/qgspdalprovider.cpp
```diff
@@ -0,0 +1,151 @@
...
+  pdal::Options las_opts;
+  las_opts.add( las_opt );
+  pdal::PointTable table;
+  pdal::LasReader las_reader;
+  las_reader.setOptions( las_opts );
+  las_reader.prepare( table );
+  pdal::PointViewSet point_view_set = las_reader.execute( table );
+  pdal::PointViewPtr point_view = *point_view_set.begin();
+  pdal::Dimension::IdList dims = point_view->dims();
+  pdal::LasHeader las_header = las_reader.header();
```
nyalldawson, line 89: Is this part relatively stable? (ie. not expected to change before release?) If so, I can add the code required to read and auto populate the layer metadata from this LasHeader object.

nyalldawson:
> Alternatively -- if the pdal provider metadata implemented the filters() method (with a new FilterType of Point Cloud) then we could use QgsFileUtils::extensionsFromFilter here to automatically generate a list of supported extensions directly from the provider metadata.

Hold off on this -- I'm working on some related improvements which I'll adapt for the pdal provider after merge

src/core/providers/ept/qgseptdataitems.cpp
nyalldawson: On this same note, I think we should only have a single Point Cloud source select widget. It's just clutter in the data source manager if we separate EPT/PDAL. I can fix this post merge though!

src/providers/pdal/qgspdalprovider.cpp
```diff
@@ -0,0 +1,151 @@
...
+bool QgsPdalProvider::load( const QString &uri )
+{
+  pdal::Option las_opt( "filename", uri.toStdString() );
+  pdal::Options las_opts;
+  las_opts.add( las_opt );
+  pdal::PointTable table;
+  pdal::LasReader las_reader;
+  las_reader.setOptions( las_opts );
+  las_reader.prepare( table );
+  pdal::PointViewSet point_view_set = las_reader.execute( table );
```
wonder-sk, line 86: Oh I think we need to avoid this call - it will read all points from the file and store them in memory - this can take many seconds. It should be possible to get the list of dimensions without actually reading everything. E.g. Entwine displays the list of dimensions before starting indexing.

src/providers/pdal/qgspdalprovider.cpp
PeterPetrik: not sure how @wonder-sk want to approach it. Do we support "ALL" PDAL supported formats? in that case do we need some QGIS abstraction to header/readers (this one is LasHeader -> only for laz/las files). 
Detto for the input source dialog. For EPT format it is easy. For PDAL-supported formats you may need to select (manually) driver in the dialog. You may want to select if you you want/do not want to automatically index it. Also you may want to add JSON file with the PDAL filters to let PDAL/EPT know how to read/index it... etc.

src/providers/pdal/qgspdalprovider.cpp
PeterPetrik: removed, we do not need it here at all

src/providers/pdal/qgspdaldataitems.cpp
PeterPetrik: added

src/core/providers/ept/qgseptdataitems.cpp
PeterPetrik: ok, lets do after merge (if needed -> see comment below)

src/providers/pdal/qgspdalprovider.cpp
nyalldawson: I still think we can avoid two seperate source select widgets. I'm thinking of an approach like what is done for vector layers now, where format specific options are shown after picking the input file. But I'll do handle this after this pr is merged, I've got a good idea of exactly how to handle this.

PeterPetrik: great!

PR #39781, opened 2020-11-03
[Oracle] Fix generated columns when evaluate default value is on

troopa81:
## Description

We must not evaluate generated columns when the option *Evaluate default value on provider side* is checked, else any feature creation will fail.

### DIFF
--- a/src/providers/oracle/qgsoracleprovider.cpp
+++ b/src/providers/oracle/qgsoracleprovider.cpp
@@ -1214,7 +1214,7 @@ QVariant QgsOracleProvider::defaultValue( int fieldId ) const
 QString QgsOracleProvider::defaultValueClause( int fieldId ) const
 {
   QString defVal = mDefaultValues.value( fieldId, QString() ).toString();
-  bool isGenerated = mAlwaysGenerated.value( fieldId, false );
+  const bool isGenerated = mAlwaysGenerated.value( fieldId, false );
 
   if ( isGenerated )
   {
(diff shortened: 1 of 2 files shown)


src/providers/oracle/qgsoracleprovider.cpp
```diff
@@ -1212,6 +1213,12 @@ QVariant QgsOracleProvider::defaultValue( int fieldId ) const
 QString QgsOracleProvider::defaultValueClause( int fieldId ) const
 {
   QString defVal = mDefaultValues.value( fieldId, QString() ).toString();
+  bool isGenerated = mAlwaysGenerated.value( fieldId, false );
```
elpaso, line 1216:
```suggestion
  const bool isGenerated = mAlwaysGenerated.value( fieldId, false );
```

tests/src/python/providertestbase.py
```diff
@@ -1206,3 +1208,19 @@ def testGeneratedColumns(self):
...
+
+        # Test insertion with default value evaluation on provider side to be sure
+        # it doesn't fail generated columns
+        vl.dataProvider().setProviderProperty(QgsDataProvider.EvaluateDefaultValues, True)
+
+        vl.startEditing()
+        feature = QgsVectorLayerUtils.createFeature(vl, QgsGeometry(), {0: 8})
+        vl.addFeature(feature)
+        self.assertTrue(feature.id() < 0)
+        # to be fixed
```
elpaso, line 1220: ?

troopa81: It's related to #38032 . Management of read-only fields has to be improved to uncomment this line. I propose to wait for  #38032 to be merged and I'll rebase/uncomment this line.

And I'll backport manually without it in 3.16

troopa81:
I rebased the PR, and everything is now ready to be merged

PR #39793, opened 2020-11-04
Optimise ept point cloud decoding

nyalldawson:
Add a bunch of optimisations to QgsPointCloudAttribute handling to optimise ept point cloud decoding.

### DIFF
--- a/src/core/pointcloud/qgspointcloudattribute.cpp
+++ b/src/core/pointcloud/qgspointcloudattribute.cpp
@@ -27,21 +27,6 @@ QgsPointCloudAttribute::QgsPointCloudAttribute( const QString &name, DataType ty
   updateSize();
 }
 
-QString QgsPointCloudAttribute::name() const
-{
-  return mName;
-}
-
-size_t QgsPointCloudAttribute::size() const
-{
-  return mSize;
-}
-
-QgsPointCloudAttribute::DataType QgsPointCloudAttribute::type() const
-{
-  return mType;
-}
-
 void QgsPointCloudAttribute::updateSize()
 {
   switch ( mType )
--- a/src/core/pointcloud/qgspointcloudattribute.cpp
+++ b/src/core/pointcloud/qgspointcloudattribute.cpp
@@ -69,18 +69,17 @@ void QgsPointCloudAttribute::updateSize()
 QgsPointCloudAttributeCollection::QgsPointCloudAttributeCollection() = default;
 
 QgsPointCloudAttributeCollection::QgsPointCloudAttributeCollection( const QVector<QgsPointCloudAttribute> &attributes )
-  : mAttributes( attributes )
 {
-  for ( int i = 0; i < mAttributes.size(); ++i )
+  mAttributes.reserve( attributes.size() );
+  for ( const QgsPointCloudAttribute &attribute : attributes )
   {
-    const QgsPointCloudAttribute &attribute = mAttributes.at( i );
-    mSize += attribute.size();
+    push_back( attribute );
   }
-
 }
 
 void QgsPointCloudAttributeCollection::push_back( const QgsPointCloudAttribute &attribute )
 {
+  mCachedAttributes.insert( attribute.name(), CachedAttributeData( mAttributes.size(), mSize ) );
   mAttributes.push_back( attribute );
   mSize += attribute.size();
 }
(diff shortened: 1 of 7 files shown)


nyalldawson:
Ping @PeterPetrik

src/core/pointcloud/qgspointcloudattribute.cpp
```diff
@@ -69,18 +54,17 @@ void QgsPointCloudAttribute::updateSize()
...
-  : mAttributes( attributes )
 {
-  for ( int i = 0; i < mAttributes.size(); ++i )
+  mAttributes.reserve( attributes.size() );
+  for ( const QgsPointCloudAttribute &attribute : attributes )
   {
-    const QgsPointCloudAttribute &attribute = mAttributes.at( i );
-    mSize += attribute.size();
+    push_back( attribute );
   }
```
PeterPetrik, line 62: you need to initialize mCachedAttributes and mSize in this constructor too

src/core/pointcloud/qgspointcloudattribute.h
```diff
@@ -96,11 +96,23 @@ class CORE_EXPORT QgsPointCloudAttributeCollection
     const QgsPointCloudAttribute *find( const QString &attributeName, int &offset ) const;
 
     //! Returns total size of record
-    int pointRecordSize() const { return mSize; }
+    std::size_t pointRecordSize() const { return mSize; }
```
PeterPetrik, line 99: not a fan... we work with qt containers that all work on int indices (https://doc.qt.io/qt-5/qvector.html#size). not sure if we win anything here by using site_t?

src/core/providers/ept/qgseptdecoder.cpp
```diff
@@ -121,37 +121,67 @@ bool _serialize( QByteArray &data, size_t outputPosition, const QgsPointCloudAtt
...
+    {}
+
+    int inputOffset;
+    int inputSize;
+    QgsPointCloudAttribute::DataType inputType;
+    int requestedSize;
+    QgsPointCloudAttribute::DataType requestedType;
+  };
+
+  std::vector< AttributeData > attributeData;
```
PeterPetrik, line 153: qvector?

src/core/pointcloud/qgspointcloudattribute.cpp
nyalldawson: That should be handled inside push_back, right?

src/core/pointcloud/qgspointcloudattribute.h
nyalldawson: Qt 6 compatibility for one :P

nyalldawson: But the actual rationale here was to match the type of QgsPointCloudAttribute::size() -- if we mix it gets annoying to deal with things which are sensitive to the type, like the QCOMPARE macro (which breaks the build on travis for some reason and doesn't auto cast to int)

src/core/providers/ept/qgseptdecoder.cpp
nyalldawson: Since it's only internal, I trust the compiler here to better optimise this vs Qt :laugh: . Also means we don't incur the cost of some qt container checks which are expensive on debug builds. They are skipped/optimised out on release builds, but I almost always use debug builds and want faster stuff anyway!

src/core/pointcloud/qgspointcloudattribute.cpp
PeterPetrik: of course, misread it 🤦

src/core/providers/ept/qgseptdecoder.cpp
PeterPetrik: micro-optimization? :)

PeterPetrik (approved):

src/core/providers/ept/qgseptdecoder.cpp
nyalldawson: Maybe... but it **did** make a difference in my test builds!

Read the upstream discussion around the qt containers. They are disliked by the qt developers and I gather that they'd happily remove them all and force everyone to use std containers.

I'm 50/50 on this. I prefer std containers wherever possible, but the qt containers have two nice things which we use in qgis a lot:
1. implicit sharing
2. sip/python compatibility

That's why my general rule of thumb is (in the absence of all other factors) use std:: for private stuff, use Qt for public stuff.

(lastly -- qhash is often better then the std container, and will be even better on qt 6 builds. There's no easy answers!)

PR #39802, opened 2020-11-04
cmake: Catch missing sip5 binary

marisn:
I somehow can not compile QGIS on my new Ubuntu 20.10 box due to messed up dependencies. Turns out it is possible to pass SIP detection test when sip5 binary is missing. It gets set to None and CMake is OK with it. This PR prevents reporting SIP found if sip5 binary is missing.

### DIFF
--- a/cmake/FindSIP.py
+++ b/cmake/FindSIP.py
@@ -30,6 +30,8 @@
 # Redistribution and use is allowed according to the terms of the BSD license.
 # For details see the accompanying COPYING-CMAKE-SCRIPTS file.
 
+import sys
+
 try:
     import sipbuild
 except ImportError:  # Code for SIP v4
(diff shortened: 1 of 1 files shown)


cmake/FindSIP.py
```diff
@@ -48,12 +50,15 @@
     else:
         print("sip_module_dir:%s" % sipcfg.sip_mod_dir)
 else:  # Code for SIP v5
+    import shutil
+    sip_bin = shutil.which("sip5")
+    if not sip_bin:
+        sys.exit(1)
```
m-kuhn, line 56: Would it make sense to print an error message here or is that covered already by cmake?
Can sip_bin be manually specified somewhere as a cmake variable? This code assumes it's on system path.

m-kuhn:
Thanks for this !

cmake/FindSIP.py
marisn: Missing sip5 will result in an error "SIP not found". It could be more verbose but then it has to be changed in FindSIP.cmake. FindSIP.cmake is doing really lousy work on detecting if values coming back from FindSIP.py make any sense at all thus sys.exit(1) is the only safe option. Fiddling with cmake files is a bit out of my league and I lack free time to learn cmake.
Threat this PR not as an end but as a step into right direction.

m-kuhn: Just wondering if a simple print or print to stderr in python would help?

And for the other question it's likely that this script is not even called if a sip binary has already been specified, so no worries, that works for me.

nyalldawson:

PR #39810, opened 2020-11-05
Point cloud UX improvements

nyalldawson:
A bunch of UX improvements, and general nicer handling of things via the provider registry instead of hardcoded logic.

Including:

- Only expose a single generic "Point Cloud" tab in data source manager, and use provider registry methods to auto select the appropriate provider to use for a file
- Hide a bunch of unwanted files from browser, such as the broken ogr entries for "ept.json" files and the internal "ept-build.json" files.


nyalldawson:
Ping @PeterPetrik

cmake/FindPDAL.cmake
```diff
@@ -17,28 +17,26 @@
...
+
 FIND_PATH(PDAL_INCLUDE_DIR pdal.hpp
-  "$ENV{LIB_DIR}/include/"
+  "$ENV{LIB_DIR}/include"
   "$ENV{LIB_DIR}/include/pdal"
+  "/usr/include/pdal"
   c:/msys/local/include
   NO_DEFAULT_PATH
   )
-FIND_PATH(PDAL_INCLUDE_DIR pdal.hpp)
```
PeterPetrik, line 26: I would keep it, see comment on lines 14-19. (we have the same in other Find* files)

src/core/qgsfileutils.h
```diff
@@ -46,6 +46,14 @@ class CORE_EXPORT QgsFileUtils
      */
     static QStringList extensionsFromFilter( const QString &filter );
 
+    /**
+     * Giiven a \a filter string like "GeoTIFF Files (*.tiff *.tif)", extracts
```
PeterPetrik, line 50: Given

PeterPetrik (approved):
I suggest to merge and revert the FindPDAL on master?

cmake/FindPDAL.cmake
nyalldawson: Ok, reverted

src/core/qgsfileutils.h
nyalldawson: FIxed

PR #39812, opened 2020-11-05
traverse tree in 2D rendering by map error

PeterPetrik:
Determines how many nodes level to load in 2D renderer based on screen error. Same approach that is already implemented in 3D renderer.

progressive_2d_renderer

see https://github.com/wonder-sk/point-cloud-experiments/issues/15

### DIFF
--- a/src/core/pointcloud/qgspointcloudrenderer.cpp
+++ b/src/core/pointcloud/qgspointcloudrenderer.cpp
@@ -87,26 +87,28 @@ void QgsPointCloudRendererConfig::setColorRamp( const QgsColorRamp *value )
   mColorRamp.reset( value->clone() );
 }
 
+float QgsPointCloudRendererConfig::maximumScreenError() const
+{
+  return mMaximumScreenError;
+}
+
 ///@endcond
 
 QgsPointCloudLayerRenderer::QgsPointCloudLayerRenderer( QgsPointCloudLayer *layer, QgsRenderContext &context )
   : QgsMapLayerRenderer( layer->id(), &context )
   , mLayer( layer )
 {
-  // TODO: we must not keep pointer to mLayer (it's dangerous) - we must copy anything we need for rendering
-  // or use some locking to prevent read/write from multiple threads
-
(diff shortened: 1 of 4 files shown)


PeterPetrik:
ping @nyalldawson

nyalldawson:
Looks good to me

src/core/pointcloud/qgspointcloudrenderer.h
```diff
@@ -71,10 +71,14 @@ class CORE_EXPORT QgsPointCloudRendererConfig
...
     void setColorRamp( const QgsColorRamp *value );
 
+    //! Returns maximum allowed screen error in pixels
+    float maximumScreenError() const;
+
   private:
     double mZMin = 0, mZMax = 0;
     int mPenWidth = 1;
     std::unique_ptr<QgsColorRamp> mColorRamp;
+    float mMaximumScreenError = 5;
```
wonder-sk, line 81: let's set it to 1px

src/core/pointcloud/qgspointcloudrenderer.cpp
```diff
@@ -159,6 +169,37 @@ bool QgsPointCloudLayerRenderer::render()
...
+    float maxErrorPixels,
+    float nodeErrorPixels )
+{
+  QList<IndexedPointCloudNode> nodes;
+
+  if ( context.renderingStopped() )
+  {
+    qDebug() << "canceled";
+    return nodes;
+  }
```
wonder-sk, line 184: no need to pass context to this method and check context.renderingStopped() here - the traversal is very fast

wonder-sk (approved):

nyalldawson:
@PeterPetrik I think something is not working quite right here. Checkout what happens when I do tiny zoom increments in/out of this cloud:

Peek 2020-11-06 11-24

The tile in the bottom left is showing at a lower level of detail then its neighbours, and it's only when I zoom in a bit closer that it kicks in with the correct amount of detail. Any ideas what could be causing this?

Issue #34868, opened 2020-03-04
Respect image orientation in layout(picture), HTML widget, map tips


Houska1:
QGIS display of photos (via an attribute referencing a file or URL) does not reflect the image's EXIF orientation tag in several places:

1. HTML widget in forms (it **is** reflected in the QML widget--thank you!)
2. Map tips
3. Picture item in a layout

EXIF orientation is of course a bugbear in HTML, with at best dubious normal browser support, and images intended for web use are best saved pre-rotated. However, given the increasing integration of QGIS with fieldwork, where photos with cameras that use the orientation tag are pretty typical, out-of-the-box support would be helpful.

I'm guessing the places orientation is reflected are where QGIS uses `QImageReader` with `setAutoTransform=true` as opposed to an HTML engine rendering an `<img>` . So I'm hoping fixing #3 above might be easy, while #1 and #2 are likely harder?

(There are, of course, workarounds, but they're a hassle -- for now I'm extracting the Orientation tag as a separate attribute in the layer and then using data-driven rotations and resizing to fit inside the specified rectangle in a layout.)

A prior commit that seems to have fixed this for the attachment widget is https://github.com/qgis/QGIS/commit/574a6549f8d929c885b531fce827993c26fb034d

nyalldawson:
Yes - fixing the picture item should be relatively straightforward, using the same approach as 574a654. As you've rightly guessed, the HTML based widgets are a lot trickier. Looks like adding custom js to your html to handle this is the only viable approach...

davidafuller01:
+1

pellegoeg:
Hi,

I have made a fix for this issue (my first pull request, so hopefully I have done the right job!).

Its adding a extra attribute to the photo table with exif orientation.

The could then be rotated in maptips by using som case when eg:

`<img src="file:///[% photo %]" width="350" height="250"
style="[% CASE
WHEN orientation = 2 THEN 'transform: scaleX(-1);'
WHEN orientation = 3 THEN 'transform:rotate(180deg);'
WHEN orientation = 4 THEN 'transform:rotate(180deg) scaleX(-1);'
WHEN orientation = 5 THEN 'transform:rotate(-90deg) scaleX(-1);'
WHEN orientation = 6 THEN 'transform:rotate(90deg);'
WHEN orientation = 7 THEN 'transform:rotate(90deg) scaleX(-1);'
WHEN orientation = 8 THEN 'transform:rotate(270deg);'
ELSE ''
END %]" />
[...]

PR #39817, opened 2020-11-05
Extend import geotagged photos to include exif_orientation

pellegoeg:
## Description

In the danish user group, the missing image rotation have been a topic. 

 "Import geotagged photos" is updated so the exif_orientation is extracted to the output table.

This will make it posible to get the rotation in maptips and other html widgets eg:

``

I have added at 8 new test fotos to the repos showing the letter F as used to descripe the orientation in exif documentation like: https://www.impulseadventure.com/photo/exif-orientation.html

This fixes 1 and 2 in issue: #34868


pellegoeg:
Test failed, im fixing the code in a couple of hours

nyalldawson:
Thanks @pellegoeg ! Given that the new files aren't actually being used in tests, can you remove these? May as well keep the git repo size as small as possible :smile: 

One thing I'm wondering -- do you think it would be worthwhile also adding a "rotation" field here, where we fill this in based on the orientation value? It would be easier for users that way, as they wouldn't have to lookup what the raw orientation codes mean. (I realise we can't handle flipped images by a rotation alone, but I think those would be extremely rare anyway!)

pellegoeg:
Thank you for the comment @nyalldawson 

I actually started with the solution you mention by converting the exif_orientation code to a rotation, but thought, maybe there is some specialised use cases where they needed the flipped orientation and the qgis projected wanted a more true præsentation of the actual metadata?

In my latest commit, I have cleaned up the test images and I have converted the orientation to a rotation.

I have attached the photos so It can be manualy testet in this pull request.
[orientation_photos.zip](https://github.com/qgis/QGIS/files/5499873/orientation_photos.zip)

pellegoeg:
and the code for maptips is now:

`

pellegoeg:
Hi @nyalldawson 

Is there some work I have forgot to do, or should I wait for you or another maintainer to accept the pull request?

nyalldawson (approved):

nyalldawson:
Looks great, thank you!

PR #39819, opened 2020-11-05
Fix server virtual layers load

elpaso:
Another chapter of the tragic https://github.com/qgis/QGIS/pull/38488 saga.

This happens because virtual layers ctor calls reloadData which calls project::instance but on server it just calls the wrong instance (because initially is nullptr and on subsequent calls it might be and instance of a different project).

### DIFF
--- a/src/server/qgsconfigcache.cpp
+++ b/src/server/qgsconfigcache.cpp
@@ -43,7 +43,11 @@ const QgsProject *QgsConfigCache::project( const QString &path, QgsServerSetting
 {
   if ( ! mProjectCache[ path ] )
   {
+
     std::unique_ptr<QgsProject> prj( new QgsProject() );
+
+    // This is required by virtual layers that call QgsProject::instance() inside the constructor :(
+    QgsProject::setInstance( prj.get() );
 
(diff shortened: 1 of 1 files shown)


nyalldawson:
@elpaso 

Server specifics aside, what do you think is a good solution to this? Potentially we could add the QgsProject reference to QgsDataProvider::ProviderOptions so that the provider can access the actual project it's associated with. I think that's probably the cleanest approach we'll find, but I'm open for other ideas!

m-kuhn:
Isn't this very specific to virtual layers?

elpaso:
@m-kuhn AFAIK it's virtual layers only, the other solution would be to use lazy loading of dependencies inside the virtual layer provider, I see it's implemented but only used for tasks. I don't know if it could work in the the general case.

In general I think we should avoid accessing the project from within a data provider, I understand the special case of virtual layers.

Anyway, I would recommend to merge and backport this urgent patch immediately because it breaks virtual layers on LTR server  while we think about how to reduce our dependency on project instances in general.

m-kuhn:
@elpaso Agreed on all points :+1:

PR #39820, opened 2020-11-06
PDAL provider fixes and improvements

nyalldawson:
1. Report PDAL version information in qgis about screen and processing debug logs (for future point cloud processing algorithms :wink: )
2. Add nice flexible API to QgsProviderRegistry/QgsProviderMetadata so that in the absence of all other information we can pick the best provider and layer type to use to open a specific URI
3. Use this API to handle opening point cloud layers from drag and dropped files, allowing las/laz files to be opened by dropping them onto QGIS.


nyalldawson:
Ping @PeterPetrik 

Also ping @elpaso -- this has some more provider registry changes you may find interesting

PeterPetrik (approved):

roya0045:
With the extra funding, were the processing related elements planned to be released in 3.18 or was that later?

PR #39827, opened 2020-11-06
Add WITH_EPT optional cmake switch

nyalldawson:
(includes https://github.com/qgis/QGIS/pull/39820)

Switch from silently building broken Entwine point cloud support when dependancies are missing (libzstd, lazperf) to an explicit
WITH_EPT optional cmake switch which requires those dependancies

This makes it easier people to determine whether they have the complete set of necessary dependancies to build Entwine point
cloud support, as cmake will show an error if they enable WITH_EPT but don't have them...


nyalldawson:
@PeterPetrik mind having a look over https://github.com/qgis/QGIS/pull/39827/commits/333d5f495961d6db9093819957dc05abd73f9ea7 ?

wonder-sk:
:+1:  Looks good to me.

Originally we thought EPT would be a required provider, that's why it ended up in src/core/providers. Shall we also move it to src/providers or are you happy with it being a part of qgis_core?

nyalldawson:
Let's leave it... It's not doing any harm there 😆

PeterPetrik (approved):

PR #39846, opened 2020-11-06
Allow loading oracle layers from the browser

### LINKED ISSUE #39759: [Oracle] Cannot load any oracle layer
**Describe the bug**

There is the following error when loading an oracle layer 

```
Layer is not valid : The layer dbname='XEPDB1' host=localhost port=1521 user='QGIS' password='qgis' srid=5698 type=Polygon25D table="QGIS"."POLY_DATA" (GEOM) is not a valid layer and can not be added to the map. Reason: Invalid provider
```

**How to Reproduce**

1. Create an oracle connection
2. Double click on a table from the browser
3. the error message appears, no layer is loaded

**QGIS and OS versions**

QGIS master
Debian testing

**Additional context**

This is due to #39656. It's because not all providers implement [correctly decodeUri method](https://github.com/qgis/QGIS/blob/master/src/core/qgsprovidermetadata.h#L332), like Oracle for instance. I don't know if Oracle is the only one not implementing it. Should we just implement it in Oracle or deal with the case where the provider doesn't implement it?

cc @rduivenvoorde

rduivenvoorde:
See #39759

This commit fixes the absence of encodeUri and decodeUri which made loading an Oracle layer via the browser fail.

Also loading a layer via iface.addVectorLayer is now working again. Eg if you start the Oslandia test  Oracle db in a docker with the QGIS test data (see issue) the following works:

```
iface.addVectorLayer('dbname=\'XEPDB1\' host=localhost port=1521 user=\'QGIS\' password=\'qgis\' key=\'pk\' estimatedmetadata=true srid=4326 type=Polygon table="QGIS"."SOME_POLY_DATA" (GEOM)', 'oraclelayer', 'oracle')
```

Some remarks:

- Some 'fields' seem not to be used (yet?) by the OracleProvider: `selectatid`, `sslmode`. I kept these for now.

- Some other providers, notably db2, gpx, geonode and grass also do not implement  encodeUri and decodeUri yet... I could try do gpx here also, as that is easy to test. But for the others I would need assistance. 

I also see that loading an (in 3.16) valid WFS layer, does not load in current master:
```
iface.addVectorLayer(
"https://geodata.nationaalgeoregister.nl/bestuurlijkegrenzen/wfs?SERVICE=WFS&VERSION=1.0.0&REQUEST=GetFeature&TYPENAME=bestuurlijkegrenzen:provincies&SRSNAME=EPSG:28992",
"wfs example",
"WFS")
```

I will now investigate if this is the same issue...

### DIFF
--- a/src/providers/oracle/qgsoracleprovider.cpp
+++ b/src/providers/oracle/qgsoracleprovider.cpp
@@ -3820,6 +3820,8 @@ QVariantMap QgsOracleProviderMetadata::decodeUri( const QString &uri ) const
     uriParts[ QStringLiteral( "port" ) ] = dsUri.port();
   if ( ! dsUri.service().isEmpty() )
     uriParts[ QStringLiteral( "service" ) ] = dsUri.service();
+  if ( ! dsUri.param( "dbworkspace" ).isEmpty() )
+    uriParts[ QStringLiteral( "dbworkspace" ) ] = dsUri.param( "dbworkspace" );
   if ( ! dsUri.username().isEmpty() )
     uriParts[ QStringLiteral( "username" ) ] = dsUri.username();
   if ( ! dsUri.password().isEmpty() )
@@ -3828,8 +3830,6 @@ QVariantMap QgsOracleProviderMetadata::decodeUri( const QString &uri ) const
     uriParts[ QStringLiteral( "authcfg" ) ] = dsUri.authConfigId();
   if ( dsUri.wkbType() != QgsWkbTypes::Type::Unknown )
     uriParts[ QStringLiteral( "type" ) ] = dsUri.wkbType();
-  if ( uri.contains( QStringLiteral( "selectatid=" ), Qt::CaseSensitivity::CaseInsensitive ) )
-    uriParts[ QStringLiteral( "selectatid" ) ] = ! dsUri.selectAtIdDisabled();
   if ( ! dsUri.table().isEmpty() )
     uriParts[ QStringLiteral( "table" ) ] = dsUri.table();
   if ( ! dsUri.schema().isEmpty() )
@@ -3840,10 +3840,12 @@ QVariantMap QgsOracleProviderMetadata::decodeUri( const QString &uri ) const
     uriParts[ QStringLiteral( "srid" ) ] = dsUri.srid();
   if ( uri.contains( QStringLiteral( "estimatedmetadata=" ), Qt::CaseSensitivity::CaseInsensitive ) )
     uriParts[ QStringLiteral( "estimatedmetadata" ) ] = dsUri.useEstimatedMetadata();
-  if ( uri.contains( QStringLiteral( "sslmode=" ), Qt::CaseSensitivity::CaseInsensitive ) )
-    uriParts[ QStringLiteral( "sslmode" ) ] = dsUri.sslMode();
+  if ( ! dsUri.param( "includegeoattributes" ).isEmpty() )
+    uriParts[ QStringLiteral( "includegeoattributes" ) ] = dsUri.param( "includegeoattributes" );
   if ( ! dsUri.sql().isEmpty() )
     uriParts[ QStringLiteral( "sql" ) ] = dsUri.sql();
+  if ( ! dsUri.param( "checkPrimaryKeyUnicity" ).isEmpty() )
+    uriParts[ QStringLiteral( "checkPrimaryKeyUnicity" ) ] = dsUri.param( "checkPrimaryKeyUnicity" );
   if ( ! dsUri.geometryColumn().isEmpty() )
     uriParts[ QStringLiteral( "geometrycolumn" ) ] = dsUri.geometryColumn();
   return uriParts;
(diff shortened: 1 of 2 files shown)


src/providers/oracle/qgsoracleprovider.cpp
```diff
@@ -3807,4 +3807,88 @@ QGISEXTERN QgsProviderGuiMetadata *providerGuiMetadataFactory()
...
+    uriParts[ QStringLiteral( "selectatid" ) ] = ! dsUri.selectAtIdDisabled();
+  if ( ! dsUri.table().isEmpty() )
+    uriParts[ QStringLiteral( "table" ) ] = dsUri.table();
+  if ( ! dsUri.schema().isEmpty() )
+    uriParts[ QStringLiteral( "schema" ) ] = dsUri.schema();
+  if ( ! dsUri.keyColumn().isEmpty() )
+    uriParts[ QStringLiteral( "key" ) ] = dsUri.keyColumn();
+  if ( ! dsUri.srid().isEmpty() )
+    uriParts[ QStringLiteral( "srid" ) ] = dsUri.srid();
+  if ( uri.contains( QStringLiteral( "estimatedmetadata=" ), Qt::CaseSensitivity::CaseInsensitive ) )
```
troopa81, line 3841: Out of curiosity, do you know why we just don't use `dsUri.useEstimatedMetadata().isEmpty()` like other condition?
It look like we want to not care about case but in the end we still get `dsUri.useEstimatedMetadata()`. 

I noticed it's the same in postgres.

troopa81, line 3841: Out of curiosity, do you know why we just don't use dsUri.useEstimatedMetadata().isEmpty() like other condition?
It look like we want to not care about case but in the end we still get dsUri.useEstimatedMetadata().

I noticed it's the same in postgres.

troopa81 (changes requested):
Thanks for taking care of this.

I think some fields are missing: 
- [includegeoattributes](https://github.com/qgis/QGIS/blob/master/src/providers/oracle/qgsoracleprovider.cpp#L88) 
- [workspace](https://github.com/qgis/QGIS/blob/master/src/providers/oracle/qgsoracleprovider.cpp#L236)

> Some 'fields' seem not to be used (yet?) by the OracleProvider: selectatid, sslmode. I kept these for now.

I'm not sure this really interesting to keep the code to deal with this parameter while they are actually not managed the provider

src/providers/oracle/qgsoracleprovider.cpp
rduivenvoorde: @troopa81 I do not have a very clear conceptual view of the need of the encodeUri and decodeUri. But to me it looked like that we were actually 'marshalling' an uri to a QgsDataSourceUri (https://qgis.org/api/qgsdatasourceuri_8h.html) and viceversa. Because postgis kept all members of that class I though to let Oracle also keep all.
And I did not see the generic setParam yet... so the qgsdatasource can actually hold a whole set of custom params/key-value pairs.
I will add those, and I will comment the not used (just to be clear?). Thanks for checking!
Mmm probably I should also add that to postgis provider?

You question about "why we just don't use dsUri.useEstimatedMetadata().isEmpty()" I do not know, maybe has something todo with True vs true (python vs cpp?)? I tried git blame but could not find info about it.

rduivenvoorde: Mmm, creating an Oracle connection, and then looking into the project file to check the attributes of the Oracle (*connection*?) element I get the following items:
```
name="docker_localhost" 
host="localhost" 
port="1521" 
schema="" 
username="QGIS" 
password="qgis" 
saveUsername="true" 
savePassword="true" 
allowGeometrylessTables="true" 
userTablesOnly="true" 
database="XEPDB1" 
dbworkspace="" 
geometryColumnsOnly="true" 
estimatedMetadata="true" 
dboptions=""
```
If I create a layer from that connection and request the uri via python:
```
iface.mapCanvas().currentLayer().dataProvider().uri().uri()
```
I get (both in 3.16 and in current master)
```
'dbname=\'XEPDB1\' 
host=localhost 
port=1521 
user=\'QGIS\' 
password=\'qgis\' 
key=\'pk\' 
estimatedmetadata=true 
srid=4326 
type=Polygon 
table="QGIS"."SOME_POLY_DATA" (GEOM)'
```
I'm a little puzzled about what I can/should put in the QgsDataSourceUri now. I also see that QgsDataSourceUri itself as an encodedUri and setEncodedUri: https://github.com/qgis/QGIS/blob/master/src/core/qgsdatasourceuri.cpp#L610-L659
THAT one only returns the username password of the connection?

I removed the selectatid and sslmode, added dbworkspace and includegeoattributes (and checkPrimaryKeyUnicity to one side to make it symmetric)

troopa81:
> I'm a little puzzled about what I can/should put in the QgsDataSourceUri now.

I think the *decodeUri* has to deal with parameters known by the provider, no more, no less. So the way it is now looks fine to me. 

I see you add 2 more fixes (delete from browser, and the CHAR issue), do you mind creating another PR for those (and add non regression unit test for this 2 issues), so we can merge this PR and continue discussing the other.

troopa81 (approved):
Thansks for the modifications. LGTM

rduivenvoorde:
@troopa81 can you merge? Or should I ask Nyall or Matthias?

troopa81:
@rduivenvoorde I don't have merge rights. @nyalldawson @m-kuhn Is that OK for you?

m-kuhn:
Thanks @rduivenvoorde and @troopa81

PR #39859, opened 2020-11-06
Fix random raster Float32/64 support

### LINKED ISSUE #39845: Create random raster layer (uniform distribution) does not create pixels with decimal values
@root676 

I'm not sure if this is a bug report or a feature request, but I can't create a random raster layer, for instance, with values from 0.0 to 1.0, even choosing Float32 or Float64 data types. 

Also, following the default behaviour (By default, the values will range between the minimum and maximum value of the specified output raster type), if I choose Byte and keep min and max bounds as zero, the output is a raster will all pixels with zero.

So, it seems some kind of a bug.

root676:
## Description
This PR fixes #39845. The Uniform random raster algorithm was not able to produce double values due to a mis-assignment of variables. Also, using the min and max values for the respective data type of the raster didn't take effect due mis-assignment. Both issues were fixed.

Again thanks to @PedroVenancio for reporting this! :+1: 

Fixes #39845


nyalldawson (approved):

PedroVenancio:
Impressive the quick fix!! Thank you very much @root676 !

PR #39861, opened 2020-11-06
Embed a copy of laz-perf, and use if system laz-perf is not found

nyalldawson:
Laz-perf is a relatively "underground" library, and is not packaged
for many major distributions. Its also tiny and trivial to include with
the other external libraries.

This avoids requiring users who want point cloud support to have to
compile their own laz-perf


nyalldawson:
ping @PeterPetrik

PeterPetrik (approved):

